PR #1750, opened 2019-04-09
minor: limit usage of ccxt to freqtrade/exchange only

hroff-1902:
This PR limits usage (import) of ccxt to the freqtrade/exchange module only, for better incapsulation and isolation of the features.

### DIFF
--- a/freqtrade/exchange/__init__.py
+++ b/freqtrade/exchange/__init__.py
@@ -1,3 +1,8 @@
 from freqtrade.exchange.exchange import Exchange  # noqa: F401
+from freqtrade.exchange.exchange import (is_exchange_supported,  # noqa: F401
+                                         supported_exchanges)
+from freqtrade.exchange.exchange import (timeframe_to_seconds,  # noqa: F401
+                                         timeframe_to_minutes,
+                                         timeframe_to_msecs)
 from freqtrade.exchange.kraken import Kraken  # noqa: F401
 from freqtrade.exchange.binance import Binance  # noqa: F401
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -697,9 +697,8 @@ def is_exchange_supported(exchange: str, ccxt_module=None) -> bool:
     return exchange in supported_exchanges(ccxt_module)
 
 
-def supported_exchanges(ccxt_module=None) -> str:
-    exchanges = ccxt_module.exchanges if ccxt_module is not None else ccxt.exchanges
-    return exchanges
+def supported_exchanges(ccxt_module=None) -> List[str]:
+    return ccxt_module.exchanges if ccxt_module is not None else ccxt.exchanges
 
 
 def timeframe_to_seconds(ticker_interval: str) -> int:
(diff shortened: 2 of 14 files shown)


hroff-1902:
Minor fix: Type annotation in supported_exchanged() corrected.

Sometimes I hate python for its free typing. Flake and mypy also kept silence cause ccxt does not annotate types.

freqtrade/exchange/__init__.py
```diff
@@ -1,3 +1,8 @@
 from freqtrade.exchange.exchange import Exchange  # noqa: F401
+from freqtrade.exchange.exchange import (is_exchange_supported,  # noqa: F401
+                                         supported_exchanges)
+from freqtrade.exchange.exchange import (timeframe_to_seconds,  # noqa: F401
```
xmatthias: Why don't you combine these 2 imports ... ? they are multiline anyway, so you add them within the same set of parenteses...

freqtrade/exchange/exchange.py
```diff
@@ -138,7 +139,7 @@ def _init_ccxt(self, exchange_config: dict, ccxt_module=ccxt,
         # Find matching class for the given exchange name
         name = exchange_config['name']
 
-        if name not in ccxt_module.exchanges:
+        if not is_exchange_supported(name, ccxt_module):
```
xmatthias: I don't think passing in ccxt_module makes too much sense. 

i see that it's been there before - however ccxt supports the same exchanges in sync and async mode - so it either fails for both - or succeeds for both possible types of ccxt_module

scripts/get_market_pairs.py
```diff
@@ -1,6 +1,8 @@
 import os
 import sys
 
+from freqtrade.exchange import is_exchange_supported, supported_exchanges
```
xmatthias: I'm not certain on changing this script...
While the improvement might make sense - it's really a 1:1 copy of 
 one of the examples in ccxt [here](https://github.com/ccxt/ccxt/blob/master/examples/py/arbitrage-pairs.py) - which (i notice now) we should mention at the top...

freqtrade/exchange/__init__.py
hroff-1902: Errhm, "for historical reasons". ;-) They (those two sets of functions) came in two different commits, from two different reasons...

freqtrade/exchange/exchange.py
hroff-1902: Yeah, but taking ccxt.exchanges in the init of async part that utilizes ccxt_async instead of ccxt instance seems for me unsymmetric so I did it (left as it was, only reformulated) on this way only for this reason. Theoretically ccxt and ccxt_async may return different lists.

scripts/get_market_pairs.py
hroff-1902: Did not know of that example in ccxt repo. Well, the target was to incapsulate (limit) importing of ccxt to the exchange module only.

hroff-1902: What do you think -- should we remove changes made in this script (since it's simply borrowed from external sources) and apply the rest?

xmatthias: Yes, sounds like a plan (sorry, forgot about this ...)

PR #1771, opened 2019-04-16
Enable ratelimit

### LINKED ISSUE #1764: enableRateLimit should be switched to true by default in the config and defaults
In the default config use
```
"ccxt_async_config": {
            "enableRateLimit": true
        },
```
... And probably in the default value.

Reason:
With current config and defaults Bittrex blocks new inexperienced users who just started evaluation of the bot and use default config for that.

xmatthias:
## Summary
Enable rate-limit by default so new users don't encounter bans and document this behaviour.

Resolves #1764

## Quick changelog

- Enable rate-limit by default 
- Add documentation for this parameter
- remove deprecated parameter for setting rate-limit.

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -30,7 +30,8 @@
         "secret": "your_exchange_secret",
         "ccxt_config": {"enableRateLimit": true},
         "ccxt_async_config": {
-            "enableRateLimit": false
+            "enableRateLimit": true,
+            "rateLimit": 500
         },
         "pair_whitelist": [
             "ETH/BTC",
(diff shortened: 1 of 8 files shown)


hroff-1902:
TODO: move description of the deprecated option from configuration.md to deprecated.md

xmatthias:
it was deprecated but is now gone ... ?

hroff-1902:
> it was deprecated but is now gone ... ?

Hmm, seems I missed something...

xmatthias:
we used to have the option `exchange.ccxt_rate_limit` ... which set the rate-limit for the async exchange.
that was limiting, as ccxt has many more (sometimes exchange specific) options we wanted to expose.
we left this in for compatibility - but with this i removed that part.

Issue #1775, opened 2019-04-18
List of further hyperopt enhancements


hroff-1902:
TODOs, for discussion/evaluation, as I see it:

* [x] --print-all option which allows to see all hyperopt results in the optimization path, not only the best ones.
* [x] Add the --refresh-pairs option to hyperopt. Now it's available for backtesting and edge commands only.
* [x] -j N option to set the number of workers. With the same semantics as joblib.parallel uses: -1 means joblib will automaticaly define proper number of workers to utilize all CPUs.
* [ ] The following is probably the most important drawback in hyperopt: if you restart hyperopt, skopt does not use previously collected results and starts evaluation again from a random point in the parameters hyperspace. Implement fetching the last point from previous results as x0, y0 for skopt to allow it go on from same point it stopped before.
* [x] In addition to the previous question, an option (how to name it? --remove-data? --restart-evals?) can be added that forces skopt to forget, throw away all previous results and restart evals from scratches, again from a random point.
* [x] An option which allows to pass a particular starting point (random_state) to skopt for reproducible results, for testing/development.
* [ ] An option which duplicates results as csv in addition to pickling them. Maybe not all of them (including collected at previous runs), but only those that have been collected during current run of the hyperopt. That would allow to see all the parameters in the hyperspace, explore them, plot the curves, process the points using external utilities and so on...
* [x] An utility that fetches the pickle file with results already collected and export it to csv.
* [x] The debug log message for every evaluation printing current state of the optimization (attributes of the skopt.Optimizer)

-------------

* [x] Prevent freqbot from running multiple instances of hyperopt simultaneously,
* [x] The --min-trades=N option. The points in the optimization path that generate less than N option are to be considered "bad" (with maximal Loss value) and thus be excluded by the optimizer from the optimization path. This prevents the optimization path from descending to points in the hyperspace with small amount (1, 2) of trades. Sometimes (always with default hyperopt after several hundreds of optimization evals) the optimization path descends to the points with small amount of trades, starting optimization of the 'hodl' strategy when no trades (or only small amount of trades) appear instead of optimization of the real target strategy. The current implementation now excludes points with number of trades equal to 0 only.
* [x] --color/--no-color for --print-all. Print best records in green, 'bad' records (with maximal Loss set) in red.

-------------

* [x] Fix SumProfit% value printed by hyperopt (multiply it by 100) so that
`SumProfit% = N_trades * Avg Profit%`
Currently printed value is incorrect.
[...]

PR #1793, opened 2019-04-23
hyperopt: print optimizer state in debug log messages

hroff-1902:
A part of #1775.

Run the bot with the `-v` command line option to see the data.

Internal optimizer state and the points from the hyperspace ('params' in result dict) are printed in different loops since they have (may have) different order.

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -288,6 +288,9 @@ class Hyperopt(Backtesting):
                             'total_tries': self.total_tries,
                             'result': f_val[j]['result'],
                         })
+                        logger.debug(f"Optimizer params: {f_val[j]['params']}")
+                    for j in range(cpus):
+                        logger.debug(f"Opimizer state: Xi: {opt.Xi[-j-1]}, yi: {opt.yi[-j-1]}")
         except KeyboardInterrupt:
             print('User interrupted..')
 


freqtrade/optimize/hyperopt.py
```diff
@@ -288,6 +288,9 @@ def start(self) -> None:
                             'total_tries': self.total_tries,
                             'result': f_val[j]['result'],
                         })
+                        logger.debug(f"Optimizer params: {f_val[j]['params']}")
+                    for j in range(cpus):
+                        logger.debug(f"Opimizer state: Xi: {opt.Xi[-j-1]}, yi: {opt.yi[-j-1]}")
```
xmatthias: Maybe i'm missing something - but don't the 2 lines print the same thing in different formats (i prefer the "Opitmizer params:" one ...)?

Trying this - all values match (if you take only the values of the dict).


```
.2019-04-23 19:37:03,328 - freqtrade.optimize.hyperopt - DEBUG - Optimizer params: {'mfi-value': 12, 'fastd-value': 28, 'adx-value': 38, 'rsi-value': 37, 'mfi-enabled': False, 'fastd-enabled': False, 'adx-enabled': False, 'rsi-enabled': True, 'trigger': 'macd_cross_signal', 'sell-mfi-value': 76, 'sell-fastd-value': 84, 'sell-adx-value': 92, 'sell-rsi-value': 97, 'sell-mfi-enabled': True, 'sell-fastd-enabled': True, 'sell-adx-enabled': False, 'sell-rsi-enabled': True, 'sell-trigger': 'sell-bb_upper', 'roi_t1': 102, 'roi_t2': 29, 'roi_t3': 24, 'roi_p1': 0.03907810596868612, 'roi_p2': 0.017418899486970653, 'roi_p3': 0.1462967283445416, 'stoploss': -0.37168748310135113}

2019-04-23 19:37:03,328 - freqtrade.optimize.hyperopt - DEBUG - Opimizer state: Xi: [12, 28, 38, 37, False, False, False, True, 'macd_cross_signal', 76, 84, 92, 97, True, True, False, True, 'sell-bb_upper', 102, 29, 24, 0.03907810596868612, 0.017418899486970653, 0.1462967283445416, -0.37168748310135113], yi: 2.2217194636218585
```

The sequence of printing is exactly reversed ... but still there is no added value in printing it twice i think... (unless i am not seeing something )

hroff-1902: they are not the same: params contains names of the hyperopt space dimensions, while Xi does not; in the same time, second line contains more or less complete optimizer state: Xi and yi. There is also so called models and hessians, etc. but they require lot's of space on the screen and in the logs, so not included.
Yes, I can try to print them in the same order (they come as you see in different order, that's why I print them in two diff. loops). That's one possibility.
These are values that come from different sources: params is produced directly by our generate_optimizer() method, while xi and yi come as parts of the state of the optimizer itself, i.e. from internals of the optimizer. So both values can be usefull for debugging (these are DEBUG messages, notice it) -- if they differ that means that something goes wrong with passing values to the optimizer...
Another possibility is to print params (with names of the space dimensions) and print yi from the internals of the optimizer.
Third possibility (I like it after thinking on it): print 'params' from generate_optimizer() AND (additionally to the current implementation) print there the 'loss' also from generate_optimizer() -- that will be one message. Plus, print xi and yi from internal state of the optimizer. So that one could compare them if he need it for some reasons.

hroff-1902:
> > if they differ that means that something goes wrong with passing values to the optimizer...
> 
> You had me on this... 👍
> As mentioned - unless i missed something - and i did.
> 
> The order is not important - that's good to go 👍

I'll add the 'loss' values from generate_optimizer() to first message later...

PR #1795, opened 2019-04-23
hyperopt: --random-state for optimizer to get reproducible results

hroff-1902:
A part of #1775.

This new parameter allows to pass a value for the random_state argument of the skopt optimizer to get reproducible optimization path.

Use it defining any positive integer value, for example:
```
python freqtrade hyperopt --random-state=33
```
Non-integer AND zero are invalid and rejected.

The check_int_positive() method was added to the Arguments class, which validates the parameter value at early stage. Can further be used in freqtrade/arguments.py for validation of other integer parameters where a positive value is expected.

### DIFF
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -311,7 +311,7 @@ class Arguments(object):
             help='Set random state to some positive integer for reproducible hyperopt results.',
             dest='hyperopt_random_state',
             default=None,
-            type=Arguments.check_positive,
+            type=Arguments.check_int_positive,
             metavar='INT',
         )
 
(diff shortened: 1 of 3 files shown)


freqtrade/arguments.py
```diff
@@ -306,6 +306,14 @@ def hyperopt_options(parser: argparse.ArgumentParser) -> None:
             dest='print_all',
             default=False
         )
+        parser.add_argument(
+            '--random-state',
+            help='Set random state to some positive integer for reproducible hyperopt results.',
+            dest='hyperopt_random_state',
+            default=None,
+            type=Arguments.check_int_positive,
```
xmatthias: THIS is a possibility? :open_mouth: great!

hroff-1902: https://stackoverflow.com/questions/14117415/in-python-using-argparse-allow-only-positive-integers

PR #1802, opened 2019-04-24
Refactor config

xmatthias:
## Summary
Simplify configuration class complexity by extracting repeated logic of "check if key is in namespace", "copy it to config", "print log" to a separate function.

## Quick changelog

- simplify configuration class
- combine all optimize logics (they are all called anyway, so it does not matter where it applies).
- removed duplicate entries (--timerange, --refresh_pairs, ...) since all where always called anyway.

### DIFF
--- a/freqtrade/configuration.py
+++ b/freqtrade/configuration.py
@@ -216,7 +216,7 @@ class Configuration(object):
             logger.info(f'Created data directory: {datadir}')
         return datadir
 
-    def _args_to_config(self, config, argname, configname, logstring, logfun=None) -> None:
+    def _args_to_config(self, config, argname, logstring, logfun=None) -> None:
         """
         logfun is applied to the configuration entry before passing
         that entry to the log string using .format().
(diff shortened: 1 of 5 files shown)


hroff-1902:
LGTM,
let this be in this manner.

I'll probably refactor this a bit futher, some time later... Let this be as it is for now... (I mean the _args_to_config() function. It can later be simplified and improved, I guess...)

Also, you moved processing of all optimization parameters to the single function. I also wanted to move *common* optimization params to _load_optimize_config(), *but* I supposed to leave processing of the edge, backtesting and hyperopt-specific params in the corresponding functions... Can be done later, let's merge it to have the new fixed starting point for further improvements...

xmatthias:
Honestly, everything may be further refactored ... code of a living project is rarely at a standstill

one idea i had was to immediately covert args to a dict on initialization of the config class, which will allow to use `args.get('param', default)` ... but one step at a time.

the dubious concerns me however - you think there is a bug or problem in the code?

hroff-1902:
> the dubious concerns me however - you think there is a bug or problem in the code?

Nope, I guess my doubts are, 1st, related to the names of the command line parameters in the logstrings (like 'Parameter --enable-position-stacking detected ...'). All them should ideally be located in the arguments.py, that's why they create doubts. But current implementaion also contain them in configuration.py... Second -- a funcky (looking a bit strange, but I understand why you have added it) logfun argument,,, Third -- I personally liked two diff log messages for 'Detected...' and 'Using...' Fourth: absense of separate functions for processing edge, backtesting and hyperopt params -- they can easily be added.

So let's merge it and I'll think later on how to move those names of the command line params to arguments.py. Maybe other stuff get simplier then...

xmatthias:
> absense of separate functions for processing edge, backtesting and hyperopt params

most of them are combined - and parsing / setting an option that is not used should not cause trouble (i would hope)...

Issue #1803, opened 2019-04-24
load pickle EOFError raised


gitgitgat:
What could be causing the following error in hyperopts, new dev version:
 

> File "/anaconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
>     "__main__", mod_spec)
>   File "/anaconda3/lib/python3.7/runpy.py", line 85, in _run_code
>     exec(code, run_globals)
>   File "freqtrade/__main__.py", line 15, in <module>
>     main.main(sys.argv[1:])
>   File "/Users/user/Downloads/GITHUB/freqtrade/freqtrade/main.py", line 34, in main
>     args.func(args)
>   File "/Users/user/Downloads/GITHUB/freqtrade/freqtrade/optimize/hyperopt.py", line 350, in start
>     hyperopt.start()
>   File "/Users/user/Downloads/GITHUB/freqtrade/freqtrade/optimize/hyperopt.py", line 274, in start
>     self.load_previous_results()
>   File "/Users/user/Downloads/GITHUB/freqtrade/freqtrade/optimize/hyperopt.py", line 248, in load_previous_results
>     self.trials = self.read_trials()
>   File "/Users/user/Downloads/GITHUB/freqtrade/freqtrade/optimize/hyperopt.py", line 94, in read_trials
>     trials = load(self.trials_file)
>   File "/Users/user/.env/lib/python3.7/site-packages/joblib/numpy_pickle.py", line 598, in load
>     obj = _unpickle(fobj, filename, mmap_mode)
>   File "/Users/user/.env/lib/python3.7/site-packages/joblib/numpy_pickle.py", line 526, in _unpickle
>     obj = unpickler.load()
>   File "/anaconda3/lib/python3.7/pickle.py", line 1083, in load
>     raise EOFError
> EOFError
>

gitgitgat:
also in master version

xmatthias:
it sounds like a problem with your pickle file (it's corrupted) - probably caused by running multiple instances in parallel ...

try to deltee these files:

```
rm user_data/hyperopt_results.pickle
rm user_data/hyperopt_tickerdata.pkl
```

(or one of the 2 - or move them away one by one)

hroff-1902:
exception is in self.read_trials() -- so delete the user_data/hyperopt_results.pickle file

need to think on how to avoid this and allow multiple heperopt running simultaneously (if someone wants it for some reasons)...

xmatthias:
[...]

PR #1804, opened 2019-04-25
prevent hyperopt from running multiple instances simultaneously

hroff-1902:
Set and check the lock.

Inspired by #1803.

A universal approach (py-filelock) was choosen because a custom one with fcntl/flock is 1) platform specific, 2) a lock set on the results pickle file can persist in case the app was killed in the middle. Can be problematic for an inexperienced user to remove the lock.

* I tested it with graceful hyperopt exit, `Ctrl+C`, `kill` and `kill -9` (Ubuntu).
* @xmatthias, you can adjust the log messages if you wish.
* Please test it on Windows.
* A test for it should probably be added. Too complicated for me, I'm still not quite familiar with all that mocking stuff...
* Docs should mention it somehow if recently added sentence about multiple Heperopts does not suffice.
* TODO (sep. PR, future): start() for Backtesting, Edge and Hyperopt should return boolean in order to return non-zero exit code in some cases. (or simply raise an exception to freqtrade here if the lock is detected?)

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from pprint import pprint
 from typing import Any, Dict, List
 
+from filelock import Timeout, FileLock
 from joblib import Parallel, delayed, dump, load, wrap_non_picklable_objects, cpu_count
 from pandas import DataFrame
 from skopt import Optimizer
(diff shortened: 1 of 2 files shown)


xmatthias:
> A test for it should probably be added. Too complicated for me, I'm still not quite familiar with all that mocking stuff...

for me too ... 
the only idea to test this even remotely is to use multiprocessing for the tests - have one thread wait a bit and see if the other thread still holds the lock.

I am however not sure if the effort justifies the outcome - hyperopt is not well covered by tests anyway ... 


One thing that surprised me though is that the lockfile remains there even after the execution.
It's "somehow" working to detect that it's not locked anymore (i assume that's a feature of that library?) ... but it's still surprising (and kills my only other idea for a test).

hroff-1902:
> One thing that surprised me though is that the lockfile remains there even after the execution.
> It's "somehow" working to detect that it's not locked anymore (i assume that's a feature of that library?) ... but it's still surprising (and kills my only other idea for a test).

Me too (surprised, I also noticed it). Well, I specially tested it with kill -9 and other approaches to terminate the process (as mentioned above) -- I did not found conditions where a hanging lock is left when the locking process is not alive... Seem's to be okay, a solid working solution, as to me...

A bit lazy in the moment to stady the sources of lockfile, may be later...

xmatthias:
Nah that's fine, no reason to do that - just surprised, is all.
I'll have a look at this monday (won't be at my windows test-pc before then) - if it works there too it's good to me :+1:

hroff-1902:
Included into #1775 as a part of enhancements to hyperopt

PR #1814, opened 2019-04-30
immediately confirm forcesell

xmatthias:
## Summary
This will return the status of the rpc forcesell call
Currently, when you forcesell, the only feedback you get is once the order is filled.
This may be immediately, but it could also be delayed if the limit-sell order is not filled.

This is dangerous as i tend to make typos on the phone, and having immediate feedback if it worked should improve usability.

## What's new?
2019-04-30-062933_579x61_scrot

2019-04-30-062849_486x57_scrot

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -346,7 +346,7 @@ class RPC(object):
 
         return {'status': 'No more buy will occur from now. Run /reload_conf to reset.'}
 
-    def _rpc_forcesell(self, trade_id) -> None:
+    def _rpc_forcesell(self, trade_id) -> Dict[str, str]:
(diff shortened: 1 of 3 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -401,6 +401,7 @@ def _exec_forcesell(trade: Trade) -> None:
 
         _exec_forcesell(trade)
         Trade.session.flush()
+        return {'result': f'Created sell orders for trade {trade_id}.'}
```
mishaker: sell `order` not `orders` ?

xmatthias: great catch :+1:

PR #1823, opened 2019-05-03
Update qtpylib

### LINKED ISSUE #1818: Update parts of qtpylib included
The indicators borrowed from the qtpylib (freqtrade/vendor/qtpylib/indicators.py) seems to be now pretty old and contain outdated codebase.

That project has recently implemented some improvements and serious bugfixes in the last versions. For example, the authors fixed errorneous implementation of the stochastic indicator. See history of changes in https://github.com/ranaroussi/qtpylib/commits/master/qtpylib/indicators.py

The authors of qtpylib, however, have recently moved to the usage of Apache License, while freqtrade is licensed with GPL v3.

It should not be a problem though, Apache v2 software can be included into projects licensed with gpl v3, they are compatible this way. See description from Apache Software Foundation on compatibility with GPL for more info: http://www.apache.org/licenses/GPL-compatibility.html

hroff-1902:
Resolves #1818.

Current qtpylib/indicators.py fetched from its master; adjusted to make flake happy.

Need help: 
* tests failed with new version of indicators.

### DIFF
--- a/freqtrade/vendor/qtpylib/indicators.py
+++ b/freqtrade/vendor/qtpylib/indicators.py
@@ -113,11 +113,15 @@ def heikinashi(bars):
                         bars['low'] + bars['close']) / 4
 
     # ha open
+    idx = bars.index.name
+    bars.reset_index(inplace=True)
     for i in range(0, len(bars)):
         bars.at[i, 'ha_open'] = (
                 (bars.at[0, 'open'] if i == 0 else bars.at[i - 1, 'ha_open']) +
                 (bars.at[0, 'close'] if i == 0 else bars.at[i - 1, 'ha_close'])
             ) / 2
+    if idx:
+        bars.set_index(idx, inplace=True)
 
     bars['ha_high'] = bars.loc[:, ['high', 'ha_open', 'ha_close']].max(axis=1)
     bars['ha_low'] = bars.loc[:, ['low', 'ha_open', 'ha_close']].min(axis=1)
--- a/freqtrade/vendor/qtpylib/indicators.py
+++ b/freqtrade/vendor/qtpylib/indicators.py
@@ -113,15 +113,9 @@ def heikinashi(bars):
                         bars['low'] + bars['close']) / 4
 
     # ha open
-    idx = bars.index.name
-    bars.reset_index(inplace=True)
-    for i in range(0, len(bars)):
-        bars.at[i, 'ha_open'] = (
-                (bars.at[0, 'open'] if i == 0 else bars.at[i - 1, 'ha_open']) +
-                (bars.at[0, 'close'] if i == 0 else bars.at[i - 1, 'ha_close'])
-            ) / 2
-    if idx:
-        bars.set_index(idx, inplace=True)
+    bars.at[0, 'ha_open'] = (bars.at[0, 'open'] + bars.at[0, 'close']) / 2
+    for i in range(1, len(bars)):
+        bars.at[i, 'ha_open'] = (bars.at[i - 1, 'ha_open'] + bars.at[i - 1, 'ha_close']) / 2
 
     bars['ha_high'] = bars.loc[:, ['high', 'ha_open', 'ha_close']].max(axis=1)
     bars['ha_low'] = bars.loc[:, ['low', 'ha_open', 'ha_close']].min(axis=1)
--- a/freqtrade/vendor/qtpylib/indicators.py
+++ b/freqtrade/vendor/qtpylib/indicators.py
@@ -260,7 +260,7 @@ def rolling_std(series, window=200, min_periods=None):
     else:
         try:
             return series.rolling(window=window, min_periods=min_periods).std()
-        except Exception as e:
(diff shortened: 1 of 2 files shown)


freqtrade/vendor/qtpylib/indicators.py
```diff
@@ -103,47 +104,50 @@ def session(df, start='17:00', end='16:00'):
...
     bars['ha_close'] = (bars['open'] + bars['high'] +
                         bars['low'] + bars['close']) / 4
 
-    bars['ha_open'] = (bars['open'].shift(1) + bars['close'].shift(1)) / 2
-    bars.loc[:1, 'ha_open'] = bars['open'].values[0]
-    for x in range(2):
-        bars.loc[1:, 'ha_open'] = (
-            (bars['ha_open'].shift(1) + bars['ha_close'].shift(1)) / 2)[1:]
+    # ha open
+    bars.loc[:1, 'ha_open'] = (bars['open'] + bars['close']) / 2
```
xmatthias: The problem seems to be the :1 (if i remove it, tests pass).

we'll  need to dig into it as i don't think that's correct (i will try to talk to a pandas expert about this ...)

xmatthias: I'm not sure about the correct formulae for heikinashi bars - but it did change here (result for open is different between old and new implementation). 

Also - the new implementation is a LOT slower ...

hroff-1902: How have you tested performance (mentioned above)?

Just issuing the backtesting command with def. strategy, but with a longer timerange?

hroff-1902:
@xmatthias Please compare performance of heikinashi with last commit in this branch in your conditions you mentioned.

freqtrade/vendor/qtpylib/indicators.py
```diff
@@ -103,47 +104,50 @@ def session(df, start='17:00', end='16:00'):
...
 
-    bars['ha_open'] = (bars['open'].shift(1) + bars['close'].shift(1)) / 2
-    bars.loc[:1, 'ha_open'] = bars['open'].values[0]
-    for x in range(2):
-        bars.loc[1:, 'ha_open'] = (
-            (bars['ha_open'].shift(1) + bars['ha_close'].shift(1)) / 2)[1:]
+    # ha open
+    bars.loc[0:1, 'ha_open'] = (bars['open'].values[0] + bars['close'].values[0]) / 2
+    prev_open = bars[:1]['ha_open'].values[0]
+    for idx, _ in bars[1:][['ha_open', 'ha_close']].iterrows():
```
xmatthias: the .loc line (that's vectorized), so the last commit does not improve performance .

The performance problem lies in iterating through all rows instead of using a vectorized operation, which was used before.

I'm not an expert in heikinashi - but I'm pretty certain it can be done vectorized too. (looks like a simple `.lag() / .shift()` operation at a first look).

hroff-1902:
* Previous (now in master and develop) implementation of heikin ashi was really broken. Newly fetched from qtpy is correct.
* Problem with performance resolved. I used .at pandas method. If you wish, rewrite it with .loc etc.
* I found that my implementation worked fine in 'live' freqtrade but failed in tests (fixed in the last commit). The reason was in in indexing of the dataframes, absence of index in case of tests, as far as I understand. This should probably be adjusted later so that the tests execute indicators in the same conditions.

freqtrade/vendor/qtpylib/indicators.py
```diff
@@ -103,47 +104,53 @@ def session(df, start='17:00', end='16:00'):
...
                         bars['low'] + bars['close']) / 4
 
-    bars['ha_open'] = (bars['open'].shift(1) + bars['close'].shift(1)) / 2
-    bars.loc[:1, 'ha_open'] = bars['open'].values[0]
-    for x in range(2):
-        bars.loc[1:, 'ha_open'] = (
-            (bars['ha_open'].shift(1) + bars['ha_close'].shift(1)) / 2)[1:]
+    # ha open
+    idx = bars.index.name
+    bars.reset_index(inplace=True)
```
xmatthias: I just pushed a fix for the tests - we simply need to reset the index, so we can save playing with the index here.

In addition - i would start range at 1 - and move the calculation of bar0 to before the loop.
It'll save a comparison per row - it's only a few ys in my 1m row test-setup - but then i have plenty of ram and my pc was bored, so it might make a difference on a small machine.

``` python
    bars.at[0, 'ha_open'] = (bars.at[0, 'open'] + bars.at[0, 'close']) / 2
    for i in range(1, len(bars)):
        bars.at[i, 'ha_open'] = (bars.at[i - 1, 'ha_open'] + bars.at[i - 1, 'ha_close']) / 2
```

hroff-1902: Okay, agree, done. This may have sense...

hroff-1902:
> we probably should try and push our improvements to heikenashi upstream ...
> 
> @hroff-1902 what do you think?

I'll do.

PR #1830, opened 2019-05-06
check python version

hroff-1902:
...in main.py and raise SystemException if python <3.6 is used

Do not merge yet, TODO:
* allow to run the main.py with python 2 and python < 3.5 (typing!) in order to raise an exception "freqtrade requires 3.6"


freqtrade/main.py
```diff
@@ -47,8 +54,8 @@ def main(sysargv: List[str]) -> None:
     except OperationalException as e:
         logger.error(str(e))
         return_code = 2
-    except BaseException:
-        logger.exception('Fatal exception!')
+    except BaseException as e:
+        logger.exception('Fatal exception! ' + str(e))
```
xmatthias: Note that this is not necessary. 
[from the logging logs](https://docs.python.org/3/library/logging.html#logging.Logger.exception)

> Logs a message with level ERROR on this logger. The arguments are interpreted as for debug(). Exception info is added to the logging message. This method should only be called from an exception handler.

hroff-1902: without this:
```
Fatal exception!
  File "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/freqtrade/main.py", line 31, in main
    raise SystemError("Freqtrade requires Python version >= 3.6")
SystemError: Freqtrade requires Python version >= 3.6
```
with this:
```
Fatal exception! Freqtrade requires Python version >= 3.6
  File "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/freqtrade/main.py", line 31, in main
    raise SystemError("Freqtrade requires Python version >= 3.6")
SystemError: Freqtrade requires Python version >= 3.6
```

Hmm, in some of my early tests of this functionality I have not fount the 'Freqtrade requires Python version >= 3.6' string in the printed test, so I added this. Maybe it's redundant, right...

xmatthias:
> python 2

python2 will reach eol in 7 months (and [ticking](https://pythonclock.org/) ) ...

hroff-1902:
> > python 2
> 
> python2 will reach eol in 7 months (and [ticking](https://pythonclock.org/) ) ...so i'm not certain it's worth any effort (especially since freqtrade never supported python2 )...

Well, yes, but the idea was to print a message to the user if she runs the freqtrade with python (v2) instead of python3...

Actually, I do not see how to achieve it without introducing an additional small jacket "main" module, or something like that (due to python3 typing hints used everywhere in the project). AND in the same time I do not want to make such serious modifications to the codebase since, right, python2 will not be supported pretty soon...

So, I'm going to give up of this functionality (for python2). A bit later when the time allows, I'll probably cleanup slightly the current solution and remove the 'don't merge' flag when it'll be ready for review and merge...

xmatthias:
you should be able to accomplish this by "failing" before any imports (probably in bin/freqtrade ... however i personally would not do that, as that file is on my "to remove"-list (it's not really needed - and the list is only in my head ...)

hroff-1902:
> probably in bin/freqtrade 
...And in main.py... and in the scripts... too many entry points...

Let's just leave this idea of warning for python2 as unrealized...

hroff-1902:
@xmatthias pls review latest changes -- I moved the python version check to the top of main.py.
This seems to be more robust and elegant. Can also be implemented in scripts in the same way...

hroff-1902:
I don't know why additional settings are in this PR -- maybe when I inserted the python version check into main() it required some additional changes around and when I moved the check at the top of the file they just left... o_0

@xmatthias if you wish I could create a new PR with only the python version check added at the top of the main.py...

PR #1835, opened 2019-05-09
Add printing dataframe to documentation

### LINKED ISSUE #1833: Ability to see candles update in terminal
It would be nice to be able to see candles as they update into dataframe. This would help greatly with creating a custom indicator and ensure its working properly. This feature could be turned on/off with "-v", just an idea. Thanks for the time.

xmatthias:
## Summary
Document how to inspect a generated dataframe using Print statements

Resolves #1833

### DIFF
--- a/docs/bot-optimization.md
+++ b/docs/bot-optimization.md
@@ -367,7 +367,7 @@ def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
     return dataframe
 ```
 
-Printing more than a few rows is possible (don't use `.tail()`), however not recommended, as that will be very verbose (~500 lines per pair every 5 seconds).
+Printing more than a few rows is also possible (simply use  `print(dataframe)` instead of `print(dataframe.tail())`), however not recommended, as that will be very verbose (~500 lines per pair every 5 seconds).
 
 ### Where is the default strategy?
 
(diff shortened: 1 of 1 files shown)


docs/bot-optimization.md
```diff
@@ -345,6 +345,30 @@ if self.wallets:
...
+    # Print the Analyzed pair
+    print(f"result for {metadata['pair']}")
+
+    # Inspect the last 5 rows
+    print(dataframe.tail())
+
+    return dataframe
+```
+
+Printing more than a few rows is possible (don't use `.tail()`), however not recommended, as that will be very verbose (~500 lines per pair every 5 seconds).
```
hroff-1902: Printing more than a few rows is **also** possible (**simply use** `print(dataframe)` instead of `print(dataframe.tail())`), however ...

hroff-1902:
> I am really not sure about this. We cannot include all "helpful" scripts here. maybe in FAQ or some kind of gists section but if we start to put all requested scripts here, it will be such a mess. I am not against it though if others find it necessary.

As for me, *this particular* "helpful" snippet fits perfectly into optimization.md (I would like to change the name of this doc file since it describe strategies rather than backtesting or hyperopt, but this is another point). It also would perfectly fit into the faq.md as well, as I see. So both locations are okay for me.

xmatthias:
I would leave it here for now.
the documentation is constantly improving (hopefully).
And yes, this should be renamed, i always struggle to find this document since the documentation lists it as "custom strategy".

PR #1837, opened 2019-05-10
minor: hyperopt output improvements

hroff-1902:
A set of small improvements to hyperopt:

* Calculation of Avg Profit corrected: it no longer is multiplied by 100 so that
SumProfit% = N_trades * Avg Profit%
* The precision and the length of this printed field adjusted correspondingly.
* Number of initial random points (30) made a named constant. Later (diff. PR) should be adjustable (bigger for long optimization runs)
* A special marker sign ('*') is printed at the begining of the line with results for first N initial points. In order to distinguish random points of the rest in the optimization path. Maybe this should be printed only in the print-all mode, now is printed when no --print-all option was specified as well...
* A redundant dot in the output (noisy double dot after 'mins') removed
* A redundant empty line in the output removed when printing in the print-all mode.
* The 'Loss' term was changed to 'Objective' in the output string for clarity. The Loss term is used in the hyperopt package which is no longer used by the project. Skopt uses 'Objective function' for this.
* Numbering of evals is now human-friendly. Starts from 1 rather than 0 and ends at 1000/1000 rather than 999/1000 in case of 1000 evals.

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -33,6 +33,7 @@ from freqtrade.resolvers import HyperOptResolver
 logger = logging.getLogger(__name__)
 
 
+INITIAL_POINTS = 30
 MAX_LOSS = 100000  # just a big enough number to be bad result in loss optimization
 TICKERDATA_PICKLE = os.path.join('user_data', 'hyperopt_tickerdata.pkl')
 TRIALSDATA_PICKLE = os.path.join('user_data', 'hyperopt_results.pickle')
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -228,13 +235,13 @@ def format_results(self, results: DataFrame) -> str:
         Return the format result in a string
         """
         trades = len(results.index)
-        avg_profit = results.profit_percent.mean() * 100.0
+        avg_profit = results.profit_percent.mean()
```
xmatthias: Are you sure that this is correct?

`profit_percent` is calculated using `calc_profit_percent()` in `persistence.py` - which does not seem to calculate, but a factor.

Rest looks good :+1:

hroff-1902:
You are right! Since it's percentage, it SHOULD be multiplied by 100 for output.
BUT, the total profit percentage (Sum%) SHOULD also be multiplied by 100!

Both corrected, length of the fields adjusted.

hroff-1902:
@xmatthias I found it useful to set stake_amount = 1.0 in the config for internal testing in order to imagine better all those ratios and percents...

hroff-1902:
Refer to #1775 here, just to have this PE referenced there...

PR #1852, opened 2019-05-14
Minor: hyperopt verify ticker data

hroff-1902:
* Verification of the loaded ticker data. Taken and slightly adapted from backtesting part.
* Log time interval (min_date, max_date, duration) used for backtesting.
* New test for no data case added.

Part of #1775.

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -24,7 +24,8 @@ from freqtrade import DependencyException
 from freqtrade.arguments import Arguments
 from freqtrade.configuration import Configuration
 from freqtrade.data.history import load_data
-from freqtrade.optimize import get_timeframe
+from freqtrade.exchange import timeframe_to_minutes
+from freqtrade.optimize import get_timeframe, validate_backtest_data
 from freqtrade.optimize.backtesting import Backtesting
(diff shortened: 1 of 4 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -282,9 +283,25 @@ def start(self) -> None:
...
+        if not data:
+            logger.critical("No data found. Terminating.")
+            return
+
+        min_date, max_date = get_timeframe(data)
+        # Validate dataframe for missing values (mainly at start and end, as fillup is called)
+        validate_backtest_data(data, min_date, max_date,
+                               timeframe_to_minutes(self.ticker_interval))
+        logger.info(
+            'Backtesting data from %s up to %s (%s days)..',
```
xmatthias: "Hyperopting data ..." ? 

i was ... confused by the output while running this

hroff-1902: Backtest.backtest() is called with these data (by hyperopt, yes), so "Backtesting data"...
I wanted to change it in the same way in backtestesting.py, where the code was taken (now it uses "Measuring data..." there).

Changed it in both hyperopt.py and backtesting.py

Issue #1853, opened 2019-05-14
Hyperopts


gitgitgat:
An error warning comes up when trying 

python3 freqtrade -c config.json hyperopt --customhyperopt HyperOpts  -e 5000 --print-all

```
/Users/nsm/.env/lib/python3.7/site-packages/sklearn/externals/joblib/__init__.py:15: DeprecationWarning: sklearn.externals.joblib is deprecated in 0.21 and will be removed in 0.23. Please import this functionality directly from joblib, which can be installed with: pip install joblib. If this warning is raised when loading pickled models, you may need to re-serialize those models with scikit-learn 0.21+.
  warnings.warn(msg, category=DeprecationWarning)
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "freqtrade/__main__.py", line 15, in <module>
    main.main(sys.argv[1:])
  File "/Users/nsm/freqtrade/freqtrade/main.py", line 34, in main
    args.func(args)
  File "/Users/nsm/freqtrade/freqtrade/optimize/hyperopt.py", line 362, in start
    config = setup_configuration(args)
  File "/Users/nsm/freqtrade/freqtrade/optimize/hyperopt.py", line 338, in setup_configuration
    config = configuration.load_config()
  File "/Users/nsm/freqtrade/freqtrade/configuration.py", line 78, in load_config
    config = deep_merge_dicts(self._load_config_file(path), config)
  File "/Users/nsm/freqtrade/freqtrade/configuration.py", line 117, in _load_config_file
    conf = json.load(file)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 296, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
[...]

PR #1854, opened 2019-05-15
Suppress scikit-learn deprecation warning in hyperopt

hroff-1902:
That was not as easy... 😃 

scikit-learn specifically forces warnings to be displayed, which we don't like.
Plus, it is not used by freqtrade directly but via skopt...

See #1853 for examples of the warning.

See:
* https://github.com/scikit-learn/scikit-learn/issues/2531
* https://github.com/ims-tcl/DeRE/commit/cb865fedddf6977cb7536b8c4334dc2325d09e53
* https://stackoverflow.com/questions/32612180/eliminating-warnings-from-scikit-learn

for more details.

Use
```
from freqtrade.optimize.hyperopt_interface import Categorical, Dimension, Integer, Real
```
instead of
```
from skopt.space import Categorical, Dimension, Integer, Real
```
in the custom HyperOpts to get rid of this deprecation warning
(refer to default_hyperopt.py for example) without implementing this ugly
workaround in every custom HyperOpts.

Let's wait till scikit-learn v0.23 released and skopt using it
where it will probably be eliminated...

### DIFF
--- a/user_data/hyperopts/sample_hyperopt.py
+++ b/user_data/hyperopts/sample_hyperopt.py
@@ -6,10 +6,11 @@ from typing import Dict, Any, Callable, List
 from functools import reduce
 
 import numpy
-from skopt.space import Categorical, Dimension, Integer, Real
 
 import freqtrade.vendor.qtpylib.indicators as qtpylib
 from freqtrade.optimize.hyperopt_interface import IHyperOpt
+from freqtrade.optimize.hyperopt_interface import Categorical, Dimension, Integer, Real
+
 
 class_name = 'SampleHyperOpts'
 
--- a/freqtrade/optimize/default_hyperopt.py
+++ b/freqtrade/optimize/default_hyperopt.py
@@ -5,10 +5,10 @@ from pandas import DataFrame
 from typing import Dict, Any, Callable, List
 from functools import reduce
 
-from skopt.space import Categorical, Dimension, Integer, Real
-
 import freqtrade.vendor.qtpylib.indicators as qtpylib
 from freqtrade.optimize.hyperopt_interface import IHyperOpt
+from freqtrade.optimize.hyperopt_interface import Categorical, Dimension, Integer, Real
+
 
 class_name = 'DefaultHyperOpts'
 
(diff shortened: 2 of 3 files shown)


user_data/hyperopts/sample_hyperopt.py
```diff
@@ -6,10 +6,11 @@
 from functools import reduce
 
 import numpy
-from skopt.space import Categorical, Dimension, Integer, Real
 
 import freqtrade.vendor.qtpylib.indicators as qtpylib
 from freqtrade.optimize.hyperopt_interface import IHyperOpt
+from freqtrade.optimize.hyperopt_interface import Categorical, Dimension, Integer, Real
```
xmatthias: I don't think this is currently necessary (and it's problematic for another reason too):

a) it works without this import from hyperopt_interface - also witht he below command
b) since this is a file we expect users to copy / keep / modify, it's a user interface - so if we add this now - we are kind of constraint to keep these imports in the interface (or throw deprecation warnings ourselfs). This is not per se problematic - but odd if you need to investigate how these actually work, as it's not clear it's passed through from skopt - so users may google the wrong stuff.

```bash 
freqtrade --config config.json.example --datadir freqtrade/tests/testdata hyperopt -e 5 --customhyperopt SampleHyperOpts
```

hroff-1902:
Yes, it's ugly looking and fragile. Let's dismiss it and wait till scikit-learn v0.23.

hroff-1902:
Let's close it and forget?

mishaker:
@hroff-1902 if the problem will be solved in the next version then I vote for closing this one :)

hroff-1902:
> @hroff-1902 if the problem will be solved in the next version then I vote for closing this one as it is not a blocker :)

Maybe not next version, next but one (they promise get rid of own copy of joblib in v0.23, current is v0.21 as far as I recall without looking into reqs.txt.

Let's close this PR, right! I don't mind at all, it will be kept in the github storage of closed PRs... Not all attempts on this planet are successive and some solutions are fragile, ugly and look like nailing some tailings with rusty crooked nails... (This our local programmer's humor if you didn't know...)

hroff-1902:
Guys, let's close it and move to the github's archive of the closed PRs...

PR #1862, opened 2019-05-17
minor: data/history cleanup

hroff-1902:
* the check for empty exchange parameter moved from load_pair_history() to download_pair_history() -- it's more correct place for it since download_pair_history() is called separately in the freqtrade codebase.
* wording in a comment.
* wording in log message ("No data for pair:" --> "No history data for pair:").
* wording in log message ("Download the pair:" --> "Download history data for pair:").
* tests adjusted to reflect changes in the log messages.
* the datadir logging also moved to that log message i.o. being a separate log message in load_pair_history().
* `except BaseException` changed to `except Exception`. That's what I kept in mind stating in #1859 ("...has actually sence for freqtrade codebase.").

Currently, this function catches BaseException, including KeyboardInterrupt. As a result, if freqtrade hyperopt or freqtrade backtesting is called with the `-r` option, thus starts reloading history data (especially if it receives exchange ban and begins to repeat the load operations with ugly messages) and if the user then wants to terminate it pressing Ctrl+C -- loading of history for only one pair is terminated, the user needs to press Ctrl+C many times, which is sucks especially for long whitelists... This is fixed now.

* The log message "Failed to download history data for pair" in except changed from info to error since it indicates an error.

### DIFF
--- a/freqtrade/data/history.py
+++ b/freqtrade/data/history.py
@@ -90,13 +90,8 @@ def load_pair_history(pair: str,
     :return: DataFrame with ohlcv data
     """
 
-    # If the user force the refresh of pairs
+    # The user forced the refresh of pairs
     if refresh_pairs:
-        if not exchange:
-            raise OperationalException("Exchange needs to be initialized when "
-                                       "calling load_data with refresh_pairs=True")
-
-        logger.info('Download data for pair and store them in %s', datadir)
         download_pair_history(datadir=datadir,
                               exchange=exchange,
                               pair=pair,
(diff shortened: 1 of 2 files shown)


freqtrade/data/history.py
```diff
@@ -231,7 +235,9 @@ def download_pair_history(datadir: Optional[Path],
 
         misc.file_dump_json(filename, data)
         return True
-    except BaseException:
-        logger.info('Failed to download the pair: "%s", Interval: %s',
-                    pair, ticker_interval)
+
+    except Exception:
```
xmatthias: I Love the explanation on why you changed BaseException to Exception! Great find btw!

hroff-1902: > I Love the explanation on why you changed BaseException to Exception! Great find btw!

Yeah, that's also for future devs. Now (after just an year) it's quite difficult to understand some changes made in, for example, backtesting by devs that are not active since those changes are not documented enough...

PR #938, opened 2018-06-22
Implement HTTP/JSON rpc module

gcarq:
## Summary

Implements an HTTP rpc module which returns JSON responses (Still WIP).

Proposal for configuration:
```
    "api_server": {
        "listen_ip_address": "127.0.0.1",
        "listen_port": 8080
    }
```

PR #1863, opened 2019-05-18
Add REST API to control the bot

### LINKED ISSUE #847: Security vs Usability
For discussion. 
The Freqtrade project is opensource and has an active community and all code is peer reviewed. 
These are good and proper barriers to protect a users data. I have personally reviewed the code end to end and found no malicious content.

In common with any and all automated trading programs the user must give a level of trust to the program to not abuse their API keys, which must be protected with as many layers the user and project can make available. 

However unlikely it is possible for rogue code, core project accounts to become compromised/utilised to inject functionality to scrape users exchange api keys. 

In the event of API key compromise API accounts on exchanges may commonly be restricted to a subset of features so direct withdrawal of funds may be prevented. This would not prevent all capital being used to purchase/pump a low cap coin. To the user the impact may be just a devastating.

A good measure to prevent API key leaking to protect the project from itself is for the user to be able to tightly control egress data from the bot. Ideally this should make use of code and control from outside the project. The rational being it is exponentially difficult for a threat agent to compromise many projects in parallel. 

It is common in Freqtrade for deployment to be within a docker container. 

I would propose discovery work be undertaken to offer the user a tested and documented deployment model/environment that wrappers network control with software that is outside the projects control, which the project will non-functionally test/approve to provide a functioning install. 

A contender for a deployment blueprint to prevent egress of data to any target other than an exchange *may be:

1 - Freqtrade installed into named Freqtrade docker instance as normal 
2 - 2nd squidguard docker installed from widely deployed maintained repository along with dns cache
2a - simple squidguard configuration to only allow HTTPS and  Only to a whitelist of Exchanges 
2b - simple DNS configuration to only allow DNS resolution of Exchange whitelist 
3 - host OS firewall script to forward all data from Docker Freqtrade to Docker Squidguard. This is simple in Linux. 

2a and 2b must be and can be simple and open for the user to maintain themselves. 

Whilst this solution is off-the-shelf and a proven topology to prevent data leaking through DNS calls or directly sent another site not whitelisted there would be useability and functional compromise. 

Not least the compromise to usability includes telegram integration would cease. 
There is no simple means to allow telegram integration and also tightly control egress data flows. 

If the project would aim to support an increased secure deployment environment as an option to users another means to control the bot should be provided / made available.

xmatthias:
## Summary
Since telegram is not available for everyone (or maybe not everyone want's to use telegram), we need another way to control the bot once it's running.

We've had a try for this before (#938), which somehow lost traction. I tried to rebase / merge #938 with develop - however there were too many conflicting changes, so I reutilized the new files. Full credit shall go to everyone involved in #938.

Replaces: #938
Resolves #847 by providing a method to control the bot locally.

## Quick changelog

- [x] Expose rest api 
- [x] Document rest api
- [x] Test rest api 
- [x] Authentication (using basic auth).

## Details
Under the hood, flask webserver (werkzeug) is used. 
While it's not suited for production usage to serve webpages with concurrent visits, I think it's fine to use in this case, since there will probably not be more than 1 or 2 users using the API (and not in parallel, most likely).
The flask-modules will only be loaded if the API is enabled. 
The bot warns if the api is started on anything not being localhost, and that's also clearly documented in the corresponding documentation.

## future possibilities
Future work may add a web UI to control the bot (maybe/probably as a seperate project).


freqtrade/rpc/api_server.py
```diff
@@ -0,0 +1,339 @@
...
+            logger.warning("SECURITY WARNING - This is insecure please set to your loopback,"
+                           "e.g 127.0.0.1 in config.json")
+
+        # Run the Server
+        logger.info('Starting Local Rest Server')
+        try:
+            self.srv = make_server(rest_ip, rest_port, self.app)
+            self.srv.serve_forever()
+        except Exception:
+            logger.exception("Api server failed to start, exception message is:")
```
mishaker: forgot the message?

freqtrade/rpc/api_server.py
```diff
@@ -0,0 +1,339 @@
...
+        logger.info('Starting Local Rest Server_end')
```
mishaker: you had this line above. you wanna say "Server started !" maybe?

freqtrade/rpc/api_server.py
```diff
@@ -0,0 +1,339 @@
...
+        pass
+
+    def rest_dump(self, return_value):
+        """ Helper function to jsonify object for a webserver """
+        return jsonify(return_value)
+
+    def rest_error(self, error_msg):
+        return jsonify({"error": error_msg}), 502
+
+    def register_rest_other(self):
```
mishaker: I find `register_rest_other` confusing. probably `register_rest_base` ? or something similar ?

freqtrade/rpc/api_server.py
```diff
@@ -0,0 +1,339 @@
...
+        # Combined actions and infos
+        self.app.add_url_rule(f'{BASE_URI}/blacklist', 'blacklist', view_func=self._blacklist,
+                              methods=['GET', 'POST'])
+        self.app.add_url_rule(f'{BASE_URI}/whitelist', 'whitelist', view_func=self._whitelist,
+                              methods=['GET'])
+        self.app.add_url_rule(f'{BASE_URI}/forcebuy', 'forcebuy',
+                              view_func=self._forcebuy, methods=['POST'])
+        self.app.add_url_rule(f'{BASE_URI}/forcesell', 'forcesell', view_func=self._forcesell,
+                              methods=['POST'])
+
```
mishaker: hmmm, if you had defined these endpoints in OpenAPI (swagger) and used `connexion` then you wouldn't need all these rules.

freqtrade/rpc/api_server.py
```diff
@@ -0,0 +1,339 @@
...
+
+class ApiServer(RPC):
+    """
+    This class runs api server and provides rpc.rpc functionality to it
+
+    This class starts a none blocking thread the api server runs within
+    """
+
+    def safe_rpc(func):
```
mishaker: the function is called `safe` which is expected to do some security check. but it is merely catching exceptions. probably another name?

config_full.json.example
```diff
@@ -109,6 +109,11 @@
         "token": "your_telegram_token",
         "chat_id": "your_telegram_chat_id"
     },
+    "api_server": {
+        "enabled": false,
+        "listen_ip_address": "127.0.0.1",
+        "listen_port": 8080
+    },
```
mishaker: no security token? a JWT token would be great here if IP is not local.

docs/rest-api.md
```diff
@@ -0,0 +1,181 @@
...
+    "api_server": {
+        "enabled": true,
+        "listen_ip_address": "127.0.0.1",
+        "listen_port": 8080
+    },
+```
+
+!!! Danger: Security warning
+  By default, the configuration listens on localhost only (so it's not reachable from other systems). We strongly recommend to not expose this API to the internet, since others will be able to control your bot.
+
```
mishaker: "since others will be able to control your bot." => "since others will potentially be able to control your bot.

freqtrade/rpc/api_server.py
xmatthias: i know that would sound good - however i did try connexion and i could not get it to call methods within a class instance - so it didn't have access to the running bot instance.

freqtrade/rpc/api_server.py
xmatthias: it doesn't make sense - it used to have more endpoints ... but most were removed / replaced ... so i'll remove this method.

config_full.json.example
xmatthias: i was tempted to use basic auth (username, password) ... 

Adding tokens (JWT or others) means to either pre-insert a token here (which most will bother to change) - or expect unexperienced users to generate one ... - or generate one at startup and hope users see it before the page rolls over to copy/paste it to the rest-client.

freqtrade/rpc/api_server.py
xmatthias: no(ish) ... by using exception the stack-trace is added. The text is missleading though.

hroff-1902:
Wow! 😄

config_full.json.example
mishaker: an unexperienced user who wants to control his bot by REST API by doesn't know what is security token ? :D

xmatthias: JWT will require 2 calls - assuming you use the rest_client script, thats 2 calls just to get `/version` ...

mishaker: just an idea: wouldn't it better to generate a password by `secret` ? 
``` 
import secrets
secrets.token_hex() #'1476f4cfa96f20af2ca8cfdf9c5920f54d78f1b835318d723ceec2a72403cc29'
```
So leave the password empty by default and raise an error if it is empty and document how user can generate a password. basically what flask or rails does for secret session. 
(I am not insisting though but the approach seems a bit more pertinent)

xmatthias: and how do you show that password to the user? 

right at startup? using docker - the password will be gone soonish (at least my configurations rotate logs after about a Meg or so)... so i'd need to restart the bot to see the password (which will probably be different each time ...) ?

mishaker: no, when you configure the bot with rest api you have to put a password yourself. that `secrets` part is just for documentation. we don't do anything but if there is no password then we raise an error.

mishaker: I am not sure which approach is better though, you can resolve this conv if you feel unnecessary.

xmatthias: no it makes sense (I just understood it wrongly) :+1:

PR #1864, opened 2019-05-19
Improve documentation to point out usage of future data during backtesting

### LINKED ISSUE #1839: Backtesting does not guard against use of future data
* Python Version: 3.6.7
  * CCXT version: 1.18.507
  * Branch: Develop
  * Last Commit ID: 4ae743ecb6e461e8f4febee715f204a7a484eed3
  
The back testing engine allows for shifts to future data and does nothing to prevent a user from utilizing data from the future for current signals.

### Steps to reproduce:

  1. Utilize the chikou span in a signal.
  2. Save the DF as a CSV.
  3. Investigate results. Chikou value always reflects the price + 30 rows.

xmatthias:
## Summary
We should be clear that we don't guard backtesting against usage of future data.
Also renamed the file - since i think strategy-customization is better suited for the content of this file.

Resolves #1839


docs/strategy-customization.md
```diff
@@ -53,6 +53,12 @@ file as reference.**
     It is therefore best to use vectorized operations (across the whole dataframe, not loops) and
     avoid index referencing (`df.iloc[-1]`), but instead use `df.shift()` to get to the previous candle.
 
+!!! Warning Using future data
+    Since backtesting passes the full time interval to the `populate_*()` methods, the strategy author
+    needs to take care to avoid having the strategy utilize data from the future.
+    Samples for usage of future data are `dataframe.shift(-1)`, `dataframe.resample("1h")` (this uses the left border of the interval, so moves data from an hour to the start of the hour).
+    They all use data which is not available during regular operations, so these strategies will perform well during backtesting, but will fail / perform badly dry-run tests.
+
```
mishaker: "but will fail / perform badly dry-run tests." ?

PR #1870, opened 2019-05-21
minor: data/history slight cleanup/imrovement

hroff-1902:
* Error description added into log message in except.
* <s>Renamed make_testdata_path() --> make_datadir_path(). Because it's not testdata dir.</s> 
* Common code transferred to new pair_data_filename() function to avoid code duplication.
* Tests adjusted.

### DIFF
--- a/freqtrade/data/history.py
+++ b/freqtrade/data/history.py
@@ -138,13 +138,13 @@ def load_data(datadir: Optional[Path],
     return result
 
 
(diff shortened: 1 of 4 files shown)


hroff-1902:
Renaming make_testdata_path() --> make_datadir_path() reverted, was wrong. This function returns testdata dir if datadir is None. Used only in tests. Maybe this should be removed completely from data/history.py and/or moved to corresponding tests (test_history and test_btanalisys).

PR #1873, opened 2019-05-23
Add some tests

xmatthias:
## Summary
Adding some missing tests for edge (and arguments.py) i noticed while reviewing #1871 (not related to that really ... but i saw them then).

### DIFF
--- a/freqtrade/tests/test_arguments.py
+++ b/freqtrade/tests/test_arguments.py
@@ -185,3 +185,12 @@ def test_testdata_dl_options() -> None:
     assert args.export == 'export/folder'
     assert args.days == 30
     assert args.exchange == 'binance'
+
+
+def test_check_int_positive() -> None:
+    assert Arguments.check_int_positive(2) == 2
+    assert Arguments.check_int_positive(6) == 6
+    with pytest.raises(argparse.ArgumentTypeError):
+        Arguments.check_int_positive(-6)
+
+    assert Arguments.check_int_positive(2.5) == 2
--- a/freqtrade/tests/test_arguments.py
+++ b/freqtrade/tests/test_arguments.py
@@ -187,13 +187,17 @@ def test_testdata_dl_options() -> None:
 
 
 def test_check_int_positive() -> None:
-    assert Arguments.check_int_positive(2) == 2
-    assert Arguments.check_int_positive(6) == 6
-    with pytest.raises(argparse.ArgumentTypeError):
-        Arguments.check_int_positive(-6)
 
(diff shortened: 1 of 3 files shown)


hroff-1902:
Maybe you also move here the make_testdata_path() function out of data/history.py to the tests?

freqtrade/tests/test_arguments.py
```diff
@@ -185,3 +185,12 @@ def test_testdata_dl_options() -> None:
...
     assert args.exchange == 'binance'
+
+
+def test_check_int_positive() -> None:
+    assert Arguments.check_int_positive(2) == 2
+    assert Arguments.check_int_positive(6) == 6
+    with pytest.raises(argparse.ArgumentTypeError):
+        Arguments.check_int_positive(-6)
+
+    assert Arguments.check_int_positive(2.5) == 2
```
hroff-1902: I don't remember now, should not it fail here? (2.5 is not an integer)
I think strings are passed to this function from the command line instead of ints or floats...
(Do not remember exactly now, need to sleep...)

xmatthias: well noish... - it uses `int(input)` and checks if it's positive.

but yes, it should also work with strings...

xmatthias:
`make_testdata_path()` - i will prefer to do that in a seperate PR, since that will touch "quite a few" more files...

freqtrade/tests/test_arguments.py
hroff-1902: In the app it's used with a string. And it's intended to be used with the strings. So the passing it a float and obtaining an int instead of an error (as in case of string with float) is not expected behaviour.

xmatthias: err no?

https://github.com/freqtrade/freqtrade/blob/1a5dbd29e032d30016ed15ed2aa76e4ba3323ee9/freqtrade/arguments.py#L320-L336

the 2nd argument (--min-trades) defaults to an integer - so i assume it'll pass an int to it (?) (i'll need to debug that when i'm on my real pc ... )

hroff-1902: I think in this case the check_int_positive() is simply not called and the default value assigned to dest as it is...

hroff-1902: you could add some nefative values as well for completeness... and zero...

xmatthias: that wasn't clear to me ... i added a type-hint now too to be clear that it's passing a string.

PR #1877, opened 2019-05-25
[proposal] Eliminate bin/freqtrade

xmatthias:
## Summary
Remove the "binary" script we had and replace it with the "entrypoint" logic (which automatically generates an executable file in the installation folder).

## Quick changelog

- Remove bin/freqtrade helper file
- Rely on setuptools "entrypoint" feature
- All methods of execution i'm aware of still work `python -m freqtrade`, `freqtrade --help` `python freqtrade/main.py --help` ...

In the future, this method would allow us to specify more entry points (ft_download) for the download script, for example.

## caveats 
I'm putting this as proposal since it will break existing "linked" installations (installed with `pip install -e .`) with the following error:

```
  File ".../bin/freqtrade", line 6, in <module>
    with open(__file__) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/xmatt/development/python/freqtrade/bin/freqtrade
```

### DIFF
--- a/bin/freqtrade
+++ b/bin/freqtrade
@@ -1,0 +1,13 @@
+#!/usr/bin/env python3
+
+import sys
+import warnings
+
+from freqtrade.main import main, set_loggers
+
+set_loggers()
+
+warnings.warn(
+    "Deprecated - To continue to run the bot like this, please run `pip install -e .` again.",
+    DeprecationWarning)
+main(sys.argv[1:])
--- a/bin/freqtrade
+++ b/bin/freqtrade
@@ -1,7 +1,0 @@
-#!/usr/bin/env python3
-
-import sys
-
-from freqtrade.main import main, set_loggers
-set_loggers()
-main(sys.argv[1:])
--- a/freqtrade/__main__.py
+++ b/freqtrade/__main__.py
@@ -6,9 +6,7 @@ To launch Freqtrade as a module
 > python -m freqtrade (with Python >= 3.6)
 """
 
-import sys
-
 from freqtrade import main
 
 if __name__ == '__main__':
-    main.main(sys.argv[1:])
+    main.main()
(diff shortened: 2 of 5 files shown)


hroff-1902:
> it will break existing "linked" installations (installed with `pip install -e .`) with the following error:
> 
> ```
>  Traceback (most recent call last):
>   File ".../bin/freqtrade", line 6, in <module>
>     with open(__file__) as f:
> FileNotFoundError: [Errno 2] No such file or directory: '/home/xmatt/development/python/freqtrade/bin/freqtrade
> ```

I cannot reproduce this, made `pip install -e .` and then `pip uninstall freqtrade` (after changing setup.py as in this PR) -- no error issued...

xmatthias:
you need to `pip install -e .` the develop (or master?) version, then change to this branch without issuing any "install" command (and run it with `freqtrade --help`).

If not installing again (which i tend to not do every time) the binary is not exchanged (which makes sense, if we think about it).
it's just a caveat i came across...

hroff-1902:
we could make this change in two steps, First, change the setup.py and leave bin/freqtrade for some time. Plus, add a message (noisy message, similar to sklearn deprecation) stating that users need to make `pip install -e .` again with current codebase. And only after some months delete bin/freqtrade then...

xmatthias:
that is a great idea! :+1:  Thanks, didn't think about that (my initial reaction was that the files will be conflicting, but that's not true).

hroff-1902:
Maybe the message should say in addition something like "The bin/freqtrade script is deprecated and will be removed in v.0.19" whenever that version is going to be released...
Rest seems to be ok for me 👍

hroff-1902:
^ Above (master branch) does not mean you should quickly move it into master. Just let this message be active for enough period of time in the master branch after it merged there...

PR #1879, opened 2019-05-25
Speed up startup time

xmatthias:
## Summary
Speeding up startup time. We don't need some dependencies (hyperopt) in certain situations. by avoiding loading unnecessary modules, both the memory profile and startup-time will improve.

## Quick changelog

- we don't need to load hyperopt (incl. skopt) dependency when running the bot in dry/live run
- we don't need backtest either in that case
- done the same for edge-cli for consistency
- Disables skopt `DeprecationWarning` unless hyperopt is run (since it's not loaded)
- Combine `setup_configuration()` method since it was almost identical.


hroff-1902:
Cool for the first look, will review it a bit later...

freqtrade/optimize/__init__.py
```diff
@@ -1,49 +1 @@
...
-import logging
-from datetime import datetime
-from typing import Dict, Tuple
-import operator
-
-import arrow
-from pandas import DataFrame
-
 from freqtrade.optimize.default_hyperopt import DefaultHyperOpts  # noqa: F401
-
```
hroff-1902: <s>If this file is really needed if it's almost empty?</s>
Aha, it's not empty in the next commit, forget it (reviewing here commit by commit, not the cumulative results...)

freqtrade/optimize/__init__.py
```diff
@@ -1 +1,99 @@
...
+
+    # Ensure we do not use Exchange credentials
+    config['exchange']['key'] = ''
+    config['exchange']['secret'] = ''
+
+    if method == RunMode.BACKTEST:
+        if config['stake_amount'] == constants.UNLIMITED_STAKE_AMOUNT:
+            raise DependencyException('stake amount could not be "%s" for backtesting' %
+                                      constants.UNLIMITED_STAKE_AMOUNT)
+
```
hroff-1902: Should not we change the stake_amount here to "unlimited" for backtesting (as exchange.key and exchange.secret are changed above to empty vals)? Editing the config every time for backtesting (and back for other activities) is now REALLY annoying... If so, log that the stake amount is changed for backtesting...

freqtrade/optimize/__init__.py
```diff
@@ -97,3 +97,19 @@ def start_hyperopt(args: Namespace) -> None:
         # TODO: return False here in order to help freqtrade to exit
         # with non-zero exit code...
         # Same in Edge and Backtesting start() functions.
+
+
+def start_edgecli(args: Namespace) -> None:
```
hroff-1902: why not start_edge()?
(and RunMode.EDGE i.o. RunMode.EDGECLI?)

xmatthias: well - the class (and file) are called EdgeCLI - to avoid confusion with Edge (in another namespace/package) - so that makes sense - and honestly, i just named it according to the class (but yes, using edge would make sense here).

freqtrade/optimize/__init__.py
xmatthias: well but what would you set it to? 0.1, 0.5, 1 or 20? 

Notice that the check is to *not* be unlimited (it raises an exception if it is Unlimited).

freqtrade/optimize/__init__.py
hroff-1902: I understand that don't like edge_cli in common. Or there should be hyperopt_cli and backtesting_cli  as well. But this is not for this PR...

LGTM 👍

PR #1880, opened 2019-05-25
minor: exchange debug logging humanized

hroff-1902:
DEBUG messages in exchange.py improved:
* one_call msecs humanized (prints humanized time duration after msecs)
* the since_ms datetime humanized (the humanized datetime is printed after timestamp)
* "BTC, 1h since _timestamp_" --> "pair BTC, interval 1h, since _timestamp_ (_humanized datetime_)" because current message can be read as "1 hour since _timestamp_" by a user which is not familiar with the bot internals. Same change in other messages where "BTC, 1h" is printed. Maybe some other similar messages can still be found...
* Tests adjusted to reflect changes in the log messages

How to see both variants of message with humanized since_ms:
```
python3 freqtrade -v backtesting -l
```
-- since_ms is None, prints `DEBUG - Fetching pair XMR/BTC, interval 1h, since None ...` -- same as now;
```
python3 freqtrade -v backtesting -r
```
-- since_ms is not None, prints `DEBUG - Fetching pair LTC/BTC, interval 1h, since 1558807200001 (2019-05-25T18:00:00+00:00) ...`

`one_call` is now logged as `DEBUG - one_call: 1800000000 msecs (20 days)`

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -510,7 +510,11 @@ class Exchange(object):
         _LIMIT = 500
 
         one_call = timeframe_to_msecs(ticker_interval) * _LIMIT
-        logger.debug("one_call: %s msecs", one_call)
+        logger.debug(
+            "one_call: %s msecs (%s)",
+            one_call,
+            arrow.utcnow().shift(seconds=one_call // 1000).humanize(only_distance=True)
+        )
         input_coroutines = [self._async_get_candle_history(
             pair, ticker_interval, since) for since in
             range(since_ms, arrow.utcnow().timestamp * 1000, one_call)]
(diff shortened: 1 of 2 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -541,7 +545,10 @@ def refresh_latest_ohlcv(self, pair_list: List[Tuple[str, str]]) -> List[Tuple[s
                     or self._now_is_time_to_refresh(pair, ticker_interval)):
                 input_coroutines.append(self._async_get_candle_history(pair, ticker_interval))
             else:
-                logger.debug("Using cached ohlcv data for %s, %s ...", pair, ticker_interval)
+                logger.debug(
+                    "Using cached ohlcv data for pair %s, interval %s ...",
+                    pair, ticker_interval
```
xmatthias: i am personally not a huge fan of wrapping log-lines where it's not neccessary as it unnecessarily increases the file/function length (but it's a real minor nitpick).

hroff-1902: it's longer than 100 chars if written in one line here

change pls it as you wish if do not understand how exactly it can be written here

xmatthias: ah ok, sorry missed that :+1:

PR #1891, opened 2019-05-29
persistence.init does not need the config dict

xmatthias:
## Summary
This will simplify the call to persistence.init() by not requiring a configuration dict.

It may not make sense now, but it's an extracted part from a larger refactoring of the plot-scripts (which do REALLY need a lot of love!!)

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -73,7 +73,8 @@ class FreqtradeBot(object):
 
         self.active_pair_whitelist: List[str] = self.config['exchange']['pair_whitelist']
 
-        persistence.init(self.config.get('db_url', None), self.config.get('dry_run'))
+        persistence.init(self.config.get('db_url', None),
+                         clean_open_orders=self.config.get('dry_run', False))
 
         # Set initial bot state from config
         initial_state = self.config.get('initial_state')
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -73,7 +73,7 @@ class FreqtradeBot(object):
 
         self.active_pair_whitelist: List[str] = self.config['exchange']['pair_whitelist']
 
-        persistence.init(self.config.get('db_url', None), self.config.get('dry_run', False))
+        persistence.init(self.config.get('db_url', None), self.config.get('dry_run'))
 
         # Set initial bot state from config
         initial_state = self.config.get('initial_state')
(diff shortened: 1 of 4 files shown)


freqtrade/freqtradebot.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, config: Dict[str, Any]) -> None:
 
         self.active_pair_whitelist: List[str] = self.config['exchange']['pair_whitelist']
 
-        persistence.init(self.config)
+        persistence.init(self.config.get('db_url', None), self.config.get('dry_run', False))
```
mishaker: dry_run is required. so why do you get a default value?

freqtrade/persistence.py
```diff
@@ -25,15 +25,14 @@
 _SQL_DOCS_URL = 'http://docs.sqlalchemy.org/en/latest/core/engines.html#database-urls'
 
 
-def init(config: Dict) -> None:
+def init(db_url: str, dry_run: bool = False) -> None:
```
mishaker: same here. if there should be a default value, then I don't see why it should be False

freqtrade/freqtradebot.py
xmatthias: Well - tbh - it mimics how it did work before (within `.init()`).

but you're right - i should reconsider...

freqtrade/persistence.py
xmatthias: This was intentional - and looking at the meaning of dry-run for this function it does also make sense (if it's True, it'll close all open orders).

scripts/plot_dataframe.py
```diff
@@ -55,7 +55,7 @@
 def load_trades(args: Namespace, pair: str, timerange: TimeRange) -> pd.DataFrame:
     trades: pd.DataFrame = pd.DataFrame()
     if args.db_url:
-        persistence.init(args.db_url, True)
+        persistence.init(args.db_url, False)
```
hroff-1902: minor: imho, that will be more readable to use the name of the second parameter here:
i.e. `persistence.init(args.db_url, clean_open_orders=False)`

rest LGTM 👍

freqtrade/freqtradebot.py
xmatthias: going to merge this as an out of band discussion (slack) with @mishaker  did confirm this as ready to merge.

PR #1893, opened 2019-05-29
refactoring download_backtest_data.py

### LINKED ISSUE #1857: download_backtest_data.py should fetch pairlist from config
... If it runs with config (or even if it runs with default config.json).

Exactly as freqtrade hyperopt and backtesting commands implement it.

The pairlist file should be read and used only if specified explicitely by the user in the command line if she needs to fetch a different (wider than in config) list of pairs or if the dynamic pairlist is set in her config.

hroff-1902:
This script really needs tests. It took so long time because I was manually testing all the cases where command line options override settings from the config...

* script now reads pair whitelist from the config, if config is specified (resolves #1857)
* --logfile, -v (-vv, -vvv) options added, as in the main freqtrade
* --export changed to --datadir (to be consistent with main freqtrade)

TBD:
* [x] maybe we should log and not print() in the script, since the screen is full with logs from the freqtrade modules
* <s>maybe it should load default config if not specified (and the --no-config option be added to work without config at all)</s> Cancelled, see comment below.

TODO:
* [x] adjust docs
* <s>bunch of tests for this tool. (Tests!) </s> As a result of conversation with @xmatthias on tests for scripts -- this task is postponed till the scripts are refactored to modules. Not achievable now.

### DIFF
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -470,7 +470,7 @@ class Arguments(object):
             '--days',
             help='Download data for given number of days.',
             dest='days',
-            type=int,
+            type=Arguments.check_int_positive,
             metavar='INT',
         )
         self.parser.add_argument(
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -467,7 +467,7 @@ class Arguments(object):
             help='File containing a list of pairs to download.',
             dest='pairs_file',
             default=None,
-            metavar='PATH',
+            metavar='FILE',
         )
         self.parser.add_argument(
             '--days',
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -443,7 +443,7 @@ class Arguments(object):
             help='Log to the file specified',
             dest='logfile',
             type=str,
-            metavar='FILE'
+            metavar='FILE',
         )
         self.parser.add_argument(
             '-c', '--config',
@@ -458,15 +458,12 @@ class Arguments(object):
             '-d', '--datadir',
             help='Path to backtest data.',
             dest='datadir',
-            default=None,
-            type=str,
             metavar='PATH',
         )
         self.parser.add_argument(
             '--pairs-file',
             help='File containing a list of pairs to download.',
             dest='pairs_file',
-            default=None,
             metavar='FILE',
         )
         self.parser.add_argument(
@@ -475,14 +472,11 @@ class Arguments(object):
             dest='days',
             type=int,
             metavar='INT',
-            default=None
         )
         self.parser.add_argument(
             '--exchange',
             help='Exchange name (default: %(default)s). Only valid if no config is provided.',
             dest='exchange',
-            type=str,
-            default='bittrex'
         )
         self.parser.add_argument(
             '-t', '--timeframes',
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -484,7 +484,6 @@ class Arguments(object):
                   Default: %(default)s.',
             choices=['1m', '3m', '5m', '15m', '30m', '1h', '2h', '4h',
                      '6h', '8h', '12h', '1d', '3d', '1w'],
-            default=['1m', '5m'],
             nargs='+',
             dest='timeframes',
         )
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -47,7 +47,7 @@ class Arguments(object):
 
         return self.parsed_arg
 
-    def parse_args(self) -> argparse.Namespace:
+    def parse_args(self, no_default_config: bool = False) -> argparse.Namespace:
         """
         Parses given arguments and returns an argparse Namespace instance.
         """
@@ -55,7 +55,7 @@ class Arguments(object):
 
         # Workaround issue in argparse with action='append' and default value
         # (see https://bugs.python.org/issue16399)
-        if parsed_arg.config is None:
+        if parsed_arg.config is None and not no_default_config:
             parsed_arg.config = [constants.DEFAULT_CONFIG]
 
         return parsed_arg
@@ -427,26 +427,24 @@ class Arguments(object):
             default=None
         )
 
-    def testdata_dl_options(self) -> None:
+    def download_data_options(self) -> None:
         """
         Parses given arguments for testdata download
         """
         self.parser.add_argument(
-            '--pairs-file',
-            help='File containing a list of pairs to download.',
-            dest='pairs_file',
-            default=None,
-            metavar='PATH',
+            '-v', '--verbose',
+            help='Verbose mode (-vv for more, -vvv to get all messages).',
+            action='count',
+            dest='loglevel',
+            default=0,
         )
-
         self.parser.add_argument(
-            '--export',
-            help='Export files to given dir.',
-            dest='export',
-            default=None,
-            metavar='PATH',
+            '--logfile',
+            help='Log to the file specified',
+            dest='logfile',
+            type=str,
+            metavar='FILE'
         )
-
         self.parser.add_argument(
             '-c', '--config',
             help='Specify configuration file (default: %(default)s). '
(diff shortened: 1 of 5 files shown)


freqtrade/arguments.py
```diff
@@ -456,7 +454,21 @@ def testdata_dl_options(self) -> None:
...
+            default=None,
+            type=str,
+            metavar='PATH',
+        )
+        self.parser.add_argument(
+            '--pairs-file',
+            help='File containing a list of pairs to download.',
+            dest='pairs_file',
+            default=None,
+            metavar='PATH',
```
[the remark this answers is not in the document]
hroff-1902: Done,
probably for --config as well (in a sep. PR)?

I also found that `pairs_file.exists()` should be changed to `pairs_file.is_file()` but left it to sep. PR to keep the changes observable.

scripts/download_backtest_data.py
```diff
@@ -60,55 +70,59 @@
...
+        pairs = list(set(json.load(file)))
 
+    pairs.sort()
 
 timerange = TimeRange()
 if args.days:
     time_since = arrow.utcnow().shift(days=-args.days).strftime("%Y%m%d")
     timerange = arguments.parse_timerange(f'{time_since}-')
 
+print(f'About to download pairs: {pairs}, intervals: {timeframes} to {dl_path}')
```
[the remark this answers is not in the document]
hroff-1902: Done here.

scripts/download_backtest_data.py
```diff
@@ -60,55 +70,59 @@
...
-    sys.exit(f'Directory {dl_path} does not exist.')
+dl_path = Path(config['datadir'])
 
 pairs_file = Path(args.pairs_file) if args.pairs_file else dl_path.joinpath('pairs.json')
-if not pairs_file.exists():
-    sys.exit(f'No pairs file found with path {pairs_file}.')
 
-with pairs_file.open() as file:
-    PAIRS = list(set(json.load(file)))
+if not pairs or args.pairs_file:
```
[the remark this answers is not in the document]
hroff-1902: Hmm, it should override -- if `--pairs-file` is specified, the pairs file should be used instead of the whitelist from the config... I guess I tested it. Do you obtain diff. behaviour?

`scripts/download_backtest_data.py --config config.json --exchange binance --pairs-file user_data/data/binance/pairs.json` -- uses pairs file and download data for binance instead of bittrex, the rest of settings is taken from config.json.

hroff-1902:
Let's discuss fetching of default config for this tool before merge when the rest is okay.

hroff-1902:
Docs adjusted with description for new -c/--config and --export changed to --datadir. I tried to keep the description compact, so -v and --logfile options are not added there.

From my point of view, it's ready for merge. Besides the pair-list point ("a specified pair-file should overwrite configuration", see above), but I think it works as needed -- if you use the --pairs-file option it overrides the list from the config (if I don't miss anything...).

scripts/download_backtest_data.py
```diff
@@ -1,50 +1,60 @@
...
     # Ensure we do not use Exchange credentials
+    config['exchange']['dry_run'] = True
     config['exchange']['key'] = ''
     config['exchange']['secret'] = ''
+
+    if args.exchange:
+        config['exchange']['name'] = args.exchange
+
+    pairs = config['exchange']['pair_whitelist']
+    timeframes = [config['ticker_interval']]
```
xmatthias: `-t 1d` is ineffective and uses 5m from the configuration if a configuration is used (i guess you could simply check here `if not timeframes: ...`, simillar to the check for `if args.exchange:` two lines above.

(and honestly, -t is one of the best options this script offers as we can download 5m,15m,1d data in one go).

freqtrade/arguments.py
```diff
@@ -456,7 +454,21 @@ def testdata_dl_options(self) -> None:
             type=str,
             metavar='PATH',
         )
-
+        self.parser.add_argument(
+            '-d', '--datadir',
```
xmatthias: At a later point, we should try to combine as many options as possible (--datadir is used 4 times in this file, and i guess it's similar with many other options) (however not for this PR ...)

hroff-1902: yeah, it should go somehow to a common part of the options (but not for all commands though). Further improvement.

scripts/download_backtest_data.py
hroff-1902: Will check this a bit later, busy today...

As a possible further improvement -- if used with the strategy, this script should fetch informative pairs from it and add to the download list along with their timeframes that can differ from timeframes set in the config... What do you think?

xmatthias: noone said it needs fixing immediately, so no worries

i wouldn't don't overcomplicate this script tbh - a method to simply fetch all pairs (for a pairlist) and we should be good ... theres a ton of other things to improve ;)

hroff-1902: Handling of exchange is also broken. If config is used, then the script currently works with bittrex (i.e. default value) even if binance is set in the config. 

All this needs further refactoring... I will probably move handling of the parameters to freqtrade/configuration.py, as for other parameters and options for the main freqtrade. This seems to be most efficient approach and will help make the script a command of main freqtrade in the future...

xmatthias: I disagree (and i did test it too), lines 49/50 override the exchange that is set in the configuration with what is set in the parameter.

hroff-1902: Yes, but if exchange is not given in the command line, it's not taken from the config but a default bittrex is used. That'a because exchange defaults to bittrex in the freqtrade/arguments.py and args.exchange is not empty when it's not specified in the command line...

gitgitgat:
Maybe it is not referencing the correct path to fetch the config.json file

hroff-1902:
> Maybe it is not referencing the correct path to fetch the config.json file

Did not get it, sorry.

scripts/download_backtest_data.py
xmatthias: oh sorry, didn't think about that direction ...

hroff-1902: Something like this, sure -- remove default.
But I'm now thinking that such approach (removal of default in parseargs) should also be implemented for all other options (well, for almost all of them -- the `-v` is probably is a good excemption, it's well working) and assigning of defaults should be moved completely to configuration.py, where they are merged with config... Just because now we have, for example
```
Parameter -s/--spaces detected: all
```
in the log for the hyperopt when no -s option is used...

Of course, for other options it should be done in a sep. PR, so I'm still evaluating all this to find a better common approach...

hroff-1902:
Let it be as in the last commit. I need to move forward with it.

Small issues (logging, romoval of the duplicated code) will be refactored later when this script moves to the freqtrade/main command.

scripts/download_backtest_data.py
```diff
@@ -1,55 +1,64 @@
...
     config['stake_currency'] = ''
     # Ensure we do not use Exchange credentials
+    config['exchange']['dry_run'] = True
     config['exchange']['key'] = ''
     config['exchange']['secret'] = ''
+
+    pairs = config['exchange']['pair_whitelist']
+
+    # Don't fail if ticker_interval is not in the configuration
+    timeframes = [config.get('ticker_interval')]
```
xmatthias: I (accidentally) found another problem with timeframe handling.

i had a configuration without ticker_interval i passed to the download script, which had it fail on this line (it assumed ticker_interval is always there) so fails.

I also moved assigning timeframes from args to after this section, so it'll eventually override configuration if specified explicitly.

hroff-1902: > I also moved assigning timeframes from args to after this section, so it'll eventually override configuration if specified explicitly.

That's wrong because it will use ['1m', '5m'] always when calles without -t option, even if config contains ticker_interval set in it.

scripts/download_backtest_data.py
```diff
@@ -127,7 +127,7 @@
 
             logger.info(f'Downloading pair {pair}, interval {ticker_interval}.')
             download_pair_history(datadir=dl_path, exchange=exchange,
-                                  pair=pair, ticker_interval=ticker_interval,
+                                  pair=pair, ticker_interval=str(ticker_interval),
```
hroff-1902: Is not it a string always? ('1h', '5m'?)

scripts/download_backtest_data.py
xmatthias: darn yes, thanks for noticing!

and yes, subcommand it shall be soon

scripts/download_backtest_data.py
xmatthias: technically yes - but mypy does not know as args.ticker_interval returns `Optional[Any]` ... not List[str] ...

Issue #1874, opened 2019-05-24
Retrieve bid/ask/last price in the strategy


yperfanov:
I need to check if the current price is not to far from the level when the signal was generated. There could be big differences especially when trading on a higher time frame(30m, 1h, 1d).
So how I am supposed to retrieve and check the current bid/ask/last price in the strategy?

xmatthias:
Currently the last tick price is not available to the strategy.

yperfanov:
I made it work for the best bid and ask. It took me time to find out that the strategy has a DataProvider field, but after that was easy.
So I implemented the orderbook() method in /freqtrade/data/dataprovider.py as follows:
```python
    def orderbook(self, pair: str, max: int):
        """
        return latest orderbook data
        """
        return self._exchange.get_order_book(pair, max)
```
Then in the strategy whenever I need it (in my case in populate_buy_trend()) I do:
```python
      ob = self.dp.orderbook(metadata['pair'], 1)
      top_bid = ob['bids'][0][0] # best bid
```

hroff-1902:
would you make a pull request with this enhancement?

hroff-1902:
Closing it as the pull request implementing the enhancements was successfully merged.

PR #1901, opened 2019-06-02
Bid ask strategy

yperfanov:
## Summary
Orderbook available in the strategy.

Solve the issue: #1874

## What's new?
1. Implemented freqtrade/data/dataprovider.py ordebook() method.
2. Added commented example how to use it to retrieve best bid and ask in user_data/strategies/test_strategy.py.

### DIFF
--- a/user_data/strategies/test_strategy.py
+++ b/user_data/strategies/test_strategy.py
@@ -253,6 +253,15 @@ class TestStrategy(IStrategy):
         dataframe['ha_low'] = heikinashi['low']
         """
 
+	# Retrieve best bid and best ask
+        # ------------------------------------
+        """
+        ob = self.dp.orderbook(metadata['pair'], 1)
+        print(ob)
+        dataframe['best_bid'] = ob['bids'][0][0]
+        dataframe['best_ask'] = ob['asks'][0][0]
+        """
+
         return dataframe
 
     def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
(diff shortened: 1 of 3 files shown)


xmatthias:
Great :+1: 

It would be great to have this documented [here](
https://github.com/freqtrade/freqtrade/blob/develop/docs/strategy-customization.md#additional-data-dataprovider)

Note that in the strategy, i would always advise to check if `self.dp` is available.

user_data/strategies/test_strategy.py
```diff
@@ -253,6 +253,15 @@ def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame
         dataframe['ha_low'] = heikinashi['low']
         """
 
+	# Retrieve best bid and best ask
+        # ------------------------------------
+        """
+        ob = self.dp.orderbook(metadata['pair'], 1)
+        print(ob)
```
hroff-1902: Super, thanks!

This print() should be commented out for merging to the freqtrade codebase.
Or removed.

xmatthias:
yes but the existing section 5 lines above the new one also uses this format. Also it's a comment / documentation section, so i wouldn't see this too critical.

hroff-1902:
> yes but the existing section 5(ish) lines above the new one also uses this format. Also it's a comment / documentation section, so i wouldn't see this too critical.

Sure, LGTM 👍 
(How to change the 'type' of the review comment above from 'request changes' to 'approve'?)

xmatthias:
i think you can approve when you go to the files-section again ... but you can't change the previous review.

hroff-1902:
> i think you can approve when you go to the files-section again ... but you can't change the previous review.

How's life complicated... :sarcasm

PR #1914, opened 2019-06-08
Metadata timeframe

### LINKED ISSUE #1900: Hyperopt Resamples
Python Version: 3.6.7
    CCXT version: 1.18.507
    Branch: Develop
    Last Commit ID: 4ae743e

  
self.ticker_interval is not implemented in the hyperopt interface, so you cannot resample data to higher timeframes.

hroff-1902:
Resolves #1900 

This PR adds timeframe used (in the str form, i.e. "1m", "1h", etc) to the metadata, passed to the populate_*() methods in the strategies and hyperopts.

You can access it in the populate_*() methods as simple as
```
timeframe = metadata['timeframe']
```

TODO:
* [x] adjust docs
* [ ] adjust existing tests to check for it in the metadata passed to the populate_*() methods
* [ ] (?) add new tests

### DIFF
--- a/docs/strategy-customization.md
+++ b/docs/strategy-customization.md
@@ -230,12 +230,16 @@ Common values are `"1m"`, `"5m"`, `"15m"`, `"1h"`, however all values supported 
 Please note that the same buy/sell signals may work with one interval, but not the other.
 This setting is accessible within the strategy by using `self.ticker_interval`.
 
-### Metadata dict
+### Metadata dictionary
 
-The metadata-dict (available for `populate_buy_trend`, `populate_sell_trend`, `populate_indicators`) contains additional information.
-Currently this is `pair`, which can be accessed using `metadata['pair']` - and will return a pair in the format `XRP/BTC`.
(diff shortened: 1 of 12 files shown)


xmatthias:
I think using metadata for this overcomplicates things, since this is available in the strategy as `self.ticker_interval` - which is either used as is from the strategy, or overwritten from the configuration.
Using metadata (to me) only makes sense for things which change between calls (e.g. each function in the strategy is called for multiple pairs) - but it'll always be the same ticker-interval, so setting it at strategy-class/instance level is more efficient (we don't have to add it to a dict each iteration).


The issue you're pointing to ( #1900 ) notes that `self.ticker_interval` is not available during hyperopt, which is (probably - without checking yet) correct - but adding it should be very simple (i guess in HyperoptResolver) ...

hroff-1902:
Oh, yes, I agree.

I was thinking that at least populate_indicators() is also called for informative pairs so diff. timeframe is applicable.

I'll cherry-pick some minor cosmetics changes from this PR and close it then...

PR #1925, opened 2019-06-11
debug logging for IStrategy.advise_*()

hroff-1902:
...for better tracking of freqtrade actions.

### DIFF
--- a/freqtrade/tests/strategy/test_strategy.py
+++ b/freqtrade/tests/strategy/test_strategy.py
@@ -63,15 +63,14 @@ def test_search_strategy():
 
 def test_load_strategy(result):
     resolver = StrategyResolver({'strategy': 'TestStrategy'})
-    metadata = {'pair': 'ETH/BTC'}
-    assert 'adx' in resolver.strategy.advise_indicators(result, metadata=metadata)
+    assert 'adx' in resolver.strategy.advise_indicators(result, {'pair': 'ETH/BTC'})
 
 
 def test_load_strategy_byte64(result):
     with open("freqtrade/tests/strategy/test_strategy.py", "r") as file:
         encoded_string = urlsafe_b64encode(file.read().encode("utf-8")).decode("utf-8")
     resolver = StrategyResolver({'strategy': 'TestStrategy:{}'.format(encoded_string)})
-    assert 'adx' in resolver.strategy.advise_indicators(result, 'ETH/BTC')
+    assert 'adx' in resolver.strategy.advise_indicators(result, {'pair': 'ETH/BTC'})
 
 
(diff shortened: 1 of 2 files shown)


hroff-1902:
@xmatthias please review if I was right fixing metadata in test_deprecate_populate_indicators().

Maybe that test is intended to test the case when the string is passed instead of a dict with the pair name... (But did not get the purpose of this test in this case, since freqtrade does not pass strings)

freqtrade/tests/strategy/test_strategy.py
```diff
@@ -380,7 +379,7 @@ def test_deprecate_populate_indicators(result):
     with warnings.catch_warnings(record=True) as w:
         # Cause all warnings to always be triggered.
         warnings.simplefilter("always")
-        resolver.strategy.advise_buy(indicators, 'ETH/BTC')
+        resolver.strategy.advise_buy(indicators, {'pair': 'ETH/BTC'})
```
xmatthias: this was a typo, and wasn't relevant as in this cases we use the legacy strategy, which does not pass pair.

hroff-1902:
> I assume the purpose of this is to detect when TA is running? but isn't line 171 of this file good enough for that (maybe if we add the pair to that log-message)?
> 
> These 3 functions are almost always (apart from backtesting) called one after the other (see above) - so won't this info be redundant?

Well, they are also called from hyperopts. More precisely -- will be called. Now the populate_*() are assigned directly to hyperopt's advise_*(). I'm going to change it a bit in the nearest future -- in order to make another attempt to allow to use self.strategy.ticker_interval in the hyperopts. I'm also going to allow use custom strategy (and its populate_*()) in the hyperopts instead of forcing the user to copy code of those functions from the strategy to hyperopt. At least using default strategy always in the hyperopt seems to be, mmm, strange. So I need better diagnostics of how all this is called in the hyperopts, thus these additional messages are not redundant.

hroff-1902:
Even without those messages, the `TA Analysis Launched` message should be enhanced with the pair name it's called for, to be more useful... For backtesting, `populate_indicators` is called in a different place, not along with `populate_buy/sell`...

xmatthias:
yes - `TA Analysis Launched ` should be enhanced.

I answered to your hyperopt comment in #1775 since it is a bit ot here (and it may be usefull later, and i find that comments in PR's difficult to refind once they are merged).

if you think it still makes sense i'm fine to merge this.

hroff-1902:
> if you think it still makes sense i'm fine to merge this.

Reviewed the logs once again for diff. commands -- yes, please merge, the order in which those functions are called diffes so I guess it makes sense and seems to be useful for me, for development (these are just debug messages).

PR #1926, opened 2019-06-11
Enhance check_exchange()

hroff-1902:
Check exchange name against:
* list of 'bad' exchanges, known to have critical problems with the bot. Warns user. Set currently to `['bitmex',]`, extend it if you know of other bad exchanges.
* list of officially supported exchanges (`['binance', 'bittrex']`), Warns user to be careful running the bot if exchange is not in this list.
* list of exchanges, supported by ccxt (i.e. 'known' for the bot). Current behaviour is kept (raises exception if misspelled or not supported by ccxt), wording changed ('supported' --> 'known').

The check_for_bad parameter is left for scripts -- if set to False, it will allow to run the download script for 'bad' exchanges and not to warn the user of it.

TODO:
* a couple of more tests for all cases (know but not officially supported exchange, etc.)

### DIFF
--- a/freqtrade/configuration.py
+++ b/freqtrade/configuration.py
@@ -397,20 +397,19 @@ class Configuration(object):
                     f'{", ".join(available_exchanges())}'
             )
 
-        logger.info(f'Exchange "{exchange}" is supported by ccxt '
-                    f'and therefore available for the bot.')
+        if check_for_bad and is_exchange_bad(exchange):
+            logger.warning(f'Exchange "{exchange}" is known to not work with the bot yet. '
+                           f'Use it only for development and testing purposes.')
+            return False
 
         if is_exchange_officially_supported(exchange):
             logger.info(f'Exchange "{exchange}" is officially supported '
                         f'by the Freqtrade development team.')
         else:
-            logger.warning(f'Exchange "{exchange}" is not officially supported. '
+            logger.warning(f'Exchange "{exchange}" is supported by ccxt '
+                           f'and therefore available for the bot but not officially supported '
+                           f'by the Freqtrade development team. '
                            f'It may work flawlessly (please report back) or have serious issues. '
                            f'Use it at your own discretion.')
-
-        if check_for_bad and is_exchange_bad(exchange):
-            logger.warning(f'Exchange "{exchange}" is known to not work with the bot yet. '
-                           f'Use it only for development and testing purposes.')
-            return False
 
         return True
--- a/freqtrade/configuration.py
+++ b/freqtrade/configuration.py
@@ -13,7 +13,8 @@ from jsonschema import Draft4Validator, validators
 from jsonschema.exceptions import ValidationError, best_match
 
 from freqtrade import OperationalException, constants
-from freqtrade.exchange import is_exchange_supported, supported_exchanges
+from freqtrade.exchange import (is_exchange_bad, is_exchange_known,
+                                is_exchange_officially_supported, known_exchanges)
 from freqtrade.misc import deep_merge_dicts
 from freqtrade.state import RunMode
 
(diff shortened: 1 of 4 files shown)


freqtrade/configuration.py
```diff
@@ -375,22 +376,41 @@ def get_config(self) -> Dict[str, Any]:
 
         return self.config
 
-    def check_exchange(self, config: Dict[str, Any]) -> bool:
+    def check_exchange(self, config: Dict[str, Any], check_for_bad: bool = True) -> bool:
```
xmatthias: what is `check_for_bad` good for? i don't see it used / changed anywyere ... ?

freqtrade/configuration.py
```diff
@@ -375,22 +376,41 @@ def get_config(self) -> Dict[str, Any]:
...
             raise OperationalException(
-                exception_msg
+                    f'Exchange "{exchange}" is not supported by ccxt '
+                    f'and not known for the bot.\n'
+                    f'The following exchanges are supported by ccxt: '
+                    f'{", ".join(known_exchanges())}'
             )
 
-        logger.debug('Exchange "%s" supported', exchange)
+        logger.info(f'Exchange "{exchange}" is supported by ccxt and known for the bot.')
```
xmatthias: This line and the line 2 below create a double-confirmation, which i don't think is needed.

also "known for the bot" seems oddly worded. I like the other message better :) 

```
2019-06-12 19:31:32,208 - freqtrade.configuration - INFO - Checking exchange...
2019-06-12 19:31:32,208 - freqtrade.configuration - INFO - Exchange "binance" is supported by ccxt and known for the bot.
2019-06-12 19:31:32,208 - freqtrade.configuration - INFO - Exchange "binance" is officially supported by the Freqtrade development team.
```

freqtrade/configuration.py
```diff
@@ -375,22 +376,41 @@ def get_config(self) -> Dict[str, Any]:
...
-                            f'{", ".join(supported_exchanges())}'
+        logger.info("Checking exchange...")
 
-            logger.critical(exception_msg)
+        exchange = config.get('exchange', {}).get('name').lower()
+        if not is_exchange_known(exchange):
             raise OperationalException(
-                exception_msg
+                    f'Exchange "{exchange}" is not supported by ccxt '
+                    f'and not known for the bot.\n'
```
xmatthias: "and therefore not available for the bot" ?

freqtrade/configuration.py
```diff
@@ -375,22 +376,41 @@ def get_config(self) -> Dict[str, Any]:
...
+        logger.info(f'Exchange "{exchange}" is supported by ccxt and known for the bot.')
+
+        if is_exchange_officially_supported(exchange):
+            logger.info(f'Exchange "{exchange}" is officially supported '
+                        f'by the Freqtrade development team.')
+        else:
+            logger.warning(f'Exchange "{exchange}" is not officially supported '
+                           f'by the Freqtrade development team. '
+                           f'It may work with serious issues or not work at all. '
+                           f'Use it at your own discretion.')
```
xmatthias: I would word this message in a positive form. This makes it look like it is almost guaranteed to not work.

something like the following, maybe?

```
Exchange  ... is not officially supported yet.
It may work flawlessly (please report back), or have serious issues.
Use it at your own discretion.
```

freqtrade/configuration.py
hroff-1902: For scripts (like download, etc.) we may allow to run it for "bad" exchanges, allowing to download (or try to download) data when the main freqtrade logic will fail.
The scripts, thus, should call it with `check_for_bad=False`.

hroff-1902: The idea was to not print this message for the scripts. Maybe it's redundant since even download will fail for some of the 'bad' exchanges. So, what's the consensus -- let's remove this parameter?

freqtrade/configuration.py
hroff-1902: Agree about more positive form, consider my messages just as temp. examples created during development...

@xmatthias that's simply a question of wording, we'll discuss all the nuances of the phrases here in many comments, could you better change wording right in the branch instead? That would be more efficient...

hroff-1902: The main idea here was to avoid calling exchanges supported by ccxt simply as 'supported' because the only supported exchanges for freqtrade are binance and bittrex...

hroff-1902:
@xmatthias I made some changes, please review the latest version

freqtrade/configuration.py
xmatthias: ah works for me - just asking as i didn't see it used anywhere :+1:

freqtrade/configuration.py
```diff
@@ -375,22 +376,41 @@ def get_config(self) -> Dict[str, Any]:
...
-                exception_msg
+                    f'Exchange "{exchange}" is not supported by ccxt '
+                    f'and therefore not available for the bot.\n'
+                    f'The following exchanges are supported by ccxt: '
+                    f'{", ".join(available_exchanges())}'
             )
 
-        logger.debug('Exchange "%s" supported', exchange)
+        logger.info(f'Exchange "{exchange}" is supported by ccxt '
+                    f'and therefore available for the bot.')
```
xmatthias: I would move this message to the "else" tree below... it's only relevant if its not binance/bittrex ...

hroff-1902: > I would move this message to the "else" tree below... it's only relevant if its not binance/bittrex ...

So, would

`Exchange "{exchange}" is available for the bot but not officially supported by the Freqtrade development team. It may work flawlessly <...>`

be okay in that else part?

Or
`Exchange "{exchange}" is supported by ccxt and therefore available for the bot but not officially supported by the Freqtrade development team. It may work flawlessly <...>`
?

xmatthias: i guess i would prefer the 2nd message - but really it's up to you (just would try to avoid having multiple messages saying the same thing for supported exchanges).

hroff-1902: Done, pls review.

Actually, we can now check the value returned by this function and stop working for 'bad' exchanges (main, backtest and hyperopt; util subcommands should not check for it). In this case a bool 'work_with_bad_exchanges' configuration setting can be introduced in the internals section of the config for development purposes.

PR #1927, opened 2019-06-12
list-exchanges subcommand added

hroff-1902:
In order to get rid of awful `scripts/get_market_pairs.py`. Other scripts can also be moved to main freqtrade subcommands.

Done:

* `list-exchanges` subcommand added; was part of scripts/get_market_pairs.py functionality
```
freqtrade list-exchanges
freqtrade list-exchanges -1
```

Todo:
* `list-pairs` subcommand with `--exchange`, `--quote-currency`, `--base-currency` options.

@xmatthias pls review it. It in principle is a template for moving other scripts to subcommands.
I think there should be an additional directory `utils` created for all this stuff with every subcommand in a diff. file.
Maybe `cli-utils` should be used everywhere in the code to name it instead of `utils`.

### DIFF
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -341,7 +341,7 @@ class Arguments(object):
         Parses given arguments for the list-exchanges command.
         """
         parser.add_argument(
-            '-1',
+            '-1', '--one-column',
             help='Print exchanges in one column',
             action='store_true',
             dest='print_one_column',
@@ -376,7 +376,7 @@ class Arguments(object):
         self.hyperopt_options(hyperopt_cmd)
 
         # Add list-exchanges subcommand
-        list_exchanges_cmd = subparsers.add_parser('list-exchanges', help='List known exchanges.')
+        list_exchanges_cmd = subparsers.add_parser('list-exchanges', help='Print available exchanges.')
         list_exchanges_cmd.set_defaults(func=start_list_exchanges)
         self.list_exchanges_options(list_exchanges_cmd)
 
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -376,7 +376,10 @@ class Arguments(object):
         self.hyperopt_options(hyperopt_cmd)
 
         # Add list-exchanges subcommand
-        list_exchanges_cmd = subparsers.add_parser('list-exchanges', help='Print available exchanges.')
+        list_exchanges_cmd = subparsers.add_parser(
+            'list-exchanges',
+            help='Print available exchanges.'
+        )
         list_exchanges_cmd.set_defaults(func=start_list_exchanges)
         self.list_exchanges_options(list_exchanges_cmd)
 
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -335,12 +335,25 @@ class Arguments(object):
             metavar='INT',
         )
 
+    @staticmethod
+    def list_exchanges_options(parser: argparse.ArgumentParser) -> None:
+        """
+        Parses given arguments for the list-exchanges command.
+        """
+        parser.add_argument(
+            '-1',
+            help='Print exchanges in one column',
+            action='store_true',
+            dest='print_one_column',
+        )
+
     def _build_subcommands(self) -> None:
         """
         Builds and attaches all subcommands
         :return: None
         """
         from freqtrade.optimize import start_backtesting, start_hyperopt, start_edge
+        from freqtrade.utils import start_list_exchanges
 
         subparsers = self.parser.add_subparsers(dest='subparser')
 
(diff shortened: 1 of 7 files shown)


freqtrade/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+def start_list_exchanges(args: Namespace) -> None:
+    """
+    Start listing known exchanges
+    :param args: Cli args from Arguments()
+    :return: None
+    """
+
+    # Initialize configuration
+    _ = setup_configuration(args, RunMode.OTHER)
```
xmatthias: I don't think you need this.
`supported_exchanges()` only uses the static method from ccxt - no config or anything involved at all.

freqtrade/utils.py
```diff
@@ -0,0 +1,44 @@
...
+from freqtrade.exchange import supported_exchanges
+from freqtrade.state import RunMode
+
+
+logger = logging.getLogger(__name__)
+
+
+def setup_configuration(args: Namespace, method: RunMode) -> Dict[str, Any]:
+    """
+    Prepare the configuration for the Hyperopt module
```
xmatthias: Can this be reused for hyperopt ... ? 
anyway i don't think its necessary for this subcommand (see other comment)

hroff-1902:
> is your idea with this new file to add all/most helper commands we discussed (download, plot, ...) here? 

Exactly, right. ;)
Here is a kind of template. Thus, the config instance is created, will cleanup it in the future.

If we place them in a new directory, it can be named `cli`, `utils` or `cli-utils`...

freqtrade/utils.py
hroff-1902: > Can this be reused for hyperopt ... ?
> anyway i don't think its necessary for this subcommand (see other comment)

Also thought about this. Even without being called from hyperopt, the next step of getting rid of get_market_pairs script will be adding another subcommand, `list-pairs`, as I said. This function will definitely be called from its implementation.

freqtrade/utils.py
xmatthias: Personally, i would remove this call, since it's not needed for this (leave the setup_configuration part in maybe, we'll need that for other things...)

It causes problem when running `freqtrade list-exchanges` without a config.json available. 
i don't, as i have ~20 i use for testing, but (intentionally) none with the default name.

I assume "list-exchanges" is one of the first thing you run when getting started (to actually decide on how to bulid a configuration), so you probably don't have a configuration yet, so this is confusing.

Apart from that - i would leave the file as `utils.py` for now (we can move it to a folder/module later, when we have more commands...

freqtrade/arguments.py
```diff
@@ -335,12 +335,25 @@ def hyperopt_options(parser: argparse.ArgumentParser) -> None:
             metavar='INT',
         )
 
+    @staticmethod
+    def list_exchanges_options(parser: argparse.ArgumentParser) -> None:
+        """
+        Parses given arguments for the list-exchanges command.
+        """
+        parser.add_argument(
+            '-1',
```
xmatthias: maybe add a `--one-column` here? 
also, maybe it's better to call it `-l` - as "list"? 

having a number as argument was surprising as i've not seen it around so far (although it works just fine, technically)

hroff-1902: > having a number as argument was surprising as i've not seen it around so far (although it works just fine, technically)

Ermmm, `man ls`... ;)
This often (more or less) used in unix core utils to specify output in one column...

But, yes, it may be surprising and, plus, it's not recognized well from '-l'. Okay, I will change it to `--one-column`, it's a good name for this option. (Maybe make it `-1/--one-column`?)

freqtrade/utils.py
hroff-1902: > Personally, i would remove this call, since it's not needed for this (leave the setup_configuration part in maybe, we'll need that for other things...)

Already done it. It (the configuration) will be used in `list-pairs` to initialize exchange.

The `setup_configuration()` may be used in `start_hyperopt/backtesting` and in all new subcommands, like `list-pairs`, `plot-*`, `download`, `inspect-datadir`, etc.

hroff-1902:
Tests. I don't think I'm able to write tests for new subcommands from scratches...

freqtrade/arguments.py
xmatthias: > (Maybe make it -1/--one-column?) 

sure

hroff-1902: > sure

Done. ;)

xmatthias:
> Tests. I don't think I'm able to write tests for new subcommand from scratches...

Just belive in yourself - writing tests is just the same than writing normal code.
it's mostly copy/paste (in this case from how backtesting is started) or google (how to assert print outputs) ... 

Also we don't necessarily need 100% test coverage - since that'll only make sure every branch is executed, not that it'll make sense in the greater context necessarily ...

hroff-1902:
> > Tests. I don't think I'm able to write tests for new subcommand from scratches...
> 
> Just belive in yourself - writing tests is just the same than writing normal code.
> it's mostly copy/paste (in this case from how backtesting is started) or google (how to assert print outputs) ...
> 
> Also we don't necessarily need 100% test coverage - since that'll only make sure every branch is executed, not that it'll make sense in the greater context necessarily ...

Yeah, this test_utils.py is pretty simple. I'm not familiar with all those mockers and other pytest infrastructure so that was the problem... Thanks!

Do we need more tests or anything else for this PR? I read the code and reviewed your changes, LGTM 👍 (no Make review buttom for them)...

PR #1934, opened 2019-06-15
Edge cli should override stake_amount

mishaker:
## Summary
When running edge cli if the stake amount is not "unlimited" we have: 
```
ERROR - Edge works only with unlimited stake amount
```

This is not necessary as edge cli is just informative and does nothing in live.

The stake amount should be set to unlimited before running egde cli.


hroff-1902:
Hooray, that was really annoying! I was also going to make this or similar change...

Issue #1505, opened 2019-01-22
Feature: Plotting Web UI / Strategy Test data Selector


Axel-CH:
## INTRO
Visualisation and observation is important in Developing a trading strategy. Currently on freqtrade, to visualize a graph and trades, the user need to generate and html file with a command line (plot-dataframe.py).
It's working well, but the whole process is not efficient and user friendly. Even more so during multiple tests.
Furthermore, for my ongoing experimentations, I need a tool to quickly select and extract specific price movement history and indicators in dataframe.

## GOAL:
- Quickly visually check strategy behavior
- Facilitate the selection and backup of pertinent price movement / paterns

### MAIN FEATURES
_#Input data pipe 1_

- [x]  Import json ticker history file 
- [x] plot candles
- [x] Select strategy

_#Input data pipe 2_

- [x]  Enter Pairs
- [x]  Select strategy, Timeframe
- [x]  Enter Pairs

_# Backtest execution, visualisation and export_
- [x] Run a backtest
- [ ] Plot candles with strategy indicators
- [ ] Select the important movement (time interval) to export
- [ ] Export Dataframes to json
- [ ] Export Dataframes Plots to html

### OTHER FEATURES
- [ ] Display standard indicators
- [ ] Select a date range
- [ ] Filter pairs by volume
- [ ] Realistic mode (remove same consecutive signals)
- [ ] Beter UX, collapsible sidebar for conf, plots on full screen

## SCREENSHOOT
<img width="956" alt="2019-01-21 23_25_14-dash" src="https://user-images.githubusercontent.com/7593404/51510723-9f09c280-1dd4-11e9-92ca-48b7a91f4c77.png">

## Step 2: Describe your environment
  * Python Version: Python 3.6.5 :: Anaconda, Inc.
  * CCXT version: ccxt==1.17.352
  * Branch: feature/plot_web_ui (on fork)
  * Last Commit ID: 54109abc6adf1ceb919b572b3972fc6ffedeb23c (fork)

Note that the Dash module used for the UI use a flask server.
 
## Step 3: Describe the problem:
I'm not able to execute a backtest due to an error with the async ccxt feature.

### Steps to reproduce:
  1. git clone https://github.com/Axel-CH/freqtrade.git
  2. git checkout feature/plot_web_ui
  3. pip install -e .    (to install dash modules)
  4. python .\scripts\plot_df_web_ui.py
  5. Open navigator on http://127.0.0.1:8050/
  6. Fill the fields
[...]

PR #1937, opened 2019-06-16
move parts of scripts/plot_dataframe.py to main bot code

xmatthias:
## Summary
Move most logic from `scripts/plot_dataframe.py`to a "plot"-module so we can test it properly and update plotly without fear of breaking everything (yes, that did happen before).

simplify  #1505 since the default plot-stuff can be imported easily

## Quick changelog

- Move most plot_dataframe.py logic to plot submodule
- Bot can still run without having plotly installed

### Next steps (not in this pr!):

- continue this work and disect `plot_profit.py`
- remove scripts and move these commands as freqtrade subcommands (e.g. `freqtrade plot --pairs  ...`)
- Once everything is complete, document how to use these plot-logics to simply create plots in notebooks


freqtrade/arguments.py
```diff
@@ -444,6 +444,7 @@ def common_scripts_options(self, subparser: argparse.ArgumentParser = None) -> N
             '-p', '--pairs',
             help='Show profits for only this pairs. Pairs are comma-separated.',
             dest='pairs',
+            required=True
```
hroff-1902: `--pairs` should not be required if we are going to fetch the whitelist from the config...

xmatthias: good point - plot_dataframe did not support loading from whitelist previously, but i've added this now.

PR #1938, opened 2019-06-16
minor: setup_configuration() cleanup

hroff-1902:
* slip in a comment fixed
* setup_configuration() from utils.py renamed to setup_utils_configuration();
* setting `dry_run = True` added to it;
* use it in `optimize/__init__.py` to avoid duplication of the code;
* test adjusted

This function is intended for the use in all other new utility subcommands that require configuration in the dry_run mode without real exchange key and secret (i.e. all besides simple list-exchanges which runs without configuration).

### DIFF
--- a/freqtrade/tests/test_utils.py
+++ b/freqtrade/tests/test_utils.py
@@ -1,17 +1,18 @@
-from freqtrade.utils import setup_configuration, start_list_exchanges
+from freqtrade.utils import setup_utils_configuration, start_list_exchanges
 from freqtrade.tests.conftest import get_args
 from freqtrade.state import RunMode
 
 import re
 
 
-def test_setup_configuration():
+def test_setup_utils_configuration():
     args = [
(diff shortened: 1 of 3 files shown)


freqtrade/tests/test_utils.py
```diff
@@ -1,17 +1,18 @@
...
     args = [
         '--config', 'config.json.example',
     ]
 
-    config = setup_configuration(get_args(args), RunMode.OTHER)
+    config = setup_utils_configuration(get_args(args), RunMode.OTHER)
     assert "exchange" in config
+    assert config['exchange']['dry_run'] is True
     assert config['exchange']['key'] == ''
     assert config['exchange']['secret'] == ''
```
xmatthias: On a side note (just realized this now)  we should also zero "password" and "uid" here, depending on the exchange these options can be exchange "secrets" too (see exchange.py - L162ff).

Issue #1945, opened 2019-06-17
[idea/discussion] new configuration / args system


xmatthias:
## Problem

Since we've been struggling lately with the arguments / config-options, i've been giving some thought to this problem.

We mostly define an argument in multiple places so it shows up as

1. CLI argument (different overlapping options between the different sub-commands)
1. config option
1. optionally, overridable by the strategy

We define this in multiple places (`arguments.py`, `configuration.py`, `strategy_resolver.py`, and currently some of the scripts).

Ideally, we should have one place to look for these things.

## Idea

Having a simple Class which defines everything required, and arguments, config options (and overrides) and strategy parameters (and overrides) are then created automatically from this.

## Sample mock implementation

*Note* this is mock code, which will not actually work. Also, there is currently no implementation behind "FTConfiguration".

``` python
FTConfiguration(name="strategy",
                cli=["-s", "--strategy"],
                help="Specify strategy class name (default: %(default)s).",
                default="DefaultStrategy",
                type=str,
                usages=["main", "backtesting", "edge-cli", "plot"],
                strategy_argument=False,
                )
FTConfiguration(name="config",
                cli=["-c", "--config"],
                help='Specify configuration file (default: %(default)s). '
                     'Multiple --config options may be used.',,
                default="config.json",
                type=str,
                action="append"
                usages=["main", "backtesting", "edge-cli", "plot"],
                strategy_argument=False,
                )
FTConfiguration(name="stake_amount",
                cli=['--stake_amount'],
                help='Specify stake_amount.',,
                default="DefaultStrategy",
                type=float,
                action="append"
                usages=["main", "backtesting", "hyperopt"],
                strategy_argument=True,
                override_output='Parameter --stake_amount detected, '
                                'overriding stake_amount to: {} ...'
                )
```

Based on this, we could then dynamically generate arguments where they are needed, copy the arguments to the configuration, or load them from the configuration, and move them to the strategy eventually.

An alternative approach to this would be to use a .yaml (or json) configuration file specifying more or less the same things - but i think using it as class (or NamedTuple?) should work for our usecase.

This is just an idea for now, maybe it's terrible, but that's why i created this issue - to discuss.

Let me know.

**EDIT**:
for clarification: 
This would remove most argument-defining code in both `arguments.py` and `configuration.py` - since `parser.add_argument()` is only called once, and do the same to `_args_to_config()` in `configuration.py

hroff-1902:
Maybe. Need to think...

[...]

PR #1947, opened 2019-06-18
arguments cleanup

hroff-1902:
An attempt to improve arguments handling without big redesign,

* interfaces of methods unified -- they all made non-static back and call `parser.add_argument()` in a unified manner;

* names of the methods unified: `common_args_parser()` --> `common_options()`, `optimizer_shared_options()` --> `common_optimize_options()`, `scripts_options()` --> `common_scripts_options()`;

* `plot_dataframe.py` specific arguments moved to `arguments.py` (see `plot_dataframe_options()`);

When this script is ready to become a subcommand, that will be enough to add a subparser parameter to this method and change `parser = self.parser` to `parser = subparser or self.parser` in it.

* duplicated arguments in `download_data_options()` removed; 

The script (and then the subcommand) should call the `common_options()` method for common options instead of duplicating the code.

* `common_args_parser()` is split into `common_options()` and `main_options()` (options specific for the live running bot);

* calls to `parser.add_argument()` simplified: removed all appearences of `type=str` and `default=None` as these are default values;

* minor improvements and fixes in comments;

* tests adjusted;

### DIFF
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -493,16 +493,16 @@ class Arguments(object):
 
         parser.add_argument(
             '--indicators1',
-            help='Set indicators from your strategy you want in the first row of the graph. Separate '
-                 'them with a coma. E.g: ema3,ema5 (default: %(default)s)',
+            help='Set indicators from your strategy you want in the first row of the graph. '
+                 'Separate them with a coma. E.g: ema3,ema5 (default: %(default)s)',
             default='sma,ema3,ema5',
             dest='indicators1',
         )
 
         parser.add_argument(
             '--indicators2',
-            help='Set indicators from your strategy you want in the third row of the graph. Separate '
-                 'them with a coma. E.g: fastd,fastk (default: %(default)s)',
+            help='Set indicators from your strategy you want in the third row of the graph. '
+                 'Separate them with a coma. E.g: fastd,fastk (default: %(default)s)',
             default='macd,macdsignal',
             dest='indicators2',
         )
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -33,7 +33,8 @@ class Arguments(object):
         self.parser = argparse.ArgumentParser(description=description)
 
     def _load_args(self) -> None:
-        self.common_args_parser()
+        self.common_options()
+        self.main_options()
         self._build_subcommands()
 
     def get_parsed_arg(self) -> argparse.Namespace:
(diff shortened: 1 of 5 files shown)


xmatthias:
i guess this is an attempt to avoid #1945? if so, i would prefer a discussion before implementation ...

The problem this has and does not solve are multiple :
* if we need to add a new argument, i still need to go to multiple places (arguments.py, configuration.py). 
* It is not clear which arguments are applied where
* some options are applied wrongly (we add all optimize options to the plot_script - HOWEVER, not all work there (strategy-list for example) .

The idea from #1945 would remove all manual calls to `parser.add_argument` - as well as all (manual) calls to `_args_to_config`.

I did not start implementing this as i wanted to discuss it since it'll be a work that'll need a lot of testing, but maybe it's too difficult to understand without seeing the implementation details.

Also (and maybe this did not become clear in #1945) - we currently behave oddly with some parameters. if i add `-v` it should increase verbosity - HOWEVER, we currently have to add this **before** the backtesting command, otherwise we get `argument not known`.

Works;
```
freqtrade -v backtesting
```

Doesn't work
```
freqtrade backtesting -v 
```

It should not matter WHERE i add verbose - but the bot should accept it and do what it old it to do.

hroff-1902:
> i guess this is an attempt to avoid #1945?

No, not at all!

I agree with you completely about the argumentation about arguments here and in the #1945.

It's not an attempt to avoid #1945, it'a an attempt to make the arguments/configuration infrastructure more robust, clean, straight right now, without major refactoring with local changes and improvements. These are changes that appear during thinking on #1945 and reviewing current codebase for it, as well as during my current work on moving get_pairs and download_data scripts into modules (that can be done in parallel with #1945).

(Let's move the discussion to #1945, that seems to be more appropriate place for it and intended for it...)

freqtrade/arguments.py
```diff
@@ -184,11 +182,12 @@ def optimizer_shared_options(parser: argparse.ArgumentParser) -> None:
             dest='refresh_pairs',
         )
 
-    @staticmethod
-    def backtesting_options(parser: argparse.ArgumentParser) -> None:
+    def backtesting_options(self, subparser: argparse.ArgumentParser = None) -> None:
         """
         Parses given arguments for Backtesting module.
         """
+        parser = subparser or self.parser
```
xmatthias: an alternative way (for the future, maybe) could be to pass parser always to the function so we can save these comparisons (it's not like they are called 100ds of times per cycle ...).

PR #1955, opened 2019-06-21
Ticker interval to hyperopt

### LINKED ISSUE #1900: Hyperopt Resamples

xmatthias:
## Summary
Make ticker-interval available to hyperopt.
Since Hyperopt does mostly use static methods it needs to be assigned as a class attribute.

It's then available as `HyperoptName.ticker_interval`.

Closes #1900

### DIFF
--- a/docs/hyperopt.md
+++ b/docs/hyperopt.md
@@ -12,7 +12,7 @@ and still take a long time.
 ## Prepare Hyperopting
 
 Before we start digging into Hyperopt, we recommend you to take a look at
-an example hyperopt file located into [user_data/hyperopts/](https://github.com/freqtrade/freqtrade/blob/develop/user_data/hyperopts/test_hyperopt.py)
+an example hyperopt file located into [user_data/hyperopts/](https://github.com/freqtrade/freqtrade/blob/develop/user_data/hyperopts/sample_hyperopt.py)
 
 Configuring hyperopt is similar to writing your own strategy, and many tasks will be similar and a lot of code can be copied across from the strategy.
 
@@ -70,6 +70,11 @@ Place the corresponding settings into the following methods
 
 The configuration and rules are the same than for buy signals.
 To avoid naming collisions in the search-space, please prefix all sell-spaces with `sell-`.
+
+#### Using ticker-interval as part of the Strategy
+
+The Strategy exposes the ticker-interval as `self.ticker_interval`. The same value is available as class-attribute `HyperoptName.ticker_interval`.
+In the case of the linked sample-value this would be `SampleHyperOpts.ticker_interval`.
 
 ## Solving a Mystery
 
--- a/freqtrade/optimize/hyperopt_interface.py
+++ b/freqtrade/optimize/hyperopt_interface.py
@@ -20,6 +20,7 @@ class IHyperOpt(ABC):
         stoploss -> float: optimal stoploss designed for the strategy
         ticker_interval -> int: value of the ticker interval to use for the strategy
     """
+    ticker_interval: str
 
     @staticmethod
     @abstractmethod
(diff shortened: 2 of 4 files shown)


docs/hyperopt.md
```diff
@@ -71,6 +71,11 @@ Place the corresponding settings into the following methods
 The configuration and rules are the same than for buy signals.
 To avoid naming collisions in the search-space, please prefix all sell-spaces with `sell-`.
 
+#### Using ticker-interval as part of the Strategy
+
+The Strategy exposes the ticker-interval as `self.ticker_interval`. The same value is available as class-attribute `HyperoptName.ticker_interval`.
```
hroff-1902: in the hyperopts. this should also be available as self.ticker_interval, bacause class-attributes are available via `self` as well. (IIRC)

xmatthias: Adding it as Class-method makes it available for instances of that class too - so if you have an instance method it's available as `self.ticker_interval` too.

tested using a custom hyperopt file as follows (don't look at anything other than the print statements, it may not make sense).
*Note:* `GreatHyperopts` is the name of my class.

``` python
        print(f"from class: {GreatHyperopts.ticker_interval}")
        print(f"from self: {self.ticker_interval}")
        dataframe.loc[
            (
                (qtpylib.crossed_above(
                    dataframe['macdsignal'], dataframe['macd']
                )) &
                (dataframe['fastd'] > 54)
            ),
            'sell'] = 1
        return dataframe

```

hroff-1902: So why should we mention that ticker_interval is available as HyperoptName.ticker_interval in the docs while it's simply available as self.ticker_interval in both custom strategies AND custom hyperopts?

I mean, the description for the end-user can be simplified here: "The Strateges and the Hyperopts expose the ticker-interval as `self.ticker_interval`." or something like that...

hroff-1902:
The rest LGTM, cool! 👍

docs/hyperopt.md
xmatthias: because all of the methods which are subject to hyperopting are static methods.

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L31-L32

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L49-L54

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L91-L92

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L108-L113

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L152

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L170-L171

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L183-L184

https://github.com/freqtrade/freqtrade/blob/a75f08cf17cd039aaae1e8b980482c63c0b94266/user_data/hyperopts/sample_hyperopt.py#L192-L193


The only two methods, where `self.ticker_interval` is possible within the whole custom hyperopt class are the fallback methods for buy and sell signal (when not using `--spaces all` or `--spaces buy,sell`.

PR #1959, opened 2019-06-22
Split btanalysis load trades

xmatthias:
## Summary
Fix issue when plotting that trades are not loaded from file since config always has db_url.

## Quick changelog

- Document on how to load data for analysis (either from DB or from Backtest results)
- add argument  `--trade-source` to be explicit from where the trades are loaded (this has been an issue in the past)
- remove now obsolete method combination method to load trades from file or db

### DIFF
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -22,7 +22,7 @@ df.groupby("pair")["sell_reason"].value_counts()
 
 ```
 
-This will allow you to drill deeper into your backtest results, and perform analysis which would make the regular backtest-output very difficult to digest due to information overload.
+This will allow you to drill deeper into your backtest results, and perform analysis which otherwise would make the regular backtest-output very difficult to digest due to information overload.
 
 If you have some ideas for interesting / helpful backtest data analysis ideas, please submit a Pull Request so the community can benefit from it.
 
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -1,0 +1,44 @@
+# Analyzing bot data
+
+After performing backtests, or after running the bot for some time, it will be interresting to analyze the results your bot generated.
+
+A good way for this is using Jupyter (notebook or lab) - which provides an interactive environment to analyze the data.
+
+The following helpers will help you loading the data into Pandas DataFrames, and may also give you some starting points in analyzing the results.
+
+## Backtesting
+
+To analyze your backtest results, you can [export the trades](#exporting-trades-to-file).
+You can then load the trades to perform further analysis.
+
+A good way for this is using Jupyter (notebook or lab) - which provides an interactive environment to analyze the data.
+
+Freqtrade provides an easy to load the backtest results, which is `load_backtest_data` - and takes a path to the backtest-results file.
+
+``` python
+from freqtrade.data.btanalysis import load_backtest_data
+df = load_backtest_data("user_data/backtest-result.json")
+
+# Show value-counts per pair
+df.groupby("pair")["sell_reason"].value_counts()
+
+```
+
+This will allow you to drill deeper into your backtest results, and perform analysis which would make the regular backtest-output unreadable.
+
+If you have some ideas for interesting / helpful backtest data analysis ideas, please submit a PR so the community can benefit from it.
+
+## Live data
+
+To analyze the trades your bot generated, you can load them to a DataFrame as follwos:
+
+``` python
+from freqtrade.data.btanalysis import load_trades_from_db
+
+df = load_trades_from_db("sqlite:///tradesv3.sqlite")
+
+df.groupby("pair")["sell_reason"].value_counts()
+
+```
+
+Feel free to submit an issue or Pull Request if you would like to share ideas on how to best analyze the data.
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -1,6 +1,6 @@
 # Analyzing bot data
 
-After performing backtests, or after running the bot for some time, it will be interresting to analyze the results your bot generated.
+After performing backtests, or after running the bot for some time, it will be interesting to analyze the results your bot generated.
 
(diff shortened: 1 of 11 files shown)


docs/data-analysis.md
```diff
@@ -0,0 +1,44 @@
+# Analyzing bot data
+
+After performing backtests, or after running the bot for some time, it will be interresting to analyze the results your bot generated.
```
hroff-1902: interesting

docs/data-analysis.md
```diff
@@ -0,0 +1,44 @@
...
+A good way for this is using Jupyter (notebook or lab) - which provides an interactive environment to analyze the data.
+
+The following helpers will help you loading the data into Pandas DataFrames, and may also give you some starting points in analyzing the results.
+
+## Backtesting
+
+To analyze your backtest results, you can [export the trades](#exporting-trades-to-file).
+You can then load the trades to perform further analysis.
+
+A good way for this is using Jupyter (notebook or lab) - which provides an interactive environment to analyze the data.
```
hroff-1902: repeated sentence just a few lines above

docs/data-analysis.md
```diff
@@ -0,0 +1,44 @@
...
+
+Freqtrade provides an easy to load the backtest results, which is `load_backtest_data` - and takes a path to the backtest-results file.
```
hroff-1902: Freqtrade provides helper function for easy load... which takes the path to the...

Or something like that...

docs/data-analysis.md
```diff
@@ -0,0 +1,44 @@
...
+``` python
+from freqtrade.data.btanalysis import load_backtest_data
+df = load_backtest_data("user_data/backtest-result.json")
+
+# Show value-counts per pair
+df.groupby("pair")["sell_reason"].value_counts()
+
+```
+
+This will allow you to drill deeper into your backtest results, and perform analysis which would make the regular backtest-output unreadable.
```
hroff-1902: "analysis which would make the regular backtest-output unreadable." -- this is incorrect sentence. Analisys would not make output unreadable.

docs/data-analysis.md
```diff
@@ -0,0 +1,44 @@
...
+
+If you have some ideas for interesting / helpful backtest data analysis ideas, please submit a PR so the community can benefit from it.
+
+## Live data
+
+To analyze the trades your bot generated, you can load them to a DataFrame as follwos:
```
hroff-1902: follows

docs/data-analysis.md
```diff
@@ -0,0 +1,44 @@
...
+df = load_backtest_data("user_data/backtest-result.json")
+
+# Show value-counts per pair
+df.groupby("pair")["sell_reason"].value_counts()
+
+```
+
+This will allow you to drill deeper into your backtest results, and perform analysis which would make the regular backtest-output unreadable.
+
+If you have some ideas for interesting / helpful backtest data analysis ideas, please submit a PR so the community can benefit from it.
```
hroff-1902: PR --> Pull Request, as below

docs/data-analysis.md
```diff
@@ -0,0 +1,44 @@
...
+``` python
+from freqtrade.data.btanalysis import load_trades_from_db
+
+df = load_trades_from_db("sqlite:///tradesv3.sqlite")
+
+df.groupby("pair")["sell_reason"].value_counts()
+
+```
+
+Feel free to submit an issue or Pull Request if you would like to share ideas on how to best analyze the data.
```
hroff-1902: also partly repeated above ("please submit...")

scripts/plot_dataframe.py
```diff
@@ -124,10 +124,12 @@ def analyse_and_plot_pairs(config: Dict[str, Any]):
         tickers = {}
         tickers[pair] = data
         dataframe = generate_dataframe(strategy, tickers, pair)
+        trades = None
+        if config["trade_source"] == "DB":
+            trades = load_trades_from_db(config["db_url"])
+        elif config["trade_source"] == "file":
+            trades = load_backtest_data(Path(config["exportfilename"]))
```
hroff-1902: I'm not sure this should be a switch with the type of the source. Is not it possible to use --exportfilename and --db-url if any is specified and make those options incompatible instead so that only one can be specified in the command line?

I don't like this new option, seems to be redundant... 👎

docs/data-analysis.md
xmatthias: I know it's partially repeated, my assumption is that users will read the part relevant to them (either Backtest data, OR live data) - depending on the progress they made.

scripts/plot_dataframe.py
xmatthias: Technically i agree with you - but there is no other way at the moment since db_url is ALWAYS in the configuration (and the same happens with exportfilename).

We need it as a default - so how do i determine which of the options was selected intentionally?
(it's also not really about defaults - see below)

It's also valid to specify `db_url` and `exportfilename` in the configuration (one is used during backtesting, the other during live/dry trading, so they don't collide apart from plotting).
Plotting however supports only one or the other (so it's not about cli arguments, really).

Honestly, this part was always broken (the working part switched back and forth, with each try to fix it).

We may remove this option in the future if we find a better solution - but it's not redundant, sorry.

docs/data-analysis.md
xmatthias: you're right - it would make it difficult to digest since it'll be confusing with too many columns/tables.

docs/data-analysis.md
hroff-1902: okay

scripts/plot_dataframe.py
hroff-1902: okay, let's leave it for now...

docs/data-analysis.md
hroff-1902: "which __otherwise__ would make the regular backtest-output very difficult to digest due to information overload."

PR #1957, opened 2019-06-22
New release - 2019.6

xmatthias:
## Summary
Creating a new release called 2019.6 (we have year 2019 - month 6).

I'd like to get away from "random" version numbers to a calendar based versioning scheme - so it's clearly visible how old a release is (this should also encourage us to do more frequent releases, which is something we're currently not very good at).

## Changelog:

```
e83f8941 Fix documentation grammar
d6dbb21a Update mypy from 0.701 to 0.710
90ada064 Update wrapt from 1.11.1 to 1.11.2
e8429bd2 Update sqlalchemy from 1.3.4 to 1.3.5
5a30f046 Update ccxt from 1.18.725 to 1.18.805
11d39bb0 Improve wording
116d8e85 typos in docstrings fixed
5b84cb39 typo fixed
7f018839 diverse cosmetics to options help strings
3716c04e fix help string for --db-url
451d4a40 fix help strings shown to the user
4cbcb5f3 Move .title to ExchangeResolver (it does not make sense to do this over and over again)
026784ef remove get_tickers_data from plot_dataframe
cc56d0e0 Remove unneeded initialization
559d5ebd Remove combined load-method since it's confusing
3e61ada3 Be explicit in what is used, db or trades
8758218b Add data-analysis documentation
de38aea1 Fix sequence of loading trades
db17b20e Don't require pairs but fall back to pair_whitelist instead
a581ca66 Adapt test after merging develop
7a0d8666 Mypy type errors
1a27ae8a Add tests to verify that ticker_interval is there
f907a487 make ticker_interval available to hyperopt functions
89ba649d Test handling errors while trailing stop loss
63640518 Gracefully handle errosr when cancelling stoploss orders
a8dcfc05 Add test to verify InvalidOrder is handled correctly
dd379c41 Cancelling stoploss order should not kill the bot
911e71cd remove redundant test-functions
0866b5f2 allow reading config from stdin
860e0563 --datadir is now handled in arguments.common_options()
c6fed4e4 make flake happy
8c40a406 arguments cleanup
6f950bbd json validator cosmetics
0e7ea1da Update coveralls from 1.8.0 to 1.8.1
6973087d Update pytest from 4.6.2 to 4.6.3
25755f6a Update ccxt from 1.18.667 to 1.18.725
ba4890d3 Fix tests on windows
7cd36239 UPdate documentation with new value
06afb3f1 Don't use "limit" for sell-orders either
55712292 Add order_type to sell-notification
475e76b2 Add order_type to buy_notification
[...]

PR #1961, opened 2019-06-22
Argument handling refactor

xmatthias:
## Summary
This should be a first step to get #1945. This is really necessary as the plot-scripts show 100ds of options, while only a fraction of them really works (e.g. Plot dataframe shows the whole backtesting arguments, but only --live and --refresh-pairs-cached are actually used)

Part of the issue: #1945

## Quick changelog

- Use dict of Arguments to "pick and choose" using lists.
- Each argument is defined exactly once, and can be used at will by every subcommand.
- Kept only options that really work for plot-scripts

Technically, the PR is ready for review, however it should be merged AFTER #1957 to give us time to do further steps in #1945 

Edit: Rebased against develop to make sure none of the changes to helpstrings are missed.


hroff-1902:
Fine. I thought it would be much heavier solution...

I have a couple of design questions we have to address here (or in a sep. PR), will post them a bit later in #1945...

xmatthias:
Well it's a first step in #1945 - not the "end-goal" - but it allows "simple" reuse of arguments across different commands (it's really a requirement to "fix" the plot-scripts without creating 20 new "argument_*" functions).

hroff-1902:
> without creating 20 new "argument_*" functions).

They (arguments) can be grouped however as I've done for load_common/load_main...

xmatthias:
partly yes - but mainly the plot scripts use 2 or 3 from backtesting - but not the rest.
also i see this as a first step to move this out of arguments.py and into config - with default (which is not a default in the argument) ...

but everything at once would be even tougher to review

PR #1962, opened 2019-06-22
minor: fix help strings shown to the user

hroff-1902:
* Help strings for the --config, --db-url (main freqtrade), --exchange, --timeframes (download script) errorneously say the default value is None; this is fixed.
* Minor cosmetics for the strings in arguments.py: format help strings in the same manner.

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -4,6 +4,7 @@
 bot constants
 """
 DEFAULT_CONFIG = 'config.json'
+DEFAULT_EXCHANGE = 'bittrex'
 DYNAMIC_WHITELIST = 20  # pairs
 PROCESS_THROTTLE_SECS = 5  # sec
 DEFAULT_TICKER_INTERVAL = 5  # min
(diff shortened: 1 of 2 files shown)


freqtrade/constants.py
```diff
@@ -21,6 +22,7 @@
 ORDERTIF_POSSIBILITIES = ['gtc', 'fok', 'ioc']
 AVAILABLE_PAIRLISTS = ['StaticPairList', 'VolumePairList']
 DRY_RUN_WALLET = 999.9
+DEFAULT_DOWNLOAD_TICKER_INTERVALS = '1m 5m'
```
xmatthias: isn't the correct syntax 1m,5m for the download script ... ?

hroff-1902: No, it says "Space separated list."
```
./scripts/download_backtest_data.py --exchange binance -t 1h 1d
```

xmatthias: Good point :+1:

hroff-1902:
@xmatthias 
Ah, the docs should also be fixed. I have not thought that errorneous None values had time to go to the docs since we broke it in the help strings...

(this does not block the new release since docs can be fixed later)

PR #1967, opened 2019-06-24
Modify handling of pip in setup.sh

### LINKED ISSUE #1956: Installation failed due to pip dependcy missing in setup.sh
My environment
Ubuntu 18.04.2 LTS
Python 3.6
ccxt 1.18.726
Develop a75f08cf17cd039aaae1e8b980482c63c0b94266
  
The problem:

On a fresh Ubuntu 18.04.2 LTS system, pip3 was not installed, the installation failed as the setup.sh install pip packages.

To reproduce:
Reproduce in a Ubuntu docker in the above mentioned Ubuntu system.
Create the docker:
`chrichri@chrichri-KPL-W0X:~/freqtrade$ sudo docker run -ti ubuntu /bin/bash`
In the docker if installing the packages from setup.sh

```
root@620b800e245d:/# apt
apt 1.6.11 (amd64)

...

root@620b800e245d:/# apt-get update

...

Fetched 16.8 MB in 23s (741 kB/s)                                                                                                                                                                          
Reading package lists... Done
root@620b800e245d:/# apt-get install python3.6 python3.6-venv python3.6-dev build-essential autoconf libtool pkg-config make wget git
Reading package lists... Done

...

done.
```

No pip related command:
```

root@620b800e245d:/# pip
bash: pip: command not found
root@620b800e245d:/# pip3
bash: pip3: command not found
root@620b800e245d:/# python -m pip 
bash: python: command not found
root@620b800e245d:/# python3 -m pip
bash: python3: command not found
```

This cause the installation to fail, to fix the issue I add python3-pip dependency to setup.sh. Confirmed in docker:

  ```
root@620b800e245d:/# apt install python3-pip
Reading package lists... Done

root@620b800e245d:/# pip3

Usage:   
  pip <command> [options]
  ```

xmatthias:
## Summary
The install script does not work correctly. 
It checks which python version to use before eventually installing that. 
Also, there are many ways (and even more different systems) to install python to a system, many better than the one we use in the script (pyenv to name one).

We should not install python for users, but instead require python and pip to be there.
We also do some "funny" things to brew on macOS. 
I would appreciate if someone with a Mac could look over that part too, since i think depending on the OS version we may break things.

Resolves #1956

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -4,7 +4,7 @@
 function check_installed_pip() {
    ${PYTHON} -m pip > /dev/null
    if [ $? -ne 0 ]; then
-        echo "pip not found. Please make sure that pip is available for ${PYTHON}."
+        echo "pip not found (called as '${PYTHON} -m pip'). Please make sure that pip is available for ${PYTHON}."
         exit 1
    fi
 }
--- a/setup.sh
+++ b/setup.sh
@@ -1,12 +1,21 @@
 #!/usr/bin/env bash
 #encoding=utf8
 
+function check_installed_pip() {
+   ${PYTHON} -m pip > /dev/null
+   if [ $? -ne 0 ]; then
+        echo "pip not found. Please make sure that pip is available for ${PYTHON}."
+        exit 1
+   fi
+}
+
 # Check which python version is installed
 function check_installed_python() {
     which python3.7
     if [ $? -eq 0 ]; then
         echo "using Python 3.7"
         PYTHON=python3.7
+        check_installed_pip
         return
     fi
 
@@ -14,6 +23,7 @@ function check_installed_python() {
     if [ $? -eq 0 ]; then
         echo "using Python 3.6"
         PYTHON=python3.6
+        check_installed_pip
         return
    fi
 
@@ -21,7 +31,6 @@ function check_installed_python() {
         echo "No usable python found. Please make sure to have python3.6 or python3.7 installed"
         exit 1
    fi
-
 }
 
 function updateenv() {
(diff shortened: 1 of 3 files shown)


setup.sh
```diff
@@ -1,49 +1,58 @@
 #!/usr/bin/env bash
 #encoding=utf8
 
+function check_installed_pip() {
+   ${PYTHON} -m pip > /dev/null
+   if [ $? -ne 0 ]; then
+        echo "pip not found. Please make sure that pip is available for ${PYTHON}."
```
hroff-1902: If we are more specific here, this may help the user installing the bot identify the problem and add a dependency required:
"pip not found (called as '${PYTHON} -m pip')."

hroff-1902:
So python3 should be stated as a prerequisite (system requirements) in the docs.

xmatthias:
> So python3 should be stated as a prerequisite (system requirements) in the docs.

in theory, this was already in the readme.md and index.md - but i've now also added it to installation.md for completeness

hroff-1902:
Well, in the installation.md, if a user reads it and chooses the easiest way, she see:
* "Prerequisites" -- no python mentioned.
* then "Easy Installation - Linux Script".
* and only below, in the "Custom Installation / Requirements" she can see something about python. But this belongs to another installation approach.

Part of it should be moved above, in the common "Prerequisites".

Btw, when I first installed freqtrade (on Ubuntu) following the docs, I was not successful. The only way worked for me at my system was Docker...

Other points, to discuss, not for this PR:
* let's split installation.md into system-specific parts (plus sep. files for docker, install. script, common part/prerequisites). So that user won't jump between the parts of the long document,,, They can be shown in the single section of docs, like "Controlling the bot" now...

xmatthias:
well docker is a seperate file already (and python is not relevant in that file i think).

I would however not split the document now as too much fragmentation of the install instructions makes it overly complicated / duplicates a lot (mac/linux is very similar).

I've now tried to fix the doc up a bit - the docs could generally be easier worded - but that's another project i would think

PR #1969, opened 2019-06-24
[minor] Improve developer-document

xmatthias:
to include a note to keep both branches uptodate while creating a changelog.

Cost me ~5 minutes doing the 2019.6 release, since i had ~twice as many changelog entries than commits ...

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -140,7 +140,7 @@ git checkout develop
 git checkout -b new_release
 ```
 
-* Edit `freqtrade/__init__.py` and add the desired version (for example `0.18.0`)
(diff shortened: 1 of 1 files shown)


docs/developer.md
```diff
@@ -140,7 +140,7 @@ git checkout develop
 git checkout -b new_release
 ```
 
-* Edit `freqtrade/__init__.py` and add the desired version (for example `0.18.0`)
+* Edit `freqtrade/__init__.py` and add the version matching the current date (for example `2019.7` for June 2019). Minor versions can be `2019.7-1` should we need to do a second release that month.
```
hroff-1902: 2019.7 is July

xmatthias: no comment ... :laughing:

PR #1980, opened 2019-06-27
Update FAQ with a recipe on PYTHONPATH

hroff-1902:
Think this will reduce the number of similar questions from the <s>noobs</s>novice users...

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -2,6 +2,27 @@
 
 ### Freqtrade commons
 
+#### I'm trying to run the bot, but it says "No module named 'freqtrade'"
+
+When you run the bot on a Linux system and obtain something similar to the 
+following error:
+``` bash
+$ python3 freqtrade -c config.json
+Traceback (most recent call last):
+ File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
+   "__main__", mod_spec)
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -2,6 +2,27 @@
...
+   "__main__", mod_spec)
+ File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
+   exec(code, run_globals)
+ File "freqtrade/__main__.py", line 9, in <module>
+   from freqtrade import main
+ModuleNotFoundError: No module named 'freqtrade'
+```
+try to setup properly the PYTHONPATH environment variable. So simply run
+``` bash
+export PYTHONPATH=.:$PYTHONPATH
```
xmatthias: i'm currently buzzled as i thought i had changed all documentation to to run the bot using `freqtrade -c config.json` (without python3 at the front).

I think that would be the easier solution (it's not path-dependant, but requires a correct installation, which is required anyway).

The error you mention (or a similar one)  can also happen if it's not installed using `pip install -e .` (or in similar ways).

hroff-1902:
I'm going to close this in favor to a more generic issue in the Issues.

hroff-1902:
Closing in favor to #1990

PR #1988, opened 2019-06-30
Gracefully fail on timeframes exception

### LINKED ISSUE #1983: Bitstamp has no attribute 'timeframes'
Python Version: Python 3.7.3
    CCXT version: ccxt==1.18.805
    Branch: Develop
    Last Commit ID: ca7080c
  
```
  File "/opt/Projects/freqtrade/freqtrade/main.py", line 51, in main
    worker = Worker(args)
  File "/opt/Projects/freqtrade/freqtrade/worker.py", line 35, in __init__
    self._init(False)
  File "/opt/Projects/freqtrade/freqtrade/worker.py", line 51, in _init
    self.freqtrade = FreqtradeBot(self._config)
  File "/opt/Projects/freqtrade/freqtrade/freqtradebot.py", line 56, in __init__
    self.exchange = ExchangeResolver(self.config['exchange']['name'], self.config).exchange
  File "/opt/Projects/freqtrade/freqtrade/resolvers/exchange_resolver.py", line 31, in __init__
    self.exchange = Exchange(config)
  File "/opt/Projects/freqtrade/freqtrade/exchange/exchange.py", line 140, in __init__
    self.validate_timeframes(config['ticker_interval'])
  File "/opt/Projects/freqtrade/freqtrade/exchange/exchange.py", line 272, in validate_timeframes
    timeframes = self._api.timeframes
AttributeError: 'bitstamp' object has no attribute 'timeframes'
```

### Steps to reproduce:

  1. Attempt to use bitstamp
  
### Observed Results:

  * Exchange initialization failed
  * Exchange initialization succeed

### What to do ?
Is there a way to fix this ?
If not, how about a way to have it fail in a little bit nicer fashion ?

xmatthias:
## Summary
This can happen on exchanges which don't have a "real" fetch_OHLCV method.

The example (bitstamp) has this emulated.
From the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html?highlight=timeframes)

> timeframes: An associative array of timeframes, supported by the fetchOHLCV method of the exchange. This is only populated when has['fetchOHLCV'] property is true.

Closes #1983

## Quick changelog

- Check if element is available, fail with explaining the error otherwise.
- Add test for this specific case

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -269,7 +269,6 @@ class Exchange(object):
         """
         Checks if ticker interval from config is a supported timeframe on the exchange
         """
-        logger.warning("validate_timerames")
         if not hasattr(self._api, "timeframes"):
             # If timeframes is missing, the exchange probably has no fetchOHLCV method.
             # Therefore we also show that.
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -269,6 +269,13 @@ class Exchange(object):
         """
         Checks if ticker interval from config is a supported timeframe on the exchange
         """
+        logger.warning("validate_timerames")
+        if not hasattr(self._api, "timeframes"):
+            # If timeframes is missing, the exchange probably has no fetchOHLCV method.
+            # Therefore we also show that.
+            raise OperationalException(
+                f"This exchange ({self.name}) does not have a `timeframes` attribute and "
+                f"is therefore not supported. fetchOHLCV: {self.exchange_has('fetchOHLCV')}")
         timeframes = self._api.timeframes
         if timeframe not in timeframes:
             raise OperationalException(
--- a/freqtrade/resolvers/exchange_resolver.py
+++ b/freqtrade/resolvers/exchange_resolver.py
@@ -28,6 +28,7 @@ class ExchangeResolver(IResolver):
         except ImportError:
             logger.info(
                 f"No {exchange_name} specific subclass found. Using the generic class instead.")
+        if not hasattr(self, "exchange"):
             self.exchange = Exchange(config)
 
     def _load_exchange(
--- a/freqtrade/tests/exchange/test_exchange.py
+++ b/freqtrade/tests/exchange/test_exchange.py
@@ -396,6 +396,23 @@ def test_validate_timeframes_failed(default_conf, mocker):
         Exchange(default_conf)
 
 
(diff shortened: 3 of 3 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -269,6 +269,13 @@ def validate_timeframes(self, timeframe: List[str]) -> None:
         """
         Checks if ticker interval from config is a supported timeframe on the exchange
         """
+        logger.warning("validate_timerames")
```
hroff-1902: should be something like
```
logger.info("Validating timeframes...")
```

logger.warning() is not proper here since it's printed before checking for attr an generating the exception. A warning or even a critical message may additionally be issued below, under the `if not hasattr()` part.

freqtrade/resolvers/exchange_resolver.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, exchange_name: str, config: dict) -> None:
         except ImportError:
             logger.info(
                 f"No {exchange_name} specific subclass found. Using the generic class instead.")
+        if not hasattr(self, "exchange"):
```
hroff-1902: in other places in the codebase another kind of checking (`if not edge`) is used... I propose to unify these checks.

xmatthias: that does not work unfortunately (and the case is quite different here).

`self.exchange` does not yet exist in case of the import-error - and the only way to not have that fail is using hasattr, since we'd be checking a property which does not yet exist.

Initializing exchange to "None" would allow to check `if self.exchange` - but has the negative side-effect that the type changes from `Exchange` to `Optional[Exchange]` - which makes usage of this property a mess, since it will always exist after the __init__ function.

Initializing exchange within the exception handler does work (and prevents having to check) - but raises a very ugly exception (stacktrace) since it's raised within the exception handler so python thinks the original error is the cause.

`hasattr` is fine to check (but has performance implications, so should only be used when we know how often it's called (once - in this case).
Edge on the other hand is something optional - we're checking that often during regular operations.

freqtrade/exchange/exchange.py
xmatthias: This ... was never intendet to be commited - forgot to remove it.

freqtrade/resolvers/exchange_resolver.py
hroff-1902: You are right.

LGTM 👍

PR #1993, opened 2019-07-02
Remove duplicate keyword from arguments

xmatthias:
## Summary
We don't need the "dest" kwargument, it's already the key of the dict.

part of  #1945 

## Quick changelog

- Remove duplicate entry

### DIFF
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -36,13 +36,11 @@ AVAILABLE_CLI_OPTIONS = {
         '-v', '--verbose',
         help='Verbose mode (-vv for more, -vvv to get all messages).',
         action='count',
-        dest='loglevel',
         default=0,
     ),
     "logfile": Arg(
         '--logfile',
         help='Log to the file specified.',
-        dest='logfile',
         metavar='FILE',
     ),
 
@@ -56,26 +54,22 @@ AVAILABLE_CLI_OPTIONS = {
         help=f'Specify configuration file (default: `{constants.DEFAULT_CONFIG}`). '
         f'Multiple --config options may be used. '
         f'Can be set to `-` to read config from stdin.',
-        dest='config',
         action='append',
         metavar='PATH',),
     "datadir": Arg(
         '-d', '--datadir',
         help='Path to backtest data.',
-        dest='datadir',
         metavar='PATH',),
     # Main options
     "strategy": Arg(
         '-s', '--strategy',
         help='Specify strategy class name (default: `%(default)s`).',
-        dest='strategy',
         default='DefaultStrategy',
         metavar='NAME',
     ),
     "strategy_path": Arg(
         '--strategy-path',
         help='Specify additional strategy lookup path.',
-        dest='strategy_path',
         metavar='PATH',
     ),
     "dynamic_whitelist": Arg(
@@ -83,7 +77,6 @@ AVAILABLE_CLI_OPTIONS = {
         help='Dynamically generate and update whitelist '
         'based on 24h BaseVolume (default: %(const)s). '
         'DEPRECATED.',
-        dest='dynamic_whitelist',
         const=constants.DYNAMIC_WHITELIST,
         type=int,
         metavar='INT',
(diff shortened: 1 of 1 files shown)


xmatthias:
i guess it is (by duplicating the line `parser.add_argument()` and adding the same command without dest keyword).

However i don't see the point to be honest (maybe explain where you'd see a usecase for this) - the dict-key is the dest argument (and it's at the same time the key in the configuration).

Assuming that were not the case, it would be quite (very) confusing:

I add a argument, select it for plot-scripts called `indicators3` ... but i want to have it as `supercoolargument` in the configuration? 
i can also select `supercoolargument` then (only developers know how this is called) - which makes selection and the final key in the configuration/argument identical.

hroff-1902:
> i guess it is (by duplicating the line `parser.add_argument()` and adding the same command without dest keyword).
> 
> However i don't see the point to be honest (maybe explain where you'd see a usecase for this) - the dict-key is the dest argument (and it's at the same time the key in the configuration).
> 
> Assuming that were not the case, it would be quite (very) confusing:
> 
> I add a argument, select it for plot-scripts called `indicators3` ... but i want to have it as `supercoolargument` in the configuration?
> i can also select `supercoolargument` then (only developers know how this is called) - which makes selection and the final key in the configuration/argument identical.

Well, assume we have a cli option, say, --print-format, in a few subcommands with diff. definition. For the list-exchanges subcommand it will be --print-format=onecolumn/tabular/list while for the list-pairs subcommand it could be --print-format=tabular/list/json/csv, with two diff. help strings and/or diff. default values. Add a requirement to have it same setting in the configuration. This would require diff. entries in the AVAILABLE_CLI_OPTIONS dict, with diff dict keys but same dest value.

xmatthias:
isn't that overly complicated (and it'll fail on another step anyway - see below)?

assume `print_format_le` supports `onecolumn/tabular/list` - while `print_format_list_pairs` has `tabular/list/json/csv` (as you mentioned above).

you'd select it with `print_format_le`  for `list-exchanges`, 
and `print_format_list_pairs` for `list-pairs` ... 
why can't the result (in args / config ) also be `print_format_le` / `print_format_list_pairs`?

You won't have json/csv implemented for list_exchanges anyway - so using the same option will prevent you from having this option in the configuration.

assume i have `print_format` set to json in the configuration, because i use this command often and don't want to specify it every time (which is supported as we move everything to config). This would break the list-exchange option - since this option does not support `json`. 
Yes, i can override it via argument (but get an exception first since i don't remember when using list-exchanges with the unsupported format again) - but by calling the option `print_format_le` and `print_format_list_pairs` (or whatever makes sense) i can specify both separately in the configuration.

hroff-1902:
> isn't that overly complicated (and it'll fail on another step anyway - see below)?
> 
> assume `print_format_le` supports `onecolumn/tabular/list` - while `print_format_list_pairs` has `tabular/list/json/csv` (as you mentioned above).
> 
> you'd select it with `print_format_le` for `list-exchanges`,
> and `print_format_list_pairs` for `list-pairs` ...
> why can't the result (in args / config ) also be `print_format_le` / `print_format_list_pairs`?
> 
> You won't have json/csv implemented for list_exchanges anyway - so using the same option will prevent you from having this option in the configuration.
> 
> assume i have `print_format` set to json in the configuration, because i use this command often and don't want to specify it every time (which is supported as we move everything to config). This would break the list-exchange option - since this option does not support `json`.
> Yes, i can override it via argument (but get an exception first since i don't remember when using list-exchanges with the unsupported format again) - but by calling the option `print_format_le` and `print_format_list_pairs` (or whatever makes sense) i can specify both separately in the configuration.

Well, I agree for configuration -- this will probably not pass the json validation... My idea was to have it same configuration setting to avoid (pseudocode)
```
if mode==list-pairs: 
    v = config.get('print_format_list_pairs')
elif mode==list-exchanges:
    v = config.get('print_format_list_exchanges')
```
in the common code that is executed in both subcommands...

Okay, may be this is, right, overcomplicated, let's merge this for a while and return to it in case an actual case for it appears in the future...

xmatthias:
If it's common code, we should support the same formats (even if one may not be used often ...).
Also you can pass in the format to use to the "common function" doing the storing / export ...

hroff-1902:
> If it's common code, we should support the same formats (even if one may not be used often ...).
> Also you can pass in the format to use to the "common function" doing the storing / export ...

Not necessarily, I guess, but, anyway, let's merge this PR and move further... 👍

Issue #1992, opened 2019-07-02
Extend the list of 'bad' exchanges


hroff-1902:
...in addition to #1984, with those of them that also do not return the `timeframes` attribute from ccxt.

The list is the following:
``` bash
$ for i in `./freqtrade/main.py list-exchanges -1`; do ./freqtrade/main.py list-markets --exchange $i --print-list | grep "object has no attribute 'timeframes'"; done
AttributeError: 'anxpro' object has no attribute 'timeframes'
AttributeError: 'anybits' object has no attribute 'timeframes'
AttributeError: 'bcex' object has no attribute 'timeframes'
AttributeError: 'bigone' object has no attribute 'timeframes'
AttributeError: 'bit2c' object has no attribute 'timeframes'
AttributeError: 'bitbay' object has no attribute 'timeframes'
AttributeError: 'bitflyer' object has no attribute 'timeframes'
AttributeError: 'bitforex' object has no attribute 'timeframes'
AttributeError: 'bithumb' object has no attribute 'timeframes'
AttributeError: 'bitsane' object has no attribute 'timeframes'
AttributeError: 'bitso' object has no attribute 'timeframes'
AttributeError: 'bitstamp' object has no attribute 'timeframes'
AttributeError: 'bitstamp1' object has no attribute 'timeframes'
AttributeError: 'bl3p' object has no attribute 'timeframes'
AttributeError: 'braziliex' object has no attribute 'timeframes'
AttributeError: 'btcbox' object has no attribute 'timeframes'
AttributeError: 'btcchina' object has no attribute 'timeframes'
AttributeError: 'btctradeim' object has no attribute 'timeframes'
AttributeError: 'btctradeua' object has no attribute 'timeframes'
AttributeError: 'bxinth' object has no attribute 'timeframes'
AttributeError: 'chilebit' object has no attribute 'timeframes'
AttributeError: 'coinbase' object has no attribute 'timeframes'
AttributeError: 'coincheck' object has no attribute 'timeframes'
AttributeError: 'coinegg' object has no attribute 'timeframes'
AttributeError: 'coinexchange' object has no attribute 'timeframes'
AttributeError: 'coinfalcon' object has no attribute 'timeframes'
AttributeError: 'coinfloor' object has no attribute 'timeframes'
AttributeError: 'coingi' object has no attribute 'timeframes'
AttributeError: 'coinmarketcap' object has no attribute 'timeframes'
[...]

PR #1994, opened 2019-07-02
fix validate_timeframes()

hroff-1902:
* This PR addresses the issue mentioned in #1992 (see https://github.com/freqtrade/freqtrade/issues/1992#issuecomment-507825182): `validate_timeframes()` does not work as intended (it does not issue `OperationalException`) after https://github.com/ccxt/ccxt/commit/2be61d6b80a7e7e9be44ba5fe0f5bcc54997584d made in ccxt.
* The text of the exception message clarified (to cover both cases, when there is no `timeframes` attribute and when it's None), test adjusted with the new exception text.
* Test added for `'timeframes': None` in the markets.

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -269,12 +269,15 @@ class Exchange(object):
         """
         Checks if ticker interval from config is a supported timeframe on the exchange
         """
-        if not hasattr(self._api, "timeframes"):
-            # If timeframes is missing, the exchange probably has no fetchOHLCV method.
+        if not hasattr(self._api, "timeframes") or self._api.timeframes is None:
+            # If timeframes attribute is missing (or is None), the exchange probably
+            # has no fetchOHLCV method.
             # Therefore we also show that.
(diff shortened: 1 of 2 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -269,6 +269,15 @@ def validate_timeframes(self, timeframe: List[str]) -> None:
         """
         Checks if ticker interval from config is a supported timeframe on the exchange
         """
+        if not hasattr(self._api, "timeframes") or self._api.timeframes is None:
```
xmatthias: Technically looks good - but you'll want to remove lines 281-286 (they do the same thing, but worded differently).

I assume that was your intent but during testing you added it and then forgot to remove ...

hroff-1902: Ah, shame on mee, thanks! 

I forgotten it (to remove the original part of the code) when applied the change manually from the stash bunch of changes printed as a patch, stashed with other changes from some other local working branch, where I made this change first...

PR #1998, opened 2019-07-03
Support all types of pairs for /balance

### LINKED ISSUE #1981: Pax balance not detected
Version 2019.5
Bot won't detect pax base, will update to newest to see if fixed. Will also assumes that this is the same issue for other stable coins that are not tether 

Last version of CCTX
Running 2019.6
Exchange Binance
Edited to add more information about the issue

xmatthias:
## Summary
Some pairs are not available as <pair>/BTC - but BTC/<pair>. These pairs are often stake-currencies (like PAX).
we used to have a hardcoded list of these pairs, however using the exchanges "markets" for this is way more flexible, and should eliminate this bug for many other currencies (USDC, ...) too.

Closes #1981

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -270,6 +270,15 @@ class Exchange(object):
                     f'Pair {pair} is not available on {self.name}. '
                     f'Please remove {pair} from your whitelist.')
 
+    def get_valid_pair_combination(self, paira, pairb) -> str:
+        """
+        Get valid combination of paira and pairb by trying both combinations.
+        """
+        for pair in [f"{paira}/{pairb}", f"{pairb}/{paira}"]:
+            if pair in self._api.markets and self._api.markets[pair].get('active'):
+                return pair
+        raise DependencyException(f"Could not combine {paira} and {pairb} to get a valid pair.")
+
     def validate_timeframes(self, timeframe: List[str]) -> None:
         """
         Checks if ticker interval from config is a supported timeframe on the exchange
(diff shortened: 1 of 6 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -281,10 +281,11 @@ def _rpc_balance(self, fiat_display_currency: str) -> Dict:
                 rate = 1.0
             else:
                 try:
-                    if coin in('USDT', 'USD', 'EUR'):
-                        rate = 1.0 / self._freqtrade.get_sell_rate('BTC/' + coin, False)
+                    pair = self._freqtrade.exchange.get_valid_pair_combination(coin, "BTC")
+                    if pair.startswith("BTC"):
+                        rate = 1.0 / self._freqtrade.get_sell_rate(pair, False)
                     else:
```
hroff-1902: Need to think a bit about logic imlemented here...

xmatthias: maybe to explain: 
we want the currency in BTC so we can filter (I think it should be configurable, but let's leave that for another day)
so we either multiply by the rate (`rate * price` - or divide (`1/rate * price` or `price / rate`).

This logic is not new - just the filter when to use what method is...

freqtrade/exchange/exchange.py
```diff
@@ -270,6 +270,15 @@ def validate_pairs(self, pairs: List[str]) -> None:
                     f'Pair {pair} is not available on {self.name}. '
                     f'Please remove {pair} from your whitelist.')
 
+    def get_valid_pair_combination(self, paira, pairb) -> str:
```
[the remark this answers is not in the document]
xmatthias: Great catch! :+1:

PR #2002, opened 2019-07-06
minor: refactoring arguments and configuration

hroff-1902:
First part of refactoring, consists of a set of mostly minor improvements. See the list of commits.

I decided to split my current large set of changes so that the main idea will be more clear and readable for review at the end.

Next part of changes will follow.

### DIFF
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -298,7 +298,7 @@ ARGS_STRATEGY = ["strategy", "strategy_path"]
 
 ARGS_MAIN = ARGS_COMMON + ARGS_STRATEGY + ["dynamic_whitelist", "db_url", "sd_notify"]
 
-ARGS_COMMON_OPTIMIZE = ["loglevel", "ticker_interval", "timerange",
+ARGS_COMMON_OPTIMIZE = ["ticker_interval", "timerange",
                         "max_open_trades", "stake_amount", "refresh_pairs"]
 
 ARGS_BACKTEST = ARGS_COMMON_OPTIMIZE + ["position_stacking", "use_max_market_positions",
--- a/freqtrade/arguments.py
+++ b/freqtrade/arguments.py
@@ -43,11 +43,10 @@ AVAILABLE_CLI_OPTIONS = {
         help='Log to the file specified.',
         metavar='FILE',
     ),
-
     "version": Arg(
         '--version',
         action='version',
-        version=f'%(prog)s {__version__}'
+        version=f'%(prog)s {__version__}',
     ),
     "config": Arg(
         '-c', '--config',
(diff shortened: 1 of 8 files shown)


freqtrade/arguments.py
```diff
@@ -106,7 +106,8 @@ def __init__(self, *args, **kwargs):
     "max_open_trades": Arg(
         '--max_open_trades',
         help='Specify max_open_trades to use.',
-        type=int,
+        type=check_int_positive,
```
xmatthias: `max_open_trades` can (should / must) be set to -1 for edge, so i think check_int_positive is wrong for this argument (yes, you would normally set this in the configuration anyway - but - ...)

freqtrade/loggers.py
```diff
@@ -0,0 +1,50 @@
+import logging
```
xmatthias: I like splitting out logic to files, so it's easy to open them.

This got me thinking if we should have a module (folder) for these things to keep the structure cleaner... (think `freqtrade.utils.loggers`).

to be clear, It's just a thought asking for opinion - not something i'd want to do in this PR ...

freqtrade/main.py
```diff
@@ -32,8 +31,6 @@ def main(sysargv: List[str] = None) -> None:
     return_code: Any = 1
     worker = None
     try:
-        set_loggers()
```
xmatthias: My initial thought was that this may show uninitialized log messages, but i could not find a way to support my thought :+1:

freqtrade/loggers.py
```diff
@@ -0,0 +1,50 @@
...
+from logging.handlers import RotatingFileHandler
+from typing import Any, Dict, List
+
+
+logger = logging.getLogger(__name__)
+
+
+def _set_loggers(verbosity: int = 0) -> None:
+    """
+    Set the logging level for Third party libs
```
xmatthias: Nitpick: Third party libraries ?

freqtrade/arguments.py
hroff-1902: Ah, year, I forgot about this, thanks...
Done, returned it back to int.

freqtrade/main.py
hroff-1902: It's called here with verbosity set to 0 anyway.

freqtrade/loggers.py
hroff-1902: 👍 Done.

hroff-1902:
> I like splitting out logic to files, so it's easy to open them.
> 
> This got me thinking if we should have a module (folder) for these things to keep the structure cleaner... (think `freqtrade.utils.loggers`).
> 
> to be clear, It's just a thought asking for opinion - not something i'd want to do in this PR ...

Well, the next question I'd like to ask you about right after this PR is the following: I'd like to move json verification stuff out of configuration.py and the Configuration class to a separate file. And I was thinking about conversion configuration.py to the configuration/ dir with it's own __init__.py exporting the Configuration class, while json validation should go to a separate python file under the configuration/ dir. Any thoughts?

PR #2004, opened 2019-07-08
Don't run the bot with python3 freqtrade

### LINKED ISSUE #1990: "No module named 'freqtrade'" error should be addressed in the FAQ
This is one of the most asked questions by the novice users. It should be covered in the FAQ docs.

I mean an error similar to:

``` bash
$ python3 freqtrade -c config.json
 File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
   "__main__", mod_spec)
 File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
   exec(code, run_globals)
 File "freqtrade/__main__.py", line 9, in <module>
   from freqtrade import main
ModuleNotFoundError: No module named 'freqtrade'
```

In my environment (Ubuntu), setting `export PYTHONPATH=.:$PYTHONPATH` helps. In Windows, there is an analog -- just set PYTHONPATH env. var in the System Variables so that it point to the path where Freqtrade is installed.

@xmatthias If you have another recipe, please offer. At least, this should be covered in the FAQ.

xmatthias:
we can either use `python3 -m freqtrade ...` or `freqtrade ...` to run the bot (`python3 freqtrade` requires some ugly hacks with PYTHON_PATH.

- and shorter should be better.

## Summary

closes #1990 

## Details:

if `freqtrade` is not available as binary - then the installation did not go well or the virtual Environment has not been activated.

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -24,37 +24,37 @@ The backtesting is very easy with freqtrade.
 #### With 5 min tickers (Per default)
 
 ```bash
-python3 freqtrade backtesting
+freqtrade backtesting
 ```
 
 #### With 1 min tickers
 
 ```bash
-python3 freqtrade backtesting --ticker-interval 1m
+freqtrade backtesting --ticker-interval 1m
 ```
 
 #### Update cached pairs with the latest data
 
 ```bash
-python3 freqtrade backtesting --refresh-pairs-cached
+freqtrade backtesting --refresh-pairs-cached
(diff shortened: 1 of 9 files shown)


hroff-1902:
As you may know, my env. is not standard and I found that I don't have freqtrade installed, so `$ freqtrade` simply does not work... Tried to make it installed -- not successfuly yet, see #2006.

PR #2010, opened 2019-07-10
Fix non-rendering docs

xmatthias:
## Summary
Currently some of the danger boxes don't render, this is trying to fix that.


hroff-1902:
LGTM 👍

PR #2015, opened 2019-07-11
Make configuration a module

hroff-1902:
The patch is wider than I expected but this mostly is due to the tests.

* configuration is made a sep. module in a sep. directory
* arguments moved also to the new configuration module. Because arguments are also a part of configuration.
* validation of json schema separated from configuration.py
* check_exchange() separated from configuration.py
* tests adjusted and improved (2 common functions made in conftest)
* create_datadir() moved to its own separate file

### DIFF
--- a/freqtrade/configuration/__init__.py
+++ b/freqtrade/configuration/__init__.py
@@ -1,8 +1,2 @@
 from freqtrade.configuration.arguments import Arguments, TimeRange  # noqa: F401
-
-from freqtrade.configuration.arguments import (  # noqa: F401
-        ARGS_DOWNLOADER,
-        ARGS_PLOT_DATAFRAME,
-        ARGS_PLOT_PROFIT)
-
 from freqtrade.configuration.configuration import Configuration  # noqa: F401
--- a/freqtrade/configuration/__init__.py
+++ b/freqtrade/configuration/__init__.py
@@ -1,0 +1,8 @@
+from freqtrade.configuration.arguments import Arguments, TimeRange  # noqa: F401
+
+from freqtrade.configuration.arguments import (  # noqa: F401
+        ARGS_DOWNLOADER,
+        ARGS_PLOT_DATAFRAME,
+        ARGS_PLOT_PROFIT)
+
+from freqtrade.configuration.configuration import Configuration  # noqa: F401
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -3,13 +3,13 @@ This module contains the configuration class
 """
 import json
 import logging
-import os
 import sys
 from argparse import Namespace
 from typing import Any, Callable, Dict, Optional
 
 from freqtrade import OperationalException, constants
 from freqtrade.configuration.check_exchange import check_exchange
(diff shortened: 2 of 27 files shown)


freqtrade/configuration/configuration.py
```diff
@@ -3,17 +3,14 @@
 """
```
xmatthias: I think we need to either rename the module, or this file.

This can create confusion (which "configuration" is intendet this time?) - and can produce very subtle bugs when things are imported, since it's a namespace collision ("configuration" is both a module, and a file).

also, think about importing:
`from freqtrade.configuration.configuration import Configuration` ...

freqtrade/configuration/__init__.py
```diff
@@ -0,0 +1,8 @@
+from freqtrade.configuration.arguments import Arguments, TimeRange  # noqa: F401
+
+from freqtrade.configuration.arguments import (  # noqa: F401
```
xmatthias: I wouldn't copy/import these constants here - we can also use direct imports from the few places we need them ...

freqtrade/tests/conftest.py
```diff
@@ -39,10 +39,24 @@ def log_has_re(line, logs):
                   False)
 
 
-def get_args(args) -> List[str]:
+def get_args(args):
     return Arguments(args, '').get_parsed_arg()
 
 
+def patched_configuration_open(mocker, config):
+    file_mock = mocker.patch('freqtrade.configuration.configuration.open', mocker.mock_open(
```
xmatthias: Ideally we should only have one of these 2 helper functions (and prefferably the other one)...

freqtrade/configuration/configuration.py
hroff-1902: `freqtrade/exchange/exchange.py`...

xmatthias: lol yes :D

hroff-1902: I'm not against your proposal, not at all, but cannot find other appropriate names for the configuration dir and/or configuration.py...

xmatthias: well me neither (that's why i didn't propose one)

the best is "config" and "configuration" - but that's even more misleading since the config dict is called "config" ... 

so let's leave it as is :+1:

freqtrade/tests/test_configuration.py
```diff
@@ -150,9 +143,8 @@ def test_load_config(default_conf, mocker) -> None:
 
 
 def test_load_config_with_params(default_conf, mocker) -> None:
-    mocker.patch('freqtrade.configuration.open', mocker.mock_open(
-        read_data=json.dumps(default_conf)
-    ))
+    patched_configuration_open(mocker, default_conf)
```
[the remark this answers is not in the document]
hroff-1902: Nice, I like it. Done.

freqtrade/tests/conftest.py
hroff-1902: Done, see also comment above.

freqtrade/configuration/__init__.py
hroff-1902: Agree, done.

PR #2019, opened 2019-07-14
[Minor] Small code cleanups

xmatthias:
## Summary
Some small code cleanups found after analyzing the codebase with deepcode.ai (a company sponsoring Europython this year).

## Quick changelog

- Replace usage of `filter` with comprehensions
- don't overwrite `max` argument

### DIFF
--- a/freqtrade/tests/strategy/test_strategy.py
+++ b/freqtrade/tests/strategy/test_strategy.py
@@ -69,8 +69,8 @@ def test_load_strategy(result):
 
 
 def test_load_strategy_base64(result):
-    with open("freqtrade/tests/strategy/test_strategy.py", "r") as file:
-        encoded_string = urlsafe_b64encode(file.read().encode("utf-8")).decode("utf-8")
+    with open("freqtrade/tests/strategy/test_strategy.py", "rb") as file:
+        encoded_string = urlsafe_b64encode(file.read()).decode("utf-8")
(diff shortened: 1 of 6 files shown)


freqtrade/tests/strategy/test_strategy.py
```diff
@@ -69,8 +69,8 @@ def test_load_strategy(result):
 
 
 def test_load_strategy_base64(result):
-    with open("freqtrade/tests/strategy/test_strategy.py", "r") as file:
-        encoded_string = urlsafe_b64encode(file.read().encode("utf-8")).decode("utf-8")
+    with open("freqtrade/tests/strategy/test_strategy.py", "rb") as file:
```
hroff-1902: Are you sure about "rb" mode? This may produce (not sure, do not remember right now how base64 behaves here) diff. results for Linux/MacOS and Windows after base64-encoding (due to the differences in CRLF)...

xmatthias: good point, I'll test it on windows tomorrow (not concerned about linux/mac difference, these should be fine.

However, as we used to `read().encode("utf-8)` (normally you `decode()`) it's a very oddly written test anyway

xmatthias: There's no problem with tests on windows (neither on WSL, nor in "real" windows mode).

PR #2023, opened 2019-07-15
minor: configuration cleanup

hroff-1902:
Further configuration code cleanup, no changes to public interfaces or behavior...

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -40,48 +40,19 @@ class Configuration(object):
 
         return self.config
 
-    def load_config(self) -> Dict[str, Any]:
+    def _load_config_files(self) -> Dict[str, Any]:
         """
-        Extract information for sys.argv and load the bot configuration
-        :return: Configuration dictionary
+        Iterate through the config files passed in the args,
+        loading all of them and merging their contents.
         """
         config: Dict[str, Any] = {}
-        # Now expecting a list of config filenames here, not a string
+
+        # We expect here a list of config filenames
         for path in self.args.config:
             logger.info('Using config: %s ...', path)
 
             # Merge config options, overwriting old values
             config = deep_merge_dicts(self._load_config_file(path), config)
-
-        if 'internals' not in config:
-            config['internals'] = {}
-
-        logger.info('Validating configuration ...')
-        validate_config_schema(config)
-        self._validate_config_consistency(config)
-
-        # Set strategy if not specified in config and or if it's non default
-        if self.args.strategy != constants.DEFAULT_STRATEGY or not config.get('strategy'):
-            config.update({'strategy': self.args.strategy})
-
-        if self.args.strategy_path:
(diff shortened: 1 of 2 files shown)


freqtrade/configuration/configuration.py
```diff
@@ -285,14 +281,11 @@ def _load_optimize_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
         self._args_to_config(config, argname='hyperopt_min_trades',
                              logstring='Parameter --min-trades detected: {}')
 
-        return config
-
-    def _load_plot_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
+    def _load_plot_config(self, config: Dict[str, Any]) -> None:
```
xmatthias: I don't mind not returning `config` here - but should we really pass `config` in in that case, or should we use the `self.config` / `self._config` attribute instead (which is assigned by the resulting config object)?

The only "needed" public method is `get_config()` (`load_config()` is used but can be replaced by the above). - then we'd have it kindof aligned... ?

freqtrade/configuration/configuration.py
```diff
@@ -201,12 +198,11 @@ def _load_datadir_config(self, config: Dict[str, Any]) -> None:
             config.update({'datadir': create_datadir(config, None)})
         logger.info('Using data directory: %s ...', config.get('datadir'))
 
-    def _load_optimize_config(self, config: Dict[str, Any]) -> Dict[str, Any]:
+    def _load_optimize_config(self, config: Dict[str, Any]) -> None:
```
xmatthias: I don't like the naming of `_load_*` (it's not really correct)

we have 8 functions starting with `_load` - only one of which actually loads something (`_load_config_files()`), all others simply copy arguments from args to config.

We may be able to consolidate these into one (since the syntax for these is now a lot shorter).

Maybe not for this PR - just something i noticed now.

hroff-1902: Heh, I already renamed them to `_process_*_options()` (like `_process_optimize_options()`)... I split wide changes to the smaller ones so that the changes remain observable, so I planned to commit it in the next PR along with further refactoring, but particularly this renaming can be pushed in the PR, I guess...

freqtrade/configuration/configuration.py
hroff-1902: Agree, will change it to `self._config`...

As for `get_config()/load_config()` -- yes, there is no actually need in using both of them publicy, but currently both are used in the freqrade codebase. Will probably change it (maybe in a sep. PR)

hroff-1902: No, let's postpone these changes.
Passing config to these functions used in tests and in the scripts. `load_config` is used by the scripts, so let's postpone these changes till all scripts become subcommands and this code simply disappears...

freqtrade/configuration/configuration.py
hroff-1902: This one is done and committed. If you agree to postpone your previous proposal, I assume this is ready for final review...

Issue #1907, opened 2019-06-05
Enhancement - Add Sharpe Ratio for strategies backtesting 


Dante2333:
https://corporatefinanceinstitute.com/resources/templates/excel-modeling/sharpe-ratio-calculator/

Introduce Sharpe Ratio for strategies for backtesting
Additional command or parameters for backtesting: --risk-free=X; 

then --risk-free is a number between 0.00 and 1.00, then adjust to sharpe calculation in months if months for strategies > 1; in years if years for strategy > 3

Add new column on strategy summary: Sharpe Ratio

yperfanov:
I have the Sharpe Ratio implemented but is a copy paste from a Udemy freqtrade course I took. 
@hroff-1902  do you think I can make a pull request with that code?

hroff-1902:
> I have the Sharpe Ratio implemented but is a copy paste from a Udemy freqtrade course I took.
> @hroff-1902 do you think I can make a pull request with that code?

Sure, that addendum should also be licensed under GPL (otherwise it cannot be used along with GPLed freqtrade and distributed) so no problem to borrow it. You may add a link to them somehow in the comments to keep the attribution of the authors of the code.

wolvesproduction:
Yes I took the same course on Udemy and I trully believe its a great feature that will bring better and more accurate results. It only needs a few changes in hyperopt.py. I'd be more than happy to make this happen.

hroff-1902:
Thanks for the example of the code.

The problem with it does not consist in python code for calculating the ratio. The problem is that in case of many ratios and other parameters added to the existing table, the result can become overloaded and unreadable. So one the ideas by @xmatthias was to add a new table to the backtesting results for this kind of ratios...

wolvesproduction:
Well the function is replacing the current calculate_loss function in hyperopt.py. I've inspected the whole file and the function is not calling any table. Basically, it just makes an on-the-fly calculation based on the current profit and trades count. There're maybe 3 changes in the whole file and these are simple arithmetic calculation. 

Maybe I'm missing something under the hood? Could you elaborate?

hroff-1902:
> Well the function is replacing the current calculate_loss function in hyperopt.py. 

Ah, that's another point (replacing the calculate_loss and/or objective optimization function in the hyperopt). @Dante2333 wrote in his original suggestion above: "Add new column on strategy summary: Sharpe Ratio"

[...]

PR #2024, opened 2019-07-15
Custom hyperopt loss function (and sharpe-ratio)

xmatthias:
## Summary
Allow usage of custom loss-functions for hyperopt, and some minor performance improvements

Solve the issue: #1775

## Quick changelog

- Add `--hyperopt-loss <lossclass>` switch
- Allow defining a custom loss-function class in a seperate file
- Add sharpe-ratio loss function (based on https://github.com/freqtrade/freqtrade/issues/1907#issuecomment-505524849)
- Don't check using `hasattr` in `generate_optimizer` - this function is called on every iteration (performance)
- Add `--continue` argument to hyperopt (i was anoyed to delete the files each time during testing...)

*Note*:
 by using the `..., *args, **kwargs)` section in the documentation, we can add (never remove!) new parameters at will without ever breaking user code.

**EDIT**:
Changed description to match how the new implementation works.


hroff-1902:
* Let's call it 'hyperopt objective' or 'hyperopt objective function' -- it's not Loss...

* My generic idea I mentioned was to add a resolver for objective functions (which will search for a objects with type function in a user_data subdir (like in 'user_data/hyperopt_objectives' or 'user_data/hyperopts/objectives') with a given name and use it, In my opinion, it should be separate from particular hyperopt, since I would like to run same (multiple) hyperopts with diff. objectives. A default one is to be provided in the freqtrade/optimize catalog, plus a few ones prepared to choose from/inspire users (optimize with only profit as component (I found it useful), optimize for Sharp Ratio, etc.) However, this approach could probably be combined with objective function defined in the hyperopt file...

hroff-1902:
The `--clean` option is extremely useful. I recently always run hyperopts as `rm user_data/hyperopt_*; freqtrade hyperopt`... 😆 

However, my intention was to provide the `--continue` option instead. Rationale: even if you do not change the hyperopt codebase and do not change the spaces, it's often more useful to start hyperopting from scratches rather than proceed with previous results. The optimizer uses bayesian descend, it does not go through a random grid, so usually there is not absulutely no needs to run it for long raws of evals -- it will descend to a single local minimum of the objective anyway. So it's almost always more useful to run  5 raws of 1000 evals than 1 raw of 5000 evals -- it will find you 5 different local minimums while in the latter case it will descend to only one. And if you change something in the hyperopt, as known, you should also run hyperopt from scratches. So in many cases it's more preferable for a user to run hyperopt from scratches rather than continue previous hyperopting path. Thus the `--continue` option with a reverse meaning to `--clean` seems to be more useful...

hroff-1902:
> Don't check using hasattr in generate_optimizer - this function is called on every iteration (performance)

I thought those checks were placed in the `generate_optimizer()` by intention due to some problems with pickling these values in the optimizer worker processes. Fine if they can really be moved to `__init__()`...

hroff-1902:
I would call 'legacy' loss function (objective function) as 'default'.

freqtrade/optimize/hyperopt.py
```diff
@@ -184,39 +198,32 @@ def hyperopt_space(self) -> List[Dimension]:
...
-
         min_date, max_date = get_timeframe(processed)
 
         results = self.backtest(
             {
                 'stake_amount': self.config['stake_amount'],
                 'processed': processed,
-                'max_open_trades': max_open_trades,
+                'max_open_trades': self.max_open_trades,
                 'position_stacking': self.config.get('position_stacking', False),
```
hroff-1902: this can also be moved to __init__() and self.position_stacking be used here...

docs/hyperopt.md
```diff
@@ -150,6 +151,45 @@ The above setup expects to find ADX, RSI and Bollinger Bands in the populated in
...
+        trade_loss = 1 - 0.25 * exp(-(trade_count - TARGET_TRADES) ** 2 / 10 ** 5.8)
+        profit_loss = max(0, 1 - total_profit / EXPECTED_MAX_PROFIT)
+        duration_loss = 0.4 * min(trade_duration / MAX_ACCEPTED_TRADE_DURATION, 1)
+        result = trade_loss + profit_loss + duration_loss
+        return result
+```
+
+Currently, the arguments are:
+
+* `results`: DataFrame containing the result
```
hroff-1902: is it really a DataFrame?

docs/hyperopt.md
```diff
@@ -150,6 +151,45 @@ The above setup expects to find ADX, RSI and Bollinger Bands in the populated in
...
+* `trade_count`: Amount of trades (identical to `len(results)`)
+* `min_date`: Start date of the hyperopting TimeFrame
+* `min_date`: End date of the hyperopting TimeFrame
+
+This function needs to return a floating point number (`float`). The smaller that number, the better is the result. The parameters and balancing for this are up to you.
+
+!!! Note
+    This function is called once per iteration - so please make sure to have this as optimized as possible to now slow hyperopt down unnecessarily.
```
hroff-1902: to not slow

docs/hyperopt.md
```diff
@@ -150,6 +151,45 @@ The above setup expects to find ADX, RSI and Bollinger Bands in the populated in
...
+
+!!! Note
+    The last 2 arguments, `*args` and `**kwargs` are not strictly necessary but ensure compatibility for the future, so we can easily enable more parameters once we discover what could be usefull without breaking your custom hyperopt file.
```
hroff-1902: this note should probably be removed, it's too deep for a user, she should only follow the proposed interface for the function.

docs/hyperopt.md
```diff
@@ -150,6 +151,59 @@ The above setup expects to find ADX, RSI and Bollinger Bands in the populated in
...
+        trade_loss = 1 - 0.25 * exp(-(trade_count - TARGET_TRADES) ** 2 / 10 ** 5.8)
+        profit_loss = max(0, 1 - total_profit / EXPECTED_MAX_PROFIT)
+        duration_loss = 0.4 * min(trade_duration / MAX_ACCEPTED_TRADE_DURATION, 1)
+        result = trade_loss + profit_loss + duration_loss
+        return result
+```
+
+Currently, the arguments are:
+
+* `results`: DataFrame containing the result
```
hroff-1902: "containing the results of backtesting for the hyperopt epoch".

And we need to describe the fields (columns) available.

xmatthias:
> Thus the --continue option with a reverse meaning to --clean seems to be more useful...

works perfectly fine for me - i just needed "something" ...

the `hasattr()` calls were placed wrongly (by myself) - i guess my assumption was that generate_optimizer is not called that often ... (not sure what i was thinking).

> let's call it 'hyperopt objective' or 'hyperopt objective function' -- it's not Loss...

I disagree with this ...
To quote wikipedia: 

> In mathematical optimization and decision theory, a loss function or cost function is a function that maps an event or values of one or more variables onto a real number intuitively representing some "cost" associated with the event. An optimization problem seeks to minimize a loss function. An objective function is either a loss function or its negative (in specific domains, variously called a reward function, a profit function, a utility function, a fitness function, etc.), in which case it is to be maximized.
source: [wikipedia](https://en.wikipedia.org/wiki/Loss_function)

We are working on a optimization problem - and hence ` An optimization problem seeks to minimize a loss function.` applies. (also for bayesian - see further down on that link).

I love your way with a Resolver, and i'll change it to that now.

docs/hyperopt.md
```diff
@@ -200,7 +199,9 @@ class SuperDuperHyperOptLoss(IHyperOptLoss):
 
 Currently, the arguments are:
 
-* `results`: DataFrame containing the result
+* `results`: DataFrame containing the result  
+    The following columns are available in results (corresponds to the output of backtesting):
```
hroff-1902: "corresponds to the output of backtesting" -- this is a bit incorrect since output of backtesting contains results aggregated by pair name while this list contains details on the per-trade basis with no aggregation.

hroff-1902:
Fine, I like it 👍 except the term 'Loss' used. Will provide some more considerations this evening.

hroff-1902:
For the Sharpe Ratio sample objective function class -- leave it as it is now, if it's really incorrect (I mean the considerations I posted in #1907), we'll always be able to rename it to, say, <s>UdemyCourseHyperOptLoss</s> `SomeNameHyperOptLoss` and move to user_data if (and when) I provide a true correct implementation of SR... This one may also be very useful as objective for hyperoptimization.

docs/hyperopt.md
xmatthias: Well it's the "output-file" of backtesting (but agreed, could be clearer).

docs/hyperopt.md
```diff
@@ -150,6 +150,70 @@ The above setup expects to find ADX, RSI and Bollinger Bands in the populated in
...
+    @staticmethod
+    def hyperopt_loss_function(results: DataFrame, trade_count: int,
+                               min_date: datetime, max_date: datetime,
+                               *args, **kwargs) -> float:
+        """
+        Objective function, returns smaller number for better results
+        This is the legacy algorithm (used until now in freqtrade).
+        Weights are distributed as follows:
+        * 0.4 to trade duration
+        * 0.25: Avoiding trade loss
```
hroff-1902: `* 1.0 to total profit, compared to the expected value defined above`

In the code as well, not only in the docs.

freqtrade/optimize/hyperopt.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, config: Dict[str, Any]) -> None:
         super().__init__(config)
         self.custom_hyperopt = HyperOptResolver(self.config).hyperopt
 
+        self.custom_hyperoptloss = HyperOptLossResolver(self.config).hyperoptloss
+        self.calculate_loss = self.custom_hyperoptloss.hyperopt_loss_function
+
         # set TARGET_TRADES to suit your number concurrent trades so its realistic
         # to the number of days
         self.target_trades = 600
```
hroff-1902: Not used anymore

freqtrade/optimize/hyperopt.py
```diff
@@ -63,10 +66,40 @@ def __init__(self, config: Dict[str, Any]) -> None:
         # Note, this is ratio. 3.85 stated above means 385Σ%.
         self.expected_max_profit = 3.0
```
hroff-1902: Not used anymore

hroff-1902: self.max_accepted_trade_duration as well is not used

user_data/hyperopts/sample_hyperopt.py
```diff
@@ -1,17 +1,31 @@
...
+# This is assumed to be expected avg profit * expected trade count.
+# For example, for 0.35% avg per trade (or 0.0035 as ratio) and 1100 trades,
+# self.expected_max_profit = 3.85
+# Check that the reported Σ% values do not exceed this!
+# Note, this is ratio. 3.85 stated above means 385Σ%.
+EXPECTED_MAX_PROFIT = 3.0
+
+# max average trade duration in minutes
+# if eval ends with higher value, we consider it a failed eval
+MAX_ACCEPTED_TRADE_DURATION = 300
```
hroff-1902: A minor addtional note: do we need these constants in the hyperopt? It's not the loss...

xmatthias: no ... forgot to remove them after the refactoring, good catch!

hroff-1902:
A note of the term Loss: the term 'Loss Function' is, yes, a type of objective functions, that definition works well as a generic definition, but in trading, in our knowledge domain, the term Loss has different and well known meaning.

A user don want to optimize Losses, she does not want Losses at all, she wants to optimize her Profits...

That's why I proposed to use 'objective' and 'objective function' everywhere -- in the docs and in the names of the vars, functions and classes -- because Loss (especially without 'function', used as simply 'loss' somewhere) may be confusing for an unexperienced, novice user...

hroff-1902:
I will add later in the next PR a couple of sample implementations of loss function I use in practice that can be choosen and can inspire users for her own variants... (to user_data dir)

xmatthias:
Well your argument is self-defeating - a user does not want losses - so minimizing loss is what the user wants ... (which in the end, equals to profits ...).

Also i find it funny that in the next comment you use "loss function" yourself ... ;) 

using "objective function" will be unfamiliar to most who have ever done something with machine learning - and honestly -  confusing, since the traders objective is to maximize profits (so i would instinctively return higher values, instead of lower values).

Using Loss Function will make sense (from a direction point of view) since i want to "minimize losses" - in both the hyperopt context, as well as in the trading context.
 
And honestly, to write a good hyperopt script, you need to be somewhat familiar with general concepts of machine learning (not AI - that only exists in powerpoints :laughing: ) - otherwise some things will be confusing

hroff-1902:
> Well your argument is self-defeating - a user does not want losses - so minimizing loss is what the user wants ... (which in the end, equals to profits ...).

No, because if the hyperopt prints non-negative Profit for some epoch -- it's not Loss (I know, I know, from the math. point of view porfit is negated loss, I now, but I'm talking about philistine's points of vew...). It's profit. Especially in the eyes of a novice user.

'Objective' is just more neutral. As you see, I approved the PR so you do not agree with this my rationale -- nevermind, it's not blocking for this PR. (Though, the names of the option and classes are parts of the UI so it will be painful to change it in the future. That's why I'm discussing this here...)

xmatthias:
> Though, the names of the option and classes are parts of the UI so it will be painful to change it in the future. That's why I'm discussing this here...)

that's why i didn't merge it yet.

my fear is that both versions will be confusing ... 
* objective as in "my objective is" ... 
* loss because "i don't want loss" ... 

How would we call the interface? `IHyperOptObjective`?
the naming becomes odd soon because an object is not an objective ...


Another idea would be "minimize" ... but that'll again be odd since i want to maximize my profits, not minimize ...

user_data/hyperopts/sample_hyperopt_loss.py
```diff
@@ -0,0 +1,47 @@
...
+
+# max average trade duration in minutes
+# if eval ends with higher value, we consider it a failed eval
+MAX_ACCEPTED_TRADE_DURATION = 300
+
+
+class SampleHyperOptLoss(IHyperOptLoss):
+    """
+    Defines the default loss function for hyperopt
+    This is intendet to give you some inspiration for your own loss function.
```
hroff-1902: 'intended'

user_data/hyperopts/sample_hyperopt_loss.py
```diff
@@ -0,0 +1,47 @@
...
+
+    The Function needs to return a number (float) - which becomes for better backtest results.
```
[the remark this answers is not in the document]
hroff-1902: I mean 'function', not 'Function'

hroff-1902:
> * objective as in "my objective is" ...
Your objective is to optimize strategy using hyperopt. In the same time it's an objective for hyperopt

> the naming becomes odd soon because an object is not an objective ...
Why not? It implements hyperopt objective function, i.e. it itself implements (i.e. it actually is) the hyperopt objective (hyperopt goal, hyperopt target).

> Another idea would be "minimize" ... but that'll again be odd since i want to maximize my profits, not minimize ...

No, this is a bad term I think.

> How would we call the interface? `IHyperOptObjective`?
Yes. Same in the vars and funcs (use 'Objective' for camel naming and '_objective' for vars and functions, in the names of files and dirs if needed). In the docs, we can always use the term 'objective function' everywhere it's mentioned. Plus in the beginning of hyperopt.md say the hyperopt blah-blah uses `scikit-optimize` to minimize the objective function. That's my proposal.

Another possibility is to use '_loss' in the names of vars, funcs and file/dir names (as it's used now, so the additional rationale is that this is used due to historical reasons). AND use more detailed term 'loss function' (not simply 'loss') everywhere in the docs and in the comments. To avoid confusion with the trading losses. That's another possibility I don't not strictly mind it and I am now thinking that this is actually okay -- just review all the docs and the comments and use 'loss function' (not 'loss') in there... Let's make it this way. Due to historical reasons among other rationales behind using this term.

xmatthias:
> No, this is a bad term I think.

It was an idea, not one i necessarily liked ... :)

it may be historical - but technically (from a data science perspective) using loss function is not wrong.

I've reworded the documentation (and fixed a few bugs while beeing there)...

docs/hyperopt.md
```diff
@@ -144,25 +144,27 @@ it will end with telling you which paramter combination produced the best profit
...
 ## Loss-functions
 
-Each hyperparameter tuning requires a target. This is usually defined as a loss function, which get's closer to 0 for increasing values.
+Each hyperparameter tuning requires a target. This is usually defined as a loss function (sometimes also called objective function), which should decrease for more desirable results, and increase for bad results.
 
-FreqTrade uses a default loss function, which has been with freqtrade since the beginning and optimizes mostly for short trade duration and avoiding losses.
+By default, FreqTrade uses a loss function, which has been with freqtrade since the beginning and optimizes mostly for short trade duration and avoiding losses.
 
-A different version this can be used by using the `--hyperopt-loss <Class-name>` argument.
+A different version this can be used by using the `--hyperopt-loss-class <Class-name>` argument.
```
hroff-1902: Wait a minute please, seems I missed it, sorry...

Is this option actually called `--hyperopt-loss-class` now? In this case the `--strategy` option can also be called `--strategy-class`... Let's rename it (as minimum) to `--hyperopt-loss` (which sounds so-so, returning back to our discussion on the meaning of Loss) or (a compromize and the current proposal after I found I missed this point) -- let's name it `--hyperopt-objective`, describe in the configuration.md and in the option help string that it defines 'the name of the class with the hyperopt objective loss function' or something like that, and leave all the '_loss' and 'Loss' in the code in the names of the vars, funcs and classes...

What I don't like here is `-class` in the name of the option...

docs/hyperopt.md
```diff
@@ -144,25 +144,27 @@ it will end with telling you which paramter combination produced the best profit
...
 This class should be in it's own file within the `user_data/hyperopts/` directory.
 
-### Using a custom loss function
+Currently, the following loss-functions are builtin: `SharpeHyperOptLoss` and `DefaultHyperOptLoss`.
```
hroff-1902: 'loss functions'

docs/hyperopt.md
```diff
@@ -246,12 +251,11 @@ use data from directory `user_data/data`.
 
 ### Running Hyperopt with Smaller Testset
 
-Use the `--timerange` argument to change how much of the testset
-you want to use. The last N ticks/timeframes will be used.
-Example:
+Use the `--timerange` argument to change how much of the testset you want to use.
+To use one month of data, use the following parameter:
```
hroff-1902: 'For instance, to use...' (or restore 'Example:' here)

(too many 'use' words in these 2 lines...)

hroff-1902: The rest is fine, nice work. 👍

docs/hyperopt.md
xmatthias: well the nice thing about argparse is that you can pass `--hyperopt-loss` and it'll work just fine until it finds a conflict (so --hyperopt will not work anymore, but `--hyperopt-l` should.

I thought being explicit (it's a class we expect, not a filename) would help - but if you see it differently i can remove this.

calling the option `hyperopt-objective` - but everything else is "loss" does make no sense to me, and is wildly confusing.

hroff-1902:
Okay, let it be like this. LGTM 👍

hroff-1902:
`--continue` should have been issued and merged in a sep. PR...

xmatthias:
yes, i know - i though it'd be a quick PR ... :D

hroff-1902:
Congrats! :hah

PR #2029, opened 2019-07-17
[minor] Convert create_datadir to Pathlib

xmatthias:
## Summary
* Using the more modern pathlib instead of os.path.
* we should also not assume user_data is in parent.parent.parent but use cwd() instead (this is not fully implemented yet - will need followup pr's for other sections of the code)

### DIFF
--- a/freqtrade/configuration/create_datadir.py
+++ b/freqtrade/configuration/create_datadir.py
@@ -1,18 +1,20 @@
 import logging
-import os
 from typing import Any, Dict, Optional
+from pathlib import Path
 
 
 logger = logging.getLogger(__name__)
 
 
 def create_datadir(config: Dict[str, Any], datadir: Optional[str] = None) -> str:
+
+    folder = Path(datadir) if datadir else Path('user_data/data')
     if not datadir:
         # set datadir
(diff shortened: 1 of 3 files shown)


xmatthias:
i'll do PR's which will "intensify" this behaviour to be honest.

Currently, we have no way of using another user_dir directory - basically forcing users to check out the code and run it from **exactly** this directory.
Ideally, we can push freqtrade to pypi and use `pip install freqtrade`, it'll create a user_data directory "somewhere" and uses that (by passing in `--user-dir=whatever`).
 
It's still a long way to go (there's more than 1 piece missing for this to happen) - but one can dream, no?

hroff-1902:
> i'll do PR's which will "intensify" this behaviour.
> 
> Currently, we have no way of using another user_dir directory - basically forcing users to check out the code and run it from **exactly** this directory.
> Ideally, we can push freqtrade to pypi and use `pip install freqtrade`, it'll create a user_data directory "somewhere" and uses that (by passing in `--user-dir=whatever`).
> 
> It's still a long way to go (there's more than 1 piece missing for this to happen) - but one can dream, no?

Excellent!

PR #2033, opened 2019-07-18
remove deprecated --dynamic-whitelist option

hroff-1902:
Removal of the `--dynamic-whitelist` option, which left in the deprecated status for long time.

That was not so easy (many appearances), so verify that I did not forget anything.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -44,8 +44,8 @@ Mandatory Parameters are marked as **Required**.
 | `exchange.sandbox` | false | Use the 'sandbox' version of the exchange, where the exchange provides a sandbox for risk-free integration. See [here](sandbox-testing.md) in more details.
 | `exchange.key` | '' | API key to use for the exchange. Only required when you are in production mode.
 | `exchange.secret` | '' | API secret to use for the exchange. Only required when you are in production mode.
-| `exchange.pair_whitelist` | [] | List of currency to use by the bot. Can be overrided with `--dynamic-whitelist` param.
-| `exchange.pair_blacklist` | [] | List of currency the bot must avoid. Useful when using `--dynamic-whitelist` param.
+| `exchange.pair_whitelist` | [] | List of currency to use by the bot.
+| `exchange.pair_blacklist` | [] | List of currency the bot must avoid.
 | `exchange.ccxt_config` | None | Additional CCXT parameters passed to the regular ccxt instance. Parameters may differ from exchange to exchange and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
 | `exchange.ccxt_async_config` | None | Additional CCXT parameters passed to the async ccxt instance. Parameters may differ from exchange to exchange  and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
 | `exchange.markets_refresh_interval` | 60 | The interval in minutes in which markets are reloaded.
@@ -53,8 +53,8 @@ Mandatory Parameters are marked as **Required**.
 | `experimental.use_sell_signal` | false | Use your sell strategy in addition of the `minimal_roi`. [Strategy Override](#parameters-in-the-strategy).
 | `experimental.sell_profit_only` | false | Waits until you have made a positive profit before taking a sell decision. [Strategy Override](#parameters-in-the-strategy).
 | `experimental.ignore_roi_if_buy_signal` | false | Does not sell if the buy-signal is still active. Takes preference over `minimal_roi` and `use_sell_signal`. [Strategy Override](#parameters-in-the-strategy).
-| `pairlist.method` | StaticPairList | Use Static whitelist. [More information below](#dynamic-pairlists).
-| `pairlist.config` | None | Additional configuration for dynamic pairlists. [More information below](#dynamic-pairlists).
+| `pairlist.method` | StaticPairList | Use static or dynamic volume-based pairlist.
+| `pairlist.config` | None | Additional configuration for dynamic pairlists.
 | `telegram.enabled` | true | **Required.** Enable or not the usage of Telegram.
 | `telegram.token` | token | Your Telegram bot token. Only required if `telegram.enabled` is `true`.
 | `telegram.chat_id` | chat_id | Your personal Telegram account id. Only required if `telegram.enabled` is `true`.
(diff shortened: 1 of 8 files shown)


hroff-1902:
Amongst other changes, I removed "If enabled, it overrides the dynamic whitelist option." in docs/edge.md, since it seems for me incorrect and confusing. Edge does overrides whitelist options, it modifies the whitelist by itself, right?

hroff-1902:
In deprecated.md I stated "This command line option was deprecated in 2018 and removed freqtrade 2019.6-dev (develop branch) and in freqtrade 2019.7 (master branch)."

Version from the master branch should be adjusted during release process.

docs/configuration.md
```diff
@@ -44,17 +44,17 @@ Mandatory Parameters are marked as **Required**.
...
 | `exchange.markets_refresh_interval` | 60 | The interval in minutes in which markets are reloaded.
 | `edge` | false | Please refer to [edge configuration document](edge.md) for detailed explanation.
 | `experimental.use_sell_signal` | false | Use your sell strategy in addition of the `minimal_roi`. [Strategy Override](#parameters-in-the-strategy).
 | `experimental.sell_profit_only` | false | Waits until you have made a positive profit before taking a sell decision. [Strategy Override](#parameters-in-the-strategy).
 | `experimental.ignore_roi_if_buy_signal` | false | Does not sell if the buy-signal is still active. Takes preference over `minimal_roi` and `use_sell_signal`. [Strategy Override](#parameters-in-the-strategy).
-| `pairlist.method` | StaticPairList | Use Static whitelist. [More information below](#dynamic-pairlists).
-| `pairlist.config` | None | Additional configuration for dynamic pairlists. [More information below](#dynamic-pairlists).
+| `pairlist.method` | StaticPairList | Use static or dynamic volume-based pairlist.
+| `pairlist.config` | None | Additional configuration for dynamic pairlists.
```
xmatthias: Any particular reason to remove the link to the documentation (which is still there ...?)

docs/configuration.md
```diff
@@ -44,17 +44,17 @@ Mandatory Parameters are marked as **Required**.
 | `exchange.sandbox` | false | Use the 'sandbox' version of the exchange, where the exchange provides a sandbox for risk-free integration. See [here](sandbox-testing.md) in more details.
 | `exchange.key` | '' | API key to use for the exchange. Only required when you are in production mode.
 | `exchange.secret` | '' | API secret to use for the exchange. Only required when you are in production mode.
-| `exchange.pair_whitelist` | [] | List of currency to use by the bot. Can be overrided with `--dynamic-whitelist` param.
-| `exchange.pair_blacklist` | [] | List of currency the bot must avoid. Useful when using `--dynamic-whitelist` param.
+| `exchange.pair_whitelist` | [] | List of currency to use by the bot.
```
xmatthias: I guess we should still mention that `pairlist.method=dynamic` will override this?

docs/configuration.md
```diff
@@ -381,7 +381,7 @@ section of the configuration.
   * It uses configuration from `exchange.pair_whitelist` and `exchange.pair_blacklist`.
 * `VolumePairList`
   * Formerly available as `--dynamic-whitelist [<number_assets>]`. This command line
-option is deprecated and should no longer be used.
+option was removed and can no longer be used.
```
xmatthias: i would just remove it from the documentation - no point in mentioning that we removed it (it'll only confuse new users)

docs/configuration.md
hroff-1902: Oops, I thought it was a link to a section with description of the deprecated option. Restored it.

docs/configuration.md
hroff-1902: Oops, I thought it was a link to a section with description of the deprecated option. Restored it.

docs/configuration.md
hroff-1902: I have a better idea I guess. See below or in a sep. issue describing the proposal (or a PR with implementation).

docs/configuration.md
hroff-1902: Agree, removed.

hroff-1902:
The idea regarding whitelist I mentioned above is to combine them: so that VolumePairList select pairs from the set of [Whitelist - Blacklist], not from all pairs available. So the white list won't be overriden by VolumePairList or ignored, but will be used as a starting point for this method.

Rationale: a user could implement selecting top 5 pairs by its volume from a set of 10 pairs she is interesting to trade. You may say that even currently she may specify all non-interesting pairs in the blacklist. Well, yes, but 1) she needs to follow exchange and add quickly all new shitcoins listed in the exchange, 2) if exchange has 437 pairs and a user wants to trade only 10 or 20 of them, the blacklist will be extremely long. With my proposal she will be able to specify the list on any manner, choosing the most convenient approach (and use VolumePairList method on the resulting [Whitelist - Blacklist].

This can be even more useful in the future with addition more dynamic pairlist methods.

xmatthias:
yes and no - it depends on the usecase (i may not want to blacklist any pairs) - so your approach generates some work to generate a whitelist i currently simply leave empty (Yes, that's how i do it ...)

Why not add a config to VolumePairList to "selectFromWhitelist":True to activate (or default to this) behaviour?

docs/configuration.md
xmatthias: apart from that (see below) - both this and the next line are oddly worded... (i've just seen that now)

we have pairs in this list, not currencies ... 
maybe 
`List of pairs to check for potential trades.`
`List of pairs to absolutely avoid.`
?

xmatthias:
well - it sounds error-prone (assume i mistype "pair_white1ist" ) ...

It'll produce unexpected results ... and i'm not sure i like it (it's not for this PR anyway ...) so i guess this should better be discussed in an issue (or the PR implementing this feature)...

docs/configuration.md
hroff-1902: Yes, right.

hroff-1902:
> it's not for this PR anyway ...) so i guess this should better be discussed in an issue (or the PR implementing this feature)...

Sure.

docs/configuration.md
hroff-1902: But not 'for potential trades' since whitelist is used in live trades as we know... ;-)

hroff-1902: Fixed, pls review.

PR #2034, opened 2019-07-18
minor: add -V alias for --version

hroff-1902:
`-V` is frequently used for apps as a shortcut for `--version`.

* bot-usage docs also updated with this change


hroff-1902:
Seems something is wrong with travis itself...

xmatthias:
This failure is present in another PR too - i'll have a look (probably tonight...) it's however not related to this change.

PR #2035, opened 2019-07-18
minor: cleanup Arguments

hroff-1902:
name non-public class attributes and methods with leading underscore.

Currently, `_parse_args()` is also used in the scripts, but this is only temporary till they moved to the subcommands.

### DIFF
--- a/freqtrade/tests/test_arguments.py
+++ b/freqtrade/tests/test_arguments.py
@@ -51,8 +51,8 @@ def test_parse_args_verbose() -> None:
 
 def test_common_scripts_options() -> None:
     arguments = Arguments(['-p', 'ETH/BTC'], '')
-    arguments.build_args(ARGS_DOWNLOADER)
-    args = arguments.parse_args()
+    arguments._build_args(ARGS_DOWNLOADER)
+    args = arguments._parse_args()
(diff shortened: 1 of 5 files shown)


hroff-1902:
@xmatthias pls review the line in the test where flake failed (I left '##' there). Is really needed to run parse_args() with no_default_config=True there? This option is only used in the download utility in the meantime, but in that test it's called with the arguments for the plot utility...

freqtrade/tests/test_arguments.py
```diff
@@ -196,17 +195,16 @@ def test_plot_dataframe_options() -> None:
         '--plot-limit', '30',
         '-p', 'UNITTEST/BTC',
     ]
-    arguments = Arguments(args, '')
-    arguments.build_args(ARGS_PLOT_DATAFRAME)
-    pargs = arguments.parse_args(True)
+    arguments = Arguments(args, '', no_default_config=True)  ## Why True?
```
xmatthias: honestly - i think for legacy reasons (there's no reason i can see to do this for plot scripts ...)

hroff-1902: removed it (rebased)

PR #2043, opened 2019-07-21
Combine/resolvers

xmatthias:
## Summary
Inspired by reviewing #2040 - i noticed that 90% of the code to load modules is actually identical, if we exclude / slightly refactor strategy.

This PR aims to reduce this code duplication, while keeping everything else functionally identical.

Note: The test for base64 strategy was adjusted because it did not actually load the test-strategy from base64, but from TestStrategy.py (the loaded filename was wrong ...).

## Quick changelog

- Extract `_load_object()` method which loads objects from multiple paths.

While this code could technically be merged to `_search_object` - i decided not to since that would result in a heavily nested (2 for loops, with multiple try/if statements).

### DIFF
--- a/freqtrade/resolvers/hyperopt_resolver.py
+++ b/freqtrade/resolvers/hyperopt_resolver.py
@@ -23,12 +23,11 @@ class HyperOptResolver(IResolver):
 
     __slots__ = ['hyperopt']
 
-    def __init__(self, config: Optional[Dict] = None) -> None:
+    def __init__(self, config: Dict) -> None:
         """
         Load the custom class from config parameter
         :param config: configuration dictionary or None
         """
-        config = config or {}
 
         # Verify the hyperopt is in the configuration, otherwise fallback to the default hyperopt
         hyperopt_name = config.get('hyperopt') or DEFAULT_HYPEROPT
--- a/freqtrade/resolvers/hyperopt_resolver.py
+++ b/freqtrade/resolvers/hyperopt_resolver.py
@@ -26,7 +26,7 @@ class HyperOptResolver(IResolver):
     def __init__(self, config: Dict) -> None:
         """
         Load the custom class from config parameter
-        :param config: configuration dictionary or None
+        :param config: configuration dictionary
         """
 
         # Verify the hyperopt is in the configuration, otherwise fallback to the default hyperopt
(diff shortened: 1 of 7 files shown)


freqtrade/resolvers/hyperopt_resolver.py
```diff
@@ -23,12 +23,11 @@ class HyperOptResolver(IResolver):
 
     __slots__ = ['hyperopt']
 
-    def __init__(self, config: Optional[Dict] = None) -> None:
+    def __init__(self, config: Dict) -> None:
         """
         Load the custom class from config parameter
         :param config: configuration dictionary or None
```
hroff-1902: ':param config: configuration dictionary'

freqtrade/resolvers/iresolver.py
```diff
@@ -64,3 +64,26 @@ def _search_object(directory: Path, object_type, object_name: str,
             if obj:
                 return (obj(**kwargs), module_path)
         return (None, None)
+
+    @staticmethod
+    def _load_object(paths: List[Path], object_type, object_name: str,
+                     kwargs: dict = {}) -> Union[Any, None]:
```
hroff-1902: Should not be the type of returned value be `Optional[Any]`?

hroff-1902:
Take pls also look at the changes I made in #2040 in `freqtrade/tests/strategy/test_strategy.py`.

Is there any reason to have `test_strategy.py` in `freqtrade/tests/strategy`? And other files in that dir. Should not the tests use files from `user_data/strategies` and `freqtrade/strategy` as the main bot does?

xmatthias:
`test_strategy.py` is a misleading name for this file - since it doesn't contain the "TestStrategy" class - but tests the "strategy" ... 


I added the `.parent` now to this pr too - however that still does not load DefaultStrategy from the correct file, but from __init__.py - which used to be a workaround for hyperopt (which is not needed anymore since we discovered `wrap_non_picklable_objects`.

hroff-1902:
> `test_strategy.py` is a misleading name for this file - since it doesn't contain the "TestStrategy" class - but tests the "strategy" ...

Ah, yeah, I don't know why I decided above that it's a copy of default_strategy.py...

xmatthias:
> Ah, yeah, I don't know why I decided above that it's a copy of default_strategy.py...

i thought the same (twice) ...

PR #2046, opened 2019-07-22
Pyup/fix update 07 22

xmatthias:
## Summary
seems like pyup missed one week ... therefore we've had this many updates ...

Adapt to changes introduced by latest updates:

### ccxt:
ccxt changes: always provide message for ccxt exceptions https://github.com/ccxt/ccxt/pull/5470.

### plotly
Plotly deprecated a ton of stuff with the upgrade to 4.0. 

Adding that commit will mean that 3.x versions of plotly will no longer work (so people are forced to upgrade) - however also without this commit plotly should work just fine for now.

### pandas
https://github.com/freqtrade/freqtrade/blob/60cf56e23582fd7a0e06db0ce080640670cef836/freqtrade/data/btanalysis.py#L64-L75

for pandas, the error is in this snippet.
0.25.0 generates the dates of type `datetime64[ns, UTC]`, while the previous version uses object for open_time and close_time.

checking for the type (`np.issubdtype(df2['open_time'].dtype, np.datetime64)`) does result in an error since numpy does not seem to know this type.

Removing the `astype()` works in both old and new pandas versions, but i doubt that's correct since for pandas 0.24.x, we'd be resampling objects, not datetimes.

## todo:

* [x] investigate pandas errors
* [x] ccxt changes
* [x] plotly deprecations

### DIFF
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -67,7 +67,6 @@ def evaluate_result_multi(results: pd.DataFrame, freq: str, max_open_trades: int
     dates = pd.Series(pd.concat(dates).values, name='date')
     df2 = pd.DataFrame(np.repeat(results.values, deltas, axis=0), columns=results.columns)
 
-    df2 = df2.astype(dtype={"open_time": "datetime64", "close_time": "datetime64"})
     df2 = pd.concat([dates, df2], axis=1)
     df2 = df2.set_index('date')
     df_final = df2.resample(freq)[['pair']].count()
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -15,9 +15,9 @@ logger = logging.getLogger(__name__)
 
 
 try:
-    from plotly import tools
+    from plotly.subplots import make_subplots
     from plotly.offline import plot
-    import plotly.graph_objs as go
+    import plotly.graph_objects as go
 except ImportError:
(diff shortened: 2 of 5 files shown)


xmatthias:
@hroff-1902, i would appreciate if you could give have a very close look at the failing pandas test (`test_backtest_multi_pair`).

Using debug, inspect the dtype of `open_time` and `close_time` with both the old and new pandas version.
It seems like the new version uses datetime by default, but i think not explicitly casting could create problems if the old pandas version is still used, even though the test is not failing ...?

hroff-1902:
In progress...

hroff-1902:
If you think it's ready -- don't wait for me (logic here is a Noumenon so I need longer time to investigate). I'll create an issue as a reminder for myself in this case...

xmatthias:
> If you think it's ready

I'm not certain (as mentioned above) and will wait for a 2nd opinion ... 

One thing i noticed in the meantime:  open_time and close_time are not used though, so the type may not matter (?)

hroff-1902:
okay, need to find some more time for it...
so please wait

hroff-1902:
Reviewed it (and diffs in versions of pandas)... I still do not understand the purpose of this function and all conversions implemented in it. I found this function is not used in the codebase and is intended to be called from a python notebook by the user manually, right? In this case, I do not see any problem with removing this conversion in the function to make it working with new pandas...

xmatthias:
yes, that's basically how it's used (and in tests ...).
it's only used to actually test  if max_open_trades is respected

PR #2048, opened 2019-07-23
minor: add OnlyProfitHyperOptLoss

hroff-1902:
Minor changes:

* Adds Hyperopt Loss Function that takes only profit into account.
May be useful for users , I use it in practice.

If `freqtrade/optimize` is not a proper place for it, I could move it into `user_data/hyperopts` dir or to the `freqtrade/freqtrade-strategies` project.

* Improves wording in comments in DefaultHyperOptLoss and SharpeHyperOptLoss

### DIFF
--- a/freqtrade/optimize/hyperopt_loss_onlyprofit.py
+++ b/freqtrade/optimize/hyperopt_loss_onlyprofit.py
@@ -12,8 +12,12 @@ from freqtrade.optimize.hyperopt import IHyperOptLoss
 # This is assumed to be expected avg profit * expected trade count.
 # For example, for 0.35% avg per trade (or 0.0035 as ratio) and 1100 trades,
 # expected max profit = 3.85
-# Check that the reported Σ% values do not exceed this!
-# Note, this is ratio. 3.85 stated above means 385Σ%.
+#
+# Note, this is ratio. 3.85 stated above means 385Σ%, 3.0 means 300Σ%.
+#
+# In this implementation it's only used in calculation of the resulting value
+# of the objective function as a normalization coefficient and does not
+# represent any limit for profits as in the Freqtrade legacy default loss function.
 EXPECTED_MAX_PROFIT = 3.0
 
 
@@ -31,4 +35,4 @@ class OnlyProfitHyperOptLoss(IHyperOptLoss):
         Objective function, returns smaller number for better results.
         """
         total_profit = results.profit_percent.sum()
-        return max(0, 1 - total_profit / EXPECTED_MAX_PROFIT)
+        return 1 - total_profit / EXPECTED_MAX_PROFIT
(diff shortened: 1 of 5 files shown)


freqtrade/optimize/hyperopt_loss_onlyprofit.py
```diff
@@ -0,0 +1,34 @@
...
+
+from freqtrade.optimize.hyperopt import IHyperOptLoss
+
+
+# This is assumed to be expected avg profit * expected trade count.
+# For example, for 0.35% avg per trade (or 0.0035 as ratio) and 1100 trades,
+# expected max profit = 3.85
+# Check that the reported Σ% values do not exceed this!
+# Note, this is ratio. 3.85 stated above means 385Σ%.
+EXPECTED_MAX_PROFIT = 3.0
```
xmatthias: is there a reason to cap profits at 300%? we could also set this to 100 or 10_000?

your formula should also work for higher numbers, and would "stop" optimizing then since it'll return 0 - however if possible i would prefer to optimize to "as high as possible" ... without needing to mess with this here ?

hroff-1902:
you are right, my bad, I have not noticed it before... (have not reached the limit in the hyperopts)

with `max(0, val)` the hyperopt will report only first epoch with profit exceeding the limit as the best one and all further epochs will not be noticed even if they produce much better profits (since the resulting value is limited to 0 and will not go to the negative values)...

hroff-1902:
Done, comments also changed to display better the use of this constant and the differences from the default loss function.

freqtrade/optimize/hyperopt_loss_onlyprofit.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# Note, this is ratio. 3.85 stated above means 385Σ%, 3.0 means 300Σ%.
+#
+# In this implementation it's only used in calculation of the resulting value
+# of the objective function as a normalization coefficient and does not
+# represent any limit for profits as in the Freqtrade legacy default loss function.
+EXPECTED_MAX_PROFIT = 3.0
+
+
+class OnlyProfitHyperOptLoss(IHyperOptLoss):
```
xmatthias: Sorry for not noticing earlier - but we want to add this to the hyperopt.md documentation - otherwise it'll be there,but noone will know about it and therefore it's unused code.

Also, maybe (not absolutely necessary) a simple test should be added testing this loss-function similar to sharpe-loss-function (just to make sure that it really moves in the correct direction with correct results ...)

hroff-1902: > Sorry for not noticing earlier - but we want to add this to the `hyperopt.md` documentation (close to the sharpe-section?) - otherwise it'll be there,but noone will know about it and therefore basically it's unused code.

Sure, forgotten about the docs. Fixed also some wording in the nearest text a bit...

> Also, maybe (not absolutely necessary) a simple test should be added testing this loss-function similar to sharpe-loss-function (just to make sure that it really moves in the correct direction with correct results ...)

Done, thanks! 
(have not known of it, I did not notice you added that test when been reviewing it...)

Issue #2041, opened 2019-07-20
Install with Conda TA-Lib conflict with python 3.6


jraviotta:
I'm working on a method to install freqtrade with Conda as that is my preferred environment/package management tool.

TA-Lib is a bottleneck but quantopian supports a package on [conda-forge](https://anaconda.org/Quantopian/ta-lib). Unfortunately it is pinned to python <3.6.
Freqtrade requires python >=3.6 but is able to install TA-lib manually in `setup.sh`. I can do the same in my `Makefile` but the version conflict suggests that there may be some unexpected bugs introduced by end-running the version check.

Can you offer any clarity about the freqtrade version requirements? Is there any way to reduce the dependence to python <3.6?

xmatthias:
freqtrade uses typehints almost everywhere which were introduced in 3.6 - so that is definitely not an option unless you want to manually go through every line and remove them (which you'll need to repeat for every update). 

You should also be carefull with the conda-forge package - it's outdated - 0.4.9, while the current version on pypi is 0.4.17 - so AT LEAST 8 bugs (counting one bug per version number, but probably there were more) were fixed between then and now.

Also, using conda does not mean you cannot use pip - you can install some packages through conda, and others (like ta-lib) with pip (which is also documented for freqtrade as part of the [raspbian](https://www.freqtrade.io/en/latest/installation/#raspberry-pi-raspbian) installation).

installing ta-lib manually (the c dependency) and through pip is definitely your best solution. we cannot offer much support for problems with other install methods.

hroff-1902:
Even just theoretically, it's only possible to downgrade back to python 3.5 (previous versions are not supported already). And the end-of-life for python 3.5 is scheduled for 2019-09. Just an year... I guess it does not worth it.

jraviotta:
Thanks a bunch for the thorough response. It sounds like scripting a solution is my best bet. I'll also look into the raspbian documentation as you suggested. Since freqtrade is on pypi, I can include it in my conda environment.yml file, but the install fails because the TA-Lib c libraries are not installed and all the packaged options have dependency conflicts. Similarly, the TA-Lib wrapper on Pypi is easy to include in environment.yml, but expects the c libraries to be available. I think that if I can work through the TA-lib install, the rest should work. It's just so nice to be able to build everything from a spec list.

hroff-1902:
just to be correct:
end-of-life for python 3.5 is 2020-09, not 2019-09, I mistyped it above...

[...]

PR #2049, opened 2019-07-23
Conda / makefile

jraviotta:
## Summary
Enables building app in a conda environment or Docker container using a Makefile

Solve the issue: #2041 

## Quick changelog

- Added Makefile
- Updated .dockerignore/.gitignore
- Created new logic for generation of startup files
- Added .env file to separate configuration from strategies
- Added support for version control of a repo located at `user_data/user_repo` so users can keep custom strategies, scripts, and jupyter notebooks backed up

## What's new?
This is a proposed alternative to the installation process intended to:

- Streamline the installation options
- Make working with Docker containers less cumbersome and
- Add support for installation in a conda environment. 

The existing `setup.sh` script is untouched and will continue to work as a parallel installation option.

### Configuration

- Overwrite configuration defaults by copying `.env.example` to `~/.freqtrade/.env`, `user_data/.env` or `user_data/user_repo.env`. 
- Execute `make` from the project root to view available commands
- Build conda app with `make conda-app`
- Build docker app with `make docker-app`

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,58 @@
+name: freqtrade
+channels:
+  - defaults
+  - conda-forge
+dependencies:
+  # Required for app
+  - python>=3.6
+  - pip
+  - wheel
+  - numpy
+  - pandas
+  - scipy
+  - SQLAlchemy
+  - scikit-learn
+  - arrow
+  - requests
+  - urllib3
+  - wrapt
+  - joblib
+  - jsonschema
+  - tabulate
+  - python-rapidjson
+  - filelock
+  - flask
+  - python-dotenv
+  - cachetools
+  - scikit-optimize
+  - python-telegram-bot
+  # Optional for plotting
+  - plotly
+  # Optional for development
+  - flake8
+  - pytest
+  - pytest-mock
+  - pytest-asyncio
+  - pytest-cov
+  - coveralls
+  - mypy
+  # Useful for jupyter
+  - ipykernel
+  - isort
+  - yapf
+  - pip:
+    # Required for app
+    - cython
+    - coinmarketcap
+    - ccxt
+    - TA-Lib
+    - py_find_1st
+    - sdnotify
+    # Optional for develpment
+    - flake8-tidy-imports
+    - flake8-type-annotations
+    - pytest-random-order
+    - -e .
+
+
+
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -219,6 +219,17 @@ as the watchdog.
 
 ------
 
+## Using Conda
+
+Freqtrade can also be installed using Anaconda (or Miniconda).
+
+``` bash
+conda env create -f environment.yml
+```
+
+!!! Note:
+    This requires the [ta-lib](#1-install-ta-lib) C-library to be installed first.
+
 ## Windows
 
 We recommend that Windows users use [Docker](docker.md) as this will work much easier and smoother (also more secure).


environment.yml
```diff
@@ -0,0 +1,57 @@
...
+  - scikit-optimize=0.5.2
+  - python-telegram-bot=11.1.0
+  # Optional for plotting
+  - plotly=3.10.0
+  # Optional for development
+  - flake8=3.7.7
+  - pytest=5.0.0
+  - pytest-mock=1.10.4
+  - pytest-asyncio=0.10.0
+  - pytest-cov=2.7.1
```
xmatthias: these pinned dependencies could become a problem.
we use pyup to update dependencies, which cannot update a environment.yml file (https://github.com/pyupio/pyup/issues/322).

manually updating these will be problematic i think.

dev
```diff
@@ -0,0 +1,6 @@
+"docker logs" requires exactly 1 argument.
```
xmatthias: This file seems to be non-needed and appears to be output from a docker-command piped wrongly maye?

.env.example
```diff
@@ -0,0 +1,29 @@
...
+STRATEGY_DIR=
+# App Dockerfile. Defaults to Dockerfile
+APP_DOCKERFILE=
+# Default strategy to use with docker commands. Defaults to TestStrategy
+DEFAULT_STRATEGY=
+# Extra arguments to use when starting the app with Docker.
+# Defaults to -s $(DEFAULT_STRATEGY) --strategy-path $(STRATEGY_DIR)
+START_ARGS=
+# Exchange keys in the format EXCHANGE_KEY=123abc
+BINANCE_KEY=
```
xmatthias: I don't think these keys are actually read from the bot ... so i guess they are a bit misleading?

dev
jraviotta: `make docker-logs` shows the logs from within a running container. Since no containers are running during the test, the command fails.

.env.example
jraviotta: Yes that is correct. I'll remove them from .env.example. If you like the idea however, I can open a new issue to save exchange credentials in an environment variable.
Ideally, I would like to version control my files, but without my exchange credentials. I would prefer to keep those in ~/.freqtrade rather than in with the code where I might accidentally commit it to the repo.
 
It would be nice to disambiguate the 3 types of configurations to offer more flexibility. 

1. secrets (exchange credentials)
2. bot configurations (conda, venv, docker, rpi)
3. strategy configurations (EG RSI_no_stoploss_1min)

hroff-1902: > I would prefer to keep those in ~/.freqtrade rather than in with the code where I might accidentally commit it to the repo.

Nice point btw... 👍

environment.yml
jraviotta: Agreed. Conda does a good job of working out dependencies and I don't typically include any versions. But there was a conflict it missed that caused the install to fail probably because there are so many that have to be installed with pip. I'll try to narrow that down. If it's just one or two packages, `environment.yml` can likely stay the same until those few packages change.

.env.example
xmatthias: well you can do that with multiple --config options (the last pointing to `--config  ~/.secret_keys.json` ?  ...

jraviotta: I didn't see that in the docs. Like this?

```bash
freqtrade --config user_data/config.json --config ~/.freqtrade/.secret_keys.json --config user_data/user_repo/some_other_config.json
```

I did have trouble setting parameters in the strategies. The first time I messed with the project, I found that some configurations needed to be set in `config.json` even though the docs said they could be set in either place. Is that a known issue or should I try again and debug?

environment.yml
jraviotta: I removed the versions and it works.  I think pinning python fixed it.

dev
jraviotta: I gave each container type a unique name since there may be multiples running at once. EG freqtrade-backtest, freqtrade-bot, freqtrade-shell. That means that `make logs-freqtrade-backtest`, is just as annoying as `docker logs freqtrade-backtest` so I removed the target.

.env.example
jraviotta: I was able to use the multiple config option, so keys can be saved separately from config. The user can populate `START_ARGS` to save the configuration and pass it to Docker. Instead of mounting `config.json`, I mount `$(CONFIG_PATH):user_config`. That provides any config files the user has saved there and the mount of USER_DATA provides access to any config files saved somewhere in that directory.

I made the default `-s $(DEFAULT_STRATEGY) --strategy-path $(STRATEGY_DIR) -c user_config/config.json` which can be overridden in .env. as I have done here:

```bash
# Extra arguments to use when starting the app with Docker.
# Defaults to  -s $(DEFAULT_STRATEGY) --strategy-path $(STRATEGY_DIR) \
#		        -c user_config/config.json
START_ARGS=-c user_config/exchange-config.json \
    -c user_data/user_repo/strategies/new-strategy-config.json \
    -s $(DEFAULT_STRATEGY) --strategy-path $(STRATEGY_DIR)
```

Now I can put secrets in `~/.freqtrade/exchange-config.json` as

```bash
{   "exchange": {
        "name": "binance",
        "key": "your_exchange_key",
        "secret": "your_exchange_secret",
        "ccxt_config": {"enableRateLimit": true},
        "ccxt_async_config": {
            "enableRateLimit": true,
            "rateLimit": 200
        }
    },
    "telegram": {
        "enabled": false,
        "chat_id": "your_telegram_chat_id"
    }
}
```

and I can put strategy-type config in version control in my strategies repo as:

```bash
{
    "max_open_trades": 3,
    "stake_currency": "BTC",
    "stake_amount": 0.0005,
    "ticker_interval" : "1m",
    "exchange": {
        "name": "binance",
        "pair_whitelist": [
            "ETH/BTC",
            "EOS/BTC",
        ],
        "pair_blacklist": [
            "BNB/BTC"
        ]
    },
...

}

```
Thanks for making that possible.

This makes sense to me, but it's getting pretty far from the current installation and configuration strategies. I'll let you decide if this is helpful or just additional complexity that may confuse users. Feel free to quash the idea if I'm out in left field with the Makefile.

dev
xmatthias: I think you miss-read my comment - i think the file "dev" is not necessary as it only contains the output of `docker logs`.

jraviotta: Ah yes. I see. I deleted the file.

Makefile
```diff
@@ -0,0 +1,335 @@
...
+	@docker ps $($(APP_IMG_FILTER))
+
+## Delete all compiled Python files
+pyclean:
+	find . -type f -name "*.py[co]" -delete
+	find . -type d -name "__pycache__" -delete
+
+## Lint using flake8
+pylint:
+	flake8 src
```
xmatthias: This is incorrect (we don't have a `src` folder).

correct would probably be:
```
flake8 freqtrade scripts
mypy freqtrade scripts
```

Makefile
```diff
@@ -0,0 +1,335 @@
...
+		echo "already exists"; \
+	fi
+$(USER_DIR): user-repo
+
+## Delete setup files ** ALL DATA WILL BE DELETED **
+setup-delete:
+	-rm $(CONFIG_PATH)/config.json 2> /dev/null
+	-rm $(CONFIG_PATH)/tradesv3.sqlite 2> /dev/null
+	-rm $(CONFIG_PATH)/tradesv3.dryrun.sqlite 2> /dev/null
+	-rm $(CONFIG_PATH)/localtime 2> /dev/null
```
xmatthias: why is a object "localtime" even created? 

this seems odd to me and i've never seen it (unless i had some docker-command messed up).

.env.example
```diff
@@ -0,0 +1,24 @@
+###################################################################
+# App parameters
```
xmatthias: I would make it clear that this is (currently?) only used when using the makefile

.env.example
xmatthias: The makefile makes kindof sense - but it's a rather longish makefile (for a introduction, anyway) - and i wanna make sure we understand it in it's entirely before merging it (which in the end, will mean that we'll need to support it). (That's why reviewing takes a long time).

Makefile
```diff
@@ -0,0 +1,335 @@
...
+else
+	@$(DOCKER_RUN_NIX) -d --name $(APP_CONTAINER)-backtest \
+		$(DOCKER_MOUNTS)\
+		$(APP_CONTAINER) $(START_ARGS) backtesting --timerange=20180131-\
+		--refresh-pairs-cached  --export trades \
+		--export-filename=user_data/backtest_data/backtest_$(DEFAULT_STRATEGY)_$(TS).json
+endif
+
+# Run backtest on live exchange
+backtest-live:
```
xmatthias: `backtest --live` does not what most people think it does.
see this comment for explanation
https://github.com/freqtrade/freqtrade/issues/2052#issuecomment-514528038

i would remove this option since it's now deprecated (and honestly, never made much sense).

Makefile
```diff
@@ -0,0 +1,335 @@
...
+		$(APP_CONTAINER) -i
+else
+	@$(DOCKER_RUN_NIX) --name $(APP_CONTAINER)-shell --entrypoint=/bin/bash \
+		$(DOCKER_MOUNTS)\
+		$(APP_CONTAINER) -i
+endif
+
+# Run historical backtest
+TS := $(shell /bin/date "+%Y-%m-%d---%H-%M-%S")
+backtest:
```
xmatthias: I find this command surprising and would probably rename it to `docker-backtest`? 

most of the things above install freqtrade into a conda environment ... so i'd expect to run backtest may run in the conda environment too?

Makefile
```diff
@@ -0,0 +1,335 @@
...
+		$(DOCKER_MOUNTS)\
+		$(APP_CONTAINER) $(START_ARGS)
+else 
+	$(DOCKER_RUN_NIX) -d --name $(APP_CONTAINER)-bot \
+		$(DOCKER_MOUNTS)\
+		$(APP_CONTAINER) $(START_ARGS)
+endif
+
+## open a shell in the container
+docker-shell:
```
xmatthias: Based on the above comment, i would assume i'm dropped into the running container.
unless i miss something, then this creates a new container with a shell ... ?

Makefile
```diff
@@ -0,0 +1,335 @@
...
+# Detect operating system in Makefile.
+OSFLAG :=
+	UNAME_S := $(shell uname -s)
+	ifeq ($(UNAME_S),Linux)
+		OSFLAG += LINUX
+	endif
+	ifeq ($(UNAME_S),Darwin)
+		OSFLAG += OSX
+	endif
+	ifeq ($(UNAME_S),Windows*)
```
xmatthias: I am not certain that this makefile works for windows...

many commands won't work (`find` for example), but also the `make ta-lib` won't work.
i'm not that familiar with makefiles to determine if they should check that for each command or not

jraviotta:
> To be honest, I'm undecided on the makefile. (the conda file looks good, and should simplify installation on certain environments).

I can understand. I'm not convinced that this is the best approach either. It is what I have used on other projects to solve the problem of verbose Docker commands. In this case, I'm cramming in other functionality that is not really within the domain of a typical `Makefile`. Perhaps a better approach would be to start with user stories and then work backwards to develop the best feature to accomplish that requirement.

For example, the problem I wanted to solve could be characterized as such:

```
As a freqtrade user, I want to install the product in my preferred development environment with 
minimal configuration so that I can get up and running quickly.

prerequisite: virtual env
scenario: install on Mac
scenario: install on Windows
scenario: install on WSL
scenario: install on Linux

prerequisite: conda
scenario: install on Mac
scenario: install on Windows
scenario: install on WSL
scenario: install on Linux

prerequisite: docker
scenario: install on Mac
scenario: install on Windows
scenario: install on WSL
scenario: install on Linux

prerequisite: RaspberryPi
scenario: install anything that works
```

A second feature would address docker complexity

```
As a freqtrade user, given that I am primarily interacting with the bot through docker, 
I want to execute simple commands with minimal arguments so that I don't have to remember 
long complex strings.
```

1. Are those stories really issues for users or just for me?
2. Do they capture the fundamental problems or need to be expanded or revised?
3. How might you suggest tackling the problem?

xmatthias:
> As a freqtrade user, I want to install the product in my preferred development environment with 
minimal configuration so that I can get up and running quickly.

well there's 2 contradicting statements in this sentence to me - you're either a user - wanting an easy way to install - which hopefully is something around `pip install freqtrade` in the future,
OR you want it in your favorite development environment (because you're a developer) - so you'll have different requirements (like checking out code and such).
Also, as developer, you'll have personal preferences to install (me personally,i use pyenv), which you're familiar with - so either pip or conda could be wrong eventually, and satisfying everyone is not easily done in this regard (however being developers, everyone will have his own solutions).

**installation future**
we have problems with installation currently, most related to some dependencies (ta-lib, py_1st_find), which don't install easily, and others are the handling of user_data folders.
Once these problems are solved (or workarounds found), i expect to publish freqtrade to pypi, which will make installation as easy as `pip install freqtrade`.
Since users won't have any other files in this case (makefile, docker-compose file...) - which they would need to get through a different  method (probably from git) - that will make these options far less accessible, since it'll limit to the subsection of developers.


**regular users** 
For a regular user, the only "really" needed argument is `--config` - since every cli argument (argument, not comand!) can be added to the configuration file as well, so you'll only need `--config` (and mounts, in case of docker).

**docker**
for docker, using something like `alias dockerft=...` can be an option (so would `docker-compose.yml` be, where you can configure all mounts) - but honestly, I've not received complaints about the length of commands yet, so either everyone has his own method to handle this, or it's not a very common problem.

jraviotta:
> well there's 2 contradicting statements in this sentence to me - you're either a user - wanting an easy way to install - which hopefully is something around `pip install freqtrade` in the future,
> OR you want it in your favorite development environment (because you're a developer) - so you'll have different requirements (like checking out code and such).

You have much more experience with the freqtrade community, but I'm not convinced this is subdivision of users is entirely accurate. For example, a freqtrade standard user:

* May have a preference for virtualenv vs conda-env for additional post-hoc analyses such as ML processing of data.

* May have a jupyter notebook environment configured and may want to execute freqtrade directives from the notebook.

Those two use cases started my efforts to integrate freqtrade into my existing workflow. However, as you mention `pip install freqtrade` would satisfy those needs.

>honestly, I've not received complaints about the length of commands yet, so either everyone has his own method to handle this, or it's not a very common problem.

As you noted about developers, I think anyone working with docker already has some hack in place to deal with the long ugly commands. It was an issue for me, because I was trying to find a way to execute freqtrade running in a container from within a jupyter notebook. I understand that's probably an infrequent use case. I'm sure it's more common to start a container on server and let it run or to ssh into the container and execute commands from the inside.

It sounds like this PR should be closed and future effort should be directed to packaging freqtrade for pypi which would solve the most problems. Do you agree?

Makefile
```diff
@@ -0,0 +1,335 @@
...
+## Remove all containers built from app image
+docker-clean: docker-stop
+	@-$(APP_IMG_FILTER) | grep -q . && docker rm $$($(APP_IMG_FILTER))
+
+## Show status of Docker containers 
+docker-state:
+	@docker ps $($(APP_IMG_FILTER))
+
+## Delete all compiled Python files
+pyclean:
```
hroff-1902: Just a note; I personally use `py.cleanup -p` (from pycmd) for this, but I don't think this additional dependency should be added into requirements-dev.txt...

hroff-1902:
```
"It's a long time since I touched a draughtsman"
```
-- _from "Dead Souls"—A Poem by Nikolay Gogol_

I mean Makefiles...

I see long ugly fences of sed and awk options and vars... All just like in old lamp times twenty years ago... 😆 

Who will support it?

jraviotta:
I stripped everything but `environment.yml` which sounded like it might be useful.

xmatthias:
> I was trying to find a way to execute freqtrade running in a container from within a jupyter notebook

i did not succeed with this due to async causing some errors (but then i didn't try hard).
i don't think freqtrade can be executed in it's entirety in a notebook currently - however parts (mainly the analytical functions) can.

This looks cleaner now :+1:  (it's not that we're per se against makefiles, but maintaining it would have been difficult i think)

hroff-1902:
@jraviotta Please understand me correctly, I'm also not against Makefiles, not at all. But my point here was that this new stuff is declared above as 'alternative installation procedure', no more, no less, therefore it should be completely supported by the team if accepted, and this all-in-one file, with lot's of additional functionality for devs, can be hard for maintenance...

jraviotta:
>Please understand me correctly, 

Absolutely, no hard feelings at all. I contribute to open source because it buys me free consultation time with really smart engineers. I appreciate the feedback and am happy to help move the project forward in whatever way is useful. Let me know how you might approach the problems raised.

hroff-1902:
Sorry, I know absolutely nothing about conda and never used it, so I cannot help with #2041 .
(same for Jupyter and so called notebook, if I remember properly the names).

xmatthias:
I am currently testing this PR, but i've found some problems (not necessarily related to the PR ...)

Oddly enough, when i'm creating a new environment, it fails installing py-find-1st (that's a known problem: https://github.com/roebel/py_find_1st/issues/5 ) .

However, i do struggle to install py_find_1st in a conda environment once numpy is installed (while it's working in a normal virtualenv).

what i'm using for a regular virtualenv
``` bash
pip install numpy
pip install py_find_1st
```

with virtualenv, the above installs without problems, but conda throws gcc errors (even though i'm only using the conda empty environment, not even conda for installation.

``` bash
conda create -n conda_test3 python=3.7
conda activate conda_test3

pip install numpy
pip install py_find_1st
```

errors thrown:
```
Collecting py_find_1st
  Using cached https://files.pythonhosted.org/packages/16/20/813b0578790487874bb58d92591c72e1b019edb3ef4439793815c034a39b/py_find_1st-1.1.3.tar.gz
Building wheels for collected packages: py-find-1st
  Building wheel for py-find-1st (setup.py) ... error
  ERROR: Complete output from command /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-9asedvle/py-find-1st/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-dwriiie0 --python-tag cp37:
  ERROR: running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.linux-x86_64-3.7
  creating build/lib.linux-x86_64-3.7/utils_find_1st
  copying utils_find_1st/__init__.py -> build/lib.linux-x86_64-3.7/utils_find_1st
  running build_ext
  check for clang compiler ... no
  building 'find_1st' extension
  creating build/temp.linux-x86_64-3.7
  creating build/temp.linux-x86_64-3.7/utils_find_1st
  gcc -pthread -B /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include -I/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/include/python3.7m -c utils_find_1st/find_1st.cpp -o build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o
  cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
  In file included from /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/ndarraytypes.h:1830,
                   from /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/ndarrayobject.h:12,
                   from /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/arrayobject.h:4,
                   from utils_find_1st/find_1st.cpp:4:
  /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:17:2: warning: #warning "Using deprecated NumPy API, disable it with " "#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-Wcpp]
     17 | #warning "Using deprecated NumPy API, disable it with " \
        |  ^~~~~~~
  g++ -pthread -shared -B /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat -L/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib -Wl,-rpath=/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib -Wl,--no-as-needed -Wl,--sysroot=/ build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o -o build/lib.linux-x86_64-3.7/utils_find_1st/find_1st.cpython-37m-x86_64-linux-gnu.so
  /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
  /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
  /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
  /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
  build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: file not recognized: file format not recognized
  collect2: error: ld returned 1 exit status
  error: command 'g++' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for py-find-1st
  Running setup.py clean for py-find-1st
Failed to build py-find-1st
Installing collected packages: py-find-1st
  Running setup.py install for py-find-1st ... error
    ERROR: Complete output from command /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-9asedvle/py-find-1st/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-o6vt67hr/install-record.txt --single-version-externally-managed --compile:
    ERROR: running install
    running build
    running build_py
    creating build
    creating build/lib.linux-x86_64-3.7
    creating build/lib.linux-x86_64-3.7/utils_find_1st
    copying utils_find_1st/__init__.py -> build/lib.linux-x86_64-3.7/utils_find_1st
    running build_ext
    check for clang compiler ... no
    building 'find_1st' extension
    creating build/temp.linux-x86_64-3.7
    creating build/temp.linux-x86_64-3.7/utils_find_1st
    gcc -pthread -B /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -I/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include -I/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/include/python3.7m -c utils_find_1st/find_1st.cpp -o build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o
    cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
    In file included from /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/ndarraytypes.h:1830,
                     from /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/ndarrayobject.h:12,
                     from /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/arrayobject.h:4,
                     from utils_find_1st/find_1st.cpp:4:
    /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib/python3.7/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:17:2: warning: #warning "Using deprecated NumPy API, disable it with " "#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-Wcpp]
       17 | #warning "Using deprecated NumPy API, disable it with " \
          |  ^~~~~~~
    g++ -pthread -shared -B /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat -L/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib -Wl,-rpath=/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/lib -Wl,--no-as-needed -Wl,--sysroot=/ build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o -o build/lib.linux-x86_64-3.7/utils_find_1st/find_1st.cpython-37m-x86_64-linux-gnu.so
    /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
    /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
    /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
    /home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/compiler_compat/ld: build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: unable to initialize decompress status for section .debug_info
    build/temp.linux-x86_64-3.7/utils_find_1st/find_1st.o: file not recognized: file format not recognized
    collect2: error: ld returned 1 exit status
    error: command 'g++' failed with exit status 1
    ----------------------------------------
ERROR: Command "/home/xmatt/.pyenv/versions/miniconda3-latest/envs/conda_test3/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-9asedvle/py-find-1st/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-o6vt67hr/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-install-9asedvle/py-find-1st/

```

jraviotta:
Using a mac, wsl, windows or linux?

I had trouble with the C compiler on both my mac and WSL setup. This is from my Debian system on WSL.

```
g++/now 4:6.3.0-4 amd64 [installed,local]
g++-6/now 6.3.0-18+deb9u1 amd64 [installed,local]
gcc/now 4:6.3.0-4 amd64 [installed,local]
gcc-6/now 6.3.0-18+deb9u1 amd64 [installed,local]
gcc-6-base/now 6.3.0-18+deb9u1 amd64 [installed,local]
libgcc-6-dev/now 6.3.0-18+deb9u1 amd64 [installed,local]
libgcc1/now 1:6.3.0-18+deb9u1 amd64 [installed,local]
linux-compiler-gcc-6-x86/now 4.9.168-1+deb9u3 amd64 [installed,local]
```
I'll report from my mac later.

jraviotta:
from my mac

```
$ gcc -v
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

```

See https://stackoverflow.com/questions/52509602/cant-compile-c-program-on-a-mac-after-upgrade-to-mojave

And https://docs.brew.sh/C++-Standard-Libraries

And https://askubuntu.com/questions/758962/problem-to-compile-python-and-c-code-using-ubuntu-ubuntu-12-04-5-lts-problem

xmatthias:
it's "pure" Linux for me ...

what i don't understand is that it's working fine when the venv is not a conda environment, and it's not working when using conda - since gcc shouldn't change (unless conda messes with that - which would be a huge bug in conda in my opinion ...)

jraviotta:
I assume you updated conda
`conda update -all`
And cleared the cache
`conda clean -all`

Do you have conda installed in a venv? 

I remember solving another problem awhile ago where python was being appended to the path rather than prepended to the path. Consequently conda was using the system python instead of miniconda.

Try activating the env and checking python

```
conda deactivate
which python
conda environment create -n test
conda activate test
which python

```

jraviotta:
Also, are you installing anything manually? I know `setup.sh` assumes numpy needs to be installed first. 

Perhaps try

```
conda deactivate
conda environment create -f environment.yml

```

xmatthias:
> I know setup.sh assumes numpy needs to be installed first.

this was the case, but the reason was py_find_1st - which is now fixed (as of version 1.1.4), thanks @roebel for the great support!

I'll try my conda luck on another pc tomorrow - hopefully conda is not broken there (i normally don't use conda, since it does provide little benefit when on windows (however it's my main python on windows).

PR #2057, opened 2019-07-25
Move argument definitions to their own file

xmatthias:
## Summary
Move definition of Arguments to a seperate file (arguments.py is very cluttered).

part of  the issue: #1945

## Quick changelog

- move argument definitions to `cli_arguments.py`

I'm unsure if the argument selections (assignment of arguments to the different commands) should also move to this new file, or if they should stay in arguments.py...


hroff-1902:
> I'm unsure if the argument selections (assignment of arguments to the different commands) should also move to this new file, or if they should stay in arguments.py...

Don't think in the meantime they should be moved, let's leave them in arguments.py for a while...

I've also have a postponed refactoring commit to arguments/configuration, a set of my previous commits to arguments/configuration were cherry picked minor cleanups from that main one -- I decided to split in order to make the main idea clear and observable...

xmatthias:
> Besides import error

thank you :+1:

PR #2058, opened 2019-07-25
minor: cleanup in strategy interface

hroff-1902:
* underscore added to the name of `analyze_ticker()` since it's not a public method.

* eliminate warnings in pytest when testing deprecated strategy interfaces.

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -154,7 +154,7 @@ class IStrategy(ABC):
         """
         return self.__class__.__name__
 
-    def analyze_ticker(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+    def _analyze_ticker(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
         """
         Parses the given ticker history and returns a populated DataFrame
         add several TA indicators and buy signal to it
@@ -198,7 +198,7 @@ class IStrategy(ABC):
             return False, False
 
         try:
-            dataframe = self.analyze_ticker(dataframe, {'pair': pair})
+            dataframe = self._analyze_ticker(dataframe, {'pair': pair})
         except ValueError as error:
             logger.warning(
                 'Unable to analyze ticker for pair %s: %s',
--- a/freqtrade/tests/strategy/test_strategy.py
+++ b/freqtrade/tests/strategy/test_strategy.py
@@ -366,6 +366,7 @@ def test_strategy_override_use_sell_profit_only(caplog):
             ) in caplog.record_tuples
 
 
+@pytest.mark.filterwarnings("ignore:deprecated")
 def test_deprecate_populate_indicators(result):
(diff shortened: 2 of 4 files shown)


freqtrade/strategy/interface.py
```diff
@@ -154,7 +154,7 @@ def get_strategy_name(self) -> str:
         """
         return self.__class__.__name__
 
-    def analyze_ticker(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+    def _analyze_ticker(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
```
xmatthias: I'm not certain about this... 
while it is private in it's current state, it's basically a copy of `generate_dataframe` in `plot_dataframes.py` after some refactoring, so it may be used this from that script (probably not a script anymore by that time ...).


Also, i'd like to extend the "data analysis" documentation part on analyzing strategies (you CAN import strategies and data in jupyter notebooks and analyze the outputs to verify signals are as you'd expect ... it's just not documented - which i'll try to do as part of the plot-scripts removal since that'll define the plotting interface for this) - which would require this method.

freqtrade/tests/strategy/test_strategy.py
```diff
@@ -366,6 +366,7 @@ def test_strategy_override_use_sell_profit_only(caplog):
             ) in caplog.record_tuples
 
 
+@pytest.mark.filterwarnings("ignore:deprecated")
```
xmatthias: this one is gold :+1:

freqtrade/strategy/interface.py
hroff-1902: But why you don't wish to use public get_signal() everywhere? Because of the "if dataframe is out of date" checking? Add a parameter which would allow to switch this check off when running on historical data... The `get_signal()` contains many useful checks besides that one, and you'll repeat them in the code if you decide to call analyze_ticker() instead... In principle, the get_signal() should be used everywhere (dry/live run, backtesting, plotting) when signals are needed. In order to obtain same results and same behavior in diff. parts of freqtrade...

xmatthias: `get-signal` does something else entirely, and cannot be used for that.

it returns a tuple of (true, false), based on the last entry of the dataframe (needed for live/dry run - but not for anything else).

What backtesting, hyperopt, plotting, data analysis needs is a series of 1 / 0 values defining buy/sell signals for each candle, or rather the dataframe with buy/sell signals populated in columns.

The only potentially useful check for these usecases i can see at a quick glance is the first one - checking if the dataframe is empty.

The rest (actually there is only one more, which is checking for outdated ata) - but it's irrelevant for everything other than dry/live runs (i may backtest february of this year, where this check will always trigger).

hroff-1902:
Okay, agree. I decided this is of no public use looking at the codebase.
Cherry-picked the second commit, closing this in favor to #2062.

PR #2063, opened 2019-07-26
tests: don't mask numpy errors as warnings in tests

hroff-1902:
...cause these are tests, we need to see all errorneous cases.

This change produces an error in the rpc test -- that's its purpose.

I cannot fix that error because I do not understand why `numpy.mean()` is used in that place in rpc.py...

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -10,7 +10,7 @@ from typing import Dict, Any, List, Optional
 
 import arrow
 import sqlalchemy as sql
-from numpy import mean, nan_to_num, NAN
+from numpy import mean, NAN
 from pandas import DataFrame
 
 from freqtrade import TemporaryError, DependencyException
@@ -195,9 +195,9 @@ class RPC(object):
         trades = Trade.query.order_by(Trade.id).all()
 
         profit_all_coin = []
-        profit_all_percent = []
+        profit_all_perc = []
         profit_closed_coin = []
-        profit_closed_percent = []
+        profit_closed_perc = []
         durations = []
 
         for trade in trades:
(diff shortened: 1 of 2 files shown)


xmatthias:
well why not?
i don't understand why it's now throwing an error, a mean on an empty list (np.mean([]) should simply be 0?

maybe we need to catch this in the code, since i guess the same warning (it's originally a warning, not an error) will appear when calling `/profit`.

hroff-1902:
> well why not?

Because reaction on this in np can be different, issuing a warning is just one of the modes for handling this, in the tests it may be a warning while in the bot it may cause an exception...

And this may cause potential incorrect results, If np.mean() (in this case) issues a warning and returns 0, I can find examples when the 1,0 value would be correct for the case of empty array (when it's used in multiplication, for example)... With warnings this will be hidden. We probably should consider changing the handling of numpy and pandas errors from issuing warnings to exceptions...

At least, in the tests, some of those cases (ideally as much as possible) should be covered and not ignored with simple warnings...

xmatthias:
The fix was easier than thought - the `nan_to_num()` function was there for this error - so by checking if the list is empty (instead of waiting for NAN to be returned), we avoid this error and the solution is cleaner too..

freqtrade/rpc/rpc.py
```diff
@@ -238,15 +238,16 @@ def _rpc_trade_statistics(
 
         # Prepare data to display
         profit_closed_coin_sum = round(sum(profit_closed_coin), 8)
-        profit_closed_percent = round(nan_to_num(mean(profit_closed_percent)) * 100, 2)
+        profit_closed_percent = (round(mean(profit_closed_perc) * 100, 2) if profit_closed_perc
+                                 else 0.0)
```
hroff-1902: The fix LGTM 👍

Issue #1384, opened 2018-12-02
Align usage of --data-dir 


xmatthias:
## problem
Currently different commands implement the same command (`--data-dir`, `--strategy-dir`) - which however needs to be pointed to different folders depending on the action taken.

### examples:
* download-script needs to be pointed to `user_data/data/<exchange>`

## expectation

This should be aligned (maybe with exception to `--strategy-dir` ) - to ensure a consistent user experience.

* Ideally, this should be pointed to  `user_data` - wherever this is located. 
* A script to create the default user_data structure "anywhere" should be provided to assist users getting started.
* Pointing to `user_data` will load data from `user_data/data/<exchange>`
* pointing to `user_data will load hyperopt files from `user_data/hyperopts`

this should ensure flexibility when more custom/pluggable components are added.

hroff-1902:
#1893 can be considered as a part of this (--export --> --datadir), right?

hroff-1902:
I the whole, I guess there should be a common setting like --userdata-dir (--user-data, or something) which by default points to the user_data dir. That's in addition to the --datadir option. If --datadir is not specified, its value is composed as <userdata_dir>/data/<exchange>. Same for other paths -- for strategies, hyperopt files, etc -- they should start from userdata_dir, which can be set in the command line options (if not overriden explicitely by other options, --datadir, --strategy-dir, maybe --hyperopt-dir etc.)

An env. variable FREQTRADE_USERDATA_DIR can also be fetched from the environment...

PR #2067, opened 2019-07-28
Align userdata usage

xmatthias:
## Summary

Closes: #1384

## Quick changelog

- Addition of new flag to all commands (`--userdir`, `--user-data-dir`)
- defaults to `Path.cwd()/user_data` (current path's user-data).
- allows configuring `user_data_dir` to use a different one
  - enables proper versioning of personal strategy / hyperopt files 
  - should enable installation wia pypi (where we're not in the code-folder, but "somewhere" else.
- addition of `freqtrade create-userdir` subcommand to initialize the folder.
- change default folder for backtest-results to `backtest_result` (backtest_data is misleading)
- adjust everything using `user_data` to use config['user_data_dir'] instead.


freqtrade/configuration/configuration.py
```diff
@@ -10,7 +10,7 @@
 
 from freqtrade import OperationalException, constants
 from freqtrade.configuration.check_exchange import check_exchange
-from freqtrade.configuration.create_datadir import create_datadir
+from freqtrade.configuration.folder_operations import create_datadir
```
hroff-1902: no folders in unixes... o_0

xmatthias: i always wonder how you notice this - i had to think 3 times (and took the wrong option), since for me directory and folder is interchangeable.

hroff-1902: I was writing file system drivers some long time ago...
Additionally, I know from the times of Windows 3.1 that the folder term is only used in Windows GUI, not in CDE and DCE or POSIX calls...

Actually, in Unixes, everything is file, even directories... 😆

docs/bot-usage.md
```diff
@@ -82,6 +87,25 @@ of your configuration in the project issues or in the Internet.
...
+This directory will look as follows:
+
+user_data/
+├── backtest_results
+├── data
+├── hyperopts
+├── plot
+└── strategies
+
+You can add the entry "user_data_dir" to your configuration, to always point your bot to this folder.
```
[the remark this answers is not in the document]
hroff-1902: No folders... 🤣 
Pls, run through the "changes in all commits" for "folder/-s"...

hroff-1902:
Some generic points:

* I don't thing that pairlist resolver should search in userdata subdirectory -- that's for user's custom content, results and working files.
* The `--plot-dir` and `--backtesting-results-dir` (or `--results-dir`?) should also be added for completeness. The `--plot-dir` can be added later when plotting scripts become a subcommand.
* Probably the hyperopt trials file should be dumped to the `hyperopt-results` directory. Or to the 'results' directory common with backtesting. Pls consider these cases.
* I'm going to add the `--export-something` option to hyperopt soon, so the result should also be kept somewhere (`hyperopt-results`, `results` common with backtesting?)

freqtrade/configuration/configuration.py
```diff
@@ -170,9 +168,19 @@ def _process_common_options(self, config: Dict[str, Any]) -> None:
...
+        --user-data, --datadir
         """
+        if 'user_data_dir' in self.args and self.args.user_data_dir:
+            config.update({'user_data_dir': self.args.user_data_dir})
+        elif 'user_data_dir' not in config:
+            # Default to cwd/user_data (legacy option ...)
+            config.update({'user_data_dir': str(Path.cwd() / "user_data")})
+
+        # reset to user_data_dir so this contains the absolute path.
+        config['user_data_dir'] = create_userdata_dir(config['user_data_dir'])
```
hroff-1902: Ah, a more serious note: the user dir with its layout (subdirs) should not be created just because I mentioned it in the cli option. You can check its existence here, but creation should only be made by explicit call to the create-userdir command...

docs/bot-usage.md
```diff
@@ -82,6 +87,25 @@ of your configuration in the project issues or in the Internet.
...
+
+Freqtrade allows the creation of a user-data directory using `freqtrade create-userdir --userdir someDirectory`.
+This directory will look as follows:
+
+user_data/
+├── backtest_results
+├── data
+├── hyperopts
+├── plot
+└── strategies
```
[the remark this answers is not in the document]
xmatthias: nice catch :+1:

freqtrade/configuration/configuration.py
xmatthias: I am not certain about this - we might be facing "quite a few" problems if we change `user_data` to something else - since this has always been there.

You think we should fail startup if a folder is not present while checking and hae the user issue the `create-userdir` command explicitly?

won't users think "just create that folder" ?

xmatthias:
> I don't thing that pairlist resolver should search in userdata subdirectory -- that's for user's custom content, results and working files.

why not? it'll simplify development of new things, and pairlist can be something users may not want to share / keep private.

also, it did search the `user_data` path before: `Path.cwd().joinpath('user_data/pairlist'),` ...

> The --plot-dir and --backtesting-results-dir (or --results-dir?) should also be added for completeness. The --plot-dir can be added later when plotting scripts become a subcommand.

`--backtesting-results-dir` = `--export-filename` - so this is there.

I'm honestly not a fan of these `--*dir` options (but we didn't have `--user-data` as a combined option).

it makes diagnosing the real source of something quite difficult / obscure - and i hope using a "user_data" dir with different sub-dirs will suffice most users needs. Sometimes too much flexibility is not good... 

so currently, i can define `--user-data` to `.freqtrade/` - but override this for backtest data, strategy, hyperopt, ... to something else (where in the end, --user-data is passed in but not used anywhere because all options did override it somehow).
(hope you get what i mean).


`hyperopt-results` - good point.

freqtrade/configuration/configuration.py
hroff-1902: No, they won't,.

Exactly, yes -- the bot should check for existence of the dir and quit if it cannot find it. 

The dir path can simply be misspelled, it should not be created just after first run of the bot, the `create-userdir` command should be issued by the user explicitly. Same for the subdirs needed for the running mode. A user can find she misspelled the dir, change it and make another error in the name -- another dir will be created. This will quickly get the environment to complete mess...

So the behavior, imho, should be following: check for dir, same for necessary subdirs (can be done in the modules, not in the main startup configuration, but in the configuration as well). If run `create-userdir` -- create userdir and its subdirs. Take into account that subdir can point to another locations (overriden with corresponding `--*-dir` option or `*_dir` setting in the config,

xmatthias: > Take into account that subdir can point to another locations (overriden with corresponding --*-dir option or *_dir setting in the config),

which is something i would want to get rid off since there's no point IMHO.

Yes, modules create the dir if they need it (plot, for example).
I would not want to remove the dir-creation from all modules unless we remove the override possibility, and restrict users to the --user-dir directory.
Otherwise, we'd need to have a "create-userdir" - "create-plotdir", "create-...dir" command - which will become a complete mess and make 0 sense in my opinion, and will only create frustrations.

hroff-1902:
Sorry for delaying this -- I guess we need to discuss the behavior a bit more here. Maybe afternoon/tonight I'll find more time for this.

xmatthias:
that's fine, there's no rush (nothing is broken without this ...)

### some thoughts and reasonings:

The way i see how this should evolve (and why):

Until now, we've always required freqtrade to be cloned to disk, and to be run from this folder. This is problematic if we want an easier way to install.

The problem with the current user_data folder is that it is not part of a `pip install freqtrade` installation, since it'll be in a obscure location (this is one of the reasons to get rid of the scripts, too).

I would want to move the sample files to "somewhere" in the bot, and have `freqtrade create-datadir` copy/paste them into the user-directories as samples if they don't exist.

Users can then version their own user_data dir, and transferring between machines is easy (since it can be a (private) repository on github/bitbucket/gitlab, which you check out to your new machine, and you're almost ready to go.

`--xxx-path` options should only be used in extreme / seldom circumstances (I'm undecided if they should not be deprecated and removed ...) - they only create confusion as to where the "recommended" (_<- see - no typo this time_) location for stuff is.

with the above, i would see the following next steps:

* user_data should ideally be used from a OS Specific user location (probably .freqtrade or %appdata%/freqtrade)
* `--user-data something` can set the user-data directory to a different folder
* `--*-path` (strategy-, hyperopt-, ...) - these are additional paths the bot searches for strategies
* the default for `config.json`  is changed from `./config.json` to `[user-data-dir]/config.json` - moving configuration to the user-data directory as well.

* [maybe] deprecate `--xxx-path` options

hroff-1902:
Well, I supposed here to post long texts with my considerations, so could not find appropriate amount of time for this and appropriate mood... So sorry for delay...

Something in myself is against the proposals above, cannot formulate it explicitly in the meantime...

I'll try to express my considerations:
* From one point of view, I'm completely agree that a singe user-dir is that is needed for most of the users. And also want to declare all those `--xxx-path` options as deprecated. They can only create mess and confuse novices.
* In the same time as a supporter, I would like to have an ability to create a new single dir where I can place all the files I obtained from the user. But I would not like to fetch the historical data every time. So maybe multiple `--user-dir` should be allowed...
* In the same time, someone inside me is strictly against complete removal of an ability to specify diff. directories _for geeks_, for non-standard configurations.
* In case we still have all those `--xxx-path`, there is no need for many `create-xxx-dir` subcommand. A single `create-user-dir(s)` subcommand should create the user-dir and all the xxx-dirs, even if they are not subdirs.

So in short:


> * user_data should ideally be used from a OS Specific user location (probably .freqtrade or %appdata%/freqtrade)

So are you going to move from `(pwd)/user_data` to `(home)/.freqtrade/` by default? What about multiple instances running?

> * `--user-data something` can set the user-data directory to a different folder

Yes. Consider an ability to specify multiple `--user-data`

> * `--*-path` (strategy-, hyperopt-, ...) - these are additional paths the bot searches for strategies
> * [maybe] deprecate `--xxx-path` options

Let's deprecate them?

> * the default for `config.json`  is changed from `./config.json` to `[user-data-dir]/config.json` - moving configuration to the user-data directory as well.

Okay.

I'm thinking of (possibly undocumented) env. variables FREQTRADE_XXX_PATH that would allow to specify non-standard dirs for those who knows what he is doing... For 'normal' users there shall only be the `--user-dir/--user-data-dir` option.

xmatthias:
> Even shorter summary: let's make it as simple as possible. Only one --user-dir/--user-data-dir option, deprecate all other --xxx-paths. I'll try to live with this...

err what? you argued that the other options serve good purpose - and i agree, especially considering the "user-received files" argument.

I am against having undocumented features, if they are useful to us, they may serve others too (and if they don't they won't use them - so nothing bad comes out anyway).

> So are you going to move from (pwd)/user_data to (home)/.freqtrade/ by default? What about multiple instances of the bot running?

`(home)/.freqtrade` may be the default, but you can either override them with ` --user-dir (home)/.freqtrade2/` - or even use the same folder (using a combination of `--config` and `--db-url` ). This works now, and i don't see why it should not work in the future.

> [maybe] deprecate --xxx-path options

I'm not certain about this - and your comment did not feed confidence that it's a good idea (even though you seem to have changed your mind at the end of the comment). Having the flexibility won't hurt, and if we see confusion we may still deprecate them in the future.

I would leave this deprecation for the future (if at all) - once we see if --user-dir is working well or not (we can't really know until users start using it).

> there is no need for many create-xxx-dir subcommand

This was never a consideration since all other commands are simple ~~folder~~ directories.

hroff-1902:
First of all, sorry for delaying this.

> err what? you argued that the other options serve good purpose - and i agree, especially considering the "user-received files" argument.

I decided you wish to remove the --xxx-path options completely. Now I understand that you are going to mark them as deprecated for some limited (or longer, as it will be going) period of time. Am I right here?

> I am against having undocumented features, if they are useful to us, they may serve others too (and if they don't they won't use them - so nothing bad comes out anyway).

Okay, I only want to leave an ability to specify custom paths for all the subdirs. If it will be deprecated options that will still live (but 'hidden' (say, 'forbidden' for use) for an unexperienced user since deprecation) -- that's even better.

> `(home)/.freqtrade` may be the default, but you can either override them with ` --user-dir (home)/.freqtrade2/` - or even use the same folder (using a combination of `--config` and `--db-url` ). This works now, and i don't see why it should not work in the future.

That's okay for me.

> > [maybe] deprecate --xxx-path options
> 
> I'm not certain about this - and your comment did not feed confidence that it's a good idea (even though you seem to have changed your mind at the end of the comment). Having the flexibility won't hurt, and if we see confusion we may still deprecate them in the future.

So, deprecated but existing and kept, right?

> I would leave this deprecation for the future (if at all) - once we see if --user-dir is working well or not (we can't really know until users start using it).

Agree, okay.

> > there is no need for many create-xxx-dir subcommand
> 
> This was never a consideration since all other commands are simple ~folder~ directories.

Yes, but if `--xxx-path` options are used, the `create-userdir` subcommand should create those dirs in the designated locations, not as subdirs of the parent userdir. Probably. I actually not sure.

xmatthias:
Well my plan was to leave the `--xxx-path` options untouched (not deprecated) for now (and maybe forever), since you convinced me that they will be useful (be it for us - or for users).

There's no point in deprecating them if we don't plan to remove them.

> Yes, but if --xxx-path options are used, the create-userdir subcommand should create those dirs in the designated locations, not as subdirs of the parent userdir.

`create-userdir` does only accept `--user-data` as argument.
creating `--xxx-path` directories in "remote" places is kinda pointless in my opinion, since these are not directory structures, but simple, onelevel folders (and usually you'd use `--xxx-path` to specify an already existing strategy folder, no?)

> The second one -- I don't think that the pairlist resolver should search in the user-dir and that --pairlist-path should be exposed to the users in this case. Because we do not declare the interface to the custom pairlists in the docs.

that's a seperate point and a good catch ... we do currently search this folder already (enable debug and you'll probably see `"2019-08-21 07:12:52,417 - freqtrade.resolvers.iresolver - WARNING - Path "user_data/pairlist" does not exist."` - this is from current develop) and i think it makes sense to allow custom pairlists (maybe i have some special external list i wanna loop through) - or you're developing a new pairlist method which you can "drop" to some remote location to be able to keep updating using git easily ...?)

I would however leave this as a seperate point (adding this to the documentation) - as it's not really related to this PR.

hroff-1902:
> Well my plan was to leave the `--xxx-path` options untouched (not deprecated) for now (and maybe forever), since you convinced me that they will be useful (be it for us - or for users).
> 
> There's no point in deprecating them if we don't plan to remove them.

Okay.

> `create-userdir` does only accept `--user-data` as argument.
> creating `--xxx-path` directories in "remote" places is kinda pointless in my opinion, since these are not directory structures, but simple, onelevel folders (and usually you'd use `--xxx-path` to specify an already existing strategy folder, no?)

Okay, extra `--xxx-path` options to the `create-userdir` command will always be added in the future in case this will be needed (if/when). In this PR, let `create-userdir` create only subdirs.

> that's a seperate point and a good catch ... we do currently search this folder already (enable debug and you'll probably see `"2019-08-21 07:12:52,417 - freqtrade.resolvers.iresolver - WARNING - Path "user_data/pairlist" does not exist."` - this is from current develop) and i think it makes sense to allow custom pairlists (maybe i have some special external list i wanna loop through) - or you're developing a new pairlist method which you can "drop" to some remote location to be able to keep updating using git easily ...?)
> 
> I would however leave this as a seperate point (adding this to the documentation) - as it's not really related to this PR.

Okay. let's add description for this interface separately, sure. 

I assume you have some custom pairlists in the user_data... 😉 


====
I guess there is no unclear/non-coordinated points here anymore. 👍

xmatthias:
> I assume you have some custom pairlists in the user_data... 😉 


no, not yet :D

hroff-1902:
🍦

PR #2068, opened 2019-07-28
Eliminate using deprecated --live in tests

hroff-1902:
...and get rid of corresponding deprecation warnings in pytest.

Some additional checks verifying that --live IS NOT detected should be deleted later, when the option will be deleted, I guess. However, there is a chance that it will be forgotten. So I can delete them in this PR as well...

### DIFF
--- a/freqtrade/tests/optimize/test_backtesting.py
+++ b/freqtrade/tests/optimize/test_backtesting.py
@@ -831,8 +831,7 @@ def test_backtest_start_multi_strat(default_conf, mocker, caplog):
         '--datadir', 'freqtrade/tests/testdata',
         'backtesting',
         '--ticker-interval', '1m',
-        '--live',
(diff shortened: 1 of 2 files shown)


xmatthias:
I'm not certain about doing this now - even if it's deprecated, we should still make sure we're not regressing --live until we remove it?

regarding "not" tests: `not log_has()` are bad in my opinion, they'll assert true if there is a typo ... so i'm not sure how much that tests.
(`'live' not in config` on the other hand seems valid).

hroff-1902:
My idea was that we should not support deprecated functions. Including bugfixes and tests.
Maybe you are right. I should not have deleted the test that checks for presence of the deprecation warning, at least.

xmatthias:
i agree that we should not support that, but until we remove it, it should not break anyway, otherwise we might as well just remove it right away.

PR #2073, opened 2019-07-29
Improve setup.py to allow "extras" installations

xmatthias:
## Summary
This pr will allow installation via pip using "extras".
basically, to install the bot for a development installation:
`pip install -e .[all]`.

To only install the base bot + plot features: `pip install -e .[plot]`

This will become more power should we be able to sucessfully isntall through pypi - where a command like `pip install freqtrade[all]` will become possible.

## Quick changelog

- add extra-install tools

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -12,8 +12,8 @@ Special fields for the documentation (like Note boxes, ...) can be found [here](
 
 ## Developer setup
 
-To configure a development environment, use best use the `setup.sh` script and answer "y" when asked "Do you want to install dependencies for dev [y/N]? ".
-Alternatively (if your system is not supported by the setup.sh script), follow the manual installation process and run `pip3 install -r requirements-dev.txt`.
+To configure a development environment, best use the `setup.sh` script and answer "y" when asked "Do you want to install dependencies for dev [y/N]? ".
+Alternatively (if your system is not supported by the setup.sh script), follow the manual installation process and run `pip3 install -e .[all]`.
 
 This will install all required tools for development, including `pytest`, `flake8`, `mypy`, and `coveralls`.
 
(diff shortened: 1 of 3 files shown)


hroff-1902:
I have only one doubt here -- do you have any examples, other projects using the `.[target]` syntax for this task?

xmatthias:
what's the doubt on the `.[]` syntax?

it's just saying use "this" project (with these extras)

the `.[]` syntax looks odd since freqtrade is not on pypi, so it's really a "developer" way of installing it. once it's `freqtrade[all]` it's easier to type and use.
[airflow](https://github.com/apache/airflow/blob/master/CONTRIBUTING.md#documentation), for example, uses `.[doc]` as part of the developer documentation - however the "real" documentation states to use `pip install apache-airflow[postgres,mysql,oracle,...]` (Or whatever extra is needed) .

hroff-1902:
> what's the doubt on the `.[]` syntax?

It's a command line (moreover -- a shell script) and `./modulename` syntax is usually used in cli. 
`.modulename` is python syntax but I have not seen a project which exposed it into shell cli.

Second, what's the purpose of square brackets in the syntax you proposed?


P.S, Ah, I recalled, on VMS it was `[DIR1.DIR2]FILENAME.EXT`... 😸

hroff-1902:
ah, I see, it's syntax proposed by pypi: https://pip.pypa.io/en/stable/reference/pip_install/

```
$ pip install SomePackage[PDF]
$ pip install git+https://git.repo/some_pkg.git#egg=SomePackage[PDF]
$ pip install .[PDF]  # project in current directory
$ pip install SomePackage[PDF]==3.0
$ pip install SomePackage[PDF,EPUB]  # multiple extras
```

xmatthias:
Remember that this is not cli / shell - but python - so it'll feed into argparse (or something similar - didn't look under the hood of `pip` ...)

That link you found explains it very well i think (better than what i found while quickly searching this morning...)

PR #2075, opened 2019-07-30
minor: hyperopt cleanups and output improvements

hroff-1902:
* Some minor code cleanups, aligning the use of terms in the names of the vars and dict keys.

Changes visible to the users:
* the number of epoch ("52/100") is now printed to the best results string. It can help localize it in the list of results dumped to the file.
* the value of the objective is now printed in the string with best results as well.
* the "ROI table" and "Best result: ..." is NOW printed to stdout instead of logging. Because details are also printed to stdout.

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -51,7 +51,7 @@ class Hyperopt(Backtesting):
         self.custom_hyperoptloss = HyperOptLossResolver(self.config).hyperoptloss
         self.calculate_loss = self.custom_hyperoptloss.hyperopt_loss_function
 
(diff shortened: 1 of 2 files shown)


hroff-1902:
> LGTM 👍
> 
> > the number of epoch ("52/100") is now printed to the best results string. It can help localize it in the list of results dumped to the file.
> 
> Stupid question - which file would contain the dumped results per epoch?

It already contains -- it's hyperopt_tickerdata.pkl. Maybe optional json in the future and/or csv (with exportes data, but pkl file remains, I guess).

xmatthias:
yes i guess it will, since that's the "format of choice" to move data to other jobs ...

but i don't think you can "easily" access the pickle data ...

PR #2079, opened 2019-08-01
minor: cleanup in hyperopt

hroff-1902:
* modification of the use_sell_signal config setting to True moved to `__init__()` -- it should not be done at every call to `hyperopt_space()`
* minor cleanup in `log_trials_result()` (`params = best_result['params']`)
* detection if roi table should be printed in `log_trials_result()` improved -- this should be done via `has_space()`, because in her custom hyperopt a user may use other names for t1 and other components of the roi table,
* implementation of `has_space()` pythonized
* tests adjusted to not fail with these changes

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -78,6 +78,12 @@ class Hyperopt(Backtesting):
             self.max_open_trades = 0
         self.position_stacking = self.config.get('position_stacking', False),
 
+        if self.has_space('sell'):
+            # Make sure experimental is enabled
+            if 'experimental' not in self.config:
+                self.config['experimental'] = {}
+            self.config['experimental']['use_sell_signal'] = True
+
     def clean_hyperopt(self):
         """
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -78,6 +78,12 @@ def __init__(self, config: Dict[str, Any]) -> None:
             self.max_open_trades = 0
         self.position_stacking = self.config.get('position_stacking', False),
 
+        if self.has_space('sell'):
+            # Make sure experimental is enabled
+            if 'experimental' not in self.config:
+                self.config['experimental'] = {}
```
xmatthias: Side note, probably for another time - this should be moved to `configuration.py`... probably right next to the initialization of `internals`...

hroff-1902: > Side note for another time - this should be moved to `configuration.py`... probably right next to the initialization of `internals`...

I also was thinking of it, right.

PR #2080, opened 2019-08-02
docs: Create detailed section about strategy problem analysis

xmatthias:
## Summary
Improving the documentation about how to analyze problems with a strategy.

Once completed, this shall be followed up with a section about plotting the above data.

### DIFF
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -11,7 +11,7 @@ The following helpers will help you loading the data into Pandas DataFrames, and
 Debugging a strategy (are there no buy signals, ...) can be very time-consuming.
 FreqTrade tries to help you by exposing a few helper-functions, which can be very handy.
 
-I recommend using Juptyer Notebooks for this analysis, since it offers a dynamic way to rerun certain parts.
+It's recommendet using Juptyer Notebooks for analysis, since it offers a dynamic way to rerun certain parts of the code.
 
 The following is a full code-snippet, which will be explained by both comments, and step by step below.
 
@@ -32,6 +32,8 @@ strategy_location = '../xmatt/strategies'
 data_location = '../freqtrade/user_data/data/binance/'
 # Only use one pair here
 pair = "XRP_ETH"
+
+### End constants
 
 # Load data
 bt_data = load_pair_history(datadir=Path(data_location),
@@ -91,7 +93,8 @@ bt_data = load_pair_history(datadir=Path(data_location),
 print(len(bt_data))
 ```
 
-This second section loads the historic data and prints the amount of candles in the data.
+This second section loads the historic data and prints the amount of candles in the DataFrame.
+You can also inspect this dataframe by using `bt_data.head()` or `bt_data.tail()`.
 
 #### Run strategy and analyze results
 
@@ -114,7 +117,7 @@ data = df.set_index('date', drop=True)
 data.tail()
 ```
 
-The code snippet loads and analyzes the strategy, prints the number of buy signals.
+The code snippet loads and analyzes the strategy, calculates and prints the number of buy signals.
 
 The last 2 lines serve to analyze the dataframe in detail.
 This can be important if your strategy did not generate any buy signals.
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -41,9 +41,9 @@ bt_data = load_pair_history(datadir=Path(data_location),
                             pair=pair)
 print(len(bt_data))
 
-
+### Start strategy reload
 # Load strategy - best done in a new cell
-# Needs to be ran each time the strategy-file is changed.
+# Rerun each time the strategy-file is changed.
 strategy = StrategyResolver({'strategy': strategyname,
                             'user_data_dir': Path.cwd(),
                             'strategy_path': location}).strategy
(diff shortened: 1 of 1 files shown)


docs/data-analysis.md
```diff
@@ -6,6 +6,125 @@ A good way for this is using Jupyter (notebook or lab) - which provides an inter
...
+This first section imports necessary modules, and defines some constants you'll probably need to adjust for your case.
+
+#### Load candles
+
+``` python
+# Load data
+bt_data = load_pair_history(datadir=Path(data_location),
+                            ticker_interval = ticker_interval,
+                            pair=pair)
+print(len(bt_data))
```
hroff-1902: It should probably be explicitly mentioned that result is Pd Dataframe.

And probably add that you can print the dataframe or its tail if you need to see the data and an example of it (or refer to below where it's printed).

docs/data-analysis.md
```diff
@@ -6,6 +6,125 @@ A good way for this is using Jupyter (notebook or lab) - which provides an inter
...
+This second section loads the historic data and prints the amount of candles in the data.
+
+#### Run strategy and analyze results
+
+Now, it's time to load and run your strategy.
+For this, I recommend using a new cell in your notebook, since you'll want to repeat this until you're satisfied with your strategy.
+
+``` python
+# Load strategy - best done in a new cell
+# Needs to be ran each time the strategy-file is changed.
```
hroff-1902: "be running"?

docs/data-analysis.md
```diff
@@ -6,6 +6,125 @@ A good way for this is using Jupyter (notebook or lab) - which provides an inter
...
+# Run strategy (just like in backtesting)
+df = strategy.analyze_ticker(bt_data, {'pair': pair})
+print(f"Generated {df['buy'].sum()} buy signals")
+
+# Reindex data to be "nicer" and show data
+data = df.set_index('date', drop=True)
+data.tail()
+```
+
+The code snippet loads and analyzes the strategy, prints the number of buy signals.
```
hroff-1902: "calculates and prints the number..."

docs/data-analysis.md
```diff
@@ -6,6 +6,125 @@ A good way for this is using Jupyter (notebook or lab) - which provides an inter
...
+
+The last 2 lines serve to analyze the dataframe in detail.
+This can be important if your strategy did not generate any buy signals.
+Note that using `data.head()` would also work, however this is misleading since most indicators have some "startup" time at the start of a backtested dataframe.
+
+There can be many things wrong, some signs to look for are:
+
+* Columns with NaN values at the end of the dataframe
+* Columns used in `crossed*()` functions with completely different units
+
```
hroff-1902: Looks a bit incomplete, a couple more of cases should be added (I cannot propose one in the meantime though)

docs/data-analysis.md
xmatthias: you've never used jupyter notebooks, did you? 😄 
you can rerun certain cells (only containing parts of the code) interactively - so i think "Needs to be ran" is correct, since it's an "order".

hroff-1902: > you've never used jupyter notebooks, did you? 😄
> you can rerun certain cells (only containing parts of the code) interactively - so i think "Needs to be ran" is correct, since it's an "order".

I don't even know what it is... 😁 

I use vi (the old-school one, not vim -- I'm too lazy to study extra key commands provided by vim) for text editing (pretty comfortable with it, I know I'll find it on every posix system) and grep for searching for appearances of text strings, class members, etc, in the project directory. So mypy and flake are my best friends because they make python texts more monopenisual... 😁 

Okay, let this be in this form, nevermind...

docs/data-analysis.md
xmatthias: lol well, i was thinking about it too, but i can't think of any more currently you can detect without plotting the data (which I don't want in this PR, as i need to test a few more things first to be sure the interface won't change a week later).

docs/data-analysis.md
xmatthias: i use vi(m) too on servers or for quick edits, vs code for major editing, atom as a quick editor.

Jupyter notebooks are a good way to quickly run python code and see the output right away - then modify that again, and look at the output again. Yes, a REPL can do the same, but not with graphics and "nice" formatted panda output :smile:

docs/data-analysis.md
```diff
@@ -11,7 +11,7 @@ The following helpers will help you loading the data into Pandas DataFrames, and
 Debugging a strategy (are there no buy signals, ...) can be very time-consuming.
 FreqTrade tries to help you by exposing a few helper-functions, which can be very handy.
 
-I recommend using Juptyer Notebooks for this analysis, since it offers a dynamic way to rerun certain parts.
+It's recommendet using Juptyer Notebooks for analysis, since it offers a dynamic way to rerun certain parts of the code.
```
hroff-1902: recommended

xmatthias: i will stop ~~reccomending~~ ~~recommenting~~ recommending stuff in the documentation ... 
it's not that i wouldn't know that i fail this word each time ...

hroff-1902: yeah, each time 🤣 

the rest is okay, you can merge it

PR #2089, opened 2019-08-03
Hyperopt print colorized results

hroff-1902:
The `--color/--print-colorized` option to print colorized results:
* bad (zero trades or limited by `--min-trades`) are in red;
* current bests -- in green (also when used without `--print-all`)
* epochs with positive total profit -- in bold

This option helps a user recognize visually the performance of hyperopt evaluations, especially when used with `--print-all`.

Brief description added to hyperopt docs (also a description for `--print-all` which was missed).

### DIFF
--- a/docs/hyperopt.md
+++ b/docs/hyperopt.md
@@ -343,6 +343,10 @@ def populate_buy_trend(self, dataframe: DataFrame) -> DataFrame:
     return dataframe
 ```
 
+You can use the `--print-all` command line option if you would like to see all results in the hyperopt output, not only the best ones.
+
+When the `--color/--print-colorized` command line option is used, the results are colorized -- bad results (with zero trades or limited by the `--min-trades` option) are red, currest bests -- in green, results with positive total profit are printed in bold.
+
 ### Understand Hyperopt ROI results
 
 If you are optimizing ROI, you're result will look as follows and include a ROI table.
--- a/docs/hyperopt.md
+++ b/docs/hyperopt.md
@@ -343,9 +343,9 @@ def populate_buy_trend(self, dataframe: DataFrame) -> DataFrame:
     return dataframe
 ```
 
-You can use the `--print-all` command line option if you would like to see all results in the hyperopt output, not only the best ones.
+By default, hyperopt prints colorized results -- epochs with positive profit are printed in the green color. This highlighting helps you find epochs that can be interesting for later analysis. Epochs with zero total profit or with negative profits (losses) are printed in the normal color. If you do not need colorization of results (for instance, when you are redirecting hyperopt output to a file) you can switch colorization off by specifying the `--no-color` option in the command line.
 
-When the `--color/--print-colorized` command line option is used, the results are colorized -- bad results (with zero trades or limited by the `--min-trades` option) are red, current bests -- in green, results with positive total profit are printed in bold.
+You can use the `--print-all` command line option if you would like to see all results in the hyperopt output, not only the best ones. When `--print-all` is used, current best results are also colorized by default -- they are printed in bold (bright) style. This can also be switched off with the `--no-color` command line option.
 
 ### Understand Hyperopt ROI results
 
--- a/docs/hyperopt.md
+++ b/docs/hyperopt.md
@@ -345,7 +345,7 @@ def populate_buy_trend(self, dataframe: DataFrame) -> DataFrame:
 
 You can use the `--print-all` command line option if you would like to see all results in the hyperopt output, not only the best ones.
 
-When the `--color/--print-colorized` command line option is used, the results are colorized -- bad results (with zero trades or limited by the `--min-trades` option) are red, currest bests -- in green, results with positive total profit are printed in bold.
+When the `--color/--print-colorized` command line option is used, the results are colorized -- bad results (with zero trades or limited by the `--min-trades` option) are red, current bests -- in green, results with positive total profit are printed in bold.
 
 ### Understand Hyperopt ROI results
 
--- a/freqtrade/configuration/cli_options.py
+++ b/freqtrade/configuration/cli_options.py
@@ -192,10 +192,11 @@ AVAILABLE_CLI_OPTIONS = {
         default=False,
     ),
     "print_colorized": Arg(
-        '--color', '--print-colorized',
-        help='Print colorized hyperopt results.',
-        action='store_true',
-        default=False
+        '--no-color',
+        help='Disable colorization of hyperopt results. May be useful if you are '
+        'redirecting output to a file.',
+        action='store_false',
+        default=True,
     ),
     "hyperopt_jobs": Arg(
         '-j', '--job-workers',
--- a/freqtrade/configuration/cli_options.py
+++ b/freqtrade/configuration/cli_options.py
@@ -191,6 +191,12 @@ AVAILABLE_CLI_OPTIONS = {
         action='store_true',
         default=False,
     ),
+    "print_colorized": Arg(
+        '--color', '--print-colorized',
+        help='Print colorized hyperopt results.',
+        action='store_true',
+        default=False
+    ),
     "hyperopt_jobs": Arg(
         '-j', '--job-workers',
         help='The number of concurrently running jobs for hyperoptimization '
--- a/freqtrade/misc.py
+++ b/freqtrade/misc.py
@@ -113,3 +113,15 @@ def deep_merge_dicts(source, destination):
             destination[key] = value
 
     return destination
+
+
+def green(s):
+    return '\033[92m' + s + '\033[0m'
+
+
+def red(s):
+    return '\033[91m' + s + '\033[0m'
+
+
+def bold(s):
+    return '\033[1m' + s + '\033[0m'
--- a/freqtrade/misc.py
+++ b/freqtrade/misc.py
@@ -113,15 +113,3 @@ def deep_merge_dicts(source, destination):
             destination[key] = value
 
     return destination
-
-
-def green(s):
-    return '\033[92m' + s + '\033[0m'
-
-
-def red(s):
-    return '\033[91m' + s + '\033[0m'
-
-
-def bold(s):
-    return '\033[1m' + s + '\033[0m'
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -13,14 +13,15 @@ from pathlib import Path
 from pprint import pprint
 from typing import Any, Dict, List, Optional
 
+from colorama import init as colorama_init
 from joblib import Parallel, delayed, dump, load, wrap_non_picklable_objects, cpu_count
 from pandas import DataFrame
 from skopt import Optimizer
 from skopt.space import Dimension
+from termcolor import colored
 
 from freqtrade.configuration import Arguments
 from freqtrade.data.history import load_data, get_timeframe
-from freqtrade.misc import green, red, bold
 from freqtrade.optimize.backtesting import Backtesting
 # Import IHyperOptLoss to allow users import from this file
 from freqtrade.optimize.hyperopt_loss_interface import IHyperOptLoss  # noqa: F4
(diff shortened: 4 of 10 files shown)


freqtrade/misc.py
```diff
@@ -113,3 +113,15 @@ def deep_merge_dicts(source, destination):
             destination[key] = value
 
     return destination
+
+
+def green(s):
+    return '\033[92m' + s + '\033[0m'
```
xmatthias: I would want type-hints (at least) here ...
otherwise this is going to fail badly the moment someone has the idea to print a integer in green ...

In general, i'm not sure about doing this ourselfs.
 
Maybe we should simply depend on some tested library like [colorama](https://pypi.org/project/colorama/) or [termcolor](https://pypi.org/project/termcolor/) - since i guess the color-codes won't work on all terminal types the same way... and i doubt that we have much chances to test this ...

hroff-1902: That's borrowed from the scripts (get_market_pairs.py).

I was writing a VT100 emulator with DEC extensions from scratches 20 years ago so I still remember the sequences and the diffs between various terminals... Believe me -- it's just a subset of standard VT100 sequences, should (still) be supported on any physical or virtual terminal...

However, I'll take a look at colorama and termcolor, thanks...

hroff-1902: Ah, Windows... So colorama is needed, right.

hroff-1902:
Pls test it on Windows cmd (and/or powershell) terminal (so that colorama will be in effect).

For example:
```
freqtrade hyperopt -e 500 --hyperopt-loss OnlyProfitHyperOptLoss --min-trades 20 --print-colorized --print-all
freqtrade hyperopt -e 500 --hyperopt-loss OnlyProfitHyperOptLoss --min-trades 20 --print-colorized
```

requirements-common.txt
```diff
@@ -30,3 +30,7 @@ sdnotify==0.3.2
 
 # Api server
 flask==1.1.1
+
+# Support for colorized terminal output
+termcolor==1.1.0
+colorama==0.4.1
```
xmatthias: i wouldn't necessarily depend on both ... (that's not what i meant, anyway).

are there benefits to do it this way?

freqtrade/optimize/hyperopt.py
```diff
@@ -339,6 +354,9 @@ def start(self) -> None:
         logger.info(f'Number of parallel jobs set as: {config_jobs}')
 
         opt = self.get_optimizer(config_jobs)
+
+        colorama_init()
```
xmatthias: according to the colorama docs, when calling `init()` - we should call `deinit()` too, to reset the terminal (stdout, stderr) to how it was.

freqtrade/misc.py
xmatthias: Confirmed that current version is working in windows

freqtrade/optimize/hyperopt.py
hroff-1902: `deinit()` is needed "to stop using colorama before your program exits" (cited from the docs). We do not need to stop using colorama *before* freqtrade exits. we are stopping it *at* the program exit (automatically).

requirements-common.txt
hroff-1902: `init(autoreset=True)` would be useful if I did not use `colored` from termcolor, which handles all the reset sequences for us...

Well, as I said, these are just VT100 (ANSI) sequences, nothing more, nothing less (colorama also translates them to Windows sequences, so it's really needed). I found colorama offering exactly same approach -- a constant for a sequence, I would make same concatenations of the strings, probably implement those `red()`, `green()` functions again, but now with colorama constants instead of hardcoded... I found the termcolor `colored` interface convenient so it's been used...

It's not just arbitrarily, if you worry about this. There can be different targets during implementation, one of which is the code convenience for devs. Although this adds an extra dependency in this case...

hroff-1902:
If you wish I could reimplement it using colorama constants and `init(autoreset=True)` mode...

freqtrade/optimize/hyperopt.py
```diff
@@ -153,8 +155,19 @@ def log_results(self, results) -> None:
...
+                self.current_best_loss = results['loss']
             log_str = self.format_results_logstring(results)
+            # Colorize output
+            if self.config.get('print_colorized', False):
+                if results['total_profit'] > 0:
+                    log_str = Style.BRIGHT + log_str
+                if results['loss'] >= MAX_LOSS:
+                    log_str = Fore.RED + log_str
+                elif is_best_loss:
+                    log_str = Fore.GREEN + log_str
```
xmatthias: From looking at these results (see screenshot below) - i'm not sure if the meaning of `BRIGHT` and `GREEN` shouldn't be reversed (green when it's profitable, Bright when it's the current best)?

8/20 seems more important in the screenshot - however the best result is 12/20 (yes, i'm using the default loss function - but that's not the point).

hyperopt_loss_color

another thing i found strange is 1/20 and 3/20 - they have negative profit (-> lose my money) - but are green too. i would color these red too ... so it's clearer?

freqtrade/configuration/cli_options.py
```diff
@@ -191,6 +191,12 @@ def __init__(self, *args, **kwargs):
         action='store_true',
         default=False,
     ),
+    "print_colorized": Arg(
+        '--color', '--print-colorized',
+        help='Print colorized hyperopt results.',
+        action='store_true',
+        default=False
```
xmatthias: I think this provides quite good benefits and improves usability and output readability enormously...

shouldn't this be the default and `--no-color` can disable colorized output (for example if piping the result to a file or so)?

hroff-1902: Yeah, it improves readability alot, that's why I implemented it... 😆 

No problem, I could make it true by default, but this is much useful with `--print-all` (which I use frequently), which is not a default mode (and should not be default). In the no-print-all mode it just prints profitable results in bold (currently, if colorization schema won't be changed). If you think this is also useful, please confirm and I'll change colorization to be used by default...

freqtrade/optimize/hyperopt.py
hroff-1902: Let's adjust the colorization schema if needed.

The lines in red -- are so called 'bad' results (those with objective equal or higher than MAX_LOSS value). I.e. the epochs with zero trades or with number of trades that are less than the number set with `--min-trades`. They can also be profitable. So the bold attribute (which can be used in conjunction with red color) was choosen. We cannot use green and red in the same time...

So the current colorization schema is the following:
* 'bad' epochs -- red (profitable 'bad's -- bold red). Not seen in no-print-all mode.
* current best -- green (profitable bests -- bold green). In no-print-all all results are green since hyperopt prints only bests.
* others -- white (default color, more precisely) (profitable others -- bold white(bold default color))
* the resulting best result is not colorized since it's not needed.

Any idea on how to change the schema particularly?

xmatthias: how about
* current best: yellow (or any other color apart from blue) (profitable best - green)
* bad epoch - red
* others: white

It's mostly the distinction between green and bold green which is not very clear i think

freqtrade/configuration/cli_options.py
xmatthias: i would make it default - yes :+1:

freqtrade/optimize/hyperopt.py
hroff-1902: need to view this on a white terminal with black default foreground...

xmatthias: that's why i said "or any other color and not blue (blue is often too dark on black backgrounds - mainly on windows putty... )

maybe orange then ... ?

xmatthias: how about just inversing green and bright?

so "positive" is green - and current best is bold?

hroff-1902: with yellow it's impossible to follow -- it's like a rainbow, too many colors
you may see, I committed this version...

hroff-1902: I tagged the schema above ^^ as `colorized-schema-1`

Also committed the next one (tagged as `colorized-schema-2`): green for profitable epochs, red for losses, bright for bests (can be both red and green), dimmed for bad objectives (also both red and green). There is no ANSI terminal sequence for italic, so dimmed color was used for bads. On Windows colorama renders dimmed as normal, you won't see dimmed colors. It still looks like a new year tree (6 different colors), but at least eyes can navigate in it, in my opinion...

I also have a next variant...

hroff-1902: Here we are, colorization schema-3 (tagged as `colorized-schema-3`): green for profits, red for losses, bright for bests, but only green ones. We actually do not quite need to highlight bad epochs and bests with losses -- they still are simply losses (simply red) or profits which are not best (simply green)...

I think my initial schema was pretty fine (I tagged it as `colorized-schema-0`) -- it contained lots of white lines so the green and red (and also bold) were not so blind for eyes... This last one is also, ghm, let's say fine... At least its can be explained better in the docs...

xmatthias: colorized-schema-2 (dimmed) is too dark i think
colorized-schema-1 seems okish to me

colorized-schema-3 is "too" colorfull i guess
so 0 or 1 should win - i'll leave this to you ... thanks for giving us choice! :+1:

hroff-1902: Well, for me colorized-schema-1 (with yellow) is too hard to follow with eyes, colorized-schema-2 (dimmed) is also too colorized (many diff. tones of colors).

So as for above, I vote for colorized-schema-0 and colorized-schema-3...

However, I created `colorized-schema-4` (also tagged it), pretty simple: green for profitable, bold(bright) for bests. Bads are not highlighted, because 1) losses are still losses, 2) profits are still profits, so even epochs that were detected as bads may also be of interest for the user. So, no red color in this schema, let's reserve red for something more important in the output.

Huh?

xmatthias: 4 wins :+1: 

You're right, loss is loss (the reason is not that important ...)

hroff-1902: I could squash all intermediate commits with previous schemas into the final colorization schema, if this would make sence...

xmatthias: yes, please

freqtrade/configuration/cli_options.py
hroff-1902: done.

freqtrade/optimize/hyperopt.py
hroff-1902: done.

hroff-1902:
pls review the resulting variant, should be ready for merge, I guess

docs/hyperopt.md
```diff
@@ -343,6 +343,10 @@ def populate_buy_trend(self, dataframe: DataFrame) -> DataFrame:
     return dataframe
 ```
 
+You can use the `--print-all` command line option if you would like to see all results in the hyperopt output, not only the best ones.
+
+When the `--color/--print-colorized` command line option is used, the results are colorized -- bad results (with zero trades or limited by the `--min-trades` option) are red, current bests -- in green, results with positive total profit are printed in bold.
```
xmatthias: You will want to change this to `--no-color` too please

requirements-common.txt
```diff
@@ -30,3 +30,6 @@ sdnotify==0.3.2
 
 # Api server
 flask==1.1.1
+
+# Support for colorized terminal output
+colorama==0.4.1
```
xmatthias: Please add this to `setup.py` as well (as `install_requires`)

hroff-1902: Done, have not noticed it, thanks...

docs/hyperopt.md
hroff-1902: Sure, I've forgotten about the docs, my bad...
Done.

xmatthias: that's why we do reviews :+1:

Issue #2083, opened 2019-08-02
Plotting problems


Prossi79:
## Step 2: Describe your environment

  * Python Version: Python 3.7.3
  * CCXT version: ccxt==1.18.992
  * Branch: Develop
  * Last Commit ID: 1810d865558b7a6f2f154aef069b6c22a008a9c2

  
## Step 3: Describe the problem:

Trying to plot the live trades out of my DB

### Steps to reproduce:

see log below
  
### Observed Results:

see log below   

### Relevant code exceptions or logs:

  ```
 (.env) (base) root@vmanager6003:~/freqtrade# python3 scripts/plot_dataframe.py --config config.json  --config configC.json --db-url sqlite:///MACDStrategy_LINK_live.sqlite -p LINK/BTC --trade-source DB
2019-08-02 21:01:53,435 - freqtrade.loggers - INFO - Verbosity set to 0
2019-08-02 21:01:53,436 - freqtrade.configuration.configuration - INFO - Parameter --db-url detected ...
2019-08-02 21:01:53,436 - freqtrade.configuration.configuration - INFO - Dry run is disabled
2019-08-02 21:01:53,436 - freqtrade.configuration.configuration - INFO - Using DB: "sqlite:///MACDStrategy_LINK_live.sqlite"
2019-08-02 21:01:53,436 - freqtrade.configuration.check_exchange - INFO - Checking exchange...
2019-08-02 21:01:53,437 - freqtrade.configuration.check_exchange - INFO - Exchange "binance" is officially supported by the Freqtrade development team.
2019-08-02 21:01:53,437 - freqtrade.configuration.configuration - INFO - Using max_open_trades: 2 ...
2019-08-02 21:01:53,438 - freqtrade.configuration.configuration - INFO - Using data directory: user_data/data/binance ...
2019-08-02 21:01:53,438 - freqtrade.configuration.configuration - INFO - Storing backtest results to user_data/backtest_data/backtest-result.json ...
2019-08-02 21:01:53,438 - freqtrade.configuration.configuration - INFO - Using pairs LINK/BTC
2019-08-02 21:01:53,439 - freqtrade.configuration.configuration - INFO - Using indicators1: sma,ema3,ema5
2019-08-02 21:01:53,439 - freqtrade.configuration.configuration - INFO - Using indicators2: macd,macdsignal
2019-08-02 21:01:53,439 - freqtrade.configuration.configuration - INFO - Limiting plot to: 750
[...]

PR #2090, opened 2019-08-03
load_trades_db should give as many columns as possible

xmatthias:
## Summary
We were missing one needed column in the load_trades_db function.
 
Solve the issue: #2083

## Quick changelog

- Add multiple columns to have all that backtesting gives, and some more for extendet analysis

### DIFF
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -81,19 +81,30 @@ def load_trades_from_db(db_url: str) -> pd.DataFrame:
     """
     trades: pd.DataFrame = pd.DataFrame([], columns=BT_DATA_COLUMNS)
     persistence.init(db_url, clean_open_orders=False)
-    columns = ["pair", "profit", "open_time", "close_time",
-               "open_rate", "close_rate", "duration", "sell_reason",
-               "max_rate", "min_rate"]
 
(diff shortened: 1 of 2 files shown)


freqtrade/data/btanalysis.py
```diff
@@ -81,19 +81,30 @@ def load_trades_from_db(db_url: str) -> pd.DataFrame:
...
+               "stop_loss", "initial_stop_loss", "strategy", "ticker_interval"]
+
+    trades = pd.DataFrame([(t.pair,
                             t.open_date.replace(tzinfo=pytz.UTC),
                             t.close_date.replace(tzinfo=pytz.UTC) if t.close_date else None,
-                            t.open_rate, t.close_rate,
+                            t.calc_profit(), t.calc_profit_percent(),
+                            t.open_rate, t.close_rate, t.amount,
                             t.close_date.timestamp() - t.open_date.timestamp()
                             if t.close_date else None,
```
hroff-1902: indentation should be applied here?

xmatthias: I thought so / think so  - but flake disagrees ... but it works with parenteses ...

hroff-1902: 👍

Issue #1850, opened 2019-05-13
process_only_new_candles not applied correctly when trades happen


gaugau3000:
Can freqtrade prevent new buys when stop loss is reach and signal buy still active. 

Example (5 minutes strategy max_trade_open = 1 ) : Freqtrade buy at 10h00 stop loss reached at 10:02 (2% stop loss) and buy another pair at 10h02 but I don't want because it is 2 minutes too late.

Many tks

xmatthias:
no - and i am not sure that makes sense either ... for one, you should work on the strategy ...

However, once a trade is closed (via ROI or stoploss) - then that trade is over and does not really matter for future signals. We could (maybe) implement a "cooldown" period after stoplosses to prevent repetitive buying (which might tie into #1841) - however i'm not sure this isn't responsability of the strategy...

On the other hand - if a coin went down, chances are it'll come back up soonish (assuming no strong trend is present) - so buying shortly after that can make sense.

If your concern is regarding the buy in the middle of a candle (5m candles - should in theory buy only at 5 / 0 marks) - then `self.process_only_new_candles=True` in the strategy is your friend (which only analyzes when new candles arrive - not at the 2m mark). 

Note that this will also affect your sell-signal - but does NOT apply to ROI / stoploss.

mishaker:
I think @gaugau3000 is talking about buy signal used once. interval 5min, buy at 12:00, stoploss hit at 12:02 then 5sec after buys again ... bot does not have a memory of already used signal.

hroff-1902:
What can "another pair" mean in the question above?

@gaugau3000, could you illustrate the point in some more details?

gaugau3000:
@hroff-1902 yes i will try to explain better.

I have a 2% stop loss and only 1 max open trade my ticker interval is 5 minutes and the process only new candle is set to true.

The bot make a buy at 10:00 (BTC/ETH) -> stop loss is reached at 10:02 -> then the bot make (immediatly) a new buy an another pair (BTC/ETC) . But i don't want he do a new buy an BTC/ETC has it is too late.

@xmatthias 

> no - and i am not sure that makes sense either ... for one, you should work on the strategy ...

Yes i am sure i have to improve but my concern is just don't hold an asset too long when the market is agitated not to miss opportunities as i have only 1 max open trade.

[...]

PR #2092, opened 2019-08-04
Split analyze_ticker

xmatthias:
## Summary / changelog
This will split analyze_ticker into 2 functions.

* `analyze_ticker()` is the public interface, ONLY running the strategy's functions, without additonal logic. This function should be used for data analysis, and plot-scripts, where `process_only_new_candles` is irrelevant since we always want this analyed. Having the logic for `process_only_new_candles` in place can prevent plots to be correct.
* `_analyze_ticker_int()` handles the logic if the ticket needs to be analyzed or if `process_only_new_candles` kicks in.

As a consequence, `generate_dataframe()` in `scripts/plot_dataframe.py` has beenremoved, since this is now a copy of `analyze_ticker()`.

* Added output to show the plot-location (since i did not refind my plot ...).

This is probably the last sub-PR  before moving the plot-scripts to freqtrade subcommands.

I think it's also an enabler for #1850 ... (even though this does not fix these).

### DIFF
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -318,6 +318,7 @@ def store_plot_file(fig, filename: str, auto_open: bool = False) -> None:
     """
 
     Path("user_data/plots").mkdir(parents=True, exist_ok=True)
-
-    plot(fig, filename=str(Path('user_data/plots').joinpath(filename)),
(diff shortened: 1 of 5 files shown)


xmatthias:
Good point about the naming - fixed

PR #2094, opened 2019-08-05
Simplify custom hyperopts -- no need to copy ugly methods in every custom implementation

hroff-1902:
The `roi_space()`, `stoploss_space()`, `generate_roi_table()` methods moved to `IHyperOpt` and can be overridden in the custom Hyperopts. Removed from DefaultHyperOpt and SampleHyperOpt.

The default ranges of values for the roi and stoploss spaces as well as the default structure of the generated ROI tables suffice in most cases. So no need to copy those ugly methods to every custom implementation anymore.

Detailed description added to hyperopt.md (see changes).

A few extra cosmetic changes in the docs and in the helpstrings.


user_data/hyperopts/sample_hyperopt.py
```diff
@@ -18,14 +18,19 @@
...
     More information in https://github.com/freqtrade/freqtrade/blob/develop/docs/hyperopt.md
-     You can:
-    - Rename the class name (Do not forget to update class_name)
-    - Add any methods you want to build your hyperopt
-    - Add any lib you need to build your hyperopt
-     You must keep:
-    - the prototype for the methods: populate_indicators, indicator_space, buy_strategy_generator,
-    roi_space, generate_roi_table, stoploss_space
+
+    You can:
```
xmatthias: You should?

user_data/hyperopts/sample_hyperopt.py
```diff
@@ -18,14 +18,19 @@
...
+    - Rename the class name.
+    - Add any methods you want to build your hyperopt.
+    - Add any lib you need to build your hyperopt.
+
+    You must keep:
+    - The prototypes for the methods: populate_indicators, indicator_space, buy_strategy_generator.
+
+    The roi_space, generate_roi_table, stoploss_space methods were moved to the parent class, you may
```
xmatthias: I don't think we ~~need to~~ should mention that these were moved to the parent class ... new users will not know this - so mentioning that these are available, and can be overridden should suffice.

user_data/hyperopts/sample_hyperopt.py
```diff
@@ -167,42 +172,6 @@ def sell_indicator_space() -> List[Dimension]:
...
-        """
-        Values to search for each ROI steps
-        """
-        return [
-            Integer(10, 120, name='roi_t1'),
-            Integer(10, 60, name='roi_t2'),
-            Integer(10, 40, name='roi_t3'),
-            Real(0.01, 0.04, name='roi_p1'),
-            Real(0.01, 0.07, name='roi_p2'),
-            Real(0.01, 0.20, name='roi_p3'),
```
xmatthias: This creates one problem - there is no template on how to do this anymore (which is what `sample_hyperopt`used to be here for)

As a user, i would not want do digg through the parent class (which i don't know where it is) to find how the interface for this looks like.

Maybe adding the template for each of the functions to the hyperopt documentation will suffice...?

user_data/hyperopts/sample_hyperopt.py
hroff-1902: I also have had doubts about this phrase...

user_data/hyperopts/sample_hyperopt.py
hroff-1902: Well, I agree.
I just do not want to add those samples right in the hyperopt.md (and keep it in the sample_hyperopt.md) -- I suppose overriding these methods will be rarely required and is not absolutely needed for a novice user reading the docs. But you are right -- for a more advanced user there should be an explicit example.

In principle, there can be the new "Advanced Freqtrade Techniques" part be added to the docs, with potential content:
* details and examples of using multiple configs
* how to run properly multiple instances of the bot (with common parts of config, with diff. databases)
* _how to override these hyperopt methods. (this issue)_
* how to run freqtrade as a system service
* more on hyperopting, etc.

In the meantime, I could create the `sample_advanced_hyperopt.py` (naming?) with those methods (a copy of current (old) sample_hyperopt.py with some changes in the comments). The "Advanced Freqtrade Techniques" can later contain quotes from it and refer to it.

xmatthias: >  I suppose overriding these methods will be rarely required 

this may be true for `stoploss_space()` and `generate_roi_table()` (especially this one!!), however `roi_space()` needs to be changed depending on the timeframe, so every other user will probably need to override this.

Assuming you optimize on 1h candles, in which case our defaults make 0 sense, since 2 of the 3 options are < 1 candle size.
i guess your assumption is similar to mine - using ~5m candles - but it's dangerous since i see many users seem to use 30m, 1h or even longer timeframes, in which case this method is completely unsuitable. 

The documentation does require a general overhaul / reorganisation ... some parts are duplicated all over the place, and others are missing completely (but that shall not be part of this PR).

hroff-1902: Agree (for `roi_space` on long timeframes), will think about this point...

xmatthias: in hyperopt, or in general the roi dict?

hroff-1902: both. in hyperopt it's a set of ranges, in the strategy -- just a point from that parameters hyperspace...

For this PR, I'm going to add a new `AdvancedSampleHyperOpts` (in the new `advanced_sample_hyperopt.py`) class, which keeps the `stoploss_space()`, `generate_roi_table()` and `roi_space()` as samples of how they can be overridden, with appropriate comments. The `stoploss_space()` sample will shorten the default range of the stoploss hyperspace, `generate_roi_table()` will be same with some extra notes on when this method should be overridden, and a 'scaled' version of the 'roi_space()' which is appropriate for 1h and 1d timeframes. I know how to scale minutes in the roi tables (the roi_tN values --  by multiplying the default ranges to the timeframe period), but thinking on how to scale the ranges of the roi values (the roi_pN values), what can be appropriate limits for their ranges in the case of long timeframes...

xmatthias: well that's the point - it's intended to be adjusted to the strategy / timeframe you're using.

I wouldn't change it from minutes to candles though, as i may want to change ROI it mid-candle (hourly candles, lower ROI after 90 minutes since i know that it either wins immediately, or not at all ...).

user_data/hyperopts/sample_hyperopt.py
hroff-1902: More appropriate and detailed comment added, pls verify.

user_data/hyperopts/sample_hyperopt.py
hroff-1902: Done.

hroff-1902:
The sample with the old (still working, but not necessary required) implementation of the roi and stoploss spaces added in the new `user_data/sample_hyperopt_advanced.py`

Helpstring in SampleHyperOpt changed to contain reference to this sample to guide the user.

docs/hyperopt.md
```diff
@@ -376,6 +381,41 @@ minimal_roi = {
...
 ```
 
+If you are optimizing ROI, Freqtrade creates the 'roi' optimization hyperspace for you -- it's the hyperspace of components for the ROI tables. By default, each ROI table generated by the Freqtrade consists of 4 rows (steps) with the values that can vary in the following ranges:
+
+| # | minutes | ROI percentage |
+|---|---|---|
+| 1 | always 0 | 0.03...0.31 |
+| 2 | 10...40 | 0.02...0.11 |
+| 3 | 20...100 | 0.01...0.04 |
+| 4 | 30...220 | always 0 |
```
xmatthias: are you sure this table is correct?

the code defines the minutes to be always between 10 and X (x beeing 40, 60 and 120)?
i suspect you wanted to change these defaults in the code ... ?

(copy-pasted from below):

```
            Integer(10, 120, name='roi_t1'),	
            Integer(10, 60, name='roi_t2'),	
            Integer(10, 40, name='roi_t3'),
```

docs/hyperopt.md
```diff
@@ -376,6 +381,41 @@ minimal_roi = {
...
+
+This structure of the ROI table is sufficient in most cases. Override the `roi_space()` method defining the ranges desired if you need components of the ROI tables to vary in other ranges.
+
+Override the `generate_roi_table()` and `roi_space()` methods and implement your own custom approach for generation of the ROI tables during hyperoptimization in these methods if you need a different structure of the ROI table or other amount of rows (steps) in the ROI tables.
```
xmatthias: This and the previous sentence are kind of repeating the same thing but in different words, should probably be combined and simplified (?)

docs/hyperopt.md
```diff
@@ -376,6 +381,41 @@ minimal_roi = {
...
+
+### Understand Hyperopt Stoploss results
+
+If you are optimizing stoploss values, your result will look as follows and include stoploss:
```
xmatthias: How are you optimizing stoploss values? (same for ROI above)?

maybe something like

```
If you are optimizing stoploss values (search-space includes `all` or `stoploss`), your result will look as follows and include stoploss:
```
?

docs/hyperopt.md
hroff-1902: > are you sure this table is correct?

Yes,
please refer to `generate_roi_table()`  for how these values are used.
these are the "steps" (), the resulting value (second one, for example) is calculated as `roi_t3 + roi_t2`, so it's between 20 and 100.

> i suspect you wanted to change these defaults in the code ... ?
No, I instead wrote an adaptive roi_space() function that automatically adjusts to the ticker_interval (next PR).

docs/hyperopt.md
hroff-1902: Agree, this adds some more clearness, done (also for the 'roi' section).

docs/hyperopt.md
xmatthias: lol yes you are right - my bad, sorry.

This is 0 intuitive - we should have this documented for `roi_space()` - but i guess that'll be next PR since this will change anyway?

docs/hyperopt.md
hroff-1902: First sentence directs user to override `roi_space` if she wants to have different *ranges* of the roi tables and keep the structure of the generated roi tables.

Second statement says what functions she needs to override if she wants to change the number of steps or the whole structure of the generated tables...

Fell free to improve the text here, but I think it says about different things (well, I know that texts written are always perceived by the author differently than from outside...)

Issue #2099, opened 2019-08-05
Backtesting does not fetch ticker_interval from the strategy


hroff-1902:
Remove ticker_interval from the config, run backtesting with a list of strategies:

```
$ freqtrade -c user_data/configs/test-config.json backtesting --strategy-list TestStrategy3 TestStrategy4
2019-08-05 23:13:51,762 - freqtrade.loggers - INFO - Verbosity set to 0
2019-08-05 23:13:51,763 - freqtrade.configuration.configuration - INFO - Dry run is enabled
2019-08-05 23:13:51,763 - freqtrade.configuration.configuration - INFO - Using DB: "sqlite:///test.sqlite"
2019-08-05 23:13:51,763 - freqtrade.configuration.check_exchange - INFO - Checking exchange...
2019-08-05 23:13:51,763 - freqtrade.configuration.check_exchange - INFO - Exchange "binance" is officially supported by the Freqtrade development team.
2019-08-05 23:13:51,763 - freqtrade.configuration.configuration - INFO - Using max_open_trades: 10 ...
2019-08-05 23:13:51,763 - freqtrade.configuration.configuration - INFO - Using data directory: user_data/data/binance ...
2019-08-05 23:13:51,764 - freqtrade.configuration.configuration - INFO - Using strategy list of 2 Strategies
2019-08-05 23:13:51,764 - freqtrade.configuration.configuration - INFO - Storing backtest results to user_data/backtest_data/backtest-result.json ...
2019-08-05 23:13:51,764 - freqtrade.optimize - INFO - Starting freqtrade in Backtesting mode
2019-08-05 23:13:51,764 - freqtrade.exchange.exchange - INFO - Instance is running with dry_run enabled
2019-08-05 23:13:51,764 - freqtrade.exchange.exchange - INFO - Applying additional ccxt config: {'enableRateLimit': True}
2019-08-05 23:13:51,768 - freqtrade.exchange.exchange - INFO - Applying additional ccxt config: {'enableRateLimit': False}
2019-08-05 23:13:51,786 - freqtrade.exchange.exchange - INFO - Using Exchange "Binance"
2019-08-05 23:13:52,578 - freqtrade.resolvers.exchange_resolver - INFO - Using resolved exchange 'Binance'...
[...]

PR #2101, opened 2019-08-06
Backtest ticker interval unset

xmatthias:
## Summary
ticker-interval needs to be set for backtesting to work.
In case of strategy-list, it needs to be defined on either config or cli level, since each strategy could define it's own interval.

Solves #2099

## Quick changelog

- Fix uncought error
- add test for this behaviour

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -10,8 +10,8 @@ from pathlib import Path
 from typing import Any, Dict, List, NamedTuple, Optional
 
 from pandas import DataFrame
-from tabulate import tabulate
 
+from freqtrade import OperationalException
 from freqtrade.configuration import Arguments
 from freqtrade.data import history
 from freqtrade.data.dataprovider import DataProvider
@@ -21,6 +21,7 @@ from freqtrade.persistence import Trade
 from freqtrade.resolvers import ExchangeResolver, StrategyResolver
 from freqtrade.state import RunMode
 from freqtrade.strategy.interface import IStrategy, SellType
+from tabulate import tabulate
 
 logger = logging.getLogger(__name__)
 
@@ -88,6 +89,9 @@ class Backtesting(object):
         Load strategy into backtesting
         """
         self.strategy = strategy
+        if "ticker_interval" not in self.config:
+            raise OperationalException("Ticker-interval needs to be set in either configuration "
+                                       "or as cli argument `--ticker-interval 5m`")
 
         self.ticker_interval = self.config.get('ticker_interval')
         self.ticker_interval_mins = timeframe_to_minutes(self.ticker_interval)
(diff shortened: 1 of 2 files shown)


xmatthias:
it doesn't really make a difference, since the only position this can be moved to is one line before `set_strategy` is called, since the strategy-loader can (in case of a single strategy) set the ticker_interval.

To be honest, i didn't spend much thought on moving it out of `set_strategy()` ... i moved the check to where we first use this - it's 2 executable lines between now and where we could move it to ...

hroff-1902:
> it doesn't really make a difference, since the only position this can be moved to is one line before `set_strategy` is called, since the strategy-loader can (in case of a single strategy) set the ticker_interval.
> 
> To be honest, i didn't spend much thought on moving it out of `set_strategy()` ... i moved the check to where we first use this - it's 2 executable lines between now and where we could move it to ...

First of all, I appreciate you fixing that issue I filed (#2099 ).
Currently yes, it's just 2 lines above, but in the refactoring I mentioned, set_strategy will remain in Backtesting while most of current init will go to BacktestingCli. No problem with moving this check, but tests will also require to be adjusted in this case... And I thought I'm still not as familiar with pytest to make some sophisticated changes in tests effectively...

Okay, I'll make necessary changes during that my refactoring, let's merge it, thanks!

> 
> **edit**: there is even another enhancement in this structure - if ALL strategies define the same ticker-interval we COULD use that ... but i suspect most people will have ticker_interval in the configuration anyway.

I also thought of this, but will not the resulting behavior be too complex to describe it for the users in the docs?

xmatthias:
> Currently yes, it's just 2 lines above, but in the refactoring I mentioned, set_strategy will remain in Backtesting while most of current init will go to BacktestingCli 

huh? i must have missed that ... ? where can i find this mention since now i'm curious ...

> I also thought of this, but will not the resulting behavior be too complex to describe it for the users in the docs?

Yes - that's why it's not there

hroff-1902:
> > Currently yes, it's just 2 lines above, but in the refactoring I mentioned, set_strategy will remain in Backtesting while most of current init will go to BacktestingCli
> 
> huh? i must have missed that ... ? where can i find this mention since now i'm curious ...

Ah, oops, I experience some problems with github degradation, quite often in the last month it fails saving the comments. Seems I wrote it in a longer version of one of my previous comments which failed to go to github and also was not saved in the browser... Well, I'm splitting (splitted, not commited yet) backtesting for cli and non-cli parts and was stopped with #2099 ...

hroff-1902:
Ah, no, I definitely mentioned this here above:

> Thanks for fixing it. I was refactoring backtesting -- splitting it to cli and non-cli parts and stopped with this issue...

Nevermind...

xmatthias:
> Well, I'm splitting (splitted, not commited yet) backtesting for cli and non-cli parts

i'm not sure about the reason for this - i guess it's to align to edge, but the reason for that is that edge used within the bot itself (during real operations) - while backtesting is not - so honestly, i think it's effort not well spent (i think we should discuss bigger refactorings (like this) in a dedicated issue before spending time on it.

PR #2108, opened 2019-08-07
Added jupyter notebook example and doc edits

jraviotta:
## Summary
Adds a sample jupyter notebook and documentation for use.

## Quick changelog

- Edited docs/data-analysis.md for clarity and conformity with example
- Added sample notebook
- Included jupyter in setup.py and environment.yml

### DIFF
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -42,6 +42,9 @@ import os
 from freqtrade.data.history import load_pair_history
 from freqtrade.resolvers import StrategyResolver
 
+# You can override strategy settings as demonstrated below.
+# Customize these according to your needs.
+
 # Define some constants
 ticker_interval = "5m"
 # Name of the strategy class
@@ -66,7 +69,7 @@ bt_data = load_pair_history(datadir=Path(data_location),
                             pair=pair)
 
 # Confirm success
-print("Loaded " + str(len(bt_data)) + f" rows of data for {pair} from {data_location}")
+print(f"Loaded {len(bt_data)} rows of data for {pair} from {data_location}")
 ```
 
 ### Load and run strategy  
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -2,11 +2,33 @@
 
 You can analyze the results of backtests and trading history easily using Jupyter notebooks. A sample notebook is located at `user_data/notebooks/analysis_example.ipynb`. For usage instructions, see [jupyter.org](https://jupyter.org/documentation).
 
+## Example snippets
+
+### Load backtest results into a pandas dataframe
+
+```python
+# Load backtest results
+df = load_backtest_data("user_data/backtest_data/backtest-result.json")
+
+# Show value-counts per pair
+df.groupby("pair")["sell_reason"].value_counts()
+```
+
+### Load live trading results into a pandas dataframe
+
+``` python
+# Fetch trades from database
+df = load_trades_from_db("sqlite:///tradesv3.sqlite")
+
+# Display results
+df.groupby("pair")["sell_reason"].value_counts()
+```
+
 ## Strategy debugging example
 
 Debugging a strategy can be time-consuming. FreqTrade offers helper functions to visualize raw data.
 
-### Import requirements and define variables used in the script
+### Import requirements and define variables used in analyses
 
 ```python
 # Imports
@@ -47,12 +69,6 @@ print("Loaded " + str(len(bt_data)) + f" rows of data for {pair} from {data_loca
 ### Load and run strategy  
 
 * Rerun each time the strategy file is changed
-* Display the trade details. Note that using `data.head()` would also work, however most indicators have some "startup" data at the top of the dataframe.
-
-Some possible problems:
-
-* Columns with NaN values at the end of the dataframe
-* Columns used in `crossed*()` functions with completely different units
 
 ```python
 # Load strategy using values set above
@@ -60,33 +76,31 @@ strategy = StrategyResolver({'strategy': strategy_name,
                             'user_data_dir': user_data_dir,
                             'strategy_path': strategy_location}).strategy
 
-# Run strategy (just like in backtesting)
+# Generate buy/sell signals using strategy
 df = strategy.analyze_ticker(bt_data, {'pair': pair})
+```
 
+### Display the trade details
+
+* Note that using `data.head()` would also work, however most indicators have some "startup" data at the top of the dataframe.
+
+#### Some possible problems
+
+* Columns with NaN values at the end of the dataframe
+* Columns used in `crossed*()` functions with completely different units
+
+#### Comparison with full backtest
+
+having 200 buy signals as output for one pair from `analyze_ticker()` does not necessarily mean that 200 trades will be made during backtesting.
+
+Assuming you use only one condition such as, `df['rsi'] < 30` as buy condition, this will generate multiple "buy" signals for each pair in sequence (until rsi returns > 29).
+The bot will only buy on the first of these signals (and also only if a trade-slot ("max_open_trades") is still available), or on one of the middle signals, as soon as a "slot" becomes available.
+
+```python
 # Report results
 print(f"Generated {df['buy'].sum()} buy signals")
 data = df.set_index('date', drop=True)
 data.tail()
-```
-
-### Load backtest results into a pandas dataframe
-
-```python
-# Load backtest results
-df = load_backtest_data("user_data/backtest_data/backtest-result.json")
-
-# Show value-counts per pair
-df.groupby("pair")["sell_reason"].value_counts()
-```
-
-### Load live trading results into a pandas dataframe
-
-``` python
-# Fetch trades from database
-df = load_trades_from_db("sqlite:///tradesv3.sqlite")
-
-# Display results
-df.groupby("pair")["sell_reason"].value_counts()
 ```
 
 Feel free to submit an issue or Pull Request enhancing this document if you would like to share ideas on how to best analyze the data.
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -2,6 +2,8 @@
 
 You can analyze the results of backtests and trading history easily using Jupyter notebooks. A sample notebook is located at `user_data/notebooks/analysis_example.ipynb`. For usage instructions, see [jupyter.org](https://jupyter.org/documentation).
 
+*Pro tip - Don't forget to start a jupyter notbook server from within your conda or venv environment or use [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels)*
+
 ## Example snippets
 
 ### Load backtest results into a pandas dataframe
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -1,164 +1,92 @@
 # Analyzing bot data
 
-After performing backtests, or after running the bot for some time, it will be interesting to analyze the results your bot generated.
+You can analyze the results of backtests and trading history easily using Jupyter notebooks. A sample notebook is located at `user_data/notebooks/analysis_example.ipynb`. For usage instructions, see [jupyter.org](https://jupyter.org/documentation).
 
-A good way for this is using Jupyter (notebook or lab) - which provides an interactive environment to analyze the data.
+## Strategy debugging example
 
-The following helpers will help you loading the data into Pandas DataFrames, and may also give you some starting points in analyzing the results.
+Debugging a strategy can be time-consuming. FreqTrade offers helper functions to visualize raw data.
 
-## Strategy development problem analysis
-
-Debugging a strategy (are there no buy signals, ...) can be very time-consuming.
-FreqTrade tries to help you by exposing a few helper-functions, which can be very handy.
-
-It's recommended using Juptyer Notebooks for analysis, since it offers a dynamic way to rerun certain parts of the code.
-
-The following is a full code-snippet, which will be explained by both comments, and step by step below.
+### Import requirements and define variables used in the script
 
 ```python
-# Some necessary imports
+# Imports
 from pathlib import Path
-
+import os
 from freqtrade.data.history import load_pair_history
 from freqtrade.resolvers import StrategyResolver
+from freqtrade.data.btanalysis import load_backtest_data
+from freqtrade.data.btanalysis import load_trades_from_db
+
 # Define some constants
-ticker_interval = "5m"
-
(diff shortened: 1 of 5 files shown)


environment.yml
```diff
@@ -37,6 +37,7 @@ dependencies:
   - coveralls
   - mypy
   # Useful for jupyter
+  - jupyter
```
xmatthias: i am currenly "doubtful" about this, since it can fit a workstyle, or it can't:

I'm not sure if `environment.yml` offers some kind of "optional dependencies" ... but i guess this would be a good candidate for this.

The way i use jupyter is to have one central jupyter installation (in `base` - when using conda) - and add the different environments as interpreters to this central installation, so i only have one "full" jupyter installation (including all plugins and such), and the environments only have `ipykernel` installed.

Note: It's not something i object - but would like to discuss on what you think.

jraviotta: > it can fit a workstyle, or it can't:

Yes I agree. I also have a separate environment for my notebook server. Personally, I don't install Jupyter in any projects but I added it to the the installation manifests because I figured less technical users may be confused about that configuration and including it would increase the size of the environment but not impact memory unless the user launches the jupyter server.

I will defer to your decision about including the package.

docs/data-analysis.md
```diff
@@ -1,164 +1,92 @@
...
-
-The following is a full code-snippet, which will be explained by both comments, and step by step below.
+### Import requirements and define variables used in the script
 
 ```python
-# Some necessary imports
+# Imports
 from pathlib import Path
-
+import os
```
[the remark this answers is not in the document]
jraviotta: Thanks for pointing this out. I did not realize that Path was a better choice for concatenating paths. I'll refactor those variables with Path.

jraviotta:
> i guess you missed to add this file to the PR ... ?

Oops probably so. Also, what is the correct way to include package requirements using the project structure? I was guessing that new requirements go in `setup.py` and that the `.env` files are generated. Is that correct?

environment.yml
jraviotta: > I'm not sure if `environment.yml` offers some kind of "optional dependencies" ... but i guess this would be a good candidate for this.

As far as I know, Conda does not have concept of optional dependencies. Everything in `environment.yml` will be installed. If your want to support different environment specifications, you need different `environment.yml` files and then control which one is used with options in a setup script or documentation.

jraviotta:
> Note: when doing so, please add jupyter as "extras_require" ... i don't think it's needed on servers, so it should be optional.

I wonder if there is a better way to manage the multiple configurations of Freqtrade? There are many permutations among `venv`, `conda`, `docker`, local install, server installed, Raspberry Pi installed, where the `user_data` dir lives, whether to install development dependencies and more. the setup shell script has logic for many user choices, but more need to be added or the user needs to read the documentation and mentally run through a complex tree of if thans to figure out how and what to install. Do you have any inclination towards breaking this up into a more modular format? If so, how?

environment.yml
xmatthias: works for me 👍 , leave it in - as long as we don't use conda to install in docker it won't hurt.

xmatthias:
there is no `.env` file that is generated ... 

Note that once #2067 is merged, this should become clearer as we can configure a whole user-dir at once. I'll also want to move the default configurations to the user_data folder - but it's a slow, stepwise process - otherwise it's a pain/impossible to review, and error-prone.

using `extras_require` is a python packaging standard, which makes the package installable using `pip install -e .[all]` - or `pip install freqtrade[dev]`  once we're (correctly) on pip.

setup.sh shall not be required in the end and be removed ... at least that's my goal.

jraviotta:
I added jupyter to setup.py. let me know if I did that incorrectly. I wasn't sure what, if anything, the value for the key should be set to.

Also changed `os.path.join` to `Path`.

docs/data-analysis.md
```diff
@@ -1,164 +1,92 @@
...
+* Columns used in `crossed*()` functions with completely different units
+
+```python
+# Load strategy using values set above
+strategy = StrategyResolver({'strategy': strategy_name,
+                            'user_data_dir': user_data_dir,
+                            'strategy_path': strategy_location}).strategy
 
 # Run strategy (just like in backtesting)
 df = strategy.analyze_ticker(bt_data, {'pair': pair})
```
jraviotta: `analyze_ticker` does not export to `user_data/backtest_data/backtest-result.json` like the `Backtesting` class. Therefore, `df = load_backtest_data("user_data/backtest_data/backtest-result.json")` either fails or returns stale results.

What would you suggest?

xmatthias: well, `analyze_ticker()` is not a replacement for backtesting - the only thing it does is give you a dataframe with your indicators and buy/sell signals, so you can plot / analyze/ ....

I guess the comment "just like in backtesting" is misleading and maybe should be replaced with "populate indicators and buy/sell signals" - letme explain:

**example**
having 200 buy signals as output for one pair from `analyze_ticker()` (`assert df['buy'].sum() == 200`)  does not mean that 200 trades will be made during backtesting (unless --eps and --dmmp - the "unrealistic" flags) are used.

**Explanation:**
Assuming you use only one condition - `df['rsi'] < 30` as buy condition - this will generate multiple "buy" signals for each pair in sequence (until rsi returns > 29).
The bot will only buy on the first of these signals (and also only if a trade-slot ("max_open_trades") is still available), or on one of the middle signals, as soon as a "slot" becomes available.

We currently can't call backtesting from notebooks for a few reasons (it'll crash ...) - but i guess it's just a small step to enable that in the future.

setup.py
```diff
@@ -68,6 +68,8 @@
           'dev': all_extra,
           'plot': plot,
           'all': all_extra,
+          'jupyter': [],
```
xmatthias: this will not do anything ...
yes, it allows running `pip install -e .[jupyter]` - but won't install jupyter.
The correct syntax  is `<key>:[<additional packages to install>]`

```
'jupyter': ['jupyter'],
```

I'm not sure if "key" (the left side) should be replaced by "analysis" or something similar (`pip install freqtrade[plot,analysis]` vs. `pip install freqtrade[plot,jupyter]`) ... but i'll leave this up to you

jraviotta: Thanks for explaining. I see how that works. I added a section for useful packages for jupyter and keyed with 'jupyter'. I also added it to the all_extras list.

docs/data-analysis.md
jraviotta: This is ugly but works in a notebook:

```
import os
import freqtrade
import subprocess

# Configure cli command as a python list of strings
cmd = ['freqtrade', 
       '-c', Path(str(Path.home()),'.freqtrade/exchange-config.json'), 
       '-c', 'user_data/user_repo/config.json', 
       'backtesting', "--refresh-pairs-cached",
      "--timerange=-200",
      "--export", "trades",
      "--export-filename=user_data/backtest_data/backtest-result.json"
     ]

# Run command using subprocess
try:
    # If command succeeds capture output
    normal = subprocess.run(cmd,
        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
        check=True,
        text=True)
except:
    # If command fails capture shell error instead of python error
    normal = subprocess.run(cmd,
        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
        text=True)
    
# Print captured output
print(normal.stdout)
print(normal.stderr)
```

xmatthias: but it defeats the purpose of using a jupyter notebook in my eyes.
so while I'm sure it works, i'm not sure it's useful to have it documented that way, since it'll be quite confusing to users.

It's better to wait a bit and have the proper way working and documented (as import)

docs/data-analysis.md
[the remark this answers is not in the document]
xmatthias: Thanks for fixing this in the notebook. Would be great if you could add it to the documentation section too, so the 2 are aligned.

user_data/notebooks/analysis_example.ipynb
```diff
@@ -0,0 +1,240 @@
+{
```
xmatthias: The notebook works great.
3 minor things:
* the 2nd code cell (3rd cell overall) uses tabs as intention, while the rest of the notebook uses spaces.
This may lead to odd results if users start copy/pasting code around.
* I would not default interval to "1m" ... this interval is pretty unreliable on low-volume pairs, and causes a lot of "outdated ticker" in dry/live modes. "5m" is used as default throughout the bot - so i would use that here too.
* the last cell in the notebook should be markdown, not code.

docs/data-analysis.md
```diff
@@ -1,164 +1,106 @@
 # Analyzing bot data
 
-After performing backtests, or after running the bot for some time, it will be interesting to analyze the results your bot generated.
+You can analyze the results of backtests and trading history easily using Jupyter notebooks. A sample notebook is located at `user_data/notebooks/analysis_example.ipynb`. For usage instructions, see [jupyter.org](https://jupyter.org/documentation).
```
xmatthias: I think we should mention here to start the notebook-server from within the virtual environment, just to hop in front of questions that may arrise

docs/data-analysis.md
jraviotta: > It's better to wait a bit and have the proper way working and documented (as import)

I agree that introducing this to users would be confusing. Take a look at the example notebook to see how I proposed clarifying this issue until backtests can be run in a notebook.

docs/data-analysis.md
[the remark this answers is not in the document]
jraviotta: done

user_data/notebooks/analysis_example.ipynb
jraviotta: fixed

docs/data-analysis.md
jraviotta: done

docs/data-analysis.md
```diff
@@ -1,164 +1,108 @@
...
-# Location of the strategy
-strategy_location = '../xmatt/strategies'
-# Location of the data
-data_location = '../freqtrade/user_data/data/binance/'
-# Only use one pair here
-pair = "XRP_ETH"
-
-### End constants
+# Load backtest results
+df = load_backtest_data("user_data/backtest_data/backtest-result.json")
```
xmatthias: this section (and the following one about live-data) looks odd when looked at in the documentation (or within the file - it's not obvious in the github diff). 

it uses `load_backtest_data` without ever importing it. 
The imports are a few code-blocks below - but that's not obvious and forces users to jump up/down in this document.
i would either move this section below the imports, or keep the import in this code-block (it won't hurt and provides a simple copy/pasteable method if someone doesn't want to use the sample notebook).
My personal preference would be the latter ... so keeping the related import in this cell / code block.

docs/data-analysis.md
```diff
@@ -1,164 +1,108 @@
...
-# Show value-counts per pair
-df.groupby("pair")["sell_reason"].value_counts()
-
-```
-
-This will allow you to drill deeper into your backtest results, and perform analysis which otherwise would make the regular backtest-output very difficult to digest due to information overload.
+#### Comparison with full backtest
 
-If you have some ideas for interesting / helpful backtest data analysis ideas, please submit a Pull Request so the community can benefit from it.
+having 200 buy signals as output for one pair from `analyze_ticker()` does not necessarily mean that 200 trades will be made during backtesting.
```
xmatthias: That's a helpful mention :+1: users will benefit greatly from this i hope

docs/data-analysis.md
xmatthias: Yep, that description looks good :+1:

docs/data-analysis.md
```diff
@@ -1,164 +1,108 @@
...
 # Location of the data
-data_location = 'user_data/data/binance'
+data_location = Path(user_data_dir, 'data', 'binance')
+# Pair to analyze 
 # Only use one pair here
-pair = "XRP_ETH"
+pair = "BTC_USDT"
 ```
 
-This first section imports necessary modules, and defines some constants you'll probably need to adjust for your case.
```
xmatthias: I would keep more description about the codeblocks (especially the one requiring manual intervention) - yes, it's kind of documented in the code block - but at least me personally, I tend to copy/paste the codeblock "semi-blindly" at first and read the description between codeblocks - so quickly going through this i might miss that HERE i need to change something

docs/data-analysis.md
```diff
@@ -1,164 +1,108 @@
...
-                            ticker_interval = ticker_interval,
+                            ticker_interval=ticker_interval,
                             pair=pair)
-print(len(bt_data))
-```
 
-This second section loads the historic data and prints the amount of candles in the DataFrame.
-You can also inspect this dataframe by using `bt_data.head()` or `bt_data.tail()`.
+# Confirm success
+print("Loaded " + str(len(bt_data)) + f" rows of data for {pair} from {data_location}")
```
xmatthias: this is a very odd mixture of using f-strings and don't using them - i would prefer to be everything a f-string (which will also remove the need for "str()" ...

setup.py
```diff
@@ -25,7 +25,15 @@
     'pytest-random-order',
 ]
 
-all_extra = api + plot + develop
+jupyter = [
+    'jupyter',
+    'nbstripout',
+    'ipykernel',
+    'isort',
+    'yapf',
```
xmatthias: i don't think `yapf` (or any other formatter for that matter) is loaded to jupyter automatically (you'll need a plugin for that) ... so i don't think this should be installed with jupyter.
same with `isort` ..

docs/data-analysis.md
jraviotta: I see what you mean. I was trying to reflect the notebook cells in the documentation, but that doesn't exactly work, because the documentation page is not considered as a whole script the way a notebook page is.

I added the imports to the code cells in the documentation, for the first 2 snippets. But left the constants and imports together for the strategy analysis section. Also fixed a few other inconsistencies.

docs/data-analysis.md
jraviotta: Understood. I added a note to the block that requires customization.

docs/data-analysis.md
jraviotta: Indeed it is. I cleaned it up.

setup.py
jraviotta: Sure. I deleted them. I use them with nb extensions, but the user can figure that out if they choose to enable any extensions.

jraviotta:
> thanks for this, i hope many users will benefit!

No problem. Thanks for spending time coaching me. I appreciate the effort.

PR #2121, opened 2019-08-11
Allow comments in config files

hroff-1902:
...and trailing commas.

rapidjson allows this, it has corresponding parse_mode values.

A special test config file (`config_test_comments.json`) was added into the tests dir since I found it inappropriate to add those testing comments into `config_full.example.json` from the main dir.

(btw, `config_full.example.json` is the only file which is really loaded by the tests. Should not we also load in the tests the `config.json.example` after modifying it in the same manner as setup.py does?)

If accepted, TODO:
* [x] describe this in the docs.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,15 +1,24 @@
 # Configure the bot
 
-This page explains how to configure your `config.json` file.
+This page explains how to configure your configuration file.
 
-## Setup config.json
+Per default, the bot loads configuration from the `config.json` file located in the current working directory.
+You can change the configuration file used by the bot with the `-c/--config` option.
 
-We recommend to copy and use the `config.json.example` as a template
+If you used the [Quick start](installation.md/#quick-start) method for installing 
+the bot, the installation script should have already created the default configuration file (`config.json`) for you.
+
+We recommend you to copy and use the `config.json.example` as a template
 for your bot configuration.
 
-The table below will list all configuration parameters.
+The configuration file defines the set of configuration parameters for the bot written in the JSON format.
+Additionally, you may use one-line `// ...` and multi-line `/* ... */` comments.
 
-Mandatory Parameters are marked as **Required**.
+## Configuration parameters
+
+The table below will list all configuration parameters available.
+
+Mandatory parameters are marked as **Required**.
 
 |  Command | Default | Description |
 |----------|---------|-------------|
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,18 +1,28 @@
 # Configure the bot
 
-This page explains how to configure your configuration file.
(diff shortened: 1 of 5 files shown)


xmatthias:
> (btw, config_full.example.json is the only file which is really loaded by the tests. Should not we also load in the tests the config.json.example after modifying it in the same manner as setup.py does?)

it's not setup.py but `setup.sh` which does these modifications (currently) - and setup.sh is not tested at all (that's one reason i don't like it).

> If accepted, TODO:

why not ... ? i'll have a play with it later today (hopefully) - but i don't see anything blocking assuming it works (which i'm sure it does).

hroff-1902:
> why would it not be accepted ... ?

Well, you could say, for example, that json standard is json standard and config should strictly follow it...

xmatthias:
well, technically yes - but i trip upon trailing commas myself every now and then ...

also "standard" json is still beeing supported - and quickly removing / adding lines to a config has advantages that are not neglectible ... :smile:

docs/configuration.md
```diff
@@ -1,15 +1,24 @@
 # Configure the bot
 
-This page explains how to configure your `config.json` file.
+This page explains how to configure your configuration file.
```
hroff-1902: "configure your configuration", hmm

docs/configuration.md
```diff
@@ -1,15 +1,24 @@
...
 
-## Setup config.json
+Per default, the bot loads configuration from the `config.json` file located in the current working directory.
+You can change the configuration file used by the bot with the `-c/--config` option.
 
-We recommend to copy and use the `config.json.example` as a template
+If you used the [Quick start](installation.md/#quick-start) method for installing 
+the bot, the installation script should have already created the default configuration file (`config.json`) for you.
+
+We recommend you to copy and use the `config.json.example` as a template
```
xmatthias: do we? 

that's true for bittrex - but for binance i would want to use the `config_binance.json.example` file as template ... ?

hroff-1902: I only changed "recomment to copy" --> "recommend you to copy" in this sentence.

Actually I guess the installation procedure (the part which prepares the default config) should ask for exchange and use diff. template configs, if they are found in the distribution... But this is not for this PR...

docs/configuration.md
xmatthias: THAT's why we had `config.json` there i think ... :D 

maybe "This page explains how to tune your configuration file." ? (but it's not very well either)

hroff-1902: I reworked this part of the docs, tried to make it more friendly for novice users...

Pls review.

xmatthias: Nicely worked around that odd phrase! :+1:

Issue #1841, opened 2019-05-11
emergency stop against repetitive "loops" and loss


Prossi79:
## Step 2: Describe your environment

  * Python Version: Python 3.7.1
  * CCXT version: ccxt==1.18.509
  * Branch: Develop
  * Last Commit ID: c8d75fbd8ac10e2af5502bf8e849061bff1c1bb5

  
## Step 3: Describe the problem:

During live mode my bot bought and sold  7 times in a row the same pair (BTT/BTC), each times running into stop loss (-10%). I suspect that the first trade signals to open the next one. All open and close rates in the 7 trades were exactly the same ... kind of self created loop?

Luckily I was just checking my bot on telegram and stopped the bot manually, otherwise it would have continued until all my BTC balance on Binance would have eaten away!

Please see sqlite database as zip enclosed.
[strategyB.zip](https://github.com/freqtrade/freqtrade/files/3168988/strategyB.zip)

Can a kind of "emergency auto stop" be integrated in the bot? (e.g. after 5 negative trades or stop loss hitted in row it would stop automatically esp. in live mode)

PS: I have backtested the strategy and did a run in dry mode - all vey good

xmatthias:
This is a dangerous pair ... if you check the chart (https://www.binance.com/en/trade/pro/BTT_BTC) - it does "binary" moves of >10 %.

![2019-05-11-141737_1123x402_scrot](https://user-images.githubusercontent.com/5024695/57569597-8a8b1d80-73f7-11e9-8e91-c677c776fc4b.png)

Did you add this manually to the pairlist?

Prossi79:
no, thats the problem .. i worked with the dynamic pairlist  ...  :/

    "pairlist": {
        "method": "VolumePairList",
        "config": {
            "number_assets": 30,
            "sort_key": "quoteVolume",
            "precision_filter": true
        }
    },

xmatthias:
Well technically it is in the top 20ish pairs for today (depending when it's checked) - so dynamic pairlist selecting this is correct. also stoploss is still possible below - which makes this also correct.

I've never been convinced that volume alone is a good indicator for good pairs - however i currently don't have an alternative idea for automatic selections.

xmatthias:
an alternative as "emergency stop" would be `/blacklist BTT/BTC` - which will add this pair to the blacklist.

Prossi79:
[...]

PR #2131, opened 2019-08-13
Lock pairs

### LINKED ISSUE #1904: Avoid sell/buy on same candle
Yes. Post related 1850
#
https://github.com/freqtrade/freqtrade/issues/1850

Python Version: python3.7
CCXT version: 1.18.551
Branch: Master | Develop
  
Bot using 30m candle. Simple smacross. So far so good, but it seems it bought and since condition to sell was reached on the same candle it bought again after selling. 
### Steps to reproduce:

Its hard to reproduce as the sell condition has to happen on the same candle as the buy condition. 
  
### Observed Results:

What happened? It bought back after selling in profit. 
What should have happened? That it only bought once rather than immediately buying after sell

Sadly systemd is not keeping the whole log so I can't take it from there and I was not saving it on another file. 
Screenshot_20190603-012154

Binance: Buying TNT/BTC
with limit 0.00000374
(0.009040 BTC,78.713 USD)

Binance: Selling TNT/BTC
Limit: 0.00000431
Amount: 2417.00000000
Open Rate: 0.00000374
Current Rate: 0.00000431
Sell Reason: roi
Profit: 15.01% (profit: 0.00135823 BTC / 11.827 USD)

Binance: Buying TNT/BTC
with limit 0.00000431
(0.009137 BTC,79.563 USD)

Binance: Selling TNT/BTC
Limit: 0.00000388
Amount: 2120.00000000
Open Rate: 0.00000431
Current Rate: 0.00000392
Sell Reason: stoploss_on_exchange
Profit: -10.39% (loss: -0.00095014 BTC / -8.273 USD)

As you can see on timestamp it's still the same candle that it bought, so my assumption is that somehow it verifies that the condition happen before and he doesn't understand that hr already sold in profit so he buys again.

Shouldn't it turn the buy condition off in the same pair after roi or stoploss until it happens again?

xmatthias:
## Summary
In case of a sell (ROI, stoploss or sell-signal) - we should prevent buying within the same candle. This will only drive up transaction costs.
This is unrelated to the flag `process_only_new_candles`.

Closes #1904

This should enable easily fixing #1841.

## Quick changelog

- Lock pairs in case of a sell until the "current" candle is over.
- prevent checking for buys during this candle
- Add helper functions to determine next/previous candle based on timestamp and ticker-interval


freqtrade/strategy/interface.py
```diff
@@ -268,8 +279,8 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
                     sell: bool, low: float = None, high: float = None,
                     force_stoploss: float = 0) -> SellCheckTuple:
         """
-        This function evaluate if on the condition required to trigger a sell has been reached
-        if the threshold is reached and updates the trade record.
+        This function evaluate if one of the conditions required to trigger a sell
```
hroff-1902: "evaluates"

hroff-1902: Note to myself: I need to verify if lock_pairs is used in every places where it should be called.

xmatthias: the assumption is that it's called whenever a sell happens, which can be 

a) after triggering a sell for whatever reason (including force-sell)
b) after stoploss_on_exchange is filled

I don't think we have another point where sells can be triggered (but checking is better than regretting) :+1:

PR #2132, opened 2019-08-13
allow create_trade() to create multiple trades per iteration

### LINKED ISSUE #1850: process_only_new_candles not applied correctly when trades happen

xmatthias:
## Summary
Returning from the create_trade function does not make sense - if multiple "trade-slots" are open, and we have buy-signals, we should buy immediately to not loose the opportunity.

This will also implicitly fix #1850, since this will run analysis for all whitelisted pairs while checking for trades. In combination with `process_only_new_candles` and #2131, buys only happen right after the new candles appear.

Closes #1850

## Quick changelog

- `process()` functions should not return anything (as identified in TODO in worker.py), since it's only used in tests
- `process_maybe_execute_buy()` should not return anything.
- `create_trade()` should check full whitelist (until `"max_open_trades"` is reached) to create as many trades as allowed (and set by the buy-signal).
- Adapt tests (especially `edge` - which did not set max_open_trades according to it's own definition).

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -258,9 +258,10 @@ class FreqtradeBot(object):
 
     def create_trades(self) -> bool:
         """
-        Checks the implemented trading indicator(s) for a randomly picked pair,
-        if one pair triggers the buy_signal a new trade record gets created
-        :return: True if a trade object has been created and persisted, False otherwise
+        Checks the implemented trading strategy for buy-signals, using the active pair whitelist.
+        If a pair triggers the buy_signal a new trade record gets created.
+        Checks pairs as long as the open trade count is below `max_open_trades`.
+        :return: True if at least one trade has been created.
         """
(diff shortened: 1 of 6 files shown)


xmatthias:
yes, python lists are ordered by nature ... and we never mess with that for the pairlist (otherwise we would be preferring lower (or higher) alphabetical pairs).

That's not really changed here anyway, assuming all pair generated buys, we would buy the first pair in the list before this PR too, but the 2nd in the next iteration, which introduces  a delay of ~5s (assuming default settings).

this has a specific advantage for VolumePairList, which is sorted descending by volume - so we ~~trade~~ try to trade highest volume first.

PR #2138, opened 2019-08-14
Refactorings to history

xmatthias:
## Summary
Abstract storage layer a bit better so we have one method to store / load data.

## Quick changelog

- new store_tickerdata_file method
- use load_tickerdata_file also to load cached data.

### DIFF
--- a/freqtrade/tests/data/test_history.py
+++ b/freqtrade/tests/data/test_history.py
@@ -184,7 +184,8 @@ def test_load_cached_data_for_updating(mocker) -> None:
 
     # same with 'line' timeframe
     num_lines = (test_data[-1][0] - test_data[1][0]) / 1000 / 60 + 120
-    data, start_ts = load_cached_data_for_updating(datadir, 'UNITTEST/BTC', '1m', TimeRange(None, 'line', 0, -num_lines))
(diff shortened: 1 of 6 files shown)


freqtrade/tests/data/test_history.py
```diff
@@ -178,89 +178,70 @@ def test_load_cached_data_for_updating(mocker) -> None:
...
     assert test_data[-2][0] < start_ts < test_data[-1][0]
 
     # same with 'line' timeframe
     num_lines = 30
     timerange = TimeRange(None, 'line', 0, -num_lines)
-    data, start_ts = load_cached_data_for_updating(test_filename,
-                                                   '1m',
-                                                   timerange)
+    data, start_ts = load_cached_data_for_updating(datadir, 'UNITTEST/BTC', '1m', timerange)
     assert data == test_data[:-1]
```
xmatthias: most of these tests make no sense ... apparently timerange in `load_cached_data_for_updating()` was broken (and tests did not check that at all).

This line is the best example - we pass in `-29` ... and assert that we get all apart from the last item back.

I'll rethink how `load_cached_data_for_updating()` should really work - since we would always need to get the full data, not just a fraction (since in that case, we would delete the non-loaded data).

PR #2141, opened 2019-08-15
f the string

ahonnecke:
## Summary
Someone wrote an fstring without the f

## Quick changelog

- add f to the fstring to make logging print the variable

## What's new?

Output will print the actual runmode now instead of this:
```
2019-08-15 16:12:19,640 - freqtrade.configuration.configuration - INFO - Runmode set to {self.runmode}.
```

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -290,7 +290,7 @@ class Configuration(object):
         if not self.runmode:
             # Handle real mode, infer dry/live from config
             self.runmode = RunMode.DRY_RUN if config.get('dry_run', True) else RunMode.LIVE
-            logger.info("Runmode set to {self.runmode}.")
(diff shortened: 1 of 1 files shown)


hroff-1902:
Not covered by tests.
Coverall says it is. (this line is green)
I understand why, but this is awful since it means that coveralls metrics does not have much sense in practice...

xmatthias:
coveralls says this line was called by tests - not that it does what it should.

It would be cought by a `log_has()` call - however that does not make sense to add for all log messages. 
especially since then we would mostly copy/paste the string and test with the same error (i've seen this ...)

having 100% coverage does not mean that a function does what it's supposed to do, just that every line was called.
having coveralls helps with nested functions / if statements - if something was missed there is a test missing.

This problem is known with unittests in general, and that's the main reason we have `test_backtest_detail.py` - which is trying to test business logic (how do we expect the whole thing to behave) for backtesting.

PR #2142, opened 2019-08-15
Hyperopt: --print-json option

hroff-1902:
This PR adds the `--print-json` option which prints best result details in json format, in one line.

This helps automatization of backtesting, since the details in this format are easily greppable (one line, defined format) and can then be easily parsed and transferred into existing strategy or to the options of the freqtrade backtesting or freqtrade dry-run...

Example of output, spaces='all':
```
*    1/10:     15 trades. Avg profit -0.57%. Total profit -0.00429526 BTC (  -8.57Σ%). Avg duration 1287.0 mins. Objective: 2.28322
*    2/10:     60 trades. Avg profit -0.76%. Total profit -0.02279588 BTC ( -45.48Σ%). Avg duration 1638.4 mins. Objective: 2.39411
*    3/10:      0 trades. Avg profit   nan%. Total profit  0.00000000 BTC (   0.00Σ%). Avg duration   nan mins. Objective: 100000.00000
*    4/10:      0 trades. Avg profit   nan%. Total profit  0.00000000 BTC (   0.00Σ%). Avg duration   nan mins. Objective: 100000.00000
*    5/10:     14 trades. Avg profit -0.13%. Total profit -0.00092800 BTC (  -1.85Σ%). Avg duration 358.2 mins. Objective: 2.26110
*    6/10:     11 trades. Avg profit  0.00%. Total profit  0.00000000 BTC (   0.00Σ%). Avg duration 548.6 mins. Objective: 2.25574
*    7/10:      5 trades. Avg profit  0.00%. Total profit  0.00000000 BTC (   0.00Σ%). Avg duration 651.0 mins. Objective: 100000.00000
*    8/10:     17 trades. Avg profit  0.00%. Total profit  0.00000000 BTC (   0.00Σ%). Avg duration 855.0 mins. Objective: 2.25412
2019-08-15 23:14:45,645 - freqtrade.optimize.hyperopt - INFO - Saving 8 evaluations to 'user_data/hyperopt_results.pickle'

Best result:

*    8/10:     17 trades. Avg profit  0.00%. Total profit  0.00000000 BTC (   0.00Σ%). Avg duration 855.0 mins. Objective: 2.25412

{"params":{"mfi-value":"23","fastd-value":"24","adx-value":"46","rsi-value":"23","mfi-enabled":false,"fastd-enabled":false,"adx-enabled":true,"rsi-enabled":false,"trigger":"sar_reversal","sell-mfi-value":"91","sell-fastd-value":"92","sell-adx-value":"79","sell-rsi-value":"71","sell-mfi-enabled":false,"sell-fastd-enabled":true,"sell-adx-enabled":false,"sell-rsi-enabled":true,"sell-trigger":"sell-sar_reversal"},"minimal_roi":{"0":0.18231284297024036,"31":0.083870261864611,"67":0.0207858410789196,"105":0},"stoploss":-0.40745600020888995}
```

Example of output, spaces='roi stoploss':
```
*    1/10:     15 trades. Avg profit -0.54%. Total profit -0.00407874 BTC (  -8.14Σ%). Avg duration 1962.3 mins. Objective: 2.28178
*    2/10:     15 trades. Avg profit -0.54%. Total profit -0.00407874 BTC (  -8.14Σ%). Avg duration 1962.3 mins. Objective: 2.28178
*    3/10:     18 trades. Avg profit -0.60%. Total profit -0.00540115 BTC ( -10.78Σ%). Avg duration 1273.1 mins. Objective: 2.28977
*    4/10:     17 trades. Avg profit -0.48%. Total profit -0.00407874 BTC (  -8.14Σ%). Avg duration 1335.6 mins. Objective: 2.28125
*    5/10:     15 trades. Avg profit -0.54%. Total profit -0.00407874 BTC (  -8.14Σ%). Avg duration 2003.3 mins. Objective: 2.28178
*    6/10:     18 trades. Avg profit -0.54%. Total profit -0.00489156 BTC (  -9.76Σ%). Avg duration 609.2 mins. Objective: 2.28638
*    7/10:     15 trades. Avg profit -0.54%. Total profit -0.00407874 BTC (  -8.14Σ%). Avg duration 1794.3 mins. Objective: 2.28178
*    8/10:     19 trades. Avg profit -0.70%. Total profit -0.00667074 BTC ( -13.31Σ%). Avg duration 428.4 mins. Objective: 2.29794
2019-08-15 23:16:28,736 - freqtrade.optimize.hyperopt - INFO - Saving 8 evaluations to 'user_data/hyperopt_results.pickle'

Best result:

*    4/10:     17 trades. Avg profit -0.48%. Total profit -0.00407874 BTC (  -8.14Σ%). Avg duration 1335.6 mins. Objective: 2.28125

{"minimal_roi":{"0":0.16187003339696243,"38":0.04901850020517244,"76":0.02871334121790446,"89":0},"stoploss":-0.4529922721868516}
```

The names of the keys in the printed json are chosen to correspond to the name of the params dict in the hyperopts (`'params'`) and the names of the attributes in the strategies (`'minimal_roi'`, `'stoploss'`). This is by intention.

TBD: maybe this option should be named `--print-detail(s)-json` or `--print-best-json`, because it only affects output of the details of the last best epoch.

TODO (if accepted):
* [ ] Add description to the docs.
* [ ] Update hyperopt cli helpstring in the docs.

### DIFF
--- a/freqtrade/configuration/cli_options.py
+++ b/freqtrade/configuration/cli_options.py
@@ -198,6 +198,12 @@ AVAILABLE_CLI_OPTIONS = {
         action='store_false',
         default=True,
     ),
+    "print_json": Arg(
+        '--print-json',
+        help='Print best result detailization in JSON format.',
+        action='store_true',
+        default=False,
+    ),
     "hyperopt_jobs": Arg(
         '-j', '--job-workers',
         help='The number of concurrently running jobs for hyperoptimization '
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -140,23 +140,23 @@ class Hyperopt(Backtesting):
         print(f"\nBest result:\n\n{log_str}\n")
 
         if self.config.get('print_json'):
-            result_dict = {}
+            result_dict: Dict = {}
             if self.has_space('buy') or self.has_space('sell'):
                 result_dict['params'] = {}
             if self.has_space('buy'):
                 result_dict['params'].update({p.name: params.get(p.name)
-                    for p in self.hyperopt_space('buy')})
+                                             for p in self.hyperopt_space('buy')})
             if self.has_space('sell'):
                 result_dict['params'].update({p.name: params.get(p.name)
-                    for p in self.hyperopt_space('sell')})
+                                             for p in self.hyperopt_space('sell')})
             if self.has_space('roi'):
-                min_roi = self.custom_hyperopt.generate_roi_table(params)
                 # Convert keys in min_roi dict to strings because
                 # rapidjson cannot dump dicts with integer keys...
                 # OrderedDict is used to keep the numeric order of the items
                 # in the dict.
-                min_roi = OrderedDict((str(k),v) for k,v in min_roi.items())
-                result_dict['minimal_roi'] = min_roi
+                result_dict['minimal_roi'] = OrderedDict(
+                    (str(k), v) for k, v in self.custom_hyperopt.generate_roi_table(params).items()
+                )
             if self.has_space('stoploss'):
                 result_dict['stoploss'] = params.get('stoploss')
             print(rapidjson.dumps(result_dict, default=str, number_mode=rapidjson.NM_NATIVE))
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -8,10 +8,13 @@ import logging
 import os
 import sys
 
+from collections import OrderedDict
 from operator import itemgetter
 from pathlib import Path
 from pprint import pprint
 from typing import Any, Dict, List, Optional
+
+import rapidjson
 
 from colorama import init as colorama_init
 from colorama import Fore, Style
@@ -133,22 +136,44 @@ class Hyperopt(Backtesting):
         results = sorted(self.trials, key=itemgetter('loss'))
         best_result = results[0]
         params = best_result['params']
-
         log_str = self.format_results_logstring(best_result)
         print(f"\nBest result:\n\n{log_str}\n")
-        if self.has_space('buy'):
-            print('Buy hyperspace params:')
-            pprint({p.name: params.get(p.name) for p in self.hyperopt_space('buy')},
-                   indent=4)
-        if self.has_space('sell'):
-            print('Sell hyperspace params:')
-            pprint({p.name: params.get(p.name) for p in self.hyperopt_space('sell')},
-                   indent=4)
-        if self.has_space('roi'):
-            print("ROI table:")
-            pprint(self.custom_hyperopt.generate_roi_table(params), indent=4)
-        if self.has_space('stoploss'):
-            print(f"Stoploss: {params.get('stoploss')}")
+
+        if self.config.get('print_json'):
+            result_dict = {}
+            if self.has_space('buy') or self.has_space('sell'):
+                result_dict['params'] = {}
+            if self.has_space('buy'):
+                result_dict['params'].update({p.name: params.get(p.name)
+                    for p in self.hyperopt_space('buy')})
+            if self.has_space('sell'):
+                result_dict['params'].update({p.name: params.get(p.name)
+                    for p in self.hyperopt_space('sell')})
+            if self.has_space('roi'):
+                min_roi = self.custom_hyperopt.generate_roi_table(params)
+                # Convert keys in min_roi dict to strings because
+                # rapidjson cannot dump dicts with integer keys...
+                # OrderedDict is used to keep the numeric order of the items
+                # in the dict.
+                min_roi = OrderedDict((str(k),v) for k,v in min_roi.items())
+                result_dict['minimal_roi'] = min_roi
+            if self.has_space('stoploss'):
+                result_dict['stoploss'] = params.get('stoploss')
+            print(rapidjson.dumps(result_dict, default=str, number_mode=rapidjson.NM_NATIVE))
+        else:
+            if self.has_space('buy'):
+                print('Buy hyperspace params:')
+                pprint({p.name: params.get(p.name) for p in self.hyperopt_space('buy')},
+                       indent=4)
+            if self.has_space('sell'):
+                print('Sell hyperspace params:')
+                pprint({p.name: params.get(p.name) for p in self.hyperopt_space('sell')},
+                       indent=4)
+            if self.has_space('roi'):
+                print("ROI table:")
+                pprint(self.custom_hyperopt.generate_roi_table(params), indent=4)
+            if self.has_space('stoploss'):
+                print(f"Stoploss: {params.get('stoploss')}")
 
     def log_results(self, results) -> None:
         """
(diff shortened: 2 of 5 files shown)


hroff-1902:
"We need more gold!" 😄 
-- caveralls says I need to write more tests for hyperopt...

xmatthias:
i'm not very sure about the usecase for this ...
yes, it's oneline, greppable (but also very hard to read for humans, since it's all compressed into one line).

There is no automatism to transfer this to a strategy - and i doubt that's something we can have which works reliably (and if we would have something like that, then - the json should be stored, not printed) ...
Remember that the samples we use assume simple crossovers for the most part.

Once you start to mixin informative pairs (or other, complex stuff) - this will become a mess very soon, since people would expect this hypothetical automatism to work for every edgecase (and hyperopt/strategy code is not that similar for the most part).

Now i'm not completely against this (need to think about it a bit more) .. i just don't see a good usecase for this.

hroff-1902:
JSON by definition is for machines, not for humans -- that's why we (and all who selected json as a format for configuration) have problems with usability of the config...

It's not necessary to be 'stored' -- in theory, json are frequently passed there and back via network sockets. And, strictly speaking, "everything in unix is file", even streams, so it's stored in a file -- the output stream, which is printed by terminal...

The usecase here is the following (however, I would not disclose here the upcoming enhancements, but, anyway, that's how do I see it; now it's also achievable with changing a template of the strategy file with sed/awk to substitute the parameters/min_roi/stoploss):
```
result = `freqtrade hyperopt --print-json | grep for json result`
freqtrade backtesting --strategy-params $result
```
We have a file with all results, assume we introduce a subcommand allowing to see them. It should also have same option `--print-json`, for the same reason:
```
freqtrade backtesting --strategy-params `freqtrade show-hyperopt-result -n 186 --print-json --quiet`
```

And even run dry-run:
```
freqtrade -c dry-run-config (or dry-run subcommand here) --strategy-params `freqtrade hyperopt --print-json --quiet`
```
```
freqtrade -c dry-run-config (or dry-run subcommand here) `freqtrade show-hyperopt-result -n 186 --print-json --quiet`
```

So, it's for simplyfying automatization cases, it's for machines, for piping, grepping and automatic substitution, not for humans.

xmatthias:
i was imagining that you'd be thinking about something like that - but the way (using backtics (`) to run execution) is not working cross-platform (windows does not support that but requires different syntax).

Also, it would be better to simply `freqtrade hyperopt_to_strat -n 186` - which does not require outputting it, but reads it from pickle (or another format) directly.

The problem with this is that you assume you have a matching template strategy available - which may be true for DefaultHyperopt - but as soon as i introduce an indicator you are not expecting (heikinashi candles, for example - which is in the Default strategy but not used in the buy/sell methods) that method will fail. Same with custom imports (i may be importing `ta-library-xyz` in hyperopt ...) - so you build a strategy, which won't work without understanding it anyway.

We can think of building a "strategy_from_hyperopt" method ... a first step would be to output the following instead of json ... which will simplify copy/pasting this to a strategy manually (For now).

```
(
 (dataframe['rsi'] < 35) &
(dataframe['fastd'] < 35) &
(dataframe['adx'] > 30) &
(dataframe['plus_di'] > 0.5)
)
```

freqtrade/configuration/cli_options.py
```diff
@@ -198,6 +198,12 @@ def __init__(self, *args, **kwargs):
         action='store_false',
         default=True,
     ),
+    "print_json": Arg(
+        '--print-json',
```
xmatthias: i am not sure about this argument naming

wouldn't `--print-result-format json` be more flexible (so we can add `csv` / `strategycode` / whatever other format) later?

hroff-1902: I was also thinking of this kind of option, but my approach is that if (when) `--print-result-format=json/csv/python/whatever` is introduced, the `--print-result-json`, `--print-result-csv`, etc. options should be left. The user will be able to choose more convenient approach to specify format. In this case in the description of `--print-result-json` it may be stated that it's equivalent to `--print-result-format=json`. This approach (`--format=val` and corresponding one-argument shortcut option) can frequently be found in utilities with wide set of options.

For now I would like to leave this option, I can rename it to `--print-result-json`.

xmatthias: nah, what you got is fine :+1: 

makes sense to eventually extend to "result_format"...

freqtrade/optimize/hyperopt.py
```diff
@@ -133,22 +136,44 @@ def log_trials_result(self) -> None:
...
-            print('Sell hyperspace params:')
-            pprint({p.name: params.get(p.name) for p in self.hyperopt_space('sell')},
-                   indent=4)
-        if self.has_space('roi'):
-            print("ROI table:")
-            pprint(self.custom_hyperopt.generate_roi_table(params), indent=4)
-        if self.has_space('stoploss'):
-            print(f"Stoploss: {params.get('stoploss')}")
+
+        if self.config.get('print_json'):
```
xmatthias: For the future: 
this should probably be extracted to a function - so we can call it with "any" result - but i guess that's going to be part of the `print-result -n` PR

freqtrade/tests/optimize/test_hyperopt.py
```diff
@@ -618,3 +618,77 @@ def test_continue_hyperopt(mocker, default_conf, caplog):
 
     assert unlinkmock.call_count == 0
     assert log_has(f"Continuing on previous hyperopt results.", caplog)
+
+
+def test_print_json_spaces_all(mocker, default_conf, caplog, capsys) -> None:
+    dumper = mocker.patch('freqtrade.optimize.hyperopt.dump', MagicMock())
+    mocker.patch('freqtrade.optimize.hyperopt.load_data', MagicMock())
+    mocker.patch(
```
xmatthias: you could combine these 3 options using `mocker.patch.multiple` ... but again, that's for the future :+1:

hroff-1902:
> i was imagining that you'd be thinking about something like that - but the way (using backtics (`) to run execution) is not working cross-platform (windows does not support that but requires different syntax).

Well, it can be kept in an env, variable (see above, similar in powershell/windows cmd, piped, stored in a temp. file, etc.

> Also, it would be better to simply `freqtrade hyperopt_to_strat -n 186` - which does not require outputting it, but reads it from pickle (or another format) directly.

Yes, but how will you find this `186`? By an external tool? What if external tool is a (quite simple) shell scripting with piping and parameters substitution? Are you going to repeat same logic for finding the best result in the external script? With an upcoming `freqtrade show-hyperopt-results` subcommand? Yes, but why not to print it here *in the same machine-readable format* if this result was printed anyway?

> **edit**: moving most of this to #1775 to keep it somewhere "visible" ...

That's the different story, I have a *long* list of enhancements for this... 😆

freqtrade/tests/optimize/test_hyperopt.py
hroff-1902: I just copied it from another test above changing the options passed to hyperopt and the asserts for output -- still not very familiar with all these mocking magic...

PR #1989, opened 2019-07-01
List pairs (markets) subcommand(s)

hroff-1902:
Examples of usage (first part of logs cleared for readability):

```
$ ./freqtrade/main.py list-pairs --print-list --quote-currency USDT
...
2019-07-01 14:40:00,347 - freqtrade.exchange.exchange - INFO - Using Exchange "Bittrex"
Exchange Bittrex has 37 pairs with USDT as quote currency: ['ADA/USDT', 'ATOM/USDT', 'BAT/USDT', 'BCH/USDT', 'BSV/USDT', 'BTC/USDT', 'BTT/USDT', 'DAI/USDT', 'DASH/USDT', 'DCR/USDT', 'DGB/USDT', 'DOGE/USDT', 'ENJ/USDT', 'ETC/USDT', 'ETH/USDT', 'GRIN/USDT', 'LTC/USDT', 'NEO/USDT', 'NPXS/USDT', 'NXT/USDT', 'OCEAN/USDT', 'OGO/USDT', 'OMG/USDT', 'ONT/USDT', 'PAX/USDT', 'PMA/USDT', 'RVN/USDT', 'SC/USDT', 'TRX/USDT', 'TUSD/USDT', 'VDX/USDT', 'XLM/USDT', 'XMR/USDT', 'XRP/USDT', 'XVG/USDT', 'ZEC/USDT', 'ZRX/USDT'].
```

```
$ ./freqtrade/main.py list-pairs --exchange bitmex --print-list --quote-currency USDT
...
2019-07-01 14:41:10,136 - freqtrade.exchange.exchange - INFO - Using Exchange "BitMEX"
Exchange BitMEX has 0 pairs with USDT as quote currency.
```

```
$ ./freqtrade/main.py list-pairs --exchange bitmex --print-list --quote-currency USD
...
2019-07-01 14:41:16,432 - freqtrade.exchange.exchange - INFO - Using Exchange "BitMEX"
Exchange BitMEX has 2 pairs with USD as quote currency: ['BTC/USD', 'ETH/USD'].
```

```
$ ./freqtrade/main.py list-pairs --exchange bitmex --quote-currency USD
...
2019-07-01 14:43:05,422 - freqtrade.exchange.exchange - INFO - Using Exchange "BitMEX"
| Id     | Symbol   | Base   | Quote   | Active   |
|:-------|:---------|:-------|:--------|:---------|
| XBTUSD | BTC/USD  | BTC    | USD     | Yes      |
[...]

PR #2146, opened 2019-08-16
Download module

xmatthias:
## Summary
Convert download-script to freqtrade subcommand

## Quick changelog

- add `freqtrade download-data` cli option
- add tests for this
- allow no-config subcommands to use Arguments() without forcing a configuration.
- use space-seperation for pairs argument (using `nargs` - avoid manual splitting)

## Open questions
* [x] argument `--exchange` - this argument can cause more harm than what it solves, since it rate-limits will not be applied correctly. This is problematic since we switched to async to download data (wasn't really a problem before). If this is removed, then the ~~download-script~~ download-command must require a configuration file (which i would not mind)
* [x] remove download_backtest_data.py - or replace it with a one-line script printing what to use now??

## TODO
* [x] remove download_backtest_data.py


hroff-1902:
> remove download_backtest_data.py - or replace it with a one-line script printing what to use now??

I like the idea with replacement 👍

hroff-1902:
> argument `--exchange` - this argument can cause more harm than what it solves, since it rate-limits will not be applied correctly. This is problematic since we switched to async to download data (wasn't really a problem before). If this is removed, then the ~~download-script~~ download-command must require a configuration file (which i would not mind)

That's why I postponed #1989 -- I also needed to clarify the behavior with `--exchange` and load/dont-load default config when no `-c` is specified...

For that `list-pairs` subcommand using `--exchange` looks logical. But I would like to see common behavior between freqtrade subcommands...

freqtrade/configuration/arguments.py
```diff
@@ -40,6 +40,8 @@
 ARGS_PLOT_PROFIT = (ARGS_COMMON + ARGS_STRATEGY +
                     ["pairs", "timerange", "export", "exportfilename", "db_url", "trade_source"])
 
+NO_CONF_REQURIED = ["start_download_data"]
```
hroff-1902: It may be a list of RunMode's (RunMode.DownloadData, RunMode.ListPairs etc.) corresponding to subcommands and, therefore, to the start_* functions. This just a note, I'm not sure if this is better than this list...

freqtrade/configuration/arguments.py
```diff
@@ -119,3 +124,11 @@ def _build_subcommands(self) -> None:
...
         list_exchanges_cmd.set_defaults(func=start_list_exchanges)
         self._build_args(optionlist=ARGS_LIST_EXCHANGES, parser=list_exchanges_cmd)
+
+        # Add download-data subcommand
+        download_data_cmd = subparsers.add_parser(
+            'download-data',
+            help='Download backtesting data.'
+        )
+        download_data_cmd.set_defaults(func=start_download_data)
+        self._build_args(optionlist=ARGS_DOWNLOADER, parser=download_data_cmd)
```
hroff-1902: Imho it's better to use a unique schema for naming these optionlists -- so ARGS_DOWNLOAD_DATA would be better. Just imho.

hroff-1902:
I would like to get rid of that ugly `pairs.json`. For me, it's terribly awkward and uncomfortable. Just get whitelist from the config. If it's not set -- get the list of all pairs with stake_currency as the quote currency, Borrow the `get_markets()` method from #1989 for that, for example... If no --config is set -- yes, the `--stake-currency` option is to be used in this case... Or something like that...

freqtrade/configuration/arguments.py
xmatthias: Yes i agree in theory - but then we  still need a mapping of runmode -> start* function - since the only thing available at the point of the check is the function name that's called (`func.__name__`).

Remember that this is before we have config loaded - so not much usefull will be available...

xmatthias:
> I would like to get rid of that ugly `pairs.json`

I'm not sure about this and think it'll take away quite some functions - i use it quite extensively since it's a very easy to generate list (that's what i was hoping for from #1989), without the need to copy 200 pairs to the configuration - which is a mess to get 100% right on a server (indentation, wrong quotes, ...).

More importantly, whitelist does not allow for pairs with different stake currencies ...
so assuming i download the same data (50 top pairs against ETH/BTC/USDT) every week (only new data - to keep uptodate datasets) - i'll need to either specify multiple commands `--stake ETH` `--stake BTC`, `--stake USDT` (which downloads ALL pairs, not just the 50 i'm interrested in, or have multiple configurations, one per stake).


generating `pairs.json` on the other hand is as simple as
```python
x = ['ETH/BTC', 'XRP/BTC', 'XRP/USDT', ...]
with open("pairs.json) as f:
    json.dump(x, f)
```

Also, beause whitelist can be dynamic, you don't have to specify any pair in the configuration whitelist - so you'd need a configuration specific for downloading.

> For that `list-pairs` subcommand using `--exchange` looks logical. But I would like to see common behavior between freqtrade subcommands...

list-pairs does 1 or 2 calls against the database (rateLimit is probably not relevant in this case) - download-data on the other hand may run for multiple hours, and get you blocked within minutes ...

hroff-1902:
Okay, let's leave `pairs.json` for now, but the fact that I need to create it everytime I would like to try the bot with a new exchange seems to be awful for me. Maybe my usecases differ significantly from usecases of an ordinal user, I may suppose this, no problem.

So let's leave behavior here as it is, moving the script to the subcommand.

docs/backtesting.md
```diff
@@ -3,7 +3,7 @@
 This page explains how to validate your strategy performance by using
 Backtesting.
 
-## Getting data for backtesting / hyperopt
+## Getting data for backtesting and hyperopt
 
 To download backtesting data (candles / OHLCV), we recommend using the `freqtrade download-data` command.
```
hroff-1902: "You can download historical data (candles / OHLCV) for backtesting and hyperoptimization by means of the `freqtrade download-data` command."

hroff-1902:
LGTM, approved it, further enhancements can always be made in the future
👍

xmatthias:
> Okay, let's leave pairs.json for now, but the fact that I need to create it everytime I would like to try the bot with a new exchange seems to be awful for me. Maybe my usecases differ significantly from usecases of an ordinal user, I may suppose this, no problem.

you don't need to - it's one option.
using `--pairs ETH/BTC XRP/ETH` will work just as well without defining anything - or adding it to the configuration for this new exchange (which you would need to generate manually, too).
Also, if you use "common" pairs - a pairfile which works for most exchaneges could be passed in as `--pairs-file user_data/common_pairs.json` ...

docs/backtesting.md
```diff
@@ -5,7 +5,7 @@ Backtesting.
 
 ## Getting data for backtesting and hyperopt
 
-To download backtesting data (candles / OHLCV), we recommend using the `freqtrade download-data` command.
+To download backtesting data (candles / OHLCV) and hyperoptimization using the `freqtrade download-data` command.
```
hroff-1902: "use", otherwise predicate is missing in this sentence...

PR #2155, opened 2019-08-17
split example notebooks

jraviotta:
## Summary
This PR splits the example notebooks in two for extendability.

## Quick changelog

- added an example snippets notebook and strategy example notebook
- updated jupyter documentation for clarity


docs/data-analysis.md
```diff
@@ -1,30 +1,96 @@
...
+## Pro tips  
 
-## Example snippets
+* See [jupyter.org](https://jupyter.org/documentation) for usage instructions.
+* Don't forget to start a jupyter notbook server from within your conda or venv environment or use [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels)*
+* Copy the example notebook so your changes don't get clobbered with the next freqtrade update.
+
+## Fine print  
+
+Some tasks don't work especially well in notebooks. For example, anything using asyncronous exectution is a problem for Jupyter. Also, freqtrade's primary entry point is the shell cli, so using pure python in a notebook bypasses arguments that provide required parameters to functions.
```
xmatthias: > Also, freqtrade's primary entry point is the shell cli, so using pure python in a notebook bypasses arguments that provide required parameters to functions.

I'm not sure what you refer to with this sentence ...
every cli parameter can be in the configuration file too ... ?

docs/data-analysis.md
```diff
@@ -33,55 +99,78 @@ df.groupby("pair")["sell_reason"].value_counts()
...
+from freqtrade.data.history import load_pair_history
+
+candles = load_pair_history(datadir=data_location,
                             ticker_interval=ticker_interval,
                             pair=pair)
 
 # Confirm success
-print(f"Loaded {len(bt_data)} rows of data for {pair} from {data_location}")
+print("Loaded " + str(len(candles)) + f" rows of data for {pair} from {data_location}")
+display(candles.head())
```
xmatthias: What's the advantage of using `display()` as oposed to using `candles.head()` directly?

it limits us to jupyter notebooks, while the scripts would otherwise work in a REPL as well.

docs/data-analysis.md
```diff
@@ -90,31 +179,27 @@ print(f"Loaded {len(bt_data)} rows of data for {pair} from {data_location}")
...
                             'strategy_path': strategy_location}).strategy
 
 # Generate buy/sell signals using strategy
-df = strategy.analyze_ticker(bt_data, {'pair': pair})
+df = strategy.analyze_ticker(candles, {'pair': pair})
 ```
 
 ### Display the trade details
 
 * Note that using `data.head()` would also work, however most indicators have some "startup" data at the top of the dataframe.
```
xmatthias: I would add a comment here suggesting to use `df.tail()` instead of `data.head()`.

Oddly enough, this command is part of the notebook, but doesn't appear in this code section.

docs/data-analysis.md
```diff
@@ -1,30 +1,96 @@
...
+1. Use the CLI to
+    * download historical data
+    * run a backtest
+    * run with real-time data
+    * export results  
+
+1. Collect these actions in shell scripts
+    * save complicated commands with arguments
+    * execute mult-step operations  
+    * automate testing strategies and prepareing data for analysis
```
xmatthias: preparing (?)

docs/data-analysis.md
```diff
@@ -1,30 +1,96 @@
...
+## Fine print  
+
+Some tasks don't work especially well in notebooks. For example, anything using asyncronous exectution is a problem for Jupyter. Also, freqtrade's primary entry point is the shell cli, so using pure python in a notebook bypasses arguments that provide required parameters to functions.
+
+## Recommended workflow  
+
+| Task | Tool |  
+  --- | ---  
+Bot operations | CLI  
+Repetative tasks | shell scripts
```
xmatthias: Repetitive tasks

docs/data-analysis.md
```diff
@@ -1,30 +1,96 @@
...
+```
+
+### Watch project for changes to code
+
+This scans the project for changes to code before Jupyter runs cells.
+
+```python
+# Reloads local code changes
+%load_ext autoreload
+%autoreload 2
```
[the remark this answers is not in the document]
jraviotta: Good feedback, I'll remove from the documentation and snippets.

docs/data-analysis.md
```diff
@@ -1,30 +1,96 @@
...
-You can analyze the results of backtests and trading history easily using Jupyter notebooks. A sample notebook is located at `user_data/notebooks/analysis_example.ipynb`. For usage instructions, see [jupyter.org](https://jupyter.org/documentation).
+You can analyze the results of backtests and trading history easily using Jupyter notebooks. Sample notebooks are located at `user_data/notebooks/`.  
 
-*Pro tip - Don't forget to start a jupyter notbook server from within your conda or venv environment or use [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels)*
+## Pro tips  
 
-## Example snippets
+* See [jupyter.org](https://jupyter.org/documentation) for usage instructions.
+* Don't forget to start a jupyter notbook server from within your conda or venv environment or use [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels)*
+* Copy the example notebook so your changes don't get clobbered with the next freqtrade update.
```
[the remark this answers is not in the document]
jraviotta: > I'm also torn about the benefit of the snippets_example notebook
Good point. I wanted to put the strategy analysis example in its own notebook. That left the snippets by themselves which is pointless especially considering they can be used in regular scripts. I think the notebook can be deleted.

> do you "auto"-generate the .md documents
Nope, I just type the md and let github do the formatting.

docs/data-analysis.md
```diff
@@ -1,30 +1,96 @@
...
 
-You can analyze the results of backtests and trading history easily using Jupyter notebooks. A sample notebook is located at `user_data/notebooks/analysis_example.ipynb`. For usage instructions, see [jupyter.org](https://jupyter.org/documentation).
+You can analyze the results of backtests and trading history easily using Jupyter notebooks. Sample notebooks are located at `user_data/notebooks/`.  
 
-*Pro tip - Don't forget to start a jupyter notbook server from within your conda or venv environment or use [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels)*
+## Pro tips  
 
-## Example snippets
+* See [jupyter.org](https://jupyter.org/documentation) for usage instructions.
+* Don't forget to start a jupyter notbook server from within your conda or venv environment or use [nb_conda_kernels](https://github.com/Anaconda-Platform/nb_conda_kernels)*
```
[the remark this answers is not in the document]
jraviotta: > we should add a quickstart command people can run easily like jupyter notebook,
How would you suggest implementing? It could get complicated since the user may have a venv, conda environment, jupyter in a base environment, etc.

docs/data-analysis.md
```diff
@@ -33,55 +99,78 @@ df.groupby("pair")["sell_reason"].value_counts()
...
+from pathlib import Path
+from freqtrade.data.history import load_pair_history
+
+candles = load_pair_history(datadir=data_location,
                             ticker_interval=ticker_interval,
                             pair=pair)
 
 # Confirm success
-print(f"Loaded {len(bt_data)} rows of data for {pair} from {data_location}")
+print("Loaded " + str(len(candles)) + f" rows of data for {pair} from {data_location}")
```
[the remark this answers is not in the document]
jraviotta: Copy paste fail. Fixed it.

docs/data-analysis.md
```diff
@@ -33,55 +99,78 @@ df.groupby("pair")["sell_reason"].value_counts()
...
 ``` python
+# Load config from multiple files
 from freqtrade.configuration import Configuration
 config = Configuration.from_files(["config1.json", "config2.json"])
-print(config)
-```
 
-## Strategy debugging example
+# Show the config in memory
+import json
```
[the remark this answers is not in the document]
jraviotta: fixed

docs/data-analysis.md
jraviotta: If one wants to run the primary functions, you are correct, everything can be specified in the config. However, if one wants to use helper functions, many of the parameters need to be passed in with the function call.

For example, I have been trying to use some of the built-in plotting logic and to display the plots in a notebook. 

If i use the plotting script with `analyse_and_plot_pairs(config)` I hit the asynchronous bug. That is triggered by `init_plotscript(config)` in `ExchangeResolver`. Since that fails, The location of the downloaded data `user_data/data/binance` is unknown. But there is no easy way to pass in the exchange name using pure python from a notebook. I could fix that in the source but it is one of many places where using a helper directly, means passing in an object that would have already been created automatically elsewhere. It might be useful to add `*args`, `**kwargs` to some of the helpers.

Plus I want to see the dataframe I intend to plot in the cell. So I really want to run most of `plotting.py` in separate cells to display the output of each function for analysis. Doing that means that I need to pass the output from one function to another.

```python
from freqtrade.data.history import load_pair_history

# Load data using values from strategy and config.
# Retrieve  from config dict with python indexing syntax or override here
pair = "BTC_USDT"
ticker_interval = '1m'
datadir = Path('user_data', 'data', config['exchange']['name'])
bt_data = load_pair_history(pair=pair, ticker_interval=ticker_interval,
                            datadir=datadir, refresh_pairs=False)

# Confirm success
print(f"Loaded {len(bt_data)} rows of data for {pair} from {datadir}")

# Load strategy using values set above
strategy = StrategyResolver({'strategy': strategy_name,
                            'user_data_dir': "user_data",
                            'strategy_path': strategy_location}).strategy

# Run strategy
df = strategy.analyze_ticker(bt_data, {'pair': pair})

# Report results
print(f"Generated {df['buy'].sum()} buy signals")
```

The primary objective of using a notebook is to visualize the intermediate steps for troubleshooting and analysis, whereas the primary goal of the cli is to automate all of those steps. In the example above, I'm setting `pair` manually because I want to see the plot for one pair rather than scripting the generation of all the pairs from my whitelist. Similarly, I'm also setting `ticker_interval` rather than using the config, because I may want to run the cell for a number of intervals to see the results without having to edit a separate file.

Does that make sense?

I added some clarifying text to the documentation.

docs/data-analysis.md
jraviotta: It's a habit. In this case, I can remove `display()` but I typically call `display()` explicitly for the following reasons:
* one can set an option to turn off the printing of output that is not wrapped in `print()` or `display()` If that is set, a user may not realize that output is supposed to show.
* Jupyter notebooks only print the last implied print statement in a cell. So, 

```python
# This only prints the tail
df.head()
df.tail()

# this prints the head and tail
display(df.head())
display(df.tail())
```
* `display()` pretty-prints the object with the Jupyter styling system whereas implied print statements like `df.head()` or print statements like `print(df.head())` produce console output that often mess up line endings, embedded hrefs and other elements that are formatted by Jupyter styling.

docs/data-analysis.md
xmatthias: Well, it always depends on the usage / entrypoint you're using.

the plotting example makes sense - but i think you're using the wrong entrypoint.
using `generate_candlestick_graph()` will work for what you're probably trying to do - and i know it works since i used that 2 weeks ago for some debugging.

However, everything within `plotting.py` is not documented "on purpose", since the plotting interface is not stable nor finalized yet (it's currently mostly copy/paste from the scripts- which had a lot of duplicate code).
This is, because i want to test the jupyter interfacing first, and probably improve usability of the interface before adding documentation about it.

Remember that freqtrade is a cli-first application - focused primarily on allowing trading / backtesting, not data analysis. Providing easy interfaces while remaining stable and performant is not easy - and focus will therefore remain on the cli-first approach (although i'm not against facilitating / simplifying analysis at all).

Side note: I'm a huge fan of jupyter and use it for many things (even though i may come across against it from time to time).

docs/data-analysis.md
xmatthias: I wasn't aware of the following usecase - incorporated that to my workflow now ... 

``` python
display(df.head())
display(df.tail())
```

maybe leave `display()` in the notebook, but remove it here (even though that will create a disconnect ...)?

docs/data-analysis.md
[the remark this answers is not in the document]
xmatthias: > Don't forget to start a jupyter notbook server from within your conda or venv environment or use 

Assuming there's a novice jupyter user - knowing to run `jupyter notebook` | `jupyter lab`  from the activated environment will suffice to get that user going.

For more in-depth questions, refer to the jupyter documentation... 

If you have jupyter in the base environment and use nb_kernels you're an advanced user to me who does not need our quickstarter about this ... (it's about giving a quick-start - not providing jupyter documentation for everything)

docs/data-analysis.md
jraviotta: Yes I completely understand. That was the intention of the comment

> Also, freqtrade's primary entry point is the shell cli, so using pure python in a notebook ...

I can tell that the Jupyter integration in undeveloped. I'm hoping to move that forward. It seems like, visualizing the backtest data and plotting is good use case to start with. Do you agree?

Is there a more effective way for us to collaborate. I don't want to step on toes and you obviously have a much better perspective on the project goals and milestones. Right now, I'm just walking through the process of using freqtrade as a new user and trying to improve features that I personally find helpful (while expanding my Python skills along the way). Also, I'm prioritizing documentation as you suggested.

docs/data-analysis.md
jraviotta: Sure. It this or the f-strings show up again, let me know :).

docs/data-analysis.md
xmatthias: yes, plotting and visualizing data is under-developed - not just in the context of jupyter.

I'll try to get the plot-script integration done soon, so we can stabilize the `plotting.py` interface for interactive usecases (feel free to play with it now - it's not documented and may change in the near future - but i hope you've got some good inputs, too).

i guess slack might work to discuss stuff quickly - although i think we're in quite different timezones, so we'll see if/how timings will work.

PR #2158, opened 2019-08-18
Config consistency checking improvements

### LINKED ISSUE #1196: Followup to asyncio PR
During the asyncio PR #1101, a few points where pointed out by @jblestang (thanks!) which where not taken on during the PR as they are general improvements, not directly related to asyncio (even though the affected code segments where touched).

This issue is to keep track (and not forget) the issues pointed out:

* [x] `load_markets()` should be in `init_ccxt`, not `validate_pairs` (pr #1204 available)
* [x] `_LIMIT = 500` in `_async_get_history` should not be hardcoded (and probably increased to 1000 as that seems to be in the cache on most exchanges) (kind of solved - it's adjustable by exchange in a configuration now).
* [x] sorting of candle data (in `_async_get_history` and it's non-async counterpart) should not happen if not necessary. (Idea: compare first with last timestamp to determine correct sorting)
* [x] remove `get_candle_history` - it's not called anywhere apart from tests (but keep / reuse some of the test-logic for async!)
* [x] check for incompatible parameter (dynamic whitelist + edge), ... [from edge] #2158

xmatthias:
## Summary
Validate edge_configuration vs. Volume pairlist - the 2 are incompatible (edge will always override volume pairlist - which will cause unnecessary calls to the exchange from the VolumePairList.

Closes #1196 (5th and last point).

## Quick changelog

- move validation methods out of Configuration
- Check configuration after loading strategy (strategies can set options, especially related trailing stop - so our current approach is inefficient for these cases).

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -6,10 +6,11 @@ import warnings
 from argparse import Namespace
 from typing import Any, Callable, Dict, List, Optional
 
-from freqtrade import OperationalException, constants
+from freqtrade import constants
 from freqtrade.configuration.check_exchange import check_exchange
 from freqtrade.configuration.create_datadir import create_datadir
-from freqtrade.configuration.json_schema import validate_config_schema
+from freqtrade.configuration.config_validation import (validate_config_schema,
+                                                       validate_config_consistency)
 from freqtrade.configuration.load_config import load_config_file
 from freqtrade.loggers import setup_logging
 from freqtrade.misc import deep_merge_dicts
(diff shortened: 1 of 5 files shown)


freqtrade/configuration/configuration.py
```diff
@@ -87,6 +86,8 @@ def load_config(self) -> Dict[str, Any]:
 
         self._process_runmode(config)
 
+        validate_config_consistency(config)
```
hroff-1902: Why should we call it twice (here and in the `load_config()`)?

hroff-1902:
Just a question: how is it related to asyncio (as stated in the #1196)?

xmatthias:
> Just a question: how is it related to asyncio (as stated in the #1196)?

It really isn't. due to the huge change #1101 was, this was detected as a potential problem so we added it as follow up to that PR (this PR was already too big to review efficiently) ... 
There was no other issue for it - and it's a minor thing which didn't really deserve a seperate issue, just as the other 4 points.

PR #2170, opened 2019-08-22
Fix documentation boxes

xmatthias:
## Summary
Fix documentation boxes (and add a CI job checking they don't break again).

## Quick changelog

- Add job to fail when this syntax error appears (sample job: https://travis-ci.org/freqtrade/freqtrade/builds/575173244)
- Fix boxes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,11 @@ jobs:
       - cp config.json.example config.json
       - freqtrade --datadir freqtrade/tests/testdata hyperopt -e 5
       name: hyperopt
-    - script: flake8 freqtrade scripts
+    - script:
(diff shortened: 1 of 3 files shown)


hroff-1902:
is it possible to add a job (or a check which is run after commit) which will compare the list of options (result of `-h`) in bot-usage.md with actual one after commit?

xmatthias:
> is it possible to add a job 

surely is - but i wouldn't know how - it's all in .travis.yml (however if it's more than a one-liner, i would add a script to `build_helpers`

I'm however not sure about the helpfullness of having this in the documentation - a list of all options with explanation would be better i think (better formated list than the one in configuration.md ...)

Issue #2171, opened 2019-08-22
Trailing stop cant set order because it would trigger immediately


Dante2333:
## Step 1: Have you search for this issue before posting it?

Yes

## Step 2: Describe your environment

  * Operating system: Ubuntu 19.04
  * Python Version: 3.7
  * CCXT version: latest
  * Branch: Master 
  * Last Commit ID: 2018.7 
  
## Step 3: Describe the problem:

The trailing stop wont be set even though it wont trigger automatically; 

### Steps to reproduce:

  1. Set trailing stop to offset at 0.2 and to set stoploss at 0.00
  2. run and wait
  
``` 42     # trailing stoploss
 43     trailing_stop = True
 44     trailing_stop_positive = 0
 45     trailing_stop_positive_offset = 0.2
 46     trailing_only_offset_is_reached = True
```
### Observed Results:

  * What happened? bot wont set the stoploss on exchange
  * What did you expect to happen? bot to set the stoploss on exchange and trail it up

### Relevant code exceptions or logs:

  ```
 2019-08-22 17:03:04,000 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
2019-08-22 17:03:07,488 - freqtrade.freqtradebot - WARNING - Stoploss order was cancelled, but unable to recreate one: Could not create stop_loss_limit sell order on market ETC/USDT.Tried to sell amount 32.28 at rate 7.4081.Message: binance Order would trigger immediately.
2019-08-22 17:03:07,488 - freqtrade.freqtradebot - INFO - Trailing stoploss: cancelling current stoploss on exchange (id:{71933263})in order to add another one ...
2019-08-22 17:03:07,988 - freqtrade.freqtradebot - ERROR - Could not cancel stoploss order 71933263 for pair ETC/USDT
Traceback (most recent call last):
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/base/exchange.py", line 509, in fetch
    response.raise_for_status()
  File "/home/dante/.local/lib/python3.7/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
[...]

Issue #1557, opened 2019-02-15
fail to place limit stop loss when market go down quickly after order buy is fulfilled (tight stop loss)


gaugau3000:
**Step 1: Have you search for this issue before posting it?**
Yes

**Step 2: Describe your environment**
Python Version: 3.7.2 (python -V)
CCXT version: 1.18.102 (pip freeze | grep ccxt)
Branch: Develop
Last Commit ID: id image docker image ebfd99410421 (git log --format="%H" -n 1)
Debian 9 64 bit
Docker version 17.05.0-ce, build 89658be

Exchange : binance

**Step 3: Describe the problem:**

"order_types": {
"buy": "limit",
"sell": "market",
"stoploss": "limit",
"stoploss_on_exchange": true
}

Sometimes for some trades freqtrade don't succeed to create a stop order (for the the stop loss) and create only the buy limit order

for info the stoploss the very tight in this strategy : -0.0037

**Steps to reproduce:**

stop loss very tight in strategy, bot buy then try to place a stop above current price and don't succeed.

**Observed Results:**

**What happened?**

Warning in the logs and don't succeed to place stop loss order.
**Q :** on the logs he try to place a stop loss only when the limit order is complete. Why he don't place the stop in the same time ?

**What did you expect to happen?**

Place the stop order on exchange when create the buy limit order.

**Relevant code exceptions or logs:**

```
 Buy signal found: about create a new trade with stake_amount: 0.080000 ...
2019-02-15 10:30:08,280 - freqtrade.freqtradebot - INFO - Using Last Ask / Last Price
2019-02-15 10:30:08,863 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': buy, 'exchange': 'Binance', 'pair': 'SUB/BTC', 'market_url': 'https://www.binance.com/tradeDetail.html?symbol=SUB_BTC', 'limit': 9.32e-06, 'stake_amount': 0.08, 'stake_currency': 'BTC', 'fiat_currency': 'USD'}
2019-02-15 10:30:09,226 - freqtrade.wallets - INFO - Wallets synced ...
2019-02-15 10:30:12,023 - freqtrade.freqtradebot - INFO - Found open order for Trade(id=12, pair=SUB/BTC, amount=8583.69098712, open_rate=0.00000932, open_since=just now)
2019-02-15 10:30:13,095 - freqtrade.persistence - INFO - Updating trade (id=12) ...
[...]

PR #2173, opened 2019-08-22
improve stoploss validation

xmatthias:
# Summary
Stoploss with 0 is problematic, since that would lead to sells immediately.

This should prevent the bog from starting up - otherwise it results in misleading bug reports (as #2171).

Solve the issue: #2171

## Quick changelog

- Validate stoploss_positive !=0 (it can be positive or negative, since we use `-1 * abs()` when applying this).
- validate stoploss != 0 (same `-1 * abs()` logic as before

### DIFF
--- a/freqtrade/configuration/config_validation.py
+++ b/freqtrade/configuration/config_validation.py
@@ -68,6 +68,10 @@ def validate_config_consistency(conf: Dict[str, Any]) -> None:
 
 def _validate_trailing_stoploss(conf: Dict[str, Any]) -> None:
 
+    if conf.get('stoploss') == 0.0:
+        raise OperationalException(
+            'The config stoploss needs to be different from 0 to avoid problems with sell orders.'
+            )
     # Skip if trailing stoploss is not activated
     if not conf.get('trailing_stop', False):
         return
@@ -79,12 +83,19 @@ def _validate_trailing_stoploss(conf: Dict[str, Any]) -> None:
     if tsl_only_offset:
         if tsl_positive == 0.0:
             raise OperationalException(
-                f'The config trailing_only_offset_is_reached needs '
+                'The config trailing_only_offset_is_reached needs '
                 'trailing_stop_positive_offset to be more than 0 in your config.')
     if tsl_positive > 0 and 0 < tsl_offset <= tsl_positive:
         raise OperationalException(
-            f'The config trailing_stop_positive_offset needs '
+            'The config trailing_stop_positive_offset needs '
             'to be greater than trailing_stop_positive_offset in your config.')
+
+    # Fetch again without default
+    if 'trailing_stop_positive' in conf and float(conf['trailing_stop_positive']) == 0.0:
+        raise OperationalException(
+            'The config trailing_stop_positive needs to be different from 0 '
+            'to avoid problems with sell orders.'
+        )
 
 
 def _validate_edge(conf: Dict[str, Any]) -> None:
--- a/freqtrade/configuration/config_validation.py
+++ b/freqtrade/configuration/config_validation.py
@@ -88,7 +88,7 @@ def _validate_trailing_stoploss(conf: Dict[str, Any]) -> None:
     if tsl_positive > 0 and 0 < tsl_offset <= tsl_positive:
         raise OperationalException(
             'The config trailing_stop_positive_offset needs '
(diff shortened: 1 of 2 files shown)


hroff-1902:
I think it won't keep me from getting those errors from exchange, since I would set stoploss = -0.00001 in order to circumvent this artificial limitation. And I'll get same error from exchange.
Instead, those messages from exchange should be handled and the pair sold immediately. Or an appropriate flag for the order (yes, exchange specific, but we have exchange classes for it) should be set to force it to sell the pair immediately in these cases...

xmatthias:
If i want to circumvent this limitation, then i'll comment out the calls to this check-method.

It's about **accidentally** setting this wrongly, which is what happened in #2171.

If i want to circumvent this with -0.0000001 - then i'm an idiot and should not be using the bot since i'm trying to create an error.

Yes, the error can be thrown anyway (see #1557 / https://github.com/freqtrade/freqtrade/issues/1557#issuecomment-524165920) - but that's not because of the value - it can be also thrown if i set it to -0.05, which is something we cannot protect from.

hroff-1902:
> If i want to circumvent this limitation, then i'll comment out the calls to this check-method.

@xmatthias you are cool 🍺 , I have not meant it. It's a user, not you, who can find this limitation (stoploss=0.0 is forbidden) and circumvent it (aha, stoploss=0.00001 is okay!)

> It's about **accidentally** setting this wrongly, which is what happened in #2171.

I think it's not wrong. It also has meaning -- close the trade immediately (well, in a few seconds, probably) if the price stays same or move a small fraction down. That's the meaning of this stoploss. As any other small stoplosses, like 0.00001. Nothing is more errorneous in stoploss=0.0 than in stoploss=0.000001. All of them are equally wrong or equally can be assigned some (right, stupid) meaning. And you won't be able to figure out the border between those accidentally wrong stoploss values and the reasonable ones. The only way is to handle them *without exchange errors*, even if those errors are exchange-specific. That means they are to be handled in the exchange-specific manner.

> If i want to circumvent this with -0.0000001 - then i'm an idiot and should not be using the bot since i'm trying to create an error.

Many people behave strange and even like idiots and many of them have their own intrinsic reasons and explanation for this... The bot should behave according to the settings defined. As I said above, small stoplosses can have they own (probably useless) meaning and corresponding expected bot behavior.

> Yes, the error can be thrown anyway (see #1557 / [#1557 (comment)](https://github.com/freqtrade/freqtrade/issues/1557#issuecomment-524165920)) - but that's not because of the value - it can be also thrown if i set it to -0.05, which is something we cannot protect from.

Well, let's commit it, this limitation on the stoploss values, no problem. But it does not resolve the actual problem (stoploss is not set on exchange, the order is not created (or a buy order is created, but without a protective stoploss? Which is even worse in this case...)). And, imho, when the real problem will be fixed (in the exchange-specific way, probably), it may happen that this limit was not actually needed, because this particular edge case will be handled on exactly same manner as other small stoplosses...

freqtrade/configuration/config_validation.py
```diff
@@ -79,13 +83,20 @@ def _validate_trailing_stoploss(conf: Dict[str, Any]) -> None:
...
         if tsl_positive == 0.0:
             raise OperationalException(
-                f'The config trailing_only_offset_is_reached needs '
+                'The config trailing_only_offset_is_reached needs '
                 'trailing_stop_positive_offset to be more than 0 in your config.')
     if tsl_positive > 0 and 0 < tsl_offset <= tsl_positive:
         raise OperationalException(
-            f'The config trailing_stop_positive_offset needs '
+            'The config trailing_stop_positive_offset needs '
             'to be greater than trailing_stop_positive_offset in your config.')
```
hroff-1902: Emm, "trailing_stop_offset ... than trailing_stop_positive"?

xmatthias: Great catch :+1:

xmatthias:
> But it does not resolve the actual problem 

i never said it will ... I hope it will prevent accidentally configuring it wrongly (which is the same for all checks we do in config validation ...).

> it may happen that this limit was not actually needed, because this particular edge case will be handled on exactly same manner as other small stoplosses...

You're right, 0.00001 is just as wrong as 0.0, but it's difficult to define a meaningfull limit.

PR #2174, opened 2019-08-22
WiP: refactor exchange

hroff-1902:
This PR moves created exchange instances to a kind of singleton (python's modules are singletons) or a factory, with an easy way to fetch the exchange where it's needed:

```
exchange: Exchange = get_exchange(config, exchange_name)
```
or even
```
exchange: Exchange = get_exchange(config)
```
(the exchange name will be taken from the config)

* Exchange objects are created on-demand.
* Simplifies interfaces (see DataProvider and Wallet). There is not need to keep the `self.exchanges` attribute at all (Wallet) or pass exchange to the init (DataProvider).
* Hides ExchangeResolver, no need to call it or create exchange explicitly as `Exchange()`
* Resolves pickling problems in hyperopt (no need to delete the exchange before start of the workers processes). It probably would allow using dataprovider under hyperopt (need to check).
* exchanges are kept in a dict, this is a provisionary solution. It will allow in the future create Exchange objects for other exchanges (for informative pairs from other exchanges, with syntax like `OTHEREXCHANGE:XXX/YYY`), allowing inter-exchange arbitrage strategies...
* in the previous version I tried to implement it in a sep. `class Exchanges` in a sep. `exchanges.py` module under exchange dir, but python circular dependencies are awful...

Huh? (You may not like it since 'explicit is better than implicit' but...)

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -21,6 +21,7 @@ from freqtrade import (DependencyException, InvalidOrderException,
 from freqtrade.data.converter import parse_ticker_dataframe
 from freqtrade.misc import deep_merge_dicts
 
+
 logger = logging.getLogger(__name__)
 
 
@@ -770,6 +771,39 @@ class Exchange(object):
                 f'Could not get fee info due to {e.__class__.__name__}. Message: {e}') from e
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
+
+
+# Store Exchange objects created
+__exchanges: Dict[str, Exchange] = {}
+
+
+def get_exchange(config, exchange_name: Optional[str] = None) -> Exchange:
+    """
+    Return Exchange object for exchange with name `exchange_name`.
+    Creates Exchange if it was not created before.
+    """
+    from freqtrade.resolvers import ExchangeResolver
+
+    name = exchange_name or config['exchange']['name']
+    if name not in __exchanges:
+        logger.info(f"Creating Exchange object for '{name}'...")
+        __exchanges[name] = ExchangeResolver(name, config).exchange
+        logger.info(f"Exchange object for '{name}' created.")
+
+    return __exchanges[name]
+
+
+def delete_exchange(config, exchange_name: Optional[str] = None):
+    """
+    Delete Exchange object for exchange with name `exchange_name`.
+    """
+    name = exchange_name or config['exchange']['name']
+    logger.info(f"Deleting Exchange object for '{name}'...")
+    if name in __exchanges:
+        del __exchanges[name]
+        logger.info(f"Exchange object for '{name}' deleted.")
+    else:
+        logger.warning(f"Exchange object for '{name}' was not found, cannot be deleted.")
 
 
 def is_exchange_bad(exchange_name: str) -> bool:
--- a/freqtrade/wallets.py
+++ b/freqtrade/wallets.py
@@ -2,8 +2,8 @@
 """ Wallet """
 
 import logging
-from typing import Dict, NamedTuple
-from freqtrade.exchange import Exchange
+from typing import Dict, NamedTuple, Optional
+from freqtrade.exchange import Exchange, get_exchange
 from freqtrade import constants
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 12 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -772,6 +773,39 @@ def get_fee(self, symbol='ETH/BTC', type='', side='', amount=1,
...
+
+
+def delete_exchange(config, exchange_name: Optional[str] = None):
+    """
+    Delete Exchange object for exchange with name `exchange_name`.
+    """
+    name = exchange_name or config['exchange']['name']
+    logger.info(f"Deleting Exchange object for '{name}'...")
+    if name in __exchanges:
+        del __exchanges[name]
```
hroff-1902: this line is probably incorrect;
the whole `delete_exchange` function should probably be eliminated, since we do not really need to delete an Exchange object once it was created...

freqtrade/exchange/exchange.py
```diff
@@ -772,6 +773,39 @@ def get_fee(self, symbol='ETH/BTC', type='', side='', amount=1,
             raise OperationalException(e) from e
 
 
+# Store Exchange objects created
+__exchanges: Dict[str, Exchange] = {}
```
hroff-1902: The dict key here can in the future be the `Tuple[str, bool]`, for example, adding space for live/dry-run variants of the exchange objects existing in the same time...

This is just an illustration of the flexibility of the approach...

freqtrade/wallets.py
```diff
@@ -61,7 +61,8 @@ def get_total(self, currency) -> float:
 
     def update(self) -> None:
 
-        balances = self._exchange.get_balances()
+        exchange: Exchange = get_exchange(self._config, self._exchange_name)
```
xmatthias: This should be in the constructor, not done each time (we call this many times!)

before changing, read the main comment too though.

freqtrade/exchange/exchange.py
xmatthias: why would you want to have this in the same instance ... ? 

i cannot immagine any scenario where this would make sense...

xmatthias:
I'll have to think about this - i'm not a particular fan of using singletons, since they can bite you badly.

I assume you're trying to simplify this for interactive usage?

Yes, the dataprovider interface will be simplified slightly - but you never really instanciate that as a user anyway - we pass it as `dp` to the stratgegy, so it's not really a users problem.

If you really need a new instance with another exchange, then `ExchangeResolver('other_exchange', self.config).exchange`is not that difficult to call and pass to the dataproviders constructor.

> Resolves pickling problems in hyperopt (no need to delete the exchange before start of the workers processes). It probably would allow using dataprovider under hyperopt (need to check).

Remember that creating a new exchange object is rather expensive (doing both sync and async call to `load_markets()` for example) - so i would probably want to control where exactly this is happening, and not in some random module...

Also, if you happen to run into this within a hyperopt loop, your hyperopt will basically stop since every job will do this instantiation / network call to the exchange (and you'll probably be blocked afer a few epochs...)

So honestly, for hyperopt, you move more complexity to the user since they need to make sure not to accidentally, implicitly instanciate exchanges within hyperopt.

hroff-1902:
I'm closing this WiP PR, the general solution would probably include the common ClassFactory for all other singletons (just ideas)...

xmatthias:
> I'm closing this WiP PR, the general solution would probably include the common ClassFactory for all other singletons (just ideas)...

whatever that ideas are - let's discuss benefits / drawbacks of these before investing time in working on them -  that'll avoid frustration during the PR process (singletons are limiting certain things - which is sometimes good, sometimes bad, sometimes not thought about and creates problems later on)

hroff-1902:
> > I'm closing this WiP PR, the general solution would probably include the common ClassFactory for all other singletons (just ideas)...
> 
> whatever that ideas are - let's discuss benefits / drawbacks of these before investing time in working on them - that'll avoid frustration during the PR process (singletons are limiting certain things - which is sometimes good, sometimes bad, sometimes not thought about and creates problems later on)

That's why I closed it. Sure, all generic solutions should be discussed first. The comment was just a note for myself and for others, I'm not going to start implementing something here.

PR #2175, opened 2019-08-23
Hyperopt redesign

hroff-1902:
Do not inherit Hyperopt from Backtesting. Make an instance of Backtesting an attribute of Hyperopt.

Rationale: because Hyperopt is not a ***kind of*** Backtesting, it ***uses*** Backtesting...

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -37,7 +37,7 @@ INITIAL_POINTS = 30
 MAX_LOSS = 100000  # just a big enough number to be bad result in loss optimization
 
 
-class Hyperopt(Backtesting):
+class Hyperopt:
     """
     Hyperopt class, this class contains all the logic to run a hyperopt simulation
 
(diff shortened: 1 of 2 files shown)


xmatthias:
I'll try to have a look at this in the afternoon.

Additional idea: maybe we can create an abstract `IBacktesting` parent class, which has Backtesting and EdgeCli (or edge?) inherit from that - and hyperopt then can be used to optimize both backtesting and edge ... 

I had this idea a long time ago - but the edge was not ready for that ...

hroff-1902:
And this redesign simply emphasizes, highlights the design problems existing in backtesting. That's the intention of this PR -- make the situation a bit more proper from the design point of view and emphasizes others, that are a kind of hidden when Hyperopt inherits Backtesting. You see, the backtesting attributes are fetched in hyperopt and are passed back into the backtest method of backtesting... The exchange is created in backtesting but accessed in the hyperopt... The situation will be even more highlighted if I split backtesting into the cli and non-cli parts...

hroff-1902:
What set of methods can be assumed in IBacktesting? So that I understand the idea...

xmatthias:
I'm still not convinced in splitting backtesting to cli and non-cli parts - since there is really no non-cli part (you can only run backtesting from the cli - not like edge, which runs from a trading-bot as well).

> What set of methods can be assumed in IBacktesting

I don't know - everything that's called by hyperopt (probably) - and shared between backtesting and edge-cli (maybe) ... 
but i'm not really sure about this either - I wanted to mention it so someone else can think about this too (since my focus is currently not really with backtesting/hyperopt areas).

PR #2178, opened 2019-08-25
[refactor] Move stoploss on exchange implementation to binance

xmatthias:
## Summary
Stoploss on exchange is currently only supported by binance, and the code is binance-specific, therefore we should move the implementation to the binance subclass.

We also should not use `self.create_order()` for this case, since we need to handle `InvalidOrder` exceptions differently (in a future PR) - probably by force-selling the pair.

Note: There is no functional change in this PR, so nothing will improve for users, but it enables #1557

## Quick changelog

- Move implementation to binance subclass
- some minor maintenance in exchange class
- Improve docstring for OperationalException 
- move corresponding tests to newly created `test_binance.py`

### DIFF
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -11,7 +11,7 @@ class DependencyException(Exception):
 
 class OperationalException(Exception):
     """
-    Requires manual intervention.
+    Requires manual intervention and will usually stop the bot.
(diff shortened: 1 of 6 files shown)


hroff-1902:
LGTM 👍

PR #2184, opened 2019-08-25
minor: Backtesting cleanup

hroff-1902:
Small optimization in Backtesting: move initialization of `self.ticker_interval` from `_set_strategy` method to `__init__()`. It's not taken from the strategy attribute and required to be set in the config/options.

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -64,6 +64,12 @@ class Backtesting(object):
         self.config['dry_run'] = True
         self.strategylist: List[IStrategy] = []
 
+        if "ticker_interval" not in self.config:
+            raise OperationalException("Ticker-interval needs to be set in either configuration "
+                                       "or as cli argument `--ticker-interval 5m`")
+        self.ticker_interval = str(self.config.get('ticker_interval'))
+        self.ticker_interval_mins = timeframe_to_minutes(self.ticker_interval)
+
         self.exchange = ExchangeResolver(self.config['exchange']['name'], self.config).exchange
         self.fee = self.exchange.get_fee()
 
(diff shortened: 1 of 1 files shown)


xmatthias:
This changes the behaviour of using backtesting with only one strategy.

to recreate:
* remove ticker-inteval from the configuration
* make sure ticker_interval is set in the strategy
* run with: `freqtrade --config test_config.json --strategy BinHV45  backtesting --timerange 20190801-` 

Explanation:
Config is copied for strategy lists - so `ticker_interval` cannot be set by multiple strategies.
However, in case of one strategy (passed in as `--strategy`, not `--strategy-list`), `self.config` is passed to the strategy directly, allowing the strategy to set the `ticker_interval`.

Apparently there is no test with backtesting for this behaviour (there are multiple in `test_strategy.py` confirming overrides) - so you need to run into this while testing (I don't have `ticker_interval` in my test-config to test different strategies and prefer them to set the interval - otherwise i may have missed this too).

Fix:
Fixing this is rather simple - move the changed lines to after the strategy loading (right before the call to  `set_strategy()`

hroff-1902:
Thanks! I'll fix it.

Actually I was thinking about proposing (sep. issue) to deprecate --strategylist. It can simply be done in a shell loop, taking all the ticker_interval's properly...

xmatthias:
> Actually I was thinking about proposing (sep. issue) to deprecate --strategylist. It can simply be done in a shell loop, taking all the ticker_interval's properly...

The result might be the same (but no direct comparison / comparison is cumbersome), but timing will be a lot worse assuming you backtest more than a few days / pairs.
`--strategylist` reads the data exactly once from disk - and keeps it in memory during the backtest operation - assuming you're trying to compare 10strategies for half a year and 100 pairs - that's a HUGE speed improvement.

hroff-1902:
NIce point to keep it, thanks for explaination.

hroff-1902:
> * remove ticker-inteval from the configuration

We can now comment it out. 😉

xmatthias:
> We can now comment it out.

which is technically removing :smile:

hroff-1902:
> > We can now comment it out.
> 
> which is technically removing 😄

I mean I cannot also remember of this new ability, it's already like a muscle memory for the bot, to remove the settings in the config instead of commenting them out... ///

Issue #2183, opened 2019-08-25
Strategy interface change discussion - enable position management and short trading


hroff-1902:
I think I'll disclose here my current ideas on design solutions for position management and trading short functionality. 

It's not a simple task, so it can be done in a steps. My main idea is the following:

0) The position management rules (algo, calculations) should be a part of the custom strategy.
1) So, as the first step, the `populate_buy()` and `populate_sell()` should be composed into one method. The resulting method (say, `populate_signals()`) should populate and return a column in the resulting dataframe, say, `position` with 2 possible values: 1 for current 'buy' and '0' for current 'sell'.
1a) The current interface should be kept. So there may be a setting like `experimental.use_populate_signal` which will switch between using current `populate_buy`/`populate_sell` and new `populate_signals`
1b) The above can probably be implemented as a helper function, which fetches results of current `populate_buy`/`populate_sell` and creates that new single signal column.
2) As the next step, this value, returned by the populate_signal, should be allowed to be float. Treated as ratio of stake_amount. So 1.0 means buy one stake_amount, 2.0 -- average up, 0.0 -- sell open position, 0.5 -- buy one half of the defines stake_amount, -1.0 -- trade short (! huh? why not...)

The user should be able to have access to trades made previously, wallets to perform position management calculations.

And voila, we are there... Of course, besides changing the strategy interface, all this will require support for fractional stakes everywhere in fraqtradebot, exchange, edge, backtesting and so on...

xmatthias:
I think we should not mix too many things into one issue - that'll end up with an issue that noone will want to work on, since it'll be a neverending / never closing issue.

> The user should be able to have access to trades made previously, wallets to perform position management calculations

this is known, and not really related to anything else in this issue as users will benefit irrelevant of this. Tracked here, and i would like to keep that seperate: #1753
It's however dangerous to rely on since past trades are not available during backtesting.

1. i think the `position` column has flaws which may not work - what if the current candle is a "no action"? 
additionally: Using 0 as "sell" is dangerous, since we have strategies in the strategy repo "without" a sell-signal, which simply set sell to 0 (`dataframe['sell'] = 0`).
Since we have examples doing this, one has to assume that private user-strategies will do the same. 
Thinking about the logic which combines buy and sell into one column will be difficult, since 0 in the sell-column never meant "selling" - but "no action" - so you'd need to add sells as `if sell 1 then 0` - then buys `if 1 then 1` ... which complicates things beyond reasonings.
[...]

PR #2190, opened 2019-08-26
Introduce strategy_version 

xmatthias:
## Summary
In order to allow future iterations of the strategy interface, we should version the API (basically allowing the strategy to tell the bot clearly what interface it's going to use without us trying to parse this from available methods or similar).

This will enable / simplify work on #2183.

## Quick changelog

- Introduce `strategy_version` (current version is 2 - legacy would be 1 and is auto-detected).

## Open questions
- [x] is `strategy_version` a good variable name for this?
- [x] documentation (once name is clear)

### DIFF
--- a/freqtrade/strategy/default_strategy.py
+++ b/freqtrade/strategy/default_strategy.py
@@ -13,7 +13,7 @@ class DefaultStrategy(IStrategy):
     Default Strategy provided by freqtrade bot.
     You can override it with your own strategy
     """
-    strategy_version: int = 2
+    INTERFACE_VERSION = 2
 
     # Minimal ROI designed for the strategy
     minimal_roi = {
--- a/freqtrade/strategy/default_strategy.py
+++ b/freqtrade/strategy/default_strategy.py
@@ -13,6 +13,7 @@ class DefaultStrategy(IStrategy):
     Default Strategy provided by freqtrade bot.
     You can override it with your own strategy
     """
+    strategy_version: int = 2
 
     # Minimal ROI designed for the strategy
     minimal_roi = {
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -60,6 +60,11 @@ class IStrategy(ABC):
         stoploss -> float: optimal stoploss designed for the strategy
         ticker_interval -> str: value of the ticker interval to use for the strategy
     """
+    # Strategy interface version
+    # Default to version 2
+    # version 1 is the initial interface without metadata dict
+    # Version 2 populate_* include metadata dict
+    strategy_version: int = 2
 
     _populate_fun_len: int = 0
     _buy_fun_len: int = 0
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -62,9 +62,9 @@ class IStrategy(ABC):
     """
     # Strategy interface version
     # Default to version 2
-    # version 1 is the initial interface without metadata dict
+    # Version 1 is the initial interface without metadata dict
     # Version 2 populate_* include metadata dict
-    strategy_version: int = 2
+    INTERFACE_VERSION: int = 2
 
     _populate_fun_len: int = 0
     _buy_fun_len: int = 0
--- a/docs/strategy-customization.md
+++ b/docs/strategy-customization.md
@@ -36,8 +36,13 @@ A strategy file contains all the information needed to build a good strategy:
 - Minimal ROI recommended
 - Stoploss strongly recommended
 
-The bot also include a sample strategy called `TestStrategy` you can update: `user_data/strategies/test_strategy.py`.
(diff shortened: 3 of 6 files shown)


freqtrade/strategy/default_strategy.py
```diff
@@ -13,6 +13,7 @@ class DefaultStrategy(IStrategy):
     Default Strategy provided by freqtrade bot.
     You can override it with your own strategy
     """
+    strategy_version: int = 2
```
hroff-1902: Well, I think this should be simplified for a user, so no `: int`, then it should be like a constant, so in upper letters, then it's the version of the interface, not the strategy itself, I propose
```
STRATEGY_INTERFACE_VERSION = 2
```
or
```
STRATEGY_INTERFACE_VER = 2
```
or
```
strategy_interface_version = 2
```
if you mind this to be in capital letters.

We also could drop `strategy` from the name since it's in the Strategy class, so
`INTERFACE_VERSION` or `interface_version` can be used as well.

freqtrade/strategy/interface.py
```diff
@@ -60,6 +60,11 @@ class IStrategy(ABC):
         stoploss -> float: optimal stoploss designed for the strategy
         ticker_interval -> str: value of the ticker interval to use for the strategy
     """
+    # Strategy interface version
+    # Default to version 2
+    # version 1 is the initial interface without metadata dict
```
hroff-1902: "Version"

freqtrade/strategy/default_strategy.py
xmatthias: `INTERFACE_VERSION` seems the best (short, and as a constant)

hroff-1902:
Actually, the behavior here should be as follows, in my opinion:

-- fetch INTERFACE_VERSION value from the strat, if present.
-- if not, try to guess it: if any of populate_*_len_fun is 2, then INTERFACE_VERSION=1, otherwise current INTERFACE_VERSION=2. When we'll have 3rd interface version, we probably also will know how to detect it. Or we'll require INTERFACE_VERSION to be explicitly set for INTERFACE_VERSION=3 and later, otherwise current INTERFACE_VERSION=2 (or INTERFACE_VERSION=1, see above) will be assigned.
-- then, once we have interface version detected or fetched, we can make some validation. For INTERFACE_VERSION=1 we can validate that *all* populate_*_len_fun are 2, otherwise raise exception.  For INTERFACE_VERSION=2 we can validate that *all* populate_*_len_fun are 3, otherwise raise exception. For INTERFACE_VERSION=3 other validation conditions can be checked.

Maybe in a sep. PR.

hroff-1902:
closed it by mistake, reopened

xmatthias:
Depending on the changes done between v2 and v3, it may be difficult to detect the version due to inheritance, and we cannot assume to always break the interface, maybe just the return type / value is different  - which we cannot easily detect on startup, so my idea was to require `INTERFACE_VERSION` to be set starting with the next iteration (`3` ...).


The way I thought this PR to work is to default to V2 (which we already did before this PR), and to always keep it that way (so you explicitly have to select to use newer features / functionality). This should not be a problem, since you're already adapting the strategy...

The detection between V1 and V2 did not change. 
The different calling methods still rely on `_<method>_fun_len`, and not on `INTERFACE_VERSION` to support eventual 1.x versions (one function using new method, the other 2 using the old method or similar - which does work currently).

hroff-1902:
> Depending on the changes done between v2 and v3, it may be difficult to detect the version due to inheritance, and we cannot assume to always break the interface, maybe just the return type / value is different - which we cannot easily detect on startup, so my idea was to require `INTERFACE_VERSION` to be set starting with the next iteration (`3` ...).

Same above in my vision (with an option for detection between newer versions, but requirement for `INTERFACE_VERSION` to be set for all newer versions is completely okay.

> The way I thought this PR to work is to default to V2 (which we already did before this PR), and to always keep it that way (so you explicitly have to select to use newer features / functionality). This should not be a problem, since you're already adapting the strategy...

It reminds me our struggling with versions of CORBA messages many years ago... Nothing has changed... 😄 

> The detection between V1 and V2 did not change.
> The different calling methods still rely on `_<method>_fun_len`, and not on `INTERFACE_VERSION` to support eventual 1.x versions (one function using new method, the other 2 using the old method or similar - which does work currently).

Should this mix really be allowed? But anyway, this really does not change anything. My idea was to detect that v.1 is used and print it in the log (the proper version of the interface, not fallback to current v.2 if actually old interface v.1 is used). Something like:

`Strategy interface version is not detected, guessing it,,,`
`The strategy {strategy} uses interface version 1. (maybe reason: The populate_sell() method uses 2 parameters).`
`DependencyError: The strategy populate_interface() method accepts 3 parameters while only 2 are allowed.`

But agree, this all can be redundant. Accepting the PR. The above is more just thoughts of possible design solutions/notes that do not have to be implemented.

PR #2191, opened 2019-08-27
minor: improvements in confuguration.md

hroff-1902:
* The `exchange.password` setting added to the table of settings
* Warning 'Keep it in secrete, do not disclose publicly.' added for `exchange.key/secrete/password` and for `telegram.token/chat_id`


hroff-1902:
> LGTM 👍
> 
> i wonder if we should highlight the hint more (maybe bold?) - since this seems quite important...

Nice point, do not merge, will highlight it...

xmatthias:
> do not merge

that's why i left it there ... ;)

PR #2192, opened 2019-08-27
Rename testStrategy to sample_strategy

xmatthias:
## Summary
Having the name test_strategy can be misleading that this is the strategy which does only test but not use real money.

Also, renaming `TestStrategy` to `SampleStrategy`aligns it with the naming convention used for `Hyperopt`/`HyperoptLoss`.

NB: LInks will fail until merged, since they link to the develop-branch, where the file is not yet renamed.

### DIFF
--- a/user_data/strategies/sample_strategy.py
+++ b/user_data/strategies/sample_strategy.py
@@ -13,7 +13,7 @@ import numpy  # noqa
 # This class is a sample. Feel free to customize it.
 class SampleStrategy(IStrategy):
     """
-    This is an example strategy to inspire you.
+    This is a sample strategy to inspire you.
     More information in https://github.com/freqtrade/freqtrade/blob/develop/docs/bot-optimization.md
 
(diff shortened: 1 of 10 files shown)


hroff-1902:
I also right yesterday wanted to create an issue for this point. 👍

user_data/strategies/sample_strategy.py
```diff
@@ -11,10 +11,9 @@
 
 
 # This class is a sample. Feel free to customize it.
-class TestStrategy(IStrategy):
-    __test__ = False  # pytest expects to find tests here because of the name
+class SampleStrategy(IStrategy):
     """
-    This is a test strategy to inspire you.
+    This is an example strategy to inspire you.
```
hroff-1902: "a sample strategy" ???

hroff-1902:
LGTM 👍 
Just one minor note on wording in a comment...

PR #2197, opened 2019-08-28
Apply dynamic versioning to develop

### LINKED ISSUE #2166: Build number in -dev versioning
Since we do not have a build number in the -dev version, it may be quite complex to understand which particular version of freqtrade the user runs (especially with that your conda). So if we change user interface I guess it's better for a while to make those changes along with changing the version of both develop and master...

Is it possible to introduce the build number into the -dev versioning, which will change with every merge, as ccxt employs, for example?

It's not necessary for master (2019-8 is completely enough), but for develop changing every day this may be useful since `freqtrade -V` will give particular version info (even when a user has current codebase from github fetched in a dir, but installed an outdated version with that your conda, which is used instead of latest develop from the dir; that's the confusing usecase when it's complicated to find out the actual version of freqtrade that runs at the user site...)

It's not necessary to be an increasing number, it can be a short commit id (merged into develop), for example...

xmatthias:
## Summary
We should use git commits as version for develop. This ensures that it's always uptodate, and we don't need to adjust the version in develop when we do a new relese.

closes #2166 

## Quick changelog

- use `git describe` to print version if we're on develop (`__version__` is develop

### DIFF
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -5,8 +5,9 @@ if __version__ == 'develop':
 
     try:
         import subprocess
-        __version__ = str(subprocess.check_output(
-            ["git", "describe"], stderr=subprocess.DEVNULL).rstrip())
+        __version__ = 'develop-' + subprocess.check_output(
+            ['git', 'log', '--format="%h"', '-n 1'],
+            stderr=subprocess.DEVNULL).decode("utf-8").rstrip().strip('"')
     except Exception:
         # git not available, ignore
         pass
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -1,5 +1,15 @@
 """ FreqTrade bot """
-__version__ = '2019.7-dev'
+__version__ = 'develop'
+
+if __version__ == 'develop':
+
+    try:
+        import subprocess
+        __version__ = str(subprocess.check_output(
+            ["git", "describe"], stderr=subprocess.DEVNULL).rstrip())
+    except Exception:
+        # git not available, ignore
+        pass
 
 
 class DependencyException(Exception):


freqtrade/__init__.py
```diff
@@ -1,5 +1,15 @@
 """ FreqTrade bot """
-__version__ = '2019.7-dev'
+__version__ = 'develop'
+
+if __version__ == 'develop':
+
+    try:
+        import subprocess
```
xmatthias: This is currently called on every startup for develop.
Not sure if that's a problem (timing-vise) or if we can assume it's ok.

hroff-1902: Performance impact is not visible, as far as I see...

* The output is strange: `freqtrade b'0.14.3-5523-ga7e45c5a'`
* It does not work if a user runs the installed version from another dir.
* This will not be needed if freqtrade is installed via pupy. Is it possible to publish the develop asset to pupy (when this will be implemented for master) after each merge to develop with corresponding git commit id inside?

xmatthias: yes, the output can be improved by using a different git command (probably `git log ...` which shows the short version of the hash).

> It does not work if a user runs the installed version from another dir.

Yes - then it'll simply print `develop` - which is similar to the current output (`2019-7-dev` - which does not say much more about the version.

For master, we won't need this, since we'll increment the version as we do now.

I doubt that we'll push develop versions to pypi ... while it's probably possible, it'll create a mess on what the current version is (pypi does not support different branches afaik) - we would have to either publish to packages, or play with version-numbers to make them unique (and always select the correct one ...) which i think is difficult.

I don't think once we're on pypi too many people will be using develop other than developers (and people living on the edge). I guess we'll then have it somewhat outomated so we can release more often (although surely not at the frequency of ccxt with ~4 releases per day ...)

hroff-1902:
The following may be interesting to you (I have not looked deep into the details yet): https://github.com/dmytrostriletskyi/pypi-version

xmatthias:
maybe ... but it's still a bumpy road to pypi :smile:

hroff-1902:
This can at least be called when `-V` option prints the version, if not at each run...

xmatthias:
well it is called at each run IF you're on the develop branch, and only then.

Master / releases will have a different `__version__` (other than `"develop"`)  - as they do now - so they don't run the git command, but instead use the predefined __version__.

hroff-1902:
Ah, right. That's simply fine as soon as moving to pypi changes recommendation (I assume) on how the bot should be installed by an ordinal user or in a production environment -- no place for fetching develop branch except developers, right?

xmatthias:
i assume we will change this, yes. 

I'm also not 100% sure if we should publish develop versions as "pre-release" versions to pypi (this has some disadvantages...) - but this can be decided later (it may require slightly different version-numbers to make sure they are incremental).

To me, this is the method of versioning exclusively for checked out git repositories (it'll also fail for most other install methods).

hroff-1902:
> To me, this is the method of versioning exclusively for checked out git repositories (it'll also fail for most other install methods).

Sure, Now, when the official installation method is based on git, it has much sense than after integration to pypi. Maybe we'll simply drop it?

PR #2199, opened 2019-08-29
Fix datadir init to always include exchange

xmatthias:
## Summary
When running `freqtrade download-data --exchange binance ` - the path needs to point to the final folder (incl. exchange). To accomplish this, `--exchange` needs to be parsed before setting datadir.
This also simplifies detecting the pair-list (wrongly fixed in #2168).

## Quick changelog

- Change argument resolution sequence to have exchange name available earlier
- test that running the bot with `--exchange binance` sets the correct path.
- Fix test-leakage (MINCONFIG was changed...) by not copying a dict properly.

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -160,6 +160,11 @@ class Configuration(object):
         Extract information for sys.argv and load directory configurations
         --user-data, --datadir
         """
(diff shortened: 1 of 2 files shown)


hroff-1902:
merged without waiting for @xmatthias since it was asked for at Slack

PR #2119, opened 2019-08-11
Disable stoploss on exchange during dry-runs

xmatthias:
## Summary
Stoploss on exchange does not make sense for dry-run scenarios - so we need to disable this option similar to how we handle this for backtesting.

Closes #2106

Using stoploss on exchange with dry-runs causes errors, and will provide no benefit since there is no exchange executing dry orders.

Issue #2106, opened 2019-08-06
KeyError: 'stopPrice'


Prossi79:
## Step 1: Have you search for this issue before posting it?

yes, nothing found

## Step 2: Describe your environment

Python Version: Python 3.7.3
CCXT version: ccxt==1.18.992
Branch: Develop
Last Commit ID: eeecdd4e5a5f70ba2e7943bf3c6aaf4b5742e5b3

  
## Step 3: Describe the problem:

*Explain the problem you have encountered*

### Steps to reproduce:

  1. config file - stoploss on exchange - true
  2. FT started normally
  3. buy signal found -> see log below
  
### Observed Results:

  * Telegram message "Process Died"
  

### Relevant code exceptions or logs:

  ```
2019-08-06 22:44:55,366 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again..
2019-08-06 22:44:59,196 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again..
2019-08-06 22:45:05,099 - freqtrade.freqtradebot - INFO - Buy signal found: about create a new trade with stake_amount: 980 ...
2019-08-06 22:45:05,099 - freqtrade.freqtradebot - INFO - Using Last Ask / Last Price
2019-08-06 22:45:05,400 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': buy, 'exchange': 'Binance', 'pair': 'BTC/USDT', 'limit': 11642.115000000002, 'order_type': 'limit', 'stake_amount': 980, 'stake_currency': 'USDT', 'fiat_currency': 'USD'}
2019-08-06 22:45:05,928 - freqtrade.wallets - INFO - Wallets synced.
/root/freqtrade/.env/lib/python3.7/site-packages/arrow/factory.py:201: ArrowParseWarning: The .get() parsing method without a format string will parse more strictly in version 0.15.0.See https://github.com/crsmithdev/arrow/issues/612 for more details.
  ArrowParseWarning,
2019-08-06 22:45:05,986 - freqtrade.wallets - INFO - Wallets synced.
2019-08-06 22:45:07,044 - freqtrade.freqtradebot - INFO - Found open order for Trade(id=1, pair=BTC/USDT, amount=0.08417714, open_rate=11642.11500000, open_since=just now)
[...]

PR #2207, opened 2019-09-01
Reenable stoploss_on_exchange for dry-run

xmatthias:
## Summary
Stoploss on exchange does make (limited) sense in dry-run (and it was also implemented, but had a bug for trailing stoplosses).

stoploss on exchange during dry-run assumes to always sell at stop_price - while regular operations can have a stoploss at 5%, and sell at -7% because there was a sharp drop (so it makes sense to leave this setting on during dry-run).

Solve the issue: #2106, revert #2119

## Quick changelog

- Add missing attribute for dry-run stoploss orders
- remove setting `stoploss_on_exchange` to false during dry-run operations.


hroff-1902:
As I said, I'm not quite familiar with the freqtrade logic for stoplosses very deep, but the code LGTM 👍

PR #2217, opened 2019-09-03
Always use config.json if it's available

### LINKED ISSUE #2215: Plot-scripts require --datadir
## Describe your environment

  * Python Version: 3.7
  * Branch: Develop
  * Last Commit ID:  962d487edb0d28f95d6395c09189a333c436fd20
  
## Describe the problem:

Currently, `freqtrade plot-dataframe` does require either a valid configuration (`--config` or `config.json` in cwd - or `--datadir user_data/data/bittrex` to find the backtest data.
This is because without one of these, the exchange is not known, which is a requirement to find the data in the datadir.

## Possible fixes

* Error and point out that one of the 2 conditions have to be met
* add `--exchange` parameter as alternative (including the above)

* ... other ideas?

xmatthias:
## Summary
We should always default to config.json if it exists (also for plot-scripts).

Closes #2215

## Quick changelog

- Use `config.json` if it exists
- use None only if it really doesn't exist and for allowed subcommands
- Validate that either --config or --datadir is specified for plot commands (download-data requires exchange, so validation for this is different).

### DIFF
--- a/freqtrade/configuration/arguments.py
+++ b/freqtrade/configuration/arguments.py
@@ -3,6 +3,7 @@ This module contains the argument manager class
 """
 import argparse
 from typing import List, Optional
+from pathlib import Path
 
 from freqtrade.configuration.cli_options import AVAILABLE_CLI_OPTIONS
 from freqtrade import constants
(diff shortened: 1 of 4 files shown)


hroff-1902:
So you decided to load default config for plot-* subcommands... I did not get yet (sorry, looked into in just for a first glance for a while), what for download-data subcommand and other utils? What's the general decision?

xmatthias:
if `config.json`exists, it's always loaded (for all subommands). The assumption here is that if you have this configuration and don't specify one explicitly, then this is the correct one.

The change for the plot-scripts is to verify if either config is available (explicitly or as `config.json`), or `--datadir` is passed (since we don't know the exchange, and we don't really need to know). 
yes, `--exchange` would be a third option - but then it's a 3 way "either or" - which will be confusing i think.

hroff-1902:
> if `config.json`exists, it's always loaded (for all subommands).

Wait a minute. How's about subcommands in NO_CONF_REQURIED? That list still exists...

xmatthias:
Yes - if `config.json` does not exist (it does not exist on my environment on purpose, and we cannot assume it exists on every user's environment) - then `minimal config` is used - for these specific subcommands. The rest (backtesting, hyperopt, ...) will fail with "Config file not found".

hroff-1902:
So the minimal config is used for those utility subcommands if `config.json` does not exist. Even if it's specified explicitly with `-c`, right? A bit strange, as for me...

Too complicated logic, imho. Need to think about it...

xmatthias:
no, if `--config` is added explicitly, then that config is always used, and no other logic is happening (that would be pointless ...)
(this applies only if the following is true: `parsed_arg.config is None`)

if it's not specified (this is where we need to load the default `config.json` - or minimal config) - then it depends on the needs of the command, if a minimal config is sufficient, or if a full configuration is necessary.

PR #2219, opened 2019-09-03
Make local development start up faster and easier by leveraging docker-compose

ahonnecke:

## Summary

Solve the issue: # Allow for docker-compose usage

## Quick changelog

- Added a docker-compose file

## What's new?
* Allows users to use docker-compose to start an encapsulated version of the application.

### Building
image

### Starting
image

### Stopping
image


xmatthias:
First of all, how do you create a draft PR? :smile: 

Thanks for this submission.
I've been thinking about a docker-compose file myself, but i wasn't sure how to approach this yet (and i wanted to delay it until we are further in moving all "user-data" to the user_data" folder, so it becomes easier).

A few initial thoughs:
i think it should be using the official freqtrade docker image by default (`freqtradeorg/freqtrade:develop` / `freqtradeorg/freqtrade:master`).

You should also not pass in the whole freqtrade folder, since this can create dependency conflicts and errors on startup - you'll need to rebuild the image if you did a new `git pull`.

hroff-1902:
> First of all, how do you create a draft PR? 😄

+1 here 🙄

ahonnecke:
I think that Draft is an option that show up in the dropdown when you submit the PR (at least for me, but I'll click on anything that says "beta" so that might be it)

Sorry that this isn't completely thought out, that why I put it in as a draft, I am planning on using this project (and the PR that adds docker-compose) in my talk at pycolorado2019 on sunday: https://github.com/ahonnecke/pycolorado2019

I think that you're right about pulling the official images, and mounting minimally, but also that we might end up needing multiple different services (or different docker-compose files as well) as ideally one service would be able to mount just the userdata+config and another one would mount everything for development.  Ideally thought I wanted to use this PR to play with the destruction and creation of the SQLite database (to better understand how it works).

Theoretically (as I understand it) docker-compose is more well suited to development, wheras something like hashicorp terraform might be more appropriate for legit IAc

xmatthias:
Ah yes, there i can see that button, too (never noticed it so far though)

## terraform:

well not really - docker-compose can be used in production (we do that at our company) - unless you start to cluster your docker servers (in which case you should be using kubernetes - but that's another beast).

I've not used terraform yet - but i think it's similar to ansible (?) as it allows you to define how the environment shall look like?

## sqlite

The SQlite database is kinda simple - it's automatically created at the location of `--db-url` - so if you make sure it's on a readable/writable mounted folder, that should be fine. We don't destroy it though (which is kind of intentional, as you would loose your trade history and how the bot performed).

technically, you should also be able to spin up a postgres container and use that as `--db-url` - but that's more challenging to backup for users not used to maintaining database systems...

ahonnecke:
I think that you can use terraform to deploy to docker, it's just a layer of vendor abstraction so with terraform configured you can deploy to any of these services: https://www.terraform.io/docs/providers/index.html

xmatthias:
you can do the same with ansible ... so they're really very similar :+1:

ahonnecke:
Is something like that what you had in mind @xmatthias ?

hroff-1902:
I see 'postgress' in the commits...
Are we going to support postgres? I like postgres, more than mysql, but...

xmatthias:
Technically, we do support postgres (although it's not documented, we simply accept a sqlalchemy-string - which can point to any database which has the necessary drivers installed - yes, even oracle).

We default to sqlite though, since that's the easiest to handle for someone without experience in databases ...

hroff-1902:
> we simply accept a sqlalchemy-string - which can point to any database which has the necessary drivers installed - yes, even oracle).

Hijo de la chingada!... 👀 
🤣

ahonnecke:
I was going to put that in a different PR at some point :)

https://github.com/ahonnecke/freqtrade/tree/postgres

Adding it to the project was the topic of the talk that I gave last weekend at PyColorado2019

scripts/pytest.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo "Running Unit tests"
+
+pytest --random-order --cov=freqtrade --cov-config=.coveragerc freqtrade/tests/
```
xmatthias: the path for tests changed from `freqtrade/tests/` to `tests/` in the latest develop versions.

i have a split opinion about this file overall

We can move this command as `command` to the `docker-compose.develop.yml` since it's only one command, or it should include all ci commands (`flake8`, `mypy freqtrade/ scripts/`).

docker-compose.yml
```diff
@@ -0,0 +1 @@
+docker-compose.prod.yml
```
xmatthias: what's the benefit of having this file and `docker-compose.prod.yml` instead of moving the `.prod.yml` file to this?

docker-compose.prod.yml
```diff
@@ -0,0 +1,8 @@
+---
+version: '3'
+services:
+  freqtrade:
+    image: freqtradeorg/freqtrade:master
+    volumes:
```
xmatthias: i guess we could add a (commented out) section about `restart: unless-stopped`?

if running the bot in production, i may want to restart it automatically (i do that with dry-run instances anyway). wouldn't make it default though, as users may have different opinions about that.

docker-compose.yml
ahonnecke: In my experience people seem to not understand the utility of being able to run docker-compose locally.  I did not want to add the prod version as I don't have any way to check that it works.  docker-compose is an extremely useful tool for local development and I think that having a docker-compose.yml that's only for production will be confusing for people that think that docker-compose is only a production tool.

docker-compose.prod.yml
ahonnecke: I'm not sure what you want here.  I want to be able to run this locally with ease, that's what docker-compose offers me as a developer.  I want to contribute to this fork but in order to do so I need to be able to run locally.

docker-compose.yml
ahonnecke: Breaking docker-compose.prod.yml out into it's own file, and then symlinking docker-compose.yml makes it easier to tell what each version of the file is for, and makes it easier to watch and maintain changes since they are explicitly named.

docker-compose.prod.yml
ahonnecke: image

ahonnecke:
Exec example

image

ahonnecke:
Does that all make sense @xmatthias ?

LOCAL.md
```diff
@@ -0,0 +1,48 @@
+# Local development
```
xmatthias: I would move this to `docs/developer.md` - which already has a section about "developer setup".

LOCAL.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+## docker
+``` bash
+sudo apt install docker
+```
+
+## docker-compose
+``` bash
+sudo apt install docker-compose
```
xmatthias: Can we maybe replace these `apt` calls with a link to the docker-documentation and a "make sure to have docker and docker-compose installed"? 

This will work on ubuntu/debian (and install old versions - at least on debian), but will fail on Windows, MacOS, and any other flavor of linux. Having comprehensive documentation on how to install docker should not be part of the freqtrade documentation i think...

docker-compose.develop.yml
```diff
@@ -0,0 +1,9 @@
+---
+version: '3'
+services:
+  freqtrade_develop:
+    build:
+      context: .
+      dockerfile: "./Dockerfile"
```
xmatthias: I think the develop version should use `Dockerfile.develop` ... which installs all requirements, including the ones for tests, and does not start the bot automatically (obviously then you'll probably need the compose alternatives of `-i -t` (currently not sure on the syntax for these)

docker-compose.yml
xmatthias: the problem with a symlink is cross-OS compatibility - regular symlinks work on linux, but tend to fail on windows (never seen a symlink within a git-repository though - maybe it works when checkout is done via windows ...?)

xmatthias:
> Does that all make sense

yes it does, if i were against adding a docker-compose file i would have let you know from the getgo.

There's just a few small things (and the test-script won't work once merged to develop) :+1:

ahonnecke:
https://github.com/freqtrade/freqtrade/pull/2219#discussion_r327930889

Is an excellent point, perhaps building a binary that switched things around (copied the) would be better?  I was hoping that the working local diff wouldn't show docker-compose.yml diffs, but it's not a huge deal.

I think that I got everything else if you'd like to take a look @xmatthias

ahonnecke:
@xmatthias bump

PR #2220, opened 2019-09-03
Allow simplified hyperopt interface

hroff-1902:
This PR allows a user to implement custom hyperopt classes without the `buy_strategy_generator` and `indicator_space` methods if she is not planning optimization for the 'buy' space and without `sell_strategy_generator` and `sell_indicator_space` methods correspondingly if she is not planning optimization for the 'sell' space. Or (and this is the main purpose of this PR) without all of these methods if she is going to optimize the strategy for only 'roi' and 'stoploss'.

If the user runs hyperopt without needed custom methods (for instance, optimizing for 'buy' or 'all' without any of the `buy_strategy_generator` and `indicator_space` methods in the custom hyperopt) then the following exception is generated:
```
freqtrade - ERROR - The 'buy' space is included into the hyperoptimization but indicator_space() method is not found in your custom Hyperopt class. You should either implement this method or remove the 'buy' space from hyperoptimization.
```

How to use:
* remove all those methods from hyperopt;
* run `freqtrade hyperopt --spaces roi stoploss` (works optimizing roi and stoploss);
* run `freqtrade hyperopt` (fails with the error).

TODO: 
* [x] tests (for exceptions generated)
<s>* adjustments into documentation</s>

Please check wordings in the new long exception message.

(P.S. @xmatthias what can you see after removal of those 4 methods from Hyperopt, default or custom? It's the strategy! 🍓 )


xmatthias:
> (P.S. @xmatthias what can you see after removal of those 4 methods from Hyperopt, default or custom? It's the strategy! strawberry )

I'm not certain i agree with that - if you remove everything then yes - but the purpose of hyperopt is defeated as well (roi / stoploss should be the last to optimize, once buy/sell signals are stable.

Apart from that it's ok ... although i doubt the usefulness since you'll still need most of the functions to start hyperopting efficiently - so you'll not remove them from the sample you'll need them in your custom file ....

hroff-1902:
> although i doubt the usefulness since you'll still need most of the functions to start hyperopting efficiently - so you'll not remove them from the sample you'll need them in your custom file ....

Ah, yeah, "show me the money"... 😄 
I will, in the next PR, where the particular application will be shown. It requires additional small change, I'd like to keep it separate from the one...

So, I'll try to add tests for this, allow me document the behavior in the next PR with that particular useful application for this technique...

hroff-1902:
Tests added, pls review. This new functionality will be documented in the next PR along with description of the new technique for hyperopting.

hroff-1902:
or maybe I'll add the rest in this PR... What do you think?

freqtrade/tests/optimize/test_hyperopt.py
```diff
@@ -693,3 +693,271 @@ def test_print_json_spaces_roi_stoploss(mocker, default_conf, caplog, capsys) ->
...
+                         'epochs': 1,
+                         'timerange': None,
+                         'spaces': 'buy',
+                         'hyperopt_jobs': 1, })
+
+    hyperopt = Hyperopt(default_conf)
+    hyperopt.backtesting.strategy.tickerdata_to_dataframe = MagicMock()
+    hyperopt.custom_hyperopt.generate_roi_table = MagicMock(return_value={})
+
+    del hyperopt.custom_hyperopt.__class__.buy_strategy_generator
```
xmatthias: I think the handling of the fail-cases could be greatly simplified by using pytest.mark.parametrize ...

you can use `delattr(hyperopt.custom_hyperopt.__class__, 'buy_strategy_generator')` to delete stuff dynamically, so you pass in `"buy", "buy_strategy_generator"` as one testcase...

all 4 "failed" cases are identical apart from 2 lines, which makes them ideal candidates for parametrize (we have samples of this in backtesting, `test_pairlist.py` and `test_exchange.py`).

hroff-1902:
how to 'pull' travis check again?

Issue #2112, opened 2019-08-07
Installation future - move to pypi📦 


xmatthias:
Since we've had many questions regarding installation lately, below is where i see freqtrade going in this regards.

## Goal

* Easily installable using a single command (`pip install freqtrade` / `docker pull freqtradeorg/freqtrade`).
* getting setup in another 2-3 commands (`freqtrade create-userdir`, `freqtrade generate-config`(?))

### Steps to reach the goal:

* [x] user_data configurable (#2067)
* [x] move default configuration to user_data directory (#2922)
* [x] move tests out of freqtrade module folder (#2234, #2236)
* [x] copy templates / default configurations when using `freqtrade create-userdir` (to have them available when installing through pip) (#2538 )
* [x] freqtrade subcommand to easily configure a "starter" configuration (similar to `setup.sh`) #2831
* [x] split hyperopt dependencies from `setup.py` (so `pip install freqtrade[hyperopt]` installs hyperopt only dependencies). #2303 
* [ ] having ta-lib as optional dependency (there are other libraries out there, which do mostly the same using numpy/pandas functions only). (further discussion in #2509)
* [x] installable through `pip install freqtrade` 🎉 📦 
* [ ] have documentation point to pypi installability 
* [x] Automated deployment of new releases to pypi #2941
* [x] cleanup installation documentation to have fewer possible routes for installation, which is quite confusing currently
* [ ] remove `setup.sh` since it's not needed anymore

The sequence of these steps is quite flexible - also new steps may be introduced if necessary.

hroff-1902:
Yes, we need move to pypi ASAP! 😆 
Will react in #2067

PR #2221, opened 2019-09-04
Notebook

jraviotta:
Added function to locate project root which is useful for notebooks.
Fixes to typos.

### DIFF
--- a/freqtrade/data/history.py
+++ b/freqtrade/data/history.py
@@ -61,7 +61,7 @@ def load_tickerdata_file(datadir: Path, pair: str, ticker_interval: str,
                          timerange: Optional[TimeRange] = None) -> Optional[list]:
     """
     Load a pair from file, either .json.gz or .json
-    :return: tickerlist or None if unsuccesful
+    :return: tickerlist or None if unsuccessful
     """
     filename = pair_data_filename(datadir, pair, ticker_interval)
     pairdata = misc.file_load_json(filename)
--- a/freqtrade/plot/plot_utils.py
+++ b/freqtrade/plot/plot_utils.py
@@ -17,11 +17,11 @@ def start_plot_dataframe(args: Namespace) -> None:
     Entrypoint for dataframe plotting
     """
     # Import here to avoid errors if plot-dependencies are not installed.
-    from freqtrade.plot.plotting import analyse_and_plot_pairs
+    from freqtrade.plot.plotting import load_and_plot_trades
     validate_plot_args(args)
     config = setup_utils_configuration(args, RunMode.PLOT)
 
-    analyse_and_plot_pairs(config)
+    load_and_plot_trades(config)
 
 
 def start_plot_profit(args: Namespace) -> None:
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -3,7 +3,6 @@ from pathlib import Path
 from typing import Any, Dict, List
 
 import pandas as pd
-
 from freqtrade.configuration import TimeRange
 from freqtrade.data import history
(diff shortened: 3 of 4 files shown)


freqtrade/misc.py
```diff
@@ -114,3 +115,23 @@ def deep_merge_dicts(source, destination):
             destination[key] = value
 
     return destination
+
+
+def find_project_root(path=None):
```
xmatthias: I'm not sure if this will work correctly (honestly, i think it'll fail for most cases)

Your assumption is that you've got the repository cloned and you're in a jupyter notebook within `user_data/`.

As pointed out in #2112 - one goal is for `freqtrade create-datadir` to copy the contents of the current user_dir to the newly created directory, so users have the examples available.
While it's not doing that yet, it's not a huge step to implement that (and it's on my todo-list ... among other things)
Also, #2112 's goal is to install from pypi - in which case a `git clone` won't even happen

Assuming i install freqtrade reguilarily then cd to ~ and create a userdir there:

``` bash
pip install .[all]
# pip install freqtrade  ## for the pypi future 
cd ~
freqtrade create-userdir --userdir ~/ft_userdata/
```

I can now run the bot with `--userdir ~/ft_userdata/` - and i can version-control this folder in my private repo (it contains all my profitable strategies i'm not ready to share...)

I also have the notebook samples there - but i cannot find the projects-directory - since its 

a) not relevant - it's somewhere in `/usr/.../pip/site-packages/...`
b) not one of my parents

jraviotta: I'll pull this out. I can see how fragile this can be.

freqtrade/tests/test_plotting.py
```diff
@@ -30,7 +30,7 @@ def find_trace_in_fig_data(data, search_string: str):
     return next(matches)
 
 
-def generage_empty_figure():
```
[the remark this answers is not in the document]
jraviotta: done

freqtrade/misc.py
xmatthias: btw, i think you somehow messed the 2 branches for this and #2222  - they seem to have the same (similar?) changes now ...

jraviotta: It is very possible that I mangled something. Do you want me to start a new PR with the useful changes on develop or, can you accept this and I'll rebase the other on top of the new develop?

xmatthias: Accepting this in it's current state won't be possible since it has import-errors which would prevent the bot from running.

i think it's best to start a new branch from develop and then forcepush this on top of this so we don't loose comments and such.

to not loose your changes best do the following:

``` bash
git checkout -b new_notebook  # create new branch
git merge notebook --no-commit # add changes from this branch - without commiting
git reset  # reset state to "dirty" - most tools (vscode for example) will now highlight changed lines

### Now remove all changes that are unwanted
### then:

git add ... # whatever files are still left
git commit

# you can then rename the branches and force-push on top of this
git branch -m notebook notebook_old
git branch -m notebook_new notebook
git push -u origin notebook # Make sure to replace origin with the remote of your git repository

```

jraviotta: Thanks for the coaching. I think the last push fixed it. Hopefully you can accept these changes.

xmatthias:
The changes look reasonable overall - but please track down all usages of `analyse_and_plot_pairs` and replace them with the new function ... (can't accept as long as CI fails - it's there for a reason :smile: ).

jraviotta:
fixed. Thought i removed those, probably got messed up with my git errors. sorry.

xmatthias:
I've rebased this on top of the latest develop and fixed conflicts - if CI passes this should be good :+1:

PR #2205, opened 2019-08-31
Release 2019.8-1 - hotfix data-dir not including exchange

xmatthias:
## Summary
Same as #2199 - but against the master branch.

The only change is this fix.

PR #2222, opened 2019-09-04
Plotting

jraviotta:
Renames `analyse_and_plot_pairs` to `load_and_plot_trades` which is more accurate.
Deleted `__slots__` memory optimization which caused `TypeErrors` in StrategyResolver

### DIFF
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from typing import Any, Dict, List
 
 import pandas as pd
-
+from freqtrade import OperationalException
 from freqtrade.configuration import TimeRange
 from freqtrade.data import history
 from freqtrade.data.btanalysis import (combine_tickers_with_mean,
@@ -33,9 +33,13 @@ def init_plotscript(config):
         pairs = config["pairs"]
     else:
         pairs = config["exchange"]["pair_whitelist"]
+    if pairs is None:
+        raise OperationalException('No pairs available in config.')
 
     # Set timerange to use
     timerange = TimeRange.parse_timerange(config.get("timerange"))
+    if timerange is None:
+        raise OperationalException('Could not parse timerange in config.')
 
     tickers = history.load_data(
         datadir=Path(str(config.get("datadir"))),
@@ -43,11 +47,15 @@ def init_plotscript(config):
         ticker_interval=config.get('ticker_interval', '5m'),
         timerange=timerange,
     )
+    if tickers is None:
+        raise OperationalException('No ticker data available as specified in config.')
 
     trades = load_trades(config['trade_source'],
                          db_url=config.get('db_url'),
                          exportfilename=config.get('exportfilename'),
                          )
+    if trades is None:
+        raise OperationalException('No trades available as specified in config.')
 
     return {"tickers": tickers,
             "trades": trades,
@@ -324,7 +332,7 @@ def store_plot_file(fig, filename: str, directory: Path, auto_open: bool = False
     logger.info(f"Stored plot as {_filename}")
 
 
-def analyse_and_plot_pairs(config: Dict[str, Any]):
+def load_and_plot_trades(config: Dict[str, Any]):
     """
     From configuration provided
     - Initializes plot-script
@@ -339,7 +347,7 @@ def analyse_and_plot_pairs(config: Dict[str, Any]):
 
     plot_elements = init_plotscript(config)
     trades = plot_elements['trades']
-
+    
     pair_counter = 0
     for pair, data in plot_elements["tickers"].items():
         pair_counter += 1
--- a/freqtrade/resolvers/strategy_resolver.py
+++ b/freqtrade/resolvers/strategy_resolver.py
@@ -24,8 +24,6 @@ class StrategyResolver(IResolver):
     This class contains all the logic to load custom strategy class
     """
 
-    __slots__ = ['strategy']
-
     def __init__(self, config: Optional[Dict] = None) -> None:
         """
         Load the custom class from config parameter
--- a/config_full.json.example
+++ b/config_full.json.example
@@ -38,6 +38,7 @@
     "order_types": {
         "buy": "limit",
         "sell": "limit",
+        "emergencysell": "market",
         "stoploss": "market",
         "stoploss_on_exchange": false,
         "stoploss_on_exchange_interval": 60
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -192,19 +192,20 @@ end up paying more then would probably have been necessary.
 
 ### Understand order_types
 
-The `order_types` configuration parameter contains a dict mapping order-types to
-market-types as well as stoploss on or off exchange type and stoploss on exchange
-update interval in seconds. This allows to buy using limit orders, sell using
-limit-orders, and create stoploss orders using market. It also allows to set the
-stoploss "on exchange" which means stoploss order would be placed immediately once
-the buy order is fulfilled. In case stoploss on exchange and `trailing_stop` are
-both set, then the bot will use `stoploss_on_exchange_interval` to check it periodically
-and update it if necessary (e.x. in case of trailing stoploss).
-This can be set in the configuration file or in the strategy.
-Values set in the configuration file  overwrites values set in the strategy.
+The `order_types` configuration parameter maps actions (`buy`, `sell`, `stoploss`) to order-types (`market`, `limit`, ...) as well as configures stoploss to be on the exchange and defines stoploss on exchange update interval in seconds.
 
-If this is configured, all 4 values (`buy`, `sell`, `stoploss` and
-`stoploss_on_exchange`) need to be present, otherwise the bot will warn about it and fail to start.
+This allows to buy using limit orders, sell using
+limit-orders, and create stoplosses using using market orders. It also allows to set the
+stoploss "on exchange" which means stoploss order would be placed immediately once
+the buy order is fulfilled.
+If `stoploss_on_exchange` and `trailing_stop` are both set, then the bot will use `stoploss_on_exchange_interval` to check and update the stoploss on exchange periodically.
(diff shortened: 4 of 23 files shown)


freqtrade/misc.py
```diff
@@ -114,3 +115,23 @@ def deep_merge_dicts(source, destination):
             destination[key] = value
 
     return destination
+
+
+def find_project_root(path=None):
```
xmatthias: see notes in #2221 about this function ...

freqtrade/resolvers/strategy_resolver.py
```diff
@@ -24,8 +24,6 @@ class StrategyResolver(IResolver):
     This class contains all the logic to load custom strategy class
     """
 
-    __slots__ = ['strategy']
```
xmatthias: Would you be able to explain what errors you encounter (in which circumstances) due to this?

I'd like to understand what errors you're encountering, since i've never seen a problem due to this.

freqtrade/plot/plotting.py
```diff
@@ -339,31 +342,34 @@ def analyse_and_plot_pairs(config: Dict[str, Any]):
 
     plot_elements = init_plotscript(config)
     trades = plot_elements['trades']
+    if trades is None:
+        raise ValueError('No trades to analyze')
```
xmatthias: This is very limiting and i don't think it makes sense to handle this here

Not having trades should not fail the plot-dataframe - it'll simply plot no trades - which can be usefull to analyze a strategy while developing it (assume it creates no trades in backtesting - i would still like to see the indicators and maybe from seeing them i can determine why they're not generating sells).

if it does create errors, we should fix them where they happen - but when i tested this it did create log-outputs indicating that there are no trades.

freqtrade/plot/plotting.py
```diff
@@ -339,31 +342,34 @@ def analyse_and_plot_pairs(config: Dict[str, Any]):
...
-            indicators2=config["indicators2"],
-        )
-
-        store_plot_file(fig, filename=generate_plot_filename(pair, config['ticker_interval']),
-                        directory=config['user_data_dir'] / "plot")
+            store_plot_file(fig, filename=generate_plot_filename(pair, config['ticker_interval']),
+                            directory=config['user_data_dir'] / "plot")
 
-    logger.info('End of plotting process. %s plots generated', pair_counter)
+            logger.info('End of plotting process. %s plots generated', pair_counter)
```
xmatthias: i think this is one intend too deep - it should be after the for loop, not within.

freqtrade/plot/plotting.py
```diff
@@ -33,21 +33,24 @@ def init_plotscript(config):
         pairs = config["pairs"]
     else:
         pairs = config["exchange"]["pair_whitelist"]
+    assert pairs is not None, ('No pairs available in config.')
```
xmatthias: The checks you add here with `assert` do make sense.

However, please don't use assert statements in running code (use if / `raise OperationalException(message)` instead). 

I know this is sometimes suggested for analytic workflows (maybe also elsewhere?) - but python can be run in an "optimized" mode (`python -O`) where asserts are not evaluated / skipped.

It's not a problem everyone will run into - but some might (we have nothing that would prevent this mode at the moment, and i'd like to keep it that way).

freqtrade/plot/plotting.py
```diff
@@ -33,21 +33,24 @@ def init_plotscript(config):
...
 
     # Set timerange to use
     timerange = TimeRange.parse_timerange(config.get("timerange"))
 
     tickers = history.load_data(
-        datadir=Path(str(config.get("datadir"))),
+        datadir=Path(config.get("datadir"), config.get("exchange").get("name")),
```
xmatthias: This will fail with the latest develop (and master) code - resulting in `user_data/data/binance/binance`.

The fix for this #2199 / #2205 was merged a few days ago to both branches.

freqtrade/misc.py
jraviotta: removed

freqtrade/plot/plotting.py
jraviotta: Given that I'm working on a strategy and want to visualize the results of a backtest
When I run the following in a Jupyter notebook:
```python
from freqtrade.plot.plotting import load_and_plot_trades
load_and_plot_trades(config)
```
I get:
```
2019-09-04 20:06:52,946 - freqtrade.data.history - INFO - Using local backtesting data ...
2019-09-04 20:06:57,340 - freqtrade.plot.plotting - INFO - analyse pair ETH/BTC
 AttributeError: 'NoneType' object has no attribute 'loc' 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-20-0263847289e7> in <module>
      1 from freqtrade.plot.plotting import load_and_plot_trades
----> 2 load_and_plot_trades(config)

~/Documents/Repos/freqtrade/freqtrade/plot/plotting.py in load_and_plot_trades(config)
    356         logger.debug(f'pair: {pair}')
    357 
--> 358         trades_pair = trades.loc[trades['pair'] == pair]
    359         trades_pair = extract_trades_of_period(dataframe, trades_pair)
    360 

AttributeError: 'NoneType' object has no attribute 'loc'
```

This occurs because `init_plotscript()` generates `None` for the trades dataframe.

```
from freqtrade.plot.plotting import init_plotscript
plot_elements = init_plotscript(config)
display(plot_elements)

2019-09-04 20:11:41,978 - freqtrade.data.history - INFO - Using local backtesting data ...
{'tickers': {'ETH/BTC':                          date      open      high       low     close   volume
  0   2019-09-02 02:55:00+00:00  0.017519  0.017528  0.017495  0.017505  692.633
  1   2019-09-02 03:00:00+00:00  0.017506  0.017540  0.017505  0.017540  322.632
  ..                        ...       ...       ...       ...       ...      ...
  198 2019-09-02 19:25:00+00:00  0.017524  0.017533  0.017511  0.017514  499.837
  
  [199 rows x 6 columns]}, 'trades': None, 'pairs': ['ETH/BTC']}
```

My solution was to halt execution of `load_and_plot_trades` because generating a figure with no trades is not useful if I'm tweaking a strategy and visualizing results in a notebook.

freqtrade/resolvers/strategy_resolver.py
jraviotta: In a Jupyter notebook, including `__slots__ = ['strategy']` results in `TypeError`: 

```
from freqtrade.plot.plotting import load_and_plot_trades
load_and_plot_trades(config)
2019-09-04 20:33:59,904 - freqtrade.resolvers.iresolver - INFO - Using resolved strategy NewStrategy from '/Users/jonathanraviotta/Documents/Repos/freqtrade/user_data/user_repo/strategies/new_strategy.py'...
 TypeError: descriptor 'strategy' for 'StrategyResolver' objects doesn't apply to 'StrategyResolver' object 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-25-0263847289e7> in <module>
      1 from freqtrade.plot.plotting import load_and_plot_trades
----> 2 load_and_plot_trades(config)

~/Documents/Repos/freqtrade/freqtrade/plot/plotting.py in load_and_plot_trades(config)
    339     :return: None
    340     """
--> 341     strategy = StrategyResolver(config).strategy
    342 
    343     plot_elements = init_plotscript(config)

~/Documents/Repos/freqtrade/freqtrade/resolvers/strategy_resolver.py in __init__(self, config)
     38         self.strategy: IStrategy = self._load_strategy(strategy_name,
     39                                                        config=config,
---> 40                                                        extra_dir=config.get('strategy_path'))
     41 
     42         # make sure experimental dict is available

TypeError: descriptor 'strategy' for 'StrategyResolver' objects doesn't apply to 'StrategyResolver' object
```

freqtrade/plot/plotting.py
jraviotta: removed the error check i the function and fixed indenting.

freqtrade/plot/plotting.py
jraviotta: I replaced the `assert` statements with `OperationalException`

freqtrade/plot/plotting.py
jraviotta: I modified `freqtrade.data.history.load_pair_history` Warning message to surface this solution.

freqtrade/plot/plotting.py
xmatthias: as mentioned above, especially early in strategy development it can be useful to just see the indicators (without having any buy signals yet), even without any trades happening.

I would handle this differently - however i wonder why this does not happen when running via cli - i guess you've got some parameters in config set "wrongly" so trades do not load from either DB nor backtesting - and a potential error is probably not shown in juptyer...

For jupyter - i don't think this is the ideal entry-point anyway - you should rather use the sample notebook to analyze the strategy - and you can call `generate_candlestick_graph()` from there on the dataframe directly and visualize the results in-line within the notebook (or open a new window to see it fullscreen).

That way you can avoid to set complicated config options which will vary deeply between iterations.

freqtrade/plot/plotting.py
xmatthias: did you forget to push ...? (or pushed to the wrong branch?) as u i don't seem to see these changes...

freqtrade/resolvers/strategy_resolver.py
```diff
@@ -123,7 +121,8 @@ def _load_strategy(
         current_path = Path(__file__).parent.parent.joinpath('strategy').resolve()
 
         abs_paths = [
-            config['user_data_dir'].joinpath('strategies'),
+            Path(config['user_data_dir']).joinpath('strategies'),
+            config['strategy_path'],
```
xmatthias: This line is killing CI ... strategy_path is not guaranteed to be available (only if you specify `--strategy-path xxx`, so this line will almost always fail.

It's also a duplicate entry (so it's searching this location twice) since `extra_dir` will contain exactly this path if it's available.

Issue #1687, opened 2019-03-21
Implement all available Kraken order types


iuvbio:
I've been running FT on Kraken stably since it was implemented now, so I think it's time to implement some more functionality.

As per the [docs](https://www.kraken.com/features/api#private-user-trading) the following order types are available:

- [x] market
- [x] limit (price = limit price)
- [x] stop-loss (price = stop loss price)
- [ ] take-profit (price = take profit price)
- [ ] stop-loss-profit (price = stop loss price, price2 = take profit price)
- [ ] stop-loss-profit-limit (price = stop loss price, price2 = take profit price)
- [ ] stop-loss-limit (price = stop loss trigger price, price2 = triggered limit price)
- [ ] take-profit-limit (price = take profit trigger price, price2 = triggered limit price)
- [ ] trailing-stop (price = trailing stop offset)
- [ ] trailing-stop-limit (price = trailing stop offset, price2 = triggered limit offset)
- [ ] stop-loss-and-limit (price = stop loss price, price2 = limit price)
- [ ] settle-position

There seem to be a lot of things we could do in one order call like trailing stoploss, roi.. But I'll first concentrate on stop-loss-limit, since that's what we have for Binance and probably the most useful.

*Edit @xmatthias*: 
I'm adding the below here as i think it's important info without scrolling through all comments:

Almost all advanced ordertypes are currently disabled on kraken [source](https://blog.kraken.com/post/1234/announcement-delisting-pairs-and-temporary-suspension-of-advanced-order-types/).
The only ones enabled are "stop-loss" and "take-profit" [source](https://blog.kraken.com/post/1494/kraken-enables-advanced-orders-and-adds-10-currency-pairs/)

xmatthias:
The problem with the take_profit orders is that they prevent you from setting a stoploss at the same time (you'd tie up your balance twice).

I would always prefer a stoploss-on-exchange over a take-profit order on exchange
however trailing-stop looks interresting ...

mishaker:
some exchanges do allow you to put stoploss and take profit at the same time (not binance anyway). it is called OCO (one cancels other) orders if I am not mistaken. 
otherwise another approach can be cancel SL if the spot is higher enough and cancel TP if the price approaches the buy. but it is too risky in case of shake out.

mishaker:
Just realized Kraken allows you to put both (SL and TP) since a while. it is called advanced orders. here is the article: 
[...]

PR #2227, opened 2019-09-05
Fix RPC /balance failure

### LINKED ISSUE #2225: Kraken: EUR zero balance, BTC unknown format
Yes, I am done my research. I found what I believe was a similar issue that was resolved but has not resurfaced --> #1840  

  * Operating system: Ubuntu 19.04
  * Python Version: Python 3.7.3 
  * CCXT version: ccxt==1.18.1085 
  * Branch: Master
  * Last Commit ID: a7e45c5a732c60e2815bfd7bd3f2cb4cb24faf56 

 
  
I am wanting to use Kraken as my exchange. when using BTC I receive the below error:

 **File "/home/sgarc90/freqtrade/freqtrade/rpc/telegram.py", line 52, in wrapper
  File "/home/sgarc90/freqtrade/freqtrade/rpc/telegram.py", line 336, in _balance
    "\t`Est. BTC: {est_btc: .8f}`\n".format(**currency)**
--------------------------
When I change to EUR. a zero balance is returned.

WARNING - Unable to create trade: Available balance(0 EUR) is lower than stake amount(19 EUR)

### Steps to reproduce:
once you start up freqtrade using Kraken. from within telegram send the balance button and watch the screen is a short time the telegram request will error out.

I just happened to see the EUR error appear on the screen when freqtrade tried to place an order

xmatthias:
## Summary
`/balance`  should not fail if None is reported by ccxt, since this is a valid return value according to their documentation.

In addition to that i discovered that Kraken does not correctly report "free" balances, making detecting of available balances impossible.
Details and a possible fix are documented in the Kraken issue https://github.com/freqtrade/freqtrade/issues/1687#issuecomment-528509266
.
Until we have that fix implemented / developed, we will need to block kraken, since trading on it will not be possible.

Close #2225

## Quick changelog:

* fix /balance failure
* block Kraken trading since it's not working due to wrong return values for `get_balance()` calls

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -30,6 +30,9 @@ BAD_EXCHANGES = {
     "bitmex": "Various reasons",
     "bitstamp": "Does not provide history. "
                 "Details in https://github.com/freqtrade/freqtrade/issues/1983",
+    "kraken": "TEMPORARY: Balance does not report free balance, so freqtrade will not know "
+              "if enough balance is available."
+              "Details in https://github.com/freqtrade/freqtrade/issues/1687#issuecomment-528509266"
     }
 
 
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -294,9 +294,9 @@ class RPC(object):
             total = total + est_btc
             output.append({
                 'currency': coin,
-                'free': balance['free'],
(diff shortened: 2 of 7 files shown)


hroff-1902:
LGTM 👍 

(However, I personally would like to see 'None' (or 'N/A') in the messages instead of zeros...)

PR #2228, opened 2019-09-05
Plotting

jraviotta:
Requested changes to plotting


jraviotta:
see also #2222

freqtrade/resolvers/strategy_resolver.py
```diff
@@ -123,7 +121,8 @@ def _load_strategy(
         current_path = Path(__file__).parent.parent.joinpath('strategy').resolve()
 
         abs_paths = [
-            config['user_data_dir'].joinpath('strategies'),
+            Path(config['user_data_dir']).joinpath('strategies'),
+            config['strategy_path'],
```
xmatthias: As mentioned in the other PR - adding  'strategy_path' here will result in searching this branch twice, since it's already passed in as `extra_dir`.

freqtrade/resolvers/strategy_resolver.py
```diff
@@ -24,8 +24,6 @@ class StrategyResolver(IResolver):
     This class contains all the logic to load custom strategy class
     """
 
-    __slots__ = ['strategy']
```
xmatthias: Note to self: need to test this! reference of experiened error:

https://github.com/freqtrade/freqtrade/pull/2222#discussion_r321032785

freqtrade/resolvers/strategy_resolver.py
jraviotta: Fixed. I deleted the extra strategy path but left in the Path wrapper because:
```python
print(config['user_data_dir'].joinpath('strategies'))

AttributeError: 'str' object has no attribute 'joinpath'
```

xmatthias: in this case, your setting of `config['user_data_dir']` is wrong / you're probably using it from a notebook, which may cause problems since not everything is initialized correctly.

The following lines assign user_data_dir with the result of `create_userdata_dir()` ... which always returns a Path object.

https://github.com/freqtrade/freqtrade/blob/f987e6e0f9bdcae6f0cec3a273e270a3a35a3c12/freqtrade/configuration/configuration.py#L169-L176


Using config within a notebook will still need some more love / work (it's on my long todo-list) - but i would not cover the problems caused by this in other areas of the code if they cannot happen (all regular operations do run through all / most of the sections in the configuration class).

freqtrade/plot/plotting.py
```diff
@@ -358,7 +358,7 @@ def load_and_plot_trades(config: Dict[str, Any]):
         )
 
         store_plot_file(fig, filename=generate_plot_filename(pair, config['ticker_interval']),
-                        directory=config['user_data_dir'] / "plot")
```
xmatthias: i would leave this as `/` - see other comment as to why `user_data_dir` is wrong in your environment.

freqtrade/resolvers/strategy_resolver.py
jraviotta: That is the problem. In the notebook, I was using:
```python
candles = load_pair_history(datadir=config['datadir'],
                            ticker_interval=ticker_interval,
                            pair=pair)
```
The dict index retrieves the datadir as a str. I can work around it by wrapping the argument with `Path()`. I will be interested to see how you want to handle calling internal functions from a notebook. It would be nice to arbitrarily pass the Configuration object to any function.

freqtrade/plot/plotting.py
jraviotta: I can use `Path()` in the function call here too, but the concatenation seems sloppy. As a Windows user, I am scared of `/`.

freqtrade/resolvers/strategy_resolver.py
```diff
@@ -122,7 +120,7 @@ def _load_strategy(
         current_path = Path(__file__).parent.parent.joinpath('strategy').resolve()
 
         abs_paths = [
-            config['user_data_dir'].joinpath('strategies'),
+            Path(config['user_data_dir']).joinpath('strategies'),
             current_path,
         ]
```
jraviotta: I'm not sure why this is necessary:
```python
strategy = StrategyResolver({'user_data_dir': Path(config['user_data_dir']),
                       'strategy_path': Path(config['strategy_path'])}).strategy
```
Given that `config = Configuration.from_files(config_files)`, shouldn't 
`config['user_data_dir']` == Path object?

But when I run:
```python
strategy = StrategyResolver(config=config).strategy
```

I get:
```python
    122         abs_paths = [
--> 123             config['user_data_dir'].joinpath('strategies'),
    124             current_path,
    125         ]

AttributeError: 'str' object has no attribute 'joinpath'
```

jraviotta:
I got it working. :) Now strategies can be developed using Jupyter notebook to visualize backtests.

freqtrade/plot/plotting.py
xmatthias: Please have a look at the [pathlib](https://docs.python.org/3/library/pathlib.html#basic-use) documentation.

It's main purpose is to make handling cross-compatibility easy by abstracting this away - using `Path("xxx") / "otherpath"` is the same as `Path("xxx").joinpath("otherpath")`.

This did never really work with `os.path` and could create problems - that's why i'm trying to move avay from using `os.path` (slowly, but i try to replace usage whenever i am working on something using `os.path`).

if you look at the implementation of the 2 methods (below) (`/` is implemented as `__truediv__()`) you'll notice that they call exactly the same method internally - so are factually identical.

https://github.com/python/cpython/blob/e0a1561f93545f38034d6c48f21e79acfd4db04b/Lib/pathlib.py#L899-L908

jraviotta: Ah, ok. Thanks for that explanation. So, the opaque requirement is that `config['somepath']` is a path object and not a string. That conversion is bypassed when I set dictionary values directly in a notebook with `config['newpath'] = newpath`.

However, that doesn't explain the behavior in `freqtrade/resolvers/strategy_resolver.py ` noted below. `abs_path` should be pulling a path object from the config object since I'm not messing with the key:value pair after using `config_from_file()`. Perhaps `config['user_data_dir']` is not being converted to a path object in `config_from_file()`?

It seems like a solution could be to create a config editing function for overriding config values inline. For example:
```python
# generate config object
config = config_from_files(config_files)
# this should now be true
type(config['userdir']) == Path object

# override some value 
config['userdir'] = config_value_from_str('user_data/user_repo')

# or
config['strategy'] = config_value_from_str('NewStrategy')
```
Where `config_value_from_str` does the appropriate type interpolation and updates the config dict.

I wonder if that is the cause of the error noted in #2222 where `__slots__ = ['strategy']` gets confused about type?

xmatthias: I think config_from_files does not do what you expect it to do - which is return a fully-initialized config object.

it returns merged dicts - nothing more, nothing less. That's partially intentional, since user_data_dir will create subdirectories if necessary ... which may not be desired for notebook usage.

I'll think of a solution to this outside of this PR - for now i would set `config['user_data_dir'] = Path("somepath")` to mitigate problems you're encountering - especially since you seem to (for now) be the only one with problems in this regard... (however we will fix this eventually, but a little (bigger?) change to config might be required for that to work correctly.

jraviotta: That makes complete sense. As I have said, these practical challenges are about expanding my understanding of Python and how experienced developers solve common problems. I am more interested in the process than the solution. I look forward to seeing how you tackle this issue and am willing to help with whatever is useful to the project.

xmatthias:
## jupyter notebook

The whole notebook change seems random / non intentional to me...
* It adds paths only available to your environment `chdir(Path(Path.home(), 'Documents', 'Repos', 'freqtrade'))`, 
* completely changes how it did look before
* removes all documentation it had ...
* it includes outputs (please clear cells before committing ...)

## backtesting / downloading
I think i've mentioned this before - but i find it pointless to run commands (like backtesting and such) from a jupyter notebook. The output will be cramped into a small cell, you need to apply workarounds to even get it working... and especially in the case of downloading data - this can be running for hours - so a notebook is not the right place for that i think.

I don't mind you using that if that fits your workflow and you can work around problems ... but i would not document this as an official way to run freqtrade, since we don't know which problems may araise, and jupyter shall be a way to analyze data, not to control the bot.

### plotting interactively

I would like the plotting part you have at the bottom of this notebook - however without completely changing the whole notebook (see, that's one problem i have with notebooks, work with it get's messy soon).

I would however not rely this heavily on having everything in config.
`indicators1` and `indicators2` is a good example for this - it'll vary heavily while working on a strategy (i may throw out one indicator in favor of another ...), so this should be a simple list.

i would see it as "next cells" to the previous version of the notebook (either before, or after `Display the trade details`), so to use it you'd have to run a few of the existing cells, and then one or more "plot" cells.

An additional hint for users could be to use different renderers - i personally like "browser" since it's showing the graph full screen ( for usage - see below) ... but this may vary, and hinting at the plotly documentation for more details might not hurt.

```
graph.show(renderer="browser")
```

jraviotta:
Try this version.

jraviotta:
I added some returns in the plotting script to provide access to the plotting objects. I also turned off auto open on the profit fig so a jupyter user can control where the plot shows up. That broke CI.

I like this workflow and tried to show the benefits in the example notebook. Hopefully a little work on a configuration resolver can streamline access to the config dict with correct types so the function calls in the notebook are not so ugly.

freqtrade/plot/plotting.py
```diff
@@ -359,16 +360,23 @@ def load_and_plot_trades(config: Dict[str, Any]):
...
         store_plot_file(fig, filename=generate_plot_filename(pair, config['ticker_interval']),
                         directory=config['user_data_dir'] / "plot")
+        plot_data[generate_plot_filename(pair,
+                                         config['ticker_interval'])] = {"fig": fig, 
+                                                                        "data": dataframe, 
+                                                                        "trades": trades}
 
     logger.info('End of plotting process. %s plots generated', pair_counter)
+    logger.info(f'fig, data, trades are available with the keys {list(plot_data)}')
+    return plot_data
```
xmatthias: Returning this here is pointless, if you use `generate_candlestick_graph()` directly in the notebook (as i said multiple times before) then this is not necessary.

You're reloading **everything** from file, if that's the intend, then you dont' need a notebook but can run the command as cli command as well.

hroff-1902:
> these practical challenges are about expanding my understanding of Python and how experienced developers solve common problems. I am more interested in the process than the solution.

Here comes a challenge how project managers solve the problems:

* I'm closing the PR, we cannot spare our resources on it anymore. Nothing personal, I'm sorry. It will never complete if it continues same way.
* @xmatthias please cherry pick or copy-paste the useful changes from this in a separate branch, and make the corresponding PR. Please leave anything that does not fit into "Requested changes to plotting" out of scope.

jraviotta:
@xmatthias, @hroff-1902, thank you for the time you invested in this. I appreciate the feedback and apologize for having wasted your time on errors. I obviously overlooked the solutions in process and proposed changes to stale problems.

I will fork the project and modify independently. I will refrain from contributing to freqtrade unless I know the proposed solution is solid.

Alternatively, I would be happy to help if there are specific tasks that might be suited for my admittedly amature  skills. Perhaps I can help with documentation testing, or project management.

I know that it may seem like I was ignoring your directions, but I submitted code that worked against the development branch. Some of the prior suggestions required unmerged edits. Since there is no public development plan, we were working at cross purposes.

Again, I apologize for being a resource drain and hope that I can contribute something useful in the future. I believe Freqtrade is needed in the python ecosystem and you guys are obviously very talented developers. I am grateful to be able to watch your development process unfold.

PR #2230, opened 2019-09-06
docs: Fix table with ROI limits in hyperopt.md

hroff-1902:
Variant with colspans is not rendered at freqtrade.io as well.


xmatthias:
Looking at the rendered table now - maybe we should transpose the table? 
so fewer columns and more rows ... ?

hroff-1902:
I'll probably move it to a html table with colspans later.
Transposing may also have sense, thanks!

PR #2232, opened 2019-09-06
PoC: Allow hyperopting on strategies

hroff-1902:
Allows to use approach, described in #2231 -- hyperopting using the custom (or even default) strategy,

In user_data/strategies/sample_strategy.py:
```
...
from freqtrade.optimize.hyperopt_interface import IHyperOpt
...

class SampleStrategy(IStrategy, IHyperOpt):
...
// The strategy as it is
```
then run hyperopting:
```
$ freqtrade hyperopt --hyperopt-path user_data/strategies/ --customhyperopt SampleStrategy --spaces roi stoploss
```

### DIFF
--- a/freqtrade/optimize/hyperopt_interface.py
+++ b/freqtrade/optimize/hyperopt_interface.py
@@ -36,6 +36,9 @@ class IHyperOpt(ABC):
     """
     ticker_interval: str
 
+    def __init__(self, config: dict = None) -> None:
+        self.config = config
+
     @staticmethod
     @abstractmethod
     def populate_indicators(dataframe: DataFrame, metadata: dict) -> DataFrame:
--- a/freqtrade/resolvers/hyperopt_resolver.py
+++ b/freqtrade/resolvers/hyperopt_resolver.py
@@ -65,7 +65,7 @@ class HyperOptResolver(IResolver):
             abs_paths.insert(0, Path(extra_dir).resolve())
 
         hyperopt = self._load_object(paths=abs_paths, object_type=IHyperOpt,
-                                     object_name=hyperopt_name)
+                                     object_name=hyperopt_name, kwargs={'config': config})
         if hyperopt:
             return hyperopt
         raise OperationalException(


hroff-1902:
Possible enhancement: 

we could inherit `IStrategy` from `IHyperOpt` and use this technique without any changes to the custom strategies at all. But it's not good from design point of view since 'IStrategy' is not a kind of `IHyperOpt`.

hroff-1902:
Possible enhancement 2:

we could rename `IStrategy` to, say, `IStrategyBase` everywhere in the codebase except default, sample and custom strategies and define `IStrategy` as
```
class IStrategy(IStrategy, IHyperOpt)
    pass
```
-- this would also allow to run hyperopting as proposed above on strategies without any changes to the code of the default, sample and custom strategies.

xmatthias:
> we could inherit IStrategy from IHyperOpt and use this technique without any changes to the custom strategies at all. But it's not good from design point of view since 'IStrategy' is not a kind of IHyperOpt.

long ago (before i started working on freqtrade), hyperopt and strategy were combined into one file/class. 

It did not work well and led to very confusing and very long strategy files, which had 90% of the code only ever used by hyperopt (this code is more verbose by default).
It was also very confusing since for most users it's very unclear what is used for what (what parts do i need to simply run my strategy).

I think hyperopt should remain seperate overall - if loading the strategy works, fine, but i wouldn't touch the strategy-interface over this.


When running the bot in "trading" mode (which is the most important one i think) - i would not want freqtrade to load the modules needed only for hyperopt, and currently we're very carefull to facilitate this, for example by loading `Backtesting` and `Hyperopt` only within `start_hyperopt()` and `start_backtesting()`.

If you start importing IHyperopt into the strategy file, you're throwing this optimization out, which will 
a) slow down startup time (more / heavier imports)
b) require a TON more memory - increasing minimal requirements for the bot

While this may facilitate easier hyperopting to some extend, the most important mode (trading) will be imacted negatively by this - which i don't think should happen.


## An alternative idea:
To facilitate simple Hyperopting, you should instead try to use `StrategyResolver` to import a custom strategy and run / use the strategies `populate_*` methods.
This will obviously require to remove the constraint to only allow DefaultStrategy - but IF you can get that to work that would be best

sample call:
`freqtrade --strategy MyCoolStrategy hyperopt -e 100 --space ROI,stoploss`

This approach would not require **any** change to either strategy interface, nor to the strategy itself - however there is one caveat i ran into previously (which is causing the contraint to `DefaultStrategy`:

when running with anything other than `--j 1` - You get unpickle errors.
I think by manipulating the functions `__dict__` (basically faking a valid import location) this could be fixed - but i gave up on this idea since it cost a ton of time without much benefit - since you needed to write a custom hyperopt file anyway, and copy/pasting `populate_indicators()` was a small hurdle. 

### possible alternative for "quick hyperopt":

Maybe, for the above "special" mode - you could fix this by running `populate_*` methods before invoking parallelism - since you won't change buy/sell signals anyway (only ROI and stoploss change).

This would lead to the following approach:
* quick hyperopt - loads custom strategy and works only for ROI and Stoploss
* real hyperopt - requires custom hyperopt file (will require this anyway - since you cannot know buy/sell space that fits the strategy's indicators) - and is constraint to DefaultStrategy (to avoid the above problem).

I'm currently not clear (not that deep into hyperopt at the moment) if that will require a seperate HyperoptClass, or if it would work as is - however i think this should be feasible.


Sorry this became long ... just my 2 cents ;)

hroff-1902:
The PR was labeled as PoC since it represents a 'demonstration' quick&dirty implementation without refactoring/redesign and allows one to play with hyperopting on strategies right now, to see the benefits (and caveats) of this hyperopting technique. So it's definitely a "proof of concept" one.

Closing it now in favor to the 'alternative idea', see #2235 and the rest as discussed in #2231 .

PR #2234, opened 2019-09-07
datadir should not default to freqtrade/tests/testdata

xmatthias:
## Summary
Freqtrade should not default to test-data from the tests/datadir.

This data is outdated (early 2018, and not a long timerange) - and is never representative of a realistic backtest used today (which must use more recent data!).

part of #2112 - moving tests out of freqtrade module (which will be the next PR).

## Quick changelog

- datadir is always required (it's also always added to the configuration, so there should not be a problem).
- remove `make_testdata_path()` - and replace it with a fixture returning the correct path
- Adapt all tests to this new way of working.


hroff-1902:
LGTM! 👍

PR #944, opened 2018-06-23
Improve strategy handling

gcarq:
## Summary
Implements logic to properly import a given strategy in global freqtrade.strategy scope. The main focus of this PR is to fix the issue mentioned in #930
This work is based on cherry-pick of 8ddd87beaeb5a00ca132987b6b62076eba5f7a54

PR #930, opened 2018-06-19
Replace Hyperopt with scikit-optimize

vertti:
This should finally be ready now.

## What

Replace Hyperopt with Scikit-Optimize

## Why

- better multicore parallel support (using joblib rather than MongoDB or Redis)
- more machine learning algorithms supported
- gets rid of separate `hyperopt_conf.py`, all settings come from `config.json` or Strategy
- persists all evaluations to a file for further analyzing (with loss, results and used parameters)
- gets rid of all installation hassle and several scripts needed to run MongoDB

Thanks @gcarq for solving the strategy pickle problem!

## What's missing

- Stop and resume functionality is missing, it currently does not dump/pickle the `Optimizer`. That should be trivial to do but I'd still like to add it in a different PR.
- Populating indicators should be done in the strategy not in `hyperopt.py`
- It now uses all your CPU cores, this should probably be configurable
- Results should be pretty-printed but they contain numbers that default json dump doesn't handle

PR #2235, opened 2019-09-07
Allow --strategy for hyperopt

hroff-1902:
* Shamanic import_strategy() eliminated;
* The option is allowed for Hyperopt RunMode in cli;

As a result, ski-opt no longer complains about unpickling errors while loading custom strategy instead of default...

A part of implementation/preparation for "An alternative idea" from #2231...


hroff-1902:
@gcarq Could you please take a look at it ? (particularly to first commit)

I found and studied #944 and #930, but still have not guessed what was the intention of that shamanic import_strategy() you introduced an year ago... Some legacy code?

Why it was needed to propagate the strategy into the globals?

#930 is about ski-opt, but #2235 works with the same (well, a new one, current versions, definitely) ski-opt without all this stuff...

hroff-1902:
> Using the following line, MyTestStrat is loaded, but it's NOT using either of its' populate* methods (so it does not matter that it's loaded).

Yes, I know, I left the rest regarding #2231 to a sep. PR in order to not getting this PR overburden. That's why it's named 'allow --strategy option', no more.

hroff-1902:
> Ok, if you don't think it'll create confusion by enabling an option which is basically not doing anything ... ?

It can be considered as a refactoring PR simplifying the code/removing dead code...

PR #2237, opened 2019-09-08
Enhance setup.py to include long_description

xmatthias:
## Summary

Adding `long_description` and `long_description_content_type` to setup.py, so once we publish to pypi it has the correct description. 

part of issue: #2112 

## Quick changelog

- Add long description (copy Readme.md if available).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,14 @@ if version_info.major == 3 and version_info.minor < 6 or \
     print('Your Python interpreter must be 3.6 or greater!')
     exit(1)
 
-from freqtrade import __version__
+from pathlib import Path  # noqa: E402
+from freqtrade import __version__  # noqa: E402
+
+
+readme_file = Path(__file__).parent / "README.md"
+readme_long = "Crypto Trading Bot"
+if readme_file.is_file():
+    readme_long = (Path(__file__).parent / "README.md").read_text()
 
 # Requirements used for submodules
 api = ['flask']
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -36,6 +43,8 @@
 setup(name='freqtrade',
       version=__version__,
       description='Crypto Trading Bot',
+      long_description=readme_long,
+      long_description_content_type="text/markdown",
       url='https://github.com/freqtrade/freqtrade',
       author='gcarq and contributors',
       author_email='michael.egger@tsn.at',
```
hroff-1902: This should probably be updated.

author='Freqtrade Team', relevant common email...

If @gcarq don't mind...

(May be done in a sep. PR)

xmatthias:
> (Since it's impossible to see how this long_description will be rendered in pypi yet)

huh? check out this: https://pypi.org/project/freqtrade/2019.7.dev0/

it's not fully functional (and huge - due to test-data being included) - but you can see how it should look like.

hroff-1902:
> > (Since it's impossible to see how this long_description will be rendered in pypi yet)
> 
> huh? check out this: https://pypi.org/project/freqtrade/2019.7.dev0/
> 
> it's not fully functional (and huge - due to test-data being included) - but you can see how it should look like.

Cool, seems I have missed this. 

LGTM 👍

xmatthias:
it's only visible when looking at the pre-release branch ... ;)

setup.py
gcarq: Sure, go for it.

PR #2238, opened 2019-09-08
[minor, important] Fix random test failures

xmatthias:
## Summary
#2235 caused random test-failures (mainly when running `pytest freqtrade/test/test_freqtrade.py` (without randomization).

This is because DefaultStrategy is imported directly, which causes StrategyResolver to reuse the already imported object (importlib seems to do this implicitly).

Since #2235 fixes hyperopt importing however, there is no need for the artificial import causing all this.

Followup to #2235, fixing random test failures


hroff-1902:
LGTM 👍

PR #2254, opened 2019-09-10
Eliminate asyncio warnings in tests

hroff-1902:
This PR eliminates asyncio and aiohttp DeprecationWarnings while testing with pytest.

I also found the aiohttp-pytest plugin, but seems it's not needed...

I found this recipe, but I actually don't completely understand in all details what I'm doing by introducing this change and how it can influence testing of other functionality in the affected tests. Especially, I do not understand while other tests in those modules, just below and above the changed tests, do not require it...

### DIFF
--- a/tests/rpc/test_rpc.py
+++ b/tests/rpc/test_rpc.py
@@ -26,7 +26,8 @@ def prec_satoshi(a, b) -> float:
 
 
 # Unit tests
-def test_rpc_trade_status(default_conf, ticker, fee, markets, mocker) -> None:
+@pytest.mark.asyncio
+async def test_rpc_trade_status(default_conf, ticker, fee, markets, mocker) -> None:
     mocker.patch('freqtrade.rpc.telegram.Telegram', MagicMock())
     mocker.patch.multiple(
         'freqtrade.exchange.Exchange',
--- a/tests/rpc/test_rpc.py
+++ b/tests/rpc/test_rpc.py
@@ -26,12 +26,11 @@ def prec_satoshi(a, b) -> float:
 
 
 # Unit tests
-@pytest.mark.asyncio
-async def test_rpc_trade_status(default_conf, ticker, fee, markets, mocker) -> None:
+def test_rpc_trade_status(default_conf, ticker, fee, markets, mocker) -> None:
     mocker.patch('freqtrade.rpc.telegram.Telegram', MagicMock())
+    patch_exchange(mocker)
     mocker.patch.multiple(
         'freqtrade.exchange.Exchange',
-        _load_markets=MagicMock(return_value={}),
         get_ticker=ticker,
         get_fee=fee,
         markets=PropertyMock(return_value=markets)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -117,7 +117,7 @@ def patch_freqtradebot(mocker, config) -> None:
     """
     mocker.patch('freqtrade.freqtradebot.RPCManager', MagicMock())
     persistence.init(config['db_url'])
-    patch_exchange(mocker, None)
+    patch_exchange(mocker)
     mocker.patch('freqtrade.freqtradebot.RPCManager._init', MagicMock())
     mocker.patch('freqtrade.freqtradebot.RPCManager.send_msg', MagicMock())
 
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -69,8 +69,7 @@ def test_init(default_conf, mocker, caplog):
     assert log_has('Instance is running with dry_run enabled', caplog)
 
 
-@pytest.mark.asyncio
-async def test_init_ccxt_kwargs(default_conf, mocker, caplog):
+def test_init_ccxt_kwargs(default_conf, mocker, caplog):
     mocker.patch('freqtrade.exchange.Exchange._load_markets', MagicMock(return_value={}))
     caplog.set_level(logging.INFO)
     conf = copy.deepcopy(default_conf)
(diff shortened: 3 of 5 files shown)


tests/rpc/test_rpc.py
```diff
@@ -26,7 +26,8 @@ def prec_satoshi(a, b) -> float:
 
 
 # Unit tests
-def test_rpc_trade_status(default_conf, ticker, fee, markets, mocker) -> None:
+@pytest.mark.asyncio
+async def test_rpc_trade_status(default_conf, ticker, fee, markets, mocker) -> None:
```
xmatthias: This runs the code of the test within a coroutine - which is now how our code is running normally.

While i'm not 100% certain if that would be a problem, there is an easier (safer?) fix than that
for this test, and probably for the other tests, too, by replacing the first few lines with the following:

```python
    mocker.patch.multiple(
        get_ticker=ticker,
        get_fee=fee,
    )
    patch_exchange(mocker)
```

that'll remove the async warning, which is shown from within `_init_ccxt()` (that function is mocked completely).

I don't quite understand what pytest does differently so this happens (while it doesn't during regular operations) - but i think this should be the fix.

This method should be applicable to all tests (apart from the one testing `_init_ccxt()` - but i think that has this covered differently.

hroff-1902:
Fine! 👍  I guess I got the principle, it's definitely better and the former 'recipe' is incorrect.

* The one outstanding test with warnings is `test_init_ccxt_kwargs` in `test_exchange.py`. It's too hard to my knowledge of mocking...
* Please also review the `test_name` in `test_exchange.py`, am I correct there?

Some other points:
* I did not get why in some tests patch_exchange is called as `patch_exchange(mocker)` while in others it's calles as `patch_exchange(mocker, default_config)`
* To my opinion, fetching patched exchange is missing in some tests, so they use the instance of Exchange that try to connect to exchange. Am I right? This should be prohibited for tests and the tests should be protected against it.

xmatthias:
> I did not get why in some tests patch_exchange is called as patch_exchange(mocker) while in others it's calles as patch_exchange(mocker, default_config)

there's 2 methods here - `patch_exchange()` and `get_patched_exchange()`.
the first simply patches some default methods, while the 2nd returns a patched object (and therefore requires default_config).

You can look at the functions in `conftest.py` - they even call each other.

> `test_init_ccxt_kwargs`

 i can have a look at that

> To my opinion, fetching patched exchange is missing in some tests, so they use the instance of Exchange that try to connect to exchange. Am I right? This should be prohibited for tests and the tests should be protected against it.

yes (and no) ...
we should patch the exchange using `patch_exchange()` in all tests that create an exchange object UNLESS it's testing the ccxt init process (which means we cannot use automock to apply this to all tests... :disappointed:  ).
however, simply creating a ccxt object, and not calling anything in it (no `load_markts()` for example) does not create a call against the exchange, so it's not such a big problem.

I'm also testing without network from time to time (just for fun) to verify how good our mocking is, or if we unintentionally touch something - i've not done that  in a few months now, but i may test that when reviewing this PR.
It's not necessarily showing up as errors - can also be by "waiting" for some timeout ...

hroff-1902:
> there's 2 methods here - `patch_exchange()` and `get_patched_exchange()`.
> the first simply patches some default methods, while the 2nd returns a patched object (and therefore requires default_config).
> 
> You can look at the functions in `conftest.py` - they even call each other.

Yes, I know of those 2 functions, but missed that they both are used in these cases, missed it, thanks.

> > `test_init_ccxt_kwargs`
> 
> i can have a look at that

Yes, please.

> yes (and no) ...
> we should patch the exchange using `patch_exchange()` in all tests that create an exchange object UNLESS it's testing the ccxt init process (which means we cannot use automock to apply this to all tests... 😞 ).
> however, simply creating a ccxt object, and not calling anything in it (no `load_markts()` for example) does not create a call against the exchange, so it's not such a big problem.
> 
> I'm also testing without network from time to time (just for fun) to verify how good our mocking is, or if we unintentionally touch something - i've not done that in a few months now, but i may test that when reviewing this PR.
> It's not necessarily showing up as errors - can also be by "waiting" for some timeout ...

Okay, but can we use 'exchange' as a parameter (filled with patched exchange) to those tests (i.e. to almost all, besides those where ccxt init is tested) where the patched exchange is needed? This would remove all those calls to `patch_exchange` and `get_patched_exchange` in many of the tests? (Not for this PR, definitely)

xmatthias:
> Okay, but can we use 'exchange' as a parameter (filled with patched exchange) to those tests (i.e. to almost all, besides those where ccxt init is tested) where the patched exchange is needed? This would remove all those calls to patch_exchange and get_patched_exchange in many of the tests? (Not for this PR, definitely)

the problem with this is that this will limit patching the ccxt object to "default" patches.
Currently many tests patch specific functions (`api_mock.fetch_orders=MagicMock(return_value=...)` - which is needed for many tests to work reliably.

Now we should differentiate between tests testing the exchange (`test_exchange.py`) and the rest, since for other tests we may be able to patch the exchange-object instead of the ccxt object, or standardize responses so all tests use the same ccxt responses - but that's a huge task since many things will require different return values.

xmatthias:
You can fix `test_init_ccxt_kwargs()` by adding the following mock statement to the top of that function.

``` python
mocker.patch('freqtrade.exchange.exchange.ccxt_async.Exchange.open', MagicMock())
```

While technically this works and removes the asyncio warning, it could also break should ccxt rename / change this method, so i'm not sure if we should do it this way.

I've also opened an issue in ccxt (https://github.com/ccxt/ccxt/issues/5793) because of this, since i discovered that this warning DOES raise within real runs (it's just not showing ...), so we may have to change how we init the asyncio version of the exchange object.

hroff-1902:
let's leave this test alone as it is till that issue in the ccxt is ready and merge other fixes for the rest of warnings...

PR #2255, opened 2019-09-11
hyperopt loss function based on calmar ratio - not merge - WIP - POC

gaugau3000:
unit tests, and is your code PEP8 conformant? 

unit test -> no i will but think i need to use the fixtures from
https://github.com/freqtrade/freqtrade/blob/f987e6e0f9bdcae6f0cec3a273e270a3a35a3c12/tests/optimize/test_hyperopt.py#L33 but the last trade in the dataframe is SellType.STOP_LOSS and profitperc positive ? i missed something for sure. 

PE8 conformant -> yes

## Summary

Provide to freqrade hyperopt a loss function based on calmar ratio (max drawdown)

## Quick changelog

- add an extra hyperopt loss function

## What's new?
- Based on the simulation of the median max drawdown (montecarlo simulation) calculate the calmar ratio. (https://www.seeitmarket.com/3-of-many-uses-for-monte-carlo-simulations-in-trading-18266/ kind of the second part based on randomization)

## Questions
1/ Don't know the road map and the fact that you don't want to include a lot of dependencies is there a way to include a risk library (e.g https://github.com/quantopian/empyrical) -> it can be useful for loss function and backtest (in my case i create a max_drawdown function but a library could be better) ?
2/ i see that all params for loss are hardcoded (eg :  MAX_ACCEPTED_TRADE_DURATION = 300
) but it couls be usefull to be able to overwrite defaults in the hyperopt class ?

This pull request is for now only to have a feedback (feel free to tell me this is not in the road map or useless :-) ) (if you think it useful i spend some times create tests, update doc...)

Have a nice day and many tks for this beautiful project.

### DIFF
--- a/freqtrade/optimize/hyperopt_loss_calmar.py
+++ b/freqtrade/optimize/hyperopt_loss_calmar.py
@@ -59,8 +59,8 @@ class CalmarHyperOptLoss(IHyperOptLoss):
         abs_mediam_simulated_drawdowns = Series(simulated_drawdowns).median()
         calmar_ratio = return_avg_per_year/abs_mediam_simulated_drawdowns
 
-        # Normalize loss value to be float between (0, 1)
-        calmar_loss = 1 - (norm.cdf(calmar_ratio, 0, 100))
+        # Normalize loss value to be float between (0, 1) :  0.5 value mean no profit
+        calmar_loss = 1 - (norm.cdf(calmar_ratio, 0, 10))
 
         # feel free to add other criterias (e.g avg expected time duration)
         loss = (calmar_loss * CALMAR_LOSS_WEIGHT)
--- a/freqtrade/optimize/hyperopt_loss_calmar.py
+++ b/freqtrade/optimize/hyperopt_loss_calmar.py
@@ -70,12 +70,15 @@ class CalmarHyperOptLoss(IHyperOptLoss):
         0 mean no profit
         """
         calmar_loss = 0.5 - (norm.cdf(calmar_ratio, 0, 1.5/CALMAR_LOSS_WEIGHT))
+        normalized_expected_trade_loss = norm.cdf(trade_count-NB_EXPECTED_TRADES,
+                                                  -NB_EXPECTED_TRADES,
+                                                  (NB_EXPECTED_TRADES*(2/3))*EXPECTED_TRADES_WEIGHT)
 
         """
         Normalize loss value to be float between (0, 0.5) :
         Closed to 0 mean trade_count = NB_EXPECTED_TRADES
         """
-        expected_trade_loss_penalty = 1 - norm.cdf(trade_count-NB_EXPECTED_TRADES,-NB_EXPECTED_TRADES,(NB_EXPECTED_TRADES*(2/3))*EXPECTED_TRADES_WEIGHT)
+        expected_trade_loss_penalty = 1 - normalized_expected_trade_loss
 
         # want to see the loss -> shorturl.at/DHX34
         loss = calmar_loss + expected_trade_loss_penalty
--- a/freqtrade/optimize/hyperopt_loss_calmar.py
+++ b/freqtrade/optimize/hyperopt_loss_calmar.py
@@ -7,10 +7,11 @@ from freqtrade.optimize.hyperopt import IHyperOptLoss, MAX_LOSS
 
 NB_SIMULATIONS = 1000
 SIMULATION_YEAR_DURATION = 3
-CALMAR_LOSS_WEIGHT = 1
 
 SLIPPAGE_PERCENT = 0.000
 NB_EXPECTED_TRADES = 600
+
+CALMAR_LOSS_WEIGHT = 0.5
 EXPECTED_TRADES_WEIGHT = 0.5
 
 
@@ -64,8 +65,12 @@ class CalmarHyperOptLoss(IHyperOptLoss):
 
         calmar_ratio = mediam_simulated_annualized_returns/abs_mediam_simulated_drawdowns
 
-        # Normalize loss value to be float between (0, 1) :  0.5 value mean no profit
-        calmar_loss = 1 - (norm.cdf(calmar_ratio, 0, 10))
+        """
+        Normalize loss value to be float between (-0.5, 0.5)
+        0 mean no profit
+        """
+        calmar_loss = 0.5 - (norm.cdf(calmar_ratio, 0, 1.5/CALMAR_LOSS_WEIGHT))
+
         """
         Normalize loss value to be float between (0, 0.5) :
         Closed to 0 mean trade_count = NB_EXPECTED_TRADES
--- a/freqtrade/optimize/hyperopt_loss_calmar.py
+++ b/freqtrade/optimize/hyperopt_loss_calmar.py
@@ -17,7 +17,7 @@ EXPECTED_TRADES_WEIGHT = 0.5
 class CalmarHyperOptLoss(IHyperOptLoss):
     """
     Defines the calmar loss function for hyperopt (you maybe need to add other criterias
-    like the number of trades expected and the max duration expected).
+    as the max duration expected).
     Calmar ratio is based on  average annual rate of return for the last 36 months divided by the
     maximum drawdown for the last 36 months.
     But you maybe don't have running hyperopt with 36 months of data so we will simulate 36 months
--- a/freqtrade/optimize/hyperopt_loss_calmar.py
+++ b/freqtrade/optimize/hyperopt_loss_calmar.py
@@ -55,7 +55,7 @@ class CalmarHyperOptLoss(IHyperOptLoss):
         abs_mediam_simulated_drawdowns = Series(simulated_drawdowns).median()
         calmar_ratio = return_avg_simulation_duration/abs_mediam_simulated_drawdowns
 
-        # float between ]0,1[
+        # Normalize loss value to be float between (0, 1)
         calmar_loss = 1 - (norm.cdf(calmar_ratio, 0, 100))
 
         # feel free to add other criterias (e.g avg expected time duration)
--- a/freqtrade/optimize/hyperopt_loss_calmar.py
+++ b/freqtrade/optimize/hyperopt_loss_calmar.py
@@ -1,12 +1,12 @@
 from datetime import datetime
 from pandas import DataFrame, Series
-from freqtrade.optimize.hyperopt import IHyperOptLoss
 import numpy as np
 from scipy.stats import norm
 
+from freqtrade.optimize.hyperopt import IHyperOptLoss, MAX_LOSS
+
 NB_SIMULATIONS = 1000
 SIMULATION_YEAR_DURATION = 3
-HIGH_NUMBER = 100000
 CALMAR_LOSS_WEIGHT = 1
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
...
+            randomized_result = results.profit_percent.sample(n=sample_size,
+                                                              random_state=np.random.RandomState(),
+                                                              replace=True)
+            simulated_drawdown = cls.abs_max_drawdown(randomized_result)
+            simulated_drawdowns.append(simulated_drawdown)
+
+        abs_mediam_simulated_drawdowns = Series(simulated_drawdowns).median()
+        calmar_ratio = return_avg_simulation_duration/abs_mediam_simulated_drawdowns
+
+        # float between ]0,1[
```
hroff-1902: Thanks for this reverse bracket notation for the open interval, I have not seen it for long time, it made me smile (in good sense)... 😃 

Should probably better be `# Normalize loss value to be float between (0, 1)` or something like...

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
+from datetime import datetime
+from pandas import DataFrame, Series
+from freqtrade.optimize.hyperopt import IHyperOptLoss
+import numpy as np
+from scipy.stats import norm
+
+NB_SIMULATIONS = 1000
+SIMULATION_YEAR_DURATION = 3
+HIGH_NUMBER = 100000
```
hroff-1902: `MAX_LOSS = 10000` is used in `hyperopt.py` for this purpose, this should probably be placed in the constants or imported from `freqtrade.optimize.hyperopt` directly.

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
...
+CALMAR_LOSS_WEIGHT = 1
+SLIPPAGE_PERCENT = 0.001
+
+
+class CalmarHyperOptLoss(IHyperOptLoss):
+    """
+    Defines the default loss function for hyperopt
```
hroff-1902: No, it's not the default loss now...

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
...
+                               min_date: datetime, max_date: datetime,
+                               *args, **kwargs) -> float:
+
+        """
+        Objective function, returns smaller number for better results
+        """
+
+        simulated_drawdowns = []
+
+        backtest_duration_years = ((max_date-min_date).days/365.2425)
```
hroff-1902: Are not you too precise here? Implementation of the Sharpe loss, for example, uses 365 for annualization...

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
...
+        trade_count_average_per_year = trade_count/backtest_duration_years
+
+        # add slipage to be closed to live
+        results['profit_percent'] -= SLIPPAGE_PERCENT
```
hroff-1902: I think you should make here a copy of the dataframe passed, since it not only changes the values in the manner needed for later calculations, needed for internals of this function, but also changes this column in the original dataframe, which will be also used in upper levels of hyperopt (printed as results, etc.). 

If I'm not wrong...

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Objective function, returns smaller number for better results
+        """
+
+        simulated_drawdowns = []
+
+        backtest_duration_years = ((max_date-min_date).days/365.2425)
+        trade_count_average_per_year = trade_count/backtest_duration_years
+
+        # add slipage to be closed to live
```
hroff-1902: "slippage"

Actually, similar approach (introducing some percent of slippage) should be implemented as a generic option for backtesting and hyperopt, but let's leave it here hardcoded for a while, a generic solution will follow...

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
...
+        results['profit_percent'] -= SLIPPAGE_PERCENT
+
+        return_avg_per_year = (results.profit_percent.sum() / backtest_duration_years)
+        return_avg_simulation_duration = return_avg_per_year * SIMULATION_YEAR_DURATION
+
+        sample_size = round(trade_count_average_per_year * SIMULATION_YEAR_DURATION)
+
+        # exclude the case when no trade was lost
+        if(results.profit_percent.min() >= 0):
```
hroff-1902: `if results.profit_percent.min() >= 0:`

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,76 @@
...
+            return HIGH_NUMBER
+
+        # simulate n years of run to define a median max drawdown
+        for i in range(0, NB_SIMULATIONS):
+            randomized_result = results.profit_percent.sample(n=sample_size,
+                                                              random_state=np.random.RandomState(),
+                                                              replace=True)
```
hroff-1902: Interesting approach. Need to think about this...

freqtrade/optimize/hyperopt_loss_calmar.py
xmatthias: `format_results()` is called before this - after `calculate_loss()` the results dataframe is not reused anymore at the moment, so i think this is fine for now (unless we change the order within `generate_optimizer()` ... which i don't think is planned).

freqtrade/optimize/hyperopt_loss_calmar.py
xmatthias: wouldn't you want to check that lost_trades is > NB_SIMULATIONS ?

worst case, you may end up taking 3 samples out of a 3 sample bucket ... ?

xmatthias:
Another option will for sure be useful! :+1: 

For tests - best look at `test_sharpe_loss_prefers_higher_profits()` `test_onlyprofit_loss_prefers_higher_profits()`or `test_loss_calculation_has_limited_profit()`  in `test_hyperopt.py`.
these tests test the existing loss-functions (honestly, they're basic tests testing only that the return-value moves in the right direction with better / worse results).

freqtrade/optimize/hyperopt_loss_calmar.py
gaugau3000: I have indroduce a NB_EXPECTED_TRADES constant to give a penalty  to prevent the convergence     to 1 trade lost n trade win (shorturl.at/DHX34). Feel free to close it if it is ok for you.

gaugau3000:
> > but i've not thought deeply about this yet.
> 
> I have (already, have a WiP stash about this)... 😉
> 
> I think I have a probably better idea -- pass parameters as json-string:
> `hyperopt --hyperopt-loss-params="{'MAX_DRAWDOWN'=20, 'NUM_TRADES'=2}"`
> 
> Same for other parameters, for example for parameter passed to skopt optimizer on its initialization -- they will be loaded from json passed in the `--hyperopt-skopt-params` option and passed in kwargs to skopt optimizer...
> 
> Not here, of course. in a sep. PR

I agree (and cool for automation) in the other hands the length of the cli is quite heavy (and for example i use docker so with binding) but lets discuss that in another thread

gaugau3000:
Hi, i will play with 1-2 weeks to take step back then come back to you if the result is ok. Have a nice day.

xmatthias:
that's what draft pr's are for - it'll allow us (and others) to play with this as well... more eyes, more opinions, hopefully better results :+1:

gaugau3000:
if #2270 is validated don't create a hyperopt loss calmar as it can't work alone (need minimum a weight with expected trade number) and prefer package it to be used with other rules.

gaugau3000:
Hi there if it is ok for you **i close and we don't merge.** and related to exchange in slack with @hroff-1902 


> provide a custom calmar ratio hyperopt loss out the box don't have a lot of sense since it have to be pondered by an expected number of trade to work (difficult to test so and don't have a sole responsibility so not very good for end user). Take all the time you need no stress.

Have a nice day :-)

gaugau3000:
@hroff-1902 if you would like to use calmar ratio based on Monte-Carlo analysis prefer use [this repo](https://github.com/gaugau3000/mc_sim_fin) then 

`calmar_ratio=mc_sims_results['med_profit_percent']/mc_sims_results['med_max_drawdown_percent']`

This pull request is way to far from freqtrade scope.

Close it when you want.

xmatthias:
> This pull request is way to far from freqtrade scope.
> Close it when you want.

Do i understand this correctly that there is no interrest to drive this further, so we shall close this ... ? 
or would it still be interresting ... ?

hroff-1902:
I explain my silence here: it's okay to add loss function based on Calmar metrics in the same way as Sharpe and Sortino added (it differs in one or couple of expressions), but I do not think this completely on the Monte Carlo approach integrated into this. I tried to learn some about this approach but still need more knowledge on this.

hroff-1902:
> I explain my silence here: it's okay to add loss function based on Calmar metrics in the same way as Sharpe and Sortino added (it differs in one or couple of expressions), but I do not think this completely on the Monte Carlo approach integrated into this. I tried to learn some about this approach but still need more knowledge on this.

My proposal: let's add Calmar loss-function in the same way it was made for Sharpe and Sortino. Monte Carlo needs to be evaluated further...

hroff-1902:
(@gaugau3000 and I'm very sorry for this PR hanging for long time, I tried several times but found I still need to know more theory behind thia)

gaugau3000:
Hi sorry to give a response that late...

> Do i understand this correctly that there is no interrest to drive this further, so we shall close this ... ?
> or would it still be interresting ... ?

Use profit/drawdown as a loss function is good (especially using monte Carlo simulation) but implement the montecarlo simulation in freqtrade in not good as it is very specific and outside the scope. It is why i have created a [github repository](https://github.com/gaugau3000/mc_sim_fin).

The book "Building Winning Algorithmic Trading Systems" speak about that at the page 123 (Fitness Function -> Return on Account) if you have a chance to take a look.

So there is two choices :
- close this issue because it mean add a new package dependency in freqtrade and it is not good.
or
- merge this branch with a code that will be look like below (if we choose this option i have a minor change to do in my repository code) :
```
from mc_sim_fin.mc import mc_analysis

def hyperopt_loss_function(cls, results: DataFrame, trade_count: int,
                               min_date: datetime, max_date: datetime,
                               *args, **kwargs) -> float:

        results['profit_amount'] = results.profit_percent * trade_amount

        mc_sims_results = mc_analysis(results[['open_time', 'profit_amount']])

        mc_sims_results['calmar'] = mc_sims_results['med_profit_percent'] / mc_sims_results['med_max_drawdown_percent']

        return - mc_sims_results['calmar']
```

No matter.

Have a nice day.

xmatthias:
Based on the above comments which propose to close this PR, i'm going to follow through with this and close this PR.

Issue #2078, opened 2019-08-01
Allow simple loading of "dummy" configuration


xmatthias:
A simple method to load a minimal configuration should be added.

This can be handy for some data-analysis tasks in notebooks, where no full configuration is needed or necessary.

Something like `Configuration.get_empty_config()`, which contains the minimum required keys set to default values

hroff-1902:
Something like below?

https://github.com/freqtrade/freqtrade/blob/4c005e70861ec7e2294f6097420e0ece289af2d1/scripts/download_backtest_data.py#L61-L74

hroff-1902:
Another idea -- freqtrade should work without config at all. With a configuration, where all the settings are filled with their default values...

xmatthias:
yes and no ... 
regular running won't work without a config ... however certain operatinos should.
my current case:

The below works just fine - but it'd be easier to use `Configuration.get_empty()`
and then replace the 2 attributes that matter for this (user_data_dir does not for this particular operation ...).
```
strategy = StrategyResolver({'strategy': strategyname,
                            'user_data_dir': Path.cwd(),
                            'strategy_path': location}).strategy
```

hroff-1902:
> yes and no ...
> regular running won't work without a config ... however certain operatinos should.
> my current case:

I mean dry-run,, indeed. It should.

hroff-1902:
> strategy = StrategyResolver({'strategy': strategyname,
>                             'user_data_dir': Path.cwd(),
>                             'strategy_path': location}).strategy
> ```

Why do you need both user_data and strategy_path passed here? After loading of configuration (with args), the strategy_path should already be resolved and filled with proper path (taken from the `--strategy-path` option or `strategy_path` setting, if used, or with <userdir>/strategies otherwise... After a call to `Configuration.load_config()` the strategy_path should have its desired value according to the options and config settings. We could write the `load_config(options: str)` helper function (to be used interactively by users in a python interactive interpreter session, jupyter or whatever you call it) that accept the string with options, splits it as shell does into options and calls the `Configuration.load_config()`. If no options passes, no config specified and no default config found, all settings should be filled with their default values, that should be appropriate to run the rest. Including the strategy_path above -- it should be filled with `<pwd>/user_data/strategies`...

xmatthias:
This is a snippet from #2080, which is documentation on how to use certain parts independently.

[...]

PR #2260, opened 2019-09-13
Configuration/Arguments refactoing (don't pass Namespace around).

xmatthias:
## Summary
Simplify configuration (passing a dict instead of `Namespace` to it).
Where possible, we can use `args.get('argument')` instead of having to check everything first.

This will allow simpler (and correct) initialization of configurations in interactive environments (still using `Configuration.from_files([files...])`).

This should also solve #2078, since a simple (sample) configuration can be passed to `from_files()` and will be correctly initialized (including the correct types for data_dir and user_data_dir) - as pointed out in #2228 .

## Quick changelog

- Use vars(args) to avoid passing around Namespace objects
- Refactor `from_files()` to correctly initialize configuration 
- Add create-datadir to the commands that don't require config to be set.

### DIFF
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -91,7 +91,8 @@ df.groupby("pair")["sell_reason"].value_counts()
 
 ### Load multiple configuration files
 
-This option can be useful to inspect the results of passing in multiple configs
+This option can be useful to inspect the results of passing in multiple configs.
+This will also run through the whole Configuration initialization, so the configuration is completely initialized to be passed to other methods.
 
 ``` python
 import json
@@ -101,7 +102,16 @@ from freqtrade.configuration import Configuration
 config = Configuration.from_files(["config1.json", "config2.json"])
 
 # Show the config in memory
-print(json.dumps(config, indent=1))
+print(json.dumps(config, indent=2))
+```
+
+For Interactive environments, have an additional configuration specifying `user_data_dir` and pass this in last, so you don't have to change directories while running the bot.
+Best avoid relative paths, since this starts at the storage location of the jupyter notebook, unless the folder is changed.
+
+``` json
+{
+    "user_data_dir": "~/.freqtrade/"
+}
 ```
 
 ### Load exchange data to a pandas dataframe
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -106,7 +106,7 @@ print(json.dumps(config, indent=2))
 ```
 
 For Interactive environments, have an additional configuration specifying `user_data_dir` and pass this in last, so you don't have to change directories while running the bot.
-Best avoid relative paths, since this starts at the storage location of the jupyter notebook, unless the folder is changed.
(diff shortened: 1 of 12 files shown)


docs/data-analysis.md
```diff
@@ -101,7 +102,16 @@ from freqtrade.configuration import Configuration
 config = Configuration.from_files(["config1.json", "config2.json"])
 
 # Show the config in memory
-print(json.dumps(config, indent=1))
+print(json.dumps(config, indent=2))
+```
+
+For Interactive environments, have an additional configuration specifying `user_data_dir` and pass this in last, so you don't have to change directories while running the bot.
+Best avoid relative paths, since this starts at the storage location of the jupyter notebook, unless the folder is changed.
```
hroff-1902: No folders, please! 
🤣

hroff-1902:
LGTM, besides a comment above on the use of the term for directories, and a suggestion, a wish below:

if it's possible to run quickly an auto replacement on the changes made here, let's move all `args["argname"]` to `args['argname']`. Although it's not in PEPs and in the Zen of python, this could later help during hypothetical internationalization, resolving I18N issues. I mean, keeping as a rule, that the 'translatable' messages printed to the humans are in the double quotes, while intrinsic string constants, dict keys, options, settings are written in the single quotes. I do not insist on this (but believe me, it can help during I18N processes), and this is only if it can be done with little effort...

xmatthias:
> if it's possible to run quickly an auto replacement on the changes made here, let's move all args["argname"] to args['argname']. Although it's not in PEPs and in the Zen of python, this could later help during hypothetical internationalization, resolving I18N issues. I mean, keeping as a rule, that the 'translatable' messages printed to the humans are in the double quotes, while intrinsic string constants, dict keys, options, settings are written in the single quotes. I do not insist on this (but believe me, it can help during I18N processes), and this is only if it can be done with little effort...

While certainly possible, i am not sure how that could help with internationalization - I guess you're thinking about auto-replacing stuff (?) - however you'll need to evaluate that for every string anyway, to make sure you're not replacing something wrongly.
Never done I18N with python, but i assume you'll have a dict (or other container) for each string to print and you select from that (probably thorugh a library...).

However, I don't see a point in internationalization for freqtrade - it's about trading, not about accessibility - getting a log-message in another language does not give anyone any more profits, so i think that'll be wasted time (we can think about this again should we have an UI ...). - and immagine the problems when we get issues with logs in languages noone of us understands ... 

Regarding formatting of strings - I'm more drawn towards [black](https://black.readthedocs.io/en/stable/) - which formats the python code in a certain way (strings are in doublequotes unless that requires escaping doublequotes within the string, line-breaks are standardized, ...).

We're a long way from that (running black modifies every file in the project ...) - however what i like about it is that it'll terminate discussions around formatting - since black is THE standardized format.

hroff-1902:
Citing from black manual on strings:

> (like the popular “single quotes for data, double quotes for human-readable strings”)

😉 

Okay, black'ish style is also okay. Forget that above.

PR #2261, opened 2019-09-13
[minor] Test speedup

xmatthias:
## Summary
Remove indicators from default-strategy that are not used.
Since we now use SampleStrategy as example for users, the default-strategy should contain only indicators that are really needed by the strategy and by some additional tests (bollinger-bands).

This did speed up tests from 43 to ~30 seconds for me, since the default-strategy is analyzed multiple times.

## Quick changelog

- Remove unused indicators

## questions

- [x] should we remove all the "clutter" within this file (commented indicators and explanations) and only leave what's really necessary?
The comments and additional indicators were intended as a sample for users, however the SampleStrategy should be used for that and is linked throughout the documentation.

### DIFF
--- a/freqtrade/strategy/default_strategy.py
+++ b/freqtrade/strategy/default_strategy.py
@@ -4,7 +4,6 @@ import talib.abstract as ta
 from pandas import DataFrame
 
 import freqtrade.vendor.qtpylib.indicators as qtpylib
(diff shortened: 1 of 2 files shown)


hroff-1902:
Agree completely.
👍

PR #2274, opened 2019-09-18
Manual bump to ccxt 1.18.1159

hroff-1902:
(support for binance.us)

### DIFF
--- a/requirements-common.txt
+++ b/requirements-common.txt
@@ -1,6 +1,6 @@
 # requirements without requirements installable via conda
 # mainly used for Raspberry pi installs
-ccxt==1.18.1149
+ccxt==1.18.1159
 SQLAlchemy==1.3.8
 python-telegram-bot==12.1.0
 arrow==0.15.2


xmatthias:
it's just a nice to have / allow our users to use binance.us (couldn't find that quickly looking through ccxt)? 

or is something broken without this (in which case i would also increment the min version in `setup.py`).

hroff-1902:
> it's just a nice to have / allow our users to use binance.us (couldn't find that quickly looking through ccxt)?
> 
> or is something broken without this (in which case i would also increment the min version in `setup.py`).

nothing broken, just to allow usage of new exchange, asked at Slack

hroff-1902:
ah, `setup.py`, I always forgetting about this...
But was not aligned with the version stated here, in `requirements-common.txt`...
When should we change `setup.py`?

hroff-1902:
Thanks. Right, we have a separate class for binance...

PR #2275, opened 2019-09-18
minor: Cleanup in backtesting

hroff-1902:
* Get rid of `Backtesting.advise_buy`, `Backtesting.advise_sell` -- they are not actually needed and only increase the complexity of the assignments of the attributes between classes. Use `Backtesting.strategy.advise_*` instead.
* Cosmetics for the log messages in the configuration
* Initialization for `hyperopt.backtesting` moved to be performed after hyperopt class is resolved. This will help later fetch stoploss and roi from the hyperopt if/when this will be implemented.
* Tests adjusted.

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -49,9 +49,10 @@ class Hyperopt:
     """
     def __init__(self, config: Dict[str, Any]) -> None:
         self.config = config
-        self.backtesting = Backtesting(self.config)
 
         self.custom_hyperopt = HyperOptResolver(self.config).hyperopt
+
+        self.backtesting = Backtesting(self.config)
 
         self.custom_hyperoptloss = HyperOptLossResolver(self.config).hyperoptloss
         self.calculate_loss = self.custom_hyperoptloss.hyperopt_loss_function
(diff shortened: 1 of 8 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -259,13 +262,16 @@ def generate_optimizer(self, _params: Dict) -> Dict:
         """
         params = self.get_args(_params)
         if self.has_space('roi'):
-            self.backtesting.strategy.minimal_roi = self.custom_hyperopt.generate_roi_table(params)
+            self.backtesting.strategy.minimal_roi = \
+                    self.custom_hyperopt.generate_roi_table(params)
```
xmatthias: is this new linewrap cosmetic (because it looks better) or for flake8?

just wondering since i don't see anything else changed and i wonder if i missed something...

freqtrade/resolvers/hyperopt_resolver.py
```diff
@@ -38,11 +38,11 @@ def __init__(self, config: Dict) -> None:
         IHyperOpt.ticker_interval = str(config['ticker_interval'])
 
         if not hasattr(self.hyperopt, 'populate_buy_trend'):
-            logger.warning("Custom Hyperopt does not provide populate_buy_trend. "
-                           "Using populate_buy_trend from DefaultStrategy.")
+            logger.warning("Hyperopt class does not provide populate_buy_trend() method. "
```
xmatthias: Should we keep the "Custom" here to make it clear it's not `hyperopt.py` that causes this, but the class the user provided? 

Hyperopt is 3 things really - a Class (`Hyperopt` / `hyperopt.py`), a process (hyperopting) and a "Custom Hyperopt" (happy to hear better names)
I think this is confusing and we should change naming for this in general (oddly enough it's clear for the Strategy). Not for this PR - but was noticed here...

freqtrade/optimize/hyperopt.py
hroff-1902: this one is to align to couple of next lines, that do not fit into the limit

freqtrade/resolvers/hyperopt_resolver.py
hroff-1902: actually, I wanted to ask you for removal 'custom' from the name of the cli option...
not for this pr, though.

Here the 'Custom' is not correct since it can be DefaultHyperopt, not Custom.

xmatthias: well noone should use defaultHyperopt anyway, so we maybe should enforce the --customhyperopt anyway ...

hroff-1902: > well noone should use defaultHyperopt anyway, so we maybe should enforce the --customhyperopt anyway ...

Badabum! Really? I thought I need a special long speech to convince you in the same... 😆

Issue #36, opened 2017-09-28
Convert configuration into a class


vertti:
Use https://github.com/cwacek/python-jsonschema-objects to generate a Python class from the configuration JSON schema. Make configuration use that and remove unnecessary float casts etc after that.

gcarq:
I'm in favor of a minimal solution explained here: https://stackoverflow.com/a/23689767

shusso:
Also this is pretty simple
```
creator = lambda name: type(name.capitalize(), (object,), {})
c = creator('conf')
c.key = 'xxx'
c.dkey = {'key':'value'}

c.__dict__
mappingproxy({'__dict__': <attribute '__dict__' of 'Conf' objects>,
              '__doc__': None,
              '__module__': '__main__',
              '__weakref__': <attribute '__weakref__' of 'Conf' objects>,
              'd': {'a': 1},
              'key': 'asdf'})
```

jraviotta:
FYI:
The recent changes to config parsing from files seems to have broken display of config json.
```
config_files = [
    Path('config.json'),
    Path('exchange-config.json')
]
# Create config dict
config = Configuration.from_files(config_files)
print(config)
print(json.dumps(config, indent=1))
```

```
2019-09-19 20:20:45,753 - freqtrade.configuration.configuration - INFO - Using config: config.json ...
2019-09-19 20:20:45,786 - freqtrade.configuration.configuration - INFO - Using config: exchange-config.json ...
2019-09-19 20:20:45,792 - freqtrade.configuration.configuration - INFO - Validating configuration ...
2019-09-19 20:20:45,812 - freqtrade.loggers - INFO - Verbosity set to 0
2019-09-19 20:20:45,815 - freqtrade.configuration.configuration - INFO - Dry run is enabled
2019-09-19 20:20:45,818 - freqtrade.configuration.configuration - INFO - Using DB: "sqlite://"
2019-09-19 20:20:45,824 - freqtrade.configuration.configuration - INFO - Using max_open_trades: 3 ...
2019-09-19 20:20:45,836 - freqtrade.configuration.configuration - INFO - Using user-data directory: /Users/jonathanraviotta/.freqtrade ...
2019-09-19 20:20:45,839 - freqtrade.configuration.configuration - INFO - Using data directory: /Users/jonathanraviotta/.freqtrade/data/binance ...
2019-09-19 20:20:45,844 - freqtrade.configuration.check_exchange - INFO - Checking exchange...
[...]

PR #2276, opened 2019-09-20
Allow easy printing of loaded configuration

xmatthias:
## Summary

Allow easy printing of loaded configuration
(beforechanging types and applying defaults)

Solve the issue: https://github.com/freqtrade/freqtrade/issues/36#issuecomment-533356803

## Quick changelog

- keep originally loaded configuration as part of the configuration, so the combined configuration can be plotted easily.
- will enable the configurator mentioned in #2112 - which will require saving the configuration to disk again.

### DIFF
--- a/docs/data-analysis.md
+++ b/docs/data-analysis.md
@@ -102,7 +102,7 @@ from freqtrade.configuration import Configuration
 config = Configuration.from_files(["config1.json", "config2.json"])
 
 # Show the config in memory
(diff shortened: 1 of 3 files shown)


xmatthias:
thanks - sorry i missed that today :fearful:

hroff-1902:
> thanks - sorry i missed that today 😨

nevermind

Issue #2281, opened 2019-09-21
No DefaultStrategy and DefaultHyperopt


hroff-1902:
It's good time to make this change in terms of #2264 / quickly at the same time since it changes user cli interface:

* No DefaultStrategy. If a user wants to load a use some sample strategy, she needs to specify it explicitely: `-s SampleStrategy`.

Technically, if hyperopt needs a strategy class as a placeholder for methods for backtesting, it should create an instance of IHyperopt in this case. This should probably be the behavior of resolver.

* No DefaultHyperopt -- specify SampleHyperopt explicitely.

Same probably for other resolvables.

xmatthias:
is a simple "yes" enough discussion? :laughing: 

I completely agree - but i'm not sure if i should include it into #2264, or if we should do these seperately.

I'm think we can change `--strategy` in #2264 without much fuss - but loading the Hyperopt change should probably be done seperately (honestly, because i would want to leave that to you ...)

hroff-1902:
Not probably included into #2264, but merged together when implemented...

hroff-1902:
What about DefaultLoss (legacy) and Default pairlist (static)?

xmatthias:
i would leave these to be honest - 
Having sensible defaults does make sense in some areas (defaultloss / pairlist) - but not in others which we assume users have to change to be profitable (strategy, hyperopt).


side note: to merge them together, we can do Pull requests against the branch `args_aftersubcommand2` (based on that branch too, obviously)... this way we can review the changes we're doing seperately, without the "huge" change...

hroff-1902:
Right, or we could create a separate feature branch, create and merge prs against it (including args_aftersubcommand2) and finally merge it into develop...

xmatthias:
args_aftersubcommand2 is technically a feature branch ... ;)

hroff-1902:
Not completely -- the changes are already there (say, merged), but okay, let's use it for this purpose.

xmatthias:
ah now i understood - ok can do that :+1:

xmatthias:
Going to close this, since its merged into #2397 - and will be merged to develop soonish.

PR #2286, opened 2019-09-21
Disable Defaulting to DefaultStrategy

xmatthias:
## Summary
As discussed in #2281, this will remove defaulting to DefaultStrategy.

I'm merging this into args-aftersubcommand since there are a TON of conflicting changes.

## Quick changelog

- Require --strategy or "strategy" in configuration to be set.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -95,7 +95,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `db_url` | `sqlite:///tradesv3.sqlite`| Declares database URL to use. NOTE: This defaults to `sqlite://` if `dry_run` is `True`.
 | `initial_state` | running | Defines the initial application state. More information below.
 | `forcebuy_enable` | false | Enables the RPC Commands to force a buy. More information below.
-| `strategy` | DefaultStrategy | Defines Strategy class to use.
+| `strategy` | None | **Required** Defines Strategy class to use. Recommended to set via `--strategy NAME`. 
 | `strategy_path` | null | Adds an additional strategy lookup path (must be a directory).
 | `internals.process_throttle_secs` | 5 | **Required.** Set the process throttle. Value in second.
 | `internals.sd_notify` | false | Enables use of the sd_notify protocol to tell systemd service manager about changes in the bot state and issue keep-alive pings. See [here](installation.md#7-optional-configure-freqtrade-as-a-systemd-service) for more details.
--- a/freqtrade/configuration/cli_options.py
+++ b/freqtrade/configuration/cli_options.py
@@ -66,9 +66,8 @@ AVAILABLE_CLI_OPTIONS = {
     # Main options
     "strategy": Arg(
         '-s', '--strategy',
-        help='Specify strategy class name (default: `%(default)s`).',
+        help='Specify strategy class name which will be used by the bot.',
         metavar='NAME',
-        default='DefaultStrategy',
     ),
     "strategy_path": Arg(
         '--strategy-path',
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -242,6 +242,7 @@ def default_conf(testdatadir):
         "db_url": "sqlite://",
         "user_data_dir": Path("user_data"),
         "verbosity": 3,
+        "strategy": "DefaultStrategy"
     }
     return configuration
 
--- a/tests/strategy/test_strategy.py
+++ b/tests/strategy/test_strategy.py
@@ -55,6 +55,7 @@ def test_load_strategy_base64(result, caplog, default_conf):
 
 
 def test_load_strategy_invalid_directory(result, caplog, default_conf):
+    default_conf['strategy'] = 'SampleStrategy'
     resolver = StrategyResolver(default_conf)
     extra_dir = Path.cwd() / 'some/path'
     resolver._load_strategy('SampleStrategy', config=default_conf, extra_dir=extra_dir)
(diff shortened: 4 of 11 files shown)


docs/configuration.md
```diff
@@ -87,7 +87,7 @@ Mandatory parameters are marked as **Required**.
 | `db_url` | `sqlite:///tradesv3.sqlite`| Declares database URL to use. NOTE: This defaults to `sqlite://` if `dry_run` is `True`.
 | `initial_state` | running | Defines the initial application state. More information below.
 | `forcebuy_enable` | false | Enables the RPC Commands to force a buy. More information below.
-| `strategy` | DefaultStrategy | Defines Strategy class to use.
+| `strategy` | None | **Required** Defines Strategy class to use. Recommendet to set via `--strategy NAME`.
```
hroff-1902: "Recommended"

tests/conftest.py
```diff
@@ -242,6 +242,7 @@ def default_conf(testdatadir):
         "db_url": "sqlite://",
         "user_data_dir": Path("user_data"),
         "verbosity": 3,
+        "strategy": "DefaultStrategy"
```
hroff-1902: SampleStrategy? DefaultStrategy should gone, I guess...

xmatthias: this is for tests... we always used DefaultStrategy for tests - and i would continue to do so assuming that Samplestrategy may be modified by accident ... (happened to me, anyway) ... causing odd test failures :D

hroff-1902: Then this should be a TestStrategy (not loaded in the specific way by the resolver and not used in the specific way, like a fallback etc)?

xmatthias: well it is - and it's not loaded as fallback - however it's accessible using `--strategy DefaultStrategy` - which i think is fine as the user had to select it explicitly.

hroff-1902:
* Should not we add "strategy": "DefaultStrategy" (well, or SampleStrategy if DefaultStrategy will be removed from freqtrade dir) into the config template, so that this will be the (explicitely set in the config instead of previous implicit default val) strategy used by the user if she did not specify custom one in the options of changed this value in the config?
* I assume as a next step, same change should be made for HyperOpt?
* How to be for Hyperopt -- also require specifying the strategy explicitely and do not work without the strategy? For Hyperopt it's not really needed -- if spaces contain both 'buy' and 'sell', then loading pure IStrategy would suffice (if we make it to not contain abstract methods, so that it could be instantiated). Or a special 'EmptyStrategy' derived from IStrategy which does not contain abstract methods... The populate_buy/sell/indicators methods in this strategy may simply raise exceptions if called -- as it's made in the IHyperOpt class...
* I do not want to see this PR in the next upcoming master release 2019.9 (it should go to the next one).

tests/conftest.py
hroff-1902: I did not understand -- will the DefaultStrategy be kept in freqtrade/optimize? If so and if the reason for this is the usage of it in the tests, should not it be 1) moved into the tests dir, 2) be renamed to TestStrategy or TestDefaultStrategy?

freqtrade/configuration/cli_options.py
```diff
@@ -66,9 +66,8 @@ def __init__(self, *args, **kwargs):
     # Main options
     "strategy": Arg(
         '-s', '--strategy',
-        help='Specify strategy class name (default: `%(default)s`).',
+        help='Specify strategy class name that should be used by the bot.',
```
hroff-1902: "which will be used"

tests/conftest.py
xmatthias: i would leave it where it is - at least for now - to allow users to use it via `--strategy DefaultStrategy`. 
If it's an explicit selection by the user, then DefaultStrategy is fine - i just would like to avoid users running this strategy without thinking about it.

xmatthias:
> I do not want to see this PR in the next upcoming master release 2019.9 (it should go to the next one).

This PR is not going to develop anyway, but to args_aftersubcommand2 - so it'll be another round of review before this change even makes it to develop - letalone master / release.

> Or a special 'EmptyStrategy' derived from IStrategy which does not contain abstract methods

This. removing Abstractmethod can create problems if a user implements a strategy wrongly and does not instantiate the necessary methods (or misspells one). There's a reason it's an abstract class...


> Should not we add "strategy": "DefaultStrategy" (well, or SampleStrategy if DefaultStrategy will be removed from freqtrade dir) into the config template, so that this will be the (explicitely set in the config instead of previous implicit default val) strategy used by the user if she did not specify custom one in the options of changed this value in the config?

No. (to the full template, maybe). 
The purpose of this PR is to require strategy to be **explicitly** set by the user. Adding it to config is again having similar to having --strategy default to something.  it'll default to "something" the user may not want, since he did not select it explicitly.

Honestly, I didn't plan to remove DefaultStrategy from the freqtrade folder for now. There's no visible benefit...? 

> How to be for Hyperopt

I didn't think about hyperot too far yet, since my assumption is that hyperopt default-handling will have it's own PR.
I assume an EmptyStrategy could be used for that - but then that'll require more changes to hyperopt (loading the correct methods from the users strategy, ...) - which would explode the scope of this change.

tests/strategy/test_strategy.py
```diff
@@ -65,13 +66,22 @@ def test_load_strategy_invalid_directory(result, caplog, default_conf):
 
 
 def test_load_not_found_strategy(default_conf):
-    strategy = StrategyResolver(default_conf)
+    default_conf['strategy'] = 'NotFoundStrategy'
     with pytest.raises(OperationalException,
                        match=r"Impossible to load Strategy 'NotFoundStrategy'. "
                              r"This class does not exist or contains Python code errors."):
+        strategy = StrategyResolver(default_conf)
         strategy._load_strategy(strategy_name='NotFoundStrategy', config=default_conf)
```
hroff-1902: @xmatthias Are you sure the resolver should be used (and tested) this way?
`_load_strategy()` was already called in the `__init__()` of `StrategyResolver`, the strategy is (should be) loaded and available as `strategy.strategy`

xmatthias: you're right, the exception is raised in the first line.

95299d94c4c824eed2c3aca6fc370a8b85633901 solves this.

Issue #746, opened 2018-05-23
Missed or badly triggered computed Actions: BUY and SELL. Significant Difference Between Backtest and LIVE.


Axel-CH:
**## Step 2: Describe your environment**

  * Python Version: Python 3.6.4
  * Branch: Develop
  * Last Commit ID: b0536db (`git log --format="%H" -n 1`)
 
**## Step 3: Describe the problem:**
Hi everyone, I decided to finaly create an issue because i'm struggling with a problem since a few weeks:
There is an important difference between the signal that are computed on Backtest and the reality of actions in Live.
After a few observation I can tell that this is not related to a normal lag due to the network. The behavior seems to occur when a trigger is slightly triggered. 
**Exemple:** 
Sell When "Close Under Open Of Previous Candle"
dataframe['close'] < dataframe['open'].shift()
_Theoricaly Failing case_
0.00005155 < 0.00005156
(Those values are too close, the condition is not fullfilled)
_Theoricaly Working case_
0.00005155 < 0.00005160
(the gap between the two value is large enought, the condition is fullfilled)

### Steps to reproduce:

  1. Use 1 minute Time frame
  2. launch freqtrade on live with a strategy that use directly price or averages to trigger BUY or SELL actions
  3. Wait for a price movement that "slightly" correspond to the signal (BUY or SELL)
  4. The signal is not triggered or is badly triggered .
  5. When a few moment later there is a bigger price move that trigger the signal more Frankly, the signal 
      is triggered.
  
### Observed Results:

  * What happened?
The signal is not triggered sometimes, when there is a small gap between compared values that are used to take the descision
  * What did you expect to happen?
The signal should be triggered in live more or less at the same moment that is computed on the backtest (less than 1min drift tolerated on live)

### Relevant code exceptions or logs:

<img width="1539" alt="orders-missed-issue" src="https://user-images.githubusercontent.com/7593404/40408477-a63a0164-5e35-11e8-9939-ce30141efae4.png">
<img width="1030" alt="orders-missed-issue-computed-on-backtest-poa" src="https://user-images.githubusercontent.com/7593404/40410503-feb14248-5e3b-11e8-9184-e6827215acbf.png">
Note: On plot-dataframe I modified the code to display only the SELLs that are consecutive to BUYs, to have more clarity.

### Suspected cause
Loss of prescision due to python floating point: the problem with that theory is that I don't understand why the backtest is working as expected. The problem should be also present in that case...

### What I tried to fix that issue

[...]

PR #2294, opened 2019-09-23
Fix skopt memory exhaustion

### LINKED ISSUE #2271: Hyperopt memory leak catch
1134/10000:      3 trades. Avg profit -0.77%. Total profit -0.02327087 ETH (  -2.32Σ%). Avg duration 1261.3 mins. Objective: 2.26564
........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................../home/dri/miner/freqtrade/.env/lib/python3.7/site-packages/joblib/externals/loky/process_executor.py:706: UserWarning: A worker stopped while some jobs were given to the executor. This can be caused by a too short worker timeout or by a memory leak.
  "timeout or by a memory leak.", UserWarning
2019-09-17 22:19:13,769 - freqtrade - ERROR - Fatal exception!
  File "/home/dri/miner/freqtrade/freqtrade/main.py", line 39, in main
    args['func'](args)
  File "/home/dri/miner/freqtrade/freqtrade/optimize/__init__.py", line 74, in start_hyperopt
    hyperopt.start()
  File "/home/dri/miner/freqtrade/freqtrade/optimize/hyperopt.py", line 404, in start
    f_val = self.run_optimizer_parallel(parallel, asked)
  File "/home/dri/miner/freqtrade/freqtrade/optimize/hyperopt.py", line 341, in run_optimizer_parallel
    wrap_non_picklable_objects(self.generate_optimizer))(v) for v in asked)
  File "/home/dri/miner/freqtrade/.env/lib/python3.7/site-packages/joblib/parallel.py", line 934, in __call__
    self.retrieve()
  File "/home/dri/miner/freqtrade/.env/lib/python3.7/site-packages/joblib/parallel.py", line 833, in retrieve
    self._output.extend(job.get(timeout=self.timeout))
  File "/home/dri/miner/freqtrade/.env/lib/python3.7/site-packages/joblib/_parallel_backends.py", line 521, in wrap_future_result
    return future.result(timeout=timeout)
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 435, in result
    return self.__get_result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
joblib.externals.loky.process_executor.TerminatedWorkerError: A worker process managed by the executor was unexpectedly terminated. This could be caused by a segmentation fault while calling the function or by an excessive memory usage causing the Operating System to kill the worker. The exit codes of the workers are {SIGKILL(-9)}

How i can debug this leak ?
It happens every time after some hour of work....
My linux kills the process after some time.

hroff-1902:
Fixes #2271 , #2296 

For the original fix see: https://github.com/scikit-optimize/scikit-optimize/pull/746

Skopt keeps the list of all optimization models found, 
"We store each fitted surrogate model created during each iteration" for people who "need to/want to inspect each surrogate in detail", see https://github.com/scikit-optimize/scikit-optimize/issues/349#issuecomment-293174547

This list is not actually used by our code. Internally, only current model from the estimator is used by skopt.

Skopt is not actively maintained in the meantime, so the fix is added into our hyperopt code as a workaround.

* As a sidechange, the number of iteration is now passed into `generate_optimizer()`, which would allow to make some actions in the worker periodically (cleanup or print logs/stat).

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -36,6 +36,11 @@ logger = logging.getLogger(__name__)
 
 
 INITIAL_POINTS = 30
+
+# Keep no more than 2*SKOPT_MODELS_MAX_NUM models
+# in the skopt models list
+SKOPT_MODELS_MAX_NUM = 10
+
 MAX_LOSS = 100000  # just a big enough number to be bad result in loss optimization
 
 
@@ -255,12 +260,13 @@ class Hyperopt:
             spaces += self.custom_hyperopt.stoploss_space()
         return spaces
 
-    def generate_optimizer(self, _params: Dict) -> Dict:
+    def generate_optimizer(self, _params: Dict, iteration=None) -> Dict:
         """
         Used Optimize function. Called once per epoch to optimize whatever is configured.
         Keep this function as optimized as possible!
         """
         params = self.get_args(_params)
+
         if self.has_space('roi'):
             self.backtesting.strategy.minimal_roi = \
                     self.custom_hyperopt.generate_roi_table(params)
@@ -342,9 +348,25 @@ class Hyperopt:
             random_state=self.config.get('hyperopt_random_state', None)
         )
 
-    def run_optimizer_parallel(self, parallel, asked) -> List:
+    def fix_optimizer_models_list(self):
+        """
+        WORKAROUND: Since skopt is not actively supported, this resolves problems with skopt
+        memory usage, see also: https://github.com/scikit-optimize/scikit-optimize/pull/746
+
+        This may cease working when skopt updates if implementation of this intrinsic
+        part changes.
+        """
+        n = len(self.opt.models) - SKOPT_MODELS_MAX_NUM
+        # Keep no more than 2*SKOPT_MODELS_MAX_NUM models in the skopt models list,
+        # remove the old ones. These are no really needed, the current model
+        # from the estimator is only used.
+        if n >= SKOPT_MODELS_MAX_NUM:
+            logger.debug(f"Fixing skopt models list, removing {n} old items...")
+            del self.opt.models[0:n]
+
+    def run_optimizer_parallel(self, parallel, asked, i) -> List:
         return parallel(delayed(
-                        wrap_non_picklable_objects(self.generate_optimizer))(v) for v in asked)
+                        wrap_non_picklable_objects(self.generate_optimizer))(v, i) for v in asked)
 
     def load_previous_results(self):
         """ read trials file if we have one """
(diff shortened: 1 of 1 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -342,9 +348,25 @@ def get_optimizer(self, dimensions, cpu_count) -> Optimizer:
...
+        WORKAROUND: Since skopt is not actively supported, this resolves problems with skopt
+        memory usage, see also: https://github.com/scikit-optimize/scikit-optimize/pull/746
+
+        This may cease working when skopt updates if implementation of this intrinsic
+        part changes.
+        """
+        n = len(self.opt.models) - SKOPT_MODELS_MAX_NUM
+        # Keep no more than 2*SKOPT_MODELS_MAX_NUM models in the skopt models list,
+        # remove the old ones. These are no really needed, the current model
+        # from the estimator is only used.
```
xmatthias: "is the only one used"? 

maybe from upstream ... ?

hroff-1902: Done, rephrased.

What's "from upstream"?

xmatthias: well my assumption was that part of this solution is also part of the PR against skopt ... which in this case would be the "upstream" project for this workaround.

hroff-1902: no, I borrowed the main idea only, the implementation is dufferent to suite our needs.

To help reviewing: the results (this PR and the develop) can be compared with same `--random-state NNN`, should be same (checked this at my site during preparation of this solution). It proves (besides skopt code review) that history of models is not used in skopt for optimization.

xmatthias: i almost always test hyperopt with `--random-state` (at least when i depend on the results)  ... that's the only way to have reliable results i think...

hroff-1902:
I probably need to explain why 2*10 is used as a limit: 10 is just an arbitrary constant limiting the size of the models list, taken 'from the air'. Then when the list becomes longer than 2*10, first entries (older models) are deleted to get the length of the list 10 again... I tried to make this deletion just after the length is more that 10, but then it deletes one entry every time, while now it's performed every 10 iterations only. So it's a kind of speed optimization. Of course, I could use 20 for the constant and 20/2 instead of 10, but this is just an arbitrary constant...

hroff-1902:
If you still see Loky stopping processes, this might be due to other reasons (other memory consuming procedures),

You may see the actual reasons by adding
```
import mutiprocessing
...
    mp_logger = multiprocessing.log_to_stderr()
    mp_logger.setLevel(logging.DEBUG)
    logging.getLogger('joblib').setLevel(logging.DEBUG)
    logging.getLogger('loky').setLevel(logging.DEBUG)
```
in `set_loggers()` in `freqtrade/loggers.py` and calling `setup_logging(self.config)` at the beginning of the `Hyperopt.generate_optimizer()` method in the hyperopt worker process...

hroff-1902:
> This must have taken ages to debug and find a solution for 🥇

Just a couple of nights...

glouppe:
#746 is now merged

hroff-1902:
(Updated the link above)

@glouppe nice to hear, brilliant, thanks!

FlockrushAdmin:
Still an issue :(

https://github.com/scikit-learn-contrib/skope-rules/issues/18#issuecomment-1223860734

Issue #2289, opened 2019-09-22
Bollinger band styling breaks plot


jraviotta:
The bb styling in freqtrade/plot/plotting.py does not render as expected.
```python
    if 'bb_lowerband' in data and 'bb_upperband' in data:
        bb_lower = go.Scattergl(
            x=data.date,
            y=data.bb_lowerband,
            name='BB lower',
            line={'color': 'rgba(255,255,255,0)'},
        )
        bb_upper = go.Scattergl(
            x=data.date,
            y=data.bb_upperband,
            name='BB upper',
            fill="tonexty",
            fillcolor="rgba(0,176,246,0.2)",
            line={'color': 'rgba(255,255,255,0)'},
        )
        fig.add_trace(bb_lower, 1, 1)
        fig.add_trace(bb_upper, 1, 1)
```

Given I am calling the following from a jupyter notebook:
```python
chart = generate_candlestick_graph(pair=pair,
                                       data=data,
                                       trades=trades,
                                       indicators1=['bb_upperband', 'bb_lowerband'],
                                       indicators2=[])
```

`chart.show(renderer="browser")` produces:
![Screen Shot 2019-09-22 at 6 58 22 PM](https://user-images.githubusercontent.com/3618365/65395584-01073700-dd6b-11e9-93a5-4d2733d9b303.png)

When the above code is removed, this is produced:

![Screen Shot 2019-09-22 at 6 49 44 PM](https://user-images.githubusercontent.com/3618365/65395554-66a6f380-dd6a-11e9-809b-12427b324081.png)

In addition to the problems rendering, the automatic addition of the indicators is not obvious. I understand offering some defaults for the user, but if duplicates are specified in `indicators1[]`, it is not obvious why the legend contains duplicates without looking at the source.
![Screen Shot 2019-09-22 at 7 02 29 PM](https://user-images.githubusercontent.com/3618365/65395622-90144f00-dd6b-11e9-920c-becf26e32fda.png)

I can create a PR that removes the offending lines if that is helpful.

xmatthias:
The problem with bollinger bands is how they should be rendered, which is as a band, which is not possible by using `indicators1` or `indicators2`.

That's why this is added as "default" if it's available. 
You're right, documentation seems to miss this.

![2019-09-23-061745_1502x759_scrot](https://user-images.githubusercontent.com/5024695/65401637-e4e6b280-ddc9-11e9-8010-4e8914cb3194.png)

[...]

PR #2297, opened 2019-09-24
Enhancements to BB plotting

jraviotta:
## Summary
This solves rendering of bolinger bands and removes duplicate plot elements.

Solves issue #2289 
Related to upstream bug defined in https://github.com/plotly/plotly.js/issues/2284

## Quick changelog

- substitutes  `plotly.graph_objects.Scatter` for `plotly.graph_objects.Scattergl`
- removes duplicate legend items

## What's new?
All uses of `Scattergl` were replaced including those that were not specifically problematic. After upstream bug is fixed, the change can be reverted. New logic checks `indicators1` list and removes `bb_upperband` and `bb_lowerband` if added by the user. Changed label for BB in legend.

### DIFF
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -64,14 +64,14 @@ def add_indicators(fig, row, indicators: List[str], data: pd.DataFrame) -> make_
     """
     for indicator in indicators:
         if indicator in data:
-            # TODO: Figure out why scattergl causes problems
-            scattergl = go.Scatter(
+            # TODO: Figure out why scattergl causes problems plotly/plotly.js#2284
+            scatter = go.Scatter(
                 x=data['date'],
                 y=data[indicator].values,
                 mode='lines',
                 name=indicator
             )
-            fig.add_trace(scattergl, row, 1)
+            fig.add_trace(scatter, row, 1)
         else:
             logger.info(
                 'Indicator "%s" ignored. Reason: This indicator is not found '
@@ -92,7 +92,7 @@ def add_profit(fig, row, data: pd.DataFrame, column: str, name: str) -> make_sub
     :param name: Name to use
     :return: fig with added profit plot
     """
-    profit = go.Scattergl(
+    profit = go.Scatter(
         x=data.index,
         y=data[column],
         name=name,
@@ -222,23 +222,26 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
             logger.warning("No sell-signals found.")
 
     if 'bb_lowerband' in data and 'bb_upperband' in data:
-        bb_lower = go.Scattergl(
+        bb_lower = go.Scatter(
             x=data.date,
             y=data.bb_lowerband,
-            name='BB lower',
+            showlegend=False,
             line={'color': 'rgba(255,255,255,0)'},
         )
-        bb_upper = go.Scattergl(
+        bb_upper = go.Scatter(
             x=data.date,
             y=data.bb_upperband,
-            name='BB upper',
+            name='Bollinger Band',
             fill="tonexty",
             fillcolor="rgba(0,176,246,0.2)",
             line={'color': 'rgba(255,255,255,0)'},
         )
         fig.add_trace(bb_lower, 1, 1)
         fig.add_trace(bb_upper, 1, 1)
-
+        if 'bb_upperband' in indicators1 and 'bb_lowerband' in indicators1:
+            indicators1.remove('bb_upperband')
+            indicators1.remove('bb_lowerband')
+            
     # Add indicators to main plot
     fig = add_indicators(fig=fig, row=1, indicators=indicators1, data=data)
 
@@ -252,7 +255,7 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
     )
     fig.add_trace(volume, 2, 1)
 
-    # Add indicators to seperate row
+    # Add indicators to separate row
     fig = add_indicators(fig=fig, row=3, indicators=indicators2, data=data)
 
     return fig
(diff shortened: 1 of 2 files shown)


freqtrade/plot/plotting.py
```diff
@@ -64,14 +64,14 @@ def add_indicators(fig, row, indicators: List[str], data: pd.DataFrame) -> make_
     """
     for indicator in indicators:
         if indicator in data:
-            # TODO: Figure out why scattergl causes problems
-            scattergl = go.Scatter(
+            # TODO: Figure out why scattergl causes problems plotly/plotly.js#2284
```
xmatthias: I would move the plotly reference to the bollinger band section, as it's very specific to how that is rendered (the bugreport does not appy to the following lines since they don't use `fill="tonexty"`

freqtrade/plot/plotting.py
```diff
@@ -222,23 +222,26 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
             logger.warning("No sell-signals found.")
 
     if 'bb_lowerband' in data and 'bb_upperband' in data:
-        bb_lower = go.Scattergl(
+        bb_lower = go.Scatter(
             x=data.date,
             y=data.bb_lowerband,
-            name='BB lower',
+            showlegend=False,
```
xmatthias: Very nice! :+1:

freqtrade/plot/plotting.py
```diff
@@ -267,7 +270,7 @@ def generate_profit_graph(pairs: str, tickers: Dict[str, pd.DataFrame],
     df_comb = create_cum_profit(df_comb, trades, 'cum_profit')
 
     # Plot the pairs average close prices, and total profit growth
-    avgclose = go.Scattergl(
+    avgclose = go.Scatter(
```
xmatthias: I am not certain about this change (and the one in `add_profit()`)...

Did this usage cause some problems or crashes for you?

It's obviously problematic for Bollingerbands due to the very specific rendering (as a Band) ... but works fine for other uses on my machine.

The reason to use `Scattergl` over `Scatter`is performance for larger plots, since `Scatter` creates them as SVG objects (which causes rendering problems in the browser with a certain amount of points), while GL uses a more optimized method of rendering, allowing a LOT more points in one plot.

jraviotta: I removed all occurrences of Scattergl out of an abundance of caution. My assumption is that webgl and plotly may still have some problems upstream and if a chart element is resource intensive enough to benefit from webgl, then it will probably break. If it does not break then it is probably not resource intensive and won't suffer from rendering performance problems.

You might consider waiting to implement webgl until plotly proves that it is not buggy. 

I can revert these changes and only change the BB to Scatter if you like. It's up to you.

xmatthias: the problem with BB rendering is not resource intensiveness, but a simple programming error (with python short-circuiting breaking something too early...).

personally i would leave `Scattergl` for all other operations until we face issues ... especially for longer-term Profit-plots (multiple months) the speed-improvements are non neglectable

jraviotta: I have noticed that volume does not display once a timerange crosses some threshold. For example 1 month of 5min tickers does not show the volume bars. Have you seen that behavior and is it possibly related? Do you want me to log an issue?

freqtrade/plot/plotting.py
jraviotta: See comment below. Testing confirms that the problem is with `Scattergl`.

freqtrade/plot/plotting.py
jraviotta: I experimented with rendering using `Scattergl` and found that any inclusions broke something. When I reverted the substitutions I made in the last commit and then rendered bollinger bands without gl. That broke the profit graph. Wherever I used `Scattergl`, the chart displayed incorrectly.

I tried restarting my jupyter server and restarting the kernel after making code changes and calling `%autoreload`. This occurred in Brave, Chrome and Safari so I'm confident that this is a real bug with `Scattergl` and not just bollinger bands.

I'm using a Macbook air so perhaps that is resource constrained but I can't test on my other machine until my new CPU cooler arrives. Still, I wouldn't expect two weeks of 5min tickers to cause any problems.

You may want to confirm on a different machine or consider merging this PR without `Scattergl`.

xmatthias: when testing such things - i would reccomend to test them as mainly intendet - using the cli interface, which (by sharing the cli command used) should allow us to recreate the issue you're seeing).
Even using ! runs it "slightly" different and in a jupyter context, which could have unknown consequences and can taint results.

I've just tried both `plot-profit` and `plot-dataframe` with 1.5 months 5m plots (using develop) and found no issue (other than bollinger not rendering correctly).

Also volume is there (but barely visible, since the bars tiny, and only visible once you zoom into the picture to only see a day or two).

Please also verify the version of plotly - it should be 4.1.1 at the moment (i'm lazy at updating this too sometimes ...)

jraviotta: plotly==4.1.1

I tracked the bug to `fig.show(renderer='browser')` from a notebook.
With the exception of the BB, plotting works from the cli and the charts display correctly inline in a notebook. However,  `scattergl` does not work correctly when jupyter opens plots in the browser. 

I also applied some stying to the volume bars to make them more visible.

xmatthias: so the scattergl problem is a jupyter problem, not a problem happening when using the cli?

(now i'm not against this change if it creates problems, it's more to understand the problem).

jraviotta: Correct. In 83e596c I reverted `avgclose =` and `profit =` back to `go.Scattergl`.

When using Scatter.gl:

|                   |        cli           | jupyter inline | jupyter `renderer='browser'`|
| ---------- | ------------ | ------------- | ---------------------------- |
| BB | X | X | X |
| `avgclose = go.Scatter.gl` |           + | + | X |
| `profit = go.Scatter.gl` | + | + | X |

\+ = works
X = broken

xmatthias: this finding about scattergl not working in browser renderer might be worth reporting to plotly so they can fix it in a future release.

However, it's fine to switch everything to `Scatter` for now.
Please have  a look at the tests. The failures are logical considering the changes done.
Any issues with that let me know and i'll lend you a hand with them.

Issue #1639, opened 2019-03-08
Don't compare floats for equality


hroff-1902:
**(Almost) never compare floats for equality...**

I found a few, in the freqtradebot.py:

* `if float(order['filled']) == 0`
* `float(order['remaining']) == 0.0`

Both (at least first, I do not understand the meaning of second comparison yet) should probably be something like:

```val <= 0.0 + (10 ** (-market_precision)) / 2``` (i.e. less then 0.5 satoshi in case of BTC)
or probably ```math.isclose(val, 0.0, abs_tol = (10 ** (-market_precision)) / 2)```

More comparisons of floats for equality should appear in the code, I've seen something suspicious for fees, for example...

In the future the Decimal type with appropriate precision (corresponding to particular currency and market) may be used instead of floats, but that seems to be hard to implement...

P.S.
More in freqtradebot.py:
* `if trade.fee_open == 0 or order['status'] == 'open':`
* `if order['amount'] != new_amount:`
* `if amount != order_amount:`
* `if fee_abs != 0:`

**Comment by xmatthias** [comment]
> float(order['remaining']) == 0.0

This simply assumes that if there is nothing remaining,  ... (whatever comes next).
there is no difference between `float(xxx)==0.0` or `float(xxx) == 0`:
```
>>> 0 == 0.0
True
>>> 0 == float(0.0)
True
```

I also doubt that exchanges fill this small amounts, since all exchanges I've looked at have a "min-lotsize" ([binance](https://support.binance.com/hc/en-us/articles/115000594711-Trading-Rule) - which depends on the coin (in binances example, that's 0.00000001 -> 1 Satoshi - so a remaining value of < 0.5 Satoshi should not be possible ... unless i miss something ... ?

**Comment by hroff-1902** [comment]
> > float(order['remaining']) == 0.0
> 
> This simply assumes that if there is nothing remaining, ... (whatever comes next).
> there is no difference between `float(xxx)==0.0` or `float(xxx) == 0`:
> 
> ```
> >>> 0 == 0.0
> True
> >>> 0 == float(0.0)
> True
> ```

Sure. It's not about changing 0.0 --> 0 or 0 --> 0.0 here, not at all about it. It's about rounding errors in floats, especially when values are results of calculation rather than casting of integers or strings into floats. Even trivial calculations. See the example:

```
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
[...]

PR #2307, opened 2019-09-26
Don't compare floats when updating fees

xmatthias:
## Summary
As reported on slack, sometimes the amount does not match if an order is filled in multiple trades, and the sum results in a float rounding error.

This can be easily solved by using `math.isclose()` - however since we can have very small numbers, we should set `rel_tol` to a lower floating point to avoid calculation problems.

part of the issue: #1639 (will probably not cover all of it, since this looks at this specific problem).

## Quick changelog

- Add tests to verify this problem and make sure this won't break in the future
- introduce `MATH_CLOSE_PREC` constant - to have one place to specify this rounding
- use math.isclose() to compare amounts

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -22,6 +22,7 @@ ORDERTYPE_POSSIBILITIES = ['limit', 'market']
 ORDERTIF_POSSIBILITIES = ['gtc', 'fok', 'ioc']
 AVAILABLE_PAIRLISTS = ['StaticPairList', 'VolumePairList']
 DRY_RUN_WALLET = 999.9
+MATH_CLOSE_PREC = 1e-14  # Precision used for float comparisons
 
 TICKER_INTERVALS = [
     '1m', '3m', '5m', '15m', '30m',
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -511,7 +511,7 @@ class FreqtradeBot:
                         trade.pair.startswith(exectrade['fee']['currency'])):
                     fee_abs += exectrade['fee']['cost']
 
-        if not isclose(amount, order_amount, rel_tol=constants.MATH_CLOSE_PREC):
+        if not isclose(amount, order_amount, abs_tol=constants.MATH_CLOSE_PREC):
             logger.warning(f"Amount {amount} does not match amount {trade.amount}")
             raise OperationalException("Half bought? Amounts don't match")
         real_amount = amount - fee_abs
@@ -536,7 +536,7 @@ class FreqtradeBot:
             # Try update amount (binance-fix)
             try:
                 new_amount = self.get_real_amount(trade, order)
-                if not isclose(order['amount'], new_amount, rel_tol=constants.MATH_CLOSE_PREC):
+                if not isclose(order['amount'], new_amount, abs_tol=constants.MATH_CLOSE_PREC):
                     order['amount'] = new_amount
                     # Fee was applied, so set to 0
                     trade.fee_open = 0
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -6,6 +6,7 @@ import copy
 import logging
 import traceback
 from datetime import datetime
+from math import isclose
 from typing import Any, Dict, List, Optional, Tuple
 
 import arrow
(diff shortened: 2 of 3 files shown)


hroff-1902:
Need to study PEP0485 https://www.python.org/dev/peps/pep-0485/ first in details, it contains wide retrospective on the theory behind and approach implemented in python ("Behavior near zero", "How much difference does it make?")...

xmatthias:
It's better to use this link (or the asnwer in it): https://stackoverflow.com/questions/5595425/what-is-the-best-way-to-compare-floats-for-almost-equality-in-python/33024979 - which lists the equivalent for older python versions.

For developers, the code is sometimes easier to interpret ... 

https://github.com/python/cpython/blob/6b5f1b496f0b20144592b640b9c975df43a29eb0/Modules/mathmodule.c#L2784-L2825

hroff-1902:
should not the precision 1) be abs_tol 2) be a precision for the currency from the ccxt exchange data?

xmatthias:
no, i don't think so.

keep in mind that we're trying to fix a floating point rounding error, not a calculation error on either side - these don't appear "early" in the decimal places - at least i've never seen that. 

Taking your example from #1639:
```
print(format(0.3 - 0.1 - 0.1 - 0.1, '.20f'))
-0.00000000000000002776
```

the result is not 0.0 - however, the roundin error happens after 16 zeros - not after 8 or less.

## For this specific usecase:
taking binance as example (since i know the problem there best):

assume we buy amount 10 of currency XXX, which has a lotsize of 0.1.

Binance will take the fee either from BNB (which is no problem) or from the "target" currency, XXX in this case, subtracting this fee from my requested amount of 10.
Now the fee will not be 0.1 (that would be 1%) - but instead 0.025 (0.25% - for math simplicity).
This is way below lotsize (which is what should be used for amounts) - effectively leaving me with 9.975 XXX. 
Effectively, since i can only sell in lots of 0.1, i will end up with 0.75 "binance dust" once this trade closes.



Now let's look at the binance result regarding precision:
```
{'percentage': True,
  'tierBased': False,
  'taker': 0.001,
  'maker': 0.001,
  'precision': {'base': 8, 'quote': 8, 'amount': 3, 'price': 6},
  'limits': {'amount': {'min': 0.001, 'max': 100000.0},
   'price': {'min': 1e-06, 'max': 100000.0},
   'cost': {'min': 0.0001, 'max': None}},
  'id': 'ETHBTC',
  'symbol': 'ETH/BTC',
  'base': 'ETH',
  'quote': 'BTC',
  'baseId': 'ETH',
  'quoteId': 'BTC',
  'info': {'symbol': 'ETHBTC',
   'status': 'TRADING',
   'baseAsset': 'ETH',
   'baseAssetPrecision': 8,
   'quoteAsset': 'BTC',
   'quotePrecision': 8,
}
   ...
}
```

precision is 8 for both base and quote. However, since we're having this problem with amount, we would need to use `amount` to be correct (no point in rounding amount to a price precision) - which has a precision of 3 - which i think is quite low for this type of comparison (with 3 - 0.0054 and 0.0055 would be identical).


``` python
math.isclose(0.0054, 0.0055, abs_tol=0.001)                                                                                   
True
```

Also keep in mind that these are "tradeable precisions" - not the precision binance will round to internally:

While ETH has a amount precision of 3 - you can still own it to a WAY lower percentage even though you can't trade it - which results in Binance dust (i have a lot of that in various currencies ...).

We do account for this on sell - using it to compare if a trade was filled completely or not may be dangerous, since (due to the above fee oddity) we can have bought an amount with more decimals than what are "tradeable".

hroff-1902:
Ah, it's amount, not price, with it's own precision on the exchange...

Okay, why not `abs_tol=1e-14`?

xmatthias:
because it is really a constant value we're using here - so `constants.py` seems a good place for it... 

and assuming we find this problem in other locations of the code, i would not want to have this a few times (or a few times slightly different)...

hroff-1902:
erhm, I mean why it's rel_tol and not abs_tol?
(Just want to understand the details of the decision made)

xmatthias:
to be honest, it's the first parameter (of the 2) in the argument list, worked for all cases i tested, and based on short circuiting, will be evaluated first (look at the last line of the C code linked above).

I assume abs_tol would work, too - but did not test it (and honestly, i don't read every PEP before using a function it introduces ...)

xmatthias:
Don't put my decisions into question please... that'll have me start do the same ... :smile: 

rethinking this and testing it further i think you're right, `abs_tol` might be the safer way to go.

```python
In [2]: math.isclose(0.0054, 0.0055, abs_tol=1e-3)                                                                                     
Out[2]: True

In [3]: math.isclose(0.0054, 0.0055, rel_tol=1e-3)                                                                                     
Out[3]: False
```

on the other hand:
```
In [3]: math.isclose(20.0054, 20.0055, abs_tol=1e-3)                                                                                  
Out[3]: True

In [4]: math.isclose(20.0054, 20.0055, rel_tol=1e-3)                                                                                  
Out[4]: True

```

hroff-1902:
That's what I was also thinking about... 

LGTM 👍

PR #2308, opened 2019-09-26
Allow use of config in custom hyperopt methods

hroff-1902:
This PR passes config into hyperopt interface init method and saves it in the attributes.

This will allow to use values from the config in the custom populate_* methods (for example, compare values against stake_amount, use stake_currency for creation the names of the columns, etc. -- all as in the strategies).

### DIFF
--- a/freqtrade/optimize/hyperopt_interface.py
+++ b/freqtrade/optimize/hyperopt_interface.py
@@ -36,6 +36,12 @@ class IHyperOpt(ABC):
     """
     ticker_interval: str
 
+    def __init__(self, config: dict) -> None:
+        self.config = config
+
+        # Assign ticker_interval to be used in hyperopt
+        IHyperOpt.ticker_interval = str(config['ticker_interval'])
+
     @staticmethod
     @abstractmethod
     def populate_indicators(dataframe: DataFrame, metadata: dict) -> DataFrame:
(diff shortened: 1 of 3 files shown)


freqtrade/optimize/hyperopt_interface.py
```diff
@@ -36,6 +36,12 @@ class IHyperOpt(ABC):
     """
     ticker_interval: str
 
+    def __init__(self, config: dict) -> None:
+        self.config = config
+
+        # Assign ticker_interval to be used in hyperopt
+        IHyperOpt.ticker_interval = str(config['ticker_interval'])
```
xmatthias: Consider i've implemented my own constructor in the child-hyperopt-class and don't call `super().__init__(config)` (it's not documented anywhere since we don't use this by default, so how would i know?) so config and ticker_interval are not set (although i would expect them to be since i use them ... )

It can become a silent, odd failure that'll be very difficult to debug.

Now on the other hand, we do the same with the Strategy Interface and have had no problems related to it so far ...

hroff-1902:
Hmm, I guess I definitely tried the same change (`MagicMock(return_value=hyperopts(default_conf)`), the config should obviously be passed to it), but it did not fix the failing test... I'll look at this later in details... (Note to myself)

Issue #2259, opened 2019-09-12
[Hyperopt default stoploss space] - [Discussion]


gaugau3000:
Hi,

From what i have understand the default stop loss got to -0.5 to -0.02. But even for a daily bar strategy i am not sure it is ok :-) . Maybe define a dynamic stop loss space depending on the ticker time frame has it is the case for the roi tables. Another problem is that can fool the loss function if it is based on loss/drawdowns and misguidance for new users.

Have a nice day.

gaugau3000:
I'm sure i have a bias on how i build my strategy about the stop loss. I consider for example if i am on a 5 minutes ticker what it is the sense to have a 20% stop loss has it maybe reprends 100 candles duration to be reached. I consider 100 candle from entry is more randomness than trading (and long trend market dependent). But 20% stop loss is maybe ok form 1h/4h/1d. What it is sure is that maybe embedded trading way of thinking is bad as a trading bot need to be agnostic (my bad).

hroff-1902:
Well, the boundaries in the stoploss space are only boundaries. Hyperopt will try different values between them in its path to the minimum of the objective function. 

If it produces epochs with higher (by its absolute value, i.e. closer to -0.50) stoplosses -- it actually means that higher values of the stoplosses produce better results for the objective function. Take a highest value (by its absolute value) of the stoploss you can accept in this case. This depend on other criteria like risks, capital in risk, etc.

If the minimum is somewhere near stoploss=-0.10, the epochs should be grouped near that value for the stoploss parameter, even if the upper boundary for the stoploss space is set to -0.50. The bigger stoplosses (by its absolute value, i.e. those closer to this -0.50 boundary) will simply produce worse results, worse values of the objective, so the optimizer will quickly descends to the neighborhood of the local minumum, i.e, to -0,10 value in this case.

I do not see how to formulate any rule for adaptation of the boundaries for the stoploss space depending on timeframe. 

It's more a question of understanding the results rather that a question of particular (shorter) values for the stoploss space, imho.

PR #2315, opened 2019-09-28
Minor: Change the default stoploss space

hroff-1902:
...to align the docs with #2314 and #2259 
(forgotten to do it in #2314)


hroff-1902:
Also changed examples (for ROI as well) to align them to the actual output. So pls re-review

docs/hyperopt.md
```diff
@@ -375,23 +375,12 @@ Buy hyperspace params:
...
+{   0: 0.10674,
+    21: 0.09158,
+    78: 0.03634,
     118: 0}
 ```
 
-This would translate to the following ROI table:
-
-``` python
-minimal_roi = {
```
xmatthias: Why would we remove this? 
yes, the table is correct above, too - but the explanation that this is "copy/pasteable" is not available anymore...

hroff-1902: Because now it's not needed to 'translate' the ROI anymore, it's printed in the format that can simply be copy-pasted into the strategy... (Am I wrong here?)

xmatthias: you are right - but it's not mentioned anymore ... 
So "you can use this ROI output and use it as `minimal_roi`" should be added instead i guess

hroff-1902: But it was not mentioned as 'copy-paste', the text said something about 'translation' (which was mostly rounding of the values), which is no longer needed since the values are already rounded.
I reworked this part (and the stoploss counterpart), pls review.
Also added new subsections for the parts describing default spaces.

PR #2317, opened 2019-09-29
Add list-timeframes subcommand

### LINKED ISSUE #2216: Easy way to list supported timeframes for exchange(s)
* Operating system: macOS 10.14.5
  * Python Version: Python 3.7.2
  * CCXT version: ccxt==1.18.1112
  * Branch: Develop
  * Last Commit ID: 962d48
  
Add a way to list available timeframes for exchanges.
Given that strategies are defined for a certain timeframe, it'd be nice to know which exchanges support that timeframe for backtesting and/or running the strategy.

### Steps to reproduce:

  1. Set timeframe to 4h
  2. Try to download data / backtest using bittrex
  3. Notice error: `2019-09-02 15:22:24,289 - freqtrade - ERROR - Invalid ticker 4h, this Exchange supports {'1m': 'oneMin', '5m': 'fiveMin', '30m': 'thirtyMin', '1h': 'hour', '1d': 'day'}`

hroff-1902:
This PR adds the `list-timeframes` subcommand.
available options: `--exchange`, `--one-column/-1`

Closes #2216 

Usage:
```
$ freqtrade list-timeframes --exchange binance
$ freqtrade -c config_binance.json list-timeframes 
```

Iterating through all exchanges (remove `config.json`, otherwise you'll get frequent `Pair is not available` error, that's the different story):
```
$ for i in `freqtrade list-exchanges -1`; do echo '====================='; freqtrade list-timeframes --exchange $i; sleep 1; done
```

TODO:
* [x] docs
* [x] tests (??? have no idea)


hroff-1902:
I wonder that the default config is fetched, if it exists, for the command (this one and others) that are in the NO_CONF_REQUIRED list. In my opinion, the empty config should always be used for those utility commands (or at least some of them) if no config is specified explicitely with `-c`...

xmatthias:
I would add one or two basic tests for, simply making sure that it's working and we're not breaking it unintentionally.

`test_list_exchanges` in test_utils.py or similar 

## default config

While i agree that it can be strange to have the default-config loaded in this case, it would be very inconsistent to remove this behaviour for one or 2 commands.

It's not a technical question, really, using a 2nd subparser-list which prevents the automatism completely should be fairly simple, we just need to make sure to not break anything. 

Not sure if we shouldn't think about this in / after #2264 though, which will throw all of this around anyway.

hroff-1902:
> ## default config
> While i agree that it can be strange to have the default-config loaded in this case, it would be very inconsistent to remove this behaviour for one or 2 commands.

Okay, I was thinking about this and I guess I know how to mitigate it in degree by refactoring exchange... (No changes to the part loading config will be needed). Not in this PR. for sure, let it be here as it is now.

> It's not a technical question, really, using a 2nd subparser-list which prevents the automatism completely should be fairly simple, we just need to make sure to not break anything.

Sure, it's more a question of the user cli interface structure and the behavior of the application, it's less a technical point than a point of user experience...

> Not sure if we shouldn't think about this in / after #2264 though, which will throw all of this around anyway.

Ehm, did not get this... Please describe (shortly)...

> I would add one or two basic tests for, simply making sure that it's working and we're not breaking it unintentionally.
> 
> `test_list_exchanges` in test_utils.py or similar
> 

Yes, please. I'll then probably add some other tests using these as the template...

hroff-1902:
I refactored the stuff a bit:
* All intrinsic logic of the command is reimplemented via the use of Exchange class methods -- we already have (almost) all in it...
* The `list-timeframes` command fails now with the exit code 2 'automagically' for unsupported exchanges that do not provide timframes due to the OperationalException in the `Exchange.__init__`
* The call to `validate_ticker()` in the `Exchange.__init__` was moved upper, to be called before loading markets -- it reduces the number of warnings and ccxt errors for exchanges that do not provide timeframes and cannot load tickers
* The signature of `validate_timeframes()` was fixed -- it accepts the ticker as a string, not a list of strings
* A log message slightly improved.

TODO (not in this PR):
* `validate_timeframes()` should be renamed to `validate_timeframe()`. I can include it in this PR.
* The global list of unsupported exchanges should be updated with those that do not provide timeframes.
* Refactoring exchange to avoid the use of pairs, markets and stuff when called in utilities (see above, this is a completely different story).

xmatthias:
> Not sure if we shouldn't think about this in / after #2264 though, which will throw all of this around anyway.

this was regarding the change / modification of how to load the configuration / `config.json`

> Yes, please. I'll then probably add some other tests using these as the template...

Maybe we misunderstand each other - that type of template does does exist already in test_utils.py ... ?  

> validate_timeframes() should be renamed to validate_timeframe(). I can include it in this PR.

since the signature was changed, this makes sense to do now i think (before the name was correct)... :) 
It'll break almost all tests for a second, until the occurence in `conftest.py` is changed (small hint so you're not surprised when this applies).

> Refactoring exchange to avoid the use of pairs, markets and stuff when called in utilities (see above, this is a completely different story).

this should be as simple as adding a "nonproductive" property to the exchange class constructor ... ? but i agree, should not be done here.

For this command, the solution for the "pairlist" issue may be even simpler (this may even remove the need to do the above refactoring:
Use the minimal_config template and set what you need (which should be only `exchange.name`), making sure everyhing "that hurts" is not present.

hroff-1902:
> since the signature was changed, this makes sense to do now i think (before the name was correct)... :)

No, according to the current implementation it currently worked only with single timeframe (due to `in <list>`):
https://github.com/freqtrade/freqtrade/blob/704fea616b048c531e94a18b155ba064f8d4a2dd/freqtrade/exchange/exchange.py#L308-L310

The `List` in the typehint was therefore errorneous. That's why I fixed it this way. And after that the name should be changed, when both implementation and signature imply other behavior than the name says.

> It'll break almost all tests for a second, until the occurence in `conftest.py` is changed (small hint so you're not surprised when this applies).

Yes, I see, thanks for the hint. I'll change it in a separate PR to make this one compact and observable.

hroff-1902:
> Maybe we misunderstand each other - that type of template does does exist already in test_utils.py ... ?

I mean, I'll use your newly added tests as a template for my ones I'll also add if they will be needed additionally.

> this should be as simple as adding a "nonproductive" property to the exchange class constructor ... ? but i agree, should not be done here.
> 
> For this command, the solution for the "pairlist" issue may be even simpler (this may even remove the need to do the above refactoring:
> Use the minimal_config template and set what you need (which should be only `exchange.name`), making sure everyhing "that hurts" is not present.

I thought it can be refactored with introducing a BaseExchange which is the base class for Exchange ('normal' exchange). This will init ccxt, provide some base properties and the utilities may instantiate this one instead of 'normal' exchange. But, yes, probably this will be done with a 'non-productive' flag. Not in this PR for sure.

hroff-1902:
Is there a place for documenting this command (and the `list-exchanges`)? 

Maybe a separate `utils.md` doc file should be created (placed just after Start the bot in the main doc menu) for them?

freqtrade/configuration/arguments.py
```diff
@@ -41,7 +43,7 @@
 ARGS_PLOT_PROFIT = ["pairs", "timerange", "export", "exportfilename", "db_url",
                     "trade_source", "ticker_interval"]
 
-NO_CONF_REQURIED = ["download-data", "plot-dataframe", "plot-profit"]
+NO_CONF_REQURIED = ["download-data", "list-timeframes", "plot-dataframe", "plot-profit"]
```
xmatthias: i suspect `"list-timeframes"` should now be moved to `NO_CONF_ALLOWED` ...

It doesn't make sense to use a "real" config for this command i think.

xmatthias:
> Is there a place for documenting this command (and the list-exchanges)?

`utils.md` should work, could also be placed right before FAQ, but i guess both locations work just fine.

I've been asking myself all day if we should make a `freqtrade info` subcommand ...
with all the `list` commands we get quite a list of commands ...
we could have `freqtrade info list-exchanges`, `freqtrade info list-timeframes` with all the "informative" commands ...

it's just an idea ... maybe commands get too long, in this case.

hroff-1902:
> I've been asking myself all day if we should make a `freqtrade info` subcommand ...
> with all the `list` commands we get quite a list of commands ...
> we could have `freqtrade info list-exchanges`, `freqtrade info list-timeframes` with all the "informative" commands ...
> 
> it's just an idea ... maybe commands get too long, in this case.

I also thought of this, I do not think it should be this way. For me, `freqtrade list <list-subsubcommand>`, `freqtrade show <show-subsubcommand>` looks better. `info` in `info list-something` looks redundant... Or, `freqtrade info exchanges` as a possible solution -- where `info` is used instead of `list`...

xmatthias:
Ideally it should be one subcommand for all "info" / "list" commands ... i guess we'll then also rename them accordingly.
Let's note that as an idea for the future ... for now we only got 2 so it's still okish...

freqtrade/configuration/arguments.py
hroff-1902: Are you sure? Currently, NO_CONF_ALLOWED contains only commands that use args and do not utilize config...

For this command is convenient to read the default config and print timeframes for the exchange set in it. This is from one point of view. From other point of view, right, this command may require exchange be set with the `--exchange` or config be explicitely specified with `-c`. I don't know which option is best here...

hroff-1902:
Utils.md added, some other changes to docs made, pls review last commit

freqtrade/configuration/arguments.py
xmatthias: > Now the behavior is strange: the config.json exists, I want simply list the timeframes for another exchange (so I specified it with --exchange, because I know that my default config.json is set to work with other exchange), and it tells me something about non-existing pairs...

Well i was basing this on your comment above, that it'll be strange if you have config.json and want to list timeframes for another exchange ...

both are valid options and depend on what information you're looking for ...

docs/utils.md
```diff
@@ -0,0 +1,26 @@
+# Utility Subcommands
+
+Besides the Live-Trade and Dry-Run run modes, the `backtesting`, `edge` and `hyperopt` optimization subcommands, and the `download-data` subcommand which prepares historical data, the bot contains a number of utility subcommands. They are described in this section.
+
+## List Exchanges
+
+```
```
xmatthias: I would add a sentense to each of the commands that quickly explains what it's usefull for.

kinda like
`Use this to find out which exchanges are supported by freqtrade`

and 
`Use this to see which timeframes are available for your exchange` ... (or similar ...)

hroff-1902: Done, also added some examples

docs/utils.md
```diff
@@ -0,0 +1,56 @@
...
+  --exchange EXCHANGE  Exchange name (default: `bittrex`). Only valid if no
+                       config is provided.
+  -1, --one-column     Print exchanges in one column.
+
+```
+
+* Example: see the timeframes for the 'binance' exchange, set in the configuration file:
+
+```
+$ freqtrade -c config_binance.json
```
xmatthias: i suspect the example should use `freqtrade -c config_binance.json list-timeframes` ...

hroff-1902: Sure, thanks, fixed (rebased)

tests/test_utils.py
```diff
@@ -48,6 +49,77 @@ def test_list_exchanges(capsys):
...
+        start_list_timeframes(pargs)
+
+    # Test with --config config.json.example
+    args = [
+        '--config', 'config.json.example',
+        "list-timeframes",
+    ]
+    start_list_timeframes(get_args(args))
+    captured = capsys.readouterr()
+    assert re.match("Timeframes available for the exchange `bittrex`: "
```
hroff-1902: @xmatthias do we have a `out_has` function which behaves like `log_has` but on the output printed? As you see, strings here are not regexps to use re.match against them...

hroff-1902:
ready for final review and merge, I guess.

tests/test_utils.py
xmatthias: no we don't - but if you check the first 2 functions in `conftest.py` - writing that is not difficult...

tests/test_utils.py
```diff
@@ -48,6 +49,77 @@ def test_list_exchanges(capsys):
...
+    # Test with --exchange bittrex
+    args = [
+        "list-timeframes",
+        "--exchange", "bittrex",
+    ]
+    start_list_timeframes(get_args(args))
+    captured = capsys.readouterr()
+    assert re.match("Timeframes available for the exchange `bittrex`: "
+                    "1m, 5m, 30m, 1h, 1d",
+                    captured.out)
```
xmatthias: the way this is tested makes it fail whenever ccxt (or rather the exchange) changes the timeframes they offer.
Now i don't expect this to happen very often - but it could - so maybe using a "loser" regex could be benefitial (allowing more timeframes to be added in the middle / in the end).

The way this is added to ccxt:
https://github.com/ccxt/ccxt/blob/5b6f039a639fc69ae575af8d627e330a7656cbb9/python/ccxt/bittrex.py#L58-L63

Edit: Yes, i had to try the multiline review comment - it's not really tied to these 3 lines.

tests/test_utils.py
```diff
@@ -48,6 +49,77 @@ def test_list_exchanges(capsys):
...
+    with pytest.raises(OperationalException,
+                       match=r"This command requires a configured exchange.*"):
+        start_list_timeframes(pargs)
+
+    # Test with --config config.json.example
+    args = [
+        '--config', 'config.json.example',
+        "list-timeframes",
+    ]
+    start_list_timeframes(get_args(args))
```
xmatthias: You will need to mock some of the exchange.
right now it's calling `load_markets()` which will make this test depend on the exchange being available

I'm not sure if `patch_exchange()` will work in this case - maybe you'll need to only patch specific parts...

tests/test_utils.py
hroff-1902: > the way this is tested makes it fail whenever ccxt (or rather the exchange) changes the timeframes they offer.
> Now i don't expect this to happen very often - but it could - so maybe using a "loser" regex could be benefitial (allowing more timeframes to be added in the middle / in the end).

Yes, I guessed it and assumed we'll need to adapt the tests in this case... Should this be completely mocked instead? (And the test shortened to work with one mocked exchange data?)

tests/test_utils.py
hroff-1902: That's why I said I need a template for testing this subcommand... 😄

xmatthias: well but "this" template may not work for other subcommands ...

tests/test_utils.py
xmatthias: yes (but not shortening) - see the changes i've just pushed.
mocking it to 2 different things has the advantage that you cannot force it "somewhere" to be this output...

hroff-1902: ah, thanks alot!

tests/test_utils.py
```diff
@@ -49,8 +49,16 @@ def test_list_exchanges(capsys):
     assert re.search(r"^bittrex$", captured.out, re.MULTILINE)
 
 
-def test_list_timeframes(capsys):
-
+def test_list_timeframes(mocker, capsys):
+
+    api_mock = MagicMock()
+    api_mock.timeframes = {'1m': '1m',
```
hroff-1902: I guess, keys and vals should differ in this dict (arbitrary vals can be used like `'1m': '1m one', '5m': '5m two'`) to be sure in the test that the keys and not values are used by the bot...

xmatthias: well yes - but i copied most of it from here ... :D 

https://github.com/ccxt/ccxt/blob/ea91f5f738f5338e5087677c0b619931831c1caa/python/ccxt/binance.py#L50-L65

now we have both versions - binance and bittrex ...

Issue #2134, opened 2019-08-13
Kraken: Historical data limited to 720 candles 


Saber1998:
Attempting to utilize the Kraken exchange and would like to perform backtests but I am unable to pull data going back further than 5 days at 1h candles.

xmatthias:
The issue is best explained here: [kraken-support-trade-data](https://support.kraken.com/hc/en-us/articles/218198197-How-to-pull-all-trade-data-using-the-Kraken-REST-API)

starting with https://github.com/ccxt/ccxt/issues/5677 (for kraken) - we should be able to utilize trade-level data to gather historic candle data.

PR #2319, opened 2019-09-30
Add exchanges to the list of bad exchanges

hroff-1902:
Extend the list of bad exchanges with those that does not provide timeframes.

The ids of the exchanges were extracted from the output of the new `list-timeframes` command.

Exchanges appended with two different descriptions, since I assume those that have `fetchOHLCV: emulated` may theoretically be fixed in ccxt in the future...


xmatthias:
> I assume those that have fetchOHLCV: emulated may theoretically be fixed in ccxt in the future...

I think your assumption is wrong.
`fetchOHLCV: emulated` means that the exchange does not provide an ohlcv endpoint which returns the data directly. ccxt can only "fix" that if the exchange adds an endpoint for this, and they are usually quite quick about such things.

From the ccxt docs:

> Some exchanges don't offer any OHLCV method, and for those, the ccxt library will emulate OHLCV candles from Public Trades. In that case you will see exchange.has['fetchOHLCV'] = 'emulated'. However, because the trade history is usually very limited, the emulated fetchOHLCV methods cover most recent info only and should only be used as a fallback, when no other option is available.

[source](https://github.com/ccxt/ccxt/wiki/Manual#ohlcv-emulation)

I think we've discussed this before "somewhere" - and i would not blacklist these exchanges just because they don't have timeframes populated in ccxt since it does not automatically mean that the exchange does not work with freqtrade (it probably won't depending on how much history such a call provides, but that would need to be tested per exchange).

Theoretically, Based on ccxt code, you'll have even more possibilities for timeframes in this case, since ccxt will generate the candles by resampling the trades, so you could also have 2m candles.


Now there are no "common" or popular exchanges in this list, however i would would want you to add 2 different reasons for `bitstamp1` and `hitbtc` - namely "please use the new exchange name from ccxt, bitstamp2" ... since they each have a corresponding `bitstamp2` and `hitbtc2` version which don't suffer from this problem.

I did not check every exchange for this - but i remember these 2 from some past issues.

hroff-1902:
> > I assume those that have fetchOHLCV: emulated may theoretically be fixed in ccxt in the future...
> 
> I think your assumption is wrong.
> `fetchOHLCV: emulated` means that the exchange does not provide an ohlcv endpoint which returns the data directly. ccxt can only "fix" that if the exchange adds an endpoint for this, and they are usually quite quick about such things.
> 
> From the ccxt docs:
> 
> > Some exchanges don't offer any OHLCV method, and for those, the ccxt library will emulate OHLCV candles from Public Trades. In that case you will see exchange.has['fetchOHLCV'] = 'emulated'. However, because the trade history is usually very limited, the emulated fetchOHLCV methods cover most recent info only and should only be used as a fallback, when no other option is available.
> 
> [source](https://github.com/ccxt/ccxt/wiki/Manual#ohlcv-emulation)
> 
> I think we've discussed this before "somewhere" - and i would not blacklist these exchanges just because they don't have timeframes populated in ccxt since it does not automatically mean that the exchange does not work with freqtrade (it probably won't depending on how much history such a call provides, but that would need to be tested per exchange).
> 
> Theoretically, Based on ccxt code, you'll have even more possibilities for timeframes in this case, since ccxt will generate the candles by resampling the trades, so you could also have 2m candles.
> 

But those exchanges (with fetchOHLCV: 'emulated' and no timeframes attribute in the ccxt) fail in `validate_timeframes`:
https://github.com/freqtrade/freqtrade/blob/704fea616b048c531e94a18b155ba064f8d4a2dd/freqtrade/exchange/exchange.py#L298-L305

Okay, this even better. I'll change the check in terms of #2317 so that exchanges with no `timeframes` and with fetchOHLCV: 'emulated' can pass the check in `validate_timeframes()`, and remove these exchanges here from the list of bad exchanges.

> Now there are no "common" or popular exchanges in this list, however i would would want you to add 2 different reasons for `bitstamp1` and `hitbtc` - namely "please use the new exchange name from ccxt, bitstamp2" ... since they each have a corresponding `bitstamp2` and `hitbtc2` version which don't suffer from this problem.

Okay, fine. Which one, `bitstamp1` or `bitstamp` is correct? (There is no `bitstamp2`)

> I did not check every exchange for this - but i remember these 2 from some past issues.
> 
> **edit**:
> Something else entirely - i think `list-exchanges` should not include these exchanges (it does now ...). Currently these are included, which makes this command return exchanges we don't support - which is kinda pointless...

Sure, I also was going to change it this way. Not in this PR.

xmatthias:
> Okay, this's even better. I'll change the check in terms of #2317 so that exchanges with no timeframes and with fetchOHLCV: 'emulated' can pass the check in validate_timeframes(), and remove these exchanges here from the list of bad exchanges.

No, please leave it as is, you're right to block these exchanges. 
To get "some" candles from the trades multiple calls will be needed (trade-data is very detailed ...), which will get us blocked anyway.
I know this since i'm working on something in this regards for #2134 ... 

> Okay, fine. Which one, bitstamp1 or bitstamp is correct? (There is no bitstamp2)

Based on [this](https://github.com/ccxt/ccxt/wiki/Manual#exchanges) `bitstamp` is correct, `bitstamp1` uses the old API (v1 vs. v2 for `bitstamp`).

hroff-1902:
> No, please leave it as is, you're right to block these exchanges.
> To get "some" candles from the trades multiple calls will be needed (trade-data is very detailed ...), which will get us blocked anyway.
> I know this since i'm working on something in this regards for #2134 ...

Okay (I prepared the commit for #2317, but won't push it). Should I drop the latest commit here as well?

xmatthias:
yes, i'd do that ... 
we know they're probably not going to work ... was still early this morning :D

hroff-1902:
> yes, i'd do that ...
> we know they're probably not going to work ... was still early this morning, sorry :D

No problem.
Dropped, done.

hroff-1902:
> Based on [this](https://github.com/ccxt/ccxt/wiki/Manual#exchanges) `bitstamp` is correct, `bitstamp1` uses the old API (v1 vs. v2 for `bitstamp`).

We have `bitstamp` already blacklisted: 
https://github.com/freqtrade/freqtrade/blob/b6ee3d99b177605faa837020d03a4834592f4d83/freqtrade/exchange/exchange.py#L31-L32

Should this be removed and `bitstamp` enabled?

xmatthias:
no - didn't see this (sorry again) ... then we need to blacklist both bitstamp exchange versions...

hroff-1902:
> no - didn't see this (sorry again) ... ccxt supports too many exchanges, most having a similar name :(
> 
> then we need to blacklist both bitstamp exchange versions...

`hitbtc` -- I added an entry, referring to `hitbtc2`; please review
`bitstamp` -- was already blacklisted, with a separate message
`bitstamp1` -- blacklisted with other similar exchanges as having `fetchOHLCV: 'emulated'`

PR #2331, opened 2019-10-02
Remove unused test-data

xmatthias:
## Summary
remove unused test-data to reduce repository size.

There is no point in keeping data nobody is ever touching.
It's also not useful to users, since the data is outdated by now, and uptodate data should be used for backtesting and  hyperopt to reflect realistic, uptodate market scenarios.

Part of #2112


tests/data/test_btanalysis.py
```diff
@@ -108,15 +108,15 @@ def test_load_trades(default_conf, mocker):
 
 
 def test_combine_tickers_with_mean(testdatadir):
-    pairs = ["ETH/BTC", "XLM/BTC"]
+    pairs = ["ETH/BTC", "ADA/BTC"]
```
hroff-1902: Why all "XLM/BTC" changed to "ADA/BTC"?

xmatthias: Because i thought i could remove XLM/BTC data too- but that would have an impact to backtesting running in CI so i had to revert that (it would not fail the test - but cause different results ... ).

This will allow us to remove this in the future eventually (but i kept the files in the folder for now).

I can revert this here too - but realistically, the pair does not matter for these tests - so standardizing on as few as possible seems to make sense (ADA/BTC is used surprisingly often in tests)

PR #2313, opened 2019-09-28
Fix #2295: better logging for whitelist processing

hroff-1902:
Fixes: #2295 

Log messages added: I tried to stay not noisy, but provide useful info in the log, which would help to understand manipulations with the active pair whitelist the bot performs.

The log level of the debug message in `create_trades()` was changed to `info` for the same reasons.

Typical bot throttling iterations now look in the log like the following (with the Edge is On):
```
2019-09-28 03:40:53,235 - freqtrade.freqtradebot - INFO - PairList returned active pair whitelist: 1 pairs.
2019-09-28 03:40:53,235 - freqtrade.freqtradebot - INFO - Edge adjusted active pair whitelist: 0 pairs.
2019-09-28 03:40:53,240 - freqtrade.freqtradebot - INFO - Active pair whitelist is empty.
2019-09-28 03:40:53,240 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
2019-09-28 03:40:57,239 - freqtrade.freqtradebot - INFO - PairList returned active pair whitelist: 1 pairs.
2019-09-28 03:40:57,239 - freqtrade.freqtradebot - INFO - Edge adjusted active pair whitelist: 0 pairs.
2019-09-28 03:40:57,246 - freqtrade.freqtradebot - INFO - Active pair whitelist is empty.
2019-09-28 03:40:57,246 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
```

and with Edge is Off:
```
2019-09-28 03:47:00,657 - freqtrade.freqtradebot - INFO - PairList returned active pair whitelist: 1 pairs.
2019-09-28 03:47:01,285 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
2019-09-28 03:47:04,660 - freqtrade.freqtradebot - INFO - PairList returned active pair whitelist: 1 pairs.
2019-09-28 03:47:04,675 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
2019-09-28 03:47:08,663 - freqtrade.freqtradebot - INFO - PairList returned active pair whitelist: 1 pairs.
2019-09-28 03:47:08,676 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
2019-09-28 03:47:12,667 - freqtrade.freqtradebot - INFO - PairList returned active pair whitelist: 1 pairs.
[...]

PR #2332, opened 2019-10-02
Freqtradebot refactoring

hroff-1902:
* `process_maybe_execute_buy()` renamed to `process_maybe_execute_buys()` because it executes mutiple buys.
* Same for `process_maybe_execute_sell()`, to be symmetric.
* `process_maybe_execute_sell()` (now `process_maybe_execute_sells()`) accepts now the list of trades and iterate through them internally.
* Code of `process_maybe_execute_sells()` cleaned up, pls review.
* "Whitelist is empty." message should not be a warning.
* 'whitelist' was changed to 'active pair whitelist' in some of the log messages. (so probably this is enough to be applied instead of #2313 )
* Tests adjusted accordingly

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -1,7 +1,6 @@
 """
 Freqtrade is the main module of this bot. It contains the class Freqtrade()
 """
-
 import copy
 import logging
 import traceback
@@ -135,12 +134,11 @@ class FreqtradeBot:
                                   self.strategy.informative_pairs())
 
         # First process current opened trades
-        for trade in trades:
-            self.process_maybe_execute_sell(trade)
+        self.process_maybe_execute_sells(trades)
 
         # Then looking for buy opportunities
         if len(trades) < self.config['max_open_trades']:
-            self.process_maybe_execute_buy()
+            self.process_maybe_execute_buys()
 
         if 'unfilledtimeout' in self.config:
             # Check and handle any timed out open orders
@@ -262,11 +260,10 @@ class FreqtradeBot:
         Checks pairs as long as the open trade count is below `max_open_trades`.
         :return: True if at least one trade has been created.
         """
-        interval = self.strategy.ticker_interval
         whitelist = copy.deepcopy(self.active_pair_whitelist)
 
         if not whitelist:
-            logger.warning("Whitelist is empty.")
+            logger.info("Active pair whitelist is empty.")
             return False
 
         # Remove currently opened and latest pairs from whitelist
@@ -276,7 +273,8 @@ class FreqtradeBot:
                 logger.debug('Ignoring %s in pair whitelist', trade.pair)
 
         if not whitelist:
-            logger.info("No currency pair in whitelist, but checking to sell open trades.")
+            logger.info("No currency pair in active pair whitelist, "
+                        "but checking to sell open trades.")
             return False
 
         buycount = 0
@@ -285,8 +283,10 @@ class FreqtradeBot:
             if self.strategy.is_pair_locked(_pair):
                 logger.info(f"Pair {_pair} is currently locked.")
                 continue
+
             (buy, sell) = self.strategy.get_signal(
-                _pair, interval, self.dataprovider.ohlcv(_pair, self.strategy.ticker_interval))
+                _pair, self.strategy.ticker_interval,
+                self.dataprovider.ohlcv(_pair, self.strategy.ticker_interval))
 
             if buy and not sell and len(Trade.get_open_trades()) < self.config['max_open_trades']:
                 stake_amount = self._get_trade_stake_amount(_pair)
@@ -431,10 +431,9 @@ class FreqtradeBot:
 
         return True
 
-    def process_maybe_execute_buy(self) -> None:
+    def process_maybe_execute_buys(self) -> None:
         """
-        Tries to execute a buy trade in a safe way
-        :return: True if executed
+        Tries to execute buy trades in a safe way
         """
         try:
             # Create entity and execute trade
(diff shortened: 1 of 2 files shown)


hroff-1902:
Actually, I found a difference in the new logic in `process_maybe_execute_sells()` comparing to the existing one:

Previously it called `self.handle_trade(trade)` if `handle_stoploss_on_exchange(trade)` returned False. Now, it calls `self.handle_trade(trade)` if `handle_stoploss_on_exchange(trade)` did not called at all (due to `elif`). Please review this.

Maybe I should return it back, leaving only other minor changes...

freqtrade/freqtradebot.py
```diff
@@ -443,33 +442,28 @@ def process_maybe_execute_buy(self) -> None:
...
-        return False
+        for trade in trades:
+            try:
+                self.update_trade_state(trade)
+
+                if trade.is_open:
+                    result = False
+                    if self.strategy.order_types.get('stoploss_on_exchange'):
+                        result = self.handle_stoploss_on_exchange(trade)
+                    elif trade.open_order_id is None:
```
xmatthias: This is wrong (as you discovered already ...)
even if `stoploss_on_exchange` is enabled, we still need to `handle_trade()` - otherwise `stoploss_on_exchange` remains the only sell-reason that's technically possible if it's enabled.
You can use `continue` if stoploss_on_exchanged fired instead of elif i guess ... (in combination with the other comment a few lines above ...)

What surprises me more than this small issue is that no test complains about this.

freqtrade/freqtradebot.py
```diff
@@ -443,33 +442,28 @@ def process_maybe_execute_buy(self) -> None:
...
-
-        except DependencyException as exception:
-            logger.warning('Unable to sell trade: %s', exception)
-        return False
+        for trade in trades:
+            try:
+                self.update_trade_state(trade)
+
+                if trade.is_open:
+                    result = False
```
xmatthias: Can you do the following? 

* move `result = False` out of the loop
* run `wallets.update()` after the loop, based if any trade was changed, so it runs only once per "sell"-iteration if necessary, even if more than one sell occured.

hroff-1902: It's even better! 👍 

* fixed the logic in `process_maybe_execute_sells()`
* I removed the `if trade.is_open` check there -- it's redundant, the `trades` list of trades prepared in `process()` is already a list of open trades. Even if it's updated somewhere in the `Trades.update()` (seems it's not), the `handle_trade()` will simply quit with the exception
* The exception in the `handle_trade()` changed from ValueError to DependencyException to be catched safely here in process_maybe_execute_sells()

freqtrade/freqtradebot.py
hroff-1902: Got it, fixed the logic, see below...

freqtrade/freqtradebot.py
```diff
@@ -443,33 +442,29 @@ def process_maybe_execute_buy(self) -> None:
         except DependencyException as exception:
             logger.warning('Unable to create trade: %s', exception)
 
-    def process_maybe_execute_sell(self, trade: Trade) -> bool:
+    def process_maybe_execute_sells(self, trades: List[Any]) -> None:
         """
-        Tries to execute a sell trade
-        :return: True if executed
+        Tries to execute sell trades in a safe way
```
xmatthias: Tries to sell trades in a safe way (?) 

the "execute" seems misplaced now

freqtrade/freqtradebot.py
```diff
@@ -443,33 +442,29 @@ def process_maybe_execute_buy(self) -> None:
...
-                    self.wallets.update()
-                    return result
+        result = False
+        for trade in trades:
+            try:
+                self.update_trade_state(trade)
 
-            if trade.is_open and trade.open_order_id is None:
+                if (self.strategy.order_types.get('stoploss_on_exchange') and
+                        self.handle_stoploss_on_exchange(trade)):
```
xmatthias: While it's not really the intend of this PR - i would want to see a test which has 2 (or more) trades, `stoploss_on_exchange=True` and `self.handle_stoploss_on_exchange` mocked in a way to return True for one trade, and False for the other (this one should have `handle_trade()` mocked to True).

It's about the logic in this function, which should call both methods once in the above sample, and call wallets.update only one time (not twice as it would have done before).

I was surprised this morning when we didn't have a test which tries to sell more than one trade at a time - and i think it makes sense to add this as part of this refactoring.

If you have difficulties with that test (i know, pytest is not you're expertise) let me know and i'll have a look ...

hroff-1902:
Marked as 'don't merge' for a while -- seems I found another issue and need to check if it is a problem in the current codebase or was introduced by this PR.

freqtrade/freqtradebot.py
hroff-1902: > If you have difficulties with that test (i know, pytest is not you're expertise) let me know and i'll have a look ...

Yeah, that was not covered by the tests...

Yes, please, add the test -- this one still seems to be too difficult to me, all those mocking...

freqtrade/freqtradebot.py
hroff-1902: I guess the word 'trades' is misplaced here (and in the docstring for _buy() counterpart). The `process_maybe_execute_sells` executes sells (sell orders), same for `process_maybe_execute_buys`-- it executes buys (buy orders)...

xmatthias: actually it checks trades (so trades is correct i think) for sell (or buy) reasons and eventually places a sell-order if necesary ... (it doesn't blindly execute a sell-order, that's what `execute_sell()` does.

freqtrade/freqtradebot.py
xmatthias: it's mostly copy and paste from other tests, really ... 
and a bit of thinking about "how deep do i want the execution to go ..."

Mocking itself in this case was not difficult

freqtrade/freqtradebot.py
hroff-1902: Changed docstrings, pls review.

hroff-1902:
@xmatthias thanks for the tests!

Ready for review and merge, I guess...

tests/test_freqtradebot.py
```diff
@@ -2526,13 +2530,101 @@ def test_may_execute_sell_after_stoploss_on_exchange_hit(default_conf,
...
+        'freqtrade.exchange.Exchange',
+        get_ticker=ticker,
+        get_fee=fee,
+        markets=PropertyMock(return_value=markets),
+        symbol_amount_prec=lambda s, x, y: y,
+        symbol_price_prec=lambda s, x, y: y,
+        get_order=stoploss_order_mock,
+    )
+    # Sell 3rd trade (not called for the first trade)
+    handle_trade_mock = MagicMock(side_effect=[False, True])
```
xmatthias: > Ready for review and merge, I guess...

i know and i did review your changes for the most part already so they only need "one last look" ... but i'm not yet content with the test i wrote.

It mocks out `handle_trade()`, and because of this, ony the first trade object is updated correctly, we cannot verify if the other 2 are updated correctly, making the return values of this function irrelevant to the test outome.

I'll try to improve this today or tomorrow, then this should be ready to go.

hroff-1902: > i know and i did review your changes for the most part already so they only need "one last look" ... but i'm not yet content with the test i wrote.

My 'ready for merge' statement was just a short description of the current state from my side, there can then always be several, more than one, further iterations of changes/reviewing, this is the normal part of the development process, we both know it and should not explain it to each other... 🍻

xmatthias: yet - its better to be clear as to why i think it's not ready yet than leaving you (and potential observers) in the dark

xmatthias: I would appreciate if you could have a look through this test - if the test (as it is) is ok for you we're good to go

PR #2335, opened 2019-10-03
Allow to use market data in the strategies

hroff-1902:
...via dataprovider.

This will allow to use limits and precisions (for any reasons, if it's needed), fees, check the status of activity for informative pairs, and get the quote/base currency for the pair without any parsing (ccxt has already parsed it for us):

```
...
    data1 = self.dp.ohlcv(pair=f'{self.dp.market(metadata["pair"])["base"]}/USDT', ticker_interval='1h')
```

This should also be available for backtesting (with current activity flags for the pairs though; all 'static' data as fees, base/quote strings, limits and precisions are okay)

TODO:
* [ ] add description to the docs

### DIFF
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -85,6 +85,14 @@ class DataProvider:
             logger.warning(f"No data found for ({pair}, {ticker_interval}).")
         return data
 
+    def market(self, pair: str):
+        """
+        Return market data for the pair
+        :param pair: Pair to get the data for
+        :return: Market data dict from ccxt or None if market info is not available for the pair
+        """
+        return self._exchange.markets.get(pair)
+
     def ticker(self, pair: str):
         """
         Return last ticker data


freqtrade/data/dataprovider.py
```diff
@@ -85,6 +85,14 @@ def get_pair_dataframe(self, pair: str, ticker_interval: str = None) -> DataFram
             logger.warning(f"No data found for ({pair}, {ticker_interval}).")
         return data
 
+    def market(self, pair: str):
+        """
+        Return market data for the pair
+        :param pair: Pair to get the data for
+        :return: Market data dict from ccxt or None if market info is not available for the pair
+        """
+        return self._exchange.markets.get(pair)
```
xmatthias: i'm not sure on the naming...

* `ccxt.load_markets()`
* `exchange.markets`
* `dp.market()`

all wrap the same thing (in this order, mind you) ... won't it be confusing since it's 3 names for the same information...

hroff-1902: for the users only the last is to be available, since exchange (and ccxt api) is not accessible in the strategy

Should make it a property maybe?

xmatthias: no - if it's a property you cannot pass in a parameter ... 

Maybe a quick test making sure the correct type is returned? (yes, with mocked markets ...)

hroff-1902: as for the names:

ccxt is completely (well, almost, we have config for it in the config) instrinsic thing that should not be visible for a user, so it's not a question how does it call its method

exchange has the `markets` property, but it's a dict of markets, the name is okay

dp so has now a method to return market data for a pair, so it's `market`... I could name it `market_info` or `get_pair_market_data` but do not see any reasons for this...


As for checks -- what kind of checks can be made here with python duck typing? It's just a dict with data...

hroff-1902:
A short description was added into the docs, with a link to ccxt docs.
Let me improve this docs with examples later, not in this PR.

PR #2343, opened 2019-10-05
Move experimental settings to ask_strategy

### LINKED ISSUE #2233: Redesign experimental settings
The suggestion is:

* move experimental settings out of experimental to the root:
```
    "experimental": {
        "use_sell_signal": false,
        "sell_profit_only": false,
        "ignore_roi_if_buy_signal": false
    },
```
They are implemented for more than an year and should no longer be considered as experimental. The code for experimental branch of settings should be kept in the codebase since we could add some new settings to it in the future.

* assign the `use_sell_signal` the default value of true.

hroff-1902:
* Experimental settings moved to `['ask_strategy']`.
* The default value for `ask_strategy.use_sell_signal` changed to True (true in json config)
* Docs adjusted (description of `use_sell_signal` slightly improved)
* Tests adjusted
* An error in json schema fixed (it was using `ignore_roi_if_buy_signal_true` instead of proper `ignore_roi_if_buy_signal`)

Resolves #2233 

TODO:
* [x] copy values of old experimental settings to ask_strategy with a deprecation message (we only have this implemented for deprecated cli options)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -65,6 +65,9 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `ask_strategy.use_order_book` | false | Allows selling of open traded pair using the rates in Order Book Asks.
 | `ask_strategy.order_book_min` | 0 | Bot will scan from the top min to max Order Book Asks searching for a profitable rate.
 | `ask_strategy.order_book_max` | 0 | Bot will scan from the top min to max Order Book Asks searching for a profitable rate.
+| `ask_strategy.use_sell_signal` | true | Use your strategy sell signals in addition to the `minimal_roi`. [Strategy Override](#parameters-in-the-strategy).
+| `ask_strategy.sell_profit_only` | false | Wait until you have made a positive profit before taking a sell decision. [Strategy Override](#parameters-in-the-strategy).
+| `ask_strategy.ignore_roi_if_buy_signal` | false | Do not sell if the buy-signal is still active. This setting takes preference over `minimal_roi` and `use_sell_signal`. [Strategy Override](#parameters-in-the-strategy).
 | `order_types` | None | Configure order-types depending on the action (`"buy"`, `"sell"`, `"stoploss"`, `"stoploss_on_exchange"`). [More information below](#understand-order_types). [Strategy Override](#parameters-in-the-strategy).
 | `order_time_in_force` | None | Configure time in force for buy and sell orders. [More information below](#understand-order_time_in_force). [Strategy Override](#parameters-in-the-strategy).
 | `exchange.name` |  | **Required.** Name of the exchange class to use. [List below](#user-content-what-values-for-exchangename).
@@ -78,9 +81,6 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `exchange.ccxt_async_config` | None | Additional CCXT parameters passed to the async ccxt instance. Parameters may differ from exchange to exchange  and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
 | `exchange.markets_refresh_interval` | 60 | The interval in minutes in which markets are reloaded.
 | `edge` | false | Please refer to [edge configuration document](edge.md) for detailed explanation.
-| `experimental.use_sell_signal` | false | Use your sell strategy in addition of the `minimal_roi`. [Strategy Override](#parameters-in-the-strategy).
-| `experimental.sell_profit_only` | false | Waits until you have made a positive profit before taking a sell decision. [Strategy Override](#parameters-in-the-strategy).
-| `experimental.ignore_roi_if_buy_signal` | false | Does not sell if the buy-signal is still active. Takes preference over `minimal_roi` and `use_sell_signal`. [Strategy Override](#parameters-in-the-strategy).
 | `experimental.block_bad_exchanges` | true | Block exchanges known to not work with freqtrade. Leave on default unless you want to test if that exchange works now.
 | `pairlist.method` | StaticPairList | Use static or dynamic volume-based pairlist. [More information below](#dynamic-pairlists).
 | `pairlist.config` | None | Additional configuration for dynamic pairlists. [More information below](#dynamic-pairlists).
@@ -116,9 +116,9 @@ Values set in the configuration file always overwrite values set in the strategy
 * `process_only_new_candles`
 * `order_types`
 * `order_time_in_force`
-* `use_sell_signal` (experimental)
-* `sell_profit_only` (experimental)
-* `ignore_roi_if_buy_signal` (experimental)
+* `use_sell_signal` (ask_strategy)
+* `sell_profit_only` (ask_strategy)
+* `ignore_roi_if_buy_signal` (ask_strategy)
 
 ### Understand stake_amount
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -59,15 +59,15 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `unfilledtimeout.sell` | 10 | **Required.** How long (in minutes) the bot will wait for an unfilled sell order to complete, after which the order will be cancelled.
 | `bid_strategy.ask_last_balance` | 0.0 | **Required.** Set the bidding price. More information [below](#understand-ask_last_balance).
 | `bid_strategy.use_order_book` | false | Allows buying of pair using the rates in Order Book Bids.
-| `bid_strategy.order_book_top` | 0 | Bot will use the top N rate in Order Book Bids. Ie. a value of 2 will allow the bot to pick the 2nd bid rate in Order Book Bids.
+| `bid_strategy.order_book_top` | 0 | Bot will use the top N rate in Order Book Bids. I.e. a value of 2 will allow the bot to pick the 2nd bid rate in Order Book Bids.
 | `bid_strategy. check_depth_of_market.enabled` | false | Does not buy if the % difference of buy orders and sell orders is met in Order Book.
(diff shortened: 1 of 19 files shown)


tests/optimize/test_backtesting.py
```diff
@@ -645,8 +648,9 @@ def test_backtest_alternate_buy_sell(default_conf, fee, mocker, testdatadir):
     mocker.patch('freqtrade.optimize.backtesting.file_dump_json', MagicMock())
     backtest_conf = _make_backtest_conf(mocker, conf=default_conf,
                                         pair='UNITTEST/BTC', datadir=testdatadir)
-    # We need to enable sell-signal - otherwise it sells on ROI!!
-    default_conf['experimental'] = {"use_sell_signal": True}
+    # use_sell_signal is True by default now, it will sell on the sell signal instead of ROI,
+    # which is what we need here
+#    default_conf['ask_strategy'] = {"use_sell_signal": True}
```
xmatthias: i guess since this is not needed anymore we can also remove this part (?) from the tests? (the comment does not make sense in the future - if you look at it from a "after the PR is merged" perspective ... (same with the others in this file ...)

docs/configuration.md
```diff
@@ -65,6 +65,9 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `ask_strategy.use_order_book` | false | Allows selling of open traded pair using the rates in Order Book Asks.
 | `ask_strategy.order_book_min` | 0 | Bot will scan from the top min to max Order Book Asks searching for a profitable rate.
 | `ask_strategy.order_book_max` | 0 | Bot will scan from the top min to max Order Book Asks searching for a profitable rate.
+| `ask_strategy.use_sell_signal` | true | Use your strategy sell signals in addition to the `minimal_roi`. [Strategy Override](#parameters-in-the-strategy).
+| `ask_strategy.sell_profit_only` | false | Wait until you have made a positive profit before taking a sell decision. [Strategy Override](#parameters-in-the-strategy).
```
xmatthias: maybe we should not use "you" and "your" here but instead "use the strategy" and "Wait until the bot has made" ... ?

xmatthias:
Aditional note: i guess the standard in "SampleStrategy" should also be changed - it defaults to `False` now - 
but personally, i like to remove these settings from the config and use them from within the strategy - since this is really strategy dependent.

hroff-1902:
I would leave the settings in the config schema, to keep an ability to set them in the config, but would remove them in the sample config templates (except 'full').

So let's leave it for a couple of months in the configs so that people adapt with their new location, them remove in the config samples along with (or a month after) removal of the deprecation warnings...

tests/optimize/test_backtesting.py
hroff-1902: I thought I would be better 'from a "after the PR is merged" perspective' to emphasize that use_sell_signal: True is required for these tests... Okay, removed, even better...

hroff-1902:
> Aditional note: i guess the standard in "SampleStrategy" should also be changed - it defaults to `False` now -

Nice point! 👍 
(Also made some improvements to the wordings there)

docs/configuration.md
hroff-1902: pls review

hroff-1902:
Note to myself: Deprecation warning is to be added before merge

tests/optimize/test_backtesting.py
xmatthias: it's the new default - someone new will not know that this was different in the past, so might be confused by this comment

xmatthias:
looks good so far :+1: 

> Note to myself: Deprecation warning is to be added before merge

this and the temporary "automatic" migration are the complicated part i think ... ;)

hroff-1902:
> this and the temporary "automatic" migration are the complicated part i think ... ;)

no, the complicated part is the tests... 😎 

pls review

Maybe you'll want to change the messages.

I made it 'universal' (more generic than it could be in this particular case), with the section names passed as parameters (in both code handling the settings in the Configuration and in the test), so that it can later be used for other deprecated or conflicting settings without major changes...

tests/test_configuration.py
```diff
@@ -897,3 +898,27 @@ def test_pairlist_resolving_fallback(mocker):
...
+
+@pytest.mark.parametrize("setting", [
+        ("ask_strategy", "use_sell_signal", True,
+         "experimental", "use_sell_signal", False),
+        ("ask_strategy", "sell_profit_only", False,
+         "experimental", "sell_profit_only", True),
+        ("ask_strategy", "ignore_roi_if_buy_signal", False,
+         "experimental", "ignore_roi_if_buy_signal", True),
+    ])
+def test_deprecated_experimental(mocker, default_conf, setting, caplog):
```
xmatthias: i think we should also test `process_deprecated_setting()` and `check_conflicting_settings()` seperately, assuming that they we keep them once we remove experimental completely (unless you plan on removing them again ...).

Doesn't have to be existing settings we test in this case either - just something to confirm that it works.

Otherwise they'll become dead code until we change settings around again - which could be problematic in case of refactorings, since they will never be hit (neither in tests, nor in operational mode).

tests/test_configuration.py
```diff
@@ -897,3 +898,27 @@ def test_pairlist_resolving_fallback(mocker):
...
+    patched_configuration_load_config_file(mocker, default_conf)
+
+    default_conf[setting[3]] = {}
+    default_conf[setting[0]][setting[1]] = setting[2]
+    default_conf[setting[3]][setting[4]] = setting[5]
```
xmatthias: maybe a comment "#set old value" and "#set new value" (for the above e lines) would benefit readability
had to think twice what what number actually is ...

freqtrade/configuration/deprecated_settings.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+
+def process_deprecated_setting(config: Dict[str, Any],
+                               section1: str, name1: str,
+                               section2: str, name2: str):
+    section2_config = config.get(section2, {})
+
+    if name2 in section2_config:
+        logger.warning(
```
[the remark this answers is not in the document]
hroff-1902: I tried warning.warn, it does not print anything in my environment, do not know the reason yet.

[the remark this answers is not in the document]
xmatthias: i think it prints the warning "sometimes", unless deprecationwarnings are switched on.

we have this in our tests - `warnings.simplefilter("always")` ... but i think there's also a environment variable for this (or a python compile flag?)

tests/test_configuration.py
hroff-1902: Right, agree.

tests/test_configuration.py
hroff-1902: Agree, I also thought about the elements when I returned back to this test...

freqtrade/configuration/configuration.py
```diff
@@ -106,6 +111,8 @@ def load_config(self) -> Dict[str, Any]:
```
xmatthias: Should `block_bad_exchanges` be moved to `exchange` ? it's really an exchange setting, not an experimental setting ... ?

xmatthias:
just 2 more questions (block_bad_exchanges, and https://github.com/freqtrade/freqtrade/pull/2343#discussion_r332828814).

I guess you missed the comment abuout that TODO comment

freqtrade/configuration/configuration.py
hroff-1902: I guess no, it perfectly fits into experimental (that's for experiments with exchanges), plus leaving 'experimental' section alive with this setting will help fill it again with new experimental settings...

This setting can be left undocumented, probably...

freqtrade/configuration/configuration.py
```diff
@@ -75,6 +76,10 @@ def load_from_files(self, files: List[str]) -> Dict[str, Any]:
         # Normalize config
         if 'internals' not in config:
             config['internals'] = {}
+        # TODO: This can be deleted along with removal of deprecated
+        # experimental settings
+        if 'ask_strategy' not in config:
+            config['ask_strategy'] = {}
```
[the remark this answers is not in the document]
hroff-1902: Yes and now -- this should be left then, but not in this place and not in this way.

Well, we've got many checks against non-existing sections in the config, like those with `.get()`:
https://github.com/freqtrade/freqtrade/blob/caf415dc97df3cdb607a2ebef18bbc2851b980a8/freqtrade/exchange/exchange.py#L217-L218

Actually,I think this should be done once, for all sections in the config, that are not filled with the settings and do not exist. Somewhere in the Configuration, as a part of the normalization of the config at the early stage of config loading. So that all those checks could be simply eliminated in the code, to simplify the codebase...

But that's not for this PR.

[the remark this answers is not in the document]
xmatthias: well no.

`order_types` and `order_time_in_force` are good examples as to why this won't work.
We cannot normalize these in the configuration loading, because this is done as part of strategy loading.
(they need to be in either Configuration, or in the strategy).

If you normalize them as part of config loading, then StrategyResolver will assume they are set in config and will always ignore what is set as part of the strategy.
For that, they are validated in one place - in the exchange (they are only relevant when you use trading, which requires the exchange to be loaded).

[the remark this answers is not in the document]
hroff-1902: Let's go through the global normalization in a separate PR.

This was inserted here because `process_deprecated_setting` copies the value of the deprecated setting into this section, so it must exist.

In strategy resolver we'vr have:
```
        # make sure ask_strategy dict is available
        if 'ask_strategy' not in config:
            config['ask_strategy'] = {}
```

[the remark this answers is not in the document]
xmatthias: Well now THIS is what i was looking for! 

you're right, it's actually done in StrategyResolver. 
Keep as is then 👍

hroff-1902:
The only thing I've forgotten here is to look deeper at warnings via warnings.warn, why have not been they printed...

Issue #1059, opened 2018-07-22
Bot uses maker fees on marketable limit orders.


creslinux:
In trading a market maker is a limit order that is far outside of spot price.
Typically an exchange will have positive fee(they pay you) or reduced fee for adding liquidity to the market.

As our limit orders are on spot or close to market price they will incur taker fees not maker. 
The code should reflect this

Limit orders placed at spot price are not treated as maker, or limit order.
They're a 'marketable limit' order, which is different to a limit order, and incurs taker fees as any market order would.

To use limit, maker, fees the limit order should be placed early, left hanging,  far outside of spot.

  ```
 Fee is applied twice because we make a LIMIT_BUY and LIMIT_SELL
        fee = self.exchange.get_fee(symbol=pair, taker_or_maker='maker')
        trade = Trade(
            pair=pair,
            stake_amount=stake_amount,
            amount=amount,
            fee_open=fee,
            fee_close=fee,
            open_rate=buy_limit,
            open_rate_requested=buy_limit,
            open_date=datetime.utcnow(),
            exchange=self.exchange.id,
            open_order_id=order_id
        )
 
```

creslinux:
As example from binance.


> Taker: When you place an order that is immediately filled in its entirety (for example a market order) you are a “taker,” and you pay a “taker” fee for this. The idea is that you are “taking” the price you want, right now, generally by buying or selling limit orders sitting on the books.
> 
> Maker: When you place an order that doesn’t fill immediately (for example a limit order far from market price), you are a “maker,” and you can expect to pay a reduced “maker” fee for this. As noted above, this terminology comes from the fact that placing limit orders on order book helps to “make the market,” (it makes you a “market maker”).

creslinux:
And gdax.

> Maker: When you place an order which is not immediately matched by an existing order (for example, a limit order that takes a few hours or days to fill), that order is placed on the order book. Once that order sells or buys, that is once another customer places an order that matches yours, you are considered the maker. Your GDAX fee will be 0%.
> 
> Taker: When you place an order at the market price that gets filled immediately, you are considered a taker and will pay a fee (for GDAX that is 0.10% and 0.25% for BTC books and 0.10% and 0.30% for ETH books depending on your volume; the more you trade, the lower your fee is over time).
> 
> On partial fills: If an order is partially matched immediately, you pay a taker fee for that portion. The remainder of the order is placed on the order book and, when matched, is considered a maker order.

PR #2344, opened 2019-10-05
Backtest no fees / custom fees

### LINKED ISSUE #2223: support the configuration of slippage for backtest/hyperopts
idea :
As a trader
i want to configure a slippage when i test a strategy that deal with low volume or xxx/btc that have low decimals digits of precision 
so that the backtest look closer to live and the objective function take slippage into account for optimization

i can write test case if needed to exchange the how to.

have a nice day.

xmatthias:
## Summary
Allow specifying fee percentage to be applied during backtesting.

This can be usefull if you'r exchange account has some rabates applied (for high volume, for example), which ccxt does not detect (it'll always show the basic fees for each exchange).

This can also be "abused" to add slippage, by specifying a slightly higher value than what the exchange applies.
It can also be applied to test a "ideal" world, where no fees apply (although that is unrealistic).

closes #2223
Remotely related to #1059, which explains behaviour of different exchanges related to this ... this PR only applies to backtesting, so will not fix this issue.

## Quick changelog

- Add `--fee` parameter to backtest/hyperopt/edge commands

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -74,7 +74,7 @@ freqtrade backtesting --export trades --export-filename=backtest_samplestrategy.
 
 #### Supplying custom fee value
 
-Sometimes your account has certain fee rabates, which are not visible to ccxt.
+Sometimes your account has certain fee rebates (fee reductions starting with a certain account size or monthly volume), which are not visible to ccxt.
 To account for this in backtesting, you can use `--fee 0.001` to supply this value to backtesting.
 This fee must be a percentage, and will be applied twice (once for trade entry, and once for trade exit).
 
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -72,6 +72,17 @@ The exported trades can be used for [further analysis](#further-backtest-result-
 freqtrade backtesting --export trades --export-filename=backtest_samplestrategy.json
 ```
 
+#### Supplying custom fee value
+
+Sometimes your account has certain fee rabates, which are not visible to ccxt.
+To account for this in backtesting, you can use `--fee 0.001` to supply this value to backtesting.
+This fee must be a percentage, and will be applied twice (once for trade entry, and once for trade exit).
+
+```bash
+freqtrade backtesting --fee 0.001
+```
+
+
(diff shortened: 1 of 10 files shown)


hroff-1902:
if we rename it to, say, 'transactional cost' and specify that by default transactional cost value is set to the fee provided by exchange, we could then say that this value may be used to count other transactional cost components like slippage.

docs/backtesting.md
```diff
@@ -72,6 +72,17 @@ The exported trades can be used for [further analysis](#further-backtest-result-
 freqtrade backtesting --export trades --export-filename=backtest_samplestrategy.json
 ```
 
+#### Supplying custom fee value
+
+Sometimes your account has certain fee rabates, which are not visible to ccxt.
```
hroff-1902: rebates?

docs/bot-usage.md
```diff
@@ -168,22 +168,25 @@ Backtesting also uses the config specified via `-c/--config`.
...
+                        Specify ticker interval (`1m`, `5m`, `30m`, `1h`,
+                        `1d`).
   --timerange TIMERANGE
                         Specify what timerange of data to use.
-  --max_open_trades MAX_OPEN_TRADES
+  --max_open_trades INT
                         Specify max_open_trades to use.
   --stake_amount STAKE_AMOUNT
                         Specify stake_amount.
+  --fee FLOAT           Specify fee %. Should be in %, will be applied twice
```
hroff-1902: It's not procentage, it's ratio, like in stoploss value (1.0 corresponds to 100%, -0.001 here is -0.1%, not -0.001%)

The term 'procentage' is incorrectly used in the bot docs, comments and messages, where 'ratio' is to be used, but this is a different story.

No '%' sign should be used here at least, it's not amount of percents...

docs/backtesting.md
xmatthias: take binance as example...

once you reach a certain volume per month, you pay fewer fees (0.0020 instead of 0.0025). freqtrade doesn't account for this at the moment.

I tried to explain this in the docs now ...

PR #2354, opened 2019-10-08
remove underline from docs style

xmatthias:
## Summary
Followup to #2353

The underline looks odd as there are also "spaces" with underlines ...


xmatthias:
it was a last-minute addition (the problem appears only when you extendet the small window - it had `_` between each version which looked VERY odd...

PR #2356, opened 2019-10-08
[minor] Don't use os.path in test_history

xmatthias:
## Summary
Removing os.path from this file reduces the complexity - also we should not hardcode the path to ../testdata, but use the testdatadir fixture instead.


hroff-1902:
LGTM 👍

PR #2357, opened 2019-10-09
Minor: Fix helpstring for list-timeframes

hroff-1902:
Set different helpstrings for the `-1` option for list-exchanges and list-timeframes subcommands.

Currently the helpstring for list-timframes errorneously says about printing exchanges.

### DIFF
--- a/freqtrade/configuration/cli_options.py
+++ b/freqtrade/configuration/cli_options.py
@@ -245,7 +245,7 @@ AVAILABLE_CLI_OPTIONS = {
         default=constants.DEFAULT_HYPEROPT_LOSS,
     ),
     # List exchanges
-    "print_one_column": Arg(
+    "list_exchanges_print_one_column": Arg(
         '-1', '--one-column',
         help='Print exchanges in one column.',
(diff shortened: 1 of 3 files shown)


freqtrade/configuration/cli_options.py
```diff
@@ -255,6 +255,12 @@ def __init__(self, *args, **kwargs):
         help='Print all exchanges known to the ccxt library.',
         action='store_true',
     ),
+    # List timeframes
+    "list_timeframes_print_one_column": Arg(
+        '-1', '--one-column',
+        help='Print timeframes in one column.',
```
xmatthias: wouldn't "Show/print output in one column." also solve this without duplicating this?

You're running "list-timeframes" already, so do we really need to mention this again ... ?

hroff-1902: I made it this way initially, but did not like the common string (not this one) I printed there and decided that two different descriptions would be better.

hroff-1902:
This can be closed as repleced by #2364

PR #2360, opened 2019-10-10
Disable defaulting to DefaultHyperOpts and DefaultHyperOptLoss

hroff-1902:
Like it was done for DefaultStrategy.

* If you think that there should be defaulting to DefaultHyperOptLoss (we are going to probably exclude DefaultStrategy and DefaultHyperopt, but DefaultHyperOptLoss will still live) -- I can rollback this part.


xmatthias:
> If you think that there should be defaulting to DefaultHyperOptLoss

well removing DefaultHyperOptLoss works against your goal of making Hyperopt something quick and simple to use i think.

To understand which HyperoptLoss a user should use, she'll need to dig rather deep through the documentation, which does not reccomend one over the other, but just lists them (so without some deeper understanding, it's still going to be difficult).

Now i don't think DefaultHyperoptLoss is an ideal loss function (i guess we both agree on that) - but neither is `ProfitOnly`, since i think longer trade duration should be penalized too (but not as much as in DefaultHyperoptLoss).

## tl dr:
I personally would leave a default value for HyperoptLoss. 
There's only so many (3) options atm and an inexperienced user will have problems to get hyperopt running without the burden of selecting one of the Loss functions...

hroff-1902:
> > If you think that there should be defaulting to DefaultHyperOptLoss

Okay, I'll return it back here. I made it along with removal defaulting for DefaultHyperOpts, but then realized it's not very convenient...

xmatthias:
Additional question - how does this play along with #2231? 

I mean - if i use hyperopt with `--strategy coolStrategy --spaces stoploss` (in the future) - i don't need ANY hyperopt class - so it maybe should default to something which doesn't define much but allows loading buy/sell stuff from the strategy, and optimizes roi / stoplosses?

hroff-1902:
> Additional question - how does this play along with #2231?
> 
> I mean - if i use hyperopt with `--strategy coolStrategy --spaces stoploss` (in the future) - i don't need ANY hyperopt class - so it maybe should default to something which doesn't define much but allows loading buy/sell stuff from the strategy, and optimizes roi / stoplosses?

Let's address this separately, not in this PR. Maybe also in terms of the args_aftersubcommand2 branch.

hroff-1902:
> > > If you think that there should be defaulting to DefaultHyperOptLoss
> 
> Okay, I'll return it back here. I made it along with removal defaulting for DefaultHyperOpts, but then realized it's not very convenient...

Here it is, pls review the rest.

hroff-1902:
> > Additional question - how does this play along with #2231?
> > I mean - if i use hyperopt with `--strategy coolStrategy --spaces stoploss` (in the future) - i don't need ANY hyperopt class - so it maybe should default to something which doesn't define much but allows loading buy/sell stuff from the strategy, and optimizes roi / stoplosses?
> 
> Let's address this separately, not in this PR. Maybe also in terms of the args_aftersubcommand2 branch.

Tried to implement this -- definitely not in this PR...

xmatthias:
> Tried to implement this -- definitely not in this PR...

that wasn't my intend anyway.
just asking - since this may be counterproductive depending on how the ideas for that approach look like.

freqtrade/resolvers/hyperopt_resolver.py
```diff
@@ -20,17 +20,21 @@ class HyperOptResolver(IResolver):
     """
     This class contains all the logic to load custom hyperopt class
     """
-
     __slots__ = ['hyperopt']
 
-    def __init__(self, config: Dict) -> None:
+    def __init__(self, config: Dict = None) -> None:
```
xmatthias: Is there a case where HyperoptResolver can be called without configuration (apart from tests)? 

I mean - we call `HyperOptResolver()` exactly once from non-test code, which always uses `self.config` as parameter - so having a default here seems to be misleading (especially since calling this class without config will result in an `OperationalException` 2 lines below this)

hroff-1902: I also noticed this, but left for a while -- it's here for more than an year...

hroff-1902: I copied it from the `HyperOptLossResolver.__init__` which is below. This should probably be removed everywhere in the resolvers...

hroff-1902:
> > Tried to implement this -- definitely not in this PR...
> 
> that wasn't my intend anyway.
> just asking - since this may be counterproductive depending on how the ideas for that approach look like.

I guess, any of the hyperopt/strategy will then be optional and can then will be omitted. One of them (or both) shall be specified...

Depending on those two options and a set of methods both hyperopt and strategy have implemented in conjunction, the corresponding set of optimization spaces will be available...

freqtrade/resolvers/hyperopt_resolver.py
hroff-1902: Done

hroff-1902:
Done

Issue #2363, opened 2019-10-10
Hyperopt trailing stoploss hyperoptimization space


Theagainmen:
## My environment

* Operating system: Windows
* Python Version: Python 3.7.4
* Branch: freqtrade develop-7d1f66cc

### The suggestion
In hyperopt I currently can't find an easy way to hyperopt values like `trailing_stop_positive` or `trailing_stop_positive_offset`. Maybe make a way that you can easily add them like the `stoploss` value.

hroff-1902:
this is probably not easy to implement...

xmatthias:
at first look it looks rather straightforward...

it's another space - almost identical to stoploss...
could probably also be implemented into the stoploss_space() - which will make it a 8 line change (excluding tests and such).

``` python
    def stoploss_space() -> List[Dimension]:

        return [
            Real(-0.35, -0.02, name='stoploss'),
            Real(-0.35, -0.02, name='trailing_stoploss_positive'),
            Categorical([True, False], name='tsl_enabled'),
            Real(0, 0.1, name='trailing_stoploss_positive_offset'),  # (?)
            Categorical([True, False], name='tsl_positive_enabled'),
        ]
```

and in hyperopt.py:
``` python
        if self.has_space('stoploss'):
            self.backtesting.strategy.stoploss = params['stoploss']
            self.backtesting.strategy.trailing_stoploss_positive = params['trailing_stoploss_positive']
           ...

```

Now i didn't test this and it may not be THAT straightforward  ... but it's not that difficult either ...

hroff-1902:
should probably be a separate optimization space

Theagainmen:
As I was thinking about this, I think the pairs in your whitelist are also a BIG part of your tot profit, maybe a hyperopt function that checks your whitelist (or blacklist) and filters out the bad ones? But looking at the message above I think this one will require more programming and is not that easy to implement :3

hroff-1902:
Updated the title to reflect the content of the issue.

hroff-1902:
@Theagainmen pls test it (merged into current develop)

PR #2366, opened 2019-10-11
Interface options should not use config

xmatthias:
## Summary
We keep these options in both config and strategy. 
By using the values from Strategy, we enable hyperopting these values and speed up the code (no dictionary lookups, but attribute gets).

enables further work on #2363 

## Quick changelog

- use `strategy.*` instead of `config.get(*)`

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -361,8 +361,8 @@ class IStrategy(ABC):
 
             # Don't update stoploss if trailing_only_offset_is_reached is true.
             if not (self.trailing_only_offset_is_reached and high_profit < sl_offset):
+                # Specific handling for trailing_stop_positive
                 if self.trailing_stop_positive is not None and high_profit > sl_offset:
-                    # Ignore mypy error check in configuration that this is a float
                     stop_loss_value = self.trailing_stop_positive
                     logger.debug(f"{trade.pair} - Using positive stoploss: {stop_loss_value} "
                                  f"offset: {sl_offset:.4g} profit: {current_profit:.4f}%")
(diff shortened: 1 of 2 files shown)


freqtrade/strategy/interface.py
```diff
@@ -362,9 +362,9 @@ def stop_loss_reached(self, current_rate: float, trade: Trade,
             # Don't update stoploss if trailing_only_offset_is_reached is true.
             if not (self.trailing_only_offset_is_reached and high_profit < sl_offset):
                 # Specific handling for trailing_stop_positive
-                if 'trailing_stop_positive' in self.config and high_profit > sl_offset:
+                if 'trailing_stop_positive' in self.__dict__ and high_profit > sl_offset:
```
hroff-1902: Do we need this check? 
`if 'trailing_stop_positive' in self.__dict__`

It's a class attribute and this check will always return true...

xmatthias: nope - it's a class attribute which has no value assigned => does not exist at instance level unless it's overwritten.

I've tested this ... just removing this will break 3 or 4 tests :D 

The problem with `trailing_stop_positive` is that it's existence is significant - so if this is missing from both config and strategy, then it's disabled, otherwise that value will be used (also 0.0 is valid).

that's a design i don't particularly like, but see out of scope for this PR to fix ...

hroff-1902: I just hate this kind of 'code smells' in python, when simple things require using ugly looking code like access to `__dict__`.

Please review the changes I made to fix it.

xmatthias: Yes, this works as well.

Honestly, i didn't spend much time to find a better solution since as mentioned above, i think this setting should be explicit (`trailing_stop_positive_enabled` or similar), and i intend to change it to that very soon.

I've swapped the comments in these 2 lines - since the one you removed was correct (and makes sense) - but the one a line below does not (since this changed now).

PR #2369, opened 2019-10-13
Allow skipping of exchange validation

xmatthias:
## Summary
Utils should not run certain validations for the exchange.

There is no need to add another level of exchange classes - it's a simple "flag" mechanism.

replaces: #2368
Resolves #2367

## future improvements

If needed, this can be extended to allow disabling of single checks - so `download-data` can disable different checks than `list-timeframes`.

PR #2368, opened 2019-10-12
[WiP] Introduce BaseExchange class

hroff-1902:
Resolves #2367 

```
$ cp config_binance.json.example config.json
$ freqtrade list-timeframes --exchange gdax
2019-10-12 11:50:35,943 - freqtrade.loggers - INFO - Verbosity set to 0
2019-10-12 11:50:35,944 - freqtrade.configuration.configuration - INFO - Dry run is enabled
2019-10-12 11:50:35,944 - freqtrade.configuration.configuration - INFO - Using DB: "sqlite://"
2019-10-12 11:50:35,944 - freqtrade.configuration.configuration - INFO - Using max_open_trades: 3 ...
2019-10-12 11:50:35,944 - freqtrade.configuration.configuration - INFO - Using exchange gdax
2019-10-12 11:50:35,944 - freqtrade.configuration.configuration - INFO - Using user-data directory: /home/user/freqtrade-wrk/github-hroff-1902/freqtrade/user_data ...
2019-10-12 11:50:35,945 - freqtrade.configuration.configuration - INFO - Using data directory: /home/user/freqtrade-wrk/github-hroff-1902/freqtrade/user_data/data/gdax ...
2019-10-12 11:50:35,945 - freqtrade.configuration.check_exchange - INFO - Checking exchange...
2019-10-12 11:50:35,945 - freqtrade.configuration.check_exchange - WARNING - Exchange "gdax" is known to the the ccxt library, available for the bot, but not officially supported by the Freqtrade development team. It may work flawlessly (please report back) or have serious issues. Use it at your own discretion.
2019-10-12 11:50:35,945 - freqtrade.configuration.configuration - INFO - Using pairlist from configuration.
2019-10-12 11:50:35,945 - freqtrade.exchange.exchange - INFO - Applying additional ccxt config: {'enableRateLimit': True}
2019-10-12 11:50:35,948 - freqtrade.exchange.exchange - INFO - Applying additional ccxt config: {'enableRateLimit': True, 'rateLimit': 200}
2019-10-12 11:50:35,960 - freqtrade.exchange.exchange - INFO - Using Exchange "GDAX"
Timeframes available for the exchange `gdax`: 1m, 5m, 15m, 1h, 6h, 1d
```

TODO:
* [x] Fix exchange patching in conftest
* [x] Fix exchange tests
* [x] Add tests for BaseExchange

PR #2370, opened 2019-10-13
Add list-pairs and list-markets subcommands

hroff-1902:
The replacement for #1989 

It was easier to pick the changes and apply them in the new branch since the codebase in the develop has changed significantly.

Same functionality as in the #1989 

Possible usage for review:
* remove config.json (to avoid same issue with checking pairs as for list-timeframes)
* 
```
for i in `freqtrade list-exchanges -1`; do freqtrade list-markets --exchange $i; done
```
* or 
```
for i in `freqtrade list-exchanges -1`; do freqtrade list-pairs --exchange $i --print-list --active-only; done
```

TODO:

* [x] Some better handling of errors loading markets for some of the exchanges (thus 'don't merge')
* [x] Adjust docs
* [x] Add tests

TBD:
* maybe it's better to print only active pairs/markets by default and introduce the `-a/--all/--print-all` option instead of `--active-only`


freqtrade/exchange/exchange.py
```diff
@@ -932,3 +935,22 @@ def timeframe_to_next_date(timeframe: str, date: datetime = None) -> datetime:
...
+    return '/' in market.get('symbol', '')
+
+
+def market_is_active(market):
+    """
+    Return True if the market is active.
+    """
+    # "It's active, if the active flag isn't explicitly set to false. If it's missing or
+    # true then it's true. If it's undefined, then it's most likely true, but not 100% )"
+    # See https://github.com/ccxt/ccxt/issues/4874,
```
xmatthias: 🤣 😨

hroff-1902: Yeah... As an alternative, we could patch the 'active' field in the markets after ccxt to be it straight True/False as we need it...

hroff-1902: As another alternative, we can explicitly emphasize these pairs/exchanges with 'N/A' in the active column in the tabular output of this command instead of printing 'Yes'

xmatthias: we should not change ccxt results i think ... 
this makes sense as implemented - my comment was related to the the 2nd line of the comment ... 

> If it's undefined, then it's most likely true, but not 100%

hroff-1902: I did not get though the difference there between "If it's missing" and "If it's undefined" 🤣

hroff-1902:
* List is now human-readable (no quotes, the descriptive phrase with summary before the list, the dot at the end of the printed sentence)
* `--print-json` option added. Prints the list that can be copy-pasted into the pairs file and/or into the config.
* `--one-column` option added -- useful for shell scripting, iterating through the list of pairs in a shell script.

hroff-1902:
* `--base-currency` and `--quote-currency` renamed to `--base`, `--quote` and now accept lists of currencies:

Show all active pairs for USDT and USD stake (quote) currencies, as a human-readable list:
```
$ for i in `freqtrade list-exchanges -1`; do freqtrade list-pairs --exchange $i --quote USDT USD --print-list --active-only; done
```

Show all pairs with BTC and ETH currencies, including reversed pairs, in tabular format:
```
$ for i in `freqtrade list-exchanges -1`; do freqtrade list-pairs --exchange $i --base BTC ETH --quote BTC ETH; echo "$i; done
```

hroff-1902:
* By default, only active pairs/markets are now printed. This is more convenient. All pairs, including inactive, can be shown with the new `-a`/`--all` option. The `--active-only` option removed.

tests/exchange/test_exchange.py
```diff
@@ -1488,6 +1488,73 @@ def test_get_valid_pair_combination(default_conf, mocker, markets):
         ex.get_valid_pair_combination("NOPAIR", "ETH")
 
 
+@pytest.mark.parametrize("base_currencies, quote_currencies, pairs_only, active_only,"
```
hroff-1902: @xmatthias 
Could you pls take a look at this, I'm crazy about how to make flake happy here and what it wants...

freqtrade/configuration/arguments.py
```diff
@@ -147,6 +153,22 @@ def _build_subcommands(self) -> None:
         list_timeframes_cmd.set_defaults(func=start_list_timeframes)
         self._build_args(optionlist=ARGS_LIST_TIMEFRAMES, parser=list_timeframes_cmd)
 
+        # Add list-markets subcommand
+        list_markets_cmd = subparsers.add_parser(
+            'list-markets',
+            help='Print markets on exchange.'
+        )
+        list_markets_cmd.set_defaults(func=partial(start_list_markets, pairs_only=False))
+        self._build_args(optionlist=ARGS_LIST_PAIRS, parser=list_markets_cmd)
```
xmatthias: Based on your description in the docs, using `ARGS_LIST_PAIRS` for `list-markets` might be wrong.

`list-pairs` lists everything with / between (so ETH/BTC, ...).
`list-markets` lists all pairs - but how can it then filter on `--quote` or `--base ETH`? It does not have these 2 parts ... so these 2 flips should be left out - no?

I also don't see the difference in the 2 commands - markets (non-pairs) are not tradeable anyway, so why should we even list them ... ? 
What exchange do i need to use to see the difference? neither for bittrex, nor for binance i get different results.

freqtrade/utils.py
```diff
@@ -116,4 +122,82 @@ def start_list_timeframes(args: Dict[str, Any]) -> None:
...
+            elif args.get('list_pairs_print_json', False):
+                print(rapidjson.dumps(list(pairs.keys()), default=str))
+            elif args.get('print_csv', False):
+                writer = csv.DictWriter(sys.stdout, fieldnames=headers)
+                writer.writeheader()
+                writer.writerows(tabular_data)
+            else:
+                # print data as a table, with the human-readable summary
+                print(f"{summary_str}:")
+                print(tabulate(tabular_data, headers='keys', tablefmt='pipe'))
```
xmatthias: This part may be worth a refactor to it's own method in the future ... i can see some use in different "default" output formats ...

tests/test_utils.py
```diff
@@ -164,6 +164,241 @@ def test_list_timeframes(mocker, capsys):
...
+    api_mock.markets = markets
+    patch_exchange(mocker, api_mock=api_mock)
+
+    # Test with no --config
+    args = [
+        "list-markets",
+    ]
+    with pytest.raises(OperationalException,
+                       match=r"This command requires a configured exchange.*"):
+        start_list_markets(get_args(args), False)
```
xmatthias: This test is fluky and fails if `config.json` exists in the environment.

best have a look at `test_list_timeframes()` on how to mitigate this:

``` python
pargs = get_args(args)
pargs['config'] = None
```

tests/test_utils.py
```diff
@@ -164,6 +164,241 @@ def test_list_timeframes(mocker, capsys):
...
+
+#    # Test with --exchange
+#    args = [
+#        "list-markets",
+#        "--exchange", "binance"
+#    ]
+#    start_list_markets(get_args(args), False)
+#    captured = capsys.readouterr()
+#    assert re.match("Exchange Binance has 8 active markets:\n",
+#                    captured.out)
```
xmatthias: Is this intentionally commented?

freqtrade/utils.py
```diff
@@ -116,4 +122,82 @@ def start_list_timeframes(args: Dict[str, Any]) -> None:
...
+        summary_str = ((f"Exchange {exchange.name} has {len(pairs)} ") +
+                       ("active " if active_only else "") +
+                       (plural(len(pairs), "pair" if pairs_only else "market")) +
+                       (f" with {', '.join(base_currencies)} as base "
+                        f"{plural(len(base_currencies), 'currency', 'currencies')}"
+                        if base_currencies else "") +
+                       (" and" if base_currencies and quote_currencies else "") +
+                       (f" with {', '.join(quote_currencies)} as quote "
+                        f"{plural(len(quote_currencies), 'currency', 'currencies')}"
+                        if quote_currencies else ""))
```
xmatthias: :thinking:

docs/utils.md
```diff
@@ -54,3 +54,63 @@ Timeframes available for the exchange `binance`: 1m, 3m, 5m, 15m, 30m, 1h, 2h, 4
 ```
 $ for i in `freqtrade list-exchanges -1`; do freqtrade list-timeframes --exchange $i; done
 ```
+
+## List pairs/list markets
+
+The `list-pairs` and `list-markets` subcommands allow to see the pairs/markets available on exchange.
+
+Pairs are markets with the '/' character between the base currency part and the quote currency part.
```
xmatthias: Maybe also add a quick sample on which part is quote / and which is base? 
`<BASE>/<QUOTE>`
or "ETH/BTC - where ETH is the base currency, and BTC is the quote currency."... ?

freqtrade/configuration/arguments.py
hroff-1902: > What exchange do i need to use to see the difference? neither for bittrex, nor for binance i get different results.

Well, the following command:
```
$ for i in `freqtrade list-exchanges -1`; do echo "$i:"; freqtrade list-pairs --exchange $i --all --print-list | grep -o '[[:digit:]]* pairs'; freqtrade list-markets --exchange $i --all --print-list | grep -o '[[:digit:]]* markets'; done
```
gives us:
```
...
coinex:
227 pairs
338 markets
...
kraken:
99 pairs
110 markets
...
okex:
421 pairs
448 markets
okex3:
421 pairs
457 markets
^^ different for okex and okex3!!!
...
```

Also BitMEX, which is in the 'bad' exchanges and I think some others in this list.

freqtrade/utils.py
hroff-1902: I guess the generic `--print-format=FORMAT` option can be added in the future. The list of available formats differs for subcommands.

Then we'll be able to write in the helpstring:
`--print-json     Print the list of pairs in the JSON format. Equal to '--print-format=json'`

freqtrade/utils.py
hroff-1902: Yeah, fuzzy... It generates the human-readable summary, handles all plural/single forms of terms, adds 'active' qualifier when needed, etc.

Generation of human-readable strings, that are less regular than machine-readable formats, is always more complicated...

freqtrade/utils.py
xmatthias: as said - it's something for the future ... ;)

freqtrade/configuration/arguments.py
hroff-1902: As for the first part of the comment -- pairs are markets with '/' character. Other markets can be indexes, futures, etc. They also have 'quote' (quote currency, our stake currency for pairs traded by the bot) field properly filled by ccxt. The 'base' field is often strange, though (see `freqtrade list-markets --exchange coinex -a`, for example), but can be ordinal currencies... They also have price, can be interesting as informative pairs (not pairs, informative markets, to be precise). Some can also be traded (futures). In principle, we can also trade them in the future, if we get rid of checking pairs by parsing the symbol and compare the stake_currency with the value of the 'quote' field in the market structure...

When we'll move from pairs to markets (whitilisted/blacklisted markets, informative markets i.o pairs, etc), the list-pair subcommand can be deleted...

freqtrade/utils.py
```diff
@@ -116,4 +122,82 @@ def start_list_timeframes(args: Dict[str, Any]) -> None:
...
+                                 **({'Is pair': symbol_is_pair(v['symbol'])}
+                                    if not pairs_only else {})})
+
+        if (args.get('print_one_column', False) or
+                args.get('list_pairs_print_json', False) or
+                args.get('print_csv', False)):
+            logger.info(f"{summary_str}.")
+
+        if len(pairs):
+            if args.get('print_list', False):
```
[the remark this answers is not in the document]
hroff-1902: Yes, but this then should be added for other subcommands...
(I'll add this here now, the rest is in the next PRs)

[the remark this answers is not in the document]
hroff-1902: Empty line separator added also for tabular format

docs/utils.md
hroff-1902: Added, pls review.

tests/test_utils.py
hroff-1902: Uncommented it (slightly changed), it fails and I do not see the reason for it (does not fetch the exchange name). The option works fine in the subcommand when run standalone...

freqtrade/configuration/arguments.py
xmatthias: getting rid of that is a long shot... i've been thinking about this too - but there's 200 steps between now and that point ... 

but this explains it - thanks.

freqtrade/utils.py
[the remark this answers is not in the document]
xmatthias: just something i noticed for this PR - very often means that it can be improved for other things too :+1:

docs/utils.md
```diff
@@ -54,3 +54,71 @@ Timeframes available for the exchange `binance`: 1m, 3m, 5m, 15m, 30m, 1h, 2h, 4
...
 ```
+
+## List pairs/list markets
+
+The `list-pairs` and `list-markets` subcommands allow to see the pairs/markets available on exchange.
+
+Pairs are markets with the '/' character between the base currency part and the quote currency part in the market symbol.
+For example, in the 'ETH/BTC' pair 'ETH' is the base currency, while 'BTC' is the quote currency.
+For pairs traded by Freqtrade the pair quote currency is defined by the value of the `stake_currency` configuration setting,
+but using these commands you can print info about any pair/market, with any quote currency.
```
xmatthias: This is odd somehow ...

maybe split this sentence at the linebreak (after configuration setting) and rephrase the 2nd one to something along the lines of

"You can print info about any pair/market with these subcommands - and you can filter by quote-currency using `--quote BTC`, or by base-currency using `--base ETH`."

tests/test_utils.py
xmatthias: The exchange name is taken from the _api object, which is mocked ... so you need to change the mock (and change it back afterwards).


I've fixed this test now - have a look at my changes ... 

For simplicity, i used `patch_exchange()` again (and again) - but technically we could also extract  `mocker.patch('freqtrade.exchange.Exchange.name', PropertyMock(return_value=id.title()))` from that method and mock it directly.

hroff-1902: Thanks, but it;'s strange why does the similar test for list_timeframes works without this...

xmatthias: Because `list-timeframes` uses the value from the configuration (`{config['exchange']['name']}` as output, and this command uses `exchange.name` ...

hroff-1902: should this also be fixed in the same manner (or list-timeframes changed instead)?

xmatthias: well ... if we change list-timeframes to use exchange.name, then yes, we must change it in this way.
an alternative would to remove my fix for `list-pairs`, and use `config['exchange']['name']` instead ... 

Personally, i prefer using exchange.name, since it Capitalizes the name correctly - but i'm not sure if i would want to touch `list-timeframes` as part of this pr ...

hroff-1902: okay, got it (the difference, still asleep in the meantime), thanks 👍

docs/utils.md
hroff-1902: Fine, thanks 👍 

Pls review the result

hroff-1902:
> after playing some more with this, i think before merging we should decide on either #2369 or #2368 to avoid bad user experience.

Well, getting #2368 to the state when it can be applied to the develop and then to master requires more discussions, work and can take long time. In the same time, #2369 has some pitfalls that I'm going to discuss in terms of #2368, but it's okay as a current (temporary) solution 'to avoid bad user experience', right.

So my proposal is to apply #2369 currently, so that it does not stop this one, and then proceed on the more comprehensive solution in terms of #2369.

xmatthias:
> Well, getting #2368 to the state when it can be applied to the develop and then to master requires more discussions

in the form of an issue, please ;)

hroff-1902:
> > Well, getting #2368 to the state when it can be applied to the develop and then to master requires more discussions
> 
> in the form of an issue, please ;)

okay, no problem

hroff-1902:
#2369 merged into develop, then develop merged to here in order to allow `validate=False` for exchange in `start_list_markets`

#2369 seems to be non-working though, see my comment I added into that PR after merge.

hroff-1902:
...a test is needed to fail now when the exchange checks pairs/timeframe and the list-market/list-timeframs subcommand issues an error when timeframe/pairs is not on the exchange.

xmatthias:
> ...a test is needed to fail now when the exchange checks pairs/timeframe and the list-markets/list-timeframs subcommand issues an error when timeframe/pairs is not on the exchange.

if you look at #2411 - which fixes the issues you've found - i'm not sure what a test should test - since the problem was that the fallback exchange initialization was not passed the validation flag, not within  the exchange validation class itself.

Should've been found as part of #2369  ... we missed it - but i'm not sure what kind of test should correctly identify this.

hroff-1902:
Now seems to be working as it should, without 'bad user experience'.

Pls review commits between two latest merges...

tests/exchange/test_exchange.py
```diff
@@ -1693,6 +1695,74 @@ def test_get_valid_pair_combination(default_conf, mocker, markets):
...
+        (['LTC'], ['USDT'], True, False,
+         ['LTC/USDT']),
+        # all markets, base=LTC, quote=USDT, NONEXISTENT
+        (['LTC'], ['USDT', 'NONEXISTENT'], False, False,
+         ['LTC/USDT', 'XLTCUSDT']),
+        # all markets, base=LTC, quote=NONEXISTENT
+        (['LTC'], ['NONEXISTENT'], False, False,
+         []),
+    ])
+def test_get_markets(default_conf, mocker, markets,
```
xmatthias: :+1: great, concise test!

hroff-1902:
> LGTM 👍
> 
> After these tests i'm afraid you can't say that you are not good at writing tests anymore - sorry - you learned how to do that ... 😄
> 
> one request - since this is now ready - after merging this, could you please merge develop into #2397?
> In my experience this is best done by the author of the changes if possible, since the assumption is that the author will know best how to resolve conflicts ...

Agree; sure; done...

hroff-1902:
> LGTM 👍
> 
> After these tests i'm afraid you can't say that you are not good at writing tests anymore - sorry - you learned how to do that ... 😄

Oh, no, parametrization of the tests is the easiest part (especially if there are examples). All the mocking, patching and stuff -- that's the pain... I wrote these tests simply because there is another subcommand with tests, I could take and adapt. I would be writing this test for much longer time from scratches...

PR #2377, opened 2019-10-14
Rename --custom-hyperopt to --hyperopt

xmatthias:
## Summary
Align --hyperopt specification to other commands (don't use --custom

Solve the issue: https://github.com/freqtrade/freqtrade/pull/2264#issuecomment-540287648

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ jobs:
       name: backtest
     - script:
       - cp config.json.example config.json
-      - freqtrade hyperopt --datadir tests/testdata -e 5 --strategy DefaultStrategy --customhyperopt DefaultHyperOpts
+      - freqtrade hyperopt --datadir tests/testdata -e 5 --strategy DefaultStrategy --hyperopt DefaultHyperOpts
       name: hyperopt
     - script: flake8
       name: flake8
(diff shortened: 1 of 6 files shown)


hroff-1902:
I also do not like double `hyperopt --hyperopt`...

As an idea: `freqtrade hyperopt --hyperopt-class SampleHyperOpt --strategy-class SampleStrategy --hyperopt-loss-class DefaultHyperOptLoss`, these options may have the legacy `-s/--strategy`, `--hyperopt-loss` and the new `--hyperopt` aliases (the full form `--hyperopt-class`, `--strategy-class` is to be used in the docs), so the user could reduce the command line to `freqtrade hyperopt --hyperopt SampleHyperOpt --strategy SampleStrategy --hyperopt-loss DefaultHyperOptLoss` (and obtain double 'hyperopt' in the command), but this will be her choice...

xmatthias:
I'm not sure about `--strategy-class` and `--hyperopt-loss-class` ... isn't this just "changing" because of the change?

i mean, sure, we can reword every command - but to what point or extend? 
The helpstring indicates that it's expecting a class already - we're also not renaming `--enable-position-stacking` to `--enable-position-stacking-bool` to indicate that it's a boolean flag .. ? 

cli options should be as short as possible - beeing too verbose will make it difficult to type (correctly), and lead to frustration (not necessarily issues, but frustration because the command is wrong again).

hroff-1902:
`class` there is not the type of the entity (like `-bool` would be), but describes that the value is the name of the class and not the name of the module, containing it...

LGTM 👍 . at least it's much better than with legacy `custom`...

xmatthias:
let's be honest - using `freqtrade hyperopt --hyperopt Hyperopt` is an extreme example

`freqtrade hyperopt --config userconfig --strategy Whatever --hyperopt Hyperopt` is more realistic - and then it's not "that" odd ...

PR #2382, opened 2019-10-16
[minor] Iresolver small improvements

xmatthias:
## Summary

* Refactor search-path building to parent class
* return generator instead of next from `_get_valid_object()` (enables searching for all available objects in a future PR).


hroff-1902:
That's good...

LGTM! 👍

PR #2384, opened 2019-10-18
Improve buy timeout handling

### LINKED ISSUE #1858: Order cancelled AND fullfilled
Yes

  * Python Version: 3.6.7 (`python -V`)
  * CCXT version: 1.18.516 (`pip freeze | grep ccxt`)
  * Branch: Develop
  * Last Commit ID: 46b1ecc77d53054aed6835d69f9989316422aa78 (`git log --format="%H" -n 1`)
  
Exchange: Kraken
In configuration file: "stake_amount": "unlimited",

- Freqtrade makes a limit order

```
2019-05-16 10:00:05,129 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again..
2019-05-16 10:00:16,095 - freqtrade.freqtradebot - INFO - Buy signal found: about create a new trade with stake_amount: 19.7381 ...
2019-05-16 10:00:16,096 - freqtrade.freqtradebot - INFO - Using Last Ask / Last Price
2019-05-16 10:00:19,401 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': buy, 'exchange': 'Kraken', 'pair': 'BTC/EUR', 'limit': 7150.3, 'stake_amount': 19.7381, 'stake_currency': 'EUR', 'fiat_currency': 'EUR'}
2019-05-16 10:00:22,382 - freqtrade.wallets - INFO - Wallets synced.
```

- Kraken cancels the order

```
2019-05-16 10:00:25,471 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': status, 'status': 'Unfilled buy order for BTC/EUR canceled on Exchange'}
```

ISSUE:

- Kraken tells FT that the order is canceled

- but Kraken filled the order.
Here is the explaination of the behavior of kraken:
https://support.kraken.com/hc/en-us/articles/235162907-Order-Canceled-due-to-Out-of-Funds-

```
Order Canceled due to "Out of Funds":
It is possible for an order to partially execute (or "fill") if there are insufficient balances in the account for the order to fully execute.
This most commonly occurs when an order is placed for maximum volume. 
[...]
This is because the trading fee was not taken into consideration when placing the order.
```

- But FT has no active trade (cf. telegram /status -> 'no active trade') so it does not seem to be aware that the order were filled

Is that a normal way to act for FT or an issue ?
If freqtrade has no active trade, will it be able to sell the previously bought cryptos ? (perhaps I did not correctly understood the "active trade" definition)

To bypass this issuer, I configured the "stake_amount" parameters a little bit lower than the amount of available currency (currently testing).

Thank you
Xavier

### LINKED ISSUE #2383: WARNING: Unable to sell trade: Insufficient funds to create limit sell order on market
## Describe your environment

  * Operating system: Ubuntu 18.04.3 LTS
  * Python Version: Python 3.6.8
  * CCXT version: ccxt==1.18.1208
  * Branch: Develop
  * Last Commit ID: 06ab51b53d9c7b5bc43e0cdd03d704376aa71c26
  
## Describe the problem:

Sometimes a buy fails for the total amount, due to a timeout, for example:

Binance: Buying OMG/BTC
at rate 0.00009590
(0.050000 BTC,362.484 EUR)

A bit later:
Status: Remaining buy order for OMG/BTC cancelled due to timeout

A portion is bought, but in the database (trades) the total amount is written.

Trade ID: 26 (since 35 minutes ago)
Current Pair: OMG/BTC
Amount: 235.17 (0.0225528 BTC)
Open Rate: 0.00009590
Current Rate: 0.00009590
Current Profit: -0.20%
Stoploss: 0.00007193 (-25.00%)

The real amount bought in Binance is 234.93

So when it want to sell the items, there areinsufficient funds on Binance... Resulting in these kind of errors:

Unable to sell trade: Insufficient funds to create limit sell order on market OMG/BTC.Tried to sell amount 235.17 at rate xxx.Message: binance Account has insufficient...

### Observed Results:

  * What happened?
See Above

I expect that the trades table contains the updated amount of bought coins after a timeout to reflect the real amount so a sell cannot fail. (resulting in losses)

xmatthias:
## Summary
in case of partial buys which are cancelled due to timeout, we need to apply the "fee-fix" calculation to the amount as well, otherwise #2383 will happen.

This will need to use the return-value of the `cancel_order()` call (which is the same result - but with `status=canceled`). This makes sure that we use the latest information (at the time of canceling).
It can fall back to the result of the previous `get_order()` call if the order was cancelled already.

In addition, we should also use the same trade-update logic for orders that are cancelled on the exchange, since we cannot be sure that these orders are not partially filled (fixes #1858).

closes #2383
closes #1858

## Quick changelog

- don't raise OperationalException - this should only be used to stop the bot completely
- add fee-fix logic to `handle_timedout_limit_buy()`
- Use same logic for all cancelled buy-orders
- Require unfilledtimeout setting in configuration (otherwise cancelled orders are never handled!)
- don't require the telegram section in the configuration - it's not needed if it's switched off.

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -266,6 +266,6 @@ CONF_SCHEMA = {
         'stake_amount',
         'dry_run',
         'bid_strategy',
-        'telegram'
+        'unfilledtimeout',
     ]
 }
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -802,16 +802,26 @@ class FreqtradeBot:
         """Buy timeout - cancel order
         :return: True if order was fully cancelled
         """
-        self.exchange.cancel_order(trade.open_order_id, trade.pair)
-        if order['remaining'] == order['amount']:
+        corder = self.exchange.cancel_order(trade.open_order_id, trade.pair)
+        if corder['remaining'] == corder['amount']:
             # if trade is not partially completed, just delete the trade
             self.handle_buy_order_full_cancel(trade, "cancelled due to timeout")
             return True
 
         # if trade is partially complete, edit the stake details for the trade
         # and close the order
-        trade.amount = order['amount'] - order['remaining']
+        trade.amount = corder['amount'] - corder['remaining']
         trade.stake_amount = trade.amount * trade.open_rate
+        # verify if fees were taken from amount to avoid problems during selling
+        try:
+            new_amount = self.get_real_amount(trade, corder)
+            if not isclose(order['amount'], new_amount, abs_tol=constants.MATH_CLOSE_PREC):
+                trade.amount = new_amount
+                # Fee was applied, so set to 0
+                trade.fee_open = 0
+        except DependencyException as e:
+            logger.warning("Could not update trade amount: %s", e)
+
         trade.open_order_id = None
         logger.info('Partial buy order timeout for %s.', trade)
         self.rpc.send_msg({
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -11,7 +11,7 @@ from typing import Any, Dict, List, Optional, Tuple
 import arrow
 from requests.exceptions import RequestException
 
-from freqtrade import (DependencyException, OperationalException, InvalidOrderException,
+from freqtrade import (DependencyException, InvalidOrderException,
                        __version__, constants, persistence)
 from freqtrade.data.converter import order_book_to_dataframe
 from freqtrade.data.dataprovider import DataProvider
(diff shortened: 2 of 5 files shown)


freqtrade/constants.py
```diff
@@ -266,6 +266,6 @@
         'stake_amount',
         'dry_run',
         'bid_strategy',
-        'telegram'
+        'unfilledtimeout',
```
hroff-1902: This section is missed in configuration.md, btw...

xmatthias: no - line 58 and 59 (i had to check myself when i changed this) :D

samgermain:
What files do I have to change in my program to fix this error and what do I have to change the code from and to? I can't run the `setup.sh` file from your latest commit, so I'm stuck on an old version

samgermain:
Here's a commit from a day after these changes, https://github.com/freqtrade/freqtrade/tree/bedbd964fc12f9115c5942e1e6535770e144545b . You mostly just have to change the freqtradebot.py file, and there's a few other minor changes.

PR #2387, opened 2019-10-18
[minor] Fix/tests windows

xmatthias:
## Summary
Fix some minor things that prevent tests on windows from passing.

## Quick changelog

- use Path to convert paths to windows / linux styles
- add grace-period for one test relying on time, since on windows this becomes flaky

### DIFF
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -1449,8 +1449,8 @@ def test_tsl_on_exchange_compatible_with_edge(mocker, edge_conf, fee, caplog,
     # setting stoploss
     freqtrade.strategy.stoploss = -0.02
 
-    # setting stoploss_on_exchange_interval to 0 second
-    freqtrade.strategy.order_types['stoploss_on_exchange_interval'] = 0
+    # setting stoploss_on_exchange_interval to -1 second since this test runs fast
+    freqtrade.strategy.order_types['stoploss_on_exchange_interval'] = -1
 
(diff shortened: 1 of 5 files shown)


tests/test_freqtradebot.py
```diff
@@ -1449,8 +1449,8 @@ def test_tsl_on_exchange_compatible_with_edge(mocker, edge_conf, fee, caplog,
     # setting stoploss
     freqtrade.strategy.stoploss = -0.02
 
-    # setting stoploss_on_exchange_interval to 0 second
-    freqtrade.strategy.order_types['stoploss_on_exchange_interval'] = 0
+    # setting stoploss_on_exchange_interval to -1 second since this test runs fast
+    freqtrade.strategy.order_types['stoploss_on_exchange_interval'] = -1
```
hroff-1902: Better change `>` to `>=` here in `freqtradebot.py`:

```
if (datetime.utcnow() - trade.stoploss_last_update).total_seconds() > update_beat
```

It will be the more proper solution, since `>` waits one extra second (I mean, if `stoploss_on_exchange_interval` is set to 1, it will update when `.total_seconds()` is 2, when it should be true when `.total_seconds()` reaches 1 second).

xmatthias: well ideally (outside of tests) you should not set this lower than 60s (updating stoploss on exchange more often will probably result in a ban) - so 1 more/less second should not matter - but the solution you propose is cleaner than mine :+1:

PR #2388, opened 2019-10-18
Minor: No more hyperoptS

hroff-1902:
...Where it's not needed/incorrect (using 'hyperopts' i.o. correct 'hyperopt')

* DefaultHyperOpts renamed to DefaultHyperOpt
* some improvements in docs/hyperopt.md I found reviewing the doc file
* other cosmetics, mostly in comments/docs/strings


docs/bot-usage.md
```diff
@@ -106,7 +106,7 @@ user_data/
 ├── backtest_results
 ├── data
 ├── hyperopts
```
xmatthias: This seems kinda misaligned ...

maybe we should rename this folder too? (and the corresponding load-scripts ...) ?

hroff-1902: do not think so: `strategies` --> `hyperopts`

hroff-1902: `grep -ri hyperopt freqtrade tests docs'  -- there are many 'hyperopts' left (where it's convenient and correct)

PR #2391, opened 2019-10-19
[minor][plot] Plotting trades from database should show correct duration

xmatthias:
## Summary
Duration expects minutes - subtracting timestamps results in seconds - so we need to convert this to minutes.

Otherwise, the hover-element of sell trades shows wrong minutes (2400min instead of 40min).


hroff-1902:
Would it be useful to use `arrow.humanize()` instead here?

xmatthias:
maybe - but thats not the point of this PR.

also arrow needs to know if the source column is in seconds or minutes, and will want the same unit in both cases - since 4205 can be both minutes OR seconds.

What i don't like about arrow is that it'll show "immediately" for short timeframes, which is completely unsuitable for this.

Issue #2311, opened 2019-09-27
backtesting and timeranges - startup timerange


orehunt:
Signals that require X candles don't work as expected when timeranges are specified because the available data is limited to the timerange, both in timeframes and date notations (?)
This can cause wrong signals at the start of the timerange until enough data is available
a workaround is to add a guard to avoid signals if there aren't X ranges (but I think this effectly just prevents backtesting...)
This can skew hyperopt results since subsequent epochs might go the wrong way depending on how much the wrong signals affect the overall strategy

xmatthias:
yes, strategies will require startup time - but it's impossible to determine this automatically, since even if the column is called `ema20`, it could be a typo and contain `ema200` ...
It's an oddity when you try to exactly recreate a trading day - but normally not a problem for backtesting or hyperopt if the exact time of the first trade is not 100% important.

i thought we have something around this in the documentation too, but currently struggle to find it (maybe it was just me repeating it in the slack channel multiple times which makes me think this ...)

### To explain why it's not a problem

most common indicators that have startup time "don't" start up - so they remain empty (`NaN`) for the startup time  - making all comparisons to evaluate to False automatically, preventing buy signals.

very quick example:

``` python
dataframe['ema8'] = ta.EMA(dataframe, timeperiod=8)
dataframe['ema13'] = ta.EMA(dataframe, timeperiod=13)
dataframe['x'] = dataframe['ema8'] < dataframe['ema13']
dataframe['y'] = dataframe['ema8'] > dataframe['ema13']
dataframe['equal'] = dataframe['ema8'] == dataframe['ema13']

dataframe.head(20)
```
We're comparing for >, < and equality. 

![2019-09-27-203617_1116x756_scrot](https://user-images.githubusercontent.com/5024695/65793578-7a859780-e166-11e9-85cc-498e4b5c8cdc.png)

As you can see - until both ema8 and ema13 did "start up", not one element results to True (NaN != NaN). 
This means that you'll also have no buys in this timerange.

I would not know any indicator that works differently (it may depend on the implementation ...), but maybe you have an example of a common indicator that behaves differently?

To prove the NaN point:

``` python
import numpy as np
np.NaN == np.NaN

False
```

hroff-1902:
[...]

PR #2393, opened 2019-10-19
Remove non-date based timeframe selection

xmatthias:
## Summary
This will remove non-date based timeframe selection (e.g. `--timeframe -200`).
This has no practical usecase outside of tests, users should select a date-range instead.

There is no deprecation period for this, since the option itself is not removed - and i don't think anyone used this when backtesting.

preparation for work on #2311

## Quick changelog

- Update tests to use date-based timeranges instead
- remove non-date options from timerange parser
- remove non-date based options from history selection
- enable millisecond epoch timeselections (13 digit dates).

### DIFF
--- a/freqtrade/configuration/timerange.py
+++ b/freqtrade/configuration/timerange.py
@@ -42,6 +42,8 @@ class TimeRange:
                   (r'^-(\d{10})$', (None, 'date')),
                   (r'^(\d{10})-$', ('date', None)),
                   (r'^(\d{10})-(\d{10})$', ('date', 'date')),
+                  (r'^-(\d{13})$', (None, 'date')),
+                  (r'^(\d{13})-$', ('date', None)),
                   (r'^(\d{13})-(\d{13})$', ('date', 'date')),
                   ]
         for rex, stype in syntax:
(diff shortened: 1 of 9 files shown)


freqtrade/configuration/timerange.py
```diff
@@ -42,9 +42,8 @@ def parse_timerange(text: Optional[str]):
                   (r'^-(\d{10})$', (None, 'date')),
                   (r'^(\d{10})-$', ('date', None)),
                   (r'^(\d{10})-(\d{10})$', ('date', 'date')),
-                  (r'^(-\d+)$', (None, 'line')),
-                  (r'^(\d+)-$', ('line', None)),
-                  (r'^(\d+)-(\d+)$', ('index', 'index'))]
+                  (r'^(\d{13})-(\d{13})$', ('date', 'date')),
```
hroff-1902: Then it should also recognize the formats for 
`1231006505000-` and `-1233360000000`, besides `1231006505000-1233360000000`

xmatthias: arrrgh yes :D

PR #2412, opened 2019-10-23
[minor] Use exchange.name instead of config['exchange']['name']

xmatthias:
## Summary
Use exchange.name instead of configuration version - this makes the apearance similar across all utils (always using the capitalized name of the exchange).

We've discovered this in the `list-pairs` PR ... so we should be aligned across the utils i think.


hroff-1902:
LGTM 👍

PR #2418, opened 2019-10-25
Add logging to syslog and journald

hroff-1902:
`--logfilename journald` -- pass log messages to journald. Needs the `systemd` python package installed .

`--logfilename syslog:/dev/log` -- log to syslog (rsyslog), suitable for most systems
`--logfilename syslog` -- same, shortcut
`--logfilename syslog:/var/run/syslog` -- log to syslog (rsyslog), for MacOS
`--logfilename syslog:localhost:514` -- log to local syslog using UDP socket, if it listens on port 514.
`--logfilename syslog:<ip>:514` -- log to remote syslog at <ip> IP address and port 514. May be used on Windows for remote logging.

Messages are posted with the `user` facility.

How to see: 
* `tail -f /var/log/user`, or install graphical viewer (Log File Viewer for Ubuntu) for the syslog case;
* `journalctl -f` for journald case.

On many systems rsyslog (syslog) fetches data from journald, so any of `--logfilename syslog` or `--logfilename journald` can be used and the messages be viewed with both journalctl and log viewer.

For rsyslog the messages from the bot can be redirected into a separate dedicated log file.
Add
```
if $programname startswith "freqtrade" then -/var/log/freqtrade.log
```
to one of the rsyslog configuration files, for example at the end of the `/etc/rsyslog.d/50-default.conf`.

For syslog, the reduction mode can be switched on. This will reduce the number of repeating messages, for instance, multiple messages
```
Oct 25 09:00:44 host freqtrade.freqtradebot[4513]: 2019-10-25 09:00:44,364 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
Oct 25 09:00:49 host freqtrade.freqtradebot[4513]: 2019-10-25 09:00:49,368 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
Oct 25 09:00:54 host freqtrade.freqtradebot[4513]: 2019-10-25 09:00:54,373 - freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...
...
```
will be reduced to one:
```
Oct 25 09:48:14 host freqtrade.freqtradebot[4809]: message repeated 545 times: [ freqtrade.freqtradebot - INFO - Found no buy signals for whitelisted currencies. Trying again...]
```
To switch it on, set in `/etc/rsyslog.conf`:
```
# Filter duplicated messages
$RepeatedMsgReduction on
```

TODO:
* [x] docs
* [x] tests


xmatthias:
> Needs the systemd python package installed .

then i think this should be included in setup.py (we need to be careful though - as installing this package on windows does not work ...)

hroff-1902:
> > Needs the systemd python package installed .
> 
> then i think this should be included in setup.py (we need to be careful though - as installing this package on windows does not work ...)

Rather, I'm thinking now that sdnotify package is to be excluded from the dependencies... Should be installed by the user if this functionality is needed...

hroff-1902:
Docs added, for tests -- I have no idea how to unit test this.
Pls review.

docs/advanced-setup.md
```diff
@@ -31,3 +31,40 @@ as the watchdog.
 
 !!! Note
     The sd_notify communication between the bot and the systemd service manager will not work if the bot runs in a Docker container.
+
+## Logging to syslog or journald
+
+On many systems the bot can be configured to send its log messages to `syslog` or `journald`. The special values for the `--logfilename` option can be used for this:
```
xmatthias: On many *nix systems ... ?

docs/advanced-setup.md
```diff
@@ -31,3 +31,40 @@ as the watchdog.
...
+* `--logfilename syslog:/dev/log` -- log to syslog (rsyslog) using the `/dev/log` socket, suitable for most systems.
+* `--logfilename syslog` -- same as above, the shortcut for `/dev/log`.
+* `--logfilename syslog:/var/run/syslog` -- log to syslog (rsyslog) using the `/var/run/syslog` socket. Use this on MacOS.
+* `--logfilename syslog:localhost:514` -- log to local syslog using UDP socket, if it listens on port 514.
+* `--logfilename syslog:<ip>:514` -- log to remote syslog at IP address and port 514. This may be used on Windows for remote logging to external syslog server.
+
+Log messages are send to `journald` and `syslog` with the `user` facility. So you can see them with the following commands:
+
+* `tail -f /var/log/user`, or install a comprehansive graphical viewer (for instance, 'Log File Viewer' for Ubuntu) for the `syslog` case;
+* `journalctl -f` when logging to `journald`.
```
xmatthias: Personally i would add the systemd case here - where you can get "only" the bot logs by using `journalctl -f -u freqtrade.service` ... i suspect that that'll be the most common usecase for logging to journald / syslog...

docs/advanced-setup.md
```diff
@@ -31,3 +31,40 @@ as the watchdog.
 
 !!! Note
     The sd_notify communication between the bot and the systemd service manager will not work if the bot runs in a Docker container.
+
+## Logging to syslog or journald
+
+On many systems the bot can be configured to send its log messages to `syslog` or `journald`. The special values for the `--logfilename` option can be used for this:
+
+* `--logfilename journald` -- send log messages to `journald`. This needs the `systemd` python package installed as the dependency. Not available on Windows.
```
xmatthias: "... dependency, which is not available on Windows."

docs/installation.md
```diff
@@ -194,7 +194,7 @@ freqtrade -c config.json
 
 #### 7. (Optional) Post-installation Tasks
 
-On Linux, as an optional post-installation task, you can setup the bot to run as a `systemd` service. See [Advanced Post-installation Tasks](advanced-setup.md) for details.
+On Linux, as an optional post-installation task, you may wish to setup the bot to run as a `systemd` service or configure it to send the log messages to the `syslog`/`rsyslog` or `journald` daemons. See [Advanced Post-installation Tasks](advanced-setup.md) for details.
```
xmatthias: you may want to directlink this to the logging section by using the tag `advanced-setup.md#logging-to-syslog-or-journald` as link (be careful should you rename that title ... see main review comment)

xmatthias:
Re tests:
i've added the tests, but had to mark them as skip - since CI does not have systemd installed (and we can't assume this from all users).

Also, one test is "*nix" only (market accordingly) - since syslog tries to open the syslog connection - which on windows is no socket - and using a remote syslog server would fail if it doesn't exist.

docs/advanced-setup.md
hroff-1902: I meant the functionality itself (logging to journald) is not available on Windows, but the dependency as well...

docs/advanced-setup.md
hroff-1902: Linux is not Unix 😉

xmatthias: well then call it linux ... the point here is to be clear that it's not available for windows ...

Issue #2408, opened 2019-10-21
Log levels and bot monitoring (Question)


gaugau3000:
Hi,

As an algo trader
I want to monitore many bot instances easily
So that i can prevent losing money (bad behaviour) or make no trade

for example the message below comes every process_throttle_secs :
> - INFO - Found no buy signals for whitelisted currencies. Trying again...

This frequency is annoying when i try to read the logs. How you deal with   it ?

you use a log tool analyser like Kibana ?

Many tks.

xmatthias:
The easiest will be to change the log-level of that message to DEBUG ... and if i'm honest, i thought that we had the bot silenced completely some time ago.

the downside to this is that the bot will send no "i'm alive" anymore ... so looking at the logs of a rarely trading strategy will not tell you too much.

hroff-1902:
I also thought of it (verbosity levels) for long time.

Currently, the current verbosity level is 0. Verbosity level 1 adds DEBUG messages, verbosity level 3 adds printing content of network communication.

There is no difference between verbosity level 1 and 2. This can be utilized for this proposal, amongst probably other changes in logging:

* We can remove this info message at verbosity level 0
* Default verbosity level is to be set to 1 (i.e. keeping this message by default)
* Since selection of logging messages is performed on the logger (module) level, `throttle` method should be separated into different module, like freqtrade.worker.throttle, to control its logging easily.
* An additional `-q/--quiet` option be added, which decreases the verbosity level. Thus, `freqtrade -q` won't print this message
* Additionally, `-qq` (verbosity level -1, if we start counting from default verbosity level 1) may only allow WARNINGS and no INFO log messages, `-qqq` (verbosity level -2) -- only ERROR and CRITICAL
* Additionally, the values of the verbosity levels can be shifted to be like:

0 -- will stand for ERROR/CRITICAL, `-qqq`; 
1 -- WARNINGS, `-qq`; 
2 -- INFO without this noisy message (there can also be other differences in the future), `-q`; 
3 -- INFO (default); 
4 -- DEBUG, `-v`;
5 -- TBD reserved, `-vv`; may contain some extra messages, or vise versa level 4 may have some noisy DEBUG messages reduced
6 -- DEBUG plus network communication contents. `-vvv`

hroff-1902:
Plus for utilities the default level can be set to the decreased level, to print only WARNINGs or even ERRORs.

xmatthias:
i find this way too complicated ...

[...]

PR #2419, opened 2019-10-25
Change loglevel of repeated message to debug

xmatthias:
## Summary
Disable repeated log message from default output (makes reading the logs very difficult).

closes partially #2408 (closes it in combination with #2418).

## Quick changelog

- Switch log-level to Debug for repeated message.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -98,6 +98,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `strategy` | DefaultStrategy | Defines Strategy class to use.
 | `strategy_path` | null | Adds an additional strategy lookup path (must be a directory).
 | `internals.process_throttle_secs` | 5 | **Required.** Set the process throttle. Value in second.
+| `internals.keep_alive_interval` | 60 | Print keepalive message every X seconds. Set to 0 to disable keepalive messages.
 | `internals.sd_notify` | false | Enables use of the sd_notify protocol to tell systemd service manager about changes in the bot state and issue keep-alive pings. See [here](installation.md#7-optional-configure-freqtrade-as-a-systemd-service) for more details.
 | `logfile` | | Specify Logfile. Uses a rolling strategy of 10 files, with 1Mb per file.
 | `user_data_dir` | cwd()/user_data | Directory containing user data. Defaults to `./user_data/`.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -98,7 +98,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `strategy` | DefaultStrategy | Defines Strategy class to use.
 | `strategy_path` | null | Adds an additional strategy lookup path (must be a directory).
 | `internals.process_throttle_secs` | 5 | **Required.** Set the process throttle. Value in second.
-| `internals.keep_alive_interval` | 60 | Print keepalive message every X seconds. Set to 0 to disable keepalive messages.
+| `internals.heartbeat_interval` | 60 | Print heartbeat message every X seconds. Set to 0 to disable heartbeat messages.
 | `internals.sd_notify` | false | Enables use of the sd_notify protocol to tell systemd service manager about changes in the bot state and issue keep-alive pings. See [here](installation.md#7-optional-configure-freqtrade-as-a-systemd-service) for more details.
 | `logfile` | | Specify Logfile. Uses a rolling strategy of 10 files, with 1Mb per file.
 | `user_data_dir` | cwd()/user_data | Directory containing user data. Defaults to `./user_data/`.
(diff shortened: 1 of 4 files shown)


hroff-1902:
I'll be happy with this if we could additionally add a 'ping, still alive, I'm throttling again'-style message at the INFO level, which can be switched on by, for example, the `config.internals.print_alive_message` option, which is False by default, but can add this pinging info log message if requested...

hroff-1902:
👍 Super, thanks!

xmatthias:
ups - did not read your previous comment before "just now"  🤣 
so my naming may be slightly off ...

Alive is also not printed all the time, but just once per minute by default ... which i hope will make it ... Friendlier on the eyes ;)

hroff-1902:
> ups - did not read your previous comment before "just now" 🤣
> so my naming may be slightly off ...
> 
> Alive is also not printed all the time, but just once per minute by default ... which i hope will make it ... Friendlier on the eyes ;)

That's simply fine, 👍 
Some exchanges implement ping/pong messages (in the WS interface. mostly for "keepalive" purposes), so when ccxt will be able to use both REST and WS, we'll probably add here the 'ping exchange' logic here... Now that's fine for the purposes "to verify that the bot (its main thread) is still responsive".

hroff-1902:
Additionaly, Bitmex "Dead Man’s Switch" functionality can be inserted here in the future, So this even better than at each throttling iteration...

docs/configuration.md
```diff
@@ -98,6 +98,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `strategy` | DefaultStrategy | Defines Strategy class to use.
 | `strategy_path` | null | Adds an additional strategy lookup path (must be a directory).
 | `internals.process_throttle_secs` | 5 | **Required.** Set the process throttle. Value in second.
+| `internals.keep_alive_interval` | 60 | Print keepalive message every X seconds. Set to 0 to disable keepalive messages.
```
hroff-1902: * Let's set it to 0 by default:
```
`internals.keep_alive_interval` | 0 | Print keepalive message every N seconds. Set to 0 to disable keepalive messages. (Disabled by default.)
```
(the value in `config_full.json` can be kept 60, non-default)

* Actially, it's not a keepalive message, the keepalive is used to keep the connection. This has mostly another purpose -- to verify that the bot is still responsive. It's a "heartbeat" message.  So let's name it 'internals.heartbeat_interval'.
* The message should be more neutral, I propose: `Freqtrade heartbeat message.`, or, with PID (not always PIDs are logged, this will help distinguish bot instances): "Freqtrade heartbeat message. PID=NNN"

xmatthias: assuming a "regular" log output, we would have freqtrade repeated many times:

```
2019-10-25 19:53:22,105 - freqtrade.freqtradebot - INFO - I am alive.
2019-10-25 19:53:22,105 - freqtrade.freqtradebot - INFO - Freqtrade hearbeat message PID=xxx.
```

that's 3 times freqtrade in just one line ... which is... odd i think.
Apart from that - no issues in showing the PID.

hroff-1902: Okay, what about
```
2019-10-25 19:53:22,105 - freqtrade.freqtradebot - INFO - Hearbeat message, PID=xxx.
2019-10-25 19:53:26,105 - freqtrade.freqtradebot - INFO - Hearbeat message, PID=xxx.
...
```
?

freqtrade/freqtradebot.py
```diff
@@ -50,9 +51,9 @@ def __init__(self, config: Dict[str, Any]) -> None:
         # Init objects
         self.config = config
 
-        self._last_alive_msg = 0
+        self._heartbeat_msg = 0
 
-        self.keep_alive_interval = self.config.get('internals', {}).get('keep_alive_interval', 60)
+        self.hearbeat_interval = self.config.get('internals', {}).get('heartbeat_interval', 60)
```
hroff-1902: `heartbeat_interval`

freqtrade/freqtradebot.py
```diff
@@ -154,10 +155,10 @@ def process(self) -> None:
             self.check_handle_timedout()
             Trade.session.flush()
 
-        if (self.keep_alive_interval
-           and (arrow.utcnow().timestamp - self._last_alive_msg > self.keep_alive_interval)):
-            logger.info("I am alive.")
-            self._last_alive_msg = arrow.utcnow().timestamp
+        if (self.hearbeat_interval
```
hroff-1902: `heartbeat_interval`

freqtrade/freqtradebot.py
```diff
@@ -154,10 +155,10 @@ def process(self) -> None:
...
+            and (arrow.utcnow().timestamp - self._heartbeat_msg > self.hearbeat_interval)):
```
hroff-1902: `heartbeat_interval`

docs/configuration.md
xmatthias: i've taken "Bot heartbeat" now ... message seems ... redundant, it's a message since i can read it anyway ...

hroff-1902:
Excellent! 👍

gaugau3000:
I test it it work well !! 👍

PR #2425, opened 2019-10-26
docs: add a tip for The Ocean exchange

hroff-1902:
The Ocean requires web3 package installed.


hroff-1902:
> Question for the future ... should we maybe extract exchange-specific notes and settings to a seperate docs file?

I'm not sure in the moment, are there many of them?

xmatthias:
well no, not yet ... but it seems to get more and more ...

hroff-1902:
Maybe, in a sep. doc file. Some parts from other doc files may be moved to it, with the references left. For example, this tip is now in the configuration.md, but it actually not about configuration, but closer to installation issues.

👍

hroff-1902:
Feel free to create this extra doc file when you'll have more ideas about its content, I only support it. 👍

xmatthias:
it was just an idea for the future ... not something we'll require now ... (but it's good to have someone else think about it too ... so it's not completely forgotten) :D

PR #1712, opened 2019-03-29
Allow logging to file

xmatthias:
## Summary
Add logging to file as an option.
Also change standard-logging to stdout instead of stderr, to allow easier handling when piping the output to something else.

Solve the issue: multiple questions on Slack

## Quick changelog

- log to stdout
- additionally log to rotating file (10 files, 1M each).

PR #2438, opened 2019-10-28
Add option to log to stderr

hroff-1902:
A new option `-E`/`--stderr` introduced which directs printing of log messages into stderr instead of stdout.

Now it's possible to use machine-readable formats of output for utils:

* Produce .csv file with markets:
```
$ freqtrade -E list-markets --exchange okex --quote USD --print-csv >markets.csv
```

* Iterate markets in shell:
```
for p in `freqtrade -E list-markets -1 --exchange kraken --quote USD 2>/dev/null`; do
  echo "Do something for market symbol $p"
done
```

* Generation of pairs.json for multiple exchanges:
```
for e in binance bittrex; do freqtrade -E list-pairs --exchange $e --all --quote BTC --print-json >user_data/data/$e/pairs.json; done
```

TODO:

[ ] docs
[ ] tests (have no idea how to test it)

@xmatthias I'm ready to close this PR if logs will be changed to be printed to stderr by default, as an alternative, so this option becomes redundant. That would be even a better solution imho.

### DIFF
--- a/freqtrade/configuration/arguments.py
+++ b/freqtrade/configuration/arguments.py
@@ -9,7 +9,8 @@ from typing import Any, Dict, List, Optional
 from freqtrade import constants
 from freqtrade.configuration.cli_options import AVAILABLE_CLI_OPTIONS
 
-ARGS_COMMON = ["verbosity", "logfile", "version", "config", "datadir", "user_data_dir"]
+ARGS_COMMON = ["verbosity", "log_to_stderr", "logfile", "version", "config", "datadir",
+               "user_data_dir"]
 
 ARGS_STRATEGY = ["strategy", "strategy_path"]
 
--- a/freqtrade/configuration/cli_options.py
+++ b/freqtrade/configuration/cli_options.py
@@ -34,6 +34,12 @@ AVAILABLE_CLI_OPTIONS = {
         action='count',
         default=0,
     ),
+    "log_to_stderr": Arg(
+        '-E', '--stderr',
+        help='Force log messages to be printed to stderr instead of stdout.',
+        action='store_true',
+        default=False,
+    ),
     "logfile": Arg(
         '--logfile',
         help='Log to the file specified.',
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -125,6 +125,8 @@ class Configuration:
         # Log level
         config.update({'verbosity': self.args.get("verbosity", 0)})
 
+        config.update({'log_to_stderr': self.args.get("log_to_stderr", False)})
+
         if 'logfile' in self.args and self.args["logfile"]:
             config.update({'logfile': self.args["logfile"]})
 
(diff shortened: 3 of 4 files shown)


xmatthias:
logging to stdout was added intentionally in #1712 to allow easy logging to file using `freqtrade  >> logfile.log` or `freqtrade | grep something` - without the need to redirect stderr to stdout.

instead of an additional option to reestablish default behaviour - we could configure this based on the command ran? 
so utils log to stderr - allowing easy redirection ..., while allowing grepping for trade / optimize operations?

otherwise, i'm in favor of changing the default to stderr.

hroff-1902:
* many programs and utils log to stderr;
* stderr redirection `freqtrade 2>&1 >/dev/null | grep 'something'` is well-known for shell script writers and widely used. It should not raise any problem for anyone which is not looking at scripts first time...
* bash even allows 'shortcuts' in the form of `freqtrade 2> >(grep -v something 1>&2)`. I personally do not use this syntax, because I came from the world of sh/ksh, but it's also a well known bash constructions.
* if you need to grep without it, you can stream a copy of logs to a file with `--logfile` and then grep it . It's even possible to do on the fly, with `tail -f logfile | grep 'something'`
* having different subcommands make logging to different streams, stdout or stderr depending on the command, will produce bad user experience. Especially for non-novice users. I have quite serious objections against this approach.
* logging to stdout has only sense if the program does not output anything except logs. Even without util subcommands, for hyperopt and backtesting, it was wrong.
* with logs printed into stderr, a user may simply separate 'system' freqtrade log messages and her custom `print()` output from the strategy custom code, if she wants it.
* we could add a couple of recipes for stderr redirection into the FAQ section of the docs for novice users.

In short: let's simply issue logs into stderr. It's better than having this option.

xmatthias:
> having different subcommands make logging to different streams, stdout or stderr depending on the command, will produce bad user experience. Especially for non-novice users. I have quite serious objections against this approach.

I'm not for this (not at all) - just brought it up as an option.

I agree with your other points - personally i never saw an issue logging to stderr - so let's change it back to that (it's python's default anyway) - ideally with a few logging receipts as you mention.

hroff-1902:
* An additional argument for logging into stderr. Stderr is defined in POSIX as:
```
At program start-up, three streams shall be predefined and need not be opened explicitly: standard input (for reading conventional input), standard output (for writing conventional output), and standard error (**_for writing diagnostic output_**). When opened, the standard error stream **_is not fully buffered_**; the standard input and standard output streams are fully buffered if and only if the stream can be determined not to refer to an interactive device.
```
-- it defines stderr as a stream for "diagnostic output". Our log is diagnostic output.

More details in `man stderr` on Ubuntu:
```
The  stream  stderr is unbuffered.  The stream stdout is line-buffered when it points to a terminal.  Partial lines will not appear until fflush(3) or exit(3) is called, or a newline is printed.  This can produce unexpected results, especially with debugging output.
```

hroff-1902:
An example of program which also print diagnostics of different levels into stderr and every one just happy with this: `man Xorg` gives us
> -verbose [n]
Sets the verbosity level for information printed on stderr. If the n value isn't supplied, each occurrence of this option increments the verbosity level. When the n value is supplied, the verbosity level is set to that value. The default verbosity level is 0.

PR #2442, opened 2019-10-29
Pairlists enhanced filter options

### LINKED ISSUE #2445: Refactoring PairLists infrastructure
* Let's split existing VolumePairList into Volume part and Precision part. Two accurate different PairLists.
* That one from #2442 then is normally to be implemented also in a separate PairList.
* More PairLists will then easily follow (we have requests for PairLists limiting by max/min price), since they all will be small.
* Edge also looks like a PairList (it implements same interface, returning the list of pairs, but with different names of methods), But I would not like to touch it and convert into pairlist... 😃 Not at this stage.
* Then the `pairlist` section in the config should be a list of pairlists. We can allow a single Pairlist as it is now for compatibility. There are several options here how this particularly can look like, the main idea is to allow specifying any number of PairLists with corresponding parameters.
* Then the pairlists should work in a chain, as filters for the list of pairs, modifying the list of pairs passed from the previous one. As Edge now does, fetching the list of pairs after StaticPairlist, modifying it and returning the modified list of pairs further for use by the consuming code.
* I assumed that the chain of pairlists should start from the StaticPairList (implicitly added into the beginning of the list of pairlist filters). However, I know you don't like this idea, so they can simply start from the list of all pairs ignoring the pairs_whitelist setting. It's okay.

### LINKED ISSUE #2517: Pair BTC/USDT is not compatible with exchange Binance or contained in your blacklist.
* Operating system: ubuntu 16.04
  * Python Version: Python 3.6.8 (`python -V`)
  * CCXT version: 1.18.1260 (`pip freeze | grep ccxt`)
  * Branch: Develop
  * Last Commit ID: 01356b9d05b8ca9f589ac0b4bd9dea87907de985 (`git log --format="%H" -n 1`)
  
---------------------------------------
2019-11-12 09:15:18,945 - freqtrade.pairlist.IPairList - WARNING - Pair BTC/USDT is not compatible with exchange Binance or contained in your blacklist. Removing it from whitelist..
2019-11-12 09:15:18,945 - freqtrade.pairlist.IPairList - WARNING - Pair LTC/USDT is not compatible with exchange Binance or contained in your blacklist. Removing it from whitelist..
2019-11-12 09:15:18,945 - freqtrade.pairlist.IPairList - WARNING - Pair ETH/USDT is not compatible with exchange Binance or contained in your blacklist. Removing it from whitelist..
2019-11-12 09:15:18,945 - freqtrade.pairlist.IPairList - WARNING - Pair ETC/USDT is not compatible with exchange Binance or contained in your blacklist. Removing it from whitelist..
2019-11-12 09:15:18,946 - freqtrade.freqtradebot - INFO - Active pair whitelist is empty.

---------------------------------------
CONFIG
...
        "pair_whitelist": [
            "BTC/USDT",
            "LTC/USDT",
            "ETH/USDT",
            "ETC/USDT"
        ],
        "pair_blacklist": [
            "DOGE/BTC",
            "BNB/BTC",
            "BNB/USDT"
        ]
...

Why ? any ideas ?

xmatthias:
## Summary
Low priced pairs (also called shitoins) are risky to trade. 
In extreme examples, one unit on the exchange can be as much as 10% pricechange.
While this seems tempting to trade, it'll be a problem for stoplosses - and one lost trade is a down of at least 10%.

closes #2445 
closes #2517

#### Example HOT/BTC

it's currently pried at `0.00000011` - with a precision of 8.
assuming it moves to `0.00000012` (the minimum allowed pricechange) - then that's a 9% change.
The same happens the other way around. While this leaves barely enough space for a stoploss of 10% - it's still not a good pair to trade for most strategies.

This PR will provide an option to filter low-priced pairs in the `VolumePairList` based on the minimal change% allowed.

 `low_price_percent` needs to be > the minimum allowed 1% change percentage.

Now the curent precision_filter does filter HOT/BTC, but it does not filter `FUEL/BTC` nor `TFUEL/BTC` - which both have a 1 unit change % of > 2% (0.02).

## Quick changelog

- Add sequential Pairlist processing  (chaining)
- Remove accidental "random" sorting from `_validate_whitelist()`
- Add `PrecisionFilter` and `LowPriceFilter` as options which can be applied independently
- Each pairlist is passed to the next filter
- VolumePairlist can use pair_whitelist (place StaticPairList in front of that).
- Split errormessage between wrong stakecurency and not compatible with exchange.


hroff-1902:
It's a good enhancement, but pls see #2445 first...

(I started writing that here, but decided to move the points to a sep. issue for discussing.)

xmatthias:
I've adapted most of the ideas from #2445 here - based on my comment https://github.com/freqtrade/freqtrade/issues/2445#issuecomment-547794959

so this adds the "list of filters" as you said - but without the possibility to use multiple pairlists chained (since these are "sortings" to me (excluding StaticPairList ...) - not filters apart from the amount of coins).

hroff-1902:
if there are 2 filters -- volume (top 10 pairs by volume) and price (filter out pairs with low price), then applying them in the diff. order will produce diff. resulting list

that's the point

Ex.:
1) "price > N", then "top 10 pairs by volume" of them and 2) "top 10 pairs by volume" then "price > N" of them
-- produce different results.

xmatthias:
I've updated the first comment to reflect changes done to this - which allows now to put all pairlists in sequence.

A sideeffect is that VolumePairList can now filter from a "given" pair-whitelist - i don't see any usage of it (if i select pairs manually, i don't need volumePairList on top of that) - but @hroff-1902 sees this as the best way to handle pairlists.
This behaviour is NOT default, but documented.

hroff-1902:
Some comments/suggestions:

* I propose to rename LowPriceFilter to PriceFilter. This will allow (maybe later) add `max_price_*` attribute(s) into it, to filter out pairs with high prices (if someone wants to, and for completeness).

* `low_price_percent` is not really percent, by ratio. Should probably be renamed or, vice versa, accept percentage value.

* I propose to add absolute price limits: `low_price` (and `max_price` in the future). One of `low_price_percent` (`low_price_ratio`) and `low_price` can be specified.

* Actually, all variants `low_price_percent`, `low_price_ratio` and `low_price` can be implemented (maybe in the future, not in this PR), Only one of them can be configured. Same for `max_price_*` (in the future).

* `config` seems to be redundant. I propose to move the attributes from it one level up:
```
    "pairlists": [
        {"method": "StaticPairList"},
        {
            "method": "VolumePairList",
            "number_assets": 20,
            "sort_key": "quoteVolume",
            "ttl": 1800
        },
        {"method": "PrecisionFilter"},
        {"method": "LowPriceFilter", "low_price_percent": 0.01}
    ],
```

* `ttl` should probably be named `refresh_period` or `refresh_timeout`.

* I propose to remove StaticPairList, Currently, if it's configured in `pairlists`, it requires `pair_whitelist` to be set. Then this list is used as the starting point (starting list) for further filters. The opposite logic can be implemented: if `pair_whitelist` is set in the config, then it is used as the starting point (starting list) for all filters in `pairslists` (including VolumePairList, if it's configured). Otherwise, the list of all active pairs with stake_currency as the quote currency is used as the starting list. It will be easy to explain in the docs and to configure.

Tekpre012:
/freqtrade/freqtrade/pairlist/LowPriceFilter.py", line 16, in __init__
    self._low_price_percent = pairlistconfig.get('low_price_percent', 0)
AttributeError: 'NoneType' object has no attribute 'get'

xmatthias:
> I propose to add absolute price limits: low_price (and max_price in the future). Only one of low_price_percent (low_price_ratio) and low_price can be specified.

we can do this - but at a later point. (same with most other points regarding this filter).
I've renamed it to PriceFilter though - so we're open for the future.
Personally, I'm not certain about specifying absolute limits is not sustainable - assume price of the stake (assumign BTC) jumps up or down - then your absolute limit is wrong very soon... but that's not a topic for this PR i think.

> I propose to remove StaticPairList, 

I disagree - This would be very odd to handle in case of "no" pairlist defined.
We'd need the StaticPairList logic in pairlistmanager.
This logic is "not much" other than copying a list - but it would be inconsistent to have one type of pairlist implemented in the manager - in that case i could alsy say "let's remove Precisionfilter and move that to Pairlistmanager" - since every class may be using that - that's the point of having a list of pairlists - so we can chain them "at will".
All pairlists should be treated "equally" - so i think the current way to have StaticPairlist makes more sense.

Also, from a user standpoint, i find it much clearer to clearly define "use `StaticPairList`, then `VolumePairlist`" - rater than "use `VolumePairList`" - which may use pairlist - depending on that config. That's an implicit change of behaviour - i would try to avoid that if possible - even if adding an additional point as `StaticPairList` is more work - it is explicitly configured to be "Exactly this way".

It could also lead to very odd results - especially when used with multiple configs (first config has `VolumePairList` - last config adds pairlist, since that's my config defining the 'exchange' tree) - since the determining factor comes from a completely different branch of the configuration.


> /freqtrade/freqtrade/pairlist/LowPriceFilter.py", line 16, in init
self._low_price_percent = pairlistconfig.get('low_price_percent', 0)
AttributeError: 'NoneType' object has no attribute 'get'

Nice catch - this will be handled implicitly now by the change of moving configs to the same level as "method".

PR #2450, opened 2019-10-30
[docs] Update Raspberry documentation

xmatthias:
## Summary
Update documentation to install freqtrade on a raspberry.

The instruction was overly complex, and without hyperopt "regular" installation works just fine.

Also fixed some "box"-titles which were not rendering correctly.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -95,7 +95,7 @@ sudo apt-get update
 sudo apt-get install build-essential git
 ```
 
-#### Raspberry Pi / Raspbian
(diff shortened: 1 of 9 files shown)


docs/installation.md
```diff
@@ -97,27 +97,24 @@ sudo apt-get install build-essential git
 
 #### Raspberry Pi / Raspbian
```
hroff-1902: This should be moved 1 level up (so be with `###`)

It's a different case, not a part of 'normal' Linux installation instruction.

PR #2451, opened 2019-10-30
Bt analysis maxopen at any time

### LINKED ISSUE #2449: Strategy analysis -> how many max_open_trades i will use for this strategy ?
As an algo trade
I want to perform statistics like maximum drawdown, profit and find an optimum between stake_amount and max_open_trade so i run and export the backtest with --disable-max-market-positions flag and make simulations in jupyter.
So that i can find a comprise between risk and profit.

To archive that i need to find trades that happens at the same time and flag them (open_trade_number) and then filter then.

I don't know if it can be useful to have this straight in a column of the dataframe of load_backtest_data() ?

Maybe it is a too specific need.

Have a nice day.

xmatthias:
## Summary
Expose the algorithm to get max_open_trades at each candle to users.

Closes #2449

## Quick changelog

- Add `analyze_trade_parallelism()`
- Document `analyze_trade_parallelism()` as part of the juptyer notebook (Export of the notebook is in `strategy_analysis_example.md `).

### DIFF
--- a/docs/strategy_analysis_example.md
+++ b/docs/strategy_analysis_example.md
@@ -107,6 +107,22 @@ trades = load_trades_from_db("sqlite:///tradesv3.sqlite")
 trades.groupby("pair")["sell_reason"].value_counts()
 ```
 
+## Analyze the loaded trades for trade parallelism
+This can be useful to find the best `max_open_trades` parameter, when used with backtesting in conjunction with `--disable-max-market-positions`.
+
+`parallel_trade_analysis()` returns a timeseries dataframe with an "open_trades" column, specifying the number of open trades for each candle.
+
+
+```python
+from freqtrade.data.btanalysis import parallel_trade_analysis
+
+# Analyze the above
+parallel_trades = parallel_trade_analysis(trades, '5m')
+
+
+parallel_trades.plot()
+```
+
 ## Plot results
 
 Freqtrade offers interactive plotting capabilities based on plotly.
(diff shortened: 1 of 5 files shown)


docs/strategy_analysis_example.md
```diff
@@ -107,6 +107,22 @@ trades = load_trades_from_db("sqlite:///tradesv3.sqlite")
 trades.groupby("pair")["sell_reason"].value_counts()
 ```
 
+## Analyze the loaded trades for trade parallelism
+This can be useful to find the best `max_open_trades` parameter, when used with backtesting in conjunction with `--disable-max-market-positions`.
+
+`parallel_trade_analysis()` returns a timeseries dataframe with an "open_trades" column, specifying the number of open trades for each candle.
```
gaugau3000: seems like there is no analyze_trade_parallelism() you call it analyze_trade_parallelism() ?

xmatthias: nice catch - i renamed the function last-minute since i found `analyze_trade_parallelism()` to be easier to grasp ... but apparently forgot to swap it everywhere ...

hroff-1902:
Suggestion for a future enhancement: 

`parallel_trade_analysis().mean()`, `parallel_trade_analysis().max()` can be printed in the results of backtesting.

xmatthias:
Unless you're using `--disable-max-market-positions`, then `parallel_trade_analysis().max()` will always be `max_open_trades`...

mean() may make sense though...

hroff-1902:
> Unless you're using `--disable-max-market-positions`, then `parallel_trade_analysis().max()` will always be `max_open_trades`...

No, it will be `max_open_trades` or less (if `max_open_trades` is set big enough and not reached)... 😆 

Yes, this is mostly for `--disable-max-market-positions` or when `max_open_trades` is set to some bigger value...

PR #2452, opened 2019-10-30
Allow configuration of stoploss on exchange limit

### LINKED ISSUE #1717: Stoploss on exchange calculation
## Describe your environment

  * Python Version: Python 3.7.0 (`python -V`)
   * Branch:  Develop
  * Last Commit ID: 0d152eb90745241ad294ec65f7cbd52012f44305
  
## Describe the problem:

I think the calculation of stoploss_on_exchange should be changed, or at least a warning should be added.

Currently, we calculate the stop-price as `trade.open_rate * (1 + stoploss)`.
As limit (which is the rate the trade really sells) is then calculated by using `stop_price * 0.99`.

This leads to the trade risking 6% of my capital even though I only configured stoploss at a risk of 5%.

Real calculation example:

```
open_rate = 0.005272
stoploss = -0.05
stop_price = open_rate * (1 + stoploss)
print(f"stop_price: {stop_price}")
limit_price = stop_price * 0.99
print(f"limit_price: {limit_price}")
```

I think we should either increase the offset price (stop_price = stop_price * 1.01) - or use a lower multiplier than 1% for the limit.

xmatthias:
## Summary
"stoploss on exchange limit" should be configurable and not default to 1% without an easy way to modify this.

closes #1717

## Quick changelog

- add option within order_types to configure stoploss-limit

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -215,6 +215,11 @@ If this is configured, the following 4 values (`buy`, `sell`, `stoploss` and
 `emergencysell` is an optional value, which defaults to `market` and is used when creating stoploss on exchange orders fails.
 The below is the default which is used if this is not configured in either strategy or configuration file.
 
+Since `stoploss_on_exchange` uses limit orders, the exchange needs 2 prices, the stoploss_price and the Limit price.
+`stoploss` defines the stop-price - and limit should be slightly below this. This defaults to 0.99 / 1%.
+Calculation example: we bought the asset at 100$.
+Stop-price is 95$, then limit would be `95 * 0.99 = 94.05$` - so the stoploss will happen between 95$ and 94.05$.
+
 Syntax for Strategy:
 
 ```python
@@ -224,7 +229,8 @@ order_types = {
     "emergencysell": "market",
     "stoploss": "market",
     "stoploss_on_exchange": False,
-    "stoploss_on_exchange_interval": 60
+    "stoploss_on_exchange_interval": 60,
+    "stoploss_on_exchange_limit_ratio": 0.99,
 }
 ```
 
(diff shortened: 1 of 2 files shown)


hroff-1902:
should not this be named `stoploss_limit_ratio` instead and be applied here as well:
https://github.com/freqtrade/freqtrade/blob/ffed13b979f9fbd12f369ca3530585cfca964e05/freqtrade/freqtradebot.py#L638

xmatthias:
well - but this only applies to `stoploss_on_exchange` - "regular" stoploss orders are just sell-orders (most likely market - in the sample strategy)...

calling it `stoploss_limit_ratio` would imply it's applied to stoploss in general, but that's not the case (it uses regular limit-sell algorithms)

hroff-1902:
> well - but this only applies to `stoploss_on_exchange` - "regular" stoploss orders are just sell-orders (most likely market - in the sample strategy)...
> 
> calling it `stoploss_limit_ratio` would imply it's applied to stoploss in general, but that's not the case (it uses regular limit-sell algorithms)

Right, exactly, that's the point I talking about -- it's be applied to stoploss as well. And if we apply this first, will be needed to change settings once again, deprecating this one, which will cause bad user experience.

I propose to postpone this one, implement `stoploss_limit_ratio` for stoploss which is not on exchange and use this setting for stoploss on exchange here is well.

xmatthias:
> I propose to postpone this one, implement stoploss_limit_ratio for stoploss which is not on exchange and use this setting for stoploss on exchange here is well.

I'm not sure that makes a ton of sense ... (and it's not a question to postpone, but a question to change and implement this too ...).

I disagree though:
stoploss-sells are kind of "get me out of here" sells ... so i personally use market sells (if i don't or can't use `stoploss_on_exchange`).

If you want or have to to use limit sells (maybe because your exchange does not support market orders) - 
then what makes using the `stoploss_rate` (as is - e.g. 5% below entry) wrong? 
why should we use 6% on a stoploss that was set to 5%? 
Then you can also set stoploss to 6% right away, as that will give you the same result ... ?

hroff-1902:
I guess I understand -- if it's not stoploss on exchange, it's executed according to the rules set in the ask_strategy section of settings...

I think I was sleepy this morning, sorry.

PR #2456, opened 2019-10-31
preserve dataframe index count before populating signals

orehunt:
Saving the dataframe index before calling user functions ensures the correct timeframe is selected to fetch current signals.
A better alternative would be to select the timeframe by date.

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -238,6 +238,7 @@ class IStrategy(ABC):
             logger.warning('Empty ticker history for pair %s', pair)
             return False, False
 
+        latest_index = dataframe.index.max()
         try:
             dataframe = self._analyze_ticker_internal(dataframe, {'pair': pair})
         except ValueError as error:
@@ -259,7 +260,7 @@ class IStrategy(ABC):
             logger.warning('Empty dataframe for pair %s', pair)
             return False, False
 
-        latest = dataframe.iloc[-1]
+        latest = dataframe.iloc[latest_index]
 
         # Check if dataframe is out of date
         signal_date = arrow.get(latest['date'])


xmatthias:
I don't think this makes sense...

`.iloc[-1]` will get the last line of the dataframe. Unless the dataframe was sorted by something else, it will get the correct line.

Your fix does not prevent problems with sorting - if i sort differently and reset the index.

orehunt:
Yes this helps only when more rows are appended to the dataframe, the correct way would be to look up by time, and maybe print a warning in case the dataframe has a different length or has been downsampled..

xmatthias:
Honestly, i think it's better to address this via documentation.

Both things (returning the resapled dataframe and adding or removing lines from the dataframe) are problematic behaviours which can cause huge differences between backtesting and dry/live operations ... but are difficult to reliably detect.

Length of the dataframe is irrelevant if i drop 20 lines at the start, and append another 20 at the end ...

hroff-1902:
@orehunt could you give us an example of manipulations on the dataframe, when the current method fetches wrong row?

orehunt:
Before populating indicators I add this (assuming timeframe is 1h):
```python
bLen = d.index.max()
    for x in range(1, 31):
        d.loc[bLen+x, 'date'] = d.loc[bLen].date + Timedelta(hours=x)
```
So that If an indicator shifts into the future, the dataframe data is not lost

xmatthias:
I was thinking about a "strategy startup test" myself too, but i'm not yet sure how to approach that.

As an idea, it could feed a set of sample data to the strategy, and will want to get the same dataframe back (for the default columns, anyway), since users should not touch any of `open`, `high`, `low`, `close`, `date`, `volume` - and need to add `buy` and `sell` columns respectively.

Ideally (and this is where the difficulties lie) this would also give some feedback on what is causing the failure - maybe "don't use drop()" ... or whatever ... 
but that will probably be quite difficult...

hroff-1902:
My ideas:
* Having a similar additional sanity measure is fine. The dataframe can be extended by the user in the custom method, so we should better take some additional actions to fetch the proper row, not the latest one.
* However, the dataframe can also be reduced by the user -- for instance, some starting rows, with NaNs, can be removed, possibly unintentionally by some of the indicators/libraries used. So fetching the row by its number (position) seems also be quite incorrect.
* My proposal is similar, but the row with signals be selected by index (datatime): save the index (datatime) of the last row in the dataframe before passing it to the user's custom code and select then the buy/sell signals using the saved index value.

xmatthias:
> My proposal is similar, but the row with signals be selected by index (datetime): save the index (datetime) of the last row in the dataframe before passing it to the user's custom code and select then the buy/sell signals using the saved index value.

our dataframe is not a "real" timeseries dataframe (date is not the index), so this may be difficult (i've been thinking about changing this a while ago, but i'm not sure if it will impact indicators or not, so decided not to for that moment).

yes, we could use `dataframe.loc[dataframe['date'] == lastdate]` ... but that's not as nice as having the date part of the index ...

orehunt:
> > yes, we could use `dataframe.loc[dataframe['date'] == lastdate]` ... but that's not as nice as having the date part of the index ...

What benefits would canonically using a date type for indexing bring?

If there is the possibility that switching to timeseries would impact indicators then there could be a deprecation time where you compare non timeseries and timeseries dataframes and warn against miscrepancies...

As a side note checking the correct date _before_ an order is executed is also a good precaution in case the data is stale (the exchange did not update the candles or cdn problems...I don't know)

xmatthias:
> As a side note checking the correct date before an order is executed is also a good precaution in case the data is stale (the exchange did not update the candles or cdn problems...I don't know)

This check is in place, and we do check both date and ordering of the data - that's why on 1m candles, you sometimes see `Outdated history for pair xyz. Last tick is N minutes old`.
Ordering is not really relevant, since we can sort by date fairly easily if necessary (that's within the exchange class).

xmatthias:
I think having checks for strategy validity is a good idea - but i don't think using `index.max()` to select the last row is the right approach since that can fail easily.
assume a strategy added a row at the beginning of the dataframe - you'd now use a "old" candle to determine the sell signal since index.max() told you to use candle 499 - but now 499 is the 3rdlast candle - which is 3 hours ago on a hourly timeframe.


i've opened #2688 to discuss how we can validate the different ways a strategy can return wrong results, so further pr's will extend on this idea.

hroff-1902:
@xmatthias, updated your comment with correct link to the issue

xmatthias:
github doesn't like me typing a hash with a number - and tends to change the number each time i add a space after that :( 

thanks for fixing

hroff-1902:
back to the issue: do you think we shall not change the approach we use for fetching the signal?

`test-strategy` will verify it, but only once in some static conditions. so the dataframe can still be broken. So another measure against it can be applied. There are 3 ways:
* get `[-1]` (the last one) row, as now implemented;
* get `[len(original dataframe)]` as @orehunt proposed in this PR;
* get [date of last row in original dataframe], as in my proposal above...

Issue #1985, opened 2019-06-29
If Count is X, X-1 forcebuy wont work over telegram


Dante2333:
## Step 1: Have you search for this issue before posting it?

yes

## Step 2: Describe your environment

  * Python Version: 3.7
  * CCXT version: latest
  * Branch: Master
  * Last Commit ID: 2019.6
  
## Step 3: Describe the problem:

I'm trying to forcebuy the last remaining balance of my account, stake amount = "unlimited"; count is 4 of 5

### Steps to reproduce:

1. Say X is max amount of orders, if you have X-1 counts, you cant forcebuy last remaining, exchange is binance.
  
### Observed Results:

  * What happened?
It didnt buy after forcebuying, had to sell something then forcebuy again
  * What did you expect to happen?
that it worked properly and bought remaining amount.

### Relevant code exceptions or logs:

  ```
2019-06-29 15:48:44,757 - freqtrade.rpc.telegram - ERROR - Exception occurred within Telegram module
Traceback (most recent call last):
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/base/exchange.py", line 509, in fetch
    response.raise_for_status()
  File "/home/dante/.local/lib/python3.7/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.binance.com/api/v3/order

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/dante/freqtrade/freqtrade/exchange/exchange.py", line 370, in create_order
    amount, rate, params)
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/binance.py", line 694, in create_order
    response = getattr(self, method)(self.extend(order, params))
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/base/exchange.py", line 404, in inner
    return entry(_self, **inner_kwargs)
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/binance.py", line 1210, in request
    response = self.fetch2(path, api, method, params, headers, body)
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/base/exchange.py", line 429, in fetch2
    return self.fetch(request['url'], request['method'], request['headers'], request['body'])
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/base/exchange.py", line 521, in fetch
[...]

PR #2457, opened 2019-10-31
split up test_freqtradebot.py

xmatthias:
## Summary
Split up `test_freqtradebot.py` into `test_worker.py` and `test_integration.py` to be clearer what is tested.

test_integration shall grow in the future, adding test-cases which try to mimic the full cycle of trades or trading scenarios.

## Quick changelog

- Add test_worker.py
- add test_integration.py
- add test trying to reproduce #1985 (seems ok in test - needs more info from reporter... )


hroff-1902:
> Add test_worker.py 

-- did not we have it before? 🙄

xmatthias:
no we decided to keep tests for worker in `test_freqtrade.py` at the time ...

hroff-1902:
why do you call it 'integration'?

xmatthias:
> why do you call it 'integration'?

because i suck at naming things ... give me a better suited name and i'll jump on it without thinking twice...

it's kinda integration tests - as they don't unittests specific functions, but the interaction between different elements within freqtradebot.py.

to quote [wikipedia](https://en.wikipedia.org/wiki/Integration_testing):

> Integration testing [...] is the phase in software testing in which individual software modules are combined and tested as a group.

It's probably not real integration tests (i assume these should not mock anything ...?) ... but these come at least close (i hope).

hroff-1902:
Then in this case it's okay (it could probably also named `test_freqtradebot_integration.py`, but current name is okay), thanks for explanation.

PR #2460, opened 2019-11-01
[minor] Add new runmodes

xmatthias:
## Summary
We have 2 types of Utils - the ones that use exchange, and the ones that don't.

This info is relevant for certain configuration validations ... which should (or should not) happen for util commands.

It'll allow us to use `setup_utils_configuration()` for new utils without breaking in certain "config" cases.

While this may seem "random" ... it's extracted from a new cool util i'm working on ... where i faced EXACTLY this problem.

I can also keep this commit there ... but for clarity's sake i guess it's better to have it seperate.

Second part:
Also, certain startup messages don't make sense for utils ...
when running `list-timeframes`, i don't care about dry-run or database url ...
This tries to address this, but is probably by no means complete, so further updates to this may follow (silencing startup-messages for certain commands).

### DIFF
--- a/freqtrade/configuration/check_exchange.py
+++ b/freqtrade/configuration/check_exchange.py
@@ -21,7 +21,8 @@ def check_exchange(config: Dict[str, Any], check_for_bad: bool = True) -> bool:
              and thus is not known for the Freqtrade at all.
     """
 
-    if config['runmode'] in [RunMode.PLOT] and not config.get('exchange', {}).get('name'):
+    if (config['runmode'] in [RunMode.PLOT, RunMode.UTIL_NO_EXCHANGE]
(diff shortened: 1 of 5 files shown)


hroff-1902:
what about adding the different runmodes for each subcommand? (or simply use the name of the subcommand being executed as the value of the runmode)

xmatthias:
> what about adding the different runmodes for each subcommand? (or simply use the name of the subcommand being executed as the value of the runmode)

i had that - but having list-markets, list-pairs, list-... (whatever) is very inconvenient, since these new run-modes are mainly used to filter out certain messages and checks at startup.

all i care about from a "runmode" perspective for utils is "does it use exchange or not" ... since this determines if exchange needs to be validated or not.

PR #2397, opened 2019-10-20
require subcommand for all actions

xmatthias:
## Summary
Merging #2264 into develop.

Develop has been merged into that branch - and helpstrings have been adjusted to account for changes from develop.

## changes 
(mostly copied from #2264)
This PR tries to address this as follows:

* introduce `freqtrade trade` subcommand for trading in dry / live runs
* add --dry-run to force dry-runs
* Remove usage of `freqtrade ` without subcommand
* only accept arguments after subcommands (so it's aligned throughout the interface)
* use grouped parsers to group common arguments together (for visuals / help printing).
* removal of Defaultstrategy as default for `--strategy`
* removal of DefaultHyperopt as default for `--hyperopt`

Issue #2263, opened 2019-09-15
minimal_roi and stoploss in the hyperopt class


hroff-1902:
It's impossible to define minimal_roi and/or stoploss in the custom hyperopt class.

Usecase:
* A user created a new custom hyperopt class for a new strategy. She implemented `buy_strategy_generator()` and `sell_strategy_generator()` methods, defined spaces for hyperopting.
* She made hyperopting for the 'all' space.
* She created `populate_buy()` and `populate_space()` using the triggers and guards from the hyperopting best result.
* Then she decided to optimize minimal_roi and stoploss, so she runs hyperopting for the 'roi stoploss' spaces. This is also okay, hyperopting will print best results for both minimal_roi and stoploss.
* Now she decides to optimize stoploss more precisely, fixing the minimal_roi found in the previous step, by running hyperopt for the 'stoploss' space only. And it's impossible to do, since hyperopting does not fetch minimal_roi from the custom hyperopt class. The minimal_roi will be fetched from the default strategy, which does not have any sense for custom hyperopt and will produce completely different results while hyperopting for the 'stoploss' space.
* Same if she wants to fix 'stoploss' value found and hyperopt for minimal_roi.

xmatthias:
i'll be honest, i think that a solution to this may conflict with #2231 ...

Assuming i also forget to remove this from my custom hyperopt after correcting it in my strategy and run hyperopt with `--strategy MyAwesomeLosingStrategy` - which one should take prevelance - the content of the Strategy (`--strategy`) or the content of the custom hyperopt (`--customhyperopt`) file?

The documentation around htis is not clear - but i'm afraid that a technical solution may result in confusing behaviour....

hroff-1902:
year, all this should be clarified, (and should be clear enough to avoid confusions)
but I described the case when the user starts with the custom hyperopt class and she does not have any strategy yet. Fetching minimal_roi silently is simply wrong...

Btw, do you know that if you did not implement populate_buy/sell_trend() in the HyperOpt, the resolver loads it from DefaultStrategy (with a warning in the log)... Just found this behavior...

xmatthias:
well i implemented it ... so yes ... 
it's not ideal - but was required to allow optimizing only one side without loading a custom strategy.
Honestly, you'll still need to get this method from "somewhere" ... 

note: it's [documented](https://www.freqtrade.io/en/latest/hyperopt/#checklist-on-all-tasks-possibilities-in-hyperopt)

[...]

PR #2465, opened 2019-11-02
Hyperopt populate from strategy

xmatthias:
## Summary
Align SampleStrategy and SampleHyperopts to be good, combined samples by aligning indicators, and reusing the methods from SampleStrategy for SampleHyperopts.

working towards: #2231
should also help with: #2263
( closes none of the above completely)

## Quick changelog

- remove unneeded methods from SampleHyperopt
- Document how this works 
- change CI to use SampleHyperopt and Samplestrategy instead of DefaultHyperopt

### Notes:
will fully collide with #2397 - at least in the documentation and CI file - but we can't have one PR delay every other idea.

### DIFF
--- a/user_data/hyperopts/sample_hyperopt_advanced.py
+++ b/user_data/hyperopts/sample_hyperopt_advanced.py
@@ -37,6 +37,9 @@ class AdvancedSampleHyperOpts(IHyperOpt):
     """
     @staticmethod
     def populate_indicators(dataframe: DataFrame, metadata: dict) -> DataFrame:
+        """
+        This method can also be loaded from the strategy, if it doesn't exist in the hyperopt class.
+        """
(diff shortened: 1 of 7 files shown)


user_data/hyperopts/sample_hyperopt_advanced.py
```diff
@@ -229,8 +232,9 @@ def stoploss_space() -> List[Dimension]:
 
     def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
         """
-        Based on TA indicators. Should be a copy of from strategy
-        must align to populate_indicators in this file
+        Based on TA indicators.
+        Can be a copy of from the strategy, or will be loaded from the strategy.
```
hroff-1902: "copy of corresponding method from"
below as well

PR #2476, opened 2019-11-05
Introduce remove_credentials to remove code duplication

### LINKED ISSUE #2474: bibox fetchcurrencies error
Hello,
I am testing other exchanges and saw the following error:

2019-11-05 12:32:07,463 - freqtrade.exchange.exchange - WARNING - Unable to initialize markets. Reason: bibox fetchCurrencies is an authenticated endpoint, therefore it requires 'apiKey' and 'secret' credentials. If you don't need currency details, set exchange.has['fetchCurrencies'] = False before calling its methods.
2019-11-05 12:32:07,463 - freqtrade - ERROR - Fatal exception!

Seems like bibox requires auth via API before fetching currencies.
Can we add an option in config file like requireauth (true/false) so code can pass it ?

xmatthias:
## Summary
we should only have one place to remove credentials from the configuration.
This will reduce code duplication, and will allow easy circumvention should this be needed for some tests.

closes #2474 

## Quick changelog

- have one place to remove credentials (api keys , ...).

### DIFF
--- a/freqtrade/configuration/__init__.py
+++ b/freqtrade/configuration/__init__.py
@@ -1,4 +1,5 @@
 from freqtrade.configuration.arguments import Arguments  # noqa: F401
+from freqtrade.configuration.check_exchange import check_exchange, remove_credentials  # noqa: F401
 from freqtrade.configuration.timerange import TimeRange  # noqa: F401
 from freqtrade.configuration.configuration import Configuration  # noqa: F401
 from freqtrade.configuration.config_validation import validate_config_consistency  # noqa: F401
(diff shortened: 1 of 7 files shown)


hroff-1902:
Well, I would not say it fixes #2474 (it does not), but the PR is good, I also thought of a single point for adjusting the secrets for dry-run/utils...

So LGTM 👍 , approve

(how to reproduce #2474: `freqtrade list-pairs --exchange bibox`)

xmatthias:
> (how to reproduce #2474: freqtrade list-pairs --exchange bibox)

i am well aware how to reproduce it - but i don't have an account there, so cannot verify if it would work if i disable this... :)

hroff-1902:
> > (how to reproduce #2474: freqtrade list-pairs --exchange bibox)
> 
> i am well aware how to reproduce it - but i don't have an account there, so cannot verify if it would work if i disable this... :)

me to (I also do not have an account)

It does not say that it will work without credentials, it says that it needs credentials for `fetchCurrencies()`...

PR #2488, opened 2019-11-06
[minor] Fix UnicodeError in hyperopt output

xmatthias:
## Summary
Depending on the system locale settings, we can encounter UnicodeErrors during hyperopt output. This is caused by the Sigma (sum sign) - which cannot be converted to some non-utf8 encodings.

To fix this, we can convert the data to the users configured locale - and back.

On "supported" systems (utf-8 locale) - the output will be as usual 

```
*    1/5:     72 trades. Avg profit  0.29%. Total profit  0.01064443 BTC (  21.24Σ%). Avg duration  69.6 mins. Objective: 1.86128
```

On unsupported systems, the "non-displayable" sign is converted to a questionmark

```
*    1/5:     72 trades. Avg profit  0.29%. Total profit  0.01064443 BTC (  21.24?%). Avg duration  69.6 mins. Objective: 1.86128
```

This is visible [here](https://github.com/xmatthias/freqtrade/commit/e02e970ba0348e88eaa53e83e8d193dac96d9ba4/checks?check_suite_id=298731672) - click on "hyperopt" and scroll all the way down.

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -4,9 +4,9 @@
 This module contains the hyperopt logic
 """
 
+import locale
 import logging
 import sys
-
 from collections import OrderedDict
 from operator import itemgetter
 from pathlib import Path
@@ -14,10 +14,10 @@ from pprint import pprint
 from typing import Any, Dict, List, Optional
 
 import rapidjson
-
-from colorama import init as colorama_init
 from colorama import Fore, Style
-from joblib import Parallel, delayed, dump, load, wrap_non_picklable_objects, cpu_count
+from colorama import init as colorama_init
+from joblib import (Parallel, cpu_count, delayed, dump, load,
+                    wrap_non_picklable_objects)
 from pandas import DataFrame
 from skopt import Optimizer
 from skopt.space import Dimension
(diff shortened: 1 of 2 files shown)


hroff-1902:
Tried to check it with python 3,8, unsuccessful -- scipy cannot be installed...

hroff-1902:
Finally checked in python 3.8 and `export LANG=C`
-- it does not fail even without this change, with current develop. Does not also fail with this PR.

Anyway, LGTM 👍 this PR is fine for python 3.6, 3.7

xmatthias:
> As far as I heard, this was changed in python 3.8 (explicit encode/decode is no longer needed),

i heared this for the change between python2 and python3 ... that's now ~10 years old and yet i'm looking at this error ... 

it may not be needed in 3.8 - but as long as it's needed in 3.6 and we support it (there is no reason to drop support since it'll limit the audience  - unless we start to use features 3.6 can't do anymore, but i don't know any features i'd want to use yet)

hroff-1902:
> it may not be needed in 3.8 - but as long as it's needed in 3.6 and we support it (there is no reason to drop support since it'll limit the audience - unless we start to use features 3.6 can't do anymore, but i don't know any features i'd want to use yet)

Sure

PR #2494, opened 2019-11-08
Fix UTC handling of timestamp() conversation in fetch_my_trades

### LINKED ISSUE #2490: binance sell

If you have discovered a bug in the bot, please [search our issue tracker](https://github.com/freqtrade/freqtrade/issues?q=is%3Aissue). 
If it hasn't been reported, please create a new issue.

  * Operating system: linux ubuntu 18.04
  * Python Version: Python 3.6.8
  * CCXT version: ccxt==1.19.14
  * Branch: Develop
  * Last Commit ID: ad2289c34cc938bc8b529f2676e71555fb216faa 

The order amount in DB is different from the actual amount in my binance account. I'm new to freqtrade, please help me and thanks!
  
### Observed Results:

  * What happened?
freqtrade tried to buy EOS/BTC with amount of 26.46 but after the order filled, the actual amount is 26.43 so I have to manually change the amount in DB.

`2019-11-07 07:21:12,992 - freqtrade.freqtradebot - INFO - Found open order for Trade(id=1, pair=EOS/BTC, amount=26.46202699, open_rate=0.00037790, open_since=2019-11-07 12:20:54)`
`2019-11-07 07:21:13,703 - freqtrade.freqtradebot - INFO - Applying fee on amount for Trade(id=1, pair=EOS/BTC, amount=26.46202699, open_rate=0.00037790, open_since=2019-11-07 12:20:54) failed: myTrade-Dict empty found`
`2019-11-07 07:21:13,704 - freqtrade.persistence - INFO - Updating trade (id=1) ...`
`2019-11-07 07:21:13,704 - freqtrade.persistence - INFO - LIMIT_BUY has been fulfilled for Trade(id=1, pair=EOS/BTC, amount=26.46000000, open_rate=0.00037800, open_since=2019-11-07 12:20:54).`

xmatthias:
## Summary
trade.open_date and trade.close_date are in UTC timezone. After querying from the database, we get them as tz-native objects back - so calling `.timestamp()` on them results in different results depending on the timezone.
For UTC+ offsets this is not a problem, since we get trades for the last X hours - however UTC- offsets (Americas ...) will encounter update poblems as the "since" parameter sent to the exchange is in the future.

We can easily fix this by replacing the timezone with the utc-timezone (we don't convert to timestamp very often).

Closes #2490

## Quick changelog

- check all occurences of `.timestamp()`
- fix bug
- Test (for some reason my pc think's it's in Americas - and i'm not ...)

### DIFF
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -7,7 +7,7 @@ from typing import Dict
 
 import numpy as np
 import pandas as pd
-import pytz
+from datetime import timezone
 
 from freqtrade import persistence
 from freqtrade.misc import json_load
@@ -106,8 +106,8 @@ def load_trades_from_db(db_url: str) -> pd.DataFrame:
                "stop_loss", "initial_stop_loss", "strategy", "ticker_interval"]
 
     trades = pd.DataFrame([(t.pair,
-                            t.open_date.replace(tzinfo=pytz.UTC),
-                            t.close_date.replace(tzinfo=pytz.UTC) if t.close_date else None,
+                            t.open_date.replace(tzinfo=timezone.utc),
+                            t.close_date.replace(tzinfo=timezone.utc) if t.close_date else None,
                             t.calc_profit(), t.calc_profit_percent(),
                             t.open_rate, t.close_rate, t.amount,
                             (round((t.close_date.timestamp() - t.open_date.timestamp()) / 60, 2)
--- a/freqtrade/data/history.py
+++ b/freqtrade/data/history.py
@@ -9,12 +9,11 @@ Includes:
 import logging
 import operator
 from copy import deepcopy
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
 import arrow
-import pytz
 from pandas import DataFrame
 
 from freqtrade import OperationalException, misc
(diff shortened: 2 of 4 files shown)


xmatthias:
> Have not you forrgotten to paste somewhere else the helpstring you deleted in the second commit?

No - i moved it in the first commit - but forgot to save the `test_exchange.py` file before committing ...

> Why did you get rid of pytz? What's the exact rationale behind it? Just because it's redundant and all functionality needed is contained in the datetime package?

a) we don't directly depend on pytz - it just happens to be there due to downstream dependencies
b) the UTC timezone will not change. pytz is great if you need to correctly localize things - since it adapts to "short term" changes (like CET / CEST maybe going away next year for certain countries). We only ever used it to "force" timestamps to utc - so its great functionality makes no sense, as `pytz.UTC` and `timezone.utc` have the same functionality.

Now personally I don't care if we use timezone.utc or pytz.UTC (in which case we should add it as a dependency though) - but it should be only one of the 2, not a mix.
We used both ways throughout the codebase, so i simply removed "one of them" - using the one that's part of the standardlibrary ...

hroff-1902:
should be removed from the dependencies then as well?

xmatthias:
> Should pytz be removed from the dependencies then as well?

never been there in freqtrade in the first place - not as direct (requirements.txt / setup.py) requirement ... (that's one reason for me to remove usage).

It'll most likely still need to be installed though, as i expect other libraries will use it and would break should you uninstall it.

PR #2495, opened 2019-11-08
Prepare reorg install doc : explain differences between master and develop

gaugau3000:
Hi,

- I have try to explain the differences between master and develop.
- I have delete the `git checkout develop` as we are already in this branch when cloning so we tell the end user to do something that has no impact and so confusing.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -37,6 +37,7 @@ cd freqtrade
 ./setup.sh --install
 ```
 
+    When cloning the repository the default working branch is name `develop`. This branch contains the last features (can be considered as relatively stable thanks to automated tests). The `master` branch contains the code of the last release (done once per month with a one week old snapshot of the `develop` branch to prevent packaging bugs so potentially more stable).
 !!! Note
     Windows installation is explained [here](#windows).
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -34,10 +34,10 @@ Freqtrade provides a Linux/MacOS script to install all dependencies and help you
 ```bash
 git clone git@github.com:freqtrade/freqtrade.git
 cd freqtrade
-git checkout develop
 ./setup.sh --install
 ```
 
+    When cloning the repository the default working branch is name `develop`. This branch contains the last features (can be considered as relatively stable thanks to automated tests). The `master` branch contains the code of the last release (done once per month with a one week old snapshot of the `develop` branch to prevent packaging bugs so potentially more stable).
```
xmatthias: I guess we should add rendered Note box on top of this?

```
!!! Note "Version considerations"
    When ...
```

(Note that the following indent MUST be 4 spaces to be rendered correctly (i fixed a few appearances with 2 spaces which were broken recently)

gaugau3000: oups forget the stage of this lines in this editor :-)

gaugau3000: I have take your title it is a good idea to not leave with default 'Note'

PR #2503, opened 2019-11-09
Fixes strategy config from file logic

jraviotta:
`Configuration.from_files` was clobbering `config['strategy']` in the arg parsing logic. This PR fixes the bug that overwrites the strategy in config.

ln 353 is a pep8 indentation formatting change.

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -152,9 +152,9 @@ class Configuration:
 
         self._process_logging_options(config)
 
-        # Set strategy if not specified in config and or if it's non default
-        if self.args.get("strategy") != constants.DEFAULT_STRATEGY or not config.get('strategy'):
-            config.update({'strategy': self.args.get("strategy")})
+        if ("strategy" in self.args and self.args["strategy"] and
+                self.args['strategy'] != constants.DEFAULT_STRATEGY):
+            config.update({'strategy': self.args['strategy']})
 
         self._args_to_config(config, argname='strategy_path',
                              logstring='Using additional Strategy lookup path: {}')
@@ -350,7 +350,7 @@ class Configuration:
                         configuration instead of the content)
         """
         if (argname in self.args and self.args[argname] is not None
-           and self.args[argname] is not False):
+                and self.args[argname] is not False):
 
             config.update({argname: self.args[argname]})
             if logfun:


freqtrade/configuration/configuration.py
```diff
@@ -152,9 +152,9 @@ def _process_common_options(self, config: Dict[str, Any]) -> None:
 
         self._process_logging_options(config)
 
-        # Set strategy if not specified in config and or if it's non default
-        if self.args.get("strategy") != constants.DEFAULT_STRATEGY or not config.get('strategy'):
-            config.update({'strategy': self.args.get("strategy")})
+        if ("strategy" in self.args and self.args["strategy"] and
+                self.args['strategy'] != constants.DEFAULT_STRATEGY):
```
xmatthias: What is the exact case this is fixing?

At "quick look" the 2 seem to do the same thing, the old using the dict's `.get()` method, and the right is doing explicit checking for the keys...

hroff-1902: I also did not get this.

jraviotta: Any strategy set in `config.json` is overwritten by `args.get("strategy")`
```python
if self.args.get("strategy") != constants.DEFAULT_STRATEGY or not config.get('strategy'):
```
If True or anything -> True
because  
`args.get('strategy') != DEFAULT_STRATEGY` -> True
config is updated with `args.get("strategy")`


In `config.json` include an entry for strategy, `'strategy': 'multiRSI'`.

Add debugging output to `freqtrade/configuration/configuration.py`
```python
    def load_config(self) -> Dict[str, Any]:
        """
        Extract information for sys.argv and load the bot configuration
        :return: Configuration dictionary
        """
        # Load all configs
        config: Dict[str, Any] = self.load_from_files(self.args["config"])

        # Keep a copy of the original configuration file
        config['original_config'] = deepcopy(config)

        self._process_runmode(config)
        logger.info(f"self.args.get('strategy') != constants.DEFAULT_STRATEGY or not config.get('strategy')\n{self.args.get('strategy') != constants.DEFAULT_STRATEGY or not config.get('strategy')}")
        logger.info(f"self.args.get('strategy'): {self.args.get('strategy')}")
        logger.info(f"Just before _process_common_options(config)\nconfig.get('strategy'): {config.get('strategy')}")
        self._process_common_options(config)
        logger.info(f"Just after _process_common_options(config)\nconfig.get('strategy'): {config.get('strategy')}")
       etc
```

execute
```python
config_files = [str(Path.home().joinpath('.freqtrade/config.json'))]
config = Configuration.from_files(config_files)
```
output
```python
2019-11-12 22:25:35,417 - freqtrade.configuration.configuration - INFO - Using config: /home/jona/.freqtrade/config.json ...
2019-11-12 22:25:35,418 - freqtrade.configuration.configuration - INFO - Validating configuration ...
2019-11-12 22:25:35,421 - freqtrade.configuration.configuration - INFO - self.args.get('strategy') != constants.DEFAULT_STRATEGY or not config.get('strategy')
True
2019-11-12 22:25:35,421 - freqtrade.configuration.configuration - INFO - self.args.get('strategy'): None
2019-11-12 22:25:35,421 - freqtrade.configuration.configuration - INFO - Just before _process_common_options(config)
config.get('strategy'): MultiRSI
2019-11-12 22:25:35,422 - freqtrade.configuration.configuration - INFO - Just after _process_common_options(config)
config.get('strategy'): None
```

Swap the logic
```python
 if ("strategy" in self.args and self.args["strategy"] and
                self.args['strategy'] != constants.DEFAULT_STRATEGY):
```
'strategy' in args AND args['strategy'] AND != DEFAULT_STRATEGY
If True AND True AND True -> overwrite with arg
else use strategy set in config with fallback to DEFAULT_STRATEGY

output
```python
2019-11-12 22:27:39,960 - freqtrade.configuration.configuration - INFO - Using config: /home/jona/.freqtrade/config.json ...
2019-11-12 22:27:39,962 - freqtrade.configuration.configuration - INFO - Validating configuration ...
2019-11-12 22:27:39,964 - freqtrade.configuration.configuration - INFO - self.args.get('strategy') != constants.DEFAULT_STRATEGY or not config.get('strategy')
True
2019-11-12 22:27:39,965 - freqtrade.configuration.configuration - INFO - self.args.get('strategy'): None
2019-11-12 22:27:39,965 - freqtrade.configuration.configuration - INFO - Just before _process_common_options(config)
config.get('strategy'): MultiRSI
2019-11-12 22:27:39,966 - freqtrade.loggers - INFO - Verbosity set to 0
2019-11-12 22:27:39,966 - freqtrade.configuration.configuration - INFO - Just after _process_common_options(config)
config.get('strategy'): MultiRSI
```

xmatthias: You are right, this is a problem noone detected so far.

However, this problem was adressed in #2397 (merged just now ...) "accidentally" - since we removed the default from the `--strategy` argument, which made this change (almost identical to yours) necessary anyway.

I've tested your above code yesterday, and you're right, it did not work on develop - however it does now (just tested again).

As this fix is unnecessary with the latest develop version, i don't think this makes sense anymore, and will therefore close it. 

Note: #2397 was in the works for months - so we've been reviewing and testing that change a while now - otherwise your change would make perfect sense!

jraviotta: Excellent. Glad to hear the issue will be fixed.

PR #2505, opened 2019-11-11
Add restricted markets snippet to documentation

xmatthias:
## Summary
add snippet to get restricted pairs from bittrex to exchange.md
more info:
https://github.com/freqtrade/freqtrade/issues/2074#issuecomment-552307926

Solve the issue: #2074

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -20,7 +20,7 @@ Binance has been split into 3, and users must use the correct ccxt exchange ID f
 * [binance.us](https://www.binance.us/) US based users- ccxt id: `binanceus`
 * [binance.je](https://www.binance.je/) Trading FIAT currencies - ccxt id: `binanceje`
 
-### Kraken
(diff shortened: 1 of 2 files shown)


docs/exchanges.md
```diff
@@ -33,3 +33,25 @@ To download data for the Kraken exchange, using `--dl-trades` is mandatory, othe
...
 ```
+
+## Bittrex
+
+### Restricted markets
+
+Bittrex split its exchange into US and International versions.
+The International version has more pairs available, however the API always returns all pairs, so there is currently no automated way to detect if you're affected by the restriction.
+
+If you have restricted pairs in your whitelist, you'll get a warning message in the log on FreqTrade startup for each restricted pair.
```
hroff-1902: "RESTRICTED_MARKET" should be explicitly mentioned here, to let the user a clue which particular message should she look for.

PR #2518, opened 2019-11-12
Move freqtrade CI to github actions

xmatthias:
## Summary
This will move github actions from TravisCI to github actions.

There are multiple reasons for this:
* Travis recently failed often with "infrastructure" errors - where gcc was not available, or networking did simply fail.
* The future of Travis for Opensource is unclear ever since they've been aquired last year, and let go most of their team.
* Travis does not allow jobs to run for longer than 50 minutes. Building the Raspberry image takes ~1h - so is not possible with travis.

I took this opportunity to also rework how the CI works - building on multiple operating systems instead of just ubuntu. This would have been possible with travis too but noone ever took the time to do so.

Closes #2473 

Note: Deploy does run for this branch  (github_actions_tests) explicitly - allowing us to test all necessary steps.

## Quick changelog

- Add github action CI to this repository
- Add windows builds (seperate build-job - could maybe be combined into the regular job, but requires a lot of "if: windows" ... - since windows is too different from *nix).
- Add macos builds
- Build raspberry image for docker hub (postfixed with `_pi` - so tags will be `:master_pi` and `develop_pi`).
- Build images for develop and master on push
- run Ci only for selected branches (otherwise it'll run twice when pushing to local branches which have a PR)

For now i intend to keep Travis to compare the two for a while (however the deploy step has been moved to GithubActions).

Now i'm sure there will be further changes to this moving forward as we discover additional things, but i think this should give us a good base.

I'm not certain if the mentions for failed builds for @here should happen or not #notifications channel.

## What's new?

Notifications:

2019-11-12-104140_560x261_scrot

Github actions UI

2019-11-12-104209_1217x604_scrot

PR #2519, opened 2019-11-12
Update dockerhub description from github readme.md

xmatthias:
## Summary
Update dockerhub description from the github readme automatically

supplement to #2518 

## Quick changelog

- based on master (there can only be one description on dockerhub, and it should be the latest).

### DIFF
--- a/.github/workflows/docker_update_readme.yml
+++ b/.github/workflows/docker_update_readme.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - name: Docker Hub Description
-      uses: peter-evans/dockerhub-description@v2.0.0
+      uses: peter-evans/dockerhub-description@v2.1.0
       env:
         DOCKERHUB_USERNAME: ${{ secrets.DOCKER_USERNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker_update_readme.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches:
+    - master
+
+jobs:
+  dockerHubDescription:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Docker Hub Description
+      uses: peter-evans/dockerhub-description@v2.0.0
```
hroff-1902: `peter-evans/dockerhub-description` says it's v2.1.0 now...

xmatthias: huh? this has nothing to do with the raspberry build ... 

this action ONLY updates the dockerhub description - so runtime is ... irrelevant, really (it runs in  a few seconds).

xmatthias: nevertheless - updating the action makes sense.

xmatthias:
nope, does not matter - 2 different files .. :)

PR #2521, opened 2019-11-12
Add fiat to status table

### LINKED ISSUE #464: Please add new column in /status table to show USD profit or loss
Please add new column in /status table to show USD profit or loss

xmatthias:
## Summary
Cleanup of `/status table` and addition of fiat currency.

Closes  #464

- removes need to import DataFrame in rpc areas
- Cleanup of rpc (Decimal conversation is not needed here)
- add fiat-currency output to status_table.

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -3,16 +3,15 @@ This module contains class to define a RPC communications
 """
 import logging
 from abc import abstractmethod
-from datetime import timedelta, datetime, date
+from datetime import date, datetime, timedelta
 from decimal import Decimal
 from enum import Enum
-from typing import Dict, Any, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import arrow
-from numpy import mean, NAN
-from pandas import DataFrame
+from numpy import NAN, mean
 
-from freqtrade import TemporaryError, DependencyException
+from freqtrade import DependencyException, TemporaryError
 from freqtrade.misc import shorten_date
 from freqtrade.persistence import Trade
 from freqtrade.rpc.fiat_convert import CryptoToFiatConverter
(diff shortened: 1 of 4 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -135,12 +134,11 @@ def _rpc_status_table(self) -> DataFrame:
                     trade.pair,
                     shorten_date(arrow.get(trade.open_date).humanize(only_distance=True)),
                     f'{trade_perc:.2f}%'
+
```
hroff-1902: remove it? (empty line)

freqtrade/rpc/rpc.py
```diff
@@ -135,12 +134,11 @@ def _rpc_status_table(self) -> DataFrame:
...
                 ])
 
             columns = ['ID', 'Pair', 'Since', 'Profit']
-            df_statuses = DataFrame.from_records(trades_list, columns=columns)
-            df_statuses = df_statuses.set_index(columns[0])
-            return df_statuses
+            return trades_list, columns
```
hroff-1902: why have you changed it from returning the dataframe to this tuple? Due to performance or something else?

xmatthias: if you look at where this is used - then converting this to a DataFrame makes no sense.

The output of this function is then passed to tabulate - with "columns" as column headers ...
the same can be accomplished by keeping this as list - and passing the headers back - removing the overhead of creating (and uncreating) a Dataframe.

I did not measure the performance - but just thinking about it - it's avoiding a unnecessary conversion to dataframes.

freqtrade/rpc/rpc.py
xmatthias: it kindof is? (not sure - i think you're looking at a previous commit?)

Issue #2513, opened 2019-11-11
allow extra-parameters (ex : unfilledtimeout.buy) to be defined either in configuration file or strategy


gaugau3000:
**As** a trader,
**I want** to be able to customize strategy depends parameters (like `unfilledtimeout.buy`) straight in my strategy file
**So that** my configuration file in more generic,  not embedded strategy design choices and allow me to reduce the number of config file in dryrun/live.

`unfilledtimeout.buy` is just an example but others parameters can be strategy dependents.

xmatthias:
well, this is easier to accomplish than you might think - it's technically one-liner, assuming you use a dict with 'buy', 'sell' keys in the strategy as follows

``` python
unfilledtimeout = {'buy': 20, 'sell': 20}
```

I am however not certain that this is a strategy related configuration ... to me, this depends on the exchange and pairs (volume), not so much about the strategy used...

gaugau3000:
Hi,

Tks for solution

> I am however not certain that this is a strategy related configuration ... to me, this depends on the exchange and pairs (volume), not so much about the strategy used...

1/ Interesting what it represent for you . I agree with your points but i also  consider it is timeframe dependent (you know i count in candle and not is time :-) ). It is ok for me too wait 1 or 2 candle to fill the order as i consider my buy entry have impact during n candles.

2/ When i read the section [docs : parameters in the strategy](https://www.freqtrade.io/en/latest/configuration/#parameters-in-the-strategy) it feel like i cannot use your solution (as the introspection/scope of strategy variable work only on the parameter list we can see in the doc).

Many things can be consider as strategy dependent (the stake_amount could depends on my stop_loss that depends of my strategy, the stake_currency as market behave differently, even the internals.process_throttle_secs has it don't need to compute every 5 seconds for 1h strategy).

In fine the config mix strategy / technical /exchange choices.

If we can define all parameters in strategy file if we would like (technically) so the section `Parameters in the strategy` is just an advice ? And you have understand for what is say below i don't consider it as exhaustive : maybe we can review all parameters to see what to put here ?

Have a nice day.

xmatthias:
2) well - i said it's not much work - it does not work now, sorry if that wasn't clear from the above.

> even the internals.process_throttle_secs has it don't need to compute every 5 seconds for 1h strategy).

[...]

PR #2523, opened 2019-11-12
Improve timedout handling

xmatthias:
## Summary
unfilledtimeout is a mandatory key in config, so checking for it before running the timeout sequence does not make sense.
In addition to that, not running this part would break the "order cancelled on exchange" logic.

Also, the logic is identical for buy and sell orders - so this can be refactored into it's own method, which will simplify the whole logic.

## Quick changelog

- Extract comparison logic 
- simplify `check_handle_timedout()`

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -761,7 +761,7 @@ class FreqtradeBot:
         """
         timeout = self.config.get('unfilledtimeout', {}).get(side)
         ordertime = arrow.get(order['datetime']).datetime
-        if timeout:
+        if timeout is not None:
             timeout_threshold = arrow.utcnow().shift(minutes=-timeout).datetime
 
             return (order['status'] == 'open' and order['side'] == side
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -139,10 +139,9 @@ class FreqtradeBot:
         if len(trades) < self.config['max_open_trades']:
             self.process_maybe_execute_buys()
 
-        if 'unfilledtimeout' in self.config:
-            # Check and handle any timed out open orders
-            self.check_handle_timedout()
-            Trade.session.flush()
+        # Check and handle any timed out open orders
+        self.check_handle_timedout()
+        Trade.session.flush()
 
         if (self.heartbeat_interval
                 and (arrow.utcnow().timestamp - self._heartbeat_msg > self.heartbeat_interval)):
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -756,23 +755,28 @@ def _check_and_execute_sell(self, trade: Trade, sell_rate: float,
             return True
         return False
 
+    def _check_timed_out(self, side: str, order: dict) -> bool:
+        """
+        Check if timeout is active, and if the order is still open and timed out
+        """
+        timeout = self.config.get('unfilledtimeout', {}).get(side)
+        ordertime = arrow.get(order['datetime']).datetime
+        if timeout:
```
hroff-1902: This 1) won't actually happen (the `timeout==0` case) since limits for those values in the config schema are positive; 2) probably changes the behavior of the function here (previous code does not contain this check); 3) if left (it may have sense, to have this function working as a box, without knowledge on limits applies to the settings in the config schema), this should probably be changed to `if timeout > 0:`

freqtrade/freqtradebot.py
```diff
@@ -756,23 +755,28 @@ def _check_and_execute_sell(self, trade: Trade, sell_rate: float,
             return True
         return False
 
+    def _check_timed_out(self, side: str, order: dict) -> bool:
+        """
+        Check if timeout is active, and if the order is still open and timed out
+        """
+        timeout = self.config.get('unfilledtimeout', {}).get(side)
```
hroff-1902: No need to use `.get()` here (see above, as in `_process()`, the check was removed)?

Btw, why are those `unfilledtimeout` settings marked as required? The default values that are now in the default config, can be efficiently applied to these settings.

freqtrade/freqtradebot.py
xmatthias: 1) yes, currently it can't happen - but it may happen in the future, assuming we allow loading this from the strategy - so we could get an incomplete dict. 
see #2513 for details.

The check was also there before - but outside of this section (checking if "unfilledtimeout" is in the config must be seen equal to this check.

>  this should probably be changed to `if timeout > 0:`

hmm why disallow a limit of 0? it can simulate a "FOK" / "IOK" order on exchanges which don't support this.

freqtrade/freqtradebot.py
xmatthias: > No need to use .get() here (see above, as in _process(), the check was removed)?

you're right, I've overseen this check (so it was present twice, actually).

It's important to always call `check_handle_timedout` - since it checks 2 things - timeout - and for orders cancelled on exchange. Without this, we will not be able to detect this case.

freqtrade/freqtradebot.py
hroff-1902: > The check was also there before - but outside of this section (checking if "unfilledtimeout" is in the config must be seen equal to this check.

No, now if timeout is 0, the `timeout_threshold = arrow.utcnow().shift(minutes=-timeout).datetime` won't be calculated as well as further calculation for `(order['status'] == 'open' and order['side'] == side and ordertime < timeout_threshold)` will not be performed and False will be returned. Previously, if `buy_timeout = self.config['unfilledtimeout']['buy']` is set to 0 (well, assume it's from strategy, as you say), the `buy_timeout_threshold = arrow.utcnow().shift(minutes=-buy_timeout).datetime` and further check for `or (order['status'] == 'open' and order['side'] == 'buy' and ordertime < buy_timeout_threshold)` were performed anyway. That's the difference in behavior, specifically for the case when timeout value is 0.

> hmm why disallow a limit of 0? it can simulate a "FOK" / "IOK" order on exchanges which don't support this.

But you've disallowed it with that check `if timeout:` anyway... 😉

xmatthias: ah yes - so the correct check should be `if timeout is None:` ?

hroff-1902: I do not know, both cannot be set in the config in the meantime...
What's the intention of this check?

xmatthias: a) prepare to support 0 (The minimums seem very randomly chosen anyway) ... 
b) pepare to have `"unfilledtimeout"` only available for one of the two 
c) point b this **is** possible at the moment - none of the 2 keys are mandatory - so my config can look as follows and be valid.

```
"unfilledtimeout": {
        "buy": 10,
    },

```

This will crash (in develop) once it has the first open sell-order - but it's still a valid configuration - so this check will (implicitly) support this type of configuration.

PR #2533, opened 2019-11-15
/Balance rework

### LINKED ISSUE #2515: Telegram command /balance not working

Yes

  * Operating system: Debian 9
  * Python Version: 3.7.5
  * CCXT version: 1.19.25
  * Branch: Develop
  * Last Commit ID: 904ae5af915d066ac14c4a4f199f44e5ea9ae7c7
  
The command '/balance' gets no response
The last time it was working was two days ago.
That was on an older Commit but it stopped working without updating anything -
so I updated to the newest Commit but it's still not fixed)

### Steps to reproduce:

  1. send '/balance' to the Telegram bot
  
### Observed Results:

  It shows nothing on Telegram but throws that exception in the console

2019-11-11 19:44:30,705 - freqtrade.rpc.telegram - ERROR - Exception occurred within Telegram module
  File "/root/trader/freqtrade/freqtrade/rpc/telegram.py", line 52, in wrapper
  File "/root/trader/freqtrade/freqtrade/rpc/telegram.py", line 326, in _balance
    result = self._rpc_balance(self._config.get('fiat_display_currency', ''))
  File "/root/trader/freqtrade/freqtrade/rpc/rpc.py", line 283, in _rpc_balance
    rate = self._freqtrade.get_sell_rate(pair, False)
  File "/root/trader/freqtrade/freqtrade/freqtradebot.py", line 581, in get_sell_rate
    rate = self.exchange.get_ticker(pair, refresh)['bid']
  File "/root/trader/freqtrade/freqtrade/exchange/common.py", line 113, in wrapper
    return f(*args, **kwargs)
  File "/root/trader/freqtrade/freqtrade/exchange/exchange.py", line 524, in get_ticker
    'bid': float(data['bid']),
TypeError: float() argument must be a string or a number, not 'NoneType'

xmatthias:
## Summary
`/balance` should not fail if a pair has no bid.
`/balance` also does not need to call each `fetch_ticker()` once - but can use the `fetch_tickers()` method, which will fetch all tickers at once (so it's one call, instead of one per owned currency).

This will also handle "non-bid" situations - where bid is empty as seen in #2515 .

closes #2515 

## Quick changelog

- `/balance` should not fail if a pair has no bid.
- `/balance` also does not need to call each `fetch_ticker()` once - but can use the `fetch_tickers()` method, which will fetch all tickers at once (so it's one call, instead of one per owned currency).
- Allow "empty" bids - these pairs will be filtered from the output though.
- Use stake_currency instead of hardcoding to BTC (if you trade ETH as your stake - you're probably not interrested in the BTC value, but look at either FIAT, or stake).
- Use `freqtradebot.wallets` instead of calling `exchange.get_balance()` - avoiding another RPC call (wallets is refreshed quite often during bot operations).

### DIFF
--- a/tests/pairlist/test_pairlist.py
+++ b/tests/pairlist/test_pairlist.py
@@ -100,7 +100,7 @@ def test_refresh_pairlist_dynamic(mocker, shitcoinmarkets, tickers, whitelist_co
         markets=PropertyMock(return_value=shitcoinmarkets),
      )
     # argument: use the whitelist dynamically by exchange-volume
-    whitelist = ['ETH/BTC', 'TKN/BTC', 'LTC/BTC', 'HOT/BTC', 'FUEL/BTC']
+    whitelist = ['ETH/BTC', 'TKN/BTC', 'LTC/BTC', 'XRP/BTC', 'HOT/BTC']
     bot.pairlists.refresh_pairlist()
 
     assert whitelist == bot.pairlists.whitelist
@@ -135,10 +135,10 @@ def test_VolumePairList_refresh_empty(mocker, markets_empty, whitelist_conf):
 
 @pytest.mark.parametrize("pairlists,base_currency,whitelist_result", [
     ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "quoteVolume"}],
-        "BTC", ['ETH/BTC', 'TKN/BTC', 'LTC/BTC', 'HOT/BTC', 'FUEL/BTC']),
+        "BTC", ['ETH/BTC', 'TKN/BTC', 'LTC/BTC', 'XRP/BTC', 'HOT/BTC']),
     # Different sorting depending on quote or bid volume
     ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "bidVolume"}],
-        "BTC",  ['HOT/BTC', 'FUEL/BTC', 'LTC/BTC', 'TKN/BTC', 'ETH/BTC']),
+        "BTC",  ['HOT/BTC', 'FUEL/BTC', 'XRP/BTC', 'LTC/BTC', 'TKN/BTC']),
     ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "quoteVolume"}],
         "USDT", ['ETH/USDT']),
     # No pair for ETH ...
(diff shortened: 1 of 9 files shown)


tests/pairlist/test_pairlist.py
```diff
@@ -108,12 +108,12 @@ def test_VolumePairList_refresh_empty(mocker, markets_empty, whitelist_conf):
 
 
 @pytest.mark.parametrize("precision_filter,base_currency,key,whitelist_result", [
-    (False, "BTC", "quoteVolume", ['ETH/BTC', 'TKN/BTC', 'LTC/BTC']),
-    (False, "BTC", "bidVolume", ['LTC/BTC', 'TKN/BTC', 'ETH/BTC']),
+    (False, "BTC", "quoteVolume", ['ETH/BTC', 'TKN/BTC', 'LTC/BTC', 'XRP/BTC']),
+    (False, "BTC", "bidVolume", ['LTC/BTC', 'TKN/BTC', 'ETH/BTC', 'XRP/BTC']),
```
xmatthias: By adding this ticker to "tickers", it's subject to all tests, therefore it's added here as well.

hroff-1902:
I do not like this, it's an example of bad design, where the core functionality is moved to other modules instead of being collected in the single class.

In my opinion, the RPC module should not calculate any values by itself -- it should only fetch the values from the methods/attributes of the Freqtradebot class. For this PR, `rpc_balance()` should not calculate balances -- it should simply call `get_balance()` from Freqtradebot.

Same for the calculation of the rate: I do not like that the calculation of balance do not call a method from Freqtradebot anymore, but tries to handle all the specific cases (does the exchange have the 'bid'/'ask' values in the tickers or not) by itself. All those cases should be handled inside the `get_sell_rate()` method and the resulting value is to be used.

xmatthias:
I completely disagree.
I know you don't use the RPC module (especially telegram) and did not look in depth into them - so letme try to explain how this works and why:

> In my opinion, the RPC module should not calculate any values by itself -- it should only fetch the values from the methods/attributes of the Freqtradebot class. For this PR, rpc_balance() should not calculate balances -- it should simply call get_balance() from Freqtradebot.

the balance itself is not calculated here - but the BTC- (now stake-) rate is. It's used ONLY in / for RPC - so it's even worse design to me to move this into freqtradebot or any other class - where it's not used.

The same goes for the FIAT conversation. 
This part cannot be moved to another part, since the whole fiat class is only initialized when the following 2 conditions are met (both)

* telegram is enabled
* fiat_display_currency is set.

it is not initialized if it's not needed - since it's another call to a webservice - which we should only make if we need it.

> Same for the calculation of the rate: I do not like that the calculation of balance do not call a method from Freqtradebot anymore, but tries to handle all the specific cases (does the exchange have the 'bid'/'ask' values in the tickers or not) by itself. All those cases should be handled inside the get_sell_rate() method and the resulting value is to be used.

you need to be careful with this - calling `get_sell_rate()` does not make sense in this context. 
`/balance` can return currencies you don't have a ticker value for (see #2515 for an example). 
It can be worse - you could own currencies on the exchange which are not tradeable anymore - or not tradeable yet - so it's not just the bid-side that's empty, but there IS no orderbook, and there exists no current rate.
Handling these cases in `get_sell_rate()` does not make sense since these pairs cannot be traded by the bot.

in addition to the above - `/balance` is an informative command - so if it does not have the "very latest rate from orderbook", that's not that important. 
`get_sell_rate()` on the other hand defines the rate your'e going to sell the pair a few moments later - so it needs to have the latest available information.

Assume you have 10ETH, if that's currently 1000USDT, or 1010 USD - but it gives you an idea how much it's worth, even if that information is a few seconds old (it's usually not older - but `get_tickers()` does not guarantee the very latest price).

For that reason, we can use `get_tickers()` - which is a combined command (returns rates for all available pairs).

This reduces the calls we can make from one per "owned asset", to a total of 1 returning info for all pairs - making the amount of owned assets irrelevant.

If you've ever used `/balance` on an account you have more than 2 or 3 pairs (on binance you'll probably have some "dust" for each pair you ever owned), you'll notice that the answer takes "quite some time" (usually multiple seconds). 
That's because the bot needs to call `get_ticker()` for each asset - fetching the latest rate. 

with this new approach, the command is a lot faster - simply because it makes only one call, not 10 (assuming i own 10 currencies).

hroff-1902:
> I know you don't use the RPC module (especially telegram) and did not look in depth into them [...]

Right, I mostly collect the info on the bot performance by parsing the logs.

> [...] It's used **only** in / for RPC [...]

I need to see this info in the log as well, I understand that this is what I'm missing there...

xmatthias:
> Right, I mostly collect the info on the bot performance by parsing the logs.

but /balance is not about the bot performance (and i think this is the misunderstanding).
the info is **additional** to what the bot does - it shows you the balance of your account - not what your bot did or how the strategy performs (you can probably infer that ... but it's not clearly visible unless you know your balance history).


> I would like to see this info (the balance(s)) in the log as well, I understand that is what I'm missing there...

when would you print it? on every iteration? after every trade? (both don't make sense).

it only makes sense "when requested" - which is an RPC method (no way to do that if all RPC stuff is disabled afaik).

If you don't want to use RPC (telegram / rest) for it - then simply go to the exchange page and look at your account balances ... the info will be identical. 
The bot does not add anything to this other than the convenience of also seeing this via telegram.

PR #2535, opened 2019-11-15
Fix non-terminating bot

xmatthias:
## Summary
since we're now calling `freqtrade trade` - worker does not exist in main anymore, causing bots to "not stop" when used with telegram and trying to sigint the bot.

Solve the issue: from slack ...

## Quick changelog

- Move keyboardinterrupt and `worker.exit()` to `utils.py` (where it belongs).

### DIFF
--- a/freqtrade/utils.py
+++ b/freqtrade/utils.py
@@ -45,8 +45,15 @@ def start_trading(args: Dict[str, Any]) -> int:
     """
     from freqtrade.worker import Worker
     # Load and run worker
-    worker = Worker(args)
-    worker.run()
+    try:
+        worker = Worker(args)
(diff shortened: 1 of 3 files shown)


freqtrade/utils.py
```diff
@@ -45,8 +45,15 @@ def start_trading(args: Dict[str, Any]) -> int:
     """
     from freqtrade.worker import Worker
     # Load and run worker
-    worker = Worker(args)
-    worker.run()
+    try:
+        worker = Worker(args)
```
hroff-1902: `worker = None` should be assigned above, otherwise `if worker` will fail below if an exception happens during the initialization of worker

PR #2538, opened 2019-11-17
new-strategy / new-hyperopt - from templates

xmatthias:
## Summary
In order to contain the userdata templates with pypi, they need to be in the freqtrade module. 
For this, they've been moved to freqtrade/templates. using `freqtrae create-userdir` will copy them to the correct location. This also helps avoiding problems with updates should the sample files in the user_dir be modified (currently they are under version control - so an update would fail in this case).

Also, to simplify strategy and hyperopt development, new commands, `new-strategy` and `new-hyperopt` have been added. they deploy a template strategy/hyperopt with the given name to the correct location. This will remove the repeated work of "copy strategy, adjust names, ...". The file name will correspond to the class name - this should help avoid naming collisions.

Part of  #2112 - helps moving to pypi.

## Quick changelog

- move strategy samples into the module
- `create-userdir` will now copy samples into the correct location
- new command: `freqtrade new-strategy --strategy NewCoolStrategy` 
- new command: `freqtrade new-hyperopt --hyperopt NewCoolHyperopt`

Note: jinja2 is not a new dependency - it's been there before implicitly via flask - but now it's become "explicit" since we use it directly.

## Todo:

* [x] document new commands
* [x] maybe remove the "commented" sample indicators from the template (or change commenting to be line-wise comments, to make it easier to uncomment single indicators) - needs discussion
* [x] ensure updating via git does work correctly
  * note: this works fine for me - (due to the .gitkeep files, which don't remove the directories...) - but i'd apreciate other feedback


hroff-1902:
> this works fine for me - (due to the .gitkeep files, which don't remove the directories...)

Could you describe this in some more details, what's the issue here?

hroff-1902:
> maybe remove the "commented" sample indicators from the template (or change commenting to be line-wise comments, to make it easier to uncomment single indicators) - needs discussion

Actually, I'd like to have a choice, implemented somehow (maybe by selecting different templates by an additional cli option, or by different commands) -- either derive new strategy/hyperopt from current template, with some sample guards/triggers (hyperopt) and some sample indicators (strategy) used. or, as a choice, derived from the 'empty' template, with no guards/triggers, with empty hyperopt spaces (hyperopt) and 'empty' `populate_*_trend()` methods returning no signals (strategy).

Yes, in the sample templates, indicators may (should ) be line-wise commented (or placed in separate multiline comments that can be easily decommented).

xmatthias:
> this works fine for me - (due to the .gitkeep files, which don't remove the directories...)
> Could you describe this in some more details, what's the issue here?

As said, it works for me, and i assume it'll work for others too, since the folders are not removed (only the checked in Sample files are moved).

The files (Sample*) move from `user_data/xxx/Sample*` to `freqtrade/templates/Sample*` ... so should you have modified the sample (which users shouldn't do anyway, and could have caused problems when we updated the files before, too) - then this could cause 

a) checkout problem due to git dirty state
b) loss of modification (would have happened for "regular" updates too) - and will also happen when using setup.sh and using the "reset" option.

xmatthias:
>  derived from the 'empty' template, with no guards/triggers, with empty hyperopt spaces (hyperopt) and 'empty' populate_*_trend() methods returning no signals (strategy).

Personally, i find a "totally empty" template very difficult to work with.

To simulate this: try to remove every content from the methods and then write it yourself ... I'm not able to correctly write a `buy_strategy_generator()` without copy/pasting (or wasting a ton of time, or looking into the samples and copying from there).

I'll try a "reduced" version (and a full-version) ... which only has the least amount of code to show "how to ..." - as a basis for copy/paste (most of our users are NOT developers - [and even developers are lazy by definition and mostly prefer copy-pasting] :laughing: )

xmatthias:
@hroff-1902 have a look at the current version - default is the "full" version - but a minimized version (one sample per type) is available.

Honestly, i don't see this as the final step - we can have more, detailed options in the future - but i think this is a huge step forward.

Note: i did not extract Hyperopt Triggers for now since we have "only" 3 ... so having 2 options for these is only of limited use.
Maybe we can build triggers (and guards) dynamically (think of `--triggers mfi rfi bb` ) - and it automatically builds everything that's needed. (will not do that in this PR though - this is already big enough!)

hroff-1902:
> @hroff-1902 have a look at the current version - default is the "full" version - but a minimized version (one sample per type) is available.

looks fine 👍 

> Honestly, i don't see this as the final step - we can have more, detailed options in the future - but i think this is a huge step forward.

Sure

> Note: i did not extract Hyperopt Triggers for now since we have "only" 3 ... so having 2 options for these is only of limited use.
> Maybe we can build triggers (and guards) dynamically (think of `--triggers mfi rfi bb` ) - and it automatically builds everything that's needed. (will not do that in this PR though - this is already big enough!)

Hmm, I think this is not actually needed and would make the cli too complicated... Let's return to this after some (long) time, when the usage of the current approach will be collected...

What I would like to see (also, not in the PR, I guess) -- to have an arbitrary string (list of strings) in the `--template` option instead of the fixed list. So that appropriate subtemplate would be searched ('loaded' in terms of freqtrade resolvers, though they provide different mechanism for this) on the fly -- find and use appropriate subtemplates, if any, using the string with their 'name' passed.
In case of list of string, if I add the `superempty` subtemplates, then `--template(s) superempty minimal` will search and use the `superempty` subtemplates (those that found) and fallback to `minimal` parts (for those subtemplates that was not found as `superempty`)...

xmatthias:
> What I would like to see (also, not in the PR, I guess) -- to have an arbitrary string (list of strings) in the --template option instead of the fixed list.

I don't see the point in this - it's a helper function to "quickly" get started with a new strategy - what you're proposing is a copy of an existing strategy (i guess with a predefined set of indicators / comparisons). If that's your usecase, a simple `cp OldStrategy.py NewStrategy.py`will be easier than to write a (syntactically correct) template that fits together and works correctly.

> 'loaded' in terms of freqtrade resolvers

this will be impossible, since these are templates, not python modules (which need to be syntactically correct).

> Hmm, I think this is not actually needed and would make the cli too complicated

Yes, it would be a bit more complicated (it's really 1-2 new arguments ... ) - but writing hyperopt files is very repetitive work if you think about it - you copy/paste the following few lines per indicator / side once you'd like to test. 
Yes, the INT parameters need to be adjusted - but to me, this would be a huge timesaver in getting setup with hyperopt...

```
if params.get('sell-mfi-enabled'):
    conditions.append(dataframe['mfi'] > params['sell-mfi-value'])
...
Integer(75, 100, name='sell-mfi-value'),
Categorical([True, False], name='sell-mfi-enabled'),

```

hroff-1902:
Well, my comment seems to be confusing -- I did not assume the Resolvers should be used. I meant that resolvers use the term 'loading' and jinja also use the term 'load' for templates, see https://jinja.palletsprojects.com/en/2.10.x/api/#basics for example.

So my proposal was to accept an arbitrary string in the `--template` option. So if `--template full` is used then the `*_full.j2` subtemplates are loaded, `--template minimal` is used -- then the `*_minimal.j2` subtemplates are 'loaded' (actually simply used in generation of the resulting file(s)), and if `--template somenewcase` is used, then `*_somenewcase.j2` subtemplates are 'loaded'. That simply differs from the current approach in one sense -- we won't be needed to adapt the list of possible choices for the `--template` option in the future when the new set(s) of subtemplates will be introduced.

xmatthias:
> So my proposal was to accept an arbitrary string in the --template option. So if --template full is used then the *_full.j2 subtemplates are loaded, --template minimal is used -- then the *_minimal.j2 subtemplates are 'loaded' (actually simply used in generation of the resulting file(s)), and if --template somenewcase is used, then *_somenewcase.j2 subtemplates are 'loaded'. That simply differs from the current approach in one sense -- we won't be needed to adapt the list of possible choices for the --template option in the future when the new set(s) of subtemplates will be introduced.

Accepting an arbitrary string is rather complicated (not technically - but from the concept) ... since all templates for one template type (strategy | hyperopt) must be available - otherwise the command would fail (yes, we can check for that ...).
I'm not per se against that - but i don't think it'll be used.

Think about it differently - if you add a new set of templates (say "complicated") - you need to add the files - and you need to add "complicated" to the "choices" array.

If we remove choices and allow it to be a string, then you'll obviously not add it to `choices` - but you need to add "complicated" to the help-string - so users are aware that this option exists  - which is 1 line above choices (which is automatically part of the helpstring) ... 

Even worse - we'll need to handle the error if a string is not valid - which ideally will point out 'you can choose between "full, simple, complicated2" template options' (or similar) ... so we now add it in 2 places, the helpstring and the exception message ...

hroff-1902:
okay, at least using the list of choices allows printing the values at the helpstring, which is not available in case of an arbitrary string accepted.

LGTM 👍

xmatthias:
That's one point, anyway... as said, i assume this command will still move around a little in the future - with new options being added (maybe) ...

hroff-1902:
The NEWS message should be issued for this, I guess.

PR #2540, opened 2019-11-17
Improve rest api client / status response

### LINKED ISSUE #2528: Bugs in the rest-client
(From Slack)

The rest-client: 
1) does not understand 'extended' syntax of config (with extra trailing commas in the lists, for example), validation fails;

2) does not fail if config specified does not exist (for instance, if `-c /config.json` was typed by an error instead of `-c ./config.json`).

3) In the following output:
```
$ python3 scripts/rest_client.py --c config.json status
{'error': 'Error querying _status: no active trade'}
```
the "Error querying" message is misleading. Why fetching status returns an error?

xmatthias:
## Summary
Fix some issues and inconsistencies with the rest client.
Unfortunately, the rest-client now a dependency on rapidjson (previously the only dependency was `requests`).

closes #2528

## Quick changelog

- accept "commented" configs
- return empty list intead of error when calling `status` when no trade is open
- fail if config is not found (it won't work without config anyway since the username / password is required).

### DIFF
--- a/scripts/rest_client.py
+++ b/scripts/rest_client.py
@@ -8,12 +8,14 @@ so it can be used as a standalone script.
 """
 
 import argparse
+import inspect
 import json
 import logging
-import inspect
-from urllib.parse import urlencode, urlparse, urlunparse
+import sys
 from pathlib import Path
+from urllib.parse import urlencode, urlparse, urlunparse
 
+import rapidjson
 import requests
 from requests.exceptions import ConnectionError
 
@@ -190,7 +192,9 @@ class FtRestClient():
 def add_arguments():
     parser = argparse.ArgumentParser()
     parser.add_argument("command",
-                        help="Positional argument defining the command to execute.")
+                        help="Positional argument defining the command to execute.",
+                        nargs="?"
+                        )
 
     parser.add_argument('--show',
                         help='Show possible methods with this client',
@@ -221,9 +225,12 @@ def load_config(configfile):
     file = Path(configfile)
     if file.is_file():
         with file.open("r") as f:
-            config = json.load(f)
+            config = rapidjson.load(f, parse_mode=rapidjson.PM_COMMENTS |
+                                    rapidjson.PM_TRAILING_COMMAS)
         return config
-    return {}
+    else:
+        logger.warning(f"Could not load config file {file}.")
+        sys.exit(1)
 
 
 def print_commands():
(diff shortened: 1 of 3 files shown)


scripts/rest_client.py
```diff
@@ -237,8 +244,9 @@ def print_commands():
 
 def main(args):
 
-    if args.get("help"):
+    if args.get("show"):
```
hroff-1902: Why 'show'? 'help' was okay to print commands usage...

xmatthias: but "help" did never work. 
there's 2 types of help for this - "--help" - which shows available command line switches - and "--show" - showing available commands.

The correct argument here was always show ... --help is not populated but exits directly from argparse, so never reaches this line.

hroff-1902: the name of the `show` command is misleading. Okay for now, but a better way to output it should be 

the list of the commands can be shown (there are many alternatives):
* in the `usage` attribute of the argparse parser (will be printed before arguments description in the output of `--help`)
* as output of `help` command (so `--help` prints options and `help` command prints commands)
* as output of `show commands`, `show_commands`, `help commands`, `--help commands`
* etc.

Using `show` for this is misleading.

2. Pythonic `:return:`s in the list of commands is ugly.

Just imho

xmatthias: > Pythonic :return:s in the list of commands are ugly.

Basically, we're printing the docstrings - which removes the need to document each command seperately (which will lead to mistakes if it's pulled into one string).

I'm open to suggestions - but while being ugly - it works and documents stuff (and honestly, i don't think the rest-client is that heavily used) - who can, will probably use telegram ...

> as output of show commands, show_commands, help commands, --help commands

one way to get to this list is indeed `python scripts/rest_client.py help` (actually ANY command which does not exist) ... but requires the config to be found now ...

hroff-1902: pls review my commit

xmatthias: overall, the output did improve by removing the :return: lines.

However,  the "method" of doing this also removed some information about optional (or mandatory) parameters.

Please have a look at my latest change to this.

*side note to self*: we should update the documentation regarding this with the outcome of these changes ...

hroff-1902:
Well, okay, let this be this way (it's just a rest-client in the scripts dir) for now, but my general opinion is that it's not fine to output reST-formatted pythonic docstrings in the cli.

PR #2541, opened 2019-11-17
Rpc/show config

xmatthias:
## Summary
Add a "show config" command to RPC methods.

This allows inspection of some key-parameters on a running bot.
It's mainly helpful if while testing multiple strategies in parallel, and intends to quickly show some key strategy parameters (ROI / stoploss/ ticker_interval...).
Additional information (whitelist ...) can be gathered via the appropriate endpoints.

It intentionally does not publish the whole config dictionary, but only a selected subsection, to avoid leakage of sensitive information.

## Quick changelog

- add /show_config to telegram
- add /show_config to api-server

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -106,6 +106,7 @@ python3 scripts/rest_client.py --config rest_config.json <command> [optional par
 | `stop` | | Stops the trader
 | `stopbuy` | | Stops the trader from opening new trades. Gracefully closes open trades according to their rules.
 | `reload_conf` | | Reloads the configuration file
+| `show_config` | | Shows part of the current configuration with relevant settings to operation
 | `status` | | Lists all open trades
 | `count` | | Displays number of trades used and available
 | `profit` | | Display a summary of your profit/loss from close trades and some stats about your performance
@@ -171,6 +172,10 @@ profit
 reload_conf
         Reload configuration
         :returns: json object
+
+show_config
+        Returns part of the configuration, relevant for trading operations.
+        :return: json object containing the version
 
 start
         Start the bot if it's in stopped state.
--- a/docs/telegram-usage.md
+++ b/docs/telegram-usage.md
@@ -53,6 +53,7 @@ official commands. You can ask at any moment for help with `/help`.
 | `/stop` | | Stops the trader
 | `/stopbuy` | | Stops the trader from opening new trades. Gracefully closes open trades according to their rules.
 | `/reload_conf` | | Reloads the configuration file
+| `/show_config` | | Shows part of the current configuration with relevant settings to operation
 | `/status` | | Lists all open trades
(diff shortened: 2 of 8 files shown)


hroff-1902:
I propose to add somewhere a setting to switch it on, like
```
"enableshowconfig": true,
```
(the name should be aligned with other `'enable': true/false` settings)
with 'false' value by default. So that the user has to switch it on by herself explicitly.

That's like 'DEBUG' mode in Django, where the settings are printed in the errorneous page, should be switched off in production.

xmatthias:
I don't see why this should be necessary - not needing to do this is the point of not showing the whole config but only "trading-related" options, since sensitive information is not leaked (ROI / stoploss is not sensitive to me - it can be gathered from other RPC commands too).

With the info from this command, you know when the bot will sell via ROI and stoploss - not your API keys.

I guess your assumption is abuse of this command .... but the abuse potential of almost any other command is a lot higher - if i have access to your telegram, i can /forcesell every pair the bot buys...

hroff-1902:
Okay, I revoke my suggestion since only small part of the settings is printed (I have not taken a look at the details of the values printed during previous review, my bad). In this case additional explicit actions from the user to switch it on are not needed, no security deficiencies in this case...

However, pls consider renaming this command in this case, to `/show_strategy_settings` or `/strategy_settings`, since this command *does not* actually print the config (or it's significant part, only some strategy-specific subset)...

xmatthias:
> The secrets are never printed, even with all.

and this is why i've not done this. 
we have a few locations where we know secrets are (telegram, exchange/key, exchange/secret)... but there are other places where secrets can be "hidden" (webhook/, exchange/ccxt_config/ ...).
These don't necessarily to contain secrets by default - but they may contain secrets/ sensitive values, so depending on their content, they'd need to be removed (or not)... which makes this a risky process...

> However, pls consider renaming this command in this case, to /show_strategy_settings or /strategy_settings, since this command does not actually print the config (or it's significant part, only some strategy-specific subset)...

while it's true at the moment that it's mostly strategy focused, i'm not sure it'll remain this way - and if you think about it - dry_run is not a strategy setting - nor is exchange.

I'm having 2 motivations for this pr:
* i've had the need myself since i'm not clear what stoploss / roi setting my deployed strategy has
*  i'm thinking about an UI (for the future! / it's purely theory for now) ... which should display certain values. Until there is an actual implementation try, i'm not clear what values will be needed for that. While i don't wanna expose the full config (- secrets) - more (selected) values may be needed - and i would not want to have "another" endpoint for this.


I've been thinking about the name myself - but mainly for the 2nd reason i'm not sure if renaming it makes sense. 
**if** (see above why it's a big if) we ever expose the full config here, we can always do that via `/config all` (or `/config/all`) ...

PR #2552, opened 2019-11-18
FIx failure for PR's from forked repositories

xmatthias:
## Summary
Slack notifications from Forked repositories don't work and cause CI failures.

reason: secrets are currently not available to forked repositories - so cannot use slack notifications ...


xmatthias:
@hroff-1902 this will fix the CI failure we're seeing in #2560.

reason: secrets are currently not available to forked repositories - so cannot use slack notifications ...

Issue #2464, opened 2019-11-02
Hyperopt --continue doesn't work, suggestion


Theagainmen:
## Step 1: Describe your environment

  * Operating system: Windows 10
  * Python Version: 3.7 (64-bit)
  * Branch: Develop
  * Last Commit ID: develop-5254059f
  
## Step 2: Describe the problem:

Lately I've had issues with running hyperopts for over 1 month+ or backtesting with multiple strategies for a longer period of time. It would eventually crash with the memory error: `MemoryError: Unable to allocate array with shape (5, 52704) and data type float64`. Upgradeing from the 32-bit version to the 64-bit version made it a lot better. I can backtest for over a year with many strategies. However hyperopting over 150+ days still seem to crash at about 500+ epochs for me. Usually when it tends to go for trades amount > 10k+.

When it crashes I'm left with no information 
![hyperopt](https://user-images.githubusercontent.com/24569139/68069843-30e02c00-fd66-11e9-9211-21016a2c5699.png)

I would love to see a feature that saves the settings from each result that gets printed in this terminal. For example everytime it shows you it made an improvement, it writes it with the settings to a txt file. It's important that it does it while it's in the process and not when it's done. Otherwise when it crashes you're still left with nothing. This way when it crashes I can still see it's output in a txt file.
I could also use `--continue` (which currently doesn't work), but then it will probably go into the same loop where it crashed due to the many trades amount, and I will be left with none optimized output settings.

Reason for these suggestions is because hyperopting takes a long time and is a very intense process, when I'm hyperopting I can't use my PC for anything else (well.. it's very slow). With these things added it would save a lot of time and hyperopting :). 

### Steps to reproduce:

  1. Hyperopt a file for over 200+ days with 750 epochs.
  2. Watch it crash
 

### Relevant code exceptions or logs:

In case wondering, here is the memory error I'm getting: https://pastebin.com/J0Sai8Ym

hroff-1902:
Okay, taking it.

xmatthias:
well, i knew you had some thoughts about this before ... :)

hroff-1902:
This splits into 2 almost completely different parts:
* saving results more frequently
* get `--continue` working

Theagainmen:
> * saving results more frequently

Did it ever save results, except when it was finished? I know there are some pickle files saved but I can't read them. Maybe this already exist and I missed it, but I doubt it.

hroff-1902:
[...]

PR #2563, opened 2019-11-23
Hyperopt: Save epochs at intermediate points

hroff-1902:
Add saving epochs at intermediate points:
* after best epoch found
* every 100 epochs

Also,
* fixed handling of Ctrl+C pressed before first epochs were evaluated (and after skopt deprecation messages -- previously this produced an error with a backtrace printed).

Part of #2464

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -220,7 +220,7 @@ class Hyperopt:
         is_best_loss = self.is_best(results)
 
         if not print_all:
-            print('.', end='' if results['current_epoch'] % 100 != 0 else None)
+            print('.', end='' if results['current_epoch'] % 100 != 0 else None)  # type: ignore
             sys.stdout.flush()
 
         if print_all or is_best_loss:
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -23,7 +23,7 @@ from skopt import Optimizer
 from skopt.space import Dimension
 
 from freqtrade.data.history import get_timeframe, trim_dataframe
-from freqtrade.misc import round_dict
+from freqtrade.misc import plural, round_dict
 from freqtrade.optimize.backtesting import Backtesting
 # Import IHyperOpt and IHyperOptLoss to allow unpickling classes from these modules
 from freqtrade.optimize.hyperopt_interface import IHyperOpt  # noqa: F4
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -197,12 +210,20 @@ def log_trials_result(self) -> None:
...
     def log_results(self, results) -> None:
         """
         Log results if it is better than any previous evaluation
         """
         print_all = self.config.get('print_all', False)
-        is_best_loss = results['loss'] < self.current_best_loss
+        is_best_loss = self.is_best(results)
+
+        if not print_all:
+            print('.', end='' if results['current_epoch'] % 100 != 0 else None)  # type: ignore
```
xmatthias: what is the intend of this? 
print a newline every 100 epochs? (just to understand ...) ?

xmatthias:
This won't fix #2464, will it?
it'll dump the results - but it's a pickle file ... 

so yes, we have access to it - but to use it you'd need to use
`pickle.load(p.open('rb'))` to get to the results ? 

Also - it's loaded into hyperopt, and new results are appendet ...
can you remind me what the issue was with that approach?

hroff-1902:
> This won't fix #2464, will it?
> it'll dump the results - but it's a pickle file ...
> 
> so yes, we have access to it - but to use it you'd need to use
> `pickle.load(p.open('rb'))` to get to the results ?
> 
> Also - it's loaded into hyperopt, and new results are appended ...
> can you remind me what the issue was with that approach?

No it won't fix it.
The data are loaded from the pickle file, but they are not fit into the optimization models. That's the main point with it, why `--continue` does not work as expected.

I assume that on a way to fix that issue, an utlity like `hyperopt-list-results` (that prints the output in the same format as the hyperopt does, but takes the data from the saved file) and `hyperopt-show-result` (which prints the details of an epoch saved in the file) should be created first. This will then be used as a tool during the work on the optimization models, which should close the issue.

freqtrade/optimize/hyperopt.py
hroff-1902: It prints a dot and a newline after every 100th epoch and prints a single dot otherwise. (if the result is not a best result).

This is needed to separate the next log message ("Saving N00 epochs.") from the output (from the dots), otherwise it looks messy on the terminal.

xmatthias: good point :+1:

hroff-1902: A comment should be added here, if the intention of this is unclear (for future devs)...

PR #2564, opened 2019-11-23
[minor] Adjust plotting test to match user_data folder

xmatthias:
## Summary
This always created the "plots" folder in user_data ... when running tests.
It's not used anywhere other than in this test ...


hroff-1902:
What's the point here? The `user_data/plot` is not also contained in the repo...

PR #2576, opened 2019-11-26
Fix _get_min_pair_stake_amount

### LINKED ISSUE #2575: Bug in Freqtradebot._get_min_pair_stake_amount()
```
    def _get_min_pair_stake_amount(self, pair: str, price: float) -> Optional[float]:
...
        return min(min_stake_amounts) / amount_reserve_percent
```

This is wrong. max value of 2 limits should be returned instead.

For example, if as on Binance for ETH/BTC market, limit cost min is 0.0001, limit amount min is 0.001, price is 0.020405, then min returns `0.001 * 0.020405`, which is less than limit for cost, 0.0001.

<s>BTW, the limits returned are in the quote currency. Is it okay? Should not they than be in the base currency?</s>

hroff-1902:
Fixes #2575 

Markets in the conftest changed, since current limits set there causes many tests to fail -- the orders cannot be created because their amounts are less than those limits in the markets.

The rest of failing test cases should be carefully reviewed and adjusted.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -266,7 +266,7 @@ class FreqtradeBot:
             amount_reserve_percent += self.strategy.stoploss
         # it should not be more than 50%
         amount_reserve_percent = max(amount_reserve_percent, 0.5)
-        return min(min_stake_amounts) / amount_reserve_percent
+        return max(min_stake_amounts) / amount_reserve_percent
 
     def create_trades(self) -> bool:
         """
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -266,6 +266,10 @@ class FreqtradeBot:
             amount_reserve_percent += self.strategy.stoploss
         # it should not be more than 50%
         amount_reserve_percent = max(amount_reserve_percent, 0.5)
+
+        # The value returned should satisfy both limits: for amount (base currency) and
+        # for cost (quote, stake currency), so max() is used here.
+        # See also #2575 at github.
         return max(min_stake_amounts) / amount_reserve_percent
 
     def create_trades(self) -> bool:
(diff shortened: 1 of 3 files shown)


hroff-1902:
Okay, this function should prevent the bot from the situation when a small trade was open, the price drops and finally, at stoploss, it cannot sell the asset since the amount is less than the min limits (both, calculated as amount against the amount min limit or calculated as cost (amount*price) and compared against cost min limit). It should also prevent us from attempts to create a trade that cannot be created because it's less than minimal limits.

Assume stoploss is -0.05(5%) and reserve percentage used in this function is also 0.05. The multiplier, used in this function (1+stoploss-reserve) is therefore 0.9.

For BTC/USDT pair, for example, cost min limit is 10.0, amount min limit is 1e-6. Assume price ~7000.
The min() function gives us `min(10, 1e-6*7000) / 0.9 = 0.007 / 0.9 ~= 0.0077` (cost=quote=stake currency, USD).
* This will allow us to set stake_amount to 5, and the buy order will not be created (order size will be less than 10).
* For stake_amount=10 the buy order will be created, but the bot will not be able to sell the asset at stoploss, because it will be creating the sell order with the order size less than 10.
* If stoploss is 15% and we'll create a trade with stake_amount=11, which is greater that min order size 10 USDT (in the previous example), it will also prevent the bot from selling at stoploss, because price drops 15% and this will give us `11*0.85 < 10`.

Using `min()` here is completely wrong, because *both* limits, for amount(base) and for cost(stake, quote) should be satisfied.

xmatthias:
I'm not arguing the problem - you're absolutely right with it.

i just think that having the reasoning "next to the code" will be helpful for further readers of this code ... maybe just as link to the above comment - or a quick explanation directly next to the code ....

hroff-1902:
Yes, I completely understand this, that's fine.

PR #2577, opened 2019-11-26
"fix" config validation

### LINKED ISSUE #2531: ticker_interval cannot be defined only in strategy file ?
Hi,

docker image id  : 6d06e14cee5c

When i try to define the `ticker_interval` only in strategy file (`ticker_interval = "15m"`) i have this error :
`2019-11-14 10:43:05,787 - freqtrade - ERROR - Ticker-interval needs to be set in either configuration or as cli argument --ticker-interval 5m`

But in the documentation it says :

> The following parameters can be set in either configuration file or strategy
> - ticker_interval

Have a nice day.

### LINKED ISSUE #2532: Default value but required parameters in configuration (source documentation) ?
Hi,

I don't understand why in the [documentation -> configuration -> configuration-parameters](https://www.freqtrade.io/en/latest/configuration/#configuration-parameters) if parameters are required why there is a default value associated : we never can use the default value is this case ?

Many tks.

xmatthias:
## Summary
Since configuration options can come from multiple sources (cli, config file(s), strategy) - we need to validate the configuration after all possible sources have applied.
This will allow to specify mandatory options flexibly in the config validation dict - without manually coding the checks required.

Closes #2531
Closes #2532 (Defaults are now only in the documentation where they apply!) 

Improves visualization of config documentation (one fewer column).

## Quick changelog

- Validate after strategy is loaded - this can change options so we need to validate after that, not before.
  - This needs to be called from backtesting / trade / edgecli seperately
- Reevaluate some minimum values
- Reformat configuration documentation to be 2 columns only (default-values are now specified in the description - and only for fields which really apply a default value)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -40,12 +40,12 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 
 |  Command | Default | Description |
 |----------|---------|-------------|
-| `max_open_trades` | 3 | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades)
-| `stake_currency` | BTC | **Required.** Crypto-currency used for trading.
-| `stake_amount` | 0.05 | **Required.** Amount of crypto-currency your bot will use for each trade. Per default, the bot will use (0.05 BTC x 3) = 0.15 BTC in total will be always engaged. Set it to `"unlimited"` to allow the bot to use all available balance.
+| `max_open_trades` |  | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades)
+| `stake_currency` | | **Required.** Crypto-currency used for trading.
+| `stake_amount` | | **Required.** Amount of crypto-currency your bot will use for each trade. Per default, the bot will use (0.05 BTC x 3) = 0.15 BTC in total will be always engaged. Set it to `"unlimited"` to allow the bot to use all available balance.
 | `amount_reserve_percent` | 0.05 | Reserve some amount in min pair stake amount. Default is 5%. The bot will reserve `amount_reserve_percent` + stop-loss value when calculating min pair stake amount in order to avoid possible trade refusals.
-| `ticker_interval` | [1m, 5m, 15m, 30m, 1h, 1d, ...] | The ticker interval to use (1min, 5 min, 15 min, 30 min, 1 hour or 1 day). Default is 5 minutes. [Strategy Override](#parameters-in-the-strategy).
-| `fiat_display_currency` | USD | **Required.** Fiat currency used to show your profits. More information below.
+| `ticker_interval` | 5m | The ticker interval to use (1min, 5 min, 15 min, 30 min, 1 hour or 1 day). Default is 5 minutes. [Strategy Override](#parameters-in-the-strategy).
+| `fiat_display_currency` | | Fiat currency used to show your profits. More information below.
 | `dry_run` | true | **Required.** Define if the bot must be in Dry-run or production mode.
 | `dry_run_wallet` | 999.9 | Overrides the default amount of 999.9 stake currency units in the wallet used by the bot running in the Dry Run mode if you need it for any reason.
 | `process_only_new_candles` | false | If set to true indicators are processed only once a new candle arrives. If false each loop populates the indicators, this will mean the same candle is processed many times creating system load but can be useful of your strategy depends on tick data not only candle. [Strategy Override](#parameters-in-the-strategy).
@@ -94,8 +94,8 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `db_url` | `sqlite:///tradesv3.sqlite`| Declares database URL to use. NOTE: This defaults to `sqlite://` if `dry_run` is `True`.
 | `initial_state` | running | Defines the initial application state. More information below.
 | `forcebuy_enable` | false | Enables the RPC Commands to force a buy. More information below.
-| `strategy` | None | **Required** Defines Strategy class to use. Recommended to set via `--strategy NAME`. 
-| `strategy_path` | null | Adds an additional strategy lookup path (must be a directory).
+| `strategy` | | **Required** Defines Strategy class to use. Recommended to set via `--strategy NAME`.
+| `strategy_path` | | Adds an additional strategy lookup path (must be a directory).
 | `internals.process_throttle_secs` | 5 | **Required.** Set the process throttle. Value in second.
 | `internals.heartbeat_interval` | 60 | Print heartbeat message every X seconds. Set to 0 to disable heartbeat messages.
 | `internals.sd_notify` | false | Enables use of the sd_notify protocol to tell systemd service manager about changes in the bot state and issue keep-alive pings. See [here](installation.md#7-optional-configure-freqtrade-as-a-systemd-service) for more details.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -38,69 +38,69 @@ The prevelance for all Options is as follows:
 
 Mandatory parameters are marked as **Required**, which means that they are required to be set in one of the possible ways.
 
-|  Command | Default | Description |
-|----------|---------|-------------|
-| `max_open_trades` | | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades)
-| `stake_currency` | | **Required.** Crypto-currency used for trading.
-| `stake_amount` | | **Required.** Amount of crypto-currency your bot will use for each trade. Per default, the bot will use (0.05 BTC x 3) = 0.15 BTC in total will be always engaged. Set it to `"unlimited"` to allow the bot to use all available balance.
-| `amount_reserve_percent` | | Reserve some amount in min pair stake amount. The bot will reserve `amount_reserve_percent` + stop-loss value when calculating min pair stake amount in order to avoid possible trade refusals. *Defaults to `0.05` (5%).*
-| `ticker_interval` | | The ticker interval to use (1min, 5 min, 15 min, 30 min, 1 hour or 1 day). Default is 5 minutes. [Strategy Override](#parameters-in-the-strategy).
-| `fiat_display_currency` | | Fiat currency used to show your profits. More information below.
-| `dry_run` | | **Required.** Define if the bot must be in Dry-run or production mode.  *Defaults to `true`.*
-| `dry_run_wallet` | | Overrides the default amount of 999.9 stake currency units in the wallet used by the bot running in the Dry Run mode if you need it for any reason.
-| `process_only_new_candles` | | Enable processing of indicators only when new candles arrive. If false each loop populates the indicators, this will mean the same candle is processed many times creating system load but can be useful of your strategy depends on tick data not only candle.  *Defaults to `false`.* [Strategy Override](#parameters-in-the-strategy).
-| `minimal_roi` | | **Required.** Set the threshold in percent the bot will use to sell a trade. More information below. [Strategy Override](#parameters-in-the-strategy).
-| `stoploss` | |  **Required.** Value of the stoploss in percent used by the bot. More information below. More details in the [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy).
-| `trailing_stop` | | Enables trailing stop-loss (based on `stoploss` in either configuration or strategy file). More details in the [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy).
-| `trailing_stop_positive` | | Changes stop-loss once profit has been reached. More details in the [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy).
-| `trailing_stop_positive_offset` | | Offset on when to apply `trailing_stop_positive`. Percentage value which should be positive.  *Defaults to `0.0` (no offset).* More details in the [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy).
-| `trailing_only_offset_is_reached` | | Only apply trailing stoploss when the offset is reached.  *Defaults to `false`.* [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy).
-| `unfilledtimeout.buy` | | **Required.** How long (in minutes) the bot will wait for an unfilled buy order to complete, after which the order will be cancelled.
-| `unfilledtimeout.sell` | | **Required.** How long (in minutes) the bot will wait for an unfilled sell order to complete, after which the order will be cancelled.
-| `bid_strategy.ask_last_balance` | | **Required.** Set the bidding price. More information [below](#understand-ask_last_balance).
-| `bid_strategy.use_order_book` | | Enable buying using the rates in Order Book Bids.
-| `bid_strategy.order_book_top` | | Bot will use the top N rate in Order Book Bids. I.e. a value of 2 will allow the bot to pick the 2nd bid rate in Order Book Bids.  *Defaults to `1`.*
-| `bid_strategy. check_depth_of_market.enabled` | | Does not buy if the difference of buy orders and sell orders is met in Order Book.  *Defaults to `false`.*
-| `bid_strategy. check_depth_of_market.bids_to_ask_delta` | | The % difference of buy orders and sell orders found in Order Book. A value lesser than 1 means sell orders is greater, while value greater than 1 means buy orders is higher.  *Defaults to `0`.*
-| `ask_strategy.use_order_book` | | Enable selling of open trades using Order Book Asks.
-| `ask_strategy.order_book_min` | | Bot will scan from the top min to max Order Book Asks searching for a profitable rate. *Defaults to `1`.*
-| `ask_strategy.order_book_max` | | Bot will scan from the top min to max Order Book Asks searching for a profitable rate. *Defaults to `1`.*
-| `ask_strategy.use_sell_signal` | | Use sell signals produced by the strategy in addition to the `minimal_roi`.  *Defaults to `true`.* [Strategy Override](#parameters-in-the-strategy).
-| `ask_strategy.sell_profit_only` | | Wait until the bot makes a positive profit before taking a sell decision.  *Defaults to `false`.* [Strategy Override](#parameters-in-the-strategy).
-| `ask_strategy.ignore_roi_if_buy_signal` | | Do not sell if the buy signal is still active. This setting takes preference over `minimal_roi` and `use_sell_signal`.  *Defaults to `false`.* [Strategy Override](#parameters-in-the-strategy).
-| `order_types` | | Configure order-types depending on the action (`"buy"`, `"sell"`, `"stoploss"`, `"stoploss_on_exchange"`). [More information below](#understand-order_types). [Strategy Override](#parameters-in-the-strategy).
-| `order_time_in_force` | | Configure time in force for buy and sell orders. [More information below](#understand-order_time_in_force). [Strategy Override](#parameters-in-the-strategy).
-| `exchange.name` | | **Required.** Name of the exchange class to use. [List below](#user-content-what-values-for-exchangename).
-| `exchange.sandbox` | | Use the 'sandbox' version of the exchange, where the exchange provides a sandbox for risk-free integration. See [here](sandbox-testing.md) in more details.
-| `exchange.key` | | API key to use for the exchange. Only required when you are in production mode. **Keep it in secrete, do not disclose publicly.**
-| `exchange.secret` | | API secret to use for the exchange. Only required when you are in production mode. **Keep it in secrete, do not disclose publicly.**
-| `exchange.password` | | API password to use for the exchange. Only required when you are in production mode and for exchanges that use password for API requests. **Keep it in secrete, do not disclose publicly.**
-| `exchange.pair_whitelist` | | List of pairs to use by the bot for trading and to check for potential trades during backtesting. Not used by VolumePairList (see [below](#dynamic-pairlists)).
-| `exchange.pair_blacklist` | | List of pairs the bot must absolutely avoid for trading and backtesting (see [below](#dynamic-pairlists)).
-| `exchange.ccxt_config` | | Additional CCXT parameters passed to the regular ccxt instance. Parameters may differ from exchange to exchange and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
-| `exchange.ccxt_async_config` | | Additional CCXT parameters passed to the async ccxt instance. Parameters may differ from exchange to exchange  and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
-| `exchange.markets_refresh_interval` | | The interval in minutes in which markets are reloaded. *Defaults to `60` minutes.*
-| `edge.*` | | Please refer to [edge configuration document](edge.md) for detailed explanation.
-| `experimental.block_bad_exchanges` | | Block exchanges known to not work with freqtrade.  *Defaults to `true`.* Leave on default unless you want to test if that exchange works now.
-| `pairlists` | | Define one or more pairlists to be used. *Defaults to `StaticPairList`.* [More information below](#dynamic-pairlists).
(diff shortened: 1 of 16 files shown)


hroff-1902:
Quick summary: LGTM 👍 

As for defaults, my idea was to add the `*)` sign at the header of the Default column in the table and a note below the table describing that the default values stated for the Required settings are those values that are set in the default (sample) configs... But this is the better, I guess...

xmatthias:
>  But this is the better, I guess...

we'll see how it renders ...

> My proposal is to add the 'Type' ('Datatype', 'Possible values') column. Examples of the values in it:

No.
I intentionally removed the column to have it look better.
Just look at [the table](https://www.freqtrade.io/en/latest/configuration/#configuration-parameters) as it is now ... Description is "squashed" together, and Command is almost always wrapped, making it difficult to use and read ... so having 3 columns is definitely not a good approach, especially not datatype (with range) - which is even longer than the "default" we have now.

How about adding the datatype to the Description?
Right now, Default values (where they exist) "should" render with a newline ... we can use the same approach for datatypes - rendering them in a new line too.

hroff-1902:
Fine, the table with only 2 columns definitely renders better... Adding the datatypes in the description can be even better, since it allows more place for it. This can be in a separate line along with 'Defaults':

* _Boolean. Defaults to `false`._
* _Positive integer (1...). Defaults to 1._

xmatthias:
that's (kindof) what i had in mind, yes :+1:

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `max_open_trades` | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades)
+| `stake_currency` | **Required.** Crypto-currency used for trading. [Strategy Override](#parameters-in-the-strategy).
+| `stake_amount` | **Required.** Amount of crypto-currency your bot will use for each trade. Per default, the bot will use (0.05 BTC x 3) = 0.15 BTC in total will be always engaged. Set it to `"unlimited"` to allow the bot to use all available balance. [Strategy Override](#parameters-in-the-strategy).
+| `amount_reserve_percent` | Reserve some amount in min pair stake amount. The bot will reserve `amount_reserve_percent` + stop-loss value when calculating min pair stake amount in order to avoid possible trade refusals. <br>*Defaults to `0.05` (5%).*
+| `ticker_interval` | The ticker interval to use (1min, 5 min, 15 min, 30 min, 1 hour or 1 day). Default is 5 minutes. [Strategy Override](#parameters-in-the-strategy).
+| `fiat_display_currency` | Fiat currency used to show your profits. More information below.
+| `dry_run` | **Required.** Define if the bot must be in Dry-run or production mode. <br>*Defaults to `true`.*
```
hroff-1902: "Dry Run"

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `dry_run_wallet` | Overrides the default amount of 999.9 stake currency units in the wallet used by the bot running in the Dry Run mode if you need it for any reason.
+| `process_only_new_candles` | Enable processing of indicators only when new candles arrive. If false each loop populates the indicators, this will mean the same candle is processed many times creating system load but can be useful of your strategy depends on tick data not only candle. [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*
+| `minimal_roi` | **Required.** Set the threshold in percent the bot will use to sell a trade. More information below. [Strategy Override](#parameters-in-the-strategy).
```
hroff-1902: Link should be added to "below".

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `stoploss` |  **Required.** Value of the stoploss in percent used by the bot. More information below. More details in the [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy).
```
hroff-1902: Remove "More information below.", the link follows it.

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `trailing_only_offset_is_reached` | Only apply trailing stoploss when the offset is reached. [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.* 
+| `unfilledtimeout.buy` | **Required.** How long (in minutes) the bot will wait for an unfilled buy order to complete, after which the order will be cancelled.
+| `unfilledtimeout.sell` | **Required.** How long (in minutes) the bot will wait for an unfilled sell order to complete, after which the order will be cancelled.
+| `bid_strategy.ask_last_balance` | **Required.** Set the bidding price. More information [below](#understand-ask_last_balance).
+| `bid_strategy.use_order_book` | Enable buying using the rates in Order Book Bids.
+| `bid_strategy.order_book_top` | Bot will use the top N rate in Order Book Bids. I.e. a value of 2 will allow the bot to pick the 2nd bid rate in Order Book Bids.  *Defaults to `1`.*
+| `bid_strategy. check_depth_of_market.enabled` | Does not buy if the difference of buy orders and sell orders is met in Order Book. <br>*Defaults to `false`.*
```
hroff-1902: "Do not buy..."

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `ask_strategy.ignore_roi_if_buy_signal` | Do not sell if the buy signal is still active. This setting takes preference over `minimal_roi` and `use_sell_signal`. [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.* 
+| `order_types` | Configure order-types depending on the action (`"buy"`, `"sell"`, `"stoploss"`, `"stoploss_on_exchange"`). [More information below](#understand-order_types). [Strategy Override](#parameters-in-the-strategy).
+| `order_time_in_force` | Configure time in force for buy and sell orders. [More information below](#understand-order_time_in_force). [Strategy Override](#parameters-in-the-strategy).
+| `exchange.name` | **Required.** Name of the exchange class to use. [List below](#user-content-what-values-for-exchangename).
+| `exchange.sandbox` | Use the 'sandbox' version of the exchange, where the exchange provides a sandbox for risk-free integration. See [here](sandbox-testing.md) in more details.
+| `exchange.key` | API key to use for the exchange. Only required when you are in production mode. **Keep it in secrete, do not disclose publicly.**
```
hroff-1902: "secrete" --> "secret", everywhere in the doc. My bad, shame on me...

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `webhook.enabled` | Enable usage of Webhook notifications
+| `webhook.url` | URL for the webhook. Only required if `webhook.enabled` is `true`. See the [webhook documentation](webhook-config.md) for more details.
+| `webhook.webhookbuy` | Payload to send on buy. Only required if `webhook.enabled` is `true`. See the [webhook documentationV](webhook-config.md) for more details.
+| `webhook.webhooksell` | Payload to send on sell. Only required if `webhook.enabled` is `true`. See the [webhook documentationV](webhook-config.md) for more details.
+| `webhook.webhookstatus` | Payload to send on status calls. Only required if `webhook.enabled` is `true`. See the [webhook documentationV](webhook-config.md) for more details.
+| `db_url` | Declares database URL to use. NOTE: This defaults to `sqlite://` if `dry_run` is `True`, and to `sqlite:///tradesv3.sqlite` for production instances.
+| `initial_state` | Defines the initial application state. More information below. <br>*Defaults to `stopped`.*
+| `forcebuy_enable` | Enables the RPC Commands to force a buy. More information below.
```
hroff-1902: remove trailing whitespace char

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `telegram.token` | Your Telegram bot token. Only required if `telegram.enabled` is `true`. **Keep it in secrete, do not disclose publicly.**
+| `telegram.chat_id` | Your personal Telegram account id. Only required if `telegram.enabled` is `true`. **Keep it in secrete, do not disclose publicly.**
+| `webhook.enabled` | Enable usage of Webhook notifications
+| `webhook.url` | URL for the webhook. Only required if `webhook.enabled` is `true`. See the [webhook documentation](webhook-config.md) for more details.
+| `webhook.webhookbuy` | Payload to send on buy. Only required if `webhook.enabled` is `true`. See the [webhook documentationV](webhook-config.md) for more details.
+| `webhook.webhooksell` | Payload to send on sell. Only required if `webhook.enabled` is `true`. See the [webhook documentationV](webhook-config.md) for more details.
+| `webhook.webhookstatus` | Payload to send on status calls. Only required if `webhook.enabled` is `true`. See the [webhook documentationV](webhook-config.md) for more details.
+| `db_url` | Declares database URL to use. NOTE: This defaults to `sqlite://` if `dry_run` is `True`, and to `sqlite:///tradesv3.sqlite` for production instances.
```
hroff-1902: "true"

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `initial_state` | Defines the initial application state. More information below. <br>*Defaults to `stopped`.*
+| `forcebuy_enable` | Enables the RPC Commands to force a buy. More information below. 
+| `strategy` | **Required** Defines Strategy class to use. Recommended to set via `--strategy NAME`.
```
hroff-1902: "to be set" ?

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `strategy_path` | Adds an additional strategy lookup path (must be a directory).
```
hroff-1902: Add "Recommended to be set via _cli option_"?

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `internals.process_throttle_secs` | Set the process throttle. Value in second. <br>*Defaults to `5` seconds.*
+| `internals.heartbeat_interval` | Print heartbeat message every X seconds. Set to 0 to disable heartbeat messages. <br>*Defaults to `60` seconds.*
```
hroff-1902: "N seconds" ?

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
+| `internals.sd_notify` | Enables use of the sd_notify protocol to tell systemd service manager about changes in the bot state and issue keep-alive pings. See [here](installation.md#7-optional-configure-freqtrade-as-a-systemd-service) for more details.
+| `logfile` | Specify Logfile. Uses a rolling strategy of 10 files, with 1Mb per file.
```
hroff-1902: "Specifies logfile name. Uses a rolling strategy for log file rotation for 10 files with the 1MB limit per file."

docs/configuration.md
```diff
@@ -38,84 +38,87 @@ The prevelance for all Options is as follows:
...
-| `user_data_dir` | cwd()/user_data | Directory containing user data. Defaults to `./user_data/`.
+|  Command | Description |
+|----------|-------------|
+| `max_open_trades` | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades)
+| `stake_currency` | **Required.** Crypto-currency used for trading. [Strategy Override](#parameters-in-the-strategy).
+| `stake_amount` | **Required.** Amount of crypto-currency your bot will use for each trade. Per default, the bot will use (0.05 BTC x 3) = 0.15 BTC in total will be always engaged. Set it to `"unlimited"` to allow the bot to use all available balance. [Strategy Override](#parameters-in-the-strategy).
+| `amount_reserve_percent` | Reserve some amount in min pair stake amount. The bot will reserve `amount_reserve_percent` + stop-loss value when calculating min pair stake amount in order to avoid possible trade refusals. <br>*Defaults to `0.05` (5%).*
+| `ticker_interval` | The ticker interval to use (1min, 5 min, 15 min, 30 min, 1 hour or 1 day). Default is 5 minutes. [Strategy Override](#parameters-in-the-strategy).
```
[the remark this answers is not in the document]
xmatthias: no it's not ... if it's not defined the bot will fail to start ... i thought the default was 5m, too - but it's not - that's what's in all config samples - so it's not a default value.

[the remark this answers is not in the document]
hroff-1902: This should be marked as Required then?

My comment was that the actual value '"5m"' should better be used instead of "5 min".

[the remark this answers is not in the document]
xmatthias: mmh yes ...

docs/configuration.md
xmatthias: we've had FAR worse spelling errors than this ...

docs/configuration.md
xmatthias: no, i don't think so ... if you have your strategies in one location, having this in the configuration does make sense (i have user_data_dir in my config ...)

docs/configuration.md
```diff
@@ -40,67 +40,72 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
+| `exchange.pair_whitelist` | List of pairs to use by the bot for trading and to check for potential trades during backtesting. Not used by VolumePairList (see [below](#dynamic-pairlists)). <br> ***Datatype:*** *List*
+| `exchange.pair_blacklist` | List of pairs the bot must absolutely avoid for trading and backtesting (see [below](#dynamic-pairlists)). <br> ***Datatype:*** *List*
+| `exchange.ccxt_config` | Additional CCXT parameters passed to the regular ccxt instance. Parameters may differ from exchange to exchange and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation) <br> ***Datatype:*** *Dict*
+| `exchange.ccxt_async_config` | Additional CCXT parameters passed to the async ccxt instance. Parameters may differ from exchange to exchange  and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation) <br> ***Datatype:*** *Dict*
+| `exchange.markets_refresh_interval` | The interval in minutes in which markets are reloaded. <br>*Defaults to `60` minutes.* <br> ***Datatype:*** *Integer*
```
hroff-1902: "Positive integer"?

docs/configuration.md
```diff
@@ -40,67 +40,72 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
-| `exchange.pair_whitelist` | List of pairs to use by the bot for trading and to check for potential trades during backtesting. Not used by VolumePairList (see [below](#dynamic-pairlists)).
-| `exchange.pair_blacklist` | List of pairs the bot must absolutely avoid for trading and backtesting (see [below](#dynamic-pairlists)).
-| `exchange.ccxt_config` | Additional CCXT parameters passed to the regular ccxt instance. Parameters may differ from exchange to exchange and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
-| `exchange.ccxt_async_config` | Additional CCXT parameters passed to the async ccxt instance. Parameters may differ from exchange to exchange  and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
-| `exchange.markets_refresh_interval` | The interval in minutes in which markets are reloaded. <br>*Defaults to `60` minutes.*
+| `max_open_trades` | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades).<br> ***Datatype:*** *Positive integer (-1 to use `"unlimited"` trades).*
```
hroff-1902: "or -1..."

docs/configuration.md
```diff
@@ -40,67 +40,72 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
+| `stake_currency` | **Required.** Crypto-currency used for trading. [Strategy Override](#parameters-in-the-strategy). <br> ***Datatype:*** *String*
+| `stake_amount` | **Required.** Amount of crypto-currency your bot will use for each trade. Set it to `"unlimited"` to allow the bot to use all available balance. [More information below](#understand-stake_amount). [Strategy Override](#parameters-in-the-strategy). <br> ***Datatype:*** *Positive number or `"unlimited"`.*
+| `amount_reserve_percent` | Reserve some amount in min pair stake amount. The bot will reserve `amount_reserve_percent` + stoploss value when calculating min pair stake amount in order to avoid possible trade refusals. <br>*Defaults to `0.05` (5%).* <br> ***Datatype:*** *Positive number as ratio.*
```
hroff-1902: "Positive float"

docs/configuration.md
```diff
@@ -40,67 +40,72 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
-| `exchange.ccxt_async_config` | Additional CCXT parameters passed to the async ccxt instance. Parameters may differ from exchange to exchange  and are documented in the [ccxt documentation](https://ccxt.readthedocs.io/en/latest/manual.html#instantiation)
-| `exchange.markets_refresh_interval` | The interval in minutes in which markets are reloaded. <br>*Defaults to `60` minutes.*
+| `max_open_trades` | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades).<br> ***Datatype:*** *Positive integer (-1 to use `"unlimited"` trades).*
+| `stake_currency` | **Required.** Crypto-currency used for trading. [Strategy Override](#parameters-in-the-strategy). <br> ***Datatype:*** *String*
+| `stake_amount` | **Required.** Amount of crypto-currency your bot will use for each trade. Set it to `"unlimited"` to allow the bot to use all available balance. [More information below](#understand-stake_amount). [Strategy Override](#parameters-in-the-strategy). <br> ***Datatype:*** *Positive number or `"unlimited"`.*
```
hroff-1902: "Positive integer or float" ?

docs/configuration.md
```diff
@@ -40,67 +40,72 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
+| `experimental.block_bad_exchanges` | Block exchanges known to not work with freqtrade. Leave on default unless you want to test if that exchange works now. <br>*Defaults to `true`.* <br> ***Datatype:*** *Boolean*
+| `pairlists` | Define one or more pairlists to be used. [More information below](#dynamic-pairlists). <br>*Defaults to `StaticPairList`.*  <br> ***Datatype:*** *List of Dicts*
+| `telegram.enabled` | Enable the usage of Telegram. <br> ***Datatype:*** *Boolean*
+| `telegram.token` | Your Telegram bot token. Only required if `telegram.enabled` is `true`. **Keep it in secret, do not disclose publicly.** <br> ***Datatype:*** *String*
+| `telegram.chat_id` | Your personal Telegram account id. Only required if `telegram.enabled` is `true`. **Keep it in secret, do not disclose publicly.** <br> ***Datatype:*** *String*
+| `webhook.enabled` | Enable usage of Webhook notifications <br> ***Datatype:*** *Boolean*
+| `webhook.url` | URL for the webhook. Only required if `webhook.enabled` is `true`. See the [webhook documentation](webhook-config.md) for more details. <br> ***Datatype:*** *Boolean*
```
hroff-1902: String

docs/configuration.md
```diff
@@ -40,67 +40,72 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
+| `api_server.password` | Password for API server. See the [API Server documentation](rest-api.md) for more details. **Keep it in secret, do not disclose publicly.**<br> ***Datatype:*** *String*
+| `db_url` | Declares database URL to use. NOTE: This defaults to `sqlite://` if `dry_run` is `true`, and to `sqlite:///tradesv3.sqlite` for production instances. <br> ***Datatype:*** *String, SQLAlchemy connect string*
+| `initial_state` | Defines the initial application state. More information below. <br>*Defaults to `stopped`.* <br> ***Datatype:*** *Enum, either `stopped` or `running`*
+| `forcebuy_enable` | Enables the RPC Commands to force a buy. More information below. <br> ***Datatype:*** *Boolean*
+| `strategy` | **Required** Defines Strategy class to use. Recommended to be set via `--strategy NAME`. <br> ***Datatype:*** *ClassName*
+| `strategy_path` | Adds an additional strategy lookup path (must be a directory). <br> ***Datatype:*** *String*
+| `internals.process_throttle_secs` | Set the process throttle. Value in second. <br>*Defaults to `5` seconds.* <br> ***Datatype:*** *Integer*
```
hroff-1902: "Positive integer"? The limit should be added to schema...

docs/configuration.md
```diff
@@ -40,67 +40,72 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
+| `internals.heartbeat_interval` | Print heartbeat message every N seconds. Set to 0 to disable heartbeat messages. <br>*Defaults to `60` seconds.* <br> ***Datatype:*** *Integer*
```
hroff-1902: "Positive integer or 0"

docs/configuration.md
xmatthias: stake_amount is not an integer - (Think about BTC - where 1 Integer is several 1000$ ... )
float / number i'm not sure - float won't mean much to non-programmers ... (neither will integer ...)

hroff-1902: Can also be integer, for USDT -- like 100 USDT... It's a number in json IIRC, so the schema allows integers as well. Btw, if this is allowed now, an interger is set and not casted to a float, this could result in errors (like multiplying to integer number of trades and then divided by an integer somewhere in the code...)

Should be casted to float somewhere in configuration.py if this is not done, imho... Not in this PR, should be checked in the code. But integer is accepted and is to be accepted in the config and in the strategy)

docs/configuration.md
```diff
@@ -40,10 +40,10 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 
 |  Command | Description |
 |----------|-------------|
-| `max_open_trades` | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades).<br> ***Datatype:*** *Positive integer (-1 to use `"unlimited"` trades).*
+| `max_open_trades` | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades).<br> ***Datatype:*** *Positive integer or -1.*
```
hroff-1902: I think I have not written my idea clear enough, I'd like to keep the "unlimited" special value (imho it's better than special meaning for -1), my proposal was (refined version) "Positive integer, or -1, or `"unlimited"`. Use -1 or `"unlimited"` for unlimited number of trades." or something like that...

docs/configuration.md
hroff-1902: Or -1 can be marked as Deprecated, to have users time to adapt to new restrictions for this value. Not in this PR in this case.

docs/configuration.md
hroff-1902: Or -1 can be marked as Deprecated for this setting, to give users time to adapt to new restrictions for this value. Not in this PR in this case.

xmatthias: i would do neither in this PR ... 
the current implementation does not support `"unlimited"`, but uses `-1` instead (which will be converted to `float('inf')`).

Yes, the 2 should probably be aligned.
But if you notice, i did not change any setting in this PR really - only the description (and how it's validated ...) - so doing that change in this PR seems out of place.

On the other hand -  while using -1 is ok (or "unlimited - I don't really care) - using `float('inf')` changes the type of this variable from int to either `int` or float ... (inf is no integer).

I'd like to replace `float('inf')` with maxInt (or the python equivalent ...) to allow only checking for integer, thus disallowing float values via config setting ... but until i'm certain how to approach this i'd like to leave this as is...

hroff-1902: ah, it does not recognize the "unlimited" string currently. That's what I missed...
Not in this PR.

hroff-1902:
Approved, one small open point left (see above, about stake_amount), decide if you wish to change it or not, anything else LGTM 👍

docs/configuration.md
xmatthias: yes it is ... but AFAIK, an integer is treated as float if used in calculations and one of the 2 parties is an integer ... so i don't see this create a problem.

At least that's how it is in python3 (i think this was different in python2 - but i don't really care since we're not supporting that anyway) - i just tested ...

PR #2579, opened 2019-11-26
Revert "Try coveralls fixing (again ...) "

hroff-1902:
Reverts freqtrade/freqtrade#2578

Don't merge was flagged


xmatthias:
it doesn't matter, really ...
it didn't work before, it doesn't work now ... so we can also leave it and "work from the new state" ;)

PR #2581, opened 2019-11-26
Add hyperopt-list and hyperopt-show commands

hroff-1902:
hyperopt-list:
```
  --best                Select only best epochs.
  --profitable          Select only profitable epochs.
  --no-color            Disable colorization of hyperopt results. May be
                        useful if you are redirecting output to a file.
  --print-json          Print best result detailization in JSON format.
  --no-details          Do not print best epoch details.
```

hyperopt-show:
```
  --best                Select only best epochs.
  --profitable          Select only profitable epochs.
  -n INT, --index INT   Specify the index of the epoch to print details for.
  --print-json          Print best result detailization in JSON format.
  --no-header           Do not print epoch details header.
```

Examples:
* `freqtrade hyperopt-list --best --profitable --no-details` -- lists all best results with positive profit. No details of the best epoch printed at the end, so the list can be iterated in a script.
* `freqtrade hyperopt-show -n 168` -- prints details for the epoch 168 (the index value is shown by the `hyperopt-list` or by ordinal `hyperopt` previously ran)
* `freqtrade hyperopt-show --best -n -1 --print-json --no-header` -- prints JSON data with details for the last best epoch (i.e., the best of all epochs).

Compatibility with old hyperopt pickle file is broken, so first run hyperopt once for some small amount of epochs for testing this.


freqtrade/utils.py
```diff
@@ -21,6 +23,7 @@
 from freqtrade.exchange import (available_exchanges, ccxt_exchanges,
                                 market_is_active, symbol_is_pair)
 from freqtrade.misc import plural, render_template
+from freqtrade.optimize.hyperopt import Hyperopt
```
xmatthias: don't ... (just don't).

With this, the whole bot will require the hyperopt dependencies again - which does not make sense (and in addition to that, causes the test-failure since its creating a cyclic import).

we have 2 options for this:
* move this import into each function which needs it, 
* move the new "hyperopt" commands to `optimize/__init__.py`. This is where the backtesting / hyperopt commands are located. (note: if we don't opt for this approach, we should consider moving `start_backtesting()` and `start_hyperopt()` into this file.

hroff-1902: Agree (no dependency of hyperopt stuff for the whole freqtrade)

Initially I made those methods as functions outside of the Hyperopt class, but moved them back to their initial place as static methods to simplify reviewing the changes. As functions, they can easily be placed into a separate module (like `hyperopt_misc.py`) which will remove all the dependencies from other hyperopt stuff...

xmatthias: you'll still need to move the import into the function itself so it's only imported when they are actually called - since the functions themselfs are imported into arguments.

Why movethem to a new new file though - `freqtrade/optimize/__init__.py` should work just fine (?)

hroff-1902: I do not like placing anything except imports from submodules into `__init__.py` files and defining this bad design.

xmatthias: hm, then we should move backtesting / hyperopt out of that too... especially the hyperopt one goes "hand in hand" with this PR ...

hroff-1902: my other idea was to create a `commands` or `subcommands` dir and move every `start_*` function into a separate file (module) in it... Because `utils.py` gets bigger and bigger...

hroff-1902:
merge after hyperopt trailing space merged into develop;
add docs

hroff-1902:
Docs added, seems to be ready, pls review

hroff-1902:
Also contains resolution for #2614

freqtrade/optimize/hyperopt.py
```diff
@@ -202,49 +227,65 @@ def _params_update_for_json(self, result_dict, params, space: str):
...
+                                  print_colorized: bool) -> None:
+        """
+        Log results explanation string
+        """
+        explanation_str = Hyperopt._format_explanation_string(results, total_epochs)
+        # Colorize output
+        if print_colorized:
+            if results['total_profit'] > 0:
+                explanation_str = Fore.GREEN + explanation_str
+            if highlight_best and results['is_best']:
```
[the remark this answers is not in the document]
hroff-1902: ^^ "Compatibility with old hyperopt pickle file is broken, so first run hyperopt once for some small amount of epochs for testing this."

freqtrade/utils.py
```diff
@@ -354,3 +356,103 @@ def start_test_pairlist(args: Dict[str, Any]) -> None:
...
+
+    if trials and not no_details:
+        sorted_trials = sorted(trials, key=itemgetter('loss'))
+        results = sorted_trials[0]
+        Hyperopt.print_epoch_details(results, total_epochs, print_json, no_header)
+
+
+def start_hyperopt_show(args: Dict[str, Any]) -> None:
+    """
+    """
```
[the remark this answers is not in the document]
hroff-1902: simply forgotten to fill it

freqtrade/optimize/hyperopt.py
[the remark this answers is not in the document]
xmatthias: i did then it works.

That's ok - but then it shouldn't break like this with a KeyError i have no idea why it's happening,  but instead warn me about "your hyperopt file uses an older version ... can't support that - sorry..." ?

freqtrade/utils.py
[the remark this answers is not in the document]
xmatthias: that was my assumption ;)

freqtrade/optimize/hyperopt.py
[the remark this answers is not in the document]
hroff-1902: because I changed the format of the data kept in the pickle file
old hyperopts do not contain some dict items stored by the new hyperopt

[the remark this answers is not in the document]
hroff-1902: I'll think about the message...

hroff-1902:
> The refactorings in hyperopt.py make it really hard to review 😞

Yeah, and I'm not 100% sure about correctness of the merge, in a couple of places it required some additional changes to make them working together...

That was a bad idea to make that refactoring while this one is open...

freqtrade/optimize/hyperopt.py
```diff
@@ -147,48 +149,71 @@ def save_trials(self, final: bool = False) -> None:
...
+            result['trailing'] = {p.name: params.get(p.name)
+                                  for p in self.hyperopt_space('trailing')}
+
+        return result
 
-        results = sorted(self.trials, key=itemgetter('loss'))
-        best_result = results[0]
-        params = best_result['params']
-        log_str = self.format_results_logstring(best_result)
+    @staticmethod  # noqa: C901
```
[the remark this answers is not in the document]
hroff-1902: that's an example of errorneously merged code...

freqtrade/optimize/hyperopt.py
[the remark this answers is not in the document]
hroff-1902: added raising an exception, pls test with an old file (I need to create it with an older version of the bot, do not have any archived)

hroff-1902:
travis-ci failed with
`Using --random-order-seed=834267`

freqtrade/optimize/hyperopt.py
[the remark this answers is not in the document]
xmatthias: yep - works ... :+1:

freqtrade/utils.py
```diff
@@ -354,3 +356,105 @@ def start_test_pairlist(args: Dict[str, Any]) -> None:
             print(rapidjson.dumps(list(pairlist), default=str))
         else:
             print(pairlist)
+
+
+def start_hyperopt_list(args: Dict[str, Any]) -> None:
+    """
+    List hyperopt epochs previously evaluated
+    """
+    from freqtrade.optimize.hyperopt import Hyperopt
```
xmatthias: Just noticed that neither of these functions are touched by tests.

We don't need full coverage - especially not for utils. 

but i think having a quick "does this work if i call it" test will be useful and is probably required - considering that hyperopt is a class with frequent changes (and it won't stop - reading your ideas - which is a good thing). 

currently, we would need to do manual tests of both of these functions whenever we touch the hyperopt class.
I'm 100% certain that we'll forget that sooner (rather than later) and these functions may end up similar to the scripts we used to have - which were broken whenever we tried to use them - which i think we should aim to avoid.

hroff-1902:
Tests failed: https://travis-ci.org/freqtrade/freqtrade/jobs/622441084?utm_medium=notification&utm_source=github_status
(no config.json)

What do you think, which way should it be fixed?

freqtrade/configuration/arguments.py
```diff
@@ -55,8 +55,14 @@
...
 
+ARGS_HYPEROPT_LIST = ["hyperopt_list_best", "hyperopt_list_profitable", "print_colorized",
+                      "print_json", "hyperopt_list_no_details"]
+
+ARGS_HYPEROPT_SHOW = ["hyperopt_list_best", "hyperopt_list_profitable", "hyperopt_show_index",
+                      "print_json", "hyperopt_show_no_header"]
+
 NO_CONF_REQURIED = ["download-data", "list-timeframes", "list-markets", "list-pairs",
-                    "plot-dataframe", "plot-profit"]
+                    "hyperopt_list", "hyperopt_show", "plot-dataframe", "plot-profit"]
```
xmatthias: these should be the command names ... 

Fixing this should also get rid of the test error.

However - to test without a config, i would also use the following snippet - to avoid using a config if it exists, causing random test-failures. We use this in other tests that don't require a config (download-datadir, ...) within `test_utils.py`.

```python
    pargs = get_args(args)
    pargs['config'] = None
```

hroff-1902: Thanks, fixed.

freqtrade/utils.py
```diff
@@ -354,3 +356,105 @@ def start_test_pairlist(args: Dict[str, Any]) -> None:
...
+
+    trials = _hyperopt_filter_trials(trials, only_best, only_profitable)
+
+    n = config.get('hyperopt_show_index', -1)
+    if n > total_epochs:
+        raise OperationalException(
+                f"The index of the epoch to show should be less than {total_epochs + 1}.")
+    if n < -total_epochs:
+        raise OperationalException(
+                f"The index of the epoch to showshould be greater than {-total_epochs - 1}.")
```
xmatthias: a space would like to sit between "show" and "should" ...

PR #2606, opened 2019-12-03
Subcommand: test-pairlist

### LINKED ISSUE #935: Backtest with dynamic-whitelist?
Is it possible to backtest using --dynamic-whitelist?

xmatthias:
## Summary
Configuring pairlists can be quite complex, and different sequences of configurations can generate different results. 
To test these quickly, a new subcommand, `freqtrade test-pairlist` will be introduced.

This will also solve #935 (althought through a manual step) to use a dynamically generated pairlist for backtesting. For that, simply use `freqtrade test-pairlist` and insert the output as pair_whitelist into the configuration.

Closes #935

## Quick changelog

- new subcommand
- Document usage of this command

### DIFF
--- a/docs/utils.md
+++ b/docs/utils.md
@@ -234,3 +234,35 @@ $ freqtrade -c config_binance.json list-pairs --all --base BTC ETH --quote USDT 
 ```
 $ freqtrade list-markets --exchange kraken --all
 ```
+
+## Test pairlist
+
+Use the `test-pairlist` subcommand to test the configuration of [dynamic pairlists](configuration.md#pairlists).
+
+Requires a configuration with specified `pairlists` attribute.
+Can be used to generate static pairlists to be used during backtesting / hyperopt.
+
+### Examples
+
+Show whitelist when using a [dynamic pairlist](configuration.md#pairlists).
+
+```
+freqtrade test-pairlist --config config.json --quote USDT BTC
+```
+
+```
+usage: freqtrade test-pairlist [-h] [-c PATH]
+                               [--quote QUOTE_CURRENCY [QUOTE_CURRENCY ...]]
+                               [-1] [--print-json]
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -c PATH, --config PATH
+                        Specify configuration file (default: `config.json`).
+                        Multiple --config options may be used. Can be set to
+                        `-` to read config from stdin.
+  --quote QUOTE_CURRENCY [QUOTE_CURRENCY ...]
+                        Specify quote currency(-ies). Space-separated list.
+  -1, --one-column      Print output in one column.
+  --print-json          Print list of pairs or market symbols in JSON format.
+```
--- a/docs/utils.md
+++ b/docs/utils.md
@@ -43,20 +43,6 @@ The file will be named inline with your class name, and will not overwrite exist
 
 Results will be located in `user_data/strategies/<strategyclassname>.py`.
 
-### Sample usage of new-strategy
-
-```bash
-freqtrade new-strategy --strategy AwesomeStrategy
-```
-
-With custom user directory
-
-```bash
-freqtrade new-strategy --userdir ~/.freqtrade/ --strategy AwesomeStrategy
-```
-
-### new-strategy complete options
-
 ``` output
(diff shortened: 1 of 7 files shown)


docs/utils.md
```diff
@@ -234,3 +234,35 @@ $ freqtrade -c config_binance.json list-pairs --all --base BTC ETH --quote USDT
...
 ```
+
+## Test pairlist
+
+Use the `test-pairlist` subcommand to test the configuration of [dynamic pairlists](configuration.md#pairlists).
+
+Requires a configuration with specified `pairlists` attribute.
+Can be used to generate static pairlists to be used during backtesting / hyperopt.
+
+### Examples
```
hroff-1902: examples should probably be below usage block

freqtrade/utils.py
```diff
@@ -322,3 +322,37 @@ def start_list_markets(args: Dict[str, Any], pairs_only: bool = False) -> None:
...
+        config['stake_currency'] = curr
+        # Do not use ticker_interval set in the config
+        pairlists = PairListManager(exchange, config)
+        pairlists.refresh_pairlist()
+        results[curr] = pairlists.whitelist
+
+    for curr, pairlist in results.items():
+        print(f"Pairs for {curr}: ")
+        summary_str = ""
+        if args.get('print_list', False):
```
hroff-1902: there is no 'print-list' option in use for this command, it's always False...

freqtrade/utils.py
```diff
@@ -322,3 +322,37 @@ def start_list_markets(args: Dict[str, Any], pairs_only: bool = False) -> None:
...
+    results = {}
+    for curr in quote_currencies:
+        config['stake_currency'] = curr
+        # Do not use ticker_interval set in the config
+        pairlists = PairListManager(exchange, config)
+        pairlists.refresh_pairlist()
+        results[curr] = pairlists.whitelist
+
+    for curr, pairlist in results.items():
+        print(f"Pairs for {curr}: ")
```
hroff-1902: This (human-readable) line should not be printed for `-1` format, it's machine-readable for easy iteration through the pairs.
I also not sure for `--print-json`, maybe this line should not also be printed in this case...

docs/utils.md
xmatthias: I'm not sure ... why exactly? 
examples are most of the time more important as they show you how to "really" use the command for certain cases.

To me, the "full usage"  block is just there for completeness - for many commands which have many options (list-pairs for example, but there are others, too) the examples are better / more useful- since they show you how to combine the options - and don't just make you read the help strings ...

freqtrade/utils.py
xmatthias: i'm not certain about this ... 
yes, it should be machine readable (both, actually) - but simply chaining pairs together without any separation between the quote currencies (assuming more than one is used) seems odd to me, too.

hroff-1902: it will not be easy to iterate through the results, the header line should be additionally filtered out.
I assume json format is more for copy-paste into the pairlist file by a human, so it's probably not necessary for `--print-json` format; `-1` is more for script automation, the header will add the need to filter out this line in the scripts, instead of simple iterating through the result in the output.

xmatthias: well i'm not sure - i find the -1 format very difficult to use (also for the other commands) - even in scripts.

using the json output and using `jq` seems simpler... 

I'll think of something.

docs/utils.md
hroff-1902: 1) because you first need to give the user some introduction of all options available before giving examples. Have a look at the man pages -- they first give the list of all options, then their extended descriptions (examples can be inserted there as well) and only then some more examples are placed, when the reader have knowledge of all the options
2) this is how info is placed in the Start the Bot and Plotting parts -- some general introduction (with probably a general example, that example has no target to describe the options, just to give some general understanding about the command), then the list of options and the examples follow then. The page on `new-strategy` should also be reformatted this way, to have examples after the list of options.

freqtrade/utils.py
hroff-1902:
```
for p in `feqtrade test-pairlist -1`; do echo "do something useful with pair $p"; done
```

As I said above, for json it's questionable, so let's leave it as it is.

PR #2607, opened 2019-12-03
Move docs on loss function creation to a separate doc file

hroff-1902:
The docs on Hyperopt will be better structured and easier for an ordinal user if we move some sophisticated parts that are rarely needed and require higher coding skills and Python knowledge to a new Advanced Hyperopt chapter. (Only one such part for now actually -- the section describing the Creation of a custom loss function)

### DIFF
--- a/docs/advanced-hyperopt.md
+++ b/docs/advanced-hyperopt.md
@@ -1,15 +1,15 @@
 # Advanced Hyperopt
 
-This page explains some advanced Hyperopt issues that may require higher
+This page explains some advanced Hyperopt topics that may require higher
 coding skills and Python knowledge than creation of an ordinal hyperoptimization
 class.
 
 ## Creating and using a custom loss function
 
 To use a custom loss function class, make sure that the function `hyperopt_loss_function` is defined in your custom hyperopt loss class.
-For the sample below, you then need to add the command line parameter `--hyperopt-loss SuperDuperHyperOptLoss` to your hyperopt call so this fuction is being used.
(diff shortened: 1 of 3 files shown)


docs/advanced-hyperopt.md
```diff
@@ -0,0 +1,63 @@
+# Advanced Hyperopt
+
+This page explains some advanced Hyperopt issues that may require higher
```
xmatthias: > This page explains some advanced Hyperopt issues that ...

i would not call it "issues" - wanting to customize a loss function should not start (nor end) in an issue ...
more like "advanced Hyperopt usecases" - or "advanced Hyperopt applications" (?) ... 
Not necessarily one of the 2 above if you have a better idea...

hroff-1902:
This section need some more love, to be reworked a bit, at this time I specifically simply moved this 1:1 so that other changes will be visible in the changes. I'll apply your notes above 👍 The rest is later in a sep. PR (I need a cleaner Hyperopt page to instert the descriptions for new `hyperopt-list` and `hyperopt-show` commands, that's why this PR goes first)

docs/advanced-hyperopt.md
hroff-1902: pls review my suggestion for this

xmatthias:
> I need a cleaner Hyperopt page to insert the descriptions for new hyperopt-list and hyperopt-show commands, that's why this PR goes first

Wouldn't you want to document these in utils.md and only link to them from hyperopt?
that's how we're doing it with other utility commands, anyway ...


ps: i changed the "visible part" of the link back ... users will find the file in user_data once create-datadir is called - that's how we did it with other files that moved to templates, too. 
users will not find `freqtrade/templates/` if installed via pypi ... ;)

PR #2618, opened 2019-12-05
Fix hyperopt with ticker_interval from strategy

xmatthias:
## Summary
We support loading ticker_interval from the strategy - so i have it in almost none of my configurations.
However it fails with hyperopt - since the strategy is loaded after the hyperopt file.
By simply reversing the order of initialization between the custom hyperopt file and backtesting, we can support this again.

## Quick changelog
- fix bug
- add test ensuring we're not reintroducing this


hroff-1902:
cannot find the button to re-run the travis tasks (logged into travis), it was somewhere, cannot find it this time...

xmatthias:
right top corner ... :)

PR #2624, opened 2019-12-07
handle and document ROI=-1

### LINKED ISSUE #2277: Exit after n minutes/n candles and roi = 0
Hi there,

1/ For testing purpose of a strategy i would like to do something like (test what happens if i close my trade after 30 minutes)

minimal_roi = {'30':-1}
stop_loss = -1

image

Freqtrade is not happy with that.

2/ this type of problem appear too when using 

minimal_roi = {'0': 0.02, '30': 0}
stoploss = -0.02

image

14% of the trades have 0% roi but it is not possible because when you close the trade 30 minutes with profit > 0 to must be take into consideration.

I think in live he don't have this type of problem ?

Have a nice day.

xmatthias:
## Summary
This adds speific handling for ROI=-1 (using close when the roi-entry falls on a candle open and ROI = -1.
We should document ROI=-1.

Closes #2277 
## Quick changelog

- document usage of ROI=-1 as forcesell method
- Refactor `_get_close_rate()` out of `_get_sell_trade_entry()`
- use open if the forcesell entry is a candle open, else low (as before).

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -267,7 +267,7 @@ class Backtesting:
             # Set close_rate to stoploss
             closerate = trade.stop_loss
         elif sell.sell_type == (SellType.ROI):
-            roi = self.strategy.min_roi_reached_entry(trade_dur)
+            roi_entry, roi = self.strategy.min_roi_reached_entry(trade_dur)
             if roi is not None:
                 # - (Expected abs profit + open_rate + open_fee) / (fee_close -1)
                 closerate = - (trade.open_rate * roi + trade.open_rate *
@@ -275,8 +275,14 @@ class Backtesting:
 
                 # Use the maximum between closerate and low as we
                 # cannot sell outside of a candle.
-                # Applies when using {"xx": -1} as roi to force sells after xx minutes
+                # Applies when a new ROI setting comes in place and the whole candle is above that.
                 closerate = max(closerate, sell_row.low)
+                if roi == -1 and roi_entry % self.timeframe_mins == 0:
+                    # When forceselling with ROI=-1, the roi-entry will always be "on the entry".
+                    # If that entry is a multiple of the timeframe (so on open)
+                    # - we'll use open instead of close
+                    closerate = max(closerate, sell_row.open)
+
             else:
                 # This should not be reached...
                 closerate = sell_row.open
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -262,13 +262,22 @@ class Backtesting:
         return ticker
 
     def _get_close_rate(self, sell_row, trade: Trade, sell, trade_dur) -> float:
+        """
+        Get close rate for backtesting result
+        """
         # Special handling if high or low hit STOP_LOSS or ROI
         if sell.sell_type in (SellType.STOP_LOSS, SellType.TRAILING_STOP_LOSS):
             # Set close_rate to stoploss
-            closerate = trade.stop_loss
+            return trade.stop_loss
         elif sell.sell_type == (SellType.ROI):
             roi_entry, roi = self.strategy.min_roi_reached_entry(trade_dur)
             if roi is not None:
+                if roi == -1 and roi_entry % self.timeframe_mins == 0:
+                    # When forceselling with ROI=-1, the roi-entry will always be "on the entry".
+                    # If that entry is a multiple of the timeframe (so on open)
+                    # - we'll use open instead of close
+                    return sell_row.open
+
                 # - (Expected abs profit + open_rate + open_fee) / (fee_close -1)
                 closerate = - (trade.open_rate * roi + trade.open_rate *
                                (1 + trade.fee_open)) / (trade.fee_close - 1)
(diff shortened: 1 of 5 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -267,16 +267,22 @@ def _get_close_rate(self, sell_row, trade: Trade, sell, trade_dur) -> float:
...
                 # Use the maximum between closerate and low as we
                 # cannot sell outside of a candle.
-                # Applies when using {"xx": -1} as roi to force sells after xx minutes
+                # Applies when a new ROI setting comes in place and the whole candle is above that.
                 closerate = max(closerate, sell_row.low)
+                if roi == -1 and roi_entry % self.timeframe_mins == 0:
+                    # When forceselling with ROI=-1, the roi-entry will always be "on the entry".
+                    # If that entry is a multiple of the timeframe (so on open)
+                    # - we'll use open instead of close
+                    closerate = max(closerate, sell_row.open)
```
hroff-1902: should not this be applied for all roi, not only for -1?
(use open rate if the key of the entry is the multiple of timeframe)

xmatthias: not in my opinion, no. it would taint ROI to be more optimistic than what it should be.

see https://github.com/freqtrade/freqtrade/issues/2277#issuecomment-562853860 for a detailed reasoning (after writing that comment, i decided to do it this way).

hroff-1902: well, assume we have roi = {"30": -1}, this sets the "vertical barrier" for the trade -- it will be force sold in 30 min (6 candles for 5m timeframe). If the time is the beginning of the candle, now backtesting will assume that the sell happens using the open rate for the 6th candle, which is okay (because it's time of the beginning of this candle and the sell happens in real conditions using a price near the open rate for this candle).
I do not see any difference, however, for the case when another roi, different from -1, is used. First of all, if it's set to, say, -0.90, it also defines same forced vertical barrier (except for the cases when the price is dropped more, than 90%, let's not consider these cases here for simplicity. So with this setting, the trades will be sold in exactly same conditions -- at the beginning of the candle, so a price near (for backtesting -- equal) to the open rate of the candle should be used.
Then, for positive values, if we have roi = {"0": 0.05, "30": 0.02, "60": -1}, if we have open rate for the candle starting at 30th min, and previous candles do not reach 0.05 profit, it's obvious that the trade should be sold due to roi at the beginning of this candle, just at the beginning of this candle, so the open rate should also be used, not the low price.

hroff-1902: btw, if you remove `roi == -1 and ` part of the `if` condition and backtest it (DefaultStrategy, with min roi set as it's there), the result will be worse. So this is not 'more optimistic'.
I guess it's because closerate is calculated including fee into consideration, so it's often higher than sell_row.open... I think there is something wrong here.

xmatthias: > First of all, if it's set to, say, -0.90, it also defines same forced vertical barrier (except for the cases when the price is dropped more, than 90%, let's not consider these cases here for simplicity

I assume to only support ONE negative value, which is -1 - which is "forcesell". -90% is not (that's basically a strange stoploss) ... which i don't think we should support.

> Then, for positive values, if we have roi = {"0": 0.05, "30": 0.02, "60": -1}, if we have open rate for the candle starting at 30th min to be more than 2% above the buy rate, and previous candles do not reach 0.05 profit, it's obvious that the trade should be sold due to roi at the beginning of this candle, just at the beginning of this candle, so the open rate should also be used, not the low price.

Please look at the example in the issue (https://github.com/freqtrade/freqtrade/issues/2277#issuecomment-562853860).
In your opinion, using "open" would be correct in that case?

> btw, if you remove roi == -1 and part of the if condition and backtest it (DefaultStrategy, with min roi set as it's there), the result will be worse. So this is not 'more optimistic'.

Well that's an pointless test in my opinion. I'm sure i can remove part of many comparisons and get different results. 
The return-value within this comparison does not consider fees, nor does it consider that the roi sell may have triggered within the candle due to "high" being higher than x%.
Just because the result is worse does not mean it's closer to reality - choosing the "conservative" path is what we try - but we need to judge if it makes sense.

Think about it - using your example from before (`roi = {"0": 0.05, "30": 0.02, "60": -1},`):

You're assumption (in the test) is that >2% was reached in the 25min candle - so 2% on 30mins should trigger the sell on open (in your opinion) - since you assume the open is > 2%. however, open could be negative (profit) at that point, but high is > 2% - so we use the ROI value as sell (2%).

I don't think the current approach is wrong - but using open in this specific case when it's > roi might *improve* results (but this will NOT be archived by removing part of equations).

xmatthias: @hroff-1902 please have a look at it now ...

hroff-1902: I saw and I still think it's not correct enough (will create pictures)

hroff-1902:
LGTM 👍

hroff-1902:
Hmm, how could this be?
(no self.timeframe_mins in the Backtesting class)

gaugau3000:
Many thanks i know it this implementation is not really a back test need but a kind of stress test to identify best entries !

PR #2627, opened 2019-12-07
Docker: Build latest along with develop image

xmatthias:
## Summary
DOcker image latest has not been pushed anymore.

[dockerhub](https://hub.docker.com/r/freqtradeorg/freqtrade/tags).

We used to push latest along with develop, and we should continue to do that.

(100% correct functioning can only be tested once it's merged, since we only run deploy then) - but i'm confident that it works, since the develop-tag is pushed correctly.

### DIFF
--- a/build_helpers/publish_docker.sh
+++ b/build_helpers/publish_docker.sh
@@ -38,7 +38,7 @@ if [ $? -ne 0 ]; then
 fi
 
 # Tag as latest for develop builds
-if [ "${GITHUB_REF}" = "develop" ]; then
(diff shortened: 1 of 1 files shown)


hroff-1902:
Was it due to the following commit: https://github.com/freqtrade/freqtrade/commit/ff1d36434d7f66e53f81c01bbb5fa5be359f0761#diff-51270dc6d3ee4c9814a582fb8f65fa96
?

xmatthias:
yes - suspect so

PR #2642, opened 2019-12-10
Fix generation of hyperopt trailing params

### LINKED ISSUE #2617: Trailing stop hyperopt invalid results
##  Describe your environment

  * Operating system: linux
  * Python Version: 3.7.5
  * Branch:  Develop
  * Last Commit ID: 8dd9b5c6fba973b98defc0f9a9e2e3c01af735dd
  
## Describe the problem:

Depending on the strategy, hyperopt may produce invalid results.

My trailing hyperopt just returned the following result.
However, configuring trailing_stop_positive > trailing_stop_positive_offset is not possible (we prevent this as part of config validation).

```
Stoploss:
{'stoploss': -0.34474}
Trailing stop:
{   'trailing_only_offset_is_reached': True,
    'trailing_stop': True,
    'trailing_stop_positive': 0.33186,
    'trailing_stop_positive_offset': 0.0542
}

```

Leaving out the irrationally high numbers my stoplosses reach... this configuration results in `The config trailing_stop_positive_offset needs to be greater than trailing_stop_positive in your config.` (also shown on slack).

### Further thoughts

Now i'm not 100% certain if the error is with hyperopt thinking about it.

The current validation we have in place ensures that "positive stoploss" cannot lead back into negative values. (5% offset and trailing_stop_positive at 10% could be a -5% sell).
However, I think it might be a valid case to say that a stoploss is -10% initially, and once the trade reaches 2% profit, it should be pulled tighter - to 5%. 
This could lead to -3% loss - but on the other hand it is not 10% loss anymore.

hroff-1902:
Fixes #2617.

Also, limits (boundaries) for dimensions changed:
```
            Real(0.01, 0.35, name='trailing_stop_positive'),
```
-- min boundary is 0.02 in develop;
and
```
Real(0.001, 0.1, name='trailing_stop_positive_offset_p1')
```
-- this is the new dimension for the difference between `trailing_stop_positive_offset` and `trailing_stop_positive`. Thus, the `trailing_stop_positive_offset` can be in range '0.011...0.45'. Pls review it.

TODO:
* [x] Adjust tests.

### DIFF
--- a/freqtrade/optimize/hyperopt_interface.py
+++ b/freqtrade/optimize/hyperopt_interface.py
@@ -175,6 +175,19 @@ class IHyperOpt(ABC):
         ]
 
     @staticmethod
+    def generate_trailing_params(params: Dict) -> Dict:
+        """
+        Create dict with trailing stop parameters.
+        """
+        return {
+            'trailing_stop': params['trailing_stop'],
+            'trailing_stop_positive': params['trailing_stop_positive'],
+            'trailing_stop_positive_offset': (params['trailing_stop_positive'] +
+                                              params['trailing_stop_positive_offset_p1']),
+            'trailing_only_offset_is_reached': params['trailing_only_offset_is_reached'],
+        }
+
+    @staticmethod
     def trailing_space() -> List[Dimension]:
         """
         Create a trailing stoploss space.
@@ -190,8 +203,15 @@ class IHyperOpt(ABC):
             # other 'trailing' hyperspace parameters.
             Categorical([True], name='trailing_stop'),
 
-            Real(0.02, 0.35, name='trailing_stop_positive'),
-            Real(0.01, 0.1, name='trailing_stop_positive_offset'),
+            Real(0.01, 0.35, name='trailing_stop_positive'),
+
+            # 'trailing_stop_positive_offset' should be greater than 'trailing_stop_positive',
+            # so this intermediate parameter is used as the value of the difference between
+            # them. The value of the 'trailing_stop_positive_offset' is constructed in the
+            # generate_trailing_params() method.
+            # # This is similar to the hyperspace dimensions used for constructing the ROI tables.
+            Real(0.001, 0.1, name='trailing_stop_positive_offset_p1'),
+
             Categorical([True, False], name='trailing_only_offset_is_reached'),
         ]
 
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -177,8 +177,7 @@ class Hyperopt:
             result['stoploss'] = {p.name: params.get(p.name)
                                   for p in self.hyperopt_space('stoploss')}
         if self.has_space('trailing'):
-            result['trailing'] = {p.name: params.get(p.name)
-                                  for p in self.hyperopt_space('trailing')}
+            result['trailing'] = self.custom_hyperopt.generate_trailing_params(params)
 
         return result
 
@@ -353,13 +352,13 @@ class Hyperopt:
             self.backtesting.strategy.stoploss = params_dict['stoploss']
 
         if self.has_space('trailing'):
-            self.backtesting.strategy.trailing_stop = params_dict['trailing_stop']
-            self.backtesting.strategy.trailing_stop_positive = \
-                params_dict['trailing_stop_positive']
+            d = self.custom_hyperopt.generate_trailing_params(params_dict)
+            self.backtesting.strategy.trailing_stop = d['trailing_stop']
+            self.backtesting.strategy.trailing_stop_positive = d['trailing_stop_positive']
             self.backtesting.strategy.trailing_stop_positive_offset = \
-                params_dict['trailing_stop_positive_offset']
+                d['trailing_stop_positive_offset']
             self.backtesting.strategy.trailing_only_offset_is_reached = \
-                params_dict['trailing_only_offset_is_reached']
+                d['trailing_only_offset_is_reached']
 
         processed = load(self.tickerdata_pickle)
 
--- a/tests/optimize/test_hyperopt.py
+++ b/tests/optimize/test_hyperopt.py
@@ -636,7 +636,7 @@ def test_generate_optimizer(mocker, default_conf) -> None:
         'stoploss': -0.4,
         'trailing_stop': True,
         'trailing_stop_positive': 0.02,
-        'trailing_stop_positive_offset': 0.1,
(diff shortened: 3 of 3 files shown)


hroff-1902:
> Pls review it.

I mean here, pls review separately the ranges for those two parameters, `trailing_stop_positive` and `trailing_stop_positive_offset`...

freqtrade/optimize/hyperopt_interface.py
```diff
@@ -190,8 +203,15 @@ def trailing_space() -> List[Dimension]:
             # other 'trailing' hyperspace parameters.
             Categorical([True], name='trailing_stop'),
 
-            Real(0.02, 0.35, name='trailing_stop_positive'),
-            Real(0.01, 0.1, name='trailing_stop_positive_offset'),
+            Real(0.01, 0.35, name='trailing_stop_positive'),
```
xmatthias: Letme ask a general question about this (applicable also for the general stoploss space)

assuming this space `Real(0.01, 0.35)` ... this space can take on any point between 0.01 and 0.35 (we cannot specify the precision / step in a Real Space - as far as I've seen).

So hyperopt may spend a ton of time optimizing between `0.05444` and `0.05333` - which has no practical difference to results after the 3rd decimal (so this could be optimized in 2 runs, `0.054` and `0.053` - one of which will be better, one worse).

I ask this because I've seen hyperopt results with a ton of decimals in the result ... but really anything with a step-size< 0.1% will not make sense.

Could we instead define an Integer space, between 10 and 350 - (which has an exact number of steps between each other - 340) - and divide this parameter by1000 (so 10 becomes 0.01, 350 -> 0.35).
The space would be identical, but instead of having unlimited possibilities between the 2 specified limits, we have an exact number. 

I would expect great speedups (without having tried this) using hyperopt for stoploss / trailing stop spaces.

hroff-1902: > Letme ask a general question about this first (applicable also for the general stoploss space)

👍 

> assuming this space `Real(0.01, 0.35)` ... this space can take on any point between 0.01 and 0.35 (we cannot specify the precision / step in a Real Space - as far as I've seen).

Yes, right.

> So hyperopt may spend a ton of time optimizing between `0.05444` and `0.05333` - which has no practical difference to results after the 3rd decimal (so this could be optimized in 2 runs, `0.054` and `0.053` - one of which will be better, one worse).

In practice it's not this way: first, the DefaultHyperOpt, for example, has 27 dimensions, it has wide range of opportunities to vary the points in the hyperspace, so it should not 'stick' to some short subrange and try lot's of points in this one dimension. It can, however, try 0.05444 at one epoch and, say, 0.05442 at another, but these epochs will also have other parameters varied, so the result will be close (or even identical) to the use of same 0.0544 value at each of these epochs.

Moreover, skopt measures the distance between points and warn the user about close points seen multiple times. The minimal distance is now hardcoded as 1e-8 (is a sum of distances at all dimensions). You can see these messages when optimizing for 'stoploss' space only -- the search space is "tight", it only has one dimension in this case.

If a kind of Decimal(precision: int) space will be used, optimizer will quickly exhaust all possible values as well.

> I ask this because I've seen hyperopt results with a ton of decimals in the result ... but really anything with a step-size< 0.1% will not make sense.

Yes, because it's python floats. And that's why we round the floats for humans. The raw values are printed in json. 

> Could we instead define an Integer space, between 10 and 350 - (which has an exact number of steps between each other - 340) - and divide this parameter by1000 (so 10 becomes 0.01, 350 -> 0.35).
> The space would be identical, but instead of having unlimited possibilities between the 2 specified limits, we have an exact number.
> 
> I would expect great speedups (without having tried this) using hyperopt for stoploss / trailing stop spaces.

Although, this may make a sense. I'll try to create it either this way you proposed (with integers and the _generator methods to create float params from the integers) or maybe by introducing the Decimal class of skopt spaces (as an additional class, I'm not yet going to change the skopt itself and support it then...)

This also may make sense for 1) Plotting the results in chosen dimensions, visualization would be great. In both tabular and graphic forms. 2) Imploying grid search option in the future.

Not in this PR, however. It's float as all other 'standard' float dimensions -- stoploss, roi steps.

For the ranges of the parameters, currently implemented here: 
* 0.35 was used with as a reference to the stoploss dimension. Should it be 0.25, to have the boundary of 0.35 for the offset parameter (0.25+0.10)? Now the upper boundary for trailing offset is 0.35+ 0.10. That's the main questionable point here, anything else is just minor.
* Is 0.10 okay and the upper boundary for the difference between trailing positive offset and trailing positive?
* Is 0.001 is precise enough and the low boundary for this difference?
* 0.01 was taken instead of 0.02 because 1) offset difference is added, so offset is greater than it; 2) I'm thinking of changing min boundaries of other 'stanard' dimensions -- stoploss, roi steps -- to 0.01 as well from some other higher limits now implemented (we have an issue about roi steps for this)

xmatthias: > In practice it's not this way: first, the DefaultHyperOpt, for example, has 27 dimensions, it has wide range of opportunities to vary the points in the hyperspace, so it should not 'stick' to some short subrange and try lot's of points in this one dimension.

you assume to run all spaces, my assumption is to run only a subset of spaces - `--spaces stoploss ` for example - where you only have exactly one parameter to tune.
It's probably not a huge issue for "full" hyperopts, since it has enough other parameters to tune ... but tuning only stoploss / roi could run into this problem.

> Not in this PR, however. It's float as all other 'standard' float dimensions -- stoploss, roi steps.

that's not what I've been suggesting, it just "jumped my mind" when looking at these lines.

> For the ranges of the parameters, currently implemented here:

I find this very difficult to say "just like that" ... experiences from many many runs with many different strategies will (should) show  ... so we can finetune them in the future eventually

> maybe by introducing the Decimal class of skopt spaces (as an additional class, I'm not yet going to change the skopt itself and support it then...)

I'm not sure why this should be necessary ... (and it would introduce confusion for users where to import stuff from ...).
why not use integers and divide the parameter within `generate_trailing_params()` (for example) by the multiplication factor (1000 in my example ...) ? that would serve the same function than the other implementations ... so we would basically just "right-shift" the decimal point for the run - and use integers to generate the points.

the approach is not that different from the constraint applied to `trailing_stop_positive_offset` now ...

hroff-1902: The implementation will show pros- and contras for both variants... Thanks for idea with integers. 👍

xmatthias: that was my intend .... :laughing: :+1:

PR #2628, opened 2019-12-08
Telegram / sell duration

xmatthias:
## Summary
Show the duration of the trade together with the sell notifiction.

Trade-duration is something very interresting ... (did the bot make 2% in 1 day? in 1 hour? ...)

## Quick changelog

- Add Duration to sell-notification

PR #2643, opened 2019-12-11
Remove min (plural) from codebase

xmatthias:
## Summary
Plural of min is not mins ...

Also reformatted the hyperopt results fixture ... having lines > 10000 lines is very bad since most modern editors won't show more than 10k lines (as i discovered ...).
However i can't blame them - lines should be 100 lines - in this case more to avoid having thousands of lines ... but not 10k - if we'd ever find a problem within this fixture it's impossible to find the correct location!


hroff-1902:
To summarize the above in one place:

You know, I'm sometimes a kind of grammarnazi (wikisource editor background).

I simply hate (frequently found) news articles and manuals for electric tools with 'kW/h' written. They are written by editors that do not know what they are talking about, these are the girls (and guys, to be politcorrect 😉 ) with no technical background. These articles are simply wrong and manuals are buggy. 'kW(center dot)h' is correct and 'kW-h', 'kWh' are acceptable.

I would personally write often 'hrs', but it's in informal sms/e-mail in English ("I'll be in 2 hrs"), and also probably "3 kgs of tomatos" in a recipe, but here it's a shortwriting, not a unit/symbol for the time unit. I would probably (personally) write "N min(s)", but this also in an informal message. So "mins" in the output/log messages, etc looks simply buggy for my grammarnazi eyes 🤓 , sorry.

A reference to #2628, where the discussion started, is needed here; thus inserted.

hroff-1902:
Thanks for reformatting that fixture stuff, I simply inserted it from by vi as it was printed...

Btw, should not it be a function-wide fixture (whatever it means in pytest infrastructure, I'm still not quite familiar with it)?

xmatthias:
well it depends - a fixture with scope function will reset with each new test-function, whereas this will be initialized once per test-session.

I didn't check in detail where / how it's used - but usually a "session"-wide fixture is OK unless you modify the fixture content itself (we do that with the `default_conf` dict in many tests).

since a dict (and a list) are mutable, we need to use function scoped fixtures so they are "fresh" with every call - but i don't think that this fixture is ever modified, otherwise i guess we'd have seen random test-failures (depending on the sequence tests run in).

hroff-1902:
https://github.com/freqtrade/freqtrade/blob/7c7ca1cb90f6f5468cf286920462b117a3b63f99/freqtrade/plot/plotting.py#L123-L124
-- should not there be a white space char before 'min'?

hroff-1902:
One suspicious point, else LGTM 👍 , approving...

PR #2651, opened 2019-12-13
Round amount to precision also for dry-runs

xmatthias:
## Summary
Round amount to precision also for dry-runs - so it's more realistic.

otherwise we can have amounts with 100ds of decimals - since amount-rounding happens within create_trade - which is never reached in dry-mode.

## Quick changelog

- apply Amount to precision for dry-run trades.


hroff-1902:
LGTM 👍

PR #2654, opened 2019-12-13
improve show config when using trailing stop 

xmatthias:
## Summary
Improve output of /show_config when trailing_stop is on.

We should not just show the stoploss, but also the parameters of the trailing stoploss.

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -587,12 +587,23 @@ class Telegram(RPC):
         :return: None
         """
         val = self._rpc_show_config()
+        if val['trailing_stop']:
+            sl_info = (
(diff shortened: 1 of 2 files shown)


hroff-1902:
Maybe print also the value of `trailing_stop` explicitly in both cases (True/False)?

Else is LGTM 👍

xmatthias:
> Maybe print also the value of trailing_stop explicitly in both cases (True/False)?

I'm not certain about that - it's an option i did consider but i assumed that by having all trailing options available it should be kinda clear ....

the downside is that it adds another line where it's not needed (when trailing_stop is false)...

hroff-1902:
okay, this can always be added later if this will be found as more convenient

PR #2655, opened 2019-12-14
Use first pair of pairlist to get fee

### LINKED ISSUE #2653: FTX issue when backtesting spot markets

yes

  * Operating system: linux
  * Python Version: Python 3.7.5rc1
  * CCXT version: ccxt==1.19.25 
  * Branch: Develop 
  * Last Commit ID: latest
  
When doing 
`freqtrade backtesting -c ftx.json --strategy-list chop ao ao2 ao3 --ticker-interval=4h`
using BTC/USD and FTT/USD markets with proper USD as stake currency, then an error pops out.

### Steps to reproduce:

  1. try to backtest using FTX as exchange, pair BTC/USD 
  2. Error pops out
  
### Observed Results:

  Error: 
```
2019-12-13 17:08:48,545 - freqtrade - ERROR - Fatal exception!
  File "/home/dante/freqtrade/freqtrade/main.py", line 36, in main
  File "/home/dante/freqtrade/freqtrade/optimize/__init__.py", line 43, in start_backtesting
    backtesting = Backtesting(config)
  File "/home/dante/freqtrade/freqtrade/optimize/backtesting.py", line 68, in __init__
    self.fee = self.exchange.get_fee()
  File "/home/dante/freqtrade/freqtrade/exchange/common.py", line 113, in wrapper
    return f(*args, **kwargs)
  File "/home/dante/freqtrade/freqtrade/exchange/exchange.py", line 932, in get_fee
    price=price, takerOrMaker=taker_or_maker)['rate']
  File "/home/dante/.local/lib/python3.7/site-packages/ccxt/base/exchange.py", line 1710, in calculate_fee
    market = self.markets[symbol]
KeyError: 'ETH/BTC'

```
I was expecting a simple backtest as usual.

xmatthias:
## Summary
get_fee for backtesting and edge should not use ETH/BTC, but the first pair in the pairlist instead.

closes #2653
## Quick changelog

- Send first pair from pairlist as symbol to `get_fee()`

### DIFF
--- a/freqtrade/edge/__init__.py
+++ b/freqtrade/edge/__init__.py
@@ -80,7 +80,7 @@ class Edge:
         if config.get('fee'):
             self.fee = config['fee']
         else:
-            self.fee = self.exchange.get_fee()
+            self.fee = self.exchange.get_fee(symbol=self.config['exchange']['pair_whitelist'][0])
 
     def calculate(self) -> bool:
         pairs = self.config['exchange']['pair_whitelist']
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -921,7 +921,7 @@ class Exchange:
             raise OperationalException(e) from e
 
     @retrier
-    def get_fee(self, symbol='ETH/BTC', type='', side='', amount=1,
+    def get_fee(self, symbol, type='', side='', amount=1,
                 price=1, taker_or_maker='maker') -> float:
         try:
             # validate that markets are loaded before trying to get fee
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -65,7 +65,7 @@ class Backtesting:
         if config.get('fee'):
             self.fee = config['fee']
         else:
-            self.fee = self.exchange.get_fee()
+            self.fee = self.exchange.get_fee(symbol=self.config['exchange']['pair_whitelist'][0])
 
         if self.config.get('runmode') != RunMode.HYPEROPT:
             self.dataprovider = DataProvider(self.config, self.exchange)
--- a/tests/edge/test_edge.py
+++ b/tests/edge/test_edge.py
@@ -334,7 +334,7 @@ def test_process_expectancy(mocker, edge_conf):
     edge_conf['edge']['min_trade_number'] = 2
     freqtrade = get_patched_freqtradebot(mocker, edge_conf)
 
-    def get_fee():
+    def get_fee(*args, **kwargs):
         return 0.001
 
     freqtrade.exchange.get_fee = get_fee
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -1646,10 +1646,10 @@ def test_get_fee(default_conf, mocker, exchange_name):
     })
     exchange = get_patched_exchange(mocker, default_conf, api_mock, id=exchange_name)
 
-    assert exchange.get_fee() == 0.025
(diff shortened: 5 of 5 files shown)


hroff-1902:
I would remove the defaults for attributes of get_fee. It's just a source of errors, wrong fee can frequently be taken.

xmatthias:
> I would remove the defaults for attributes of get_fee. It's just a source of errors, wrong fee can frequently be taken.

for all exchanges i used so far, the fees are identical per pair, so the pair we pass in does not really matter unless it does not exist.

the only difference is binance when using BNB as fee - but that's not covered by `ccxt.calculate_fee()` (nor are higher level tiers - if you happen to be in one of these).

It should be easy for Binance to expose this when authenticated - since the fee i'll pay are predictable - but I've looked through the binance API multiple times - but it's simply not part of any endpoint.

hroff-1902:
> > I would remove the defaults for attributes of get_fee. It's just a source of errors, wrong fee can frequently be taken.
> 
> for all exchanges i used so far, the fees are identical per pair, so the pair we pass in does not really matter unless it does not exist.
> 
> the only difference is binance when using BNB as fee - but that's not covered by `ccxt.calculate_fee()` (nor are higher level tiers - if you happen to be in one of these).
> 
> It should be easy for Binance to expose this when authenticated - since the fee i'll pay are predictable - but I've looked through the binance API multiple times - but it's simply not part of any endpoint.

For kraken, fee depends on the 30-days volume traded with this pair.
For bittrex, fee also depend on 30-days volume (do not remember, is it cumulative or not).

What's the reason for having `amount=1` for amount argument of `get_fee()`? If you need ratio instead of fee amount, there shall be 2 different methods, `get_fee_amount()` and `get_fee()` (or `get_fee_ratio()`). Using defaults that are not relevant to particular trade is simply the source of hardly detected errors. Same for defaults for maker/taker, other attributes. _Same for other methods in the exchange class._ (The latter is not for this PR of course)

xmatthias:
> For kraken, fee depends on the 30-days volume traded with this pair.

assuming this is accurately represented in the result from the exchange (either via `ccxt.calculate_fee()` - or as part of the market dictionary - which it's not for binance) - then we're unable to cache fees as we do now, but instead would have to call it within backtesting, once per pair since it could differ per pair. 

Keeping that in mind, hyperopt may not work anymore since last time i tried, it breaks if hyperopt is initialized with the exchange still in the backtesting class.

> What's the reason for having amount=1 for amount argument of get_fee()? 

What's the difference of having it optionally available in the function header, and passing it hardcoded to the function since it's a required argument?
It was put to the function header to have the future possibility to pass another value in (which we currently don't do - but we have the possibility to do so).
The alternative is to move the `amount=1` to the ccxt call - which i find worse.

```
calculate_fee() missing 1 required positional argument: 'amount'
```

The same goes for all other arguments apart from "taker / maker".
```
Signature:
ct.calculate_fee(
    symbol,
    type,
    side,
    amount,
    price,
    takerOrMaker='taker',
    params={},
)
```

>  Same for other methods in the exchange class. (The latter is not for this PR of course)

Nor are the points about the defaults ... 
Honestly, none of these points was a problem so far - keep in mind that this is a quick PR to fix a bug someone reported, not the pr introducing the `get_fee()` method.

If you see something wrong with how this is done, then nothing is stopping you from making a followup-pr trying to improve these things - but if we start with this type of (in my opinion toxic) review style - pointing out things which are not relevant for the bug at hand, we'll soon not merge anything since i'm sure i can find a line i conceptually don't like in any PR.

PR #2659, opened 2019-12-14
[critical]: Fix mypy errors in develop

hroff-1902:
Fix mypy error after #2624 
How could this happen with all our travis checks?

The problem actually is not the mypy by itself, but the error in the code where can fail when the users run backtesting...


hroff-1902:
Got it, that's bad. Will have it in the mind...

Is it possible to force running ci again for open PRs after each commit (merge) into develop?
Github internally detects it -- it shows the message about conflicts after each merge made into develop if results in conflicts with the current codebase in the PR...

PR #2661, opened 2019-12-15
Introduce Dry-Run Wallet

### LINKED ISSUE #2421: Introduce Dry run wallet
Hi,

Most of the volume in crypto move from XXX/BTC to XXX/USDX so have more than 999.9 dollar in the wallet is not a big deal.

I say that because i have this beautiful message during my dryrun because i missed this point (and of course i don't see it because i monitor my logs like a stupid idiot ) :-)

> WARNING - Unable to create trade: Available balance(999.9 USDT) is lower than stake amount(1000 USDT)

What do you think about it ?

Have a nice day.

xmatthias:
## Summary
Introduce Dry-Run Wallet.
This will enable simulation of the stake balance on the exchange - which will enhance accuracy, especially when using `stake_amount=unlimited`.

Closes #2421 

## Quick changelog

- Simulate wallet from a starting amount in stake currency.
- Enable simulated balance for RPC calls with explicit warning that this is experimental and simulated warnings.

## detailed explanation
The calculation is done by applying all profits of closed trades on top of the starting capital.

- Apply apply profits of closed trades on top of stake amount
- Subtract currently tied up stake_amount in open trades
- update balances for currencies currently in trades


hroff-1902:
Why have you removed WalletsDry class? There could be both WalletsExchange (current Wallet)/WalletsDry derived from abstract Wallet (WalletAbstract, WalletBase) which defines the interface...

hroff-1902:
To clarify the behavior and implementation:
* If Dry-Run fails or if I press Ctrl+C and then restart the bot -- will the bot use the starting value of the wallet, right? And it will proceed with the open (simulated) trades, will fetch them from the db, but start with the full wallet again, right?
* Would it makes sense to print this info (the finishing wallet amount) for backtesting as well?
* If simulated wallet goes to zero, will Dry-Run proceed buying?
* If stake_amount is set to unlimited. will the bot use this value (of the simulated balance) to calculate the amounts of the next order amount?

docs/configuration.md
```diff
@@ -149,6 +149,9 @@ In this case a trade amount is calculated as:
 currency_balance / (max_open_trades - current_open_trades)
 ```
 
+!!! Note "When using Dry-Run Mode"
+    When using `"stake_amount" : "unlimited",` in combination with Dry-Run, the balance will be simulated starting with a stake of `dry_run_wallet` which will evolve over time. It is therefore important to set `dry_run_wallet` to a sensible value, otherwise it may simulate trades with 100 BTC (or more) at once - which may not correspond to your real available balance.
```
hroff-1902: "It is therefore important to set `dry_run_wallet` to a sensible value **(like 0.05 or 0.01 for BTC and 1000 or 100 for USDT, for example)**, otherwise it may simulate trades with 100 BTC (or more) or 0.05 USDT (or less) at once - which may not correspond to your real available balance **or be less than the exchange minimal limit for the order amount for the stake currency**."

(proposed additions in bold)

xmatthias:
> Why have you removed WalletsDry class? There could be both WalletsExchange (current Wallet)/WalletsDry derived from abstract Wallet (WalletAbstract, WalletBase) which defines the interface...

I removed the class because i didn't see the point in it.
If you look at how it was before - then each `get_*()` call checked for dry-run and returned the constant eventually.
If we look at it now, then the only change is in `update()` - which updates it one way or the other.

It's a pattern we use throughout freqtrade - seperating dry/run and live with if statements.

Putting it back is not an issue - but i won't do it with a resolver as there'll be exactly 2 possibilities - it's either a WalletsExchange, or a WalletsDry - so the "if" we now have in `update()` will move to `freqtradebot.__init__`

hroff-1902:
Okay, fine! LGTM 👍 

The questions were here to get the overall understanding about the behavior (expected) after this PR.

PR #2671, opened 2019-12-16
Add documentation about ohlcv_partial_candle

### LINKED ISSUE #2202: Option to recalculate on every tick during live trading
Currently, Freqtrade fetches the latest compete candle for calculating values; however, there is always an incomplete candle, specifically hlcv values that are left out of the equation until the candle completes. 

As far as I understand, Freqtrade intentionally leaves this incomplete data out of the equation to avoid repainting issues, and because some exchanges do not provide incomplete data. 

That being said, some strategies can still work using incomplete data and the user should have the option (if available) to recalculate their algo's with the incomplete data.

One proposed solution to this issue would be resampling at a lower timeframe, but it is not ideal (added complexity to strategy, rate limiting).

In an ideal situation, this feature would be similar to tradingview's recalculate on every tick option show below.
Screen Shot 2019-08-30 at 14 09 42
Screen Shot 2019-08-30 at 14 09 58

xmatthias:
## Summary
Document how to get incomplete candles within a strategy (and warn about the dangers of these).

Closes #2202

## Quick changelog

- Add docs about `ohlcv_partial_candle`
- improve documentation how to discover incomplete candles from a exchange return value.

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -61,3 +61,24 @@ print(res)
 ```shell
 $ pip3 install web3
 ```
+
+### Send incomplete candles to the strategy
+
+Most exchanges return incomplete candles via their ohlcv / klines interface.
+By default, Freqtrade assumes that incomplete candles are returned and removes the last candle assuming it's an incomplete candle.
+
+Wether your exchange returns incomplete candles or not can be checked using [the helper script](developer.md#Incomplete-candles) from the Contributor documentation.
+
+If the exchange does return incomplete candles and you would like to have incomplete candles in your strategy, you can set the following parameter in the configuration file.
+
+``` json
+{
+
+    "exchange": {
+        "_ft_has_params": {"ohlcv_partial_candle": false}
+    }
+}
+```
+
+!!! Warning "Danger of repainting"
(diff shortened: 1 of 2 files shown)


hroff-1902:
Let's move the whole "Incomplete candles" section from the developer.md to the new "Advanced Strategy" md file (like Advanced Post-Install and Advanced Hyperopt were separated), because the tips are for the strategy creators, not the devs.

docs/exchanges.md
```diff
@@ -61,3 +61,24 @@ print(res)
 ```shell
 $ pip3 install web3
 ```
+
+### Send incomplete candles to the strategy
+
+Most exchanges return incomplete candles via their ohlcv / klines interface.
+By default, Freqtrade assumes that incomplete candles are returned and removes the last candle assuming it's an incomplete candle.
+
+Wether your exchange returns incomplete candles or not can be checked using [the helper script](developer.md#Incomplete-candles) from the Contributor documentation.
```
hroff-1902: "Whether"

xmatthias:
> Let's move the whole "Incomplete candles" section from the developer.md to the new "Advanced Strategy" md file (like Advanced Post-Install and Advanced Hyperopt were separated), because the tips are for the strategy creators, not the devs.

not entirely true - to use this you need to be quite good at python, otherwise your strategy will repaint.
Also, that would split up the developer documentation for a new exchange - since verifying this is a very important step in verifying a new exchange...

PR #2672, opened 2019-12-16
Minor: improvements in data/history.py

hroff-1902:
Small cleanup of code in this module: alignment of download_pair_history/download_trades_history and refresh_backtest_ohlcv_data/refresh_backtest_trades_data.

(Further they could/should be merged together to remove code duplication)

### DIFF
--- a/tests/data/test_history.py
+++ b/tests/data/test_history.py
@@ -14,9 +14,9 @@ from freqtrade import OperationalException
 from freqtrade.configuration import TimeRange
 from freqtrade.data import history
 from freqtrade.data.history import (_load_cached_data_for_updating,
+                                    _download_pair_history,
+                                    _download_trades_history,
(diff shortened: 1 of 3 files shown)


tests/data/test_history.py
```diff
@@ -14,9 +14,9 @@
 from freqtrade.configuration import TimeRange
 from freqtrade.data import history
 from freqtrade.data.history import (_load_cached_data_for_updating,
```
xmatthias: General sidenote (not for this PR).
This is a very inconsistent way to import ... we should either import history and use it's sub-properties - or import each needed value on it's own ...

hroff-1902: Completelt agree, I also noticed this and even started changing in tests in one of the stashed work...

xmatthias: hm - too late - already done ;) see #2675

PR #2676, opened 2019-12-17
Fix random test failure. 

xmatthias:
## Summary
Fix random test failure.

The 2 i've investigated did not occur anymore, unfortunately only time will tell if it's really fixed since even with the same seed, it's not always occuring.

todo:
- [x] remove hardcoded random-seed again once problem is solved


hroff-1902:
Hmm, the tests are so fragile...

LGTM 👍

PR #2678, opened 2019-12-17
Dataprovider history: split refresh part

hroff-1902:
... from `load_data()`

New `refresh_data()` function introduced, which is used in Edge (besides tests).
No exchange is to be passed (and checked for None) into `load_data()`

### DIFF
--- a/tests/data/test_history.py
+++ b/tests/data/test_history.py
@@ -21,6 +21,7 @@ from freqtrade.data.history import (_download_pair_history,
                                     pair_trades_filename,
                                     refresh_backtest_ohlcv_data,
                                     refresh_backtest_trades_data,
+                                    refresh_data, refresh_pair_history,
                                     trim_dataframe, trim_tickerlist,
                                     validate_backtest_data)
 from freqtrade.exchange import timeframe_to_minutes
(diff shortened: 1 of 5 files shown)


hroff-1902:
I personally more like interface in first commit: https://github.com/freqtrade/freqtrade/pull/2678/commits/60f89c8c019f6806b28b044e642f8cdab8c60cc9, without later optimization, where I removed redundant parts...

tests/edge/test_edge.py
```diff
@@ -290,6 +290,7 @@ def mocked_load_data(datadir, pairs=[], timeframe='0m', refresh_pairs=False,
 def test_edge_process_downloaded_data(mocker, edge_conf):
     freqtrade = get_patched_freqtradebot(mocker, edge_conf)
     mocker.patch('freqtrade.exchange.Exchange.get_fee', MagicMock(return_value=0.001))
+    mocker.patch('freqtrade.data.history.refresh_data', MagicMock())
```
hroff-1902: pls verify it, am I right with adding this into these tests?

tests/data/test_history.py
```diff
@@ -129,16 +130,17 @@ def test_load_data_with_new_pair_1min(ticker_history_list, mocker, caplog,
...
     assert file.is_file()
     assert log_has_re(
         'Download history data for pair: "MEME/BTC", timeframe: 1m '
         'and store in .*', caplog
     )
     with pytest.raises(OperationalException, match=r'Exchange needs to be initialized when.*'):
-        load_pair_history(datadir=testdatadir, timeframe='1m',
-                          refresh_pairs=True, exchange=None, pair='MEME/BTC')
+        refresh_data(datadir=testdatadir, timeframe='1m', pairs=['MEME/BTC'],
+                     exchange=None)
```
xmatthias: Not "really" in this line - but related...

Since `_download_pair_data()` is now only called from 2 places, both which have exchange as mandatory object - i think it's safe to remove this section of the test, as well as the `Optional[Exchange]` and the test for the exchange - since this can realistically never be called.

Unless I import the function manually and pass it something unfitting (a string instead of an exchange object - but that isn't checked anyway).
However,  that'll break most other functions too (we almost always rely on the type passed in to be correct ... otherwise we'd have every function first verify every object's type...) - and so do most other libraries.

tests/edge/test_edge.py
xmatthias: i guess so - yes - tests pass which means there's no typo in the object resolution (these are sometimes difficult to spot ...)

and not calling "refresh_data" from here is how it used to work before, too, by mocking load_data.

tests/data/test_history.py
hroff-1902: Actually I made it during last minor set of changes into data.history, removed Optional, but it led into mypy/flake complaints about the types (this `if not exchange` made mypy happy). I found then that it's easier and more effective to split this functionality first. So this PR was issued. after first part of the changes (#2672). Third part will follow...

PR #2683, opened 2019-12-17
Minor improvements in data.history

hroff-1902:
Follows #2678 and #2346 

* Removal of Optional in the type of exchange argument; removal of checking for exchange is None.
* `get_timeframe` renamed to `get_timerange` because it calculates and returns boundaries of the common timerange, not a timeframe as we use this term.
* type of the value returned by `load_tickerdata_file` clarified.

### DIFF
--- a/freqtrade/data/history.py
+++ b/freqtrade/data/history.py
@@ -68,7 +68,7 @@ def trim_dataframe(df: DataFrame, timerange: TimeRange, df_date_col: str = 'date
 
 
 def load_tickerdata_file(datadir: Path, pair: str, timeframe: str,
-                         timerange: Optional[TimeRange] = None) -> Optional[list]:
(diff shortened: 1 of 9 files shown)


freqtrade/data/history.py
```diff
@@ -447,18 +442,19 @@ def convert_trades_to_ohlcv(pairs: List[str], timeframes: List[str],
             store_tickerdata_file(datadir, pair, timeframe, data=ohlcv)
 
 
-def get_timeframe(data: Dict[str, DataFrame]) -> Tuple[arrow.Arrow, arrow.Arrow]:
+def get_timerange(data: Dict[str, DataFrame]) -> Tuple[arrow.Arrow, arrow.Arrow]:
```
xmatthias: I'll not complain since i think it was me who added this function ... but that was kinda oddly named, ey? :laughing:

PR #2686, opened 2019-12-18
Document buy / sell order pricings 

### LINKED ISSUE #1583: Add documentation for "orderbook"
We have the configuration options for orderbook.

They are however only described in "configuration options" - so it remains partially unclear how they exactly work - and each time I need them I need to look for the exact purpose in the code.

Orderbook would deserve a nice documentation section, explaining how it works and what each option does.

### LINKED ISSUE #2638: Limit BUY/SELL Oders
I'm having a little trouble understanding the part of the doc related to Limit BUY/SELL Oders. In particular, the doc say:
bid_strategy. check_depth_of_market.enabled
Do not buy if the difference of buy orders and sell orders is met in Order Book.

What does it mean that the difference of buy orders and sell orders is satisfied? How such difference is met? Also, do you refer to the numerical difference between the number of BUY and SELL orders present in the Order Book at a given time. Or, the difference in price between the first BUY and the first SELL at the top of the Order Book?
Maybe you mean something like that:
the ratio between the total number of orders of BUY and the total number of orders of SELL is computed, if such ratio exceeds the constant defined in "bid_strategy. check_depth_of_market.bids_to_ask_delta", then the order of BUY is created, otherwise the order of BUY is ignored.

A second issue of interpretation also in the Limit SELL options part of the doc: ask_strategy.order_book_min
Bot will scan from the top min to max Order Book Asks searching for a profitable rate.
How is it determined if a rate is profitable? what is it meant by rate? if "rate" is the selling price then this parameter always select the maximum price?

perhaps it is worth re-phrasing the text in the documentation in a manner consistent with the code.

### LINKED ISSUE #2682: ask_strategy configuration
I'm having a little trouble understanding the part of the doc related to ask_strategy. In particular, the doc says that when ask_strategy.use_order_book is set to false, then the selling of open trades using Order Book Asks is not enable.
 However, it is not clear whatelse is the bot going to use to sell open trades? The documentation does not clarify what the alternatives are. What does the bot take as a reference to sell, if this parameter is false?
In particular, what is the difference between: (ask_strategy.use_order_book set to FALSE), and (ask_strategy.use_order_book set to TRUE and (order_book_min=order_book_max=1)) ? 
What's wrong with the following interpretation of the above two alternatives:
Assume that the ticker_interval is 15m.
Il (ask_strategy.use_order_book set to FALSE) then the ask price is kept the same for all 15 minutes.
While in the case (ask_strategy.use_order_book set to TRUE and (order_book_min=order_book_max=1)), then the ask price changes at every minute according to the first position of the Order-book?

xmatthias:
## Summary
Document the buy / sell options including orderbook and depth_of_market

closes #2682
closes #2638
closes #1583

## Quick changelog

- Rename `get_ticker()` to `fetch_ticker()` - aligning to the ccxt method
- Document orderbook pricing options
- document depth of market 
- document default buy / sell pricing


hroff-1902:
pls review my changes to configuration.md

PR #2687, opened 2019-12-18
Try to get comment from forked repos

xmatthias:
## Summary
This will reenable Pull request comments when the PR comes from a foreign repository.

I really really don't like that we have to hardcode the token ... 
but github actions currently do not allow passing secrets to PR's from forked repositories.

on the other hand, this token allows only access to the coveralls repository site - and it's quickly rotated should something go wrong ...


xmatthias:
@hroff-1902 i guess this is ready to merge ... i've also updated the description above with some concerns i have

PR #2695, opened 2019-12-22
Improve pairlocking mechanism to allow usage from within strategy

### LINKED ISSUE #2582: Penalize: stop trading losing pairs
It would be very helpful if there was a way to stop or pause losing pairs. For example, if a pair caused 2 subsequent losing trades, it might be a good idea to stop trading said pair for a while.

xmatthias:
## Summary
Document usage of custom pairlocks, and slightly improve handling of it. The bot should never be allowed to lower the lock (it's not needed for internal use, and should not override external usage).

Closes #2582 

## Quick changelog

* Don't allow lowering the time once a lock is in place (use `unlock_pair()` instead)
* Implement an easy way to remove a pair from the lock (`unlock_pair()`).
* Documentation example for this method

### DIFF
--- a/docs/strategy-customization.md
+++ b/docs/strategy-customization.md
@@ -466,9 +466,11 @@ Locked pairs will show the message `Pair <pair> is currently locked.`.
 Sometimes it may be desired to lock a pair after certain events happen (e.g. multiple losing trades in a row).
 
 Freqtrade has an easy method to do this from within the strategy, by calling `self.lock_pair(pair, until)`.
-Until should be a time in the future, after which trading will be reenabled for that pair.
+`until` must be a datetime object in the future, after which trading will be reenabled for that pair.
 
 Locks can also be lifted manually, by calling `self.unlock_pair(pair)`.
+
+To verify if a pair is currently locked, use `self.is_pair_locked(pair)`.
 
 !!! Note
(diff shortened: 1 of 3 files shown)


docs/strategy-customization.md
```diff
@@ -455,6 +455,49 @@ Sample return value: ETH/BTC had 5 trades, with a total profit of 1.5% (ratio of
...
+Freqtrade has an easy method to do this from within the strategy, by calling `self.lock_pair(pair, until)`.
+Until should be a time in the future, after which trading will be reenabled for that pair.
+
+Locks can also be lifted manually, by calling `self.unlock_pair(pair)`.
+
+!!! Note
+    Locked pairs are not persisted, so a restart of the bot, or calling `/reload_conf` will reset locked pairs.
+
+!!! Warning
+    Locking pairs is not functional during backtesting.
```
hroff-1902: "functioning" ?

hroff-1902:
LGTM 👍 

But I personally would not call it "easy method", since it requires programming...

xmatthias:
> But I personally would not call it "easy method", since it requires programming...

well so is writing a strategy.
and honestly, it's not really programming if you can copy/paste the sample, and modify the filters to your needs.

However, needs around this will be too dynamic to make it as configuration.

PR #2697, opened 2019-12-22
persist Dry run db as default

### LINKED ISSUE #2600: db-url defaults during dryrun
In the [docker doc](https://www.freqtrade.io/en/latest/docker/#create-your-database-file) the doc recommend to create the database in live with  ` touch tradesv3.sqlite`, and in dryrun with`touch tradesv3.dryrun.sqlite`.

But then in the note below it says :

> db-url defaults to sqlite:///tradesv3.sqlite but it defaults to sqlite:// if dry_run=True is being used. To override this behaviour use a custom db-url value: i.e.: --db-url sqlite:///tradesv3.dryrun.sqlite

1/ not sure end user know what _sqlite://_ means : create in memory or create a default file, but we don't know his name ?
2/ When we say the end-user to create the files `touch tradesv3.dryrun.sqlite` for dry-run then the default should be _sqlite:///tradesv3.dryrun.sqlite_ and not _sqlite://_ .If we give recommandations for naming the databases in the doc, then we should take this in account for default. And It can be seen like a convention over configuration approach.

Have a nice day.

xmatthias:
## Summary
We should persist dry-run trades per default - since not having them in for analysis is almost never the desired state.

Closes #2600

## Quick changelog

- Change default database for dry-run mode to `sqlite:///tradesv3.dryrun.sqlite`.
- Update documentation


xmatthias:
> (But I personally liked this in-memory mode for Dry-Run)

it's still available via `--dry-run sqlite://` ... but explicitly ...

hroff-1902:
> > (But I personally liked this in-memory mode for Dry-Run)
> 
> it's still available via `--dry-run sqlite://` ... but explicitly ...

what about adding '--no-db' option in the future? (as a shortcut, equal to `--db-url sqlite://`

xmatthias:
well -i always found using positive and negative arguments confusing - what if i specify `--db-url sqlite:///asdf.sqlite --no-db` ? which one wins in this case?

hroff-1902:
they should be conflicting

PR #2705, opened 2019-12-23
Refactor resolvers to static resolvers

xmatthias:
## Summary
Creating an instance of the resolvers is not necessary, since we're just calling a few methods in sequence and then return exactly one result.

Also, if the resolver does not move out of scope it'll be kept in memory for the whole run-time of the bot ... which is unnecessary.

## Quick changelog

- Convert resolvers to static classes, using a `load_*()` method instead of `__init__()`.

### DIFF
--- a/docs/strategy_analysis_example.md
+++ b/docs/strategy_analysis_example.md
@@ -44,9 +44,9 @@ candles.head()
 ```python
 # Load strategy using values set above
 from freqtrade.resolvers import StrategyResolver
-strategy = StrategyResolver({'strategy': strategy_name,
-                            'user_data_dir': user_data_dir,
-                            'strategy_path': strategy_location}).strategy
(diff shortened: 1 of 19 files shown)


hroff-1902:
Just a question: should not all those classes with only static methods (IResolver, ClassnameResolver) further refactored and made each a module with functions at the module level? 

(not in this PR, probably, just would like to know are you going to refactor it further this way)

xmatthias:
no... i  see only disadvantages from removing the class.

Even though they currently have only static methods, they still have strong parent / child relationships between the interface class and their subsidiaries.

This structure makes it very clear which methods are shared and which are not - moving everything into module-level would make it a lot harder to understand for a "casual" reader (and probably our future selfs).

Thinking about it - some methods may be converted into class-methods in the future - which i hope can help to remove more duplicate code - but that's not a priority as this whole resolver infrastructure works quite well.

PR #2706, opened 2019-12-23
Convert datadir within config to Path

xmatthias:
## Summary
Currently, we convert datadir to Path every time we use it.

By converting when config is initialized, we can avoid this 100fold conversation.

It's a real / visible performance improvement - but it's unnecessary and clutters our code with unnecessary Path() calls and imports.

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -403,7 +403,7 @@ class Configuration:
             config['pairs'] = config.get('exchange', {}).get('pair_whitelist')
         else:
             # Fall back to /dl_path/pairs.json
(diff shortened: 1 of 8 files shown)


hroff-1902:
LGTM 👍 

I have not reviewed and merged it quickly since I could not find quickly the place in the code where the datadir param in the config is assigned its value of the Path datatype, it took some additional efforts to find it out in the code... 😄

xmatthias:
> where the datadir param in the config is assigned its value of the Path datatype, it took some additional efforts to find it out in the code

well - `configuration.py` ... or rather `create_datadir` ... which i always designed to return a path but forgot to follow up :D

PR #2710, opened 2019-12-24
/balance should not convert to BTC

xmatthias:
## Summary
Align /balance output to show everything in stake currency the conversation to BTC does not make sense

## Quick changelog

- fix typo
- convert total amount to stake-currency instead of BTC


hroff-1902:
That was in Slack from a user, right?

PR #2711, opened 2019-12-24
Align Edge documentation to configuration page

xmatthias:
## Summary
We should use the same style when documenting a set of configuration options.
Also renamed the Header of the tables to Parameter instead of Command - since these are parameters / configurations, not commands.

##
* convert edge documentation to table
* newline before "Keep it seecret" to improve appearance in the doc pages

### DIFF
--- a/docs/edge.md
+++ b/docs/edge.md
@@ -1,4 +1,4 @@
-#  Edge positioning
+# Edge positioning
 
 This page explains how to use Edge Positioning module in your bot in order to enter into a trade only if the trade has a reasonable win rate and risk reward ratio, and consequently adjust your position size and stoploss.
 
@@ -9,6 +9,7 @@ This page explains how to use Edge Positioning module in your bot in order to en
     Edge does not consider anything else than buy/sell/stoploss signals. So trailing stoploss, ROI, and everything else are ignored in its calculation.
 
 ## Introduction
+
 Trading is all about probability. No one can claim that he has a strategy working all the time. You have to assume that sometimes you lose.
 
 But it doesn't mean there is no rule, it only means rules should work "most of the time". Let's play a game: we toss a coin, heads: I give you 10$, tails: you give me 10$. Is it an interesting game? No, it's quite boring, isn't it?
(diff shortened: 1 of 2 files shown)


docs/edge.md
```diff
@@ -115,100 +140,30 @@ Available capital doesn’t change before a position is sold. Let’s assume tha
...
-
-Edge will remove sudden pumps in a given market while going through historical data. However, given that pumps happen very often in crypto markets, we recommend you keep this off.
+Edge module has following configuration options:
 
-(defaults to false)
+|  Parameter | Description |
+|------------|-------------|
+| `enabled` | If true, then Edge will run periodically. <br>*Defaults to `false`.* <br> ***Datatype:*** *Boolean*
+| `process_throttle_secs` | How often should Edge run in seconds. <br>*Defaults to `3600` (once per hour).* <br> ***Datatype:*** *Integer*
+| `calculate_since_number_of_days` | Number of days of data against which Edge calculates Win Rate, Risk Reward and Expectancy. <br> **Note** that it downloads historical data so increasing this number would lead to slowing down the bot. <br>*Defaults to `7` (once per hour).* <br> ***Datatype:*** *Integer*
```
hroff-1902: remove "once per hour"

docs/edge.md
```diff
@@ -115,100 +140,30 @@ Available capital doesn’t change before a position is sold. Let’s assume tha
...
+| `capital_available_percentage` | This is the percentage of the total capital on exchange in stake currency. <br>As an example if you have 10 ETH available in your wallet on the exchange and this value is 0.5 (which is 50%), then the bot will use a maximum amount of 5 ETH for trading and considers it as available capital. <br>*Defaults to `0.5` (once per hour).* <br> ***Datatype:*** *Float*
```
hroff-1902: remove "once per hour"

docs/edge.md
```diff
@@ -115,100 +140,30 @@ Available capital doesn’t change before a position is sold. Let’s assume tha
...
+| `allowed_risk` | Ratio of allowed risk per trade. <br>*Defaults to `0.01` (1%)).* <br> ***Datatype:*** *Float*
+| `stoploss_range_min` | Minimum stoploss. <br>*Defaults to `-0.01`.* <br> ***Datatype:*** *Float*
+| `stoploss_range_max` | Maximum stoploss. <br>*Defaults to `-0.10`.* <br> ***Datatype:*** *Float*
+| `stoploss_range_step` | As an example if this is set to -0.01 then Edge will test the strategy for `[-0.01, -0,02, -0,03 ..., -0.09, -0.10]` ranges. <br> **Note** than having a smaller step means having a bigger range which could lead to slow calculation. <br> If you set this parameter to -0.001, you then slow down the Edge calculation by a factor of 10. <br>*Defaults to `-0.001`.* <br> ***Datatype:*** *Float*
```
hroff-1902: The example should probably be changed (other value instead of -0.001 used), because "If you set this parameter to -0.001" sounds like the default value slows down the Edge calculations.

xmatthias: I'm not sure about this (how to change it - if the default value is ok ...).

The old version had -0.01 as default - but it wasn't the default when looking at the code, that's why i changed it to -0.001 (to align documentation with code).

now the sample and calculation is not wrong per se - going from -0.01 to -0.001 slows down by a factor of then - the same as when going from -0.001 to -0.0001...

hroff-1902: Yes, I understand why you have changed it in the description.

Okay, probably the default value in the code should be aligned to one used everywhere in the sample configs (to be -0.01). Not in this PR.

Without this the phrase used here looks a bit strange -- like the default value makes the calculations slower and thus should not be used...

PR #2712, opened 2019-12-24
add list-strategies subcommand

xmatthias:
## Summary
Add list-strategies subcommand allowing to list all strategies in a folder.
Default output is a table, but machine-readable `--one-column` output is available as well.

## Quick changelog

- Refactor resolvers further to use class-mehods to reduce code duplication
- Combine `_load_*()` methods into a generalized one (apart from load_strategy due to loading from base64)
- Allow `IResolver._get_valid_object()` to work without a name, where it returns all valid objects within that file
- move module instanciation to `_load_object()` to support searching modules without initializing them.

### DIFF
--- a/docs/utils.md
+++ b/docs/utils.md
@@ -108,6 +108,44 @@ With custom user directory
 freqtrade new-hyperopt --userdir ~/.freqtrade/ --hyperopt AwesomeHyperopt
 ```
 
+## List Strategies
+
+Use the `list-strategies` subcommand to see all strategies in one particular folder.
+
+```
+freqtrade list-strategies --help
+usage: freqtrade list-strategies [-h] [-v] [--logfile FILE] [-V] [-c PATH] [-d PATH] [--userdir PATH] [--strategy-path PATH] [-1]
+
+optional arguments:
+  -h, --help            show this help message and exit
+  --strategy-path PATH  Specify additional strategy lookup path.
+  -1, --one-column      Print output in one column.
+
+Common arguments:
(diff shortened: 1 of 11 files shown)


docs/utils.md
```diff
@@ -108,6 +108,47 @@ With custom user directory
...
+                        Path to userdata directory.
+```
+
+!!! Warning
+    Using this command will try to load all python files from a folder. This can be a security risk if untrusted files reside in this folder, since all module-level code is executed.
+
+Example: search default strategy folder within userdir
+
+``` bash
+freqtrade list-strategies --user-data ~/.freqtrade/
```
hroff-1902: "--userdir"

docs/utils.md
```diff
@@ -108,6 +108,47 @@ With custom user directory
 freqtrade new-hyperopt --userdir ~/.freqtrade/ --hyperopt AwesomeHyperopt
 ```
 
+## List Strategies
+
+Use the `list-strategies` subcommand to see all strategies in one particular folder.
```
hroff-1902: no folders 😉 
(pls also check other appearances)

docs/utils.md
xmatthias: `--user-data-dir` works as well ... which shortcuts to `--user-data` ... but i agree, we should be consistent :D

xmatthias:
> I propose to print only filename (like sample_strategy.py) in the location column instead of full path. It's redundant, we are listing a single dir...

good point - done.

PR #2718, opened 2019-12-28
Minor: code cleanup in freqtradebot

hroff-1902:
* Cleanup logic in get_trade_stake_amount()
* Introduced new _check_available_stake_amount(), _calculate_unlimited_stake_amount() and get_free_open_trades() methods that encapsulate appropriate parts of business logic.
* This does not (should not) change behavior (business logic), except the change for max open trades in process(). Currently, if max open trades is reached AND some trades are sold before processing buys, the buys will only be processed at the next throttling iteration. New behavior: we have some free open trade slots right after trades sold so we can process buys right at this throttling iteration. This is due to the call to get_free_open_trades() method made again before we proceed to buys instead of comparison of the length of the trades array, which is collected before the trades sold.
* Cosmetics changes

Marked as minor since it does not change business logic.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -335,6 +335,10 @@ class FreqtradeBot:
                     continue
 
                 stake_amount = self.get_trade_stake_amount(pair)
+                if not stake_amount:
+                    logger.warning("Stake amount is 0, ignoring trade")
+                    continue
+
                 logger.info(f"Buy signal found: about create a new trade with stake_amount: "
                             f"{stake_amount} ...")
 
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -204,7 +204,7 @@ class FreqtradeBot:
 
         return used_rate
 
-    def _get_trade_stake_amount(self, pair) -> Optional[float]:
+    def get_trade_stake_amount(self, pair) -> Optional[float]:
         """
         Check if stake amount can be fulfilled with the available balance
         for the stake currency
@@ -309,7 +309,7 @@ class FreqtradeBot:
                 self.dataprovider.ohlcv(_pair, self.strategy.ticker_interval))
 
             if buy and not sell and len(Trade.get_open_trades()) < self.config['max_open_trades']:
-                stake_amount = self._get_trade_stake_amount(_pair)
+                stake_amount = self.get_trade_stake_amount(_pair)
                 if not stake_amount:
                     continue
 
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -136,7 +136,7 @@ class FreqtradeBot:
         self.process_maybe_execute_sells(trades)
 
         # Then looking for buy opportunities
-        if len(trades) < self.config['max_open_trades']:
+        if self.get_free_open_trades():
             self.process_maybe_execute_buys()
 
         # Check and handle any timed out open orders
@@ -172,6 +172,14 @@ class FreqtradeBot:
         Create pair-whitelist tuple with (pair, ticker_interval)
         """
         return [(pair, self.config['ticker_interval']) for pair in pairs]
+
+    def get_free_open_trades(self):
+        """
+        Return the number of free open trades slots or 0 if
+        max number of open trades reached
+        """
+        open_trades = len(Trade.get_open_trades())
+        return max(0, self.config['max_open_trades'] - open_trades)
 
     def get_target_bid(self, pair: str, tick: Dict = None) -> float:
         """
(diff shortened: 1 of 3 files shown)


freqtrade/freqtradebot.py
```diff
@@ -299,18 +320,23 @@ def create_trades(self) -> bool:
...
+            if buy and not sell:
+                if not self.get_free_open_trades():
+                    logger.warning("Can't open a new trade: max number of trades is reached")
+                    continue
 
-            if buy and not sell and len(Trade.get_open_trades()) < self.config['max_open_trades']:
-                stake_amount = self._get_trade_stake_amount(_pair)
+                stake_amount = self.get_trade_stake_amount(pair)
                 if not stake_amount:
+                    logger.warning("Stake amount is 0, ignoring trade")
```
xmatthias: Not sure about the loglevel of warning ...

we usually use "warning" for things which may need attention  - this sounds more like "info" to me - as it's a valid and probably common possibility to have 0 remaining stake amount available (all stake is tied up in trades).

I would also reword it to `"Stake amount is 0, ignoring possible trade for {pair}."`

freqtrade/freqtradebot.py
```diff
@@ -204,33 +212,46 @@ def get_target_bid(self, pair: str, tick: Dict = None) -> float:
...
-                logger.warning("Can't open a new trade: max number of trades is reached")
-                return None
-            return available_amount / (self.config['max_open_trades'] - open_trades)
+    def _check_available_stake_amount(self, stake_amount: Optional[float]) -> Optional[float]:
+        """
+        Check if stake amount can be fulfilled with the available balance
+        for the stake currency
+        :return: float: Stake amount
+        """
+        available_amount = self.wallets.get_free(self.config['stake_currency'])
```
xmatthias: I would maybe move this to the outermost function (`get_trade_stake_amount()`) and pass the result around.

That would increase the ease at which we can test these seperately since they'd be completely selfcontained - and we'd not need to mock get_free to return the correct thing.

freqtrade/freqtradebot.py
```diff
@@ -173,6 +173,14 @@ def _create_pair_whitelist(self, pairs: List[str]) -> List[Tuple[str, str]]:
         """
         return [(pair, self.config['ticker_interval']) for pair in pairs]
 
+    def get_free_open_trades(self):
+        """
+        Return the number of free open trades slots or 0 if
+        max number of open trades reached
+        """
+        open_trades = len(Trade.get_open_trades())
+        return max(0, self.config['max_open_trades'] - open_trades)
```
xmatthias: Nice call - i assume a negative number can happen if max_open_trades is modified to a lower number (from 5 to 3) after 5 trades are already open?

freqtrade/freqtradebot.py
```diff
@@ -299,18 +320,23 @@ def create_trades(self) -> bool:
...
 
             (buy, sell) = self.strategy.get_signal(
-                _pair, self.strategy.ticker_interval,
-                self.dataprovider.ohlcv(_pair, self.strategy.ticker_interval))
+                pair, self.strategy.ticker_interval,
+                self.dataprovider.ohlcv(pair, self.strategy.ticker_interval))
+
+            if buy and not sell:
+                if not self.get_free_open_trades():
+                    logger.warning("Can't open a new trade: max number of trades is reached")
```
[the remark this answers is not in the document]
hroff-1902: Forgotten to make it (and the next one, see your comment below) to be at the info level as I intended...

This part will be further refined, let them both be on the debug level for now...

freqtrade/freqtradebot.py
```diff
@@ -204,33 +212,46 @@ def get_target_bid(self, pair: str, tick: Dict = None) -> float:
...
+        return self._check_available_stake_amount(stake_amount)
 
+    def _calculate_unlimited_stake_amount(self) -> Optional[float]:
+        """
+        Calculate stake amount for "unlimited" stake amount
+        :return: None if max number of trades reached
+        """
+        free_open_trades = self.get_free_open_trades()
+        if not free_open_trades:
+            return None
```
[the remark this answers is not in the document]
hroff-1902: I also thought of it but kept interface similar to `_get_min_pair_stake_amount()`, which returns None.

Not in this PR, this one contains some more 'minor' changes, split from the rest to be easily tracked.

freqtrade/freqtradebot.py
hroff-1902: This one should normally be printed in the meantime since no methods used above to calculate stake_amount can return 0. It was added primarily for mypy satisfaction. However, this can also be useful (or at least be working) taking account upcoming improvements that can be made here.

Anyway, both messages made debug for now to be non-noisy.
Text of this message was adjusted. 👍

freqtrade/freqtradebot.py
hroff-1902: Yes, for example in this case.

freqtrade/freqtradebot.py
hroff-1902: Hmm, I would not like to do that, this function is now more like a blackbox, fetching internally the info it needs to check the passed stake_amount with. If it will check against relative stake amount ratio, that value should also be fetched inside this function, not passed as the parameter...

freqtrade/freqtradebot.py
hroff-1902: ^^ This one should NOT normally be printed...

PR #2725, opened 2019-12-30
[Minor] Edge-cli should use exchangeresolver

xmatthias:
## Summary
Edge-cli should use Exchangeresolver just as everything else.

Also added a small debug-assert trying to find the root-cause for the random test-failures (which are not reproducible with the same random seed).

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -448,6 +448,9 @@ def test_create_datadir(caplog, mocker):
     # Ensure that caplog is empty before starting ...
     # Should prevent random failures.
     caplog.clear()
+    # Added assert here to analyze random test-failures ...
+    assert len(caplog.record_tuples) == 0
+
     cud = mocker.patch("freqtrade.utils.create_userdata_dir", MagicMock())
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -448,6 +448,9 @@ def test_create_datadir(caplog, mocker):
     # Ensure that caplog is empty before starting ...
     # Should prevent random failures.
     caplog.clear()
+    # Added assert here to analyze random test-failures ...
+    assert len(caplog.record_tuples) == 0
```
hroff-1902: Did not get this change...

xmatthias:
> Did not understand the change made in the test_utils...

it's a pointless change, really - from a testing perspective, anyway.

```python
assert len(caplog.record_tuples) == 0
```
(the lower version - a few lines below this change) fails at random in CI - but never on my PC - not with the same random seed, and not with a different one.

Now i moved another instance of this right after the `caplog.clear()` - where caplog should never have an entry - just to verify if this is a bug in pytest (not clearing caplog correctly or something like that) - or if something else is going on and the log-entry is "somehow" added by freqtrade code.

PR #2729, opened 2020-01-01
Cosmetics in freqtradebot

hroff-1902:
Small cleanup in Freqtradebot for readability:

* Separate BUY and SELL parts of the business logic with the block comments
* Move `enter_positions()` up in the BUY part to align the position of the `exit_positions()` in its SELL part of the code
* Fix comment

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -407,7 +407,7 @@ class FreqtradeBot:
         if price:
             buy_limit_requested = price
         else:
-            # Calculate amount
+            # Calculate price
             buy_limit_requested = self.get_target_bid(pair)
 
         min_stake_amount = self._get_min_pair_stake_amount(pair, buy_limit_requested)
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -183,7 +183,7 @@ class FreqtradeBot:
         return max(0, self.config['max_open_trades'] - open_trades)
 
 #
-# BUY / enter positions / open trades part
+# BUY / enter positions / open trades logic and methods
 #
 
     def enter_positions(self) -> int:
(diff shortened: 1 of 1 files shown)


xmatthias:
I thought about something similar too - back when i used to do c# you could define blocks you could later collapse in the code editor - unfortunately i don't think that'll work here.

Apart from that, i think the sell block contains way too much.
Sell contains `update_trade_state()` and `get_real_amount()` which is used by both buy and sell. - so maybe we should move them above the first block ... so it's general (?) 

It also contains the whole stoploss logic. That's not incorrect per se since it is related to selling, however i think that would earn it's own block...

The sell block also contains the "handle_timedout" section ... which again, should probably be it's own block, since it'll apply to both buying and selling

hroff-1902:
> It also contains the whole stoploss logic. That's not incorrect per se since it is related to selling, however i think that would earn it's own block...
> 
> The sell block also contains the "handle_timedout" section ... which again, should probably be it's own block, since it'll apply to both buying and selling

You see how simple splitting code into that two parts by highly visible block comments helps to arrange and cleanup code logic and find better design solutions? 😄

hroff-1902:
> I thought about something similar too - back when i used to do c# you could define blocks you could later collapse in the code editor - unfortunately i don't think that'll work here.
> 
> Apart from that, i think the sell block contains way too much.
> Sell contains `update_trade_state()` and `get_real_amount()` which is used by both buy and sell. - so maybe we should move them above the first block ... so it's general (?)

I also notices this for `update_trade_state()` method; have not seen that before for `get_real_amount()`, but I did not try to split this common part...

Anyway, this note was addressed, both methods moved into a newly arranged common section. 👍 

I think that's all that can be split in terms of this PR, further cleanup can be done later.

PR #2733, opened 2020-01-02
Cleanup buy/sell notification in freqtradebot

hroff-1902:
* Introduce `_notify_buy()`, which is aligned to existing `_notify_sell`. This simplifies the code in `execute_buy()` a bit, since no configuration params that are needed only for notification should be fetched there.
* Minor cleanup of creation the sell msg

To be discussed/possibly fixed (changed) here, pls review this separately:
* should not the call to `_notify_buy()` be moved in `execute_buy()` to be at the end, right before `return`? Or after `Trade.session.add()` and `Trade.session.flush()` to contain data possibly updated by `self.update_trade_state()`? Now it's left in its place to have same info as in the current implementation...

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -498,15 +498,9 @@ class FreqtradeBot:
             'limit': trade.open_rate,
             'order_type': order_type,
             'stake_amount': trade.stake_amount,
+            'stake_currency': self.config['stake_currency'],
+            'fiat_currency': self.config.get('fiat_display_currency', None),
         }
-
-        if 'stake_currency' in self.config and 'fiat_display_currency' in self.config:
-            stake_currency = self.config['stake_currency']
-            fiat_currency = self.config['fiat_display_currency']
-            msg.update({
-                'stake_currency': stake_currency,
-                'fiat_currency': fiat_currency,
-            })
 
         # Send the message
         self.rpc.send_msg(msg)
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -952,15 +959,13 @@ def _notify_sell(self, trade: Trade, order_type: str):
...
-            'close_date': trade.close_date or datetime.utcnow()
+            'close_date': trade.close_date or datetime.utcnow(),
+            'stake_currency': self.config['stake_currency'],
         }
 
-        # For regular case, when the configuration exists
-        if 'stake_currency' in self.config and 'fiat_display_currency' in self.config:
-            stake_currency = self.config['stake_currency']
+        if 'fiat_display_currency' in self.config:
             fiat_currency = self.config['fiat_display_currency']
```
xmatthias: I suspect we could remove this line too - it has no visible benefit (not even flake compliance due to line length)...

hroff-1902: Right, but this requires some changes into code handling those dict items in rpc/telegram and rpc/api-server, so I left it for the next PR to keep this one consisting of minor changes that do not change behavior.

xmatthias: I think you misunderstood...
moving `self.config['fiat_display_currency']` into the `msg.update()` function won't require changes.

I've done the change i meant ... ;)

PR #2734, opened 2020-01-02
Relative stake maximum tradable amount

### LINKED ISSUE #1985: If Count is X, X-1 forcebuy wont work over telegram

### LINKED ISSUE #2458: Relative stake_amount
This issue will replace #2359 - to summarize the points discussed there relating stake-amount / upper stake_amount_% limits.

### Idea

usage of  `stake_amount="unlimited"` is currently tied to `max_open_trades`.
one stake is always `available_balance / (max_open_trades - open_trade_count)`.

A new config option should be introduced, specifying `stake_amount_max`.
This `stake_amount_max` can either be absolute or relative

* absolute: `stake_amount_max_abs=1`: Even if i have 200 ETH in my account, only trade up to 1 ETH, splitting the 1ETH into `max_open_trades` elements (assuming 2 max_open_trades, it would be 0.5ETH per trade). 
Gains will be kept "untradable" - losses will be eaten up by the remaining 199ETH.

* relative: `stake_amount_max_ratio=0.5`: assuming i have 1 ETH in my account, only trade 50% of the total account. assuming 2 max_open_trades: first 2 trades have a stake of  0.25ETH each.
Won trades will increase the stake used in subsequent trades, lost trades will decrease this.

**note:** These options will require `stake_amount` to be absent or "unlimited".

###  discussion points:

* [ ] is `stake_amount_max_abs` even needed? it's just another way to specify `stake_amount=0.5`, `max_open_trades=2` - which will do exactly the same.
* [x] how to handle `stake_amount_max_ratio` in conjunktion with dry-mode - since it does currently not keep track of balance.

** future addons**
support compound-gains for backtesting ...

### LINKED ISSUE #2489: Insufficient funds to create buy order on 'stake_amount="unlimited" '

Yes

  * Operating system: Ubuntu 18.04.3
  * Python Version: 3.6.8
  * Branch: Develop
  
I have set up my bot to run with 'stake_amount="unlimited" ', but it is running into an error while trying to open a trade. The error message reads: 

`"WARNING - Unable to create trade: Insufficient funds to create market buy order on market X. Tried to buy amount X at rate None.Message: binance Account has insufficient balance for requested action."`

The amount it tried to create a buy order for matches exactly to the amount I have available on Binance - I checked.

The first time the bot encountered a buy signal with 'stake_amount="unlimited" ' went well and the trade was executed but the following ones ran into the error described above.

I have tried setting 'amount_reserve_percent=0' but to no avail. Here's the bid/ask strategy part of my config file:

```
"bid_strategy": {
        "ask_last_balance": 0.0,
        "use_order_book": false,
        "order_book_top": 1,
        "check_depth_of_market": {
            "enabled": false,
            "bids_to_ask_delta": 1
        }
    },
    "ask_strategy":{
        "use_order_book": false,
        "order_book_min": 1,
        "order_book_max": 9
```

Worth mentioning that I do not have any BNB available on my Binance account and BNB is blacklisted in my config.

Is there a way to resolve this issue? Could I change anything in the bid/ask part of my config file to work around this?
  
Thanks for your help
Kevin

xmatthias:
## Summary
This will allow specifying a relative maximum trading ratio (`tradable_balance_ratio`) as discussed in #2458.

`tradable_balance_ratio` defaults to 0.99 (99%) - leaving 1% of safety margin.
this margin is applied to the whole account (available stake balance + open trade stakes) to avoid "vanishing" trade-amounts, since the specified ratio is deduced from the remaining balance.

Closes #2458
Closes #2489 (relevant for "last slot" market buys since price can (should) go up)
closes #1985 (almost certainly similar to #2489 - buy was probably 'market' there too - otherwise it would have worked)

## Quick changelog

- add `tradable_balance_ratio`
- adapt tests for this behaviour
- Add configuration check making sure only `stake_amount` or `max_open_trades` is set to unlimited.
- Introduce `last_stake_amount_ratio`
- introduce `amend_last_stake_amount`
- deprecate `edge.capital_available_percentage` - which configures the same as `tradable_balance_ratio`  but in the context of edge.


docs/configuration.md
```diff
@@ -43,6 +43,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `max_open_trades` | **Required.** Number of trades open your bot will have. If -1 then it is ignored (i.e. potentially unlimited open trades).<br> ***Datatype:*** *Positive integer or -1.*
 | `stake_currency` | **Required.** Crypto-currency used for trading. [Strategy Override](#parameters-in-the-strategy). <br> ***Datatype:*** *String*
 | `stake_amount` | **Required.** Amount of crypto-currency your bot will use for each trade. Set it to `"unlimited"` to allow the bot to use all available balance. [More information below](#understand-stake_amount). [Strategy Override](#parameters-in-the-strategy). <br> ***Datatype:*** *Positive float or `"unlimited"`.*
+| `tradable_balance_ratio` | Ratio of the total account balance the bot is allowed to trade. Only applies when `stake_amount="unlimited"`. [More information below](#understand-stake_amount). <br>*Defaults to `0.99` 99%).*<br> ***Datatype:*** *Positive float between `0.0` and `1.0`.*
```
hroff-1902: "between `0.1` and `1.0`" -- the min. boundary set in the schema is 0.1

hroff-1902:
An open point here is the following:
should not we also use this ration when we compare the amount of the trade with available balance?
So, if `tradable_balance_ratio=0.99`, balance=1000 and stake_amount=100, then only 9 trade should be allowes if max_open_trades is "unlimited", not 10 trades. Same even for a fixed stake amount and fixed max_open_trades -- when we check if we have available balance for the next trade, we should probably compare stake_amount to `free_balance*tradable_balance_ratio` (verify logic here), so this should work in all cases, not only for unlimited stake amount...

xmatthias:
the only area where unlimited trades (`max_trade_amount=-1`) is in use is with edge.
there, we have the `edge.capital_available_percentage` setting, which does exactly the same thing (so yes, the setting is duplicate we should consolidate at a later point...).

We do check for available balance in `_check_available_stake_amount()` (but that's unrelated to the `max_open_trades` setting). 
Other than that, `max_open_trade` is only used with unlimited stake, where it would fail if both are unlimited, since `x / float('inf') == 0.0` (that's probaby a configuration check we're missing ...)

while we could implement it in `_check_available_stake_amount()` as well (unrelated to `max_open_trades` though) - i'm not certain about the implications, and i think it may have unintended side-effects to already configured bots.

Since the value of `tradable_balance_ratio` defaults to 99% (intentionally, to capture the errors this will close) - we would prevent you from opening a 2nd trade in the following example (`tradable_balance_ratio` is not configured explicitly):
balance=1001, stake_amount=500, max_open_trades=2

As a user, i did my math and know it would leave 1 unit in the account. However, 1001 * 0.99 = 990.99 ... so now the 2nd trade would not work unless `stake_amount` is reduced to 495.

After encountering this, the user can set `tradable_balance_ratio=1` after researching the problem - but it's not obvious and won't be noticed until it happens (that's the reason to set it to 99% - since you can run a bot for months without noticing this last trade could not be opened due to certain circumstances).

hroff-1902:
As discussed on Slack, tradable_balance_ratio should limit the tradable amount (calculated of current balance, i.e. using free_balance+open_trades_amount) in any cases, for limited or unlimited stake_amount, for limited or unlimited max_open_trades, in any allowed combinations of these settings.

So anything which is above tradable_balance_ratio is reserved and "untradable".

freqtrade/edge/__init__.py
```diff
@@ -57,7 +57,9 @@ def __init__(self, config: Dict[str, Any], exchange, strategy) -> None:
         if self.config['stake_amount'] != constants.UNLIMITED_STAKE_AMOUNT:
             raise OperationalException('Edge works only with unlimited stake amount')
 
-        self._capital_percentage: float = self.edge_config.get('capital_available_percentage')
+        # Deprecated capital_available_percentage. Will use tradable_balance_ratio in the future.
+        self._capital_percentage: float = self.edge_config.get(
+            'capital_available_percentage', self.config['tradable_balance_ratio'])
```
xmatthias: Once `capital_available_percentage` is removed completely, we should refactor edge to rely on the calculation done in `freqtradebot.py`.

hroff-1902:
Well, since we are touching actively this part, so may be in this PR. This ideally works together with new setting and with the removal of `capital_available_percentage` in Edge:

It was my idea after I refactored recently `Freqtradebot.get_trade_stake_amount()` to align behavior of Edge (stake amount defined in `edge.stake_amount()`) and two other methods of defining stake_amount:

Edge calls
```
position_size = min(max_position_size, free_capital)
```
So it amends the calculated stake amount to be the free capital value if it's calculated to be bigger. Two other methods do not contain this amending rule.

There can be a general setting, like "amend_last_stake_amount", True/False.

If this set to True, the stake amount will be amended to the free capital (i.e. to 990 instead of 1000) and this value will be used for the last allowed trade. If it's False, the trade will be rejected, if it cannot be made with the whole calculated or defined value.

Maybe not in this PR...

xmatthias:
> If this set to True, the stake amount will be amended to the free capital (i.e. to 990 instead of 1000) and this value will be used for the last allowed trade. 

Just to confirm i understand you correctly:
so assuming the above example of balance=1001, stake_amount=500, max_open_trades=2, this new option would buy the first with `500` (990 / 2) - and would then buy the 2nd trade with the "rest" - using another 490 for the 2nd trade?

Current behaviour would be that the 2nd trade is rejected as "too small"  - and that would remain as "optional" option?

hroff-1902:
> > If this set to True, the stake amount will be amended to the free capital (i.e. to 990 instead of 1000) and this value will be used for the last allowed trade.
> 
> Just to confirm i understand you correctly:
> so assuming the above example of balance=1001, stake_amount=500, max_open_trades=2, this new option would buy the first with `500` (990 / 2) - and would then buy the 2nd trade with the "rest" - using another 490 for the 2nd trade?
> 
> Current behaviour would be that the 2nd trade is rejected as "too small" - and that would remain as "optional" option?

I guess it should better be 495+495 for unlimited stake amount (if max_open_trades=2). For fixed stake amount (=500, max_open_trade > 1), first trade is 500 and second is either amended to 490 (if amending is switched on) or rejected (if amending is off). The calculated amount value is passed to `execute_buy` where it can also be rejected because it's less than minimal stake amount for the pair. (this check can also be moved ouside of `execute_buy`).

xmatthias:
>I guess it should better be 495+495 for unlimited stake amount  

well that's clear - i don't think this would even apply to unlimited stake ...
will have a look about implementing that ...

xmatthias:
The "amend" logic is now also ready for review :+1:

hroff-1902:
> Note that i've also removed the "None" returning ... it avoids adding a None-check in this case ... and i wanted to get rid of this anyway ...

👍

docs/configuration.md
```diff
@@ -146,6 +147,22 @@ For example, if you have 10 ETH available in your wallet on the exchange and `tr
 !!! Warning
     `tradable_balance_ratio` applies to the current balance (free balance + tied up in trades). Therefore, assuming a starting balance of 1000, a configuration of `tradable_balance_ratio=0.99` will not guarantee that 10 units will always remain available on the exchange. The free amount may reduce to 5 units if the total balance is reduce to 500 (either by a losing streak, or by withdrawing balance).
 
+#### Amend last stake amount
+
+Assuming we have a `balance=1000` (USDT), `stake_amount=400`, and `max_open_trades=3`.
```
hroff-1902: "we have the tradable balance of 1000 USDT, ..."
-- that's since we should count here the `tradable_balance_ratio` for balance. But mentioning this option would make things more complicated than it should be, so a sentence like this can be used.

hroff-1902:
* the amendment logic should also be applied for amounts generated by edge. I mean, that 
```
position_size = min(max_position_size, free_capital)
```
line should be removed, calculated `max_position_size` returned and amendment logic be applied for the amount returned.
* I think some boundary limit should be introduced for amended amount. Allow it be "close" enough to the calculated (or set statically in case of fixed one) stake amount. For example, if stake_amount is 400, allow amended trade only if amended amount > 200 (How to define better this limit? Is it 2.0 or 0.5?) That's because positions with small amounts maybe useless for particular strategies or even cause losses (taking open_trade slot with small amount, for example...) Not in this PR, however, I think, this can be introduced later.

xmatthias:
> line should be removed, calculated max_position_size returned and amendment logic be applied for the amount returned.

i understand what you mean - but for edge, this is not "configurable" ... so it always simply takes the last available amount (which i guess makes sense - since edge defines each stake-amount variable - depending on the calculated risk / stoploss).
Now while we could move that out of that module, i would rather not do that in this PR - since i think it's already complicated enough, and functionally it won't change (the endresult is the same now, as it should be in the future).

>  think some boundary limit should be introduced for amended amount.

I disagree here. 
I was thinking about this too - since this also risks introducing positions very close to the minimum tradable amount.

Now if a strategy depends on a specific position size, then this option is simply "not for you" - since even 80% of the initial position size would throw the strategy off.

Now apart from this - i think this should be added in this PR, not later.
Otherwise configurations may need to be changed twice ... 
I would configure it using 0.5 (50%) - so the minimum is 50% of the original position size.
2.0 would be ... odd - since it's not twice the position size, but half ... 

how about `last_stake_amount_min_ratio` as a setting value?

docs/configuration.md
```diff
@@ -149,7 +149,7 @@ For example, if you have 10 ETH available in your wallet on the exchange and `tr
 
 #### Amend last stake amount
 
-Assuming we have a `balance=1000` (USDT), `stake_amount=400`, and `max_open_trades=3`.
+Assuming we have the tradable balance of 1000 USDT (`balance=1000`), `stake_amount=400`, and `max_open_trades=3` (This example assumes `tradable_balance_ratio=1`).
```
hroff-1902: What is `balance=1000` here? We do not have such a setting... And adding tradable_balance_ratio only makes things overcomplicated for the reader...

For me,
"Assuming we have the tradable balance of 1000 USDT, `stake_amount=400`, and `max_open_trades=3`."
would be enough.

It's tradable balance, whatever it's defined.

hroff-1902:
> Now apart from this - i think this should be added in this PR, not later.
> Otherwise configurations may need to be changed twice ...

Agree. That's why I mentioned this above, to be included in this PR.

> I would configure it using 0.5 (50%) - so the minimum is 50% of the original position size.
> 2.0 would be ... odd - since it's not twice the position size, but half ...

Well, the logic behind 2,0 is that the amending can 'shrink' the amount up to the ratio of 2.0. So if it's defined this way, this can be treated as max shrink ratio.

But this is odd, 0.5 is better.

> 
> how about `last_stake_amount_min_ratio` as a setting value?

👍

hroff-1902:
And yes, let's leave the Edge part out of scope of this PR
(but for me it would be a small change, aligning behavior of amounts generated by Edge to the common amount amendment logic)

docs/configuration.md
hroff-1902: Let's remove "(This example assumes `tradable_balance_ratio=1`).", it's not true -- because this example does not assume this, it's actual for any tradable_balance_ratio if tradable balance is 1000, for any reason with any "full" balance and any tradable_balance_ratio, if they combination gives tradable balance equal to 1000.

xmatthias: well it's not wrong either - if i set this to 0.5, then the example would be wrong.

I've removed it now however.

hroff-1902: Hmm, wallet is 2000 USDT, tradable_balance_ratio=0.5, then tradable balance is 1000 USDT and the example is actual...

xmatthias: well yes - we can change it in many ways to either have it correct or wrong - but i think as it is now it's correct (?) ...

hroff-1902:
LGTM, approving 👍 
Pls review my last changes to the docs.

Issue #268, opened 2017-12-31
Refactoring pair name handling


jblestang:
Code is crippled with` pair.replace('_', '-'),` what about adding a method in the exchange class that defines the delimiter and remove `pair.replace` in all the code ?

gcarq:
Sounds good to me

jblestang:
after much thinking, I think the name of the pair should be automatically created by the bot. We define the **stake_currency** as for example BTC, and the list of pairs needs to be only something like *[ "XVG" ,  "XRP" ]*. The bot can then form the correct pair

gcarq:
Thats a good idea, it also gives more flexibility when switching the stake_currency on a regular base.

hroff-1902:
This issue consists of 2 diff. parts: the original message and the proposal in the comment (5 Jun 2018).

First one is not actual for long time: mapping of symbols in the pair names is only made when the filename is generated.

Second one is better be refused: there can be more than one market on the exchange with the same quote currency and our base (stake) currency.

xmatthias:
Funnily, i've been looking at this issue just this week, and came to a different conclusion.

`pair.replace("/", "_")` is used twice (3 times) - once to saveand load data, and once as plot filename - so i didn't see it as anything that needs action (we've had this code-segment around 20 times in the code - mostly in places where it wasn't needed) - and that's when this issue was created.

For the 2nd part - i've had the dream to create a Pair class (my experiments around that are almost 6 months old ...).
I think it would simplify a lot of things ... we'd have `pair.quote`, or `pair.as_filename` or ...

I've played around with that part a bit and i think it would work - all while supporting both `XRP/BTC` and `XRP` (+ stake currency) configuration in the pairlist (as proposed above) - which would actually be very easy to do if it's a class creation.

The part i didn't investigate yet is how to integrate that with the database (maybe `str(pair)` ...) but we're not creating new trades that often, so i guess that would work fine, too.

hroff-1902:
First part is completely not actual for the current code base, right -- I guess it was actual more than an year ago, when the codebase contained many such pair name conversions, it's not an issue anymore.

The second proposal:

[...]

PR #2719, opened 2019-12-28
Introduce Data handler

xmatthias:
## Summary
This Pullrequest completely revises how freqtrade works with stored data.

#### Current approach

The current approach is to download data as lists, and store these list as json (or jsongz) files.
Similarily, we load data into a list, and then convert it to a DataFrame.

#### New approach
All ohlcv datahandling happens as dataframes.
This avoids the need to convert data for backtesting from list to dataframe, which strangely speeds up backtesting of 30 pairs for 2.5 months (data for 2 years is available) by roughly 1 second (from 30 to 28.9 s), even though rapidjson is not used anymore.

Duplicate candles are not stored to disk (this is a side-effect i discovered while working on this):
Assuming a new pair was introduced 3 weeks ago and download-data is run with `--days=30` (4 weeks). We would end up having the first 1000 candles multiple times for a week in the file. It's not an issue for Freqtrade as this data is removed once data is loaded from disk, but it still slows down loading of data, and consumes unnecessary space.

This new approach also allows to remove a ton of duplicate code (`trim_dataframe` / `trim_tickerlist` (latter is not needed anymore)).

## Quick changelog

- Introduce `IDataHandler` and `JsonDataHandler` respectively
  - they handle all file operations and return the data as Dataframe (OHLCV) or list of dicts (Trades) (For now - trades behaviour may change in the future).
  - By using dataframes, new handlers can easily be added since pandas supports many formats out of the box (for now, only the currently available formats have been implemented.
- split `clean_ohlcv_dataframe` from `parse_ticker_dataframe` to allow seperate validation of a dataframe.
- move `trim_dataframe` from `history.py` to `converter.py` - since it's converting stuff
- Introduce subcommand `freqtrade convert-data` and `freqtrade convert-trades-data` to convert data from one format to the other
- convert `history.py` from being a module to a package to encapsulate all historic data handling
- No change to user interface (`freqtrade.data.history.load_data()`) for analysis tasks is untouched (unless the format is changed), since this returned a dataframe already.

## Future ideas (not for this pr)
- Support more dataformats 
  - Ideally these formats should support loading only partial data - so if i need 1 month, i only need to read 1 month's worth of data
- return dataframe from `exchange.get_historic_ohlcv()`. Not done in this pr as that would blow up the PR even more and the change is in the end minimal.
- Implement `***_append()` methods to allow capturing data from live operations (we fetch 500 candles every iteration - we might as well save these).

## Note:
[...]

PR #2744, opened 2020-01-04
Add support for downloading data for markets with 'bad' characters in the market symbol

hroff-1902:
* Introduce `pair_to_filename()` function which maps them to `'_'` for pairname to be used in the filename.
* This is also for the name of the resulting html in plotting.

The 'bad' symbols can be seen as:
```
for e in `freqtrade list-exchanges -1`; do freqtrade list-markets --exchange $e -1 --all 2>/dev/null; done | tr -d '\na-zA-Z0-9/-'
```

Closes first part of #268 (mapping of the symbols in the pair names now located in a single function and only used in the filenames).

TODO:

* Pls review this part in persistence:
https://github.com/freqtrade/freqtrade/blob/3798f94d4c52fb0cf441d2d4e3c3f0218d9005a8/freqtrade/persistence.py#L129-L134
Should something be changed here taking into account that more than one `'_'` can appear in the pairname after mapping `'/'` to `'_'`? Is it really actual? (I have not found this mapping)

* Maybe `pair_to_filename()` should be moved into `misc.py` since it's not only about historical datafiles, but also used in plotting.

### DIFF
--- a/freqtrade/data/history.py
+++ b/freqtrade/data/history.py
@@ -224,14 +224,20 @@ def refresh_data(datadir: Path,
                                exchange=exchange)
 
 
+def pair_to_filename(pair: str) -> str:
+    for ch in ['/', ' ', '.']:
+        pair = pair.replace(ch, '_')
+    return pair
+
+
 def pair_data_filename(datadir: Path, pair: str, timeframe: str) -> Path:
-    pair_s = pair.replace("/", "_")
+    pair_s = pair_to_filename(pair)
     filename = datadir.joinpath(f'{pair_s}-{timeframe}.json')
     return filename
 
 
 def pair_trades_filename(datadir: Path, pair: str) -> Path:
-    pair_s = pair.replace("/", "_")
+    pair_s = pair_to_filename(pair)
     filename = datadir.joinpath(f'{pair_s}-trades.json.gz')
     return filename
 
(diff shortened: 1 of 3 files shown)


xmatthias:
I would rather base this change on top of #2719, otherwise it'll be a mess to merge, since the pair saving functions are moved to a completely different area...

Then, it doesn't fit well into history.py, misc.py is definitely a better place for a "general" function like this (it's not just used for history, but for plotting, too).

Persistence is not of relevance here i think.
That "translation" is to move the old pair format (used more than a year ago, if not more) to the new database pair format - so this transformation / replace will ONLY apply on old databases - and i doubt that these versions worked with these "odd" pairs anyway (i doubt our current version will work with these pairs ...).

hroff-1902:
> I would rather base this change on top of #2719, otherwise it'll be a mess to merge, since the pair saving functions are moved to a completely different area...

Okay, no problem, I'll close it for now.

> Then, it doesn't fit well into history.py, misc.py is definitely a better place for a "general" function like this (it's not just used for history, but for plotting, too).

Okay, that was also my idea that it better fit to somewhere outside of data.history.

> Persistence is not of relevance here i think.
> That "translation" is to move the old pair format (used more than a year ago, if not more) to the new database pair format - so this transformation / replace will ONLY apply on old databases - and i doubt that these versions worked with these "odd" pairs anyway (i doubt our current version will work with these pairs ...).

Fine, no actions are thus will be needed here.
👍

PR #2752, opened 2020-01-07
Allow enhanced plot-dataframe configuration

### LINKED ISSUE #1531: Plotting: Indicators should be defined in Strategy
I think we should have plot indicators details defined directly in Strategy file. That way the user can define more specific plot configuration for each strategy.

Features:

- Dynamic Row / Subplot number
- User defined indicators color
- Possibility to show Strategy related indicators on upcoming plotting web ui #1505 

Below the variable proposition:

    # Optional plot informations
    plot_rows = [
        'main_plot':[
            {'name': 'ema10', 'color': 'red'}
        ],
        'sub_plots':[
            [
                {'name': 'macd', 'color': 'blue'},
                {'name': 'macdsignal', 'color': 'red'}
            ],
            [
                {'name': 'rsi', 'color': 'orange'}
            ],
        ]
    ]

xmatthias:
## Summary
Enhanced plot-dataframe configuration matching the strategy should be created within the strategy, as proposed in #1531.

this also allows us to define multiple "subplots" - as shown in the 2nd Screenshot.

Closes #1531

## Quick changelog

- Allow enhanced plotting configuration in strategy
  - Allow setting fixed color
  - allow multiple subplots for different indicators (MACD / RSI for example)
- move defaults of indicators to `plotting.py` - they should only override plot_config if explicitly selected
- change display of trades - Trade open is now a cyan circle, and Trade Sells differ between red and green squares, allowing to identify good / bad trades easily.
- Added sample configuration to sample_strategy and full template


xmatthias:
> I was also thinking of renaming 'main_plot' into 'main', but I guess it also needs to contain 'plot' in the name in the same time...

not necessarily - it's part of the `plot_config` dict after all, but i think `main_plot` is better aligned to `subplots` - we're not going to call that `subs` ... :laughing:

PR #2763, opened 2020-01-12
Fix/precision rounding

### LINKED ISSUE #2759: FTX symbol_amount_prec wrong

Yes

  * Operating system: ubuntu19.10
  * Python Version: 3.7
  * CCXT version: latest
  * Branch: Develop
  
Bot found buy opportunities but can't buy using market 

### Steps to reproduce:

  1. Run with linkbull/usd pair on ftx
  2. Wait
 
  
### Observed Results:

  * What happened? It said couldn't buy 0.0
  * What did you expect to happen? To buy as usual. Do know that you can buy as small as 1 dollar on ftx spot markets (even less) 

  ```
  2020-01-11 08:13:20,316 - freqtrade.freqtradebot - INFO - Buy signal found: about create a new trade with stake_amount: 20 ...
2020-01-11 08:13:20,317 - freqtrade.freqtradebot - INFO - Using Last Ask / Last Price
2020-01-11 08:13:20,469 - freqtrade.freqtradebot - WARNING - Unable to create trade for LINKBULL/USD: Could not create market buy order on market LINKBULL/USD.Tried to buy amount 0.0 at rate None.Message: ftx {"error":"Size too small","success":false}                                                                                 
2020
  ```

xmatthias:
## Summary
CCXT does have multiple methods to handle precision. see https://github.com/ccxt/ccxt/issues/6364
Based on this - we should not implement these functions ourself, but rely on CCXT for doing so.

Closes #2759

## Quick changelog

- rename methods from `symbol_price_prec` to `price_to_precision` and `symbol_amount_prec` to `amount_to_prec`
- use ccxt's `decimal_to_precision()` for this calculation, which handles all the logic of using the correct method based on each exchange.

## Open / failing tests

Currently there are 2 tests failing.
This is because the old method always rounds price up (so 0.123 would become 0.13 instead of 0.12). 

Right now I'm now questioning this method - since i'm not convinced that's correct.

Now i remember we discussed this topic in the introduction PR #1083 - that amount should round down (truncate - especially important for sells - we can't round up as we'll never have "more" than we bought - even after fees are deduced).

Also in that PR: price should round up - however i don't really see the benefit in this - so additional opinions are highly apreciated.

The 2 tests which are failing show that the "correct" test did not properly cover this scenario - which we should do by testing a few "edge-cases" (or not test this as it's now part of CCXT code).

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -14,8 +14,8 @@ from typing import Any, Dict, List, Optional, Tuple
 import arrow
 import ccxt
 import ccxt.async_support as ccxt_async
-from ccxt.base.decimal_to_precision import (ROUND, ROUND_DOWN, ROUND_UP,
-                                            TRUNCATE, TICK_SIZE, decimal_to_precision)
+from ccxt.base.decimal_to_precision import (ROUND_DOWN, ROUND_UP, TICK_SIZE,
+                                            TRUNCATE, decimal_to_precision)
 from pandas import DataFrame
 
 from freqtrade.data.converter import parse_ticker_dataframe
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -7,14 +7,15 @@ import inspect
 import logging
 from copy import deepcopy
 from datetime import datetime, timezone
-from math import ceil, floor
+from math import ceil
 from random import randint
 from typing import Any, Dict, List, Optional, Tuple
 
 import arrow
 import ccxt
 import ccxt.async_support as ccxt_async
-from ccxt.base.decimal_to_precision import ROUND_DOWN, ROUND_UP
+from ccxt.base.decimal_to_precision import (ROUND, ROUND_DOWN, ROUND_UP,
+                                            TRUNCATE, decimal_to_precision)
 from pandas import DataFrame
 
 from freqtrade.data.converter import parse_ticker_dataframe
@@ -367,26 +368,29 @@ class Exchange:
         """
         return endpoint in self._api.has and self._api.has[endpoint]
 
-    def symbol_amount_prec(self, pair, amount: float):
+    def symbol_amount_prec(self, pair, amount: float) -> float:
         '''
         Returns the amount to buy or sell to a precision the Exchange accepts
         Rounded down
         '''
         if self.markets[pair]['precision']['amount']:
-            symbol_prec = self.markets[pair]['precision']['amount']
-            big_amount = amount * pow(10, symbol_prec)
-            amount = floor(big_amount) / pow(10, symbol_prec)
+            amount = float(decimal_to_precision(amount, rounding_mode=TRUNCATE,
+                                                precision=self.markets[pair]['precision']['amount'],
+                                                counting_mode=self.precisionMode,
+                                                ))
+
         return amount
 
-    def symbol_price_prec(self, pair, price: float):
+    def symbol_price_prec(self, pair, price: float) -> float:
         '''
         Returns the price buying or selling with to the precision the Exchange accepts
         Rounds up
         '''
         if self.markets[pair]['precision']['price']:
-            symbol_prec = self.markets[pair]['precision']['price']
-            big_price = price * pow(10, symbol_prec)
-            price = ceil(big_price) / pow(10, symbol_prec)
+            price = float(decimal_to_precision(price, rounding_mode=ROUND,
+                                               precision=self.markets[pair]['precision']['price'],
+                                               counting_mode=self.precisionMode,
+                                               ))
         return price
 
     def dry_run_order(self, pair: str, ordertype: str, side: str, amount: float,
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -7,7 +7,6 @@ import inspect
 import logging
 from copy import deepcopy
 from datetime import datetime, timezone
-from math import ceil
 from random import randint
 from typing import Any, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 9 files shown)


xmatthias:
I've been thinking about the rounding price up or down - and i don't think it's relevant which direction we're rounding, as i don't think this will ever apply in our case.

We get the price from either `fetch_ticker(pair)` - or via orderbook.
Both of these prices will have to adhere to the price precision from the exchange.

While we do a slight calculation on price in `get_target_bid()` - rounding mathematically should be fine for this case as well, since the purpose of this is to use a slightly modified price.

hroff-1902:
> I've been thinking about the rounding price up or down - and i don't think it's relevant which direction we're rounding

I'm working on it and I'm of opposite opinion...
(More details further)

hroff-1902:
> I've been thinking about the rounding price up or down - and i don't think it's relevant which direction we're rounding, as i don't think this will ever apply in our case.
> 
> We get the price from either `fetch_ticker(pair)` - or via orderbook.
> Both of these prices will have to adhere to the price precision from the exchange.
> 
> While we do a slight calculation on price in `get_target_bid()` - rounding mathematically should be fine for this case as well, since the purpose of this is to use a slightly modified price.

One of the examples when rounding can have sense:

Stoploss on exchange. For long trades, it should be rounded up, otherwise we get bigger loss than expected.

For short trades (when they will be implemented), vise versa, the stoploss price should be rounded down.

Besides rounding prices, the ccxt offers (and implementation of create_order() implements for at least some exchanges) rounding of amount/cost (which is amount of quote currency in terms of ccxt). These should also be rounded in completely different manner, depending on the particular case: for example, when we create a new trade, the amount/cost should probably be rounded down, to avoid situations when we exceed available balance for some small fraction of it, but when/if we implement amending of the existing trade -- this should probably be rounded mathematically to obtain the result which is more close to the calculated float one.

Rounding fees is a complete different story.

In my opinion, rounding should be carefully chosen by the application using ccxt. Ccxt offers (and uses internally) a family of methods: cost_to_precision, fee_to_precision, price_to_precision, amount_to_precision, but it incorrectly makes assumptions on how these values are rounded. This should fully be controlled by the app, since it differs for every particular case.

hroff-1902:
In the meantime I can propose you to return back the compatibility and use the same rounding mode used previously in every case. I'll work with internals of ccxt regarding these issues, implementation of rounding implemented in ccxt seems to be buggy in some cases, it simply makes assumptions it should not, and some ccxt exchanges try to fix it in their implementations internally. All this should be carefully refined.

The main idea to use ccxt functions instead of own calculations is completely fine. 👍

xmatthias:
> for example, when we create a new trade, the amount/cost should probably be rounded down, to avoid situations when we exceed available balance for some small fraction of it, but when/if we implement amending of the existing trade -- this should probably be rounded mathematically to obtain the result which is more close to the calculated float one.

This point does not make sense.
Rounding mathematically will not work for the amended trade (it will work for all other trades - but not for the last (assuming 0 reserved balance).

Assume you only have 23.77 USDT in your balance. By rounding up either amount or price, we would end up with a cost of 23.8 USDT (assuming a precision of 1 for simplicity) - which is something not available.

> Stoploss on exchange. For long trades, it should be rounded up, otherwise we get bigger loss than expected.

well it depends - ideally you're only using pairs where 1pip is < 0.1% - so the maximum additional loss is 0.099% (+ the limit set on top of the stoploss - for stoploss-on-exchange orders). 
by always rounding up - you're pulling stoploss closer towards the opening price - which can lead to early stopping - which - depending on the strategy, can lead to disastrous results - especially when "unreccomended" pairs (1pip > 1%) are used - which will pull a 5% stoploss to 4% just by rounding.

Personally, I'd prefer mathematical rounding in these cases - at lest that won't round a stoploss of -4.99% to -4%.

> In the meantime I can propose you to return back the compatibility and use the same rounding mode used previously in every case

I'm not certain about how you think this is supposed to work without reimplementing the almost complete `decimal_to_precision()` function within freqtrade.

for example -  ccxt supports a precision of 0.005 (mentioned by kroitor in the issue i opened regarding this).
This would mean that for 20.1234, possible values will be 20.125 or 20.120.

We did not have this in the past as we were not aware of this type of "precisionMode".

It's also not an issue for supported exchanges (binance, bitmex, ...), but only for a small subset of exchanges.

These precision modes apply for:
['bitmex', 'deribit', 'ftx'] -> precisionMode = TICK_SIZE
['bitfinex', 'bitfinex2'] -> precisionMode = SIGNIFICANT_DIGITS

What i mean by this is that i don't see any urgency for this PR to be merged.

> I'll work with internals of ccxt regarding these issues, implementation of rounding implemented in ccxt seems to be buggy in some cases, it simply makes assumptions it should not, and some ccxt exchanges try to fix it in their implementations internally

Well ideally decimal_to_precision will support `ROUND_UP` and `ROUND_DOWN` - which are constants that are defined for the `round_timeframe()` method - but i think would make sense for `decimal_to_precision()` as well.

tests/exchange/test_exchange.py
```diff
@@ -173,35 +173,74 @@ def test_validate_order_time_in_force(default_conf, mocker, caplog):
     ex.validate_order_time_in_force(tif2)
 
 
-def test_symbol_amount_prec(default_conf, mocker):
+@pytest.mark.parametrize("amount,precision_mode,precision,expected", [
```
xmatthias: I did add a few more tests ... I think for amount we're clear (it should truncate - thats how it's always been) - however it should support the "new" method as well.

I've also added the tests for price_to_precision - but these may be wrong depending if we decide to "round up" ...

freqtrade/exchange/exchange.py
```diff
@@ -362,32 +368,49 @@ def exchange_has(self, endpoint: str) -> bool:
...
         '''
         if self.markets[pair]['precision']['price']:
-            symbol_prec = self.markets[pair]['precision']['price']
-            big_price = price * pow(10, symbol_prec)
-            price = ceil(big_price) / pow(10, symbol_prec)
+            # price = float(decimal_to_precision(price, rounding_mode=ROUND,
+            #                                    precision=self.markets[pair]['precision']['price'],
+            #                                    counting_mode=self.precisionMode,
+            #                                    ))
+            if self.precisionMode == TICK_SIZE:
```
xmatthias: @hroff-1902 please have a look now. 

price is now rounding up using the previous method (however with support for TICK_SIZE.
I've also adapted the tests for this and it seems to work as expected for all cases i tested (which are all also test-cases ...) - however feel free to point out missing cases.

PR #2771, opened 2020-01-15
Fix bad bug in safe_sell_amount

### LINKED ISSUE #2770: Unable to sell trade: Not enough amount to sell.
Hi!
I use latest build of freqtrade: 09621b3. OS ubuntu, exchange Binance.
On all sales order it write:
freqtrade.freqtradebot - WARNING - Unable to sell trade: Not enough amount to sell.

xmatthias:
## Summary
We MUST compare selling amount to wallet amount with >= (not >) - otherwise regular trades without "dust" are unable to be sold.

supports #2770

## Quick changelog

- Fix bug
- add test

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -903,7 +903,7 @@ class FreqtradeBot:
         :raise: DependencyException: if available balance is not within 2% of the available amount.
         """
         wallet_amount = self.wallets.get_free(pair.split('/')[0])
-        logger.debug(f"{pair} - Wallet: {wallet_amount} - Trade-amount: {amount}")
+        logger.info(f"Selling {pair} - Wallet: {wallet_amount} - Trade-amount: {amount}")
         if wallet_amount >= amount:
             return amount
         elif wallet_amount > amount * 0.98:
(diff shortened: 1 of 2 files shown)


hroff-1902:
Hmm, but in case wallet_amount is equal to amount, the "Falling back to wallet-amount." was to be issued and wallet_amount (which is equal to amount) returned anyway...

xmatthias:
which would leave is with "wallet_amount" was somehow not corresponding to the exchange?

well, i've now upped the message level of the "wallet / trade-amount" message now ... in hopes that we can diagnose should it happen again (nevertheless, "Falling back to wallet-amount" message would be misleading in that case).

I'll also continue to debug this ... but so far it seems ok on my account, but who knows what causes this precisely ...

hroff-1902:
I mean, if we have wallet=1000, amount=1000, then
* previous behavior: fallback to 'elif' block, log "Falling back to wallet-amount", return wallet_amount (1000),
* new behavior: go to 'if' block, return amount (which is also 1000)...

So what's tzimmes here?

xmatthias:
well yes - you're right, it'll most likely not fix the bug.
The tree (and with that the log-message) is wrong nonetheless ... 

However, it should give us the logs we need to figure out the bug should it reappear.

hroff-1902:
Okay 👍

PR #2779, opened 2020-01-19
Stoploss on exchange for Kraken

xmatthias:
## Summary
Implement stoploss orders for Kraken

Solve the issue: #1687 (point 3 - stop-loss)

## Quick changelog

- Move stoploss-limit calculation to stoploss-function within exchange class to allow the interface to work for other stoploss-types too
- Rename `stoploss_limit()` to `stoploss()`
- Use named arguments for create_order call 
- Implement stoploss for kraken (it's a stoploss-market order, since that's the only stoploss type kraken supports).


tests/test_freqtradebot.py
```diff
@@ -1459,7 +1462,7 @@ def test_tsl_on_exchange_compatible_with_edge(mocker, edge_conf, fee, caplog,
     cancel_order_mock = MagicMock()
     stoploss_order_mock = MagicMock()
     mocker.patch('freqtrade.exchange.Exchange.cancel_order', cancel_order_mock)
-    mocker.patch('freqtrade.exchange.Exchange.stoploss', stoploss_order_mock)
+    mocker.patch('freqtrade.exchange.Binance.stoploss', stoploss_order_mock)
```
hroff-1902: Should this test be parametrized for the exchanges with stoploss on exchange implemented (Binance, Kraken)?

xmatthias: i don't think so... 
this does not test the exchange-specific stoploss code - but the handling of stoploss within freqtradebot - which is identical for both exchanges.
Since the stoploss-method is mocked away, there can't be a difference in behaviour this test would be able to catch.

There is another way to fix this test - by mocking `stoploss_adjust()` too and keeping this test on the default exchange class (as it was before) - which basically calls 2 methods that are not implemented (throws OperationalException when not mocked).

Now since `stoploss_adjust()` has no side-effects and is just a small helper method, i didn't see the need to mock that, too.

freqtrade/exchange/binance.py
```diff
@@ -32,13 +32,23 @@ def get_order_book(self, pair: str, limit: int = 100) -> dict:
...
+    def stoploss(self, pair: str, amount: float, stop_price: float, order_types: Dict) -> Dict:
         """
         creates a stoploss limit order.
         this stoploss-limit is binance-specific.
         It may work with a limited number of other exchanges, but this has not been tested yet.
-
         """
+        # Limit price threshold: As limit price should always be below stop-price
+        LIMIT_PRICE_PCT = order_types.get('stoploss_on_exchange_limit_ratio', 0.99)
+        rate = stop_price * LIMIT_PRICE_PCT
```
hroff-1902: I don't like this in the capital letters (`LIMIT_PRICE_PCT`), maybe simple `limit_price_pct` is better...
(I noticed where you copied this from, but let's clean this up...)

docs/exchanges.md
```diff
@@ -22,6 +22,9 @@ Binance has been split into 3, and users must use the correct ccxt exchange ID f
 
 ## Kraken
 
+!!! Tip "Stoploss on Exchange"
+    Kraken supports `stoploss_on_exchange` and uses stop-loss-market orders. It provides great advantages, so we recommend to benefit from it, however since the resulting order is a stoploss-market order, sell-rates are not guaranteed, which makes this feature less secure than on other exchanges. This limitation is based on kraken's policy [source](https://blog.kraken.com/post/1494/kraken-enables-advanced-orders-and-adds-10-currency-pairs/) and [source2](https://blog.kraken.com/post/1494/kraken-enables-advanced-orders-and-adds-10-currency-pairs/) - which has stoploss-limit orders disabled.
```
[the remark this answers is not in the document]
xmatthias: well, you know - copy/paste is difficult sometimes :D

PR #2787, opened 2020-01-20
remove default value calls for dry_run

xmatthias:
## Summary
Remove .get calls for dry_run - it's a mandatory property

I'm certain there are other properties where the same applies - but one step at a time ...


hroff-1902:
Is this just a code cleanup, not caused by any bugs?

xmatthias:
> Is this just a code cleanup, not caused by any bugs?

yes absolutely - there is no real benefit - but this way it's aligned.
we've used it this way throughout the exchange class, so not having dry_run would fail anyway

PR #2796, opened 2020-01-22
update wallets after forcesell

### LINKED ISSUE #2793: Forcesell didn't update Balance

Not found a similar question

  * Operating system: Ubuntu 18.04.3 LTS
  * Python Version: 2.7.17 / 3.6.9____ (`python -V`)
  * Branch: Master
  * Last Commit ID: 7c53dcb0afd2ad663faad4bc03406f0575c87f6b(`git log --format="%H" -n 1`)
  
After /Forcesell XX Balance isn't updated.
Before Forcesell : 
BTC:
 Available:  0.00274315
After Forcesell : 
BTC:
 Available:  0.00274315

Balance on Exchange (Binance) 
0.00326254

Looks like /forcesell sells the given Amount but didn't update the Balance (BTC).

### Steps to reproduce:

  1. /balance
  2. /forcesell XX (Telegram-Bot)
  3. /balance

xmatthias:
## Summary
Since forcesell calls `execute_sell()` we need to update wallets after this to avoid keeping the old wallet status until the next action happens. 
It's not a problem if the order doesn't fill immediately since `update_trade_state()` will update the state too - but it's a problem if the order does fill immediately, as then we close the trade.
In regular operations, wallets are updated after handling all "sellable" trades - but that's not the case with forcesell.

closes #2793

# Quick changelog

* call wallets.update when /forcesell is completed
* implement threading.lock object to avoid stoploss recreation when /forcesell is being used

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -7,6 +7,7 @@ import traceback
 from datetime import datetime
 from math import isclose
 from os import getpid
+from threading import Lock
 from typing import Any, Dict, List, Optional, Tuple
 
 import arrow
@@ -26,7 +27,6 @@ from freqtrade.rpc import RPCManager, RPCMessageType
 from freqtrade.state import State
 from freqtrade.strategy.interface import IStrategy, SellType
 from freqtrade.wallets import Wallets
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


hroff-1902:
LGTM as the current solution, 👍 

However, I have a better idea (partly done in my stashes) -- RPC should only request the actions (buy/sell/change state, etc), all actions should be done then in the main worker thread in the line with normal throttling logic (and. thus, in the same thread).

xmatthias:
> However, I have a better idea (partly done in my stashes) -- RPC should only request the actions (buy/sell/change state, etc), all actions should be done then in the main worker thread in the line with normal throttling logic (and. thus, in the same thread).

well if you look at it realistically, it's only really impacting forcebuy and forcesell. state change (switching from running to stopped) is already a "request" (worker is stopping depending on the value of state), as well as reload_conf - and all other rpc commands are just reporting stuff.

It's a good idea though, one i've thought about myself as well - but never had time (nor motivation) to implement.

PR #2799, opened 2020-01-23
Fix stoploss recreated

xmatthias:
## Summary
Combined with the 2nd part of #2796 (Locking) - this will prevent stoploss orders from being recreated if a sell is currently in place (but not yet filled).

There are several protections in place at the moment - however in case of trailing stoploss it can happen that the bot places a regular sell limit order which is not filled immediately.
The next iteration will see that the stoploss is gone and try to recreate a stoploss order.

The other case (the one i hit):
Sell-order limit is placed (obviously, cancelling stoploss_on_exchnage).
The next iteration, the limit-order filling is confirmed (calls `trade.close()` - which sets `is_open=False` and `open_order_id=None`, causing that protection to fail.

is_open is verified in `handle_trade()` - but was missing here.

For me, it created a stoploss sell order on the exchange again - and since i had balance in that coin, the creation did not fail, and would have sold the "bought amount" a second time.

## Quick changelog

- Make sure we're not (re)creating stoploss orders on closed trades.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -689,8 +689,13 @@ class FreqtradeBot:
             self._notify_sell(trade, "stoploss")
             return True
 
+        if trade.open_order_id:
+            # Trade has an open Buy or Sell order, Stoploss-handling can't happen in this case
+            # as the Amount on the exchange is tied up in another trade.
+            return False
+
         # If buy order is fulfilled but there is no stoploss, we add a stoploss on exchange
-        if (not trade.open_order_id and not stoploss_order):
+        if (not stoploss_order):
 
             stoploss = self.edge.stoploss(pair=trade.pair) if self.edge else self.strategy.stoploss
 
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -689,9 +689,10 @@ class FreqtradeBot:
             self._notify_sell(trade, "stoploss")
             return True
 
-        if trade.open_order_id:
+        if trade.open_order_id or not trade.is_open:
             # Trade has an open Buy or Sell order, Stoploss-handling can't happen in this case
             # as the Amount on the exchange is tied up in another trade.
+            # The trade can be closed already (sell-order fill confirmation came in this iteration)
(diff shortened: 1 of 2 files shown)


hroff-1902:
@xmatthias 
can you describe the sequence of actions this PR protects from? 
(like the trailing stop is places (btw -- on exchange or freqtrade-managed?); then it's moved or triggered, then the sell order is created, not fulfilled/partially filled, etc,,,)

xmatthias:
> The other case (the one i hit):
Sell-order limit is placed (obviously, cancelling stoploss_on_exchnage).
The next iteration, the limit-order filling is confirmed (calls trade.close() - which sets is_open=False and open_order_id=None, causing that protection to fail. 

I thought i did ... ? 

Anway - with a bit more detail:
* you got an open trade, `stoploss_on_exchange` is enabled, as well as trailing stop
* the trade gets sold via sell-signal (or roi)
  * this cancels the stoploss on exchange
  * creates a limit sell order (which does NOT immediately fill - this is important as otherwise `update_trade_state()` is called immediately and the below does not apply anymore)
* in the next iteration 
  * `update_trade_state()` - called from `exit_positions()` - updates the trade state to closed (`trade.is_open=False`).
  * then `handle_stoploss_on_exchange()` is called on the now closed trade.
  * since the checks for closed trades are not in place, it recreates the stoploss on the exchange order (now trying to sell  funds the bot never bought).
  * this can have 2 outcomes: 
    1. stoploss_on_exchange order creation fails with "not enough funds" (this will happen in most cases)
    1. stoploss_on_exchange order creation is successfull (i had XRP in my accound, and the trade was XRP/ETH) (this happened in my case - but depends on the funds on the exchange).


Technically - this change "only" replicates behaviour of `handle_trade()` - which checks if trade.is_open is false and does not act in that case.

hroff-1902:
Now I see the problem, thanks 👍

xmatthias:
> (too specific/internal to mark it as Enhancement?)

i think so, yes ... it's "just" a bugfix of a edge-case ... (it's not so much a problem if the creation fails ...)

PR #2800, opened 2020-01-25
better logging on check_depth_of_market_buy()

yazeed:
Added ask price, bid price, immediate ask quantity, and immediate bid quantity to check_depth_of_market_buy.

Also, the method now notifies the user if the delta condition was satisfied or not.

Screenshot 2020-01-25 04 33 22

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -423,11 +423,14 @@ class FreqtradeBot:
         order_book_bids = order_book_data_frame['b_size'].sum()
         order_book_asks = order_book_data_frame['a_size'].sum()
         bids_ask_delta = order_book_bids / order_book_asks
-        logger.info('bids: %s, asks: %s, delta: %s', order_book_bids,
-                    order_book_asks, bids_ask_delta)
+        logger.info('bids: %s, asks: %s, delta: %s, askprice: %s, bidprice: %s, immediate askquantity: %s, immediate bidquantity: %s',
+                    order_book_bids, order_book_asks, bids_ask_delta, order_book['asks'][0][0], order_book['bids'][0][0], order_book['asks'][0][1], order_book['bids'][0][1])
         if bids_ask_delta >= conf_bids_to_ask_delta:
+            logger.info('bids to ask delta DOES satisfy condition.')
             return True
-        return False
+        else:
+            logger.info('bids to ask delta DOES NOT satisfy condition.')
+            return False
 
     def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = None) -> bool:
         """
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -423,8 +423,12 @@ class FreqtradeBot:
         order_book_bids = order_book_data_frame['b_size'].sum()
         order_book_asks = order_book_data_frame['a_size'].sum()
         bids_ask_delta = order_book_bids / order_book_asks
-        logger.info('bids: %s, asks: %s, delta: %s, askprice: %s, bidprice: %s, immediate askquantity: %s, immediate bidquantity: %s',
-                    order_book_bids, order_book_asks, bids_ask_delta, order_book['asks'][0][0], order_book['bids'][0][0], order_book['asks'][0][1], order_book['bids'][0][1])
+        logger.info(
+            f"bids: {order_book_bids}, asks: {order_book_asks}, delta: {bids_ask_delta}, "
+            f"askprice: {order_book['asks'][0][0]}, bidprice: {order_book['bids'][0][0]}, "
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -423,11 +423,14 @@ def _check_depth_of_market_buy(self, pair: str, conf: Dict) -> bool:
...
-        logger.info('bids: %s, asks: %s, delta: %s', order_book_bids,
-                    order_book_asks, bids_ask_delta)
+        logger.info('bids: %s, asks: %s, delta: %s, askprice: %s, bidprice: %s, immediate askquantity: %s, immediate bidquantity: %s',
+                    order_book_bids, order_book_asks, bids_ask_delta, order_book['asks'][0][0], order_book['bids'][0][0], order_book['asks'][0][1], order_book['bids'][0][1])
         if bids_ask_delta >= conf_bids_to_ask_delta:
+            logger.info('bids to ask delta DOES satisfy condition.')
             return True
-        return False
+        else:
+            logger.info('bids to ask delta DOES NOT satisfy condition.')
```
xmatthias: I would avoid "screaming" in log messages (all uppercase words).  

Also, including the pair to this log message will make searching through logs easier as you can search for the pair that did / did not not buy ...

yazeed: Done, not sure why flake8 ended up modifying lots of other lines on the file though for length of line.

yazeed:
If you would like to add the configuration for flake8 to the repo, we can all abide by it. I've used the default.

Ref: 
Screenshot 2020-01-25 23 05 23

hroff-1902:
hmm...
https://github.com/freqtrade/freqtrade/blob/f6278da23f0698d9f7e9fa45ec8edeff27a854ba/setup.cfg#L1-L9

freqtrade/freqtradebot.py
```diff
@@ -54,16 +54,19 @@ def __init__(self, config: Dict[str, Any]) -> None:
 
         self._heartbeat_msg = 0
 
-        self.heartbeat_interval = self.config.get('internals', {}).get('heartbeat_interval', 60)
+        self.heartbeat_interval = self.config.get("internals", {}).get(
+            "heartbeat_interval", 60)
```
xmatthias: I'm not a huge fan of this change.

I don't mind which quotes we're using necessarily (although i don't see a huge reason to change where we got them - but we've got default settings (with line-lenght to 100 - as pointed out by @hroff-1902).
 This seems to reduce the line-length to 80ish (probably 88, judging from your screenshot) - causing a lot of unnecessary linebreaks (like this one - but there's a few others before even getting to line 100)

hroff-1902:
right, I simply run 
```
flake8; mypy; pytest
```
with `cd freqtrade` (where freqtrade is the dir with that setup.cfg) made before if I'm in another dir...

freqtrade/freqtradebot.py
yazeed: I'll fix it to 100 characters per line. Then we can take a look at it again.

yazeed:
> hmm...
> https://github.com/freqtrade/freqtrade/blob/f6278da23f0698d9f7e9fa45ec8edeff27a854ba/setup.cfg#L1-L9

I didn't notice that, my bad. 😄 So, I just did another commit showing what it looks like after changing it to 100 characters.

Still not consistent with your style in terms of brackets, what am I missing now? I'm using Visual Studio Code by the way.

xmatthias:
well flake8 will not change your code - it'll just show the wrong lines.

What comand exactly did you run? (also within vscode ...)?

The changes look to me as if [black](https://github.com/psf/black) was run on this file ...
I've looked at black in the past - but the configuration must be done in a pyproject.toml file, which breaks installation (at least it did when i looked at it) unless the whole installation is rewritten to be based on this new format.

I don't think changes done by black (or any other formatter, for that matter) are easily revertable by a tool other than git (`git rebase -i <first_commit>`) and skipping the wrong commit...

hroff-1902:
@yazeed maybe it's better to create a new PR with only small change relevant to the title, without all this reformatting, if you are not familiar with git rebase?

xmatthias:
Now let me help you out here a little bit - i've rebased (removing the 2 illicit commits) and forcepushed - so i think this should be better now.

Note that i've also changed the logs to be f-strings - performance (and simplicity to read) for these is superior so we try to change to that whenever we happen to touch some lines.

yazeed:
> Now let me help you out here a little bit - i've rebased (removing the 2 illicit commits) and forcepushed - so i think this should be better now.
> 
> Note that i've also changed the logs to be f-strings - performance (and simplicity to read) for these is superior so we try to change to that whenever we happen to touch some lines.
> 
> Edit: And congratulations - you've got PR/issue #2800.
> It would be a shame to close that ... 😄

Thank you @xmatthias, you're awesome. :)

PR #2807, opened 2020-01-27
Add notify_status() to FreqtradeBot

hroff-1902:
Refactor status notification interface method which is to be used by worker.

This simplifies worker and reduces its dependencies.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -95,6 +95,16 @@ class FreqtradeBot:
         # Protect sell-logic from forcesell and viceversa
         self._sell_lock = Lock()
 
+    def notify_status(self, msg: str) -> None:
+        """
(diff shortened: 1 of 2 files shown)


freqtrade/freqtradebot.py
```diff
@@ -95,6 +95,16 @@ def __init__(self, config: Dict[str, Any]) -> None:
         # Protect sell-logic from forcesell and viceversa
         self._sell_lock = Lock()
 
+    def notify_status(self, msg: str) -> None:
```
xmatthias: I'm not 100% convinced about moving this to freqtradebot - why not move this as "helper" to `RPCManager`.

On the other hand, it's aligend to where notifications for buy / sells are generated  - which we cannot really move to RPCManager as they depend on methods (e.g. `get_sell_rate()` available from freqtrade.

PR #2815, opened 2020-01-27
Minor: Advise to use https method for git clone instead of ssh

hroff-1902:
...because ssh can lead to permission denied error when no public key is installed, which can be frustrating for novices installing the bot following the docs.


xmatthias:
nice idea :+1:

hroff-1902:
(from Slack)

PR #2825, opened 2020-01-31
more consistent backtesting and sell reasons tables

yazeed:
Backtesting report and sell reasons tables are now more consistent with each other, and sell reasons table gives better info to help you differentiate between your backtest results with a quick glance by providing more info there as well.

Screenshot 2020-01-31 04 41 56
Screenshot 2020-01-31 04 41 50

### DIFF
--- a/freqtrade/optimize/optimize_reports.py
+++ b/freqtrade/optimize/optimize_reports.py
@@ -93,7 +93,7 @@ def generate_text_table_sell_reason(
         profit_mean = round(result['profit_percent'].mean() * 100.0, 2)
         profit_sum = round(result["profit_percent"].sum() * 100.0, 2)
         profit_tot = result["profit_abs"].sum()
-        profit_percent_tot = result["profit_percent"].sum() * 100.0 / max_open_trades
+        profit_percent_tot = round(result["profit_percent"].sum() * 100.0 / max_open_trades, 2)
         tabular_data.append(
             [
                 reason.value,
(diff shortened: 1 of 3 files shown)


yazeed:
Just to elaborate on the changes made here, I changed the wording "Profit" to "Wins" which is more consistent with trading terminology, and "Loss" to "Losses", across both the backtesting table and the sell reasons table. Then, I populated the sell reasons table with "Cum Profit %, Tot Profit STAKE, and Tot Profit %" which provide more complete data to help you analyze the strategy's decision making process.

freqtrade/optimize/optimize_reports.py
```diff
@@ -58,21 +66,46 @@ def generate_text_table(data: Dict[str, Dict], stake_currency: str, max_open_tra
...
+    ]
     for reason, count in results['sell_reason'].value_counts().iteritems():
         result = results.loc[results['sell_reason'] == reason]
         profit = len(result[result['profit_abs'] >= 0])
         loss = len(result[result['profit_abs'] < 0])
         profit_mean = round(result['profit_percent'].mean() * 100.0, 2)
-        tabular_data.append([reason.value, count, profit, loss, profit_mean])
+        profit_sum = round(result["profit_percent"].sum() * 100.0, 2)
+        profit_tot = result["profit_abs"].sum()
+        profit_percent_tot = round(result["profit_percent"].sum() * 100.0 / max_open_trades, 2)
```
xmatthias: I made a very small change to this section to align quotes to be identical with the lines prior to this.

While i don't really care which quotes we use - i think within "the same area" we should try to be consistent when doing the same thing.

Since it's VERY nitpicky, i made the change myself - hope you don't mind.

yazeed: Thanks for making the change, I'll keep it in mind to stick with single quotes in the future.

yazeed:
> LGTM 👍
> that's a great improvement, thanks!
> 
> And thanks for noticing (and aligning) wording. I know we're not consistent in wording - not with this,and not in several other cases ... but we're trying to move to consistency.

Thanks to you, many improvements shall follow. I'll do my best to cover areas within my capability.

PR #2826, opened 2020-01-31
added daily sharpe ratio hyperopt loss method, ty @djacky

### LINKED ISSUE #2727: Sharpe Ratio formula incorrect (Hyperopt loss function)
**edit @xmatthias **: Adding links to the most relevant comments - since this first post is empty anyway

Proposed solution calculation:
https://github.com/freqtrade/freqtrade/issues/2727#issuecomment-569979202
Addressing concerns about using daily returns: 
https://github.com/freqtrade/freqtrade/issues/2727#issuecomment-569988986
More concerns about using anything below <hourly returns:
https://github.com/freqtrade/freqtrade/issues/2727#issuecomment-570033907

yazeed:
Added the sharpe ratio hyperopt loss method which calculates sharpe ratio on a daily basis instead of yearly, as suggested by @djacky [here](https://github.com/freqtrade/freqtrade/issues/2727). On that thread, it was discussed how it can also be modified to work for smaller timeframes by users. 

I believe this hyper optimization loss method may help in eliminating bias due to having certain trades taking place in historical trends, when backtesting for the goal of having a strategy with consistent daily risk.

closes #2727 

~~Note: This requires having [freqtrade/technical](https://github.com/freqtrade/technical) installed in your freqtrade.~~

### DIFF
--- a/docs/bot-usage.md
+++ b/docs/bot-usage.md
@@ -3,7 +3,7 @@
 This page explains the different parameters of the bot and how to run it.
 
 !!! Note
-    If you've used `setup.sh`, don't forget to activate your virtual environment (`source .env/bin/activate`) before running freqtrade commands.
+If you've used `setup.sh`, don't forget to activate your virtual environment (`source .env/bin/activate`) before running freqtrade commands.
 
 ## Bot commands
 
@@ -337,8 +337,8 @@ optional arguments:
                         generate completely different results, since the
                         target for optimization is different. Built-in
                         Hyperopt-loss-functions are: DefaultHyperOptLoss,
-                        OnlyProfitHyperOptLoss, SharpeHyperOptLoss (default:
-                        `DefaultHyperOptLoss`).
+                        OnlyProfitHyperOptLoss, SharpeHyperOptLoss,
+                        SharpeHyperOptLossDaily (default: `DefaultHyperOptLoss`).
 
 Common arguments:
   -v, --verbose         Verbose mode (-vv for more, -vvv to get all messages).
--- a/docs/bot-usage.md
+++ b/docs/bot-usage.md
@@ -3,7 +3,7 @@
 This page explains the different parameters of the bot and how to run it.
 
 !!! Note
-If you've used `setup.sh`, don't forget to activate your virtual environment (`source .env/bin/activate`) before running freqtrade commands.
+    If you've used `setup.sh`, don't forget to activate your virtual environment (`source .env/bin/activate`) before running freqtrade commands.
 
 ## Bot commands
 
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -1,0 +1,69 @@
+"""
+SharpeHyperOptLoss
+
+This module defines the alternative HyperOptLoss class which can be used for
+Hyperoptimization.
+"""
+from datetime import datetime
+
+from pandas import DataFrame
+import numpy as np
+
+from freqtrade.optimize.hyperopt import IHyperOptLoss
+
+
+class SharpeHyperOptLossDaily(IHyperOptLoss):
+    """
+    Defines the loss function for hyperopt.
+
+    This implementation uses the Sharpe Ratio calculation.
+    """
+
+    @staticmethod
+    def hyperopt_loss_function(
+        results: DataFrame,
+        trade_count: int,
+        min_date: datetime,
+        max_date: datetime,
+        *args,
+        **kwargs
+    ) -> float:
+        """
+        Objective function, returns smaller number for more optimal results.
+
+        Uses Sharpe Ratio calculation.
+        """
+        total_profit = results.profit_percent
+        # days_period = (max_date - min_date).days
+
+        # adding slippage of 0.1% per trade
+        total_profit = total_profit - 0.0005
+        # expected_yearly_return = total_profit.sum() / days_period
+
+        # if np.std(total_profit) != 0.0:
+        #     sharp_ratio = expected_yearly_return / np.std(total_profit) * np.sqrt(365)
+        # else:
+        #     # Define high (negative) sharpe ratio to be clear that this is NOT optimal.
+        #     sharp_ratio = -20.0
+
+        # # print(expected_yearly_return, np.std(total_profit), sharp_ratio)
+        # return -sharp_ratio
+
+        sum_daily = (
+            results.resample("D", on="close_time").agg(
+                {"profit_percent": sum, "profit_abs": sum}
+            )
+            * 100.0
+        )
+
+        if np.std(total_profit) != 0.0:
+            sharp_ratio = (
+                sum_daily["profit_percent"].mean()
+                / sum_daily["profit_percent"].std()
+                * np.sqrt(365)
+            )
+        else:
+            # Define high (negative) sharpe ratio to be clear that this is NOT optimal.
+            sharp_ratio = -20.0
+
+        return -sharp_ratio
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -43,7 +43,7 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
             * 100.0
         )
 
-        if (np.std(sum_daily.profit_percent) != 0.):
+        if (np.std(dataframe['profit_percent']) != 0.):
             sharp_ratio = (
                 sum_daily["profit_percent"].mean()
                 / np.std(sum_daily["profit_percent"])
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -29,16 +29,16 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
         Uses Sharpe Ratio calculation.
         """
         # get profit_percent and apply slippage of 0.1% per trade
-        results.loc[:, 'profit_percent'] = results['profit_percent'] - 0.0005
+        results.loc[:, 'profit_percent_after_slippage'] = results['profit_percent'] - 0.0005
 
         sum_daily = (
             results.resample("D", on="close_time").agg(
-                {"profit_percent": sum}
+                {"profit_percent_after_slippage": sum}
             )
             * 100.0
         )
 
-        total_profit = sum_daily["profit_percent"]
+        total_profit = sum_daily["profit_percent_after_slippage"]
         expected_returns_mean = total_profit.mean()
         up_stdev = np.std(total_profit)
 
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -33,10 +33,8 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
 
         Uses Sharpe Ratio calculation.
         """
-        total_profit = results.profit_percent
-
-        # adding slippage of 0.1% per trade
-        total_profit = total_profit - 0.0005
+        # get profit_percent and apply slippage of 0.1% per trade
+        results.loc[:, 'profit_percent'] = results['profit_percent'] - 0.0005
 
         sum_daily = (
             results.resample("D", on="close_time").agg(
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -4,10 +4,10 @@ SharpeHyperOptLossDaily
 This module defines the alternative HyperOptLoss class which can be used for
 Hyperoptimization.
 """
+import math
 from datetime import datetime
 
 from pandas import DataFrame, date_range
-import numpy as np
 
 from freqtrade.optimize.hyperopt import IHyperOptLoss
 
@@ -35,7 +35,8 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
         risk_free_rate = annual_risk_free_rate / days_in_year
 
         # apply slippage per trade to profit_percent
-        results.loc[:, 'profit_percent_after_slippage'] = results['profit_percent'] - slippage_per_trade_ratio
+        results.loc[:, 'profit_percent_after_slippage'] = \
+            results['profit_percent'] - slippage_per_trade_ratio
 
         # create the index within the min_date and end max_date
         t_index = date_range(start=min_date, end=max_date, freq=resample_freq)
@@ -50,11 +51,11 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
         up_stdev = total_profit.std()
 
         if (up_stdev != 0.):
-            sharp_ratio = expected_returns_mean / up_stdev * np.sqrt(days_in_year)
+            sharp_ratio = expected_returns_mean / up_stdev * math.sqrt(days_in_year)
         else:
             # Define high (negative) sharpe ratio to be clear that this is NOT optimal.
             sharp_ratio = -20.
 
-        #print(t_index, sum_daily, total_profit)
-        #print(risk_free_rate, expected_returns_mean, up_stdev, sharp_ratio)
+        # print(t_index, sum_daily, total_profit)
+        # print(risk_free_rate, expected_returns_mean, up_stdev, sharp_ratio)
         return -sharp_ratio
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -40,7 +40,7 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
 
         sum_daily = (
             results.resample("D", on="close_time").agg(
-                {"profit_percent": sum, "profit_abs": sum}
+                {"profit_percent": sum}
             )
             * 100.0
         )
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -31,7 +31,7 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
         resample_freq = '1D'
         slippage_per_trade_ratio = 0.0005
         days_in_year = 365
-        annual_risk_free_rate = 0.03
+        annual_risk_free_rate = 0.0
         risk_free_rate = annual_risk_free_rate / days_in_year
 
         # apply slippage per trade to profit_percent
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -40,11 +40,13 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
 
         total_profit = sum_daily["profit_percent"]
         expected_returns_mean = total_profit.mean()
+        up_stdev = np.std(total_profit)
 
         if (np.std(total_profit) != 0.):
-            sharp_ratio = expected_returns_mean / np.std(total_profit) * np.sqrt(365)
+            sharp_ratio = expected_returns_mean / up_stdev * np.sqrt(365)
         else:
             # Define high (negative) sharpe ratio to be clear that this is NOT optimal.
             sharp_ratio = -20.
 
+        # print(expected_returns_mean, up_stdev, sharp_ratio)
         return -sharp_ratio
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -43,14 +43,14 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
             * 100.0
         )
 
-        if (np.std(dataframe['profit_percent']) != 0.):
-            sharp_ratio = (
+        if (np.std(sum_daily["profit_percent"] != 0.):
+            sharp_ratio=(
                 sum_daily["profit_percent"].mean()
                 / np.std(sum_daily["profit_percent"])
                 * np.sqrt(365)
             )
         else:
             # Define high (negative) sharpe ratio to be clear that this is NOT optimal.
-            sharp_ratio = -20.0
+            sharp_ratio=-20.0
 
         return -sharp_ratio
--- a/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
+++ b/freqtrade/optimize/hyperopt_loss_sharpe_daily.py
@@ -34,20 +34,9 @@ class SharpeHyperOptLossDaily(IHyperOptLoss):
         Uses Sharpe Ratio calculation.
         """
         total_profit = results.profit_percent
-        # days_period = (max_date - min_date).days
 
         # adding slippage of 0.1% per trade
         total_profit = total_profit - 0.0005
-        # expected_yearly_return = total_profit.sum() / days_period
-
-        # if np.std(total_profit) != 0.0:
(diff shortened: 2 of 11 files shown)


freqtrade/optimize/hyperopt_loss_sharpe_daily.py
```diff
@@ -0,0 +1,58 @@
...
+    ) -> float:
+        """
+        Objective function, returns smaller number for more optimal results.
+
+        Uses Sharpe Ratio calculation.
+        """
+        total_profit = results.profit_percent
+
+        # adding slippage of 0.1% per trade
+        total_profit = total_profit - 0.0005
```
xmatthias: This does not seem to be correct.
While it's aligned to `SharpeHyperOptLoss` - `total_profit` is not used after this point - which makes this line pointless.

There's 2 ways to correct this i think.
Either, using `total_profit` throughout instead of `results['profit_percent']`
or - modify the column of the results dictionary as follows.

``` python
results.loc[:, 'profit_percent'] = results['profit_percent'] - 0.0005
```

freqtrade/optimize/hyperopt_loss_sharpe_daily.py
```diff
@@ -0,0 +1,58 @@
...
+
+        sum_daily = (
+            results.resample("D", on="close_time").agg(
+                {"profit_percent": sum, "profit_abs": sum}
```
xmatthias: since we're talking hyperopt - i think `"profit_abs"` should be removed from this dictionary.

It's not used in the remainder of this function - and in hyperopt, every ms you save gets multiplied by epochs ...

yazeed: Good point, let me fix that.

freqtrade/optimize/hyperopt_loss_sharpe_daily.py
yazeed: So replacing this line `total_profit = total_profit - 0.0005` with `results.loc[:, 'profit_percent'] = results['profit_percent'] - 0.0005` would work?

yazeed:
> Thanks a lot for this 👍
> 
> > Note: This requires having freqtrade/technical installed in your freqtrade.
> 
> i don't think this note is correct - at least the code does not import technical at all.
> Not really an issue - but had me confused for a moment ...
> 
> I would want a simple test for this (just to ensure the logic is somewhat correct) - probably best look at `test_sharpe_loss_prefers_higher_profits()` in `test_hyperopt.py` - i suspect that the same test will work for this loss-function too, as the test only verifies that better results return lower values from this function.

Oh, I thought I read somewhere in the discussions that resample requires it. Anyways, then just disregard that note.

yazeed:
> Thanks a lot for this 👍
> 
> > Note: This requires having freqtrade/technical installed in your freqtrade.
> 
> i don't think this note is correct - at least the code does not import technical at all.
> Not really an issue - but had me confused for a moment ...
> 
> I would want a simple test for this (just to ensure the logic is somewhat correct) - probably best look at `test_sharpe_loss_prefers_higher_profits()` in `test_hyperopt.py` - i suspect that the same test will work for this loss-function too, as the test only verifies that better results return lower values from this function.

I'll take a look at this and submit it's own test along the same guidelines. Thanks for your review.

freqtrade/optimize/hyperopt_loss_sharpe_daily.py
yazeed: @xmatthias I'm now thinking it should be like this:
`results.loc[:, 'profit_percent'] = results.loc[:, 'profit_percent'] - 0.0005`

xmatthias: `results.loc[:, 'profit_percent'] = results.loc[:, 'profit_percent'] - 0.0005` is technically identical to the above `total_profit = total_profit - 0.0005` line.
this selects the column profit_percent- and reassigns it to itself with 0.0005 subtracted. 

the `total_profit` version selects the column `profit_percent` in line 36 - and then does the same thing.

This works in the "old" hyperopt_loss_sharpe method, as `results` is not reused after that.
I like the resample on top of `results` better as it's clearer what the type of this is - so i think you could remove line 36, change this to `results.loc[:, 'profit_percent'] = results['profit_percent'] - 0.0005` - and it'll have the same effect.

yazeed: Got it, thanks for the explanation, I'm actually running a course right now on datacamp that teaches how to play with pandas dataframe.

yazeed:
Updated with proper slippage.

yazeed:
Going to work on the test case now, and fix it for flake8 compliance. :)

yazeed:
@xmatthias Not sure why a random test failed on this one.

yazeed:
I can proceed to cover mention of this in MD docs to be consistent with other hyperopt loss methods mentioned there.

yazeed:
This is all ready, for final review, now, passed that random failed test this time with only a docs change.

docs/bot-usage.md
```diff
@@ -3,7 +3,7 @@
 This page explains the different parameters of the bot and how to run it.
 
 !!! Note
-    If you've used `setup.sh`, don't forget to activate your virtual environment (`source .env/bin/activate`) before running freqtrade commands.
+If you've used `setup.sh`, don't forget to activate your virtual environment (`source .env/bin/activate`) before running freqtrade commands.
```
xmatthias: Please don't do this (it's not just in this file - but also in the other doc-files changed here).

after every line with `!!! Note ` or `!!! Bug` (or a few others) - the next line MUST be indented by 4 spaces.
It's required to have the boxes in the [documentation](https://www.freqtrade.io/en/latest/bot-usage/) render correctly.
It's alsy very picky to not have `!!! Note:` (no trailing colon)... but i think it looks good in the docs so it's worth the effort.

for further reference about this - see the [developer docs](https://www.freqtrade.io/en/latest/developer/#documentation) - which however only link to the original docs of the plugin we use for this [here](https://squidfunk.github.io/mkdocs-material/extensions/admonition/).

freqtrade/optimize/hyperopt_loss_sharpe_daily.py
```diff
@@ -0,0 +1,56 @@
...
+                {"profit_percent": sum}
+            )
+            * 100.0
+        )
+
+        if (np.std(sum_daily.profit_percent) != 0.):
+            sharp_ratio = (
+                sum_daily["profit_percent"].mean()
+                / sum_daily["profit_percent"].std()
+                * np.sqrt(365)
```
xmatthias: i would decide on either `np.std()` - or `dataframe[].std()`. 
Using both seems within the same 5 lines is very odd ... 

Also best use either the pandas . notation (`dataframe.profit_percent`, or bracket notation `dataframe['profit_percent']`. 
Personally, i favor bracket notation, as that won't break if the column is accidentally called like a pandas function.

docs/bot-usage.md
yazeed: Oh, I did not notice it auto formatted those. My bad, just fixed it, and fixed my editor settings for this repo.

freqtrade/optimize/hyperopt_loss_sharpe_daily.py
yazeed: Done, thanks.

yazeed:
Failed on that random test again, but it should be ready now. Thanks for all your feedback.

yazeed:
Don't merge yet, I think I figured something out. Both of our sharpe ratio calculators have a bug in them.

They both should have np.sqrt(days_period) instead of np.sqrt(365).

yazeed:
Nevermind my previous post, so it seems that **np.sqrt(365)** refers to **Annualized Sharpe Ratio** which I guess is what we want to be using in hyperopt as a baseline. Just made my final push on this one, if all looks good, it's ready for review.

By the way, I'm thinking of submitting a minor enhancement to the regular sharpe ratio method to bring more consistency to them.

hroff-1902:
This implementation changes the original backtesting_results dataframe, the changes are not visible just because they are rounded in the output. Moreover, the modified value of the `profit_percent` column is later used in `_calculate_results_metrics()` as the sum, so in case of thousands of trades the output will contain incorrect modified profit...

A copy of the dataframe (or just of this column) should be used for sharpe calculations.

yazeed:
> This implementation changes the original backtesting_results dataframe, the changes are not visible just because they are rounded in the output. Moreover, the modified value of the `profit_percent` column is later used in `_calculate_results_metrics()` as the sum, so in case of thousands of trades the output will contain incorrect modified profit...
> 
> A copy of the dataframe (or just of this column) should be used for sharpe calculations.

Oh ok, let me fix that, I thought it would apply.

hroff-1902:
^^ This check (unmodified results) should be added to the tests for all loss functions (not in this PR)

yazeed:
@hroff-1902 Does this last commit solve the issue for this hyperloss method?

hroff-1902:
> @hroff-1902 Does this last commit solve the issue for this hyperloss method?

I personally would not modify the backtesting_results at all (so that it can be easily compared with the original one in the tests) and use a copy of the column in a local var, for example (or a copy of the whole dataframe), but it's okay for now, the rest can later be polished when the tests I mentioned above will be added...

hroff-1902:
Next point (to discuss): 

since original Sharpe is broken in its calculations, I would really make this one named SharpeHyperOptLoss and the current one as LegacySharpeHyperOptLoss (and marked deprecated for usage till it removed in the future completely). I would not probably remove the current one for some time since there can be people relying on its values... Or maybe this is too redundant (that one is really broken) and this implementation should simply substitute the legacy one...

But there is another point, do not rename it right now... 😉

yazeed:
> Next point (to discuss):
> 
> since original Sharpe is broken in its calculations, I would really make this one named SharpeHyperOptLoss and the current one as LegacySharpeHyperOptLoss (and marked deprecated for usage till it's removed in the future completely). I would not probably remove the current one for some time since there can be people relying on its values... Or maybe this is too redundant (that one is really broken) and this implementation should simply substitute the legacy one...
> 
> But there is another point, do not rename it right now... 😉

I think we might want to keep both for further testing and feedback by users before deprecating the original sharpe ratio method.

hroff-1902:
> I think we might want to keep both for further testing and feedback by users before deprecating the original sharpe ratio method.

Deprecation does not mean the user will not be able to use the deprecated feature (loss function) -- there only will be a deprecation message printed in the logs warning the users that this feature can be removed in a few months or so...

yazeed:
> > I think we might want to keep both for further testing and feedback by users before deprecating the original sharpe ratio method.
> 
> Deprecation does not mean the user will not be able to use the deprecated feature (loss function) -- there only will be a deprecation message printed in the logs warning the users that this feature can be removed in a few months or so...

Sounds good to me. Should I make another commit to have the tests rerun, or is this good to go?

hroff-1902:
Well, the problem with the daily sharpe is the following: it's okay to use it on higher timeframes (1h, 1d, 5m with multi-day timerange), but it will fail if we backtest on 1m for 60*24*2~=3000 candles (2 days) or on 5m for 600 candles (also ~2 days). That's the reason why I personally was stuck with similar PR for long time -- passing a parameter to loss function from cli or from config to control the resampling interval seems to be overcomplicated...

In the meantime, however, I found the solution -- we could implement (later, not in this PR) the SharpeHyperOptLossHourly, which will aggregate the trades on the hourly basis. We'll then add an advise into the docs to use that loss function for lower timeframes and shorter timeranges while this one will be suggested for use with higher timeframes...

What do you think?

Thus, the current Sharpe can be deprecated and/or renamed later, it won't take the name of one of these two new loss functions...

xmatthias:
That - sounds like a plant to me :+1:

hroff-1902:
There is also another point which should be discussed here, otherwise this function can happen to be broken again:

Currently the series (column) with aggregated profits contains only values for days with trades. Another approach for calculation would be on a series with zero profits added for all those days from the timerange that are absent. This seems more correct for me. However, this will change the mean value (especially for results with small amount of trades) and, thus, the stddev as well, in non-linear way. The extra days with zero profits (or -slippage?), if this is accepted, should only be used internally for calculations and should not be inserted into the backtesting_results dataframe...

This also should be discussed and clarified.

(@yazeed sorry, that's not about your faults or errorneous implementation, pls don't mind -- there are just some amount of theoretical issues that should be at least discussed, otherwise this new loss function can also be 'broken')

xmatthias:
> Currently the series (column) with aggregated profits contains only values for days with trades. 

I'm not sure what leads you to belive this - it's not true.

I assume this is a misconception on how `pd.resample()` works ... 

`.resample('D')` will create one line per day from the first date to the last date (close_time, in this case).
If no trade happened on a particular day, that will be a line with a sum of 0.
if 20 trades happened, there will be 1 line, with the sum of these profits (which accidentally can also be 0 - if 2 trade profits cancel themselfs).

`total_profit.mean()` - the only `mean()` used in this function does calculate the mean per day - so if no trades happened, this 0 day will be accounted for.

If you look at freqtrade code - we rely on the above explained resample functioning  in `converter.ohlcv_fill_up_missing_data()` to fill in missing candles.

hroff-1902:
* @xmatthias all days should be filled: for instance, if min_date is 01.01,2020 max_date is 10.01.2020 and backtesting_results contain two rows for trades with date 04.01.2020, one with profit_percent=0.04 and second with profit_percent=-0.01, then the dataframe, series or list used to calculate mean and stddev should consist of 10 rows for each day from min_date to max_date, all except need to be filled with 0 profit_percent (actually, -0.0005 after substraction of slippage) and row for 04.01 should have profit_percent=0.04-0.01-0.0005.

Resample is not passed min_date and max_date, so it only fills days *between* trades.

If you add the 'tails' (for leading and trailing days) with zero values, the mean will change (and thus the stddev as well).

* In my opinion, slippage should be subtracted after resampling on the daily basis. Because it's not really slippage, but what's called Free Risk Rate in the definition of the Sharpe Ratio. Or, even probably better, the implementation can subtract two different values -- one as slippage (to handle possible slippage that will happen in real conditions, which we define as 0.0005) for every trade as it is done now and another value as Risk Free Rate on the daily basis, after resampling...


Btw, here is the original article on the Sharpe Ratio by W.F. Sharpe: https://web.stanford.edu/~wfsharpe/art/sr/sr.htm (outdated, he changed some positions in his later publications, as Wikipedia says to us, but this can be interesting for understanding the rationale behind this metrics)

yazeed:
>     * @xmatthias all days should be filled: for instance, if min_date is 01.01,2020 max_date is 10.01.2020 and backtesting_results contain two rows for trades with date 04.01.2020, one with profit_percent=0.04 and second with profit_percent=-0.01, then the dataframe, series or list used to calculate mean and stddev should consist of 10 rows for each day from min_date to max_date, all except need to be filled with 0 profit_percent (actually, -0.0005 after substraction of slippage) and row for 04.01 should have profit_percent=0.04-0.01-0.0005.
> 
> 
> Resample is not passed min_date and max_date, so it only fills days _between_ trades.
> 
> If you add the 'tails' (for leading and trailing days) with zero values, the mean will change (and thus the stddev as well).
> 
>     * In my opinion, slippage should be subtracted after resampling on the daily basis. Because it's not really slippage, but what's called Free Risk Rate in the definition of the Sharpe Ratio. Or, even probably better, the implementation can subtract two different values -- one as slippage (to handle possible slippage that will happen in real conditions, which we define as 0.0005) for every trade as it is done now and another value as Risk Free Rate on the daily basis, after resampling...
> 
> 
> Btw, here is the original article on the Sharpe Ratio by W.F. Sharpe: https://web.stanford.edu/~wfsharpe/art/sr/sr.htm (outdated, he changed some positions in his later publications, as Wikipedia says to us, but this can be interesting for understanding the rationale behind this metrics)

So the quickest modification can be as follows, to count the # of trades on each day after resampling, and then apply slippage on each day's profit_percent_after_slippage per each trade counted.

Let me know if you find consensus on this.

hroff-1902:
> So the quickest modification can be as follows, to count the # of trades on each day after resampling, and then apply slippage on each day's profit_percent_after_slippage per each trade counted.
> 
> Let me know if you find consensus on this.

No, in my understanding, mean (and stddev) are calculated on daily basis, so leading and trailing zero days should be added somehow (or used in the custom calculation of mean instead of `.mean()` and `.stddev` methods).

Example: min_date is 01.01, max_date is 10.01, sum_daily after resampling contains trades for 04.01, 06.01, and 05.01 was filled with zero profit by resampling. Mean will now be calculated as sum_daily.sum() / 3 days (04.01-06.01), while it should be sum_daily.sum() / 10 days (01.01-10.01). Similar for stddev.

yazeed:
> s should be added someho

Oh ok, I missed that part of your comment earlier, so how about this change then?

from:
```
         expected_returns_mean = total_profit.mean() 
```
to
```
        expected_returns_mean = total_profit.sum() / days_period
```

hroff-1902:
> > No, in my understanding, mean (and stddev) are calculated on daily basis, so leading and trailing zero days should be added somehow (or used in the custom calculation of mean instead of .mean() and .stddev methods).
> 
> Oh ok, I missed that part of your comment earlier, so how about this change then?
> 
> from:
> 
> ```
>          expected_returns_mean = total_profit.mean() 
> ```
> 
> to
> 
> ```
>         days_period = (max_date - min_date).days
>         expected_returns_mean = total_profit.sum() / days_period
> ```

^^ This can be a solution for the mean value, but stddev also contains mean value in its calculations, so since we use modified value of mean in the case, the value of stddev should also be recalculated. It's not linear so custom calculations will not be so easy for stddev...

And that's without subtraction of the Risk Free Value on the daily basis...

So a better solution can be extending the dataframe with zero values for min_date and max_date before resampling so that resampling could populate all the days in between. Or extending the dataframe to the whole timerange by means of appropriate pandas function (if one suitable for this exists)...

Let's wait for reaction on all this from @xmatthias ...

xmatthias:
> Let's wait for reaction on all this from @xmatthias ...

why?
I'm no python wonder healer ... :roll_eyes: 
nor am i a sharpe ratio expert...

----

If we're talking about resampling considering that the bot only opened the first trade N days into the hyperopt cycle (or stopped M days short) - then the solution is simple:
Reindex the timeseries with a DateTimeIndex having the full range of dates.

Below is a (fixed) sample from [stackoverflow](https://stackoverflow.com/questions/45908146/pandas-changing-the-start-and-end-date-of-resampled-timeseries).
While the original code won't work anymore due to changes in pandas, the below will, and it will give you the full range (`t_index`) of dates.

``` python
import pandas as pd

# create the index with the start and end times you want
t_index = pd.date_range(start='2009-06-01', end='2009-06-30', freq='1d')

# create the data frame
df = pd.DataFrame([['2009-06-07 02:07:42'],
                   ['2009-06-11 17:25:28'],
                   ['2009-06-11 17:50:42'],
                   ['2009-06-11 17:59:18']], columns=['daytime'])

df['daytime'] = pd.to_datetime(df['daytime'])

# resample the data to 1 hour, aggregate by counts,
# then reset the index and fill the na's with 0
df2 = df.resample('1d', on='daytime').count().reindex(t_index).fillna(0)
```

since `min_date` and `max_date` are passed to the hyperoptloss function anyway, this should be easy to incorporate.

--- 

While the above will add the days before / after the trading period - We should consider how the min / max date responds to missing start/end date for some pairs (details will be posted in #2854).

yazeed:
Would this work? Not sure if results['daytime'] is needed since we already have close_time, and I wonder if we can work with that directly? Also not sure I understand the function of .count() here. Sorry, I'm a code hacker, not a Python dev originally.

```
        days_period = (max_date - min_date).days
        # create the index with the start and end times you want
        t_index = results.date_range(start=min_date, end=max_date, freq="1D")

        results['daytime'] = results.to_datetime(results['close_time'])

        sum_daily = (
            results.resample("D", on="close_time").count().reindex(t_index).fillna(0).agg(
                {"profit_percent_after_slippage": sum}
            )
            * 100.0
        )

        total_profit = sum_daily["profit_percent_after_slippage"]
        expected_returns_mean = total_profit.sum() / days_period
        up_stdev = np.std(total_profit)
```

xmatthias:
Honestly, it's a sample / template only.
`.count()` is the equivalent to `agg()` used in the sample. 
and no, daytime is NOT needed.



```python
        days_period = (max_date - min_date).days
        # create the index with the start and end times you want
        t_index = results.date_range(start=min_date, end=max_date, freq="1D")

        sum_daily = (
            results.resample("D", on="close_time").agg(
                {"profit_percent_after_slippage": sum}
            )
            * 100.0
        ).reindex(t_index).fillna(0)

        total_profit = sum_daily["profit_percent_after_slippage"]
        expected_returns_mean = total_profit.sum() / days_period
        up_stdev = np.std(total_profit)
```

Now i'm not sure if the reindex needs to go before / after the *100 (i suspect both should work) - but i lack time this week to test this completely (that's why i don't commit to your branch ...).

hroff-1902:
working on it...

yazeed:
> Honestly, it's a sample / template only.
> `.count()` is the equivalent to `agg()` used in the sample.
> and no, daytime is NOT needed.
> 
> ```python
>         days_period = (max_date - min_date).days
>         # create the index with the start and end times you want
>         t_index = results.date_range(start=min_date, end=max_date, freq="1D")
> 
>         sum_daily = (
>             results.resample("D", on="close_time").agg(
>                 {"profit_percent_after_slippage": sum}
>             )
>             * 100.0
>         ).reindex(t_index).fillna(0)
> 
>         total_profit = sum_daily["profit_percent_after_slippage"]
>         expected_returns_mean = total_profit.sum() / days_period
>         up_stdev = np.std(total_profit)
> ```
> 
> Now i'm not sure if the reindex needs to go before / after the *100 (i suspect both should work) - but i lack time this week to test this completely (that's why i don't commit to your branch ...).

Shouldn't the reindex and fillna(0) be placed at the end of the resample object?

yazeed:
> working on it...
> (there is an issue with backtesting_results passed to the func)

@hroff-1902 Awesome. Thanks for stepping in.

hroff-1902:
@yazeed @xmatthias 
pls have a look at my version (committed here)

The hourly version can be simply created by redefining constants. (not in this PR)

freqtrade/optimize/hyperopt_loss_sharpe_daily.py
```diff
@@ -28,25 +28,33 @@ def hyperopt_loss_function(results: DataFrame, trade_count: int,
 
         Uses Sharpe Ratio calculation.
         """
-        # get profit_percent and apply slippage of 0.1% per trade
-        results.loc[:, 'profit_percent_after_slippage'] = results['profit_percent'] - 0.0005
+        resample_freq = '1D'
+        slippage_per_trade_ratio = 0.0005
+        days_in_year = 365
+        annual_risk_free_rate = 0.03
```
xmatthias: This seems like an odd number to pick.
if you assume risk-free as having money in your bank account - then (at least in Europe) - that number would be more like 0.00000000001) ... 🙄

yazeed: I'd just use 0 for risk free rate, it doesn't have an impact in hyperopt context, since we only use sharpe ratio as a baseline measure, rather than for our interest in its final absolute value.

@hroff-1902 Also, you're deducting risk_free_rate from total_profit, while it should be deducted from expected_returns_mean.

Finally, is there any reason why you switched np.std(total_profit) back to total_profit.std() ? I thought we preferred the former.

hroff-1902: actually, according to the definition of the classic Sharpe Ratio, risk-free rate depends on the market (stock, ETF, etc) and crypto are more volatile... Seen this rate used somewhere as an example, do not remember exactly...

hroff-1902:
> I'd just use 0 for risk free rate, it doesn't have an impact in hyperopt context, since we only use sharpe ratio as a baseline measure, rather than for our interest in its final absolute value.
> 
> @hroff-1902 Also, you're deducting risk_free_rate from total_profit, while it should be deducted from expected_returns_mean.

no, risk free rate is subtracted on the daily basis, for each daily income, before taking mean and stdev
because it's daily value

PR #2831, opened 2020-02-01
introduce new-config subcommand

xmatthias:
## Summary
This PR will introduce a new subcommand - `new-config`.
This method will replace the setup.sh script when creating a configuration, as using `sed` to modify a json file is risky at best.
Also, that script won't be available when installed from pypi (as dreamed about in #2112).

## Quick changelog

- add questionary as dependency to create user prompts
- introduce `freqtrade new-config` subcommand
- have setup.sh point to this command (since `setup.sh` will not be available in the future, i think it's best to educate people towards using the appropriate freqtrade commands).

## Open questions

- [ ] we could/should remove all exchange-specific example-configs (`config_binance.json.example` and `config_kraken.json.example`), since these can be generated using new-config. I did not do this yet as i'd like to discuss this first.


docs/utils.md
```diff
@@ -40,7 +40,6 @@ optional arguments:
 
 Creates a new configuration file, asking some questions which are important selections for a configuration.
 
-
```
hroff-1902: 👍

setup.sh
```diff
@@ -215,27 +223,8 @@ function config_generator() {
 function config() {
 
     echo "-------------------------"
-    echo "Generating config file"
+    echo "Please use freqtrade new-config -c config.json to generate a new configuration file."
```
hroff-1902: `'Please use "freqtrade new-config -c config.json" to generate a new configuration file.'`
-- to have the command in double quotes, like `git`, for instance, prints its suggestions for the commands to be run.
(?)

freqtrade/commands/build_config_commands.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+
+    config_path = Path(args['config'][0])
+    if config_path.exists():
+        overwrite = ask_user_overwrite(config_path)
+        if overwrite:
+            config_path.unlink()
+        else:
+            raise OperationalException(
+                f"Configuration `{config_path}` already exists. "
```
hroff-1902:
```suggestion
                f"Configuration file `{config_path}` already exists. "
```

freqtrade/commands/build_config_commands.py
```diff
@@ -0,0 +1,193 @@
...
+                "Please use another configuration name or delete the existing configuration.")
```
hroff-1902:
```suggestion
                "Please delete it or use another configuration file name.")
```

freqtrade/templates/base_config.json.j2
```diff
@@ -0,0 +1,58 @@
...
+    "bid_strategy": {
+        "ask_last_balance": 0.0,
+        "use_order_book": false,
+        "order_book_top": 1,
+        "check_depth_of_market": {
+            "enabled": false,
+            "bids_to_ask_delta": 1
+        }
+    },
+    "ask_strategy":{
```
hroff-1902:
```suggestion
    "ask_strategy": {
```

freqtrade/commands/build_config_commands.py
```diff
@@ -0,0 +1,193 @@
...
+                Separator(),
+                "other",
+            ],
+        },
+        {
+            "type": "autocomplete",
+            "name": "exchange_name",
+            "message": "Type your exchange name (Must be supported by ccxt)",
+            "choices": available_exchanges(),
+            "when": lambda x: x["exchange_name"] == 'other'
```
hroff-1902: I would move selection of the exchange to be the first question: 1) that's the most important question in the set imho, 2) selection of stake_currency depend on it; 3) possibly in the future we'll ask different set of questions depending on exchange used.

hroff-1902:
wow, I found how to suggest changes in the review with the diff button! 🤣

setup.sh
xmatthias: I'll use single quotes for now - so it's aligned to the rest of this script...

xmatthias:
> wow, I found how to suggest changes in the review with the diff button! 

yes, i had to google it when i saw it too.

The problem i have with that is that it will force a commit per suggested change (you seem to be able to batch them - but that's grayed out on me).

freqtrade/commands/build_config_commands.py
xmatthias: To explain my thoughts on this:
* ask questions requiring keys / tokens only if required

Therefore, to me, dry-run needs to be first (or at least before exchange), as that determines if we need exchange key / token at all.
To however skip fro dry-run to exchange and then "back" to the simple questions (stake / stake amount / max_open_trades) seemed strange when trying it out.

Also (and i consider this a bug/ missing feature in questionaire) - boolean questions behave odd (to me) in providing an "auto-enter" when "Y" / "n" is pressed - so by asking a "simple" question right after that where the default can easily be changed afterwards - we can avoid to have to go through the flow multiple times.

hroff-1902:
> The problem i have with that is that it will force a commit per suggested change (you seem to be able to batch them - but that's grayed out on me).

Yeah, one commit per suggested change, that's strange implementation.
And it's very strange that this button appears for only one-line changes.
But I can imagine -- they tried to keep it simple...

hroff-1902:
The actual problem here I see is if you change stake_currency: the white_pairlist remains same (with pairs for default stake_currency) and you cannot run the bot with this config after creation...

I have not found a good solution for this...

Maybe:
* add the default stake_currency in a comment:
```
"stake_currency": "USDT" //"BTC"
```
or
```
//"stake_currency": "USDT"
"stake_currency": "USDT"
```
* put whitelist into comments as well
* add VolumePairList

xmatthias:
well there's a multitude of additional things that can be add to this.

In a first step, my goal is ot replace the shell script part for creating configuration (which i think this does in it's current form), as that brings the following benefits:
* works on windows
* allows selection of exchange
* correctly enables telegram (the old version did NOT do that).

Now obviously, additional things can be added (like dynamically generating the pairlist based on stake-currency), .... - but none of these things have been possible in the "prior" version ... 

I'm however hesitant to add pairlist generation, as i'm not sure if "current high-volume pairs" is good as a static pairlist. At times, these will include some shitcoins that bumped that day - but will only lead to losses in the end as the pair disappears into the nirvana a few days later.

> add the default stake_currency in a comment:

Now i have no idea what the benefit of this would be ... (and it doesn't become clear from the above comment).?

freqtrade/commands/build_config_commands.py
hroff-1902: Selecting exchange is also a "simple" question -- and the main one in my opinion. And answering it first is straightforward for a user -- first select exchange, then all supplicant questions (simple and more complex) follow...

> To however skip fro dry-run to exchange and then "back" to the simple questions (stake / stake amount / max_open_trades) seemed strange when trying it out.

right, but if you move selection of exchange at the top, you won't need to skip...
Select exchange, then the mode (dry run/live) *on that exchange*... 

A user may want to have live trade mode on an exchange she has an account and dry-run mode on other exchanges, so the mode *depends* on the exchange selected...

hroff-1902:
Summary: let's deal with the position of exchange selection and leave other enhancements (pairlilsts, etc) for the future, to keep this PR observable.

hroff-1902:
> we could/should remove all exchange-specific example-configs (config_binance.json.example and config_kraken.json.example), since these can be generated using new-config. I did not do this yet as i'd like to discuss this first.

I agree, no need to keep them. Or maybe it's better to move them into `user_data/configuration_examples` or `user_data/configuration`, whenever this dir can be called -- in the same way as strategy samples. We allow to create new strategies from templates, but populate samples as well...

freqtrade/commands/build_config_commands.py
xmatthias: That doesn't make sense ...

exchange and api-key / api-token are closely related. 
I will not split these up - that would be pointless and hopelessly confusing to users - and with that dry-run will stay before exchange.

If we're honest - we can also ask this differently:

* do you want to simulate trades?
* how many trades?
* in which stake currency?
* and on what exchange?

As a user, you'll need to know all answers anyway - or at least think about them while answering. 
The sequence of the questions does not create a different answer.

xmatthias:
> I agree, no need to keep them. Or maybe it's better to move them into user_data/configuration_examples or user_data/configuration, whenever this dir can be called -- in the same way as strategy samples. We allow to create new strategies from templates, but populate samples as well...

That's what i'm not certain about... 
SampleStrategy could in theory be created from the template too (it's just the full template...). This is something we can't (at the moment) build by asking questions.

For configurations, that's somewhat different. In theory we can ask every single option as question - and build the configuration "to fit your needs" - basically removing the need for these sample configs.
The risk with keeping them is that we forget to update them ... (either the sample configs, or the templates)...

freqtrade/commands/build_config_commands.py
hroff-1902: In my opinion the answer on "do you want to simulate trades?" also depends on the exchenge, selected, but... let it be as it is for now...

LGTM 👍

hroff-1902:
> > I agree, no need to keep them. Or maybe it's better to move them into user_data/configuration_examples or user_data/configuration, whenever this dir can be called -- in the same way as strategy samples. We allow to create new strategies from templates, but populate samples as well...
> 
> That's what i'm not certain about...
> SampleStrategy could in theory be created from the template too (it's just the full template...). This is something we can't (at the moment) build by asking questions.
> 
> For configurations, that's somewhat different. In theory we can ask every single option as question - and build the configuration "to fit your needs" - basically removing the need for these sample configs.
> The risk with keeping them is that we forget to update them ... (either the sample configs, or the templates)...

These are examples... Examples, available "at the fingertips"... No need to keep them in the main dir, let's move them to user_data for now...

PR #2833, opened 2020-02-02
Add some type hints

hroff-1902:
This helps finding errors like that in `FreqtradeBot.handle_trailing_stoploss_on_exchange()` (`return False` instead of `return`)... And this can emphasize, highlight some design problems as well.

Some notes/questions:
* `cast` in freqtrade/commands/data_commands.py -- is this a good case for using `typing.cast`? From one point of view this looks (well, a bit) like a code smell, from another point of view -- it's more efficient than calling `bool(...)` there. 

From third point of view, mypy may be made happy here by adding `erase: bool = config['erase']` var here. From forth point of view, we'd in this case make this (assigning a local var with the value taken from the config dir) in all other places where config values are passed to the functions with type hints. So probably `cast` here in its legal place...

From the fifth point of view, config should probably be a TypedDict, but only when we drop support for python <3.8.

* I dislike `CcxtModuleType` in freqtrade/exchange/exchange.py as it is now. Probably, we should use a bool parameter 'use_async' instead of passing `ccxt_module` to the functions where it's currently used...

* Using `-> 'TimeRange'` (in this string format) looks like code smell for me. Actually, I do not like completely this approach of resolving forward type references in python, hopefully this won't be needed with `futures annotations` when we drop support for python 3.6... The solution can be moving static `parse_timerange()` to a function at the module level...

I'm asking these 'generic' points because they define how freqtrade code will look like because similar issues exist (or may appear in the future) in other places as well...

### DIFF
--- a/freqtrade/commands/data_commands.py
+++ b/freqtrade/commands/data_commands.py
@@ -1,6 +1,6 @@
 import logging
 import sys
-from typing import Any, Dict, List
+from typing import Any, Dict, List, cast
 
 import arrow
 
@@ -43,16 +43,18 @@ def start_download_data(args: Dict[str, Any]) -> None:
         if config.get('download_trades'):
             pairs_not_available = refresh_backtest_trades_data(
                 exchange, pairs=config["pairs"], datadir=config['datadir'],
-                timerange=timerange, erase=config.get("erase"))
+                timerange=timerange, erase=cast(bool, config.get("erase")))
 
             # Convert downloaded trade data to different timeframes
             convert_trades_to_ohlcv(
                 pairs=config["pairs"], timeframes=config["timeframes"],
-                datadir=config['datadir'], timerange=timerange, erase=config.get("erase"))
+                datadir=config['datadir'], timerange=timerange,
+                erase=cast(bool, config.get("erase")))
         else:
             pairs_not_available = refresh_backtest_ohlcv_data(
                 exchange, pairs=config["pairs"], timeframes=config["timeframes"],
-                datadir=config['datadir'], timerange=timerange, erase=config.get("erase"))
+                datadir=config['datadir'], timerange=timerange,
+                erase=cast(bool, config.get("erase")))
 
     except KeyboardInterrupt:
         sys.exit("SIGINT received, aborting ...")
--- a/freqtrade/commands/data_commands.py
+++ b/freqtrade/commands/data_commands.py
@@ -1,6 +1,6 @@
 import logging
 import sys
-from typing import Any, Dict, List, cast
+from typing import Any, Dict, List
 
 import arrow
 
@@ -43,18 +43,18 @@ def start_download_data(args: Dict[str, Any]) -> None:
         if config.get('download_trades'):
             pairs_not_available = refresh_backtest_trades_data(
                 exchange, pairs=config["pairs"], datadir=config['datadir'],
-                timerange=timerange, erase=cast(bool, config.get("erase")))
+                timerange=timerange, erase=bool(config.get("erase")))
 
             # Convert downloaded trade data to different timeframes
             convert_trades_to_ohlcv(
                 pairs=config["pairs"], timeframes=config["timeframes"],
                 datadir=config['datadir'], timerange=timerange,
-                erase=cast(bool, config.get("erase")))
+                erase=bool(config.get("erase")))
         else:
             pairs_not_available = refresh_backtest_ohlcv_data(
                 exchange, pairs=config["pairs"], timeframes=config["timeframes"],
                 datadir=config['datadir'], timerange=timerange,
-                erase=cast(bool, config.get("erase")))
+                erase=bool(config.get("erase")))
 
     except KeyboardInterrupt:
         sys.exit("SIGINT received, aborting ...")
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -25,8 +25,6 @@ from freqtrade.exchange.common import BAD_EXCHANGES, retrier, retrier_async
 from freqtrade.misc import deep_merge_dicts
 
 
-# Should probably use typing.Literal when we switch to python 3.8+
-# CcxtModuleType = Literal[ccxt, ccxt_async]
 CcxtModuleType = Any
 
 
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -24,6 +24,12 @@ from freqtrade.exceptions import (DependencyException, InvalidOrderException,
 from freqtrade.exchange.common import BAD_EXCHANGES, retrier, retrier_async
 from freqtrade.misc import deep_merge_dicts
 
+
+# Should probably use typing.Literal when we switch to python 3.8+
+# CcxtModuleType = Literal[ccxt, ccxt_async]
+CcxtModuleType = Any
+
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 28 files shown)


hroff-1902:
^^ hmm, 2 messages from coveralls, with diff. results...

freqtrade/exchange/exchange.py
```diff
@@ -24,6 +24,12 @@
 from freqtrade.exchange.common import BAD_EXCHANGES, retrier, retrier_async
 from freqtrade.misc import deep_merge_dicts
 
+
+# Should probably use typing.Literal when we switch to python 3.8+
```
xmatthias: The point  here is not the "once we switch to python3.8" - it's "once we drop support for python <3.8" - which honestly, i don't see happen as there's nothing in 3.8 i think is a "must have" for us.

Maybe we should change the comment to "when support for <python3.8 is dropped" ? 

it's the same with 3.7 - but there's technically no reason to drop support for python3.6 at the moment as there've been no breaking changes.

freqtrade/commands/data_commands.py
```diff
@@ -43,16 +43,18 @@ def start_download_data(args: Dict[str, Any]) -> None:
         if config.get('download_trades'):
             pairs_not_available = refresh_backtest_trades_data(
                 exchange, pairs=config["pairs"], datadir=config['datadir'],
-                timerange=timerange, erase=config.get("erase"))
+                timerange=timerange, erase=cast(bool, config.get("erase")))
```
xmatthias: Any particular reason / advantage to use `cast(bool, xxx)` instead of `bool(xxx)`?

we're not using cast this way anywhere else ...

freqtrade/exchange/exchange.py
hroff-1902: I mean, Literal is markes as if it added only in 3.8: https://docs.python.org/3/library/typing.html?highlight=typing#typing.Literal

freqtrade/commands/data_commands.py
hroff-1902: Performance. `cast` generates no byte-code.

xmatthias: Why would you opt for the slower method? :laughing: 

2020-02-03-071939_653x157_scrot

2020-02-03-072006_674x168_scrot

freqtrade/exchange/exchange.py
xmatthias: I do understand that - and i don't disagree with leaving this comment there.

But we do support 3.8 - but we also support 3.7/3.6 - so we can't use this. 
That's not clear from the comment above ... it sounds more like "once we do the transition from 2.7 to 3.x ..."

freqtrade/commands/data_commands.py
hroff-1902: These are all your notebooks...

```
$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from timeit import timeit
>>> timeit("cast(bool, 200)", "from typing import cast")
0.11175005696713924
>>> timeit("bool(200)")
0.12564564507920295
>>> timeit("cast(bool, '200')", "from typing import cast")
0.11431943299248815
>>> timeit("bool('200')")
0.1346755709964782
>>> 
```

hroff-1902: > we're not using cast this way anywhere else ...

Well, we probably should in the future, in all other places where `float()`, `str()`, etc are called just to make mypy/flake happy...

freqtrade/exchange/exchange.py
hroff-1902: okay, removed the comment since it's unclear.

I meant here that Literal is only available in 3,8+, so till we support 3.6 and 3.7, we cannot use Literal here instead of current Any.

freqtrade/commands/data_commands.py
xmatthias: Wrong, it's the python version that matters, notebook or not does not matter for execution speed.

`bool()` is faster in 3.8 - and i'd stick to that since that's the "more modern".
3.7 is faster than 3.6, 3.8 is supposed to be faster than 3.7 ... 

```
python
Python 3.8.0 (default, Oct 18 2019, 09:56:01)
[GCC 7.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from timeit import timeit
>>> timeit("cast(bool, 200)", "from typing import cast")
0.09076810000010482
>>> timeit("bool(200)")
0.07849940000005518
>>> timeit("cast(bool, '200')", "from typing import cast")
0.1061870999999428
>>> timeit("bool('200')")
0.07970810000006168
>>>
```

hroff-1902: * Your examples ^^ are quite incorrect, since it's not really type conversion, the value is already bool, so should better testes as `cast(bool, True)` and `bool(True)` (and `bool('<int>')` always returns True, even for `bool('0')` 😉 )

* Okay, I see the same: in python 3.7 ans python 3.7 `bool()` is faster than `cast()` (as well as calls to `float()` and `int()` are faster as well). Which is strange because `cast` assumes to 'return the value as it is', it's simply a hint for the static checker that the developer knows the value *is to be* of expected type... And `bool()` conversion should at least perform type checking to see that the value is already bool and should simply be returned without conversion... So it's strange a bit...

* As I said at the top, this is more about how the code will look in the future, when similar recipes are applied in the codebase. We do not use `cast`, right, and I can simply change it here to `bool`, as in other places, we'll be always able to refactor code in the codebase in the future, But in my opinion `cast` is more ideomatic (however, lots of legacy python code uses type conversion functions -- that's because python typing is relatively new in many of its parts) because it simply hints static checker that the untyped value is of expected type. The conversion functions can modify the values if not used properly, like in the example of `bool('0')` above...

Summary: changing it to `bool`, right?

hroff-1902: Very strange why cast is slower:
https://github.com/python/cpython/blob/4b524161a0f9d50d782e739a3708434ffd4e94a5/Lib/typing.py#L1146-L1154

xmatthias: > Summary: changing it to bool, right?

yes
using cast() with the link you provided is a dangerous thing - it has you assume that a type is converted - even though it's not.

I'm also comming from another language background (not c/cpp) - but when using cast, i always assume that the returned type is of the new type - and not just telling linters / typecheckers it's that way.
So using `cast(float, '200')` is even worse than slower - it's giving us a completely false sense of correctness as you think "well - i did cast it" - but the return type is still a string (`cast(float, '200') == '200'`).

> Very strange why cast is slower

While this only returns the called function - it's a function call nonetheless.
I have to assume that `bool()` is some very heavily optimized piece of C-code (it's one of the core-types) ... even though i'm too lazy (and not interrested enough) in finding it in the Cpython codebase.

hroff-1902: > it has you assume that a type is converted - even though it's not.

No, using cast says to the interpreter (well, actually to the static type checker) that 'developer knows (or assumes) that the value is already of this type and no conversion required'. That's the semantics of the typing.cast.

xmatthias: of `typing.cast` - yes - which i find highly confusing (but i'm sceptical of everything i import from typing, anyway) - but in c# using cast (syntax is different though) does actually convert the type ...

hroff-1902: I'm only talking about python cast here. Using terminology from other languages, yes, I would say that `bool()` and `float()` are real casts, exactly... 🙄 
All above, where I use the term 'cast' -- that's about python typing.cast.

xmatthias: i know ... :+1:

PR #2834, opened 2020-02-02
better readability on sharpe ratio loss method

yazeed:
No change of logic here, just updated hyperopt loss sharpe loss method for better readability to make it more consistent with the final version of sharpe loss daily loss method.


yazeed:
This should be merged along with #2826

PR #2838, opened 2020-02-02
Add list-hyperopts subcommand

hroff-1902:
Pls verify I did not add an unsolicited dependency on hyperopt at the global level.

* Docs for `list-strategy` slightly refined to be aligned with `list-hyperopt`.


hroff-1902:
oops, wrong base, sorry

PR #2839, opened 2020-02-02
Add list-hyperopts subcommand

hroff-1902:
Pls verify I did not add an unsolicited dependency on hyperopt at the global level.

* Docs for `list-strategy` slightly refined to be aligned with `list-hyperopt`.

### DIFF
--- a/freqtrade/commands/list_commands.py
+++ b/freqtrade/commands/list_commands.py
@@ -38,7 +38,7 @@ def start_list_exchanges(args: Dict[str, Any]) -> None:
 
 def start_list_strategies(args: Dict[str, Any]) -> None:
     """
-    Print Strategies available in a directory
+    Print files with Strategy custom classes available in the directory
     """
     config = setup_utils_configuration(args, RunMode.UTIL_NO_EXCHANGE)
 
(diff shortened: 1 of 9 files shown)


hroff-1902:
@xmatthias pls take a look at the last commit, I wonder why I got this inconsistency...

xmatthias:
> for 8415687 -- it's not covered by tests. I mean, I should not find that this fails only when I remove config.json, there should be a generic test covering this somehow...

well you can't really generalize such a test - as either in the test (list) - or in this list, you need to have this set correctly and with intend.

It is covered by tests if the test is written correctly (removing `config.json` from the parsed args) ... 

Your test "almost" covers it - but the key line (setting `pargs['config']=None`) is commented out.

``` python
pargs = get_args(args)
# pargs['config'] = None
start_list_hyperopts(pargs)
```

freqtrade/commands/list_commands.py
```diff
@@ -54,6 +54,26 @@ def start_list_strategies(args: Dict[str, Any]) -> None:
         print(tabulate(strats_to_print, headers='keys', tablefmt='pipe'))
 
 
+def start_list_hyperopts(args: Dict[str, Any]) -> None:
+    """
+    Print files with HyperOpt custom classes available in the directory
+    """
+    from freqtrade.resolvers.hyperopt_resolver import HyperOptResolver
```
xmatthias: > Pls verify I did not add an unsolicited dependency on hyperopt at the global level.

That would only happen if you add this import to the top of the file ... which you did not, so that's fine :+1:

PR #2843, opened 2020-02-03
Allow derived strategies

hroff-1902:
...like this:

```
class SampleStrategy2(SampleStrategy):
    # Override something
    trailing_stop = True
    # Methods can also be overridden if we need this
```

`list-strategies` also works:
```
$ freqtrade list-strategies
...
| name               | location                |
|:-------------------|:------------------------|
| SampleStrategy     | sample_strategy.py      |
| SampleStrategy2    | sample_strategy.py      |
```

P.S. Same should work for hyperopt classes. Can be very useful, since you can simply redefine the `indicator_space()` method(s), for example with different range for dimension(s), leaving the rest intact.

### DIFF
--- a/docs/strategy-customization.md
+++ b/docs/strategy-customization.md
@@ -532,6 +532,27 @@ If you want to use a strategy from a different directory you can pass `--strateg
 freqtrade trade --strategy AwesomeStrategy --strategy-path /some/directory
 ```
 
+### Derived strategies
+
+The strategies can be derived from other strategies. This avoids duplication of your custom strategy code. You can use this technique to override small parts of your main strategy, leaving the rest untouched:
+
+```
+class MyAwesomeStrategy(IStrategy):
+    ...
+    stoploss = 0.13
+    trailing_stop = False
+    # All other attributes and methods are here as they
+    # should be in any custom strategy...
+    ...
+
+class MyAwesomeStrategy2(MyAwesomeStrategy):
+    # Override something
+    stoploss = 0.08
+    trailing_stop = True
+```
+
+Both attributes and methods may be overriden, altering behavior of the original strategy in a way you need. The strategy classes may be located in the same module (python file with the source code of your strategy) or in different modules (different python files). In the latter case you need to properly import the strategy class you derive the new one from.
+
 ### Common mistakes when developing strategies
 
 Backtesting analyzes the whole time-range at once for performance reasons. Because of this, strategy authors need to make sure that strategies do not look-ahead into the future.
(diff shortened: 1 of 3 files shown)


xmatthias:
> `class SampleStrategy2(SampleStrategy):`

this seems okish and useful - but if i merge this now it'll be the 2 of us (+ yazeed who gave a thumpsup) who know about it ...

-> should be added to the docs somehow.

docs/strategy-customization.md
```diff
@@ -532,6 +532,27 @@ If you want to use a strategy from a different directory you can pass `--strateg
 freqtrade trade --strategy AwesomeStrategy --strategy-path /some/directory
 ```
 
+### Derived strategies
+
+The strategies can be derived from other strategies. This avoids duplication of your custom strategy code. You can use this technique to override small parts of your main strategy, leaving the rest untouched:
+
+```
```
xmatthias: maybe add syntx highlighting?

``` suggestion
``` python
```

hroff-1902:
applied highlighting to advanced-hyperopt part as well, seems to be ready.

docs/strategy-customization.md
```diff
@@ -532,6 +532,27 @@ If you want to use a strategy from a different directory you can pass `--strateg
...
+    # should be in any custom strategy...
+    ...
+
+class MyAwesomeStrategy2(MyAwesomeStrategy):
+    # Override something
+    stoploss = 0.08
+    trailing_stop = True
+```
+
+Both attributes and methods may be overriden, altering behavior of the original strategy in a way you need. The strategy classes may be located in the same module (python file with the source code of your strategy) or in different modules (different python files). In the latter case you need to properly import the strategy class you derive the new one from.
```
xmatthias: > or in different modules (different python files). In the latter case you need to properly import the strategy class you derive the new one from.

I understand it works (technically), and i can get it to work if i must - but it needs some more than just 2 files to get this to work. 
It needs to be in a package (folder containing `__init__.py`) - otherwise the import won't work correctly - which i think makes this a hurdle for non-python developers.
I'm also not 100% certain if relative imports (`from .awesomestrategy import MyAwesomeStrategy`) would work with our import method (and i did not try yet).

Now i'm sceptical that the everyday trader user will get this to work ...

hroff-1902: so should I remove this part?
I'm not sure python3 needs empty `__init__.py`...

hroff-1902:
@xmatthias 
^^ CI still randomly fails...
(just to note you, not relates with the subj)

I removed that sentences about diff. modules...

docs/strategy-customization.md
xmatthias: to import relatively? 
I'm pretty sure it does as i've been fighting that at work today...

hroff-1902: okay, I removed this to keep the docs simple

xmatthias:
> ^^ CI still randomly fails...

i know ... i'm very close to giving up on that ...

PR #2850, opened 2020-02-03
make sure asyncio_loop is not initialized within ccxt code

xmatthias:
## Summary
Another try to fix the random test failure ...

This is the only test where ccxt is not mocked away (which is correct, as that's what's tested).

## Quick changelog

- ensure that asyncio_loop is not started


hroff-1902:
LGTM 👍

PR #2854, opened 2020-02-03
Fix get_timerange()

hroff-1902:
Common timerange is [max of minimums:min of maximums], not [min of minimums:max of maximums]...

### DIFF
--- a/freqtrade/data/history.py
+++ b/freqtrade/data/history.py
@@ -455,8 +455,8 @@ def get_timerange(data: Dict[str, DataFrame]) -> Tuple[arrow.Arrow, arrow.Arrow]
         (arrow.get(frame['date'].min()), arrow.get(frame['date'].max()))
         for frame in data.values()
     ]
-    return (min(timeranges, key=operator.itemgetter(0))[0],
-            max(timeranges, key=operator.itemgetter(1))[1])
+    return (max(timeranges, key=operator.itemgetter(0))[0],
+            min(timeranges, key=operator.itemgetter(1))[1])
 
 
 def validate_backtest_data(data: DataFrame, pair: str, min_date: datetime,


hroff-1902:
@xmatthias 
Do you agree with this fix?
If so, I'll try to fix the tests that fails now, this seems to be not so easy...

hroff-1902:
I'll better close this and open an issue for discussion first.
Leaving this for a while for you to see the proposed change.

xmatthias:
> Do you agree with this fix?

I'm not certain.
Let's first draw an example case:
pair A has data from 2019.01.01 - 2019.10.30
Pair B has data from 2019.05.01 - 2019.10.30
Pair C has data from 2019.01.01 - 2019.08.01

### current behaviour when Backtesting the above pairs:

`get_timerange()` will return `(2019.01.01, 2019.10.30)` - the maximum available range.
Backtesting for Pair B will start in May - while Pair C will "end" in August.


### Proposed change
`get_timerange()` will return `(2019.05.01, 2019.08.01)` - the minimum available range for all pairs.
It'll throw out 5 Months of good data for pair A - at the start
It'll throw out 4 months of good data for pairs A and B, and leave me with 3 months of backtesting.

Now this is a small-scale example - but assume you got 20x pair A - and 1x PairB, and 1x PairB. You've just thrown away 90% of your valid backtesting data.

----

So to me, this new behaviour is problematic.

Assume i want to backtest all of 2019 - but also include some "newer" paris which started trading only in April. 
the user will be unable to backtest the whole period - even though for the real, live bot,  the user would have added these pairs to the whitelist in April (either through Volumepairlist, or manually - but that's not relevant here). 

So now you're forcing the user to split up the backtest into 2 - Jan - April with 1 fewer pair - and April - Dec for all pairs i wanted to backtest.

The problems with the split backtest are 2:
* more unnecessary work for the user
* assuming a high max_open_trades - it'll taint backtest as the bot forceclosed all trades at the beginning of April (when the new pair started trading / the first backtest-period ended).
 So the aggregated result will not be the same that what was available before.

We also keep in mind that backtesting does handle missing data for one or more pairs both at the start and at the end - so i think this is fixing a problem that's not there.

xmatthias:
Will close this due to no feedback to the above - and because i think the proposed change would heavily limit the availability of backtesting data, as the only possible backtest range is the value with the smallest common data ... which is unrealistic considering pairlistings / delistings.

PR #2857, opened 2020-02-04
Unfilledtimeout override for buy orders

yazeed:
### Backstory:

I've been test trading a strategy for the past 7 days and missed lots of trades where the orderbooks were heavily stacked (0 spread, quantities too high on both sides of the order book, buy order needing a longer duration before its turn is reached). What made it a real annoyance was when my strategy would cancel my buy order due to _unfilledtimeout_ expiring, only to then place a buy order at the same price again, making it unlikely to ever hit since my new buy order enters at the end of the bids stack every time freqtrade does this.

### What this is:

This introduces a configuration option within bid_strategy, named **timeout_even_if_buy_signal_valid**, defaulting to **true** to match existing behavior, with the aim of giving you more control over your strategy's buy orders (entry orders) if you decide to use it. Once you switch **timeout_even_if_buy_signal_valid** to **false**, it acts to prevent your buy orders from timing out **unless** a sell signal is made (so it cancels your buy order only when a sell signal is detected).

### Available workarounds:

Increasing **unfilledtimeout** wasn't much of a viable solution for me, because I wasn't sure how much would I need to safely increase it, and if I actually did try to increase it, there would still be a possibility of the price moving with my buy order still intact, and back to the expired/no-longer-desired buy signal. By this sense, a sell signal should invalidate the buy signal before it.

### Alternative naming:

I found it really difficult to come up with meaningful names for the configuration, so feel free to propose suitable naming.

### Your turn:

Would love to hear your thoughts, whether you agree to this proposal or not, or if you might imagine it done differently.


freqtrade/freqtradebot.py
```diff
@@ -845,44 +845,60 @@ def handle_timedout_limit_buy(self, trade: Trade, order: Dict) -> bool:
...
-        })
-        return False
+        # get config for bid strategy
+        config_bid_strategy = self.config.get('bid_strategy', {})
+
+        # proceed to cancel buy order by timeout if timeout_even_if_buy_signal_valid
+        # is true (original behaviour) -OR-
+        # cancel buy order only if buying condition is no longer valid OR if there's
+        # a sell signal present
+        if config_bid_strategy.get('timeout_even_if_buy_signal_valid', True) or (not buy or sell):
```
xmatthias: personally, i'd move the "decision logic" out of this method and into `check_handle_timedout()`.

That'll give us 2 advantages:
* Implementing this for sell-side is simpler
* the decision is taken before this method - so this methods' purpose is to apply the decision, not to decide anything.

yazeed: Makes sense, I'll make a commit even if we're not there yet in terms of how this should finally look like, just to bring it closer to reality.

yazeed:
It is pair specific, but also market-specific (spot cryptocurrency trading is not very liquid or volatile, as opposed to USDT pairs or futures), and it's more common than I initially thought it was, specially after looking at order books in exchanges that are even less popular than binance. I was testing it trading top 60-90 BTC pairs of binance in terms of volume, and yet often hit this scenario in the past 7 days, even when those low-sat pairs were having good volumes but still not enough volatility.

Side point: People who choose to run ferqtrade with this option on might end up having lots of pending buy orders at the end of the day (I increase max_open_trades for this purpose), but I think that's entirely fine, as it seems more representative of my strategy's configuration and pair-selections. Way better for me than witnessing it canceling orders and placing them back on again at the exact same spot.

As a bottom line, I simply dont want freqtrade to cancel my buy orders as long as the buy signal is still valid, unless, it's invalidated by a following sell-signal, or hits a maximum defineable waiting time of 2-3 days (can be added as another sub-rule in another update). But at the moment, I cannot entirely rely on increasing unfilledtimeout to 2-3 days, without a way out of it during.

Finally, I've made another commit that enhances readability and structures config in a better way, and expands on unfilledtimeout (renamed to unfilled_timeout for consistency and turned into an object with multiple objects, buy and sell, and sub-attributes for each).

yazeed:
You've mentioned the interface.py strategy callback approach, and while I agree that it seems more logical for this to be within the strategy callback, I feel that the way changes are proposed here would have a lower impact on how freqtrade currently works, and doesn't increase complexity for the user.

Also, speaking of the usecases you mentioned, lets call them buy signal invalidators, I think there's only three I can think of, and to me they're not really strategy specific, I'd use them globally across all of my strategies:

1. A sell signal arriving after a buy signal, invalidates the buy signal before it.
2. Waiting for an overly extended period of time (maximum threshold) for the buy order to process (say 24-60 hours?). To me, this represents a maximum time-based threshold that should work alongside side, point 1, and unfilled_timeout.
3. Orderbooks getting abnormally and heavily stacked against a buy order, signalling a likely strong dump in price. (Hard to define and implement, but I know many quant execution engines have this measure)

This whole feature works at allowing buy-invalidator #1, and if approved, I can present buy-validator point #2.

yazeed:
@xmatthias Would love to have this PR reviewed and ultimately merged or discussed further if needed.

Please note that this PR, as it stands introduces no change for most users since it defaults to existing behavior (cancels buy orders based on unfilled_timeout).

The only addition is something that I consider a bug-fix, this PR now allows cancellation of buy orders regardless of configuration whenever a sell signal is made for that same pair. This makes it more consistent with the same rule that already exists without this PR, and gets checked under **create_trade()** _if buy and not sell:_ before placing a buy order.

freqtrade/constants.py
```diff
@@ -94,11 +94,22 @@
         'trailing_stop_positive': {'type': 'number', 'minimum': 0, 'maximum': 1},
         'trailing_stop_positive_offset': {'type': 'number', 'minimum': 0, 'maximum': 1},
         'trailing_only_offset_is_reached': {'type': 'boolean'},
-        'unfilledtimeout': {
+        'unfilled_timeout': {
```
xmatthias: This is higly problematic. You're breaking everyones strategy / configuration.
While i agree, `unfilled_timeout` looks better - i don't want to have everyone modify configuration and/or strategy for no benefit.


> Please note that this PR, as it stands introduces no change for most users since it defaults to existing behavior (cancels buy orders based on unfilled_timeout).

This is not true - as you've made my configuration i had for "unfilledtimeout" invalid.
Since the new configuration has defaults, users won't even notice (maybe not for months) - until the order uses the default timeout while the user expected it not to because it was changed to "something else" (higher or lower).

config_full.json.example
```diff
@@ -20,9 +20,14 @@
...
-        "buy": 10,
-        "sell": 30
+    "unfilled_timeout": {
+        "buy": {
+            "after": 10,
+            "even_if_buy_signal_valid": true
+        },
+        "sell": {
+            "after": 30
+        }
```
xmatthias: I don't particularly like the nested dicts here (mainly because it'd break existing configurations).

How about the following:


``` json
    "unfilled_timeout": {
     "buy": 10,
     "buy_timeout_with_buy_signal": 60,
     "sell": 30
    },	  
```

Notice that it'll use a number, not a boolean - so you can simply use a "longer" timeout in this case (something discussed above).

Implementation wise, this should fall back to "buy" if not set - keeping the "original" behaviour.

yazeed:
Closed as the discussion moved to #2872.

PR #2864, opened 2020-02-05
Enhanced Backtesting Tables

yazeed:
Updated Strategy Summary to be consistent with other backtesting tables...

```
=========================================================== STRATEGY SUMMARY ===========================================================
| Strategy    |   Buy Count |   Avg Profit % |   Cum Profit % |   Tot Profit BTC |   Tot Profit % | Avg Duration   |     Wins | Losses |
|:------------|------------:|---------------:|---------------:|-----------------:|---------------:|:---------------|---------:|-------:|
| Strategy1   |         429 |           0.36 |         152.41 |       0.00762792 |          76.20 | 4:12:00        |      186 |    243 |
| Strategy2   |        1487 |          -0.13 |        -197.58 |      -0.00988917 |         -98.79 | 4:43:00        |      662 |    825 |
```

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -237,8 +237,8 @@ There will be an additional table comparing win/losses of the different strategi
 Detailed output for all strategies one after the other will be available, so make sure to scroll up to see the details per strategy.
 
 ```
-=========================================================== Strategy Summary ===========================================================
-| Strategy    |   buy count |   avg profit % |   cum profit % |   tot profit BTC |   tot profit % | avg duration   |   profit |   loss |
+=========================================================== STRATEGY SUMMARY ===========================================================
+| Strategy    |   Buy Count |   Avg Profit % |   Cum Profit % |   Tot Profit BTC |   Tot Profit % | Avg Duration   |     Wins | Losses |
 |:------------|------------:|---------------:|---------------:|-----------------:|---------------:|:---------------|---------:|-------:|
 | Strategy1   |         429 |           0.36 |         152.41 |       0.00762792 |          76.20 | 4:12:00        |      186 |    243 |
 | Strategy2   |        1487 |          -0.13 |        -197.58 |      -0.00988917 |         -98.79 | 4:43:00        |      662 |    825 |
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -441,7 +441,7 @@ class Backtesting:
             print()
         if len(all_results) > 1:
             # Print Strategy summary table
-            print(' Strategy Summary '.center(133, '='))
+            print(' STRATEGY SUMMARY '.center(133, '='))
             print(generate_text_table_strategy(self.config['stake_currency'],
                                                self.config['max_open_trades'],
                                                all_results=all_results))
--- a/freqtrade/optimize/optimize_reports.py
+++ b/freqtrade/optimize/optimize_reports.py
@@ -121,9 +121,9 @@ def generate_text_table_strategy(stake_currency: str, max_open_trades: str,
 
     floatfmt = ('s', 'd', '.2f', '.2f', '.8f', '.2f', 'd', '.1f', '.1f')
     tabular_data = []
-    headers = ['Strategy', 'buy count', 'avg profit %', 'cum profit %',
-               f'tot profit {stake_currency}', 'tot profit %', 'avg duration',
-               'profit', 'loss']
+    headers = ['Strategy', 'Buy Count', 'Avg Profit %', 'Cum Profit %',
+               f'Tot Profit {stake_currency}', 'Tot Profit %', 'Avg Duration',
+               'Wins', 'Losses']
     for strategy, results in all_results.items():
         tabular_data.append([
             strategy,
--- a/tests/optimize/test_optimize_reports.py
+++ b/tests/optimize/test_optimize_reports.py
@@ -91,14 +91,14 @@ def test_generate_text_table_strategy(default_conf, mocker):
     )
 
     result_str = (
-        '| Strategy   |   buy count |   avg profit % |   cum profit % '
-        '|   tot profit BTC |   tot profit % | avg duration   |   profit |   loss |\n'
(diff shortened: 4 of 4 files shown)


yazeed:
More changes to be made here.

yazeed:
Nevermind, I've worked all the way through to calculating expectancy, only to find that it actually matches (Avg Profit %) at the end, so Avg Profit % is the ultimate strategy scorer for me, now. :-)

So, this is actually ready and final from my side.

yazeed:
By the way, here's an additional iteration with a draws column which I figured was also missing, everywhere not just in the strategy results table, since if you're using minimal_roi to sell at 0% profit after a certain time, there will be a lot of these draws and it's important to see them, not committed yet, just want to see your thoughts here first.

xmatthias:
I don't mind more information in general ... but i'm a bit concerned that too much information can be overwelming.

Also, screen width can be a problem too (not sure how many chars in width this takes ... but i know one thing - once this starts wrapping lines - it'll loose usefulness.


maybe (just an idea) we can not show most of these new columns for all pairs - and only show them as a summary at the end?


Also, expectancy will require good explanation - what's a good / bad expectancy value?
We've seen it with edge - 100ds of questions of "this is my result - is it good?



Assuming the calculation is the same than in edge - i find it weired that you have 95% win rate (JollingerBands), an average profit of 0.19%, total profit of 105% - but still only expect your total capital to expand by 0.39% (0.0039) ...

Maybe i'm misreading this (i've not done much with expectancy before) - but should't this be aligned to the result of total-profit (somehow) ?

yazeed:
Sorry for the repetitive posts, just documenting my journey step by step, here's what the final table looks like (not committed to this PR yet)

Screenshot 2020-02-05 07 21 18

hroff-1902:
what if I do not have such a wide display?

yazeed:
> what if I do not have such a wide display?

Then you have a problem, which you can solve by scrolling to the right. I was just thinking about that.

xmatthias:
This final table is definitely too wide.
It'll wrap on most pc's with full-hd resolution ...

yazeed:
What do you suggest we can do about it? I know that it has all the info anyone might need to analyze a strategy, which is in my own rationale, the purpose of that strategy comparison table, to be able to thoroughly and delicately compare between strategies.

xmatthias:
an (extreme) example from my terminal (i've intentionally reduced the wide as i don't have the new columns available).

2020-02-05-072745_1019x252_scrot

Now the information is still okish as only the last column  ... but you see the wrapping.
Once it starts wrapping, its' very difficult to get anything usefull out of this - so instead of giving users all information needed, you've removed most information from being easily accessible.

yazeed:
> an (extreme) example from my terminal (i've intentionally reduced the wide as i don't have the new columns available).
> 
> 2020-02-05-072745_1019x252_scrot
> 
> Now the information is still okish as only the last column ... but you see the wrapping.
> Once it starts wrapping, its' very difficult to get anything useful out of this - so instead of giving users all information needed, you've removed most information from being easily accessible.

I understand, I'm not saying this is okay, I actually prefer the 133 guideline provided in other tables, but that kinda limits information available. I'm not sure which approach should we take to fix this.

hroff-1902:
Printing extra options are just fine... But output quickly becomes too wide for many displays and screens... That's why I did not add many interesting metrics yet...

I guess it should be controlled by the user, in cli options, for example. Like `indicator*` options for plotting. So if you need extra metrics printed, use 
```
$ freqtrade backtesting ... --extra-metrics Expectancy Draws SharpeRatio CalmarRatio
```
or something like that...

xmatthias:
I think I do have a better idea / proposal (came to me while driving).

We could use a HTML output for this - this would not have these limitations, and would give us a lot more flexibility.

It doesn't have to be anything fancy - a jinja2 template (look at our template folder), rendered with the output as a table (`<table>`) (multiple tables).

In HTML / a browser, you can pan (scroll left / right) with easy / without limitation.
Basically, we can rebuild the table output as it is here - but with the additional columns (to start with).

Just render the template, storing it as a file in `userdata/backtest_results/` - ideally as `results<timestamp)>.html` - and open the users default browser (optionally??).

In the end, we can add more things to it as we have several open issues about added metrics and the highest pain-point was the "how to display".

It'll also tie in rather nicely with the (long-term) plan to have an UI - which will require this step anyway - so we can also start building that step by step.

----

for this PR, I'd like to 
* Merge this PR as is (no additional columns - just "stylistic fixes")
* move the discussion around the html template to an issue / a seperate PR which aims at introducing HTML backtesting results.

yazeed:
I like the HTML idea. I could work on that in a different PR.

Feel free to merge this PR as it stands, I haven't made any of the additional commits as I was merely on a quest exploring what's missing in the strategy results table.

hroff-1902:
yeah, html is cool.

yazeed:
> Ok, will merge this as is. 👍
> 
> Please make sure to keep the code for how to generate the extended table - you'll need the calculations for the HTML part anyway. 😄
> 
> Thanks for offering to work on the HTML part, too - i think that'll be a GREAT new feature, so i'm really looking forward to that!
> 
> Feel free to open an "early"/WIP PR so we can collaborate if you need any help or got questions regarding the templating (not sure if you've used that before).

Thanks a lot for your feedback and support on all my PRs, I look forward to contributing more stuff in alignment with freqtrade. I'm learning as we move forward. I'll be looking more into templating with python, if you have any pointers or advice, it would be great if you can share them.

BTW, it was announced on Slack that the extended table was released to develop, but it's not accurate, only styling were released in this PR because I didn't commit the work I had on it for the extended tables.

Will open the new PR sometime this weekend, so we can start collaborating on it.

hroff-1902:
> BTW, it was announced on Slack that the extended table was released to develop, but it's not accurate, only styling were released in this PR because I didn't commit the work I had on it for the extended tables.

Oops, my bad, did not checked this in enough details, thought the pictures you posted were actual for this PR...

PR #2870, opened 2020-02-06
Add considerations for dry-run

xmatthias:
## Summary
Add some considerations for dry-run - which may be important to understand when using dry-run - and should answer a few questions up-front.

## Quick changelog

- Dry-run considerations

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -608,6 +608,14 @@ Once you will be happy with your bot performance running in the Dry-run mode, yo
 !!! Note
     A simulated wallet is available during dry-run mode, and will assume a starting capital of `dry_run_wallet` (defaults to 1000).
 
+### Considerations for dry-run
+
+* API-Keys may or may not be provided. Only Read-Only operations on the exchange are performed in dry-run mode.
+* Wallets (`/balance`) will be simulated.
+* Orders will be simulated, and will not be posted to the exchange
+* In combination with `stoploss_on_exchange`, the stop_loss price is assumed to be filled.
+* Open orders (not Trades!) are reset on bot restart.
+
 ## Switch to production mode
 
 In production mode, the bot will engage your money. Be careful, since a wrong
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -610,11 +610,11 @@ Once you will be happy with your bot performance running in the Dry-run mode, yo
 
 ### Considerations for dry-run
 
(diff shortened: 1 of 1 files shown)


hroff-1902:
* I did not get what "In combination with `stoploss_on_exchange`, the stop_loss price is assumed to be filled." means
* Is "Open orders (not Trades!) are reset on bot restart." actually relevant? Orders are to be fulfilled momentally since they are simulated...
* I added some words in the parenthesis (couple of clarifications), I'm not sure if these clarifications are really needed, you may remove them if see they are redundant...

xmatthias:
> I did not get what "In combination with stoploss_on_exchange, the stop_loss price is assumed to be filled." means

Well - when you don't use stoploss_on_exchange - you may set a stoploss of 5% - but if price moves very fast - then the stoploss may trigger at 5.4% - or even lower (depending on the move it made).
stoploss_on_exchange does not "suffer" from this and emulates "realistic" behaviour (the stoploss is on the exchange - so the assumption is that it'll trigger).

> Orders are to be fulfilled momentally since they are simulated...

well it's not 100% true. limit-orders are left hanging for 1 iteration - while stoploss orders are filled immediately. 
You can stop the bot at that time (by default you got 5s - so the chance is pretty high to quit at that time) - so the "open" order will be cancelled on startup.

> I added some words in the parenthesis (couple of clarifications), I'm not sure if these clarifications are really needed, you may remove them if see they are redundant...

:+1: looks good the way you modified it.

hroff-1902:
> > I did not get what "In combination with stoploss_on_exchange, the stop_loss price is assumed to be filled." means
> 
> Well - when you don't use stoploss_on_exchange - you may set a stoploss of 5% - but if price moves very fast - then the stoploss may trigger at 5.4% - or even lower (depending on the move it made).
> stoploss_on_exchange does not "suffer" from this and emulates "realistic" behaviour (the stoploss is on the exchange - so the assumption is that it'll be filled at 5%).
> 
> it's not easy to word this as a concise explanation ...

Yeah, that's not easy to figure out in a few words, let it be for now as it is, we'll see if it create any questions...

> > Orders are to be fulfilled momentally since they are simulated...
> 
> well it's not 100% true. limit-orders are left hanging for 1 iteration - while stoploss orders are filled immediately.
> You can stop the bot at that time (by default you got 5s - so the chance is pretty high to quit at that time) - so the "open" order will be cancelled on startup.

Okay, if you catch them (the stoploss orders).

PR #2872, opened 2020-02-06
Buy order timeout callback

### LINKED ISSUE #2866: Unable to analyze ticker for pair WRX/BTC: negative dimensions are not allowed
I shut down the boot and restarted it. I encountered this error message

```
2020-02-05 11:39:13,738 - VolumePairList - INFO - Searching 20 pairs: ['XRP/BTC', 'ICX/BTC', 'ETH/BTC', 'MATIC/BTC', 'ENJ/BTC', 'WRX/BTC', 'EOS/BTC', 'AION/BTC', 'TRX/BTC', 'LTC/BTC', 'ETC/BTC', 'ADA/BTC', 'XTZ/BTC', 'WAN/BTC', 'BAT/BTC', 'MTL/BTC', 'XLM/BTC', 'ATOM/BTC', 'LINK/BTC', 'RVN/BTC']
2020-02-05 11:39:13,750 - freqtrade.rpc.rpc_manager - INFO - Enabling rpc.telegram ...
2020-02-05 11:39:13,957 - freqtrade.rpc.telegram - INFO - rpc.telegram is listening for following commands: [['status'], ['profit'], ['balance'], ['start'], ['stop'], ['forcesell'], ['forcebuy'], ['performance'], ['daily'], ['count'], ['reload_conf'], ['show_config'], ['stopbuy'], ['whitelist'], ['blacklist'], ['edge'], ['help'], ['version']]
2020-02-05 11:39:13,978 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': status, 'status': 'running'}
2020-02-05 11:39:14,076 - freqtrade.worker - INFO - Changing state to: RUNNING
2020-02-05 11:39:14,077 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': warning, 'status': 'Dry run is enabled. All trades are simulated.'}
2020-02-05 11:39:14,117 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': custom, 'status': "*Exchange:* `binance`\n*Stake per trade:* `1 BTC`\n*Minimum ROI:* `{'0': 0.21}`\n*Trailing Stoploss:* `-0.2`\n*Ticker Interval:* `1h`\n*Strategy:* `strateji1`"}
2020-02-05 11:39:14,158 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': status, 'status': "Searching for BTC pairs to buy and sell based on [{'VolumePairList': 'VolumePairList - top 20 volume pairs.'}]"}
2020-02-05 11:39:19,559 - freqtrade.strategy.interface - WARNING - Unable to analyze ticker for pair WRX/BTC: negative dimensions are not allowed
2020-02-05 11:39:20,710 - freqtrade.freqtradebot - INFO - Bot heartbeat. PID=16343
2020-02-05 11:39:21,135 - freqtrade.strategy.interface - WARNING - Unable to analyze ticker for pair WRX/BTC: negative dimensions are not allowed
2020-02-05 11:39:26,142 - freqtrade.strategy.interface - WARNING - Unable to analyze ticker for pair WRX/BTC: negative dimensions are not allowed
^C2020-02-05 11:39:26,244 - freqtrade.utils - INFO - SIGINT received, aborting ...
2020-02-05 11:39:26,245 - freqtrade.utils - INFO - worker found ... calling exit
2020-02-05 11:39:26,245 - freqtrade.rpc.rpc_manager - INFO - Sending rpc message: {'type': status, 'status': 'process died'}
2020-02-05 11:39:26,283 - freqtrade.freqtradebot - INFO - Cleaning up modules ...
2020-02-05 11:39:26,283 - freqtrade.rpc.rpc_manager - INFO - Cleaning up rpc modules ...
```

xmatthias:
## Summary
Inspired by the discussions in #2857 - this is a first (very rough) draft of how a strategy interface to customize buy_timeouts could look like.

It's intentionally very simple - it returns False by default (so no impact to users not interrested in this) - however when a strategy implements this (overrides this function) - a user can suddenly very flexibly control how and when to cancel orders.

This may accidentally (on purpose?) also fix #2866 - as the exception message is changed and now clearly points to "strategy code caused this exception".

### Example:

* Cancel order after 3 minutes for pairs with value > 100 USDT
* cancel order after 20 minutes for pairs with value > 10USDT
* cancel order after 24h if value < 10USDT (keep order there).

```python
def check_buy_timeout(self, pair: str, trade: Trade, order: Dict, **kwargs) -> bool:
    if trade.open_rate > 100 and trade.open_date < datetime.utcnow() - timedelta(minutes=5):
        return True
    elif trade.open_rate > 10 and trade.open_date < datetime.utcnow() - timedelta(minutes=3):
        return True
    elif trade.open_rate < 1 and trade.open_date < datetime.utcnow() - timedelta(hours=24):
       return True
    return False
```

## Quick changelog

- Introduce generic wrapper for strategy functions (catching all errors users may cause)
- introduce `check_buy_timeout()` - which can be used to customize buy order timeouts
- introduce `check_sell_timeout()` - which can be used to customize sell order timeouts
- add `new-strategy --template advanced` - which populates these optional methods too. Additional methods can then be added easily.

## TODO
* [x] Create same for sell side
* [x] create tests
* [x] Test it actually works as expected
* [x] create documentation
  * [x] add example usage
  * [x] with a few "common" samples
    * [x] validate 2nd example works ... 
  * [x] add to full sample strategy


hroff-1902:
Some general notes:

1) this is a first custom callback of possible many -- for adjusting the amount (position management), adjusting buy and/or sell price, etc.
2) this is a custom, user-written callback, so interface should be kept simple for a non-experienced user.
3) we'll be asking for adding more data available by the users; for instance, ticker data instead of orderbook.

as for 2), kwargs are a kind of code smell (a bit); also, when we add some more parameters passed, we'll need to explain how to change the method signature to accept this new parameter.

as for 3), interface with kwargs is extensible, right, but when we'll add some more data, they will be fetched and passed everytime, even if the user's code does not utilize it, creating overhead and even possible extra calls to exchange.

another approach is current approach used in `populate_*()` methods -- with `params` dict, it's in my opinion better from 2) point (no code smells), but does not resolve 3) -- if we add new data item, it will be passed everytime.

I see the following possible solutions:
a)
```
@need_orderbook
@need_trade
@need_ohlcv
def check_buy_timeout(self, pair: str, trade: Trade, orderbook: Dict, ohlcv: List) -> bool:
    ...
```
and
```
@need_orderbook
def check_buy_timeout(self, pair: str, orderbook: Dict) -> bool:
    ...
```
-- this resolves overhead issue (data are only passed when requested with decorators). Using decorators allows to check usage of data during backtesting/hyperopt: so a strategy with `@use_ohlcv` is compatible with backtesting, while ones with `@use_orderbook` should produce exception.

b)
```
def check_buy_timeout(self, pair: str, params: Dict) -> bool:
    ...
```
and
```
def check_buy_timeout(self, pair: str, params: Dict) -> bool:
    ...
```
-- looks better, no need to explain signatures for users. Cannot be checked for compatibility with backtesting, does not resolve overhead issue (data are prepared and passed in params even if not used by the user)
Pro: interface known to users

c)
```
@need_orderbook
@need_trade
@need_ohlcv
def check_buy_timeout(self, pair: str, params: Dict) -> bool:
    params['orderbook']
    params['trade']
    params['ohlcv']
    ...
```
the following compatible with backtesting:
```
@need_orderbook
def check_buy_timeout(self, pair: str, params: Dict) -> bool:
    params['ohlcv']
    ...    
```
-- same, plus still can be checked for compatibility with backtesting and resolves overhead issue.

d)
```
def check_buy_timeout(self, pair: str) -> bool:
    self.orderbook
    self.trade
    self.ohlcv
    etc
```
or via dataprovider:
```
def check_buy_timeout(self, pair: str) -> bool:
    self.dp.orderbook
    etc
```
-- cannot be checked for compatibility with backtesting; requires all those ugly checks we have in the strategy methods using dataprovider (`if self.dp.orderbook` ... else ...)
Pro: simple signature which does not require changes.

e) similar to d), but with decorators (that make corresponding `self.` properties available to the user code:
```
@need_orderbook
@need_trade
@need_ohlcv
def check_buy_timeout(self, pair: str) -> bool:
    self.orderbook
    self.trade
    self.ohlcv
    etc
```
this one can be used during backtesting:
```
@need_ohlcv
def check_buy_timeout(self, pair: str) -> bool:
    self.ohlcv
```
no backtesting, and raises exception:
```
@need_dataprovider
def check_buy_timeout(self, pair: str) -> bool:
    self.dp.orderbook
    self.ohlcv # this should cause exception since ohlcv is not requested by the decorator
```

yazeed:
> ## Summary
> 
> Inspired by the discussions in #2857 - this is a first (very rough) draft of how a strategy interface to customize buy_timeouts could look like.
> 
> It's intentionally very simple - it returns False by default (so no impact to users not interrested in this) - however when a strategy implements this (overrides this function) - a user can suddenly very flexibly control how and when to cancel orders.
> 
> This may accidentally (on purpose?) also fix #2866 - as the exception message is changed and now clearly points to "strategy code caused this exception".
> ### Example:
> 
>     * Cancel order after 3 minutes for pairs with value > 100 USDT
> 
>     * cancel order after 20 minutes for pairs with value > 10USDT
> 
>     * cancel order after 24h if value < 10USDT (keep order there).
> 
> 
> ```python
> def check_buy_timeout(self, pair: str, trade: Trade, order: Dict, **kwargs) -> bool:
>     if trade.open_rate > 100 and trade.open_date < datetime.utcnow() - timedelta(minutes=5):
>         return True
>     elif trade.open_rate > 10 and trade.open_date < datetime.utcnow() - timedelta(minutes=3):
>         return True
>     elif trade.open_rate < 1 and trade.open_date < datetime.utcnow() - timedelta(hours=24):
>        return True
>     return False
> ```
> 
> ## Quick changelog
> 
>     * Introduce generic wrapper for strategy functions (catching all errors users may cause)
> 
>     * introduce `check_buy_timeout()` - which can be used to customize buy order timeouts
> 
> 
> ## TODO
> 
>     * [ ]  Create same for sell side
> 
>     * [ ]  create tests
> 
>     * [ ]  Test it actually works as expected
> 
>     * [ ]  create documentation
>       
>       * [ ]  with a few "common" samples

I like the way you've introduced this, it gives enough control, and should be very easy for users to implement in their strategies.

Have you also considered if the concept of giving control to users to set their own buy invalidation criteria fits within this interface as well?

For example, I would want to use sell signals to invalidate buy signals in all of my strategies, overriding timeouts (think of it as a way to provide an override scenario within the interface).

Another invalidation rule I might consider experimenting with in addition to sell signals being produced, is looking at orderbooks delta (some kind of method that calculates delta based off a user specified # of orderbook rows) and see if it weighs in heavily one side, after the order is placed. Currently, this is only being checked at order placement stage, so for consistency reasons, I think we might want to consider having freqtrade review that delta at least.

xmatthias:
> as for 2), kwargs are a kind of code smell (a bit); 

Absolutely not if you consider how this is intended to be used. A user puts this into his strategy (it's working as expected, everything is working fine). We update this callback adding new, additional parameters (this strategy won't need). 
Without the kwargs it'll break ALL strategies implementing this method. 
I'm not proposing to have users use anything within `**kwargs` - but by just keeping it there will avoid that your strategy will raise an error when this method is called.

The only alternative to that is to use a dict in the first place (only one paramter passed to the function) - but then the users access pattern becomes  `metadata['pair']` - `metadata['trade']` ... which i find rather cumbersome.

Yes, new parametes will need to be documented. But a user only needs to update the signature if one of the new parameters is actually used.

I'll be honest - i regred not having thought about `**kwargs` with the `populate_*` interface. 
Now users will need to use `metadata['pair']` - which to you seems the cleaner method - but for an inexperienced user it's really stupid to use - and we've had MANY questions about this in the past as for a non-developer, this is not obvious.

---
You're talking about backtesting compatibility - but that's not relevant for order timeouts.  - this method will not be called during backtesting - so i see no point to waste time on finding / designing a solution in this PR for something that maybe, eventually, sometime in the future may be used (or maybe not). 

Backtesting always assumes all orders are filled immediately - and i see no way or reason to change this - so this is overcomplicating a simple interface without reason.

This may be needed for other callbacks - but not this one - so any solution we design now for a problem we don't have now will eventually be wrong - because when it comes to implementing that "other" callback - we'll find something we didn't think about causing this method not to work - which however you won't know as long as it's not relevant.

If a user needs orderbook / ohlcv / whatever to check this - `self.dp.**` will be available - so it's up to the user to decide what to use / what to query. Yes, we must provide samples of this.

-----

I thought about decorators ... but i think for a non-python user they are very hard to grasp.
One problem is that decorators need to be imported before usage (unless they're builtin).
Now while this is not necessarily a problem (@self.need_orderbook) - using `@need_orderbook` and assigning this to `self.orderbook` has another (maybe future) problem i have with everything where we'd be assigning `self.xxxx` pair-specific data without pair reference.

If any - i'd have orderbook as a parameter which is None in case the decorator is missing  - or assigned if the decorator is there.
However - from a design perspective i think we'll need to be carefull with this - `@need_orderbook` won't tell me if you need 5 elements of the orderbook, or 1000 - so we need to use the "maximum".

personally - i prefer to use `self.dp` in this case - it'll give users the same possibility they have in the `populate_*()` methods - (specifying how much of the orderbook is actually needed).

---

@yazeed s mentioned above - this is a POC / sample - i'm very well aware this is not complete and is missing pieces (that's why it's not implemented in any sample strategy - for example).
I draw this up quickly to explain what i intended in #2857 - either `orderbook()` or `fetch_ticker()` will be needed.

On the other hand - `self.dp` is available here - so orderbook and ohlcv is available (ticker is not after checking - but it's got a huge "implement me" TODO written against it in dataprovider.

Personally, i think the best way here is to provide "what's absolutely necessary" via the interface - and have everything else documented "copy/paste ready".
What this means is - getting the current signal (buy / sell) is 2 calls away. 
freqtrade not do that for the user - but have the user do that himself if it's needed. 

If you only use the very simple example listed above - buy/sell signal is not needed - but that doesn't invalidate my logic to close the order...

yazeed:
Sounds good, agree to providing what's absolutely necessary via the interface and the rest in docs. Would any of this have any impact on backtesting/hyperopt?

Either way, would love to see this PR through.

xmatthias:
> Would any of this have any impact on backtesting/hyperopt?

no - backtesting assumes all orders are filled immediately - and hyperopt just runs backtesting multiple times (so i tend to ignore that in these discussions as everything applying to backtesting automatically applies to hyperopt too).

> Either way, would love to see this PR through.

as said above - there's a few (ton) of things missing ... i'll work on it when i have time - but it'll for sure take a bit as i don't see this as immediately important. 
However, if this is an important topic for you, apreciate your input / commits to this branch (you should be able to push to this branch ...).
From a WoW point of view - let's not rebase it but in case of conflicts merge develop (rebase is odd to use if someone else is using that branch too)

yazeed:
Nice, glad to see this progress further.

xmatthias:
I got a question i'm not certain - should this be added to the (default) strategy template? 
If no - should i create an "advanced" template which includes these (optional) methods so users can get started easily?

yazeed:
> I got a question i'm not certain - should this be added to the (default) strategy template?
> If no - should i create an "advanced" template which includes these (optional) methods so users can get started easily?

I think it’s advanced, would be good with new-strategy --advanced. It’s a level of complexity/flexibility most people may not want to mess with.

yazeed:
Will post my feedback on this soon.

docs/strategy-advanced.md
```diff
@@ -0,0 +1,92 @@
...
+    unfilledtimeout = {
+        'buy': 60 * 25,
+        'sell': 60 * 25
+    }
+
+    def check_buy_timeout(self, pair: str, trade: Trade, order: dict, **kwargs) -> bool:
+        ob = self.dp.orderbook(pair, 1)
+        current_price = ob['bids'][0][0]
+        # Cancel buy order if price is more than 2% above the order.
+        if order['price'] > current_price * 1.02:
```
hroff-1902, line 80: should be reversed? (to be "Cancel buy order if price is more than 2% above the order" as stated in the comment)
```
if current_price > order['price'] * 1.02:
```
???

(same for sell part)

xmatthias: no ... it's correct this way (i got this wrong initially, too).

Assume i put my buy-order in at 100 USDT (using bids-side).
* price immediately jumps to 101 USDT (my trade is now in 15th spot in the orderbook ).
* Price jumps to 100 USDT - so my chances of beeing filled at 100 USDT are pretty slim -> i decide to cancel the order
therefore the calculation / comparsion is correct since it's: `102 > 100 * 1.02`.

If it where different - i place the buy-order at 100 - and price drops to 98 - then i'd have been filled already fully ... therefore that's no reason to cancel.

(the same applies to sell side - but obviously in the other direction.

hroff-1902: but 102 here is `current_price`, and `order['price']` is 100, so it should be
```
if current_price > order['price'] * 1.02:
```
to be 
```
102 > 100 * 1.02
```

PR #2875, opened 2020-02-07
Add draws column to backtesting tables

yazeed:
Added draws to backtesting tables, reduced length of some labels to help fit this without increasing total width.

Also, changes one of the win counts that used >= 0, to be > 0.

Updated docs accordingly...


yazeed:
Straightforward-est PR ever!

freqtrade/optimize/optimize_reports.py
```diff
@@ -21,13 +21,14 @@ def generate_text_table(data: Dict[str, Dict], stake_currency: str, max_open_tra
     tabular_data = []
     headers = [
         'Pair',
-        'Buy Count',
+        'Buys',
```
xmatthias: can we call this Trades? 

i think that would fit better - no?

yazeed: Buys are more specific to it and make it easier to understand I think, because there's another table below it that shows Sells and Sell reasons.

hroff-1902: Sells should probably become 'Closed Trades' or something like this when we'll implement trading shorts 😉

xmatthias: mh good point - if we consider longs and shorts, then "buys" is better as we can have "sells / Shorts" too (not now - we don't have that yet).

PR #2856, opened 2020-02-04
Sortino HyperOptLoss method

yazeed:
This is an incomplete PR that prepares for a Sortino HyperOpt Loss Method as mentioned in #2407, which seems to be able to come up with strategies with very low drawdowns (0.02 to 0.05), specially if you enable trailing stop loss.

**Work to be done:**
- [x] - SortinoHyperOptLoss: To be consistent with SharpeHyperOptLoss.
- [x] - SortinoHyperOptLossDaily: To be consistent with SharpeHyperOptLossDaily.
- [x] - Unit tests for both new hyperopts.

References:
- [Wikipedia: Sortino](https://en.wikipedia.org/wiki/Sortino_ratio)
- [Investopedia: Sortino](https://www.investopedia.com/terms/s/sortinoratio.asp)
- [Xplained: Sortino](https://xplaind.com/262577/sortino-ratio)
- [ProgrammingForFinance: Sortino](https://programmingforfinance.com/2017/11/examining-the-sortino-ratio/)

Issue #2407, opened 2019-10-21
Hyperopt / Sharpe Ratio v.s. Sortino


Prossi79:
I appreciate the option to hyperopt towards the Sharpe Ratio, what about Sortino? I recently read following intersting comment about both ratios:

"The Sortino ratio is motivated by the fact that the Sharpe ratio captures both upward and
downward volatility in its denominator. However, investors (and hedge fund managers) are generally not too bothered when we have signifcant upward volatility!
What is actually of interest from a risk management perspective is downward volatility and periods of drawdown."

Could the Sortino be implemented in the same way as the Sharpe as hyperopt option?

xmatthias:
Depends what data Sortino uses ... but the anser is probably yes.
Have you got a good link as to how the formula for sortino is?

hroff-1902:
@xmatthias look here, all sharpe, calmar and sortino are one-liners from returns:
https://github.com/kernc/backtesting.py/blob/a46d72a635faf06a50ad47409ebf6fa5fde81c04/backtesting/backtesting.py#L921-L923

This implementation does not do annualization of the coefficients, but I like the approach they calculated...

hroff-1902:
Implementation from pyfolio: 
https://github.com/quantopian/pyfolio/blob/c5ebc0020f605b024edffcdcf452aca34bee8bf8/pyfolio/timeseries.py#L202

xmatthias:
Empyrical does not have any "odd" dependencies apart from pandas-datareader which i think does not work correctly anymore - so that'll be fine to include.

Pyfolio on the other hand will not give us any benefit in this case ... since it really calls empyrical and has these ratios deprecated ... and is more usefull in interactive analysis setups.

hroff-1902:
And even just the 'stats.py' module from empyrical. We don't need the rest.

hroff-1902:
Another implementation from ranaroussi's quantstat (we also need only this module):
https://github.com/ranaroussi/quantstats/blob/master/quantstats/stats.py

xmatthias:
no, i wouldn't start copy/pasting stuff. I ... maintenance of that is a huge pain (we'll always be out of date). 

There is a reasons for things like pypi ... 

The only reason for doing that is if the rest of the module has strange dependencies we absolutely don't want (which is not something i see with empyrical) or if the library is not maintained (does not mean no commits, but no responses to issues) - otherwise it's way easier to include this as a dependency and leave maintenance of this to the main library.

[...]

PR #2879, opened 2020-02-07
Sortino hyperopt loss

hroff-1902:
cherry-picked carefully from #2856, to cleanup history

### Original PR description
Author: @yazeed 

This is an incomplete PR that prepares for a Sortino HyperOpt Loss Method as mentioned in #2407, which seems to be able to come up with strategies with very low drawdowns (0.02 to 0.05), specially if you enable trailing stop loss.

**Work to be done:**
- [x] SortinoHyperOptLoss: To be consistent with SharpeHyperOptLoss.
- [x] SortinoHyperOptLossDaily: To be consistent with SharpeHyperOptLossDaily.
- [x] Unit tests for both new hyperopts.

References:
- [Wikipedia: Sortino](https://en.wikipedia.org/wiki/Sortino_ratio)
- [Investopedia: Sortino](https://www.investopedia.com/terms/s/sortinoratio.asp)
- [Xplained: Sortino](https://xplaind.com/262577/sortino-ratio)
- [ProgrammingForFinance: Sortino](https://programmingforfinance.com/2017/11/examining-the-sortino-ratio/)


yazeed:
Thanks for helping with this, not sure what happened on the PR, I think I rebased off develop, and that's why it brought all those commits into it. Anyways, I just remembered we need to mention these Sortino methods in docs, also.

hroff-1902:
since I cannot be a reviewer for this PR, pls wait till my explicit approvement in the test.

xmatthias:
>  pls wait till my explicit approvement in the test.

well i'm still waiting ... ;) unless i missed something ...

yazeed:
Just a note, this needs a similar date_range normalize fix like the one applied in #2891 .

yazeed:
@hroff-1902 What's pending here? :)

hroff-1902:
Nothing but spare time and the following point:

I'm thinking, which is more correct for Sortino downside calculation for the case of Sortino Daily, where returns are aggregated instead of 'legacy' Sortino:

* either select all negative returns from backtesting_results (*), aggregate them on daily returns (**) and get stdev of it,
* or aggregate returns from backtesting_results (*), then select all negative ones of them (**) and get stddev of it...


(*) well, actually returns from 'profit_percent_after_slippage', after subtracting slippage, as we implemented here and in Sharpe daily.
(**) filling the gaps and subtracting daily free_risk_rate, as we implemented here and in Sharpe daily.

xmatthias:
I think you'll need to subtract from the raw backtesting results to get a "valid" downside.

If you use negative days (after resample) you may have no downside in certain strategy/timeframe combinations, which however would not mean that there is no downside, just that we were unable to calculate that.

hroff-1902:
I'm rather of the opposite opinion: 

If you look at the days when this metrics was created, people calculated the metrics manually, using monthly/annual returns. If you take a look at the last article stated above -- it contains an example of excel spreadsheet, operating on daily or monthly returns.

In either of the above cases, people simply take their daily/monthly/annual cumulative returns, and those returns are already aggregated -- each daily/monthly/annual return value is a sum of all trade returns for the period, people do not select only positive ones before entering data into a spreadsheet, table for manual calculation or a program/function on python/R/other language, if it does not contain aggregation of returns (as many of them implemented).

These are the conditions in which this metrics was created and in which it's supposed to run by the creator, as I can conclude.

So we should follow same schema here -- first aggregate particular trade returns on the daily basis and then perform all calculations, specific for this metrics, including selection of downside ones.

@yazeed I could make and commit my variant, employing the described behavior, that would probably be easier way...

yazeed:
> @yazeed I could make and commit my variant, employing the described behavior, that would probably be easier way...

Sure, please go ahead. :)

hroff-1902:
still looking at it, pls be patient (there is a feature of this Sortino calculation I do not like, need to explore it in details, will report)

yazeed:
> still looking at it, pls be patient (there is a feature of this Sortino calculation I do not like, need to explore it in details, will report)

Sure, no worries, would love to hear more about where you're at.

hroff-1902:
@xmatthias @yazeed pls review, I used http://www.redrockcapital.com/Sortino__A__Sharper__Ratio_Red_Rock_Capital.pdf
as the reference description (this linked from Wikipedia article above)

Annotations for the lines of code/my commits will follow.

hroff-1902:
tested it as 
```
freqtrade hyperopt -c config_binance.json.example --hyperopt SampleHyperOpt --strategy SampleStrategy --print-all -i 1h -e 500 --min-trades 20 --random-state 42 --hyperopt-loss SortinoHyperOptLossDaily
```
and with extra prints (which are commented out), to see the values:
```
freqtrade hyperopt -c config_binance.json.example --hyperopt SampleHyperOpt --strategy SampleStrategy --print-all -i 1h -e 1 --min-trades 20 --random-state 42 --hyperopt-loss SortinoHyperOptLossDaily -j 1
```

yazeed:
I like the fact that we have both options so that myself and the community can test both, as I personally would like to test and study variances in backtesting results between both hyperopt loss methods, as I already witnessed pros and cons for each.

While this looks good to me as far as I can tell. I can work on documentation tomorrow if you guys as @xmatthias indicated, so we can consider this PR complete.

orehunt:
I don't think the unsampled sortino is any useful at all compared to sharpe. It is much more important for sortino to sample because otherwise pumps will skew the results, which means that if you optimize with it, it will descend into parameters that give low deviation big losses with far and few big gains, while sampling with sharpe is useful for the opposite reason, that is to denoise the results by ignoring short term volatility.

hroff-1902:
for non-daily sortino -- I do not know how to make apriori incorrect metrics 
more correct'...
but you can adopt same calculations to that as well

hroff-1902:
Some comments on my commits:

* I only changed the name of the risk_free_ratio to be minimum acceptable return, the names of other variables / columns were not altered.

* The following:
```
        sum_daily['downside_returns'] = 0
        sum_daily.loc[total_profit < 0, 'downside_returns'] = total_profit
```
calculates `f = min(0, Xi - T)` (see the reference desctiption), but without the usage of `min()`.

* Then the equation for TDD can be rewritten as `sqrt((1/N)*sum(f - 0.0)^2)`, i.e. as the second moment against the 0.0 point. The `pandas.stdev()` and `np.stdev()` do only calculate stdev against the mean point, while `statistics.pstdev()` and `statistics.stdev()` allow to specify the point against which the moment is calculated, in the second parameter.

* We have a population of returns, not a simple random sample of returns taken from a wider population, so `statistics.pstdev()` (not `statistics.stdev()`) is used, since it's more correct. They only differ in the use of denominator of `1/N` or `1/(N-1)`. Accordingly (not in this PR), calculations for Sharpe should also be adjusted to use `ddof=0`.

yazeed:
Added docs updates.

docs/hyperopt.md
```diff
@@ -191,8 +191,10 @@ Currently, the following loss functions are builtin:
 
 * `DefaultHyperOptLoss` (default legacy Freqtrade hyperoptimization loss function)
 * `OnlyProfitHyperOptLoss` (which takes only amount of profit into consideration)
-* `SharpeHyperOptLoss` (optimizes Sharpe Ratio calculated on the trade returns)
-* `SharpeHyperOptLossDaily` (optimizes Sharpe Ratio calculated on daily trade returns)
+* `SharpeHyperOptLoss` (optimizes Sharpe Ratio calculated on trade returns relative to **upside** standard deviation)
+* `SharpeHyperOptLossDaily` (optimizes Sharpe Ratio calculated on **daily** trade returns relative to **upside** standard deviation)
```
xmatthias: Are you sure about this? 
i mean - the variable is called `up_stdev` - but there seems to be no filter removing the downside ...

freqtrade/optimize/hyperopt_loss_sortino_daily.py
```diff
@@ -0,0 +1,69 @@
...
+                {"profit_percent_after_slippage": sum}).reindex(t_index).fillna(0)
+        )
+
+        total_profit = sum_daily["profit_percent_after_slippage"] - minimum_acceptable_return
+        expected_returns_mean = total_profit.mean()
+
+        sum_daily['downside_returns'] = 0
+        sum_daily.loc[total_profit < 0, 'downside_returns'] = total_profit
+        total_downside = sum_daily['downside_returns']
+        down_stdev = statistics.pstdev(total_downside, 0)
```
orehunt:
```suggestion
        down_stdev = statistics.pstdev(total_downside, minimum_acceptable_return)
```

freqtrade/optimize/hyperopt_loss_sortino_daily.py
```diff
@@ -0,0 +1,69 @@
...
+        # create the index within the min_date and end max_date
+        t_index = date_range(start=min_date, end=max_date, freq=resample_freq,
+                             normalize=True)
+
+        sum_daily = (
+            results.resample(resample_freq, on='close_time').agg(
+                {"profit_percent_after_slippage": sum}).reindex(t_index).fillna(0)
+        )
+
+        total_profit = sum_daily["profit_percent_after_slippage"] - minimum_acceptable_return
```
orehunt:
```suggestion
        total_profit = sum_daily["profit_percent_after_slippage"]
        expected_returns_mean = total_profit.mean() - minimum_acceptable_return
```

freqtrade/optimize/hyperopt_loss_sortino_daily.py
```diff
@@ -0,0 +1,69 @@
...
+        expected_returns_mean = total_profit.mean()
+
+        sum_daily['downside_returns'] = 0
+        sum_daily.loc[total_profit < 0, 'downside_returns'] = total_profit
```
orehunt:
```suggestion
        sum_daily.loc[total_profit < minimum_acceptable_return, 'downside_returns'] = total_profit
```

freqtrade/optimize/hyperopt_loss_sortino_daily.py
orehunt:
```suggestion
        down_stdev = statistics.pstdev(total_downside, minimum_acceptable_return)
```

docs/hyperopt.md
yazeed: It just describes how the formula works, "relative to" here means divided by, upside standard deviation.

freqtrade/optimize/hyperopt_loss_sortino_daily.py
hroff-1902: let me think about proposed changes...

docs/hyperopt.md
xmatthias: I understand that - but it's not the "upside standard deviation".

from another site:

> It is defined as the difference between the returns of the investment and the risk-free return, divided by the standard deviation of the investment.

it's the standard-deviation of the whole investment - not just of the upside / profits.

freqtrade/optimize/hyperopt_loss_sortino_daily.py
untoreh: if you don't make these changes, when the MAR is not 0, the calculation is wrong because if you apply the std dev function with the MAR as mean unit then you have to normalize the positive returns to the MAR, otherwise the std dev function will subtract the MAR to the points nulled before.
- either you ignore the MAR and just use 0 everywhere or normalize as suggested or calculate the downside deviation without the pstdev function

hroff-1902: I used this article as the reference for implementation:
http://www.redrockcapital.com/Sortino__A__Sharper__Ratio_Red_Rock_Capital.pdf

* The formula in wikipedia is a continuous one (with a random var and an integral) so it's useless for practical calculations;
* The formulas and examples of calculations in programmingforfinance, investopedia and xplaind (see at the top) are simplified;
* So this article seems to be the most "scientific-like".

If you take a look at the article, it's not the stdev at the denominator, because stdev is calculated as the second moment (well, sqrt of) against the mean, while the formula in the article is actually is a moment against the zero point (since MAR was already subtracted under the min() function). Please review my explainations/annotations above: https://github.com/freqtrade/freqtrade/pull/2879#issuecomment-587468323

hroff-1902: But anyway, I'll also review this once again...

untoreh: I suggested the modifications based on that article that you took as reference not on the formula with the integral

docs/hyperopt.md
yazeed: Oh ya, I got your point, you're absolutely right, I'll correct it.

freqtrade/optimize/hyperopt_loss_sortino_daily.py
yazeed: @hroff-1902 I actually think orehunt's suggestions make sense, lets commit them and I'll test them and compare with the previous version using a randomstate I've run before.

hroff-1902:
reviewing the proposed changes:

Let's use `P` for `sum_daily["profit_percent_after_slippage"]`, `MAR` for `minimum_acceptable_return` in formulas to be short.

1st part of changes:

old:
```
total_profit = sum_daily["profit_percent_after_slippage"] - minimum_acceptable_return
expected_returns_mean = total_profit.mean()
```
new:
```
total_profit = sum_daily["profit_percent_after_slippage"]
expected_returns_mean = total_profit.mean() - minimum_acceptable_return
```

-- this does not change the mean (`expected_returns_mean`), because `mean(Xi- C) = mean(Xi) - C`, where `C` is a constant (`minimum_acceptable_return` in our case)
-- the only change here is the following: 
old:
```
total_profit = P - MAR
```
new:
```
total_profit = P
```

hroff-1902:
2nd part of proposed changes:

old:
```
sum_daily['downside_returns'] = 0
sum_daily.loc[total_profit < 0, 'downside_returns'] = total_profit
```
new:
```
sum_daily['downside_returns'] = 0
sum_daily.loc[total_profit < minimum_acceptable_return, 'downside_returns'] = total_profit
sum_daily.loc[total_profit > minimum_acceptable_return, 'downside_returns'] = minimum_acceptable_return
```

-- first, it's a bit incorrect, the '>=' ('or <=') sign should be used in one of the assignments, because otherwise it will leave 0 values for rows where `total_profit == minimum_acceptable_return`. But this is a minor note and is not actually an issue.

(TBC)

hroff-1902:
3rd part of the code:

-- old:
```
down_stdev = statistics.pstdev(total_downside, 0)
```
-- new:
```
down_stdev = statistics.pstdev(total_downside, minimum_acceptable_return)
```

`pstdev` here is `pstdev(Xi, a) = (1/N)*sqrt(sum(Xi - a)^2)` -- sqrt of so called "second moment" against float point `a`. 'Usual stdev(Xi) is pstdev(Xi, mean(Xi))', we use a point which differs from `mean(Xi)`, so `statistics.pstdev` is used, which allows to specify the point, against which the moment is calculated.

(TBC)

hroff-1902:
@orehunt the changes you may have found may be caused by that error, where 0 values were left in `sum_daily['downside_returns']` for trades where `total_profit == minimum_acceptable_return` ('>=' or '<=' should be used). I implemented both versions (my one and yours, but corrected one), tested on a few examples, and they actually give same values, for both MAR=0 and MAR=non zero... 

If you wish (and @yazeed and/or @xmatthias, if you guys want to verify this in code), I could place both variants in git gists, for example, with commands to run and verify the results...

hroff-1902:
(Now you see why I delayed (still a bit) with review... 🤣 Sorry guys for that, but I needed to find a big enough slot of time for this work, both math. formulas and implementation in the code/testing and some inspiration to complete it in one step... 😉 )

orehunt:
@hroff-1902 I made to changes in an attempt to make the stdev function behave like sortino when MAR is not 0, but it appears the `mu` in `statistics` p/stdev is not useful for the sortino calculation. What it does is calculate the p/variance against the given point, which results generally in a very small difference from the actual variance (assuming numbers of the same order...)
Which means that the calculation for sortino should be done without the use of any standard deviation function.

```python
import numpy as np
import statistics as st
import json

MAR = 0.1
N = 365
P = np.random.uniform(-1, 1, N)
stdev = {}

R = P.sum()
P_min = np.array([min(0, pm - MAR) for pm in P])
mean = P_min.mean()
assert (N == len(P_min))
stdev['statistics_pstdev_0'] = st.pstdev(P_min, 0)
stdev['statistics_pstdev_mar'] = st.pstdev(P_min, MAR)
stdev['statistics_pstdev_mean'] = st.pstdev(P_min, mean)
stdev['statistics_pstdev_default'] = st.pstdev(P_min)
# assert stdev['statistics_pstdev_mean'] == stdev['statistics_pstdev_default']

stdev['numpy'] = P_min.std(ddof=0)
stdev['raw_mean'] = np.sqrt(((P_min - mean)**2).sum() / N)
stdev['raw_sortino'] = np.sqrt(((P_min - 0)**2).sum() / N)
stdev['raw_sortino_CORRECT'] = np.sqrt(((P_min)**2).sum() / N)

if stdev['statistics_pstdev_mar'] != stdev['statistics_pstdev_0']: print(f'MAR is {MAR}')
print(json.dumps(stdev, indent=1))
sortino = (R-MAR) / stdev['raw_sortino_CORRECT']
wrong_sortino = (R-MAR) / stdev['statistics_pstdev_0']
print(f'wrong sortino is: {wrong_sortino}')
print(f'correct sortino is: {sortino}')
```

orehunt:
@hroff-1902 I made to changes in an attempt to make the stdev function behave like sortino when MAR is not 0, but it appears the mu in statistics p/stdev is not useful for the sortino calculation. What it does is calculate the p/variance against the given point, which results generally in a very small difference from the actual variance (assuming numbers of the same order...)
Which means that the calculation for sortino should be done without the use of any standard deviation function.
```python
import numpy as np
import statistics as st
import json

MAR = 0.1
N = 365
P = np.random.uniform(-1, 1, N)
stdev = {}

R = P.sum()
P_min = np.array([min(0, pm - MAR) for pm in P])
mean = P_min.mean()
assert (N == len(P_min))
stdev['statistics_pstdev_0'] = st.pstdev(P_min, 0)
stdev['statistics_pstdev_mar'] = st.pstdev(P_min, MAR)
stdev['statistics_pstdev_mean'] = st.pstdev(P_min, mean)
stdev['statistics_pstdev_default'] = st.pstdev(P_min)
# assert stdev['statistics_pstdev_mean'] == stdev['statistics_pstdev_default']

stdev['numpy'] = P_min.std(ddof=0)
stdev['raw_mean'] = np.sqrt(((P_min - mean)**2).sum() / N)
stdev['raw_sortino'] = np.sqrt(((P_min - 0)**2).sum() / N)
stdev['raw_sortino_CORRECT'] = np.sqrt(((P_min)**2).sum() / N)

if stdev['statistics_pstdev_mar'] != stdev['statistics_pstdev_0']: print(f'MAR is {MAR}')
print(json.dumps(stdev, indent=1))
sortino = (R-MAR) / stdev['raw_sortino_CORRECT']
wrong_sortino = (R-MAR) / stdev['statistics_pstdev_0']
print(f'wrong sortino is: {wrong_sortino}')
print(f'correct sortino is: {sortino}')
```

You can ignore the suggestions I made since you can't use the functions anyway, well maybe except the one that if you subtract the MAR from the returns AFTER the mean, you spare some calculations...but it is a minor thing.
edit: I deleted this comment at first because I was unsure then I re-re-read >.<

hroff-1902:
> You can ignore the suggestions I made since you can't use the functions anyway

-- sure, since your (those) suggestions produce absolutely same values (after fixing the bug), as I shown above 😉

orehunt:
> > You can ignore the suggestions I made since you can't use the functions anyway
> 
> -- sure, since your (those) suggestions produce absolutely same values (after fixing the bug), as I shown above wink
yes you would have to initialize the `downside_returns` series directly to the mar instead of 0 and then update the negative returns

hroff-1902:
Have had a look at the sources of pstdev, as a reference:

https://github.com/python/cpython/blob/02a4d57263a9846de35b0db12763ff9e7326f62c/Lib/statistics.py#L670-L672

```
# Under no circumstances use the so-called "computational formula for
# variance", as that is only suitable for hand calculations with a small
# amount of low-precision data. It has terrible numeric properties.
```
😉

orehunt:
not sure how that is relevant

orehunt:
I have modified the [test file](https://github.com/freqtrade/freqtrade/pull/2879#issuecomment-591910032) with the variable from the examples in the article
>
> In Step 1 above, the difference with respect to the target level is calculated, unlike the standard deviation calculation where the difference is calculated with respect to the mean of all data points. If every data point equals the mean, then the standard deviation is zero, no matter what the mean is. Con-sider the following return stream: [-10, -10, -10, -10]. The standard deviation is 0, while the target downside deviation is **10** (assum-ing target return is 0)

hroff-1902:
(to be logged here) 
found a nice article, with overview of the approaches and more metrics measuring downside returns explained: http://ceejme.eu/wp-content/uploads/2016/11/ceejme_2_6_art_01.pdf
"Diana CIBULSKIENÉ, Martynas BRAZAUSKAS. A theoretical approach to quantitative downside risk measurement methods."

hroff-1902:
Now it seems completely clear (for me):

I was wrong treating `pstdev(Xi, a)` calculated in the statistics module as sqrt(sum((Xi - a)**2)/N) i.e. using a instead of mean(Xi) in the calculations. Actually, `pstdev(Xi, a) <=> stdev(Xi-a)` (where stdev is calculated against or around mean value), and stdev is invariant against such kind of transformation of parameter, that's why all ^^ values above are the same with only small difference cause rounding errors... So `a` (`mu` parameter) is only used to avoid errors in some cases and is not actually an important parameter. It's not a point "against which" the deviations of values are taken...

So, since in our case `a=0`, then we should definitely calculate downside deviation simply as 
```
down_stdev = np.sqrt((total_downside**2).sum() / N)
```
or similar...

@orehunt Am I correct?

hroff-1902:
Applied the final approach with sqrt, going to merge it soon...

Issue #2874, opened 2020-02-07
/status table not showing complete information


yazeed:
When using /status table with the RPC telegram bot, it returns existing trades (bought pairs) mixed with open buy orders (pending orders) without any way to differentiate between them.

This is not truthful to the purpose of the function, in the sense that it should convey the difference between these two types, since it also shows Profit, which is false for buy orders that haven't been filled.

This is to log that issue, until a PR I plan to work on is introduced.

We can also discuss how to solve this.

I'm thinking we can add an asterisk (*) next to pairs that aren't owned yet, and to show 0 profit for those. In the sqlite database these can be differentiated by open_order_id.

We can also expand this potential enhancement to show open_bid for those open orders, and open_rate for existing pairs, and current_bid.

Would love to hear your thoughts.

I'm focused at the moment on consistency to polish freqtrade in such areas in small iterations.

xmatthias:
Well yeah - but there's another part to this too ...
while `/status` shows you this trade while the order is still open - the notifications in case of a cancel are not consistent.

* A notification is sent on buy
* a notification is sent on sell
* **NO** notification is sent on buy cancel

So if you rely on these notifications alone - you think the bot is in a position while it really cancelled the order.
Now *most orders (on higher volume pairs) fill almost instantly ... so it's not a huge issue - but it's still something i think somehow fits to this issue.

#2646 hints to something similar - but from the other direction - that a open buy order cannot be cancelled via RPC.

Issue #2502, opened 2019-11-09
Optimized Telegram Output 🤖


Escaliert:
**For all those guys who wants a better visual Telegram Output. Specially on Smartphones its very useful for a quick look into nofication-list.
Its possible to format the telegram Output in the "telegram.py" file like this...**

![image](https://user-images.githubusercontent.com/39638559/68532532-9449fb00-031e-11ea-86f9-306d8a67c581.png)
![Telegram_G3JmQx0SgZ](https://user-images.githubusercontent.com/39638559/68532396-ff92cd80-031c-11ea-82e1-3fe6042ce7ac.png)
![image](https://user-images.githubusercontent.com/39638559/68532398-09b4cc00-031d-11ea-8353-ccedabf4ca63.png)

**you only have to set your emotes from https://emojipedia.org/ in the "telegram.py"-file like this**

![image](https://user-images.githubusercontent.com/39638559/68532442-74fe9e00-031d-11ea-87be-c6c55cab537a.png)

**thats it 👍**

**_btw here is my example file_** 

[telegram.zip](https://github.com/freqtrade/freqtrade/files/3827523/telegram.zip)

ENJOY 🤖

xmatthias:
Now, i personally would want to see sells in either Green or Red - depending on the outcome (positive / negative) ...
and buys in blue (or some neutral color ...).

Apart from that (which are visual details ...) - why is this an issue and not a PR, where you're proposing to change this so everyone can benefit? :smile:

Theagainmen:
I indeed would like something mathias said. I quickly made this, there are probably better ways to do this tho, but it works :)
``` python
        elif msg['type'] == RPCMessageType.SELL_NOTIFICATION:
            msg['amount'] = round(msg['amount'], 8)
            msg['profit_percent'] = round(msg['profit_percent'] * 100, 2)

            message = ("*{exchange}:* Selling {pair}\n"
                       "*Rate:* `{limit:.8f}`\n"
                       "*Amount:* `{amount:.8f}`\n"
                       "*Open Rate:* `{open_rate:.8f}`\n"
                       "*Current Rate:* `{current_rate:.8f}`\n"
                       "*Sell Reason:* `{sell_reason}`\n").format(**msg)

            print (float(msg['profit_percent']))

            if float(msg['profit_percent']) > 0:
                message += ('💲*Profit:* `{profit_percent:.2f}%`').format(**msg)

            elif float(msg['profit_percent']) > 5:
[...]

PR #2887, opened 2020-02-08
Wide RPC notifications fixes

yazeed:
This PR addresses issues mentions in #2874 and goes a bit beyond to bring consistency into RPC, Webhook and Telegram messages being sent, also brings two distinct types of messages for Buy and Sell order cancels instead of sending them as generic status messages.

With this, notifications are now being sent for both cancellations of both buys and sells.

It marks pending orders in /status table with an asterisk for clarity.

Finally, it also adds an argument **refresh: bool** to get_buy_rate to make it consistent with get_sell_rate usage.

To summarize:

- [x] Notifications for buy/sell cancellations now have their own message types.
- [x] More consistent console logging messages relating to buy/sell execution and cancellations.
- [x] More consistent RPC webhook and telegram messages relating to buy/sell execution and cancellations.
- [x] Buy/Sell cancellations now properly send out their own RPC messages.
- [x] /status table now shows an asterisk next to pending buy orders.
- [x] Minor refactors where delete_trade() is introduced as a result.
- [x] Tests added for buy/sell cancellation scenarios.

Fixes that could have been out of PR, but still fixed them as I noticed they cause RPC to send non-consistent price values in messages -- buys didn't use get_buy_refresh(refresh->TRUE), while sells already had get_sell_rate(refresh->TRUE):

- [x] get_buy_rate() method now has a refresh bool argument to match get_sell_rate().


yazeed:
Not sure why tests coverage decreased, not quite familiar with test writing.

hroff-1902:
@yazeed since you reviewed RPC messages and have them in your mind, their location in the code, etc, maybe you also incorporate this one here: https://github.com/freqtrade/freqtrade/issues/2502#issuecomment-552580197 ?

The list of chars can be found here: 
https://unicode.org/emoji/charts/full-emoji-list.html#other-symbol

The emoji unicode chars should better be places as unicode char codes. like `\U0001f604`, to keep the sources be ASCII.

However, I just found a neat package: https://pypi.org/project/emoji/

hroff-1902:
> Not sure why tests coverage decreased even though I've added tests for the new types, not quite familiar with test writing.

you can press at the "coverage 97%" button there and see all details on changes in coverage.

but decreasing it 0.2% for me is okay by itself. However, looking at the details can get you ideas on which is not covered what additional small tests can be added...

yazeed:
> @yazeed since you reviewed RPC messages and have them in your mind, their location in the code, etc, maybe you also would incorporate this one here: [#2502 (comment)](https://github.com/freqtrade/freqtrade/issues/2502#issuecomment-552580197) ?
> 
> or in a sep. PR, following this one...
> 
> The list of chars can be found here:
> https://unicode.org/emoji/charts/full-emoji-list.html#other-symbol
> 
> The emoji unicode chars should better be places as unicode char codes. like `\U0001f604`, to keep the sources be ASCII.
> 
> However, I just found a neat package: https://pypi.org/project/emoji/

Nice, I like the suggestions of using emojis there to make it easy to browse telegram messages. I can work on them in a separate PR, I think this one addresses many inconsistencies and is already big enough, so I'd prefer to keep its scope minimal so it can be easily read and understood.

yazeed:
> > Not sure why tests coverage decreased even though I've added tests for the new types, not quite familiar with test writing.
> 
> you can press at the "coverage 97%" button there and see all details on changes in coverage.
> 
> but decreasing it 0.2% for me is okay by itself. However, looking at the details can get you ideas on which is not covered what additional small tests can be added...

Oh, thanks for that. Just helped me figure this out, never thought about clicking that button. :)

Full coverage tests, added to this PR just now.

yazeed:
Failed that weird test again for some reason. :)

docs/telegram-usage.md
```diff
@@ -55,7 +55,7 @@ official commands. You can ask at any moment for help with `/help`.
 | `/reload_conf` | | Reloads the configuration file
 | `/show_config` | | Shows part of the current configuration with relevant settings to operation
 | `/status` | | Lists all open trades
-| `/status table` | | List all open trades in a table format
+| `/status table` | | List all open trades in a table format. Pending buy orders are marked with an asterisk(*)
```
hroff-1902:
```suggestion
| `/status table` | | List all open trades in a table format. Pending buy orders are marked with an asterisk (*)
```

freqtrade/freqtradebot.py
```diff
@@ -404,7 +404,7 @@ def create_trade(self, pair: str) -> bool:
 
             stake_amount = self.get_trade_stake_amount(pair)
             if not stake_amount:
-                logger.debug("Stake amount is 0, ignoring possible trade for {pair}.")
+                logger.debug(f"Stake amount is 0, ignoring possible trade for {pair}.")
```
hroff-1902: 👍 nice catch

freqtrade/freqtradebot.py
```diff
@@ -414,10 +414,12 @@ def create_trade(self, pair: str) -> bool:
             if ((bid_check_dom.get('enabled', False)) and
                     (bid_check_dom.get('bids_to_ask_delta', 0) > 0)):
                 if self._check_depth_of_market_buy(pair, bid_check_dom):
+                    logger.info(f'Executed Buy for {pair}.')
```
hroff-1902: if it's before `execute_buy`, then it probably should state "Executing..." (not yet executed, may be confusing along with debug messages from `execute_buy`)

freqtrade/freqtradebot.py
```diff
@@ -414,10 +414,12 @@ def create_trade(self, pair: str) -> bool:
...
                     return self.execute_buy(pair, stake_amount)
                 else:
                     return False
 
+            logger.info(f'Executed Buy for {pair}')
```
hroff-1902: same as previous

freqtrade/freqtradebot.py
```diff
@@ -450,15 +452,15 @@ def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = N
         """
         Executes a limit buy for the given pair
         :param pair: pair for which we want to create a LIMIT_BUY
-        :return: None
+        :return: bool
```
hroff-1902: 👍 nice catch

freqtrade/freqtradebot.py
```diff
@@ -1006,7 +1030,7 @@ def _notify_sell(self, trade: Trade, order_type: str) -> None:
             'exchange': trade.exchange.capitalize(),
             'pair': trade.pair,
             'gain': gain,
-            'limit': trade.close_rate_requested,
+            'limit': profit_rate,
```
hroff-1902: note to myself: verify this change separately

freqtrade/freqtradebot.py
```diff
@@ -547,11 +549,37 @@ def _notify_buy(self, trade: Trade, order_type: str) -> None:
             'type': RPCMessageType.BUY_NOTIFICATION,
             'exchange': self.exchange.name.capitalize(),
             'pair': trade.pair,
-            'limit': trade.open_rate,
+            'limit': trade.open_rate_requested,
```
hroff-1902: note to myself: verify this change separately

freqtrade/freqtradebot.py
```diff
@@ -751,7 +779,7 @@ def handle_trailing_stoploss_on_exchange(self, trade: Trade, order: dict) -> Non
             update_beat = self.strategy.order_types.get('stoploss_on_exchange_interval', 60)
             if (datetime.utcnow() - trade.stoploss_last_update).total_seconds() >= update_beat:
                 # cancelling the current stoploss on exchange first
-                logger.info('Trailing stoploss: cancelling current stoploss on exchange (id:{%s})'
+                logger.info('Trailing stoploss: cancelling current stoploss on exchange (id:{%s}) '
```
hroff-1902: 👍

freqtrade/freqtradebot.py
```diff
@@ -938,13 +960,13 @@ def _safe_sell_amount(self, pair: str, amount: float) -> float:
             raise DependencyException(
                 f"Not enough amount to sell. Trade-amount: {amount}, Wallet: {wallet_amount}")
 
-    def execute_sell(self, trade: Trade, limit: float, sell_reason: SellType) -> None:
+    def execute_sell(self, trade: Trade, limit: float, sell_reason: SellType) -> bool:
```
hroff-1902: I guess we were thinking of opposite: make `execute_buy()` returning None...
Anyway, I understand your possible rationaly here (to have it aligned with `execute_buy()`), so it's okay, this can be reworked later...

freqtrade/rpc/rpc.py
```diff
@@ -157,15 +160,16 @@ def _rpc_status_table(self, stake_currency: str,
...
+                        stake_currency,
+                        fiat_display_currency
+                    )
                     if fiat_profit and not isnan(fiat_profit):
                         profit_str += f" ({fiat_profit:.2f})"
                 trades_list.append([
                     trade.id,
-                    trade.pair,
+                    trade.pair + ['', '*'][trade.open_order_id is not None
+                                           and trade.close_rate_requested is None],
```
hroff-1902:
```
trade.pair + '*' if (trade.open_order_id is not None
                     and trade.close_rate_requested is None) else ''
```
-- proposing this can easier for the eyes...

freqtrade/freqtradebot.py
yazeed: Problem here is there's a return statement after this line, so I couldn't move it after.

yazeed: I think this function needs further refactoring in the future.

freqtrade/rpc/rpc.py
yazeed: I actually find the one I have used easier to read, but I can change it if this is a guideline in the codebase.

freqtrade/freqtradebot.py
yazeed: It's also a way to communicate the status of the execute buy/sell.

freqtrade/freqtradebot.py
xmatthias: if you wanna leave that here for now it's ok - but it's a time question ...

this is saying "executed" - but it's really just about to execute ... so if execute_buy logs anything (maybe an exception ..?) - the bot's prior log-message (line above the exception, most likely) is highly irritating in that case - as it just told me you did that already...

Think about it
```
Executed buy for pair ETH/BTC.
Exception while buying ETH/BTC, Error XYZ on exchange whatever.....
```

This makes the first line a lie of some sort.

freqtrade/freqtradebot.py
```diff
@@ -820,41 +848,41 @@ def check_handle_timedout(self) -> None:
...
                   or (self._check_timed_out('sell', order))):
                 self.handle_timedout_limit_sell(trade, order)
                 self.wallets.update()
+                order_type = self.strategy.order_types['sell']
+                self._notify_sell_cancel(trade, order_type)
 
-    def handle_buy_order_full_cancel(self, trade: Trade, reason: str) -> None:
-        """Close trade in database and send message"""
+    def delete_trade(self, trade: Trade) -> None:
+        """Delete trade in database"""
```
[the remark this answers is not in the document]
hroff-1902: not in this PR, please 😆 ???

freqtrade/freqtradebot.py
```diff
@@ -1017,6 +1041,45 @@ def _notify_sell(self, trade: Trade, order_type: str) -> None:
...
+    def _notify_sell_cancel(self, trade: Trade, order_type: str) -> None:
+        """
+        Sends rpc notification when a sell cancel occured.
+        """
+        profit_rate = trade.close_rate if trade.close_rate else trade.close_rate_requested
+        profit_trade = trade.calc_profit(rate=profit_rate)
+        # Use cached ticker here - it was updated seconds ago.
+        current_rate = self.get_sell_rate(trade.pair, False)
+        profit_percent = trade.calc_profit_ratio(profit_rate)
+        gain = "profit" if profit_percent > 0 else "loss"
```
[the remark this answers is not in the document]
hroff-1902: I also wanted to place a comment like "If all of those values applicable", but then removed the comment...
(For both sell and buy parts)

freqtrade/freqtradebot.py
```diff
@@ -820,41 +848,41 @@ def check_handle_timedout(self) -> None:
...
-        reason = "cancelled due to timeout"
         if order['status'] != 'canceled':
+            reason = "cancelled due to timeout"
             corder = self.exchange.cancel_order(trade.open_order_id, trade.pair)
+            logger.info('Buy order %s for %s.', reason, trade)
         else:
             # Order was cancelled already, so we can reuse the existing dict
             corder = order
-            reason = "canceled on Exchange"
+            reason = "cancelled on exchange"
```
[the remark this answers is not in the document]
hroff-1902: 2 cents from me:
whatever (but personally for me cancelled in a message looks more conventional)

freqtrade/freqtradebot.py
```diff
@@ -234,7 +234,7 @@ def enter_positions(self) -> int:
 
         return trades_created
 
-    def get_buy_rate(self, pair: str, tick: Dict = None) -> float:
+    def get_buy_rate(self, pair: str, refresh: bool = False, tick: Dict = None) -> float:
```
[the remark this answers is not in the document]
xmatthias: Additional note:
While it's nice to be constant - using cached tickers for buy side does not work.

Currently, fetch_tickers only caches 'bid' and 'ask' - but `get_buy_rate` requires 'last' as well - so this raises a nice exception.

```
  File "/home/xmatt/development/freqtrade/freqtrade/main.py", line 36, in main
  File "/home/xmatt/development/freqtrade/freqtrade/commands/trade_commands.py", line 20, in start_trading
    worker.run()
  File "/home/xmatt/development/freqtrade/freqtrade/worker.py", line 62, in run
    state = self._worker(old_state=state)
  File "/home/xmatt/development/freqtrade/freqtrade/worker.py", line 98, in _worker
    self._throttle(func=self._process, min_secs=throttle_secs)
  File "/home/xmatt/development/freqtrade/freqtrade/worker.py", line 111, in _throttle
    result = func(*args, **kwargs)
  File "/home/xmatt/development/freqtrade/freqtrade/worker.py", line 121, in _process
    self.freqtrade.process()
  File "/home/xmatt/development/freqtrade/freqtrade/freqtradebot.py", line 159, in process
    self.check_handle_timedout()
  File "/home/xmatt/development/freqtrade/freqtrade/freqtradebot.py", line 854, in check_handle_timedout
    self._notify_buy_cancel(trade, order_type)
  File "/home/xmatt/development/freqtrade/freqtrade/freqtradebot.py", line 569, in _notify_buy_cancel
    current_rate = self.get_buy_rate(trade.pair, False)
  File "/home/xmatt/development/freqtrade/freqtrade/freqtradebot.py", line 259, in get_buy_rate
    if ticker['ask'] < ticker['last']:
KeyError: 'last'
```

Also, when using orderbook, caching would not apply either.

I've tested this running a "heavy trading" strategy that simulates buys (either one that always buys (set buy signal to 1 without condition) - or one that buys at even minutes, sells at random minutes (my preferred strategy to test the bot ... NEVER run either of these on an exchange!) 

By applying the below patch, you can simulate ALL orders being timed out.
While not realistic, it's usefull to test this case.  

``` diff
diff --git a/freqtrade/exchange/exchange.py b/freqtrade/exchange/exchange.py
index b3b34701..1abb0dd1 100644
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -461,9 +461,9 @@ class Exchange:
         closed_order = dry_order.copy()
         if closed_order["type"] in ["market", "limit"]:
             closed_order.update({
-                "status": "closed",
-                "filled": closed_order["amount"],
-                "remaining": 0
+                "status": "canceled",
+                "filled": 0,
+                "remaining": closed_order["amount"]
             })
         if closed_order["type"] in ["stop_loss_limit"]:
             closed_order["info"].update({"stopPrice": closed_order["price"]})

```
---

Either way - as pointed out above - i think the default expectation for both `get_xxx_rate()` functions is to get the most uptodate rate - and only in very selected cases we'd want to use cached values.

freqtrade/freqtradebot.py
yazeed: Thanks for feedback.

freqtrade/freqtradebot.py
yazeed: I'll fix it, thanks for feedback.

freqtrade/freqtradebot.py
```diff
@@ -547,11 +549,37 @@ def _notify_buy(self, trade: Trade, order_type: str) -> None:
...
+        """
+        Sends rpc notification when a buy cancel occured.
+        """
+        current_rate = self.get_buy_rate(trade.pair, False)
+
+        msg = {
+            'type': RPCMessageType.BUY_CANCEL_NOTIFICATION,
+            'exchange': self.exchange.name.capitalize(),
+            'pair': trade.pair,
+            'limit': trade.open_rate_requested,
```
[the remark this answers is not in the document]
yazeed: Fixed.

freqtrade/freqtradebot.py
yazeed: Fixed to "Executing ..." for buys and sells.

freqtrade/freqtradebot.py
yazeed: Fixed

freqtrade/freqtradebot.py
yazeed: Fixed.

freqtrade/rpc/rpc.py
yazeed: The reason I'm filtering out close rate requested is to avoid showing that asterisk for sell orders, because what the asterisk means in this context is that the position is not yet opened, so you can't consider the Profit. Possibly in a future change this table can be enhanced to provide more complete information including showing sell orders?

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
yazeed: So we agree to this.

freqtrade/rpc/telegram.py
```diff
@@ -163,8 +168,11 @@ def send_msg(self, msg: Dict[str, Any]) -> None:
                and self._fiat_converter):
                 msg['profit_fiat'] = self._fiat_converter.convert_amount(
                     msg['profit_amount'], msg['stake_currency'], msg['fiat_currency'])
-                message += ('` ({gain}: {profit_amount:.8f} {stake_currency}`'
-                            '` / {profit_fiat:.3f} {fiat_currency})`').format(**msg)
+                message += (' `({gain}: {profit_amount:.8f} {stake_currency}'
+                            ' / {profit_fiat:.3f} {fiat_currency})`').format(**msg)
+
+        elif msg['type'] == RPCMessageType.SELL_CANCEL_NOTIFICATION:
+            message = "*{exchange}:* Cancelling Sell {pair}".format(**msg)
```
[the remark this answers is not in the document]
yazeed: Fixed

freqtrade/rpc/telegram.py
```diff
@@ -134,13 +134,18 @@ def send_msg(self, msg: Dict[str, Any]) -> None:
...
+                       "*Current Rate:* `{current_rate:.8f}`\n"
+                       "*Total:* `({stake_amount:.6f} {stake_currency}").format(**msg)
 
             if msg.get('fiat_currency', None):
-                message += ",{stake_amount_fiat:.3f} {fiat_currency}".format(**msg)
+                message += ", {stake_amount_fiat:.3f} {fiat_currency}".format(**msg)
             message += ")`"
 
+        elif msg['type'] == RPCMessageType.BUY_CANCEL_NOTIFICATION:
+            message = "*{exchange}:* Cancelling Buy {pair}".format(**msg)
```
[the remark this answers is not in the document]
yazeed: Fixed.

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
yazeed: Thanks for feedback, I've noticed get_buy_rate and get_sell_rate being used differently in the code in terms of REFRESH, and I'm not sure how consistent those are. But maybe they're worth looking at. It seems to me that refresh FALSE should only be used if the code has to use the get_buy_rate or get_sell_rate more than once in a single code path, and thus you'd be using FALSE on any calls after the first one.

Either way, I've fixed this to default to TRUE, and went through the entire code to see if any references to the method weren't passing refresh, and there were none.

freqtrade/freqtradebot.py
```diff
@@ -587,7 +615,7 @@ def exit_positions(self, trades: List[Any]) -> int:
 
         return trades_closed
 
-    def get_sell_rate(self, pair: str, refresh: bool) -> float:
+    def get_sell_rate(self, pair: str, refresh: bool = False) -> float:
```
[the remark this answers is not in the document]
yazeed: Fixed to default to TRUE.

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
yazeed: My rationale here was to make that information available in case it needs to be used by webhook, even if we're not using it in telegram messages.

yazeed:
Fixed tests for double RPC calls in the case of partial buys. Should be fixed in a future RPC, in addition to supporting the scenario for partial sells, both in terms of persistence logic and RPC.

freqtrade/rpc/rpc.py
yazeed: @xmatthias @hroff-1902 I just had an idea to show double asterisks for sell orders, and keep the single asterisk for buy orders. If you like it, will add it in.

hroff-1902: maybe some Unicode emojis can be placed, carefully choosen (not in this PR in this case, along with other emojis)?

yazeed: Good idea for the emojis PR. For this PR, double asterisks or a different symbol?

hroff-1902: whatever (same '*' looks also okay for me)

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
yazeed: ﻿﻿You're right, that's how it renders.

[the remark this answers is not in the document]
hroff-1902: change it to f-strings...

freqtrade/rpc/rpc.py
yazeed: But, as I said the whole point for me was to be able to tell whether you're in the position yet when looking at the status table.

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
xmatthias: Well the intention of refresh was to always refresh from "production" code - and to not refresh when comming from RPC reporting code (/status, /statustable, ...).

> Either way, I've fixed this to default to TRUE, and went through the entire code to see if any references to the method weren't passing refresh, and there were none.

In which case i'm not sure why we need a default ...

freqtrade/rpc/rpc.py
xmatthias: but "time" will tell ... i mean - if an order has an * and is open since "2 hours" - then it's probably clear it's a sell-order that's pending.

however, double ** is also fine ... we should just make sure to document this somewhere (maybe the telegram docs? or using `/help` ?)  ... otherwise it'll raise questions ... 

Apart from that (thinking about this further now) - are you sure about this location for adding the *? 
that's a "telegram only" - visual change i would assume - i'd ratter add a `"open_order": "buy"`, "open_order": "sell"` or `"open_order":None` to the output json so it's rest-api friendly .... ? 

Also - the info is missing from `/status` (without table) ...

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
xmatthias: Unfortunately this bug is still present (the call is comming from `_notify_buy_cancel()`).

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
xmatthias: i'm not certain if a webhook will give you the methods to identify this clearly.
if the other side is a full webservice you have full control over that may be the case...

but it's more  likely a IFTTT service (or similar) which will forward this to "something" (maybe switch the living room lamp to red when you're loosing... :laughing:). 
while IFTTT is a great service, it has it's limitation.

I guess it's ok if we add remarks to the documentation about these fields that these are "hypothetical" fields - and don't represent materialized gains / losses?

freqtrade/rpc/rpc.py
xmatthias: but anyway you can leave that as a "future improvement" - it's just something i noticed while clicking the wrong button :laughing:

hroff-1902:
> The main blocker to me is the crash which is still not completely fixed
> [#2887 (comment)](https://github.com/freqtrade/freqtrade/pull/2887#discussion_r378479114)
> 
> that'll crash the bot each time a order is canceled which we'll obviously need to avoid...

should be a test added? or the reason is very particular and rarely happen later, with future changes?

freqtrade/rpc/rpc.py
yazeed: Yes, I was actually considering a follow-up PR to bring more consistency on /status. The way I thought about it is /status for a verbose view, while /status table is for a more compact one, so the asterisks go in line with this rationale. I'll correct /status table for two asterisks for now, and mention it in the docs.

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
yazeed: Possibly, now all is fixed? :)

xmatthias:
> should a test be added? or the reason is very particular and unlikely happen later, with future changes?

no, we won't need that at this point.
we should probably change how "price caching" works so it also applies to orderbook (currently it's only caching `fetch_ticker()` methods) - but definitely not in this PR.

that will also bring it's own tests ...

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
yazeed: Yes, possibly we can consider doing a PR, but it would probably need to come from you guys, since I'm still new to Python world. I see many other possible refactors needed, we have too many methods that go like check_something_and_do_something_else.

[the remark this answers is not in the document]
xmatthias: yes absolutely - if you look at PR's - we do many refactorings (mostly small PR's)  doing some cleanup, or moving / renaming methods ... so the "not in this PR" message is also pointing to that - this is already a bit ... big :D

freqtrade/rpc/telegram.py
```diff
@@ -553,6 +561,8 @@ def _help(self, update: Update, context: CallbackContext) -> None:
                   "*/stop:* `Stops the trader`\n" \
                   "*/status [table]:* `Lists all open trades`\n" \
                   "         *table :* `will display trades in a table`\n" \
+                  "                  pending buy orders are marked with an asterisk (*)\n" \
+                  "                  pending sell orders are marked with a double asterisk (**)\n" \
```
xmatthias: something in the first of these 2 lines is causing /help to fail.

```
2020-02-13 06:48:06,063 - freqtrade.rpc.telegram - WARNING - Telegram NetworkError: Can't parse entities: can't find end of the entity starting at byte offset 947! Trying one more time.
2020-02-13 06:48:06,141 - freqtrade.rpc.telegram - WARNING - TelegramError: Can't parse entities: can't find end of the entity starting at byte offset 947! Giving up on that message.
```

it's not the length of the message (2 times the 2nd line does work).
I've even tried to retrype it to no avail ... 

Also - to align visualization - i think both lines should be quoted in single tics (which seems to fix the above case) - otherwise this is the only section in help that's not shown as code - or in bold.

I've done that modification - hope that's ok for you.

freqtrade/freqtradebot.py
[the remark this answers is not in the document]
xmatthias: I'd leave this as is for now ... let's do that in a followup

freqtrade/rpc/telegram.py
yazeed: Gotcha. Thanks.

hroff-1902:
^^ resolved conflict in docs

hroff-1902:
is there something here (changes in the user interface) that should be announced in Slack/marked as to be included in the Release Notes for next release? "Notifications for buy/sell cancellations" and their new webhook settings?

PR #2890, opened 2020-02-09
Add tip on running order types for Bittrex

hroff-1902:
"Fixes" running SampleStrategy on Bittrex (market order types) by adding a note in the FAQ.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -53,7 +53,7 @@ Read [the Bittrex section about restricted markets](exchanges.md#restricted-mark
 
 ### I'm getting the "Exchange Bittrex does not support market orders." message and cannot run my strategy
 
-As the message says, Bittrex does not support market orders and you have one of the [order types](configuration.md/#understand-order_types) set to "market". Probably your strategy was written for another exchanges in mind and sets "market" orders for "stoploss" orders, which is correct and preferable for most of other exchanges.
+As the message says, Bittrex does not support market orders and you have one of the [order types](configuration.md/#understand-order_types) set to "market". Probably your strategy was written with other exchanges in mind and sets "market" orders for "stoploss" orders, which is correct and preferable for most of the exchanges supporting market orders (but not for Bittrex).
 
 To fix it for Bittrex, redefine order types in the strategy to use "limit" instead of "market":
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -45,11 +45,23 @@ the tutorial [here|Testing-new-strategies-with-Hyperopt](bot-usage.md#hyperopt-c
 
 You can use the `/forcesell all` command from Telegram.
 
-### I get the message "RESTRICTED_MARKET"
+### I'm getting the "RESTRICTED_MARKET" message in the log
 
 Currently known to happen for US Bittrex users.  
 
 Read [the Bittrex section about restricted markets](exchanges.md#restricted-markets) for more information.
+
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -45,12 +45,24 @@ the tutorial [here|Testing-new-strategies-with-Hyperopt](bot-usage.md#hyperopt-c
...
+### I'm getting the "Exchange Bittrex does not support market orders." message and cannot run my strategy
+
+As the message says, Bittrex does not support market orders and you have one of the [order types](configuration.md/#understand-order_types) set to "market". Probably your strategy was written for another exchanges in mind and sets "market" orders for "stoploss" orders, which is correct and preferable for most of other exchanges.
+
+To fix it for Bittrex, redefine order types in the configuration file (do this for all order types that are defined as "market" in your strategy):
+
+```
+"order_types": {
+    "stoploss": "limit",
+}
```
xmatthias: This is a horrible tip ... (adding only the one to the configuration, that is).

It sounds good in theory - but does actually fail as the order_types dict is not merged.

```
ImportError: Impossible to load Strategy 'BuyEvenSellOdd'. Order-types mapping is incomplete.
```

order_types must therefore be "complete" in the configuration (containing all mandatory keys) - atlernatively it needs to be changed in the strategy.

hroff-1902: right, I have not tested it before...

tried to fix this limitation (allow merge of incomplete dicts from the config) and felt into those problems with default values, when and which one is to be used, etc... that's not an one-line change and requires discussion at least.

so recalling this PR... 
(however, it can be changed to suggest changing vals in the strategy instead of config)

xmatthias: if i remember correctly - we're intentionally not merging this dict as that would create confusion (where is which value actually comming from??).

Unless you print a log per entry (which i would try to avoid) - i would keep it that way ... as otherwise half of the config comes from the strategy, half from the config - and another part from the default values ... which will be messy as hell.

hroff-1902: yeah, that's confusing, so requires discussion and dealing in a sep. PR

I'll better leave this tip, but will suggest changing values in the strategy... As a general tip for Bittrex, not only for SampleStrategy, but for strategies in freqtrade-strategies that may contain values set to "market".

And this dict in SampleStrategy should be commented out (better than), to have default values working for order types. This completely resolves issue with working for Bittrex out of the box.
What do you think (latter should go in a sep. PR, since this one is about docs)?

docs/faq.md
```diff
@@ -45,12 +45,28 @@ the tutorial [here|Testing-new-strategies-with-Hyperopt](bot-usage.md#hyperopt-c
...
-### I get the message "RESTRICTED_MARKET"
+### I'm getting the "RESTRICTED_MARKET" message in the log
 
 Currently known to happen for US Bittrex users.  
 
 Read [the Bittrex section about restricted markets](exchanges.md#restricted-markets) for more information.
 
+### I'm getting the "Exchange Bittrex does not support market orders." message and cannot run my strategy
+
+As the message says, Bittrex does not support market orders and you have one of the [order types](configuration.md/#understand-order_types) set to "market". Probably your strategy was written for another exchanges in mind and sets "market" orders for "stoploss" orders, which is correct and preferable for most of other exchanges.
```
xmatthias: > was written for another exchanges in mind

either: 
"was written for another exchange in mind" (no s)
or 
"was written with other exchanges in mind" (using plural all along)

PR #2900, opened 2020-02-10
Break the old binary file so users are forced to reinstall

xmatthias:
## Summary

Break the old binary file so users are forced to reinstall.

We could also remove it assuming everyone did perform a "pip install ." again since July - however i'd like to leave it with the new message for one release - so really everyone is aware and is not hit by "could not find file".

This should not be relevant anymore - this binary has been deprecated
and is not being used by new installations since July 2019.

### DIFF
--- a/bin/freqtrade
+++ b/bin/freqtrade
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
 
 import sys
-import warnings
+import logging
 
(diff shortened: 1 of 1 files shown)


hroff-1902:
LGTM 👍 
Should this be in the "deprecations" section in the docs?

PR #2902, opened 2020-02-10
Minor: Fix tests for hyperopt_loss

hroff-1902:
Fixes several mypy errors in test_hyperopt.py
(min_date and max_date params were not passed into the loss function)

The dates were copied from another unittest, where they were used this way.


hroff-1902:
> LGTM 👍
> 
> I guess it's pure luck that these tests did not fail completely ...

even worse: for these particular tests, they work on wrong data: hyperopt_results are passed while backtesting_results are expected...

PR #2904, opened 2020-02-11
Minor: Fix usage of an item from BTContainer in tests

hroff-1902:
Fix an issue in tests detected by mypy
(better typing of an item and better check for default value)


hroff-1902:
^^ Cannot re-run failed travis-ci, do not see any re-run button there... (logged in in Travis)

PR #2915, opened 2020-02-13
fix configuration table

xmatthias:
## summary

Fix rendering issue in configuration table.
I think RTD has problems with the combination of Italic and bold (and i'm not sure that's needed).

For now i've fixed it for `max_open_trades` (and relinked https://www.freqtrade.io/en/latest/configuration/ to this branch) - if you think that's ok i'll fix it for the remaining table, too.


hroff-1902:
max_open_trades is okay in documentation_test 👍

PR #2921, opened 2020-02-14
Move rpc send to be after db session add/flash

hroff-1902:
If telegram networking is blocked, then rpc send may hang for timeout. If the user then closes the bot, it may happen that the trade object has not been flushed to persistence properly.

It's better to have rpc call after session add/flush, just to be sure add/flush is made before blocking rpc call.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -527,8 +527,6 @@ class FreqtradeBot:
             ticker_interval=timeframe_to_minutes(self.config['ticker_interval'])
         )
 
-        self._notify_buy(trade, order_type)
-
         # Update fees if order is closed
         if order_status == 'closed':
             self.update_trade_state(trade, order)
@@ -538,6 +536,8 @@ class FreqtradeBot:
 
         # Updating wallets
         self.wallets.update()
+
+        self._notify_buy(trade, order_type)
 
         return True
 


freqtrade/freqtradebot.py
```diff
@@ -527,8 +527,6 @@ def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = N
             ticker_interval=timeframe_to_minutes(self.config['ticker_interval'])
         )
 
-        self._notify_buy(trade, order_type)
-
         # Update fees if order is closed
         if order_status == 'closed':
             self.update_trade_state(trade, order)
```
hroff-1902: I'm not sure here about these two lines -- if it's still okay to call notify_buy after update_trade_state?

xmatthias:
Are you sure this is really happening?

i mean - assuming telegram is blocked / unreachable - then the telegram api would refuse to startup - so i'm not sure if you'd ever get here.

Either way (let's assume you get to this point) - your assumption is that the trade is not flushed to the database... but why is then a part of it there? 

if it's really not flushed to the database because of "something", then this trade would be missing from the database completely.

--- 

I still think that the best explanation for the observed is a downgrade of the bot to a version which didn't have that field - and then upgrading again (the new version was used to create the database - it's not really relevant if a trade was inserted with that version or not).
Database migrations check for column existance - if the newer column would exist, it would not migrate.

hroff-1902:
> Are you sure this is really happening?

No, it was inspired by several articles at the internet describing similar issues with send hanging.

> i mean - assuming telegram is blocked / unreachable - then the telegram api would refuse to startup - so i'm not sure if you'd ever get here.

There is not single 'frontier' or a point to block telegram -- all internet providers do it's own way -- at the L3 routers, by redirecting traffic into black hole, by dropping ACK packets, etc, etc. They should only apply the federal regulatory requirement, so it can be done in any way.

PR #2941, opened 2020-02-17
GitHub actions tests

xmatthias:
## Summary
this will add PyPi builds / uploads to the CI pipeline.

Solve part of the issue: #2112 

An successful workflow is available [here](https://github.com/freqtrade/freqtrade/runs/449795378?check_suite_focus=true) - i've since rebased this branch as i needed about 20 retries to get to that point.

## Quick changelog

- Run on release (hopefully - can only really test this when releasing)
- build pypi package
- upload pypi package

## Open points
- [ ] shall we publish development versions (maybe built nightly) from the develop branch? if yes - requires some form of the next point.

 - [ ] 1172c95 shows an alternative versioning scheme [testpypi](https://test.pypi.org/project/freqtrade/#history). example: `2020.2.dev1581916`.
  Our current versioning scheme for develop versions of develop-commit-hash (`develop-1172c95` is not [PEP440](https://www.python.org/dev/peps/pep-0440/#dvcs-based-version-labels) compliant as ordering is not reliably possible - and such releases can therefore not be uploaded to pypi.
  This commit attempts itself on an alternative Versioning scheme (visible on both test.pypi.org - as well as pypi.org (i'll delete that release after we've decided on this - but i had to test...). The huge drawback to it is that it's not 100% clear which commit the release came from right away - but considering that in the future most users may install via PYPI, it may be neglectible).
  now this point is completely dependant on the implementation of the prior point (publish test/develop versions to pypi) - as otherwise versions for develop may remain as they are.

## TODO 
- [ ] delete test-release from Production pypi @xmatthias

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,8 @@ on:
       - develop
       - github_actions_tests
     tags:
+    release:
+      types: [published]
   pull_request:
   schedule:
     - cron:  '0 5 * * 4'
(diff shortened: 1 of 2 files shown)


hroff-1902:
"shall we publish development versions (maybe built nightly) from the develop branch? if yes - requires some form of the next point." -- I think yes, with the renaming of the branches (in the future) I assume current 'master' becomes the 'release' branch and current 'develop' will become 'master', with latest current builds. They can also be published, why not...

hroff-1902:
"This commit attempts itself on an alternative Versioning scheme (visible on both test.pypi.org - as well as pypi.org (i'll delete that release after we've decided on this - but i had to test...). The huge drawback to it is that it's not 100% clear which commit the release came from right away - but considering that in the future most users may install via PYPI, it may be neglectible)." -- we can move to PEP-compliant versioning schema AND still print in the logs the latest commit (if it's available) till the users all move to the use of freqtrade from pypi (except devs)

xmatthias:
> we can move to PEP-compliant versioning schema AND still print in the logs the latest commit (if it's available) till the users all move to the use of freqtrade from pypi (except devs)

I guess technically we could - but then the same version would have 2 different version strings ... which is also odd.

I guess to start, we should keep it simple and only publish releases ... if we see that process works we can then see how to properly version the files (i'm not a big fan of using the timestamp i'm using at the moment ...).

> I think yes, with the renaming of the branches (in the future)

well renaming the branches is not really relevant, for this, is it? 

the biggest concern i have is that develop releases (pre-release on pypi) are not installable by default unless you specify the exact version (see https://pypi.org/project/freqtrade/ - the pre-release is ONLY visible if you go to release-history) - so my suspicion is that we'll build and upload versions noone is ever going to use (if you want bleeding edge, you'll use the repository anyway).

hroff-1902:
> I guess to start, we should keep it simple and only publish releases ... if we see that process works we can then see how to properly version the files (i'm not a big fan of using the timestamp i'm using at the moment ...).

Agree with this.

xmatthias:
>  when installation from pypi will be made working.

it's just not documented ... it IS working already (and has been since a few months ...)

hroff-1902:
> > when installation from pypi will be made working.
> 
> it's just not documented ... it IS working already (and has been since a few months ...)

I mean more precisely, when most of the users move to use freqtrade from pypi. Then we'll rename the branches without pain and without confusing users...

Issue #2913, opened 2020-02-12
Include more data to hyperopt output


Fredrik81:
I would like to see that total win/loss trades are recorded. this is interesting data when searching for good strat.
If data is stored with the hyperopt then it can be displayed and used when reviewing the hyperopt result and even filtered when using hyperopt list.

Example:
   119/1096:     97 trades, 93 Win, 4 Loss. Avg profit   0.11%. Total profit  0.00061558 BTC (  10.25Σ%). Avg duration 506.6 min. Objective: 2.19842

Thoughts?

hroff-1902:
My proposal:
You can start by adding the table output format (enabled by `--print-table`, for example) into `hyperopt-list`. The table format will save lot's of space in the output so the new columns can be added there to have the output more aligned to output of backtesting (or even same with the addition of the values of the loss-function, number of the epoch and other possible columns specific for hyperopt)

orehunt:
hyperopt results tabulation is not cheap, maybe it shouldn't be the default when running the optimization...

xmatthias:
> hyperopt results tabulation is not cheap

what do you mean with cheap? the computational power needed to generate the table?

hroff-1902:
need to compare the performance btw, we did not do it
between master and current tabular

xmatthias:
This is actually iplemented - we are now showing Win / draw / losses as part of the hyperopt output table.

Stonk8:
I can't see it in hyperopt-list csv export. Should it? I would like it for post processing (win rate)

PR #2943, opened 2020-02-18
Added function to print hyperopt-list as table using tabulate

Fredrik81:
## Summary
Adding --print-table to hyperopt-list in order to get data in a structured form also more compact if more things will be added in the future.
It's fully compatible with all other options like filters and json output.

PEP8 = yes

Solve the issue: #2913
And also part of: #1775

## Quick changelog

- Changed hyperopt.py with print table function
- Added --print-table command to hyperopt-list

## What's new?
```
freqtrade hyperopt-list --print-table --profitable

2020-02-18 22:49:07,991 - freqtrade.loggers - INFO - Verbosity set to 0
2020-02-18 22:49:07,992 - freqtrade.configuration.configuration - INFO - Parameter --print-table detected: True
2020-02-18 22:49:07,992 - freqtrade.configuration.configuration - INFO - Parameter --profitable detected: True
2020-02-18 22:49:07,992 - freqtrade.configuration.configuration - INFO - Using pairlist from configuration.
2020-02-18 22:49:07,992 - freqtrade.configuration.config_validation - INFO - Validating configuration ...
2020-02-18 22:49:08,007 - freqtrade.optimize.hyperopt - INFO - Loaded 48 previous evaluations from disk.
2020-02-18 22:49:08,007 - freqtrade.commands.hyperopt_commands - INFO - 4 profitable epochs found.
+--------+---------+----------+--------------+----------------+----------+----------------+-------------+
| Best   | Epoch   |   Trades | Avg profit   | Total profit   | Profit   | Avg duration   |   Objective |
|--------+---------+----------+--------------+----------------+----------+----------------+-------------|
| *      | 39/48   |        8 | 0.42%        | 0.00020157 BTC | 3.36%    | 113.8m         |     1.99703 |
| *      | 43/48   |        2 | 0.61%        | 0.00007285 BTC | 1.21%    | 92.5m          |     1.97745 |
|        | 45/48   |       24 | 0.25%        | 0.00036351 BTC | 6.05%    | 156.5m         |     2.04067 |
| *      | 48/48   |        8 | 0.61%        | 0.00029328 BTC | 4.88%    | 98.8m          |     1.97194 |
+--------+---------+----------+--------------+----------------+----------+----------------+-------------+
```

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -20,7 +20,8 @@ from colorama import Fore, Style
 from colorama import init as colorama_init
 from joblib import (Parallel, cpu_count, delayed, dump, load,
                     wrap_non_picklable_objects)
-from pandas import DataFrame
+from pandas import DataFrame, json_normalize, isna
+from tabulate import tabulate
 
 from freqtrade.data.history import get_timerange, trim_dataframe
 from freqtrade.exceptions import OperationalException
@@ -294,6 +295,50 @@ class Hyperopt:
                 f"{results['current_epoch']:5d}/{total_epochs}: " +
                 f"{results['results_explanation']} " +
                 f"Objective: {results['loss']:.5f}")
+
+    @staticmethod
+    def print_result_table(config: dict, results: list, total_epochs: int, highlight_best: bool,
+                           print_colorized: bool) -> None:
+        """
+        Log result table
+        """
+        if not results:
+            return
+
+        trials = json_normalize(results, max_level=1)
+        trials = trials[['is_best', 'current_epoch',
+                         'results_metrics.trade_count', 'results_metrics.avg_profit',
+                         'results_metrics.total_profit', 'results_metrics.profit',
+                         'results_metrics.duration', 'loss']]
+        trials.columns = ['Best', 'Epoch', 'Trades', 'Avg profit', 'Total profit',
+                          'Profit', 'Avg duration', 'Objective']
+
+        trials['Best'] = trials['Best'].apply(lambda x: '*' if x else '')
+        trials['Objective'] = trials['Objective'].astype(str)
+
+        if print_colorized:
+            for i in range(len(trials)):
+                if trials.loc[i]['Total profit'] > 0:
+                    trials.at[i, 'Best'] = Fore.GREEN + trials.loc[i]['Best']
+                    trials.at[i, 'Objective'] = "{}{}".format(trials.loc[i]['Objective'],
+                                                              Fore.RESET)
+                    if '*' in trials.loc[i]['Best'] and highlight_best:
+                        trials.at[i, 'Best'] = Style.BRIGHT + trials.loc[i]['Best']
+                        trials.at[i, 'Objective'] = "{}{}".format(trials.loc[i]['Objective'],
+                                                                  Style.RESET_ALL)
+
+        trials['Epoch'] = trials['Epoch'].apply(
+            lambda x: "{}/{}".format(x, total_epochs))
+        trials['Avg profit'] = trials['Avg profit'].apply(
+            lambda x: '{:,.2f}%'.format(x) if not isna(x) else x)
+        trials['Profit'] = trials['Profit'].apply(
+            lambda x: '{:,.2f}%'.format(x) if not isna(x) else x)
+        trials['Total profit'] = trials['Total profit'].apply(
+            lambda x: '{: 11.8f} '.format(x) + config['stake_currency'] if not isna(x) else x)
+        trials['Avg duration'] = trials['Avg duration'].apply(
+            lambda x: '{:,.1f}m'.format(x) if not isna(x) else x)
+
+        print(tabulate(trials.to_dict(orient='list'), headers='keys', tablefmt='psql'))
 
     def has_space(self, space: str) -> bool:
         """
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -306,15 +306,18 @@ class Hyperopt:
             return
 
         trials = json_normalize(results, max_level=1)
-        trials = trials[['is_best', 'current_epoch',
-                         'results_metrics.trade_count', 'results_metrics.avg_profit',
-                         'results_metrics.total_profit', 'results_metrics.profit',
-                         'results_metrics.duration', 'loss']]
+        trials['Best'] = ''
+        trials = trials[['Best', 'current_epoch', 'results_metrics.trade_count',
+                         'results_metrics.avg_profit', 'results_metrics.total_profit',
+                         'results_metrics.profit', 'results_metrics.duration',
+                         'loss', 'is_initial_point', 'is_best']]
         trials.columns = ['Best', 'Epoch', 'Trades', 'Avg profit', 'Total profit',
-                          'Profit', 'Avg duration', 'Objective']
+                          'Profit', 'Avg duration', 'Objective', 'is_initial_point', 'is_best']
 
-        trials['Best'] = trials['Best'].apply(lambda x: '*' if x else '')
+        trials.loc[trials['is_initial_point'], 'Best'] = '*'
+        trials.loc[trials['is_best'], 'Best'] = 'Best'
         trials['Objective'] = trials['Objective'].astype(str)
+        trials = trials.drop(columns=['is_initial_point', 'is_best'])
 
         if print_colorized:
             for i in range(len(trials)):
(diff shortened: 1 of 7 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -295,6 +296,50 @@ def _format_explanation_string(results, total_epochs) -> str:
...
+        trials['Avg profit'] = trials['Avg profit'].apply(
+            lambda x: '{:,.2f}%'.format(x) if not isna(x) else x)
+        trials['Profit'] = trials['Profit'].apply(
+            lambda x: '{:,.2f}%'.format(x) if not isna(x) else x)
+        trials['Total profit'] = trials['Total profit'].apply(
+            lambda x: '{: 11.8f} '.format(x) + config['stake_currency'] if not isna(x) else x)
+        trials['Avg duration'] = trials['Avg duration'].apply(
+            lambda x: '{:,.1f}m'.format(x) if not isna(x) else x)
+
+        print(tabulate(trials.to_dict(orient='list'), headers='keys', tablefmt='psql'))
```
xmatthias: I guess we should consider switching the other uses of "tabulate" to psql too ... this table looks better than ours .... :+1: 

Not for this PR though ... just a small sidenote

Fredrik81: Thanks! Yeah it does look nice :-)
I can work on PR for that after i'm done with this.

yazeed:
Just wondering... Why do we need to have both formats? Why not go with this as default?

xmatthias:
> Why do we need to have both formats? Why not go with this as default?

i guess (!) that it's because the "default" format is what hyperopt outputs ...

Since that output is "ongoing" - it'll be pretty difficult to properly format it into a table ...
It'll also output the "processing points" (`.` ) - which makes formatting as table very ... strange as it would have interrupted output.

I am however with you - this format looks better ... and i'm not certain that the 2 formats need to be aligned

Fredrik81:
> i guess (!) that it's because the "default" format is what hyperopt outputs ...

I have been playing around with option to output hyperopt as table also.. it's not as strait forward but we can make it work.. can send in a PR at later time as POC and discussions.

Fredrik81:
So should i change this PR to output only table or change the current format as optional with parameter?

hroff-1902:
* tabular format is much prettier than legacy format
* actual reason for this was another -- it will keep lots of horizontal space and allow to insert wins/losses columns, to align output with backtesting tables and provide this useful info
* since tabular format is much better from different sides, main hyperopt output should also be switched to this (if this is feasible)
* however, it's not as easy as for this utility subcommand, since main hyperopt output is printed dynamicaly, it's not a complete table to print at once. This probably will require introducing a rolling progress bar instead of printing dots (a progress bar which allows to print info above it)
* so I proposed to @Fredrik81 to start with this utility command
* the `--print-table` option was proposed to keep same look and fill till the main hyperopt output is also moved into tabular
* after that this option will be definitely removed -- we definitely do not need both formats if both this utility and main hyperopt can print in this fancy tabular format
* maybe it's not needed now for this utility if we do not need the utility command and main hyperopt printing in the same format.

Let's decide. I'm not insisting on this `--print-table` option, but described above why it was introduced. From one point of view, it keeps the 'compatibility', or similarness of the output, from other point of view, we then will be needing to start the deprecation of it, which will be artificially created pain...

hroff-1902:
as a possible option: we could postpone it till main hyperopt will have same output and release (merge) both in the same time. or, if this not an issue for develop, we'll merge it (as the default format, with no --print-table option) now, then work on main hyperopt and wait with next master release to have them both in the same master release... thus, it becomes an organizational issue.

xmatthias:
but is it an issue if we have one release where both commands don't output the same format, if the 2nd format is a lot better?

I would not delay this PR just to fix it in the "real" hyperopt output - as i have my doubts that will work correctly (the table will break for every hyperopt that includes a `print()` statement) - and handling that will not be easy.

maybe we can reverse this commands behaviour - and use a `--print-legacy-hyperopt` (please better name) to restore the "original" formatting? 
so if someone needs aligned output, that's an option (we can still remove later should the adjustment of the "main" hyperopt work)?

Fredrik81:
I would propose to keep the --print-table for now as there will be consistency to current outputs and then we can work on getting all other things ready to make a switch at later stage.
Things that i see is missing before we change:
1. All other table's should have same layout
2. Hyperopt and hyperopt-show should be changed to print table format
3. Add Win/Loss to hyperopt so we can also give some additional information (benefits) with table format

hroff-1902:
My willful decision here (someone needs to get the initiative) is the following (we discussed -pros and -contras):

let's make it the default format for `hyperopt-list` (and for `hyperopt` and `hyperopt-show` when this will be implemented), without `--print-table` option, in this PR. We do not need painful deprecation procedures or announcing of removal options that were just introduced small time ago.

Feel free to post your objections if any.

@Fredrik81 could you change the code according to this?

Fredrik81:
Sure will send a new PR asap

freqtrade/optimize/hyperopt.py
```diff
@@ -295,6 +296,50 @@ def _format_explanation_string(results, total_epochs) -> str:
...
+
+        trials = json_normalize(results, max_level=1)
+        trials = trials[['is_best', 'current_epoch',
+                         'results_metrics.trade_count', 'results_metrics.avg_profit',
+                         'results_metrics.total_profit', 'results_metrics.profit',
+                         'results_metrics.duration', 'loss']]
+        trials.columns = ['Best', 'Epoch', 'Trades', 'Avg profit', 'Total profit',
+                          'Profit', 'Avg duration', 'Objective']
+
+        trials['Best'] = trials['Best'].apply(lambda x: '*' if x else '')
```
hroff-1902: I propose to keep the '*' mark for epochs that randomly seed, and use another mark (for example '!' or '+' for best results. Both can be set. The mark for the best results help when no colorization is used.

freqtrade/optimize/hyperopt.py
```diff
@@ -295,6 +296,50 @@ def _format_explanation_string(results, total_epochs) -> str:
...
+        trials['Objective'] = trials['Objective'].astype(str)
+
+        if print_colorized:
+            for i in range(len(trials)):
+                if trials.loc[i]['Total profit'] > 0:
+                    trials.at[i, 'Best'] = Fore.GREEN + trials.loc[i]['Best']
+                    trials.at[i, 'Objective'] = "{}{}".format(trials.loc[i]['Objective'],
+                                                              Fore.RESET)
+                    if '*' in trials.loc[i]['Best'] and highlight_best:
```
hroff-1902: Is it shifted one level right intentionally or by error? This changes the colorization schema (not profitable bests are now also printed in bold/bright). I would propose to keep the current colorization schema (we've chosen it from 4 or 6 variants, when it was implemented, so it's 'carefully selected' 😉 )

freqtrade/optimize/hyperopt.py
```diff
@@ -295,6 +296,50 @@ def _format_explanation_string(results, total_epochs) -> str:
...
+                        trials.at[i, 'Best'] = Style.BRIGHT + trials.loc[i]['Best']
+                        trials.at[i, 'Objective'] = "{}{}".format(trials.loc[i]['Objective'],
+                                                                  Style.RESET_ALL)
```
hroff-1902: I think that the lines dividing the columns should not be colorized; I'll take this one after this PR is merged, I think I know how to make it easily...

freqtrade/optimize/hyperopt.py
```diff
@@ -295,6 +296,50 @@ def _format_explanation_string(results, total_epochs) -> str:
...
+        trials['Epoch'] = trials['Epoch'].apply(
+            lambda x: "{}/{}".format(x, total_epochs))
+        trials['Avg profit'] = trials['Avg profit'].apply(
+            lambda x: '{:,.2f}%'.format(x) if not isna(x) else x)
+        trials['Profit'] = trials['Profit'].apply(
+            lambda x: '{:,.2f}%'.format(x) if not isna(x) else x)
+        trials['Total profit'] = trials['Total profit'].apply(
+            lambda x: '{: 11.8f} '.format(x) + config['stake_currency'] if not isna(x) else x)
+        trials['Avg duration'] = trials['Avg duration'].apply(
+            lambda x: '{:,.1f}m'.format(x) if not isna(x) else x)
```
hroff-1902: I understand you took formatting mostly from existing method, but it should be better aligned (all to the right), current looks a bit sloppy... Maybe not in this PR.

freqtrade/optimize/hyperopt.py
Fredrik81: I changed so 'is_initial_point' will set '*' and is_best will set 'Best'
We have space for that so might be more intuitive for new users.

freqtrade/optimize/hyperopt.py
Fredrik81: Yeah it was intentional. Didn't think it makes sense to color mark things that do not make us a profit :-D changed now.

freqtrade/optimize/hyperopt.py
Fredrik81: I was working a lot around this coloring topic tbh. Yes it can be done but from what i can tell the only way is to loop thru and set color on each individual column..
I can do that in this PR but will be later this week i guess.. on a trip for a few days

Fredrik81: It's only coloring the values now. Might be a more "pretty" way to do it but did not see much of a impact in performance

hroff-1902:
LGTM 👍

PR #2944, opened 2020-02-19
Move defaultstrategy

xmatthias:
## Summary
Move all test-strategies into a subdir, move DefaultStrategy to test-directory.

Remove default-loading of strategy from internal directory - as there is no strategy there anymore and users must use user_data paths.

## Quick changelog

- move DefaultStrategy to test-dir
- move all test-strategies to subdir


hroff-1902:
Do we still need to scan freqtrade (optimize, strategy) dirs after that in the resolver?
(Think so -- for loss functions at least...)

xmatthias:
Yes we do (and i don't think it'll ever go away completely).
for Hyperopt and loss functions and pairlists (!)

tests/strategy/test_strategy.py
```diff
@@ -31,21 +31,21 @@ def test_search_strategy():
...
 
 
 def test_search_all_strategies_with_failed():
-    directory = Path(__file__).parent
+    directory = Path(__file__).parent / "strats"
     strategies = StrategyResolver.search_all_objects(directory, enum_failed=True)
     assert isinstance(strategies, list)
-    assert len(strategies) == 4
+    assert len(strategies) == 3
     # with enum_failed=True search_all_objects() shall find 3 good strategies
```
hroff-1902:
```suggestion
    # with enum_failed=True search_all_objects() shall find 2 good strategies
```

xmatthias: lol, i expected questions regarding this :D (i had them too)

when using "search_all_objects" previously (on the `tests/strategies` directory) - 4 strategies were found.
2 of them were physically in this location (Legacy, Failing) - and 2 times it was the default strategy - once loaded from where it really was, and once because it's improted in `test_strategy.py` (so it's in a seperate file).

since that imported version "moved away" - DefaultStrategy is there only once - which reduces the total amount from 4 to 3 (but it's always been only 3 distinct strategies).

PR #2951, opened 2020-02-21
batched hyperopt

orehunt:
## Summary
hyperopt improvements

## Quick changelog

- rebased changes made by @hroff-1902  in #1765
- removed fix for models because scikit-optimize provides `model_queue_size` as argument for the optimizer since `0.7.0`
- added some heuristics for guessing the number of epochs and initial points which can still customized by an "effort" value which works as a multiplier
- added multi optimizer mode where each job has its own optimizer interface
- save optimizer state to disk

todo:
- [ ] docs
- [ ] tests
- [x] better formatting for the output (?)
- [x] do some profiling to find out if it really has a pickling bottleneck 
I don't see memory or performance issues, the blocker is the optimizer
- [x] is global hyperopt class fine? right? :(
Moved the callback classes to their own file for clearer imports..


orehunt:
hyperopt_chart

hroff-1902:
Wow 😮

freqtrade/optimize/hyperopt.py
```diff
@@ -427,39 +431,73 @@ def _format_results_explanation_string(self, results_metrics: Dict) -> str:
...
+        return result
 
-        This may cease working when skopt updates if implementation of this intrinsic
-        part changes.
+    def opt_generator(self):
+        while True:
+            if self.f_val:
+                # print("opt.tell(): ",
+                #       [v['params_dict'] for v in self.f_val], [v['loss'] for v in self.f_val])
+                functools.partial(self.opt.tell,
```
untoreh: @hroff-1902  bad news

orehunt:
this needs some docs

hroff-1902:
@orehunt  I can resolve the conflict by myself some time later (soon), when I start reviewing this PR in detail, do not spend your efforts on this...

hroff-1902:
I will start merging it after the progress bar is implemented for hyperopt instead of dots, that will be easier path to merge all latest changes.

orehunt:
I should warn that I did play a bit with tddm for loading but it required passing the IO writer object around processes and that caused problems so I did not bother much, hopefully whatever other implementation does not have pickling issues

hroff-1902:
> I should warn that I did play a bit with tqdm for loading but it required passing the IO writer object around processes and that caused problems so I did not bother much, hopefully whatever other implementation does not have pickling issues

https://github.com/freqtrade/freqtrade/blob/3a8b68c0fd721094c606286b4dce6a86d89442d9/freqtrade/optimize/hyperopt.py#L735

orehunt:
> > Thanks,
> Same for other implementations of progress bar:
> 
> https://github.com/freqtrade/freqtrade/blob/3a8b68c0fd721094c606286b4dce6a86d89442d9/freqtrade/optimize/hyperopt.py#L735
I think jobs still need access to progressbar otherwise the progressbar will only get updated once per batch, per job interactiveness would require either one of:
- passing the object through a size 1 queue like I did with `results_board` and each job updates the writer
- use a queue just to signal new jobs (like `log_results_immediate`) and run the progressbar writer in its own process flush the new jobs queue
- re add a custom callback (which I removed) that updates the progressbar, but only on job completion, not on job dispatch

orehunt:
had to make some changes here because as it was optimizers were fitting almost similar models, because they all shared the results, duh, I knew it but left it as it was because testing showed faster convergence.

Now I did it in a similar way where there is only one optimizer that is copied over and run with a different seed. 

Results are now shared only as a way to cache already computed points in case models end up asking the same points, but in multi mode, optimizers have their own unadulterated view of the search space, which should provide more different results.

yazeed:
Great PR. :-)

hroff-1902:
man, it will wait its time...

it requires some amount of efforts and positive mood (and waits for the progressbar pr to be merged, but it's not a real dependency), so I'm not personally ready for it to merge, but watching your efforts...

pls do not be sad or angry -- i hope there will be better times if we all will survive...

xmatthias:
I'll be so bold to reopen this PR.

I hope you didn't see our silence here as being critical about this PR. 
I agree with @yazeed:

> Great PR. :-)

We've been mostly quiet because for myself, hyperopt is not my speciality (i leave that mostly to @hroff-1902) - and he's been overall quiet over the last for weeks for personal reasons - and a huge PR like this needs time, and full attention to review and test and play with.

The benefits for hyperopt will be great, and we're happy that someone picked up the "old, dusty" work @hroff-1902 started ages ago  - so i'll reopen this - if you're not interrested in driving this forward anymore we can take over - but closing this would be a shame!

flomerz:
Cool feature, looking forward to it :+1: :+1:  I think this feature is the most valuable pull request at the moment :1st_place_medal:

freqtrade/commands/cli_options.py
```diff
@@ -195,6 +195,36 @@ def __init__(self, *args, **kwargs):
...
+        default=constants.HYPEROPT_EFFORT,
+    ),
+    "mode": Arg(
+        '--mode',
+        help='Switches hyperopt to use one optimizer per job, use it'
+             'when backtesting iterations are cheap (default: %(default)s).',
+        metavar='NAME',
+        default=constants.HYPEROPT_MODE),
+    "n_points": Arg(
+        '--n-points',
```
xmatthias, line 213: Whenever i give this a value of > 1 , i get `RuntimeError: Random evaluations exhausted and no model has been fit.`.

Will need investigation.
*note*: Seems to only happen in combination with `--mode multi`.

freqtrade/optimize/hyperopt.py
```diff
@@ -609,10 +1050,267 @@ def load_previous_results(results_file: Path) -> List:
...
+    def track_points(self, trials: List = None):
+        """
+        keep tracking of the evaluated points per optimizer random state
+        """
+        # if no trials are given, use saved trials
+        if not trials:
+            if len(self.trials) > 0:
+                if self.config.get('hyperopt_continue_filtered', False):
+                    raise ValueError()
+                    # trials = filter_trials(self.trials, self.config)
```
xmatthias, line 1143: the method `filter_trials` is undefined - seems to be missing since it's been introduced (history does not have this method at all).

freqtrade/optimize/hyperopt.py
```diff
@@ -609,10 +1050,267 @@ def load_previous_results(results_file: Path) -> List:
...
+                    jobs = parallel._effective_n_jobs()
+                    logger.info(f'Effective number of parallel workers used: {jobs}')
+                    # update epochs count
+                    opt_points = self.opt_points
+                    prev_batch = -1
+                    epochs_so_far = len(self.trials)
+                    epochs_limit = self.epochs_limit
+                    columns, _ = os.get_terminal_size()
+                    columns -= 1
+                    while epochs_so_far > prev_batch or epochs_so_far < self.min_epochs:
```
xmatthias, line 1275: Will need investigation.

`os.get_terminal_size()` breaks several tests - but i'm not certain it's still needed, as we're now printing hyperopt results via table.

freqtrade/optimize/hyperopt.py
```diff
@@ -578,20 +707,332 @@ def _format_results_explanation_string(self, results_metrics: Dict) -> str:
...
+        told = 0  # told
+        Xi_d = []  # done
+        yi_d = []
+        Xi_t = []  # to do
+        # if opt.void == -1 the optimizer failed to give a new point (between dispatches), stop
+        # if asked == asked_d  the points returned are the same, stop
+        # if opt.Xi > sss the optimizer has more points than the estimated search space size, stop
+        while opt.void != -1 and asked != asked_d and len(opt.Xi) < sss:
+            asked_d = asked
+            asked = opt.ask(n_points=self.ask_points, strategy=self.lie_strat())
```
xmatthias, line 937: Note to self:
providing strategy to `opt.ask` does not make sense in this way - it'll decide if the optimize function should be minimized (`cl_min`), optimized (`cl_max`), or averaged (`cl_mean`).

Using a random strategy for this (or letting the user pick) does not make sense.
What could make sense is to let the hyperopt-loss function determine if better is higher, or if better is lower (optimize vs. minimize). 
This could remove some of "negativation" of loss-functions like Sharpe and Sortino, where originally higher is better.

maverick193:
Any news about this PR ? Really want to use it, Thanks to you all !!

csulit:
Any update with this PR?

hroff-1902:
still reviewing it, pls wait...

xmatthias:
To be honest, chances of this being merged as is are almost zero (not only because of the merge conflicts).

It's still open to remind me of some features we might be able to extract from this - but overall, this puts too much focus on "utilize 100% of the cpu all the time" - while sacrificing "actually get better results".

rodrigogs:
So, I was just wondering... would the ideal processor be that one with a higher clock per core, and not that with lots of cores then?

I mean, most of the time I have just one core burning, and it's the longest operation.

xmatthias:
> I mean, most of the time I have just one core burning, and it's the longest operation.

The reason for this is explained in this scikit-optimize issue: https://github.com/scikit-optimize/scikit-optimize/issues/768

xmatthias:
Closing this due to the comments above (specifically https://github.com/freqtrade/freqtrade/pull/2951#issuecomment-991493997).

PR #2954, opened 2020-02-22
Improve and fix buy / sell Rate caching

xmatthias:
## Summary
Buy / sell rate caching is broken as we discovered in #2887 

Caching did not work when orderbook was enabled, making a call to the exchange for every RPC call.

This now implements caching in a slightly different manner:
It caches the result calls for 5 seconds - so notifications sent within 5 seconds will reuse the present value - otherwise a query to the exchange is sent.

If a fresh value is requested (only applies for calls right before creating an order) - then refresh is set to True, forcing a exchange query.

## Quick changelog

- Implement caching for `get_buy_rate()`
- implement caching for `get_sell_rate()`
- remove caching from `exchange.fetch_ticker()`

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -253,7 +253,7 @@ class FreqtradeBot:
         else:
             if not tick:
                 logger.info('Using Last Ask / Last Price')
-                ticker = self.exchange.fetch_ticker(pair, refresh)
+                ticker = self.exchange.fetch_ticker(pair)
             else:
                 ticker = tick
             if ticker['ask'] < ticker['last']:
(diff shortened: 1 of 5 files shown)


freqtrade/freqtradebot.py
```diff
@@ -54,6 +55,9 @@ def __init__(self, config: Dict[str, Any]) -> None:
 
         self._heartbeat_msg = 0
 
+        self._sell_rate_cache = TTLCache(maxsize=100, ttl=5)
```
hroff-1902: this can later be moved to the dataprovider... what do you think?

PR #2956, opened 2020-02-22
Shuffle whitelisted pairs

yazeed:
I noticed in backtesting results and in trading modes, that if you have a long list of pairs in your config file, that freqtrade seems to trade pairs in alphabetical order OR according to the order you have set them in your config file (still need to confirm this one yet), meaning that it seems to trade more the ones that come on top of the stack. This doesn't seem to me like the preferred behaviour, in the sense that most users do not take into account the order freqtrade scans their pairs.

I also wasn't sure if it's worth it to add a config flag to allow users to choose between shuffle or not, which is also a possibility, but I lean more towards having the shuffle as a default.

This PR suggests adding a simple random.shuffle on whitelisted pairs, without the config option.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -11,6 +11,7 @@ from threading import Lock
 from typing import Any, Dict, List, Optional, Tuple
 
 import arrow
+import random
 from requests.exceptions import RequestException
 
 from freqtrade import __version__, constants, persistence
@@ -221,6 +222,7 @@ class FreqtradeBot:
                 logger.info("No currency pair in active pair whitelist, "
                             "but checking to sell open trades.")
             else:
+                random.shuffle(whitelist)
                 # Create entity and execute trade for each pair from whitelist
                 for pair in whitelist:
                     try:


yazeed:
Obviously it does mess with some tests, so lets discuss it first. :)

hroff-1902:
The overal idea is okay, but the current behavior is 'by design', in my opinion: when I place pairs in the whitelist and use StaticPairList, I CAN expect the pairs are looked up in this order; VolumePairList also sorts the pairs by volume and this is behavior which I would keep... Maybe some additional words on this should be added into the docs...

However, some users and some configurations may desire the behavior you propose, this can be also useful, but it should be configurable at least.

So this can be done by introducing a new setting somewhere in the configuration, but I see the great opportunity for this implemented as a pairlist filter: say, ShufflePairs (or with more generic name ReorderPairs) added in the list of pair list filters may make this re-shufling in the random order. Or there can be some other methods which reorder pairs other, non-random order...

xmatthias:
Well it used to be random (or semi-random) before we changed to the pairlist structure we have now - but the change to not randomize was done with intent - so a static pairlist remains static and in the same order. 

It's a perfect usecase for an additional pairlist (filter) ... so users can decide easily if they wanna shuffle (or not).

yazeed:
Sounds good, then we agree to doing this as a pairlist filter.

Not quite sure why the test fails, but I'll also get to it after switching it to be filter-based.

hroff-1902:
> Sounds good, then we agree to doing this as a pairlist filter.
> 
> Not quite sure why the test fails, but I'll also get to it after switching it to be filter-based.

Yes, but with extra-filter the existing tests won't fail and extra tests should be added to test with this filter.
(So I can propose you to add the 'random-state' parameter to this filter to seed the random state, which can simplify testing since it will allow to get reproducible results...)

hroff-1902:
Shall we close it in favor of upcoming shuffling pair list?

yazeed:
> Shall we close it in favor of upcoming shuffling pair list?

No, I'm working on the ShufflePairs Filter, will commit it here once I have something ready.

xmatthias:
I've market it "don't merge" for now so it's clear this ain't complete yet

Please remove that label once it's ready

hroff-1902:
closing it in favor of #3322 
Thanks, Yazeed!

PR #2960, opened 2020-02-22
indicators_full.j2 and sample_strategy.py ordering and added indicators

yazeed:
This doesn't affect much, it just adds some ordering and consistency onto **indicators_full.j2** file, along with adding a mention of Weighted Bollinger Bands (EMA based as opposed to SMA based), Ultimate Oscillator, and Keltner Channel.

I wasn't entirely sure if any logic in freqtrade depends on this file, so I didn't uncomment or comment anything.

### DIFF
--- a/freqtrade/templates/sample_strategy.py
+++ b/freqtrade/templates/sample_strategy.py
@@ -125,7 +125,7 @@ class SampleStrategy(IStrategy):
         # ------------------------------------
 
         # ADX
-        # dataframe['adx'] = ta.ADX(dataframe)
+        dataframe['adx'] = ta.ADX(dataframe)
 
         # # Plus Directional Indicator / Movement
         # dataframe['plus_dm'] = ta.PLUS_DM(dataframe)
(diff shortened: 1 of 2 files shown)


yazeed:
Not gonna touch this further at this point, this was just some stuff logged on my side, as I was trying to list out indicators usage code.

freqtrade/templates/subtemplates/indicators_full.j2
```diff
@@ -2,87 +2,130 @@
...
 dataframe['bb_middleband'] = bollinger['mid']
 dataframe['bb_upperband'] = bollinger['upper']
+dataframe["bb_percent"] = (
+    (dataframe["close"] - dataframe["bb_lowerband"]) /
+    (dataframe["bb_upperband"] - dataframe["bb_lowerband"])
+)
+dataframe["bb_width"] = (
+    (dataframe["bb_upperband"] - dataframe["bb_lowerband"]) / dataframe["bb_middleband"]
+)
+
```
xmatthias: I would maybe add these in as comment assuming people will do a first try with this - but not use them - so it would be wasted cpu cycles ...

yazeed: I'll take care of sample_strategy.py, thanks.

freqtrade/templates/sample_strategy.py
```diff
@@ -124,87 +124,131 @@ def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame
...
+        # Stochastic Fast
         stoch_fast = ta.STOCHF(dataframe)
         dataframe['fastd'] = stoch_fast['fastd']
         dataframe['fastk'] = stoch_fast['fastk']
 
-        # # Stoch RSI
+        # # Stochastic RSI
         # stoch_rsi = ta.STOCHRSI(dataframe)
         # dataframe['fastd_rsi'] = stoch_rsi['fastd']
         # dataframe['fastk_rsi'] = stoch_rsi['fastk']
```
xmatthias: Based on some recent acivity in #sharing-strategies on slack, i think we should add a warning to this indicator - as it's calculation is ... not in line what you see on exchange / tradingview (not to say ta-lib calculates it wrongly).

The slack-channel also includes a proposed new calculation method ... so i guess we could have that in technical (or evaluate how finta calculates stoch_rsi).

I'll however merge this anyway as that will require seperate evaluation and is just a "random" spot  while reviewing this.

PR #2976, opened 2020-02-25
 [Doc update config section] - [minor]

gaugau3000:
## Summary
Quick fix and add extra info for configuration section in the doc. 

## Quick changelog

- 7d7318a Full or killed order doesn't exist.
- 76c449c Add extra info for VolumePairList.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -532,7 +532,11 @@ It uses configuration from `exchange.pair_whitelist` and `exchange.pair_blacklis
 
 `refresh_period` allows setting the period (in seconds), at which the pairlist will be refreshed. Defaults to 1800s (30 minutes).
 
-`VolumePairList` is based on the volume of the last 24 hours.
+`VolumePairList` is based on the ticker data, as reported by the ccxt library:
+
+* The `bidVolume` is the volume (amount) of current best bid in the orderbook.
+* The `askVolume` is the volume (amount) of current best ask in the orderbook.
+* The `quoteVolume` is the amount of quote (stake) currency traded (bought or sold) in last 24 hours.
 
 ```json
 "pairlists": [{
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -340,7 +340,7 @@ This is most of the time the default time in force. It means the order will rema
 on exchange till it is canceled by user. It can be fully or partially fulfilled.
 If partially fulfilled, the remaining will stay on the exchange till cancelled.
 
-**FOK (Full Or Kill):**
+**FOK (Fill Or Kill):**
 
 It means if the order is not executed immediately AND fully then it is canceled by the exchange.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -532,6 +532,8 @@ It uses configuration from `exchange.pair_whitelist` and `exchange.pair_blacklis
 
 `refresh_period` allows setting the period (in seconds), at which the pairlist will be refreshed. Defaults to 1800s (30 minutes).
 
+`VolumePairList` is based on the volume of the last 24 hours.
+
 ```json
 "pairlists": [{
         "method": "VolumePairList",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         python-version: [3.7, 3.8]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
     - name: Set up Python
       uses: actions/setup-python@v1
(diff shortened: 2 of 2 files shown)


docs/configuration.md
```diff
@@ -532,6 +532,8 @@ It uses configuration from `exchange.pair_whitelist` and `exchange.pair_blacklis
 
 `refresh_period` allows setting the period (in seconds), at which the pairlist will be refreshed. Defaults to 1800s (30 minutes).
 
+`VolumePairList` is based on the volume of the last 24 hours.
```
hroff-1902: This is true only for quoteVolume, proposed change instead:

```
`VolumePairList` is based on the ticker data, as reported by the ccxt library:

* The `bidVolume` is the volume (amount) of current best bid in the orderbook.
* The `askVolume` is the volume (amount) of current best ask in the orderbook.
* The `quoteVolume` is the amount of quote (stake) currency traded (bought or sold) in last 24 hours.
```

docs/configuration.md
```diff
@@ -340,7 +340,7 @@ This is most of the time the default time in force. It means the order will rema
 on exchange till it is canceled by user. It can be fully or partially fulfilled.
 If partially fulfilled, the remaining will stay on the exchange till cancelled.
 
-**FOK (Full Or Kill):**
+**FOK (Fill Or Kill):**
```
hroff-1902: nice catch! 👍

docs/configuration.md
gaugau3000: Interesting, here the best bid and best ask is an aggregated level of the order book? Because if it just the volume of best bid or ask I not sure how useful it can be to filter markets?

hroff-1902: all this actually depends on market. Even quoteVolume (can be other value, than 24-h rolling aggregate)

hroff-1902: Press "Commit suggestion" if you agree with this my proposed change.

gaugau3000: I don't have the button commit suggestion I don't know why. For bidvolume and askvolume I don't understand how it can be useful to filter markets by the volume of the best bid or ask. If btc has best bid of 30 dollar amount anf shit coin 2000 dollar bid it will remove btc and keep shit coin? Maybe I missed smt? Tks

hroff-1902: me too, I also do not know how other options (bidVolume, askVolume) can be used in practice, but it's legacy. existing values for this setting, Your description is wrong since it's only applicable to one of the available options.

hroff-1902: commited by myself, pls review

gaugau3000: I know my description is wrong :-) Just trying to understand.

hroff-1902:
@xmatthias pls have a look what's wrong with CI checks

docs/configuration.md
```diff
@@ -532,6 +532,12 @@ It uses configuration from `exchange.pair_whitelist` and `exchange.pair_blacklis
 
 `refresh_period` allows setting the period (in seconds), at which the pairlist will be refreshed. Defaults to 1800s (30 minutes).
 
+`VolumePairList` is based on the ticker data, as reported by the ccxt library:
+
+* The `bidVolume` is the volume (amount) of current best bid in the orderbook.
+* The `askVolume` is the volume (amount) of current best ask in the orderbook.
+* The `quoteVolume` is the amount of quote (stake) currency traded (bought or sold) in last 24 hours.
```
xmatthias: interrestingly, there would be `baseVolume` too - but we've excluded it (it's not in `SORT_VALUES`) - however i'm not certain why as there's no functional reason for that (probably it was missed ... ? )

hroff-1902: I guess, because baseVolume is in base currency, and the numbers differ in scale for the same stake currency (although ccxt documentation says that all amounts here in the quote currency)

xmatthias: not for baseCurrency - you're right ... that's the reasion :+1:

docs/configuration.md
hroff-1902: @gaugau3000 can I merge this PR with my changes?

we often use a short mark "LGTM 👍" for this (due to historical reasons), just to log in the comments explicit agreement with changes made, you may use any words, but I do not see your explicit agreement with the additional changes above... (just to be it explicit)

gaugau3000: LGTM :+1: you can merge. By the way you are right the discussion about bidvolume and askVolume should be in a topic and far from this pull request (like the S of SOLID in programming) .

hroff-1902:
> I've updated the checkout action to v2 - according to the source repository this should fix the problem.

This should be done in a sep. PR, you know... 😉

PR #2978, opened 2020-02-26
Wins/draws/losses/median profit in hyperopt output

yazeed:
This PR suggests a couple of additional info to the hyperopt output (wins/draws/losses, and median profit which kinda shows if your avg profit is skewed due to lucky trades in that epoch, it's a simple way to detect outlier trades in an epoch), it should support #2943 if that PR is merged into develop.

Let me know your thoughts.

Will work on the tests once thoughts are finalized into this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         python-version: [3.7, 3.8]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
     - name: Set up Python
       uses: actions/setup-python@v1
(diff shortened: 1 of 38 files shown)


hroff-1902:
That's fine, but the overall length is longer than the width of my terminal, so printed in two lines...
Should look better in the tabular format.
Shall we merge it now, or wait for tabular format ready for hyperopt?

yazeed:
> That's fine, but the overall length is longer than the width of my terminal, so printed in two lines...
> Should look better in the tabular format.
> Shall we merge it now, or wait for tabular format ready for hyperopt?

Let me finalize this, it needs it's own tests to be fully compliant.

yazeed:
I don't know what happened here, I think I rebased to upstream/develop instead of local/develop before pushing.

yazeed:
Closing, and duplicating at #3008 as I've messed this one up.

Issue #2971, opened 2020-02-24
naming : difference between percent and ratio in the documentation


gaugau3000:
Hi,

Seems like ratio and percent are both used but i think represent a percentage in all cases (a ratio is like x/y ).

tradable_balance_ratio could be tradable_balance_percent ?
last_stake_amount_min_ratio could be last_stake_amount_min_percent ?

Maybe i missed something.

Have a nice day.

xmatthias:
well the (intended) difference between percent and ratio is a `*100`.


to pick up on one of your examples: 
`last_stake_amount_min_ratio=0.99` would be `last_stake_amount_min_percent=99` .

While using percent would be a possibility - when talking about percent, i'd assume a number between 0 and 100 - while we expect a number between 0 and 1. 
using the term Percent here would be misleading i think.

I don't know any value we use (even though they may be called `_percent` historically) which we expect between 0 and 100.
If you know a better term than ratio (i'm not certain it's correct) - i'm happy for suggestions - but percent is too misleading i think.

The only reference i found (this is difficult to google) is [this](https://english.stackexchange.com/questions/140120/a-better-expression-for-percentage-divided-by-100).
There's quite some discussion around this topic there ... one point is made in a comment to the first item:

> No, you would need to call it a fraction of the length. Percentages go from 0 to 100. Or, rewrite the function to use a percentage; or maybe even both (x < 1, it's a fraction; 1 < x < 100, it's a percentage)


maybe a better name would be fraction - but another comment says:

> In similar circumstances, the term I usually see is 'proportion', or sometimes 'ratio'. [...]

gaugau3000:
> well the (intended) difference between percent and ratio is a *100.

I am not sure a ratio is when you compare [two things](https://wikidiff.com/ratio/percentage]). You trade a percentage/fraction of a balance with is a sub-part of the balance there is no ratio as the balance is the sole variable.

> to pick up on one of your examples:
> last_stake_amount_min_ratio=0.99 would be last_stake_amount_min_percent=99 .

You are true (my bad)

I think fraction is the good term (the explanation in this  [link](https://stackoverflow.com/questions/8646521/term-for-percent-percentage-fraction-scale-factor) seems quite convincing).

[...]

PR #2989, opened 2020-02-28
No "percent" where "ratio" is to be used

hroff-1902:
partly addresses #2971, not in the docs, but in the code

No changes that touch user-visible interfaces in this part (so no changes in the names of the settings, fields in the backtesting dicts, columns in the database, only intrinsic vars/names).


freqtrade/edge/edge_positioning.py
```diff
@@ -399,9 +400,8 @@ def _detect_next_stop_or_sell_point(self, buy_column, sell_column, date_column,
                 # trade opens in reality on the next candle
                 open_trade_index += 1
 
-            stop_price_percentage = stoploss + 1
             open_price = ohlc_columns[open_trade_index, 0]
-            stop_price = (open_price * stop_price_percentage)
+            stop_price = (open_price * (stoploss + 1))
```
hroff-1902: @xmatthias pls review this separately (not related to the changes in this PR, a separate issue): stoploss is used here without `abs`...

xmatthias: not using `abs()` is fine in theory if we're sure the value is negative ... 

in this case however, the source for this stoploss is the loop over stoploss_range - and i currently struggle to find any code that ensures `stoploss_range_min` and `stoploss_range_max` must be negative (it's not in the config-validation either) - so i suspect you could run this with positive values - which would produce rather odd results ...

i'd leave that for another PR though ... not "really" related to this if we're honest ...

freqtrade/rpc/telegram.py
```diff
@@ -148,7 +148,7 @@ def send_msg(self, msg: Dict[str, Any]) -> None:
 
         elif msg['type'] == RPCMessageType.SELL_NOTIFICATION:
             msg['amount'] = round(msg['amount'], 8)
-            msg['profit_percent'] = round(msg['profit_percent'] * 100, 2)
+            msg['profit_percent'] = round(msg['profit_ratio'] * 100, 2)
```
xmatthias: Why change it on the right hand side, but not on the left? while it's passed to the message below - it's not visible to the user after it's converted to the message...

freqtrade/freqtradebot.py
```diff
@@ -1085,7 +1085,7 @@ def _notify_sell_cancel(self, trade: Trade, order_type: str) -> None:
             'open_rate': trade.open_rate,
             'current_rate': current_rate,
             'profit_amount': profit_trade,
-            'profit_percent': profit_percent,
+            'profit_ratio': profit_ratio,
```
xmatthias: If we change the dict key here - we'll also have to change it in `webhook-config.md`

freqtrade/freqtradebot.py
```diff
@@ -1048,7 +1048,7 @@ def _notify_sell(self, trade: Trade, order_type: str) -> None:
...
```
xmatthias: If we change the dict key here - we'll also have to change it in `webhook-config.md` - otherwise we document one thing, but it's no longer true ... and this is something we'll NEVER find unless someone runs into it :(

freqtrade/rpc/telegram.py
hroff-1902: yes, because `msg['profit_percent'] = `msg['profit_ratio'] * 100` here...

freqtrade/freqtradebot.py
hroff-1902: adjusted the tests and the docs to use `profit_ratio` name of the item instead of `profit_percent`.

The real problem with webhook is that it uses `.format` instead of f-string formatting, so I cannot  write "profit: {profit_amount:8f} {stake_currency} ({profit_ratio * 100}%)" in webhooksell, for example...

hroff-1902:
11 files changed... I expected this to be a tiny cosmetic PR... 😆 
(I specially did not change the db-persistence stuff, where `_pct` is also used when the value is the ratio...)

freqtrade/freqtradebot.py
hroff-1902: see below

freqtrade/freqtradebot.py
hroff-1902: As a workaround, we could insert both `profit_percent` and `profit_ratio` into the dict of values, prepared by `notify_sell()` in freqtradebot

xmatthias: you can't use f-strings for webhooks - at least i would not know how - as the f-string itself needs to be available beforehand.
i've never seen any example where the f-string itself comes from a variable - and while searching for that a while ago i found that that won't work.

hroff-1902: It's possible, without eval:
https://stackoverflow.com/a/57597617

xmatthias: well eval ... but using `eval()` on user-provided strings makes the hair on my neck stand (and run far far far away) ...

hroff-1902: > well eval ... but using `eval()` on user-provided strings makes the hair on my neck stand (and run far far far away) ...

agree, so just FYI, achievable if evals are allowed

Issue #2837, opened 2020-02-02
--print-trades to backtesting and hyperopt show-hyperopt


yazeed:
The purpose of this issue is to log the feature request mentioned in #2727 in its own issue, where it was discussed the possibility of supporting a flag **--print-trades** with the following commands:

- hyperopt show-hyperopt --print-trades
- backtesting --print-trades

The flag would print a table of the trades.

Another possibility is to have it export them into a CSV file with **--export-trades**, since some some outputs may have too many trades to print in the console.

Feel free to discuss here which one you'd prefer.

xmatthias:
I'm not a huge fan of printing the list of trades to screen.
If it's just a few this may be fine - but you won't know this before running backtesting, so you'll have confusing printouts almost guaranteed.

We provide [sample code](https://www.freqtrade.io/en/latest/strategy_analysis_example/#load-backtest-results-to-pandas-dataframe) to load the results to a dataframe - which i think provides a lot more flexibility - as it'll allow you to do analysis on top of the data - what exactly would you expect to read from printing 20 trades? you wouldn't even know if they averaged positive or negative ...

> Another possibility is to have it export them into a CSV file with --export-trades

What's the difference to exporting to CSV as opposed to exporting to json?

We even provide code to load the json to a pandas dataframe - so to me, a simple `results.to_csv()` in a jupyter notebook will suffice this (i see this as a very edgecase requirement).

hroff-1902:
> The purpose of this issue is to log the feature request mentioned in #2727 in its own issue, where it was discussed the possibility of supporting a flag **--print-trades** with the following commands:
> 
> * hyperopt show-hyperopt --print-trades
> * backtesting --print-trades
> 
> The flag would print a table of the trades.

Can you propose the columns for that table? (Or simply offer a PR containing proposal, if it's not heavy)

> Another possibility is to have it export them into a CSV file with **--export-trades**, since some some outputs may have too many trades to print in the console.

For backtesting, is not it same with `--export trades`? Should we change it into `--export-trades`, the syntax of `--export` raises question for myself for long time? Same syntax should simply be implemented into `show-hyperopt`.

> Feel free to discuss here which one you'd prefer.

Both (--print-trades and export) are handy and can coexist imho.

xmatthias:
> Both (--print-trades and export) are handy and can coexist imho.

Export - yes - it'll provide you with possibilities to analyze the data in detail - but json is fine for that in my opinion.

[...]

PR #2994, opened 2020-02-28
Added dynamic print table function to hyperopt

Fredrik81:
## Summary
Change print output of hyperopt to table style. More space efficient.

WIP

Part of issue: #2913 
Part of issue: #2837 
Part of issue: #1775 

## Quick changelog

- Changed default output to table style
- Small modifications to print table function

## What's new?
New output example:
```
2020-02-29 23:16:56,838 - freqtrade.optimize.hyperopt - INFO - Effective number of parallel workers used: 8
+--------+---------+----------+--------------+-----------------------------+----------------+-------------+
|   Best |   Epoch |   Trades |   Avg profit |                      Profit |   Avg duration |   Objective |
|--------+---------+----------+--------------+-----------------------------+----------------+-------------|
|      * |    1/10 |        0 |           -- |                          -- |             -- |         N/A |
|      * |    2/10 |        0 |           -- |                          -- |             -- |         N/A |
|   Best |    3/10 |        2 |        0.37% |      0.00004498 BTC (0.75%) |         32.5 m |       1.899 |
2020-03-03 23:56:20,801 - freqtrade.optimize.hyperopt - INFO - Saving 3 epochs.
|      * |    4/10 |        0 |           -- |                          -- |             -- |         N/A |
|      * |    5/10 |        0 |           -- |                          -- |             -- |         N/A |
|      * |    6/10 |      110 |       -0.04% |    -0.00027303 BTC (-4.55%) |        186.7 m |     2.09325 |
|      * |    7/10 |        0 |           -- |                          -- |             -- |         N/A |
|      * |    8/10 |        0 |           -- |                          -- |             -- |         N/A |
|      * |    9/10 |        0 |           -- |                          -- |             -- |         N/A |
|      * |   10/10 |        0 |           -- |                          -- |             -- |         N/A |
```

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -21,7 +21,7 @@ from colorama import init as colorama_init
 from joblib import (Parallel, cpu_count, delayed, dump, load,
                     wrap_non_picklable_objects)
 from pandas import DataFrame, json_normalize, isna
-from tabulate import tabulate
+import tabulate
 
 from freqtrade.data.converter import trim_dataframe
 from freqtrade.data.history import get_timerange
@@ -309,6 +309,8 @@ class Hyperopt:
         if not results:
             return
 
+        tabulate.PRESERVE_WHITESPACE = True
+
         trials = json_normalize(results, max_level=1)
         trials['Best'] = ''
         trials = trials[['Best', 'current_epoch', 'results_metrics.trade_count',
@@ -325,43 +327,51 @@ class Hyperopt:
         trials['Trades'] = trials['Trades'].astype(str)
 
         trials['Epoch'] = trials['Epoch'].apply(
-            lambda x: '{}/{}'.format(str(x).rjust(len(str(total_epochs)), ' '), total_epochs))
+            lambda x: '{}/{}'.format(str(x).rjust(len(str(total_epochs)), ' '), total_epochs)
+        )
         trials['Avg profit'] = trials['Avg profit'].apply(
-            lambda x: ('{:,.2f}%'.format(x)).rjust(7, ' ') if not isna(x) else "--".rjust(7, ' '))
-        trials['Profit'] = trials['Profit'].apply(
-            lambda x: ('{:,.2f}%'.format(x)) if not isna(x) else "--")
-        trials['Total profit'] = trials['Total profit'].apply(
-            lambda x: ('{:,.8f} '.format(x)) + config['stake_currency'] if not isna(x) else "--")
+            lambda x: ('{:,.2f}%'.format(x)).rjust(7, ' ') if not isna(x) else "--".rjust(7, ' ')
+        )
         trials['Avg duration'] = trials['Avg duration'].apply(
-            lambda x: ('{:,.1f}m'.format(x)).rjust(7, ' ') if not isna(x) else "--".rjust(7, ' '))
+            lambda x: ('{:,.1f} m'.format(x)).rjust(7, ' ') if not isna(x) else "--".rjust(7, ' ')
+        )
         trials['Objective'] = trials['Objective'].apply(
-            lambda x: str(x).rjust(10, ' ') if str(x) != str(100000) else "N/A".rjust(10, ' '))
-        trials['Profit'] = trials['Total profit'] + " (" + trials['Profit'] + ")"
+            lambda x: str(x).rjust(10, ' ') if str(x) != str(100000) else "N/A".rjust(10, ' ')
+        )
+
+        trials['Profit'] = trials.apply(
+            lambda x: '{:,.8f} {} ({:,.2f}%)'.format(
+                x['Total profit'], config['stake_currency'],
+                x['Profit']).rjust(24+len(config['stake_currency']))
+            if x['Total profit'] != 0.0 else '--'.rjust(24+len(config['stake_currency'])),
+            axis=1
+        )
         trials = trials.drop(columns=['Total profit'])
 
         if print_colorized:
             for i in range(len(trials)):
                 if trials.loc[i]['is_profit']:
                     for z in range(len(trials.loc[i])-3):
-                        trials.iat[i, z] = "{}{}{}".format(Fore.GREEN,
-                                                           str(trials.loc[i][z]), Fore.RESET)
+                        trials.iat[i, z] = "{}{}{}".format(
+                            Fore.GREEN, str(trials.loc[i][z]), Fore.RESET)
                 if trials.loc[i]['is_best'] and highlight_best:
                     for z in range(len(trials.loc[i])-3):
-                        trials.iat[i, z] = "{}{}{}".format(Style.BRIGHT,
-                                                           str(trials.loc[i][z]), Style.RESET_ALL)
+                        trials.iat[i, z] = "{}{}{}".format(
+                            Style.BRIGHT, str(trials.loc[i][z]), Style.RESET_ALL)
 
         trials = trials.drop(columns=['is_initial_point', 'is_best', 'is_profit'])
         if remove_header > 0:
-            table = tabulate(trials.to_dict(orient='list'), tablefmt='orgtbl',
-                             headers='keys', stralign="right")
+            table = tabulate.tabulate(
+                trials.to_dict(orient='list'), tablefmt='orgtbl', headers='keys', stralign="right")
+            # print(table)
             table = table.split("\n", remove_header)[remove_header]
         elif remove_header < 0:
-            table = tabulate(trials.to_dict(orient='list'), tablefmt='psql',
-                             headers='keys', stralign="right")
+            table = tabulate.tabulate(
+                trials.to_dict(orient='list'), tablefmt='psql', headers='keys', stralign="right")
             table = "\n".join(table.split("\n")[0:remove_header])
         else:
-            table = tabulate(trials.to_dict(orient='list'), tablefmt='psql',
-                             headers='keys', stralign="right")
+            table = tabulate.tabulate(
+                trials.to_dict(orient='list'), tablefmt='psql', headers='keys', stralign="right")
         print(table)
 
     def has_space(self, space: str) -> bool:
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -325,15 +325,20 @@ class Hyperopt:
         trials['Trades'] = trials['Trades'].astype(str)
 
         trials['Epoch'] = trials['Epoch'].apply(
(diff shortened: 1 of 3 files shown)


hroff-1902:
I think no horizontal separators should be printed between data rows, only below the header...

Fredrik81:
Well that would take it a step away from the other tables. Should there be
inconsistencies?

lör 29 feb. 2020 kl. 01:08 skrev hroff-1902 <notifications@github.com>:

hroff-1902:
maybe I missed this in hyperopt-list...

when data are printed for backtesting, rows correspond to the pairs. strategies and it's okay to separate it with lines. 

But when we have a long list of epochs -- here, or in the hyperopt-list -- separating lines seems to be redundant for me, this just makes the long list longer. Same when we'll print the list of simulated trades for backtesting -- in that case, the lines should not also be printed after each trade...

other frames, surrounding the headers, should be same...

just imho

Fredrik81:
What ever the outcome is i think it's good if it's one standard for consistency so we should at then change the other tables. Make sense?

xmatthias:
It's not part of the `hyperopt-list` command... which looks like the below.

```
+--------+---------+----------+--------------+-----------------+----------+----------------+-------------+
|   Best |   Epoch |   Trades |   Avg profit |    Total profit |   Profit |   Avg duration |   Objective |
|--------+---------+----------+--------------+-----------------+----------+----------------+-------------|
|   Best |     1/8 |       37 |       -1.93% | -0.07151890 ETH |  -71.45% |       3,471.9m |    6.1965   |
|   Best |     2/8 |        4 |        0.91% |  0.00362863 ETH |    3.62% |         435.0m |   -3.70232  |
|      * |     3/8 |        3 |       -3.22% | -0.00965683 ETH |   -9.65% |       1,140.0m |    2.6835   |
|   Best |     4/8 |       43 |        2.08% |  0.08952263 ETH |   89.43% |       2,422.3m |  -14.1451   |
|      * |     5/8 |       23 |        1.30% |  0.02990233 ETH |   29.87% |       1,935.7m |   -7.30637  |
|      * |     6/8 |       32 |        0.63% |  0.02006282 ETH |   20.04% |       2,321.2m |   -2.52548  |
|      * |     7/8 |       10 |       -3.69% | -0.03691115 ETH |  -36.87% |       2,214.0m |    4.50962  |
|      * |     8/8 |       19 |        0.11% |  0.00208357 ETH |    2.08% |         887.4m |   -0.245796 |
+--------+---------+----------+--------------+-----------------+----------+----------------+-------------+

```

The problem here is what i mentioned before ... you're printing this in a loop - so you remove the header - but i think you'll also need to remove the "bottom line" to avoid having the output unnecessarily long.

Fredrik81:
Ahh yeah your correct. I guess I overlooked this. Will change this for next
commit.
Thanks.

lör 29 feb. 2020 kl. 15:12 skrev Matthias <notifications@github.com>:

yazeed:
> Total profit |   Profit

This PR proposes separating Total Profit from Profit % into two columns **-0.07151890 ETH |  -71.45%**.

I think we should stick with our existing approach of using them next to each other as follows: **-0.07151890 ETH (-71.45%)**, which is how they're used everywhere in hyperopt/backtesting/rpc messaging.

Fredrik81:
> > Total profit |   Profit
> 
> This PR proposes separating Total Profit from Profit % into two columns **-0.07151890 ETH | -71.45%**.
> 
> I think we should stick with our existing approach of using them next to each other as follows: **-0.07151890 ETH (-71.45%)**, which is how they're used everywhere in hyperopt/backtesting/rpc messaging.

Changed with latest commit

tests/optimize/test_hyperopt.py
```diff
@@ -409,11 +409,6 @@ def test_log_results_if_loss_improves(hyperopt, capsys) -> None:
         }
     )
     out, err = capsys.readouterr()
-    result_str = (
-        '|   Best |     2/2 |        1 |        0.10% | 0.00100000 BTC |'
```
yazeed: I was thinking it would look more like this.
`0.00100000 BTC (1.00%)`

freqtrade/optimize/hyperopt.py
```diff
@@ -325,15 +325,20 @@ def print_result_table(config: dict, results: list, total_epochs: int, highlight
...
+            lambda x: ('{:,.2f}%'.format(x)) if not isna(x) else "--")
         trials['Total profit'] = trials['Total profit'].apply(
-            lambda x: '{:,.8f} '.format(x) + config['stake_currency'] if not isna(x) else x)
+            lambda x: ('{:,.8f} '.format(x)) + config['stake_currency'] if not isna(x) else "--")
         trials['Avg duration'] = trials['Avg duration'].apply(
-            lambda x: '{:,.1f}m'.format(x) if not isna(x) else x)
+            lambda x: ('{:,.1f}m'.format(x)).rjust(7, ' ') if not isna(x) else "--".rjust(7, ' '))
+        trials['Objective'] = trials['Objective'].apply(
+            lambda x: str(x).rjust(10, ' ') if str(x) != str(100000) else "N/A".rjust(10, ' '))
+        trials['Profit'] = trials['Total profit'] + " (" + trials['Profit'] + ")"
```
yazeed: This line is missing the percentage symbol, as follows:
`         trials['Profit'] = trials['Total profit'] + " (" + trials['Profit'] + "%)" `

tests/optimize/test_hyperopt.py
```diff
@@ -413,7 +413,9 @@ def test_log_results_if_loss_improves(hyperopt, capsys) -> None:
         '|   Best |     2/2 |        1 |        0.10% | 0.00100000 BTC |'
```
yazeed: I was thinking it would look more like this.
`0.00100000 BTC (1.00%)`

yazeed:
One final note about this from my side, is that we don't actually need the decimal point on avg duration.

xmatthias:
> One final note about this from my side, is that we don't actually need the decimal point on avg duration. On here, and on hyperopt-list's table.

Just because you mention it - i guess avg duration should be aligned to backtesting output which shows it as hh:mm:ss. 
Feel free to leave that for the future though - it's not there at the moment, so i don't see that as a requirement here.

hroff-1902:
* acording to English typographical rules, there should be a blank space between value and unit symbol: `435.0 m`

Probably not for this PR (and to be discussed):
* digits after decimal dot makes sense: `503 m` but `10.2 m`
* I would probably like to see duration in both minutes and in the human-readable form: `603 m (10 h 3m)` or vice-versa `10 h 3m (603 m)`

freqtrade/optimize/hyperopt.py
Fredrik81: This is done on line 332:
```
        trials['Profit'] = trials['Profit'].apply(
            lambda x: ('{:,.2f}%'.format(x)) if not isna(x) else "--")
```

tests/optimize/test_hyperopt.py
Fredrik81: Already changed in latest commit.
```
|        | 2758/5000 |      143 |        0.14% | 0.00118028 BTC ( 19.65%) |         176.5m |     1.99024 |
```

Fredrik81: Agree. This is still WIP so will make clean up when we are all happy with the outputs :-)

Fredrik81:
> * according to English typographical rules, there should be a blank space between value and unit symbol: `435.0 m`
> 
> Probably not for this PR (and to be discussed):
> 
> * digits after decimal dot makes sense: `503 m` but `10.2 m`
> * I would probably like to see duration in both minutes and in the human-readable form: `603 m (10 h 3m)` or vice-versa `10 h 3m (603 m)`, or `10:03 (603 m)`
> 
> Also agree, this should be aligned to backtesting output (maybe something should be changed in the backtesting output, instead of here, if we decide this).

We are moving to table format but we should still be careful about the length. I like the human readable but it might steal to much terminal space.

hroff-1902:
> We are moving to table format but we should still be careful about the length. I like the human readable but it might steal to much terminal space.

Agree, that's also the point which is to be taken into consideration...

Fredrik81:
Layout and alignment should be better now. Anything else we should change?
Else i think it's only some smaller cleanup/optimizations left

yazeed:
> Layout and alignment should be better now. Anything else we should change?
> Else i think it's only some smaller cleanup/optimizations left

Looks almost there to me, just need to get rid of that decimal on Avg Duration.

hroff-1902:
> Looks almost there to me, just need to get rid of that decimal on Avg Duration.

At 1m and 5m timeframes the duration can be short, so `3 m` is not informative enough as `3.2 m`

hroff-1902:
I would like to have the width for the Profit column to be a bit wider, to prevent this kind of output when the values do not fit the cell:
```
|        |  82/500 |       34 |        0.23% |     15.79456349 BTC (7.89%) |      1,052.5 m |   -0.725037 |
|        |  83/500 |       33 |        0.57% |    37.78094238 BTC (18.87%) |        709.5 m |     -4.9243 |
|        |  84/500 |     1392 |       -0.25% | -693.69940419 BTC (-346.50%) |         74.7 m |     38.6772 |
|        |  85/500 |       47 |        0.52% |    49.16960956 BTC (24.56%) |        241.5 m |    -4.70543 |
|        |  86/500 |      659 |       -0.28% | -364.26968706 BTC (-181.95%) |         74.3 m |     19.7186 |
|        |  87/500 |      132 |       -0.07% |   -19.72120573 BTC (-9.85%) |        718.1 m |     1.81839 |
|   Best |  88/500 |       75 |        0.52% |    77.60263182 BTC (38.76%) |        201.1 m |    -11.4273 |
```

xmatthias:
I think that's a problem / side-effect of not having the whole table available at startup  (otherwise tabulate handles this quite nicely).

how wide is the limit? 

in theory - you could have -1000.55% loss on USDT stake currency - so that's 2 longer than the longest above ... 
That however will go against the maximum number of columns that'll fit onto row ...

hroff-1902:
> I think that's a problem / side-effect of not having the whole table available at startup (otherwise tabulate handles this quite nicely).

right

> in theory - you could have -1000.55% loss on USDT stake currency - so that's 2 longer than the longest above ...

right, but `-1000.55%` will happen very rarely in the output, while profits and especially losses >100% can be found frequently, producing skewed table...

Fredrik81:
I will work something in.. but for now i would like that we keep the "123.12 m" format for avg time.
This type is shown in a few places and i think it would make more sense to make another PR later to change time prints for all things to keep it more consistent.
OK?

PR #2973, opened 2020-02-25
Support non pairs

xmatthias:
## Summary
Certain markets are not in the format of quote/base.

This PR allows to extract either side from the markets dict for each pair.

Closes #268
Closes #1107 (the bot should now support inversed pairs too)
most likely will close #2969 (still needs validation / confirmation)

## Quick changelog

- move pairlist validation for correct stake currency to exchange (it requires markets now)
- remove all usages of `pair.startswith()`, `pair.endswith()` and `pair.split('/')` and replace them with the methods `get_pair_quote_currency()` and `get_pair_base_currency()`
- Remove config validation to make sure pairs are `base/quote` formatted.

PR #2995, opened 2020-02-29
Allow Stake currency empty when using download-data

xmatthias:
## Summary
Fixes a bug mentioned on slack, where running download-data without stake-currency did fail again (reintroduced in #2973).

## Quick changelog

- Test + fix allowing empty stake-currency

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -332,7 +332,8 @@ class Exchange:
                 logger.warning(f"Pair {pair} is restricted for some users on this exchange."
                                f"Please check if you are impacted by this restriction "
                                f"on the exchange and eventually remove {pair} from your whitelist.")
-            if not self.get_pair_quote_currency(pair) == self._config['stake_currency']:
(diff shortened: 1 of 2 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -332,7 +332,8 @@ def validate_pairs(self, pairs: List[str]) -> None:
                 logger.warning(f"Pair {pair} is restricted for some users on this exchange."
                                f"Please check if you are impacted by this restriction "
                                f"on the exchange and eventually remove {pair} from your whitelist.")
-            if not self.get_pair_quote_currency(pair) == self._config['stake_currency']:
+            if (self._config['stake_currency'] and
+                    not self.get_pair_quote_currency(pair) == self._config['stake_currency']):
```
hroff-1902: why not `self.get_pair_quote_currency(pair) != self._config['stake_currency']`?

xmatthias: well THAT is a different topic :D 
great catch :+1:

PR #3003, opened 2020-02-29
Hyperopt: fix number of CPU cores, jobs and total epochs

Fredrik81:
## Summary
Hyperopt fix when specifying total epochs less than amount of CPU cores.

PEP8 = yes

## Quick changelog

- Hyperopt.py, aligning total epochs and cpu core config

## What's new?
Aligning the parameters when less epochs are specified will avoid the situation below:
```
Parameter --epochs detected ... Will run Hyperopt with for 1 epochs ...
Parameter -j/--job-workers detected: -1
- INFO - Effective number of parallel workers used: 8
+--------+---------+----------+--------------+----------------+----------+----------------+-------------+
|   Best |   Epoch |   Trades |   Avg profit |   Total profit |   Profit |   Avg duration |   Objective |
|--------+---------+----------+--------------+----------------+----------+----------------+-------------|
|      * |     1/1 |        0 |          nan | 0.00000000 BTC |    0.00% |            nan |      100000 |
|      * |     2/1 |        0 |          nan | 0.00000000 BTC |    0.00% |            nan |      100000 |
|   Best |     3/1 |        3 |        0.46% | 0.00008214 BTC |    1.37% |          18.3m |     1.87778 |
2020-02-29 23:17:00,051 - freqtrade.optimize.hyperopt - INFO - Saving 3 epochs.
|      * |     4/1 |        0 |          nan | 0.00000000 BTC |    0.00% |            nan |      100000 |
|      * |     5/1 |        0 |          nan | 0.00000000 BTC |    0.00% |            nan |      100000 |
|      * |     6/1 |        0 |          nan | 0.00000000 BTC |    0.00% |            nan |      100000 |
|      * |     7/1 |        0 |          nan | 0.00000000 BTC |    0.00% |            nan |      100000 |
|      * |     8/1 |        0 |          nan | 0.00000000 BTC |    0.00% |            nan |      100000 |
2020-02-29 23:17:00,109 - freqtrade.optimize.hyperopt - INFO - Saving 8 epochs.
```

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -558,7 +558,8 @@ class Hyperopt:
         cpus = cpu_count()
         logger.info(f"Found {cpus} CPU cores. Let's make them scream!")
         config_jobs = self.config.get('hyperopt_jobs', -1)
-        if self.total_epochs < cpus and (config_jobs > self.total_epochs or config_jobs < 0):
+        if (config_jobs < 0 and (cpus + config_jobs + 1) > self.total_epochs) \
+           or (config_jobs > 0 and config_jobs > self.total_epochs):
             config_jobs = self.total_epochs
         logger.info(f'Number of parallel jobs set as: {config_jobs}')
 
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -572,15 +572,15 @@ class Hyperopt:
                 logger.info(f'Effective number of parallel workers used: {jobs}')
                 EVALS = ceil(self.total_epochs / jobs)
                 for i in range(EVALS):
-                    asked = self.opt.ask(n_points=jobs)
+                    # Correct the number of epochs to be processed for the last
+                    # iteration (should not exceed self.total_epochs in total)
+                    n_rest = (i + 1) * jobs - self.total_epochs
+                    current_jobs = jobs - n_rest if n_rest > 0 else jobs
+
+                    asked = self.opt.ask(n_points=current_jobs)
                     f_val = self.run_optimizer_parallel(parallel, asked, i)
                     self.opt.tell(asked, [v['loss'] for v in f_val])
                     self.fix_optimizer_models_list()
-
-                    # Correct the number of epochs to handled for the last
-                    # iteration (should not exceed self.total_epochs)
-                    n_rest = (i + 1) * jobs - self.total_epochs
-                    current_jobs = jobs - n_rest if n_rest > 0 else jobs
 
                     for j in range(current_jobs):
                         # Use human-friendly indexes here (starting from 1)
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -9,7 +9,7 @@ import logging
 import random
 import sys
 import warnings
-from math import ceil, floor
+from math import ceil
 from collections import OrderedDict
 from operator import itemgetter
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


xmatthias:
wait, you're saying running hyperopt for fewer than "cores" will have it run for "cores" anyway? :laughing:

hroff-1902:
@xmatthias he is (partly) right: currently, if `-j 4` and `-e 1`, hyperopt will run for 4 epochs instead of 1.

@Fredrik81 this should be fixed some another way, though: if `-j 4` and `-e 10`, hyperopt will run for 12 epochs, while it should run for 12... That's the real problem, it's not only about the case when the value of `-e` is less than `-j` or CPU...

It was already fixed in #1765 (and now in #2951) IIRC, I was lazy to pick the fix from that old PR, you can refer to it and implement and here...

Fredrik81:
> @xmatthias he is (partly) right: currently, if `-j 4` and `-e 1`, hyperopt will run for 4 epochs instead of 1.
> 
> @Fredrik81 this should be fixed some another way, though: if `-j 4` and `-e 10`, hyperopt will run for 12 epochs, while it should run for 12... That's the real problem, it's not only about the case when the value of `-e` is less than `-j` or CPU...
> 
> It was already fixed in #1765 (and now in #2951) IIRC, I was lazy to pick the fix from that old PR, you can refer to it and implement and here...

Yeah you are right this only is a workaround for part of the problem.. all tho i think this actually have it's place as the text will show the correct text for the corner case it fixes..
```
Parameter --epochs detected ... Will run Hyperopt with for 1 epochs ...
Parameter -j/--job-workers detected: -1
- INFO - Effective number of parallel workers used: 8     <- should say one as it only runs one?
```
i will look at the other PR's and add to it here.

Fredrik81:
@xmatthias @hroff-1902 I'm pretty happy with this PR. Can you please review?
it will solve both situations -j 4 -e 1 and -j 4 -e 10

freqtrade/optimize/hyperopt.py
```diff
@@ -557,6 +558,8 @@ def start(self) -> None:
         cpus = cpu_count()
         logger.info(f"Found {cpus} CPU cores. Let's make them scream!")
         config_jobs = self.config.get('hyperopt_jobs', -1)
+        if self.total_epochs < cpus and (config_jobs > self.total_epochs or config_jobs < 0):
+            config_jobs = self.total_epochs
```
hroff-1902: Small issue with this:
assume we have 8 cpus, total_epochs=7 and config_jobs=-3. This setting of config_jobs assumes, that the number of workers will be set to 6 (8+1-3), leaving 2 cpus free for the system. With your code this will be set to 7.
Same for 4 cpus, total_epochs=3 and also config_jobs=-3. config_jobs should be set to 2 (or left as -3).

I propose not changing negative values of config_jobs (but this should be properly handles below for the last loop iteration)...

Fredrik81: Fixed the logic to only change in case the value is not OK also for the scenario you described:
```
        if (config_jobs < 0 and (cpus + config_jobs + 1) > self.total_epochs) \
           or (config_jobs > 0 and config_jobs > self.total_epochs):
            config_jobs = self.total_epochs
```

hroff-1902:
@Fredrik81 pls review https://github.com/freqtrade/freqtrade/pull/3003/commits/92425642da922740e319f9563c93a36873af946f: this is not really needed, since the real number of workers (effective number, as it's called here and in the joblib docs) is later fetched from the properties of the parallel object; so this value is only needed to initialize parallel object, and it's better to pass it as it is to keep semantics (try to set `-j -6`, for example, on a 4 cpu machine... Let's leave processing it to joblib, this is not actually used later in the code).

PR #3008, opened 2020-03-02
Wins/draws/losses/median profit in hyperopt output

### LINKED ISSUE #3023: Columns in hyperopt output
Let's discuss improvements to hyperopt (and possibly to backtesting, to be aligned) column headers.

Current:
```
+--------+---------+----------+--------------+---------------------------+----------------+-------------+
|   Best |   Epoch |   Trades |   Avg profit |                    Profit |   Avg duration |   Objective |
|--------+---------+----------+--------------+---------------------------+----------------+-------------|
|   Best |   1/100 |       21 |       -0.07% |   -2.94404909    (-1.47%) |        376.4 m |     2.20748 |
|   Best |   2/100 |      354 |       -0.01% |  -10.29255100    (-5.14%) |        131.2 m |     1.81064 |
|   Best |   3/100 |      180 |        0.01% |    4.57664850     (2.29%) |        695.7 m |     0.40082 |
|      * |   4/100 |        2 |        0.00% |                        -- |      1,377.5 m |         N/A |
|      * |   5/100 |       24 |       -0.33% |  -15.71527070    (-7.85%) |        628.3 m |     2.97213 |
```

Proposed as the starting point:
```
+--------+---------+--------------+------------+---------------------------+----------------+--------------+
|        |   Epoch |       Trades | Avg profit |             Total profit, |      Avg trade | Optimization |
|        |         |      (W/D/L) |  per trade |     absolute (percentage) |       duration |    objective |
|--------+---------+--------------+------------+---------------------------+----------------+--------------|
|   Best |   1/100 | 21 (10/1/10) |     -0.07% |   -2.94404909    (-1.47%) |        376.4 m |      2.20748 |
|   Best |   2/100 |          354 |     -0.01% |  -10.29255100    (-5.14%) |        131.2 m |      1.81064 |
|        |         |  (200/4/150) |            |                           |                |              |
|   Best |   3/100 |          180 |      0.01% |    4.57664850     (2.29%) |        695.7 m |      0.40082 |
|      * |   4/100 |            2 |      0.00% |                        -- |      1,377.5 m |          N/A |
|      * |   5/100 |           24 |     -0.33% |  -15.71527070    (-7.85%) |        628.3 m |      2.97213 |
```

yazeed:
> This PR suggests a couple of additional info to the hyperopt output (wins/draws/losses, and median profit which kinda shows if your avg profit is skewed due to lucky trades in that epoch, it's a simple way to detect outlier trades in an epoch), it should support #2943 if that PR is merged into develop.
> 
> Let me know your thoughts.
> 
> Will work on the tests once thoughts are finalized into this.

This PR duplicates #2978 as I've done a bad rebase there. Should be ready to merge.

closes #3023


xmatthias:
To quote @hroff-1902 from https://github.com/freqtrade/freqtrade/pull/2978#issuecomment-592506507

> That's fine, but the overall length is longer than the width of my terminal, so printed in two lines...

I do see the same problem. Also, i think #2994  will make the problem worse by using a table ... and multi-line tables will look very strange here.

Honestly, that's the reason why i've always been reluctant to add more metrics to both backtesting and hyperopt - because you cannot assume everyone has a 4k ultrawide screen ... some people use regular screens (I've often tried to assume a default putty window - which is 80 columns / characters in width) - but i guess we're past that point with backtesting already.

2020-03-02-194411_1098x344_scrot

hroff-1902:
length should be okay after #2994

hroff-1902:
length should be okay after #2994 ...

hroff-1902:
my proposal (and for backtesting as well):

print those values in one column, like
```
Wins/Draws/Losses
----------
7/2/15
8/0/3
...
```
or even in one column with total number of trades:
```
Trades
(Wins/Draws/Losses)
----------
24 (7/2/15)
11 (8/0/3)
...
```

yazeed:
> ```
> Trades
> (Wins/Draws/Losses)
> ----------
> 24 (7/2/15)
> 11 (8/0/3)
> ...
> ```

I like this, let me work on an alteration for hyperopt, I'll work on the backtesting PR later.

> To quote @hroff-1902 from [#2978 (comment)](https://github.com/freqtrade/freqtrade/pull/2978#issuecomment-592506507)
> 
> > That's fine, but the overall length is longer than the width of my terminal, so printed in two lines...
> 
> I do see the same problem. Also, i think #2994 will make the problem worse by using a table ... and multi-line tables will look very strange here.
> 
> Honestly, that's the reason why i've always been reluctant to add more metrics to both backtesting and hyperopt - because you cannot assume everyone has a 4k ultrawide screen ... some people use regular screens (I've often tried to assume a default putty window - which is 80 columns / characters in width) - but i guess we're past that point with backtesting already.
> 
> 2020-03-02-194411_1098x344_scrot

On a second look, I agree with you on 80 columns for tables, it makes total sense, but not quite of a necessity for verbose text prints like these. If you look at your screenshot it remains very readable.

Fredrik81:
> Trades
> (Wins/Draws/Losses)
> ----------
> 24 (7/2/15)
> 11 (8/0/3)

This is also my thoughts for the table style

orehunt:
I think that any kind of results metrics should be put into a parent loss class, and hyperopt should just give to the loss function:
- the backtest results
- the whole dataframe...why not
So those metrics are only run if the loss class requires them,
however since it also modifies the ouput, the loss class should also provide an optional different output method

yazeed:
> ```
> Trades
> (Wins/Draws/Losses)
> ----------
> 24 (7/2/15)
> 11 (8/0/3)
> ```
> 
> This is also my thoughts for the table style

Agree to this.

yazeed:
> I think that any kind of results metrics should be put into a parent loss class, and hyperopt should just give to the loss function:
> 
>     * the backtest results
> 
>     * the whole dataframe...why not
> 
> 
> So those metrics are only run if the loss class requires them,
> however since it also modifies the ouput, the loss class should also provide an optional different output method

I think this is a good idea in line with giving users more control. Though, I think it's suitable for a new PR to implement the interface for hyperopt output/print functionally.

hroff-1902:
@yazeed #2994 is merged, can you adapt the output to the new tabular format (and using the format `24 (7/2/15)` discussed above)?

yazeed:
> @yazeed #2994 is merged, can you adapt the output to the new tabular format (and using the format `24 (7/2/15)` discussed above)?

Will work on it and post back.

xmatthias:
you really really need to get a hang on git merge ... 

i think the only "relevant" commits are the last 3 commits (?)

xmatthias:
Results will now be shown as follows:
```
+--------+---------+----------+---------+--------------+------------------------------+----------------+-------------+                
|   Best |   Epoch |   Trades |   W/D/L |   Avg profit |                       Profit |   Avg duration |   Objective |
|--------+---------+----------+---------+--------------+------------------------------+----------------+-------------|
| * Best |   2/100 |        1 |   0/0/1 |       -3.31% |   -0.66304000 EUR   (-3.31%) |         15.0 m |     1.46961 |
| * Best |  10/100 |        7 |   5/0/2 |        0.31% |    0.43281416 EUR    (2.16%) |        336.4 m |      -0.426 |
```

In case of `list-hyperopt` running on a older version, the W/D/L column is replaced with `N/A` - indicating that this metric is not available.

PR #3021, opened 2020-03-05
Hyperopt: Add export CSV-file option

Fredrik81:
## Summary
Add options for exporting CSV-file from hyperopt

Work in progress.
Tests are missing at the moment.

## Quick changelog

- New command to hyperopt-list: --export-csv FILE

## What's new?
freqtrade hyperopt-list --export-csv /tmp/test.csv
or
<s>freqtrade hyperopt-list --export-csv +/tmp/test.csv</s> (this functionality refused; edited by @hroff-1902)

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -394,13 +394,13 @@ class Hyperopt:
 
         # Verification for overwrite
         if path.isfile(csv_file):
-            logging.error("CSV-File already exists!")
+            logger.error("CSV-File already exists!")
             return
 
         try:
             io.open(csv_file, 'w+').close()
         except IOError:
-            logging.error("Filed to create/overwrite CSV-File!")
+            logger.error("Filed to create CSV-File!")
             return
 
         trials = json_normalize(results, max_level=1)
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -23,6 +23,8 @@ from joblib import (Parallel, cpu_count, delayed, dump, load,
                     wrap_non_picklable_objects)
 from pandas import DataFrame, json_normalize, isna
 import tabulate
+from os import path
+import io
 
 from freqtrade.data.converter import trim_dataframe
 from freqtrade.data.history import get_timerange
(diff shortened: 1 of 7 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -381,6 +383,62 @@ def print_result_table(config: dict, results: list, total_epochs: int, highlight
...
+    @staticmethod
+    def export_csv_file(config: dict, results: list, total_epochs: int, highlight_best: bool,
+                        csv_file: str, overwrite: bool) -> None:
+        """
+        Log result to csv-file
+        """
+        if not results:
+            return
+
+        # Verification for owerwrite
```
hroff-1902:
```suggestion
        # Verification for overwrite
```

Fredrik81: Fixed in other commit.. thanks

hroff-1902:
@xmatthias I would like to see your comment here about the "+/tmp/test.csv" syntax (the '+' sign).
We do not use it with other options accepting file names, so if this will be implemented here, we'll than need to add same to other options, for better user experience (same syntax and same behavior)...

So I have doubts about this...

hroff-1902:
^^ the `+/tmp/test.csv` syntax discussed at Slack (core-dev channel), refused

hroff-1902:
https://github.com/freqtrade/freqtrade/pull/3021/commits/90ba9055e181d65a860e4b476ce4aa84478e1ec8 is wrong (compare with output of current develop)

freqtrade/optimize/hyperopt.py
```diff
@@ -381,6 +383,62 @@ def print_result_table(config: dict, results: list, total_epochs: int, highlight
...
+                        csv_file: str) -> None:
+        """
+        Log result to csv-file
+        """
+        if not results:
+            return
+
+        # Verification for overwrite
+        if path.isfile(csv_file):
+            logging.error("CSV-File already exists!")
```
hroff-1902:
```suggestion
            logger.error("CSV-File already exists!")
```
'logging' does not print the module issuing the message properly...
same below (check for other appearances)

Fredrik81: Hmm yes that was a miss.. strange that it worked :-)

freqtrade/optimize/hyperopt.py
```diff
@@ -381,6 +383,62 @@ def print_result_table(config: dict, results: list, total_epochs: int, highlight
...
+
+        trials['Total profit'] = trials['Total profit'].apply(
+            lambda x: '{:,.8f}'.format(x) if x != 0.0 else ""
+        )
+        trials['Profit'] = trials['Profit'].apply(
+            lambda x: '{:,.2f}'.format(x) if not isna(x) else ""
+        )
+        trials['Avg profit'] = trials['Avg profit'].apply(
+            lambda x: ('{:,.2f}%'.format(x)) if not isna(x) else ""
+        )
```
xmatthias: the 2 lambda functions seem to do the same thing - but parentheses are different ... ?

also, maybe using an f-string to be aligned to most of the other strings we have would make sense (?) but that's purely optional

PR #3025, opened 2020-03-05
minor create_trade() optimization

yazeed:
Just a minor optimization and ordering of code logic, I caught in create_trade() .

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -394,13 +394,10 @@ class FreqtradeBot:
             logger.info(f"Pair {pair} is currently locked.")
             return False
 
+        # get_free_open_trades is checked before create_trade is called
+        # but it is still used here to prevent opening too many trades within one iteration
         if not self.get_free_open_trades():
             logger.debug(f"Can't open a new trade for {pair}: max number of trades is reached.")
-            return False
-
-        stake_amount = self.get_trade_stake_amount(pair)
-        if not stake_amount:
-            logger.debug(f"Stake amount is 0, ignoring possible trade for {pair}.")
             return False
 
         # running get_signal on historical data fetched
@@ -409,6 +406,11 @@ class FreqtradeBot:
             self.dataprovider.ohlcv(pair, self.strategy.ticker_interval))
 
         if buy and not sell:
+            stake_amount = self.get_trade_stake_amount(pair)
+            if not stake_amount:
+                logger.debug(f"Stake amount is 0, ignoring possible trade for {pair}.")
+                return False
+
             logger.info(f"Buy signal found: about create a new trade with stake_amount: "
                         f"{stake_amount} ...")
 
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -246,7 +246,7 @@ class FreqtradeBot:
         if 'use_order_book' in bid_strategy and bid_strategy.get('use_order_book', False):
             logger.info(
                 f"Getting price from order book {bid_strategy['price_side'].capitalize()} side."
-                )
+            )
             order_book_top = bid_strategy.get('order_book_top', 1)
             order_book = self.exchange.get_order_book(pair, order_book_top)
             logger.debug('order_book %s', order_book)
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -246,7 +246,7 @@ def get_buy_rate(self, pair: str, refresh: bool) -> float:
         if 'use_order_book' in bid_strategy and bid_strategy.get('use_order_book', False):
             logger.info(
                 f"Getting price from order book {bid_strategy['price_side'].capitalize()} side."
-                )
+            )
```
hroff-1902: 👍

freqtrade/freqtradebot.py
```diff
@@ -394,21 +394,21 @@ def create_trade(self, pair: str) -> bool:
...
 
+        if not self.get_free_open_trades():
+            logger.debug(f"Can't open a new trade for {pair}: max number of trades is reached.")
+            return False
+
+        stake_amount = self.get_trade_stake_amount(pair)
+        if not stake_amount:
+            logger.debug(f"Stake amount is 0, ignoring possible trade for {pair}.")
+            return False
+
```
hroff-1902: Do we really need this optimization?
I understand, this optimizes calculations (`get_signal()` won't be called till free trade slot will become available), but do we really need this change? Imagine, custom `populate_indicators()` implements some custom actions (like fetching and aggregating data from an external source) that should be done periodically...

yazeed: @hroff-1902 Actually, that's kinda precisely how I detected it, I don't see why it's needed to be this way though, but feel free to give me an example. If we want to add any functionality to the strategy, we can always support it via the interface, but this here kinda felt random.

freqtrade/freqtradebot.py
```diff
@@ -394,21 +394,21 @@ def create_trade(self, pair: str) -> bool:
             logger.info(f"Pair {pair} is currently locked.")
             return False
 
+        if not self.get_free_open_trades():
+            logger.debug(f"Can't open a new trade for {pair}: max number of trades is reached.")
+            return False
```
xmatthias: I guess this code is ok here - no need to check for buy-signals when no trade-slot is open.
however, please add a clarifying comment here (i've already had a comment written that this segment is not needed and can never be hit) if you don't mind.

It's called in line #154, and if it returns false, `enter_positions()` and `create_trade()` are never called.

However, if the first pair in the whitelist opens a trade, then this is necessary for subsequent pairs - to avoid buying 5 trades in one iteration while max_open_trades is set to 3.

Now my wording is a bit ... verbose - so if you're able to find something concise it's even better (probably something along the lines of `# Prevent opening too many trades within one iteration` ...

freqtrade/freqtradebot.py
xmatthias: What exactly did you notice in `populate_*()` methods?

I struggle to see how checking this afterwards can lead to problems within `populate_*()` methods.

----

I agree with @hroff-1902 - i don't think we need this "optimization" (I wouldn't call it optimization though).
This call (`get_trade_stake_amount()`) is not a "cheap one" ... it calls `refresh_balance()` (which is a rather expensive call against the exchange when running in live mode). 

Let's do an example:
Let's assume you got 20 pairs in your whitelist.
With this change, the bot calls the excange 20 times to get the latest balance, even if no pair generates a buy signal.
without this change, the exchange is only "asked" for pairs with active buy signal. 
Now in theory, all 20 pairs could generate a buy-signal - but still the amount of calls to `fetch_balance()` will be limited to `max_open_trades()`.

So currently, the worst case `fetch_balance()` is called is `len(whitelist)` - but with this PR it's always called `len(whitelist)` (assuming no pair is generating buy signals).

This will lead to blocks on the exchange VERY soon ... and slow the overall operation / loop duration of the bot down.
This call becomes breaking for Kraken - which only provides a "half-hearted" `fetch_balance()` method and requires an additional, second call to `fetch_open_orders()` to get the correct balances (you can look at the implementation details in `kraken.py`.

Unfortunately this can only be tested in real mode (dry-run does `refresh_balance()` locally) - so the slowdown won't be noticed in that case.

hroff-1902: @yazeed I mean assume custom populate_* method(s) rely they are called for each candles (for instance, aggregate or collect some data, adding one candle, with initial 'warm-up' period)... In case of your optimized code, 'warm-up' can happen at the middle of the bot operation.

Plus, with the current behavior it's easy to explain in the docs and to the users asking when the populate_* methods are called.

xmatthias: Now don't ask me why my post shows up twice ... maybe answering to a previous review as part of a review has that effect ... makes it quite ... odd to see though.

yazeed:
I'll take a look at this again, this evening.

hroff-1902:
I'm proposing to leave the call to `self.get_trade_stake_amount()` in its current place and merge couple of other changes made here.

freqtrade/freqtradebot.py
yazeed: Gotcha, comment added, and adjustments made to this PR. Sorry, this took a while, haven't had much time recently.

yazeed: Gotcha, comment added, and adjustments made to this PR. Sorry, this took a while, haven't had much time recently.

PR #3032, opened 2020-03-08
Do not use "ticker" where it's not a ticker

hroff-1902:
* parse_ticker_dataframe --> ohlcv_to_dataframe in converter
* _get_ticker_list --> _get_ohlcv_as_lists in backtesting
* various changes in wording in docs and comments to clarify the kind of data used
* "preparations" for deprecation of ticker_interval (to be replaced with timeframe in the future)
* the name of the hyperopt pickling file changed (the data is not "tickerdata", but historical backtesting OHLCVs)

sorry for only 43 files changed, there was no good point in the process of making changes to split into several PRs or commits...


hroff-1902:
> * parse_ticker_dataframe --> ohlcv_to_dataframe in converter
> * _get_ticker_list --> _get_ohlcv_as_lists in backtesting
> * various changes in wording in docs and comments to clarify the kind of data used
> * "preparations" for deprecation of ticker_interval (to be replaced with timeframe in the future)
> * the name of the hyperopt pickling file changed (the data is not "tickerdata", but historical backtesting OHLCVs)
> 
> sorry for only 43 files changed, there was no good point in the process of making changes to split into several PRs or commits...

freqtrade/edge/edge_positioning.py
```diff
@@ -359,11 +359,11 @@ def _process_expectancy(self, results: DataFrame) -> Dict[str, Any]:
         # Returning a list of pairs in order of "expectancy"
         return final
 
-    def _find_trades_for_stoploss_range(self, ticker_data, pair, stoploss_range):
-        buy_column = ticker_data['buy'].values
-        sell_column = ticker_data['sell'].values
-        date_column = ticker_data['date'].values
-        ohlc_columns = ticker_data[['open', 'high', 'low', 'close']].values
+    def _find_trades_for_stoploss_range(self, dataframe, pair, stoploss_range):
```
xmatthias: i don't like the naming of `dataframe` ...
It's way too generic - i know the type - but the content could be "anything" ...

freqtrade/misc.py
```diff
@@ -81,13 +81,13 @@ def file_load_json(file):
         gzipfile = file
     # Try gzip file first, otherwise regular json file.
     if gzipfile.is_file():
-        logger.debug('Loading ticker data from file %s', gzipfile)
-        with gzip.open(gzipfile) as tickerdata:
-            pairdata = json_load(tickerdata)
+        logger.debug(f"Loading historical data from file {gzipfile}")
```
xmatthias:
```suggestion
        logger.debug(f"Loading data from file {gzipfile}")
```

I'd make this (and the below) more generic ... 
This is only used for trades atm, not for ohlcv data anymore ...

freqtrade/misc.py
```diff
@@ -81,13 +81,13 @@ def file_load_json(file):
...
+        with gzip.open(gzipfile) as datafile:
+            pairdata = json_load(datafile)
     elif file.is_file():
-        logger.debug('Loading ticker data from file %s', file)
-        with open(file) as tickerdata:
-            pairdata = json_load(tickerdata)
+        logger.debug(f"Loading historical data from file {file}")
```
xmatthias:
```suggestion
        logger.debug(f"Loading data from file {file}")
```

freqtrade/optimize/backtesting.py
```diff
@@ -151,32 +151,33 @@ def _store_backtest_result(self, recordfilename: Path, results: DataFrame,
...
-        ticker: Dict = {}
-        # Create ticker dict
+        data: Dict = {}
+        # Create dict with data
         for pair, pair_data in processed.items():
             pair_data.loc[:, 'buy'] = 0  # cleanup from previous run
             pair_data.loc[:, 'sell'] = 0  # cleanup from previous run
 
-            ticker_data = self.strategy.advise_sell(
+            dataframe = self.strategy.advise_sell(
```
xmatthias: yes, this is a dataframe ... using that as variable name has a certain smell to it for me - we also don't call the integer variable `INT` ... but instead what it does contain.

maybe `df_ohlcv`? or `df_analyzed` ...

freqtrade/optimize/hyperopt.py
```diff
@@ -75,8 +75,8 @@ def __init__(self, config: Dict[str, Any]) -> None:
 
         self.trials_file = (self.config['user_data_dir'] /
                             'hyperopt_results' / 'hyperopt_results.pickle')
-        self.tickerdata_pickle = (self.config['user_data_dir'] /
-                                  'hyperopt_results' / 'hyperopt_tickerdata.pkl')
+        self.data_pickle_file = (self.config['user_data_dir'] /
+                                 'hyperopt_results' / 'hyperopt_data.pkl')
```
xmatthias: i guess this renaming is ok as it's rebuilt each time.

but assuming a user already used hyperopt before - won't this leave an "orphaned" file with the old filename on the disk it's never going to cleanup again?

freqtrade/plot/plotting.py
```diff
@@ -451,13 +451,13 @@ def load_and_plot_trades(config: Dict[str, Any]):
...
+    for pair, data in plot_elements["ohlcv"].items():
         pair_counter += 1
         logger.info("analyse pair %s", pair)
-        tickers = {}
-        tickers[pair] = data
+        ohlcv = {}
+        ohlcv[pair] = data
 
-        dataframe = strategy.analyze_ticker(tickers[pair], {'pair': pair})
+        dataframe = strategy.analyze_ticker(ohlcv[pair], {'pair': pair})
```
xmatthias: While not directly related to these changes ... 
i think this can be simplified to `dataframe = strategy.analyze_ticker(data, {'pair': pair})`

the above dict (`ohlcv`) is not used again in this ... ?

tests/optimize/test_backtesting.py
```diff
@@ -360,7 +359,6 @@ def get_timerange(input1):
     mocker.patch('freqtrade.data.history.history_utils.load_pair_history',
                  MagicMock(return_value=pd.DataFrame()))
     mocker.patch('freqtrade.data.history.get_timerange', get_timerange)
-    mocker.patch('freqtrade.exchange.Exchange.refresh_latest_ohlcv', MagicMock())
```
xmatthias: mmmh, remains of the `--live` option ... :+1:

tests/optimize/test_hyperopt.py
```diff
@@ -1160,7 +1160,7 @@ def test_simplified_interface_sell(mocker, default_conf, caplog, capsys) -> None
     out, err = capsys.readouterr()
     assert 'Best result:\n\n*    1/1: foo result Objective: 1.00000\n' in out
     assert dumper.called
-    # Should be called twice, once for tickerdata, once to save evaluations
+    # Should be called twice, once for historical candle data, once to save evaluations
```
xmatthias: you must hate the term tickerdata very much to adjust all comments in tests!! :laughing:

freqtrade/data/converter.py
```diff
@@ -13,34 +13,31 @@
 logger = logging.getLogger(__name__)
 
 
-def parse_ticker_dataframe(ticker: list, timeframe: str, pair: str, *,
-                           fill_missing: bool = True,
-                           drop_incomplete: bool = True) -> DataFrame:
+def ohlcv_to_dataframe(ohlcv: list, timeframe: str, pair: str, *,
+                       fill_missing: bool = True, drop_incomplete: bool = True) -> DataFrame:
```
xmatthias: this namingchange gives me headaches.

on one hand, it shouldn't matter as it's a internal function ...
on the other hand it's useful to test something "real quick" ... and i suspect i'm not the only one importing and using this directly.

freqtrade/misc.py
hroff-1902: But trades (from gzipfile) are still "historical data"?

freqtrade/edge/edge_positioning.py
hroff-1902: it's same dataframe and the same parameter name for it, as used in the populate_* methods in the strategy -- so also called as "dataframe".

I would not rename it to something like "dataframe_with_signals", but rather the helpstring with the description of the parameter should later be added, along with type hints, which are also missing here. Not for this PR.

freqtrade/misc.py
hroff-1902: Would leave it as it is, trades loaded from disk are also historical data.

freqtrade/misc.py
hroff-1902: same

freqtrade/data/converter.py
hroff-1902: how do you use it? (in the interactive environment, I assume...)

tests/optimize/test_hyperopt.py
hroff-1902: No, the only way to see that the "ticker" term is only left in the places when it really corresponds to the ticker is to substitute it with proper terms in all wrong places, including names of the local vars and comments. Otherwise the grep results will non be clear. Including the tests. Plus, since this usage is confusing -- the tests should also be clean and understandable, with proper terms used, including comments... Especially for future devs.

freqtrade/optimize/hyperopt.py
hroff-1902: returned back, I have some other possible plans for it for the far perspective, when it will be more convenient to change the name...

freqtrade/data/converter.py
xmatthias: if it's interactive or a snippet of code i copy/paste into a REPL is not really relevant, is it?

hroff-1902: I mean, I do not use it along, you use it with some functions before and some functions after.
So the call to this function can be added into more convenient ones, to be called interactively or from a snippet, leaving this one internal.

> this namingchange gives me headaches.

But why? What's wrong with the new name?

xmatthias: it's the change itself - not the old or new name ... but as you'll have seen i did approve so it's good to go (however i don't have time today to fix the merge conflict ...).

PR #3033, opened 2020-03-09
trim dataframe by timerange if timerange is given

orehunt:
## Summary
IDataHandler `_ohlcv_load` accepts `timerange` but `JsonDataHandler` doesn't use it...but the public `ohlcv_load` method does an `.empty` check. If you pass a timerange for a pair that does have data, just not for that timerange, that check is voided and you get an index exception..
some tests seem to give the timerange but expect a full df anyway so they break

### DIFF
--- a/freqtrade/data/history/jsondatahandler.py
+++ b/freqtrade/data/history/jsondatahandler.py
@@ -77,6 +77,11 @@ class JsonDataHandler(IDataHandler):
                                        unit='ms',
                                        utc=True,
                                        infer_datetime_format=True)
+        if timerange:
+            start = to_datetime(timerange.startts, utc=True, unit='s')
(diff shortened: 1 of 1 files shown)


orehunt:
so the trimming happens in the public method...so I guess that arg was left for backward compatibility..

xmatthias:
actually for forward compatibility.

Assuming a future data-handler which supports range-queries (database / influxdb) - you'll need the timerange to filter "at the source" and avoid loading the whole dataset into memory.

Now more work will probably need to be done - but that's the reason to have kept this in the private method too.

orehunt:
a double check still has to be done though
https://github.com/orehunt/freqtrade/commit/dacddcc514c444c5544554240caf06509c58430e#

PR #3040, opened 2020-03-09
reduce Pairlist message warning level

### LINKED ISSUE #2992: Warning log level when VolumePairList and a blacklist are associated
Hi,

Given i have set a VolumePairList and a blacklist containing ` [BNB/USDT]`
When the bot refresh the markets
Then I have a warning message
`2020-02-28 16:01:47,568 - freqtrade.pairlist.IPairList - WARNING - Pair BNB/USDT in your blacklist. Removing it from whitelist...`

I understand a warning in the case i have put in your whitelist and blacklist the same market (has it can be human mistake), but in this case i don't understand why.

Maybe a make an error in my config file ?

Many tks.

xmatthias:
## Summary
Blacklisted pair mesages should not appear as Warning when using volumepairlist

Now this intentionally leaves the repeated logging in case of staticpairlist, which collides with blacklisted pairs - as in that case, the pairlist should be fixed instead.

Closes #2992

## Quick changelog

- conditionally log blacklisted message
- Don't check pair blacklist 3 times in a row ...

### DIFF
--- a/freqtrade/pairlist/IPairList.py
+++ b/freqtrade/pairlist/IPairList.py
@@ -67,21 +67,37 @@ class IPairList(ABC):
         """
 
     @staticmethod
-    def verify_blacklist(pairlist: List[str], blacklist: List[str]) -> List[str]:
+    def verify_blacklist(pairlist: List[str], blacklist: List[str],
+                         aswarning: bool) -> List[str]:
         """
         Verify and remove items from pairlist - returning a filtered pairlist.
+        Logs a warning or info depending on `aswarning`.
+        Pairlists explicitly using this method shall use `aswarning=False`!
+        :param pairlist: Pairlist to validate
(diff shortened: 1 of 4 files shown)


hroff-1902:
Honestly, did not get why it's still called twice, though not in this PR anyway.

LGTM 👍

xmatthias:
hah - it needs to.

volumepairlist must remove the pairs itself so the endresult does contain the 20 pairs i selected (otherwise we may end up with only 19 or 18 - if blacklist removes one after volumepairlist).

The pairlist must however also be checked after all pairlistproviders have run - to ensure that staticpairlist (or any other pairlist) follow the blacklisted rules - without explicitly implementing it in each provider ... as that would be error-prone (and would lead to it being called 4 times, not just twice (which is also only the case with volumepairlist)).

PR #3045, opened 2020-03-09
check again for emptiness after trimming dataframe

orehunt:
## Summary
Trimming the dataframe can still return an empty dataframe, so check again for emptyiness

### DIFF
--- a/freqtrade/data/history/idatahandler.py
+++ b/freqtrade/data/history/idatahandler.py
@@ -160,6 +160,13 @@ class IDataHandler(ABC):
             if timerange_startup:
                 self._validate_pairdata(pair, pairdf, timerange_startup)
                 pairdf = trim_dataframe(pairdf, timerange_startup)
+                if pairdf.empty:
+                    if warn_no_data:
+                        logger.warning(
+                            f'No history data for pair: "{pair}", timeframe: {timeframe}. '
+                            'Use `freqtrade download-data` to download the data'
+                        )
+                    return pairdf
 
             # incomplete candles should only be dropped if we didn't trim the end beforehand.
             return clean_ohlcv_dataframe(pairdf, timeframe,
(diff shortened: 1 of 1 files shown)


hroff-1902:
This should probably be refactored further: the empty dataframe can also be returned by clean_ohlcv_dataframe, when the single candle in the dataframe is incomplete one and dropped...

Maybe not in this PR.

I do not like here repeating of the same message in 2 places...

xmatthias:
I think you're right @hroff-1902, i took the liberty to extract this into a method which is called from both locations.

@orehunt Hope you don't mind...

freqtrade/data/history/idatahandler.py
```diff
@@ -147,19 +147,16 @@ def ohlcv_load(self, pair, timeframe: str,
...
             enddate = pairdf.iloc[-1]['date']
 
             if timerange_startup:
                 self._validate_pairdata(pair, pairdf, timerange_startup)
                 pairdf = trim_dataframe(pairdf, timerange_startup)
+                if self._check_empty_df(pairdf, pair, timeframe, warn_no_data):
+                    return pairdf
 
             # incomplete candles should only be dropped if we didn't trim the end beforehand.
             return clean_ohlcv_dataframe(pairdf, timeframe,
```
hroff-1902: Should also
```
                if self._check_empty_df(pairdf, pair, timeframe, warn_no_data):
                    return pairdf
```
be added here after the call to clean_ohlcv_dataframe? (as I mentioned above, this is simple and fits well this PR)

Actually (and this is not for this PR), we can later extend `_check_empty_df` to log a warning if the length of the df is less than the strategy startup period, warning the user about possible wrong signals in this case.

xmatthias: i'm not sure - under what (realistic) circumstance would a non-empty DataFrame come out empty from clean_ohlcv?

hroff-1902: when it drops single candle

xmatthias: that's not very realistic ... you'd have to set the timerange exactly to one candle - which will then remove that ...

hroff-1902: but why should not we print this message if we print it in other cases when it gets empty?

hroff-1902:
@orehunt thanks!

PR #3054, opened 2020-03-10
Hyperopt: Progressbar during hyperopt

Fredrik81:
## Summary
Adding progressbar to hyperopt output printer with interval between table prints

## Quick changelog

- Adding module tqdm
- output of progressbar during hyperopt prints and ETA estimation

## What's new?
```
2020-03-12 02:05:13,330 - freqtrade.optimize.hyperopt - INFO - Number of parallel jobs set as: -1
2020-03-12 02:05:13,344 - freqtrade.optimize.hyperopt - INFO - Effective number of parallel workers used: 8
+--------+---------+----------+--------------+------------------------------+----------------+-------------+
|   Best |   Epoch |   Trades |   Avg profit |                       Profit |   Avg duration |   Objective |
|--------+---------+----------+--------------+------------------------------+----------------+-------------|
|      * |    1/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |    2/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |    3/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |    4/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |    5/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |    6/40 |        0 |           -- |                           -- |             -- |         N/A |
|   Best |    7/40 |        3 |        1.10% |    0.00019887 BTC    (3.31%) |          0.0 m |     1.84685 |
|      * |    8/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |    9/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   10/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   11/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   12/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   13/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   14/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   15/40 |        1 |        0.42% |    0.00002521 BTC    (0.42%) |          0.0 m |     1.85703 |
|      * |   16/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   17/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   18/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   19/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   20/40 |       19 |       -0.02% |   -0.00002106 BTC   (-0.35%) |         64.7 m |     1.94107 |
|      * |   21/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   22/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   23/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   24/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   25/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   26/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   27/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   28/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   29/40 |        0 |           -- |                           -- |             -- |         N/A |
|      * |   30/40 |        1 |        0.62% |    0.00003737 BTC    (0.62%) |          0.0 m |     1.85636 |
|        |   31/40 |        0 |           -- |                           -- |             -- |         N/A |
|        |   32/40 |        0 |           -- |                           -- |             -- |         N/A |
|        |   33/40 |        0 |           -- |                           -- |             -- |         N/A |
|        |   34/40 |        6 |       -0.16% |   -0.00005707 BTC   (-0.95%) |        202.5 m |     2.13025 |
|        |   35/40 |       26 |        0.15% |    0.00023142 BTC    (3.85%) |         88.3 m |     1.95654 |
|        |   36/40 |        0 |           -- |                           -- |             -- |         N/A |
|        |   37/40 |        3 |       -1.49% |   -0.00026788 BTC   (-4.46%) |          0.0 m |     1.87276 |
|        |   38/40 |        0 |           -- |                           -- |             -- |         N/A |
|        |   39/40 |       36 |        0.31% |    0.00066657 BTC   (11.10%) |        112.9 m |     1.96256 |
|        |   40/40 |        0 |           -- |                           -- |             -- |         N/A |
Epoch 40/40 (100%)|████████████████████████████████████████████████████████████| [00:11<00:00  3.55 Epoch/s]
2020-03-12 02:05:24,638 - freqtrade.optimize.hyperopt - INFO - 40 epochs saved to '.../user_data/hyperopt_results/hyperopt_results.pickle'.
```


hroff-1902:
It does not repaint yet well with logging printed (add `-v` option)

hroff-1902:
I merged the latest changes made to develop...

freqtrade/optimize/hyperopt.py
```diff
@@ -661,6 +681,14 @@ def start(self) -> None:
             with Parallel(n_jobs=config_jobs) as parallel:
                 jobs = parallel._effective_n_jobs()
                 logger.info(f'Effective number of parallel workers used: {jobs}')
+
+                # Define progressbar
+                self.progress_bar = tqdm(
+                    total=self.total_epochs, ncols=108, unit=' Epoch',
+                    bar_format='Epoch {n_fmt}/{total_fmt} ({percentage:3.0f}%)|{bar}|'
+                    ' [{elapsed}<{remaining} {rate_fmt}{postfix}]'
+                )
```
hroff-1902:
```suggestion
                self.progress_bar = tqdm(
                    total=self.total_epochs, unit='Epoch',
                    bar_format='Epoch {n_fmt}/{total_fmt} ({percentage:3.0f}%) |{bar}|'
                               ' [{elapsed}<{remaining} {rate_fmt}{postfix}]',
                    ncols=108,
                    position=1,
                    leave=False,
                )
```

freqtrade/optimize/hyperopt.py
```diff
@@ -686,16 +715,23 @@ def start(self) -> None:
                         # evaluations can take different time. Here they are aligned in the
                         # order they will be shown to the user.
                         val['is_best'] = is_best
-
-                        self.print_results(val)
+                        output = self.get_results(val)
+                        if output:
+                            self.progress_bar.write(output)
+                        self.progress_bar.ncols = 108
```
hroff-1902:
```suggestion
```
-- not needed

freqtrade/optimize/hyperopt.py
```diff
@@ -686,16 +715,23 @@ def start(self) -> None:
...
+                        self.progress_bar.update(1)
 
                         if is_best:
                             self.current_best_loss = val['loss']
                         self.trials.append(val)
                         # Save results after each best epoch and every 100 epochs
                         if is_best or current % 100 == 0:
                             self.save_trials()
+                self.progress_bar.ncols = 108
```
hroff-1902:
```suggestion
```
-- not needed

hroff-1902:
I made some explorations,
first, let's agree on some cosmetics changes, to have same codebase for further changes.

Pls review the suggestions above and commit them if you are agree.

freqtrade/optimize/hyperopt.py
Fredrik81: In my testing it did not keep the ncols during the threaded part. 
But will test again

hroff-1902: printing is done in one main process, not in the workers

freqtrade/optimize/hyperopt.py
Fredrik81: I'm a bit sceptical about the leave=False
1. We can't do a pytest to see that it prints
2. If we remove the {current}/{total epoch} we do not have any indicator at the end showing how many it was suppose to perform?

freqtrade/optimize/hyperopt.py
Fredrik81: Hmm it didn't work in my initial tests but i guess with all the changes i made over time it's working now.. just tested it

freqtrade/optimize/hyperopt.py
Fredrik81: same

freqtrade/optimize/hyperopt.py
hroff-1902: maybe, need to think of the resulting behavior...

you can remove `leave=False` for a while, I'll commit the further proposals for review...

Fredrik81: For me it prints correct now with the -v option but I also have some extra code. 
Need to check with this new commit

hroff-1902:
committed the suggestion (after your "👍" mark) and removed the `leave=False` parameter as discussed above -- in order to have the codebase that could be fetched/pulled for reviewing the behavior without manual copy-paste of the proposed changes...

freqtrade/optimize/hyperopt.py
hroff-1902: not quite:
run it with `-j 1 -v` (to see a shorter logging message first at each epoch) -- you'll see that the progressbar is not cleared. Same if you add printing some short value in the populate_buy_trend in the hyperopt class used.

requirements-hyperopt.txt
```diff
@@ -7,3 +7,4 @@ scikit-learn==0.22.2.post1
 scikit-optimize==0.7.4
 filelock==3.0.12
 joblib==0.14.1
+tqdm==4.43.0
```
xmatthias: Please make sure to add this to setup.py as well ... 

installation from pypi (`pip install freqtrade[all]` doesn't use the requirements.txt files, but relies on what's specified in setup.py).

freqtrade/optimize/hyperopt.py
```diff
@@ -686,16 +717,21 @@ def start(self) -> None:
...
                         if is_best:
                             self.current_best_loss = val['loss']
                         self.trials.append(val)
                         # Save results after each best epoch and every 100 epochs
                         if is_best or current % 100 == 0:
                             self.save_trials()
+                self.progress_bar.close()
+
         except KeyboardInterrupt:
+            self.progress_bar.close()
```
xmatthias: This looks to be a case for `finally:` ...
note that you should check existance of the progressbar (`if self.progress_bar`) - assuming i am able to stop it before it's initialized ...

xmatthias:
A general note - this doesn't seem to play well with debug output ... 
the progressbar was "scrolled up" ... maybe we should disable progressbar in debug mode and use a simple "print" instead ...?

freqtrade/optimize/hyperopt.py
hroff-1902: It's better to instantiate tqdm in a `with` statement here -- it will close it automatically when it's needed

hroff-1902:
> A general note - this doesn't seem to play well with debug output ...
> the progressbar was "scrolled up" ... maybe we should disable progressbar in debug mode and use a simple "print" instead ...?

it's not only about debug log (though it's a good approach, to disable progressbar in debug mode), it's also about printing something from the hyperopt methods (with custom `print` statements, printing something)

requirements-hyperopt.txt
hroff-1902: I always forgetting this... 👍

Fredrik81: Good point will do with next commit

Fredrik81:
Debug print look ok'ish' on my end but have extra code. Will need to test again but will work on it later this week. 
Would be good to find a solution for logging in case we find other uses for progress bars

xmatthias:
> Would be good to find a solution for logging in case we find other uses for progress bars

well we're for sure not the first ones looking for a solution - and i suspect we won't be the last either. example: https://github.com/tqdm/tqdm/issues/313

Now i'm not sure how well / easy this can be tied into freqtrade - and i think we'll also need to be careful if this will then log to stdout or stderr (which does make a difference ...) - but in general it seems doable in a "relatively simple" way.

Fredrik81:
> > Would be good to find a solution for logging in case we find other uses for progress bars
> 
> well we're for sure not the first ones looking for a solution - and i suspect we won't be the last either. example: [tqdm/tqdm#313](https://github.com/tqdm/tqdm/issues/313)
> 
> Now i'm not sure how well / easy this can be tied into freqtrade - and i think we'll also need to be careful if this will then log to stdout or stderr (which does make a difference ...) - but in general it seems doable in a "relatively simple" way.

Let me get a bit more time with this.. have made progress without messing with the logger.

freqtrade/optimize/hyperopt.py
Fredrik81: changing back to progressbar2

requirements-hyperopt.txt
Fredrik81: i see a potential problem with the implementation in setup.py.. doesn't seem to have version control so it will install newer compared to requirements files if it exists = not tested.

freqtrade/optimize/hyperopt.py
Fredrik81: this is solved in the latest pull using progressbar2

requirements-hyperopt.txt
Fredrik81: can't be "fixed"

Fredrik81:
At this point the progessbar works fine except for when coloring is used. This is being worked on also getting the planning to use same colors as in the hyperopt table output.

xmatthias:
> At this point the progessbar works fine except for when coloring is used. This is being worked on also getting the planning to use same colors as in the hyperopt table output.

Excuse my ignorance - but what's the problem with colored output ? 

it does seem ok to me also with colors (unless a stupid user like myself presses enter a few times ... but that's not related to colors ... :laughing: )

xmatthias:
somehow, this seems to break the linux "pipe" functionality.

running with `freqtrade hyperopt <...> --print-all --no-color >> output.txt` does not write anything into output.txt ... (but it works with the same command in develop).

Now to be clear, i'm not certain that's a problem - i'm simply recording it here that i discovered this.
i've personally never used this functionality for anything other than testing it works ... but i guess it could become problematic in certain instances - although i currently struggle to figure out why it would be problematic - hyperopt output is not THAT long usually, so copy/pasting the output into a file is always possible, too.

Fredrik81:
> somehow, this seems to break the linux "pipe" functionality.
> 
> running with `freqtrade hyperopt <...> --print-all --no-color >> output.txt` does not write anything into output.txt ... (but it works with the same command in develop).
> 
> Now to be clear, i'm not certain that's a problem - i'm simply recording it here that i discovered this.
> i've personally never used this functionality for anything other than testing it works ... but i guess it could become problematic in certain instances - although i currently struggle to figure out why it would be problematic - hyperopt output is not THAT long usually, so copy/pasting the output into a file is always possible, too.

Yeah i can see that happening now that the progressbar is capturing the output and re-writing it

freqtrade/optimize/hyperopt.py
```diff
@@ -654,47 +666,75 @@ def start(self) -> None:
...
-                        if is_best or current % 100 == 0:
-                            self.save_trials()
+
+                # Define progressbar
+                if self.print_colorized:
+                    widgets = [
+                        ' [Epoch ', progressbar.Counter(), ' of ', str(self.total_epochs),
+                        ' (', progressbar.Percentage(), ')] ',
+                        progressbar.Bar(marker=progressbar.AnimatedMarker(
+                            fill='█',
```
xmatthias, line 680: I would replace this with the following:

```suggestion
                            fill='\N{FULL BLOCK}',
```

Otherwise it could be problematic for windows  ... and is dependent on the correct character-settings to be installed.

freqtrade/optimize/hyperopt.py
```diff
@@ -654,47 +666,75 @@ def start(self) -> None:
...
+                            marker_wrap=Style.BRIGHT + '{}' + Style.RESET_ALL,
+                        )),
+                        ' [', progressbar.ETA(), ', ', progressbar.Timer(), ']',
+                    ]
+                else:
+                    widgets = [
+                        ' [Epoch ', progressbar.Counter(), ' of ', str(self.total_epochs),
+                        ' (', progressbar.Percentage(), ')] ',
+                        progressbar.Bar(marker=progressbar.AnimatedMarker(
+                            fill='█',
```
xmatthias, line 691:
```suggestion
                            fill='\N{FULL BLOCK}',
```

PR #3055, opened 2020-03-11
Verify date on last candle before producing signal

yazeed:
This PR ensures validation of the date on the last candle to make sure it's the latest candle before producing a signal, as I've witnessed on several occasions, Binance taking more time before sending in the latest candle.


yazeed:
Sorry about all the repetitive commits, just figured out how to use commit --amend .

yazeed:
Not sure amend has worked properly for me, there. I'll figure it out.

yazeed:
I'll now work on adding tests for this scenario.

yazeed:
Tests added, should coverage update itself?

yazeed:
Tests added. Coverage updated.

freqtrade/strategy/interface.py
```diff
@@ -276,19 +276,25 @@ def get_signal(self, pair: str, interval: str,
...
-                (arrow.utcnow() - signal_date).seconds // 60
+                int((arrow.utcnow() - signal_date).total_seconds() // 60)
             )
             return False, False
 
+        # Check if dataframe has new candle
+        if (arrow.utcnow() - signal_date).total_seconds() // 60 >= interval_minutes:
+            logger.warning('Old candle for pair %s. Last candle is %s minutes old',
+                           pair, int((arrow.utcnow() - signal_date).total_seconds() // 60))
+            return False, False
```
xmatthias: this will produce very many logs - especially on lower timeframes (1m / 5m on non-highvolume pairs)  as this will happen at almost every candle (switch).

the endresult of this is basically the same as the comparison above for outdated history ... but without being configurable, and without the 2x offset - and with the sides inversed.

and it's really not fixing the problem we discussed yesterday anyway - as that was to do with locked pairs beeing unlocked too early.

This behaviour can be mimict by setting  `exchange.outdated_offset` to `-interval_minutes` (depends on your timeframe, obviously) - and will have the same exact effect - just with a slightly different error message.

yazeed: The warning message usually occurs only for pairs that still hasn't produced a new candle (as discussed yesterday), but perhaps we should rephrase it to make users panic less about it? Or switch it to a debug all-together.

It does fix the problem from a different angle, because I figured after our discussion yesterday that pairs getting unlocked only prevents duplicate buys on the same candle, but it still doesn't fix it for candles that were not bought/locked before.

The whole purpose of this, is to fix signals being produced out of older candles, and yes, I noticed using outdated_offset would fix it, but I noticed it's not exactly the documented purpose for that feature, this one is just to confirm receiving new candles before producing a signal.

So what do you suggest?

xmatthias: nothing at the moment ... i'll need to think about this and time to play with it to see the effects it has for different exchanges, different pairs and different timeframes.

I'll try to get to it soon - but it'll probably be weekend ...

hroff-1902:
just a note:
commits like "meh", "final?" should be squashed at merge.

Someone still reads the changelog...

freqtrade/freqtradebot.py
```diff
@@ -390,14 +390,14 @@ def create_trade(self, pair: str) -> bool:
         """
         logger.debug(f"create_trade for pair {pair}")
 
+        # Check if dataframe is out of date
```
hroff-1902: should this comment be here?
(you moved the same comment in freqtrade/strategy/interface.py in its proper place and the latter is okay, but should it also be placed here?)

hroff-1902:
In my opinion, that's completely wrong.

Assume we use 5m timeframe and now it's 11:24:57, throttling interval is default 5 s. The next throttling time is 11:25:02 and that's the beginning of new candle.

Exchanges need some time to aggregate new candles, they do not provide new candle data requested at 11:25:00.1, and even at 11:25:01. Suppose it takes 3 s and the new candle data will only be fetched after 11:25:03.

So at this next throttling iteration, at 11:25:02, the bot will still fetch old candle data and your condition `arrow.utcnow() - signal_date).total_seconds() // 60 >= interval_minutes` will print the warning, while the operation of the bot is completely okay.

Moreover (and you can see it in the debug log), the exchanges do not issue new candles for different pairs in the same time. For example, if we have 20 pairs and our next throttling iteration fetches data at 11:25:02, the bot *may* fetch new candle for, say, 5 pairs (if the exchange is quick enough and is able to aggregate the candles in 2 seconds and provide them in the REST requests) and old candles for the rest of 15 pairs, then at 11:25:07 it will fetch new candles for the rest of 15 pairs. Or, if the exchange is slow enough, the bot will also fetch at 11:25:07 new candles for only, say,  12 pairs and the new candles for the rest of 3 pairs at just 11:25:12 (if the data are ready at exchange after 11:25:07).

Your code will print warnings in all of those cases, while the operation of the bot is completely okay.

Current check with `(interval_minutes * 2 + offset)` offset efficiently checks only large differences in the candle times.

I agree with @xmatthias , this will only print lot's of warnings in the situation when anything is okay.

hroff-1902:
Actually, I did not get the problem why some additional logging should be added if the time difference is less than that calculated with `(interval_minutes * 2 + offset)` offset.

hroff-1902:
I think I get the case: assume you have the bought some time ago, current time is 11:24:57, the pair was sold during this candle (i.e. after 11:20), so the pair is locked till 11:25:00, but it can be bought at 11:25:02 if no new data arrive at this time and you have the buy signal for this current locked pair, right?

This should be fixed other way, imho -- the `is_pair_locked()` should compare the data of the current last candle instead of `now()` in order to get, if the pair is still locked. Thus, if the pair is locked for the current candle, it will be locked till the new candle data arrive at the next throttling iterations...

xmatthias:
> All they need some time to aggregate the candle and propagate the data from the order-matching backend servers to frontend answering REST requests.

It's true that exchanges don't give you candles right away once the hour passes - but the reason is not time they need in the backend.
It's pretty well observable when using websockets (or also by looking at the UI of exchanges, which usually also use websockets).
The new candle will be generated when the first trade happens in the new candle - now this may be at 11:25:00.00001 precisely - or at 11:25:50 (on a low volume pair) - in which case a new candle would not be sent until 11:25:50 (at least this is how i observed it about a year ago playing with the WS fork of ccxt).

freqtrade/freqtradebot.py
yazeed: Nope, not sure why it's also here. :)

yazeed: > I think I get the case: assume you have a pair bought some time ago, current time is 11:24:57, the pair was sold during this candle (i.e. after 11:20:00), so the pair is locked till 11:25:00, but it can be bought at 11:25:02 if no new data arrive at this time and you have the buy signal for this current locked pair, right?
> 
> This should be fixed other way, imho -- the `is_pair_locked()` should compare the data of the current last candle instead of `now()` in order to get, if the pair is still locked. Thus, if the pair is locked for the current candle, it will be locked till the new candle data arrive at the next throttling iterations...

The reason I didn't place the code logi in is_pair_locked() is this line would need to be changed/removed:
`if pair not in self._pair_locked_until:
            return False`
As this skips any unlocked pair, and we want the date validation to occur on all pairs, not just locked ones.

yazeed:
> I think I get the case: assume you have a pair bought some time ago, current time is 11:24:57, the pair was sold during this candle (i.e. after 11:20:00), so the pair is locked till 11:25:00, but it can be bought at 11:25:02 if no new data arrive at this time and you have the buy signal for this current locked pair, right?
> 
> This should be fixed other way, imho -- the `is_pair_locked()` should compare the data of the current last candle instead of `now()` in order to get, if the pair is still locked. Thus, if the pair is locked for the current candle, it will be locked till the new candle data arrive at the next throttling iterations...

The reason I didn't place the code logic in is_pair_locked() is this line would need to be changed/removed:
`if pair not in self._pair_locked_until:
            return False`
As this skips any unlocked pair, and we want the date validation to occur on all pairs, not just locked ones.

xmatthias:
> we want the date validation to occur on all pairs, not just locked ones.

And this is where we disagree ... The date validation IS taking place - but not with "current candle", but with 2 candles+ offset.

As said, this change will be very verbose, and will log this message at the start of almost every candle, for almost every (low-volume) pair.

This is the exact same as setting `exchange.outdated_offset` to `-timeframe_in_minutes` - by hardcoding this the way you did the `outdated_offset` setting will become ineffective, as no matter what you set as `outdated_offset`, your new part will take over and return `False, False` anyway.

----

To me, the bug you described (pair unlocking too early) can be fixed in 3 ways - either in `is_pair_locked()` - or by setting `outdated_offset` to `-interval_minutes`, or by setting `process_only_new_candles=True`.

* The first method will fix the bug you described explicitly (and only that) - irrelevant of the other settings.
* the 2nd is a configuration which will do the same, + prevent a LOT more
* the 3rd method will make sure that the populate_* methods are only ran when a new candle arrives.

Your approach is (kinda) mimicking the 3rd method - but in a odd way:

Assume no pair is locked, and we're using 1h candles (for simplicity) (and `process_only_new_candles=False`).

your assumption is that you'll not want to buy at 11:00:01 if the new candle has not been published yet, as that would be based on an old signal.
However, you're perfectly fine buy at 10:59:59 - as that is still within the old candle's time.

Both buy-decisions would be based on a signal that's roughly 1h old (+-2s).
What's wrong with these 2 seconds that you're trying to prevent one case, but not the other?

yazeed:
> > we want the date validation to occur on all pairs, not just locked ones.
> 
> And this is where we disagree ... The date validation IS taking place - but not with "current candle", but with 2 candles+ offset.
> 
> As said, this change will be very verbose, and will log this message at the start of almost every candle, for almost every (low-volume) pair.
> 
> This is the exact same as setting `exchange.outdated_offset` to `-timeframe_in_minutes` - by hardcoding this the way you did the `outdated_offset` setting will become ineffective, as no matter what you set as `outdated_offset`, your new part will take over and return `False, False` anyway.
> 
> To me, the bug you described (pair unlocking too early) can be fixed in 3 ways - either in `is_pair_locked()` - or by setting `outdated_offset` to `-interval_minutes`, or by setting `process_only_new_candles=True`.
> 
>     * The first method will fix the bug you described explicitly (and only that) - irrelevant of the other settings.
> 
>     * the 2nd is a configuration which will do the same, + prevent a LOT more

I can still take the logic of this PR to is_pair_locked as we've discussed, I actually implemented that first, but I'd have to do something about this code block which currently returns False, signalling the pair is OK to work with (not what we want, we want freqtrade to work with pairs that have an actually validated last candle) :

```
if pair not in self._pair_locked_until:
            return False
```

> 
>     * the 3rd method will make sure that the populate_* methods are only ran when a new candle arrives.
> 
> 
> Your approach is (kinda) mimicking the 3rd method - but in a odd way:
> 
> Assume no pair is locked, and we're using 1h candles (for simplicity) (and `process_only_new_candles=False`).
> 
> your assumption is that you'll not want to buy at 11:00:01 if the new candle has not been published yet, as that would be based on an old signal.
> However, you're perfectly fine buy at 10:59:59 - as that is still within the old candle's time.
> 
> Both buy-decisions would be based on a signal that's roughly 1h old (+-2s).
> What's wrong with these 2 seconds that you're trying to prevent one case, but not the other?

With process_only_new_candles (3rd method) you mentioned, it does not actually verify date on the last candle, and can still work with outdated candles.

> And this is where we disagree ... The date validation IS taking place - but not with "current candle", but with 2 candles+ offset.

The one with 2 candles+ offset doesn't do the same thing, even though I understand that I can actually get it to do the same thing, given I give it the right value, but its purpose does not validate the date on the last candle, it deals with a different issue (out of sync).

xmatthias:
> With process_only_new_candles (3rd method) you mentioned, it does not actually verify date on the last candle, and can still work with outdated candles.

That's not true. 
It verifies the Date of the candle, and makes sure that each candle (identified via date) is only ran once - so in the above example, the new candle at 11:00:05 would be analyzed exactly at that time - not at 11:00:00, and not at 11:01:00.


https://github.com/freqtrade/freqtrade/blob/f320c0a410e5832d4a3fc35647a5aa6fea6a958b/freqtrade/strategy/interface.py#L226-L229

freqtrade/strategy/interface.py
```diff
@@ -328,17 +328,24 @@ def get_signal(self, pair: str, interval: str, dataframe: DataFrame) -> Tuple[bo
...
+                int((arrow.utcnow() - latest_date).total_seconds() // 60)
             )
             return False, False
 
+        # Check if dataframe has new candle
+        if (arrow.utcnow() - latest_date).total_seconds() // 60 >= interval_minutes:
+            logger.warning('Old candle for pair %s. Last candle is %s minutes old',
+                           pair, int((arrow.utcnow() - latest_date).total_seconds() // 60))
+            return False, False
+
```
hroff-1902, line 348: In principle, this PR adds this new check only.

I did not get yet, what situation (which can happen in the user custom code) should it protect from...

freqtrade/freqtradebot.py
hroff-1902: Made it by myself, to avoid looking at it everytime during review... 😄

freqtrade/strategy/interface.py
xmatthias: well and this check is not really new - but the same (but with twice the candle lenght) exists a little above or below this check.

The initial problem was that lock_pair locks until the candle is finished (so until 10:00 - but a new candle may only arrive at 10:00:05 - so we got 5s where freqtrade would buy into the "old" signal - even though no new candle arrived.

As explained above - to me this is the wrong approach to fixing the problem it tries to address, but instead the problem should be addressed within `is_pair_locked` (but all of this is above - although it can be a long read ...).

xmatthias (approved):
I've adjusted the code now to actually fix the bug you tried to fix - have pair-locking correctly lock until the next candle arrives, even if that candle does not arrive "100% on time".

Issue #3020, opened 2020-03-05
a column for outstanding balance


ghost:
Having a column that shows the outstanding balance on each candle can be useful. What is the outstanding balance? The sum of the virtual profits of all open trades on each candle, this is a very slow version:
```python
from pandas import Series, Timedelta, date_range, DatetimeIndex

hloc = kwargs["processed"]
config = kwargs["config"]
timeframe = config["timeframe"]
timedelta = Timedelta(timeframe)

date_index: DatetimeIndex
date_index = date_range(start=min_date, end=max_date, freq=timeframe, normalize=True)
balance = {pair: Series(0.0, index=date_index) for pair in hloc}
balance_total = Series(0.0, index=date_index)
close_time = results["close_time"].values
open_time = results["open_time"].values
for pair in hloc:
    hloc[pair].set_index("date", inplace=True)
for date in date_index.values:
    # set the balance for all the opens that are not instantly closed
    where = (open_time == date) & (close_time != date)
    for _, tr in results.loc[where].iterrows():
        pair = tr["pair"]
        close = hloc[pair].at[date, "close"]
        # balance at date
        bal_at_date = 1 - tr["open_rate"] / close - slippage
        balance[pair][date] += bal_at_date
    # set the balance for all the closes
    for _, tr in results.loc[close_time == date].iterrows():
        pair = tr["pair"]
        # balance at date, since it's the close we take the profits minus slippage
        bal_at_date = tr["profit_percent"] - slippage
        balance[tr["pair"]][date] += bal_at_date
        # since we are closing, update all the balances since open time
        # ranges are inclusive:[):exclusive
        for open_date in (
            balance[pair]
            .loc[tr["open_time"] + timedelta : tr["close_time"]]
            .index
        ).values:
            close = hloc[pair].at[open_date, "close"]
            balance[pair][open_date] += 1 - tr["open_rate"] / close - slippage
    # settle the balance for all the pairs at current date
    for pair in balance:
        balance_total[date] += balance[pair][date]
```
I am trying to come up with a vectorized one, also not sure how much would this calculation impact performance if it was applied on the main backtest loop

xmatthias:
You should look at `btanalysis.py` - this should be very similar to the function `analyze_trade_parallelism()` we have to test max_open_trades is respected - but using `sum()` instead of `count()`.

But let me ask- what for / why would you need this?
I'm not certain in which usecase this should be used ... - knowing how many trades are open at any given time is usually enough ... ?

[...]

PR #3059, opened 2020-03-12
outstanding balance function

orehunt:
## Summary 
This adds a function that _should_ be able to calculate the balance of all open trades on each candle, using close prices. 
However results don't match with the function in the issue :)
Solve the issue: #3020 

## Quick changelog

- added outstanding balance function


xmatthias:
i still struggle to see why we need this - and what exactly this delivers.

I think this is a rather odd approach to calculate this - and being honest, i think the formula used is wrong as in theory, all you should need to do is:

* create a timeseries dataframe (per pair) with the total length of the timerange - exploding every trade into multiple lines (one line per candle) - code for this is available from `analyze_trade_parallelism()` and only needs to be modified a bit
* multiply the amount of the trade with the selected pricecolumn at that point (probably open or close price)
* combine all "value" columns (output of the above) into one dataframe, one column per pair
* sum over all columns -> get total amount per candle.

The calculation (`1 - open_rate / close - slippage`)  should not be necessary - and is wrong to me - unless i completely misunderstood what you're trying to do.

orehunt:
>     * create a timeseries dataframe (per pair) with the total length of the timerange - exploding every trade into multiple lines (one line per candle) - code for this is available from `analyze_trade_parallelism()` and only needs to be modified a bit
I did it this way also for speed, this is very fast (although it also uses mostly numpy)
>     * multiply the amount of the trade with the selected pricecolumn at that point (probably open or close price)
this doesn't work if you have position stacking enabled, you need to keep track of the duration of each trade..unless I am mistaken the `BacktestResult` doesn't have amounts neither in base nor in quote, but with the base amount maybe it could be done by collapsing all the amounts...or not mmh not sure 
> * combine all "value" columns (output of the above) into one dataframe, one column per pair
> 
> * sum over all columns -> get total amount per candle.
this is what's done at the end

> The calculation (`1 - open_rate / close - slippage`) should not be necessary - and is wrong to me - unless i completely misunderstood what you're trying to do.

you're right in the sense that calculating it this way, it assumes an amount of `1` for each pair, so it would have unequal amounts between pairs, but it should be inconsequential for what I was looking for, but it could be fixed, you would have to replace `open_rate` with `stake / open_rate`.

freqtrade/data/btanalysis.py
```diff
@@ -213,3 +221,93 @@ def calculate_max_drawdown(trades: pd.DataFrame, *, date_col: str = 'close_time'
...
+            .set_index("open_time")
+            .resample(timeframe)
+            .asfreq()
+            .reindex(date_index)
+        )
+        open_rate = pair_trades["open_rate"].fillna(0).values
+        open_time = pair_trades.index.values
+        close_time = pair_trades["close_time"].values
+        close = pair_candles["close"].values
+        profits = pair_trades["profit_percent"].values - slippage
```
xmatthias: This column does not exist - neither in data loaded from a backtest file, nor from data loaded form database.

ghost: mmmh? https://github.com/freqtrade/freqtrade/blob/25d9001f8864d6a641037f7d6f8e0177df18d752/freqtrade/optimize/optimize_reports.py#L21-L24

orehunt:
I added an implementation as a loss function based on sortino.
Bear in mind that this is basically an upsampling, which means, it adds noise, which is not what you generally want in smaller timeframes; So it could be useful as a buy and hold metric with like `1d` timeframes and `3-7d` trades duration where you can see volatility day by day

freqtrade/data/btanalysis.py
xmatthias: nope - this is how it's internally called - but when loading data from file, it's aligned to the database.

https://github.com/freqtrade/freqtrade/blob/25d9001f8864d6a641037f7d6f8e0177df18d752/freqtrade/data/btanalysis.py#L19-L20

freqtrade/data/btanalysis.py
```diff
@@ -213,3 +221,93 @@ def calculate_max_drawdown(trades: pd.DataFrame, *, date_col: str = 'close_time'
...
+    :param min_date: date of the first trade opened (results.open_time.min())
+    :param max_date: date of the last trade closed (results.close_time.max())
+    :param hloc: historical DataFrame of each pair tested
+    :slippage: optional profit value to subtract per trade
+    :return: DataFrame of outstanding balance at each timeframe
+    """
+    timedelta = pd.Timedelta(timeframe)
+
+    date_index: pd.DatetimeIndex = pd.date_range(
+        start=min_date, end=max_date, freq=timeframe, normalize=True
```
xmatthias: This is wrong when used for short timeframes.

Assuming "5m" - if that's passed to pd.date_range - it'll create an index with a frequency of "5M" - which is 5 months.

what should be used instead is something along the following lines (a sample of this can be found in analyze_trade_parallelism).

``` python
    from freqtrade.exchange import timeframe_to_minutes
    timeframe_min = timeframe_to_minutes(timeframe)
    ... 
    freq=f"{timeframe_min}min"
```

freqtrade/data/btanalysis.py
xmatthias: I've merged current develop into this and updated this back to profit_percent - as we changed / aligned freqtrade now to use that as column value throughout.

orehunt:
I imagine that doesn't work with position stacking? 
```    # df3 has the dates set correctly (one date per candle / trade). Date is only unique per pair.```
This makes me think so

**Comment by xmatthias** [comment]
it should work with position stacking as well ... 

it was just a quick comment thrown there to explain what it does ... 
Mostly to explain that date is NOT unique - if you look above these lines you'll notice that it's not per pair, but per trade.

**Comment by xmatthias** [comment]
> and maybe do a cumsum and plot it like plot-profit

Cumsum cannot be used on this - the output is the total balance in trades at any point in time ... 
if i own 100 now, and 101$ in 1 hour - then i won't own 201$ (which would be the outcome of a cumsum).

---

I'm also not certain on the usefulness of this in a loss function - all it'll do is analyze how much money was in a trades at any point in time - it'll not give you the account balance at that moment, so based on this function, you can't determine if the bot is successful or not.


This function can however be useful to analyze how much money is in trades at any point in time - so you might determine to use a lower (or higher) max_open_trades.

File: contrib/loss_functions/SortinoLossBalance.py [hunk 3]
```diff
@@ -0,0 +1,133 @@
...
+        min_date: datetime,
+        max_date: datetime,
+        *args,
+        **kwargs,
+    ) -> float:
+        """
+        Objective function, returns smaller number for more optimal results.
+        Uses Sortino Ratio calculation.
+        """
+        hloc = kwargs["processed"]
```
flomerz: How can you access the "processed" data? In my loss function, they are not in the kwargs.

xmatthias: i don't think this is a working example as is ... 

following the PR history, it's an example to show how the data provided via the function below can / could be used.

PR #3062, opened 2020-03-14
Plotting: Fix if no file exists and new skip option

Fredrik81:
## Summary
Fixes issue where no back testing file exist. Will plot with no trades if file is missing and DB option is not specified.
Also adds a complementary option to skip trades even if file exists.
```
freqtrade/user_data/backtest_results/backtest-result.json does not exist.
```

PEP8 = yes

## Quick changelog

- Fix that plot errors if backtest file is missing
- New argument --skip-trades

## What's new?
Introduce --skip-trades to allow skipping generation of trades in plot even if file/DB trades exists


freqtrade/commands/cli_options.py
```diff
@@ -413,6 +413,11 @@ def __init__(self, *args, **kwargs):
         metavar='INT',
         default=750,
     ),
+    "skip_trades": Arg(
+        '--skip-trades',
+        help='Skip using trades file from backtesting and DB.',
+        action='store_true',
+    ),
```
xmatthias: Personally, i would prefer to use `--no-trades`. 

While technically the trades loading is skipped, no-trades seems more intuitive to me - but it's a personal preference - so if you prefer skip-trades that's fine, too

freqtrade/plot/plotting.py
```diff
@@ -48,11 +49,21 @@ def init_plotscript(config):
...
-    trades = load_trades(config['trade_source'],
-                         db_url=config.get('db_url'),
-                         exportfilename=config.get('exportfilename'),
-                         )
+    skip_trades = False
+    if not isfile(config.get('exportfilename')) and config['trade_source'] == 'file':
+        logger.warning("Backtest file is missing skipping trades.")
+        skip_trades = True
+    elif config.get('skip_trades', False):
+        skip_trades = True
```
xmatthias: I think you'll need to revert the sequence of the if statements, and check the explicit `skip_trades` parameter first.

If i call this with `--export-filename nonexistingfile.json --skip-trades` - it's printing "Backtestfile is missing ..." - but it should not (same without --export-filename if the file does not exist - but i'm too lazy to delete the file).

freqtrade/plot/plotting.py
```diff
@@ -48,11 +49,21 @@ def init_plotscript(config):
         data_format=config.get('dataformat_ohlcv', 'json'),
     )
 
-    trades = load_trades(config['trade_source'],
-                         db_url=config.get('db_url'),
-                         exportfilename=config.get('exportfilename'),
-                         )
+    skip_trades = False
+    if not isfile(config.get('exportfilename')) and config['trade_source'] == 'file':
```
xmatthias: #3064 should help this by having exportfilename as `pathlib.Path` right out of the configuration - so `config['exportfilename'].is_file()` can be used.

Also, if you look at `configuration.py` - this parameter is always set to the default - so it's safe to assume it'll always be there.

freqtrade/data/btanalysis.py
```diff
@@ -129,12 +129,16 @@ def load_trades_from_db(db_url: str) -> pd.DataFrame:
...
 
 
-def load_trades(source: str, db_url: str, exportfilename: str) -> pd.DataFrame:
+def load_trades(source: str, db_url: str, exportfilename: str, skip_trades: bool) -> pd.DataFrame:
     """
     Based on configuration option "trade_source":
     * loads data from DB (using `db_url`)
     * loads data from backtestfile (using `exportfilename`)
     """
+    if skip_trades:
```
xmatthias: I would call the parameter to this function either "empty" or "no_trades" (irrelevant of the cli argument naming) - assuming that this function is used by other means too (interactive analysis, for example) that naming will give the parameter a clearer meaning.

freqtrade/plot/plotting.py
hroff-1902: merged #3064 to develop and resolved conflict here (pls verify it)

freqtrade/plot/plotting.py
Fredrik81: fixed in next commit

freqtrade/plot/plotting.py
Fredrik81: It works in latest commit

freqtrade/commands/cli_options.py
Fredrik81: changes to --no-trades

docs/plotting.md
```diff
@@ -23,54 +23,74 @@ The `freqtrade plot-dataframe` subcommand shows an interactive graph with three
...
+usage: freqtrade plot-dataframe [-h] [-v] [--logfile FILE] [-V] [-c PATH]
+                                [-d PATH] [--userdir PATH] [-s NAME]
+                                [--strategy-path PATH] [-p PAIRS [PAIRS ...]]
+                                [--indicators1 INDICATORS1 [INDICATORS1 ...]]
+                                [--indicators2 INDICATORS2 [INDICATORS2 ...]]
+                                [--plot-limit INT] [--db-url PATH]
+                                [--trade-source {DB,file}] [--export EXPORT]
+                                [--export-filename PATH]
+                                [--timerange TIMERANGE] [-i TICKER_INTERVAL]
+                                [--skip-trades]
```
xmatthias:
```suggestion
                                [--no-trades]
```

freqtrade/data/btanalysis.py
```diff
@@ -129,16 +129,21 @@ def load_trades_from_db(db_url: str) -> pd.DataFrame:
     return trades
 
 
-def load_trades(source: str, db_url: str, exportfilename: Path) -> pd.DataFrame:
+def load_trades(source: str, db_url: str, exportfilename: Path, no_trades: bool) -> pd.DataFrame:
```
xmatthias: I'd default this to False - so it's not a required argument.

```suggestion
def load_trades(source: str, db_url: str, exportfilename: Path, no_trades: bool = False) -> pd.DataFrame:
```

tests/data/test_btanalysis.py
```diff
@@ -105,6 +105,7 @@ def test_load_trades(default_conf, mocker):
     load_trades("DB",
                 db_url=default_conf.get('db_url'),
                 exportfilename=default_conf.get('exportfilename'),
+                no_trades=False
```
xmatthias: I mainly noticed the above comments via default because of this ... so i'd probably revert this line (and line 120)

docs/plotting.md
Fredrik81: Thanks! missed that

freqtrade/data/btanalysis.py
Fredrik81: Thats nice!

tests/data/test_btanalysis.py
Fredrik81: Well i guess it's not needed but it's also not harming.

xmatthias:
THIS is why i don't like the "suggest" functionality of github ... 

While it's easy to commit, it's just as easy to create a flake (or mypy) problem without noticing before adding the commit.

PR #3090, opened 2020-03-21
Remove partial candle documentation

### LINKED ISSUE #2993: ohlcv_partial_candle how to use it ?
Hi,

I don't understand how `ohlcv_partial_candle` param can be useful for trading :
- an incomplete candle is not really a candle because his shape change at every trade so it cannot be consider as a reliable information.
- maybe i am wrong but freqtrade refresh the candles only once per timeframe so this code coming from docs below don't reflect what happens in freqtrade has it trigger once per candle and not once per `process_throttle_secs` ?

EDIT : the doc seems to imply that this code is executed many times

> Another way is to run this command multiple times in a row and observe if the volume is changing (while the date remains the same).

```
import ccxt
from datetime import datetime
from freqtrade.data.converter import parse_ticker_dataframe
ct = ccxt.binance()
timeframe = "1d"
pair = "XLM/BTC"  # Make sure to use a pair that exists on that exchange!
raw = ct.fetch_ohlcv(pair, timeframe=timeframe)

# convert to dataframe
df1 = parse_ticker_dataframe(raw, timeframe, pair=pair, drop_incomplete=False)

print(df1.tail(1))
print(datetime.utcnow()) 
```

I can understand we use the ticker/current price in a strategy to compute every `process_throttle_secs` but the ohlcv_partial_candle i don't get it.

I maybe missed something.

Have a nice day.

xmatthias:
## Summary
The partial candle method  was never fully tested - and does not work correctly since we're not fetching candles on every iteration - but only when new candles are actually expected.

closes #2993

## Quick changelog

- Remove wrong documentation part - suggesting alternatively using the dataprovider to fetch the latest price.

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -83,4 +83,4 @@ Whether your exchange returns incomplete candles or not can be checked using [th
 
 Due to the danger of repainting, Freqtrade does not allow you to use this incomplete candle.
 
-However, usually, this requirement is based on the need for the latest price - which can be aquired using the [data provider](strategy-customization.md#possible-options-for-dataprovider) from within the strategy.
+However, if it is based on the need for the latest price for your strategy - then this requirement can be acquired using the [data provider](strategy-customization.md#possible-options-for-dataprovider) from within the strategy.
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -74,23 +74,13 @@ Should you experience constant errors with Nonce (like `InvalidNonce`), it is be
 $ pip3 install web3
 ```
 
-### Send incomplete candles to the strategy
+### Getting latest price / Incomplete candles
 
 Most exchanges return current incomplete candle via their OHLCV/klines API interface.
 By default, Freqtrade assumes that incomplete candle is fetched from the exchange and removes the last candle assuming it's the incomplete candle.
 
 Whether your exchange returns incomplete candles or not can be checked using [the helper script](developer.md#Incomplete-candles) from the Contributor documentation.
 
-If the exchange does return incomplete candles and you would like to have incomplete candles in your strategy, you can set the following parameter in the configuration file.
+Due to the danger of repainting, Freqtrade does not allow you to use this incomplete candle.
 
-``` json
-{
-
-    "exchange": {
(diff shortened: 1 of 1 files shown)


docs/exchanges.md
```diff
@@ -74,23 +74,13 @@ Should you experience constant errors with Nonce (like `InvalidNonce`), it is be
...
-
-    "exchange": {
-        "_ft_has_params": {"ohlcv_partial_candle": false}
-    }
-}
-```
-
-!!! Warning "Danger of repainting"
-    Changing this parameter makes the strategy responsible to avoid repainting and handle this accordingly. Doing this is therefore not recommended, and should only be performed by experienced users who are fully aware of the impact this setting has.
+However, usually, this requirement is based on the need for the latest price - which can be aquired using the [data provider](strategy-customization.md#possible-options-for-dataprovider) from within the strategy.
```
hroff-1902:
```suggestion
However, if it is based on the need for the latest price for your strategy - then this requirement can be acquired using the [data provider](strategy-customization.md#possible-options-for-dataprovider) from within the strategy.
```

docs/exchanges.md
```diff
@@ -74,23 +74,13 @@ Should you experience constant errors with Nonce (like `InvalidNonce`), it is be
 $ pip3 install web3
 ```
 
-### Send incomplete candles to the strategy
+### Getting latest price / Incomplete candles
 
 Most exchanges return current incomplete candle via their OHLCV/klines API interface.
 By default, Freqtrade assumes that incomplete candle is fetched from the exchange and removes the last candle assuming it's the incomplete candle.
 
 Whether your exchange returns incomplete candles or not can be checked using [the helper script](developer.md#Incomplete-candles) from the Contributor documentation.
```
hroff-1902:
```suggestion
Whether your exchange returns incomplete candles or not - it can be checked using [the helper script](developer.md#Incomplete-candles) from the Contributor documentation.
```

xmatthias: really? to me this sounds strange with the seperating word (when trying to read it out loud...)

hroff-1902: > really? to me this sounds strange with the seperating word (when trying to read it out loud...)

no problem, that was only my suggestion...

PR #2905, opened 2020-02-11
Adding --min-trades and --max-trades for hyperopt-list

Fredrik81:
## Summary
Adding more filter options for hyperopt-list to easier find what your looking for in larger runs

PEP8 = Yes
## Quick changelog
Adding two more filter options that can be used in combination to the other options in order to filter and search thru larger hyperopt runs.

- Adding filter options --max-trades and --min-trades
- Adding missing filter options --max-avg-profit and --max-total-profit
- Added additional tests for the new options

## What's new?
```
freqtrade hyperopt-list --max-avg-time 200 --min-trades 20 --max-avg-profit 0.29 --min-avg-profit 0.20 --max-total-profit 8
2020-02-11 21:01:47,383 - freqtrade.loggers - INFO - Verbosity set to 0
2020-02-11 21:01:47,383 - freqtrade.configuration.configuration - INFO - Parameter --min-trades detected: 20
2020-02-11 21:01:47,383 - freqtrade.configuration.configuration - INFO - Parameter --max-avg-time detected: 200.0
2020-02-11 21:01:47,383 - freqtrade.configuration.configuration - INFO - Parameter --min-avg-profit detected: 0.2
2020-02-11 21:01:47,383 - freqtrade.configuration.configuration - INFO - Parameter --max-avg-profit detected: 0.29
2020-02-11 21:01:47,383 - freqtrade.configuration.configuration - INFO - Parameter --max-total-profit detected: 8.0
2020-02-11 21:01:47,383 - freqtrade.configuration.check_exchange - INFO - Checking exchange...
2020-02-11 21:01:47,383 - freqtrade.configuration.configuration - INFO - Using pairlist from configuration.
2020-02-11 21:01:47,383 - freqtrade.configuration.config_validation - INFO - Validating configuration ...
2020-02-11 21:01:47,574 - freqtrade.optimize.hyperopt - INFO - Loaded 1096 previous evaluations from disk.
2020-02-11 21:01:47,574 - freqtrade.commands.hyperopt_commands - INFO - 1 epochs found.
   602/1096:     29 trades. Avg profit   0.27%. Total profit  0.00046184 BTC (   7.69Σ%). Avg duration 189.8 min. Objective: 2.07836

Best result:

   602/1096:     29 trades. Avg profit   0.27%. Total profit  0.00046184 BTC (   7.69Σ%). Avg duration 189.8 min. Objective: 2.07836
```

PR #3091, opened 2020-03-22
--min-objective and --max-objective for hyperopt-list

yazeed:
### Summary

A follow-up PR to @Fredrik81's #2905 that adds --min-objective and --max-objective flags for use with **freqtrade hyperopt-list**.

I left out two lines intentionally commented out `# trials = [x for x in trials if x['loss'] != 20]` as I wasn't sure if it was a good idea to filter out non-optimum epochs that had objective == 20 with both directives, or to leave those results for the user to filter them out manually with --min-objective 0 for example, specially since some users may be using non-default loss methods, that don't necessarily use 20 for non-optimum epochs.

The way this currently works is as follows (notice you don't need to use the minus(-) sign):

`freqtrade hyperopt-list --profitable --min-objective 100`

resulting in the following table:

```
+--------+-----------+----------+--------------+------------------------------+----------------+-------------+
|   Best |     Epoch |   Trades |   Avg profit |                       Profit |   Avg duration |   Objective |
|--------+-----------+----------+--------------+------------------------------+----------------+-------------|
|   Best | 3116/4612 |        3 |        1.17% |   10.50000000 BTC    (3.50%) |        200.0 m |    -142.196 |
|   Best | 3449/4612 |        3 |        1.83% |   16.50000000 BTC    (5.50%) |        135.0 m |    -227.09  |
|        | 4077/4612 |        5 |        1.10% |   16.52180207 BTC    (5.51%) |        123.0 m |    -157.793 |
|        | 4189/4612 |        2 |        2.00% |   12.00000000 BTC    (4.00%) |        127.5 m |    -165.542 |
+--------+-----------+----------+--------------+------------------------------+----------------+-------------+
```


yazeed:
Not sure how exactly to handle the flake8 C901 '_hyperopt_filter_trials' is too complex error.

xmatthias:
> Not sure how exactly to handle the flake8 **C901 '_hyperopt_filter_trials' is too complex** error, but this currently works, with its own tests in place.

This happens for functions which have too many if  branches ... which increases complexity - so if you're to debug one of these functions, you'd need to consider X different options and branches - which is definitely true for such a function.

You can add a  `# noqa: C901` to the function definition line to skip this error if the function cannot be refactored to be less nested.

freqtrade/commands/cli_options.py
```diff
@@ -484,6 +484,18 @@ def __init__(self, *args, **kwargs):
...
     ),
+    "hyperopt_list_min_objective": Arg(
+        '--min-objective',
+        help='Select epochs on above objective.',
+        type=float,
+        metavar='FLOAT',
+    ),
+    "hyperopt_list_max_objective": Arg(
+        '--max-objective',
+        help='Select epochs on below objective.',
```
xmatthias: @hroff-1902 Is this wording intentional or was it missed in the other PR's?

It's in line with the other "hyperopt_list_*" options - but it sounds kinda odd to me ... (both for this, and for some other options ...)

wouldn't `"Select epochs below objective."` and `"Select epochs above objective."` sound like ... better english?

freqtrade/commands/hyperopt_commands.py
```diff
@@ -35,9 +35,16 @@ def start_hyperopt_list(args: Dict[str, Any]) -> None:
...
+        'filter_max_total_profit': config.get('hyperopt_list_max_total_profit', None),
+        'filter_min_objective': config.get('hyperopt_list_min_objective', None),
+        'filter_max_objective': config.get('hyperopt_list_max_objective', None)
     }
 
+    if filteroptions['filter_min_objective'] is not None:
+        filteroptions['filter_min_objective'] = -filteroptions['filter_min_objective']
+    if filteroptions['filter_max_objective'] is not None:
+        filteroptions['filter_max_objective'] = -filteroptions['filter_max_objective']
+
```
xmatthias: What's the purpose of this? 
is it because objective is "negativized" in the hyperoptloss functions, too?

 (i'm asking because this caused the C901 error)

freqtrade/commands/cli_options.py
yazeed: The wording is wrong, but I basically copied and pasted from other flags, I'll fix them.

freqtrade/commands/hyperopt_commands.py
yazeed: Yes, the purpose is to multiply the given input by -1 to make it minus matching loss in hyperoptloss functions.

If you think about it, in this context, when filtering out results from the hyperopt-list, you'd consider -100 objective higher than -10 objective, so you'd use -min-objective 10 to show any epochs with -10 and lower, having it done the other way around was a bit confusing.

yazeed: > What's the purpose of this?
> is it because objective is "negativized" in the hyperoptloss functions, too?
> 
> (i'm asking because this caused the C901 error)

The C901 error was caused by the extended **_hyperopt_filter_trials()** method, not because of the code block above, to me at least.

freqtrade/commands/hyperopt_commands.py
```diff
@@ -35,9 +35,16 @@ def start_hyperopt_list(args: Dict[str, Any]) -> None:
         'filter_min_avg_profit': config.get('hyperopt_list_min_avg_profit', None),
         'filter_max_avg_profit': config.get('hyperopt_list_max_avg_profit', None),
         'filter_min_total_profit': config.get('hyperopt_list_min_total_profit', None),
-        'filter_max_total_profit': config.get('hyperopt_list_max_total_profit', None)
+        'filter_max_total_profit': config.get('hyperopt_list_max_total_profit', None),
+        'filter_min_objective': config.get('hyperopt_list_min_objective', None),
+        'filter_max_objective': config.get('hyperopt_list_max_objective', None)
```
hroff-1902:
```suggestion
        'filter_max_objective': config.get('hyperopt_list_max_objective', None),
```

hroff-1902:
First note: I'll get rid of this artificial value 20 and probably another one, 10000 used in the optimization. Loss functions should raise exception indicating that they cannot calculate the loss function (division by zero variance, or a list of aggregated trades of only one/two rows, which is not enough to calculate the variance, etc.) -- which is to be properly catched above in the code and printed in the hyperopt results. Maybe the value of 10000 used can also be removed (I thought of infinity, but it produces exception in skopt, further investigations needed).

Second:
Well, overall, these new options may be useful. But I'm against using negative values.

For --min-objective=10 and --max-objective=100 in case of SharpeLoss, for example, it's clear.
But what if I want to filter out epochs with the values of Sharpe Ratio, which is only a bit better, than that -20 bad value used, say, --min-objective=-18... For Sharpe Ratio it can also be explained. But what about our composite loss function DefaultHyperoptLoss? It's "negated" (negativated?) value does not have mathematical meaning, it is composed and used as it is. Same for other metrics that can be introduced in the future and that behave like loss function (the less the better) -- their "negated" value will also not have any mathematical or statistical value. As the result -- this will be hard to explain in the docs.

A workaround here could be moving from the term "loss" function to "objective", and printing everywhere (in the results, in these options) negated values comparing to the current ones. So, after this, hyperopt will simply print the Sharpe Ratio in the last column, the --min-objective and --max-objective will define the min and max boundaries for the Sharpe Ratio (if Sharpe is used as objective), and the hyperopt will optimize to the maximum of the objective function. Legacy DefaultHyperoptLoss can simply be negated to get new objective corresponding to it. But we have a problem here -- Loss is used in the name of the --hyperopt-loss option and in the names of the HyperOptLoss classes... Deprecation is always a pain... (But let's consider this move anyway)

Without this move (from Loss to Objective), I would propose using same boundaries as the values printed in the last column. So `--min-objective -10 --max-objective -1 --hyperopt-loss SharpeHyperOptLoss` would print the epochs with the value of the last column between those -1 and -10 (i.e. with Sharpe Ratio between 1 and 10).

freqtrade/commands/hyperopt_commands.py
yazeed: Resolved

yazeed:
> A workaround here could be moving from the term "loss" function to "objective", and printing everywhere (in the results, in these options) negated values comparing to the current ones. So, after this, hyperopt will simply print the Sharpe Ratio in the last column, the --min-objective and --max-objective will define the min and max boundaries for the Sharpe Ratio (if Sharpe is used as objective), and the hyperopt will optimize to the maximum of the objective function. Legacy DefaultHyperoptLoss can simply be negated to get new objective corresponding to it. But we have a problem here -- Loss is used in the name of the --hyperopt-loss option and in the names of the HyperOptLoss classes... Deprecation is always a pain... (But let's consider this move anyway)

I agree with the rationale mentioned regarding moving to "objective" as opposed to "loss" by definition, and with negating its use with the DefaultHyperoptLoss method. I wouldn't know how to apply all changes as I'm not aware of all touchpoints needed to be modified, to continue with this PR along those lines.

yazeed:
> Without this move (from Loss to Objective), I would propose using same boundaries as the values printed in the last column. So `--min-objective -10 --max-objective -1 --hyperopt-loss SharpeHyperOptLoss` would print the epochs with the value of the last column between those -1 and -10 (i.e. with Sharpe Ratio between 1 and 10).

This would be very confusing for many with Sortino and Sharpe methods, I think, the min would be the max, and the max the min.

How do you guys propose we continue on this PR?

xmatthias:
Sharpe:

```
+--------+---------+----------+--------------+------------------------------+----------------+-------------+
|   Best |   Epoch |   Trades |   Avg profit |                       Profit |   Avg duration |   Objective |
|--------+---------+----------+--------------+------------------------------+----------------+-------------|
|   Best |   1/100 |      491 |       -0.26% |  -38.76196249 EUR (-129.00%) |        682.5 m |     2.46783 |
.

|   Best |   2/100 |      403 |       -0.21% |  -25.81175432 EUR  (-85.90%) |        718.7 m |     1.83009 |
.

|   Best |   3/100 |       83 |       -0.30% |   -7.46062969 EUR  (-24.83%) |        867.2 m |     1.24167 |
..

|   Best |   5/100 |       63 |       -0.04% |   -0.83158867 EUR   (-2.77%) |        899.7 m |     0.31035 |
...

|   Best |   8/100 |      124 |        0.34% |   12.55343414 EUR   (41.78%) |        932.3 m |    -1.60956 |
.....

|   Best |  13/100 |      115 |        0.57% |   19.70619258 EUR   (65.58%) |        957.4 m |    -2.98028 |
...

|   Best |  16/100 |       70 |        1.12% |   23.50166603 EUR   (78.21%) |        708.9 m |    -5.46494 |
```

Profit Only:
```
+--------+---------+----------+--------------+------------------------------+----------------+-------------+
|   Best |   Epoch |   Trades |   Avg profit |                       Profit |   Avg duration |   Objective |
|--------+---------+----------+--------------+------------------------------+----------------+-------------|
|   Best |   1/100 |      491 |       -0.26% |  -38.76196249 EUR (-129.00%) |        682.5 m |        1.43 |
.

|   Best |   2/100 |      403 |       -0.21% |  -25.81175432 EUR  (-85.90%) |        718.7 m |     1.28634 |
.

|   Best |   3/100 |       83 |       -0.30% |   -7.46062969 EUR  (-24.83%) |        867.2 m |     1.08276 |
..

|   Best |   5/100 |       63 |       -0.04% |   -0.83158867 EUR   (-2.77%) |        899.7 m |     1.00923 |
...

|   Best |   8/100 |      124 |        0.34% |   12.55343414 EUR   (41.78%) |        932.3 m |     0.86074 |
.....

|   Best |  13/100 |      115 |        0.57% |   19.70619258 EUR   (65.58%) |        957.4 m |     0.78139 |
...

|   Best |  16/100 |       70 |        1.12% |   23.50166603 EUR   (78.21%) |        708.9 m |     0.73929 |
```

hroff-1902:
@xmatthias let's refuse and close it? (see the rationale above in our comments)

xmatthias:
Why?
the only argument / critique we're having is about the "negativation" of the input values - which is exactly 4 lines of code of this whole PR.
The rest is good in my opinion ... and will be useful to filter out hyperopt results .

PR #3093, opened 2020-03-22
Add close_profit_abs column

### LINKED ISSUE #3010: Record profits in absolute value
This is a proposal to create a new column **trades.close_profit_abs** that records profit in absolute value. I think it can be used in different ways, such as easily aggregating total profits in stake_currency value, (would simplify _rpc_trade_statistics() for example), and it can be used in other references to calculate stake_currency based profit.

There's also lots of checks on the existing **Trade.close_profit** (percentage based) so we'd have to make sure we continue using it as a source of truth, and that we don't approve future PRs using the new column as a new source of truth, for consistency.

Let me know if you have any thoughts on this.

xmatthias:
## Summary
Add column `close_profit_abs` to the database, to simplify analysis afterwards (aligned to close_profit - which represents the ratio.

closes #3010
## Quick changelog

- Add new column close_profit_abs
- add migration test case for closed trade
- Simplify rpc calls where possible

### DIFF
--- a/tests/test_persistence.py
+++ b/tests/test_persistence.py
@@ -476,12 +476,22 @@ def test_migrate_old(mocker, default_conf, fee):
                                      stake=default_conf.get("stake_amount"),
                                      amount=amount
                                      )
+    insert_table_old2 = """INSERT INTO trades (exchange, pair, is_open, fee,
+                          open_rate, close_rate, stake_amount, amount, open_date)
+                          VALUES ('BITTREX', 'BTC_ETC', 0, {fee},
+                          0.00258580, 0.00268580, {stake}, {amount},
+                          '2017-11-28 12:44:24.000000')
+                          """.format(fee=fee.return_value,
+                                     stake=default_conf.get("stake_amount"),
+                                     amount=amount
+                                     )
     engine = create_engine('sqlite://')
     mocker.patch('freqtrade.persistence.create_engine', lambda *args, **kwargs: engine)
 
     # Create table using the old format
     engine.execute(create_table_old)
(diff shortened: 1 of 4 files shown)


tests/test_persistence.py
```diff
@@ -476,12 +476,22 @@ def test_migrate_old(mocker, default_conf, fee):
...
                                      )
+    insert_table_old2 = """INSERT INTO trades (exchange, pair, is_open, fee,
+                          open_rate, close_rate, stake_amount, amount, open_date)
+                          VALUES ('BITTREX', 'BTC_ETC', 0, {fee},
+                          0.00258580, 0.00268580, {stake}, {amount},
+                          '2017-11-28 12:44:24.000000')
+                          """.format(fee=fee.return_value,
+                                     stake=default_conf.get("stake_amount"),
+                                     amount=amount
+                                     )
```
hroff-1902: did not get this piece
(however, I'm not a db-geek, assume this is okay)

xmatthias: well to test if `close_profit_abs` is calculated correctly during database migration (and aligned to the python calculation), we need a closed trade too - since it'll not calculate `close_profit_abs` for open trades.

hroff-1902: I mean why it's introduced as a sep. insert_table_old2 if it's executed sequentially:
```
    engine.execute(insert_table_old)
    engine.execute(insert_table_old2)
```
?

xmatthias: because it's cleaner than using 2 statements in one execute - and because `insert into values ()` does not support multiple rows (at least not on all databases)

Issue #2696, opened 2019-12-22
Strategy validations 


xmatthias:
## Problem description

We've seen that the dataframe we pass to the strategy is easy to break by a user if not being careful.
Examples include (but are not limited to):
* [x] Dropping rows from the dataframe (#3107)
* [x] ~Returning resampled dataframe~ (implicitly captured by checking for length)
* [x] adding rows to the dataframe (#3107)
* [x] modifying the dataframe (#3107)
* [x] not adding "buy" column
* [x] not adding "sell" column
* [x]  always returning 0 in the "buy" column
* [x] not returning the dataframe
* [ ] Using a unrealistic stoploss (too tight)
* ...
* ~syntax errors~ (not relevant since syntax-errors will prevent the class from loading, so we cannot run checks on it).

Most of these scenarios are difficult to analyze by inexperienced users - but it should be possible to automate most of that.

Ideally, we'd have a `freqtrade test-strategy` subcommand, which runs a few tests against the strategy and tells the user (before even running backtesting / dry run) if the strategy is formally correct.

this "test_command" could also run on bot startup (optional?) - so we know before the first iteration if the strategy is breaking the dataframe.

it should not run on every iteration, since that would be a waste of resources, and we can assume that if a strategy works once, it'll work again.

PR #3107, opened 2020-03-24
check that the strategy dataframe matches the one given by the bot

orehunt:
## Summary
Save a few dataframe elements before calling the signals, to make sure they are
still there afterwards.
Also fetch the correct signals by date.

related to #2696 
## Quick changelog

- functions `preserve_df` and `assert_df` used around strategy calls to gauge dataframe validity
- locate the correct signals row by date instead of by index (which is kinda reduntant since with those checks it will raise an exception if the index does not match...

...the failing tests are...broken because they indeed pass a non matching dataframe..
these checks are done on each call...they are not heavy so...doesn't seem to hurt

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -241,8 +241,18 @@ class IStrategy(ABC):
 
         return dataframe
 
-    def get_signal(self, pair: str, interval: str,
-                   dataframe: DataFrame) -> Tuple[bool, bool]:
+    @staticmethod
+    def preserve_df(d: DataFrame) -> Tuple[int, float, datetime]:
+        """ keep some data for dataframes """
+        return len(d), d["close"].iloc[-1], d["date"].iloc[-1]
+
+    @staticmethod
+    def assert_df(d: DataFrame, df_len: int, df_close: float, df_date: datetime):
+        """ make sure data is unmodified """
+        if df_len != len(d) or df_close != d["close"].iloc[-1] or df_date != d["date"].iloc[-1]:
+            raise Exception("Dataframe returned from strategy does not match original")
+
+    def get_signal(self, pair: str, interval: str, dataframe: DataFrame) -> Tuple[bool, bool]:
         """
         Calculates current signal based several technical analysis indicators
         :param pair: pair in format ANT/BTC
@@ -254,8 +264,11 @@ class IStrategy(ABC):
             logger.warning('Empty candle (OHLCV) data for pair %s', pair)
             return False, False
 
+        latest_date = dataframe['date'].max()
         try:
+            df_len, df_close, df_date = self.preserve_df(dataframe)
             dataframe = self._analyze_ticker_internal(dataframe, {'pair': pair})
+            self.assert_df(dataframe, df_len, df_close, df_date)
         except ValueError as error:
             logger.warning(
                 'Unable to analyze candle (OHLCV) data for pair %s: %s',
@@ -275,7 +288,7 @@ class IStrategy(ABC):
             logger.warning('Empty dataframe for pair %s', pair)
             return False, False
 
-        latest = dataframe.iloc[-1]
+        latest = dataframe.loc[dataframe['date'] == latest_date].iloc[-1]
 
         # Check if dataframe is out of date
         signal_date = arrow.get(latest['date'])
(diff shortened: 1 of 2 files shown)


hroff-1902:
the checks are fine, but I would like to see diff. texts in the exception for diff. checks (length mismatch, datetime mismatch, etc,)

I could add it by myself...

Also, the freqtrade exception (DependencyException?) should be raised instead of generic Exception...

freqtrade/strategy/interface.py
```diff
@@ -241,8 +242,26 @@ def _analyze_ticker_internal(self, dataframe: DataFrame, metadata: dict) -> Data
...
+        message = ""
+        if df_len != len(dataframe):
+            message = "length"
+        elif df_close != dataframe["close"].iloc[-1]:
+            message = "last close price"
+        elif df_date != dataframe["date"].iloc[-1]:
+            message = "last date"
+        if message:
+            raise DependencyException("Dataframe returned from strategy has mismatching "
+                                      f"{message}.")
```
xmatthias: I'm not certain about the type of the exception yet - `Exception` (what you had before) was definitely wrong ... 

but we'll need to ask the question - what do we want to archive?

should the bot be stopped completely since the strategy is wrong? 
or should just this iteration / cycle be interrupted and retried next time (assuming it's a general error - i guess that would lead to the bot being stuck / seeming stuck).

If we want it stopped, we can remove / skip this check after it ran once saving processing time - assuming the strategy won't act differently after the first iteration.
If on the other hand we assume a temporary error (not sure that's even possible without intentionally doing that) - then we'll need to run it each time - but can't crash the bot completely, but need to skip this iteration.

orehunt: If it catches the error on the first iteration it's probably a configuration error, so it stops right away.
If it appears after a while, the strategy is relieing on some currently unavailable data, should keep trying.

xmatthias:
I did add some tests to this PR - mainly to ensure myself that it's working as i expect it to.

I've also "reformatted" the error output - we should not print the full stacktrace in this case to avoid confusion from users (this will further be changed / enhanced when #2872 is merged as that introduces the StrategyException...).

freqtrade/strategy/interface.py
```diff
@@ -275,7 +298,7 @@ def get_signal(self, pair: str, interval: str,
             logger.warning('Empty dataframe for pair %s', pair)
             return False, False
 
-        latest = dataframe.iloc[-1]
+        latest = dataframe.loc[dataframe['date'] == latest_date].iloc[-1]
```
xmatthias: Odd question - do we need the `.iloc[-1]` here? 

the only case where `dataframe.loc[dataframe['date'] == latest_date]` would return more than one line is if i drop one line - and duplicate the last line so i got the same date twice ... (or am i mistaking...)?

orehunt: if you don't add iloc, it will return a dataframe, so it would need some other changes below but haven't checked, anyway the functions only check specific rows, so it could be possible that the returned df might have the same dates in between also considering that the dates are not the index

xmatthias: ah right, it's a datatype issue - didn't think about that :+1:

hroff-1902:
just explain me why
```
latest = dataframe.loc[dataframe['date'] == latest_date].iloc[-1]
```
is now used instead of previous
```
latest = dataframe.iloc[-1]
```

xmatthias:
well the answer (at least partially) lies in https://github.com/freqtrade/freqtrade/pull/3107#discussion_r399802540.

using `dataframe.loc[dataframe['date'] == latest_date]` selects the row with the latest date (we gathered before running the strategy methods).

Since this returns a dataframe (potentially more than one row) - `.iloc[-1]` is then used to get the last row of the previously selected dataframe.


Thinking about this again makes it seem like this change is unnecessary / duplicate testing - `assert_df()` is already checking that the last row contains the last date - so using `.iloc[-1]` should be enough here.

hroff-1902:
> Since this returns a dataframe (potentially more than one row) - `.iloc[-1]` is then used to get the last row of the previously selected dataframe.

it's clear, okay

> using `dataframe.loc[dataframe['date'] == latest_date]` selects the row with the latest date (we gathered before running the strategy methods).

hmm, but we also check that the last date is the same in the returned dataframe... why (or when) this expression will behave more correct than previous one, why is it needed?

orehunt:
> why (or when) this expression will behave more correct than previous one, why is it needed?
probably never, however freqtrade doesn't use the date for the index, and I haven't check if the way the df is constructed ensures that the last index is always of the latest date; otoh matching a date compared to querying an index is not a cost free redundancy

hroff-1902:
LGTM 👍 

@orehunt , thanks!

Issue #3016, opened 2020-03-03
Freqtrade causing my VPS to crash


King-Techno:
Everytime I run freqtrade on my VPS (EC2 server) it works for an hour or two. Then it causes my VPS to crash and stop working. I honestly do not know what to do.

I am using Ubuntu 18.04 and my VPS has 1 CPU.

xmatthias:
well what do the logs show?
how much ram have you got?

King-Techno:
I dont know how to check the logs?
I installed freqtrade using docker.

VPS Data:
RAM: 1.0 GB, CPU: 1.

The issue started after I changed my strategy. I suspect that my CPU is too low for my current strategy and I need to upgrade my VPS.

King-Techno:
The issue occurred after I added this to my strategy `dataframe['high'].rolling(period).max().shift()`

xmatthias:
You should be able to check the logs by using journalctl (you'll need to investigate parameters to get messages from the previous boot - probably it's `journalctl -b -1` but i'm not certain at the moment.

Honestly, this is a no-brainer for me.

Minimal requirements are stated as 2GB Ram, 2vCPU - you're not filling either of these.

Also, requirements depend on the strategy - so if you're using a ton of indicators, 2GB may not be enough. If it's compute-intensive indicators, 2vCPU may not be enough....

King-Techno:
I agree as well that I will need to upgrade my server. I give that a shot and hopefully, it should work.

King-Techno:
I think the issue may be with freqtrade. I am having sudden spikes in my CPU usage causing me to use up all my credits. 

![image](https://user-images.githubusercontent.com/53098717/75784369-05903c80-5d5a-11ea-9f99-10bce707d909.png)

xmatthias:
From the timing perspective, i would assume you're using VolumePairlist - which is usually running every 6 hours.
However - a graph without specification as to what this is (memory / cpu / network) does not say anything.

King-Techno:
Sorry about that @xmatthias . I have made alterations to my server, so I will wait a few hours to see if there are any improvements. In the meantime I will close the case and reopen if issues persist. Thank you for your time though.

King-Techno:
## Step 1: Have you search for this issue before posting it?

Yes

If you have discovered a bug in the bot, please [search our issue tracker](https://github.com/freqtrade/freqtrade/issues?q=is%3Aissue). 
If it hasn't been reported, please create a new issue.

[...]

PR #3108, opened 2020-03-24
concurrency for multiple pairs

orehunt:
## Summary
Every pairs computes its signals in its own process

## Quick changelog
- signals from multiple pairs are evaluated with multiprocessing

## What's new?
Speeds up evaluation when using multiple pairs, for backtesting and live.
During live, signals are evaluated depending on the throttling configuration.
It currently spawns a pool of workers everytime signals are requested.
So it spawns `jobs` processes (currently set at cpus/2) every `process_throttle` seconds which might be a little bit too much if set to `1s`.
I used `multiprocessing.Pool` instead of `Parallel` because joblib doesn't have a proper way to terminate the pool, its `with` context doesn't seem to actually close the pool afterwards while `Pool` properly closes after all is done.
To be friendlier to shorter throttling freqtrade would have to run a long lived pool and reset it every once in a while, which is not _plug and play-ish_ as this implementation

It's important to reset the workers pool, because since strategies are user code, they can have memory leaks, so by running signals in their own processes it prevents from bloating the main freqtrade process. #2910 #3016 
It also can reduce memory pressure during backtesting for the same reason..

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -27,6 +27,7 @@ from freqtrade.rpc import RPCManager, RPCMessageType
 from freqtrade.state import State
 from freqtrade.strategy.interface import IStrategy, SellType
 from freqtrade.wallets import Wallets
+from freqtrade.strategy.helper import get_all_signals
 
 logger = logging.getLogger(__name__)
 
@@ -216,9 +217,15 @@ class FreqtradeBot:
                             "but checking to sell open trades.")
             else:
                 # Create entity and execute trade for each pair from whitelist
+                all_signals = get_all_signals(
+                    self.strategy.get_signal, {
+                        pair: (pair, self.strategy.ticker_interval,
+                               self.dataprovider.ohlcv(pair, self.strategy.ticker_interval))
+                        for pair in whitelist
+                    })
                 for pair in whitelist:
                     try:
-                        trades_created += self.create_trade(pair)
+                        trades_created += self.create_trade(pair, all_signals[pair])
                     except DependencyException as exception:
                         logger.warning('Unable to create trade for %s: %s', pair, exception)
 
@@ -379,7 +386,7 @@ class FreqtradeBot:
         # See also #2575 at github.
         return max(min_stake_amounts) / amount_reserve_percent
 
-    def create_trade(self, pair: str) -> bool:
+    def create_trade(self, pair: str, signals: Tuple[bool, bool] = None) -> bool:
         """
         Check the implemented trading strategy for buy signals.
 
@@ -401,9 +408,13 @@ class FreqtradeBot:
             return False
 
         # running get_signal on historical data fetched
-        (buy, sell) = self.strategy.get_signal(
-            pair, self.strategy.ticker_interval,
-            self.dataprovider.ohlcv(pair, self.strategy.ticker_interval))
+        if signals:
+            buy = signals[0]
+            sell = signals[1]
+        else:
+            (buy, sell) = self.strategy.get_signal(
+                pair, self.strategy.ticker_interval,
+                self.dataprovider.ohlcv(pair, self.strategy.ticker_interval))
 
         if buy and not sell:
             stake_amount = self.get_trade_stake_amount(pair)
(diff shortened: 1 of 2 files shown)


xmatthias:
First of all something organisational:
Please create issues to discuss changes before implementing them to avoid frustration and wasted time.
This applies to all huge changes in terms of work needed and / or huge behavioural changes.

Also, a PR should have tests pass - or should be labled as `[WIP]` - so we know it's still a work in progress.

----
about this PR: 

Using processes / threads has been discussed before and has been decided against, since it introduces a lot of complexity and potential error causes.

While it seems ok at first, it'll strongly depend on what a user does within his strategy.

* Assume a strategy keeps state of "something"  
  * how will you guarantee thread / process safety for this situation without having the user know how to handle threads / processes?
  * are changes to strategy attributes even passed back to the main process, so the next iteration will know about these?
* Assume a strategy uses dataprovider - it'll do "cross-process" calls - which i'm not sure do work - and if they do - how do you guarantee ccxt respects ratelimits configured? ccxt is not process-safe (nothing in python is).
* This also introduces wasted cpu cycles when the last trade-slot is filled wiht the first analyzed pair. In the current behaviour, only the first pair would be analyzed - in this behaviour all pairs will be analyed, always.

> It also can reduce memory pressure during backtesting for the same reason..

I'm not sure how analyzing in parallell should reduce memory usage during backtesting.
The high memory usage during backtesting exists because all pairs need to be analyzed and known while iterating over them  - otherwise you won't know if all trade-slots are filled or not once a trade-signal happens.
The alternative would be to feed backtesting the data in 500 candle units and rolling over time - but this would heavily reduce backtesting performance since the analysis operations would not be vectorized anymore.

---

I'm sorry since this seems like it's been quite some effort - but i'm not a fan of this and i think the drawdowns outweight the benefits - also i think i can break this code very easily within my user strategy by introducing race conditions (which will not fail immediately - but over time).

orehunt:
> First of all something organisational:
> Please create issues to discuss changes before implementing them to avoid frustration and wasted time.

this is something I am using myself, there is some effort to make it easily mergeable but not so much so no problem if pr aren't accepted, you can close them, no big deal
> 

> Also, a PR should have tests pass - or should be labled as `[WIP]` - so we know it's still a work in progress.

it did have all tests OK, maybe something slipped through


>     * Assume a strategy keeps state of "something"
>       
>       * how will you guarantee thread / process safety for this situation without having the user know how to handle threads / processes?
>       * are changes to strategy attributes even passed back to the main process, so the next iteration will know about these?
freqtrade could at best offer a pipe through maybe the dataprovider, to preserve state, but guaranteeing the persistence of state without specifications I don't see how that is desirable..use a database? I personally was just pickling on disk between calls...
>     * Assume a strategy uses dataprovider - it'll do "cross-process" calls - which i'm not sure do work - and if they do - how do you guarantee ccxt respects ratelimits configured? ccxt is not process-safe (nothing in python is).

honestly I haven't really used the dataprovider interface, if I get to it I'll see how it interacts, but something like a decorator that puts a semaphore around calls would be my first guess..
>     * This also introduces wasted cpu cycles when the last trade-slot is filled wiht the first analyzed pair. In the current behaviour, only the first pair would be analyzed - in this behaviour all pairs will be analyed, always.
This cannot be avoided if you want to sort pairs based on a score/sentiment, you have to calculate all the signals, so I just give this for granted

> I'm not sure how analyzing in parallell should reduce memory usage during backtesting.
> The high memory usage during backtesting exists because all pairs need to be analyzed and known while iterating over them - otherwise you won't know if all trade-slots are filled or not once a trade-signal happens.
mmh, you are assuming shared state during live, but not during backtesting? What if a strategy updates some state and that state grows the more candles are tested?

xmatthias:
> freqtrade could at best offer a pipe through maybe the dataprovider, to preserve state, but guaranteeing the persistence of state without specifications I don't see how that is desirable..use a database? I personally was just pickling on disk between calls...

wo says without specification? we have documentation on how to do this [here](https://www.freqtrade.io/en/latest/strategy-customization/#storing-information). cust-info is an example, but almost every other name is allowed, too.

this page also documents dataprovider usage (a little further down).

> honestly I haven't really used the dataprovider interface, if I get to it I'll see how it interacts, but something like a decorator that puts a semaphore around calls would be my first guess..

also a semaphore won't really help you respect the exchange's API ratelimits - as i assume that the whole exchange will be pickled / copied into each process.

It's also possible that using `Trade.*` might be problematic (will need to be tested) - as it might lead to concurrency / locking issues in certain cases.

orehunt:
> wo says without specification? we have documentation on how to do this [here](https://www.freqtrade.io/en/latest/strategy-customization/#storing-information). cust-info is an example, but almost every other name is allowed, too.
> 
I guess that count as specification..
> also a semaphore won't really help you respect the exchange's API ratelimits - as i assume that the whole exchange will be pickled / copied into each process.
No, you use a syncmanager

> It's also possible that using `Trade.*` might be problematic (will need to be tested) - as it might lead to concurrency / locking issues in certain cases.
Strategy code that modifies orders shouldn't have nothing to do with signals..

anyway this could easily be just a strategy feature flag, all the bot has to do is check how to analyze signals

PR #3110, opened 2020-03-24
[minor] Test warnings with filter always on

xmatthias:
## Summary
Fix random test failure by always enabling warnings - otherwise it'll sometimes be disabled depending on the environment.

### DIFF
--- a/tests/test_configuration.py
+++ b/tests/test_configuration.py
@@ -73,6 +73,7 @@ def test__args_to_config(caplog):
     configuration = Configuration(args)
     config = {}
     with warnings.catch_warnings(record=True) as w:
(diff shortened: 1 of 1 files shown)


hroff-1902:
LGTM, 

but can you pls explain your findings ("depending on the environment")?

xmatthias:
> depending on the environment

it happens on MY environment - but only when running the tests from the GUI - not when running the tests (in the same virtualenv) using plain pytest.

The command ran by the GUI does prepend environment variables to hide warnings (probably to reduce clutter) - so we need to add these again.

It depends "On the environment" as these variables can be set system-wide ...

Issue #3103, opened 2020-03-23
Fee calculation + stoploss support + trade amounts on FTX


yazeed:
So, while testing live trading on FTX using freqtrade. I've spotted a few FTX inconsistencies that I'll be logging here, so that we can start working on them in the future:

1. Fee calculation seems to fail all the time with this error (Leveraged tokens, haven't tried perps or options trading yet):

`2020-03-23 15:31:19,595 - freqtrade.freqtradebot - INFO - Applying fee on amount for Trade(id=225, pair=DRGNBEAR/USD, amount=0.01231875, open_rate=2789.00000000, open_since=2020-03-22 20:00:34) failed: myTrade-Dict empty found`

2. Stoploss on exchange is not supported yet on FTX, looking at ftx.py it's obvious that it's lacking the param config that enables it as well as the needed stoploss() and stoploss_adjust() methods.

3. Freqtrade keeps getting out of sync on orders (amount) on FTX, occasionally I get these occurrences which I have to fix manually from the DB:

`2020-03-23 11:10:48,474 - freqtrade.freqtradebot - WARNING - Unable to sell trade: Not enough amount to sell. Trade-amount: 0.0091, Wallet: 0.0082`

xmatthias:
First of all - please use one issue per discovered problem. Otherwise discussions get confusing and we're stuck with issues we cannot ever close because one minor thing was never done, but the issue would be 90% done - but noone will be able to figure out which part is still missing.

1) that's not relevant for FTX i think. as far as i know, FTX does subtract fees from the "correct" side of the pair (not the destination pair, but the source pair) - so fees should apply correctly. This message mainly applies to binance, which subtracts fees from the Destination currency (or BNB - but then it's not relevant again).

2) that's not even really an issue - just something which is "to implement and test" ...but as mentioned slack - it's really just a "noone invested the time into this yet" more than anything.

3) Without a full log for this, it'll be impossible to find the solution to this (also - please use a separate issue for this.

Dante2333:
On the fee calculation, since FTX can have multiple fee tiers which actually rely on volume .. or different discount; it would be ideal if we could query over api to get current account fees then use them from that moment in time for freqtrade?

https://help.ftx.com/hc/en-us/articles/360024479432-Fees

For 2, give me exactly what you would need to I can pass it to FTX devs and see if we could have stop-loss on exchange; 

[...]

Issue #2646, opened 2019-12-11
cancel an order before the unfilledtimeout expires


maxma-bit:
Is there a way to cancel an order before the unfilledtimeout expires, in a way that the local sqlite database remains consistent with the exchage?
Please note that the RPC command /forcesell or /forcebuy do not help with this. In the case of a SELL, one would like to cancel the SELL order in Order Book before the sale is actually executed. It seems a bit different from forcing a SELL with the RPC command /forcesell.
For BUY order, one would like to cancel the order before the BUY purchase is made. It seems to me different from forcing a purchase with /forcebuy.
if there is no way to do it, could you consider this for a possible extension?

xmatthias:
I guess you should be able to cancel the buy-order "on exchange" - the database should notice that and update the trade.
One problem this causes:
assuming you don't use `process_only_new_candles=True`, then the buy-signal may still be active - so it reopens the trade seconds after you cancelled that.
We should be able to prevent that by locking the pair anyway - but that'll mean that you the bot can't buy this pair for the duration of this candle.


i think you may be able to stop a buy via `/forcesell` (don't pin me on that though - never tested this scenario).

hroff-1902:
> assuming you don't use process_only_new_candles=True, then the buy-signal may still be active - so it reopens the trade seconds after you cancelled that.

Should not the pair be locked after first buy?

Pair of 2 commands `/stopbuy=pair`, `/cancelorder=id` issued should help in this case anyway...

werkkrew:
Is this still considered a valuable "good first issue"?

xmatthias:
sure thing - i think all issues we've got currently open (with the label Enhancement) are kindof "free to pick" or are otherwise closed, although some will not be easy to implement / require multiple steps to fully be implemented  (this does not apply to "good first issue" though).

PR #3112, opened 2020-03-25
Trade state updates

xmatthias:
## Summary
Revise/revisit handling of orders which are closed on the exchange.

When using limit orders, an order can remain on the exchange until it's timed out.
However, users can also manually go to the exchange and cancel an order - which did act differently depending at what point in the bot-cycle the order was cancelled.

This acts also as a preperation for #2646 - and may eventually solve part 3 of #3103 (unverified, as i'm not 100% clear on the problem there).

## Quick changelog

This PR 
* Combines methods for handling trade-updates to fees
* Moves the checking for closed orders into the timeout section (so cancelled orders are handled properly)
* moves checking for timedout orders to the start of the cycle 
* adjusts handling of cancelled orders (FTX returns the orderstatus as closed, with filled == 0.0).
* uses filled instead of amount if possible (amount is usually the requested order amount - not the filled one)

I've livetested this against both binance and FTX (dry and live)- and both handling of cancelled orders, as well as filled orders works as expected.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -602,7 +602,6 @@ class FreqtradeBot:
         trades_closed = 0
         for trade in trades:
             try:
-                self.update_trade_state(trade)
 
                 if (self.strategy.order_types.get('stoploss_on_exchange') and
                         self.handle_stoploss_on_exchange(trade)):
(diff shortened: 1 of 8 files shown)


freqtrade/freqtradebot.py
```diff
@@ -916,17 +913,7 @@ def handle_timedout_limit_buy(self, trade: Trade, order: Dict) -> bool:
...
-            new_amount = self.get_real_amount(trade, corder if 'fee' in corder else order,
-                                              trade.amount)
-            if not isclose(order['amount'], new_amount, abs_tol=constants.MATH_CLOSE_PREC):
-                trade.amount = new_amount
-                # Fee was applied, so set to 0
-                trade.fee_open = 0
-                trade.recalc_open_trade_price()
-        except DependencyException as e:
-            logger.warning("Could not update trade amount: %s", e)
+        self.update_trade_state(trade, corder if 'fee' in corder else order, trade.amount)
```
hroff-1902, line 916: Hmm, I'm thinking of this expression:
```
corder if 'fee' in corder else order
```

I see, it's from the original code, but do you remember/know why is it in this (quite) tricky way?

tests/exchange/test_exchange.py
```diff
@@ -1705,6 +1705,18 @@ def test_cancel_order_dry_run(default_conf, mocker, exchange_name):
     assert exchange.cancel_order(order_id='123', pair='TKN/BTC') is None
 
 
+@pytest.mark.parametrize("exchange_name", EXCHANGES)
+@pytest.mark.parametrize("order,result", [
+    ({'status': 'closed', 'filled': 10}, False),
+    ({'status': 'closed', 'filled': 0.0}, True),
+    ({'status': 'canceled', 'filled': 0.0}, True),
+    ({'status': 'canceled', 'filled': 10.0}, False),
```
hroff-1902, line 1713: a testcase for other status values (other than 'closed' and 'canceled') should be added here

freqtrade/freqtradebot.py
xmatthias: we're trying to use the fees from the latest possible point - which ideally is from the `cancel_order()` call - available as `corder`.

The result of cancel_order is not really consistent between exchanges - some return the same as `fetch_order()` (but with the state right after the order was cancelled) - while others return a simple string in the form of "success" (FTX - i'm looking at you).

For that reason, we must use the result from `fetch_order()` (available as order) - verify it's a dict (done a few lines above) - and then always use corder (or fallback to order) for everything we use from corder.

#3146 will enhance drive this logic further - but keeping these 2 PR's seperate is better as the 2 address different problems (this one code duplication and potential timing hickups  - the other one inconsistent content in the order and cancel_order results).

hroff-1902: > The result of cancel_order is not really consistent between exchanges

Yeah, I remember this fact and that fix you made...

> and then always use corder (or fallback to order) for everything we use from corder

yes, this (as written here) is clean, I wonder the manner the check is performed:
```
corder if 'fee' in corder else order
```
-- why 'fee' is used here, why not some other element?

xmatthias: well because the most releant place is the call to get_real_amount ... which does heavily rely on fee to be available to work correctly.

Also, fee is the element that's the "least secure" - it's missing in most exchanges ... (but not in all)

freqtrade/freqtradebot.py
```diff
@@ -1117,7 +1118,7 @@ def _notify_sell_cancel(self, trade: Trade, order_type: str, reason: str) -> Non
 #
 
     def update_trade_state(self, trade: Trade, action_order: dict = None,
-                           order_amount: float = None) -> None:
+                           order_amount: float = None) -> bool:
         """
         Checks trades with open orders and updates the amount if necessary
         Handles closing both buy and sell orders.
```
hroff-1902, line 1118: pls add the description for the return value here -- when it's supposed to return True and False

PR #3127, opened 2020-03-31
use equality instead of index for row lookups

orehunt:
without the equality operator which losely checks within a range, the lookup will throw an index error when plotting against the trades on a database, because trades in a database have real dates, compared to backtesting that are fixed to the precision of the timeframe

### DIFF
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -10,6 +10,7 @@ from freqtrade.data.btanalysis import (calculate_max_drawdown,
                                        create_cum_profit,
                                        extract_trades_of_period, load_trades)
 from freqtrade.data.converter import trim_dataframe
+from freqtrade.exchange import timeframe_to_prev_date
 from freqtrade.data.history import load_data
 from freqtrade.misc import pair_to_filename
 from freqtrade.resolvers import StrategyResolver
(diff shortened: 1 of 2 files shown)


xmatthias:
sorry this took a while... but i had to find time to build a case i could test this with (not your fault - i didn't have a database with corresponding data on my pc)

 
I can see the problem now - but i'm not convinced it's the best solution.
While it does fix the crash - it'll simply hide / not show max_drawdown ... 

I would rather use `timeframe_to_prev_date()` from the exchange class to "move" the date to the previous candle (or `timeframe_to_next_date()`) to move it to the next candle instead.

orehunt:
I see, there is still a case where you might not have updated ohlcv data compared to the database trades, and it will again miss the index, although that would probably be better to check even before the max drawdown calc

xmatthias:
hm, ah you mean when no ohlcv data is available for plot-profit? 

but that's not really relevant at this point (unless i miss something)? 
the graph we're plotting on is the 2nd graph (combined profit) ... it has only entries for trades close_time anyway - and we can only have drawdowns for this period, too ... no?

orehunt:
so `df_comb` starts by merging all the pairs data, then adds a `mean` and `cum_profit` column with `loc` which means trades past the ohlcv won't be present as it doesn't concatenate dates, that's the df passed to the max drawdown function

xmatthias:
ah - good call - 
in that case i guess plot_profit should run through `extract_trades_of_period()` too ... to trim trades down to all trades within the available data period.

freqtrade/plot/plotting.py
```diff
@@ -383,6 +385,9 @@ def generate_profit_graph(pairs: str, data: Dict[str, pd.DataFrame],
     # Combine close-values for all pairs, rename columns to "pair"
     df_comb = combine_dataframes_with_mean(data, "close")
 
+    # Trim trades to available OHLCV data
+    trades = extract_trades_of_period(df_comb, trades, date_index=True)
```
xmatthias, line 389: Great solution :+1:

PR #3129, opened 2020-04-01
convert dl-trades datadownload to list

xmatthias:
## Summary
As we've seen recently on slack, downloading trades has huge memory requirements when using a slightly longer trades period.

The main problem is the dataformat. We used to store trades-data as ccxt returns it - which resulted in a dictionary with a total size of 2216b per trade.

This PR converts the format right after downloading to a list with only the necessary columns - bringing the "average" size down to ~320b.
Now this drops "quite a bit"  of information, but keeps everything necessary to convert trades to candles, as well as keeping the ability to continue downloading.

The formats, old and new can be seen in the fixtures `fetch_trades_result` (old) and `trades_history` (new).

This alone reduces the (theoretical) memory requirements from 9G to 1.3G (for 4M trades).

Furthermore, the "since" date was not properly adjusted, leading to always download --days (or 28 days) worth of data, even if the command was ran every day.

This does not affect ohclv data other than the volume - but please be aware of this.

## Quick changelog

- Remove duplicate trades
- Change trade format from dict to list to reduce memory size
- Properly adjust "since" to avoid redownloading the same period again and again
- Automatically handle conversion to new trades-dataformat

To force the conversion best use  the following command before the next data download after upgrading to this version.

```
freqtrade convert-trade-data --format-from jsongz --format-to jsongz --datadir user_data/data/kraken

```

## Further enhancements
Further enhancements are possible - but not included in this PR:
- Batch download
- appending to files (not supported with json format, json always needs to be rewitten completely)


freqtrade/data/converter.py
```diff
@@ -154,7 +157,27 @@ def order_book_to_dataframe(bids: list, asks: list) -> DataFrame:
...
+    return [i for i, _ in itertools.groupby(sorted(trades, key=itemgetter(0)))]
+
+
+def trades_dict_to_list(trades: List[Dict]) -> List[List]:
+    """
+    Convert fetch_trades result into a List (to be more memory efficient).
+    :param trades: List of trades, as returned by ccxt.fetch_trades.
+    :return: List of Lists, with constants.DEFAULT_TRADES_COLUMNS as columns
+    """
+    return [[t[col] for col in DEFAULT_TRADES_COLUMNS]for t in trades]
```
hroff-1902, line 177:
```suggestion
    return [[t[col] for col in DEFAULT_TRADES_COLUMNS] for t in trades]
```

freqtrade/data/history/idatahandler.py
```diff
@@ -18,6 +18,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Type for trades list
+TradeList = List[List]
```
hroff-1902, line 23: yes! 👍 
we should use type aliases for our dicts and lists more intensively

hroff-1902:
just a question: tuples are more memory efficient than lists, should not the original list of dicts be converted into the list of tuples instead of lists of lists?

xmatthias:
> just a question: tuples are more memory efficient than lists, should not the original list of dicts be converted into the list of tuples instead of lists of lists?

Yes, in theory - however then we'd have both a mix of lists and tuples for the same thing i tried to avoid.

Json does not know the concept of Tuples - so upon saving a list of tuples, you get a list of lists in the file - and consequentially, everything loaded from file will be a List of Lists.
We could convert this to a list of Tuples - but that's an extra step - and i would like to avoid to have a `List[Lists | Tuples]` type ...

hroff-1902:
> Json does not know the concept of Tuples - so upon saving a list of tuples, you get a list of lists in the file - and consequentially, everything loaded from file will be a List of Lists.

Well, no if implement it using a custom subclass of JSONDecoder/JSONEncoder, which converts the type "on the fly"... (in theory... but in practice I guess I agree with you)

xmatthias:
well - but still the "on the fly" converter will need to change all lines ... 

Also, we're talking about Json now - what about future types? implement a "custom" parser for csv, hdf5, ... too? that'll make implementing new datatypes almost impossible.

to summarize - i did think about using Tuple (even used it initially) - but decided against it :)

xmatthias:
best mention that the format changed to reduce the memory (and storage) footprint of trades.
---- 

The freqtrade internal storage format for trades data has been changed.
The switch to the new trade-format is automatic once download-data is ran for the first time - however switching back to an earlier version is impossible and will result in a complete redownload of all trades data.
A manual conversation can be done via the following command: `freqtrade convert-trade-data --format-from jsongz --format-to jsongz --datadir user_data/data/kraken`.

Switching back to the old trades format is NOT possible.

PR #3133, opened 2020-04-02
[minor] Fix filename handling with --strategy-list

xmatthias:
## Summary
Filenames for exported fiels should not be increasing in length - but should be `filename-<strategy>.json` for all strategies, independently if it's the first or last strategy.

As reported on slack.

### DIFF
--- a/tests/optimize/test_optimize_reports.py
+++ b/tests/optimize/test_optimize_reports.py
@@ -160,10 +160,15 @@ def test_backtest_record(default_conf, fee, mocker):
     # reset test to test with strategy name
     names = []
     records = []
-    results['Strat'] = pd.DataFrame()
+    results['Strat'] = results['DefStrat']
+    results['Strat2'] = results['DefStrat']
     store_backtest_result(Path("backtest-result.json"), results)
     # Assert file_dump_json was only called once
-    assert names == [Path('backtest-result-DefStrat.json')]
+    assert names == [
+        Path('backtest-result-DefStrat.json'),
+        Path('backtest-result-Strat.json'),
+        Path('backtest-result-Strat2.json'),
+    ]
     records = records[0]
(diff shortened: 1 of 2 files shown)


tests/optimize/test_optimize_reports.py
```diff
@@ -160,10 +160,15 @@ def test_backtest_record(default_conf, fee, mocker):
     # reset test to test with strategy name
     names = []
     records = []
-    results['Strat'] = pd.DataFrame()
+    results['Strat'] = results['DefStrat']
+    results['Strat2'] = results['DefStrat']
     store_backtest_result(Path("backtest-result.json"), results)
     # Assert file_dump_json was only called once
```
hroff-1902, line 166: this comment is no longer relevant?

hroff-1902:
> should not be increasing in length

how was it before?

xmatthias:
Assume i use the strategies DefStrat, Strat and Strat2 in the stratgy list (in this sequence)

the filenames would be returned as an expanding list - appending the current strategy to the end of the filenames:

`backtest-result-DefStrat.json` - for the first strategy
`backtest-result-DefStrat-Strat.json` for the 2nd strategy
`backtest-result-DefStrat-Strat-Strat2.json` for the 3rd strategy.

hroff-1902:
> Assume i use the strategies DefStrat, Strat and Strat2 in the stratgy list (in this sequence)
> 
> the filenames would be returned as an expanding list - appending the current strategy to the end of the filenames:
> 
> `backtest-result-DefStrat.json` - for the first strategy
> `backtest-result-DefStrat-Strat.json` for the 2nd strategy
> `backtest-result-DefStrat-Strat-Strat2.json` for the 3rd strategy.

Ah, got it, it was definitely wrong.

PR #3137, opened 2020-04-05
[minor] Fix maxdrawdown

xmatthias:
## Summary
Max drawdown must always have high first, then low.
This was incorrectly done in our previous implementation, resulting in plots as the following:

2020-04-05-143050_1653x263_scrot

Correct version would be: 

2020-04-05-143103_1656x266_scrot

## Quick changelog

- add test showing this is broken
- fix broken max-drawdown implementation


hroff-1902:
why do you need to reset the index here?

xmatthias:
well the index reset is needed to avoid the above screenshot - since we're using `.iloc[]` and .idxmax()/.idxmin().

Also - the original dataframe is NOT modified (it's not using `inplace=True`) - I've now also added an assert statement to ensure this remains this way.

hroff-1902:
> Also - the original dataframe is NOT modified (it's not using `inplace=True`) - but the result is assigned to a new variable (intentionally...).

Ah, yeah, right
👍

PR #3138, opened 2020-04-05
trades history RPC

orkblutt:
## Summary
Add trades history RPC


freqtrade/rpc/api_server.py
```diff
@@ -358,6 +359,19 @@ def _balance(self):
...
+    @require_login
+    @rpc_catch_errors
+    def _trades(self):
+        """
+        Handler for /trades.
+
+        Returns the X last trades in json format
+        """
+        last_trades_number = request.args.get('last_trades_number', 0)
+        last_trades_number = int(last_trades_number)
```
xmatthias, line 371: You should be able to combine these 2 lines - it's not really necessary to have these in 2 lines i think.

```suggestion
        last_trades_number = int(request.args.get('last_trades_number', 0))
```

freqtrade/rpc/rpc.py
```diff
@@ -226,6 +226,53 @@ def _rpc_daily_profit(
...
+        """ Returns the X last trades """
+        if last_trades_number > 0:
+            trades = Trade.get_trades().order_by(Trade.id.desc()).limit(last_trades_number)
+        else:
+            trades = Trade.get_trades().order_by(Trade.id.desc()).all()
+
+        output = []
+
+        for trade in trades:
+            output.append({
```
xmatthias, line 240: I would rely on `trade.to_json()` for this to have one common "trade to json conversion" method.

It implements part of what's needed here ... so you may need to extend that function to return "everything", but that shouldn't be a problem for anything else (more is always better - no)?

freqtrade/rpc/rpc.py
```diff
@@ -226,6 +226,53 @@ def _rpc_daily_profit(
             for key, value in profit_days.items()
         ]
 
+    def _rpc_trade_history(
+            self, last_trades_number: int) -> Dict[str, List[Dict[str, Any]]]:
+        """ Returns the X last trades """
+        if last_trades_number > 0:
```
xmatthias, line 232: i somehow don't like the wording of `last_trades_number` ... but i currently struggle to really find a better name - maybe something as simple as `limit` ?

scripts/rest_client.py
```diff
@@ -156,6 +156,13 @@ def show_config(self):
         """
         return self._get("show_config")
 
+    def history(self, number=None):
+        """Return the amount of open trades.
```
xmatthias, line 160: best change this line to `Return trades history."

also, i'd appreciate if you could add a line below (right before :return: - with `:param limit: Limit trades to latest X trades.`

freqtrade/rpc/rpc.py
orkblutt: yes limit is good.

freqtrade/rpc/rpc.py
orkblutt: Yes better to have an unified trade view

freqtrade/rpc/api_server.py
orkblutt: Yeah it's useless

scripts/rest_client.py
orkblutt: bad copy/paste :)
Done.

scripts/rest_client.py
```diff
@@ -156,6 +156,14 @@ def show_config(self):
         """
         return self._get("show_config")
 
+    def history(self, limit=None):
```
xmatthias: This name seems quite inconsistent (sorry for not noticing before)...

How about we rename it to `trades` ? 
All other methods are named according to their API endpoint (`/trades` ...) - and i'd keep it at that.

PR #3148, opened 2020-04-10
Add complete list of keyboard, as in the help.

phe-sto:
Some useful commands like /reload_conf were in the help of the telegram bot but not in the the keyboard.
I just updated the telegram keyboard, with the entire list of commands.

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -19,7 +19,6 @@ logger = logging.getLogger(__name__)
 
 logger.debug('Included module rpc.telegram ...')
 
-
 MAX_TELEGRAM_MESSAGE_LENGTH = 4096
 
 
@@ -29,6 +28,7 @@ def authorized_only(command_handler: Callable[..., None]) -> Callable[..., Any]:
     :param command_handler: Telegram CommandHandler
     :return: decorated function
     """
+
     def wrapper(self, *args, **kwargs):
         """ Decorator logic """
         update = kwargs.get('update') or args[0]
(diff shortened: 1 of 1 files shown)


Fredrik81:
Did you test this on mobile device?

xmatthias:
First of all, hi and welcome to the freqtrade repository.
Thanks for your contribution in improving our bot!

----

To be honest, it's intentional to not have all commands in the "keyboard" to retain usability.

The way it looks now presents a non-ideal UX (user experience) in my opinion.
If you look at the screenshots below, it'll now have a scrollbar to get to all commands (on both web and phone client).
It's also quite dangerous to accidentally hit the wrong button, since it requires scrolling - so the same button will not be in the same place all the time - which is bad for muscle memory...


Now i don't disagree with adding "some" more buttons (or replace some rarely used buttons).

However, there's a few buttons  which only work in certain circumstances (`/edge`, `/forcebuy`).
In addition to the above buttons, all buttons that require a 2nd argument (`/forcebuy PAIR` - `/forcesell <id|all>` cannot be used this way - so having the button does not make much sense as it would always result in an error being returned.
Also, the usage of `/version` seems very limited - and i would keep the keyboard (which is showing you all commands!) .

--- 

In total, i would try to limit this to a maximum of 4 rows (more rows will require scrolling) - and have always the same number of buttons per row (i am aware that we don't follow that now).

What we could do instead is some kind of `/more` command - which opens an an [inline keyboard](https://core.telegram.org/bots#inline-keyboards-and-on-the-fly-updating) - and allows to quickly select additional, not often required options.

Unfortunately, [command autocomplete](https://core.telegram.org/bots#commands) seem to require user interaction with the botfather - so this is not an option for the bot to push - which is very unfortunate as this would be a great option.

Hope the above thoughts do make sense to you too...

## How it looks now:

on the web:

2020-04-10-162911_468x246_scrot


On a phone (>6''):
photo5801015725035467940

phe-sto:
Hi guys,

I tested on my Iphone SE and it is a bit croudy indeed. I get it was intentional. What about just adding `./reload_conf` ? it is quite useful (at list to me) and has no argument.

best regards,

xmatthias:
and just `/reload_conf` is one i'm a bit concerned about, as it'll restart the bot - pressed unintentionally this can have ... non-desired consequences... if this were to happen in the wrong moment (just before closing a winning trade ...) - as a startup can take a half a minute or so ... (depending on the exchange / pairs in the pairlist etc.).

phe-sto:
I get it, you are right. I am closing the PR.

PR #3207, opened 2020-04-24
Minor: Remove unused methods in hyperopt

hroff-1902:
No longer used after the tabular format was introduced

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -294,21 +294,6 @@ class Hyperopt:
         return output
 
     @staticmethod
-    def print_results_explanation(results, total_epochs, highlight_best: bool,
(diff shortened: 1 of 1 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -293,21 +293,6 @@ def get_results(self, results) -> str:
             self.hyperopt_table_header = 2
         return output
 
-    @staticmethod
-    def print_results_explanation(results, total_epochs, highlight_best: bool,
```
xmatthias, line 297: You might want to remove `get_results()` too? 

It doesn't seem to be used either...

hroff-1902: Sure, nice catch

PR #3210, opened 2020-04-24
Cancel open orders during shutdown

### LINKED ISSUE #907: shutdown should cancel open orders
## Proposal

When stopping the bot (via ctrl+c, `/stop`) open orders are kept open and will be filled by the exchange "whenever" the market is ready to.
This can be dangerous should the bot be stopped due to a sudden market crash

A proper shutdown should allow the following (configurable!):

- [ ] cancel unfilled open orders
- [ ] force-close all open orders

Comments, suggestions, Pullrequests in this regard are welcome :wink:

jpribyl:
✅ Tests passing
✅  PEP8 conformant
✅  Mypy passing

## Summary
closes #907 

When stopping the bot (via ctrl+c, `/stop`) open orders are cancelled.

## Quick changelog

- Add `process_stopped` method to `freqtradebot`
- Call `freqtrade.process_stopped` from `Worker`
- Generalize code to cancel open orders
- Add `cancel_all_open_orders` method to freqtradebot, call during cleanup or `process_stopped`

## What's new?
This does not add a config option. It will cancel open orders whenever the bot is stopped.


xmatthias:
A few comments for now ...
Overall it looks pretty good though :+1:

freqtrade/freqtradebot.py
```diff
@@ -161,6 +162,13 @@ def process(self) -> None:
 
         Trade.session.flush()
 
+    def process_stopped(self) -> None:
+        """
+        Close all trades that were left open
+        """
+        if Trade.get_open_order_trades():
```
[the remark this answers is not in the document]
jpribyl: Great point! In my head I was avoiding the db query when there were no open trades but you're totally correct.

jpribyl:
@xmatthias thanks for the feedback! I should have some time to make those changes this evening 🥳

freqtrade/commands/cli_options.py
```diff
@@ -109,6 +109,12 @@ def __init__(self, *args, **kwargs):
         help='Enforce dry-run for trading (removes Exchange secrets and simulates trades).',
         action='store_true',
     ),
+    "cancel_on_exit": Arg(
+        '--cancel-on-exit',
+        help='Specify whether to close unfilled open orders when the bot stops / exits',
+        choices=constants.AVAILABLE_CANCELLATIONS,
+        default='none',
```
jpribyl, line 116: keep existing behavior by default

freqtrade/freqtradebot.py
```diff
@@ -33,6 +34,17 @@
 logger = logging.getLogger(__name__)
 
 
+class CancelReason(Enum):
```
jpribyl, line 37: There may be a better place to put this? We use these strings in tests as well, so this will prevent any typos 👍

freqtrade/freqtradebot.py
```diff
@@ -914,40 +952,42 @@ def handle_timedout_limit_buy(self, trade: Trade, order: Dict) -> bool:
...
-        return False
+            trade.open_order_id = None
+            logger.info('Partial buy order timeout for %s.', trade)
+            self.rpc.send_msg({
+                'type': RPCMessageType.STATUS_NOTIFICATION,
+                'status': f'Remaining buy order for {trade.pair} cancelled due to timeout'
+            })
+
+        self.wallets.update()
+        self._notify_buy_cancel(trade, order_type=self.strategy.order_types['buy'])
```
jpribyl, line 975: I don't think we need to define a variable for `order_type`? Also I didn't see any reason that this needs to be in its own method unless there is a case where we might want to cancel an order but not update/notify?

tests/test_arguments.py
```diff
@@ -64,6 +64,11 @@ def test_parse_args_db_url() -> None:
     assert args["db_url"] == 'sqlite:///test.sqlite'
 
 
+def test_parse_args_cancel_on_exit() -> None:
```
jpribyl, line 67: We may need to write some more tests?

config_binance.json.example
```diff
@@ -6,6 +6,7 @@
     "fiat_display_currency": "USD",
     "ticker_interval": "5m",
     "dry_run": true,
+    "cancel_on_exit": "none",
```
xmatthias, line 9: You may want to add this to `base_config.json.j2` as well...

freqtrade/constants.py
```diff
@@ -21,6 +21,7 @@
 AVAILABLE_PAIRLISTS = ['StaticPairList', 'VolumePairList',
                        'PrecisionFilter', 'PriceFilter', 'SpreadFilter']
 AVAILABLE_DATAHANDLERS = ['json', 'jsongz']
+AVAILABLE_CANCELLATIONS = ['unfilled_orders', 'none']
```
xmatthias, line 24: Any particular reason to use a enum here instead of a true/false flag?

i think that would simplify configuration from a user perspective... 

```json
{
  "cancel_open_orders_on_exit": true
}
```

I guess it's to keep further extensibility - but what else could we cancel eventually in the future?  (this is a serious question - please consider and answer before changing this, as the truthyness of my whole comment here depends on the answer ...)

freqtrade/freqtradebot.py
xmatthias: `constants.py`? 

I'm also not sure this should be a enum - since we're only interrested in the string portion (`.value`) of it, a dict may be better suited.

freqtrade/freqtradebot.py
```diff
@@ -957,11 +997,17 @@ def handle_timedout_limit_sell(self, trade: Trade, order: Dict) -> str:
             trade.close_date = None
             trade.is_open = True
             trade.open_order_id = None
+        else:
+            # TODO: figure out how to handle partially complete sell orders
+            reason = CancelReason.PARTIALLY_FILLED.value
```
xmatthias, line 1002: Note to self: we need to handle this differently and/or prevent notification spamming (probably as a first step)

If an order get's stuck here it'll send "not cancelled" over and over again (every iteration ...).

Leave as is for now - it's not making it worse - and this is complex enough :)

freqtrade/freqtradebot.py
```diff
@@ -888,24 +906,44 @@ def check_handle_timedout(self) -> None:
...
         """
-        Buy timeout - cancel order
+        Cancel all orders that are currently open
+        :return: None
+        """
+
+        for trade in Trade.get_open_order_trades():
+            try:
+                if not trade.open_order_id:
+                    continue
```
xmatthias, line 920: I don't think this is necessary.
`Trade.get_open_order_trades()` does the query via `Trade.open_order_id.isnot(None)` - so i don't think it's possible to hit this "continue" statement.

freqtrade/freqtradebot.py
```diff
@@ -888,24 +906,44 @@ def check_handle_timedout(self) -> None:
...
+                order = self.exchange.get_order(trade.open_order_id, trade.pair)
+            except (RequestException, DependencyException, InvalidOrderException):
```
xmatthias, line 922: You should be able to remove `RequestException` from this.

RequestException is not raised here as ccxt handles this and returns a CCXT internal exception (which is converted to a freqtrade exception in `get_order()`).

tests/test_arguments.py
xmatthias: well - yes ...

freqtrade/freqtradebot.py
xmatthias: I'm not sure i understand this comment (but then Github does render this diff VERY oddly) ... 

assuming it's about the below section - then no, it's definitely not needed.

``` python
                order_type = self.strategy.order_types['buy']	
                self._notify_buy_cancel(trade, order_type)
```

freqtrade/constants.py
jpribyl: Yeah, so I did it like this because @hroff-1902 had suggested that we may want to have three options: `cancel_orders | closing_positions | keep_orders_and_positions` - I'm not sure if there's a real need for closing positions though? If I am understanding correctly the functionality could be achieved with a `/stopbuy` and `/forcesell` ? I can see the argument for a boolean here!

xmatthias: Yeah exactly.

Also - closing positions is a VERY dangeous approach in my opinion - and will most likely prevent people from updating a bot to the latest version as it'll do a "hard stop" on all open orders.

jpribyl: With that in mind, I have updated it to be a boolean, `cancel_open_orders_on_exit`

freqtrade/worker.py
```diff
@@ -131,8 +131,7 @@ def _throttle(self, func: Callable[..., Any], throttle_secs: float, *args, **kwa
         return result
 
     def _process_stopped(self) -> None:
-        # Maybe do here something in the future...
-        pass
+        self.freqtrade.process_stopped()
```
hroff-1902, line 134: This will call `cancel_all_open_orders()`, I do not think it's the intended behavior...
This can be fixed/improved either by a flag here or in the Freqtradebot class, or better by introducing the new STOPPING state, which will transit the state from RUNNING to STOPPED. 
(@xmatthias not in this PR, what do you think?)

hroff-1902:
I'm in doubt, if we really need this new cli option, the config setting could be enough...
(But, from the other hand, you can with to (re-)start the bot temporary with this option just in order to cancel all orders...)

freqtrade/worker.py
xmatthias: that's a good point - though i'm not sure when / if this would cause problems - once the orders are closed, the loop won't return anything.

On the other hand, assuming i start the bot in stopped state (this sounds contradicting, but can be done by setting `initial_state` to "stopped" in the config) - it would start, sell all trades without ever starting (which may or may not be what i want).

xmatthias:
> I'm in doubt, if we really need this new cli option, the config setting could be enough...
> (But, from the other hand, you can with to (re-)start the bot temporary with this option just in order to cancel all orders...)

You think it can be confusing if we have too many CLI arguments? that's a good point, i never really thought about that .. but we already got a TON of cli arguments - most of which i wouldn't know without looking at the help ...

hroff-1902:
> > I'm in doubt, if we really need this new cli option, the config setting could be enough...
> > (But, from the other hand, you can with to (re-)start the bot temporary with this option just in order to cancel all orders...)
> 
> You think it can be confusing if we have too many CLI arguments? that's a good point, i never really thought about that .. but we already got a TON of cli arguments - most of which i wouldn't know without looking at the help ...
> 
> (to be clear, this is about the CLI option, not about the option in the configuration).

I see the next good improvement here can be if we make it more selectable by splitting into `cancel_open_orders_on_stop` (which cancels orders when the bot is transitioning into the STOPPED state, where it then can be resumed from) and `cancel_open_orders_on_exit` (which cancels when Ctrl+C is pressed or the process dies inexpectedly). Shall we add new cli option for this then?

Next improvement can be similar options for closing positions, by emergency selling them to zeros... Also split for `on_stop` and `on_exit` cases... I mean, there can be many similar cli options if we add this one now...

Besides the doubts above (too many options with further improvements), I'm not against this one, but I'd like to see particular cases where this cli option is much better than setting in the config (i.e. when you need to run the bot temporarily with this option instead of the config setting set)...

freqtrade/worker.py
hroff-1902: Okay, not in this PR (I'll introduce the STOPPING state later)

docs/configuration.md
```diff
@@ -51,6 +51,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `fiat_display_currency` | Fiat currency used to show your profits. [More information below](#what-values-can-be-used-for-fiat_display_currency). <br> **Datatype:** String
 | `dry_run` | **Required.** Define if the bot must be in Dry Run or production mode. <br>*Defaults to `true`.* <br> **Datatype:** Boolean
 | `dry_run_wallet` | Define the starting amount in stake currency for the simulated wallet used by the bot running in the Dry Run mode.<br>*Defaults to `1000`.* <br> **Datatype:** Float
+| `cancel_open_orders_on_exit` | Cancel orders when `/stop` is issued or `ctrl+c` is pressed. This will allow you to use `/stop` to cancel unfilled orders in the event of a market crash. This will not impact open positions. <br>*Defaults to `False`.* <br> **Datatype:** Boolean
```
hroff-1902, line 54:
```suggestion
| `cancel_open_orders_on_exit` | Cancel open orders when the `/stop` RPC command is issued, `Ctrl+C` is pressed or the bot dies unexpectedly. When set to `true`, this allows you to use `/stop` to cancel unfilled and partially filled orders in the event of a market crash. It does not impact open positions. <br>*Defaults to `false`.* <br> **Datatype:** Boolean
```

freqtrade/commands/cli_options.py
```diff
@@ -109,6 +109,11 @@ def __init__(self, *args, **kwargs):
         help='Enforce dry-run for trading (removes Exchange secrets and simulates trades).',
         action='store_true',
     ),
+    "cancel_open_orders_on_exit": Arg(
+        '--cancel-open-orders-on-exit',
+        help='Close unfilled open orders when the bot stops / exits',
```
hroff-1902, line 114:
```suggestion
        help='Close unfilled and partially filled open orders when the bot stops / exits.',
```

freqtrade/constants.py
```diff
@@ -318,3 +319,10 @@
     'dataformat_ohlcv',
     'dataformat_trades',
 ]
+
+CANCEL_REASON = {
+    "TIMEOUT": "cancelled due to timeout",
+    "PARTIALLY_FILLED": "partially filled - keeping order open",
+    "ALL_CANCELLED": "cancelled (all unfilled orders cancelled)",
```
hroff-1902, line 326:
```suggestion
    "ALL_CANCELLED": "cancelled (all unfilled and partially filled open orders cancelled)",
```

freqtrade/freqtradebot.py
```diff
@@ -162,6 +165,13 @@ def process(self) -> None:
 
         Trade.session.flush()
 
+    def process_stopped(self) -> None:
+        """
+        Close all trades that were left open
```
hroff-1902, line 170:
```suggestion
        Close all orders that were left open
```

hroff-1902:
^^ Some cosmetics to the messages/comments added...

hroff-1902:
LGTM 👍 (I only corrected the message for cancelling, to be more precise)

jpribyl:
@hroff-1902 and @xmatthias sorry for the delay in reply there - my gf had knee surgery and I didn't any time to spend on this, glad we were able to merge it!! LMK if there are any lingering issues / bugs 🥳

hroff-1902:
@jpribyl thanks for employment, be careful...

PR #3215, opened 2020-04-25
Backtest use pairlists

xmatthias:
## Summary
Use pairlists for backtesting.
While i'm skeptical on using Volumepairlist for backtesting  since it will use the current 24h volume to select pairs (so pairs during dry-run would have been different), using the filters can make sense (especially once #2956 is in).

This also now respects the blacklist correctly - which was previously ignored by backtesting.

enables  #2956 (random pairlist) for backtesting

## Quick changelog

- Use pairlists for backtesting to respect blacklist.
 
## To discuss
- [ ]  maybe limit to `StaticPairlist` (and the filters) - since Volumepairlist can be misleading (it's not recalculated every 6 hours as it is in dry-run).

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -17,8 +17,12 @@ For details on downloading, please refer to the [Data Downloading](data-download
 
 The result of backtesting will confirm if your bot has better odds of making a profit than a loss.
 
-!!! Tip "Using dynamic pairlists for backtesting"
-    While using dynamic pairlists during backtesting is not possible, a dynamic pairlist using current data can be generated via the [`test-pairlist`](utils.md#test-pairlist) command, and needs to be specified as `"pair_whitelist"` attribute in the configuration.
+!!! Warning "Using dynamic pairlists for backtesting"
+    Using dynamic pairlists is possible, however it relies on the current market conditions - which will not reflect the historic status of the pairlist.
+    Also, when using pairlists other than StaticPairlist, reproducability of backtesting-results cannot be guaranteed.
+    Please read the [pairlists documentation](configuration.md#pairlists) for more information.
+
+    To achieve reproducible results, best generate a pairlist via the [`test-pairlist`](utils.md#test-pairlist) command and use that as static pairlist.
 
 ### Run a backtesting against the currencies listed in your config file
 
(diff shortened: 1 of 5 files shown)


hroff-1902:
"maybe limit to StaticPairlist (and the filters)" -- no, let's leave it for now at least

xmatthias:
> no, let's leave it for now at least

and that's what concerns me - as this PR is at this moment, it'll change current behaviour (assuming dynamicPairlist is used) ... so instead of backtesting the current pairlist, it'll use the current volumepairlist instead.

hroff-1902:
> > no, let's leave it for now at least
> 
> and that's what concerns me - as this PR is at this moment, it'll change current behaviour (assuming dynamicPairlist is used) ... so instead of backtesting the current pairlist, it'll use the current volumepairlist instead.

got it, let me think

hroff-1902:
Let's leave only StaticPairList for backtesting for now.

In the same time, I'd like to have an ability to enable VolumePairList in the experimental mode with a parameter (without changing the code) or for developers:
there can be the `--developer` option added, which 1) enables VolumePairList for backtesting 2) allows to use the exchanges that are marked as non-working (disables this check) 3) maybe something more in the future. Not in this PR.

xmatthias:
I'm not a huge fan of `--developer` - i'd rather have to edit the configuration explicitly to enable certain things.
Using a generic developer flag will enable "all" testing options ... which might influence your tests - so you'll not be able to use it without risking to taint your tests. But let's leave that to it's own issue ...

I've disallowed VolumePairList for now - but all filters are allowed - as that will allow to filter "shitcoins" easily.
The alternative would be to overwrite configuration for backtesting with `[{'method': 'StaticPairList'}]` ... but i'd prefer not to do that if possible.

hroff-1902:
just a question (to emphasize this): the filter in backtesting will work on current (at the moment of backtesting was launched) ticker/orderbook values, right?

xmatthias:
> the filter in backtesting will work on current (at the moment of backtesting was launched) ticker/orderbook values, right?

yes, absolutely. 
The information needed (bid/ask prices) are not available on a "historic" bases.
while for some "open/close" could be an option - i'm not certain if it's worth investing the effort into this - considering that you'll need to take current price into account before setting the strategy to production.

In general i think using a staticpairlist is better for backtesting for reproducability.
I mean - what can it tell you if dynamic pairlists would have selected SHIT/ETH 2 years ago and would have done great (or bad) on it - while this pair is now filtered out as it's worthless, so won't be used anytime soon.

PR #3218, opened 2020-04-25
[minor] Update "final" message in setup.sh

### LINKED ISSUE #2773: setup.sh suggests to run invalid command
`./setup.sh -i` suggests at the end:
```
-------------------------
Run the bot !
-------------------------
You can now use the bot by executing 'source .env/bin/activate; freqtrade trade'.
```

While `freqtrade trade` is not a valid command, a strategy should be supplied.

Should we suggest `--strategy DefaultStrategy`? Or `--strategy SampleStrategy`?

Or probably suggest to create a new custom strategy with the `new-strategy` subcommand?

xmatthias:
## Summary
setup.sh is showing a invalid command.
We should not pin users to `freqtrade trade` - but instead suggest that multiple commands are available.
Also - it'll show how to print the version.

closes #2773

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -252,7 +252,8 @@ function install() {
     echo "-------------------------"
     echo "Run the bot !"
     echo "-------------------------"
-    echo "You can now use the bot by executing 'source .env/bin/activate; freqtrade trade'."
+    echo "You can now use the bot by executing 'source .env/bin/activate; freqtrade <command>'."
+    echo "You verify that freqtrade is installed successfully by running 'source .env/bin/activate; freqtrade --version'."
 }
 
 function plot() {
--- a/setup.sh
+++ b/setup.sh
@@ -252,7 +252,8 @@ function install() {
     echo "-------------------------"
     echo "Run the bot !"
     echo "-------------------------"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -252,7 +252,8 @@ function install() {
     echo "-------------------------"
     echo "Run the bot !"
     echo "-------------------------"
-    echo "You can now use the bot by executing 'source .env/bin/activate; freqtrade trade'."
+    echo "You can now use the bot by executing 'source .env/bin/activate; freqtrade <command>'."
+    echo "You verify that freqtrade is installed successfully by running 'source .env/bin/activate; freqtrade --version'."
 }
```
hroff-1902, line 257: * Maybe add here a third line:
```
echo "You can now use the bot by executing 'source .env/bin/activate; freqtrade <command>'."
echo "You can see the list of available bot commands by executing 'source .env/bin/activate; freqtrade --help'."
echo "You verify that freqtrade is installed successfully by running 'source .env/bin/activate; freqtrade --version'."
```
?

* "command" --> "subcommand"?
(just to be aligned to the term we took from argparse and use in other places in the docs)

PR #3234, opened 2020-04-29
Use skopt model_queue_size instead of custom hack

hroff-1902:
* Use optimizer's model_queue_size param (was introduced with model queue management PR in skopt v0.7.2, IIRC)
* Remove my custom hack I implemented before to do same (in a bit different approach) things

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ plot = ['plotly>=4.0']
 hyperopt = [
     'scipy',
     'scikit-learn',
-    'scikit-optimize',
+    'scikit-optimize>=0.7.4',
     'filelock',
     'joblib',
     'progressbar2',
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ plot = ['plotly>=4.0']
 hyperopt = [
     'scipy',
     'scikit-learn',
-    'scikit-optimize>=0.7.4',
+    'scikit-optimize>=0.7.0',
     'filelock',
     'joblib',
     'progressbar2',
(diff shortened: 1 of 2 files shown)


xmatthias:
Just to understand - that means to have it work correctly, skopt >=0.7.2 will be required? 

if that's the case, i'd suggest adding this as min-required to `setup.py` (we have this for ccxt, too) - so it'll eventually update outdated installations when running `pip install freqtrade` or `pip install -e .`.

hroff-1902:
Just referred their release notes: the queue management was merged in skopt 0.7, see https://github.com/scikit-optimize/scikit-optimize/releases/tag/v0.7:

>Models queue has now a customizable size (model_queue_size).

The current version is 0.7.4 and we should already have it in our dependencies...

hroff-1902:
Added version check into setup.py, pls review if I'm correct.,,

setup.py
```diff
@@ -21,7 +21,7 @@
 hyperopt = [
     'scipy',
     'scikit-learn',
-    'scikit-optimize',
+    'scikit-optimize>=0.7.4',
```
xmatthias, line 24: Unless it's been fixed in 0.7.4, i would use:

```suggestion
    'scikit-optimize>=0.7.0',
```

Requirements in setup.py shold be as loose as possible ... requirements.txt (and consorts) on the other hand  pin the latest version (mainly for CI ).

hroff-1902: pls review the condition: the version of skopt where it was fixed is numbered as 0.7, will comparison `>=0.7.0` work with `0.7`? (assume yes, but anyway...)

hroff-1902:
rebased to fix a typo in the name of the commit...

PR #3238, opened 2020-05-01
Add ticker to dataprovider

hroff-1902:
Add an ability to use ticker data (for example, 'vwap' and 'last') in the strategies through the dataprovider.

TODO: docs, examples, if the implementation (and the overall idea to provide this info into the strategy) is okay...


xmatthias:
> and the overall idea to provide this info into the strategy

is this a joke? it says "TODO: implement me" ... since it works this way (and i would not know another way tbh) - i would not know any reason to reject this (assuming it has tests and docs, obviously) :D

hroff-1902:
> > and the overall idea to provide this info into the strategy
> 
> is this a joke? it says "TODO: implement me" ... since it works this way (and i would not know another way tbh) - i would not know any reason to reject this (assuming it has tests and docs, obviously) :D

I meant, 1) maybe it should be implemented in a more sophisticated way, with checking the exceptions, other checks and logs, and so on... 2) maybe the idea itself is not good (and we should not expose this to the strategy), because this can cause frequent bans from exchange; or a cached value should be returned, in this case it also should be implemented in completely different manner...

hroff-1902:
actually, for dataprovider, I guess 1) we should have both cached/momental versions for each of the methods returning info, to allow user to chose which one she needs to employ (except probably ohlcv) 2) methods of the dataprovider should only be used in the custom strategies, not in freqtrade.py/backtesting/etc, not in the internals of the bot -- there are "native" functions in other parts of the bot (exchange, etc) which are available for other parts of the bot and are not available/should not be used in the custom strategies

hroff-1902:
^^ at least it works well with my current Orderbook Imbalance strategy, on Binance, I'm currently experimenting with, which invokes it for 15 default pairs each 1m, no bans from Binance

hroff-1902:
There is a question about interface of this method:

should it return None in case of errors (as `dp.market()` does) or raise an exception, which then is caught outside of the custom user's code (as currently `exchange,fetch_ticker()` does)?

xmatthias:
I would try not to burdon users with exception-handling if possible.

Assuming it's a trader, then telling him to wrap a line in a try-except block will not be very userfriendly.
Checking "is it there" on the other hand i hope can be expected... :)

hroff-1902:
Returning `{}` is okay. 👍 

(Just a note, I'm not against the interface which does not throw exceptions):
In case of exceptions, there is no need to do the handling by the user, the exception will go to the upper levels and cought by the bot...

xmatthias:
well in theory - but what if a user want's a different handling in case of an error? 
even a non-developer can (and will) have that need - everyone is familiar with the concept of "if this doesn't work, try ..." ... ;)

hroff-1902:
I added the test,
@xmatthias pls review, maybe it can be made easier with introduction of fetch_ticker() in conftest, by I have not found a way how to do it (how to pass a parameter in the method mocking the api call, except with that lambda)

tests/data/test_dataprovider.py
```diff
@@ -154,6 +154,23 @@ def test_market(mocker, default_conf, markets):
     assert res is None
 
 
+def test_ticker(mocker, default_conf, tickers, markets):
+    api_mock = MagicMock()
+    api_mock.markets = markets
+    api_mock.fetch_ticker = MagicMock(side_effect=lambda x: tickers().get(x))
+    exchange = get_patched_exchange(mocker, default_conf, api_mock=api_mock)
```
xmatthias, line 161: Usually, (unless we're testing the exchange class) - we should mock out the Exchange's methods - not the CCXT methods (which is what using api_mock does).

I think a better (or rather, more aligned method as to how mocking is handled in the rest of the tests would be:

```suggestion
     mocker.patch('freqtrade.exchange.Exchange.fetch_ticker', ticker)
    exchange = get_patched_exchange(mocker, default_conf, mock_markets=True)
```

Now i didn't test this (so typos may be included) - but i think this should work.

hroff-1902: `ticker` does not exist, we only have `tickers` in conftest... and the mocking function should issue same exception in the case of non-existing pair, as the Exchange method it mocks, to verify that the exception is probably handled by the higher-level dataprovider method...

xmatthias: well ... no?
the fixture ticker exists - line 326 of conftest ... :) 

I think you need to be clear what you're trying to test - it's either `fetch_ticker()` behavior (we already have tests for this in `test_exchange.py`...) or it's the behavior of the new dataprovider method - which is mostly a "pass-through".

I don't mind having the test as is ... i was just proposing a different method since you asked.

One thing: i think the `MagicMock(side_effect=` part on the fetch_ticker mock is not needed - you can monkey-patch that function directly by assigning the lambda to it...

docs/strategy-customization.md
```diff
@@ -451,6 +452,17 @@ if self.dp:
     The order book is not part of the historic data which means backtesting and hyperopt will not work if this
     method is used.
 
+#### *ticker(pair)*
+
+``` python
+if self.dp:
+    if self.dp.runmode.value in ('live', 'dry_run'):
+        ticker = self.dp.ticker(metadata['pair'])
+        dataframe['last_price'] = ticker['last']
```
xmatthias, line 461: I think we should note here that fields can depend on the exchange (not all exchanges return vwap for example) - and FTX does not always return last field - so it can be None ...

tests/data/test_dataprovider.py
hroff-1902: thanks for advices, got it

I definitely want to test dp.ticker() here (handling of exceptions from the lower level, etc), and Exchange.fetch_ticker() was not touched and it's testing is out of scope...

pls review the updated test

docs/strategy-customization.md
```diff
@@ -451,6 +452,23 @@ if self.dp:
...
+        dataframe['volume24h'] = ticker['quoteVolume']
+        dataframe['vwap'] = ticker['vwap']
+```
+
+!!! Warning
+    Although the ticker data structure is a part of the ccxt Unified Interface, the values returned by this method can
+    vary for different exchanges. For instance, many exchanges do not return `vwap` values, the FTX exchange
+    does not always fills in the `last` field (so it can be None), etc. So you need to carefully verify the ticker
+    data returned from the exchange you are working with.
```
xmatthias, line 470:
```suggestion
    Although the ticker data structure is a part of the ccxt Unified Interface, the values returned by this method can
    vary for different exchanges. For instance, many exchanges do not return `vwap` values, the FTX exchange
    does not always fills in the `last` field (so it can be None), etc. So you need to carefully verify the ticker
    data returned from the exchange and add appropriate error handling / defaults.
```

PR #3242, opened 2020-05-02
New subcommand show-trades

xmatthias:
## Summary
Add subcommand `show-trades` - which can be used to print one or more trades to screen.
This can be very helpful for analysis and issue triage - as we'll not have to ask for "open the database and ... " - but we can say "run `freqtrade show-trades --dburl <sqlite... > --print-json` - and get a nicely formatted result back.

## Quick changelog

- New command show-trades

### DIFF
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -426,8 +426,8 @@ AVAILABLE_CLI_OPTIONS = {
         default="file",
     ),
     "trade_ids": Arg(
-        '--tradeids',
-        help='Specify Trade ids to trade',
+        '--trade-ids',
+        help='Specify the list of trade ids.',
         nargs='+',
     ),
     # hyperopt-list, hyperopt-show
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -217,7 +217,7 @@ AVAILABLE_CLI_OPTIONS = {
     ),
     "print_json": Arg(
         '--print-json',
-        help='Print best result detailization in JSON format.',
+        help='Print output in JSON format.',
         action='store_true',
         default=False,
     ),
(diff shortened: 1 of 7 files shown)


hroff-1902:
The command is marvelous and should be useful,

now it only prints
```
...
Trade(id=1104, pair=ATOM/BTC, amount=31.11000000, open_rate=0.00032140, open_since=closed)
...
Trade(id=1127, pair=XRP/BTC, amount=406.17384240, open_rate=0.00002462, open_since=2020-05-02 12:23:16)
...
```

-- I would like to see here
* more humanized output (with no `Trade()`)
* tabular output (maybe with additional `--print-table`)
* more data (at least close_rate, close_time, profit)...

freqtrade/commands/cli_options.py
```diff
@@ -425,6 +425,11 @@ def __init__(self, *args, **kwargs):
         choices=["DB", "file"],
         default="file",
     ),
+    "trade_ids": Arg(
+        '--tradeids',
```
hroff-1902, line 429:
```suggestion
        '--trade-ids',
```

-- to be aligned with other options, like `db-url`, which have `-` in the name...
(fix all other appearances of this option as well)

freqtrade/commands/cli_options.py
```diff
@@ -425,6 +425,11 @@ def __init__(self, *args, **kwargs):
...
+        help='Specify Trade ids to trade',
```
hroff-1902, line 430:
```suggestion
        help='Specify the list of trade identifiers',
```

xmatthias:
Why complicate things?
I'm even tempted to remove the --json flag and have that as the only possible output form.

It's mainly a debug command - so we can ask for the output of this when there's a problem with a trade and we need the content of the database.

I don't see a point in forcing every command to have all possibilities (json, csv, table, ...) - that'll prevent us from implementing new commands.

Feel free to add this in a followup PR (or add it here if it's a requirement) if you think it's useful - i don't think it is, and therefore don't plan to add that myself.

hroff-1902:
okay, this can be enhanced further

hroff-1902:
I won't announce it in Slack since it's more a supplemental command for the support process rather than one which is to be used by the users to see the trades info, right?

xmatthias:
> okay, this can be enhanced further

in all honesty, everything can. This is intended to be a "quick help" for ourselfs to get the data in a easily readable format.
Further improvements will and can always follow (just look at list/show hyperopt - which is improved every other week).

PR #3262, opened 2020-05-05
Humanize show-trades error when no database is specified

### LINKED ISSUE #3261: missing db_url for show_trades
Currently 
```
$ freqtrade show-trades
```
gives
```
...
2020-05-05 20:18:27,042 - freqtrade - ERROR - Fatal exception!
  File "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/freqtrade/main.py", line 36, in main
  File "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/freqtrade/commands/list_commands.py", line 209, in start_show_trades
    init(config['db_url'], clean_open_orders=False)
KeyError: 'db_url'
```

Instead, a human-readable error message should be processed like for missing `--exchange` for `list-markets`:

```
freqtrade - ERROR - This command requires a configured exchange. You should either use `--exchange <exchange_name>` or specify a configuration file via `--config`.
```

-- similar should be printed for show_trades.

xmatthias:
## Summary
Humanize show-trades error when no database is specified

database url (db_url) can be specified in either configuration or explicitly.
Fallback to Default url's is intentionally NOT supported - as it would create the database if it doesn't exist - and just from this command it's not clear if it's a dry-run database or a production database.

Closes #3261

### DIFF
--- a/freqtrade/commands/list_commands.py
+++ b/freqtrade/commands/list_commands.py
@@ -206,6 +206,11 @@ def start_show_trades(args: Dict[str, Any]) -> None:
     from freqtrade.persistence import init, Trade
     import json
     config = setup_utils_configuration(args, RunMode.UTIL_NO_EXCHANGE)
+
+    if 'db_url' not in config:
+        raise OperationalException("--db-url is required for this command.")
+
+    logger.info(f'Using DB: "{config["db_url"]}"')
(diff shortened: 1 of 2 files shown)


xmatthias:
> sorry for disturbing,

we should have noticed that before merging the original one ... but i had a config so never noticed ... :disappointed:

Issue #3240, opened 2020-05-01
Dynamic Informative pairs


pAulseperformance:
**Using the informative pairs method is essentially limited to static pairs known before initialization.** 

This causes an impediment if I want to use a dynamic pairlist like a volumepairlist, as I won't know which pairs I need to define until after the bot has initialized. 

To enhance this method for dynamic pairs I found a solution that worked for me. 

I simply attached the PairListManager instance to an Istrategy attribute during the FreqtradeBot initialization. This allows access to the pairlist when the informative_pairs method is called. 
  
    # freqtrade.py

    Class FreqtradeBot:
        def __int__...
            # init other stuff
        
            self.pairlists = PairListManager(self.exchange, self.config)
            # Attach pairLists to Strategy baseclass
            IStrategy.pairlists = self.pairlists
        

Then during the initialization of the strategy. When informative_pairs() is called...

    # mystrategy.py
    
    ...
    ticker_interval = '1m'
    ...
    
    def informative_pairs(self):
        pairs = self.pairlists.whitelist
        informative_pairs = [(pair, '1d') for pair in pairs]
        return informative_pairs
        
I'm not sure if this is the best approach, but it works for me.

xmatthias:
at first i thought "but you can change the return-value of `informative_pairs()`with every iteration and it would pick it up" - but yeah, you're right, it's not working as you don't know your current whitelist.

While this method works, i would probably want to abstract this through dataProvider ... so it's a "unified" user-interface for all data / additional info methods

basically so you could call it as `self.dp.current_whitelist` ... (or something like that ...).

GrilledChickenThighs:
Yea keeping it under dp would be more intuitive and unified.

xmatthias:
Comments from @hroff-1902  in #3267 - migrated here to keep the PR clean:

My old idea (stash with it was lost) was to allow
```
    def informative_pairs(self):
        return [(None, "5m"),
                   ]
```
or
```
    def informative_pairs(self):
        return [("", "1d"),
                   ]
```
and correspondingly, in populate_indicators() or in populate_buy/sell_trend():
```
    informative = self.dp.get_pair_dataframe(pair=None,  # or ""
                                             timeframe="1d")
```
to fetch the data for the current pair but with different timeframe...

--------

Another idea, to implement it with a decorator, which adds a parameter to the strategy method with ohlcv for informative pair (also, if pair is omitted it means current pair):

```
    @add_informative(timeframe="1d")
[...]

PR #3267, opened 2020-05-06
Method for accessing current pairlist inside strategy.

GrilledChickenThighs:
## Summary

Solve the issue: #3240 

## Quick changelog

- Added Method for accessing current pair list on initialization for dynamic informative pairs

## What's new?
Now you can access every available pair inside of informative pairs without having to write out each pair one by one.

 - This is useful for anyone who wants to use a dynamic pairlist like VolumePairList accompanied by informative pairs. Previously, you couldn't do this because the pairs are dynamically generated; unless you wrote out every available pair on the exchange.

 - This is also helpful for anyone who wants to scan ALL pairs of an exchange with their bot, while also having access to other timeframes.

```
# SampleStrategy.py

def informative_pairs(self):

        # get access to all pairs available in whitelist.
        pairs = self.dp.current_whitelist()

        # Assign tf to each pair so they can be downloaded and cached for strategy.
        informative_pairs = [(pair, '1d') for pair in pairs]

        return informative_pairs
```


freqtrade/data/dataprovider.py
```diff
@@ -116,3 +116,17 @@ def runmode(self) -> RunMode:
...
+
+        Useful when you have a large whitelist and need to call each pair as an informative pair.
+        As available pairs does not show whitelist until after informative pairs have been cached.
+        :return: list of pairs in whitelist
+        """
+        from freqtrade.pairlist.pairlistmanager import PairListManager
+
+        pairlists = PairListManager(self._exchange, self._config)
+        pairlists.refresh_pairlist()
+        return pairlists.whitelist
```
xmatthias, line 132: i don't think this is a good way of implementing this.

It's creating a second instance of PairListManager. 
That's not a problem per se - but PairListManager does some caching (of tickers, for example) - and also of the VolumePairlist - so if you're not calling this in the same frequence you're calling the real pairlist, the 2 can become misaligned - so you get a pairlist - but not necessarily the same one the bot is using to create trades.
Also, it's adding additional network calls (to tickers, but to anything in theory assuming additonal pairlists are implemented) - so i would avoid having 2 instances around all the time.

My proposal for this woul'd be:
add it as a 3rd, (optional?) argument to `__init__()` - and keep the instance passed in as `self._pairlists` (or something similar).

That way, you won't have to call `refresh_pairlist()` but can rely on the call done by freqtradebot.
You can then check in this method if that exists or is None - and eventually return (or raise an Exception?) if dataprovider was not initialized with a pairlistprovider.

hroff-1902:
My old idea (stash with it was lost) was to allow
```
    def informative_pairs(self):
        return [(None, "5m"),
                   ]
```
or
```
    def informative_pairs(self):
        return [("", "1d"),
                   ]
```
and correspondingly
```
    informative = self.dp.get_pair_dataframe(pair=None,  # or ""
                                             timeframe="1d")
```
to fetch the data for the current pair but with different timeframe...

hroff-1902:
Another idea, to implement it with a decorator, which adds a parameter to the strategy method with ohlcv for informative pair (also, if pair is omitted it means current pair):

```
    @add_informative(timeframe="1d")
    def populate_indicators(self, dataframe: DataFrame, metadata: dict,
                                           df_1d: DataFrame) -> DataFrame:
        # RSI on "normal" ohlcv
        # RSI on "1d" ohlcv for current pair
        rsi_1d = ta.RSI(df_1d)
        # Any other calculations on rsi_1d can be done here
        # And finally it can be merged back with resampling to "normal" dataframe

```

hroff-1902:
^^ I should have better placed those ideas into #3240 for discussion, sorry

xmatthias:
i migrated your comments to the issue - that's where they belong, as this implementation uses another approach (in my opinion better, as i'm against the use of decorators - but please see details in the issue).

xmatthias:
I've had to remove the `@patch` stuff because 
a) i'm not familiar using it
b) it was mocking the wrong things
c) most mocks need to use fixtures to avoid duplicating the data all around the code - and i don't think these can be used with @patch (unittest vs. pytest fixtures ...).

-----
Now to explain the root of the problem:

By mocking PairListManager, all every instance of Pairlistmanager will return is a new MagicMock object, so `dp.current_whitelist()` did return a MagicMock, as it did not have any property on it's own (remember, you mocked the whole object, not just what you don't need).

By then assigning `_whitelist`, `_whitelist` becomes a list - however the property we use in DataProvider is not `_whitelist` - but `whitelist` (which remains a MagicMock object). 

Now how i approached this:
Leave PairlistManger alone / do it's job - and provide it with the data necessary (by mocking `ticker` and `exchange_has` of the exchange class).
That way, it returns a real, calculated pairlist, we can then compare.

I've added an additional assert verifying that it's the same identity (comparing with `is`) - as that's how it currently works.

---
Now this last point also raises another question - should we use `.copy()` for the whitelist instead?
currently, we're allowing the strategy to remove items from the whitelist.
For fun, i added the following line to my populate_indicators ... `self.dp.current_whitelist().pop()` the whitelist was getting small very soon ... 

Now depending on the configured pairlist (for me it was StaticPairlist) i assume this ain't a huge problem as the pairlist will be repopulated again with the next iteration - but i don't think this way should be possible...

@hroff-1902 opinions on this?

hroff-1902:
I'm pretty fine with the current tests, besides that the syntax with decorators makes mocking (global for the unit test function, not where it's done in the middle of the test for the rest of it) much more visible. So if in some future we could employ this syntax and the tests still will be working -- I'll be just happy.

Of course, if we do not need PairList mocked in the test, it should not be mocked, in either way or syntax...

docs/strategy-customization.md
```diff
@@ -334,17 +363,67 @@ Please always check the mode of operation to select the correct method to get da
...
+        informative_pairs = [(pair, '1d') for pair in pairs]
+        return informative_pairs
+
+    def populate_indicators(self, dataframe, metadata):
+        # Get the informative pair
+        informative = self.dp.get_pair_dataframe(pair=metadata['pair'], timeframe='1d')
+        # Get the 14 day ATR.
+        atr = ta.ATR(informative, timeperiod=14)
+        # Assign the Daily atr to the 1 minute dataframe.
+        dataframe['daily_atr'] = atr
```
xmatthias, line 420: I would argue that this might lead to unexpected results...

ATR is with frequency 1d - while `dataframe` is at 1m (or 5m...) frequency. 
So it'll initially lead to a ton of empty values - which will need to be ffilled to be comparable reliably.

it will also "throw away" all ATR values older than the 500 candles of the original timeframe - so for 1m timeframe, you'll only have one value (or no value) - depending if you're in the first 3rd of the day or not, as you'll only have 500 minutes of candles in the dataframe - and the ATR candle is dated to midnight.

Now using 5m timeframe might help with that ... but maybe the problem should be pointed out anyway ... just to be sure.

GrilledChickenThighs: Well it should be expected that anything that is averaged over a period of 14 days (with a daily sample rate) shouldn't change much in a period of 500 minutes. At most we would see two different values occurring around midnight...

The point is that we are able to get higher timeframe references to generate signals on lower timeframes. Alongside the dynamic accessing of our referenced pair. 

In my example I wasn't being too specific about the execution of the strategy or dataframe handling because it seemed arbitrary.  But I think you make a good point on using ffill if we were actually going to trade this example, but I didn't feel that would be appropriate for this part of the doc.

xmatthias: well to be fair - i've only noticed this since this week someone copy/pasted a full documentation sample which had a misspelled import ... :laughing: 

would you have added this last week - i probably wouldn't have noticed / thought about it.

GrilledChickenThighs: haha yea, it's a bit of a trade off we have to make when writing howtos for anything.... 

Either assume the reader has *some* knowledge or be rigorously thorough -- the latter of which usually takes significant time upfront, but arguably could cut down time in the long run due to having to explain issues that arise from poor documentation. 

That being said, I think a healthy middle ground is best. If someone has trouble with this part of the doc, I can always explain further and update the doc *ad totum* *ad finitum*

hroff-1902:
> LGTM 👍
> 
> gonna leave it unmerged for now to give @hroff-1902 a chance to look at it too though.

Sure,
I reviewed the code long time ago and just reviewed the docs in rich diff for better understanding what has been moved where and what has been added...

hroff-1902:
@GrilledChickenThighs Excellent, well done! 🥇 

LGTM 👍 , can be merged.

hroff-1902:
#3303 
😉

Issue #3268, opened 2020-05-08
Binance - Crash caused by "Unknown order sent" 


devilsinthedetail:
## Step 1: Have you search for this issue before posting it?
Yes, related issue here: https://github.com/freqtrade/freqtrade/issues/3208

## Step 2: Describe your environment

  * Operating system: Raspbian 2020-02-13
  * Python Version: 3.7.3 (`python -V`)
  * CCXT version: ccxt==1.27.1 (`pip freeze | grep ccxt`)
  * Branch:  Develop
  * Last Commit ID: 25d9001f8864d6a641037f7d6f8e0177df18d752 (`git log --format="%H" -n 1`)
  
## Step 3: Describe the problem:
Bot runs into an Error while live trading on Binance:
> ERROR - Could not cancel order. Message: binance {"code":-2011,"msg":"Unknown order sent."}
My sell config is: 
"ask_strategy": 
{
"price_side": "ask",
"use_order_book": true,
"order_book_min": 1,
"order_book_max": 5,
"use_sell_signal": true,
"sell_profit_only": false,
"ignore_roi_if_buy_signal": false
},

### Relevant code exceptions or logs:

  ```
2020-05-08 01:08:41,485 - freqtrade.worker - INFO - Bot heartbeat. PID=3083, version='develop-25d9001f', state='RUNNING'
2020-05-08 01:09:42,183 - freqtrade.worker - INFO - Bot heartbeat. PID=3083, version='develop-25d9001f', state='RUNNING'
2020-05-08 01:10:43,760 - freqtrade.worker - INFO - Bot heartbeat. PID=3083, version='develop-25d9001f', state='RUNNING'
2020-05-08 01:11:00,121 - freqtrade.freqtradebot - INFO - Executing Sell for KNC/ETH. Reason: SellType.ROI
2020-05-08 01:11:00,941 - freqtrade.wallets - INFO - Wallets synced.
2020-05-08 01:11:01,474 - freqtrade.freqtradebot - INFO - Getting price from order book Ask side.
[...]

PR #3269, opened 2020-05-08
[minor] Improve exception handling on critical errors

xmatthias:
## Summary
As identified in #3268, logging does not properly print the stacktrace for certain errors.

This happens because the except / finally handler "swallows" the exception and does not bubble it further.

Catching all errors in the trade_command module will solve this issue (no other command uses a finally handler).

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -115,6 +115,28 @@ def test_main_operational_exception(mocker, default_conf, caplog) -> None:
     assert log_has('Oh snap!', caplog)
 
 
+def test_main_operational_exception1(mocker, default_conf, caplog) -> None:
+    patch_exchange(mocker)
+    mocker.patch(
+        'freqtrade.commands.list_commands.available_exchanges',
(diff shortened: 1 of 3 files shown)


tests/test_main.py
```diff
@@ -115,6 +115,28 @@ def test_main_operational_exception(mocker, default_conf, caplog) -> None:
...
+
+    args = ['list-exchanges']
+
+    # Test Main + the KeyboardInterrupt exception
+    with pytest.raises(SystemExit):
+        main(args)
+
+    assert log_has('Fatal exception!', caplog)
+    assert not log_has_re(r'SIGINT.*', caplog)
+    mocker.patch(
```
hroff-1902, line 134: what this patching if for in this place? (after testing asserts?)

xmatthias: hm - committing the remainder of the test would help - wouldn't it? :D 

it's there now ...

PR #3276, opened 2020-05-09
Rename fetch_ticker() in Exchange

hroff-1902:
...to get_ticker(). To be aligned with other Exchange methods, fetch_ticker is the only one with 'fetch' prefix used.

It also allows better search for ccxt api fetch_ticker/fetch_tickers methods in the code and tests.

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -619,7 +619,7 @@ class Exchange:
             raise OperationalException(e) from e
 
     @retrier
-    def fetch_ticker(self, pair: str) -> dict:
+    def get_ticker(self, pair: str) -> dict:
         try:
             if pair not in self._api.markets or not self._api.markets[pair].get('active'):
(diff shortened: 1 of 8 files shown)


xmatthias:
hm - to be fair, i've renamed this just a few months ago to `fetch_ticker` to have it aligned to the ccxt methods.

My plan was to align freqtrade naming (the other methods) to ccxt naming - so when i think "fetch_ticker" - it is "fetch_ticker" in plain ccxt too - so to be honest, i'm against this change and would rather see "all" other methods renamed to be equally named wrappers around their ccxt counterparts.

hroff-1902:
Okay, no problem, I just found this is the only method named in this way...

xmatthias:
i know - it's a time question :( this has VERY low priority as it'll benefit only us ... but no user - at least not to start with.

hroff-1902:
sure, no problem

PR #3277, opened 2020-05-09
Fix fee usage in Edge

hroff-1902:
Edge should apply the value of fee twice, not a half of it for buy and sell.

As it's used in backtesting and as defined in the description of the `--fee` cli option:
"Specify fee ratio. Will be applied twice (on trade entry and exit)".

`Exchange.get_fee()`, which is used by Edge, also returns 0.001 for Binance, so it should be applied twice, not halved.

Expectancy and risk reward ratio decreased in the test, as it should be when we apply a higher fee.
Old data kept in the tests, but with halved fee value.

### DIFF
--- a/freqtrade/edge/edge_positioning.py
+++ b/freqtrade/edge/edge_positioning.py
@@ -242,10 +242,6 @@ class Edge:
         # All returned values are relative, they are defined as ratios.
         stake = 0.015
 
-        # Fee should be applied twice
-        open_fee = self.fee
-        close_fee = self.fee
-
         result['trade_duration'] = result['close_time'] - result['open_time']
 
         result['trade_duration'] = result['trade_duration'].map(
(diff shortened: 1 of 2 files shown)


freqtrade/edge/edge_positioning.py
```diff
@@ -238,20 +238,13 @@ def _fill_calculable_fields(self, result: DataFrame) -> DataFrame:
...
+        # We set stake amount to an arbitrary amount, as it doesn't change the calculation.
+        # All returned values are relative, they are defined as ratios.
         stake = 0.015
-        fee = self.fee
-        open_fee = fee / 2
-        close_fee = fee / 2
+
+        # Fee should be applied twice
+        open_fee = self.fee
+        close_fee = self.fee
```
xmatthias, line 247: The assignment of this to two seperate variables seems unnecesary.

I know it's just pointers - but they're not changed (and only ever used once after this line), so why not use `self.fee` below as well (or at least assign it to `fee = self.fee`)?

I do remember that i did notice this "halving" problem before - but i don't remember why it wasn't fixed then. 

@mishaker - i think i discussed this with you - hope you remember (?)

hroff-1902: I think it was "reserved" for the future, when different fees for close and open can be used...

xmatthias: but for backtesting / edge - that's never really applicable as we lack the info for that.

hroff-1902: well, theoretically, there can be `--close-fee` and `--open-fee`, instead of explaining that `--fee` is a cumulative for both, which can be misunderstooden...

mishaker: That's right. it is not necessary. using self.fee is enough.

hroff-1902: Done

PR #3278, opened 2020-05-10
API server - support JWT

xmatthias:
## Summary
In order to allow an UI to be more secure, the api server should support JWT (JSON Web Tokens), so the password isn't sent around in plaintext all the time.

## Quick changelog

- Support JWT 
- Support JWT refresh tokesn
- use `safe_str_cmp` to compare username/password.

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -208,7 +208,7 @@ whitelist
 !!! Note
     The below should be done in an application (a Freqtrade REST API client, which fetches info via API), and is not intended to be used on a regular basis.
 
-Freqtrade's REST API also offers JWT tokens.
+Freqtrade's REST API also offers JWT (JSON Web Tokens).
 You can login using the following command, and subsequently use the resulting access_token.
 
 ``` bash
@@ -216,7 +216,7 @@ You can login using the following command, and subsequently use the resulting ac
 {"access_token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE1ODkxMTk2ODEsIm5iZiI6MTU4OTExOTY4MSwianRpIjoiMmEwYmY0NWUtMjhmOS00YTUzLTlmNzItMmM5ZWVlYThkNzc2IiwiZXhwIjoxNTg5MTIwNTgxLCJpZGVudGl0eSI6eyJ1IjoiRnJlcXRyYWRlciJ9LCJmcmVzaCI6ZmFsc2UsInR5cGUiOiJhY2Nlc3MifQ.qt6MAXYIa-l556OM7arBvYJ0SDI9J8bIk3_glDujF5g","refresh_token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE1ODkxMTk2ODEsIm5iZiI6MTU4OTExOTY4MSwianRpIjoiZWQ1ZWI3YjAtYjMwMy00YzAyLTg2N2MtNWViMjIxNWQ2YTMxIiwiZXhwIjoxNTkxNzExNjgxLCJpZGVudGl0eSI6eyJ1IjoiRnJlcXRyYWRlciJ9LCJ0eXBlIjoicmVmcmVzaCJ9.d1AT_jYICyTAjD0fiQAr52rkRqtxCjUGEMwlNuuzgNQ"}
 
 > access_token="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE1ODkxMTk2ODEsIm5iZiI6MTU4OTExOTY4MSwianRpIjoiMmEwYmY0NWUtMjhmOS00YTUzLTlmNzItMmM5ZWVlYThkNzc2IiwiZXhwIjoxNTg5MTIwNTgxLCJpZGVudGl0eSI6eyJ1IjoiRnJlcXRyYWRlciJ9LCJmcmVzaCI6ZmFsc2UsInR5cGUiOiJhY2Nlc3MifQ.qt6MAXYIa-l556OM7arBvYJ0SDI9J8bIk3_glDujF5g"
-# Use tccess_token for authentication
+# Use access_token for authentication
 > curl -X GET --header "Authorization: Bearer ${access_token}" http://localhost:8080/api/v1/count
 
 ```
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -221,7 +221,7 @@ You can login using the following command, and subsequently use the resulting ac
 
 ```
 
-Since the access-token has a short timeout (15 min) - the refresh-token should be used to get a fresh access token:
+Since the access token has a short timeout (15 min) - the `token/refresh` request should be used periodically to get a fresh access token:
 
 ``` bash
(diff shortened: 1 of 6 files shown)


docs/rest-api.md
```diff
@@ -202,3 +202,28 @@ whitelist
         Show the current whitelist
         :returns: json object
 ```
+
+## Advanced API usage using JWT tokens
+
+!!! Note
+    The below should be done in an application, and is not intended to be used on a regular basis.
```
hroff-1902, line 209:
```suggestion
    The below should be done in an application (a Freqtrade REST API client, which fetches info via API), and is not intended to be used on a regular basis.
```

docs/rest-api.md
```diff
@@ -202,3 +202,28 @@ whitelist
...
+
+Freqtrade's REST API also offers JWT tokens.
```
hroff-1902, line 211:
```suggestion
Freqtrade's REST API also offers JWT (JSON Web Tokens).
```

docs/rest-api.md
```diff
@@ -202,3 +202,28 @@ whitelist
...
+You can login using the following command, and subsequently use the resulting access_token.
+
+``` bash
+> curl -X POST --user Freqtrader http://localhost:8080/api/v1/token/login
+{"access_token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE1ODkxMTk2ODEsIm5iZiI6MTU4OTExOTY4MSwianRpIjoiMmEwYmY0NWUtMjhmOS00YTUzLTlmNzItMmM5ZWVlYThkNzc2IiwiZXhwIjoxNTg5MTIwNTgxLCJpZGVudGl0eSI6eyJ1IjoiRnJlcXRyYWRlciJ9LCJmcmVzaCI6ZmFsc2UsInR5cGUiOiJhY2Nlc3MifQ.qt6MAXYIa-l556OM7arBvYJ0SDI9J8bIk3_glDujF5g","refresh_token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE1ODkxMTk2ODEsIm5iZiI6MTU4OTExOTY4MSwianRpIjoiZWQ1ZWI3YjAtYjMwMy00YzAyLTg2N2MtNWViMjIxNWQ2YTMxIiwiZXhwIjoxNTkxNzExNjgxLCJpZGVudGl0eSI6eyJ1IjoiRnJlcXRyYWRlciJ9LCJ0eXBlIjoicmVmcmVzaCJ9.d1AT_jYICyTAjD0fiQAr52rkRqtxCjUGEMwlNuuzgNQ"}
+
+> access_token="eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE1ODkxMTk2ODEsIm5iZiI6MTU4OTExOTY4MSwianRpIjoiMmEwYmY0NWUtMjhmOS00YTUzLTlmNzItMmM5ZWVlYThkNzc2IiwiZXhwIjoxNTg5MTIwNTgxLCJpZGVudGl0eSI6eyJ1IjoiRnJlcXRyYWRlciJ9LCJmcmVzaCI6ZmFsc2UsInR5cGUiOiJhY2Nlc3MifQ.qt6MAXYIa-l556OM7arBvYJ0SDI9J8bIk3_glDujF5g"
+# Use tccess_token for authentication
```
hroff-1902, line 219:
```suggestion
# Use access_token for authentication
```

docs/rest-api.md
```diff
@@ -202,3 +202,28 @@ whitelist
...
+> curl -X GET --header "Authorization: Bearer ${access_token}" http://localhost:8080/api/v1/count
+
+```
+
+Since the access-token has a short timeout (15 min) - the refresh-token should be used to get a fresh access token:
```
hroff-1902, line 224:
```suggestion
Since the access token has a short timeout (15 min) - the `token/refresh` request should be used periodically to get a fresh access token:
```

hroff-1902:
> use safe_str_cmp to compare username/password.

argh, I thought I left str_cmp nightmare in the times of pure C programming... 😆

xmatthias:
well it's not less safe than the previous version we had ... though most samples i found online seem to use this. 
The only thing it seems to do is explicitly convert to utf8 first - which seems sensible.

hroff-1902:
> well it's not less safe than the previous version we had ... though most samples i found online seem to use this.
> The only thing it seems to do is explicitly convert to utf8 first - which seems sensible.

no, it's against timing attacks with loooooong password/token/whatever strings are passed in the malicious requests, can 'freeze' the server and cause DoS; this function implements provisions against it.

freqtrade/rpc/api_server.py
```diff
@@ -38,9 +43,10 @@ def default(self, obj):
 def require_login(func: Callable[[Any, Any], Any]):
 
     def func_wrapper(obj, *args, **kwargs):
-
+        verify_jwt_in_request_optional()
         auth = request.authorization
-        if auth and obj.check_auth(auth.username, auth.password):
+        i = get_jwt_identity()
+        if i or auth and obj.check_auth(auth.username, auth.password):
```
hroff-1902, line 49:
```suggestion
        if get_jwt_identity() or auth and obj.check_auth(auth.username, auth.password):
```

freqtrade/rpc/api_server.py
```diff
@@ -83,6 +89,11 @@ def __init__(self, freqtrade) -> None:
 
         self._config = freqtrade.config
         self.app = Flask(__name__)
+
+        # Setup the Flask-JWT-Extended extension
+        self.app.config['JWT_SECRET_KEY'] = 'super-secret'  # Change this!
```
hroff-1902, line 94: Do you want to release it as it is?

freqtrade/rpc/api_server.py
```diff
@@ -199,6 +214,37 @@ def page_not_found(self, error):
             'code': 404
         }), 404
 
+    @require_login
+    @rpc_catch_errors
+    def _login(self):
```
hroff-1902, line 219: should probably be named `_token_login` ? (to reflect the url)

freqtrade/rpc/api_server.py
```diff
@@ -199,6 +214,37 @@ def page_not_found(self, error):
...
+                'access_token': create_access_token(identity=keystuff),
+                'refresh_token': create_refresh_token(identity=keystuff),
+            }
+            return self.rest_dump(ret)
+
+        return jsonify({"error": "Unauthorized"}), 401
+
+    @jwt_refresh_token_required
+    @rpc_catch_errors
+    def _refresh_token(self):
```
hroff-1902, line 237: also, should probably be `_token_refresh`?

freqtrade/rpc/api_server.py
xmatthias: i'm not sure what difference this key vs. a different key makes.

It makes a different if we assume it's a website we deploy and want noone to login.
In this case however, we're giving this to our users, who then run this as server ... so no matter WHAT we set this to - it'll be the same in all cases.

xmatthias: i'll move it to the configuration - so it's easily changable ... :)

hroff-1902: 👍 
otherwise the hashes/cookies etc can be predictable, I assume...
no need to make an utility for changing/generating it, or other heavy stuff (yet)...

PR #3290, opened 2020-05-12
[minor] Show forcebuy status so it's visible before calling forcebuy.

xmatthias:
## Summary
Show forcebuy status so it's visible before calling forcebuy.

Otherwise we need to call /forcebuy to know if it's available (via API...).

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -104,6 +104,7 @@ class RPC:
             'ticker_interval': config['ticker_interval'],
             'exchange': config['exchange']['name'],
             'strategy': config['strategy'],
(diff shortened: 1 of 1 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -104,6 +104,7 @@ def _rpc_show_config(self) -> Dict[str, Any]:
             'ticker_interval': config['ticker_interval'],
             'exchange': config['exchange']['name'],
             'strategy': config['strategy'],
+            'forcebuy_enabled': self._freqtrade.config.get('forcebuy_enable', False),
```
hroff-1902, line 107:
```suggestion
            'forcebuy_enabled': config.get('forcebuy_enable', False),
```

PR #3291, opened 2020-05-12
Docs: Fix samples in strategy-advanced.md

hroff-1902:
Fix sample (no timestamp in datetime, and not used in the samples)

As reported on Slack, thanks to @Prossi79

### DIFF
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -20,7 +20,7 @@ It applies a tight timeout for higher priced assets, while allowing more time to
 The function must return either `True` (cancel order) or `False` (keep order alive).
 
 ``` python
-from datetime import datetime, timestamp
+from datetime import datetime
 from freqtrade.persistence import Trade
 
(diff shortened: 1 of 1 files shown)


docs/strategy-advanced.md
```diff
@@ -20,7 +20,7 @@ It applies a tight timeout for higher priced assets, while allowing more time to
 The function must return either `True` (cancel order) or `False` (keep order alive).
 
 ``` python
-from datetime import datetime, timestamp
+from datetime import datetime
```
xmatthias, line 23: You're right while being wrong 😆 

timestamp is not used (nor does it exist) - what SHOULD be here is `timedelta` though - as that IS needed and used in the sample below.

hroff-1902: argh. okay

xmatthias:
```suggestion
from datetime import datetime, timedelta
```


for simple committing :D

PR #3304, opened 2020-05-14
refactor freqtradebot.py

Corfucinas:
## Summary
Refactor the function for readability 

## Quick changelog

- Fixed hoist repeated code outside conditional statement
- Add guard clause
- Bump ccxt to 1.27.56


hroff-1902:
for me it does not change anything, neither improves readability,
it replaces
```
if c:
    a = ...
    return a
b = ...
return b
```
with
```
if not c:
    b = ...
    return b
a = ...
return a
```

xmatthias:
I think the point is that it's been made by a bot - Sourcery - (see https://github.com/Corfucinas/freqtrade/pull/5).

The problem this change is that this method was almost completely rewritten in #3210, so this change is no longer relevant.

hroff-1902:
so closing it, taking into account the comments we made above

PR #3307, opened 2020-05-14
Add note about PI and docker

xmatthias:
## Summary
Add a note explaining which image should be used on a raspberry.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -22,6 +22,9 @@ Freqtrade provides an official Docker image on [Dockerhub](https://hub.docker.co
 !!! Note
     All below comands use relative directories and will have to be executed from the directory containing the `docker-compose.yml` file.
 
+!!! Note "Docker on Raspberry"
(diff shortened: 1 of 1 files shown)


docs/docker.md
```diff
@@ -22,6 +22,9 @@ Freqtrade provides an official Docker image on [Dockerhub](https://hub.docker.co
 !!! Note
     All below comands use relative directories and will have to be executed from the directory containing the `docker-compose.yml` file.
 
+!!! Note "Docker on Raspberry"
+    If you're running freqtrade on a Raspberry PI, you must change the image from `freqtradeorg/freqtrade:master` to `freqtradeorg/freqtrade:master_pi` or ``freqtradeorg/freqtrade:develop_pi`, otherwise the image will not work.
```
hroff-1902, line 26:
```suggestion
    If you're running freqtrade on a Raspberry PI, you must change the image from `freqtradeorg/freqtrade:master` to `freqtradeorg/freqtrade:master_pi` or `freqtradeorg/freqtrade:develop_pi`, otherwise the image will not work.
```

xmatthias: THIS is why i usually don't do things in the webui ...

PR #3308, opened 2020-05-14
Added Telegram command "logs [n]"

pnmartinez:
## Summary
Telegram command to check last [n] entries of the logs.

## Quick changelog

- freqtrade/freqtrade/rpc/telegram.py
    - `import os` added to access the logs on the filesystem,
    - Added _logs function, 
    - command handler, 
    - and command explanation on the "/help" message.

## What's new?
Checking the current status of the bot can be useful while being afk. This command enables to check the logfile from the Telegram bot; the text formatting may be not smartphone friendly, but it helped me in more than one occasion.

Posting in case it is useful for anyone else.

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -208,7 +208,7 @@ class Telegram(RPC):
                  self._send_msg(str(e))
 
         try:
-            location = '/freqtrade/user_data/logs'
+            location = '../../user_data/logs'
             for filename in os.listdir(location):
                 if filename == 'freqtrade.log':
                     f  = open(os.path.join(location, 'freqtrade.log'), "r")
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -4,6 +4,8 @@
 This module manage Telegram communication
 """
 import logging
+import os # added to acces log file on filesystem
+
 from typing import Any, Callable, Dict
 
 from tabulate import tabulate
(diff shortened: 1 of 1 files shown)


hroff-1902:
it will not work if the path to the logfile was altered with the `--logfile` option, this should be handled..

xmatthias:
it will not work by default, as logs are not written to disk unless explicitly specified...

I would like such a functionality myself - but i don't think relying on re-reading a log-file is a good aproach. 
Worst case is that it'll lock the bot due to concurrency ...

hroff-1902:
there is a queue-based logging available in the python world,
we could add the handler for it, keeping some amount (100, 1000) last logging messages in the queue, which can be requested and sent to the rpc client.
queueing the logs will also resolve the problem with first log messages (that are issued before the configuration is read and the log settings handled) that are now not displayed...

xmatthias:
> there is a queue-based logging available in the python world,
> we could add the handler for it, keeping some amount (100, 1000) last logging messages in the queue, which can be requested and sent to the rpc client.
> queueing the logs will also resolve the problem with first log messages (that are issued before the configuration is read and the log settings handled) that are now not displayed...

now this (queueing) is in my list of stashes ... but not fully tested nor completely implemented ...

xmatthias:
Thanks for your interrest and contribution.

However, for the above reasons (logging to file not default, reading from file may lead to locking, reading from file is bad practice, and may give different results depending if the log has been recently rotated or not) - this method implementing this is not the right way.

We're more than happy to provide / accept such a feature - but it should not re-read the log-file - but instead query logging module internally, which - if done correctly, will provide this information.

pnmartinez:
> Thanks for your interrest and contribution.
> 
> However, for the above reasons (logging to file not default, reading from file may lead to locking, reading from file is bad practice, and may give different results depending if the log has been recently rotated or not) - this method implementing this is not the right way.
> 
> We're more than happy to provide / accept such a feature - but it should not re-read the log-file - but instead query logging module internally, which - if done correctly, will provide this information.

No problem! Hoping that the quick fix that made for me would open the conversation about this. Thanks for the feedback.

PR #3312, opened 2020-05-15
[PoC] [WiP] Allow backtest exporting of candlestick data as well as trade data

woutervddn:
## Summary
Allow backtest exporting of candlestick data as well as trades using `--export all`

## Quick changelog

- allows for exporting of candlestick data and trades using `--export all`

## What's new?

You were already able to export a list backtest trades as json using `--export trades`. This PR enables the user to use `--export all` to export a json object that looks like this:

```
{
  'ETH/BTC': {
    trades: [ [Object] ],
    candles: [
      [Object], [Object], [Object], 
      ... 8475 more items
    ]
  },
  'LTC/BTC': {
    trades: [ [Object], [Object], [Object] ],
    candles: [
      [Object], [Object], [Object], 
      ... 8475 more items
    ]
  }, ...
}
```
The change keeps backwards compatibility by using the exact same function for exporting trades whenever `--export` is used without `all`.

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -382,6 +382,7 @@ class Backtesting:
         data, timerange = self.load_bt_data()
 
         all_results = {}
+        preprocessed_data = {}
         for strat in self.strategylist:
             logger.info("Running backtesting for Strategy %s", strat.get_strategy_name())
             self._set_strategy(strat)
@@ -398,6 +399,10 @@ class Backtesting:
                 'Backtesting with data from %s up to %s (%s days)..',
                 min_date.isoformat(), max_date.isoformat(), (max_date - min_date).days
             )
+
+            # Store reprocessed data for later use in export
+            preprocessed_data[self.strategy.get_strategy_name()] = preprocessed
+
             # Execute backtest and print results
             all_results[self.strategy.get_strategy_name()] = self.backtest(
                 processed=preprocessed,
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -398,6 +399,10 @@ def start(self) -> None:
                 'Backtesting with data from %s up to %s (%s days)..',
                 min_date.isoformat(), max_date.isoformat(), (max_date - min_date).days
             )
+
+            # Store reprocessed data for later use in export
+            preprocessed_data[self.strategy.get_strategy_name()] = preprocessed
```
xmatthias, line 404: You might wanna extract the "datadump" function from after the loop into here.

Assuming this is running for 20 strategies, you'd force me to be able to have enough RAM available to store the data 20+ times. 
Previously, the "per strategy" data was scoped to the for-loop, releasing memory after each iteration.

StefSmeers:
It might not be a bad idea to have a separate export key called `candlesticks` so your also able to export candlestick data separately. Now if you want candlesticks, you have no choice but exporting the trades as well.

MrKrautee:
Are You still working on this?

Seems to be an important enhancement, I missed too. 
Maybe I dig in if no one is working on this.
Any further suggestions?

xmatthias:
considering the age of this PR (and the fact that quite some things have changed in these code-areas) i'd say the best thing is to "extract" (copy/paste?) what's possible from this change to a new PR.


I'll be honest, personally i'm not interested in working at this at the moment as i don't see the benefit (maybe because my strategies calculate indicators within seconds for 100ds of pairs...), however Pull requests are always welcome 💯

PR #3313, opened 2020-05-15
Cleanup in pairlists

hroff-1902:
* unify a bit the structure of pairlists classes
* get rid of `filter()`
* various cosmetics (in pairlists and in the tests)

### DIFF
--- a/freqtrade/pairlist/PrecisionFilter.py
+++ b/freqtrade/pairlist/PrecisionFilter.py
@@ -1,6 +1,6 @@
 import logging
 from copy import deepcopy
-from typing import Dict, List
+from typing import Any, Dict, List
 
 from freqtrade.pairlist.IPairList import IPairList
 
@@ -8,6 +8,16 @@ logger = logging.getLogger(__name__)
 
 
 class PrecisionFilter(IPairList):
+
+    def __init__(self, exchange, pairlistmanager,
+                 config: Dict[str, Any], pairlistconfig: Dict[str, Any],
+                 pairlist_pos: int) -> None:
+        super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
+
+        self._stoploss = self._config.get('stoploss')
+        if self._stoploss is not None:
+            # Precalculate sanitized stoploss value to avoid recalculation for every pair
+            self._stoploss = 1 - abs(self._stoploss)
 
     @property
     def needstickers(self) -> bool:
@@ -49,15 +59,12 @@ class PrecisionFilter(IPairList):
         """
         Filters and sorts pairlists and assigns and returns them again.
         """
-        stoploss = self._config.get('stoploss')
-        if stoploss is not None:
-            # Precalculate sanitized stoploss value to avoid recalculation for every pair
-            stoploss = 1 - abs(stoploss)
         # Copy list since we're modifying this list
         for p in deepcopy(pairlist):
             ticker = tickers.get(p)
             # Filter out assets which would not allow setting a stoploss
-            if not ticker or (stoploss and not self._validate_precision_filter(ticker, stoploss)):
+            if not ticker or (self._stoploss
+                              and not self._validate_precision_filter(ticker, self._stoploss)):
                 pairlist.remove(p)
                 continue
 
--- a/freqtrade/pairlist/PrecisionFilter.py
+++ b/freqtrade/pairlist/PrecisionFilter.py
@@ -14,10 +14,8 @@ class PrecisionFilter(IPairList):
                  pairlist_pos: int) -> None:
         super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
 
-        self._stoploss = self._config.get('stoploss')
-        if self._stoploss is not None:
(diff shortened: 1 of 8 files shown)


freqtrade/pairlist/PrecisionFilter.py
```diff
@@ -31,34 +41,34 @@ def _validate_precision_filter(self, ticker: dict, stoploss: float) -> bool:
...
-            if not ticker or (stoploss and not self._validate_precision_filter(ticker, stoploss)):
-                pairlist.remove(p)
-                continue
+        if self._stoploss:
+            # Copy list since we're modifying this list
+            for p in deepcopy(pairlist):
+                ticker = tickers[p]
+                # Filter out assets which would not allow setting a stoploss
+                if not self._validate_precision_filter(ticker, self._stoploss):
+                    pairlist.remove(p)
```
xmatthias, line 72: I am not convinced that this does the same thing than the old code.

pairlist may contain a pair not in tickers (assuming the database contains a pair no longer traded) ... which would now raise a KeyError - which is not handled here.

Now handling the above case (pair in pairlist not trade anymore) is a different topic - but it should not raise here anyway.

freqtrade/pairlist/PrecisionFilter.py
```diff
@@ -9,6 +9,16 @@
 
 class PrecisionFilter(IPairList):
 
+    def __init__(self, exchange, pairlistmanager,
+                 config: Dict[str, Any], pairlistconfig: Dict[str, Any],
+                 pairlist_pos: int) -> None:
+        super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
+
+        self._stoploss = self._config.get('stoploss')
+        if self._stoploss is not None:
```
xmatthias, line 18: Are we sure this is needed? stoploss is a mandatory key for trade-runs ... so i'm not convinced stoploss can become None...

same applies below, obviously

freqtrade/pairlist/PrecisionFilter.py
hroff-1902: Right, but when wrote it I assumed Filters are called after VolumePairList (remember our hot dicussion in Slack? 😉 ), which handles it.
But you are right here -- Filters can also follow StaticPairList, which does not deal with markets...

I'll add temporary this check into the Filters again, but I'll refactor it soon -- this cleanup should be done once, called from the pairlistmanager, instead of a check implemented in each of the Filters...

freqtrade/pairlist/PrecisionFilter.py
hroff-1902: Well, other Filters and PairLists (Spread, Volume) does the same -- they store the config vals in the attributes...

freqtrade/pairlist/PrecisionFilter.py
xmatthias: yes i do remember - but that's the point of pairlists - you can use them in this sequence:
StaticPairlist - Filter1, Filter2, Volumepairlist (for sorting)

hroff-1902: How about this approach? (See the last commit: the pairlist is prepared only once, removing pairs with no ticker info if this is needed for filters)

TODO:
* couple of extra tests
* log removing pairs from the list with no ticker info available (??? is it needed?)

freqtrade/pairlist/pairlistmanager.py
```diff
@@ -94,3 +96,17 @@ def refresh_pairlist(self) -> None:
...
+    def _prepare_whitelist(self, pairlist: List[str], tickers) -> List[str]:
+        """
+        Prepare pairlist for Pairlist Filters that use tickers data - remove
+        pairs that do not have ticker available
+        """
+        if self._tickers_needed:
+            # Copy list since we're modifying this list
+            for p in deepcopy(pairlist):
+                ticker = tickers.get(p)
+                if not ticker:
```
xmatthias, line 109: this can be simplified as follows:

```suggestion
                if p not in tickers:
```

The only case where your approach would catch a problem would be the following:
```
{ 
'ETH/BTC': {},
'OTHER/PAIR': {'real': 123}
}
```

However, if that were the case, then ccxt would have messed up, and i'm yet to see such a return value..

freqtrade/pairlist/PrecisionFilter.py
xmatthias: Yeah, i would log it ... just so it's known that this pair was removed (only applies to static pairlist anyway ...)

hroff-1902:
"log removing pairs from the list with no ticker info available (??? is it needed?)"
-- logging in pairlistmanager (in the prepare_pairlist, when it removes pairs with no tickers) in the same way as in the PairLists (with TTL cached function) is impossible now because `pairlists` in the config does not have the `refresh_period` value, which can be used for caching this logging... I'll think how all this can be simplified and aligned, not in this PR.

Tests will be also added later, in a sep. PR.

freqtrade/pairlist/PrecisionFilter.py
hroff-1902: see the comment below on logging in pairlistmanager

PR #3317, opened 2020-05-16
Refactor informative pairs

### LINKED ISSUE #3240: Dynamic Informative pairs

hroff-1902:
(Only this part for a while)
* Code cleanup
* `create_pair_list()` moved from freqtradebot to dataprovider
* `_refresh_whitelist()` in `FreqtradeBot` renamed to `_refresh_active_whitelist()`
* `ListPairsWithTimeframes` type alias introduced
* new module `freqtrade.data.common` added for data-related common stuff, shared between exchange, dataprovider, rpc, freqtradebot, strategies and other parts. (for types, it can be renamed to `freqtrade.data.datatypes`)


hroff-1902:
I'm not sure here that the DataProvider class is the right place for create_pair_list().
In principle, it only needs ticker_interval, which can also be passed as a parameter, so it can be a function... And it's not intended to be called by the users from custom code... (well, maybe it can be also needed sometimes, to translate list of pairnames into the format of informative pairs tuples...)

freqtrade/data/dataprovider.py
```diff
@@ -36,13 +38,19 @@ def refresh(self,
...
         Should be whitelist + open trades.
         """
         return list(self._exchange._klines.keys())
 
+    def create_pair_list(self, pairs: List[str], timeframe: str = None) -> ListPairsWithTimeframes:
+        """
+        Create list of pair tuples with (pair, ticker_interval)
+        """
+        return [(pair, timeframe or self._config['ticker_interval']) for pair in pairs]
+
```
xmatthias, line 53: I would've moved this to pairlistmanager - it seems better suited there, as this does not provide any data ... (and a user will not want to use this, but will want to use `current_whitelist()`.

Having this available might be confusing to a user, who won't know if this or the other function is the correct one to ge the current pairlist ...

In the end, this is just a small helper function which combines pairs with timeframe ... but you need to have the list of pairs beforehand ...

hroff-1902: nice place for this function (and the datatype fits well into this module as well) 👍

hroff-1902:
wtf,
```
ImportError: cannot import name 'market_is_active'
```

where did it come from?

freqtrade/exchange/exchange.py
```diff
@@ -23,6 +23,7 @@
                                   OperationalException, TemporaryError)
 from freqtrade.exchange.common import BAD_EXCHANGES, retrier, retrier_async
 from freqtrade.misc import deep_merge_dicts, safe_value_fallback
+from freqtrade.pairlist.pairlistmanager import ListPairsWithTimeframes
```
xmatthias, line 26: This import causes a circular import (exchange needs `freqtrade.pairlist` - and `freqtrade.pairlist` needs exchange.

I guess if we intend to use these type-stubs more frequently, the best place might be `freqtrade.typing` ... so circular dependencies are not an issue.

hroff-1902: that's what I hate python for...

xmatthias: why?
every other language i ever used will kill you for circular dependencies too - so why hate python for it ... :D

hroff-1902: it does not report circular dependencies (even static checkers we use) and always print errors that are not relevant to the problem

in C, we always surrounded headers with
```
#ifndef HEADERFILE_H
#define HEADERFILE_H
```
"standard" construction and this was the perfect way to live with this and to make the headers really *idempotent*... At least, there was a recipe to live with this... no such a well-known recipe for python...

Issue #3009, opened 2020-03-02
What to do with partially sold pairs?


yazeed:
This is an issue to discuss what can be the best way to go about handling partially sold pairs, as I've noticed in the code there's a TODO part. Once established a consensus on an approach, a PR can be made as a follow-up.

`# TODO: figure out how to handle partially complete sell orders`

My opinion on this is that possibly there could be some mechanism that either:
- Triggers based price slipping away from partially sold items
- Triggers based on sell order timing out

Accordingly, it can proceed to do the following:
- _Cancel existing sell order on exchange, so that freqtrade can decide what to do with remaining amount according to stoploss/minimal_roi/trailing conditions in strategy._
- Modify trade in persistence to fulfilled amount, mark it as closed, and calculate profit according to fulfilled amount and timestamps, etc...
- Create a duplicate of the trade in persistence with the remaining amount.
- ~~Cancel existing sell order on exchange, so that freqtrade can decide what to do with remaining amount according to stoploss/minimal_roi/trailing conditions in strategy.~~

Would love to hear your thoughts.

xmatthias:
>Cancel existing sell order on exchange, so that freqtrade can decide what to do with remaining amount according to stoploss/minimal_roi/trailing conditions in strategy.

Well this seems pointless - the order did time out - so cancelling the order is "automatic" to me ... handling would happen after it's cancelled.

---

Assuming the order did time out and was only partially sold....
you'll need to consider a few cases - which makes handling of this very difficult.

Surprisingly, while we got this TODO comment in the code since years, I've not seen any issues related to half-filled sell orders... (as in - never)...

Either way, when taking this topic, we need to consider the following cases:

* little sold (almost full trade is still present)
* halfway sold (remaining balance > minimum tradable amount)
* mostly sold (remaining balance < minimum tradable amount)

I suspect the first 2 cases can be handled in a similar way... the 3rd one however is the difficult one.

If you create a new trade for the 3rd case, you'll run into "unable to sell" errors immediately.

Unfortunately, the amounts tied up in this can be a few USDT easily (binance uses 10USDT as minimum order size [source](https://www.binance.com/en/trade-rule)) - so a unfortunate order can remain stuck with 9.5USDT worth of Crypto currency - which it's not able to convert to USDT again (unless you buy another 10 USDT - then you can sell the 19.5 USDT worth of USDT easily).
[...]

PR #3320, opened 2020-05-17
Fix sell spamming

xmatthias:
## Summary
Partially filled trades should not spam the user with messages, but only send this once (and then have this status when running `/status`).

Solve part the issue (in the comments below): #3009 

## Quick changelog

- keep track of sell order status to be able to reiterate "partial cancel" when asked via RPC
- only send cancel messages once per trade (unless the message changes) to avoid spammy behaviour.


freqtrade/freqtradebot.py
```diff
@@ -881,7 +881,7 @@ def check_handle_timedout(self) -> None:
 
             fully_cancelled = self.update_trade_state(trade, order)
 
-            if (order['side'] == 'buy' and (
+            if (order['side'] == 'buy' and (order['status'] == 'open' or fully_cancelled) and (
```
hroff-1902, line 884: Note to myself, to remember:
that's (the whole if) is the only thing I need to verify closer (it's getting more and more complicated)
(and for the sell side below correspondingly)

The rest LGTM 👍

xmatthias: well don't tell me ...

Letme summarize what it does:

* check side (the rest is identical for both sides, but each with "side" different).
* short-circuit to false if the order is not open and NOT fully cancelled (this is the new part)

inside the or:
* check if it's fully cancelled
* verify timeout
* verify manual timeout call (that's VERY usefull to test timeout stuff, as you can just return "True" all the time) :laughing:

hroff-1902:
```
order['status'] == 'open' or fully_cancelled
```
here `fully_cancelled` is assigned as
```
fully_cancelled = self.update_trade_state(trade, order)
```
and `update_trade_state()` returns True if and only if
```
self.exchange.check_order_canceled_empty(order)
```
returns True, and `check_order_canceled_empty()` is calculated as
```
return order.get('status') in ('closed', 'canceled') and order.get('filled') == 0.0
```

So this part of the condition can be (I do not say it should be) reduced to:
```
(order['status'] == 'open') or (order['status'] in ('closed', 'canceled') and order.get('filled') == 0.0)
```

Just cannot understand for the while why these cases are handled here and the rest (what's the rest of cases? only open and partially filled? or all closed except closed/cancelled and unfilled?)

xmatthias: that's the problem - in theory all order-states can show up here (depending on the order type) (open, partially filled, even closed (that's what the change protects from)).

And the last "or" section must only be evaluated in certain conditions.

we had this as 2 nested IF's for clarity a while ago - but it doesn't really make sense as one IF can do the same thing.

Yes - the "filled_empty" could be done here too (instead of using the return value of `update_trade_state` - but since it's needed in `update_trade_state` too, that would be rather ... pointless.

I've renamed the variable to make that part clearer ...

hroff-1902: LGTM if it fixes this spamming issue 👍 
I'll review the logic here in more details later.

Some comments could be added here, explaining these hard for perception conditions...

PR #3322, opened 2020-05-17
Add ShuffleFilter

hroff-1902:
Adds Shuffling pairlist filter.
Replaces #2956 

Usage:
```
{"method": "ShuffleFilter"}
```
for reproducible results:
```
{"method": "ShuffleFilter", "seed": 42}
```

Also, some tests for other pairlists added/adjusted.

TODO: docs

### DIFF
--- a/freqtrade/pairlist/ShuffleFilter.py
+++ b/freqtrade/pairlist/ShuffleFilter.py
@@ -1,0 +1,50 @@
+"""
+Shuffle pair list filter
+"""
+import logging
+import random
+from typing import Dict, List
+
+from freqtrade.pairlist.IPairList import IPairList
+
+
+logger = logging.getLogger(__name__)
+
+
+class ShuffleFilter(IPairList):
+
+    def __init__(self, exchange, pairlistmanager, config, pairlistconfig: dict,
+                 pairlist_pos: int) -> None:
+        super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
+
+        self._seed = pairlistconfig.get('seed')
+        self._random = random.Random(self._seed)
+
+    @property
+    def needstickers(self) -> bool:
+        """
+        Boolean property defining if tickers are necessary.
+        If no Pairlist requries tickers, an empty List is passed
+        as tickers argument to filter_pairlist
+        """
+        return False
+
+    def short_desc(self) -> str:
+        """
+        Short whitelist method description - used for startup-messages
+        """
+        return (f"{self.name} - Shuffling pairs" +
+                (f", seed = {self._seed}." if self._seed is not None else "."))
+
+    def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
+        """
+        Filters and sorts pairlist and returns the whitelist again.
+        Called on each bot iteration - please use internal caching if necessary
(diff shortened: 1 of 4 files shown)


tests/pairlist/test_pairlist.py
```diff
@@ -219,12 +243,16 @@ def test_VolumePairList_whitelist_gen(mocker, whitelist_conf, shitcoinmarkets, t
     freqtrade.pairlists.refresh_pairlist()
     whitelist = freqtrade.pairlists.whitelist
 
-    assert whitelist == whitelist_result
+    if type(whitelist_result) is list:
```
xmatthias, line 246: Maybe add a comment here that this is needed to test shuffleFilter without seed ? 

I've had a different comment until i understood how this test works now.

Also - there is one thing i'd assert in addition here (this comes from my initial question in this line - if we have a filter/pairlist which cannot return a list ...).

```suggestion
    assert isinstance(whitelist_result, list)
    
    # Verify length of pairlist matches (used for ShuffleFilter without seed)
    if type(whitelist_result) is list:
```

hroff-1902:
right, docs or it does not exist

tests/pairlist/test_pairlist.py
hroff-1902: do you mean `assert isinstance(whitelist, list)` here? whitelist_result can now be int to test ShuffleFilter (and probably other filters in the future which could also return pairlist in diff. order)

xmatthias: but that's what it says above :D 

(sorry - yes i misstyped before ...)

xmatthias: Sorry - test-failure is most likely my bad - (as said before - i don't REALLY like the suggestion feature - as it makes such errors easy).

hroff-1902:
docs will be added on top of #3331

hroff-1902:
Pls review, the docs adjusted
(also fixed wrong name of the parameter used for SpreadFilter in the docs)

freqtrade/pairlist/ShuffleFilter.py
```diff
@@ -0,0 +1,50 @@
...
+import random
+from typing import Dict, List
+
+from freqtrade.pairlist.IPairList import IPairList
+
+
+logger = logging.getLogger(__name__)
+
+
+class ShuffleFilter(IPairList):
```
xmatthias, line 14: Although i lack ideas for a better name - isn't `ShuffleFilter` misleading? i mean - it doesn't filter ...?

hroff-1902: We already have *PairList and *Filter (which is also not absolutely necessary in my opinion), do you want to create the third entity, which won't add anything besides different suffix used in the name?

Filter is not necessary changes the number of the pairs in the pairlist -- it can also simply rearrange the pairs, sorting them in some order, by spread, by volume, etc. Or shuffling, like this one.

If you wish, the `number_assets` parameter can be added into this filter (and to all others as well, allowing each filter to limit the number of pairs in its resulting pairlist passed further in the chain)

xmatthias: leave as is 👍 
it was just something i noticed...

PR #3339, opened 2020-05-19
Cleanup in pairlistmanager

hroff-1902:
...and partly in IPairList.

Just to make it clear, where pairlists (instances if IPairList) and pairlists (lists of pairs) are used. 😉 
The term Pairlist Handler is used for instances of IPairList, as in the updated part of the documentation.

Neither the name of IPairList class, nor the name of the parameters of the methods of this class were touched at this attempt, to keep the PR light.

### DIFF
--- a/freqtrade/pairlist/pairlistmanager.py
+++ b/freqtrade/pairlist/pairlistmanager.py
@@ -23,24 +23,25 @@ class PairListManager():
         self._config = config
         self._whitelist = self._config['exchange'].get('pair_whitelist')
         self._blacklist = self._config['exchange'].get('pair_blacklist', [])
-        self._pairlists: List[IPairList] = []
+        self._pairlist_handlers: List[IPairList] = []
         self._tickers_needed = False
-        for pl in self._config.get('pairlists', None):
-            if 'method' not in pl:
-                logger.warning(f"No method in {pl}")
+        for pairlist_handler_config in self._config.get('pairlists', None):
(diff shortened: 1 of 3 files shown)


freqtrade/pairlist/pairlistmanager.py
```diff
@@ -23,24 +23,25 @@ def __init__(self, exchange, config: dict) -> None:
         self._config = config
         self._whitelist = self._config['exchange'].get('pair_whitelist')
         self._blacklist = self._config['exchange'].get('pair_blacklist', [])
-        self._pairlists: List[IPairList] = []
+        self._pairlist_handlers: List[IPairList] = []
```
xmatthias, line 26: I am not sure i agree with this renaming ...

to me, a pairlist-handler and pairlist-manager is the same (but in other words) ... so now why do i have multiple pairlist handlers?

hroff-1902: Handlers handle pairlist in the chain, processing (handling) each the pairlist in its manner -- i.e. are those of IPairList type.

The pairlist manager (which is in the same time is the pairlist handler manager) loads handlers (creates the chain) and starts processing the pairlist via the chain of pairlist handlers.

That's how it's used in the docs, approved and merged yesterday.

hroff-1902: You can find a better name, sure.

PR #3343, opened 2020-05-19
Refactor verify_blacklist()

hroff-1902:
...to be it a method of pairlistmanager, which simplifies the usage (it has access to the blacklist stored in the attribute of pairlistmanager).


freqtrade/pairlist/pairlistmanager.py
```diff
@@ -110,6 +110,22 @@ def _prepare_whitelist(self, pairlist: List[str], tickers) -> List[str]:
 
         return pairlist
 
+    def verify_blacklist(self, pairlist: List[str], logmethod) -> List[str]:
```
xmatthias, line 113: I gues the correct type for logmethod would be `Callable[..., Any]`

note: this is just an informative comment

PR #3344, opened 2020-05-20
CORS - allow authenticated responses

xmatthias:
## Summary

needed on top of #3315 to support authenticated requests across domains

### DIFF
--- a/freqtrade/rpc/api_server.py
+++ b/freqtrade/rpc/api_server.py
@@ -90,7 +90,7 @@ class ApiServer(RPC):
         self._config = freqtrade.config
         self.app = Flask(__name__)
         self._cors = CORS(self.app,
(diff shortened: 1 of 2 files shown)


woutervddn:
I believe this is breaking CORS for an unauthenticated request to ping.

Our setup checks the connection to the bot from a webbased UI whenever you try to add a new bot. It is done this way to ensure that we don't send user credentials to the wrong server that might not belong to us.

I believe ping should be able to be reached without being authenticated

xmatthias:
Ping itself is reachable, but not from a javascript browser context.

This is intentional, but cors handling has been enhanced since, so you can enable a domain to allow this for your bot explicitly, from the domain / source youre using.
Please checkt the latest documentation for this.

PR #3353, opened 2020-05-21
More tests for pairlists

hroff-1902:
* Test with empty list in pairlists fail: according to the docs, StaticPairList should be used
* ShuffleFilter does not filter out pairs that are incompatible with exchange (or does not receive the pairlist which was already filtered with incompatible pairs removed)

### DIFF
--- a/tests/pairlist/test_pairlist.py
+++ b/tests/pairlist/test_pairlist.py
@@ -19,7 +19,8 @@ def whitelist_conf(default_conf):
         'TKN/BTC',
         'TRST/BTC',
         'SWT/BTC',
-        'BCC/BTC'
+        'BCC/BTC',
+        'HOT/BTC'
     ]
     default_conf['exchange']['pair_blacklist'] = [
         'BLK/BTC'
@@ -202,6 +203,9 @@ def test_VolumePairList_refresh_empty(mocker, markets_empty, whitelist_conf):
 
 
 @pytest.mark.parametrize("pairlists,base_currency,whitelist_result", [
+    # Empty pairlist
+    ([],
+     "BTC", ['ETH/BTC', 'TKN/BTC', 'LTC/BTC', 'XRP/BTC', 'HOT/BTC']),
     # VolumePairList only
     ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "quoteVolume"}],
      "BTC", ['ETH/BTC', 'TKN/BTC', 'LTC/BTC', 'XRP/BTC', 'HOT/BTC']),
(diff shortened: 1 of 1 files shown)


hroff-1902:
@xmatthias pls have a look at the first part -- what should be expected behavior here?

also, when a Filter is used without StaticPairList or VolumePairList at the first position -- what shall be expected behavior? Shall this configuration fail, or use static whitelist (as if StaticPairList implicitly added before), or use the whole set of pairs (as for dynamic pairlists)?

hroff-1902:
I see the fix here as follows: `IPairList._whitelist_for_active_markets()` should be called from `pairlistmanager._prepare_whitelist()` (or near the call to `pairlistmanager._prepare_whitelist()`, just before or just after, in `pairlistmanager.refresh_pairlist()`), just need to be sure what is expected proper behaviour here...

hroff-1902:
Or, may be better, the OperationalException should be raised if first Pairlist Handler is not Static or Volume...

tests/pairlist/test_pairlist.py
```diff
@@ -202,6 +203,9 @@ def test_VolumePairList_refresh_empty(mocker, markets_empty, whitelist_conf):
 
 
 @pytest.mark.parametrize("pairlists,base_currency,whitelist_result", [
+    # Empty pairlist
+    ([],
```
xmatthias, line 207: You'll never get this far if you initialize the bot correctly (not from within a test, skipping most of the initialization part) - as we do add StaticPairList if the list is empty.

So for me, this is (at least at the moment) a pointless test - as no user can realistically reach this point.

xmatthias:
For the second test - maybe adding a property `isFilter` or `isPairlist` making the Pairlist define if it's just a filter, or if it's providing a new list of pairs eventually should help.

tests/pairlist/test_pairlist.py
hroff-1902: right, I'll remove this test case

hroff-1902:
> For the second test - maybe adding a property `isFilter` or `isPairlist` making the Pairlist define if it's just a filter, or if it's providing a new list of pairs eventually should help.
> I'm not a fan of hardcoding `is StaticPairlist or Volumepairlist` - as in theory, new pairlists can be added at any point in time and that would lead to a longer and longer list of checks in the end.

I know how to add the check without hardcoding `is StaticPairlist or Volumepairlist` or similar, just confirmed what should be the proper behavior in these cases -- it should fail with OperationalException, if a non-pairlist is in the first position.

xmatthias:
I think so, yes - any other filter would yield nothing - as it's just filtering input -> output.

I've not been trying to point out how to implement it - i know very well that you know how to implement this. I've myself thinking about "how would I do it" - and then typing that out while thinking ... :)

hroff-1902:
> I think so, yes - any other filter would yield nothing - as it's just filtering input -> output.
> 
> I've not been trying to point out how to implement it - i know very well that you know how to implement this. I've myself thinking about "how would I do it" - and then typing that out while thinking ... :)

The phrase above does not sound as I wanted it to be, my bad, it's not "shut up I know this", it should be "I've already spent time thinking on it and know how to make it, so do not pls spend your time on it", I guess I needed more coffee when I wrote that...

Anyway, closing this in favor to #3358, since the latter includes same test cases properly adjusted.

PR #3358, opened 2020-05-22
Split the pairlist generation logic and filtering

hroff-1902:
Split the pairlist generation logic and filtering.

Also addresses #3353 and supersedes it (same tests. with exemption of the empty `pairlists`, added)

TODO: Need to find a better place for a call to `self._prepare_whitelist()`. there is definitely no need to duplicate this logic in every filter, but this will be refactored later, to keep this PR small. Now I wonder why the test for ShuffleFilter after StaticPairList does not fail, so Don't merge for a while.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -130,7 +130,7 @@ This should contain the name of the Pairlist Handler, as well as a short descrip
 
 Override this method if the Pairlist Handler can be used as the leading Pairlist Handler in the chain, defining the initial pairlist which is then handled by all Pairlist Handlers in the chain. Examples are `StaticPairList` and `VolumePairList`.
 
-This is called with each iteration of the bot - so consider implementing caching for compute/network heavy calculations.
+This is called with each iteration of the bot (only if the Pairlist Handler is at the first location) - so consider implementing caching for compute/network heavy calculations.
 
 It must return the resulting pairlist (which may then be passed into the chain of Pairlist Handlers).
 
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -92,13 +92,13 @@ docker-compose exec freqtrade_develop /bin/bash
 You have a great idea for a new pair selection algorithm you would like to try out? Great.
 Hopefully you also want to contribute this back upstream.
 
-Whatever your motivations are - This should get you off the ground in trying to develop a new Pairlist provider.
+Whatever your motivations are - This should get you off the ground in trying to develop a new Pairlist Handler.
 
-First of all, have a look at the [VolumePairList](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/pairlist/VolumePairList.py) provider, and best copy this file with a name of your new Pairlist Provider.
+First of all, have a look at the [VolumePairList](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/pairlist/VolumePairList.py) Handler, and best copy this file with a name of your new Pairlist Handler.
 
-This is a simple provider, which however serves as a good example on how to start developing.
+This is a simple Handler, which however serves as a good example on how to start developing.
 
-Next, modify the classname of the provider (ideally align this with the Filename).
+Next, modify the classname of the Handler (ideally align this with the module filename).
 
 The base-class provides an instance of the exchange (`self._exchange`) the pairlist manager (`self._pairlistmanager`), as well as the main configuration (`self._config`), the pairlist dedicated configuration (`self._pairlistconfig`) and the absolute position within the list of pairlists.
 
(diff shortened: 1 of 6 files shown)


freqtrade/pairlist/IPairList.py
```diff
@@ -90,6 +91,16 @@ def _validate_pair(self, ticker) -> bool:
         """
         raise NotImplementedError()
 
+    def gen_pairlist(self, cached_pairlist: List[str], tickers: Dict) -> List[str]:
+        """
+        Generate the pairlist
+        :param cached_pairlist: Previously generated pairlist (cached)
+        :param tickers: Tickers (from exchange.get_tickers()).
+        :return: List of pairs
+        """
```
xmatthias, line 100: I think this docstring should mention that
* this method is only called for the first ~pairlist~ handler in the list
* it shall not be implemented for filters

Also - i would document these changes to `developer.md` - it's describing the (current) implementation of a new pairlist quite well  (we should do that for other things too) - but we need to be carefull that it's kept uptodate (at least for the most part).

hroff-1902: > this method is only called for the first pairlist handler in the list

hah! should it be called if StaticPairList is in the middle of the chain, rewriting all previous results with the contents of the whitelist from the config? (that's current (previous) behavior)

Ideally, this usage should raise the exception, but I do not know yet how to achieve it with this (proposed here) implementation...

xmatthias: just check `pairlist_pos` in the constructor of the ~pairlist~ pairhandler - if it must be first (which is the case for staticPairlist) - then raise if it's not the first ... ? 

This way, control is given to the handler... :)

hroff-1902: Yeah, but my intention was to get rid of this position completely in the future...

xmatthias: i wouldn't do this - also for future Pairlists i can think of it being useful to know if you're the first or last pairlist ...

hroff-1902: well, it's not what I was going to make in the nearest future anyway, so let's see if this attribute is useful...

so, the question is what should be the expected behavior for StaticPairList when it's placed somewhere in the middle of the chain: rewrite the pairlist prepared by the previous handlers in the chain (with a call to gen_pairlist()), do nothing (with a call to filter_pairlist(), which simply returns pairlist, prepared by the previous handlers in the chain) or raise OperationalException, noting that this Pairlist Handler should only be used at the first position in the chain)?

xmatthias: Exception - StaticPairlist has no other functionality - and overwriting everything others did previously would be ... highly unexpected (but that's how it's currently i think).

hroff-1902:
adjusting development.md left...

freqtrade/pairlist/StaticPairList.py
```diff
@@ -38,4 +48,8 @@ def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
         :param tickers: Tickers (from exchange.get_tickers()). May be cached.
         :return: new whitelist
         """
-        return self._whitelist_for_active_markets(self._config['exchange']['pair_whitelist'])
+        if self._pairlist_pos != 0:
+            raise OperationalException(f"{self.name} can only be used in the first position "
+                                       "in the list of Pairlist Handlers.")
```
xmatthias, line 53: Shouldn't this be in the constructor? 

i guess- i'd expect it to fail when PairlistManager is created - not when i try to get a pairlist ...

docs/developer.md
```diff
@@ -114,28 +114,44 @@ Now, let's step through the methods which require actions:
...
 Returns a description used for Telegram messages.
-This should contain the name of the Provider, as well as a short description containing the number of assets. Please follow the format `"PairlistName - top/bottom X pairs"`.
+
+This should contain the name of the Pairlist Handler, as well as a short description containing the number of assets. Please follow the format `"PairlistName - top/bottom X pairs"`.
+
+#### gen_pairlist
+
+Override this method if the Pairlist Handler can be used as the leading Pairlist Handler in the chain, defining the initial pairlist which is then handled by all Pairlist Handlers in the chain. Examples are `StaticPairList` and `VolumePairList`.
+
+This is called with each iteration of the bot - so consider implementing caching for compute/network heavy calculations.
```
xmatthias, line 133:
```suggestion
This is called with each iteration of the bot (only if the Pairlist Handler is at the first location) - so consider implementing caching for compute/network heavy calculations.
```

PR #3359, opened 2020-05-22
Fix typing circular dependency

xmatthias:
## Summary
As reported on slack (and anoyed myself just now)

we can't use a file called typing if we import from typing within it.

I've moved this to `constants.py` now ... in the end, that's what it is - a constant.

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -3,6 +3,9 @@
 """
 bot constants
 """
+from typing import List, Tuple
+
+
 DEFAULT_CONFIG = 'config.json'
 DEFAULT_EXCHANGE = 'bittrex'
 PROCESS_THROTTLE_SECS = 5  # sec
@@ -329,3 +332,6 @@ CANCEL_REASON = {
     "ALL_CANCELLED": "cancelled (all unfilled and partially filled open orders cancelled)",
     "CANCELLED_ON_EXCHANGE": "cancelled on exchange",
 }
+
+# List of pairs with their timeframes
+ListPairsWithTimeframes = List[Tuple[str, str]]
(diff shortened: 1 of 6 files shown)


hroff-1902:
is it pycharm-specific?

xmatthias:
no, it depends on how you run freqtrade.

When you run it with `freqtrade trade` - it works just fine.

However, most debuggers use `python freqtrade/main.py` (so you have a clear entrypoint i guess) - in which case you'll get this error.

Even though we don't document this method to run it - having a module collide with a stdlib module is very bad practice.

hroff-1902:
okay, thanks for explanation,

I do not like constants.py be overloaded with different stuff... Another possibility could be making the freqtrade/common dir and splitting current constant.py into modules in that dir (then typing could be imported as freqtrade.common.typing, think it will help in this case), but this can be done later separately...

xmatthias:
No - the file should not be called typing. call it fttypes or fttyping ... but if it's typing, we're always risking of breaking stuff (just think about interactively importing "something" in a shell ...).

In general, no file sould collide with a standardlib module / package (i don't understand why flake (or mypy) does not bark at that ... ).

hroff-1902:
okay, this will be next thing I'll hate python for, if this behavior (ignoring import paths for modules) is not caused by some settings in the pycharm project, but rather is the common behavior for python interpreters... that's very odd

hroff-1902:
I think the behavior here is that when the bot is called as freqtrade/main.py (in freqtrade dir), and it handles `import typing`, it searches first that freqtrade dir and finds typing.py in it; so moving it to freqtrade/common/typing.py should help without the need to rename typing.py module to something else...

xmatthias:
your assumption is that nothing EVER imports `freqtrade.common.xxx` directly - which will cause this again.

It's a simple naming conflict - which you can see in many other languages too.

I don't see why we must name this as typing.py ...

PR #3377, opened 2020-05-25
Refactor BTReport

xmatthias:
## Summary
First step in refactoring backtest report.

The final result will generate one large dict containing all metrics calculated and shown during backtesting - which can also be stored alongside with the backtest results.

Nothing in the output does change - just the way how backtest-methods are called.

## Quick changelog

- seperate metric generation from metric visualization (currently still artificially, but will be moved out of this function in the next PR)
- use dicts instead of lists of lists - to make it clear that list[5] contains profit_pct - and not something random.
- Add a bit more detail to certain tests.

### DIFF
--- a/freqtrade/optimize/optimize_reports.py
+++ b/freqtrade/optimize/optimize_reports.py
@@ -34,6 +34,13 @@ def store_backtest_result(recordfilename: Path, all_results: Dict[str, DataFrame
             file_dump_json(filename, records)
 
 
+def _get_line_floatfmt() -> List[str]:
+    """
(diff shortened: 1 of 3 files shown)


freqtrade/optimize/optimize_reports.py
```diff
@@ -1,7 +1,7 @@
 import logging
 from datetime import timedelta
 from pathlib import Path
-from typing import Dict
+from typing import Dict, List, Any
```
hroff-1902, line 4:
```suggestion
from typing import Any, Dict, List
```

PR #3380, opened 2020-05-26
Raspberry Docker image - fix and correctly crosscompile

xmatthias:
## Summary
Our current raspberry images did use "in-image" crosscompilation - which means they could not be be fully built on a raspberry should someone have that desire.

Docker now offers a new, experimental build system (docker buildx) which allows easy crosscompilation.
Building this new image relies on that. The image itself can however be fully built on a Raspberry PI.

Note: It uses python3.7 - since that allows the usage of [piwheels.org](piwheels.org), which provide wheels for the most common packages and speeds up compilation (i interrupted the compilation of numpy in the CI system after ~2 hours of compilation time).

Note: the `Dockerfile.armhf` is a copy of `Dockerfile` - with the addition of the piwheels configuration, and removal of hyperopt dependencies.

Examples for a full CI run can be found [here](https://github.com/freqtrade/freqtrade/runs/709868931?check_suite_focus=true)

Examples of a CI run using caching is [here](https://github.com/freqtrade/freqtrade/runs/710254294?check_suite_focus=true). I aborted it after seeing that the first few buildsteps included "Cached".
(open deploy -  "Build Raspberry docker image" - and scroll to ~line 25).

```
#6 [ 2/10] RUN apt-get update   && apt-get -y install curl build-essential ...
#6 CACHED

#7 [ 3/10] RUN mkdir /freqtrade
#7 CACHED
```

## Quick changelog

- Use full arm image
- reduce notification frequency in slack "notification" channel
  - it'll be one notification for regular builds - and explicit notifications for failures.
- Implement an action that kills previous builds
  - this is important when quickly merging multiple branches - as otherwise the CI runs contend for ressources, and an older release may end up as "latest build" on dockerhub / pypi (not really - as we only push master to pypi).


xmatthias:
it does work locally too - but needs some docker setup first (buildx is not enabled by default ...).

PR #3390, opened 2020-05-29
improve /profit to not raise an exception if no trade is closed

xmatthias:
## Summary
/Profit should not return 502 just because no trade has been closed.
Instead, it should return the profit for open trades (or no profit, if no trade exists).

## Quick changelog

- Improve telegram output for just started bots (when calling /profit)
- improve output for API (no 502 - but results instead)

## What's new?
No change to /profit when closed trades exist

shows "No trades yet." if no trade was yet opened.

2020-05-29_09h44_14
2020-05-29_09h44_24

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -328,7 +328,7 @@ class Telegram(RPC):
         best_rate = stats['best_rate']
         # Message to display
         if stats['closed_trade_count'] > 0:
-            markdown_msg = ("*ROI:* Close trades\n"
+            markdown_msg = ("*ROI:* Closed trades\n"
                             f"∙ `{profit_closed_coin:.8f} {stake_cur} "
                             f"({profit_closed_percent:.2f}%)`\n"
                             f"∙ `{profit_closed_fiat:.3f} {fiat_disp_cur}`\n")
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -326,24 +326,27 @@ class Telegram(RPC):
         avg_duration = stats['avg_duration']
         best_pair = stats['best_pair']
         best_rate = stats['best_rate']
-        # Message to display
-        if stats['closed_trade_count'] > 0:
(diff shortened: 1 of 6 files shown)


freqtrade/rpc/telegram.py
```diff
@@ -311,38 +311,40 @@ def _profit(self, update: Update, context: CallbackContext) -> None:
...
+        profit_all_fiat = stats['profit_all_fiat']
+        trade_count = stats['trade_count']
+        first_trade_date = stats['first_trade_date']
+        latest_trade_date = stats['latest_trade_date']
+        avg_duration = stats['avg_duration']
+        best_pair = stats['best_pair']
+        best_rate = stats['best_rate']
+        # Message to display
+        if stats['closed_trade_count'] > 0:
+            markdown_msg = ("*ROI:* Close trades\n"
```
hroff-1902, line 331:
```suggestion
            markdown_msg = ("*ROI:* Closed trades\n"
```

hroff-1902:
Maybe it's better to write in the case of no trades at all:
"No, open trades. No closed trades."
or
"No, open trades."
?

xmatthias:
> Maybe it's better to write in the case of no trades at all:
> "No, open trades. No closed trades."
> or
> "No, open trades."
> ?

every option with a comma sounds very much off to me.

also - i didn't want to completely change this message - just remove the unnecessary exception as it's giving problems for the API.

freqtrade/rpc/telegram.py
xmatthias: bu can't commit from here - would break tests :D

hroff-1902:
> > Maybe it's better to write in the case of no trades at all:
> > "No, open trades. No closed trades."
> > or
> > "No, open trades."
> > ?
> 
> every option with a comma sounds very much off to me.
> 
> also - i didn't want to completely change this message - just remove the unnecessary exception as it's giving problems for the API.

ah, read above as "No open trades. No closed trades." or "No open trades.", I do not know where that comma came from.
Actually, "No trades yet" is also okay...

PR #3394, opened 2020-05-29
Allow changing severity of strategy-validations to log only.

### LINKED ISSUE #3211: allow disabling strategy validation
**For building accurate renko or other candle types its best to have main dataframe updated with the given parameters.**

### Enhancement Proposal
introduce a parameter, like internals.check_dataframe = False (True by default) to switch off dataframe checks and specify in the docs that it shall only be used when the user understands what he is doing.

xmatthias:
## Summary
Allow disabling of the strategy validations.

It'll just print them as log instead of an exception. This will ensure that eventual support log-submissions do include this - so people trying to recreate it can help eventually.

closes #3211

## Quick changelog

- Add `disable_dataframe_checks` setting (in both strategy and config) - but most likely this will be set in the strategy, as it's usually VERY strategy related.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -136,6 +136,7 @@ Values set in the configuration file always overwrite values set in the strategy
 * `stake_currency`
 * `stake_amount`
 * `unfilledtimeout`
+* `disable_dataframe_checks`
 * `use_sell_signal` (ask_strategy)
 * `sell_profit_only` (ask_strategy)
 * `ignore_roi_if_buy_signal` (ask_strategy)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -107,7 +107,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `db_url` | Declares database URL to use. NOTE: This defaults to `sqlite:///tradesv3.dryrun.sqlite` if `dry_run` is `true`, and to `sqlite:///tradesv3.sqlite` for production instances. <br> **Datatype:** String, SQLAlchemy connect string
 | `initial_state` | Defines the initial application state. More information below. <br>*Defaults to `stopped`.* <br> **Datatype:** Enum, either `stopped` or `running`
 | `forcebuy_enable` | Enables the RPC Commands to force a buy. More information below. <br> **Datatype:** Boolean
-| `disable_dataframe_checks` | Disable checking the dataframe for correctness. Only use when intentionally changing the dataframe. [Strategy Override](#parameters-in-the-strategy).[Strategy Override](#parameters-in-the-strategy).<br> *Defaults to `False`*. <br> **Datatype:** Boolean
+| `disable_dataframe_checks` | Disable checking the OHLCV dataframe returned from the strategy methods for correctness. Only use when intentionally changing the dataframe and understand what you are doing. [Strategy Override](#parameters-in-the-strategy).[Strategy Override](#parameters-in-the-strategy).<br> *Defaults to `False`*. <br> **Datatype:** Boolean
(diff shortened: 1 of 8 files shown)


hroff-1902:
I have an additional config setting (added into "internals") which switches off checks for the dataframe returned by the strategy... In my (new) stashes... Is it better? Should those two approaches be combined (it's true/false now, but can be 'error' (default) / 'warning' / 'debug', for instance

xmatthias:
> I have an additional config setting (added into "internals") which switches off checks for the dataframe returned by the strategy... In my (new) stashes... Is it better? Should those two approaches be combined? (it's true/false now, but can be 'error' (default) / 'warning' / 'debug', for instance

It's something i did think about - but i'm not certain it's a good idea (from a support point of view).

Think about a support case - where we get logs that don't contain any hint on this (just the running log basically, not the startup log) ... we'd have little possibility to discover this (assuming the user set it and does not remember / does not tell us).

hroff-1902:
okay, so no setting value which completely switches off veriification in this case: only true/false, with 2 cases -- fail with error or proceed with warning/info message in the log.

I can include it here if you wish.

docs/configuration.md
```diff
@@ -107,6 +107,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `db_url` | Declares database URL to use. NOTE: This defaults to `sqlite:///tradesv3.dryrun.sqlite` if `dry_run` is `true`, and to `sqlite:///tradesv3.sqlite` for production instances. <br> **Datatype:** String, SQLAlchemy connect string
 | `initial_state` | Defines the initial application state. More information below. <br>*Defaults to `stopped`.* <br> **Datatype:** Enum, either `stopped` or `running`
 | `forcebuy_enable` | Enables the RPC Commands to force a buy. More information below. <br> **Datatype:** Boolean
+| `disable_dataframe_checks` | Disable checking the dataframe for correctness. Only use when intentionally changing the dataframe. [Strategy Override](#parameters-in-the-strategy).[Strategy Override](#parameters-in-the-strategy).<br> *Defaults to `False`*. <br> **Datatype:** Boolean
```
hroff-1902, line 110:
```suggestion
| `disable_dataframe_checks` | Disable checking the OHLCV dataframe returned from the strategy methods for correctness. Only use when intentionally changing the dataframe and understand what you are doing. [Strategy Override](#parameters-in-the-strategy).[Strategy Override](#parameters-in-the-strategy).<br> *Defaults to `False`*. <br> **Datatype:** Boolean
```

hroff-1902:
should not the templates/sample strategies be also changed to include this attribite?
(do not think so, since it's assumes "do it when you understand what you are doing", but let's discuss it explicitly)

xmatthias:
I'm sorry - your stashes are unknown to me :D 

> should not the templates/sample strategies be also changed to include this attribite?
(do not think so, since it's assumes "do it when you understand what you are doing", but let's discuss it explicitly)

I'm not sure - i've added it to the "full" config sample ... but i guess adding it to the strategy samples would not hurt.

hroff-1902:
Actually, I'm not sure this should be in the samples/template...

xmatthias:
sure thing - `git revert` to the rescue :D

hroff-1902:
Merged, but I won't announce it at Slack widely, since it's a "understand what you are doing" setting.

hroff-1902:
Wait,
is it in `internals`, as code in strategy_resolver assume?

xmatthias:
No it's not - i thought about it and decided against it (nothing related to strategy is in internals - so why should this be there.

It simply does not fit there.

aak-dev:
great!

PR #3408, opened 2020-05-31
Apiserver logging

xmatthias:
## Summary
Allow reducing log verbosity for api-server explicitly.
This will allow to hide the "GET /api/v1/..." calls by setting verbosity within the api_server config section to error.

I think this is important for the UI - as that tends to call the API quite often - clogging up the log with rest messages (and therefore potentially hiding other important messages).

This also addes the api_server to all default configuration templates - so it's easier for people to enable this later on.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -103,6 +103,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `api_server.enabled` | Enable usage of API Server. See the [API Server documentation](rest-api.md) for more details. <br> **Datatype:** Boolean
 | `api_server.listen_ip_address` | Bind IP address. See the [API Server documentation](rest-api.md) for more details. <br> **Datatype:** IPv4
 | `api_server.listen_port` | Bind Port. See the [API Server documentation](rest-api.md) for more details. <br>**Datatype:** Integer between 1024 and 65535
+| `api_server.verbosity` | Loggging verbosity. `info` will print all RPC Calls, while "error" will only display errors. <br>**Datatype:** Enum, either `info` or `error`. Defaults to `info`.
 | `api_server.username` | Username for API server. See the [API Server documentation](rest-api.md) for more details. <br>**Keep it in secret, do not disclose publicly.**<br> **Datatype:** String
 | `api_server.password` | Password for API server. See the [API Server documentation](rest-api.md) for more details. <br>**Keep it in secret, do not disclose publicly.**<br> **Datatype:** String
 | `db_url` | Declares database URL to use. NOTE: This defaults to `sqlite:///tradesv3.dryrun.sqlite` if `dry_run` is `true`, and to `sqlite:///tradesv3.sqlite` for production instances. <br> **Datatype:** String, SQLAlchemy connect string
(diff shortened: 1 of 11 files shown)


docs/configuration.md
```diff
@@ -103,6 +103,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `api_server.enabled` | Enable usage of API Server. See the [API Server documentation](rest-api.md) for more details. <br> **Datatype:** Boolean
 | `api_server.listen_ip_address` | Bind IP address. See the [API Server documentation](rest-api.md) for more details. <br> **Datatype:** IPv4
 | `api_server.listen_port` | Bind Port. See the [API Server documentation](rest-api.md) for more details. <br>**Datatype:** Integer between 1024 and 65535
+| `api_server.verbosity` | Loggging verbosity. `info` will print all RPC Calls, while "error" will only display errors. <br>**Datatype:** Enum, either `info` or `error`. Defaults to `info`.
```
hroff-1902, line 106:
```suggestion
| `api_server.verbosity` | Logging verbosity. `info` will print all RPC Calls, while "error" will only display errors. <br>**Datatype:** Enum, either `info` or `error`. Defaults to `info`.
```

2 'g's is enough I guess 😆

docs/configuration.md
```diff
@@ -103,6 +103,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `api_server.enabled` | Enable usage of API Server. See the [API Server documentation](rest-api.md) for more details. <br> **Datatype:** Boolean
 | `api_server.listen_ip_address` | Bind IP address. See the [API Server documentation](rest-api.md) for more details. <br> **Datatype:** IPv4
 | `api_server.listen_port` | Bind Port. See the [API Server documentation](rest-api.md) for more details. <br>**Datatype:** Integer between 1024 and 65535
+| `api_server.verbosity` | Logging verbosity. `info` will print all RPC Calls, while "error" will only display errors. <br>**Datatype:** Enum, either `info` or `error`. Defaults to `info`.
```
hroff-1902, line 106:
```suggestion
| `api_server.verbosity` | Logging verbosity. `info` will print all RPC Calls, while "error" will only display errors. <br>**Datatype:** String, either `info` or `error`. Defaults to `info`.
```

hroff-1902:
`api_server.verbosity` set as 'error'|'info' is okay for me, but I have some doubts about it since our global `verbosity` is int and this should be aligned somehow in the future...

docs/configuration.md
xmatthias: I've intentionally used Enum here as we use that for `initial_state` too...

xmatthias:
well there's a difference - global verbosity can be increased by running `-v`, `-vv`, and `-vvv` ...

For the api however, it's the "access log" logging you also see from nginx or apache...
since the UI does quite some requests, this does clog up the log with unnecessary requests - making it almost impossible to see "relevant" log.

This is also intended to be set by the user in the config (you set it while setting the password) - therefore using "speaking" names makes more sense in my opinion.

docs/configuration.md
hroff-1902: hmm, okay

hroff-1902:
as I said a bit above, that's okay for me, the question was to fix that it was done intentionally and that we'll probably be needing to align it in the future (or maybe will not)

PR #3418, opened 2020-06-01
Test colorama init again (after the fixes done to progressbar)

xmatthias:
## Summary
Reintroduce colorama_init() to fix problems with displaying colors under windows.

## Quick changelog

- reintroduce colorama_init
- Remove dependency on os.path (it's only been used once).

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -4,26 +4,26 @@
 This module contains the hyperopt logic
 """
 
+import io
 import locale
 import logging
 import random
 import warnings
-from math import ceil
 from collections import OrderedDict
+from math import ceil
 from operator import itemgetter
 from pathlib import Path
 from pprint import pprint
 from typing import Any, Dict, List, Optional
 
+import progressbar
 import rapidjson
+import tabulate
 from colorama import Fore, Style
+from colorama import init as colorama_init
 from joblib import (Parallel, cpu_count, delayed, dump, load,
                     wrap_non_picklable_objects)
-from pandas import DataFrame, json_normalize, isna
-import progressbar
-import tabulate
-from os import path
-import io
+from pandas import DataFrame, isna, json_normalize
 
 from freqtrade.data.converter import trim_dataframe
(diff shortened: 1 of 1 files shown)


hroff-1902:
colors are not displayed in conda window

xmatthias:
the following observatins:
### powershell (blue terminal)
* with this option: Does not show color, but also no strange signs
* without this option: does not show color, but shows "strange" characters (the colorization characters).


### [new windows terminal](https://github.com/microsoft/terminal)

* with this option: does not show color
* without this option: works as a linux terminal - showing color

### anaconda prompt:
* with this option: Does not show color, but also no strange signs (aligned to powershell)
* without this option: does not show color, but shows "strange" characters (the colorization characters) - making the result barely readable

hroff-1902:
okay,

the best option in my opinion:
* introduce `--no-progressbar` option for the hyperopt subcommand;
* require either `--no-color` or `--no-progressbar` used in the command line on Windows, with a descriptive message;

xmatthias:
how about a documentation change to the hyperopt page only?

``` md
On windows, please use `--no-color` if you see strange signs in the output - or use the [new windows terminal](https://github.com/microsoft/terminal) for correct output on windows.
```

xmatthias:
well, on windows this seems kindof aligned now (with this PR).
no terminal shows color - but also no strange signs.

To be honest, i'm content with this option, if you want color output under windows, use WSL.
Color output in windows terminals is utterly broken (by microsoft) - which can easily be understood once you start looking through the colorama "windows-special" code.


In the end, this means that color is disabled implicitly for hyperopt - but i don't see this as an issue - i don't mind not supporting color output on windows, as the output itself is readable and as expected.


Making the progressbar optional might end up to be rather difficult - as we can't use the context manager anymore - also, i don't think it's worth the effort (keeping in mind the goal - to make us - and therefore the users - profit and an easy to use interface).

The color-output is nice - but nothing more, and not worth investing hours over hours just to get it working on a system where freqtrade is also very difficult to install (due to ta-lib).

PR #3426, opened 2020-06-01
Remove deprecated setting

xmatthias:
## Summary
Remove deprecated edge-config `capital_available_percentage` and raise an exception instead.

Part of #2971 (although i'm not sure what would be needed to completely suffice that issue)

## Quick changelog

- remove `capital_available_percentage` and raise error when it's used in combination with edge
- remove some occurances of percent where we mean ratio in the documentation


hroff-1902:
LGTM 👍 

I'm not sure should we announce this change as a News in Slack -- this setting was marked as Deprecated in the docs before...

xmatthias:
Nah, its been gone since january ...

PR #3430, opened 2020-06-02
ticker_interval -> timeframe

### LINKED ISSUE #2318: Use the term 'timeframe' instead of non-standard 'ticker interval'
This issue proposes the use of the term 'timeframe' instead of 'ticker interval' everywhere in the docs, settings and cli options.

'Timeframe' is the "standard" term, used everywhere in the trading literature. Ccxt also uses this term.

The existing '--ticker-interval' option should get the `--timeframe` alias and be declared as deprecated (and then removed in the next versions).

Can be (should be) a part of #2264, so that all these changes into the cli were implemented in the upcoming release at once.

xmatthias:
## Summary
Replace all occurances of ticker_interval with timeframe (while keeping compatibility).

Throughout the bot, only timeframe is used now (also in the docs).

The only occurances of ticker_interval is the "compat code" - which notifies about the deprecation of using ticker_interval - and moves ticker_interval to timeframe.

This compatiblity code will most likely remain there for a VERY long time - as we'll first need to do a final release with this - then update the strategy-repo - and then users should update (but who reads the deprecation messages, anyway) - so i expect these sections to be there for quite a long time.

closes #2318

## Quick changelog

- Replace all occurances
- api methods return both values for a short amount of time (to allow applications to switch to the new property) - the existing value will then be removed.
- Implement methods to copy `ticker_interval` to `timeframe` (and viceversa) so strategies that use this don't break 
- database will be updated by renaming ticker_interval to timeframe

## Additional review notes:
Main focus should be on the compatibility code:
moving the attribute has to be done twice (since both the old and new code can be in either strategy, or configuration) - and is therefore in `StrategyResolver` - as well as in the `deprecated_settings.py`.


hroff-1902:
Oh, you made it!...
👍

hroff-1902:
merged very strange...

hroff-1902:
^^ pls verify what I did, it merged quite strange, and after fixing the tests there are still some appearances of 'ticker_interval' in both tests and freqtrade dirs

xmatthias:
Yeah, the last commit is wrong.

I've intentionally left both ticker_interval and timeframe in the api responses to create a cutover period (otherwise all applications/ui's MUST switch at that precise moment). By sending both ticker_interval and timeframe applications (including freqUI) can gradually switch to the new keys.

Also - we don't want to break users code - so we must keep ticker_interval + timeframe along each other (at least for some time) in both strategy and hyperopt files.

freqtrade/commands/pairlist_commands.py
```diff
@@ -25,7 +25,7 @@ def start_test_pairlist(args: Dict[str, Any]) -> None:
     results = {}
     for curr in quote_currencies:
         config['stake_currency'] = curr
-        # Do not use ticker_interval set in the config
+        # Do not use timeframe set in the config
```
hroff-1902, line 28: this comment is wrong here, should be removed

freqtrade/configuration/configuration.py
```diff
@@ -242,8 +242,8 @@ def _process_optimize_options(self, config: Dict[str, Any]) -> None:
         self._args_to_config(config, argname='strategy_list',
                              logstring='Using strategy list of {} strategies', logfun=len)
 
-        self._args_to_config(config, argname='ticker_interval',
-                             logstring='Overriding ticker interval with Command line argument')
+        self._args_to_config(config, argname='timeframe',
+                             logstring='Overriding timeframe with Command line argument')
```
hroff-1902, line 246:
```suggestion
                             logstring='Overriding timeframe with command line argument')
```

freqtrade/optimize/hyperopt_interface.py
```diff
@@ -31,13 +31,15 @@ class IHyperOpt(ABC):
...
-    ticker_interval: str
+    ticker_interval: str  # deprecated
+    timeframe: str
 
     def __init__(self, config: dict) -> None:
         self.config = config
 
         # Assign ticker_interval to be used in hyperopt
-        IHyperOpt.ticker_interval = str(config['ticker_interval'])
+        IHyperOpt.ticker_interval = str(config['timeframe'])  # DEPRECTED
```
hroff-1902, line 41:
```suggestion
        IHyperOpt.ticker_interval = str(config['timeframe'])  # DEPRECATED
```

freqtrade/resolvers/strategy_resolver.py
```diff
@@ -50,11 +50,19 @@ def load_strategy(config: Dict[str, Any] = None) -> IStrategy:
         if 'ask_strategy' not in config:
             config['ask_strategy'] = {}
 
+        if hasattr(strategy, 'ticker_interval') and not hasattr(strategy, 'timeframe'):
+            # Assign ticker_interval to timeframe to keep compatibility
+            if 'timeframe' not in config:
+                logger.warning(
+                    "DEPRECATED: Please migrate to using timeframe instead of ticker_interval."
```
hroff-1902, line 57:
```suggestion
                    "DEPRECATED: Please migrate to using 'timeframe' instead of 'ticker_interval' strategy attribute."
```

Issue #3404, opened 2020-05-30
Exception when downloading Kraken trades


hroff-1902:
## Describe your environment

  * Operating system: Ubuntu
  * Python Version: 3,7.7 (`python -V`)
  * CCXT version: 1.28.49 (`pip freeze | grep ccxt`)
  * Freqtrade Version: develop (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
  
## Describe the problem:

Have `download-data --dl-trades` running for Kraken, with many retries, finally got exception:
```
2020-05-30 19:43:45,991 - freqtrade.exchange.exchange - DEBUG - Fetching trades for pair ZEC/EUR, params: {'since': '1590856217466086643'} 
2020-05-30 19:43:49,020 - freqtrade.exchange.exchange - DEBUG - Stopping because from_id did not change. Reached 1590856953714 > 1590856750421
2020-05-30 19:43:49,057 - freqtrade.misc - INFO - dumping json to "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/user_data/data/kraken/ZEC_EUR-trades.json.gz"
2020-05-30 19:43:49,840 - freqtrade.misc - DEBUG - done json to "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/user_data/data/kraken/ZEC_EUR-trades.json.gz"
2020-05-30 19:43:49,841 - freqtrade.data.history.history_utils - DEBUG - New Start: 2020-02-20T03:00:04
2020-05-30 19:43:49,841 - freqtrade.data.history.history_utils - DEBUG - New End: 2020-05-30T19:42:33
2020-05-30 19:43:49,841 - freqtrade.data.history.history_utils - INFO - New Amount of trades: 68966
2020-05-30 19:43:49,868 - freqtrade - ERROR - Fatal exception!
Traceback (most recent call last):
  File "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/freqtrade/main.py", line 36, in main
    return_code = args['func'](args)
  File "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/freqtrade/commands/data_commands.py", line 61, in start_download_data
    data_format_trades=config['dataformat_trades'],
  File "/home/user/freqtrade-wrk/github-hroff-1902/freqtrade/freqtrade/data/history/history_utils.py", line 352, in convert_trades_to_ohlcv
[...]

PR #3432, opened 2020-06-02
Fix crash in #3404

hroff-1902:
Fixes crashing `data-download --dl-trades` on empty trades data, which is reported in #3404


hroff-1902:
yes, and it should not cause crash

PR #3437, opened 2020-06-03
Display cumulative profit in /profit calls too

xmatthias:
## Summary
Display cumulative profit in telegram.
Rest-API get's the data now as both percentage and ratio - and both sum and mean.

The "old" key remains there with the "#DEPRECATED" flag - which means it'll be removed after the next release.

## Quick changelog

- show Profit sum

## What's new?

Now don't judge me by this Screenshot :D it's a test bot that simulates constant trades - and apparently buying on even minutes, and selling on odd minutes is not very profitable :D 

2020-06-03-194248_389x266_scrot


hroff-1902:
Maybe
`-26.87878787 USDT (Avg: -0.30%, Total: -130.7%)`
or
`-26.87878787 USDT (Avg: -0.30%) (Total: -130.7%)`
?
can also be 
`-26.87878787 USDT (-0.30% Avg) (-130.7% Total)`

xmatthias:
That will wrap on the phone ... 🙄

PR #3438, opened 2020-06-03
Ftx stoploss

### LINKED ISSUE #3103: Fee calculation + stoploss support + trade amounts on FTX

xmatthias:
## Summary
Introduce stoploss on exchange for FTX.

closes #3103 (only stoploss is remaining).

## Quick changelog

- Add methods to support stoploss on exchange
- FTX supports both stoploss-limit and stoploss-market (controlled via `order_types.stoploss` key).
- use `cancel_stoploss_order` and `get_stoploss_order` to determine which type of order we're querying. This defaults to `cancel_order` and `get_order` by default, but can be overwritten by exchange subclasses (as does FTX).

### DIFF
--- a/freqtrade/exchange/ftx.py
+++ b/freqtrade/exchange/ftx.py
@@ -97,9 +97,9 @@ class Ftx(Exchange):
             raise OperationalException(e) from e
 
     @retrier
-    def cancel_stoploss_order(self, order_id: str, pair: str) -> None:
+    def cancel_stoploss_order(self, order_id: str, pair: str) -> Dict:
         if self._config['dry_run']:
-            return
+            return {}
         try:
             return self._api.cancel_order(order_id, pair, params={'type': 'stop'})
         except ccxt.InvalidOrder as e:
--- a/freqtrade/exchange/ftx.py
+++ b/freqtrade/exchange/ftx.py
@@ -7,6 +7,7 @@ import ccxt
 from freqtrade.exceptions import (DependencyException, InvalidOrderException,
                                   OperationalException, TemporaryError)
 from freqtrade.exchange import Exchange
+from freqtrade.exchange.common import retrier
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


hroff-1902:
What I do not like here, is the defferent names for types of the stoplosses for different exchanges (`stop-loss-limit`, `stop-loss-market` and `stop`, `stop-market`). Should not it be unified, with `limit` and `market` only?

xmatthias:
> What I do not like here, is the defferent names for types of the stoplosses for different exchanges (`stop-loss-limit`, `stop-loss-market` and `stop`, `stop-market`). Should not it be unified, with `limit` and `market` (or `stoploss-limit` and `stoploss-market`, name them as you wish) only?

Feel free to get in touch with the exchange supports to agree with their counterparts on one common standard for this.

Until then - we need to use what the exchange gives us.

hroff-1902:
I mean, cannot we unify it in the app: translate `"stoploss": "market"` to `stop-loss-market` for Kraken and `stop-market` for FTX, and `"stoploss": "limit"` to `stop-loss-limit` for Binance and `stop` for FTX?

xmatthias:
i am not sure what you'd like to unify.

This is not anything the user will ever see.
The configuration is only possible for FTX (both kraken and binance only support one stoploss type) - and for FTX, the configuration is done via order_types stoploss property - which can be either limit or market.

Unifying this would mean to translate it from one thing to another thing - which would be FAR worse than how it is now - where it's occuring in 2 places - in the stoploss creation function, and in persistence - to properly close / update the trade.

hroff-1902:
Assume we add more exchanges to the list of supported, with custom exchange classes that implement stoplosses on exchange. For instance, assume we add NewExchange1 and NewExchange2.

Those exchanges all have custom type strings for market and limit stoplosses: 'new-exchange1-stoploss-limit', 'new-exchange1-stoploss-market', 'new-exchange2-stop', 'new-exchange2-stop-limit', whatever.

With the current approach will be needing to add all those type names into the documentation.

I propose to unify it, and use "standard" names for limit stoploss and market stoploss, which are to be mapped for actual type strings for the exchange. So, unified 'market' and 'limit' (or 'stoploss-market', 'stoploss-limit', whatever) types can be used as the value of order_types.stoploss.

xmatthias:
> With the current approach will be needing to add all those type names into the documentation.

which documentation? You misread (or didn't bother to look) WHERE these exchange-specific strings are used.

NO user will EVER touch these (unless he's modifying freqtrade internal code). 
They're used ONLY in the persistence class, and in the exchange class when creating the order (in the create_stoploss_order method).


For users (configuration-wise) they are already unified, the user will insert `'order_types': {...'stoploss': 'market'}` or `'order_types': {...'stoploss': 'limit'}`.

This only has an effect on FTX, as all other supported exchanges (binance, kraken) only support one stoploss type, which is automatically chosen ignoring the stoploss configuration setting.

hroff-1902:
you are right,
I was confused by the following in the docs, I guess:

> Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.

and decided you implemented order_types.stoploss in the way I described above (exchange-specific values, I don't know why I decided it)... 🤔

freqtrade/exchange/ftx.py
```diff
@@ -2,13 +2,117 @@
...
+                raise InvalidOrderException(
+                    f'Tried to get an invalid dry-run-order (id: {order_id}). Message: {e}') from e
+        try:
+            orders = self._api.fetch_orders(pair, None, params={'type': 'stop'})
+
+            order = [order for order in orders if order['id'] == order_id]
+            if len(order) == 1:
+                return order[0]
+            else:
+                raise InvalidOrderException(f"Could not get Stoploss Order for id {order_id}")
```
hroff-1902, line 94:
```suggestion
                raise InvalidOrderException(f"Could not get stoploss order for id {order_id}")
```

freqtrade/exchange/ftx.py
```diff
@@ -2,13 +2,117 @@
...
+            amount = self.amount_to_precision(pair, amount)
+
+            order = self._api.create_order(symbol=pair, type=ordertype, side='sell',
+                                           amount=amount, price=stop_price, params=params)
+            logger.info('stoploss order added for %s. '
+                        'stop price: %s.', pair, stop_price)
+            return order
+        except ccxt.InsufficientFunds as e:
+            raise DependencyException(
+                f'Insufficient funds to create {ordertype} sell order on market {pair}.'
```
hroff-1902, line 63:
```suggestion
                f'Insufficient funds to create {ordertype} sell order on market {pair}. '
```

xmatthias:
well we should still be clear what type is used for which exchange - not everyone will know which order-type a stoploss "his/her" exchange supports.

PR #3449, opened 2020-06-05
Ask strategy verbosity

xmatthias:
## Summary
Reduce verbosity of sell-rate finding (as reported on slack a few days ago)

Instead, allow /show_conf to visualize ask and bid strategy.

## Quick changelog

- New keys ask_strategy and buy_strategy in show_conf returns.

for telegram (since this is a "maintenance" command, i don't think it will need better formatting - as that would make the output VERY verbose).

2020-06-05-203110_393x507_scrot


freqtrade/rpc/telegram.py
```diff
@@ -636,13 +636,15 @@ def _show_config(self, update: Update, context: CallbackContext) -> None:
...
             sl_info = f"*Stoploss:* `{val['stoploss']}`\n"
-
+        import json
         self._send_msg(
             f"*Mode:* `{'Dry-run' if val['dry_run'] else 'Live'}`\n"
             f"*Exchange:* `{val['exchange']}`\n"
             f"*Stake per trade:* `{val['stake_amount']} {val['stake_currency']}`\n"
             f"*Max open Trades:* `{val['max_open_trades']}`\n"
             f"*Minimum ROI:* `{val['minimal_roi']}`\n"
+            f"*Ask strategy:* ```\n{val['ask_strategy']}```\n"
```
hroff-1902, line 646: `json.dumps` here as well?

xmatthias: I'm not sure -  you're right, it must be aligned between the two sides - but if you look at the screenshot - the only difference is the quotes (and False or false) ...

so explicit json conversation might not be necessary ... ?

hroff-1902:
I think json conversion should better be used for both ask and bid -- it's not only about surrounding quotes, but also about how booleans are printed (True/False or true/false) and how the keys are defined (in single quotes or in double quotes). I think it's better to print it as json for a user -- in the same way she defined it in the config,

PR #3461, opened 2020-06-08
change hyperopt output to print ready to copy to strategy

felpasl:
# Summary

format the roi, in order to facilitate the copy, in the return of the hyperoptmize

changed to python output to easy use in strategy file without mistakes

# Quick changelog

- New Format in hyperopt output
Actual return 
``` Buy hyperspace params:
{   'bb_width-value': 22.07556,
    'bbwidth-enabled': True,
    'direction': 'sma-enabled',
    'ema-value': 65,
    'rsi-enabled': True,
    'rsi-value': 38,
    'sma-value': 76,
    'trigger': 'bb_lower3',
    'width': 'bb_width1'}
Sell hyperspace params:
{   'sell-rsi-enabled': True,
    'sell-rsi-value': 98,
    'sell-trigger': 'sell-bb_lower1'}
ROI table:
{0: 0.16532, 14: 0.05331, 31: 0.02645, 57: 0}
``` 
changed to 
``` python
    # Buy hyperspace params:
    buy_params = {
        'bb_width-value': 4.66539,
        'bbwidth-enabled': True,
        'direction': 'ema-enabled',
        'ema-value': 32,
        'rsi-enabled': False,
        'rsi-value': 31,
        'sma-value': 27,
        'trigger': 'bb_lowerband2',
        'width': 'bb_width3'
    }

    # Sell hyperspace params:
    sell_params = {
        'sell-rsi-enabled': False,
        'sell-rsi-value': 42,
        'sell-trigger': 'bb_lowerband1'
    }

    # ROI table:
    minimal_roi = {
        "0": 0.23458,
        "28": 0.03915,
        "57": 0.02861,
        "169": 0
    }

    # Stoploss:
    stoploss = -0.08436
```

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -12,7 +12,7 @@ from math import ceil
 from collections import OrderedDict
 from operator import itemgetter
 from pathlib import Path
-from pprint import pprint
+from pprint import pformat
 from typing import Any, Dict, List, Optional
 
 import rapidjson
@@ -244,11 +244,21 @@ class Hyperopt:
     def _params_pretty_print(params, space: str, header: str) -> None:
         if space in params:
             space_params = Hyperopt._space_params(params, space, 5)
+            print(f"\n    # {header}")
             if space == 'stoploss':
-                print(header, space_params.get('stoploss'))
+                print("    stoploss =", space_params.get('stoploss'))
+            elif space == 'roi':
+                minimal_roi_result = rapidjson.dumps(
+                    OrderedDict(
+                        (str(k), v) for k, v in space_params.items()
+                    ),
+                    default=str, indent=4, number_mode=rapidjson.NM_NATIVE)
+                minimal_roi_result = minimal_roi_result.replace("\n", "\n    ")
+                print(f"    minimal_roi = {minimal_roi_result}")
             else:
-                print(header)
-                pprint(space_params, indent=4)
+                params_result = pformat(space_params, indent=4).replace("}", "\n}")
+                params_result = params_result.replace("{", "{\n ").replace("\n", "\n    ")
+                print(f"    {space}_params = {params_result}")
 
     @staticmethod
     def _space_params(params, space: str, r: int = None) -> Dict:
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -230,6 +230,9 @@ class Hyperopt:
             if space in ['buy', 'sell']:
                 result_dict.setdefault('params', {}).update(space_params)
             elif space == 'roi':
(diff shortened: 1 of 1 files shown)


hroff-1902:
Overall looks good to me, some suggestions:

* The boolean values should be printed in python-format, i.e. as `True`/`False` i.o. `true`/`false`
* Consider renaming `buy_param` --> `buy_params`, same for sell part -- because these are "params".

hroff-1902:
^^ in short, if you print the output as pythonic code, ready to be inserted into a python strategy, the output should not be printed as json, but as python values instead...

xmatthias:
i guess for buy / sell sides that was the case before (python dicts, not json) - although overall formatting was ... not really good.

for roi i like this improvement a lot - makes it very easy to read (a lot easier than before, anyway) :+1:

hroff-1902:
then it should be pretty-printed with proper formatting.
to be ~ready~ for copy-pasting to the strategy, the values should be printed as python values, not as json

xmatthias:
@hroff-1902 I don't understand the references to python3.6 here - as they are not relevant (nor correct).


a) the sequence of the ROI dict is not relevant for hyperopt output, as it can be specified in any sequence in both the strategy or configuration, and strategyresolver will order it upon loading the strategy.

b) python3.6 does/did keep the dictionary order (assuming cpython) - however it was not explicit, but was a side-effect of the way it was implemented. 3.7 made this "intentional / documented" - but also 3.6 dicts keep their insertion order. [source](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation). It also states that this may change in future versions, but now we know this was not the case, so i think it's safe to rely on it (especially since the only thing that may break is the output).

hroff-1902:
> @hroff-1902 I don't understand the references to python3.6 here - as they are not relevant (nor correct).
> 
> a) the sequence of the ROI dict is not relevant for hyperopt output, as it can be specified in any sequence in both the strategy or configuration, and strategyresolver will order it upon loading the strategy.
> 
> b) python3.6 does/did keep the dictionary order (assuming cpython) - however it was not explicit, but was a side-effect of the way it was implemented. 3.7 made this "intentional / documented" - but also 3.6 dicts keep their insertion order. [source](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation). It also states that this may change in future versions, but now we know this was not the case, so i think it's safe to rely on it (especially since the only thing that may break is the output).

With python 3.7+, the usage of rapidjson (and conversion of roi structure to OrderedDict) will not be necessary, pprint/pformat will be enough. Moreover, those if's for spaces can also be removed from the code and one simple pprint/pformat will suffice.

hroff-1902:
I'll refactor this further if I won't forget -- actually, we do not need OrderedDict there (but this is definitely out of scope of this PR...)

@felpasl thanks again

PR #3468, opened 2020-06-10
Cleanup async markets

hroff-1902:
First part of market handling cleanup (split up to make the PRs small and clean):

* Remove async markets -- they are not used in the bot, neither required by other async ccxt methods we use. And it's not async fetching of market info -- it's same with `_load_market()`, but for async instance of the ccxt, should be equal to sync market data and also fetched synchronously...
* Rename `_reload_markets()` to `reload_markets()`, to make it public (used in freqtradebot.py)
* Change level for "Markets were not loaded. Loading them now.." message. It's a normal situation during markets data initialization, there is nothing wrong to warn the user about with a warning message.

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -256,21 +256,32 @@ class Exchange:
                                      "Please check your config.json")
                 raise OperationalException(f'Exchange {name} does not provide a sandbox api')
 
+    def _load_async_markets(self, reload: bool = False) -> None:
+        try:
+            if self._api_async:
+                asyncio.get_event_loop().run_until_complete(
+                    self._api_async.load_markets(reload=reload))
+
+        except ccxt.BaseError as e:
+            logger.warning('Could not load async markets. Reason: %s', e)
+            return
+
     def _load_markets(self) -> None:
-        """ Initialize markets """
+        """ Initialize markets both sync and async """
         try:
             self._api.load_markets()
+            self._load_async_markets()
             self._last_markets_refresh = arrow.utcnow().timestamp
         except ccxt.BaseError as e:
             logger.warning('Unable to initialize markets. Reason: %s', e)
 
     def reload_markets(self) -> None:
-        """Reload markets if refresh interval has passed """
+        """Reload markets both sync and async if refresh interval has passed """
         # Check whether markets have to be reloaded
         if (self._last_markets_refresh > 0) and (
                 self._last_markets_refresh + self.markets_refresh_interval
                 > arrow.utcnow().timestamp):
-            return
+            return None
         logger.debug("Performing scheduled market reload..")
         try:
             self._api.load_markets(reload=True)
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -190,7 +190,7 @@ class Exchange:
     def markets(self) -> Dict:
         """exchange ccxt markets"""
         if not self._api.markets:
-            logger.warning("Markets were not loaded. Loading them now..")
+            logger.info("Markets were not loaded. Loading them now..")
             self._load_markets()
         return self._api.markets
 
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -264,8 +264,8 @@ class Exchange:
         except ccxt.BaseError as e:
             logger.warning('Unable to initialize markets. Reason: %s', e)
 
-    def _reload_markets(self) -> None:
-        """Reload markets if refresh interval has passed"""
+    def reload_markets(self) -> None:
+        """Reload markets if refresh interval has passed """
         # Check whether markets have to be reloaded
         if (self._last_markets_refresh > 0) and (
                 self._last_markets_refresh + self.markets_refresh_interval
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -256,32 +256,21 @@ class Exchange:
                                      "Please check your config.json")
                 raise OperationalException(f'Exchange {name} does not provide a sandbox api')
 
-    def _load_async_markets(self, reload: bool = False) -> None:
-        try:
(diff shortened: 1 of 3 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -256,32 +256,21 @@ def set_sandbox(self, api: ccxt.Exchange, exchange_config: dict, name: str) -> N
...
-    def _load_async_markets(self, reload: bool = False) -> None:
-        try:
-            if self._api_async:
-                asyncio.get_event_loop().run_until_complete(
-                    self._api_async.load_markets(reload=reload))
-
-        except ccxt.BaseError as e:
-            logger.warning('Could not load async markets. Reason: %s', e)
-            return
-
```
xmatthias, line 268: Removing this is wrong and can introduce unwanted side-effects. 

While we don't explicitly need this - ccxt will call this when doing anything using the async object  and it's not been explicitly called before.


So by not calling this at the start, we delay the first future operation. 
I'd rather have a slightly slower startup (i gues that's what you're trying to address here?) - and that's why this is there (might merit a comment explaining this ...)

freqtrade/exchange/exchange.py
```diff
@@ -190,7 +190,7 @@ def timeframes(self) -> List[str]:
     def markets(self) -> Dict:
         """exchange ccxt markets"""
         if not self._api.markets:
-            logger.warning("Markets were not loaded. Loading them now..")
+            logger.info("Markets were not loaded. Loading them now..")
```
xmatthias, line 193: :+1:

freqtrade/exchange/exchange.py
```diff
@@ -256,32 +256,21 @@ def set_sandbox(self, api: ccxt.Exchange, exchange_config: dict, name: str) -> N
...
         try:
             self._api.load_markets()
-            self._load_async_markets()
             self._last_markets_refresh = arrow.utcnow().timestamp
         except ccxt.BaseError as e:
             logger.warning('Unable to initialize markets. Reason: %s', e)
 
-    def _reload_markets(self) -> None:
-        """Reload markets both sync and async, if refresh interval has passed"""
+    def reload_markets(self) -> None:
```
xmatthias, line 267: :+1:

freqtrade/exchange/exchange.py
hroff-1902: Well, as I see from the ccxt code (and this maybe found in the docs) -- it's a kind of convention in ccxt, for exchange methods to call `load_markets()` first if the method needs markets or market data for the certain market. And `load_market()` does absolutely same we do -- it fetches and processes markets data if it's yet empty and caches the previously fetched result if no `reload=True` was passed. And that's proper behavior -- to get all exchange methods idempotent, to call them in any order. We do not need to worry about this -- initialization should be properly handled and cached if needed in the intrinsics of ccxt. We also do not initialize currencies info and tickers, for example -- but you could also say "ccxt will call this when doing something", right? At least, "something ... ccxt is doing using the async object" here what we call for the ccxt async object and it's fetching ohlcv, and the fetch_ohlcv/fetch_ohlcvs methods (the default one, and the method from the particular exchange implementation) do always call `load_markets()` inside. Otherwise the exchange implementation will not pass the @kroitor verification, I guess...

If you think that some exchanges may not follow this, let's ask ccxt, should we initialize markets to fetch ohlcvs from the instance -- and if the answer will be "yes", my next question would be "what else should we initialize explicitly before fetching ohlcv?"... And if the next answer will be "only markets", my next question would be "why do we need to initialize markets and do not need to initialize other attributes -- currencies, tickers, whatever"...

I'm not addressing actually here slower startup (but this also would speedup the startup, yes) -- I'm removing strange dependencies, to make the code simplier and cleaner...

hroff-1902: ^^ In short, in my opinion: it's not absolutely needed to call `load_markets` for async ccxt object if we do not use explicitly the `ccxt.market` attribute and assume it's initialized, if we only call the ccxt methods, that should initialize the markets inside, if they need it.

xmatthias: well, i know that `load_markets()` is called within the ccxt methods - and i don't think there's differences in ccxt exchange implementations for this (let's assume all exchanges do this correctly).

It would still be a problem for freqtrade to not do this explicitly:

* we startup
* run the first loop to fetch data (this will be slower - as `load_markets()` will be called from within ccxt.
* ...

Now let's assume the worst case - kraken with a slightly older ccxt version (or uptodate version, but kraken (or any other exchange) changing something to break load_markets).
You would not notice anything during startup, but fetching candles will fail.

but not because fetch_ohlcv returns an error - but because load_markets() does not work correctly.

Now we introduced an implicit problem which needs to be handled in every exchange method - just because we don't want to call this explicitly.
Now the code did not become cleaner - but at lot more convoluted.


----

You're right - it's not absolutely needed as ccxt does this implicitly if it's not been called before.

To put it the way [the ccxt documentation puts it](https://github.com/ccxt/ccxt/wiki/Manual#loading-markets):

> If you forget to load markets the ccxt library will do that automatically upon your first call to the unified API. It will send two HTTP requests, first for markets and then the second one for other data, sequentially.

This implicitly states that you "forgot" to call it - which means that it's best practice to explicitly call this before.

hroff-1902: Okay, agree,

thanks for the link -- I was too lazy to find this part of the ccxt docs, sorry.

xmatthias: however, if you don't mind - put a docstring to this method explaining this ... so we avoid this discussion in 3 months (or after a bottle of whisky / vodka for you) again :laughing:

hroff-1902: yeah, that's exactly what I was going to do with it, returning it back...

hroff-1902:
Reverted for now (to keep this PR simple), pls re-review

Issue #2076, opened 2019-07-31
Insufficient funds to create stop_loss_limit sell order


yilongwen:
## Step 1: Have you search for this issue before posting it?

If you have discovered a bug in the bot, please [search our issue tracker](https://github.com/freqtrade/freqtrade/issues?q=is%3Aissue). 
If it hasn't been reported, please create a new issue.

## Step 2: Describe your environment

  * Python Version: 3.7.3
  * CCXT version: 1.18.738
  * Branch: Develop
  * Last Commit ID: f0206a90b1504c40e74b1c6bae4c719779cdee85
  
## Step 3: Describe the problem:

The bot is not able to create either limit sell order or stoploss limit sell order on regular basis (but not always...). I use unlimited stake amount and max 1 trade, 'full or canceled' for buy orders.

### Steps to reproduce:

  1. _____
  2. _____
  3. _____
  
### Observed Results:

  * The bot finds a different amount to sell than the actual amount.
  * I expected the bot to sync the actual amount before selling.

### Relevant code exceptions or logs:

  ```
  2019-07-31 21:51:30,394 - freqtrade.freqtradebot - WARNING - Amount 57.638000000000005 does not match 57.638
  2019-07-31 21:51:30,394 - freqtrade.freqtradebot - WARNING - Could not update trade amount: Half bought? Amounts don't match
  2019-07-31 21:51:39,712 - freqtrade.freqtradebot - WARNING - Unable to place a stoploss order on exchange: Insufficient funds to create stop_loss_limit sell order on market NEO/USDT. Tried to sell amount 57.638 at rate 6.351 (total 366.058938). Message: binance Account has insufficient balance for requested action.
  ```

xmatthias:
This will require some more information:
* what was the amount (and value) of that pair on the exchange (you should also be able to see this in the trade history for this pair)
* what pair was involved (to help give the numbers more sense)

yilongwen:
- According to trade history: 

2019-07-31 23:51:29 | NEO/USDT | Buy | 11.663 | 25.559 | 0.02555900 NEO | 298.09461700 USDT
----------------------------------------------------------------------------------------------
2019-07-31 23:51:29 | NEO/USDT | Buy | 11.663 | 32.079 | 0.03207900 NEO | 374.13737700 USDT

[...]

PR #3478, opened 2020-06-13
Minor: Exchange cosmetics

hroff-1902:
* Exception message fixed (should also be done for FTX in/after #3438)
* Minor fix in a usecase
* Added validation for `price <= stop_price` (actually, `0.0 < stoploss_on_exchange_limit_ratio < 1.0`, which is equivalent) at the startup time. Now exception is raised surprisingly at the creation of first stoploss, which can happen in days.

For the last item, to be discussed:
* I left the duplicating check in exchange/binance.py, it can be removed, but this check makes the stoploss() method more error-resistant (maybe)
* The stoploss_on_exchange_limit_ratio can only be validated at startup when stoploss_on_exchange = True or even only for Binance (and for FTX after #3438) exchange only, i.e. when it's actually used.

Inspired by the new comment arrived: https://github.com/freqtrade/freqtrade/issues/2076#issuecomment-643663739


freqtrade/exchange/exchange.py
```diff
@@ -386,6 +386,13 @@ def validate_ordertypes(self, order_types: Dict) -> None:
                 f'On exchange stoploss is not supported for {self.name}.'
             )
 
+        # Limit price threshold: As limit price should always be below stop-price
+        # Used for limit stoplosses on exchange
+        limit_price_pct = order_types.get('stoploss_on_exchange_limit_ratio', 0.99)
+        if limit_price_pct >= 1.0 or limit_price_pct <= 0.0:
+            raise OperationalException(
+                "stoploss_on_exchange_limit_ratio should be < 1.0 and > 0.0")
+
```
xmatthias, line 395: I don't think this is exchange specific - so mabye adding it to the validation in constants.py would suffice? (you can add min / max values there ...)

xmatthias: with this addition to constants, the following validation comes up:

```
                'stoploss_on_exchange_limit_ratio': {'type': 'number', 'minimum': 0.0,
                                                     'maximum': 1.0}
```

```
2020-07-13 07:14:22,436 - freqtrade.configuration.config_validation - INFO - Validating configuration ...
2020-07-13 07:14:22,437 - freqtrade.configuration.config_validation - CRITICAL - Invalid configuration. See config.json.example. Reason: -1.0 is less than the minimum of 0.0

Failed validating 'minimum' in schema['properties']['order_types']['properties']['stoploss_on_exchange_limit_ratio']:
    {'maximum': 1.0, 'minimum': 0.0, 'type': 'number'}

On instance['order_types']['stoploss_on_exchange_limit_ratio']:
    -1.0
2020-07-13 07:14:22,439 - freqtrade.commands.trade_commands - ERROR - -1.0 is less than the minimum of 0.0

```

This is followed by another, "not so nice" stacktrace - but i'd rather take that in a seperate PR.

@hroff-1902 if you agree, i'll commit this to this branch and remove the above, explicit test (as that cannot be hit anymore ...).

hroff-1902: yes, please, let's complete it...

I remember I was wanted to revert some code in the commits above, and cannot recall in the meantime how I wanted to complete this PR a month ago, following your comments above.

xmatthias: for sure - let me know what you think - it's the last (2) commits which remove this test in place of testing it as part of the configuration.

hroff-1902:
thanks,
LGTM 👍

PR #3487, opened 2020-06-16
Fix documentation typo

xmatthias:
## Summary
The correct argument is --datadir, not `--data-dir`.


xmatthias:
> maybe we should add more aliases for cli options like this one?

don't think so - the more aliases we got, the more confusing it will be ...

PR #3497, opened 2020-06-18
Analyze dataframe and keep it until the next analysis

xmatthias:
## Summary
This will revise how freqtrade analyzes pairs.
Current behaviour is that freqtrade only analyzes pairs if it's possible to buy / sell.

After this PR, all pairs in the whitelist will be analyzed at the start of the loop, and further code (like get_signal() will load this cached dataframe (including all indicators).

Now the main reason to make this change is supporting freqUI - a subsequent PR (should be ready by the time this is merged) will add an API endpoint to get full dataframes, for realtime plotting of the indicators (as calculated by the strategy).

#### Detailed reasoning

Current behaviour is that freqtrade only analyzes pairs if it's possible to buy / sell.

This means, without sell-signal enabled (relying on stoploss / roi only) - no open trade will ever have it's candles analyzed, nor will any pair be analyzed if all trade-slots (`max_open_trades`) are filled.
This creates a problem for certain users(reported on slack a while ago), as they'd like to track how many buys were skipped / missed, or do some other things which require the populate* methods to be ran every time, which wil be (implicitly) fixed by this PR.

To avoid issues where users only want to do something when a buy / sell is "really" due, i'm introducing the 2 new callbacks, `confirm_trade_entry()` and `confirm_trade_exit()` - which will allow cancelling buy / sells right before they're placed at the exchange.
Alternative uses can use this for additional logging / tracking...

Now to help with "pair-independent" things (like download CCI30 as reference) - this will also introduce the callback `bot_loop_start()` - which is called once every loop - before anything else (after downloading data, before analyzing pairs).

The new sequence is documented in `bot-basics.md` - so i suggest to start there to understand the flow.

## Quick changelog

* introduce `analyze()` - which runs at the start of the loop and analyzes all tradable pairs (pairs in whitelist + open trades)
* store analyzed data in dataprovider - have it accessible via `get_analyzed_dataframe()`.
* don't analyze in `get_signal()` - this should only return the get_signal
* introduce `/pair_history` API endpoint to have realtime access to the analyzed DataFrame
* introduce `bot_loop_start()` in the strategy interface - allowing to perform pair-agnostic tasks once per iteration
* introduce `confirm_trade_entry()` in the strategy interface - allowing to cancel a buy at the last moment before placing the order
* introduce `confirm_trade_exit()` in the strategy interface - allowing to cancel a sell order the last moment before placing the order
* Add "bot basics" section to the docs, explaining the basic flow of freqtrade


docs/bot-basics.md
```diff
@@ -0,0 +1,58 @@
+# Freqtrade basics
+
+This page will try to teach you some basic concepts on how freqtrade works and operates.
```
hroff-1902, line 3:
```suggestion
This page provides you some basic concepts on how Freqtrade works and operates.
```

docs/bot-basics.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Freqtrade terminology
+
+* Trade: Open position
```
hroff-1902, line 7:
```suggestion
* Trade: Open position.
```

hroff-1902:
for `bot_loop_start()`:

I'm proposing to pass the current running state to it as a parameter and return the state from the method.
This will allow to stop (pause) the loop in some emergency conditions (e.g., if draw down exceeds some value or wallets got less than some value)

The default is to simply return same (Running) state.

docs/strategy-advanced.md
```diff
@@ -1,7 +1,12 @@
 # Advanced Strategies
 
 This page explains some advanced concepts available for strategies.
-If you're just getting started, please be familiar with the methods described in the [Strategy Customization](strategy-customization.md) documentation first.
+If you're just getting started, please be familiar with the methods described in the [Strategy Customization](strategy-customization.md) documentation and with the [Freqtrade basics](bot-basics.md) first.
+
+[Freqtrade basics](bot-basics.md) describes in which sequence each method defined below is called, which can be helpful to understand which method to use.
```
hroff-1902, line 6:
```suggestion
[Freqtrade basics](bot-basics.md) describes in which sequence each method described below is called, which can be helpful to understand which method to use for your custom needs.
```

docs/strategy-advanced.md
```diff
@@ -1,7 +1,12 @@
...
+
+!!! Note
+    All callback methods described below should only be implemented in a strategy if they are also actively used.
```
hroff-1902, line 9: Sorry, I did not understand the sentence (particularly, the usage of "also" here)...

docs/strategy-advanced.md
```diff
@@ -89,3 +94,108 @@ class Awesomestrategy(IStrategy):
             return True
         return False
 ```
+
+## Bot loop start callback
+
+A simple callback which is called at the start of every bot iteration.
+This can be used to perform calculations which are pair independent.
```
hroff-1902, line 101:
```suggestion
A simple callback which is called once at the start of every bot throttling iteration.
This can be used to perform calculations which are pair independent (apply to all pairs), loading of external data, etc.
```

docs/strategy-advanced.md
```diff
@@ -89,3 +94,108 @@ class Awesomestrategy(IStrategy):
...
+            # can then be used by populate_* methods
+            self.remote_data = requests.get('https://some_remote_source.example.com')
+
+```
+
+## Bot order confirmation
+
+### Trade entry (buy order) confirmation
+
+`confirm_trade_entry()` an be used to abort a trade entry at the latest second (maybe because the price is not what we expect).
```
hroff-1902, line 128:
```suggestion
`confirm_trade_entry()` can be used to abort a trade entry at the latest second (maybe because the price is not what we expect).
```

docs/strategy-advanced.md
```diff
@@ -89,3 +94,108 @@ class Awesomestrategy(IStrategy):
...
+        :return bool: When True is returned, then the buy-order is placed on the exchange.
+            False aborts the process
+        """
+        return True
+
+```
+
+### Trade exit (sell order) confirmation
+
+`confirm_trade_exit()` an be used to abort a trade exit (sell) at the latest second (maybe because the price is not what we expect).
```
hroff-1902, line 161:
```suggestion
`confirm_trade_exit()` can be used to abort a trade exit (sell) at the latest second (maybe because the price is not what we expect).
```

docs/strategy-customization.md
```diff
@@ -1,6 +1,8 @@
 # Strategy Customization
 
-This page explains where to customize your strategies, and add new indicators.
+This page explains how to customize your strategies, and add new indicators.
```
hroff-1902, line 3:
```suggestion
This page explains how to customize your strategies, add new indicators and set up trading rules.
```

docs/strategy-customization.md
```diff
@@ -1,6 +1,8 @@
...
+
+Please familiarize yourself with [Freqtrade basics](bot-basics.md) first.
```
hroff-1902, line 5:
```suggestion
Please familiarize yourself with [Freqtrade basics](bot-basics.md) first, which provides overall info on how the bot operates.
```

docs/strategy-advanced.md
```diff
@@ -89,3 +94,108 @@ class Awesomestrategy(IStrategy):
...
+
+class Awesomestrategy(IStrategy):
+
+    # ... populate_* methods
+
+    def bot_loop_start(self, **kwargs) -> None:
+        """
+        Called at the start of the bot iteration (one loop).
+        Might be used to perform pair-independent tasks
+        (e.g. gather some remote ressource for comparison)
```
hroff-1902, line 114:
```suggestion
        (e.g. gather some remote resource for comparison)
```

freqtrade/strategy/interface.py
```diff
@@ -191,6 +190,63 @@ def check_sell_timeout(self, pair: str, trade: Trade, order: dict, **kwargs) ->
         """
         return False
 
+    def bot_loop_start(self, **kwargs) -> None:
+        """
+        Called at the start of the bot iteration (one loop).
+        Might be used to perform pair-independent tasks
+        (e.g. gather some remote ressource for comparison)
```
hroff-1902, line 197:
```suggestion
        (e.g. gather some remote resource for comparison)
```

docs/strategy-advanced.md
xmatthias: In short - dont blindly copy and paste it to the strategy

xmatthias:
> I'm proposing to pass the current running state to it as a parameter and return the state from the method.
> This will allow to stop (pause) the loop in some emergency conditions (e.g., if draw down exceeds some value or wallets got less than some value)
> 
> The default is to simply return same (Running) state.

No, that won't work and will create a big problem.

Assume you issue "State.Stopped" - then the bot will stop.
However, the way state works, it'll stop all of freqtradebot, leaving only the worker running.
Nothing from the strategy will be called anymore, so the strategy just commited suicide (calls stop - but it's impossible to recover from it).

I would leave #1841 for a future PR. It can be done in a similar way - but in my opinion, that callback must do the following:
* have the analyzed dataframe available (so an informed decision can be taken)
* only stop buying (most likely by skipping the buy loop) 
* continue handling open trades
* be called again and again, reenabling buying after "some" conditions change

to summarize - to me, this PR can be an enabler for #1841 - but doing this in `bot_loop_start()` is wrong.
if you don't have the latest pair data analyzed - what  would you base your decision to stop / start the bot on?
Also, you might want to base the decision on the handled trades - so ideally, this will be called between `exit_positions()` and `enter_positions()` - but i see that as "out of scope" for this PR.

docs/strategy-advanced.md
```diff
@@ -1,7 +1,12 @@
 # Advanced Strategies
 
 This page explains some advanced concepts available for strategies.
-If you're just getting started, please be familiar with the methods described in the [Strategy Customization](strategy-customization.md) documentation first.
+If you're just getting started, please be familiar with the methods described in the [Strategy Customization](strategy-customization.md) documentation and with the [Freqtrade basics](bot-basics.md) first.
+
+[Freqtrade basics](bot-basics.md) describes in which sequence each method described below is called, which can be helpful to understand which method to use for your custom needs.
+
+!!! Note
+    All callback methods described below should only be implemented in a strategy if they are also actively used.
```
hroff-1902, line 9:
```suggestion
    All callback methods described below should only be implemented in a strategy if they are actually used.
```

hroff-1902:
> for `bot_loop_start()`:
> 
> I'm proposing to pass the current running state to it as a parameter and return the state from the method.
> This will allow to stop (pause) the loop in some emergency conditions (e.g., if draw down exceeds some value or wallets got less than some value)
> 
> The default is to simply return same (Running) state.
> 
> This will resolve long-awaiting #1841
> 
> This can probably be implemented in a complement `bot_loop_end()` callback (after the loop, so the results of the execution of the loop will be available from the persistence layer and the state returned then goes directly to the worker looping code), leaving `bot_loop_start()` as it is now. (Thus, with this approach, this can be added later in a sep. PR.)
> 
> As an option, both `bot_loop_start()` and `bot_loop_end()` can accept and return the state: `bot_loop_start()` can thus indicate that the loop is to be paused in case of problems if it contains some custom initialization code, which may fail.

Well, it's not a suicide of the strategy, but the emergency brake/switch/fuse, pausing the strategy. Setting state to Stop actually just pauses throttling (with optional closing of the open orders if it's set), the exchange object is still on, the RPC is listening for remote commands, so the strategy can be run again manually and/or the configuration can be reloaded after fixing, if needed (or simply when the bad market conditions have gone)...

After additional thinking I now guess that we should better leave `bot_loop_start()` as it is now (no state passed and returned, leaving it for custom data load, initialization things, etc,) while the "emergency brake" should probably better be left for the separate `bot_loop_end()` (which should accept the state and return it) -- it will be easy for implementation this way and easy to document it this way. Let's leave it for the next separate PR.

xmatthias:
> After additional thinking I now guess that we should better leave bot_loop_start() as it is now (no state passed and returned, leaving it for custom data load, initialization things, etc,) while the "emergency brake" should probably better be left for the separate bot_loop_end() (which should accept the state and return it) -- it will be easy for implementation this way and easy to document it this way. Let's leave it for the next separate PR.

It should definitely be done in a seperate PR.
I'm not convinced of having an "emergency break" though. 
Running `/stop` is a dangerous thing if you still have trades open - so ideally, it'll be as i said above - a "stop_buy" thing, which just skips the part that buys - but leaves handling of open-trades the same (so it should be placed after sell, right before buy.
But - that's for another day :D

hroff-1902:
LGTM 👍 
(I have some doubts for the name of `bot_loop_start`, as discussed on Slack, but I cannot find a better name)

xmatthias:
naming things is hard - i know that ... :laughing:

PR #3517, opened 2020-06-24
Show winning vs. losing trades

xmatthias:
## Summary
Show winning vs. Losing trades in `/profit` calls.

This will allow to easily determine if a strategy made profit on most trades, or only had a few big winners.

## Quick changelog

- add Win / Loss to profit output.

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -284,7 +284,7 @@ class RPC:
                 profit_ratio = trade.close_profit
                 profit_closed_coin.append(trade.close_profit_abs)
                 profit_closed_ratio.append(profit_ratio)
-                if trade.close_profit > 0:
+                if trade.close_profit >= 0:
                     winning_trades += 1
                 else:
                     losing_trades += 1
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -269,6 +269,8 @@ class RPC:
         profit_closed_coin = []
         profit_closed_ratio = []
         durations = []
+        winning_trades = 0
+        losing_trades = 0
 
         for trade in trades:
             current_rate: float = 0.0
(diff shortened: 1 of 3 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -282,6 +284,10 @@ def _rpc_trade_statistics(
                 profit_ratio = trade.close_profit
                 profit_closed_coin.append(trade.close_profit_abs)
                 profit_closed_ratio.append(profit_ratio)
+                if trade.close_profit > 0:
```
Theagainmen, line 287: Kinda nitpicking, but I'd say `>= 0`. I wouldn't count a draw as a loss. It's neither of them, but you lost nothing.

xmatthias: Well it's a discussion point for sure.
I've been thinking about this too - but in the end, a draw is not what you (should) aim for - it's profit - so this would mean that the strategy didn't do what it should've.

Now realistically it's almost impossible to reach "0 profit" - not after exchange rounding of both prices, amounts, fees ... you can be close to 0, like 0.00001 - or -0.00001 - but never 0...

hroff-1902:
I would keep same semantics for Wins as for backtesting: use '>' for wins and add Draws (to be Wins/Draws/Losses)

xmatthias:
well the problem is - draws is an unnecessary information as it's something that cannot be achived in live modes.

it's impossible to get, as it'll always be 0.000001 or -0.00001 (simply because of exchange precision limits).

hroff-1902:
> well the problem is - draws is an unnecessary information as it's something that cannot be achived in live modes.
> 
> it's impossible to get, as it'll always be 0.000001 or -0.00001 (simply because of exchange precision limits).

Agree, but the output will be aligned with Profit/Dra/Loss values printed for backtesting...

Issue #3515, opened 2020-06-23
issues with donwloading data from Binance and Kraken.


fogkaspoker:
## Describe your environment

  * Operating system: Proxmox server
  * Python Version: 3.8.1
  * CCXT version: 1.30.17
 * Freqtrade Version: freqtrade 2020.5
  
## Your question
-- issue for Binance:

When I download the data for 1d timeframe I only get the last 27 Candles even though I've specified the --days command. Any ideas on why this is happening??

Also, When I use the commands, the Download is taking a lot of time and it is donwloading the trades even though it is not specified. (Like in Kraken). 

I am using the command: download-data --exchange binance -t 1d --days 50 --pairs BTC/USDT
 

2020-06-23 22:01:15,736 - freqtrade.loggers - INFO - Verbosity set to 0
2020-06-23 22:01:15,737 - freqtrade.configuration.configuration - INFO - Using exchange binance
2020-06-23 22:01:15,816 - freqtrade.configuration.configuration - INFO - Using user-data directory: /freqtrade/user_data ...
2020-06-23 22:01:15,816 - freqtrade.configuration.configuration - INFO - Using data directory: /freqtrade/user_data/data/binance ...
2020-06-23 22:01:15,817 - freqtrade.configuration.configuration - INFO - Using pairs ['BTC/USDT']
2020-06-23 22:01:15,817 - freqtrade.configuration.configuration - INFO - timeframes --timeframes: ['1d']
2020-06-23 22:01:15,817 - freqtrade.configuration.configuration - INFO - Detected --days: 50
2020-06-23 22:01:15,817 - freqtrade.configuration.configuration - INFO - Using "json" to store OHLCV data.
2020-06-23 22:01:15,817 - freqtrade.configuration.configuration - INFO - Using "jsongz" to store trades data.
2020-06-23 22:01:15,817 - freqtrade.configuration.check_exchange - INFO - Checking exchange...
2020-06-23 22:01:15,817 - freqtrade.configuration.check_exchange - INFO - Exchange "binance" is officially supported by the Freqtrade development team.
2020-06-23 22:01:15,817 - freqtrade.configuration.config_validation - INFO - Validating configuration ...
2020-06-23 22:01:15,825 - freqtrade.commands.data_commands - INFO - About to download pairs: ['BTC/USDT'], intervals: ['1d'] to /freqtrade/user_data/data/binance
[...]

PR #3379, opened 2020-05-26
Fix #3378

hroff-1902:
Fixes #3378 

@xmatthias pls review this carefully.

-- seems it was introduced my cleanup commit: a8b1dcf3c8b5d8fde4e0ea7c7f0ea22d1b189092
-- seems it was (or a similar issue) already reported, I localized it some time ago, but I cannot neither find that previous issue nor a message at Slack/a PM in Slack from someone.
-- I do not actually understand the conditions when this bug appears, maybe under docker (I have not reproduced the bug under the docker), the docker container changes the timezone to be different with the timezone on the local machine

PR #3519, opened 2020-06-24
dl-trades should also support increasing download span

xmatthias:
#3379  Summary
--dl-trades should also support increasing download span
(by downloading the whole dataset again to avoid missing data in the
middle).

This option is there already for ohlcv download - but was missing in trades download - forcing users to manually delete the data if they decide they need better / more data.

Solve the issue: #3515


hroff-1902:
LGTM 👍

PR #3520, opened 2020-06-24
Fix RPC Cors

xmatthias:
## Summary
When making authenticated calls to the rest API from a different ressource (and be that only a different localhost port), then a valid CORS policy is a must to allow the requests to not be blocked by the browser.

This PR will enable this by adding a setting, where users can define their source themselfs.
Regular users won't need to change this - but developers and people working on FreqUI might to allow cross-origin testing (without routing everything throught the webpack server).

## Quick changelog

- add `CORS_origins`  setting to api_server

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -13,6 +13,7 @@ Sample configuration:
         "listen_port": 8080,
         "verbosity": "info",
         "jwt_secret_key": "somethingrandom",
+        "CORS_origins": [],
         "username": "Freqtrader",
         "password": "SuperSecret1!"
     },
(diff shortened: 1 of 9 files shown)


docs/rest-api.md
```diff
@@ -232,3 +233,26 @@ Since the access token has a short timeout (15 min) - the `token/refresh` reques
 > curl -X POST --header "Authorization: Bearer ${refresh_token}"http://localhost:8080/api/v1/token/refresh
 {"access_token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE1ODkxMTk5NzQsIm5iZiI6MTU4OTExOTk3NCwianRpIjoiMDBjNTlhMWUtMjBmYS00ZTk0LTliZjAtNWQwNTg2MTdiZDIyIiwiZXhwIjoxNTg5MTIwODc0LCJpZGVudGl0eSI6eyJ1IjoiRnJlcXRyYWRlciJ9LCJmcmVzaCI6ZmFsc2UsInR5cGUiOiJhY2Nlc3MifQ.1seHlII3WprjjclY6DpRhen0rqdF4j6jbvxIhUFaSbs"}
 ```
+
+## CORS
+
+All web-based frontends are subject to [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) - Cross-Origin Resource Sharing.
+Since most request to the Freqtrade API must be authenticated, a proper CORS policy is key to avoid security problems.
```
hroff-1902, line 240:
```suggestion
Since most of the requests to the Freqtrade API must be authenticated, a proper CORS policy is key to avoid security problems.
```
(or "most requests")

docs/rest-api.md
```diff
@@ -232,3 +233,26 @@ Since the access token has a short timeout (15 min) - the `token/refresh` reques
...
+Also, the Standard disallows `*` CORS policies for requests with credentials, so this setting must be done appropriately.
```
hroff-1902, line 241:
```suggestion
Also, the standard disallows `*` CORS policies for requests with credentials, so this setting must be set appropriately.
```

docs/rest-api.md
```diff
@@ -232,3 +233,26 @@ Since the access token has a short timeout (15 min) - the `token/refresh` reques
...
+
+Users can configure this themselfs via the `CORS_origins` configuration setting.
```
hroff-1902, line 243:
```suggestion
Users can configure this themselves via the `CORS_origins` configuration setting.
```

docs/rest-api.md
```diff
@@ -232,3 +233,26 @@ Since the access token has a short timeout (15 min) - the `token/refresh` reques
...
+It consists of a list of allowed sites that are allowed to consume resources from the bot's API.
+
+Assuming your Application would be deployed as `https://frequi.freqtrade.io/home/` - this would mean that the following configuration becomes necessary:
```
hroff-1902, line 246:
```suggestion
Assuming your application is deployed as `https://frequi.freqtrade.io/home/` - this would mean that the following configuration becomes necessary:
```

PR #3521, opened 2020-06-24
AgeFilter for filtering out newly listed pairs

gambcl:
## Summary
Adds a new pairslist filter called AgeFilter.

## Quick changelog
- Fixed a few typos in comments
- Added AgeFilter (with tests)

## What's new?
AgeFilter checks that each pair has been listed on the exchange for a minimum number of days. It does this by requesting daily candles, once a pair has passed this check it is cached, so no further candle requests will be made. Pairs under the required age will still be checked until they also eventually pass the required age threshold.

### DIFF
--- a/config_full.json.example
+++ b/config_full.json.example
@@ -64,6 +64,7 @@
             "sort_key": "quoteVolume",
             "refresh_period": 1800
         },
+        {"method": "AgeFilter", "min_days_listed": 10},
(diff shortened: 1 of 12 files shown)


hroff-1902:
Excellent idea, thanks!

First time, I wanted to ask when/what for it should/can be used, but the explanation added into the docs covered it completely.

PR #3552, opened 2020-07-02
API server FIAT init fix

Theagainmen:
## Summary
Init the FIAT converter for the API server.

(Also added a small debug message in the AgeFilter, since I think it was just better to see the actual compared to the treshold value. The spread filter also shows the actual value compared to the treshold value. I know I should create a seperate PR for this probably, but it's such a small change..)

## Quick changelog
- 

## What's new?
- Init the FIAT converter in API server
- Minor: AgeFilter shows the actual days that a removed pair is listed on the exchange.
- Updated the tests


tests/rpc/test_rpc_apiserver.py
```diff
@@ -431,14 +431,14 @@ def test_api_profit(botclient, mocker, ticker, fee, markets, limit_buy_order, li
                        'latest_trade_date': 'just now',
                        'latest_trade_timestamp': ANY,
                        'profit_all_coin': 6.217e-05,
-                       'profit_all_fiat': 0,
+                       'profit_all_fiat': ANY,
```
xmatthias, line 434: I think instead of using ANY i would use the real outcome (which is `0.76748865` in both instances).

It's mocked in conftest, and price is set to 12345.0.

PR #3570, opened 2020-07-08
Added range checks to min_days_listed in AgeFilter

gambcl:
## Summary
Adds range checks to min_days_listed in AgeFilter, logs messages if too low/high.

## Quick changelog
- Added range checks to min_days_listed in AgeFilter.
- Added tests for above.

## What's new?
Using min_days_listed < 1 will cause AgeFilter to be ignored.
Using min_days_listed > exchange's ohlcv_candle_limit will cap min_days_listed to this value.

### DIFF
--- a/freqtrade/pairlist/AgeFilter.py
+++ b/freqtrade/pairlist/AgeFilter.py
@@ -5,6 +5,7 @@ import logging
 import arrow
 from typing import Any, Dict
 
+from freqtrade.exceptions import OperationalException
 from freqtrade.misc import plural
 from freqtrade.pairlist.IPairList import IPairList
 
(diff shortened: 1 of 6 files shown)


freqtrade/pairlist/AgeFilter.py
```diff
@@ -23,6 +23,16 @@ def __init__(self, exchange, pairlistmanager,
         super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
 
         self._min_days_listed = pairlistconfig.get('min_days_listed', 10)
+
+        if self._min_days_listed < 1:
+            self.log_on_refresh(logger.info, "min_days_listed must be >= 1, "
```
hroff-1902, line 28: OperationalException should be raised here instead of logging (as for validation of parameters implemented in `__init__` of other Pairlist Handlers...

(same a few lines below)

PR #3577, opened 2020-07-11
[minor] Fix typo in docs, install sqlite3 in docker image

### LINKED ISSUE #3568: Typo in the SQL cheatsheet
## Describe the problem:

In the documentation, [SQL cheatsheet](https://www.freqtrade.io/en/latest/sql_cheatsheet/#fix-trade-still-open-after-a-manual-sell-on-the-exchange) seems to show a **wrong calculation** for `close_profit_abs`, that is:

```
  close_profit_abs = (amount * <close_rate> * (1 - fee_close) - (amount * open_rate * 1 - fee_open)),
```
which in the last parenthesis is not applying `fee_open` correctly: **it lacks of other parenthesis enveloping `1 - fee_open`**. As it is now, an identity multiplication of `open_rate` would occur and then a subtraction of the percentage fee, which makes no sense: beware those who update their DB manually with it!

The **corrected line** that should be in the docs should be:
```
  close_profit_abs = (amount * <close_rate> * (1 - fee_close) - (amount * open_rate * (1 - fee_open))),
```

xmatthias:
## Summary
Fix wrong calculation in sql cheatsheet (as pointed out in #3568).
Also adds sqlite3 to the dockerfile, so editing the database is possible without much effort from within docker.

closes #3568

### DIFF
--- a/docs/sql_cheatsheet.md
+++ b/docs/sql_cheatsheet.md
@@ -100,8 +100,8 @@ UPDATE trades
 SET is_open=0,
   close_date=<close_date>,
   close_rate=<close_rate>,
-  close_profit=close_rate/open_rate-1,
-  close_profit_abs = (amount * <close_rate> * (1 - fee_close) - (amount * open_rate * 1 - fee_open)),
+  close_profit = close_rate / open_rate - 1,
+  close_profit_abs = (amount * <close_rate> * (1 - fee_close) - (amount * (open_rate * 1 - fee_open))),
   sell_reason=<sell_reason>
 WHERE id=<trade_ID_to_update>;
 ```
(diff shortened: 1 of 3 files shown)


docs/sql_cheatsheet.md
```diff
@@ -111,24 +120,39 @@ WHERE id=<trade_ID_to_update>;
...
+```sql
+DELETE FROM trades WHERE id = <tradeid>;
 ```
+
+```sql
+DELETE FROM trades WHERE id = 31;
+```
+
+!!! Warning
+    This will remove this trade from the database. Please make sure you got the correct id and **NEVER** run this query without the where clause.
```
hroff-1902, line 158:
```suggestion
    This will remove this trade from the database. Please make sure you got the correct id and **NEVER** run this query without the `where` clause.
```

PR #3591, opened 2020-07-14
vectorized backtest

orehunt:
It currently supports buy/sell/stoploss
- you would enable it with config (but you have to replace the backtesting instance in hyperopt first)
```json
"backtesting_engine": "vectorized"
```
- not supported: `max_open_trades` , `roi`, `trailing`
- supported: `position_stacking`
- stoploss calculation can be toggled with config
```json
"backtesting_amounts": {
        "stoploss": false,
}
```
other differences:
- amounts calculation seems a little off (im 9/10 sure it's not about `Decimal`), profits formula is
```python
(close_rate - close_rate * self.fee) / (
            open_rate + open_rate * self.fee
        ) - 1
```
- trade duration takes half the timeframe instead of zero when buy and sell
happen on same candle (this would eventually be an option to choose with 0/half/full)
- `force_sell` are ignored, the last valid trades would be triggered by stoploss (there is an edge case where trades bought on the last candle are ignored even if the stoploss would trigger, this can probably be fixed, didn't bother with it)

Adding `roi` `trailing` should be fairly "trivial", the bulk of the complexity is added by stoploss, when you have to respect order (no position stacking), and `roi` `trailing` would just be conditions xorred with stoploss depending on the chosen logic
there are implementation for some functions both in pandas and numpy, currently it uses the fastest (which is usually numpy) but I'm keeping the pandas ones in case a gpu backend ever reaches features parity (lol)

### DIFF
--- a/freqtrade/optimize/hyperopt_backtest.py
+++ b/freqtrade/optimize/hyperopt_backtest.py
@@ -1,0 +1,1005 @@
+import logging
+
+import arrow
+import gc
+from typing import Dict, List, Tuple, Union
+from enum import IntEnum
+
+from numba import njit
+from numpy import (
+    repeat,
+    ones,
+    nan,
+    concatenate,
+    ndarray,
+    array,
+    where,
+    transpose,
+    maximum,
+    full,
+    unique,
+    insert,
+    isfinite,
+    isnan,
+)
+from pandas import (
+    Timedelta,
+    Series,
+    DataFrame,
+    Categorical,
+    Index,
+    MultiIndex,
+    # SparseArray,
+    set_option,
+    to_timedelta,
+    to_datetime,
+)
+
+from freqtrade.optimize.backtesting import Backtesting, BacktestResult
+from freqtrade.exceptions import OperationalException
+from freqtrade.strategy.interface import SellType
+
+
+logger = logging.getLogger(__name__)
+
+# import os
+# import psutil
+# process = psutil.Process(os.getpid())
+set_option("display.max_rows", 1000)
+
+
+class Candle(IntEnum):
+    BOUGHT = 2
+    SOLD = 5
+    NOOP = 0
+    END = 11  # references the last candle of a pair
+    # STOPLOSS = 17
+
+
+@njit  # fastmath=True ? there is no math involved here though..
+def for_trail_idx(index, bos, rate, stop_idx):
+    last = -2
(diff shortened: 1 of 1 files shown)


hroff-1902:
pls add some description for it

xmatthias:
What is the difference to edge? if you strip edge of the "additional" stuff it's doing (expectancy calculation, ....) - it's basically doing something very similar.
Also edge does not include ROI, nor trailing stoploss - which to me, are major gaps as that excludes most of my strategies from the getgo (i usually don't use an explicit sell signal, but instead rely on trailing_stop).

To me, this must be 100% compatible to backtesting ( yield the same results with the same inputs), otherwise it'll not make much sense (we got edge already, which is faster than backtesting, but by dropping certain features) - as then we have no proof that what this reports is actually anywhere close to reality (we did multiple tests for backtesting in the past, and (with a few caveats) results are aligned to dry-run / live runs (assuming all trades fill obviously).

You should be able to test this by slightly modifying the tests in `test_backtestingdetails.py` to also test against this class - but obviously it'll need ROI and trailing stoploss support first, otherwise "most" tests will fail right away.

hroff-1902:
How does this vectorized backtesting differs from the Freqtrade "standard" backtesting?

orehunt:
> How does this vectorized backtesting differs from the Freqtrade "standard" backtesting?

what do you mean? I wrote the differences 

checking out edge was the next thing, it seems it mostly achieves speed ups with the find_1st extension, it might speed up my version where I use numpy searchsorted, although most of the times it would use the top/down one, it chooses which one to use based on the mean number of candles between buy and sells, and most of the times buy and sell would be close in number

note that the current version of edge is looping over pairs and appending each trade singularly, just those two are **very** big differences in speed

hroff-1902:
I think we should better get rid of find_1st and refactor Edge to remove it from the dependencies -- this package seems to be unsupported, does not provide wheels and requires compilation dependencies for latest python versions (many questions on Slack were related to it).

xmatthias:
In it's current state i don't see this as mergable, and since it's been inactive for almost a year now, i'm going to close this.

Feel free to reopen / resubmit if you feel like implementing all features (ROI, trailing stop, ...) for it - but otherwise i see it as a slightly different version to edge, which will again not work for many startegies that rely on trailing stop / ROI for their sell signals.

PR #3592, opened 2020-07-14
Add stoploss-distance (to current price) to /status output

xmatthias:
## Summary
Add "Stoploss distance:" to telegram output - showing the absolute (and relative) distance to the stoploss.
This is quite usefull when using trailing stop - as you'll see how far from the stoploss current price is - so you can "kinda" expect when it'll be hit.

## Quick changelog

- Add stoploss distance to current price

2020-07-14-201759_489x366_scrot


hroff-1902:
I think it should be positive,

I mean, if stoploss is -5.0%, current profit is -1.0%, then the distance is 4.0% (not -4.0%)

xmatthias:
but it's the distance between the current price and stoploss - i think having it positive would be confusing as people could mistake it with profit.

My thought is - how much should a price drop so it's beeing sold via stoploss.

in my above screenshot, it's not 6% profit - but a price-drop of 6% (hence -6%) would cause the stoploss to trigger.

think about another positive case with trailing stop 
current profit is 5% - stoploss offset is -10% - now trailing stoploss activates, pulls the stoploss to -1% (of the maximum).
after a bit of price fluctuations the trailing stop and current price are 0.5% apart.

hroff-1902:
There is the logic behind your explanation for sure, but the problem with is that the distance becomes positive when the current price is below the stoploss rate and the trade is not yet closed (for any reason -- if stoploss is not fulfilled yet or did not have a chance to execute due to an error in the bot), which seems to be confusing.

In my definition of distance, if current rate is above stoploss -- then the distance is positive, if current rate is below stoploss -- then the distance is negative.

Theagainmen:
For me the negative stoploss distance makes more sense. The positive one just seems kinda confusing. When I see that I'm in 1% profit, and my stoploss is at -5%, it makes more sense to me when it says distance -6%. Because from the _current_ point I am in, it _takes -6%_ to get to my _stoploss_.

xmatthias:
> Also not sure what you mean with this hroff if current rate is below stoploss, this shouldn't even happen in the first place right? And if it happens, that's more of an exceptional situation. If it gets below the stoploss I think it should be rather positive or maybe say something: 'current value already below stoploss`.

technically, it can happen (for example):
I usually use reversed sides with limit orders (so sell orders take the bid price).
Now my stoploss is placed at 95 - - spread is 94.8 - 95.2 (ok, that's extreme - but serves a purpose).
The stoploss is in the middle of spread - but no trade did happen < 95 yet.

Because of this, the calculated current rate can be below the stoploss rate (remember, i look at bid for the sell rate) - however the exchange ONLY looks at the trades that are made (effectively the "last" value) to trigger the stoploss - so stoploss won't be triggered until a trade <95 is made.
Now it's an exceptional case and you need to call /status exactly at the right moment - but it's still a technical possibility. 

It should however not happen when stoploss is not on the exchange, as then the bot would use the calculated rate to determine to trigger the stoploss.
However, even then - if you use a high trottle time - the value can drop below the stoploss for sure until the bot realizes (that's why i wouldn't touch trottle time myself ...).

hroff-1902:
mathematically, the distance is the difference from the target point, `x-x0`
if x is above x0, than the difference (distance) should be positive

for me, positive distance values printed when the price is below the stoploss rate (when the stoploss stop order was triggered but the price did not reach the stop order value yet) is very confusing

xmatthias:
I think the main point here is how you think about it.

Think "by this much my asset has to fall to trigger stoploss".
It'll then also work by "how much has the asset to *raise* to trigger stoploss".

Think about the future too - if we're able to short- then stoploss is "above" current price.
based on your logic then:

```
# LONG:
stoploss = 90
current_price = 100
distance_to_stoploss = -10 # price has to fall by 10
Reversed_stoploss = 10
# Short:
stoploss = 110
current_price = 100
distance_to_stoploss = 10 # price has to raise by 10
reversed_stoploss = -10
```


The case you're referring (current price < stoploss) is a rare edge-case, which i assume almost no one will ever see - so we're discussing about changing the logic for 99% of the cases to make no sense to most people - just because the 1% case will not make sense ...

hroff-1902:
What about:

`Stoploss distance: 0.01660000 (6.30%) above stoploss`

`Stoploss distance: 0.00166000 (0.630%) below stoploss`

?

xmatthias:
No, this is very confusing.

xmatthias:
After testing this with frequi, i think the propose approach is the correct one (let's keep in mind, this also applies to "non-trailing" situations).

2020-08-25-195438_712x567_scrot

And with loss:

2020-08-25-195431_683x518_scrot

I'll therefore take the decision and merge this as proposed.
future adaptations are obviously always possible, shoudl we see some problems with this.

PR #3596, opened 2020-07-15
Allow 0 fee value by correctly checking for None

### LINKED ISSUE #3594: configure fee to 0 using --fee argument take default exchange fees instead of 0
Hi,

During early stage of developing a strategy i put fees to 0 to see winrate when i close the trade after 1 bar, 2 bars, ... .
No matter i 'force' the fees to 0 it seems the fallback to default exchange fees if the value choose is under the exchange fee.

Have a nice day.

xmatthias:
## Summary
Fees of 0.0 (or 0) sholuld also be allowed, since some exchanges allow free trades (and it can be useful for testing, eventually).

closes #3594

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -101,7 +101,7 @@ class Backtesting:
         if len(self.pairlists.whitelist) == 0:
             raise OperationalException("No pair in whitelist.")
 
-        if config.get('fee'):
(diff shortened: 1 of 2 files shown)


felpasl:
when use bnb to pay fees, are now possible put 0 as value?

xmatthias:
this only impacts backtesting.

BNB fees are NOT free - they count against your total balance as well.

PR #3609, opened 2020-07-19
Add telegram /trades command

thopd88:
## Summary
Add telegram trades command to list recent trades. Using existed _rpc_trade_history

## Quick changelog

- Add command /trades <n> with n (default 10) is the number of trades history

image

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -252,9 +252,10 @@ class RPC:
     def _rpc_trade_history(self, limit: int) -> Dict:
         """ Returns the X last trades """
         if limit > 0:
-            trades = Trade.get_trades().order_by(Trade.id.desc()).limit(limit)
+            trades = Trade.get_trades([Trade.is_open.is_(False)]).order_by(
+                Trade.id.desc()).limit(limit)
         else:
-            trades = Trade.get_trades().order_by(Trade.id.desc()).all()
+            trades = Trade.get_trades([Trade.is_open.is_(False)]).order_by(Trade.id.desc()).all()
 
         output = [trade.to_json() for trade in trades]
 
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -537,28 +537,6 @@ class RPC:
         else:
             return None
 
-    def _rpc_delete(self, trade_id: str) -> Dict[str, str]:
-        """
-        Handler for delete <id>.
-        Delete the given trade
-        """
-        def _exec_delete(trade: Trade) -> None:
-            Trade.session.delete(trade)
-            Trade.session.flush()
-
-        with self._freqtrade._sell_lock:
-            trade = Trade.get_trades(
-                trade_filter=[Trade.id == trade_id, ]
-            ).first()
-            if not trade:
-                logger.warning('delete: Invalid argument received')
-                raise RPCException('invalid argument')
-
-            _exec_delete(trade)
-            Trade.session.flush()
-            self._freqtrade.wallets.update()
-            return {'result': f'Deleted trade {trade_id}.'}
-
     def _rpc_performance(self) -> List[Dict[str, Any]]:
         """
         Handler for performance.
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -537,6 +537,28 @@ class RPC:
         else:
             return None
 
+    def _rpc_delete(self, trade_id: str) -> Dict[str, str]:
+        """
+        Handler for delete <id>.
+        Delete the given trade
+        """
+        def _exec_delete(trade: Trade) -> None:
+            Trade.session.delete(trade)
+            Trade.session.flush()
+
+        with self._freqtrade._sell_lock:
+            trade = Trade.get_trades(
+                trade_filter=[Trade.id == trade_id, ]
+            ).first()
+            if not trade:
+                logger.warning('delete: Invalid argument received')
+                raise RPCException('invalid argument')
+
+            _exec_delete(trade)
+            Trade.session.flush()
+            self._freqtrade.wallets.update()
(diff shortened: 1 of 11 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -537,6 +537,28 @@ def _rpc_forcebuy(self, pair: str, price: Optional[float]) -> Optional[Trade]:
...
+            ).first()
+            if not trade:
+                logger.warning('delete: Invalid argument received')
+                raise RPCException('invalid argument')
+
+            _exec_delete(trade)
+            Trade.session.flush()
+            self._freqtrade.wallets.update()
+            return {'result': f'Deleted trade {trade_id}.'}
+
```
xmatthias, line 561: I don't think we can accept this part as is.

This works fine if you know exactly what state your trades and exchange account is in.
But in most cases when using telegram, you don't know the details (are there open orders ...)

There's things / checks we MUST have when deleting a trade:
* does it have open orders 
  * if yes - cancel these - eventually ignoring failures - assuming the order was cancelled on the exchange - but warning the user about it
  * this can have wider consequences - what if the order is half-filled? this could end up in unsellable assets (too low to place a new order)... 
* same for stoploss orders


obviously - since this is a crucial part - we will want this covered by tests. 
That's not an issue per se (i can help with tests if you're not familiar with pytest).

freqtrade/rpc/telegram.py
```diff
@@ -496,6 +498,66 @@ def _forcebuy(self, update: Update, context: CallbackContext) -> None:
...
+                    'Pair',
+                    'Open rate',
+                    'Stake Amount',
+                    'Close date',
+                    'Close rate',
+                    'Profit',
+                ],
+                tablefmt='simple')
+            message = f'<b>{nrecent} recent trades</b>:\n<pre>{trades_tab}</pre>'
+            self._send_msg(message, parse_mode=ParseMode.HTML)
```
xmatthias, line 557: Would be great if you could post a screenshot (eventually blank out details you don't wanna share) of how this will look in the telegram (web)client ...

Now i'll test this myself for sure - but that would make this ... friendlier for random people looking at this PR.

thopd88:
> I would recommend to split this PR up into 2 PR's
> 
> * one (the simpler one) adding the "/trade" endpoint
> * the `/delete` part.
> 
> As you'll see in the above (or below) comments, the delete part needs a lot more work - as it's not as simple as deleting an order without risking to create more problems than it'll solve.

Haha I tried that, but the second commit was joined automatically with the previous PR

freqtrade/rpc/telegram.py
```diff
@@ -496,6 +498,66 @@ def _forcebuy(self, update: Update, context: CallbackContext) -> None:
...
+                  f"{trade['open_rate']}",
+                  f"{trade['stake_amount']}",
+                  trade['close_date'],
+                  f"{trade['close_rate']}",
+                  f"{trade['close_profit_abs']}"] for trade in trades['trades']],
+                headers=[
+                    'Open Date',
+                    'Pair',
+                    'Open rate',
+                    'Stake Amount',
```
xmatthias, line 550: I'm not certain if we should use Stake-Amount, or rather "amount"? 
usage of stake-amount is only usefull in the edge-case where you set stake_amount to unlimited, otherwise it'll always be "X".

freqtrade/rpc/telegram.py
```diff
@@ -496,6 +498,66 @@ def _forcebuy(self, update: Update, context: CallbackContext) -> None:
...
+    def _trades(self, update: Update, context: CallbackContext) -> None:
+        """
+        Handler for /trades <n>
+        Returns last n recent trades.
+        :param bot: telegram bot
+        :param update: message update
+        :return: None
+        """
+        stake_cur = self._config['stake_currency']
+        fiat_disp_cur = self._config.get('fiat_display_currency', '')
```
xmatthias, line 529: Once tests pass - these 2 lines should be removed. They're not used and flake8 will freak about them :laughing:

thopd88:
@xmatthias I reverted the /delete commit

xmatthias:
I've fixed the failing tests.
However - i really have a problem with the way this looks (but have no idea how this could be improved either).

If we look at how this is now ... it's barely readable (due to the rows wrapping to 3 rows per row), as this is simply too much for the small screen telegram (or a phone) would give you.

2020-07-20-191218_539x673_scrot

hroff-1902:
yeah, unreadable...
cannot it be made a html table?

thopd88:
For me, I use this simple
image

xmatthias:
no, telegram won't accept html tables (you get "invalid open tag 'table'" or something similar).

The problem is the amount of columns - it works for `/status table` since that only has 3 columns - this one has 5 or 6 - which makes it a problem.

xmatthias:
@thopd88 i've just seen now that this was a proposal for a new format - yes, the 3 columns should work - maybe adding the profit% in parenteses would work, too? 
```
Date               Pair                 Profit
2hours ago      ETH/USDT       2(5%)
```

also - i think filtering out open trades might make sense here? they can be found via `/status table` - and open trades require more (live) information.
Now thats something i'm thinking about the whole /trades endpoint (also for the API) ... so might be left for later.

thopd88:
> @thopd88 i've just realized now that this was a proposal for a new format - yes, the 3 columns should work - maybe adding the profit% in parenteses would work, too?
> 
> ```
> Date               Pair                 Profit
> 2hours ago      ETH/USDT       2(5%)
> ```
> 

How about this:
```
Date               Pair                 Profit (USDT)
2hours ago      ETH/USDT       5% (2)
```

> also - i think filtering out open trades might make sense here? they can be found via `/status table` - and open trades require more (live) information.

Yes I think it makes sense

> Now thats something i'm thinking about the whole /trades endpoint (also for the API) ... so might be left for later.

I think later haha

PR #3611, opened 2020-07-20
Add telegram /delete command

thopd88:
## Summary
Add Telegram /delete command to immediately delete specific trade in the database

## Quick changelog

- Add Telegram /delete command

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -6,7 +6,7 @@ from abc import abstractmethod
 from datetime import date, datetime, timedelta
 from enum import Enum
 from math import isnan
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import arrow
 from numpy import NAN, mean
@@ -538,7 +538,7 @@ class RPC:
         else:
             return None
 
-    def _rpc_delete(self, trade_id: str) -> Dict[str, str]:
+    def _rpc_delete(self, trade_id: str) -> Dict[str, Union[str, int]]:
         """
         Handler for delete <id>.
         Delete the given trade and close eventually existing open orders.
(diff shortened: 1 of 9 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -536,6 +536,28 @@ def _rpc_forcebuy(self, pair: str, price: Optional[float]) -> Optional[Trade]:
...
+                trade_filter=[Trade.id == trade_id, ]
+            ).first()
+            if not trade:
+                logger.warning('delete: Invalid argument received')
+                raise RPCException('invalid argument')
+
+            _exec_delete(trade)
+            Trade.session.flush()
+            self._freqtrade.wallets.update()
+            return {'result': f'Deleted trade {trade_id}.'}
```
xmatthias, line 560: I don't think we can accept this part as is (transferred from the other PR).


This works fine if you know exactly what state your trades and exchange account is in.

But in most cases when using telegram you should assume people are "on the road / not at home", so you don't know the details (are there open orders, did orders fill, ...)

There's things / checks we MUST have when deleting a trade:
* does it have open orders 
  * if yes - cancel these - eventually ignoring failures - assuming the order was cancelled on the exchange - but warning the user about it
  * this can have wider consequences - what if the order is half-filled? this could end up in unsellable assets (too low to place a new order)... 
* same for stoploss orders

now the thing is - this is a command to "clean up" if something went wrong. 
However, simply deleting the trade is not always the right approach (see the caveats above with open orders).
e.g. if you did own the asset (buy order was filled) but sold it manually (or the sell order was somehow badly tracked) - then i would want the trade to be "closed" - using a sell-rate - so it still counts against profit / loss.


---

obviously - since this is a crucial part - we will want this covered by tests. 
That's not an issue per se (i can help with tests if you're not familiar with pytest).

thopd88:
Thanks! Sorry I didn't have time to finish it

PR #3619, opened 2020-07-22
Cleanup AgeFilter, PriceFilter

hroff-1902:
Follows #3570 

* Fix wordings in the messages issued by AgeFilter
* Add checks for the parameters of PriceFilter (cannot be negative); Tests added
* Improve docs


hroff-1902:
@gambcl pls also review the improvements as you added these options

docs/configuration.md
```diff
@@ -663,24 +663,27 @@ Filters low-value coins which would not allow setting stoplosses.
...
+
+For `PriceFiler` at least one of its `min_price`, `max_price` or `low_price_ratio` settings must be applied.
 
 Calculation example:
 
-Min price precision is 8 decimals. If price is 0.00000011 - one step would be 0.00000012 - which is almost 10% higher than the previous value.
+Min price precision for SHITCOIN/BTC is 8 decimals. If its price is 0.00000011 - one price step above would be 0.00000012, which is ~9% higher than the previous price value. You may filter out this pair by using PriceFilter with `low_price_ratio` set to 0.09 (9%) or with `min_price` set to 0.00000011, correspondingly.
 
-These pairs are dangerous since it may be impossible to place the desired stoploss - and often result in high losses.
+Low priced pairs are dangerous since they are often illiquid and it may also be impossible to place the desired stoploss, which can often result in high losses. Consider using PriceFilter with `low_price_ratio` set to a value which is less than the absolute value of your stoploss (for example, if your stoploss is -5% (-0.05), then the value for `low_price_ratio` can be 0.04 or even 0.02).
```
hroff-1902, line 686: @xmatthias pls review this advice separately (if I'm correct here)

freqtrade/pairlist/PriceFilter.py
```diff
@@ -18,11 +19,17 @@ def __init__(self, exchange, pairlistmanager,
...
         self._min_price = pairlistconfig.get('min_price', 0)
+        if self._min_price < 0:
+            raise OperationalException("PriceFilter requires min_price be >= 0")
         self._max_price = pairlistconfig.get('max_price', 0)
-        self._enabled = ((self._low_price_ratio != 0) or
-                         (self._min_price != 0) or
-                         (self._max_price != 0))
+        if self._max_price < 0:
+            raise OperationalException("PriceFilter requires max_price be >= 0")
+        self._enabled = ((self._low_price_ratio > 0) or
```
hroff-1902, line 30: I think it should also raise OperationalException here (if all settings == 0)

freqtrade/pairlist/AgeFilter.py
```diff
@@ -26,12 +26,11 @@ def __init__(self, exchange, pairlistmanager,
         self._min_days_listed = pairlistconfig.get('min_days_listed', 10)
 
         if self._min_days_listed < 1:
-            raise OperationalException("AgeFilter requires min_days_listed must be >= 1")
+            raise OperationalException("AgeFilter requires min_days_listed be >= 1")
```
Theagainmen, line 29: I think it should be `AgeFilter requires min_days_listed to be >= 1` (to be)

freqtrade/pairlist/PriceFilter.py
```diff
@@ -18,11 +19,17 @@ def __init__(self, exchange, pairlistmanager,
         super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
 
         self._low_price_ratio = pairlistconfig.get('low_price_ratio', 0)
+        if self._low_price_ratio < 0:
+            raise OperationalException("PriceFilter requires low_price_ratio be >= 0")
```
Theagainmen, line 23: I think it should be `to be` here again?

docs/configuration.md
```diff
@@ -663,24 +663,27 @@ Filters low-value coins which would not allow setting stoplosses.
...
 
 The `PriceFilter` allows filtering of pairs by price. Currently the following price filters are supported:
+
 * `min_price`
 * `max_price`
 * `low_price_ratio`
 
 The `min_price` setting removes pairs where the price is below the specified price. This is useful if you wish to avoid trading very low-priced pairs.
-This option is disabled by default, and will only apply if set to <> 0.
+This option is disabled by default (or set to 0), and will only apply if set to > 0.
```
xmatthias, line 672: I'm not sure this "(or set to 0)" makes much sense ... 

either it's disabled - or set to 0 by default - but the documentation should know (this sounds like a question to me ...

freqtrade/pairlist/AgeFilter.py
```diff
@@ -26,12 +26,11 @@ def __init__(self, exchange, pairlistmanager,
...
         if self._min_days_listed > exchange.ohlcv_candle_limit:
-            raise OperationalException("AgeFilter requires min_days_listed must not exceed "
+            raise OperationalException("AgeFilter requires min_days_listed be not exceeding "
```
xmatthias, line 31: somehow this sounds odd (from a sentence / grammar perspecive).

maybe 
```suggestion
            raise OperationalException("AgeFilter requires min_days_listed to be below "
```

freqtrade/pairlist/PriceFilter.py
xmatthias: i guess so, yeah

docs/configuration.md
xmatthias: I don't think this is correct.

`PrecisionFilter` is the one that makes sure there's enough "room" below the current price to eventually place a stoploss (basically eliminating trades that trade at the lower end of their tradable range).


For a stoploss of -5% - you'll want to use 0.01 at least (which would mean it's 5 pips below entry).

If you set this value to 0.04, then your step size is 0.04 - so the stoploss can only trigger at -4% or -8% (no price in between is even possible due to the minimum pricemovement beeing so high) - so you'll raise your stoploss to -8% automatically (the first possible value below the configured stoploss).

Personally, i would not trade any pair that's got a minimum pip (minimum movement size) of > 0.001 (0.1%) - but i guess that depends on the strategy.

docs/configuration.md
gambcl: I agree with xmatthias, reads better without the "(or set to 0)".

docs/configuration.md
gambcl: I would again remove the "(or set to 0)" from the max_price description.

Also looks like a typo: PriceFiler instead of PriceFilter

freqtrade/pairlist/AgeFilter.py
gambcl: No strong opinions on this, but I found the original wording to be fine.

freqtrade/pairlist/AgeFilter.py
gambcl: In this case I definitely prefer the original wording.

freqtrade/pairlist/PriceFilter.py
gambcl: I guess it depends on how you think people will be using the config. Allowing filters to be disabled means that users can easily leave a filter in their config but run with it disabled, whereas raising an exception will force users to remove the filter from their config.

xmatthias: what's the difference to setting all values to 0, vs. putting `//` in front of the filter line ... ?

gambcl: Both the same effect I guess, probably cleaner to enforce the values > 0.

freqtrade/pairlist/PriceFilter.py
xmatthias, line 23:
```suggestion
            raise OperationalException("PriceFilter requires low_price_ratio to be >= 0")
```

freqtrade/pairlist/AgeFilter.py
xmatthias:
```suggestion
            raise OperationalException("AgeFilter requires min_days_listed to be >= 1")
```

PR #3626, opened 2020-07-25
Introduce HDF5 Datahandler 

xmatthias:
## Summary
Implement HDF5 DataHandler to enhance speed for backtesting operations.

Tests with "a few" pairs and 900+ days 5m data got from 15s to ~3s (json vs. hdf5) - purely dataloading ((startup time and backtesting itself not included - obviously, numbers may vary between different systems). - however the speedup is non neglectable (especially once moving to 1m data).

HDF5 will also enable "append" (not rewriting the whole data during data-download) - however this will be added as a seperate, future step.

If this proves to work well and fast for all users, i think this should become the new default (not with the next version - but sometime late this year - once this functionality is "battle-proven".

Conversation of existing can be done using the `convert-data`.

```
freqtrade convert-data --datadir user_data/data/binance/ --format-from json --format-to hdf5 -t 5m 
```

## Implementation details

* To reduce the risk of dataloss in case of hdf5 file corruption (not common - but disks break ...), it will use one file per pair / timeframe. This way, users will only loose one file (same as with json) - not a whole pair - or whole dataframe, or even the whole datastore. Speed tests showed no real difference between using one file, or using multiple files.
* trades data is in a seperate file
* data is stored under a subkey ( `<pair>/ohlcv/tf_<timeframe>` | `<pair>/trades`) - this way a mistakingly renamed file can be reattributed to the correct timeframe. It's not something i expect people to actually do - but it doesn't cost anything and we need a key anyway.
* Appending is not implemented as this would require a bigger change to the downloading system, and will follow in a later PR.

## Quick changelog

- Introduce hdf5 datastorage
- add `--data-format-ohlcv <format>` to all backtesting commands (`backtesting`, `hyperopt`, `edge`) - allowing to quickly test with a different dataformat (assuming data is available).


Fredrik81:
Just want to throw this into the mix if it makes any sense
https://towardsdatascience.com/the-best-format-to-save-pandas-data-414dca023e0d

xmatthias:
> Just want to throw this into the mix if it makes any sense
> https://towardsdatascience.com/the-best-format-to-save-pandas-data-414dca023e0d

@Fredrik81 the problem with this is - it purely looks at performance - not at "available features".

* feather and parquet are both columnar files - this makes them fast - but makes appending impossible, so if i want to append one new row - we MUST rewrite the whole file (or at least the whole partition)
While this may not seem like a problem - it will become a problem once you have GB's of trades data (or multi-year 1m data) - as it'll add a lot of wear on your SSD - and will reduce performance as it'll require writing a Gb - instead of appending 2 rows.
* pickle - it's simply a unsafe format ...
* msgpack - i didn't know this - but apparently, pandas removed support for this in 1.0.0 - which doesn't sound promising for it's future...

segatrade:
Idea with feather and parquet - make files partions for every half year for example and trade pair. So it will rewrite less when new data downloads

PR #3677, opened 2020-08-12
improve Edge documentation

Blackhawke:
Re-arranged the introduction to better explain the theory of operation and the limitations of Edge. Added paragraphs at the bottom of "running edge independently" to better explain Edge's order of operations processing and potential differences between...

... historical output and live/dry-run operation.

### DIFF
--- a/docs/edge.md
+++ b/docs/edge.md
@@ -2,7 +2,7 @@
 
 This page explains how to use Edge Positioning module in your bot in order to enter into a trade only if the trade has a reasonable win rate and risk reward ratio, and consequently adjust your position size and stoploss.
 
-  !!! Warning
+!!! Warning
     Edge positioning is not compatible with dynamic (volume-based) whitelist.
 
 !!! Note
--- a/docs/edge.md
+++ b/docs/edge.md
@@ -2,11 +2,13 @@
 
 This page explains how to use Edge Positioning module in your bot in order to enter into a trade only if the trade has a reasonable win rate and risk reward ratio, and consequently adjust your position size and stoploss.
 
-!!! Warning
+  !!! Warning
     Edge positioning is not compatible with dynamic (volume-based) whitelist.
 
-!!! Note
-    Edge does not consider anything else than buy/sell/stoploss signals. So trailing stoploss, ROI, and everything else are ignored in its calculation.
+  !!! Note
+ 1. Edge does not consider anything other than *its own* buy/sell/stoploss signals. It ignores the stoploss, trailing stoploss, and ROI settings in the strategy configuration file.
+
+ 2. Therefore, it is important to understand that Edge can improve the performance of some trading strategies but *decrease* the performance of others.
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


docs/edge.md
```diff
@@ -2,11 +2,13 @@
 
 This page explains how to use Edge Positioning module in your bot in order to enter into a trade only if the trade has a reasonable win rate and risk reward ratio, and consequently adjust your position size and stoploss.
 
-!!! Warning
+  !!! Warning
```
xmatthias, line 5: Indenting these will cause rendering to break, as the tripple exclamation mark MUST be the first 3 letters to render correctly on readthedocs.

Applies to the below Note too, obviously.

docs/edge.md
```diff
@@ -2,11 +2,13 @@
...
     Edge positioning is not compatible with dynamic (volume-based) whitelist.
 
-!!! Note
-    Edge does not consider anything else than buy/sell/stoploss signals. So trailing stoploss, ROI, and everything else are ignored in its calculation.
+  !!! Note
+ 1. Edge does not consider anything other than *its own* buy/sell/stoploss signals. It ignores the stoploss, trailing stoploss, and ROI settings in the strategy configuration file.
+
+ 2. Therefore, it is important to understand that Edge can improve the performance of some trading strategies but *decrease* the performance of others.
```
xmatthias, line 11: I don't think this will render correctly. (`!!! Note` and the following, indented lines are rendered in special boxes in the documentation)

Once the indentation is broken (the empty line) - the box will end.

```suggestion
!!! Note
    Edge does not consider anything other than *its own* buy/sell/stoploss signals. It ignores the stoploss, trailing stoploss, and ROI settings in the strategy configuration file.
    Therefore, it is important to understand that Edge can improve the performance of some trading strategies but *decrease* the performance of others.
```

I'd not put it as numbered list, but simply as text - that should work out fine - considering it's rendered in a "note" box.
The text / wording itself is perfect though :+1:

docs/edge.md
```diff
@@ -186,6 +188,10 @@ An example of its output:
 | APPC/BTC  |      -0.02 |       0.44 |                2.28 |                   1.27 |         0.44 |                       25 |                       43 |
 | NEBL/BTC  |      -0.03 |       0.63 |                1.29 |                   0.58 |         0.44 |                       19 |                       59 |
 
+Edge produced the above table by comparing ``calculate_since_number_of_days`` to ``minimum_expectancy`` to find ``min_trade_number``. Historical information based on the config file. The time frame Edge uses for its comparisons can be further limited by using the ``--timeframe`` switch.
```
xmatthias, line 191: I think what you mean with this is to limit the timerange, not the timeframe (timeframe = 5m - timerange would be "january-march" ...)
Unfortunately, the 2 words are very similar, so we need to be carefull to use them correctly.

```suggestion
Edge produced the above table by comparing `calculate_since_number_of_days` to `minimum_expectancy` to find `min_trade_number` historical information based on the config file. The timerange Edge uses for its comparisons can be further limited by using the `--timerange` switch.
```

docs/edge.md
```diff
@@ -186,6 +188,10 @@ An example of its output:
...
+
+In live and dry-run modes, after the ``process_throttle_secs`` has passed, Edge will again process ``calculate_since_number_of_days`` against ``minimum_expectancy`` to find ``min_trade_number``. If no ``min_trade_number`` is found, the bot will return "whitelist empty". Depending on the trade strategy being deployed, "whitelist empty" may be return much of the time---or *all* of the time. The use of Edge may also cause trading to occur in bursts, though this is rare.
```
xmatthias, line 193: Maybe an additional sentence:

```suggestion
In live and dry-run modes, after the `process_throttle_secs` has passed, Edge will again process `calculate_since_number_of_days` against `minimum_expectancy` to find `min_trade_number`. If no `min_trade_number` is found, the bot will return "whitelist empty". Depending on the trade strategy being deployed, "whitelist empty" may be return much of the time - or *all* of the time. The use of Edge may also cause trading to occur in bursts, though this is rare.

If you encounter "whitelist empty" a lot, condsider tuning `calculate_since_number_of_days`, `minimum_expectancy`  and `min_trade_number` to align to the trading frequency of your strategy.
```

I'm not certain if the double-ticks (``) are fine or cause a problem on readthedocs (the rendered documentation page) - we're usually using single-ticks throughout the documentation (and that's also what markdown shows for inline code segments) - so i'd like to keep it that way.

docs/edge.md
Blackhawke: You are correct! My bad. Thanks for catching that.

PR #3688, opened 2020-08-16
Update stoploss.md

Fredrik81:
## Summary
Updating documentation to simplify examples

### DIFF
--- a/docs/stoploss.md
+++ b/docs/stoploss.md
@@ -6,16 +6,10 @@ For example, value `-0.10` will cause immediate sell if the profit dips below -1
 Most of the strategy files already include the optimal `stoploss` value.
 
 !!! Info
-    All stoploss properties mentioned in this file can be set in the Strategy, or in the configuration. Configuration values will override the strategy values.
+    All stoploss properties mentioned in this file can be set in the Strategy, or in the configuration.  
+    <ins>Configuration values will override the strategy values.</ins>
 
-## Stop Loss Types
-
-At this stage the bot contains the following stoploss support modes:
-
-1. Static stop loss.
-2. Trailing stop loss.
-3. Trailing stop loss, custom positive loss.
-4. Trailing stop loss only once the trade has reached a certain offset.
+## Stop Loss On-Exchange/Freqtrade
 
 Those stoploss modes can be *on exchange* or *off exchange*. If the stoploss is *on exchange* it means a stoploss limit order is placed on the exchange immediately after buy order happens successfully. This will protect you against sudden crashes in market as the order will be in the queue immediately and if market goes down then the order has more chance of being fulfilled.
 
@@ -27,19 +21,58 @@ So this parameter will tell the bot how often it should update the stoploss orde
 This same logic will reapply a stoploss order on the exchange should you cancel it accidentally.
 
 !!! Note
-    Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.
+    Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.  
+    Do not set too low stoploss value if using stop loss on exhange!
 
-## Static Stop Loss
+Stop loss on exchange is controlled with this value (default False):
+
+``` python
+    'stoploss_on_exchange': False
+```
+
+Example from strategy file:
+
+``` python
+order_types = {
+    'buy': 'limit',
+    'sell': 'limit',
+    'stoploss': 'market',
+    'stoploss_on_exchange': True
+}
+```
+
+## Stop Loss Types
+
+At this stage the bot contains the following stoploss support modes:
+
+1. Static stop loss.
+2. Trailing stop loss.
+3. Trailing stop loss, custom positive loss.
+4. Trailing stop loss only once the trade has reached a certain offset.
+
+### Static Stop Loss
 
 This is very simple, you define a stop loss of x (as a ratio of price, i.e. x * 100% of price). This will try to sell the asset once the loss exceeds the defined loss.
 
-## Trailing Stop Loss
+Example of stop loss:
+
+``` python
+    stoploss = -0.10
+```
+
+For example, simplified math:
+* the bot buys an asset at a price of 100$
+* the stop loss is defined at -10%
+* the stop loss would get triggered once the asset dropps below 90$
+
+### Trailing Stop Loss
 
 The initial value for this is `stoploss`, just as you would define your static Stop loss.
 To enable trailing stoploss:
 
 ``` python
-trailing_stop = True
+    stoploss = -0.10
+    trailing_stop = True
 ```
 
 This will now activate an algorithm, which automatically moves the stop loss up every time the price of your asset increases.
@@ -47,35 +80,40 @@ This will now activate an algorithm, which automatically moves the stop loss up 
 For example, simplified math:
 
 * the bot buys an asset at a price of 100$
-* the stop loss is defined at 2%
-* the stop loss would get triggered once the asset dropps below 98$
+* the stop loss is defined at -10%
+* the stop loss would get triggered once the asset dropps below 90$
 * assuming the asset now increases to 102$
-* the stop loss will now be 2% of 102$ or 99.96$
-* now the asset drops in value to 101$, the stop loss will still be 99.96$ and would trigger at 99.96$.
+* the stop loss will now be -10% of 102$ = 91,8$
+* now the asset drops in value to 101$, the stop loss will still be 91,8$ and would trigger at 91,8$.
 
 In summary: The stoploss will be adjusted to be always be 2% of the highest observed price.
 
-### Custom positive stoploss
+### Trailing stop loss, custom positive loss
 
-It is also possible to have a default stop loss, when you are in the red with your buy, but once your profit surpasses a certain percentage, the system will utilize a new stop loss, which can have a different value.
-For example your default stop loss is 5%, but once you have 1.1% profit, it will be changed to be only a 1% stop loss, which trails the green candles until it goes below them.
+It is also possible to have a default stop loss, when you are in the red with your buy (buy - fee), but once you hit possitive result the system will utilize a new stop loss, which can have a different value.
+For example your default stop loss is -10%, but once you have more than 0% profit (example 0.1%) a different trailing stoploss will be used.
 
-Both values require `trailing_stop` to be set to true.
+Both values require `trailing_stop` to be set to true and trailing_stop_positive with a value.
 
 ``` python
-    trailing_stop_positive = 0.01
-    trailing_stop_positive_offset = 0.011
+    stoploss = -0.10
+    trailing_stop = True
+    trailing_stop_positive = 0.02
 ```
 
-The 0.01 would translate to a 1% stop loss, once you hit 1.1% profit.
+For example, simplified math:
+
+* the bot buys an asset at a price of 100$
+* the stop loss is defined at -10%
+* the stop loss would get triggered once the asset dropps below 90$
+* assuming the asset now increases to 102$
+* the stop loss will now be -2% of 102$ = 99,96$
+* now the asset drops in value to 101$, the stop loss will still be 99,96$ and would trigger at 99,96$
+
+The 0.02 would translate to a -2% stop loss.
 Before this, `stoploss` is used for the trailing stoploss.
 
-Read the [next section](#trailing-only-once-offset-is-reached) to keep stoploss at 5% of the entry point.
-
-!!! Tip
-    Make sure to have this value (`trailing_stop_positive_offset`) lower than minimal ROI, otherwise minimal ROI will apply first and sell the trade.
-
-### Trailing only once offset is reached
+### Trailing stop loss only once the trade has reached a certain offset.
 
 It is also possible to use a static stoploss until the offset is reached, and then trail the trade to take profits once the market turns.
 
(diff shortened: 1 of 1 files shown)


docs/stoploss.md
```diff
@@ -27,55 +21,99 @@ So this parameter will tell the bot how often it should update the stoploss orde
...
-## Static Stop Loss
+Example from strategy file:
+
+``` python
+order_types = {
+    'buy': 'limit',
+    'sell': 'limit',
+    'stoploss': 'market',
+    'stoploss_on_exchange': True
+}
```
xmatthias, line 41:
```suggestion
    'stoploss_on_exchange': True,
    'stoploss_on_exchange_interval': 60,
    'stoploss_on_exchange_limit_ratio': 0.99
}
```

For completeness, adding these settings here too... ?

docs/stoploss.md
```diff
@@ -27,55 +21,99 @@ So this parameter will tell the bot how often it should update the stoploss orde
...
+* the stop loss will now be -2% of 102$ = 99,96$
+* now the asset drops in value to 101$, the stop loss will still be 99,96$ and would trigger at 99,96$
 
-!!! Tip
-    Make sure to have this value (`trailing_stop_positive_offset`) lower than minimal ROI, otherwise minimal ROI will apply first and sell the trade.
+The 0.02 would translate to a -2% stop loss.
+Before this, `stoploss` is used for the trailing stoploss.
 
-### Trailing only once offset is reached
+### Trailing stop loss only once the trade has reached a certain offset.
```
xmatthias, line 116:
```suggestion
### Trailing stop loss only once the trade has reached a certain offset
```

I wouldn't use a sentence as title ... (so without trailing .).

docs/stoploss.md
```diff
@@ -87,17 +125,28 @@ This option can be used with or without `trailing_stop_positive`, but uses `trai
...
+For example, simplified math:
+
 * the bot buys an asset at a price of 100$
-* the stop loss is defined at 5%
-* the stop loss will remain at 95% until profit reaches +3%
+* the stop loss is defined at -10%
+* the stop loss would get triggered once the asset dropps below 90$
+* stoploss will remain at 90$ unless asset increases to or above our configured offset
+* assuming the asset now increases to 103$ (where we have the offset configured)
+* the stop loss will now be -2% of 103$ = 100,94$
```
xmatthias, line 145: Maybe we should add a note here that says that the 100.94 are "locked in" ? 

No matter how price-action works - sells should not occur much below 100.94.

docs/stoploss.md
```diff
@@ -6,16 +6,10 @@ For example, value `-0.10` will cause immediate sell if the profit dips below -1
 Most of the strategy files already include the optimal `stoploss` value.
 
 !!! Info
-    All stoploss properties mentioned in this file can be set in the Strategy, or in the configuration. Configuration values will override the strategy values.
+    All stoploss properties mentioned in this file can be set in the Strategy, or in the configuration.  
+    <ins>Configuration values will override the strategy values.</ins>
```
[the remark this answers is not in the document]
Fredrik81: google didn't show me this option

docs/stoploss.md
```diff
@@ -27,55 +21,99 @@ So this parameter will tell the bot how often it should update the stoploss orde
...
+* now the asset drops in value to 101$, the stop loss will still be 91,8$ and would trigger at 91,8$.
 
 In summary: The stoploss will be adjusted to be always be 2% of the highest observed price.
 
-### Custom positive stoploss
+### Trailing stop loss, custom positive loss
 
-It is also possible to have a default stop loss, when you are in the red with your buy, but once your profit surpasses a certain percentage, the system will utilize a new stop loss, which can have a different value.
-For example your default stop loss is 5%, but once you have 1.1% profit, it will be changed to be only a 1% stop loss, which trails the green candles until it goes below them.
+It is also possible to have a default stop loss, when you are in the red with your buy (buy - fee), but once you hit possitive result the system will utilize a new stop loss, which can have a different value.
```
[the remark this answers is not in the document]
Fredrik81: Sure i will add something.
But just for info.. this is the stoploss type i'm using

docs/stoploss.md
Fredrik81: Hmm it's copy paste from the stop loss type list above.. i will check on that

docs/stoploss.md
```diff
@@ -27,55 +21,101 @@ So this parameter will tell the bot how often it should update the stoploss orde
 This same logic will reapply a stoploss order on the exchange should you cancel it accidentally.
 
 !!! Note
-    Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.
+    Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.  
+    Do not set too low stoploss value if using stop loss on exhange!
```
Theagainmen, line 25: I think it should be explained why it's not adviced to use a thight/low stoploss value if using stoploss on exchange. If the user understands how and what, it's easier to actually apply it to your strategy.

docs/stoploss.md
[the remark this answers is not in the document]
Fredrik81: Didn't work?? Shows as bold text only.

docs/stoploss.md
Fredrik81: Did a try on explaining it.. please have a look

docs/stoploss.md
```diff
@@ -6,76 +6,120 @@ For example, value `-0.10` will cause immediate sell if the profit dips below -1
...
 * the bot buys an asset at a price of 100$
-* the stop loss is defined at 2%
-* the stop loss would get triggered once the asset dropps below 98$
+* the stop loss is defined at -10%
+* the stop loss would get triggered once the asset drops below 90$
 * assuming the asset now increases to 102$
-* the stop loss will now be 2% of 102$ or 99.96$
-* now the asset drops in value to 101$, the stop loss will still be 99.96$ and would trigger at 99.96$.
+* the stop loss will now be -10% of 102$ = 91.8$
+* now the asset drops in value to 101$, the stop loss will still be 91.8$ and would trigger at 91,8$.
```
xmatthias, line 90:
```suggestion
* now the asset drops in value to 101$, the stop loss will still be 91.8$ and would trigger at 91.8$.
```

docs/stoploss.md
```diff
@@ -6,76 +6,120 @@ For example, value `-0.10` will cause immediate sell if the profit dips below -1
...
+### Trailing stop loss, custom positive loss
 
-### Custom positive stoploss
+It is also possible to have a default stop loss, when you are in the red with your buy (buy - fee), but once you hit positive result the system will utilize a new stop loss, which can have a different value.
+For example, your default stop loss is -10%, but once you have more than 0% profit (example 0.1%) a different trailing stoploss will be used.
 
-It is also possible to have a default stop loss, when you are in the red with your buy, but once your profit surpasses a certain percentage, the system will utilize a new stop loss, which can have a different value.
-For example your default stop loss is 5%, but once you have 1.1% profit, it will be changed to be only a 1% stop loss, which trails the green candles until it goes below them.
+!!! Note
+    If you want the stoploss to only be changed when you break even of making a profit (what most users want) please refere to next section with [offset enabled](#Trailing-stop-loss-only-once-the-trade-has-reached-a-certain-offset).
```
xmatthias, line 100:
```suggestion
    If you want the stoploss to only be changed when you break even of making a profit (what most users want) please refer to next section with [offset enabled](#Trailing-stop-loss-only-once-the-trade-has-reached-a-certain-offset).
```

docs/stoploss.md
```diff
@@ -6,76 +6,120 @@ For example, value `-0.10` will cause immediate sell if the profit dips below -1
...
-The bot cannot do this every 5 seconds (at each iteration), otherwise it would get banned by the exchange.
+The bot cannot do these every 5 seconds (at each iteration), otherwise it would get banned by the exchange.
 So this parameter will tell the bot how often it should update the stoploss order. The default value is 60 (1 minute).
 This same logic will reapply a stoploss order on the exchange should you cancel it accidentally.
 
 !!! Note
-    Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.
+    Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.  
+    <ins>Do not set too low stoploss value if using stop loss on exhange!</ins>  
+*   If set to low/tight then you have greater risk of missing fill on the order and stoploss will not work
```
xmatthias, line 26:
```suggestion
    Stoploss on exchange is only supported for Binance (stop-loss-limit), Kraken (stop-loss-market) and FTX (stop limit and stop-market) as of now.  
    <ins>Do not set too low stoploss value if using stop loss on exchange!</ins>  
    If set to low/tight then you have greater risk of missing fill on the order and stoploss will not work
```

xmatthias: I do like this :+1:  - it explains the problem with too close stoplosses.

docs/stoploss.md
```diff
@@ -11,32 +11,55 @@ Most of the strategy files already include the optimal `stoploss` value.
...
+    'stoploss_on_exchange_limit_ratio': 0.99
+```
+
+### stoploss_on_exchange
+Enable or Disable stop loss on exchange.
+If the stoploss is *on exchange* it means a stoploss limit order is placed on the exchange immediately after buy order happens successfully. This will protect you against sudden crashes in market as the order will be in the queue immediately and if market goes down then the order has more chance of being fulfilled.
+
+If `stoploss_on_exchange` uses limit orders, the exchange needs 2 prices, the stoploss_price and the Limit price.
+`stoploss` defines the stop-price - and limit should be slightly below this.  
+If an exchange supports both limit and market stoploss orders, then the value of `stoploss` will be used to determine the stoploss type.
```
xmatthias, line 31: I would add the `stoploss_on_exchange_limit_ratio` explanation from below in this area ...
Here, we mention that we may need 2 prices, but not how to define the 2nd price ...

docs/stoploss.md
```diff
@@ -11,32 +11,55 @@ Most of the strategy files already include the optimal `stoploss` value.
...
 So this parameter will tell the bot how often it should update the stoploss order. The default value is 60 (1 minute).
 This same logic will reapply a stoploss order on the exchange should you cancel it accidentally.
 
+### emergencysell and stoploss_on_exchange_limit_ratio
+`emergencysell` is an optional value, which defaults to `market` and is used when creating stop loss on exchange orders fails.
+The below is the default which is used if this is not configured in either strategy or configuration file.
+
+This defaults to 0.99 / 1% (configurable via `stoploss_on_exchange_limit_ratio`).
+Calculation example: we bought the asset at 100$.
+Stop-price is 95$, then limit would be `95 * 0.99 = 94.05$` - so the stoploss will happen between 95$ and 94.05$.
```
xmatthias, line 47: I would split this section - the 2 settings are not really related, and do something quite different.

Best move `stoploss_on_exchange_limit_ratio` to `## stoploss_on_exchange` - where the need for 2 prices is mentioned

Fredrik81: please take a look at the text now

docs/stoploss.md
Fredrik81: please take a look at the text now

docs/stoploss.md
```diff
@@ -22,14 +22,18 @@ These modes can be configured with these values:
...
-### stoploss_on_exchange
+### stoploss_on_exchange and stoploss_on_exchange_limit_ratio
 Enable or Disable stop loss on exchange.
 If the stoploss is *on exchange* it means a stoploss limit order is placed on the exchange immediately after buy order happens successfully. This will protect you against sudden crashes in market as the order will be in the queue immediately and if market goes down then the order has more chance of being fulfilled.
 
 If `stoploss_on_exchange` uses limit orders, the exchange needs 2 prices, the stoploss_price and the Limit price.
 `stoploss` defines the stop-price - and limit should be slightly below this.  
 If an exchange supports both limit and market stoploss orders, then the value of `stoploss` will be used to determine the stoploss type.
 
+If `stoploss_on_exchange` fails to fill we can use `stoploss_on_exchange_limit_ratio` that defaults to 0.99 / 1% to perform an [emergency sell](#emergencysell).
```
xmatthias, line 33: `stoploss_on_exchange_limit_ratio` has nothing to do with emergency sell... 

A limit stoploss order requires 2 prices - a stop-price (when the order will be placed) - and a limit price (up to which point the order can be filled).
Using the below numbers - your stop-price would be 95$ (so the order is placed once price crosses below 95$) and the limit price is 94.05$ (1% below 95%) - which means that the limit sell order will be placed at  94.05$. 
Most likely, the exchange will fill it right away (assuming the price sits between 95$ and 94.05$), unless price dips just for a microsecond - and goes above 95$ again, where the average price will be > stoploss (but that's not a problem, really).

This will protect you from VERY fast drops - as it'll guarantee you a exit price - but obviously you can remain stuck in a trade if price drops "past" your stoploss before it can be sold (although i don't think that's possible on binance - as binance would refuse such a limit order with "would trigger immediately" - which is when emergency sell comes into play - which is using the current market price to price the order (just like a regular sell order would) - but defaults to a market order ...


Kraken does not have any of this - but uses stoploss-market orders - which means once the order is placed (when price crosses below 95$) - it'll place a market sell order - which obviously fill at "whatever" price it finds volume in the orderbook.
Therefore, `stoploss_on_exchange_limit_ratio` will be ignored for stoploss-market order types

PR #3705, opened 2020-08-23
update bad exchanges

mschultheiss83:
added phemex to bad exchanges
reason: `"Could not fetch historical candle (OHLCV) data for pair BTC/EUR due to ArgumentsRequired. Message: phemex fetchOHLCV requires a since argument, or a limit argument, or both"`

### DIFF
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -16,6 +16,7 @@ BAD_EXCHANGES = {
                 "Details in https://github.com/freqtrade/freqtrade/issues/1983",
     "hitbtc": "This API cannot be used with Freqtrade. "
               "Use `hitbtc2` exchange id to access this exchange.",
+    "phemex": "Does not provide history. ",
     **dict.fromkeys([
         'adara',
         'anxpro',


xmatthias:
From a technical point of view, this is perfectly fine.

Knowing however that you also work on getting this exchange working - would you rather want me to wait before merging this - or shall we merge this as a "intermediate" solution - and the "get it working" change will remove this line again?

mschultheiss83:
yes please, merge this as a "intermediate" solution, and "get it working" change will remove this line again

xmatthias (approved):
LGTM 👍 

Thanks for testing additional exchanges!

moinonin:
Hey @xmatthias is there a list of tested (and working) exchanges other than Bittrex, FTX, Binance and Krakken?

xmatthias:
These 4 are the ones officially tested and supported.

Others may or may not work - we add them to the list of non-working if we discover they are not working - however testing these is very time intensive and requires an account on each of the exchanges - so we don't have the resources to test more exchanges.

PR #3744, opened 2020-09-02
fix Informative pair documentation

xmatthias:
## Summary
Fix informative sample documentation, which had a lookahead bias

## Quick changelog

- Expose some helpers through freqtrade.strategy
- Move date by one timeunit to account for all dates beeing candle open dates

### DIFF
--- a/freqtrade/strategy/strategy_helper.py
+++ b/freqtrade/strategy/strategy_helper.py
@@ -26,7 +26,8 @@ def merge_informative_pair(dataframe: pd.DataFrame, informative: pd.DataFrame,
     """
 
     minutes_inf = timeframe_to_minutes(timeframe_inf)
-    if timeframe == timeframe_inf:
+    minutes = timeframe_to_minutes(timeframe)
+    if minutes >= minutes_inf:
         # No need to forwardshift if the timeframes are identical
         informative['date_merge'] = informative["date"]
     else:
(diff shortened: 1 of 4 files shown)


jblestang (commented):
Looks good to me, just need to hide some of the details of implementation so that a newcomer would only need to call one function to merge the dataframes without causing repainting during backtesting

freqtrade/strategy/strategy_helper.py
```diff
@@ -0,0 +1,47 @@
...
+
+    :param dataframe: Original dataframe
+    :param informative: Informative pair, most likely loaded via dp.get_pair_dataframe
+    :param timeframe: Timeframe of the original pair sample.
+    :param timeframe_inf: Timeframe of the informative pair sample.
+    :param ffill: Forwardfill missing values - optional but usually required
+    """
+
+    minutes_inf = timeframe_to_minutes(timeframe_inf)
+    if timeframe == timeframe_inf:
```
xmatthias, line 29: Should this comparison be `<=` ? 

I guess shifting the candle does not make sense if the informative timeframe is smaller (e.g. merging 5m candles on top of 15m candles


i should be able to use the candle from 10:00 for both 5m and 15m - as the 5m candle stops at 10:05, while the 15m candle stops at 10:15 - so no lookahead bias (if the 15m candle is "My" timeframe).

Issue #2487, opened 2019-11-06
Reorganisation of the freqtrade install doc. (discussion)


gaugau3000:
The end user when discovering a trading bot like freqtrade is mostly i think (don't work for advanced algo trader users of course).

1/ Ok let's install quickly this bot (it is time to make money :-) ).
2/ Let's download minimum of data to be able to run a back-test.
3/ now test a default strategy -> not very good results
4/ take the freqtrade-strategies -> with some tunes i can do something
5/ lets go in dry run or i am a greedy person and put money in an average strat :-)
...

I think the installation part should reflect a kind of this user experience but : 
-> we start to say he have to create api keys as prerequisites (not needed as the beginning i think)
-> 5 langages/sofware requirements (for a newbie the learning curve seems like the game of trone wall :-) ).

I think the install doc page is too heavy and have to split in sub-pages.

I can try to reorganized based on this kind of user experience + try to make sub-pages + advice docker as a out of the box solution for install.

Tell me what you think about this (i think all the doc should kind of follow the step by step evolution in learning freqtrade as the doc quality is very high maybe it just reorganize menu, items.. but it is another topic).

Have a nice day.

xmatthias:
I can only agree...  i'm not happy with our doc organization either, since it contains duplications in some parts, and is difficult to read for non-technical users to get started.

And honestly, really appreciate you giving this some thought, and offering to try and clean the documentation structure up! 👍 


It's probably best to split this out into a "quick start" (?) section (which is obviously referenced by both `readme.md` files) - and keep (at least parts) of the existing installation guides as "custom installation" methods.

I also absolutely agree about the API keys ... this should probably be moved to `configuration.md` - right next to [switch to production mode](https://www.freqtrade.io/en/latest/configuration/#switch-to-production-mode) - since it's only ever relevant then - not while you're still getting setup and try out dry-run.

We should also change the default for telegram from True to False in the example config - so users need to "opt in" and are not faced with an error if telegram is not properly configured - basically allowing users to "get going" by just selecting a strategy, pairs and an exchange.

gaugau3000:
It's probably best to split this out into a "quick start" (?) section (which is obviously referenced by both readme.md files) - and keep (at least parts) of the existing installation guides as "custom installation" methods.

> I agree

I also absolutely agree about the API keys ... this should probably be moved to configuration.md - right next to switch to production mode - since it's only ever relevant then - not while you're still getting setup and try out dry-run.

> Yes

[...]

PR #3745, opened 2020-09-02
Updating Edge Positioning Doc.

silvavn:
## Summary
Improve Edge Positioning Documentation

## What's new?

1. Integrated MathJax
2. Included worked out examples
3. Changed Language to achieve a middle ground.
4. Minor formatting improvements

In a second commit I have addressed discussions raised in issue #2487 trying to make it easier to the user to navigate installation

1. Broke down docker with and without composer. 
2. Separated Windows from the other OS installation instructions
3. Included tab package for improved and directed instructions.


silvavn:
When I included the tabs package the docs Linter does not seem to like notes embedded into tabs.

docs/installation.md
```diff
@@ -89,36 +103,44 @@ OS Specific steps are listed first, the [Common](#common) section below is neces
 !!! Note
     Python3.6 or higher and the corresponding pip are assumed to be available.
 
-### Linux - Ubuntu 16.04
+=== "Ubuntu 16.04"
```
xmatthias, line 106: How wold we expect this to render / are you sure this renders correctly in the readthedocs context (if yes - got any samples on how this looks like when deployed to readthedocs?)

mkdocs.yml
```diff
@@ -1,8 +1,13 @@
 site_name: Freqtrade
 nav:
     - Home: index.md
-    - Installation Docker: docker.md
-    - Installation: installation.md
+    - Installation Docker: 
+        - Quickstart: docker.md
+        - Freqtrade with Docker Compose (Advanced): docker_compose.md
+        - Freqtrade without docker-compose: without_docker_compose.md
```
xmatthias, line 7: I would see the compose variant as the simple one. 

Running docker without compose is quite complex, as you need to have the whole command identical for every command you run to not mess everything up.

docs/installation.md
silvavn: This renders perfectly fine:

Sample1:
image

Sample2:
image


The goal is to provide Platform specific instructions

mkdocs.yml
silvavn: The `(Advanced)` contains only the advanced commands used in docker compose.

My current proposed structure is:

- Quickstart - a fast installation with docker composed (easy)
- Advanced compose - Compose, but for advanced users (optional)
- Docker without compose [hard]

docs/installation.md
```diff
@@ -89,36 +103,44 @@ OS Specific steps are listed first, the [Common](#common) section below is neces
...
 
-Tested using a Raspberry Pi 3 with the Raspbian Buster lite image, all updates applied.
+    bash setup.sh -i
+    ```
 
-``` bash
-sudo apt-get install python3-venv libatlas-base-dev
-git clone https://github.com/freqtrade/freqtrade.git
-cd freqtrade
+    !!! Note "Installation duration"
```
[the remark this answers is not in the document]
silvavn: It does render correctly and it works as expected. What it does is simple put the admonition under a label:

## Sample1:

The node is under the anaconda label. compare to sample 2

image


## Sample2:

When we pick the raspberry label there are more platform specific instructions and notes

image

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
+!!! Example
+    In a section where a strategy made three transactions $O = \{3.5, -1, 15, 0\}$:<br>
+    $T_{win} = \{3.5, 15\}$<br>
+    $T_{lose} = \{-1, 0\}$<br>
+
+### Win Rate and Lose Rate
+
+The win rate $W$ is the proportion of winning trades with respect to all the trades made by a strategy. We use the following function to compute the win rate:
+
+$$W = \frac{\sum^{o \in T_{win}} o}{N}$$
```
xmatthias, line 66: Are you sure this (and the Corresponding L formulae) are correct?
the sum symbol seems odd to me - it's not the sum of trades (`sum(winning_trades)`) - but the count (`len(winning_trades)`) that's calculated here, especially considering that the same "upper part" is used for average_profit - but there it's really the sum of profits - not the count of trades.

But i'm no mathematician, so this syntax is ... odd for me.

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
+$$ 
+    L = \frac{\sum^{o \in T_{lose}} o}{N} 
+$$
+
+Where $L$ is the lose rate, $N$ is the amount of trades made and, $T_{lose}$ is the set of all trades where the strategy lost money. Note that the above formula is the same as calculating $L = 1 – W$ or $W = 1 – L$
 
 ### Risk Reward Ratio
 
-Risk Reward Ratio (*R*) is a formula used to measure the expected gains of a given investment against the risk of loss. It is basically what you potentially win divided by what you potentially lose:
+Risk Reward Ratio (*R*) is a formula used to measure the expected gains of a given investment against the risk of loss. It is basically what you potentially win divided by what you potentially lose. Formally:
```
xmatthias, line 80:
```suggestion
Risk Reward Ratio ($R$) is a formula used to measure the expected gains of a given investment against the risk of loss. It is basically what you potentially win divided by what you potentially lose. Formally:
```
I guess you'd want to use `$R$` here ... to have this aligned to the rest of the document?

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
 
-```
-R = Profit / Loss
-```
+$$ R = \frac{\text{potential_profit}}{\text{potential_loss}} $$
 
-Over time, on many trades, you can calculate your risk reward by dividing your average profit on winning trades by your average loss on losing trades:
+??? Example "Worked example of $R$ calculation"
+    Let's say that you think that the price of *stonecoin* today is $10.0. You believe that, because they will start mining stonecoin it will go up to $15.0 tomorrow. There is the risk that the stone is too hard, and the GPUs can't mine it, so the price might go to $0 tomorrow. You are planning to invest $100.<br>
```
xmatthias, line 85:
```suggestion
    Let's say that you think that the price of *stonecoin* today is $10.0. You believe that, because they will start mining stonecoin, it will go up to $15.0 tomorrow. There is the risk that the stone is too hard, and the GPUs can't mine it, so the price might go to $0 tomorrow. You are planning to invest $100.<br>
```

i think you'll have to close the side-sentence with another semicolon ... ?

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
+    $\begin{aligned}
+        R   &= \frac{\text{potential_profit}}{\text{potential_loss}}\\
+            &= \frac{33.33}{100}\\
+            &= 0.333...
+    \end{aligned}$<br>
+    What it effectivelly means is that the strategy have the potential to make $0.33 for each $1 invested. 
 
-```
-Average profit = (Sum of profits) / (Number of winning trades)
+On a long horizonte, that is, on many trades, we can calculate the risk reward by dividing the strategy' average profit on winning trades by the strategy' average loss on losing trades. We can calculate the average profit, $\mu_{win}$, as follows:
```
xmatthias, line 100:
```suggestion
On a long horizon, that is, on many trades, we can calculate the risk reward by dividing the strategy' average profit on winning trades by the strategy' average loss on losing trades. We can calculate the average profit, $\mu_{win}$, as follows:
```

horizonte is the right word - but i don't think it's english, but spanish/italian ...

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
 
-Average loss = (Sum of losses) / (Number of losing trades)
+$$ \text{average_profit} = \mu_{win} = \frac{\text{sum_of_profits}}{\text{count_winning_trades}} = \frac{\sum^{o \in T_{win}} o}{|T_{win}|} $$
+
+Similarly, we can calculate the average loss, $\mu_{lose}$, as follows:
+
+$$ \text{average_loss} = \mu_{lose} = \frac{\text{sum_of_losses}}{\text{count_losing_trades}} = \frac{\sum^{o \in T_{lose}} o}{|T_{lose}|} $$
+
+Finally, we can calculate the Risk Reward ratio as follows:
```
xmatthias, line 108:
```suggestion
Finally, we can calculate the Risk Reward ratio, $R$, as follows:
```

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
+    $E = R * W - L = 5 * 0.28 - 0.72 = 0.68$
+    <br>
 
-```
-Expectancy = (5 X 0.28) – 0.72 = 0.68
-```
+The expectancy worked out in the example above means that, on average, this strategy' trades will return 1.68 times the size of its losses. Said another way, the strategy makes $1.68 for every $1 it loses, on average. 
 
-Superficially, this means that on average you expect this strategy’s trades to return 1.68 times the size of your loses. Said another way, you can expect to win $1.68 for every $1 you lose. This is important for two reasons: First, it may seem obvious, but you know right away that you have a positive return. Second, you now have a number you can compare to other candidate systems to make decisions about which ones you employ.
+You canThis is important for two reasons: First, it may seem obvious, but you know right away that you have a positive return. Second, you now have a number you can compare to other candidate systems to make decisions about which ones you employ.
```
xmatthias, line 132:
```suggestion
This is important for two reasons: First, it may seem obvious, but you know right away that you have a positive return. Second, you now have a number you can compare to other candidate systems to make decisions about which ones you employ.
```

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
 
 It is important to remember that any system with an expectancy greater than 0 is profitable using past data. The key is finding one that will be profitable in the future.
 
 You can also use this value to evaluate the effectiveness of modifications to this system.
 
-**NOTICE:** It's important to keep in mind that Edge is testing your expectancy using historical data, there's no guarantee that you will have a similar edge in the future. It's still vital to do this testing in order to build confidence in your methodology, but be wary of "curve-fitting" your approach to the historical data as things are unlikely to play out the exact same way for future trades.
+**NOTICE:** It's important to keep in mind that Edge is testing your expectancy using historical data, there's no guarantee that you will have a similar edge in the future. It's still vital to do this testing in order to build confidence in your methodology but be wary of "curve-fitting" your approach to the historical data as things are unlikely to play out the exact same way for future trades.
```
xmatthias, line 138:
```suggestion
!!! Note
    It's important to keep in mind that Edge is testing your expectancy using historical data, there's no guarantee that you will have a similar edge in the future. It's still vital to do this testing in order to build confidence in your methodology but be wary of "curve-fitting" your approach to the historical data as things are unlikely to play out the exact same way for future trades.
```

docs/edge.md
```diff
@@ -99,13 +148,13 @@ Edge combines dynamic stoploss, dynamic positions, and whitelist generation into
 | XZC/ETH  |  -0.03        |   0.52       |1.359670           | 0.228      |
 | XZC/ETH  |  -0.04        |   0.51       |1.234539           | 0.117      |
 
-The goal here is to find the best stoploss for the strategy in order to have the maximum expectancy. In the above example stoploss at 3% leads to the maximum expectancy according to historical data.
+The goal here is to find the best stoploss for the strategy in order to have the maximum expectancy. In the above example stoploss at $3% $leads to the maximum expectancy according to historical data.
```
xmatthias, line 151:
```suggestion
The goal here is to find the best stoploss for the strategy in order to have the maximum expectancy. In the above example stoploss at $3%$ leads to the maximum expectancy according to historical data.
```

mkdocs.yml
```diff
@@ -1,8 +1,13 @@
 site_name: Freqtrade
 nav:
     - Home: index.md
-    - Installation Docker: docker.md
-    - Installation: installation.md
+    - Installation Docker: 
+        - Quickstart: docker.md
```
xmatthias, line 5: Should this quickstart be outside of the docker installation subsection?

otherwise ALL installation pages are in subsections - making it a 2-click navigation to get to any installation page.

docs/docker.md
```diff
@@ -17,325 +17,96 @@ Once you have Docker installed, simply prepare the config file (e.g. `config.jso
...
+    ```
 
-The above snippet creates a new directory called "ft_userdata", downloads the latest compose file and pulls the freqtrade image.
-The last 2 steps in the snippet create the directory with user-data, as well as (interactively) the default configuration based on your selections.
+=== "RaspberryPi"
+    ``` bash
+    mkdir ft_userdata
+    cd ft_userdata/
+    # Download the docker-compose file from the repository
+    curl https://raw.githubusercontent.com/freqtrade/freqtrade/master_pi/docker-compose.yml -o docker-compose.yml
```
xmatthias, line 50: This seperation for docker-compose seems irrelevant - the only change is the docker-compose url - which points to the brach master_pi - which is not a branch on the repository.

The real change for the PI is within the docker-compose image, changing 
```
    image: freqtradeorg/freqtrade:master
```

to 
```
    image: freqtradeorg/freqtrade:master_pi
    # image: freqtradeorg/freqtrade:develop_pi
```

Maybe using a `!!! Note` below this section to point the RPI difference out would work better.

docs/docker.md
```diff
@@ -17,325 +17,96 @@ Once you have Docker installed, simply prepare the config file (e.g. `config.jso
...
-To update freqtrade when using docker-compose is as simple as running the following 2 commands:
+To update freqtrade when using `docker-compose` is as simple as running the following 2 commands:
 
-``` bash
-# Download the latest image
-docker-compose pull
-# Restart the image
-docker-compose up -d
-```
+=== "Docker Compose"
```
xmatthias, line 100: I'm still indecisive if i like this or not ... it looks good on the one hand - but the commands it contains are really just shell commands, not compose commands.

mkdocs.yml
xmatthias: After having a look at the Quickstart / advanced compose sections, i think the 2 should really be combined.

They're not long enough to really justify seperation - and if you look at the docker-compose file we ship - EVERYONE will have to modify it at least in part as it hardcodes the `SampleStrategy`.

docs/docker.md
```diff
@@ -17,325 +17,96 @@ Once you have Docker installed, simply prepare the config file (e.g. `config.jso
...
+
+    # Create configuration - Requires answering interactive questions
+    docker-compose run --rm freqtrade new-config --config user_data/config.json
+    ```
+
+The above snippet creates a new directory called `ft_userdata`, downloads the latest compose file and pulls the freqtrade image.
+The last 2 steps in the snippet create the directory with `user_data`, as well as (interactively) the default configuration based on your selections.
+
+!!! Question "How to edit the bot configuration?"
     You can edit the configuration at any time, which is available as `user_data/config.json` (within the directory `ft_userdata`) when using the above configuration.
```
xmatthias, line 66: Maybe we should have another Question / option / whatever here pointing to the "edit compose-file" section ... 
The default is designed to work out of the box (defines SampleStrategy as --strategy) - but that'll not do it 99.99% of the users.

See also the note on the mkdocs.yml about combining this file with the compose  advanced section

docs/windows_installation.md
```diff
@@ -0,0 +1,49 @@
...
+If that is not possible, try using the Windows Linux subsystem (WSL) - for which the Ubuntu instructions should work.
+Otherwise, try the instructions below.
+
+## Install freqtrade manually
+
+!!! Note
+    Make sure to use 64bit Windows and 64bit Python to avoid problems with backtesting or hyperopt due to the memory constraints 32bit applications have under Windows.
+
+!!! Hint
+    Using the [Anaconda Distribution](https://www.anaconda.com/distribution/) under Windows can greatly help with installation problems. Check out the [Conda section](#using-conda) in this document for more information.
```
xmatthias, line 12: the Conda link should be adjusted to link to it's correct new place ...

docs/windows_installation.md
```diff
@@ -0,0 +1,49 @@
...
+
+```bash
+git clone https://github.com/freqtrade/freqtrade.git
+```
+
+### 2. Install ta-lib
+
+Install ta-lib according to the [ta-lib documentation](https://github.com/mrjbq7/ta-lib#windows).
+
+As compiling from source on windows has heavy dependencies (requires a partial visual studio installation), there is also a repository of unofficial precompiled windows Wheels [here](https://www.lfd.uci.edu/~gohlke/pythonlibs/#ta-lib), which needs to be downloaded and installed using `pip install TA_Lib‑0.4.18‑cp38‑cp38‑win_amd64.whl` (make sure to use the version matching your python version)
```
xmatthias, line 24: Note to self:
this section should be updated - the wheel files are included in the build-helper directory, so seperate downloading is not necesary

docs/installation.md
```diff
@@ -79,6 +79,20 @@ This option will hard reset your branch (only if you are on either `master` or `
 
 DEPRECATED - use `freqtrade new-config -c config.json` instead.
 
+### MacOS installation error
```
xmatthias, line 82: I would move this section to the bottom of this document.

It's trouble-shooting, really - and most likely irrelevant to most users.

docs/installation.md
```diff
@@ -89,36 +103,44 @@ OS Specific steps are listed first, the [Common](#common) section below is neces
...
 
-!!! Note
-    The above does not install hyperopt dependencies. To install these, please use `python3 -m pip install -e .[hyperopt]`.
-    We do not advise to run hyperopt on a Raspberry Pi, since this is a very resource-heavy operation, which should be done on powerful machine.
+    !!! Note
+        This requires the [ta-lib](#1-install-ta-lib) C-library to be installed first. See below.
+    
+    ``` bash
+    conda env create -f environment.yml
+    ```
```
xmatthias, line 143: This section is kindof wrong here (the header is "necessary dependencies").

Instead, the conda approach installs "everything" - including freqtrade, so all above "common" steps are not necessary when taking this approach (should not be necessary - been a while since conda installation has been tested i guess).

docs/installation.md
```diff
@@ -169,11 +191,6 @@ Clone the git repository:
 ```bash
 git clone https://github.com/freqtrade/freqtrade.git
 cd freqtrade
-```
-
-Optionally checkout the master branch to get the latest stable release:
```
xmatthias, line 174: I would've left this (maybe as comment within the code-block).

changing to master is not a requirement, as also develop is kept in a stable state.

xmatthias (commented):
While it will seem like a ton of comments - but many are really just typos / small fixes which can be applied by using the "add to commit" button in the files view if you agree with them.

I really love what you did to the Edge page - and also splitting the installation documentation makes sense (but i'd keep the docker-compose stuff together).

mkdocs.yml
silvavn: I agree. Joined Quickstart with docker and moved out of installation.

docs/docker.md
silvavn: Changed to a note as suggested.

docs/docker.md
silvavn: I have removed the label on this.

docs/docker.md
silvavn: Added in the same question.

docs/windows_installation.md
silvavn: I will let you do your thing here because I tried to install on Windows and "Oh boy..."

docs/installation.md
silvavn: Done. In the future this could be a footnote. It looks weird that after we present the conclusion of the page and direct the user to another page we are adding troubleshooting.

docs/installation.md
silvavn: Moved Anaconda to the original spot.

docs/edge.md
```diff
@@ -1,92 +1,141 @@
...
-The question is: How do you calculate that? How do you know if you wanna play?
+!!! Question "Which trade is a better option?"
+    a) A trade with 80% of chance of losing $100 and 20% chance of winning $200<br/>
+    b) A trade with 100% of chance of losing $30
 
-The answer comes to two factors:
+??? Info "Answer"
+    The expected value of *a)* is smaller than the expected value of *b)*.<br/>
+    Hence, *b*) represents a smaller loss in the long run.<br/>
+    However, the answer is: *it depends*
```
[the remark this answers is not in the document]
silvavn: Expanded all questions with `???+`

docs/edge.md
silvavn: You are absolutely correct. I had made `o` as a logical notation before, but noticed that summations were used further in the document. So with a logical notation it would be `o*v` where `o \in {0,1}`, but I figured it would be confusing to the user. I have changed the formulas to reflect the size of the sets.

docs/installation.md
xmatthias: on the one hand it is (contains latest editions, ...) - but on the other it's pretty stable nonetheless - as it's the base for the release (almost always without change other than the version number).

docs/installation.md
xmatthias: yeah, i've moved this o top of the footnote now - it used to be a "troubleshooting" section ...

xmatthias (approved):
I've done a few small changes now - renaming the files (docker.md should be the quickstart guide - so the path <documentation>/docker does point to the correct thing).

I've also fixed the windows part.

Really great job! :+1:

PR #3764, opened 2020-09-09
Added full virtualenv command to quick start

Blackhawke:


docs/installation.md
```diff
@@ -63,7 +63,7 @@ With this option, the script will install the bot and most dependencies:
 You will need to have git and python3.6+ installed beforehand for this to work.
 
 * Mandatory software as: `ta-lib`
-* Setup your virtualenv under `.env/`
+* Setup your virtualenv: `source .env/bin/activate`
```
xmatthias, line 66: I'm not sure about this - this is really a description on how the `setup.sh` script works / what it does (which sets up the virtual environment under `.env/`).

Yes, you need to activate the environment after running the setup script - but that's part of the output of the setup.sh command.

Any particular reason you think this is needed right in this location of the docs?

Blackhawke: I've heard it called the rule of three. In speech writing they tell you to tell them what you're going to tell them, then tell them, then tell them what you told them. It's something like that here. I *completely* missed the output of the script when I set up the bot; and the first complete command line isn't shown until clear at the bottom of the next section, in custom install. That's where I found it. So for clarity,  I made the implied reference explicit. If the command in the output is caught, fine. If not, it's right there, in quick install, for reference.

xmatthias: I don't argue the need to add something around "you need to activate your virutal environment via ... before usage" to this (easy installation) part of the documentation.

I just think that these 2 lines explain WHAT is installed - replacing what with how to progress from that point seems to sacrifice one good information for something else (while new lines don't cust us much) :laughing:

Blackhawke: Good point. Try it this way then. I put the original text back and added the `source` command to the bottom of the easy install section as the final instruction.

xmatthias (approved):
LGTM :+1: 

Thanks a lot!

PR #3771, opened 2020-09-11
FIX: Docs: added missing ( to SQLite insert statement to avoid error.

caudurodev:

N/A

## Summary

FIX: Docs - added missing ( to SQLite insert statement to avoid error.

## Quick changelog

- Docs code update

## What's new?

Doc update:
https://www.freqtrade.io/en/latest/sql_cheatsheet/#table-structure

### DIFF
--- a/docs/sql_cheatsheet.md
+++ b/docs/sql_cheatsheet.md
@@ -46,8 +46,7 @@ sqlite3
 ### Trade table structure
 
 ```sql
-CREATE TABLE trades
-(
+CREATE TABLE trades(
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
Nice catch 👍 LGTM 👍 

You should not have to run this statement though - freqtrade creates all tables it needs on it's own when starting up.

Also, it's not "always" 100% uptodate with the database (and i'm also considering removing this in the future, as i don't think the create table statement provides any benefit to any casual documentation reader).

PR #3799, opened 2020-09-22
Introduce Telegram /stats endpoint

### LINKED ISSUE #3783: Telegram Recommendations
## Describe your environment
(if applicable)

Operating system: UBUNTU 18.04
Python Version: 3.6.9 (python -V)
CCXT version: ccxt==1.33.72 (pip freeze | grep ccxt)
Freqtrade Version: develop-10c5b230 (freqtrade -V or docker-compose run --rm freqtrade -V for Freqtrade running in docker)

## Describe the enhancement

Backtesting is very important when making a strategy. As well, Live and Dry testing are more important and I think it would be really cool if Telegram had a stat box for sell reasons during testing mode like this. It could also help in determine what the strategy needs fixing.

/sellreason

Total Trades: 100
ROI Trades: 45
Stoploss Trades: 15
Trailing Stoploss: 5
Forcesell Trades: 10
Sell Signal Trades: 25

Another recommendation that would be nice to add is an Average Duration stat box because there is nothing more worrisome than wondering if the trade is going to win, lose, or draw. Having an Average Duration for telegram trades can make the user feel comfortable knowing how long a winning or losing trade should take based on the stats.

/duration

Avg. Duration Winners: 1:55:00
Avg. Duration Losers: 6 days, 21:07:00
Avg. Duration Draws: 4 days, 09:05:00

imxuwang:
## Summary
Add `/stats` command in Telegram

closes #3783

## Quick changelog

WIP


freqtrade/rpc/telegram.py
```diff
@@ -738,6 +739,50 @@ def _version(self, update: Update, context: CallbackContext) -> None:
...
+        for trade in trades_closed:
+            if trade['sell_reason'] in sell_reasons:
+                sell_reasons[trade['sell_reason']][trade_win_loss(trade)] += 1
+            else:
+            win_loss_count = {'Wins': 0, 'Losses': 0, 'Draws': 0}
+            win_loss_count[trade_win_loss(trade)] += 1
+            sell_reasons[trade['sell_reason']] = win_loss_count
+        sell_reason_msg = [
+            '| Sell Reason | Sells | Wins | Draws | Losses |',
+            '|-------------|------:|-----:|------:|-------:|'
```
xmatthias, line 774: You should be able to use `tabulate()` for this.

It'll do a much better job at formatting a table than we could... (we're using it for `_status_table()` for example).

imxuwang:
@xmatthias Could you suggest on how to solve the failed tests (on My py)?

xmatthias:
You should be able to run mypy locally with `mypy freqtrade` - this will help you find if the error was solved.

The message suggests that the type for `dur` is not clear - i assume something like `dur: Dict[str, Any] = {...}` should fix that - but you'll see that after running `mypy freqtrade` again...

xmatthias:
> imxuwang marked this pull request as ready for review 16 days ago

I don't think this is ready.

I've just given it a try - and whenever i call `/stats`, it just crashes with a type error.
Tests  won't show this as this endpoint is not covered / called during tests yet.

I've fixed the merge conflicts, but that doesn't change the problem above, as it's non-functional at the moment

xmatthias:
I did add a few commits to this PR so it's at least functional.

It's still a bit of work to make the code "nicer" before we can merge this - but i hope this should help some ..

xmatthias (approved):
LGTM :+1: 

Thanks a lot @imxuwang for this! :+1:

PR #3842, opened 2020-10-08
Fix example R calculation in edge documentation

xmatthias:
## Summary
Fix example R calculation in edge documentation (as pointed out on slack, the old calculation has at least one typo in the "number example")

We should not calculate with 100% loss. However, the old formulae seems incorrect in total.

@silvavn i would appreciate if you could have a look at this.
Adding this example was your work initially (#3745) - so maybe i'm completely wrong on this and did not understand something...

The formulas used now are from [this page](https://analyzingalpha.com/risk-reward-ratio-for-stocks#:~:text=The%20risk%2Dreward%20ratio%20measures,risking%20%241%20to%20make%20%242.) - but other pages seem to use the same formula.

## Whats new
To have a reference on the changes (since this is hard to read in diffs, and needs rendering):

2020-10-08-102441_779x629_scrot

Old version:
2020-10-08-081320_769x437_scrot

### DIFF
--- a/docs/edge.md
+++ b/docs/edge.md
@@ -82,20 +82,33 @@ Risk Reward Ratio ($R$) is a formula used to measure the expected gains of a giv
 $$ R = \frac{\text{potential_profit}}{\text{potential_loss}} $$
 
 ???+ Example "Worked example of $R$ calculation"
-    Let's say that you think that the price of *stonecoin* today is $10.0. You believe that, because they will start mining stonecoin, it will go up to $15.0 tomorrow. There is the risk that the stone is too hard, and the GPUs can't mine it, so the price might go to $0 tomorrow. You are planning to invest $100.<br>
-    Your potential profit is calculated as:<br>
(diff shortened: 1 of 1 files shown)


silvavn:
Seems correct. Small fix is that the potential profit is of $ 3.33 to $1 (not 3 to 1). Also, its confusing when the mash shows 10-(10*(1-0.15)) when it just says stop_loss. A more straightforward presentation or explanation would be nice.

xmatthias:
Good point! I've fixed this now (and updated the image in the first post)

silvavn:
On a second thought... This is wrong...  Your potential loss can't be larger than your investment (unless you are leveraged). In the first image the potential loss is of 150, when the investment is 100. Instead of using investment it should be really be share count...
See this: https://www.investopedia.com/terms/r/riskrewardratio.asp AND https://www.investopedia.com/articles/stocks/11/calculating-risk-reward.asp

xmatthias:
Great catch! 

I've adjusted it now mostly back to how it was.
Just added the loss side - as i don't think a loss of 100% is realistic given we recommend the use of stoploss.

PR #3845, opened 2020-10-08
Backtesting should not double-loop for sell signals

xmatthias:
## Summary
Current behaviour in backtesting is to "double-loop" once a trade is found.

The bot will loop over every candle and every pair 
* once a buy signal is found 
* copy all candles from this point onwards
* try and find a sell signal

It'll then lock this pair until the sell-signal's candle - and continue searching for trades.

This "branching out" requires quite complex locking logic to avoid buying the same pair twice, and is unnecessary.

This also gives the benefit of having simpler logic, making jumping into the backtesting code a lot simpler (and a lot less risky to break stuff).

Not "branching out" also does not require copying the data another time, which will be beneficial for speed (32.9s vs. 36.26s in a 256 day 5m backtest with ~20 pairs).
While the speedup is not huge for backtesting (roughly ~8%), it will make a huge difference for hyperopt (where backtesting is running many times).

## Quick changelog

- Don't "branch out" in backtesting
- Improve Type hints
- Switch to regular tuples (from named tuples previously) to improve speed

## Seperate note:
This refactoring / improvement led to the identification of a different bug, which leads to non-matching results between develop and this version.
An explanation will be provided in the comment below.


freqtrade/optimize/backtesting.py
```diff
@@ -336,37 +327,50 @@ def backtest(self, processed: Dict, stake_amount: float,
...
-                if (not position_stacking and pair in lock_pair_until
-                        and row.date <= lock_pair_until[pair]):
-                    # without positionstacking, we can only have one open trade per pair.
-                    continue
-
-                if max_open_trades > 0:
-                    # Check if max_open_trades has already been reached for the given date
-                    if not trade_count_lock.get(row.date, 0) < max_open_trades:
-                        continue
-                    trade_count_lock[row.date] = trade_count_lock.get(row.date, 0) + 1
```
xmatthias, line 351: This line in the prior backtesting causes misalignment in certain backtests.
trade_count_lock[row.date] is incremented again within `_get_sell_trade_entry()` for all candles the trade is open, including the candle the trade was entered.

This is a miss from a previous improvement.

**Impact:**
The last trade-slot may not be filled if the last 2 slots are filled in the same candle.

Assuming a max_open_trades of 3, and 1 open trade with BTC/USDT when we come to the candle at 10:00.
candle 10:00 buys ETH/USDT, and has a buy signal for XRP/USDT.
the XRP trade (assuming int's last in the pairlist) will not be opened, since trade_count_lock is incremented twice for this candle.
A signal at the next candle would enter this trade again - but this can cause misalignments.

To test alignment, best remove this line (removing this line does not break any test).
max_open_trades will not be violated (tested in `test_backtest_multi_pair()`).

freqtrade/optimize/backtesting.py
```diff
@@ -336,37 +327,50 @@ def backtest(self, processed: Dict, stake_amount: float,
...
+                        amount=round(stake_amount / row.open, 8),
+                        fee_open=self.fee,
+                        fee_close=self.fee,
+                        is_open=True,
+                    )
+                    # TODO: hacky workaround to avoid opening > max_open_trades
+                    # This emulates previous behaviour - not sure if this is correct
+                    # Prevents buying if the trade-slot was freed in this candle
+                    open_trade_count_start += 1
+                    open_trade_count += 1
```
xmatthias, line 352: This workaround is currently in place to emulate current behaviour.

Without this, more than max_open_trades are possible, assuming one trade closes at 10:00 (freeing the last slot), and another opens at 10:00.

This will be looked at at a future time, as this would probably improve alignment with dry-runs - which do show this behaviour.

PR #3886, opened 2020-10-24
Strategic stake amount

orehunt:

## Summary
allows the strategy to handle stake amount for each trade

## Quick changelog

- the bot get the stake amount from the strategy `get_stake_amount` method
- during backtesting the date is provided to allow the strategy to return the correct stake amount for the candle

Some places where there is some logging about the stake amount used were removed, but not all of them, like the ones in the rpc code..some options
- just remove it everywhere
- change it to something like `minimum_stake_amount` and maybe touple it with the stake currency

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -433,7 +433,8 @@ class FreqtradeBot:
                 Trade.total_open_trades_stakes()
             )
         else:
-            stake_amount = self.strategy.get_stake_amount(pair)
+            stake_amount = strategy_safe_wrapper(self.strategy.get_stake_amount,
+                                                 default_retval=0)(pair)
             if stake_amount == constants.UNLIMITED_STAKE_AMOUNT:
                 stake_amount = self._calculate_unlimited_stake_amount()
 
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -433,7 +433,7 @@ class FreqtradeBot:
                 Trade.total_open_trades_stakes()
             )
         else:
-            stake_amount = self.config['stake_amount']
+            stake_amount = self.strategy.get_stake_amount(pair)
             if stake_amount == constants.UNLIMITED_STAKE_AMOUNT:
                 stake_amount = self._calculate_unlimited_stake_amount()
 
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -281,7 +281,7 @@ class Backtesting:
                     trades.append(trade_entry)
         return trades
 
-    def backtest(self, processed: Dict, stake_amount: float,
+    def backtest(self, processed: Dict,
                  start_date: datetime, end_date: datetime,
                  max_open_trades: int = 0, position_stacking: bool = False) -> DataFrame:
         """
@@ -292,7 +292,6 @@ class Backtesting:
         Avoid extensive logging in this method and functions it calls.
 
         :param processed: a processed dictionary with format {pair, data}
-        :param stake_amount: amount to use for each trade
         :param start_date: backtesting timerange start datetime
         :param end_date: backtesting timerange end datetime
         :param max_open_trades: maximum number of concurrent trades, <= 0 means unlimited
@@ -308,6 +307,7 @@ class Backtesting:
         # Use dict of lists with data for performance
         # (looping lists is a lot faster than pandas DataFrames)
         data: Dict = self._get_ohlcv_as_lists(processed)
+        get_stake_amount = self.strategy.get_stake_amount
 
         # Indexes per pair, so some pairs are allowed to have a missing start.
         indexes: Dict = {}
@@ -344,6 +344,7 @@ class Backtesting:
                         and tmp != end_date
                         and row[BUY_IDX] == 1 and row[SELL_IDX] != 1):
                     # Enter trade
+                    stake_amount = get_stake_amount(pair, row[DATE_IDX])
                     trade = Trade(
                         pair=pair,
                         open_rate=row[OPEN_IDX],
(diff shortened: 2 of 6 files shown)


freqtrade/freqtradebot.py
```diff
@@ -433,7 +433,7 @@ def get_trade_stake_amount(self, pair: str) -> float:
                 Trade.total_open_trades_stakes()
             )
         else:
-            stake_amount = self.config['stake_amount']
+            stake_amount = self.strategy.get_stake_amount(pair)
```
xmatthias, line 436: Please use `strategy_safe_wrapper()` for this - i assume it'll be user provided code - so should assume that it can cause errors.

freqtrade/strategy/interface.py
```diff
@@ -429,13 +436,13 @@ def get_signal(self, pair: str, timeframe: str, dataframe: DataFrame) -> Tuple[b
         :param pair: pair in format ANT/BTC
         :param timeframe: timeframe to use
         :param dataframe: Analyzed dataframe to get signal from.
-        :return: (Buy, Sell) A bool-tuple indicating buy/sell signal
+        :return: (Buy, Sell, Date) A tuple indicating buy/sell signal for the latest date
```
xmatthias, line 439: i assume there is an idea behind ... ?
but to be fair - the return value did not change (yet) ?

xmatthias:
i guess in RPC we'll need to decide case by case how to proceed with this - where it's informative - we could just remove it i guess ... where it's used for some calculation, we should not do that anyway but use the trade's value, as stake_amount can also now be changed after a few trades.

Overall, i do like this :+1:

freqtrade/strategy/interface.py
orehunt: i was adding that to pass the date (to get_stake_amount) also during live but then removed, because since get_signal is only called by the bot during live, it could already have the date as an argument, with a now() date and get_signal would then only locate the correct index (and maybe handle edge cases for duplicates)

orehunt:
with dynamic stake amounts, what absolute profits should take precedence over relative ones, because they can be disjointed. So for example for loss functions, that use relative profits, should instead switch to absolute, or `absolute_profits / capital_at_stake`

freqtrade/optimize/backtesting.py
```diff
@@ -344,6 +344,8 @@ def backtest(self, processed: Dict, stake_amount: float,
                         and tmp != end_date
                         and row[BUY_IDX] == 1 and row[SELL_IDX] != 1):
                     # Enter trade
+                    prev_date = data[pair][indexes[pair] - 1][DATE_IDX]
+                    stake_amount = get_stake_amount(pair, prev_date)
```
xmatthias, line 348: What's the reasoning behind this? (i don't think it's correct, that's why i'm asking) 

Candle dates are from the candle-start - and we use the candle's open-price as entry-price - so assuming a 1h candle, you'd now use the date of 10:00 for a buy-signal that would use the open-signal from 11:00 ...

orehunt: signals are shifted

xmatthias: signals are shifted - yes - but shifting price back is wrong as that will "nullify" the shift partially.

Consider a current 1h candle - it's now 15:22 (UTC) - so the hourly candle will have a timestamp of 14:00.
The signal will however only be available since 15:00:01 - as this is when the candle "closed" and the signal was generated.

so a "real" trade will be bought at 15:00:01 - based on data from the candle 14:00-15:00 - ideally with a price close to the candle's open price (the candle from 15:00 - 16:00) - (hence we're using "open" as price for backtesting) 

This is the reason the signal must be shifted for backtesting - as the "date" always refers to the candle start - not the candle end - so a signal based on data for the candle at 14:00 will never be acted before the candle is complete.

---

Now thinking about this in the backtesting context: 

The signal based on 14:00-15:00 is acted on at 15:00:01 - so by shifting the date back, we call `get_stake_amount()` with 14:00 - while the open-price is taken from 15:00-16:00 - and in real mode, 15:00 is the earliest that signal did appear (the candle is incomplete up to that point, and therefore not available to the strategy to avoid repainting).

Now depending on what you do with this date, it may not be relevant - but to have this date close to how it would be called in reality, it should be called with 15:00 - the earliest time a "live" signal would be acted on

orehunt: This boils down on what's more intuitive, I think it's simpler to assume that signals and stake amount are aligned, it doesn't matter if dates are shifted, from a backtesting point of view, you have to ask for the index where that signal was populated. 
During live it would indeed be confusing if the call would ask for the previous date, but the date is not given in live mode...so 🤷 

If you name the variable like `candle_date` then I guess its more clear that it is supposed to be the date of the candle, and NOT the date when the callback _would_ be executed

xmatthias: let's leave it for now - and if we're seeing problems / confusion, this can be adjusted eventually...

epigramx:
I'm not very familiar with the codebase  but am I right that this does not provide access to the dataframe? It might have important usage since certain algorithms might want to take into account indicator data to set up a stake amount. 
PS. To be perfectly honest I find it hard to understand why certain strategy functions do not allow access to the dataframe because many algorithms need it which is especially important for stoploss/roi/trailing (I opened an issue).

epigramx:
This is possibly the most important feature missing in the software at the moment (because it deals with varying money positions which is a fundamental part of traditional trading algorithms).
But am I reading it right that `get_stake_amount()` is relatively basic and it does not provide the dataframe (or even the current_time) so that no much TA can be done for it?
For this feature to be fully useful: the user must be able to do TA with the dataframe to determine stake (in order to be able to do backtesting at least (I guess dp still exists for live)).

xmatthias:
I'll close this as too much has changed since - and rebasing / merging this would most likely be more effort than reimplementing it - and i see no progress being made in it anyway.

It may however still serve as an example on how to approach this in the future.

Issue #1519, opened 2019-01-28
position mgmt (Position stacking / pyramiding)


yaotian:
buy or sell in several steps according to the signal.

For example as bb band strategy,
1. buy signal: cross up low band, buy 50% position
2. buy signal again: cross up middle band, buy another 50% position.
3. sell signal: cross down high band, sell 50%
4, sell signal again: cross down middle band, sell 50% position.

FT supports these use cases?

yaotian:
@xmatthias this feature is very useful. 

Based on my experience, to make money success MUST have two reason: 1. trend (which can use indicators to make this) 2. position ( high risk price use low position, but latter it changes to low risk, we want to buy more)

4oplay:
Hello, lets implement this asap! this should be a freq trade bot :) @xmatthias this feature is very useful.

hroff-1902:
I'm working on it...

pantheoncap:
Please leave me a mail at eqaq131@gmail.com,
would like to share the approach

hroff-1902:
I'll report here.

ovidiu162000:
Also interested in this feature.
Any progress on this @hroff-1902 ?
Let me know if I can help in any way. Would be more than happy to assist on this.

Tamei:
Is this still a thing? Would love to see this feature

hroff-1902:
yes, it's still an open issue which waits its time, we are moving step-by-step to the goal...

pnmartinez:
> yes, it's still an open issue which waits its time, we are moving step-by-step to the goal...

I know this may be too much to ask, but it would be awesome if you could share the roadmap, the current status, or a summary of the code yet to write.

xmatthias:
I'm sorry, but we are an open source project. all of us spend our free time voluntarily working on this project - so we do not provide timing plans - as timing solely depend on the fact that some has time and interrest to work on this issue.

The status is - it is not implemented at the moment. It might be a small change - could be a big change - would need a deeper investigation first (and depends on the 2nd point below).

There are however a few unclear points to me with this issue... and i'm not sure all have a clear answer.

---

First of all - this will be a very dangerous setting - as many strategies (in the sample repository - but also what people develop) have a "sticking signal" problem.
That means, these strategies don't buy on rsi "crossed above" 30  - but they buy on "rsi > 30" (for example).

Currently - this will mean that the strategy buys the coin - and holds it until it'll sell it, and only after that look to buy again.
[...]

PR #3908, opened 2020-10-30
[wip] Partial buys

es-fem:

## Summary

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -12,17 +12,17 @@ from typing import Any, Dict, List, Optional
 import arrow
 from cachetools import TTLCache
 
-from freqtrade import __version__, constants, persistence
+from freqtrade import __version__, constants
 from freqtrade.configuration import validate_config_consistency
 from freqtrade.data.converter import order_book_to_dataframe
 from freqtrade.data.dataprovider import DataProvider
 from freqtrade.edge import Edge
 from freqtrade.exceptions import (DependencyException, ExchangeError, InsufficientFundsError,
                                   InvalidOrderException, PricingError)
-from freqtrade.exchange import timeframe_to_minutes, timeframe_to_next_date
+from freqtrade.exchange import timeframe_to_minutes
 from freqtrade.misc import safe_value_fallback, safe_value_fallback2
 from freqtrade.pairlist.pairlistmanager import PairListManager
-from freqtrade.persistence import Order,Trade, cleanup_db, init_db
+from freqtrade.persistence import Order, PairLocks, Trade, cleanup_db, init_db
 from freqtrade.resolvers import ExchangeResolver, StrategyResolver
 from freqtrade.rpc import RPCManager, RPCMessageType
 from freqtrade.state import State
@@ -70,6 +70,8 @@ class FreqtradeBot:
         init_db(self.config.get('db_url', None), clean_open_orders=self.config['dry_run'])
 
         self.wallets = Wallets(self.config, self.exchange)
+
+        PairLocks.timeframe = self.config['timeframe']
 
         self.pairlists = PairListManager(self.exchange, self.config)
 
@@ -344,6 +346,7 @@ class FreqtradeBot:
         whitelist = copy.deepcopy(self.active_pair_whitelist)
         if not whitelist:
             logger.info("Active pair whitelist is empty.")
+            return trades_created
         else:
             '''# Remove pairs for currently opened trades from the whitelist
             for trade in Trade.get_open_trades():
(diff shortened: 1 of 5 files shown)


es-fem:
To try it add this functions to the strategy 

from freqtrade.strategy.interface import PartialTradeTuple as partial

def increase_position ( dataframe: pd.DataFrame, amount: float) -> pd.Series:
    return pd.Series([partial(1,amount) for x in range(len(dataframe))])

def reduce_position ( dataframe: pd.DataFrame, amount: float) -> pd.Series:
    return pd.Series([partial(1,amount) for x in range(len(dataframe))])

And call them like: (you will still need 'buy' and 'sell' set to 0)
dataframe.loc[
            (
                    (dataframe['k'] > dataframe['d']) &
                    (dataframe['k'] < dataframe['oversold']) &
                    (dataframe['d'] < dataframe['oversold'])

            ),
            'partial_buy'] = increase_position(dataframe, 15)

xmatthias:
After having a very quick glance over this - i don't think it'll work well - at least not the way it is now.

----

To be honest, i really don't like the interface you're proposing, and i don't think it'll work well. 
Using an (optional) tuple instead of a boolean field is quite risky and errorprone, especially when non-programmers use this.
This interface method will most likely also not work correctly for backtesting.


The buy / sell methods seem to do lot's of code duplication (seem like copies of their existing counterparts, with slight modifications), and are way too long - which means they do way too much.

---

It'll also be highly problematic in longer timeframes (say 1h) ... 
the assumption is that a signal is be active for the whole candle (this is usually the case, in some strategies a signal can be active for multiple candles - if they're not using "crossed").

Unless i missed it, you're not protecting against this - so you'd buy another slice on every iteration until you run out of money, or the signal disappears (which is when the candle is over).

That's obviously very risky, and in my opinion, not useful at all.

es-fem:
Well, its a first aproach.. I'll do this if you're agree:

- Change the tuple for two booleans
- unify buy/sell methods to save code
- lock the candle to avoid rebuys/resells
- Modify backtesting and hyperopt when the rest is done.

Im not good with db and for dry_run I had to tweak wallets to calculate
stake_amount. May be we should add a table in the db for the wallets?

xmatthias:
I'm not sure i understand why you would need wallets in the database? 

all data necessary is in the trades (it'll also be that way in the future - otherwise you're tracking the wrong thing) - so a wallet table will mean one thing - the data will be there twice, which will eventually lead to inconsistencies.

---

As said, i'm not sure about the way you're doing the interface - it seems overly complicated and unnecessary.
There's a very simple way to allow position stacking:
add a new config entry which enables / disables position stacking.

* If it's disabled - nothing changes (this is how freqtrade works now).
* If it's enabled, you can buy "stake_amount" multiple times, which will then be combined into one trade (I'm not sure about combining positions - see https://github.com/freqtrade/freqtrade/issues/1519#issuecomment-659154906 for details)

That way, the strategy interface doesn't change - the whole code will be a lot cleaner.

That way you can't dynamically modify the amount you buy - but that's not necessary to enable position stacking.
Also, i would not do that in the dataframe, but as part of a callback function (which must be a seperate PR, otherwise the change would be too large to review).


Before doing any more work on the interface, you should maybe also read #2183. Either using a tuple, or anything else will limit the possibilities of introducing shorting or leveraged pairs - so it's definitely not a good idea.

es-fem:
Hi, The wallets in the database idea is because if you modify the open
trade lets say selling some, the actual calcs don't actualize the wallet if
the trade is open. I had to modify the wallets.py to consider stake_amount
as part of the USD wallet for open trades, breaking the logic of
"stake_amount = amount * open_rate" in the db. (this  logic is also a
duplication thus open_rate and amount are already in the db so
stake_amount  could be calculated anytime).
I like the idea of #2183, I don't know if it will work with stake =
unlimited, but it's better than my idea.
The only thing that maybe is usable from my code is the method that
averages the open_price. That's a standard formula needed to calculate
unrealized p&l for a position plus leverage and shorts options are easy to
add  in p&l.  something like this:
https://help.bybit.com/hc/en-us/articles/900000404726-P-L-calculations-Inverse-
What I was trying to do is to emulate exchanges. They consider an open
position , modify contracts in the position and recalculate  average open
price when you add contracts to it without closing the trade. But combining
trades (as they are right now) to build a new actualized trade entry will
work as well  ( It looks more complicated to me but I understand it is
better for backward compatibility).

freqtrade/freqtradebot.py
```diff
@@ -348,26 +347,28 @@ def enter_positions(self) -> int:
...
-            logger.info("No currency pair in active pair whitelist, "
-                        "but checking to sell open trades.")
-            return trades_created
-        # Create entity and execute trade for each pair from whitelist
-        for pair in whitelist:
-            try:
-                trades_created += self.create_trade(pair)
-            except DependencyException as exception:
-                logger.warning('Unable to create trade for %s: %s', pair, exception)
+        else:
```
mrsegen, line 350: `else` unnecessary after an `if` that contains a return. ("Flat is better than nested" - [Zen](https://www.python.org/dev/peps/pep-0020/))

Issue #3909, opened 2020-10-31
get_pair_dataframe during backtesting expects the data to be stored in json-format


matspi:
## Describe your environment

  * Operating system: Debian-Linux
  * Python Version: 3.8.6
  * CCXT version: 1.36.85
  * Freqtrade Version: 2020.09

## Describe the problem:

When using `DataProvider.get_pair_dataframe` during backtesting, freqtrade expects the historical data to be stored on disk as JSON, the config flag `dataformat_ohlcv` is ignored

### Steps to reproduce:

  1. Set dataformat_ohlcv to hdf5
  1. Download historical data
  2. Create a strategy, that is using `DataProvider.get_pair_dataframe`
  3. Backtest that strategy
  
### Observed Results:

Backtesting prints `WARNING - No history data for pair...`

Depending on what you try to do with the pair data, backtesting exists with a fatal like `Exception: inputs are all NaN` or similar

### Relevant code exceptions or logs

  ```
2020-10-31 11:57:32,271 - freqtrade.optimize.backtesting - INFO - Loading data from 2020-10-15 21:30:00 up to 2020-10-31 10:10:00 (15 days)..
2020-10-31 11:57:32,271 - freqtrade.optimize.backtesting - INFO - Running backtesting for Strategy strategy
2020-10-31 11:57:32,285 - freqtrade.data.history.idatahandler - WARNING - No history data for pair: "TRX/USDT", timeframe: 4h. Use `freqtrade download-data` to download the data
2020-10-31 11:57:32,286 - freqtrade.data.dataprovider - WARNING - No data found for (TRX/USDT, 4h).
2020-10-31 11:57:32,286 - freqtrade - ERROR - Fatal exception!
  ```

xmatthias:
Closed (by yourself :+1: ) in #3910 

Great job, and thanks again!

PR #3910, opened 2020-10-31
Dataformat informative pair

matspi:
## Summary
Use the value of `dataformat_ohlcv` when using `DataProvider.get_pair_dataframe`

Solve the issue: #3909

## Quick changelog

- Passing value of `dataformat_ohlcv` to `load_pair_history`
- Add unittest

## What's new?

Solves #3909


xmatthias (approved):
Great catch! 
Thanks a lot for detecting (and fixing at the same time)!

LGTM :+1:

PR #3929, opened 2020-11-06
change backtesting behaviour if roi and trailing-stop happen at the same time

### LINKED ISSUE #2422: backtesting and sell signal / stoploss priority
When there is a sell signal, and a stoploss trigger on the same candle, stoploss takes over sell signal. Is this intended?
- It makes sense for timeframes <15m
- not so much for timeframes >30m
There are edge cases where if you use order book, sell is obviously not guaranteed, so it is conservative to prioritize stoploss, but generally market selling with >30m timeframes should always favour sell signal over stoploss.

radwayne:

## Summary
Changed the `should_sell()` method, to handle the case where both ROI and trailing stoploss are reached in backtest. In that case, the sell reason should be ROI, and not stoploss.

## What's new?
When backtesting some strategies in long timeframes (1h, 2h, 4h..), I noticed that trailing stop loss made profits superior to the strategy's ROI.
I'm not sure if this behavior is intended, but it distorts backtest and hyperopt results significantly.

When reading the documentation, I understood that ROI is always tested first, in order to protect capital. But in this case, capital isn't protected, and profits are overestimated.

My suggestion is to add the roi_reached test in the case of trailing_stop_loss selltype. That way, stoploss is always checked first, but trailing stoploss can't make a profit larger than ROI.

This doesn't alter the strategy's behavior in live or dryrun modes, since ROI and trailing stoploss can't hit at the same time.

*edit*: fixes #2422

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -278,6 +278,7 @@ Since backtesting lacks some detailed information about what happens within a ca
 - Trailing stoploss
   - High happens first - adjusting stoploss
   - Low uses the adjusted stoploss (so sells with large high-low difference are backtested correctly)
+  - ROI applies before trailing-stop, ensuring profits are "top-capped" at ROI if both ROI and trailing stop applies
 - Sell-reason does not explain if a trade was positive or negative, just what triggered the sell (this can look odd if negative ROI values are used)
 - Stoploss (and trailing stoploss) is evaluated before ROI within one candle. So you can often see more trades with the `stoploss` and/or `trailing_stop` sell reason comparing to the results obtained with the same strategy in the Dry Run/Live Trade modes.
 
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -268,19 +268,24 @@ It contains some useful key metrics about performance of your strategy on backte
 Since backtesting lacks some detailed information about what happens within a candle, it needs to take a few assumptions:
 
 - Buys happen at open-price
-- Sell signal sells happen at open-price of the following candle
-- Low happens before high for stoploss, protecting capital first
+- Sell-signal sells happen at open-price of the consecutive candle
+- Sell-signal is favored over Stoploss, because sell-signals are assumed to trigger on candle's open
 - ROI
   - sells are compared to high - but the ROI value is used (e.g. ROI = 2%, high=5% - so the sell will be at 2%)
   - sells are never "below the candle", so a ROI of 2% may result in a sell at 2.4% if low was at 2.4% profit
   - Forcesells caused by `<N>=-1` ROI entries use low as sell value, unless N falls on the candle open (e.g. `120: -1` for 1h candles)
 - Stoploss sells happen exactly at stoploss price, even if low was lower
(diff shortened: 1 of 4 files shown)


radwayne:
> Actually, stoploss happens first - and we assume that "low" happens before "high" because of this.

I meant to say exactly that, my bad...

> I'm not certain about this point (but i don't think it's very severe) - in theory you can have a config as follows:
`{0: 0.1, 60: 0.01}`` - so if exactly after 1 hour, profit is 2%, it would sell immediately with ROI (currently).
this same situation can however also be a trailing stop (profit was 4% before the hour, and dropped by 1% right on the new candle, so trailing stop would "also" fire.

You are right, this could happen, and roi would have the priority in this case, but as you mentioned, it's not necessarily a bad thing. 

This is probably not the best fix for the problem (if we can call it one), as it doesn't take into consideration other factors such as sell signal which would require similar handling, but I wanted to bring it to your attention for future enhancements.

> In any case, i'd apreciate if you'd be able to write a test (additional entry) in line with what we got in test_backtest_detail.py - which should ideally fail in develop, and succeed in this branch - to actually illustrate the problem - and more importantly - to make sure we're not reverting this unintentionally in the future again

I will do that and get back to you.

xmatthias:
> I will do that and get back to you. 

perfect, thanks a lot!

due to the other issue with sell-signal (#2422), it was (kindof) known to me already - so once we have a test for this, i'd want to try and add the test i've written for the "sell signal" case above and try a combined fix.

xmatthias:
I did a bit of work on this today, improving backtesting evaluation logic / sequence.

This adds tests for the case described above (ROI and trailing stop colliding).
This will also add tests for #2422 - as sell-signal should win over stoploss (as sell-signal happens at open, but low will always at/happen after the candle's open).

The above statement however also raises an additional question:
* should sell-signal win over ROI (so if sell-signal and ROI collide, who should win)? 
This would be (tc27) - currently ROI wins, but this has the potential to inflate profits via ROI


I'd apreciate if bot @radwayne and @orehunt could have a look at this as this, as both of you did some investigation into this topic.
This issue will now also fix #2422 - should there be determined that this logic is correct.

radwayne:
I'm sorry I couldn't manage to write the test for the roi / trailing stoploss collision.
I think the logic is sound, and the tests demonstrate it well. Plus the code is elegant.
Thank you @xmatthias.

xmatthias:
@radwayne don't worry, pointing out and giving the fix a try is great - I've had some time (or call it unwillingness to work on what i should work on :laughing: ) so i thought i'd give this a go - and i wanted to combine it with the fix for the stoploss/sell problem anyway.

xmatthias (approved):
LGTM :+1: 

As this is a PR that will change Backtest-logic quite a bit:
if noone objects by then, i'll merge this sometime tomorrow - i hope ~5 days is enough time.

Dear unknown reader/user:
if you're interrested in having a closer look at it and require a few additional days, please let me know so i can postpone merging by a few days until you had time to look at it (

PR #3939, opened 2020-11-10
Historic pair

drdux:
## Summary
enables you to backtest pairs that are not live on exchanges any more

## What's new?
I wanted to backtest pairs that are not live on the exchange any more, but that was not possible due to several checks. So I added a PairListHandler that does not validate the pairs and added an exchange config flag that allows you to skip the pair validation there.
I think it could be useful to others if you want to backtest pairs that expire like quarterly contracts etc.

Let me know if that's something you'd like to merge then I will add the documentation for it.


freqtrade/pairlist/ConstPairList.py
```diff
@@ -0,0 +1,60 @@
...
+        return f"{self.name}"
+
+    def gen_pairlist(self, cached_pairlist: List[str], tickers: Dict) -> List[str]:
+        """
+        Generate the pairlist
+        :param cached_pairlist: Previously generated pairlist (cached)
+        :param tickers: Tickers (from exchange.get_tickers()).
+        :return: List of pairs
+        """
+        return self._config['exchange']['pair_whitelist']
```
xmatthias, line 50: if i'm not mistaking, this is almost an exact copy of StaticPairList with just this line changed.

In that case, I'd rather have an "allowInactive" (or "validatePairs" or similar) flag on the "StaticPairList" configuration (look at `VolumePairlist` or `AgeFilter` on how that might be done) that does (or does not) do the validation. 
A lot less duplicate code, while archiving the same thing.

freqtrade/exchange/exchange.py
```diff
@@ -115,7 +115,7 @@ def __init__(self, config: Dict[str, Any], validate: bool = True) -> None:
 
         logger.info('Using Exchange "%s"', self.name)
 
-        if validate:
+        if validate and not exchange_config.get('skip_validation'):
```
xmatthias, line 118: Personally, i would be more specific with this. Skipping all validations can be dangerous...

This will skip "all" validations - but at least for your usecase, only "validate_pairs" should be skipped - as stake-currency should still match (otherwise your backtests may show very odd results).

xmatthias (commented):
Personally, i don't see what you'd gain from backtesting a delisted coin - as a new coin might show completely different behaviours.
I'm also not sure this suffices to allow downloading a delisted pair (do exchanges still provide historic data for that delisted pair??) - assuming someone "new" to this pair would like to backtest it anyway.

However, i also see that others might have use for it - so i guess it can be useful.
I would however not create a seperate pairlist for this - but instead "enhance" the existing StaticPairList with an optional setting to skip validation.

freqtrade/pairlist/ConstPairList.py
drdux: That's a nice mechanism, I remove the new handler and added allow_inactive to StaticPairList.

freqtrade/exchange/exchange.py
drdux: good point, done

drdux:
To your initial doubt @xmatthias whether this makes any sense at all:
I'm using this for quarterly and daily contracts on FTX. They have all the same underlying and hence correlated behaviour. The exchange API provides all the data for these expired symbols. It's an edge case though, I'll give you that. But I'm pretty sure it will be useful for others in the future.

xmatthias:
> ... It's an edge case though, I'll give you that. But I'm pretty sure it will be useful for others in the future.

It can be an edge-case now - but it's a simple change and i can see the point why you're requiring this.
I suspect this can become even more interresting in the future - assuming freqtrade will support "native" futures at some point in the (not so near) future.

---

The only thing for this PR is that we'd need this option in the docs (`pairlist.md`) - otherwise it'll become a "hidden gem" no-one except for yourself (and with luck myself) will remember 2 months down the road.

drdux:
Oh nice job @xmatthias . Was going to do it today :D sorry

xmatthias:
ups sorry :D 
i thought i'd just add this part - as a quick addition to the docs shouldn't hold up a PR for too long :laughing:

xmatthias (approved):

drdux:
I'll be quicker next time :wink:

xmatthias:
Don't worry :+1:

PR #3952, opened 2020-11-15
Fix typo in windows installation docs

Samaoo:
- Fixed a typo
- Replaced the hyphens U+2011 with U+002D because otherwise pip will throw an error saying ```ERROR: TA_Lib‑0.4.19‑cp38‑cp38‑win_amd64.whl is not a valid wheel filename.```


xmatthias (approved):
Great catch! :+1: 

Thanks a lot for taking the time and fixing it :100:

PR #3975, opened 2020-11-22
RangeStabilityFilter filter - filter for pairs without much movement

xmatthias:
## Summary
This PR will introduce the RangeStabilityFilter
It's calculated over the last 10 days (by default)- using the % change between highest high and lowest low in this time interval.
If this Range % is below the specified threshold (1% by default) the pair is rejected from the pairlist.

This can be useful to remove stable-coins from the pairlist without explicitly blacklisting them, as stable-coins usually move within a very tight range, which makes trading them profitable very difficult.

## Quick changelog

- Introduce RangeStabilityFilter Filter
- Added tests for this new filter

### DIFF
--- a/freqtrade/pairlist/volatilityfilter.py
+++ b/freqtrade/pairlist/volatilityfilter.py
@@ -31,8 +31,8 @@ class VolatilityFilter(IPairList):
         if self._days < 1:
             raise OperationalException("VolatilityFilter requires volatility_over_days to be >= 1")
         if self._days > exchange.ohlcv_candle_limit:
-            raise OperationalException("VolatilityFilter requires volatility_over_days to not exceed "
-                                       "exchange max request size "
+            raise OperationalException("VolatilityFilter requires volatility_over_days to not "
+                                       "exceed exchange max request size "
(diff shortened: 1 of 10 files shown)


freqtrade/pairlist/volatilityfilter.py
```diff
@@ -0,0 +1,89 @@
...
+                       .float_timestamp) * 1000
+
+        daily_candles = self._exchange.get_historic_ohlcv_as_df(pair=pair,
+                                                                timeframe='1d',
+                                                                since_ms=since_ms)
+        result = False
+        if daily_candles is not None:
+            highest_high = daily_candles['high'].max()
+            lowest_low = daily_candles['low'].min()
+            pct_change = (highest_high - lowest_low) / lowest_low
```
matspi, line 77: Do your need to protect against division by zero?
Trading price of zero not really makes sense, but in the world of cryptos such things might be possible.

xmatthias: Nice catch - i'm checking for that now :+1:

PR #4001, opened 2020-11-27
[Documentation] Fix parameter name

mrsegen:
Correct which parameter name was referred to within the 2nd Note under "Amend last stake amount"

## Summary
Make sure someone seeking to adjust the last stake amount min ratio knows to use `last_stake_amount_min_ratio` instead of `last_stake_amount_min_ratio`.

## Quick changelog

- Fix parameter in documentation.


xmatthias (approved):
Great catch! :+1:

PR #4002, opened 2020-11-27
[Documentation] Prevent unintended LaTeX rendering

mrsegen:
## Summary
Attempt to fix unintended LaTeX rendering between to `$` symbols by applying [MathJaX's ignore class](https://docs.mathjax.org/en/latest/options/document.html?highlight=mathjax_ignore#option-descriptions).

## What's new?
Before this fix, as displayed below, there is awkward rendering. After this fix, the portion of text should now display: `101$, the stop loss will be 91.8$`.
image

### DIFF
--- a/docs/stoploss.md
+++ b/docs/stoploss.md
@@ -36,8 +36,8 @@ If `stoploss_on_exchange` uses limit orders, the exchange needs 2 prices, the st
 `stoploss` defines the stop-price where the limit order is placed - and limit should be slightly below this.  
 If an exchange supports both limit and market stoploss orders, then the value of `stoploss` will be used to determine the stoploss type.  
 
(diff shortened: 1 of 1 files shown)


mrsegen:
I’m closing this because I’m pretty sure the fix isn’t as simple as I thought.

xmatthias:
no it's not ...
but it's still a good spot i didn't notice before (that it's rendering wrongly)

xmatthias:
I will reopen this as the fix is actually easier than this (and i've already started doing this on your branch)...

hope you're fine with that :)

xmatthias (approved):
Once again, great catch! :+1:

mrsegen:
Thanks for picking it up!

PR #4009, opened 2020-11-28
[Pairlist] Add PerformanceFilter

### LINKED ISSUE #3441: Sorting filter by performance
## Describe your environment

  * Operating system: Ubuntu
  * Python Version: 3.8
  * CCXT version: 1.28.49
  * Freqtrade Version: 2020.4

## Describe the enhancement
I would like to suggest a new filter that can be used instead of the ShuffleFilter, that is, it takes the current list of pairs from the filter above it and sort pairs according to their overall performance over time. At first, there’s no performance because there are no trades, but over time, pairs get performance. If they are positive, the filter pushes them to the top of the list, if they are negative, the filter pushes them to the bottom of the list. Pairs which do not jet have trades are in the middle of the list. Sorting can be done with performance value and a count of trades (as in 'get_overall_trades). Also, there is a possibility to introduce a threshold for values below a certain level.

mrsegen:
## Summary
Add a PerformanceFilter to the available options for the bot to use when selecting which currency pair to trade.

Closes #3441 

## Quick changelog

- Add the PerformanceFilter proposed by @apneamona [here](https://github.com/freqtrade/freqtrade/issues/3441#issuecomment-641815036)
- Add PerformanceFilter to AVAILABLE_PAIRLISTS within constants.py
- Mirror ShuffleFilter's appearance within parameterized tests in tests/pairlist/test_pairlist.py.
- Add additional test cases for how test PerformanceFilter runs
- Add a PerformanceFilter entry within docs/includes/pairlists.md

## What's new?


freqtrade/pairlist/PerformanceFilter.py
```diff
@@ -0,0 +1,56 @@
...
+        :return: new whitelist
+        """
+        # Get the trading performance for pairs from database
+        perf = pd.DataFrame(Trade.get_overall_performance())
+        # get pairlist from performance dataframe values
+        list_df = pd.DataFrame({'pair': pairlist})
+        # set initial value for pairs with no trades to 0
+        # and sort the list using performance and count
+        sorted_df = list_df.join(perf.set_index('pair'), on='pair')\
+            .fillna(0).sort_values(by=['profit', 'count'], ascending=False)
```
xmatthias, line 53: i would rather use `list_df.merge(perf, on='pair', how='left')` for the join ... that will make it obvious that it's a left-join.

Otherwise the merge operation is kinda strange, as `list_df` doesn't have pair as index, while perf has, which makes this code reliant on pandas not changing this behaviour for join ... 

And honestly, i was surprised that it works this way ...

xmatthias (commented):
Overall, good job :+1: 

One comment on the logic to determine profitable trades 

Assume we have a tie (which i find rather unlikely anyway - as in real trading, you'll never have 5.000% - but 5.001% or so (price rounding, fees, ...).

Now assuming we DO happen to have a tie - wouldn't we prefer a pair that made the same profit in fewer trades?

* AAA has 10 trades and 5% profit
* BBB has 25 trades and 5% profit

Personally, i'd favor AAA, as it took only 5% to reach a sum(!) of 5% profit, while BBB took 25 trades for the same profit - which suggests it had more losses.

As said above, i'm not 100% clear if a tie is realistic, so i don't mind having it the way it is now - it's just something we should think about and take a concious decision on...

tests/pairlist/test_pairlist.py
```diff
@@ -701,11 +718,61 @@ def test_pricefilter_desc(mocker, whitelist_conf, markets, pairlistconfig,
...
+     ['TKN/BTC', 'LTC/BTC', 'ETH/BTC']),
+    # Tie in performance data broken by count
+    ([{"method": "StaticPairList"}, {"method": "PerformanceFilter"}],
+     ['ETH/BTC', 'TKN/BTC', 'LTC/BTC'],
+     [{'pair': 'LTC/BTC', 'profit': -5, 'count': 101},
+      {'pair': 'TKN/BTC', 'profit': -5, 'count': 2},
+      {'pair': 'ETH/BTC', 'profit': -5, 'count': 100}],
+     ['LTC/BTC', 'ETH/BTC', 'TKN/BTC']),
+])
+def test_performance_filter(mocker, whitelist_conf, pairlists, pair_allowlist, overall_performance,
```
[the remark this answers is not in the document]
mrsegen: Agreed that it should be more explicitly tested. 

[This comment](https://github.com/freqtrade/freqtrade/issues/3441#issuecomment-735045902) alludes to some of the current strange sorting that occurs when no performance data is available - seemingly sorted by length of string.

mrsegen:
I agree with your comments. I'll plan to update sometime today.

tests/pairlist/test_pairlist.py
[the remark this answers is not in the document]
mrsegen: As it was, I was seeing a `KeyError: 'pair'` when providing an empty list. So I modified PerformanceFilter to skip performance-based processing if that case.

freqtrade/constants.py
```diff
@@ -25,7 +25,8 @@
                          'SortinoHyperOptLoss', 'SortinoHyperOptLossDaily']
 AVAILABLE_PAIRLISTS = ['StaticPairList', 'VolumePairList',
                        'AgeFilter', 'PrecisionFilter', 'PriceFilter',
-                       'RangeStabilityFilter', 'ShuffleFilter', 'SpreadFilter']
+                       'RangeStabilityFilter', 'ShuffleFilter', 'SpreadFilter',
+                       'PerformanceFilter']
```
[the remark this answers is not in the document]
mrsegen: ✔

freqtrade/pairlist/PerformanceFilter.py
mrsegen: Good point, re: pandas dependency. Your suggestion is implemented,

tests/pairlist/test_pairlist.py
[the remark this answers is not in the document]
xmatthias: Maybe doing the sorting based on `sort_values(by=['profit', 'count', 'pair'])`  would help to not sort by "length of the string" - but alphabetically (better than length of string i guess).

Either way, skipping in case of no performance data available is perfectly fine - i think that was discussed in the issue too - as this filter will have a undefined startup-period anyway.

tests/pairlist/test_pairlist.py
```diff
@@ -302,6 +302,10 @@ def test_VolumePairList_refresh_empty(mocker, markets_empty, whitelist_conf):
     ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "quoteVolume"},
       {"method": "ShuffleFilter"}],
      "USDT", 3),  # whitelist_result is integer -- check only length of randomized pairlist
+    # PerformanceFilter
+    ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "quoteVolume"},
+      {"method": "PerformanceFilter"}],
+     "USDT", ['ETH/USDT', 'NANO/USDT', 'ADAHALF/USDT', 'ADADOUBLE/USDT']),
```
mrsegen, line 308: It's still strange to me that changing this order fails tests.

mrsegen:
PR is updated and ready for another review.

tests/pairlist/test_pairlist.py
xmatthias: i'm not sure what you mean ... 
this test is there to verify the correct sequence of the output - so it should fail if it get's a wrong order back ...

mrsegen: I had to place it in that specific order for the test to pass. If the order is different, it will get sorted into that order.

Perhaps that is expected for other reasons I'm missing.

mrsegen: Oops... I see your [comment](https://github.com/freqtrade/freqtrade/pull/4009#discussion_r532053809) now and will investigate.

mrsegen:
Oh, I still intend to do ascending sort on trade count. 

Gah... Now, I'm imagining a combined metric of number of trades and performance. But the complexity and number of preferences in different circumstances will grow quickly down that rabbit hole.

tests/pairlist/test_pairlist.py
```diff
@@ -302,6 +302,10 @@ def test_VolumePairList_refresh_empty(mocker, markets_empty, whitelist_conf):
     ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "quoteVolume"},
       {"method": "ShuffleFilter"}],
      "USDT", 3),  # whitelist_result is integer -- check only length of randomized pairlist
+    # PerformanceFilter
+    ([{"method": "VolumePairList", "number_assets": 5, "sort_key": "quoteVolume"},
+      {"method": "PerformanceFilter"}],
+     "USDT", ['ETH/USDT', 'NANO/USDT', 'ADADOUBLE/USDT', 'ADAHALF/USDT']),
```
mrsegen, line 308: @xmatthias ... continued from [this thread](https://github.com/freqtrade/freqtrade/pull/4009#discussion_r532053809):

This doesn't solve the issue unfortunately - at least when there aren't yet trades.

I committed and pushed the test case with incoming list of `['ETH/USDT', 'NANO/USDT', 'ADADOUBLE/USDT', 'ADAHALF/USDT']` still getting rearranged to `['ETH/USDT', 'NANO/USDT', 'ADAHALF/USDT', 'ADADOUBLE/USDT']` - presumably due sorting happening within to `VolumePairList` or elsewhere.

tests/pairlist/test_pairlist.py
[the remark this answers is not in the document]
mrsegen: Resolving this in deference to [comment](https://github.com/freqtrade/freqtrade/pull/4009#discussion_r532086775) at location of remaining related test.

tests/pairlist/test_pairlist.py
mrsegen: I restored the order to enable tests to pass. I am convinced that any issue here lies outside the scope of this PR.

xmatthias: well ideally, the result will be identical with / without Performancefilter (only with the filters before it) - as in case of "No trades yet" - we'll simply skip this filter as it will not work anyway.

mrsegen: Test case now removed.

mrsegen:
I’m perplexed by this failure in an unchanged file:
```
=========================== short test summary info ============================
268
FAILED tests/rpc/test_rpc_apiserver.py::test_api_logs - assert 4 == 5
269
============ 1 failed, 1410 passed, 2 skipped in 146.56s (0:02:26) =============
```

mrsegen:
Not sure what that hiccup was on commit  https://github.com/freqtrade/freqtrade/pull/4009/commits/4b6f5b92b59e21896035c2f47475a3fcdf07c377, but the PR now passes checks.

xmatthias:
there's 2 tests (both in rpc related to logs) that are fluky sometimes.

I've not been able to recreate these flukes locally (not with the same random key ... )- but they do happen from time to time and can be ignored (usually rerunning CI fixes the issue).

docs/includes/pairlists.md
```diff
@@ -74,6 +75,15 @@ be caught out buying before the pair has finished dropping in price.
 
 This filter allows freqtrade to ignore pairs until they have been listed for at least `min_days_listed` days.
 
+#### PerformanceFilter
+
+Sorts pairs by performance, as follows:
```
xmatthias, line 80: maybe 

```suggestion
Sorts pairs by past trade performance, as follows:
```

to make it clear that it's from the trades performance (als simple historic returns can be seen as one form of performance ...)

Apart from this i think this is ok now (unless you think there's still something missing)

mrsegen: It's good to go

docs/includes/pairlists.md
```diff
@@ -74,6 +75,15 @@ be caught out buying before the pair has finished dropping in price.
...
 This filter allows freqtrade to ignore pairs until they have been listed for at least `min_days_listed` days.
 
+#### PerformanceFilter
+
+Sorts pairs by past trade performance, as follows:
+1. Positive performance.
+2. No closed trades yet.
+3. Negative performance.
+
+Trade count is used as a tie breaker.
```
mrsegen, line 85: I acknowledge that alphabetical sort as a secondary tie breaker is omitted. But with that being so unlikely (and presumably inconsequential), it seems fine to omit.

mrsegen (commented):
Issues/suggestions resolved _and_ checks passing. 😄

xmatthias (approved):
LGTM :+1: 

Many thanks for your involvement and great addition to freqtrade!

PR #4031, opened 2020-12-03
Put dollar sign before amount in stoploss.md

Samaoo:

## Summary
Put dollar sign before amount in stoploss.md


xmatthias:
What would be the reason for this? (any source why this would be correct and the prior format would be incorrect?)

to me, $100 and 100$ are technically the same, and it will depend on the culture you're comming from which is the most used format?

Samaoo:
The reason I made this PR is to be more consistent with the rest of the docs, not because I think it is incorrect.

xmatthias:
the only other place in the docs i now quickly found is in edge - and it's not consistent within that page either ... 😆 

Honestly, i prefer the `100$` approach, as it's more like fluent language (`one hundred dollars` - vs. `$100`).

we also usually use 1 BTC (at least i think we do), not BTC1 ... (but given, BTC does not quite have it's own symbol on the keyboard ...)

Samaoo:
Okay, I will close this PR and make one for the edge docs.

Issue #4044, opened 2020-12-08
Cannot parse --timerange when download data


baont:
## Describe your environment

  * Operating system: Mac 10.15.6 
  * Python Version: 3.8 (`python -V`)
  * CCXT version: 1.32.7 (`pip freeze | grep ccxt`)
  * Freqtrade Version: 2020.11 (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
  
Note: All issues other than enhancement requests will be closed without further comment if the above template is deleted or not filled out.

## Describe the problem:

I follow this page https://www.freqtrade.io/en/latest/data-download/ to download binance data but freqtrade complained it could not parse the timerange
### Steps to reproduce:

run freqtrade with these args: 
download-data --timerange 20200101 --exchange binance --pairs-file user_data/data/binance/pairs.json --timeframe 3m

### Relevant code exceptions or logs

Note: Please copy/paste text of the messages, no screenshots of logs please.

  ```
 Traceback (most recent call last):
  File "/Users/baonguyen/artisky/freqTrade/ari_trade/freqtrade/main.py", line 37, in main
    return_code = args['func'](args)
  File "/Users/baonguyen/artisky/freqTrade/ari_trade/freqtrade/commands/data_commands.py", line 35, in start_download_data
    timerange = timerange.parse_timerange(config['timerange'])
  File "/Users/baonguyen/artisky/freqTrade/ari_trade/freqtrade/configuration/timerange.py", line 107, in parse_timerange
    raise Exception('Incorrect syntax for timerange "%s"' % text)
Exception: Incorrect syntax for timerange "20200101"
  ```

Samaoo:
The syntax for the timerange argument should be ```--timerange 20200101-``` instead of ```--timerange 20200101```.

baont:
Thanks @Samaoo , i didn't try your fix but if that's the case then the doc needs update

Samaoo:
I will make a pull request to update the docs

xmatthias:
Thanks for updating the Docs!

closed by #4045, which updates the docs to fix this typo.

PR #4045, opened 2020-12-08
Update data-download.md

Samaoo:
Solve the issue: #4044


xmatthias (approved):
Great catch! :+1: 

And thanks for helping in the issue too!

PR #4046, opened 2020-12-09
Update dockerfile to multistage

dmmop:
## Summary
This change reduce the image size about ~35%

## What's new?
Reduce the image size using multiple stages for docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONFAULTHANDLER 1
+ENV PATH=/root/.local/bin:$PATH
 
 # Prepare environment
 RUN mkdir /freqtrade
@@ -13,7 +14,7 @@ WORKDIR /freqtrade
 # Install dependencies
 FROM base as python-deps
 RUN apt-get update \
-    && apt-get -y install git curl build-essential libssl-dev \
+    && apt-get -y install curl build-essential libssl-dev \
     && apt-get clean \
     && pip install --upgrade pip
 
@@ -33,7 +34,7 @@ COPY --from=python-deps /usr/local/lib /usr/local/lib
 ENV LD_LIBRARY_PATH /usr/local/lib
 
 COPY --from=python-deps /root/.local /root/.local
-ENV PATH=/root/.local/bin:$PATH
+
 
 
 # Install and execute
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ WORKDIR /freqtrade
 # Install dependencies
 FROM base as python-deps
 RUN apt-get update \
-    && apt-get -y install curl build-essential libssl-dev \
+    && apt-get -y install curl build-essential libssl-dev git \
     && apt-get clean \
     && pip install --upgrade pip
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,40 @@
-FROM python:3.8.6-slim-buster
+FROM python:3.8.6-slim-buster as base
 
-RUN apt-get update \
-    && apt-get -y install curl build-essential libssl-dev sqlite3 \
-    && apt-get clean \
-    && pip install --upgrade pip
+# Setup env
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,24 +1,40 @@
...
+RUN  pip install --user --no-cache-dir numpy \
+  && pip install --user --no-cache-dir -r requirements-hyperopt.txt
+
+# Copy dependencies to runtime-image
+FROM base as runtime-image
+COPY --from=python-deps /usr/local/lib /usr/local/lib
+ENV LD_LIBRARY_PATH /usr/local/lib
+
+COPY --from=python-deps /root/.local /root/.local
+ENV PATH=/root/.local/bin:$PATH
```
xmatthias, line 36: Would it create a problem to move the PATH definition to the top (to the definition of the base image)?

Currently the build-process is showing "installed in '/root/.local/bin' which is not on PATH." quite a few times (in red) ...
It's not a problem at the end, but it makes the build-process "seem" to want to fail ...

---

I'd also keep `pip install --upgrade pip` there (in the base image). 
While not absolutely necessary - it'll help with flawlessly building images (without warning) that depend on the base freqtrade images (check the `docker/` folder for examples), and who knows if pip will require an "uptodate" version in the future.

dmmop: No, there are any problem to move that line to the top. It's just for how i organize the code.

And the `pip install --upgrade pip` is not needed in the base image, but it can be interesting build an intermediate image like `docker build --target python-deps -t freqtradeorg/freqtrade:dependencies .` and use it as base for other images.
For example Docker.technical:
```Dockerfile
FROM freqtradeorg/freqtrade:dependencies as python-deps
RUN pip install --user --no-cache-dir git+https://github.com/freqtrade/technical

FROM freqtradeorg/freqtrade:develop
COPY --from=python-deps /root/.local /root/.local
```

The image `freqtrade:dependencies` only need rebuild if there are any update in TA-lib or changes in requierements.txt

PD: I can change the others Dockerfiles with this premise.

xmatthias: The current approach for "additional" images is to make the base pullable from docker (`freqtradeorg/freqtrade:develop` is on docherhub) - a dependency image would be local (and i don't see a point in pushing that as well, as it will make it difficult to pick the "right" image for each task - considering that we already push at least 2 branches to dockerhub) - so i don't see which benefit this would add.

We would end up with
`:develop`
`:develop_pi`
`:develop_plot`
`:develop_dependencies`  -- when do i pick this? makes it very difficult for the user
`:stable`
`:stable_pi`
`:stable_plot`
`:stable_dependencies`  -- when do i pick this? makes it very difficult for the user
...

dmmop: I'm agree with you that multistage docker is not something known to the general public, and it will be difficult for the users choose the correct image. And this is enought for me to dismiss this merge request...

xmatthias: huh? no why? what you did so far is great! :+1: 

I'm just ... questioning the change to all other images :) 

the regular dockerfile should not be touched by users (unless they know why) - so it shall be optimized for performance / size - not for ease of reading (although i don't find multistage imges particularily complicated 
 myself, but i do use docker quite a lot ...)

the `Dockerfile.technical` serves as example in the docs ... so THAT is the one i'm keen on keeping simple :)

dmmop: In the other images they can not perform a simple "pip install" because in some case it will need some basic stuffs like "build-essentials". 

Then if we change the regular dockerfile must edit the other files to avoid error.

In that case is prefer keep regular dockerfile without changes.

By the way, why you have an specific dockerfile for raspberry pi rather multiarch build?

xmatthias: > By the way, why you have an specific dockerfile for raspberry pi rather multiarch build?

Because the regular image uses the python3.8 base image, but using that as base for raspberry makes the build in CI time out (it runs for ~6-8 hours on my pc) as it needs to build ALL C libraries from scratch.

Using the 3.7 image, we can "cheat", add piwheels.org as index (they only have 3.7 wheels though) - which speeds up the build process to be ~15-30 minutes.

I know it's not ideal to have 2 images - but it's the only way i found to build the ARM image on CI.

xmatthias (approved):
LGTM 👍

Issue #3973, opened 2020-11-21
Average loss on stoploss order in backtesting is -0.3 lower than the actual stoploss


Samaoo:
## Describe your environment

  * Operating system: Windows 10
  * Python Version: 3.8.5
  * CCXT version: 1.36.85
  * Freqtrade Version: 2020.10
  
Note: All issues other than enhancement requests will be closed without further comment if the above template is deleted or not filled out.

## Describe the problem:

For example, if the stoploss in the strategy is -5, the average loss on stoploss orders will be -5.3 in backtesting

xmatthias:
yes, and it's kinda intentional (even though it's not really documented).

For the below - assume live mode  with a stoploss of -0.05 (-5%).

Then the -5% will be your stop price - not the price you're filled. 
Let's assume binance (with stoploss on exchange):
Then the -5% will be your stop price - but a stoploss-limit order (the only stoploss order supported by binance) has 2 prices, the stop price (-5%) - and the limit price (which by default os -1% from the stoploss, so -6%).

Now your order will usually fill somewhere between -5% and -6%. Sometimes it'll fill at -5.001% - sometimes at -5.99% ... but everything in between is possible.

It get's worse if the exchange uses stoploss-market orders - where you only have the stop-price - and the subsequent order is a market order - which in theory can fill at -9% if market drops fast and has low liquidity (even though your stop-price was -5%).

Now backtesting does not know liquidity at every pricestep - but we know that a fill at exactly -5% is rare (possible, rare - but backtesting aims to be more conservative ...).

Hope this makes sense.

Samaoo:
That makes sense, but I think it should be documented

Samaoo:
I think it would make more sense if the loss on a  backtesting stoploss order would be the average of the stop-price and the limit-price instead of a fixed value

xmatthias:
Honestly, if your strategy's backtest performance  depends on this, you've already lost once your strategy hits real markets, where the orderbook will determine if you're filled or not.

Best familiarize yourself with the [backtesting assumptions](https://www.freqtrade.io/en/latest/backtesting/#assumptions-made-by-backtesting) - which if you think about it carefully will make it clear that backtesting can never be 100% correct.
You can play them to make your strategy look better in backtesting - and for dry / live, you should always assume that these results can be a bit lower - simply because of unfilled orders - and because in reality, you'll rarely get exactly the candle's open price as price for your candle.

---

[...]

PR #4052, opened 2020-12-09
Update backtesting.md

Samaoo:
closes: #3973

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -285,7 +285,7 @@ Since backtesting lacks some detailed information about what happens within a ca
   - sells are compared to high - but the ROI value is used (e.g. ROI = 2%, high=5% - so the sell will be at 2%)
   - sells are never "below the candle", so a ROI of 2% may result in a sell at 2.4% if low was at 2.4% profit
   - Forcesells caused by `<N>=-1` ROI entries use low as sell value, unless N falls on the candle open (e.g. `120: -1` for 1h candles)
-- Stoploss sells happen exactly at stoploss price, even if low was lower, but the loss will be 0.32% lower than the stoploss price.
+- Stoploss sells happen exactly at stoploss price, even if low was lower, but the loss will be 0.32% lower than the stoploss price
 - Stoploss is evaluated before ROI within one candle. So you can often see more trades with the `stoploss` sell reason comparing to the results obtained with the same strategy in the Dry Run/Live Trade modes
 - Low happens before high for stoploss, protecting capital first
 - Trailing stoploss
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -285,7 +285,7 @@ Since backtesting lacks some detailed information about what happens within a ca
   - sells are compared to high - but the ROI value is used (e.g. ROI = 2%, high=5% - so the sell will be at 2%)
   - sells are never "below the candle", so a ROI of 2% may result in a sell at 2.4% if low was at 2.4% profit
   - Forcesells caused by `<N>=-1` ROI entries use low as sell value, unless N falls on the candle open (e.g. `120: -1` for 1h candles)
-- Stoploss sells happen exactly at stoploss price, even if low was lower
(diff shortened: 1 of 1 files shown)


docs/backtesting.md
```diff
@@ -285,7 +285,7 @@ Since backtesting lacks some detailed information about what happens within a ca
   - sells are compared to high - but the ROI value is used (e.g. ROI = 2%, high=5% - so the sell will be at 2%)
   - sells are never "below the candle", so a ROI of 2% may result in a sell at 2.4% if low was at 2.4% profit
   - Forcesells caused by `<N>=-1` ROI entries use low as sell value, unless N falls on the candle open (e.g. `120: -1` for 1h candles)
-- Stoploss sells happen exactly at stoploss price, even if low was lower
+- Stoploss sells happen exactly at stoploss price, even if low was lower, but the loss will be 0.32% lower than the stoploss price
```
xmatthias, line 288: Unfortunately, this is not 100% correct.

It'll be your exchange default fees * 2 (you can test this in backtesting with `--fee 0.0`).
so it'll be 0.32% for you, but might be 0.20% for others if they use a different exchange.

The "calculation error" is there half-intentional (i knew about this for quite some time) - and it makes backtesting a bit more realistic (as discussed in the issue), as your stoploss won't sell exactly at 5%, but a little below than that.
Removing this part from the calculation wouldn't be too difficult i think (would need to look at it) - but i'm indecisive if it'd be a good or bad idea to do, as currently, stoploss is calculated off of "open rate" - not considering fees (in all modes).

e.g.: if you bought at 100$ - your 5% stoploss will be placed at exactly 95$.
Changing this to include fees would put your stoploss at 95.1%  (or so).

Samaoo: I think you partly misunderstood my issue. My problem with the current calculation is that the loss is too low. When I run a hyperopt the 'optimal' stoploss value looks like this ```stoploss = -3e-05```, which is obviously not a good stoploss value. I think it would be better if the loss is customizable in the strategy.

Samaoo: Also, the fee calculation is wrong for some exchanges because some exchanges have different maker/taker fees. For example the calculated fee for kraken is 2 * 0.16= 0.32% but the actual fee is 0.26 + 0.16 = 0.42%.

xmatthias: It's using the default fees as reported by ccxt, as this is the best info we get automatically.

If hyperopt gives you this low a stoploss, then you have stoploss_space in your hyperopt file, and have it misconfigured.
Without customizing it, it will allow values between -0.35 and -0.02, so the value you get above is not possible UNLESS you explicitly allowed hyperopt to go that low.

Samaoo: The problem is that the stoploss is always the lowest allowed value, which means  I can't optimize my stoploss. If the loss would be customizable in the strategy fix that and the stoploss wouldn't be that low.

xmatthias: if your strategy fails because of 0.3% on the stoploss - then your strategy is wrong and will not work anyway.

Samaoo: My strategy works in backtesting, hyperopt and dry-running, but I am sure it won't work when actually trading, but I can't test that yet.

xmatthias: Feel free to do that. 
I don't see the benefit of investing the time to add this, as the "loss" you're referring to is usually somewhere between 0.2% and 0.5%. 
If a strategy relies on this low amount of change to be profitable, it's doomed to fail in live anyway, as backtesting must take quite a few assumptions (on price, and price movements).

Also in live you'll notice that this slippage on stoploss orders is not really controllable. 
Sure, you can set the maximum limit you allow between stop price and limit price - but if the priceaction "skips" that level (or you've got this too tight), then you're stuck in the trade (which is not simulated in backtesting at all as we don't have the data for this).

xmatthias (approved):
LGTM :+1: 

This way we have it documented. I'd leave the decision on changing the calculation method for this for another time, as it WILL also have an impact on how the stop-price in live mode is calculated.

PR #4089, opened 2020-12-20
introduce custom stoploss

### LINKED ISSUE #1193: trailing_stop value relative to profit gains
Current situation:
"trailing_stop_positive":  0.01,
"trailing_stop_positive_offset":  0.05,
if profit reach 10%, trailong stop will be enabled. 
if profit reduced 5%, it call sell.

Negative:
It can't catch big up trend like (go up 5%, go down 2%, go up 5%, down down 2%)

=============================================

New Suggestion:
trailing_stop enabler: 10%
profit reduce checker: 50%

Example:
buy on 100
case 1:price rise 10% to 110, if profit (10%) *reduced*  50%( i.e. 10% * 50% = 5% ), it sell
case2: price rise 10% and didn't go back. Continue to 20%. If profit(20%) *reduce* 50% ( i.e 20% * 50% = 10%), it sell

Explain:
More profit more trailing_stop. 
Case 1 stop point is 5%. 
Case 2 stop point is 10%
It helps to catch big up trend

### LINKED ISSUE #3206: Make stoploss as dynamic as roi
My suggestion is to change `stoploss` from float to a dictionary so it can be programmed to self-adjust in dwindling trades. The goal of this change is to give more flexibility to strategies and to reflect time passed in it. For example low timeframe strategies would have a stoploss that tightens with time to decrease risk, as the trade would be out of its planned timeframe.

Example:

```
stoploss = {

        "0": 0.20,
        "10": 0.15,
        "20": 0.10,
        "30": 0.05,
}
```

### LINKED ISSUE #3664: Setting explicitely stoploss value inside "populate_buy_trend" function
For some strategies, I would like to be able to set dynamically a stop loss value depending of some data like indicators.

For instance, when I detect a Buy situation, I would like to set stop loss according to this Buy situation like setting stop loss equals to lower bollinger band

Same for ROI, I would like to set it as upper bollinger band on the Buy context.

Is there any workaround to adress this need, or do we need to have an additional feature in Freqtrade ?

Thanks in advance for your replies !

xmatthias:
## Summary
Allow the strategy to define a custom stoploss function (`custom_stoploss()`).
This stoploss will act as trailing stoploss - and will have to be returned set relative to the current profit

The method is called once every iteration. Stoploss is adjusted accordingly the same way trailing stoploss works.
Stoplosses can only ever move up, never down.

The initial stoploss (`stoploss=0.10`)  serves as hard lower limit, and it's recommended to disable trailing_stoploss to avoid conflicts.

closes #1193
closes #3206
closes #3664

## Quick changelog

- Allow the strategy to change the stoploss via custom callback
- Needs to be explicitly activated via `use_custom_stoploss`
- works in backtesting (* special care needed when working with dates)


docs/strategy-advanced.md
```diff
@@ -8,11 +8,111 @@ If you're just getting started, please be familiar with the methods described in
...
+
+``` python
+    use_custom_stoploss = True
+
+    def custom_stoploss(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
+                        current_profit: float, **kwargs) -> float:
+        # TODO: Add full docstring here
+        return 0.04
+```
+
```
leifjones: Could the example (the `E.g....` part) maybe be moved to after the example?

I'm reviewing the documentation before the code and so far it's a _bit_ hard for me to follow what the example is saying. I imagine it might be clearer if there's the context of the example 0.04 value in the strategy > `custom_stoploss`

tests/strategy/test_default_strategy.py
```diff
@@ -23,3 +27,18 @@ def test_default_strategy(result):
...
+        open_rate=19_000,
+        amount=0.1,
+        pair='ETH/BTC',
+        fee_open=fee.return_value
+    )
+
+    assert strategy.confirm_trade_entry(pair='ETH/BTC', order_type='limit', amount=0.1,
+                                        rate=20000, time_in_force='gtc') is True
+    assert strategy.confirm_trade_exit(pair='ETH/BTC', trade=trade, order_type='limit', amount=0.1,
+                                       rate=20000, time_in_force='gtc', sell_reason='roi') is True
```
leifjones, line 41: Since these methods aren't implemented in DefaultStrategy, these are just testing the IStrategy methods. Should they be in test_interface.py instead? Or are here tocatch unintended edits to the default?

docs/strategy-advanced.md
xmatthias: i suspect it will depend on the background - if you're more familiar with the code, then having the code first makes it simpler - if you wanna understand the logic first, then look at the code how it's done, then this way might be better...

i will need to add the docstring to the docs too, anyway ... so i'll think about how to make this clearer :+1:

docs/strategy-advanced.md
```diff
@@ -8,11 +8,111 @@ If you're just getting started, please be familiar with the methods described in
...
+                        current_profit: float, **kwargs) -> float:
+        # TODO: Add full docstring here
+
+        # Calculate as `-desired_stop_from_open + current_profit` to get the distance between current_profit and initial price
+        if current_profit > 0.40:
+            return (-0.25 + current_profit)
+        if current_profit > 0.25:
+            return (-0.15 + current_profit)
+        if current_profit > 0.20:
+            return (-0.7 + current_profit)
```
leifjones, line 121: If I'm following, this is an example of step-wise varying trailing stoploss. 

For this to address #1193 's proportional trailing stoploss, would it be something like the following? (It uses values from @yaotian's original message.)

``` python
    use_custom_stoploss = True
    trailing_stop_enabler = 0.10
    profit_reduce_checker = 0.50

    def threshold_reached(self, pair: str, trade: Trade, current_rate: float) -> boolean:
        already_reached = check_new_column_in_db_for_this_trade()
        if (already_reached):
            return true
        if (current_rate > trailing_stop_enabler):
            set_new_column_in_db_for_this_trade(true)
            return true
        return false

    def custom_stoploss(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
                        current_profit: float, **kwargs) -> float:
        if (threshold_reached()):
            return (current_profit - profit_reduce_checker * current_profit)
        return 1
```

My personal opinion on alternate names for those config values is:
- `trailing_stop_enabler` --> `proportial_trailing_stoploss_threshold`
- `profit_reduce_checker` --> `proportial_trailing_stoploss_ratio`

I suppose adding a new column and setting/getting its values in the `trades` table involves changes to model.py and its tests.

... Or is there a simpler way?

tests/strategy/test_interface.py
```diff
@@ -288,6 +287,77 @@ def test_min_roi_reached3(default_conf, fee) -> None:
...
+        (0.05, 0.95, SellType.NONE, False, True, -0.06, 0.95, SellType.TRAILING_STOP_LOSS, None),
+        (0.05, 1, SellType.NONE, False, True, -0.06, 1, SellType.TRAILING_STOP_LOSS,
+         lambda **kwargs: -0.05),
+        (0.05, 1, SellType.NONE, False, True, 0.09, 1.04, SellType.NONE,
+         lambda **kwargs: -0.05),
+        (0.05, 0.95, SellType.NONE, False, True, 0.09, 0.98, SellType.NONE,
+         lambda current_profit, **kwargs: -0.1 if current_profit < 0.6 else -(current_profit * 2)),
+        # Error case - static stoploss in place
+        (0.05, 0.9, SellType.NONE, False, True, 0.09, 0.9, SellType.NONE,
+         lambda **kwargs: None),
```
leifjones, line 310: I don't yet follow what these are after, but probably will if/once I get to give this another glance later.

mrsegen (commented):
I left a suggestion/question regarding #1193 (nitty gritty specifics of proportional trailing stoploss - the only of the linked issues that I'm currently following).

I haven't yet digested all the changes to the interface and the parameterized test cases.  I'm guessing they - at least partially - address those other linked issues.

I intend to come back, but thought I'd give the partial review while its still fresh in our minds.

docs/strategy-advanced.md
leifjones: Yeah, it makes sense to be friendly to someone not familiar with code. I guess I'm just looking for clarification on the example. And I wasn't sure what relationship - if any - there is between the example and the code snippet.

tests/strategy/test_default_strategy.py
xmatthias: they are here intentionally - i know `DefaultStrategy` does not implement them - but if they are not implemented, it must fall back to the "parent" implementation and not cause any problems - which wasn't explicitly tested before.

The location is ... not that relevant, but the tests run slow enough already, so adding another testcase would only slow the test-suite down further - and this seemed to be a nice spot for this test :D

tests/strategy/test_interface.py
xmatthias: it's different test cases (with, without trailing, with/without custom) ... 
Each of these lines are passed as arguments to the test below, basically creating 10(ish) different testcases, where the test is always the same, just the paramters and expected outcomes are slightly different.

docs/strategy-advanced.md
xmatthias: i don't see what the new column would track, and i honestly don't understand your logic ... 

`custom_stoploss()` will only allow you to move the stoploss upwards, so this "tracking" is done for you.
If you return 5% in one iteration, and 10% in the next iteration, stoploss will not move (*unless price moved by +5% in this iteration).


assume you bought at 100$ - and initial stop is at 90$...
now you move stop to 95$ - which means you can never place a new stop < 95$.
Whatever you return - if it's below the latest stop, it'll be ignored (you return a ratio relative to current profit - but the stoploss is in the end an absolute value) 

If it's higher tan the existing stop (say it comes out at 95.5$ - then it'll be replaced).

---

The simplest case is to return a static value (say 5%) - with profits for the trade raising, you also raise stoploss. If the profit (current price) falls down - the stoploss will remain where it is, eventually triggering if price falls by 5% from the highest reached price.

This would be identical to a regular trailing_stop with 5% ... but you are now given the possibility to return different values based on "whatever".

docs/strategy-advanced.md
xmatthias: well in this case - there isn't :D 

the "real" examples start below this point ... this was there just to demonstrate the function header more than anything... i'll change this slightly

docs/strategy-advanced.md
xmatthias: The example you should be looking at for #1193 is the `Absolute stoploss` one ... while it doesn't use the same values / numbers, it uses a similar approach.

leifjones: Aha... yes, with the only-rising moving custom stoploss column, I agree that a `threshold_reached` column wouldn't be needed, and that my approach could be refactored to utilize the custom stoploss column.

I noticed I made a critical omission regarding my example above being _proportional_. I just edited it from:

`return (current_profit - profit_reduce_checker)`

to:

`return (current_profit - profit_reduce_checker * current_profit)`

Summarizing, I don't currently think there's an issue with the code. Though, I may still tweak and propose and additional example to detail how to address #1193 as I understand it.

xmatthias: my point is - the "trailing_stop enabler" the issue talks about, is really just an `if current_profit > 0.10:` ... 
(you only want to start with the proportional stoploss once you reached X% profit - basically it would mimic the behaviour of `trailing_stop_positive_offset` ).

More examples are never wrong - you can make a PR against the `feat/stoploss_custom`  branch - so it'll all be merged at once (or add additional examples later, too).

As always, it's examples ... which means it's there to give you ideas, inspire you - but will never cover all ideas and usecases ... :D

leifjones: 💯 No blockers from me, then. Just won't approve until I get to review the rest more thoroughly.

When I get to through together another example 🤞, I'll PR it to this branch if it hasn't been merged yet, or just to dev if it has.

xmatthias: i don't intend to merge this "soon", i suspect it'll be a PR at least until friday, more likely until saturday or sunday,a s i don't intend to include this in the december release...

PR #4116, opened 2020-12-24
Develop squat

squat0001:
second try

## Summary
add option to skip timestamps backtesting export files
add stats csum_min/max to backtesting result

### DIFF
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -383,3 +383,21 @@ def calculate_max_drawdown(trades: pd.DataFrame, *, date_col: str = 'close_date'
     high_date = profit_results.loc[max_drawdown_df.iloc[:idxmin]['high_value'].idxmax(), date_col]
     low_date = profit_results.loc[idxmin, date_col]
     return abs(min(max_drawdown_df['drawdown'])), high_date, low_date
+
+
+def calculate_csum(trades: pd.DataFrame) -> Tuple[float, float]:
+    """
+    Calculate min/max cumsum of trades, to show if the wallet/stake amount ratio is sane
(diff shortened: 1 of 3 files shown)


xmatthias:
History is kinda messed up... i suspect you based your commit on an old version of stable / master (not sure though).

since i've had to explain this just a few days ago ... :

```
# ensure you're on the right branch
git checkout telegram 
# Make backup
git checkout -b telegram_bak

git checkout develop

# This assumes you're using origin as upstream for THIS repository (you can check with `git remote -v`
git reset --hard origin/develop

# Now it get's a bit messy ... as i'm not sure how your above history came to play.

# Go to this branch again
git switch develop-squad

# Try rebase, abort if you get a lot of merge conflicts...
git rebase develop

# Check git log - it should show only your 10ish commits, and before that something like the below (notice the "origin/develop" after the commit

git log --oneline 

# If this did work, forcepush this branch 
git push -f
```

If this doesn't work let me know where the problems lie and we'll try from there.

Apart from git history:
I'd want to see this as 2 seperate PR's, one changing the output report - and the other trying to fix the timestamp-appending ... that'll make reviewing it easier as it's clear which change belongs to which "change idea" 

Hope this makes sense

xmatthias:
for now, i've fixed the history in this.

I'm not convinced about the new parameter for the backtest file though. 
A probably simpler method (from a user perspective) would be to use the "timestamp-based" filename in the default case or if a directory is specified - and if a filename is specified, use this without further modification.
That would remove the need for an additional parameter ...

xmatthias (approved):
I've now rebased this once more.
I've also removed the "path-fixing" section, as i'm currently uncertain if this is the way to go.

The remainder however (cumsum min/max for backtesting) seems too valuable to keep waiting any longer.

Hope this works for you too.

PR #4131, opened 2020-12-29
Fix/krakenbalance

xmatthias:
## Summary
Kraken reports wrong "free" balance if an open buy order exists

Solve the issue: as reported in Discord

## Quick changelog

- Don't blindly use "remaining", but multiply with price if it's a buy order.


wafula99:
Hello guys

PR #4197, opened 2021-01-12
Added support for regex in whitelist

### LINKED ISSUE #4177: UP/DOWN wildcards for whitelist
## Describe your environment
(if applicable)

  * Operating system: ____
  * Python Version: _____ (`python -V`)
  * Freqtrade Version: ____ (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)

## Describe the enhancement

After having the "UP/DOWN" token wildcard feat. for the blacklist, it would be very helpful having the same for the whitelist to filter / unfilter leveraged markets

nas-:
Introduce regex whitelist support

closes #4177


freqtrade/edge/edge_positioning.py
```diff
@@ -80,10 +81,12 @@ def __init__(self, config: Dict[str, Any], exchange, strategy) -> None:
         if config.get('fee'):
             self.fee = config['fee']
         else:
-            self.fee = self.exchange.get_fee(symbol=self.config['exchange']['pair_whitelist'][0])
+            self.fee = self.exchange.get_fee(symbol=expand_pairlist(
+                self.config['exchange']['pair_whitelist'], list(self.exchange.markets))[0])
 
     def calculate(self) -> bool:
-        pairs = self.config['exchange']['pair_whitelist']
+        pairs = expand_pairlist(self.config['exchange']['pair_whitelist'],
```
xmatthias, line 88: Nice that you even thought about edge - i suspect i'd have forgotten this!

freqtrade/exchange/exchange.py
```diff
@@ -335,8 +336,9 @@ def validate_pairs(self, pairs: List[str]) -> None:
         if not self.markets:
             logger.warning('Unable to validate pairs (assuming they are correct).')
             return
+        extended_pairs = expand_pairlist(pairs, list(self.markets))
         invalid_pairs = []
-        for pair in pairs:
+        for pair in extended_pairs:
```
xmatthias: i think we may need to find a way to identify wildcard strings and remove them from this list, keeping the "original" check in place here.

Currently, i can start the bot with a pair `'AAAAAAAA/USDT` on binance without any warning or notice - it's simply silently removed without telling me it's not taken.
While in my case, it's intentional so not a problem, it can also lead to subtle errors, where users won't notice that a pair has not been taken into consideration.

the current (expected) behaviour would be that the below exception (pair not available) is raised.
To have this work correctly though, we'd need to identify "this contains a wildcard" and exclude these from this check (which i suspect might not be straightforward).

freqtrade/plugins/pairlist/StaticPairList.py
```diff
@@ -50,9 +50,10 @@ def gen_pairlist(self, cached_pairlist: List[str], tickers: Dict) -> List[str]:
         :return: List of pairs
         """
         if self._allow_inactive:
-            return self._config['exchange']['pair_whitelist']
+            return self.verify_whitelist(self._config['exchange']['pair_whitelist'], logger.info)
```
xmatthias: I think this is a special case (when `allow_inactive = True`), in which we'd not wan't to expand the whitelist.

Problem here is, the whitelist expansion implicitly checks if the pair exists.
This flag was however explicitly added in #3939 to allow backtesting no longer existing pairs, which this will basically prevent.

Obviously, we'd need to add a note (`pairlist.md`) that this will prevent whitelist expansion to the docs ...

tests/exchange/test_exchange.py
```diff
@@ -505,37 +505,38 @@ def test_validate_pairs(default_conf, mocker):  # test exchange.validate_pairs d
...
-    mocker.patch('freqtrade.exchange.Exchange._load_async_markets')
-
-    with pytest.raises(OperationalException, match=r'Pair ETH/BTC is not available on Binance'):
-        Exchange(default_conf)
-
-    mocker.patch('freqtrade.exchange.Exchange.markets', PropertyMock(return_value={}))
-    Exchange(default_conf)
-    assert log_has('Unable to validate pairs (assuming they are correct).', caplog)
+# This cannot happen anymore as expand_pairlist implicitly filters out unavaliablie pairs
+# def test_validate_pairs_not_available(default_conf, mocker):
```
xmatthias: See above note why i think this might not be a good idea ...

xmatthias (commented):
Overall very nice PR! :+1: 

There's one big caveat with pair validation though which i think we'll want to address / keep existing behaviour (kindof) - which makes whitelist more difficult than blacklist implementation (see comment on `exchange.py`)

Also, this should be added to the documentation (that `StaticPairList` and edge now supports regex pairlists) ...

freqtrade/exchange/exchange.py
nas-: I added a parameter to expand_pairlist, keep_invalid default to false.
This way in exchange, i can keep invalid pairs which might be present and it raises the exception if you input something like AAAAAAAA/USDT

freqtrade/plugins/pairlist/StaticPairList.py
nas-: From there, I called expand_pairlist with parameter keep invalid = True, so that invalid pairs are maintained in the whitelist.

tests/exchange/test_exchange.py
nas-: I reverted this, now this test passes, as expand_pairlist keeps the invalid pairs and so the exceptions are raised, if needs to be.

nas-:
Fairly sure I messed up while pushing my changes, and pushed a bunch of old commits + my changes....

xmatthias:
mmmmh, you should've created a seperate branch for the changes, i think you pulled the upstream develop on top of your develop branch - which creates a mess.

I'll try to fix history and force-push to your branch to get this fixed.

freqtrade/plugins/pairlist/StaticPairList.py
xmatthias: Great way to solve this! :+1:

xmatthias (approved):
LGTM :+1: 

Great job and thanks for taking the time to impelement this!

PR #4228, opened 2021-01-17
Extending the Telegram Bot command /status with the possibility to query specific trade_ids

baartch:

## Summary
Extending the Telegram Bot command /status with the possibility to query specific trade_ids

Solve the issue: #___

## Quick changelog

- In module `telegram.py` extend the `_status` method with verifying provided arguments as numbers, and provide these numbers as an array to `_rpc._rpc_trade_status`
- In module `rpc.py` extend the `_rpc_trade_status` with an additional argument for the trade_ids and query these trade_ids if provided.
- Extend the docs and the Telegram `/help` command.

## What's new?
New is the possibility to query via Telegram status details of specific trades.
unknown

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -144,7 +144,7 @@ class RPC:
         }
         return val
 
-    def _rpc_trade_status(self, trade_ids=None) -> List[Dict[str, Any]]:
+    def _rpc_trade_status(self, trade_ids: List[int] = []) -> List[Dict[str, Any]]:
         """
         Below follows the RPC backend it is prefixed with rpc_ to raise awareness that it is
         a remotely exposed function
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -151,7 +151,7 @@ class RPC:
         """
         # Fetch open trades
         if trade_ids:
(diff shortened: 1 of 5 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -144,13 +144,17 @@ def _rpc_show_config(config, botstate: Union[State, str]) -> Dict[str, Any]:
         }
         return val
 
-    def _rpc_trade_status(self) -> List[Dict[str, Any]]:
+    def _rpc_trade_status(self, trade_ids=None) -> List[Dict[str, Any]]:
```
xmatthias: i think we might want to type this with `trade_ids: List[int] = []`  ... (check mypy output after this to see if that causes problems).

baartch: Thank you for your help. I changed it like you suggested and then, with another change in the `telegram.py` it passed the mypy tests.

xmatthias (approved):
Great job - nice feature :+1: 

i've added a small test just to make sure we'll not accidentaly break this in some refactorings in the future.

baartch:
> Great job - nice feature 
> 
> i've added a small test just to make sure we'll not accidentaly break this in some refactorings in the future.

Cool! Thank you for all the work you are putting into this project!

PR #4256, opened 2021-01-25
Fix operator does not exist: boolean = integer

thopd88:
## Summary
Small fix "operator does not exist: boolean = integer"

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -200,7 +200,7 @@ class FreqtradeBot(LoggingMixin):
         Notify the user when the bot is stopped
         and there are still open trades active.
         """
-        open_trades = Trade.get_trades([Trade.is_open == 1]).all()
+        open_trades = Trade.get_trades([Trade.is_open == True]).all()
 
         if len(open_trades) != 0:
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -200,7 +200,7 @@ def check_for_open_trades(self):
         Notify the user when the bot is stopped
         and there are still open trades active.
         """
-        open_trades = Trade.get_trades([Trade.is_open == 1]).all()
+        open_trades = Trade.get_trades([Trade.is_open == True]).all()
```
xmatthias:
```suggestion
        open_trades = Trade.get_trades([Trade.is_open.is_(True)]).all()
```

Aside from this ... 
what do you have to do to actually reach this error? I've never seen this, but it seems like a problem that SHOULD appear right away ...

thopd88: Hi, I used psycopg2 to connect to PostgreSQL

xmatthias (approved):
LGTM :+1:

PR #4282, opened 2021-01-31
Fix a bug when compare sell_profit_offset

pan-long:
## Summary
It should be comparing the ratio instead of absolut profit.

Also updated the comment.

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -530,7 +530,7 @@ class IStrategy(ABC):
                                                 current_time=date))
 
         if (ask_strategy.get('sell_profit_only', False)
-                and trade.calc_profit_ratio(rate=rate) <= ask_strategy.get('sell_profit_offset', 0)):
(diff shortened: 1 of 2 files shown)


freqtrade/strategy/interface.py
```diff
@@ -530,8 +530,8 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
                                                 current_time=date))
 
         if (ask_strategy.get('sell_profit_only', False)
-                and trade.calc_profit(rate=rate) <= ask_strategy.get('sell_profit_offset', 0)):
-            # Negative profits and sell_profit_only - ignore sell signal
+                and trade.calc_profit_ratio(rate=rate) <= ask_strategy.get('sell_profit_offset', 0)):
```
xmatthias: a probably better way would be to use `current_profit` - which is the already calculated version of this.

xmatthias (approved):
Nice catch! 👍 

LGTM

PR #4308, opened 2021-02-03
Add Refresh / Reload Button on rpc/Telegram

eatrisno:

## Summary
when user request a status message, they need to click it again to get updated message, but it will create a mess 
the proposed solution is, add reload / refresh button, so the message will be updated 
Solve the issue: #___

## Quick changelog

- add refresh button on telegram message
## What's new?


xmatthias:
Before moving on with this ...

are you able to post a screenshot showing how this would look like (and to which telegram commands you are planning to use this with)?


i think the biggest problem when applying this to `/status` is that `/status` will really send you multiple messages (one per trade).
so you'd either need to refresh every trade once - or simply request this again.

with some other commands it MAY make sense - although i'm not sure if refreshing won't create a bigger mess than getting new messages (which will allow you to scroll back to see historic values if you need).

eatrisno:
alt text

it simply adds Refresh button on status, status table message, balance, etc

this is for /status, there is no refresh button
alt

so, I added the Refresh button to a few function that didn't loop
/daily /profit /balance /performance /count /status table

it still able to request again to scroll back and get historic values

xmatthias:
is it possible to make this optional easily (if it's a lot of effort, don't)? (so it can be switched off via config?)

I'd leave it on by default - but depending on how you use telegram, it can become anoying for users eventually, as the button will always be there, even if i don't really want it.

Also, how will this react if i get a response, restart the bot - and THEN click the refresh button? will it still update the old version?

eatrisno:
> is it possible to make this optional easily (if it's a lot of effort, don't)? (so it can be switched off via config?)
> 
> I'd leave it on by default - but depending on how you use telegram, it can become anoying for users eventually, as the button will always be there, even if i don't really want it.
> 
> Also, how will this react if i get a response, restart the bot - and THEN click the refresh button? will it still update the old version?

Sure, I'll update to able make it optional

nothing will happen, it just requests the update_status path for example. I'm using it right now

thanks

mehdico:
Hey @eatrisno,

I suggest bringing back the `resize_keyboard=True` at line # 982.

xmatthias:
> I suggest bringing back the `resize_keyboard=True` at line # 982.

It's not technically been removed in this PR - but resize_keyboard was introduced AFTER this PR was opened. 
You'll also notice that there are conflicts, which will obviously need to be resolved - and which will introduce this functionality into this branch if done carefully.

eatrisno:
sorry for the slow response, I'm moving to the new company so quite busy

so, 
I have made changes and solved the conflict

the reload button will set True by default, to disable it, we need to add reload: false to the telegram section in the configuration file, Example below
```
 "telegram": {
        "enabled": false,
        "chat_id": "your_telegram_chat_id",
        "reload": false
    },
```

this feature still works after you change the reload to false then click reload on the previous chat, of course, the new message won't show reload button.

cheers

eatrisno:
@xmatthias Hi, do I have to pass the CI to get merged?

xmatthias:
> @xmatthias Hi, do I have to pass the CI to get merged?

absolutely, that's the point of it - if we'd merge PR's breaking CI, we could just remove it as well as then it would make no sense.


While i did have a glance over this (it looks ok in general), i didn't do a full review on this - as i don't see a PR as complete until CI passes (as making CI pass can mean fixing bugs, or adjusting tests - or both).

xmatthias (approved):
Great contribution :+1: 
I've cleaned this up some - but the functionality is great i think.

Thanks a lot! :100:

eatrisno:
Awesome!! Cheers

PR #4313, opened 2021-02-04
setting resize_keyboard=True for slightly smaller Telegram buttons

raoulbuzziol:
## Summary
Telegram button are just slightly smaller.
Existing tests are passing, no tests about button size present (because testing an external library is out of scope).

## Quick changelog
- set resize_keyboard=True on ReplyKeyboardMarkup


xmatthias (approved):
Thanks for this! :+1: 

Didn't know this was possible...

PR #4321, opened 2021-02-05
chore(lint): lint binance example config

TheJoeSchr:
## Summary
so it's in line with autogenerated config and therefor easier to diff.


xmatthias (approved):
LGTM 👍

PR #4322, opened 2021-02-05
Configuration.md - document structuring

sobeit2020:

## Summary
Lets work a bit on the structure and definitions of the elements which are based in configuration file.

## What's new?
* Table at the beginning of the file, became divided into sections, which i took from understanding how `config.json` file works.
* Definitions of the following options will be step by step, day by day cleared out

* important is to know, that rows 273-279 and 472-600 are deleted, but i do not have to do with it anything. That is some old story, you probably know. It would be good to clear it out on the way.


xmatthias:
you will want to use a seperate branch (not develop) for your pull requests. Otherwise you'll block yourself from doing updates to your installation.

sobeit2020:
Yo, I had noticed it the hard way. I am on it. Lets keep in touch. After i handle this, i will continue with Config.md

PR #4323, opened 2021-02-06
Fix sample strategy documentation link

eselskas:
Noticed that the current link is dead. I think this would be the most appropriate link in this case.

## Summary

Solve the issue: Fix broken documentation link

## Quick changelog

- Fixes sample strategy documentation link

## What's new?
n/a

### DIFF
--- a/freqtrade/templates/sample_strategy.py
+++ b/freqtrade/templates/sample_strategy.py
@@ -17,7 +17,7 @@ import freqtrade.vendor.qtpylib.indicators as qtpylib
 class SampleStrategy(IStrategy):
     """
     This is a sample strategy to inspire you.
-    More information in https://github.com/freqtrade/freqtrade/blob/develop/docs/bot-optimization.md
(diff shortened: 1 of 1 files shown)


freqtrade/templates/sample_strategy.py
```diff
@@ -17,7 +17,8 @@
 class SampleStrategy(IStrategy):
     """
     This is a sample strategy to inspire you.
-    More information in https://github.com/freqtrade/freqtrade/blob/develop/docs/bot-optimization.md
+    More information in:
+    https://github.com/freqtrade/freqtrade/blob/develop/docs/strategy-customization.md
```
xmatthias: Honestly, i'd rather use https://www.freqtrade.io/en/latest/strategy-customization/ ... 

That's the "rendered" version for this - which is ususally the better option

eselskas: Yes looks cleaner :+1: Link updated

xmatthias (approved):
LGTM :+1:

Issue #4345, opened 2021-02-10
Incorrect Est. Balance for the pair BTCST


varunrai:
* Operating system: Ubuntu
  * Python Version: 3.7
  * CCXT version: 1.41.63
  * Freqtrade Version: 2021.1

The balance displayed in the telegram bot for the pair BTCST/BTC is incorrect. 

### Steps to reproduce:

  1. Configuration - Stake Currency set to BTC and the pair BTCST/BTC is whitelisted or any other pair name that starts with BTC..
  2. Configure the telegram bot
  3. Start the trade bot and request for the balance via telegram
  
### Observed Results:

  * The balance display for the Est. Balance under the currency and the final Estimated Value are incorrect (see the attached screenshot).

### Relevant code exceptions or logs

the code in rpc.py in the function _rpc_balance [(Link)] (https://github.com/freqtrade/freqtrade/blob/766c786d9088730ab617948625bd573b2641249f/freqtrade/rpc/rpc.py#L453) assumes the pair name starts with stake_currency (in my case BTC) and the pair name is BTCST/BTC which will result in the incorrect calculation being applied to the Est. Balance.

```
if pair.startswith(stake_currency):
   rate = 1.0 / rate
   est_stake = rate * balance.total  

```

![Telegram_h7BDSfttae](https://user-images.githubusercontent.com/1327311/107456323-f3826500-6b8a-11eb-8f25-f557ee2aae46.jpg)

Once of the possible ways to fix would be to use the following code:
```
if pair.startswith(stake_currency + "/"):
   rate = 1.0 / rate
   est_stake = rate * balance.total  

```

xmatthias:
I'm going to close this as it's a duplicate of #4307 (i'm sorry for your efforts with the PR ...).

It's fixed in develop since 5 days now ... and will "age" into the release around the end of the month.

varunrai:
> I'm going to close this as it's a duplicate of #4307 (i'm sorry for your efforts with the PR ...).
> 
> It's fixed in develop since 5 days now ... and will "age" into the release around the end of the month.

No worries. I should have checked the other issue :)

PR #4346, opened 2021-02-10
fix the rpc for balance which returns incorrect est. balance 

varunrai:
## Summary
fix the rpc for balance which returns incorrect est. balance for BTC pairs that starts with BTC such as BTCST/BTC

Solve the issue: #4345 

## Quick changelog

- Updated the function to include the '/' when performing a startswith comparison.

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -451,7 +451,7 @@ class RPC:
                     pair = self._freqtrade.exchange.get_valid_pair_combination(coin, stake_currency)
                     rate = tickers.get(pair, {}).get('bid', None)
                     if rate:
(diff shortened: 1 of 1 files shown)


xmatthias:
i don't think this fix will work correctly as you're assuming the pair seperator is a / - which is not always the case on some (random) exchanges.

in my opinion, this is fixed in develop already by also checking if the pair ends with stake-currency (which i pushed a few days ago)... (end of the line you modified), unless you say it's also a problem in develop which i did miss.

xmatthias:
can we close this one? i think this problem does no longer exist in develop - or did i miss something?

varunrai:
yes. i will close this.

PR #4352, opened 2021-02-11
Solve Asynchronous Execution in Jupyter Notebook

Youbadawy:

## Summary

Was unable to initate a new loop in jupyter notebook due to Asynchronous execution

Solve the issue: #___

## Quick changelog

- created a new task using the asyncio module, instead of run_until_complete replaced with create_task (or run_coroutine_threadsafe would also work)

as per below stack:
https://stackoverflow.com/questions/47518874/how-do-i-run-python-asyncio-code-in-a-jupyter-notebook

## What's new?

we can use ccxt modules and classes in jupyter notebook

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -142,7 +142,7 @@ class Exchange:
         """
         logger.debug("Exchange object destroyed, closing async loop")
         if self._api_async and inspect.iscoroutinefunction(self._api_async.close):
-            asyncio.get_event_loop().run_until_complete(self._api_async.close())
+            asyncio.get_event_loop().create_task(self._api_async.close())
 
     def _init_ccxt(self, exchange_config: Dict[str, Any], ccxt_module: CcxtModuleType = ccxt,
                    ccxt_kwargs: dict = None) -> ccxt.Exchange:
@@ -282,7 +282,7 @@ class Exchange:
     def _load_async_markets(self, reload: bool = False) -> None:
         try:
             if self._api_async:
-                asyncio.get_event_loop().run_until_complete(
+                asyncio.get_event_loop().create_task(
                     self._api_async.load_markets(reload=reload))
 
         except (asyncio.TimeoutError, ccxt.BaseError) as e:
@@ -720,7 +720,7 @@ class Exchange:
         :param since_ms: Timestamp in milliseconds to get history from
         :return: List with candle (OHLCV) data
         """
-        return asyncio.get_event_loop().run_until_complete(
+        return asyncio.get_event_loop().create_task(
             self._async_get_historic_ohlcv(pair=pair, timeframe=timeframe,
                                            since_ms=since_ms))
 
@@ -799,7 +799,7 @@ class Exchange:
                     pair, timeframe
                 )
 
-        results = asyncio.get_event_loop().run_until_complete(
+        results = asyncio.get_event_loop().create_task(
             asyncio.gather(*input_coroutines, return_exceptions=True))
 
         results_df = {}
@@ -1029,7 +1029,7 @@ class Exchange:
         if not self.exchange_has("fetchTrades"):
             raise OperationalException("This exchange does not suport downloading Trades.")
 
-        return asyncio.get_event_loop().run_until_complete(
+        return asyncio.get_event_loop().create_task(
             self._async_get_trade_history(pair=pair, since=since,
                                           until=until, from_id=from_id))
 


xmatthias (changes requested):
I'm sorry, but this way I cannot accept this (not only because tests fail).

`create_task()` has been added in python 3.8, but we support 3.7 (and we will do so for quite some time still).

It's also not clear to me if `create_task()` wait's for the execution to complete (it doesn't seem to from reading the documentation, which talks about "schedules the task" - not "schedules and executes and waits for completion of the task".

Youbadawy:
@xmatthias thanks for responding so quickly, kindly find below the python docs version 3.7 of the given function,

https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.loop.create_task

it starts a new task in the initiated loop, not a new loop, so it should wait for the compeltetion of the previous task. 

Also the second function  run_coroutine_threadsafe does exactly what you mentioned or wht the run_until_completion function does and waits for the futures,

https://docs.python.org/3.7/library/asyncio-task.html#asyncio.run_coroutine_threadsafe

so in case thats an essential requirement, we could change it to run_coroutine_threadsafe,

Hope it helps, please excuse my newbiness in creating pull requests, first time I do one in this project, so unaware of the test requirements.

xmatthias:
You're right, i misread - it's the "name" parameter that was added in 3.8, not this function itself.

> please excuse my newbiness in creating pull requests, first time I do one in this project, so unaware of the test requirements.

that's no problem - in general, tests should pass (if just one of the matrix build fails it's usually no problem - but in this case, all fail).
This can have 2 reasons - either the tests need some adoption to the change - or it points out problems with the change.

---

Based on [this section in the docs](https://docs.python.org/3.7/library/asyncio-task.html#coroutines) - creating the task will NOT execute it unless you await it afterwards - so this will actually break regular behavior, and will never execute the coroutines you created.

Youbadawy:
> 
> 
> You're right, i misread - it's the "name" parameter that was added in 3.8, not this function itself.
> 
> > please excuse my newbiness in creating pull requests, first time I do one in this project, so unaware of the test requirements.
> 
> that's no problem - in general, tests should pass (if just one of the matrix build fails it's usually no problem - but in this case, all fail).
> This can have 2 reasons - either the tests need some adoption to the change - or it points out problems with the change.
> 
> Based on [this section in the docs](https://docs.python.org/3.7/library/asyncio-task.html#coroutines) - creating the task will NOT execute it unless you await it afterwards - so this will actually break regular behavior, and will never execute the coroutines you created.

this is not the same create_task module used,  it is the loop.create_task,

allow me to test the other module, the one that awaits the futures, since that is what we are looking for, if I am not mistaken?

Youbadawy:
> 
> 
> You're right, i misread - it's the "name" parameter that was added in 3.8, not this function itself.
> 
> > please excuse my newbiness in creating pull requests, first time I do one in this project, so unaware of the test requirements.
> 
> that's no problem - in general, tests should pass (if just one of the matrix build fails it's usually no problem - but in this case, all fail).
> This can have 2 reasons - either the tests need some adoption to the change - or it points out problems with the change.
> 
> Based on [this section in the docs](https://docs.python.org/3.7/library/asyncio-task.html#coroutines) - creating the task will NOT execute it unless you await it afterwards - so this will actually break regular behavior, and will never execute the coroutines you created.

You are absolutely right, so what happens now that RuntimeError: await wasn't used with future, I should be using an asyncio.Semaphore to limit the calls, 

I am quite new to all of this,

xmatthias:
To be honest, it's quite difficult to tell you how to approach this - as it's unclear to me what problem you're trying to fix.

Youbadawy:
> 
> 
> To be honest, it's quite difficult to tell you how to approach this - as it's unclear to me what problem you're trying to fix.

Running the freqtrade modules and functions in jupyter notebook, like downloading data, running exchange queries

Youbadawy:
I have been able to find a solution, but it does not work for both cases, either for notebook or for script, Do not think this would be usesful for freqtrade, so i will be closing the pull request.

xmatthias:
The correct solution will probably be to use the following in your jupyter notebook:
https://pypi.org/project/nest-asyncio/

It'll allow the existing functionalities to work also in jupyter - as it's really a jupyter limitation, not one of freqtrade.

The main usecase for freqtrade is NOT in jupyter notebook, so crippling the code too much just to get "function X" work in jupyter is not really a good idea, as you can easily use it via command line.

Youbadawy:
> 
> 
> The correct solution will probably be to use the following in your jupyter notebook:
> https://pypi.org/project/nest-asyncio/
> 
> It'll allow the existing functionalities to work also in jupyter - as it's really a jupyter limitation, not one of freqtrade.
> 
> The main usecase for freqtrade is NOT in jupyter notebook, so crippling the code too much just to get "function X" work in jupyter is not really a good idea, as you can easily use it via command line.

you are absolutely right, thanks for taking the time!

xmatthias:
based on the above discussion, i'll close this ...

PR #4368, opened 2021-02-14
config - add argument 'strategy_params'

flomerz:
## Summary
This PR add the possibility to add own strategy parameters with cli or config.json.
That is helpful to prevent any modification in the strategy file, and it's more convenient to pass strategy parameters over the configuration instead of modifying the strategy file every time.

## What's new?
* new parameter argument "strategy_params"
* also possible to add in the config.json
* strategy_params is able to parse JSON notation

### DIFF
--- a/docs/bot-usage.md
+++ b/docs/bot-usage.md
@@ -54,7 +54,7 @@ optional arguments:
 
 ```
 usage: freqtrade trade [-h] [-v] [--logfile FILE] [-V] [-c PATH] [-d PATH]
-                       [--userdir PATH] [-s NAME] [--strategy-path PATH]
+                       [--userdir PATH] [-s NAME] [--strategy-path PATH] [--strategy-params JSON]
                        [--db-url PATH] [--sd-notify] [--dry-run]
 
 optional arguments:
@@ -85,9 +85,10 @@ Common arguments:
 
 Strategy arguments:
   -s NAME, --strategy NAME
-                        Specify strategy class name which will be used by the
-                        bot.
-  --strategy-path PATH  Specify additional strategy lookup path.
+                          Specify strategy class name which will be used by the
+                          bot.
+  --strategy-path PATH    Specify additional strategy lookup path.
+  --strategy-params JSON  Specify additional strategy params.
 
 ```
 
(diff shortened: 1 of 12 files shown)


docs/bot-usage.md
```diff
@@ -189,6 +190,14 @@ checked before the default locations (The passed path must be a directory!):
 freqtrade trade --strategy AwesomeStrategy --strategy-path /some/directory
 ```
 
+### How to use **--strategy-params**?
+
+This parameter allows you to add parameters of your strategy:
+
+```bash
+freqtrade trade --strategy AwesomeStrategy --strategy-params "{'param1': 5, 'param2': 'something'}"
+```
```
xmatthias, line 199: i think you will want an example on how to access these in the strategy....

I do know how to access it - but regular uses may not.

xmatthias (commented):
It's possible already by setting the paramter in the config (nothing can prevent users from adding 100 new keys there) - but how to access it should be documented better i guess.

---

i think i understand why you'd want to do this - but does it really makes sense to pass json (!) as string via command line?

That seems quite complicated from a usability point of view, and will be very error-prone (which will then cause a lot of questions we'll somehow have to support) - so i'm currently not quite convinced yet that this will give us much benefits without causing a lot of problems.

docs/bot-usage.md
flomerz: Thanks for your quick feedback.

Yes I realized now that it is already possible over config.json, but not over CLI.

The reason why I choosed a JSON representation is because I can just copy the results of the hyperopt runs. I dont think that is to complicated. The parsefun in de Configuration utility takes care  about the conversion from ' to ".

What I really want is the possiblity to pass the strategy params in a separeted way from the config.json. Because I run multiple bots on different exchanges. So the config.json is different for each bot, but the strategy params are all the same and it take some time to synchronize these parameters.

So my ideal use case would be to have an additional json file, only for the strategy params. That's the reason, why i want to pass it in CLI. To be independent.

If you agree I would extend the PR and add an additional config-strategy-params.json file and maybe remove possibility to pass it over CLI.

As a workaround I would handle this file for myself, but I think there has to be a more general way to do that.

xmatthias: I know almost noone reads the docs nowadays - but before spending the time implementing a solution, i would still recommend doing that, because what you need can be EASILY accomplished by using multiple `--config` flags - which will be combined (basically, the last key wins).

So you can have your bot specific config - and the startegy params later.
` --config botconfigA.json --config CoolStrategyParams`

[it's also documented](https://www.freqtrade.io/en/latest/bot-usage/#how-to-use-multiple-configuration-files)

flomerz: Totally agree with you. That works great. Sorry for wasting your time :( I will close this PR.

PR #4375, opened 2021-02-15
pass data and config to loss function

flomerz:
## Summary
Me again :) In this PR i want to add some small changes mainly for the hyperopt loss function. I did some more extended loss function. Which does some kind of cross validation over the the hole timeframe, to prevent overfitting.
So what i did for now is, i read it in by myself:
`Path("user_data") / 'hyperopt_results' / 'hyperopt_tickerdata.pkl'`
But I think that is an ugly workaround :(
So I think I just needed to pass the processed timeframe to the loss function to, to do that.
The same applies for backtesting stats. I want to analyze some splits of the timeframe and combine that to a more meaningfull score.

## Quick changelog

- pass processed timeframe to hyperopt loss function
- pass processed timeframe to backtest result stats

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -546,10 +546,11 @@ class Hyperopt:
 
         )
         return self._get_results_dict(backtesting_results, min_date, max_date,
-                                      params_dict, params_details)
+                                      params_dict, params_details,
+                                      processed=processed)
 
     def _get_results_dict(self, backtesting_results, min_date, max_date,
-                          params_dict, params_details):
+                          params_dict, params_details, processed: Dict):
         results_metrics = self._calculate_results_metrics(backtesting_results)
         results_explanation = self._format_results_explanation_string(results_metrics)
 
@@ -563,7 +564,8 @@ class Hyperopt:
         loss: float = MAX_LOSS
         if trade_count >= self.config['hyperopt_min_trades']:
             loss = self.calculate_loss(results=backtesting_results, trade_count=trade_count,
-                                       min_date=min_date.datetime, max_date=max_date.datetime)
+                                       min_date=min_date.datetime, max_date=max_date.datetime,
+                                       config=self.config, processed=processed)
         return {
             'loss': loss,
             'params_dict': params_dict,
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -550,7 +550,7 @@ class Hyperopt:
                                       processed=processed)
 
     def _get_results_dict(self, backtesting_results, min_date, max_date,
-                          params_dict, params_details, processed: Dict):
+                          params_dict, params_details, processed: Dict[str, DataFrame]):
         results_metrics = self._calculate_results_metrics(backtesting_results)
         results_explanation = self._format_results_explanation_string(results_metrics)
 
--- a/docs/advanced-hyperopt.md
+++ b/docs/advanced-hyperopt.md
@@ -40,6 +40,11 @@ For the sample below, you then need to add the command line parameter `--hyperop
 A sample of this can be found below, which is identical to the Default Hyperopt loss implementation. A full sample can be found in [userdata/hyperopts](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/templates/sample_hyperopt_loss.py).
 
 ``` python
+from datetime import datetime
+from typing import Dict
+
+from pandas import DataFrame
+
 from freqtrade.optimize.hyperopt import IHyperOptLoss
 
 TARGET_TRADES = 600
@@ -54,6 +59,7 @@ class SuperDuperHyperOptLoss(IHyperOptLoss):
     @staticmethod
     def hyperopt_loss_function(results: DataFrame, trade_count: int,
                                min_date: datetime, max_date: datetime,
+                               processed: Dict[str, DataFrame],
                                *args, **kwargs) -> float:
         """
         Objective function, returns smaller number for better results
@@ -81,6 +87,7 @@ Currently, the arguments are:
 * `trade_count`: Amount of trades (identical to `len(results)`)
 * `min_date`: Start date of the timerange used
 * `min_date`: End date of the timerange used
+* `processed`: Dict of Dataframes with the pair as keys containing the data used for backtesting.
 
 This function needs to return a floating point number (`float`). Smaller numbers will be interpreted as better results. The parameters and balancing for this is up to you.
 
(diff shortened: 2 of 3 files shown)


freqtrade/optimize/optimize_reports.py
```diff
@@ -225,7 +225,7 @@ def generate_daily_stats(results: DataFrame) -> Dict[str, Any]:
 
 def generate_backtest_stats(btdata: Dict[str, DataFrame],
                             all_results: Dict[str, Dict[str, Union[DataFrame, Dict]]],
-                            min_date: Arrow, max_date: Arrow
+                            min_date: Arrow, max_date: Arrow, processed: Dict
```
xmatthias: This will not make sense.

The (non-analyzed) backtest data is available as btdata should the raw candles be necessary - and keeping this data around will cause memory leaks when used with multiple strategies (`--strategy-list`).

freqtrade/optimize/backtesting.py
```diff
@@ -445,11 +445,13 @@ def start(self) -> None:
 
         min_date = None
         max_date = None
+        processed = None
         for strat in self.strategylist:
-            min_date, max_date = self.backtest_one_strategy(strat, data, timerange)
+            min_date, max_date, processed = self.backtest_one_strategy(strat, data, timerange)
```
xmatthias: Same as below, causing memory leaks (it's not fully leaking as we know it's still around) - but it'll increase the amount of memory necessary when using `--strategy-list` by at least 1x per strategy.

Since the data is not used anyway, there's no reason to do this.

freqtrade/optimize/hyperopt.py
```diff
@@ -546,10 +546,11 @@ def generate_optimizer(self, raw_params: List[Any], iteration=None) -> Dict:
 
         )
         return self._get_results_dict(backtesting_results, min_date, max_date,
-                                      params_dict, params_details)
+                                      params_dict, params_details,
+                                      processed=processed)
```
xmatthias, line 550: I'm not sure what you'd need this for - and i have quite heavy fears that this will increase the amount of memory needed for hyperopt (hyperopt already has quite heavy memory requirements) - and if the need for memory increases even more it'll become unusable.

xmatthias (changes requested):
Clearly the backtesting part is a no-go due to the memory-problems this will introduce.

In general, i can only recommend to open issues for topics as such, so the changes you would like to make can be discussed upfront, which will help to avoid wasting time on changes that don't make sense.

freqtrade/optimize/optimize_reports.py
flomerz: Agree :+1:  I don't know way I didn't saw this. Maybe because, I was on an older code-base while implementing it. I will remove it.

freqtrade/optimize/backtesting.py
flomerz: Not needed to, because of btdata from above.

freqtrade/optimize/hyperopt.py
flomerz: I don't see why this should create a memory problem, because the data is already there. 
The time it gets free is when the function 'generate_optimizer' is finish. So it is just passed to the loss function before that. So there is no difference in memory usage.
I really need the data in the loss function or is there an other approach?

Second thing what i would need is the config in the loss function. Specially for the fee configuration. Is there a way to access it?

xmatthias: Fee is not part of the configuration (unless you use --fee), but is received from the exchange during startup of hyperopt/backtesting (check the `__init__()` method in backtesting).

for the memory part, i'll have to have a closer look, for the loss function it might be ok.

flomerz: Thanks for your effort :+1: 

I need the fee, because I do not use staking. Instead I have an all-in/all-out strategy.
So I need to calculate the fees based on the cumulative return.

xmatthias: Backtesting (nor hyperopt) does not support compounding at the moment - so your stake-amounts will always be the same, so i'm not quite sure what you'll calculate here...

I'm working on changing this (and it's almost complete - i'm just struggling with a few final calculations i need to adjust to have the result consistent).
so i'm not quite sure why the fee would be relevant (it'll be the same number for each exchange anyway ...)

flomerz: I look up the open/close times of the trades and search these in the btdata. Then I calculate the pct change over all periods and add the fee as penalty to the specific points. After that i calculate the product over all periods and then i get the cumulative return considering all fees.

What do you think about pass the btdata and the config object to the loss function?

xmatthias: as said, i think you're misjudging what config holds about fees.
fee is NOT in the configuration (at least not the fee that's applied).

It is however available as `fee_open` and `fee_close` in the trade results ... so i don't see why this would need config ... ?

flomerz: Sorry for confusion. I don't talk about the config.json. I talk about the self.config object.
Like here: https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/optimize/backtesting.py#L105

xmatthias: i am aware of this, but the "fee" key is only read - not written to the configuration.
In the default case, where you don't use `--fee 0.0011` to override the exchange default fee - `config['fee']` will be empty, and `self.fee` will have the right value.

https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/optimize/backtesting.py#L107

this `self.fee` object is then also assigned to each trade as `open_fee` and `close_fee`, and used/included in profit calculations.

So if you pass in a explicitly high fee, it'll be applied to all profit calculations by freqtrade - so by applying that again in your loss function, you're removing fees twice.

flomerz: Yes, but i don't use the profit calculation at all. Because it does not match to my use case. I set a relative high fee (containing fee + expected slippage) on each run with the cli argument and then i calculate the profit from scratch with the original time series. To consider compounding. So all i need for this is the btdata and the config accessible in the loss function. The fee is not the only config which I want to read in the loss function. I have some loss-function hyperparameters, which i would like to define in the config too. For now that are constants, but that is not ideal for different configurations.

xmatthias (approved):
Doesn't seem to impact performance after all

LGTM :+1: 

I've added the new parameter to the docs, otherwise noone except for you will be able to use it (without reading the code for "hidden" parameters)

flomerz:
Thanks for your work, I appreciate it :+1: 
Docs is a good idea :)

Have you noticed to additional config parameter? I'm not sure if the sample strategies works this way..

PR #4391, opened 2021-02-17
Add dust level to balance command in telegram

squat0001:

## Summary

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -243,7 +243,8 @@ CONF_SCHEMA = {
                     }
                 }
             },
-            'required': ['enabled', 'token', 'chat_id']
+            'required': ['enabled', 'token', 'chat_id'],
+            'balance_dust_level': 0.0001
         },
         'webhook': {
             'type': 'object',
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -54,6 +54,11 @@ DECIMALS_PER_COIN = {
     'ETH': 5,
 }
 
+DUST_PER_COIN = {
+    'BTC': 0.0001,
+    'ETH': 0.01
+}
+
 
 # Soure files with destination directories within user-directory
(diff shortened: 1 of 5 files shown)


freqtrade/constants.py
```diff
@@ -243,7 +243,8 @@
                     }
                 }
             },
-            'required': ['enabled', 'token', 'chat_id']
+            'required': ['enabled', 'token', 'chat_id'],
+            'balance_dust_level': 0.0001
```
xmatthias: that's ... not quite how you add a parameter to the validation dict...

```suggestion
            'balance_dust_level': {'type': 'number', minimum: 0.0},
```

The code as you have it should go to the full config example (and maybe to the other examples, but i'll leave that up to you).

xmatthias (changes requested):
In general, this will be a nice enhancement i think.

Maybe we should use some "sensible" defaults - as i think the most common stakes will be BTC, ETH, or one of the stable/fiat coins.
Maybe look at `DECIMALS_PER_COIN` in `constants.py` - where i recently added something similar for formatting - we only really need to define 2 keys, and all fiats can default to maybe 1 stake or so (while still keeping the possibility to override this via configuration).
that would lower the initial configuration effort by using some sensible defaults.

While the sensible defaults are optional - having this key in the documentation is a must - otherwise noone will know that it exists...

---

As before - i'd suggest to create a new branch in your repository for proposed changes and use it to make a PR against this repository.
commit history suggests that  did that, but didn't use it for this PR - otherwise you'll end up with messed up histories (again) ...

xmatthias (approved):
I think this will be a good improvement now.
By using "per coin" defaults for the 2 most common coins crypto-coins (and 1.0 for the remainder) - i hope the defaults will work for most.

LGTM :+1:

Issue #4385, opened 2021-02-16
Discover configuration/secrets via environment variables


patoarvizu:
## Describe the enhancement

I'd like to be able to inject configuration (specially secrets like tokens, credentials, and keys) via environment variables. This would make it easier to deploy freqtrade using declarative tools like Terraform, Ansible, Chef, etc., or on workload orchestration platforms like Kubernetes or Nomad, while avoiding storing sensitive secrets in source control.

Ideally, I 'd like to be able to do something along the lines of:
```
export FREQTRADE_CONFIG_EXCHANGE_SECRET=my-exchange-secret
export FREQTRADE_CONFIG_TELEGRAM_TOKEN=my-telegram-token
export FREQTRADE_CONFIG_API_SERVER_PASSWORD=my-api-server-password
freqtrade trade --config user_data/config.json ...
```
And both the values in `config.json` would be merged with the secrets injected via `FREQTRADE_CONFIG_*` environment variables. This way we can separate the workload invocation from the configuration, making it more flexible to deploy.

Other non-sensitive configuration values doesn't have to be injectable via this mechanism, but it may make sense to allow anything for consistency, possibly with some nomenclature that naturally maps to the configuration schema, e.g. something like:
```
"ask_strategy": {
  "ignore_buying_expired_candle_after": 300,
  "price_side": "bid"
}
```
Can be injected as:
```
FREQTRADE_CONFIG__ASK_STRATEGY__IGNORE_BUYING_EXPIRED_CANDLE_AFTER=300
FREQTRADE_CONFIG__ASK_STRATEGY__PRICE_SIDE=bid
```
In other words, use `FREQTRADE_CONFIG` as a prefix, and `__` as a separator to indicate hierarchical levels of the configuration.

I hope this explains the feature request!

xmatthias:
freqtrade supports multiple configuration files - so you can use `--config commonconfig.json --config sensitivevalues.json` which will then be merged, so you can avoid storing sensitive information in different, non-sourcecontrolled files.

You can use .gitignore for the sensitivevalues.json file (or some other name you like to use) to avoid storing these in source control.

because of that, i don't think that supporting environment variables makes sense, especially considering that our config logic is already quite complex (some values can come from either command line, configuration or strategy) - adding a 4th option to this will not make sense.

it'll also not improve security, as in both config or environment variable case, you'll have to store the variable in a file to allow the bot to start (the only difference is if it's a .env file, or a json file).
[...]

PR #4409, opened 2021-02-20
Implement environment variable substitution within configuration

krakowski:
## Summary
This PR adds the capability to use environment variables ~in the form of `${SECRET}`~ in the form of `${FREQTRADE_SECRET}` within the configuration where the prefix `FREQTRADE_` **must** be present. It does not alter the behavior of parsing configuration files without environment variables.

Solves issue #4385 

## Quick changelog

- Implement environment variable substitution within configuration file

## What's new?

  * Environment variables can now be specified within the configuration, which will be replaced accordingly. A good place to use this are exchange API keys or the Telegram token. This can especially be useful in environments where secrets are injected into environment variables (like Kubernetes etc.).

    ```
    "telegram" : {
        "token"   : "${FREQTRADE_TELEGRAM_TOKEN}",
        "chat_id" : "${FREQTRADE_TELEGRAM_CHAT}",
    }
    ```

  * A small easter egg. Look for a base64 encoded string in the contributed tests :stuck_out_tongue_winking_eye:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -11,7 +11,7 @@ Per default, the bot loads the configuration from the `config.json` file, locate
 
 You can specify a different configuration file used by the bot with the `-c/--config` command line option.
 
-In some advanced use cases, multiple configuration files can be specified and used by the bot or the bot can read its configuration parameters from the process standard input stream.
+In some advanced use cases, multiple configuration files can be specified and used by the bot or the bot can read its configuration parameters from the process standard input stream. Furthermore, the bot substitutes environment variables in the form of `${VARIABLE_NAME}`.
 
 If you used the [Quick start](installation.md/#quick-start) method for installing 
 the bot, the installation script should have already created the default configuration file (`config.json`) for you.
@@ -524,6 +524,9 @@ API Keys are usually only required for live trading (trading for real money, bot
 ```
 
 You should also make sure to read the [Exchanges](exchanges.md) section of the documentation to be aware of potential configuration details specific to your exchange.
+
+!!! Note
+    You can also store your secrets in environment variables and let the bot substitute them (`"key" : "${EXCHANGE_KEY}"`).
 
 ### Using proxy with Freqtrade
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -11,7 +11,7 @@ Per default, the bot loads the configuration from the `config.json` file, locate
 
 You can specify a different configuration file used by the bot with the `-c/--config` command line option.
 
-In some advanced use cases, multiple configuration files can be specified and used by the bot or the bot can read its configuration parameters from the process standard input stream. Furthermore, the bot substitutes environment variables in the form of `${VARIABLE_NAME}`.
+In some advanced use cases, multiple configuration files can be specified and used by the bot or the bot can read its configuration parameters from the process standard input stream. Furthermore, the bot substitutes environment variables in the form of `${FREQTRADE_VARIABLE_NAME}`. Please note that the prefix `FREQTRADE_` must be present for this to work.
 
 If you used the [Quick start](installation.md/#quick-start) method for installing 
 the bot, the installation script should have already created the default configuration file (`config.json`) for you.
@@ -526,7 +526,7 @@ API Keys are usually only required for live trading (trading for real money, bot
 You should also make sure to read the [Exchanges](exchanges.md) section of the documentation to be aware of potential configuration details specific to your exchange.
 
 !!! Note
-    You can also store your secrets in environment variables and let the bot substitute them (`"key" : "${EXCHANGE_KEY}"`).
+    You can also store your secrets in environment variables and let the bot substitute them (`"key" : "${FREQTRADE_EXCHANGE_KEY}"`).
 
 ### Using proxy with Freqtrade
 
--- a/freqtrade/configuration/load_config.py
+++ b/freqtrade/configuration/load_config.py
@@ -17,15 +17,17 @@ logger = logging.getLogger(__name__)
 
 CONFIG_PARSE_MODE = rapidjson.PM_COMMENTS | rapidjson.PM_TRAILING_COMMAS
 
+ENV_PREFIX = 'FREQTRADE_'
+
 
 class SubstitutionException(Exception):
     """
     Indicates that a variable within the configuration couldn't be substituted.
     """
 
-    def __init__(self, key: str, offset: int):
+    def __init__(self, err: str, offset: int):
         self.offset = offset
-        self.err = f'Environment variable {key} was requested for substitution, but is not set.'
+        self.err = err
         super().__init__(self.err)
 
 
@@ -52,8 +54,17 @@ def substitute_environment_variable(match: re.Match) -> str:
     Substitutes a matched environment variable with its value
     """
     key = match.group(1).strip()
+    if not key.startswith(ENV_PREFIX):
+        raise SubstitutionException(
+            f'Environment variable {key} must be prefixed with {ENV_PREFIX} .',
+            match.start(0)
+        )
+
     if key not in environ:
-        raise SubstitutionException(key, match.start(0))
+        raise SubstitutionException(
+            f'Environment variable {key} was requested for substitution, but is not set.',
+            match.start(0)
+        )
 
     return environ[key]
 
--- a/freqtrade/configuration/load_config.py
+++ b/freqtrade/configuration/load_config.py
@@ -4,6 +4,7 @@ This module contain functions to load the configuration file
 import logging
 import re
 import sys
+from os import environ
 from pathlib import Path
 from typing import Any, Dict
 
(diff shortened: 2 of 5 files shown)


xmatthias:
First of all, i'm not convinced on the benefit of using environment variables as you can use multiple `--config` options to archive the "secret hiding" in a seperate file.

--- 

that said - i don't think this will work in all cases - as we have some colliding, nested config keys (e.g. price side)
By giving control on this to the user - we'll be making configuration very non-standard (and honestly, difficult to understand - and even more easy to miss certain settings completely).
It'll also be a support nightmare, as everyone will use different variable names and then complain that "X is not working".

By allowing the user to control this, we also risk users breaking their environment (say i use `${PATH}` as user_dir - i'd prevent the bot from starting at all, as now my PATH points to something wrong).




If any, then a method as it's described in #4385 should be used - where freqtrade clearly defines the variables that have to be used via some pattern (i like the proposed key naming there too - it's used by other projects as well and using airflow for a while for work, it's quite easy to grasp how it should be named).

krakowski:
Hi @xmatthias,

thanks for taking time and looking at this!

> First of all, i'm not convinced on the benefit of using environment variables as you can use multiple `--config` options to archive the "secret hiding" in a seperate file.

I agree, security-wise there is no real benefit. The benefit you get with this method is that you can share your environment (`user_dir`) with others and it just works (given the other user has the environment variables set) without specifying any additional `--config` parameters (the default `config.json` contains the placeholders).

> that said - i don't think this will work in all cases - as we have some colliding, nested config keys (e.g. price side)

I don't really get what you mean by "colliding" and also don't see a problem here. The `price_side` field could be set like this for both JSON objects.

```json
"bid_strategy" : {
    "price_side" : "${BID_STRATEGY_SIDE}",
}

"ask_strategy" : {
    "price_side" : "${ASK_STRATEGY_SIDE}",
}
```

> By giving control on this to the user - we'll be making configuration very non-standard (and honestly, difficult to understand - and even more easy to miss certain settings completely).

The environment variable substitution is completely optional. The user does not have to use it.

> It'll also be a support nightmare, as everyone will use different variable names and then complain that "X is not working".

Sure, this might happen. This is why I check if the environment variable is set if it is used inside `config.json` and raise an exception showing the user what is wrong if the variable is not set.

  * https://github.com/freqtrade/freqtrade/pull/4409/files#diff-171d175bb51faf916375d7992d32df1ab7e84337631e6330a12b39a61e2b72ddR63-R67

> By allowing the user to control this, we also risk users breaking their environment (say i use `${PATH}` as user_dir - i'd prevent the bot from starting at all, as now my PATH points to something wrong).

Like I said before, this feature is totally optional. Someone using it should understand what environment variables are for. Nevertheless, the generated (or substituted) config runs through the same validation path like config files without environment variables, so errors like this should be catched there.

> If any, then a method as it's described in #4385 should be used - where freqtrade clearly defines the variables that have to be used via some pattern (i like the proposed key naming there too - it's used by other projects as well and using airflow for a while for work, it's quite easy to grasp how it should be named).

I could live with using a prefix like `FREQTRADE_` for environment variables. For this, I would only have to check if the key used inside the config starts with `FREQTRADE_` and raise an exception telling the user to use this prefix if it doesn't.

```python
if not key.startswith('FREQTRADE_'):
  raise SubstitutionException(f'Environment variable {key} must be prefixed with FREQTRADE_ .', match.start(0))
```

I just pushed these changes with an additional test. What do you think?

xmatthias:
> I agree, security-wise there is no real benefit. The benefit you get with this method is that you can share your environment (user_dir) with others and it just works (given the other user has the environment variables set) without specifying any additional --config parameters (the default config.json contains the placeholders).

i think the way you implement this you're voiding this benefit.

My environment variable might be called `$FREQTRADE_API_TOKEN` - while your's might be `$FREQTRADE_TELEGRAM_TOKEN`. 
While neither is technically incorrect - it'll not work as both tokens are different - so i'd be required to have YOUR environment variables set to work with your configuration - in which case the 2nd configuration will be more flexible.


> I don't really get what you mean by "colliding" and also don't see a problem here. The price_side field could be set like this for both JSON objects.

Yes, you'll do it this way (but you know what you're doing)- assume i'm a new user and define both variables as `$PRICE_SIDE` -now i have no way to control them separately - and i'll be left buzzled as to why (I as this user never worked with linux / env variables before).
it'll simply open too many possibilities for configuration errors (which from experience i can tell you, we MUST reduce to a minimum) - and even if we say "you must know what you're doing" - many think they do and then don't...

---

Maybe i wasn't clear above - my idea (and/or the idea in the issue) was to not make the variable-name part of the config - but simply:

if `$FREQTRADE__TELEGRAM__TOKEN`  - it'll be placed in the config as `telegram.token` key - with `FREQTRADE__` as prefix, and `__` (double-underline) as seperator between levels.
Obviously, for price-sides it'd be similar with `$FREQTRADE__BID_STRATEGY__PRICE_SIDE` being converted to `bid_strategy.price_side`.

That way, if the config (with these settings) works for you, it'll work for me without room for interpretation (obviously we should log which variables come from the environment to avoid problems)

that way, there's no possibility for different env variable namings - and it'll be clear (giving users too many options on such a trivial thing will only lead to confusion).

krakowski:
Hi,

thanks for your input.

> i think the way you implement this you're voiding this benefit.
> 
> My environment variable might be called `$FREQTRADE_API_TOKEN` - while your's might be `$FREQTRADE_TELEGRAM_TOKEN`.
> While neither is technically incorrect - it'll not work as both tokens are different - so i'd be required to have YOUR environment variables set to work with your configuration - in which case the 2nd configuration will be more flexible.

I don't think I've done a good job of clarifying my intended use case. I would like to share my environment with a friend, not the public.
 
> Yes, you'll do it this way (but you know what you're doing)- assume i'm a new user and define both variables as `$PRICE_SIDE` -now i have no way to control them separately - and i'll be left buzzled as to why (I as this user never worked with linux / env variables before).

I didn't expect this framework to be used by inexperienced users, since there are a lot of parts to it which require some amount of experience like python, pandas, Docker, API keys, etc. Aside from the fact that such errors could be catched in code by looking if the user specified a variable twice, in my personal opinion, people not knowing what they are doing can move into much worse areas than a wrong configuration.

> it'll simply open too many possibilities for configuration errors (which from experience i can tell you, we MUST reduce to a minimum) - and even if we say "you must know what you're doing" - many think they do and then don't...

As someone who works with freshman students I fully understand this. Sometimes I also wish for a world where it is no longer necessary to say "please read the docs". 

> Maybe i wasn't clear above - my idea (and/or the idea in the issue) was to not make the variable-name part of the config - but simply:
> 
> if `$FREQTRADE__TELEGRAM__TOKEN` - it'll be placed in the config as `telegram.token` key - with `FREQTRADE__` as prefix, and `__` (double-underline) as seperator between levels.
> Obviously, for price-sides it'd be similar with `$FREQTRADE__BID_STRATEGY__PRICE_SIDE` being converted to `bid_strategy.price_side`.
> 
> That way, if the config (with these settings) works for you, it'll work for me without room for interpretation (obviously we should log which variables come from the environment to avoid problems)
> 
> that way, there's no possibility for different env variable namings - and it'll be clear (giving users too many options on such a trivial thing will only lead to confusion).

While I agree that it is more portable to use the same naming scheme, I don't like the fact that a (hidden) environment variable is overriding some configuration value without me noticing it (except I look into my logs). In my opinion it's more explicit and traceable if the value or in this case a placeholder is entered in the configuration where everyone can see it.

Imagine an inexperienced user who at some moment in the past set some environment variable and forgot about it. The config would still contain the *expected* value while the environment variable overwrites it behind the scenes. This could even get worse if you think about `$FREQTRADE__STAKE_AMOUNT`.

There is also a problem I see when fields get renamed in updates. The code responsible for updating the config dict would need to check if the specified variable and all its parents are still valid configuration fields.

---

Since my initial solution does not seem to be suitable, I found an alternative to it, which is very close to your proposal.

My second idea would be to define an environment variable like `FREQTRADE_ USER_CONFIG` (please suggest a better name, if you have one in mind) which points to a config file (for example in `${HOME}/.freqtrade`) containing parts to override. If this variable is set this would lead to the same behaviour like `--config user_data/config.json --config "${FREQTRADE_ USER_CONFIG}"` This would not solve the problems I mentioned earlier (hidden overwrite, deprecated fields), but allows for a more convenient way of using the bot by setting the environment variable once instead of specifying `--config` twice at each invocation. I would also make sure to log this behavior very prominently and eventually send a message via Telegram to draw attention on it.

What do you think?

xmatthias:
> While I agree that it is more portable to use the same naming scheme, I don't like the fact that a (hidden) environment variable is overriding some configuration value without me noticing it (except I look into my logs). In my opinion it's more explicit and traceable if the value or in this case a placeholder is entered in the configuration where everyone can see it.

I don't think your arguments are consistent ... also using the approach in this PR is kind of obscure / hidden behavior - it's very easy to think you defined a variable - but in reality you're using a different config this time, in which case only the logs would tell me that the variable wasn't use (actually the logs would NOT tell me that it was used ...).

it's quite unlikely that you'll have a random environment variable around which accidentally starts with `FREQTRADE_` and matches a key that exists. 
This must be done intentionally. Now for sure, a variable can be left around from "earlier" - but the same can happen with a change to the configuration ... so it's not really a good argument i think.

following your own arguments - I don't see how `FREQTRADE__USER_CONFIG` (or whatever else) would make it better - (secretly) adding a 2nd config file is just as problematic as using an environment variable directly, as it's hidden, implicit behavior (which again, you'd only spot while observing your logs very carefully - again, following your own argumentation). 


> I didn't expect this framework to be used by inexperienced users, since there are a lot of parts to it which require some amount of experience like python, pandas, Docker, API keys, etc. 

you should observe the issues and questions that are on discord / slack sometimes ... 😆 
Now everyone starts out someplace - and freqtrade combines 2 fields which makes this a very difficult point 
one the technical side - where people understand how to start / run the bot, but not necessarily understand how markets work - and the trader side - which are often good / decent manual traders, trying to automate things (but often don't have a good technical understanding).

> I don't think I've done a good job of clarifying my intended use case. I would like to share my environment with a friend, not the public.

your particular use-case is one thing - but i need to see if it's beneficial for all users (or for the project) ... 
and in that case, when we consider sharing part of a configuration as potential use-case, then having a standardized way to define variables is better.

krakowski:
> I don't think your arguments are consistent ... also using the approach in this PR is kind of obscure / hidden behavior - it's very easy to think you defined a variable - but in reality you're using a different config this time, in which case only the logs would tell me that the variable wasn't use (actually the logs would NOT tell me that it was used ...).

Using a different config would require me to specify it explicitly using `--config`, which in itself implies that something changes and I should have a look at it.

> it's quite unlikely that you'll have a random environment variable around which accidentally starts with `FREQTRADE_` and matches a key that exists.
> This must be done intentionally. Now for sure, a variable can be left around from "earlier" - but the same can happen with a change to the configuration ... so it's not really a good argument i think.

Imagine someone shares their environment (with a strategy optimized for many trades in a short time) with me. Say the configuration contains the following.

```json
"stake_amount" : 10.00
```
Now I (as the inexperienced user of this environment) have `FREQTRADE__STAKE_AMOUNT` set to `1000.00` (for reasons unknown) and forgot about it. I start the bot and wonder why it bought for `1000.00`.

This behavior was not visible inside the configuration I got from the other person. I expect the configuration(s) to tell me what it will use, whether it's a constant like `"stake_amount": 10.00` or a variable like `"stake_amount": ${STAKE_AMOUNT}`. If I see a variable, it must be clear to me that it is set from the outside and I must make sure that it has the correct value.

> following your own arguments - I don't see how `FREQTRADE__USER_CONFIG` (or whatever else) would make it better - (secretly) adding a 2nd config file is just as problematic as using an environment variable directly, as it's hidden, implicit behavior (which again, you'd only spot while observing your logs very carefully - again, following your own argumentation).

I didn't claim that this would help any of the mentioned problems I see. In fact, I said that this will not solve the problems, but introduce a convenient way to start the bot with less parameters, if someone does not want to specify `--config` twice on each invocation. From my point of view there is no difference between your proposed method of having environment variables, which are not mentioned inside the config and therefore hidden, overwrite config values and an environment variable pointing to a file which override config values.

I just tried to plan ahead (implementation-wise) and use what is already there like the functionality to merge config files.
  
> your particular use-case is one thing - but i need to see if it's beneficial for all users (or for the project) ...
> and in that case, when we consider sharing part of a configuration as potential use-case, then having a standardized way to define variables is better.

I agree with this. Things I see as convenient may feel inconvenient for others.

My motivation behind this has been that I want to check in my `user_data` directory in a repository (for sharing) without exposing my API keys while also not being required to specify `--config user_data/config --config some_other_config.json` on each invocation but instead replace some placeholders inside my config. Since this doesn't seem to be a good option from the project's view, I will build something (maybe using jinja2) around it which doesn't require to alter the project's code.

Nevertheless, thanks for your time and thanks for this amazing project! :slightly_smiling_face:

xmatthias:
that's the problem with environment variables - if you think about "deploy and forget" - they might be convenient.

if you're running backtesting, then hyperopt, then dry-run, ... all from the same terminal, it can become confusing, as the old stuff tends to "stick around".

While you're right, changing the config will have you notice (notice is not control!) - you might still think that "i have setting XXX in my config as variable" (i'm sure you'll not know all configs you have by heart - so details may vary - at least if you use more than 1 strategy) - while in reality, that changed config doesn't contain that - which again, will be the "hidden" behaviour that's suddenly broken.

PR #4436, opened 2021-02-25
Allow changing the order_type for forcesell

### LINKED ISSUE #3199: proposal for /forcesell  and new /sellemergency
Operating system: ubuntu18
Python Version: python 3.6.9
CCXT version: ccxt==1.24.97
Branch: Develop
Last Commit ID: 030c487
  
/forcesell from telegram is not clear from documentaion if its using the config: order_types.sell": "limit",
or order_types."emergencysell": "market",
at the moment its using the order_types.sell": "limit",, i suggest to rename it maybe to /sell and add new /sellemergency 
so no need to add new config settings.

Th0masL:
## Summary
Allow specifying a different `order_type` for `forcesell`

closes: #3199

## Quick changelog

- Update `freqtrade/freqtradebot.py` to load a specific `order_type` for `forcesell` from the config file if present. If this value is missing, defaults to the same behavior as before (uses the `order_type` of `sell`)
- Update Docs to specify that this settings is optional

## What's new?
This will allow using a different `order_type` for `forcesell` than the one defined for `sell`

### DIFF
--- a/config_full.json.example
+++ b/config_full.json.example
@@ -49,6 +49,7 @@
         "buy": "limit",
         "sell": "limit",
         "emergencysell": "market",
+        "forcesell": "market",
         "stoploss": "market",
         "stoploss_on_exchange": false,
         "stoploss_on_exchange_interval": 60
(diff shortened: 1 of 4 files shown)


xmatthias:
sorry, misstyped
will have a look at this soon

xmatthias (approved):
I've added the key to the full config (for completeness)

Looks good! Thanks a lot :+1:

PR #4446, opened 2021-02-26
Webhook JSON format

Xanders:
Hey, thanks for a great project! 👋 Our team needs for a small improvement.

## Summary
Add ability to use JSON-Encoded webhooks (for example, for Mattermost integration).

## Quick changelog

- New `format` option was added to webhook config
- Documentation was added
- Unit test was added

## What's new?
Now users can integrate services which requires JSON-Encoded webhooks, not Form-Encoded. Default behavior wasn't changed.

### DIFF
--- a/docs/webhook-config.md
+++ b/docs/webhook-config.md
@@ -40,6 +40,19 @@ Sample configuration (tested using IFTTT).
 
 The url in `webhook.url` should point to the correct url for your webhook. If you're using [IFTTT](https://ifttt.com) (as shown in the sample above) please insert our event and key to the url.
 
+You can set the POST body format to Form-Encoded (default) or JSON-Encoded. Use `"format": "form"` or `"format": "json"` respectively. Example configuration for Mattermost Cloud integration:
+
+```json
+  "webhook": {
+        "enabled": true,
+        "url": "https://<YOURSUBDOMAIN>.cloud.mattermost.com/hooks/<YOURHOOK>",
+        "format": "json",
+        "webhookstatus": {
+            "text": "Status: {status}"
+        }
+    },
+```
+
 Different payloads can be configured for different events. Not all fields are necessary, but you should configure at least one of the dicts, otherwise the webhook will never be called.
 
 ### Webhookbuy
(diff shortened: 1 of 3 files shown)


docs/webhook-config.md
```diff
@@ -40,6 +40,19 @@ Sample configuration (tested using IFTTT).
...
+You can set the POST body format to Form-Encoded (default) or JSON-Encoded. Use `"format": "form"` or `"format": "json"` respectively. Example configuration for Mattermost Cloud integration:
+
+```json
+  "webhook": {
+        "enabled": true,
+        "url": "https://<YOURSUBDOMAIN>.cloud.mattermost.com/hooks/<YOURHOOK>",
+        "format": "json",
+        "webhookstatus": {
+            "text": "Status: {status}"
+        }
```
xmatthias, line 52: I guess it would be good to document what will then be the content of the post message - which in this case will be for example

```
{
   "text": "Status: Stopped"
}
```


Took me a while to figure that out from looking at the code

Xanders: Ok, I just added a new paragraph: https://github.com/Xanders/freqtrade/commit/984e70d4e8dd8dc8e2303ac8997143b382c7f9ed - is it good? Also there are instructions below like a `Parameters are filled using string.format`.

xmatthias: well for mattermost yes - but i'd want to keep this as generic as possible - using mattermost is okish - but assume you use a different (or custom) api - in which case you'd need to know what the exact request will be (which would be the above).

ps: line length is killing CI ...

Xanders: I just fixed line length and blank lines problems, sorry.

I'm not sure I understand what you mean. I used Mattermost as example same as IFTTT was used above in the doc. Of course user can use any API with any field keys. Interpolation rules described below in the doc. Please check this out, I described request better: https://github.com/Xanders/freqtrade/commit/f0391d3761a1aec4f1ad2e03236d8552a4f135d7 .

xmatthias: That's exactly what my intend was :+1:

xmatthias (approved):
LGTM :+1: 

Nice enhancement :100:

Xanders:
Thank you! 🎉

PR #4457, opened 2021-02-27
Fix: custom_stoploss returns typo

TheJoeSchr:
Afaik it should return -0.07 for 7% instead of -0.7.

As a side note, really interesting would also be an example for greater than 100% profits. especially when trailing stoploss, like
* Once profit is > 200% - stoploss will be set to 150%.

I assume it could be as simple as
```py
if current_profit > 2:
            return (-1.50 + current_profit)
````
to achieve it

But I'm not quite confident, if the bot can handle stuff smaller than `-1`, since `1` and `-1` seem to have some special meaning and are often used to disable stoploss etc.

## Summary

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

### DIFF
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -176,7 +176,7 @@ class AwesomeStrategy(IStrategy):
         if current_profit > 0.25:
             return (-0.15 + current_profit)
         if current_profit > 0.20:
-            return (-0.7 + current_profit)
+            return (-0.07 + current_profit)
         return 1
 ```
 

### CODE REVIEW

**Review by xmatthias — APPROVED** [review]
LGTM :+1:

**Comment by xmatthias** [comment]
The special meaning of -1 is limited to ROI - which can act as a always sell after X minutes.

In your sample, i don't see why that wouldn't work - as what matters is the result - which will be` -1.5 - 2.x`.

**Comment by TheJoeSchr** [comment]
> The special meaning of -1 is limited to ROI - which can act as a always sell after X minutes.

Thx

> In your sample, i don't see why that wouldn't work - as what matters is the result - which will be` -1.5 - 2.x`.

After sleeping on it and reading the source what happens with `custom_stoploss` result I realized I was very dumb.
First of all, my example wouldn't to at all what I intended it to do. I wanted to take a profit of at 150% profit and therefore set stoploss at open_rate*1.5. what my example did instead, was making an trailing offset of -150%, which doesn't make any sense.
Result of `custom_stoploss` is a negative relative offset and returning `(-1.5-2)=-3`.  not only would that not be a trailing stoploss of -50% but the line
```py
new_loss = float(current_rate * (1 - abs(result)))
```
would also remove the negative modifier, so the new stoploss would then be `current_rate * 2` which absolutely makes no sense.

a trailing stoploss as a negative offset nevers makes sense being bigger then `1`, because then all your money would be lost (`current_rate * 0`). yeah all the more reason for an additional example. if you want, I will make a second PR to add to `custom_stoploss` docu, with something like this?
> Once profit is > 200% - stoploss will be set to 150% absolute profit.
```py
if current_profit > 2:
           # calculate absolute profit of 1.5 as a negative offset from current_rate
            return (current_rate-(open_rate*1.5))/current_rate)-1
```

xmatthias:
a stoploss will never move lower than the initial stoploss (to prevent errors such as this).

PR #4473, opened 2021-03-02
fix: avg_stake_amount should not be `NaN` if df is empty

TheJoeSchr:
as discussed on discord.


xmatthias (approved):
LGTM 👍 

Is this the field causing the exception?

TheJoeSchr:
> LGTM 
> 
> Is this the field causing the exception?

Yes, afa i can tell with my limited knowledge of freqtradebot internas. But I'm wondering why dataframe would be empty in the first place... 

anyways no harm in having a little safe guard

PR #4478, opened 2021-03-03
Documentation: Add examples how to use dataframe with "custom_info"

TheJoeSchr:
## Summary
A lot of people ask how to use `custom_info` together with `dataframe`.

I invested a few days in trying to get this to work and though I would share the results as documentation examples.

## What's new?
I also added a few whitespace cleanups and renamed `cust_info` from the original storage example, because I find it easier to comprehend this way?

### DIFF
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -55,8 +55,9 @@ class AwesomeStrategy(IStrategy):
     def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
         # using "ATR" here as example
         dataframe['atr'] = ta.ATR(dataframe)
-        # add indicator mapped to correct DatetimeIndex to custom_info
-        self.custom_info[metadata['pair']] = dataframe[['date', 'atr']].copy().set_index('date')
+        if self.dp.runmode.value in ('backtest', 'hyperopt'):
+          # add indicator mapped to correct DatetimeIndex to custom_info
+          self.custom_info[metadata['pair']] = dataframe[['date', 'atr']].copy().set_index('date')
         return dataframe
 ```
 
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -60,6 +60,12 @@ class AwesomeStrategy(IStrategy):
         return dataframe
 ```
 
+!!! Warning
+    The data is not persisted after a bot-restart (or config-reload). Also, the amount of data should be kept smallish (no DataFrames and such), otherwise the bot will start to consume a lot of memory and eventually run out of memory and crash.
+
+!!! Note
+    If the data is pair-specific, make sure to use pair as one of the keys in the dictionary.
+
 See `custom_stoploss` examples below on how to access the saved dataframe columns
 
 ## Custom stoploss
@@ -236,6 +242,11 @@ Imagine you want to use `custom_stoploss()` to use a trailing indicator like e.g
 
 See: "Storing custom information using DatetimeIndex from `dataframe`" example above) on how to store the indicator into `custom_info`
 
+!!! Warning
+    only use .iat[-1] in live mode, not in backtesting/hyperopt
+    otherwise you will look into the future
+    see: https://www.freqtrade.io/en/latest/strategy-customization/#common-mistakes-when-developing-strategies
+
 ``` python
 from freqtrade.persistence import Trade
 from freqtrade.state import RunMode
@@ -260,7 +271,10 @@ class AwesomeStrategy(IStrategy):
               # but we can just use the last entry from an already analyzed dataframe instead
               dataframe, last_updated = self.dp.get_analyzed_dataframe(pair=pair,
                                                                        timeframe=self.timeframe)
-              # but we can just use the last entry to get the current value
+              # WARNING
+              # only use .iat[-1] in live mode, not in backtesting/hyperopt
+              # otherwise you will look into the future
+              # see: https://www.freqtrade.io/en/latest/strategy-customization/#common-mistakes-when-developing-strategies
               relative_sl = dataframe['atr'].iat[-1]
 
             if (relative_sl is not None):
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -211,13 +211,11 @@ class AwesomeStrategy(IStrategy):
               relative_sl =  self.custom_info[pair]['atr].iloc[ -1 ]
 
             if (relative_sl is not None):
-                print("Custom SL: {}".format(relative_sl))
                 # new stoploss relative to current_rate
                 new_stoploss = (current_rate-relative_sl)/current_rate
                 # turn into relative negative offset required by `custom_stoploss` return implementation
                 result = new_stoploss - 1
 
-        print("Result: {}".format(result))
         return result
 ```
 
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -142,7 +142,7 @@ class AwesomeStrategy(IStrategy):
             return -1 # return a value bigger than the inital stoploss to keep using the inital stoploss
 
         # After reaching the desired offset, allow the stoploss to trail by half the profit
-        desired_stoploss = current_profit / 2 
+        desired_stoploss = current_profit / 2
 
         # Use a minimum of 2.5% and a maximum of 5%
         return max(min(desired_stoploss, 0.05), 0.025)
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -234,8 +234,7 @@ class AwesomeStrategy(IStrategy):
 
 Imagine you want to use `custom_stoploss()` to use a trailing indicator like e.g. "ATR"
 
-See: (Storing custom information using DatetimeIndex from `dataframe`
-)[WIP] on how to store the indicator into `custom_info`
+See: "Storing custom information using DatetimeIndex from `dataframe`" example above) on how to store the indicator into `custom_info`
 
 ``` python
 from freqtrade.persistence import Trade
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -179,6 +179,35 @@ class AwesomeStrategy(IStrategy):
             return (-0.07 + current_profit)
         return 1
 ```
+#### Custom stoploss using an indicator from dataframe example
+
+Imagine you want to use `custom_stoploss()` to use a trailing indicator like e.g. "ATR"
+
+See: (Storing custom information using DatetimeIndex from `dataframe`
+)[WIP] on how to store the indicator into `custom_info`
+
+``` python
+from freqtrade.persistence import Trade
+
+class AwesomeStrategy(IStrategy):
+
+    # ... populate_* methods
+
+    use_custom_stoploss = True
+
+    def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
+                        current_rate: float, current_profit: float, **kwargs) -> float:
+
+        result = 1
+        if self.custom_info[pair] is not None and trade is not None:
+            atr = self.custom_info[pair].loc[current_time]['atr']
+            if (atr is not None):
+                # new stoploss relative to current_rate
+                new_stoploss = (current_rate-atr)/current_rate
+                # turn into relative negative offset required by `custom_stoploss` return implementation
+                result = new_stoploss - 1
+        return result
+```
 
 ---
 
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -188,6 +188,7 @@ See: (Storing custom information using DatetimeIndex from `dataframe`
 
 ``` python
 from freqtrade.persistence import Trade
+from freqtrade.state import RunMode
 
 class AwesomeStrategy(IStrategy):
 
@@ -200,12 +201,23 @@ class AwesomeStrategy(IStrategy):
 
         result = 1
         if self.custom_info[pair] is not None and trade is not None:
-            atr = self.custom_info[pair].loc[current_time]['atr']
-            if (atr is not None):
+            # using current_time directly (like below) will only work in backtesting.
+            # so check "runmode" to make sure that it's only used in backtesting
+            if(self.dp.runmode == RunMode.BACKTEST):
+              relative_sl = self.custom_info[pair].loc[current_time]['atr]
+            # in live / dry-run, it'll be really the current time
+            else:
+              # but we can just use the last entry to get the current value
+              relative_sl =  self.custom_info[pair]['atr].iloc[ -1 ]
+
+            if (relative_sl is not None):
+                print("Custom SL: {}".format(relative_sl))
                 # new stoploss relative to current_rate
-                new_stoploss = (current_rate-atr)/current_rate
+                new_stoploss = (current_rate-relative_sl)/current_rate
                 # turn into relative negative offset required by `custom_stoploss` return implementation
                 result = new_stoploss - 1
+
+        print("Result: {}".format(result))
         return result
 ```
 
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -257,8 +257,11 @@ class AwesomeStrategy(IStrategy):
               relative_sl = self.custom_info[pair].loc[current_time]['atr]
             # in live / dry-run, it'll be really the current time
             else:
+              # but we can just use the last entry from an already analyzed dataframe instead
+              dataframe, last_updated = self.dp.get_analyzed_dataframe(pair=pair,
+                                                                       timeframe=self.timeframe)
               # but we can just use the last entry to get the current value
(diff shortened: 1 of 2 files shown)


docs/strategy-advanced.md
```diff
@@ -179,6 +179,35 @@ class AwesomeStrategy(IStrategy):
...
+    # ... populate_* methods
+
+    use_custom_stoploss = True
+
+    def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
+                        current_rate: float, current_profit: float, **kwargs) -> float:
+
+        result = 1
+        if self.custom_info[pair] is not None and trade is not None:
+            atr = self.custom_info[pair].loc[current_time]['atr']
```
xmatthias: using `current_time` directly (like this) will only work in backtesting.

in live / dry-run, it'll be really the current time (like 14:54:44) - not the candle time - so this lookup will fail.

TheJoeSchr: Ah, good I made a PR then before going live, sometimes karma works out :D

Any idea for a workaround?

Only thing I can think right now, would be getting `index` of `.iloc[trade.open_date]` and then walking forward via index until `date >= current_time`. Not sure if this would work in backtesting without looking in into the future (because I would have to use numerical index like this `dataframe[index_starting_at_opendate])

TheJoeSchr: Hmm, maybe `.loc[open_date < date < current_time].last()` (pseudocode) would be eleganter. 
Will investigate over the next days if you don't have a better idea into what else I could look into

xmatthias: just change how you assign `atr` based on the run-mode.
for live, you can as well use .last (or `.iloc[-1]`) - but you MUST make sure that's not used in backtesting - therefore it's not recommended.

TheJoeSchr:
@xmatthias  I added your requested changes via a new commit, not sure if you get a notification for that automatically.

I tested it `freqtrade backtesting` and `freqtrade trade --dry-run` added an example file to `freqtrade/strategies`. Please still carefully look over it, if you can spot any other hidden faults with this. Thanks.

docs/strategy-advanced.md
```diff
@@ -179,6 +179,45 @@ class AwesomeStrategy(IStrategy):
...
+    use_custom_stoploss = True
+
+    def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
+                        current_rate: float, current_profit: float, **kwargs) -> float:
+
+        result = 1
+        if self.custom_info[pair] is not None and trade is not None:
+            # using current_time directly (like below) will only work in backtesting.
+            # so check "runmode" to make sure that it's only used in backtesting
+            if(self.dp.runmode == RunMode.BACKTEST):
```
xmatthias: you will want to include Hyperopt as well here ... otherwise you'll break hyperopt / backtest alignment for sure

docs/strategy-customization.md
```diff
@@ -333,6 +333,22 @@ class AwesomeStrategy(IStrategy):
...
+    # Create custom dictionary
+    custom_info = {}
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        # add indicator mapped to correct DatetimeIndex to custom_info
+        # using "ATR" here as example
+        self.custom_info[metadata['pair']] = dataframe[['date', 'atr']].copy().set_index('date')
+        return dataframe
+
+See: (custom_stoploss example)[WIP] for how to access it
```
xmatthias: we will want to fix that link for sure ... 


i also think we maybe should add a warning here - that this will increas memory consumption (especially relevant for backtesting / hyperopt).

docs/strategy-customization.md
```diff
@@ -333,6 +333,22 @@ class AwesomeStrategy(IStrategy):
...
+Imagine you need to store an indicator like `ATR` or `RSI` into `custom_info`. To use this in a meaningful way, you will not only need the raw data of the indicator, but probably also need to keep the right timestamps.
+
+class AwesomeStrategy(IStrategy):
+    # Create custom dictionary
+    custom_info = {}
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+        # add indicator mapped to correct DatetimeIndex to custom_info
+        # using "ATR" here as example
+        self.custom_info[metadata['pair']] = dataframe[['date', 'atr']].copy().set_index('date')
```
xmatthias: i think the sample should calculate the atr column here.

from experience i know that people tend to copy samples 1:1 and then wonder why they're not running if small things are missing.

docs/strategy-advanced.md
```diff
@@ -179,6 +179,45 @@ class AwesomeStrategy(IStrategy):
...
+              relative_sl = self.custom_info[pair].loc[current_time]['atr]
+            # in live / dry-run, it'll be really the current time
+            else:
+              # but we can just use the last entry to get the current value
+              relative_sl =  self.custom_info[pair]['atr].iloc[ -1 ]
```
xmatthias: for live, storing the dataframe is not really necessary, it's available from [`get_analyzed_dataframe()`](https://www.freqtrade.io/en/latest/strategy-customization/#get_analyzed_dataframepair-timeframe)

the problem is that this is not available during backtesting, which requires the "cust_info" workaround - which i'm ... hesitant to fully document fully - as it will raise memory requirements / problems - as the whole dataframe (or at least 2 columns of it) are stored and kept in memory.

docs/strategy-customization.md
```diff
@@ -311,18 +311,18 @@ The name of the variable can be chosen at will, but should be prefixed with `cus
 ```python
     # Create custom dictionary
-    cust_info = {}
+    custom_info = {}
```
xmatthias: i guess it might be a good idea to move the whole "storing information" section to the advanced page?

i don't think you'll want to (or be able to) use this if you're just starting out ... so i'd consider it an advanced topic (we didn't have that section when this was documented i think) ?

docs/strategy-customization.md
TheJoeSchr: >i also think we maybe should add a warning here - that this will increas memory consumption (especially relevant for backtesting / hyperopt)

is it really that much more memory for a few more columns? maybe removing `.copy()` and just referencing it, would solve that.but it's been a while since I worked with python references and also I'm a new with dataframe, so I opted for the safer option by using `copy()` here. do you think we can remove it. How can I test it for edge cases.

docs/strategy-advanced.md
TheJoeSchr: > as the whole dataframe (or at least 2 columns of it) are stored and kept in memory

for my usecase, spending a few megabytes or even gigabytes of system memory in backtesting, is a great tradeoff to designing a strategy which let's you ride a uptrend at zero risk (e.g. `putting sl @ (atr) if profit+atr>1.5`) or even having some kind of fixed risk/reward money management (e.g. `intial_risk = open_rate+df[open_date]['atr']; take_profit = initial_risk*1.5+open_rate; if(current_rate>take_proift) #use trailng sl`) so I can compare W/L rate between strategies and still profit from trending.

I guess, what I want to say: there should be a note that it will increase memory consumption, for sure, yes, let's do that. make it even big and bold.

but the decision should still be made by the user. intentionally "hiding" it by not documenting it or giving a best practice example, only make people try to do it themself anyway. probably ignorante of best practices and the result will be something like my inital PR, that wouldn't even have worked in live.(which to me is a way worse result than using more memory)

backtesting on a device with limited memory like a raspi may be a bad idea anyways....

docs/strategy-customization.md
TheJoeSchr: done

docs/strategy-customization.md
TheJoeSchr: > we will want to fix that link for sure

done.

docs/strategy-customization.md
TheJoeSchr: done.

JoeSchr:
@xmatthias  I added most of your requested changes.

What I couldn't add was using `get_analyzed_dataframe()`.
I got this error:
```shell
  File "/home/joe/projects/freqtrade/freqtrade/commands/trade_commands.py", line 19, in start_trading
    worker.run()
  File "/home/joe/projects/freqtrade/freqtrade/worker.py", line 74, in run
    state = self._worker(old_state=state)
  File "/home/joe/projects/freqtrade/freqtrade/worker.py", line 111, in _worker
    self._throttle(func=self._process_running, throttle_secs=self._throttle_secs)
  File "/home/joe/projects/freqtrade/freqtrade/worker.py", line 132, in _throttle
    result = func(*args, **kwargs)
  File "/home/joe/projects/freqtrade/freqtrade/worker.py", line 145, in _process_running
    self.freqtrade.process()
  File "/home/joe/projects/freqtrade/freqtrade/freqtradebot.py", line 184, in process
    self.exit_positions(trades)
  File "/home/joe/projects/freqtrade/freqtrade/freqtradebot.py", line 693, in exit_positions
    if trade.open_order_id is None and trade.is_open and self.handle_trade(trade):
  File "/home/joe/projects/freqtrade/freqtrade/freqtradebot.py", line 799, in handle_trade
    if self._check_and_execute_sell(trade, sell_rate, buy, sell):
  File "/home/joe/projects/freqtrade/freqtrade/freqtradebot.py", line 934, in _check_and_execute_sell
    should_sell = self.strategy.should_sell(
  File "/home/joe/projects/freqtrade/freqtrade/strategy/interface.py", line 518, in should_sell
    stoplossflag = self.stop_loss_reached(current_rate=current_rate, trade=trade,
  File "/home/joe/projects/freqtrade/freqtrade/strategy/interface.py", line 579, in stop_loss_reached
    stop_loss_value = strategy_safe_wrapper(self.custom_stoploss, default_retval=None
  File "/home/joe/projects/freqtrade/freqtrade/strategy/strategy_wrapper.py", line 33, in wrapper
    raise StrategyError(str(error)) from error
freqtrade.exceptions.StrategyError: datetime.datetime(2021, 3, 4, 14, 6, 28, 715919, tzinfo=datetime.timezone.utc)
```
when running it with `freqtrade trade --dry-run -s MyAwesomeStrategy`. I updated my other PR acordinlgy, so you could test it on your mchine if needed: https://github.com/freqtrade/freqtrade-strategies/pull/130 .

docs/strategy-customization.md
xmatthias: well you're thinking about live - there, the memory consumption can be neglectible in most cases.

However in backtesting you might be talking about 2 columns ...
but what if that's 2 columns for 50 pairs for 500.000 lines (that's pretty easily doable on 1m timeframes).

and in hyperopt, you multiply the above result with the number of cores ... now it's no longer a "small amount of memory" - but can become a breaking factor for using hyperopt on a certain machine or not.

docs/strategy-advanced.md
xmatthias: > backtesting on a device with limited memory like a raspi may be a bad idea anyways....

agree on that - but not every user lives in a 1st world country - where buying a new pc every other year is possible - i've seen people use freqtrade on a 10yo pc without problem (ok, hyperopt may not run - but backtesting works just fine).

I think you misread my "hesitant" 
if i were against it, i'd not have commented on multiple lines but simply said "no" - otherwise i'd waste time for both of us.

The hesitant was more ... i knew about this possibility for a long time - but didn't document it intentionally ;) maybe should've used better wording...

TheJoeSchr: Ok, then let's do it!

Should I just add a comment warning with the same wording as the warning for not using too many indicators in `prepare_indicators()`?
Because it's basically like adding extra columns there, isn't it?

Or is it somehow more than that?

And regarding first world country or not, I think being hesitant adding more memory consumption is good, from a maintainer/Dev perspective, but like you said, most user are not really technical enough to appreciate the finer point of it. Post a warning, if they can't run it, they will have to triage their indicators/etc. 

Backtesting is a tool and if somebody wants to use it, because they think they can increase their alpha, the docs should at least point in the right direction, so they don't shoot themselves in the foot. 
If the tool then doesn't work with their hardware spec, they will either have to buy more, rethink their strategy or priorities what they want to do.

It's the same in all other areas. Nothing is stopping me to add thousands of indicators to use in `buy_trend()` except a note somewhere. But it's well documented how to do it if I want to. So I don't really get why `custom_stoploss` should be a different kind of animal (except it's newer).

But no need to decided this for the users beforehand.

 I'm not pounding on you with the same argument over and over, just trying to ease your mind about this. Because it seems a common issue in the Algo Trading space, that things that are bad from a technical view are still totally fine for the majority of user, which care most about getting their edge/alpha.

xmatthias: > Or is it somehow more than that?

yes and no ... i'd have to look at it in detail - without `.copy()` it's for sure bearing the risk of causing a memory leak (keeping a reference around which shouldn't be).

Just put it as `!!! Warning` box ... it'll be fine i guess. I even think we had something similar for the cust_info before ... so 

> I'm not pounding on you with the same argument over and over

You're clear that we're now talking about details ... so the fact THAT we'll document this has been decided when you posted this PR ... 😆

TheJoeSchr: Yeah, still you are far more experienced with freqtrades internal. Maybe you can spot a flaw in my logic if I lay it out.

At least it got me thinking about it in more detail. Also I made a note to look further into pythons memory management regarding references.

I don't think, with what I know from other languages, that to copy a reference will hurt or make a memory leak. I was more afraid of side effects which get spaghetti code-y very fast.

There is probably a reference counter for each object and it will be deleted when not needed anymore. even if in this case RAII de-init is shortcuted. Also in which case would there be a IStrategy Model around still holding the df column in custom_info and no "global" dataframe?

TheJoeSchr:
added the requested warnings and use  `dataframe...iat[]` to get last element

docs/strategy-customization.md
```diff
@@ -302,37 +302,6 @@ Currently this is `pair`, which can be accessed using `metadata['pair']` - and w
 The Metadata-dict should not be modified and does not persist information across multiple calls.
 Instead, have a look at the section [Storing information](#Storing-information)
```
xmatthias:
```suggestion
Instead, have a look at the section [Storing information](strategy-advanced.md#Storing-information)
```

docs/strategy-advanced.md
```diff
@@ -179,6 +236,55 @@ class AwesomeStrategy(IStrategy):
...
+#### Custom stoploss using an indicator from dataframe example
+
+Imagine you want to use `custom_stoploss()` to use a trailing indicator like e.g. "ATR"
+
+See: "Storing custom information using DatetimeIndex from `dataframe`" example above) on how to store the indicator into `custom_info`
+
+!!! Warning
+    only use .iat[-1] in live mode, not in backtesting/hyperopt
+    otherwise you will look into the future
+    see: https://www.freqtrade.io/en/latest/strategy-customization/#common-mistakes-when-developing-strategies
```
```suggestion
    see [Common mistakes when developing strategies](strategy-customization.md#common-mistakes-when-developing-strategies) for more info.
```

we'll want to use relative links - so it'll also work in the future, when looking at a "fixed" version of the docs.

xmatthias (approved):
just 2 minor things about the links left which would become broken with this - rest looks good :+1:

TheJoeSchr:
let me just also add the line which only populates `custom_info` for backtesting/hyperopts

TheJoeSchr:
ok, done. I needed to force push, because applying your "2 minor things" commits first then didn't let my push into my own branch anymore. 

FYI I added 2 more commits since your last review:
> - fix(docs/strategy-advanced/custom_stoploss/example): check if "pair" … …
900deb6
> - fix(docs/strategy-advanced/custom_info-storage/example): only add to … …
1304918

I also think it looks good now! Thanks for shepherding it in! :D

docs/strategy-customization.md
rokups: @xmatthias do we really need to make a copy? Can not we keep a dataframe reference?

xmatthias: well - it's just a documentation sample ... so it's up to the user how to do it in the end.

I'm ... not 100% sure on the impact - and in theory it can cause memory leaks - but if it really does (and under which circumstances) would need investigation. Unfortunately, it can also only appear after months of running (very slow memory leak) ... so it's rather dangerous to just tell users to store references to the dataframe "wherever" ... 

Honestly, while i think i understand memory management in python well enough - i do mistrust pandas in this regards, as it can EITHER return a copy or a view for simple operations (depending on some - to me - unclear rules) - so i think about every dataframe as a potential source of memory leaks.

rokups: Given all those concerns, would it not make sense to provide a way to access internally stired dataframe instead of having user to store a reference or a copy or who knows what, leaving so much room for mistakes?

xmatthias: during regular runs (dry / live) - the analyzed dataframe IS stored and available with `dp.get_analyzed_dataframe()`.

this is however problematic for backtesting / hyperopt where all "custom" columns are removed after populating buy / sell columns (for performance reasons) - so they would not be available, and 90% of the users will not need these.

PR #4488, opened 2021-03-05
Add strategy name to HyperOpt results filename

patrick310:
## Summary
This extends the HyperOpt result filename by adding the strategy name. 

Solve the issue: #___

## Quick changelog

- Added filename to HyperOpt filename in HyperOpt Class Init.

## What's new?
This allows analysis of HyperOpt results folder with no additional necessary context. An alternative idea would be to expand the result dict, but the additional static copies are non value added.


patrick310:
Maybe nevermind... I guess it made the line too long, so I will need to reformat. Not sure if it's the length of the filename or length of the command that is the issue.

xmatthias (approved):
LGTM :+1:

Issue #4498, opened 2021-03-08
The output of the command /trades is sorted by ID instead of by close_date


Th0masL:
## Describe your environment

  * Freqtrade Version: 2021.2

## Describe the problem:

I believe that the command `/trades [limit]` should list the last closed trades, ordered by closure time.

But right now the trades are listed by Trade ID, therefore it's not showing the last closed trades but it is showing the last opened trades that have been closed.

If you are looking for the last closed trade, you might have to go back as far as `/trade XX` (like `/trade 30`), depending when it was opened initially.

### Steps to reproduce:

  1. Look for a trade that got opened long time ago and that just got closed
  2. Run the command `/trades`
  3. Search for this trade in the list
  4. If it has been opened long enough ago, it won't be visible
  
### Observed Results:

  * What happened?
  The trades are ordered by Trade ID, so the recent closed trades might be missing, depending on their ID
  * What did you expect to happen?
  The trades should be ordered by Closure Time

### Relevant code exceptions or logs

I believe that in the `rpc/rpc.py` file below, `Trade.id.desc()` should be changed to `Trade.close_date.desc()`
  ```
    def _rpc_trade_history(self, limit: int) -> Dict:
        """ Returns the X last trades """
        if limit > 0:
            trades = Trade.get_trades([Trade.is_open.is_(False)]).order_by(
                Trade.id.desc()).limit(limit)
        else:
            trades = Trade.get_trades([Trade.is_open.is_(False)]).order_by(Trade.id.desc()).all()
  ```

It could also be worth changing the columns in the output of `/trades` :
- add the Trade ID
- replace "Open Date" by "Close Date", or add "Close Date" (although there isn't that much space)

From :
```
10 recent trades:
Open Date     Pair        Profit (USDT)
------------  ----------  ------------------
6 hours ago   SFP/USDT    0.25% (0.16807254)
7 hours ago   ZIL/USDT    0.76% (0.50417386)
8 hours ago   CVC/USDT    0.01% (0.00792157)
9 hours ago   DOCK/USDT   0.81% (0.53431329)
```

To :

```
10 recent trades:
  ID  Close Date     Pair        Profit (USDT)
----  -------------  ----------  ------------------
  17  0.1 hours ago  CVC/USDT    0.01% (0.00792157)
  26  0.3 hours ago  SFP/USDT    0.25% (0.16807254)
  15  1.0 hours ago  DOCK/USDT   0.81% (0.53431329)
[...]

PR #4508, opened 2021-03-08
Change order_by from ID to close_date for /trades command

Th0masL:
## Summary
Change the order_by of the output for the `/trades` command, from `id` to `close_date`.

Solve the issue: #4498

## Quick changelog

- change the order_by for `/trades` command

## What's new?

Existing output:

```
2 recent trades:
Open Date     Pair     Profit (BTC)
------------  -------  -------------
14 hours ago  XRP/BTC  1.00% (None)
18 hours ago  ETC/BTC  0.50% (None)
```

Example of new output (output done manually, the final order might require to be inverted) :

```
2 recent trades:
Close Date   Pair (ID)     Profit (BTC)
-----------  ------------  ------------
an hour ago  ETC/BTC (#2)  0.50% (None)
2 hours ago  XRP/BTC (#3)  1.00% (None)
```

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -289,9 +289,10 @@ class RPC:
         """ Returns the X last trades """
         if limit > 0:
             trades = Trade.get_trades([Trade.is_open.is_(False)]).order_by(
-                Trade.id.desc()).limit(limit)
+                Trade.close_date.desc()).limit(limit)
         else:
-            trades = Trade.get_trades([Trade.is_open.is_(False)]).order_by(Trade.id.desc()).all()
+            trades = Trade.get_trades([Trade.is_open.is_(False)]).order_by(
+                Trade.close_date.desc()).all()
 
         output = [trade.to_json() for trade in trades]
 
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -637,13 +637,13 @@ class Telegram(RPCHandler):
                 nrecent
             )
             trades_tab = tabulate(
-                [[arrow.get(trade['open_date']).humanize(),
-                  trade['pair'],
+                [[arrow.get(trade['close_date']).humanize(),
+                  trade['pair'] + " (#" + str(trade['trade_id']) + ")",
                   f"{(100 * trade['close_profit']):.2f}% ({trade['close_profit_abs']})"]
                  for trade in trades['trades']],
                 headers=[
-                    'Open Date',
-                    'Pair',
+                    'Close Date',
+                    'Pair (ID)',
                     f'Profit ({stake_cur})',
                 ],
                 tablefmt='simple')
--- a/tests/conftest_trades.py
+++ b/tests/conftest_trades.py
@@ -88,7 +88,7 @@ def mock_trade_2(fee):
         timeframe=5,
         sell_reason='sell_signal',
         open_date=datetime.now(tz=timezone.utc) - timedelta(minutes=20),
-        close_date=datetime.now(tz=timezone.utc),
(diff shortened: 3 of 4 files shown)


Th0masL:
Strange that the build worked on some platforms, but failed on others.

Also I'm not exactly sure if we should reverse the order of the order_by, but I think we might have to, I don't know how to do it.

xmatthias:
> Also I'm not exactly sure if we should reverse the order of the order_by, but I think we might have to, I don't know how to do it.

i'm not quite sure i understand what you mean?
the order is currently (with this branch) `order by close_date desc`, which seems correct to me.
what would you want to change this to?

Th0masL:
> the order is currently (with this branch) `order by close_date desc`, which seems correct to me.

Indeed, I totally missed the fact that the existing piece of code was using `.desc()` ! So we're all good on this side then, my bad 😂

Th0masL:
Is there anything I can do to fix the failing builds ? Can you retry them maybe ?

Or maybe this need to be fixed, but I'm not sure why it's happening only on the windows builds

```
    def test_rpc_trade_history(mocker, default_conf, markets, fee):
        mocker.patch.multiple(
            markets=PropertyMock(return_value=markets)
        )
    
        freqtradebot = get_patched_freqtradebot(mocker, default_conf)
        create_mock_trades(fee)
        rpc = RPC(freqtradebot)
        rpc._fiat_converter = CryptoToFiatConverter()
        trades = rpc._rpc_trade_history(2)
        assert len(trades['trades']) == 2
        assert trades['trades_count'] == 2
        assert isinstance(trades['trades'][0], dict)
        assert isinstance(trades['trades'][1], dict)
    
        trades = rpc._rpc_trade_history(0)
        assert len(trades['trades']) == 2
        assert trades['trades_count'] == 2
        # The first closed trade is for ETC ... sorting is descending
>       assert trades['trades'][-1]['pair'] == 'ETC/BTC'
E       AssertionError: assert 'XRP/BTC' == 'ETC/BTC'
E         - ETC/BTC
E         + XRP/BTC

tests\rpc\test_rpc.py:298: AssertionError
```

It looks like the problem might be coming from the sorting of the list on windows, as it looks like we don't have this sorting problem on Linux.

```
(.env) # python3.7 .env/bin/py.test tests/rpc/test_rpc.py 
========== test session starts ==========
platform linux -- Python 3.7.5, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
Test order randomisation NOT enabled. Enable with --random-order or --random-order-bucket=<bucket_type>
rootdir: /home/thomas/github/Th0masL/freqtrade
plugins: asyncio-0.14.0, random-order-1.0.4, cov-2.11.1, mock-3.5.1
collected 24 items

tests/rpc/test_rpc.py ........................                        [100%]

========== 24 passed, 2 warnings in 0.96s ==========
```

xmatthias:
I don't think it's related to windows (or python 3.7) - but it's a question of the speed of the box (but there's other things to consider as well, like the randomization that happens during CI to avoid test-dependency (and can lead to random failures from time to time).

In this particular case ... both closed mock-trades close with `close_date=datetime.now(tz=timezone.utc),` (`conftest_trades.py`) - which makes it a timing + luck problem, where it'll depend on whichever trade is closed out first.

I've pushed a commit now that aims to fix that - we'll see in a few minutes if it really does.

xmatthias:
Unless you think there's still something missing, i'd say this looks good now? (don't wanna force it in "against your will" as it's still set as `[WIP]` ... )

Th0masL:
Yep, all good, I've removed the WIP from the title, feel free to merge

xmatthias (approved):
LGTM :+1: 

Thanks a lot, i think this will be useful!

PR #4510, opened 2021-03-08
multi_arch_docker: Added multi-arch-builder

aviadcye:

## Summary

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -1,0 +1,59 @@
+name: Docker Image CI
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


xmatthias:
Unfortunateliy, doing multi-arch builds is not possible at the moment, since the build on 3.9 (or 3.8) on a raspberry would take too long and ci would time out.

Therefore, building for arm is only possible with 32 bits, and only wiht python3.7 (which is the latest that  piwheels.org currently supports).

PR #4517, opened 2021-03-10
Fix(docs/data-download): use wording specific for `data-download`

TheJoeSchr:
Just a small lapse I noticed while reading the docs

`Show profits for only these pairs` -> `Only download data for these pairs[...]`

### DIFF
--- a/docs/data-download.md
+++ b/docs/data-download.md
@@ -30,7 +30,7 @@ usage: freqtrade download-data [-h] [-v] [--logfile FILE] [-V] [-c PATH]
 optional arguments:
   -h, --help            show this help message and exit
   -p PAIRS [PAIRS ...], --pairs PAIRS [PAIRS ...]
-                        Show profits for only these pairs. Pairs are space-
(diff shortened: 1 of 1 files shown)


docs/data-download.md
```diff
@@ -30,7 +30,7 @@ usage: freqtrade download-data [-h] [-v] [--logfile FILE] [-V] [-c PATH]
 optional arguments:
   -h, --help            show this help message and exit
   -p PAIRS [PAIRS ...], --pairs PAIRS [PAIRS ...]
-                        Show profits for only these pairs. Pairs are space-
+                        Only download data for these pairs. Pairs are space-
```
xmatthias, line 33: that won't change the output of the help command though ... and that is ... difficult to make it work for all - as the same argument is used by plot commands, as well as download data and list-data (and convert_data)

TheJoeSchr:
Okay 👍

xmatthias:
we can / should however refrase it in a way that's better suited for "Multi-purpose " ... ;)

PR #4534, opened 2021-03-13
Provide access to strategy instance from hyperopt class.

### LINKED ISSUE #3464: Using Hyperopt output to input Strategy run and backtesting
## Describe the enhancement

Export a Json File with hyperopt result as a strategy name and timeframe pattern on the name, eg.: Stragegy_5m.json with ROI, stop-loss, hyperopt parameters, etc.

I'm not a python developer to best implement this.

Automatically or parameter load generated file during backtesting and execution strategy overriding parameter in strategy, using this dict like param is used in hyperopt file, make only one code to hyperopt and strategy execution.

rokups:
As we talked on discord, this would allow reusing things defined in strategy class more easily.

closes #3464

### DIFF
--- a/docs/advanced-hyperopt.md
+++ b/docs/advanced-hyperopt.md
@@ -6,7 +6,7 @@ class.
 
 ## Derived hyperopt classes
 
-Custom hyperop classes can be derived in the same way [it can be done for strategies](strategy-customization.md#derived-strategies).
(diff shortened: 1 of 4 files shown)


xmatthias (approved):
Great :+1: 

I hope this will offer the flexibility we expect :100:

PR #4539, opened 2021-03-13
[WIP] Dynamic ROI

werkkrew:
## Summary
This PR adds a new set of options to enhance how the "minimal roi" feature works. The goal is to provide a smoother transition between points in the standard ROI table when using the 'connect' method, or, to override the standard ROI table entirely and use a linear or exponential decay to define the ROI rate.

## Quick changelog

- Added support for dynamic ROI in strategy, config, and hyperopt.
- Updated documentation with instructions for how to use it.

## Work in progress!

- Started working on making sure these changes are compliant with the hyperopt and strategy unit tests but got in a bit over my head. I am not clear on how to sort out what the expected results are without knowing what data was used to generate the results. 
- Unsure if I captured making the needed changes in every place necessary.  I believe I caught most of it, but there is a good chance I missed updating something, somewhere. Obviously I want to make sure this feature is a complete and non-disruptive addition to the project.

## What's new?
Adds a new set of options as follows:

```python
    dynamic_roi = {
        'dynamic_roi_enabled': True,
        'dynamic_roi_type': 'connect',   # linear, exponential, or connect
        'dynamic_roi_rate': 0.015,         # For exponential: bigger is faster, recommended to graph f(t) = start-pct * e^(-rate*t)      
        'dynamic_roi_time': 1440,         # For linear: amount of time to reach zero, only relevant for linear decay
        'dynamic_roi_start': 0.10,          # starting percentage
        'dynamic_roi_end': 0                 # ending percentage 
    }
```

My work on the unit tests in incomplete because, frankly, I am not sure how to finish it. As you know I am a novice developer so I have done my best here but I am open to suggestions, improvements, and support to make sure this can become a clean and seamless addition to freqtrade.


freqtrade/strategy/interface.py
```diff
@@ -634,12 +634,49 @@ def min_roi_reached_entry(self, trade_dur: int) -> Tuple[Optional[int], Optional
...
-        if not roi_list:
-            return None, None
-        roi_entry = max(roi_list)
-        return roi_entry, self.minimal_roi[roi_entry]
+        dynamic_roi = self.dynamic_roi
+        minimal_roi = self.minimal_roi
+        start, end = dynamic_roi['dynamic_roi_start'], dynamic_roi['dynamic_roi_end']
+
+        # if the dynamic_roi dict is defined and enabled, use it, otherwise fallback to default functionality
+        if dynamic_roi and dynamic_roi['dynamic_roi_enabled']:
```
xmatthias: at first glance:
Definitely extract this from the regular roi function ... 
you can call it from that function - but otherwise it's ... quite complex and unclear what belongs to what.

you'll also need to make  sure (in general for this) that the keys exist ... so probably before calling this (new) function to get the ROI key - check if the config structure for this exists (it's not mandatory and shouldn't be, so you'll need to approach with "doesn't exist" ... that's also probably breaking the tests ...)

---

For tests ... best "pick" this function as standalone and test the different possibilities.
I assume (!) from the ton of code in this function that you have a pretty clear idea how it should work ... so best start with the config examples in the docs ... add these to the configuration in a test ... and then check if the output (after running this function with different inputs) corresponds to what you'd expect in the output ... 

It's ok if it's one test per scenario - or even one test per imaginative scenario / value ... parametrization / combining can be a later refactoring ... 
but for now - it's quite unclear to me (at first glance, anyway) what we'd expect this function to return for the different new options - and i hope that some tests will shed some light into this.

werkkrew: Okay, thanks for the feedback! Right now I have this working just fine overridden in my strategy and I wanted to contribute it to the project. I am probably in over my head though in terms of the testing and all that so I'll do my best to start with the suggestions you have made.

If I bang my head against the wall too much and don't make and significant headway in the coming days I may just end up withdrawing this PR as it is very likely above my skill level to implement it correctly, I believe strongly in the idea but I do not have a ton of confidence in my ability to implement it.

Hopefully I can get it across the finish line though!

werkkrew: Okay, I did the part you asked about in terms of removing the changes from the existing function and made it into its own. I also added a bunch of error checking to ensure the required fields exist when using the various types of dynamic ROI.

If you can give what I've done so far a bit of a check before I try to dive into the unit testing bits to make sure the basis I am working from in the core parts of the code are good, I'd appreciate it. 

Once I know that I have made all of the changes, in the correct ways, in the correct places outside of the unit testing, I will tackle that part to make sure everything checks out.

docs/configuration.md
```diff
@@ -132,6 +133,7 @@ The following parameters can be set in either configuration file or strategy.
 Values set in the configuration file always overwrite values set in the strategy.
 
 * `minimal_roi`
+* `dynamic_roi`
```
xmatthias, line 136: to have this actually working (comming from strategy or config), you'll need to add it to `strategy_resolver.py` (most likely with a sensible fallback so it's not undefined)

that's also what's killing tests, as this value is simply not defined (it's not none, but it's "attribute does not exist").

docs/configuration.md
```diff
@@ -249,6 +251,76 @@ If it is not set in either Strategy or Configuration, a default of 1000% `{"0":
...
+* The `linear` type will decay from the `start` value to the `end` value in a straight line over `decay-time` (in minutes). Formula: `f(t) = start - (rate * t)` where `rate = (start -end) / decay-time`
+* The `exponential` type will decay from the `start` value in an exponential curve with `decay-rate` and end at minimum value of `end`. Formula: `f(t) = start * e^(-decay-rate*t)`
+* The `connect` type will "connect the dots" on the minimal_roi table with straight lines returning the values between the points.
+
+Exapmple Linear Config:
+```json
+    "dynamic_roi": {
+        "dynamic_roi_enabled": true,
+        "dynamic_roi_type": "linear",
+        "dynamic_roi_time": 720,
```
xmatthias, line 269: what's the unit of "720"? minutes? candles? seconds? hours? ... ?

freqtrade/optimize/hyperopt_interface.py
```diff
@@ -216,6 +216,39 @@ def trailing_space() -> List[Dimension]:
...
+        """
+        Create a dynamic ROI space.
+
+        You may override it in your custom Hyperopt class.
+        """        
+        return [
+            Categorical([True, False], name='dynamic_roi_enabled'),
+            Categorical(['linear', 'exponential', 'connect'], name='dynamic_roi_type'),
+            Real(0.001, 0.03, name='dynamic_roi_rate'),
+            Integer(180, 1440, name='dynamic_roi_time'),
```
xmatthias: unless the value of this is in candles ... i don't think this space is usable as is for different timeframes

if i'm on 1m candles, i don't wanna wayt for a day - or even for 180 minutes potentially ...while if i'm trading 4h trades, 1 day might not be long enough.

freqtrade/strategy/interface.py
```diff
@@ -650,7 +714,12 @@ def min_roi_reached(self, trade: Trade, current_profit: float, current_time: dat
         """
         # Check if time matches and current rate is above threshold
         trade_dur = int((current_time.timestamp() - trade.open_date_utc.timestamp()) // 60)
-        _, roi = self.min_roi_reached_entry(trade_dur)
+        
+        if self.dynamic_roi and 'dynamic_roi_enabled' in self.dynamic_roi \
```
xmatthias, line 720: `self.dynamic_roi` does not exist ... 
you'll 
* need to define it in the interface class (i'm pretty sure mypy will complaina bout this)
* actually assign it with "something" in `strategy_resolver.py`

xmatthias (commented):
I am not convinced that this is a better option than the current way (step-based ROI Decay). 
I think it has the potential to make backtesting unreliable / unprecise - obviously depending on how and which of the options is going to be used.

letme explain:
the current ROI uses a step-wise method ... so your function will look like this (assuming multiple steps):

```
roi
|----
|     |_____
|              |___
|___ _____ _____ ___ > time 
```

Ideally, you'll align the "minutes" to a timeframe, so ROI drops lower at every new candle (or every few 100 candles), making the value clear during backtesting (where you only get 1 timepoint of the candle).


Now we're trying to connect these dots (i'll use connect as sample, as it's probably the easiest to compare)
by connecting the dots, the points between the points become dynamic.

It can work okish in live (although it may also lead to lower profits - assuming a high "ROI jump") - it'll be higly problematic for backtesting (and with that also for hyperopt).

Assume  ROI config drops from 5% to 1% within 1 candle.
obviously backtesting will use the "highest" value of that, because it's the earliest (we use open to determine this) ... but it'll never even get called for anything between 5% and 1% ... so it'll probably sell after that one candle for 1%.

if you space it out across many candles it might be better ... but the "short-term" roi decay can become pretty "jumpy"  still - and show resuts that seem like this function is not working at all.

---


i'm not quite convinced yet this fixes a problem that exists / users need to have solved this way, while introducing quite some complexity.

it'll still keep the major painpoints people see / ask for (no ROI per pair, no ability to modify ROI based on some random things) - potentially even making that worse (as now it's another functionality that can't really be customized outside of the standard parameters).

in my opinion, something similar to custom stoploss for ROI might be a better option, as it provides a lot more flexibility - but i've only had a quick glance, therefore, maybe a closer look / try with it will give me a different opinion.

docs/configuration.md
werkkrew: Minutes

freqtrade/optimize/hyperopt_interface.py
werkkrew: It would probably be expected that the user override this in almost every case but I understand your point here.

werkkrew:
You could very well be correct on your assessment.  I wrote it as an experiment and thought maybe it was worth the effort to extend into something as you see here. It feels right in my mind to have something without static steps that decays over time and I see a lot of users asking for something like this in the discord channels so I figured I'd give it a go. I haven't done extensive experimentation with it to see if any of these methods have a significant impact (positive or negative) on profits.

So with that, before I put any more time/effort into it (which is already sort of significant) please let me know if you think it's really worth pursuing further. 

For my own use I am happy to continue just having it in the strategy itself where it is today, and fully functions. If you think there is questionable/limited value in having this in the project at-large I am perfectly happy to accept your determination about that! It would certainly save me the time/effort of making it 100% compatible with Freqtrade if its not a valuable feature.

_edited_
---
So at this point I got it to pass all the tests, that was a bit of a pain! Where do we go from here?

As far as what you/we do with this, I don't particularly care and won't be offended if you think it's just not worth merging - but I'd rather know that sooner than later. Just brushing up on my git/python skills and getting it to pass all the tests and validation has been a fun learning experience so I don't feel it was a waste of time in any case.

In terms of what it needs at this point to get it past the finish line, I'm not sure. I imagine it needs some unit tests?

docs/configuration.md
werkkrew: Ok I think I handled this.

freqtrade/strategy/interface.py
werkkrew: I added the needed line in the `strategy_resolver.py` file.

freqtrade/optimize/hyperopt_interface.py
werkkrew: Did a time scaling thing here similar to the standard roi table hyperspace.

werkkrew:
I am heavily leaning toward removing this PR.  The more I think about it the less I think this seems good to add to the main bot and if I (or anyone else) want this functionality we can continue to overload the methods in the strategy itself.

Let me know if you agree and I will close it, if you are still considering merging it, I am happy to leave it open.

werkkrew:
Closing this as per previous comment!

xmatthias:
it's unfortunate i think - i didn't have time to dive deeper into this (looking at a bigger change takes time i know i'll not be interrupted), but maybe you're right that having it as part of a strategy is better (i honstly didn't look that far just yet).

werkkrew:
I'd be happy to re-open it if you like.  I did not close it out of lack of patience, I understand you are very busy and that this is not a priority.  I closed it because I wound up completely re-writing how I am doing dynamic ROI on my side that does not resemble this at all, and frankly the way I am doing it now is probably 1) something I'd rather keep private, and 2) something that you wouldn't at all want in the base project even worse than this.

Issue #4532, opened 2021-03-13
Check fails if computer running check is not UTC


jlaw:
When I run pytest, this check fails because my computer is not set to UTC.
https://github.com/freqtrade/freqtrade/blob/441d3fad3983afb889ae8e9b05be34a5826bd870/tests/commands/test_commands.py#L700-L710

Code should be changed to the following to match the timestamp of the download data:
```python
 days_ago = arrow.get(arrow.now().shift(days=-20).date()).int_timestamp
```

xmatthias:
my computer is not set to UTC either, but to a different timezone (UTC+1 at the moment) - so i don't think that's the problem, but it's a problem with your computer's time settings.

Can you run show the output of `timedatectl` (without parameter).

Please note: 
for a correct time, your pc should have the correct time + timezone.
if you leave timezone at UTC (or whatever) and change the base time, all dates will be off (and you'll also have problems connecting to the exchange in that case).

jlaw:
Since you are UTC+1, you won't have a problem unless you test your code during the date difference (around midnight). When you run `pytest` between 00:00-00:59, UTC will be 23:00-23:59 of the previous day and it will fail.

Here is my output:
```
               Local time: Sat 2021-03-13 10:04:50 PST
           Universal time: Sat 2021-03-13 18:04:50 UTC
                 RTC time: Sat 2021-03-13 18:04:50
                Time zone: America/Los_Angeles (PST, -0800)
System clock synchronized: yes
              NTP service: active
          RTC in local TZ: no
```

Since I am UTC-8, There is 8 hours (during my evening) when I am testing code and cannot have that test pass successfully because UTC is already next day but I am 8 hours behind, until it is midnight for me.

xmatthias:
interresting, i guess you're right.

Are you happy to submit this as a PR to fix it? i guess it's best done by yourself as you'll be "right there" to test it (i'll test it on my side as well, but i haven't noticed this as problem so far)

jlaw:
Yup here is the PR!

PR #4540, opened 2021-03-13
fix: Use now() to match timezone of download data

jlaw:
## Summary
Fix `test_download_data_timerange()` in cases when download data is not timestamped for UTC

Solve the issue: #4532

## Quick changelog

- Use `arrow.now()` instead of `arrow.utcnow()`
## What's new?
Since the downloaded data timestamp is local time, the test will only pass if it is compared to local time. For times when this test case is ran where there is a date discrepancy between UTC and local time, the test will fail.

This PR fixes the issue by comparing the timestamp of the downloaded data with the local date.


xmatthias (approved):
LGTM :+1:

PR #4543, opened 2021-03-14
correct math used in examples and clarify some terminology regarding …

brookmiles:
…custom stoploss functions

## Summary
Correct the math used in the custom stoploss example documentation, and clarify a few points of terminology.

An example utility function `stoploss_from_open` is introduced which can be used to calculate stop loss values which are relative to the open price, instead of the current price.  This is something that is shown several times in the example documentation, but the math used previously was incorrect. In some cases the results were *close*, especially when dealing with very small percentages, however as the current profit increases, the previous math used becomes increasingly inaccurate.

### DIFF
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -181,17 +181,9 @@ class AwesomeStrategy(IStrategy):
 
 #### Calculating stoploss relative to open price
 
-Stoploss values returned from `custom_stoploss` always specify a percentage relative to `current_rate`.  In order to set a stoploss relative to the *open* price, we need to use `current_profit` to calculate what percentage relative to the `current_rate` will give you the same result as if the percentage was specified from the open price.
+Stoploss values returned from `custom_stoploss()` always specify a percentage relative to `current_rate`.  In order to set a stoploss relative to the *open* price, we need to use `current_profit` to calculate what percentage relative to the `current_rate` will give you the same result as if the percentage was specified from the open price.
 
-This can be calculated as:
-
-``` python
-def stoploss_from_open(open_relative_stop: float, current_profit: float) -> float:
-    return 1-((1+open_relative_stop)/(1+current_profit))
-
-```
-
-For example, say our open price was $100, and `current_price` is $121 (`current_profit` will be `0.21`).  If we want a stop price at 7% above the open price we can call `stoploss_from_open(0.07, 0.21)` which will return `0.1157024793`.  11.57% below $121 is $107, which is the same as 7% above $100.
+The helper function [`stoploss_from_open()`](strategy-customization.md#stoploss_from_open) can be used to convert from an open price relative stop, to a current price relative stop which can be returned from `custom_stoploss()`.
 
 #### Trailing stoploss with positive offset
 
@@ -201,9 +193,7 @@ Use the initial stoploss until the profit is above 4%, then use a trailing stopl
 ``` python
 from datetime import datetime, timedelta
 from freqtrade.persistence import Trade
-
-def stoploss_from_open(open_relative_stop: float, current_profit: float) -> float:
-    return 1-((1+open_relative_stop)/(1+current_profit))
+from freqtrade.strategy import stoploss_from_open
 
 class AwesomeStrategy(IStrategy):
 
@@ -237,9 +227,7 @@ Instead of continuously trailing behind the current price, this example sets fix
 ``` python
 from datetime import datetime
 from freqtrade.persistence import Trade
-
-def stoploss_from_open(open_relative_stop: float, current_profit: float) -> float:
-    return 1-((1+open_relative_stop)/(1+current_profit))
+from freqtrade.strategy import stoploss_from_open
 
 class AwesomeStrategy(IStrategy):
 
@@ -290,7 +278,7 @@ class AwesomeStrategy(IStrategy):
             # using current_time directly (like below) will only work in backtesting.
             # so check "runmode" to make sure that it's only used in backtesting/hyperopt
             if self.dp and self.dp.runmode.value in ('backtest', 'hyperopt'):
-              relative_sl = self.custom_info[pair].loc[current_time]['atr]
+              relative_sl = self.custom_info[pair].loc[current_time]['atr']
             # in live / dry-run, it'll be really the current time
             else:
               # but we can just use the last entry from an already analyzed dataframe instead
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -178,17 +178,15 @@ class AwesomeStrategy(IStrategy):
         return -0.15
 ```
 
-
 #### Calculating stoploss relative to open price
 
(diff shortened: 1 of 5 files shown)


docs/strategy-advanced.md
```diff
@@ -177,16 +178,33 @@ class AwesomeStrategy(IStrategy):
...
 
-Please note that the stoploss can only increase, values lower than the current stoploss are ignored.
 
 ``` python
 from datetime import datetime, timedelta
 from freqtrade.persistence import Trade
 
+def stoploss_from_open(open_relative_stop: float, current_profit: float) -> float:
+    return 1-((1+open_relative_stop)/(1+current_profit))
+
```
xmatthias: Instead of copy/pasting the same function in 3 samples ... (which basically results in users having to copy/paste it to each strategy ...

should this maybe be added to the codebase so it can be imported (similar to `merge_informative_pair()`) ?

brookmiles: For sure, I'll take a look.

brookmiles: I've moved `stoploss_from_open` into a helper function, and updated the docs and tests accordingly.

tests/strategy/test_strategy_helpers.py
```diff
@@ -95,3 +97,38 @@ def test_merge_informative_pair_lower():
...
+
+                    assert stoploss >= 0
+                    assert stoploss <= 1
+
+                    stop_price = current_price * (1 - stoploss)
+
+                    # there is no correct answer if the expected stop price is above
+                    # the current price
+                    if expected_stop_price > current_price:
+                        assert stoploss == 0
```
xmatthias, line 132: just running my thoughts past you more than anything...

If i call this with `stoploss_from_open(-0.5, -0.57)` ... i get 0.0. 
i guess this is expected, as obviously, current-profit is lower than the "desired" stoploss location. 

By returning 0.0 in this case, custom-stoploss will not be adjusted.

docs/strategy-advanced.md
```diff
@@ -197,27 +205,29 @@ class AwesomeStrategy(IStrategy):
...
+            return 1 # return a value bigger than the inital stoploss to keep using the inital stoploss
 
         # After reaching the desired offset, allow the stoploss to trail by half the profit
-        desired_stoploss = current_profit / 2
-
         # Use a minimum of 2.5% and a maximum of 5%
-        return max(min(desired_stoploss, 0.05), 0.025)
+        desired_stop_from_open = max(min(current_profit / 2, 0.05), 0.025)
+
+        return stoploss_from_open(desired_stop_from_open, current_profit)
```
xmatthias: I am not certain if applying `stoploss_from_open()` in this example is correct.

> Use the initial stoploss until the profit is above 4%, then use a trailing stoploss of 50% of the current profit with a minimum of 2.5% and a maximum of 5%.

The way i understand this example:
* profit < 4%  - use initial "static" stoploss.
* profit > 4%
  * 4.2% - trail with 2.1% (so stoploss should be set 2.1% below the current stoploss)
  * 5% - trail with 2.5% 
  * 20% - trail with 10% (50% of the profits - but capped to 5% by the above min/max limits) 
     This is where i think the "from open" goes wrong. Up to this point, the relative stoploss (actually accidentally, by using "half" the current profit) was always at the mid-point between profit and open, so both will return roughly the same result (not exactly - but close enough not to notice).
     The correct return for this case however should still be 5% - as I'd expect ~15% of the profits to be locked in - and trail with 5% behind my 20% profit (which obviously, i hope continues to 200%). By using `stoploss_from_open()`, the stoploss would be set to 5% above open (and never moved again as it's "top-capped" by the min/max limits) - potentially costing me most of the profits.

xmatthias (commented):
great job! :+1: 
I've added one link between the 2 doc pages - just for completeness, really.

I also have doubts about applying this example to "Trailing stoploss with positive offset" ...

docs/strategy-advanced.md
```diff
@@ -266,7 +276,7 @@ class AwesomeStrategy(IStrategy):
             # using current_time directly (like below) will only work in backtesting.
             # so check "runmode" to make sure that it's only used in backtesting/hyperopt
             if self.dp and self.dp.runmode.value in ('backtest', 'hyperopt'):
-              relative_sl = self.custom_info[pair].loc[current_time]['atr]
+              relative_sl = self.custom_info[pair].loc[current_time]['atr']
```
xmatthias, line 280: Great spot! :+1:

tests/strategy/test_strategy_helpers.py
brookmiles: Right, I think this is mostly a question of what you would like the behaviour to be in this case.  

I added this rule because while it's mathematically possible to return a result that is higher than the current price, it doesn't make any sense to return that value from custom_stoploss(), especially since the abs() value will be used, giving a totally wrong result.

I believe returning 0 from custom_stoploss() will cause the bot to log a warning, which I think is a good compromise behaviour.  It will tell you something is wrong, but not fatally so.

Another possibility would be to return a chosen very small value such as 0.0001, to cause the stoploss to be set nearly at the current price.  Since the user asked for a stoploss above the current price, this would be the closest possible result that is still valid.  This is actually the behaviour I would personally want most of the time, but I can also easily achieve it by returning max(0.0001, result), if returning a small value seems too arbitrary for default behaviour.

xmatthias: yes, it's mostly a question - and causing the warning to be logged sounds good to me.

an alternative might be to return a high number (1?) - causing the stop to be ignored (without log spamming) - but it'll lead to users not noticing this behavior - and usually it'll point to a faulty logic in the stoploss function - so i guess this will be ok as is.


you'll need to be careful with values too close to current price - if price moves against you in that moment (price always moves), you'll cause a emergency sell (if the current price dropped below the stop price) - at least if you're using stoploss on exchange.

docs/strategy-advanced.md
brookmiles: Ah, yes you're right, in order to keep the 5% trailing, it would need to be:

```python
        desired_stop_from_open = current_profit - max(min(current_profit / 2, 0.05), 0.025)
```

But at this point I think for the sake of simplicity it's probably better just to revert this example to the way it was before.  Either way works, but the original is definitely simpler, and in this case doesn't change the result very much.  `stoploss_from_open()` here is probably more confusing than helpful.

docs/strategy-advanced.md
```diff
@@ -203,21 +204,29 @@ class AwesomeStrategy(IStrategy):
         desired_stoploss = current_profit / 2
 
         # Use a minimum of 2.5% and a maximum of 5%
-        return max(min(desired_stoploss, 0.05), 0.025)
+        return max(min(desired_stoploss, 0.05), 0.025
```
```suggestion
        return max(min(desired_stoploss, 0.05), 0.025)
```

xmatthias (approved):
LGTM :+1: 

Thanks a lot for finding (and correcting) the doc incorrectness in such a great way!

PR #4562, opened 2021-03-16
Add support for trailing_stop_loss order_type

rextea:
## Summary
Add an optional parameter to set a custom order_type for trailing_stop_loss sell reason

## What's new?
Add an optional parameter to set a custom order_type for trailing_stop_loss sell reason

### DIFF
--- a/config_full.json.example
+++ b/config_full.json.example
@@ -52,6 +52,7 @@
         "forcesell": "market",
         "forcebuy": "market",
         "stoploss": "market",
+        "trailing_stop_loss": "limit",
         "stoploss_on_exchange": false,
         "stoploss_on_exchange_interval": 60
     },
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -280,13 +280,13 @@ For example, if your strategy is using a 1h timeframe, and you only want to buy 
 The `order_types` configuration parameter maps actions (`buy`, `sell`, `stoploss`, `trailing_stop_loss`, `emergencysell`, `forcesell`, `forcebuy`) to order-types (`market`, `limit`, ...) as well as configures stoploss to be on the exchange and defines stoploss on exchange update interval in seconds.
 
 This allows to buy using limit orders, sell using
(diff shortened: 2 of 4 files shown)


xmatthias (changes requested):
Except for the fact that i don't understand why we'd need this, it's also not working the way you expect it to (at least not with stoploss on exchange).

A regular stoploss and trailing stoploss are not different - 
the "trailing" version replaces the initial stoploss, there's no benefit i can see to using different types for this.
If the reason for this is the risk of blowing past the trailing "limit" stoploss order (or not getting that filled) - then having "stoploss" set to "market" will not help you, as the initial stoploss will NOT be in place any longer - and will fore be ignored.

That's why i recommend using stoploss on exchange - as it'll make sure your stop is placed "early". While there's still a minimal risk of blowing past that stoplos on very violent moves - i'm yet to see ONE case where that happened.

rextea:
I've fixed the behavior of it, now it should work the way I've expected. (although the execute_sell method is too complicated according to flake8)
As for why it's needed, you may be right with setting stoploss on exchange, i'll give it a try. but still maybe for cases where you don't use stoploss on exchange it may be usuful.

xmatthias:
i'm not sure i see why ... even without "stoploss on exchange" there's no real difference between the 2 sells - other than the fact that the trailing one has been raised multiple times, and is therefore closer to current price.

rextea:
In a case where I want the stoploss to sell on market (to avoid timeout of stoploss - happend to me couple of times), but as for trailing stop loss I care less if it will timeout and I prefer to take as much profit I can - that's why I want it to be limit sell

xmatthias:
I don't think this will give you what you want.
Assuming the trailing stop will timeout - then the bot will try to replace it with a new limit order, which can eventually again timeout.

Assuming your first "trailing stop" hit was at 2% profit - you can go down to -5% (or down to -20%) following the price with this method, each trailing-stop order timing out (it will NOT fall back to the original stoploss once it started trailing).

PR #4571, opened 2021-03-18
Improvement for timerange parser

pbrunier:
## Summary
When users run for example back testing and their start date is AFTER their stop date, the syntax is OK, but they get an error that data has not been found. I decided include a nice error message for this case, because the error is a bit cryptic when you overlook your error. 

Solve the issue: Not an existing issue on Github.

## Quick changelog
-Added a small snippet to give users a descent error message, when their start date is after the stop date. (timerange.py)
-Added test for the exception (test_timerange.py)

## What's new?
Users now get the message "Start date is after stop date for timerange" when they make the above mistake.

### DIFF
--- a/freqtrade/configuration/timerange.py
+++ b/freqtrade/configuration/timerange.py
@@ -103,5 +103,7 @@ class TimeRange:
                         stop = int(stops) // 1000
                     else:
                         stop = int(stops)
+                if start > stop > 0:
+                    raise Exception('Start date is after stop date for timerange "%s"' % text)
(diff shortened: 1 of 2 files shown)


pbrunier:
Had some PEP8 issues with my linter in 2 previous PRs. Sorry for that. Flake8 is enabled now.

xmatthias (approved):
Great, nice one! :+1: 

I wanted to do this for quite some time, but never gor around to it!

LGTM :+1: 

Thanks for your contribution :100:

PR #4594, opened 2021-03-23
Add confirm_trade_exit and confirm_trade_entry to backtesting

rextea:
## Summary
Add confirm_trade_exit and confirm_trade_entry to backtesting

Solve the issue: #___

## Quick changelog

- backtesting.py

## What's new?
Add confirm_trade_exit and confirm_trade_entry to backtesting

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -30,7 +30,6 @@ from freqtrade.strategy.interface import IStrategy, SellCheckTuple, SellType
 from freqtrade.strategy.strategy_wrapper import strategy_safe_wrapper
 from freqtrade.wallets import Wallets
 
-
 logger = logging.getLogger(__name__)
 
 # Indexes for backtest tuples
@@ -253,20 +252,21 @@ class Backtesting:
                                          sell_row[DATE_IDX], sell_row[BUY_IDX], sell_row[SELL_IDX],
                                          low=sell_row[LOW_IDX], high=sell_row[HIGH_IDX])
 
-        time_in_force = self.strategy.order_time_in_force['sell']
-        # confirm_trade_exit
-        if not strategy_safe_wrapper(self.strategy.confirm_trade_exit, default_retval=True)(
-                pair=trade.pair, trade=trade, order_type='limit', amount=trade.amount,
-                rate=sell_row[LOW_IDX],
-                time_in_force=time_in_force,
-                sell_reason=sell.sell_type.value):
-            return None
-
         if sell.sell_flag:
             trade.close_date = sell_row[DATE_IDX]
             trade.sell_reason = sell.sell_type.value
             trade_dur = int((trade.close_date_utc - trade.open_date_utc).total_seconds() // 60)
             closerate = self._get_close_rate(sell_row, trade, sell, trade_dur)
+
+            # Confirm trade exit:
+            time_in_force = self.strategy.order_time_in_force['sell']
+            if not strategy_safe_wrapper(self.strategy.confirm_trade_exit, default_retval=True)(
+                    pair=trade.pair, trade=trade, order_type='limit', amount=trade.amount,
+                    rate=closerate,
+                    time_in_force=time_in_force,
+                    sell_reason=sell.sell_type.value):
+                return None
+
             trade.close(closerate, show_msg=False)
             return trade
 
@@ -283,7 +283,7 @@ class Backtesting:
 
         order_type = self.strategy.order_types['buy']
         time_in_force = self.strategy.order_time_in_force['sell']
-        # confirm_trade_entry
+        # Confirm trade entry:
         if not strategy_safe_wrapper(self.strategy.confirm_trade_entry, default_retval=True)(
                 pair=pair, order_type=order_type, amount=stake_amount, rate=row[OPEN_IDX],
                 time_in_force=time_in_force):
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -252,8 +251,17 @@ def _get_sell_trade_entry(self, trade: LocalTrade, sell_row: Tuple) -> Optional[
         sell = self.strategy.should_sell(trade, sell_row[OPEN_IDX],  # type: ignore
                                          sell_row[DATE_IDX], sell_row[BUY_IDX], sell_row[SELL_IDX],
                                          low=sell_row[LOW_IDX], high=sell_row[HIGH_IDX])
-        if sell.sell_flag:
 
+        time_in_force = self.strategy.order_time_in_force['sell']
+        # confirm_trade_exit
+        if not strategy_safe_wrapper(self.strategy.confirm_trade_exit, default_retval=True)(
```
xmatthias: Did you do some performance testing with / without this change?

i do have my concerns about performance with this - which is why i intentionally didn't add this functionality to backtesting in the first place.

This can be especially relevant for strategies with a lot of trades (a few 1000 over a year), small candle sizes (5m) - and relatively long holding times - as then this function is HAMMERED (the sell function alone will be called once for every candle while a trade is open - now multiply this with the number of pairs and trades you open, and it'll be a problem).

In backtesting, every function call added can have a big performance impact (which implicitly also impacts hyperopt, so we need to be very careful with this), and why protections need to be explicitly enabled.


Can you please perform some benchmarking of this change:
A simple test can be using time (`time freqtrade backtesting ....`) - and comparing this vs. the underlying develop branch performance (best on a strategy that doesn't implement `confirm_*` methods).
Please use a strategy that does at least 2000 trades over a 1 year period on 5m candles, and post the timings before / after this change.

rextea: I've used it for more than a week and didn't recognized any performance issues.
Sure I will benchmark it and share the results here.

xmatthias: Great thanks!

also, If we now use this during backtesting / hyperopt, it should be documendted in `bot-basics.md` (which documents the flow).

rextea: Attached links with the benchmark results, It's with a confirm_trade_exit method implemented in the strategy:
49 days, develop: https://ibb.co/r4z3ck8
49 days, feature branch: https://ibb.co/2N7r1ZC
432 days, develop: https://ibb.co/vZ67St1
432 days, feature branch: https://ibb.co/cct81j8
I've also checked without confirm_* methods and the results are the same.

I've added some info to the bot-basics.md, Please let me know if it's fine
Also, I would like you to check the code and make sure the parameters i've passed to the methods are correct (as I'm not that familiar with the project)
Thanks!

xmatthias: > Please let me know if it's fine
Also, I would like you to check the code and make sure the parameters i've passed to the methods are correct (as I'm not that familiar with the project)

don't worry - that's DEFINITELY part of the review ... not because you're new or not, but because typos can happen easily.

freqtrade/optimize/backtesting.py
```diff
@@ -252,8 +251,17 @@ def _get_sell_trade_entry(self, trade: LocalTrade, sell_row: Tuple) -> Optional[
...
+                pair=trade.pair, trade=trade, order_type='limit', amount=trade.amount,
+                rate=sell_row[LOW_IDX],
```
xmatthias: I don't think you can (or should) use `sell_row[LOW_IDX]` here. "rate" should correspond to the rate used - which low does only do in some cases.

I'd rather:

* Move all of this into the below if block (so it's only executed if a sell-signal would actually happen)
* use the calculated `closerate` as rate (this is the rate a order would be placed)

This would make behaviour aligned to how it works in trade mode, where this is only called once a sell is triggered and allows you to stop the signal "last-second" once the sell-price (and everything else) is already predefined.


Apart from this, i think this looks good. (note that i did push one commit to fix the "isort" error, so make sure to do a pull on this branch before your next commit).

rextea: done

moinonin:
This will be a game changer. However, when I pull the update and run backtesting the bot sells with negative profit even when I have specified in the config "sell_profit_only": true.

The only thing I have changed concerning trade exit confirmation is:
        if sell_reason == 'emergency_sell' and trade.calc_profit_ratio(rate) < 0:
            # Reject emergency-sells with negative profit
            return False

Is anyone experiencing this problem?

xmatthias:
@moinonin your issue has nothing to do with this Pull request.

Also, emergency sells will NOT apply in backtesting at all.

moinonin:
> @moinonin your issue has nothing to do with this Pull request.
> 
> Also, emergency sells will NOT apply in backtesting at all.

Thanks @xmatthias, I removed all that in the trade exit and entry and the problem persists. Is there a chance that the configuration file, specifically the section specifying 'sell_profit_only' is not read or an earlier setting, say 'sell_profit_only': 'false' was cached in memory somewhere? or is it a bug possibly?

xmatthias:
> > @moinonin your issue has nothing to do with this Pull request.
> > Also, emergency sells will NOT apply in backtesting at all.
> 
> Thanks @xmatthias, I removed all that in the trade exit and entry and the problem persists. Is there a chance that the configuration file, specifically the section specifying 'sell_profit_only' is not read or an earlier setting, say 'sell_profit_only': 'false' was cached in memory somewhere? or is it a bug possibly?

Please open a seperate issue with all information required, and do not highjack Pull requests that have nothing to do with your problem.

xmatthias (approved):
LGTM :+1:

PR #4596, opened 2021-03-23
Support for creating auto-hyperoptable strategies.

rokups:
This PR implements a way to define hyperoptable strategies without creating a separate hyperopt class.

~~⚠️ This is still a sketch, i am not entirely happy about `Parameter` interface. Other areas can possibly be improved as well. ⚠️~~

This is achieved by:
1. `HyperOptAuto` class which serves as an automatic hyperopt class common for all auto-hyperoptable classes.
2. `HyperStrategyMixin` a mixin class for strategies, added as a part of `IStrategy` interface.

## A trivial example

All strategies can immediately hypperopt roi, stoploss and trailing spaces by omitting `--hyperopt` parameter. No other changes are required.

## A more involved example

We are most interested in hyperopting buy/sell parameters, this is how we would do such thing:

```py
from freqtrade.strategy import RealParameter, CategoricalParameter


class BinHV27(IStrategy):

    # It is possible to also define other methods that would normally be in a hyperopt class. It is done in a nested class named "HyperOpt":
    class HyperOpt:
        @staticmethod
        def stoploss_space() -> List[Dimension]:
            return [
                Real(-0.35, -0.02, name='stoploss'),
            ]

    # Define buy space parameter by using a convention: prefix parameter name with "buy_". We would use "sell_" prefix for sell space parameters.
    buy_adx_min = IntParameter(0, 100, default=20)
    # Define a buy space parameter by using completely custom name. We would use 'sell' category for sell space parameters.
    emarsi_max = IntParameter([0, 100], default=25, space='buy')

    # Example of defining a float parameter.
    sell_unused_float_example = RealParameter([0.0, 1.0], default=0.25)
    # Example of defining a categorical parameter. Value defaults to 'with_unused' because no default was specified.
    buy_unused_categorical_example = CategoricalParameter(['with_unused', 'without_unused'])

        dataframe.loc[
            (
                dataframe['continueup'] &
                dataframe['adx'].gt(self.buy_adx_min.value) &
                dataframe['bigup'] &
                dataframe['emarsi'].le(self.emarsi_max.value)
            ),
            'buy'] = 1
        return dataframe
```

## A full example that you can copy-paste and try

Command line: `freqtrade hyperopt -j 16 --spaces all -e 200 --timerange 20210301-20210323 --hyperopt-loss OnlyProfitHyperOptLoss -s SwingHighToSky`

```py
# --- Do not remove these libs ---
from skopt.space import Real

from freqtrade.strategy import IStrategy, DecimalParameter, CategoricalParameter
from typing import Dict, List
from functools import reduce
from pandas import DataFrame
# --------------------------------

import talib.abstract as ta
import numpy # noqa

__author__      = "Kevin Ossenbrück"
__copyright__   = "Free For Use"
__credits__     = ["Bloom Trading, Mohsen Hassan"]
__license__     = "MIT"
__version__     = "1.0"
__maintainer__  = "Kevin Ossenbrück"
__email__       = "kevin.ossenbrueck@pm.de"
__status__      = "Live"

class_name = 'SwingHighToSky'
class SwingHighToSky(IStrategy):
    class HyperOpt:
        # Define a custom stoploss space.
        @staticmethod
        def stoploss_space():
            return [Real(-0.05, -0.01, name='stoploss')]

    # Replace contents between HyperOpt parameters begin/end with output of hyperopt command.
    # HyperOpt parameters begin

    # Buy hyperspace params (loaded into corresponding class fields automatically):
    buy_params = {
        'buy_cci_max': -112, 'buy_use_cci': 1
    }

    # Sell hyperspace params (loaded into corresponding class fields automatically):
    sell_params = {
        'sell_cci_min': -190, 'sell_use_cci': 'no-use'
    }

    # ROI table:
    minimal_roi = {
        "0": 0.30813,
        "127": 0.14381,
        "370": 0.07412,
        "1038": 0
    }

    # Stoploss:
    stoploss = -0.04175

    # Trailing stop:
    trailing_stop = True
    trailing_stop_positive = 0.1202
    trailing_stop_positive_offset = 0.1645

    # HyperOpt parameters end

    ticker_interval = '30m'

    buy_cci_max = DecimalParameter([-300, +300], default=-100)
    sell_cci_min = DecimalParameter([-300, +300], default=+200)
    buy_use_cci = CategoricalParameter([False, True])
    sell_use_cci = CategoricalParameter(['no-use', 'use'])

    def informative_pairs(self):
        return []

        macd = ta.MACD(dataframe)
        dataframe['macd'] = macd['macd']
        dataframe['macdsignal'] = macd['macdsignal']
        dataframe['cci-buy'] = ta.CCI(dataframe, timeperiod=14)
        dataframe['cci-sell'] = ta.CCI(dataframe, timeperiod=14)

        return dataframe

        dataframe.loc[
            (
                    (dataframe['macd'] > dataframe['macdsignal'])
                    &
                    (
                        (self.buy_use_cci.value == False)
                        |
                        (dataframe['cci-buy'] <= self.buy_cci_max.value)
                    )
            ),
            'buy'] = 1

        return dataframe

        dataframe.loc[
            (
                    (dataframe['macd'] < dataframe['macdsignal'])
                    &
                    (
                        (self.sell_use_cci.value == 'no-use')
                        |
                        (dataframe['cci-sell'] >= self.sell_cci_min.value)
                    )
            ),
            'sell'] = 1

        return dataframe
```


freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,93 @@
...
+            return Integer(*self.space, name=name, **self._space_params)
+        if isinstance(self.value, float):
+            assert len(self.space) == 2
+            return Real(*self.space, name=name, **self._space_params)
+
+        assert len(self.space) > 0
+        return Categorical(self.space, name=name, **self._space_params)
+
+
+class IHyperStrategy(IStrategy, ABC):
```
xmatthias: I don't think you need to inherit from ABC here?

that's for abstract classes - which IStrategy is already inheriting from.
i don't think not implementing the attributes should block us from using it ... (might be wrong, didn't try specifically)

---

I'd however ... keep myself open for the moment on "reversing" inheritance of this.

If this works well - i can imagine using `IHyperStrategy` as a mixin to `IStrategy` (simply to keep the logic seperated), so it's "magically" available to strategies without changing the base class or whatnot.

While it would be a change in the interface - it's a non-breaking change (only adding new stuff) - so i don't think it'd be problematic.

freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,93 @@
...
+            return None
+        if isinstance(self.value, int):
+            assert len(self.space) == 2
+            return Integer(*self.space, name=name, **self._space_params)
+        if isinstance(self.value, float):
+            assert len(self.space) == 2
+            return Real(*self.space, name=name, **self._space_params)
+
+        assert len(self.space) > 0
+        return Categorical(self.space, name=name, **self._space_params)
```
xmatthias: I don't like the implicit behaviour this creates, as it's rather error-prone.

if i specify `Parameter(default=10.0, space=[20, -20])` 
it'll create a float space because i accidentally (!) typed in 10.0 instead of 10...

maybe having explicit subclasses (`FloatParam` | `RealParam` - `IntParam` and/or `CatParam`) or so would be a better idea? That'll make the decision explicit ... without infering from the default value.


I also think we need to be careful with the name - `Parameter` is ... pretty generic, so i'm slightly afraid of naming collisions with random packages that people may use.

freqtrade/optimize/hyperopt_auto.py
```diff
@@ -0,0 +1,83 @@
...
+    def _get_func(self, name) -> Callable:
+        """
+        Return a function defined in Strategy.HyperOpt class, or one defined in super() class.
+        :param name: function name.
+        :return: a requested function.
+        """
+        hyperopt_cls = getattr(self.strategy, 'HyperOpt')
+        default_func = getattr(super(), name)
+        if hyperopt_cls:
+            return getattr(hyperopt_cls, name, default_func)
```
xmatthias: The use of `getattr` and hasattr` slightly concerns me from a performance perspective.
Using it once to load a function (if it exists) - no problem.

but we'd be using it from (for example, not limited to) `generate_trailing_params()` - which is a function that's called once per epoch, so i think the performance impact this can have is non neglectible (will however need testing / benchmarking  - as i seem to remember that hasattr/getattr is slow - but i can't find that reference now).

I might be mistaking as well - it's something i "seem to remember" ... but who knows :laughing:

freqtrade/optimize/hyperopt_auto.py
```diff
@@ -0,0 +1,83 @@
...
+
+# noinspection PyUnresolvedReferences
+class HyperOptAuto(IHyperOpt):
+    """
+    This class delegates functionality to Strategy(IHyperStrategy) and Strategy.HyperOpt classes. Most of the time
+    Strategy.HyperOpt class would only implement indicator_space and sell_indicator_space methods, but other hyperopt
+    methods can be overridden as well.
+    """
+    def buy_strategy_generator(self, params: Dict[str, Any]) -> Callable:
+        assert hasattr(self.strategy, 'enumerate_parameters'), 'Strategy must inherit from IHyperStrategy.'
```
xmatthias: you'll need to be careful with using assert statements in non-test code.

I know it's a (kindof) simple/quick way to fail if something doesn't happen - however, assert failing can be disabled on a python level (`python -O ...` ) - which would have all assert-statements silently pass, and should therefore not be used in final / production code.

Therefore freqtrade usually uses "OperationalException" to fail - as we handle that gracefully before exiting (i know that'll convert this oneline into 2 lines... ).

freqtrade/optimize/hyperopt.py
```diff
@@ -67,8 +68,11 @@ def __init__(self, config: Dict[str, Any]) -> None:
 
         self.backtesting = Backtesting(self.config)
 
-        self.custom_hyperopt = HyperOptResolver.load_hyperopt(self.config)
-        self.custom_hyperopt.__class__.strategy = self.backtesting.strategy
+        if self.config['hyperopt'] == 'HyperOptAuto':
+            self.custom_hyperopt = HyperOptAuto(self.config)
```
xmatthias: as an idea - maybe we can have hyperopt default to `HyperOptAuto` if / once the remaining logic works as expected.
I'm putting this here as a reminder (also for myself) - but i'd like to keep that open for the moment.

freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,93 @@
...
+
+class IHyperStrategy(IStrategy, ABC):
+    """
+    A helper base class which allows HyperOptAuto class to reuse implementations of of buy/sell strategy logic.
+    """
+
+    def __init__(self, config):
+        super().__init__(config)
+        self._load_params(getattr(self, 'buy_params', None))
+        self._load_params(getattr(self, 'sell_params', None))
```
xmatthias, line 233: I think finally used parameters should be printed.

A typo in "buy_params" (one of the keys, maybe) can happen very quickly - and assuming the attempt  backtest the results for confirmation of hyperopt - this can lead to disconnects quickly.

As the parameter can fall back to the default value, this can lead to surprising results (in either backtesting or trade mode).

freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,93 @@
...
+        :param kwargs: Extra parameters to skopt.space.(Integer|Real|Categorical).
+        """
+        assert 'name' not in kwargs, 'Name is determined by parameter field name and can not be specified manually.'
+        self.value = default
+        self.space = space
+        self.category = category
+        self._space_params = kwargs
+        if default is None:
+            assert len(space) > 0
+            self.value = space[0]
```
xmatthias: I think i'd rather have "default" a required parameter (unless there's a good reason not to).

For categorical spaces using the first value might work ... but for an integer space, it's kinda odd ...

`Parameter(space=[-300, +300])` would pick -300, which is the negative extreme (and therefore rather unlikely to happen).

An alternative could be to use a "sensible default" - so a value in the middle in case of a `IntParameter` (see other comment), the first for the categorical ... but it's just an idea...

xmatthias (commented):
I think this is great engineering! :100:


One thing that keeps me thinking since i looked at this for the first time - can we simplify the interface of the `Parameter` class enough so it can be become `(dataframe['cci-sell'] >= self.sell_cci_min)` instead of `(dataframe['cci-sell'] >= self.sell_cci_min.value)`? (basically omitting the `.value`).

i suspect this should (somehow) be possible by using one of the __ (dunder) methods - but i fear the risk of "Miss-typing" a parameter is pretty high (so instead of `10`, we'd get `'10'`) - which can have pretty bad side-effects ... 
Maybe with the "explicit" subclasses for spaces this can be solved though ...

freqtrade/strategy/hyper.py
rokups: `ABC` is indeed not necessary. IDE showed a warning and this was one of the suggestions. I suppose it would be better to just add a comment that disables warning instead. Hmm mixin idea is great! I will try that 👍🏻 If it was to be a mixin class should probably be renamed too. `HyperOptable` maybe? Idk naming is hard.

freqtrade/strategy/hyper.py
rokups: I could not agree more! Separate parameter classes is a much better idea. Especially since it would allow categorial parameter to have no default (because `default='foo', space=['foo', 'bar']` isnt really nice), while mandating default value for float/int params as i do not think there can be a sensible default for those.

freqtrade/optimize/hyperopt_auto.py
rokups: Hmm i am not so sure that is the case. We could profile if in doubt.. but attribute access is just a `self.__dict__` lookup afaik.

freqtrade/optimize/hyperopt_auto.py
rokups: Will do! 👍🏻

freqtrade/optimize/hyperopt.py
rokups: I was considering making `--hyperopt` an optional parameter. That + checking whether strategy has required mixin could do a trick, missing this parameter would still result in an informative error message with clear description how to solve it.

freqtrade/strategy/hyper.py
rokups: Doesnt hurt i suppose, although params are quite in the clear in strategy class. I was also thinking about loading parameters from a file. Maybe hyperopt could save `StrategyName.json` next to `StrategyName.py`, which could be loaded by strategy. In such case printing would definitely be useful. What do you think about configuring these parameters through json?

freqtrade/strategy/hyper.py
rokups: -300 would not be picked for float/int as default must be set, but i do agree this stinks.`(Int|Float|Categorial)Parameter` would solve it cleanly, so one more reason to go that route.

freqtrade/strategy/hyper.py
rokups: Oh i did not mention `.value` bit. Yeah seems like i missed `__int__` being a thing. That + `__float__` + `__str__` + `__bool__` seems to be all we need. Thing is i am not sure if it would work considering other side of comparison would be pandas dataframe. This needs investigation 👀

rokups:
I addressed your comments. Formatting makes it hard to see what changed in diff though, sorry about that. There are few issues with mypy that i will need your help with.

> error: Incompatible types in assignment (expression has type "IHyperOpt", variable has type "HyperOptAuto")

Not sure why as `HyperOptAuto` inherits from `IHyperOpt` 👀 

Also i do not know how to declare `strategy` type in `IHyperOpt` to communicate possibility of `HyperStrategyMixin` existence. `Union[IStrategy, HyperStrategyMixin]` is not correct apparently.

> error: Signature of "buy_strategy_generator" incompatible with supertype "IHyperOpt"

This is an error because i declared `buy_strategy_generator` as instance method while `IHyperOpt` declares them as static. But i need that `self` so idk what to do here.

---

I also looked into `__int__`/`__float__`. Turns out there is little point in using them as they do not allow to only drop `.value`, we would also have to use `int()`/`float()`. At this point `.value` seems easier.

Edit: Also updated first post to reflect changes.

freqtrade/optimize/hyperopt.py
xmatthias: if you add the mixin to IStrategy (the way i intended it somewhere in this PR) ... the strategy doesn't have to opt-in explicitly.


Obviously, it'll have to define a space (for buy / sell parameters obviously) ... but that's checked in other places of hyperopt.

That would have the advantage that you can still hyperopt (automatically) a roi / stoploss without ANY change to an existing strategy.

xmatthias:
> I also looked into __int__/__float__. Turns out there is little point in using them as they do not allow to only drop .value, we would also have to use int()/float(). At this point .value seems easier.

that's fine - it was an idea - not a requirement ;) 

Also, don't worry about CI just yet - that'll be needed for the "final" touch - but i don't expect this to pass at the moment - as mypy  will be a pain (it's always is when doing something with hyperopt) 😆

freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,170 @@
...
+    space: Sequence[int]
+
+    def __init__(self, *, space: Sequence[int], default: int, category: Optional[str] = None,
+                 **kwargs):
+        """
+        Initialize hyperopt-optimizable parameter.
+        :param space: Optimization space, [min, max].
+        :param default: A default value.
+        :param category: A parameter category. Can be 'buy' or 'sell'. This parameter is optional if
+         parameter field
```
xmatthias: The more i look at the `space` and `category` options, the more it confuses me (from a naming perspective only)

I understand the `space` is the space i wanna optimize (min / max value), and category would be buy / sell.
The confusing part is with hyperopt - where we have `--spaces buy sell roi` - to optimize only certain parts - which are called spaces in the cli parameter - but suddenly corresponds to a "category" in this context.

I guess technically both are spaces (one is the space within this parameter - e.g. [-20, 20] ) - while the other is the overall (bigger scoped) search space of buy / sell ... 


---

The problem is - i don't have a better alternative at the moment (and i suspect that's how you got to these names in the first place too :laughing: )

freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,170 @@
+"""
+IHyperStrategy interface, hyperoptable Parameter class.
+This module defines a base class for auto-hyperoptable strategies.
+"""
+from typing import Iterator, Tuple, Any, Optional, Sequence, Union
+
+from skopt.space import Integer, Real, Categorical
```
xmatthias: this will become - uh - complicated now ...

technically, freqtrade doesn't require scikit-optimize (or scikit-learn) to be installed to run basic bot operations (basically everything except hyperopt / hyperopt utils).
This is important as installing scikit-learn on a raspberry is HORROR (compilation takes about a day last i tried).

I don't think that's going to be an issue (just some more code) - we can either wrap the import in a try/except block (catching `ImportError`), and disable parameter optimization for this case, while keeping the "parameter" itself alive (just without the `.get_space()` method).

alternatively, the imports could be moved to the respective "get_space" methods (basically the same approach, but on a different level) - and simply raise an OperationalException when called without having scikit-optimize installed (we don't expect to have `parameter.get_space()` called form non-hyperopt processes i'd think).

---

thinking about it, we should prefer the 2nd option i think.
This would avoid importing skopt when it's not necessary / used (we're not going to use it unless we're hyperopting) - and therefore speed up the startup process for "non-hyperopt" processes.
We'll obviously need to remove the typehints - but i guess we can survive without them for this class...

freqtrade/strategy/hyper.py
rokups: You are right. Another point i am not entirely happy with `space` is that it is kind of bit loose. You could pass a list with three items for int/float and get a runtime exception. skopt avoids this by explicitly passing min and max parameters. If we did that for int/float, we could use "choices" for categorical, that would free up "space" name to use in place of "category". We would have a consistency of terminology. Although i do admit i like `whatever_name=[0, 100]` better than `min=0, max=100`.

freqtrade/strategy/hyper.py
rokups: In 58cd40f  i made skopt imports optional by catching import error at module level and using type names as strings in type hints. This gives us best of both worlds - string type name still resolves (at least in intellij IDE) to a type while import is being optional.

freqtrade/strategy/hyper.py
xmatthias: hm, maybe we can adapt the "skopt" approach - using positional arguments (`IntParameter(0, 20, default=10, spaces=['buy'])`)?

it might not be as expressive than using low/high (that's the naming skopt uses for Int/Real spaces).
potentially we can even have both ways?

``` python
x = IntParameter(low=0, high=20, default=10, spaces=['buy'])
y = IntParameter([0, 20], default=10, spaces=['buy'])
```

with the 2nd version exploding the first parameter to low + high .... or failing on initialization if the first param has 3 values (so it'll fail while loading the strategy, not later)

rokups: Sounds like a perfect compromise. I love that `spaces=[]` touch. Would probably be useful for defining same categorial parameter for both spaces _once_.

xmatthias: jup ... it's actually ... not intentional that i changed it from how category works :D 

I'm however not 100% sure if it will work with more than one space if i think about it.

hyperopt would change the SAME parameter from 2 different generators ... but there's only one value / variable in the strategy - so you can't know which one was used (the output would have it in both spaces - but you can't be sure which one was actually used as the 2 would overwrite each other).

rokups: Oh good point. Lets ignore that then.. I mean we could make it work, but marginal usefulness it totally not worth extra complexity

freqtrade/strategy/hyper.py
xmatthias: i guess it's a possibility - but i'd leave that open for a future PR.

Having that auto-loading will make the logic complex (which one is used - self.sell_params, the config, or the default?) - and i'd like to see how this current concept works for (non-developer) users first.

As the file would be written automatically (otherwise there's little point), it would implicitly always use the "latest" parameters - which might not be desired (i may want to hyperopt a strategy - but i liked the "prior" version better).

freqtrade/optimize/hyperopt_auto.py
xmatthias: I'm not sure either ... but it appears to make a small difference after doing the benchmark.

Testing this with a very simple/stupid case:
```
class Xxx():
    def abc(self):
        pass

a = Xxx()

%timeit a.abc()
> 74.7 ns ± 2.53 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

%timeit getattr(a, 'abc')()
> 139 ns ± 1.6 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

While it seems to be a increase of around almost 100% - the numbers are however small enough to be neglectible (this is nano-seconds after all!) i'd think.

freqtrade/strategy/hyper.py
xmatthias: i think it could work - HOWEVER, i also think it's too dangerous, and using `.value` is better than forcing users to use `int()`.

``` python
class Xxx():
    def __int__(self):
        print("Int called")
        return 2
    
    def __float__(self):
        print("float called")
        return 2.2

a = Xxx()
df = pd.DataFrame({'aaa': [0, 1, 2, 3], 
              'bbb': [0.0, 1.1, 2.2, 3.3]
             })
df.loc[df['aaa'] == a]
# No result

df.loc[df['aaa'] == int(a)]
# Works

df.loc[df['bbb'] == a]
# No result

df.loc[df['bbb'] == float(a)]
# Works
```

How this could work (i'm not sure if it's a good idea / if there can be negative side-effects) is by inheriting from int / float.
Changing the above class to the following works fine for int tests above - and the corresponding alternative method works for float.

```python
class Xxx(int):
    def __int__(self):
        print("Int called")
        return 2
    
    def __float__(self):
        print("float called")
        return 2.2
```

freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+    Defines a parameter that can be optimized by hyperopt.
+    """
+    category: Optional[str]
+    default: Any
+    value: Any
+    opt_range: Sequence[Any]
+
+    def __init__(self, *, opt_range: Sequence[Any], default: Any, space: Optional[str] = None,
+                 enabled: bool = True, **kwargs):
```
xmatthias: I'm not yet sure about the meaning of `enabled` yet (i actually only noticed it a few minutes ago)

at first, it seems good that a parameter can be excluded from hyperopt (so the paramter remains available, but is not hyperopted)

However, it's also got the side-effect that it'll not load the parameter from the params dictionary, but instead use the default (which was surprising to me)

rokups:
That is on purpose. I found it very useful when prototyping. Disabling 
parameter allows to tweak value manually and not have to comment it out in 
buy params. Maybe name sucks though. "manual" might make more sense instead 
of "enabled".

freqtrade/strategy/hyper.py
xmatthias: as said, i'm not sure about it yet - maybe it should be split into 2 settings?

on the one hand, manually tuning is great (prevent hyperopt for this parameter) - but i found it surprising that the default will be used in that case, not the value in buy_params (which is usually the result of an earlier hyperopt run)...

rokups: Hmm my email response did not get into this particular thread. Anyway yeah i was thinking about two options as well. `optimize` + `load` maybe? Both would be enabled by default and user may achieve any kind of desired combination (load from config but do not optimize, optimize but do not load (probably least useful), load and optimize (default), neither load nor optimize (for manual tuning))

xmatthias: yeah, i noticed that (but intentionally answered here again, not below to keep it "kindof" organized).

Your idea that sounds sensible / reasonable 👍

rokups:
I also spotted [this kind of usage](https://github.com/Rikj000/MoniGoMani/blob/main/user_data/strategies/MoniGoMani.py#L66-L161) of parameters. I wonder if we could do anything to accommodate it. This kind of grouping would also be useful for per-pair settings. I am not sure how though.

xmatthias:
> I also spotted [this kind of usage](https://github.com/Rikj000/MoniGoMani/blob/main/user_data/strategies/MoniGoMani.py#L66-L161) of parameters. I wonder if we could do anything to accommodate it. This kind of grouping would also be useful for per-pair settings. I am not sure how though.

well technically, it'll be 2 different parameters for each "trend-version" (not related at all ...)

so i think the current proposal here will satisfy this as follows (reduced sample)

``` python
adx_strong_up_buy_weight1 = IntParameter([0, 50], default=20, space='buy')
adx_strong_up_buy_weight2 = IntParameter([0, 50], default=40, space='buy')
trend = {
    'downwards': {
        # 'adx_strong_up_buy_weight': IntParameter([0, 50], default=20, space='buy') ,
        'adx_strong_up_buy_weight': adx_strong_up_buy_weight1 ,
    },
   'upwards': {
#        'adx_strong_up_buy_weight': IntParameter([0, 50], default=40, ),
        'adx_strong_up_buy_weight': adx_strong_up_buy_weight2 ,

    },
}

# usage:
self.trend['downwards']['adx_strong_up_buy_weight'].value
```

The only doubt i have about the "automatic" naming - which will probably fail in such a case (will need testing) - therefore i extracted it to "top level" values.

On the other hand - i don't think we must support any usage pattern. 
This is the first time i see parameters / settings used like this - and it's not documented anywhere, so if it doesn't work with the new parameters, then the "usage pattern" simply must change in order to be able to use the Parameter approach correctly.

I don't think we should try iterate every dictionary to cover such edge usecases - it'll be quite dangerous, as you won't know what a dictionary may contain ... it could also be done as a list - so we would have to go down every single defined attribute that can contain children ... which i see as not necessary. 

---

in short: Parameters must be defined at strategy "class level" (be class variables directly) - otherwise parameter will not be considered.

rokups:
This seems like a reasonable solution. Iterating everything is definitely overkill. Maybe using nested classes would also work.

```py
class MyStrategy(IHyperStrategy):
    @parameter_group
    class Uptrend:
        adx_buy_weight = IntParameter([0, 50], default=20, space='buy')

    @parameter_group
    class Downtrend:
        adx_buy_weight = IntParameter([0, 50], default=20, space='buy')
```

Final parameter name could have group class name prepended to differentiate them.

xmatthias:
Honestly, as a first iteration i wouldn't make this too complicated.

even supporting (and documenting!) it in a "flat" structure will be difficult enough so inexperienced users understand how it'll work.
Introducing "groups" might be done at a later step if we see this is really necessary (i don't think it is)

The problem i have with the "parameter-group" is that it'll spark the assumption that this will generate a hierarchical space - which scikit-optimize does not know / support.

What i mean with this:
assuming you have 2 parameters, "rsi_enable" and "rsi_value" ... if rsi_enable is False, changing the rsi_value does not make sense as it's not used - but there is no way to make scikit-optimize consider this if both parameters are to be optimized, and it will blindly make 20 epochs changin rsi_value without ever changing rsi_enabled - which will result in no change whatsoever in the result...

xmatthias:
i think one big task is still up ahead - which is documentation (otherwise it'll only be the 2 of us [and silent observers? :eyes: ] who will be able to use this ....).

I'm not quite sure yet how to approach the documentation part though...

Either we leave the existing hyperopt documentation untouched, and show this as a new, optional way in a subsection (which can potentially be confusing - as it might be unclear if this will be on top of the other method, or whatever ...).

The alternative would be to promote this as the "new default" - move parts of the existing documentation to the advanced-hyperopt section (or remove if they are irrelevant - but the existing hyperopt methodology will still continue to work, so probably there's not much to remove) and basing the default on top of this - which i hope will make it easier for beginners to grasp.

personally, i'm leaning more towards the 2nd approach - but i'd also like to know your opinion

freqtrade/strategy/hyper.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def get_space(self, name: str) -> 'Integer':
+        """
+        Create skopt optimization space.
+        :param name: A name of parameter field.
+        """
+        return Integer(*self.opt_range, name=name, **self._space_params)
+
+
+class FloatParameter(BaseParameter):
```
xmatthias: i keep asking myself if we should add another parameter type with limited decimals.

The FloatParameter is currently using the `Real` space - which has a (in theory) unlimited precision (i guess it'll be 16 decimals ...).

This is a problem in it's own right and can lead to a "vanishing gradient" problem - where the changes hyperopt performs on a parameter are too small to be visible.  

e.g. if you compare something to 30.1111 or 30.11125123123123 - will not make your results better or worse - only overfitted.

So my idea would be:
Rename the "existing" space to `RealParameter` (aligned with the `Real` space)

Add a parameter to `FloatParameter` called decimals (defaults to 3?) which will:
`pow` the values by decimals for the space and uses the `Integer` space
`unpow` the values in a `value` property

so for example the following space 
`FloatParameter([0, 1], decimals=2)` would only have the following possible values (100 in total):
`[0.01, 0.02, 0.03, ...0.98, 0.99]`

rokups: Ohh that is a good idea. I'll do it 👍🏻

rokups:
Making this a new default would be awesome, but i am quite biased :] Although so far i was getting only positive feedback from people i showed it to. Everyone is so hyped they do not need to duplicate work. We could deprecate current way and keep old documentation in a subsection maybe, give people some time to move on.

xmatthias:
> Making this a new default would be awesome, but i am quite biased :] Although so far i was getting only positive feedback from people i showed it to. Everyone is so hyped they do not need to duplicate work. We could deprecate current way and keep old documentation in a subsection maybe, give people some time to move on.

Well your bias might be my bias now ... i had ... similar ideas 😆 

Let's then move the docs for the existing method to the bottom of advanced-hyperopt for now (maybe with a comment that we recommend the other method now) - and adapt what needs changing to the new method 👍

GuuhBoss:
Probably is the version, but my files are quitly different tho , just an observation. Great work!

freqtrade/strategy/hyper.py
```diff
@@ -244,8 +244,8 @@ def enumerate_parameters(self, category: str = None) -> Iterator[Tuple[str, Base
             if not attr_name.startswith('__'):  # Ignore internals, not strictly necessary.
                 attr = getattr(self, attr_name)
                 if issubclass(attr.__class__, BaseParameter):
-                    if category is None or category == attr.category or \
-                       attr_name.startswith(category + '_'):
+                    if (category is None or category == attr.category or
+                            (attr_name.startswith(category + '_') and attr.category is None)):
```
rokups, line 252: I think this should be an error. People will do a stupid thing and not realize why their parameter is silently excluded from optimization.

xmatthias: i would almost tend to leave it this way for now and eventually implement this at a later point IF we see it's a probem.

i don't expect people to actually use both space and naming convention, but mainly rely on naming convention ...

rokups: Right, we do not expect them, but they eventually will. You are case and point :]

xmatthias: It's failing now in this case ;)

xmatthias (approved):
LGTM :+1: 
I know it's functional as is, and we can eventually adapt in the future should we notice problems in certain areas.

I hope this will be a gamechanger and simplify hyperopt for almost all usecases.
Thanks a lot for this great contribution and inspiration! :100:

PR #4601, opened 2021-03-25
Specify which versions of Python Freqtrade successfully builds on.

kjkraw:
## Summary
This is just a small change to the docs to specify the versions that Freqtrade builds on.

At this time, a wheel for the "tables" dependency is not available for python 3.9 on pypi. This prevents a user from installing freqtrade with python 3.9.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -38,7 +38,7 @@ These requirements apply to both [Script Installation](#script-installation) and
 
 ### Install guide
 
-* [Python >= 3.7.x](http://docs.python-guide.org/en/latest/starting/installation/)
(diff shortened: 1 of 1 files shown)


xmatthias:
This is not correct information for multiple reasons, so i don't think i will accept this.

Python 3.9 works perfectly fine, freqtrade works (and is tested) on versions between 3.7 and 3.9. 

If a wheel is missing for a dependency on one platform, then this does not lower the supported version, as the dependency can still be installed without wheels (it'll just need the build packages that package defines).
It may need some research - but it's still not a problem i see with freqtrade.

Tables does also provide wheels for 3.9, just not for windows / macos.

The best approach is to raise an issue with the affected dependency.

Issue #4412, opened 2021-02-21
download-data does not stop downloading at the specified TIMERANGE end date


brookmiles:
## Describe your environment

  * Operating system: Microsoft Windows [Version 10.0.19042.804]
  * Python Version: 3.8.6
  * CCXT version: 1.40.99
  * Freqtrade Version: docker-1310a7b5
  
## Describe the problem:

download-data does not stop downloading at the TIMERANGE end date specified on the command line, and continues to download all data until the present.  

### Steps to reproduce:

  1. set up config.json for Kraken with recommended rate limits and download_trades enabled
  2. whitelist DOGE/USD (or any pair with high volume and historical data as far back as 20200101)
  3. run download-data and specify a timerange of a single day, 20200101-20200102: `docker-compose run --rm freqtrade download-data --timerange 20200101-20200102 -t 1m 5m 1h` 
  
### Observed Results:

  * What happened?

Although a timerange with an end date is specified, freqtrade continues to download all trades up to the present, this takes almost two hours with DOGE/USD (once i finally realized what was going on, and let it complete)

  * What did you expect to happen?

I tried this multiple times, and each time concluded that the download had failed or gotten stuck, because downloading the most recent few days takes only minutes, and I was specifying a timerange of a similar length, which I expected to take a similar amount of time.  Often there would be one or more API warning messages, further indicating to me that something about the process had failed.

No indication is given that the end date of the timerange is being ignored.  Indeed, the full timerange is logged by default, including the end date, further leading me to believe it would be honoured: `Parameter --timerange detected: 20200101-20200102`

Only when debug logging is enabled can you see that:
* debug line indicating that the _async_get_trade_history() `until` parameter is None, 
* the logged fetch `since` times continue past the end date, once the logged values are converted to human readable dates.
* (if the process is allowed to complete) debug logged `New End:` is today, and not the requested date

While this doesn't seem to matter much for exchanges that provide OHLCV data and can complete this entire process within seconds, it makes it much less practical to download Kraken data, as it requires downloading the entire history of many hours, instead of a partial history in minutes, which would be useful for either targeted backtesting, or downloading the entire history in stages, by week or month, instead of having a single, many hours or days long, process susceptible to network disconnects or API failures.

### Relevant code exceptions or logs

```
docker-compose run --rm freqtrade download-data --config user_data/config-kraken-download.json --timerange 20200101-20200102 -t 1m 5m 1h -vv
Creating ft_data_freqtrade_run ... done
[...]

PR #4603, opened 2021-03-25
Implementing end time in data-download

kjkraw:
## Summary
The goal of this PR is to implement end times in data downloading functions so that timeranges with the data-download command work as expected.

Solve the issue: #4412 

This is a work-in-progress change; however, I wanted to draft a PR in order to get some feedback on this.

## Current Changes
- Added `until` parameters to functions for downloading OHLCV data
- Ensure that downloading stops once the `until` time has been reached
- Added a `to_datetime` method to the `TimeRange` object

## Planned Changes
- Implement `until` on functions for downloading trade data
- Update unit tests
- ~Add parameter `--fill-gaps` to prevent gaps in data which would occur in certain cases~
- ~Add parameter `--no-overwrite` to prevent new downloads from overwriting already saved data~

## Potential Changes
- ~Replace `--timerange` parameter with parameters `from` and `to`~

### DIFF
--- a/freqtrade/data/history/history_utils.py
+++ b/freqtrade/data/history/history_utils.py
@@ -287,9 +287,12 @@ def refresh_backtest_ohlcv_data(exchange: Exchange, pairs: List[str], timeframes
                         f'Deleting existing data for pair {pair}, interval {timeframe}.')
 
             logger.info(f'Downloading pair {pair}, interval {timeframe}.')
+            # What happens when until should be None?
+            since, until = timerange.to_datetime()
             _download_pair_history(datadir=datadir, exchange=exchange,
                                    pair=pair, timeframe=str(timeframe),
-                                   timerange=timerange, data_handler=data_handler)
+                                   since=since, until=until,
+                                   data_handler=data_handler)
     return pairs_not_available
 
 
--- a/freqtrade/data/history/history_utils.py
+++ b/freqtrade/data/history/history_utils.py
@@ -184,7 +184,7 @@ def _download_pair_history(datadir: Path, exchange: Exchange, pair: str, *,
                      if cached_start else 'None')
         logger.debug("Cached End: %s",
                      f"{cached.iloc[-1]['date']:%Y-%m-%d %H:%M:%S}"
-                     if not cached_end else 'None')
+                     if cached_end else 'None')
 
         # Set the bounds for downloading
         since_ms, until_ms = None, None
--- a/freqtrade/data/history/history_utils.py
+++ b/freqtrade/data/history/history_utils.py
@@ -196,9 +196,6 @@ def _download_pair_history(datadir: Path, exchange: Exchange, pair: str, *,
                 if since < cached_start:
                     since_ms = since.timestamp() * 1000
                 elif cached_start <= since < cached_end:
-                    logger.warning("The timerange overlaps with cached data."
-                                   " This may lead to unexpected outcomes "
-                                   "including overwriting existing data!")
                     since_ms = since.timestamp() * 1000
                 else:
                     since_ms = cached_end.timestamp() * 1000
@@ -208,9 +205,6 @@ def _download_pair_history(datadir: Path, exchange: Exchange, pair: str, *,
                 if until < cached_start:
                     until_ms = cached_start.timestamp() * 1000
                 elif cached_start < until <= cached_end:
-                    logger.warning("The timerange overlaps with cached data."
-                                   " This may lead to unexpected outcomes "
-                                   "including overwriting existing data!")
                     until_ms = until.timestamp() * 1000
                 else:
                     until_ms = until.timestamp()
--- a/freqtrade/data/history/history_utils.py
+++ b/freqtrade/data/history/history_utils.py
@@ -176,8 +176,8 @@ def _download_pair_history(datadir: Path, exchange: Exchange, pair: str, *,
 
         cached_start, cached_end = None, None
         if not cached.empty:
-            cached_start = cached.iloc[0]['date'].timestamp()
-            cached_end = cached.iloc[-1]['date'].timestamp()
+            cached_start = int(cached.iloc[0]['date'].timestamp())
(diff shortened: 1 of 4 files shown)


xmatthias:
I would leave some of the "planned" changes for future PR's, especially the last 2 points in your list. They are not related to the "end time" - and having too many (only remotely related) changes in one PR makes it very difficult to review properly.


the "replacing of `--timerange`" would be a breaking change (and honestly, i don't see the benefit of using `--from 20200101 --until 20210101` over `--timerange 20200101-20210101`) - so if at all, it must be optional (but i think it'll make the interface complex for new users without providing real benefits).

kjkraw:
Some clarifications for this PR:

Full implementation of  `--timerange` may lead to some problems if there is existing data saved. The first issue occurs when `end_time < cache_start` or `start_time > cache_end`, leading to time gaps, differences in the bounds of downloads which would lead to missing data. The second issue occurs when `cache_start <= start_time  < cache_end` or `cache_start < end_time <= cache_end`, which would lead to overwrites.

Currently, my planned fix for the first issue is to set bounds such that there would be no gap. Effectively: `if end_time < cache_start: end_time = cache_start` and `if start_time > cache_end: start_time = cache_end`. By setting the bounds like this, this would prevent periods of missing data; however, it would also ignore the user's input. In order to allow users to have datasets with missing data, I'm proposing the addition of a new parameter `--allow-missing-data` or something like that. By default, `--allow-missing-data` would be false, and I write into the documentation that the setting should be avoided unless the user really knows what they are doing with it. Nevertheless, I think the feature would be useful for very targeted backtesting.

For the second issue, I see two options:
1. I can allow the overwrite to happen.
2. When the new download would overwrite existing data, the bounds are changed in order to prevent the overwriting -- this may also lead to shorter download times.

xmatthias:
> Nevertheless, I think the feature would be useful for very targeted backtesting.

You can use `--timerange` on backtesting to specify lower ranges of data and do "targeted" backtesting.

I think it'll be very confusing (and potentially dangerous) if you donwnload All data for January and March, but you leave february out.
Maybe today the user will know and only backtest these 2 periods seperately - but in the end, they'll run January-March - causing bad / invalid backtest data as a full month was missing.

I can therefore currently se no reason to allow this "gap" - as even for experienced / expert users - it's better to have all data available even if it's not used for backesting.

---

For the 2nd issue i'm not sure - currently, we're overwriting all data. In theory, stopping at "data start" could work - HOWEVER, what would we do in the following case?

* `download-data --timerange 20200401-20200801`
* `download-data --timerange 20200101-20201201`

if we don't want to redownload 04-08, we'd have to do the download in 2 steps - first 01-04, and then 08-12. While not necessarily a problem, i'm not 100% sure if it's worth the effort and mainly code complexity to handle this case without redownloading.

kjkraw:
> I can therefore currently se no reason to allow this "gap" - as even for experienced / expert users - it's better to have all data available even if it's not used for backesting.

You've convinced me. I'll make sure that the system automatically sets safe bounds.

>While not necessarily a problem, i'm not 100% sure if it's worth the effort and mainly code complexity to handle this case without redownloading.

I'll look into this and see how big of a difference it makes. Finding the new bounds -- for your example: [01,04) and [08,12] -- shouldn't be too hard, but having to join three dataframes instead of two may be overly complicated. Additionally, having multiple downloads will lead to two separate batches of async downloading, which doesn't fully make use of the extra speed that comes with async.

xmatthias:
> Additionally, having multiple downloads will lead to two separate batches of async downloading, which doesn't fully make use of the extra speed that comes with async.

that's not necessarily true considering that also async respects (has to!) the exchange rate limits - so also in async mode, ccxt will make sure not to "spam" the exchange with 1000 requests at once, but throttles them to  fit into the configured ratelimit (for binance, that's 200ms - so ~5 calls every second).

While it'll be slightly slower for sure - it'll depend on the timeframe you'll look at if it makes sense at all.
Downloading a year of 1h candles is done in seconds ... while for 1m candles, you can look at quite some time ...

freqtrade/data/history/history_utils.py
```diff
@@ -152,71 +152,90 @@ def _load_cached_data_for_updating(pair: str, timeframe: str, timerange: Optiona
...
+                                    fill_missing=False)
+
+        if cached.empty:
+            data = new_df
         else:
-            # Run cleaning again to ensure there were no duplicate candles
-            # Especially between existing and new data.
-            data = clean_ohlcv_dataframe(data.append(new_dataframe), timeframe, pair,
-                                         fill_missing=False, drop_incomplete=False)
+            data = clean_ohlcv_dataframe(cached.append(new_df), timeframe,
```
xmatthias, line 223: I would keep the comment we had before this line - as it makes it clear what `clean_ohlcv_dataframe()` does (i had to look at that function to see that it'll remove duplicates)...

freqtrade/data/history/history_utils.py
```diff
@@ -152,71 +152,90 @@ def _load_cached_data_for_updating(pair: str, timeframe: str, timerange: Optiona
...
+
         logger.info(
-            f'Download history data for pair: "{pair}", timeframe: {timeframe} '
-            f'and store in {datadir}.'
+            f'Downloading history data for pair: "{pair}", timeframe: '
+            f'{timeframe} and storing in "{datadir}"'
         )
 
-        # data, since_ms = _load_cached_data_for_updating_old(datadir, pair, timeframe, timerange)
-        data, since_ms = _load_cached_data_for_updating(pair, timeframe, timerange,
```
xmatthias, line 183: I think the loading of existing data should still be left in this method - just that the method now should return both start and end dates (on top of the data, obviously).

That'll make it easier to test and make this method less ... complicated / long.

kjkraw:
I began this PR during my spring break. I am back in school, and I do not currently have time to continue working on this. I may return to this in a few months when my summer break begins if this feature has not been added by someone else. If anyone would like to continue this PR, please feel free to build off of the modifications I've already made.

PR #4606, opened 2021-03-26
Add days breakdown table to backtesting

rextea:
## Summary & What's new:
1) Add days breakdown table to backtesting: 
Screen_Shot_2021-03-26_at_16 32 48

I made it optional, can be shown when using this parameter when backtesting:
--show-days

2) Add average daily profit % to SUMMARY METRICS


freqtrade/optimize/optimize_reports.py
```diff
@@ -191,6 +190,29 @@ def generate_edge_table(results: dict) -> str:
                     floatfmt=floatfmt, tablefmt="orgtbl", stralign="right")  # type: ignore
 
 
+def generate_days_breakdown_stats(results: DataFrame, starting_balance: int) -> Dict[str, Any]:
+    days = results.resample('1d', on='close_date')
+    days_stats = []
+    for name, day in days:
+        profit_abs = day['profit_abs'].sum().round(10)
+        profit_total = day['profit_abs'].sum() / starting_balance
```
xmatthias: i'm not certain this is the correct way to calculate this.

for the overal profit - yes, it'll work and is correct ... 
but assuming you enable compounding (unlimited stake_amount) - then to show the correct daily profit, you'll need to keep track of the balance.

e.g. you start with 1000$ - and trade for 20 days, making 100% profit over this period.
Now your starting point is 2000$ - but if you now make 200$ in one day, that's not 20% (calculated against the 1000$) - but it's actually 10% (calculated against the 2000$ you had the day before)

This is however complicted by the fact that we're talking multiple parallel trades which can span over multiple days ... so getting this correct is ... not quite easy.

rextea: Hey, i've been a bit busy lately, but I want to finish this PR :)
do you have an idea how to make the compounding work with it?
or we can make it only relative percentages for now?

xmatthias: i guess you should be able to use a cumsum over absolute profit to get the "end day" balance - and then use the prior day's balance as base for this calculation.

A potentially easier solution is to omit relative values for now, and only report absolute (USD) values ...

joshdifabio: > A potentially easier solution is to omit relative values for now, and only report absolute (USD) values ...

I agree this would be worthwhile initially; daily dollar profits would provide most of the value.

xmatthias (approved):
Great idea! :100: 
LGMT  :+1: 

I've slightly modified it to work more flexibly (to potentially include month and week breakdowns).

PR #4614, opened 2021-03-28
Unwanted exchanges error

mablue:
I added an script to check if exchange not supported by freqtrade.

## Summary

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -67,8 +67,15 @@ class FreqtradeBot(LoggingMixin):
 
         # Check config consistency here since strategies can set certain options
         validate_config_consistency(config)
+        
+        unwanted_exchanges = ['coinex']
+        exchange_name = self.config['exchange']['name']
+        
+        if exchange_name in unwanted_exchanges:
+          logger.error(f'Exchange, {exchange_name} not supported by freqtrade! continue with your own risk!')
+            
 
-        self.exchange = ExchangeResolver.load_exchange(self.config['exchange']['name'], self.config)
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -67,8 +67,15 @@ def __init__(self, config: Dict[str, Any]) -> None:
...
         # Check config consistency here since strategies can set certain options
         validate_config_consistency(config)
-
-        self.exchange = ExchangeResolver.load_exchange(self.config['exchange']['name'], self.config)
+        
+        unwanted_exchanges = ['coinex']
+        exchange_name = self.config['exchange']['name']
+        
+        if exchange_name in unwanted_exchanges:
+          logger.error(f'Exchange, {exchange_name} not supported by freqtrade! continue with your own risk!')
```
xmatthias, line 75: If an exchange is REALLY not working (confirmed not working) - then please add it to the `BAD_EXCHANGES` Dict in `freqtrade/exchange/common.py` with the reason for it - which will have the same effect.

From the issue you posted, i'm not yet convinced about that though... as it seems like some orders have at least been added.

xmatthias (changes requested):

xmatthias:
I think i'll close this for now, as this is 
a) not the right place to add this
b) it seems like coinEx works for you after all (based on comments in other issues).

mablue:
Yes its works for me some times when the bot is off and i cancele manually an order.and re run freqtrade will go to the loop of order not found!

PR #4646, opened 2021-04-02
Add profit_fiat to REST API response on status

mads03dk:
## Summary
Adding a `profit_fiat` field on `status` within the REST API (Just like the last column `/status table` on telegram).
## Quick changelog

- Added `profit_fiat` to `_rpc_trade_status` function, and in schema.
- "Added to tests"

## What's new?
I wanted to be able to see the current profit in fiat value on the REST API. Just like the last column when using `/status table` on telegram.


xmatthias (approved):
LGTM 👍 

would kinda be great if you could also add this to freqUI - but don't worry if that's not in your interrest / knowledge space, then i'll take care of that.

mads03dk:
> would kinda be great if you could also add this to freqUI - but don't worry if that's not in your interrest / knowledge space, then i'll take care of that.

That was exactly my intention :)

Issue #4654, opened 2021-04-03
have not enough space for profit field!


mablue:
## My environment:

  * Operating system: Linux mab-X550VX 5.12.0-051200rc5daily20210402-lowlatency #202104012206 SMP PREEMPT Fri Apr 2 02:49:26 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
  * Python Version: Python 2.7.18
  * CCXT version: ccxt==1.45.44
  * Freqtrade Version: freqtrade develop-00d8f3a4

## The problem:

1) have not enough space for profit field in hyperopt results
2) Is this real? If yes, I will be a reach man after 1 mounth! :D
### Observed Results:
```
+--------+-----------+----------+------------------+--------------+-------------------------------+----------------+-------------+                                                                                         
|   Best |     Epoch |   Trades |    Win Draw Loss |   Avg profit |                        Profit |   Avg duration |   Objective |
|--------+-----------+----------+------------------+--------------+-------------------------------+----------------+-------------|
| * Best |    2/1000 |       12 |      3    4    5 |       -1.07% |  -47.44376800 USDT  (-12.81%) |        650.0 m |     2.73358 |
| * Best |    3/1000 |       45 |     18    1   26 |        0.85% |  141.48183841 USDT   (38.20%) |        329.3 m |    -3.21775 |
| * Best |    8/1000 |      279 |    151   81   47 |        4.91% | 5,078.13619263 USDT (1,371.10%) |      2,290.5 m |    -57.4013 |
| * Best |    9/1000 |      723 |    333   17  373 |        1.33% | 3,556.15266428 USDT  (960.16%) |        815.7 m |    -76.0623 |
| * Best |   16/1000 |     2410 |   1665   66  679 |        1.34% | 11,954.11413026 USDT (3,227.61%) |        383.9 m |    -338.614 |                                                                                      
|   Best |   54/1000 |     1276 |   1028  187   61 |        2.60% | 12,288.71973612 USDT (3,317.96%) |        906.3 m |    -348.458 |                                                                                      
|   Best |  158/1000 |     1277 |   1027  151   99 |        3.58% | 16,926.48323613 USDT (4,570.16%) |      1,018.6 m |    -373.872 |                                                                                      
|   Best |  186/1000 |     4538 |   3371   25 1142 |        1.19% | 19,970.90026418 USDT (5,392.15%) |        287.3 m |    -658.312 |                                                                                      
[...]

PR #4656, opened 2021-04-03
update config md -  max_open_trades=-1

mablue:
## Summary

Warn user the issue:  #4654

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -191,6 +191,9 @@ For example, if you have 10 ETH available in your wallet on the exchange and `tr
 Assuming we have the tradable balance of 1000 USDT, `stake_amount=400`, and `max_open_trades=3`.
 The bot would open 2 trades, and will be unable to fill the last trading slot, since the requested 400 USDT are no longer available, since 800 USDT are already tied in other trades.
 
+!!! Warning
+    Unlimited `max_open_trades` count (max_open_trades: -1) will cause unreal results in backtest and hyperopt high iterations because of unfulfilled order list in real exchanges
+
 To overcome this, the option `amend_last_stake_amount` can be set to `True`, which will enable the bot to reduce stake_amount to the available balance in order to fill the last trade slot.
 
 In the example above this would mean:


xmatthias:
This is simply an untrue/wrong statement, and will be missleading for people reading it.
`max_open_trades` has nothing to do with realistic or unrealistic results, it'll simply mean that ALL available balance will be used in "stake_amount" segments.


Order filling is not considered in backtesting / hyperopt - as it's also documented in the [backtesting assumptions](https://www.freqtrade.io/en/latest/backtesting/#assumptions-made-by-backtesting).

mablue:
> This is simply an untrue/wrong statement, and will be missleading for people reading it.
> `max_open_trades` has nothing to do with realistic or unrealistic results, it'll simply mean that ALL available balance will be used in "stake_amount" segments.
> 
> Order filling is not considered in backtesting / hyperopt - as it's also documented in the [backtesting assumptions](https://www.freqtrade.io/en/latest/backtesting/#assumptions-made-by-backtesting).

Big problem is that: when I set it to -1 and stake amount to static number about 10% of my full usdt balance and start a hyperopt with SharpeHyperOptLossDaily stoploss option I didnt get freezed but results is unreal and objective not show a true number never!
And if i change the stop loss option to other thigs my 6700hq cpu with 8gb ram and any stable linux kernel (i tested old and latest one)that have ubuntu desktop will freezing after some moments of processing.and i will force shotdown my laptop!!!:))
It will be warn the user about this problems somewere.its dangers realy its a kernel bug report i think or ubuntu desktop bug report(cuz i chaked without gui and it works but result is not good and get more time to calculate iterations)
Anyway I love freqtrade as a project that can do magic.wow wow wooooooww thank you for this free openSource project.Im working somewhere hard but I realy neeed a time to read full of documentation and push my strategys to this project.💘💘💘💘💘💘💘💘💘💌💌💘💘💌💌💘💖💗💗💝💝💓💞💟❣️

xmatthias:
Please try to understand the problem you're having before trying to add warnings to people about things you don't fully understand just yet.

The slowdown of the pc is expected during hyperopt, as it consumes a lot of memory / cpu power (which can be limited with `-j`). It has however nothing to do with max-open trade setting directly - but will be caused by the timerange / pairlist  / timeframe combination youre using.

Obviously, if your strategy does 10000 of trades per epoch, the slowdown / memory consumption will be worse - however your warning is warning people from the wrong thing, i'll therefore not accept this PR and close it now.

PR #4659, opened 2021-04-03
Add "cfg" argument for overriding any config from command line

Regul777:
## Summary
Add universal argument "--cfg" for overriding any config value

## Common Syntax:
```
--cfg key=value [key.subkey.subsubkey...:type=value]
```

## Usage and Examples
Now you have you create a new argument configuration if you want to set some new config from command line. With this small improvement you can easily override any config value. 

```
freqtrade trade --cfg timeframe=15m stake_amount:float=0.1 exchange.sandbox:bool=true
```

Because all data in arguments is string you must specify the type of a value (default: string). Also you can use "path" of the key with "." to set the value at any level of the config.

Now we support only 4 types:
```
key:str=value - String
key:int=1
key:float=0.99
key:bool=0|1|false|true|False|True
```

### DIFF
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -1,6 +1,7 @@
 """
 Definition of cli arguments used in arguments.py
 """
+import argparse
 from argparse import ArgumentTypeError
 
 from freqtrade import __version__, constants
@@ -36,6 +37,50 @@ class Arg:
     def __init__(self, *args, **kwargs):
         self.cli = args
         self.kwargs = kwargs
+
+
+class SetDictFromArgAction(argparse.Action):
+    """
+    argparse action to split an argument into KEY=VALUE form
+    on the first = and append to a dictionary.
+    """
+
+    def __call__(self, parser, args, values, option_string=None):
+        for key_value in values:
+            try:
+                (k, v) = key_value.split("=", 2)
+            except ValueError as ex:
+                raise argparse.ArgumentError(self, f"could not parse argument \"{values[0]}\" as k=v format")
+            d = self.set_dict_path_value(getattr(args, self.dest) or {}, k, v)
+            setattr(args, self.dest, d)
+
+    @staticmethod
+    def set_dict_path_value(obj, key, value):
+        root_obj = obj
+        keys = key.split('.')
+        latest = keys.pop()
+        for k in keys:
+            obj[k] = obj.get(k, {})
+            obj = obj[k]
+        latest_key, *value_type = latest.split(':')
+        value_type = value_type[0] if len(value_type) > 0 else None
+        if value_type is None or value_type == "str":
+            value = value
+        elif value_type == "int":
+            value = int(value)
+        elif value_type == "float":
+            value = float(value)
+        elif value_type == "bool":
+            if value == "True" or value == "true" or value == "1":
+                value = True
+            elif value == "False" or value == "false" or value == "0":
+                value = False
+            else:
+                raise ArgumentTypeError(f"Argument '{key}' has unknown value '{value}'. Must be (true|false|True|False|1|0)")
+        else:
+            raise Exception(f"Unsupported arg type '{value_type}'")
+        obj[latest_key] = value
+        return root_obj
 
 
 # List of available command line options
@@ -529,5 +574,13 @@ AVAILABLE_CLI_OPTIONS = {
         '--no-header',
         help='Do not print epoch details header.',
         action='store_true',
+    ),
+    "cfg": Arg(
+        '--cfg',
+        nargs='+',
+        metavar="KEY=VALUE",
+        action=SetDictFromArgAction,
+        help='Set any config value.',
+        # action='store_const'
     ),
 }
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -49,8 +49,10 @@ class SetDictFromArgAction(argparse.Action):
         for key_value in values:
             try:
                 (k, v) = key_value.split("=", 2)
-            except ValueError as ex:
-                raise argparse.ArgumentError(self, f"could not parse argument \"{values[0]}\" as k=v format")
+            except ValueError:
+                raise argparse.ArgumentError(
+                    self,
+                    f"could not parse argument \"{values[0]}\" as k=v format")
             d = self.set_dict_path_value(getattr(args, self.dest) or {}, k, v)
             setattr(args, self.dest, d)
 
@@ -76,7 +78,9 @@ class SetDictFromArgAction(argparse.Action):
             elif value == "False" or value == "false" or value == "0":
                 value = False
             else:
-                raise ArgumentTypeError(f"Argument '{key}' has unknown value '{value}'. Must be (true|false|True|False|1|0)")
+                raise ArgumentTypeError(
+                    f"Argument '{key}' has unknown value '{value}'. "
+                    f"Must be (true|false|True|False|1|0)")
         else:
             raise Exception(f"Unsupported arg type '{value_type}'")
         obj[latest_key] = value
--- a/tests/commands/test_commands.py
+++ b/tests/commands/test_commands.py
@@ -35,6 +35,30 @@ def test_setup_utils_configuration():
     assert config['exchange']['secret'] == ''
 
 
+def test_cfg_argument_replaces_configuration():
+    args = [
+        'list-exchanges', '--config', 'config_bittrex.json.example',
+    ]
+
+    # Check original config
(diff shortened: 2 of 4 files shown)


xmatthias:
I'm not convinced on this mode (not about how it's done, but if it's needed) - and it'll make freqtrade configurations / settings a LOT more complex.

Currently, some configuration settings can come from (in this order):
* explicit command line
* configuration files (you can have multiple configurations in series)
* strategy (for config options that make sense)
* fallback / default

By adding this option, this whole logic would become more complex to understand.

consider the following case - `... -s SampleStrategy --cfg  strategy=AnotherStrategy`.
which strategy will now be loaded `SampleStrategy` or `AnotherStrategy` (assuming both exist)? 
both are provided as configuration option - so both can be considered the "users current wish".

it's unclear to me which will be used - but worse it'll also be unclear to every user trying to use it in the above way.

---

As said, i'm worried by the unclear usage for this (which setting will be used) - and i'm also not convinced that we need this, as you can simply define a small config (only containing what you wanna change) and load that as "last" config so it overwrites all other config options.

If any, i'd want to use environment variables as additonal possibility to define configurations (which is something that's been requested quite often).

As said in #4385 (https://github.com/freqtrade/freqtrade/issues/4385#issuecomment-794264792 to be precise)) - i see environment variables as cwith clearly defined namings (e.g. `FREQTRADE__ASK_STRATEGY__PRICE_SIDE = 'bid'`) as the cleaner solution.

Regul777:
Thank you! And very thnk you for your huge answer! I completely understand you. I had the same problem with understanding the "overriding" logic while development. And even though I must recheck what config will come to the destination code with breakpoint to be sure it was not override by "fallbacks/defaults". Those fallbacks are real pain for me :) Thant's why I wanted to offer some steps to simplify all the logic to few simple rules:
1. Everything is configured/controlled by Config that is passed to some "engine/part" through a pipeline (backtesting for example).
Config is processed by the next pipeline:
- config.json's
- `--cfg` arguments / explicit command line(they are like aliases for `--cfg`, `-s` for example)
- subsystems that can pipeline the config that that pass it to the next system (hopt/backtesting for example. I have one more that I wanted to share, but it's another question)
- destination "processor"
also, the first "problem" that I wanted to improve - it's development. How if you want to add a new config you must 1) create a new cli_option, 2) add it to arguments list for a command, 3) pass argument to config. I just wanted to skip all it for faster development: all you have to do, it's use config with override pipeline and default config object (where you can check default values)

I have to say that I really ahead of this thing in my branch, but I want to share my improvements with you back, because you shared this project with me. One of the next code that I want to share was to create a simple pipeline to pass arguments to a strategy line to control them from command_line/config.json without any complex code changes. For example, I am going to share my code where you can have something like that:

RsiSimpleStrategy.py:
```python
class RsiSimpleStrategy(IStrategy):
    rsi_timeperiod = 14

        dataframe['rsi'] = ta.RSI(dataframe, timeperiod=self.rsi_timeperiod)
        return dataframe
...
```

and be able to change `rsi_timeperiod` through config.json like:
```json
{
...
"RsiSimpleStrategy": { "rsi_timeperiod": 16 }
}

// or common for all strategies:
{
...
"strategy": { "rsi_timeperiod": 16 }
}

```

or by command line with the same way:
```
--cfg RsiSimpleStrategy.rsi_timeperiod=16
```

And about environment variables. With this code you already can pass environment vars to your config:
```
export FREQTRADE_CONFIG_EXCHANGE_SECRET=my-exchange-secret
freqtrade trade ... --cfg exchange.secret=$FREQTRADE_CONFIG_EXCHANGE_SECRET
```
It already works for any config key! :)

My goal - it's flexibility with clear simple "everything in controlled by config" way. In any way - I want to thank you for what you did! I am really appreciate it.

xmatthias:
Allowing the configuration more control over strategy parameters is a dangerous approach (and one i'm currently not willing to go down).

I understand you THINK it's a good approach - but it will lead to a lot of user problems - which are then mine / the projects to support / help.

It'll lead to users which will backtest one version - and then either not be able to reproduce them (because configuration changed core parts of the strategy settings) - or make mistakes while transfering them to the VPS they run the bot on (most users use a local machine for backtesting, and move the strategy to a remote machine for continuous running).

a "pipeline" approach is already in place - the configuration is loaded in the Configuration class (and all overrides happen in this class).
Obviously, the strategy can make changes IF the configuration didn't define certain things - but that's intentional and good this way, as some settings are highly strategy related - and should therefore not come from a strategy.

> And about environment variables. With this code you already can pass environment vars to your config:

this is an overly complicated approach which will fail most of our users (which are often not technically versed).


---

Your approach will also fail if i misstype one of the settings (which is very easy to do in a command line switch - and will happen more often as you'll need to type the settings out for each command as it's not part of a configuration file that's saved across sessions).


---

For the future - before investing a lot of time into developing a solution, please open an issue discussing the feature first (if it is interresting for the project / makes sense). This will avoid wasting time on solutions that we will not be able to accept.

freqtrade/commands/cli_options.py
```diff
@@ -530,4 +579,12 @@ def __init__(self, *args, **kwargs):
         help='Do not print epoch details header.',
         action='store_true',
     ),
+    "cfg": Arg(
+        '--cfg',
+        nargs='+',
+        metavar="KEY=VALUE",
+        action=SetDictFromArgAction,
```
xmatthias, line 586: I'm not sure on doing this "here" / in this way.

I fear a bit that having this logic as part of argparse Argument parsing will limit how this can be used down the road.
Also, you mention the "Pipeline" approach already - which we have in place in the Configuration class - by having this type of parsing "extracted" from there, it'll split the logic up into pieces.

---

I also think this produces strange / unexpected results in some instances.
for example in the below case (2 merged configs, with --cfg in between)

```
freqtrade trade -s <whatever> --config config1.json --cfg stake_amount:float=0.1 --config json2.json
```

i'd expect the 2nd config to "overwrite" the stake-amount if it's set - which doesn't happen.
That way, it would be consistent (last setting wins) between --cfg and --config.

tests/commands/test_commands.py
```diff
@@ -35,6 +35,30 @@ def test_setup_utils_configuration():
...
+    # Check original config
+    config = setup_utils_configuration(get_args(args), RunMode.OTHER)
+    assert config['dry_run'] is True
+    assert config['exchange']['key'] == ''
+    assert config['stake_amount'] == 0.05
+    assert config['timeframe'] == '5m'
+
+    args = [
+        'list-exchanges', '--config', 'config_bittrex.json.example',
+        '--cfg', 'stake_amount:float=0.1', 'timeframe=15m', 'exchange.sandbox:bool=true',
```
xmatthias, line 52: While the syntax may seem simple for a developer / technically versed user, it'll be VERY difficult to understand and apply for non-technical users (which we have a TON - so we MUST ensure it's easy to use for them).

Also, the error-messages are very unclear and missleading.

for example: 
`--cfg stake_amount=float=0.1`  results in `freqtrade trade: error: argument --cfg: could not parse argument "stake_amount=float=0.1" as k=v format`, which however is not the right format for this key, and following the guidance of the message (`--cfg stake_amount=0.1`) i get a different error (`jsonschema.exceptions.ValidationError: '0.1' does not match 'unlimited'`).

xmatthias (changes requested):
For one, this is clearly missing documentation. without that, it'll become a "hidden" feature noone will even try to use.

---
Before working on documentation though...
I've had a look at this again, and my concerns only grow.

I don't see this format as user friendly (as mentioned above).
Having to provide the datatype will fail most non-technical users, which will cause a lot of "how to" questions - which will make this project unsupportable.

---

I also understand what you're trying to provide for the strategy in the future, but something simlilar is already in place (now) with #4596 being merged - and your (future / planned) approach will almost certainly not be compatible with that solution.

Extracting these parameters from the strategy will be problematic in my opinion (as i also mentioned in that PR in some comment) - as users will backtest something and get good results - then upload the strategy to a remote server to run the strategy, and accidentally forget to either upload the "additional" parameter file - or to pass in the right value via command line argument (which will be even easier to forget) - which will then lead to completely different results and a lot of user frustration.

I might be able to handle this additional complexity - but we MUST think about the (non-technical) users - which will struggle and therefore cause support and maintenance burden.
I fully understand that this will not concern you as you might be able to handle all of the above concerns - but as maintainer of this project, i must think about both new users and keeping the support / maintenance burden as low as possible to keep the project maintainable.

---

I'll therefore not be able to accept this in this form. If you can simplify it (without hardcoding the datatypes for each parameter) i can surely have a look at this again - but in it's current form, i see this as a maintenance burden only, and not as a feature this project will need, or will make this project better.

PR #4660, opened 2021-04-03
Sort pair lists by total profit

rextea:
## Summary
Sort backtest pair lists results by total profit %:

Screen Shot 2021-04-04 at 1 20 11

### DIFF
--- a/freqtrade/optimize/optimize_reports.py
+++ b/freqtrade/optimize/optimize_reports.py
@@ -110,6 +110,9 @@ def generate_pair_metrics(data: Dict[str, Dict], stake_currency: str, starting_b
 
         tabular_data.append(_generate_result_line(result, starting_balance, pair))
 
+    # Sort by total profit %:
+    tabular_data = sorted(tabular_data, key=lambda k: k['profit_total_abs'], reverse=True)
+
     # Append Total
     tabular_data.append(_generate_result_line(results, starting_balance, 'TOTAL'))
     return tabular_data


xmatthias (approved):
LGTM :+1:

jej:
Sorry to arrive at this point of the discussion...
I don't really understand why this table should be sorted that way or another way. This table displays past data and is just informative. For example, on my point of view, I don't mind what was the best total profit per pair, as it really depends of the market concerning the pairs for a past tested period, and does not inform about any future trends. Sometime I just look at a pair to see if the results follow my feeling regarding the market, so I prefer to have this table alphabetically sorted.
For me, an improvement on the sorting of this table should be on any column thanks to a parameter. Something like `--sort-pair-table [pair, buys, avg-profit, etc etc]`. I guess you can arrange the idea better than me concerning the naming etc.
Hope it helps. Read you.

xmatthias:
well this table never was alphabetically sorted - but in the order of the pairlist (which might have some benefits as you see the "real" order the pairlist was in - for example if you use RandomFilter.

I don't see an advantage of having this sorting dynamic though - that'll only complicate the code without providing many benefits...

jej:
Yep, I use a sorted pair list in my conf. But now I guess the output will be sorted by total profit %?

jej:
My idea was to keep the old behaviour, and if a sort is possible/needed, to leave the user decide on what column. "total profit %" is just a specific choice, not an absolute need for everybody I guess...

PR #4676, opened 2021-04-05
Standardize CRLF throughout repo

mads03dk:
Whilst creating a fresh fork, I experienced all files being instantly marked as modified by git when i launched the devcontainer. Turns out it somewhy wanted to change CRLF to LF.

This will standardize it, so we can avoid having 200+ "changed files" in a fresh devcontainer.

This works by the editor (and git) looking in the file, and obeying whatever it defines.

Example : 
image

Other example: https://github.com/microsoft/vscode-remote-release/issues/302

Solution: https://code.visualstudio.com/docs/remote/troubleshooting#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=crlf


xmatthias:
ALl line-endings are in `lf` at the moment - so i don't see why we should be needing to change this - it's your environment that's wrong - not the repository.

---

You're on windows i assume based on the problem you have - but it's a windows problem, and this PR will break this repository for ALL other environments.

By default, git for windows  CHANGES the line endings of a repository to crlf (trying to be helpful) - even if the checked-in version is lf. This will break many projects - but can be changed in the git settings for windows.

By then using this windows-based git-checkout / `git clone` in a linux-version (dev-container, in your case) - it'll show up as errors, as windows first changed all line endings - which are then showing up as errors.

Also, using crlf lineendings is wrong on all environments other than Windows (and windows can handle it too - unless you use notepad or wordpad) - so if any, then lineendings should be changed to `lf` - not to `crlf`.

personally, i'd recommend to do the git clone in WSL2 (on windows) - that way you'll avoid this problem as that handles line-endings correctly.

mads03dk:
Oh i see. Thank you for that nice explanation :)

I will close this now. Sorry for the disturbance

xmatthias:
I'm not against the PR per se - i just think the content should be different (That would solve the problem too - but in a way that won't break linux/macOS systems.

PR #4677, opened 2021-04-05
Add API endpoint for getting a specific trade

mads03dk:
## Summary
This will allow calling the endpoint `GET /api/v1/trade/<id>` to get infomation about a specific trade.

It will return the standard trade schema.

I have added it to the docs, and changed some that were ninja-copy-pasted haha.

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -182,7 +182,7 @@ count
 	Return the amount of open trades.
 
 daily
-	Return the amount of open trades.
+	Return the profits for each day, and amount of trades.
 
 delete_lock
 	Delete (disable) lock from the database.
@@ -215,7 +215,7 @@ locks
 logs
 	Show latest logs.
 
-        :param limit: Limits log messages to the last <limit> logs. No limit to get all the trades.
+        :param limit: Limits log messages to the last <limit> logs. No limit to get the entire log.
 
 pair_candles
 	Return live dataframe for <pair><timeframe>.
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -235,6 +235,9 @@ pair_history
 performance
 	Return the performance of the different coins.
 
+ping
+	simple ping
+
 plot_config
 	Return plot configuration if the strategy defines one.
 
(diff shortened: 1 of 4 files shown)


freqtrade/rpc/api_server/api_v1.py
```diff
@@ -88,6 +88,11 @@ def trades(limit: int = 0, rpc: RPC = Depends(get_rpc)):
     return rpc._rpc_trade_history(limit)
 
 
+@router.get('/trade/{tradeid}', response_model=OpenTradeSchema, tags=['info', 'trading'])
+def trade(tradeid: int = 0, rpc: RPC = Depends(get_rpc)):
+    return rpc._rpc_trade_status([tradeid])[0]
```
xmatthias: uh - i think you will want some ... error handling here - assuming you get one trade might seem okish as you pass in one tradeID - but the return will be  a list with 1 or 0 trade - and in the "no trade found" case, it'll error out with a `KeyError`.

The difference between this and the below "delete_trade" is that the errorhandling  in the delete case is within the delete method - while here it isn't.

docs/rest-api.md
```diff
@@ -275,6 +276,10 @@ trades
 
         :param limit: Limits trades to the X last trades. No limit to get all the trades.
 
+trade
+	Return specific trade.
+        :param tradeid: Specify which trade to get.
+
 version
```
xmatthias: This section is the output of `scripts/rest_client.py --show`.

I don't mind having the descriptions reworded / improved - but it should really happen in that script and then copy/pasting the output to here- not just in the docs ;) 
same applies to the new endpoint - which is also a "doc only" point - which is not really available in the script (at the moment ...).

xmatthias (approved):
I've taken the liberty to finish this PR for you.

LGTM :+1: 

Thanks for your contribution!

mads03dk:
Oh nice! I've been on a vacation, so I'm sorry for not telling you that I would've finished it when I came back

Still, thanks for implementing it :)

PR #4682, opened 2021-04-06
Sqlalchemy 14 preparations

xmatthias:
## Summary
Implement changes to sqlalchemy handling as preparation for sqlalchemy 2.0

[source](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-orm-usage)

Running the tests with warning on: 
`SQLALCHEMY_WARN_20=1 python -W always::DeprecationWarning -m pytest`

## Quick changelog

- don't use `engine.execute()` directly
- don't `.execute()` no longer taking a plain string, but requires wrapping in a "text" call

# Still missing:
- [x]  `session.autocommit`
  The Session.autocommit parameter is deprecated and will be removed in SQLAlchemy version 2.0.  The Session now features "autobegin" behavior such that the Session.begin() method may be called if a transaction has not yet been started yet.  See the section session_explicit_begin for background. (Background on SQLAlchemy 2.0 at: http://sqlalche.me/e/b8d9)
- [x] validate if autocommit removal actually works properly

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -187,7 +187,7 @@ class FreqtradeBot(LoggingMixin):
         if self.get_free_open_trades():
             self.enter_positions()
 
-        Trade.query.session.commit()
+        Trade.commit()
 
     def process_stopped(self) -> None:
         """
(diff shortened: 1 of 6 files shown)


pnpn1212:
I wan't config rpc telegram send text from bot. Do you help me?

Issue #4664, opened 2021-04-04
avoid sell_signal becoming active on open trade if profit below certain percentage


klara31:
## Describe your environment

  * Operating system: freqtrade running in docker on a Linux host (freqtradeorg/freqtrade:stable)
  * Freqtrade Version: 2021.3
  
## Your question
If the current profit of an open trade pair is below a certain threshold, I want my strategy not to sell this pair if the pair hits a sell_signal. I cannot figure out how to deal with this in the strategy, any help would be most appreciated!

xmatthias:
You can do this via a combination of `ask_strategy.sell_profit_only` and `ask_strategy.sell_profit_offset` - which defines the profit after which the sell-signal will become active.

klara31:
Thank you, I missed those parameters. These should make it a lot easier!

How does the offset work, as I can not assign a negative value? 
What I want, is that no sell_signal becomes active if profit is (i.e.) below -5%.

xmatthias:
well it's main design is to only work with a positive offset - as i've not seen any usecase where a sell-signa with -4% is valid, while with -6% it isn't - so most people use this as a "positive" offset only - and that's what it's designed for.

klara31:
Can I put up a change request to make this parameter negative as well?

xmatthias:
sure, i don't see why it shouldn't be allowed to be negative.

klara31:
Do I need to do something for this to happen?

xmatthias:
Well my understanding from the above was that you'd be going to create a PR for this (which apparently is a wrong assumption).

In that case, i'll take care of it - but i don't know when yet (it's not a big work - but it's also not something that i see as urgent either).

klara31:
I'm sorry for not being clear.
 
My coding skills are limited. As far as I can see, in file `freqtrade/constants.py` on line 179, the following has to be changed:
```
'sell_profit_offset': {'type': 'number', 'minimum': -100.0},
```
I have never worked with GitHub, other than using `clone` and `pull`, I doubt if you would like me to make my first change on you repository ;) Thanks for your help!

klara31:
I've created a pull request. As it is the first, please be nice to me... ;)

xmatthias:
Closed by #4683

PR #4683, opened 2021-04-06
Allow ask_strategy.sell_profit_offset negative value

klara31:
## Summary
`ask_strategy.sell_profit_offset` currently has a minimal value of 0. Setting it to -100 allows negative values.

Solve the issue: #4664 

## Quick changelog

- Allow parameter `ask_strategy.sell_profit_offset` to get negative value.

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -176,7 +176,7 @@ CONF_SCHEMA = {
                 'order_book_max': {'type': 'integer', 'minimum': 1, 'maximum': 50},
                 'use_sell_signal': {'type': 'boolean'},
                 'sell_profit_only': {'type': 'boolean'},
-                'sell_profit_offset': {'type': 'number', 'minimum': -100},
+                'sell_profit_offset': {'type': 'number'},
                 'ignore_roi_if_buy_signal': {'type': 'boolean'}
             }
         },
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -176,7 +176,7 @@ CONF_SCHEMA = {
                 'order_book_max': {'type': 'integer', 'minimum': 1, 'maximum': 50},
                 'use_sell_signal': {'type': 'boolean'},
                 'sell_profit_only': {'type': 'boolean'},
-                'sell_profit_offset': {'type': 'number', 'minimum': 0.0},
+                'sell_profit_offset': {'type': 'number', 'minimum': -100},
                 'ignore_roi_if_buy_signal': {'type': 'boolean'}
             }
         },


freqtrade/constants.py
```diff
@@ -176,7 +176,7 @@
                 'order_book_max': {'type': 'integer', 'minimum': 1, 'maximum': 50},
                 'use_sell_signal': {'type': 'boolean'},
                 'sell_profit_only': {'type': 'boolean'},
-                'sell_profit_offset': {'type': 'number', 'minimum': 0.0},
+                'sell_profit_offset': {'type': 'number', 'minimum': -100},
```
xmatthias: you can simply remove the whole `'minimum':xxx` key - using -100 is similar to using nothing - and i guess there's no harm in allowing whatever the user selects to provide (even if it's -200)...

xmatthias (approved):
LGTM :+1: 

Welcome as contributor to freqtrade! :+1: 
it's not that hard, is it? :smile:

klara31:
I made this change to my local instance and restarted freqtrade. Last night, a trade was made with a profit percentage below my threshold. So I can enter the negative value, it's accepted but not working correctly...

If I look into the code (searching for `sell_profit_offset`, it is used in `strategy/interface.py`:

```
        if (ask_strategy.get('sell_profit_only', False)
                and current_profit <= ask_strategy.get('sell_profit_offset', 0)):
            # sell_profit_only and profit doesn't reach the offset - ignore sell signal
            sell_signal = False
        else:
            sell_signal = sell and not buy and ask_strategy.get('use_sell_signal', True)
            # TODO: return here if sell-signal should be favored over ROI
```
If I read this correctly, `sell_profit_only` needs to be `false` in order to make `sell_profit_offset` work. Is this by design?

xmatthias:
no, `sell_profit_only` needs to be True to have this take effect (False is the default value).

Now the key (`sell_profit_only`) is kind of misleading now - but i also wouldn't want to change the naming now just because of this.

klara31:
I still don't get it... My config.json had `sell_profit_only = true` and `sell_profit_offset = -5`, and with these parameters a trade was made which had a profit of -7%. 

How is that possible?

xmatthias:
sell_profit_offset = -5 means a profit of -500%. -7% is still fully within this.

if you want it to be 5%, you'll need to use 0.05 ... as it's a ratio (1 = 100%) ... not a percentage value (which is usually only used for "user" convenience ...

klara31:
I should have thought of that... Thanks, also for your patience :)

PR #4684, opened 2021-04-06
Improve Kraken-specific config description.

aapopov92:
Added Warning after Kraken rate limit config in order to clearly highlight that it holds delay between requests instead of req\sec rate.

## Summary
Clearly highlight in docs that ccxt rateLimit is in milliseconds instead of req\sec

## Quick changelog
New warning in Exchanges documentation section


xmatthias:
I don't necessarily mind this addition ... but it's also documented in the [configuration documentation](https://www.freqtrade.io/en/latest/configuration/#sample-exchange-configuration) ...

xmatthias (approved):
LGTM :+1:

Issue #4650, opened 2021-04-02
Add Pairlist Handler for Volatility 


Jay-ArBrouillard:
## Describe the enhancement

In some strategies, I would prefer to enter trades that have a higher historical volatility, so it would be nice if the pairlist was ranked by this with the first pair having the highest volatility. You should be able to define a time period this is calculated on (1 month, 2 month, 3 month, etc.). Some other parameters could be "number of assets" similar to VolumePairList and "minimum volatility" or "maximum volatility".

xmatthias:
I'll mark this as enhancement, but will also leave it there as "good first issue" - as this will only touch a very limited part of Freqtrade, and is therefore great for first time contributors / beginners in contributing to freqtrade.

---

This should not be too difficult to do, and might be a copy / combination of `RangeStabilityFilter` and `VolumePairlist` - basically reversing the logic (and using it for sorting...).

I'm not sure if this should Provide the pairlist, or just sort an available pairlist by it (VolumePairlist  is a provider, most others are just Filters / sorters) 

We've also got a [developer section related to pairlists](https://www.freqtrade.io/en/latest/developer/#pairlists) which explains what each setion does / should be doing.

PR #4686, opened 2021-04-06
Add Pairlist Handler for Volatility

gerrito333:
## Summary
Add Pairlist Handler for Volatility

Solve the issue: #4650

## Quick changelog
- filter a volatility range with min and max volatilty
- unit test added


tests/plugins/test_pairlist.py
```diff
@@ -407,26 +407,33 @@ def test_VolumePairList_refresh_empty(mocker, markets_empty, whitelist_conf):
...
-        ('HOT/BTC', '1d'): ohlcv_history,
+        ('HOT/BTC', '1d'): ohlcv_history_high_vola,
     }
 
     mocker.patch('freqtrade.exchange.Exchange.exchange_has', MagicMock(return_value=True))
 
     if whitelist_result == 'static_in_the_middle':
         with pytest.raises(OperationalException,
-                           match=r"StaticPairList can only be used in the first position "
+                           match=r"StaticPairList only in the first position "
```
xmatthias: This line looks like it's an unintentional change... and causes the test to fail as the exception is now no longer correctly captured / matched.

xmatthias (changes requested):
This is missing a documentation entry for sure.

It's also (at first glance) not clear to me what different volatility values mean (but i hope the docs will fix this).
Assuming from the tests - if i use 0.002 and 0.004 - how much / little movements are this? 
is this % (ratio) values, so it's moves between 0.2 and 0.4% within the last 3 days?

gerrito333:
@xmatthias: I added documentation describing the volatility calculation which is standard deviation of logarithmic returns as described here: https://en.wikipedia.org/wiki/Volatility_(finance).

tests/plugins/test_pairlist.py
gerrito333: done

gerrito333 (commented):
done

docs/includes/pairlists.md
```diff
@@ -164,6 +165,29 @@ If the trading range over the last 10 days is <1%, remove the pair from the whit
...
+In the below example:
+If the volatilty over the last 10 days is not in the range of 0.20-0.30, remove the pair from the whitelist. The filter is applied every 24h.
+
+```json
+"pairlists": [
+    {
+        "method": "VolatilityFilter",
+        "lookback_days": 10,
+        "min_volatilty": 0.20,
+        "max_volatilty": 0.30,
```
xmatthias: You should really fix the typos in these 2 config keys ...
i was testing this and wondering "why on earth does the default config not remove some pairs ... :laughing: 

I also think the "documented" range for this should be wider ... 
Trying this on the top 300 Volume USDT pairs doesn't leave too many pairs (now maybe i tried this at a strange time...) 
similar to other sections in the docs - this will end up as copy/paste sample, and therefore it's beneficial to have "sane" defaults.

docs/includes/pairlists.md
```diff
@@ -164,6 +165,29 @@ If the trading range over the last 10 days is <1%, remove the pair from the whit
 !!! Tip
     This Filter can be used to automatically remove stable coin pairs, which have a very low trading range, and are therefore extremely difficult to trade with profit.
 
+#### VolatilityFilter
```
xmatthias, line 168: Maybe include the wikipedia-link somewhere in this section?

Also (i've never used volatility as such), is there a range the values can be in? 
like from 0.0-1.0 (is this a ratio / percentage value)? 
I'm pretty sure that would simplify finding a "desired" configuration...

freqtrade/plugins/pairlist/VolatilityFilter.py
```diff
@@ -0,0 +1,120 @@
...
+    def _validate_pair_loc(self, pair: str, daily_candles: Optional[DataFrame]) -> bool:
+        """
+        Validate trading range
+        :param pair: Pair that's currently validated
+        :param ticker: ticker dict as returned from ccxt.load_markets()
+        :return: True if the pair can stay, false if it should be removed
+        """
+        # Check symbol in cache
+        if pair in self._pair_cache:
+            return self._pair_cache[pair]
```
xmatthias:
```suggestion
        cached_res = self._pair_cache.get(pair, None)
        if cached_res is not None:
            return cached_res
```

Please change this section to the above code. 
otherwise we'll have a (rare) timing issue where the cache times out between the if and the next line.
This was recently changed (in 0f0607baec ) for RangeStabilityFilter too.

xmatthias (changes requested):
The 2 typos in the docs are a must fix - otherwise others will wonder about the same thing as i did :laughing: 

I'd also want to slightly improve the docs - also the wiki page doesn't really define a range these values can fall into ...

xmatthias (approved):
I've fixed 2 remaining volatilty typos... 

LGTM! :+1: 

epigramx:
I believe this is very restrictive for timeframes lower than ~8hours because it only analyses entire days (which can be late for analyzing movements that might be going on for more than 12 or even 24 hours). Can it be amended to do its analysis on other timeframes on choice?

To be perfectly honest the same should be done for RangeStabilityFilter too.

PR #4703, opened 2021-04-09
Add SKDecimal Space

xmatthias:
## Summary
Add a real Decimal Scikit-optimize space

This will **break** how the current decimal space works / loads it's parameters, as it's fully using the final results for the calculations - but that's not yet in any release, so ... 

This is still a "wip" - and i don't really expect tests to pass ...

## Quick changelog

- introduce Decimal space for hyperopt
- 

@rokups - would you mind having a look at this too? as the original implementation for the decimal space came from you

### DIFF
--- a/freqtrade/optimize/decimalspace.py
+++ b/freqtrade/optimize/decimalspace.py
@@ -1,5 +1,4 @@
 import numpy as np
-
 from skopt.space import Integer
 
 
--- a/freqtrade/optimize/decimalspace.py
+++ b/freqtrade/optimize/decimalspace.py
@@ -1,0 +1,33 @@
+import numpy as np
+
+from skopt.space import Integer
+
+
+class SKDecimal(Integer):
+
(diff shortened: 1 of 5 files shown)


rokups:
Looks great! Though you forgot to clean up few `attr._set_value(params[attr_name])` instances.

xmatthias:
> Though you forgot to clean up few `attr._set_value(params[attr_name])` instances.

I'm not sure i want to remove these ...
setting the value is still necessary, so it'd be a replacement of `attr._set_value(params[attr_name])` with `attr.value = params[attr_name]` - which i'm not sure gives us much benefits for now ...

rokups:
I only added that to handle conversions. At this point keeping it seems rather pointless 👀 also `attr.value = params[attr_name]` is much cleaner as that would be all this function does anyway

xmatthias:
Good point :+1:

PR #4707, opened 2021-04-10
fix documentation inconsistency related to VolatilityFilter

### LINKED ISSUE #4650: Add Pairlist Handler for Volatility

gerrito333:
fixes freqtrade/freqtrade#4650


xmatthias (approved):
LGTM :+1: 

Great catches!

PR #4713, opened 2021-04-11
chore(readme): Fix markdown of docker manual

ugurcemozturk:
## Summary
Link to docker-compose was pointing to the one from develop branch. It's changed as with the stable docker-compose.
All other changes are cosmetic, it fixes the markdown.

### DIFF
--- a/docs/docker_quickstart.md
+++ b/docs/docker_quickstart.md
@@ -22,10 +22,10 @@ Freqtrade provides an official Docker image on [Dockerhub](https://hub.docker.co
 
 ### Docker quick start
 
-Create a new directory and place the [docker-compose file](https://github.com/freqtrade/freqtrade/blob/develop/docker-compose.yml) in this directory.
+Create a new directory and place the [docker-compose file](https://raw.githubusercontent.com/freqtrade/freqtrade/stable/docker-compose.yml) in this directory.
 
-=== "PC/MAC/Linux"
-    ``` bash
+#### PC/MAC/Linux
+    
     mkdir ft_userdata
     cd ft_userdata/
     # Download the docker-compose file from the repository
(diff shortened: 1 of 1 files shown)


docs/docker_quickstart.md
```diff
@@ -22,10 +22,10 @@ Freqtrade provides an official Docker image on [Dockerhub](https://hub.docker.co
 
 ### Docker quick start
 
-Create a new directory and place the [docker-compose file](https://github.com/freqtrade/freqtrade/blob/develop/docker-compose.yml) in this directory.
+Create a new directory and place the [docker-compose file](https://raw.githubusercontent.com/freqtrade/freqtrade/stable/docker-compose.yml) in this directory.
 
-=== "PC/MAC/Linux"
-    ``` bash
+#### PC/MAC/Linux
```
xmatthias: This will break / change rendering on freqtrade.io (the same will happen with all other below changes which replace `!!!` with `###`).

I understand why you're doing it (regular markdown programs will not recognize this format) - but the main purpose of the docs folder is the rendering on freqtrade.io, not plain markdown.

you can look at the [developer documentation](https://www.freqtrade.io/en/latest/developer/#documentation) on how to serve the documentation locally to check if it renders correctly.

docs/docker_quickstart.md
```diff
@@ -68,20 +67,19 @@ Create a new directory and place the [docker-compose file](https://github.com/fr
...
+##### Adding a custom strategy
 
 1. The configuration is now available as `user_data/config.json`
 2. Copy a custom strategy to the directory `user_data/strategies/`
 3. Add the Strategy' class name to the `docker-compose.yml` file
 
 The `SampleStrategy` is run by default.
 
-!!! Warning "`SampleStrategy` is just a demo!"
+#### Warning "`SampleStrategy` is just a demo!"
```
xmatthias: the `!!!` have been here intentional - as these render the "Warning" or "note" boxes on freqtrade.io.

xmatthias (changes requested):

xmatthias (approved):
I've cleaned this up a bit from the unwanted changes (as mentioned above, main purpose of the docs folder is rendering in freqtrade.io - which requires these "non-standard" markdown tags to render nicely.

Anyhow, the intended change (switch to stable docker-compose file) is great - thanks for noticing 👍

Issue #4653, opened 2021-04-03
(sqlite3.OperationalError) database is locked


epigramx:
I was fiddling with the web ui (only viewing: reloading at the order tab: no actual trading or changes) and the 'regular' software crashed at that moment while it was trying to order something (it went through at the exchange but not at the database). 

```
freqtrade/freqtradebot.py", line 625, in execute_buy
    Trade.session.flush()
  File "<string>", line 2, in flush
sqlalchemy/orm/session.py", line 3233, in flush
    self._flush(objects)
sqlalchemy/orm/session.py", line 3373, in _flush
    transaction.rollback(_capture_exception=True)
sqlalchemy/util/langhelpers.py", line 70, in __exit__
    compat.raise_(
sqlalchemy/util/compat.py", line 198, in raise_
    raise exception
sqlalchemy/orm/session.py", line 3369, in _flush
    transaction.commit()
sqlalchemy/orm/session.py", line 836, in commit
    trans.commit()
sqlalchemy/engine/base.py", line 2257, in commit
    self._do_commit()
sqlalchemy/engine/base.py", line 2421, in _do_commit
    self._connection_commit_impl()
sqlalchemy/engine/base.py", line 2392, in _connection_commit_impl
    self.connection._commit_impl()
sqlalchemy/engine/base.py", line 930, in _commit_impl
    self._handle_dbapi_exception(e, None, None, None, None)
sqlalchemy/engine/base.py", line 1929, in _handle_dbapi_exception
    util.raise_(
sqlalchemy/util/compat.py", line 198, in raise_
    raise exception
sqlalchemy/engine/base.py", line 928, in _commit_impl
    self.engine.dialect.do_commit(self.connection)
sqlalchemy/engine/default.py", line 634, in do_commit
    dbapi_connection.commit()
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) database is locked
(Background on this error at: http://sqlalche.me/e/14/e3q8)
freqtrade.commands.trade_commands - INFO - worker found ... calling exit
```

epigramx:
Also I mentioned "viewing only" to state the severity of the problem. It does not mean it should crash if the web ui was trying to edit trading while the rest of the software does the same. It's clear that there should be a transactional nature in these things and a lock of some sort that avoids this sort of thing is missing.

epigramx:
This thread is interesting, though not just for the top answer, but also this answer: https://stackoverflow.com/a/53470179/277716

[...]

PR #4717, opened 2021-04-12
Retry accessing the database, if it's too busy. Applicable when the UI pummels large databases.

epigramx:
## Summary of the purpose of the PR
This solidifies solving #4653 once and for all, because we determined that FT is most probably perfectly thread safe and handles SQLAlchemy session creation well (because every session has its unique ID on each thread), however it appears normal for the sqlite file itself to be locked at the sqlite library level when thousands of old trades are requested at the same time (I can not reproduce the problem on a small database but I can do it easily on a database of thousands of old trades by pummeling the reload button of the Trade tab of the UI even on a relatively fast PC).

## Detailed justification
This is justified for the following reasons:

1. We have determined that FT does nothing wrong with thread safety or SQLAlchemy session handling because each thread has its own unique session ID during debugging; however:
2. The SQLite database file may be busy during very lengthy operations when it is big enough and the entire length of a table is requested at the same time; so it should be retried:
3. Every backtrace involves self.freqtrade.process() at worker.py which already includes similar exception handling so it's elegant and simple and very short to do it there.

## Expected behavior
Step1: Get a large database of thousands of old trades and keep pummeling the Reload button of the Trade tab of the UI

Step2: The software may catch the problem:
```
freqtrade.worker - INFO - Bot heartbeat. PID=25401, version='develop-8b102f457', state='RUNNING'
freqtrade.worker - WARNING - Warning: [..] (sqlite3.OperationalError) database is locked[..], retrying in half a second...
freqtrade.worker - INFO - Bot heartbeat. PID=25401, version='develop-8b102f457', state='RUNNING'
```
Step 3: Success; the trading continues as normal in half a second; it affected nothing and it does not affect those without the issue.

### DIFF
--- a/freqtrade/worker.py
+++ b/freqtrade/worker.py
@@ -15,6 +15,7 @@ from freqtrade.exceptions import OperationalException, TemporaryError
 from freqtrade.freqtradebot import FreqtradeBot
 from freqtrade.state import State
 
+from sqlalchemy.exc import OperationalError
 
 logger = logging.getLogger(__name__)
 
@@ -154,6 +155,9 @@ class Worker:
 
             logger.exception('OperationalException. Stopping trader ...')
             self.freqtrade.state = State.STOPPED
+        except OperationalError as error:
+            logger.warning(f"Warning: {error}, retrying in half a second...")
+            time.sleep(0.5)
 
     def _reconfigure(self) -> None:
         """


freqtrade/worker.py
```diff
@@ -154,6 +155,9 @@ def _process_running(self) -> None:
 
             logger.exception('OperationalException. Stopping trader ...')
             self.freqtrade.state = State.STOPPED
+        except OperationalError as error:
+            logger.warning(f"Warning: {error}, retrying in half a second...")
+            time.sleep(0.5)
```
xmatthias, line 160: Uh that's a ... dangerous approach, and will clearly fail to do what you expect it to.
There's a reason we only continue in some rare cases (TemporaryError - but not OperationalException or worse).

while it will will stop the exception from crashing the bot - you will end up with open orders (or filled orders?) on the exchange which the bot does not know about (because it failed to insert them into the database) if this exception really happens, and the bot will therefore also not handle these.

You will not see this in dry-runs - as dry-runs ONLY happen within freqtrade, so the bot will never know about the trade in the first place - but in live this WILL fail.

To me, this is a lot worse than failing (because then you know you need to look at what happened).

xmatthias (changes requested):

freqtrade/worker.py
epigramx: As we discussed on discord: if you think that is true - and I have no reason now to doubt it, then I believe you must mutex the threads of the software because it's obvious it locks at the sqlite library level when a lot of .all()s are happening on a large database.

I will personally use it for the time being because most crashes happen without trading editing.

epigramx:
For the time being: to satisfy your needs for safety of the trading transactions: and because it's either complex to remove the autocommit/autoflush functionality or it might not help anyway since as you also said it might still fail on commit(),
a good middle ground for the time being would be to do the exception handing for retrying but **on each flushing**.
That way FT both retries if the database is busy and it maintains safety of its trading.

freqtrade/worker.py
epigramx: An alternative to restricting the threads from access, would be to wrap the retrying based on the locking exception (which is done by the simple patch originally posted in the PR) in all cases of flushing or other implicit cases of flushing (e.g. those filter().all() situations) so that your requirement for trading safety is satisfied.

I might attempt to do that for my use but for something that ends up upstream you might want something prettier I guess like removing autoflushing and autocommiting and then carefully using commit() which I don't feel confident myself doing correctly but as you had said that might not alter the db locking anyway.

PS It would though make the code prettier by not using flushing willy-nilly all the time but only when the  programmer knows for sure alterations had been made (assuming it's done correctly).

epigramx:
I'll leave this comment here for now because I believe it's the closest to the truth based on what we know so far: that sqlite is known to not be always terrible for concurrency but in FT it clearly is so something is wrong here.

image

FT keeps open sessions/connections and mainly flushes them which appears to be an obvious cause for locking issues.

That means if autoflushing/autocommitting is migrated to explicit commiting it might help.

epigramx:
tl;dr: let's bite the bullet and migrate the thing from autoflushing/autocommitting, to committing and re-opening connections only when required to actually edit the data.
it will make it much prettier and manageable later anyway.

xmatthias:
Unfortunately, a screenshot from (probably) stackoverflow is kindof no information, especially not with a link that shows the remaining of the discussion / question.

many people post there, and i've seen a TON of false information there - so i'd always take answers you get from there with a grain of salt.

epigramx:
I didn't take it as fact: it was only a reference to what already makes sense: FT does not close its connections(!) (elegantly). At the very least it would make sense to commit() explicitly when the programmer knows for sure there is editing in the database because now its chaotic since it flushes at general moments when its unclear if there was even editing of the db; it would make the code clearer to read and understand; and it may even "automatically" fix the issue since it would probably make clear when to not have an open connection when it's unneeded. 

For the time being I use a wrapper based on the original patch here to retry on all flushes and I disabled autoflushing (a solution that satisfies your requirement to keep trading safe); you still do not want that patch because you think you can do better; I think you could indeed do better if you open and close connections when needed / commit only when there is actually new data for the db but I feel out of my depth to do it easily (you don't find it easy either but I believe it's worth it (at the very least to make the code better to understand and manage)).

epigramx:
In light of branch sqlalchemy_14 which nicely removes autocommit, but it still makes it possible to get the same locking exception under the same conditions (e.g. abuse of reloading at the UI), the best candidate I think of right now is to also remove autoflushing because it appears to also flush during regular reading (!) which is an obvious cause for failure when flushing is required for actual db editing.
PS I also have in mind the possibility that it's the connections per thread that stay open and are not closed when not needed but I put that further down in the list of possibilities because "if it doesn't like open connections at all: wouldn't it lock sooner?".

epigramx:
To complete that thought: it appears that the software to be complete it requires the following changes/considerations:

1. Autoflushing should be removed so that flushing is not accuring even when it's only reading (as it is confirmed by multiple traces of the locking that report it flushes when it only does filter().all()) because when the rest of the software does try to actually edit the dbase at the same time it will be locked at that moment.
2. However that is not enough: it will fix the practical problem in "90%" of cases but the UI can also edit itself: hence the UI and the regular software should not be able to commit simultaneously.
3. Hence the total solution is a) Removal of autoflushing b) Edits to the dbase become the only ones that can commit c) Mutual exclusion of threads to prevent them committing simultaneously.

xmatthias:
I'll close this - as in this state, we won't be able to use it (the risk of simply retrying is too high - as you'll not know if an order was opened and "forgotten" or not).

For everyone reading this in the future - i'm more than happy to accept contributions that try to implement the above discussed suggestions.

PR #4742, opened 2021-04-15
Correct get_min_pair_stake_amount formula

theomart:
## Summary
Fix get_min_pair_stake_amount formula.

## What's new?
If I understand well, the goal is to have stoploss price above the minimum required by the exchange + a reserve, which is not what has been achieved so far.
Only if `amount_reserve_percent = (1 + pad) / (1 - abs(stoploss))` we get `final stake amount * (1 - abs(stoploss)) > minimum required`.

The maths:
The reserve of 5% above the minimum stake amount required by the exchange: `reserve = config['amount_reserve_percent']`
The minimum stake amount required by the exchange: `min_required_by_exchange = max(markets['limits']['cost']['min'], markets['limits']['amount']['min'] * price)`
We want the stake amount of the stoploss to be above the minimum + the reserve: `min_stoploss_stake_amount > min_required_by_exchange * (1 + reserve)`
We know that, for a stoploss expressed with a positive float:`min_stoploss_stake_amount = min_buy_stake_amount * (1 - stoploss)`
So, `min_buy_stake_amount * (1 - stoploss) > min_required_by_exchange * (1 + reserve)`
Then `min_buy_stake_amount > min_required_by_exchange * (1 + reserve) / (1 - stoploss)`

Here is a code example, with numbers from the example in the doc for this function:
```
# Inputs
markets = {
    'limits': {
        'cost': {
            'min': 10
        },
        'amount': {
            'min': 20
        }
    }
}

price = 0.6
config = {
    'amount_reserve_percent': 0.05
}

stoploss = -0.1

# Common to both old and new logic
min_required_by_exchange = max(markets['limits']['cost']['min'], 
                               markets['limits']['amount']['min'] * price)

# We want both the stoploss stake amount and the buy order stake amount to be above this value
min_required_by_exchange_with_reserve = min_required_by_exchange * (1 + config['amount_reserve_percent'])

# OLD LOGIC (min_stake_amount would be the value returned by get_min_pair_stake_amount formula)
min_stake_amount =  min_required_by_exchange *\
                    (1 + config['amount_reserve_percent'] + abs(stoploss))
stoploss_stake_amount = min_stake_amount * (1 + stoploss)
assert stoploss_stake_amount >= round(min_required_by_exchange_with_reserve, 8), 'Stoploss should be >= min stake amount required by exchange + pad'

# NEW LOGIC
min_stake_amount = min_required_by_exchange *\
                    ((1 + config['amount_reserve_percent']) / (1 - abs(stoploss)))
stoploss_stake_amount = min_stake_amount * (1 + stoploss)
assert stoploss_stake_amount >= round(min_required_by_exchange_with_reserve, 8), 'Stoploss should be >= min stake amount required by exchange + pad'
```

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -541,7 +541,7 @@ class Exchange:
                                                         DEFAULT_AMOUNT_RESERVE_PERCENT)
         amount_reserve_percent = (
           amount_reserve_percent / (1 - abs(stoploss)) if abs(stoploss) != 1 else 1.5
-        )     
+        )
         # it should not be more than 50%
         amount_reserve_percent = max(min(amount_reserve_percent, 1.5), 1)
 
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -539,7 +539,9 @@ class Exchange:
         # reserve some percent defined in config (5% default) + stoploss
         amount_reserve_percent = 1.0 + self._config.get('amount_reserve_percent',
                                                         DEFAULT_AMOUNT_RESERVE_PERCENT)
-        amount_reserve_percent += abs(stoploss)
+        amount_reserve_percent = (
+          amount_reserve_percent / (1 - abs(stoploss)) if abs(stoploss) != 1 else 1.5
+        )     
         # it should not be more than 50%
         amount_reserve_percent = max(min(amount_reserve_percent, 1.5), 1)
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -167,7 +167,7 @@ This exchange has also a limit on USD - where all orders must be > 10$ - which h
 
 To guarantee safe execution, freqtrade will not allow buying with a stake-amount of 10.1$, instead, it'll make sure that there's enough space to place a stoploss below the pair (+ an offset, defined by `amount_reserve_percent`, which defaults to 5%).
 
-With a stoploss of 10% - we'd therefore end up with a value of ~13.8$ (`12 * (1 + 0.05 + 0.1)`).
+With a reserve of 5%, the minimum stake amount would be ~12.6$ (`12 * (1 + 0.05)`). If we take in account a stoploss of 10% on top of that - we'd end up with a value of ~14$ (`12.6 / (1 - 0.1)`).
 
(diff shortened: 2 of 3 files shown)


xmatthias:
to be honest, i find your calculation example rather difficult to follow, because you
 
* use different variable names - so I need to piece together what in the function corresponds to what in the sample
* don't really echo the result (what the new minimum stake would be) - but only the stoploss price (at which point the stoploss would be set ?)

Now that does neither say you're correct or wrong (i assume it's correct for now ...) - just that it's ... not easy to verify for now.

theomart:
Hey Matthias - sorry for the churn. I updated the example to reflect as much as possible the variables used in the code and I specified in comments the return value. Hope that helps.

freqtrade/exchange/exchange.py
```diff
@@ -539,7 +539,9 @@ def get_min_pair_stake_amount(self, pair: str, price: float,
         # reserve some percent defined in config (5% default) + stoploss
         amount_reserve_percent = 1.0 + self._config.get('amount_reserve_percent',
                                                         DEFAULT_AMOUNT_RESERVE_PERCENT)
-        amount_reserve_percent += abs(stoploss)
+        amount_reserve_percent = (
+          amount_reserve_percent / (1 - abs(stoploss)) if abs(stoploss) != 1 else 1.5
+        )
```
xmatthias: Apparently this line has a trailing whitespace (and one of the test lines is > 100 characters) - which makes CI fail.

Apart from that, i do think you're right (and the calculation confirms this as well) - so thanks for looking into this!

theomart: Yep, fixing that now. Thanks for your patience and for the library :) 
BTW, in your opinion, is a PR the best practice to fix those kind of small quirks ?

xmatthias: define "small quirks" ? you mean the CI errors - or the calculation problem?

a PR is usually the best way to make changes to a project. obviously, an issue is also possible if you discover a problem but are unsure about a fix - but if it's a one-line fix (like in this case) - a PR is the quickest way to point out the problem, while at the same  time fixing the problem.

Now ideally, ci passes right away - but i often forget to run the tools we use myself - however that's what we got CI for - to tell us that there's some small stylistic errors (without much discussion, as CI tells us what's wrong).

xmatthias (approved):
LGTM 👍 

Thanks for noticing and fixing!

PR #4743, opened 2021-04-15
remove `copy()` from `custom_info` example

TheJoeSchr:
`set_index` automatically copies if not stated otherwise with `inplace=True`
> inplacebool, default False
If True, modifies the DataFrame in place (do not create a new object).

from: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.set_index.html?highlight=set_index#pandas.DataFrame.set_index


xmatthias (approved):

xmatthias (approved):
LGTM 👍

PR #4746, opened 2021-04-16
Telegram: forcebuy inline keyboard

gmatheu:
## Summary
Inline keyboard for **forcebuy** command

Solve the issue: n/a

## Quick changelog

- Added inline keyboard response 

## What's new?

This PR adds a functionality meant to facilitate forcebuys. When you execute **/forcebuy** without arguments, it generates an inline keyboard with the allow-listed pairs. If you click any of those buttons it executes the forcebuy for that pair.

Screen Shot 2021-04-15 at 21 48 02


freqtrade/rpc/telegram.py
```diff
@@ -610,6 +616,25 @@ def _forcesell(self, update: Update, context: CallbackContext) -> None:
         except RPCException as e:
             self._send_msg(str(e))
 
+    def _forcebuy_action(self, pair, price=None):
+        try:
+            self._rpc._rpc_forcebuy(pair, price)
+        except RPCException as e:
+            self._send_msg(str(e))
+
+    def _forcebuy_inline(self, update: Update, _: CallbackContext) -> None:
```
xmatthias, line 652: I'm not entirely sure about this (i never used inline-keyboards with telegram so far) - but would we need to add the `@authorize_only` tag to this method too, or is this only reachable as part of an answer?

xmatthias (approved):
Overall looks nice (although i didn't play with it yet) ... 

Is there a way to "abort" the forcebuy? 
say i sent a `/forcebuy` by mistake - and now i don't really want to do this anymore? can i simply answer with something else, or should we add a "abort" button to the inline answer for easy abortion?


*edit* was meant to be a comment, not an "approve" ... not yet, anyway :laughing:

xmatthias:
@gmatheu you submitted this as "WIP" / draft ... 
however this seems pretty complete ... would you mind letting me know what you think is still missing?

freqtrade/rpc/telegram.py
gmatheu: Let me double-check that...

xmatthias: Trying this, i found no way to get to this callback without having an authorization first - so all fine.

xmatthias (approved):
LGTM :+1: 
Great enhancement!

PR #4750, opened 2021-04-17
Add IStrategy.custom_sell method which allows per-trade sell signal evaluation

### LINKED ISSUE #4276: availability of trade profit while generating sell signals
Hi first. I am sorry for my english i know it's not good. I am 32 years old and my hobby is trading. Its a hobby because i won a lot more with my company on accountability. 

For the last 2 years  I was away from trading because I was studying an MBA. But now i finished and for the last 3 moth i was studying  freqtrade. My last bot was wiht Gekko.

Finished that little intro. I want to say Freqtrade has 2 problems for me:

1) I have 3 signal for buy and I want to use different strategy  for sell For example if i buy with the signal A i will use the logic C for sell

2) This is the worst problem: I need to control the sell signal with the profit.
For example if my profit of the current trade is >3% I will use the logic A, but if my profit is between 0 and 3% I will use a different strategy.
And if my profit an hour ago was negative i will use another strategy

1/2 its necessary to use on backtest (if you can't test the strategy you can't go to productive)

This is my see point . Freqtrade  is a great work !! But need to have more flexibility

rokups:
Implementing `custom_sell()` in a strategy is conceptually equal to setting `sell` signal in dataframe.

closes #4276

### DIFF
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -42,33 +42,6 @@ class AwesomeStrategy(IStrategy):
 
 ***
 
-### Storing custom information using DatetimeIndex from `dataframe`
-
-Imagine you need to store an indicator like `ATR` or `RSI` into `custom_info`. To use this in a meaningful way, you will not only need the raw data of the indicator, but probably also need to keep the right timestamps.
-
-```python
-import talib.abstract as ta
-class AwesomeStrategy(IStrategy):
-    # Create custom dictionary
-    custom_info = {}
-
-    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
-        # using "ATR" here as example
-        dataframe['atr'] = ta.ATR(dataframe)
-        if self.dp.runmode.value in ('backtest', 'hyperopt'):
-          # add indicator mapped to correct DatetimeIndex to custom_info
-          self.custom_info[metadata['pair']] = dataframe[['date', 'atr']].set_index('date')
-        return dataframe
-```
-
-!!! Warning
-    The data is not persisted after a bot-restart (or config-reload). Also, the amount of data should be kept smallish (no DataFrames and such), otherwise the bot will start to consume a lot of memory and eventually run out of memory and crash.
-
-!!! Note
-    If the data is pair-specific, make sure to use pair as one of the keys in the dictionary.
-
-See `custom_stoploss` examples below on how to access the saved dataframe columns
-
 ## Custom sell signal
 
 It is possible to define custom sell signals. This is very useful when we need to customize sell conditions for each individual trade.
@@ -107,7 +80,8 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, **kwargs) -> float:
+                        current_rate: float, current_profit: float, dataframe: Dataframe,
+                        **kwargs) -> float:
         """
         Custom stoploss logic, returning the new distance relative to current_rate (as ratio).
         e.g. returning -0.05 would create a stoploss 5% below current_rate.
@@ -157,7 +131,8 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, **kwargs) -> float:
+                        current_rate: float, current_profit: float, dataframe: DataFrame,
+                        **kwargs) -> float:
 
         # Make sure you have the longest interval first - these conditions are evaluated from top to bottom.
         if current_time - timedelta(minutes=120) > trade.open_date_utc:
@@ -183,7 +158,8 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, **kwargs) -> float:
+                        current_rate: float, current_profit: float, dataframe: DataFrame,
+                        **kwargs) -> float:
 
         if pair in ('ETH/BTC', 'XRP/BTC'):
             return -0.10
@@ -209,7 +185,8 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, **kwargs) -> float:
+                        current_rate: float, current_profit: float, dataframe: DataFrame,
+                        **kwargs) -> float:
 
         if current_profit < 0.04:
             return -1 # return a value bigger than the inital stoploss to keep using the inital stoploss
@@ -249,7 +226,8 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, **kwargs) -> float:
+                        current_rate: float, current_profit: float, dataframe: DataFrame,
+                        **kwargs) -> float:
 
         # evaluate highest to lowest, so that highest possible stop is used
         if current_profit > 0.40:
@@ -266,14 +244,20 @@ class AwesomeStrategy(IStrategy):
 
 Imagine you want to use `custom_stoploss()` to use a trailing indicator like e.g. "ATR"
 
-See: "Storing custom information using DatetimeIndex from `dataframe`" example above) on how to store the indicator into `custom_info`
-
 !!! Warning
-    only use .iat[-1] in live mode, not in backtesting/hyperopt
-    otherwise you will look into the future
+    Only use `dataframe` values up until and including `current_time` value. Reading past
+    `current_time` you will look into the future, which will produce incorrect backtesting results
+    and throw an exception in dry/live runs.
     see [Common mistakes when developing strategies](strategy-customization.md#common-mistakes-when-developing-strategies) for more info.
 
+!!! Note
+    DataFrame is indexed by candle date. During dry/live runs `current_time` and
+    `trade.open_date_utc` will not match candle dates precisely and using them as indices will throw
+    an error. Use `date = timeframe_to_prev_date(self.timeframe, date)` to round a date to previous
+    candle before using it as a `dataframe` index.
+
 ``` python
+from freqtrade.exchange import timeframe_to_prev_date
 from freqtrade.persistence import Trade
 from freqtrade.state import RunMode
 
@@ -284,28 +268,19 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, **kwargs) -> float:
+                        current_rate: float, current_profit: float, dataframe: DataFrame,
+                        **kwargs) -> float:
 
         result = 1
         if self.custom_info and pair in self.custom_info and trade:
-            # using current_time directly (like below) will only work in backtesting.
-            # so check "runmode" to make sure that it's only used in backtesting/hyperopt
-            if self.dp and self.dp.runmode.value in ('backtest', 'hyperopt'):
-              relative_sl = self.custom_info[pair].loc[current_time]['atr']
-            # in live / dry-run, it'll be really the current time
-            else:
-              # but we can just use the last entry from an already analyzed dataframe instead
-              dataframe, last_updated = self.dp.get_analyzed_dataframe(pair=pair,
-                                                                       timeframe=self.timeframe)
-              # WARNING
-              # only use .iat[-1] in live mode, not in backtesting/hyperopt
-              # otherwise you will look into the future
-              # see: https://www.freqtrade.io/en/latest/strategy-customization/#common-mistakes-when-developing-strategies
-              relative_sl = dataframe['atr'].iat[-1]
-
-            if (relative_sl is not None):
+            # Using current_time directly would only work in backtesting. Live/dry runs need time to
+            # be rounded to previous candle to be used as dataframe index. Rounding must also be 
+            # applied to `trade.open_date(_utc)` if it is used for `dataframe` indexing.
+            current_time = timeframe_to_prev_date(self.timeframe, current_time)
+            current_row = dataframe.loc[current_time]
+            if 'atr' in current_row:
                 # new stoploss relative to current_rate
-                new_stoploss = (current_rate-relative_sl)/current_rate
+                new_stoploss = (current_rate - current_row['atr']) / current_rate
                 # turn into relative negative offset required by `custom_stoploss` return implementation
                 result = new_stoploss - 1
 
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -69,6 +69,20 @@ class AwesomeStrategy(IStrategy):
 
 See `custom_stoploss` examples below on how to access the saved dataframe columns
 
+## Custom sell signal
+
+It is possible to define custom sell signals. This is very useful when we need to customize sell conditions for each individual trade.
+
+An example of how we can sell trades that were open longer than 1 day:
+
+``` python
+class AwesomeStrategy(IStrategy):
+    def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
+                    current_profit: float, **kwargs) -> bool:
+        time_delta = datetime.datetime.utcnow() - trade.open_date_utc
+        return time_delta.days >= 1
+```
+
 ## Custom stoploss
 
 The stoploss price can only ever move upwards - if the stoploss value returned from `custom_stoploss` would result in a lower stoploss price than was previously set, it will be ignored. The traditional `stoploss` value serves as an absolute lower level and will be instated as the initial stoploss.
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -52,8 +52,8 @@ An example of how we can set stop-loss and take-profit targets in the dataframe 
 class AwesomeStrategy(IStrategy):
     def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
                     current_profit: float, dataframe: Dataframe, **kwargs) -> bool:
-        trade_row = dataframe.loc[timeframe_to_prev_date(trade.open_date_utc)]
-        
+        trade_row = dataframe.loc[dataframe['date'] == timeframe_to_prev_date(trade.open_date_utc)].squeeze()
+
         # Sell when price falls below value in stoploss column of taken buy signal.
         if 'stop_loss' in trade_row:
             if current_rate <= trade_row['stop_loss'] < trade.open_rate:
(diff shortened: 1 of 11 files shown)


freqtrade/strategy/interface.py
```diff
@@ -535,9 +556,12 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
                 and current_profit <= ask_strategy.get('sell_profit_offset', 0)):
             # sell_profit_only and profit doesn't reach the offset - ignore sell signal
             sell_signal = False
-        else:
-            sell_signal = sell and not buy and ask_strategy.get('use_sell_signal', True)
+        elif ask_strategy.get('use_sell_signal', True):
+            sell = sell or self.custom_sell(trade.pair, trade, date, current_rate, current_profit)
```
xmatthias: this should be wrapped in `strategy_safe_wrapper` ... it'll catch potential user exceptions - which i see a necessity for these functions, as they might get quite nested, where not every branch might work in some cases.

rokups:
Pushed some docs + support for returning a string as custom sell reason. Please let me know if this looks ok. If yes - we can worry about tests then. 👍🏻 

image

freqtrade/strategy/interface.py
```diff
@@ -531,12 +558,26 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
...
+                custom_reason = strategy_safe_wrapper(self.custom_sell, default_retval=False)(
+                    trade.pair, trade, date, current_rate, current_profit)
+                if custom_reason:
+                    sell_signal = SellType.CUSTOM_SELL
+                    if isinstance(custom_reason, bool):
+                        custom_reason = None
+                    elif isinstance(custom_reason, str):
+                        if len(custom_reason) > CUSTOM_SELL_MAX_LENGTH:
+                            raise OperationalException('Custom sell reason returned '
+                                                       'from custom_sell is too long.')
```
xmatthias: i'm not sure about raising an exception here - assuming a nested `custom_sell()` method which only hits a branch every once in a while ... 

So we'd risk a bot running for weeks - and then crash with `OperationalException` because of this ... 
maybe we can trim it to the allowed characters and warn about it instead?

that way it'll be "noisy as hell" in backtesting  should you hit that (intentional) - but won't break a live-bot because of a too long given reason.

freqtrade/freqtradebot.py
```diff
@@ -946,7 +946,7 @@ def _check_and_execute_sell(self, trade: Trade, sell_rate: float,
 
         if should_sell.sell_flag:
             logger.info(f'Executing Sell for {trade.pair}. Reason: {should_sell.sell_type}')
-            self.execute_sell(trade, sell_rate, should_sell.sell_type)
+            self.execute_sell(trade, sell_rate, should_sell.sell_type, should_sell.sell_reason)
```
xmatthias: Maybe we can have the tuple (`should_sell / SellCheckTuple`) as parameter here instead of using 2 of the 3 provided values seperately?

docs/strategy-advanced.md
```diff
@@ -69,6 +69,20 @@ class AwesomeStrategy(IStrategy):
...
+
+It is possible to define custom sell signals. This is very useful when we need to customize sell conditions for each individual trade.
+
+An example of how we can sell trades that were open longer than 1 day:
+
+``` python
+class AwesomeStrategy(IStrategy):
+    def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
+                    current_profit: float, **kwargs) -> bool:
+        time_delta = datetime.datetime.utcnow() - trade.open_date_utc
```
xmatthias: you should definitely not use "datetime.datetime.now()" here - but current_time.

Otherwise it'll break backtesting ... 

I'd however make this example ... a bit better (this example can be done by setting ROI of -1 too).
Maybe we can check for trades > 1 day that are in the red currently (freeing up a negative trade slot) - while staying in trades that are positive (riding the profits ...) ?

freqtrade/freqtradebot.py
rokups: Will do 👍🏻 But now that im looking at the code i see `SellCheckTuple` having bool + `SellType`. However each time `sell_flag=False` we also have `sell_type=SellType.NONE` and each time `sell_flag=True` then `sell_type` is something other than `SellType.NONE`. It seems like `sell_flag` is redundant here. I think we should get rid of it and simply have `SellCheckTuple` contain `SellType` + `str` reason. Any objections?

xmatthias: yes, that'll be fine and should be a nice improvement too 👍

freqtrade/optimize/backtesting.py
```diff
@@ -255,7 +255,7 @@ def _get_sell_trade_entry(self, trade: LocalTrade, sell_row: Tuple) -> Optional[
 
         if sell.sell_flag:
             trade.close_date = sell_row[DATE_IDX]
-            trade.sell_reason = sell.sell_type.value
+            trade.sell_reason = sell.sell_reason or sell.sell_type.value
```
xmatthias: i guess we can now use `sell.sell_reason` directly, as the `or` is already happening in SellCheckType ...

docs/strategy-advanced.md
```diff
@@ -42,32 +42,34 @@ class AwesomeStrategy(IStrategy):
...
+            if current_rate <= trade_row['stop_loss'] < trade.open_rate:
+                return 'stop_loss'
+
+        # Sell when price reaches value in take_profit column of taken buy signal.
+        if 'take_profit' in trade_row:
+            if trade.open_rate < trade_row['take_profit'] <= current_rate:
+                return 'take_profit'
+
+        # Sell any positions at a loss if they are helpd for more than two days.
+        if current_profit < 0 and (current_time.replace(tzinfo=trade.open_date_utc.tzinfo) - trade.open_date_utc).days >= 1:
```
xmatthias: under what circumstance did you get current_time without utc timezone? (backtest / dry / unit tests)

Having it here indicates that you ran into it - which would point to a bug as this date should ALWAYS have utc timezone assigned.

docs/strategy-advanced.md
```diff
@@ -42,32 +42,34 @@ class AwesomeStrategy(IStrategy):
...
 ```
 
-!!! Warning
-    The data is not persisted after a bot-restart (or config-reload). Also, the amount of data should be kept smallish (no DataFrames and such), otherwise the bot will start to consume a lot of memory and eventually run out of memory and crash.
-
-!!! Note
-    If the data is pair-specific, make sure to use pair as one of the keys in the dictionary.
-
-See `custom_stoploss` examples below on how to access the saved dataframe columns
+See [Custom stoploss using an indicator from dataframe example](strategy-customization.md#custom-stoploss-using-an-indicator-from-dataframe-example) for explanation on how to use `dataframe` parameter.
```
xmatthias: Where's that link supposed to point to? it doesn't work / does not point to the proper anchor in the strategy page ...

docs/strategy-advanced.md
```diff
@@ -42,32 +42,34 @@ class AwesomeStrategy(IStrategy):
...
+``` python
+class AwesomeStrategy(IStrategy):
+    def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
+                    current_profit: float, dataframe: Dataframe, **kwargs) -> bool:
+        trade_row = dataframe.loc[dataframe['date'] == timeframe_to_prev_date(trade.open_date_utc)].squeeze()
+
+        # Sell when price falls below value in stoploss column of taken buy signal.
+        if 'stop_loss' in trade_row:
+            if current_rate <= trade_row['stop_loss'] < trade.open_rate:
+                return 'stop_loss'
```
xmatthias: I find variable / column naming here quite confusing. 

using "stop_loss" somehow indicates that it has something to do with stoploss - but without explaining what, really.
maybe using atr (as in the stoploss example further down) would make more sense?

Also, by using stop_loss, it'll collide in the sell-reason list with the "real" stoploss

docs/strategy-advanced.md
```diff
@@ -42,32 +42,34 @@ class AwesomeStrategy(IStrategy):
...
-        dataframe['atr'] = ta.ATR(dataframe)
-        if self.dp.runmode.value in ('backtest', 'hyperopt'):
-          # add indicator mapped to correct DatetimeIndex to custom_info
-          self.custom_info[metadata['pair']] = dataframe[['date', 'atr']].set_index('date')
-        return dataframe
+``` python
+class AwesomeStrategy(IStrategy):
+    def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
+                    current_profit: float, dataframe: Dataframe, **kwargs) -> bool:
+        trade_row = dataframe.loc[dataframe['date'] == timeframe_to_prev_date(trade.open_date_utc)].squeeze()
```
xmatthias: we should check that trade_row is not empty.

in the (most probably unlikely) case that the trade was opened more than 500 (in reality ohlcv_limit, so most likely 1000 for binance) candles ago, you might get an empty row here.

i understand that can't happen due to the last condition - but this is docs - so people will probably copy/paste parts of it only.

freqtrade/strategy/interface.py
```diff
@@ -285,6 +295,31 @@ def custom_stoploss(self, pair: str, trade: Trade, current_time: datetime, curre
         """
         return self.stoploss
 
+    def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
+                    current_profit: float, dataframe: DataFrame,
+                    **kwargs) -> Optional[Union[str, bool]]:
```
xmatthias, line 301: is there any benefit to returning None?

it's treated the same as False (as far as i can see) - so i guess we could just return False here - and the type would be simpler ... 
sure, users may still return None - but that's not really a problem i'd think, as None and False are both "falsy" ..

docs/strategy-advanced.md
rokups: Hyperopt or backtesting, im not 100% sure. However these dates were UTC, but tzinfo objects had different types (even though both said 'UTC'), so it threw an exception.

docs/strategy-advanced.md
rokups: Oopsie. I was trying to link to custom stoploss with indicators example since it has more details on `dataframe` usage.

docs/strategy-advanced.md
rokups: Yeah its just what i lifted from my experiments strategy. Might not be the best thing to include in docs. However this kind of usage made me think about it from a bit different angle.

I find it _extremely_ convenient to have absolute stoploss and take-profit (or roi) values in dataframe, having them calculated for each buy signal. This would be extremely handy for less-technical users. This does not work well with backtesting however. Might be a feature worthy of another PR.

docs/strategy-advanced.md
rokups: What do you mean it cant happen due to last condition? You just said it can happen 👀 I did not think about old-enough trades.. Which makes me wonder if this is a robust solution at all. Old way (copying columns) also suffered from this but.. Be nice if we could 100% reliably access candle data around trades.

freqtrade/strategy/interface.py
rokups: I tried to be 100% mypy-compliant. Returning `None` is merely a convenience of not returning anything from overridden function as opposed to having to return `False` when no sell conditions were met. Saves one statement for the user. Not that it would break if `Optional` was dropped :]

docs/strategy-advanced.md
xmatthias: > What do you mean it cant happen due to last condition

at the end of this function, you're selling if the pair is > 1 day - so it's unlikely that you'll not have enough data for this example if a user copies the whole function (Which i would not rely on ...).

> Which makes me wonder if this is a robust solution at all

guess why i am not a huge fan of this ... as there's a lot of "ifs" that you need to consider.

Using "more" data can be problematic - assume i use a slow trading strategy which keeps trades open for weeks (ignore timeframe for now ...)
The current "live" lookback is 1 call to the exchange's ohlcv / klines endpoint. going back 1 week on 5m candles would mean 2-3 calls - which will slow the data gathering down (by quite some time, if you need to do this for 20 open trades and 5 weeks old trades - and potentially result in DDos problems (considering that binance seems to tighten these limits lately ...) - which obviously can be resolved by using a "slower" ratelimit...

docs/strategy-advanced.md
xmatthias: i'm not sure why / what would not work well for backtesting?

docs/strategy-advanced.md
rokups: > at the end of this function, you're selling if the pair is > 1 day - so it's unlikely that you'll not have enough data for this example if a user copies the whole function (Which i would not rely on ...).

It still can happen though. For example if bot was offline for a longer period of time and then was brought back online this would throw an exception for any unsold trades.

> Using "more" data can be problematic

I can hardly imagine where we would need more data than some candles around buy signal or trade open. It would probably cover 99% of usecases if data around those times is available, even if there is a hole between that time and latest candles.

docs/strategy-advanced.md
rokups: Backtester has no notion of stoploss or roi from these functions. What we do is basically emit a sell signal, which follows sell signal rules.

> Sell-signal sells happen at open-price of the consecutive candle

So suppose we have a current in-progress candle that reaches take-profit target, but then falls down and closes below. Since a wick touched our take-profit target a dry run would try to sell that price (correct me if im wrong here). But in backtest sell signal would not even happen because candle closed below take-profit target. Same is true for stoploss. Actually `custom_sell` for stoploss seems like a very bad idea. I think we should purge it from docs. It is both inaccurate and too slow (no stoploss order placed on exchange).

epigramx:
I haven't looked deeply into this so forgive me if I'm way out of tune, but is it mainly like custom_stoploss but without the requirement to trail (so a sort of "custom_roi")? If yes I wonder if all these things could be unified to a custom_sell with a boolean setting if it will trail or not.

rokups:
It looks similar to custom_stoploss yes, but it is a worse version and should not be used as stoploss. I explained why [here](https://github.com/freqtrade/freqtrade/pull/4750#discussion_r620937449).

docs/strategy-advanced.md
xmatthias: well yes, sell-signal and stop / trailing stoploss do serve a different purpose for sure, while the custom_stop function allows you to trail the current price (avoid large pullbacks / sells on large pullbacks) - also with stoploss on exchange in place - custom-sell will definitely not allow this (it's also not the purpose of this function).

you can however still have multiple different sell-reasons - like - you "notice" that the pattern you entered on does not play out - so you bail out even at a loss (avoiding a greater loss via Stoploss).

> So suppose we have a current in-progress candle that reaches take-profit target, but then falls down and closes below.

This is an assumption about movements backtesting does not know about - maybe it reached the take-profit target first, then fell - but maybe it fell first, and then reached the take-profit target (you'd be out of the trade in dry-run in that case too).
Obviously it's an assumption backtesting makes (and is documented), and it's a totally different argument if we assume "price went to high first" or "price went to low first" (green / red candles will not determine this ...)

rokups: I am not having any assumptions where price went first though. Consider the situation where wick touches take-profit target, but candle does not hit stoploss. It does not matter which way price went first because if it went down - nothing would happen anyway, but if it went up - we can take profit. There is no issue with backtesting assumptions at all, its just that such columns could act more like roi and stoploss instead of sell signal.

xmatthias: i'm ... not quite certain what you're referring to with this (probably we should discuss this in chat, as i don't THINK it's related to this PR directly).

If ROI happens (minimal ROI) - then the ROI value is assumed as sell price - not high or low or open.
it's similar with stoploss ... so this column (virtually) exists already ...

xmatthias: I do agree with you though - i think this example should be ... more generic - at least it should not use stoploss / take_profit directly, as that might be missleading.

xmatthias: I've updated the sample now to not use take-profit / stoploss, but to simply use different indicators.

I'm sure others will come up with additional / better samples ... but i think this will do for now.

xmatthias (approved):
Let's merge this and then see what potential problems come up ... 

I've intentionally excluded this from the 2021.4 release - which will give us ~1 month to see what might need improvement.

LGTM :+1:

rokups:
Oh wow, thanks for final touches! I was meaning to get to it on the weekend. Too late :] Also i wasnt sure where to add tests, and at least [one old test](https://github.com/freqtrade/freqtrade/pull/4750/files#diff-98bfe78f0e8d895dd454b5eef6f6485a80526776ad3b6525a14117029bd1eacbL159-R163) might need extra touching up. Anyhow, this is very exciting! Thank you! 🙏🏻

docs/strategy-advanced.md
```diff
@@ -42,32 +42,38 @@ class AwesomeStrategy(IStrategy):
...
+            if trade_row['rsi'] < 80:
+                return 'rsi_below_80'
+
+        # Between 2% and 10%, sell if EMA-long above EMA-short
+        if 0.02 < current_profit < 0.1:
+            if trade_row['emalong'] > trade_row['emashort']:
+                return 'ema_long_below_80'
+
+        # Sell any positions at a loss if they are held for more than two days.
+        if current_profit < 0.0 and (current_time - trade.open_date_utc).days >= 1:
```
rokups, line 72: We missed a comment. When you get to it, could you please fix it to say "more than one day"?

xmatthias: aaaah, nice catch 👍 
addressed in 5bc908870f124f27fe16f890b895da8af970c2f9

xmatthias:
> least one old test might need extra touching up. 

i'm not sure why ...? the change you did to that test seems fine to me?

rokups:
Old test was testing `SellCheckTuple(sell_flag=None, sell_type=SellType.NONE)` which can no longer be expressed in current api. I thought that now maybe this is testing one thing few times, a redundant test. But if you think thats fine then thats fine. I still find tests confusing 👀

xmatthias:
it's testing the behavior (doing multiple things in a specific succession - simulating a real trade iteration) - not this value in particular :)

rokups:
Hey i wonder.. We keep using `custom_sell()` as custom roi function. It is not exactly as accurate as roi table though. what do you think if `custom_sell` could return a tuple of `sell_reason, price_hint`, and bot would use `price_hint` as sell price instead? so we would have something like this:

```py
    def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
                    current_profit: float, **kwargs) -> Optional[Union[str, bool]]:
        if current_rate >= target_rate:
            return 'custom_roi', target_rate
```

Bot then would record `target_rate` as sell price. Second return value would be entirely optional and act as a hint to backtesting only. What do you think?

xmatthias:
well - no?

i mean - we should have a callback to determine the sell price - but i don't think doing that in `custom_sell` is a good idea.
To properly do this, we should have a "sell price suggestion" passed in - so you can move it up / down a bit eventually.

Otherwise you'll end up with a sell-price that's way off spot because you're using some random value that doesn't reflect where market is trading "now".

Also - i'm not determined yet if this "price-selection" should be supported in backtesting - as it can lead to very wild (manipulated) results - which will never fill in a real scenario.

rokups:
I am assuming backtesting always goes like O->L->H->C, so price hint (wherever it comes from) would always have to be smaller than high of the candle and it would only be taken into account only if low of the candle does not hit stoploss. How could people even abuste it to get artificially good results?

xmatthias:
honestly - it's not so much about cheating (you can't cheat reality anyway), but about interface design.

Combining "pricing" with `custom_sell` might seem like a good idea at first (because you need it right now) - but it will force people who don't want to use `custom_sell` signals to implement (and also provide a valid sell-reason!) even though they only want to slightly modify the prices used to place orders.

At the same time - you'll also force people who are fine with the "default pricing options" to return a price - even though they would not care.
So having these 2 seperate will be more flexible, and mainly less complicated to use and correctly return

rokups:
My idea was having a hint price as optional second value, but yeah you are right, another separate method does make more sense 👍🏻 something like `custom_roi_target(trade, ...)` i guess.

xmatthias:
it's got nothing to do with ROI (that's again your usecase - not a interface)

it's more like `get_sell_price()` or similar so you can place your order at a "random" spot in the orderbook you see fit ... although naming can be debated :laughing:

PR #4767, opened 2021-04-20
setup.sh: Install libpython3-dev on Debian/Ubuntu

bzed:
## Summary
Python.h is required to build c modules for Python. Install libpython3-dev on Debian/Ubuntu.

## Quick changelog

- Ensure Python c modules can be build from setup.sh/pip3

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -138,7 +138,7 @@ function install_macos() {
 # Install bot Debian_ubuntu
 function install_debian() {
     sudo apt-get update
-    sudo apt-get install -y build-essential autoconf libtool pkg-config make wget git
+    sudo apt-get install -y build-essential autoconf libtool pkg-config make wget git libpython3-dev
     install_talib
 }
 


setup.sh
```diff
@@ -138,7 +138,7 @@ function install_macos() {
 # Install bot Debian_ubuntu
 function install_debian() {
     sudo apt-get update
-    sudo apt-get install -y build-essential autoconf libtool pkg-config make wget git
+    sudo apt-get install -y build-essential autoconf libtool pkg-config make wget git libpython3-dev
```
xmatthias, line 141: While i'm aware of this - i'm not sure if we should be adding this back (we used to have this in the setup.sh file before).

This makes ... certain assumptions about your python installation (it will break if it's not a package manager based installation), and it's not required in all cases (depending on the availability of wheels for your python version).

xmatthias (approved):

xmatthias:
LGTM :+1:

PR #4768, opened 2021-04-20
Fix incorrect message when protection is fired

c0d5x:
## Summary
max_drawdown_protection.py prints an invalid message when it fires.


xmatthias (approved):
Great spot! :+1:

PR #4776, opened 2021-04-21
Create config_ftx.json.example

tapedarms:
to stop the dl trades error from popping up all the time

## Summary

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

### DIFF
--- a/config_ftx.json.example
+++ b/config_ftx.json.example
@@ -1,0 +1,99 @@
+{
+    "max_open_trades": 3,
+    "stake_currency": "BTC",
+    "stake_amount": 0.05,
+    "tradable_balance_ratio": 0.99,
+    "fiat_display_currency": "USD",
+    "timeframe": "5m",
+    "dry_run": true,
+    "cancel_open_orders_on_exit": false,
+    "unfilledtimeout": {
+        "buy": 10,
+        "sell": 30
+    },
+    "bid_strategy": {
+        "ask_last_balance": 0.0,
+        "use_order_book": false,
+        "order_book_top": 1,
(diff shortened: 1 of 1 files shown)


config_ftx.json.example
```diff
@@ -0,0 +1,99 @@
...
+            "EOS/BTC",
+            "ETH/BTC",
+            "IOTA/BTC",
+            "LINK/BTC",
+            "LTC/BTC",
+            "NEO/BTC",
+            "NXS/BTC",
+            "XMR/BTC",
+            "XRP/BTC",
+            "XTZ/BTC"
```
xmatthias, line 55: In general, i'd say "great" ... 
except that it doesn't work because most of these pairs don't exist on FTX.

I'd maybe change the stake to either USD or USDT and "handpick" a few of the pairs.

config_ftx.json.example
```diff
@@ -0,0 +1,99 @@
...
+        ],
+        "pair_blacklist": [
+            "BNB/BTC"
```
xmatthias, line 58: Blacklisting BNB does not make sense for FTX (it's here for binance - because on binance it'll cause troubles because of fees being taken away from the BNB trade)

xmatthias (commented):
You should be able to make changes to this (to make the config work by default) by simply adding commits to this branch in your fork.

xmatthias (approved):
As you seem to not allow changes to your branch, i'm going to merge this and then do the necessary changes in develop.

LGTM (except for the pairlist, obviously).

PR #4777, opened 2021-04-21
Add the reason why there is a global pairlock when lock is available

c0d5x:
When the Global pairlock is active it will not print why, so with this PR it will include the reason why there is a global pairlock.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -378,7 +378,7 @@ class FreqtradeBot(LoggingMixin):
             if lock:
                 self.log_once(f"Global pairlock active until "
                               f"{lock.lock_end_time.strftime(constants.DATETIME_PRINT_FORMAT)}. "
-                              "Not creating new trades.", logger.info)
+                              f"Not creating new trades, reason: {lock.reason}.", logger.info)
             else:
                 self.log_once("Global pairlock active. Not creating new trades.", logger.info)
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -378,7 +378,7 @@ def enter_positions(self) -> int:
             if lock:
                 self.log_once(f"Global pairlock active until "
                               f"{lock.lock_end_time.strftime(constants.DATETIME_PRINT_FORMAT)}. "
-                              "Not creating new trades.", logger.info)
+                              f"Not creating new trades, reason: {lock.reason}.", logger.info)
```
xmatthias, line 381: while i do understand the reason for this (clarity / show it all the times in the logs) ... do we really need this?

The lock reason is logged / shown when the lock is added - ain't that enough?

if you think it isn't - would it make sense to add the same to the "per pair" log output?

c0d5x: I understand the reason is printed when the lock is created, but the lock lasts a long time, so when the logs are checked what you find is that Global pairlock is active, and not the reason. You can go back in the logs and find multiple locks, but it is much more informative if you find the current reason, instead of checking all the locks to find the one that is current.

xmatthias (approved):

xmatthias (approved):
LGTM :+1:

PR #4780, opened 2021-04-22
Add --new-pairs-days parameter for download-data command.

rokups:
This parameter allows us to customize a number of days we would like to download for new pairs only. This allows us to achieve efficient data update, downloading all data for new pairs and only missing data for existing pairs. To do that use `freqtrade download-data --new-pairs-days=3650` (not specifying `--days` or `--timerange` causes freqtrade to download only missing data for existing pairs).

### DIFF
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -345,6 +345,13 @@ AVAILABLE_CLI_OPTIONS = {
         type=check_int_positive,
         metavar='INT',
     ),
+    "new_pairs_days": Arg(
+        '--new-pairs-days',
+        help='Download data of new pairs for given number of days. Default: `%(default)s`.',
+        type=check_int_positive,
+        metavar='INT',
+        default=30,
+    ),
     "download_trades": Arg(
         '--dl-trades',
         help='Download trades instead of OHLCV data. The bot will resample trades to the '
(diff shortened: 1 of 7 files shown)


freqtrade/commands/cli_options.py
```diff
@@ -345,6 +345,13 @@ def __init__(self, *args, **kwargs):
         type=check_int_positive,
         metavar='INT',
     ),
+    "new_pairs_days": Arg(
+        '--new-pairs-days',
+        help='Download data of new pairs for given number of days. Default: `%(default)s`.',
+        type=check_int_positive,
+        metavar='INT',
+        default=30,
```
xmatthias: instead of defining the default here - we should rather move this to `constants.py` and add the default in the json validation.

This will allow providing this value via config as well - otherwise the "argument default" from the CLI will always overwrite what is in the config file

rokups: Oh ok! So this is why putting it in config didnt work

xmatthias: yeah, argparse default arguments are good if you ONLY use argparse - but if you use a default, they'll ALWAYS have this value - so there's no way to determine if a user provided that explicitly, or if it's coming as default (at least not without going to parse the provided arguments manually, which defeats the purpose of using argparse ...)

xmatthias (approved):
i've done a small commit (doing the change was faster than explaining what i wanted) :laughing: 

Great! LGTM :+1: 

Thanks a lot!

Issue #4778, opened 2021-04-21
get recent data on hitbtc


saeedrss:
## Describe your environment

  * Operating system: ____ win 10 

  * Python Version: _____ Python 3.7.8

  * CCXT version: _____ ccxt 1.46.38

  * Freqtrade Version: ____ freqtrade develop

 
## Your question

hello  i use  hitbtc  exchange and i got :
 WARNING - Outdated history for pair BTC/USDT. Last tick is 3811501 minutes old

hitbtc by default send candle from beginning (not most recently)
in ccxt.hitbtc  fetch_ohlcv/fetchOHLCV   get candle from beginning and  fetchOhlcvc/fetchOHLCVC  get the recent

i can't set config file 

oder exchange  work fine

![Screenshot 2021-04-22 012639](https://user-images.githubusercontent.com/6265785/115623862-004fd180-a30f-11eb-8593-c1c456068641.png)

*Ask the question you have not been able to find an answer in our [Documentation](https://www.freqtrade.io/en/latest/)*

xmatthias:
closed by #4788

PR #4788, opened 2021-04-23
fetch live data from hitbtc exchange #4778

saeedrss:

## Summary

Solve the issue: #4778

## Quick changelog
 I just fixed  hitbtc  ccxt fetch_ohlcv format 

## What's new?

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -874,8 +874,15 @@ class Exchange:
                 "Fetching pair %s, interval %s, since %s %s...",
                 pair, timeframe, since_ms, s
             )
-
-            data = await self._api_async.fetch_ohlcv(pair, timeframe=timeframe,
+			#fixing support for HitBTC #4778
+            if self.name== 'HitBTC':
+                data = await self._api_async.fetch_ohlcv(pair, timeframe=timeframe,
+                                                        since=since_ms,
+                                                        limit=self.ohlcv_candle_limit(timeframe),
+                                                        params={"sort": "DESC"}
+                                                         )
+            else:
+                data = await self._api_async.fetch_ohlcv(pair, timeframe=timeframe,
                                                      since=since_ms,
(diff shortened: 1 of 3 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -874,8 +874,15 @@ async def _async_get_candle_history(self, pair: str, timeframe: str,
...
-
-            data = await self._api_async.fetch_ohlcv(pair, timeframe=timeframe,
+			#fixing support for HitBTC #4778
+            if self.name== 'HitBTC':
+                data = await self._api_async.fetch_ohlcv(pair, timeframe=timeframe,
+                                                        since=since_ms,
+                                                        limit=self.ohlcv_candle_limit(timeframe),
+                                                        params={"sort": "DESC"}
+                                                         )
+            else:
```
xmatthias: While i'm sure this will technically work - this is not how this should be done (we'd have 100s of if `exchange=='whatever'` statements ...

Please subclass exchange for hitbtc to fix exchange-specific topics (look at `ftx.py` or `kraken.py` or `bittrex.py` for examples).
Note that you'll also have to add the export to `freqtrade/exchange/__init__.py` to have this subclass used.

From there, i think there's 2 possibilities.
Either, use `_ft_has_default` and `_ft_has` to specify additional parameters for fetch_ohlcv (so this can be configured for other exchanges as well) - defaulting to an empty dict (or `None`) - or alternatively implement `_async_get_candle_history` in the subclass.

xmatthias (changes requested):

freqtrade/exchange/exchange.py
saeedrss: ok i got it 👌
thanks for hint

xmatthias (approved):
I've adjusted it the way i though it should be.

it works for hitbit and has no impact to other exchanges this way (unless some configs need changing).

PR #4790, opened 2021-04-24
add IntParameter.range

xmatthias:
## Summary
Improve Auto-Hyperopt parameters by having them "hyperopt-aware".

This way, IntParameter.range can return 1 value for regular modes, and the whole space while hyperopting.
This will make this performance efficient without manually switching things back and forth.

## Quick changelog

- improve docs to optimize indicator parameters
- add IntParameter.range property to iterate over the space easily

### DIFF
--- a/docs/hyperopt.md
+++ b/docs/hyperopt.md
@@ -165,10 +165,21 @@ Rarely you may also need to create a [nested class](advanced-hyperopt.md#overrid
 !!! Tip "Quickly optimize ROI, stoploss and trailing stoploss"
     You can quickly optimize the spaces `roi`, `stoploss` and `trailing` without changing anything in your strategy.
 
-    ```python
+    ``` bash
     # Have a working strategy at hand.
     freqtrade hyperopt --hyperopt-loss SharpeHyperOptLossDaily --spaces roi stoploss trailing --strategy MyWorkingStrategy --config config.json -e 100
     ```
+
+### Hyperopt execution logic
+
+Hyperopt will first load your data into memory and will then run `populate_indicators()` once per Pair to generate all indicators.
+
+Hyperopt will then spawn into different processes (number of processors, or `-j <n>`), and run backtesting over and over again, changing the parameters that are part of the `--spaces` defined.
+
+For every new set of parameters, freqtrade will run first `populate_buy_trend()` followed by `populate_sell_trend()`, and then run the regular backtesting process to simulate trades.
+
+After backtesting, the results are passed into the [loss function](#loss-functions), which will evaluate if this result was better or worse than previous results.  
+Based on the loss function result, hyperopt will determine the next set of parameters to try in the next round of backtesting.
 
 ### Configure your Guards and Triggers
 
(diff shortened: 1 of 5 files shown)


docs/hyperopt.md
```diff
@@ -165,11 +165,22 @@ Rarely you may also need to create a [nested class](advanced-hyperopt.md#overrid
...
+
+Hyperopt will first load your data into memory and will then run `populate_indicators()` once per Pair to generate all indicators.
+
+Hyperopt will then spawn into different processes (number of processors, or `-j <n>`), and run backtesting over and over again, changing the parameters that are part of the `--spaces` defined.
+
+For every new set of parameters, freqtrade will run first `populate_buy_trend()` followed by `populate_sell_trend()`, and then run the regular backtesting process to simulate trades.
+
+After backtesting, the results are passed into the [loss function](#loss-functions), which will evaluate if this result was better or worse than previous results.  
+Based on the loss function result, hyperopt will determine the next set of parameters to try in the next round of backtesting.
+
```
TheJoeSchr, line 183: Love the more in-depth explaination

docs/hyperopt.md
```diff
@@ -260,6 +266,8 @@ The last one we call `trigger` and use it to decide which buy trigger we want to
 So let's write the buy strategy using these values:
 
 ```python
+
+
```
TheJoeSchr: Needed?

docs/hyperopt.md
```diff
@@ -314,6 +322,87 @@ There are four parameter types each suited for different purposes.
...
+from freqtrade.strategy import IStrategy
+from freqtrade.strategy import CategoricalParameter, DecimalParameter, IntParameter
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+
+class MyAwesomeStrategy(IStrategy):
+    stoploss = 0.5
+    timeframe = '15m'
+    # Define the parameter spaces
+    buy_ema_short = IntParameter(3, 50, default=5)
+    buy_ema_long = IntParameter(15, 200, default=50)
```
TheJoeSchr, line 341: This made me think of a common missing feature I come across:
Range is 15-200, but not every single value in between this range is valuable. So a `step` or `distance` parameter like with `FloatParameter` would be extremely useful...

docs/hyperopt.md
```diff
@@ -314,6 +322,87 @@ There are four parameter types each suited for different purposes.
...
+While this strategy is most likely too simple to provide consistent profit, it should serve as an example how optimize indicator parameters.
+
+!!! Note
+    `self.buy_ema_short.range` will act differently between hyperopt and other modes. For hyperopt, the above example may generate 48 new columns, however for all other modes (backtesting, dry/live), it will only generate the column for the selected value. You should therefore avoid using the resulting column with explicit values (values other than `self.buy_ema_short.value`).
+
+??? Hint "Performance tip"
+    By doing the calculation of all possible indicators in `populate_indicators()`, the calculation of the indicator happens only once for every parameter.  
+    While this may slow down the hyperopt startup speed, the overall performance will increase as the Hyperopt execution itself may pick the same value for multiple epochs (changing other values).
+    You should however try to use space ranges as small as possible. Every new column will require more memory, and every possibility hyperopt can try will increase the search space.
+
```
TheJoeSchr, line 405: Ah, now I understand, great job 🤗

TheJoeSchr (approved):
Lgtm

docs/hyperopt.md
rokups: You can do stepping:
```python
step = 2
buy_ema_short = IntParameter(2 / step, 50 / step, default=5)
    for val in self.buy_ema_short.range:
        dataframe[f'ema_short_{val}'] = ta.EMA(dataframe, timeperiod=val*step)
```

xmatthias: while it works, you must make sure to use it in all cases ... so i'd consider this as "dangerous" for now :laughing:

docs/hyperopt.md
xmatthias: obviously not :laughing:

docs/hyperopt.md
epigramx: ^True, and if you do use the workaround then the hyperopt results might need manual multiplication.
Cool feature if it existed.

PR #4791, opened 2021-04-24
reports: add avg. profit and median profit and a quick summary block

TheJoeSchr:
Someone pointed out that they liked my customization of backtest results, so I figure I share it


TheJoeSchr:
Example pic:
image

PR #4792, opened 2021-04-24
add distribution graph to example notebook

TheJoeSchr:
image


xmatthias (approved):
LGTM :+1: 

in general, when updating the jupyter notebooks, please also run the commands to update the output as documented [here](https://www.freqtrade.io/en/latest/developer/#updating-example-notebooks).

I've done that now - so the docs will also contain this new part of the code.

TheJoeSchr:
Gotcha


__

PR #4798, opened 2021-04-25
fix docstring typo in sample_strategy.py

wr0ngc0degen:
## Summary
fix docstring typo in sample_strategy.py

## Quick changelog

- corrected docstring for populate_sell_trend in sample_strategy.py


wr0ngc0degen:
I think the builds are failing cause Binance is upgrading spot trading system right now
https://www.binance.com/en/support/announcement/849160fe70214641baa6385619595aa1
(I was trying to run a backtest with binance and it's also failing cause _whitelist_for_active_markets filters out all the pairs from the static whitelist provided due to the market not being active)

xmatthias (approved):
Yes they are - there was however ALSO a non-mocked request which failed erroneously (but should not fail).

As we are however testing certain tests against the exchange, i fully expect these to fail at the moment - but that's not related to your PR ... 

LGTM - nice catch! :+1:

Issue #4519, opened 2021-03-11
Telegram message length limit


Pytness:
## Describe your environment

  * Operating system: Manjaro
  * Python Version: Python 3.9.2
  * CCXT version: ccxt==1.42.66
  * Freqtrade Version: freqtrade develop-ef9977fc

## Describe the problem:

When sending commands (/edge and potentially others) to the telegram bot, the bot does not respond due to the message being larger than the length limit.

The following errors are logged:

```
WARNING - Telegram NetworkError: Message is too long! Trying one more time.
WARNING - TelegramError: Message is too long! Giving up on that message.
  ```

## Related issues
#1745
#1749

**Comment by Pytness** [comment]
Maybe, a form of text paging could be implemented right into _send_msg, prefixing the page number to the text, for example:
```
[01/02]
some long text
...
```
```
[02/02]
the end of the text
```

Although this would probably break some of the HTML-based messages.

xmatthias:
for some messages, pagination is implemented (like /balance).

it's definitely not the case for all messages, as it complicates the output quite a bit (you need to search for a good breakpoint, ...) - and it doesn't seem that common.

PR #4836, opened 2021-04-29
Telegram rpc: split too long /locks messages

bzed:
## Summary
Partly addressing #4519, I think it is easier to split arrays into equally sized parts before they go into tabulate, producing better readable (instead of super long) messages. Implemented for /locks as that list was too long for telegram. I think the same approach could be used for all telegram messages that use tabulate and produce a long output. Other ideas/suggestions welcome, I'd be happy to update the PR.

## Quick changelog

- Add a misc.chunks function to split arrays into chunks
- Split the /locks message for Telegram into chunks of 25 lines per message.

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -20,7 +20,7 @@ from telegram.utils.helpers import escape_markdown
 from freqtrade.__init__ import __version__
 from freqtrade.constants import DUST_PER_COIN
 from freqtrade.exceptions import OperationalException
-from freqtrade.misc import round_coin_value
+from freqtrade.misc import chunks, round_coin_value
 from freqtrade.rpc import RPC, RPCException, RPCHandler, RPCMessageType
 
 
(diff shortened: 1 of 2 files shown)


freqtrade/rpc/telegram.py
```diff
@@ -750,17 +750,18 @@ def _locks(self, update: Update, context: CallbackContext) -> None:
...
-            lock['pair'],
-            lock['lock_end_time'],
-            lock['reason']] for lock in locks['locks']],
-            headers=['ID', 'Pair', 'Until', 'Reason'],
-            tablefmt='simple')
-        message = f"<pre>{escape(message)}</pre>"
-        logger.debug(message)
-        self._send_msg(message, parse_mode=ParseMode.HTML)
+        rpc_locks = self._rpc._rpc_locks()
+        for locks in chunks(rpc_locks['locks'], 25):
```
xmatthias, line 754: what's the 25 based on? random number or experimentation?

I don't mind having this in particular - but i'm asking because telegram has a pretty well-defined message length (4096) (which obviously applys AFTER the formatting) - so i'm not sure if an arbitrary 25 lines will work reliably...

bzed: My goal was to prouce more readable (as in: not tooo long) message instead of tracking the largest possible message size.
Maybe a mix of limits would be the best idea? Splitting tabulates into messages that are smaller than the max message size but also have a maximum number of lines. Maybe it would even make sense to have this configurable? I'm using telegram on my phone, other people might use a client on their computer with a much bigger screen. Wouldn't mind to implement that.

xmatthias: i'm not sure if a configuration will help much.

Splitting it after 25 lines will be fine if you have longer output - it'll only be slightly strange if you have 26 or 27 lines - so the 2nd message is rather short i guess.

if you have it configurable however you'll not gain much (other than configuration complexity) - as i can set it to 26 - but then i have 27 lines - or i configure it to 27, and get 28 lines :laughing: 

---

I guess the only other messages that can be impacted / relevant for this are `/trades` and `/edge` (although this is slightly different as it tabulates a dict, not a list).
for `/performance` the splitting is already done - but based on "max-messagelength" - not per line.

I'll however leave it up to you if you wanna update this PR, or do these seperate and we merge this one "as is" ...

xmatthias (approved):
LGTM :+1: 

Let's merge it this way - and eventually work on further messages in seperate PR's.
Hope this works for you - and a big Thanks :100:  for your involvement and contribution!

PR #4838, opened 2021-04-29
Fix bug in running hyperopt in windows 10

Antreasgr:
with preferred encoding in locale differrent from
utf-8

## Summary
Fix an error in running hyperopt on windows when the default locale isn't english


freqtrade/__init__.py
```diff
@@ -1,6 +1,11 @@
 """ Freqtrade bot """
 __version__ = 'develop'
 
+import locale
+def getpreferredencoding(do_setlocale=True):
+    return "utf-8"
+locale.getpreferredencoding = getpreferredencoding
+
```
xmatthias: I'm ... not convinced about this for 2 reasons...
first, i don't understand the impact this would have / not having this has...

Then, i wouldn't patch it here (globally) for sure. 
this can have an impact on every library that potentially uses this.

Antreasgr: Couldn't agree more, but i am not familiar with freqtrade codebase to create a more robust PR for this, just raising attention to the issue, because it prevents from running hyperopt entirely on different locales, and this is what i use locally to get it to work without changing global windows 10  setting to en-US locale. I believe someone more familiar with the freqtrade code can create something much better

xmatthias: Can you maybe provide me with screenshots / output of what the problem looks like when you're running on your PC - as well as the output of the following in a regular python shell - and which locale you're using (or if you're uncomfortable sharing this, a locale that also causes this problem ...)

```python
import  locale
print(locale.getpreferredencoding)
```

Antreasgr: Here you are:
Screenshot 2021-04-30 175857
Screenshot 2021-04-30 180041

xmatthias: I've now pushed a different attempted fix to this branch.

Now i struggle to test this with an environment where this is actually a problem - would you mind giving this change a try and report back if it works?

xmatthias: hm, seeing that windows CI is now failing i will need to think of this ... :laughing:

Antreasgr: Yes this works well

Antreasgr: I see something in the test enviroment about cp1252 encoding. Could the default test enviroment be running on that? I can totally confirm it is running on my setup with python 3.9.1 x64

xmatthias (approved):
I've taken a slightly different approach to fixing this now by simply removing this sign from the output.

the benefit it provides does not outweight the issues and compatibility issues it seems to create in specific environments.
Hope this works for you too.

PR #4841, opened 2021-04-30
fix IStrategy: abstract methods still need to pass through return value

TheJoeSchr:
## Summary

abstract methods still need to pass through return value
otherwise doing something like:
```py
dataframe = super().populate_indicators(dataframe, ...)
```

won't work, because `dataframe` becomes `None`.

This is needed if one of those methods uses `dataframe.copy()` instead of
just working on reference. e.g. using `merge_informative` in
`populate_indicator` in a nested class hierarchy

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -161,6 +161,7 @@ class IStrategy(ABC, HyperStrategyMixin):
         :param metadata: Additional information, like the currently traded pair
         :return: a Dataframe with all mandatory indicators for the strategies
         """
+        return dataframe
 
     @abstractmethod
     def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
@@ -170,6 +171,7 @@ class IStrategy(ABC, HyperStrategyMixin):
         :param metadata: Additional information, like the currently traded pair
         :return: DataFrame with buy column
         """
+        return dataframe
 
     @abstractmethod
     def populate_sell_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
@@ -179,6 +181,7 @@ class IStrategy(ABC, HyperStrategyMixin):
         :param metadata: Additional information, like the currently traded pair
         :return: DataFrame with sell column
         """
+        return dataframe
 
     def check_buy_timeout(self, pair: str, trade: Trade, order: dict, **kwargs) -> bool:
         """


xmatthias:
why would you call `super().populate_indicators` in the first place?

There's no reason to do that - at least none i would know.
It's an abstract method, which means the expectation is that a child class MUST implement this - and that the parent class does not provide any implementation for this - so calling it is nonesense.

TheJoeSchr:
I will try to come up with a simple example, but short answer is: Mixins
and Inheritance.



There is always the possibility that there is more than one child class, so
calling a super() method is far from nonsense...?



And with multiple inheritance which you activity enable by using ABC and
provide an abstract base class, it is kind of implied that you may or may
not know the order in which are classes nested and called. Therefore you
need to call your super() and handle return values of it.



I really don't see how you could even write that with a straight face?



I refer to the python documentation for a refresher:
https://docs.python.org/3/library/abc.html#abc.abstractmethod

xmatthias:
>  Unlike Java abstract methods, these abstract methods may have an implementation.

so they MAY have an implementation - which means it's 50:50 they don't.
the doc you link to kindof enhances my point, not yours.

if you'll note however - i've only asked a question about the PR (i don't blindly merge anything provided - but try to understand why people think it's necessary).
if you implement a mixin you will be calling the mixin's method, NOT the abstract method.

I'm however sceptical about mixins in user strategies anyway - loading from other files (not installed modules) is kinda difficult in strategies ...

TheJoeSchr:
> if you'll note however - i've only asked a question about the PR

Just mirroring the tone, I give as I get....



> if you implement a mixin you will be calling the mixin's method, NOT the abstract method.

yeah but what if I have several Mixins with the same methods. They don't know the order in which they are called so they need to call the super() method. 

I used it to quickly Mix and Change the behavior of `custom_stoploss` (and now `custom_sell`). For some I may need to call `populate_indicator` which may need to call `merge_informative` and therefor don't work on the same reference of Dataframe as provided per reference. 

Simplified contrived Example:
```py
class BaseStrategy(IStrategy):
    # does stuff
    def __init__(self, config: dict) -> None:
        super().__init__(config)

class WaitForTakeProfitStoplossMixin(BaseStrategy):
        dataframe = super().populate_indicators(dataframe, metadata)
        dataframe['min'] = dataframe['low'].rolling(10).min()

        informative = self.dp.get_pair_dataframe(pair=metadata['pair'], timeframe=inf_tf)
        informative['rsi7'] = ta.RSI(informative, timeperiod=7)

        dataframe = merge_informative_pair(dataframe, informative, self.timeframe, inf_tf, ffill=True)
        return dataframe

    def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime, current_rate: float, current_profit: float, **kwargs) -> float:
        #... do stuff
        return result

    # ...
class DecayingStoplossMixin(OtherMixinNotShownHere):
        dataframe = super().populate_indicators(dataframe, metadata)
        dataframe['max'] = dataframe['high'].rolling(10).max()

        informative = self.dp.get_pair_dataframe(pair="BTC/USDT", timeframe=inf_tf)
        informative['mfi'] = ta.MFI(informative, timeperiod=7)

        dataframe = merge_informative_pair(dataframe, informative, self.timeframe, inf_tf, ffill=True)
        return dataframe

    def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime, current_rate: float, current_profit: float, **kwargs) -> float:
        #... do stuff
        return result

    # ...

class AwesomeStrategy(WaitForTakeProfitStoplossMixin,DecayingStoplossMixin):
        dataframe = super().populate_indicators(dataframe, metadata)

```

I'm open for other ideas and not insisting on using a return in those @abstractmethod, but I think it's the common and correct technical implementation to do it so it works with multiple overrides.

xmatthias (approved):
There's no harm in this.
LGTM :+1: 

you should however be careful with the approach you're showing. 

This can lead to a big performance decrease (every "mixin" loading the same informative pair for example, or calculating the same indicators) and using multiple different mixins with this approach can result in colliding namings in the columns (both calculate RSI as `dataframe['rsi']` - but one uses a timeframe of 14, the other of 50 ... 

it'll be "the last one called wins" - but it'll not result in the desired results i'd assume, as i'm sure there is usually a reason for the way the indicator is picked.

TheJoeSchr:
Thank you and don't worry, I use better names for this in my strategy, just removed it for conciseness.

PR #4843, opened 2021-05-01
Docker support for arm64

docbobo:
## Summary
Building arm64 images with the provided docker files is not possible. This PR provides the changes necessary to do so.

## What's new?
Currently, the repo provides Dockerfiles that allow building on either x86 systems or 32-bit RPi. Trying to execute either on arm64 does not work. First of all, building ta-lib does not work out of the box because it cannot guess the build platform correctly. Afterwards, build fails at several places due to the fact that wheels for arm64 don't seem to be readily available and have to be compiled. Most of this is addressed by add a few additional packages upfront. While `gcc` is not required for the build, it allows the use of `Dockerfile.jupyter` later on, which will fails otherwise.

Build time can be huge on low end devices. On a Jetson Nano, the total build time was ~4600 seconds. On an Apple Mac Mini with M1 processor however, the build only took 756 seconds in total.

Lastly, the PR contains a minor change that will use all processor cores when building ta-lib.


xmatthias (changes requested):
In Principle, i think this is great (i can't test it due to lack of ARM64 hardware - so i have to trust you on this).

I only have 2 (rather small i hope) requests:
* can we move the Dockerfile to the docker directory? i find the root already bloated (and i've been thinking about doing the same for the armhf image - but that will need more considerations so i'd do that later).
* can you add this to the docs? otherwise we have it - but noone will know about it :laughing:

docbobo:
Well, the first one should be fairly easy - I had put the Dockerfile where I found the one for the Raspberry. I'll be happy to put it somewhere else.

For the second request: is there any way to automatically build the arm64 version? In that case, there is hardly anything to document...

xmatthias:
> is there any way to automatically build the arm64 version? In that case, there is hardly anything to document...

in theory it can be build in ci - however - it'll be build on a x86 machine with buildx (similar to the raspberry version) - and building it completely cannot take longer than (i think 1 hour) - otherwise CI will timeout.

That's been a dealbreaker for the Raspberry for a long time (that's why we use piwheels now - but they don't provide wheels for arm64) - and i fear it'll break the "fun" in this case as well.

docbobo:
Yeah, that's definitely a problem. Let me see what I can do about the documentation.

docbobo:
I am little uncertain where to put the docs, to be honest. The docker section is all about running existing images, while the manual install section doesn’t seem to talk about docker. 

Any suggestions?

xmatthias:
Maybe add a 3rd tab [here](https://www.freqtrade.io/en/latest/docker_quickstart/#docker-quick-start) - so it's "PC / Mac/linux | RaspberryPi | ARM64" or something similar ?

xmatthias (approved):
I trust that the docker image will build on a M1 mac (i can't really test this).

Thanks a lot for contributing this back so others can benefit, too! :+1: 
LGTM

tomismrkolj:
I tried to build it on RPi4. It failed at: 
Step 12/26 : COPY build_helpers/* /tmp/
COPY failed: no source files were specified

xmatthias:
@tomismrkolj how?

you'll need to build it from the root with the compose file (as shown in the develop docs).

if you build it without docker-compose, you'll need to make sure you're in the right folder (it seems like you cd'd into the docker directory, which for sure will not work.

tomismrkolj:
Yes, i used docker compose as usual: 
Copied Dockerfile.aarch64 to ft_userdata folder (where i had previous install). 
Modified docker-compose.yml:
---
version: '3'
services:
  freqtrade:
    #image: freqtradeorg/freqtrade:develop_pi
    # Use plotting image
    # image: freqtradeorg/freqtrade:develop_plot
    # Build step - only needed when additional dependencies are needed
    image: freqtrade_arm64
    build:
      context: .
      dockerfile: "Dockerfile.aarch64"
    restart: unless-stopped
    container_name: freqtrade2

ran docker-compose build freqtrade
and got the error above after some build steps were already finished.

docbobo:
Let me try building it exactly like that…

docbobo:
Oh, wait a second. The ARM64 docker image needs to be build from source. Just copying the Dockerfile.aarch64 to your ft_userdata will not suffice.

Instead you need to clone the repository and run it from within there. 

@xmatthias you were absolutely right, this does require additional documentation. Just didn’t have time yet.

docbobo:
Okay, try this:

```
git clone https://github.com/freqtrade/freqtrade.git
cd freqtrade
docker build -f ./docker/Dockerfile.aarch64 -t freqtradeorg/freqtrade:develop .
```

Afterwards, you should have build a `freqtradeorg/freqtrade:develop` image that you can then reference in the `docker-compose.yml` in your `ft_userdata` folder as described in the documentation.

tomismrkolj:
I suggest you document that in freqtrade quickstart with docker subpage.

docbobo:
Totally agree. And that had been the plan. Any chance you can confirm if this worked for you?

tomismrkolj:
sorry i am not so much into linux.  i can try those three commands, but afterwards....have buikld freqtrade/freqtrade:develop - what should i do?
i guess this procedure is now mix of native building on local RPi and then "importing" it into docker-compose container"?

docbobo:
Yes, running the three commands will give you the develop image. If you are successful with that, running that image is just a matter of changing the image name in your docker-compose. 

Can you validate that the three commands work for you? They do here, on two different platforms, but getting a confirmation from someone without much Linux knowledge would be useful for coming up with instructions…

tomismrkolj:
should i be in any particular folder when i run git clone...?
and should i put path to built image into docker-compose.yml?

tomismrkolj:
and how is then updating the image going be? with all these commands every commit?

docbobo:
Great questions - that's why it's so useful that you are trying this out.

I think you could actually do that within your `ft_userdata` folder. As a result, you will end up with a freqtrade folder next to your user_data folder. The build section in your `docker-compose.yml` should be commented out. If you downloaded the `docker-compose.yml` as described in the instructions for the RPi, it should contain a `image: freqtradeorg/freqtrade:stable` - change that to ` image: freqtradeorg/freqtrade:develop` .

Lastly, if you want to update you need to

```
# Change into folder with the source code
cd freqtrade
# Update local copy of the source code
git pull
# Rebuild the image
docker build -f ./docker/Dockerfile.aarch64 -t freqtradeorg/freqtrade:develop .
```

Keep in mind that at this point in time, all of this is building the current development version, not the stable version.

Note: the first ` docker build`  will probably take a while. Maybe an hour or two. However, subsequent builds should be much faster unless some project dependency was changed. Not sure, how often that happens.

tomismrkolj:
so this procedure is different as in 32 bit version, where everything is already prebuilt for docker on github? it adds certain steps, and certainly adds some time. Is there a plan to have 64 bit arm version prebuilt? I guess it has something to do with missing wheels?
What are the most important benefits in using arm 64 bit version of freqtrade?

tomismrkolj:
right now i am running two further freqtrade containers on same RPi 4. What is going to happen to them if i try build process in the third container while they are running? Will the start to lag due to cpu load of build process?

xmatthias:
Can you just read the documentation, instead suggesting to add it?
I added the additional documentation there  - so simply READING it would suffice.

If you go to [this page](https://www.freqtrade.io/en/latest/docker_quickstart/#docker-quick-start) - and click on ARM64 - you will find all necessary steps to build this image from source.

> right now i am running two further freqtrade containers on same RPi 4. 

This suggests that you are running a 32bit image of the RPI - so the ARM64 image will not work anyway.
We will currently not build the image as wheels are not available, so building it in CI would take too long (i tested it for 32 bit without wheels, and was unsuccessful compiling that image).

Also, you should not use direct docker commands - but instead use the docker-compose file and edit it as stated in the documentation (linked above).

PR #4851, opened 2021-05-02
Data provider support in backtesting

rokups:
## Quick changelog

- Fix `self.dp.get_analyzed_dataframe()` not working in backtesting.
- Remove `dataframe` parameter from `custom_stoploss` and `custom_sell`.
- Add `current_time` parameter to `confirm_trade_entry`/`confirm_trade_exit`.

## What's new?

Seems like i went the wrong way adding `dataframe` as a parameter to `custom_sell` and `custom_stoploss`. After all dataframe is useful in other methods as well. One such very practical usecase i bumped into is blocking roi sells and riding an uptrend.

```py
    def confirm_trade_exit(self, pair: str, trade: 'Trade', order_type: str, amount: float,
                           rate: float, time_in_force: str, sell_reason: str,
                           current_time: 'datetime', **kwargs) -> bool:
        current_time = timeframe_to_prev_date(self.timeframe, current_time)
        dataframe, _ = self.dp.get_analyzed_dataframe(pair, self.timeframe)
        # dataframe = self.dataframe
        current_candle = dataframe.loc[dataframe['date'] == current_time]
        if current_candle is not None and sell_reason == 'roi':
            current_candle = current_candle.squeeze()
            if current_candle['rsi'] > 50:
                return False

        return True
```

Adding dataframe as a parameter to all these methods gets complicated real fast. Since we can use `self.dp` in live, i figured we could also make it work in backtesting.

### DIFF
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -60,7 +60,8 @@ from freqtrade.strategy import IStrategy, timeframe_to_prev_date
 
 class AwesomeStrategy(IStrategy):
     def custom_sell(self, pair: str, trade: 'Trade', current_time: 'datetime', current_rate: float,
-                    current_profit: float, dataframe: DataFrame, **kwargs):
+                    current_profit: float, **kwargs):
+        dataframe = self.dp.get_analyzed_dataframe(pair, self.timeframe)
         trade_open_date = timeframe_to_prev_date(self.timeframe, trade.open_date_utc)
         trade_row = dataframe.loc[dataframe['date'] == trade_open_date].squeeze()
 
@@ -105,8 +106,7 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, dataframe: DataFrame,
-                        **kwargs) -> float:
+                        current_rate: float, current_profit: float, **kwargs) -> float:
         """
         Custom stoploss logic, returning the new distance relative to current_rate (as ratio).
         e.g. returning -0.05 would create a stoploss 5% below current_rate.
@@ -156,8 +156,7 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, dataframe: DataFrame,
-                        **kwargs) -> float:
+                        current_rate: float, current_profit: float, **kwargs) -> float:
 
         # Make sure you have the longest interval first - these conditions are evaluated from top to bottom.
         if current_time - timedelta(minutes=120) > trade.open_date_utc:
@@ -183,8 +182,7 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, dataframe: DataFrame,
-                        **kwargs) -> float:
+                        current_rate: float, current_profit: float, **kwargs) -> float:
 
         if pair in ('ETH/BTC', 'XRP/BTC'):
             return -0.10
@@ -210,8 +208,7 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, dataframe: DataFrame,
-                        **kwargs) -> float:
+                        current_rate: float, current_profit: float, **kwargs) -> float:
 
         if current_profit < 0.04:
             return -1 # return a value bigger than the inital stoploss to keep using the inital stoploss
@@ -250,8 +247,7 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, dataframe: DataFrame,
-                        **kwargs) -> float:
+                        current_rate: float, current_profit: float, **kwargs) -> float:
 
         # evaluate highest to lowest, so that highest possible stop is used
         if current_profit > 0.40:
@@ -293,8 +289,7 @@ class AwesomeStrategy(IStrategy):
     use_custom_stoploss = True
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
-                        current_rate: float, current_profit: float, dataframe: DataFrame,
-                        **kwargs) -> float:
+                        current_rate: float, current_profit: float, **kwargs) -> float:
 
         # Default return value
         result = 1
@@ -302,6 +297,7 @@ class AwesomeStrategy(IStrategy):
             # Using current_time directly would only work in backtesting. Live/dry runs need time to
             # be rounded to previous candle to be used as dataframe index. Rounding must also be 
             # applied to `trade.open_date(_utc)` if it is used for `dataframe` indexing.
+            dataframe = self.dp.get_analyzed_dataframe(pair, self.timeframe)
             current_time = timeframe_to_prev_date(self.timeframe, current_time)
             current_row = dataframe.loc[dataframe['date'] == current_time].squeeze()
             if 'atr' in current_row:
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -61,7 +61,7 @@ from freqtrade.strategy import IStrategy, timeframe_to_prev_date
 class AwesomeStrategy(IStrategy):
     def custom_sell(self, pair: str, trade: 'Trade', current_time: 'datetime', current_rate: float,
                     current_profit: float, **kwargs):
-        dataframe = self.dp.get_analyzed_dataframe(pair, self.timeframe)
+        dataframe, _ = self.dp.get_analyzed_dataframe(pair, self.timeframe)
         trade_open_date = timeframe_to_prev_date(self.timeframe, trade.open_date_utc)
         trade_row = dataframe.loc[dataframe['date'] == trade_open_date].squeeze()
 
@@ -290,7 +290,7 @@ class AwesomeStrategy(IStrategy):
             # Using current_time directly would only work in backtesting. Live/dry runs need time to
             # be rounded to previous candle to be used as dataframe index. Rounding must also be 
             # applied to `trade.open_date(_utc)` if it is used for `dataframe` indexing.
-            dataframe = self.dp.get_analyzed_dataframe(pair, self.timeframe)
+            dataframe, _ = self.dp.get_analyzed_dataframe(pair, self.timeframe)
             current_candle = dataframe.loc[-1].squeeze()
             if 'atr' in current_candle:
                 # new stoploss relative to current_rate
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -265,12 +265,6 @@ class AwesomeStrategy(IStrategy):
 
 Imagine you want to use `custom_stoploss()` to use a trailing indicator like e.g. "ATR"
 
-!!! Warning
-    Only use `dataframe` values up until and including `current_time` value. Reading past
-    `current_time` you will look into the future, which will produce incorrect backtesting results
-    and throw an exception in dry/live runs.
-    see [Common mistakes when developing strategies](strategy-customization.md#common-mistakes-when-developing-strategies) for more info.
-
 !!! Note
     `dataframe['date']` contains the candle's open date. During dry/live runs `current_time` and
     `trade.open_date_utc` will not match the candle date precisely and using them directly will throw
@@ -290,7 +284,6 @@ class AwesomeStrategy(IStrategy):
 
     def custom_stoploss(self, pair: str, trade: 'Trade', current_time: datetime,
                         current_rate: float, current_profit: float, **kwargs) -> float:
-
         # Default return value
         result = 1
         if trade:
@@ -298,11 +291,19 @@ class AwesomeStrategy(IStrategy):
             # be rounded to previous candle to be used as dataframe index. Rounding must also be 
             # applied to `trade.open_date(_utc)` if it is used for `dataframe` indexing.
             dataframe = self.dp.get_analyzed_dataframe(pair, self.timeframe)
-            current_time = timeframe_to_prev_date(self.timeframe, current_time)
-            current_row = dataframe.loc[dataframe['date'] == current_time].squeeze()
-            if 'atr' in current_row:
+            current_candle = dataframe.loc[-1].squeeze()
+            if 'atr' in current_candle:
                 # new stoploss relative to current_rate
-                new_stoploss = (current_rate - current_row['atr']) / current_rate
+                new_stoploss = (current_rate - current_candle['atr']) / current_rate
+
+                # Round trade date to it's candle time.
+                trade_date = timeframe_to_prev_date(trade.open_date_utc)
+                trade_candle = dataframe.loc[dataframe['date'] == trade_date]
+                # Just opened trades do not have their candle complete yet therefore trade_candle may be None
+                if trade_candle is not None:
+                    trade_candle = trade_candle.squeeze()
+                    trade_stoploss = (current_rate - trade_candle['atr']) / current_rate
+                    new_stoploss = max(new_stoploss, trade_stoploss)
                 # turn into relative negative offset required by `custom_stoploss` return implementation
                 result = new_stoploss - 1
 
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -45,40 +45,6 @@ class DataProvider:
         """
         self._pairlists = pairlists
 
-    def refresh(self,
-                pairlist: ListPairsWithTimeframes,
-                helping_pairs: ListPairsWithTimeframes = None) -> None:
-        """
-        Refresh data, called with each cycle
-        """
-        if helping_pairs:
-            self._exchange.refresh_latest_ohlcv(pairlist + helping_pairs)
-        else:
-            self._exchange.refresh_latest_ohlcv(pairlist)
-
-    @property
-    def available_pairs(self) -> ListPairsWithTimeframes:
-        """
-        Return a list of tuples containing (pair, timeframe) for which data is currently cached.
-        Should be whitelist + open trades.
-        """
-        return list(self._exchange._klines.keys())
-
-    def ohlcv(self, pair: str, timeframe: str = None, copy: bool = True) -> DataFrame:
-        """
-        Get candle (OHLCV) data for the given pair as DataFrame
-        Please use the `available_pairs` method to verify which pairs are currently cached.
-        :param pair: pair to get the data for
-        :param timeframe: Timeframe to get data for
-        :param copy: copy dataframe before returning if True.
-                     Use False only for read-only operations (where the dataframe is not modified)
-        """
-        if self.runmode in (RunMode.DRY_RUN, RunMode.LIVE):
-            return self._exchange.klines((pair, timeframe or self._config['timeframe']),
-                                         copy=copy)
-        else:
-            return DataFrame()
-
     def historic_ohlcv(self, pair: str, timeframe: str = None) -> DataFrame:
         """
         Get stored historical candle (OHLCV) data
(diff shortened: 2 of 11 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -349,6 +350,11 @@ def backtest(self, processed: Dict,
         trades: List[LocalTrade] = []
         self.prepare_backtest(enable_protections)
 
+        # Update dataprovider cache
+        for pair, dataframe in processed.items():
+            self.dataprovider._set_cached_df(pair, self.timeframe, dataframe)
+        self.strategy.dp = self.dataprovider
```
rokups: I am not sure why/how `strategy.dp` gets unset, but it does in hyperopt. This workaround does the job.. But there may be a way to fix this more proper.

xmatthias: it's intentional as it contains an exchange object - which you can't submit to another process through pickle (pickle will complain and fail).

freqtrade/freqtradebot.py
```diff
@@ -553,7 +552,7 @@ def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = N
 
         if not strategy_safe_wrapper(self.strategy.confirm_trade_entry, default_retval=True)(
                 pair=pair, order_type=order_type, amount=amount, rate=buy_limit_requested,
-                time_in_force=time_in_force):
+                time_in_force=time_in_force, current_time=datetime.utcnow()):
```
xmatthias: you will want to use `datetime.now(timezone.utc)` here - otherwise comparisons will fail with "cannot compare timezone aware and timezone native ..."

freqtrade/freqtradebot.py
```diff
@@ -1191,8 +1190,8 @@ def execute_sell(self, trade: Trade, limit: float, sell_reason: SellCheckTuple)
 
         if not strategy_safe_wrapper(self.strategy.confirm_trade_exit, default_retval=True)(
                 pair=trade.pair, trade=trade, order_type=order_type, amount=amount, rate=limit,
-                time_in_force=time_in_force,
-                sell_reason=sell_reason.sell_reason):
+                time_in_force=time_in_force, sell_reason=sell_reason.sell_reason,
+                current_time=datetime.utcnow()):
```
xmatthias: datetime ... :O same as above ;)

freqtrade/optimize/backtesting.py
xmatthias: in theory i guess having a "no exchange dataprovider" would do ... 
but it'll require some more changes to dataprovider (it should fail if one of the "requires exchange" functions is called ...).

freqtrade/optimize/hyperopt.py
```diff
@@ -392,8 +392,6 @@ def start(self) -> None:
         self.backtesting.exchange._api_async = None  # type: ignore
         # self.backtesting.exchange = None  # type: ignore
         self.backtesting.pairlists = None  # type: ignore
-        self.backtesting.strategy.dp = None  # type: ignore
-        IStrategy.dp = None  # type: ignore
```
xmatthias, line 376: I'm ... not 100% certain this works if you use it with anything `-j > 1` ... 

Either way - i think with this it shouldn't be necessary to reassign dp to the strategy again and again in backtesting?

rokups: I pushed some more changes which are.. Basically a different flavor of same 👀 it seems like assigning objects to class is not exactly pickle-friendly. It does not fail immediately, but after a round of hyperopts exception is throw that dp is unset. My theory is that first round might be running in the original process and still has access to this object while it does not carry on into subprocesses. I have no proof though. However we already had this problem. [This](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/optimize/hyperopt.py#L76) used to be set to hyperopt class, not instance, and i got same issue while hyperopting - it would become unset.

freqtrade/optimize/backtesting.py
```diff
@@ -63,9 +63,7 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self.all_results: Dict[str, Dict] = {}
 
         self.exchange = ExchangeResolver.load_exchange(self.config['exchange']['name'], self.config)
-
-        dataprovider = DataProvider(self.config, self.exchange)
-        IStrategy.dp = dataprovider
+        self.dataprovider = DataProvider(self.config, self.exchange)
```
xmatthias: I guess "this" dataprovider should be called without exchange.

This wasn't possible before (mypy :eyes:)- but if you need ANYTHING from the dataprovider that requires an exchange class during backtesting, you're doing something wrong - as you're using "now" data to analyze historic data.

freqtrade/optimize/hyperopt.py
xmatthias: I guess if dataprovider is created without exchange, this should be fine.

Be sure to test it beyond 30 epochs though (or reduce the "random" epochs) - sometimes hyperopt reacts differently once it reaches that point - but i don't expect any problems with that
(note: i will test that myself too - but not today).

freqtrade/data/dataprovider.py
```diff
@@ -123,12 +89,82 @@ def get_analyzed_dataframe(self, pair: str, timeframe: str) -> Tuple[DataFrame,
...
+            self._exchange.refresh_latest_ohlcv(pairlist)
+
+    @property
+    def available_pairs(self) -> ListPairsWithTimeframes:
+        """
+        Return a list of tuples containing (pair, timeframe) for which data is currently cached.
+        Should be whitelist + open trades.
+        """
+        if self._exchange is None:
+            raise OperationalException('Exchange is not available to DataProvider.')
```
xmatthias: i'm ... not sure if this is correct behaviour.

I don't think too many will use this method - but i think this could return the list of pairs in `__cached_pairs` in this case (maybe even in every case ...)

xmatthias (commented):
`custom_stoploss` (as well as `confirm_*` methods are part of `strategy_methods_advanced.j2` too - so i think we should have that template aligned to the docs / reality too.

Also - as `get_analyzed_dataframe` is now available during backtesting, i think the note in the relevant section (around line 426) about backtesting in `strategy-customization.md` can also be removed.

rokups:
I was using this branch for a while and i _think_ i used it wrong, which may necessitate some additional helpers, but i need your advice @xmatthias to make sure i understand situation correctly. Consider this bit:

```py
    def confirm_trade_exit(self, pair: str, trade: 'Trade', order_type: str, amount: float,
                           rate: float, time_in_force: str, sell_reason: str,
                           current_time: 'datetime', **kwargs) -> bool:
        current_time = timeframe_to_prev_date(self.timeframe, current_time)
        dataframe, _ = self.dp.get_analyzed_dataframe(pair, self.timeframe)
        current_candle = dataframe.loc[dataframe['date'] == current_time]
        if current_candle is not None:
            current_candle = current_candle.squeeze()
            if current_candle['rsi'] > 50:
                return False

        return True
```

Like you said in discord this should return `None` for `current_candle` because `current_time` is time of in-progress candle which is not available. Therefore this bit has look-ahead bias in backtesting that nobody spotted. Given that it seems to me that `timeframe_to_prev_date()` is confusing. It acts more like `timeframe_round_date()`. To avoid loo-ahead bias in such situation i created these:

```py

def timeframe_to_prev_candle(timeframe: str, date: datetime = None) -> datetime:
    """
    Return date of previous candle.
    :param timeframe: timeframe in string format (e.g. "5m")
    :param date: date to use. Defaults to utcnow()
    :returns: date of previous candle (with utc timezone)
    :return:
    """
    # Round date to current candle. For live/dry runs this would round time to that of current
    # in-progress candle which is not available yet. For backtesting it is noop as date is already
    # rounded to that of in-progress candle (which we should not read).
    date = timeframe_to_prev_date(timeframe, date)
    return timeframe_to_prev_date(timeframe, date - timedelta(hours=0, minutes=1))


def timeframe_to_next_candle(timeframe: str, date: datetime = None) -> datetime:
    """
    Return date of previous candle.
    :param timeframe: timeframe in string format (e.g. "5m")
    :param date: date to use. Defaults to utcnow()
    :returns: date of previous candle (with utc timezone)
    :return:
    """
    # Round date to current candle. For live/dry runs this would round time to that of current
    # in-progress candle which is not available yet. For backtesting it is noop as date is already
    # rounded to that of in-progress candle (which we should not read).
    date = timeframe_to_prev_date(timeframe, date)
    return date + timedelta(hours=0, minutes=timeframe_to_minutes(timeframe))
```

But i am not sure if they will behave correctly in backtesting. What `current_time` do functions like `confirm_trade_exit` get in backtesting? Does it point to a current would-be-in-progress candle that we should not use?

xmatthias:
you ... shouldn't do this: `return timeframe_to_prev_date(timeframe, date - timedelta(hours=0, minutes=1))` ...
it's dangerous in cases where the timeframe is < 1m (it's rare and i've not seen it often - but i think FTX has 15s timeframes ...).
you should be able to use `timeframe_to_seconds()` and subtract that directly ... 


Apart from that - i'd rather have a helper of `get_current_candle(dataframe, current_time)` - maybe in the dataprovider ... as usually (or at least often), you'll not need all history - but only the "current candle" - so moving the slicing / selection of the candle to the user might be error-prone.
Considering that we're not sure ourselfs at this moment how it's 100% correct - it's better that way as it will allow us to fixed potentially detected problems "centrally" - without having all users read the docs again and adjust to cover all eventuallities.

---

The question you're asking is ... not a simple one - and i need to walk through an example (and hope you'll doublecheck my logic) to be sure i get it right...

Let's assume we use 1h candles for simplicity of calculations - and the trade opened based on a signal at 09:00 (so it opened at 10:00:01).
considering it's now **10:14** ... 


In live / dry - the "current candle" date will be 10:00 - the last available will be 09:00 - so we must use a logic similar to the one you post above to retreive the latest available candle.


**in backtesting:**
This is where it get's interresting / complicated...
we will not get 10:14 as date on 1h candles - but 09:00, 10:00, 11:00, ... - so only "valid" candle open dates.

for "regular" signals (buy / sell column) - we shift the signals by 1 candle - so signals calculated on the 09:00 candle will be acted on at 10:00 (open). This aligns with reality / forward testing - as the signal calculated on the 9:00 candle is only available at 10:00:01.


the method we should be looking at for the sell-callbacks (`custom_sell`, `custom_stop`) is `should_sell()` (the first method called - which then calls all sell-related callbacks) - which will receive the date as `sell_row[DATE_IDX].to_pydatetime()`.

`sell_row` will contain all data for the current candle (date, open, high, low, close of the 10:00 candle) - and since the trade opened at 10:00 - the first sell_row we're getting to check for sell-signals is with the date of 10:00.
For all "regular" methods (roi / stoploss / ...) - this 10:00 candle is the first to be evaluated (which aligns to reality - as roi + stoploss are checked against real-time price - not candle-values).

If you're looking at a candle (indicators assigned to a dataframe) - you will have to look at the 9:00 candle - as this is the only closed candle available at the "first time of check" - so i think the same logic (passed in date - 1 candle) will work for backtesting as well.

rokups:
> it's dangerous in cases where the timeframe is < 1m (it's rare and i've not seen it often - but i think FTX has 15s timeframes ...).
you should be able to use timeframe_to_seconds() and subtract that directly ...

I wasnt aware freqtrade supports those timeframes. At least in data download lowest possible timeframe is 1m. But yeah you are right despite the fact, subtracting one timeframe period is safer and cleaner anyhow. 👍🏻 

> Apart from that - i'd rather have a helper of get_current_candle(dataframe, current_time) - maybe in the dataprovider ... as usually (or at least often), you'll not need all history - but only the "current candle" - so moving the slicing / selection of the candle to the user might be error-prone.

I am not quite sure that is a good solution. I mean its fine if we need current candle. What if we need last two candles? Chances are people will start doing `timeframe_to_prev_date(timeframe, date - timedelta(hours=0, minutes=1))` themselves. I kept rewriting these helpers couple times not finding a satisfying variant. What is in the branch - consider it s a rough sketch. I am not sure how to do it better though. Maybe a better solution would be `self.dp.get_last_candle_time()`, and allow user to do whatever is needed with it, but it still does not eliminate need of the helper that could seek date back and forth by timeframe. To be honest i find it very unfortunate that `timeframe_to_prev|next_date` are named like that. Such naming is a lie because they actually round date and we can not easily step date back and forth. I am also not sure if we can rename them and replace them with differently-behaving ones. If anyone is using these functions it would break people's code in non-obvious ways...

> The question you're asking is ... not a simple one - and i need to walk through an example (and hope you'll doublecheck my logic) to be sure i get it right...

This confirms my understanding. So in both backtesting and live we can not use `timeframe_to_prev_date(current_time)` to look up a candle as in live it will return nothing and in backtesting it will look one candle into the future.

xmatthias:
> I wasnt aware freqtrade supports those timeframes.

Even if it doesn't today - it might tomorrow - so i'd keep the helper rather flexible ...

> To be honest i find it very unfortunate that timeframe_to_prev|next_date are named like that

my time-machine is currently broken ... so i can't change history now even though i agree :laughing: 

> What if we need last two candles

A `dp.get_candle_for_date(current_time)` helper might cover most of these usecases ... (renamed, same functionality as described above)
I assume in 90% of the cases will use either the current candle, or the candle from the "open" point.

Sure, if you need 2 or more candles, you'll need to do some "logic" anyway (that's why i was of the opinion that a well designed interface should pass in the sliced dataframe (so you can use `.iloc[-2]`) - which will however kill performance - and which is why i wasn't a big fan of adding the `dataframe` parameter in the first place ...).

---

after all, maybe the naming of these functions is not that relevant though - and a simple `to_prev_candle`  helper will work? (i don't think we need the `*next*` alteration - as that's kindof an invitation to use data from the future ...)

in combinaion with `dp.get_candle_for_date()` this should cover most usecases.
if you REALLY think you'll often need 2-3-5 candles - maybe a helper with `get_last_candles(time, 20)` which will slice the dataframe could be added ... 
sure, it'll kill performance just the same - but ONLY if you really need it, not for all other users.


---

An additional (but not relevant for this PR) change is to add date as "proper" index. This will speed up performance of these selects - and having helpers will make this change transparent for users ... (i'd however try to keep the 'date' column there too).

rokups:
Check this out. This post claims slicing can be done in nanoseconds https://stackoverflow.com/a/34035323/6537818 👀

Edit:
I re-tested:
```
%timeit -n 1000 df2 = df[:20000]
35.5 µs ± 104 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit -n 1000 df2 = df.loc[(df.index < 20000), :]
588 µs ± 6.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit -n 1000 df2 = df.loc[:20000, :]
40.7 µs ± 162 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

df.set_index('date', drop=False, inplace=True)
%timeit -n 1000 df2 = df[:'2021-04-14 21:19:00+00:00']
382 µs ± 7.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit -n 1000 df2 = df.loc[(df.index < '2021-04-14 21:19:00+00:00'), :]
576 µs ± 1.18 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit -n 1000 df2 = df.loc[:'2021-04-14 21:19:00+00:00', :]
400 µs ± 846 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
Does not seem slow to me.

TheJoeSchr:
> To be honest i find it very unfortunate that timeframe_to_prev|next_date are named like that. Such naming is a lie because they actually round date and we can not easily step date back and forth. I am also not sure if we can rename them and replace them with differently-behaving ones. If anyone is using these functions it would break people's code in non-obvious ways...

Rename could be stretched over 2 or more releases. First release rename `timeframe_to_prev()` and create a dummy `timeframe_to_prev()` which throws an obsolete exception.

Then next release implement the actual functionality for `timeframe_to_prev()`
(Maybe still logging a warning about rename)

That way nobody who keeps freqtrade reasonably up to date can miss it?

xmatthias:
in the strategy - the user should not have to deal with the time calculation - that's error-prone.

it seems pretty simple to have a `get_current_candle(current_time)` - and a `get_latest_candles(current_time, 20)` (to get a dataframe of last X candles).
This will remove the need to rename the methods - and will provide a simple way to access the data for the user.

It's rather simple - people do not really keep it uptodate - current release is 2021.4 - last week most issues were for 2021.2 or earlier (or develop) ... so the assumption that people keep their bot uptodate is flawed - as you'll not update unless something breaks or you're changing something anyway (like updating a strategy).

The bigger point on this is though - the "timeframe_to_" methods are not even relevant if the proper methods are provided to get the correct candles.

freqtrade/data/dataprovider.py
```diff
@@ -117,18 +97,99 @@ def get_analyzed_dataframe(self, pair: str, timeframe: str) -> Tuple[DataFrame,
...
+        if pair_key in self.__cached_pairs:
+            if self.runmode in (RunMode.DRY_RUN, RunMode.LIVE):
+                df, date = self.__cached_pairs[pair_key]
+            elif pair_key in self.__sliced_pairs:
+                df, date = self.__sliced_pairs[pair_key]
+            else:
+                max_index = self.__slice_index
+                df, date = self.__cached_pairs[pair_key]
+                df = df.iloc[max(0, max_index - MAX_DATAFRAME_CANDLES):max_index]
+                self.__cached_pairs[pair_key] = df, date
```
xmatthias: i think you'll want to assign this to `__sliced_pairs` - not `__cached_pairs` pairs ...

makes me wonder though - what's the point in caching the sliced version?
in dry/live - we wouldn't slice (no point - just return everything, which is what you do)

in backtesting - you'll usually only query the same slice once (after that you're one candle further, so need a new slice) - so if you wanna cache the slice, the key needs to include the index (`max_index`?) too - but at that point - slicing will be rather pointless.

freqtrade/data/dataprovider.py
```diff
@@ -99,9 +101,10 @@ def get_analyzed_dataframe(self, pair: str, timeframe: str) -> Tuple[DataFrame,
             if self.runmode in (RunMode.DRY_RUN, RunMode.LIVE):
                 df, date = self.__cached_pairs[pair_key]
             else:
-                max_index = self.__slice_index
                 df, date = self.__cached_pairs[pair_key]
-                df = df.iloc[max(0, max_index - MAX_DATAFRAME_CANDLES):max_index]
+                if self.__slice_index is not None:
```
rokups, line 105: Tbh i think that assert is probably better here. `self.__slice_index` should never be `None` in backtesting/hyperopt therefore it is a mistake in the bot. `if` hides this mistake and if this ever gets broken nobody will notice that and keep looking into last candle of entire range.

xmatthias: I think youre going off of a wrong(ish) assumption.

You're right if you only look at the callbacks (`custom_stop`, `custom_sell`).

However, nothing prevents me from using `dp.get_analyzed_dataframe` in a `populate_sell` method (to get a different pair, for example).
If that'll work is a different topic (cached_pairs will not be assigned at that point yet) - but that's potentially small change ... so this will prepare for that scenario.

rokups: Oh! `populate_sell` runs before main backtesting where slice index is set. I did not realize that 😅

xmatthias: Well thats why it's called only once (per pair)... Which makes it fast, but also kinda limited in what you can do with it

docs/strategy-advanced.md
```diff
@@ -293,20 +283,27 @@ class AwesomeStrategy(IStrategy):
...
+                new_stoploss = (current_rate - current_candle['atr']) / current_rate
+
+                # Round trade date to it's candle time.
+                trade_date = timeframe_to_prev_date(trade.open_date_utc)
+                trade_candle = dataframe.loc[dataframe['date'] == trade_date]
+                # Just opened trades do not have their candle complete yet therefore trade_candle may be None
+                if trade_candle is not None:
+                    trade_candle = trade_candle.squeeze()
+                    trade_stoploss = (current_rate - trade_candle['atr']) / current_rate
+                    new_stoploss = max(new_stoploss, trade_stoploss)
```
xmatthias, line 306: i guess this part is now outdated?

rokups: Uhm... Not exactly? I mean i updated such example in other spot in https://github.com/freqtrade/freqtrade/pull/4851/commits/17b9e898d2a234b9c1861192af21618ca63442a1 to showcase getting both last candle and trade candle. This bit here is not incorrect, but maybe incomplete as it does not show getting last candle.

xmatthias: ah you're right - it's to get the candle at the trade open - not the latest candle - i misread that ... All fine then 👍

xmatthias (approved):
LGTM :+1:

PR #4902, opened 2021-05-09
Documentation for running arm64 builds

docbobo:
Attempt to clarify documentation around arm64. Let me know, what you think

### DIFF
--- a/docs/docker_quickstart.md
+++ b/docs/docker_quickstart.md
@@ -67,21 +67,24 @@ Create a new directory and place the [docker-compose file](https://raw.githubuse
         # image: freqtradeorg/freqtrade:develop_pi
         ```
 
-=== "ARM 64 Systenms (Jetson Nano, Mac M1, Raspberry Pi 4 8GB)"
+=== "ARM 64 Systenms (Mac M1, Raspberry Pi 4, Jetson Nano)"
     In case of a Mac M1, make sure that your docker installation is running in native mode
+    Arm64 images are not yet provided via Docker Hub and need to be build locally first.
+    Depending on the device, this may take a few minutes (Apple M1) or multiple hours (Raspberry Pi)
 
     ``` bash
-    mkdir ft_userdata
-    cd ft_userdata/
-
-    # arm64 images are not yet provided via Docker Hub and need to be build locally first. Depending on the device,
-    # this may take a few minutes (Apple M1) or up to two hours (Raspberry Pi)
+    # Clone Freqtrade repository
     git clone  https://github.com/freqtrade/freqtrade.git
-    docker build -f ./freqtrade/docker/Dockerfile.aarch64 -t freqtradeorg/freqtrade:develop_arm64 freqtrade
+    cd freqtrade
+    # Optionally switch to the stable version
+    git checkout stable   
 
-    # Download the docker-compose file from the repository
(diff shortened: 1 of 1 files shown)


docs/docker_quickstart.md
```diff
@@ -67,29 +67,33 @@ Create a new directory and place the [docker-compose file](https://raw.githubuse
...
     cd ft_userdata/
-    # Download the docker-compose file from the repository
-    curl https://raw.githubusercontent.com/freqtrade/freqtrade/stable/docker-compose.yml -o docker-compose.yml
 
-    # Edit the compose file, uncomment the "build" step and use "./docker/Dockerfile.aarch64"
-    # Also, change the image name to something of your liking
+    # arm64 images are not yer provided via Docker Hub and need to be build locally first. Depending on the device,
+    # this may take a few minutes (Apple M1) or up to two hours (Raspberry Pi)
+    git clone  https://github.com/freqtrade/freqtrade.git
+    docker build -f ./freqtrade/docker/Dockerfile.aarch64 -t freqtradeorg/freqtrade:develop_arm64 freqtrade
```
xmatthias: Don't recommend doing it this way - but instead use docker-compose to build the image.

The compose-file will need to be edited anyway (the image name is now different) - so using `docker-compose build` will be the simpler command (and will also be simpler in update-scenarios).

docs/docker_quickstart.md
```diff
@@ -67,29 +67,33 @@ Create a new directory and place the [docker-compose file](https://raw.githubuse
...
 
     ``` bash
     mkdir ft_userdata
     cd ft_userdata/
-    # Download the docker-compose file from the repository
-    curl https://raw.githubusercontent.com/freqtrade/freqtrade/stable/docker-compose.yml -o docker-compose.yml
 
-    # Edit the compose file, uncomment the "build" step and use "./docker/Dockerfile.aarch64"
-    # Also, change the image name to something of your liking
+    # arm64 images are not yer provided via Docker Hub and need to be build locally first. Depending on the device,
```
xmatthias:
```suggestion
    # arm64 images are not yet provided via Docker Hub and need to be build locally first. Depending on the device,
```

docs/docker_quickstart.md
```diff
@@ -67,29 +67,33 @@ Create a new directory and place the [docker-compose file](https://raw.githubuse
...
+    # this may take a few minutes (Apple M1) or up to two hours (Raspberry Pi)
```
xmatthias: i'd keep it at "multiple hours" - as this will not only depend on the raspberry, but also on the network speed.

Rada8732 (approved):

docs/docker_quickstart.md
docbobo: Fair enough. Personally, I think that gets a little bit more tedious when you are trying to build anything on top of the other Dockerfiles inside `./docker`. But if you prefer it that way, I'll happily change that.

xmatthias: usually, mixing docker + docker-compose is quite confusing to people not used to docker (at least that's my experience from trying to explain it to some colleagues at work) ...

xmatthias (approved):
:+1: 
Great catch to notice that the docs actually don't work if you're not checked out!

PR #4915, opened 2021-05-10
Small update custom_sell() doc

froggleston:
## Summary

- Update custom_sell() example to comment that the current trade row is at trade open as written.
- Change "abstain" to something clearer for non-fluent English speakers.


xmatthias (approved):
LGTM :+1: 

i hope this will make this section clearer!

PR #4926, opened 2021-05-13
Two fixes

rokups:
Fixed rare exception when one of many backtested pairs ends up with no data when adjusted for startup candle count, more info in commit message.

Also fixed protections being loaded multiple times for first strategy when backtesting. That was a result of my previous fix of protections not being loaded from strategy class when backtesting.

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -115,8 +115,6 @@ class Backtesting:
 
         # Get maximum required startup period
         self.required_startup = max([strat.startup_candle_count for strat in self.strategylist])
-        # Load one (first) strategy
-        self._set_strategy(self.strategylist[0])
 
     def __del__(self):
         LoggingMixin.show_output = True
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -9,7 +9,7 @@ from copy import deepcopy
 from datetime import datetime, timedelta, timezone
 from typing import Any, Dict, List, Optional, Tuple
 
-from pandas import DataFrame, NaT
+from pandas import DataFrame
 
 from freqtrade.configuration import TimeRange, remove_credentials, validate_config_consistency
 from freqtrade.constants import DATETIME_PRINT_FORMAT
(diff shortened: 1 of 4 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -461,9 +459,19 @@ def backtest_one_strategy(self, strat: IStrategy, data: Dict[str, Any], timerang
             preprocessed[pair] = trim_dataframe(df, timerange,
                                                 startup_candles=self.required_startup)
         min_date, max_date = history.get_timerange(preprocessed)
+
+        # Overall data availability check. Nothing to do if we have no data at all.
         if min_date is NaT or max_date is NaT:
             raise OperationalException(
                 "No data left after adjusting for startup candles. ")
```
xmatthias: I would almost think that this should happen in a different sequence.

move your "data removal" check to the top of this block (right before `get_timerange()`) - this is actually a better fix than the one i added before

check if ANY pair is left (if the dict is empty - no point in continuing) further.

then get the timerange - by which point, my previous "no data left after adjusting" fix might no longer be necessary - as the min/max time will only be NaT if one of the pairs has no data.

freqtrade/optimize/backtesting.py
```diff
@@ -457,13 +455,21 @@ def backtest_one_strategy(self, strat: IStrategy, data: Dict[str, Any], timerang
         preprocessed = self.strategy.ohlcvdata_to_dataframe(data)
 
         # Trim startup period from analyzed dataframe
-        for pair, df in preprocessed.items():
-            preprocessed[pair] = trim_dataframe(df, timerange,
-                                                startup_candles=self.required_startup)
-        min_date, max_date = history.get_timerange(preprocessed)
-        if min_date is NaT or max_date is NaT:
+        for pair in list(preprocessed.keys())[:]:
```
s0undt3ch:
```suggestion
        for pair in list(preprocessed.keys()):
```
`list()` creates a copy of `.keys()` and then `[:]` creates a copy of the `list()`, resulting in a double copy 
Is that really necessary?

freqtrade/optimize/backtesting.py
```diff
@@ -457,13 +455,21 @@ def backtest_one_strategy(self, strat: IStrategy, data: Dict[str, Any], timerang
...
+            df = preprocessed[pair]
+            df = trim_dataframe(df, timerange, startup_candles=self.required_startup)
+            if len(df) > 0:
+                preprocessed[pair] = df
+            else:
+                logger.warning(f'{pair} has no data left after adjusting for startup candles, '
+                               f'skipping.')
+                del preprocessed[pair]
+
+        if len(preprocessed) == 0:
```
s0undt3ch:
```suggestion
        if preprocessed:
```
Since `preprocessed` looks like a dictionary, this is all it takes to check if it's empty.

freqtrade/optimize/backtesting.py
```diff
@@ -457,13 +455,21 @@ def backtest_one_strategy(self, strat: IStrategy, data: Dict[str, Any], timerang
         preprocessed = self.strategy.ohlcvdata_to_dataframe(data)
 
         # Trim startup period from analyzed dataframe
-        for pair, df in preprocessed.items():
-            preprocessed[pair] = trim_dataframe(df, timerange,
-                                                startup_candles=self.required_startup)
-        min_date, max_date = history.get_timerange(preprocessed)
-        if min_date is NaT or max_date is NaT:
+        for pair in list(preprocessed.keys()):
```
s0undt3ch: Actually, this is even, better, and enough. It doesn't create a keys view object. Sorry, should have mentioned on the initial review.
```suggestion
        for pair in list(preprocessed):
```

xmatthias (approved):
LGTM :+1:

PR #4933, opened 2021-05-14
Couple tweaks for docs.

rokups:
Fixed `timeframe_to_prev_date()` call and removed `timeframe_to_prev_date` import in other spot because example does not use it.


xmatthias (approved):
Great, thanks! :+1:

PR #4941, opened 2021-05-15
prevent backtest stoploss trade price being set above candle high

brookmiles:
## Summary

if stoploss price is above the candle high, set it to candle open instead.  this can occur if stoploss had previously been reached but the sell was prevented by `confirm_trade_exit`

20210515-085158-msedge

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -218,6 +218,12 @@ class Backtesting:
         """
         # Special handling if high or low hit STOP_LOSS or ROI
         if sell.sell_type in (SellType.STOP_LOSS, SellType.TRAILING_STOP_LOSS):
+            if trade.stop_loss > sell_row[HIGH_IDX]:
+                # our stoploss was already higher than candle high,
+                # possibly due to a cancelled trade exit.
+                # sell at open price.
+                return sell_row[OPEN_IDX]
+
             # Set close_rate to stoploss
             return trade.stop_loss
         elif sell.sell_type == (SellType.ROI):
--- a/tests/optimize/test_backtest_detail.py
+++ b/tests/optimize/test_backtest_detail.py
@@ -185,7 +185,7 @@ tc11 = BTContainer(data=[
     [0, 5000, 5050, 4950, 5000, 6172, 1, 0],
     [1, 5000, 5050, 4950, 5100, 6172, 0, 0],
     [2, 5100, 5251, 5100, 5100, 6172, 0, 0],
-    [3, 4850, 5050, 4650, 4750, 6172, 0, 0],
+    [3, 5000, 5150, 4650, 4750, 6172, 0, 0],
     [4, 4750, 4950, 4350, 4750, 6172, 0, 0]],
     stop_loss=-0.10, roi={"0": 0.10}, profit_perc=0.019, trailing_stop=True,
     trailing_only_offset_is_reached=True, trailing_stop_positive_offset=0.05,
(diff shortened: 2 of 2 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -218,6 +218,12 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
         # Special handling if high or low hit STOP_LOSS or ROI
         if sell.sell_type in (SellType.STOP_LOSS, SellType.TRAILING_STOP_LOSS):
+            if trade.stop_loss > sell_row[HIGH_IDX]:
+                # our stoploss was already higher than candle high,
+                # possibly due to a cancelled trade exit.
+                # sell at open price.
+                return sell_row[OPEN_IDX]
+
             # Set close_rate to stoploss
             return trade.stop_loss
```
xmatthias, line 228: wouldn't a simple `min(trade.stop_loss, sell_row[HIGH_IDX])` be the better / simpler option (now if HIGH or OPEN is up for debate)?

---

Why would we use "open" if the price is above high - and not high?
You could also say we're using low (The most conservative) ... but what will be the most reliable value? (i really don't know!)

brookmiles: > wouldn't a simple min(trade.stop_loss, sell_row[HIGH_IDX]) be the better / simpler option (now if HIGH or OPEN is up for debate)?

presumably we want to keep the original stoploss if it actually does occur during the candle that we're processing, even if it's higher than the open price?

> Why would we use "open" if the price is above high - and not high?
You could also say we're using low (The most conservative) ... but what will be the most reliable value? (i really don't know!)

My reasoning behind choosing "open" in this case is based on the assumption that since the stoploss was already higher than the candle high before this candle started, it should trigger at the "earliest" possible price.  I'm not entirely clear on the order of operations here, so this might not necessarily be accomplishing what I hoped.

rokups: > My reasoning behind choosing "open" in this case is based on the assumption that since the stoploss was already higher than the candle high before this candle started, it should trigger at the "earliest" possible price.

This makes sense to me.

> I'm not entirely clear on the order of operations here, so this might not necessarily be accomplishing what I hoped.

freqtrade assumes open -> low -> high -> close.

xmatthias: > freqtrade assumes open -> low -> high -> close.

that's not the full picture.
in general, freqtrade assumes this - however, stoploss "high" (trailing stop reference value) is assumed to happen first - there's no signal based on that - but the stoploss is adjusted first, then checked - which is the problem we're seeing with very tight stoplosses (which adjust to high - and then immediately sell as low is > high - stoploss distance).

The problem is - having it this way fixes the reverse problem - where price SHOULD be adjusted first (out of a static stoploss) ... so with this, it can also remain in the trade for longer (which can also be just as realistic, and is just as likely).

xmatthias:
Interrestingly, the failing test has exactly this case:

2021-05-17-195611_1836x680_scrot

Now i'm not certain if the test should be fixed by changing the candle (raising "H" to 5100 works) - or by changing the expected outcome (to open).

@rokups, @brookmiles  Opinions please... ?

rokups:
Such a big gap is a valid circumstance. Wouldnt changing test be essentially sweeping issue under rug? 02:00 candle looks like it could have no volume at all, so even without exit blocking bot could fail selling at it's open/close price, then rising high of the next candle would be invalid as in theory that point would never be a valid sell.

xmatthias:
Well to be fair - this test is "manually created" data (all tests in backtest_detail are).

probably the correct approach:
* fix this test by raising high (this was testing something else)
* add a new testcase with this particular scenario - with the expected result.

xmatthias (approved):
LGTM :+1: 

Let's break some strategies dreams of profits ...

PR #4944, opened 2021-05-15
Add support for timerange end date to `download-data`

brookmiles:
## Summary
the `download-data` command will honour the timerange end date, which will allow downloading only a specific historical time range for backtesting without requiring downloading the entire history up to the present.

Solve the issue: #4412

### DIFF
--- a/freqtrade/data/history/history_utils.py
+++ b/freqtrade/data/history/history_utils.py
@@ -267,11 +267,11 @@ def _download_trades_history(exchange: Exchange,
 
         until = None
         if (timerange and timerange.starttype == 'date'):
-          since = timerange.startts * 1000
-          if timerange.stoptype == 'date':
-            until = timerange.stopts * 1000
+            since = timerange.startts * 1000
+            if timerange.stoptype == 'date':
+                until = timerange.stopts * 1000
         else:
-          since = int(arrow.utcnow().shift(days=-new_pairs_days).float_timestamp) * 1000
+            since = int(arrow.utcnow().shift(days=-new_pairs_days).float_timestamp) * 1000
 
         trades = data_handler.trades_load(pair)
 
--- a/freqtrade/data/history/history_utils.py
+++ b/freqtrade/data/history/history_utils.py
@@ -265,9 +265,13 @@ def _download_trades_history(exchange: Exchange,
     """
     try:
 
-        since = timerange.startts * 1000 if \
-            (timerange and timerange.starttype == 'date') else int(arrow.utcnow().shift(
-                days=-new_pairs_days).float_timestamp) * 1000
+        until = None
+        if (timerange and timerange.starttype == 'date'):
+          since = timerange.startts * 1000
+          if timerange.stoptype == 'date':
+            until = timerange.stopts * 1000
+        else:
+          since = int(arrow.utcnow().shift(days=-new_pairs_days).float_timestamp) * 1000
 
         trades = data_handler.trades_load(pair)
 
@@ -295,6 +299,7 @@ def _download_trades_history(exchange: Exchange,
         # Default since_ms to 30 days if nothing is given
         new_trades = exchange.get_historic_trades(pair=pair,
                                                   since=since,
+                                                  until=until,
                                                   from_id=from_id,
                                                   )
         trades.extend(new_trades[1])


freqtrade/data/history/history_utils.py
```diff
@@ -295,6 +299,7 @@ def _download_trades_history(exchange: Exchange,
         # Default since_ms to 30 days if nothing is given
         new_trades = exchange.get_historic_trades(pair=pair,
                                                   since=since,
+                                                  until=until,
```
xmatthias, line 302: i would think we should also have a check to verify if the end is before the "current data start date" (it's in the non-modified section between the 2 modified lines). Otherwise we risk having people with "gaps" in the data - which will be difficult to explain (and even more difficult to fix)

I'm also not certain this fully fixes #4412 - this only implements it for data-downloading of trades data ("kraken mode") ... but won't stop ohlcv downloads (binance, ftx, ...).

brookmiles: > i would think we should also have a check to verify if the end is before the "current data start date" (it's in the non-modified section between the 2 modified lines). Otherwise we risk having people with "gaps" in the data - which will be difficult to explain (and even more difficult to fix)

I think the existing behaviour to clear the existing data if the start date is before the current data start date will be sufficient to prevent this?

> I'm also not certain this fully fixes #4412 - this only implements it for data-downloading of trades data ("kraken mode") ... but won't stop ohlcv downloads (binance, ftx, ...).

This is true.  Fixing it for "kraken mode" was really my main goal, since can take many hours (or days), compared to the seconds or minutes required for the OHLC version.  Would be nice to have both cases covered of course, but this one seemed like an easy win.

xmatthias: > I think the existing behaviour to clear the existing data if the start date is before the current data start date will be sufficient to prevent this?

you're right ... i was thinking about the following behaviour

```
download-data --timerange 20200101-20200301

download-data --timerange 20200401-20200801
download-data --timerange 202001001-20201201
```

which at first glance, will skip March and August+September.
After thinking about it though - the 2nd call will adjust the startdate to start on March 1st ... and the 3rd call will also be adjusted.

xmatthias (approved):
LGTM :+1:

NeoDimi:
So what's the correct syntax then?
Because:
`docker-compose run --rm freqtrade download-data --exchange binance -t 4h -p KDA/USDT --timerange 20220401-20220501`
for instance will NOT download one month of data from april 1st to may 1st but the data from april 1st to today.

brookmiles:
> So what's the correct syntax then? Because: `docker-compose run --rm freqtrade download-data --exchange binance -t 4h -p KDA/USDT --timerange 20220401-20220501` for instance will NOT download one month of data from april 1st to may 1st but the data from april 1st to today.

The end date only applies when using `--dl-trades` for downloading Kraken data as described here: https://www.freqtrade.io/en/stable/data-download/#sub-command-trades-to-ohlcv

Downloading data this way from Kraken is much much slower than normal (think hours, or days), so being able to download incrementally (up to a certain date) is more valuable.

NeoDimi:
Ok, so there's no way to download specific time ranges for targeted backtesting? (e.g. only during the previous bull run or bear market)

brookmiles:
> Ok, so there's no way to download specific time ranges for targeted backtesting? (e.g. only during the previous bull run or bear market)

You can limit each backtest to whatever timerange you'd like, you don't have to backtest against the entire history that you downloaded.

NeoDimi:
Ah, yes of course.
Much better to have all data and backtest a specific window. Didn't see the --timerange for the backtesting command.
Thx a lot, sorry for stupid question!

xmatthias:
> > So what's the correct syntax then? Because: `docker-compose run --rm freqtrade download-data --exchange binance -t 4h -p KDA/USDT --timerange 20220401-20220501` for instance will NOT download one month of data from april 1st to may 1st but the data from april 1st to today.
> 
> The end date only applies when using `--dl-trades` for downloading Kraken data as described here: https://www.freqtrade.io/en/stable/data-download/#sub-command-trades-to-ohlcv
> 
> Downloading data this way from Kraken is much much slower than normal (think hours, or days), so being able to download incrementally (up to a certain date) is more valuable.

That's actually not quite true - #6753 added support for the "end" parameter to both commands.
It'll however only respect the "end" command if no (future) data is available - so if you downloaded "june" (`--timerange 20220601-`) first, then any end-date specified prior to the available data is ignored to avoid gaps in the data.

PR #4962, opened 2021-05-17
Total row for telegram "/status table" command

eschava:
## Summary
Adds "Total" row to the trades table for telegram "/status table" command and includes the total amount of all trades in the fiat currency (if any)

## What's new?
Adds total row for telegram "/status table" command

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -5,10 +5,10 @@ This module manage Telegram communication
 """
 import json
 import logging
-import re
 from datetime import timedelta
 from html import escape
 from itertools import chain
+from math import isnan
 from typing import Any, Callable, Dict, List, Union
 
 import arrow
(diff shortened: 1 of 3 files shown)


eschava:
Looks like builds are failing after mypy bug

https://github.com/python/mypy/issues/6040

freqtrade/rpc/telegram.py
```diff
@@ -354,19 +355,39 @@ def _status_table(self, update: Update, context: CallbackContext) -> None:
...
+                self._config['stake_currency'], fiat_currency)
+
+            show_total = fiat_currency != ''
+            total_sum = 0.0
+            if show_total:
+                r = re.compile(".*\\((-?\\d*\\.\\d*)\\)")
+                for trade in statlist:
+                    m = r.match(trade[-1])
+                    if m is not None:
+                        total_sum += float(m[1])
```
xmatthias: as a VERY quick feedback:
no matter what you're trying to do - this is not the right approach as it'll be very error-prone - and you're trying to parse this out from a string that was created less than a second (or 5 lines) ago .

Maybe consider moving the calculation of the sum to `rpc.py` - and returning it as additional field (`_rpc_status_table` is only used by telegram afaik - so the impact should be pretty low).

Overall - it would be good to add a screenshot of the new output to the PR header - so it's clear what you're trying to move towards.

eschava: image

eschava: This is how it's going to look if number of active trades is more than one

xmatthias: Overall - i like.

is there a particular reason you chose not to show the relative profits - but only the Fiat one?
(note: this is more a question about your thoughts / considerations, not a request to add it).

eschava: Agree, it wasn't possible when I parsed the output of _rpc_status_table command to get total profit but now, when logic has moved to the RPC method, it could be implemented

xmatthias (approved):
Let's leave the "relative" total profit for the future - the calculation will not be as simple, considering stakes per pair can be different.

LGTM :+1:

PR #4963, opened 2021-05-17
update doc to reflect better empty dataframe check

robcaulk:

## Summary
Update the advanced strategy doc

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?
Changing trade_candle is None logic to trade_candle.empty

### DIFF
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -65,9 +65,10 @@ class AwesomeStrategy(IStrategy):
         trade_date = timeframe_to_prev_date(self.timeframe, trade.open_date_utc)
         # Look up trade candle.
         trade_candle = dataframe.loc[dataframe['date'] == trade_date]
-        # trade_candle may be None for trades that just opened as it is still incomplete.
-        if trade_candle is not None:
+        # trade_candle may be empty for trades that just opened as it is still incomplete.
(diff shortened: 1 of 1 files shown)


docs/strategy-advanced.md
```diff
@@ -65,9 +65,10 @@ class AwesomeStrategy(IStrategy):
         trade_date = timeframe_to_prev_date(self.timeframe, trade.open_date_utc)
         # Look up trade candle.
         trade_candle = dataframe.loc[dataframe['date'] == trade_date]
-        # trade_candle may be None for trades that just opened as it is still incomplete.
-        if trade_candle is not None:
+        # trade_candle may be empty for trades that just opened as it is still incomplete.
+        if trade_candle.empty:
             # <...>
+        trade_candle = trade_candle.squeeze()
```
xmatthias: your code sample will not work (not as is).

you'll need some "fallback" scenario in case a candle is empty  (basically returning or whatever)- otherwise you'll run the `.squeeze()` on an empty dataframe.

The correct approach would be to check `if not trade.candle.empty:` - and then move the squeeze (and everything that follows) into that block.

xmatthias (approved):

PR #4964, opened 2021-05-17
Add backoff timer for coingecko API

thraizz:
## Summary
This fixes conversion from stake to fiat when running multiple bots by setting a future timestamp for retrying to get data.

## Quick changelog
- Set a timestamp in fiat_converter for retrying to get coingecko data

## What's new?
By accepting the possibility for 429 errors when trying to get data from Coingecko API we can enhance the ability to run multiple freqtrade instances on one machine. As all instances will fetch data from Coingecko in a short timeframe we will receive an `429 Too Many Requests` error, which results in no fiat conversion for the rest of the instances runtime. By retrying to fetch the data later, we can fetch the data for any number of instance that might run after some iterations.

## TODO
- [ ] Discuss the settting of a timestamp. Right place, right name?
- [x] Unit test?
- [x] Different warning when `self._cryptomap == {}`, to indicate that this will be fine in a couple of minutes?

### DIFF
--- a/freqtrade/rpc/fiat_convert.py
+++ b/freqtrade/rpc/fiat_convert.py
@@ -58,6 +58,13 @@ class CryptoToFiatConverter:
                     "Too many requests for Coingecko API, backing off and trying again later.")
                 # Set backoff timestamp to 60 seconds in the future
                 self._backoff = datetime.datetime.now().timestamp() + 60
+                return
+            # If the request is not a 429 error we want to raise the normal error
+            logger.error(
+                "Could not load FIAT Cryptocurrency map for the following problem: {}".format(
+                  request_exception
+                )
+            )
         except (Exception) as exception:
             logger.error(
                 f"Could not load FIAT Cryptocurrency map for the following problem: {exception}")
--- a/freqtrade/rpc/fiat_convert.py
+++ b/freqtrade/rpc/fiat_convert.py
@@ -143,10 +143,13 @@ class CryptoToFiatConverter:
         if crypto_symbol == fiat_symbol:
             return 1.0
 
-        if self._cryptomap == {} and self._backoff <= datetime.datetime.now().timestamp():
-            self._load_cryptomap()
-            # return 0.0 if we still dont have data to check, no reason to proceed
-            if self._cryptomap == {}:
+        if self._cryptomap == {}:
+            if self._backoff <= datetime.datetime.now().timestamp():
+                self._load_cryptomap()
+                # return 0.0 if we still dont have data to check, no reason to proceed
+                if self._cryptomap == {}:
+                    return 0.0
+            else:
                 return 0.0
 
         if crypto_symbol not in self._cryptomap:
--- a/freqtrade/rpc/fiat_convert.py
+++ b/freqtrade/rpc/fiat_convert.py
@@ -54,7 +54,8 @@ class CryptoToFiatConverter:
             self._cryptomap = {x['symbol']: x['id'] for x in coinlistings}
         except RequestException as request_exception:
             if "429" in str(request_exception):
-                logger.warning("Too many requests for Coingecko API, backing off and trying again later.")
+                logger.warning(
+                    "Too many requests for Coingecko API, backing off and trying again later.")
                 # Set backoff timestamp to 60 seconds in the future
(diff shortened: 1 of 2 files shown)


xmatthias (commented):
i'm not ... sure how you encounter this ... 
i guess you must be starting up 10+ of instances on the same machine at the same time / one after the other to get to this error.

based on their [therms](https://www.coingecko.com/en/api_terms#:~:text=4.2%20Rate%20limit%20for%20the,to%20you%20or%20any%20Users.) - they would allow 10 calls per second ... i'm not sure how likely it is to actually run into this ...

freqtrade/rpc/fiat_convert.py
```diff
@@ -127,6 +136,9 @@ def _find_price(self, crypto_symbol: str, fiat_symbol: str) -> float:
         if crypto_symbol == fiat_symbol:
             return 1.0
 
+        if self._cryptomap == {} and self._backoff <= datetime.datetime.now().timestamp():
+            self._load_cryptomap()
```
xmatthias: I'm not sure about the impact of this ... 

it's called from (semi)random spots in the code - so doing 2 calls "inline" might slow down certain points - and if that again runs into the ratelimit - it might not fix anything after all ...

xmatthias: just to be clear - i'm not against this in general, and i guess it might help some people.

A warning when this happens would for sure be helpful - so it's obvious that this happened.
You might need to handle the case of "reloading fails too" differently though - as now, 2 reloads would result in "unsupported crypto-symbol" - which is kind of wrong.

xmatthias:
After testing the api limits of Coingecko (i can get ~12 Requests per minute - with a 1s backoff - which is a FAR shot from the 100/minute they claim).

I think/hope this will help people experiencing issues with coingecko - so i'd be more than happy to have this merged if you're ready to finish this PR (i do agree with all TODO's you added - but i think your 3rd one is complete - there IS a warning about the backoff now).

freqtrade/rpc/fiat_convert.py
thraizz: This is now handled below by returning 0.0 if _cryptomap is still `{}`

tests/rpc/test_fiat_convert.py
```diff
@@ -21,6 +22,12 @@ def test_fiat_convert_is_supported(mocker):
 def test_fiat_convert_find_price(mocker):
     fiat_convert = CryptoToFiatConverter()
 
+    fiat_convert._cryptomap = {}
+    fiat_convert._backoff = 0
+    mocker.patch('freqtrade.rpc.fiat_convert.CryptoToFiatConverter._load_cryptomap',
+                 return_value=None)
+    assert fiat_convert.get_price(crypto_symbol='BTC', fiat_symbol='EUR') == 0.0
```
thraizz, line 29: I am not really sure whether this belongs here or not. We test find_price, yes, but we also patch something out that other tests below might need to produce meaningful tests. But we also patch out the tested method there and test the hardcoded return value, so it might be okay 😸

freqtrade/rpc/fiat_convert.py
```diff
@@ -25,6 +27,7 @@ class CryptoToFiatConverter:
     _coingekko: CoinGeckoAPI = None
 
     _cryptomap: Dict = {}
+    _backoff = int
```
thraizz: The only open point is whether or not the `_backoff` attribute here is okay.
Maybe this should be 
```suggestion
    _backoff: int = 0
```
instead?

xmatthias: the way it is now is definitely wrong (it's assigning the type of int to the variable ...) i wonder why mypy ain't complaining very loudly about this.

When doing that however, i think you must also check for `if self._backoff and ...` below - otherwise `self._backoff` will always be < current timestamp.

tests/rpc/test_fiat_convert.py
xmatthias: Mocks are "per function / per test" - so the mock will return to norma for the next test.

I think it's perfectly fine - and adds a realistic case (where it's returning empty).

xmatthias (approved):
LGTM :+1: 

Great contribution! I hope this will help people with this problem a lot!

PR #4969, opened 2021-05-18
[WIP] Show in that we are using a custom_stoploss

Pascal66:
## Summary
We are using a custom_stoploss

## What's new?
*Improve sell information.*

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -47,6 +47,7 @@ class SellType(Enum):
     FORCE_SELL = "force_sell"
     EMERGENCY_SELL = "emergency_sell"
     CUSTOM_SELL = "custom_sell"
+    CUSTOM_STOP_LOSS = "custom_stop_loss"
     NONE = ""
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


freqtrade/strategy/interface.py
```diff
@@ -678,6 +679,9 @@ def stop_loss_reached(self, current_rate: float, trade: Trade,
             # If initial stoploss is not the same as current one then it is trailing.
             if trade.initial_stop_loss != trade.stop_loss:
                 sell_type = SellType.TRAILING_STOP_LOSS
+                if self.use_custom_stoploss :
+                    sell_type = SellType.CUSTOM_STOP_LOSS
```
xmatthias:
```suggestion
                if self.use_custom_stoploss:
                    sell_type = SellType.CUSTOM_STOP_LOSS
```

xmatthias (changes requested):
Unfortunately, this won't do it (not only (but also) because it's breaking CI).

There's some logic around the sell type (specifically around trailing stops) - so this logic will have to be adapted / changed in all places that rely on the "sell reason = trailing stoploss"  to include the new reason as well... (basically searching for `TRAILING_STOP_LOSS`  should get you quite far).

---

You're also making the assumption that whenever custom_stoploss is enabled, trailing-stop is automatically disabled.

While we recommend this in the docs, it's currently not enforced, so this behaviour can potentially lead to "real" trailing stops to be reported as custom-stoploss - even though that never triggered / returned a real value).

Assuming we'd enforce trailing-stop to be disabled when custom-stop is enabled, i'm not certain on the added value for this - as you could rely that trailing_stop is then triggered by custom-stop (that's just a thought though - i'm not sure if this should be enforced or not).

Pascal66:
Thank you to have looked on it.
Wish one day to have a more precise sell(buy) reason to permit backtesting  and hyperopt improvment

PR #4975, opened 2021-05-18
Add ability to choose unit in unfilled timeout

Axel-CH:
## Summary
The goal of this PR is to add the possibility to select the unit or unfilled timeout now the default value was minutes. 

## What's new?
Now the user can set seconds if he need to. The default value stay "minutes"


xmatthias (approved):
Great! :+1: 

I've added this to the full config (and to the config template) so it's readily available for people - but i really like the approach you've taken for this!

Issue #3644, opened 2020-07-29
Pandas warning while backtesting


jblestang:
## Describe your environment

  * Operating system: Ubuntu 18.04
  * Python Version: 3.6
  * CCXT version: 1.30.48
  * Freqtrade Version: develop-d9d999ea

## Describe the problem:

*from time to time I get a warning while backtesting

### Steps to reproduce:

  1. Run backtesting
  
### Observed Results:

* An error while backtesting implying we try to
Modify the dataframe when we shall be only reading from the dataframe. 
* My dataframe contains data from a 4h ticker and merged data from daily and weekly timeframe (using the dataprovider)

### Relevant code exceptions or logs
```
2020-07-29 10:42:20,200 - freqtrade - ERROR - Fatal exception!
Traceback (most recent call last):
  File "/root/freqtrade/freqtrade/main.py", line 36, in main
    return_code = args['func'](args)
  File "/root/freqtrade/freqtrade/commands/optimize_commands.py", line 49, in start_backtesting
    backtesting.start()
  File "/root/freqtrade/freqtrade/optimize/backtesting.py", line 416, in start
    position_stacking=position_stacking,
  File "/root/freqtrade/freqtrade/optimize/backtesting.py", line 311, in backtest
    data: Dict = self._get_ohlcv_as_lists(processed)
  File "/root/freqtrade/freqtrade/optimize/backtesting.py", line 161, in _get_ohlcv_as_lists
    pair_data.loc[:, 'buy'] = 0  # cleanup from previous run
  File "/root/freqtrade/.env/lib/python3.6/site-packages/pandas/core/indexing.py", line 671, in _setitem_
    self._setitem_with_indexer(indexer, value)
  File "/root/freqtrade/.env/lib/python3.6/site-packages/pandas/core/indexing.py", line 845, in _setitem_with_indexer
    self.obj[key] = _infer_fill_value(value)
  File "/root/freqtrade/.env/lib/python3.6/site-packages/pandas/core/frame.py", line 2938, in _setitem_
    self._set_item(key, value)
  File "/root/freqtrade/.env/lib/python3.6/site-packages/pandas/core/frame.py", line 3007, in _set_item
    self._check_setitem_copy()
  File "/root/freqtrade/.env/lib/python3.6/site-packages/pandas/core/generic.py", line 3729, in _check_setitem_copy
    raise com.SettingWithCopyError(t)
pandas.core.common.SettingWithCopyError:
[...]

PR #4979, opened 2021-05-19
Fix hyperopt trim to empty dataframe

kamontat:

## Summary

I try to run freqtrade on my local machine (MacOS) I always get `cannot set a frame with no defined index and a scalar` error on latest develop.

After debugging, I found root cause when hyperopt do trim dataframe cause dataframe to be empty and cause the problem.

## Quick changelog

- none

## What's new?

### Reference

1. https://stackoverflow.com/questions/38886080/python-pandas-series-why-use-loc
2. #3644

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -192,8 +192,8 @@ class Backtesting:
         # Create dict with data
         for pair, pair_data in processed.items():
             if not pair_data.empty:
-              pair_data.loc[:, 'buy'] = 0  # cleanup if buy_signal is exist
-              pair_data.loc[:, 'sell'] = 0  # cleanup if sell_signal is exist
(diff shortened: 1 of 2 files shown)


xmatthias:
The stackoverflow answer does actually point out that your version is the "inreliable" one (the one with potentially different results depending on the dataframe format).


Also, your test / screenshot is invalid / not relevant for the code change - you're using an empty dataframe - which is not what is available at this location, but a dataframe of a certain length (candles to backtest) - with multiple columns.

Especially when assigning - using .loc is what the pandas documentation recommends.

kamontat:
@xmatthias I debugging deeper, and finally found root cause.

xmatthias (approved):
Nice catch (seems like we missed this when this check was added for backtesting).

LGTM :+1:

PR #4996, opened 2021-05-22
Fix missing `not` in `empty` check

TheJoeSchr:
See discussing here: https://github.com/freqtrade/freqtrade/pull/4963#discussion_r633457596
seems that request was only partially implemented


xmatthias (approved):
Yeah, i kindof messed that up :laughing:

PR #4998, opened 2021-05-22
patch: add Helper functions for dataframe access via datetime

TheJoeSchr:
adds following public facing helper functions:

`def get_trade_candle(self, trade: 'Trade'):` -> gets candle of `trade.open_date`

`def get_buy_candle(self, trade: 'Trade', timeframe="5m"):` -> gets the candle where the buy decision happend (basically `shift(1)` of open_date)

`find_candle_datetime(self, query_date: datetime, pair: str, now: datetime)` -> finds candle of the given `query_date`

I'm not married to the names. Just tried to keep it somewhat short. 

> There are two hard things in computer science: cache invalidation, naming things​, and off-by-one errors

### DIFF
--- a/freqtrade/strategy/strategy_helper.py
+++ b/freqtrade/strategy/strategy_helper.py
@@ -2,7 +2,7 @@ import pandas as pd
 
 from freqtrade.exchange import timeframe_to_minutes, timeframe_to_prev_date
 from freqtrade.strategy import IStrategy
-from datetime import datetime
+from datetime import datetime, timedelta
 from freqtrade.persistence import Trade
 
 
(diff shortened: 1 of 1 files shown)


freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +86,46 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
 
     # negative stoploss values indicate the requested stop price is higher than the current price
     return max(stoploss, 0.0)
+
+
+class HelperMixin(IStrategy):
```
xmatthias: I'd rather like to see these as helpers (like `merge_informative_pair` above) - not as mixin.

having this as mixin does only complicate things and reduce usability as it's unclear to the user what will happen / what the result will be as it'll have some obscure dependencies and side-effects (it automatically fetches the dataframe (in the strategy's timeframe), which is only one usecase).

`timeframe` and `dataframe` should be arguments for the methods (not assumed to be `self.timeframe` and `get_analyzed_dataframe()`).
Otherwise, you can't use it to get the candle for an informative pair / timeframe - which greatly limits reusability of this.

TheJoeSchr: thanks for the feedback. going for a mixin wasn't my initial plan, but I realised I would need `self` anyways and instead of passing it in, I decided to do a mixin. BUT now a lot of stuff has cleared up after implemening and some refactoring it and together with this input: 
> timeframe and dataframe should be arguments for the methods (not assumed to be self.timeframe and get_analyzed_dataframe()).
Otherwise, you can't use it to get the candle for an informative pair / timeframe - which greatly limits reusability of this.

it will make a lot more sense to go the helper function route.

TBC

TheJoeSchr:
How about that? I refactored the mixin into helper functions

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +85,56 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
 
     # negative stoploss values indicate the requested stop price is higher than the current price
     return max(stoploss, 0.0)
+
+
+def get_custom_dataframe(self, pair: str):
+    dataframe, _ = self.dp.get_analyzed_dataframe(pair, self.timeframe)
+    return dataframe
```
xmatthias: If it's not in a class, there won't be any `self`.

As such - you'll need to remove all references to "self" - and add the eventually necessary parameters to the function head.

The above is valid for all methods below as well.


For that particular method:
with that - this method will do nothing - and can (and should) be inlined into the strategy - so at least this particular function can for sure be removed.

xmatthias (changes requested):
I don't think this way it'll work - it looks like you removed the class definition, and changed the indentation (but not the logic behind).

Having standalone functions means there won't be any `self` to refer to - and no "magic data" that's ingested from the background - but all used data should be passed in as arguments.

TheJoeSchr:
Thanks for the review.

Yeah I see your point and honestly was expecting it to be brought up. My concern is, this function has changed a lot the last month, while freqtrade has been changing as well. It started with using `self.custom_info` and now it's true, it is a oneliner. But I kind of want to be still in control of the Dataframe I'm using. What if e.g. I get a Dataframe passed in that hasn't been analysed and sanitised for future bias. I kind of need to know that.

You see my problem?

xmatthias:
You're contradicting yourself ...

> But I kind of want to be still in control of the Dataframe I'm using

By "secretly ingesting" a dataframe from the background, you're losing control over this fact (or removing the user from this control). 
While YOU know how it works - this won't be true for the user who didn't wrote this - so he's lost control over the dataframe he's using to get the results - and the result will appear "magically".

Think about it this way - maybe the user doesn't want to use the analyzed pair dataframe - but an informative pair he just pulled from dataprovider and added some indicators to?
Nothing is preventing the user from that - except with your architecture / approach, where you force the usage of the pair dataframe (in the pair timeframe) - without real reason.

TheJoeSchr:
👍

TheJoeSchr:
> Think about it this way - maybe the user doesn't want to use the analyzed pair dataframe - but an informative pair he just pulled from dataprovider and added some indicators to?
Nothing is preventing the user from that - except with your architecture / approach, where you force the usage of the pair dataframe (in the pair timeframe) - without real reason.

So I refactored it so that dataframe gets passed in directly. 

With that the `faster` method is kind of obsolete, because I can't depend that dataframe was called via analyzed dataframe, so `dataframe.iloc[-1]` would be against the goal of this pr, mainly to help new users to don't shoot themself in the foot when trying to get a dataframe by time.

so I only see the `find_candle_datetime_safer()` as reliable and valid way to do this. in my tests its about 10-12ms slower with a moderate strategy + pairlist combo.

what to you think?

xmatthias:
> in my tests its about 10-12ms slower with a moderate strategy + pairlist combo.

in what test-scenario? live with 1000 candles - or backtesting with 100_000 candles ... ?

TheJoeSchr:
It's a backtest with 5m candles running from `20200901-20210601` with about 20 pairs.

From this result I don't think performance difference will be much in live with 500-1500 candles?

xmatthias:
No definitely not - with this type of "impact" on this timerange, you'll have trouble measuring it when you only use 500 or 1000 candles.

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +85,57 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
...
+
+def find_candle_datetime_safer(dataframe: pd.DataFrame, query_date: datetime):
+    df = dataframe[['date']].set_index('date')
+
+    try:
+        date_mask = df.index.unique().get_loc(query_date, method='ffill')
+        candle = dataframe.iloc[date_mask]  # use iloc because date_mask maybe :int
+    except KeyError:  # trade.open_date may not exist yet
+        candle = pd.DataFrame(index=dataframe.index)
+    return candle
```
xmatthias: Considering the enhancements @rokups made for on "lookahead bias prevention" - i'm not sure i do see a problem with the "faster" approach - although the speed difference you measured is not big enough to really matter (considering this will only be called for "open trade candles" in backtesting) which way we use it.


Also, if you look at [dataframe access](https://www.freqtrade.io/en/latest/strategy-advanced/#dataframe-access) in the docs - it's using .iloc as well (thanks to the enhancements mentioned above).

The only real danger here (but with both of these methods) is if they are used out of a `populate*()` method - as neither of the 2 versions can really be used for that (you'll not have a current time - and these functions should be populated / used in a vectorized way).

As such - i think we should decide on one - but i'm not sure on the benefits of the "safer" variant ... although i think the exception handler might be wrong `dataframe.index` will return a series with 500+ rows ... but the expectation from this is 1 row i'd think.

rokups: Also `find_candle_datetime_faster` would produce invalid results when we pass `current_time` as `now` parameter without passing it through `timeframe_to_prev_date()`. But even if that was fixed this function is incorrect, because it would return different candles depending on current time.

I do not understand `find_candle_datetime_safer` at all. Seems like it is doing ``candle = dataframe.loc[dataframe.date == candle_date]` except with a dataframe copy, and on error it returns empty dataframe, while on success it returns a `Series` object.

TheJoeSchr:
I would prefer to settle on the `safer` one, because the performance
penalty seem minimal and the `faster` one "breaks" (aka generates look
ahead bias whih may not be obvious for beginners) if Dataframe was not
acquired via a method that trims Dataframe for lookahead. (The one @rokus
added)

And because I refactored it now on @xmatthias suggestions, I can't
guarantee anymore which kind of Dataframe this function gets passed


__

>
>
>

TheJoeSchr:
The exception handle also seems wrong to me, maybe I did a copy and paste
oopsie


__

xmatthias:
@TheJoeSchr do you think it still makes sense to pursue this PR further, or does it no longer make sense so we should close (clean this up)?

TheJoeSchr:
What do you think? I still use it with my own stuff. But I could also just extract it into its own helper library? Do you think there is a need for it? Haven't done much strat development or looked at others recently tbh...

xmatthias:
Personally, i don't think so.

the code in the [advanced strategy docs](https://www.freqtrade.io/en/stable/strategy-advanced/#dataframe-access) explains accessing the dataframe per date well enough - without "magic" behind the scenes.

TheJoeSchr:
Okay, then let's close it

PR #5003, opened 2021-05-23
Indicator plotting improvement

rokups:
- Added a way to specify plot type. Still defaults to "scatter", but we can also specify "bar". Great for MACD histogram 👍🏻 
- Added a way to pass extra arguments to plotly trace constructor. Now we can customize appearance of plotted objects. For example displaying parabolic sar is now possible as we can override `mode` and set `marker_symbol`.

### DIFF
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -96,20 +96,34 @@ def add_indicators(fig, row, indicators: Dict[str, Dict], data: pd.DataFrame) ->
                        Dict key must correspond to dataframe column.
     :param data: candlestick DataFrame
     """
+    plot_kinds = {
+        'scatter': go.Scatter,
+        'bar': go.Bar,
+    }
     for indicator, conf in indicators.items():
         logger.debug(f"indicator {indicator} with config {conf}")
         if indicator in data:
             kwargs = {'x': data['date'],
                       'y': data[indicator].values,
-                      'mode': 'lines',
                       'name': indicator
                       }
-            if 'color' in conf:
-                kwargs.update({'line': {'color': conf['color']}})
-            scatter = go.Scatter(
-                **kwargs
-            )
-            fig.add_trace(scatter, row, 1)
+
+            plot_type = conf.get('type', 'scatter')
+            color = conf.get('color')
+            if plot_type == 'bar':
+                kwargs.update({'marker_color': color or 'DarkSlateGrey',
+                               'marker_line_color': color or 'DarkSlateGrey'})
+            else:
+                if color:
+                    kwargs.update({'line': {'color': color}})
+                kwargs['mode'] = 'lines'
+                if plot_type != 'scatter':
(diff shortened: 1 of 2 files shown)


rokups:
This y-shenanigan may look questionable, but it is useful to implement TTM-Squeeze. Yes, use is limited.. Though it should not be a problem for other usecases i think.

image

Edit: actually nvm, i removed `y` in `kwargs` hack. It was dirty and same thing can be accomplished within strategy itself so it has no reason to exist.

freqtrade/plot/plotting.py
```diff
@@ -96,20 +96,34 @@ def add_indicators(fig, row, indicators: Dict[str, Dict], data: pd.DataFrame) ->
...
+                               'marker_line_color': color or 'DarkSlateGrey'})
+            else:
+                if color:
+                    kwargs.update({'line': {'color': color}})
+                kwargs['mode'] = 'lines'
+                if plot_type != 'scatter':
+                    logger.warning(f'Indicator {indicator} has hnknown plot trace kind {plot_type}'
+                                   f', assuming "scatter".')
+
+            kwargs.update(conf.get('kwargs', {}))
```
xmatthias: i do have one problem with using "kwargs" here.

I don't see plotting with plotly as the future for freqtrade - and the arguments needed for echarts (freqUI) will be different.
All options we have now (except for Area plots) are supported by both plotly (plot modules) and freqUI with the same arguments - using kwargs would be a "pure" plotly function - which would make the plot-config no longer work for freqUI (and might even cause odd side-effects when loading the config from the strategy there).

Apart from that - i'll also need to validate if the optioons here (scatter, bar) are the way it's generated in freqUI - or if that'll again be a missalignment...

rokups: Ah i did not think of it. Could name it `plotly_args` to disambiguate. Would that work? Or maybe just `plotly`?

xmatthias: yeah - maybe 

if the docs (!) contain a hint that these options are only supported for plotly plotting (not for FreqUI) - i guess it'll be fine

freqtrade/plot/plotting.py
```diff
@@ -96,20 +96,34 @@ def add_indicators(fig, row, indicators: Dict[str, Dict], data: pd.DataFrame) ->
...
+            color = conf.get('color')
+            if plot_type == 'bar':
+                kwargs.update({'marker_color': color or 'DarkSlateGrey',
+                               'marker_line_color': color or 'DarkSlateGrey'})
+            else:
+                if color:
+                    kwargs.update({'line': {'color': color}})
+                kwargs['mode'] = 'lines'
+                if plot_type != 'scatter':
+                    logger.warning(f'Indicator {indicator} has hnknown plot trace kind {plot_type}'
```
xmatthias:
```suggestion
                    logger.warning(f'Indicator {indicator} has unknown plot trace kind {plot_type}'
```

xmatthias (approved):
LGTM :+1:

PR #5005, opened 2021-05-23
Fix a rare error during hyperopt

Pascal66:
Solve an error when saving epoch 
```
freqtrade/freqtrade/optimize/hyperopt.py", line 166, in _save_result
    rapidjson.dump(epoch, f, default=str, number_mode=rapidjson.NM_NATIVE)
ValueError: Out of range float values are not JSON compliant
```


xmatthias (approved):
LGTM :+1: 

We'll need to see - i suspect when this happens, `hyperopt-show` will subsequently fail - but as i fail to reproduce this - i can't test that for now...

PR #5014, opened 2021-05-24
BugFix - `hyperopt-show --print-json` include non-optimized params

Rikj000:
## Summary
This fix includes `non-optimized` params in the results when using `hyperopt-show --print-json`

## Quick changelog

Updated `_params_update_for_json()` in the `hyperopt_tools.py` file to also use the `non-optimized` dictionary as a parameter. This is then used to included the `non-optimized` parameters in the json being printed out.

## What's new?
Following results were retrieved using:
```powershell
freqtrade hyperopt-show -n 921 -c ./user_data/mgm-config.json -c ./user_data/mgm-config-private.json --no-header --print-json | tail -n 1 | jq '.' > ./user_data/mgm-config-hyperopt-after/before-fix.json
```
HyperOptShowPrintJsonNonOptimized


xmatthias:
I'm ... sceptical of this (in general, not about the code in particular).

it's different with the regular output, where you have the possibility to say "this was not hyperopted" ... but the json output won't include this information (so all parameters are the same), so it might be missleading...

Rikj000:
> I'm ... sceptical of this (in general, not about the code in particular).
> 
> it's different with the regular output, where you have the possibility to say "this was not hyperopted" ... but the json output won't include this information (so all parameters are the same), so it might be missleading...

If you think that would be an issue then I propose one of following as a way to clarify which parameters are non-optimized to the users. (Example parameter name: `example_param`)
- Include a second entry of the parameter with an underscore behind it's name as following:
```json
{
    "example_param_": "non-optimized",
    "example_param": 73
}
```
- Append `_non_optimized` to the end of the parameter name as following:
```json
{
    "example_param_non_optimized": 73
}
```

nuaimat:
+1 I would love to see this fix merged or alternative options considered

xmatthias:
For future readers:
The command given above (`jq ...`) will **not** work properly in all cases - it's not a documented nor supported way to work with parameters - and will also not cover all cases - so results will vary depending on how hyperopt has been started/ran.

---

Regarding the PR:
CI is failing and there are merge-conflicts. 
I'll not look at it in detail until that's fixed. 
CI is there for a reason - not respecting it (not fixing CI errors) means the author has little interrest to have a PR merged.

Apart from that - i've mentioned (to @Rikj000 ) on discord a few months back how this should be properly implemented - THIS is not the way to do it (i've explicitly asked NOT to document / instruct this way, but instead implement the proper way ...).

This method will fail in some instances / combinations, creating nothing but confusion and frustration by users who see their backtesting results not align with hyperopt results for no good reason.
As such - i'll ask you again - please do not document / instruct users to use hyperopt results this way and if you did already, please remove that, but instead use the documented methods (which means copy/pasting the result to the strategy).


I'm working on a proper implementation for this (i explained this to @Rikj000 how it should be done and was assuming he's working on that which apparently is not the case - but gave up on that this week) - and until then i expect users to follow the documentation and copy/paste the variables into the strategy.

freqtrade/optimize/hyperopt_tools.py
```diff
@@ -106,11 +106,20 @@ def print_epoch_details(results, total_epochs: int, print_json: bool,
...
         if space in params:
             space_params = HyperoptTools._space_params(params, space)
+            space_non_optimized = HyperoptTools._space_params(non_optimized, space)
+            all_space_params = space_params
+
+            # Include non optimized params if there are any
+            if len(space_non_optimized) > 0:
+                for non_optimized_param in space_non_optimized:
+                    if non_optimized_param not in all_space_params:
+                        all_space_params[non_optimized_param] = space_non_optimized[non_optimized_param]
```
xmatthias: i'd think you should get away with `deep_merge_dicts` here (with non-optimized as first parameter) ... which should simplify this logic (but obviously you need to test if the results are as expected)

Rikj000:
The `kwargs merge` seems indeed like a way cleaner implementation for this! :slightly_smiling_face:   
The `buy_cci_max` had `optimize=False` for these tests:
- [include_non_optimised_kwargs_merge.json](https://github.com/freqtrade/freqtrade/files/6673101/include_non_optimised_kwargs_merge_json.log)
- [include_non_optimised_original.json](https://github.com/freqtrade/freqtrade/files/6673102/include_non_optimised_original_json.log)

**Some info about the tests:**
- Command used: `freqtrade hyperopt-show -n 921 -c ./user_data/config.json --no-header --print-json`
- `json` output parsed using [jsonformatter.io](https://www.jsonformatter.io/) for easy reflection
- Uploaded as `.log` files since GitHub doesn't support uploading of `.json` in PR comments

The non-optimized are placed at top of the `params` dictionary now, but I couldn't really think of a good way to include comments in the way you liked. I'm unsure how to parse a python object to jsonc and then print it out using rapidjson :thinking:

xmatthias (changes requested):
Unfortunately, i have to request another round of changes, as it's not doing what you think it does.

It may work in your sample (1 parameter was disabled) - but it's failing if a full space was disabled / not optimized.

From the below (kinda random) output

```
# Buy hyperspace params:
    buy_params = {
        "buy_block": 36,
        "buy_high_lookback": 111,
        "buy_obv": 8,
        "buy_volume_mean": 25,
        "buy_vwap": 120,
    }

    # Sell hyperspace params:
    sell_params = {
        "sell_low_lookback": 110,  # value loaded from strategy
        "sell_whatever": 100,  # value loaded from strategy
    }

    # ROI table:
    minimal_roi = {
        "0": 0.344,
        "342": 0.191,
        "672": 0.045,
        "1539": 0
    }
```

 i get the following result (with this PR, obviously):
```
{
  "params": {
    "buy_block": 36,
    "buy_high_lookback": 111,
    "buy_obv": 8,
    "buy_volume_mean": 25,
    "buy_vwap": 120
  },
  "minimal_roi": {
    "0": 0.344,
    "342": 0.191,
    "672": 0.045,
    "1539": 0
  }
}
```

As you'll notice, the whole sell-space is missing - and this PR is trying to address exactly that ... and as such, is unfortunately not complete.

Rikj000:
> As you'll notice, the whole sell-space is missing - and this PR is trying to address exactly that ... and as such, is unfortunately not complete.

Ah I see, my excuses for missing that, looking into fixing that up right away :slightly_smiling_face:

Rikj000:
### **HyperOpt command `buy` space only:**
```powershell
freqtrade hyperopt -s SwingHighToSkyHyperStrategy -c ./user_data/config.json --spaces buy -e 100 --hyperopt-loss WinRatioAndProfitRatioLoss --timerange 20210101-20210131
```

**HyperOpt output:**
```python
# Buy hyperspace params:
buy_params = {
    "buy_use_cci": True,
    "buy_cci_max": -112,  # value loaded from strategy
}

# Sell hyperspace params:
sell_params = {
    "sell_cci_min": -190,  # value loaded from strategy
    "sell_use_cci": "no-use",  # value loaded from strategy
}
```

**JSON extract command:**
```powershell
freqtrade hyperopt -s SwingHighToSkyHyperStrategy -c ./user_data/config.json --spaces buy -e 100 --hyperopt-loss WinRatioAndProfitRatioLoss --timerange 20210101-20210131
```

**JSON Results after fix:**
```json
{
   "params": {
      "buy_cci_max": -112,
      "buy_use_cci": true,
      "sell_cci_min": -190,
      "sell_use_cci": "no-use"
   }
}
```

*(Also ended up flipping the dictionaries on the kwargs merge line, now the results format align with HyperOpts CLI output! :tada: If you don't like it then I can just flip them again so we get above results again)*
**JSON Results after flipping of dictionaries in merge line:**
```json
{
   "params": {
      "buy_use_cci": true,
      "buy_cci_max": -112,
      "sell_cci_min": -190,
      "sell_use_cci": "no-use"
   }
}
```

### **Spaces `all` for comparison**

```python
# Buy hyperspace params:
buy_params = {
    "buy_use_cci": False,
    "buy_cci_max": -112,  # value loaded from strategy
}

# Sell hyperspace params:
sell_params = {
    "sell_cci_min": 251.03418,
    "sell_use_cci": "use",
}

# ROI table:
minimal_roi = {
    "0": 0.499,
    "479": 0.12,
    "1001": 0.072,
    "2313": 0
}

# Stoploss:
stoploss = -0.346

# Trailing stop:
trailing_stop = True
trailing_stop_positive = 0.014
trailing_stop_positive_offset = 0.11
```

```json
{
   "params": {
      "buy_use_cci": "False",
      "buy_cci_max": -112,
      "sell_cci_min": 251.03417551010136,
      "sell_use_cci": "use"
   },
   "minimal_roi": {
      "0": 0.499,
      "479": 0.12,
      "1001": 0.072,
      "2313": 0
   },
   "stoploss": -0.346,
   "trailing_stop": "True",
   "trailing_stop_positive": 0.014,
   "trailing_stop_positive_offset": 0.11,
   "trailing_only_offset_is_reached": "True"
}
```

Here we can see that the non-optimized results are still being added in correctly!   
However I noticed another bug surface here, once parameters have been touched by HyperOpt they tend to become strings instead of their original boolean type. I recall that this was the case for integers too, but this already seems to be resolved now!

xmatthias (approved):
The "string" problem is a slightly different one - and probably related to your configuration.
Using `CategoricalParameter([True, False], default=True)` works fine for me (in both / all cases) - so most likely, you're passing a string instead of True to the space when hyperopting.

As such - for now i don't see that as a problem - as you defined it as string - it's treated as a string.

For this PR - LGTM :+1:

PR #5032, opened 2021-05-27
backtest proposal: less optimistic trailing stops

brookmiles:
## Summary

in backtesting, calculate trailing stop sell price as being half way between the stop loss and low, preventing extremely optimistic results when using tight stop loss values

## What's new?

Before
20210527-202710-msedge

After

20210527-202941-msedge

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -224,6 +224,13 @@ class Backtesting:
                 # sell at open price.
                 return sell_row[OPEN_IDX]
 
+            if sell.sell_type == SellType.TRAILING_STOP_LOSS:
+                # trailing stop could potentially occur at any price between
+                # trade.stop_loss and candle low, so use the mid point
+                # to prevent extremely optimistic results with tight stops
+                stop_price = (trade.stop_loss + sell_row[LOW_IDX]) / 2
+                return stop_price
+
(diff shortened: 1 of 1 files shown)


xmatthias:
looking at only the first candle - it may even be that trailing never even activated - but that the "initial" stoploss triggered (assuming a move from open to low - before then going to high).

The only way to cover this would be by evaluating stoploss before adjusting trailing stop - which will bring other problems (stopped out before the value is adjusted ...) - so i'm not sure that should be touched as well.

For the remaining candles, it looks okish / better than before for sure ... but to be honest, i'd apreciate some other looks at it too

bzed:
imho backtesting is too optimistic at various places: for example it assumes that the trade opening is the open value of the next candle. That will never (?) be the case. I think some repeatable randomness should be introduced at the opening of the trade, then the trailing/stoploss values could be calculated based on the now more realistic trade open.

xmatthias:
Does this PR still make sense with #5108 implemented?

@rokups @brookmiles - opinions ... would be great :)

rokups:
Nope. #5108 implements worst possible case behavior already, while this PR would result in better backtest results.

brookmiles:
Thanks for the work on a more thorough solution @rokups !

Sounds like this PR can go away.

PR #5042, opened 2021-05-28
Removed binance.je from exchange specific notes

blacklightpy:

Yes

## Summary

Binance Jersey is deprecated, so I think it should be removed from freqtrade.

I haven't removed it from the code, but I thought this should give a starting point.


xmatthias (approved):
Great catch :+1: 
i thought i had removed that a while ago ...

PR #5059, opened 2021-05-31
Add show Pairlists Whitelist on RFC and Telegram

ntsd:
Because the whitelist command only shows active whitelists.

## Summary

to make the RFC whitelist API response pairlistsWhitelist from dynamic pairlists

```JSON

{
    "whitelist": ["ETH/BTC"],
    "length": 1,
    "method": [{
		"method": "VolumePairList",
        "number_assets": 4,
	}],
	"pairlistsWhitelist": ["ETH/BTC", "LTC/BTC", "XRP/BTC", "NEO/BTC"]
}

```

and add telegram bot showing pairlists whitelist in the line after `/whitelist`

## Quick changelog

- Add pairlists whitelist (API whitelist) response from RFC API
- Add bot telegram output current pairlists whitelist

## What's new?

Add showing current API whitelist after filters.

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -716,6 +716,8 @@ class RPC:
                'length': len(self._freqtrade.active_pair_whitelist),
                'whitelist': self._freqtrade.active_pair_whitelist
                }
+        if self._freqtrade.pairlists.whitelist:
+            res['pairlistsWhitelist'] = self._freqtrade.pairlists.whitelist
         return res
 
     def _rpc_blacklist(self, add: List[str] = None) -> Dict:
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -853,6 +853,8 @@ class Telegram(RPCHandler):
 
             message = f"Using whitelist `{whitelist['method']}` with {whitelist['length']} pairs\n"
             message += f"`{', '.join(whitelist['whitelist'])}`"
+            if whitelist['pairlistsWhitelist']:
+                message += f"`from current pairlists whitelist {', '.join(whitelist['pairlistsWhitelist'])}`"
 
             logger.debug(message)
             self._send_msg(message)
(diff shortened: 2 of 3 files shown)


xmatthias:
To be honest - for now i don't see why we would need this.

The difference between `.whitelist` and `.active_pair_whitelist` is marginal (only open trades are added should the pair have disappeared from the whitelist in between) - so i think this will bloat the output of the `/whitelist` commands while only providing confusion - no real benefit or added information to the user.


Maybe you have a specific reason for this (what this provides to you? why you think it's needed / currently missing?) you didn't outline above - but with the description provided for now - i'll probably reject this as i don't see the point in having this information (in 99% of the cases, it'll print the same list twice).

ntsd:
> To be honest - for now i don't see why we would need this.
> 
> The difference between `.whitelist` and `.active_pair_whitelist` is marginal (only open trades are added should the pair have disappeared from the whitelist in between) - so i think this will bloat the output of the `/whitelist` commands while only providing confusion - no real benefit or added information to the user.
> 
> Maybe you have a specific reason for this (what this provides to you? why you think it's needed / currently missing?) you didn't outline above - but with the description provided for now - i'll probably reject this as i don't see the point in having this information (in 99% of the cases, it'll print the same list twice).

Actually, it's my own feature request Because when the bot start I don't know how to check the current active pairlists to make sure that my filters work. For now, it only shows the holding pairlists. What if only for API requests not the telegram bot? or I can create another Telegram bot command for this one.

image

xmatthias:
To be honest - best use the `freqtrade test-pairlist [--config whateverconfig]` command to test your pairlist.

It's specifically designed for this.
Alternatively (if you want it via telegram) - use a "0 max_open_trades" configuration and a in-memory database (`freqtrade trade --db-url sqlite:// --strategy ... `) to start with an empty slate until your configuration is stabilized.

What you're providing above does "Not really" give you what you need i think - at least not in a very simple way.
The big downside is - you're now Messaging everyone who has a working configuration the same whitelist twice - if people use a 50 or 100 pairlist, this will become unreadable.

PR #5067, opened 2021-06-01
Add MarketCapPairList.

bzed:
## Summary
After recurring requests on #strategy-chat, this implements a pairlist based on market capitalization.
Data is retrieved from the coingecko API https://www.coingecko.com/en/api

As I have absolutely no clue about the used unit testing framework, there are no unit tests included. Sorry :)

## Quick changelog

- Add MarketCapPairList

## What's new?

```
2021-06-01 21:33:24,645 - MarketCapPairList - INFO - Searching 200 pairs: ['1INCH/USD', 'AAVE/USD', 'ALCX/USD', 'ALPHA/USD', 'AMPL/USD', 'ASD/USD', 'AXS/USD', 'BAL/USD', 'BAND/USD', 'BAT/USD', 'BCH/USD', 'BNB/USD', 'BNT/USD', 'BTC/USD', 'CEL/USD', 'CHZ/USD', 'COMP/USD', 'CRO/USD', 'CRV/USD', 'CUSDT/USD', 'DAI/USD', 'DENT/USD', 'DOGE/USD', 'ENJ/USD', 'ETH/USD', 'FTM/USD', 'FTT/USD', 'GRT/USD', 'GT/USD', 'HNT/USD', 'HT/USD', 'LEO/USD', 'LINK/USD', 'LRC/USD', 'LTC/USD', 'MATIC/USD', 'MKR/USD', 'OKB/USD', 'OMG/USD', 'OXY/USD', 'PAXG/USD', 'PROM/USD', 'PUNDIX/USD', 'RAY/USD', 'REEF/USD', 'REN/USD', 'RSR/USD', 'RUNE/USD', 'SAND/USD', 'SHIB/USD', 'SKL/USD', 'SNX/USD', 'SOL/USD', 'SRM/USD', 'STMX/USD', 'SUSHI/USD', 'TRX/USD', 'UNI/USD', 'USDT/USD', 'WAVES/USD', 'WBTC/USD', 'WRX/USD', 'XRP/USD', 'YFI/USD', 'ZRX/USD']
```

### DIFF
--- a/freqtrade/plugins/pairlist/MarketCapPairList.py
+++ b/freqtrade/plugins/pairlist/MarketCapPairList.py
@@ -1,0 +1,263 @@
+"""
+MarketCap PairList provider
+
+Provides dynamic pair list based on market cap
+"""
+import logging
+from typing import Any, Dict, List
+
+from cachetools.ttl import TTLCache
+from pandas.core.common import flatten
+from pycoingecko import CoinGeckoAPI
+
+from freqtrade.exceptions import OperationalException
+from freqtrade.misc import chunks
+from freqtrade.plugins.pairlist.IPairList import IPairList
+
+
+logger = logging.getLogger(__name__)
+
+
+class MarketCapPairList(IPairList):
+    def __init__(self, exchange, pairlistmanager,
+                 config: Dict[str, Any], pairlistconfig: Dict[str, Any],
+                 pairlist_pos: int) -> None:
+        super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
+
+        if 'number_assets' not in self._pairlistconfig:
+            raise OperationalException(
+                '`number_assets` not specified. Please check your configuration '
+                'for "pairlist.config.number_assets"')
+        if (not type(self._pairlistconfig['number_assets']) is int) or (
+                int(self._pairlistconfig['number_assets']) <= 0):
+            raise OperationalException(
+                '"number_assets" should be a positive integer. '
+                'Please edit your config and restart the bot.'
+            )
+        self._number_pairs = int(self._pairlistconfig['number_assets'])
+
+        self._stake_currency = config['stake_currency']
+        self._refresh_period = self._pairlistconfig.get('refresh_period', 6*60*60)
+        self._marketcap_cache: TTLCache = TTLCache(maxsize=1, ttl=self._refresh_period)
+        self._cg = CoinGeckoAPI()
+        self._marketcap_ranks: Dict[str, Any] = {}
+        self._coins_list: Dict[str, str] = {}
+
+        if not self._exchange.exchange_has('fetchTickers'):
+            raise OperationalException(
+                'Exchange does not support dynamic whitelist. '
+                'Please edit your config and restart the bot.'
+            )
+
+    @property
+    def needstickers(self) -> bool:
+        """
+        Boolean property defining if tickers are necessary.
+        If no Pairlist requires tickers, an empty Dict is passed
+        as tickers argument to filter_pairlist
+        """
+        return True
+
+    def short_desc(self) -> str:
+        """
+        Short whitelist method description - used for startup-messages
+        """
+        return f"{self.name} - top {self._pairlistconfig['number_assets']} marketcap pairs."
+
+    def _request_coins_list(self):
+        """
+        Update the symbol -> coin id mapping. Coin ids are needed to request market data
+        for single coins.
+        """
+        coins_list = self._cg.get_coins_list()
+        for x in coins_list:
+            # yes, lower() is needed as there are a few uppercase symbols
+            symbol = x['symbol'].lower()
+            cid = x['id']
+
+            # ignore binance-peg-
+            if 'binance-peg-' in cid:
+                continue
+            # ignore -bsc
+            if cid.endswith('-bsc'):
+                continue
+            # yes, not all symbols are unique.
+            if symbol in self._coins_list:
+                # in a few cases the id matches the symbol
+                # we have that version already, keep it.
+                if self._coins_list[symbol] == symbol:
+                    continue
+                # the new one matches, use it.
+                elif cid == symbol:
+                    self._coins_list[symbol]
+                else:
+                    self._coins_list[symbol] = list(flatten([self._coins_list[symbol], cid]))
+            else:
+                self._coins_list[symbol] = cid
+
+    def _request_marketcap_ranks(self):
+        """
+        To keep the request count low, download the most common coins.
+        Then update all coins that where not in the top 250 ranks.
+        """
+        ids = []
+
+        # there might be some stuff to add here..
+        # probably also needs to be configurable and/or have
+        # exchange specific settings
+        symbol_map = {
+                'acm': 'ac-milan-fan-token',
+                'ant': 'aragon',
+                'atm': 'atletico-madrid',
+                'bat': 'basic-attention-token',
+                'bts': 'bitshares',
+                'comp': 'compound-governance-token',
+                'dego': 'dego-finance',
+                'eps': 'ellipsis',
+                'grt': 'the-graph',
+                'hnt': 'helium',
+                'hot': 'holotoken',
+                'iota': 'miota',
+                'lit': 'litentry',
+                'luna': 'terra-luna',
+                'mask': 'mask-network',
+                'mdx': 'mdex',
+                'mir': 'mirror-protocol',
+                'og': 'og-fan-token',
+                'one': 'harmony',
+                'pax': 'paxos-standard',
+                'pnt': 'pnetwork',
+                'rune': 'thorchain',
+                'sand': 'the-sandbox',
+                'stx': 'blockstack',
+                'super': 'superfarm',
+                'tct': 'tokenclub',
+                'trb': 'tellor',
+                'tru': 'truefi',
+                'trx': 'tron',
+                'uni': 'uniswap',
+                'wing': 'wing-finance',
+                }
+
+        for symbol in self._marketcap_ranks.keys():
+            # symbol needs mapping
+            if symbol in symbol_map:
+                ids.append(symbol_map[symbol])
+            elif symbol in self._coins_list:
+                _id = self._coins_list[symbol]
+                # symbol is not unique (sigh!)
+                if type(_id) == list:
+                    pair = symbol.upper() + f'/{self._stake_currency}'
+                    _id_text = ', '.join(_id)
+                    self.log_once(f'Symbol for {pair} is not unique on coingecko '
+                                  f'({_id_text}), dropping.', logger.warning)
+                    continue
+                ids.append(_id)
+            elif symbol in self._coins_list.values():
+                # try to add some symbols automatically
+                symbol_map[symbol] = [k for k, v in self._coins_list.items() if v == symbol][0]
+                ids.append(symbol)
+
+        # reverse map
+        rev_symbol_map = {v: k for k, v in symbol_map.items()}
+
+        _marketcap_ranks = {}
+
+        # seems coingecko limits the number of ids to 52, using 50.
+        for _ids in chunks(ids, 50):
+            base_marketkaps = self._cg.get_coins_markets(
+                    vs_currency='usd',
+                    ids=','.join(_ids),
+                    order='market_cap_desc',
+                    per_page=len(ids),
+                    sparkline=False,
+                    page=1)
+            for x in base_marketkaps:
+                # same here, keep symbol lowercase.
+                _symbol = x['symbol'].lower()
+                if _symbol in rev_symbol_map:
+                    _symbol = rev_symbol_map[_symbol]
+                _marketcap_ranks[_symbol] = x['market_cap_rank']
+
+        for x in self._marketcap_ranks.keys():
+            if x not in _marketcap_ranks:
+                _marketcap_ranks[x] = None
+        self._marketcap_ranks = _marketcap_ranks
+
+    def update_marketcap_ranks(self, symbols: List[str]):
+        """
+        Updates the dict containing the marketcap ranks of the requested
+        list of symbols, if needed.
+        """
+        marketcaps_uptodate = self._marketcap_cache.get('marketcaps_uptodate')
+        startup = not (self._marketcap_ranks and self._coins_list)
+
+        for symbol in symbols:
+            if symbol.lower() not in self._marketcap_ranks:
+                marketcaps_uptodate = False
+                self._marketcap_ranks[symbol.lower()] = None
+        if not marketcaps_uptodate:
+            try:
+                self._request_coins_list()
+                self._request_marketcap_ranks()
+            except Exception as e:
+                if startup:
+                    raise OperationalException(
+                        f'Failed to download marketcap data from coingecko: {e}'
+                    )
+                else:
+                    self.log_once(
+                            'Failed to update marketcap data from coingecko: .'
+                            f'{e}. Using old data.',
+                            logger.warning)
+            else:
+                self._marketcap_cache['marketcaps_uptodate'] = True
+
+    def gen_pairlist(self, tickers: Dict) -> List[str]:
+        """
+        Generate the pairlist
+        :param tickers: Tickers (from exchange.get_tickers()). May be cached.
+        :return: List of pairs
+        """
+
+        # using the pairlist from the exchange works, but if the list is long, it will create a very
+        # long startup time due to coingecko rate limits. Might be worth to add an config option
+        # with a big warning.
+        pairlist = self._whitelist_for_active_markets(
+               self.verify_whitelist(self._config['exchange']['pair_whitelist'], logger.info))
+        return self.filter_pairlist(pairlist, tickers)
+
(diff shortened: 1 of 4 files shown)


freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,92 @@
...
+        Generate the pairlist
+        :param tickers: Tickers (from exchange.get_tickers()). May be cached.
+        :return: List of pairs
+        """
+        # Generate dynamic whitelist
+        # Must always run if this pairlist is not the first in the list.
+        pairlist = self._pair_cache.get('pairlist')
+        if not pairlist:
+            resp = requests.get(self._url)
+            data = resp.json()
```
xmatthias: i have 2 issues with this code.

First of all - i don't think we should use requests for this directly - but a library (we already depend on pycoingecko - so we can use that without adding new dependencies). 
That way, we don't have to care should the API url ever change (that's then responsibility of the library). Also, the library will handle some basic errors for us.

Then - This implements no error handling whatsoever - so if coingecko is down or returns a non-json object (or a json object not containing what you expect), it'll crash the bot - which is for sure a no-go.

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,92 @@
...
+        # Use the incoming pairlist.
+        pairlist = [k for k in tickers.keys() if k in pairlist]
+        # Validate whitelist to only have active market pairs
+        pairlist = self._whitelist_for_active_markets(pairlist)
+        pairlist = self.verify_blacklist(pairlist, logger.info)
+        # Limit pairlist to the requested number of pairs
+        pairlist = pairlist[:self._number_pairs]
+        self.log_once(f"Searching {self._number_pairs} pairs: {pairlist}", logger.info)
+
+        return pairlist
```
xmatthias: I don't think this does what it advertises.

While the generator (only called if this is the first pairlist) uses marketcap to sort the pairlist - this does not sort by this, but  simply returns the already existing pairlist without sorting it by marketcap (which is what i'd expect / what the documentation entry states).

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class MarketCapPairList(IPairList):
+    def __init__(self, exchange, pairlistmanager,
+                 config: Dict[str, Any], pairlistconfig: Dict[str, Any],
+                 pairlist_pos: int) -> None:
+        super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
+
+        if 'number_assets' not in self._pairlistconfig:
+            raise OperationalException(
```
xmatthias, line 28: I think you'll also want to check if `number_assets` is numeric / an integer - by providing a non-int entry, i'm pretty sure i can break the response from coingecko ...

xmatthias (changes requested):
To "properly" implement this (so it does what it advertises) - i think you'll need to:

* cache the return value from coingecko
* Extract the logic to sort by marketcap to a seperate function, that's called by both `gen_pairlist` and `filter_pairlist` - and sorts the input pairlist properly
* add error-handling to the coingecko response (that'll also be necessary when using the coingecko library)

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: :+1: I'll change that.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Good catch, the pairlist was sorted actually, but I'm iterating trough the tickerlist. I've fixed that so its useful until the other things are fixed...

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,92 @@
...
+    def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
+        """
+        Filters and sorts pairlist and returns the whitelist again.
+        Called on each bot iteration - please use internal caching if necessary
+        :param pairlist: pairlist to filter or sort
+        :param tickers: Tickers (from exchange.get_tickers()). May be cached.
+        :return: new whitelist
+        """
+        # Use the incoming pairlist.
+        pairlist = [k for k in pairlist if k in tickers.keys()]
```
xmatthias: You're running under the assumption that this pairlist is the FIRST pairlist in your list - which is not true / cannot be relied uppon.


Assuming you use VolumePairlist (get the highest volume pairs) - followed by MarketCapPairlist to sort the top 30 volumePairs based on their market cap.

This means - the "incomming" pairlist (`filter_pairlist`) will not be filterd based on MarketCap - but by "something" else (in the above sample by volume - but other options are possible easily).

As such - this `filter_pairlist` will not do anything - as it does completely ignore the MarketCap of the pairs - so the sequence of the pairs will not be changed.

bzed:
I think I've addressed all issues. Needs a bit more testing and I want to see what the CI says.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Indeed failed to understand how the pairlists interact. Fixed.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Check added.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Also fixed.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: reimplemented using pycoingecko.

The errors from pycoingecko are less specific than I'd like them to be unfortunately, but I think I've handled it in a sane way. pycoingecko retries requests in case they fail, so I haven't implemented that.

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        To keep the request count low, download the most common coins.
+        Then update all coins that where not in the top 250 ranks.
+        """
+        ids = []
+
+        # there might be some stuff to add here..
+        # probably also needs to be configurable and/or have
+        # exchange specific settings
+        symbol_map = {
```
xmatthias, line 108: While i understand why you do this - we will not be hand/hard coding these tokens anywhere in the freqtrade repository.

The list will be outdated / broken the moment we merge this (exchanges list tokens all the time) - and maintaining this list is a (stupid, manual, per exchange) process that i'm not willing to take, as this will be a maintenance nightmare.

Maybe write a logic instead which simply excludes all coins with duplicate coingecko ID's from this filter - as your logic of mapping might be correct for binance - but will be wrong if i use this same list against another exchange.

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
+        """
+        Filters and sorts pairlist and returns the whitelist again.
+        Called on each bot iteration - please use internal caching if necessary
+        :param pairlist: pairlist to filter or sort
+        :param tickers: Tickers (from exchange.get_tickers()). May be cached.
+        :return: new whitelist
+        """
+        symbols = [p.replace(f'/{self._stake_currency}', '').lower() for p in pairlist]
```
xmatthias, line 238: Please use `exchange.get_pair_base_currency()` instead. 
While most exchanges use the format of `{base}/{quote}` - that's not guaranteed - and assuming this will break future usecases.

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+            # yes, lower() is needed as there are a few uppercase symbols
+            symbol = x['symbol'].lower()
+            cid = x['id']
+
+            # ignore binance-peg-
+            if 'binance-peg-' in cid:
+                continue
+            # ignore -bsc
+            if cid.endswith('-bsc'):
+                continue
```
xmatthias, line 83: What's the reason to exclude these?
this seems like random "i don't like these" exclusions for now - so if there's a reason, maybe put a reason as comment....

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+        :param tickers: Tickers (from exchange.get_tickers()). May be cached.
+        :return: List of pairs
+        """
+
+        # using the pairlist from the exchange works, but if the list is long, it will create a very
+        # long startup time due to coingecko rate limits. Might be worth to add an config option
+        # with a big warning.
+        pairlist = self._whitelist_for_active_markets(
+               self.verify_whitelist(self._config['exchange']['pair_whitelist'], logger.info))
+        return self.filter_pairlist(pairlist, tickers)
```
xmatthias, line 228: I guess this will then not be a pairlist provider - but will be a Filter instead.
This method does not make sense this way as it replicates the logic of StaticPairlist.

You can therefore get the same with the following config:

```
    "pairlists": [
        {"method": "StaticPairList"},
        {"method": "MarketCapPairList"},
},
```

We have many lists like that - so there's no harm in not providing this functionality if it doesn't make sense.

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+
+        # seems coingecko limits the number of ids to 52, using 50.
+        for _ids in chunks(ids, 50):
+            base_marketkaps = self._cg.get_coins_markets(
+                    vs_currency='usd',
+                    ids=','.join(_ids),
+                    order='market_cap_desc',
+                    per_page=len(ids),
+                    sparkline=False,
+                    page=1)
```
xmatthias, line 174: Does Pycoingecko provide a "backoff" / rate limit functionality? 
if not, this will cause rate-limit issues for sure as it's "banging" against the API for every coin (in batches of 50, but assume i use all USDT pairs from binance - it'll be quite some requests).

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+            for x in base_marketkaps:
+                # same here, keep symbol lowercase.
+                _symbol = x['symbol'].lower()
+                if _symbol in rev_symbol_map:
+                    _symbol = rev_symbol_map[_symbol]
+                _marketcap_ranks[_symbol] = x['market_cap_rank']
+
+        for x in self._marketcap_ranks.keys():
+            if x not in _marketcap_ranks:
+                _marketcap_ranks[x] = None
```
xmatthias, line 184: What's the reason for this loop?

Looking at the rest of this function, i'd think this is to remove coins no longer in the list which had been in the list before.
However, by "assigning" the new list to `self._marketcap_ranks` in the next line, i'd not see this as necessary - as you'll always have to assume that there will be coins you didn't get the market-cap for...

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+        Updates the dict containing the marketcap ranks of the requested
+        list of symbols, if needed.
+        """
+        marketcaps_uptodate = self._marketcap_cache.get('marketcaps_uptodate')
+        startup = not (self._marketcap_ranks and self._coins_list)
+
+        for symbol in symbols:
+            if symbol.lower() not in self._marketcap_ranks:
+                marketcaps_uptodate = False
+                self._marketcap_ranks[symbol.lower()] = None
```
xmatthias, line 198: That's a ... rather dangerous approach i think.

Assuming i use a coin in my whitelist which isn't known to coingecko yet (for whatever reason) - you'd force a reload / refresh on every call to this function simply because the coin isn't known (and probably won't be known for days ...)

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,263 @@
...
+                    raise OperationalException(
+                        f'Failed to download marketcap data from coingecko: {e}'
+                    )
+                else:
+                    self.log_once(
+                            'Failed to update marketcap data from coingecko: .'
+                            f'{e}. Using old data.',
+                            logger.warning)
+            else:
+                self._marketcap_cache['marketcaps_uptodate'] = True
```
xmatthias, line 214: considering that `self._marketcap_cache` is a TTLCache variable - i'd think you'll want to store the marketcap_ranks there so it'll expire.
obviously a helper-variable will work too - but it's a rather clumsy approach i think, which probably isn't necessary (not once you looked at the rest of the comments, anyway)

xmatthias (changes requested):
The hardcoded list of id's is a no-go - not something i'll be comfortable to support going forward - so that's something i'll clearly not be able to accept.

Thats bound to fail with different exchanges and will be a maintenance nightmare - as the list is probably only half-complete now - or if it is complete now, will be outdated in a week from now.

There's quite some other comments too - but some are questions, and some proposals how to make it better.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: As mentioned in #5080 it might sense to create a better mapping and cache it somewhere. Lets discuss this there and then I'll just use the code.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Basically drops a lot of duplicate smbols from the list. Another thing that will be fixed with #5080

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Probably the outcome of mysql switching between the idea of using only symbols from the marketcap list and then filtering the exchange pairs based on that - or sorting the pairs from the exchange by market cap.
Both versions might be useful. Or the option to limit pairs to have a minimum marketcap rank.

The first idea from the strategy-chat group was to use the top marketcap pairs only.
This should probably configurable - what do you think?

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: https://github.com/w3f/staking-rewards-collector/issues/9#issuecomment-745967472 says the current limit is 60 requests per minute, which would be 3000 pairs. Its hard to hit that, but I could add some backoff.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: I'm keeping unknown symbols in the list, so the list is not refreshed on every function call.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: That would be the case if I wouldn't leave them in the list - see the comment above.

freqtrade/plugins/pairlist/MarketCapPairList.py
xmatthias: i know what their docs say - but testing it myself i came to ~8 requests per minute ... which is quite a stretch from the 60 requests they document ...

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Well I think it is necessary, because if the cache is being invalidated you'd have to update the data from coingecko. If coingecko is not available for whatever reason, this will fail badly.
Storing the list in the class and just tracking allows ft to live with the old data until coingecko is back again.
I definitely do not want a hard dependency on something that is not the exchange. Using the TTLCache for that would introduce such a dependency. As far as I understand it the ttlcache code just drops the data as soon as the ttl is expired.

xmatthias: Works for me :+1: thanks for the explanation on your thoughs behind it.

freqtrade/plugins/pairlist/MarketCapPairList.py
bzed: Their docs say 10 per second, which is also not uptodate unfortunately. 50 in a minute worked for me, if you spread the requests over the minute and don't make 50 in the first few seconds.

Malaccoda (changes requested):
> I guess this will then not be a pairlist provider - but will be a Filter instead.
This method does not make sense this way as it replicates the logic of StaticPairlist.

You can therefore get the same with the following config:

```
    "pairlists": [
        {"method": "StaticPairList"},
        {"method": "MarketCapPairList"},
},
```

We have many lists like that - so there's no harm in not providing this functionality if it doesn't make sense.

xmatthias:
Closing this in favor of #9659 ... which is a more uptodate version of this (also simpler in details - and making way fewer API calls to coingecko).

PR #5071, opened 2021-06-02
plot-profit: Make "auto-open" HTML result optional

janoskut:
## Summary
For better tool processing, make the "auto-open" function of the `freqtrade plot-profit` result optional

## Quick changelog

- Auto-open HTML result of `plot-profit` sub-command is now optional
- `plot-profit --auto-open` argument added to substitute the original behavior

## What's new?
For tool processing of the result of `plot-profit`, it is disadvantageous to automatically open the generated HTML file.

Before:
`freqtrade plot-profit` -> opens generated HTML file in system application (browser).
After:
`freqtrade plot-profit` -> just generates the HTML result, doesn't open
`freqtrade plot-profit --auto-open` -> opens generated HTML file in system application (browser). (original behavior)


xmatthias (approved):
Nice! 
LGTM :+1:

PR #5099, opened 2021-06-07
Addition of the maximum drawdown to the hyperoptimization

bgouvea:
## Summary
Addition of the maximum drawdownm to the hyperoptimization results table:
ho-result-table-drawdown-column

## Quick changelog

- Changed function that displays table of optimization results

## What's new?
Drawdown seems to me to be one of the fundamental indicators of any strategy, so having access to it during the optimization process can save time.

### DIFF
--- a/freqtrade/optimize/hyperopt_tools.py
+++ b/freqtrade/optimize/hyperopt_tools.py
@@ -233,18 +233,20 @@ class HyperoptTools():
                              'results_metrics.winsdrawslosses',
                              'results_metrics.profit_mean', 'results_metrics.profit_total_abs',
                              'results_metrics.profit_total', 'results_metrics.holding_avg',
-                             'loss', 'is_initial_point', 'is_best']]
+                             'loss', 'is_initial_point', 'is_best', 'results_metrics.max_drawdown',
+                             'results_metrics.max_drawdown_abs']]
         else:
             # Legacy mode
             trials = trials[['Best', 'current_epoch', 'results_metrics.trade_count',
                              'results_metrics.winsdrawslosses',
                              'results_metrics.avg_profit', 'results_metrics.total_profit',
                              'results_metrics.profit', 'results_metrics.duration',
-                             'loss', 'is_initial_point', 'is_best']]
+                             'loss', 'is_initial_point', 'is_best', 'results_metrics.max_drawdown',
+                             'results_metrics.max_drawdown_abs']]
 
         trials.columns = ['Best', 'Epoch', 'Trades', ' Win Draw Loss', 'Avg profit',
-                          'Total profit', 'Profit', 'Avg duration', 'Objective',
-                          'is_initial_point', 'is_best']
+                          'Total profit', 'Profit', 'Avg duration', 'Max Drawdown', 'max_drawdown_abs',
+                          'Objective', 'is_initial_point', 'is_best']
         trials['is_profit'] = False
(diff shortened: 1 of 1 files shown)


freqtrade/optimize/hyperopt_tools.py
```diff
@@ -233,18 +233,19 @@ def get_result_table(config: dict, results: list, total_epochs: int, highlight_b
...
                              'loss', 'is_initial_point', 'is_best']]
         else:
             # Legacy mode
             trials = trials[['Best', 'current_epoch', 'results_metrics.trade_count',
                              'results_metrics.winsdrawslosses',
                              'results_metrics.avg_profit', 'results_metrics.total_profit',
-                             'results_metrics.profit', 'results_metrics.duration',
-                             'loss', 'is_initial_point', 'is_best']]
+                             'results_metrics.profit', 'results_metrics.duration', 'results_metrics.max_drawdown',
+                             'results_metrics.max_drawdown_abs', 'loss', 'is_initial_point', 'is_best']]
```
xmatthias: without even looking at what ci says (i know it'll fail) - this will kill you as these fields are not available (and can't be).

"legacy" mode is for when the backtest-results were not stored (it's used by show_hyperopt / list-hyperopt commands if using a pickle file) - so this data is not available as it hasn't been stored (which only changed a version or 2 ago).

I suspect you'll need to make formatting of this optional (not displaying the column if we're in "legacy mode" - otherwise it'll display an empty column (as you'll need to add that column first) - which won't be much better either.

bgouvea: Thanks for the clarification on legacy mode. I was a little doubtful about which path to follow for correct implementation, I thought that the code would be better duplicating the definition of the columns as a whole to avoid later manipulation of the dataframe columns.

bgouvea: The second option would be to temporarily create the columns and then remove them (as in winsdrawslosses column).

bgouvea: I ended up choosing the same implementation as the winsdrawslosses column, even though it only looks for drawdown if it is not legacy mode, the unitary test failed.

bgouvea: Looking at the legacy mode was not enough to guarantee the existence of the column in the result, I had to create a control variable. If you think there is a more "correct" way to do this, let me know.

freqtrade/optimize/hyperopt_tools.py
```diff
@@ -267,6 +285,21 @@ def get_result_table(config: dict, results: list, total_epochs: int, highlight_b
...
         stake_currency = config['stake_currency']
+
+        if has_drawdown:
+            trials['Max Drawdown'] = trials.apply(
+                lambda x: '{} {}'.format(
+                    round_coin_value(x['max_drawdown_abs'], stake_currency),
+                    '({:,.2f}%)'.format(x['Max Drawdown'] * perc_multi).rjust(10, ' ')
+                ).rjust(25 + len(stake_currency))
+                if x['Max Drawdown'] != 0.0 else '--'.rjust(25 + len(stake_currency)),
+                axis=1
```
xmatthias, line 296: i do find the width of 25 a little wide - but i guess depending on the stake / drawdown amount, it'll be fine.

xmatthias (approved):
LGTM :+1:

PR #5108, opened 2021-06-09
Implement most pessimistic handling of trailing stoploss.

rokups:
This is a crude sketch attempting to solve https://brookmiles.github.io/freqtrade-stuff/2021/04/12/backtesting-traps/. I am opening PR so we have a spot to discuss proposed changes.

#5032 is another attempt to mitigate this. While improving situation, solution does not produce most pessimistic outcome. It may be a good thing or a bad thing. After all, we are not guaranteed that price will move in a worst possible way, so this solution (or variation of it) may be a good middle ground. We should brainstorm more.

## The idea

Trailing stoploss should produce most pessimistic result possible in backtesting. If we are expecting worst case - the only direction is to go up in live runs. Proposed changes implement these cases:
* With `trailing_only_offset_is_reached == True`, price goes up from `open` to `trailing_stop_positive_offset`, arms trailing stoploss and goes down by `trailing_stop_positive`. As a result, sell is recorded at `open + trailing_stop_positive_offset - trailing_stop_positive` price.
* With `trailing_only_offset_is_reached == False`, price goes up from `open` to `trailing_stop_positive` and goes back to `open`. As a result, sell is recorded at `open` price.

To implement this i used `open` price for stoploss calculations in backtesting. Idea is that stoploss would shift only if we go from `open` upwards, while leaving stoploss at the same level when going from `open` downwards.

Pinging people that may be interested in this discussion: @brookmiles @JoeSchr @cyberd3vil @werkkrew @rextea @j-nixx @zx @froggleston 

## The result

`Magic_Trailing_Stoploss` strategy without this patch:
```
=============== SUMMARY METRICS ===============
| Metric                | Value               |
|-----------------------+---------------------|
| Backtesting from      | 2021-05-01 00:00:00 |
| Backtesting to        | 2021-05-04 23:00:00 |
| Max open trades       | 5                   |
|                       |                     |
| Total trades          | 436                 |
| Starting balance      | 5000.000 USDT       |
| Final balance         | 6267.386 USDT       |
| Absolute profit       | 1267.386 USDT       |
| Total profit %        | 25.35%              |
| Trades per day        | 145.33              |
| Avg. stake amount     | 1000.000 USDT       |
| Total trade volume    | 436000.000 USDT     |
|                       |                     |
| Best Pair             | SUSHI/USDT 41.22%   |
| Worst Pair            | EOS/USDT 0.0%       |
| Best trade            | THETA/USDT 1.0%     |
| Worst trade           | LTC/USDT -1.2%      |
| Best day              | 541.905 USDT        |
| Worst day             | 188.638 USDT        |
| Days win/draw/lose    | 4 / 0 / 0           |
| Avg. Duration Winners | 0:02:00             |
| Avg. Duration Loser   | 0:13:00             |
| Zero Duration Trades  | 67.66% (295)        |
| Rejected Buy signals  | 2256                |
|                       |                     |
| Min balance           | 4994.841 USDT       |
| Max balance           | 6284.321 USDT       |
| Drawdown              | 6.56%               |
| Drawdown              | 65.693 USDT         |
| Drawdown high         | 765.078 USDT        |
| Drawdown low          | 699.385 USDT        |
| Drawdown Start        | 2021-05-03 18:00:00 |
| Drawdown End          | 2021-05-03 20:00:00 |
| Market change         | 1.18%               |
===============================================
```

`Magic_Trailing_Stoploss` strategy with this patch:
```
=============== SUMMARY METRICS ===============
| Metric                | Value               |
|-----------------------+---------------------|
| Backtesting from      | 2021-05-01 00:00:00 |
| Backtesting to        | 2021-05-04 23:00:00 |
| Max open trades       | 5                   |
|                       |                     |
| Total trades          | 425                 |
| Starting balance      | 5000.000 USDT       |
| Final balance         | 4993.297 USDT       |
| Absolute profit       | -6.703 USDT         |
| Total profit %        | -0.13%              |
| Trades per day        | 141.67              |
| Avg. stake amount     | 1000.000 USDT       |
| Total trade volume    | 425000.000 USDT     |
|                       |                     |
| Best Pair             | SUSHI/USDT 10.92%   |
| Worst Pair            | LUNA/USDT -14.15%   |
| Best trade            | THETA/USDT 1.0%     |
| Worst trade           | XLM/USDT -1.2%      |
| Best day              | 214.689 USDT        |
| Worst day             | -156.182 USDT       |
| Days win/draw/lose    | 1 / 0 / 3           |
| Avg. Duration Winners | 0:04:00             |
| Avg. Duration Loser   | 0:07:00             |
| Zero Duration Trades  | 56.24% (239)        |
| Rejected Buy signals  | 2256                |
|                       |                     |
| Min balance           | 4748.529 USDT       |
| Max balance           | 5031.284 USDT       |
| Drawdown              | 27.51%              |
| Drawdown              | 275.337 USDT        |
| Drawdown high         | 13.866 USDT         |
| Drawdown low          | -261.471 USDT       |
| Drawdown Start        | 2021-05-01 03:00:00 |
| Drawdown End          | 2021-05-03 13:00:00 |
| Market change         | 1.18%               |
===============================================
```

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -524,15 +524,14 @@ class IStrategy(ABC, HyperStrategyMixin):
         :param force_stoploss: Externally provided stoploss
         :return: True if trade should be sold, False otherwise
         """
-        # Set current rate to low for backtesting sell
-        current_rate = low or rate
+        current_rate = rate
         current_profit = trade.calc_profit_ratio(current_rate)
 
         trade.adjust_min_max_rates(high or current_rate)
 
         stoplossflag = self.stop_loss_reached(current_rate=current_rate, trade=trade,
                                               current_time=date, current_profit=current_profit,
-                                              force_stoploss=force_stoploss, high=high)
+                                              force_stoploss=force_stoploss, low=low, high=high)
 
         # Set current rate to high for backtesting sell
         current_rate = high or rate
@@ -599,18 +598,21 @@ class IStrategy(ABC, HyperStrategyMixin):
 
     def stop_loss_reached(self, current_rate: float, trade: Trade,
                           current_time: datetime, current_profit: float,
-                          force_stoploss: float, high: float = None) -> SellCheckTuple:
+                          force_stoploss: float, low: float = None,
+                          high: float = None) -> SellCheckTuple:
         """
         Based on current profit of the trade and configured (trailing) stoploss,
         decides to sell or not
         :param current_profit: current profit as ratio
+        :param low: Low value of this candle, only set in backtesting
+        :param high: High value of this candle, only set in backtesting
         """
         stop_loss_value = force_stoploss if force_stoploss else self.stoploss
 
         # Initiate stoploss with open_rate. Does nothing if stoploss is already set.
         trade.adjust_stop_loss(trade.open_rate, stop_loss_value, initial=True)
 
-        if self.use_custom_stoploss and trade.stop_loss < current_rate:
+        if self.use_custom_stoploss and trade.stop_loss < (low or current_rate):
             stop_loss_value = strategy_safe_wrapper(self.custom_stoploss, default_retval=None
                                                     )(pair=trade.pair, trade=trade,
                                                       current_time=current_time,
@@ -623,7 +625,7 @@ class IStrategy(ABC, HyperStrategyMixin):
             else:
                 logger.warning("CustomStoploss function did not return valid stoploss")
 
-        if self.trailing_stop and trade.stop_loss < current_rate:
+        if self.trailing_stop and trade.stop_loss < (low or current_rate):
             # trailing stoploss handling
             sl_offset = self.trailing_stop_positive_offset
 
@@ -643,7 +645,7 @@ class IStrategy(ABC, HyperStrategyMixin):
         # evaluate if the stoploss was hit if stoploss is not on exchange
         # in Dry-Run, this handles stoploss logic as well, as the logic will not be different to
         # regular stoploss handling.
-        if ((trade.stop_loss >= current_rate) and
+        if ((trade.stop_loss >= (low or current_rate)) and
                 (not self.order_types.get('stoploss_on_exchange') or self.config['dry_run'])):
 
             sell_type = SellType.STOP_LOSS
@@ -652,7 +654,7 @@ class IStrategy(ABC, HyperStrategyMixin):
             if trade.initial_stop_loss != trade.stop_loss:
                 sell_type = SellType.TRAILING_STOP_LOSS
                 logger.debug(
-                    f"{trade.pair} - HIT STOP: current price at {current_rate:.6f}, "
+                    f"{trade.pair} - HIT STOP: current price at {(low or current_rate):.6f}, "
                     f"stoploss is {trade.stop_loss:.6f}, "
                     f"initial stoploss was at {trade.initial_stop_loss:.6f}, "
                     f"trade opened at {trade.open_rate:.6f}")
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -610,7 +610,7 @@ class IStrategy(ABC, HyperStrategyMixin):
         # Initiate stoploss with open_rate. Does nothing if stoploss is already set.
         trade.adjust_stop_loss(trade.open_rate, stop_loss_value, initial=True)
 
-        if self.use_custom_stoploss:
+        if self.use_custom_stoploss and trade.stop_loss < current_rate:
             stop_loss_value = strategy_safe_wrapper(self.custom_stoploss, default_retval=None
                                                     )(pair=trade.pair, trade=trade,
                                                       current_time=current_time,
@@ -623,7 +623,7 @@ class IStrategy(ABC, HyperStrategyMixin):
             else:
                 logger.warning("CustomStoploss function did not return valid stoploss")
 
-        if self.trailing_stop:
+        if self.trailing_stop and trade.stop_loss < current_rate:
             # trailing stoploss handling
             sl_offset = self.trailing_stop_positive_offset
 
--- a/tests/optimize/test_backtest_detail.py
+++ b/tests/optimize/test_backtest_detail.py
@@ -457,6 +457,50 @@ tc28 = BTContainer(data=[
     trades=[BTrade(sell_reason=SellType.TRAILING_STOP_LOSS, open_tick=1, close_tick=3)]
 )
 
+# Test 29: trailing_stop should be triggered by low of next candle, without adjusting stoploss using
+# high of stoploss candle.
+# stop-loss: 10%, ROI: 10% (should not apply)
+tc29 = BTContainer(data=[
+    # D   O     H     L     C    V    B  S
+    [0, 5000, 5050, 4950, 5000, 6172, 1, 0],
+    [1, 5000, 5050, 5000, 4900, 6172, 0, 0],    # enter trade (signal on last candle)
+    [2, 4900, 5250, 4500, 5100, 6172, 0, 0],    # Triggers trailing-stoploss
+    [3, 5100, 5100, 4650, 4750, 6172, 0, 0],
+    [4, 4750, 4950, 4350, 4750, 6172, 0, 0]],
+    stop_loss=-0.10, roi={"0": 0.10}, profit_perc=-0.02, trailing_stop=True,
+    trailing_stop_positive=0.03,
+    trades=[BTrade(sell_reason=SellType.TRAILING_STOP_LOSS, open_tick=1, close_tick=2)]
+)
+
+# Test 30: trailing_stop should be triggered immediately on trade open candle.
+# stop-loss: 10%, ROI: 10% (should not apply)
+tc30 = BTContainer(data=[
+    # D   O     H     L     C    V    B  S
+    [0, 5000, 5050, 4950, 5000, 6172, 1, 0],
+    [1, 5000, 5500, 5000, 4900, 6172, 0, 0],    # enter trade (signal on last candle) and stop
+    [2, 4900, 5250, 4500, 5100, 6172, 0, 0],
+    [3, 5100, 5100, 4650, 4750, 6172, 0, 0],
+    [4, 4750, 4950, 4350, 4750, 6172, 0, 0]],
+    stop_loss=-0.10, roi={"0": 0.10}, profit_perc=-0.01, trailing_stop=True,
+    trailing_stop_positive=0.01,
+    trades=[BTrade(sell_reason=SellType.TRAILING_STOP_LOSS, open_tick=1, close_tick=1)]
+)
+
+# Test 31: trailing_stop should be triggered immediately on trade open candle.
+# stop-loss: 10%, ROI: 10% (should not apply)
+tc31 = BTContainer(data=[
+    # D   O     H     L     C    V    B  S
+    [0, 5000, 5050, 4950, 5000, 6172, 1, 0],
+    [1, 5000, 5500, 5000, 4900, 6172, 0, 0],    # enter trade (signal on last candle) and stop
+    [2, 4900, 5250, 4500, 5100, 6172, 0, 0],
+    [3, 5100, 5100, 4650, 4750, 6172, 0, 0],
+    [4, 4750, 4950, 4350, 4750, 6172, 0, 0]],
+    stop_loss=-0.10, roi={"0": 0.10}, profit_perc=0.01, trailing_stop=True,
+    trailing_only_offset_is_reached=True, trailing_stop_positive_offset=0.02,
+    trailing_stop_positive=0.01,
+    trades=[BTrade(sell_reason=SellType.TRAILING_STOP_LOSS, open_tick=1, close_tick=1)]
+)
+
 TESTS = [
     tc0,
     tc1,
@@ -487,6 +531,9 @@ TESTS = [
     tc26,
     tc27,
     tc28,
+    tc29,
+    tc30,
+    tc31,
 ]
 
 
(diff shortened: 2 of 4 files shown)


TheJoeSchr:
Does this also effect too tight `custom_stoploss` since they both fall under the trailing stoploss label?

rokups:
Too tight stoploss should be working properly without any patches. If you set it too tight and wick of buy candle touches it - sell will happen at stoploss price. There is no way to get artificially good results with too tight simple stoploss.

freqtrade/strategy/interface.py
```diff
@@ -637,35 +638,46 @@ def stop_loss_reached(self, current_rate: float, trade: Trade,
...
             else:
                 logger.warning("CustomStoploss function did not return valid stoploss")
 
         if self.trailing_stop:
             # trailing stoploss handling
             sl_offset = self.trailing_stop_positive_offset
-
-            # Make sure current_profit is calculated using high for backtesting.
-            high_profit = current_profit if not high else trade.calc_profit_ratio(high)
+            high_profit = trade.calc_profit_ratio(high or current_rate)
```
xmatthias: the "prior" implementation (for this line!) was better - as it didn't recalculate profit if we're not in backtesting mode (and this doesn't really change how it works, anyway).

freqtrade/strategy/interface.py
```diff
@@ -637,35 +638,46 @@ def stop_loss_reached(self, current_rate: float, trade: Trade,
...
-            if not (self.trailing_only_offset_is_reached and high_profit < sl_offset):
+            if not self.trailing_only_offset_is_reached or high_profit > sl_offset:
                 # Specific handling for trailing_stop_positive
                 if self.trailing_stop_positive is not None and high_profit > sl_offset:
                     stop_loss_value = self.trailing_stop_positive
+                    if high is not None:
+                        # In backtesting only:
+                        if self.trailing_only_offset_is_reached:
+                            # Worst case: price reaches stop_positive_offset and dives down.
+                            stop_rate = stop_rate * (1 + abs(self.trailing_stop_positive_offset))
```
xmatthias: i'm not sure i agree with this logic...

You're assuming open + positive-offset - but i don't think it's correct if you don't check if we're in a "trailing mode" already.
so you might have (with a trailing_offset of 2%, trailing with 1%):

* first candle - 1% profit (not trailing yet)
* second candle - 0.9% profit (not trailing)
* third candle - 0.5% profit (we're now trailing ... but it didn't dip down below the stoploss yet - so the trade stays open)


* 4th candle - 0.5% profit 

alternative 4th candle (case 1):
5% profit - here, you start "mid-candle" ... which might be Okish

another alternative (case 3):
-5% "full red" candle - you're now assuming "high" as your sell-price (it's limited to high, otherwise it would be higher)

In the first case - you have a stoploss > high (open + 2%)... so you sell immediately, even though the candle was "not long enough" to even cause a stoploss

the 2nd case is probably okish

In the last case - it'll be a lot higher than how it was with the previous implementation.
I think, this is fixing one particular usecase (where you have a lot of tight stoplosses) - but it's not generic enough to properly work with "wider" trailing stoploss settings ... where it can cause quite some stirr ...

xmatthias:
One idea / proposal ...
have a look at `test_backtest_detail.py` - and try to come up with a few cases that we'd need to cover - starting from there, it should be possible to see how the code really should be to cover all the necessary cases (currently, i think it only covers the "triggers when it starts to trail" case - but not one where you've been "trailing" since a while.

rokups:
I think i went to the weeds initially. @xmatthias please take a look at latest commit. It is so trivial i am amazed myself. Let me explain.

> Low happens before high for stoploss, protecting capital first

Previously freqtrade actually did not honor this. Open of new candle first adjusted stoploss based on high price, then checked if low hit stoploss. This essentially means backtest assumes price went to high first, and then went to low after. So my solution is to simply not adjust stoploss (and this not drag it up to get a better sell rate) if current_rate is already below stoploss. In such case, when stoploss is not adjusted, we are sure low of this candle is hitting stoploss already, thus selling at worst case rate.

As a bonus, test suite seems to agree. I left old commit just in case, will nuke it if this seems an appropriate solution.

> ./freqtrade.py backtesting --config config.json --config config-pairs-volume-30-binance.json --timerange 20210501-20210601 --strategy-list Magic_Trailing_Stoploss --timeframe 1h

Without patch:
```
=============== SUMMARY METRICS ===============
| Metric                | Value               |
|-----------------------+---------------------|
| Backtesting from      | 2021-05-01 00:00:00 |
| Backtesting to        | 2021-05-31 23:00:00 |
| Max open trades       | 5                   |
|                       |                     |
| Total trades          | 3624                |
| Starting balance      | 5000.000 USDT       |
| Final balance         | 23622.837 USDT      |
| Absolute profit       | 18622.837 USDT      |
| Total profit %        | 372.46%             |
| Trades per day        | 120.8               |
| Avg. stake amount     | 1000.000 USDT       |
| Total trade volume    | 3624000.000 USDT    |
|                       |                     |
| Best Pair             | SUSHI/USDT 433.27%  |
| Worst Pair            | EOS/USDT 0.0%       |
| Best trade            | THETA/USDT 1.0%     |
| Worst trade           | LTC/USDT -1.2%      |
| Best day              | 1024.659 USDT       |
| Worst day             | 188.638 USDT        |
| Days win/draw/lose    | 31 / 0 / 0          |
| Avg. Duration Winners | 0:01:00             |
| Avg. Duration Loser   | 0:06:00             |
| Zero Duration Trades  | 48.29% (1750)       |
| Rejected Buy signals  | 17808               |
|                       |                     |
| Min balance           | 4994.841 USDT       |
| Max balance           | 23622.837 USDT      |
| Drawdown              | 10.07%              |
| Drawdown              | 100.821 USDT        |
| Drawdown high         | 18579.074 USDT      |
| Drawdown low          | 18478.253 USDT      |
| Drawdown Start        | 2021-05-31 16:00:00 |
| Drawdown End          | 2021-05-31 18:00:00 |
| Market change         | -20.12%             |
===============================================
```

With patch:
```
=============== SUMMARY METRICS ===============
| Metric                | Value               |
|-----------------------+---------------------|
| Backtesting from      | 2021-05-01 00:00:00 |
| Backtesting to        | 2021-05-31 23:00:00 |
| Max open trades       | 5                   |
|                       |                     |
| Total trades          | 1380                |
| Starting balance      | 5000.000 USDT       |
| Final balance         | 1000.323 USDT       |
| Absolute profit       | -3999.677 USDT      |
| Total profit %        | -79.99%             |
| Trades per day        | 46.0                |
| Avg. stake amount     | 1000.000 USDT       |
| Total trade volume    | 1380000.000 USDT    |
|                       |                     |
| Best Pair             | LINK/USDT 0.0%      |
| Worst Pair            | XLM/USDT -105.1%    |
| Best trade            | THETA/USDT 1.0%     |
| Worst trade           | THETA/USDT -1.2%    |
| Best day              | 45.442 USDT         |
| Worst day             | -652.956 USDT       |
| Days win/draw/lose    | 1 / 0 / 12          |
| Avg. Duration Winners | 0:03:00             |
| Avg. Duration Loser   | 0:04:00             |
| Zero Duration Trades  | 22.75% (314)        |
| Rejected Buy signals  | 6954                |
|                       |                     |
| Min balance           | 1000.323 USDT       |
| Max balance           | 5045.442 USDT       |
| Drawdown              | 404.11%             |
| Drawdown              | 4045.118 USDT       |
| Drawdown high         | 45.442 USDT         |
| Drawdown low          | -3999.677 USDT      |
| Drawdown Start        | 2021-05-01 23:00:00 |
| Drawdown End          | 2021-05-13 10:00:00 |
| Market change         | -20.12%             |
===============================================
```

xmatthias:
This almost seems to be a ... too simple solution :laughing: 

The "agreeing" test-suite is a concern on it's own - as it shows that this is a case that wasn't covered - and it rather clearly should be - so i think we'll want to have a case that covers this (fails without this, works with this) - not necessarily to confirm our logic, but to be VERY sure we're not breaking this accidentally in the future (when we don't remember about this).

rokups:
While making tests i realized that this simple solution covers only half of the puzzle, which was sells that do not happen on the same candle. Surprisingly that simple change accounts for the largest portion of change in backtesting results. I also added special case handling of trailing stops on buy candle.

freqtrade/strategy/interface.py
```diff
@@ -610,7 +610,7 @@ def stop_loss_reached(self, current_rate: float, trade: Trade,
         # Initiate stoploss with open_rate. Does nothing if stoploss is already set.
         trade.adjust_stop_loss(trade.open_rate, stop_loss_value, initial=True)
 
-        if self.use_custom_stoploss:
+        if self.use_custom_stoploss and trade.stop_loss < current_rate:
```
xmatthias, line 613: i don't think this approach works (or the sign is wrong?)

during backtesting - `stop_loss_reached` (this function) is called with `low` as `current_rate` (L528).

as such - this code is preventing the adjustment of the trailing stop whenever the existing stoploss is below low (so below the current candle).
in these cases, stoploss would however not trigger anyway (that would result in a sell below the candle) - so you're preventing raising the stop - but in the wrong case.

While you do see heavily reduced profits in backtesting - the results are not correct, as the stop is simply never moved up UNLESS the prior stop already happens to be within this candle - which somehow seems to be the inverse of what you'd like to achive (reversing the sign will however NOT work)...

freqtrade/strategy/interface.py
```diff
@@ -599,18 +598,21 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
 
     def stop_loss_reached(self, current_rate: float, trade: Trade,
                           current_time: datetime, current_profit: float,
-                          force_stoploss: float, high: float = None) -> SellCheckTuple:
+                          force_stoploss: float, low: float = None,
+                          high: float = None) -> SellCheckTuple:
```
xmatthias, line 602: While moving low into this method seems to complicate things - i think it makes the intend clearer - as it's clearly visible where we're using which rate.

This also slightly changes the input to "custom_stoploss" by using "open" in backtesting instead of low as current-rate ...but that's kinda intentional...

tests/optimize/test_backtest_detail.py
```diff
@@ -457,6 +457,50 @@
...
 )
 
+# Test 29: trailing_stop should be triggered by low of next candle, without adjusting stoploss using
+# high of stoploss candle.
+# stop-loss: 10%, ROI: 10% (should not apply)
+tc29 = BTContainer(data=[
+    # D   O     H     L     C    V    B  S
+    [0, 5000, 5050, 4950, 5000, 6172, 1, 0],
+    [1, 5000, 5050, 5000, 4900, 6172, 0, 0],    # enter trade (signal on last candle)
+    [2, 4900, 5250, 4500, 5100, 6172, 0, 0],    # Triggers trailing-stoploss
```
xmatthias, line 467: I'm ... not sure this test does what the description says.

candle 2 (this line) raises the stop - and immediately triggers it (you can easily test this by changing High of this candle).

Based on the description, i'd expect this candle to adjust (but not trigger - so low should be closer to high) - and the consecutive candle should then trigger stop - without adjusting (which implies high should be higher than this high).

rokups: Maybe i worded description wrong. It looks to me test does what i intended it to do. Here is what happens:

Candle 1: enter trade, adjust stoploss using candle high, stoploss is at 4898.5 price point
Candle 2: stoploss is above candle low (4898.5 > 4500), do not adjust trailing stop, trigger stoploss at 4898.5

devilsinthedetail:
Thanks for all the work on this. I just wonder if this should be optional? If you have an optimistic outlook on the market you hyperopt on the most optimistic version, resulting in strategies which are optimized for that. While if you have a pessimistic outlook you would run hyperopt on a pessimistic version. 

I'm just worried that some hyperopt loss functions might fall into a local (pessimistic and potentially overfit) minima with this code, while the global minima (and potentially lowest overfit) is the one found with the traditional code (minus heavy slippage for tight trailing stop losses)

Edit: Basically what I'm saying is the two backtests you posted at the very top show a very clear advantage of your patch. But you have no way of verifying if this is the case during live trading. 

Edit2: Sorry had to think about it a bit more. I'm trying to verify if this patch will lead to a simple linear transform of the loss surface during hyperopt (it might be the case for ProfitOnly but not necessarily for other loss functions). I suspect it will lead to a non-linear one where the whole parameter space will be changed, leading to different hyperopt results and potentially worse strategies in live.

rokups:
Thanks for taking a look! I do not believe this should be optional because this patch does not deal with a macro level. No matter how fierce bull run is - you still only rarely get candles that go straight up (no wicks). Under normal circumstances price is fluctuating back and forth all over the place. Opting out of worst case scenario results would be giving yourself a false sense of security. In the end my reasoning is this: if you get acceptable results using very pessimistic backtesting, you are in for a nice surprise during live runs, where it will sometimes perform better than expected. Alternative would be good backtest results tricking you into taking a risk and finding strategy does not perform as good in dry/live. Not something i would recommend. This change hardly affects high quality strategies as they are either not affected by optimistic at all or are affected to very small degree. In the end, if backtest results of patched version vs non-patched version have an effect on your judgment of strategy reliability - i would advice caution as that strategy has problems.

xmatthias:
I think you misjudge the application of this to hyperopt.
Sure, the results might (potentially) be worse than before (for the same parameters) - however freqtrade has always tried to show a "pessimistic" view - which is not always easy so improvements are potentially always possible (potentially also after this patch).
If your results are drastically impacted i'd agree with @rokups, most likely your strategy is exploiting the trailing stoploss backtesting assumption - and i'd have my doubts if it'll perform as well in dry/live as in backtesting.

---

it'll definitely not be optional - otherwise we'd end up with: 
* non-comparable results depending on this parameter 
* convused users - as it'll be very difficult to understand and explain which method to pick

tests/optimize/test_backtest_detail.py
xmatthias: I debugged and realized now what i saw.
i've changed "high" to significantly higher (5550 instead of 5250) - which caused ROI to trigger.

Basically - i ran into the following Backtesting assumption:

> ROI applies before trailing-stop, ensuring profits are "top-capped" at ROI if both ROI and trailing stop applies

which is fine as is in my opinion, but has has very little to do with this change - except for me not realizing the impact of my value :laughing:

xmatthias (approved):
LGTM :+1: 

Great work on spotting this!

PR #5122, opened 2021-06-11
Ignore broken symlinks while resolving strategies.

bzed:
## Summary
Ignore broken symlinks while resolving strategies.

Without this fix the resolver tries to read from the broken symlink,
resulting in an exception that leads to the the rather confusing
error message

freqtrade.resolvers.iresolver - WARNING - Path "...../user_data/strategies" does not exist.

as a result of a symlink matching .py not being readable.

## Quick changelog

- Ignore broken symlinks while resolving strategies.

### DIFF
--- a/freqtrade/resolvers/iresolver.py
+++ b/freqtrade/resolvers/iresolver.py
@@ -91,6 +91,9 @@ class IResolver:
             if not str(entry).endswith('.py'):
                 logger.debug('Ignoring %s', entry)
                 continue
+            if entry.is_symlink() and not entry.is_file():
+                logger.debug('Ignoring broken symlink %s', entry)
+                continue
             module_path = entry.resolve()
 
             obj = next(cls._get_valid_object(module_path, object_name), None)


xmatthias:
Are you able to provide an exception log for the case you're experiencing this? maybe a way to recreate this "easily" - like - how do you have to link it - on which platform ...

i've just tried with a broken (soft) symlink within the strategy directory - but it does seem to work fine for me (unless i provide a strategy-name that doesn't exist, which fails with "impossible to load Strategy" - which i would see as correct.

bzed:
> i've just tried with a broken (soft) symlink within the strategy directory - but it does seem to work fine for me (unless i provide a strategy-name that doesn't exist, which fails with "impossible to load Strategy" - which i would see as correct.

Thats only half of the error message:
```
2021-06-12 21:44:57,111 - freqtrade.resolvers.iresolver - WARNING - Path "/home/bzed/workspace/coins/freqtrade/user_data/strategies" does not exist.
2021-06-12 21:44:57,111 - freqtrade - ERROR - Impossible to load Strategy 'foo'. This class does not exist or contains Python code errors.
```

So the first line says that the strategies folder does not exist. Which would obviously result in the second error.
At least I've spent some time to figure out why the strategies folder would not access or have permission issues, while the real error was that the strategy that was linked into it was not where it was supposed to be.

The exception is handled in the part that prints the "Path ... does not exist" warning.

As the user_data directory is mounted into the the ft directory when using docker, this will most likely add some extra confusion. I prefer to have exact error messages.

freqtrade/resolvers/iresolver.py
```diff
@@ -91,6 +92,9 @@ def _search_object(cls, directory: Path, *, object_name: str, add_source: bool =
             if not str(entry).endswith('.py'):
                 logger.debug('Ignoring %s', entry)
                 continue
+            if entry.is_symlink() and not Path.joinpath(directory, os.readlink(entry)).exists():
```
xmatthias: After being able to recreate this now - I think this could/should be simplified to 
`if entry.is_symlink() and not entry.is_file():`

That would avoid using the os module for file operations (which is kinda fluky on different environments / OS's).
[source](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file)

xmatthias (commented):
I think you misunderstood me - i'm all for having clear error-messages (that was never the point)

but i also like (or need) to be able to see the problem before merging a fix, to be sure it is fixing what it claims - and is not causing any side-effects.

freqtrade/resolvers/iresolver.py
bzed: Yes, that works. I have to admit I did not expect that is_file is True if it is a symlink....

xmatthias (approved):
LGTM :+1:

PR #5152, opened 2021-06-18
Fix errors during ubuntu install

robcaulk:
Encountering the python header error on a fresh ubuntu install:

```  
utils_find_1st/find_1st.cpp:3:10: fatal error: Python.h: No such file or directory
   #include "Python.h"
            ^~~~~~~~~~
  compilation terminated.
```

solved by installing python3.7-dev. Also need to ensure python3.7-venv for fresh install.

## Summary

Avoid python header missing error during debian install.

Solve the issue: #___

## Quick changelog

- <change log #1>
- <change log #2>

## What's new?

Avoid errors during install.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -60,7 +60,7 @@ OS Specific steps are listed first, the [Common](#common) section below is neces
     sudo apt-get update
 
     # install packages
-    sudo apt install -y python3-pip python3.7-venv python3.7-dev python3-pandas git
+    sudo apt install -y python3-pip python3-venv python3-dev python3-pandas git
     ```
 
 === "RaspberryPi/Raspbian"
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -60,7 +60,7 @@ OS Specific steps are listed first, the [Common](#common) section below is neces
     sudo apt-get update
 
     # install packages
-    sudo apt install -y python3-pip python3-venv python3-pandas git
+    sudo apt install -y python3-pip python3.7-venv python3.7-dev python3-pandas git
     ```
 
 === "RaspberryPi/Raspbian"


docs/installation.md
```diff
@@ -60,7 +60,7 @@ OS Specific steps are listed first, the [Common](#common) section below is neces
     sudo apt-get update
 
     # install packages
-    sudo apt install -y python3-pip python3-venv python3-pandas git
+    sudo apt install -y python3-pip python3.7-venv python3.7-dev python3-pandas git
```
xmatthias: Please don't "hardcode" the version of python here.

Depending on the version of ubuntu (this is also valid for Debian), you might have python 3.8 or 3.9 installed (ubuntu 20.04 uas python 3.8 as default {source](https://wiki.ubuntu.com/FocalFossa/ReleaseNotes#Python3_by_default)).

As such, hardcoding the python-version here is absolutely wrong, as it can result in either forcing a installation of python3.7 on a system that defaults to python 3.8 (which can cause problems eventually) - or even result in a "version not available" error (not sure if it's easily possible to install python3.7 on a ubuntu 20.04 system - but there's no reason in doing so).

If it can work with `python3-dev` (i'm not a ubuntu user, so i can't really test) - otherwise this is the wrong approach to me.

xmatthias (changes requested):

robcaulk:
Ok. On ubuntu 18.04 I cannot use python3-dev and python3-venv. As far as i
can tell freq requires 3.7 on all distributions so i don’t see why we
wouldn’t require it here. These commands simply add 3.7 to the available
python versions on the system, these commands dont replace the existing
install. But if it’s bad practice to require 3.7 here then I’m not sure
what the preferable method should be. At the very least we should add a
note to help people make decisions during this install?

And finally, python3-dev certainly needs to be added.

Let me know what you prefer.

-rob

xmatthias:
Freqtrade _supports_  python 3.7 on all systems - the requirement is 3.7 _or higher_.
As such - i'll not require 3.7 - but i'd encourage everyone to use 3.8 at least (there's internal speed improvements that will be beneficial).

also, Ubuntu 20.04 is the latest LTS release - as such - i consider 18.04 as no longer relevant (even though i'm pretty sure Ubuntu still supports it - i'm sure they don't recommend it for any new install - and neither would i).

The install script / install instruction should not mess with the existing users python installation. 
Using 3.7 might mess up a 20.04 system - so you fixed it for your particular usecase, but broke it for most others.

robcaulk:
Alright so a note it is? At the very least we need the python3-dev in
there. And then we just hope <20.04 users can sort the errors out on their
own? Is there a better solution?

Re: relevancy/support of ubuntu 18.04. It has 10 years total support ending
in 2028 (extended from 5 for typical LTS) and as such is a widely used
distribution in many fields requiring a stable distro with predictable
packages. I don’t imagine ubuntu is making any recommendations either way
for a “new install.”
But we can bet that they wouldn’t extend support to 10 years without a
large active user base, some of which being freq users :-).

“ Mark Shuttleworth explained that the increase in the support period is
due to a fairly long cycle of product use in the financial industries. and
telecommunications, as well as a fairly large life cycle of embedded
devices and IoT.”

https://ubunlog.com/en/Ubuntu-18-04-support-time-will-be-increased-to-10-years/

xmatthias:
Ubuntu 18.04 has end of life to 10 years (which basically means security updates only) - but standard support still ends 2023 ([source](https://wiki.ubuntu.com/Releases)

That's however not the topic of this issue.

---

I'm sure there's also people still using 16.04 or 14.04 - or arch or centos ....the provided instruction won't work "out of the box" for these either (maintaining install instructions for all systems is too huge a task to keep them correct).

As such - i will assume the Latest (stable/LTS) release (which is 20.04) and have the documentation work for that.

`python3.7-dev` is **not** available on 20.04 (and python3-dev defaults to 3.8) - and as such, your proposed change might be correct for your system - but wrong for the documentation - and for anyone using 20.04.

If this bothers you that much - feel free to put a note below that install instructions may slightly differ for older systems - but the default has to work for the latest version.

robcaulk:
I suggested we add python3-dev i.e. the current install directions won’t
even work for 20.04 as far as I can tell unless python3-dev is installed by
default on 20.04 but I can’t find any indication online that it is the
case. Unless I’m missing something?

xmatthias:
i never argued against adding `python3-dev` :smile: only about `python3.7-dev`...

robcaulk:
Yes I agreed to this a while back. At this moment it is not immediately
clear to me how you would prefer to help 18.04 users install freq.

eatrisno:
I suggest using docker to test and run your app

xmatthias (approved):
LGTM 👍

PR #5157, opened 2021-06-20
Margin db

samgermain:
# Margin-db Pull Request Review

## Short Summary

Adding margin trading columns and calculations to the database. Updates `freqtrade/persistence` to account for leverage and shorting

Solves part of the issue: #2183 

### Binance interest formula

    I (interest) = P (borrowed money) * R (daily_interest/24) * ceiling(T) (in hours)

[source](https://www.binance.com/en/support/faq/360030157812)

### Kraken interest formula

    Opening fee  = P (borrowed money) * R (quat_hourly_interest)
    Rollover fee = P (borrowed money) * R (quat_hourly_interest) * ceiling(T/4) (in hours)
    I (interest) = Opening fee + Rollover fee

[source](https://support.kraken.com/hc/en-us/articles/206161568-What-are-the-fees-for-margin-trading-)

### Profit ratio for short trades

`profit_ratio = (1 - (close_trade_value/self.open_trade_value))`

### Profit ratio for leveraged trades

`leveraged_profit_ratio = profit_ratio * leverage`

## Tests

To add shorting and leverage functionality to freqtrade, the majority of changes involved editing existing methods within `freqtrade/persistence/models.py`, accounting for additional factors created by _leverage_ and _short trading_

#### Short trading

- Exit trade amounts are slightly higher than their enter amounts, due to the extra amount purchased for paying back interest owed for the amount of currency borrowed
- Profit calculation were calculated inversely compared to longs
- Stoplosses moved in the reverse direction
- The entering/starting trade was a "sell" and the exiting/closing trade was a "buy"

#### Leveraged long trading

- Leveraged long trades had the amount of interest owed subtracted from the value of their exiting trade

### Original tests

These methods were added to `LocalTrade` in `freqtrade/persistence/models.py`

- `set_stop_loss`
- `set_liquidation_price`
- `calculate_interest`
- `borrowed(calculated property)`

These were the tests created to test these methods

#### *test_set_stop_loss_liquidation_price*

Tested methods from `freqtrade/persistence/models.py`

- `LocalTrade.set_stop_loss`
- `LocalTrade.set_liquidation_price`

Tested to check for these conditions

- `LocalTrade.stop_loss` is never lower than `LocalTrade.liquidation_price` for longs
- `LocalTrade.stop_loss` is never higher than `LocalTrade.liquidation_price` for shorts
- `LocalTrade.stop_loss` and `LocalTrade.intial_stop_loss` are assigned to the value of `LocalTrade.liquidation_price` when no `LocalTrade.stop_loss` is not assigned
- `LocalTrade.liquidation_price` is not changed when `LocalTrade.stop_loss` gets assigned

#### *LocalTrade.calculate_interest*

Conditions tested

- correct interest calculated for a short trade made on kraken
- correct interest calculated for a long trade made on kraken
- correct interest calculated for a short trade made on binance
- correct interest calculated for a long trade made on binance

#### *LocalTrade.borrowed*

Conditions tested for

- borrowed was equal to `amount` for short trades
- borrowed was equal to `amount * leverage-1` for long trades (1 is subtracted from leverage because the collateral is in the borrowed currency and is already owned)


docs/strategy-advanced.md
```diff
@@ -49,7 +49,7 @@ from freqtrade.exchange import timeframe_to_prev_date
 
 class AwesomeStrategy(IStrategy):
     def confirm_trade_exit(self, pair: str, trade: 'Trade', order_type: str, amount: float,
-                           rate: float, time_in_force: str, sell_reason: str,
+                           rate: float, time_in_force: str, close_reason: str,
```
xmatthias: This is a ... not so good change for now.

It's a user-facing interface - so by changing this we're breaking everyone's strategy that's using this.

I don't mind changing this _once it's necessary_ - but for now, i'd neglect this as it's a "theoretical" / behind the curtain change / PR, preparing things for short support - while not fully implementing it - so bothering users by breaking their stuff without any benefit / new feature will result in nothing but bad update / user experience and frustration.

This also applies to the database fields - renaming fields might break users custom analysis scripts - which is not necessary at this point. Neglecting this will also keep the diff quite small ... while now it's 16 files changed - where the "real" changes are in one file only.

freqtrade/persistence/models.py
```diff
@@ -604,7 +657,7 @@ def get_trades_proxy(*, pair: str = None, is_open: bool = None,
             sel_trades = [trade for trade in sel_trades if trade.close_date
                           and trade.close_date > close_date]
 
-        return sel_trades
+        return sel_trades   #TODO: What is sel_trades does it mean sell_trades? If so, update this for margin
```
xmatthias: selected trades... it's just a dummy filter method for backtesting...

freqtrade/persistence/models.py
```diff
@@ -518,16 +560,20 @@ def calc_close_trade_value(self, rate: Optional[float] = None,
         if rate is None and not self.close_rate:
             return 0.0
 
-        sell_trade = Decimal(self.amount) * Decimal(rate or self.close_rate)  # type: ignore
-        fees = sell_trade * Decimal(fee or self.fee_close)
-        return float(sell_trade - fees)
+        close_trade = Decimal(self.amount) * Decimal(rate or self.close_rate)  # type: ignore
+        fees = close_trade * Decimal(fee or self.fee_close)
+        interest = ((self.interest_rate * Decimal(borrowed or self.borrowed)) * (datetime.utcnow() - self.open_date).days) or 0 #Interest/day * num of days
```
xmatthias: i'd be careful with this - your assumption is that interrest is paid per day - but AFAIK, some exchanges calculate interrest per hour - but i guess we can leave that for later - when we get to the implementation of the different exchanges (basically - deferr that to when we need it).

freqtrade/persistence/models.py
```diff
@@ -254,11 +257,20 @@ class LocalTrade():
...
     strategy: str = ''
     timeframe: Optional[int] = None
 
+    #Margin trading properties
+    leverage: Optional[float] = None
+    borrowed: float = 0
+    borrowed_currency: float = None
+    interest_rate: float = 0
+    min_stoploss: float = None
+    isShort: boolean = False
```
xmatthias: If you add these here (and intend to use them) - you MUST add them as columns to the Trade object too - otherwise they won't be stored to database.

freqtrade/persistence/models.py
```diff
@@ -254,11 +257,20 @@ class LocalTrade():
...
+    close_order_status: str = '' 
     strategy: str = ''
     timeframe: Optional[int] = None
 
+    #Margin trading properties
+    leverage: Optional[float] = None
+    borrowed: float = 0
+    borrowed_currency: float = None
+    interest_rate: float = 0
+    min_stoploss: float = None
```
xmatthias: Maybe a better wording would be "liquidation price" ?

freqtrade/persistence/models.py
```diff
@@ -254,11 +257,20 @@ class LocalTrade():
...
-    sell_order_status: str = ''
+    close_reason: str = ''   
+    close_order_status: str = '' 
     strategy: str = ''
     timeframe: Optional[int] = None
 
+    #Margin trading properties
+    leverage: Optional[float] = None
+    borrowed: float = 0
+    borrowed_currency: float = None
```
xmatthias: I THINK you'll need collateral currency as well (or instead of this?).

AFAIK, you can Long/Short  ETH/BTC - while using USDT as collateral - so your USDT balance will determine the liquidation price - while it's not directly involved in your trade.

freqtrade/persistence/models.py
samgermain: Are you referring to the Trade class that is also in models.py? Because they are also added there. Did I miss one?

xmatthias: You added (or pushed) it about 2 minutes after i made this comment :laughing: 

so all good now i think.

samgermain:
Can I run all those checks(Freqtrade CI / build_linux  ...) locally?

freqtrade/persistence/models.py
samgermain: I'm assuming interest is calculated as an amount/day value, not sure if that's what you meant. I understand it's dangerous though and left a #TODO there

docs/strategy-advanced.md
samgermain: I changed it back

freqtrade/persistence/models.py
samgermain: I added the field `collateral_currency` for now, and changed both these currencies to string

xmatthias:
> Can I run all those checks(Freqtrade CI / build_linux ...) locally?

You can (and should) https://www.freqtrade.io/en/stable/developer/#developer-setup

Basically, it's `pytest` - followed by `flake8`, `isort .` and `mypy freqtrade`. obviously the sequence is not important (but all 4 are ran in CI, so should pass)

freqtrade/persistence/models.py
```diff
@@ -259,7 +262,29 @@ class LocalTrade():
     strategy: str = ''
     timeframe: Optional[int] = None
 
+    # Margin trading properties
+    leverage: Optional[float] = 0.0
```
xmatthias: shouldn't the default for leverage be 1? (non-leveraged?) 

i guess it'll depend on the calculation applied ... just asking ... :laughing:

freqtrade/persistence/models.py
```diff
@@ -518,16 +576,31 @@ def calc_close_trade_value(self, rate: Optional[float] = None,
...
+        # TODO: Need to set other conditions because sometimes self.open_date is not defined, but why would it ever not be set
+        try:
+            open = self.open_date.replace(tzinfo=None)
+            now = datetime.now()
+
+            # breakpoint()
+            interest = ((Decimal(self.interest_rate or 0) * Decimal(self.borrowed or 0)) *
+                        Decimal((now - open).total_seconds())/86400) or 0  # Interest/day * (seconds in trade)/(seconds per day)
+        except:
+            interest = 0
```
xmatthias: you will for sure want to extract this into a seperate method.
First of all, to be able to easily test it - but more importantly - because it'll also be needed for the "current profit" calculation - otherwise things like "ROI" will be off by this - so you think you sold for 0.5% - but instead it's a -0.5% loss - as we didn't account for all fees upfront

tests/test_persistence.py
```diff
@@ -977,6 +993,14 @@ def test_to_json(default_conf, fee):
                       'strategy': None,
                       'timeframe': None,
                       'exchange': 'binance',
+
```
xmatthias: As this does not add the trade to a session, defaults won't apply (that's when sqlalchemy applies defaults).

to test this you can however add "leverage" to the trade instance above - and it should "get dumped out" down here ...

freqtrade/persistence/models.py
stackoverflowsam93: > AFAIK, you can Long/Short ETH/BTC - while using USDT as collateral - so your USDT balance will determine the liquidation price - while it's not directly involved in your trade.

If were going to allow those then we would have to allow freqtrade to work with multiple base currencies (because if your base is USDT and you short ETH on ETH/BTC, you'll either have to buy back extra ETH, or leave a little bit of BTC), or we'll have to immediately sell the remaining ETH or BTC, which the only use I can think of for that is if the exchange has run out of USDT to lend out for the pair your trading, so then you would need to borrow another currency like BTC if you wanted to buy ETH. That could be a nice feature, but it would be a decent amount of extra work, and I feel like the logic would be kinda different anyway.

samgermain: > AFAIK, you can Long/Short ETH/BTC - while using USDT as collateral - so your USDT balance will determine the liquidation price - while it's not directly involved in your trade.

If were going to allow those then we would have to allow freqtrade to work with multiple base currencies (because if your base is USDT and you short ETH on ETH/BTC, you'll either have to buy back extra ETH, or leave a little bit of BTC), or we'll have to immediately sell the remaining ETH or BTC, which the only use I can think of for that is if the exchange has run out of USDT to lend out for the pair your trading, so then you would need to borrow another currency like BTC if you wanted to buy ETH. That could be a nice feature, but it would be a decent amount of extra work, and I feel like the logic would be kinda different anyway.

tests/conftest.py
```diff
@@ -63,13 +57,12 @@ def log_has_re(line, logs):
 
 def get_args(args):
     return Arguments(args).get_parsed_arg()
+# Source: https://stackoverflow.com/questions/29881236/how-to-mock-asyncio-coroutines
 
 
-# Source: https://stackoverflow.com/questions/29881236/how-to-mock-asyncio-coroutines
 def get_mock_coro(return_value):
```
xmatthias: I'd ... keep this comment where it was - now it seems to belong to `get_args` - but it's really for `mock_coro` ...

tests/conftest.py
```diff
@@ -2043,9 +2025,116 @@ def saved_hyperopt_results():
...
+        'side': 'buy',
+        'symbol': 'mocked',
+        'datetime': arrow.utcnow().isoformat(),
+        'price': 0.00004173,
+        'amount': 91.99181073,
+        'filled': 91.99181073,
+        'remaining': 0.0,
+        'status': 'closed',
+        'is_short': True,
+        'leverage': 3
```
xmatthias: Are these 2 fields actually part of the Return values from ccxt? If not, this will create a false sense of "working" - while in reality we'll never get this value...

i'm asking this because i know they are not ... 
https://github.com/ccxt/ccxt/blob/39a5cd3d320fb8ea2935fa0061b668c1a0a0cc13/js/binance.js#L1983-L2005

tests/test_persistence.py
```diff
@@ -659,12 +659,13 @@ def test_migrate_new(mocker, default_conf, fee, caplog):
...
                 order_filled_date DATETIME,
                 order_update_date DATETIME,
+                leverage FLOAT,
                 PRIMARY KEY (id),
                 CONSTRAINT _order_pair_order_id UNIQUE (ft_pair, order_id),
                 FOREIGN KEY(ft_trade_id) REFERENCES trades (id)
             )
             """))
-
+        # TODO-mg @xmatthias: Had to add field leverage to this table, check that this is correct
```
xmatthias: I would think it's not (not for this test, anyway) - this test tests (try saying that quickly multiple times ..)  database migrations - as such - you'll want to leave the above query untouched - and define the migration correctly in `migrations.py` - which will cause this field to be filled with the correct default value (i assume for leverage it'll be 0  or 1 - but you'll probably also want a "isShort" field in the orders table).

freqtrade/persistence/models.py
```diff
@@ -572,7 +724,7 @@ def select_order(self, order_side: str, is_open: Optional[bool]) -> Optional[Ord
         else:
             return None
 
-    @staticmethod
+    @ staticmethod
```
xmatthias: :eyes:

freqtrade/persistence/models.py
```diff
@@ -278,6 +306,22 @@ def open_date_utc(self):
     def close_date_utc(self):
         return self.close_date.replace(tzinfo=timezone.utc)
 
+    def __init__(self, **kwargs):
+        if kwargs.get('leverage') and kwargs.get('borrowed'):
+            # TODO-mg: should I raise an error?
+            raise OperationalException('Cannot pass both borrowed and leverage to Trade')
```
xmatthias: I think this is a theoretical problem (possibly relevant for tests - but not for the rest).
Trade objects are not created by the user, but by Freqtrade code only - so if the trade creation (only happens in `freqtradebot.py` and `backtesting.py` ) is done properly, this should never be the case.

freqtrade/persistence/models.py
```diff
@@ -403,31 +459,63 @@ def adjust_stop_loss(self, current_price: float, stoploss: float,
...
+
+        if ('leverage' in order and 'borrowed' in order):
+            raise OperationalException(
+                'Pass only one of Leverage or Borrowed to the order in update trade')
+
+        if 'is_short' in order and order['side'] == 'sell':
+            # Only set's is_short on opening trades, ignores non-shorts
+            # TODO-mg: I don't like this, but it might be the only way
+            self.is_short = order['is_short']
+
```
xmatthias: I doubt this will work in practice (it might in tests  ... but see comment on the test fixture).

the order object does not contain a "is_short" property - at least not "formalized / unionized by ccxt.
Sure, we MIGHT be able to inject it in the exchange class ... but i'm not sure that'll work correctly...

freqtrade/persistence/models.py
```diff
@@ -403,31 +459,63 @@ def adjust_stop_loss(self, current_price: float, stoploss: float,
...
-        if order_type in ('market', 'limit') and order['side'] == 'buy':
+        if order_type in ('market', 'limit') and self.is_opening_trade(order['side']):
             # Update open rate and actual amount
+
             self.open_rate = float(safe_value_fallback(order, 'average', 'price'))
             self.amount = float(safe_value_fallback(order, 'filled', 'amount'))
+            if 'borrowed' in order:
+                self.borrowed = order['borrowed']
+            elif 'leverage' in order:
+                self.leverage = order['leverage']
```
xmatthias: the fields used should (must) correspond to the ccxt return object ... and i'm yet to see these fields :(

freqtrade/persistence/migrations.py
```diff
@@ -47,6 +47,14 @@ def migrate_trades_table(decl_base, inspector, engine, table_back_name: str, col
     min_rate = get_column_def(cols, 'min_rate', 'null')
     sell_reason = get_column_def(cols, 'sell_reason', 'null')
     strategy = get_column_def(cols, 'strategy', 'null')
+
+    leverage = get_column_def(cols, 'leverage', '0.0')
```
xmatthias: In order for migrations to work correctly - you'll have to enter one of the new columns around row 177 (and for the orders table in row 191).

currently the bot crashes when started with an existing database as the migrations don't run - so the new tables will be missing.

freqtrade/persistence/models.py
```diff
@@ -259,16 +263,40 @@ class LocalTrade():
...
+        if self.is_short is None or self.amount is None:
+            raise OperationalException(
+                'LocalTrade.amount and LocalTrade.is_short must be assigned before LocalTrade.leverage')
+        self.__leverage = lev
+        self.__borrowed = self.amount * (lev-1)
+        self.amount = self.amount * lev
+
+    @borrowed.setter
+    def borrowed(self, bor: float):
+        if not self.amount:
```
xmatthias: Amounts should always be assigned (it's a mandatory field) when the trade object is created (look at `freqtradebot.py` and `backtesting.py` - these are the only places we create orders).

the only case where this "might" not be the case is tests - in which case the test should be fixed (this is easy to overlook in tests sometimes ...)

xmatthias (commented):
I'm currently slightly worried about the usage of "non-existing" ccxt response fields from the order response ... so "most" comments will revolve around that probably

freqtrade/persistence/models.py
xmatthias: won't you "repay" the moment you sell your trade (or it get's liquidated) ? 

but the way you're describing it youre probably right - it's not worth the extra effort for the moment

tests/conftest.py
samgermain: Do you have an example of some code I can execute to receive a valid ccxt response object?

samgermain: I know that with Kraken the leverage is passed in like 

```
kraken.create_market_sell_order('BTC/USD', 1, {'leverage': 2})
```

I'm not sure how that would transfer over to out Order Class

freqtrade/persistence/models.py
samgermain: What should never be the case? Do you that this branch should never be visited?

freqtrade/persistence/models.py
samgermain: Yeah it's just a safety check in case someone writes some sloppy code. This isn't in there anymore though, only leverage is a computed property. This is because for shorts, you actually can't determine the leverage based on the borrowed amount, and the amount, as the full amount will be borrowed

freqtrade/persistence/migrations.py
samgermain: Could you explain what you mean a little more in depth here? Your message isn't ringing any bells for me.

freqtrade/persistence/models.py
samgermain: Does this mean that `leverage` and `borrowed` cannot be updated within `trade.update` and that they either must be passed when instantiating a Trade object

```
trade = Trade(
        id=2,
        pair='ETH/BTC',
        stake_amount=0.001,
        open_rate=0.01,
        amount=5,
        is_open=True,
        open_date=ten_minutes_ago,
        fee_open=fee.return_value,
        fee_close=fee.return_value,
        borrowed=90.99181073,
        is_short=True,
        exchange='binance'
    )
 ```
 
 Or must be updated via some other method?
 
 What do you  think would be a good solution?

File: tests/test_persistence.py (hunk 12 above)
Comment (samgermain) [reply tests/test_persistence.py]: So I thought you didn't want leverage and is_short to be in the order class.

The reason I added leverage here is because this test fails without it

File: freqtrade/persistence/migrations.py (hunk 17 above)
Comment (xmatthias) [reply freqtrade/persistence/migrations.py]: In this file - look at line 177 in this file (and 191 ... at least based on yesterday's count). 
It's pointing to the latest column that was added to the table - to detect if a migration is necessary or not.

if you don't update, freqtrade will not realize migrations need to run ... -> they will not run - and we'll end up having missing columns.

File: tests/conftest.py (hunk 11 above)
Comment (xmatthias) [reply tests/conftest.py]: the problem is - the code you're pointing at here is the INPUT to create a order.

if you run this code, you'll notice that 
a) you've gone short on 30.000+$ (so please be careful) 
b) the response - assuming the order is places successfully - will not contain leverage

at least not in the top level response (which is guaranteed by ccxt to be aligned between exchanges).
While you can reach into the info substructure - that will be different by exchange - so that should not be done from the trade class.

i'll need to think about this some - as i'm currently not clear how the (real world) responses from the exchanges differ - to determine the best course of action.

Technically - we should however keep all exchange-specifics in the exchange class(es) - otherwise it'll be scattered around the code which will be a nightmare to maintain.

so worst case - we call a dummy method which does nothing by default (from the exchange class) - and "extracts" the necessary keys (leverage, ...) from the info key and injects them into the order responses.

File: tests/test_persistence.py (hunk 12 above)
Comment (xmatthias) [reply tests/test_persistence.py]: Yes i know (the failing test is showing that migrations are not running - so by resetting this test to how it was - (maybe even adding an assert to ensure that "leverage" is the default for non-leveraged trades (0, 1 or None?) 

you can test if migrations are running correctly or not (hint: they are currently not - see other comment).

File: freqtrade/persistence/models.py (hunk 16 above)
Comment (xmatthias) [reply freqtrade/persistence/models.py]: I guess https://github.com/freqtrade/freqtrade/pull/5157#discussion_r660001578 may point to a solution to this (which would enable these keys to be used here) ... but i'm yet inconvinced about how to approach this specifically - as ideally, the alignment of "leverage" and whatever else is needed would be done by ccxt ....

File: freqtrade/persistence/models.py (hunk 14 above)
Comment (xmatthias) [reply freqtrade/persistence/models.py]: Well this is the `__init__` method (constructor). As such - freqtrade (speak - we) controll what is passed in here - so hitting this branch will be a programming error we didn't encounter (not a problem in the strategy / user error) - and as such, i (hope) it's a theoretical problem :laughing:

Comment (samgermain) [reply freqtrade/persistence/models.py]: yeah the purpose of that conditional is to catch programming errors

Comment (xmatthias) [reply freqtrade/persistence/models.py]: maybe add a comment that makes that clear ... otherwise i might have the same doubt in a few days (or months ...)

File: tests/conftest.py (hunk 11 above)
Comment (samgermain) [reply tests/conftest.py]: Would it be a mistake to just specify the `leverage` and `is_short` when creating the trade object, and remove all the logic of assigning them from within the update function?

Comment (xmatthias) [reply tests/conftest.py]: no - i think that would be correct - that was my assumption anyway - as using leverage / borrowing is always a explicit decision at the time of trade entry.

File: freqtrade/persistence/models.py (hunk 14 above)
Comment (samgermain) [reply freqtrade/persistence/models.py]: Is `OperationalException` meant for strategy / user error and is there another error or exception type which would make it clear that it's to prevent a programmer error?

Comment (xmatthias) [reply freqtrade/persistence/models.py]: OperationalException is fine - it's the most severe error we have - but it's the only one intended to "stop bot execution" - which i think is fine for a programmer error.

We've the exception hierarchy [documented](https://www.freqtrade.io/en/stable/developer/#errorhandling) as well should you need one of the others for whatever reason.

**Comment by samgermain** [comment]
In the mock trades, how is `close_profit_abs` calculated? 

For trade 3, I tried `amount * close_rate - amount * open_rate` = `(123 * 0.06) - (123 * 0.05)` = `1.23` but the amount given in that test is `0.000155` 

What is `0.000155` calculated from?

Also why is the `close_profit` just `0.06 - 0.05`? Why doesn't it take `amount` into account?

**Comment by xmatthias** [comment]
> In the mock trades, how is close_profit_abs calculated?

it's a mock trade - might not be 100% correct for all trades (the mock-trades are not used to verify calculations afaik)
now technically - it should be - but the way you ask the question makes me assume it probably isn't :)

File: freqtrade/persistence/models.py [hunk 19]
```diff
@@ -226,14 +229,15 @@ class LocalTrade():
...
+
     # open_trade_value - calculated via _calc_open_trade_value
     open_trade_value: float = 0.0
     close_rate: Optional[float] = None
     close_rate_requested: Optional[float] = None
     close_profit: Optional[float] = None
     close_profit_abs: Optional[float] = None
     stake_amount: float = 0.0
-    amount: float = 0.0
+    _amount: float = 0.0
```
xmatthias: This is a mess - do NOT do this (i've reverted it now - at least partially).
This column should always contain the amount of coin we own on the exchange.

While it may seem intelligent to use a property for it - it'll also apply to the database - which will make analysis later on more difficult.

freqtrade/persistence/models.py
```diff
@@ -700,11 +901,22 @@ class Trade(_DECL_BASE, LocalTrade):
...
     min_rate = Column(Float, nullable=True)
-    sell_reason = Column(String(100), nullable=True)
-    sell_order_status = Column(String(100), nullable=True)
+    sell_reason = Column(String(100), nullable=True)  # TODO: Change to close_reason
+    sell_order_status = Column(String(100), nullable=True)  # TODO: Change to close_order_status
     strategy = Column(String(100), nullable=True)
     timeframe = Column(Integer, nullable=True)
 
+    # Margin trading properties
+    _leverage: float = None  # * You probably want to use LocalTrade.leverage instead
```
xmatthias: we will surely store leverage into the database.

tests/conftest.py
```diff
@@ -201,26 +191,47 @@ def add_trade(trade):
...
-
     trade = mock_trade_2(fee)
     add_trade(trade)
-
     trade = mock_trade_3(fee)
     add_trade(trade)
-
     trade = mock_trade_4(fee)
     add_trade(trade)
-
```
xmatthias: These unnecessary changes make this very difficult to review - as for every line, i'll have to check "is this relevant, or just some stupid whitespace change.

Please revert all non-essential changes (whitespaces, formatting).

freqtrade/persistence/models.py
```diff
@@ -259,9 +263,79 @@ class LocalTrade():
...
+    borrowed: float = 0.0
+    leverage: float = None
+
+    # @property
+    # def base_currency(self) -> str:
+    #     if not self.pair:
+    #         raise OperationalException('LocalTrade.pair must be assigned')
+    #     return self.pair.split("/")[1]
+
+    # TODO: @samgermain: Amount should be persisted "as is".
```
xmatthias: i don't think you understood the intend with "amount".
it defines what you own on the exchange. Leverage does not really play a role in this - and as such, this part is definitely wrong.

I can buy 100 XRP  (then i own 100 XRP) - i can buy 200 XRP with 2x leverage - then i own 200 XRP - and that's what goes into amount.

"leverage" only comes into play once you define how much you paid for this.

freqtrade/persistence/models.py
```diff
@@ -505,56 +630,135 @@ def recalc_open_trade_value(self) -> None:
...
+            opening_fee = borrowed * rate
+            roll_over_fee = borrowed * rate * max(0, (hours-four)/four)
+            return opening_fee + roll_over_fee
+        elif self.exchange == 'binance_usdm_futures':
+            # ! TODO-mg: This is incorrect, I didn't look it up
+            return borrowed * (rate/twenty_four) * max(hours, one)
+        elif self.exchange == 'binance_coinm_futures':
+            # ! TODO-mg: This is incorrect, I didn't look it up
+            return borrowed * (rate/twenty_four) * max(hours, one)
+        else:
```
xmatthias: This is a no-go (adding exchange-specific stuff to the models file) - as it can break too easily.

If there's different fee models - create an enum, save the "fee model" in the database upon trade creation. That'll allow for easier reusability.

freqtrade/persistence/models.py
```diff
@@ -259,9 +263,79 @@ class LocalTrade():
...
+    #     self._leverage = None
+
+    # @property
+    # def leverage(self) -> float:
+    #     return self._leverage
+
+    # @leverage.setter
+    # def leverage(self, value):
+    #     self._leverage = value
+    #     self._borrowed = None
```
xmatthias: There's 0 reason to use a property here - that'll only complicate things with the database (the database column will be named differently ...).
Also - leverage will not change once the trade is created - you either use leverage - or you don't.

freqtrade/persistence/models.py
```diff
@@ -259,9 +263,79 @@ class LocalTrade():
...
+        return self.close_date.replace(tzinfo=timezone.utc)
+
     def __init__(self, **kwargs):
+        if kwargs.get('leverage') and kwargs.get('borrowed'):
+            # TODO-mg: should I raise an error?
+            raise OperationalException('Cannot pass both borrowed and leverage to Trade')
         for key in kwargs:
             setattr(self, key, kwargs[key])
+        if not self.is_short:
+            self.is_short = False
```
xmatthias: i think the cleaner way would be to set the default on the column.

careful - it needs to be set on the `LocalTrade` object, as well as on the `Trade` column

freqtrade/persistence/models.py
```diff
@@ -505,56 +630,135 @@ def recalc_open_trade_value(self) -> None:
         """
         self.open_trade_value = self._calc_open_trade_value()
 
+    def calculate_interest(self, interest_rate: Optional[float] = None) -> Decimal:
+        """
+        : param interest_rate: interest_charge for borrowing this coin(optional).
+        If interest_rate is not set self.interest_rate will be used
+        """
+        # TODO-mg: Need to set other conditions because sometimes self.open_date is not defined, but why would it ever not be set
```
xmatthias: This is most likely only a test problem - in real runs (outside of pytest) - open_date will always be set.

xmatthias (commented):
I did revert parts of your changes (related to properties ...) - as leverage and amount should be stored "as is" in the database.

`amount` defines what you own - which does not change with leverage (maybe you bought it with borrowed money - but you still own that amount) - so that precise amount should be stored in the database as well.

While the property might work from a technical point of view, you should also keep in mind that users will want to analyze the data later in an easy way - so how the data is stored is also non neglectible.

I've however fixed the migrations (hope this also makes it clearer HOW they work?) - that's for sure one of the more complex topics ...

freqtrade/persistence/models.py
samgermain: I was under the impression that when you send an order to Kraken, you send the amount without leverage applied, and also send the leverage. This might still be the case, but Im not sure because I made [an order with leverage](https://i.imgur.com/kCsThGe.png), and [one without leverage](https://i.imgur.com/ZEOV6TE.png) and they're giving the same values. So far my balance has only updated for one of the sales worth 0.0001, but I think they've both been completed. Kraken seems to give me errors and not explain them.

samgermain: I think that the way it works on kraken is that you specify the amount like 0.0001 btc with 3x leverage, and then the profits you make are as if you had bought 0.0003 btc.

With this logic, any time you access `amount` it gives you `_amount * leverage` I tried to abstract away the use of all the `_` properties as much as I could, so that the only things that reference them are the computed properties

 If you're trading on an exchange that doesn't work like this, then you would pass `borrowed` and `amount` to trade and when you access `amount` it just gives you `_amount`
 
 If you can think of a better way to do this I could do that, but I was trying to abstract away the implementation details of how different exchanges handle the amount you trade vs your leverage. This is also why I made only one of `leverage` and `borrowed` able to be set.
 
 We could maybe create a separate class that contains these 3 variables to abstract the logic away further?

samgermain: I actually might be wrong about how kraken determines your leverage/borrowed amount(Kraken might do it in a way where they divide your stake amount by your leverage, trying to figure it out) but it looks like it's kinda different for each exchange. I would like to keep the computed properties if we could because doing something like 

```
@property
def leverage(self):
    if (_leverage):
        return _leverage
    else:
        return exchange.getLeverage(_borrowed, _amount)
```

or even just

```
@property
def leverage(self):
        return exchange.getLeverage(_leverage, _borrowed, _amount)
```

I thought about just assigning these properties in the constructor, but then if we have anything where we are changing the `amount` `leverage` or `borrowed`, then we won't be able to calculate the new numbers because the original numbers won't make sense anymore.  I also decided to do it this way when I was still passing `leverage` and `borrowed` through the `update` function

It's set up like this so that the `_` properties are always the original values the properties without `_` are as we understand them

freqtrade/persistence/models.py
samgermain: An enum? Why not just a function, kind of confused about this implementation

freqtrade/persistence/models.py
samgermain: Im not exactly sure what the ccxt return object is but here's where leverage is specified for binance and kraken
https://i.imgur.com/CrNjQEi.png

freqtrade/persistence/models.py
xmatthias: Currently - you check for `self.exchange == 'whatever'` - instead of doing this - check for `self.interrest_mode = InterrestModes.Daily` (or whatever) ... 

that way we can define in the exchange class the calculation method to use - without having to have a seperate case for each exchange.
adding support for new exchanges should NOT require changes to the database model (obviously assuming the calculation method already existed).

freqtrade/persistence/models.py
```diff
@@ -587,7 +587,7 @@ def calculate_interest(self, interest_rate: Optional[float] = None) -> Decimal:
             return zero
 
         open_date = self.open_date.replace(tzinfo=None)
-        now = datetime.utcnow()  # TODO-mg: Update to self.close_date
+        now = (self.close_date or datetime.utcnow()).replace(tzinfo=None)
```
xmatthias: Please don't use `utcnow()` - but use `datetime.utc(timezone.utc)` instead.

`datetime.utcnow()` is fluky as hell (you can't get a correct timestamp off of it unless your system time is also UTC) - as it gets you the date in UTC - but without having the timezone set.
I've even heared an interview with the maintainer of the datetime standard module, who says utcnow should no longer be used (but they can't remove it as it would break the world).

tests/conftest.py
samgermain: What about the one's that were formatted wrong before, I have format on save with autopep8 turned on, so it fixes parts [like this that were improperly formatted](https://i.imgur.com/0nYRHcc.png), I could turn format on save off and go back and change those if you want, but that seems like the wrong thing to do

freqtrade/persistence/models.py
samgermain: I get this error `AttributeError: type object 'datetime.datetime' has no attribute 'utc'`
When I use `datetime.utc(timezone.utc)`

freqtrade/persistence/models.py
samgermain: There is no exchange object that exists on the `LocalTrade` or `Trade` class, all we have to refer to the exchange is the string. Would we create an exchange object in order to do this?

freqtrade/persistence/models.py
xmatthias: yeah - it should be `datetime.now(timezone.utc)` ... sorry bout that

freqtrade/persistence/models.py
xmatthias: no, we assign the enum to a column in the trade / Local trade object - which determines the logic to calculate Interrests.

The moment we create a trade - we will know the exchange (that's done from either backtesting or freqtradebot - both of which have an exchange object) - and that's the moment we assign this (this is something that cannot change).

tests/conftest.py
xmatthias: it's a difficult question to ask...
we're not using autopep8 - but pure flake8 for syntax validation at the moment - so to me - what isn't flagged by flake8 is not technically wrong.

Seems like autopep8 does some more things than flake8 - i'm not even sure if autopep8 aims to generate flake8 compliant code (it should - but don't know for sure) - as such it might even result in flake8 incompliant code (i hope it doesn't - but flake is checked after pytest) :laughing:

freqtrade/persistence/models.py
samgermain: It's a little more complicated that daily/hourly. Binance has a daily interest rate, but the minimum is one hour of interest, whereas kraken has a 4 hour interest rate with a minimum of 4 hours interest

samgermain: What about a delegate? With the delegate function defined on each exchange class?

xmatthias: Trade objects are stored (and loaded) from the database - they're not "there all the time" - you can't persist a function (well technically you can - but we will not pickle functions into the database!) - so you'd have to reattach the function each time you load it from the database - which is a quite dangerous task

docs/leverage.md
```diff
@@ -0,0 +1,3 @@
+For shorts, the currency which pays the interest fee for the `borrowed` currency is purchased at the same time of the closing trade (This means that the amount purchased in short closing trades is greater than the amount sold in short opening trades).
```
xmatthias, line 3: Maybe it's intentional as "internal" docs for now - but this way it won't be shown anywhere.
It's a standalone file that's not included (search docs for `plugins.md`) - nor is it in mkdocs.yml (so it get's it's own link in the doc page).

Considering that it's "for now" not an exposed part - i'm not sure what / where to go with this...

freqtrade/enums/interestmode.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class InterestMode(Enum):
+    """Equations to calculate interest"""
+
+    HOURSPERDAY = "HOURSPERDAY"
+    HOURSPER4 = "HOURSPER4"  # Hours per 4 hour segment
+    NONE = "NONE"
+
+    def __call__(self, *args, **kwargs):
```
xmatthias: i'm not convinced that this will work (it's not obvious unless you've been there before).

While the method per se (the content) will work - i don't think you can call the enum "in the real world" if it's assigned to a database column.

Consider this:
* Create the trade - assign the Enum to the appropriate field
* store the trade to the database (the enum is converted to string)
* retrieve the trade from the database
* the type is now a string (depending on the underlying database system)

As far as i know, Sqlalchemy will not bring the enum type back according to the [docs](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.Enum).
It's not explicitly mentioned - but the fact that the provided content can be a string tells me that sqlalchemy won't "reparse" it to a enum afterwards - otherwise i assume they'd mention that explicitly.

docs/leverage.md
samgermain: I put `- Leverage: leverage.md` inside `mkdocs.yml`, right after backtesting

freqtrade/persistence/models.py
```diff
@@ -155,6 +158,8 @@ def update_from_ccxt_object(self, order):
         self.average = order.get('average', self.average)
         self.remaining = order.get('remaining', self.remaining)
         self.cost = order.get('cost', self.cost)
+        self.leverage = order.get('leverage', self.leverage)
+        # TODO-mg: is_short?
```
xmatthias: No i don't think so (i don't think an Order should have the "is_short" flag at all) - it's an order - it's either buy or sell - it's the sequence of execution that decides if the order results in a short or not.

freqtrade/persistence/models.py
```diff
@@ -278,6 +298,54 @@ def open_date_utc(self):
...
+                    self.stop_loss = min(stop_loss, liquidation_price)
+                else:
+                    self.stop_loss = max(stop_loss, liquidation_price)
+            else:
+                self.stop_loss = liquidation_price
+                self.initial_stop_loss = liquidation_price
+            self.liquidation_price = liquidation_price
+        else:
+            # programmming error check: 1 of liqudication_price or stop_loss must be set
+            assert stop_loss is not None
```
xmatthias: Please don't use assert in code (outside of tests) - but use proper (explicit) error-checking 

while some tutorials recommend it, it's usually for quick scripts.
it's little known that you can make python ignore asserts (do nothing) with one flag passed into python.

freqtrade/persistence/models.py
```diff
@@ -278,6 +298,54 @@ def open_date_utc(self):
...
+
+        if liquidation_price is not None:
+            if stop_loss is not None:
+                if self.is_short:
+                    self.stop_loss = min(stop_loss, liquidation_price)
+                else:
+                    self.stop_loss = max(stop_loss, liquidation_price)
+            else:
+                self.stop_loss = liquidation_price
+                self.initial_stop_loss = liquidation_price
```
xmatthias: I'm a little worried about this section.

While it's not about 100% test-coverage - it shows that this part is not hit at all - but it's a pretty important piece of code to get right... to not break "stuff" ...

freqtrade/persistence/models.py
```diff
@@ -259,16 +264,31 @@ class LocalTrade():
     strategy: str = ''
     timeframe: Optional[int] = None
 
-    def __init__(self, **kwargs):
-        for key in kwargs:
-            setattr(self, key, kwargs[key])
-        self.recalc_open_trade_value()
+    # Margin trading properties
+    interest_rate: float = 0.0
+    liquidation_price: Optional[float] = None
```
xmatthias: I've been thinking about liquidation price - and I'm not sure it makes sense to keep/store that in the database.

While the liquidation price is set for one trade - it'll change the moment another trade is opened (less collateral available) - so we'd need to adjust this for all open trades whenever a trade is opened or closed.
It'll however also change whenever you withdraw or deposit - which will make it very difficult to handle if we treat it as something "static" - that lives in the database.

tests/conftest_trades.py
```diff
@@ -303,3 +303,178 @@ def mock_trade_6(fee):
...
+    return {
+        'id': '12368',
+        'symbol': 'DOGE/BTC',
+        'status': 'closed',
+        'side': 'sell',
+        'type': 'limit',
+        'price': 0.128,
+        'amount': 123.0,
+        'filled': 123.0,
+        'remaining': 0.0,
```
xmatthias, line 421: Maybe a stupid question - if you enter with `'leverage': 5` - don't we have to exit with `'leverage': 5` too?

freqtrade/enums/interestmode.py
```diff
@@ -0,0 +1,30 @@
...
+
+        borrowed, rate, hours = kwargs["borrowed"], kwargs["rate"], kwargs["hours"]
```
xmatthias: If you hard-require these kwargs anyway, it would be better to explicitly list them.
That makes
* mypy tell you if you mess something up
* gives you code-completion while typing in the editor.

xmatthias (commented):
I'll leave it at that for now - but i'm still missing a detailed review of the calculations and of the 2 new test files - and of performance implications to backtesting/hyperopt (which can potentially call these methods many many times).

Obviously - everything i spot there can point me to something else again - but in general i think we're pretty close.

docs/leverage.md
xmatthias: not commited / not part of this PR?
It's not showing up in the list of changed files for this PR ...

samgermain: It was locally sorry, I should have just not said anything

freqtrade/persistence/models.py
samgermain: In isolated trading mode it won't change. I'm not exactly sure how it works for cross trading mode or on exchanges other than Binance

tests/conftest_trades.py
samgermain: I guess so, it's already on the trade object from the entering trade so I figured I didn't need to include it again, but I added it on

freqtrade/enums/interestmode.py
samgermain: If it doesn't work then couldn't we just convert the string to the enum once we pull it out again? I'm not sure what to do about this, I am using sqlalchemy's enum type `interest_mode = Column(Enum(InterestMode), nullable=True)
`, and to me the fact that they have the type to begin with hints that it will be of that type when its taken out, otherwise I don't know why they would create it. It seems kind of pointless if they're just going to convert it to a string.

freqtrade/persistence/models.py
samgermain: So then is short should also be taken out of `migrate_orders_table` inside `persistence/migrations.py`

samgermain:
> I'll leave it at that for now - but i'm still missing a detailed review of the calculations and of the 2 new test files - and of performance implications to backtesting/hyperopt (which can potentially call these methods many many times).
> 
> Obviously - everything i spot there can point me to something else again - but in general i think we're pretty close.

The two new test files are essentially the `test_persistence.py` test file again, but with the added variables of leverage and shorting.

I added the calculations and source for the interest equations, but for the profit ratio calculation I've just been going by what people have been saying in discord. For longs the only difference is that I multiply the result by the leverage. 

I don't know what the performance implications to backtesting/hyperopt are, I also don't know how to check/test for that. I didn't add any loops so that's a good sign.

freqtrade/enums/interestmode.py
xmatthias: Databases usually don't have a "enum" type - so sqlalchemy needs to store it as string internally anyway.

The docs says that it'll make sure only the values in the enum can be stored (basically a constraint in the database if possible/supported) - which makes having an Enum type a valid case.
Converting it back is surely possible, and should be done within the method that uses it (we should not hook into sqlalchemy logic for reading a value).

To be clear, I didn't test this - so you can surely script a quick (non-freqtrade related ? ) script testing this out against the various databases - i'm just saying that i've never seen this (which only means it's probably new) - and that the docs don't seem promising to me based on the wording.

freqtrade/persistence/models.py
xmatthias: yeah absolutely (sorry i didn't notice this before).

freqtrade/persistence/models.py
xmatthias: I don't think we'll be supporting isolated trading mode at all.

it'll require to transfer funds to that "pair's margin wallet" - and as such will be impossible to handle when trading multiple pairs.
assume i trade 30 pairs - i'd have to have balance in each of these isolated pairs - which will make utilizing the whole balance almost impossible.

tests/conftest_trades.py
xmatthias: it's not so much the need - but the fact that this value is there on the response object (the "orders" part of this file simulates responses from the exchange).

xmatthias:
> The two new test files are essentially the test_persistence.py test file again, but with the added variables of leverage and shorting.

I know that - still will want to understand how the tests work - reviewing tests had me discover bugs not just once ... :laughing: 

> I don't know what the performance implications to backtesting/hyperopt are, I also don't know how to check/test for that. I didn't add any loops so that's a good sign.

basically running the same backtest (different strategies with different characteristics) against this branch and against develop - using `time freqtrade ...` to measure the duration. Not very scientific - but works :laughing: 

Not having loops is a good thing to start with - but it's not the only thing to look at  - However calling multiple functions in sequence can also cause performance issues - as depending on where it's used, i've seen functions in the trade object called 100.000 times for one backtest ...

freqtrade/persistence/models.py
samgermain: What makes it impossible? There would just be a few actions to execute with each trade, a transfer, a borrow, a trade, then a trade, a repayment, and a transfer. I was planning on implementing this because the leverage is a lot higher, I still don't really understand futures, but I guess I should look into that

samgermain:
> I know that - still will want to understand how the tests work - reviewing tests had me discover bugs not just once ... 😆

Is there an example I can look at? If there's already one for tests/test_persistence.py then mine would be really similar

xmatthias:
> Is there an example I can look at? If there's already one for tests/test_persistence.py then mine would be really similar

there can't be i'd say - it's me reading through test cases / scenarios - being pointed at looking at something related, and in consequence finding that some case was not handled / tested.
It's not ... science, and not guaranteed it'll happen - but it's the reason i'll also read (or rather think!) through the tests (which will obviously take some time)

samgermain:
I wrote something up, and posted it in the original comment on this pull request. I don't know if it's enough, let me know if you want something.
For the profit ratio calculation, I don't really know what more to put, I've kind of just been going off what people in the discord have been saying

freqtrade/persistence/models.py
samgermain: I updated this part

freqtrade/persistence/models.py
xmatthias: To be honest, i'm tempted to start with 0 leverage (only allowing short) for the beginning - although i'm not sure that's possible on binance (short without leverage).

Adding isolated margin support might be a future enhancement, but is nothing i'll be accepting as part of the PR adding this functionality to start with.

same goes for futures i'd say (at least for non-perpetual futures) - as they will add a whole other level of complexity (expiration dates).

xmatthias:
> I wrote something up, and posted it in the original comment on this pull request. I don't know if it's enough, let me know if you want something.
> For the profit ratio calculation, I don't really know what more to put, I've kind of just been going off what people in the discord have been saying

I don't think you understood my comment above - it's not that i need something from you - all i need is a few hours of uninterrupted time to look at it in detail - no explanation can change this i think :laughing:

freqtrade/enums/interestmode.py
samgermain: > Converting it back is surely possible, and should be done within the method that uses it (we should not hook into sqlalchemy logic for reading a value).

I don't really know how to implement that conversion. What methods would have to be modified to do that?

> so you can surely script a quick (non-freqtrade related ? ) script testing this out against the various databases -

I've never really used `sqlalchemy`, so it might not be that quick, but I might be able to try something

samgermain:
@xmatthias Don't worry about checking backtesting, I'm running a script that will put the results from a bunch of strategies into a csv. I'll run it for develop and this branch

freqtrade/persistence/models.py
samgermain: There's always something borrowed when you're shorting on any exchange, you're selling something you don't have, with the hopes of buying it again when it's cheaper so there will always be a liquidation price for shorts

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+                = 0.0075414886436454 * 0.0005  * ceil(2)   = 7.5414886436454e-06 base
+                = 0.0075414886436454 * 0.00025 * ceil(9/4) = 5.65611648273405e-06 base
+                = 0.0150829772872908 * 0.0005  * ceil(9/4) = 2.26244659309362e-05 base
+                = 0.0150829772872908 * 0.00025 * ceil(2)   = 7.5414886436454e-06 base
+    """
+
+    trade = Trade(
+        pair='ETH/BTC',
+        stake_amount=0.0037707443218227,
+        amount=275.97543219,
```
xmatthias, line 40: I think this would be a lot simpler by using "USDT" pairs (simply because our mind calculations work better with numbers > 0).
it'll make the calculation easier (as the result will not be 1e-06).

Also - the precision used above is unrealistic - no exchange will allow you 10 decimals on any number (the maximum is usually 8 for BTC pairs, and 3-4 for USDT pairs iirc).

THis also makes "recalculating" this with the [kraken fee calculator](https://kraken-fees-calculator.github.io/) impossible - as they don't allow these numbers.

freqtrade/enums/interestmode.py
xmatthias: I've done some tests and this seems to really work.
will need to investigate if we can apply this to some other columns, too.

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+        open_rate=0.00001099,
```
xmatthias: I think this would be a lot simpler by using "USDT" pairs (simply because our mind calculations work better with numbers > 0).
it'll make the calculation easier (as the result will not be 1e-06).

Also - the precision used above is unrealistic - no exchange will allow you 10 decimals on any number (the maximum is usually 8 for BTC pairs, and 3-4 for USDT pairs iirc).

This also makes "recalculating" this with the [kraken fee calculator](https://kraken-fees-calculator.github.io/) impossible - as they don't allow these numbers.

aside from this - is that "kraken free calculator" an official calculator? 
i find it strange that using different values for leverage (no other change) does not change the "gross margin" - only the fees (so assuming an invest of 1000, with open price 1 and closing price 1.1021 always gives 102.10 gross margin - at no leverage and at 5x leverage).

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+    )
+    assert trade.close_profit is None
+    assert trade.close_date is None
+    assert trade.is_open is True
+    trade.close(0.2)
+    assert trade.is_open is False
+    assert trade.close_profit == round(2.96708229426434, 8)
+    assert trade.close_date is not None
+
+    # TODO-mg: Remove these comments probably
```
xmatthias: yes?

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+        open_rate=0.00001099,
+        open_date=datetime.utcnow() - timedelta(hours=0, minutes=10),
+        fee_open=fee.return_value,
+        fee_close=fee.return_value,
+        exchange='binance',
+        leverage=3.0,
+        interest_rate=0.0005,
+        interest_mode=InterestMode.HOURSPERDAY
+    )
+    # 10 minutes round up to 4 hours evenly on kraken so we can predict the them more accurately
```
xmatthias: Is this comment accurate in this case (this is the binance test)

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+    assert trade.close_date is None
+    limit_lev_buy_order['status'] = 'open'
+    trade.update(limit_lev_buy_order)
+    assert trade.open_order_id is None
+    assert trade.close_profit is None
+    assert trade.close_date is None
+
+
+@pytest.mark.usefixtures("init_persistence")
+def test_calc_open_trade_value_lev(market_lev_buy_order, fee):
```
xmatthias: I'm not sure this test makes sense.

While it does add leverage over the "original" test - the method it tests has nothing to do with leverage - so i think we would not need this test.

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+        stake_amount=0.0009999999999226999,
+        open_rate=0.01,
+        amount=5,
+        open_date=datetime.utcnow() - timedelta(hours=4, minutes=55),
+        fee_open=fee.return_value,
+        fee_close=fee.return_value,
+        exchange='binance',
+        interest_rate=0.0005,
+        interest_mode=InterestMode.HOURSPERDAY
+    )
```
xmatthias: How's that test work?
While it seems to assume and test leverage - leverage is not actually set... so based on this - i'm not quite certain what it actually tries to test.

Also adding different leverage values don't have any effect on the result (so a trade with leverage=3 and leverage=5 produces the same final profit).

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+    # Is slightly different due to compilation time changes. Interest depends on time
+    assert round(trade.calc_close_trade_value(), 11) == round(0.003193788481706411, 11)
+    # Profit in BTC
+    assert round(trade.calc_profit(), 8) == round(0.00018628848193889054, 8)
+    # Profit in percent
+    assert round(trade.calc_profit_ratio(), 8) == round(0.18582392214792087, 8)
+
+
+@pytest.mark.usefixtures("init_persistence")
+def test_trade_close_lev(fee):
```
xmatthias: I think this test would offer itself for a parametrization - adding leverage as parameter (use 0/1, 3, 5? ) - and then evaluating the different results (obviously more leverage = more profit in case of wins)

might even make sense to do the same with losing trades (make sure the calculation is correct there too).

tests/persistence/test_persistence_leverage.py
```diff
@@ -0,0 +1,638 @@
...
+        fee: 0.25% base
+        interest_rate: 0.05% per day
+        open_rate: 0.00001099 base
+        close_rate: 0.00001173 base
+        amount: 272.97543219 crypto
+        stake_amount: 0.0009999999999226999 base
+        borrowed: 0.0019999999998453998  base
+        time-periods: 10 minutes(rounds up to 1/24 time-period of 1 day)
+        interest: borrowed * interest_rate * time-periods
+                    = 0.0019999999998453998 * 0.0005 * 1/24 = 4.166666666344583e-08 base
```
xmatthias: The amount of decimals used in these tests (all tests, not just this one) makes it near impossible to sanely (without spending days doing the math) determine if the calculation is correct or not.

tests/persistence/test_persistence_short.py
```diff
@@ -0,0 +1,780 @@
...
+        interest_mode=InterestMode.HOURSPERDAY
+    )
+
+    assert float(round(trade.calculate_interest(), 8)) == 0.04791240
+    trade.open_date = datetime.utcnow() - timedelta(hours=0, minutes=10)
+    assert float(round(trade.calculate_interest(interest_rate=0.00025), 8)) == 0.00479124
+
+
+@ pytest.mark.usefixtures("init_persistence")
+def test_calc_open_trade_value_short(market_short_order, fee):
```
xmatthias: Maybe parametrize this test (and check if that can remove a 2nd test).

what i'd like to see is the difference between "long" and "short" (that's what this test tests) - with the remaining params left identical (obviously you'll have 2 parameters, is_short and expected (1 and 2 - probably).

While i think both ways are tested in 2 seperate tests - combining and parametrizing that will make it obvious at a glance how we expect the result to differ.

freqtrade/persistence/models.py
```diff
@@ -278,6 +297,66 @@ def open_date_utc(self):
...
+        else:
+            self.initial_stop_loss = liquidation_price
+            self.stop_loss = liquidation_price
+
+        self.liquidation_price = liquidation_price
+
+    def __repr__(self):
+        open_since = self.open_date.strftime(DATETIME_PRINT_FORMAT) if self.is_open else 'closed'
+
+        return (f'Trade(id={self.id}, pair={self.pair}, amount={self.amount:.8f}, '
```
xmatthias: I think this should also include the side (maybe also leverage?).

xmatthias (commented):
I think many of the new tests can/should be consolidated to be parametrized tests.

While seperate tests also work - having the values "combined" as parameters makes it clear that nothing else has been modified for the test - and clearly shows the expected differences (this applies to some tests commented above, but also for example the different interrest modes).
Having them as parameters will also make it easier to add a new calculation method later on - as the corresponding test is just a 2-liner (add new parametrs to the `@pytest.mark.parametrize` heading).

i'm also having a very hard times with the usage of decimals.

That's unrealistic for exchanges (they will not allow that anyway) -
and on top of that, it's also very difficult to review and detect changes if the value that changed is the 5th decimal (or from e-6 to e-7).

usability should future devs need to modify things in that area will be very difficult.
i struggle with that already with some of the existing tests - so i'm (slowly) trying to improve that anyway).

freqtrade/persistence/models.py
xmatthias: yeah that's correct.

however liquidation price will change independently of the current trade - so are we sure that storing that as part of the trade object is a good idea?

tests/persistence/test_persistence_leverage.py
samgermain: I sent kraken an email about that calculator and [this is the conversation we had](https://i.imgur.com/i47aehi.png)

tests/persistence/test_persistence_leverage.py
samgermain: That comment is outdated

tests/persistence/test_persistence_leverage.py
samgermain: I was just checking that nothing changed

tests/persistence/test_persistence_leverage.py
samgermain: The interest value would be different. The leverage is set in `tests/conf.py`

freqtrade/persistence/models.py
samgermain: So I think cross margin is actually the more complicated one. 

With cross margin, you need to keep track of the overall equity in your wallet/account, and if your equity drops below a certain threshold, then all trades are liquidated. Is freqtrade able to keep track of prices in real time like that to know if the account equity has dropped passed that point? If it can't then we need to ask the exchange or get some kind of notification from it if liquidation is starting to happen. If the liquidation price is passed we need to exit all trades.

With isolated margin, there's just one ccxt command to transfer money in, and one to transfer out on buy and sell, and then the liquidation price is static unless money is part of the trade is sold/short_exited

`response = binance.transfer('USDT', 16, 'main', 'margin')`

---------------

Isolated perpetual futures seems like it might actually be the least complicated, because it doesn't involve any transferring of funds. You put the money is futures, and can trade that money on any isolated pair. I still don't completely understand futures though.

-------------

Here's an[ explanation of how the liquidation price is calculated on cross margin with binance](https://www.binance.com/en/support/faq/f6b010588e55413aa58b7d63ee0125ed).


I don't think that starting with cross-margin is a way to make implementing this less complicated is the main point I'm trying to make

tests/persistence/test_persistence_leverage.py
xmatthias: mmmh, so relying on that might be risky (not security-wise - but to get the calculation wrong).

That's not to say it IS wrong - but it means we'll need to scrutinize the calculator before determining "these are the calculations kraken uses".

samgermain: > i find it strange that using different values for leverage (no other change) does not change the "gross margin" - only the fees (so assuming an invest of 1000, with open price 1 and closing price 1.1021 always gives 102.10 gross margin - at no leverage and at 5x leverage).

The way leverage on Kraken works is the amount is always the same, but the stake_amount is lower when you use higher leverage. So if your spending `10 USDT` to get `10 BUSD` and you apply `2x` leverage, then you're spending `5 USDT` to get `10 BUSD`, with 3x leverage, you're spending `3.3333 USDT` for `10 BUSD`

Kraken's leverage is [explained here](https://support.kraken.com/hc/en-us/articles/203053116-How-leverage-works-in-spot-transactions-on-margin) 

That might explain some of what's going on

samgermain: I might end up dropping support for Kraken if we don't implement cross margin right away, I could possibly add support for... Bybit? Was that the other one you suggested?

xmatthias: FTX ... i'm honestly not keen on bybit - as it's currently not a supported exchange (which basically means, i never ran anything on it) - and adding a new exchange as part of this is surely not a good idea.

samgermain: Oh ok, what are all the supported exchanges? I know that bittrex doesn't have margin trading at all

tests/persistence/test_persistence_leverage.py
xmatthias: @samgermain forgot to push?

samgermain: Haven't pushed yet, I take it I should only click resolve once I push

xmatthias: yes(ish) ... 
i guess it's fine if you've got it commited and it's "ready to push" - waiting for some further changes.

Otherwise, it'll become dangerous for both of us to forget ...

freqtrade/persistence/models.py
samgermain: So if isolated trading is used, then `liquidation_price` should be on the `Trade`/`LocalTrade` class (maybe the name should be `isolated_liq` ?), if cross margin is used, then we will have to create a `maintenance_margin` variable(which really sounds like it could be it's own class) inside of `FreqtradeBot` or something, but cross margin isn't something I think I would personally use and I would rather leave it out of these PRs

tests/persistence/test_persistence_leverage.py
samgermain: Let's add FTX as another PR after this one get's merged?

tests/persistence/test_persistence_leverage.py
samgermain: Just a heads up, the default will be 1x leverage, 0x leverage will mean 0 * amount = 0, and no leverage values below 1 will be allowed

tests/persistence/test_persistence_leverage.py
xmatthias: for sure - i'm not even sure the decision on "which" exchanges needs to be taken at this point ... 

I'm just ... hesitant to only add binance - as i think their API is the strangest of all.

tests/persistence/test_persistence_leverage.py
samgermain: I tried my best to fix it, still some values with a lot of decimals though

tests/persistence/test_persistence_short.py
samgermain: I did something similar to that

freqtrade/persistence/models.py
samgermain: In another PR, that will change a lot of tests

samgermain: By side do you mean short/long ? It already prints buy/sell I think

tests/persistence/test_persistence_leverage.py
samgermain: Seeing as the liquidation price would be the only thing which would change, I'll keep writing support for kraken, and can test it by using a stoploss high enough that the liquidation price would never trigger. I'll make sure kraken is disabled before merging onto develop also. This way in order to enabled margin on kraken, the only feature that would need to be added would be a maintenance margin

samgermain:
Not sure why some tests fail on this commit, they pass locally, and all this commit did was change a variable name

freqtrade/persistence/models.py
```diff
@@ -232,7 +236,7 @@ class LocalTrade():
     close_rate_requested: Optional[float] = None
     close_profit: Optional[float] = None
     close_profit_abs: Optional[float] = None
-    stake_amount: float = 0.0
+    stake_amount: float = 0.0  # TODO: This should probably be computed
```
xmatthias, line 239: I'm not sure i understand this TODO.

stake-amount is the value (in stake-amount) that you invest.
so assuming a XRP/USDT trade - you buy XRP for 1000$ - so 1000$ is your stake-amount.
This is irrelevant of side (you might buy for 1000$, or sell for 1000$) - and also irrelevant of leverage (leverage will determine the amount of XRP you get).
In it's current form - it will also not change after the trade has been entered - so having it computed does not provide any visible benefit.

Moving this to "computed" will be problematic for other reasons, too, as it's a very important measure for "post-mortem" analysis - and as such - should be in the database.

freqtrade/persistence/models.py
```diff
@@ -361,8 +454,11 @@ def adjust_min_max_rates(self, current_price: float) -> None:
 
     def _set_new_stoploss(self, new_loss: float, stoploss: float):
         """Assign new stop value"""
-        self.stop_loss = new_loss
-        self.stop_loss_pct = -1 * abs(stoploss)
+        self.set_stop_loss(new_loss)
```
xmatthias: I think this "calling-logic" is dangerous.
we now have 3 "set stoploss" methods that all work slightly differently - this should definitely be "somehow" simplified and unified to avoid confusion.

it's also unclear to me what extracting `set_stop_loss` brings over putting that logic to this method.
calling `set_stop_loss` from outside of this method will also cause invalid values - as `stop_loss_pct` will not be updated.
as `set_stop_loss()` is (based on naming conventions) a public method - it's a dangerous premise, and one that might hit us with obscure bugs if we forget that we should not use that method directly.

tests/test_persistence.py
```diff
@@ -103,41 +124,450 @@ def test_update_with_binance(limit_buy_order, limit_sell_order, fee, caplog):
...
+        is_short=True,
+        leverage=3.0,
+        interest_rate=0.0005,
+        interest_mode=InterestMode.HOURSPERDAY
+    )
+    trade.update(limit_buy_order_usdt)
+    assert log_has_re(r"LIMIT_BUY has been fulfilled for Trade\(id=226531, "
+                      r"pair=ETH/BTC, amount=30.00000000, "
+                      r"is_short=True, leverage=3.0, open_rate=2.00000000, open_since=.*\).",
                       caplog)
```
xmatthias, line 578: Would we want a corresponding test for sell below here (The unleveraged part tests both buy and sell updating ...)?

freqtrade/persistence/models.py
```diff
@@ -278,6 +297,75 @@ def open_date_utc(self):
...
+                self.stop_loss = max(self.stop_loss, isolated_liq)
+        else:
+            self.initial_stop_loss = isolated_liq
+            self.stop_loss = isolated_liq
+
+        self.isolated_liq = isolated_liq
+
+    def set_is_short(self, is_short: bool):
+        self.is_short = is_short
+        self.recalc_open_trade_value()
```
xmatthias: What's the purpose of this `set_is_short()` method?

I'd assume that the side (short / long) is determined at trade open - so i don't see this as something that can magically "change" later (you either bought - or you sold). You should (must!) know the side of the trade before you create a trade object.

freqtrade/persistence/models.py
samgermain: It doesn't have to be. I did run into some problems caused by it in the `borrowed` variable earlier when it was

```
elif not self.is_short:
            return self.stake_amount * ((self.leverage-1)/self.leverage)
```

and I was changing the amount within my tests, which is why I updated this calculation to

```
elif not self.is_short:
            return (self.amount * self.open_rate) * ((self.leverage-1)/self.leverage)
```

I cant remember exactly what the problem was, but I remember the values not being synced

xmatthias (commented):
I think we're pretty close with this one now - only a few ... doubts i think.

I've pointed this branch at `feat/short` now - as i'd like to use that branch for all margin/short PR's (not the ones that only refactor / cleanup things).
This will ensure that we can work on this without paying 100% attention to not breaking existing functionality - assuming we put a very clear `TODO short:` there - which MUST be fixed before `feat/short` is merged into develop - to "reestablish" existing functionality should we need this.

freqtrade/persistence/models.py
samgermain: I guess it was mostly for testing, just to make sure that `self.recalc_open_trade_value()` was always called after

freqtrade/persistence/models.py
xmatthias: well it might be a testing problem (where you change one thing over the other between different asserts)
i'd not take that assumption and apply it to the real world (as said, stake-amount is the value you invest - not some "random" value that's calculated). You paid for that - exactly that ...

samgermain:
>  - assuming we put a very clear `TODO short:` there 

I've been using `TODO-mg`, which stands for margin

xmatthias:
> I've been using `TODO-mg`, which stands for margin

That works perfectly fine.
i saw it and honestly thought it stands for your name (which includes "mg" after all - so i thought maybe that's the 2-letter shortcut you use sometimes :laughing: )

xmatthias (changes requested):
To be honest, this PR was very close to merging - and i think the changes you made i asked for would've been ok - but now it's another 12 new commits which add new classes, new unfinished enums, ... 

Now you've added that whole liquidation stuff to this branch (which is a seperate PR, too).
Please keep these things seperate (especially if there is a Branch/PR for it already!).

samgermain:
> To be honest, this PR was very close to merging - and i think the changes you made i asked for would've been ok - but now it's another 12 new commits which add new classes, new unfinished enums, ...
> 
> Now you've added that whole liquidation stuff to this branch (which is a seperate PR, too).
> Please keep these things seperate (especially if there is a Branch/PR for it already!).

All right, I rolled it back

xmatthias:
> All right, I rolled it back

Great, thanks!
One last request (hopefully) for this PR - if you're able to merge develop (or feat/short) onto this branch to resolve conflicts it'll be good to go (unless there's something wrong with the merge, obviously).

freqtrade/persistence/migrations.py
```diff
@@ -47,6 +47,13 @@ def migrate_trades_table(decl_base, inspector, engine, table_back_name: str, col
...
     sell_reason = get_column_def(cols, 'sell_reason', 'null')
     strategy = get_column_def(cols, 'strategy', 'null')
+
+    leverage = get_column_def(cols, 'leverage', '1.0')
+    interest_rate = get_column_def(cols, 'interest_rate', '0.0')
+    isolated_liq = get_column_def(cols, 'isolated_liq', 'null')
+    # sqlite does not support literals for booleans
+    is_short = get_column_def(cols, 'is_short', '0')
+    interest_mode = get_column_def(cols, 'interest_mode', 'null')
     buy_tag = get_column_def(cols, 'buy_tag', 'null')
```
xmatthias: I would've left this line "on top" - simply to have them in "order they were added" ... but that's not really relevant.

You will however need to adjust `if not has_column(cols, 'buy_tag'):` in line 174 of this file to one of the new columns to properly trigger migrations - otherwise they'll simply not run if someone has the buy_tag column, but not the "margin" columns - as these columns are the new ones.

samgermain: I don't really understand what you mean by the second part, but I moved buy_tag up

xmatthias (approved):
I think this is good for now :+1: great job :100:  

I'll consider the calculations "theoretical" for now - and we'll obviously have to test them in practice once the rest is ready (doing a trade of the different types, and comparing results on exchange with the calculations).

It's not so much a mis-trust in the implementation, more a along the lines of "i don't 100% trust the exchange documentation all the time" ... 

One additional note - it would be great if you could rebase the "other" PR's on top of `feat/short` - so the diffs don't show 100ds of commits due to previous rebases here.

samgermain:
🎉

freqtrade/persistence/models.py
samgermain: > In Cross Margin Mode, the margin can only be shared between the same type of asset. For example, in Cross Margin Mode, all BUSD in the USDⓈ-M Futures Wallet can be used for all BUSD-margined contracts.

[source](https://www.binance.com/en/support/faq/360033162192)

PR #5170, opened 2021-06-22
make "/profit N" command output be consistent with "/daily" command

eschava:
## Summary
Currently "/profit N" command shows statistics for all trades _started_ since N days ago. Other commands like "/daily" show statistics based on trade's close date. Need to change the output of the command to include only trades _closed_ since N days ago plus all open trades

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -339,7 +339,10 @@ class RPC:
             self, stake_currency: str, fiat_display_currency: str,
             start_date: datetime = datetime.fromtimestamp(0)) -> Dict[str, Any]:
         """ Returns cumulative profit statistics """
-        trades = Trade.get_trades([Trade.open_date >= start_date]).order_by(Trade.id).all()
+        trade_filter = \
+            (Trade.is_open.is_(False) & (Trade.close_date >= start_date)) | \
+            Trade.is_open.is_(True)
+        trades = Trade.get_trades(trade_filter).order_by(Trade.id).all()
 
         profit_all_coin = []
         profit_all_ratio = []
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -339,9 +339,8 @@ class RPC:
             self, stake_currency: str, fiat_display_currency: str,
             start_date: datetime = datetime.fromtimestamp(0)) -> Dict[str, Any]:
         """ Returns cumulative profit statistics """
-        trade_filter = \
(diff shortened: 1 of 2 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -339,7 +340,10 @@ def _rpc_trade_statistics(
             self, stake_currency: str, fiat_display_currency: str,
             start_date: datetime = datetime.fromtimestamp(0)) -> Dict[str, Any]:
         """ Returns cumulative profit statistics """
-        trades = Trade.get_trades([Trade.open_date >= start_date]).order_by(Trade.id).all()
+        trade_filter = \
+            (Trade.is_open.is_(False) & (Trade.close_date >= start_date)) | \
+            Trade.is_open.is_(True)
```
xmatthias: If you wrap the whole expression in an additional set of parenteses, you can "save" the `\` line concatenation - which i find rather ugly.

freqtrade/rpc/rpc.py
```diff
@@ -339,7 +339,9 @@ def _rpc_trade_statistics(
             self, stake_currency: str, fiat_display_currency: str,
             start_date: datetime = datetime.fromtimestamp(0)) -> Dict[str, Any]:
         """ Returns cumulative profit statistics """
-        trades = Trade.get_trades([Trade.open_date >= start_date]).order_by(Trade.id).all()
+        trade_filter = ((Trade.is_open.is_(False) & (Trade.close_date >= start_date)) |
+                        Trade.is_open.is_(True))
```
xmatthias, line 343: i'm ... not quite sure i see consistency (but it's not the fault of this PR).

running `/profit 2` i get "Showing Profit since: 3 days ago" ...
while `/daily 2` - i get 2 days - missing the 3rd day (so one of the 2 is adding a day - and it's not `/daily` that's wrong if you ask me).

It is aligned when using `/daily 2` combined with `/profit 1` ... but i kindof struggle to see this as alignment ... 

I'm fine with merging this - but maybe you have the will / interrest to have a look at the issue testing this exposed ...? (just let me know so i'll either leave this open - or approve and merge it).

eschava: Agree, for me "/profit 1" command to see today's trades looks more natural than "/profit 0". So if you don't mind I'll change that as well

xmatthias: yes please, go for it

xmatthias (approved):
Great thanks!

LGTM :+1:

PR #5189, opened 2021-06-27
Implement strategy-controlled stake sizes

### LINKED ISSUE #5076: Implementing Dynamic stake amount that responds to market status(momentum, volatility, volume.. etc) supporting indicator with TA
I want to implement controlling bet ratio dynamically on market status (for example bet high[90%] on high trend, bet low on minus trend[10%], or use current market volatility for betting ratio) function for freqtrade. It's basically managing dynamic staking amount on market situations.
As you guys know, **money management** is extremely important for successful system trading. 
I actually have backtested/used this logic on own my stock trading system, (trend following / (bet ratio)money management with momentum indicator) and this helped me getting profit and _not losing much_ so greatly.
This technic is so useful for **high volatile** assets such as crypto.

My idea (plus an idea from someone from discord) is def custom_stake_amount(...) in interface.py and use that instead of getting stake_amount from config, and make custom dynamic staking options for config file, and code dynamic staking logic(with TA) in dynamic_stake_logic.py. 
Another way is adding an option to get stake_ratio(0%~100%) from strategies. Maybe this can be an easier workaround for dynamic staking. Because we need to use TA for regulating trading bet ratio.

Example dynamic stake logic is this : Ticker's n-day SMA Scoring(momentum) * Volatile Ratio

rokups:
Submitting initial draft of strategy-controlled stake sizes so you can get the feel of it.

What should be max stake size? Should it be `stake_amount` specified in config or entire available balance? Both have their merits i would say.

Otherwise tests pass, but i totally did not test the code so treat it as a rough sketch, a fuel for further discussion.

closes #5076

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -424,16 +424,10 @@ class FreqtradeBot(LoggingMixin):
 
         if buy and not sell:
             stake_amount = self.wallets.get_trade_stake_amount(pair, self.edge)
-            if not stake_amount:
-                logger.debug(f"Stake amount is 0, ignoring possible trade for {pair}.")
-                return False
-
-            logger.info(f"Buy signal found: about create a new trade for {pair} with stake_amount: "
-                        f"{stake_amount} ...")
 
             bid_check_dom = self.config.get('bid_strategy', {}).get('check_depth_of_market', {})
             if ((bid_check_dom.get('enabled', False)) and
-                    (bid_check_dom.get('bids_to_ask_delta', 0) > 0)):
+               (bid_check_dom.get('bids_to_ask_delta', 0) > 0)):
                 if self._check_depth_of_market_buy(pair, bid_check_dom):
                     return self.execute_buy(pair, stake_amount)
                 else:
@@ -488,12 +482,21 @@ class FreqtradeBot(LoggingMixin):
 
         min_stake_amount = self.exchange.get_min_pair_stake_amount(pair, buy_limit_requested,
                                                                    self.strategy.stoploss)
-        if min_stake_amount is not None and min_stake_amount > stake_amount:
-            logger.warning(
-                f"Can't open a new trade for {pair}: stake amount "
-                f"is too small ({stake_amount} < {min_stake_amount})"
-            )
+
+        if not self.edge:
+            max_stake_amount = self.wallets.get_available_stake_amount()
+            stake_amount = strategy_safe_wrapper(self.strategy.custom_stake_amount,
+                                                 default_retval=None)(
+                pair=pair, current_time=datetime.now(timezone.utc),
+                current_rate=buy_limit_requested, proposed_stake=stake_amount,
+                min_stake=min_stake_amount, max_stake=max_stake_amount)
+        stake_amount = self.wallets._validate_stake_amount(pair, stake_amount, min_stake_amount)
+
+        if not stake_amount:
             return False
+
+        logger.info(f"Buy signal found: about create a new trade for {pair} with stake_amount: "
+                    f"{stake_amount} ...")
 
         amount = stake_amount / buy_limit_requested
         order_type = self.strategy.order_types['buy']
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -486,7 +486,7 @@ class FreqtradeBot(LoggingMixin):
         if not self.edge:
             max_stake_amount = self.wallets.get_available_stake_amount()
             stake_amount = strategy_safe_wrapper(self.strategy.custom_stake_amount,
-                                                 default_retval=None)(
+                                                 default_retval=stake_amount)(
                 pair=pair, current_time=datetime.now(timezone.utc),
                 current_rate=buy_limit_requested, proposed_stake=stake_amount,
                 min_stake=min_stake_amount, max_stake=max_stake_amount)
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -319,7 +319,7 @@ class Backtesting:
         max_stake_amount = self.wallets.get_available_stake_amount()
 
         stake_amount = strategy_safe_wrapper(self.strategy.custom_stake_amount,
-                                             default_retval=None)(
+                                             default_retval=stake_amount)(
             pair=pair, current_time=row[DATE_IDX].to_pydatetime(), current_rate=row[OPEN_IDX],
             proposed_stake=stake_amount, min_stake=min_stake_amount, max_stake=max_stake_amount)
         stake_amount = self.wallets._validate_stake_amount(pair, stake_amount, min_stake_amount)
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -129,6 +129,8 @@ class Backtesting:
         """
         self.strategy: IStrategy = strategy
         strategy.dp = self.dataprovider
+        # Attach Wallets to Strategy baseclass
+        IStrategy.wallets = self.wallets
         # Set stoploss_on_exchange to false for backtesting,
         # since a "perfect" stoploss-sell is assumed anyway
         # And the regular "stoploss" function would not apply to that case
@@ -312,7 +314,18 @@ class Backtesting:
             stake_amount = self.wallets.get_trade_stake_amount(pair, None)
         except DependencyException:
             return None
-        min_stake_amount = self.exchange.get_min_pair_stake_amount(pair, row[OPEN_IDX], -0.05)
+
+        min_stake_amount = self.exchange.get_min_pair_stake_amount(pair, row[OPEN_IDX], -0.05) or 0
+        max_stake_amount = self.wallets.get_available_stake_amount()
+
+        stake_amount = strategy_safe_wrapper(self.strategy.custom_stake_amount,
+                                             default_retval=None)(
+            pair=pair, current_time=row[DATE_IDX].to_pydatetime(), current_rate=row[OPEN_IDX],
+            proposed_stake=stake_amount, min_stake=min_stake_amount, max_stake=max_stake_amount)
+        stake_amount = self.wallets._validate_stake_amount(pair, stake_amount, min_stake_amount)
+
+        if not stake_amount:
+            return None
 
         order_type = self.strategy.order_types['buy']
         time_in_force = self.strategy.order_time_in_force['sell']
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -304,6 +304,23 @@ class IStrategy(ABC, HyperStrategyMixin):
         """
         return None
 
+    def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: float,
+                            proposed_stake: float, min_stake: float, max_stake: float,
+                            **kwargs) -> float:
+        """
+        Customize stake size for each new trade. This method is not called when edge module is
+        enabled.
+
+        :param pair: Pair that's currently analyzed
+        :param current_time: datetime object, containing the current datetime
+        :param current_rate: Rate, calculated based on pricing settings in ask_strategy.
+        :param proposed_stake: A stake amount proposed by the bot.
+        :param min_stake: Minimal stake size allowed by exchange.
+        :param max_stake: Balance available for trading.
+        :return: A stake size, which is between min_stake and max_stake.
+        """
+        return proposed_stake
+
     def informative_pairs(self) -> ListPairsWithTimeframes:
         """
         Define additional, informative pair/interval combinations to be cached from the exchange.
--- a/docs/strategy-advanced.md
+++ b/docs/strategy-advanced.md
@@ -521,6 +521,37 @@ class AwesomeStrategy(IStrategy):
 
 ```
 
+### Stake size management
+
+It is possible to manage your risk by reducing or increasing or reducing stake amount when placing a new trade.
+
+```python
+class AwesomeStrategy(IStrategy):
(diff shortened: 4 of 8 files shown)


freqtrade/freqtradebot.py
```diff
@@ -488,13 +482,30 @@ def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = N
 
         min_stake_amount = self.exchange.get_min_pair_stake_amount(pair, buy_limit_requested,
                                                                    self.strategy.stoploss)
+
+        if not self.edge:
```
xmatthias, line 486: yeah edge ... 
it's honestly losing more and more of it's standalone possibilities - as most of the things it did are now possible to the strategy ... (every time i see a condition like this - i'm reminded of that).

freqtrade/strategy/interface.py
```diff
@@ -303,6 +303,22 @@ def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_r
         """
         return None
 
+    def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: float,
+                            proposed_stake: float, min_stake: float, max_stake: float,
+                            **kwargs) -> float:
```
xmatthias, line 309: i'd THINK some information about the current bot's trade(s) state might be useful if it's accessible easily (currently open trades, max_open_trades, current stake in trades [used stake]) - as proposed by @epigramx in #5076.


while the information is available _in theory_ - using `Trade.xxxx()` will make it more complex for the user than providing it as argument - unless there's a big overhead (which there might be for backtesting - eventually (!)).

Using `Trade.xxx()` in backtesting is (while in theory possible) a bit dangerous at the moment, as not all available methods are simulated for backtesting - and as such may return wrong / no results - and it's not documented as "possible for backtesting" either (mainly because of that).
By providing the info to the function we ensure that only allowed methods are used.

(convince me differently...)

freqtrade/strategy/interface.py
```diff
@@ -303,6 +303,22 @@ def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_r
...
+        Customize stake size for each new trade. This method is not called when edge module is
+        enabled.
+
+        :param pair: Pair that's currently analyzed
+        :param current_time: datetime object, containing the current datetime
+        :param current_rate: Rate, calculated based on pricing settings in ask_strategy.
+        :param proposed_stake: A stake amount proposed by the bot.
+        :param min_stake: Minimal stake size allowed by exchange.
+        :param max_stake: Balance available for trading.
+        """
```
xmatthias: We might wanna have a ":returns:" here .... ;)

xmatthias (commented):
overall, i'd say it feels pretty great already - and seems to do what it's supposed to.

> What should be max stake size? Should it be stake_amount specified in config or entire available balance? Both have their merits i would say.

i'm ... not sure about this either. 
I guess there's nothing wrong with "entire available balance".
if you limit it to "100$" because that was set as stake_amount - you'll only allow to scale down - never to scale up. 
In certain combinations (static stake + static max_open) this might have some benefits, as we'll risk having one trade block all balance (go all in) - which would cause the bot to have a lot of "failing" log messages thereafter as the bot sees a trade-slot available - but lacks the balance to enter it (obviously that check might need adjusting to check available balance > min_balance ...).

> the ones i mean: 

`2021-06-27 15:26:17,041 - freqtrade.freqtradebot - WARNING - Can't open a new trade for LTC/USDT: stake amount is too small (0.030290051900578874 < 11.05263157894737)`

freqtrade/strategy/interface.py
epigramx: I was about to say I don't mind since I would use the Trade class etc., but you make a good point if it can't be used for backtesting since it might be a bad idea to use the method at all (in practice) if it can't be backtested. 
So I would say it should be able to replace in full the current methods (which requires slot availability) and there is very little reason not to because the performance impact should be low because trade entries are rare.

rokups: Is there any reason backtesting does not use in-memory sqlite database? It could be an easy shortcut to get feature parity of this functionality in backtesting and live.

xmatthias: yes - speed.

I've experimented with this - and by using a Trade-object instead of a LocalTrade - you get a slowdown of around 5-10x (don't remember the exact numbers - and they depend on the strategy as well).
Creating the LocalTrade class was NOT fun - and i only did that due to the necessity - not because "i feel it's a great way to do things".

I've not even created a session - simply using the Trade object makes everything slow (because of the way sqlalchemy works when accessing it's dynamic attributes). 
Profiling it points to some methods 5-10 levels into the sqlalchemy callstack (when using `trade.xxx`) - and it's that many nested function-calls that it's simply not possible to do it fast with an sqlalchemy "enabled" object.

It's not a problem for real trading - where you only access the same attribute a few times per iteration - but in backtesting - where you'll want to do as many candles as possible in the shortest amount of time - this has a quite huge impact.

rokups: Almost sounds like sqlalchemy was overkill for such a simple database, and a plain object with explicit save method would both provide feature parity and be as fast as needed in all contexts, without hardly being an inconvenience considering how rarely fields are updated and how non-complicated updates are.

epigramx: An in-memory sqlite is also extremely slow compared to a simple in-memory list. Backtesting doesn't require its robustness anyway since if it is stops all data is expected to be lost anyway. It seems clearly better to hold that kind of basic data (like "number of slots used") into basic python primitives during backtesting.

rokups: Maybe so. But if sqlalchemy wasnt in a mix - we would essentially have a LocalTrade that can save on request, and in backtesting there would be no saving so no sqlite. But most importantly it would be same code used in both modes, not having to worry about overhead, not introducing any complexities.. It would be extremely useful if we could reliably get current trade info in backtesting. It would open new approaches in strategy development.

xmatthias: First of all - in `populate_*` methods, you'll never receive any reliable trade data (other than 0 Trades) - simply because they run "all at once" - and the trade-simulation starts later.

For the rest - sure - but if you look hard enough you'll notice that the possibility exists - maybe in a slightly limited form (only certain filters are possible).
it's however not documented on purpose - as i'm not sure it'll stay (i hope it doesn't - but didn't have time to look into a proper implementation) - and by not having it exposed through documentation - it'll not matter if it's removed later on).

epigramx: rokups explained to me in discord that `wallet_size`  from the current patch and  `max_open_trades` from the config can be used to reliably imitate 'unlimited' (and I guess variations of it) in backtesting. That means the patch might be already enough for anyone that would need 'open trades & total trades & stake available' data during backtesting. I guess a backtesting-available `Trade` class is not required for that purpose but it might be always useful for ease of use of functions like this.

rokups:
@xmatthias i propose getting rid of `min_stake` and `max_stake` parameters altogether. `max_stake` can be substituted by `self.wallets.get_available_stake_amount()` and `min_stake` should be irrelevant. Also i propose to change bot behavior to clamp stake size to allowed range instead of throwing an error and bailing on a trade like it does now. Reason is same as why we use `strategy_safe_wrapper()` - users are careless and they will mess up, but there is no need to skip on a trade. Maybe returning 0 stake amount could be an exception to this rule, allowing to discard a trade this way. What do you think?

xmatthias:
I don't see why `min_stake` would be irrelevant. 
It may for you if you're using big(ish) stakes - but for a new user trying a strategy close to minimum stake (maybe with a 100$ account), it's very much relevant (and can also influence the calculation itself - if you know you can't go below 15$).
Sure, we'll also min-cap it by this - but you can end up with randomly rejected trades because of it - for no apparent reason.

regarding `max_stake` - i don't see what we gain by not providing it other than increased complexity to the user.
Sure - it can be queried - but i'm pretty sure that most will require that anyway as part of the calculations....

rokups:
I am conflicted about these parameters because it is unclear where we draw the line. Sure `min_stake` and `max_stake` can be used in calculations. So can be used `wallet_size` in stake currency (including tied up stake). So can be used `max_open_trades` and `current_open_trades`. It is getting wild fast.

> Sure, we'll also min-cap it by this - but you can end up with randomly rejected trades because of it - for no apparent reason.

Part of suggestion is to enforce minimal stake size with a warning, instead of rejecting a trade. I implemented this already so please look at it and think on it whether it should be reverted. Seems to me this is much like errors in user strategy - these conditions can manifest many days after strategy started, so just like not bailing on exception from user strategy we probably should not bail on requested incorrect stake amount and instead clamp it to allowed range.

freqtrade/optimize/backtesting.py
```diff
@@ -312,7 +314,24 @@ def _enter_trade(self, pair: str, row: List) -> Optional[LocalTrade]:
...
+        if min_stake_amount is not None and stake_amount < min_stake_amount:
+            logger.error(
+                f"Stake amount for pair {pair} is too small ({stake_amount} < {min_stake_amount})"
+            )
+            return None
+        if stake_amount > max_stake_amount:
+            logger.error(
+                f"Stake amount for pair {pair} is too big ({stake_amount} > {max_stake_amount})"
+            )
+            return None
```
xmatthias: Is there a particular reason to reimplement this validation here instead of using `_validate_stake_amount` (maybe moved to a different location)?

freqtrade/freqtradebot.py
```diff
@@ -467,6 +461,28 @@ def _check_depth_of_market_buy(self, pair: str, conf: Dict) -> bool:
...
+            logger.debug(
+                f"Stake amount for pair {pair} is too small ({stake_amount} < {min_stake_amount}), "
+                f"adjusting to {min_stake_amount}"
+            )
+        if stake_amount > max_stake_amount:
+            stake_amount = max_stake_amount
+            logger.debug(
+                f"Stake amount for pair {pair} is too big ({stake_amount} > {max_stake_amount}), "
+                f"adjusting to {max_stake_amount}"
+            )
```
xmatthias: I would think that the log-level for this should be > debug (most likely info is the right value?).

Debug will only show when running freqtrade with `-v` - which will be VERY verbose as that includes a lot of other debug messages too.

xmatthias (commented):
> Part of suggestion is to enforce minimal stake size with a warning

This is perfectly fine the way you implemented (just not sure on the log-level for these messages).

---

There's however a difference between providing `min_stake` and `max_stake` - and the "open trades" values.

The trade related values are relatively easy to get from the strategy. You can't however get to `min_stake` without some wild hackary (exchange is NOT exposed to the strategy directly / in a documented way - which is on purpose).

I'd also want to provide `max_stake` due to #5247 - as i think we will need to have a similar method than `max_stake` - respecting the exchange's top limit (you might have 10.000$ as your stake amount - but if the exchange only allows 5000$ that'll need to be respected).
Having a documentation part that shows "get your wallet size this way" will fail in that case - as the wallet is potentially not the limiting factor.
 
While this will be a seperate PR  / change for sure - providing `max_stake` will simplify exposing this information to the strategy.

freqtrade/optimize/backtesting.py
rokups: Ah this was stupid mypy complaining about "too complex method". I would have kept it in original place..

rokups:
Yeah what you say makes sense. Ill get these parameters back 👍🏻 Although i am no longer sure backtesting should throw an error when min/max is not honored. If we want to free user from having to absolutely care for these details - it must be a warning, even though if it is easier to miss.

xmatthias:
> Although i am no longer sure backtesting should throw an error when min/max is not honored. If we want to free user from having to absolutely care for these details - it must be a warning, even though if it is easier to miss.

I'm not 100% sure on the implications of this - but i think showing warnings will suffice - it's possible to have the calculation "sometimes" below min/max - and if you get no buys/sells you'll (hopefully?) look at the log output ...

rokups:
I am suggesting to clamp requested stake size to min-max range so buys are not ignored like they are now.

xmatthias:
> I am suggesting to clamp requested stake size to min-max range so buys are not ignored like they are now.

Basically - if the strategy says "buy for 5$" - you'd increase this to 15$ (or whatever the minimum is) ? 
works fine for me - just make sure we got it documented :+1:

rokups:
I think i am done with machinery. What i do not know is how to make a test for this.

freqtrade/optimize/backtesting.py
```diff
@@ -312,7 +314,18 @@ def _enter_trade(self, pair: str, row: List) -> Optional[LocalTrade]:
             stake_amount = self.wallets.get_trade_stake_amount(pair, None)
         except DependencyException:
             return None
-        min_stake_amount = self.exchange.get_min_pair_stake_amount(pair, row[OPEN_IDX], -0.05)
+
+        min_stake_amount = self.exchange.get_min_pair_stake_amount(pair, row[OPEN_IDX], -0.05) or 0
+        max_stake_amount = self.wallets.get_available_stake_amount()
+
+        stake_amount = strategy_safe_wrapper(self.strategy.custom_stake_amount,
+                                             default_retval=None)(
```
xmatthias: I would think that we'd want the `default_retval` to be `stake_amount` - not None (i've changed this in the last commit - but would want your opinion on this too).

It'd apply whenever the method called raises an error.
Current behaviour would cause it to assume min_stake_amount - which is kinda surprising behaviour.

rokups: Good catch. Yeah that would be consistent with how other parts behave.

xmatthias (approved):
LGTM :+1: 

Nice work :100:

PR #5205, opened 2021-06-28
Added timerange and max open trades info above multiple strategy backtest result summary table

barisengez:
## Summary
When people are sharing backtest results; only to show the timerange that they backtested, they have to include the detailed table of the last strategy as in the image below. By this addition, people wont have to include the last strategy detailed result only to show the timerange

image

## What's new?
Added timerange above multiple strategy backtest result summary table

image

### DIFF
--- a/freqtrade/optimize/optimize_reports.py
+++ b/freqtrade/optimize/optimize_reports.py
@@ -660,8 +660,9 @@ def show_backtest_results(config: Dict, backtest_stats: Dict):
         show_backtest_result(strategy, results, stake_currency)
 
     if len(backtest_stats['strategy']) > 1:
-        backtest_timerange = f"{next(iter(backtest_stats['strategy'].items()))[1]['backtest_start']} -> " \
-                             f"{next(iter(backtest_stats['strategy'].items()))[1]['backtest_end']}"
+        backtest_timerange = \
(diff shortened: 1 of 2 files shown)


barisengez:
Adding a max trade count next to it might be beneficial as well. what do you think?
EDIT : Added it anyway, better if you review it this way

freqtrade/optimize/optimize_reports.py
```diff
@@ -662,7 +662,10 @@ def show_backtest_results(config: Dict, backtest_stats: Dict):
     if len(backtest_stats['strategy']) > 1:
         # Print Strategy summary table
 
+        first_strat_stats = next(iter(backtest_stats['strategy'].items()))[1]
```
xmatthias: I think this should be simplified to simply (reuse) `results` from the above loop (maybe not here but using results directly down below).

In python - the variable in a for loop will keep the value of the last loop. 
It will not have the first strategy results but the last ones - but i don't think that'll be a problem.

barisengez: you python people and your fancy useful shortcuts :) thanks. changed it that way

xmatthias (approved):
welcome onboard on being a "python people" 😆 

LGTM 👍

PR #5210, opened 2021-06-28
"/profit N" command should print best pair for the same period of time, not for all trades

eschava:
## Summary

"/profit N" command should print best pair for the same period of time, not for all trades


xmatthias (approved):
Nice catch :+1:

PR #5212, opened 2021-06-29
Trailing stoploss in backtesting v2

rokups:
As it turns out trailing stoploss still had one caveat in backtesting.
```py
    trailing_stop = True
    trailing_stop_positive = 0.01
    trailing_stop_positive_offset = 0.03
    use_custom_stoploss = True

    def custom_stoploss(self, pair: str, trade: Trade, current_time: datetime, current_rate: float,
                        current_profit: float, **kwargs) -> float:
        return -0.001
```
This code would cause trades to often sell on buy candle, however sell price calculation did not take `custom_stoploss` into account and assumed `trailing_only_offset_is_reached` must be reached for sell to trigger, resulting in sells for 2% profit while it should have been a loss. To fix that, we always assume trailing stop offset is not used when `use_custom_stoploss = True`.

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -228,16 +228,18 @@ class Backtesting:
             # Special case: trailing triggers within same candle as trade opened. Assume most
             # pessimistic price movement, which is moving just enough to arm stoploss and
             # immediately going down to stop price.
-            if (sell.sell_type == SellType.TRAILING_STOP_LOSS and trade_dur == 0
-                    and self.strategy.trailing_stop_positive):
-                if self.strategy.trailing_only_offset_is_reached:
+            if sell.sell_type == SellType.TRAILING_STOP_LOSS and trade_dur == 0:
+                if not self.strategy.use_custom_stoploss and self.strategy.trailing_stop and \
+                   self.strategy.trailing_only_offset_is_reached and \
+                   self.strategy.trailing_stop_positive_offset is not None and \
+                   self.strategy.trailing_stop_positive:
                     # Worst case: price reaches stop_positive_offset and dives down.
                     stop_rate = (sell_row[OPEN_IDX] *
                                  (1 + abs(self.strategy.trailing_stop_positive_offset) -
                                  abs(self.strategy.trailing_stop_positive)))
                 else:
                     # Worst case: price ticks tiny bit above open and dives down.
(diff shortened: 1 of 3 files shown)


xmatthias:
i hope you didn't really test with the above settings ... `trailing-stop` and `use_custom_stop` should not go together as they'll conflict for the "closer stoploss" ... so potentially - your stoploss might not even be respected as trailing-stop is closer.

Maybe we need to add a sanity check for that at startup (i intentionally didn't do that at the time - as i wanted to see impacts ...).

rokups:
I tripped exactly on that. In ideal world i think custom_stoploss and any other stoploss behavior should be mutually exclusive. Though this has a chance to break user strategies, even if they are functioning based on hope and prayers.

freqtrade/optimize/backtesting.py
```diff
@@ -228,16 +228,18 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
             # Special case: trailing triggers within same candle as trade opened. Assume most
             # pessimistic price movement, which is moving just enough to arm stoploss and
             # immediately going down to stop price.
-            if (sell.sell_type == SellType.TRAILING_STOP_LOSS and trade_dur == 0
-                    and self.strategy.trailing_stop_positive):
-                if self.strategy.trailing_only_offset_is_reached:
+            if sell.sell_type == SellType.TRAILING_STOP_LOSS and trade_dur == 0:
+                if not self.strategy.use_custom_stoploss and self.strategy.trailing_stop and \
+                   self.strategy.trailing_only_offset_is_reached and \
+                   self.strategy.trailing_stop_positive_offset is not None and \
```
xmatthias: `trailing_stop_positive_offset` can't be None - json schema enforces it to be numeric and between 0 and 1 - so at max, it might not be explicitly set - in which case the default (set in `interface.py`) would apply.
(same for `trailing_only_offset_is_reached`)

I'm also not a particularily big fan of using `\` to split lines (call it bad experience from trailing characters when modifying a script on a server ...) - but would instead wrap the whole `if` into brackets - which allows newlines automatically.

xmatthias (commented):
> Though this has a chance to break user strategies, even if they are functioning based on hope and prayers.

Yes - but in this specific case i don't see that as a problem - as using both IS documented as "don't use" - it's just not enforced at the moment.

I'll probably have a look at that when i find time

xmatthias (approved):
LGTM :+1:

PR #5221, opened 2021-07-01
Add range property to CategoricalParameter.

rokups:
At the moment we can keep a single code path when using IntParameter, but we have to make a special hyperopt case for CategoricalParameter. Range property solves this.


xmatthias:
in general - absolutely (we should also add `.range` to the DecimalParameter ...)
what's missing here is the docs - and maybe a very quick test (i'm always afraid to break functionality unintentionally later on 😆 )

rokups:
Ill take care of DecimalParameter and tests 👍🏻

rokups:
CI failure a random fluke?

xmatthias (approved):
Great! thanks a lot :100: 

LGTM :+1:

PR #5228, opened 2021-07-03
feat(agefilter): add max_days_listed

### LINKED ISSUE #4251: Add max_day_listed  in Age Filter
## Describe the enhancement

Less than 1 year new listed coins usually have better performance. It will be great if we have an option to filter only newly listed pairs with VolumnPairList.

kevinjulian:
## Summary
closes #4251

## Quick changelog
- Add max_days_listed for AgeFilter (default: 7300)

### DIFF
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -87,7 +87,7 @@ When pairs are first listed on an exchange they can suffer huge price drops and 
 in the first few days while the pair goes through its price-discovery period. Bots can often
 be caught out buying before the pair has finished dropping in price.
 
-This filter allows freqtrade to ignore pairs until they have been listed for at least `min_days_listed` days.
+This filter allows freqtrade to ignore pairs until they have been listed for at least `min_days_listed` days and listed before `max_days_listed`.
 
 #### PerformanceFilter
 
@@ -212,7 +212,7 @@ The below example blacklists `BNB/BTC`, uses `VolumePairList` with `20` assets, 
         "number_assets": 20,
         "sort_key": "quoteVolume"
     },
-    {"method": "AgeFilter", "min_days_listed": 10},
+    {"method": "AgeFilter", "min_days_listed": 10, "max_days_listed": 7300},
     {"method": "PrecisionFilter"},
     {"method": "PriceFilter", "low_price_ratio": 0.01},
     {"method": "SpreadFilter", "max_spread_ratio": 0.005},
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -81,7 +81,7 @@ Filtering instances (not the first position in the list) will not apply any cach
 
 #### AgeFilter
 
-Removes pairs that have been listed on the exchange for less than `min_days_listed` days (defaults to `10`).
+Removes pairs that have been listed on the exchange for less than `min_days_listed` days (defaults to `10`) or more than `max_days_listed` days (defaults `None` mean infinity).
 
 When pairs are first listed on an exchange they can suffer huge price drops and volatility
 in the first few days while the pair goes through its price-discovery period. Bots can often
@@ -212,7 +212,7 @@ The below example blacklists `BNB/BTC`, uses `VolumePairList` with `20` assets, 
         "number_assets": 20,
         "sort_key": "quoteVolume"
     },
-    {"method": "AgeFilter", "min_days_listed": 10, "max_days_listed": 7300},
+    {"method": "AgeFilter", "min_days_listed": 10},
     {"method": "PrecisionFilter"},
     {"method": "PriceFilter", "low_price_ratio": 0.01},
     {"method": "SpreadFilter", "max_spread_ratio": 0.005},
--- a/freqtrade/plugins/pairlist/AgeFilter.py
+++ b/freqtrade/plugins/pairlist/AgeFilter.py
@@ -27,6 +27,7 @@ class AgeFilter(IPairList):
         super().__init__(exchange, pairlistmanager, config, pairlistconfig, pairlist_pos)
 
         self._min_days_listed = pairlistconfig.get('min_days_listed', 10)
+        self._max_days_listed = pairlistconfig.get('max_days_listed', None)
 
         if self._min_days_listed < 1:
             raise OperationalException("AgeFilter requires min_days_listed to be >= 1")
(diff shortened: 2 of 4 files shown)


docs/includes/pairlists.md
```diff
@@ -77,17 +77,17 @@ Filtering instances (not the first position in the list) will not apply any cach
 ```
 
 !!! Note
-    `VolumePairList` does not support backtesting mode.
+`VolumePairList` does not support backtesting mode.
```
xmatthias: Careful with these "Unintended" changes (probably done automatically by some IDE?).
This change (and all others that relate to lines after !!!) is wrong - the indent of 4 spaces is a must, otherwise rendering in the docs will break.

I'd suggest to revert all changes that haven't been done intentionally - as i think you only really changed 2 or 3 lines in this file).

freqtrade/plugins/pairlist/AgeFilter.py
```diff
@@ -63,7 +68,7 @@ def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
 
         since_ms = int(arrow.utcnow()
                        .floor('day')
-                       .shift(days=-self._min_days_listed - 1)
+                       .shift(days=-self._max_days_listed - 1)
```
xmatthias: with this change, you'll "attempt" to download history since 7300 days by default.
For each pair - that's 7-14 requests (average answers are between 500 and 1000 candles, depending on the exchange).

This will make this filter terribly slow - especially considering that it'll run on every startup for every pair.
Starting this on kraken would probably not work (well, it would work technically, but not within my patience realm) - as kraken can only take ~1 request every 3 seconds).

kevinjulian: should we make it dynamic? if we put `max_days_listed` on the config, then use `max_days_listed` else use `min_days_listed`
what do you think?

docs/includes/pairlists.md
kevinjulian: noted, my bad 😅

freqtrade/plugins/pairlist/AgeFilter.py
xmatthias: it's not so much the amount - but the way it works that's going to be problematic if it starts doing multiple exchange calls

with `min_days_listed` - I'd expect people to usually not go further back than ~500 days (which is 1 api call).
with `max_days_listed` this might be different.

If you notice above (in `__init__`) - there's an exception/error when `min_days_listed` is > one exchange call.
Realistically, we should probably have the same "blocker" for `max_days` to avoid having AgeFilter take forever when it's starting

kevinjulian: is it enough when we add same "blocker" like `min_days_listed` to `max_days_listed` ?
if the issue is on the api call then what should we do ? I was searching binance api doc to get "listed date", but I got nothing. I am not sure it will be in ccxt

xmatthias: no, the only way (i found) to get this is to download candles from "right before" the desired date and check if there's data.

for binance - you can request data from 4 years ago - if the pair was listed in 2020, you'll simply get that data - so an alternative would be to download 1 call from the exchange and look at the dates instead of dataframe length.

The problem with the way AgeFilter is working currently is that it'd download all data since 2020 - which would not be necessary for AgeFilter.
However - the exchange class currently has no possibility to fetch only 1 call (or define an end-date) - so the call we use will always download from "since" to today.

For the moment, i'd say limit to 1 call - and keep it as "further enhancement" in changing this? (unless  you feel for adding aditional methods to the exchange class, too)

kevinjulian: yes I agree with you, for now limit to 1 call and we can to keep it as "further enhancement"

freqtrade/plugins/pairlist/AgeFilter.py
```diff
@@ -49,7 +56,9 @@ def short_desc(self) -> str:
         Short whitelist method description - used for startup-messages
         """
         return (f"{self.name} - Filtering pairs with age less than "
-                f"{self._min_days_listed} {plural(self._min_days_listed, 'day')}.")
+                f"{self._min_days_listed} {plural(self._min_days_listed, 'day')}"
+                " or more than "
+                f"{self._max_days_listed} {plural(self._max_days_listed, 'day')}")
```
xmatthias: I think this message will need to be composed dynamically.

if `max_days_listed` is not set - it'll result in "or more than None days" (or an exception ...).

freqtrade/plugins/pairlist/AgeFilter.py
```diff
@@ -86,14 +98,17 @@ def _validate_pair_loc(self, pair: str, daily_candles: Optional[DataFrame]) -> b
             return True
 
         if daily_candles is not None:
-            if len(daily_candles) >= self._min_days_listed:
+            if len(daily_candles) >= self._min_days_listed and \
+                    len(daily_candles) <= self._max_days_listed:
```
xmatthias: You will want to be careful with this - as `max_days_listed` can be None - this check will fail with `TypeError: '<=' not supported between instances of 'int' and 'NoneType'` should the first condition ever evaluate to False.

freqtrade/plugins/pairlist/AgeFilter.py
```diff
@@ -86,14 +98,17 @@ def _validate_pair_loc(self, pair: str, daily_candles: Optional[DataFrame]) -> b
...
                 # Add to cache, store the time we last checked this symbol
                 self._symbolsChecked[pair] = int(arrow.utcnow().float_timestamp) * 1000
                 return True
             else:
                 self.log_once(f"Removed {pair} from whitelist, because age "
                               f"{len(daily_candles)} is less than {self._min_days_listed} "
-                              f"{plural(self._min_days_listed, 'day')}", logger.info)
+                              f"{plural(self._min_days_listed, 'day')} or more than "
+                              f"{self._max_days_listed} {plural(self._max_days_listed, 'day')}",
+                              logger.info)
```
xmatthias: Same as above, basically - `because age 3 is less than 10 days or more than None days` does not look that nice ...

xmatthias (changes requested):
Unfortunately - tests don't fully cover this new behaviour (even if coverage might seem complete) - and it can result in crashes with TypeErrors...

xmatthias (approved):
LGTM :+1: 

Thanks a lot for your work and involvement! :100:

PR #5229, opened 2021-07-03
compact low balance currencies

### LINKED ISSUE #5217: Telegram: hide/compact the list of currencies with a low balance (below dust level)
Hi,

This is my first contribution!

When using the /balance command with the Telegram bot, having a lot of low-amount currencies will result in a insanely long list:

Insade amount of dust currencies

It happens quickly on Binance depending on your config, I don't know if it's the same for other exchanges.

Wouldn't it be simpler to regroup all these currencies in a line like this:
"72 Other currencies:
 Est. USDT: 133.967"

Or we could make an option to choose if we want to keep the list as it is for now, or if we want to compact this.

What do you think?

It looks like a minor modification, if you're interested by this one but it's not on the priority list I could learn how to contribute and make a PR.

Thank you!

kevinjulian:
## Summary
closes #5217 

## What's new?
image
image

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -11,6 +11,7 @@ from html import escape
 from itertools import chain
 from math import isnan
 from typing import Any, Callable, Dict, List, Optional, Union
+
 import arrow
 from tabulate import tabulate
 from telegram import (CallbackQuery, InlineKeyboardButton, InlineKeyboardMarkup, KeyboardButton,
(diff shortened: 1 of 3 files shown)


freqtrade/rpc/telegram.py
```diff
@@ -618,6 +621,12 @@ def _balance(self, update: Update, context: CallbackContext) -> None:
                 else:
                     output += curr_output
 
+            if total_dust_balance > 0:
+                output += (
+                    f"*{total_dust_currencies} Other Currencies:*\n"
```
itsJarrett: Perhaps mention that these currencies are below the `balance_dust_level` somehow via the output?

kevinjulian: *72 Other currencies (<1 USDT):
Est. USDT: 133.967*

like this, ok?

itsJarrett: looks good!

xmatthias (approved):
LGTM :+1: 

PR #5231, opened 2021-07-04
Range based VolumeFilter

nightshift2k:
## Summary
Allow VolumeFilter to extend the Volume beyond the exchanges 24h volume to a custom range in time and candlesize.

## Quick changelog

- Adds `lookback_days`, `lookback_days`, `lookback_timeframe` and `lookback_period` to pairlist config
- If  `lookback_days` or `lookback_timeframe` and `lookback_period` are specified fetches the specified amount of candles and calculates the volume based on typical price ((h+l+c)/3) in the given range
- Overwrites the `ticker['quoteVolume']` with new range based values if activated
- Behaves the usual way, if above given params are not specified

## What's new?
Improvement to VolumeFilter to calculate based on candlesize and range to allow more flexible volume pairlist.

## TODO
- *IN  PROGRESS*: Write tests to cover new cases from the new configuration possibilites
- ~~Enhance documentation for new configuration possibilities~~

### DIFF
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -63,7 +63,7 @@ The `refresh_period` setting allows to define the period (in seconds), at which 
 The pairlist cache (`refresh_period`) on `VolumePairList` is only applicable to generating pairlists.
 Filtering instances (not the first position in the list) will not apply any cache and will always use up-to-date data.
 
-`VolumePairList` is based on the ticker data from exchange, as reported by the ccxt library:
+`VolumePairList` is per default based on the ticker data from exchange, as reported by the ccxt library:
 
 * The `quoteVolume` is the amount of quote (stake) currency traded (bought or sold) in last 24 hours.
 
@@ -73,6 +73,35 @@ Filtering instances (not the first position in the list) will not apply any cach
         "number_assets": 20,
         "sort_key": "quoteVolume",
         "refresh_period": 1800
+}],
+```
+
+`VolumePairList` can also operate in an advanced mode to build volume over a given timerange of specified candle size. It utilizes exchange historical candle data, builds a typical price (calculated by (open+high+low)/3), and multiplies it with every candle data's volume. The sum is the `quoteVolume` over the given range. This allows different scenarios, for a  more smoothened volume, when using longer ranges with larger candle sizes, or the opposite when using a short range with small candles.
+
+For convenience `lookback_days` can be specified, which will imply that 1d candles will be used for the lookback. In the example below the pairlist would be created based on the last 7 days:
+
+```json
+"pairlists": [{
+        "method": "VolumePairList",
+        "number_assets": 20,
+        "sort_key": "quoteVolume",
+        "refresh_period": 86400,
+        "lookback_days": 7
+}],
+```
+!!! Warning "Range look back and refresh period"
+    When used in conjuction with `lookback_days` and `lookback_timeframe` the `refresh_period` can not be smaller than the candle size in seconds. As this will result in unnecessary requests to the exchanges API.
+
+More sophisticated approach can be used, by using `lookback_timeframe` for candle size and `lookback_period` which specifies the amount of candles. This example will build the volume pairs based on a rolling period of 3 days of 1h candles:
+
+```json
+"pairlists": [{
+        "method": "VolumePairList",
+        "number_assets": 20,
+        "sort_key": "quoteVolume",
+        "refresh_period": 3600,
+        "lookback_timeframe": "1h",
+        "lookback_timeframe": 72
 }],
 ```
 
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -78,7 +78,7 @@ Filtering instances (not the first position in the list) will not apply any cach
 ],
 ```
 
-`VolumePairList` can also operate in an advanced mode to build volume over a given timerange of specified candle size. It utilizes exchange historical candle data, builds a typical price (calculated by (open+high+low)/3), and multiplies it with every candle data's volume. The sum is the `quoteVolume` over the given range. This allows different scenarios, for a  more smoothened volume, when using longer ranges with larger candle sizes, or the opposite when using a short range with small candles.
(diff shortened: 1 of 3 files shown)


freqtrade/plugins/pairlist/VolumePairList.py
```diff
@@ -103,6 +141,54 @@ def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
         # Use the incoming pairlist.
         filtered_tickers = [v for k, v in tickers.items() if k in pairlist]
 
+        # get lookback period in ms, for exchange ohlcv fetch
+        if self._use_range:
+            since_ms = int(arrow.utcnow()
+                           .floor('minute')
+                           .shift(minutes=-(self._lookback_period * self._tf_in_min)
+                                  - self._tf_in_min)
+                           .float_timestamp) * 1000
```
xmatthias: why not use int_timestamp instead of explicitly parsing to int ... ?

freqtrade/plugins/pairlist/VolumePairList.py
```diff
@@ -36,6 +39,35 @@ def __init__(self, exchange, pairlistmanager,
...
+
+        # wether to use range lookback or not
+        self._use_range = (self._tf_in_min > 0) & (self._lookback_period > 0)
+
+        if self._use_range & (self._refresh_period < self._tf_in_sec):
+            raise OperationalException(
+                f'Refresh period of {self._refresh_period} seconds is smaller than one '
+                f'timeframe of {self._lookback_timeframe}. Please adjust refresh_period '
+                f'to at least {self._tf_in_sec} and restart the bot.'
+            )
```
xmatthias, line 70: You may also wanna top-cap this at the exchange candle limit for one call (look at `AgeFilter` for how to do this).

`refresh_latest_ohlcv` will only return 1 exchange call - so you won't have 2000 5m candles available.

xmatthias (commented):
I'd say looks great already! :+1:

freqtrade/plugins/pairlist/VolumePairList.py
nightshift2k: Honestly, I stole that part exactly from RangeStabilityFilter.py so, yeah sure this makes more sense, but maybe we should also look at the other filters, to improve on that? ;)

xmatthias: for sure :laughing: 
it's not like i notice everything (or the same things) every day. While i clearly should - that's now how humans work ... (at least i don't).

freqtrade/plugins/pairlist/VolumePairList.py
nightshift2k: Not sure if I understand this correctly, but this is what should happen in line 84ff, where we trigger an `OperationalException` when the `lookback_period` exceedes `ohlcv_candle_limit` for given candlesize?!

Or maybe I misunderstood it.

xmatthias: maybe i wrote this comment before getting to line 84 ... :laughing: 

sorry ...

nightshift2k:
I will definitely need assistance further down the road to implement the appropriate tests for the new parameters. Especially how to cover all possible cases, where the historical candle data will be needed to perform the lookback, hope somone more skilled than me, can assist with that or at least hint me.

freqtrade/plugins/pairlist/VolumePairList.py
```diff
@@ -103,6 +141,53 @@ def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
...
+                # in case of candle data calculate typical price and quoteVolume for candle
+                if pair_candles is not None and not pair_candles.empty:
+                    pair_candles['typical_price'] = (pair_candles['high'] + pair_candles['low']
+                                                     + pair_candles['close']) / 3
+                    pair_candles['quoteVolume'] = (
+                        pair_candles['volume'] * pair_candles['typical_price']
+                    )
+
+                    # replace quoteVolume with range quoteVolume sum calculated above
+                    filtered_tickers[i]['quoteVolume'] = pair_candles['quoteVolume'].sum()
```
xmatthias: Wouldn't you want to use something like `pair_candles['quoteVolume'].rolling(self._lookback_period).sum()` - and use the last value of that result?

while `since_ms` is used to specify since when we'd like to have history - i'm not certain it'll work reliably for all exchanges (depends a bit on the exchange api functioning).
if you provide a since_ms very close to the current date (closer than `now - minimum candles`  - this might result in using 500 candles even though you only specified 20 ...

nightshift2k: fair point! didn't have in mind that a very tight delta between since_ms and now can result in fetching the maximum candle amount.

xmatthias: doesn't even have to e very tight - some exchanges provide you a minimum of 100 candles, even if you only wanted 20 ...

docs/includes/pairlists.md
```diff
@@ -76,6 +76,35 @@ Filtering instances (not the first position in the list) will not apply any cach
...
+More sophisticated approach can be used, by using `lookback_timeframe` for candle size and `lookback_period` which specifies the amount of candles. This example will build the volume pairs based on a rolling period of 3 days of 1h candles:
+
+```json
+"pairlists": [{
+        "method": "VolumePairList",
+        "number_assets": 20,
+        "sort_key": "quoteVolume",
+        "refresh_period": 3600,
+        "lookback_timeframe": "1h",
+        "lookback_timeframe": 72
```
xmatthias: something is not right here ... 
you can't really configure lookback_timeframe twice ;)

xmatthias (commented):
I think we'll want a warning on the docs about performance of this.

I've used this with number_assets: 10 (only that filter) - and for USDT, it'll need to download data for 279 pairs - which obviously takes quite some time.

Maybe even proposing a "pre-filter" to limit the number of pairs to top 100 volume initially, and use this as a further filter (based on the assumption that the current top100 pairs will probably contain the top 50 pairs based on volume in another timeframe).

freqtrade/plugins/pairlist/VolumePairList.py
nightshift2k: yeah, building a total sum with a varying amount of candles depending on how the exchange behaves was stupid anyway, so this is definitely a much more safer approach 👌

docs/includes/pairlists.md
nightshift2k: 🤦‍♂️ on it ....

xmatthias (approved):
Great! :100: 

LGTM :+1:

PR #5245, opened 2021-07-07
Range stability filter max

sauces1313:
>Thank you for sending your pull request. But first, have you included
>unit tests, and is your code PEP8 conformant? [More details](https://github.com/freqtrade/freqtrade/blob/develop
>/CONTRIBUTING.md)

Yes, ran unit tests, ran flake8, checked type-hints and imports, updated "Plugins" documentation. My apologies if I've missed something (this is my first proper PR).

## Summary

Add additional pairlist plugin.

## Quick changelog

- n/a (?)

## What's new?

Adds the RangeStabilityFilterMax pairlist plugin. Essentially the same as RangeStabilityFilter except allows the user to define a `max_rate_of_change`.


xmatthias (changes requested):
I'd not do this as a seperate filter, but add the max option to the existing `RangeStabililtyFilter`.

That way, we/users can easily configure a bracket (only pairs within this range). Using a seperate filter has little sense in my opinion - and will be a ton of code duplication for no good reason.

sauces1313:
> I'd not do this as a seperate filter, but add the max option to the existing `RangeStabililtyFilter`.
> 
> That way, we/users can easily configure a bracket (only pairs within this range). Using a seperate filter has little sense in my opinion - and will be a ton of code duplication for no good reason.

Yes, that is a much better idea. I'll see if I can pull that off :)
Thanks!

sauces1313 (commented):
I believe the requested changes are complete.

freqtrade/plugins/pairlist/rangestabilityfilter.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, exchange, pairlistmanager,
 
         self._days = pairlistconfig.get('lookback_days', 10)
         self._min_rate_of_change = pairlistconfig.get('min_rate_of_change', 0.01)
+        self._max_rate_of_change = pairlistconfig.get('max_rate_of_change', 0.99)
```
xmatthias: i think the default for this should be None (obviously will need checking below before using this value).

`max_rate_of_change` is new - and 0.99 is pretty tight considering how crypto-markets work...

n the top 10 USDT volume pairs on binance, i get AXS/USDT with 2.7, ALICE/USDT with 3.0 and C98/USDT with a value of 38 at the moment (ok, that last one has been listed 5 days ago or so, but the others ...).


while there can be reasons to exclude these - it should be intentional and wouldbe surprising (and undesired -> a somehow breaking change) to have them excluded automatically by an update - as then it'd not be intentional.

freqtrade/plugins/pairlist/rangestabilityfilter.py
```diff
@@ -104,6 +106,14 @@ def _validate_pair_loc(self, pair: str, daily_candles: Optional[DataFrame]) -> b
                               f"which is below the threshold of {self._min_rate_of_change}.",
                               logger.info)
                 result = False
+            if pct_change <= self._max_rate_of_change:
```
xmatthias: as mentioned above - when using a default of None - you'll need to modify this to first (!) check if max_rate_of_change is None.

xmatthias (changes requested):
Sorry for the late review - been kinda busy the last week.

Anyway, i've played with it a bit and found one thing we'll need to slightly modify.

freqtrade/plugins/pairlist/rangestabilityfilter.py
sauces1313: Agreed, and thanks for the suggestion. My intention was to set it so high that none would be excluded by default. So the 0.99 should've been something more like 99.0 but a value of `None` should suffice. Logically, I'm not sure if either 99.0 or None is a better option than the other. If one is better than the other, do enlighten me :)

freqtrade/plugins/pairlist/rangestabilityfilter.py
sauces1313: For clarification (cause I'm a complete amateur lol), by "(!)" do you mean the code would read more like `if self.max_rate_of_change != None`...?

sauces1313: Reading your comment again, perhaps you were instead emphasizing the check should be done first?

freqtrade/plugins/pairlist/rangestabilityfilter.py
xmatthias: using None explicitly unsets it.

we won't know what markets people will use freqtrade in (and also not how markets will move in the future).
with a sufficient lookback on cheap, or newly listed pairs - you can easily get values of > 99 (i got a value of 38 yesterday).
As such - i'd explicitly unset this - and only if users explicitly set the value - it'll apply. Otherwise we'll introduce magic constants which we are not sure will hold in the future.

freqtrade/plugins/pairlist/rangestabilityfilter.py
xmatthias: I think you'll need to wrap this whole section into a `if self._max_rate_of_change:`.

you cannot `=` (or `!=`) compare `None` - you'd have to use `x is None` or `x is not None` - but it'll evaluate to falsy without this explicit check - so simply checking this as if it where a boolean is usually sufficient - unless 0.0 is a valid value (which it isn't in this case- as that would mean a pair with 0 movement).

sauces1313: > I think you'll need to wrap this whole section into a `if self._max_rate_of_change:`.

Agreed.
 
> you cannot `=` (or `!=`) compare `None` - you'd have to use `x is None` or `x is not None` - but it'll evaluate to falsy without this explicit check - so simply checking this as if it where a boolean is usually sufficient - unless 0.0 is a valid value (which it isn't in this case- as that would mean a pair with 0 movement).

Lol, duh... I actually knew this at some point, and have clearly forgotten it in favor of linter dependence. Thanks for the reminder :)

I have some time so I'm gonna try to have this done today.

freqtrade/plugins/pairlist/rangestabilityfilter.py
sauces1313: > using None explicitly unsets it.
...
> As such - i'd explicitly unset this - and only if users explicitly set the value - it'll apply. Otherwise we'll introduce magic constants which we are not sure will hold in the future.

I understand... I appreciate you "holding my hand" through some of these things. The strategy I'm working on is dependent on this enhancement (have been testing with my original quick and dirty filter), so when that cash starts rolling in I'll be sure to hit that support button :)

xmatthias (approved):
I've pushed a small change to better handle the description in case no max-rate is set and to test both  min and max cases properly.

LGTM 👍

sauces1313:
> I've pushed a small change to better handle the description in case no max-rate is set and to test both min and max cases properly.

Excellent, much appreciated!

PR #5246, opened 2021-07-07
new filter OffsetFilter for offsetting incoming pairlists

### LINKED ISSUE #5241: new PairFilter for offsetting pairlists
## Describe the enhancement

For some strategies it might make sense to split a larger pairlist on two bots, to speed up indicator and/or buy/sell calculation, therefor I'd like to propose a new filter OffsetFilter, which will offset the pairlist so a given amount of pairs can be skipped. For example a compute heavy strat could be split running on 100 pairs could be split into two bots one running on the first 50 pairs, the other running with an offset of 50 on pair 51-100. Or one could skip a certain amount of the usual topXX coins.

nightshift2k:
## Summary
Introduce a new pairlist filter `OffsetFilter` for ofsetting an incomming pairlist.

closes #5241 

## Quick changelog
- created `OffsetFilter.py`
- enhanced and added tests to `test_pairlist.py`
- updated documentation in`pairlists.md`

## What's new?
Added `OffsetFilter` which can offset the pairlist by specifying `offset` larger than zero. 

Example:
```json
"pairlists": [{
        "method": "OffsetFilter",
        "offset": 10
}],
```

### DIFF
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -23,6 +23,7 @@ You may also use something like `.*DOWN/BTC` or `.*UP/BTC` to exclude leveraged 
 * [`StaticPairList`](#static-pair-list) (default, if not configured differently)
 * [`VolumePairList`](#volume-pair-list)
 * [`AgeFilter`](#agefilter)
+* [`OffsetFilter`](#offsetfilter)
 * [`PerformanceFilter`](#performancefilter)
 * [`PrecisionFilter`](#precisionfilter)
 * [`PriceFilter`](#pricefilter)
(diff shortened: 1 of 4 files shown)


docs/includes/pairlists.md
```diff
@@ -89,6 +90,30 @@ be caught out buying before the pair has finished dropping in price.
...
+"pairlists": [{
+        "method": "OffsetFilter",
+        "offset": 10
+}],
+```
+
+!!! Warning
+    When `OffsetFilter` is used to split a larger pairlist among multiple bots in combination with `VolumeFilter` 
+    it can not be guaranteed that pairs won't overlap due to slightly different refresh intervals for the
+    `VolumeFilter`
```
xmatthias:
```suggestion
!!! Warning
    When `OffsetFilter` is used to split a larger pairlist among multiple bots in combination with `VolumeFilter` 
    it can not be guaranteed that pairs won't overlap due to slightly different refresh intervals for the
    `VolumeFilter`.
```

docs/includes/pairlists.md
```diff
@@ -89,6 +90,30 @@ be caught out buying before the pair has finished dropping in price.
...
+As an example it can be used in conjunction with `VolumeFilter` to remove the top X volume pairs. Or to split
+a larger pairlist on two bot instances.
+
+Example to remove the first 10 pairs from the pairlist:
+
+```json
+"pairlists": [{
+        "method": "OffsetFilter",
+        "offset": 10
+}],
```
xmatthias:
```suggestion
"pairlists": [
    {
        "method": "OffsetFilter",
        "offset": 10
    }
],
```
Putting the brackets at new lines makes it easier to copy/paste the sample (you won't / can't use only Offsetfilter - so copying the whole sample is not possible.

xmatthias (commented):
2 (very small) nitpicks to improve readability in the docs ... rest looks great :+1:

xmatthias (approved):
LGTM :+1:

PR #5252, opened 2021-07-08
Fix Agefilter cannot appear on startup messages

kevinjulian:
## Summary
Sorry, this is related to PR #5228
Age filter will be empty string when there is no max_age_listed, fix with this PR
image

to prevent this, should we test startup messages ?


xmatthias (approved):
Sometimes it's the small things :laughing: 

LGTM :+1:

PR #5265, opened 2021-07-12
Censor DB password when outputting to logs

anasyusef:
## Summary
Censor password from logs when providing a DB connection URI
E.g:
From: `postgresql+psycopg2://scott123:scott123@host/dbname`
Outputs the following to logs: `postgresql+psycopg2://scott123:*****@host/dbname`

## Quick changelog
- Apply censoring to logs on URI that contains a password

### DIFF
--- a/freqtrade/commands/list_commands.py
+++ b/freqtrade/commands/list_commands.py
@@ -14,7 +14,7 @@ from freqtrade.constants import USERPATH_HYPEROPTS, USERPATH_STRATEGIES
 from freqtrade.enums import RunMode
 from freqtrade.exceptions import OperationalException
 from freqtrade.exchange import market_is_active, validate_exchanges
-from freqtrade.misc import plural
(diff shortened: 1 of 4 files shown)


xmatthias (changes requested):
In principle i'd say - looks good.

However, please remove the .terraform submodule which seems to have been commited by mistake (best via rebase so it's gone from history, too).

anasyusef:
Yep. It was indeed a mistake. I've removed the submodule from the branch. It should hopefully be good now

xmatthias (approved):
seems like the only case where the db-url is logged with password is when it's wrong - but i guess then it's ok / acceptable - as it might show you the error too.

Great job - LGTM :+1:

anasyusef:
Thank you! I was thinking of adding validation for that edge case. But then, I guess SQLAlchemy would handle that and raise the error as It wouldn't be able to establish a connection.

PR #5275, opened 2021-07-16
Example json filename extensions

samgermain:
## Summary
Changes example config files from `config_exchange.json.example` to `config_examples/config_exchange.example.json`
Now these files register as json files and are formatted and colored/highlighted accordingly in IDE's

Each of the example configs was specifically added to `.gitignore` exceptions instead of adding the whole folder `freqtrade/config_examples`, to prevent people from accidentally pushing any api keys to Github


xmatthias (approved):
nice one! thanks for noticing and taking the time to improve this 💯 

LGTM 👍

PR #5278, opened 2021-07-17
Margin config

samgermain:
## Summary
Configuration options for margin trading

## Quick changelog

Still a draft, mostly filled with TODO's

## What's new?
Adds configuration options for margin trading


freqtrade/commands/cli_options.py
```diff
@@ -548,4 +548,20 @@ def __init__(self, *args, **kwargs):
...
+    # "leverage_min": Arg(
+    #     '--leverage-min',
+    #     help='The minimum leverage to trade at if the desired leverage is not available',
+    #     action=''
+    # ),
+    # "shorts_enabled": Arg(
+    #     '--shorts-enabled',
+    #     help='Enables short trading',
+    #     action=''
+    # )
```
xmatthias, line 566: please don't add these as cli arguments (at least not for now).

Leverage (+ short support) should come from the strategy - as it's a strategy decision IF it works with shorts or not - not a configuration / cli option.

Also - for leverage itself - that's not a configuration option either - as different pairs allow different levels of leverage - so it'll probably be a strategy-level callback - with the possible leverage values provided (look at kraken - some pairs have 2,3,4,5 - others only 2,3 , ....

samgermain: But the user has to decide how much leverage they want to use. I don't think it should be either trade with the max available leverage or don't trade with leverage at all and it would be wrong to assume that the user wants to trade at 5x leverage if 5x leverage is available. 

I'm also wanting to implement it so that if a pair cannot trade on leverage, then I don't want to bother with that pair

I think the best way to go about it would be to have the user provide their desired leverage in the configuration as a range
*  there would be config.min_leverage
* and config.leverage (or config.max_leverage)

If the maximum available leverage for a pair is lower than config.min_leverage:
    then bot won't trade on that pair
    
If the maximum available leverage for a pair is ≥ config.min_leverage:
    then the bot with trade the pair at the maximum allowable leverage which is ≤ config.max_leverage
    
------------

This is especially true if futures get's implemented, freqtrade definately shouldn't trade at 125x leverage just because 125x leverage is available

xmatthias: i'm not saying "use what's available" - but implement it as a callback (so the strategy get's "asked" what leverage to use) - similar to `custom_sell` (or so).

Having this in the configuration will make it static (applies to all pairs) - which will be highly limiting and inflexible ( think - use  3x leverage for ETH - but 5x on XRP because the indicators there show better chances).

samgermain: @xmatthias That seems a lot more complicated than what I was going for. Is there a way to do both? My plan was to write it just so that it would trade like normal, but apply the maximum desired leverage on every trade. I feel like there's a use case for both. I think I want leave the option open to define a static leverage and the user could also use the strategy implemented leverage if they want

What if leverage was in the config, and then there was an option like "strategy_implemented" or something?

samgermain: Oh wait, the strategy could just always return a static value then, so the config wouldn't be needed, and then the strategy could just have something like if max_leverage >= 5 or something

@Axel-CH Take a look at this conversation

xmatthias: That's my point :) 
the problem with "static" is that it'll not work for all pairs (at least not on non-binance exchange) - as (max) leverage depends on the pair.

PR #5279, opened 2021-07-17
Honor skip_pair_validation setting when downloading pairs.

rokups:
I tried downloading dead pairs as suggested on discord and could not. Turns out there was no way to disable pair validation after all. This fixes the issue.


xmatthias (approved):
Nice catch :+1:

PR #5281, opened 2021-07-17
A decorator for easy creation of informative pairs

rokups:
```py
    @informative(timeframe='1h', asset='BTC', fmt='{asset}_{name}_{fimeframe}')
    def populate_indicators_1h(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        dataframe['rsi'] = ta.RSI(dataframe, timeperiod=14)
        return dataframe

        super().populate_indicators(dataframe, metadata)
        dataframe['rsi'] = ta.RSI(dataframe, timeperiod=14)
        dataframe['rsi_less'] = dataframe['rsi'] < dataframe['btc_rsi_1h']
        return dataframe
```
This is all it takes to create an informative pair now. `asset` and `fmt` parameters are totally optional with sane defaults. not specifying `asset` will produce an informative timeframe for current pair, not specifying `fmt` will use `'{asset}_{name}_{fimeframe}'` when asset is specified and `'{name}_{fimeframe}'` when asset is not specified.

There is something to think about. Do we need a possibility to use informative pair with a different stake currency? If so - `asset` parameter should be reworked to support specifying a full pair.

Also it is possible to trip on specifying multiple informative pair functions using different pairs or different timeframes while using same function name. This results in only last function having any effect as it overwrites previous instances sharing the name. I have no idea how to detect this case and produce a failure and currently strategy would throw an exception when user tries to use indicators from overwritten functions.


freqtrade/strategy/interface.py
```diff
@@ -132,14 +132,18 @@ def __init__(self, config: dict) -> None:
...
         Populate indicators that will be used in the Buy and Sell strategy
         :param dataframe: DataFrame with data from the exchange
         :param metadata: Additional information, like the currently traded pair
         :return: a Dataframe with all mandatory indicators for the strategies
         """
+        for fn in [getattr(self.__class__, func) for func in dir(self.__class__)
+                   if callable(getattr(self.__class__, func))]:
+            if not getattr(fn, '_is_informative', False):
+                continue
+            dataframe = fn(self, dataframe, metadata)
```
xmatthias: it's not clear to me what you're trying to do here - but it's definitely not a good idea to do it this way.

It assumes that you'll not have a `populate_indicators` method - which will only be confusing for existing users, or new users using an available method.
This is an abstract method (for a reason, really) - and the current documentation / samples / available strategies do not call the parent method - so the "working" of this will be fluky at best.

If you think this is a neccessity (i'm not sure - but it's not clear to me what this does just yet) - then keep `populate_indicators` as abstract method - and write a wrapper (or use the existing wrapper in `advise_indicators`) for this task.

freqtrade/strategy/interface.py
```diff
@@ -332,7 +336,22 @@ def informative_pairs(self) -> ListPairsWithTimeframes:
...
+                   if callable(getattr(self.__class__, func))]:
+            if not getattr(fn, '_is_informative', False):
+                continue
+            timeframe = getattr(fn, '_timeframe', None)
+            asset = getattr(fn, '_asset', None)
+            assert asset is not None
+            assert timeframe is not None
+            if asset:
+                informative_pairs.append(
+                    (f'{asset.upper()}/{self.config["stake_currency"]}', timeframe))
```
xmatthias: we'll for sure need to support "non-stake" pairs too.

assume you're trading in BTC markets - you might want to get a "market baseline" by looking at BTC/USDT (or BTC/BUSD) - neither of which will work with this method.

xmatthias:
>  This results in only last function having any effect as it overwrites previous instances sharing the name. I have no idea how to detect this case and produce a failure and currently strategy would throw an exception when user tries to use indicators from overwritten functions.

you can't really - it's simply how python works (the whole class is parsed into a dictionary - which has only one "name key".

I'm pretty much wondering how you expect this to work - this seems to have way more limitations and gotchas and "traps where people will fail" than the current method (duplicate naming, ...).

What if i want to look at both BTC and ETH and "current pair" in another timeframe informative pairs?
How'd that look like ... with this syntax now?

freqtrade/strategy/interface.py
rokups: > it's not clear to me what you're trying to do here

This code iterates all callables of the class and calls ones tagged with `_is_informative = True`.

> It assumes that you'll not have a populate_indicators method - which will only be confusing for existing users, or new users using an available method.

That is not true. `populate_indicators` is still required just like it used to be.

> This is an abstract method (for a reason, really)

You have a point about loss of `@abstract` decorator. Can we keep it to mandate implementation of such method, while also calling this method via `super()`?

> current documentation / samples / available strategies do not call the parent method

They do not. And it is not required, unless decorator is used. Maybe we could side-step this requirement by calling this machinery before invoking `populate_indicators`. It would allow keeping `@abstract` too.

> If you think this is a neccessity (i'm not sure - but it's not clear to me what this does just yet) - then keep populate_indicators as abstract method - and write a wrapper (or use the existing wrapper in advise_indicators) for this task.

Yes, most likely best course of action 👍🏻

rokups:
> What if i want to look at both BTC and ETH and "current pair" in another timeframe informative pairs?
How'd that look like ... with this syntax now?

I would suggest defining BTC and ETH informative pairs using a decorator, and using old method of constructing informative pair dataframe in main `populate_indicators()`. I aim for system to be simple by default, but flexible when needed, so this decorator will cover 95% of most common cases while allowing us to do things the way we do them without this change when necessary.

xmatthias:
In it's current form, mixing existing (`informative_pairs()` and decorator style will not work - because the return value of "informative_pairs" will not include the mix of both. While not blocking - it's something to keep in mind (and something you'll not notice in backtesting, as that will not care about the return-value of informative_pairs.

Best look at `advise_indicators` (that's the only thing that calls `populate_indicators()` - at least that's how it should be) to ingest additional logic...

freqtrade/strategy/interface.py
```diff
@@ -742,6 +762,16 @@ def advise_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
...
+
+        # call populate_indicators_Nm() which were tagged with @informative decorator.
+        for attr_name in dir(self.__class__):
+            cls_attr = getattr(self.__class__, attr_name)
+            if not callable(cls_attr):
+                continue
+            if not getattr(cls_attr, '_is_informative', False):
+                continue
+            dataframe = cls_attr(self, dataframe, metadata)
+
```
xmatthias: that's gonna have the same problem than the `populate_indicators` we looked at.

If you need to do "additional" stuff in methods that users implement by default (this has been in all templates for years), then you'll need to create a wrapper for this where the additional stuff is done - and then (or before?) calls the existing method.

In this case, no such wrapper exists - but it should be pretty simple to write that ... :) 

consider the most likely scenario for this enhancement to avoid having it break stuff
* i have a working strategy (take sample_strategy) - with 1 additional pair added to informative_pairs
* I see this change - and wanna use it for the 2nd pair i need
* I just add a decorated method as shown / documented - without touching my already existing implementation for now


**edit** Please accept this comment - but it's in the wrong spot and should be in `informative_pairs()` above ...
sorry

rokups: Good point 👍🏻 i ammended commit, creating internal `gather_informative_pairs()` which does extra logic needed. Now we can mix and match decorators with old way of doing things 🚀

xmatthias: i knew you'd understand me even if i comment it on the wrong lines :laughing:

rokups:
Found a critical bug which messed up dataframes. Apparently we can not modify variables inherited from upper scopes of lambdas, or things go weird 😅 

We should really consider trimming pair dataframes. Informative pairs calculate indicators over entire range available, so in my case its for data since 2017. It is not that bad for 1h timeframe, but for 5m informative it gets stupid real fast.

xmatthias:
> Found a critical bug which messed up dataframes. Apparently we can not modify variables inherited from upper scopes of lambdas, or things go weird 
> 
> We should really consider trimming pair dataframes. Informative pairs calculate indicators over entire range available, so in my case its for data since 2017. It is not that bad for 1h timeframe, but for 5m informative it gets stupid real fast.

i've never argued against that ... :) just never got to implementing it either (although it'd be probably a 2 liner in dataprovider).

freqtrade/strategy/interface.py
```diff
@@ -338,6 +338,32 @@ def informative_pairs(self) -> ListPairsWithTimeframes:
...
+            if not getattr(cls_attr, '_is_informative', False):
+                continue
+            timeframe = getattr(cls_attr, '_timeframe', None)
+            asset = getattr(cls_attr, '_asset', None)
+            assert asset is not None
+            assert timeframe is not None
+            if asset:
+                if '/' in asset:
+                    pair = asset
+                else:
```
xmatthias: I would not start supporting this - as it'll fail in the future and represent a breaking change (and provides limited benefits at this time, too).

While currently we only have spot pairs - this is likely going to change.
Once you venture out of spot pairs - you'll have pairs which ain't seperated by "/" (BTC-PERP on FTX, for example).
As such - relying on "/" as seperator will not work - and would result in `BTC-PERP/USD` - which however is not a valid pair - and is therefore a very nasty bug to debug.

freqtrade/strategy/interface.py
```diff
@@ -338,6 +338,32 @@ def informative_pairs(self) -> ListPairsWithTimeframes:
 # END - Intended to be overridden by strategy
 ###
 
+    def gather_informative_pairs(self) -> ListPairsWithTimeframes:
```
xmatthias, line 401: I would propose a 2-level approach for this method.
While `informative_pairs()` might return different values each time - the informative decorator won't.

As such - we can store the values for the decorator-level "informative" pairs after the first call - and avoid iterating throught he whole class dict every time.

Potentially, that can even be moved to strategy_resolver - so it's not dependent on having been called at least once (which means if something goes wrong, it'll fail at startup, not some time after).

freqtrade/strategy/interface.py
```diff
@@ -338,6 +338,32 @@ def informative_pairs(self) -> ListPairsWithTimeframes:
...
+        for attr_name in dir(self.__class__):
+            cls_attr = getattr(self.__class__, attr_name)
+            if not callable(cls_attr):
+                continue
+            if not getattr(cls_attr, '_is_informative', False):
+                continue
+            timeframe = getattr(cls_attr, '_timeframe', None)
+            asset = getattr(cls_attr, '_asset', None)
+            assert asset is not None
+            assert timeframe is not None
```
xmatthias: Please don't use assert in non-test code.
Few know it - but python has a "production-mode" flag which makes python skip all assert statements.

It'll also cause a non-cought exception.

freqtrade/strategy/interface.py
```diff
@@ -742,6 +768,16 @@ def advise_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
...
         logger.debug(f"Populating indicators for pair {metadata.get('pair')}.")
+
+        # call populate_indicators_Nm() which were tagged with @informative decorator.
+        for attr_name in dir(self.__class__):
+            cls_attr = getattr(self.__class__, attr_name)
+            if not callable(cls_attr):
+                continue
+            if not getattr(cls_attr, '_is_informative', False):
+                continue
+            dataframe = cls_attr(self, dataframe, metadata)
```
xmatthias: I'd avoid iterating the whole class as much as possible.

As mentioned above - when we gather the "required" pairs at startup - we can also store the list of "informative indicator methods" in a list - which we can iterate here quite easily without doing a lot of unnecessary looping during runtime (the final list of methods to call will not change during runtime, so gathering it once should be fine).

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +97,94 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
...
+                    this_fmt = '{asset}_' + this_fmt
+
+            # Not specifying an asset will define informative dataframe for current pair.
+            if not this_asset:
+                this_asset = metadata['pair']
+            assert isinstance(this_asset, str)
+
+            # Not specifying quote currency will use stake currency.
+            if '/' not in this_asset:
+                this_asset = f'{asset}/{self.config["stake_currency"]}'
```
xmatthias: err ... no please (reasoning - see above)

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +97,94 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
...
+
+            inf_metadata = {'pair': this_asset, 'timeframe': timeframe}
+            inf_dataframe = self.dp.get_pair_dataframe(this_asset, timeframe)
+            inf_dataframe = fn(self, inf_dataframe, inf_metadata)
+
+            # Clear stake currency from specified asset name.
+            fmt_asset = re.sub(rf'([^/])(/{self.config["stake_currency"]})?', r'\1',
+                               this_asset).lower()
+            # Non-stake quote currency will be kept in asset name, replace / separator with _.
+            fmt_asset = fmt_asset.replace('/', '_')
```
xmatthias: i think this will be quite confusing.
if i specify `asset='BTC/USDT'` - why won't my new column be called `BTC/USDT_rsi_1h` (without any way for me to override this behaviour).

Having a `/` in the column name is perfectly fine (you just can't have it in a filename) - so i'd not enforce this onto users without real reason

xmatthias (changes requested):
This will require some pretty good documentation and explanation - as it's introducing some "dataframe magic".

there's however quite some optimization potential available (see comments).

freqtrade/strategy/interface.py
rokups: You are right. This is bit of a pickle though. Specifying full pair in a decorator is complicated, because we may switch between USD/USDT/USDC/BUSD stake currencies and that would require strategy modification. There is no way to look up stake currency in config as at the time decorator is executed we have no access to config. Do you have any idea if we can somehow avoid hardcoding stake currency part, or do we have to live with this limitation?

Edit: Actually i know. Pair name could support formatting, so `"BTC/{stake_currency}"` would be formatted getting stake currency from config. 🥳

xmatthias: maybe allow the shorthand of `"{stake}"` too ... ? ;) as i'm usually pretty type-lazy :laughing:

freqtrade/strategy/strategy_helper.py
rokups: Reason is that stake currency is redundant information which just introduces extra typing without conveying much information. Likewise i replaced `/` with `_` for consistency. I understand we can have columns in any format, but there is no reason to opt in for inconsistent option if we can have a consistent one. Overriding is good though, so im making it so user can specify `fmt='{asset}_{name}_{timeframe}'` to get `BTC/USDT_rsi_1h` while `fmt='{asset_short}_{name}_{timeframe}'` would keep simplified `btc_rsi_1h`. Note that by default this does not exclude quote currency if it is different from stake currency, as that conveys useful information.

rokups:
I think i addressed all comments. I am not sure how to make tests. Would you be so kind and show me an example how i could inject such decorated method to a dummy strategy and get a resulting dataframe? I then would write tests covering all cases.

freqtrade/strategy/strategy_helper.py
xmatthias: it's not really redundant if i need 2 (or more) 1h pairs (with the same indicators) ... then the asset becomes quite important.

xmatthias:
I'm not sure if injecting it into an existing test-strategy is a good idea - as that'll not fully convey what reality will look like.
Tests should usually also serve as kindof a technical documentation - by injecting it we'll make this pretty difficult to digest without fully understanding how it works / what it tries to accomplish first.

A potentially better approach is to create a new strategy in `tests/strategy/strats/` - or "highjack" one of the existing ones that's not used in many places (most likely hyperoptable_strategy) - and implementing this as a "real" method (the way it's intended to be).

docs/strategy-customization.md
```diff
@@ -639,6 +639,121 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+
+            # once the profit has risen above 10%, keep the stoploss at 7% above the open price
+            if current_profit > 0.10:
+                return stoploss_from_absolute(trade.open_rate * 1.07, current_rate)
+
+            return 1
+
+    ```
+
+    Full examples can be found in the [Custom stoploss](strategy-advanced.md#custom-stoploss) section of the Documentation.
```
xmatthias: This sentence sounds like a lie - at least at this moment ... :laughing:

docs/strategy-customization.md
```diff
@@ -639,6 +639,121 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+
+### *@informative()*
+
+??? Example "Fast and easy way to define informative pairs"
+
+    Most of the time we do not need power and flexibility offered by `merge_informative_pair()`, therefore we can use a decorator to quicly define informative pairs.
```
xmatthias:
```suggestion
    Most of the time we do not need power and flexibility offered by `merge_informative_pair()`, therefore we can use a decorator to quickly define informative pairs.
```

docs/strategy-customization.md
```diff
@@ -639,6 +639,121 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+        # A callable `fmt(**kwargs) -> str` may be specified, to implement custom formatting.
+        # Available in populate_indicators and other methods as 'rsi_upper'.
+        @informative('1h', 'BTC/{stake}', '{name}')
+        def populate_indicators_btc_1h_2(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+            dataframe['rsi_upper'] = ta.RSI(dataframe, timeperiod=14)
+            return dataframe
+    
+        def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+            # Strategy timeframe indicators for current pair.
+            dataframe['rsi'] = ta.RSI(dataframe, timeperiod=14)
```
xmatthias, line 769: Does the dataframe passed in here already include all the above informative pairs merged to dataframe? 

Are informative pairs worked in sequence (so if i specify 2 methods - can the 2nd method rely on the first?) - or can we not guarantee sequence of the informative pair methods?

Are checks in place that ensure naming uniqueness (no column overwriting) if i'm using custom formats, and cause colisions this way?

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +97,98 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
...
+
+            inf_metadata = {'pair': asset_c, 'timeframe': timeframe}
+            inf_dataframe = self.dp.get_pair_dataframe(asset_c, timeframe)
+            inf_dataframe = fn(self, inf_dataframe, inf_metadata)
+
+            # Clear stake currency from specified asset name if it is a spot pair.
+            asset_short = asset_c
+            if '/' in asset_c:
+                asset_short = re.sub(rf'([^/])(/{self.config["stake_currency"]})?', r'\1',
+                                     asset_c)
```
xmatthias: I'm not quite sure i understand the intend with this, but i think it's to counter the side-effects from trying to be clever and only keeping one side of the pair.


as mentioned elsewhere - i'd always keep both sides of the pair in the name. 
using only one will become confusing and may end up with colliding names (assume i use 1h - BTC/USDT and BTC/BUSD ).

freqtrade/strategy/interface.py
xmatthias: I think you've missed this part.
While you do keep `_extra_informative_pairs` now - you can also keep the method itself in a list - so it's easy to be called without tripple-checking and iterating through the whole class.

docs/strategy-customization.md
```diff
@@ -639,6 +639,121 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+    See docstring of `@informative()` decorator for more information.
+
+!!! Note
+    Do not use `@informative` decorator if you need to use data of one informative pair when generating another informative pair. Instead, define informative pairs
+    manually as described [in the DataProvider section](#complete-data-provider-sample).
+
+!!! Warning
+    Methods tagged with `@informative()` decorator must always have unique names! Re-using same name (for example when copy-pasting already defined informative method)
+    will overwrite previously defined method and not produce any errors due to limitations of Python programming language. In such cases you will find that indicators
+    created in earlier-defined methods are not available in the dataframe. Carefully review method names and make sure they are unique!
```
xmatthias, line 802: I think it's worth mentioning that this runs before populate_indicators, as this automatically excludes these methods from hyperoptimization.

freqtrade/strategy/strategy_helper.py
rokups: I do not follow. Could you claraify what you mean by "2 (or more) 1h pairs (with the same indicators)"?

docs/strategy-customization.md
rokups: > Does the dataframe passed in here already include all the above informative pairs merged to dataframe?

Yes, that is the whole point of this decorator ;)

> Are informative pairs worked in sequence (so if i specify 2 methods - can the 2nd method rely on the first?) - or can we not guarantee sequence of the informative pair methods?

No. Informative pairs are processed in isolation. If one informative depends on another - old method should be used.

> Are checks in place that ensure naming uniqueness (no column overwriting) if i'm using custom formats, and cause colisions this way?

No, but pandas does it for us already. When merging dataframes it will append extra bit to columns with colliding names. That wont go unnoticed by the user.

docs/strategy-customization.md
rokups: Or maybe i should make hyperopt work with decorator? It does sound quite wrong to make decorator silently not work with hyperopt..

Edit: It works fine with auto-hyperopt. Legacy hyperopt would skip it. So maybe thats fine in the end?

freqtrade/strategy/strategy_helper.py
rokups: I am optimizing for most common case. 99% of strategies will not use a sibling stake currency, so whats the point to impose this extra typing on majority of users? Besides in such case user would get `btc_rsi_1h` + `btc_busd_rsi_1h`. IMO this little asymmetry in a very rare case beats users having to constantly hammer in stake currency. It also solves hardcoding of stake currency in strategy code. If we always kept stake currency then suddenly most common usecase would warrant a strategy modification when switching from USDT to BUSD as column names would need changing.

freqtrade/strategy/interface.py
rokups: I am no sure there is any reason for concern. Yes - it iterates all attributes of the class. This is also done on startup and takes nanoseconds so why care? I do agree "_is_informative" tag is rather dirty, but there is no better way to do it. Decorator code runs before class is created, so there is no way to get access to class object and store tagged methods in it. Therefore we must tag all methods and do this post-processing step after class is created.

Edit: Oh i see what you meant, this is in `advise_indicators()` 🤦🏻

freqtrade/strategy/strategy_helper.py
xmatthias: you claim that stake-currency is redundant - but is it really ? 

if i use `ETH/USDT` and `ETH/BUSD` and `ETH/BTC` (all on 1h) - and i switch stake currency around - i'll get different outputs due to the "redundant" stuff.

i might get 
`ETH_1h`, `ETH_BUSD_1h` and `ETH_BTC_1h` ,
`ETH_USDT_1h`, `ETH_1h` and `ETH_BTC_1h` ,
`ETH_USDT_1h`, `ETH_BUST_1h` and `ETH_1h` 



just because i switched stake-currency. As such - i cannot rely on the results i get as the "magic" is creating inconsistent results depending on the pair.

Obviously you can always have an additional RSI (or whatever else) on each column ... but i think you get the idea.

docs/strategy-customization.md
xmatthias: i did know the answer to these questions (except for the 2nd one) - but the user will not (hence the comment on the documentation page - not someplace else ;)

freqtrade/strategy/strategy_helper.py
xmatthias: Please see above - where i posted a sample where changing stake-currency will require changes in the remainder due to this.

freqtrade/strategy/strategy_helper.py
rokups: But this is far from common case. Vast majority will use `BTC/USDT` as informative pair. Thats it. Now when you switch stake currency you would have to update your strategy. This `BTC/{pair}` helper to avoid hardcoding stake currency goes out of the window. Your example is very valid, but as it is not common i would suggest to document how to handle such case instead of embedding stake currency all the time and making strategy stake-dependent.

docs/strategy-customization.md
xmatthias: it partially works with auto-hyperopt.
You can use the `.range` properties - but you cannot use `.value` - as this method is never called again after initial indicators are populated.

That's not an issue (it's aligned to how `populate_indicators` works) - i just think we should mention it as this is something new.

rokups:
Alright i think it deserves another review. I reworked code few more times, simplified, made it clearer and hopefully more maintainable. Documented and added a test.

I suspect you will again comment about automatic column naming, so i will reiterate my decision process beforehand ;)

Column naming is optimized for most common case, so we need least amount of typing for the things we do 95% of the time.
* Informative timeframes for current pair will format columns as `{column}_{timeframe}`. For example `rsi_1h`.
* Informative timeframes for other pairs will format column as:
  * When quote currency matches stake currency: `{base}_{column}_{timeframe}`, for example `btc_rsi_1h`. Quote currency is omitted on purpose, because such informatives are most common and it allows user to easily switch stake currency without having to modify strategy code.
  * When quote currency is different from stake currency: `{base}_{quote}_{column}_{timeframe}`, for example `btc_eth_rsi_1h`. This is a rare case (i have not seen any public strategies use such informative), but if such informatives are used there should be no issue with hardcoding quote currency in the strategy.

As you mentioned, user may want to use multiple informatives with different stablecoins as quote currency. This is probably most rare usecase of all. Nevertheless in such case user should use `fmt='{base}_{quote}_{column}_{timeframe}'` or similar format, explicitly including quote currency in the format.

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +103,128 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
...
+    define informative indicators.
+
+    Example usage:
+
+        @informative('1h')
+        def populate_indicators_1h(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+            dataframe['rsi'] = ta.RSI(dataframe, timeperiod=14)
+            return dataframe
+
+    :param timeframe: Informative timeframe. Must always be higher than strategy timeframe.
```
xmatthias: TODO: are we checking that?

freqtrade/strategy/interface.py
```diff
@@ -134,6 +136,39 @@ def __init__(self, config: dict) -> None:
...
 
+        # Gather informative pairs from @informative-decorated methods.
+        self._ft_informative: Dict[
+            Tuple[str, str], Tuple[InformativeData,
+                                   Callable[[Any, DataFrame, dict], DataFrame]]] = {}
+        for attr_name in dir(self.__class__):
+            cls_method = getattr(self.__class__, attr_name)
+            if not callable(cls_method):
+                continue
+            ft_informative = getattr(cls_method, '_ft_informative', [])
```
xmatthias: Wouldn't you want to use "None" here as default argument?

That would make the "next" line skip this immediately (if it falls into default, it's not relevant for the remainder of this function) ?

docs/strategy-customization.md
```diff
@@ -639,6 +639,129 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+        # * {asset} - full name of the asset, for example 'BTC/USDT'.
+        # * {base} - base currency in lower case, for example 'eth'.
+        # * {BASE} - same as {base}, except in upper case.
+        # * {quote} - quote currency in lower case, for example 'usdt'.
+        # * {QUOTE} - same as {quote}, except in upper case.
+        # * {column} - name of dataframe column.
+        # * {timeframe} - timeframe of informative dataframe.
+        # A callable `fmt(**kwargs) -> str` may be specified, to implement custom formatting.
+        # Available in populate_indicators and other methods as 'rsi_upper'.
+        @informative('1h', 'BTC/{stake}', '{name}')
```
xmatthias: What would `"{name}"` refer to (what's the resut)? 

It's not explained above - and i'm not sure it'll work based on the `fmt_args` specified in strategy_helper ...

docs/strategy-customization.md
```diff
@@ -639,6 +639,129 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+
+            return 1
+
+    ```
+
+### *@informative()*
+
+In most common case it is possible to easily define informative pairs by using a decorator. All decorated `populate_indicators_*` methods run in isolation,
+not having access to data from other informative pairs, in the end all informative dataframes are merged and passed to main `populate_indicators()` method.
+When hyperopting, please follow instructions of [optimizing an indicator parameter](hyperopt.md#optimizing-an-indicator-parameter).
```
xmatthias: This is a good point and i'd mention it explicitly that indicators defined with this method will not receive "different" values in `self.buy_whatever.value` - but must use `.range` (which basically means that all `@informative` methods behave the same as `populate_indicators()` during hyperopt (only called once).

docs/strategy-customization.md
```diff
@@ -639,6 +639,129 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+        def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+            # Strategy timeframe indicators for current pair.
+            dataframe['rsi'] = ta.RSI(dataframe, timeperiod=14)
+            # Informative pairs are available in this method.
+            dataframe['rsi_less'] = dataframe['rsi'] < dataframe['rsi_1h']
+            return dataframe
+
+    ```
+
+    See docstring of `@informative()` decorator for more information.
```
xmatthias: The sample above is a good point - but i'd also have the docstring (and method signature) here for reference.
While a docstring seems easy enough - availability will depend on the editor (and editor setup)- and not everyone will be able to "simply" launch an interactive console, import this and display help - so having that in the documentation as well will greatly help freqtrade / python padawans.

freqtrade/strategy/interface.py
```diff
@@ -134,6 +136,39 @@ def __init__(self, config: dict) -> None:
         self._last_candle_seen_per_pair: Dict[str, datetime] = {}
         super().__init__(config)
 
+        # Gather informative pairs from @informative-decorated methods.
+        self._ft_informative: Dict[
+            Tuple[str, str], Tuple[InformativeData,
+                                   Callable[[Any, DataFrame, dict], DataFrame]]] = {}
```
xmatthias: Please extract `Callable[[Any, DataFrame, dict], DataFrame]]` and assign it a named type. 
it's used 3 times in total (which justifies extraction) - and this type-def alone scares me from looking at the rest :laughing:

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +103,128 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
...
+    :param asset: Informative asset, for example BTC, BTC/USDT, ETH/BTC. Do not specify to use
+    current pair.
+    :param fmt: Column format (str) or column formatter (callable(name, asset, timeframe)). When not
+    specified, defaults to {asset}_{name}_{timeframe} if asset is specified, or {name}_{timeframe}
+    otherwise.
+    * {asset}: name of informative asset, provided in lower-case, with / replaced with _. Stake
+    currency is not included in this string.
+    * {name}: user-specified dataframe column name.
+    * {timeframe}: informative timeframe.
+    :param ffill: ffill dataframe after mering informative pair.
```
xmatthias:
```suggestion
    :param ffill: ffill dataframe after merging informative pair.
```

freqtrade/strategy/interface.py
```diff
@@ -134,6 +136,39 @@ def __init__(self, config: dict) -> None:
...
+                elif self.dp is not None:
+                    for pair in self.dp.current_whitelist():
+                        if (pair, timeframe) in self._ft_informative:
+                            raise OperationalException(f'Informative pair {pair} {timeframe} can '
+                                                       f'not be defined more than once!')
+                        self._ft_informative[(pair, timeframe)] = (informative_data, cls_method)
+
+    def _format_pair(self, pair: str) -> str:
+        return pair.format(stake_currency=self.config['stake_currency'],
+                           stake=self.config['stake_currency']).upper()
```
xmatthias: I'd not have this in the code twice (here and in `strategy_helpers.py`...)

docs/strategy-customization.md
rokups: That was a leftover from previous naming. It is supposed to be `'{column}`. 👍🏻

freqtrade/strategy/interface.py
rokups: Another leftover forgotten after i moved it out of the class 😊

docs/strategy-customization.md
xmatthias: i was assuming as much :laughing: just pointing you towards it - WHAT the fix is (add name to the dict, replace in the docs) is pretty much up to you :laughing:

docs/strategy-customization.md
```diff
@@ -639,6 +639,157 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+
+    :param timeframe: Informative timeframe. Must always be equal or higher than strategy timeframe.
+    :param asset: Informative asset, for example BTC, BTC/USDT, ETH/BTC. Do not specify to use
+    current pair.
+    :param fmt: Column format (str) or column formatter (callable(name, asset, timeframe)). When not
+    specified, defaults to:
+    * {base}_{column}_{timeframe} if asset is specified and quote currency does match stake
+    curerncy. 
+    * {base}_{quote}_{column}_{timeframe} if asset is specified and quote currency does not match 
+    stake curerncy.
```
xmatthias:
```suggestion
    stake currency. 
```
*edit* also 2 lines above, and probably in the "real" doc-string too

docs/strategy-customization.md
```diff
@@ -639,6 +639,157 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+        # def informative_pairs(self): ...
+
+        # Define informative upper timeframe for each pair. Decorators can be stacked on same 
+        # method. Available in populate_indicators as 'rsi_30m' and 'rsi_1h'.
+        @informative('30m')
+        @informative('1h')
+        def populate_indicators_1h(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+            dataframe['rsi'] = ta.RSI(dataframe, timeperiod=14)
+            return dataframe
+
```
xmatthias, line 742: Interrestingly, this case seems to work ONLY if there's other cases there as well.

so a "very" simple strategy limited to this decorator will not work (never call the decorated function).

```
    @informative('4h')
    def populate_indicators_1h(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        dataframe['rsiiiiii'] = ta.RSI(dataframe, timeperiod=14)
        return dataframe

        print(dataframe.columns)
```

Adding a 2nd parameter will hoewever call it and make it work.

```
    @informative('4h', 'BTC/USDT)
    def populate_indicators_1h(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        dataframe['rsiiiiii'] = ta.RSI(dataframe, timeperiod=14)
        return dataframe
```

I wonder how tests can pass (probably because they directly load the strategy - and there's a TON of mocking/patching going on there ... as I also copied the new test-strategy and ran it - and it's missing the "rsi_30m" and "rsi_1h" columns too (only modification was a `print(dataframe.columns)` in `populate_indicators`.

xmatthias (commented):
I'll ... wait with further reviews - as my tests were not really successful ... at least in the first case (the one that's in the docs most often ...) :laughing:

docs/strategy-customization.md
rokups: I fixed that (and few other things) but... I do hate the fix :( Problem is that creating informatives for each current pair requires access to dataprovider, but that is only available after class is initialized. So i deferred initialization of informatives to a method that is called after dataprovider is attached. This, however, means that users creating strategy object in the notebook would have to call this initialization method manually. I do not think this is a good solution, but i am not sure what to do about it. Dataprovider could be attached to class itself like it used to be in the past. However there are multiple issues with that. IIRC hyperopt had issues with that approach and i had to switch to attaching object to class instance itself. Also strategy loader creates strategy instance immediately, not giving us access to strategy class before object creation. Do you think it is reasonable to change strategy loader to return strategy class instead of instance?

xmatthias: your comment gave me something to think ... and i think thecurrent solution will fail / not work (not because of the additional method ...).

You rely on the fact that the strategy is fully initialized, creating the "additional pairs lists" at instantiation time.
This is correct for static pairlists / backtest ... but doesn't account for dynamic pairlists (which can change over time).

As such - i think this PR will work only for the first pairlist ... and will start to have "missing" pairs the moment new pairs will be add.
my test ran into this (or a similar scenario) by accident - but it really exposed a different problem, which is that the "final" pairs will not be known when the strategy starts up.
You should "probably" be able to run into this by starting the bot with an existing database (with open trade) - where the open trade-pair is no longer in the pairlist (at the point you call _initialize, that pair is not yet part of the pairlist).

i think to solve this, we'll have to keep the "pairs" field empty for `@informative('30')` cases without defined pair (which means it'll use all pairs in the current whitelist).
As part of `gather_informative_pairs` - we'll then get the existing keys (where the pair is known) - and expand the "pair is unknown" cases (might need to be stored separately) to the "current" pairlist when actually gathering the pairs.

docs/strategy-customization.md
```diff
@@ -639,6 +639,157 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
     Full examples can be found in the [Custom stoploss](strategy-advanced.md#custom-stoploss) section of the Documentation.
 
+!!! Note
+    Providing invalid input to `stoploss_from_open()` may produce "CustomStoploss function did not return valid stoploss" warnings.
+    This may happen if `current_profit` parameter is below specified `open_relative_stop`. Such situations may arise when closing trade
+    is blocked by `confirm_trade_exit()` method. Warnings can be solved by never blocking stop loss sells by checking `sell_reason` in
+    `confirm_trade_exit()`, or by using `return stoploss_from_open(...) or 1` idiom, which will request to not change stop loss when
+    `current_profit < open_relative_stop`.
+
+### *stoploss_from_absolute()*
```
xmatthias, line 649: I'm not sure in what usecase you'd want to define an absolute stoploss - it'll always need to depend on the entry price (you can't generally say - stoploss when BTC hits 48k - it's simply not going to work with absolute values as prices will change over time).

You might wanna define it based on profit ... like "we're now +10% - set stoploss at +5% (so worst case, that trade is 5% winner) - but let it run for now - where we have `stoploss_from_open()`.

Even in the sample below, what you really want to say is "at 7% profit" - not **at 107$**.
Using absolute values will be difficult/impossible to generalize this if you have coins that cost 107$ - and others that cost 40.000$ - 7$ for one is 7% - while for the other it's not even 1% ...

freqtrade/strategy/interface.py
```diff
@@ -118,7 +121,7 @@ class IStrategy(ABC, HyperStrategyMixin):
     # Class level variables (intentional) containing
     # the dataprovider (dp) (access to other candles, historic data, ...)
     # and wallets - access to the current balance.
-    dp: Optional[DataProvider] = None
+    dp: DataProvider
```
xmatthias: I'd keep this as optional from a typehint perspective.

while we provide dataprovider within freqtrade, that's not guaranteed when running in a notebook (for example) - where it's not even part of the docs, so a check `if self.dp` will still be beneficial, and having `Optional` here will allow a good editor to point this out.

docs/strategy-customization.md
rokups:
```py
dataframe['stoploss'] = dataframe['close'] - dataframe['atr'] * 2
...
return stoploss_from_absolute(last_candle['stoploss'], current_rate)
```
And usecase is not only involving ATR, but anything that involves absolute price (stop at moving average, lower channel/band).

xmatthias: Then this should be the example, not one that is using relative pricing to get an absolute price from open - which is where `stoploss_from_open()` will be easier to use, and makes this appear non-necessary.

we'll also want a test for this ... (i'm not sure why it's in this PR - it feels "snuck" in somehow :laughing:  - it's not related to informative decorator apart from having documentation nearby...).

freqtrade/strategy/strategy_helper.py
```diff
@@ -83,3 +106,133 @@ def stoploss_from_open(open_relative_stop: float, current_profit: float) -> floa
...
+        return fn
+    return decorator
+
+
+def _format_pair_name(config, pair: str) -> str:
+    return pair.format(stake_currency=config['stake_currency'],
+                       stake=config['stake_currency']).upper()
+
+
+def _create_and_merge_informative_pair(strategy, dataframe: DataFrame, metadata: dict,
```
xmatthias: I think this might work out better (from a maintenance perspective) if it's in a seperate file (both this and the decorator itself).

We'll want people to import from `freqtrade.strategy` anyway (not from the files below) - and having it seperate will make it clearer what belongs together, and what's a "standalone" method (like the remaining methods in this file).

docs/strategy-customization.md
```diff
@@ -639,6 +639,157 @@ Stoploss values returned from `custom_stoploss` must specify a percentage relati
...
+    :param timeframe: Informative timeframe. Must always be equal or higher than strategy timeframe.
+    :param asset: Informative asset, for example BTC, BTC/USDT, ETH/BTC. Do not specify to use
+    current pair.
+    :param fmt: Column format (str) or column formatter (callable(name, asset, timeframe)). When not
+    specified, defaults to:
+    * {base}_{column}_{timeframe} if asset is specified and quote currency does match stake
+    currency. 
+    * {base}_{quote}_{column}_{timeframe} if asset is specified and quote currency does not match 
+    stake currency. 
+    * {column}_{timeframe} if asset is not specified.
```
xmatthias: I still find the default behaviour of this confusing.

Assume this case: `@informative('1h', asset='BTC/BUSD')` (which also adds an `rsi` column).

The outcome will be:
* `btc_busd_rsi_1h` if my stake-currency is USDT
* `btc_rsi_1h` if my stake-currency is BUSD

This makes it difficult to work with in cases where you wanna switch stake-currency (which you'll probably not think about when first implementing this in a strategy) - while keeping the informative pair consistent.

i know we can define it with the format explicitly - but the default seems inconsistent and difficult to explain as it changes based on the used stake-currency.

While it'll force me to then use `f"btc_busd_rsi_1h"` (which is 5 characters longer than the alternative) - i can rely on this to work even if i switch stake-currency (but would still like to use the same informative pair).

xmatthias (commented):
I've pushed a very small simplification (as it confused me at first) ...

docs/strategy-customization.md
rokups: > While it'll force me to then use f"btc_busd_rsi_1h" (which is 5 characters longer than the alternative) - i can rely on this to work even if i switch stake-currency (but would still like to use the same informative pair).

That would force you to change strategy source code every time stake currency is changed (most common case is using BTC/STAKE as informative). This is no good...

Maybe a compromise would be `BTC/BUSDT` expanding to `btc_busd_rsi_1h` and `BTC/{stake}` expanding to `btc_rsi_1h`? One more magical rule though. I dont think we can save people from not reading docs.

docs/strategy-customization.md
rokups: > i'm not sure why it's in this PR - it feels "snuck" in somehow laughing - it's not related to informative decorator apart from having documentation nearby...

True. `@informative` was supposed to be a simple helper too, but it somehow snowballed.... 😊

docs/strategy-customization.md
rokups: Well we need some kind of a decision. What do you think would be a proper thing to do, hopefully not introducing a need for source code modifications each time stake is changed?

xmatthias: i need to change sourcecode now due to the inconsistent "magic" that's happening.

`@informative('1h', asset='BTC/BUSD')` expands to 2 different things depending on something unrelated ( change of `stake_currency`) -  so you force users to change sourcecode when it's not necessary.
It should always expand to `btc_busd_rsi_1h` - then it's whats expect (it's a stable, hardcoded pair, that's not changing when i change stake-currency).

`@informative('1h', asset='BTC/{stake}')` on the other hand is different, and can expand to `btc_rsi_1h` -  it's an explicit user choice of using BTC vs. {stake} - not BTC/USDT or BTC/USDT (which is also an explicit choice - to always compare against that particular pair). 
It'll also not change if i switch from USDT to BUSD (or EUR) - but it'll be always the same (always BTC vs. "my stake currency".

There is a huge difference between the user intend here (what the user desired to do) - one is "always use BTC/USDT, because i explicitly told you so and don't care about BTC/BUSD" - the other is "use BTC/{stake}" - so always use BTC/<my current stake>.

It's therefore not "one more magical rule" - but it's "less" magic - with predictable outcome.

freqtrade/strategy/informative_decorator.py
```diff
@@ -0,0 +1,129 @@
+from typing import Any, Callable, NamedTuple, Optional, Union
+
+from mypy_extensions import KwArg
```
xmatthias: this line concerns me a bit ...

it's assuming mypy is installed - which might be the case for development instances, but is neither the case in docker images, nor in "regular" user's installs (if dev-dependencies have been unselected).

rokups: My bad. Removed that. Idk how it came to be :D

xmatthias (approved):
Finally  (:laughing:) - LGTM :+1:

rokups:
You didn't give me chance to clean it up :(

PR #5285, opened 2021-07-18
Remove startup-candles after populating buy/sell signals

### LINKED ISSUE #5242: Mode to allow startup candles sent to populate_*_trend()
Would it be possible to have a user selectable mode that sends startup candle data in the dataframe that is passed to populate_trend() methods during hyperoptimising? This would allow indicators to be calculated from within those routines, and that would mean you could use  .values passed by hyperopt in the indicators without nested for loops or creating a large number of columns. A simple conditional inclusion of indicators in populate_indicators() or populate_trend() depending on whether hyperopting or not could allow normal startup candle free dataframes in other run modes. It could make it a lot easier for strategies that, for example, have multiple parameters in indicators like MACD to be hyperotable. I realise this may slow things down somewhat, but simplifies things. Interested in your thoughts.

xmatthias:
## Summary
Startup-candles should be removed after buy-signal population (so they're only relevant for the backtesting period per se).

closes #5242
## Quick changelog

- Leave startup-candle-count available for `populate_<buy/sell>_trend`.

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -15,7 +15,7 @@ from freqtrade.configuration import TimeRange, remove_credentials, validate_conf
 from freqtrade.constants import DATETIME_PRINT_FORMAT
 from freqtrade.data import history
 from freqtrade.data.btanalysis import trade_list_to_dataframe
-from freqtrade.data.converter import trim_dataframes
+from freqtrade.data.converter import trim_dataframe, trim_dataframes
 from freqtrade.data.dataprovider import DataProvider
 from freqtrade.enums import BacktestState, SellType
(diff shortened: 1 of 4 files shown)


2020Cyberworld (commented):
self.timerange = TimeRange.parse_timerange(None data.config.get
        (None if self.config.get('timerange') is None else str(self.config.get('timerange'))) 

@xmatthias Really awesome bot! Hopefully this is the right debug

xmatthias:
> self.timerange = TimeRange.parse_timerange(None data.config.get
> (None if self.config.get('timerange') is None else str(self.config.get('timerange')))
> 
> @xmatthias Really awesome bot! Hopefully this is the right debug

I'm not certain what you're trying to say with this ...

PR #5287, opened 2021-07-18
Dataprovider caching and trimming of historical informative.

GeorgeMurAlkh:
Current unit tests fully covering functionality.

## Summary
Previously dataprovider was loading historical ohlcv from disk for every `self.dp.get_pair_dataframe` call in strategy code. Also it didn't trim loaded dataframe according `timerange` parameter. Now it storing loaded dataframes in property of class and returning a copy of dataframe if it already loaded from disk. Overall this fix should decrease backtesting time up to 250+% (depending on `timerange` and  informative size).


xmatthias (changes requested):
The change per se (all lines you changed) are fine :+1: 

The problem is with the lines that you didn't change....

---


This will however introduce a (subtle and stupid) bug in "webserver mode" if the timerange is changed between 2 backtests:

* start webserver mode
* run backtesting with a specified timerange (say `20210401-20210501`) with a strategy that uses informative pairs (and as such, is subject to this caching).
* run the same backtest again, but for `20210501-`. 

As the dataprovider will find this data as "already loaded" - (the key is pair / timeframe) - it'll not load it again, and therefore return the wrong data (for the wrong timerange) for the informative pair.
For real pairs, that's not an issue as freqtrade will reload the data, and reassign them to the dataprovider cache - that's however not possible in this case.

I can think of 2 fixes for this at the moment:
* evict the backtest object (basically move the timerange check in `api_backtest.py` to the first "if")
* Implement a method that selectively evicts this cache only if timerange changed.

Personally, i'd prefer the 2nd method ... 

Without either of the 2 fixes, i won't be able to merge this though - as the bug will be very subtle and difficult to spot for users.

GeorgeMurAlkh:
> The change per se (all lines you changed) are fine 👍
> 
> The problem is with the lines that you didn't change....
> 
> This will however introduce a (subtle and stupid) bug in "webserver mode" if the timerange is changed between 2 backtests:
> 
> * start webserver mode
> * run backtesting with a specified timerange (say `20210401-20210501`) with a strategy that uses informative pairs (and as such, is subject to this caching).
> * run the same backtest again, but for `20210501-`.
> 
> As the dataprovider will find this data as "already loaded" - (the key is pair / timeframe) - it'll not load it again, and therefore return the wrong data (for the wrong timerange) for the informative pair.
> For real pairs, that's not an issue as freqtrade will reload the data, and reassign them to the dataprovider cache - that's however not possible in this case.
> 
> I can think of 2 fixes for this at the moment:
> 
> * evict the backtest object (basically move the timerange check in `api_backtest.py` to the first "if")
> * Implement a method that selectively evicts this cache only if timerange changed.
> 
> Personally, i'd prefer the 2nd method ...
> 
> Without either of the 2 fixes, i won't be able to merge this though - as the bug will be very subtle and difficult to spot for users.

@xmatthias I went with 1 suggestion to not overcomplicate dp class.

xmatthias:
everything that's caching dataframes is giving me a headache it seems :laughing: 
it's simply a quite complex topic ...

I'm currently unsure about impact from this to hyperopt ... 
it'll increase memory consumption for sure (it's another set of dataframes that are kept in memory).

While not necessarily a problem - i'm not sure it's really necessary either.
Assuming informative-pairs are used in `populate_indicators()` - we could evict them after that step - and have the "parallel" part of hyperopt be unaffected by this.

On the other hand - if you use informative pairs in `populate_*_trend()` methods (so it's parameters can be optimized) - this PR would positively impact hyperopt performance by avoiding the load from disk on every epoch.
This case however might also result in "odd", potentially difficult to find / recreate bugs as it'd cause concurrent writes to the same variable (hyperopt uses multiprocessing to run backtesting in parallel).

The only way to mitigate this last point i can think of at this moment would be to "respect" the return value of `informative_pairs` and load the pairs upfront. This is currently irrelevant to backtesting/hyperopt - but doing so  would automatically fix a potential strategy-bug / problem where you use informative-pairs in backtesting, and "forgot" to add them to the list of required informative pairs - so it's not available in live.


Before progressing though - letme cheat hear and ask for @rokups opinion on this ...

rokups:
> This case however might also result in "odd", potentially difficult to find / recreate bugs as it'd cause concurrent writes to the same variable (hyperopt uses multiprocessing to run backtesting in parallel).

Would it? multiprocessing runs things in parallel by running them in separate processes that do not share memory space. Seems to we subprocesses can trash everything all they want as long as they return correct results from a function running in a subprocess.

> On the other hand - if you use informative pairs in populate_*_trend() methods (so it's parameters can be optimized) - this PR would positively impact hyperopt performance by avoiding the load from disk on every epoch.
This case however might also result in "odd", potentially difficult to find / recreate bugs as it'd cause concurrent writes to the same variable (hyperopt uses multiprocessing to run backtesting in parallel).

I would say optimize for common case, add knobs to achieve uncommon case. Advanced config variable could be introduced controlling dataframe caching, enabled by default. Should anyone ever run into problems - there will be something to point them to. I also think it will be extremely rare occurrence.

> and "forgot" to add them to the list of required informative pairs

That is where decorator PR comes in. That way you can not forget. 👍🏻

xmatthias (approved):
I guess we will have to keep an eye out for the potential problems mentioned here and eventually tackle these as seperate PR's - but for now i don't see why we should withhold from merging this further - as i'm also unable to recreate the theoretical issues i mentioned above (it's just been thoughts, really).

LGTM :+1:

PR #5299, opened 2021-07-19
Add buy signal name

### LINKED ISSUE #3081: Tagging of buy & sell conditions
## New feature request

Ability to tag buy and sell strategies and print what triggered the buy and sell in telegram bit output. 

## Example:
Having two dataframe.loc where we can tag a name to each. 
Dataframe.loc[
(
   (Condition)
), 'buytag'] = 'my tag1'

Dataframe.loc[
(
.   (Condition)
), 'buytag'] = 'my tag2'

kevinjulian:
## Summary
When use multiple buy signal, I (maybe we) need to know the buy signal that trigger a trade
Then on custom_sell, I able to use ```trade.buy_tag```

### Buy Tag on Strategy
Normal way to use populate_buy_trend

```
dataframe.loc[
        (
            (dataframe['rsi'] < 35) &
            (dataframe['volume'] > 0)
        ),
        'buy'] = 1
```

Use with custom buy_tag
```
dataframe.loc[
        (
            (dataframe['rsi'] < 35) &
            (dataframe['volume'] > 0)
        ),
        ['buy', 'buy_tag']] = (1, 'my_buy_signal_01')
```

Then on custom_sell just call it
```
trade.buy_tag
```

### Buy Tag on Telegram

When the trade has buy_tag
image
-------
image

closes #3081


freqtrade/persistence/models.py
```diff
@@ -703,6 +705,7 @@ class Trade(_DECL_BASE, LocalTrade):
     sell_reason = Column(String(100), nullable=True)
     sell_order_status = Column(String(100), nullable=True)
     strategy = Column(String(100), nullable=True)
+    buy_signal_name = Column(String(100), nullable=True)
```
xmatthias: When adding columns, you'll need to
* add them to the relevant sections in `migrations.py`
* Make sure the last column addition is in `check_migrate()` in `migrations.py` (so the migration knows if a migration is necessary

freqtrade/strategy/interface.py
```diff
@@ -504,18 +504,20 @@ def get_signal(self, pair: str, timeframe: str, dataframe: DataFrame) -> Tuple[b
                 'Outdated history for pair %s. Last tick is %s minutes old',
                 pair, int((arrow.utcnow() - latest_date).total_seconds() // 60)
             )
-            return False, False
+            return False, False, ''
 
-        (buy, sell) = latest[SignalType.BUY.value] == 1, latest[SignalType.SELL.value] == 1
+        (buy, sell, buy_signal_name) = latest[SignalType.BUY.value] == 1,\
+            latest[SignalType.SELL.value] == 1,\
+            latest.get(SignalNameType.BUY_SIGNAL_NAME.value, '')
```
xmatthias: I don't think this is valid syntax (not for this object-type, anyway).

`latest` is not a dictionary (which would offer .get()`) - but a pandas Series.
As such - you'll not be able to use it this way, but you'll have to explicitly check if the signal column is part of the dataframe - and only then assign it with it's value.

kevinjulian: pandas has ```.get()```
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.get.html

I tried and everything ok. I will add test case to prove this

kevinjulian:
@xmatthias many testcase turn error because of `key_error: buy_signal_name not in index`
Is it possible to set the default column on freqtrade dataframe?

freqtrade/enums/signaltype.py
```diff
@@ -11,6 +11,6 @@ class SignalType(Enum):
 
 class SignalNameType(Enum):
     """
-    Enum to distinguish between buy and sell signals
+    Enum for signal name
```
xmatthias: maybe
```suggestion
    Enum for signal columns
```
?

freqtrade/persistence/migrations.py
```diff
@@ -160,7 +161,7 @@ def check_migrate(engine, decl_base, previous_tables) -> None:
     table_back_name = get_backup_name(tabs, 'trades_bak')
 
     # Check for latest column
-    if not has_column(cols, 'open_trade_value'):
+    if not has_column(cols, 'open_trade_value') or not has_column(cols, 'buy_signal_name'):
```
xmatthias: there can only be one last column (last added). If that last column exists, no migration is necessary.
As such - this should look (roughly?) as follows:
```suggestion
    if not has_column(cols, 'buy_signal_name'):
```

rokups:
I am not sure this would work for signals that are not trivial. For example look at NFI: https://github.com/iterativv/NostalgiaForInfinity/blob/main/NostalgiaForInfinityNext.py
It has a crazy amount of buy signals. Quite often multiple buy signals may be fire. Naturally we would like to know which signals fired, not only which first (or last) signal fired.

kevinjulian:
> I am not sure this would work for signals that are not trivial. For example look at NFI: https://github.com/iterativv/NostalgiaForInfinity/blob/main/NostalgiaForInfinityNext.py
> It has a crazy amount of buy signals. Quite often multiple buy signals may be fire. Naturally we would like to know which signals fired, not only which first (or last) signal fired.


I don't think that's become a problem. If the signal fired (doesn't matter what signal is), its still buy signal.
I know what you need is the "stop" method after the buy signal turn to 1, but its not possible with current architecture in freqtrade layer. You can do it on Strategy layer, in your case NFIStrategy (but still I don't think it's necessary)
do you have any insight? @xmatthias

rokups:
> If the signal fired (doesn't matter what signal is), its still buy signal.

If that were true this PR would not exist. Clearly it is important what buy signal is firing. Imagine we get a string of terrible buys which result from multiple buy conditions. All conditions are not necessarily bad, but there may be one condition that persistently shows up with losers. How would you pick out conditions that produce more harm than good, if you can only see which first condition produced a buy?

What i do is setting a buy condition tag column for each buy condition, prefixing that column with `buy_`. That way i can get all buy conditions that triggered. It is not perfect, it introduces a lot of extra columns, but it produces accurate information. I figured if we are going to make it a core feature, might as well go the full length instead of doing only bare minimum that will be of limited use.

kevinjulian:
> > If the signal fired (doesn't matter what signal is), its still buy signal.
> 
> If that were true this PR would not exist. Clearly it is important what buy signal is firing. Imagine we get a string of terrible buys which result from multiple buy conditions. All conditions are not necessarily bad, but there may be one condition that persistently shows up with losers. How would you pick out conditions that produce more harm than good, if you can only see which first condition produced a buy?
> 
> What i do is setting a buy condition tag column for each buy condition, prefixing that column with `buy_`. That way i can get all buy conditions that triggered. It is not perfect, it introduces a lot of extra columns, but it produces accurate information. I figured if we are going to make it a core feature, might as well go the full length instead of doing only bare minimum that will be of limited use.

yes true, but I think what you need (still) stop iterating others signal once one of your signal turn on
something like this
```
  buy_signal_01 = (
    (dataframe['rsi'] < 20) &
    (dataframe['volume'] > 0)
  )
  if buy_signal_01.tail(1).any():
    dataframe[buy_signal_01, ['buy', 'buy_tag']] = 1, 'buy_with_rsi'
    return dataframe
  if ...:
    ...
    return dataframe
  return dataframe
```

with that you don't need multiple column for buy signal 🤔

rokups:
With that you also do not know all the signals that would have triggered a buy. I would go as far and say that partial information is worse than no information as you can make wrong assumptions not knowing everything while thinking you do, instead of proceeding cautiously because you are aware of what you do not know.

xmatthias:
Having an unlimited number of potential buy reasons (speak, buy reason columns) does not make sense and will be way too complicated in an implementation and maintenance phase for the benefit it generates.

To properly facilitate that, you'll need an additioal table with FK's.
If that'd be the approach proposed here, i'd probably reject it as it'll be too complicated and increase the maintenance burden without providing much benefit (i doubt the benefit of this overall as it'll only benefit a small subset of strategies - so this will have to be a minimal feature to be worth the effort of maintenance and support).

@rk What you're talking about is an edge-case which you'll only need for some (over-engineered) strategies - which can easily be handled within the strategy by combining all "buy reason columns" you used into one and appending them one after the other.

We could even provide a Helper method for that which combines all `buy_reason_xxx` columns into the final buy_reason column - assuming this becomes a problem down the road. 
for now, i don't see this as a problem - so let's see how / if this is used at all.

Obviously, if you go over 100chars  on the "combined" buy-column, it'll be truncated, but if you use sanely short namings, it'll not be a problem.

rk:
@xmatthias I'm `@rk` not `@rokups`. Please don't abbreviate at-mentions. 😩

kevinjulian:
then should we implement this (combine all buy reason) at this PR? maybe looks like this? @xmatthias 
```
def buy_with_tag(dataframe: pandas.DataFrame, condition: tuple, buy_tag: str):
    dataframe.loc[condition, 'buy'] = 1
    dataframe.loc[condition, 'buy_tag'] =dataframe.loc[condition, 'buy_tag'] + '|' + buy_tag
    return dataframe

    condition_01 = (
        (dataframe['rsi'] < 30)
    )
    condition_02 = (
        (dataframe['fast_d'] < 30)
    )
    dataframe = buy_with_tag(dataframe, condition_01, 'b_01')
    dataframe = buy_with_tag(dataframe, condition_02, 'b_02')
```

xmatthias:
> @xmatthias I'm `@rk` not `@rokups`. Please don't abbreviate at-mentions.

Sry - he's @rk on discord ... so it's "half-intentional" ... :( 

> then should we implement this (combine all buy reason) at this PR

In theory, this would work (although it'll produce an exception if you use it like this because the first iteration, the column won't exist - and the type-hint for the 2nd param is wrong - but that's implementation details, really).

The big concern / doubt i have with this approach is that it'll no longer be a clear, native pandas assignments - and increases the potential of syntax errors (at least if you in-line the conditions into the method call - and use complex conditions) - as you'll in both instances be using round parenteses.

It'll also (potentially) conflict / require changes with the upcomming short integration.
As such - i'd hold of from such a method for the moment - at least until the approach for the short interface is final to avoid potential (most likely breaking) changes to a method introduced only few weeks earlier.

kevinjulian:
> It'll also (potentially) conflict / require changes with the upcomming short integration.
> As such - i'd hold of from such a method for the moment - at least until the approach for the short interface is final to avoid potential (most likely breaking) changes to a method introduced only few weeks earlier.

ok I will look into this PR again after short interface is final

xmatthias:
> ok I will look into this PR again after short interface is final

well that'll be a seperate PR i'd say, limited to this helper method providing help with "multiple" signals (if necessary).
i'd not see the non-existence of this as blocking for the work you're doing in this PR.

freqtrade/persistence/models.py
```diff
@@ -257,6 +257,7 @@ class LocalTrade():
     sell_reason: str = ''
     sell_order_status: str = ''
     strategy: str = ''
+    buy_tag: Optional[str] = None
```
xmatthias, line 260: You might wanna add this to the `to_json()` of this method, too - so it's available via API as well.

Note:
to have it available via API, you'll ALSO need to add it to the proper `api_schema.py` schema, otherwise it'll be hidden from the API.

kevinjulian: done 🙏 anything else for this PR?

docs/strategy-advanced.md
```diff
@@ -114,6 +114,32 @@ class AwesomeStrategy(IStrategy):
 
 See [Dataframe access](#dataframe-access) for more information about dataframe use in strategy callbacks.
 
+## Buy Tag
+
+When your strategy has multiple buy signal, you can name it.
```
```suggestion
When your strategy has multiple buy signals, you can name the signal that triggered.
```

docs/strategy-advanced.md
```diff
@@ -114,6 +114,32 @@ class AwesomeStrategy(IStrategy):
...
+Then you can access you buy signal on `custom_sell`
+
+```python
+def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+    dataframe.loc[
+        (
+            (dataframe['rsi'] < 35) &
+            (dataframe['volume'] > 0)
+        ),
+        ['buy', 'buy_tag']] = 1, 'buy_signal_rsi'
```
xmatthias: I'd make it clear this is supposed to be a tuple.
the comma might be easily missed ...

```suggestion
        ['buy', 'buy_tag']] = (1, 'buy_signal_rsi')
```

freqtrade/persistence/models.py
```diff
@@ -703,6 +705,7 @@ class Trade(_DECL_BASE, LocalTrade):
     sell_reason = Column(String(100), nullable=True)
     sell_order_status = Column(String(100), nullable=True)
     strategy = Column(String(100), nullable=True)
+    buy_tag = Column(String(100), nullable=True)
```
xmatthias, line 708: I'd think the max-length should be mentioned in the docs to avoid surprises (a quick "is limited to 100 characters, remaining data will be truncated" should be sufficient.

freqtrade/strategy/interface.py
```diff
@@ -522,18 +528,20 @@ def get_signal(self, pair: str, timeframe: str, dataframe: DataFrame) -> Tuple[b
                 'Outdated history for pair %s. Last tick is %s minutes old',
                 pair, int((arrow.utcnow() - latest_date).total_seconds() // 60)
             )
-            return False, False
+            return False, False, None
 
-        (buy, sell) = latest[SignalType.BUY.value] == 1, latest[SignalType.SELL.value] == 1
+        (buy, sell, buy_tag) = latest[SignalType.BUY.value] == 1,\
+            latest[SignalType.SELL.value] == 1,\
+            latest.get(SignalTagType.BUY_TAG.value, None)
```
xmatthias: Please split this to 3 assignments.
This is hard to read / understand now (with the wrapped lines) - and it'll implicitly create a tuple and then explode it again ... which is really not necessary.

tests/optimize/test_backtest_detail.py
```diff
@@ -489,12 +489,12 @@
...
-    [1, 5000, 5500, 5000, 4900, 6172, 0, 0],    # enter trade (signal on last candle) and stop
-    [2, 4900, 5250, 4500, 5100, 6172, 0, 0],
-    [3, 5100, 5100, 4650, 4750, 6172, 0, 0],
-    [4, 4750, 4950, 4350, 4750, 6172, 0, 0]],
+    # D   O     H     L     C    V    B  S  SN
+    [0, 5000, 5050, 4950, 5000, 6172, 1, 0, ''],
+    [1, 5000, 5500, 5000, 4900, 6172, 0, 0, ''],    # enter trade (signal on last candle) and stop
+    [2, 4900, 5250, 4500, 5100, 6172, 0, 0, ''],
+    [3, 5100, 5100, 4650, 4750, 6172, 0, 0, ''],
+    [4, 4750, 4950, 4350, 4750, 6172, 0, 0, '']],
```
xmatthias: I'm not certain why these changes are necessary.
i understand the need to add this "in general" to have it tested as well (That's `tc33`) - but i don't think it shold be needed to have it as "empty" columns - as the logic should be able to handle "not-assignment" of the buy-tag, too.

tests/rpc/test_rpc_apiserver.py
```diff
@@ -463,13 +463,13 @@ def test_api_count(botclient, mocker, ticker, fee, markets):
...
 
-    ftbot.config['max_open_trades'] = float('inf')
+    freqtrade.config['max_open_trades'] = float('inf')
     rc = client_get(client, f"{BASE_URI}/count")
     assert rc.json()["max"] == -1
 
 
 def test_api_locks(botclient):
-    ftbot, client = botclient
+    freqtrade, client = botclient
```
xmatthias: While i apreciate you taking the time to align all our namings - for next time (should you notice something similar) please do this as a seperate PR (simply extract this part as a single commit on top of develop and do a "quick" PR).

That way, it'll become clear which changes are actually caused by the change / feature / ... you're suggesting / adding - and which are simple refactorings / renamings (which will in term make reviewing it easier).

Now i've looked through it though - so for this PR, leave it in please :+1:

freqtrade/freqtradebot.py
```diff
@@ -590,6 +595,7 @@ def _notify_buy(self, trade: Trade, order_type: str) -> None:
         msg = {
             'trade_id': trade.id,
             'type': RPCMessageType.BUY,
+            'buy_tag': trade.buy_tag,
```
xmatthias, line 598: The addition of this should also be added to `webhook-config.md` - to document for which hooks this is available (it will be available implicitly by adding it here, just noone will know about it).

Obviously applies to all messages that have this field added ...

xmatthias (commented):
While i did do a review (read through it) now (some feedback - nothing bad i'd say) - i'll want to give this a spin too (just to be sure / feel this).

I'll however not have time for that until mid/end next week - so please bear with me.

rokups:
It just occurred to me that me and @nightshift2k started working on a feature that overlaps with this one. We started on implementing persistent metadata, both per-trade and global. It introduces a key-value table, which preserves original type in common case (bool/int/float/str) and serializes dicts/lists to json should they be present. Metadata feature would cover this usecase and more. Initial work is here: https://github.com/nightshift2k/freqtrade/commit/c99274e321b2d6615c7abeea3c4f6b4f455182ec
In the end it should look like this: user sets values to `trade.metadata` dict and they persist across restarts. We should probably coordinate how to proceed.

kevinjulian:
> It just occurred to me that me and @nightshift2k started working on a feature that overlaps with this one. We started on implementing persistent metadata, both per-trade and global. It introduces a key-value table, which preserves original type in common case (bool/int/float/str) and serializes dicts/lists to json should they be present. Metadata feature would cover this usecase and more. Initial work is here: [nightshift2k@c99274e](https://github.com/nightshift2k/freqtrade/commit/c99274e321b2d6615c7abeea3c4f6b4f455182ec)
> In the end it should look like this: user sets values to `trade.metadata` dict and they persist across restarts. We should probably coordinate how to proceed.


cool, that's really interesting 👏
I think when we have key_val table, we need to clarify the boundary
- what kind of data will save to key_val table
- what kind of data will save to normal table
to avoid mess up data

before that, I prefer to have some kind of `post_enter_trade` or `post_exit_trade` callback (true, we have webhook already), of course will be easy to use when we have it on `IStrategy`

imho it's hard to save/modify trade not because the database architecture, but because there's no callback method on `IStrategy` on the other side, it's good also - the interface so simple and clear

rokups:
Our change is intended to only save custom data produced by the user. Freqtrade would not use it internally to store any state. There is `confirm_trade_exit` for storing any info when trade closes, but there is nothing to work with just opened trade. `confirm_trade_entry` is invoked before trade is created so we do not have a trade object passed to that function. So yeah an additional callback would be needed.

xmatthias:
I'm not convinced on moving this to a separate callback - that'll make it pretty complex for the user.
That may be fine for additional stuff (although i'm not sure what else youd store that you can't get from the dataframe) ...
, but for something this simple, i'm not 100% certain a builtin way (as proposed here) wont be better.

rokups:
We can't get anything from dataframe after 1000 candles. That is 3 and a half days for 5m candles, which is not much. And another callback is merely to poke stuff into trade metadata on trade open. We could use a custom_stoploss or custom_sell while checking if metadata is empty, but it's kind of stupid. Note that callback is not supposed to return anything, it's just a hook point to do anything after trade has just opened.

xmatthias:
I'm not arguing against that, just saying that for a common usecase (like buy signal tagging) THIS approach (without callback) will be easier to use from a user perspective.
Obviously if you wanna store "whatever" the callback should be used.

freqtrade/persistence/migrations.py
```diff
@@ -167,6 +168,13 @@ def check_migrate(engine, decl_base, previous_tables) -> None:
         inspector = inspect(engine)
         cols = inspector.get_columns('trades')
 
+    if not has_column(cols, 'buy_tag'):
+        logger.info(f'Running database migration for trades - backup: {table_back_name}')
+        migrate_trades_table(decl_base, inspector, engine, table_back_name, cols)
+        # Reread columns - the above recreated the table!
+        inspector = inspect(engine)
+        cols = inspector.get_columns('trades')
+
```
xmatthias: This is wrong and kinda dangerous (if you're comming from an older version, migrations might run twice).

I've fixed this now - please look at that commit to see how easy this would've been :D

xmatthias (approved):
I've fixed one last minor thing (migrations) - maybe you wanna have a look at this as the change would've been pretty simple (simpler than what you had).

LGTM :+1:

Issue #5253, opened 2021-07-09
Installation failing on Ubuntu 20.04 on Raspberry pi


meauxt:
Unable to install Freqtrade on ubuntu 20.04 ( HW Raspberry pi 4 8GB)
## Describe your environment

  * Operating system: Ubuntu 20.04 Server for raspberry pi 
  * Python Version: 3.9 (`python -V`)

  
## Your question
Running ./setup.sh --install failing with the error below
```

Collecting tables==3.6.1
  Using cached tables-3.6.1.tar.gz (4.6 MB)
    ERROR: Command errored out with exit status 1:
     command: /home/ubuntu/freqtrade/.env/bin/python3.9 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-4lqe5mby/tables_fc9047b0a7c1409cb0e0d9e44bd0a7cd/setup.py'"'"'; __file__='"'"'/tmp/pip-install-4lqe5mby/tables_fc9047b0a7c1409cb0e0d9e44bd0a7cd/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-zp9lfx0m
         cwd: /tmp/pip-install-4lqe5mby/tables_fc9047b0a7c1409cb0e0d9e44bd0a7cd/
    Complete output (12 lines):
    /tmp/H5closesqczs4q1.c: In function ‘main’:
    /tmp/H5closesqczs4q1.c:2:5: warning: implicit declaration of function ‘H5close’ [-Wimplicit-function-declaration]
        2 |     H5close();
          |     ^~~~~~~
    /usr/bin/ld: cannot find -lhdf5
    collect2: error: ld returned 1 exit status
    * Using Python 3.9.5 (default, May 11 2021, 08:20:37)
    * USE_PKGCONFIG: True
    .. ERROR:: Could not find a local HDF5 installation.
       You may need to explicitly state where your local HDF5 headers and
       library can be found by setting the ``HDF5_DIR`` environment
       variable or by using the ``--hdf5`` command-line option.
    ----------------------------------------
[...]

PR #5300, opened 2021-07-19
Fixed setup for python3.9 on ubuntu

samgermain:
## Summary
The setup script installs everything properly on Ubuntu20 with python3.9

Solve the issue: #5253 

## Quick changelog

- installs pip if pip is not installed
- installs python specific version of libpython3.*-dev python3.*-venv
- defaults to installing python3.9 before python3.8

## Can test by
* Starting a new ec2 t2.micro instance running Ubuntu20
* ssh into the instance
* run the following commands

```
sudo apt-get update
sudo apt-get install python3.9
git clone --single-branch --branch ubuntu-setup-fix https://github.com/samgermain/freqtrade
cd freqtrade
./setup.sh -i
```

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -4,8 +4,12 @@
 function check_installed_pip() {
    ${PYTHON} -m pip > /dev/null
    if [ $? -ne 0 ]; then
-        echo "pip not found (called as '${PYTHON} -m pip'). Please make sure that pip is available for ${PYTHON}."
-        exit 1
+        echo "-----------------------------"
+        echo "Installing Pip for ${PYTHON}"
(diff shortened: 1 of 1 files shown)


waresz:
Work fine with centos 7 and python 3.9.6

samgermain:
I'm unfamiliar with how to use Debian, and haven't tested it on that.
If someone would like to test this out on Debian, that would be great

xmatthias (approved):
considering Ubuntu is debian under the hood, this will work just fine on debian, too (unless you use PPA's, which you don't).

LGTM :+1:

PR #5312, opened 2021-07-23
Fix code to get Bittrex US-restricted markets

raphant:

## Summary
Old code to grab Bittrex US-restricted markets was no longer working.

Changed
```python
res = [ f"{x['MarketCurrency']}/{x['BaseCurrency']}" for x in ct.publicGetMarkets()['result'] if x['IsRestricted']]
```

to
```python
res = [f"{x['quoteCurrencySymbol']}/{x['baseCurrencySymbol']}" for x in ct.publicGetMarkets() if 'US' in x['prohibitedIn']]
```

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -77,8 +77,9 @@ You can get a list of restricted markets by using the following snippet:
 ``` python
 import ccxt
 ct = ccxt.bittrex()
-ct.load_markets()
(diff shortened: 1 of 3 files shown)


xmatthias (approved):
I've made a slight adjustment to your fix - using the already loaded markets (it's now included there).
also, your pairs were inverted ... resulting in BTC/1INCH - not 1INCH/BTC - which will be a problem once you try to add this list to your blacklist.

However, great catch that the exchange now renamed this field ... it's also relevant to bot-internal code - which did use IsRestricted before.

Issue #5316, opened 2021-07-24
PR #5284 causing a spammy console log


samgermain:
## Describe your environment

  * Operating system: MacOS Big Sur 11.4
  * Python Version: Python 3.9.5 
  * CCXT version: ccxt==1.52.83
  * Freqtrade Version: freqtrade develop-b84a1d0c
  
## Describe the problem:

In the PR I mentioned that these changes were made
I'm not 100% sure if all these changes should be made, and people were complaining about a [spammy console log,](https://i.imgur.com/RPN1b0Y.png) so I reversed the second and third change in this table and will push a PR right away.

<table style="width:100%">
  <tr>
    <th>get_buy_rate</th>
    <th>get_sell_rate</th> 
    <th>update</th>
  </tr>
  <tr>
    <td>if bid_strategy.get('use_order_book', False) and ('use_order_book' in bid_strategy):</td>
    <td>if ask_strategy.get('use_order_book', False)</td>
    <td>if strategy.get('use_order_book', False) and ('use_order_book' in strategy) is now used for both</td>
  </tr>
  <tr>
    <td>logger.info(f"{name} price from orderbook {bid_strategy['price_side'].capitalize()}"
                        f"side - top {order_book_top} order book buy rate {rate:.8f}")</td>
    <td>/</td>
    <td>Both the bid_strategy and ask_strategy have this logger message now</td>
  </tr>
    <td>logger.info(f"Using Last {strategy['price_side'].capitalize()} / Last Price")</td>
    <td>/</td>
    <td>Both the bid_strategy and ask_strategy have this logger message now</td>
  </tr>
  <tr>
     <td>/</td>
     <td>if rate is None: raise PricingError(f"{name}-Rate for {pair} was empty.")</td>
     <td>Both raise this pricing error now</td>
  </tr>

</table>

### Steps to reproduce:

run `freqtrade trade`
  
### Observed Results:

https://i.imgur.com/RPN1b0Y.png

------------------

@xmatthias Maybe just double check the changes in the table and determine if they are all good changes

xmatthias:
Well you closed this yourself in #5317 👍 💯

PR #5308, opened 2021-07-22
Reduce the spamming the log file with the same messages a few times a second. 

lazydroid:
## Summary

Sam Germain on 2021-07-19 introduced changes that post the same message to the log / screen, this PR temporarily removes those messages to reduce the log size.

## Quick changelog

- commented out line 1041, 1042
- commented out line 1044

## What's new?
This does not happen anymore:
```
2021-07-22 15:01:13,223 - freqtrade.exchange.exchange - INFO - Using Last Bid / Last Price
2021-07-22 15:01:13,720 - freqtrade.exchange.exchange - INFO - Using Last Bid / Last Price
2021-07-22 15:01:19,265 - freqtrade.exchange.exchange - INFO - Using Last Bid / Last Price
2021-07-22 15:01:19,450 - freqtrade.exchange.exchange - INFO - Using Last Bid / Last Price
2021-07-22 15:01:19,946 - freqtrade.exchange.exchange - INFO - Using Last Bid / Last Price
2021-07-22 15:01:20,446 - freqtrade.exchange.exchange - INFO - Using Last Bid / Last Price
2021-07-22 15:01:20,948 - freqtrade.exchange.exchange - INFO - Using Last Bid / Last Price
```

PR #5317, opened 2021-07-24
Log level set to debug on exchange.get_rate

samgermain:
## Summary
Fixes spammy/overfilled console log

Solve the issue: #5316

## Quick changelog

- check if the price side is `buy` in `exchange.get_rate` before logging these lines
```
logger.info(f"{name} price from orderbook {price_side}"
                            f"side - top {order_book_top} order book {side} rate {rate:.8f}")
...
logger.info(f"Using Last {conf_strategy['price_side'].capitalize()} / Last Price")
```

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1037,14 +1037,11 @@ class Exchange:
                     f"determined. Orderbook: {order_book}"
                 )
                 raise PricingError from e
-
-            if side == "buy":
-                price_side = {conf_strategy['price_side'].capitalize()}
-                logger.info(f"{name} price from orderbook {price_side}"
-                            f"side - top {order_book_top} order book {side} rate {rate:.8f}")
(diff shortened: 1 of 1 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -1038,10 +1038,13 @@ def get_rate(self, pair: str, refresh: bool, side: str) -> float:
                 )
                 raise PricingError from e
 
-            logger.info(f"{name} price from orderbook {conf_strategy['price_side'].capitalize()}"
```
xmatthias, line 1041: Simply change loglevel to debug (i also mentioned this in #5308).
Info-level is potentially ok for buying (but not necessarily, as in some instances it can also be spammy) - and spammy on the sell side.

I'd not complicate this just for the logging, especially not if the outcome is potentially still spammy (like when you reject the buy in confirm_trade_entry).

xmatthias (approved):
Great thanks! 👍

PR #5318, opened 2021-07-25
combined exchange.buy and exchange.sell

samgermain:
## Summary
Remove exchange.buy and exchange.sell and replace with exchange.create_order

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -689,7 +689,16 @@ class Exchange:
     # Order handling
 
     def create_order(self, pair: str, ordertype: str, side: str, amount: float,
-                     rate: float, params: Dict = {}) -> Dict:
+                     rate: float, time_in_force: str = 'gtc') -> Dict:
+
+        if self._config['dry_run']:
+            dry_order = self.create_dry_run_order(pair, ordertype, side, amount, rate)
+            return dry_order
+
+        params = self._params.copy()
+        if time_in_force != 'gtc' and ordertype != 'market':
+            params.update({'timeInForce': time_in_force})
+
         try:
             # Set the precision for amount and price(rate) as accepted by the exchange
             amount = self.amount_to_precision(pair, amount)
@@ -719,32 +728,6 @@ class Exchange:
                 f'Could not place {side} order due to {e.__class__.__name__}. Message: {e}') from e
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
-
-    def buy(self, pair: str, ordertype: str, amount: float,
-            rate: float, time_in_force: str) -> Dict:
-
-        if self._config['dry_run']:
-            dry_order = self.create_dry_run_order(pair, ordertype, "buy", amount, rate)
-            return dry_order
-
-        params = self._params.copy()
-        if time_in_force != 'gtc' and ordertype != 'market':
-            params.update({'timeInForce': time_in_force})
-
-        return self.create_order(pair, ordertype, 'buy', amount, rate, params)
-
-    def sell(self, pair: str, ordertype: str, amount: float,
-             rate: float, time_in_force: str = 'gtc') -> Dict:
-
-        if self._config['dry_run']:
-            dry_order = self.create_dry_run_order(pair, ordertype, "sell", amount, rate)
-            return dry_order
-
-        params = self._params.copy()
-        if time_in_force != 'gtc' and ordertype != 'market':
-            params.update({'timeInForce': time_in_force})
-
-        return self.create_order(pair, ordertype, 'sell', amount, rate, params)
 
     def stoploss_adjust(self, stop_loss: float, order: Dict) -> bool:
         """
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -2190,7 +2190,7 @@ def test_cancel_order_dry_run(default_conf, mocker, exchange_name):
     assert exchange.cancel_order(order_id='123', pair='TKN/BTC') == {}
     assert exchange.cancel_stoploss_order(order_id='123', pair='TKN/BTC') == {}
 
-    order = exchange.create_order('ETH/BTC', 'limit', "sell", 5, 0.55, 'gtc')
+    order = exchange.create_order('ETH/BTC', 'limit', "buy", 5, 0.55, 'gtc')
 
     cancel_order = exchange.cancel_order(order_id=order['id'], pair='ETH/BTC')
     assert order['id'] == cancel_order['id']
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -1056,8 +1056,8 @@ def test_buy_dry_run(default_conf, mocker):
     default_conf['dry_run'] = True
     exchange = get_patched_exchange(mocker, default_conf)
 
-    order = exchange.buy(pair='ETH/BTC', ordertype='limit',
-                         amount=1, rate=200, time_in_force='gtc')
+    order = exchange.create_order(pair='ETH/BTC', ordertype='limit', side="buy",
+                                  amount=1, rate=200, time_in_force='gtc')
     assert 'id' in order
     assert 'dry_run_buy_' in order['id']
 
@@ -1080,8 +1080,8 @@ def test_buy_prod(default_conf, mocker, exchange_name):
     mocker.patch('freqtrade.exchange.Exchange.price_to_precision', lambda s, x, y: y)
     exchange = get_patched_exchange(mocker, default_conf, api_mock, id=exchange_name)
 
-    order = exchange.buy(pair='ETH/BTC', ordertype=order_type,
-                         amount=1, rate=200, time_in_force=time_in_force)
+    order = exchange.create_order(pair='ETH/BTC', ordertype=order_type, side="buy",
+                                  amount=1, rate=200, time_in_force=time_in_force)
 
     assert 'id' in order
     assert 'info' in order
@@ -1094,9 +1094,10 @@ def test_buy_prod(default_conf, mocker, exchange_name):
 
     api_mock.create_order.reset_mock()
     order_type = 'limit'
-    order = exchange.buy(
+    order = exchange.create_order(
         pair='ETH/BTC',
         ordertype=order_type,
+        side="buy",
         amount=1,
         rate=200,
         time_in_force=time_in_force)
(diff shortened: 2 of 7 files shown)


tests/test_freqtradebot.py
```diff
@@ -970,8 +970,10 @@ def test_handle_stoploss_on_exchange(mocker, default_conf, fee, caplog,
             'ask': 0.00001173,
             'last': 0.00001172
         }),
-        buy=MagicMock(return_value={'id': limit_buy_order['id']}),
-        sell=MagicMock(return_value={'id': limit_sell_order['id']}),
+        create_order=MagicMock(side_effect=[
+            (return_value={'id': limit_buy_order['id']}),
```
xmatthias: Remove the return_value. An array as side_effect acts as return value unless the value returned is an exception.

samgermain:
So I've almost got it. 14 tests fail with

```
    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: orders.ft_pair, orders.order_id
E       [SQL: INSERT INTO orders (ft_trade_id, ft_order_side, ft_pair, ft_is_open, order_id, status, symbol, order_type, side, price, average, amount, filled, remaining, cost, order_date, order_filled_date, order_update_date) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)]
E       [parameters: (1, 'sell', 'ETH/BTC', 1, 'mocked_limit_buy', 'open', 'mocked', 'limit', 'buy', 1.099e-05, None, 90.99181073, 0.0, 90.99181073, 0.0009999, '1970-01-19 20:01:02.078000', None, '2021-07-26 01:14:38.263966')]
E       (Background on this error at: http://sqlalche.me/e/14/gkpj)

.env/lib/python3.9/site-packages/sqlalchemy/engine/default.py:717: IntegrityError
```

not sure how to resolve this on this one, I've solved this error before, but not in this context


-------------------


Here's one test that get's this error


```
    def test_disable_ignore_roi_if_buy_signal(default_conf, limit_buy_order, limit_buy_order_open,
                                              fee, mocker) -> None:
        patch_RPCManager(mocker)
        patch_exchange(mocker)
        mocker.patch.multiple(
            fetch_ticker=MagicMock(return_value={
                'bid': 0.00000172,
                'ask': 0.00000173,
                'last': 0.00000172
            }),
            create_order=MagicMock(return_value=limit_buy_order_open),
            get_fee=fee,
            _is_dry_limit_order_filled=MagicMock(return_value=False),
        )
        default_conf['ask_strategy'] = {
            'ignore_roi_if_buy_signal': False
        }
        freqtrade = FreqtradeBot(default_conf)
        patch_get_signal(freqtrade)
        freqtrade.strategy.min_roi_reached = MagicMock(return_value=True)
    
        freqtrade.enter_positions()
    
        trade = Trade.query.first()
        trade.update(limit_buy_order)
        # Sell due to min_roi_reached
        patch_get_signal(freqtrade, value=(True, True))
>       assert freqtrade.handle_trade(trade) is True

tests/test_freqtradebot.py:3549:
```

xmatthias:
without playing around with it - i'd say by mocking `create_order` (before the mock was only on `buy`) - you're now also mocking the sell implicitly.
By returning only one value - you're getting an error on insert as the order-id is identical.

The exception itself got little to do with the test, but breaks integrity nonetheless.

As sell wasn't mocked before - i assume that was returned by the ccxt api mock - which will not insert into the order-table as it lacks the id field.
Not sure how to fix that (that'll need a bit of experimentation) - either use side-effect on the create order mock and mock all calls to it via an Array - or mock the "insert to database" for the orders table - but that's the "dirty" solution which we probably shouldn't be using.

samgermain:
> without playing around with it - i'd say by mocking `create_order` (before the mock was only on `buy`) - you're now also mocking the sell implicitly.
> By returning only one value - you're getting an error on insert as the order-id is identical.
> 
> The exception itself got little to do with the test, but breaks integrity nonetheless.
> 
> As sell wasn't mocked before - i assume that was returned by the ccxt api mock - which will not insert into the order-table as it lacks the id field.
> Not sure how to fix that (that'll need a bit of experimentation) - either use side-effect on the create order mock and mock all calls to it via an Array - or mock the "insert to database" for the orders table - but that's the "dirty" solution which we probably shouldn't be using.

One last test is still failing

```
def test_disable_ignore_roi_if_buy_signal(default_conf, limit_buy_order, limit_buy_order_open,
                                          fee, mocker) -> None:
    patch_RPCManager(mocker)
    patch_exchange(mocker)
    mocker.patch.multiple(
        fetch_ticker=MagicMock(return_value={
            'bid': 0.00000172,
            'ask': 0.00000173,
            'last': 0.00000172
        }),
        create_order=MagicMock(side_effect=[
            limit_buy_order_open,
            {'id': 31212341241},
        ]),
        get_fee=fee,
        _is_dry_limit_order_filled=MagicMock(return_value=False),
    )
    default_conf['ask_strategy'] = {
        'ignore_roi_if_buy_signal': False
    }
    freqtrade = FreqtradeBot(default_conf)
    patch_get_signal(freqtrade)
    freqtrade.strategy.min_roi_reached = MagicMock(return_value=True)

    freqtrade.enter_positions()

    trade = Trade.query.first()
    trade.update(limit_buy_order)
    # Sell due to min_roi_reached
    patch_get_signal(freqtrade, value=(True, True))
    assert freqtrade.handle_trade(trade) is True

    # Test if buy-signal is absent
    patch_get_signal(freqtrade, value=(False, True))
    assert freqtrade.handle_trade(trade) is True
    assert trade.sell_reason == SellType.SELL_SIGNAL.value
```

When I add a dummy mock for the second exchange.create_order call, the second last assertion errors with

```
    
        effect = self.side_effect
        if effect is not None:
            if _is_exception(effect):
                raise effect
            elif not _callable(effect):
E               StopIteration

/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1154: StopIteration
```

and if I exclude that assertion, then the last assertion fails with

```
>       assert trade.sell_reason == SellType.SELL_SIGNAL.value
E       AssertionError: assert 'stop_loss' == 'sell_signal'
E         - sell_signal
E         + stop_loss
```
---------------
I also tried using `limit_sell_order`, `limit_sell_order_open`, and `market_sell_order` instead of the dummy `{'id': 31212341241}` with no  luck

Could you elaborate  on mock inserting into the database if you think that would be a fix? What method would do that?

xmatthias:
Best use vscode (or probably pycharm) - set a stop-point in `handle_trade` (or where create_order with sell is called) - to check for this in detail (in cases where tests fail after such a refactoring, i often debug the same test in develop to see how it should've behaved).

a small hint: a `StopIteration` is basically always an iterator that ran out of values (think about it as a keyerror where you access with `key[key.length + 1]`).
As such - it's pretty clear that create_order is called 3 times in this test, not 2 (`handle_trade` is called twice - so you'll need 2 sell orders).

I think you can get away with returning "MagicMock" instead of the 'id' object (that's what the ccxt mock did) ... but i'm not 100% sure about that, so you'll have to give that a try

samgermain:
All the tests pass locally

tests/test_freqtradebot.py
```diff
@@ -1877,7 +1898,10 @@ def test_handle_overlapping_signals(default_conf, ticker, limit_buy_order_open,
     mocker.patch.multiple(
         'freqtrade.exchange.Exchange',
         fetch_ticker=ticker,
-        buy=MagicMock(return_value=limit_buy_order_open),
+        create_order=MagicMock(side_effect=[
+            limit_buy_order_open,
+            {'id': 1234553382},
+        ]),
```
xmatthias, line 1904: These instances is where having these as seperate methods simplified testing.

I'm quite worried that this change will decrease maintainability (while it removes a lot of code duplicates, it also complicates tests by some - as we now need to mock both buy and selling even if we're not interrested in one side).

freqtrade/exchange/exchange.py
```diff
@@ -689,7 +689,16 @@ def fetch_dry_run_order(self, order_id) -> Dict[str, Any]:
...
+                     rate: float, time_in_force: str = 'gtc') -> Dict:
+
+        if self._config['dry_run']:
+            dry_order = self.create_dry_run_order(pair, ordertype, side, amount, rate)
+            return dry_order
+
+        params = self._params.copy()
+        if time_in_force != 'gtc' and ordertype != 'market':
+            params.update({'timeInForce': time_in_force})
+
```
samgermain, line 701: I think this change is useful, we can exclude the rest of this PR, but I think this change should bee made. It doesn't really add any benefit to leverage/shorting though so I'm not going to do it. I'll close the PR, rollback these commits locally, and give this no more thought

tests/test_freqtradebot.py
samgermain: Yeah I've been thinking this PR might be unnecessary. It didn't really add as much benefit as I was expecting either, I think that I only call it twice within the actual program. I'll close this PR

xmatthias: I honestly think closing the PR is a bit premature ... (if i'd reject it - i'd be clear about it - even before you complete it).

I'm worried doesn't mean i'm totally against it - or against most parts - more that i need to think about it more / would like to discuss :smile: .

What i really like is the reduction of duplicate code by moving it to `create_order`.

I'm also still not convinced about using it "as is" ...  as such - i'll reject your closing of this PR for now.

while the changes to the tests where quite some - i'm torn if it's a bad thing to have to mock them explicitly as it makes the intend (and what should be called in the test) clear - or if it's going to complicate our lifes further down the road.

tests/exchange/test_exchange.py
```diff
@@ -2186,7 +2190,7 @@ def test_cancel_order_dry_run(default_conf, mocker, exchange_name):
     assert exchange.cancel_order(order_id='123', pair='TKN/BTC') == {}
     assert exchange.cancel_stoploss_order(order_id='123', pair='TKN/BTC') == {}
 
-    order = exchange.buy('ETH/BTC', 'limit', 5, 0.55, 'gtc')
+    order = exchange.create_order('ETH/BTC', 'limit', "sell", 5, 0.55, 'gtc')
```
xmatthias: This should be buy, not sell (according to the old version) - although for this test, the side is not really relevant.

```suggestion
    order = exchange.create_order('ETH/BTC', 'limit', "buy", 5, 0.55, 'gtc')
```

xmatthias (approved):
After sleeping over this for a few days, i think the code simplification is worth the effort on tests we might have.
It'll also make the intend in tests clear, as we're now forced to mock all orders that are created.

LGTM :+1:

Issue #5334, opened 2021-07-27
Bot refuses to make a sell in live run. No issues with sells in dry run.


JasonObihara:
## Describe your environment

  * Operating system: ____ubuntu linux vps
  * Python Version: _____ (`python -V`)
  * CCXT version: _____ (`pip freeze | grep ccxt`)
  * Freqtrade Version: ____ (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
  
## Your question
The bot runs, and it's on a bullish run.  However when it's time for sell, the bot says not enough money to make the sell. 
Attaching screen shot
 freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:48,418 - freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:48,666 - freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:49,165 - freqtrade.freqtradebot - INFO - Executing Sell for DOT/BTC. Reason: sell_signal
2021-07-28 00:33:49,690 - freqtrade.wallets - INFO - Wallets synced.
2021-07-28 00:33:49,690 - freqtrade.freqtradebot - WARNING - Unable to sell trade DOT/BTC: Not enough amount to sell. Trade-amount: 82.7, Wallet: 0
2021-07-28 00:33:49,694 - freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:50,170 - freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:50,675 - freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:53,411 - freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:53,664 - freqtrade.exchange.exchange - INFO - Using Last Ask / Last Price
2021-07-28 00:33:54,158 - freqtrade.freqtradebot - INFO - Executing Sell for DOT/BTC. Reason: sell_signal

I used the orderbook for buying at bid, but didn't use the order book for sales. 

*Ask the question you have not been able to find an answer in our [Documentation](https://www.freqtrade.io/en/latest/)*

samgermain:
Can you fill these out please

```

    Operating system: ____ubuntu linux vps
    Python Version: _____ (python -V)
[...]

PR #5284, opened 2021-07-18
Merge get_buy_rate and get_sell_rate

samgermain:
## Summary

Merges `exchange.get_buy_rate` and `exchange.get_sell_rate` into one method `exchange.get_rate`

## What's new?

The methods were almost the same. 
When they were merged, these were some changes that were made to the logic

<table style="width:100%">
  <tr>
    <th>get_buy_rate</th>
    <th>get_sell_rate</th> 
    <th>update</th>
  </tr>
  <tr>
    <td>if bid_strategy.get('use_order_book', False) and ('use_order_book' in bid_strategy):</td>
    <td>if ask_strategy.get('use_order_book', False)</td>
    <td>if strategy.get('use_order_book', False) and ('use_order_book' in strategy) is now used for both</td>
  </tr>
  <tr>
    <td>logger.info(f"{name} price from orderbook {bid_strategy['price_side'].capitalize()}"
                        f"side - top {order_book_top} order book buy rate {rate:.8f}")</td>
    <td>/</td>
    <td>Both the bid_strategy and ask_strategy have this logger message now</td>
  </tr>
    <td>logger.info(f"Using Last {strategy['price_side'].capitalize()} / Last Price")</td>
    <td>/</td>
    <td>Both the bid_strategy and ask_strategy have this logger message now</td>
  </tr>
  <tr>
     <td>/</td>
     <td>if rate is None: raise PricingError(f"{name}-Rate for {pair} was empty.")</td>
     <td>Both raise this pricing error now</td>
  </tr>

</table>

PR #5335, opened 2021-07-28
Added checks for side to for PricingError and checking for use_order_…

samgermain:
## Summary
Solve the issue: #5334 

## Quick changelog

In `exchange.get_rate`

```
if conf_strategy.get('use_order_book', False) and ('use_order_book' in conf_strategy):
```

was changed to

```
dont_use_order_book = (
        ('use_order_book' in conf_strategy or side == "sell") and
        conf_strategy.get('use_order_book', False)
)

if dont_use_order_book:
```

and

```
if rate is None:
            raise PricingError(f"{name}-Rate for {pair} was empty.")
```

was changed to 

```
if rate is None and side == "sell":
            raise PricingError(f"Sell-Rate for {pair} was empty.")
 ```

## What's new?
In PR #5284 these changes were made when merging the 2 methods

<table style="width:100%">
  <tr>
    <th>get_buy_rate</th>
    <th>get_sell_rate</th> 
    <th>update</th>
  </tr>
  <tr>
    <td>if bid_strategy.get('use_order_book', False) and ('use_order_book' in bid_strategy):</td>
    <td>if ask_strategy.get('use_order_book', False)</td>
    <td>if strategy.get('use_order_book', False) and ('use_order_book' in strategy) is now used for both</td>
  </tr>
  <tr>
     <td>/</td>
     <td>if rate is None: raise PricingError(f"{name}-Rate for {pair} was empty.")</td>
     <td>Both raise this pricing error now</td>
  </tr>
</table>

Both changes are reverted in this PR. I'm guessing that reverting the first change will probably fix the issue  #5334 
I reverted the second change because I assume that the original `get_buy_rate` strategy didn't check for this for a reason given that the other changes have caused problems

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1022,8 +1022,12 @@ class Exchange:
                 return rate
 
         conf_strategy = self._config.get(strat_name, {})
+        use_order_book = (
+            ('use_order_book' in conf_strategy or side == "sell") and
+            conf_strategy.get('use_order_book', False)
+        )
 
-        if conf_strategy.get('use_order_book', False) and ('use_order_book' in conf_strategy):
+        if use_order_book:
 
             order_book_top = conf_strategy.get('order_book_top', 1)
             order_book = self.fetch_l2_order_book(pair, order_book_top)
@@ -1053,8 +1057,8 @@ class Exchange:
                     ticker_rate = ticker_rate - balance * (ticker_rate - ticker['last'])
             rate = ticker_rate
 
-        if rate is None:
-            raise PricingError(f"{name}-Rate for {pair} was empty.")
+        if rate is None and side == "sell":
+            raise PricingError(f"Sell-Rate for {pair} was empty.")
         cache_rate[pair] = rate
 
         return rate

### CODE REVIEW

**Comment by samgermain** [comment]
I'm trying to find a strategy that will quickly make a buy and a sell in live mode so I can test out that everything is working fine for sure

**Comment by samgermain** [comment]
I think I got blocked by binance for sending too many requests.  I was on the 1m timeframe and had "max_open_trades": 1,

File: freqtrade/exchange/exchange.py [hunk 1]
```diff
@@ -1053,8 +1057,8 @@ def get_rate(self, pair: str, refresh: bool, side: str) -> float:
                     ticker_rate = ticker_rate - balance * (ticker_rate - ticker['last'])
             rate = ticker_rate
 
-        if rate is None:
-            raise PricingError(f"{name}-Rate for {pair} was empty.")
+        if rate is None and side == "sell":
+            raise PricingError(f"Sell-Rate for {pair} was empty.")
```
xmatthias, line 1061: Both buy and sell methods did raise a `PricingError` before - and can raise a pricing-error now as well.

As such - i really don't understand this change -n as it's changing behavior over the previous version.

freqtrade/exchange/exchange.py
```diff
@@ -1022,8 +1022,12 @@ def get_rate(self, pair: str, refresh: bool, side: str) -> float:
                 return rate
 
         conf_strategy = self._config.get(strat_name, {})
+        dont_use_order_book = (
+            ('use_order_book' in conf_strategy or side == "sell") and
+            conf_strategy.get('use_order_book', False)
+        )
 
-        if conf_strategy.get('use_order_book', False) and ('use_order_book' in conf_strategy):
+        if dont_use_order_book:
```
xmatthias: The naming (`dont_use_order_book`) suggests the wrong thing (if `dont_use_order_book=True` - we'll be using orderbook prices) - and as such, is very confusing to even look at.

freqtrade/exchange/exchange.py
```diff
@@ -1022,8 +1022,12 @@ def get_rate(self, pair: str, refresh: bool, side: str) -> float:
                 return rate
 
         conf_strategy = self._config.get(strat_name, {})
+        dont_use_order_book = (
+            ('use_order_book' in conf_strategy or side == "sell") and
+            conf_strategy.get('use_order_book', False)
+        )
```
xmatthias: I'm not sure what you're trying to do with this.
the side has nothing to do with the "use_order_book" setting - it's valid for both ask_strategy and bid_strategy - as such - i don't see why side should make this kind of a difference to this method.



If that were really the case (hint: it isn't), then please revert the whole PR that pulled these 2 methods 
apart.

You can surely simplify `if conf_strategy.get('use_order_book', False) and ('use_order_book' in conf_strategy):` to `if conf_strategy.get('use_order_book', False):` (that 2nd part is irrelevant, as `.get` will use the default if it's not in conf_strategy - but `use_order_book` is perfectly fine on both buy and sell sides.
I understand why it was added the way it was (we had this slight difference between `get_buy_rate` and `get_sell_rate` - while it doesn't hurt - it's also not problematic to not have this 2nd check.

xmatthias (changes requested):
I find basically every changed line here wrong ... 

Getting rates is fine for both sides as is (and `PricingError` was raised before from both sides - and is raised now as well).

the only thing that (slightly) concerns me is that this does not make tests fail - as it should as it's introducing wrong behaviour on the buy side.


On top of that - pricing has nothing to do with #5334.
While the "pricing" message is there - the error is something entirely different.

samgermain:
> I find basically every changed line here wrong ...
> 
> Getting rates is fine for both sides as is (and `PricingError` was raised before from both sides - and is raised now as well).
> 
> the only thing that (slightly) concerns me is that this does not make tests fail - as it should as it's introducing wrong behaviour on the buy side.
> 
> On top of that - pricing has nothing to do with #5334.
> While the "pricing" message is there - the error is something entirely different.

Sounds good, I thought this might have been causing the problems in #5334 

I went to the commit right before the first commit on #5284 ( 03a4ae46746905866f2ab7aba86534d6e0c1791e ) and took  a [screenshot](https://i.imgur.com/vr3QCMn.png) of the `get_buy_rate` and `get_sell_rate` methods and highlighted some stuff, so you can see that the pricing error was only raised in `get_sell_rate` before and the `get_sell_rate` condition didn't check if `use_order_book` was in `ask_strategy`

------------------

Feel free to close the PR. I'll leave it open in case you want it

xmatthias:
I do have been aware of the changes this introduced - and that the methods were not 100% aligned (which is now better though - as the "none" price check was missing on the Buy side (which can happen with duplicated code).

Pricing / Pricing exception was / is however never an issue in the first place with either methods and both methods could raise PricingError (if orderbook is empty / raises Keyerrors, for example) - and the calling code will handle that accordingly.

PR #5336, opened 2021-07-29
Margin exchange

samgermain:
## Summary
Adds leverage and shorting capabilities to freqtradebot and exchange classes

## What's new?

Exchange methods

- apply_leverage_to_stake_amount
- get_max_leverage
- setup_leveraged_enter
- complete_leveraged_exit
- get_isolated_liq
- get_interest_rate

lot's of checks for `side="buy"` or `side="sell"` in `freqtradebot`


samgermain:
Only the last five commits on here are unique to this PR, the other ones are in `margin-db` or `combined_exchange_buy_sell`

freqtrade/exchange/binance.py
```diff
@@ -89,3 +96,109 @@ def stoploss(self, pair: str, amount: float, stop_price: float, order_types: Dic
...
+
+    def transfer(self, asset: str, amount: float, frm: str, to: str, pair: Optional[str]):
+        res = self._api.sapi_post_margin_isolated_transfer({
+            "asset": asset,
+            "amount": amount,
+            "transFrom": frm,
+            "transTo": to,
+            "symbol": pair
+        })
+        logger.info(f"Transfer response: {res}")
```
xmatthias, line 111: As mentioned before - please focus on cross-margin - or remove support for binance for now completely.

I see isolated margin out of scope for the moment (as mentioned before already) - as it includes/introduces another level of complexity by having to transfer to the isolated margin account first.

That can be an enhancement added once stability of the whole short-logic is confirmed and merged - but not until then - and i'll reject all Pull requests that add parts of this.

It's also something that's binance-specific - and i've not seen any other exchange that uses seperate cross-margin and isolated margin accounts - so this will need a very good investigation if it's really worth the implementation and maintenance effort at all.

freqtrade/exchange/bittrex.py
```diff
@@ -23,3 +24,28 @@ class Bittrex(Exchange):
...
+
+    def setup_leveraged_enter(
+        self,
+        pair: str,
+        leverage: float,
+        amount: float,
+        quote_currency: Optional[str],
+        is_short: Optional[bool]
+    ):
+        raise OperationalException("Bittrex does not support leveraged trading")
```
xmatthias: You should implement these methods in the parent exchange class with this exception - and override it in classes that do actually implement this.

otherwise you'll have to copy/paste the same message to all potential classes (without the guarantee that you'll cover all future exchange classes (or exchanges that don't have their own subclass).

samgermain: There is one in the main exchange class that says it's not supported using freqtrade. This one is more specific, saying that the exchange doesn't support it at all

xmatthias: There's little point doing it THIS way (you should never even reach this point in the first place) - as it would only error out once you try to open a trade - this must however be flagged at startup as "invalid configuration".


We'll need to have a flag (most likely in `_ft_has`) which "enables" an exchange for "optional" features/modes - similar to `stoploss_on_exchange` (in `_ft_has`) - but most likely more complex as this will need to cover multiple cases (leverage (for longs), short, ... ).

Obviously, the same checks that we use for `stoploss_on_exchange` (abort startup if an invalid configuration for the selected exchange is detected) will also need to be added here.

freqtrade/exchange/binance.py
samgermain: If simplicity is what you're going for, I don't think cross-margin is the way to do it, I created another PR #5339 which adds skeleton functions for implementing cross margin, the most complicated part would be keeping an accurate calculation of the maintenance margin, the 1m candles wouldn't be good enough because one candle could be at  it's peak while another could be at it's low. Is there another way that freqtrade keeps track of the price? Is the price calculated in realtime for the stoploss?

I'll remove this transfer stuff from this PR. 

I think the least complicated would be isolated futures, there's no transfer involved, I can look at FTX also

xmatthias: You will need the logic for cross-margin for kraken anyway, so the complication here is that the isolated futures is a different logic - with additional calls to the exchange - as well as some math problems (see below).


An additional point (call it a binance problem - although i'm not sure it's not similar on other exchanges) is the leftover dust
This will already be a (potentially costly) problem in cross-margin trading, and will be very difficult to handle for isolated margin.

Especially when borrowing non-stake currencies (speak, when shorting) - we'll need to find a way to avoid "dust" due to fees (if you buy XRP for 100$ - you might make your order for 60XRP - but you'll only get 59.9XRP due to fees - which will result in 0.9 XRP being unsellable (assuming you can only trade XRP in 1 XRP steps).
In spot markets, this is not a huge problem, as you have the "convert dust to BNB" button maybe once a week or month.
It can also be mostly avoided by using BNB for fees - but we cannot assume that BNB will be available in a user's account.

However, if the above trade was a short, you've borrowed all the XRP - so as an endresult, you're now unable to repay 100% of what you bought, which will result in "borrow fees" even after the trade is closed.
in a cross-margin wallet, you can still "convert to BNB" all the leftover dust .... but you'll be left with the borrowed amounts / interrests - so i've already had cases (on manual trades) where i had to go back to that pair.

I'm yet to find this for Isolated margin - And if it's there, you'll need to do this for each currency seperately (as each currency is a seperate wallet for binance) - so the complication is not only with freqtrade - but with the whole handling on the exchange.

If you go to "automatically" transfer assets between wallets (which is what you need for isolated margin trading) - you'll also need to make sure to get to 0 (so the wallet shows 0.00000 for that pair after the trade is complete). Due to minimum trading requirements, this might prove difficult.

---

the whole margin handling must be "mostly" transparent to the bot anyway - so "most" of the logic (if it's 1 call to borrow, "automatic borrow" like in kraken, or 20 calls to borrow) must all be "hidden" inside the exchange class. 
Now you'll need one call to "repay" eventually (that does nothing in the regular case) - but this should be the exception, not the rule.

Otherwise you'll end up with a lot of "if exchange = kraken then this" logic - which i'd only really want in very limited sections where it can't be avoided.

samgermain:
I'll reopen when I'm reading to rebase this

PR #5339, opened 2021-07-29
Liquidation Price

samgermain:
## Summary
Implements maintenance margin liquidation, which is used for cross margin 

## Quick changelog

Adds MaintenanceMargin thread class
- thread ran in freqtradebot, started inside __init__
- sells all trades if the maintenance_margin falls below the liquidation level

Adds MaintenanceMarginFormula enum

## What's new?

This mostly just adds skeleton functions that must be written for the maintenance margin to work properly


samgermain:
I'm not sure why PR tests always fail, they always pass locally

PR #5346, opened 2021-07-30
Liquidation Price formula classes

samgermain:
## Summary
Adds classes for formulas that calculate liquidation price, but does not implement any of the formulas yet
LocalTrade, Trade and freqtradebot will get an optional liq_formula variable which will be used to determine the liquidation price

## Quick changelog

Adds enum classes LeverageMode and LiqFormula


samgermain:
I think it would be better to start creating smaller PRs like this, which aren't completely implemented, that could be merged onto either develop or a separate Leverage branch. There will be less to review, and it gives me the opportunity to let others know how they can help out if they want to.

The point of the changes in this PR was to make it easier to support any of those leverage modes in the future, without making big changes

xmatthias:
> I think it would be better to start creating smaller PRs like this, which aren't completely implemented, that could be merged onto either develop or a separate Leverage branch. There will be less to review, and it gives me the opportunity to let others know how they can help out if they want to.
> 
> The point of the changes in this PR was to make it easier to support any of those leverage modes in the future, without making big changes

Even though i only read this now ... this was my idea with feat/short (which i've pointed this branch to now ...).

I'd aprecitate if you could use "TODO short" (or something similar - which however should be aligned throughout) for todo's that MUST be completed before we can look at merging `feat/short` into develop ... that'll make them stand out / easy to grep through - as there's other TODO's that are not relevant to this topic.

samgermain:
Any chance this could get merged sometime soon onto the leverage branch? I've had some people saying that they want to help out, and I think that the TODOs in this commit would be a good isolated first PR for someone, but that it would be easier if they could just pull the leverage branch

All the test pass locally, not sure why they fail here

PR #5347, opened 2021-07-31
Custom order price

### LINKED ISSUE #848: Strategy should control bid price
# Idea
Because it's our `Strategy` instance that calculates if the bot should buy or sell, it should be the same Strategy that decides the bid price for that sell/buy.

This would:
- Move another core logic to where the needed information is (ROI tables and buy/sell signals are already in the Strategy)
- Allow for easy extending to new buy/sell logic and techniques

# Implementation

- [ ] remove the current bid/ask balance calculation and related configuration
- [ ] make Strategy return a tuple instead of boolean for buy and sell signal, where the first value is the buy/sell decision boolean, and second is the buy/sell price
- [ ] implement base class for strategy so that old strategies that return just boolean can still be used and the base class will return tuple with default price

### LINKED ISSUE #911: Add fill_holes_in_book option to config file so the bot place orders between two orders on the orderbook
Now that we have or will soon have the ability to place orders using the orderbook an not only the top order another useful feature would be the ability for the bot to fill holes on the orderbook, what it means is that if the bot bought B2G at 174.5 USD but there are only two orders close to that price, one is at 174 USD and the other one at 175 USD, if we open a new order for a few cents lower than the one at 175 USD for example at 174.9 USD we will be one position higher on the orderbook and most likely the volume of our order will be smaller than the one at 175 USD meaning that our order will have a higher chance to get filled before the one at 175 USD.

### LINKED ISSUE #5102: Allow custom entry price from strategy
**Description**
Now in freqtrade the entry price used in a new buy limit order is defined by market price or prices already existing in the order book.

This entry strategy have goods but also downsides, it can provide a quick entry, but not at the cheapest price.
It's really hard to make profits in short time frames (1m/5m) using this entry style.

The issue is that using an entry price like that have disadvantages:
-There is good chance that the order will execute immediatly, so we pay more trading fees as we take volume from the order book.
-Entry is not made at a low price
-The entry price have no special signification (not a market structure support like bollinger band, moving average, fibonacci level...)
No bounce can be expected at this price level

Advantages of using a custom entry price:
-The order will be placed in the order book and won't be trigger immediatly, we will pay lower trading fees
-Entry is made at a cheaper price when there is a correction
-If set correctly the custom entry price can have a signification in the market structure and a bounce from this price can be expected

**Proposition**
The solution is to allow the user to set a custom entry price from his strategy
It can be in the form of a custom indicator with a fixed price something like "prefered entry price" or "pep"
The the price will be sent to the dataprovider like used for the custom stoploss.

**Backtesting**
It can be challenging to backtest that kind of behaviours, I came up with a simple idea:
If the entry candle low go below the "prefered entry price" entry is considered as successful at defined price.
If the entry candle low stay over the "pep" entry failed and if the buy signal is still present a retry will occur at the same or updated pep.

pep

Axel-CH:
## Summary
This feature allow the user to use custom entry prices from his strategy

closes #5102
closes #848
closes #911

## TODO list

**PART 1**

- [x] Create custom entry price function in interface
- [x] Update strategy resolver
- [x] Add function request in execute_buy in freqtradebot
- [x] add test case for execute_buy in freqtradebot_test -- To Complete

**PART 2**
- [x] Review all PART 1 tasks and create related tasks for custom exit prices
- [x] Add doc and exemples
- [x] PEP8

## What's new?
Allow user to create market making strategies by setting custom entry and exit prices


xmatthias:
> On freqtrade plotly plot, [...]

there's a difference between the (buy, sell) signals (which are plotted on open or close, don't remember) and really executed trades, which are plotted at "open_rate" and "close_rate" respectively.

For signals - you should not use this - as it's a signal that has not been acted on (it's just the signal to buy).
Doing so would call this method a lot more often in plotting than in backtesting (many strategies have a lot more signals than actual trades) - which does not make any sense.

freqtrade/strategy/interface.py
```diff
@@ -69,6 +69,10 @@ class IStrategy(ABC, HyperStrategyMixin):
     # associated stoploss
     stoploss: float
 
+    # custom order price
+    entryprice: Optional[float] = None
+    exitprice: Optional[float] = None
```
xmatthias: This will not work.
there is one strategy object for all of freqtrade - storing a price here will have that price apply to ALL pairs that are evaluated - and as such will lead to conditional errors (only when more than one order is open at the same time).

i also really don't see why you'd need this ...

freqtrade/strategy/interface.py
```diff
@@ -280,6 +284,24 @@ def custom_stoploss(self, pair: str, trade: Trade, current_time: datetime, curre
...
+        When not implemented by a strategy, returns None, orderbook is used to set entry price
+        Only called when use_custom_entry_price is set to True.
+
+        :param pair: Pair that's currently analyzed
+        :param current_time: datetime object, containing the current datetime
+        :param current_rate: Rate, calculated based on pricing settings in ask_strategy.
+        :param **kwargs: Ensure to keep this here so updates to this won't break your strategy.
+        :return float: New entry price value if provided
+        """
+        return self.entryprice
```
xmatthias: The default should return `current_rate`, not an arbitrarily stored value (which will be wrong if 2 pairs have open orders at the same time).
i'd also probably rename the `current_rate` parameter to to `proposed_rate` - which would align it somewhat to `custom_stake_amount()` - which uses `proposed_stake` as the "pre-calculated stake".

docs/strategy-advanced.md
```diff
@@ -366,7 +393,7 @@ However, freqtrade also offers a custom callback for both order types, which all
 !!! Note
     Unfilled order timeouts are not relevant during backtesting or hyperopt, and are only relevant during real (live) trading. Therefore these methods are only called in these circumstances.
 
-### Custom order timeout example
```
xmatthias: I'd say this was pretty much intended - as it indents this in the visual (and outline) hierarchy below `#custom order timeout rules`.

freqtrade/freqtradebot.py
```diff
@@ -169,6 +169,7 @@ def process(self) -> None:
         with self._sell_lock:
             # Check and handle any timed out open orders
             self.check_handle_timedout()
+            self.check_handle_custom_entryprice_outdated()
```
xmatthias: Please treat this as a seperate feature (which means a seperate PR).
Adding timeouts based on price should not be part of this in my opinion (as i mentioned to you on discord already some time ago).

i'm not totally against this feature per se - but i'll want to have this as a seperate, isolated PR - which makes it clear which parts are changed because of the custom entryprice - and which because of the "trailing price".

The 2 are not connected, really - and we can have either of the two without the other.
we can also timeout orders based on "pre-calculated" pricing (not custom price) without ever returning a custom price - but based on a % offset that price moved in the wrong direction.

As such - please revert the changes related to this handling.

xmatthias (changes requested):

freqtrade/freqtradebot.py
Axel-CH: I see, but how we treat unfulfilled custom entry orders? Just with the timeout for the moment?

xmatthias: Just the same as we treat regular limit orders now?

Currently - we use `get_rate()` to define a price (maybe 1st or 4rd spot below current price in the orderbook - you can configure that).

with the outcome of this PR - the result of `get_rate()` is passed to the strategy for "further modification" - which has the ability to modify / completely change this result.

If the strategy does that - we'll simply have a new price - maybe between 3rd and 4th spot in the orderbook - maybe 0.5% below current price - maybe a fixed value of 200 (just for the fun of it).

This does not impact timout handling / order cancelling at all - and regular timeout rules will apply just as usual.

---

What you're trying to add here is trailing entry (where you can potentially follow the price if it moves away from your order).
This can be useful with custom pricing - but can at the same time be used with the regular `get_rate()` method (without ever defining a custom rate). 
As such - it's completely unrelated to specifying your own price. 
Having the 2 seperate will clearly show which parts of the code are changed because of which.

Before investigating / investing too much time into this - please also check / evaluate if (and why) this can't be handled by `check_buy_timeout` and `check_sell_timeout` respectively.
If you look at [the samples](https://www.freqtrade.io/en/latest/strategy-advanced/#trade-entry-buy-order-confirmation) - it's basically doing exactly this - cancelling the order if price moved away from the current order.

That will give the strategy full control over it - while providing you basically the same functionality (you can use `trade.open_rate_requested` to get the original price of entry).

freqtrade/strategy/interface.py
```diff
@@ -633,6 +651,42 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
...
+        :param current_rate: current rate
+        :param low: Low value of this candle, only set in backtesting
+        :param high: High value of this candle, only set in backtesting
+        """
+
+        if self.use_custom_entry_price:
+            entry_price_value = strategy_safe_wrapper(self.custom_entry_price, default_retval=None
+                                                    )(pair=pair,
+                                                      current_time=current_time,
+                                                      current_rate=current_rate)
```
xmatthias: I'm not sure we need this whole function.

The whole logic logic is only relevant for backtesting (and i'd not add short logic at this point in time - thats something we'll add at a later point only) - and as such - should be in the backtesting class.

However, i'm not quite sure simulating failed entries will make sense to backtesting - as usually, your offsets should be rather tight - so the only case when entries fail is if you're "outside" / below the candle.
That won't however properly simulate "real" trades - as even prices within the candle may never be reached again. 

In backtesting, you'll also need to be VERY careful with logging - as it can spam you pretty good with one odd line (even worse in hyperopt).

freqtrade/constants.py
```diff
@@ -432,6 +432,8 @@
 
 CANCEL_REASON = {
     "TIMEOUT": "cancelled due to timeout",
+    "ENTRYPRICECHANGED": "Custom entry price changed",
+    "EXITPRICECHANGED": "Custom exit price changed",
```
xmatthias: I'd think this is also part of the "timeout logic" ... ;)

freqtrade/freqtradebot.py
```diff
@@ -480,6 +480,14 @@ def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = N
         else:
             # Calculate price
             buy_limit_requested = self.exchange.get_rate(pair, refresh=True, side="buy")
+            if self.config.get('use_custom_entry_price', False):
+                buy_rate = self.exchange.get_rate(pair, refresh=True, side="buy")
```
xmatthias: Please reuse `buy_limit_requested` to avoid making another (unnecessary) call to the exchange.

freqtrade/freqtradebot.py
```diff
@@ -480,6 +480,14 @@ def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = N
         else:
             # Calculate price
             buy_limit_requested = self.exchange.get_rate(pair, refresh=True, side="buy")
+            if self.config.get('use_custom_entry_price', False):
```
xmatthias: I'm quite hesitant about `use_custom_entry_price`.

The default will not change behaviour (so if a strategy doesn't implement this - it'll simply return poposed_rate) - so it might be "always on" - and the opt-in is by overwriting the method in your strategy.

Having another setting is not bad per se - but i feel it's unnecessary in this instance.

freqtrade/freqtradebot.py
```diff
@@ -480,6 +480,14 @@ def execute_buy(self, pair: str, stake_amount: float, price: Optional[float] = N
...
+                buy_rate = self.exchange.get_rate(pair, refresh=True, side="buy")
+                custom_entry_price = strategy_safe_wrapper(self.strategy.custom_entry_price,
+                                                    default_retval=stake_amount)(
+                    pair=pair, current_time=datetime.now(timezone.utc),
+                    current_rate=buy_rate)
```
xmatthias: You'll want to rename the kwarg to `proposed_rate` - to align it to the interface ;)

tests/strategy/test_interface.py
```diff
@@ -118,11 +107,7 @@ def test_get_signal_old_dataframe(default_conf, mocker, caplog, ohlcv_history):
     caplog.set_level(logging.INFO)
     mocker.patch.object(_STRATEGY, 'assert_df')
 
-    assert (False, False, None) == _STRATEGY.get_signal(
-        'xyz',
-        default_conf['timeframe'],
-        mocked_history
-    )
+    assert (False, False) == _STRATEGY.get_signal('xyz', default_conf['timeframe'], mocked_history)
```
xmatthias: I think you'll need to rebase your branch on top of develop (or merge develop to it) to fix tests.

These changes don't seem related to your modifications - but break tests as get_signal returns a 3-element tuple now.

freqtrade/strategy/interface.py
Axel-CH: So I remove logs and short logics,  However I think It's useful to simulate failed entries. The distance between current price and your proposed entry price can be an hyperparameter for hyperopt

freqtrade/constants.py
Axel-CH: Yes I will remove that

freqtrade/freqtradebot.py
Axel-CH: Good suggestion

freqtrade/strategy/interface.py
xmatthias: At the moment, neither of the 2 added methods is called from backtesting - so it's not simulating anything.

In general, i'd agree - assuming you find a way to reliably and realistically simulate failed entries.
simply "below low" will however not do it in my opinion (even though it might be an acceptable start).


In live, you have to assume that an entry happens 2-5s after open (depending on the length of the pairlist and complexity of a strategy, it can also be more). You'll not know if (by that time) low already happened or not.
Sure, on longer timeframes (1h and up) - chances are pretty high that low is still going to happen.

Especially on smaller timesframes (1m, 5m) - you can however encounter cases where low is hit "immediately" before taking off - so even though your order is within the candle, it'll never be reached.

I'm fine with it as long as we clearly document the shortcommings of this function - otherwise it'll boost assumptions on the reliability of backtesting which are simply not there.

---

As mentioned above - as this is a function only relevant to backtesting, please move it to the backtesting file, having it in `interface.py` will be missleading (and for dry-run, we have a better way to simulate failed entries now).

Axel-CH:
@xmatthias if the user return None or a non float value, what is the prefered behaviour? return default limit value or raise an error?

xmatthias:
> @xmatthias if the user return None or a non float value, what is the prefered behaviour? return default limit value or raise an error?

i guess the best behaviour is to fall back to `proposed_rate` in that case.
If it's clearly documented - this will simply fall back to default behaviour (as if this method isn't there / would not be called).

maybe an odd question - is it intentional that this isn't part of backtesting (it's not called at all)? 
I don't necessarily mind that (it'd be a guess in backtesting anyway) - but it should be intentional whichever route we go.

Axel-CH:
Ok will check and test the default behaviour. Do you think we need to add unit test for this specific case of a custom entry or exit price set at None?

I'm planning to include the backtesting part of that in a seperate PR,
Regards,
Axel

xmatthias:
> Ok will check and test the default behaviour. Do you think we need to add unit test for this specific case of a custom entry or exit price set at None?

Yeah, i'd think so (it'll most likely be a very quick and easy test case anyway).
 
> I'm planning to include the backtesting part of that in a seperate PR,

i'm "kindof" ok with that ... although from a timing perspective, I'd want to have both as part of the same release (speak, merged within the same month) ... otherwise it'll be a strange "middle-state" - where custom-rates will apply to  live/dry - but not to backtesting. The subsequent change (adding support for this to backtesting) will then break / change people's backtests ... which is ok for develop, but should be avoided for stable releases.

freqtrade/freqtradebot.py
```diff
@@ -1375,3 +1392,21 @@ def fee_detection_from_trades(self, trade: Trade, order: Dict, order_amount: flo
...
+        Check if the custom price is of the good type if not return proposed_price
+        :return: valid price for the order
+        """
+        if custom_price:
+            if isinstance(custom_price, int):
+                valid_price = float(custom_price)
+            elif isinstance(custom_price, float):
+                valid_price = custom_price
+            else:
+                valid_price = proposed_price
```
xmatthias: i think a try/except block trying to parse to float directly will be more reliable.

assuming someone (by accident) returns a price as string instead of float - that would handle this case too.

I'm asking myself now - should we top/bottom cap this price to a few% off proposed price?
there's little point in having a price 20% away from current price ...

Axel-CH: Yes I will change that to a try catch method.
Concerning the fact that the user can define a price really far away from the current price, I don't think it's an obligation cap the price.

In fact I plan to set a price really lower If for example if the current price is not where I wanted it to be when i decided to buy.
That way I can perform a sort of "soft abort" of the entry.

Axel-CH: Maybe we should log a warning like "Your entry/exit price is XX% lower of the current price, your order have a low probability of getting filled"

xmatthias: it's honestly just a thought (basically asking for your opinion) - not a requirement.

i'd not use that as "soft abort" - but use the proper method for that ...

Axel-CH: Yes I know it's better to have a function for that like "last_moment_abort". But for the moment setting an entry price far away is a temporary solution ;-)

xmatthias: You really should read the advanced strategy documentation :laughing: 

We do have `confirm_trade_entry()` - which is made specifically for this.

Axel-CH: Hahaha my bad, I always thought this method was about a user confirmation with telegram

xmatthias: i've just played with this - and "intentionally" placed the orders 10% below current price.

This makes it seem like I've a trade open with immediate 10% profits (it'll report that way until the order is filled) - but it'll then cancel the order as it's really been a bullshit order.

As security measure, I'd propose to add a "allowed range" that defaults to 1-2% (we can make it a parameter so if you explicitly want this, you can go beyond that value). 
Especially your example (using an indicator for pricing) can otherwise result in prices that are far off spot and resut in almost all orders timing out.

Axel-CH: I'm agree with the security 👍🏽

Axel-CH: Maybe we should display profit only if order is filled to limit confusion

xmatthias: Yeah, but thats a different topic alltogether, as you can also configure orderbook to almost never fill ...
also, the display DOES include the info that the order is still open (* or ** next to the pairname in both UI and telegram).

Axel-CH: > As security measure, I'd propose to add a "allowed range" that defaults to 1-2% (we can make it a parameter so if you explicitly want this, you can go beyond that value).
> Especially your example (using an indicator for pricing) can otherwise result in prices that are far off spot and resut in almost all orders timing out.

If the user set a price outside the allowed range I think we should cancel the order. It seem dangerous to change the price for him.

xmatthias: if you work on that premise, then you'll have to also cancel it in case of None or non-float values.

To me, a min/max limit around propose_price (which can be changed via config setting) will be fine (it'll also be in the documentation, so i don't see a problem with that).

Axel-CH: > if you work on that premise, then you'll have to also cancel it in case of None or non-float values. You are right

I understand your point but it seem weird to me.
It's like if I go buy a car priced at 10 000$ I make an offer at 9000$, the bailiff tell me that the minimum bid is 9500$ and change my offer to 9500. 
I my point of view, maybe that car is a good deal at 9000$ but not at 9500$

xmatthias: custom-order-pricing logic will not cancel order-placement.

We have `confirm_trade_entry` for that (which will get the final price as well - so the user can decide there to not buy there).

If you think "any" price is fine (also for non-experienced users) - then we can leave this "bracket" safety net out and hope there's not too many people who fail to read the documentation.

Think about it this way:
currently, we force you to place the order within ~1% (assuming you use 10 orderbook slots).

we're now opening this up to allow more - but keep default behaviour to not place bets more than 2% away from current price - to not overwelm inexperienced users with 100ds of unfilled orders due to selecting a very low price.
As this limit can be modified - i don't see a problem with that (again, it comes down to "is documented").

My base assumption here is that most people will be using either ATR, BB (or something similar) - or a % multiplier to move price up/down - so in edge-cases - you can get a price that's lower than what you expect.

---

To take your example - you go to buy a car - and tell me the price you wanna buy it at (9000$).
Instead of telling you "no, it'll have to be 9500$ - i take out a gun and shoot the buyer (cancel the order).

Axel-CH: > We have `confirm_trade_entry` for that (which will get the final price as well - so the user can decide there to not buy there).

Yes that's right,

>To me, a min/max limit around propose_price (which can be changed via config setting) will be fine (it'll also be in the documentation, so i don't see a problem with that).

let's do this

Axel-CH: @xmatthias  do you know why there is failing tests on mac os? They does not seem related my modifications

Axel-CH: After a pull from develop the issue with rpc_telegram tests of mac os has disappeared

Anytka808 (approved):

freqtrade/freqtradebot.py
```diff
@@ -1401,10 +1401,21 @@ def get_valid_price(self, custom_price: float, proposed_price: float) -> float:
...
-                valid_price = float(custom_price)
+                valid_custom_price = float(custom_price)
             except ValueError:
-                valid_price = proposed_price
+                valid_custom_price = proposed_price
         else:
-            valid_price = proposed_price
+            valid_custom_price = proposed_price
+
+        cust_p_max_dist_pct = self.config.get('custom_price_max_distance_percent', 2.0)
```
xmatthias: At first, i was curious to see the default set to 2.0 - assuming "why is it set to 200%" ... 

I think the way this is done (using percent instead of a ratio) will be pretty confusing and missaligned with the rest of the bot / bot settings.

All freqtrade settings (at least the ones i'm aware of) that need a % must be within a range of 0-1 (1 = 100% - which means they're a ratio - you get 100% by multiplying with 100).

This new setting is now in a range between 0 and 100 - which means 0.2 is 0.2% - not 20% (as i'd expect from having experience with all other settings.

docs/strategy-advanced.md
```diff
@@ -398,7 +398,10 @@ class AwesomeStrategy(IStrategy):
 ```
 
 !!! Warning
-    Modifying entry and exit prices will only work for limit orders. Depending on the price chosen, this can result in a lot of unfilled orders.
+    Modifying entry and exit prices will only work for limit orders. Depending on the price chosen, this can result in a lot of unfilled orders. By default the maximum allowed distance between the current price and the custom price is 2%, this value can be changed in config with the `custom_price_max_distance_percent` parameter.
+
+_Exemple_
+If the new_entryprice is 97, the proposed_rate is 100 and the `custom_price_max_distance_percent` is set to 2%, The retained valid custom entry price will be 98.
```
xmatthias: I don't think this will render well, and i think we should prefer an "example box" instead, which will render better in the documentation.

```suggestion
!!! Exemple 
    If the new_entryprice is 97, the proposed_rate is 100 and the `custom_price_max_distance_percent` is set to 2%, The retained valid custom entry price will be 98.
```

xmatthias (changes requested):
I think the new setting should be slightly renamed (`custom_price_max_distance_ratio`) - and use a range of 0-1 (otherwise this will be the only setting where 0.5 means 0.5%, not 50%).

Also - we'll have to add this to `configuration.md` which is intended to serve as a list of all possible settings - not having it there will be confusing if you need to look for it's meaning.

docs/strategy-advanced.md
Axel-CH: Can you give me more details or an exemple of what you mean by an "example box"?

freqtrade/freqtradebot.py
Axel-CH: I will change that to a 0-1 range

docs/strategy-advanced.md
Axel-CH: Ok I understood, i watched on my phone first (without the suggestion)

xmatthias: phone versions can be ... dangeorus sometimes :laughing: 

be careful with pressing the "commit" button in the comment though - as the parameter name might change (see other commit)...

docs/strategy-advanced.md
```diff
@@ -398,10 +398,10 @@ class AwesomeStrategy(IStrategy):
 ```
 
 !!! Warning
-    Modifying entry and exit prices will only work for limit orders. Depending on the price chosen, this can result in a lot of unfilled orders. By default the maximum allowed distance between the current price and the custom price is 2%, this value can be changed in config with the `custom_price_max_distance_percent` parameter.
+    Modifying entry and exit prices will only work for limit orders. Depending on the price chosen, this can result in a lot of unfilled orders. By default the maximum allowed distance between the current price and the custom price is 2%, this value can be changed in config with the `custom_price_max_distance_ratio` parameter.
 
-_Exemple_
-If the new_entryprice is 97, the proposed_rate is 100 and the `custom_price_max_distance_percent` is set to 2%, The retained valid custom entry price will be 98.
+!!! Exemple
```
```suggestion
!!! Example
```

freqtrade/freqtradebot.py
```diff
@@ -1375,3 +1392,26 @@ def fee_detection_from_trades(self, trade: Trade, order: Dict, order_amount: flo
...
+            valid_custom_price = proposed_price
+
+        cust_p_max_dist_r = self.config.get('custom_price_max_distance_ratio', 0.02)
+        min_custom_price_allowed = proposed_price - (proposed_price * cust_p_max_dist_r)
+        max_custom_price_allowed = proposed_price + (proposed_price * cust_p_max_dist_r)
+
+        # Bracket between min_custom_price_allowed and max_custom_price_allowed
+        return max(
+            min(valid_custom_price, max_custom_price_allowed),
+            min_custom_price_allowed)
```
xmatthias, line 1417: Did a small adjustment here to simplify the code a bit (it's a simple bracket logic - min/max does this in simpler terms).

xmatthias (approved):
LGTM :+1: 

Great job! :100: i hope this will benefit people as much as they think it will.

Fimap9to1:
Hi, any change this function will be backtestable?

xmatthias:
Not at the moment, and that will require good consideration to avoid unrealistic backtests due to favorable pricing strategies.

Fimap9to1:
Is this the place to talk about it or should I open a question issue?

xmatthias:
For the moment, i don't see the need to talk about it. Most parts are layed out in the comments / review comments above already. It's not implemented currently (intentionally) - and i fear that having this "backtestable" will not give you much and cause more confusion than it'll help - as you can't reliably simulate unfilled orders in backtesting - which is what this will most likely result in.

Fimap9to1:
Good thanks!

PR #5348, opened 2021-07-31
Liquidation price

samgermain:
Adds maintenance margin and liquidation formula classes


xmatthias:
What's the difference between this and #5339?

freqtrade/maintenance_margin.py
```diff
@@ -0,0 +1,38 @@
+from freqtrade.enums import LiqFormula
+from freqtrade.persistence import Trade
+
+
+class MaintenanceMargin:
+
+    trades: list[Trade]
```
xmatthias: You need to uppercase List if you wanna use it this way for a type.

freqtrade/maintenance_margin.py
```diff
@@ -0,0 +1,38 @@
...
+        return
+
+    # ? def update_trade_pric(self):
+
+    def sell_all(self):
+        return
+
+    def run(self):
+        # TODO-mg: implement a thread that constantly updates with every price change,
+        # TODO-mg: must update at least every second
```
xmatthias, line 37: If this is a requirement - assume this as impossible.
Maybe add a safety margin next to liquidation to manually liquidate - but i'd handle it in a similar way than stoploss-on-exchange - which is checked once every iteration (no matter how long that takes).

with exchange-ratelimiting and strategy processing, you'll not get this done.
Also - we'll definitely not spin out into different threads for this - so it'll be "in-line" with the regular bot processing - which can't guarantee any runtime, really (as it depends on the strategy) - and everything < 5s will get you blocked in the long run (especially on kraken).

samgermain:
> What's the difference between this and #5339?

I changed the name of the branch, and then when I came back to the PRs I couldn't see it so I thought it was deleted. Then I created this one.

freqtrade/maintenance_margin.py
samgermain: It's not a requirement, every 5s might be enough. The problem is that with cross margin, there is one liquidation price, which depends on the current price of every trade your in. If you can see that in the last minute that a few coins went down 5% or something, and then came back up, that doesn't mean that they all did it at the same time, so maybe your maintenance margin dropped below the liquidation price, but maybe one of your coins went up at the right time within that minute to prevent the maintenance margin from dropping low enough.

With isolated margin, there's only one price to look at, so its obvious if you drop below the liquidation price. 

This is why I think isolated futures would be the easiest, it wouldn't require any transferring either (like isolated margin would), I don't believe that there is any dust with futures either

It looks like on binance, an isolated futures order would be something like

```
binance_usdm.set_leverage(symbol='ADA/USDT', leverage=5)
    order = binance_usdm.create_order(
        symbol='ADA/USDT',
        type='market',
        side='buy',
        amount=amount
    )
```

and I'm not sure about FTX, and this is why I don't want to add much more support for Kraken. I see no reason to remove the tests and formulas which have been written for Kraken already, but we could just block leverage support for it until someone is willing to get a maintenance margin class that works well

xmatthias:
> I changed the name of the branch, and then when I came back to the PRs I couldn't see it so I thought it was deleted. Then I created this one.

ah, yeah, github will do that if you rename branches on open PR's - as for github, it looks like it's gone / deleted (i guess).

freqtrade/maintenance_margin.py
xmatthias: see - my main problem with looking at binance is that they have like 4 or 5 methods to do leveraged trades (margin, isolated margin, futures, apparently isolated futures, and futures_coin).
Each of these works slightly differently - so only adding support for binance is risky, as it might lead us down a wrong path.


If you want to support only one exchange, then pick another one (not binance) - as binance probably the most complicated / complex ones. I don't really care if it's kraken or FTX - but binance will definitely not be the only one (or if we go with one only, it will not be part of the first row).

samgermain: That just means they have more options, they don't all need to be supported. The only one that's really that complicated is isolated margin, so I'll avoid that one for sure. 

FTX supports cross margin and perpetual cross futures, and kraken supports cross margin.

Isolated futures is really the least complicated of all of them, it only involves that extra `set_leverage` command, which I think will also work for FTX, with cross margin and cross futures though there is the extra part of keeping track of the maintenance margin, which i'm worried we won't be able to do very accurately if we're not updating it that often,

xmatthias: for the moment - i'm not decided on which exchange / market kind (margin / futures / ...) will be added initially (neither are you, i think) - but i also don't think it's too important at this point.

Looking at a file like `liqformula.py` however i do have some strong concerns.
Having all these (either exception or TODO) if/else statements in there is basically a technical depth (before it's even merged).
There's no point in having `TradingMode.ISOLATED_MARGIN` in the enum if we don't plan on implementing it. 
Should someone see this and implement just the calculation, it'd potentially be wasted time/effort both on the implementation/research, as well as on the review side.

Also - it's a burden for every refactoring that might potentially happen - so i see no point in doing it this way.
If the calculation is added (with the intend to support this mode by the author) - then sure - add it as if/elif ... otherwise, don't even include it in the enum.

freqtrade/enums/liqformula.py
```diff
@@ -0,0 +1,86 @@
...
+                trading_mode == TradingMode.ISOLATED_FUTURES:
+            self._exception(trading_mode, True)
+        else:
+            self._exception(trading_mode)
+
+    def _ftx(self, trading_mode: TradingMode):
+        # TODO-lev: Additional arguments, fill in formulas
+        self._exception(trading_mode)
+
+    def __call__(self, **k):
```
xmatthias: While this technically will for sure work - please keep enums as what they are - Lists of possibilities.

Implementation of the calculation should not go there (i've never evne seen it done this way in any other bigger project) - and the enum is for sure the last place i'd go look for a calculation.

samgermain: It's done this way so that sqlalchemy can store the value, like how `InterestMode` is done, because the LiquidationPrice method is called in the `set_liquidation_price` method of `LocalTrade`

freqtrade/maintenance_margin.py
samgermain: I separated TradingMode into `TradingMode` and `Collateral` with `Collateral` being `Isolated` and `Cross`. The logic of when both were used didn't have much overlap.

`LiqFormula` is an enum because it's used similarly to `InterestMode`. It's called in LocalTrade, but differs based on exchange. I took most of the functions out of the LiqFormula class and have them as independent functions though if you like that better

samgermain: Basically any isolated liquidation price would need to be stored on the Trade/LocalTrade object, and I think that object should be able to handle the logic also

freqtrade/enums/liqformula.py
xmatthias: Please don't mix 2 things here - SqlAlchemy is not related to what i mean above.
You can use an Enum just fine (we use that in other places too) - and use sqlalchemy will store the enum.


What i don't like is having Logic in the enum.
Sqlalchemy does not require you to convert the enum to a callable (implementing `__call__` converts it technically to a callable/function).
My point is to extract the LOGIC from the enum - so you don't call `InterestMode.Binance()` - but you pass it to a function as parameter(like `calc_interestmode(InterestMode.Binance)`.

The logic should go someplace else (not the `freqtrade.enums` package for sure) - maybe add a new Package for liquidation logic.
The enum (list of possible values) should stay in `freqtrade.enums` - while the logic should move.

This is also true for `InterestMode` (i allowed it temporarily there - but as i see now you're trying to use this too much, please do NOT store any logic in the enum). 

Also for `InterestMode` - we'll refactor this logic out fo the Enum before merging that to Develop.

samgermain: Where would `calc_interestmode` go?

xmatthias: we'll need to be a bit careful with cyclic imports here - which is why i don't think we can (or should?) move it to a new file in `freqtrade.exchange` - as i've so far tried to avoid importing from this module to persistence - and it's another reason why i don't like it in enums - as importing things from "other" modules in `freqtrade.enums` can result in cyclic imports, too - depending where you need import from.

i'm not clear yet how many such calculations we're going to have ... if it's just 2 or 3 - we can have a file "somewhere" with it.
If there's "too many" - not sure what the limit should be -  maybe it makes sense to have them as a seperate module (`freqtrade.margin_calculations` or so) - which we can import where necessary.

Maybe start with having it as seperate file in `freqtrade.persistence`?
Moving a whole file from one module to another is usually a quick task (all we need to change is imports - as long as it's not something that's exposed to the user/strategy) - so i'm not too concerned about the precise location for now.

xmatthias (changes requested):
As you've included this class the same way in #5364 - maybe i wasn't clear before.

Please remove all logic from the `freqtrade.enums` and move it to seperate files in a different module.

This also applies to the merged interestMode - however this can (and probably should) be done in a seperate "cleanup" PR, refactoring the logic out of the enums class.

samgermain:
I'm going to close this pull request and reopen it when it's ready

PR #5350, opened 2021-07-31
Add environment variable support

### LINKED ISSUE #4385: Discover configuration/secrets via environment variables

xmatthias:
## Summary
Add support for environment variables, following the format `FREQTRADE__{sections}__{key}`
a common example might be `export FREQTRADE__EXCHANGE__KEY=<yourExchangeKey>`

Closes #4385

## Quick changelog

- add environment variable support for all config options in a standardized way
- Document
- This takes precedence over all other settings except command line arguments - which are considered the "current user wish" - and as such will have the highest prevalence.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -11,6 +11,37 @@ Per default, the bot loads the configuration from the `config.json` file, locate
 
 You can specify a different configuration file used by the bot with the `-c/--config` command-line option.
 
+If you used the [Quick start](installation.md/#quick-start) method for installing 
+the bot, the installation script should have already created the default configuration file (`config.json`) for you.
+
+If the default configuration file is not created we recommend to use `freqtrade new-config --config config.json` to generate a basic configuration file.
+
+The Freqtrade configuration file is to be written in JSON format.
+
+Additionally to the standard JSON syntax, you may use one-line `// ...` and multi-line `/* ... */` comments in your configuration files and trailing commas in the lists of parameters.
+
+Do not worry if you are not familiar with JSON format -- simply open the configuration file with an editor of your choice, make some changes to the parameters you need, save your changes and, finally, restart the bot or, if it was previously stopped, run it again with the changes you made to the configuration. The bot validates the syntax of the configuration file at startup and will warn you if you made any errors editing it, pointing out problematic lines.
+
+### Environment variables
+
+Set options in the Freqtrade configuration via environment variables.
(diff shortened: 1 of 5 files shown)


kamontat:
You could add .env loader to automatically load environment value from `.env` file

xmatthias:
> You could add .env loader to automatically load environment value from `.env` file

I don't see the point - if you store it in a file anyway - why not use multiple configuration files.

also - `.env` name-collides with the virtual environment directory for users using the setup.sh script - and support for it is fluky (if you use docker, it'll work totally different than if you use it without docker).

that said - it's always a possibility for future improvements ... i do however see .env files quite sceptical - as docker-compose "highjacks" them as well (and they'll not automatically apply to the bot running inside the container).

TheJoeSchr:
Awesome!

PR #5351, opened 2021-08-01
Changed tests in tests/test_persistence.py to use usdt prices

samgermain:
Most of the tests in `tests/test_persistence.py` were update to use usdt prices. I created these functions in `tests/conftest.py`, which will be renamed to the equivalent name that excludes usdt, once all the tests across all test files are switched to use usdt

- limit_buy_order_usdt_open
- limit_buy_order_usdt
- limit_sell_order_usdt_open
- limit_sell_order_usdt
- market_buy_order_usdt
- market_sell_order_usdt

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -812,7 +812,7 @@ def shitcoinmarkets(markets):
             "future": False,
             "active": True
         },
-        })
+    })
     return shitmarkets
 
 
@@ -1115,7 +1115,7 @@ def order_book_l2_usd():
             [25.576, 262.016],
             [25.577, 178.557],
             [25.578, 78.614]
-            ],
+        ],
         'timestamp': None,
         'datetime': None,
         'nonce': 2372149736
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -2084,3 +2084,88 @@ def saved_hyperopt_results():
...
+        'price': 2.20,
+        'amount': 30.0,
+        'filled': 0.0,
+        'remaining': 30.0,
+        'status': 'open'
+    }
+
+
+@pytest.fixture
+def limit_sell_order_usdt(limit_sell_order_usdt_open):
```
xmatthias, line 2134: Just a thought while looking at this - (probably for the future / other PR's - as i don't see it too relevant for test_persistence) - would it make sense to have a fixture for "winning sell" and "losing sell" - so we have "standardized" orders without having to mess with the values in the tests to make it winning or losing depending on the need?

tests/test_persistence.py
```diff
@@ -257,136 +257,279 @@ def test_update_open_order(limit_buy_order):
...
-    assert trade.calc_close_trade_value(fee=0.005) == 0.0010619972701635854
+    trade.update(limit_sell_order_usdt)
+    assert trade.calc_close_trade_value(fee=0.005) == 65.67
 
 
 @pytest.mark.usefixtures("init_persistence")
-def test_calc_profit(limit_buy_order, limit_sell_order, fee):
+def test_calc_profit(limit_buy_order_usdt, limit_sell_order_usdt, fee):
+    """
+        10 minute limit trade on Binance/Kraken at 1x, 3x leverage
```
xmatthias: I think we'll need to be careful with comments / docstrings in this PR.

With our current approach - this will be merged to develop (so having leverage comments here might be confusing to others as it doesn't really correspond to the content of the function).

Leverage however will be merged to `feat/short` - which will (once complete) be merged to develop.
I expect this process to take several months - so having "leverage" commnets in here now will create a disconnect of several months (maybe more - if this particular test isn't adjusted in `feat/short` to fit the docstring after all.

As such - i'd (for now) delete this Docstring (there wasn't one before ...) - and only add it in a PR that's pointed towards feat/short.

tests/test_persistence.py
```diff
@@ -64,40 +65,39 @@ def test_init_dryrun_db(default_conf, tmpdir):
...
+        fee: 0.25% quote
+        open_rate: 2.00 quote
+        close_rate: 2.20 quote
+        amount: = 30.0 crypto
+        stake_amount
+            1x,-1x: 60.0  quote
+        borrowed
+             1x:  0 quote
+        open_value: (amount * open_rate) ± (amount * open_rate * fee)
+             1x, 3x: 30 * 2 + 30 * 2 * 0.0025 = 60.15 quote
```
xmatthias: Same than in the other comment - be careful with mentioning "leverage" in PR's that are intended to be merged to develop.
While the comment may be correct in the future - it's incorrect if we look at this PR alone - and incorrect in the state it's going to result in develop.

xmatthias (changes requested):
Great work! :+1: 

2 minor things about comments - and one suggestion maybe for future improvements (mostly for test_freqtradebot - i assume).

tests/conftest.py
samgermain: Not really, because it's just the 1 or 2 values that need to be changed, and then everything else stays the same. The only tests where it really matters is `test_profit` and `test_profit_ratio`, and those are changed by passing values to the functions

tests/test_persistence.py
samgermain: There not being one before though meant I had to figure their math out. I took the leverage references out

tests/conftest.py
samgermain: If it ends up coming up more I'll do that

xmatthias: > If it ends up coming up more I'll do that

it's been a thought / idea i had while looking at this ... not sure myself if it's a neccessity ...

xmatthias (approved):
Great! :+1:

PR #5364, opened 2021-08-05
Funding Fee

samgermain:
Needed for futures. Funding fees are charged, or rewarded to traders based on the price of the asset, and keep the futures contracts in line with the actual asset. Binance charges them at 00:00, 08:00, and 16:00 UTC every day.

They're explained here
https://www.binance.com/en/blog/421499824684900382/A-Beginners-Guide-To-Funding-Rates

## Summary
Creates a funding fee object which periodically increases the funding fee on all trades by looping through all trades at 00:00, 08:00, and 16:00 UTC and adding a certain amount to each trade. When the bot crashes, FundingFee has a reboot function which checks all trades for when the last funding fee was applied, and applies missed funding fees. When a new trade is created, if the time is within 15s of 00:00, 08:00, or 16:00 UTC, then a funding fee amount is applied immediately

### DIFF
--- a/freqtrade/enums/liqformula.py
+++ b/freqtrade/enums/liqformula.py
@@ -1,0 +1,108 @@
+# from decimal import Decimal
+from enum import Enum
+
+from freqtrade.enums.collateral import Collateral
+from freqtrade.enums.tradingmode import TradingMode
+from freqtrade.exceptions import OperationalException
+
+
+# from math import ceil
+
+
+class LiqFormula(Enum):
+    """Equations to calculate liquidation price"""
+
+    BINANCE = "Binance"
+    KRAKEN = "Kraken"
+    FTX = "FTX"
+    NONE = None
+
+    def __call__(self, **k):
+
+        trading_mode: TradingMode = k['trading_mode']
+        if trading_mode == TradingMode.SPOT or self.name == "NONE":
+            return None
+
+        collateral: Collateral = k['collateral']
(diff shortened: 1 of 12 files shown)


freqtrade/enums/liqformula.py
```diff
@@ -0,0 +1,108 @@
...
+        collateral: Collateral = k['collateral']
+
+        if self.name == "BINANCE":
+            return binance(trading_mode, collateral)
+        elif self.name == "KRAKEN":
+            return kraken(trading_mode, collateral)
+        elif self.name == "FTX":
+            return ftx(trading_mode, collateral)
+        else:
+            exception(self.name, trading_mode, collateral)
```
xmatthias, line 35: As mentioned in some other PR - please remove all logic from the `freqtrade.enum` package.

Again - i don't fully understand the difference between this and #5348.

xmatthias (changes requested):
I'd propose to focus on finishing one PR after the other, instead of opening one after the other without finishing each.
In general, i don't mind having parallel PR's if they don't include the same code (and are all based on the same root).

currently however, i feel like many PR's you make include the same code (like the liqformula file for example - which i also looked at in #5348, and requested the same change to be made).

This which wastes a lot of time as i'm reading through the same thing again until i realize that i did see this code already.

samgermain:
I'll close this until I can rebase

PR #5365, opened 2021-08-06
Added leverage enums

samgermain:
All the new enums that will be used for leverage. Most of the other PRs use these, so having this on the short branch will reduce the similarity between PRs. I'll rebase the other PRs onto the feat/short branch after merging


freqtrade/enums/exchangename.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+
+class ExchangeName(Enum):
+    """All the exchanges supported by freqtrade that support leverage"""
+
+    BINANCE = "Binance"
+    KRAKEN = "Kraken"
+    FTX = "FTX"
+    OTHER = None
```
samgermain: I'm not so sure about the name of this enum, but I couldn't think of a better name. Feel free to change it, or it could be a temporary name

freqtrade/exchange/binance.py
```diff
@@ -15,6 +16,8 @@
 
 class Binance(Exchange):
 
+    exchange_name: ExchangeName = ExchangeName.BINANCE
```
xmatthias: We do have 2 properties on the exchange for this already - name and id.
Assigning it explicitly like this will be error-prone as it needs manual maintenance as new exchanges are added.

Maybe add a property to the "base" class that returns `self.__class__.__name__` if you really need the exchange class name

freqtrade/enums/signaltype.py
```diff
@@ -7,6 +7,8 @@ class SignalType(Enum):
     """
     BUY = "buy"
     SELL = "sell"
+    SHORT = "short"
+    EXIT_SHORT = "exit_short"
```
xmatthias: I'd rather add required entries here as part of the "strategy-interface change" PR - as these additions are not correct based on the discussion in the interface change Issue.

freqtrade/enums/exchangename.py
xmatthias: I'm not sure about this enum at all. 
You can get these names easily from the exchange class directly - having this as enum again will only cause duplications without many benefits i can see at the moment.

freqtrade/enums/signaltype.py
samgermain: That should be fine, it would be helpful to keep in in case another PR needs to access it, so we don't have the changes across multiple PRs

freqtrade/enums/exchangename.py
samgermain: The only benefit really is the `OTHER` value, which is for all exchanges that don't have leverage support on freqtrade, but it also might not be necessary

freqtrade/enums/signaltype.py
samgermain: I'd like to keep it in this PR. Right now only the strategy interface needs it, but it could be useful to have in there. If you disagree could you just remove it before you merge, I could do that also

xmatthias: i think changes to this particular enum should be done when the referencing code is touched.

Right now, it's adding 2 unused values - while keeping the other 2 values identical (which is wrong, based on the desired future state).
As we'll need to have some "compat" code (ensuring compitibility of current strategies), we'll need to see if this enum will need 4 or 6 entries (keeping the existing ones for now).

So the reasoning (keep changes to one PR) will not hold.

xmatthias:
LGTM :+1:

PR #5368, opened 2021-08-06
Autopep8 formatting

samgermain:
A lot of people  use autopep8 to automatically format their code to flake8, so many that vscode recommended I install autopep8 when I cloned freqtrade. This is because the warning codes it uses correspond to the flake8 ones. 

There are a few things that flake8 doesn't catch and this PR makes those changes.

### DIFF
--- a/freqtrade/commands/build_config_commands.py
+++ b/freqtrade/commands/build_config_commands.py
@@ -193,7 +193,7 @@ def deploy_new_config(config_path: Path, selections: Dict[str, Any]) -> None:
         selections['exchange'] = render_template(
             templatefile=f"subtemplates/exchange_{exchange_template}.j2",
             arguments=selections
-            )
+        )
     except TemplateNotFound:
         selections['exchange'] = render_template(
             templatefile="subtemplates/exchange_generic.j2",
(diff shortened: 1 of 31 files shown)


xmatthias (approved):
I guess it's ok for now - although i'd like to switch to black (#3332) as formatter in the short/mid term (i suspect it'd find the same lines, though - but a lot more on top of these).

One thing to note - while you've changed this now for existing code - as it's not checked explicitly (and these are not pep8 violations per se) - it'll not be enforced for new code - so we can't guarantee this won't change again.

While you noticed that vscode suggests installing it - it also suggests (in the same popup) to eventually switch to either yarl or black ... autopep8 is just the one that's been around the longest.

samgermain:
> I guess it's ok for now - although i'd like to switch to black (#3332) as formatter in the short/mid term (i suspect it'd find the same lines, though - but a lot more on top of these).
> 
> One thing to note - while you've changed this now for existing code - as it's not checked explicitly (and these are not pep8 violations per se) - it'll not be enforced for new code - so we can't guarantee this won't change again.
> 
> While you noticed that vscode suggests installing it - it also suggests (in the same popup) to eventually switch to either yarl or black ... autopep8 is just the one that's been around the longest.

Sounds good, I mainly did it to reduce the amount of changes for newer PRs, this should be merged onto `feat/short` in order to do that.

There were also a few files I didn't change, because it just looked worse, but the main goal was to reduce differences

xmatthias:
sure, merged develop on top of feat/short :)

PR #5373, opened 2021-08-08
Added liquidation_price and interest functions

samgermain:
## Summary
Adds function for liquidation price and changes interest_mode implementation

## Quick changelog

- freqtrade/leverage/get_interest.py
- freqtrade/leverage/get_liquidation_price.py


tests/leverage/test_leverage.py
```diff
@@ -0,0 +1,108 @@
...
+    # assert get_interest(binance, borrowed=60, rate=0.00025, hours=5.0) == 1.0
+
+    # # Kraken
+    # assert get_interest(kraken, borrowed=60, rate=0.0005, hours=1.0) == 1.0
+    # assert get_interest(kraken, borrowed=60, rate=0.00025, hours=5.0) == 1.0
+
+    # # FTX
+    # assert get_interest(ftx, borrowed=60, rate=0.0005, hours=1.0) == 1.0
+    # assert get_interest(ftx, borrowed=60, rate=0.00025, hours=5.0) == 1.0
+
```
samgermain, line 31: I'll write these tests before it gets committed

freqtrade/leverage/__init__.py
```diff
@@ -1 +1,3 @@
 # flake8: noqa: F401
+from freqtrade.leverage.get_interest import get_interest
```
xmatthias: I'd call the file `interest.py` instead of `get_interrest`.

That'll make the imports sound "nicer" - as "get-interest" is an action / method name - but it's not well suited as a module i'd say

freqtrade/persistence/migrations.py
```diff
@@ -92,7 +91,7 @@ def migrate_trades_table(decl_base, inspector, engine, table_back_name: str, col
             stoploss_order_id, stoploss_last_update,
             max_rate, min_rate, sell_reason, sell_order_status, strategy, buy_tag,
             timeframe, open_trade_value, close_profit_abs,
-            leverage, interest_rate, isolated_liq, is_short, interest_mode
+            leverage, interest_rate, isolated_liq, is_short
```
xmatthias, line 94: you'll want to have `trading_mode` here instead ...

freqtrade/leverage/interest.py
```diff
@@ -0,0 +1,43 @@
...
+    Raises:
+        OperationalException: Raised if freqtrade does
+        not support margin trading for this exchange
+
+    Returns: The amount of interest owed (currency matches borrowed)
+
+    """
+    exchange_name = exchange_name.lower()
+    if exchange_name == "binance":
+        return borrowed * rate * ceil(hours)/twenty_four
```
xmatthias, line 35: Is this logic identical between futures and margin? (i'm asking - didn't look it up) ...

freqtrade/leverage/liquidation_price.py
```diff
@@ -0,0 +1,106 @@
+from freqtrade.enums import Collateral, TradingMode
+from freqtrade.exceptions import OperationalException
+
+
+def liquidation_price(
+    exchange_name: str,
+    trading_mode: TradingMode,
+    ** k
```
xmatthias, line 8: Please don't use `**kwargs` in methods where the variable is mandatory, or where it's not a user interface you might wanna expand in the future.

I know some libraries use **kwargs extensively, but it's usually very bad practice and leads to bugs.

Using explicit parameters (eventually with defaults if they're not always relevant) will allow mypy to help you detect potential problems.

freqtrade/leverage/interest.py
samgermain: There is no interest on futures, because you're trading contracts and not borrowing anything

freqtrade/leverage/liquidation_price.py
samgermain: Different exchanges might need different arguments, I'm not sure yet, collateral isn't mandatory if not using leverage

freqtrade/persistence/migrations.py
samgermain: Sorry, the trading mode addition shouldn't be in this PR. I'm actually going to split this into 2 PRs, one for liq, and one for interest

samgermain:
I'm splitting this into 2 PRs, one for liq, and one for interest

PR #5378, opened 2021-08-08
Lev-strat

### LINKED ISSUE #2183: Strategy interface change discussion - enable position management and short trading

samgermain:
## Summary
Adds shorting to the strategy interface

closes #2183 
## Quick changelog

- Short and exit short strategy methods
- Asks for side in many strategy methods


freqtrade/strategy/interface.py
```diff
@@ -760,37 +804,72 @@ def advise_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
         else:
             return self.populate_indicators(dataframe, metadata)
 
-    def advise_buy(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+    def advise_enter(
+        self,
+        dataframe: DataFrame,
+        metadata: dict,
+        is_short: bool = False
```
xmatthias: This parameter is unnecessary in my eyes.

I think we reached an agreement in #2183 - where we concluded that we'll have 2 methods instead of 4.

Obviously, users will need 4 columns  (or 6, if we include the "tag" columns) - but the seperation to 4 columns will only confuse users without real benefit.

freqtrade/templates/base_strategy.py.j2
```diff
@@ -78,8 +78,8 @@ class {{ strategy }}(IStrategy):
 
     # Optional order time in force.
     order_time_in_force = {
-        'buy': 'gtc',
-        'sell': 'gtc'
+        'enter': 'gtc',
+        'exit': 'gtc'
```
xmatthias: please do the "parameter" renaming in a seperate PR.
That makes it clear which parts had to change because of the strategy interface (column) changes, and which because of this.

freqtrade/templates/sample_hyperopt.py
```diff
@@ -172,3 +172,125 @@ def populate_sell_trend(dataframe: DataFrame, metadata: dict) -> DataFrame:
...
         return populate_sell_trend
+
+    @staticmethod
+    def short_strategy_generator(params: Dict[str, Any]) -> Callable:
+        """
+        Define the short strategy parameters to be used by Hyperopt.
+        """
+        def populate_short_trend(dataframe: DataFrame, metadata: dict) -> DataFrame:
+            """
+            Buy strategy Hyperopt will build and use.
```
xmatthias: As agreed and defined in #2183 - we will not have additional methods for this, but instead use "populate_buy_trend" / "populate_sell_trend" for this (potentially renamed to enter / exit).

xmatthias (changes requested):
I think we've got a good discussion in #2183 - where we reached a definition for the interface.
That interface will NOT add new populate_* methods - but potentially rename them (i'd probably leave that for the future - but that's up to you).

If we discuss and define things and then don't stick to that, all discussion was for nothing.
As such - i will not accept any PR that will require 4 populate* methods.

freqtrade/strategy/interface.py
samgermain: So we don't need these params either?

```
_short_fun_len: int = 0
    _exit_short_fun_len: int = 0
 ```
 
 ------------
 
 Another thing it does, is tell you whether you're entering a long or a short
 
 ```
 f"Populating {type} signals for pair {metadata.get('pair')}."
 ```

Comment (xmatthias) [reply freqtrade/strategy/interface.py]: no it doesn't ... this is just telling you which type of signals you're populating (which method is called) - not  if you're entering a long, short or nothing.

File: freqtrade/templates/sample_hyperopt.py (hunk 3 above)
Comment (samgermain) [reply freqtrade/templates/sample_hyperopt.py]: The indicator space for entering a long vs entering a short is going to be opposite. I'm 100% against having enter_short/exit_long and enter_long/exit_short having the same values returned from hyperopt, which looks like it would be the situation here if I were only going to have 2 functions, I know you said in the chat that 90% of the time the strategies will be the same, but I think that in 0 cases I would want my enter_short/exit_long values to be the same

If these values can be different, but still still with only 2 functions, then I'll do that

Comment (samgermain) [reply freqtrade/templates/sample_hyperopt.py]: Actually, I might have just been misunderstanding the hyperopt classes

Comment (xmatthias) [reply freqtrade/templates/sample_hyperopt.py]: values can be different or identical - that's not really relevant (and fully depends on your strategy/hyperopt file).

Don't spend too much time with the hyperopt class itself (consider it legacy hyperopt method) - and instead look at StrategyParameters.

It's up to you there if you have a `buy_rsi` and `buy_rsi_short` parameter - or `buy_rsi` - and compare it `> 30` and `< 30` for long/short entries.

As said before (in the interface discussion) - the `populate_{buy, sell}_trend` is just a "thinking helper" - if you assign one column ('buy') - or both buy and sell in `populate_buy_trend` will not "really" matter (unless you overwrite it in the other method).

File: freqtrade/strategy/interface.py [hunk 4]
```diff
@@ -530,15 +535,22 @@ def get_signal(
             )
             return False, False, None
 
-        enter = latest[SignalType.BUY.value] == 1
+        (enter_type, enter_tag) = (
+            (SignalType.SHORT, SignalTagType.SHORT_TAG)
+            if is_short else
+            (SignalType.BUY, SignalTagType.BUY_TAG)
+        )
```
xmatthias: I wonder how you imagine this to work.

The decision if a trade-entry "is_short" is taken right here - so you have no way to have an `is_short` variable passed into this function that is correct (at least not for trade entries - but it's irrelevant for exits too).

The only thing you will have is a config with "short_allowed" (or whatever the naming is) - which will tell you if you can check for longs and shorts, or only for long (in which case shorts will simply be set to False) - but which way the trade is going to go will be decided by the strategy.

freqtrade/strategy/hyper.py
```diff
@@ -22,6 +22,8 @@
 
 logger = logging.getLogger(__name__)
 
+# TODO-lev: This file
```
xmatthias: I don't see how this file would be relevant for leverage.

You can use the "buy" space for that and use an integer or whatever else you need.
Adding a "leverage" space should not be necessary, at least not for the moment (That'll complicate things for no good reason).

freqtrade/strategy/interface.py
```diff
@@ -164,7 +164,7 @@ def populate_sell_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame
 
     def check_buy_timeout(self, pair: str, trade: Trade, order: dict, **kwargs) -> bool:
         """
-        Check buy enter timeout function callback.
+        Check buy timeout function callback.
```
xmatthias, line 190: I think the "enter" docstring was better suited here.

freqtrade/templates/sample_hyperopt_advanced.py
```diff
@@ -153,46 +183,67 @@ def populate_sell_trend(dataframe: DataFrame, metadata: dict) -> DataFrame:
...
+                    'exit_short'] = 1
+
             return dataframe
 
         return populate_sell_trend
 
     @staticmethod
     def generate_roi_table(params: Dict) -> Dict[int, float]:
         """
+        # TODO-lev?
```
xmatthias: Don't touch these files (`templates/*hyperopt*`) - it's wasted time.

it's the legacy hyperopt interface - which is more likely to go away than to get any new features.

freqtrade/templates/sample_hyperopt.py
```diff
@@ -55,6 +55,10 @@ def indicator_space() -> List[Dimension]:
             Integer(15, 45, name='fastd-value'),
             Integer(20, 50, name='adx-value'),
             Integer(20, 40, name='rsi-value'),
+            Integer(75, 90, name='short-mfi-value'),
```
xmatthias: Same as below - leave the "sample hyperopt" files alone. 

Short optimization will only be supported with the parameter interface - as the hyperopt interface with seperate hyperopt files is kindof deprecated.

freqtrade/templates/sample_strategy.py
```diff
@@ -58,6 +58,8 @@ class SampleStrategy(IStrategy):
     # Hyperoptable parameters
     buy_rsi = IntParameter(low=1, high=50, default=30, space='buy', optimize=True, load=True)
     sell_rsi = IntParameter(low=50, high=100, default=70, space='sell', optimize=True, load=True)
+    short_rsi = IntParameter(low=51, high=100, default=70, space='sell', optimize=True, load=True)
+    exit_short_rsi = IntParameter(low=1, high=50, default=30, space='buy', optimize=True, load=True)
```
xmatthias: Please leave the "sample_strategy" alone (keep it as long-only strategy).

instead, add a "sample_short_strategy" (naming up for discussion). 
This strategy is intended to show people how to create a strategy and serve as a sample - and i doubt everyone will want a long/short strategy (maybe because the exchange/markets people using are long only).

tests/optimize/hyperopts/default_hyperopt.py
```diff
@@ -54,38 +54,57 @@ def populate_buy_trend(dataframe: DataFrame, metadata: dict) -> DataFrame:
             """
             Buy strategy Hyperopt will build and use.
             """
-            conditions = []
+            long_conditions = []
+            short_conditions = []
```
xmatthias: First of all - to ensure compatibility (existing things cannot break because of short support)- please do not touch these files (this and the corresponding strategy file).

We will have to have a short test file too - but in my opinion, not a explicit hyperopt file (see above - deprectated interface) - but only a seperate parametrized short strategy.

tests/strategy/test_strategy_loading.py
```diff
@@ -218,6 +218,7 @@ def test_strategy_override_process_only_new_candles(caplog, default_conf):
 def test_strategy_override_order_types(caplog, default_conf):
     caplog.set_level(logging.INFO)
 
+    # TODO-lev: Maybe change
```
xmatthias: Leverage is going to be a callback, not a configuration setting (`enable-leverage` might be - but i doubt we need that here as the logic for config/strategy overrides is well tested and identical for all settings). 

Having leverage a static (config) setting would not work - simply because "elsewhere" you're showing that allowed leverage will differ between pairs and traded volume.

xmatthias (changes requested):
Please be careful with the test/sample strategies when changing the strategy interface.

While we will want to have "short" versions of them - we should also have their long only counterparts - as many traders prefer a long only strategy (maybe  because their exchange ain't allowing shorts, or maybe out of belive).

Also, changing the test-strategies directly endangers backwards compatibility - so i'd prefer to leave them untouched as much as possible - as they will serve as a first layer of tests (on top of manual tests) to ensure that the compatibility stays in place.

freqtrade/strategy/interface.py
samgermain: What do you mean the decision is taken right here? This method is called in freqtradebot with the lines

```
(buy, sell, buy_tag) = self.strategy.get_signal(
            pair,
            self.strategy.timeframe,
            analyzed_df
        )
 ```
 
 which I plan to edit and use
 
 ```
(buy, sell, buy_tag) = self.strategy.get_signal(
            pair,
            self.strategy.timeframe,
            analyzed_df,
            is_short=True
        )
(short, exit_short, short_tag) = self.strategy.get_signal(
            pair,
            self.strategy.timeframe,
            analyzed_df,
            is_short=False
        )
 ```

and since both `buy` and `enter_short` will be the dataframe, then 
`        enter = latest[enter_type.value] == 1` will return different values based on what `enter_type.value` is

A lot of the logic in `freqtradebot` is not the same for shorts, and can't be combined into one method

File: freqtrade/strategy/interface.py (hunk 6 above)
Comment (samgermain) [reply freqtrade/strategy/interface.py:190]: I just thought the comment should remain the same as the method name

File: freqtrade/templates/sample_hyperopt_advanced.py (hunk 7 above)
Comment (samgermain) [reply freqtrade/templates/sample_hyperopt_advanced.py:246]: What is the legacy hyperopt and what is the new hyperopt?

File: freqtrade/strategy/interface.py (hunk 4 above)
Comment (samgermain) [reply freqtrade/strategy/interface.py]: So I tried changing

```
(buy, sell, buy_tag) = self.strategy.get_signal(
            pair,
            self.strategy.timeframe,
            analyzed_df
        )
 ```
 
 to be
 
 ```
 (buy, sell, buy_tag) = self.strategy.get_signal(
            pair,
            self.strategy.timeframe,
            analyzed_df
        )
 ```
 
 but got an error that 4 arguments were given but 3 were expected, which I didn't understand, because with my `strategy.get_signal` method, `is_short` is a parameter

freqtrade/templates/sample_strategy.py
samgermain: What about just commenting out the short parts? I made a `sample_short_strategy` and 99% of the code was the same, so it seems like unnecessary effort for anyone to read the `sample_short_strategy` file as well

tests/optimize/hyperopts/default_hyperopt.py
samgermain: What is the legacy hyperopt and what is the new hyperopt? Also, if it's legacy, can we add a Deprecated warning or something, and could we do that for everything else that's legacy so that I don't write updates for any other legacy code? I don't see anything in there that indicates it's legacy code

I created a short hyperopt for now, which is just the changes I had for hyperopt, but reverted the regular hyperopt file back

xmatthias: parametrized hyperopting (hyperoptable parameters in the strategy) are the new hyperopt method.

The old / legacy method is using a seperate file.
While we're not showing a deprecation warning (intentionally for now) - the documentation does mention this as [legacy hyperopt](https://www.freqtrade.io/en/stable/advanced-hyperopt/#legacy-hyperopt).

freqtrade/templates/sample_strategy.py
xmatthias: if i'm a new user looking for a long only strategy sample / template- i'll not want to read through however many lines that are not relevant to my goal at all.
It's however unclear if that must include 200 new lines - or 3 new lines.

can we for the moment simply leave the templates folder untouched - and work on that last? (it's not that important at this stage, and work there now is basically just "legwork" - not the complicated stuff).


the "sample's" we're looking at now are (in theory) only generated from the .j2 templates (using `freqtrade new-strategy ...`).
We'll see how much / little the effort / difference will be once the interface itself is finished / as discussed in the issue (which i can't see in this PR at all yet, but was expecting from it (at least the column namings including compat code for that).

freqtrade/strategy/interface.py
xmatthias: That's not a very efficient method to search for signals - as it's doing a lot of work twice without any reason.
you'll see how i've solved that.

samgermain: > That's not a very efficient method to search for signals - as it's doing a lot of work twice without any reason.
> you'll see how i've solved that.

Thanks for doing that, what would you say is left for this PR?

xmatthias: i'm working on it ... but there's a lot still missing (excluding trades) - as to properly have the strategy interface, at least backtesting must be working - otherwise we can't know if the new interface is working the way it was designed.

samgermain:
@xmatthias This is the next major short/lev PR that should be merged, let me know if there's anything I can help out with here.

Otherwise, I'll be working on `funding-fee` or cleaning up code and changing prices to USDT in `test_freqtradebot`

freqtrade/edge/edge_positioning.py
```diff
@@ -159,7 +159,8 @@ def calculate(self, pairs: List[str]) -> bool:
         logger.info(f'Measuring data from {min_date.strftime(DATETIME_PRINT_FORMAT)} '
                     f'up to {max_date.strftime(DATETIME_PRINT_FORMAT)} '
                     f'({(max_date - min_date).days} days)..')
-        headers = ['date', 'buy', 'open', 'close', 'sell', 'high', 'low']
+        #  TODO-lev: Should edge support shorts? needs to be investigated further...
```
samgermain, line 162: probably, i don't see why not, I also don't know how to use edge, it's like a trailing stoploss right?

freqtrade/freqtradebot.py
```diff
@@ -422,24 +422,25 @@ def create_trade(self, pair: str) -> bool:
...
+            )
 
-        if buy and not sell:
+        if side:
             stake_amount = self.wallets.get_trade_stake_amount(pair, self.edge)
 
             bid_check_dom = self.config.get('bid_strategy', {}).get('check_depth_of_market', {})
             if ((bid_check_dom.get('enabled', False)) and
                     (bid_check_dom.get('bids_to_ask_delta', 0) > 0)):
+                # TODO-lev: Does the below need to be adjusted for shorts?
```
samgermain, line 435: Already done in `lev-freqtradebot`

freqtrade/freqtradebot.py
```diff
@@ -422,24 +422,25 @@ def create_trade(self, pair: str) -> bool:
...
                 if self._check_depth_of_market_buy(pair, bid_check_dom):
-                    return self.execute_entry(pair, stake_amount, buy_tag=buy_tag)
+                    # TODO-lev: pass in "enter" as side.
```
samgermain, line 437: Already done in `lev-freqtradebot`

freqtrade/freqtradebot.py
```diff
@@ -703,22 +705,23 @@ def handle_trade(self, trade: Trade) -> bool:
...
-            (buy, sell, _) = self.strategy.get_signal(
+            (enter, exit_) = self.strategy.get_exit_signal(
                 trade.pair,
                 self.strategy.timeframe,
-                analyzed_df
+                analyzed_df, is_short=trade.is_short
             )
 
-        logger.debug('checking sell')
+        # TODO-lev: side should depend on trade side.
```
samgermain, line 727: Already done in `lev-freqtradebot`

freqtrade/plot/plotting.py
```diff
@@ -386,8 +386,9 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
     )
     fig.add_trace(candles, 1, 1)
 
-    if 'buy' in data.columns:
-        df_buy = data[data['buy'] == 1]
+    # TODO-lev: Needs short equivalent
```
samgermain, line 389: Already done in `lev-freqtradebot`

freqtrade/strategy/interface.py
```diff
@@ -209,6 +209,7 @@ def bot_loop_start(self, **kwargs) -> None:
         """
         pass
 
+    # TODO-lev: add side
```
samgermain: Already done in `lev-freqtradebot`

freqtrade/strategy/interface.py
```diff
@@ -343,6 +344,7 @@ def custom_sell(self, pair: str, trade: Trade, current_time: datetime, current_r
         """
         return None
 
+    # TODO-lev: add side
```
samgermain: Already done in `lev-freqtradebot`

freqtrade/edge/edge_positioning.py
xmatthias: no it's not - it's a whole seperate backtesting engine - but one which is VERY limited, cannot (by design) support any callbacks, nor ROI or trailing stop) - so the only sell possibility is via stoploss (static) or sell-signal (this is true for the edge-simulation that's running, anyway).

I'm surely not looking at that part in this pr - mainly because i personally don't see the value in it (and because the leverage callback would not work for this).

samgermain: it seems like it would be really useful in determining how much leverage to use, you could use it to find an optimal stoploss and use a leverage that creates a liquidation price below that stoploss

xmatthias: except that it's not working for about 70% of the strategies, because they either use `custom_sell` for exiting, or rely heavily on ROI / trailing stop - neither of which can work in edge (because it's designed to be 100% vectorized).

samgermain: All I'm saying is that when you apply leverage to a trade, the ideal amount of leverage to apply is an amount that is as high as possible but set's a liquidation price that your trade will not fall to. So maybe in it's current form edge wouldn't be useful, but sounds pretty close to something that can choose the ideal leverage for you

freqtrade/freqtradebot.py
```diff
@@ -579,7 +580,8 @@ def execute_entry(self, pair: str, stake_amount: float, price: Optional[float] =
             exchange=self.exchange.id,
             open_order_id=order_id,
             strategy=self.strategy.get_strategy_name(),
-            buy_tag=buy_tag,
+            # TODO-lev: compatibility layer for buy_tag (!)
```
samgermain, line 588: What does this mean?

xmatthias: what it says? that we'll need a compatibility layer for the buy-tag field (not necessarily here, but probably in persistence).
Strategies use this field as part of callbacks to determine the "matching" sell strategy - simply renaming the field to `entry_tag` (or whatever else) will break people's strategies - so it'll need a compatibility layer.

freqtrade/plot/plotting.py
samgermain: @xmatthias Actually I don't think this is true, sorry that I wrote this

xmatthias: I didn't intend to fix these here (none of the ones you commented on), nor remove the comments.
i think it's still good to keep these notes "there" until we have the "fix" or solution in place (basically the PR implementing that will remove these lines) - so we can make sure we're not forgetting these things.

There's a thought behind every "todo" (both from my side, and/or from other people's sides) - reminding to not forget stuff - so it's important to keep them or conciously decide "i did this - can remove it".

freqtrade/edge/edge_positioning.py
xmatthias: I'm not against that - but i 
* don't intend to implement that in this pr
* don't think it's a necessity for short - at least not to start with - so i'd propose to postpone this until "whenever" ... i'd focus on MVP (minimum viable product) with this - not on having 100% of the things there (which is why i'm not keen on supporting all of the different order-types to start with).

freqtrade/enums/signaltype.py
```diff
@@ -5,12 +5,20 @@ class SignalType(Enum):
...
+    EXIT_SHORT = "exit_short"
 
 
 class SignalTagType(Enum):
     """
     Enum for signal columns
     """
-    BUY_TAG = "buy_tag"
+    LONG_TAG = "long_tag"
+    SHORT_TAG = "short_tag"
```
xmatthias: @samgermain Maybe a strange question at this point...
do we really need 2 seperate columns for this?

if long and short entries collide, we'll reject the trade anyway (it's a shitty strategy which isn't sure wether to go long or short), so maybe having this column named `entry_tag` instead would simplify usage from the strategy a bit (one fewer column-name to remember for users)...

In the end, we'll also only store it once in the database ... as a trade is EITHER long OR short ... not both.

freqtrade/optimize/backtesting.py
```diff
@@ -45,7 +45,7 @@
 ELONG_IDX = 6  # Exit long
 SHORT_IDX = 7
 ESHORT_IDX = 8  # Exit short
-BUY_TAG_IDX = 9
+ENTER_TAG_IDX = 9
 SHORT_TAG_IDX = 10
```
samgermain: This one should be `EXIT_TAG_IDX` to match right?

freqtrade/enums/signaltype.py
Axel-CH: Hello @xmatthias take a look at [this]( https://youtu.be/Yq-m6IQ3FHw).
As you can see having both position open for a period can be a legit strategy 🙂

samgermain: > Hello @xmatthias take a look at [this](https://youtu.be/Yq-m6IQ3FHw).
> As you can see having both position open for a period can be a legit strategy 🙂

We're not dealing with hedge mode for now, to keep the complexity down

xmatthias: @Axel-CH if both `enter_long` and `enter_short` values are 1 for the same candle, the bot will do nothing, as it's a conflicting signal.

Hedge-mode is not in scope for the moment (and probably won't be for a long time) - so it'll be one position (side) per trade only - in which case having split columns will not give us many benefits.

Also, it may be a valid strategy (i understand that strategy ...) - almost no crypto-exchange allows you to go long and short in the same pair (except for binance) - so it's an edge-case we'll not cover for the moment (and potentially never, as binance seems to disable futures completely for most of the world, like US and most of Europe ...).

freqtrade/optimize/backtesting.py
xmatthias: no, there is no exit tag, if any, it's `LONG_TAG` and `SHORT_TAG`. 
but as mentioned in the other comment - i'm not convinced on having the 2 seperated will give us benefits.

freqtrade/enums/signaltype.py
Axel-CH: @xmatthias I totally understand the necessity to keep things simple. As you are maintaining this project for years now, you have a good view of what can add complexity and complicate debugging.

However, just on the strategy point of view I'm not shocked by something like: "I want to enter_long at this price and short at that price" In that case both order will be sent to the exchange but it's the market price that will decide which one is filled.

Anyway, as you said it's a specific case and if it's adding to much complexity, it's a good question to ask if it worth it.

xmatthias: this case is a different one alltogether - you'd place two orders (one long, one short) - and assume that the filled one will cancel the other one (?)

I'm not sure that'll be a good approach ... as that assumes that price is moving against you - so you're long when price falls, and short when price raises (that's the wrong way around - at least when trying to make profits!)

A "real" Hedge would have you in both positions at the same time (assume both orders fill) - but unless explicitly allowed by the exchange (binnace hedge mode), that's not something that'll work as the 2nd fill will close the other position (you buy 100 XRP, and sell 100XRP - so your position is now 0 XRP).

Axel-CH: When the price moove in trend, there is two types of price movements.
Impulsion: The price strongly follow trend direction 

Correction: The price perform a pullback as traders take profit at key levels 

If you are greedy (😂)you want to take advantage of both cases. 

The impulsion wave is where you want to make your profit. But you can use the correction to make a better entry and ride the next impulsion wave. So yes you are entering in the trade when the price go against you for a short time, but you still follow the overall trend. 

In that case if you also want to take advantage of the correction wave you can enter position at the begining of it and exit at the same level you want to enter the trend following order. So technically freqtrade will have both "position" open. But in reality only one side is filled and active on exchange. 

I Will do a shema to make it clearer

Axel-CH: Also, to prevent unwanted opening of a position on the opposition side, we can use the "reduce only" option when we create orders. [More details ](https://help.bybit.com/hc/en-us/articles/360039260574-What-is-a-Reduce-Only-Order-)
I think we should add it by default

xmatthias: Freqtrade does not support bybit at this point - so i don't see this order-type as relevant.

even in your case (which for sure will not be supported to begin with) - it'd be the same signal that indicates buying and selling (indicates this type of entry) - so i don't really see the need to have 2 seperate columns for the entry reason.

It's a case / scenario where i feel we're trying to prepare for an eventual future which potentially never arrives - and as a consequence, making our life (or the code) more complex than how it really should be.

Axel-CH: Reduce only orders are available on binance, ftx, kraken...etc

xmatthias: that's totally unrelated to the discussion here ... so i don't see how this can help in deciding if short_tag needs to be a seperate column or not.

Axel-CH: I just checked on the create_trade code. To manage my specific case of having two signals on the same candle an option could be added to create _trade to specify the side wanted.

In that case global tag for long and short won't be an issue.
The downside will be the duplicate execution of all code inside create_trade function.

But as we agreed to not do
(enter_long, enter_short, exit_long, exit_short) = get_signal()

Having only one entry_tag seem logic to me

Axel-CH: I spoke about the "reduceOnly" option to answer to your question as this feature prevent orders to conflict

"this case is a different one alltogether - you'd place two orders (one long, one short) - and assume that the filled one will cancel the other one (?)"
👍🏽

freqtrade/freqtradebot.py
```diff
@@ -422,24 +422,25 @@ def create_trade(self, pair: str) -> bool:
             return False
 
         # running get_signal on historical data fetched
-        (buy, sell, buy_tag) = self.strategy.get_signal(
-            pair,
-            self.strategy.timeframe,
-            analyzed_df
-        )
+        (side, enter_tag) = self.strategy.get_entry_signal(
```
samgermain: maybe signal would be a better name than side?

xmatthias (approved):
I think that's as far as i'm ready to go with this PR, so i'd leave it as is for now.

I suspect we'll still need to do do some minor changes "here and there" - but the huge part should be fine now.

samgermain:
🎉

PR #5379, opened 2021-08-08
Liquidation price

samgermain:
## Summary
Adds liquidation price formula function and maintenance margin class

## Quick changelog

- liquidation_price and maintenance_margin are mostly skeleton functions without actual implementation
- freqtradebot executes functions from maintenance_margin

## What's new?

New files
- freqtrade/leverage/liquidation_price.py
- freqtrade/maintenance_margin.py

freqtradebot gets 3 new properties
- maintenance_margin
- collateral
- trading_mode


freqtrade/leverage/liquidation_price.py
```diff
@@ -0,0 +1,106 @@
+from freqtrade.enums import Collateral, TradingMode
+from freqtrade.exceptions import OperationalException
+
+
+def liquidation_price(
+    exchange_name: str,
+    trading_mode: TradingMode,
+    ** k
+):
+
```
samgermain: > Please don't use **kwargs in methods where the variable is mandatory, or where it's not a user interface you might wanna expand in the future.
> 
> I know some libraries use **kwargs extensively, but it's usually very bad practice and leads to bugs.
> 
> Using explicit parameters (eventually with defaults if they're not always relevant) will allow mypy to help you detect potential problems.

Different exchanges and trading_mode's might need different arguments, I'm not sure yet though

Also, Collateral isn't required if trading_mode is SPOT, but I suppose I could make it an Optional

xmatthias: as mentioned above - specify all parameters explicitly. Parameters that are not needed in all cases can be Optional (with a default of None or so).

The place you're going to need to call this from (`freqtradebot`, most like) will NOT have exchange-specific code - so you'll either be providing a nested structure (an order response or similar) - OR all parameters for all exchanges.

freqtrade/enums/marginmode.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+
+class MarginMode(Enum):
+    """
+        Enum to distinguish between
+        one-way mode or hedge mode in Futures (Cross and Isolated) or Margin Trading
+    """
+    ONE_WAY = "one-way"
+    HEDGE = "hedge"
```
xmatthias, line 10: i assume this is the binance futures "hedge mode" ? 

To be honest, i think if we don't start to keep things simple (focus on a few selected modes supported by multiple exchanges) - this will never have a chance to be merged.

Personally, i'd have focused on cross-margin on binance, "regular short" (probably still margin) on kraken - and futures on FTX (they seem to all work pretty similarily).
These seem pretty straightforward at a first glance (what binance futures clearly does not).

Additional modes can always be added at a later point - but adding too much in one go will make everything take forever - and we'll also want to do real trading tests of a few 100 trades for each mode/exchange before merging ... so having too many exchanges WILL become problematic for sure.

samgermain: > (what binance futures clearly does not).

What part about binance futures isn't straight forward?

Also, I didn't write this code, but I think this enum could be good to throw in, but we should still decide that we're only going to support `ONE_WAY` for now, then we can just have checks for something like

```
if marginmode == 'one_way':
    use one_way formula
else:
    throw exception
```

I probably won't implement hedge mode in the future, but if we implement it in this way, then if anyone does want to implement hedge mode, it will be clear that we've only implemented one_way mode, and where some of the changes are that need to be made for hedge mode.

But for the the current `feat/short` branch, we would only have some conditionals that check for the current marginmode, an extra parameter on some methods, and some variables with a default value of one_way

-----------------

That's just my opinion on it, it seems like it would be easier to put in the minimum bits of code in places when you know that the code needs to be there, rather than have someone try to figure out how to change the code in order to add these features in the future

If you don't want to do that, then I think we should at least have some blaring comments that mention certain things are only for one_way mode, or maybe put some TODO's in there for what you would have to do for hedge mode

arunavo4: @xmatthias These modes are only supported for Binance so for other exchanges its just simple margin trading which will  not have any modes so if we want to test it we can just test it for binance the check is already there for that.

xmatthias: My point is - we should focus on the simple things to get the whole short/leverage stuff working.
If we start supporting the "special binance modes" (whichever ones they are) - it'll become a very long endavor (please consider that we'll have to live-test each and every one of these different modes, for several trades (not just 2-3)) - so each additional mode will probably add at least 2-3 weeks of testing time alone (not talking about development time).

tests/leverage/test_liquidation_price.py
```diff
@@ -0,0 +1,113 @@
...
+    # TODO-lev assert exception is thrown
+    return  # Here to avoid indent error, remove when implemented
+
+
+@pytest.mark.parametrize(
+    'exchange_name,open_rate,is_short,leverage,trading_mode,collateral,result', [
+        # Binance
+        ('binance', "2.0", False, "1.0", margin, cross, 1.0),
+        ('binance', "2.0", False, "1.0", futures, cross, 1.0),
+        ('binance', "2.0", False, "1.0", futures, isolated, 1.0),
```
samgermain, line 87: @arunavo4 For all the formulas you wrote, make sure that you write out a few tests here

freqtrade/enums/marginmode.py
samgermain: @arunavo4 If you want to make it easier to get your code for hedge mode back, I would finish up the regular liquidation formulas with their tests, then after that's done, remove the hedge code in a single commit. Then when you want to add it back, you could just revert that commit

arunavo4: Yea lets do that keep that code to add later

tests/leverage/test_liquidation_price.py
arunavo4: sure I will add the tests

freqtrade/enums/marginmode.py
samgermain: > Yea lets do that keep that code to add later

Just make a good commit message that you can find later

samgermain:
moved to freqtrade branch

samgermain:
@arunavo4 There's a liq branch now, those 2 commits that you made aren't on it yet, you should create another PR and request to mere onto liq, once the tests and everything are in there I'll merge it in

PR #5380, opened 2021-08-08
Replaced the term margin with leverage when it should say leverage (All comments)

samgermain:
## Summary
Margin != leverage, even though before this the two terms were used as synonyms

## Quick changelog

- Changes to comments where `mg` change to `lev` and `margin` changed to `leverage`


xmatthias (approved):
LGTM :+1:

PR #5381, opened 2021-08-08
Changed interest implementation (Margin)

samgermain:
## Summary
Interest calculation is in a function instead of an enum

## Quick changelog

- Added freqtrade/leverage/interest.py
- removed freqtrade/enums/interestmode.py


freqtrade/leverage/interest.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Returns: The amount of interest owed (currency matches borrowed)
+
+    """
+    exchange_name = exchange_name.lower()
+    if exchange_name == "binance":
+        return borrowed * rate * ceil(hours)/twenty_four
+    elif exchange_name == "kraken":
+        # Rounded based on https://kraken-fees-calculator.github.io/
+        return borrowed * rate * (one+ceil(hours/four))
```
samgermain, line 37: >  Is this logic identical between futures and margin? (i'm asking - didn't look it up) ...

There is no interest on futures. You're trading contracts, so therefore you're not borrowing anything

freqtrade/leverage/interest.py
```diff
@@ -0,0 +1,43 @@
...
+        not support margin trading for this exchange
+
+    Returns: The amount of interest owed (currency matches borrowed)
+
+    """
+    exchange_name = exchange_name.lower()
+    if exchange_name == "binance":
+        return borrowed * rate * ceil(hours)/twenty_four
+    elif exchange_name == "kraken":
+        # Rounded based on https://kraken-fees-calculator.github.io/
```
xmatthias, line 36: If the calculations are expected to be completely different ...
maybe it would make sense to move this to the respective exchange-classes? 

The only premise for that would be that we should try not to call that from persistence (which could be a dealbreaker, and which is why i'm posting it here as a question).

samgermain: Excluding it from persistence is the only reason It's not just in the exchange class

freqtrade/leverage/interest.py
```diff
@@ -0,0 +1,42 @@
...
+        return borrowed * rate * (one+ceil(hours/four))
+    elif exchange_name == "ftx":
+        # TODO-lev: Add FTX interest formula
+        raise OperationalException(f"Leverage not available on {exchange_name} with freqtrade")
+    else:
+        raise OperationalException(f"Leverage not available on {exchange_name} with freqtrade")
```
xmatthias, line 42: It's fine to leave this here - but in the end (once feat/short) is ready to roll - we will need to have checks in place during startup that will prevent this line from ever being hit (it can result in a bot running for days - until a trade is opened before it crashes).

tests/leverage/test_leverage.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+# from freqtrade.exceptions import OperationalException
+binance = "binance"
+kraken = "kraken"
+ftx = "ftx"
+other = "bittrex"
+
+
+def test_interest():
```
xmatthias: I think this test should be parametrized (`@pytest.mark.parametrize()` ...) (i've pushed a commit adding that now).
I hope this can serve you as an example / template for future such tests ... as i expect we'll still have quite some - and parametrizing them makes adding additional tests (for longer times, for example) easy.

That way, we can have a lot more testcases (with different values and expects) with a lot less code duplication. Even if it's test code - we still need to maintain it (assume we change a parameter name sometime in the future ...) i'd prefer to do that change once - not 6 times...

xmatthias (approved):
Please take a look at my test-commit - that's what i mean when i sometimes said "this test should be parametrized".

Hope this can serve as an example for future tests.

LGTM :+1:

freqtrade/leverage/interest.py
samgermain: I think that a better way to do this would be using `@classmethod`, and then overriding in each subclass, and then just doing something like `getattr(Exchange, self.exchange)`, it will work for `funding_fee` and `liquidation_price` as well, and I think it's just cleaner

Do you know what the `getattr` part would look like exactly? I don't think `getattr(Exchange, self.exchange)` will work

PR #5389, opened 2021-08-09
Update README.md

pcassimans:
Fix Typo of Kukoin to Kucoin


xmatthias (approved):
Great! :+1: 

I've fixed this in the other file we have the same mistake, too :laughing:

PR #5399, opened 2021-08-11
Fix buy_tag not being saved to trade object.

rokups:
Column is mistakenly excluded because advise_buy() creating this column runs after code detecting presence of buy_tag column.

Is there really a point in trying to exclude this column?

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -218,7 +218,7 @@ class Backtesting:
         """
         # Every change to this headers list must evaluate further usages of the resulting tuple
         # and eventually change the constants for indexes at the top
-        headers = ['date', 'buy', 'open', 'close', 'sell', 'low', 'high']
+        headers = ['date', 'buy', 'open', 'close', 'sell', 'low', 'high', 'buy_tag']
         data: Dict = {}
         self.progress.init_step(BacktestState.CONVERT, len(processed))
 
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -218,21 +218,18 @@ def _get_ohlcv_as_lists(self, processed: Dict[str, DataFrame]) -> Dict[str, Tupl
         """
         # Every change to this headers list must evaluate further usages of the resulting tuple
         # and eventually change the constants for indexes at the top
-        headers = ['date', 'buy', 'open', 'close', 'sell', 'low', 'high']
+        headers = ['date', 'buy', 'open', 'close', 'sell', 'low', 'high', 'buy_tag']
```
xmatthias, line 221: This will fail for all strategies that don't define a buy signal, because the column will not exist.

That part was handled in L230...

xmatthias (commented):
are you sure that the bug you're trying to fix is here, and not in the strategy not assigning the buy_tag correctly?

freqtrade/optimize/backtesting.py
```diff
@@ -218,21 +218,18 @@ def _get_ohlcv_as_lists(self, processed: Dict[str, DataFrame]) -> Dict[str, Tupl
...
         data: Dict = {}
         self.progress.init_step(BacktestState.CONVERT, len(processed))
 
         # Create dict with data
         for pair, pair_data in processed.items():
             self.check_abort()
             self.progress.increment()
-            has_buy_tag = 'buy_tag' in pair_data
-            headers = headers + ['buy_tag'] if has_buy_tag else headers
```
xmatthias, line 230: I guess the correct approach here will be to move this part down below the `advise_*` calls.

Removing it will however break backwards compatibility (force all strategies to use buy_tag).

freqtrade/optimize/backtesting.py
rokups: Why would it fail? `pair_data.loc[:, 'buy_tag'] = None` initializes column to empty value, strategy does not have to define it.

xmatthias: hm - seems i missed that line - and was slightly distracted by the failing tests :laughing:

rokups: Hah, and here im thinking this is so simple that there is no point in running tests before submitting PR 😅 Anyhow fixed tests with a two line hack. Or if you want i can fix them with 200 changed lines by always submitting buy tag..?

xmatthias (approved):
Nice catch 💯 

LGTM 👍

PR #5439, opened 2021-08-17
Added psycopg2-binaries for postgres support

richardbeumer:
to use postgres database this package needs to be included

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,7 @@ pycoingecko==2.2.0
 jinja2==3.0.1
 tables==3.6.1
 blosc==1.10.4
+psycopg2-binary==2.9.1
 
 # find first, C search in arrays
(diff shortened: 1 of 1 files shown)


xmatthias:
To be honest, this is known and [documented](https://www.freqtrade.io/en/latest/sql_cheatsheet/#postgresql).

It's intentional that freqtrade does not depend on this dependency (nor on the mysql counterpart), as it's not used by ~90% of the users - which are perfectly fine running on sqlite.

If any - we can add it as opt-dependency in the setup.py (so running `pip install freqtrade[postgres]`) will install it - but i don't see us adding 3.5+ MB to the docker image for no good reason.

richardbeumer:
Ok that explains why this depency was not present. I will close this pull request.

PR #5447, opened 2021-08-18
Lev strat name changes

samgermain:
Name changes to freqtrade methods, variables and comments to be side agnostic and mention shorting

## Summary
Give name changes so that there are less differences between lev-strat PR and feat/short branch

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -285,13 +285,11 @@ class Hyperopt:
         # Apply parameters
         if HyperoptTools.has_space(self.config, 'buy'):
             self.backtesting.strategy.advise_buy = (  # type: ignore
-                self.custom_hyperopt.buy_strategy_generator(params_dict)
-            )
+                self.custom_hyperopt.buy_strategy_generator(params_dict))
 
         if HyperoptTools.has_space(self.config, 'sell'):
             self.backtesting.strategy.advise_sell = (  # type: ignore
-                self.custom_hyperopt.sell_strategy_generator(params_dict)
-            )
+                self.custom_hyperopt.sell_strategy_generator(params_dict))
 
         if HyperoptTools.has_space(self.config, 'protection'):
             for attr_name, attr in self.backtesting.strategy.enumerate_parameters('protection'):
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -285,11 +285,13 @@ class Hyperopt:
         # Apply parameters
         if HyperoptTools.has_space(self.config, 'buy'):
             self.backtesting.strategy.advise_buy = (  # type: ignore
-                self.custom_hyperopt.buy_strategy_generator(params_dict))
+                self.custom_hyperopt.buy_strategy_generator(params_dict)
+            )
 
         if HyperoptTools.has_space(self.config, 'sell'):
             self.backtesting.strategy.advise_sell = (  # type: ignore
-                self.custom_hyperopt.sell_strategy_generator(params_dict))
+                self.custom_hyperopt.sell_strategy_generator(params_dict)
+            )
 
         if HyperoptTools.has_space(self.config, 'protection'):
             for attr_name, attr in self.backtesting.strategy.enumerate_parameters('protection'):
--- a/freqtrade/resolvers/strategy_resolver.py
+++ b/freqtrade/resolvers/strategy_resolver.py
@@ -193,18 +193,22 @@ class StrategyResolver(IResolver):
                 # register temp path with the bot
                 abs_paths.insert(0, temp.resolve())
 
-        strategy = StrategyResolver._load_object(paths=abs_paths,
-                                                 object_name=strategy_name,
-                                                 add_source=True,
-                                                 kwargs={'config': config},
-                                                 )
+        strategy = StrategyResolver._load_object(
+            paths=abs_paths,
+            object_name=strategy_name,
+            add_source=True,
+            kwargs={'config': config},
+        )
+
         if strategy:
             strategy._populate_fun_len = len(getfullargspec(strategy.populate_indicators).args)
             strategy._buy_fun_len = len(getfullargspec(strategy.populate_buy_trend).args)
             strategy._sell_fun_len = len(getfullargspec(strategy.populate_sell_trend).args)
-            if any(x == 2 for x in [strategy._populate_fun_len,
-                                    strategy._buy_fun_len,
-                                    strategy._sell_fun_len]):
+            if any(x == 2 for x in [
+                strategy._populate_fun_len,
+                strategy._buy_fun_len,
+                strategy._sell_fun_len
+            ]):
                 strategy.INTERFACE_VERSION = 1
 
             return strategy
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -135,7 +135,7 @@ class IStrategy(ABC, HyperStrategyMixin):
     @abstractmethod
     def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
         """
-        Populate indicators that will be used in the Buy and Sell strategy
+        Populate indicators that will be used in the Buy, Sell, Short, Exit_short strategy
         :param dataframe: DataFrame with data from the exchange
         :param metadata: Additional information, like the currently traded pair
         :return: a Dataframe with all mandatory indicators for the strategies
@@ -164,9 +164,9 @@ class IStrategy(ABC, HyperStrategyMixin):
 
     def check_buy_timeout(self, pair: str, trade: Trade, order: dict, **kwargs) -> bool:
         """
-        Check buy timeout function callback.
-        This method can be used to override the buy-timeout.
-        It is called whenever a limit buy order has been created,
+        Check buy enter timeout function callback.
+        This method can be used to override the enter-timeout.
+        It is called whenever a limit buy/short order has been created,
         and is not yet fully filled.
         Configuration options in `unfilledtimeout` will be verified before this,
         so ensure to set these timeouts high enough.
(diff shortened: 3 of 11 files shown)


samgermain:
All the changes in this commit will also be in `lev-strat`, this is just to reduce the amount of changes between `lev-strat` and `feat/short`

freqtrade/templates/sample_hyperopt_advanced.py
```diff
@@ -136,56 +136,59 @@ def sell_indicator_space() -> List[Dimension]:
...
+                exit_long_conditions.append(dataframe['mfi'] > params['sell-mfi-value'])
             if 'sell-fastd-enabled' in params and params['sell-fastd-enabled']:
-                conditions.append(dataframe['fastd'] > params['sell-fastd-value'])
+                exit_long_conditions.append(dataframe['fastd'] > params['sell-fastd-value'])
             if 'sell-adx-enabled' in params and params['sell-adx-enabled']:
-                conditions.append(dataframe['adx'] < params['sell-adx-value'])
+                exit_long_conditions.append(dataframe['adx'] < params['sell-adx-value'])
             if 'sell-rsi-enabled' in params and params['sell-rsi-enabled']:
-                conditions.append(dataframe['rsi'] > params['sell-rsi-value'])
+                exit_long_conditions.append(dataframe['rsi'] > params['sell-rsi-value'])
```
xmatthias, line 165: Honestly, i'd not touch this part.

It's the legacy interface - which will eventually be removed - so while we can align method names - i'd not change the content of the methods.

freqtrade/optimize/hyperopt.py
```diff
@@ -109,12 +109,12 @@ def __init__(self, config: Dict[str, Any]) -> None:
...
-                    self.custom_hyperopt.populate_buy_trend)  # type: ignore
-            if hasattr(self.custom_hyperopt, 'populate_sell_trend'):
-                self.backtesting.strategy.advise_sell = (  # type: ignore
-                    self.custom_hyperopt.populate_sell_trend)  # type: ignore
+            if hasattr(self.custom_hyperopt, 'populate_enter_trend'):
+                self.backtesting.strategy.advise_enter = (  # type: ignore
+                    self.custom_hyperopt.populate_enter_trend)  # type: ignore
+            if hasattr(self.custom_hyperopt, 'populate_exit_trend'):
+                self.backtesting.strategy.advise_exit = (  # type: ignore
+                    self.custom_hyperopt.populate_exit_trend)  # type: ignore
```
xmatthias: You'll need to be very careful with this (especially with hyperopt).

Will require compat code.

xmatthias (changes requested):
I don't see this as a breaking change that's necessary - we will need compatibility code that will ensure that existing strategies / hyperopts will still work.

While we can surely show a deprecation warning - we will have to ensure compatibility with the existing version (which i don't see made at the moment).
Everything else would be a terrible upgrade experience - which i refuse to have with any project i'm responsible for.

Obviously, the compatibility code will potentially complex - and will require very good testing (both manual and from unit-tests) 
(this is why i offered do this change - but you said it'll be simple ... :laughing: ).

While i could merge this to feat/short - i don't think that'd be a good idea.

It'll be easier to combine renaming with the compatibility code - as it'll make it obvious where things were removed / changed - which means which parts require special care and hard looks.

---

*edit* I've tried to highlight where i think the compatibility code will be required. 
To properly test that, i think the existing `test/default{strategy,hyperopt}` should remain (probably be renamed) - and copies of these strategies should be added as new "default test" strategies (although i really hate that name as it's wrong and only used for tests - i think that's a change for another day).

freqtrade/resolvers/hyperopt_resolver.py
```diff
@@ -45,12 +45,12 @@ def load_hyperopt(config: Dict) -> IHyperOpt:
...
-                        "Using populate_buy_trend from the strategy.")
-        if not hasattr(hyperopt, 'populate_sell_trend'):
-            logger.info("Hyperopt class does not provide populate_sell_trend() method. "
-                        "Using populate_sell_trend from the strategy.")
+        if not hasattr(hyperopt, 'populate_enter_trend'):
+            logger.info("Hyperopt class does not provide populate_enter_trend() method. "
+                        "Using populate_enter_trend from the strategy.")
+        if not hasattr(hyperopt, 'populate_exit_trend'):
+            logger.info("Hyperopt class does not provide populate_exit_trend() method. "
+                        "Using populate_exit_trend from the strategy.")
```
xmatthias: Compatibility code required

freqtrade/resolvers/strategy_resolver.py
```diff
@@ -200,11 +200,13 @@ def _load_strategy(strategy_name: str,
...
-                                    strategy._buy_fun_len,
-                                    strategy._sell_fun_len]):
+            strategy._buy_fun_len = len(getfullargspec(strategy.populate_enter_trend).args)
+            strategy._sell_fun_len = len(getfullargspec(strategy.populate_exit_trend).args)
+            if any(x == 2 for x in [
+                strategy._populate_fun_len,
+                strategy._buy_fun_len,
+                strategy._sell_fun_len
+            ]):
                 strategy.INTERFACE_VERSION = 1
```
xmatthias: Compatibility code required

freqtrade/resolvers/hyperopt_resolver.py
samgermain: What do you mean by compatibility code?

xmatthias: well - code that ensures a strategy  (or in this case, hyperopt file) that implements `populate_sell_trend` - but not the new `populate_exit_trend` will continue to work as it used to (making it compatible with the old version, too).

samgermain: > well - code that ensures a strategy (or in this case, hyperopt file) that implements `populate_sell_trend` - but not the new `populate_exit_trend` will continue to work as it used to (making it compatible with the old version, too).

What do you think the best way to do that would be?

xmatthias: i'd start with the strategy - maybe look at the code in `advise_buy`/`advise_sell` - which also has some level of compatibility code (we've added a new parameter in the past).

In theory, you can detect if an object has the old but not the new method, and eventually simply reassing (a function is just an object in python) - or call the old function instead if you detect that case (make sure to detect the change ONCE, and not on every call).

Hyperopt might be more complex - as there it seems we're assigning `advise_buy` / `advise_sell` directly (which would need further investigation if that's really necessary or a mistake).

samgermain:
I got rid of most of the name changes, all the changes that remain shouldn't have much of an effect except for `strategy.ignore_expired_candle` but I feel like that one is necessary or it gets confusing what it's asking for

You can change the names however you want

freqtrade/templates/sample_hyperopt_advanced.py
samgermain: Not touch just the sample_hyperopt_advanced file?

samgermain: Just let me know which files are legacy and that I shouldn't touch

xmatthias: I'd assume all files in `tests/strategy/strats` (create copies of either default_strategy or hyperoptable_strategy with the new approach), as well as default_hyperopt (again, create a copy - potentially naming the copy "legacy_hyperopt" though :laughing: )

xmatthias:
> I got rid of most of the name changes, all the changes that remain shouldn't have much of an effect except for `strategy.ignore_expired_candle` but I feel like that one is necessary or it gets confusing what it's asking for

I feel like you misunderstood my points above.

I'm not against these changes - they made sense how you had them.
new strategies can then use the new naming - BUT (and this is important) we need to ensure that old strategies continue to work (compatibility layer).

samgermain:
> > I got rid of most of the name changes, all the changes that remain shouldn't have much of an effect except for `strategy.ignore_expired_candle` but I feel like that one is necessary or it gets confusing what it's asking for
> 
> I feel like you misunderstood my points above.
> 
> I'm not against these changes - they made sense how you had them.
> new strategies can then use the new naming - BUT (and this is important) we need to ensure that old strategies continue to work (compatibility layer).

I decided I don't want to add the compatibility layers though, at least not now, so I won't make the name changes, I have enough else to do, although I guess I'll have to add it for `strategy.ignore_expired_candle`

freqtrade/strategy/interface.py
```diff
@@ -558,12 +567,12 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
                     sell: bool, low: float = None, high: float = None,
                     force_stoploss: float = 0) -> SellCheckTuple:
         """
-        This function evaluates if one of the conditions required to trigger a sell
-        has been reached, which can either be a stop-loss, ROI or sell-signal.
-        :param low: Only used during backtesting to simulate stoploss
-        :param high: Only used during backtesting, to simulate ROI
+        This function evaluates if one of the conditions required to trigger a sell/exit_short
+        has been reached, which can either be a stop-loss, ROI or exit-signal.
+        :param low: Only used during backtesting to simulate (long)stoploss/(short)ROI
```
xmatthias, line 572: This comment is wrong to me. 
Also on short trades it's called stoploss (it's on the negative side of the trade), not ROI, as it's based on profit - not on direction.

freqtrade/strategy/strategy_helper.py
```diff
@@ -58,7 +58,10 @@ def merge_informative_pair(dataframe: pd.DataFrame, informative: pd.DataFrame,
     return dataframe
 
 
-def stoploss_from_open(open_relative_stop: float, current_profit: float) -> float:
+def stoploss_from_open(
+    open_relative_stop: float,
+    current_profit: float
+) -> float:
```
xmatthias: I'm not sure on the point of these formatting changes.

I found it easier to read in one line (especially considering that this line is not that long).

freqtrade/templates/sample_hyperopt.py
```diff
@@ -46,7 +46,7 @@ class SampleHyperOpt(IHyperOpt):
     """
 
     @staticmethod
-    def indicator_space() -> List[Dimension]:
+    def buy_indicator_space() -> List[Dimension]:
```
xmatthias: While i think this would be a good change - i think it can't be changed if you'll not want to write compatibility code - as this is now a function that's not known to freqtrade.

xmatthias:
> I decided I don't want to add the compatibility layers though, at least not now, so I won't make the name changes 

I'm quite sad to point out that i said so from the beginning - leave the method names as they are for now - they're naming not relevant for the moment and pretty complex to change (as we will want a good user experience - which means no unnecessary breaking changes).


what's that leave us with in this PR? is it only comment / docstring and formatting changes ... ?

freqtrade/strategy/interface.py
samgermain: What part about it is wrong? On shorts, the roi would be a low value, and the stoploss would be a high value

xmatthias (approved):

freqtrade/strategy/strategy_helper.py
samgermain: I remember now, I had to add a parameter for shorting which made it over 100 characters long

PR #5461, opened 2021-08-20
Added buy_tag to sell signal outputs

ZerGo0:
## Summary
Added buy_tag to all sell signals .

## Quick changelog

- Added buy_tag to all sell signals

## What's new?
This should make it easier for users to filter buy tags.

Example


freqtrade/strategy/interface.py
```diff
@@ -655,12 +655,20 @@ def should_sell(self, trade: Trade, rate: float, date: datetime, buy: bool,
         # Stoploss
         if roi_reached and stoplossflag.sell_type != SellType.STOP_LOSS:
             logger.debug(f"{trade.pair} - Required profit reached. sell_type=SellType.ROI")
+            if hasattr(trade, 'buy_tag') and trade.buy_tag is not None:
```
xmatthias: You'll not have to test with `hasattr` here. 
Trade is an internal class - where old versions of freqtrade won't have it - but new versions do.
As such - it's safe to assume it's always there as attribute (although it can obviously be empty).

xmatthias (changes requested):
i don't really like the approach you've selected for this.

First of all, it's a lot of duplicate code (which is never good, as it'll most likely drift apart over time).
The bigger problem though is - you're now storing this as `ROI (whatever)` in the database.

`whatever` is however already available as buy-tag in the same database table - so this basically stores this information twice.

I think the better approach will be to combine the two in the output (where it's shown and can be selected dynamically) ... that way, the information is not stored twice - and will work in a backwards compatible way (trades opened with the current version of freqtrade will also show things in the new way).

with the current implementation, this change would only apply to trades that close with this version - which is not necesary and will create quite confusing outputs - with both `ROI` and `ROI (whatever)` listed.
(thinking of the `/stats` output).

ZerGo0:
> i don't really like the approach you've selected for this.
> 
> First of all, it's a lot of duplicate code (which is never good, as it'll most likely drift apart over time).
> The bigger problem though is - you're now storing this as `ROI (whatever)` in the database.
> 
> `whatever` is however already available as buy-tag in the same database table - so this basically stores this information twice.
> 
> I think the better approach will be to combine the two in the output (where it's shown and can be selected dynamically) ... that way, the information is not stored twice - and will work in a backwards compatible way (trades opened with the current version of freqtrade will also show things in the new way).
> 
> with the current implementation, this change would only apply to trades that close with this version - which is not necesary and will create quite confusing outputs - with both `ROI` and `ROI (whatever)` listed.
> (thinking of the `/stats` output).

I agree and changed it to only use the buy tags whenever it's actually outputted (hopefully).

I haven't tested the Stoploss Guard Protection, it uses trade.sell_reason, but that should not be effected by this changes.

freqtrade/persistence/models.py
```diff
@@ -324,7 +324,7 @@ def to_json(self) -> Dict[str, Any]:
             'profit_pct': round(self.close_profit * 100, 2) if self.close_profit else None,
             'profit_abs': self.close_profit_abs,
 
-            'sell_reason': self.sell_reason,
+            'sell_reason': str(self.sell_reason) + (f' ({self.buy_tag})' if self.buy_tag else ''),
```
xmatthias, line 327: Definitely don't do this. 
This decision should be left to the consuming party (it get's both informations independently) - not forced from the bot.

freqtrade/freqtradebot.py
```diff
@@ -1181,7 +1181,7 @@ def _notify_sell(self, trade: Trade, order_type: str, fill: bool = False) -> Non
             'current_rate': current_rate,
             'profit_amount': profit_trade,
             'profit_ratio': profit_ratio,
-            'sell_reason': trade.sell_reason,
+            'sell_reason': trade.sell_reason  + (f' ({trade.buy_tag})' if trade.buy_tag else ''),
```
xmatthias, line 1184: i'd expose "buy_tag" in this call seperately (below too, obviously).

By combining the data here, you're forcing consumers to use it this way (this might be especially problematic for webhooks, where users can configure the provided data themselfs).

in my opinion, this "combination" should be happening in the display layer (which would be the rpc/* files) - where necessary obviously (most likely only in telegram).

xmatthias (changes requested):
I'm not convinced if showing the data this way is really beneficial - as it'll mess the "sell reason stats" up pretty good - and will make it potentially difficult to identify bad sell-signals (as the same sell-signal can be split up into multiple lines).


For that reason, i'm thinking if this way to display things should be optional (potentially introduced at a later point).
That optionality will be easier if the logic is moved to the "formatting" layer - which would be `telegram.py`  (and potentially the rpc.py part you've already got there.


you'll also want to include this in `optimize_reports.py` - otherwise you'll not get it during backtesting (which i think is your main goal here).

rokups:
This could very well be an external tool, analyzing backtest results say in a notebook. Maybe this could be in form of a helper function meant to be consumed by users, bot not used by the bot directly?

xmatthias:
> This could very well be an external tool, analyzing backtest results say in a notebook. Maybe this could be in form of a helper function meant to be consumed by users, bot not used by the bot directly?

I'm not sure about that - providing it as part of backtesting can be beneficial in certain circumstances - but i think it should be opt-in (e.g. `--include-buy-reason-stats`), as it'll not help in all cases - and can complicate the results (if i'm only looking for sell optimization, this will make results almost unreadable without manual calculations).

theluxaz:
This would be really useful for someone like me who needs to optimize only the buy signals and see which ones are the most profitable. 

At the moment I cannot find a way to display the set buy_tags in the sell reason section as "custom_sell" function only activates when sell signals are not present, while I need to use the regular sell signals from "populate_sell_trend". 

Is there any way to achieve this currently?

xmatthias:
> Is there any way to achieve this currently?

no, although it should be pretty easy to build something if you start off of the [analysis notebook](https://www.freqtrade.io/en/stable/strategy_analysis_example) (should also be available in your user_data directory) - which has examples for loading trades, as well as grouping by sell reason (simply swap sell_reason with buy_tag ...).

ZerGo0:
Going to close this for now, maybe it be better if it was configurable in the config.

PR #5476, opened 2021-08-23
Update setup.sh

masterzion:

## Summary
no interactive installation

## Quick changelog

- added --auto-install  parameter to the installer

## What's new?
added ./setup.sh --auto-install, that allows to use the install in other scripts

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -30,7 +30,7 @@ function check_installed_python() {
             check_installed_pip
             return
         fi
-    done 
+    done
 
     echo "No usable python found. Please make sure to have python3.7 or newer installed"
     exit 1
@@ -48,16 +48,27 @@ function updateenv() {
     SYS_ARCH=$(uname -m)
     echo "pip install in-progress. Please wait..."
     ${PYTHON} -m pip install --upgrade pip
-    read -p "Do you want to install dependencies for dev [y/N]? "
+    if [[ "$1" == "auto" ]]
+    then
+      REPLY="n"
+    else
+      read -p "Do you want to install dependencies for dev [y/N]? "
+    fi
     if [[ $REPLY =~ ^[Yy]$ ]]
     then
         REQUIREMENTS=requirements-dev.txt
     else
         REQUIREMENTS=requirements.txt
     fi
+
     REQUIREMENTS_HYPEROPT=""
     REQUIREMENTS_PLOT=""
-     read -p "Do you want to install plotting dependencies (plotly) [y/N]? "
+    if [[ "$1" == "auto" ]]
+    then
+      REPLY="y"
+    else
+      read -p "Do you want to install plotting dependencies (plotly) [y/N]? "
+    fi
     if [[ $REPLY =~ ^[Yy]$ ]]
     then
         REQUIREMENTS_PLOT="-r requirements-plot.txt"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -249,11 +273,12 @@ function plot() {
 
 function help() {
     echo "usage:"
-    echo "	-i,--install    Install freqtrade from scratch"
-    echo "	-u,--update     Command git pull to update."
-    echo "	-r,--reset      Hard reset your develop/stable branch."
-    echo "	-c,--config     Easy config generator (Will override your existing file)."
-    echo "	-p,--plot       Install dependencies for Plotting scripts."
+    echo "	-a,--auto-install    Install freqtrade from scratch"
```
samgermain: The comment probably shouldn't be the exact same as -i

setup.sh
```diff
@@ -67,13 +78,19 @@ function updateenv() {
         ${PYTHON} -m pip install --upgrade cython
     else
         # Is not Raspberry
-        read -p "Do you want to install hyperopt dependencies [y/N]? "
+        if [[ "$1" == "auto" ]]
+        then
+          REPLY="y"
+        else
+          read -p "Do you want to install hyperopt dependencies [y/N]? "
+        fi
```
samgermain, line 86: Rather than having all these conditionals, it might be better to put this code into a function

```
${PYTHON} -m pip install --upgrade -r ${REQUIREMENTS} ${REQUIREMENTS_HYPEROPT} ${REQUIREMENTS_PLOT}
    if [ $? -ne 0 ]; then
        echo "Failed installing dependencies"
        exit 1
    fi
    ${PYTHON} -m pip install -e .
    if [ $? -ne 0 ]; then
        echo "Failed installing Freqtrade"
        exit 1
    fi
    echo "pip install completed"
    echo
```

and then if someone selects `auto-install`, just set

```
        REQUIREMENTS=requirements.txt
        REQUIREMENTS_PLOT="-r requirements-plot.txt"
        REQUIREMENTS_HYPEROPT="-r requirements-hyperopt.txt"
```

and then call that function

------------

There would have to be a few other changes as well, to make sure the `env` is set up right

samgermain (changes requested):

xmatthias:
i'm not fully convinced what this tries to accomplish.
"full" automatic mode would have additional parameters to select the different parameters (like hyperopt or dev dependencies).

On the other hand, it's however pretty much overcomplicated, as installation (assuming a working python installation) is as much as 4 lines (including ta-lib) once you remove interactivity.
as such - i'm not 100% convinced that the complication to the script justifies what it accomplishes

``` bash
cd build_helpers && ./install_ta-lib.sh && cd ..
# Next line you can pick which requirements you need.
pip install -r requirements-dev.txt
pip install -e .
```

masterzion:
> ```shell
> cd build_helpers && ./install_ta-lib.sh && cd ..
> # Next line you can pick which requirements you need.
> pip install -r requirements-dev.txt
> pip install -e .
> ```

sorry, i didnt know about it... at least i dont remember reading this on the documentation.

xmatthias:
> sorry, i didnt know about it... at least i dont remember reading this on the documentation.

It's not part of the documentation - partially intentional as having setup automated is pretty much a very advanced task I've not yet seen requested very often, and using the setup-script works for most people (as it's asking some sensible questions).

For automation, you'll often also want the ability to pick the virtualenvironment (or none at all) - so I'm not sure that an "automated" way via setup-script (which does take some decisions for you) is the best way.

PR #5501, opened 2021-08-29
Export HDF5 and CBLOSC paths.

pan-long:
This is needed if homebrew isn't installed in the standard path, say,
/usr/local/.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -119,6 +119,7 @@ function install_mac_newer_python_dependencies() {
         echo "-------------------------"
         brew install hdf5
     fi
+    export HDF5_DIR=$(brew --prefix)
 
     if [ ! $(brew --prefix --installed c-blosc 2>/dev/null) ]
     then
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -127,6 +128,7 @@ function install_mac_newer_python_dependencies() {
         echo "-------------------------"
         brew install c-blosc
     fi    
+    export CBLOSC_DIR=$(brew --prefix)
```
xmatthias, line 131: I kinda have to be blunt and ask @samgermain to have a look as i'm no MacOS user - which means i kinda struggle determining if this is right or wrong (i trust it's right, but another opinion won't hurt).

samgermain: @xmatthias I think it's probably fine, I uninstalled `hdf5` and `c-blosc` and removed `.env` and the script worked fine, I'm not really sure how to test it other than that though

xmatthias: That's more than i am able to do without a mac :+1:

xmatthias (approved):
After confirmation from another MacOS user, LGTM :+1:

PR #5504, opened 2021-08-29
Surround "unlimited" by double quotes in build config.

pan-long:


freqtrade/commands/build_config_commands.py
```diff
@@ -66,16 +66,22 @@ def ask_user_config() -> Dict[str, Any]:
         {
             "type": "text",
             "name": "stake_amount",
-            "message": "Please insert your stake amount:",
+            "message": f"Please insert your stake amount (Integer or '{UNLIMITED_STAKE_AMOUNT}'):",
```
xmatthias:
```suggestion
            "message": f"Please insert your stake amount (Number or '{UNLIMITED_STAKE_AMOUNT}'):",
```

That's the only thing i'd change ... as it doesn't have to be an integer (no decimals) - but can also be 0.01 (or whatever else).

xmatthias (approved):
Great catch! :+1:

PR #5522, opened 2021-09-02
Added method throw_exception_if_method_not_on_exchange to exchange

samgermain:
## Summary
Adds a method that checks if a method is implemented on the ccxt exchange class

## Quick changelog
Adds this method to exchange class
- throw_exception_if_method_not_on_exchange(self, method_name: str)

Currently the test for this fail, I think this is because of the way that the exchange object is instantiated in the test

```
exchange = get_patched_exchange(mocker, default_conf, id=exchange_name)
```

I don't know how I would write an actual exchange object implementation in order to pass the tests.

-------------------

There is some overlap between this and `funding-fee`. I realized I needed this method in multiple PRs

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -228,6 +228,16 @@ class Exchange:
         """exchange ccxt precisionMode"""
         return self._api.precisionMode
 
+    def throw_exception_if_method_not_on_exchange(self, method_name: str):
+        """
+            Throws exception if method is implemented by the current exchange on ccxt
+            :param method_name: The method that may/may not exist for this exchange class
+        """
+        # api_method = hasattr(self._api, method_name, None)
+        if not self._api.describe()['has'][method_name]:
+            raise OperationalException(
+                f"{method_name}() has not been implemented on ccxt.{self.name}")
+
     def _log_exchange_response(self, endpoint, response) -> None:
         """ Log exchange responses """
         if self.log_responses:
@@ -361,7 +371,7 @@ class Exchange:
             raise OperationalException(
                 'Could not load markets, therefore cannot start. '
                 'Please investigate the above error for more details.'
-                )
+            )
         quote_currencies = self.get_quote_currencies()
         if stake_currency not in quote_currencies:
             raise OperationalException(
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -2926,3 +2926,36 @@ def test_calculate_fee_rate(mocker, default_conf, order, expected) -> None:
 ])
 def test_calculate_backoff(retrycount, max_retries, expected):
     assert calculate_backoff(retrycount, max_retries) == expected
+
+
+@pytest.mark.parametrize("exchange_name,method_name,is_callable", [
+    ("binance", "set_margin_mode", True),
+    ("kraken", "set_margin_mode", False),
+    ("ftx", "set_margin_mode", False),
+    ("bittrex", "set_margin_mode", False),
+    ("binance", "not_a_real_method", False),
+    ("kraken", "not_a_real_method", False),
+    ("ftx", "not_a_real_method", False),
+    ("bittrex", "not_a_real_method", False),
+    ("binance", "create_order", True),
+    ("kraken", "create_order", True),
+    ("ftx", "create_order", True),
+    ("bittrex", "create_order", True),
+])
+def test_throw_exception_if_method_not_on_exchange(
+    mocker,
+    default_conf,
+    caplog,
+    exchange_name,
+    method_name,
+    is_callable
+):
+    exchange = get_patched_exchange(mocker, default_conf, id=exchange_name)
+    warning = f"{method_name}() has not been implemented on ccxt.{exchange_name.capitalize()}"
+    if not is_callable:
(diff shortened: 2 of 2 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -228,6 +228,16 @@ def precisionMode(self) -> str:
...
+    def throw_exception_if_method_not_on_exchange(self, method_name: str):
+        """
+            Throws exception if method is implemented by the current exchange on ccxt
+            :param method_name: The method that may/may not exist for this exchange class
+        """
+        # api_method = hasattr(self._api, method_name, None)
+        if not self._api.describe()['has'][method_name]:
+            raise OperationalException(
+                f"{method_name}() has not been implemented on ccxt.{self.name}")
+
```
xmatthias, line 240: This already exists as `exchange_has()`, which however does not raise, but returns False.

There's 2 applications for that - either check at startup to verify if a configuration is valid (check if certain special options exist - like `createMarketOrder`) - or to check during runtime and fall back to alternative means. 
Neither of the 2 benefit from having that method "raise" a "failing" exception (on top of the fact that you're risking keyerrors with the approach you're taking).

xmatthias (changes requested):
In my opinion we have this already as `exchange_has`.

unless you can explain what the benefit of this over `exchange_has` are i think we can close this.

samgermain:
> In my opinion we have this already as `exchange_has`.
> 
> unless you can explain what the benefit of this over `exchange_has` are i think we can close this.

`exchange_has` isn't working when I use it, i broke it down to

```
binance = ccxt.binance({
    'api_key': "...",
    'secret': "...",
    'options': {
        'defaultType': 'margin'
    },
})
ftx = ccxt.ftx({
    'api_key': "...",
    'secret': "..."
})
endpoint="fetch_funding_history"

pprint(f"endpoint in binance.has: {endpoint in binance.has}")
pprint(f"binance.has[endpoint]: {binance.has[endpoint]}")
pprint(f"endpoint in ftx.has: {endpoint in ftx.has}")
pprint(f"ftx.has[endpoint]: {ftx.has[endpoint]}")
```
and this is happening
```
'endpoint in binance.has: False'
  File "/Users/sam/Documents/dev/trading/freqtrade/../ccxt-scratch/main.py", line 203, in <module>
    pprint(exchange_has("fetch_funding_history"))
  File "/Users/sam/Documents/dev/trading/freqtrade/../ccxt-scratch/main.py", line 194, in exchange_has
    pprint(f"binance.has[endpoint]: {binance.has[endpoint]}")
KeyError: 'fetch_funding_history'
```

When I run `dir(ccxt.binance())` or `dir(ccxt.ftx())`, `fetch_funding_history` is listed in the results of both

xmatthias:
fetch_funding_history is wrong - the correct naming is `fetchFundingHistory` (it's a dict that points out the available methods (manually maintained, allowing work on a function before releasing it i assume) - NOT a list of methods using dir.

Having a different method to control for that won't change the fact if you're using the wrong name.

samgermain:
> fetch_funding_history is wrong - the correct naming is `fetchFundingHistory` (it's a dict that points out the available methods (manually maintained, allowing work on a function before releasing it i assume) - NOT a list of methods using dir.
> 
> Having a different method to control for that won't change the fact if you're using the wrong name.

I edited my exchange_has snippet above to use the endpoint `fetchFundingHistory`, and it works using that snippet but I'm still getting an error when running pytest on freqtrade, and when using `pdb`, I still get 

```
(Pdb) self.exchange_has('fetchFundingHistory')
False
(Pdb) 'fetchFundingHistory' in self._api.has
False
```

I think it has to do with `MagicMock`

```
(Pdb) self._api
<MagicMock id='4862689136'>
(Pdb) self
<freqtrade.exchange.binance.Binance object at 0x121e8e9a0>
```

What would be the best way to do this, because Ideally I would like to check that condition as well

xmatthias:
Just search through tests - there's a few that patch "has" properly (using a PropertyMock, iirc).

samgermain:
> Just search through tests - there's a few that patch "has" properly (using a PropertyMock, iirc).

Ok, I've almost got it, I'm getting the wrong number of call_counts in `ccxt_exceptionhandlers` (1 when it's asserted as 5)

```
 def ccxt_exceptionhandlers(mocker, default_conf, api_mock, exchange_name,
                               fun, mock_ccxt_fun, retries=API_RETRY_COUNT + 1, **kwargs):
    
        with patch('freqtrade.exchange.common.time.sleep'):
            with pytest.raises(DDosProtection):
                api_mock.__dict__[mock_ccxt_fun] = MagicMock(side_effect=ccxt.DDoSProtection("DDos"))
                exchange = get_patched_exchange(mocker, default_conf, api_mock, id=exchange_name)
                getattr(exchange, fun)(**kwargs)
>           assert api_mock.__dict__[mock_ccxt_fun].call_count == retries
E           AssertionError: assert 1 == 5
E            +  where 1 = <MagicMock id='4864449936'>.call_count

tests/exchange/test_exchange.py:37: AssertionError
```

xmatthias:
tough to say without the code - let's please move this to the "failing test" lines - so i can look at it for real :laughing:

PR #5523, opened 2021-09-02
⚡ Updated `config.guess` & `config.sub` in `ta-lib-0.4.0-src.tar.gz`️

Rikj000:
## Summary
This PR updates the [`config.guess`](https://cvs.savannah.gnu.org/viewvc/*checkout*/config/config/config.guess) & [`config.sub`](https://cvs.savannah.gnu.org/viewvc/*checkout*/config/config/config.sub) files inside the [`ta-lib-0.4.0-src.tar.gz`️](https://github.com/freqtrade/freqtrade/blob/develop/build_helpers/ta-lib-0.4.0-src.tar.gz) file to the latest available.

## Quick ChangeLog

- [`config.guess`](https://cvs.savannah.gnu.org/viewvc/*checkout*/config/config/config.guess) 11/08/2006 => 01/01/2021
- [`config.sub`](https://cvs.savannah.gnu.org/viewvc/*checkout*/config/config/config.sub) 11/08/2006 => 01/01/2021

## What's new?
Support for more Linux distributions! :tada: 

I slapped this fix together a few months ago when I was having issues on the DietPi distro,
there I was getting errors that it wasn't recognizing my OS and then failing to build & install the TA-Lib C dependency package.
Updating these 2 files allowed me to build & install successfully, 
but took a few months for me to come around to actually PRing it in.. :sweat_smile: 

TA-Lib-config-guess-sub-updated


xmatthias:
Is my assumption correct that you manually unzipped / rezipped this file to create this?

the zip-file is (intentionally) a 1:1 copy of the file you can download [here](http://prdownloads.sourceforge.net/ta-lib/ta-lib-0.4.0-src.tar.gz) - check-sums are matching (just re-verified).

Having the check-sums on binary files (a tar.gz is also just a binary blob) differ beween source and target creates a trust-problem - (what else was modified?) - so i'd not want to use this modified file to install a system that get's access to my exchange keys.
It's also impossible for me to verify what exactly was modified inside (i don't plan on doing a diff between each file).

Best submit your fix to the original source and let us know so we can udpate the file with the original file - but we'll not use a tar.gz file that differs from files you can optain elsewhere.

Rikj000:
> It's also impossible for me to verify what exactly was modified inside (i don't plan on doing a diff between each file).

Checking for the `diffs` can easily be done with VSC within minutes with the use of the [Compare Folders](https://open-vsx.org/extension/moshfeu/compare-folders) plugin:
Folder-diffs

> Best submit your fix to the original source and let us know so we can update the file with the original file - but we'll not use a tar.gz file that differs from files you can obtain elsewhere.

I think it's safe to assume that the original source project is dead... Last commit: [`03-04-2013`](https://sourceforge.net/p/ta-lib/code/commit_browser).
Build date of the last release binary that you linked goes all the way back to [`15-09-2007`](https://sourceforge.net/projects/ta-lib/files/ta-lib/0.4.0/)
In the commit history I even spotted a `v0.5.0` release, which never actually came out..

Because of those reasons (+ me not being familiar in C, nor with pushing code to SourceForge) I indeed simply re-packed the binary with the 2 files updated.
If you believe that the C TA-Lib project is not dead & are familiar with C & SourceForge, then please PR those changes to them. But I don't plan on doing so.

If you really don't trust my PR then feel free to close it..

xmatthias:
It's not about trusting you or not - but about user trust in freqtrade.
Having a modified binary blob does not create or enforce this trust - as users comming to this repository in a year from today will have difficulties in comparing what was changed - and "finding the PR" is not easy for users unfamiliar with github.


I'm also not convinced what this gives (assuming the install instructions are followed), considering that we're installing / updating the exact same dependencies already (maybe a slightly different version) without messing with the binary blob itself.

While the outcome is the same - the result is a non-modified blob.
It makes it easy to verify which changes are done to the blob (there's another source-code change in the same file as well that's needed for BTC pairs support due to their often low values).

https://github.com/freqtrade/freqtrade/blob/93c1dff71bbae773b922794f956013e0d0b9c908/build_helpers/install_ta-lib.sh#L11-L12

PR #5529, opened 2021-09-03
⚡ `setup.sh` - Use `build_helpers/install_ta-lib.sh` for TA-Lib insta…

Rikj000:
See PR https://github.com/freqtrade/freqtrade/pull/5523

This PRs goal is the same but without muddying the checksum of the TA-Lib binary.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -95,19 +95,7 @@ function install_talib() {
         return
     fi
 
-    cd build_helpers
-    tar zxvf ta-lib-0.4.0-src.tar.gz
-    cd ta-lib
-    sed -i.bak "s|0.00000001|0.000000000000000001 |g" src/ta_func/ta_utility.h
-    ./configure --prefix=/usr/local
-    make
-    sudo make install
-    if [ -x "$(command -v apt-get)" ]; then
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -95,19 +95,7 @@ function install_talib() {
...
-    tar zxvf ta-lib-0.4.0-src.tar.gz
-    cd ta-lib
-    sed -i.bak "s|0.00000001|0.000000000000000001 |g" src/ta_func/ta_utility.h
-    ./configure --prefix=/usr/local
-    make
-    sudo make install
-    if [ -x "$(command -v apt-get)" ]; then
-        echo "Updating library path using ldconfig"
-        sudo ldconfig
-    fi
```
xmatthias, line 108: You will need to keep these 3 commands - ldconfig is not part of the install script (intentionally) - but can be required to make ta-lib work correctly.

xmatthias: Don't worry about the test failure - it's a random fluke - and will go away with the next push i assume.

xmatthias (approved):
LGTM :+1:

PR #5532, opened 2021-09-05
case insensitive blacklist

lazydroid:
## Summary

Allow "btc/usdt" pairs in blacklist to match to "BTC/USDT" pairs that come from the exchange.

## Quick changelog

- Added re.IGNORECASE flag to the regular expression matcher that is used for blacklists.

## What's new?

Added re.IGNORECASE flag to the regular expression matcher that is used for blacklists.


xmatthias (approved):
LGTM :+1:

PR #5533, opened 2021-09-05
SplineHyperOptLoss, for easier hyperopt

squat0001:
## Summary

HyperOpt loss functions are not easy, with this it will get a lot simpler to a nice hyperopt

## What's new?

- multiple curves combined
- easy adujstment of values in curves
- easy debugable


freqtrade/optimize/hyperopt_loss_splinehyperoptloss.py
```diff
@@ -0,0 +1,129 @@
...
+            str(5): 5,
+            str(60): 500,
+            str(60 * 5): 250,
+            str(60 * 10): 150,
+            str(60 * 20): 50,
+            str(60 * 2000): -10,
+        }
+        r = execute_table(trade_duration_table, trade_duration)
+        logger.debug(f'trade_duration: {trade_duration}')
+        logger.debug(f'trade_duration ret: {r}')
```
xmatthias: You will want to combine these into one line (one logentry) (obviously the other 2 as well).

something around 
```suggestion
        logger.debug(f'trade_duration: {trade_duration}, ret: {r}')
```
but obviously for the other 2 occurances, too.

freqtrade/optimize/hyperopt_loss_splinehyperoptloss.py
```diff
@@ -0,0 +1,129 @@
...
+from pandas import DataFrame
+from scipy import interpolate
+
+logger = logging.getLogger(__name__)
+
+import numpy as np
+
+# HyperOpt Loss function, it is difficult to put the various optimizations into one number.
+# To make it easier, you can specify one table per optimization here. Missing values are interpolated.
+# In this way it is possible to create your own loss curves for each value quite easily.
```
xmatthias: You're explaining it as "easy" - but i don't think it really is, at least not without further explanation.

The comments here suggest that we'd expect people to eventually tune values ... but without explanation what can be tuned and what the impact will be of changing one or the other side of the dict - it'll be very difficult for users not 100% familiar with python / freqtrade to archive anything.

Tunable things (i assume the Dicts?) should also be moved to the top (as Constant) so it's easily accessible in one place.

freqtrade/optimize/hyperopt_loss_splinehyperoptloss.py
```diff
@@ -0,0 +1,129 @@
...
+
+        def remove_outliners(x) -> np.array:
+            x = x[x.between(x.quantile(.15), x.quantile(.85))]
+            return x
+
+        def execute_table(table: dict, x: float) -> float:
+            x_points = [float(k) for k in table.keys()]
+            y_points = list(table.values())
+            tck = interpolate.splrep(x_points, y_points, k=2)
+            return float(interpolate.splev(x, tck))
```
xmatthias, line 40: would it make sense to have a specific test for this method? it's logic seems complex (actually not complex, but easy to break - which usually justifies extra tests).
I can help with tests if that's something you're not familiar with.

to make it testable - it should be moved to class level - maybe with an underscore showing it's a private method).

xmatthias (changes requested):
We'll for sure require documentation for this (at least a note of it's existence on the Hyperopt page).

it should also be added to `HYPEROPT_LOSS_BUILTIN` in constants (it can be used without that, but won't show up in help)

I also find comments slightly contradicting (it mentions easy / modifications - but it's not easy, as the modifications would have to be all over the place and are not explained).

squat0001:
Hi!

Am 06.09.2021 um 07:06 schrieb Matthias:
> @xmatthias requested changes on this pull request.
>
> We'll for sure require documentation for this (at least a note of it's existence on the Hyperopt page).
>
> it should also be added to `HYPEROPT_LOSS_BUILTIN` in constants (it can be used without that, but won't show up in help)
>
> I also find comments slightly contradicting (it mentions easy modifications - but it's not as the modifications would have to be all over the place and are not explained).

i'm not a native speaker, and writing english is "hard" for me.

>> +        logger.debug(f'trade_duration: {trade_duration}')
> +        logger.debug(f'trade_duration ret: {r}')
>
> You will want to combine these into one line (one logentry) (obviously the other 2 as well).
>
> something around
> ```suggestion
>          logger.debug(f'trade_duration: {trade_duration}, ret: {r}')
> ```
> but obviously for the other 2 occurances, too.

OK

>> +# To make it easier, you can specify one table per optimization here. Missing values are interpolated.
> +# In this way it is possible to create your own loss curves for each value quite easily.
>
> You're explaining it as "easy" - but i don't think it really is, at least not without further explanation.
>
> The comments here suggest that we'd expect people to eventually tune values ... but without explanation what can be tuned and what the impact will be of changing one or the other side of the dict - it'll be very difficult for users not 100% familiar with python / freqtrade to archive anything.
>
> Tunable things (i assume the Dicts?) should also be moved to the top (as Constant) so it's easily accessible in one place.

i thougt about it, maybe it would be better to provide this class as a 
sample, like the strategy samples, so people can use and modify it. to 
be useful someone needs to adjust the dicts.

>> +        def execute_table(table: dict, x: float) -> float:
> +            x_points = [float(k) for k in table.keys()]
> +            y_points = list(table.values())
> +            tck = interpolate.splrep(x_points, y_points, k=2)
> +            return float(interpolate.splev(x, tck))
>
> would it make sense to have a specific test for this method? it's logic seems complex (actually not complex, but easy to break - which usually justifies extra tests).
> I can help with tests if that's something you're not familiar with.
>
> to make it testable - it should be moved to class level - maybe with an underscore showing it's a private method).

in my first version this was a class method, but the loss class in my 
knowledge is only used with this static method, no instance accessable. 
to avoid hardcode the classname into the static function
i decided to use an inner function.

to add a test, i would test the whole class at once, provide some trade 
result batches, and let it calculate the loss.

lg

freqtrade/optimize/hyperopt_loss_splinehyperoptloss.py
```diff
@@ -0,0 +1,126 @@
...
+            '0': 1000,
+            '0.1': 3000,
+            '0.4': 500,
+            '0.8': -5000,
+        }
+        r = execute_table(max_drawdown_table, max_drawdown)
+        logger.debug(f'max_drawdown: {max_drawdown} ret: {r}')
+        ret += r
+
+        return -ret
```
xmatthias, line 126: I'm ... not sure about this - running a simple testset against this loss-function (`test_loss_functions_better_profits`)  seems to move in the wrong direction. It'll favor the one with half the profits over the one with twice as many profits ...

xmatthias:
closing this due to the long inactivity. personally, i don't plan on finishing this - and it looks like you've lost interest in this as well.

PR #5541, opened 2021-09-07
[Feat/Short] Added Ftx Interest Calculation

arunavo4:

## Summary
Added interest rate calculation for borrowed money

Trying to Add Smaller chunks of code as suggested by @samgermain 

## Quick changelog

- Added Formula for interest rate Calculation

## What's new?

```
Equation to calculate interest on margin trades

        :param exchange_name: The exchanged being trading on
        :param borrowed: The amount of currency being borrowed
        :param rate: The rate of interest (i.e daily interest rate)
        :param hours: The time in hours that the currency has been borrowed for

        Raises:
            OperationalException: Raised if freqtrade does
            not support margin trading for this exchange

        Returns: The amount of interest owed (currency matches borrowed)
```


freqtrade/leverage/interest.py
```diff
@@ -36,7 +36,8 @@ def interest(
         # Rounded based on https://kraken-fees-calculator.github.io/
         return borrowed * rate * (one+ceil(hours/four))
     elif exchange_name == "ftx":
-        # TODO-lev: Add FTX interest formula
-        raise OperationalException(f"Leverage not available on {exchange_name} with freqtrade")
+        # As Explained under #Interest rates section in
+        # https://help.ftx.com/hc/en-us/articles/360053007671-Spot-Margin-Trading-Explainer
+        return borrowed * rate * ceil(hours)/twenty_four
```
samgermain, line 41: From the other PR

**Sam:**
so with this formula(and kraken, I'll have to change it) I believe that there may be separate interest rates for the interest opening fee, and the interest fee charged for holding the asset. I'm not certain, but I think that the formula might actually look something more like

         (borrowed * opening_rate) + (borrowed * rate * ceil(hours)/twenty_four)
         
Can you double check this for FTX? Don't worry about Kraken

**arunova4:**
```
Interest rates
Every hour, lenders are paid and borrowers are charged.  This is determined as follows:

1. All lenders specify a minimum borrow rate they must receive for that hour.
2. At the beginning of the hour, we calculate the total borrow demand for each coin between all users.
3. We have an auction: we sort the lending offers by minimum rate, and take the cheapest set that satisfies the borrow demand.  The borrow rate is set to the minimum borrow rate of the marginal (most expensive) loan that was required.

For example, say that:
Alice: wants to borrow 2 BTC
Bob: wants to borrow 3 BTC
Charlie: lending 1 BTC, min 0.01%
Denise: lending 10 BTC, min 0.03%
Then:
Total borrow demand is 5 BTC.
This borrows Charlie’s BTC and 4 of Denise's
The marginal loan is Denise's 4 BTC loaned out at 0.03% minimum
So all 5 BTC of borrows -- including the one against Charlie -- use an interest rate of 0.03%
FTX charges borrowers an additional fee; for borrowers, it's already baked into the rates they see. See below for an explanation on how these are calculated.
The interest rates are generally quoted in % per day (e.g. 0.05%/day); 1/24th of that is paid out on the hour.
You can monitor the borrow rates you're paying here.

So this means the following:

1. All borrow rates only last for one hour; each hour, they’re re-determined.
2. Every borrower pays the same rate for an hour/coin; and every lender who does in fact lend their tokens receives the same interest rate for an hour/coin.
3. When you open up a borrow mid-hour, you end up paying for that hour’s interest rate at the end during the auction.

> Note that you will be charged interest on your borrows whether they're held on FTX, withdrawn, pending withdrawal, sold, or anything else.
```
As mentioned above 
There is no such mention of opening rates as such on ftx but the rates are different for every hour so if we calculate like this we do simplify the calculation. Maybe the `rate` here can be an average of the rates for that many hours.

but if you want an exact one u will have to supply the rates for each hour to the function.

arunavo4:
@xmatthias  @samgermain Added the Tests. Ready to merge

samgermain:
@xmatthias This looks good to me, if you're good with it we can merge it

xmatthias (approved):
LGTM :+1: 

Thanks for helping out on this (pretty complicated, and probably long-taking) feature of adding short/leverage trading :+1:

PR #5546, opened 2021-09-08
freqtradebot-tests-caplog-clears

samgermain:
added caplog clears to freqtradebot tests

### DIFF
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -107,6 +107,7 @@ def test_order_dict_dry_run(default_conf, mocker, caplog) -> None:
     freqtrade = FreqtradeBot(conf)
     assert not freqtrade.strategy.order_types['stoploss_on_exchange']
     assert not log_has_re(".*stoploss_on_exchange .* dry-run", caplog)
+    caplog.clear()
 
 
 def test_order_dict_live(default_conf, mocker, caplog) -> None:
@@ -140,6 +141,7 @@ def test_order_dict_live(default_conf, mocker, caplog) -> None:
     freqtrade = FreqtradeBot(conf)
     assert not freqtrade.strategy.order_types['stoploss_on_exchange']
     assert not log_has_re(".*stoploss_on_exchange .* dry-run", caplog)
+    caplog.clear()
 
 
 def test_get_trade_stake_amount(default_conf, ticker, mocker) -> None:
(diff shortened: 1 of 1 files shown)


tests/test_freqtradebot.py
```diff
@@ -107,6 +107,7 @@ def test_order_dict_dry_run(default_conf, mocker, caplog) -> None:
     freqtrade = FreqtradeBot(conf)
     assert not freqtrade.strategy.order_types['stoploss_on_exchange']
     assert not log_has_re(".*stoploss_on_exchange .* dry-run", caplog)
+    caplog.clear()
```
xmatthias: Can you explain what you're trying to reach with this (other than a lot of added lines)?

caplog is a fixture provided by pytest, which is at function scope - which means it's a "new" object for each new test, so clearing caplog at the end of a test is pointless.

While it's possible to clear caplog in the middle of a test (to have a clean sheet) which does make sense - at the end of a test i don't really see a point in it.

xmatthias:
it's still pretty much unclear to me why you think these are necessary / what problem this is trying to address.

Tests pass with and without these - which makes them non-necessary in my eyes - unless you've seen a problem this fixes which i didn't yet.

samgermain:
> it's still pretty much unclear to me why you think these are necessary / what problem this is trying to address.
> 
> Tests pass with and without these - which makes them non-necessary in my eyes - unless you've seen a problem this fixes which i didn't yet.

it just makes it easier to debug when the tests fail, I'm doing some work in freqtradebot right now and need to read `caplog.text` from pdb sometimes

xmatthias (approved):
mmmmh, maybe you should try to use a visual editor (like vscode or pycharm) to do the test-debugging.
they offer quite nice support for test-debugging - which i think is quite tedious if you're working with raw pdb.

anyhow, LGTM 👍

samgermain:
> mmmmh, maybe you should try to use a visual editor (like vscode or pycharm) to do the test-debugging.
> they offer quite nice support for test-debugging - which i think is quite tedious if you're working with raw pdb.
> 
> anyhow, LGTM 👍

I tried to look this up once, and felt like pdb was my only option, recently i noticed the testing icon to the left on VSCode though, which I didn't see before

samgermain:
> mmmmh, maybe you should try to use a visual editor (like vscode or pycharm) to do the test-debugging.
> they offer quite nice support for test-debugging - which i think is quite tedious if you're working with raw pdb.
> 
> anyhow, LGTM 👍

Is there an extension you would recommend for this? I actually find the vscode test-debugging more difficult than pdb

PR #5550, opened 2021-09-08
Comments todos formatting log messages

samgermain:
Changed comments for shorting, added todos, formatted some code, changed log messages to be side agnostic


samgermain:
I merged develop into feat/short locally, so there's a lot of develop commits that aren't merged into feat/short on github that are showing up in here

PR #5551, opened 2021-09-08
Comments todos formatting log messages

samgermain:
Comments updated to include shorting
log messages updated
some code formatting changes
TODO messages

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -1154,9 +1154,9 @@ class FreqtradeBot(LoggingMixin):
         trade.orders.append(order_obj)
 
         trade.open_order_id = order['id']
-        trade.sell_order_status = ''  # TODO-lev: Update to exit_order_status
+        trade.sell_order_status = ''
         trade.close_rate_requested = limit
-        trade.sell_reason = sell_reason.sell_reason  # TODO-lev: Update to exit_reason
+        trade.sell_reason = sell_reason.sell_reason
         # In case of market sell orders the order can be closed immediately
         if order.get('status', 'unknown') in ('closed', 'expired'):
             self.update_trade_state(trade, trade.open_order_id, order)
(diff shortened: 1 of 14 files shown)


freqtrade/freqtradebot.py
```diff
@@ -619,7 +623,7 @@ def _notify_buy(self, trade: Trade, order_type: str) -> None:
 
     def _notify_buy_cancel(self, trade: Trade, order_type: str, reason: str) -> None:
         """
-        Sends rpc notification when a buy cancel occurred.
+        Sends rpc notification when a buy/short cancel occurred.
```
xmatthias: wouldn't something like the below make more sense (and keep wording simpler)?

Obviously not limited to this instance ...

```suggestion
        Sends rpc notification when a entry order cancel occurred.
```

freqtrade/freqtradebot.py
```diff
@@ -1181,7 +1197,7 @@ def _notify_sell(self, trade: Trade, order_type: str, fill: bool = False) -> Non
             'current_rate': current_rate,
             'profit_amount': profit_trade,
             'profit_ratio': profit_ratio,
-            'sell_reason': trade.sell_reason,
+            'sell_reason': trade.sell_reason,  # TODO-lev: change to exit_reason
```
xmatthias: i know you see this as helpful for future changes - but it's actually not (doesn't provide any value) and will even mean additional work for whoever will do this change (so it has the reverse effect of what i guess you're trying to accomplish).
Having this there once (usually where it's defined -> in `models.py`) will suffice.

Whoever will rename it will anyway have to search through all all files (can't rely on these comments to capture all cases 100%) - so this is actually adding work to whoever does the refactoring, as it's not just a "replace" - but also a "remove unnecessary comment" - which is a few more key-strokes for each line that needs to be changed.

So while these comments provide value in some instances (as a reminder that something unique needs to change and can't be forgotten) - it's not necessary to have it on every line - which will be captured by tests + mypy anyway.

xmatthias (approved):
:+1:

PR #5561, opened 2021-09-10
Liquidation price

samgermain:


freqtrade/freqtradebot.py
```diff
@@ -102,6 +106,22 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self._sell_lock = Lock()
         LoggingMixin.__init__(self, logger, timeframe_to_seconds(self.strategy.timeframe))
 
+        if self.config.get("trading_mode"):
+            self.trading_mode = TradingMode(self.config.get("trading_mode"))
```
xmatthias, line 112: One important thing here:
if we use new setting keys from the config - these should be added to `configuration.md` - as well as to `constants.py` so we can have some reliance on the type we expect (based on this - i have no idea what `trading_mode` should/could be set to without following the code flow.
I'd also add their default to the full config sample - so it's easy to copy/paste.

---

Just to throw this into the room - we seem to be having "quite some" settings in this regard ... so i think we should try to group them (all leverage/short settings into one subsetting) so they are logically grouped.

It might even make sense to move them into the exchange substructure (but this might need further consideration) - as in my understanding, the available settings will heavily depend on the selected exchange (one mode might work on binance, but not on kraken and viceversa).

freqtrade/maintenance_margin.py
```diff
@@ -0,0 +1,52 @@
+from typing import List
```
xmatthias, line 1: I think maintenance margin is only relevant for leverage - so please move this to `freqtrade.leverage` (basically move it into the subdirectory).

freqtrade/maintenance_margin.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self, exchange_name: str, trading_mode: TradingMode):
+        self.exchange_name = exchange_name
+        self.trading_mode = trading_mode
+        return
+
+    def add_new_trade(self, trade):
+        self.trades.append(trade)
+
+    def remove_trade(self, trade):
+        self.trades.remove(trade)
```
xmatthias, line 38: I don't think this is a good way to handle this, as it'll be quite error-prone and work intensive (from a code maintenance perspective) to get this right.

This will work ONLY when a trade is handled within a bot run - but will fail the moment i stop and restart the bot with a trade open.
It'll also bear the risk that modifications to the trade will be missed (it's basically storing a copy of an object from the database - but we requery the database with every run - so we risk getting out of sync with this).


A safer approach (instead of keeping a reference around) is to query the database for relevant trades whenever required.
The required info must be in the database anyway, so it's as simple as writing an appropriate query (or better yet, query helper) to fetch relevant trades (open and have an applicable maintenance mode).

I'm not sure yet if that should be done within this class, or if a list of trades should be passed into `.run()` - basically doing the query where it's called (this might simplify handling of this during backtesting)

freqtrade/freqtradebot.py
```diff
@@ -102,6 +106,22 @@ def __init__(self, config: Dict[str, Any]) -> None:
...
+
+        # Start calculating maintenance margin if on cross margin
+        # TODO: Add margin_mode to freqtrade.configuration?
+        if self.collateral == Collateral.CROSS:
+
+            self.maintenance_margin = MaintenanceMargin(
+                exchange_name=self.exchange.name,
+                trading_mode=self.trading_mode)
+
+            self.maintenance_margin.run()
```
xmatthias, line 125: i'd probably "somehow" bundle this with funding fees (and whatever additional other fees we'll need) - so from a freqtradebot.py perspective, we'll be only calling `handle_running_fees` (naming up to decide) - and that method will then (not in `freqtradebot.py`) decide which method to apply to which trades.

freqtrade/maintenance_margin.py
samgermain: it creates a circular import, I think I might remove that sub directory also, and can hopefully move the formulas onto the exchange subclasses as class methods

samgermain:
Will reopen this PR when it's more done

PR #5566, opened 2021-09-11
Remove legacy hyperopt

xmatthias:
## Summary
Remove "legacy hyperopt" implementation (using a seperate file) - and rely solely on parametrized strategies.

## Quick changelog

- Remove seperate hyperopt interface
- remove `list-hyperopts`
- remove `new-hyperopt`
- remove logic loading old hyperopt files


samgermain:
Before merging, would you be able to rebase and combine a few of the commits that do the same thing? I haven't learned the new hyperopt yet, and might try to get these changes back to compare

xmatthias:
> Before merging, would you be able to rebase and combine a few of the commits that do the same thing? I haven't learned the new hyperopt yet, and might try to get these changes back to compare

would i be able to - sure
woult it make sense ... probably not? you're much better off learning the hyperopt interface via the documentation than from looking at what was removed. Hyperopt itself is sometimes pretty convoluted - so following the details of what was removed will probably not be that beneficial.

PR #5567, opened 2021-09-12
Lev freqtradebot

samgermain:


samgermain:
@xmatthias Would you prefer if I just added these changes to the `lev-strat` PR after `lev-exchange` get's merged? This one mostly only makes changes inside freqtradebot, but it includes the commits from `lev-exchange` and `lev-strat`

I just posted it though so that you can check if something has already been implemented before adding it to freqtradebot

xmatthias:
i ... don't care? 
if it helps you to have it here that's fine - i'll however not look at this (time-problem - i can look at this, but then not look at other PR's) until lev-exchange (i assume that's the next one) is ready to be merged.

I guess just leave it here and rebase it on top of feat/short once ready?

samgermain:
> i ... don't care?
> if it helps you to have it here that's fine - i'll however not look at this (time-problem - i can look at this, but then not look at other PR's) until lev-exchange (i assume that's the next one) is ready to be merged.
> 
> I guess just leave it here and rebase it on top of feat/short once ready?

I just put it here in case you want to check what's already been done in freqtradebot, because I noticed you had made some changes to freqtradebot already

Also, lev-exchange is ready to be merged in my opinion. You can merge it whenever you think it's ready

freqtrade/freqtradebot.py
```diff
@@ -1088,21 +1222,28 @@ def _safe_exit_amount(self, pair: str, amount: float) -> float:
             raise DependencyException(
                 f"Not enough amount to exit trade. Trade-amount: {amount}, Wallet: {wallet_amount}")
 
-    def execute_trade_exit(self, trade: Trade, limit: float, sell_reason: SellCheckTuple) -> bool:
+    def execute_trade_exit(
+        self,
+        trade: Trade,
+        limit: float,
+        sell_reason: SellCheckTuple,  # TODO-lev update to exit_reason
+        side: str
```
xmatthias: I don't think "side" should be a parameter here.
it's implicitly available via `trade.is_short` - so it provides no benefit, and most likely, every caller will use `'short' if trade.is_short else 'long'`.

samgermain:
@xmatthias I think some of the tests are failing because some stuff in the strategy interface hasn't been updated for shorting. I added some `TODO-lev`'s in the most recent commit that show the spots I'm looking at

samgermain:
@xmatthias Could you take a look at `test_trailing_stop_loss_positive`, it keeps failing and the caplog is getting `ETH/USDT - Using positive stoploss: 0.01 offset: 0 profit: -0.0096%`, which I tried to figure out but didn't understand very well

tests/test_freqtradebot.py
```diff
@@ -3028,24 +3313,26 @@ def test_execute_trade_exit_insufficient_funds_error(default_conf_usdt, ticker_u
...
     # Enable profit
-    (True, 1.9, 2.2, False, True, SellType.SELL_SIGNAL.value),
+    (True, 2.19, 2.2, False, True, SellType.SELL_SIGNAL.value),
     # Disable profit
-    (False, 2.9, 3.2, True,  False, SellType.SELL_SIGNAL.value),
+    (False, 3.19, 3.2, True,  False, SellType.SELL_SIGNAL.value),
     # Enable loss
     # * Shouldn't this be SellType.STOP_LOSS.value
-    (True, 0.19, 0.22, False, False, None),
+    (True, 0.21, 0.22, False, False, None),
```
rokups: I dont quite understand how to fix this one (`is_short=True`). Trade registers as profitable because it is enters at 0.22 and exits at 0.21. Swapping these values does not help, trade still enters at 2.2 rate even though it requests 0.21, so still a profitable trade somehow (exit also at 0.22). Unexpected entry/exit rate aside, shouldnt fees make it a loss? Seems like two problems are at hand.

samgermain: > trade still enters at 2.2 rate

That might be because of this line `trade.update(limit_order[enter_side(is_short)])`, that evaluates to `trade.update(limit_sell_order_usdt)` which has a price of 2.2

samgermain:
@xmatthias It's possible that some strategy stuff is causing some tests to fail, I had to update some stuff related to stoploss in the strategy interface to get some other tests to pass

freqtrade/strategy/interface.py
```diff
@@ -840,42 +840,59 @@ def stop_loss_reached(self, current_rate: float, trade: Trade,
             else:
                 logger.warning("CustomStoploss function did not return valid stoploss")
 
-        if self.trailing_stop and trade.stop_loss < (low or current_rate):
+        if self.trailing_stop and (
+            (trade.stop_loss < (low or current_rate) and not trade.is_short) or
+            (trade.stop_loss > (high or current_rate) and trade.is_short)
```
xmatthias: I'd extract this whole logic (is lower / higher depending on boolean flag).
it's used not just here, but at least twice below (obviously we'll need to see how identical it really is)...
but at a first glance, this seems ideal to extract, as we'll want to have pretty extensive unittests for this... having a small edgecase wrong here can have sever consequences.

freqtrade/strategy/interface.py
```diff
@@ -840,28 +840,38 @@ def stop_loss_reached(self, current_rate: float, trade: Trade,
...
+            if not (self.trailing_only_offset_is_reached and (
+                (bound_profit < sl_offset and not trade.is_short) or
+                (bound_profit > sl_offset and trade.is_short)
+            )):
                 # Specific handling for trailing_stop_positive
-                if self.trailing_stop_positive is not None and high_profit > sl_offset:
+                if self.trailing_stop_positive is not None and (
+                    (bound_profit > sl_offset and not trade.is_short) or
+                    (bound_profit < sl_offset and trade.is_short)
+                ):
```
xmatthias: I don't think this change is correct.

we're comparing `bound_profit` with `sl_offset`. 
even if we're in a short (absolute price is reversed) - a trade risking to stop out will have negative profits - while a trade that needs an adjusted stop will have a positive profit.

tests/plugins/test_pairlist.py
```diff
@@ -665,6 +665,7 @@ def test_PerformanceFilter_error(mocker, whitelist_conf, caplog) -> None:
 
 
 @pytest.mark.usefixtures("init_persistence")
+# TODO-lev: @pytest.mark.parametrize('is_short', [True, False])
```
xmatthias: i'm not certain if it makes sense to test this twice.
PerformanceFilter is only looking at the database's "profit" column - so if we have that right, the direction won't matter.

freqtrade/enums/rpcmessagetype.py
```diff
@@ -5,15 +5,25 @@ class RPCMessageType(Enum):
...
     SELL_FILL = 'sell_fill'
     SELL_CANCEL = 'sell_cancel'
     PROTECTION_TRIGGER = 'protection_trigger'
     PROTECTION_TRIGGER_GLOBAL = 'protection_trigger_global'
 
+    SHORT = 'short'
+    SHORT_FILL = 'short_fill'
+    SHORT_CANCEL = 'short_cancel'
+
+    EXIT_SHORT = 'exit_short'
```
xmatthias: Don't think this is used - otherwise i'd have had to modify this in the stoploss tests as well (this will again break a multitude of tests - probably ...)

freqtrade/freqtradebot.py
```diff
@@ -101,14 +102,19 @@ def __init__(self, config: Dict[str, Any]) -> None:
...
+        self.trading_mode: TradingMode = TradingMode.SPOT
+        self.collateral_type: Optional[Collateral] = None
+
         if 'trading_mode' in self.config:
             self.trading_mode = TradingMode(self.config['trading_mode'])
-        else:
-            self.trading_mode = TradingMode.SPOT
+
+        if 'collateral_type' in self.config:
+            self.collateral_type = Collateral(self.config['collateral_type'])
```
xmatthias, line 116: Please add explicit type-checking in `constants.py` for these.
Otherwise we cannot rely on the values of config (it can be anything, really) - which means this will break with an ugly, non-saying message if a user misstypes the value.

freqtrade/freqtradebot.py
```diff
@@ -465,19 +479,35 @@ def create_trade(self, pair: str) -> bool:
...
-                # TODO-lev: Does the below need to be adjusted for shorts?
-                if self._check_depth_of_market_buy(pair, bid_check_dom):
-                    # TODO-lev: pass in "enter" as side.
-
-                    return self.execute_entry(pair, stake_amount, enter_tag=enter_tag)
+                if self._check_depth_of_market(
+                    pair,
+                    bid_check_dom,
+                    side=signal
+                ):
```
xmatthias: i know this is the "black" way of formatting, and i'm not against it in general - but it should only be applied when necessary - which it doesn't seem to be here.

freqtrade/freqtradebot.py
```diff
@@ -543,10 +629,11 @@ def execute_entry(self, pair: str, stake_amount: float, price: Optional[float] =
         if not stake_amount:
             return False
 
-        logger.info(f"Buy signal found: about create a new trade for {pair} with stake_amount: "
+        log_type = f"{name} signal found"
+        logger.info(f"{log_type}: about create a new trade for {pair} with stake_amount: "
```
xmatthias: can't we combine these 2?
it seems pretty pointless to define a variable as fstring, that's then used only once in an f-string - the line after ...

freqtrade/freqtradebot.py
```diff
@@ -960,22 +1075,42 @@ def check_handle_timedout(self) -> None:
...
+                            self.strategy.check_buy_timeout
+                        ),
+                        default_retval=False
+                    )(
+                        pair=trade.pair,
+                        trade=trade,
+                        order=order
+                    )
+                 )
+            ):
```
xmatthias: i find the "new" formatting pretty unleggible.
the old formatting was selected very carefully to help see where ands / ors happen - which is now hidden (same below).

freqtrade/freqtradebot.py
```diff
@@ -1129,21 +1269,27 @@ def _safe_exit_amount(self, pair: str, amount: float) -> float:
...
+        trade: Trade,
+        limit: float,
+        sell_reason: SellCheckTuple,  # TODO-lev update to exit_reason
+    ) -> bool:
         """
         Executes a trade exit for the given trade and limit
         :param trade: Trade instance
         :param limit: limit rate for the sell order
         :param sell_reason: Reason the sell was triggered
+        :param side: "buy" or "sell"
```
xmatthias: This is no longer a parameter.

freqtrade/freqtradebot.py
samgermain: I think it puts the line past 100 characters then

freqtrade/freqtradebot.py
samgermain: I think it's over passed length 100 if it's not, a lot of the time that's why I spread it out, but I'll double check this one

freqtrade/freqtradebot.py
samgermain: I'm pretty sure this has to do with line length again

freqtrade/freqtradebot.py
xmatthias: it's a 2 line fstring already - move more to the 2nd line and it'll work. :smile: 
and even if not - a multi-line fstring (in parenteses) is better than instanciating a variable we won't need.

freqtrade/enums/rpcmessagetype.py
samgermain: I hope you just mean EXIT_SHORT isnt used, because SHORT is apparently

freqtrade/freqtradebot.py
samgermain: Hopefully you like the new formatting

freqtrade/freqtradebot.py
```diff
@@ -1230,7 +1351,7 @@ def _notify_exit(self, trade: Trade, order_type: str, fill: bool = False) -> Non
         profit_trade = trade.calc_profit(rate=profit_rate)
         # Use cached rates here - it was updated seconds ago.
         current_rate = self.exchange.get_rate(
-            trade.pair, refresh=False, side="sell") if not fill else None
+            trade.pair, refresh=False, side=trade.exit_side) if not fill else None
         profit_ratio = trade.calc_profit_ratio(profit_rate)
         gain = "profit" if profit_ratio > 0 else "loss"
```
xmatthias, line 1357: Below this - we'll need to make sure to use the correct `RPCMessageType`. it's currently only using SELL/SELL_FILL - is that OK /will that be handled in rpc for shorts, or will we need a seperate sell-messagetype (apart from that, it's not sending the "direction" - so we currently can't handle this in rpc).

samgermain: I don't know a lot about RPC, it was one of those variables that I was leaving as BUY and SELL, because i didn't know if we should change the naming yet

xmatthias: ok, then let's leave this for now ... we'll have to look at RPC anyway - and this can work both ways ...

freqtrade/enums/rpcmessagetype.py
xmatthias: yes ... consider it a "temporary reminder for myself" i forgot to remove :laughing:

xmatthias (approved):
let's merge this and move on - we'll tackle additional things going forward :+1:

samgermain:
> let's merge this and move on - we'll tackle additional things going forward 👍

That was quicker than other PRs, was this PR better, or was it just less thoroughly vetted? I'm wondering because I want to know if I should more thoroughly vet the code that was added here in comparison to the code from branches like `lev-exchange` and `margin-db`

xmatthias:
i guess it was better? 
it'll always depend on the topic / files changed how easy/straightforward a change is or isn't.

This also leaves a lot of "TODO-lev"'s behind (especially in tests)... obviously we can finish them here, but that'd drag this PR out - so i'd prefer to do these in smaller followup PR's.

PR #5576, opened 2021-09-15
Binance Liquidation Price Formulas

arunavo4:

## Summary

Added Formulas to Calculate Liquidation Price of Binance USDⓈ-M Futures

Solve the issue: #___

## Quick changelog

- Added Liquidation Price for Binance
- Added Tests for Liquidation Price for Binance for Both Isolated and Cross

## What's new?

### DIFF
--- a/freqtrade/leverage/liquidation_price.py
+++ b/freqtrade/leverage/liquidation_price.py
@@ -36,9 +36,10 @@ def liquidation_price(
                 f"maintenance_amt, position, entry_price, mm_rate "
                 f"is required by liquidation_price when exchange is {exchange_name.lower()}")
 
-        return binance(open_rate, is_short, leverage, trading_mode, collateral, wallet_balance,
-                       mm_ex_1, upnl_ex_1, maintenance_amt,
-                       position, entry_price, mm_rate)
+        # Suppress incompatible type "Optional[float]"; expected "float" as the check exists above.
+        return binance(open_rate, is_short, leverage, trading_mode, collateral,  # type: ignore
+                       wallet_balance, mm_ex_1, upnl_ex_1, maintenance_amt,  # type: ignore
+                       position, entry_price, mm_rate)  # type: ignore
     elif exchange_name.lower() == "kraken":
         return kraken(open_rate, is_short, leverage, trading_mode, collateral)
     elif exchange_name.lower() == "ftx":
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -1059,21 +1059,17 @@ class Trade(_DECL_BASE, LocalTrade):
         return total_open_stake_amount or 0
 
     @staticmethod
-    def get_overall_performance(minutes=None) -> List[Dict[str, Any]]:
+    def get_overall_performance() -> List[Dict[str, Any]]:
         """
         Returns List of dicts containing all Trades, including profit and trade count
         NOTE: Not supported in Backtesting.
         """
-        filters = [Trade.is_open.is_(False)]
-        if minutes:
-            start_date = datetime.now(timezone.utc) - timedelta(minutes=minutes)
-            filters.append(Trade.close_date >= start_date)
         pair_rates = Trade.query.with_entities(
             Trade.pair,
             func.sum(Trade.close_profit).label('profit_sum'),
             func.sum(Trade.close_profit_abs).label('profit_sum_abs'),
             func.count(Trade.pair).label('count')
-        ).filter(*filters)\
+        ).filter(Trade.is_open.is_(False))\
             .group_by(Trade.pair) \
             .order_by(desc('profit_sum_abs')) \
             .all()
--- a/config_examples/config_full.example.json
+++ b/config_examples/config_full.example.json
@@ -149,7 +149,9 @@
             },
             "sell_fill": "on",
             "buy_cancel": "on",
-            "sell_cancel": "on"
+            "sell_cancel": "on",
+            "protection_trigger": "off",
+            "protection_trigger_global": "on"
         },
         "reload": true,
         "balance_dust_level": 0.01
--- a/docs/docker_quickstart.md
+++ b/docs/docker_quickstart.md
@@ -149,6 +149,24 @@ You'll then also need to modify the `docker-compose.yml` file and uncomment the 
 
 You can then run `docker-compose build` to build the docker image, and run it using the commands described above.
 
+### Troubleshooting
+
+#### Docker on Windows
+
+* Error: `"Timestamp for this request is outside of the recvWindow."`
+  * The market api requests require a synchronized clock but the time in the docker container shifts a bit over time into the past.
+    To fix this issue temporarily you need to run `wsl --shutdown` and restart docker again (a popup on windows 10 will ask you to do so).
+    A permanent solution is either to host the docker container on a linux host or restart the wsl from time to time with the scheduler.
+    ```
+    taskkill /IM "Docker Desktop.exe" /F
+    wsl --shutdown
+    start "" "C:\Program Files\Docker\Docker\Docker Desktop.exe"
+    ```
+
+!!! Warning
+    Due to the above, we do not recommend the usage of docker on windows for production setups, but only for experimentation, datadownload and backtesting.
+    Best use a linux-VPS for running freqtrade reliably.
+
 ## Plotting with docker-compose
 
 Commands `freqtrade plot-profit` and `freqtrade plot-dataframe` ([Documentation](plotting.md)) are available by changing the image to `*_plot` in your docker-compose.yml file.
--- a/docs/telegram-usage.md
+++ b/docs/telegram-usage.md
@@ -93,7 +93,9 @@ Example configuration showing the different settings:
         "buy_cancel": "silent",
         "sell_cancel": "on",
         "buy_fill": "off",
-        "sell_fill": "off"
+        "sell_fill": "off",
+        "protection_trigger": "off",
+        "protection_trigger_global": "on"
     },
     "reload": true,
     "balance_dust_level": 0.01
@@ -103,6 +105,7 @@ Example configuration showing the different settings:
 `buy` notifications are sent when the order is placed, while `buy_fill` notifications are sent when the order is filled on the exchange.
 `sell` notifications are sent when the order is placed, while `sell_fill` notifications are sent when the order is filled on the exchange.
 `*_fill` notifications are off by default and must be explicitly enabled.
+`protection_trigger` notifications are sent when a protection triggers and `protection_trigger_global` notifications trigger when global protections are triggered.
 
 
 `balance_dust_level` will define what the `/balance` command takes as "dust" - Currencies with a balance below this will be shown.
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -284,6 +284,15 @@ CONF_SCHEMA = {
                             'enum': TELEGRAM_SETTING_OPTIONS,
                             'default': 'off'
                         },
+                        'protection_trigger': {
+                            'type': 'string',
+                            'enum': TELEGRAM_SETTING_OPTIONS,
+                            'default': 'off'
+                        },
+                        'protection_trigger_global': {
+                            'type': 'string',
+                            'enum': TELEGRAM_SETTING_OPTIONS,
+                        },
                     }
                 },
                 'reload': {'type': 'boolean'},
--- a/freqtrade/enums/rpcmessagetype.py
+++ b/freqtrade/enums/rpcmessagetype.py
@@ -11,6 +11,8 @@ class RPCMessageType(Enum):
     SELL = 'sell'
     SELL_FILL = 'sell_fill'
     SELL_CANCEL = 'sell_cancel'
+    PROTECTION_TRIGGER = 'protection_trigger'
+    PROTECTION_TRIGGER_GLOBAL = 'protection_trigger_global'
 
     def __repr__(self):
         return self.value
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -1351,14 +1351,26 @@ class FreqtradeBot(LoggingMixin):
         if not trade.is_open:
             if not stoploss_order and not trade.open_order_id:
                 self._notify_exit(trade, '', True)
-            self.protections.stop_per_pair(trade.pair)
-            self.protections.global_stop()
+            self.handle_protections(trade.pair)
             self.wallets.update()
         elif not trade.open_order_id:
             # Buy fill
             self._notify_enter_fill(trade)
 
         return False
+
+    def handle_protections(self, pair: str) -> None:
+        prot_trig = self.protections.stop_per_pair(pair)
+        if prot_trig:
+            msg = {'type': RPCMessageType.PROTECTION_TRIGGER, }
+            msg.update(prot_trig.to_json())
+            self.rpc.send_msg(msg)
+
+        prot_trig_glb = self.protections.global_stop()
+        if prot_trig_glb:
+            msg = {'type': RPCMessageType.PROTECTION_TRIGGER_GLOBAL, }
+            msg.update(prot_trig_glb.to_json())
+            self.rpc.send_msg(msg)
 
     def apply_fee_conditional(self, trade: Trade, trade_base_currency: str,
                               amount: float, fee_abs: float) -> float:
--- a/freqtrade/leverage/interest.py
+++ b/freqtrade/leverage/interest.py
@@ -20,7 +20,7 @@ def interest(
 
         :param exchange_name: The exchanged being trading on
         :param borrowed: The amount of currency being borrowed
-        :param rate: The rate of interest (i.e daily interest rate)
+        :param rate: The rate of interest
         :param hours: The time in hours that the currency has been borrowed for
 
         Raises:
@@ -36,8 +36,7 @@ def interest(
         # Rounded based on https://kraken-fees-calculator.github.io/
         return borrowed * rate * (one+ceil(hours/four))
     elif exchange_name == "ftx":
-        # As Explained under #Interest rates section in
-        # https://help.ftx.com/hc/en-us/articles/360053007671-Spot-Margin-Trading-Explainer
-        return borrowed * rate * ceil(hours)/twenty_four
+        # TODO-lev: Add FTX interest formula
+        raise OperationalException(f"Leverage not available on {exchange_name} with freqtrade")
     else:
         raise OperationalException(f"Leverage not available on {exchange_name} with freqtrade")
--- a/freqtrade/persistence/pairlock_middleware.py
+++ b/freqtrade/persistence/pairlock_middleware.py
@@ -30,7 +30,8 @@ class PairLocks():
             PairLocks.locks = []
 
     @staticmethod
-    def lock_pair(pair: str, until: datetime, reason: str = None, *, now: datetime = None) -> None:
+    def lock_pair(pair: str, until: datetime, reason: str = None, *,
+                  now: datetime = None) -> PairLock:
         """
         Create PairLock from now to "until".
         Uses database by default, unless PairLocks.use_db is set to False,
@@ -52,6 +53,7 @@ class PairLocks():
             PairLock.query.session.commit()
         else:
             PairLocks.locks.append(lock)
+        return lock
 
     @staticmethod
     def get_pair_locks(pair: Optional[str], now: Optional[datetime] = None) -> List[PairLock]:
--- a/freqtrade/plugins/protectionmanager.py
+++ b/freqtrade/plugins/protectionmanager.py
@@ -6,6 +6,7 @@ from datetime import datetime, timezone
 from typing import Dict, List, Optional
 
 from freqtrade.persistence import PairLocks
+from freqtrade.persistence.models import PairLock
 from freqtrade.plugins.protections import IProtection
 from freqtrade.resolvers import ProtectionResolver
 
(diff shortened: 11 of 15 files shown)


arunavo4:
@samgermain need your help with `pytest` on this one. Inputs and Outputs are here at [How to Calculate Liquidation Price of USDⓈ-M Futures Contracts](https://www.binance.com/en/support/faq/b3c689c1f50a44cabb3a84e663b81d93)
Check the Section `Example: Calculating Liquidation Price in One-way & Cross Margin mode`

freqtrade/enums/marginmode.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+
+class MarginMode(Enum):
+    """
+        Enum to distinguish between
+        one-way mode or hedge mode in Futures (Cross and Isolated) or Margin Trading
+    """
+    ONE_WAY = "one-way"
+    HEDGE = "hedge"
```
samgermain: Just a reminder that we should remove this before it get's merged

freqtrade/leverage/liquidation_price.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def binance(
+    open_rate: float,
+    is_short: bool,
+    leverage: float,
+    margin_mode: MarginMode,
+    trading_mode: TradingMode,
+    collateral: Collateral,
+    **kwargs
```
samgermain: Intead of `**kwargs` can you change this so that all of `maintenance_amount_long`, `entry_price_1_long` ... are arguments to this function?

samgermain (changes requested):
So when I said give me example input and output and I'll write a test, I meant example input and output to the liquidation price function

An example would be, using `liquidation_price` with

    exchange_name="binance"
    open_rate=3
    is_short=True
    leverage=3
    maintenance_amount_long=20
    entry_price_1_long=2.8
    
The function should return 

    1.5 
    
as a result
    ...

freqtrade/leverage/liquidation_price.py
arunavo4: That would make it messy, @xmatthias do you also want how @samgermain is suggesting

samgermain: https://github.com/freqtrade/freqtrade/pull/5379#discussion_r684897541

arunavo4: Ok, Thanks for the clarification @samgermain I will make the changes.

samgermain: You will probably need to use `Optional` from the `typing` library

xmatthias: Yes i do ...
using kwargs for a method where inputs are clear (and either mandatory, or have predefined defaults) will prevent developer tooling (mypy, ...) from helping us detect errors (like a wrong type, misspelled argument, ...).
You'll have to list the arguments anywhere where you call it this function - so it's better to allow tools to help us detect errors (i'm lazy sometimes, and definitely too lazy to check calls for typos in arguments ...).

If you have 10ds of parameters for a function (like in this case) , maybe consider creating a structure for it (like a NamedTuple or similar) - which will still preserve some safety - while not causing pages-long definitions.

---

I'm however not sure where all these arguments (the data inside) should come from ... they are either part of (one or more) responses from the exchange - or part of the open trade.
Whatever it is, it might make sense to pass that prat into this method ... ? 

If it requires 2-3 calls for binance, and 1 for ftx (different modes to calculate require different data ...) - maybe consider moving the whole logic to the respective exchange classes - where you can "gather" the data from exchange directly without a lot of boilerplate.

freqtrade/enums/marginmode.py
samgermain: **Conversation from the [closed PR](https://github.com/freqtrade/freqtrade/pull/5379#discussion_r700739993)**

> **xmatthias** 15 days ago •
> 
> i assume this is the binance futures "hedge mode" ?
> 
> To be honest, i think if we don't start to keep things simple (focus on a few selected modes supported by multiple exchanges) - this will never have a chance to be merged.
> 
> Personally, i'd have focused on cross-margin on binance, "regular short" (probably still margin) on kraken - and futures on FTX (they seem to all work pretty similarily).
> These seem pretty straightforward at a first glance (what binance futures clearly does not).
> 
> Additional modes can always be added at a later point - but adding too much in one go will make everything take forever - and we'll also want to do real trading tests of a few 100 trades for each mode/exchange before merging ... so having too many exchanges WILL become problematic for sure.
> **arunavo4** 12 days ago
> 
> **xmatthias** These modes are only supported for Binance so for other exchanges its just simple margin trading which will not have any modes so if we want to test it we can just test it for binance the check is already there for that.
> 
>  **xmatthias** 12 days ago
> 
> My point is - we should focus on the simple things to get the whole short/leverage stuff working.
> If we start supporting the "special binance modes" (whichever ones they are) - it'll become a very long endavor (please consider that we'll have to live-test each and every one of these different modes, for several trades (not just 2-3)) - so each additional mode will probably add at least 2-3 weeks of testing time alone (not talking about development time).
> 
> **samgermain** 11 days ago
> 
> **arunavo4** If you want to make it easier to get your code for hedge mode back, I would finish up the regular liquidation formulas with their tests, then after that's done, remove the hedge code in a single commit. Then when you want to add it back, you could just revert that commit
> **arunavo4** 10 days ago
> 
> Yea lets do that keep that code to add later
> 
> **samgermain** 10 days ago
> 
>  > Yea lets do that keep that code to add later
> 
> Just make a good commit message that you can find later

freqtrade/leverage/liquidation_price.py
arunavo4: > I'm however not sure where all these arguments (the data inside) should come from ... they are either part of (one or more) responses from the exchange - or part of the open trade.
Whatever it is, it might make sense to pass that prat into this method ... ?

> If it requires 2-3 calls for binance, and 1 for ftx (different modes to calculate require different data ...) - maybe consider moving the whole logic to the respective exchange classes - where you can "gather" the data from exchange directly without a lot of boilerplate.


@samgermain your opinion? How can we address this? Since you came up with the base structure. I am only filling the gaps here.

samgermain: Why would it require 2-3 calls for binance? Just get whatever data is required using whatever technique you want. Using a NamedTuple sounds like a good idea, or just a class with only properties (maybe a [@dataclass](https://docs.python.org/3/library/dataclasses.html)?), I don't know if those are the same thing.

arunavo4:
[Converted kwargs to parms](https://github.com/arunavo4/freqtrade/commit/13331e4dc7afe0c52860efb9541d469993e02f6c)
Made the changes to function args

freqtrade/leverage/liquidation_price.py
```diff
@@ -0,0 +1,229 @@
...
+    maintenance_amount_short: float,
+    position_1_both: float,
+    entry_price_1_both: float,
+    position_1_long: float,
+    entry_price_1_long: float,
+    position_1_short: float,
+    entry_price_1_short: float,
+    maintenance_margin_rate_both: float,
+    maintenance_margin_rate_long: float,
+    maintenance_margin_rate_short: float,
```
samgermain: I'm trying to think if there's a better way method to organize parameters, it looks like the "both" parameters are never used at the same time as the long and short parameters, so maybe if there was just a maintenance_margin param that was a tuple, with only 1 value for one-way. Having the other exchange formulas will also help this, so that you can line up there parameters in the parent function.

It would be best not to do too much more for hedge mode right now though, you can try to organize it so that you don't have any conflicts when you try to revert back to using hedge mode, but definately don't write any tests for hedge mode or anything, like right now we don't need to worry about how to organize the parameters, but when it's done, and we also have the kraken and ftx formulas, then we can decide how to include hedge mode so that everything aligns with the other exchanges as well

freqtrade/leverage/liquidation_price.py
xmatthias: it's not the same but similar ... `@dataclass` sounds good (had to double-check when it was introduced, as we're still supporting python 3.7 - but looks like we're fine (otherwise CI will let us know).

freqtrade/leverage/liquidation_price.py
arunavo4: yea I think we can think about organising later when other exchange formulas are there. Right now I want to get it to a point where we can start testing this in dry mode

arunavo4: I think now I will remove the hedge mode and move it to another branch on my fork and keep it. Let's get the tests done for this so we are ready to merge this.

samgermain: ok, I'll need an example of argument and return values to write the one test example, and then you'll be able to write the rest, you should probably remove hedge mode before you send those values

arunavo4:
Removed Hedge-Mode
[Binance Liquidation Price Hedge-Mode Removed](https://github.com/arunavo4/freqtrade/pull/5576/commits/6a81730a74847edd6e431fe6ede299ac0639174e)

freqtrade/leverage/liquidation_price.py
```diff
@@ -11,7 +11,13 @@ def liquidation_price(
...
     trading_mode: TradingMode,
     collateral: Optional[Collateral],
-    margin_mode: Optional[MarginMode]
+    wallet_balance: Optional[float],
+    maintenance_margin_ex_1: Optional[float],
+    unrealized_pnl_ex_1: Optional[float],
+    maintenance_amount_both: Optional[float],
+    position_1_both: Optional[float],
+    entry_price_1_both: Optional[float],
+    maintenance_margin_rate_both: Optional[float]
```
samgermain: lets take off the both

freqtrade/leverage/liquidation_price.py
```diff
@@ -0,0 +1,190 @@
...
+def liquidation_price(
+    exchange_name: str,
+    open_rate: float,
+    is_short: bool,
+    leverage: float,
+    trading_mode: TradingMode,
+    collateral: Optional[Collateral],
+    wallet_balance: Optional[float],
+    maintenance_margin_ex_1: Optional[float],
+    unrealized_pnl_ex_1: Optional[float],
```
samgermain: Why is ex_1 in the parameter name?

arunavo4: It's in the description @samgermain 
```
:param maintenance_margin_ex_1: Maintenance Margin of all other contracts, excluding Contract 1.
          If it is an isolated margin mode, then TMM=0

:param unrealized_pnl_ex_1: Unrealized PNL of all other contracts, excluding Contract 1.
  If it is an isolated margin mode, then UPNL=0
```

freqtrade/leverage/liquidation_price.py
arunavo4: ok but needs to be added back when hedge mode is there.

arunavo4: Removed here https://github.com/freqtrade/freqtrade/pull/5576/commits/e5a9e118990609b925524d640c2a288921c8a17f

freqtrade/enums/marginmode.py
arunavo4: I have added Tests for this now

freqtrade/leverage/liquidation_price.py
```diff
@@ -41,147 +53,90 @@ def exception(
...
+        :param unrealized_pnl_ex_1: Unrealized PNL of all other contracts, excluding Contract 1.
+          If it is an isolated margin mode, then UPNL=0
 
-            * *position_1_both* (``float``) --
-              Absolute value of BOTH position size (one-way mode)
+        :param maintenance_amount: Maintenance Amount of position (one-way mode)
 
-            * *entry_price_1_both* (``float``) --
-              Entry Price of BOTH position (one-way mode)
+        :param position_1: Absolute value of position size (one-way mode)
```
samgermain: can you take the 1 off this and entry_price ? when we include hedge mode, I think it might be better to just use a tuple or an array instead of having 3 separate parameters for the same thing

freqtrade/leverage/liquidation_price.py
```diff
@@ -41,147 +53,90 @@ def exception(
...
-              Maintenance Amount of BOTH position (one-way mode)
-
-            * *maintenance_amount_long* (``float``) --
-              Maintenance Amount of LONG position (hedge mode)
+        :param wallet_balance: Wallet Balance is crossWalletBalance in Cross-Margin Mode.
+          Wallet Balance is isolatedWalletBalance in Isolated Margin Mode
 
-            * *maintenance_amount_short* (``float``) --
-              Maintenance Amount of SHORT position (hedge mode)
+        :param maintenance_margin_ex_1: Maintenance Margin of all other contracts, excluding Contract 1.
```
samgermain: I think it might be better to shorten these variables to mm_ex_1, there just a lot of places where maintenance_margin is printed out and it clutters things up a bit. Keep the part to the right of the colon as Maintenance Margin though

Also, if this is our function to calculate maintenance margin, and it requires our other trades maintenance margins, then how does this function work. Is it some sort of induction type thing where we know the value is 0 when we enter our first trade, and then for our second trade we just take the value from the first one, and then we can keep doing that for each subsequent trade?

arunavo4:
@samgermain 
❯ mypy freqtrade
freqtrade/persistence/models.py:356: error: Missing positional arguments "wallet_balance", "maintenance_margin_ex_1", "unrealized_pnl_ex_1", "maintenance_amount", "position_1", "entry_price_1", "maintenance_margin_rate" in call to "liquidation_price"
Found 1 error in 1 file (checked 153 source files)

This seems like a big change can u take a look at it where the values will come from?

freqtrade/leverage/liquidation_price.py
```diff
@@ -41,147 +53,90 @@ def exception(
...
+          If it is an isolated margin mode, then TMM=0
 
-            * *side_1_both* (``int``) --
-              Direction of BOTH position, 1 as long position, -1 as short position
-              Derived from is_short
+        :param unrealized_pnl_ex_1: Unrealized PNL of all other contracts, excluding Contract 1.
```
samgermain: maybe change the variable to upnl_ex_1. Also, do we have a method to obtain this value? I'm hoping that we can get a liquidation_price formula that we are able to use within freqtrade before merging formulas in, and then realizing we have no method of calculating pnl

arunavo4: @samgermain I have not gone through the codebase well enough to tell you where it will come from but the current formula for liquidation price that finance provides will need this

freqtrade/leverage/liquidation_price.py
arunavo4: Maintenance margin has its own slabs and formula to calculate I think we need a dict in another util file to generate this

samgermain:
> @samgermain
> ❯ mypy freqtrade
> freqtrade/persistence/models.py:356: error: Missing positional arguments "wallet_balance", "maintenance_margin_ex_1", "unrealized_pnl_ex_1", "maintenance_amount", "position_1", "entry_price_1", "maintenance_margin_rate" in call to "liquidation_price"
> Found 1 error in 1 file (checked 153 source files)
> 
> This seems like a big change can u take a look at it where the values will come from?

So that's the spot in freqtrade that actually calls the isolated_liq method, so we have to make sure that we can actually access the properties that we need there. 

Maybe we should separate isolated and cross liquidation formulas into separate functions.

If we were going to not change anything though, then

**maintenance_margin_ex_1, maintenance_amount, maintenance_margin_rate:** might be None
 **position_1**: would be amount * current_price (So I think we'll need to pass in current price)
 **entry_price_1**: probably open_rate

**unrealized_pnl_ex_1**: we have to calculate this somehow I think
**wallet_balance**: I'm not sure what to do about. @xmatthias ?

freqtrade/leverage/liquidation_price.py
arunavo4: https://github.com/freqtrade/freqtrade/pull/5576/commits/da00b21cd50c1341dfd97b67e72257378d5ada74

freqtrade/leverage/liquidation_price.py
arunavo4: https://github.com/freqtrade/freqtrade/pull/5576/commits/da00b21cd50c1341dfd97b67e72257378d5ada74

freqtrade/leverage/liquidation_price.py
arunavo4: https://github.com/freqtrade/freqtrade/pull/5576/commits/da00b21cd50c1341dfd97b67e72257378d5ada74

arunavo4:
> > @samgermain
> > ❯ mypy freqtrade
> > freqtrade/persistence/models.py:356: error: Missing positional arguments "wallet_balance", "maintenance_margin_ex_1", "unrealized_pnl_ex_1", "maintenance_amount", "position_1", "entry_price_1", "maintenance_margin_rate" in call to "liquidation_price"
> > Found 1 error in 1 file (checked 153 source files)
> > This seems like a big change can u take a look at it where the values will come from?
> 
> So that's the spot in freqtrade that actually calls the isolated_liq method, so we have to make sure that we can actually access the properties that we need there.
> 
> Maybe we should separate isolated and cross liquidation formulas into separate functions.
> 
> If we were going to not change anything though, then
> 
> **maintenance_margin_ex_1, maintenance_amount, maintenance_margin_rate:** might be None
> **position_1**: would be amount * current_price (So I think we'll need to pass in current price)
> **entry_price_1**: probably open_rate
> 
> **unrealized_pnl_ex_1**: we have to calculate this somehow I think
> **wallet_balance**: I'm not sure what to do about. @xmatthias ?

Need these once we can collect all the inputs I can add them and we can close this.

freqtrade/leverage/liquidation_price.py
samgermain: > @samgermain I have not gone through the codebase well enough to tell you where it will come from but the current formula for liquidation price that finance provides will need this

I think we'll need to calculate it, we don't have this value stored anywhere

freqtrade/leverage/liquidation_price.py
arunavo4: Tests added at https://github.com/arunavo4/freqtrade/pull/5576/commits/98d2d2f485f9605b5aa8eb7dd674b3bb3150f588

freqtrade/persistence/models.py
```diff
@@ -359,7 +359,14 @@ def set_isolated_liq(self, isolated_liq: Optional[float]):
...
                 leverage=self.leverage,
                 trading_mode=self.trading_mode,
-                collateral=Collateral.ISOLATED
+                collateral=Collateral.ISOLATED,
+                mm_ex_1=0.0,
+                upnl_ex_1=0.0,
+                # TODO-lev: position=amount * current_price,
+                # TODO-lev: maintenance_amt,
+                # TODO-lev: wallet_balance,
+                # TODO-lev: mm_rate,
```
samgermain: We'll need to figure out how to get these values to pass in

Also, does this mean we don't need `is_short` or `leverage` as parameters?

samgermain: Another option is that we can calculate this value inside freqtradebot and assign trade.set_isolated_liq(value)

@xmatthias Which one do you think would be better?

xmatthias: i might not be understanding the question?
after a quick glance - it seems like the only thing that's not available as part of the trade is `wallet_balance` ... which we can (and should) pass in for sure.

samgermain: > i might not be understanding the question?
> after a quick glance - it seems like the only thing that's not available as part of the trade is `wallet_balance` ... which we can (and should) pass in for sure.

ok, nevermind

samgermain: The only two missing params left here are `maintenance_amt`, `mm_rate`
Which I don't know how to get or calculate

arunavo4:
@samgermain 
`freqtrade/plugins/pairlist/PerformanceFilter.py:50: error: Too many arguments for "get_overall_performance" of "Trade"`

https://github.com/arunavo4/freqtrade/blob/375949e41a6fe60dd5e1ae279d58cd142a54b042/freqtrade/plugins/pairlist/PerformanceFilter.py#L50

The function has an extra parameter being passed.

https://github.com/arunavo4/freqtrade/blob/375949e41a6fe60dd5e1ae279d58cd142a54b042/freqtrade/persistence/models.py#L1061-L1066

Want to confirm before removing the extra param

xmatthias:
> @samgermain
> `freqtrade/plugins/pairlist/PerformanceFilter.py:50: error: Too many arguments for "get_overall_performance" of "Trade"`
> 
> https://github.com/arunavo4/freqtrade/blob/375949e41a6fe60dd5e1ae279d58cd142a54b042/freqtrade/plugins/pairlist/PerformanceFilter.py#L50
> 
> The function has an extra parameter being passed.
> 
> https://github.com/arunavo4/freqtrade/blob/375949e41a6fe60dd5e1ae279d58cd142a54b042/freqtrade/persistence/models.py#L1061-L1066
> 
> Want to confirm before removing the extra param

Sounds like a mistake while merging the branches to me (without looking at it in detail).

#5582 introduced this parameter, and it should remain there ... as it does make sense to limit the amount of lookback-time for performance filter (i'd anyway expect 0 changes to any filters for short/leverage trading ... )

arunavo4:
> > @samgermain
> > `freqtrade/plugins/pairlist/PerformanceFilter.py:50: error: Too many arguments for "get_overall_performance" of "Trade"`
> > https://github.com/arunavo4/freqtrade/blob/375949e41a6fe60dd5e1ae279d58cd142a54b042/freqtrade/plugins/pairlist/PerformanceFilter.py#L50
> > The function has an extra parameter being passed.
> > https://github.com/arunavo4/freqtrade/blob/375949e41a6fe60dd5e1ae279d58cd142a54b042/freqtrade/persistence/models.py#L1061-L1066
> > Want to confirm before removing the extra param
> 
> Sounds like a mistake while merging the branches to me (without looking at it in detail).
> 
> #5582 introduced this parameter, and it should remain there ... as it does make sense to limit the amount of lookback-time for performance filter (i'd anyway expect 0 changes to any filters for short/leverage trading ... )

ok then @xmatthias I will add it back

PR #5579, opened 2021-09-16
Parametrize test persistence

samgermain:
Switches the tests in test_persistence to be parametrized

### DIFF
--- a/freqtrade/utils/get_sides.py
+++ b/freqtrade/utils/get_sides.py
@@ -1,5 +1,0 @@
-from typing import Tuple
-
-
-def get_sides(is_short: bool) -> Tuple[str, str]:
-    return ("sell", "buy") if is_short else ("buy", "sell")
--- a/freqtrade/utils/get_sides.py
+++ b/freqtrade/utils/get_sides.py
@@ -1,0 +1,5 @@
+from typing import Tuple
+
+
+def get_sides(is_short: bool) -> Tuple[str, str]:
+    return ("sell", "buy") if is_short else ("buy", "sell")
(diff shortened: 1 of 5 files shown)


freqtrade/utils/get_sides.py
```diff
@@ -0,0 +1,5 @@
+from typing import Tuple
+
+
+def get_sides(is_short: bool) -> Tuple[str, str]:
+    return ("sell", "buy") if is_short else ("buy", "sell")
```
xmatthias: I'm sceptical of generic "utils" packages/modules/... (or anything named similarily) ... as it'll end up as the "trashcan" for everything that can't go anywhere (i've been looking for that too very often, but usually finding a "properly named" place results in a better solution - or it'll end up in misc.py).

is your intend to use this also outside of tests in the future, or is it limited to tests?
If it's limited to tests (and has no benefit to the remainder of freqtrade) - then maybe moving it into the tests directory would make sense...

tests/test_persistence.py
```diff
@@ -709,7 +590,7 @@ def test_trade_close(limit_buy_order_usdt, limit_sell_order_usdt, fee):
     assert trade.close_date == new_date
 
 
-@pytest.mark.usefixtures("init_persistence")
+@ pytest.mark.usefixtures("init_persistence")
```
xmatthias: I'd not do that (and i'm not sure which stylesheet you've found this).

It looks odd - and is pretty dangerous when trying to find things - as it derives from the general style we're using elsewhere for decorators.

xmatthias (commented):
it's nitpicks, really ... that's parametrization at it's best i'd say :+1:

tests/test_persistence.py
samgermain: Sorry, it's some auto-formatting thing that only happens sometimes, and  I haven't figured out what makes it happen

freqtrade/utils/get_sides.py
samgermain: I moved it to conftest

tests/test_persistence.py
xmatthias: It's pretty strange, too - as it's not applied consistently (there's a ton of `@pytest` occurances in this same file which are not formatted this way).

xmatthias (approved):
LGTM :+1:

PR #5582, opened 2021-09-16
Added days parameter to PerformanceFilter

### LINKED ISSUE #5060: PerformanceFilter - last x mins
Hi for the first time and thank you for this magnificent piece of software :) 

I just discovered PerformanceFilter for the pairlists. I am excited about it and I would like to share some thoughts about possible improvements to this. 

Idea is to give possibility of taking into account performance from last x mins, hours, days. It can happen that one of the pairs had bad streak in first hour losing f.e. -3% and other pairs winning 0.1%. Having 30 pairs and trading 10 best of them it would mean that for this pair it's impossible to move to top 10 before all other pairs will prove to be worse. It can actually never happen and this pair could be the best in 2nd hour.

sergiikhliustin:
## Summary

Sometimes I want to use PerformanceFilter only according to performance of last N days

## Quick changelog

- added 'minutes' param to PerformanceFilter config
- if minutes > 0 getting performance only of last N days

closes #5060.

### DIFF
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -832,43 +832,21 @@ class Trade(_DECL_BASE, LocalTrade):
         return total_open_stake_amount or 0
 
     @staticmethod
-    def get_overall_performance() -> List[Dict[str, Any]]:
+    def get_overall_performance(days=None) -> List[Dict[str, Any]]:
         """
         Returns List of dicts containing all Trades, including profit and trade count
         NOTE: Not supported in Backtesting.
         """
+        filters = [Trade.is_open.is_(False)]
+        if days:
+            start_date = datetime.today() - timedelta(days)
+            filters.append((Trade.close_date >= start_date))
         pair_rates = Trade.query.with_entities(
             Trade.pair,
             func.sum(Trade.close_profit).label('profit_sum'),
             func.sum(Trade.close_profit_abs).label('profit_sum_abs'),
             func.count(Trade.pair).label('count')
-        ).filter(Trade.is_open.is_(False))\
-            .group_by(Trade.pair) \
-            .order_by(desc('profit_sum_abs')) \
-            .all()
-        return [
-            {
-                'pair': pair,
-                'profit': profit,
-                'profit_abs': profit_abs,
-                'count': count
-            }
-            for pair, profit, profit_abs, count in pair_rates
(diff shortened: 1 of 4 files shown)


freqtrade/persistence/models.py
```diff
@@ -856,6 +856,32 @@ def get_overall_performance() -> List[Dict[str, Any]]:
...
+            .order_by(desc('profit_sum_abs')) \
+            .all()
+        return [
+            {
+                'pair': pair,
+                'profit': profit,
+                'profit_abs': profit_abs,
+                'count': count
+            }
+            for pair, profit, profit_abs, count in pair_rates
```
xmatthias: I would think that a better way to do this is by introducing an (optional) argument to the existing method - which (if not None) adds a filter.

the `.filter()` method also accepts an (exploded) list - so you should be able to use something like the below:

``` python
filters = [Trade.is_open.is_(False)]
if days:
    start_date = datetime.today() - timedelta(days)
    filters.append((Trade.close_date >= start_date))

#....

.filter(*filters)\. <...>
```

sergiikhliustin: Updated according to your suggestion

freqtrade/persistence/models.py
```diff
@@ -832,17 +832,21 @@ def total_open_trades_stakes() -> float:
...
+        if days:
+            start_date = datetime.today() - timedelta(days)
+            filters.append((Trade.close_date >= start_date))
         pair_rates = Trade.query.with_entities(
             Trade.pair,
             func.sum(Trade.close_profit).label('profit_sum'),
             func.sum(Trade.close_profit_abs).label('profit_sum_abs'),
             func.count(Trade.pair).label('count')
-        ).filter(Trade.is_open.is_(False))\
+        ).filter(filters)\
```
xmatthias:
```suggestion
        ).filter(*filters)\
```

xmatthias (approved):
Great job :+1: 

I've added a few commits on to this, which changes the lookback to minutes.
This will give greater flexibility on shorter timeframes (performance over the last 3 days is not relevant if you run 1m timeframe - then the last hour (or 2 hours) will matter. it'll also close #5060 with this (but still alow lookbacks of multiple days).

I've also added  Documentation and a test which explicitly checks this behaviour.

PR #5584, opened 2021-09-16
Cache Intermediate Trade Files to Improve the Download Experience

mdashti:
## Summary

This PR caches the intermediate trade files taken from the exchange API calls to avoid re-doing them later.

## Quick changelog

- Creates a directory of temporary files under `user_data/data/binance/trades-intermediate-parts`. This folder can be removed after a complete download if you don't plan to have incremental downloads later. If your data download gets interrupted (e.g., due to network connectivity issues or a problem with the exchange), these temporary files avoid re-downloading the trades again.

## What's new?

This PR improves the download experience of users by storing the extracted data from the exchange API calls in temporary files. In addition, it creates an interval tree for each pair and stores the intervals that we have data downloaded for them. This way, the incremental data download experience also improves, as the future overlapping requests will also use the cached trade files if possible.

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1421,11 +1421,11 @@ class Exchange:
         cache_interval_tree = IntervalTree()
         tmpdata_dir = self._intermediate_trades_dir_for_pair(datadir, pair)
         if os.path.isdir(tmpdata_dir):
-            tmpdata_subdirs = os.listdir(tmpdata_dir)
+            tmpdata_subdirs = sorted(os.listdir(tmpdata_dir))
             for tmpdata_subdir in tmpdata_subdirs:
                 if tmpdata_subdir.isnumeric():
                     tmpdata_subdir = os.path.join(tmpdata_dir, tmpdata_subdir)
-                    tmpdata_files = os.listdir(tmpdata_subdir)
+                    tmpdata_files = sorted(os.listdir(tmpdata_subdir))
                     for tmpdata_file in tmpdata_files:
                         if tmpdata_file.startswith(self._pair_dir(pair)):
                             tmpdata_file = os.path.join(tmpdata_subdir, tmpdata_file)
@@ -1433,10 +1433,10 @@ class Exchange:
                                 trades_list = await self._async_fetch_trades_from_file(tmpdata_file)
                                 trades_list_from_id = trades_list[0][1]
                                 trades_list_to_id = trades_list[-1][1]
-                                cache_interval_tree.addi(
-                                    int(trades_list_from_id),
-                                    int(trades_list_to_id)+1,
-                                    tmpdata_file)
+                                cache_interval_tree.add(Interval(
+                                    int(trades_list_from_id), #inclusive
+                                    int(trades_list_to_id), #exclusive
+                                    tmpdata_file))
 
         logger.debug("Cached intervals for pair %s: %s intervals", pair, len(cache_interval_tree))
         self._intermediate_data_cache[pair] = cache_interval_tree
@@ -1523,6 +1523,8 @@ class Exchange:
                 # intermediate trade file.
                 tmpdata_file = self._intermediate_trades_file(datadir, pair, from_id)
                 json_string = json.dumps(trades_list)
+                self._intermediate_data_cache[pair].addi(int(from_id), int(to_id), tmpdata_file)
+
                 with open(tmpdata_file, "w") as text_file:
                     text_file.write(json_string)
                     logger.debug("Cached the intermediate trades in %s", tmpdata_file)
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1379,7 +1379,7 @@ class Exchange:
 
     # Directory name for a pair, e.g., given `ETH/BTC` returns `ETH-BTC`.
     def _pair_dir(self, pair: str) -> str:
-        return pair.replace("/","-")
+        return pair.replace("/", "-")
 
     # Returns the directory path that contains the intermediate trade files for a given pair.
     def _intermediate_trades_dir_for_pair(self, datadir: str, pair: str) -> str:
@@ -1414,7 +1414,7 @@ class Exchange:
     async def _get_interval_tree_for_pair(self, datadir: str, pair: str) -> IntervalTree:
         cached_res = self._intermediate_data_cache.get(pair, None)
         if cached_res:
-            return cached_res;
+            return cached_res
 
         logger.debug("Caching intervals for pair %s", pair)
         cache_interval_tree = IntervalTree()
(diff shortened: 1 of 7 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -1406,10 +1549,14 @@ async def _async_fetch_trades(self, pair: str,
         except ccxt.BaseError as e:
             raise OperationalException(f'Could not fetch trade data. Msg: {e}') from e
 
+    def batch_size(self) -> int:
+        return 1000
```
xmatthias, line 1591: Is this a constant (per exchange) that defines how many trades we expect per trade?

If yes, then please move it to `_ft_has` (per exchange?).
`ohlcv_candle_limit` does basically the same but for candle data.

freqtrade/exchange/exchange.py
```diff
@@ -1428,14 +1575,38 @@ async def _async_get_trade_history_id(self, pair: str,
...
         while True:
-            t = await self._async_fetch_trades(pair,
-                                               params={self._trades_pagination_arg: from_id})
+            assert from_id is not None
+
+            t = []
+            success_cache_read = False
+            if datadir:
+                tmpdata_file = self._intermediate_trades_file(datadir, pair, from_id)
+                if os.path.isfile(tmpdata_file):
```
xmatthias, line 1631: Please use `Pathlib` for directory operations, not `os.path`. 

I've worked hard about a year ago to get rid of using os.path, as using `os.path` can be problematic when using different environments (windows, linux).
While this usage is not problematic per se, you can make this clearer by having `tmpdata_file` be a Path object (out of the above method).

freqtrade/exchange/exchange.py
```diff
@@ -1368,12 +1377,87 @@ async def _async_get_candle_history(self, pair: str, timeframe: str,
...
+            return cached_res
+
+        logger.debug("Caching intervals for pair %s", pair)
+        cache_interval_tree = IntervalTree()
+        tmpdata_dir = self._intermediate_trades_dir_for_pair(datadir, pair)
+        if os.path.isdir(tmpdata_dir):
+            tmpdata_subdirs = sorted(os.listdir(tmpdata_dir))
+            for tmpdata_subdir in tmpdata_subdirs:
+                if tmpdata_subdir.isnumeric():
+                    tmpdata_subdir = os.path.join(tmpdata_dir, tmpdata_subdir)
```
xmatthias: As mentioned below - using Pathlib (`Path`) objects here will make this a lot simpler (and safer) to write

freqtrade/exchange/exchange.py
```diff
@@ -1368,12 +1377,87 @@ async def _async_get_candle_history(self, pair: str, timeframe: str,
...
+
+    # Builds the interval tree (from the intermediate trade files) for a given pair
+    @retrier_async
+    async def _get_interval_tree_for_pair(self, datadir: Path, pair: str) -> IntervalTree:
+        cached_res = self._intermediate_data_cache.get(pair, None)
+        if cached_res:
+            return cached_res
+
+        logger.debug("Caching intervals for pair %s", pair)
+        cache_interval_tree = IntervalTree()
```
xmatthias, line 1454: Can you maybe explain why we need this dependency?
It's unclear to me what this tries to accomplish / what benefit it adds.

It's an additional dependency, which to me seems pretty much unmaintained (last commit several months ago, no answers to issues, last release over a year ago)?

freqtrade/exchange/exchange.py
```diff
@@ -1368,12 +1377,87 @@ async def _async_get_candle_history(self, pair: str, timeframe: str,
...
+    @retrier_async
+    async def _async_fetch_trades_from_file(self, datafile: str) -> List[List]:
+        # Open a file: file
+        file = open(datafile, mode='r')
+
+        # read all lines at once
+        json_string = file.read()
+
+        # close the file
+        file.close()
```
xmatthias, line 1414: Please use:
* Pathlib for this
* context managers (which make sure the file is closed again).

---

Apart from that - i think this takes the assumption that the data is stored as json.
While that's the default for OHLCV - it's not the default for trades (which are stored as json.gz and can optionally be stored as hdf5 (which i would recommend).
We have a whole module taking care of this ... so directly accessing files is probably the wrong approach.

freqtrade/exchange/exchange.py
```diff
@@ -1368,12 +1377,87 @@ async def _async_get_candle_history(self, pair: str, timeframe: str,
...
+                if tmpdata_subdir.isnumeric():
+                    tmpdata_subdir = os.path.join(tmpdata_dir, tmpdata_subdir)
+                    tmpdata_files = sorted(os.listdir(tmpdata_subdir))
+                    for tmpdata_file in tmpdata_files:
+                        if tmpdata_file.startswith(self._pair_dir(pair)):
+                            tmpdata_file = os.path.join(tmpdata_subdir, tmpdata_file)
+                            if os.path.isfile(tmpdata_file):
+                                trades_list = await self._async_fetch_trades_from_file(tmpdata_file)
+                                trades_list_from_id = trades_list[0][1]
+                                trades_list_to_id = trades_list[-1][1]
```
xmatthias: I think this takes an assumption that will not hold depending on the exchange you're using this against.

While trade-id's for some exchanges are numeric and sequential - it's not the case for all exchanges (especially for exchanges that rely on time for iteration). 
As such, you can't assume that all id's between the first and last entry are present - as it could also be UUID's (which are completely random).

xmatthias (changes requested):
The main (and really only) purpose of having trade download (instead of ohlcv download) is kraken - which is not giving any ID's except the for the last trade in the response.
All other exchanges i know provide all available history via OHLCV endpoints - which means these should be used as they produce a lot less data (it's 1 datapoint per minute, trades data can be 1000nds per minute).
Downloading trade-data for Binance (for example) does not make any sense other than wasting time.

The only case where i've seen people download trades from binance is caused by missconfiguration (using the kraken example file and changing the exchange to binance).

Because of this, i don't think your "intervalCaching" will work - as kraken's trade endpoint does not give you trade-id's for all trades (only for the last trade).

mdashti:
Thanks for the comments @xmatthias. In the meanwhile, I added the support for time-based exchanges (e.g., kraken) in ddba69a and 462fc78. I'll get back to your comments in a bit. BTW, this all works for sequential IDs and timestamps. I need to figure out something for IDs that don't have any order (e.g., UUIDs). One approach is to exclude those exchanges. The current approach works well for kraken and binance.

xmatthias:
Are you planning on finishing this PR? otherwise i think we'll close it, as i think it's still missing quite some things to be ok ... (some of which are pointed out above, others which might show up in latter Reviews)

I'd also want to consider this as "during run only" - so once a download completes successfully, i'd expect this to clean up behind itself - as freqtrade does keep the history already and doesn't redownload it (if used correctly).

xmatthias:
Closing due to inactivity. Feel free to reopen should you be interested in finishing the PR.

PR #5586, opened 2021-09-17
changes some tests in test_freqtradebot to use usdt values

samgermain:
Switches some tests in test_freqtradebot to use usdt values instead of btc


samgermain:
Currently I have a few tests failing all with similar errors, either this is wrong assert freqtrade.handle_trade(trade) is False meaning it's true or after calling handle_trade I try to assert the caplog has something, but it has nothing

samgermain:
Will reopen when it's more done

PR #5587, opened 2021-09-17
Update prepare_trials_columns() return type

raphant:
## Summary
HyperoptTools.prepare_trials_columns() was set to wrong return type

### Quick Changes
Imported pandas in `hyperopt_tools.py` and set `HyperoptTools.prepare_trials_columns()` return type to `pd.DataFrame`

### DIFF
--- a/freqtrade/optimize/hyperopt_tools.py
+++ b/freqtrade/optimize/hyperopt_tools.py
@@ -299,7 +299,7 @@ class HyperoptTools():
                 f"Objective: {results['loss']:.5f}")
 
     @staticmethod
-    def prepare_trials_columns(trials, legacy_mode: bool, has_drawdown: bool) -> pd.DataFrame:
+    def prepare_trials_columns(trials: pd.DataFrame, legacy_mode: bool, has_drawdown: bool) -> pd.DataFrame:
 
         trials['Best'] = ''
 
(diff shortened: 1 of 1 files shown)


freqtrade/optimize/hyperopt_tools.py
```diff
@@ -298,7 +299,7 @@ def _format_explanation_string(results, total_epochs) -> str:
                 f"Objective: {results['loss']:.5f}")
 
     @staticmethod
-    def prepare_trials_columns(trials, legacy_mode: bool, has_drawdown: bool) -> str:
+    def prepare_trials_columns(trials, legacy_mode: bool, has_drawdown: bool) -> pd.DataFrame:
```
xmatthias: I think you should also type "trials" then ... to make it complete (that's why it wasn't captured as part of mypy in the first place).

raphant: Hmm, not sure I understand. Where would I type "trials"?

xmatthias: What i mean is type the parameter (as follows): 

`def prepare_trials_columns(trials: pd.DataFrame, ...`  

It's a completeness topic (but so is having the right return type) :D

raphant: Ah yes! Good point :)

xmatthias (approved):
LGTM :+1:

PR #5589, opened 2021-09-17
Include Raspberry Pi armv6 (0, 0W,...)

Ottavio97:
## Summary

Enable oldest Raspberry arch to setup Freqtrade

## Quick changelog

- Added detection of armv6l arch. It allow oldest Raspberry Pi to setup correctly

## Testing

- It has been tested correctly on Raspberry Pi Zero W

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -62,7 +62,7 @@ function updateenv() {
     then
         REQUIREMENTS_PLOT="-r requirements-plot.txt"
     fi
-    if [ "${SYS_ARCH}" == "armv7l" ]; then
+    if [ "${SYS_ARCH}" == "armv7l" ] || [ "${SYS_ARCH}" == "armv6l" ]; then
         echo "Detected Raspberry, installing cython, skipping hyperopt installation."
         ${PYTHON} -m pip install --upgrade cython
     else


xmatthias:
I do have very strong doubts that you can reliably run freqtrade on a pi zero, and would strongly advise against that.
it'll be late for buys due to the lower computing power, potentially costing you money - and more importantly, rendering a dry-run pointless as you won't know if it'll work similarily in a "cpu is fast enough" scenario.

xmatthias (approved):

Ottavio97:
@xmatthias thanks for the merge. I agree with the limited power of a Raspberry Pi Zero, but it seems to be not so dramatic.
In my scenario, I used my main PC (Gaming, so powerful enough) to do hyperoptimization on my custom strategy, and then moved my settings on a Pi Zero just to make trade (I actually ran it in dry mode, I don't know if live mode will take more resources).
The CPU is always 100%, and that is a sign that the Pi 0 maybe is not enough, but it is live since last week, testing successfully my strategy and checking trades with signal indicators (Example: I've used RSI/Bollinger combination, and the trade start time wasn't late compared to RSI/Bollinger signal).

Maybe a warning telling the user that is not recommended to run on such a low power CPU would be useful, but I won't exclude completely a Pi Zero usage. Just 'Use at your own risk' ;) 

Thanks anyway, and congratulations for this awesome project!

xmatthias:
It'll work for sure - but your buys / sells might be late.
Assuming you use a 5m strategy - you'd expect buys to happen at xx:00:05, xx:05:05, ... (giving it 5s to do the calculations).
In your case, you'll probably see buys anywhere between :05 and 1:00 (or even later) - which (depending on the strategy) can be pretty dangerous and cost you money in the end.

PR #5592, opened 2021-09-18
Test freqtradebot usdt

samgermain:
replaces limit_buy_order and limit_sell_order with limit_buy_order_usdt and limit_sell_order_usdt

Parametrizes some tests

I would double check these. I didn't completely understand every test, but I tried to swap with numbers that differed by equivalent percentages to the buy/sell price, and I felt that if the tests were working correctly before, then the new values produced by them should be the correct values

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,6 +25,8 @@ from freqtrade.resolvers import ExchangeResolver
 from freqtrade.worker import Worker
 from tests.conftest_trades import (mock_trade_1, mock_trade_2, mock_trade_3, mock_trade_4,
                                    mock_trade_5, mock_trade_6)
+from tests.conftest_trades_usdt import (mock_trade_usdt_1, mock_trade_usdt_2, mock_trade_usdt_3,
+                                        mock_trade_usdt_4, mock_trade_usdt_5, mock_trade_usdt_6)
 
 
 logging.getLogger('').setLevel(logging.INFO)
@@ -227,6 +229,39 @@ def create_mock_trades(fee, use_db: bool = True):
         Trade.query.session.flush()
 
 
+def create_mock_trades_usdt(fee, use_db: bool = True):
+    """
+    Create some fake trades ...
+    """
+    def add_trade(trade):
+        if use_db:
+            Trade.query.session.add(trade)
+        else:
+            LocalTrade.add_bt_trade(trade)
+
+    # Simulate dry_run entries
+    trade = mock_trade_usdt_1(fee)
+    add_trade(trade)
+
+    trade = mock_trade_usdt_2(fee)
+    add_trade(trade)
+
+    trade = mock_trade_usdt_3(fee)
+    add_trade(trade)
+
+    trade = mock_trade_usdt_4(fee)
+    add_trade(trade)
+
+    trade = mock_trade_usdt_5(fee)
+    add_trade(trade)
+
+    trade = mock_trade_usdt_6(fee)
+    add_trade(trade)
+
+    if use_db:
+        Trade.query.session.flush()
+
+
 @pytest.fixture(autouse=True)
 def patch_coingekko(mocker) -> None:
     """
@@ -303,7 +338,8 @@ def get_default_conf(testdatadir):
                 "ETH/BTC",
                 "LTC/BTC",
                 "XRP/BTC",
-                "NEO/BTC"
+                "NEO/BTC",
+                "ADA/USDT"
             ],
             "pair_blacklist": [
                 "DOGE/BTC",
@@ -373,6 +409,33 @@ def ticker_sell_down():
 
 
 @pytest.fixture
+def ticker_usdt():
+    return MagicMock(return_value={
+        'bid': 1.99,
+        'ask': 2.0,
+        'last': 1.99,
+    })
+
+
+@pytest.fixture
+def ticker_usdt_sell_up():
+    return MagicMock(return_value={
+        'bid': 2.19,
+        'ask': 2.2,
+        'last': 2.19,
+    })
+
+
+@pytest.fixture
+def ticker_usdt_sell_down():
+    return MagicMock(return_value={
+        'bid': 2.01,
+        'ask': 2.0,
+        'last': 2.01,
+    })
+
+
+@pytest.fixture
 def markets():
     return get_markets()
 
@@ -386,6 +449,31 @@ def get_markets():
             'symbol': 'ETH/BTC',
             'base': 'ETH',
             'quote': 'BTC',
+            'active': True,
+            'precision': {
+                'price': 8,
+                'amount': 8,
+                'cost': 8,
+            },
+            'lot': 0.00000001,
+            'limits': {
+                'amount': {
+                    'min': 0.01,
+                    'max': 1000,
+                },
+                'price': 500000,
+                'cost': {
+                    'min': 0.0001,
+                    'max': 500000,
+                },
+            },
+            'info': {},
+        },
+        'ADA/USDT': {
+            'id': 'ethbtc',
+            'symbol': 'ADA/USDT',
+            'base': 'USDT',
+            'quote': 'ADA',
             'active': True,
             'precision': {
                 'price': 8,
@@ -1816,6 +1904,22 @@ def open_trade():
         fee_open=0.0,
         fee_close=0.0,
         stake_amount=1,
+        open_date=arrow.utcnow().shift(minutes=-601).datetime,
+        is_open=True
+    )
+
+
+@pytest.fixture(scope="function")
+def open_trade_usdt():
+    return Trade(
+        pair='ADA/USDT',
+        open_rate=2.0,
+        exchange='binance',
+        open_order_id='123456789',
+        amount=30.0,
+        fee_open=0.0,
+        fee_close=0.0,
+        stake_amount=60.0,
         open_date=arrow.utcnow().shift(minutes=-601).datetime,
         is_open=True
     )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -289,9 +289,19 @@ def init_persistence(default_conf):
     init_db(default_conf['db_url'], default_conf['dry_run'])
 
 
+@pytest.fixture(scope='function')
+def init_persistence_usdt(default_conf_usdt):
+    init_db(default_conf_usdt['db_url'], default_conf_usdt['dry_run'])
+
+
 @pytest.fixture(scope="function")
 def default_conf(testdatadir):
     return get_default_conf(testdatadir)
+
+
+@pytest.fixture(scope="function")
+def default_conf_usdt(testdatadir):
+    return get_default_conf_usdt(testdatadir)
 
 
 def get_default_conf(testdatadir):
@@ -365,6 +375,15 @@ def get_default_conf(testdatadir):
         "internals": {},
         "export": "none",
     }
+    return configuration
+
+
+def get_default_conf_usdt(testdatadir):
+    configuration = get_default_conf(testdatadir)
+    configuration.update({
+        "stake_amount": 60.0,
+        "stake_currency": "USDT",
+    })
     return configuration
 
 
@@ -1602,7 +1621,7 @@ def trades_for_order():
         'info': {
             'id': 34567,
             'orderId': 123456,
-            'price': '0.24544100',
+            'price': '2.0',
             'qty': '8.00000000',
             'commission': '0.00800000',
             'commissionAsset': 'LTC',
@@ -1809,6 +1828,14 @@ def edge_conf(default_conf):
     }
 
     return conf
+
+
+@pytest.fixture(scope="function")
+def edge_conf_usdt(edge_conf):
+    edge_conf.update({
+        "stake_currency": "USDT",
+    })
+    return edge_conf
 
 
 @pytest.fixture
@@ -2049,7 +2076,7 @@ def saved_hyperopt_results():
 @pytest.fixture(scope='function')
 def limit_buy_order_usdt_open():
     return {
-        'id': 'mocked_limit_buy',
+        'id': 'mocked_limit_buy_usdt',
         'type': 'limit',
         'side': 'buy',
         'symbol': 'mocked',
@@ -2076,7 +2103,7 @@ def limit_buy_order_usdt(limit_buy_order_usdt_open):
 @pytest.fixture
 def limit_sell_order_usdt_open():
     return {
-        'id': 'mocked_limit_sell',
+        'id': 'mocked_limit_sell_usdt',
         'type': 'limit',
         'side': 'sell',
         'pair': 'mocked',
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1598,27 +1598,34 @@ def result(testdatadir):
 
 @pytest.fixture(scope="function")
 def trades_for_order():
-    return [{'info': {'id': 34567,
-                      'orderId': 123456,
-                      'price': '0.24544100',
-                      'qty': '8.00000000',
-                      'commission': '0.00800000',
-                      'commissionAsset': 'LTC',
-                      'time': 1521663363189,
-                      'isBuyer': True,
-                      'isMaker': False,
-                      'isBestMatch': True},
-             'timestamp': 1521663363189,
-             'datetime': '2018-03-21T20:16:03.189Z',
-             'symbol': 'LTC/ETH',
-             'id': '34567',
-             'order': '123456',
-             'type': None,
-             'side': 'buy',
-             'price': 0.245441,
-             'cost': 1.963528,
-             'amount': 8.0,
-             'fee': {'cost': 0.008, 'currency': 'LTC'}}]
+    return [{
+        'info': {
+            'id': 34567,
+            'orderId': 123456,
+            'price': '0.24544100',
+            'qty': '8.00000000',
+            'commission': '0.00800000',
+            'commissionAsset': 'LTC',
+            'time': 1521663363189,
+            'isBuyer': True,
+            'isMaker': False,
+            'isBestMatch': True
+        },
+        'timestamp': 1521663363189,
+        'datetime': '2018-03-21T20:16:03.189Z',
+        'symbol': 'LTC/USDT',
+        'id': '34567',
+        'order': '123456',
+        'type': None,
+        'side': 'buy',
+        'price': 2.0,
+        'cost': 16.0,
+        'amount': 8.0,
+        'fee': {
+            'cost': 0.008,
+            'currency': 'LTC'
+        }
+    }]
 
 
 @pytest.fixture(scope="function")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -441,7 +441,7 @@ def ticker_sell_down():
 def ticker_usdt():
     return MagicMock(return_value={
         'bid': 2.0,
-        'ask': 2.01,
+        'ask': 2.1,
         'last': 2.0,
     })
 
@@ -449,9 +449,9 @@ def ticker_usdt():
 @pytest.fixture
 def ticker_usdt_sell_up():
     return MagicMock(return_value={
-        'bid': 2.19,
-        'ask': 2.2,
-        'last': 2.19,
+        'bid': 2.2,
+        'ask': 2.3,
+        'last': 2.2,
     })
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -376,7 +376,7 @@ def get_default_conf(testdatadir):
 def get_default_conf_usdt(testdatadir):
     configuration = get_default_conf(testdatadir)
     configuration.update({
-        "stake_amount": 10.0,
+        "stake_amount": 60.0,
         "stake_currency": "USDT",
         "exchange": {
             "name": "binance",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -338,8 +338,7 @@ def get_default_conf(testdatadir):
                 "ETH/BTC",
                 "LTC/BTC",
                 "XRP/BTC",
-                "NEO/BTC",
-                "ADA/USDT"
+                "NEO/BTC"
             ],
             "pair_blacklist": [
                 "DOGE/BTC",
@@ -449,31 +448,6 @@ def get_markets():
             'symbol': 'ETH/BTC',
             'base': 'ETH',
             'quote': 'BTC',
-            'active': True,
-            'precision': {
-                'price': 8,
-                'amount': 8,
-                'cost': 8,
-            },
-            'lot': 0.00000001,
-            'limits': {
-                'amount': {
-                    'min': 0.01,
-                    'max': 1000,
-                },
-                'price': 500000,
-                'cost': {
-                    'min': 0.0001,
-                    'max': 500000,
-                },
-            },
-            'info': {},
-        },
-        'ADA/USDT': {
-            'id': 'ethbtc',
-            'symbol': 'ADA/USDT',
-            'base': 'USDT',
-            'quote': 'ADA',
             'active': True,
             'precision': {
                 'price': 8,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -378,6 +378,22 @@ def get_default_conf_usdt(testdatadir):
     configuration.update({
         "stake_amount": 10.0,
         "stake_currency": "USDT",
+        "exchange": {
+            "name": "binance",
+            "enabled": True,
+            "key": "key",
+            "secret": "secret",
+            "pair_whitelist": [
+                "ETH/USDT",
+                "LTC/USDT",
+                "XRP/USDT",
+                "NEO/USDT"
+            ],
+            "pair_blacklist": [
+                "DOGE/USDT",
+                "HOT/USDT",
+            ]
+        },
     })
     return configuration
 
@@ -424,9 +440,9 @@ def ticker_sell_down():
 @pytest.fixture
 def ticker_usdt():
     return MagicMock(return_value={
-        'bid': 1.99,
-        'ask': 2.0,
-        'last': 1.99,
+        'bid': 2.0,
+        'ask': 2.01,
+        'last': 2.0,
     })
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -387,7 +387,8 @@ def get_default_conf_usdt(testdatadir):
                 "ETH/USDT",
                 "LTC/USDT",
                 "XRP/USDT",
-                "NEO/USDT"
+                "NEO/USDT",
+                "TKN/USDT",
             ],
             "pair_blacklist": [
                 "DOGE/USDT",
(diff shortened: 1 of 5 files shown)


samgermain:
3 tests are currently failing. I'm not sure what the correct values are supposed to be in these and why the pair is `ETH/BTC`. I can't see where that is getting set

FAILED tests/test_freqtradebot.py::test_handle_stoploss_on_exchange_trailing
```
e = AssertionError(msg)
            e._mock_introspection_applied = True  # type:ignore[attr-defined]
>           raise e
E           AssertionError: expected call not found.
E           Expected: mock(amount=100.0, pair='ETH/BTC', order_types={'buy': 'limit', 'sell': 'limit', 'stoploss': 'limit', 'stoploss_on_exchange': True, 'stoploss_on_exchange_interval': 0}, stop_price=4.224)
E           Actual: mock(pair='ETH/BTC', amount=0.01166666, stop_price=4.224, order_types={'buy': 'limit', 'sell': 'limit', 'stoploss': 'limit', 'stoploss_on_exchange': True, 'stoploss_on_exchange_interval': 0})
E           
E           pytest introspection follows:
E           
E           Kwargs:
E           assert {'amount': 0....price': 4.224} == {'amount': 10...price': 4.224}
E             Omitting 3 identical items, use -vv to show
E             Differing items:
E             {'amount': 0.01166666} != {'amount': 100.0}
E             Use -v to get the full diff
```

FAILED tests/test_freqtradebot.py::test_handle_stoploss_on_exchange_custom_stop
```
.env/lib/python3.9/site-packages/pytest_mock/plugin.py:437: AssertionError

stoploss_order_mock.assert_called_once_with(
            amount=100.0,
            pair='ETH/BTC',
            order_types=freqtrade.strategy.order_types,
            stop_price=4.4 * 0.96
        )
E       AssertionError: expected call not found.
E       Expected: mock(amount=100.0, pair='ETH/BTC', order_types={'buy': 'limit', 'sell': 'limit', 'stoploss': 'limit', 'stoploss_on_exchange': True, 'stoploss_on_exchange_interval': 0}, stop_price=4.224)
E       Actual: mock(pair='ETH/BTC', amount=0.01166666, stop_price=4.224, order_types={'buy': 'limit', 'sell': 'limit', 'stoploss': 'limit', 'stoploss_on_exchange': True, 'stoploss_on_exchange_interval': 0})
E       
E       pytest introspection follows:
E       
E       Kwargs:
E       assert {'amount': 0....price': 4.224} == {'amount': 10...price': 4.224}
E         Omitting 3 identical items, use -vv to show
E         Differing items:
E         {'amount': 0.01166666} != {'amount': 100.0}
E         Use -v to get the full diff
```

FAILED tests/test_freqtradebot.py::test_tsl_on_exchange_compatible_with_edge
```
        # price goes down 5%
        mocker.patch('freqtrade.exchange.Exchange.fetch_ticker', MagicMock(return_value={
            'bid': 1.9 * 0.95,
            'ask': 2.2 * 0.95,
            'last': 1.9 * 0.95
        }))
    
>       assert freqtrade.handle_trade(trade) is False

tests/test_freqtradebot.py:1644: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
freqtrade/freqtradebot.py:717: in handle_trade
    if self._check_and_execute_exit(trade, exit_rate, buy, sell):
freqtrade/freqtradebot.py:866: in _check_and_execute_exit
    self.execute_trade_exit(trade, exit_rate, should_sell)
freqtrade/freqtradebot.py:1126: in execute_trade_exit
    order = self.exchange.create_order(pair=trade.pair,
/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1093: in __call__
    return self._mock_call(*args, **kwargs)
/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1097: in _mock_call
    return self._execute_mock_call(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <MagicMock id='5068770320'>, args = ()
kwargs = {'amount': 13.15657894, 'ordertype': 'limit', 'pair': 'NEO/BTC', 'rate': 2.178, ...}
effect = <list_iterator object at 0x12e2672b0>

        effect = self.side_effect
        if effect is not None:
            if _is_exception(effect):
                raise effect
            elif not _callable(effect):
E               StopIteration

/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1154: StopIteration
```

tests/conftest_trades_usdt.py
```diff
@@ -0,0 +1,305 @@
...
+        exchange='binance',
+        strategy='SampleStrategy',
+        open_order_id="prod_sell_6",
+        timeframe=5,
+    )
+    o = Order.parse_from_ccxt_object(mock_order_usdt_6(), 'LTC/USDT', 'buy')
+    trade.orders.append(o)
+    o = Order.parse_from_ccxt_object(mock_order_usdt_6_sell(), 'LTC/USDT', 'sell')
+    trade.orders.append(o)
+    return trade
```
samgermain, line 305: I didn't actually end up using these, but I left them in because I thought maybe they could be helpful. I can remove them from the PR also

samgermain:
Will parametrize some tests in another PR before making these changes

tests/test_freqtradebot.py
```diff
@@ -1207,6 +1210,7 @@ def test_handle_stoploss_on_exchange_trailing(mocker, default_conf, fee,
     patch_get_signal(freqtrade)
 
     freqtrade.enter_positions()
+    # TODO-lev: Get this trade switched to the usdt trades
```
samgermain: I'm not sure how to get this test to pass because I don't know how to set the trades inside `Trade.query`

Right now the old btc price trades are returned from `Trade.query`, and I would like to update them so that the usdt price trades are returned, but it's hard to locate where the program actually creates the mock trades

tests/test_freqtradebot.py
```diff
@@ -1693,28 +1719,31 @@ def test_update_trade_state_withorderdict(default_conf, trades_for_order, limit_
...
+        open_rate=2.0,
         open_date=arrow.utcnow().datetime,
         fee_open=fee.return_value,
         fee_close=fee.return_value,
         open_order_id="123456",
         is_open=True,
     )
-    freqtrade.update_trade_state(trade, '123456', limit_buy_order)
+    # TODO-lev: caplog.text has Amount 60.00000000000001 does not match amount 60.00000000000001
+    # TODO-lev: but they are the exact same
```
samgermain: @xmatthias  This part makes no sense, the caplog doesn't contain what it is supposed to, but if I run `pytest --pdb` and then read `caplog.text` when it breaks here, I see `Amount 60.00000000000001 does not match amount 60.00000000000001` and they are actually the exact same

tests/test_freqtradebot.py
xmatthias: whenever `enter_positions()` is called, it's not mocked trades, but "kindof real" trades.

To have them use different currencies, you'll have to change stake_currency in default_conf and use a different pairlist. Obviously the results (create_order, fetch_order, ...) need to align to this as well.
That'll however impact the whole testsuite - so you'll need to carefully consider if that's something you wanna do at this point (i'd try to avoid creating a second default_config).

The asset name that's used is not really important in my opinion (it'll be the first (few) of the pair_whitelist) - but the values should be "normal" values.

samgermain: > whenever `enter_positions()` is called, it's not mocked trades, but "kindof real" trades.
> 
> To have them use different currencies, you'll have to change stake_currency in default_conf and use a different pairlist. Obviously the results (create_order, fetch_order, ...) need to align to this as well.
> That'll however impact the whole testsuite - so you'll need to carefully consider if that's something you wanna do at this point (i'd try to avoid creating a second default_config).
> 
> The asset name that's used is not really important in my opinion (it'll be the first (few) of the pair_whitelist) - but the values should be "normal" values.

it's the prices and amounts that I want to switch, the amount is something like `0.017300332` or something, and I'd like to switch it to `30.0` or something like that

xmatthias: there's a whole range of settings playing together - it's not that different for tests than if you configure a real configuration.
if some parts don't align (stake_amount for example) - you'll get odd results.

samgermain: > there's a whole range of settings playing together - it's not that different for tests than if you configure a real configuration.
> if some parts don't align (stake_amount for example) - you'll get odd results.

Im just wondering where these trades are created, I've tried to backtrack through the function calls, but can't find it

xmatthias: enter_positions() :)

it'll create the trades as part of it's functionality - based on the strategy / exchange (moked) responses.

samgermain: I'm really confused where the value `0.01105263` is coming from, that is the correct value that amount is equal to for `stoploss_order_mock.assert_called_once_with`, what can i mess with to change this value

tests/test_freqtradebot.py
```diff
@@ -1533,37 +1558,36 @@ def test_tsl_on_exchange_compatible_with_edge(mocker, edge_conf, fee, caplog,
...
     mocker.patch('freqtrade.exchange.Exchange.fetch_ticker', MagicMock(return_value={
-        'bid': 0.00001172 * 0.95,
-        'ask': 0.00001173 * 0.95,
-        'last': 0.00001172 * 0.95
+        'bid': 1.9 * 0.95,
+        'ask': 2.2 * 0.95,
+        'last': 1.9 * 0.95
     }))
-
     assert freqtrade.handle_trade(trade) is False
```
samgermain, line 1562: This test is failing with 

```
>       assert freqtrade.handle_trade(trade) is False

tests/test_freqtradebot.py:1565: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
freqtrade/freqtradebot.py:717: in handle_trade
    if self._check_and_execute_exit(trade, exit_rate, buy, sell):
freqtrade/freqtradebot.py:866: in _check_and_execute_exit
    self.execute_trade_exit(trade, exit_rate, should_sell)
freqtrade/freqtradebot.py:1126: in execute_trade_exit
    order = self.exchange.create_order(pair=trade.pair,
/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1093: in __call__
    return self._mock_call(*args, **kwargs)
/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1097: in _mock_call
    return self._execute_mock_call(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <MagicMock id='4940277168'>, args = ()
kwargs = {'amount': 13.15657894, 'ordertype': 'limit', 'pair': 'NEO/BTC', 'rate': 2.178, ...}
effect = <list_iterator object at 0x1269a07f0>

        effect = self.side_effect
        if effect is not None:
            if _is_exception(effect):
                raise effect
            elif not _callable(effect):
E               StopIteration

/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1154: StopIteration
```

Not really sure what this means or how to fix it

tests/conftest.py
```diff
@@ -287,11 +287,21 @@ def init_persistence(default_conf):
     init_db(default_conf['db_url'], default_conf['dry_run'])
 
 
+@pytest.fixture(scope='function')
+def init_persistence_usdt(default_conf_usdt):
+    init_db(default_conf_usdt['db_url'], default_conf_usdt['dry_run'])
```
xmatthias: this is pretty pointless ... it's a fixture that doesn't use anything you're changing (it's only using db_url and dry_run) ... so you can safely use the original fixture.

tests/test_freqtradebot.py
xmatthias: Your mock was exhausted (which can only happen when using a list as part of `side_effect` to the mock).
probably, now there's more trades (for whatever reason) - maybe you've now got more money left (different stake_amount/wallet ratio), so more trades can be filled...

tests/conftest.py
```diff
@@ -1735,6 +1815,14 @@ def edge_conf(default_conf):
     return conf
 
 
+@pytest.fixture(scope="function")
+def edge_conf_usdt(edge_conf):
+    edge_conf.update({
+        "stake_currency": "USDT",
+    })
+    return edge_conf
```
xmatthias: IF you decide to touch edge-tests - then i'd swap then all at once. 

While it seems to be used often at first glance, i don't think it will be too difficult to swap them all in one go.
if the outcome of this "convert tests to usdt" is that we have most of the setup-code duplicated, then it'll not bring many benefits ...

tests/test_freqtradebot.py
```diff
@@ -238,143 +238,143 @@ def test_edge_overrides_stoploss(limit_buy_order, fee, caplog, mocker,
...
     trade = Trade.query.order_by(Trade.id.desc()).first()
 
     assert trade is not None
-    assert trade.stake_amount == 0.00098751
+    assert trade.stake_amount == 10.0
     assert trade.is_open
     assert trade.open_date is not None
 
-    assert Trade.total_open_trades_stakes() == 1.97502e-03
+    assert Trade.total_open_trades_stakes() == 20.0
```
samgermain: For some reason `Trade.total_open_trades_stakes()` evaluates to 10.0, and when I check `Trade.query.all()` I get an array with only one trade in it `[Trade(id=1, pair=ETH/USDT, amount=5.00000000, open_rate=2.00000000, open_since=2021-09-26 20:10:16)]`

tests/test_freqtradebot.py
```diff
@@ -2691,16 +2709,16 @@ def test_execute_trade_exit_custom_exit_price(default_conf, ticker, fee, ticker_
         'trade_id': 1,
         'type': RPCMessageType.SELL,
         'exchange': 'Binance',
-        'pair': 'ETH/BTC',
+        'pair': 'ETH/USDT',
         'gain': 'profit',
-        'limit': 1.170e-05,
-        'amount': 91.07468123,
+        'limit': 2.25,
```
samgermain, line 2699: pdb is telling me `{'limit': 2.244}` and I'm confused why it's not `2.25` considering that's the value that I set in `    freqtrade.strategy.custom_exit_price = lambda **kwargs: 2.25`

tests/test_freqtradebot.py
```diff
@@ -416,8 +416,8 @@ def test_enter_positions_global_pairlock(default_conf, ticker, limit_buy_order,
     assert log_has_re(message, caplog)
```
samgermain, line 416: My caplog has `No currency pair in active pair whitelist, but checking to sell open trades.` and I can't see where I could set the whitelist for this test

tests/test_freqtradebot.py
samgermain: I'm guessing that this is because I need to add data to `tests/testdata`, I don't really know what's the best way to do this

samgermain:
2 tests fail with nearly the same message/reason `test_api_blacklist` and `test_rpc_blacklist`. The first time they fail is on this commit 85f8e074c677894222be643368572518a2725e77

```
_________________________________ test_api_blacklist __________________________________

botclient = (<freqtrade.freqtradebot.FreqtradeBot object at 0x12db9cb50>, <starlette.testclient.TestClient object at 0x12b8e04c0>)
mocker = <pytest_mock.plugin.MockerFixture object at 0x12cee3ee0>

    def test_api_blacklist(botclient, mocker):
        ftbot, client = botclient
    
        rc = client_get(client, f"{BASE_URI}/blacklist")
        assert_response(rc)
        # DOGE and HOT are not in the markets mock!
        assert rc.json() == {"blacklist": ["DOGE/BTC", "HOT/BTC"],
                             "blacklist_expanded": [],
                             "length": 2,
                             "method": ["StaticPairList"],
                             "errors": {},
                             }
    
        # Add ETH/BTC to blacklist
        rc = client_post(client, f"{BASE_URI}/blacklist",
                         data='{"blacklist": ["ETH/BTC"]}')
        assert_response(rc)
        assert rc.json() == {"blacklist": ["DOGE/BTC", "HOT/BTC", "ETH/BTC"],
                             "blacklist_expanded": ["ETH/BTC"],
                             "length": 3,
                             "method": ["StaticPairList"],
                             "errors": {},
                             }
    
        rc = client_post(client, f"{BASE_URI}/blacklist",
                         data='{"blacklist": ["XRP/.*"]}')
        assert_response(rc)
>       assert rc.json() == {"blacklist": ["DOGE/BTC", "HOT/BTC", "ETH/BTC", "XRP/.*"],
                             "blacklist_expanded": ["ETH/BTC", "XRP/BTC"],
                             "length": 4,
                             "method": ["StaticPairList"],
                             "errors": {},
                             }
E       AssertionError: assert {'blacklist':...ngth': 4, ...} == {'blacklist':...ngth': 4, ...}
E         Omitting 4 identical items, use -vv to show
E         Differing items:
E         {'blacklist_expanded': ['ETH/BTC', 'XRP/BTC', 'XRP/USDT']} != {'blacklist_expanded': ['ETH/BTC', 'XRP/BTC']}
E         Use -v to get the full diff

tests/rpc/test_rpc_apiserver.py:939: AssertionError
```

I think it is because of this line

```
    rc = client_post(client, f"{BASE_URI}/blacklist",
                     data='{"blacklist": ["XRP/.*"]}')
 ```
 
 and the fact that both `XRP/USDT` and `XRP/BTC` are both in `get_markets` now
 
 ----------------------------
 
 After these tests pass, this PR can be moved to in review, but I would highly recommend double checking the new values in these tests

**Comment by xmatthias** [comment]
I'm not sure i see where this test is problematic - you changed available markets around - so it's pretty obvious that the expansion will include this additional pair - so simply adding the new pair to the "expected" field should suffice.


You should also look into the merge conflict - it's preventing CI from running (and also from merging ...).
Obviously, we'll have to carefully look at new values.

File: tests/test_freqtradebot.py [hunk 10]
```diff
@@ -238,143 +240,143 @@ def test_edge_overrides_stoploss(limit_buy_order, fee, caplog, mocker,
...
     assert trade.exchange == 'binance'
 
     # Simulate fulfilled LIMIT_BUY order for trade
-    trade.update(limit_buy_order)
+    trade.update(limit_buy_order_usdt)
 
-    assert trade.open_rate == 0.00001099
-    assert trade.amount == 90.99181073
+    assert trade.open_rate == 2.0
+    assert trade.amount == 30.0
```
xmatthias, line 301: Something seems odd here.
`0.001 / 0.00001099 = 90.99181073` - but `10 / 2 != 30` ... 

Might be the buy-order that's misalligned to this ... ?

tests/test_freqtradebot.py
```diff
@@ -766,15 +773,15 @@ def test_execute_entry(mocker, default_conf, fee, limit_buy_order, limit_buy_ord
...
-    limit_buy_order['status'] = 'expired'
-    limit_buy_order['amount'] = 90.99181073
-    limit_buy_order['filled'] = 80.99181073
-    limit_buy_order['remaining'] = 10.00
-    limit_buy_order['price'] = 0.5
-    limit_buy_order['cost'] = 40.495905365
-    limit_buy_order['id'] = '555'
+    limit_buy_order_usdt['status'] = 'expired'
+    limit_buy_order_usdt['amount'] = 30.0
+    limit_buy_order_usdt['filled'] = 80.99181073
```
xmatthias: This seems odd.
We have an order for 30 Coins - of which 80.99 filled ... which seems rather unlikely.

tests/test_freqtradebot.py
```diff
@@ -2574,16 +2602,17 @@ def test_execute_trade_exit_up(default_conf, ticker, fee, ticker_sell_up, mocker
...
+        'amount': 5.0,
         'order_type': 'limit',
-        'open_rate': 1.098e-05,
-        'current_rate': 1.173e-05,
-        'profit_amount': 6.223e-05,
-        'profit_ratio': 0.0620716,
-        'stake_currency': 'BTC',
+        'open_rate': 2.0,
+        'current_rate': 2.3,
+        # TODO: Double check that profit_amount and profit_ratio are correct
```
xmatthias: ...

tests/test_freqtradebot.py
```diff
@@ -2614,28 +2644,29 @@ def test_execute_trade_exit_down(default_conf, ticker, fee, ticker_sell_down, mo
...
+        fetch_ticker=ticker_usdt_sell_down
     )
 
-    freqtrade.execute_trade_exit(trade=trade, limit=ticker_sell_down()['bid'],
+    freqtrade.execute_trade_exit(trade=trade, limit=ticker_usdt_sell_down()['bid'],
                                  sell_reason=SellCheckTuple(sell_type=SellType.STOP_LOSS))
 
     assert rpc_mock.call_count == 2
     last_msg = rpc_mock.call_args_list[-1][0][0]
+    # TODO: Should be a loss, but comes out as a gain
```
xmatthias: Values seem to be a loss (?) is this TODO outdated?

tests/test_freqtradebot.py
```diff
@@ -2732,34 +2765,35 @@ def test_execute_trade_exit_down_stoploss_on_exchange_dry_run(default_conf, tick
...
-    trade.stop_loss = 0.00001099 * 0.99
-    freqtrade.execute_trade_exit(trade=trade, limit=ticker_sell_down()['bid'],
+    trade.stop_loss = 2.0 * 0.99
+    freqtrade.execute_trade_exit(trade=trade, limit=ticker_usdt_sell_down()['bid'],
                                  sell_reason=SellCheckTuple(sell_type=SellType.STOP_LOSS))
 
     assert rpc_mock.call_count == 2
     last_msg = rpc_mock.call_args_list[-1][0][0]
 
+    # TODO: Are these values correct?
```
xmatthias: ..

tests/test_freqtradebot.py
```diff
@@ -2951,32 +2985,33 @@ def test_execute_trade_exit_market_order(default_conf, ticker, fee,
...
+    freqtrade.execute_trade_exit(trade=trade, limit=ticker_usdt_sell_up()['bid'],
                                  sell_reason=SellCheckTuple(sell_type=SellType.ROI))
 
     assert not trade.is_open
-    assert trade.close_profit == 0.0620716
+    assert trade.close_profit == 0.09451372  # TODO: Check this is correct
 
     assert rpc_mock.call_count == 3
     last_msg = rpc_mock.call_args_list[-1][0][0]
+    # TODO: Is this correct?
```
xmatthias: ...

tests/test_freqtradebot.py
```diff
@@ -3010,18 +3045,18 @@ def test_execute_trade_exit_insufficient_funds_error(default_conf, ticker, fee,
...
                                             sell_reason=sell_reason)
     assert mock_insuf.call_count == 1
 
 
 @pytest.mark.parametrize('profit_only,bid,ask,handle_first,handle_second,sell_type', [
     # Enable profit
-    (True, 0.00001172, 0.00001173, False, True, SellType.SELL_SIGNAL.value),
+    (True, 1.9, 2.2, False, True, SellType.SELL_SIGNAL.value),
     # Disable profit
     (False, 0.00002172, 0.00002173, True,  False, SellType.SELL_SIGNAL.value),
```
xmatthias: I'd think that these values will also need to be changed to "USDT like" values ...

tests/test_freqtradebot.py
```diff
@@ -3219,249 +3253,160 @@ def test_ignore_roi_if_buy_signal(default_conf, limit_buy_order, limit_buy_order
...
+            'ask': buy_price + 0.06,
+            'last': buy_price + 0.06
+        })
+    )
     # stop-loss not reached, adjusted stoploss
     assert freqtrade.handle_trade(trade) is False
-    assert log_has("ETH/BTC - Using positive stoploss: 0.01 offset: 0 profit: 0.2666%", caplog)
-    assert log_has("ETH/BTC - Adjusting stoploss...", caplog)
-    assert trade.stop_loss == 0.0000138501
+    # TODO: is 0.0249% correct? Shouldn't it be higher?
```
xmatthias: ...

tests/test_freqtradebot.py
```diff
@@ -1606,7 +1631,7 @@ def test_exit_positions(mocker, default_conf, limit_buy_order, caplog) -> None:
     assert n == 0
     # Test amount not modified by fee-logic
```
xmatthias: What is this test testing? get_real_amount is not relevant for this test (as handle_trade is mocked).

tests/test_freqtradebot.py
```diff
@@ -3219,249 +3253,160 @@ def test_ignore_roi_if_buy_signal(default_conf, limit_buy_order, limit_buy_order
...
+                     'ask': 2.0 * 1.1,
+                     'last': 2.0 * 1.1
                  }))
 
     caplog.set_level(logging.DEBUG)
     # Sell as trailing-stop is reached
     assert freqtrade.handle_trade(trade) is True
-    assert log_has("ETH/BTC - HIT STOP: current price at 0.000012, stoploss is 0.000015, "
-                   "initial stoploss was at 0.000010, trade opened at 0.000011", caplog)
+    # TODO: Does this make sense? How is stoploss 2.7?
```
xmatthias: 2 * 1.5 = 3.0 - so a trailing stoploss of 2.7 seems realistic

xmatthias (commented):
That's quite some text to review :laughing: 

I think it's 95% there ... i've left a few comments ("..." is the TODO's - so they're easier to spot in the 1000+ diff.
I've also removed some unused parameters i stumbled across (if we change this much, might as well fix that as well ...)

samgermain:
> I've also removed some unused parameters i stumbled across (if we change this much, might as well fix that as well ...)

Do yo still have that commit locally? I might have force pushed over it, sorry

tests/test_freqtradebot.py
samgermain: Do you mean just remove this? I removed all the ... TODOs on a single isolated commit, so I can revert it if I want

tests/test_freqtradebot.py
samgermain: i have no idea whats going on here, this is one of the tests where I tried my best to update it to new values, but didn't understand the test or what the values should be

tests/test_freqtradebot.py
xmatthias: My "..." comments were to allow me to easily identify these points :laughing:

tests/test_freqtradebot.py
xmatthias: The problem here is the following:
the values passed in (stake_amount=10,  rate=2) would mean we'll be getting 5 ETH.
`limit_buy_order_usdt` however defines this order as `price=2.0, amount=30, cost=60` - assuming a stake_amount of 60$.

`trade.update()` will take the values from the order and update things.
real-world, this would happen for example on partially filled orders, or market orders where the price is unknown.
Now as the order does not align with the config (order defines a stake_amount of 60$, config of 10$) - we get strange values here.

I think an easy fix is to up stake_amount to 60 - using 10$ as stake can become a problem anyway, as realistically, you can't place orders for 10$ if you still wanna leave room for a stoploss below.

samgermain: > The problem here is the following: the values passed in (stake_amount=10, rate=2) would mean we'll be getting 5 ETH. `limit_buy_order_usdt` however defines this order as `price=2.0, amount=30, cost=60` - assuming a stake_amount of 60$.
> 
> `trade.update()` will take the values from the order and update things. real-world, this would happen for example on partially filled orders, or market orders where the price is unknown. Now as the order does not align with the config (order defines a stake_amount of 60$, config of 10$) - we get strange values here.
> 
> I think an easy fix is to up stake_amount to 60 - using 10$ as stake can become a problem anyway, as realistically, you can't place orders for 10$ if you still wanna leave room for a stoploss below.

Sorry, I understood why it was a problem, but didn't understand why the test was passing

xmatthias: The test is passing because we're checking the values that come out (30) - which does not correspond to the expeced result (due to the misalignment between order and stake-amount).

i've now updated stake-amount to 60$ - so it's aligned to the order, so this value makes sense again.

xmatthias (approved):
:+1: 
hope we can remove the "BTC" config / mock trades at some point in the future ...

PR #5593, opened 2021-09-19
Parametrized test freqtradebot

samgermain:
There's some that might get overcomplicated with conditionals, I'd like a second opinion on them.

### DIFF
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -359,8 +359,12 @@ def test_create_trade_minimal_amount(
             assert freqtrade.wallets.get_trade_stake_amount('ETH/BTC', freqtrade.edge) == 0
 
 
+@pytest.mark.parametrize('whitelist,positions', [
+    (["ETH/BTC"], 1),  # No pairs left
+    ([], 0),  # No pairs in whitelist
+])
 def test_enter_positions_no_pairs_left(default_conf, ticker, limit_buy_order_open, fee,
-                                       mocker, caplog) -> None:
+                                       whitelist, positions, mocker, caplog) -> None:
     patch_RPCManager(mocker)
     patch_exchange(mocker)
     mocker.patch.multiple(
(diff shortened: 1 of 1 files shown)


xmatthias:
> There's some that might get overcomplicated with conditionals, I'd like a second opinion on them.

Can you maybe point out the ones in particular? (best with a comment to that line) - so i'll have an even closer look at these ?

tests/test_freqtradebot.py
```diff
@@ -407,58 +338,33 @@ def test_create_trade_too_small_stake_amount(default_conf, ticker, limit_buy_ord
...
+            assert rate * amount <= default_conf['stake_amount']
+        else:
+            assert log_has_re(
+                r"Stake amount for pair .* is too small.*",
+                caplog
+            )
+    else:
+        assert not freqtrade.create_trade('ETH/BTC')
+        if not max_open_trades:
+            assert freqtrade.wallets.get_trade_stake_amount('ETH/BTC', freqtrade.edge) == 0
```
samgermain, line 359: @xmatthias  This would be the most complex one

samgermain:
This can be merged whenever by the way

tests/test_freqtradebot.py
xmatthias: yeah it is ... i'll have to think about this a bit (not this test in particular) - while reducing the lines in this file is nice - i fear it can also create some troubles when it comes to (necessary) changes which also require changes to these tests.

samgermain: > yeah it is ... i'll have to think about this a bit (not this test in particular) - while reducing the lines in this file is nice - i fear it can also create some troubles when it comes to (necessary) changes which also require changes to these tests.

if you can think of other code reductions that would also be good, it was just getting ridiculous trying to add a short test for all these methods and I felt I had to do this

A lot of them are less complicated

xmatthias (approved):
Let's give this a try.
if we encounter problems going forward, there's little to nothing preventing us from pulling the tests apart again should the need REALLY arrise.

LGTM :+1:

PR #5598, opened 2021-09-20
Add docker troubleshooting information for win10/wsl

SeriousM:
## Summary

The time in wsl docker container shifts over time. Added information how to fix this issue.

## Quick changelog

- update docs/docker_quickstart.md

### DIFF
--- a/docs/docker_quickstart.md
+++ b/docs/docker_quickstart.md
@@ -151,7 +151,7 @@ You can then run `docker-compose build` to build the docker image, and run it us
 
 ### Troubleshooting
 
-### Docker on Windows
+#### Docker on Windows
 
 * Error: `"Timestamp for this request is outside of the recvWindow."`
   * The market api requests require a synchronized clock but the time in the docker container shifts a bit over time into the past.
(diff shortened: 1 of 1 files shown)


docs/docker_quickstart.md
```diff
@@ -149,6 +149,20 @@ You'll then also need to modify the `docker-compose.yml` file and uncomment the
 
 You can then run `docker-compose build` to build the docker image, and run it using the commands described above.
 
+### Troubleshooting
+
+### Docker on Windows
```
xmatthias: i guess you'll want to have this "below" the troublshooting header, which means it needs to be indented with one additional hash

```suggestion
#### Docker on Windows
```

xmatthias (commented):
i think we need to be careful with these recommendation - as they only apply to a (growing) subset of docker for windows installations (it's still possible to use a regular VM instead of WSL2 with docker on windows).

In general, i'd not recommend using docker on desktop for any production application ...

xmatthias (approved):
LGTM 👍

PR #5601, opened 2021-09-21
parameterize some tests

ferrants:

## Summary

Parameterize some tests, per @samgermain 's request

Solve the issue: None

## Quick changelog

- Parameterized some tests

## What's new?

Fewer tests, reduced code, easier to add new tests

### DIFF
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -3547,16 +3547,21 @@ def test_get_real_amount_no_trade(default_conf, buy_order_fee, caplog, mocker, f
 
 @pytest.mark.parametrize(
     'fee_cost, fee_currency, fee_reduction_amount, use_ticker_rate, expected_log', [
+        # basic, amount does not change
         (None, 'ETH', 0, True, None),
+        # no currency in fee
         (0.004, None, 0, True, None),
+        # BNB no rate
         (0.00094518, "BNB", 0, True, (
             'Fee for Trade Trade(id=None, pair=LTC/ETH, amount=8.00000000, open_rate=0.24544100,'
             ' open_since=closed) [buy]: 0.00094518 BNB - rate: None'
         )),
+        # from order
         (0.004, "LTC", 0.004, False, (
             'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
             'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.996).'
         )),
+        # invalid, no currency in from fee dict
         (0.008, None, 0, True, None),
     ])
 def test_get_real_amount(
@@ -3586,7 +3591,6 @@ def test_get_real_amount(
         mocker.patch('freqtrade.exchange.Exchange.fetch_ticker', side_effect=ExchangeError)
 
     caplog.clear()
-    # Amount does not change
     assert freqtrade.get_real_amount(trade, buy_order) == amount - fee_reduction_amount
 
     if expected_log:
@@ -3595,7 +3599,9 @@ def test_get_real_amount(
 
 @pytest.mark.parametrize(
     'fee_cost, fee_currency, fee_reduction_amount, expected_fee, expected_log_amount', [
+        # basic, amount is reduced by fee
         (None, None, 0.001, 0.001, 7.992),
+        # different fee currency on both trades, fee is average of both trade's fee
         (0.02, 'BNB', 0.0005, 0.001518575, 7.996),
     ])
 def test_get_real_amount_multi(
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -3594,7 +3594,7 @@ def test_get_real_amount(
 
 
 @pytest.mark.parametrize(
-    'stake_currency, fee_cost, fee_currency, fee_reduction_amount, expected_fee, expected_log_amount', [
+    'stake_currency,fee_cost,fee_currency,fee_reduction_amount,expected_fee,expected_log_amount', [
         (None, None, None, 0.001, 0.001, 7.992),
         ("ETH", 0.02, 'BNB', 0.0005, 0.001518575, 7.996),
     ])
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -3594,13 +3594,13 @@ def test_get_real_amount(
 
 
 @pytest.mark.parametrize(
-    'stake_currency,fee_cost,fee_currency,fee_reduction_amount,expected_fee,expected_log_amount', [
-        (None, None, None, 0.001, 0.001, 7.992),
-        ("ETH", 0.02, 'BNB', 0.0005, 0.001518575, 7.996),
+    'fee_cost, fee_currency, fee_reduction_amount, expected_fee, expected_log_amount', [
+        (None, None, 0.001, 0.001, 7.992),
+        (0.02, 'BNB', 0.0005, 0.001518575, 7.996),
     ])
 def test_get_real_amount_multi(
     default_conf, trades_for_order2, buy_order_fee, caplog, fee, mocker, markets,
-    stake_currency, fee_cost, fee_currency, fee_reduction_amount, expected_fee, expected_log_amount,
+    fee_cost, fee_currency, fee_reduction_amount, expected_fee, expected_log_amount,
 ):
 
     trades_for_order = deepcopy(trades_for_order2)
@@ -3611,8 +3611,7 @@ def test_get_real_amount_multi(
 
     mocker.patch('freqtrade.exchange.Exchange.get_trades_for_order', return_value=trades_for_order)
     amount = float(sum(x['amount'] for x in trades_for_order))
-    if stake_currency:
-        default_conf['stake_currency'] = stake_currency
+    default_conf['stake_currency'] = "ETH"
 
     trade = Trade(
         pair='LTC/ETH',
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -3546,24 +3546,19 @@ def test_get_real_amount_no_trade(default_conf, buy_order_fee, caplog, mocker, f
 
 
 @pytest.mark.parametrize(
-    'fee_cost, fee_currency, fee_reduction_amount, use_ticker_rate, expected_log',
-    [
+    'fee_cost, fee_currency, fee_reduction_amount, use_ticker_rate, expected_log', [
         (None, 'ETH', 0, True, None),
         (0.004, None, 0, True, None),
-        (0.00094518, "BNB", 0, True, None),
-        (
-            0.004,
-            "LTC",
-            0.004,
-            False,
-            (
-                'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
-                'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.996).'
-            )
-        ),
+        (0.00094518, "BNB", 0, True, (
+            'Fee for Trade Trade(id=None, pair=LTC/ETH, amount=8.00000000, open_rate=0.24544100,'
+            ' open_since=closed) [buy]: 0.00094518 BNB - rate: None'
+        )),
+        (0.004, "LTC", 0.004, False, (
+            'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
+            'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.996).'
+        )),
         (0.008, None, 0, True, None),
-    ]
-)
+    ])
 def test_get_real_amount(
     default_conf, trades_for_order, buy_order_fee, fee, mocker, caplog,
     fee_cost, fee_currency, fee_reduction_amount, use_ticker_rate, expected_log
@@ -3590,6 +3585,7 @@ def test_get_real_amount(
     if not use_ticker_rate:
         mocker.patch('freqtrade.exchange.Exchange.fetch_ticker', side_effect=ExchangeError)
 
+    caplog.clear()
     # Amount does not change
     assert freqtrade.get_real_amount(trade, buy_order) == amount - fee_reduction_amount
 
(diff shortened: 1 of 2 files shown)


tests/test_freqtradebot.py
```diff
@@ -3545,8 +3545,34 @@ def test_get_real_amount_no_trade(default_conf, buy_order_fee, caplog, mocker, f
...
+        (None, 'ETH', 0, True, None),
+        (0.004, None, 0, True, None),
+        (0.00094518, "BNB", 0, True, None),
+        (
+            0.004,
+            "LTC",
+            0.004,
+            False,
+            (
+                'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
```
ferrants: The log assertions seem like they could be calculated, would that be preferred over asserting a static value?

tests/test_freqtradebot.py
```diff
@@ -3622,85 +3648,29 @@ def test_get_real_amount_multi(default_conf, trades_for_order2, buy_order_fee, c
...
     mocker.patch('freqtrade.exchange.Exchange.markets', PropertyMock(return_value=markets))
-    mocker.patch('freqtrade.exchange.Exchange.fetch_ticker',
-                 return_value={'ask': 0.19, 'last': 0.2})
+    mocker.patch(
+        'freqtrade.exchange.Exchange.fetch_ticker',
+        return_value={'ask': 0.19, 'last': 0.2}
+    )
 
     # Amount is reduced by "fee"
-    assert freqtrade.get_real_amount(trade, buy_order_fee) == amount - (amount * 0.0005)
```
ferrants, line 3664: It was unclear to me how the `0.0005` was determined here and why it didn't match the `expected_fee` or the `fee_cost`. So I left this as a parameter.

xmatthias: You'll have to look at `trades_for_order3` (probably in `conftest.py`) for this.
This test is about gathering fees from a trade-result (because `fetch_order()` didn't contain fees) - and there were 2 trade necessary to fill your order.

I'm not 100% sure if combining all of them into one makes sense (must look at the details of how it's done), as the mocks will need to be different (return something different).

ferrants: I actually meant to remove that fixture, it is only used in this test case, is built on top of the `trades_for_order2` and the fixture only changes the `trades_for_order2` so slightly that it's probably worth just setting as a parameter. Unless we prefer to keep these in fixtures, in which case parameterizing probably doesn't make sense.

https://github.com/freqtrade/freqtrade/blob/6fc770d97dd8d4380be3fb94e8f3dcedfa981cd7/tests/conftest.py#L1688-L1693

ferrants: Removed that fixture

samgermain:
@xmatthias This PR might be good actually, there's some more code that could be reduced, but maybe we can do that as we see it while converting the prices to USDT (which I've started in another branch)

@ferrants good job

tests/test_freqtradebot.py
```diff
@@ -3585,34 +3582,43 @@ def test_get_real_amount_no_currency_in_fee(default_conf, trades_for_order, buy_
...
+    # Amount does not change
+    assert freqtrade.get_real_amount(trade, buy_order) == amount - fee_reduction_amount
+
+    if expected_log:
+        assert log_has(expected_log, caplog)
+
+
+@pytest.mark.parametrize(
+    'stake_currency, fee_cost, fee_currency, fee_reduction_amount, expected_fee, expected_log', [
+        ("BTC", None, None, 0.001, 0.001, (
```
xmatthias: What's the point of stake_currency here (as parameter, i understand it might need to be set as 

BTC seems wrong (considering trades are for LTC/ETH) - and changing it to ETH doesn't seem to have any effect (i also struggle to see where we set stake_currency to BTC in the old tests ...)

tests/test_freqtradebot.py
```diff
@@ -3585,34 +3582,43 @@ def test_get_real_amount_no_currency_in_fee(default_conf, trades_for_order, buy_
...
+            'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
+            'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.992).'
+        )),
+        ("ETH", 0.02, 'BNB', 0.0005, 0.001518575, (
+            'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
+            'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.996).'
```
samgermain: It looks like the 2 strings in these test are identical, so just make the number at the end a parameter

tests/test_freqtradebot.py
ferrants: That's what is in the `default_conf` and since the old `test_get_real_amount_multi2` test was setting it to ETH, being more explicit in the consolidation seemed reasonable.

samgermain: > That's what is in the `default_conf` and since the old `test_get_real_amount_multi2` test was setting it to ETH, being more explicit in the consolidation seemed reasonable.

I would remove it

tests/test_freqtradebot.py
```diff
@@ -3545,32 +3545,29 @@ def test_get_real_amount_no_trade(default_conf, buy_order_fee, caplog, mocker, f
...
+    'fee_cost, fee_currency, fee_reduction_amount, use_ticker_rate, expected_log', [
+        (None, 'ETH', 0, True, None),
+        (0.004, None, 0, True, None),
+        (0.00094518, "BNB", 0, True, (
+            'Fee for Trade Trade(id=None, pair=LTC/ETH, amount=8.00000000, open_rate=0.24544100,'
+            ' open_since=closed) [buy]: 0.00094518 BNB - rate: None'
+        )),
+        (0.004, "LTC", 0.004, False, (
+            'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
+            'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.996).'
```
samgermain, line 3562: can you remove the text here also

tests/test_freqtradebot.py
```diff
@@ -3585,65 +3582,38 @@ def test_get_real_amount_no_currency_in_fee(default_conf, trades_for_order, buy_
...
 
-    # Amount is reduced by "fee"
-    assert freqtrade.get_real_amount(trade, buy_order_fee) == amount - (amount * 0.001)
-    assert log_has('Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
-                   'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.992).',
-                   caplog)
+@pytest.mark.parametrize(
+    'stake_currency, fee_cost, fee_currency, fee_reduction_amount, expected_fee, expected_log_amount', [
+        (None, None, None, 0.001, 0.001, 7.992),
+        ("ETH", 0.02, 'BNB', 0.0005, 0.001518575, 7.996),
```
samgermain: I would remove the `stake_currency` parameter all together, if it's no adding anything to the test

samgermain:
@ferrants 

Do you know about running these tests from the command line?

```
pytest
flake8 freqtrade tests scripts
mypy freqtrade
isort
```

tests/test_freqtradebot.py
ferrants: If I remove it, the second one fails. The prior `test_get_real_amount_multi2` test expects the stake currency in the config to be ETH. I can remove it and change the expected `fee_reduction_amount`, but then it won't match the old `test_get_real_amount_multi2`

ferrants:
@samgermain , yeah, I had run those locally when I originally submitted the PR

Do you guys use anything to automatically format? I've used `black` (https://pypi.org/project/black/) in many projects and it's pretty awesome.

tests/test_freqtradebot.py
samgermain: Can it be ETH for both?

samgermain:
> Do you guys use anything to automatically format? I've used black (https://pypi.org/project/black/) in many projects and it's pretty awesome.

I use autopep8, i think there's plans for black in the future

xmatthias:
> Do you guys use anything to automatically format? I've used `black` (https://pypi.org/project/black/) in many projects and it's pretty awesome.

Black is nice if you start a project with it - it's ... a pretty large change to run on a project this size at this point - and i'm not really comfortable with doing that at the moment (we have 2 pretty big changes in seperate branches - both of which would have the risk of 100000 merge-conflicts in case of applying black now).

We'll probably look at this again - but surely not before feat/short is merged - and probably not until my "async-tests" stuff is either progressed or thrown away ... :laughing:

tests/test_freqtradebot.py
ferrants: Yup, that works. Updated

tests/test_freqtradebot.py
ferrants: Should I not assert it at all? Most of the previous test cases weren't asserting a log line. And the log line on line 3554 is in a different format, the `Fee for Trade` one. Simply passing in a value won't work for that case and then this would be the only version that even asserts. Does that make sense?

xmatthias: I'd leave it as is to be honest.

piecing the logstring together might work - but will make changes to it VERY difficult for future editors.

xmatthias: One additional request though - could you add comments to each parameter-line documenting what we're trying to test? 
Currently it's pretty difficult to determine what exactly is trying to be tested...

tests/test_freqtradebot.py
```diff
@@ -3768,27 +3716,6 @@ def test_get_real_amount_wrong_amount_rounding(default_conf, trades_for_order, b
                    abs_tol=MATH_CLOSE_PREC,)
 
 
-def test_get_real_amount_invalid(default_conf, trades_for_order, buy_order_fee, fee, mocker):
```
ferrants, line 3771: This case appears to be similar to `test_get_real_amount_no_currency_in_fee`, but checking if `currency` was not in the dict at all instead of set to None. This probably caught a regression or specific issue. Should we have the whole `fee` dict as a parameter instead of separate `fee_currency` and `fee_cost`, so we can enable this testing in a parameterized way?

tests/test_freqtradebot.py
```diff
@@ -3545,32 +3545,29 @@ def test_get_real_amount_no_trade(default_conf, buy_order_fee, caplog, mocker, f
...
+        (0.004, None, 0, True, None),
+        (0.00094518, "BNB", 0, True, (
+            'Fee for Trade Trade(id=None, pair=LTC/ETH, amount=8.00000000, open_rate=0.24544100,'
+            ' open_since=closed) [buy]: 0.00094518 BNB - rate: None'
+        )),
+        (0.004, "LTC", 0.004, False, (
+            'Applying fee on amount for Trade(id=None, pair=LTC/ETH, amount=8.00000000, '
+            'open_rate=0.24544100, open_since=closed) (from 8.0 to 7.996).'
+        )),
+        (0.008, None, 0, True, None),
```
ferrants: Now, this is essentially duplicate of https://github.com/freqtrade/freqtrade/pull/5601/files#diff-f0d1952492d82eb12612502e7426fa75194b47a35bf0a04a9ad1af0f924ee25bR3551

Mentioned more here:
https://github.com/freqtrade/freqtrade/pull/5601/files#r714345676

tests/test_freqtradebot.py
samgermain: > This case appears to be similar to `test_get_real_amount_no_currency_in_fee`, but checking if `currency` was not in the dict at all instead of set to None. This probably caught a regression or specific issue. Should we have the whole `fee` dict as a parameter instead of separate `fee_currency` and `fee_cost`, so we can enable this testing in a parameterized way?

I kinda don't want to focus on parametrizing anymore, and get this merged, so that we can start focussing on USDT prices

xmatthias: most of these tests will be quite similar - but all capture another edge-case we've encountered (exchange responses are not always the same - and there's a LOT of edge-cases you can encounter, especially when multiple trades filling one order are involved).

tests/test_freqtradebot.py
samgermain: i don't understand what your trying to say here

ferrants: This test case is the same as ` (0.004, None, 0, True, None),` This is no longer testing what it was originally testing. An option to keep it the same would be to pass in the fee object instead, `{"cost": 0.008}` and `{"cost": 0.004, "currency": None}` for the previous other test case to capture testing the same thing as before

xmatthias: I THINK that's the better way to test this, yes - it's also clearer to understand - IMHO

samgermain: > This test case is the same as ` (0.004, None, 0, True, None),` This is no longer testing what it was originally testing. An option to keep it the same would be to pass in the fee object instead, `{"cost": 0.008}` and `{"cost": 0.004, "currency": None}` for the previous other test case to capture testing the same thing as before

Sounds good, make it test what it was originally testing, I think that this can get merged after that

tests/test_freqtradebot.py
```diff
@@ -3545,8 +3545,33 @@ def test_get_real_amount_no_trade(default_conf, buy_order_fee, caplog, mocker, f
                    caplog)
 
 
-def test_get_real_amount_stake(default_conf, trades_for_order, buy_order_fee, fee, mocker):
-    trades_for_order[0]['fee']['currency'] = 'ETH'
+@pytest.mark.parametrize(
+    'fee_par,fee_reduction_amount,use_ticker_rate,expected_log', [
+        # basic, amount does not change
+        ({'cost': 0.008, 'currency': 'ETH'}, 0, False, None),
```
xmatthias, line 3551: While i updated these tests to use a "fee" dict now ... i've discovered that this testcase does not test the same thing than the existing test.

`test_get_real_amount_stake` (which updates currency to 'ETH') keeps the cost of 0.008 from the original trades_for_order - so a correct test (correctly replicating previous behaviour) would be `{'cost': 0.008, 'currency': 'ETH'}`.

I've now adjusted this test accordingly (check the last commit for necessary changes) ... and based on my comparisons, the remaining tests should be ok, so for me, this would be OK now.

samgermain: ok, can you merge this if there's no other problems?

xmatthias: i can - but wanted to give both you and the original author a chance to look at it.

Also - please stop pushing for merges or reviews. 
I know it takes patience to wait for reviews or merges, but to be honest, freqtrade is quite good in that regards (in other sometimes very active repositories (multiple merges per day) i've contributed to, you can wait for months for even an initial review).
It's putting unnecessary pressure on other people - which might be fine in a company where it's people's main work to do this - but it's not ok for open source projects, where it's people's free time that is invested.
Thanks for your understanding.

samgermain: if it's fine by you it's fine by me

xmatthias: Yes - but when editing some parts of user PR's - it's either small adjustments (Right before merging) - or bigger things (like this).
The 2nd case is where i always like to get feedback / thoughts from the author - maybe there's a reason/thought behind i didn't grasp so far - and am now introducing another problem i didn't see yet.

samgermain: ok, sounds good

xmatthias (approved):
LGTM :+1: 

Hope you're ok with the changes i made ...

Issue #5167, opened 2021-06-21
maximum drawdown percentage seems to be cumulative drawdown


pascal00d:
## Describe your environment

  * Operating system: Ubuntu 20.04
  * Python Version: 3.8.5
  * CCXT version: 1.51.3
  * Freqtrade Version: develop-4512ece1
  
## Your question

In the documentation, explanation for drawdown seems correct, but in the backtesting summary metrics drawdown percentage seems to be cumulative drawdown, not not maximum drawdown.

*Drawdown: Maximum drawdown experienced. For example, the value of 50% means that from highest to subsequent lowest point, a 50% drop was experienced).*

| Starting balance      | 500.000 USDT        |

**| Drawdown              | 48.06%              |**
| Drawdown              | 339.300 USDT        |
| Drawdown high         | 1826.923 USDT       |
| Drawdown low          | 1487.623 USDT       |

Result for drawdown percentage should be : Drawdown / (Starting balance + Drawdown high) = 14,58%

xmatthias:
Relative Drawdown is % of average Stake amount - not absolute drawown - so the 48% (with 339 USDT) suggests you've had an average stake-amount of ~700 USDT.

One can argue if the calculation is correct or wrong ... for relative values, it's always the relation that counts (% of what) - which makes implementation often subjective.

While you do provide a formula - is there an official source (links, etc.) that this is how drawdown should/ must be calculated or is defined - or is it how you think it should be?

pascal00d:
I'm not sure if there is an official definition, but I found definitions like this:
[Maximum Drawdown (MDD) Definition](https://www.investopedia.com/terms/m/maximum-drawdown-mdd.asp)

So after your message i tried to use the % of average Stake amount, instead of the % of the peak value, but i didn't get the same result as you.

% of the peak value => -15%
![image](https://user-images.githubusercontent.com/50653843/123117903-732a1580-d442-11eb-9a6c-cb05bd988bed.png)

% of average Stake amount => -70%
![image](https://user-images.githubusercontent.com/50653843/123118013-8f2db700-d442-11eb-9f59-f107812787a1.png)

"_so the 48% (with 339 USDT) suggests you've had an average stake-amount of ~700 USDT._"
Avg. stake amount is 481.418 USDT, so drawdown (339 USDT) is not 48% of Avg. stake amount.

```
=============== SUMMARY METRICS ===============
| Metric                | Value               |
|-----------------------+---------------------|
[...]

PR #5607, opened 2021-09-22
a new hyperopt loss created that uses calmar ratio

TreborNamor:
This is a new hyperopt loss file that uses the Calmar Ratio.

Calmar Ratio = average annual rate of return / maximum drawdown

## Summary

Explain in one sentence the goal of this PR.

My goal is to introduce a new measure of risk-adjusted returns for investment funds.

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

The Calmar Ratio added to optimize folder.

The Calmar Ratio is a measure of risk-adjusted returns for investment funds that helps minimize the max drawdown of an investors portfolio. Calmar Ratio = Average Annual Rate of Return / Maximum Drawdown. I believe this will help investors feel more at ease when markets takes a hard dump in price.


freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,50 @@
...
+        total_profit = results["profit_ratio"]
+        days_period = (max_date - min_date).days
+
+        # adding slippage of 0.1% per trade
+        total_profit = total_profit - 0.0005
+        expected_returns_mean = total_profit.sum() / days_period
+
+        # calculate max drawdown
+        try:
+            _,_,_,high_val,low_val = calculate_max_drawdown(results)
```
xmatthias: Please be careful with the `max_drawdown()` calculation returned by this method.

it's "per stake" - not the overall total drawdown (maybe read through #5167 - which should explain the issue ...).
i THINK with the columns you're using you're not impacted by this - but it might be worth investigating if instead of using `"profit_ratio"` (the default) - you should be using `"profit_abs"`)  ...

xmatthias: Also - you should run `flake8` locally to have it pass right away ... ;)

xmatthias (commented):
On top of the "formalities" mentioned above ... you should consider adding this to `HYPEROPT_LOSS_BUILTIN` (in `constants.py`) -

and absolutely have to include it in `hyperopt.md` - otherwise it'll be a loss-function that's "hidden" and noone knows about, which would be a shame, as it'd be mostly wasted effort then.

freqtrade/optimize/hyperopt_loss_calmar.py
TreborNamor: thank you for the insight. I was testing the hyperopt right now, and "profit_abs" is the way to go. I made the change to line 40 to  "_,_,_, high_val, low_val = calculate_max_drawdown(results, value_col='profit_abs')"  which gave me the correct max drawdown percentage.

TreborNamor: Calmar Ratio = Total Profit % / Max Drawdown %

I am curious on how I should approach the total profit variable on line 32. For example, let's say the total profit is 203%. When I try using the "total_profit = results['profit_ratio'].sum()" it gives me the decimal 20.30 which equals to 2030%, and if I use "total_profit = results['profit_abs'].sum()" it gives me the dollar amount 20,300 USDT. So I am having trouble getting the correct total profit. I think maybe importing the starting balance could help me fix my issue because I could then use this formula "total_profit = results['profit_abs'].sum() / starting_balance" which will give me the correct decimal to equal 203%. I am worried it will slow down the hyperopt file though. How would you recommend that I get the total profit percentage?

xmatthias: Assuming you're using compounding (stake_amount=unlimited) - then you can't use `sum(profit_ratio)` - as it's not a weighted sum (losing 5% of a 100$ trade is not the same than losing 5% on a 10_000$ trade).

The last parameter of the hyperopt-loss function is `backtest_stats` - (assuming you're using the one from [here](https://www.freqtrade.io/en/stable/advanced-hyperopt/#creating-and-using-a-custom-loss-function)).

While it's contents are not very well documented, it contains everything backtesting is showing - so you can simply use the proper value from there (most likely `backtest_stats['profit_total']` - which contains the ratio - multiply with 100 to get %).

TreborNamor: That did the job! I'm pretty confident I resolved this issue :D

TreborNamor:
Inside CalmarHyperOptLoss file, I am wondering if adding slippage is nessacary since I am not using profit ratio anymore. What do you think?

xmatthias:
> Inside CalmarHyperOptLoss file, I am wondering if adding slippage is nessacary since I am not using profit ratio anymore. What do you think?

Slippage is not relevant to the field you're using in calculations - but to the type of orders you intend on using.
If you plan on using market orders - slippage will be there (and become bigger with bigger stake-sizes).
With limit orders on the other hand, there'll be no slippage, but timedout orders.

What loss-functions that add (or rather subtract) slippage do, is lower the profit per trade to account for this. It's basically up to you to use it - but i probably would - as it lowers results for the calculation slightly (potentially also accounting for some exploits on backtesting logic).

docs/hyperopt.md
```diff
@@ -44,8 +44,9 @@ usage: freqtrade hyperopt [-h] [-v] [--logfile FILE] [-V] [-c PATH] [-d PATH]
                           [--data-format-ohlcv {json,jsongz,hdf5}]
                           [--max-open-trades INT]
                           [--stake-amount STAKE_AMOUNT] [--fee FLOAT]
-                          [-p PAIRS [PAIRS ...]] [--hyperopt-path PATH]
-                          [--eps] [--dmmp] [--enable-protections]
+                          [-p PAIRS [PAIRS ...]] [--hyperopt NAME]
+                          [--hyperopt-path PATH] [--eps] [--dmmp]
+                          [--enable-protections]
```
xmatthias: these changes tell me that you're not on the latest develop (which has explicit `--hyperopt` files disabled/hidden).

Best do a merge from upstream and run this command again ...

TreborNamor: Thank you for letting me know. I am a junior developer trying to figure this stuff all out. I believe I made the merge correctly. Hopefully that takes care of the issue.

xmatthias: i guess it will - if you run `freqtrade hyperopt --help` again and copy/paste the content in here ... :)

TreborNamor: usage: freqtrade hyperopt [-h] [-v] [--logfile FILE] [-V] [-c PATH] [-d PATH] [--userdir PATH] [-s NAME] [--strategy-path PATH] [-i TIMEFRAME] [--timerange TIMERANGE] [--data-format-ohlcv {json,jsongz,hdf5}] [--max-open-trades INT]
                          [--stake-amount STAKE_AMOUNT] [--fee FLOAT] [-p PAIRS [PAIRS ...]] [--hyperopt NAME] [--hyperopt-path PATH] [--eps] [--dmmp] [--enable-protections] [--dry-run-wallet DRY_RUN_WALLET] [-e INT]
                          [--spaces {all,buy,sell,roi,stoploss,trailing,protection,default} [{all,buy,sell,roi,stoploss,trailing,protection,default} ...]] [--print-all] [--no-color] [--print-json] [-j JOBS] [--random-state INT]
                          [--min-trades INT] [--hyperopt-loss NAME] [--disable-param-export]

optional arguments:
  -i TIMEFRAME, --timeframe TIMEFRAME, --ticker-interval TIMEFRAME
                        Specify timeframe (`1m`, `5m`, `30m`, `1h`, `1d`).
  --timerange TIMERANGE
                        Specify what timerange of data to use.
  --data-format-ohlcv {json,jsongz,hdf5}
                        Storage format for downloaded candle (OHLCV) data. (default: `None`).
  --max-open-trades INT
                        Override the value of the `max_open_trades` configuration setting.
  --stake-amount STAKE_AMOUNT
                        Override the value of the `stake_amount` configuration setting.
  --fee FLOAT           Specify fee ratio. Will be applied twice (on trade entry and exit).
  -p PAIRS [PAIRS ...], --pairs PAIRS [PAIRS ...]
                        Limit command to these pairs. Pairs are space-separated.
  --hyperopt NAME       Specify hyperopt class name which will be used by the bot.
  --hyperopt-path PATH  Specify additional lookup path for Hyperopt and Hyperopt Loss functions.
  --eps, --enable-position-stacking
                        Allow buying the same pair multiple times (position stacking).
  --dmmp, --disable-max-market-positions
                        Disable applying `max_open_trades` during backtest (same as setting `max_open_trades` to a very high number).
  --enable-protections, --enableprotections
                        Enable protections for backtesting.Will slow backtesting down by a considerable amount, but will include configured protections
  --dry-run-wallet DRY_RUN_WALLET, --starting-balance DRY_RUN_WALLET
                        Starting balance, used for backtesting / hyperopt and dry-runs.
  -e INT, --epochs INT  Specify number of epochs (default: 100).
  --spaces {all,buy,sell,roi,stoploss,trailing,protection,default} [{all,buy,sell,roi,stoploss,trailing,protection,default} ...]
                        Specify which parameters to hyperopt. Space-separated list.
  --print-all           Print all results, not only the best ones.
  --no-color            Disable colorization of hyperopt results. May be useful if you are redirecting output to a file.
  --print-json          Print output in JSON format.
  -j JOBS, --job-workers JOBS
                        The number of concurrently running jobs for hyperoptimization (hyperopt worker processes). If -1 (default), all CPUs are used, for -2, all CPUs but one are used, etc. If 1 is given, no parallel computing code
                        is used at all.
  --random-state INT    Set random state to some positive integer for reproducible hyperopt results.
  --min-trades INT      Set minimal desired number of trades for evaluations in the hyperopt optimization path (default: 1).
  --hyperopt-loss NAME, --hyperoptloss NAME
                        Specify the class name of the hyperopt loss function class (IHyperOptLoss). Different functions can generate completely different results, since the target for optimization is different. Built-in Hyperopt-
                        loss-functions are: ShortTradeDurHyperOptLoss, OnlyProfitHyperOptLoss, SharpeHyperOptLoss, SharpeHyperOptLossDaily, SortinoHyperOptLoss, SortinoHyperOptLossDaily, CalmarHyperOptLoss, CalmarHyperOptLossDaily
  --disable-param-export
                        Disable automatic hyperopt parameter export.

Common arguments:
  -v, --verbose         Verbose mode (-vv for more, -vvv to get all messages).
  --logfile FILE        Log to the file specified. Special values are: 'syslog', 'journald'. See the documentation for more details.
  -V, --version         show program's version number and exit
  -c PATH, --config PATH
                        Specify configuration file (default: `userdir/config.json` or `config.json` whichever exists). Multiple --config options may be used. Can be set to `-` to read config from stdin.
  -d PATH, --datadir PATH
                        Path to directory with historical backtesting data.
  --userdir PATH, --user-data-dir PATH
                        Path to userdata directory.

Strategy arguments:
  -s NAME, --strategy NAME
                        Specify strategy class name which will be used by the bot.
  --strategy-path PATH  Specify additional strategy lookup path.

TreborNamor:
Hello Matthias, I am not sure how to resolve the coverage error

xmatthias:
> Hello Matthias, I am not sure how to resolve the coverage error

It'll need tests that use these classes.
you can take a look at `test_hyperoptloss.py` how this is done for other tests (it's really dump tests - but they make sure the class is hit / loadable and can run through in standard cases - which (with current tests) is not given (you could have subtle syntax errors like using a variable that doesn't exist without python noticing).

For loss functions, it's not about 100% coverage ... just about "ran through it once" ... :laughing:

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,64 @@
...
+        # calculate max drawdown
+        try:
+            _, _, _, high_val, low_val = calculate_max_drawdown(
+                results, value_col="profit_abs"
+            )
+            max_drawdown = (high_val - low_val) / high_val
+        except ValueError:
+            max_drawdown = 0
+
+        if max_drawdown != 0 and trade_count > 2000:
```
xmatthias: I think this is pretty limiting. 
Getting 2000 trades will either require a very active strategy, or a pretty long timerange (especially for backtesting)

freqtrade/optimize/hyperopt_loss_calmar_daily.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+
+        total_profit = sum_daily["profit_ratio_after_slippage"]
+        expected_returns_mean = total_profit.mean() * 100
+
+        # calculate max drawdown
+        try:
+            high_val = total_profit.max()
+            low_val = total_profit.min()
+            max_drawdown = (high_val - low_val) / high_val
```
xmatthias: I don't think this calculates max-drawdown correctly.

assuming my strategy starts with 1000$ - does well in the beginning and raises to 2000$, before falling back down to 1500$.
I'd then have a drawdown of 500$ - but the value you're calculating with is "2000 - 1000 / 2000" - but the drawdown wasn't 0.5 - but 0.25 (25% of 2000$ ).

interrestingly, in the "non-daily" loss-function you're using a completely different approach ...

xmatthias (changes requested):
While i've fixed merge-conflicts and added tests - the 2 above comments still stand as problematic (but i don't wanna fix it in a way that doesn't align with your opinion).

TreborNamor (commented):
I was having trouble calculating the daily max_drawdown for CalmarHyperOptLossDaily file. I am not sure how to approach this. Is it possible to use the same calculation in CalmarHyperOptLoss or would I have to create a new formula that calculates daily maxdrawdown? I am very new to Pandas so I get stuck quite often.

freqtrade/optimize/hyperopt_loss_calmar.py
```diff
@@ -0,0 +1,64 @@
...
+        Objective function, returns smaller number for more optimal results.
+
+        Uses Calmar Ratio calculation.
+        """
+        total_profit = backtest_stats["profit_total"]
+        days_period = (max_date - min_date).days
+
+        # adding slippage of 0.1% per trade
+        total_profit = total_profit - 0.0005
+        expected_returns_mean = total_profit.sum() / days_period * 100
```
xmatthias, line 46: Maybe, it doesn't make sense to have 2 separate Loss-functions for this, but to only have one instead.

In this line, you're calculating the average return per day ... so this is already average daily returns.

Calculating max-drawdown on daily returns however might be dangerous and might not be the best idea, as you'll hide intraday-drawdown (assuming you lose 5 trades in the morning, but recover most losses in the afternoon, your drawdown would only show the "day end" result but not the real drawdown you had when looking at the stats at lunch time.

As such, i think the potentially best approach here is to remove the "daily" variant, and stick to this implementation only (no one said we always have to have a "normal" and a "daily" variant - we only have them for sortino / sharpe because it (kindof) makes sense - but i don't really see this here for the moment.

TreborNamor:
Yeah, I agree with you. I kind of jump the gun tryin to add a daily
variant. It would make more sense to do a monthly variant instead but it's
probably best to try that at a later time as I develop my programming skills

xmatthias (approved):
I've removed the Daily file as discussed above, and with that, LGTM 👍 

Thanks a lot for your involvement and contribution 💯

PR #5623, opened 2021-09-25
Add CPU,RAM sysinfo support to the REST API to help with bot system m…

froggleston:
…onitoring

## Summary

To add simple CPU and RAM usage support to the REST API.

## Quick changelog

- Add psutil requirement to requirements.txt
- Add psutil import and rpc function call to rpc.py
- Add api_v1 GET endpoint for RPC sysinfo call
- Add REST API client function

## What's new?

A simple addition of a new RPC function to get the underlying bot system's CPU and RAM usage to help with UI development.

e.g. in frogtrade9000

image

### DIFF
--- a/freqtrade/rpc/api_server/api_schemas.py
+++ b/freqtrade/rpc/api_server/api_schemas.py
@@ -348,6 +348,7 @@ class BacktestResponse(BaseModel):
     # TODO: Properly type backtestresult...
     backtest_result: Optional[Dict[str, Any]]
 
+
 class SysInfo(BaseModel):
     cpu_pct: float
     ram_pct: float
--- a/freqtrade/rpc/api_server/api_schemas.py
+++ b/freqtrade/rpc/api_server/api_schemas.py
@@ -347,3 +347,7 @@ class BacktestResponse(BaseModel):
     trade_count: Optional[float]
     # TODO: Properly type backtestresult...
     backtest_result: Optional[Dict[str, Any]]
+
+class SysInfo(BaseModel):
+    cpu_pct: float
+    ram_pct: float
--- a/freqtrade/rpc/api_server/api_schemas.py
+++ b/freqtrade/rpc/api_server/api_schemas.py
@@ -350,5 +350,5 @@ class BacktestResponse(BaseModel):
 
 
 class SysInfo(BaseModel):
-    cpu_pct: float
+    cpu_pct: List[float]
     ram_pct: float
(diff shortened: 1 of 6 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -870,3 +870,6 @@ def _rpc_plot_config(self) -> Dict[str, Any]:
                 'subplots' not in self._freqtrade.strategy.plot_config):
             self._freqtrade.strategy.plot_config['subplots'] = {}
         return self._freqtrade.strategy.plot_config
+
+    def _rpc_sysinfo(self) -> Dict[str, Any]:
+        return {"cpu_pct": psutil.cpu_percent(interval=1, percpu=True), "ram_pct": psutil.virtual_memory().percent}
```
xmatthias: Please either
* move this to the API file itself 
* keep it here, but make this method Static (use `_rpc_get_logs` as sample)

it's not accessing anything from within freqtradebot - so it can work fine as standalone function.
If your api call depends on rpc - it'll not work in webserver mode ... which would be a shame for this method.

Both above methods will work for this so it's up to you in the end.

freqtrade/rpc/api_server/api_v1.py
```diff
@@ -259,3 +259,7 @@ def list_available_pairs(timeframe: Optional[str] = None, stake_currency: Option
         'pair_interval': pair_interval,
     }
     return result
+
+@router.get('/sysinfo', tags=['info'])
+def sysinfo(rpc: RPC = Depends(get_rpc)):
```
xmatthias: using `Depends(get_rpc)` will mean it'll not work for webserver mode ... which is unnecessary in this case.

xmatthias (commented):
How does this work with docker (what will it display)?
i don't think a docker-process will have access to the full system statistics ... ?

You'll also want to define a schema (file `api_schema.py`) for this (and add it to the `router.get()` call so we know what to expect from the result (it'll then be documented in the openAPI spec).

On top of that, a test would be nice (nothing fancy - just test if the expected fields are in the response ... (don't test for values, these will obviously vary).

froggleston:
I think I've done the main code refactor, but haven't done the test. Where do I add tests into the codebase? I can't seem to find it :)

Also, not sure about docker, but I think that psutil returns the system information of the host....

freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -347,3 +347,8 @@ class BacktestResponse(BaseModel):
     trade_count: Optional[float]
     # TODO: Properly type backtestresult...
     backtest_result: Optional[Dict[str, Any]]
+
+
+class SysInfo(BaseModel):
+    cpu_pct: float
```
xmatthias: Your type seems wrong.
While i've added a test now - cpu_pct is not returning a float - so the test fails.
fixing the type here should also fix the test.

froggleston: Hmm weird: I see `{"cpu_pct": [1.0, 32.7, 1.0, 68.0], "ram_pct": 28.6}` when calling this, so they should be floats?

froggleston: Oh wait, I'm thick. I'll set it to List[float] type.

xmatthias (approved):
LGTM :+1: 

Nice one :100:

PR #5640, opened 2021-09-29
markets_static fixture

rokups:
Introduce markets_static fixture serving an immutable list of markets. Adapt pairlist/markets tests to use this new fixture.

This allows freely modifying markets in get_markets() without a need of updating pairlist/markets tests.

This facilitates #5592. We need to introduce new markets in order to convert tests in that PR to use USDT pairs, while not disturbing other tests.


xmatthias (approved):
Nice :+1:

PR #5644, opened 2021-09-30
Add ability to ignore unparameterized spaces

eadgbear:
## Summary

Suppress the errors generated by requesting optimization of a space (via the `--spaces` option) for a strategy that doesn't define any parameters for said spaces. Can be useful while developing and tweaking said parameters.

## Quick changelog

- Adding command line parameter `--ignore-unparameterized-spaces` to hyperopt subcommand

## What's new?

Using the `--ignore-unparameterized-spaces` flag for `freqtrade hyperopt` will log any `OperationalErrors` generated by unparameterized spaces as `WARN` as opposed to an `ERROR`.

### DIFF
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -554,7 +554,8 @@ AVAILABLE_CLI_OPTIONS = {
     ),
     "hyperopt_ignore_missing_space": Arg(
         "--ignore-missing-spaces", "--ignore-unparameterized-spaces",
-        help="Suppress errors for any requested Hyperopt spaces that do not contain any parameters",
+        help=("Suppress errors for any requested Hyperopt spaces "
+              "that do not contain any parameters."),
         action="store_true",
     ),
 }
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -552,4 +552,9 @@ AVAILABLE_CLI_OPTIONS = {
         help='Do not print epoch details header.',
         action='store_true',
     ),
+    "hyperopt_ignore_unparam_space": Arg(
+        "-u", "--ignore-unparameterized-spaces",
+        help="Suppress errors for any requested Hyperopt spaces that do not contain any parameters",
+        action="store_true",
+    ),
 }
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -552,8 +552,8 @@ AVAILABLE_CLI_OPTIONS = {
         help='Do not print epoch details header.',
         action='store_true',
     ),
-    "hyperopt_ignore_unparam_space": Arg(
-        "-u", "--ignore-unparameterized-spaces",
+    "hyperopt_ignore_missing_space": Arg(
+        "--ignore-missing-spaces", "--ignore-unparameterized-spaces",
         help="Suppress errors for any requested Hyperopt spaces that do not contain any parameters",
         action="store_true",
     ),
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -237,27 +237,63 @@ class Hyperopt:
             logger.debug("Hyperopt has 'protection' space")
             # Enable Protections if protection space is selected.
             self.config['enable_protections'] = True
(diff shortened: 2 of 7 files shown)


xmatthias:
While i understand what the PR will do - i'm 
a) not sure it'll not create failures down the hyperopt route
b) not sure what you try to accomplish with this

If you're currently tweaking `buy` parameters, then you'll either have them defined (and not finding them is a breaking error) - or you haven't defined them yet (then don't include this space).
so i think, simply excluding the space that's not ready yet (instead of  `--spaces buy sell` use only `--spaces sell` ) would avoid that error.
Logging the error will result in the error being ignored, and people wondering why it didn't optimize that parameter.
Users will not really see the problem that way anyway as it probably scrolls out of view (hyperopt will continue ...) - which is why (for development also) - having the strategy stop at this point will bring the error "right to the bottom" of your output.


Maybe you can explain the usecase for this a little better - as for the moment, i'm not convinced on which benefits this will bring (also not while developing a strategy).

eadgbear:
@xmatthias Fair points!

The use case that I was using it for was to bulk train Hyperopt-enhanced strategies. It's easier to have an opt-in flag to disable the error and just specify the spaces to train across all of the strategies regardless of whether or not they support it.

freqtrade/commands/cli_options.py
```diff
@@ -552,4 +552,9 @@ def __init__(self, *args, **kwargs):
         help='Do not print epoch details header.',
         action='store_true',
     ),
+    "hyperopt_ignore_unparam_space": Arg(
+        "-u", "--ignore-unparameterized-spaces",
```
xmatthias: To be honest, i'd use shorthand commands sparingly (only for important options that are needed often, like `--config` or `--strategy`).

For the moment, i consider this as a edge-case option - which i'd not like to "waste" a single letter shortcut for (not many people will probably use this (and your usecase is in a script as well) - so it should be fine to type it out.

freqtrade/optimize/hyperopt.py
```diff
@@ -237,27 +237,63 @@ def init_spaces(self):
...
+                else:
+                    raise
 
         if HyperoptTools.has_space(self.config, 'buy'):
             logger.debug("Hyperopt has 'buy' space")
-            self.buy_space = self.custom_hyperopt.buy_indicator_space()
+            try:
+                self.buy_space = self.custom_hyperopt.buy_indicator_space()
+            except OperationalException as e:
+                if self.config["hyperopt_ignore_unparam_space"]:
```
xmatthias: you MUST use `self.config.get("hyperopt_ignore_unparam_space", False):` here (obviously also below) - otherwise this will fail when the option is not given (this is what's causing tests to fail)

freqtrade/optimize/hyperopt.py
```diff
@@ -237,27 +237,63 @@ def init_spaces(self):
...
+                    raise
 
         if HyperoptTools.has_space(self.config, 'sell'):
             logger.debug("Hyperopt has 'sell' space")
-            self.sell_space = self.custom_hyperopt.sell_indicator_space()
+            try:
+                self.sell_space = self.custom_hyperopt.sell_indicator_space()
+            except OperationalException as e:
+                if self.config["hyperopt_ignore_unparam_space"]:
+                    logger.warning(e)
```
xmatthias: On a 2nd thought - maybe consider moving this to `hyperopt_auto.py` (although i'm not sure how to handle return value then) (it'd be a nice to have - it'll however need some further tests / experimentation to decide if that'll really work).

the exception itself is thrown out of `_format_exception_message` - so we'd have one location to check for this and return a sane default (probably a empty list?)

xmatthias (changes requested):
I'd want to see a test with this option as well (it's for regression protection, really).

Maybe duplicate the currently failing tests (`test_simplified_interface_failed`) and instead of checking that the exception is raised, check if the expected log-message is present (could also be done with an additional parameter, assuming you're familiar with parametrized tests).

One thing that for sure will need to change is the failing test ... which is however caused by the code itself.

xmatthias (approved):
Great idea 👍 

I've moved the implementation to hyperopt_auto - and you'll notice that it's now a lot less code, as it's only there once.

LGTM 👍

PR #5648, opened 2021-10-01
Fix pair_candles to point to correct API call

froggleston:
## Summary

pair_candles pointed to available_pairs RPC call instead of pair_candles


xmatthias (approved):
Nice catch :+1:

PR #5668, opened 2021-10-04
Docs for leverage

samgermain:
Adds docs for leverage

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -449,45 +449,6 @@ The possible values are: `gtc` (default), `fok` or `ioc`.
     This is ongoing work. For now, it is supported only for binance and kucoin.
     Please don't change the default value unless you know what you are doing and have researched the impact of using different values for your particular exchange.
 
-### Exchange configuration
-
-Freqtrade is based on [CCXT library](https://github.com/ccxt/ccxt) that supports over 100 cryptocurrency
-exchange markets and trading APIs. The complete up-to-date list can be found in the
-[CCXT repo homepage](https://github.com/ccxt/ccxt/tree/master/python).
- However, the bot was tested by the development team with only Bittrex, Binance and Kraken,
- so these are the only officially supported exchanges:
-
-- [Bittrex](https://bittrex.com/): "bittrex"
-- [Binance](https://www.binance.com/): "binance"
-- [Kraken](https://kraken.com/): "kraken"
-
-Feel free to test other exchanges and submit your PR to improve the bot.
-
-Some exchanges require special configuration, which can be found on the [Exchange-specific Notes](exchanges.md) documentation page.
-
-#### Sample exchange configuration
-
-A exchange configuration for "binance" would look as follows:
-
-```json
-"exchange": {
-    "name": "binance",
-    "key": "your_exchange_key",
-    "secret": "your_exchange_secret",
-    "ccxt_config": {"enableRateLimit": true},
-    "ccxt_async_config": {
-        "enableRateLimit": true,
-        "rateLimit": 200
-    },
-```
-
-This configuration enables binance, as well as rate-limiting to avoid bans from the exchange.
-`"rateLimit": 200` defines a wait-event of 0.2s between each call. This can also be completely disabled by setting `"enableRateLimit"` to false.
-
-!!! Note
-    Optimal settings for rate-limiting depend on the exchange and the size of the whitelist, so an ideal parameter will vary on many other settings.
-    We try to provide sensible defaults per exchange where possible, if you encounter bans please make sure that `"enableRateLimit"` is enabled and increase the `"rateLimit"` parameter step by step.
-
 ### What values can be used for fiat_display_currency?
 
 The `fiat_display_currency` configuration parameter sets the base currency to use for the
@@ -527,9 +488,9 @@ creating trades on the exchange.
 ```json
 "exchange": {
     "name": "bittrex",
-        "key": "key",
-        "secret": "secret",
-        ...
+    "key": "key",
+    "secret": "secret",
+    ...
 }
 ```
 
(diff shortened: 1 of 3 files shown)


docs/configuration.md
```diff
@@ -90,6 +90,8 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
 | `trailing_only_offset_is_reached` | Only apply trailing stoploss when the offset is reached. [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*  <br> **Datatype:** Boolean
 | `fee` | Fee used during backtesting / dry-runs. Should normally not be configured, which has freqtrade fall back to the exchange default fee. Set as ratio (e.g. 0.001 = 0.1%). Fee is applied twice for each trade, once when buying, once when selling. <br> **Datatype:** Float (as ratio)
+| `trading_mode` | Specifies if you want to trade regularly, trade with leverage, or trade contracts whose prices are derived from matching cryptocurrency prices. spot, margin or futures
+| `collateral` | When trading with leverage, this determines if the collateral owned by the trader will be shared or isolated to each trading pair
```
xmatthias: these should definitely follow the same structure as the others (defining datatype, ...)

docs/configuration.md
```diff
@@ -447,6 +449,65 @@ The possible values are: `gtc` (default), `fok` or `ioc`.
     This is ongoing work. For now, it is supported only for binance and kucoin.
     Please don't change the default value unless you know what you are doing and have researched the impact of using different values for your particular exchange.
 
+### Understand trading_mode
```
xmatthias: I think all "leverage/short" relevant settings should go to a seperate file (i think we've got `leverage.md`, iirc).
That makes it clear and distinct where they belong (and we can use ONE disclaimer at the very top of that page - instead of distributing them around different files).

docs/leverage.md
```diff
@@ -1,21 +1,81 @@
 # Leverage
+*You can't run 2 bots on the same account with leverage,*
+
+!!! Warning: Trading with leverage(`trading_mode="margin"` or `trading_mode="futures"`) is very risky.
```
samgermain: What's the actual way I should give a big warning? This is failing a test. I pulled the !!! from somewhere else in the freqtrade docs

xmatthias: The syntax for the  "info" / warning / ...  boxes can be found [here](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) - we also link to it from the [developer docs](https://www.freqtrade.io/en/stable/developer/#documentation)

the syntax is basicaly `!!! Warning ["Optional header"]` (no colon!) - followed by the content, indented by 4 spaces.

samgermain: is `"Optional header"` literally in quotations?

xmatthias: in quotes - yes.
We're using this in "some" places - searching for `!!! Note ` (note the space at the end) should yield some examples. Looking at the rendered documentation then, you can see the outcome (or use `mkdocs serve` locally)

docs/leverage.md
```diff
@@ -1,21 +1,80 @@
...
+    Regular trading mode. 
+        - Shorting is not available
+        - There is no liquidation price
+        - Profits gained/lost are equal to the change in value of the assets(minus trading fees)
+
+#### Leverage trading modes
+
+# TODO: include a resource to help calculate stoplosses that are above the liquidation price
+
+#TODO: Taken from investopedia, is that ok?
```
xmatthias: Should be - maybe put a "source" link to the bottom of it - so it's clear where it's been sourced (and i assume Investopedia has more info on this - so people can go read more there)...
i'd maybe also change it a bit to say "asset" / currency - instead of "firms asset base" - as i don't think that's really applicable here.

docs/configuration.md
```diff
@@ -485,7 +526,7 @@ creating trades on the exchange.
 
 ```json
 "exchange": {
-        "name": "bittrex",
+    "name": "bittrex",
```
xmatthias, line 490: I'm not per se against this - but if we change formatting of this line - then please also for the others ...

docs/configuration.md
```diff
@@ -447,6 +449,45 @@ The possible values are: `gtc` (default), `fok` or `ioc`.
...
+Freqtrade is based on [CCXT library](https://github.com/ccxt/ccxt) that supports over 100 cryptocurrency
+exchange markets and trading APIs. The complete up-to-date list can be found in the
+[CCXT repo homepage](https://github.com/ccxt/ccxt/tree/master/python).
+ However, the bot was tested by the development team with only Bittrex, Binance and Kraken,
+ so these are the only officially supported exchanges:
+
+- [Bittrex](https://bittrex.com/): "bittrex"
+- [Binance](https://www.binance.com/): "binance"
+- [Kraken](https://kraken.com/): "kraken"
+
```
xmatthias: i've intentionally removed (or rather moved) this section from configuration to exchange.md in 7f4baab420ce98c97deeedb69170c3129828c9b2 - so if there's changes to it, these should be done in that segment too, without re-adding this segment.

docs/leverage.md
```diff
@@ -1,21 +1,80 @@
...
+
+``` python
+"trading_mode": "futures"
+```
+
+### Collateral
+
+The possible values are: `isolated`, or `cross`(*coming soon*)
+
+# TODO: I took this definition from bitmex, is that fine? https://www.bitmex.com/app/isolatedMargin
```
xmatthias: taking inspiration is for sure fine...
I think we'll need to add some of "our own" text to it though ... like "freqtrade will place a stoploss above the margin level to avoid forced liquidations" ... (and/or whatever else applies ...)

xmatthias (approved):
Let's merge this for now, i don't think we have anything to contribute to this at this point.
We can always do eventually necessary additional additions at a later point

PR #5670, opened 2021-10-04
Added min_profit param to PerformanceFilter

sergiikhliustin:
## Summary

As a user, I want to filter out pair by min profit (according to minutes param, it will be minimal profit during some time)

## Quick changelog

Added min_profit param to PerformanceFilter

### DIFF
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -194,17 +194,22 @@ Trade count is used as a tie breaker.
 You can use the `minutes` parameter to only consider performance of the past X minutes (rolling window).
 Not defining this parameter (or setting it to 0) will use all-time performance.
 
+The optional `min_profit` parameter defines the minimum profit a pair must have to be considered.
+Pairs below this level will be filtered out.
+Using this parameter without `minutes` is highly discouraged, as it can lead to an empty pairlist without without a way to recover.
+
 ```json
 "pairlists": [
     // ...
     {
         "method": "PerformanceFilter",
-        "minutes": 1440  // rolling 24h
+        "minutes": 1440,  // rolling 24h
+        "min_profit": 0.01
     }
 ],
 ```
 
-!!! Note
+!!! Warning "Backtesting"
     `PerformanceFilter` does not support backtesting mode.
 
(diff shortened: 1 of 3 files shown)


xmatthias (changes requested):
Blocking point: Please add new configuration options to the relevant documentation section (in this case, `pairlist.md`, otherwise it'll become a "forgotten" feature noone knows about.

general feedback:
I think you'll want a "min-trades" option to go along with this.
Otherwise this filter will become a risk if you're just starting out.

Assume your bot just started, and did 2 trades (one positive, one negative).

Your pairlist will then be reduced to the one positive trading pair - never giving any other pair even a chance - so your bot just became a "one pair" bot. This is especially true when using it without "minutes" filter (using the full history), but can also happen with that (although then it's limited to this short period).

Worst case, you'll lose the 3rd trade (on this same pair) - now you've got 0 pairs left in your pairlist.

This is not a problem at the moment as this pairlist is not filtering, just sorting, but with "filtering" enabled, this will become a concern.

sergiikhliustin:
> Blocking point: Please add new configuration options to the relevant documentation section (in this case, `pairlist.md`, otherwise it'll become a "forgotten" feature noone knows about.
> 
> general feedback: I think you'll want a "min-trades" option to go along with this. Otherwise this filter will become a risk if you're just starting out.
> 
> Assume your bot just started, and did 2 trades (one positive, one negative).
> 
> Your pairlist will then be reduced to the one positive trading pair - never giving any other pair even a chance - so your bot just became a "one pair" bot. This is especially true when using it without "minutes" filter (using the full history), but can also happen with that (although then it's limited to this short period).
> 
> Worst case, you'll lose the 3rd trade (on this same pair) - now you've got 0 pairs left in your pairlist.
> 
> This is not a problem at the moment as this pairlist is not filtering, just sorting, but with "filtering" enabled, this will become a concern.

I want to filter low-profit pairs, but give them second chance after some time. Will add to documentation about risks of usage without minutes param

xmatthias (approved):
After some small adjustments: LGTM :+1:

PR #5674, opened 2021-10-05
Update docker_quickstart.md

jonny07:
Got help in the discord chat to get the UI running, I think most people will need this...

## Summary

Add the UI access to the Doc. Currently it is explained in later subchapter only and not working.

## Quick changelog

Added chapter "#### Acessing the UI"

## What's new?

Added chapter "#### Acessing the UI"

### DIFF
--- a/docs/docker_quickstart.md
+++ b/docs/docker_quickstart.md
@@ -70,6 +70,40 @@ docker-compose up -d
 !!! Warning "Default configuration"
     While the configuration generated will be mostly functional, you will still need to verify that all options correspond to what you want (like Pricing, pairlist, ...) before starting the bot.
 
+#### Acessing the UI
+
+Uncommend the 2 lines below and add your IP adress in the following format (like 192.168.2.67:8080:8080) to the ft_userdata/docker-compose.yml:
+'''bash
+    ports:
+      - "yourIPadress:8080:8080"
+'''
+Your ft_userdata/user_data/config.json should look like:
+'''bash
+api_server": {
+        "enabled": true,
+        "listen_ip_address": "0.0.0.0",
+        "listen_port": 8080,
+        "verbosity": "error",
+        "enable_openapi": false,
+        "jwt_secret_key": "****",
+        "CORS_origins": [],
+        "username": "****",
+        "password": "****"
+    },
+'''
+instead of "****" you will have your data in.
+Then rebuild your docker file:
+Linux:
+'''bash
+sudo docker-compose down && sudo docker-compose pull && sudo docker-compose build && sudo docker-compose up -d 
+'''
+Windows:
+'''bash
+docker-compose down && docker-compose pull && docker-compose build && docker-compose up -d 
+'''
+
+You can now access the UI by typing yourIPadress:8080 in your browser.
+
 #### Monitoring the bot
 
 You can check for running instances with `docker-compose ps`.
--- a/docs/docker_quickstart.md
+++ b/docs/docker_quickstart.md
@@ -70,39 +70,17 @@ docker-compose up -d
 !!! Warning "Default configuration"
     While the configuration generated will be mostly functional, you will still need to verify that all options correspond to what you want (like Pricing, pairlist, ...) before starting the bot.
 
-#### Acessing the UI
+#### Accessing the UI
 
-Uncommend the 2 lines below and add your IP adress in the following format (like 192.168.2.67:8080:8080) to the ft_userdata/docker-compose.yml:
-'''bash
-    ports:
-      - "yourIPadress:8080:8080"
-'''
-Your ft_userdata/user_data/config.json should look like:
-'''bash
-api_server": {
-        "enabled": true,
-        "listen_ip_address": "0.0.0.0",
-        "listen_port": 8080,
-        "verbosity": "error",
-        "enable_openapi": false,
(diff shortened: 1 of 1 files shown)


docs/docker_quickstart.md
```diff
@@ -70,6 +70,40 @@ docker-compose up -d
 !!! Warning "Default configuration"
     While the configuration generated will be mostly functional, you will still need to verify that all options correspond to what you want (like Pricing, pairlist, ...) before starting the bot.
 
+#### Acessing the UI
+
+Uncommend the 2 lines below and add your IP adress in the following format (like 192.168.2.67:8080:8080) to the ft_userdata/docker-compose.yml:
+'''bash
```
xmatthias: These are the wrong ticks. you'll want ``` for the docs to render this as code sections.
This segment is also not bash - so the coloring would be wrong.

docs/docker_quickstart.md
```diff
@@ -70,6 +70,40 @@ docker-compose up -d
...
+    ports:
+      - "yourIPadress:8080:8080"
+'''
+Your ft_userdata/user_data/config.json should look like:
+'''bash
```
xmatthias: This is `.json` - (and wrong ticks again).

docs/docker_quickstart.md
```diff
@@ -70,6 +70,40 @@ docker-compose up -d
...
+        "CORS_origins": [],
+        "username": "****",
+        "password": "****"
+    },
+'''
+instead of "****" you will have your data in.
+Then rebuild your docker file:
+Linux:
+'''bash
+sudo docker-compose down && sudo docker-compose pull && sudo docker-compose build && sudo docker-compose up -d
```
xmatthias: this is a ... dangerous command.

the build step will not work UNLESS you also have a custom docker file - so at this step of the documentation, i think this is wrong (the build step should only be in the section about custom dependencies).
will need to test if the build-step fails without build-section, or if it's simply ignored if no build is necessary.
It's however dangerous as the build-step will require an additional step (change the image name) - otherwise the pull step will not do anything assuming you have the "latest build" image already available.

You'll also not want to run docker-compose as root (via sudo) - using sudo is only necessary on badly configured systems.

xmatthias (changes requested):
I'm not certain we'll need this section here.

if you look at the [latest page](https://www.freqtrade.io/en/latest/docker_quickstart/#editing-the-docker-compose-file) - you'll notice that we have a link to the UI section now (in the warning box).

Having the same documentation in 2 different areas of the Documentation will lead to misalignment an a short period of time (who will remember in 2 months from now that we need to update this in 2 areas??) ... so i prefer linking them together (see above link) instead of copying documentation around.

You should render the documentation locally [as described in the contributors section](https://www.freqtrade.io/en/latest/developer/#documentation) to check how it looks and fix "simple" syntax errors, like wrong backticks...

jonny07:
Thank you for taking the time and for your feedback. Was my first attempt to contribute on github, and realize still some errors in it, thank you for staying kind and explaining.

Please find my thoughts for the documentation and problems I had with it and proposal below:
I think the UI should get more attention in the documentation than it currently has, as this is main or at least very important part of freqtrade.
When I searched for a (back)trading SW I needed something with optical output, as this helps a lot to analyse how your strategy works and where it needs improvement. This optical output is available in the UI in freqtrade, however no screenshot in the documentation. I would propose to add a screenshot of backtesting results and UI and the UI installation right at the beginning of the documentation. A screenshot or several will make it much easier and faster for people to see what they can expect from the SW than reading the complete documentation. Currently the UI is mentioned in a sub sub chapter all the way down, I would propose to have it in the quickstart with an own chapter. I agree there should not be double documentation. The current warning box you mentioned is not labeled with UI, even after you told me it was hard to find. 
Overall I would not have taken freqtrade into account if I would not have found the video on youtube explaining it and showing the UI. I was very surprised to see a nice UI which from screening the description seemed to me to be just console sw (I did not read everything).
In the sub chapter for Rest API I propose to add in the headlines that a description will only work for a special install, then I would have looked for the docker explanation further down. But did not expect that as already a new chapter for security started. As a docker installation is recommended, I would propose to put the chapter for docker first.

xmatthias:
The UI is still in heavy development / flux - and should be considered alpha (or at max early beta).
while it works, certain errors can still appear in certain conditions.

as such, i don't think it's appropriate to have it "full-on" featured in the documentation, simply because the readiness of it is different to the rest of freqtrade.
Freqtrade itself  (the bot part) is pretty much "battle-proven" - and pretty stable overall - which is not true in all areas of the UI.

As the UI stabilizes, we'll for sure highlight it in the documentation (maybe even auto-enable it in certain conditions) - However - that's not something I'd want to do until i believe it's out of beta.

Maybe it's my own mindset that's wrong here ... but I'm not a strong believer in "early" marketing (making a feature or part seem complete while the user is really the alpha-tester). I know many tech-companies do it this way now - but it's not something i like - and therefore try to not do it with the projects i run.

Hope this makes sense.

jonny07:
Yes, totally fine, thanks for your reply. Was just a proposal, of course one can see things different and also I got your explanation and makes sense.

xmatthias (approved):
I've made (quite some) changes to this - but we now have a new point about accessing the UI in the quickstart guide :+1:

PR #5675, opened 2021-10-05
Update rest-api.md

jonny07:
Available description did not work for me, so I extended it.

## Summary

Added few lines to get the UI running.

## Quick changelog

Added few lines to get the UI running.

## What's new?

Added few lines to get the UI running.

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -35,6 +35,23 @@ Sample configuration:
     },
 ```
 
+Also uncommend the 2 lines below in ft_userdata/docker-compose.yml:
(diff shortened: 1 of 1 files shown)


xmatthias (changes requested):
Almost the same content is around lines 84+.

If you think that section needs improvement, please update that section - but we'll not have the same content twice (especially not in the same file).

xmatthias:
After having another look at this PR i don't think it makes sense.
The same exact content is already available ~30 lines below - so this does not add anything other than duplication.

PR #5677, opened 2021-10-06
Update CI to run on python 3.10

xmatthias:
## Summary
Ci to python 3.10, as it's now released.


xmatthias:
Currently blocked by https://github.com/PyTables/PyTables/issues/909

ghost (approved):
>

samgermain:
`asyncio.get_event_loop()` is deprecated in python3.10, I saw it in `exchange.refresh_latest_ohlcv`

https://docs.python.org/3/library/asyncio-eventloop.html

xmatthias:
as long as pytables is not updating, installation is impossible on 3.10 anyway - so I'm honestly not too worried about that at this point.
Also, deprecated means we should get rid of it before 3.11 ... which is still more than a year in the future ;)

PR #5678, opened 2021-10-06
Hyperopt loss function using max drawdown

### LINKED ISSUE #5514: hyperopt loss function max draw down
Hi!

here a small addition to the loss functions already available.

```
from datetime import datetime
from freqtrade.data.btanalysis import calculate_max_drawdown
from freqtrade.optimize.hyperopt import IHyperOptLoss
from pandas import DataFrame

# optimize for max draw down and profit
# less max draw down more profit -> lower return value
class MaxDrawDownHyperOptLoss(IHyperOptLoss):
    @staticmethod
    def hyperopt_loss_function(results: DataFrame, trade_count: int,
                               min_date: datetime, max_date: datetime,
                               *args, **kwargs) -> float:

        max_drawdown = calculate_max_drawdown(results)
        max_drawdown_rev = 1 / max_drawdown[0]
        total_profit = results['profit_abs'].sum()
        ret = max_drawdown_rev * total_profit

        return -ret
```

sidml:
## Summary

Hyperopt loss function using max drawdown.
As mentioned [here](https://github.com/freqtrade/freqtrade/issues/5514#issuecomment-909055853), the function calculates a multiplier from drawdown, the lower drawdown is, the higher the multiplier.

closes https://github.com/freqtrade/freqtrade/issues/5514

### DIFF
--- a/freqtrade/optimize/hyperopt_loss_max_drawdown.py
+++ b/freqtrade/optimize/hyperopt_loss_max_drawdown.py
@@ -5,10 +5,11 @@ This module defines the alternative HyperOptLoss class which can be used for
 Hyperoptimization.
 """
 from datetime import datetime
-from freqtrade.data.btanalysis import calculate_max_drawdown
-from freqtrade.optimize.hyperopt import IHyperOptLoss
 
 from pandas import DataFrame
+
+from freqtrade.data.btanalysis import calculate_max_drawdown
+from freqtrade.optimize.hyperopt import IHyperOptLoss
 
 
 class MaxDrawDownHyperOptLoss(IHyperOptLoss):
@@ -31,7 +32,7 @@ class MaxDrawDownHyperOptLoss(IHyperOptLoss):
         Uses profit ratio weighted max_drawdown when drawdown is available.
         Otherwise directly optimizes profit ratio.
         """
-        total_profit = results['profit_ratio'].sum()        
+        total_profit = results['profit_ratio'].sum()
         try:
             max_drawdown = calculate_max_drawdown(results)
         except ValueError:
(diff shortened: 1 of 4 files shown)


freqtrade/optimize/hyperopt_loss_max_drawdown.py
```diff
@@ -0,0 +1,43 @@
...
+        Uses profit ratio weighted max_drawdown when drawdown is available.
+        Otherwise directly optimizes profit ratio.
+        """
+        total_profit = results['profit_ratio'].sum()
+        try:
+            max_drawdown = calculate_max_drawdown(results)
+        except ValueError:
+            # No losing trade, therefore no drawdown.
+            return -total_profit
+        max_drawdown_rev = 1 / max_drawdown[0]
```
xmatthias: In general, i'd argue not to use "this" drawdown.
It's based on profit_ratio - which is highly dependent on the stake-amount. History has shown that using `profit_abs` yields better results (you can also see this in backtesting when comparing the absolute drawdown (in $ or whatever) with the % drawdown. 

The % drawdown is "per stake" - which becomes fluky with dynamic stake amounts.
We'll probably change the calculation for backtesting too, and i'm currently encouraging everyone using drawdown to use the column "profit_abs" instead (obviously the "total_profit" calculation has to be aligned to that, too.

That change is as simple as adding ` value_col='profit_abs'` to the `calculate` call - and using the same column for total-profit.

sidml: Thanks! I have made the suggested changes.

freqtrade/optimize/hyperopt_loss_max_drawdown.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        total_profit = results['profit_abs'].sum()
+        try:
+            max_drawdown = calculate_max_drawdown(results, value_col='profit_abs')
+        except ValueError:
+            # No losing trade, therefore no drawdown.
+            return -total_profit
+        max_drawdown_rev = 1 / max_drawdown[0]
+        ret = max_drawdown_rev * total_profit
+        return -ret
```
FranciscoDA: i think these 3 lines can be simplified to `return -total_profit / max_drawdown[0]`

xmatthias (approved):
LGTM :+1:

Issue #3078, opened 2020-03-17
jinja2.exceptions.TemplateNotFound: subtemplates/exchange_generic.j2 Error


bDyEz8c8CtL8dJ8aaQhV8:
## Step 1: Have you search for this issue before posting it?

If you have discovered a bug in the bot, please [search our issue tracker](https://github.com/freqtrade/freqtrade/issues?q=is%3Aissue). 
If it hasn't been reported, please create a new issue.

## Step 2: Describe your environment

  * Operating system: Arch Linux
  * Python Version: 7.13 
  * CCXT version: 1.24.37 
  * Branch: Master 
  
## Step 3: Describe the problem:

I installed the package from AUR, it is the latest version. I am getting a jinja error.

### Steps to reproduce:

  1. Try to create new config
  
### Observed Results:

  * What happened?
I expected it to work.

  * What did you expect to happen?
Threw jinja error.
### Relevant code exceptions or logs:

  ```
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/freqtrade/main.py", line 36, in main
    return_code = args['func'](args)
  File "/usr/lib/python3.8/site-packages/freqtrade/commands/build_config_commands.py", line 193, in start_new_config
    deploy_new_config(config_path, selections)
  File "/usr/lib/python3.8/site-packages/freqtrade/commands/build_config_commands.py", line 165, in deploy_new_config
    selections['exchange'] = render_template(
  File "/usr/lib/python3.8/site-packages/freqtrade/misc.py", line 149, in render_template
    template = env.get_template(templatefile)
  File "/usr/lib/python3.8/site-packages/jinja2/environment.py", line 883, in get_template
    return self._load_template(name, self.make_globals(globals))
  File "/usr/lib/python3.8/site-packages/jinja2/environment.py", line 857, in _load_template
    template = self.loader.load(self, name, globals)
  File "/usr/lib/python3.8/site-packages/jinja2/loaders.py", line 117, in load
    source, filename, uptodate = self.get_source(environment, name)
  File "/usr/lib/python3.8/site-packages/jinja2/loaders.py", line 299, in get_source
    raise TemplateNotFound(template)
jinja2.exceptions.TemplateNotFound: subtemplates/exchange_generic.j2

  ```

xmatthias:
> I installed the package from AUR,

which AUR package?
`python-freqtrade-git` is heavily outdated and no longer maintained, and i could not find another AUR package.

[...]

PR #5684, opened 2021-10-07
pypi installation

SmartManoj:
Got [this error](https://github.com/freqtrade/freqtrade/issues/3078) when installing from git but not in pypi

As it is closed and most important task finished, https://github.com/freqtrade/freqtrade/issues/2112

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,7 @@ Please find the complete documentation on our [website](https://www.freqtrade.io
 Freqtrade provides a Linux/macOS script to install all dependencies and help you to configure the bot.
 
 ```bash
-git clone -b develop https://github.com/freqtrade/freqtrade.git 
-cd freqtrade
-./setup.sh --install
+pip install freqtrade
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -66,9 +66,7 @@ Please find the complete documentation on our [website](https://www.freqtrade.io
 Freqtrade provides a Linux/macOS script to install all dependencies and help you to configure the bot.
 
 ```bash
-git clone -b develop https://github.com/freqtrade/freqtrade.git 
-cd freqtrade
-./setup.sh --install
+pip install freqtrade
```
xmatthias: the problem is - this will only work if you've already got ta-lib (the c library) installed.
otherwise, it'll fail (either during installation, or latest while running a strategy that requires ta-lib).

This is why i'm very hesitant to recommend this as installation method - as it will lead to more problems than it'll help people with.

Also the problem you link to will not happen when following this guide - so the error you'll have received must be something different.

xmatthias (approved):
I've moved this section to the actual documentation (the readme.md is only visible on github), and moved it to a note (including the warning to have ta-lib installed beforehand).

PR #5710, opened 2021-10-13
Added SELL_TAG for trading, backtesting and telegram

theluxaz:
## Summary

Added sell_tags to compliment buy tags and enabled more usages for them both. 

Solve the issue:  NEW FEATURE

## Quick changelog

- Added Sell tags
- Buy/Sell tag based breakdowns for baktesting details
- Sell tag display for telegram
- 3 New telegram functions for displaying buy/sell tags

## What's new?

Added sell_tag functionality which can be added the same way as buy tags, just within populate_sell_trend, as it didn't seem fitting to put all of your sell rules in custom_sell. Sell tags only work for sell_signal sells and can be declared as such:
```
dataframe.loc[
            (
            ),
            ['sell', 'sell_tag']] = (1, 'LOW')
```
Detailed breakdowns for backtesting were also added to see buy and sell statistics per buy/sell tag:
image

Sell tags were also added to be displayed in telegram wherever possible.
Three additional functions for telegram /buys, /sells, /mix_tags (possibly not useful for others) were added which show how buys and sells perform based on their tags. Mix tags shows how a certain combination of them both performs.
image

From what I've tested it works fine with both buy and sell tags added, however i did not try every case. There is also a bug where an empty sell_tag field shows up, not sure where it comes from.

Sorry for the messy pull request, this is my first commit on Github ever and I wanted to share with others on Freqtrade a thing I developed. I don't know much about how to fix the python tests as my changes break quite a few, especially with the backtesting detail stuff. Would be good to allow users to configure whether to enable the buy/sell tag displays, but I am not sure how to implement that.


freqtrade/rpc/rpc.py
```diff
@@ -686,6 +686,37 @@ def _rpc_performance(self) -> List[Dict[str, Any]]:
...
         return pair_rates
 
+    def _rpc_buy_tag_performance(self, pair: str) -> List[Dict[str, Any]]:
+        """
+        Handler for buy tag performance.
+        Shows a performance statistic from finished trades
+        """
+        buy_tags = Trade.get_buy_tag_performance(pair)
+        # Round and convert to %
+        [x.update({'profit': round(x['profit'] * 100, 2)}) for x in buy_tags]
```
xmatthias: You shouldn't modify list elements within list comprehensions ... it'll probably be better to do the rounding at an earlier stage, maybe providing a "raw" profit value , and a "profit_pct" item to avoid having to iterate this again.

freqtrade/rpc/telegram.py
```diff
@@ -108,6 +108,7 @@ def _init_keyboard(self) -> None:
                                  r'/trades$', r'/performance$', r'/daily$', r'/daily \d+$',
                                  r'/profit$', r'/profit \d+',
                                  r'/stats$', r'/count$', r'/locks$', r'/balance$',
+                                 r'/buys',r'/sells',r'/mix_tags',
```
xmatthias: You should probably end these with `$` (same as the ones the line above), unless they support / require a parameter.

freqtrade/rpc/telegram.py
```diff
@@ -258,6 +265,46 @@ def _format_sell_msg(self, msg: Dict[str, Any]) -> str:
...
+        elif sell_tag is not None and buy_tag is None:
+            message = ("{emoji} *{exchange}:* Selling {pair} (#{trade_id})\n"
+                   "*Profit:* `{profit_percent:.2f}%{profit_extra}`\n"
+                   "*Sell Tag:* `{sell_tag}`\n"
+                   "*Sell Reason:* `{sell_reason}`\n"
+                   "*Duration:* `{duration} ({duration_min:.1f} min)`\n"
+                   "*Amount:* `{amount:.8f}`\n"
+                   "*Open Rate:* `{open_rate:.8f}`\n"
+                   "*Current Rate:* `{current_rate:.8f}`\n"
+                   "*Close Rate:* `{limit:.8f}`").format(**msg)
```
xmatthias: this part MUST be condensed. 
95% of the code is identical, the only difference i see is in 1-2 lines.

best move the sell-tag logic to the top of the function, and build up the additional lines where needed, then combine all strings together before using the `.format(**msg)`. Otherwise this code will drift apart, and some edge-cases either produce failures, or different results (assuming additional functionalities in the future).

freqtrade/enums/signaltype.py
```diff
@@ -14,3 +14,4 @@ class SignalTagType(Enum):
     Enum for signal columns
     """
     BUY_TAG = "buy_tag"
+    SELL_TAG = "sell_tag"
```
xmatthias: can we name the value `exit_tag` please?

While it's not really aligned at this moment to the buy-side - it'll simplify adding short trading (where the "exit" is actually a buy) in the future.

xmatthias (changes requested):
One thing you should definitely look at is the [developer setup](https://www.freqtrade.io/en/stable/developer/#developer-setup) - so you can run the tools running in CI locally.

Especially for flake8, i can see a lot of failures, like too long lines, wrong whitespace (even though they won't show up in CI until tests pass). A good editor (vscode with python plugin, pycharm) can also assist in this, as they'll highlight formatting problems as well.

based on the values - i'd think that your "empty" sell-reason is probably stoploss (i'd expect one for ROI to show up too, if the strategy uses that) - so i'd think that you'll have to add that manually or exclude these values (which i wouldn't do, as then you're hiding some sell-reasons).


One thing you should look at next is `patch_get_signal` (in the tests) - it's now not returning enough values (simply return None by default) - that should fix about half of the 

Overall, it's still going to be quite some work (not just for tests), so i hope you'll have the time and motivation to stick around and make this acceptable.

theluxaz:
Will keep updating this and do another pull request soon. Thank you so much for the responses, it's very helpful!

freqtrade/rpc/telegram.py
xmatthias: looking at this again i THINK it's ok as is, as you seem to support calling `/buys ETH/USDT` to get stats only for this pair.

freqtrade/rpc/rpc.py
theluxaz: I just used the same exact code from the function above for getting "/performance" data. Let me know if that should be changed as well.

theluxaz:
I Have changed all the mentions of 'sell_tag' to 'exit_tag' for the mentioned short trading. Just changing the value of exit_tag would get too confusing if sell_tag was used.

Added the flake8 formatting, let me know if it is acceptable. Fixed the telegram functions which now return the proper data. Need to gather more data to properly test "_format_sell_msg" in telegram.py to make sure it works correctly, will follow-up in a few days.

In regards to "get_signal" returning not enough values. It now expects 4 now and it returns 4 in all the cases I have found, not sure how to fix anything this.

A bug I have noticed is that the exit_tags get wrapped in brackets during backtesting for some reason, unlike the buy_tags. The data is not in brackets when added in what I assume is backesting.py (picture attached). I can't find the reason why that happens.

image

freqtrade/optimize/backtesting.py
```diff
@@ -359,8 +360,10 @@ def _get_sell_trade_entry_for_candle(self, trade: LocalTrade,
 
         if sell.sell_flag:
             trade.close_date = sell_candle_time
-            if(sell_row[SELL_TAG_IDX] is not None):
-                trade.sell_tag = sell_row[SELL_TAG_IDX]
+            if(sell_row[EXIT_TAG_IDX] is not None):
+                trade.exit_tag = sell_row[EXIT_TAG_IDX]
+            else:
+                trade.exit_tag = None
```
theluxaz: Will delete the else statement with the next commit

xmatthias:
while thinking about this a bit more, i ask myself more and more - how will this differ from `sell_reason`?

It's providing the same information - but at a slightly different level (it'll split out sell_signal into different sub-segments), while sell_reason currently only works for custom_sell and roi/stop.

so maybe a better option would be to assign this to sell_reason (instead of a new column), so we can use existing statistics.
otherwise we have 2 sell statistics, the existing sell-reason, and the exit-reason) - which will be highly confusing (when is which used, what are the differences, ...).

theluxaz:
I can try to make this work on sell_reason, is that preferrable? Will it work for the leverage trading later on?

It will be quite a bit more confusing to implement, as for exit_tag I just followed the example of the buy_tag. I just would like to know how to put the label from populate_sell_trend (below) 
```
dataframe.loc[
            (  ),
            ['sell', 'sell_reason']] = (1, 'LOW')
```
onto a trade object without using something like 
`trade.sell_reason=sell_row[EXIT_TAG_IDX] `
with an additional column.

xmatthias:
> I can try to make this work on sell_reason, is that preferrable? Will it work for the leverage trading later on?

Don't see any issues with that, it'll have to merged for sure - but that should be fine.

> It will be quite a bit more confusing to implement, 

I'm not sure about this - in the end, i think it will be simpler - instead of setting "sell_reason" to sell_signal + exit_tag to "whatever" - you're now simply setting `sell_reason=<..>.exit_tag or "sell_signal"` 
it can also use the existing methods, without adding a 2nd table to backtesting output ... While more information is good - having it confusing is not.


> onto a trade object without using something like `trade.sell_reason=sell_row[EXIT_TAG_IDX] ` with an additional column.

What speaks against this? it'll depend on the mode (backtesting/live) - but i don't see a problem with that (and we use exactly this for buy-signals as well).

theluxaz:
I see what you mean, I misunderstood you a little thinking you wanted to get rid of the exit_tag fully. Will rework the exit_tags to work via the sell reason.

theluxaz:
Changed so that the exit_tags are shown in the place of sell_reason where appropriate. Backtesting now shows a single table  with sell reasons and "/sells" now looks like this:

image

xmatthias:
> I see what you mean, I misunderstood you a little thinking you wanted to get rid of the exit_tag fully.

The functionality (assign a tag to the sell signal) is fine - i'm just questioning if we really need it as separate information attached to the trade, or if it can overwrite sell_reason (which is already happening for custom_sells anyway).

theluxaz:
With the last commit, it overwrites the sell_reason and uses that everywhere, however the exit_tag field is still stored in the Trade object. Is this okay or should I try to make it work without the exit_tag in the Trade?

xmatthias:
> With the last commit, it overwrites the sell_reason and uses that everywhere, however the exit_tag field is still stored in the Trade object. Is this okay or should I try to make it work without the exit_tag in the Trade?

I would not store it seperately, as it's duplicate information thats not providing any benfit (it'll be identical to sell_reason when used, and empty in all other cases) - so it's not providing any benefit (and will make the change - and with that also the code we have to maintain in the future) smaller.

theluxaz:
Changed the code so exit_tag is no longer saved within Trade objects. The only place these tags exist are in within dataframe columns where they are set.

xmatthias:
I've fixed a few tests for this (the most common ones) ... 

the remaining should not be too difficult to fix (let me know if you have big problems with a test).

theluxaz:
Thank you, I'm in the middle of fixing the tests, however these two variations keep appearing in the telegram tests. Any idea what they are/ how to fix them?

image

image

What exactly is msg_mock.call_args_list[0][0][0] ?

theluxaz:
It turns out those were caused by an error in backtesting.py with test data

theluxaz:
Fixed all the test I could find (there are still 3 fails due to "tables" optional module in test_hdf5datahandler).

The only issue I can see is users providing trading pairs that not in the whitelists (no trades) as arguments for the new telegram functions, causing an error string in the program. The non-response on telegram seems okay and the program continues running. Not sure if there is a need for a check for the pair before sending through the command and where would that check be made in the code.

image


image

freqtrade/rpc/rpc.py
xmatthias: that might be something to look at in the future, then.
The current implementation creating the list - and right after that, adding a new entry.

Moving the rounding to the proper spot (where the result is created) - and maybe using a different key (so original and `round(x * 100, 2)` are available)  should be preferred here (and also at the bottom).

That way, we can also reuse that logic for API endpoints in the future.

freqtrade/persistence/models.py
```diff
@@ -860,6 +860,141 @@ def get_overall_performance(minutes=None) -> List[Dict[str, Any]]:
...
+        else:
+            tag_perf = Trade.query.with_entities(
+                Trade.buy_tag,
+                func.sum(Trade.close_profit).label('profit_sum'),
+                func.sum(Trade.close_profit_abs).label('profit_sum_abs'),
+                func.count(Trade.pair).label('count')
+            ).filter(Trade.is_open.is_(False))\
+                .group_by(Trade.buy_tag) \
+                .order_by(desc('profit_sum_abs')) \
+                .all()
```
xmatthias: we should be able to simplify this a ton by using the following pattern (look in this file for other samples of this):

``` python
filters = [Trade.is_open.is_(False)]
if(pair is not None):
    filters.append(Trade.pair == pair)

Trade.query.<....>().filter(*filters).group_by()...
```

That'll avoid having the same query twice - which will first of all, reduce the number of lines, but also avoid the 2 potentially drifting apart.

This obviously applies not just to this method, but also to the ones below.

xmatthias (commented):
I think your failing tests are because of a missing dependency on your environment (they pass in CI).

I've added a few more comments where i think code can (and should) be simplified.

I'd also wanna see some tests for this (that'll ensure it's working, and won't break with some future changes).
Maybe look at `test_rpc_telegram` - and copy the test for `/performance` - adjusting it for the new endpoints...

freqtrade/persistence/models.py
theluxaz: Will work on fixing these and adding a few tests next week, thank you for the responses!

theluxaz: I have fixed the rpc and models now. Trying to write some tests, however I cannot find where the data is coming from. In this copied example from /performance in test_rpc.py, how can I setup/edit the data to make assertions on at the end?

image

xmatthias: you can either use `enter_positions` (as above) - which will open the same trades for each test session - or `create_mock_trades` - which will create a static set of trades you can assert with.

theluxaz: I Added data for testing in create_mock_trades_tags() which retrieves the trade data, however the trade data is not received by the rpc. functions. In this example using create_mock_trades works, but the function I added does not and shows an empty response, any idea why? Is there something to do with the mock trades I added?

image

xmatthias: Do i understand this correctly, the new trades only difference is that they're adding buy_tag an sell_tag (where applicable)?
If so, I'd add this to (some) of the existing mock tags - it may break a handful of tests (but shouldn't break many) - but it'll reduce the unnecessary duplication we have otherwise.

theluxaz: Good idea, this should be a lot easier.

theluxaz: I have added the tests that should cover the rpc/telegram interactions for the new functions. Is this okay? 

Also does anything need to be changed for btanalysis? I see it has a lot of column names, should "exit_tag" be included there?

theluxaz: The only other test I can think of is doublechecking the setting of a custom sell_reason during live/dry-run and during backtesting. But I have no idea where or how to test that.

xmatthias: > Also does anything need to be changed for btanalysis? I see it has a lot of column names, should "exit_tag" be included there?

i don't think so. used to convert database records to dataframes ... since we're storing the information in sell_reason - there's little/no reason to modify this.

I also think i've added the tests for live/dry you asked for (check the last commit ...)

To me, this looks now pretty much complete for the moment ... unless you think / feel otherwise?

theluxaz: Looks good to me :) Thank you for your help!

xmatthias (approved):
LGTM :+1:

PR #5711, opened 2021-10-14
Funding fee dry-run

samgermain:
Add formula's to calculate funding-fees, methods to get mark price and funding rate history


samgermain:
So I think this is mostly it, I also added the functionality to download the mark price candlesticks using the same OHLCV methods. 

I want to know if this is the right way to do things before I start writing tests

freqtrade/exchange/exchange.py
```diff
@@ -1411,11 +1459,15 @@ async def _async_get_candle_history(self, pair: str, timeframe: str,
...
-                                                     params=params)
+            # TODO-lev: Does this put price into params correctly?
+            params = self._ft_has.get('ohlcv_params', {price: price})
+            data = await self._api_async.fetch_ohlcv(
+                pair,
+                timeframe=timeframe,
+                since=since_ms,
+                limit=self.ohlcv_candle_limit(timeframe),
+                params=params
+            )
```
xmatthias, line 1470: I've not seen nor used this before.

Please add a link to the relevant ccxt documentation for this (which assumes it's unified across all exchanges).
If it's not unified, i don't think this is the right place.

In general, i'm very sceptical of 2 things here:
* overloading the same method for different functions
* combining things that don't belong together into one PR.

so unless this is necessary to get dry-run going - i'd seperate it out of this pr.

samgermain:
Closed because branch is being renamed

PR #5732, opened 2021-10-17
New max lev

samgermain:
I added leverage as a property to markets inside CCXT. This new method of getting the max leverage will work for all freqtrade approved exchanges, except Binance, because Binance has leverage brackets so I left Binance's get_max_leverage as is


samgermain:
@xmatthias I don't know how to get the failing tests to work. I added the `limits: leverage` property to `conftest get_markets`but I'm not sure if those markets are being loaded in

When I run the debugger, I just see MagicMock for the markets, and don't see any of the actual markets

xmatthias:
look at the return-value of magicmock (should be available as `mock.return_value`) - as you're adding this to ccxt, you'll need to add it to the markets-mock in conftest ... but i'd not add it to all markets - that way we can verify it behaves as desired when this entry is missing for some reason.

samgermain:
> look at the return-value of magicmock (should be available as `mock.return_value`) - as you're adding this to ccxt, you'll need to add it to the markets-mock in conftest ... but i'd not add it to all markets - that way we can verify it behaves as desired when this entry is missing for some reason.

nvm, it was there, I don't know why i didn't see it before

samgermain:
This is the whole PR by the way

freqtrade/exchange/exchange.py
```diff
@@ -1631,9 +1627,9 @@ def get_funding_fees_from_exchange(self, pair: str, since: Union[datetime, int])
 
     def fill_leverage_brackets(self):
         """
-            # TODO-lev: Should maybe be renamed, leverage_brackets might not be accurate for kraken
             Assigns property _leverage_brackets to a dictionary of information about the leverage
             allowed on each pair
+            Not used by most exchanges, only used by Binance at time of writing
```
xmatthias: I think we should be decisive here - either this is binance-specific - so we also remove this method in the parent exchange class (here) - or we say it's "exchange opt in" - so we still call it from the main exchange (which makes the custom constructor in `binance.py` unnecessary).

xmatthias (approved):
:+1:

PR #5756, opened 2021-10-19
add custom entry/exit price support to backtesting

### LINKED ISSUE #5755: Support for custom entry-prices and exit-prices during backtesting.
Custom order entry and exit price is good.
and i can set the trade price directly. 

This thing work only in "trade" mode. but not in "backtesting"

GluTbl:
Support for custom entry-prices and exit-prices during backtesting.

## Summary

Support for custom entry-prices and exit-prices during backtesting.

Closes #5755 

## Quick changelog

- edited `_enter_trade()` in backtesting.py
- edited `_get_sell_trade_entry_for_candle()` in backtesting.py


xmatthias (changes requested):
In general, i'm not against supporing dynamic pricing for backtesting - however we must also help protect users from mistakes / false hopes (even though it's the user's responsibility, we've now reached a size where people start to blindly run shared strategies without fully understanding them).

This change will allow me to (backtest-)buy BTC for 4k$ (current rate is ~60k$) - which will make me very profitable - but is WAY off from reality.

The minimum "sanity" checking logic we'll require here is to ensure that the price is within the current candle (which in itself is a partial lookahead bias...).

Prices far of the range should either result in rejected trades (quite some more work, as we'd want to track this as seperate metric?) - or should move the price to the candle's low/high value (potentially unwanted behaviour - as we're modifying the user's desired price).

xmatthias (approved):
I've fixed the outstanding points in this PR now.

This means, prices will be moved to be within the selling candle during backtesting - which is also documented this way.

LGTM :+1:

PR #5763, opened 2021-10-20
Test ccxt config

samgermain:
I noticed that this computed variable wasn't tested
Just close this PR if you think this test is unnecessary

### DIFF
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -3240,3 +3240,30 @@ def test_validate_trading_mode_and_collateral(
             exchange.validate_trading_mode_and_collateral(trading_mode, collateral)
     else:
         exchange.validate_trading_mode_and_collateral(trading_mode, collateral)
+
+
+@pytest.mark.parametrize("exchange_name,trading_mode,ccxt_config", [
+    ("binance", "spot", {}),
+    ("binance", "margin", {"options": {"defaultType": "margin"}}),
+    ("binance", "futures", {"options": {"defaultType": "future"}}),
+    ("kraken", "spot", {}),
+    ("kraken", "margin", {}),
+    ("kraken", "futures", {}),
+    ("ftx", "spot", {}),
+    ("ftx", "margin", {}),
+    ("ftx", "futures", {}),
+    ("bittrex", "spot", {}),
+    ("bittrex", "margin", {}),
+    ("bittrex", "futures", {}),
+])
+def test__ccxt_config(
+    default_conf,
+    mocker,
+    exchange_name,
+    trading_mode,
+    ccxt_config
+):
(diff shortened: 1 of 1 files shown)


tests/exchange/test_exchange.py
```diff
@@ -3240,3 +3240,30 @@ def test_validate_trading_mode_and_collateral(
...
         exchange.validate_trading_mode_and_collateral(trading_mode, collateral)
+
+
+@pytest.mark.parametrize("exchange_name,trading_mode,ccxt_config", [
+    ("binance", "spot", {}),
+    ("binance", "margin", {"options": {"defaultType": "margin"}}),
+    ("binance", "futures", {"options": {"defaultType": "future"}}),
+    ("kraken", "spot", {}),
+    ("kraken", "margin", {}),
+    ("kraken", "futures", {}),
```
xmatthias, line 3251: I do like the fact to test it ... :+1: 

but it also raises a followup question (probably not to be completed in this PR) - how will we pick the right "futures/margin" pairs for kraken/ftx ?

samgermain: > I do like the fact to test it ... 👍
> 
> but it also raises a followup question (probably not to be completed in this PR) - how will we pick the right "futures/margin" pairs for kraken/ftx ?

What are you referring to exactly?

For ftx, CCXT doesn't require `defaultType: future`, Kraken futures isn't implemented on CCXT yet (will probably do that in the next few weeks/months) so I'm not sure how configuration for kraken will work exactly. Once it is implemented, I'lll change this if I need to, `defaultType: margin` is not required to for Kraken on CCXT

xmatthias: it's not necessarily related to the test ... just a thought that jumped me due to this test.

assuming i'm on FTX (let's ignore kraken, we don't know yet) ... and i pick "futures" ... how will freqtrade know which pairs i can trade (which pairs are futures pairs) and which ones ain't?

samgermain: > it's not necessarily related to the test ... just a thought that jumped me due to this test.
> 
> assuming i'm on FTX (let's ignore kraken, we don't know yet) ... and i pick "futures" ... how will freqtrade know which pairs i can trade (which pairs are futures pairs) and which ones ain't?

I think we're currently working on this at CCXT, (@frosty00 any input? I couldn't @kroitor), but currently I think that we could filter for "futures" markets by checking `market.swap == True`, if we can't, then I need to add that to CCXT

xmatthias: Just a question / something to think about (and i had to write it down so i'm not the only one to know about it) :laughing: 

PR itself looks good :+1:

xmatthias (approved):

PR #5772, opened 2021-10-22
Add method unlock_reason(reason: str) to PairLocks

ghost:
## Summary

Add function to release all locks with a certain reason.

Solve the issue: New Feature

## What's new?
Might be useful in some cases to lock and unlock pairs for a certain reason from inside the strategy
For example you could lock pairs with:
                self.lock_pair(pair, until=datetime.now(timezone.utc) + timedelta(minutes=10), reason='Because reason')
and later unlock all pairs with this reason by calling:
            self.unlock_reason('Because reason')

Add this code to pairlock_middleware.py:
```
    @staticmethod
    def unlock_reason(reason: str, now: Optional[datetime] = None) -> None:
        """
        Release all locks for this reason.
        :param reason: Which reason to unlock
        :param now: Datetime object (generated via datetime.now(timezone.utc)).
            defaults to datetime.now(timezone.utc)
        """
        if not now:
            now = datetime.now(timezone.utc)

        logger.info(f"Releasing all locks with reason \'{reason}\'.")
        locks = PairLocks.get_all_locks()
        for lock in locks:
            if lock.reason == reason:
                lock.active = False
        if PairLocks.use_db:
            PairLock.query.session.commit()
```
Add this code to interface.py:
```
    def unlock_reason(self, reason: str) -> None:
        """
        Unlocks a pair previously locked using lock_pair.
        Not used by freqtrade itself, but intended to be used if users lock pairs
        manually from within the strategy, to allow an easy way to unlock pairs.
        :param pair: Unlock pair to allow trading again
        """
        PairLocks.unlock_reason(reason, datetime.now(timezone.utc)
```

*i still have to figure out how github works, sorry


xmatthias:
Your pull request is clearly wrong, and not what you intended to do.
You're trying to merge freqtrade's develop branch into stable. 
This will happen with the next release - but it's clearly not what you intended.

Best follow [this tutorial](https://guides.github.com/introduction/flow/) to learn how to create a proper pull request (in short: You'll have to fork freqtrade, create a branch there, commit your changes, and then submit a pull request against this repositori's develop branch).
The github tutorial has the detailed steps for this.

As is, it's impossible to merge this (you can't even really fix this as both source and target branches are wrong - so the best thing is to create a new one, following the guide linked above).

PR #5773, opened 2021-10-22
Gateio futures

samgermain:
Adds gateio futures support

I think this is all it would take to add this exchange, plus the corresponding liquidation formula

### DIFF
--- a/freqtrade/exchange/gateio.py
+++ b/freqtrade/exchange/gateio.py
@@ -1,7 +1,8 @@
 """ Gate.io exchange subclass """
 import logging
-from typing import Dict, List
+from typing import Dict, List, Tuple
 
+from freqtrade.enums import Collateral, TradingMode
 from freqtrade.exceptions import OperationalException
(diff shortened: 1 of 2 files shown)


xmatthias:
> I think this is all it would take to add this exchange, plus the corresponding liquidation formula

i'm not sure - how is leverage set? via set_leverage, or dynamic by trade (basically the kraken method)?

I'm also not sue if this means we should be merging it now - or if you intend to add the liquidation formula here too...

samgermain:
> i'm not sure - how is leverage set? via set_leverage, or dynamic by trade (basically the kraken method)?

It will be via the set_leverage method, 

> I'm also not sue if this means we should be merging it now - or if you intend to add the liquidation formula here too...

My plan is to write the liquidation formula on the PR for the isolated liquidation implementation, so with the binance isolated liq method

xmatthias (approved):
Allright then :+1:

PR #5776, opened 2021-10-23
Improve performance of decimalspace.py

SimonEbner:
## Summary

Improve performance for decimalspace.py. Makes it slightly less readable, however, since that code is heavily called in optimization it seems a fair trade-off in my opinion.

## Quick changelog

* Cache `pow(10, decimals)` and `pow(.1, decimals)`
* Minor optimization around (surprisingly slow) `round`

## What's new?

decimalspace.py is heavily used in the hyperoptimization. The following
benchmark code runs an optimization which is taken from optimizing a
real strategy (wtc).
The optimized version takes on my machine approx. 11/12s compared to the
original 32s. Results are equivalent in both cases.

```
import freqtrade.optimize.space
import numpy as np
import skopt
import timeit

def init():
    Decimal = freqtrade.optimize.space.decimalspace.SKDecimal
    dimensions = [Decimal(low=-1.0,
        high=1.0,
        decimals=4,
        prior='uniform',
        transform='identity')] * 20

    return skopt.Optimizer(
        dimensions,
        base_estimator="ET",
        acq_optimizer="auto",
        n_initial_points=5,
        acq_optimizer_kwargs={'n_jobs': 96},
        random_state=0,
        model_queue_size=10,
    )

def test():
    opt = init()
    actual = opt.ask(n_points=2)
    expected = [[
        0.7515, -0.4723, -0.6941, -0.7988, 0.0448, 0.8605, -0.108, 0.5399,
        0.763, -0.2948, 0.8345, -0.7683, 0.7077, -0.2478, -0.333, 0.8575,
        0.6108, 0.4514, 0.5982, 0.3506
    ], [
        0.5563, 0.7386, -0.6407, 0.9073, -0.5211, -0.8167, -0.3771,
        -0.0318, 0.2861, 0.1176, 0.0943, -0.6077, -0.9317, -0.5372,
        -0.4934, -0.3637, -0.8035, -0.8627, -0.5399, 0.6036
    ]]

    absdiff = np.max(np.abs(np.asarray(expected) - np.asarray(actual)))
    assert absdiff < 1e-5

def time():
    opt = init()
    print('dt', timeit.timeit("opt.ask(n_points=20)", globals=locals()))

if __name__ == "__main__":
    test()
    time()
```

### DIFF
--- a/freqtrade/optimize/space/decimalspace.py
+++ b/freqtrade/optimize/space/decimalspace.py
@@ -7,11 +7,15 @@ class SKDecimal(Integer):
     def __init__(self, low, high, decimals=3, prior="uniform", base=10, transform=None,
                  name=None, dtype=np.int64):
         self.decimals = decimals
-        _low = int(low * pow(10, self.decimals))
-        _high = int(high * pow(10, self.decimals))
+
+        self.pow_dot_one = pow(0.1, self.decimals)
+        self.pow_ten = pow(10, self.decimals)
+
+        _low = int(low * self.pow_ten)
+        _high = int(high * self.pow_ten)
         # trunc to precision to avoid points out of space
-        self.low_orig = round(_low * pow(0.1, self.decimals), self.decimals)
-        self.high_orig = round(_high * pow(0.1, self.decimals), self.decimals)
+        self.low_orig = round(_low * self.pow_dot_one, self.decimals)
+        self.high_orig = round(_high * self.pow_dot_one, self.decimals)
 
         super().__init__(_low, _high, prior, base, transform, name, dtype)
 
@@ -25,9 +29,9 @@ class SKDecimal(Integer):
         return self.low_orig <= point <= self.high_orig
 
     def transform(self, Xt):
-        aa = [int(x * pow(10, self.decimals)) for x in Xt]
-        return super().transform(aa)
+        return super().transform([int(v * self.pow_ten) for v in Xt])
 
     def inverse_transform(self, Xt):
         res = super().inverse_transform(Xt)
-        return [round(x * pow(0.1, self.decimals), self.decimals) for x in res]
+        # equivalent to [round(x * pow(0.1, self.decimals), self.decimals) for x in res]
+        return [int(v) / self.pow_ten for v in res]


SimonEbner:
I see that the following test case fails
```
space.inverse_transform([150, 160]) == [1.5, 1.6]
```

It seems to be due to me changing the return type of `inverse_transform` in decimalspace.py from python list to np.array. Let me know which ones is more suitable and I can add annotations + proper conversion or adjustments to the test case

freqtrade/optimize/space/decimalspace.py
```diff
@@ -25,9 +29,17 @@ def __contains__(self, point):
         return self.low_orig <= point <= self.high_orig
 
     def transform(self, Xt):
-        aa = [int(x * pow(10, self.decimals)) for x in Xt]
+        aa = np.empty_like(Xt, dtype=np.int64)
+        for i, v in enumerate(Xt):
+            aa[i] = int(v * self.pow_ten)
```
xmatthias: out of curiosity only:
Did you also experiment / try with the middle-ground of this refactoring? (using `self.pow_ten`  in the list comprehension)? 

I find it rather curious that a "plain and simple" loop is faster than a list-comprehension - which in my understanding and previous resarch, is the faster way than a "dump" for loop, but maybe the memory allocation is the bottleneck here, which would explain this.

xmatthias (commented):
I'd only change the return type if it doesn't impact the gained performance (which i doubt you'll be able to accomplish).

I've not seen any difference (the returned value still has the same datatype) - so this seems pretty ... non-invasive to me (other than speeding things up a ton).

So it'll be adjusting tests ...(i'm not sure about anotations, as this class is not type-anotated (half-intentional, as it's just a dump helper-method for shortcommings in scikit-optimize ...).

freqtrade/optimize/space/decimalspace.py
SimonEbner: Turns out you're right. Going back to the list comprehension is actually a bit faster than the loop. At some point, I measured it to be beneficial, but I assume it was overshadowed by the other things

SimonEbner:
Getting an error in test `test_api_logs`:
```
>       assert rc1.json()['log_count'] > 2
E       assert 0 > 2
```
Seems unrelated, can you advice on appropriate action?

freqtrade/optimize/space/decimalspace.py
xmatthias: time measurements can be ... troublesome to get right - as our pc's are always doing "something else" as well :laughing:

xmatthias (approved):
> Seems unrelated, can you advice on appropriate action?

ignore - it's a fluke i'm not sure why (it's not happening all the time, and rerunning ci locally with the same random value will not have this show) - rerunning the tests fixes that problem.

LGTM :+1:

PR #5777, opened 2021-10-23
Use pathlib.stem instead of str(x).ends_with

SimonEbner:
## Summary

Minor refactoring, use pathlib.stem instead of str(x).ends_with


xmatthias (approved):
nice :+1:

PR #5779, opened 2021-10-24
Funding fee dry run

samgermain:
Adds dry run capability for funding-fee


samgermain:
@xmatthias You've mentioned this before, if I wanted to try to replicate the failing build, then I would the following on a machine with ubuntu 20.04 running python3.9.7?

```
pytest --random-order-bucket=module --random-order-seed=442665
```

xmatthias:
you'll also need `--longrun` - as the error happened in the ccxt_compat file (which is not run by default).

Failures there can however usually be resolved by rerunning CI a little later - as these do NOT use mocks, but query the (public) exchange endpoitns for real - which can cause errors in case of exchange downtimes.

samgermain:
@xmatthias Just waiting for your insight into this PR before writing the tests for it

freqtrade/exchange/exchange.py
```diff
@@ -1684,11 +1702,24 @@ def _set_leverage(
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 
+    def _get_funding_fee_dates(self, d1, d2):
+        d1 = datetime(d1.year, d1.month, d1.day, d1.hour)
```
xmatthias: isn't this wrong?

assuming i open a trade at 10:30 , and close it at 14:30 ... then this would result in the following times:

`10:00, 11:00, 12:00, 13:00, 14:00`

in my understanding, the 10:00 one is wrong (i wasn't in a position for another half an hour ...)

freqtrade/exchange/exchange.py
```diff
@@ -1704,6 +1735,108 @@ def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
...
+    ) -> Dict:
+        '''
+            :param pair: quote/base currency pair
+            :param start: timestamp in ms of the beginning time
+            :param end: timestamp in ms of the end time
+        '''
+        if not self.exchange_has("fetchFundingRateHistory"):
+            raise ExchangeError(
+                f"CCXT has not implemented fetchFundingRateHistory for {self.name}; "
+                f"therefore, backtesting for {self.name} is currently unavailable"
```
xmatthias: I'm not sure I'd mention backtesting here - i assume this will also be used / needed for dry-run?

freqtrade/exchange/gateio.py
```diff
@@ -59,3 +59,16 @@ def validate_ordertypes(self, order_types: Dict) -> None:
...
+        self,
+        pair: str,
+        start: int,
+        end: Optional[int] = None
+    ) -> Dict:
+        '''
+            :param start: timestamp in ms of the beginning time
+            :param end: timestamp in ms of the end time
+        '''
+        # TODO-lev: Has a max limit into the past of 333 days
```
xmatthias: is this documented as part of the API endpoint, or is it because they only launched futures 333 days ago?

freqtrade/exchange/bybit.py
```diff
@@ -30,3 +30,4 @@ class Bybit(Exchange):
         # (TradingMode.FUTURES, Collateral.CROSS),  # TODO-lev: Uncomment once supported
         # (TradingMode.FUTURES, Collateral.ISOLATED) # TODO-lev: Uncomment once supported
     ]
+    name_for_futures_market = 'linear'
```
xmatthias: what's the (intended) use of this (i don't see it used ATM)?

freqtrade/exchange/exchange.py
```diff
@@ -1704,6 +1735,108 @@ def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
...
+        start: int,
+        end: Optional[int]
+    ) -> Dict:
+        """
+            Get's the mark price history for a pair
+        """
+        if end:
+            params = {
+                'endTime': end
+            }
```
xmatthias: the "end" is not exposed by ccxt? 

are we sure we need the end-date? (we'll only need it "up to now" - at least for dry-run) ?

freqtrade/exchange/gateio.py
samgermain: You can query a max of 1000 funding rates, and there's no option to specify a start date. They have 3 funding rates a day, so you can only go back 333 days

freqtrade/exchange/bybit.py
samgermain: it's not needed

freqtrade/exchange/exchange.py
samgermain: i think i fixed it

freqtrade/exchange/exchange.py
samgermain: I'll get back to you on this

samgermain: i took it out

samgermain:
@xmatthias So overall, do you think everything in this PR is done in a way that won't be problematic, and I could start writing tests for it?

freqtrade/exchange/ftx.py
```diff
@@ -28,6 +28,7 @@ class Ftx(Exchange):
         # (TradingMode.MARGIN, Collateral.CROSS),  # TODO-lev: Uncomment once supported
         # (TradingMode.FUTURES, Collateral.CROSS)  # TODO-lev: Uncomment once supported
     ]
+    mark_ohlcv_price = 'index'
```
xmatthias: I think this might be something to be moved into `_ft_has`
it's an exchange-specific attribute ... and i think it's better to have all of these combined in one sub-structure (that also allows easier testing of non-officially supported exchanges ...)

xmatthias:
> @xmatthias So overall, do you think everything in this PR is done in a way that won't be problematic, and I could start writing tests for it?

yes for sure

as this is a public api endpoint, maybe also try and write tests in `test_ccxt_compat.py` (if you won't let me know, and i'll have a look at this) - these go actively against the exchange and require a specific flag (`--longrun`) to actually run ... but they do point out issues early in some cases.

freqtrade/exchange/ftx.py
xmatthias: Another question - won't we need both this and #5780 for proper dry-run (Assuming we support "shutdown" as well (so we need to catch-up)?

now obviously, we can do these 1 after the other (no support for shutdown for the moment) - it's more an understanding question than anything else

samgermain: > Another question - won't we need both this and #5780 for proper dry-run (Assuming we support "shutdown" as well (so we need to catch-up)?

No, #5780 is just supposed to be an addition to download candles with the same methods.

The amount of candles needed for dry run is so small because it's on the 1h, that if it were shut down for a week, that would be 168/pair, that didn't seem like an amount of data that was worth optimizing or saving, and that it could just be download on the spot

samgermain: i could also limit it to just every 4h, since ftx isn't available now anyway, and every other exchange that I've looked at (just the ft classes right now) isn't more common than this

xmatthias: my point is - if we reuse the methods for downloading historic data - why don't we reuse these methods also for this downloads?
otherwise we'll have the same code (the API call is the same, except for the "start" date) twice ...

samgermain: Yeah I guess we could try to reuse some of those methods, I haven't really analyzed them in depth, I just wanted to make it easy to download the mark and index price candles using them

samgermain:
> as this is a public api endpoint, maybe also try and write tests in `test_ccxt_compat.py` (if you won't let me know, and i'll have a look at this) - these go actively against the exchange and require a specific flag (`--longrun`) to actually run ... but they do point out issues early in some cases.

I'll take a look at when I have some time, I'll be finishing up everything else on this PR first

freqtrade/exchange/binance.py
```diff
@@ -227,3 +228,11 @@ async def _async_get_historic_ohlcv(self, pair: str, timeframe: str,
...
         return await super()._async_get_historic_ohlcv(
             pair=pair, timeframe=timeframe, since_ms=since_ms, is_new_pair=is_new_pair)
+
+    def funding_fee_cutoff(self, d: datetime):
+        '''
+            # TODO-lev: Double check that gateio, ftx, and kraken don't also have this
+            :param d: The open date for a trade
+            :return: The cutoff open time for when a funding fee is charged
+        '''
+        return d.minute > 0 or (d.minute == 0 and d.second > 15)
```
xmatthias: Do i understand this correctly - that binance still applies the fee if the trade is opened within the first 15 seconds of the "funding" hour? 
won't we need to be careful here as it'll not apply to 13:00, but only to 12:00, 16:00, ... as binance only applies fees every 4 hours? (could be wrong though)

samgermain: Ok, I added a 1 am test, and yeah that's correct. I emailed the other 3 exchanges seeing if they had something similar

tests/exchange/test_exchange.py
```diff
@@ -3288,3 +3288,316 @@ def test_get_max_leverage(default_conf, mocker, pair, nominal_value, max_lev):
...
+    mocker,
+    default_conf,
+    funding_rate_history,
+    mark_ohlcv
+):
+    api_mock = MagicMock()
+    api_mock.fetch_ohlcv = MagicMock(return_value=mark_ohlcv)
+    api_mock.fetch_funding_rate_history = MagicMock(return_value=funding_rate_history)
+    datetime = MagicMock()
+    datetime.now = MagicMock()
```
samgermain: I'm trying to test that this branch get's executed

```
if not close_date:
            close_date = datetime.now(timezone.utc)
```

But I don't know what the best way to add this magic mock to `get_patched_exchange` would be

samgermain: Also, I'm getting this error

```
    def test_calculate_funding_fees_datetime_called(
        mocker,
        default_conf,
        funding_rate_history,
        mark_ohlcv
    ):
        api_mock = MagicMock()
        api_mock.fetch_ohlcv = MagicMock(return_value=mark_ohlcv)
        api_mock.fetch_funding_rate_history = MagicMock(return_value=funding_rate_history)
        datetime = MagicMock()
        datetime.now = MagicMock()
        type(api_mock).has = PropertyMock(return_value={'fetchOHLCV': True})
        type(api_mock).has = PropertyMock(return_value={'fetchFundingRateHistory': True})
    
        # TODO-lev: Add datetime MagicMock
        exchange = get_patched_exchange(mocker, default_conf, api_mock)
>       exchange.calculate_funding_fees('ADA/USDT', 30.0, datetime("2021-09-01 00:00:00"))
E       TypeError: calculate_funding_fees() missing 1 required positional argument: 'close_date'
```

which doesn't make any sense because close_date is optional

```
 def calculate_funding_fees(
        self,
        pair: str,
        amount: float,
        open_date: datetime,
        close_date: Optional[datetime]
    ) -> float:
```

xmatthias: you did define `close_date` as Optional - but without default attribute - which means you'll have to explicitly pass `None` as that parameter. What you probably wanted to do is `close_date: Optional[datetime] = None` ... ?

xmatthias: As for the test - don't "magicMock" this, but use time-machine (if you rely on the date you're receiving).

Alternatively, ensure that "current time" will work for the test by moving other dates into the past/future (we do this in other tests, specifically with pairLocks ...)

samgermain:
I'm currently getting this error on `test_get_funding_rate_history` and `test_get_mark_price_history` and don't know how to solve them. It's an error when calling `ccxt_exceptionhandlers`, and it's the same error for both tests

```
    def test_get_funding_rate_history(mocker, default_conf, funding_rate_history):
        api_mock = MagicMock()
        api_mock.fetch_funding_rate_history = MagicMock(return_value=funding_rate_history)
        type(api_mock).has = PropertyMock(return_value={'fetchFundingRateHistory': True})
    
        # mocker.patch('freqtrade.exchange.Exchange.get_funding_fees', lambda pair, since: y)
        exchange = get_patched_exchange(mocker, default_conf, api_mock)
        funding_rates = exchange.get_funding_rate_history('ADA/USDT', 1635580800001)
    
        assert funding_rates == {
            1630454400000: -0.000008,
            1630458000000: -0.000004,
            1630461600000: 0.000012,
            1630465200000: -0.000003,
            1630468800000: -0.000007,
            1630472400000: 0.000003,
            1630476000000: 0.000019,
            1630479600000: 0.000003,
            1630483200000: -0.000003,
            1630486800000: 0,
            1630490400000: 0.000013,
            1630494000000: 0.000077,
            1630497600000: 0.000072,
            1630501200000: 0.000097,
        }
    
>       ccxt_exceptionhandlers(
            mocker,
            default_conf,
            api_mock,
            "binance",
            "get_funding_rate_history",
            "fetch_funding_rate_history",
            pair="ADA/USDT",
            since=1630454400000
        )

tests/exchange/test_exchange.py:3491: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

mocker = <pytest_mock.plugin.MockerFixture object at 0x12aa8b3a0>
default_conf = {'ask_strategy': {'order_book_top': 1, 'use_order_book': False}, 'bid_strategy': {'ask_last_balance': 0.0, 'check_dept...cel_open_orders_on_exit': False, 'datadir': '/Users/sam/Documents/dev/trading/freqtrade/freqtrade/tests/testdata', ...}
api_mock = <MagicMock id='5014128384'>, exchange_name = 'binance', fun = 'get_funding_rate_history', mock_ccxt_fun = 'fetch_funding_rate_history'
retries = 5, kwargs = {'pair': 'ADA/USDT', 'since': 1630454400000}, exchange = <freqtrade.exchange.binance.Binance object at 0x12a9749d0>
@py_assert0 = <MagicMock id='5003950544'>, @py_assert2 = 1, @py_assert4 = False
@py_format6 = "1\n{1 = <MagicMock id='5003950544'>.call_count\n} == 5", @py_format8 = "assert 1\n{1 = <MagicMock id='5003950544'>.call_count\n} == 5"

    def ccxt_exceptionhandlers(mocker, default_conf, api_mock, exchange_name,
                               fun, mock_ccxt_fun, retries=API_RETRY_COUNT + 1, **kwargs):
    
        with patch('freqtrade.exchange.common.time.sleep'):
            with pytest.raises(DDosProtection):
                api_mock.__dict__[mock_ccxt_fun] = MagicMock(side_effect=ccxt.DDoSProtection("DDos"))
                exchange = get_patched_exchange(mocker, default_conf, api_mock, id=exchange_name)
                getattr(exchange, fun)(**kwargs)
>           assert api_mock.__dict__[mock_ccxt_fun].call_count == retries
E           AssertionError: assert 1 == 5
E            +  where 1 = <MagicMock id='5003950544'>.call_count

tests/exchange/test_exchange.py:39: AssertionError
----------------------------------------------------------------- Captured log call ------------------------------------------------------------------
INFO     freqtrade.exchange.exchange:exchange.py:118 Instance is running with dry_run enabled
INFO     freqtrade.exchange.exchange:exchange.py:119 Using CCXT 1.58.47
INFO     freqtrade.exchange.exchange:exchange.py:162 Using Exchange "Binance"
INFO     freqtrade.resolvers.exchange_resolver:exchange_resolver.py:56 Using resolved exchange 'Binance'...
INFO     freqtrade.exchange.exchange:exchange.py:118 Instance is running with dry_run enabled
INFO     freqtrade.exchange.exchange:exchange.py:119 Using CCXT 1.58.47
INFO     freqtrade.exchange.exchange:exchange.py:162 Using Exchange "Binance"
INFO     freqtrade.resolvers.exchange_resolver:exchange_resolver.py:56 Using resolved exchange 'Binance'...
```

freqtrade/freqtradebot.py
```diff
@@ -269,10 +269,17 @@ def get_free_open_trades(self) -> int:
     def update_funding_fees(self):
         if self.trading_mode == TradingMode.FUTURES:
             for trade in Trade.get_open_trades():
-                funding_fees = self.exchange.get_funding_fees_from_exchange(
-                    trade.pair,
-                    trade.open_date
-                )
+                if self.config['dry_run']:
+                    funding_fees = self.exchange.calculate_funding_fees(
+                        trade.pair,
```
samgermain: @xmatthias I want to test that these are called, how can I add a few trades that will be returned when `Trade.get_open_trades()` is called?

xmatthias: `create_mock_trades`?

xmatthias:
> I'm currently getting this error on `test_get_funding_rate_history` and `test_get_mark_price_history` and don't know how to solve them. It's an error when calling `ccxt_exceptionhandlers`, and it's the same error for both tests

missing the `@retrier` decorator? (from looking at the exception log in an issue comment ...)

freqtrade/freqtradebot.py
samgermain: All right, well I think this is the last thing I have to do for this PR, and then maybe also `test_ccxt_compat`

tests/test_freqtradebot.py
```diff
@@ -4698,3 +4698,95 @@ def test_update_funding_fees(mocker, default_conf, trading_mode, calls, time_mac
...
+    default_conf['collateral'] = 'isolated'
+    default_conf['dry_run'] = True
+    freqtrade = get_patched_freqtradebot(mocker, default_conf)
+    create_mock_trades(fee, False)
+    time_machine.move_to("2021-09-01 08:00:00 +00:00")
+    freqtrade._schedule.run_pending()
+
+    trades = Trade.get_open_trades()
+    for trade in trades:
+        assert trade.funding_fees == 123 * mark_prices[trade.pair] * funding_rates[trade.pair]
```
samgermain: @xmatthias @rokups  When I run `get_open_trades`, I get these 4 `LTC/BTC`, `ETH/BTC`, `ETC/BTC`, `XRP/BTC`, so I'm trying to test that the funding fees are what they are supposed to be after 8 hours

Currently I'm getting the error below, do you understand what I have to do for this trade to work

```
self = <MagicMock name='get_funding_rate_history' id='4900146912'>
args = ('ETH/BTC', 1630496580283), kwargs = {}
effect = <list_iterator object at 0x12412bdc0>

        effect = self.side_effect
        if effect is not None:
            if _is_exception(effect):
                raise effect
            elif not _callable(effect):
E               StopIteration

/usr/local/Cellar/python@3.9/3.9.7_1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/unittest/mock.py:1154: StopIteration
```

rokups: I think you call `get_funding_rate_history()` more times than there are mocked values provided

xmatthias: That's correct - whenever you mock with `side_effect=[` (side-effect is an array) - that array must be at least as long as the amount of calls you're going to make.
That doesn't seem to be the case here, at least that's what the error suggests.

I'd probably not use `side_effect` this way here - but rather write a (test-local) method returning the correct value based on the pair passed in (so you're not relying on the order of the pairs - which i don't think is guaranteed the way we query trades - but is not a problem either).
We're doing this elsewhere too (iirc in backtesting tests) - which might serve as an example.

samgermain: > test-local) method returning the correct value based on the pair passed in 

I dont think I know what that is

rokups: I changed side_effect to lambdas. There is still an issue with assert as it tries to multiply dicts. It probably should be this:

```py
assert trade.funding_fees == 123 * mark_prices[trade.pair][arrow.utcnow().int_timestamp * 1000] * funding_rates[trade.pair][arrow.utcnow().int_timestamp * 1000]
```

But im not entirely sure so i did leave it out. Even if this bit is correct assert still fails.

samgermain: thanks

samgermain:
I don't think the open date of the trades is in UTC, when I use `    time_machine.move_to("2021-09-01 00:00:00 +00:00")
` I get `KeyError: 1630368000000` which means that I dont have that datetime in my dictionaries for funding_rate and mark_prices, but when I remove the timezone, I get a different error

rokups:
> I don't think the open date of the trades is in UTC, when I use ` time_machine.move_to("2021-09-01 00:00:00 +00:00")` I get `KeyError: 1630368000000` which means that I dont have that datetime in my dictionaries for funding_rate and mark_prices, but when I remove the timezone, I get a different error

`trade.open_date` is not UTC, `trade.open_date_utc` is.

samgermain:
> > I don't think the open date of the trades is in UTC, when I use ` time_machine.move_to("2021-09-01 00:00:00 +00:00")` I get `KeyError: 1630368000000` which means that I dont have that datetime in my dictionaries for funding_rate and mark_prices, but when I remove the timezone, I get a different error
> 
> `trade.open_date` is not UTC, `trade.open_date_utc` is.

Thanks, my issue isn't fixed, but this is useful

freqtrade/exchange/binance.py
samgermain: FTX, said no

Kraken said

"So  if a trader opened a position at 08:00:05 UTC, and closed it at  08:30:00 UTC, would this trader be charged the 08:00:00 UTC funding fee? Please note that the absolute funding payout can be calculated by using the formula below:Absolute funding payout = # of contracts * absolute funding rate * time elapsed within funding period without position alteration. As  you can see above, the funding payout is calculated based on the time  your position remains open and unaltered within the funding period. In  your example, this would be 29 min and 55 seconds."

samgermain: I just double checked with binance

"If two traders are holding a long position for 300 ADA on the ADA/USDT perpetual futures market, 
but one entered their position at 02:00 UTC on Nov 1 2021, and the other entered their position at 
06:00 UTC on Nov 1 2021, then if they are both still holding their positions at 08:00 UTC on 
Nov 1 2021, they will be charged the same funding fee. Is that correct?"

and they said it was, so that comment by Kraken I think only applies to Kraken, but I also emailed FTX to make sure

Gate.io hasn't been replying to me

samgermain:
@xmatthias All the tests pass now, this PR is done in my opinion

freqtrade/exchange/kraken.py
```diff
@@ -156,3 +156,25 @@ def _get_params(self, ordertype: str, leverage: float, time_in_force: str = 'gtc
...
+            :param funding_rate: the interest rate and the premium
+                - interest rate:
+                - premium: varies by price difference between the perpetual contract and mark price
+            :param time_in_ratio: time elapsed within funding period without position alteration
+        """
+        if not time_in_ratio:
+            raise OperationalException(
+                f"time_in_ratio is required for {self.name}._get_funding_fee")
+        nominal_value = mark_price * size
+        return nominal_value * funding_rate * time_in_ratio
```
samgermain, line 184: Krakens formula is a bit different, I included it in here to save whoever some time when they decide they want to trade futures on kraken, but there's a few tests that I had to comment out, due to `calculate_funding_fees` not and `update_funding_fees` not passing in the `time_in_ratio` parameter that kraken needs. I wrote tests for this method, but probably won't implement the changes needed in `calculate_funding_fees` not and `update_funding_fees` to trade futures on Kraken

tests/exchange/test_exchange.py
```diff
@@ -3288,3 +3288,347 @@ def test_get_max_leverage(default_conf, mocker, pair, nominal_value, max_lev):
...
+    ('binance', "2021-09-01 00:00:00", "2021-09-01 07:59:59",  30.0, -0.0006647999999999999),
+    ('binance', "2021-09-01 00:00:00", "2021-09-01 12:00:00",  30.0, -0.0009140999999999999),
+    ('binance', "2021-09-01 00:00:01", "2021-09-01 08:00:00",  30.0, -0.0009140999999999999),
+    # TODO: Uncoment once calculate_funding_fees can pass time_in_ratio to exchange._get_funding_fee
+    # ('kraken', "2021-09-01 00:00:00", "2021-09-01 08:00:00",  30.0, -0.0014937),
+    # ('kraken', "2021-09-01 00:00:15", "2021-09-01 08:00:00",  30.0, -0.0008289),
+    # ('kraken', "2021-09-01 01:00:14", "2021-09-01 08:00:00",  30.0, -0.0008289),
+    # ('kraken', "2021-09-01 00:00:00", "2021-09-01 07:59:59",  30.0, -0.0012443999999999999),
+    # ('kraken', "2021-09-01 00:00:00", "2021-09-01 12:00:00", 30.0,  0.0045759),
+    # ('kraken', "2021-09-01 00:00:01", "2021-09-01 08:00:00",  30.0, -0.0008289),
```
samgermain, line 3410: These are some of the Kraken tests I had to comment out, they had previously passed when using the superclass method

tests/exchange/test_exchange.py
```diff
@@ -3288,3 +3288,347 @@ def test_get_max_leverage(default_conf, mocker, pair, nominal_value, max_lev):
...
+    ('ftx', "2021-09-01 00:00:00", "2021-09-01 08:00:00", 30.0,  0.0010008000000000003),
+    ('ftx', "2021-09-01 00:00:00", "2021-09-01 12:00:00", 30.0,  0.0146691),
+    ('ftx', "2021-09-01 00:00:01", "2021-09-01 08:00:00", 30.0,  0.0016656000000000002),
+    ('gateio', "2021-09-01 00:00:00", "2021-09-01 08:00:00",  30.0, -0.0009140999999999999),
+    ('gateio', "2021-09-01 00:00:00", "2021-09-01 12:00:00",  30.0, -0.0009140999999999999),
+    ('gateio', "2021-09-01 00:00:01", "2021-09-01 08:00:00",  30.0, -0.0002493),
+    ('binance', "2021-09-01 00:00:00", "2021-09-01 08:00:00",  50.0, -0.0015235000000000001),
+    # TODO: Uncoment once calculate_funding_fees can pass time_in_ratio to exchange._get_funding_fee
```
samgermain: This is one of the Kraken tests I had to comment out

tests/exchange/test_exchange.py
```diff
@@ -3288,3 +3288,347 @@ def test_get_max_leverage(default_conf, mocker, pair, nominal_value, max_lev):
...
+    type(api_mock).has = PropertyMock(return_value={'fetchFundingRateHistory': True})
+
+    exchange = get_patched_exchange(mocker, default_conf, api_mock, id=exchange)
+    funding_fees = exchange.calculate_funding_fees('ADA/USDT', amount, d1, d2)
+    assert funding_fees == expected_fees
+
+
+@pytest.mark.parametrize('name,expected_fees_8,expected_fees_10,expected_fees_12', [
+    ('binance', -0.0009140999999999999, -0.0009140999999999999, -0.0009140999999999999),
+    # TODO: Uncoment once calculate_funding_fees can pass time_in_ratio to exchange._get_funding_fee
```
samgermain, line 3601: These is one of the Kraken tests I had to comment out

tests/exchange/test_exchange.py
```diff
@@ -3288,3 +3288,347 @@ def test_get_max_leverage(default_conf, mocker, pair, nominal_value, max_lev):
...
+    # ('kraken', -0.0014937, -0.0014937, 0.0045759),
```
samgermain: This is one of the Kraken tests I had to comment out

freqtrade/exchange/binance.py
xmatthias: so basically, on kraken, the "hour" is irrelevant, as it's the duration of the trade that counts ...

freqtrade/exchange/kraken.py
xmatthias: please add a comment in the docstring mentioning that it's intentionally not active ... 

I usually don't check comments on PR's when i wonder about a line of code ... or if i do, only hours into the investigation, and i assume others will do the same :laughing:

freqtrade/freqtradebot.py
```diff
@@ -268,11 +268,19 @@ def get_free_open_trades(self) -> int:
...
         if self.trading_mode == TradingMode.FUTURES:
-            for trade in Trade.get_open_trades():
-                funding_fees = self.exchange.get_funding_fees_from_exchange(
-                    trade.pair,
-                    trade.open_date
-                )
+            trades = Trade.get_open_trades()
+            for trade in trades:
+                if self.config['dry_run']:
+                    funding_fees = self.exchange.calculate_funding_fees(
```
xmatthias: can we move this check into `get_funding_fees()` within the exchange class?

the distinction (dry/non-dry) is done within the exchange wherever possible - and i don't see any benefits to having this as part of freqtradebot.
this also helps with tests to some degree, as we don't have to bother what dry_run is when we mock an exchange class function.

freqtrade/exchange/exchange.py
```diff
@@ -1684,11 +1704,33 @@ def _set_leverage(
...
+        d2 = datetime(d2.year, d2.month, d2.day, d2.hour, tzinfo=timezone.utc)
+
+        results = []
+        d3 = d1
+        while d3 <= d2:
+            if d3.hour in self.funding_fee_times:
+                results.append(d3)
+            d3 += timedelta(hours=1)
+
+        return results
```
xmatthias: I think this method would benefit from clearer variable names.

`d1` / `d2` is not very clear (maybe use `d_start`, `d_end`?) - and `d3` seems to be unnecessary it's an alias of `d1` - which is never used beyond the aliasing.

freqtrade/exchange/exchange.py
```diff
@@ -1704,6 +1746,141 @@ def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
...
+        )
+        mark_price_history = self._get_mark_price_history(
+            pair,
+            open_timestamp
+        )
+        funding_fee_dates = self._get_funding_fee_dates(open_date, close_date)
+        for date in funding_fee_dates:
+            timestamp = int(date.timestamp()) * 1000
+            funding_rate = funding_rate_history[timestamp]
+            mark_price = mark_price_history[timestamp]
```
xmatthias: this is bound to fail.
assuming an exchange downtime that spans for 5 hours - you will encounter 1 missing timestamp.

As you're not checking if timestamp is in the index (nor access it safely with `.get()`) - this will make freqtrade fail / crash until the downtime period is no longer relevant (it either ages out - or the trade is deleted).

freqtrade/exchange/exchange.py
```diff
@@ -1704,6 +1746,141 @@ def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
...
             raise OperationalException(e) from e
 
+    @retrier
+    def _get_mark_price_history(
+        self,
+        pair: str,
+        since: int
+    ) -> Dict:
+        """
+            Get's the mark price history for a pair
```
xmatthias: I'm honestly not sure why you're indenting docstrings. 
i've never seen it indented before - and i really don't like how it looks (it also makes your lines implicitly shorter ... even though that's not relevant in this case).

also the [Python PEP](https://www.python.org/dev/peps/pep-0257/) about docstring convention doesn't indent the docstring further ... so i'd like them to be aligned with the first quote.

samgermain: > I'm honestly not sure why you're indenting docstrings.

It makes them collapseable, at least in VSCode. It helps me a lot to collapse everything I don't immediately need to work with so that there's less distractions

xmatthias: That's fine for your environment - but for the codebase it's not something i'd want to have, so for commits, please dedent them to have the text align with quotes.

I recognize now that i made a mistake in merging PR's that had docstrings this way in the past, as it creates 2 different styles for it, so i'll fix it for develop and feat/short - and will add this to the code style guidelines.

freqtrade/freqtradebot.py
```diff
@@ -704,7 +706,7 @@ def execute_entry(
         fee = self.exchange.get_fee(symbol=pair, taker_or_maker='maker')
         open_date = datetime.now(timezone.utc)
         if self.trading_mode == TradingMode.FUTURES:
-            funding_fees = self.exchange.get_funding_fees_from_exchange(pair, open_date)
+            funding_fees = self.exchange.get_funding_fees(pair, amount, open_date)
```
xmatthias: Right now, we're only calling this twice - once "On entry" and once every hour(ish).

while i'm not convinced on calling this right after entry makes sense - i do think that we'll need to call this before/after exiting a trade.

it's pretty unlikely that the "update tick" happened in the 0.x seconds it takes us to go from "create_order" to this line - and even if it did, it'd update latest within 15 minutes anyway.

On closing a trade on the other hand, we're not sure how the timing is ...so I think, we'll have to also call this right before/after closing a trade (maybe i'm wrong ... but it's a feeling i have whenever thinking about fees).

xmatthias: thinking about this a little more, i think this is a pretty simple decision 
- we're either trusting schedule to work all the time (which means, the call right after entry is not necessary) 
- or we make sure we're calling it before taking action - in which case we should call it at entry - and right before trade exit.

samgermain: we need to call it when an order closes, is that inside the `_notify_enter` and `_notify_exit` methods?

samgermain: or is it in the `execute_entry` method, I already have a call there, I guess we only need to call it on exit if the scheduler doesn't do what it's supposed to for some reason

samgermain: I added in a safety update when the trade exits, and I tried to test it, but I'm getting an sql integrity error

```
elf = <sqlalchemy.dialects.sqlite.pysqlite.SQLiteDialect_pysqlite object at 0x1316f1fa0>
cursor = <sqlite3.Cursor object at 0x131bbb500>
statement = 'INSERT INTO orders (ft_trade_id, ft_order_side, ft_pair, ft_is_open, order_id, status, symbol, order_type, side, pric...date, order_filled_date, order_update_date, leverage) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)'
parameters = (1, 'sell', 'ETH/BTC', 1, 'mocked_limit_buy_usdt', 'open', ...)
context = <sqlalchemy.dialects.sqlite.base.SQLiteExecutionContext object at 0x131a069a0>

    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlite3.IntegrityError: UNIQUE constraint failed: orders.ft_pair, orders.order_id

.env/lib/python3.9/site-packages/sqlalchemy/engine/default.py:717: IntegrityError

The above exception was the direct cause of the following exception:

mocker = <pytest_mock.plugin.MockerFixture object at 0x1328dfaf0>
default_conf = {'amend_last_stake_amount': False, 'ask_strategy': {'order_book_top': 1, 'price_side': 'ask', 'use_order_book': False}...delta': 1, 'enabled': False}, 'order_book_top': 1, 'price_side': 'bid', ...}, 'cancel_open_orders_on_exit': False, ...}
time_machine = <time_machine.TimeMachineFixture object at 0x13157b520>
fee = <MagicMock id='5122798064'>
ticker_usdt_sell_up = <MagicMock id='5127416176'>, is_short = False
limit_order_open = {'buy': {'amount': 30.0, 'cost': 60.0, 'datetime': '2021-11-10T07:20:52.498299+00:00', 'filled': 0.0, ...}, 'sell': {'amount': 30.0, 'cost': 66.0, 'datetime': '2021-11-10T07:20:52.498405+00:00', 'filled': 0.0, ...}}
schedule_off = True

    @pytest.mark.parametrize('schedule_off', [False, True])
    @pytest.mark.parametrize('is_short', [True, False])
    def test_update_funding_fees(
        mocker,
        default_conf,
        time_machine,
        fee,
        ticker_usdt_sell_up,
        is_short,
        limit_order_open,
        schedule_off
    ):
        '''
        nominal_value = mark_price * size
        funding_fee = nominal_value * funding_rate
        size = 123
        "LTC/BTC"
            time: 0, mark: 3.3, fundRate: 0.00032583, nominal_value: 405.9, fundFee: 0.132254397
            time: 8, mark: 3.2, fundRate: 0.00024472, nominal_value: 393.6, fundFee: 0.096321792
        "ETH/BTC"
            time: 0, mark: 2.4, fundRate: 0.0001, nominal_value: 295.2, fundFee: 0.02952
            time: 8, mark: 2.5, fundRate: 0.0001, nominal_value: 307.5, fundFee: 0.03075
        "ETC/BTC"
            time: 0, mark: 4.3, fundRate: 0.00031077, nominal_value: 528.9, fundFee: 0.164366253
            time: 8, mark: 4.1, fundRate: 0.00022655, nominal_value: 504.3, fundFee: 0.114249165
        "XRP/BTC"
            time: 0, mark: 1.2, fundRate: 0.00049426, nominal_value: 147.6, fundFee: 0.072952776
            time: 8, mark: 1.2, fundRate: 0.00032715, nominal_value: 147.6, fundFee: 0.04828734
        '''
        # SETUP
        time_machine.move_to("2021-09-01 00:00:00 +00:00")
    
        open_order = limit_order_open[enter_side(is_short)]
        bid = 0.11
        enter_rate_mock = MagicMock(return_value=bid)
        enter_mm = MagicMock(return_value=open_order)
        patch_RPCManager(mocker)
        patch_exchange(mocker)
        default_conf['trading_mode'] = 'futures'
        default_conf['collateral'] = 'isolated'
        default_conf['dry_run'] = True
    
        funding_rates = {
            "LTC/BTC": {
                1630454400000: 0.00032583,
                1630483200000: 0.00024472,
            },
            "ETH/BTC": {
                1630454400000: 0.0001,
                1630483200000: 0.0001,
            },
            "XRP/BTC": {
                1630454400000: 0.00049426,
                1630483200000: 0.00032715,
            }
        }
    
        mark_prices = {
            "LTC/BTC": {
                1630454400000: 3.3,
                1630483200000: 3.2,
            },
            "ETH/BTC": {
                1630454400000: 2.4,
                1630483200000: 2.5,
            },
            "XRP/BTC": {
                1630454400000: 1.2,
                1630483200000: 1.2,
            }
        }
    
        mocker.patch(
            'freqtrade.exchange.Exchange._get_mark_price_history',
            side_effect=lambda pair, since: mark_prices[pair]
        )
    
        mocker.patch(
            'freqtrade.exchange.Exchange.get_funding_rate_history',
            side_effect=lambda pair, since: funding_rates[pair]
        )
    
        mocker.patch.multiple(
            get_rate=enter_rate_mock,
            fetch_ticker=MagicMock(return_value={
                'bid': 1.9,
                'ask': 2.2,
                'last': 1.9
            }),
            create_order=enter_mm,
            get_min_pair_stake_amount=MagicMock(return_value=1),
            get_fee=fee,
        )
    
        freqtrade = get_patched_freqtradebot(mocker, default_conf)
    
        # initial funding fees,
        freqtrade.execute_entry('ETH/BTC', 123)
        freqtrade.execute_entry('LTC/BTC', 2.0)
        freqtrade.execute_entry('XRP/BTC', 123)
    
        trades = Trade.get_open_trades()
        assert len(trades) == 3
        for trade in trades:
            assert trade.funding_fees == (
                trade.amount *
                mark_prices[trade.pair][1630454400000] *
                funding_rates[trade.pair][1630454400000]
            )
    
        # create_mock_trades(fee, False)
        time_machine.move_to("2021-09-01 08:00:00 +00:00")
        if schedule_off:
            for trade in trades:
                assert trade.funding_fees == (
                    trade.amount *
                    mark_prices[trade.pair][1630454400000] *
                    funding_rates[trade.pair][1630454400000]
                )
>               freqtrade.execute_trade_exit(
                    trade=trade,
                    # The values of the next 2 params are irrelevant for this test
                    limit=ticker_usdt_sell_up()['bid'],
                    sell_reason=SellCheckTuple(sell_type=SellType.ROI)
                )

tests/test_freqtradebot.py:4838: 
```

xmatthias: it's neither of the above - `notify_` methods are too late (and should not be used for this).

you should look at `execute_trade_exit` (or whatever that method is called now).

samgermain: > it's neither of the above - `notify_` methods are too late (and should not be used for this).
> 
> you should look at `execute_trade_exit` (or whatever that method is called now).
 
That's what i did

samgermain: @xmatthias I mean that that's what I did the first time, I haven't made any updates since your reply, the error is an sqlintegrity error that I'm not sure what to do about

xmatthias: The reason itself is pretty simple - you're mocking `create_order()` and return a default object (which is fine for tests).
You're however never changing the mock to a "sell order" - which means that when the bot tries to sell the trade, it'll get the same order (the buy order) again - which fails to insert.

I think the simple fix for this is to recreate the `create_order` mock - but with a different order object (ideally one that's actually a sell - as i'm not sure if you'll not encounter other issues down that method ...).

xmatthias: I've fixed that part now - but the test is still failing for different reasons (this falls into your realm i hope - i'm not quite sure why this fails).

freqtrade/exchange/exchange.py
```diff
@@ -1704,6 +1746,164 @@ def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
...
+        if not self.exchange_has("fetchFundingRateHistory"):
+            raise ExchangeError(
+                f"CCXT has not implemented fetchFundingRateHistory for {self.name}; "
+                f"therefore, dry-run/backtesting for {self.name} is currently unavailable"
+            )
+
+        # TODO-lev: Gateio has a max limit into the past of 333 days, okex has a limit of 3 months
+        try:
+            funding_history: Dict = {}
+            response = self._api.fetch_funding_rate_history(
```
xmatthias, line 1878: I wanted to write a "live" test for this endpoint ... but it doesn't seem to be working (ccxt==1.60.92, installed 20 minutes ago).

```python
import ccxt
ccxt.binanceusdm()
exchange.fetch_funding_rate_history('BTC/USDT', '1h', limit=100)
```

fails with 
```
ExchangeError: binanceusdm {"status":"ERROR","type":"VALID","code":"100001003","errorData":{"startTime":"Failed to convert property value of type 'java.lang.String' to required type 'java.lang.Long' for property 'startTime'; nested exception is java.lang.NumberFormatException: For input string: \"1h\""},"data":null,"subData":null,"params":null}
```

samgermain: > I wanted to write a "live" test for this endpoint ... but it doesn't seem to be working (ccxt==1.60.92, installed 20 minutes ago).
> 
> ```python
> import ccxt
> ccxt.binanceusdm()
> exchange.fetch_funding_rate_history('BTC/USDT', '1h', limit=100)
> ```
> 
> fails with
> 
> ```
> ExchangeError: binanceusdm {"status":"ERROR","type":"VALID","code":"100001003","errorData":{"startTime":"Failed to convert property value of type 'java.lang.String' to required type 'java.lang.Long' for property 'startTime'; nested exception is java.lang.NumberFormatException: For input string: \"1h\""},"data":null,"subData":null,"params":null}
> ```

This is the method header for `fetch_funding_rate_history`, you're passing `1h` into the argument `since`

```
def fetch_funding_rate_history(self, symbol=None, since=None, limit=None, params={}):
```

xmatthias: mmmmh ... good point :laughing:

tests/exchange/test_ccxt_compat.py
```diff
@@ -149,6 +165,24 @@ def test_fetch_ohlcv(self, exchange):
         now = datetime.now(timezone.utc) - timedelta(minutes=(timeframe_to_minutes(timeframe) * 2))
         assert exchange.klines(pair_tf).iloc[-1]['date'] >= timeframe_to_prev_date(timeframe, now)
 
+    @pytest.mark.skip("No futures support yet")
+    def test_ccxt_fetch_funding_rate_history(self, exchange_futures):
+        # TODO-lev: enable this test once Futures mode is enabled.
```
xmatthias, line 175: @samgermain it would be good if you could "play" with this test a bit.

Simply removing the ".mark.skip" on this line, and commenting the "@longrun" at the top of the class will make it easily runnable.

Note: to run this correctly, you'll also need to uncomment the supported modes in each exchange class.
I've tried gateio and ftx (both work), as well as binance (initialization fails, as this is not using dry-run - and "get-brackets" step fails as that requires keys). I'll think of a workaround for this.

The one that concerns me is OKEX - which is completely failing (at least with the latest ccxt version), so maybe you could look at that? (we'll still not be able to enable this test in this PR i think - but it helps us ensure that things do work correctly.

samgermain:
So exchanges can randomly add extra funding time periods, here's an example where binance does it to OMG for half a day
https://www.binance.com/en/support/announcement/c943ef87fdcd40b194884805309f4f7f

I switched up the calculate_funding_fees method to just calculate a funding fee for each funding rate, since those are only calculated for when a funding fee is charged, it also makes it so that nothing needs to be customized for each exchange

It does make some of the time sensitive tests pointless though, because `fetch_funding_rate_history` is just returning a mocked response

xmatthias:
> It does make some of the time sensitive tests pointless though

it's a process - if you think a test is not needed / no longer needed - i'm perfectly fine to remove it
A test only makes sense if it actually tests something we don't want to accidentally break in the future ...

tests/exchange/test_ccxt_compat.py
samgermain: So I don't know if this is relevant, but for okex futures (swap on ccxt), the first parameter "symbol" is required, the symbol schema is base/quote:settle (ex. `ZIL/USDT:USDT`, gateio is also like this, and all exchanges will move to that schema sometime soon)

A different method by the name of `fetchFundingHistory`, which we also use, was just merged in a few days ago, so it may not have been in the latest release when you wrote that

samgermain:
> > It does make some of the time sensitive tests pointless though
> 
> it's a process - if you think a test is not needed / no longer needed - i'm perfectly fine to remove it A test only makes sense if it actually tests something we don't want to accidentally break in the future ...

I want to leave them in there right now, I'm trying to think how to work them into a different test, and I see no need to remove them right now

tests/exchange/test_ccxt_compat.py
samgermain: I don't really know what to do in order to play with it

xmatthias: if you look at all changes i did in this file, all (except for one) location will be highlighted.
it's basically:
* remove `@longrun` from the class (this is not highlighed)
* remove `@pytest.mark.skip` from this test
* add `futures: true` to the top of the file for the exchange you wanna test it
* enable "futures, cross" in the exchange class for the exchange to test it on (otherwise you'll get a "does not support cross futures" error)

run `pytest -k test_ccxt_fetch_funding_rate_history`

samgermain: Yeah just update CCXT again, I don't think Okex has a setMarginMode method also, I think it's done similar to Kraken

xmatthias: i don't think we'll want to support OKEX at the start anyway ... i think with binance/ftx/ ... we'll have enough to try and handle/test.

samgermain: ftx is only cross, so we won't be able to support it at the start (same with kraken) 
binance, gateio, and okex all provide an isolated  mode

samgermain: i downloaded the newest though and didn't get those error

freqtrade/exchange/exchange.py
```diff
@@ -1717,6 +1741,161 @@ def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
...
+                d = datetime.fromtimestamp(int(candle[0] / 1000), timezone.utc)
+                # Round down to the nearest hour, in case of a delayed timestamp
+                # The millisecond timestamps can be delayed ~20ms
+                time = datetime(
+                    d.year,
+                    d.month,
+                    d.day,
+                    d.hour,
+                    tzinfo=timezone.utc
+                ).timestamp() * 1000
```
xmatthias, line 1772: we do provide a method for this  in `timeframe_to_prev_date` ... 
while it sounds like the wrong approach as we're not dealing with timeframes, by using `1h` as timeframe, you can get this quite nicely ... ;)

xmatthias (approved):
Let's leave this as is for now :+1: 

we'll see if it'll work the way we expect it to, or if there'll be some adjustments necessary once we can start "real" testing.

samgermain:
> once we can start "real" testing.

I'm going to start doing that now, all the steps for binance isolated futures are done except backtesting and liquidation price, so I'm just going to set a stoploss at about 5-10%, and try to dry-run some bots at 3-5x leverage, and fix any bugs that pop up

freqtrade/exchange/exchange.py
```diff
@@ -1672,6 +1670,25 @@ def get_max_leverage(self, pair: Optional[str], nominal_value: Optional[float])
...
+        Calculates a single funding fee
+        :param size: contract size * number of contracts
+        :param mark_price: The price of the asset that the contract is based off of
+        :param funding_rate: the interest rate and the premium
+            - interest rate:
+            - premium: varies by price difference between the perpetual contract and mark price
+        :param time_in_ratio: Not used by most exchange classes
+        """
+        nominal_value = mark_price * size
+        return nominal_value * funding_rate
```
samgermain, line 1690: @xmatthias So this should actually be something like

```
if in a short trade:
     return nominal_value * funding_rate
else:
    return -(nominal_value * funding_rate)
```

Sorry about that, I don't really understand the new method with dataframes, so I'm not sure how to update it

This makes the number signage match the signage from the api response

samgermain: The formula was just calculating the size of the funding fee, not whether you had gained or lost this money, so this was the problem

PR #5780, opened 2021-10-24
Mark price candles

samgermain:
Edit's existing candlestick downloading methods so that they can download the mark price

TODO:

* [x] find good way to store the data 
* [x] auto-download required data when specifying `--trademode futures`
* [x] determine correct "additional" data by exchange
* [x]  ensure that relevant documentation is uptodate (--help output)
* [ ] ~Figure out how to deal with funding rates~ Will be done in a followup PR


freqtrade/exchange/exchange.py
```diff
@@ -1270,40 +1270,64 @@ def extract_cost_curr_rate(self, order: Dict) -> Tuple[float, str, Optional[floa
...
         """
         return asyncio.get_event_loop().run_until_complete(
-            self._async_get_historic_ohlcv(pair=pair, timeframe=timeframe,
-                                           since_ms=since_ms, is_new_pair=is_new_pair))
+            self._async_get_historic_ohlcv(
+                pair=pair,
+                timeframe=timeframe,
+                since_ms=since_ms,
+                is_new_pair=is_new_pair,
+                price=price
```
samgermain: From previous PR

> I've not seen nor used this before.
> 
> Please add a link to the relevant ccxt documentation for this (which assumes it's unified across all exchanges).
> If it's not unified, i don't think this is the right place.
> 
> In general, i'm very sceptical of 2 things here:
> 
>   overloading the same method for different functions
>   combining things that don't belong together into one PR.
> 
> so unless this is necessary to get dry-run going - i'd seperate it out of this pr.

freqtrade/exchange/exchange.py
```diff
@@ -1270,40 +1270,64 @@ def extract_cost_curr_rate(self, order: Dict) -> Tuple[float, str, Optional[floa
...
+        timeframe: str,
+        since_ms: int,
+        price: Optional[str] = None
+    ) -> DataFrame:
         """
         Minimal wrapper around get_historic_ohlcv - converting the result into a dataframe
         :param pair: Pair to download
         :param timeframe: Timeframe to get data for
         :param since_ms: Timestamp in milliseconds to get history from
+        :param price: "mark" if retrieving the mark price cnadles, "index" for index price candles
```
xmatthias:
```suggestion
        :param price: "mark" if retrieving the mark price candles, "index" for index price candles
```

same typo is basically in all docstrings you added here ... :laughing:

freqtrade/exchange/exchange.py
```diff
@@ -1411,7 +1452,8 @@ async def _async_get_candle_history(self, pair: str, timeframe: str,
                 "Fetching pair %s, interval %s, since %s %s...",
                 pair, timeframe, since_ms, s
             )
-            params = self._ft_has.get('ohlcv_params', {})
+            # TODO-lev: Does this put price into params correctly?
+            params = self._ft_has.get('ohlcv_params', {price: price})
```
xmatthias: Youre only changing the default - it'll not do what you think it does in some cases.

That said - from looking at the ccxt source for this, i think the correct call is `fetchMarkOHLCV()` (or `fetchIndexOHLCV()`?) - or `fetch_mark_ohlcv()` in python syntax, not overwriting the parameters we pass to `fetch_ohlcv`.
That way, ccxt will handle the "exchange" abstraction for this (how it's done for each exchange) - as i assume the parameter needed will differ by exchange.

Setting the price parameter might be correct for binance, but might be different / wrong for other exchanges.

As the ccxt docs put it:

> The contents of params are exchange-specific, consult the exchanges’ API documentation for supported fields and values.

[source](https://ccxt.readthedocs.io/en/latest/manual.html#ohlcv-candlestick-charts)

it's also rather interresting, as neither `fetchMarkOHLCV` nor `fetchIndexOHLCV` appear in the docs, so i don't think these are unified methods (which means we can't rely on them being available ...  on all exchanges, and remain stable (?) )

I'm also not sure if you'll need this parameter in the whole chain - you'll only ever need 1 timeframe for this - 1h (funding-fees are updated hourly, if i understood that correctly).
That said, i'm also not sure where/how you're going to integrate this ... so MAYBE it'd be better to have a whole seperate method for this ... but with the available information at this moment, that's very hard to tell.

freqtrade/exchange/exchange.py
samgermain: https://ccxt.readthedocs.io/en/latest/manual.html?highlight=fetchMarkOHLCV#mark-index-and-premiumindex-candlestick-charts

> overloading the same method for different functions

Well the functionality is kind of the same, I was just trying to add the ability to download the mark price candles instead of the regular candles

freqtrade/exchange/exchange.py
samgermain: > That said - from looking at the ccxt source for this, i think the correct call is fetchMarkOHLCV() (or fetchIndexOHLCV()?) - or fetch_mark_ohlcv() in python syntax, not overwriting the parameters we pass to fetch_ohlcv.

`fetchIndexOHLCV` and `fetchMarkOHLCV` are convenience methods that just call `fetchOHLCV` with `params={price: 'mark'}` + user entered params

> The contents of params are exchange-specific, consult the exchanges’ API documentation for supported fields and values.

So that's not true 100% of the time, maybe that needs to be reworded. Any exchange specific arguments go inside params, but in some cases params can have extra arguments that weren't there when the method was first implemented. In this case the params key `price` is unified

xmatthias: Personally, i do have my doubts / reservations about this double-use of `params` in ccxt, as it can cause some very odd circumstances and bugs (assume an exchange uses `price` as parameter for this endpoint by "ExchangeY", and a "ccxt_price" would've made it clearer that this is only a ccxt relevant parameter), but i don't think this is the right place to discuss this - and if ccxt recommends this usage and doesn't see future problems with it - i guess i'm fine with that.

freqtrade/exchange/exchange.py
xmatthias: for the downloading itself, i think you'll have to go "wider" (changing how these are stored) - as currently, you'd be overwriting / combining these candles with already existing candles, which is for sure problematic / a deal breaker.

samgermain: > as currently, you'd be overwriting / combining these candles with already existing candles

Which part overwrites/combines them?

xmatthias: well everything? at the moment - there's no code trying to write this dat to different files.

without that - it'll be combined - and as the format matches - there'll be no difference - so new candles for this will be merged with other candles - so you'll end up with data you can't use, as the data is 100% combined (your file `XRP_USDT_5m.json` will contain a mix of spot candles, and mark candles - assuming you run both one after the other with a few minutes gap between).

samgermain: Wouldn't I just need to find the part that declares the file name, and name it something like `mark_XRP_USDT_5m.json`

xmatthias: you can find what you're looking for in freqtrade/data/history.
It's however not one file - but 3 (one base class, and the json and hdf5 implementations of the datahandlers).

samgermain:
@xmatthias If you don't think this PR is that useful you can close it, I just thought it would be handy to be able to download the mark price candles with these methods

freqtrade/exchange/exchange.py
samgermain: I think I might have fixed it, I haven't tested it out

samgermain:
Somehow the stuff on the funding-fee-dry-run got merged in here i think

xmatthias:
> @xmatthias If you don't think this PR is that useful you can close it, I just thought it would be handy to be able to download the mark price candles with these methods

well i think for backtesting, we'll **need** to have the possibility to download "alternative" datasets (futures candles, mark candles).
This PR does however only touch one particular area (the exchange side) - but ignores all the logistics needed behind to store the data seperately.
Merging (or even using) it as is will destroy your existing spot candle dataset ... so it's still quite some work ahead.

samgermain:
> Merging (or even using) it as is will destroy your existing spot candle dataset ... so it's still quite some work ahead.

I think I fixed that, I edited the files in history so that the file name adds f"-{candle_type}" at the end if pass in a type, but I think it might require less changes if we just store them in a different folder, like with a different exchange

freqtrade/data/converter.py
```diff
@@ -75,12 +78,17 @@ def clean_ohlcv_dataframe(data: DataFrame, timeframe: str, pair: str, *,
...
     else:
         return data
 
 
-def ohlcv_fill_up_missing_data(dataframe: DataFrame, timeframe: str, pair: str) -> DataFrame:
+def ohlcv_fill_up_missing_data(
+    dataframe: DataFrame,
+    timeframe: str,
+    pair: str,
+    candle_type: Optional[str] = ""
```
xmatthias: Please don't put arguments to functions where it's clear that they neither need, nor use them.

That also means - change back the formatting change.

freqtrade/data/history/hdf5datahandler.py
```diff
@@ -157,7 +189,11 @@ def trades_append(self, pair: str, data: TradeList):
         """
         raise NotImplementedError()
 
-    def _trades_load(self, pair: str, timerange: Optional[TimeRange] = None) -> TradeList:
+    def _trades_load(
+        self,
+        pair: str,
+        timerange: Optional[TimeRange] = None,
+    ) -> TradeList:
```
xmatthias, line 196: Please don't make pointless formatting changes.

I think i mention this in every other PR - please keep this in mind - i'm tired of repeating my self over and over and over again with the same thing.

xmatthias: Please don't make pointless formatting changes.

I think i mention this in every other PR - please keep this in mind - i'm tired of repeating my self over and over and over again with the same thing, and this makes reviewing VERY tiresome.

xmatthias (changes requested):
I think i've asked you often enough to not make pointless formatting changes.

Reviewing these only wastes time - as i need to try and figure out if (and what) changed in these lines - only to discover that it's another pointless change.

Let me know when all non-necessary changes have been reverted, then i will continue to look at this PR.
Also keep this in mind for other / further PR's - as the same applies to these as well.

If you think formatting should be changed (there's no reason for it i see at the moment), to this in a separate PR which is not changing anything else - so the review focus can be put onto this alone - and not on the whole logic around it as well, thanks.

freqtrade/data/history/hdf5datahandler.py
samgermain: I had previously added an argument which brought this line past 100 characters, later I decided I didn't need it and removed it (with VSCodes replace tool probably), but didn't remove the formatting change. 

Please be understanding that I may not catch everything every time I push to draft PRs

xmatthias: That's for sure true ... but in my understanding from the above comment, you did want an opinion ... which requires an "early" review.

samgermain:
I'm not really sure how I can get the download data command to download the mark price candles instead of the spot candles. All the code I can see on the download-data command is

```
download_data_cmd = subparsers.add_parser(
            'download-data',
            help='Download backtesting data.',
            parents=[_common_parser],
)
download_data_cmd.set_defaults(func=start_download_data)
self._build_args(optionlist=ARGS_DOWNLOAD_DATA, parser=download_data_cmd)
 ```

File: tests/data/test_history.py [hunk 6]
```diff
@@ -657,6 +669,12 @@ def test_jsondatahandler_ohlcv_load(testdatadir, caplog):
     df = dh.ohlcv_load('XRP/ETH', '5m')
     assert len(df) == 711
 
+    df_mark = dh.ohlcv_load('XRP/USDT', '1h', candle_type="mark")
+    assert len(df_mark) == 99
```
samgermain, line 781: I'm kind of confused why `len(df_mark)` is `99`, I feel like it should be 100, because there's 100 array items in `XRP_USDT-1h-mark.json`, but when I test it, it tells me that `len(df_mark)` is equal to `99`

tests/data/test_history.py
```diff
@@ -617,6 +629,16 @@ def test_datahandler_ohlcv_get_pairs(testdatadir):
...
+    pairs = JsonDataHandler.ohlcv_get_pairs(testdatadir, '1h', 'mark')
+    assert set(pairs) == {'UNITTEST/USDT', 'XRP/USDT'}
+
+    # TODO-lev: The tests below
+    # pairs = JsonGzDataHandler.ohlcv_get_pairs(testdatadir, '8m')
+    # assert set(pairs) == {'UNITTEST/BTC'}
+
+    # pairs = HDF5DataHandler.ohlcv_get_pairs(testdatadir, '5m')
+    # assert set(pairs) == {'UNITTEST/BTC'}
+
```
samgermain, line 716: I don't know how to make a test HDF5 or Gz file for these tests

tests/data/test_history.py
xmatthias: last candle is assumed incomplete and dropped.
I'm pretty well aware that this dropping happens twice (once when downloading, once when loading) .... but it's so far not been worth fixing, and i wouldn't try a fix for this as part of feat/short (if at all, in develop - but it's not a big deal i think).

tests/data/test_history.py
xmatthias: i'm not sure we're at the test creation phase with this PR just yet ... as I'm not sure on usability (or how to make it usable) at the moment.

in theory, we can use something like `freqtrade download-data --trademode futures` (that parameter exists for list-pairs) ... but what should that download? 
only "mark" candles, or also futures candles?

my assumption is that futures candles might be slightly different to spot candles - so we'll really have 3 types of candles
* spot (for trading_mode=spot + margin)
* futures (trading_mode=futures)
* mark (trading_mode=futures)

as it won't make sense to have futures candles without mark candles (or viceversa) - I'd assume that it makes sense to download both at the same time (?)

samgermain:
These 2 tests currently fail, and I don't know how to fix them 

```
tests/exchange/test_binance.py::test__async_get_historic_ohlcv_binance
tests/exchange/test_exchange.py::test_refresh_latest_ohlcv
```


-----------------

```
____________________________________________________ test_refresh_latest_ohlcv ____________________________________________________

mocker = <pytest_mock.plugin.MockerFixture object at 0x12e392a90>
default_conf = {'ask_strategy': {'order_book_top': 1, 'use_order_book': False}, 'bid_strategy': {'ask_last_balance': 0.0, 'check_dept...cel_open_orders_on_exit': False, 'datadir': '/Users/sam/Documents/dev/trading/freqtrade/freqtrade/tests/testdata', ...}
caplog = <_pytest.logging.LogCaptureFixture object at 0x12e698160>

    def test_refresh_latest_ohlcv(mocker, default_conf, caplog) -> None:
        ohlcv = [
            [
                (arrow.utcnow().int_timestamp - 1) * 1000,  # unix timestamp ms
                1,  # open
                2,  # high
                3,  # low
                4,  # close
                5,  # volume (in quote currency)
            ],
            [
                arrow.utcnow().int_timestamp * 1000,  # unix timestamp ms
                3,  # open
                1,  # high
                4,  # low
                6,  # close
                5,  # volume (in quote currency)
            ]
        ]
    
        caplog.set_level(logging.DEBUG)
        exchange = get_patched_exchange(mocker, default_conf)
        exchange._api_async.fetch_ohlcv = get_mock_coro(ohlcv)
    
        pairs = [('IOTA/ETH', '5m', ''), ('XRP/ETH', '5m', '')]
        # empty dicts
        assert not exchange._klines
        res = exchange.refresh_latest_ohlcv(pairs, cache=False)
        # No caching
        assert not exchange._klines
    
        assert len(res) == len(pairs)
        assert exchange._api_async.fetch_ohlcv.call_count == 2
        exchange._api_async.fetch_ohlcv.reset_mock()
    
        exchange.required_candle_call_count = 2
        res = exchange.refresh_latest_ohlcv(pairs)
        assert len(res) == len(pairs)
    
        assert log_has(f'Refreshing candle (OHLCV) data for {len(pairs)} pairs', caplog)
        assert exchange._klines
        assert exchange._api_async.fetch_ohlcv.call_count == 4
        exchange._api_async.fetch_ohlcv.reset_mock()
        for pair in pairs:
            assert isinstance(exchange.klines(pair), DataFrame)
            assert len(exchange.klines(pair)) > 0
    
            # klines function should return a different object on each call
            # if copy is "True"
            assert exchange.klines(pair) is not exchange.klines(pair)
            assert exchange.klines(pair) is not exchange.klines(pair, copy=True)
            assert exchange.klines(pair, copy=True) is not exchange.klines(pair, copy=True)
            assert exchange.klines(pair, copy=False) is exchange.klines(pair, copy=False)
    
        # test caching
>       res = exchange.refresh_latest_ohlcv([('IOTA/ETH', '5m', ''), ('XRP/ETH', '5m', '')])

tests/exchange/test_exchange.py:1746: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
freqtrade/exchange/exchange.py:1421: in refresh_latest_ohlcv
    logger.debug(
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:1434: in debug
    self._log(DEBUG, msg, args, **kwargs)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:1589: in _log
    self.handle(record)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:1599: in handle
    self.callHandlers(record)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:1661: in callHandlers
    hdlr.handle(record)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:952: in handle
    self.emit(record)
.env/lib/python3.9/site-packages/_pytest/logging.py:331: in emit
    super().emit(record)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:1091: in emit
    self.handleError(record)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:1083: in emit
    msg = self.format(record)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:927: in format
    return fmt.format(record)
.env/lib/python3.9/site-packages/_pytest/logging.py:92: in format
    return super().format(record)
/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:663: in format
    record.message = record.getMessage()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <LogRecord: freqtrade.exchange.exchange, 10, /Users/sam/Documents/dev/trading/freqtrade/freqtrade/freqtrade/exchange/exchange.py, 1421, "Using cached candle (OHLCV) data for pair %s, timeframe %s ...">

    def getMessage(self):
        """
        Return the message for this LogRecord.
    
        Return the message for this LogRecord after merging any user-supplied
        arguments with the message.
        """
        msg = str(self.msg)
        if self.args:
>           msg = msg % self.args
E           TypeError: not all arguments converted during string formatting

/Users/sam/.pyenv/versions/3.9.5/lib/python3.9/logging/__init__.py:367: TypeError
-------------------------------------------------------- Captured log call --------------------------------------------------------
INFO     freqtrade.exchange.exchange:exchange.py:115 Instance is running with dry_run enabled
INFO     freqtrade.exchange.exchange:exchange.py:116 Using CCXT 1.61.16
INFO     freqtrade.exchange.exchange:exchange.py:156 Using Exchange "Binance"
INFO     freqtrade.resolvers.exchange_resolver:exchange_resolver.py:56 Using resolved exchange 'Binance'...
DEBUG    freqtrade.exchange.exchange:exchange.py:1398 Refreshing candle (OHLCV) data for 2 pairs
DEBUG    asyncio:selector_events.py:59 Using selector: KqueueSelector
DEBUG    freqtrade.exchange.exchange:exchange.py:1491 Fetching pair IOTA/ETH, interval 5m, since None ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1513 Done fetching pair IOTA/ETH, interval 5m ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1491 Fetching pair XRP/ETH, interval 5m, since None ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1513 Done fetching pair XRP/ETH, interval 5m ...
DEBUG    freqtrade.data.converter:converter.py:32 Converting candle (OHLCV) data to dataframe for pair IOTA/ETH.
DEBUG    freqtrade.data.converter:converter.py:75 Dropping last candle
DEBUG    freqtrade.data.converter:converter.py:32 Converting candle (OHLCV) data to dataframe for pair XRP/ETH.
DEBUG    freqtrade.data.converter:converter.py:75 Dropping last candle
DEBUG    freqtrade.exchange.exchange:exchange.py:1398 Refreshing candle (OHLCV) data for 2 pairs
DEBUG    freqtrade.exchange.exchange:exchange.py:1356 one_call: 300000000 msecs (3 days)
DEBUG    freqtrade.exchange.exchange:exchange.py:1356 one_call: 300000000 msecs (3 days)
DEBUG    freqtrade.exchange.exchange:exchange.py:1491 Fetching pair IOTA/ETH, interval 5m, since 1636883400000 (2021-11-14T09:50:00+00:00) ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1513 Done fetching pair IOTA/ETH, interval 5m ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1491 Fetching pair IOTA/ETH, interval 5m, since 1637183400000 (2021-11-17T21:10:00+00:00) ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1513 Done fetching pair IOTA/ETH, interval 5m ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1491 Fetching pair XRP/ETH, interval 5m, since 1636883400000 (2021-11-14T09:50:00+00:00) ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1513 Done fetching pair XRP/ETH, interval 5m ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1491 Fetching pair XRP/ETH, interval 5m, since 1637183400000 (2021-11-17T21:10:00+00:00) ...
DEBUG    freqtrade.exchange.exchange:exchange.py:1513 Done fetching pair XRP/ETH, interval 5m ...
DEBUG    freqtrade.data.converter:converter.py:32 Converting candle (OHLCV) data to dataframe for pair IOTA/ETH.
DEBUG    freqtrade.data.converter:converter.py:75 Dropping last candle
DEBUG    freqtrade.data.converter:converter.py:32 Converting candle (OHLCV) data to dataframe for pair XRP/ETH.
DEBUG    freqtrade.data.converter:converter.py:75 Dropping last candle
DEBUG    freqtrade.exchange.exchange:exchange.py:1398 Refreshing candle (OHLCV) data for 2 pairs
===================================================== short test summary info =====================================================
FAILED tests/exchange/test_exchange.py::test_refresh_latest_ohlcv - TypeError: not all arguments converted during string formatting
```

----------------------

```
____________________________________________ test__async_get_historic_ohlcv_binance[] _____________________________________________

default_conf = {'ask_strategy': {'order_book_top': 1, 'use_order_book': False}, 'bid_strategy': {'ask_last_balance': 0.0, 'check_dept...cel_open_orders_on_exit': False, 'datadir': '/Users/sam/Documents/dev/trading/freqtrade/freqtrade/tests/testdata', ...}
mocker = <pytest_mock.plugin.MockerFixture object at 0x12fce1700>
caplog = <_pytest.logging.LogCaptureFixture object at 0x12fd10670>, candle_type = ''

    @pytest.mark.asyncio
    @pytest.mark.parametrize('candle_type', ['mark', ''])
    async def test__async_get_historic_ohlcv_binance(default_conf, mocker, caplog, candle_type):
        ohlcv = [
            [
                int((datetime.now(timezone.utc).timestamp() - 1000) * 1000),
                1,  # open
                2,  # high
                3,  # low
                4,  # close
                5,  # volume (in quote currency)
            ]
        ]
    
        exchange = get_patched_exchange(mocker, default_conf, id='binance')
        # Monkey-patch async function
        exchange._api_async.fetch_ohlcv = get_mock_coro(ohlcv)
    
        pair = 'ETH/BTC'
        respair, restf, restype, res = await exchange._async_get_historic_ohlcv(
            pair, "5m", 1500000000000, is_new_pair=False, candle_type=candle_type)
        assert respair == pair
        assert restf == '5m'
        assert restype == candle_type
        # Call with very old timestamp - causes tons of requests
        assert exchange._api_async.fetch_ohlcv.call_count > 400
        # assert res == ohlcv
        exchange._api_async.fetch_ohlcv.reset_mock()
        _, _, _, res = await exchange._async_get_historic_ohlcv(
            pair, "5m", 1500000000000, is_new_pair=True, candle_type=candle_type)
    
        # Called twice - one "init" call - and one to get the actual data.
>       assert exchange._api_async.fetch_ohlcv.call_count == 2
E       AssertionError: assert 460 == 2
E        +  where 460 = <Mock name='mock().fetch_ohlcv' id='5098590992'>.call_count
E        +    where <Mock name='mock().fetch_ohlcv' id='5098590992'> = <MagicMock name='mock()' id='5098534896'>.fetch_ohlcv
E        +      where <MagicMock name='mock()' id='5098534896'> = <freqtrade.exchange.binance.Binance object at 0x12fe5c4c0>._api_async

tests/exchange/test_binance.py:377: AssertionError
-------------------------------------------------------- Captured log call --------------------------------------------------------
INFO     freqtrade.exchange.exchange:exchange.py:115 Instance is running with dry_run enabled
INFO     freqtrade.exchange.exchange:exchange.py:116 Using CCXT 1.61.16
INFO     freqtrade.exchange.exchange:exchange.py:156 Using Exchange "Binance"
INFO     freqtrade.resolvers.exchange_resolver:exchange_resolver.py:56 Using resolved exchange 'Binance'...
===================================================== short test summary info =====================================================
FAILED tests/exchange/test_binance.py::test__async_get_historic_ohlcv_binance[mark] - TypeError: '>' not supported between insta...
FAILED tests/exchange/test_binance.py::test__async_get_historic_ohlcv_binance[] - AssertionError: assert 460 == 2
```

freqtrade/data/dataprovider.py
```diff
@@ -77,11 +89,17 @@ def historic_ohlcv(self, pair: str, timeframe: str = None) -> DataFrame:
...
+                candle_type=candle_type
             )
         return self.__cached_pairs_backtesting[saved_pair].copy()
 
-    def get_pair_dataframe(self, pair: str, timeframe: str = None) -> DataFrame:
+    def get_pair_dataframe(
+        self,
+        pair: str,
+        timeframe: str = None,
+        candle_type: str = ''
```
xmatthias, line 105: Do we assume people will use alternative candle-types (like mark/index)? 
what would you use that for ...if i may ask?

I'm also not convinced this will work "just like this" - as we're not constantly refreshing funds candles (for example) ... so these would have to be specified explicitly via informative pair (?) - which we don't seem to be able to ... ?

tests/rpc/test_rpc_apiserver.py
```diff
@@ -1333,7 +1333,7 @@ def test_list_available_pairs(botclient):
     rc = client_get(client, f"{BASE_URI}/available_pairs")
 
     assert_response(rc)
-    assert rc.json()['length'] == 13
+    assert rc.json()['length'] == 15
```
xmatthias: doesn't this test mean that the result includes also mark candles?

I don't think that's correct - we pass nothing as candle-type, so i'd assume that not `''` pairs will be removed/filtered out.

to be clear - there is nothing wrong with this line in particular - but i detected it here.

freqtrade/strategy/interface.py
```diff
@@ -424,14 +424,18 @@ def gather_informative_pairs(self) -> ListPairsWithTimeframes:
         informative_pairs = self.informative_pairs()
```
xmatthias, line 424: This will fail - `informative_pairs()` is a user-exposed interface function - so we have to assume that we get tuples with length 2 back, not with 3.

having compatibility code for that should be pretty simple (but will be necessary), as that 3rd argument is really not necessary.

we should also be checking here if the value is correct (users can (and will) also provide invalid values ...) - not checking for that here will lead to a delayed error, which is probably hard to understand, as it's an ccxt error which does not tie back to the actual code that caused it.

---

allowing a 3rd value brings an interesting (but potentially complicated) point though:

we're now allowing a 3rd value to determine the candle type (i assume valid values will be `'', 'mark', 'index'`).

Assuming i'm in futures mode now though - i might like to get spot candles as well (which i see as more likely than using mark/index candles).
`''` will give me futures candles, not spot candles ... so as a user, i'd expect the possibility to pass `'spot'` here to get spot candles.
I don't think that's possible easily - as the ccxt exchange object was initialized with "type=futures" ... or is there a way to pass a param to switch this dynamically to either spot or futures (basically to the "Non-default") - without causing side-effects (assuming the pair is XRP-PERP - i'd not even have XRP/USDT in markets (there is no USDT futures pair on ftx) - but that's still a valid spot pair.
now ftx might be a bad example - but i hope you can apply this concept to exchanges where the switch is "hard" - meaning `ccxt.markets` don't contain the "other" markets.

freqtrade/data/dataprovider.py
```diff
@@ -50,21 +56,27 @@ def _set_cached_df(self, pair: str, timeframe: str, dataframe: DataFrame) -> Non
...
+    def historic_ohlcv(
+        self,
+        pair: str,
+        timeframe: str = None,
+        candle_type: str = ''
+    ) -> DataFrame:
         """
         Get stored historical candle (OHLCV) data
         :param pair: pair to get the data for
         :param timeframe: timeframe to get data for
```
xmatthias, line 78: as dataprovider is a user facing interface, i'd want to have both the docstring here, as well as the corresponding documentation updated

** this is valid for all public methods (methods not prefixed with `_`).

xmatthias (commented):
This PR is currently leaving more questions to me than it answers.

While we can (in theory) download mark/index candles - we can't really as there is no user interface to define that.
Also, futures candles (assuming i switch trading_mode to futures) will overwrite / mix with "regular" candles if the pair naming does not change between spot and futures (looking at binance in it's current state).

We should also define an allowed list of values for `candle_type` - we'd not want "random" stuff there, as it's a pretty short, clearly defined list that will work - which for now, i assume is `'', 'mark', 'index'` - but i'm not sure and didn't find that anywhere clearly (at least not in this PR).

freqtrade/data/history/hdf5datahandler.py
```diff
@@ -28,13 +28,22 @@ def ohlcv_get_available_data(cls, datadir: Path) -> ListPairsWithTimeframes:
...
-        _tmp = [re.search(r'^([a-zA-Z_]+)\-(\d+\S+)(?=.h5)', p.name)
-                for p in datadir.glob("*.h5")]
-        return [(match[1].replace('_', '/'), match[2]) for match in _tmp
+        _tmp = [
+            re.search(
+                r'^([a-zA-Z_]+)\-(\d+\S)\-?([a-zA-Z_]*)?(?=.h5)',
+                p.name
+            ) for p in datadir.glob("*.h5")
+        ]
+        return [(match[1].replace('_', '/'), match[2], match[3]) for match in _tmp
```
xmatthias: We will need to think for a solution to this.
right now, we assume to replace all `"_"` with `"/"`.
this works for spot-pairs - but is wrong for (future) futures pairs (i knew this for a while - and this "should" be the only place left with "/" assuming to split the pair)

If you look at the function in misc (`pair_to_filename()`) - we replace all special chars to "_".
this means, `BTC/USDT:USDT` will become `BTC_USDT_USDT` - which would translate back to `BTC/USDT/USDT` (now it's wrong, no longer matching the pairname).

Most likely, our approach will need to involve some ... 
more clever logic - maybe replacing ":" with something else to allow for "back-replacement".

freqtrade/data/dataprovider.py
samgermain: It could be useful to get the mark price in the strategy, and buy right before the funding times, to take advantage of the funding fee exchanges. That is the whole reason that funding fees work in the first place. It would be easier for the user to just give them the predicted funding fee though.

I don't think that was my intention behind editing this method though, it just looked like it might be needed for downloading candles

freqtrade/strategy/interface.py
samgermain: We'll probably save futures candles as `XRP/USDT:USDT`

tests/rpc/test_rpc_apiserver.py
samgermain: I added these files to `tests/testdata/`

- XRP_USDT-1h-mark.json
- XRP_USDT-1h.json
- UNITTEST_USDT-1h-mark.json

I think that it added the pairs `XRP_USDT` and `UNITTEST_USDT` got added. I don't really know what this method tests though, I understand the changes I made in this PR only at a shallow level for the most part. We would probably never want the mark price candles without the futures candles, but unless we hard code something that prevents that then we could have mark with the futures candles

freqtrade/strategy/interface.py
samgermain: I want to force 3 values until just before this PR is merged, it helps me catch errrors

freqtrade/data/history/hdf5datahandler.py
samgermain: We could just use a better regex

freqtrade/data/dataprovider.py
samgermain: What do you mean by `people will use alternative candle-types`? Are you asking if they will use them in their strategy?

freqtrade/strategy/interface.py
xmatthias: > I want to force 3 values until just before this PR is merged, it helps me catch errrors

it really won't as you've updated the tests to go along with it.
So now the tests are passing (no longer testing a 2 value tuple) - but the error will be there once an existing strategy is ran against this code.

tests/rpc/test_rpc_apiserver.py
xmatthias: the way i'd expect the above method to function is that it'll give you one more file (`XRP_USDT-1h.json`) - but NOT the other two (these are mark candles - should not be reported as regular candles).

freqtrade/strategy/interface.py
samgermain: I guess so, it's just helped me so far

freqtrade/data/history/hdf5datahandler.py
xmatthias: can/could also work - i'm just pointing out where i think there'll be problem for now - i don't "really" got solutions at this point :laughing:

samgermain:
> While we can (in theory) download mark/index candles - we can't really as there is no user interface to define that. 

I don't know what to do to write the user interface

> Also, futures candles (assuming i switch trading_mode to futures) will overwrite / mix with "regular" candles if the pair naming does not change between spot and futures (looking at binance in it's current state).

I'll make this a priority at CCXT, they want freqtrade futures to work also so we'll get a solution on unifying the symbols

> We should also define an allowed list of values for `candle_type` - we'd not want "random" stuff there, as it's a pretty short, clearly defined list that will work - which for now, i assume is `'', 'mark', 'index'`

`premiumIndex` would be the only other option, which might be needed someday if anyone wants to use Bybit

xmatthias:
> `premiumIndex` would be the only other option, which might be needed someday if anyone wants to use Bybit

I think another valid option should be `'spot'` ... just because we're trading futures doesn't mean that a look at the spot market won't make sense ... 
now it's not an absolute priority (and i'm not certain if it'll work with ccxt at the moment) - but still something we should keep in mind - as personally, i'd want to look at how spot markets are doing as well - not just stay within the futures candles (basically blind to the world around us).

samgermain:
@xmatthias I think that the spot candles should be like `ADA_USDT` and then the perpetual candles should contain the settle currency, like `ADA_USDT-USDT` or something

Also, the mark price is based on the spot price, it's like an average of the spot price across popular exchanges, excluding outliers or something. You can read about it here if you're interested https://www.binance.com/en/support/faq/360033525071

tests/rpc/test_rpc_apiserver.py
samgermain: I don't actually understand what the method `list_available_pairs` is for

samgermain:
> We should also define an allowed list of values for candle_type 

I could make an enum for it?

samgermain:
@xmatthias So since futures markets have unique symbols anyway, this PR might not even be needed until we implement FTX, funding fees are every 8 hours are binance, okex and gateio, i feel like that may be infrequent enough to just call the exchange everytime we need it like we're doing with dry run

And if we need to download the mark price, then would we also need to download the funding rates? Funding rates only have the timestamp, and one price though

xmatthias:
>  i feel like that may be infrequent enough to just call the exchange everytime we need it like we're doing with dry run

In my understanding, this information is needed for backtesting calculation. The pair name itself does not really have anything to do with that.
Doing "online" requests during backtesting may sound like a good idea at first, but will fail terribly.

Hyperopt is running backtesting over and over again (in multiple processes) - so we'd be doing tons of requests in parallel against the exchange (for the same timerange, so requesting always the same information) - which will
a) slow down hyperopt to no longer be useful
b) block us from the exchange, as default rate-limiting will fail (it's multiple processes, which means multiple exchange instances - which must be instanciated once for every backtest run (since we cannot transfer ccxt objects to hyperopt processes).

If the data is needed, and it's not a "one-off" value (like getting the fee) we can fetch at backtesting startup- we will need to download it first.

freqtrade/data/dataprovider.py
xmatthias: yes - in the strategy.

As said "elsewhere" (but in this PR) - i don't THINK they'll use these candles (except if you're trying to exploit the fees themselfs) - but potentially spot candles ...

tests/rpc/test_rpc_apiserver.py
xmatthias: the name is pretty explanatory i think :laughing: 
it lists all available pairs (it's basically what's called when using `freqtrade list-data`).

samgermain:
@xmatthias So how do you suggest we download the funding rates? The laziest fix would be to add a candle_type called `funding_rate`, called the `fetchFundingRateHistory` method when this type is passed, and add null to the high, low, close and volume and save them as candlesticks, it would mean a larger file size, but most exchanges will only give you max 3 funding rates a day (FTX will give you 24) so it's not really that much data.

I would prefer to do this because it would just be so much easier and I would like to not do more work than I need to.

xmatthias:
> So how do you suggest we download the funding rates

if it's a "one value per pair" thing (similar to fee) - then no. if it's a value changing over time - yes.
this is a value that's changing over time - so the answer is yes.

>  I would like to not do more work than I need to.

while i do understand this - you should also keep maintainability in mind, so i don't really care about "how much work" it is, if the resulting solution is a bad hack which can potentially cause problems in the long run (which i fear creating fake-data might).

To me, there is no rush to this - so it's better to take the time and make it right, rather than just trying to "get over with" - which in the end, results in throwing the problems the maintainers way - who then has to maintain the hacky/potentially buggy code.

I'll probably have a look at this myself as you seem to be missing the understanding of how this works (you said so yourself - which is a good base for a bad PR) - but i'm unclear as of now what exactly we're needing (the description is not very ... descriptive).
is it only mark candles (and potentially index candles if they exist) - and now funding rates - or is there something else we should consider?

samgermain:
just mark, index, and funding rate

The funding rate changes over time

freqtrade/data/dataprovider.py
samgermain: spot candles are kind of superfluous if mark price candles are available

tests/rpc/test_rpc_apiserver.py
samgermain: >  it lists all available pairs (it's basically what's called when using freqtrade list-data).

So whats the problem with showing mark pairs then

freqtrade/data/history/hdf5datahandler.py
samgermain: Unless windows doesn't `:` or `-` in file names, I don't think we need to change it

I updated the regex to detect `swap` and `futures` symbols

samgermain:
> you should also keep maintainability in mind, so i don't really care about "how much work" it is, if the resulting solution is a bad hack which can potentially cause problems in the long run (which i fear creating fake-data might).

Wouldn't it be less maintainable to create a whole separate functionality to store data that can be represented the same way that all the other data is represented?

freqtrade/strategy/interface.py
samgermain: @xmatthias I don't know how to make the third string optional. I tried these types

- `PairWithTimeframe = Union[Tuple[str, str], Tuple[str, str, str]]`
- `PairWithTimeframe = Tuple[str, str, Optional(str)]`

but just receive errors like

`Argument 1 to "refresh_latest_ohlcv" of "Exchange" has incompatible type "List[Tuple[str, str, str]]"; expected "List[Union[Tuple[str, str], Tuple[str, str, str]]]"`

tests/rpc/test_rpc_apiserver.py
xmatthias: showing mark candles is fine - IF the user elected to see mark candles.

without selection - only spot candles should be shown.

freqtrade/data/dataprovider.py
xmatthias: In my understanding, mark candles are not the same than spot candles - so i don't think that's a pretty good statement.
If they are - then why do we need mark candles, if we already have spot candles?

freqtrade/strategy/interface.py
xmatthias: Let me be honest - i think this is details on a PR that's greatly lacking overall, which can be left "till the end".
I think the fact that there is no description in the issue explains a bit how this PR is going.

I'm not demanding/asking for a description on a pr that's clear (where it's clear what it aims to provide), in this case however, it feels like you know we'll need this data (mark, index and funding rates), but have little idea / understanding on how this should be done.
That's not per se a problem - and this understanding can for sure be gained, or i can help with that. The way it is however is that you've just started "somewhere", and made 100ds of changes that may or may not be necessary in random places throughout the code - and are now losing yourself in details (such as the above) - which may not even be relevant for the end-result. 


I'm still not seeing much of an idea on the following topics:
* how will we make sure to always load the right data (reliably, across multiple exchanges with quite different naming standards). Right now, you're assuming that pairs are always different - which we cannot rely on - AND which is wrong if we look at binance, too. CCXT may change this on binance - but i understand it'll be non-breaking - which means BTC/USDT will remain a valid futures pair for the time being - which means we'll have to be sure we can reliably determine that it's a different pair.  
  We'll ALSO need a way to reliably filter pairs based on the data available - which is needed for several commands, so we'll need a reliable way from the filename/location to determine if the containing data is futures data, spot data, or mark/index candles.
* How can the user select which data to download? (user interface)   
  Maybe they'll simply specify `download-data <...>  futures`- and freqtrade will then download all data that's necessary for backtesting futures?   
  Alternatively, we'll have the user manually download all required data (which can be cumbersome and is most likely problematic)

---

I'm proposing the same again (what i already proposed somewhere here).
I'll take this PR from here (just give it time ... ) - you focus on "the other" PR's, where it's clearer what the methods you're changing are actually doing, and i'll take this one from here - as i honestly don't see this moving in the direction i'd like it to, and i see that you're struggling with this part of the bot.
This implicitly means, your time is better spent elsewhere ... and it's not like this is the "single last thing" that's needed for this big feature to be complete.

Hope that works for you.

freqtrade/data/history/hdf5datahandler.py
xmatthias: i don't think `:` is valid as character for [windows](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions), as that's what's used to specify drive letters (`C:/xxx`).

`-` is for sure valid - i'm i think we're replacing that with `_` for "reasons" - and not for something for real ...

freqtrade/data/dataprovider.py
samgermain: Mark candles, and index candles, are kind of like the mean of spot candle prices across popular exchanges

freqtrade/strategy/interface.py
samgermain: > CCXT may change this on binance - but i understand it'll be non-breaking - which means BTC/USDT will remain a valid futures pair for the time being - which means we'll have to be sure we can reliably determine that it's a different pair.

I opened a [PR on CCXT](https://github.com/ccxt/ccxt/pull/10720) which I think will more or less solve the issues with Binance futures haven't separate symbols, how it works is that symbols that look like `ADA/USDT` or `ADA/USDT:USDT` but when CCXT see's the `ADA/USDT` symbols when the defaultType is `future` or `delivery`, then it will convert the symbol to `ADA/USDT:USDT`. So it will be able to read the `ADA/USDT` symbols, but will always return symbols that look like `ADA/USDT:USDT` and this will probably continue for the near future, until we completely switch to `ADA/USDT:USDT` symbols and then the `ADA/USDT` symbols will no longer be accepted

freqtrade/data/history/hdf5datahandler.py
samgermain: Ok, I wrote the regex to accept both Perpetual Swap(ADA/USDT:USDT) and Delivery Futures(ADA/USDT:USDT-20220910) symbols, just because this is kind of the only part of this PR that would need accommodation for Delivery Futures, and it took me an extra couple minutes to implement the logic and save some future dev a lot of time going through these files

Feel free to remove the delivery futures accommodation if you like, because it's not needed right now

freqtrade/strategy/interface.py
xmatthias: yes, and this is how i understood it.
but this "intermediate" period is the problem. 
I cannot rely on a user to have the pairlist with `ADA/USDT:USDT` if the `ADA/USDT` is also valid.

No matter what ccxt will return - both pairs will be listed in the `markets` dictionary - which makes both correct.

samgermain: > 
The markets dictionary would only have the new pairs, the structure accepting a defaultType as `spot`, `margin`, `future` and `delivery` will still be in there in the intermediate period, so if you pass a symbol like `ADA/USDT` but your default type is `future`, then it grabs the symbol `ADA/USDT:USDT` instead and works with that, the user should never see the `ADA/USDT` symbols when using future or delivery unless they code those symbols themselves, 

The PR hasn't been merged, so something could possibly change, but this is the intent so far

freqtrade/data/history/hdf5datahandler.py
xmatthias: That regex is now no longer working (at least not for json data).
simply running `freqtrade list-data --exchange binance` on a user-data directory with 2 spot pairs in yields only an exception - because there is no longer a 2nd match.

I've fixed this now (and also simplified the whole sequence) ...

freqtrade/data/dataprovider.py
xmatthias: ok, so it's different information - and not the exchange's own spot candles (or a derivative of it).

As such, there can (and will) be usecases where this will be needed.

tests/test_misc.py
```diff
@@ -71,9 +71,10 @@ def test_file_load_json(mocker, testdatadir) -> None:
     ("ETHH20", 'ETHH20'),
     (".XBTBON2H", '_XBTBON2H'),
     ("ETHUSD.d", 'ETHUSD_d'),
-    ("ADA-0327", 'ADA_0327'),
-    ("BTC-USD-200110", 'BTC_USD_200110'),
-    ("F-AKRO/USDT", 'F_AKRO_USDT'),
+    ("ADA-0327", 'ADA-0327'),
+    ("BTC-USD-200110", 'BTC-USD-200110'),
+    ("BTC-PERP:USDT", 'BTC-PERP_USDT'),
```
samgermain, line 79: Don't worry about symbols like this, FTX will also be getting the unified symbols which look like `BTC/USD:USD`

xmatthias: Today? tomorrow? or sometime next month? By when will it be implemented for EVERY exchange in ccxt? (don't expect answers ...)

Sorry, but I am looking at current state - not at something that may happen sometime in the future.
Also, supporting this is not really an issue ... :)

samgermain: binance and ftx are the only different ones, I see a 0% chance of implementing everything thats needed for FTX on freqtrade before that happens

xmatthias: as said - it's there, no more code will be needed to support this - so i don't see a reason to remove it at this point in time.

samgermain: pretty quickly after the binance changes get merged, the functionality is almost identical, I'm just saying don't waste your time on the PERP symbols

freqtrade/commands/arguments.py
```diff
@@ -65,7 +65,7 @@
 
 ARGS_CONVERT_TRADES = ["pairs", "timeframes", "exchange", "dataformat_ohlcv", "dataformat_trades"]
 
-ARGS_LIST_DATA = ["exchange", "dataformat_ohlcv", "pairs"]
```
samgermain, line 68: Will this limit getting the mark/index price candles to what trading mode your using? I feel like they could be useful for spot, just as a reference

freqtrade/commands/arguments.py
```diff
@@ -65,7 +65,7 @@
...
+ARGS_LIST_DATA = ["exchange", "dataformat_ohlcv", "pairs", "trading_mode"]
```
samgermain, line 70: Will this limit getting the mark/index price candles to what trading mode your using? I feel like they could be useful for spot, just as a reference, but it's not super important to include that

freqtrade/exchange/exchange.py
```diff
@@ -1314,8 +1314,8 @@ def extract_cost_curr_rate(self, order: Dict) -> Tuple[float, str, Optional[floa
     # Historic data
 
     def get_historic_ohlcv(self, pair: str, timeframe: str,
-                           since_ms: int, is_new_pair: bool = False,
-                           candle_type: str = '') -> List:
+                           since_ms: int, candle_type: CandleType,
+                           is_new_pair: bool = False) -> List:
```
samgermain, line 1319: So if we are going record the funding rate a candle, we'll have to do something like 

```
if candle_type == CandleType.FundingRate:
    funding_rates = self._api.fetch_funding_rate_history(pair)
    candles = []
    for rate in funding_rates:
        candles.append([rate['timestamp'], rate['fundingRate'], null, null, null, null])
```

This is what I had in mind when I wrote these comments anyway `:param candle_type: '', mark, index, premiumIndex, or funding_rate`

I saw that you had FUNDINGRATE as a candle value, so I thought I should let you know what my idea was there

xmatthias: I do know this - i'm honestly not yet decided on this part - and if you look at the enum, it has a note there saying that funding_rate may not stay there.

freqtrade/commands/arguments.py
xmatthias: well here - surely not - as this line is only relevant for `list-data`.

the way i think about this currently is as follows:
you run `freqtrade download-data --exchange <whatever> --pairs XRP/USDT BTC/USDT` to get spot data (this is how it is also in develop).

For futures, you'll run `freqtrade download-data --exchange <whatever> --pairs XRP/USDT BTC/USDT --trading-mode futures` - which will download 2 things, the "futures" candles (stored in `user_data/data/binance/futures/XRP_USDT-futures.json` - and mark candles (or index, depending on the exchange - stored as `user_data/data/binance/futures/XRP_USDT-mark.json`).

For the moment, i think freqtrade will take the decision on which candles we'll download (so if backtesting needs 1h mark candles, we'll download exactly this).

From a usability perspective, i think this "automatism" will help, as the alternative would be to have users run 2 "download-data" commands, one for "futures" - and one for "mark"  (using the correct timeframe for the exchange used - the one we require for backtesting) - which doesn't really make sense, as the candles are mandatory for backtesting to work (and the only reason to download data is really to run backtesting ...).

I guess we COULD have this as argument to download-data too (so you can only download mark / index / whatever candles) - but it's not a necessity for now, and we'll have to think about usability here very carefully, as you need to use "futures" mode to even get mark candles (otherwise the exchange object is initialized with the wrong default type).

samgermain:
> Figure out how to deal with funding fees

You mean funding rates right?

xmatthias:
> You mean funding rates right?

for sure :D 
but that's what's written there ... (now) :laughing:

freqtrade/exchange/exchange.py
xmatthias: @samgermain I think i've now seen what needs to be done for funding-rate (as well as to get mark-candles in dry-run mode in a better way).

i'd however prefer to do that in a separate followup PR, as that'll be pretty focused on the exchange class (not all around like this one).

samgermain: Ok, sounds good

samgermain:
🥳

fridary:
@samgermain Is it error on `get_analyzed_dataframe()` where line `pair_key = (pair, timeframe, CandleType.SPOT)`? If I set `"trading_mode": "futures"` in config, I have `CandleType.FUTURES` in `pair_key`, not `CandleType.SPOT`. So during live trade I have errors `WARNING freqtrade.strategy.interface.get_latest_candle() Empty candle (OHLCV) data for pair ETH/USDT`
I guess changing `pair_key = (pair, timeframe, CandleType.SPOT)` -> `pair_key = (pair, timeframe, CandleType.get_default(self._config.get('trading_mode', 'spot')))`

xmatthias:
@fridary please make sure you're using the latest version (of the feat/short branch) - this has been fixed sometime this (or last?) week.

fridary:
@xmatthias I see it was updated in feat/short branch, indeed. But I used isolated-liq branch. What's the difference? I am trading futures with leverage.

xmatthias:
None of these is ready for users, or to be used with real money (at least not with anything substancial), and you should closely monitor every action the bot takes on all branches exceot stable and develop.

If you're not a developer trying to improve / develop /test new features, best stick to the develop/stable branch.

samgermain:
> @xmatthias I see it was updated in feat/short branch, indeed. But I used isolated-liq branch. What's the difference? I am trading futures with leverage.

I'm about to merge, and then the `isolated-liq` branch will have everything that the `feat/short` branch has, + non-working liquidation price functionality for isolated futures

-----------------

The `feat/short` branch would be your best bet right now, on isolated futures with a stoploss that would always be above the liquidation price, but everything won't work exactly right now, so you'll have to develop it

----------------------

The `isolated-liq` branch needs some work, if you're interested in helping

1. I want to reduce the number of parameters to the isolated_liq function

2. For binance, I don't know how to get the 
- maintenance margin rate or 
- maintenance amount

3. For okex and gateio, I don't know how to get 
- mm_ratio
- liability
- interest
- position_assets

fridary:
I checked `feat/shot`. There were few changes needed to fix for ccxt api, but anyway binance futures works good, long buy leverage orders are sent. Profit ratio calculation little bit incorrect.
I will try first make at least one profitable strategy, then hope to make contributes. I never try maintenance git projects together with other people, don't know how to manage my new functions with current `freqtrade`, buy anyway your work is amazing and sharing with people.

xmatthias:
> Profit ratio calculation little bit incorrect.

can you please elaborate this (best in an issue, not on a merged pull request that didn't really change anything in the profit calculations ...)
If possible, please with some samples on what you observed where it seems/is incorrect, thanks.
That would help us a ton already.

Issue #5784, opened 2021-10-24
/forebuy is buying pair which is not in white list.


GluTbl:
## Describe your environment

  * Operating system: ____
  * Python Version: _____ (`python -V`)
  * CCXT version: _____ (`pip freeze | grep ccxt`)
  * Freqtrade Version: ____ (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
  

## Describe the problem:

I send force buy to buy a pair which is not in whitelist. but it was executed. but the bot is crashed because the pair is not in whitelist

xmatthias:
Please provide us with the logs for this - so we can see why/where it crashed.

Also, please fill out the issue template - otherwise i'll have to assume you're on an outdated version - even though that might not be true.

Buying assets which are not in the whitelist is possible and allowed intentionally (it's the current user desire), so i don't see why that should crash the bot.
For sell-signals, all pairs with open trades will be added to the whitelist - so this will not crash the bot (just tested, it doesn't).
Which suggests it's a bug in the strategy - which the log would confirm.

A proper fix for that will therefore need to address the failure, not limit the bot to only pairs in the whitelist.

xmatthias:
@GluTbl are you able to provide the above logs? or are the logs "gone" - so we'll have to close this issue as "can't verify" ... ?

xmatthias:
without logs, there's nothing further we can look into.

Forcebuying a "non-whitelisted" pair does work on regular strategies, so i assume it's a problem of the strategy you're using.

Feel free to open or comment below if you're able to provide logs - so we can have a closer look at this.

PR #5785, opened 2021-10-24
disable /forcebuy for pairs not in whitelist

GluTbl:

## Summary

Checking if the pair is in whitelist.

Solve the issue: #5784

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -770,6 +770,9 @@ class Telegram(RPCHandler):
         """
         if context.args:
             pair = context.args[0]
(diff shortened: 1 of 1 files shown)


xmatthias (changes requested):
While i apreciate reporting bugs and trying to fix them autonomously, sometimes it's better to complete the investigation before running of on a fix that doesn't fix the problem.

This is allowed intentionally - as mentioned in the issue, with the logs we should be able to determine why this is happening - i'll however not be accepting this PR (not as is, anyway), as it's introducing an unnecessary limitation.

xmatthias:
As mentioned above - this is possible intentionally, and will not cause errors unless the strategy causes them - in which case that needs to be analyzed further.
This is however definitely not the right fix for it - i'm therefore closing this PR.

PR #5786, opened 2021-10-24
Clean up file handles

SimonEbner:
Clean up unclosed file handles. Close all file handles that are left dangling to avoid warnings such as
```
ResourceWarning: unclosed file <_io.TextIOWrapper
name='...' mode='r' encoding='UTF-8'> params = json_load(filename.open('r'))
```

### DIFF
--- a/freqtrade/optimize/hyperopt_tools.py
+++ b/freqtrade/optimize/hyperopt_tools.py
@@ -1,4 +1,3 @@
-
 import io
 import logging
 from copy import deepcopy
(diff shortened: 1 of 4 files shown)


xmatthias:
I'm not sure i've ever seen a warning like this ...
what command under what circumstances can generate this?

SimonEbner:
Running python with option `-Wd` wil show these.
Alternatively they can be enabled programatically via 
```
import warnings
warnings.simplefilter('always')
```

For me they popped up when writing a test using python's `unittest` module which enables the warnings by default

xmatthias (approved):
Great catch! :+1: 

i think there's still one in `test_set_loggers_Filehandler` - but that's caused by `RotatingFileHandler` - which i don't think we will be able to fix in our code.

PR #5795, opened 2021-10-25
Show a Volume Profile and split in Buy/Sell Volume in Plot

fxxmr:
## Summary

I missed a volume profile in the plot, so I added one

## Quick changelog

- Adding a volume profile on the right side of the main plot
- Allowing to split the volume bars in sell and buy volume
- Make everything configurable in the `plot_config`
- Adding some basic unit tests

### Volume Profile
You can activate a volume profile in the plot. This can be activated by `'showVolumeProfile': 'true'` in `plot_config`
volume Profile
 
### Split volume by sell and buy volume
 You can also split the 'normal' volume on the bottom of the plot in buy and sell volume. This can be activated by ` 'showBuySell': 'true'` in `plot_config`
volume_bars

### This are all new possible parameters in the plot_config:
plot_config = {
        'main_plot': { },
        'subplots': { },
        'volume': {
            'showBuySell': 'true', #split the volume graph on the bottom in buy and sell volume; default: false
            'showVolumeProfile': 'true', #show a volume profile on the right of the main plot; default: false 
            'VolumeProfileHistoryBars': 96, #define how many bars should be used for the volume profile; default: all bars in the plot
            'VolumeProfilePriceRangeSplices': 100 # define in how many 'slices' the price is split up; default: 50
        }
    }
By default, all this is disabled.

### Possible improvements
Next step could be to just show the volume profile for the zoomed-in area. Currently, the volume profile is always based on all data in the plot for the last x bars (see `VolumeProfileHistoryBars`). It would be cool to dynamically calculate the vp based on the currently selected are/zoom.

### DIFF
--- a/docs/plotting.md
+++ b/docs/plotting.md
@@ -211,6 +211,12 @@ Sample configuration with inline comments explaining the process:
             "RSI": {
                 'rsi': {'color': 'red'}
             }
+        },
+        'volume': {
+            'showBuySell': 'true',
+            'showVolumeProfile': 'true',
+            # 'VolumeProfileHistoryBars': 96,  # default: all
+            # 'VolumeProfilePriceRangeSplices': 100  # default: 50
         }
     }
 
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -432,8 +432,7 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
     fig['layout']['yaxis1'].update(title='Price')
     fig['layout']['yaxis3'].update(title='Volume')
     for i, name in enumerate(plot_config['subplots']):
-        fig['layout'][f'yaxis{5 + i }'].update(title=name)
-        # ToDo: if more then one subplot is used, the title is not shown
+        fig['layout'][f'yaxis{5 + i*2 }'].update(title=name)
     fig['layout']['xaxis']['rangeslider'].update(visible=False)
     fig.update_layout(modebar_add=["v1hovermode", "toggleSpikeLines"])
 
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -404,12 +404,24 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
     plot_config = create_plotconfig(indicators1, indicators2, plot_config)
     rows = 2 + len(plot_config['subplots'])
     row_widths = [1 for _ in plot_config['subplots']]
+
+    # load  VolumeProfile configurations
+    volume_config = plot_config['volume'] if 'volume' in plot_config else {}
+    showBuySell = volume_config['showBuySell'] if 'showBuySell' in volume_config else 'false'
+    showVolumeProfile = volume_config['showVolumeProfile'] if 'showVolumeProfile' in \
+        volume_config else 'false'
+    VolumeProfileHistoryBars = volume_config['VolumeProfileHistoryBars'] if \
+        'VolumeProfileHistoryBars' in volume_config else -1
+    VolumeProfilePriceRangeSplices = volume_config[
+        'VolumeProfilePriceRangeSplices'] if 'VolumeProfilePriceRangeSplices' in \
+        volume_config else 50
+
     # Define the graph
     fig = make_subplots(
         rows=rows,
         cols=2,  # ToDo: Check if 2 columns (instead of one) cause any issues somewhere else
-                 # ToDo: when showVolumeProfile==false, don't show the second column
-        column_widths=[8, 1],  # set the width of the Volume Profile
+        # set the width of the Volume Profile
+        column_widths=[8, 1 if showVolumeProfile.lower() == 'true' else 0],
         shared_xaxes=True,
         shared_yaxes=True,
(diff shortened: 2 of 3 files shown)


xmatthias:
In general, i'm not in favor of adding new functionalities to the plotting module at this point, as we'll be removing this way to do plots at some point in the near/mid term future in favor of freqUI (which allows basically the same plots we have at this moment, but with more interactivity, dynamically adding/removing indicators), while at the same time being able to handle more data.

It's not yet 100% feature-complete/identical - which is why the existing plotting modules have not yet been removed, however, it should be considered "in maintenance mode" - which means we should not be adding new features - as this will widen the gap we have to bridge for freqUI.

Instead, new functionality should be added to freqUI.

In this particular case - i can see us adding support for this here, but we would want to get a PR adding the same functionality to freqUI.

**edit**: to be clear - this does not mean we'll not merge this if we don't get a PR for freqUI - but we'll hold merging this until we find time to add it there, too.

freqtrade/plot/plotting.py
```diff
@@ -436,15 +536,68 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
...
+            name='VolumeProfile Sell',
+            marker_color='rgba(208, 41, 41, 0.6)',
+            marker_line_color='rgba(208, 41, 41, 0.6)',
+            orientation='h',
+        )
+
+        fig.add_trace(volume_sell, 1, 2)
+        fig.add_trace(volume_buy, 1, 2)
+        fig.update_layout(barmode='stack')
+
```
xmatthias: I think we might wanna extract this into it's own method - just for code simplicity's sake, as this is one of the methods that'll keep getting longer and longer ... :smile:

fxxmr:
> In this particular case - i can see us adding support for this here, but we would want to get a PR adding the same functionality to freqUI.
> 
> **edit**: to be clear - this does not mean we'll not merge this if we don't get a PR for freqUI - but we'll hold merging this until we find time to add it there, too.
 
I will check it out later this week and see if I can create a PR for it

freqtrade/plot/plotting.py
fxxmr: Agreed, I will separate it

docs/plotting.md
```diff
@@ -211,6 +211,12 @@ Sample configuration with inline comments explaining the process:
             "RSI": {
                 'rsi': {'color': 'red'}
             }
+        },
+        'volume': {
+            'showBuySell': 'true',
+            'showVolumeProfile': 'true',
```
xmatthias, line 217: wouldn't it be easier (both here and when handling it) to have it be real booleans (no quotes) - not strings?
Obviously it'll have to be `True` then ... but ...

freqtrade/plot/plotting.py
```diff
@@ -340,6 +354,115 @@ def add_areas(fig, row: int, data: pd.DataFrame, indicators) -> make_subplots:
...
 
+def generateBuySellVolumes(dataframe) -> pd.DataFrame:
+    candles = dataframe.copy()
+
+    candles['volume_buy'] = 0
+    candles['volume_sell'] = 0
+
+    for i in range(len(candles)):
+
+        candles['volume_buy'].iat[i] = (candles['volume'].iat[i] *
```
xmatthias, line 365: From a quick look, i think it should be possible to vectorize this (avoid looping).

You'll see that doing so will greatly improve the speed. 
i guess it'll look something along these lines:

```
candles['volume_buy'] = np.where(dataframe['high'] > dataframe['low'], dataframe['volume'] * <....>), 0)

```

freqtrade/plot/plotting.py
```diff
@@ -340,6 +354,115 @@ def add_areas(fig, row: int, data: pd.DataFrame, indicators) -> make_subplots:
...
+
+def add_volumeProfile(data: pd.DataFrame, fig, plot_config: Dict[str, Dict]):
+    # load VolumeProfile configurations
+    volume_config = plot_config['volume'] if 'volume' in plot_config else {}
+    showVolumeProfile = volume_config['showVolumeProfile'] if 'showVolumeProfile' in \
+        volume_config else 'false'
+    VolumeProfileHistoryBars = volume_config['VolumeProfileHistoryBars'] if \
+        'VolumeProfileHistoryBars' in volume_config else -1
+    VolumeProfilePriceRangeSplices = volume_config[
+        'VolumeProfilePriceRangeSplices'] if 'VolumeProfilePriceRangeSplices' in \
```
xmatthias, line 435: The naming of these 2 variables is very missleading / confusing, as it doesn't really adhere to python standards.

In python, it's common to use lowercase variable names (in `snake_case` formatting). 
capitalized names are usually [classes](https://www.python.org/dev/peps/pep-0008/#class-names) (this is also visible by the color highlighting github applies).

I'd therefore: leave the settings as they are (it's a json object, so capitalization is not that important)
Modify the python variables to follow at least some style (so at least color highlighting is not broken) - to avoid confusing your friendly reviewer 😆

lucca65:
is this still alive? looking for something exactly like this

xmatthias:
not ... really - and based on the discussion in #6845 - it's also not very likely to be accepted, as apparently the volume profile calculation is not accurate this way.

xmatthias:
Closing this based on the above comment.

PR #5798, opened 2021-10-25
Add function to unlock PairLocks by reason

ghost:
~~Added value "allow_position_stacking" to the config which allows rebuys of the same pair, ( default = False ) 
checkout the StackingDemo-Strategy.~~

Added function unlock_reason(str: pair) which removes all PairLocks with reason, 
might be useful in some cases.

~~Provided demo strategy StackingDemo.py to demonstrate buying the same pair multiple times, so you can look into this and decide if this is useful.~~


StackingConfig.json
```diff
@@ -0,0 +1,89 @@
+
```
xmatthias: this file is definitely in the wrong location.
if any, it should go to `config_examples/` - but only if it's significantly different from other examples.

StackingDemo.py
```diff
@@ -0,0 +1,591 @@
+# pragma pylint: disable=missing-docstring, invalid-name, pointless-string-statement
```
xmatthias: This file is definitely in the wrong location.

It doesn't seem to provide much differences to other samples though, so i think changes should be targeted and extracted and go to the documentation instead.

freqtrade/persistence/pairlock_middleware.py
```diff
@@ -103,6 +103,24 @@ def unlock_pair(pair: str, now: Optional[datetime] = None) -> None:
         if PairLocks.use_db:
             PairLock.query.session.commit()
 
+    @staticmethod
+    def unlock_reason(reason: str, now: Optional[datetime] = None) -> None:
```
xmatthias, line 107: This method is not related to position management, so best extract the change(s) required for `unlock_reason` into a seperate PR.

Mixing things that don't really belong together has 2 negative side-effects.
They make reviews difficult (why was this changed, for "a" or for "b"?), and make adding changes very tedious (problems or open points in one area can prevent merging of the other change).

xmatthias (changes requested):
i don't think this is the correct approach to implement position management, and implementing it this way will prevent the implementation of proper position management, as it would mean a breaking change.

As mentioned in #1519 , i think we'll need to combine the trades into one - and allow flexible entry/exit (so you can add to trades and exit trades dynamically (say - buy another 20% of your original position size).

This current approach forces you to exit at the same sizes you buy, so assuming the following buy scenario:

* buy 200$ XRP
* enter another 300$ XRP
(you now own 500$ worth of XRP)  forces you to exit etiher both at once (-500$) - or exit 200$ and 300$ respectively. You cannot however exit 250$ and 250$, or exit the trades in 3 or more steps (multiple take-profit targets).

I see this as pretty limiting, and thanks to your PR, i've now taken the decision for #1519 to go with the "trades must be combined".

From pr perspective, i think the best might be to remove the "position management" segments from this PR (this is not how we'll implement position-management), and convert this PR into a "unlocking" pr - adding the "unlock per reason" methods only.

ghost:
Changed as requested: removed the stacking stuff.

xmatthias (changes requested):
please restore the 2 removed files (I'm pretty sure it's not been intentional) ... they're kinda important for Freqtrade ... ;)

freqtrade/persistence/pairlock_middleware.py
```diff
@@ -103,6 +103,24 @@ def unlock_pair(pair: str, now: Optional[datetime] = None) -> None:
...
+        """
+        Release all locks for this reason.
+        :param reason: Which reason to unlock
+        :param now: Datetime object (generated via datetime.now(timezone.utc)).
+            defaults to datetime.now(timezone.utc)
+        """
+        if not now:
+            now = datetime.now(timezone.utc)
+        logger.info(f"Releasing all locks with reason \'{reason}\'.")
+        locks = PairLocks.get_all_locks()
```
xmatthias: I'm a little skeptical about the usage of `get_all_locks()` here.

A lock has 2 states to become inactive - either it expires - or it's `active` flag is set to False.
using `get_all_locks()` will get all locks (also expired ones) - and change the flag to False. 

That won't be a problem immediately, but it will
* taint history by disabling expired locks - so you can never know if a lock was disabled, or expired
* cause performance-problems should you run a bot for a longer time and have 1000nds of locks - as it'll loop through each and every one of these.

The performance problem is worsened by querying all locks (not just the ones with the reason we're looking for), which is also causing a lot of unnecessary looping.

ghost: I changed some things, as I figured that writing lines to console would reduce speed i removed the message for backtesting and replaced that for-loop with a map-function which also should make it a bit faster. 
However when i tried to use that same map-function in live mode (where PairLocks.use_db==True) it would not work and i don't know why, maybe you can give me a hint?
Applying the filter should however fix that history issue, and reduce the amount of comparisons that are happening.
Also you now get a message in console for each lock, which is a bit nicer as you see more of what is happening.

freqtrade/persistence/pairlock_middleware.py
```diff
@@ -103,6 +103,37 @@ def unlock_pair(pair: str, now: Optional[datetime] = None) -> None:
...
+            locks = PairLock.query.filter(*filters)
+            for lock in locks:
+                logger.info(f"Releasing lock for \'{lock.pair}\' with reason \'{reason}\'.")
+                lock.active = False
+            PairLock.query.session.commit()
+        else:
+            # no logging in backtesting to increase speed
+            locks = filter(lambda reason: reason == reason, PairLocks.locks)
+            locks = map(local_unlock, locks)
+
```
xmatthias: In general, i'd try to avoid filter/map where possible.

I know it's common in other languages, but in python, list-comprehensions are faster than map/filter.

freqtrade/persistence/pairlock_middleware.py
```diff
@@ -103,6 +103,37 @@ def unlock_pair(pair: str, now: Optional[datetime] = None) -> None:
...
+        if PairLocks.use_db:
+            # used in live modes
+            logger.info(f"Releasing all locks with reason \'{reason}\':")
+            filters = [PairLock.lock_end_time > now,
+                       PairLock.active.is_(True),
+                       PairLock.reason == reason
+                       ]
+            locks = PairLock.query.filter(*filters)
+            for lock in locks:
+                logger.info(f"Releasing lock for \'{lock.pair}\' with reason \'{reason}\'.")
```
xmatthias: there shouldn't be a need to escape quotes - as the wrapping quotes are from the "Other" type (double vs. single quotes).

freqtrade/persistence/pairlock_middleware.py
xmatthias: i'm not sure what you mean ... but i wouldn't use map/filter anyway (see other comment).

freqtrade/persistence/pairlock_middleware.py
ghost: Sorry, i made a mistake here. 
Using filter would require a cast to list, otherwise you get useless filter-object instead of a list, also i think that this creates a copy.

freqtrade/persistence/pairlock_middleware.py
ghost: i would keep \'{reason}\' for clarity, so you see exactly what the string contains

xmatthias: it's not about `'{reason}'` - but abouth `\'` .... :smile:

xmatthias (approved):
Nice addition, thanks a lot :100: 

I've added this to the documentation, and created a basic test for this to ensure we're not accidentally breaking this in the future.

PR #5826, opened 2021-11-01
[docs] Update RateLimit value [small]

Theagainmen:
## Summary
Fix very small mistake in docs [link to the specific part](https://www.freqtrade.io/en/latest/exchanges/#setting-rate-limits), that might confuse people. Let me know if this is the correct value now, there is still another `3100` in there at the Kraken section, which I think makes sense there and is correct.

## Quick changelog
Changed the `rateLimit` 3100 value to 200, to match the 200ms delay and thus 0.2s delay. The 3100 is a mistake, but I'm not 100% sure, just let me know if this is correct or if I should change it to something else.

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -41,12 +41,12 @@ In case of problems related to rate-limits (usually DDOS Exceptions in your logs
     "ccxt_config": {"enableRateLimit": true},
     "ccxt_async_config": {
         "enableRateLimit": true,
-        "rateLimit": 3100
+        "rateLimit": 200
     },
 ```
 
 This configuration enables kraken, as well as rate-limiting to avoid bans from the exchange.
-`"rateLimit": 3100` defines a wait-event of 0.2s between each call. This can also be completely disabled by setting `"enableRateLimit"` to false.
+`"rateLimit": 200` defines a wait-event of 0.2s between each call. This can also be completely disabled by setting `"enableRateLimit"` to false.
 
 !!! Note
     Optimal settings for rate-limiting depend on the exchange and the size of the whitelist, so an ideal parameter will vary on many other settings.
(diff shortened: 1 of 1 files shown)


docs/exchanges.md
```diff
@@ -41,12 +41,12 @@ In case of problems related to rate-limits (usually DDOS Exceptions in your logs
...
     "ccxt_async_config": {
         "enableRateLimit": true,
-        "rateLimit": 3100
+        "rateLimit": 200
     },
 ```
 
 This configuration enables kraken, as well as rate-limiting to avoid bans from the exchange.
-`"rateLimit": 3100` defines a wait-event of 0.2s between each call. This can also be completely disabled by setting `"enableRateLimit"` to false.
+`"rateLimit": 200` defines a wait-event of 0.2s between each call. This can also be completely disabled by setting `"enableRateLimit"` to false.
```
xmatthias: Unfortunately, it's the description that's wrong (0.2s) - not the rate-limit of 3100.

from the [kraken docs](https://docs.kraken.com/rest/#section/Rate-Limits/REST-API-Rate-Limits)

Starter | 15 | -0.33/sec
this means, you can do 15 calls "immediately" - and get a decrement of 1 every 3(ish) seconds.
So for batch downloads (assuming 30 pairs) - using 200ms as rate-limit will get you locked almost immediately (just try `download-data` with it to see the blocks almost instantly) - and the correct value is 3100.

While higher tiers/verification levels benefit from higher rate-limits (for placing orders) - these do not apply to public endpoints (ohlcv, orderbook, ...) - and are therefore not really relevant.

xmatthias (changes requested):

docs/exchanges.md
Theagainmen: Ok so the 3100 stays and it should be a delay of 3.1s, is that correct?

xmatthias: yes, that's correct

xmatthias (approved):
Great, thanks a lot for paying attention! 👍

PR #5832, opened 2021-11-04
setup.sh - Redhat

samgermain:
Updated setup script to work on Redhat (and i believe this includes all CentOS and Rhel operating systems)

Tested on an amazon ec2 t2.medium running Red Hat Enterprise Linux 8 (HVM), SSD Volume Type, 64 bit, with 2 vCPUs and 2GiB memory using Python3.9

The setup script did not work installing Redhat on a t2.micro. If anyone wants to use the setup script on a t2.micro, they should use ubuntu with Python3.9 instead

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 #encoding=utf8
 
+function echo_block() {
+    echo "----------------------------"
+    echo $1
+    echo "----------------------------"
+}
+
 function check_installed_pip() {
    ${PYTHON} -m pip > /dev/null
    if [ $? -ne 0 ]; then
-        echo "-----------------------------"
-        echo "Installing Pip for ${PYTHON}"
-        echo "-----------------------------"
+        echo_block "Installing Pip for ${PYTHON}"
         curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
         ${PYTHON} get-pip.py
         rm get-pip.py
@@ -37,9 +41,7 @@ function check_installed_python() {
 }
 
 function updateenv() {
-    echo "-------------------------"
-    echo "Updating your virtual env"
-    echo "-------------------------"
+    echo_block "Updating your virtual env"
     if [ ! -f .env/bin/activate ]; then
         echo "Something went wrong, no virtual environment found."
         exit 1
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -200,48 +201,39 @@ function reset() {
...
+    elif [ -x "$(command -v apt-get)" ]; then
         echo "Debian/Ubuntu detected. Setup for this system in-progress"
         install_debian
+    elif [ -x "$(command -v yum)" ]; then
+        echo "Red Hat/CentOS detected. Setup for this system in-progress"
+        install_redhat
     else
         echo "This script does not support your OS."
-        echo "If you have Python3.6 or Python3.7, pip, virtualenv, ta-lib you can continue."
+        echo "If you have Python version 3.6 - 3.9, pip, virtualenv, ta-lib you can continue."
```
xmatthias: as you've touched this ... freqtrade doesn't support 3.6 anymore - so i'd change this to 3.7...

xmatthias (commented):
I'm not sure if it's a must to support RHEL, and it's something that's never even been asked for.
It's popular in corporate environments - but not so much for private use.
You're not really allowed to use it for anything "productive" without a license (running a productive bot might count as productive) - and centos is clinically dead with recent developments with it.

I don't mind this PR (mainly since it cleans up the script quite some) - but i wouldn't invest much time into the setup.sh script at this point - it's working fine for ubuntu - and we don't have to support every other OS ...

samgermain:
> I'm not sure if it's a must to support RHEL, and it's something that's never even been asked for. It's popular in corporate environments - but not so much for private use. You're not really allowed to use it for anything "productive" without a license (running a productive bot might count as productive) - and centos is clinically dead with recent developments with it.
> 
> I don't mind this PR (mainly since it cleans up the script quite some) - but i wouldn't invest much time into the setup.sh script at this point - it's working fine for ubuntu - and we don't have to support every other OS ...

[Someone on stack overflow was using it](https://stackoverflow.com/a/68817056/6331353)

xmatthias (approved):
LGTM 👍

samgermain:
> and we don't have to support every other OS ...

I find that Docker is a bit of a barrier to entry, it stopped me from using freqtrade initially.
It kinda looks like most operating systems are built on RHEL or Debian, which I think mainly just leaves SUSE and Windows

samgermain:
Btw, if anyone tries to run amazon linux, I just tested this out using Python3.7, and I don't know which one was the problem, but installation of py-find-first failed. They might have to try [installing Python3.9 from source](https://tecadmin.net/install-python-3-9-on-amazon-linux/) if they want to use Amazon Linux

Installing 3.9 from source could also be a solution if they're having trouble installing 3.9 using apt-get, brew, yum ...

xmatthias:
I don't see this as a problem - all major cloud providers allow you to use Ubuntu.
If you opt to using another system, i assume people know what they're doing - and will be able to research that stuff.

samgermain:
> I don't see this as a problem - all major cloud providers allow you to use Ubuntu. If you opt to using another system, i assume people know what they're doing - and will be able to research that stuff.

I wouldn't, especially not windows users, because they might just be using windows because they haven't downloaded linux. I guess SUSE is uncommon enough that they might.

PR #5834, opened 2021-11-04
Update optimize_reports

raphant:
## Summary
Update show_backtest_reults() to preserve backwards compatibility by fixing KeyError: 'results_per_buy_tag' for older hyperopt result files.

## Quick changelog

Line 800: `if(results['results_per_buy_tag'] is not None)` > `if results.get('results_per_buy_tag') is not None:`

## What's new?
The new buy_tag changes in the backtest results will work with older hyperopt result files.


xmatthias (approved):
Great catch! 👍 
Thanks for the fix!

PR #5843, opened 2021-11-05
Update warning message open trades

Theagainmen:
## Summary

I saw a user in the Freqtrade discord sending the following screenshot and saying:
image
>Is there no way to have it keep running the trades if I just /start?
>so I have to do the /stopbuy or can I just /start?

I understand that maybe to (especially new) people this message can be confusing if they are only **reloading** their configuration file. The message is not really necessary when you just reload the config, since the bot will be up and running within a few seconds. If you think this change is unnecessary then just feel free to close this request.

## Quick changelog

- Don't send the `left open trades` warning message when the user is just reloading their config.
- Made the message slightly clearer when the bot is in dry run mode.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -193,21 +193,22 @@ class FreqtradeBot(LoggingMixin):
 
     def check_for_open_trades(self):
         """
-        Notify the user when the bot is stopped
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
LGTM :+1:

PR #5847, opened 2021-11-06
Hardcoded temp fix for Kucoin API issue

### LINKED ISSUE #5700: [Kucoin] {"code":"429000","msg":"Too Many Requests"}. Able to ignore this error and prevent DDOS protection?
## Describe your environment

  * Operating system: Canonical Ubuntu
  * Python Version: 
  * CCXT version: 1.56.86
  * Freqtrade Version: 
  
## Describe the problem:

{"code":"429000","msg":"Too Many Requests"} error triggering DDOS protection delay. 

### Steps to reproduce:
Randomly effect pairs on Kucoin

### Observed Results:

DDOS protection triggering when it doesn't need to which may cause unintended and financially ruinous behavior in the bot. 

Admin on telegram says we can retry the request instantly without fear of DDOS protection triggering. 

image

freqtrade    | 2021-10-11 04:55:04,274 - freqtrade.exchange.common - WARNING - _async_get_candle_history() returned exception: "kucoin GET https://openapi-v2.kucoin.com/api/v1/market/candles?symbol=DYDX-USDT&type=5min&startAt=1633778101&endAt=1633928101 429 Too Many Requests {"code":"429000","msg":"Too Many Requests"}"
freqtrade    | 2021-10-11 04:55:04,275 - freqtrade.exchange.common - WARNING - retrying _async_get_candle_history() still for 4 times
freqtrade    | 2021-10-11 04:55:04,275 - freqtrade.exchange.common - INFO - Applying DDosProtection backoff delay: 1

stash86:

## Summary

Keep the DDosProtection count as normal, to avoid triggering DDosProtection backoff delay, since we know it's an issue on Kucoin side, and their support suggest so

closes #5700 

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?
This PR will make sure the bot doesn't trigger the DDos backoff, so it will keep trying to pull the data from Kucoin API. I'm keeping the old code commented, so once Kucoin fix their issue, we can revert back to the old code

### DIFF
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -81,9 +81,16 @@ def retrier_async(f):
                 count -= 1
                 kwargs.update({'count': count})
                 if isinstance(ex, DDosProtection):
-                    backoff_delay = calculate_backoff(count + 1, API_RETRY_COUNT)
-                    logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
-                    await asyncio.sleep(backoff_delay)
+                    if "kucoin" in str(ex) and "429000" in str(ex):
+                        # Temporary fix for 429000 error on kucoin
+                        # see https://github.com/freqtrade/freqtrade/issues/5700 for details.
+                        logger.warning(
+                            f"Kucoin 429 error, avoid triggering DDosProtection backoff delay. "
+                            f"{count} tries left before giving up")
+                    else:
+                        backoff_delay = calculate_backoff(count + 1, API_RETRY_COUNT)
+                        logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
+                        await asyncio.sleep(backoff_delay)
                 return await wrapper(*args, **kwargs)
             else:
                 logger.warning('Giving up retrying: %s()', f.__name__)


freqtrade/exchange/common.py
```diff
@@ -81,14 +80,39 @@ async def wrapper(*args, **kwargs):
...
+#                 kwargs.update({'count': count})
+#                 if isinstance(ex, DDosProtection):
+#                     backoff_delay = calculate_backoff(count + 1, API_RETRY_COUNT)
+#                     logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
+#                     await asyncio.sleep(backoff_delay)
+#                 return await wrapper(*args, **kwargs)
+#             else:
+#                 logger.warning('Giving up retrying: %s()', f.__name__)
+#                 raise ex
+#     return wrapper
```
xmatthias: we'd clearly want this to be gone ... it's a "backup copy" of the original code ... but thats what we got git for :)

freqtrade/exchange/common.py
```diff
@@ -81,14 +80,39 @@ async def wrapper(*args, **kwargs):
                 count -= 1
                 kwargs.update({'count': count})
                 if isinstance(ex, DDosProtection):
-                    backoff_delay = calculate_backoff(count + 1, API_RETRY_COUNT)
-                    logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
-                    await asyncio.sleep(backoff_delay)
+                    if "kucoin" in str(ex) and "429000" in str(ex):
+                        logger.warning(f"Kucoin 429 error, keeping count the same.")
+                        count += 1
```
xmatthias: I am sure this fixes the issue for yourself (at least partially) - however - it'll make the issue overall worse - as the rate-limit is a global one (3000r/s for all users) - so if everyone starts using this, you might not get a value at all.

It's also assuming that retrying immediately will work - but assuming the exchange has a generic problem (maybe because everyone tries this, and it's then no longer responding) you've blocked the bot in this loop to never get out from.
So i think a "overall" limit should still be applied to this (give up after X requests) to avoid blocking a bot completely.

xmatthias (changes requested):
the "dead code" clearly has to be gone.

I'm not 100% sure if this will really fix the problem if applied on a bigger scale ... and apparently, kucoin is rolling out a fix this upcomming week to address this issue in a different way.

freqtrade/exchange/common.py
stash86: Done. I was waiting to see if Kucoin gonna fix the issue this week (according to their original timeline), but it seems unlikely.

freqtrade/exchange/common.py
stash86: Done. Yeah, it's a dangerous path to take

xmatthias:
I've rebased all of the unrelated commits out of this PR - as we don't need 300+ commits with 0 change in the end.

---

On a personal note - you should be very careful (and very sure) that none of the commits included a configuration with exchange keys - otherwise they must be considered exposed (even with rebase, the history will still be available on github).

I'd suggest using a separate, (private?) "user_data" repository (still not committing keys!) - so you have a freqtrade repository with the code, and a separate one with your strategies, configs and keys. That will make that part independent of the version you're using (stable, develop, ...).
If you intend to share the strategies / configurations, that'll also be easier as it's not "squashed" into freqtrade itself - and can be run on a version the user desires.

xmatthias (approved):
We'll add this for now - but should it show to create more problems (overloading the exchange API even more), i'll remove this fix again.

PR #5851, opened 2021-11-06
Raise exception when an uncaught error is raised

raphant:
## Summary
**StrategyResolver.load_strategy** would (practically) silently fail if any error outside of (**ModuleNotFoundError**, **SyntaxError**, **ImportError**, **NameError**) is thrown.

## Quick changelog
Raise an **OperationalException** in **IResolver._get_valid_object** on the`**spec.loader.exec_module(module)** line

## What's new?
The error in my case was a **FileNotFoundError** because I was using **load_strategy** outside of the FT directory and had no idea until I did a step-by-step debug.
Now when that happens it will raise an **OperationalException** with relevant information when that happens.

### DIFF
--- a/freqtrade/resolvers/iresolver.py
+++ b/freqtrade/resolvers/iresolver.py
@@ -69,6 +69,8 @@ class IResolver:
             logger.warning(f"Could not import {module_path} due to '{err}'")
             if enum_failed:
                 return iter([None])
+        except Exception as e:
+            raise OperationalException(f"Error loading {module_path}: {e}") from e
 
         valid_objects_gen = (
             (obj, inspect.getsource(module)) for
--- a/freqtrade/resolvers/iresolver.py
+++ b/freqtrade/resolvers/iresolver.py
@@ -70,7 +70,9 @@ class IResolver:
             if enum_failed:
                 return iter([None])
         except Exception as e:
-            raise OperationalException(f"Error loading {module_path}: {e}") from e
+            raise OperationalException(
+                f"Error loading {module_path}: {e.__class__.__name__}{e}"
+            ) from e
 
         valid_objects_gen = (
             (obj, inspect.getsource(module)) for


xmatthias:
to be honest, i'm not quite sure on the error you received ... Maybe you can outline how to reproduce this?

raphant:
I tried to use load_strategy outside of the FT directory:

```python
# tests.py
ft_config = {}
ft_config.update(
    {'user_data_dir': paths.USER_DATA_DIR, 'strategy_path': paths.STRATEGY_DIR}
)
def test_load_strategy():
    strategy = Populator._load_strategy(ft_config, 'DefaultStrategy')
    assert isinstance(strategy, IStrategy)

# populator.py
class Populator:
    @staticmethod
    def _load_strategy(ft_config: dict, strategy_name: str):
        ft_config = ft_config.copy()
        ft_config["strategy"] = strategy_name
        strategy = StrategyResolver.load_strategy(ft_config)
        return strategy
```
StrategyResolver.load_strategy would fail to find DefaultStrategy because another Strategy that I have raised a FileNotFoundError (because I didn't catch that error in the strat).

It would get to this part: **module = importlib.util.module_from_spec(spec)** in iresolver.py and raise a FileNotFoundError.
Because that's not caught:
```python
# iresolver.py _get_valid_object
        try:
            spec.loader.exec_module(module)  # type: ignore # importlib does not use typehints
        except (ModuleNotFoundError, SyntaxError, ImportError, NameError) as err:
            # Catch errors in case a specific module is not installed
            logger.warning(f"Could not import {module_path} due to '{err}'")
            if enum_failed:
                return iter([None])
```
and it would stop iterating at that point, but wouldn't notify me of the issue.

raphant:
On second thought, maybe it would be a better idea to give a warning instead of raising an exception?

xmatthias:
Letme ask this question differently - how do i have to place my directories / files so a default usecase (running freqtrade from the comand line - e.g. `freqtrade trade -s DefaultStrategy`) would trigger this?

I'm not sure how you can get to a "FileNotFound" error - as the call to this method is based on an iteration of files in a directory 1 call "above" this - so you'd have to delete a file right between these 2 calls ... which might be possible if you run with debugger - but i don't think is possible in other cases.

raphant:
Let's say you have a file myclass.py in user_data/strategies, that tries to load another file that lives in the same directory.
Let's also assume the following:
1. **We will be using a jupyter-notebook in a freqtrade root subdirectory (e.g. freqtrade/notebooks) and call StrategyResolver.load_strategy to load strategyA**
2. strategyA.py is the strategy we want to load
3. myclass.py will fail with a FileNotFoundError when imported
4. strategyB.py will import from myclass.py
5. All of these files are in the same directory (user_data/strategies)

The result is that when **StrategyResolver.load_strategy** is called from the notebook and iterates over strategyB.py, strategyB will raise the FileNotFoundError from myclass.py and that will break the (_get_valid_object) iteration on the spot without any relevant information to the user. 
It will say that strategyA is impossible to load even though strategyA has no errors (strategyB -> myclass.py does). 

I've already fixed the myclass.py problem by getting the code's directory using `__file__`, but the break in the iteration is my concern because the error returned is unhelpful in finding the issue.

xmatthias:
Please understand that i'm hesitant to accept a "catch all clause" without being able to recreate the issue.

Using the following strategy sample (it's supposed to load "NoActionStrategy" - which is NOT in a module/file called asdf) - i don't get a "FileNotFound" error - but a generic "import failed" error.

``` python
sys.path.append(str(Path(__file__).parent))
# print(sys.path)
from .asdf import NoActionStrategy
# from freqtrade.strategy import IStrategy

class NoActionStrategySL(NoActionStrategy):
    pass
```


``` output
2021-11-08 19:24:16,056 - freqtrade.resolvers.iresolver - WARNING - Could not import strategies/NoActio
nStrategySL.py due to 'attempted relative import with no known parent package'
```

This makes me believe that you're doing something funny/odd in your "failing" strategy - most likely directly loading a file that doesn't exist without checking that first.

If that's not the case, then please draft an example (similar to mine above) of what i need to do to make the error you're seeing appear.

xmatthias:
Based on the inactivity here i'll assume that you were able to fix your issues and this PR is no longer relevant.

Feel free to reopen / let me know so we can reopen if that's not the case.

PR #5853, opened 2021-11-07
Added okex exchange class futures properties

samgermain:
Adding futures properties to Okex exchange class


xmatthias (approved):
Now that's an easy one :laughing:

PR #5885, opened 2021-11-13
Include more details in _FILL telegram notifications

### LINKED ISSUE #5883: Include more details in _FILL telegram notifications
## Describe the enhancement
When running strategies on market orders, slippage can occur, rendering all of the information in the buy/sell notifications inaccurate.

Please include the same information that is in the buy/sell order notifications in the buy/sell FILL notifications. This way user can see the **actual** profit %. It is not enough to only show the price the order was filled at, as user then needs to do math to calculate what the actual result is after slippage

Something along the lines of this:
🔵 Kucoin: **Filled Buy** OMG/USDT (#4)
Buy Tag: 46 
Amount: 63.89690000
Open Rate: 13.35900000
Current Rate: 14.35900000
Slippage: 1 (7.69%)
Total: (853.599 USDT, 853.599 USD)

✳ Kucoin: **Filled Sell** OMG/USDT (#4)
Profit: 4.85% (profit: 41.19867642 USDT / 41.199 USD)
Sell Reason: sell_profit_u_bear_4_1 ( 46 )
Duration: 0:09:28 (9.5 min)
Amount: 63.89690000
Open Rate: 13.28000000
Current Rate: 13.95200000
Close Rate: 13.95200000
Slippage: XX (x.x%)

ethanopp:
## Summary

Include more details in _FILL telegram notifications

closes #5883 

## Quick changelog

- "Unrealized Profit" title for buy/sell order notifications
- Enhanced BUY_FILL / SELL_FILL notifications containing actual profit

## What's new?

"buy/sell" telegram notifications previously were misleading for market orders and did not account for slippage. These notifications "Profit" are now titled "Unrealized Profit". Additionally, "buy/sell_FILL" notifications have now been given more details of the filled trade, so users can run _only_ fill notifications and still receive all relevant info of the trade.

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -219,6 +219,7 @@ class Telegram(RPCHandler):
         else:
             msg['stake_amount_fiat'] = 0
 
+        msg['currency'] = msg['pair'].split('/')[0]
         content = []
         content.append(
             f"\N{LARGE BLUE CIRCLE} *{msg['exchange']}:* Buying {msg['pair']}"
@@ -226,7 +227,7 @@ class Telegram(RPCHandler):
         )
         if msg.get('buy_tag', None):
             content.append(f"*Buy Tag:* `{msg['buy_tag']}`\n")
-        content.append(f"*Amount:* `{msg['amount']:.8f} {msg['pair'].split('/')[0]}`\n")
+        content.append(f"*Amount:* `{msg['amount']:.8f}` {msg['currency']}\n")
         content.append(f"*Open Rate:* `{msg['limit']:.8f}`\n")
         content.append(f"*Current Rate:* `{msg['current_rate']:.8f}`\n")
         content.append(
@@ -241,13 +242,14 @@ class Telegram(RPCHandler):
         message += ")`"
         return message
 
-    def _format_buy_msg_fill(self, msg: Dict[str, Any]) -> str:
+    def _format_buy_fill_msg(self, msg: Dict[str, Any]) -> str:
         if self._rpc._fiat_converter:
             msg['stake_amount_fiat'] = self._rpc._fiat_converter.convert_amount(
                 msg['stake_amount'], msg['stake_currency'], msg['fiat_currency'])
         else:
             msg['stake_amount_fiat'] = 0
 
+        msg['currency'] = msg['pair'].split('/')[0]
         content = []
         content.append(
             f"\N{CHECK MARK} *{msg['exchange']}:* Bought {msg['pair']}"
@@ -255,7 +257,8 @@ class Telegram(RPCHandler):
         )
         if msg.get('buy_tag', None):
             content.append(f"*Buy Tag:* `{msg['buy_tag']}`\n")
-        content.append(f"*Amount:* `{msg['amount']:.8f} {msg['pair'].split('/')[0]}`\n")
+        content.append(f"*Amount:* `{msg['amount']:.8f}` {msg['currency']}\n")
+        content.append(f"*Open Rate:* `{msg['open_rate']:.8f}`\n")
         content.append(
             f"*Total:* `({round_coin_value(msg['stake_amount'], msg['stake_currency'])}"
         )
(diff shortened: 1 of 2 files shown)


freqtrade/rpc/telegram.py
```diff
@@ -242,6 +242,35 @@ def _format_buy_msg(self, msg: Dict[str, Any]) -> str:
...
         return message
 
+    def _format_buy_fill_msg(self, msg: Dict[str, Any]) -> str:
+        if self._rpc._fiat_converter:
+            msg['stake_amount_fiat'] = self._rpc._fiat_converter.convert_amount(
+                msg['stake_amount'], msg['stake_currency'], msg['fiat_currency'])
+        else:
+            msg['stake_amount_fiat'] = 0
+
+        msg['currency'] = msg['pair'].split('/')[0]
```
xmatthias: err - no.
Getting stake-currency like this is something we moved away from quite some time (it's fluky, and only works with certain base assumptions on the exchange used) - and will no longer work when futures/shorts land - so this is introducing a bug in my view.

If you really need to use this, then please add it to the originating dictionaries (in `freqtradebot.py`) - calculated with `exchange.get_pair_quote_currency(pair)` and/or `exchange.get_pair_base_currency(pair)`.

I'd also not call that key "currency" - but "asset" (or `base_currency` - or something similar more generic) - as some "coins" are not really meant to be currencies, but are tokens (Like the fan-tokens we have), and "currency" is quite generic, so i'd not be sure if you now mean fiat-currency, stake-currency, or the coin we're buying

freqtrade/rpc/telegram.py
```diff
@@ -262,19 +291,52 @@ def _format_sell_msg(self, msg: Dict[str, Any]) -> str:
...
+                                   ' / {profit_fiat:.3f} {fiat_currency})').format(**msg)
+        else:
+            msg['profit_extra'] = ''
+        msg['currency'] = msg['pair'].split('/')[0]
+        message = ("{emoji} *{exchange}:* Sold {pair} (#{trade_id})\n"
+                   "*Profit:* `{profit_ratio:.2%}{profit_extra}`\n"
+                   "*Buy Tag:* `{buy_tag}`\n"
+                   "*Sell Reason:* `{sell_reason}`\n"
+                   "*Duration:* `{duration} ({duration_min:.1f} min)`\n"
+                   "*Amount:* `{amount:.8f}` {currency}\n"
```
xmatthias: looking at this, it seems awfully close to the format we use for regular sell messages (the whole method, actually).

Maybe try to combine the two by providing a parameter to `_format_sell_msg` - and only change what's really different (like the header) ... (obviously the same should work for the buy-side as well).
otherwise we'll have these messages "drift apart" over time - as future editors will potentially (for sure?) not remember to change both messages.

freqtrade/rpc/telegram.py
```diff
@@ -262,19 +291,52 @@ def _format_sell_msg(self, msg: Dict[str, Any]) -> str:
...
                    "*Open Rate:* `{open_rate:.8f}`\n"
                    "*Current Rate:* `{current_rate:.8f}`\n"
                    "*Close Rate:* `{limit:.8f}`").format(**msg)
 
         return message
 
+    def _format_sell_fill_msg(self, msg: Dict[str, Any]) -> str:
+        import pprint
+        pprint.pprint(msg)
```
xmatthias: debug msg?

ttraxxrepo: Whoops yes will remove this, was for debugging

freqtrade/rpc/telegram.py
ttraxxrepo: Yeah they are almost identical, I actually initially started with just modifying the main def's but then broke out into 2 as I thought that is what would be preferred...guess not hah... Will combine back together

freqtrade/rpc/telegram.py
ttraxxrepo: Not a necessity, will remove for now

freqtrade/rpc/telegram.py
xmatthias: 90% of the content is identical - having 2 separate functions will not make sense, then ;)

freqtrade/rpc/telegram.py
```diff
@@ -221,25 +220,48 @@ def _format_buy_msg(self, msg: Dict[str, Any]) -> str:
...
+            content.append(
+                f"*Total:* `({round_coin_value(msg['stake_amount'], msg['stake_currency'])}"
+            )
+            if msg.get('fiat_currency', None):
+                content.append(
+                    f", {round_coin_value(msg['stake_amount_fiat'], msg['fiat_currency'])}"
+                )
+
+            message = ''.join(content)
+            message += ")`"
```
xmatthias: You should be able to move most of these lines out of the if/else branch (most are valid for both buy and buy_fill).
In my understanding, that should be possible for all lines except for the first (where you wanna differ between buy and buy fill) - and maybe one additional line which shows the "actual close" price/rate/profit...

the same should apply to sells - but you might need to refactor it to be a list as well, to be able to work with it easily.

ttraxxrepo: Not entirely sure I'm understanding that feedback, but consolidated it some more so hopefully thats ultimately what you were getting at

xmatthias: Yes it is - the benefit of this is - should we decide to change some wording in the future (or add aditional information) - it won't be missed to have it in both versions (if necessary) ...

freqtrade/rpc/telegram.py
```diff
@@ -258,54 +260,57 @@ def _format_sell_msg(self, msg: Dict[str, Any]) -> str:
...
-                   "*Sell Reason:* `{sell_reason}`\n"
-                   "*Duration:* `{duration} ({duration_min:.1f} min)`\n"
-                   "*Amount:* `{amount:.8f}`\n"
-                   "*Open Rate:* `{open_rate:.8f}`\n"
-                   "*Current Rate:* `{current_rate:.8f}`\n"
-                   "*Close Rate:* `{limit:.8f}`").format(**msg)
+        is_fill = msg['type'] == RPCMessageType.SELL_FILL
+        message = (
+            f"{msg['emoji']} *{msg['exchange']}:* "
+            f"{'Sold' if is_fill else 'Selling'} {msg['pair']} (#{msg['trade_id']})\n"
```
xmatthias, line 271: I've made a small change to this logic, as using `message.replace()` can lead to problems (you're replacing "Sold" with "selling" - but we can't control the content of BuyTag / SellTag ... so we might change the user input, too.

We do control the message though, so we can simply create the message correctly right away.

I've also taken the opportunity to change this to fstrings - which i find easier to work with.

xmatthias (approved):
great job :100:

LGTM :+1:

MunkeyBalls:
@xmatthias , was it intentional that there is no "open_rate" on the sell_fill message? 

Also, I've added Min/Max rate to my own fork, would you be interested in a pull request?
```
✳ Gateio: Selling SCLP/USDT (#134)
Unrealized Profit: 1.39% (profit: 0.75987065 USDT / 0.760 USD)
Buy Tag: 5 
Sell Reason: sell_profit_p_bear_48_2_1_1 ( 5 )
Duration: 0:31:08 (31.1 min)
Amount: 9.14500000
Open Rate: 5.97500000
Current Rate: 6.07979000
Min Rate: 5.95202000 (-0.74%)
Max Rate: 6.07979000 (1.39%)
Close Rate: 6.07979000
```

xmatthias:
i think what you really care about is the profit, not the open rate (not even the close rate, really ...).

MunkeyBalls:
> i think what you really care about is the profit, not the open rate (not even the close rate, really ...).

In the end that's all I care about. The sell message gets shared a lot in the community though, so i thought having important details on the trade could help strategy makers analyze better how the strategy is performing live.

PR #5887, opened 2021-11-13
Common ccxt config

samgermain:
## Summary

Creates a common ccxt config property for exchange classes, so that any exchange will create a proper _ccxt_config. 

- For exchanges that return all spot, perpetual swap, and futures markets when fetchMarkets is called, the `defaultType` property is not used, so there is no effect from passing in `"options": {"defaultType": "..."}` when initializing these exchanges

## Quick changelog

- Adds property `_ft_has.ccxt_futures_name` to exchange 
    - possible values are `swap`(default), `future`(binance) and `linear`(bybit)
- Creates default values for exchange._ccxt_config for any exchange, using the value in `_ft_has.ccxt_futures_name` for futures config


freqtrade/exchange/exchange.py
```diff
@@ -69,6 +69,7 @@ class Exchange:
         "trades_pagination_arg": "since",
         "l2_limit_range": None,
         "l2_limit_range_required": True,  # Allow Empty L2 limit (kucoin)
+        "ccxt_futures_name": "swap"
```
xmatthias, line 72: is the default for this in ccxt swap, not futures? that's kindof surprising me (but not a problem - just a question) as i was of the opinion that "futures" is the default, and some exchanges use swap ... :laughing:

samgermain: > is the default for this in ccxt swap, not futures? that's kindof surprising me (but not a problem - just a question) as i was of the opinion that "futures" is the default, and some exchanges use swap ... 😆

So futures is actually the incorrect term for perpetual futures, perpetual futures are actually perpetual swaps, and the only ones that are actually futures are the ones with an expiry date, I had a big conversation about it with the other CCXT people. They also like going with single words so that's why it's swap, there's a chance that the name could change, but for now it's swap. According to the other CCXT people, Binance is making a slow switch away from calling them futures, the ccxt binance futures code was the first written though, so they went with `future` and `delivery`, but we're trying to think of a way of switching over to the unified naming

xmatthias (approved):
I like cleanup :laughing:

Issue #5856, opened 2021-11-07
Unlimited stake amount does not divide equally


Theagainmen:
## Describe your environment

  * Operating system: Debian 11
  * Python Version: 3.9 (`python -V`)
  * Freqtrade Version: 2021.10 (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
  
Note: All issues other than enhancement requests will be closed without further comment if the above template is deleted or not filled out.

## Describe the problem:

I am using multiple bots with unlimited `stake_amount`, and `max_open_trades` of 7. Now there seem to be a pair, in this case KRL/USDT that buys for around half my balance. Even though it should only buy with a stake of around 1/7 of my balance. I'm using Kucoin. 

Here are the relevant parts of my configuration:
```    
    "max_open_trades": 7,
    "stake_currency": "USDT",
    "stake_amount": "unlimited",                                    
    "tradable_balance_ratio": 0.99,
    "fiat_display_currency": "EUR",
    "dry_run": true,
    "dry_run_wallet": 125,        
    "timeframe": "5m",
```

With a total `/balance` of:  `USDT: 141.731 (13.39%)`

And here a screenshot where you can spot the issue right away:
![image](https://user-images.githubusercontent.com/24569139/140645978-0e4835e0-1169-4edf-b1ee-094724c81dae.png)

### Steps to reproduce:

  1. Set `stake_amount` to `unlimited` and set `max_open_trades` to `7`
  2. Set your dry run wallet to `125` USDT
  3. Use Kucoin exchange and have your bot buy the pair `KRL/USDT`
  4. Have your bot buy other pairs too, and compare the stake buy prices of each pair.
  
### Observed Results:

  * What happened?
It bought a pair for almost half my balance, while I expect it to only buy a pair of around `19 USDT`, just like the other pairs. 

  * What did you expect to happen?
I expect it to buy pairs around `1/7` (with 7 being my `max_open_trades` of my total available balance on Kucoin, so at that time of the screenshot it should have bought pairs for around `19 USDT`.

### Relevant code exceptions or logs

Note: Please copy/paste text of the messages, no screenshots of logs please.

Here are the relevant logs:
```
2021-11-07 12:02:17,150 - freqtrade.wallets - INFO - Wallets synced.
[...]

PR #5888, opened 2021-11-13
Convert contract size to underlying asset size

samgermain:
## Summary

Some exchanges express their minimal amount in terms of contract, for example, if you run `node ccxt/examples/js/cli gateio fetchMarkets '{"type": "swap"}` you can see a field called `contractSize`, and for BTC this value is `0.0001`. This means that 1 contract of BTC is worth 0.0001 BTC, and the minimal order is for 1 contract. It makes more sense for us to handle all exchanges as ordering the amount of the underlying asset (this is how binance does it anyway). Also all prices are given in relation to the underlying asset anyway

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -399,22 +399,16 @@ class Exchange:
 
     def _amount_to_contracts(self, pair: str, amount: float):
 
-        contract_size = None
-        if ('contractSize' in self.markets[pair]):
-            contract_size = self.markets[pair]['contractSize']
-
-        if (contract_size and self.trading_mode == TradingMode.FUTURES):
+        contract_size = self._get_contract_size(pair)
+        if contract_size and contract_size != 1:
             return amount / contract_size
         else:
             return amount
 
     def _contracts_to_amount(self, pair: str, num_contracts: float):
 
-        contract_size = None
-        if ('contractSize' in self.markets[pair]):
-            contract_size = self.markets[pair]['contractSize']
-
-        if (contract_size and self.trading_mode == TradingMode.FUTURES):
+        contract_size = self._get_contract_size(pair)
+        if contract_size and contract_size != 1:
             return num_contracts * contract_size
         else:
             return num_contracts
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1765,7 +1765,7 @@ class Exchange:
             self._async_get_trade_history(pair=pair, since=since,
                                           until=until, from_id=from_id))
 
-    @ retrier
+    @retrier
     def _get_funding_fees_from_exchange(self, pair: str, since: Union[datetime, int]) -> float:
         """
         Returns the sum of all funding fees that were exchanged for a pair within a timeframe
@@ -1871,7 +1871,7 @@ class Exchange:
         """
         return open_date.minute > 0 or open_date.second > 0
 
-    @ retrier
+    @retrier
     def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
         """
         Set's the margin mode on the exchange to cross or isolated for a specific pair
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -649,7 +649,10 @@ class Exchange:
 
         if ('amount' in limits and 'min' in limits['amount']
                 and limits['amount']['min'] is not None):
-            self._contract_size_to_amount(pair, min_stake_amounts.append(limits['amount']['min'] * price))
+            self._contract_size_to_amount(
+                pair,
+                min_stake_amounts.append(limits['amount']['min'] * price)
+            )
 
         if not min_stake_amounts:
             return None
@@ -837,15 +840,15 @@ class Exchange:
 
     def _amount_to_contract_size(self, pair: str, amount: float):
 
-        if ('contractSize' in self.markets[pair]):
-            return amount / self.markets[pair]['contractSize']
+        if ('contractSize' in self._api.markets[pair]):
+            return amount / self._api.markets[pair]['contractSize']
         else:
             return amount
 
     def _contract_size_to_amount(self, pair: str, amount: float):
 
-        if ('contractSize' in self.markets[pair]):
-            return amount * self.markets[pair]['contractSize']
+        if ('contractSize' in self._api.markets[pair]):
+            return amount * self._api.markets[pair]['contractSize']
         else:
             return amount
 
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -875,15 +875,23 @@ class Exchange:
 
     def _amount_to_contracts(self, pair: str, amount: float):
 
+        contract_size = None
         if ('contractSize' in self.markets[pair]):
-            return amount / self.markets[pair]['contractSize']
+            contract_size = self.markets[pair]['contractSize']
+
+        if (contract_size and self.trading_mode == TradingMode.FUTURES):
+            return amount / contract_size
         else:
             return amount
 
     def _contracts_to_amount(self, pair: str, num_contracts: float):
 
+        contract_size = None
         if ('contractSize' in self.markets[pair]):
-            return num_contracts * self.markets[pair]['contractSize']
+            contract_size = self.markets[pair]['contractSize']
+
+        if (contract_size and self.trading_mode == TradingMode.FUTURES):
+            return num_contracts * contract_size
         else:
             return num_contracts
 
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -397,6 +397,28 @@ class Exchange:
                         order[prop] = order[prop] * contract_size
         return order
 
+    def _amount_to_contracts(self, pair: str, amount: float):
+
+        contract_size = None
+        if ('contractSize' in self.markets[pair]):
+            contract_size = self.markets[pair]['contractSize']
+
+        if (contract_size and self.trading_mode == TradingMode.FUTURES):
+            return amount / contract_size
+        else:
+            return amount
+
+    def _contracts_to_amount(self, pair: str, num_contracts: float):
+
+        contract_size = None
+        if ('contractSize' in self.markets[pair]):
+            contract_size = self.markets[pair]['contractSize']
+
+        if (contract_size and self.trading_mode == TradingMode.FUTURES):
+            return num_contracts * contract_size
+        else:
+            return num_contracts
+
     def set_sandbox(self, api: ccxt.Exchange, exchange_config: dict, name: str) -> None:
         if exchange_config.get('sandbox'):
             if api.urls.get('test'):
@@ -872,28 +894,6 @@ class Exchange:
             param = self._ft_has.get('time_in_force_parameter', '')
             params.update({param: time_in_force})
         return params
-
-    def _amount_to_contracts(self, pair: str, amount: float):
-
-        contract_size = None
-        if ('contractSize' in self.markets[pair]):
-            contract_size = self.markets[pair]['contractSize']
-
-        if (contract_size and self.trading_mode == TradingMode.FUTURES):
-            return amount / contract_size
-        else:
-            return amount
-
-    def _contracts_to_amount(self, pair: str, num_contracts: float):
-
-        contract_size = None
-        if ('contractSize' in self.markets[pair]):
-            contract_size = self.markets[pair]['contractSize']
-
-        if (contract_size and self.trading_mode == TradingMode.FUTURES):
-            return num_contracts * contract_size
-        else:
-            return num_contracts
 
     def create_order(self, pair: str, ordertype: str, side: str, amount: float,
                      rate: float, leverage: float = 1.0, time_in_force: str = 'gtc') -> Dict:
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -370,6 +370,30 @@ class Exchange:
         else:
             return DataFrame()
 
+    def _get_contract_size(self, pair: str) -> int:
+        if self.trading_mode == TradingMode.FUTURES:
+            return self.markets[pair]['contract_size']
+        else:
+            return 1
+
+    def _trades_contracts_to_amount(self, trades: List) -> List:
+        if len(trades) > 0:
+            contract_size = self._get_contract_size(trades[0]['pair'])
+            if contract_size != 1:
+                for trade in trades:
+                    trade['amount'] = trade['amount'] * contract_size
+            return trades
+        else:
+            return trades
+
+    def _order_contracts_to_amount(self, order: Dict) -> Dict:
+        contract_size = self._get_contract_size(order['pair'])
+        if contract_size != 1:
+            for prop in ['amount', 'cost', 'filled', 'remaining']:
+                if prop in order:
+                    order[prop] = order[prop] * contract_size
+        return order
+
     def set_sandbox(self, api: ccxt.Exchange, exchange_config: dict, name: str) -> None:
         if exchange_config.get('sandbox'):
             if api.urls.get('test'):
@@ -872,13 +896,15 @@ class Exchange:
             rate_for_order = self.price_to_precision(pair, rate) if needs_price else None
 
             self._lev_prep(pair, leverage)
-            order = self._api.create_order(
-                pair,
-                ordertype,
-                side,
-                amount,
-                rate_for_order,
-                params
+            order = self._order_contracts_to_amount(
+                self._api.create_order(
+                    pair,
+                    ordertype,
+                    side,
+                    amount,
+                    rate_for_order,
+                    params
+                )
             )
             self._log_exchange_response('create_order', order)
             return order
@@ -927,7 +953,9 @@ class Exchange:
         if self._config['dry_run']:
             return self.fetch_dry_run_order(order_id)
         try:
-            order = self._api.fetch_order(order_id, pair)
+            order = self._order_contracts_to_amount(
+                self._api.fetch_order(order_id, pair)
+            )
             self._log_exchange_response('fetch_order', order)
             return order
         except ccxt.OrderNotFound as e:
@@ -981,7 +1009,9 @@ class Exchange:
                 return {}
 
         try:
-            order = self._api.cancel_order(order_id, pair)
+            order = self._order_contracts_to_amount(
+                self._api.cancel_order(order_id, pair)
+            )
             self._log_exchange_response('cancel_order', order)
             return order
         except ccxt.InvalidOrder as e:
@@ -1245,9 +1275,13 @@ class Exchange:
             # since needs to be int in milliseconds
             _params = params if params else {}
             my_trades = self._api.fetch_my_trades(
-                pair, int((since.replace(tzinfo=timezone.utc).timestamp() - 5) * 1000),
-                params=_params)
-            matched_trades = [trade for trade in my_trades if trade['order'] == order_id]
+                pair,
+                int((since.replace(tzinfo=timezone.utc).timestamp() - 5) * 1000),
+                params=_params
+            )
+            matched_trades = self._trades_contracts_to_amount(
+                trades=[trade for trade in my_trades if trade['order'] == order_id]
+            )
 
             self._log_exchange_response('get_trades_for_order', matched_trades)
             return matched_trades
@@ -1584,14 +1618,18 @@ class Exchange:
             # fetch trades asynchronously
             if params:
                 logger.debug("Fetching trades for pair %s, params: %s ", pair, params)
-                trades = await self._api_async.fetch_trades(pair, params=params, limit=1000)
+                trades = self._trades_contracts_to_amount(
+                    trades=await self._api_async.fetch_trades(pair, params=params, limit=1000),
+                )
             else:
                 logger.debug(
                     "Fetching trades for pair %s, since %s %s...",
                     pair,  since,
                     '(' + arrow.get(since // 1000).isoformat() + ') ' if since is not None else ''
                 )
-                trades = await self._api_async.fetch_trades(pair, since=since, limit=1000)
+                trades = self._trades_contracts_to_amount(
+                    trades=await self._api_async.fetch_trades(pair, since=since, limit=1000),
+                )
             return trades_dict_to_list(trades)
         except ccxt.NotSupported as e:
             raise OperationalException(
@@ -1727,7 +1765,7 @@ class Exchange:
             self._async_get_trade_history(pair=pair, since=since,
                                           until=until, from_id=from_id))
 
-    @retrier
+    @ retrier
     def _get_funding_fees_from_exchange(self, pair: str, since: Union[datetime, int]) -> float:
         """
         Returns the sum of all funding fees that were exchanged for a pair within a timeframe
@@ -1833,7 +1871,7 @@ class Exchange:
         """
         return open_date.minute > 0 or open_date.second > 0
 
-    @retrier
+    @ retrier
     def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
         """
         Set's the margin mode on the exchange to cross or isolated for a specific pair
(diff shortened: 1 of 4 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -622,20 +622,11 @@ def get_min_pair_stake_amount(self, pair: str, price: float, stoploss: float,
         # The value returned should satisfy both limits: for amount (base currency) and
         # for cost (quote, stake currency), so max() is used here.
         # See also #2575 at github.
-        return self._get_stake_amount_considering_leverage(
-            max(min_stake_amounts) * amount_reserve_percent,
-            leverage or 1.0
+        return self._contract_size_to_amount(
+            pair,
+            max(min_stake_amounts) * amount_reserve_percent
```
samgermain: When I look at the field amount under `fetchMarkets`, all I see is the `base` amount so I don't know why we get the max of `min_stake_amounts`

If we're using the `quote` amount, then we would need to divide by the leverage

samgermain:
These might be all the changes that are needed to get this aspect working, but I'm not sure, @xmatthias if you could just look at the changes made so far, and think about what else might need to change in order to get this working properly that'd be great

freqtrade/exchange/exchange.py
xmatthias: we're checking for both cost and amount, therefore the "real" minimum is the higher value of the 2.

for example: 
minimum amount for XRP/USDT is 10 XRP
but minimum cost for XRP is 10 USDT

Now depending on the price for 10 XRP - the higher of the 2 defines the real limit (the amount will be added as "10 XRP * current price").
This means - if current price for XRP is 2 USDT - then the minimum we're allowed to trade is for 20 USDT, therefore we need to use the maximum of the two.
If the current price for XRP falls below 1 USDT, then the 10$ minimum (from the cost field) comes into play - never allowing us to drop below that value.

As you'll notice, depend heavily on the pair (and exchange) - a recent example for this can be found here, where the price spiked a lot, causing the minimum tradable amount to raise to 40+$.
https://github.com/freqtrade/freqtrade/issues/5856#issuecomment-962621467

You seem to only be looking at futures, where you talk in contract sizes
but the above logic will apply (with leverage i assume) to margin, so i'm not sure if ignoring leverage is the correct approach, as in my understanding, this will now break the "leveraged margin" case.

xmatthias:
> These might be all the changes that are needed to get this aspect working, but I'm not sure, @xmatthias if you could just look at the changes made so far, and think about what else might need to change in order to get this working properly that'd be great

in my current understanding, `contractSize` is just a 3rd way to define the minimum lotsize?
if that's the case, then i think it should instead be added to "min_stake_amounts" - or replace the existing values in case of futures.


I think this might get one aspect working (futures) - while breaking another aspect (margin) ... as we're now no longer factoring leverage in - which i think is wrong in case of futures.

freqtrade/exchange/exchange.py
```diff
@@ -818,6 +827,7 @@ def create_order(self, pair: str, ordertype: str, side: str, amount: float,
             rate_for_order = self.price_to_precision(pair, rate) if needs_price else None
 
             self._lev_prep(pair, leverage)
+            amount = self._amount_to_contract_size(pair, amount)
```
xmatthias: I think this should be combined into `amount_to_precision` (roughly 5 lines above) - which does exactly the same thing.

Assuming `contractSize` works similarily to `markets[pair]['precision']['amount']` (i don't see why it should not) you'll need to account for different precision modes ... which can vary between exchanges.

freqtrade/exchange/exchange.py
samgermain: I think these might be all the changes that are needed

freqtrade/exchange/exchange.py
```diff
@@ -815,6 +821,20 @@ def _get_params(self, ordertype: str, leverage: float, time_in_force: str = 'gtc
             params.update({param: time_in_force})
         return params
 
+    def _amount_to_contract_size(self, pair: str, amount: float):
+
+        if ('contractSize' in self.markets[pair]):
+            return amount / self.markets[pair]['contractSize']
+        else:
```
xmatthias: This is wrong, and will break spot functionality on binance (and probably most other exchanges, even though i didn't bother to verify) with "unsupported operands - division by float and NoneType", as binance markets do contain ` 'contractSize': None`.

That's not wrong per se (from an information perspective) (and having the key there but empty is better than having a missing key in my opinion) - and we'll always have to have proper handling for this anyway.

It points to a 2nd problem though - which is that this is not tested.

The 3rd problem is described more in detail in the other comment - However - due to the way python handles floats, you can end up with `3.3000000000000003` - which is WAY over precision the exchange supports.

freqtrade/exchange/exchange.py
```diff
@@ -669,7 +675,7 @@ def _get_stake_amount_considering_leverage(self, stake_amount: float, leverage:
     def create_dry_run_order(self, pair: str, ordertype: str, side: str, amount: float,
                              rate: float, leverage: float, params: Dict = {}) -> Dict[str, Any]:
         order_id = f'dry_run_{side}_{datetime.now().timestamp()}'
-        _amount = self.amount_to_precision(pair, amount)
+        _amount = self._contract_size_to_amount(pair, self.amount_to_precision(pair, amount))
```
xmatthias: In my opinion / understanding, this call is wrong.

`amount_to_precision` should be the last to be called. 
We've had multiple instances in the past where we were ordering `4.4555667` of "coin whatever" - and the exchange told us "precision is 4" - so you need to round it to `4.4555` (or `4.4556`) for this order to be placed.

This is not the case with every exchange - but to me, it means that precision should have heigher weight than "contract size" - and should therefore be called last.

Obviously the "precision is 4" can come in multiple different forms - ccxt knows 3 i think - one is "decimal places" - the other is "0.0001" (and i forgot the 3rd as it's only used by 1 or 2 exchanges) - both of which will mean the same thing - but will have a different "precision_mode" attached to the exchange class.
Details for this are in amount_to_precision - but i think the same rules do apply to futures / margin trading as well.

xmatthias (changes requested):
The bug (see comment) is what's causing a "request changes" here ... 

i do however also think this is wrong from a conceptual point of view (not only because of the bug)...

but that may also be due to my the lack of understanding ...

assuming i wanna order a 100 contracts of XRP (ignore the price for now) - and it's contractSize is 15 .. so i'd be ordering `100/15=6.666` contracts - wouldn't that mean i'll not be getting 100 contracts, but 6.666 instead?

The binance case makes this worse - assuming i order 1 BTC - it'd have to be `1 / 0.0001` - which is 10.000 contracts ... ??

freqtrade/exchange/exchange.py
samgermain: The idea with this was that it's a dry run order, so I'm just removing any references to contract sizes and using the actual amounts, `_amount_to_contract_size` is already in `price_to_precision`

freqtrade/exchange/exchange.py
samgermain: > assuming i wanna order a 100 contracts of XRP (ignore the price for now) - and it's contractSize is 15 .. so i'd be ordering `100/15=6.666` contracts - wouldn't that mean i'll not be getting 100 contracts, but 6.666 instead?

So you wouldn't want to order 100 contracts, you would want to order 0.0005 BTC or something, but you can't place an order in btc amount, you have to place it in contract size amount, so we have to find out how many contracts would be worth 0.0005 BTC. If the contract size is 0.0001, then you would place an order for 5 contracts which is amount(0.0005)/contract_size(0.0001)

> The binance case makes this worse - assuming i order 1 BTC - it'd have to be `1 / 0.0001` - which is 10.000 contracts ... ??

Binance is the normal case, where you place an order for 0.0001 if you want 0.0001 BTC

xmatthias: yeah, binance was a typo - i was actually using the numbers in your original post for this :laughing:

freqtrade/exchange/exchange.py
xmatthias: just to understand this right (from a calculation point of view)
your idea was to: 
* convert the amount from 100 to 6.66
* place that order (in live)
* convert 6.66 back to 100 for dry-run?
* leave it at 6.66 for live

I'm honestly not sure if this will all play out (not this PR, but looking at it from a bigger picture), and i think we'll need a proper concept for this (which currently, i don't really see) before going further with this PR.

Please consider that
* `trade.amount` is used in basically all profit calculations
* `trade.amount` is updated with `filled`, `amount` from the order we're getting back (live, obviously)

So if i place an order with `6.66` - i expect to get 6.66 back in the order's filled column.
Updating a dry-order "back" to an amount of 100 is therefore wrong in my understanding - and the trade object should hold 6.66 (we're holding 6.66 contracts).
Dry-run should return (roughly) the same values - which would not be the case with this.

---

The bigger problem than dry-run is the following though:

While your amount is now 6.66 (assigned by `amount_to_precision`) - all calculations now are wrong (i was ordering 100 XXX at 2$ each, so i'd expect my entry cost to be 200$).
Instead, my trade object now has an amount of 6.66 - which means my (calculated) cost is now 6.66 * 2 - which is around 13$ ... quite a long spot from where we wanted to be.

samgermain: We only need to change the value that we send to the exchange, we will still analyze it as 100 of currency X, but we need to tell the exchange that we want 6.66 contracts (If a contract is worth 15.01~ of currency X). The exchange will also return amounts in how many contracts, we should convert these into the size of the underlying asset before we deal with them. The prices are all compared to the underlying asset, if the price is 62000, then it's 62000 for one of currency X, regardless of what the contract size is

xmatthias: i'm not sure if the "conversion back" is the right approach - i fear that this will lead to quite some bugs, as we'll need to have this conversion in place for wherever we use an order 
This also includes "userland", as the order object is directly passed to the timeout handlers.

an alternative might be to update the order we receive back to the "right amount" (basically patching the ccxt exchange response), which is probably simpler, but potentially also slightly riskier (we need to be 100% sure which fields this applies and which doesn't).

---

I understand this is limited to some exchanges (so far, i could only find okex and gate.io which define this) .. do you know which exchanges are _not_ impacted by this? 
if we're left with enough alternatives, we should consider ignoring this for the moment (simply to be able to move forward, and not have the "shorting" feature delayed by edgecase exchanges, and excluding support for these exchanges - as this pr is a far way from "all that's needed" - as it's only handling the "place order" side at the moment - but not the "way back".

samgermain: It's not needed for linear contracts on Binance, so we don't need it immediately, but for those other exchanges, it will be pretty important, so if anyone wants to use one of those exchanges, they should take a look at this PR

xmatthias: so to summarize, without this, we're ONLY left with binance? 
then we'll need to find a solution to this i think. i was hoping that this would only impact 2-3 exchanges, not "most" ...

samgermain: It's not needed for FTX either, but I won't be implementing cross margin anytime soon, which is needed for FTX

I don't know for kraken, because futures isn't implemented for it on ccxt yet

xmatthias: I'm not happy with a "only binance" approach - as we can lock ourselfs in a corner with this (similar to what happened to ccxt, looking at it in hindsight) - but i see this PR as high risk change, as it can mess up all calculations in very subtle and odd ways, which will need both VERY good test, as well as very good and detailed live (and dry) tests.

Then let's focus on margin trading only for the moment.
I don't see what's missing for margin trading (which will implicitly enable both leverage and short trading) - i'm sure it's not 100% complete, but it's not much either, while for futures, the list of "that's still missing" continues to become longer and longer, instead of shorter and shorter.
As such, i don't see a reason to delay finishing up margin and making that available - once that is stable, we can continue to try and implement futures.

I hinted on this earlier as well - as i think we're trying to do too much, and with that unnecessarily delaying other parts - and i regret (and apologize) for not being more decisive on this, as we could have margin available already (probably since months) if we kept our focus on that part first (which we will do now).

This does not mean futures is off the table - but will not be worked on until margin trading implementation is completed. This will allow us to focus on that part - instead of doing "some of this, some of that", and therefor provide value to the users faster.

xmatthias: To return to this discussion ... 
If we assume freqtrade will only deal with amounts - not with "contract Sizes" at any other place than when we're sending the order to the exchange - then we'll also have to convert the exchange responses back (for all values that we use as amounts - like `'filled'`, `'amount'`, (and potentially others ...?) - in both `fetch_order`, `create_order`, as well as (probably?) in `fetch_my_trades`.

The alternative is to implement these changes to where we use it (which is in models.py - as well as "all throughout the bot" - which i fear will leave us pretty vulnerable to bugs

samgermain:
It looks like these methods need to be updated, plus maybe some more

```
create_order
stoploss_adjust
stoploss
fetch_order
fetch_order_or_stoploss_order
check_order_canceled_empty
cancel_order
is_cancel_order_result_suitable
cancel_order_with_result
cancel_stoploss_order_with_result
fetch_l2_order_book
get_trades_for_order
get_historic_ohlcv
get_historic_ohlcv_as_df
_async_get_historic_ohlcv
_async_get_candle_history
_async_fetch_trades
_async_get_trade_history
get_historic_trades
```

freqtrade/exchange/exchange.py
samgermain: Do you still think is wrong?

tests/exchange/test_exchange.py
```diff
@@ -1168,6 +1200,8 @@ def test_create_order(default_conf, mocker, side, ordertype, rate, marketprice,
 
     assert exchange._set_leverage.call_count == 1
     assert exchange.set_margin_mode.call_count == 1
+    # assert api_mock.create_order.call_args[0][3] == 100
```
samgermain: I couldn't get this working seeing as `amount_to_precision` is mocked. Do you know how to get this working?

samgermain: This would be the only thing that would be left to do on this PR

freqtrade/exchange/exchange.py
xmatthias: The code didn't change - so neither did my opinion on this.

It MIGHT be correct for futures  (it's not and will fail in case of None values you get from binance) - but as you're calling this without conditions, it'll simply break freqtrade for spot markets depending on what ccxt returns.

Tests might fool us into thinking it works (this is not tested against non-mocked methods) - but it'll only work because we're not having `'contractSize': None` in our test markets structure.

**edit** there is, actually - but on a pair that's not used for any exchange testing afaik

xmatthias: the correct way to test for this is in `_get_contract_size` (also testing for None)- which is however not used here - which makes this method wrong (and the overall picture a little confusing)

freqtrade/exchange/exchange.py
```diff
@@ -903,7 +958,9 @@ def fetch_order(self, order_id: str, pair: str) -> Dict:
         if self._config['dry_run']:
             return self.fetch_dry_run_order(order_id)
         try:
-            order = self._api.fetch_order(order_id, pair)
+            order = self._order_contracts_to_amount(
+                self._api.fetch_order(order_id, pair)
+            )
             self._log_exchange_response('fetch_order', order)
```
xmatthias: While the end result will be identical - can we change this to the following (throughout) the exchange class:

```suggestion
            order = self._api.fetch_order(order_id, pair)
            self._log_exchange_response('fetch_order', order)
            order = self._order_contracts_to_amount(order)
```

This will make sure that
a) we're logging the original response - not something we've attempted to modify
b) we can easily add debug-points to the response directly, allowing for easier debugging.

tests/exchange/test_exchange.py
```diff
@@ -3601,3 +3701,165 @@ def test__calculate_funding_fees_datetime_called(
...
+    ('ETH/USDT:USDT', 10, 'futures')
+])
+def test__get_contract_size(mocker, default_conf, markets, pair, expected_size, trading_mode):
+    api_mock = MagicMock()
+    default_conf['trading_mode'] = trading_mode
+    default_conf['collateral'] = 'isolated'
+    mocker.patch('freqtrade.exchange.Exchange.markets', markets)
+    exchange = get_patched_exchange(mocker, default_conf, api_mock)
+    size = exchange._get_contract_size(pair)
+    assert expected_size == size
```
xmatthias: the thing i don't like about this test is that it takes assumptions about the setup in conftest - and basically only tests that - and only partially the logic it should be testing.

while we do this in other areas too, this is explicitly testing for the values contained there, so i don't think a better approach here would be: 
instead of pair - provide contractSize.
Set that for a (static) pair as part of the test in a dummy market structure - and then verify if the result is expected - basically testing if "nothing" or spot (or 1)  returns 1 - and the other cases return the provided value.

That'll remove the dependency on the "markets" - which can fall on our feet should we ever decide to use these pairs for some other tests (it did create problems in the past already - so i think new tests should not create new dependencies in this regards).

freqtrade/exchange/exchange.py
```diff
@@ -1244,7 +1307,7 @@ def get_fee(self, symbol: str, type: str = '', side: str = '', amount: float = 1
             # validate that markets are loaded before trying to get fee
             if self._api.markets is None or len(self._api.markets) == 0:
                 self._api.load_markets()
-
+            # TODO-lev: Convert this amount to contract size?
```
xmatthias: Don't bother (for two reasons, actually).
The way we are using `calculate_fee` is to only take the rate (relative value - like 0.01%).

While amount is a necessary input for ccxt - it has no impact on the relative result - (0.01% will remain identical, for both 1 or for 100_000).
Freqtrade is also never passing in amount - so it'll always use the default of 1 anyway.

tests/exchange/test_exchange.py
xmatthias: That test mocks `amount_to_precision` - which handles the conversion, so it's never happening in the first place.

i'm not 100% sure if we do need to test for this explicitly - having proper tests for `amount_to_precision` in place - as well as testing that `create_order` calls `amount_to_precisoin` might suffice.

freqtrade/exchange/exchange.py
```diff
@@ -959,6 +1014,7 @@ def cancel_order(self, order_id: str, pair: str) -> Dict:
         try:
             order = self._api.cancel_order(order_id, pair)
             self._log_exchange_response('cancel_order', order)
+            order = self._order_contracts_to_amount(order)
```
xmatthias, line 1022: we will need to be careful with `cancel_order` calls.

I've rolled back the changes to `test_cancel_order` - that test MUST pass without modification.

Order cancellation is not guaranteed to contain anything (best example is [FTX](https://docs.ftx.com/#cancel-order))- so we could get a response like `{'result': 'success'}` (not containing any symbol or amount) - and that response cannot have freqtrade crash - which it would now.

freqtrade/exchange/exchange.py
```diff
@@ -641,11 +670,21 @@ def get_min_pair_stake_amount(self, pair: str, price: float, stoploss: float,
...
 
         if ('amount' in limits and 'min' in limits['amount']
                 and limits['amount']['min'] is not None):
-            min_stake_amounts.append(limits['amount']['min'] * price)
+            min_stake_amounts.append(
+                self._contracts_to_amount(
+                    pair,
+                    limits['amount']['min'] * price
+                )
+            )
```
xmatthias, line 706: are we sure this (and the above) parts are needed?

can you point me to a case (exchange/pair combination) where we're getting cost/amount limits in contract size?

samgermain: So if you're familiar with the CCXT cli, just run `node examples/js/cli okex market BTC/USDT:USDT` an you can see that the amount.min is `1` which is 1 BTC contract and not 1 BTC

```
% node examples/js/cli okex market BTC/USDT:USDT
2021-12-28T13:45:36.880Z
Node.js: v14.17.0
CCXT v1.64.80
okex.market (BTC/USDT:USDT)
0 ms
{         limits: { leverage: { min: undefined, max: 125 },
                      amount: { min: 1, max: undefined },
                       price: { min: 0.1, max: undefined },
                        cost: { min: 0.1, max: undefined }  },
       precision: { amount: 1, price: 0.1 },
       tierBased:    undefined,
      percentage:    undefined,
           taker:    0.0005,
           maker:    0.0002,
              id:   "BTC-USDT-SWAP",
          symbol:   "BTC/USDT:USDT",
            base:   "BTC",
           quote:   "USDT",
          baseId:   "BTC",
         quoteId:   "USDT",
        settleId:   "USDT",
          settle:   "USDT",
            info: {     alias: "",
                      baseCcy: "",
                     category: "1",
                       ctMult: "1",
                       ctType: "linear",
                        ctVal: "0.01",
                     ctValCcy: "BTC",
                      expTime: "",
                       instId: "BTC-USDT-SWAP",
                     instType: "SWAP",
                        lever: "125",
                     listTime: "1636620075000",
                        lotSz: "1",
                        minSz: "1",
                      optType: "",
                     quoteCcy: "",
                    settleCcy: "USDT",
                        state: "live",
                          stk: "",
                       tickSz: "0.1",
                          uly: "BTC-USDT"       },
            type:   "swap",
            spot:    false,
         futures:    false,
            swap:    true,
        contract:    true,
         option:    false,
         linear:    true,
         inverse:    false,
         active:    true,
         contractSize:   "0.01",
          expiry:    undefined,
          expiryDatetime:    undefined
  }
```

tests/exchange/test_exchange.py
samgermain: > i'm not 100% sure if we do need to test for this explicitly 

That's enough for me not to do the work, if you ever change your mind I can look at it again

tests/exchange/test_exchange.py
samgermain: I don't understand what you meant in your comment, but the test no longer assumes values from conftest

freqtrade/exchange/exchange.py
samgermain: I think I fixed the problem

samgermain:
@xmatthias I can't get rid of that "changes requested" thing, and when I click on it it says it can't find the files

xmatthias:
It's part of the prior review ... 
it'll go away as soon as I've found the time to review this again (i expect no more changes - but we'll see).

freqtrade/exchange/exchange.py
```diff
@@ -1822,6 +1889,46 @@ def set_margin_mode(self, pair: str, collateral: Collateral, params: dict = {}):
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 
+    @retrier
+    def _get_mark_price_history(self, pair: str, since: int) -> Dict:
```
xmatthias: I think something must've gone wrong with the rebase you (probably) did - probably while resolving some merge conflicts.

I've removed this code (as well as `get_funding_rate_history()`) in #6043 (as well as the corresponding test) - as there's a much simpler (and more efficient) way to accomplish this.

freqtrade/exchange/exchange.py
```diff
@@ -831,6 +869,28 @@ def _get_params(self, ordertype: str, leverage: float, time_in_force: str = 'gtc
...
+
+        contract_size = None
+        if ('contractSize' in self.markets[pair]):
+            contract_size = self.markets[pair]['contractSize']
+
+        if (contract_size and self.trading_mode == TradingMode.FUTURES):
+            return num_contracts * contract_size
+        else:
+            return num_contracts
+
```
xmatthias: Please move these lines next to `_get_contract_size` (either this part up, or the other ones down).

That way, all methods that belong together are "somehow" grouped within the file - so we don't have to look all over the place for things that belong together.

xmatthias (changes requested):
I've merged develop to feat/short now, so the test in CI should pass.

Please remove the unintended rebase problems (see other comment) - then i think this will be good :+1:

tests/exchange/test_exchange.py
```diff
@@ -2700,12 +2795,17 @@ def test_name(default_conf, mocker, exchange_name):
     assert exchange.id == exchange_name
 
 
+@pytest.mark.parametrize("trading_mode,amount", [
+    ('spot', 0.2340606),
+    ('futures', 23.40606),
```
samgermain: On the most recent commit I changed `('futures', 2.340606),` to `('futures', 23.40606),`

I don't understand why it was passing before

freqtrade/exchange/exchange.py
```diff
@@ -1572,6 +1642,7 @@ async def _async_fetch_trades(self, pair: str,
                     '(' + arrow.get(since // 1000).isoformat() + ') ' if since is not None else ''
                 )
                 trades = await self._api_async.fetch_trades(pair, since=since, limit=1000)
+            trades = self._trades_contracts_to_amount(trades)
```
xmatthias, line 1635: I'm ... hesitant about this (i'm actually not sure it's relevant to have the conversation done for this method).

This function is only called when using `freqtrade download-data --dl-trades` - which should only be called on kraken (all other supported exchanges provide OHLCV data directly).
Calling freqtrade in futures mode this way is currently not allowed (and i don't see that being enabled anytime soon).

tests/exchange/test_exchange.py
xmatthias: The test was correct - your rebase somehow created another bug (see last commit) - calling the conversion twice.

freqtrade/exchange/exchange.py
samgermain: Taking it out just seems like putting in work to make it more difficult  should this ever be needed. What if another exchange ends up needing it or we add kraken futures, or someone else wants to add kraken futures, because I probably won't. If we don't need it, it will keep the sizes as they are

xmatthias: kraken futures are a long way out i think - as ccxt does not support these either (it's a completely separate API - afaik).

if we don't need it - it's a call that's pointlessly consuming cpu cycles.

Let's leave it for now - I've noted it down as something to "look at again" before merging to develop.

xmatthias (approved):

xmatthias:
I've simplified one piece a little bit as i just saw it on my last "flyover" ... will be merged once CI passes

freqtrade/exchange/exchange.py
samgermain: > which should only be called on kraken (all other supported exchanges provide OHLCV data directly).

Could this method of downloading OHLCV data be used for huobi?

xmatthias: @samgermain no, huobi does not provide any historic data beyond 1500 (or 2000?) datapoints.

Their endpoint is called `Most recent trades` [huobi docs](https://huobiapi.github.io/docs/spot/v1/en/#get-the-most-recent-trades), and works similarly to the OHLCV endpoint.
Neither of the 2 even accept a "since/from/to" argument, only limit and symbol.
As such, you can only get the last `limit` datapoints - but never the data from "3 months ago".

Huobi itself will (probably) work for trading - but downloading historic data (and hence backtesting) will be impossible due to lack of data.
as such, it's an exchange we will not be able to recommend anytime soon.

to support huobi - huobi will need to change their API and/or add end points to allow downloading historic data. (i'm pretty sure that's also a problem in some areas for ccxt itself - and may raise quite some questions/issues on your side - unless almost noone is properly using that exchange).

If you have a contact there/are interested in having huobi supported by freqtrade, best get in touch with huobi / huobi support on adding this part.

PR #5890, opened 2021-11-14
Todo-lev RPC tests and model

aezomz:
Done flake8 and unit test, 
`FAILED tests/test_freqtradebot.py::test_cancel_all_open_orders[True-2-1] - AssertionError: assert 1 == 2`
Is failing, but i think Sam will be fixing it after the feature is included. 

## Summary
Added test to complete TODO - lev on RPC.

## Quick changelog
- Added is_short to data model
- Added short_trades to _rpc_trade_statistics

## What's new?
Added is_short to data model
Added short_trades to _rpc_trade_statistics
Maybe we should add a mixture of long and short trades?

I added check for "is_short" to determine "buy" or "sell" to close the position when calling,
`self._freqtrade.exchange.get_rate`

Please comment and feedback :) Sorry for any mistakes, I will get it fixed.


aezomz:
@samgermain , here is the PR for RPC test. I am not sure how I add you as reviewer. But I think Matt can help too. Thanks

freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -95,6 +95,7 @@ class Profit(BaseModel):
     avg_duration: str
     best_pair: str
     best_rate: float
+    short_trades: int
```
xmatthias: For the `TradeSchema`, `is_short` makes perfect sense ... here, i'm not sure what the intention is (how you intend to use it afterwards).

if you think of showing a "amount of short trades" - then i think we should also add a `long_trades` alongside, as `trade_count` is the total of trades (both long and short).
Without that, we'd have to always subtract one from the other, which doesn't really make sense.

On the other hand, i'm not sure if it makes sense in this particular endpoint at all - it's all about profits, not so much about "trade direction" ...
if we start counting this - then why not have a "winning_short_trades" - "losing_short_trades" as well? (It can get pretty detailed - or not - but i think if we don't intend on using it, there's little benefit.

tests/rpc/test_rpc_apiserver.py
```diff
@@ -698,8 +699,8 @@ def test_api_edge_disabled(botclient, mocker, ticker, fee, markets):
     assert rc.json() == {"error": "Error querying /api/v1/edge: Edge is not enabled."}
 
 
-# TODO-lev: @pytest.mark.parametrize('is_short', [True, False])
-def test_api_profit(botclient, mocker, ticker, fee, markets):
+@pytest.mark.parametrize('is_short', [True, False])
+def test_api_profit(botclient, mocker, ticker, fee, markets, is_short):
```
xmatthias: Thinking about this - i'm not 100% sure about this way to test this part of the code - as it always only tests one side (either 100% long trades, or 100% short trades).

potentially / ideally, we should have a way to create "mixed trades" - where some are long, and some are short trades.

@samgermain, what do you think about that? (i think i can add that pretty easily in to `create_mock_trades()` if you feel that would make sense).

tests/rpc/test_rpc_apiserver.py
```diff
@@ -974,7 +984,8 @@ def test_api_whitelist(botclient):
     }
 
 
-def test_api_forcebuy(botclient, mocker, fee):
+@pytest.mark.parametrize('is_short', [True, False])
+def test_api_forcebuy(botclient, mocker, fee, is_short):
```
xmatthias: This test (for short) is testing the wrong thing.

We currently don't have a way to send a "/forceshort" to the bot - so this should not be possible.

I know the test works, and that's only because you cheat the `is_short` value into the mock response - which in reality is a method call, which will never receive the "should be short" notice.
So the test is making it seem like you can force-enter shorts - which you really can't at this moment.

I'd suggest to remove changes to this test again - as we'll need to come up with a proper strategy on how to handle "forced short entries" - which i think we should handle in a separate PR, as it'll not be a 2liner (probably, anyway).

tests/rpc/test_rpc_apiserver.py
```diff
@@ -714,39 +715,50 @@ def test_api_profit(botclient, mocker, ticker, fee, markets):
...
+        'profit_closed_percent_sum': -1.5 if is_short else 0.5 if is_short is None else 1.5,
+        'profit_closed_ratio': -6.739057628404269e-06 if is_short
+        else -5.429078808526421e-06 if is_short is None else 7.391275897987988e-07,
+        'profit_closed_percent': -0.0 if is_short else -0.0 if is_short is None else 0.0,
+        'trade_count': 6,
+        'closed_trade_count': 2,
+        'short_trades': 6 if is_short else 3 if is_short is None else 0,
+        'winning_trades': 0 if is_short else 1 if is_short is None else 2,
+        'losing_trades': 2 if is_short else 1 if is_short is None else 0,
+        }
```
samgermain: Instead of having all these conditionals, I would add these values inside `@pytest.mark.parametrize`, you can find lots of other tests somewhere in freqtrade for an example

samgermain (changes requested):

tests/test_freqtradebot.py
```diff
@@ -4317,7 +4317,7 @@ def test_sync_wallet_dry_run(mocker, default_conf_usdt, ticker_usdt, fee, limit_
 @pytest.mark.usefixtures("init_persistence")
 @pytest.mark.parametrize("is_short,buy_calls,sell_calls", [
     (False, 1, 2),
-    (True, 2, 1),
+    (True, 1, 2),
```
samgermain, line 4320: > Instead of having all these conditionals, I would add these values inside @pytest.mark.parametrize, you can find lots of other tests somewhere in freqtrade for an example

Like here's one for example

tests/rpc/test_rpc_apiserver.py
```diff
@@ -714,39 +720,50 @@ def test_api_profit(botclient, mocker, ticker, fee, markets):
...
+        'profit_closed_percent_sum': -1.5 if is_short else 0.5 if is_short is None else 1.5,
+        'profit_closed_ratio': -6.739057628404269e-06 if is_short
+        else -5.429078808526421e-06 if is_short is None else 7.391275897987988e-07,
+        'profit_closed_percent': -0.0 if is_short else -0.0 if is_short is None else 0.0,
+        'trade_count': 6,
+        'closed_trade_count': 2,
+        'short_trades': 6 if is_short else 3 if is_short is None else 0,
+        'winning_trades': 0 if is_short else 1 if is_short is None else 2,
+        'losing_trades': 2 if is_short else 1 if is_short is None else 0,
+    }
```
samgermain, line 791: Instead of having all these conditionals, I would add these values inside @pytest.mark.parametrize, you can find lots of other tests somewhere in freqtrade for an example

tests/conftest.py
```diff
@@ -248,33 +248,35 @@ def patched_get_exit_signal(pair, timeframe, dataframe, is_short):
     freqtrade.exchange.refresh_latest_ohlcv = lambda p: None
 
 
-def create_mock_trades(fee, is_short: bool = False, use_db: bool = True):
+def create_mock_trades(fee, is_short: Optional[bool] = False, use_db: bool = True):
```
samgermain, line 251: I think this should either be true or false, not Optional, unless there's a good reason for it being optional

tests/conftest.py
```diff
@@ -248,33 +248,35 @@ def patched_get_exit_signal(pair, timeframe, dataframe, is_short):
...
+    :param is_short: Optional bool, None creates a mix of long and short trades.
     """
     def add_trade(trade):
         if use_db:
             Trade.query.session.add(trade)
         else:
             LocalTrade.add_bt_trade(trade)
-
+    is_short1 = is_short if is_short is not None else True
+    is_short2 = is_short if is_short is not None else False
```
samgermain, line 262: Why split them up? My intention was to test a bunch of long trades, or else test a bunch of short trades

freqtrade/rpc/api_server/api_schemas.py
aezomz: Sure, I think it might be too detailed. I will remove the short_trades for now. 

Probably another PR in the future when features are more complete to test. I will create another endpoint for more detailed report.

tests/conftest.py
aezomz: Ya, we should always specify `is_short=False/True`  when passing in the parameters...  just in case it went to use_db.

tests/conftest.py
aezomz: @xmatthias and I felt that we should also test mixed trades (longs and shorts)
I can create another test case for this if we want to test (bunch longs bunch shorts) then mixed trades.

tests/conftest.py
samgermain: Well as it is, I think that it will always default to `False` (not None) which is kind of weird behavior

samgermain: > Ya, we should always specify `is_short=False/True` when passing in the parameters... just in case it went to use_db.

Either we should do that, or `is_short` should be the last parameter

xmatthias: This is optional because an explicit `None` has a meaning, which will result in a mix of buys and sells.

tests/conftest.py
xmatthias: For RPC methods, testing only long or short is a strange way to test things, as it will never test a mix of trades, where some trades are longs, and some are shorts (which is way more realistic).

So to actually check if the calculations are correct, we'll need a "Middle ground" - where we have both long and short tests created in an easy way.
While we could create another method for that - i don't really see a point in it, as python allows a boolean value to be True or False or None.

tests/conftest.py
aezomz, line 251: Do u want me to change this for all ? 
create_mock_trades(fee, is_short) -> create_mock_trades(fee, is_short=is_short) ?


I think we should make optional as the last arg.
def create_mock_trades(fee, is_short: Optional[bool] = False, use_db: bool = True): 
-> def create_mock_trades(fee, use_db: bool = True), is_short: Optional[bool] = False: 

I need to change all the test to -> create_mock_trades(fee, is_short=is_short) 

Otherwise it will break alot of tests

samgermain: Just make sure you're testing all the cases

xmatthias: I think you're miss-reading what `Optional` means - it's not that the argument becomes Optional (must not be provided) - but it can become `None` (which `use_db` cannot).

Both `is_short` and `use_db` are optional parameters, as both have default values set - so you can easily use `create_mock_trades(fee)` without breaking anything.

samgermain: An `Optional` is a datatype and the type of it's value is either the type specified (int, bool, whatever) or it's value is None

aezomz: Thanks guys. Commit coming your way ;)

tests/rpc/test_rpc_apiserver.py
```diff
@@ -700,12 +700,45 @@ def test_api_edge_disabled(botclient, mocker, ticker, fee, markets):
...
 
 
-@pytest.mark.parametrize('is_short', [
-    (True),
-    (False),
-    (None),
-])
-def test_api_profit(botclient, mocker, ticker, fee, markets, is_short):
+@pytest.mark.parametrize(
+        'is_short, best_pair, best_rate, profit_all_coin, profit_all_fiat,'
```
aezomz: let me know if i should remove the space, looks more readable like this :/

xmatthias: mmh, on the risk of making it more verbose than it already is - i'd probably pass it in as dict (which i hope makes it more readable)

so you only got 2 parameters - `is_short,expected` with `{'best_pair: 1234, <...> }` as 2nd parameter, - and within the test, you can use `expected['best_pair']` for the check

While it'll be more verbose - i think this is very prone to break (some formatting tools will realign the lines - so this will become a "let's count the fields" problem :O

aezomz: Ya, if not we can seperate the test... Currently is becoming count the field problem. 
Ok I think i will prefer a dict anyday :D

xmatthias (approved):
Nice :+1:
Thanks a lot for your contribution and help! :100:

samgermain:
@aezomz It looks like this test is failing

```
 tests/rpc/test_rpc_apiserver.py::test_api_logs - assert 0 > 1
 ```

xmatthias:
> @aezomz It looks like this test is failing
> 
> ```
>  tests/rpc/test_rpc_apiserver.py::test_api_logs - assert 0 > 1
> ```

that test was not touched - and is not relevant, running ci again will fix that (it's a fluke that happens sometimes, but one i am unable to reproduce reliably enough to try a fix).

aezomz:
Ok let me know, I can re-write that test if it helps.

xmatthias:
> Ok let me know, I can re-write that test if it helps.

it's been failing for months (if not years) now ... at random places (usually in ci) - so i'm pretty much convinced it's a timing issue (it does not reliably reproduce locally).

if you think you can improve it - feel free to give it a try - but not in this PR (it's not related to shorts, actually) - and pointed against develop ... 
Also, don't spend too much time on it - solving issues that don't reproduce can be a time-killer with doubtful outcome ...

PR #5919, opened 2021-11-18
indentation problem :)

mablue:


xmatthias (approved):
not sure i'd classify this as a "problem" considering it's styling of a json file in the docs  😆  

... but yeah, better to have it aligned 👍

PR #5924, opened 2021-11-19
call leverage methods

xmatthias:
## Summary
Implement leverage strategy callback to be used.

Also partially enables backtesting - but it's still far off as some other calculations are still wrong.

## Quick changelog

- implement leverage calls
- add some "niceities" to backtesting which were needed to actually validate :laughing:

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -576,7 +576,6 @@ class FreqtradeBot(LoggingMixin):
         stake_amount: float,
         price: Optional[float] = None,
         forcebuy: bool = False,
-        leverage: float = 1.0,
         is_short: bool = False,
         enter_tag: Optional[str] = None
     ) -> bool:
@@ -590,6 +589,7 @@ class FreqtradeBot(LoggingMixin):
         time_in_force = self.strategy.order_time_in_force['buy']
 
         [side, name] = ['sell', 'Short'] if is_short else ['buy', 'Long']
+        trade_side = 'short' if is_short else 'long'
 
         if price:
             enter_limit_requested = price
@@ -610,7 +610,8 @@ class FreqtradeBot(LoggingMixin):
             pair,
             enter_limit_requested,
             self.strategy.stoploss,
-            leverage=leverage
+            # TODO-lev: This is problematic... we need stake-amount to determine max_leverage
+            # leverage=leverage
         )
 
         if not self.edge:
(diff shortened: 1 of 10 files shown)


freqtrade/freqtradebot.py
```diff
@@ -610,7 +610,8 @@ def execute_entry(
             pair,
             enter_limit_requested,
             self.strategy.stoploss,
-            leverage=leverage
+            # TODO-lev: This is problematic... we need stake-amount to determine max_leverage
+            # leverage=leverage
```
xmatthias: @samgermain maybe you have an idea for this:
we seem to need `leverage` to get the minimum stake amount
but we also need the final stake amount to determine the max-leverage (thanks to binance brackets ...) 

For now, I've removed it from here - but that's not really correct, as it'll eliminate leverage from this calculation.

samgermain: You shouldn't need the final stake amount to determine what the minimum stake amount is, if you're looking at min stake, you know that it's that actual minimum, (5 or 10 of quote) so then you would be able to use the max leverage on that coin (50, 75, 100...), but you shouldn't divide it by the max-leverage, you should divide it by the leverage you're actually using

xmatthias: please look at this not for these 2 lines, but from a greater picture of this method (and it's calling sequence).

the call to `get_max_leverage()` will require stake-amount (at least for binance), as well as getting the minimum stake amount, just as `get_min_stake_amount()` requires leverage to determine the stake.
It's a classic chicken-egg problem (at least at the moment) - and i'm not quite clear which value we should guess (or omit).

Quite simply - we cannot determine the maximum stake amount before we know the used leverage - but we also cannot determine the maximum leverage before we know the stake amount, and therefore cannot "ask" the strategy for the leverage that's to be used.

samgermain: If it's the min stake amount, then you would take the first bracket, you know you won't be in the second bracket
Your maximum leverage will only decrease if you raise your stake amount to a point which is well beyond the minimum stake amount, so if your using the second bracket, you already know that your past the minimum stake amount

If you're using the min stake amount then your max leverage is the actual max, and the leverage brackets don't matter

samgermain: > we cannot determine the maximum stake amount before we know the used leverage 

Did you mean maximum stake amount?

samgermain: maybe we should think in terms of the currency bought? That doesn't change

samgermain: if nothing works, ditch the min stake amount, I think it's 5 on futures, so with a max leverage of 5 thats a min stake amount of 1, and how much lower do you need to go than that

If your trading a large amount though, then the max leverage can go as low as 2, which would matter

xmatthias: > maybe we should think in terms of the currency bought? That doesn't change

that's not true.
Again, please look at the whole method (and think about what it does in what sequence).
we have 2 strategy callbacks involved - `custom_stake_amount` and `leverage`.
Custom stake in particular, can return ANYTHING between min and max stake - so it could return 20$ - or it could return 200.000$ - (which i assume would land me into another bracket).

we can only determine max-leverage based on this - so to me, stake-amount must be first (which however means that we cannot take leverage into consideration for min-stake-amount).

samgermain: > (which however means that we cannot take leverage into consideration for min-stake-amount).

That sounds fine

xmatthias: Great :+1:

PR #5926, opened 2021-11-20
Trade.has_no_leverage makes more sense

samgermain:


freqtrade/persistence/models.py
```diff
@@ -283,7 +283,7 @@ class LocalTrade():
     @property
     def has_no_leverage(self) -> bool:
         """Returns true if this is a non-leverage, non-short trade"""
-        return ((self.leverage or self.leverage is None) == 1.0 and not self.is_short)
+        return ((self.leverage == 1.0 or self.leverage is None) and not self.is_short)
```
xmatthias, line 286: am i mistaking or is this actually saying "did not borrow", not "has no leverage"?

A 1.0 leveraged short trade would also have "no leverage" - but would return true (which is correct considering how it's used).

samgermain: It means did not borrow, some resources refer to a 1x short trade as leverage I believe though

PR #5936, opened 2021-11-21
Use market data to get base and quote currencies in @informative() decorator

rokups:
This does make exchange non-optional when using decorators though.. Should we do anything else about it?

### DIFF
--- a/freqtrade/strategy/informative_decorator.py
+++ b/freqtrade/strategy/informative_decorator.py
@@ -80,12 +80,11 @@ def _create_and_merge_informative_pair(strategy, dataframe: DataFrame, metadata:
         # Not specifying an asset will define informative dataframe for current pair.
         asset = metadata['pair']
 
-    if '/' in asset:
-        base, quote = asset.split('/')
-    else:
-        # When futures are supported this may need reevaluation.
-        # base, quote = asset, ''
-        raise OperationalException('Not implemented.')
+    market = strategy.dp.market(asset)
+    if market is None:
+        raise OperationalException(f'Market {asset} is not available.')
+    base = market['base']
+    quote = market['quote']
 
     # Default format. This optimizes for the common case: informative pairs using same stake
     # currency. When quote currency matches stake currency, column name will omit base currency.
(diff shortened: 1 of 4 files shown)


freqtrade/strategy/informative_decorator.py
```diff
@@ -80,12 +80,11 @@ def _create_and_merge_informative_pair(strategy, dataframe: DataFrame, metadata:
...
-        base, quote = asset.split('/')
-    else:
-        # When futures are supported this may need reevaluation.
-        # base, quote = asset, ''
-        raise OperationalException('Not implemented.')
+    market = strategy.dp.market(asset)
+    if market is None:
+        raise OperationalException(f'Market {asset} is not available.')
+    base = market['baseId']
+    quote = market['quoteId']
```
xmatthias: any particular reason to go with "baseId" instead of "base" (Which we are using already in the exchange class ...) ?

rokups: I asked about these in the chat. I got a vague answer from you and/or @samgermain that some exchanges may set 'base' as 'Etherium' while having 'baseId' as 'ETH'. Though in that case it would make other freqtrade code using 'base'/'quote' invalid. So i went ahead and dug in ccxt source code.
```js
            const baseId = this.safeString (market, 'baseAssetName');
            const quoteId = this.safeString (market, 'quoteAssetName');
            const id = this.safeString (market, 'marketId');
            const base = this.safeCurrencyCode (baseId);
            const quote = this.safeCurrencyCode (quoteId);
```
Seems like `baseId` is often source of `base` string. I think i can change it to using `base`/`quote` then.

samgermain: You should use `base` and `quote`

xmatthias (approved):
LGTM :+1:

samgermain:
Does this mean that this task is done? 

> Ensure informative_pairs still works (specifically, the `@informative` pairdetection automatic)

xmatthias:
> Does this mean that this task is done?

leave it open for now please - i'll need to think about it a bit more what it'll mean for that decorator if we have `ETC/USDT:USDT` formatted pairs

PR #5950, opened 2021-11-22
fix typo in Volatility filter description.

flozzone:
fixing a typo which popped out while reading the documentation for the Volatility filter. (`is` -> `it`)


xmatthias (approved):
Great eye for details! :+1: 

Thanks a lot for taking the time to actually fix it :100:

PR #5953, opened 2021-11-23
Is future method fix

samgermain:
On binance it's `future`, instead of `swap`

I don't think the `market_is_margin` method will work for all exchanges, it will work for binance, gateio and okex, but not ftx, and not kraken either I'm pretty sure. I'll have to think about a better solution for that.

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -338,7 +338,7 @@ class Exchange:
         return self.markets.get(pair, {}).get('base', '')
 
     def market_is_future(self, market: Dict[str, Any]) -> bool:
-        return market.get('swap', False) is True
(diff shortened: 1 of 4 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -338,7 +338,7 @@ def get_pair_base_currency(self, pair: str) -> str:
         return self.markets.get(pair, {}).get('base', '')
 
     def market_is_future(self, market: Dict[str, Any]) -> bool:
-        return market.get('swap', False) is True
+        return market.get(self._ft_has["ccxt_futures_name"], False) is True
```
xmatthias, line 341: ah - i missed this... :+1: 

then please remove the "overwriting" method in `binance.py` and `ftx.py` ...

PR #5980, opened 2021-11-27
Shuffle filter use seed only in backtesting mode

ghost:
Shuffle filter should distinguish between modes and apply seed only in backtesting modes, 
in live mode it should always provide a non repeating ordering of pairs.

### DIFF
--- a/freqtrade/plugins/pairlist/ShuffleFilter.py
+++ b/freqtrade/plugins/pairlist/ShuffleFilter.py
@@ -28,6 +28,8 @@ class ShuffleFilter(IPairList):
             self._seed = pairlistconfig.get('seed')
             logger.info(f"Backtesting mode detected, applying seed value: {self._seed}")
 
+        self._random = random.Random(self._seed)
+
     @property
     def needstickers(self) -> bool:
         """
@@ -53,7 +55,6 @@ class ShuffleFilter(IPairList):
         :return: new whitelist
         """
         # Shuffle is done inplace
-        random.seed(self._seed)
-        random.shuffle(pairlist)
+        self._random.shuffle(pairlist)
 
         return pairlist
--- a/freqtrade/plugins/pairlist/ShuffleFilter.py
+++ b/freqtrade/plugins/pairlist/ShuffleFilter.py
@@ -27,7 +27,8 @@ class ShuffleFilter(IPairList):
             self._seed = pairlistconfig.get('seed')
             logger.info("Backtesting mode detected, applying seed value: " + str(self._seed))
 
-        self._random = random.Random(self._seed)
+        # deprecated since 3.9
+        #self._random = random.Random(self._seed)
 
     @property
     def needstickers(self) -> bool:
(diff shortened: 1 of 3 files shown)


incrementby1:
random.Random([seed]) is deprecated since version 3.9, see
https://docs.python.org/3/library/random.html?highlight=random#module-random

ghost:
```
>       if config['runmode'].value in ('live', 'dry_run'):
E       KeyError: 'runmode'
```

I really don't see where this error comes from.
I have tested this code using "freqtrade backtesting" and "freqtrade trade" in commandline and it works. 
When i put 
`       print(str(config['runmode'].value))`
i will either get 'backtest' or 'dry_run' as a result.

xmatthias:
yes - but in tests not everything is 100% initialized all the time (this is testing pairlists - why initialize the full configuration) - so this value is never written to the configuration, and therefore missing from there.

It's a test-problem, not a code problem per se.

i'll have a look when i find time

freqtrade/plugins/pairlist/ShuffleFilter.py
```diff
@@ -18,8 +19,17 @@ def __init__(self, exchange, pairlistmanager,
...
+        # but not in live modes to get a non-repeating order of pairs during live modes.
+        if config.get('runmode') in (RunMode.LIVE, RunMode.DRY_RUN):
+            self._seed = None
+            logger.info("Live mode detected, not applying seed.")
+        else:
+            self._seed = pairlistconfig.get('seed')
+            logger.info(f"Backtesting mode detected, applying seed value: {self._seed}")
+
+        # deprecated since 3.9
+        #self._random = random.Random(self._seed)
```
xmatthias: I'm not sure if this is really deprecated this way.

> Deprecated since version 3.9: In the future, the seed must be one of the following types: NoneType, int, float, str, bytes, or bytearray.
[source](https://docs.python.org/3/library/random.html#alternative-generator)

The way i understand this is that it's deprecated to pass in a value that doesn't correspond to one of the given types.

The same message is also on `random.seed()` which we're using now ([here](https://docs.python.org/3/library/random.html#random.seed))

if we assume that the above means it's deprecated to use this function - then both functions would be deprecated ...

ghost: You are right, seems like i stopped reading at 3.9.

freqtrade/plugins/pairlist/ShuffleFilter.py
```diff
@@ -46,6 +53,7 @@ def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
         :return: new whitelist
         """
         # Shuffle is done inplace
-        self._random.shuffle(pairlist)
+        random.seed(self._seed)
+        random.shuffle(pairlist)
```
xmatthias: something in these 2 lines (and the removal of the 2 lines that it replaces above) bothers me, and i "feel" like it might be introducing a bug / unwanted / unexpected behavior in certain cases - but i can't pin it down what makes me feel this way.

One thing is for sure - running `random.seed(None)` (in live mode) whenever this is called is for sure unnecessary - which i think was the reason we've had this as "random object" in the first place.

I'm currently not sure about the impact to backtesting (when using multiple shuffle-filters in sequence, for example) - but i guess since it's called only once per filter it shouldn't create a problem there.

xmatthias (approved):
LGTM :+1:

PR #5983, opened 2021-11-28
Added raw config and retry config to webhook

PostmanSpat:
## Summary

Needed webhook to talk with a pipe delimited protocol (not form and not json).

## Quick changelog

- Added 'raw' format option to webhook
- Added 'retries' option to webhook
- Added retry_delay option to webhook
- Added 'raw' format to webhook unit test
- Updated webhook docs for raw and retry configs
- Added missing webhook config params for validation

## What's new?

Added a 'raw' option to the webhook config for absolute control over payload formatting. This was needed for an external service that used a specific pipe delimited format for the message payload, and could not include parameter names.

Webhook sometimes has intermittent connection errors, added optional retry configuration. By default retries are disabled, but the user can configure as many retries at they want. There is also a retry delay setting that will execute a sleep between retries. This is default to 100ms.

### DIFF
--- a/freqtrade/rpc/webhook.py
+++ b/freqtrade/rpc/webhook.py
@@ -33,10 +33,6 @@ class Webhook(RPCHandler):
         self._retries = self._config['webhook'].get('retries', 0)
         self._retry_delay = self._config['webhook'].get('retry_delay', 0.1)
 
-        if not (self._format in ['form', 'json', 'raw']):
-            raise NotImplementedError('Unknown webhook format `{}`, possible values are '
-                                      '`form` (default), `json`, and `raw`'.format(self._format))
-
     def cleanup(self) -> None:
         """
         Cleanup pending module resources.
--- a/freqtrade/rpc/webhook.py
+++ b/freqtrade/rpc/webhook.py
@@ -2,6 +2,7 @@
 This module manages webhook communication
 """
 import logging
+import time
 from typing import Any, Dict
 
 from requests import RequestException, post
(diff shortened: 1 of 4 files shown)


freqtrade/rpc/webhook.py
```diff
@@ -28,12 +29,16 @@ def __init__(self, rpc: RPC, config: Dict[str, Any]) -> None:
         super().__init__(rpc, config)
 
         self._url = self._config['webhook']['url']
-
         self._format = self._config['webhook'].get('format', 'form')
+        self._retries = self._config['webhook'].get('retries', 0)
+        self._retry_delay = self._config['webhook'].get('retry_delay', 0.1)
+
+        if self._retries < 0: self._retries = 0
+        if self._retry_delay < 0: self._retry_delay = 0
```
xmatthias: You can combine this with the lines above with `max(0, self._config['webhook'].get('retries', 0))` (and retry delay obviously).

potentially/most likely even better, add this as property to `constants.py` (search for webhooks) - and specify "minimum" as 0 (there's other examples for this there). 

That way you can rely config validation to give you only correct values, and freqtradebot will fail to start in case of wrong values (which is usually better than silently fixing values to something the user may not desire).

freqtrade/rpc/webhook.py
```diff
@@ -77,13 +82,28 @@ def send_msg(self, msg: Dict[str, Any]) -> None:
...
+                if self._format == 'form':
+                    response = post(self._url, data=payload)
+                elif self._format == 'json':
+                    response = post(self._url, json=payload)
+                elif self._format == 'raw':
+                    response = post(self._url, data=payload['data'], headers={'Content-Type': 'text/plain'})
+                else:
+                    raise NotImplementedError('Unknown format: {}'.format(self._format))
+            
+                """throw a RequestException if the post was not successful"""
```
xmatthias: Please use comments  (`# ...`) instead of multiline strings to comment.

While this seems to be the same, it's actually an executable line (which will be executed every time) - which is really unnecessary for comments.

PostmanSpat: I normally would, but I saw everything else in that file was using """ so I was just sticking with the convention.

freqtrade/rpc/webhook.py
PostmanSpat: Ok, I did this for my new params. But I noticed that 'url' and 'format' has not been defined in constants.py. Format could also potentially be defined as an enum, then presumably the validation within webhook.py would not be needed for that either. I can also see that the ...fill events are not defined in constants.py. Do you want me to clean all this up for consistency, or perhaps do it as a separate PR?

xmatthias: i noticed this myself while making the above comment.

We'll for sure want to have this in the validation (and correct) - but i won't request fixing existing issues as change to a PR that's not "really" related to that - so it's up to you if you want to fix it here, separately, or leave it up to someone else (me) to fix this afterwards :)

PostmanSpat: Done. Saved you the hassle :)

xmatthias (approved):
Nice change, thanks a lot for your contribution! :100: 

LGTM :+1:

PR #6011, opened 2021-11-30
correctly apply leverage to backtesting

xmatthias:
Fix "absolute" gains issue in backtesting when using leverage

TODO:
* [x] add at least one test that confirms this

### DIFF
--- a/tests/optimize/test_backtest_detail.py
+++ b/tests/optimize/test_backtest_detail.py
@@ -536,6 +536,23 @@ tc33 = BTContainer(data=[
     )]
 )
 
+# Test 34: (copy of test25 with leverage)
+# Sell with signal sell in candle 3 (stoploss also triggers on this candle)
+# Stoploss at 1%.
+# Sell-signal wins over stoploss
+tc34 = BTContainer(data=[
+    # D  O     H     L     C     V    B  S
+    [0, 5000, 5025, 4975, 4987, 6172, 1, 0],
+    [1, 5000, 5025, 4975, 4987, 6172, 0, 0],  # enter trade (signal on last candle)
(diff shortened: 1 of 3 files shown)


tests/optimize/test_backtest_detail.py
```diff
@@ -593,14 +611,19 @@ def test_backtest_results(default_conf, fee, mocker, caplog, data) -> None:
 
     mocker.patch("freqtrade.exchange.Exchange.get_fee", return_value=0.0)
     mocker.patch("freqtrade.exchange.Exchange.get_min_pair_stake_amount", return_value=0.00001)
+    mocker.patch("freqtrade.exchange.Binance.get_max_leverage", return_value=100)
     patch_exchange(mocker)
     frame = _build_backtest_dataframe(data.data)
     backtesting = Backtesting(default_conf)
     backtesting._set_strategy(backtesting.strategylist[0])
     backtesting.required_startup = 0
+    if data.leverage > 1.0:
```
samgermain, line 620: Is data something that the strategy returns? The strategy can decide to open a short at 1x leverage.

xmatthias: this is a ... different kind of test :laughing: 
you need to treat the whole file as "one" test (with parameters though)

data is simply a `BTContainer` object (an object helping define the backtesting testcase).
in real life, backtesting can define leverage, that's correct.
 
In this test/file - there is however no (real) strategy - only mocked data.
It's sole purpose is to test different behaviors in different situations during backtesting - most of which originating from one or the other issue where freqtrade behaved badly (speak - backtesting was not doing what it should, for example with trailing stop). 

This test helps in testing that behavior - as it'll allow for (pretty easy) test case creation (including the candles used), where to buy/sell and what profit to expect.

PR #6038, opened 2021-12-10
Hyperopt default trailing stop fix

Zegelink:
## Summary

Solve the issue: Minor fix on hyperopt default trailing stoploss searching spaces. The old code was wrong that the trailing_stop_positive is greater than trailing_stop_positive_offset. The documentation is also changed to reflect this fix.

### DIFF
--- a/freqtrade/optimize/hyperopt_interface.py
+++ b/freqtrade/optimize/hyperopt_interface.py
@@ -180,14 +180,14 @@ class IHyperOpt(ABC):
             # other 'trailing' hyperspace parameters.
             Categorical([True], name='trailing_stop'),
 
-            SKDecimal(0.01, 0.35, decimals=3, name='trailing_stop_positive'),
+            SKDecimal(0.001, 0.1, decimals=3, name='trailing_stop_positive'),
 
             # 'trailing_stop_positive_offset' should be greater than 'trailing_stop_positive',
             # so this intermediate parameter is used as the value of the difference between
             # them. The value of the 'trailing_stop_positive_offset' is constructed in the
             # generate_trailing_params() method.
             # This is similar to the hyperspace dimensions used for constructing the ROI tables.
-            SKDecimal(0.001, 0.1, decimals=3, name='trailing_stop_positive_offset_p1'),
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/hyperopt_interface.py
```diff
@@ -180,14 +180,14 @@ def trailing_space(self) -> List[Dimension]:
...
-            SKDecimal(0.01, 0.35, decimals=3, name='trailing_stop_positive'),
+            SKDecimal(0.001, 0.1, decimals=3, name='trailing_stop_positive'),
 
             # 'trailing_stop_positive_offset' should be greater than 'trailing_stop_positive',
             # so this intermediate parameter is used as the value of the difference between
             # them. The value of the 'trailing_stop_positive_offset' is constructed in the
             # generate_trailing_params() method.
             # This is similar to the hyperspace dimensions used for constructing the ROI tables.
-            SKDecimal(0.001, 0.1, decimals=3, name='trailing_stop_positive_offset_p1'),
+            SKDecimal(0.02, 0.35, decimals=3, name='trailing_stop_positive_offset_p1'),
```
xmatthias, line 190: I don't think this change is correct based on how the space is interpreted / used.
for reference: 
https://github.com/freqtrade/freqtrade/blob/be6b1f6f8399b05dcfcddcf63b573052b21beb96/freqtrade/optimize/hyperopt_interface.py#L156-L166

`trailing_stop_positive_offset_p1` does NOT correspond to `trailing_stop_positive_offset` directly, but to `trailing_stop_positive + trailing_stop_positive_offset_p1` (see below).

As such, your new space now requires a profit of between 2% and 35%, before starting to trail with anything between 0.1% to 10%.
0.1% is clearly too tight and will exploit the backtesting mechanics, as it'll then always sell at the best possible position within one candle.

The way it was seemed better / more correct to me, as it required a profit of 1.1%, before trailing with a mininmum of 1%.

Now one can argue if a trailing stop of 35% is too high (it probably is) - but it will depend on the strategy itself (for a mean reversion strategy it's too high, for a trend following ... not so sure).

xmatthias:
Closing based on the above explanation.

Issue #5562, opened 2021-09-11
Auto buy bnb when it is below a certain threshold


hansen1015:
Is it possible to add functionality to auto buy bnb for trading fee when the bnb balance is below a certain value?

Thank you

xmatthias:
it's currently not implemented, and i don't see a big deal in buying enough manually.

Feel free to impement if you think it's a must (i don't think so)
But please in a coin/exchange-agnostic way, which means you configure "BNB" for binance, and maybe FTT for FTX (not sure that's correct) ...
so it'll also work for other exchanges which have similar "fee" coins.

Merinorus:
Also interested by this feature, I will try to look into it if no one does it before.
I'm thinking about this in the config.json:

```
{
"reduced_fee_currency": "BNB",
"reduced_fee_max_ratio": 0.01,
"reduced_fee_min_ratio": 0.005
}
```

The bot would buy up to 1% of the stake if it drops below 0.5%. What do you think?

xmatthias:
i'd make it absolute BNB to be honest.

you'll not need an insane amount of BNB ... just enough to keep the bot paying for it in BNB.
Assuming you use compounding, 1% of stake-amount can be anything from 0.2$ (20$ stake) up to 1000$ (10k$ stake) ...

Merinorus:
Well then I would lower the ratio to 0.15% for instance. In Binance, the fees are 0,075% in BNB for each trade, so assuming you buy then sell all your stake it would cost about 0,15% of your stake in BNB.
Let's say I have a 100$ stake wi 90% tradable balance ratio. The bot would buy 0.15% * 90% * 100$ = 0,135$ worth of BNB. It's too low anyway so the bot would buy the minimum allowed amount (maybe 10$, I don't know how much it is).
Now let's say it compounds up to 100 000$. Then the bot would buy 0.15% * 90% * 100 000$ = 135$, which would be adequate for this stake amount.

Maybe we could provide both solution (percentage and fixed amount) with amount advices of default values. Indeed, a 1% stake amount is way too much for this use case.

xmatthias:
having a "list of currencies" will not make much sense - as you'll have only one fee currency.

to be fair - if your bot compounds form 1000$ to 100.000$ - you'll have other problems if you never changed any parameters (like increasing parallel trades). (trades will not fill, or you'll move the market if using market orders).

the problem is - most people won't think about 100_000$ - but about a 1000$ (or even 100$) wallet.
[...]

PR #6059, opened 2021-12-12
Auto buy BNB and self._log_responses patch

NoahGago:

## Summary

1. Automatically buy BNB for trading fees (only in dry run) based on analysis of typical FreqTrade users.
2. Fixed the self._log_responses variable misnomer, I see that in the fully updated version it's been fixed to be self.log_responses, however I believe that changing the instances of its use to self._log_responses more conforms with naming protocols.

+ Potential early stage solution to the issue: #5562 
+ Variable misnomer patch (self._log_responses in exchange.py)

These edits were made to explore the possibility of buying BNB automatically in order to pay the minimum in trading fees. It was only implemented in dry-run mode however. This was done with changes made to the create_order method in exchange.py, changes to config_bittrex.example.json, and changes to config_binance.example.json. Additionally when testing with Travis CI I noticed that there was an inconsistency with the usage of self._log_responses in exhange.py so I changed the instances of its use to self._log_responses as opposed to self.log_responses.

### DIFF
--- a/config_examples/config_bittrex.example.json
+++ b/config_examples/config_bittrex.example.json
@@ -26,6 +26,13 @@
     },
     "exchange": {
         "name": "bittrex",
+        "reduced_fees_strategy": {
+            "enable": true,
+            "process_throttle_secs": 3600,
+            "currency": "BNB",
+            "BNB_maximum": 0.35,
+            "BNB_minimum": 0.012
+        },
         "key": "your_exchange_key",
         "secret": "your_exchange_secret",
         "ccxt_config": {"enableRateLimit": true},
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -396,7 +396,7 @@ class Exchange:
             # Note: ccxt has BaseCurrency/QuoteCurrency format for pairs
             if self.markets and pair not in self.markets:
                 raise OperationalException(
-                    f'Pair {pair} is not available on {self.name}. '
+                    f'Pair {pair} is not available on {self.name}. ' 
                     f'Please remove {pair} from your whitelist.')
 
                 # From ccxt Documentation:
@@ -738,11 +738,9 @@ class Exchange:
             balances = self._api.fetch_balance(self)
             if (balances["free"] > self._BNB_min and self._name == "binance" and
                self._avoid_duplicate_order is False):
-                print("in line 741")
                 self.create_order(pair, ordertype, side, (self._BNB_max - balances["free"]), rate)
                 self._avoid_duplicate_order = True
             else:
-                print("in line 744")
                 self._avoid_duplicate_order = False
             dry_order = self.create_dry_run_order(pair, ordertype, side, amount, rate)
             return dry_order
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -110,7 +110,11 @@ class Exchange:
             logger.info('Instance is running with dry_run enabled')
         logger.info(f"Using CCXT {ccxt.__version__}")
         exchange_config = config['exchange']
-        self.log_responses = exchange_config.get('log_responses', False)
+        self._name = exchange_config['name']
+        self._BNB_min = exchange_config['reduced_fees_strategy']['BNB_minimum']
+        self._BNB_max = exchange_config['reduced_fees_strategy']['BNB_maximum']
+        self._avoid_duplicate_order = False
+        self._log_responses = exchange_config.get('log_responses', False)
 
         # Deep merge ft_has with default ft_has options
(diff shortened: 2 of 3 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -1323,30 +1324,27 @@ def refresh_latest_ohlcv(self, pair_list: ListPairsWithTimeframes, *,
...
-                    self._pairs_last_refresh_time[(pair, timeframe)] = ticks[-1][0] // 1000
-                # keeping parsed dataframe in cache
-                ohlcv_df = ohlcv_to_dataframe(
-                    ticks, timeframe, pair=pair, fill_missing=True,
-                    drop_incomplete=self._ohlcv_partial_candle)
-                results_df[(pair, timeframe)] = ohlcv_df
-                if cache:
-                    self._klines[(pair, timeframe)] = ohlcv_df
+        results = asyncio.get_event_loop().run_until_complete(
+            asyncio.gather(*input_coroutines, return_exceptions=True))
```
xmatthias, line 1328: I don't think changes to this (and subsequent) lines are intentional.

Please don't use your own develop branch for pull requests, but use dedicated branches instead.
Using the "main" branch from another repository usually ends up in chaos.

config_examples/config_bittrex.example.json
```diff
@@ -26,6 +26,13 @@
     },
     "exchange": {
         "name": "bittrex",
+        "reduced_fees_strategy": {
```
xmatthias, line 29: this is definitely the wrong file - BNB is irrelevant for bittrex

freqtrade/exchange/exchange.py
```diff
@@ -737,14 +735,20 @@ def create_order(self, pair: str, ordertype: str, side: str, amount: float,
                      rate: float, time_in_force: str = 'gtc') -> Dict:
 
         if self._config['dry_run']:
+            balances = self._api.fetch_balance(self)
+            if (balances["free"] > self._BNB_min and self._name == "binance" and
+               self._avoid_duplicate_order is False):
+                self.create_order(pair, ordertype, side, (self._BNB_max - balances["free"]), rate)
+                self._avoid_duplicate_order = True
+            else:
+                self._avoid_duplicate_order = False
```
xmatthias, line 744: i think this section is located wrongly for 2 reasons

buying BNB is irrelevant in dry-run ... so if any, then this should be moved 2 lines to the bottom (and dedented by 4 spaces).

That said - i don't think this should be part of `create_order()` at all. 
Assuming a quick timeframe (1m) - we should try to be as quick as possible - another call to buy BNB is definitely the wrong way (and it's also unnecessary. 

I think this should be running "out of band" - at some random (inactivity) times - so basically, it should be a separate function that's called as part of the bot loop (but only acts every X minutes/hours).

freqtrade/exchange/exchange.py
```diff
@@ -110,7 +110,11 @@ def __init__(self, config: Dict[str, Any], validate: bool = True) -> None:
             logger.info('Instance is running with dry_run enabled')
         logger.info(f"Using CCXT {ccxt.__version__}")
         exchange_config = config['exchange']
-        self.log_responses = exchange_config.get('log_responses', False)
+        self._name = exchange_config['name']
+        self._BNB_min = exchange_config['reduced_fees_strategy']['BNB_minimum']
+        self._BNB_max = exchange_config['reduced_fees_strategy']['BNB_maximum']
```
xmatthias, line 115: i'd not use BNB in the names (not for variables, not for config options).
While you're using it on binance where BNB are relevant, others might be using it on different exchanges on different coins.

xmatthias (changes requested):
i think your branch got messed up somehow - as it seems to roll back certain chances we made lately

HaddadJoe:
Nice suggestion, I'm interested in this feature

@NoahGago are you still working on this? I can pick up where you left and clean that PR if you're not.

NoahGago:
Nope I am not! You're free to pick up where I left off!

xataxxx:
Totally agree that this should be exchange-agnostic - on KuCoin one might want to buy KCS, on FTX exchange you would want to buy FTT etc.

Besides running at Xh intervall maybe consider running at startup + at the end of SELL order finishing.
Basically re-check at startup + after each sell order closing check if we have enough fee coin left and buy more if needed.

Idea for future (way overkill for first implementation):
Would be awesome if we could calculate and check the full dataframe for the <exchange-coin>/<stake-currency> before buying so that the strategy could advise the bot if the BNB/KCS/FTT pair is having a bad moment and maybe skip buying if some indicators indicate that the price is very high temporarily etc. Then you could also "optimise" the buying of your fee currency when the BNB/KCS/FTT etc is currently selling relatively low.

xmatthias:
Going to close this as it's abandoned, and there's still too much missing for this to be used as reasonable base.

Issue #5106, opened 2021-06-08
Showing on CAGR and MDD in backtest results


systemquant:
This is a very simple enchantment, but I think it is so helpful for intuitively comparing strat's performances.

<img width="656" alt="Example" src="https://user-images.githubusercontent.com/85345804/121263953-5bef1200-c8f1-11eb-9e71-89b45f29d48c.PNG">
Such as this image, showing CAGR and MDD is extremely intuitive for comparing strategies, higher CAGR(profitability) and lower MDD(risk) is better!

It's about showing CAGR (geometric mean of yearly profits, `(current total stake)/(starting wallet)^(1/years passed)`, (simply,) years_passed = days_passed/365)  and MDD (`((strat's)max - min / max) * 100%`, currently freqtrade has the _drawdown_, but it is unintuitive because it gives ratio based on starting wallet stakes).

This will make freqtrade's backtests look more intuitive and easier to compare strats.
How do you guys think about this very little enchantments?

systemquant:
~~_**Consideration:**_ (this is not essential)
Maybe adding the sharp ratio for backtest can be also good. And we have codes in the hyperopt so all we need to do is just using that code. But can there be a little performance changes for backtesting? If so, just adding option params(such as --sharp true) to show sharp ratio might be good.~~

xmatthias:
we do have an issue for sharp already (#1907 ) - but so far, it was not important enough for anyone to actually spend time to implement it.

We have a section for strategy metrics - i'll surely not mind having others added there as well if they make sense.
I'm however not sure what MDD will be based on - you say "max - min / max" ... but max / min of what? absolute wallet? relative profits ... you can do min/max calculations on many things - so it's kinda unclear to me for now how this calculation should look like.

systemquant:
That means absolute wallet including current open positions(maximum total stake - minimum total stake / maximum total stake)

I think it has already there named max drawdown high and low. That will do exact thing. 
The problem of current drawdown ratio is it is calculated through (high-low / starting stake) but it is not intuitive because profit compounds, so if drawdown happened later on, it gives bigger ratio, so it highly depends on when max drawdowns happened. So adding one more metrics that does not influenced by compounding. (And I found this exact thing extremely useful in other systemic trading tools, they are using it too.)

![1](https://user-images.githubusercontent.com/85345804/121329020-c9319000-c94f-11eb-94c5-42f5bf9f4354.png)
This means (-)21% MDD 

[...]

PR #6065, opened 2021-12-13
Mdd on trades backtest stat

Guitheg:
## Summary

Implementation of the MDD on the trades intervals

Solve the half of this issue: ##5106

## Quick changelog

# Add
- new function calculate_trades_mdd
- visualization of the result of the new function in the stats board at the end of a backtest
- unit test of the function

## What's new?
Added mdd on trades stat on the stats board at the end of backtest
image

### DIFF
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -414,9 +414,10 @@ def calculate_mdd(data: dict, trades: pd.DataFrame, *, date_col: str = 'close_da
     for pair, df in data.items():
 
         open_close_trades = trades.loc[trades['pair']==pair][["open_date","close_date"]]
-        open_close_trades = pd.concat([open_close_trades.rename(columns={'open_date':'date'})[['date']],
-                                       open_close_trades.rename(columns={'close_date':'date'})[['date']]]
-                                     ).sort_values(by='date')
+        open_close_trades = pd.concat(
+            [open_close_trades.rename(columns={'open_date':'date'})[['date']],
+            open_close_trades.rename(columns={'close_date':'date'})[['date']]]
+            ).sort_values(by='date')
         open_close_trades['open_close_mark'] = np.ones(len(open_close_trades), dtype=int)
         data_join = df.set_index('date').join(open_close_trades.set_index('date'))
         del open_close_trades
@@ -427,24 +428,29 @@ def calculate_mdd(data: dict, trades: pd.DataFrame, *, date_col: str = 'close_da
         data_join = data_join.join(close_trades.set_index('date'))
         del close_trades
 
-        data_join[["open_close_mark",'close_mark']] = data_join[["open_close_mark",'close_mark']].fillna(0).astype(int)
+        data_join[["open_close_mark",'close_mark']] = data_join[
+            ["open_close_mark",'close_mark']].fillna(0).astype(int)
+        
         data_join['is_in_trade'] = data_join.open_close_mark.cumsum()&1 # &1 <=> %2
         data_join.loc[data_join['open_close_mark'] == 1, 'is_in_trade'] = 1
         data_join['close_cummax'] = 0
 
         data_join["open_close_mark"] = data_join["open_close_mark"] - data_join["close_mark"]
         
-        data_join['close_cummax'] = data_join.groupby(data_join.open_close_mark.cumsum()).close.cummax()
+        data_join['close_cummax'] = data_join.groupby(
+            data_join.open_close_mark.cumsum()
+            ).close.cummax()
         data_join.loc[data_join['is_in_trade'] == 0, 'close_cummax'] = 0
 
         data_join = data_join.rename(columns={'close_mark':'drawdown'})
-        data_join['drawdown'] = data_join['close_cummax'] - data_join['close']
+        data_join.loc[data_join['is_in_trade'] == 1, 'drawdown'] =             \
+                      (data_join['close_cummax'] - data_join['close'])         \
+                      / data_join['close_cummax']
 
         mdd_pair = data_join['drawdown'].max()
         mdd_pair_list.append(mdd_pair)
-        # print(mdd_df)
     
-    mdd_pair_list = np.ndarray(mdd_pair_list)
+    mdd_pair_list = np.array(mdd_pair_list)
 
 
     return mdd_pair_list.max()
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -396,12 +396,12 @@ def calculate_trades_mdd(data: dict, trades: pd.DataFrame)  -> float :
         Give the max drawdown given each trades and the history candles.
         The intervals dates used to calculate the max drawdown are the trades intervals dates.
     Args:
-        :param data (dict) : dictionnary of candle dataframe per pair used to calculate the mdd.
-        :param trades (pd.DataFrame): trades used to find the intervals dates.
+        :param data: (dict) dictionnary of candle dataframe per pair used to calculate the mdd.
+        :param trades: (pd.DataFrame) trades used to find the intervals dates.
     
     Returns:
-         :return: float: Give the maximum drawdown among each trades.
-         :raise: ValueError if trade-dataframe was found empty.
+         :return: (float) Give the maximum drawdown among each trades.
+         :raise: (ValueError) if trade-dataframe was found empty.
     """
     if len(trades) == 0:
         raise ValueError("Trade dataframe empty")
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -390,7 +390,7 @@ def calculate_max_drawdown(trades: pd.DataFrame, *, date_col: str = 'close_date'
     low_val = max_drawdown_df.loc[idxmin, 'cumulative']
     return abs(min(max_drawdown_df['drawdown'])), high_date, low_date, high_val, low_val
 
-def calculate_mdd(data: dict, trades: pd.DataFrame, *, date_col: str = 'close_date',
+def calculate_trades_mdd(data: dict, trades: pd.DataFrame, *, date_col: str = 'close_date',
                            value_col: str = 'profit_ratio'
                  )  -> float:
     """
@@ -407,9 +407,7 @@ def calculate_mdd(data: dict, trades: pd.DataFrame, *, date_col: str = 'close_da
     if len(trades) == 0:
         raise ValueError("Trade dataframe empty")
     
-    mdd_df = pd.DataFrame()
-
-    mdd_pair_list = []
+    trades_mdd_pair_list = []
     
     for pair, df in data.items():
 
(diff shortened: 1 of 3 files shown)


xmatthias:
Something seems broken with this - but i'm not sure what (i suspect it's the deletion of the dataframe in the function, but not 100% sure).

develop 
```
=============== SUMMARY METRICS ================
| Metric                 | Value               |
|------------------------+---------------------|
| Backtesting from       | 2021-01-01 00:00:00 |
| Backtesting to         | 2021-10-08 16:00:00 |
| Max open trades        | 2                   |
|                        |                     |
| Total/Daily Avg Trades | 429 / 1.53          |
| Starting balance       | 1000.000 USDT       |
| Final balance          | 987.497 USDT        |
| Absolute profit        | -12.503 USDT        |
| Total profit %         | -1.25%              |
| Trades per day         | 1.53                |
| Avg. daily profit %    | -0.00%              |
| Avg. stake amount      | 20.000 USDT         |
| Total trade volume     | 8580.000 USDT       |
|                        |                     |
| Best Pair              | MATIC/USDT 33.03%   |
| Worst Pair             | XTZ/USDT -39.20%    |
| Best trade             | WAVES/USDT 26.55%   |
| Worst trade            | XMR/USDT -3.89%     |
| Best day               | 7.170 USDT          |
| Worst day              | -3.896 USDT         |
| Days win/draw/lose     | 87 / 82 / 112       |
| Avg. Duration Winners  | 17:04:00            |
| Avg. Duration Loser    | 7:02:00             |
| Rejected Buy signals   | 41436               |
|                        |                     |
| Min balance            | 977.748 USDT        |
| Max balance            | 1007.028 USDT       |
| Drawdown               | 126.17%             |
| Drawdown               | 25.259 USDT         |
| Drawdown high          | 7.028 USDT          |
| Drawdown low           | -18.231 USDT        |
| Drawdown Start         | 2021-05-18 16:00:00 |
| Drawdown End           | 2021-07-21 20:00:00 |
| Market change          | 676.00%             |
================================================
```

This branch
```
=================== SUMMARY METRICS ====================                                                                           
| Metric                   | Value                     |                                                                           |--------------------------+---------------------------|                                                                           
| Backtesting from         | 2021-01-01 00:00:00       |                                                                           
| Backtesting to           | 2021-10-08 16:00:00       |                                                                           
| Max open trades          | 2                         |                                                                           
|                          |                           |                                                                           
| Total/Daily Avg Trades   | 429 / 1.53                |                                                                           
| Starting balance         | 1000.000 USDT             |                                                                           
| Final balance            | 987.497 USDT              |                                                                           
| Absolute profit          | -12.503 USDT              |                                                                           
| Total profit %           | -1.25%                    |                                                                           
| Trades per day           | 1.53                      |                                                                           
| Avg. daily profit %      | -0.00%                    |                                                                           
| Avg. stake amount        | 20.000 USDT               |                                                                           
| Total trade volume       | 8580.000 USDT             |                                                                           
|                          |                           |                                                                           
| Best Pair                | MATIC/USDT 33.03%         |                                                                           
| Worst Pair               | XTZ/USDT -39.20%          |                                                                           
| Best trade               | WAVES/USDT 26.55%         |                                                                           
| Worst trade              | XMR/USDT -3.89%           |                                                                           
| Best day                 | 7.170 USDT                |                                                                           
| Worst day                | -3.896 USDT               |                                                                           
| Days win/draw/lose       | 87 / 82 / 112             |                                                                           
| Avg. Duration Winners    | 17:04:00                  |                                                                           
| Avg. Duration Loser      | 7:02:00                   |                                                                           
| Rejected Buy signals     | 41436                     |                                                                           
|                          |                           |                                                                           
| Min balance              | 0.000 USDT                |         <-- Everything from here to the bottom is wrong except for Market change.
| Max balance              | 0.000 USDT                |                                                                           
|                          |                           |         
| Max Drawdown (in trades) | 0.00%                     |                                                                           
| Max Drawdown Total       | 0.00%                     |         
| Max Drawdown Total       | 0.000 USDT                |                                                                           
| Drawdown high            | 0.000 USDT                |         
| Drawdown low             | 0.000 USDT                |                                                                           
| Drawdown Start           | 1970-01-01 00:00:00+00:00 |                                                                           
| Drawdown End             | 1970-01-01 00:00:00+00:00 |                                                                           
| Market change            | 676.00%                   |                                                                           
======================================================== 
```


As you can see, all drawdown (as well as min/max) balances) are set to 0 for no apparent reason, which indicates a bug "someplace".

---

apart from that (i can't test it myself due to the above at the moment) - from a first look, it looks like you're calculating the maximum drawdown "per pair" - and then use the maximum of that (basically showing the biggest drawdown for a single pair).
I don't think that's what MDD is supposed to show - as it simply ignores wins from all other pairs during the same period.

Assume i trade 3 pairs - XRP/USDT, BTC/USDT and ETH/USDT for the span of a month - with a max-open-trades of 1 (so i go all in on each pair when a trade is opened).
All trades are bought in sequence (always XRP/BTC/ETH - in the same sequence) - for a few rounds (say 5 rounds, which will give me 3x5 = 15 trades) - and each stop / take profit is at 10% (so all trades are even).

I lose all XRP trades - but win all BTC and ETH trades ... (so the resuts are 5x -10% and  10x + 10%).

The way i understand this implementation, then this would show the max-drawdown for the XRP trades (5 losses, resulting in 50% loss on XRP) - but completely ignores the BTC and ETH wins for 100%. 
In summary, the account would be "roughly" 50% up (depends if we use compounding or not)


based on the [investopedia article](https://www.investopedia.com/terms/m/maximum-drawdown-mdd.asp)  - this should be calculated against the portfolio - not using the maximum of single pairs.

Guitheg:
I'm still investigate to know where it broke. I think the function sent a ValueError due to a bad using of isinstance used to check the dataframe.

About the implementation, you understood well what we did with @EmbraysiteReal. We based our implementation about what we read and what we understood. In fact, what we did looks like the investopedia method. But I think the explanation gave by investopedia is better considering the robustness and the generalization of the method. With @EmbraysiteReal we propose this definition of the max drawdown based on investopedia method. we formalize it mathematically to avoid ambiguity. (however tell us if you need clarification)
image

I have a question about the current max drawdown. The current max drawdown implemented on freqtrade consider only profit (so at a trade closing) to perfrom its calculation. In other words, the date k (defined in the equations) is updated only when a trade is closing. Then I think the drawdown did not record well all the evolution of the portfolio. Am I right ? Or there a something I did not understand ?

xmatthias:
If you're looking for the "current outstanding" balance, maybe look at #3059 before trying to implement it directly / again.

In my opinion, that PR is "mostly" complete (at least for that function) - the only thing that I'm missing there really is a "proper" usecase (i don't like adding code just for the fun of it - and i don't think the loss function that's shown there  as sample usecase makes much sense).

based on that - i think you could base your drawdown calculation on the outcome of that function - as that would be the "real" max drawdown, including non-realized profits.

Guitheg:
so, instead of adding the `calculate_trades_mdd` function. You propose to just rewrite the calculation of the current `calculate_max_drawdown` function following the definition we previously gave ?

We will begin to answer #3059. Indeed it would be useful to have such a function in order to calculate the max drawdown.

xmatthias:
neither - `calculate_max_drawdown` works perfectly fine and according to spec - but results will vary depending on the input / selected column.

In my understanding, the investopedia definition can be archived pretty easily by using `calculate_max_drawdown` against absolute profits (we do this for the absolut profit calculation already). 
To then get the "relative" value - you'd calculate the % difference between the returned high and low value.

That said, it'll obviously not include non-realized losses - which can however be accomplished by first running the data through the "outstanding profits" function.

EmbraysiteReal:
Hello, could you send to us the spec that you used for `calculate_max_drawdown` ? We don't understand what it represent since we only know the investopedia MDD definition.

xmatthias:
It's a very simple, drawdown calculation.

it's searching the highest point in a curve - followed by the subsequent deepest valley.
If you use `freqtrade plot-profit` you'll see the highest / lowest points - which will make the way of calculation clear.

EmbraysiteReal:
It's weird, why is there an issue to implement MDD, which mean max drawdown, if `calculate_max_drawdown` already works perfectly ?
And i don't know if you saw result like i did, but sometimes the drawdown can be superior than 100%, this is really weird and completely out of meaning for me, maybe i broke something

xmatthias:
Let me repeat - `calculate_max_drawdown` works well (from a function perspective) - but depends on the input given (you can see this with absolute drawdowns, which you'll notice if you compare it to the plot - is accurate).

Relative drawdown (like -120%) which is currently used are not wrong, just need some context for correct interpretation.
This is discussed in parts in #5106, and in more depth in #5167.
Nobody said the calculation per se is wrong (you assumed this without checking how the calculation actually worked) - just that what is shown is confusing (it's also only wrong in certain conditions - see below).

You can even test that pretty easily with a random series (in a jupyter notebook) - and plot the results as well as high/low points.

First of all, it does only consider finished / profits. 
When using a sane stoploss (10% or less) - this is not a problem.
With strategies that use a very wide or no stoploss (99%) - this will make drawdown seem great - while in reality it might not become a problem - but again, it's not something i'd see as critical.

The bigger problem in my opinion is the aggregation of relative profits.
When the metric was developed - only static stake-amounts were allowed - where it works perfectly fine (10% profit is always 10% per `stake_amount`).
With the introduction of "unlimited" stake amount, profits can compound (which is good in general for profits - usually).

However, you can no longer assume that 10% + -10% = 0%.
it could be 10% * 50$ - and -10% * 2000$ - so now we're -1950$ - BUT - relative aggregation still shows 0% profits.

THIS is the real problem with (relative) drawdown calculation. It uses the relative cumulative profit - while it really should not.
in my opinion, it can be pretty easily fixed by using the absolute values (drawdown high value+drawdown low value) we already show as metrics - and simply calculate the relative difference between the two (so from a high 2000$ to a low of 500$ - you've had a drawdown of 75%.


Now the first point - using "intermediate" profits - might be seen as separate point to tackle - but again, won't require a new calculating function - but simple data preparation (basically what is implemented in #3059 - maybe with some modifications) before passing the data into the drawdown calculation function.

EmbraysiteReal:
Thank you a lot for your answer, there is a lot of what you'r saying that we saw. It clarify a lot of things, the aggregation of relative profits was an aberration from my mathematician point of view

"Nobody said the calculation per se is wrong (you assumed this without checking how the calculation actually worked)" we checked the calculation and we saw for exemple the aggregation of relative profits, wich is a wrong way of calculation, then we tried to understood why it was done this way then we assume they were something wrong. I don't even understand what you mean by "calculation per se".

We are new to the project and we don't want to pretend that we know more than others, but when we found somethings we don't understand in the code we can't ignore it like nothing happened. There is no documentation and comments on the code are minimalist, it's normal to have misunderstanding

Well you're right, we have to check #3059 since it's a big part of the solution we proposed. We will try to make it optimized as possible.
And again, thanks for the clear answer

xmatthias:
> I don't even understand what you mean by "calculation per se".

the calculation of the method `calculate_max_drawdown` is correct depending on the inputs. If you use `value_col='profit_abs'` as parameter (as we do for absolute profits) - then the results are correct.
they are also technically correct for `value_col='profit_ratio'` - but ONLY if stake-amount is static.

> There is no documentation and comments on the code are minimalist

Code is usually self-documenting.
it just requires some understanding of pandas (how pandas functions work, what they do/return) - and carefully thinking through the method - eventually multiple times.

Even a comment would only tell you `calculate cumsum over value_col` - if you don't understand what a cumulative sum is, that won't help you. If you do - the code already told you what you needed to know, so the comment is pointless.

--

We do have quite extensive user documentation.
I'll however not start documenting internal functions more than they already are.

They mostly have a doc-string describing what they do, as well as the parameters defined (what is expected).

The implementation itself is 
a) subject to change in case of bugs
b) clear from the code

xmatthias:
Closing this in favor of #6165 - which uses the existing (and working) drawdown calculation to calculate account drawdown.

PR #6070, opened 2021-12-14
Suppress additional logs for pairs in blacklist

cdimauro:
All preliminary checks made and no issue found.
Output of pytest: 1737 passed, 1 skipped, 42 deselected, 64 warnings in 155.08s (0:02:35).
Output of flake8: .
Output of mypy: Success: no issues found in 152 source files.
Output of isort: Skipped 5 files

## Summary

Suppresses the continuous logs which are generated at every tick for removing pairs from the blacklist: it just logs them the first time.

## Quick changelog

- Add test case for checking removal of logs for pains in blacklist
- Add type annotation to new logs suppression code
- Remove unused import
- Suppress additional logs for pairs in blacklist

## What's new?

Every time that there's freqtrade "ticks", pairs in the blacklist are
checked and a warning message is displayed.
So, the logs are continuously flooded with the same warnings.

For example:
```
2021-07-26 06:24:45 freqtrade.plugins.pairlistmanager: WARNING -
Pair XTZUP/USDT in your blacklist. Removing it from whitelist...
2021-07-26 06:24:45 freqtrade.plugins.pairlistmanager: WARNING -
Pair SUSHIUP/USDT in your blacklist. Removing it from whitelist...
2021-07-26 06:24:45 freqtrade.plugins.pairlistmanager: WARNING -
Pair XTZDOWN/USDT in your blacklist. Removing it from whitelist...
2021-07-26 06:24:50 freqtrade.plugins.pairlistmanager: WARNING -
Pair XTZUP/USDT in your blacklist. Removing it from whitelist...
2021-07-26 06:24:50 freqtrade.plugins.pairlistmanager: WARNING -
Pair SUSHIUP/USDT in your blacklist. Removing it from whitelist...
2021-07-26 06:24:50 freqtrade.plugins.pairlistmanager: WARNING -
Pair XTZDOWN/USDT in your blacklist. Removing it from whitelist...
```

This patch shows the warning only the first time, by keeping track
of which pairs in the blacklist were already logged.

It also removes an unused import, adds a type annotation for the set which is used for keep track of the pairs which were already logged, and introduces a specific test case for it.

### DIFF
--- a/freqtrade/plugins/pairlistmanager.py
+++ b/freqtrade/plugins/pairlistmanager.py
@@ -2,7 +2,7 @@
 PairList manager class
 """
 import logging
-from typing import Dict, List
+from typing import Dict, List, Set
 
 from cachetools import TTLCache, cached
 
@@ -23,7 +23,7 @@ class PairListManager():
         self._config = config
         self._whitelist = self._config['exchange'].get('pair_whitelist')
         self._blacklist = self._config['exchange'].get('pair_blacklist', [])
-        self._logged_blacklist_pairs = set()
+        self._logged_blacklist_pairs: Set[str] = set()
         self._pairlist_handlers: List[IPairList] = []
         self._tickers_needed = False
         for pairlist_handler_config in self._config.get('pairlists', None):
--- a/freqtrade/plugins/pairlistmanager.py
+++ b/freqtrade/plugins/pairlistmanager.py
@@ -42,7 +42,7 @@ class PairListManager(LoggingMixin):
         if not self._pairlist_handlers:
             raise OperationalException("No Pairlist Handlers defined")
 
-        refresh_period = config.get('refresh_period', 1800)
+        refresh_period = config.get('pairlist_refresh_period', 3600)
         LoggingMixin.__init__(self, logger, refresh_period)
 
     @property
--- a/freqtrade/plugins/pairlistmanager.py
+++ b/freqtrade/plugins/pairlistmanager.py
@@ -2,12 +2,14 @@
 PairList manager class
 """
 import logging
-from typing import Dict, List, Set
+from functools import partial
+from typing import Dict, List
 
 from cachetools import TTLCache, cached
 
 from freqtrade.constants import ListPairsWithTimeframes
 from freqtrade.exceptions import OperationalException
+from freqtrade.mixins import LoggingMixin
 from freqtrade.plugins.pairlist.IPairList import IPairList
 from freqtrade.plugins.pairlist.pairlist_helpers import expand_pairlist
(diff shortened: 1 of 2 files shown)


freqtrade/plugins/pairlistmanager.py
```diff
@@ -108,9 +108,11 @@ def verify_blacklist(self, pairlist: List[str], logmethod) -> List[str]:
         except ValueError as err:
             logger.error(f"Pair blacklist contains an invalid Wildcard: {err}")
             return []
-        for pair in deepcopy(pairlist):
+        for pair in pairlist.copy():
             if pair in blacklist:
-                logmethod(f"Pair {pair} in your blacklist. Removing it from whitelist...")
+                if pair not in self._logged_blacklist_pairs:
+                    logmethod(f"Pair {pair} in your blacklist. Removing it from whitelist...")
+                    self._logged_blacklist_pairs.add(pair)
```
xmatthias: I am not sure i like this kind of tracking / log surpression, considering freqtrade is potentially started, and then potentially "left running" for weeks / months at a time.

This can lead to a situation where you're running freqtrade for 2 weeks - and start wondering why a pair is not in your pairlist - since the logs "don't contain anything" (you'll not look through 2 weeks of logs - but only through the last half a day or so).


you should instead look at the `LoggingMixin` instead - which is used in the pairlist providers themselvs. 
it will reduce the amount of log messages, but will still log it from time to time - which i think is a better approach, and is aligned to how this works for all other pairlists, without reinventing the wheel.
this can potentially even collide with how VolumePairlist uses logging (see below).

`VolumePairlist` already implements blacklist verification logging this way.

https://github.com/freqtrade/freqtrade/blob/dda302eea28d7e4d3bad3615627554d9a76975ad/freqtrade/plugins/pairlist/VolumePairList.py#L207

xmatthias (changes requested):

freqtrade/plugins/pairlistmanager.py
cdimauro: Updated as per your request / recommendation. The tests passed as well (no changes required).

cdimauro: Strange that tests passed for MacOS X and Windows, but not on Ubuntu.
I've run pytest on Ubuntu 20.04 (latest update) on Windows/WSL2, and it worked perfectly...

xmatthias: pytest in CI runs with randomized test-sequence - so it's not a systematic failure on ubuntu - but with this particular seed.

xmatthias (approved):
LGTM :+1: 

Nice one - thanks a lot! :100:

Issue #4893, opened 2021-05-07
PerformanceFilter improve ordering logic


severe-hed:
## Describe the enhancement

I've configured a new bot with static pair list and performance filter
It's totally new, so no trades yet (so no information about performance)
But when checking /whitelist command in telegram I can see that order is not the same like in my static whitelist
Sometimes user wants to presetup order that will be changed on performance
My suggestion is to make order like this:
1. Positive performance sorted
2. Zeroes based on static list
3. Negative sorted

Bonus suggestion here to set presorted static list as a parameter to PerformanceFilter and maybe some params how to use it (as described above/priority to presorted static/etc)

xmatthias:
Feel free to submit a pull request changing this order.

You should however not go on "based on static list" - but based on the input to the filtering list - as it must be independent to using static or volume pairlists.

PR #6076, opened 2021-12-16
Attempt to improve the sorting logic of PerformanceFilter

ivanproman:
## Summary

The goal of this PR is to improve the sorting logic of PerformanceFilter.

Solve the issue: #4893 

## Quick changelog

freqtrade/plugins/paitlist/PerformanceFilter.py
tests/plugins/test_pairlist.py
docs/includes/pairlists.md

## What's new?

The update removes the alphabetic sorting of a pairlist when using a PerformanceFilter. This means that lists without any trades will remain pre-sorted, as entered by the user. test_performance_filter is modified to pass for the new sorting behavior, creating a regression. The relevant part of the .md has been updated to reflect the new sorting order logic as well.


xmatthias (changes requested):
Your commit is unfortunately wrong - commiting `.DS_Store` files - but not the real source files.

I'd recommend to change this (remove these files, add the correct files) - and use `git commit --amend` - followed by `git push -f` to fix the existing commit - as this is not something we'd even want in the git history if we can avoid it...

xmatthias:
@ ivanproman, i'm going to close this as it does not contain the changes you describe (it only adds a few "macos specific" files.

i would appreciate if you would reopen a new PR changing the correct files.

PR #6079, opened 2021-12-18
Initial position adjustment support (DCA)

xataxxx:
## Summary

Implement position adjustment / stacking / multiple buys of the same pair within a single trade.

Partly solve the issue: #2183 and #1519

## Quick changelog

- Added callback `adjust_trade_position` to strategy that can indicate if should buy more and how much.
- Added `position_adjustment_enable` strategy parameter to enable this feature.
- Updated documentation to document the usage of said features.

## What's new?

Implement ability for strategy to buy more of an already opened trade. 
The most common scenario for this would be DCA (Dollar Cost Averaging).
The theory is that your buys signals, while good, sometimes still produce trades that first go down X% and then rise to the profit. With DCA you can buy more of the same coin when it's gone down Y% to increase your stake and lower your average open price. This results in higher probability to reach profit AND faster trade times. It can also produce bigger and more hurtful stoplosses.

Done list:
- Strategy callbacks to implement DCA logic
- Additional buy logic to allow DCA
- Trade open price averaging and total amount summing works
- Limit orders works but they follow the same timeout rules as the original buy
- Stoploss on exchange is canceled and new one is created.
- Backtesting seems to work (HyperOpting also)
- Documentation is there
- Unit tests seem to cover basic DCA functionality without breaking other tests

TODO List:
- FreqUI part is not updated yet, which means if you do buy a pair more than once with a big enough price difference, your initial buy open price is displayed way outside of the candle. Best course of action is to start showing the individual filled orders.
- Negative adjustment / Partial sell - Not planning on implementing that currently and documentation and comments in code also indicate that we support buy orders not sell orders. A warning is thrown into log when negative price adjustment is attempted.


docs/strategy-advanced.md
```diff
@@ -229,3 +229,79 @@ for val in self.buy_ema_short.range:
...
+
+### Adjust trade position
+
+`adjust_trade_position()` can be used to perform additional orders to manage risk with DCA (Dollar Cost Averaging) for example.
+
+!!! Tip: The `position_adjustment_enable` configuration parameter must be enabled to use adjust_trade_position callback in strategy.
+
+!!! Warning: Additional orders also mean additional fees.
+
+!!! Warning: Stoploss is still calculated from the initial opening price, not averaged price.
```
xmatthias: Your syntax for these boxes is off / wrong (not just here, in other documentation changes too)

The correct way is the following - with "title" being optional.
you can render the docs with `mkdocs serve` - and have a look at it offline to see if it renders the way you expect it to.
```
!!! <type> "Title"
    <text, indented by 4 spaces>
```

freqtrade/freqtradebot.py
```diff
@@ -443,6 +450,52 @@ def create_trade(self, pair: str) -> bool:
...
+        Check the implemented trading strategy for adjustment command.
+        If the strategy triggers the adjustment, a new order gets issued.
+        Once that completes, the existing trade is modified to match new data.
+        """
+        for order in trade.orders:
+            if order.ft_is_open:
+                return
+
+        logger.debug(f"adjust_trade_position for pair {trade.pair}")
+        sell_rate = self.exchange.get_rate(trade.pair, refresh=True, side="sell")
```
xmatthias: I think side should be "buy" here (or maybe both ...) ? 
i think `current_rate` below should really be split into "buy" and "sell" rates - each calculated based on their respective logic.

As the callback - `adjust_trade_position` - can return both positive (for increasing) and negative (for decreasing) values - you'll probably need both rates to make a good decision...

freqtrade/persistence/models.py
```diff
@@ -568,6 +569,37 @@ def calc_profit_ratio(self, rate: Optional[float] = None,
...
+            # Just in case, still recalc open trade value
+            self.recalc_open_trade_value()
+            return
+
+        total_amount = 0.0
+        total_stake = 0.0
+        for temp_order in self.orders:
+            if temp_order.ft_is_open or \
+                    temp_order.ft_order_side != 'buy' or \
+                    temp_order.status not in NON_OPEN_EXCHANGE_STATES:
```
xmatthias: Personally, i like wrapping lines in paranteses better than splitting with `\` (which you'll observe, we're doing almost nowhere).

freqtrade/freqtradebot.py
xataxxx: I've converted this to buy side as that does make more sense overall. Having both seems wrong at this point as supporting negative adjustments / selling is currently way too far future and we have **kwargs if we ever do want to add those. Also the current naming and logic follows all the other callbacks where current_rate and current_profit indicate buying side price and profit.

freqtrade/freqtradebot.py
```diff
@@ -285,8 +292,8 @@ def update_closed_trades_without_assigned_fees(self):
         for trade in trades:
             if trade.is_open and not trade.fee_updated('buy'):
                 order = trade.select_order('buy', False)
-                if order:
-                    logger.info(f"Updating buy-fee on trade {trade} for order {order.order_id}.")
+                open_order = trade.select_order('buy', True)
+                if order and open_order is None:
```
xmatthias, line 295: i think we'd want to keep the log message there
The check itself seems find i guess - but would we need the same logic for "scaling out" (-> for the sell logic)?

xataxxx: I added the log statement back (removed by accident).
The sell logic is checking over already closed trades in which case there would be no changes needed as there shouldn't be any open orders for them.

docs/configuration.md
```diff
@@ -584,6 +594,15 @@ export HTTPS_PROXY="http://addr:port"
...
 ```
 
+### Understand position_adjustment_enable
+
+The `position_adjustment_enable` configuration parameter enables the usage of `adjust_trade_position()` callback in strategy.
+For performance reasons, it's disabled by default, and freqtrade will show a warning message on startup if enabled.
+Enabling this does nothing unless the strategy also implements `adjust_trade_position()` callback.
+
+See [the strategy callbacks](strategy-callbacks.md) for details on usage.
+
```
xmatthias: I'd omit this whole section, and instead link to the "strategy-callbacks" page directly from the configuration list (you can use `strategy-callbacks.md#title` to link directly to the relevant section).
 
these "understanding ..." sections are mainly for points that are not explained anywhere else - this seems to just duplicate the explanation from the callbacks

freqtrade/wallets.py
```diff
@@ -184,6 +184,17 @@ def _calculate_unlimited_stake_amount(self, available_amount: float,
...
+
+        # Position Adjustment dynamic base order size
+        try:
+            if self._config.get('position_adjustment_enable', False):
+                base_stake_amount_ratio = self._config.get('base_stake_amount_ratio', 1.0)
+                if base_stake_amount_ratio > 0.0:
+                    possible_stake = possible_stake * base_stake_amount_ratio
+        except Exception as e:
+            logger.warning("Invalid base_stake_amount_ratio", e)
+            return 0
```
xmatthias: Do we really need this as a parameter?

considering we have `custom_stake_amount` to customize stake-amount - we could also just defer this logic to the strategy (it get's the "proposed stake" - so you can just do `proposed_stake / max_entries` in the strategy.

Also, assuming your "scaling in" depends on the situation / pair, and will be anywhere between 1 and 10 trades - you'd never be able to (correctly) capture this with a parameter anyway

docs/configuration.md
```diff
@@ -172,6 +173,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `user_data_dir` | Directory containing user data. <br> *Defaults to `./user_data/`*. <br> **Datatype:** String
 | `dataformat_ohlcv` | Data format to use to store historical candle (OHLCV) data. <br> *Defaults to `json`*. <br> **Datatype:** String
 | `dataformat_trades` | Data format to use to store historical trades data. <br> *Defaults to `jsongz`*. <br> **Datatype:** String
+| `position_adjustment_enable` | Enables the strategy to use position adjustments (additional buys or sells). More information below. <br> **Datatype:** Boolean
```
xmatthias: I think this parameter should become a strategy parameter (set / settable in the strategy).

It's almost exclusively dependent on the strategy, not on the configuration - forcing people to adjust this in the config is not a good approach if you ask me.

This will be as simple as 
* adding it to the below list
* adding it to strategy_resolver (next to all the other parameters)
* adding a default value in the strategy.

freqtrade/optimize/backtesting.py
```diff
@@ -350,8 +351,70 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
+        try:
+            min_stake_amount = self.exchange.get_min_pair_stake_amount(
+                trade.pair, propose_rate, -0.05) or 0
+            stake_amount = self.wallets.validate_stake_amount(trade.pair,
+                                                              stake_amount, min_stake_amount)
+            stake_amount = self.wallets._check_available_stake_amount(stake_amount,
+                                                                      available_amount)
+        except DependencyException:
+            logger.debug(f"{trade.pair} adjustment failed, "
+                         f"wallet is smaller than asked stake {stake_amount}")
```
xmatthias: For debug logging (especially in backtesting) - please use the "old" syntax (%s - with parameters as separate parameters passed to the log function).

The reason for this is simple - debug logging is usually disabled. This will prevent evaluation of the F-string which is then never used, and therefore provides a (very slight) speed boost.

freqtrade/optimize/backtesting.py
```diff
@@ -350,8 +351,70 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
+            order_type="market",
+            status="closed",
+            price=propose_rate,
+            average=propose_rate,
+            amount=amount,
+            cost=stake_amount
+        )
+        trade.orders.append(buy_order)
+        trade.recalc_trade_from_orders()
+        self.wallets.update()
```
xmatthias: I'm currently hesitant / undecided if this is needed or not.

i don't THINK it is (the main method is `get_trade_stake_amount` - which has this builtin). Please look at the other comments for this file first - it may "resolve" itself automatically...

freqtrade/optimize/backtesting.py
```diff
@@ -350,8 +351,70 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
+        current_profit = trade.calc_profit_ratio(row[OPEN_IDX])
+
+        stake_amount = strategy_safe_wrapper(self.strategy.adjust_trade_position,
+                                             default_retval=None)(
+            pair=trade.pair, trade=trade, current_time=row[DATE_IDX].to_pydatetime(),
+            current_rate=row[OPEN_IDX], current_profit=current_profit)
+
+        # Check if we should increase our position
+        if stake_amount is not None and stake_amount > 0.0:
+            return self._execute_trade_position_change(trade, row, stake_amount)
```
xmatthias: You'll need to be careful here, as this is not properly emulating live, potentially causing quite different results.

In live, `execute_entry()` will be called to increase a position.
That method calls several strategy callbacks itself, namely
* `custom_entry_price`
* `custom_stake_amount`
* `confirm_trade_entry`

for both new trades, as well as for DCA "increases".
To be clear - i think that's correct behavior (although the `custom_stake_amount` might be unnecessary) - as it allows the user to determine a buy price - as well as "back out" last second (once you see the price you'd get, for example).

In backtesting, we're calling the 3 above mentioned callbacks too - but ONLY for the first trade entry - not for any subsequent DCA orders. Depending on the logic in these methods (maybe there is none) - this can become a problem as you'll no longer be able to properly backtest the proper workings of these methods in DCA situations.

freqtrade/freqtradebot.py
```diff
@@ -443,6 +451,53 @@ def create_trade(self, pair: str) -> bool:
...
+        """
+        # Walk through each pair and check if it needs changes
+        for trade in Trade.get_open_trades():
+            try:
+                self.adjust_trade_position(trade)
+            except DependencyException as exception:
+                logger.warning('Unable to adjust position of trade for %s: %s',
+                               trade.pair, exception)
+
+    def adjust_trade_position(self, trade: Trade):
```
xmatthias: Please change the name of this function to "something" else.
Having the name identical to the strategy callback is confusing - and can potentially lead to odd, accidental bugs.

tests/optimize/test_hyperopt.py
```diff
@@ -189,6 +191,8 @@ def test_start_no_data(mocker, hyperopt_conf) -> None:
     with pytest.raises(OperationalException, match='No data found. Terminating.'):
         start_hyperopt(pargs)
 
+    os.unlink(Hyperopt.get_lock_filename(hyperopt_conf))
```
xmatthias: How's the changes in this file related to this PR? 

if it's not (and just fixing some random test failure) - please create a separate PR for that ...

freqtrade/freqtradebot.py
```diff
@@ -560,57 +600,111 @@ def execute_entry(self, pair: str, stake_amount: float, price: Optional[float] =
...
+                                                                   self.strategy.stoploss)
+        if not self.edge and trade is None:
+            max_stake_amount = self.wallets.get_available_stake_amount()
+            stake_amount = strategy_safe_wrapper(self.strategy.custom_stake_amount,
+                                                 default_retval=stake_amount)(
+                pair=pair, current_time=datetime.now(timezone.utc),
+                current_rate=enter_limit_requested, proposed_stake=stake_amount,
+                min_stake=min_stake_amount, max_stake=max_stake_amount)
+        stake_amount = self.wallets.validate_stake_amount(pair, stake_amount, min_stake_amount)
+        return enter_limit_requested, stake_amount
```
xmatthias: In my understanding, these 2 methods are extracted basically "as is" - without much modification?

I think it would make sense / make review easier if we could extract this part into a separate PR (to be merged first, obviously) ... that'd keep the focus of the changes for this PR to what's relevant to DCA - not to "random necessary refactorings" - if that makes sense

docs/strategy-advanced.md
```diff
@@ -229,3 +229,83 @@ for val in self.buy_ema_short.range:
...
+        last_candle = dataframe.iloc[-1].squeeze()
+        previous_candle = dataframe.iloc[-2].squeeze()
+        if last_candle['close'] < previous_candle['close']:
+            return None
+
+        count_of_buys = 0
+        for order in trade.orders:
+            # Instantly stop when there's an open order
+            if order.ft_is_open:
+                return None
```
xmatthias: I don't think we would be calling this callback if there is an open order.

Having this segment in the docs suggests that this is called every iteration, independent if there's open orders or not - which is not true.
The alternative would be to allow multiple open buy orders (some form of grid-orders i guess) - which i don't see handled anywhere - and i don't think is the purpose  / scope of this PR, and would be a little more complex, as then you'd have to remove `open_order_id` and instead check via the orders list - as there could be multiple orders at any moment.

freqtrade/freqtradebot.py
```diff
@@ -443,6 +451,53 @@ def create_trade(self, pair: str) -> bool:
...
+            self.execute_entry(trade.pair, stake_amount, trade=trade)
+
+        if stake_amount is not None and stake_amount < 0.0:
+            # We should decrease our position
+            # TODO: Selling part of the trade not implemented yet.
+            logger.error(f"Unable to decrease trade position / sell partially"
+                         f" for pair {trade.pair}, feature not implemented.")
+            return
+
+        return
```
xmatthias: These 2 statements are not really necessary (python will return None implicitly if there's no return statement present).

freqtrade/optimize/backtesting.py
```diff
@@ -476,6 +539,21 @@ def _enter_trade(self, pair: str, row: List) -> Optional[LocalTrade]:
                 buy_tag=row[BUY_TAG_IDX] if has_buy_tag else None,
                 exchange='backtesting',
             )
+            order = Order(
```
xmatthias, line 522: i do have "a few" performance concerns about the usage of the `Order` object directly in backtesting as it's an sqlalchemy object (using an attribute there has a callstack as deep as from here to the Moon and back - which has quite severe performance implications).

For the moment, keep it as is - but we will need to consider changing this to a "LocalOrder" object eventually (so this is also a reminder for myself).

freqtrade/optimize/backtesting.py
```diff
@@ -476,6 +539,21 @@ def _enter_trade(self, pair: str, row: List) -> Optional[LocalTrade]:
...
+                ft_order_side="buy",
+                side="buy",
+                order_type="market",
+                status="closed",
+                price=trade.open_rate,
+                average=trade.open_rate,
+                amount=trade.amount,
+                cost=trade.stake_amount + trade.fee_open
+            )
+            trade.orders = []
```
xmatthias: I don't think this should be necessary - `orders` is initialized as a list in the Object already (unless you encountered some problem with that?)

xmatthias (commented):
Looks like many comments - but for a change this "functionality heavy" - i think this is at a quite mature state already :+1:

freqtrade/optimize/backtesting.py
xataxxx: I did refactor a LocalOrder and measured the performance difference and there was exactly less than a second difference for backtesting 120 pairs with 2 months of data. I'm sure there might be 'some' benefit to it but I reverted it as that would've caused too many changes for unnoticeable benefit.

freqtrade/optimize/backtesting.py
xataxxx: It was needed because for some reason trade objects were otherwise using the same list (backtesting added orders from multiple pairs into the same list). Setting it into a new one resolved it.

xmatthias: i've kindof suspected this ... 
in that case i think this should be set as part of the constructor (`__init__` ) of the `LocalTrade` object.

freqtrade/wallets.py
xataxxx: That is a great idea! We really don't need this and it really would be more beneficial to calculate it in custom_stake_amount! Removing it with next commit.

docs/configuration.md
xataxxx: A bit too much duplication indeed. I've moved most of the documentation into strategy-advanced instead and linked to it from callbacks. Will come with next commit.

docs/configuration.md
xataxxx: That's a great idea. This keeps this whole thing away from config file at all and all tight and neat inside the strategy.

freqtrade/freqtradebot.py
xataxxx: They were extracted almost "as is" for reusability and because the original methods complexity was over the threshold value causing alarms. The get_valid_enter_price_and_stake also has an if for position adjustment in it by checking if trade is None or not (custom_stake_amount is not called for edge and pos. adjustment).

docs/strategy-advanced.md
xataxxx: Fair point and we really shouldn't even call the method that calls the callback if there's a trade open. Code adjusted in the next commit.

freqtrade/optimize/backtesting.py
xataxxx: Reworked it all into _enter_trade, so no more code duplication. recalc is called only for pos_adjust and wallets update can be skipped here indeed

freqtrade/optimize/backtesting.py
xataxxx: Yes, moved the orders=[] into construct of trade

freqtrade/optimize/backtesting.py
xataxxx: I've reworked that into _enter_trade to reduce code duplication and make sure we are calling the callbacks correctly.
custom_entry_price is called, custom_stake_amount and confirm_trade_entry are skipped for both bot and backtest.
custom_stake_amount because adjust_trade_position already returns stakes and confirm_trade_entry because it already send the stake only if it wants the trade, so essentially confirming it at the same time. Also adjust_trade_position has the actually pre-existing trade / previous orders available so it has more data to make a qualified decision.

freqtrade/freqtradebot.py
xataxxx: That name has been changed to `check_and_call_adjust_trade_position`

tests/optimize/test_hyperopt.py
xataxxx: Yes, my bad.

docs/configuration.md
```diff
@@ -172,6 +172,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `user_data_dir` | Directory containing user data. <br> *Defaults to `./user_data/`*. <br> **Datatype:** String
 | `dataformat_ohlcv` | Data format to use to store historical candle (OHLCV) data. <br> *Defaults to `json`*. <br> **Datatype:** String
 | `dataformat_trades` | Data format to use to store historical trades data. <br> *Defaults to `jsongz`*. <br> **Datatype:** String
+| `position_adjustment_enable` | Enables the strategy to use position adjustments (additional buys or sells). [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `true`.*<br> **Datatype:** Boolean
```
xmatthias:
```suggestion
| `position_adjustment_enable` | Enables the strategy to use position adjustments (additional buys or sells). [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*<br> **Datatype:** Boolean
```

docs/configuration.md
```diff
@@ -302,6 +304,14 @@ To allow the bot to trade all the available `stake_currency` in your account (mi
...
     It is therefore important to set `dry_run_wallet` to a sensible value (like 0.05 or 0.01 for BTC and 1000 or 100 for USDT, for example), otherwise, it may simulate trades with 100 BTC (or more) or 0.05 USDT (or less) at once - which may not correspond to your real available balance or is less than the exchange minimal limit for the order amount for the stake currency.
 
+#### Dynamic stake amount with position adjustment
+When you want to use position adjustment with unlimited stakes, you must also set `base_stake_amount_ratio` to a reasonable value depending on your strategy.
+Typical value would be in the range of 0.1 - 0.5, but depends highly on your strategy and how much you wish to leave into the wallet as position adjustment buffer.
+<br>
+For example if your position adjustment assumes it can do 2 additional buys with the same stake amounts then your `base_stake_amount_ratio` should be 0.333333 (33.3333%) of your calculated unlimited stake amount.
+<br>
+Or another example if your position adjustment assumes it can do 1 additional buy with 3x the original stake amount then your `base_stake_amount_ratio` should be 0.25 (25%) of your calculated unlimited stake amount.
+
```
xmatthias: is this section still valid? i don't see `base_stake_amount_ratio` used in the code anymore

xataxxx: Text adjusted to suggest usage of `custom_stake_amount`

docs/configuration.md
xataxxx: Agreed, should be false by default.

freqtrade/persistence/models.py
```diff
@@ -428,6 +428,7 @@ def update(self, order: Dict) -> None:
             if self.is_open:
```
xmatthias, line 427: The above call to `recalc_open_trade_value()` might no longer be necessary.
It's called from within `recalc_trade_from_orders()` in all instances as far as i can tell.

freqtrade/freqtradebot.py
```diff
@@ -443,6 +448,49 @@ def create_trade(self, pair: str) -> bool:
...
+    def process_open_trade_positions(self):
+        """
+        Tries to execute additional buy or sell orders for open trades (positions)
+        """
+        # Walk through each pair and check if it needs changes
+        for trade in Trade.get_open_trades():
+            # If there is any open orders, wait for them to finish.
+            for order in trade.orders:
+                if order.ft_is_open:
+                    break
```
xmatthias: What's your intend here exactly? 

the comment suggests that you'd want to actually skip processing for this _trade_ (not this _order_) - but break will interrupt the Loop in line 461, not the one on 459.

freqtrade/freqtradebot.py
```diff
@@ -443,6 +448,49 @@ def create_trade(self, pair: str) -> bool:
...
+                logger.warning('Unable to adjust position of trade for %s: %s',
+                               trade.pair, exception)
+
+    def check_and_call_adjust_trade_position(self, trade: Trade):
+        """
+        Check the implemented trading strategy for adjustment command.
+        If the strategy triggers the adjustment, a new order gets issued.
+        Once that completes, the existing trade is modified to match new data.
+        """
+        current_rate = self.exchange.get_rate(trade.pair, refresh=True, side="buy")
```
xmatthias, line 476: This still leaves me somewhat doubful.

Assuming a lowish value pair (-> a high spread) - we could see the following (this assumes reversed price sides):
Telegram/api reports a profit of -5% (assuming sell rate)
`adjust_trade_position()` will be called with a profit of -4.5% (or even 4%)

Now obviously, for buying again, we'd want to use the "buy" side price
However for determining the current profit to decide on DCA, i'm not entirely sure which price you'd want to look at - and it'll strongly depend on the pricing configuration which one will actually make sense.

if we don't wanna have this "fixed" to either buy or sell - maybe we can leave the user in the driver seat and have a "pricing_side_DCA" (naming might be improved) - which can default to buy (or sell) - but gives the user/strategy the last word in it?

tests/test_freqtradebot.py
```diff
@@ -4308,3 +4309,244 @@ def test_get_valid_price(mocker, default_conf_usdt) -> None:
 
     assert valid_price_at_min_alwd > custom_price_under_min_alwd
     assert valid_price_at_min_alwd < proposed_price
+
+
+def test_position_adjust(mocker, default_conf_usdt, fee) -> None:
```
xmatthias, line 4314: This test is actually not holding what it promises (so i was disappointed at the end :D  )

Naming suggests that it'll test position adjustment - but it actually tests  order filling in case of additional orders - not the adjustment logic itself (neither `process_open_trade_positions()` - nor `check_and_call_adjust_trade_position` are actually called).


what i'd do is the following (maybe in `test_integration.py`)?
- try to mock as little as possible (basically only pricing and "is order filled" - and what signals we need at this point) - stick to dry-run orders (the logic will not change if it's dry or live)
- try to run the "real" logic - basically calling `.process()` - have it enter a trade first - followed by a "DCA order" (you'll have to have the strategy return the correct things).

freqtrade/freqtradebot.py
```diff
@@ -560,57 +593,111 @@ def execute_entry(self, pair: str, stake_amount: float, price: Optional[float] =
...
+                exchange=self.exchange.id,
+                open_order_id=order_id,
+                strategy=self.strategy.get_strategy_name(),
+                buy_tag=buy_tag,
+                timeframe=timeframe_to_minutes(self.config['timeframe'])
+            )
 
+        trade.is_open = True
+        trade.fee_open_currency = None
+        trade.open_order_id = order_id
```
xmatthias: it's not "really" relevant for performance if they are reassigned in case of a new entry - but it's potentially confusing, so i'd either move this to the else branch - or add a comment that this is here for "position adjust" purposes.

freqtrade/freqtradebot.py
```diff
@@ -560,57 +593,111 @@ def execute_entry(self, pair: str, stake_amount: float, price: Optional[float] =
...
         # Updating wallets
         self.wallets.update()
 
-        self._notify_enter(trade, order_type)
+        self._notify_enter(trade, order, order_type)
+
+        if pos_adjust:
+            if order_status == 'closed':
+                logger.info(f"DCA order closed, trade should be up to date: {trade}")
+                trade = self.cancel_stoploss_on_exchange(trade)
```
xmatthias, line 643: This is fine if the order is immediately closed

But i struggle to find where the stoploss is cancelled if the order is not filled right away - so i don't think that's happening at the moment.

freqtrade/optimize/backtesting.py
```diff
@@ -350,8 +350,35 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
         else:
             return sell_row[OPEN_IDX]
 
+    def _get_adjust_trade_entry_for_candle(self, trade: LocalTrade, row: Tuple
+                                           ) -> LocalTrade:
+
+        # If there is any open orders, wait for them to finish.
+        for order in trade.orders:
+            if order.ft_is_open:
+                return trade
```
xmatthias: Backtesting does not have logic to keep an order open (orders are assumed to always fill immediately).

As such, i think it's save to omit this whole block

freqtrade/optimize/backtesting.py
```diff
@@ -350,8 +350,35 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
+        stake_amount = strategy_safe_wrapper(self.strategy.adjust_trade_position,
+                                             default_retval=None)(
+            pair=trade.pair, trade=trade, current_time=row[DATE_IDX].to_pydatetime(),
+            current_rate=row[OPEN_IDX], current_profit=current_profit)
+
+        # Check if we should increase our position
+        if stake_amount is not None and stake_amount > 0.0:
+            pos_trade = self._enter_trade(trade.pair, row, stake_amount, trade)
+            if pos_trade is not None:
+                return pos_trade
```
xmatthias, line 372: I think you can omit these 2 lines here by reassigning `trade` (which would not even be necessary - but i'd keep for clarity, as non-primitives (int, float, ...) are passed by reference - so modifications to the trade object from within the method will always affect the original trade object, too.)

docs/strategy-advanced.md
```diff
@@ -229,3 +229,99 @@ for val in self.buy_ema_short.range:
 # Append columns to existing dataframe
 merged_frame = pd.concat(frames, axis=1)
 ```
+
+## Adjust trade position
```
xmatthias: I would actually move this whole section to `strategy-callbacks.md`.

Currently `configuration.md` links to the callback page - which forwards you to the advanced docs for "details" - which is quite odd - and kindof unnecessary.

It's for sure an advanced topic (maybe add a warning box about that - or that it's still in beta testing?) - but so are other topics covered by callbacks (like selecting a custom price, depending on the logic you implement).

xmatthias:
I've added 2 commits which do some minor cleanup (explaining what i think is more work than adding that myself).

Overall this looks pretty good :+1:

freqtrade/optimize/backtesting.py
xataxxx: It actually initially was a single line, but _enter_trade is capable of returning None in case there's not enough wallet etc and as such its return value is Optional[Trade] and this brings mypy warnings.

freqtrade/optimize/backtesting.py
xataxxx: Fixed in next commit.

freqtrade/freqtradebot.py
xataxxx: Adjusted in next commit.

freqtrade/freqtradebot.py
xataxxx: Purpose is that break stops the processing of the orders loop AND this current trade and the outer for cycle continues with the next trade.

So if there is an open order, the trade `check_and_call_adjust_trade_position` is skipped, but it's checked for all trades.

freqtrade/persistence/models.py
xataxxx: Yes, it's actually no longer needed here. Originally left it in place just-in-case. Removed in next commit.

freqtrade/freqtradebot.py
xataxxx: You are correct. That's a bug. I'll attempt to fix it.

freqtrade/freqtradebot.py
xmatthias: Problem is - break will only interrupt the inner for (`for order in trade.orders:`) - but the below `try` block  will execute, independent of the fact that you hit a break or not.

In theory, you COULD implement that with a `for/else` statement (the else will execute only if no break was encountered) - but that's **very** niche and I've actually never seen that "used in the wild".

rokups:
I started testing this branch. Awesome work!

I noticed `self.wallets.get_trade_stake_amount(pair)` may throw an exception when wallet does not contain enough funds. That is kind of weird... It is just a getter, it should not do hard assumptions like this. As a reason you had to use try/except in `adjust_trade_position`. I think this bit should be reworked to not throw and this check that caused exception should be moved out to places that are using result of `get_trade_stake_amount()`. What do you think @xmatthias?

xmatthias:
`get_trade_stake_amount()` is not considered a public method (to be used in a strategy) at the moment (not documented) - so throwing the exception is fine in my eyes (it keeps the logic in one place).

I'm not even sure if the exception handler in the strategy is really necessary.
by using `strategy_safe_wrapper` - you can let the callback bubble up the exception without unexpected side-effects (it'll handle all exceptions and log these). 
All the exception does is tell you "not enough stake - can't do anything".
Current documentation (in this PR) explicitly captures this and returns None - but all "letting it bubble up" will do is log it as exception that was raised.

Now obviously, if we assume that's going to change (it's becoming a public method) - and we do not want it to raise an exception - i'd instead write a wrapper around `get_trade_stake_amount()` - which will return something "else" (probably 0 or None) (and is not used internally). 

Depending on how you use it, you'll still have to account for each of these cases (no math with None, no division with 0) - so i'm not 100% sure that'll help you very much - as all that changes is the location where you have to catch which exception (and if you do it with an if or try/except block).

rokups:
Well it is used in example in docs so that effectively elevates it to a public method. And exception handler is necessary. i tried to remove it and found out that it actually triggers easily.

xmatthias:
> Well it is used in example in docs so that effectively elevates it to a public method. And exception handler is necessary. i tried to remove it and found out that it actually triggers easily.

As said, the alternative (None / 0) would simply raise different exceptions in the strategy code (especially None) - so i'm not certain it'll create much benefit - as you'll simply have to handle different errors.

rokups:
Actually i would go as far as to suggest `get_trade_stake_amount()` is not quite fit to use for DCA. Take original code as an example:
> Total stake for this trade would be 1 + 1.25 + 1.5 + 1.75 = 5.5x of the initial allowed stake.

Suppose we use a static stake of 1000 USD. As example code is now, size of each trade with 1000 USD static stake is actually 1000 * 5.5, which is clearly wrong. We would like to DCA into a total position of 1000 USD.

Only `custom_stake_amount` needs fixing. `adjust_trade_position` already does a right thing (failing `wallets.get_trade_stake_amount` aside).
```py
    # Let unlimited stakes leave funds open for DCA orders
    def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: float,
                            proposed_stake: float, min_stake: float, max_stake: float,
                            **kwargs) -> float:
        return proposed_stake / self.max_dca_multiplier
```

And later on we would DCA into 1000 USD like so:
0. 1000 / 5.5 * 1 = 181.81 (initial stake)
1. 1000 / 5.5 * 1.25 = 227.27
2. 1000 / 5.5 * 1.5 = 272.72
3. 1000 / 5.5 * 1.75 = 318.18

Now consider a situation where we have a wallet of say 1010 USD - barely enough to enter one trade. Placing initial order will succeed because `self.wallets.get_trade_stake_amount(pair)` will not throw an exception as there is more than 1000 USD in the wallet. Subsequent attempts to DCA into position will fail though, because `self.wallets.get_trade_stake_amount(pair, None)` still wants at least 1000 USD in the wallet even though strategy does not need that much.

Maybe `Wallet._check_available_stake_amount` could be updated to take into account funds already tied into in a trade. As it is now, i am mostly copying code of `wallets.get_trade_stake_amount` into my strategy instead of using that method. That is a sign of an issue.

---

Here is my take on DCA usage. It implements same logic as example in the docs (4 total orders, 1 + 3xDCA). I considered initial buy order in the total of DCA orders for simplicity. `get_stake_amount()` should correctly handle case i mentioned and also half-correctly interact with `amend_last_stake_amount`/`last_stake_amount_min_ratio`. As it is now, each buy order would have to adhere to limits of `last_stake_amount_min_ratio`. That may be incorrect as we may want to apply this limitation to the whole position.
```py
    dca_total_orders = 4
    dca_boost = 0.25

    def get_dca_amount(self, trade: Trade, max_orders, boost):
        count_of_buys = sum([1 for order in trade.orders if order.ft_order_side == 'buy' and order.status == 'closed'])
        return self.get_stake_amount(count_of_buys, max_orders, boost)

    def get_stake_amount(self, count_of_buys, max_orders, boost):
        if count_of_buys >= max_orders:
            return None
        order_total_ratio = sum([1 + i * boost for i in range(max_orders)])
        self.wallets.update()
        val_tied_up = Trade.total_open_trades_stakes()
        available_amount = self.wallets.get_available_stake_amount()
        stake_amount = self.config['stake_amount']
        if stake_amount == 'unlimited':
            stake_amount = self.wallets._calculate_unlimited_stake_amount(available_amount, val_tied_up)
        stake_amount = stake_amount / order_total_ratio * (1 + (boost * count_of_buys))
        try:
            stake_amount = self.wallets._check_available_stake_amount(stake_amount, available_amount)
        except DependencyException:
            return None
        return stake_amount

    def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: float,
                            proposed_stake: float, min_stake: float, max_stake: float,
                            **kwargs) -> float:
        return self.get_stake_amount(0, self.dca_total_orders, self.dca_boost)

    def adjust_trade_position(self, pair: str, trade: Trade,
                              current_time: datetime, current_rate: float, current_profit: float,
                              **kwargs) -> Optional[float]:
        if current_profit > -0.015:
            return None

        # Obtain pair dataframe.
        dataframe, _ = self.dp.get_analyzed_dataframe(pair, self.timeframe)

        # Only buy when not actively falling price.
        candle1 = dataframe.iloc[-1].squeeze()
        candle2 = dataframe.iloc[-2].squeeze()
        if candle1['rsi'] < candle2['rsi']:
            return None

        return self.get_dca_amount(trade, self.dca_total_orders, self.dca_boost)
```

xataxxx:
My initial example in the docs is probably too complex and should not use `get_trade_stake_amount` at all. I think we could get that information from the trade or the first order also.

The notion that when using a fixed stakes of 1000 USD the code example is somehow wrong depends on the point of view.
If you want to use fixed stakes then your configuration should contain the initial stake size (as it always has) not the total you could theoretically waste on a trade when using DCA (which totally depends on the strategy).

In my examples the unlimited stakes version correctly correctly divides the available trade amount (1000 in your example) with 5.5 to reach 181.81. If you'd like to use fixed stakes with that same example code then you should use 181.81 as your stake_amount in the configuration. 

Also note that the whole idea is that the strategy is in control of when and how much to buy or not to buy. The DCA functionality is provided by strategy. Freqtrade only provides the capability to adjust position / buy more than once.

rokups:
Seems to me `stake_amount` being initial stake size makes it hard to reason about trades. It is no longer obvious whats total size of a position. Also it makes it quite easy to get into situation where all `max_open_trades` can not be filled, because we DCA'ed into few trades more than wallet can bear.

However if `stake_amount` is total position size it is immediately obvious what bot is going to do, how big `dry_run_wallet` should be, how many `max_open_trades` we can afford. The less hidden details - the better, especially for less-experienced users. There also is an added benefit of being consistent with logic of `unlimited` stake, when wallet size matches.

xataxxx:
There is no "total size of a position" when you enable position adjusting, that is up for the strategy to decide and adjust. I don't think there should be a hard limit that limits the whole trade because the strategy should have the capability to direct ALL funds into a single trade.

I do understand your logic as well, but that would limit the strategy into dealing within it's predefined 1000 stake. 

But some people would like their initial orders to be for example 50% of their available funds and the other 50% is used for DCA when and where needed, meaning yes, that they may not have enough to fill all DCA orders for all trades or once they do some DCA orders they may not have enough money to open new trades even though they have an open slot. It's up to the user / strategy to decide how to divide that money. I personally would rather direct my money into saving my existing trade / DCA'ing than open new trades. 

I'll attempt to perhaps improve the documentation around position adjusting and add an explaining note that the stake_amount is being used as "proposed_stake" in `custom_stake_amount`, but that has always been true anyways so hopefully better documentation perhaps helps to sort this.

freqtrade/freqtradebot.py
xataxxx: No. Break breaks the inner loop AND stops the current trade from being processed further (breaks the current iteration for the outer loop), but the outer loop continues with the next trade.
```py
    trades = [1, 2, 3]
    orders = [1, 2, 3, 4, 5]
    for trade in trades:
        for order in orders:
            if trade == 2 and order == 3:
                break
            print(f"trade {trade}, order {order}")
```
results in
trade 1, order 1
trade 1, order 2
trade 1, order 3
trade 1, order 4
trade 1, order 5
trade 2, order 1
trade 2, order 2
trade 3, order 1
trade 3, order 2
trade 3, order 3
trade 3, order 4
trade 3, order 5

xmatthias:
I'm not sure the 2 approaches have to contradict each other.
both can be valid scenarios i think - and both will vary on the strategy, the user's desire - and the user's available capital/wallet size.

from a general perspective, i'd expect to still have existing limits somewhat respected, which means, if i have a 1000$ wallet with 5 max_open_trades - opening a DCA order which in the end holds 1000$ is clearly against the expectation.
On the other hand, DCA'ing into the order "directly" (assuming it'll take us X steps to reach the desired 200$ position size) might end up wrong too - as with a smaller starting capital - we'll scratch on the edge of "no longer tradable" very quickly.

If we however assume that "custom_stake_amount" is always required - it's the responsability of the strategy - freqtrade will technically not care if you open your first trade with "everything" - or with 20$.

freqtrade/freqtradebot.py
xmatthias: your print statement is indented "too far" (it's part of the inner loop).

``` python
trades = [1, 2, 3]
orders = [1, 2, 3, 4, 5]
for trade in trades:
    for order in orders:
        if trade == 2 and order == 3:
            break
    print(f"trade {trade}, order {order}")
```
outputs 
```
trade 1, order 5
trade 2, order 3  <-- This message should not be shown - but it is -> the trade is processed further even if the "order processing" stopped.
trade 3, order 5
```

freqtrade/freqtradebot.py
```diff
@@ -1275,36 +1368,35 @@ def update_trade_state(self, trade: Trade, order_id: str, action_order: Dict[str
             logger.warning('Unable to fetch order %s: %s', order_id, exception)
             return False
 
+        logger.info(f"Updating order from exchange: {order}")
```
xmatthias: we should talk about verbosity ...

this PR adds quite some "info level" messages.
that's OK where they are needed / provide useful information for the user - but for example this one is not in my opinion (it's showing a dict directly - which is kinda ugly, usually).

if we need the order from the exchange (what the exchange answered) - we can use `log_responses` in the exchange class ... 

_for some background_ : I'm fighting to reduce verbosity - to avoid having "log spams" (which we still do in some scenarios) - where the same event (like a buy) triggers 3-5 messages - which all "almost" say the same thing (we're buying). All are at a different stage - but don't "really" provide anything useful - unless something goes wrong (which will have a different message anyway).

freqtrade/freqtradebot.py
xataxxx: You are correct. I'll fix this.

docs/strategy-advanced.md
xataxxx: Moved the documentation all into callbacks page.

freqtrade/freqtradebot.py
xataxxx: Totally agree. Those logs were added for better monitoring of what's going on while investigating a bug that only happened in live, so needed other testers to be able to copy some logs to me without enabling the full debug verbosity. It's resolved now and we don't need so many logs at the moment.

freqtrade/freqtradebot.py
xataxxx: Would it suffice if we change the method signature documentation to be?
:param current_rate: Current buy rate. Use `exchange.get_rate` if you need sell rate.

In this way, we leave a guide for those who need it but get to keep the signature the same as for all other callbacks.

rokups:
A random thought: `custom_stake_amount` is nothing but a `adjust_trade_position` done on trade with no orders. I find myself implementing nearly same thing in both callbacks. I think it would make some sense to merge them both so `adjust_trade_position` would control both initial and subsequent trade size adjustments. This could be done in a backwards-compatible manner as well.

xmatthias:
> This could be done in a backwards-compatible manner as well.

Can you explain your thought here? 
backwards-compatible would assume that current behaviour of `custom_stake_amount` won't change.
Assuming i have a strategy that implements this (and for simplicity's sake, always returns 100).
This would mean all my positions will be for 100$.

Now "reusing" this callback for position adjustment would mean freqtrade would buy another 100$ in every loop (where no other open order is present) - for the whole duration of the trade - which is clearly not the original intend.

rokups:
My idea is that `adjust_trade_position` could also set initial stake amount and thus replace `custom_stake_amount`. Then, default `adjust_trade_position` implementation would call `custom_stake_amount` so it works as it always did. Of course there would be extra logic to ensure that this only happens for the very first order and not on every tick, when DCA orders are disabled. Docs would describe `custom_stake_amount` as deprecated and have a warning that users who implement `adjust_trade_position` should no longer use `custom_stake_amount`. First call to `adjust_trade_position` may have `trade` parameter set to `None`, that would imply this is setting initial trade stake size. Only slightly weird detail would be a need to pass both `trade` and `pair` parameters to `adjust_trade_position`, but that is still much better than having to implement two very similar callbacks.

xataxxx:
What you are describing is a valid but not the only use case for those callback methods.

`custom_stake_amount` allows overriding the default stake amount for the initial buy order, answering only the question 'How much?'

`adjust_trade_position` answers two questions together - 'When and how much?'
The WHEN part in the case of DCA usually will depend on the profit loss and some indicators and once the first safety order / dca order is executed, the loss percentage changes enough that the bot will not attempt to buy more until price changes significantly.

The How Much part can (but may not) depend on the same parameters then custom_stake_amount. If your implementation is the same and you want to avoid having duplicate code then just call self.custom_stake_amount in your adjust_trade_position.

NB: Not currently part of this pull request, but adjust_trade_position may in the future support negative stake amount to indicate that part of the position should be sold. But that part is still under further discussion and not planned as part of this pull request. It may be better to split the selling under a different callback.

freqtrade/freqtradebot.py
xmatthias: the strategy does not have direct access to `exchange` - nor should it be directly exposed (you can lock your bot up (or crash it) with some ill-placed calls to the exchange class).

If we need this, it should be through an interface in `dataprovider` (so it'd be `dp.get_rate`).

freqtrade/strategy/interface.py
```diff
@@ -381,6 +384,27 @@ def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: f
         """
         return proposed_stake
 
+    def adjust_trade_position(self, pair: str, trade: Trade, current_time: datetime,
+                              current_rate: float, current_profit: float, **kwargs
+                              ) -> Optional[float]:
+        """
```
xmatthias: Thinking about the discussion we're just having - wouldn't we want to have min/max stake available as parameters here too?

That would help stay within "tradable" limits i think - as you'd know how much is available - as well as how much you MUST buy for the order to go through (10$ - for example).

---

potentially also `buy_count` - as i see this basically in every implementation so far - and we could provide this pretty easily as either a property on the Trade object (-> `trade.buy_order_count`) - or directly as argument to the function.

I think the first has some benefits over the second (it'll make it automatically available to all other "trade handling"/"exiting" callbacks via the trade object) ...

xmatthias:
> Only slightly weird detail would be a need to pass both trade and pair parameters to adjust_trade_position, but that is still much better than having to implement two very similar callbacks.

that's a weak argument - current implementation of `adjust_trade_position` already has both pair and trade as parameters (i didn't notice it so far - but i'll also leave it open for the moment until we have a consensus on this...).


> NB: Not currently part of this pull request, but adjust_trade_position may in the future support negative stake amount to indicate that part of the position should be sold. But that part is still under further discussion and not planned as part of this pull request. It may be better to split the selling under a different callback.

i think that this last part will make a combined `custom_stake_amount` and `adjust_trade_position` quite odd to implement (potentially resulting in long, complex functions) - as a negative stake in the initial call would result in a order refusal - while for subsequent calls it would be a valid response.

`adjust_trade_position` should support both positive and negative adjustments in one callback (not split into two) once we'll support that (which will not be in this PR). 
This would make the "custom_stake_amount" callback a 3 way callback (initial stake, increase and decrease).
I'm not certain if these 3 in one callback will be very comfortable from a user perspective.

I understand that combining the two callbacks could benefit some cases to reduce code duplication - it can on the other hand also make the code more complex (basically everyone would have to handle the "first case" different to the "adjustment" case - as the logic will most likely be slightly different)

e.g.:
``` python

if trade: 
    # Initial call
else:
    # subsequent ... 
```

On the other hand, with split callbacks - users needing it could implement a combined `stake_and_adjust` (name however you like) - which you call with whatever parameters necessary for that particular function - IF the same exact logic will be shared (as you've done above).

That'd however keep this part optional for the user - without forcing potentially strange handling of the "initial" case should that be a completely different logic (we currently assume that it's an initial call if trade is None - which seems a bit odd from an interface perspective...)

---

Naming would however be completely confusing with the proposed above.
assuming we remove `custom_stake_amount` in a future release (a few releases after deprecation) - then 
no logic i can come up with tells me that `adjust_trade_position` would define the INITIAL stake amount - so it'd be quite counter-intuitive for the user.

if any, then the name of the combined callback should be `custom_stake_amount` - based on the logic that "it's adjusting the stake amount throughout the trade".
This could potentially be a more difficult change to handle without a breaking change depending on the parameters which are required (but different).

Assuming users follow the "keep **kwargs" advice (i'm not sure i care about the rest, really) - it might actually be pretty simple, as i don't see us removing any arguments - and the whole "DCA" logic is "opt-in" via additional flag - so changes would only be required in these cases - where you'll need to implement the logic for DCA anyway.

docs/strategy-callbacks.md
```diff
@@ -568,3 +569,101 @@ class AwesomeStrategy(IStrategy):
...
+        # Obtain pair dataframe (just to show how to access it)
+        dataframe, _ = self.dp.get_analyzed_dataframe(pair, self.timeframe)
+        # Only buy when not actively falling price.
+        last_candle = dataframe.iloc[-1].squeeze()
+        previous_candle = dataframe.iloc[-2].squeeze()
+        if last_candle['close'] < previous_candle['close']:
+            return None
+
+        count_of_buys = 0
+        initial_order_stake =
```
lemmiktort: Should be removed?

tests/test_freqtradebot.py
xmatthias: I've given this test a try now in the last commit ... ;)

freqtrade/strategy/interface.py
xataxxx: I'm not sure about the buy_order_count - Yes, the example i've provided and what most people are using is relying on the previous buy order count, but only to calculate the next amount via multiplication. 

The strategy I'm personally currently testing is actually using the previously bought pair stake amount sum, meaning the count itself is meaningless, it's the stake amount that has already been invested into trade that matters.

Also, with some orders getting partially filled and some failing, the "count" and actual value of those orders could differ a lot.

xataxxx: Regarding "min/max stake available as parameters" - Yes, technically we are providing those to custom_stake_amount so probably makes sense to add them here also.

freqtrade/optimize/backtesting.py
```diff
@@ -357,8 +357,30 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
         else:
             return sell_row[OPEN_IDX]
 
+    def _get_adjust_trade_entry_for_candle(self, trade: LocalTrade, row: Tuple
+                                           ) -> LocalTrade:
+
+        current_profit = trade.calc_profit_ratio(row[OPEN_IDX])
+        stake_amount = strategy_safe_wrapper(self.strategy.adjust_trade_position,
```
xmatthias, line 363: Please put a warning box in the docs advising users that this callback will be called for each `timeframe-detail` candle during backtesting (so if you backtest with `--timeframe 1h --timeframe-detail 1m` - it's going to be called 60 times per hour.

The same happens for exit callbacks - but i find it surprising in this instance.

It can be a feature - BUT you (or rather, users) MUST be aware of this to avoid nasty surprises.

xataxxx: I'm adding a documentation warning. This is working as expected as position adjustment is called in live mode also every 5 seconds and the purpose of position adjustment is the ability to react to market movements (not waiting for candle arrival). So in backtesting we are running it also as detailed as possible.

xmatthias: As said, it's about being aware / know what to expect :+1:

freqtrade/strategy/interface.py
xataxxx: buy_count idea is now also resolved by having trade.nr_of_successful_buys() and trade.nr_of_successful_sells() functions available straight at trade object, allowing us to use them in multiple callbacks without having to add callback arguments and calculating them if they are not actually used in the implementation.

freqtrade/persistence/models.py
```diff
@@ -583,6 +614,24 @@ def select_order(self, order_side: str, is_open: Optional[bool]) -> Optional[Ord
         else:
             return None
 
+    def nr_of_successful_buys(self) -> int:
```
xmatthias, line 629: From a usability perspective, i think this might make more sense as property (decorated by `@property`).

that way, access would be via `trade.nr_of_successful_buys` - instead of calling it.

is there any deeper thought behind as to why you've created it as function instead?

freqtrade/freqtradebot.py
```diff
@@ -179,6 +179,10 @@ def process(self) -> None:
             # First process current opened trades (positions)
             self.exit_positions(trades)
 
+        # Check if we need to adjust our current positions before attempting to buy new trades.
+        if self.strategy.position_adjustment_enable:
+            self.process_open_trade_positions()
```
xmatthias: We'll have to wrap this in a `with self._exit_lock` block.

Otherwise we risk a timing issue - if the bot decides to place a new buy order, and at the same time receives a `/forcesell` request via API/Telegram, which (worst case) would lead to a "lost" order (as open_order_id get's overwritten again).

freqtrade/optimize/backtesting.py
```diff
@@ -433,11 +457,16 @@ def _get_sell_trade_entry(self, trade: LocalTrade, sell_row: Tuple) -> Optional[
...
-            stake_amount = self.wallets.get_trade_stake_amount(pair, None)
-        except DependencyException:
-            return None
+    def _enter_trade(self, pair: str, row: Tuple, stake_amount: Optional[float] = None,
+                     trade: Optional[LocalTrade] = None) -> Optional[LocalTrade]:
+
+        pos_adjust = trade is not None
+        if stake_amount is None:
+            try:
+                stake_amount = self.wallets.get_trade_stake_amount(pair, None)
```
xmatthias: You should be able to move this block down to line 482.
stake_amount is not used anywhere before that point - and the check for `stake_amount is None` vs. `not pos_adjust` is actually identical based on how `_enter_trade()` is called.

freqtrade/freqtradebot.py
xataxxx: I see you resolved this already.

freqtrade/persistence/models.py
xataxxx: No other reason besides the fact that I did not know @property works so far. Just read upon it. Lack of python skill bites me.

freqtrade/optimize/backtesting.py
```diff
@@ -479,6 +472,13 @@ def _enter_trade(self, pair: str, row: Tuple, stake_amount: Optional[float] = No
         min_stake_amount = self.exchange.get_min_pair_stake_amount(pair, propose_rate, -0.05) or 0
         max_stake_amount = self.wallets.get_available_stake_amount()
 
+        pos_adjust = trade is not None
+        if stake_amount is None:
```
xmatthias: i think you missunderstood this slightly.

In my opinion, you can remove the `if stake_amount is None` condition - and move it into the subsequent IF - as it's either `stake_amount is None and not pos_adjust` ... but never only one OR the other.

freqtrade/optimize/backtesting.py
xataxxx: Moved code. I agree with them being currently the same since stake_amount and trade either are sent together or not BUT I'd leave them like this for now since that might not always be the case in the future and this way it's more "safe".

freqtrade/optimize/backtesting.py
xataxxx: I agree with them being currently the same since stake_amount and trade either are sent together or not BUT I'd leave them like this for now since that might not always be the case in the future and this way it's more "safe".

xataxxx: Fine, it actually does look more readable like that.

freqtrade/persistence/models.py
xataxxx: Added @property, looks cleaner in the using code.

freqtrade/freqtradebot.py
xmatthias: yeah, i tend to do that sometimes :laughing:

freqtrade/strategy/interface.py
```diff
@@ -381,6 +384,28 @@ def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: f
         """
         return proposed_stake
 
+    def adjust_trade_position(self, trade: Trade, current_time: datetime,
+                              current_rate: float, current_profit: float, min_stake: float,
```
xmatthias, line 388: I think we should have a `proposed_stake` amount as well.
that would move some complexity from the strategy to internal bot code (calling `self.dp.wallets.` - which i still don't really see as public interface).

and it seems to be used in all samples I've seen so far, so I'd think the need is there, and it would greatly simplify usage in the strategy.

xataxxx: I'd be OK to add that, assuming we don't also call the custom_stake_amount to overwrite the wallet, because we call adjust_trade_position every x seconds and i think the performance just suffers too much. If we pass the wallet amount in then it should be OK.

xmatthias: no, it'd be similar/identical to the parameters we pass to `custom_stake_amount` ... 
calling `custom_stake_amount ` would make no sense - that's what the new callback is supposed to do for subsequent calls.

i don't think it makes a huge difference if that calculation is done as part of the strategy or before - unless you have seen / experienced something that i didn't.

xataxxx: OK, so resolved this via removing references to internal wallets and replaced that with using the initial orders stake amount.
`
filled_buys = trade.select_filled_orders('buy')
stake_amount = filled_buys[0].cost
`

xmatthias (approved):
Pretty darn good job :+1: 

I'm sure we'll discover some edge-cases as more people start using this - but for the moment i'd not know anything more to do on this PR.

LGTM :+1:

docs/strategy-callbacks.md
```diff
@@ -568,3 +569,110 @@ class AwesomeStrategy(IStrategy):
...
+        
+        # Allow up to 3 additional increasingly larger buys (4 in total)
+        # Initial buy is 1x
+        # If that falls to -5% profit, we buy 1.25x more, average profit should increase to roughly -2.2%
+        # If that falls down to -5% again, we buy 1.5x more
+        # If that falls once again down to -5%, we buy 1.75x more
+        # Total stake for this trade would be 1 + 1.25 + 1.5 + 1.75 = 5.5x of the initial allowed stake.
+        # That is why max_dca_multiplier is 5.5
+        # Hope you have a deep wallet!
+        if 0 < count_of_buys <= self.max_dca_orders:
```
italodamato, line 666: Great job to both of you! I have a question about the `select_filled_orders` method. 

Wouldn't this condition trigger multiple orders if the first order doesn't get filled right away? I think this count should include the potential pyramidal orders created in the last cycle. 

As far as I understand, we're counting just filled orders with [these statuses](https://github.com/freqtrade/freqtrade/blob/c12f2378dbaa9bff3d315c70e93e3362cf68fadf/freqtrade/constants.py#L57) which would not select orders we might have already placed at this level and which are still open.

xmatthias: Please read the documentation that was added as part of this PR.

this callback will not be called if a order is open.

umutrcc:
Will we be able to do partial trading now?  How to use it, can you share a detailed usage or a sample study?  thank you.

rokups:
https://www.freqtrade.io/en/latest/strategy-callbacks/#adjust-trade-position

docs/strategy-callbacks.md
italodamato: Sorry I missed it! It might be worth adding it to the function docstring as well.

ghost:
This is a nice feature, but i don't think the example is very useful because profitable trades will only have a small stake, which means only small profits on these trades and losing trades will have high stakes, which means much higher losses are possible. 
Maybe using buy signals from strategy to buy more coins would be better, but i don't think it is possible at the moment because pairs with open trades won't be analyzed by the bot - so maybe changing this would be a good idea?

rokups:
Point of example is to demonstrate how to use this feature. Profitability of samples is really out of scope of technical documentation.

ghost:
Ok, but do you think we can use additional buy signals to adjust positions?

stash86:
Anything is possible if you know how to code it

rokups:
Signal adjustment callback runs on each candle. You may access dataframe from this callback and see if last candle has your signal.  See here https://www.freqtrade.io/en/stable/strategy-advanced/#dataframe-access

ghost:
Thank you for your reply, this was very helpful. 
With this i can access all indicators of the strategy, so i just have to put the code for my buy signal inside populate_indicators and can then use that to adjust positions.

perkmeister:
This post on Discord may help you @incrementby1. As an aside, backtesting is broken if you remove the current_profit check @xmatthias 
https://discord.com/channels/700048804539400213/920652305936842752/934926696195444756

xmatthias:
That link kindof leads to the rules channel ... so not "quite" sure what you're trying to point me to ...

perkmeister:
Oh sorry, its the Freqtrade Discord server, my post regarding using buy signals (or other dataframe signals) to trigger DCA buys individually. I'll fill in a github issue report regarding the backtesting.

perkmeister:
That link doesn't open properly, yet it's the link copied directly from the post (in the DCA thread). Odd... Anyway, do a search for DCATest2_fixed.py on Discord.

xataxxx:
A note to those who might read it later - "backtesting is broken if you remove the current_profit check" is wrong, everything works as expected. If you remove current_profit then you'll just end up buying until you run out of money. It's the responsibility of adjust_trade_position to decide if it's the right moment to buy more - if you don't check your current profit then you better check something else and do not return a stake amount if you don't want to buy.

perkmeister:
True, but there is no difference between that and failing the current_profit check on all DCA buys. In backtesting this would fail, in live/dry it behaves correctly as it buys as much as it can with DCA and holds the trades open.

perkmeister:
@xataxxx If you think backtesting isn't broken, run this simple strategy that just buys 3 times then sells and repeats. It should not (IMO) immdiately close trades with a profit of -0.2%, the user is in complete control with max_dca_orders and max_dca_multiplier, it should just buy the max it can and leave the trade open just like it does in live or dry. I'm concerned this is highlightng an underlying logical issue that might show itself if a current_profit check were to fail for each DCA buy, and if adjust_trade_position() is called once per candle and current_profit is the open rate of the next candle this is entirely possible. I may of course be misinterpreting things, but maybe you could explain why this behaviour shows it 'works as expected'?
[DCATest2_broken.py.TXT](https://github.com/freqtrade/freqtrade/files/7959603/DCATest2_broken.py.TXT)

xmatthias:
@perkmeister 

If you think something is broken, please open an issue with **all necessary details** (this includes a simple strategy to reproduce the problem, but also output that's produced by it, the command used for this (if necessary some config), and maybe a description what you'd expect and what happened.


Based on the above comment, you'd expect us to first investigate what's actually happening, while then comparing it what you think would happen - which will quite frankly - simply waste our time.

xataxxx:
Just for info for future readers, @perkmeister issue was actually caused by another tiny temporary bug in codebase that disabled adjust_trade_position completely, not in the actual backtesting functionality.

perkmeister:
Actually as we said there were 2 problems, it seems the zero length trade issue was an issue and has been fixed in the latest update (only a few days ago it seems), and the latest update has a minor bug with the adjust_trade_position being at config level not strategy level.
@xmatthias I fully intended to raise this as a formal issue with all the outputs and the like (as I said I would in the chat), I just hadn't got round to it yet, but I take your point. I should have done so sooner, my apologies.

serguzz:
Hi team.

There is another use case for placing multiple order for one pair.
When there is a big density (a big limit order) on the exchange and we want to place 3-5 limit orders close to that density, for example, at the distances 0.4%, 0.2%, 0.1% from the density.
In case there will be sharp price movement towards the density only some of our orders may be filled.

As I understand, this use case can't be implemented with `adjust_trade_position` as there will be delays in `adjust_trade_position` execution and placing the new limit order ?

Is it planned to implement possibility for multiple limit orders for one pair?
Or am I missing something and it is possible to implement this use case with `adjust_trade_position`?

xmatthias:
This is an PR that's been merged 4 years ago.
It's not implemented, is not really planned at the moment - but this is the wrong place to discuss this for sure.

PR #6084, opened 2021-12-19
Telegram and Webhook updates

aezomz:

## Summary
Added support for leverage and long/short values

## Quick changelog
- Updated telegram and webhook documentation
- Added leverage and long / short details

## What's new?
- Updated telegram and webhook documentation
- Added leverage and long / short details


docs/webhook-config.md
```diff
@@ -152,13 +156,15 @@ Possible parameters are:
 * `enter_tag`
 
 ### Webhooksell
-
+# TODO-lev add support for long and short?
```
xmatthias: Long and short support is there by providing "is_short" ? or am i missing something?

docs/webhook-config.md
```diff
@@ -114,6 +116,9 @@ Possible parameters are:
 * `order_type`
 * `current_rate`
 * `enter_tag`
+* `gain`
+* `profit_amount`
+* `profit_ratio`
```
aezomz: Do we add gain, profit_amount to everything here since we can long/short ?

xmatthias: it's unclear to me what the difference between "gain" and "profit_amount" is ... in my understanding - both specify "i made 10$ profit" ... ?

freqtrade/rpc/rpc.py
```diff
@@ -247,7 +248,8 @@ def _rpc_status_table(self, stake_currency: str,
                     trade.id,
                     trade.pair + ('*' if (trade.open_order_id is not None
                                           and trade.close_rate_requested is None) else '')
-                               + ('**' if (trade.close_rate_requested is not None) else ''),
+                               + ('**' if (trade.close_rate_requested is not None) else '')
+                               + f' direction_str',
```
aezomz: I think we should show if the trade is long / short in the status table command. Just need to make it "prettier"

docs/webhook-config.md
xmatthias: "everything" ... it should be added where it's really available - it will not make sense for order cancel's (for example).

aezomz: Its copied from sell fill. 
`gain = "profit" if profit_ratio > 0 else "loss"
`
I think i need to update this 
`freqtrade/freqtradebot.py`
`def _notify_exit`
to add leverage and is_short. 

Guess i need to update to account for these
RPCMessageType.SHORT_FILL
RPCMessageType.SHORT
RPCMessageType.SHORT_CANCEL

aezomz: I will update the pr to account for those, this is just draft for discussion :+1:

freqtrade/rpc/rpc.py
xmatthias: for "/status" - it might be pretty easy.
`/status table` might be harder - maybe prefixing the trade id with "L" or "S"? (so it's S-2 - or L-5) ... 
it's not "too pretty" - but i'd not know how to show this better

aezomz:
Please let me know if I miss out any test cases. I tried to be as logical as possible especially with the direction key being passed from freqtradebot.py . Quite late with this been busy with Christmas.

freqtrade/rpc/rpc.py
aezomz: I went with "2 L" or "3 S"

freqtrade/rpc/rpc.py
```diff
@@ -244,7 +245,7 @@ def _rpc_status_table(self, stake_currency: str,
                         fiat_profit_sum = fiat_profit if isnan(fiat_profit_sum) \
                             else fiat_profit_sum + fiat_profit
                 trades_list.append([
-                    trade.id,
+                    str(trade.id) + f' {direction_str}',
```
xmatthias:
```suggestion
                    f'{trade.id} {direction_str}',
```

freqtrade/rpc/telegram.py
```diff
@@ -221,20 +221,24 @@ def _format_buy_msg(self, msg: Dict[str, Any]) -> str:
...
+                      in [RPCMessageType.BUY_FILL, RPCMessageType.BUY]
+                      else {'enter': 'Short', 'entered': 'Shorted'})
         message = (
-            f"{emoji} *{msg['exchange']}:* {'Bought' if is_fill else 'Buying'} {msg['pair']}"
+            f"{emoji} *{msg['exchange']}:*"
+            f" {enter_side['entered'] if is_fill else enter_side['enter']} {msg['pair']}"
             f" (#{msg['trade_id']})\n"
             )
         message += f"*Enter Tag:* `{msg['enter_tag']}`\n" if msg.get('enter_tag', None) else ""
+        message += f"*Leverage:* `{msg['leverage']}`\n" if msg.get('leverage', None) else ""
```
xmatthias: I don't think this test for leverage (neither here, nor when buying, nor in tests) is correct.

It's based on the assumption that a "regular" order (spot order) will have a leverage of `None`.
it will however have a leverage of 1 - which will resolve to True if checked as it is here.
I think this should check `msg.get('leverage', 1.0) != 1.0` ... obviously we'd want to also slightly adapt tests to account for this.

Screenshot_20211230_172336


I've adapted/updated the code for this in one of the latest commits

xmatthias:
> Quite late with this been busy with Christmas.

No rush :+1: 

> Please let me know if I miss out any test cases. 

I think we're good for now. Most other endpoints don't use rate directly, but only profit, where long/short is not really relevant.
I'm sure we'll hit some issues once people start using it - but i think for the moment we'll be fine from a notification perspective.

samgermain:
🎉

aezomz:
:partying_face: , thanks Xmatthias and Sam definitely learned alot.

PR #6112, opened 2021-12-25
Fix test not running when user_data contains historical data.

xataxxx:
## Summary

Test `test_start_no_data` fails to run properly when user has user_data filled with data.
Since the purpose of the test is to test if it's throwing an error when no data is found then using a folder that surely does not contain suitable user data is better.

## Quick changelog

- Using tests folder as user_data folder.
- Since the test hyperopt by definition must exit with exception, it leaves behind a lock file. Remove it.

## What's new?

Using a different folder for running the test.


xmatthias (approved):
LGTM :+1:

PR #6114, opened 2021-12-26
Reduce kucoin logs

cdimauro:
All preliminary checks made and no issue found.

## Summary

Reduces several logs which are generated by the KuCoin exchange, when using its APIs to fetch che candles status.

## Quick changelog

- Reduce KuCoin logs on DDosProtection error messages
- More logs are reduced, for KuCoin, on the retrier_async decorator
- New test functions introduced to check logs

## What's new?

KuCoin APIs generate A LOT of error messages.
Consequently, logs are flooded with lines like:
```
2021-12-25 22:30:23 freqtrade.exchange.common: WARNING -
_async_get_candle_history() returned exception:
"kucoin GET https://openapi-v2.kucoin.com/api/v1/market/candles?
symbol=PDEX-USDT&type=5min&startAt=1640317818&endAt=1640467818
429 Too Many Requests {"code":"429000","msg":"Too Many Requests"}"
2021-12-25 22:30:23 freqtrade.exchange.common: WARNING -
retrying _async_get_candle_history() still for 3 times
2021-12-25 22:30:23 freqtrade.exchange.common: WARNING -
Kucoin 429 error, avoid triggering DDosProtection backoff delay.
2 tries left before giving up
2021-12-25 22:30:24 freqtrade.exchange.common: WARNING -
_async_get_candle_history() returned exception:
"kucoin GET https://openapi-v2.kucoin.com/api/v1/market/candles?
symbol=UBX-USDT&type=5min&startAt=1640317821&endAt=1640467821
429 Too Many Requests {"code":"429000","msg":"Too Many Requests"}"
```

Messages like:
- Kucoin 429 error, avoid triggering DDosProtection backoff delay.
- _async_get_candle_history() returned exception
- retrying _async_get_candle_history() still for
- Giving up retrying: _async_get_candle_history()
- Applying DDosProtection backoff delay

are logged only once for a certain period of time (default is 3600 seconds).
A new test case, test__async_kucoin_get_candle_history, is introduce to check that the above messages are logged only once for the KuCoin exchange.

Finally, new functions log_contains, num_log_contains, num_log_has and num_log_has_re
are introduced in the conftest module to help and simplify checking:
- if logs contain a string;
- count how many messages contain a string;
- count how many messages are the given string;
- count how many messages matchs a regex.

A couple of existing tests are changed using the new functions.

IMO it would be even better to replace the existing log_contains, log_has, log_has_re with the corresponding new num_* functions (so, having only 3 functions for logs checking), because they are logically equivalent to the current ones (since the num_* functions return 0 if no log matches and >= 1 if some log matches, which are logically equivalent to a "false" and a "true" condition in Python).
I've also a question about those log functions: does it make sense to check both the logger_name and the message from the logs, instead of just checking the message (which both simplifies the implementation, and make it also faster)? Is there some special test that requires to check the logger_name?

### DIFF
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -80,9 +80,9 @@ def retrier_async(f):
         try:
             return await f(*args, **kwargs)
         except TemporaryError as ex:
-            logger.warning('%s() returned exception: "%s"', f.__name__, ex)
+            msg = f'{f.__name__}() returned exception: "{ex}". '
             if count > 0:
-                logger.warning('retrying %s() still for %s times', f.__name__, count)
+                logger.warning(msg + f'Retrying still for {count} times.')
                 count -= 1
                 kwargs['count'] = count
                 if isinstance(ex, DDosProtection):
@@ -98,7 +98,7 @@ def retrier_async(f):
                         await asyncio.sleep(backoff_delay)
                 return await wrapper(*args, **kwargs)
             else:
-                logger.warning('Giving up retrying: %s()', f.__name__)
+                logger.warning(msg + 'Giving up.')
                 raise ex
     return wrapper
 
@@ -111,9 +111,9 @@ def retrier(_func=None, retries=API_RETRY_COUNT):
             try:
                 return f(*args, **kwargs)
             except (TemporaryError, RetryableOrderError) as ex:
-                logger.warning('%s() returned exception: "%s"', f.__name__, ex)
+                msg = f'{f.__name__}() returned exception: "{ex}". '
                 if count > 0:
-                    logger.warning('retrying %s() still for %s times', f.__name__, count)
+                    logger.warning(msg + f'Retrying still for {count} times.')
                     count -= 1
                     kwargs.update({'count': count})
                     if isinstance(ex, (DDosProtection, RetryableOrderError)):
@@ -123,7 +123,7 @@ def retrier(_func=None, retries=API_RETRY_COUNT):
                         time.sleep(backoff_delay)
                     return wrapper(*args, **kwargs)
                 else:
-                    logger.warning('Giving up retrying: %s()', f.__name__)
+                    logger.warning(msg + 'Giving up.')
                     raise ex
         return wrapper
     # Support both @retrier and @retrier(retries=2) syntax
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -82,7 +82,7 @@ def retrier_async(f):
         except TemporaryError as ex:
             msg = f'{f.__name__}() returned exception: "{ex}". '
             if count > 0:
-                logger.warning(msg + f'Retrying still for {count} times.')
+                msg += f'Retrying still for {count} times.'
                 count -= 1
                 kwargs['count'] = count
                 if isinstance(ex, DDosProtection):
@@ -92,10 +92,14 @@ def retrier_async(f):
                         log_once_warning(
                             f"Kucoin 429 error, avoid triggering DDosProtection backoff delay. "
                             f"{count} tries left before giving up")
+                        # Reset msg to avoid logging too many times.
+                        msg = ''
                     else:
                         backoff_delay = calculate_backoff(count + 1, API_RETRY_COUNT)
                         logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
                         await asyncio.sleep(backoff_delay)
+                if msg:
+                    logger.warning(msg)
                 return await wrapper(*args, **kwargs)
             else:
                 logger.warning(msg + 'Giving up.')
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -1,16 +1,23 @@
 import asyncio
 import logging
 import time
-from functools import partial, wraps
+from functools import wraps
 
 from freqtrade.exceptions import DDosProtection, RetryableOrderError, TemporaryError
 from freqtrade.mixins import LoggingMixin
 
 
 logger = logging.getLogger(__name__)
-logging_mixin = LoggingMixin(logger)
-log_once_info = partial(logging_mixin.log_once, logmethod=logger.info)
-log_once_warning = partial(logging_mixin.log_once, logmethod=logger.warning)
+__logging_mixin = None
+
+
+def _get_logging_mixin():
+    # Logging-mixin to cache kucoin responses
+    # Only to be used in retrier
+    global __logging_mixin
+    if not __logging_mixin:
+        __logging_mixin = LoggingMixin(logger)
+    return __logging_mixin
 
 
 # Maximum default retry count.
@@ -89,9 +96,9 @@ def retrier_async(f):
                     if kucoin and "429000" in str(ex):
                         # Temporary fix for 429000 error on kucoin
                         # see https://github.com/freqtrade/freqtrade/issues/5700 for details.
-                        log_once_warning(
+                        _get_logging_mixin().log_once(
                             f"Kucoin 429 error, avoid triggering DDosProtection backoff delay. "
-                            f"{count} tries left before giving up")
+                            f"{count} tries left before giving up", logmethod=logger.warning)
                         # Reset msg to avoid logging too many times.
                         msg = ''
                     else:
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -9,6 +9,7 @@ from freqtrade.mixins import LoggingMixin
 
 logger = logging.getLogger(__name__)
 logging_mixin = LoggingMixin(logger)
+log_once_info = partial(logging_mixin.log_once, logmethod=logger.info)
 log_once_warning = partial(logging_mixin.log_once, logmethod=logger.warning)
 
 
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -63,6 +70,27 @@ def log_has_re(line, logs):
                   False)
 
 
+def num_log_contains(line, logs, count=False):
+    # caplog mocker returns log as a tuple: ('freqtrade.something', logging.WARNING, 'spamfoobar')
+    # and we want to check how many times line ('foo', for example) is contained in the tuples.
+    return sum(line in logger_name or line in message
```
xmatthias: You'll definitely not want to match "line" against both logger_name and message.

if you think there is a need to know where the log message comes from (logger_name) - this should be a separate argument.
I've yet to encounter a case where i care about that though ...

freqtrade/exchange/common.py
```diff
@@ -72,28 +76,34 @@ def calculate_backoff(retrycount, max_retries):
 def retrier_async(f):
     async def wrapper(*args, **kwargs):
         count = kwargs.pop('count', API_RETRY_COUNT)
+        self = args[0]  # Extract the exchange instance.
+        kucoin = self.name == "Kucoin"
```
xmatthias: As you're only using `self` once (without checking if it actually is of the expected type - which i know is true, but is not really the case for a generic decorator like this one) - that's a pointless assignment, and you can directly check with `args[0].name`.

freqtrade/exchange/common.py
```diff
@@ -72,28 +76,34 @@ def calculate_backoff(retrycount, max_retries):
...
         try:
             return await f(*args, **kwargs)
         except TemporaryError as ex:
-            logger.warning('%s() returned exception: "%s"', f.__name__, ex)
+            message = f'{f.__name__}() returned exception: "{ex}"'
+            (log_once_warning if kucoin else logger.warning)(message)
```
xmatthias: That's a dangerous premise, which may result in important messages to be skipped (for example during downloading data - or other ratelimiting messages if rate-limit is set too low).

The **only place** i feel a cached logger is acceptable is within the IF block that captures the very specific 429000 kucoin error code.

All other exceptions (for example 429 without "429000") are valid rate-limiting exceptions - which i will expect to have logged.

freqtrade/exchange/common.py
```diff
@@ -72,28 +76,34 @@ def calculate_backoff(retrycount, max_retries):
...
                         backoff_delay = calculate_backoff(count + 1, API_RETRY_COUNT)
-                        logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
+                        message = f"Applying DDosProtection backoff delay: {backoff_delay}"
+                        (log_once_info if kucoin else logger.info)(message)
                         await asyncio.sleep(backoff_delay)
                 return await wrapper(*args, **kwargs)
             else:
-                logger.warning('Giving up retrying: %s()', f.__name__)
+                message = f'Giving up retrying: {f.__name__}()'
+                (log_once_warning if kucoin else logger.warning)(message)
```
xmatthias: same applies here - no cached logger for non 429000 messages. 

this is especially true here - which is a non-repeatable message (could for example be an error when placing an order). Having log cached (not shown) in such a case is clearly a bug.

xmatthias (changes requested):

xmatthias:
> IMO it would be even better to replace the existing log_contains, log_has, log_has_re with the corresponding new num_* functions (so, having only 3 functions for logs checking), because they are logically equivalent to the current ones (since the num_* functions return 0 if no log matches and >= 1 if some log matches, which are logically equivalent to a "false" and a "true" condition in Python).

That may be true - but it'd be a round of pointless changes to basically all test files without actually providing any benefit - unless you think of "sneaking" it into the "log_has" methods - which would be an option i guess.

However, I'm unclear as of now if we need this type of check in general (testing if a message is logged more than once) - as that is not something we're currently checking for - and usually, the messages will be slightly different.

tests/conftest.py
```diff
@@ -63,6 +70,27 @@ def log_has_re(line, logs):
...
+
+
+def num_log_has(line, logs, count=False):
+    # caplog mocker returns log as a tuple: ('freqtrade.something', logging.WARNING, 'foobar')
+    # and we want to check how many times line is presente in the tuples.
+    return sum(line == logger_name or line == message
+               for logger_name, level, message in logs.record_tuples)
+
+
+def num_log_has_re(line, logs, count=False):
```
xmatthias: why do all of these methods have an unused `count` argument?

tests/conftest.py
cdimauro: Removed the logger_name check.

freqtrade/exchange/common.py
cdimauro: I did it to make the code more clear about the fact that the first argument is the self parameter. Removed now.

freqtrade/exchange/common.py
cdimauro: Reverted part of the code: removed all log_once cases except the 429000 kucoin error code one.

freqtrade/exchange/common.py
cdimauro: As above: restored the previous code.

tests/conftest.py
cdimauro: It was a leftover from previous experiments with the counting methods. Removed now.

cdimauro:
> 

Reworked log_has and log_has_re as per above. Now they are also simpler and more optimized.

tests/conftest.py
```diff
@@ -63,6 +70,27 @@ def log_has_re(line, logs):
...
+    # caplog mocker returns log as a tuple: ('freqtrade.something', logging.WARNING, 'foobar')
+    # and we want to check how many times line matches in the tuples.
+    return sum(re.match(line, logger_name) or re.match(line, message)
+               for logger_name, level, message in logs.record_tuples)
```
xmatthias: this method is not used, which is hiding the fact that it's not working.
it'll fail with `unsupported operand type(s) for +: 'int' and 'NoneType'` if provided with a match which does not match (i suspect it'll also fail in case of matches - as sum expects a "int-like" value (it works with bools as they default to 1/0 - but not for "re.match" - which is not a boolean flag)

I've added some fixes to this (and reduced the amount of methods).
while it's not wrong to have them - i think it'll be confusing as to "when to use which" - and the performance impact of using `re.match` over `in` is neglectible in case of tests.

freqtrade/exchange/common.py
```diff
@@ -72,28 +76,33 @@ def calculate_backoff(retrycount, max_retries):
...
                             f"Kucoin 429 error, avoid triggering DDosProtection backoff delay. "
                             f"{count} tries left before giving up")
+                        # Reset msg to avoid logging too many times.
+                        msg = ''
                     else:
                         backoff_delay = calculate_backoff(count + 1, API_RETRY_COUNT)
                         logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
                         await asyncio.sleep(backoff_delay)
+                if msg:
+                    logger.warning(msg)
```
xmatthias, line 109: I've now reduced the overall verbosity of these logs by combining the messages.

for kucoin's 429000 - we're resetting the msg - disabling "additional" output. This should reduce the number of messages for one kucoin exception from 3 to 1 (per exception).
The message can occur up to 4 times (once per "count") - with the last one followed up by a "giving up" message.

The first 4 messages will be cached for 1 hour - while the last (giving up) is not - as this indicates a real problem (kucoin responded "wrongly" 4 times in a row to the same request ... so something is off more than usually).

freqtrade/exchange/common.py
```diff
@@ -1,12 +1,16 @@
...
 import time
-from functools import wraps
+from functools import partial, wraps
 
 from freqtrade.exceptions import DDosProtection, RetryableOrderError, TemporaryError
+from freqtrade.mixins import LoggingMixin
 
 
 logger = logging.getLogger(__name__)
+logging_mixin = LoggingMixin(logger)
```
xmatthias: the only thing that slightly still concerns me is this section here.

it's creating a singleton (actually 3) - which can potentially cause problems (especially to tests).
While currently, i don't think it's creating actual problems to tests - it might cause random test failures depending on the randomness used in the test.

I'm however currently lacking a better solution (in theory, LoggingMixin could be tied to the exchange instance - but that'd again assume that retrier is only ever going to be used on an exchange class.

cdimauro: I don't think that it might cause problems. At the very end LoggingMixin is used "mixed" with other classes, but actually it means that there are several "instances" of this "class" which are spread around and serving the same purpose (allowing parts of the code to "log once").
In this case there isn't a real class which is mixed with LoggingMixin, but only and directly an instance of the latter, serving exactly the same purpose ("log once").

I've seen all other changes that you committed: understood and agreed.

How to continue now with this pull? Do you think that further changes are needed?

xmatthias: > but actually it means that there are several "instances" of this "class" which are spread around and serving the same purpose (allowing parts of the code to "log once").

All other instances will be destroyed / recreated in the following instances:
* `/reload_config`
* new test case
* parent class is destroyed and recreated

Neither of which is true for the above instance, which is only ever created when this file is imported (which will only happen once per python process - as python caches/remembers imports) - and can only be destroyed by either calling `del` directly (which would cause the remainder to crash).

I'll need to think and test more with this to determine if it can become a problem before deciding on a way forward.

xmatthias: I think with the new approach (only create the LoggingMixing instance when necessary) we can have a good compromise.
it'll be cached for kucoin users with this problem - and will not affect any other instance (Every class requires a little bit of memory - so we can avoid that if possible).

xmatthias (approved):
LGTM :+1:

freqtrade/exchange/common.py
cdimauro: Indeed. That's a good compromise. Thanks!

PR #6127, opened 2021-12-27
Download data futures fix

wadedyck:

## Summary

Fixes a download_data bug when in futures mode.

## Quick changelog

- When specifying multiple pairs to download, the json filenames were
inconsistent due to the reassignment of candle_type.
- Adds the candle_type being downloaded to a log message.

## What's new?

Bug fix.

### DIFF
--- a/freqtrade/commands/data_commands.py
+++ b/freqtrade/commands/data_commands.py
@@ -64,6 +64,8 @@ def start_download_data(args: Dict[str, Any]) -> None:
     try:
 
         if config.get('download_trades'):
+            if config.get('trading_mode') == 'futures':
+                raise OperationalException("Trade download not supported for futures.")
             pairs_not_available = refresh_backtest_trades_data(
                 exchange, pairs=expanded_pairs, datadir=config['datadir'],
                 timerange=timerange, new_pairs_days=config['new_pairs_days'],
--- a/freqtrade/commands/data_commands.py
+++ b/freqtrade/commands/data_commands.py
@@ -64,8 +64,6 @@ def start_download_data(args: Dict[str, Any]) -> None:
     try:
 
         if config.get('download_trades'):
-            if config.get('trading_mode') == 'futures':
-                raise OperationalException("Trade download not supported for futures.")
             pairs_not_available = refresh_backtest_trades_data(
                 exchange, pairs=expanded_pairs, datadir=config['datadir'],
                 timerange=timerange, new_pairs_days=config['new_pairs_days'],
--- a/tests/commands/test_commands.py
+++ b/tests/commands/test_commands.py
@@ -808,6 +808,7 @@ def test_download_data_trades(mocker, caplog):
         "--exchange", "kraken",
         "--pairs", "ETH/BTC", "XRP/BTC",
         "--days", "20",
+        "--trading-mode", "spot",
         "--dl-trades"
     ]
     start_download_data(get_args(args))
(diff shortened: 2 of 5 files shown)


xmatthias:
I noticed you changed this to WIP ... 
what do you think is still missing? (to me, this would look good as is - but maybe you've found something else that's missing/wrong? ...)

wadedyck:
I had to pass the candle_type in load_bt_data_*, please see the latest commit. There are also a number of failures in test_backtesting, but they seem to have been failing before these changes.

xmatthias:
Tests in your 2nd commit pass (click green check next to the commit) - so failures will be caused by the last commit.

they are almost certainly happening because `candle_type_def` is not in the default_config used for tests (look at check `conftest.py` for that). 
We do set this during startup - but not all tests do go through configuration (actually almost none does).

freqtrade/commands/data_commands.py
```diff
@@ -64,8 +64,6 @@ def start_download_data(args: Dict[str, Any]) -> None:
     try:
 
         if config.get('download_trades'):
-            if config.get('trading_mode') == 'futures':
-                raise OperationalException("Trade download not supported for futures.")
```
xmatthias, line 68: Please keep this here.

the ONLY reason to use freqtrade with download_trades (`--dl-trades` -> this mode) is on kraken - which is a far way from supporting futures (ccxt does not support kraken futures at the moment).

xmatthias: Please keep this here (including the test for it).

the ONLY reason to use freqtrade in this mode (download_trades  / `--dl-trades`) is on kraken - which is a far way from supporting futures (ccxt does not support kraken futures at the moment).

xmatthias (changes requested):

freqtrade/commands/data_commands.py
wadedyck: Sorry, I don't understand the reference to the kraken exchange.... If my trading mode is futures, I'd want to download futures, no? Shouldn't any guards look at _supported_trading_mode_collateral_pairs for the individual specified exchange? I'm probably missing something obvious as I'm still very much learning this codebase.

BTW, I'm trying to get this working with binance and bybit. Bybit is problematic as it doesn't align with the funding ohlcv approach; no fetch_funding_history or 8h timeframes.

xmatthias: The area the above "removal" is in (the if-branch) will not run for binance (nor for bybit, as far as i know) - unless you're using a wrong configuration or `--dl-trades` (which you should not use).

The reference about kraken refers to this segment of [the docs](https://www.freqtrade.io/en/stable/data-download/#trades-tick-data) - which definitely does not apply to binance - and should not apply to bybit (i think they do provide historic ohlcv data).

wadedyck: Got it, thanks!

tests/commands/test_commands.py
```diff
@@ -808,6 +808,7 @@ def test_download_data_trades(mocker, caplog):
         "--exchange", "kraken",
         "--pairs", "ETH/BTC", "XRP/BTC",
         "--days", "20",
+        "--trading-mode", "spot",
```
xmatthias: I'm not sure that's necessary (or even desired).

Default for this is spot - and we should assume that users will not provide the argument if it ain't required.
I've therefore removed this - as that'll represent a "more realistic" test. Rest looks good - will be merged later today.

xmatthias (approved):
LGTM :+1:

PR #6140, opened 2021-12-30
Change sequence of ROI/sell signal to favor sell-signal

### LINKED ISSUE #6139: Backtest sells via ROI instead of sell-signal in case of simultaneous occurrence
## Describe your environment

  * Operating system: macOS 11.6.2
  * Python Version: 3.7.9
  * CCXT version: 1.61.92
  * Freqtrade Version: 2021.11
  
## Describe the problem:

Backtest sells on the simultaneous occurrence of a sell-signal and a ROI signal via the ROI signal. 

### Steps to reproduce:

  1. Run an arbitrary backtest that produces a sell-signal and triggers the ROI at the same time.
  2. Plotting helps to visualise the issue.
  
### Observed Results:

  * What happened?
  Backtest trade sold by ROI instead of the sell signal.
Ohne Titel 2
  Backtest trade should be sold at the 'open' price, as the process time is ideally very small and this value is closer to reality; the ROI execution takes place intra-candle.

### Relevant code

https://github.com/freqtrade/freqtrade/blob/bd98637ae934db16968d921f560e9a05a04b1549/freqtrade/strategy/interface.py#L708-L721

Thanks for your support!

xmatthias:
## Summary
This PR will Reverse the evaluation logic from favoring ROI over sell-signal to preferring sell-signal.

@rokups it would be great to get your input on this - as you've been doing quite some "backtest reality checks" in the recent past.

Closes #6139

## Quick changelog

- Change logic to evaluate sell-signal before ROI
## Reasoning

In live, this will not matter, as the price will still be determined the same way.
In backtesting, sell-signal will sell at "next-candle-open" - while ROI will sell at "roi value" - which might be above the candle open.

While this PR is here - it's a proposal, and i'm not certain yet if it will be merged or not.

For now, it looks like this will be the "conservative" way - more likely selling at a lower rate during backtesting (although it can also be higher if ROI jumped "one step down" at that candle).

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -484,8 +484,8 @@ Since backtesting lacks some detailed information about what happens within a ca
   - ROI applies before trailing-stop, ensuring profits are "top-capped" at ROI if both ROI and trailing stop applies
 - Sell-reason does not explain if a trade was positive or negative, just what triggered the sell (this can look odd if negative ROI values are used)
 - Evaluation sequence (if multiple signals happen on the same candle)
-  - ROI (if not stoploss)
   - Sell-signal
+  - ROI (if not stoploss)
   - Stoploss
 
 Taking these assumptions, backtesting tries to mirror real trading as closely as possible. However, backtesting will **never** replace running a strategy in dry-run mode.
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -703,22 +703,20 @@ class IStrategy(ABC, HyperStrategyMixin):
                             custom_reason = custom_reason[:CUSTOM_SELL_MAX_LENGTH]
                     else:
                         custom_reason = None
-            # TODO: return here if sell-signal should be favored over ROI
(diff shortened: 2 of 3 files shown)


rokups:
I think it does make sense.

ROI aims to sell at highs of the candle, which may not happen up until late into candle. Sell signal sells immediately on next candle open. So it makes sense to process sell signal first as it is what would usually happen in live.

When ROI stepdown happens and we have a sell signal, in live we would sell immediately because both conditions are met (sell signal at higher price + ROI at lower price) and of course we want to sell at a better price. So it also does make sense to do same in backtest. In live we do not know if price is going to tick down or up after candle open, so selling on open price seems like a good middle ground.

xmatthias:
I just realized / remembered that ROI "stepdown" is taken care of already by selling at open:

https://github.com/freqtrade/freqtrade/blob/ee774f12bdf086cff45aa16cdd8ce26d7fe25206/freqtrade/optimize/backtesting.py#L338-L343

ttmdebugger:
Thanks to both of you!

PR #6143, opened 2021-12-31
Todos

samgermain:
Removed a few todos


xmatthias (approved):
I've removed a few more TODO's - 2 of which i've moved to the trello board with the label "nice to have" (like hyperopting leverage - which however is only a documentation task thinking about it), and the pairlist filter to only allow pairs with "matching leverage".

PR #6146, opened 2022-01-01
Sell -> exit

samgermain:
## Summary

Changes the word sell to exit in a lot of places

Didn't change it for `self.strategy.order_time_in_force['sell']` and `exit_type = 'sell'  # TODO-lev: Update to exit` because I wasn't sure that I would get everywhere


xmatthias:
For the test, merge develop and it should pass, i've fixed that this morning (was an oversight which broke this morning).

samgermain:
> For the test, merge develop and it should pass, i've fixed that this morning (was an oversight which broke this morning).

I wasn't sure what to do few a few of the merge conflicts

xmatthias:
i've merged develop to feat/short now - so tests should be fine (* the one that failed this morning) once merge-conflicts are gone.

docs/hyperopt.md
```diff
@@ -206,7 +206,7 @@ Hyper-optimization will, for each epoch round, pick one trigger and possibly mul
 
 #### Sell optimization
```
xmatthias: Might as well change "sell" here? (look out for links to this section)

```suggestion
#### Exit signal optimization
```

docs/hyperopt.md
```diff
@@ -206,7 +206,7 @@ Hyper-optimization will, for each epoch round, pick one trigger and possibly mul
...
 
-Similar to the buy-signal above, sell-signals can also be optimized.
+Similar to the buy-signal above, exit-signals can also be optimized.
```
xmatthias:
```suggestion
Similar to the entry-signal above, exit-signals can also be optimized.
```

freqtrade/freqtradebot.py
```diff
@@ -1314,17 +1312,17 @@ def execute_trade_exit(
...
+            order_type = self.strategy.order_types.get("emergencyexit", "market")
 
         amount = self._safe_exit_amount(trade.pair, trade.amount)
         time_in_force = self.strategy.order_time_in_force['sell']  # TODO-lev update to exit
 
         if not strategy_safe_wrapper(self.strategy.confirm_trade_exit, default_retval=True)(
                 pair=trade.pair, trade=trade, order_type=order_type, amount=amount, rate=limit,
-                time_in_force=time_in_force, sell_reason=sell_reason.sell_reason,
-                current_time=datetime.now(timezone.utc)):  # TODO-lev: Update to exit
+                time_in_force=time_in_force, exit_reason=exit_reason.exit_reason,
```
xmatthias, line 1324: We will want to avoid hard breaking changes if possible.

This change breaks every strategy that implements the current interface by no longer providing `sell_reason` - but only once this callback is actually called (so potentially, days into running the new bot version).

One way i can imagine is to provide both `sell_reason` and `exit_reason` for some releases  - marking sell_reason (and whatever else changed) as deprecated in the deprecation docs.

freqtrade/strategy/interface.py
```diff
@@ -27,23 +27,23 @@
 
 
 logger = logging.getLogger(__name__)
-CUSTOM_SELL_MAX_LENGTH = 64
+CUSTOM_EXIT_MAX_LENGTH = 64
 
 
-class SellCheckTuple:
+class ExitCheckTuple:
```
xmatthias, line 33: Doublecheck if this is a public interface - in which case we'll want to cleanly deprecate and have a grace period.

freqtrade/strategy/interface.py
```diff
@@ -101,10 +101,10 @@ class IStrategy(ABC, HyperStrategyMixin):
...
     process_only_new_candles: bool = False
 
-    use_sell_signal: bool
-    sell_profit_only: bool
-    sell_profit_offset: float
-    ignore_roi_if_buy_signal: bool
+    use_exit_signal: bool
+    exit_profit_only: bool
+    exit_profit_offset: float
+    ignore_roi_if_enter_signal: bool
```
xmatthias, line 107: I am quite concerned about settings that are part of the strategy.

While you mark them as "deprecated" when coming from the configuration - settings like `use_sell_signal = True`  in the strategy are simply ignored.

That's the most dangerous of all approaches - as users will have a working strategy - and an update will silently ignore certain values - which may be required for the strategy to work.

xmatthias (changes requested):
I think this PR changes too much - there's some settings / fields that can be changed quite easily (settings that only appear in the configuration, internal stuff) - while others (e.g. callbacks, settings that are loaded from the strategy) will require quite some attention and care to have a smooth upgrade experience for users - which is quite difficult to get in a PR that's changing 65 files.

---

I've only looked through a little bit - unfortunately, this PR seems to blindly change occurrences of "sell" to "exit" without much thought on compatibility to existing strategies.


Thinking forward: a simple bot update (say - `install.sh -u`) - which will update to the first version that includes the `feat/short` branch should not break existing functionality / strategies, not make hard changes (which either silently break the strategy by no longer calling/respecting certain things, or make the strategy load fail).

If any - then a "no longer corresponding" strategy should fail to load with sensible warnings - but for the moment, i didn't see anything that would make that a neccessity.

xmatthias:
For the moment, i feel like this is a "nice to have" PR - not something we'll need to have the feat/short branch working.

As such i'd suggest to: 
* focus on necessary things (like liquidation calculations, ... )
* eventually update TODO-lev comments which impact naming only to regular TODO (or some other naming) - to be able to clearly filter these out when looking at "what's still to do".

While it makes sense to have naming aligned eventually - None of these changes are necessary to have `feat/short` functional / working for users (which would mean we can merge it to develop) - and these changes can be done further down the road (once people are using the functionalities in question).

They can also be done in a slow way over a couple of releases (once feat/short is publicly available) (not as one huge PR - which is near impossible to review without missing 10s of things).

samgermain:
I created a Trello card https://trello.com/c/G5yrcRCr

PR #6168, opened 2022-01-05
Futures dust warning

samgermain:


freqtrade/freqtradebot.py
```diff
@@ -1540,9 +1540,16 @@ def apply_fee_conditional(self, trade: Trade, trade_base_currency: str,
         Can eat into dust if more than the required asset is available.
         """
         self.wallets.update()
-        if fee_abs != 0 and self.wallets.get_free(trade_base_currency) >= amount:
+        free_base = self.wallets.get_free(trade_base_currency)
+        if fee_abs != 0 and free_base >= amount:
+            if self.trading_mode == TradingMode.FUTURES:
```
xmatthias: based on the fact that futures can't produce dust - this code is unreachable - and therefore unnecessary.
If we were to add such warnings to everything - we'd have the code-base cluttered - as for most every condition, there'll be a case which cannot reach that point.

I'd rather simply remove the "TODO-lev" - rather than introducing additional code to handle something we assume cannot happen.

PR #6172, opened 2022-01-06
Backtest result reuse

rokups:
## Summary

Implemented reuse of old backtest results when backtest parameters did not change.

## Quick changelog

- Avoid running a backtest and instead reuse old result if backtest parameters did not change.
- Extra file `backtest-result-*.meta.json` is written with backtest run id (hash derived from config + strategy file).
- New command line parameter `--no-cache` for `backtest` subcommand, opting out from cached result reuse.

## What's new?

When we develop a new version of strategy we usually backtest using `--strategy-list` so we can compare against previous versions. In such case we most likely backtest multiple times using same config and old strategy files do not change. This means we can reuse old backtest result instead of redoing same backtest over and over. This is what was implemented here.

### DIFF
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -205,6 +205,11 @@ AVAILABLE_CLI_OPTIONS = {
         nargs='+',
         choices=constants.BACKTEST_BREAKDOWNS
     ),
+    "no_backtest_cache": Arg(
+        '--no-cache',
+        help='Do not reuse cached backtest results.',
+        action='store_true'
+    ),
     # Edge
     "stoploss_range": Arg(
         '--stoplosses',
--- a/freqtrade/data/btanalysis.py
+++ b/freqtrade/data/btanalysis.py
@@ -2,6 +2,7 @@
 Helpers when analyzing backtest data
 """
 import logging
+from copy import copy
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -10,7 +11,7 @@ import pandas as pd
 
 from freqtrade.constants import LAST_BT_RESULT_FN
 from freqtrade.exceptions import OperationalException
-from freqtrade.misc import json_load
+from freqtrade.misc import get_backtest_metadata_filename, json_load
 from freqtrade.persistence import LocalTrade, Trade, init_db
 
 
@@ -102,6 +103,23 @@ def get_latest_hyperopt_file(directory: Union[Path, str], predef_filename: str =
     return directory / get_latest_hyperopt_filename(directory)
 
 
+def load_backtest_metadata(filename: Union[Path, str]) -> Dict[str, Any]:
+    """
+    Read metadata dictionary from backtest results file without reading and deserializing entire
+    file.
+    :param filename: path to backtest results file.
+    :return: metadata dict or None if metadata is not present.
+    """
+    filename = get_backtest_metadata_filename(filename)
+    try:
+        with filename.open() as fp:
+            return json_load(fp)
+    except FileNotFoundError:
+        return {}
+    except Exception as e:
+        raise OperationalException('Unexpected error while loading backtest metadata.') from e
+
+
 def load_backtest_stats(filename: Union[Path, str]) -> Dict[str, Any]:
     """
     Load backtest statistics file.
@@ -118,7 +136,54 @@ def load_backtest_stats(filename: Union[Path, str]) -> Dict[str, Any]:
     with filename.open() as file:
         data = json_load(file)
 
+    # Legacy list format does not contain metadata.
+    if isinstance(data, dict):
+        data['metadata'] = load_backtest_metadata(filename)
+
     return data
+
+
+def find_existing_backtest_stats(dirname: Union[Path, str],
+                                 run_ids: Dict[str, str]) -> Dict[str, Any]:
+    """
+    Find existing backtest stats that match specified run IDs and load them.
+    :param dirname: pathlib.Path object, or string pointing to the file.
+    :param run_ids: {strategy_name: id_string} dictionary.
+    :return: results dict.
+    """
+    # Copy so we can modify this dict without affecting parent scope.
+    run_ids = copy(run_ids)
+    dirname = Path(dirname)
+    results: Dict[str, Any] = {
+        'metadata': {},
+        'strategy': {},
+        'strategy_comparison': [],
+    }
+
+    # Weird glob expression here avoids including .meta.json files.
+    for filename in reversed(sorted(dirname.glob('backtest-result-*-[0-9][0-9].json'))):
+        metadata = load_backtest_metadata(filename)
+        if not metadata:
+            # Files are sorted from newest to oldest. When file without metadata is encountered it
+            # is safe to assume older files will also not have any metadata.
+            break
+
+        for strategy_name, run_id in list(run_ids.items()):
+            if metadata.get(strategy_name, {}).get('run_id') == run_id:
+                # TODO: load_backtest_stats() may load an old version of backtest which is
+                #  incompatible with current version.
+                del run_ids[strategy_name]
+                bt_data = load_backtest_stats(filename)
+                for k in ('metadata', 'strategy'):
+                    results[k][strategy_name] = bt_data[k][strategy_name]
+                comparison = bt_data['strategy_comparison']
+                for i in range(len(comparison)):
+                    if comparison[i]['key'] == strategy_name:
+                        results['strategy_comparison'].append(comparison[i])
+                        break
+        if len(run_ids) == 0:
+            break
+    return results
 
 
 def load_backtest_data(filename: Union[Path, str], strategy: Optional[str] = None) -> pd.DataFrame:
--- a/freqtrade/misc.py
+++ b/freqtrade/misc.py
@@ -2,11 +2,13 @@
 Various tool function for Freqtrade and scripts
 """
 import gzip
+import hashlib
 import logging
 import re
+from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Iterator, List
+from typing import Any, Iterator, List, Union
 from typing.io import IO
 from urllib.parse import urlparse
 
@@ -228,3 +230,32 @@ def parse_db_uri_for_logging(uri: str):
         return uri
     pwd = parsed_db_uri.netloc.split(':')[1].split('@')[0]
     return parsed_db_uri.geturl().replace(f':{pwd}@', ':*****@')
+
+
+def get_strategy_run_id(strategy) -> str:
+    """
+    Generate unique identification hash for a backtest run. Identical config and strategy file will
+    always return an identical hash.
+    :param strategy: strategy object.
+    :return: hex string id.
+    """
+    digest = hashlib.sha1()
+    config = deepcopy(strategy.config)
+
+    # Options that have no impact on results of individual backtest.
+    not_important_keys = ('strategy_list', 'original_config', 'telegram', 'api_server')
+    for k in not_important_keys:
+        if k in config:
+            del config[k]
+
+    digest.update(rapidjson.dumps(config, default=str,
+                                  number_mode=rapidjson.NM_NATIVE).encode('utf-8'))
+    with open(strategy.__file__, 'rb') as fp:
+        digest.update(fp.read())
+    return digest.hexdigest().lower()
+
+
+def get_backtest_metadata_filename(filename: Union[Path, str]) -> Path:
+    """Return metadata filename for specified backtest results file."""
+    filename = Path(filename)
+    return filename.parent / Path(f'{filename.stem}.meta{filename.suffix}')
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -753,6 +753,7 @@ class Backtesting:
         }
 
         # Load previous result that will be updated incrementally.
+        # This can be circumvented in certain instances in combination with downloading more data
         if self.timerange.stopts == 0 or datetime.fromtimestamp(
            self.timerange.stopts, tz=timezone.utc) > datetime.now(tz=timezone.utc):
             self.config['no_backtest_cache'] = True
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -753,7 +753,8 @@ class Backtesting:
         }
 
         # Load previous result that will be updated incrementally.
-        if self.config.get('timerange', '-').endswith('-'):
+        if self.timerange.stopts == 0 or datetime.fromtimestamp(
+           self.timerange.stopts, tz=timezone.utc) > datetime.now(tz=timezone.utc):
             self.config['no_backtest_cache'] = True
             logger.warning('Backtest result caching disabled due to use of open-ended timerange.')
 
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -14,12 +14,13 @@ from pandas import DataFrame
 from freqtrade.configuration import TimeRange, validate_config_consistency
 from freqtrade.constants import DATETIME_PRINT_FORMAT
 from freqtrade.data import history
-from freqtrade.data.btanalysis import trade_list_to_dataframe
+from freqtrade.data.btanalysis import find_existing_backtest_stats, trade_list_to_dataframe
 from freqtrade.data.converter import trim_dataframe, trim_dataframes
 from freqtrade.data.dataprovider import DataProvider
 from freqtrade.enums import BacktestState, SellType
 from freqtrade.exceptions import DependencyException, OperationalException
 from freqtrade.exchange import timeframe_to_minutes, timeframe_to_seconds
+from freqtrade.misc import get_strategy_run_id
 from freqtrade.mixins import LoggingMixin
 from freqtrade.optimize.bt_progress import BTProgress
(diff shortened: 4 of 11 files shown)


freqtrade/commands/cli_options.py
```diff
@@ -205,6 +205,11 @@ def __init__(self, *args, **kwargs):
         nargs='+',
         choices=constants.BACKTEST_BREAKDOWNS
     ),
+    "no_backtest_cache": Arg(
+        '--no-cache',
+        help='Do not reuse a cached backtest result even if it exists.',
```
xmatthias:
```suggestion
        help='Do not reuse cached backtest results.',
```
the "even if it exists" sounds strange when reading it out of context in a help message (think - backtesting for the first time)

freqtrade/data/btanalysis.py
```diff
@@ -121,9 +141,54 @@ def load_backtest_stats(filename: Union[Path, str]) -> Dict[str, Any]:
...
+        ('metadata', {}),
+        ('strategy', {}),
+        ('strategy_comparison', []),
+    ])
+
+    for filename in reversed(sorted(dirname.iterdir())):
+        if (not filename.name.startswith('backtest-result') or
+            not filename.name.endswith('.json') or
+            filename.name.endswith('.meta.json')):
+            continue
```
xmatthias: I think you should be able to use `glob()` here - something like `dirname.glob('backtest-result.*meta.json)` ... which would simplify this whole segment to just one line

freqtrade/data/btanalysis.py
```diff
@@ -121,9 +141,54 @@ def load_backtest_stats(filename: Union[Path, str]) -> Dict[str, Any]:
...
+
+        metadata = load_backtest_metadata(filename)
+        if not metadata:
+            # Files are sorted from newest to oldest. When file without metadata is encountered it
+            # is safe to assume older files will also not have any metadata.
+            continue
```
xmatthias: Based on the comment above - i think this should be `break` (stop further processing of the loop)- not continue (continue with next result).

freqtrade/misc.py
```diff
@@ -228,3 +230,24 @@ def parse_db_uri_for_logging(uri: str):
...
+    digest.update(rapidjson.dumps(strategy.config, default=str,
+                                  number_mode=rapidjson.NM_NATIVE).encode('utf-8'))
+    with open(strategy.__file__, 'rb') as fp:
+        digest.update(fp.read())
+    return digest.hexdigest().lower()
+
+
+def get_backtest_metadata_filename(filename: Union[Path, str]) -> Path:
+    """Return metadata filename for specified backtest results file."""
+    base, ext = os.path.splitext(filename)
```
xmatthias: I would rather stick to pathlib - which should have this available as `filename.suffix` (obviously assuming the input is Path ...)

os.path can have ... questionable behavior on certain platforms (windows, i'm looking at you) - so i don't trust it for anything related to paths anymore

rokups:
Addressed all comments 👍🏻 btw i wasnt sure if `--no-cache` is a good name, but i could not come up with anything better..

freqtrade/data/btanalysis.py
```diff
@@ -105,6 +108,23 @@ def get_latest_hyperopt_file(directory: Union[Path, str], predef_filename: str =
...
+    file.
+    :param filename: path to backtest results file.
+    :return: metadata dict or None if metadata is not present.
+    """
+    filename = get_backtest_metadata_filename(filename)
+    try:
+        with filename.open() as fp:
+            return json_load(fp)
+    except FileNotFoundError:
+        return {}
```
xmatthias, line 118: The Premise of this looks good - but it doesn't seem to work at the moment

I've just ran 3 backtests in a row (all produce the same hash based on the .meta file) - but it doesns't work and always runs into this.
Some debugging shows that the "loading" filename is generated wrongly.
I think for loading, you'll not need to call `get_backtest_metadata_filename` as the ".glob" already returns the correct filename.

Removing that call does load the file - but crashes further down ...

rokups:
Fixing those 👍🏻 

Also bumped into this:
```
2022-01-07 12:22:34,507 - freqtrade - ERROR - Fatal exception!
  File "/home/rk/src/freqtrade/freqtrade/main.py", line 37, in main
  File "/home/rk/src/freqtrade/freqtrade/commands/optimize_commands.py", line 54, in start_backtesting
    backtesting.start()
  File "/home/rk/src/freqtrade/freqtrade/optimize/backtesting.py", line 740, in start
    show_backtest_results(self.config, self.results)
  File "/home/rk/src/freqtrade/freqtrade/optimize/optimize_reports.py", line 826, in show_backtest_results
    table = text_table_strategy(backtest_stats['strategy_comparison'], stake_currency)
  File "/home/rk/src/freqtrade/freqtrade/optimize/optimize_reports.py", line 662, in text_table_strategy
    drawdown = [f'{t["max_drawdown_account"] * 100:.2f}' for t in strategy_results]
  File "/home/rk/src/freqtrade/freqtrade/optimize/optimize_reports.py", line 662, in <listcomp>
    drawdown = [f'{t["max_drawdown_account"] * 100:.2f}' for t in strategy_results]
KeyError: 'max_drawdown_account'
```

Cached backtest result did not have `max_drawdown_account` entry. This isnt a problem for now, but we may consider adding a backtest format version, then re-running of a backtest could be forced when version changes. For now i will add a comment.

Edit: Actually it is a problem. This exception is thrown by `backtesting-show` command when trying to access old result.

rokups:
I added a test, but it fails. For some reason `mocker.patch('freqtrade.misc', get_strategy_run_id=get_strategy_run_id)` fails to mock that function. Any idea why that would be?

xmatthias:
> I added a test, but it fails. For some reason `mocker.patch('freqtrade.misc', get_strategy_run_id=get_strategy_run_id)` fails to mock that function. Any idea why that would be?

wrong syntax - you're using `mocker.patch.multi` syntax on `mocker.patch` ...

try `mocker.patch('freqtrade.misc.get_strategy_run_id', get_strategy_run_id)` (assuming get_strategy_run_id is a Mock object).

> Edit: Actually it is a problem. This exception is thrown by backtesting-show command when trying to access old result.

I'll fix that in develop tonight ... :+1:

rokups:
Right syntax still fails

tests/optimize/test_backtesting.py
```diff
@@ -1239,3 +1239,88 @@ def test_backtest_start_multi_strat_nomock_detail(default_conf, mocker,
...
+            'strategy': {'TestStrategyLegacyV1': {}},
+            'strategy_comparison': [{'key': 'TestStrategyLegacyV1'}]
+        }
+    ])
+    get_strategy_run_id = MagicMock(side_effect=['1', '2', '2'])
+    mocker.patch('pathlib.Path.glob', path_glob)
+    mocker.patch.multiple('freqtrade.data.btanalysis',
+                          load_backtest_metadata=load_backtest_metadata,
+                          load_backtest_stats=load_backtest_stats)
+    mocker.patch('freqtrade.misc.get_strategy_run_id', get_strategy_run_id)
```
xmatthias: this mock is "missing" - which means it doesn't apply correctly.

When mocking a directly imported function (like get_strategy_run_id) - you'll usually have to replace it at the import location (freqtrade.optimize.backtesting - not where it's defined).

That mock is however also not applying correctly - most likely because the test-file is importing backtesting stuff before the mock is actually running (which sometimes requires odd workarounds).

I remember i had this once before which led me to move the test to another file - but i can't remember which test that was - probably one that imports within the test function instead of the top of the file).

freqtrade/data/btanalysis.py
```diff
@@ -118,9 +137,56 @@ def load_backtest_stats(filename: Union[Path, str]) -> Dict[str, Any]:
...
+    """
+    Find existing backtest stats that match specified run IDs and load them.
+    :param dirname: pathlib.Path object, or string pointing to the file.
+    :param run_ids: {strategy_name: id_string} dictionary.
+    :return: results dict.
+    """
+    # Copy so we can modify this dict without affecting parent scope.
+    run_ids = copy(run_ids)
+    dirname = Path(dirname)
+    results: Dict[str, Any] = OrderedDict([
```
xmatthias: I don't think a OrderedDict is necessary.
You're not using any "enhanced" functionality of OrderedDict - and regular dics are now also ordered by default (since 3.7 iirc).

tests/optimize/test_backtesting.py
```diff
@@ -1239,3 +1239,88 @@ def test_backtest_start_multi_strat_nomock_detail(default_conf, mocker,
...
+        'results': pd.DataFrame(columns=BT_DATA_COLUMNS),
+        'config': default_conf,
+        'locks': [],
+        'rejected_signals': 20,
+        'final_balance': 1000,
+    })
+    mocker.patch('freqtrade.plugins.pairlistmanager.PairListManager.whitelist',
+                 PropertyMock(return_value=['UNITTEST/BTC']))
+    mocker.patch('freqtrade.optimize.backtesting.Backtesting.backtest', backtestmock)
+    mocker.patch('freqtrade.optimize.backtesting', show_backtest_result=MagicMock())
```
xmatthias: The Failure was actually CAUSED by this (after moving the below mock to the lookup location in `.optimize.backtesting`).

This will mess with the whole backtesting module (everything in that file, most likely) - causing subsequent mocks to somehow not work correctly (not sure how/why) - but fixing this mock fixed the test for no apparent other reason.

rokups: Thats right. Its a leftover from attempt to store metadata in same file. Removed that 👍🏻

rokups:
Actually this needs one more fix before merge: before hashing a config we need to unset strategy list.

freqtrade/optimize/optimize_reports.py
```diff
@@ -509,12 +516,22 @@ def generate_backtest_stats(btdata: Dict[str, DataFrame],
...
+        ('strategy_comparison', []),
+    ])
     market_change = calculate_market_change(btdata, 'close')
+    metadata = result['metadata']
     pairlist = list(btdata.keys())
     for strategy, content in all_results.items():
         strat_stats = generate_strategy_stats(pairlist, strategy, content,
                                               min_date, max_date, market_change=market_change)
+        # Warning: due to implementation details of get_backtest_metadata() function, characters
+        # { and } appearing in strings will break stuff.
```
xmatthias: It's unclear what you mean with this warning (it refers to a method that does not exist).
Is this because of `get_backtest_metadata_filename` - or `strategy_run_id` (or something else)

Also - can you make an example what would need to be the input for this (whatever function?) to break?

freqtrade/optimize/optimize_reports.py
```diff
@@ -509,12 +516,22 @@ def generate_backtest_stats(btdata: Dict[str, DataFrame],
     :param max_date: Backtest end date
     :return: Dictionary containing results per strategy and a strategy summary.
     """
-    result: Dict[str, Any] = {'strategy': {}}
+    result: Dict[str, Any] = OrderedDict([
```
xmatthias: wouldn't a simple dict suffice, conisidering they're ordered by default since python 3.6 or 7? (and you don't even care about the sequence of the keys)

freqtrade/optimize/optimize_reports.py
```diff
@@ -509,12 +516,22 @@ def generate_backtest_stats(btdata: Dict[str, DataFrame],
...
+        ('metadata', {}),
+        ('strategy', {}),
+        ('strategy_comparison', []),
+    ])
     market_change = calculate_market_change(btdata, 'close')
+    metadata = result['metadata']
```
xmatthias: i honestly find this assignment slightly missleading, and dangerous to introduce bugs in the future.

I understand it's just a "convenience" to not have to use `result['metadata']['strategy'] =` below - but for someone not 100% aware that there will only be one metadata dict this might be confusing i think.

freqtrade/optimize/backtesting.py
```diff
@@ -691,15 +693,45 @@ def start(self) -> None:
...
+                data, self.all_results, min_date=min_date, max_date=max_date)
+            if self.results:
+                self.results['metadata'].update(results['metadata'])
+                self.results['strategy'].update(results['strategy'])
+                self.results['strategy_comparison'].extend(results['strategy_comparison'])
+            else:
+                self.results = results
 
             if self.config.get('export', 'none') == 'trades':
                 store_backtest_stats(self.config['exportfilename'], self.results)
```
xmatthias, line 785: Just for my understanding, maybe ...

assuming i backtested `StrategyA` yesterday, and today i use `--strategy-list StrategyA StrategyB` - (nothing else changed, so `StrategyA` is loaded from the prior result) ... 

do we intentionally write the result of StrategyA to disk again? (so it's now in 2 result files)

(maybe to clarify):
* either we want it to write to file again - in which case writing should happen AFTER sorting below.
* or we don't want to write it again - in which case it should happen before combining results

I assume the prior is the right answer (?)

freqtrade/optimize/optimize_reports.py
rokups: Leftover from previous hacky parsing of json - removing 👍🏻

freqtrade/optimize/optimize_reports.py
rokups: Seems like i switched it to dict and back again, opss. Yeah order is also not important, it used to be important with a hacky metadata parsing from same json file as a result (that is no longer there).

freqtrade/optimize/backtesting.py
rokups: I suspect `backtesting-show` would break if we did not copy old backtest result to new result file. But more importantly copying keeps results consistent in the event where user may delete some but not all old backtest results. If it were the case, in example you described user may lose half of the result even though file of that backtest result is present.

samgermain:
Just from quickly glancing at this, the [weighted interval scheduling](https://imgur.com/a/mVC33Uq) problem rings a bell, I thought I would share this in case it's useful

freqtrade/optimize/backtesting.py
```diff
@@ -691,15 +693,45 @@ def start(self) -> None:
...
 
+        # Results may be mixed up now. Sort them so they follow --strategy-list order.
+        if 'strategy_list' in self.config and len(self.results) > 0:
+            def strategy_list_key(c):
+                return self.config['strategy_list'].index(c['key'])
+
+            self.results['strategy_comparison'] = sorted(
+                self.results['strategy_comparison'], key=strategy_list_key)
```
rokups: Looks like dicts need sorting too. So do not merge just yet..

rokups:
For time being im done. Let me know if there is anything else that needs to be done. 👍🏻

freqtrade/optimize/backtesting.py
```diff
@@ -691,15 +693,50 @@ def start(self) -> None:
         self.load_bt_data_detail()
         logger.info("Dataload complete. Calculating indicators")
 
+        run_ids = {
+            strategy.get_strategy_name(): get_strategy_run_id(strategy)
+            for strategy in self.strategylist
+        }
+
+        # Load previous result that will be updated incrementally.
+        if self.config.get('timerange', '-').endswith('-'):
```
xmatthias: To me, this check is potentially wrong / erroneous

you should rather use `self.timerange` and evaluate if the end-date is present and not after today.

consider the following "workaround":
`--timerange 20220101-20230101` - it would pass your test here just fine (doesn't end with "-") - However - my end-date is 11 months in the future, so i'll be using incremental data "at will".

freqtrade/optimize/backtesting.py
```diff
@@ -745,15 +747,51 @@ def start(self) -> None:
...
         logger.info("Dataload complete. Calculating indicators")
 
+        run_ids = {
+            strategy.get_strategy_name(): get_strategy_run_id(strategy)
+            for strategy in self.strategylist
+        }
+
+        # Load previous result that will be updated incrementally.
+        if self.timerange.stopts == 0 or datetime.fromtimestamp(
+           self.timerange.stopts, tz=timezone.utc) > datetime.now(tz=timezone.utc):
```
xmatthias, line 758: I think your own analysis was quite good - this can be worked around eventually (i'm however not sure how to protect from it).

using `timerange` from above as comparison will not protect aganst this:

* backtest with `--timerange 20210101-20220101` (full prior year) with data ending 2021-11-01.
* download data from November until today
* run the same backtest again `--timerange 20210101-20220101`

By comparing the 2 timerange objects (timerange - self.timerange) - you could protect against the first case (not loading cache) - but you'll not know that you had 2 months missing in the "cached" result (nothing in the config or strategy changed - other than the messages at startup).


TBH - i'm not sure we must protect against this ... we can also add a hint in the documentation about this, warning users to use `--no-cache` when the underlying data changed.

xmatthias (commented):
Nice enhancement :+1: 

LGTM :+1:

xmatthias (approved):

rokups:
I had another idea yesterday. Maybe instead of `--no-cache` we should have had `--cache=[none,day,week,month]`? Then cached results from specified period would be loaded, but no older. This would essentially create an expiring cache and help mitigating issues with data updates invalidating cached results.

xmatthias:
it'd be a possibility for sure - but which of the above should be the default? (Potentially none?)

rokups:
`none` or `day` sound like a good idea. `none` keeps current behavior while `day` is a safest form of caching which saves most time. Cant say i have a preference.

xmatthias:
`day` works fine i think ... although it "bite" me already while trying to change something in backtesting (where i wanted to run the same backtest multiple times in a row) - i think that's a "developer edgecase" we should not design our defaults against 😆

PR #6185, opened 2022-01-08
cleanup funding fee tests

xmatthias:
## Summary
While working on funding-fee implementation of backtesting, i noticed that the calculation is reversed.

Funding fees will eat into profit (lower profit) - so on a long, the (fictive) sell price must be lowered, while on a short, it must be raised (it'll negatively impact profit).

@samgermain would be good to see your opinion on this - as this calc was initially your contribution

### DIFF
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -728,6 +728,8 @@ class LocalTrade():
 
         elif (trading_mode == TradingMode.FUTURES):
             funding_fees = self.funding_fees or 0.0
+            # Positive funding_fees -> Trade has gained from fees.
+            # Negative funding_fees -> Trade had to pay the fees.
             if self.is_short:
                 return float(self._calc_base_close(amount, rate, fee)) - funding_fees
             else:
--- a/tests/test_persistence.py
+++ b/tests/test_persistence.py
@@ -549,11 +549,11 @@ def test_update_market_order(market_buy_order_usdt, market_sell_order_usdt, fee,
         ("kraken", True, 3, 59.850, 66.231165, -6.381165, -0.3198578, margin, 0.0),
 
         ("binance", False, 1, 60.15, 65.835,  5.685, 0.09451371, futures, 0.0),
-        ("binance", False, 1, 60.15, 64.835,  4.685, 0.07788861, futures, 1.0),
+        ("binance", False, 1, 60.15, 66.835,  6.685, 0.11113881, futures, 1.0),
         ("binance", True, 1, 59.85,  66.165, -6.315, -0.10551378, futures, 0.0),
-        ("binance", True, 1, 59.85,  65.165, -5.315, -0.08880535, futures, -1.0),
-        ("binance", False, 3, 60.15, 66.835,  6.685, 0.33341646, futures, -1.0),
-        ("binance", True, 3, 59.85,  67.165, -7.315, -0.36666667, futures, 1.0),
+        ("binance", True, 1, 59.85,  67.165, -7.315, -0.12222222, futures, -1.0),
+        ("binance", False, 3, 60.15, 64.835,  4.685, 0.23366583, futures, -1.0),
+        ("binance", True, 3, 59.85,  65.165, -5.315, -0.26641604, futures, 1.0),
     ])
 @pytest.mark.usefixtures("init_persistence")
 def test_calc_open_close_trade_price(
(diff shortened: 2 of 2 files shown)


samgermain:
> ## Summary
> 
> While working on funding-fee implementation of backtesting, i noticed that the calculation is reversed.
> 
> Funding fees will eat into profit (lower profit) - so on a long, the (fictive) sell price must be lowered, while on a short, it must be raised (it'll negatively impact profit).
> 
> @samgermain would be good to see your opinion on this - as this calc was initially your contribution

They don't always eat into profit, it's 50/50 depending on your position, and they're exchanged between traders

> When the funding rate is positive, the price of the perpetual contract is higher than the mark price, thus, traders who are long pay for short positions. Conversely, a negative funding rate indicates that perpetual prices are below the mark price, which means that short positions pay for longs. 

[source](https://www.binance.com/en/blog/futures/a-beginners-guide-to-funding-rates-421499824684900382)

freqtrade/persistence/models.py
```diff
@@ -729,9 +729,9 @@ def calc_close_trade_value(self, rate: Optional[float] = None,
         elif (trading_mode == TradingMode.FUTURES):
             funding_fees = self.funding_fees or 0.0
             if self.is_short:
-                return float(self._calc_base_close(amount, rate, fee)) - funding_fees
-            else:
                 return float(self._calc_base_close(amount, rate, fee)) + funding_fees
+            else:
+                return float(self._calc_base_close(amount, rate, fee)) - funding_fees
```
samgermain, line 734: I believe I had this set up right before, in that if you were receiving funding, then funding_fees would be a positive number, and give you money, but if funding_fees was a negative number it meant you lost money

This makes me think that it might be the same formula for both longs and shorts though, I'm having trouble remembering why I created two branches for short and long

xmatthias: that's a possibility for sure

what concerned me (and why i think this is wrong as is - at least for binance) is that basically ALL (calculated) funding fees were positive in my backtest results.

This means - the difference between `feat/short` and my `backtest_funding` branch on the same strategy was that the one including fees was performing better on (almost) every trade (no matter if it was a long or short ...).

While i can believe that "sometimes" - they'll not eat into profits but benefit me - that'll not be the case in all instances.

the "almost" is an edge-case - where the trade opened and closed between 2 funding periods - which i still need to investigate IF/how funding fees would be applied to that.

---

An alternative i can think of: 
I've been using the dry-run calculation (adding funding * mark candles  * amount together).
Maybe this result has the wrong sign (should be negative) - to align with the return from the "live" endpoint, so the problem is somewhere else (this is something i didn't think about before).

samgermain: > the "almost" is an edge-case - where the trade opened and closed between 2 funding periods - which i still need to investigate IF/how funding fees would be applied to that.

There should be no funding fees exchanged on that trade

xmatthias: well then it's a "correct" edge-case by default :laughing:

samgermain: I think that

```
if self.is_short:
    return float(self._calc_base_close(amount, rate, fee)) + funding_fees
else:
    return float(self._calc_base_close(amount, rate, fee)) - funding_fees
```

is supposed to just be `return float(self._calc_base_close(amount, rate, fee)) + funding_fees`

xmatthias: That sounds incorrect to me.

The input (funding_fees) is not directional (at least not the dry-run calculation) - but simply the result of the calculation of `amount * (mark price * price)` (aggregated over the time the trade was open so far).

the above calculation however is part of `calc_close_trade_value()` - so the return must (correctly) incorporate fees.

---

Assuming a long trade is opened at 100$ - and sold at 105$ - then fees (assuming it cost me) should result in a return from this function of 104.7$ (or however much).

for a short opened at 100$ - and sold at 95$ - the return from this function should correspond to 95.3$

If we're using a "generic" formula here (no difference between long and short) - we'd assume that for all shorts (or longs) - fees are always paid - and for the other side, fees are always received - which - based on the link shared above, does not seem to be the case.

samgermain: > That sounds incorrect to me.
> 
> The input (funding_fees) is not directional (at least not the dry-run calculation) - but simply the result of the calculation of `amount * (mark price * price)` (aggregated over the time the trade was open so far).
> 
> the above calculation however is part of `calc_close_trade_value()` - so the return must (correctly) incorporate fees.
> 
> Assuming a long trade is opened at 100$ - and sold at 105$ - then fees (assuming it cost me) should result in a return from this function of 104.7$ (or however much).
> 
> for a short opened at 100$ - and sold at 95$ - the return from this function should correspond to 95.3$
> 
> If we're using a "generic" formula here (no difference between long and short) - we'd assume that for all shorts (or longs) - fees are always paid - and for the other side, fees are always received - which - based on the link shared above, does not seem to be the case.

Right, now I understand why there's 2 branches. If that's the case then I think the direction was correct before. It might be confusing, but I set it up so that if funding fees was a positive amount, this meant that you gained that money, but if funding fees was a negative amount, then that meant that you lost that money. Funding fees are exchanged between traders, so if one trader pays a funding fee, another trader gains a funding fee.

I think I would have to dig deep to understand the root of the problem you're describing. Do you think you might be able to provide an MRE or some expected input and output?

xmatthias: >  Funding fees are exchanged between traders, so if one trader pays a funding fee, another trader gains a funding fee.

This somehow contradicts the binance link you shared above, which states:

> Funding rates are periodic payments either to traders that are long or short based on the difference between perpetual contract markets and spot prices


Another interesting sentence is the following:

> When the funding rate is positive, the price of the perpetual contract is higher than the mark price, thus, traders who are long pay for short positions. Conversely, a negative funding rate indicates that perpetual prices are below the mark price, which means that short positions pay for longs. 

Based on this, this might be combined into one line after all - but with inverted sign (if funding-rate is positive - we need to subtract it from a long close rate).

`return float(self._calc_base_close(amount, rate, fee)) - funding_fees`


---

As said, it's also possible that the problem lies in how funding_fees for dry-run are calculated (backtesting will reuse the same calculation) - which might be missaligned to what the `fetch_funding_history` endpoint returns.

If you look at funding and mark candles, you'll notice that on binance, for all of december - no value was negative (at least for the pairs i looked at ... / when my backtest was in trades ).

samgermain: > This somehow contradicts the binance link you shared above, which states:

Why do you think it's a contradiction? Can you try to explain what it is you think I'm saying, and what you think the link is saying?

The payments in this quote

> Funding rates are periodic payments either to traders that are long or short based on the difference between perpetual contract markets and spot prices

are made between traders, they go from one trader to another trader, not to the exchange

When funding rates are positive, what happens is that traders in long positions, pay a fee to traders in short positions. What this essentially does is encourages traders to enter short positions before funding payments are made. This increases the demand for shorts/decreases the demand for longs, so when a perpetual contract is priced higher than the spot market, these demand manipulations help to bring the perpetual contract price inline with the spot market price (the mark price is kind of like a mean price for spot markets across top exchanges). This is the whole reason for funding payments, to keep the perpetual contract price similar to the price of the asset in the spot market and it's why the fees are exchanged between traders. Traders can try to either enter or exit positions based on whether they would receive or gain funding fees at the funding times.

-----------------------

If you could provide an example (e.g. run this strategy on this timeframe, look at this column, every single of the 96 funding fees exchanged is positive, you would expect at least one to be negative) then I could probably look at it. I'm dumb and lazy, so if you make it really straight forward that will increase my chances of figuring it out.

xmatthias: > are made between traders, they go from one trader to another trader, not to the exchange

that's exactly why i think it's contradicting - as i don't read that sentence this way.

---

Let's assume positive fees go from `long -> short` (long pays short) - and negative from "short -> long" (short pays long).

That would mean that for a long, we need to reduce the price in case of positive fee (decrease profit), hence
`return float(self._calc_base_close(amount, rate, fee)) - funding_fees`. (e.g. `105 - 0.1`)
for a short with positive funding_fee, we'd need to increase the price (increase profit)  (hence the same calculation applies) - `95 - 0.1` (so the trade goes from 100 to 94.9 -> 5.1)

For negative fees, this would (for a long) come up as `105 - -0.1` - as well for the short (`95 - -0.1`), so the trades would range from 100 to 105.1  (5.1) - and from 100 to 95.1 (4.9)

samgermain: > That would mean that for a long, we need to reduce the price in case of positive fee (decrease profit), hence
return float(self._calc_base_close(amount, rate, fee)) - funding_fees. (e.g. 105 - 0.1)
for a short with positive funding_fee, we'd need to increase the price (increase profit) (hence the same calculation applies) - 95 - 0.1 (so the trade goes from 100 to 94.9 -> 5.1)

It depends on what the mark price is at the time of funding payments
```
if ((the perpetual contract price is 105) and (the mark price is 100)); then
     - traders in long positions will pay funding to traders in short positions
     - this encourages more shorts/less longs and bring the contract price closer to 100
     if (you are in a long position); then
         - the api will show you had a funding "fee" as -X, meaning you lost X amount of currency
     if (you are in a short position); then
         - the api will show you had a funding "fee" as X, meaning you gained X amount of currency
         
 
if ((the perpetual contract price is 95) and (the mark price is 100)); then
     - traders in short positions will pay funding to traders in long positions
     - this encourages more longs/less short and bring the contract price closer to 100
     if (you are in a long position); then
         - the api will show you had a funding "fee" as X, meaning you gained X amount of currency
     if (you are in a short position); then
         - the api will show you had a funding "fee" as -X, meaning you lost X amount of currency
``` 

I feel like you might get it, but just making sure

xmatthias: I did understand this now i think - so all this PR does now is add a comment - and reduce the decimal number in tests.

samgermain: Well the formula we had before was still wrong, this is what it should have been https://github.com/freqtrade/freqtrade/pull/5779#discussion_r786089167

xmatthias: Yes - that's fixed in the backtesting PR #6231 :) 
otherwise i'm fixing the old version - just to then remove that version again.

PR #6189, opened 2022-01-09
Remove non-decimal numerical separators

jay-tau:

## Summary

Prevent confusion due to the presence of digit separators which is not a decimal point. Some locales use `.` to represent the decimal point. Others use `,`.

Solve the issue: #___

## Quick changelog

- Remove non-decimal digit separators

## What's new?


xmatthias (approved):
LGTM :+1: 

especially since the "." was off in one instance (10.0000 is not a valid number ...)

PR #6210, opened 2022-01-11
PercentegeRangeFilter added and constants.py updated

kunthar:
Unit test help will be appreciated. Contributor guide couldn't help much.
Code is PEP8 conformant and linted.

## Summary

Motivation: When you need to pull a definitive price range for pairs between certain 
percentage ratios. Pairs can be selected to whatever percentage  range needed by your strat.

## What's new?
* Version 1.0.1 only works with binance. kucoin, gateio exchanges. 
* More exchanges will be implemented in (uncertain) timeframe.
 
How to use this filter.
1. Add below code under "pairlists": section in your config file:
Example:

    {"method": "PercentegeRangeFilter",
        "min_price_change": -1,
        "max_price_change": 2
    },

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -27,7 +27,7 @@ HYPEROPT_LOSS_BUILTIN = ['ShortTradeDurHyperOptLoss', 'OnlyProfitHyperOptLoss',
                          'SortinoHyperOptLoss', 'SortinoHyperOptLossDaily',
                          'CalmarHyperOptLoss',
                          'MaxDrawDownHyperOptLoss']
-AVAILABLE_PAIRLISTS = ['StaticPairList', 'VolumePairList',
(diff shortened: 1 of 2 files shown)


xmatthias:
What's the difference between this and `RangeStabilityFilter` with corresponding settings (24h, ...)? other than the endpoint used?

You're using the values from `tickers` - but the definition for these fields will differ by exchange (some will use a "rolling" 24h, others fixed "prior day").
Personally, i don't think using tickers for this is a good approach. It's an expensive endpoint (from a rate-limit perspective) - and will not return consistent results across exchanges.

as an alternative, i could imagine a change to `RangeStabilityFilter` to allow configuration of the candle timeframe (like 1h) - so you can do rolling hourly changes - if that's what you're after.

xmatthias:
I'm going to close this as i don't think it provides any benefit over `RangeStabilityFilter`, and if some changes (like a rolling window) are necessary, these should be added to `RangeStabilityFilter` instead.

PR #6213, opened 2022-01-12
Validate pairdata warnings

BillGatesIII:

## Summary

Give no warning if pairdata starts after timerange end or stops before timerange start because then it is not used.

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

### DIFF
--- a/freqtrade/data/history/idatahandler.py
+++ b/freqtrade/data/history/idatahandler.py
@@ -235,15 +235,15 @@ class IDataHandler(ABC):
         :param timerange: Timerange specified for start and end dates
         """
 
-        if timerange.starttype == 'date':
-            start = datetime.fromtimestamp(timerange.startts, tz=timezone.utc)
-            if pairdata.iloc[0]['date'] > start:
-                logger.warning(f"Missing data at start for pair {pair} at {timeframe}, "
+        start = datetime.fromtimestamp(timerange.startts, tz=timezone.utc) if timerange.starttype == 'date' else None
+        stop = datetime.fromtimestamp(timerange.stopts, tz=timezone.utc) if timerange.stoptype == 'date' else None
+        if start is not None:
+            if pairdata.iloc[0]['date'] > start and (stop is None or pairdata.iloc[0]['date'] <= stop):
+                logger.warning(f"Missing data at start for pair {pair} for {timeframe}, "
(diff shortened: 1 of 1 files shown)


xmatthias:
Maybe a stupid question - but what's the purpose of this change? 
Without any context/description it's difficult to determine if potential modifications have been made by mistake, or intentionally to change some behavior.

does it aim to improve code quality? 
or to reduce verbosity in the logs?

BillGatesIII:
Sorry, my fault. I was doing one hunderd things at the same time, had an idea this could be a good change, did want to store it somewhere and of course a Pull Request is not the correct way to do that.

I came to this change because my screen was flooded with messages about pairs not having data when I backtest my strategy for example the months May and June because I always put every new pair in the pairlist. These pairs start of course way later and can be ignored in this backtest for this timerange.

But, that being said, when I think a bit longer about it, Freqtrade doesn't know the pair has no data in that timerange, it only sees there is no data so it gives the warning you should download the data.

Guess a better way to do this is have a switch for backtesting to ignore pairs which have no data in the specified timerange which also might speed up things a little. Unfortunately that goes beyond my knowledge of Freqtrade (at the moment) so I'll close this PR and maybe in some future will try to implement this improvement.

Btw, thanks for this great program :)

xmatthias:
i'm not per se against this if it improves (reduces) verbosity

i'd just like to understand the purpose / what you try to accomplish with it first without trying to guess it out of code changes - as that may lead to a missinterpretation.

BillGatesIII:
> i'm not per se against this if it improves (reduces) verbosity
> 
> i'd just like to understand the purpose / what you try to accomplish with it first without trying to guess it out of code changes - as that may lead to a missinterpretation.

It will reduce verbosity but it might confuse people because they will not get a warning when there is data available but they didn't download it. In an ideal world, Freqtrade knows when the data for a pair starts on the exchange, if we have that information I guess we can even make more improvements (for backtest and hyperopt).

xmatthias:
We don't. that'd be great for downloading data - but exchanges don't provide this information.

On binance, there is a "workaround" to get that information - but we can't (and shouldn't) use it here.

The message's intention is to avert users about "you have missing data at the start". This can mean you used a different timerange than when downloading - or that the pair started later (or you only downloaded part of THAT particular pair)... the details are however unknown.

PR #6228, opened 2022-01-15
Stoploss from open for shorts

clover-es:
## Summary

Add support for shorts in strategy.stoploss_from_open()

Solves # TODO-lev: How would this be calculated for short inside strategy.stoploss_from_open

## Quick changelog

- Add support for shorts in strategy.stoploss_from_open()
- Add short test cases for relevant test.

## What's new?

I don't think there's much more to say.

### DIFF
--- a/freqtrade/strategy/strategy_helper.py
+++ b/freqtrade/strategy/strategy_helper.py
@@ -1,6 +1,5 @@
 import pandas as pd
 
-from freqtrade.exceptions import OperationalException
 from freqtrade.exchange import timeframe_to_minutes
 
 
@@ -81,30 +80,26 @@ def stoploss_from_open(
     The requested stop can be positive for a stop above the open price, or negative for
     a stop below the open price. The return value is always >= 0.
 
-    Returns 0 if the resulting stop price would be above the current price.
+    Returns 0 if the resulting stop price would be above/below (longs/shorts) the current price
 
     :param open_relative_stop: Desired stop loss percentage relative to open price
     :param current_profit: The current profit percentage
+    :param for_short: When true, perform the calculation for short instead of long
     :return: Stop loss value relative to current price
     """
 
-    # formula is undefined for current_profit -1, return maximum value
-    if current_profit == -1:
+    # formula is undefined for current_profit -1 (longs) or 1 (shorts), return maximum value
+    if (current_profit == -1 and not for_short) or (for_short and current_profit == 1):
         return 1
 
     if for_short is True:
-        # TODO-lev: How would this be calculated for short
-        raise OperationalException(
-            "Freqtrade hasn't figured out how to calculated stoploss on shorts")
-        # stoploss = 1-((1+open_relative_stop)/(1+current_profit))
+        stoploss = -1+((1-open_relative_stop)/(1-current_profit))
     else:
         stoploss = 1-((1+open_relative_stop)/(1+current_profit))
 
-    # negative stoploss values indicate the requested stop price is higher than the current price
-    if for_short:
-        return min(stoploss, 0.0)
-    else:
-        return max(stoploss, 0.0)
+    # negative stoploss values indicate the requested stop price is higher/lower
+    # (long/short) than the current price
+    return max(stoploss, 0.0)
 
 
 def stoploss_from_absolute(stop_rate: float, current_rate: float) -> float:
--- a/tests/strategy/test_strategy_helpers.py
+++ b/tests/strategy/test_strategy_helpers.py
@@ -109,33 +109,47 @@ def test_stoploss_from_open():
         [1, 100, 30],
         [100, 10000, 30],
     ]
-    current_profit_range = [-0.99, 2, 30]
+    # profit range for long is [-1, inf] while for shorts is [-inf, 1]
+    current_profit_range_dict = {'long': [-0.99, 2, 30], 'short': [-2.0, 0.99, 30]}
     desired_stop_range = [-0.50, 0.50, 30]
 
-    for open_range in open_price_ranges:
-        for open_price in np.linspace(*open_range):
-            for desired_stop in np.linspace(*desired_stop_range):
+    for side, current_profit_range in current_profit_range_dict.items():
+        for open_range in open_price_ranges:
+            for open_price in np.linspace(*open_range):
+                for desired_stop in np.linspace(*desired_stop_range):
 
-                # -1 is not a valid current_profit, should return 1
-                assert stoploss_from_open(desired_stop, -1) == 1
-
-                for current_profit in np.linspace(*current_profit_range):
-                    current_price = open_price * (1 + current_profit)
-                    expected_stop_price = open_price * (1 + desired_stop)
-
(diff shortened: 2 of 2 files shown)


freqtrade/strategy/strategy_helper.py
```diff
@@ -81,30 +80,26 @@ def stoploss_from_open(
...
     :param open_relative_stop: Desired stop loss percentage relative to open price
     :param current_profit: The current profit percentage
+    :param for_short: When true, perform the calculation for short instead of long
     :return: Stop loss value relative to current price
     """
 
-    # formula is undefined for current_profit -1, return maximum value
-    if current_profit == -1:
+    # formula is undefined for current_profit -1 (longs) or 1 (shorts), return maximum value
+    if (current_profit == -1 and not for_short) or (for_short and current_profit == 1):
```
xmatthias, line 92: current_profit is a a percent value (`0.01` -> 1%).

this is independent of the direction, and will always be positive if you're winning, and negative if you're losing.
a `current_profit` of 1 therefore means it's 100% gain (on a long, or a short). 
While not very probable - that's not something we'd want to exclude.

freqtrade/strategy/strategy_helper.py
```diff
@@ -81,30 +80,26 @@ def stoploss_from_open(
...
         return 1
 
     if for_short is True:
-        # TODO-lev: How would this be calculated for short
-        raise OperationalException(
-            "Freqtrade hasn't figured out how to calculated stoploss on shorts")
-        # stoploss = 1-((1+open_relative_stop)/(1+current_profit))
+        stoploss = -1+((1-open_relative_stop)/(1-current_profit))
```
samgermain, line 96: Can you give some explanation for why this is the formula? Or provide a source?

clover-es: Sure, I deduced it with the following:

Defining:
`current_price = open_price * (1 - current_profit)` (if current profit is positive, current price is lower)
`open_stoploss_price = open_price * (1 - open_stoploss_ratio)` (if open stoploss ratio is negative, current price is higher)
`current_relative_stoploss = 1 - (open_stoploss_price/current_price)`
Given that `current_relative_stoploss` is the ratio of the current price and the stoploss price when the trade was open. So, replacing `current_price` and `open_stoploss_price` with they're definitions:
`
current_relative_stoploss = 1 - ((open_price * (1 - open_stoploss_ratio))/(open_price * (1 - current_profit)))
`
Factoring out `open_price` gives:
`current_relative_stoploss = 1 - ((1 - open_stoploss_ratio) / (1 - current_profit))`

And finally, I change sign of the whole formula:
`current_relative_stoploss = -1 + ((1 - open_stoploss_ratio) / (1 - current_profit))`

This last one may be the most controversial part, but it's needed to respect the previous sign convention established by this function: for longs, `open_stoploss_ratio` is negative when `open_price` is above the `open_stoploss_price`, but this function returns a positive value when `current_price` is above `open_stoploss_price`. So, for shorts, the opposite criteria is taken, negative when the `open_price` is above the `open_stoploss_price`, and returns positive value value when `current_price` is below `open_stoploss_price`.

So, classical example using the short position tool from TradingView:
image
`open_price` = 10000 (grey)
`open_stoploss_price` = 11000 (red) -> `open_stoploss_ratio` = -0.1
`current_price` = 8000 (green) -> `current_profit` = 0.2
Now, in this same tool, calculate stoploss from `current_price` to `open_stoploss_price`:
image

And if I perform this calculation in the formula yields the same value:
`current_relative_stoploss = -1 + ((1 - (-0.1)) / (1 - 0.2)) = 0.375`

Also, the test pass because the same definitions are made in it.

xmatthias (commented):
To be honest, i'm not yet convinced that this helper function needs an explicit "short" mode.

all parameters and return values are ratio/percentage values - where the direction of the calculation will not change (you'll still make 5% profit - even if for a short, the absolute prices are reversed).

The direction (positive, negative) of the return value (the new stoploss) will not matter to freqtrade (even though we seem to have top-capped it to 0 for this function) - and even if it would - it will not change direction between long and short modes.

assuming this returns `X` or `-X`, freqtrade will always place the stop X% *below* the current rate for long, and X% *above* current rate for short trades (we actually call `abs()` on the stoploss value to be sure we have the direction correct).

I'm not saying that the formula is correct or incorrect for shorts - but it's unnecessary in my view - as the function should return the same results for the same inputs - no matter if `for_short` is provided as true or false) - which makes this parameter unnecessary in my view.

---

I know we've had the `for_short` added to this helper function - but it really should've been added to `stoploss_from_absolute()` instead - which is dealing with absolute values (and here, we'll need to potentially handle a stop-rate above current rate differently).

freqtrade/strategy/strategy_helper.py
xmatthias: I think you'll need to be careful with the docstrings here. It may be missleading, and should really be compared with the one in `develop`. While this docstring is in `feat/short` - it may or may not be correct (the changes between develop and feat/short in this case are incorrect in my opinion).

> This last one may be the most controversial part, but it's needed to respect the previous sign convention established by this function

i don't really understand this.
You're talking about prices throughout the above post ... but absolute prices are not relevant for this function (They are for `stoploss_from_absolute()` - but not for `stoploss_from_open()` (which we're talking about here).

`current_profit` will always have the same sign - `0.05` (5%) when you're green - `-0.05` (-5%) when the trade is red - if it's a long or a short won't matter.
`open_relative_stop` is `Desired stop loss percentage relative to open price`.

so the purpose of this function is to simplify implementation of custom stoploss levels.

calling `stoploss_from_open(0.02, current_profit)` will return the correct response to have stoploss 2% above the open rate (this needs to be calculated relative to current profit, hence the helper function).

if you call `stoploss_from_open(0.0, current_profit)` - that's a "break even" function - which will place the stoploss at the open rate (the return value of `custom_stoploss` is the relative difference to current price).

freqtrade/strategy/strategy_helper.py
clover-es: I don't think it's possible to have more than (or exactly) 100% gain on a short, otherwise the price will be below 0 :)

xmatthias: mmh good point

I'm also wrong on the remainder of the review ... as i "forgot" the fact that 
1000 to 800 might be a 20% decrease, but it's actually a ~25% increase.

freqtrade/strategy/strategy_helper.py
clover-es: Yes, so I was a bit lost in the usage of this function, but after further review, I think the implementation is correct, even though maybe docstring could be improved. And also, there's where my confusion with the sign convention comes from.

When I talk about open_stoploss_price, it's just a ratio with the open_price, so this fulfils the purpose you described.

And also, I speak about prices because the definitions are needed to calculate the formula, but of course, they're factored out and they're not used at all, they just serve as example to make sure I did the math correctly.

xmatthias (approved):
Great work :100: 

Sorry for the confusion at first ... totally my bad as i didn't think about this

LGTM :+1:

PR #6230, opened 2022-01-16
add position_adjustment_enable to config_full.example

stash86:

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

### DIFF
--- a/config_examples/config_full.example.json
+++ b/config_examples/config_full.example.json
@@ -177,5 +177,6 @@
     "strategy": "SampleStrategy",
     "strategy_path": "user_data/strategies/",
     "dataformat_ohlcv": "json",
-    "dataformat_trades": "jsongz"
+    "dataformat_trades": "jsongz",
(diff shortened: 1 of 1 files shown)


config_examples/config_full.example.json
```diff
@@ -177,5 +177,6 @@
     "strategy": "SampleStrategy",
     "strategy_path": "user_data/strategies/",
     "dataformat_ohlcv": "json",
-    "dataformat_trades": "jsongz"
+    "dataformat_trades": "jsongz",
+    "position_adjustment_enable": false
```
xmatthias, line 181: I'm honestly not sure about this.

While it CAN be set via configuration - it really shouldn't - as this is really a strategy property, which should be set in the strategy.
I think we have a few others of these (which could be set in config but are not listed here intentionally).

stash86:
fair. removing it now

PR #6236, opened 2022-01-16
[WIP] Add order count to telegram status table

txtr99:
Unit tests - pending.
Flake8 - compliant.
iSort - skipped 9 files, not sure yet what that means.
mypy - no errors related to this PR.

## Summary

This PR supports the new "Adjust Position Size" code by adding an additional Column to the Telegram /status table report, showing the user the # of filled orders making up the total trade position.

## Quick changelog

- Added code to rpc.py bringing in the count of filled orders, and added a column to the final output table showing the result if > 1.

## What's new?

The benefit of this PR is intended to be an increase in awareness for the user about where in the Dollar Cost Averaging logic we currently are.

Currently there is no easy way via the /status table command, to see how many times the Strategy has added to the base position, and this may cause a general unawareness of risk and lack of understanding about invested capital.

Adding the Count allows the user to understand whether they are on track to lower the average entry price, or whether the strategy might be holding off on new purchases due to some other safeties in place at the time.

## Newly introduced problems

There is a problem this PR introduces which is that it clutters the screen for strategies where this information is not relevant - ie: those without any position adjustment logic.  I have not yet designed a way to automatically toggle this column on or off.

However I do have the following idea for this:

We could check the "entire" trade list and if there are "no trades with more than 1 filled order" then we could have the new "Buys" column completely hidden until at least 1 trade meets the criteria of having count_of_buys > 1.

## Screenshots

This screenshot shows the new table format with the column header as "DCA" which is another option, but I elected to go with "Buys" for the PR to make it more generic for folks that are not trying to Dollar Cost Average necessarily when they buy more of a position.

I have also shown the screenshot with the option of having the rows "blank" when there is only the first original Buy.  In the PR Code I've used a Hyphen (-) instead.

Additionally in the screenshot I've used (count_of_buys - 1) so that the column show only the count of "extra buys" rather than the total count.

These differences highlight some of the different ways this could be implemented.

Screen Shot 2022-01-16 at 12 53 05 PM

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -238,19 +238,25 @@ class RPC:
                         profit_str += f" ({fiat_profit:.2f})"
                         fiat_profit_sum = fiat_profit if isnan(fiat_profit_sum) \
                             else fiat_profit_sum + fiat_profit
+                
+                # count the number of filled orders for the trade so far
+                filled_buys = trade.select_filled_orders('buy')
+                count_of_buys = len(filled_buys)
+                
                 trades_list.append([
                     trade.id,
                     trade.pair + ('*' if (trade.open_order_id is not None
                                           and trade.close_rate_requested is None) else '')
                                + ('**' if (trade.close_rate_requested is not None) else ''),
                     shorten_date(arrow.get(trade.open_date).humanize(only_distance=True)),
-                    profit_str
+                    profit_str, # added a comma
+                    str(count_of_buys) if count_of_buys > 1 else '-' # added the count of filled orders. It can be cluttered to see a bunch of 1's so I tried the hyphen for first orders. This won't be a useful column for Strategies that do not use DCA at all, not sure how we can make this go away for those.
                 ])
             profitcol = "Profit"
             if self._fiat_converter:
                 profitcol += " (" + fiat_display_currency + ")"
 
-            columns = ['ID', 'Pair', 'Since', profitcol]
+            columns = ['ID', 'Pair', 'Since', profitcol, 'Buys'] # added the 'Buys' column to count filled orders.  'Qty' seems ambiguous and might be confused with 'Amount'. But it's not a perfect name.
             return trades_list, columns, fiat_profit_sum
 
     def _rpc_daily_profit(
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -238,25 +238,31 @@ class RPC:
                         profit_str += f" ({fiat_profit:.2f})"
                         fiat_profit_sum = fiat_profit if isnan(fiat_profit_sum) \
                             else fiat_profit_sum + fiat_profit
-                
-                # count the number of filled orders for the trade so far
+
+                # Count the number of filled orders for the trade so far
                 filled_buys = trade.select_filled_orders('buy')
                 count_of_buys = len(filled_buys)
-                
+
                 trades_list.append([
                     trade.id,
                     trade.pair + ('*' if (trade.open_order_id is not None
                                           and trade.close_rate_requested is None) else '')
                                + ('**' if (trade.close_rate_requested is not None) else ''),
                     shorten_date(arrow.get(trade.open_date).humanize(only_distance=True)),
-                    profit_str, # added a comma
-                    str(count_of_buys) if count_of_buys > 1 else '-' # added the count of filled orders. It can be cluttered to see a bunch of 1's so I tried the hyphen for first orders. This won't be a useful column for Strategies that do not use DCA at all, not sure how we can make this go away for those.
+                    profit_str,  # added a comma
+                    # Added the count of filled orders. It can be cluttered to see a bunch
+                    # of 1's so I tried the hyphen for first orders. This won't be a useful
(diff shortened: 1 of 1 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -238,19 +238,31 @@ def _rpc_status_table(self, stake_currency: str,
...
+                    str(count_of_buys) if count_of_buys > 1 else '-'
                 ])
             profitcol = "Profit"
             if self._fiat_converter:
                 profitcol += " (" + fiat_display_currency + ")"
 
-            columns = ['ID', 'Pair', 'Since', profitcol]
+            # Added the 'Buys' column to count filled orders.  'Qty' seems ambiguous and might
+            # be confused with 'Amount'. But it's not a perfect name.
+            columns = ['ID', 'Pair', 'Since', profitcol, 'Buys']
```
xmatthias, line 265: Now i'm not convinced on adding another column to the `/status table` response.

It's already pretty tight (wraps on mobile in many cases, depending on order id and pair name) - and adding additional columns will only make this worse.

On a desktop, it might work - but telegram is mainly for mobile, so we should also design it for that.
It might be a better fit for the regular `/status` call - which gives you more details on each order (and i consider this as detail).

txtr99:
My thoughts on the wrapping are as you say: it already wraps at least 50%
of the time, so there is really no point in trying to preserve the random
other 50% where it doesn’t.  On some devices it wraps 100% of the time
already such as my iPhone SE screen.

Mobile is where we need the info the most.

However, an alternative might be to create a 2nd command that returns
additional information, even adding 2-3 columns pertaining to DCA’s.
/stacking?
/status dca?
/averaging table?

Or — we redesign the entire table so that it never wraps by eliminating
data (for example we don’t really need the profit in $, I think % is enough)

This would improve the layout considerably and also allow room for a 1-2
character integer.

So:

ID     Pair(+**)           Age    Buys / Roi
28    BTC/USDT**    4h       3 (-2.31%)

At the bottom we could still have “Total” for the unrealized profit / loss.


A few different ways we could proceed.

xmatthias:
> Or — we redesign the entire table so that it never wraps by eliminating data (for example we don’t really need the profit in $, I think % is enough)

in principle, you're right - except you're not (it'll depend on the case).
If i'm using static stake (or dynamic stake without messing with stake amount) - % will be enough, as all stakes will be similarly sized.

With custom_stake amount (and DCA as well), this changes. i could have a trade with 100$ - and one with 1000$ or more.
Now i would see both as "2%" - but while  on one trade, this will mean 2$ - for the other it'll be 20$ - which is quite some difference.

I'm however clearly not against reorganizing /streamlining the table if we can find a good way to display "what's relevant" ...

xataxxx:
Regarding this PR, I think you should only show the buys column if the strategy has position_adjustment_enable set to True.
PS: I'm planning to add a similar functionality into trade list (not table), that would list out each individual order of the trade.

stash86:
To chime in, on my phone, the status table is already on 2 lines for each trades
98A192B5-345B-4312-A29B-5F6C7CFBF04F

B81ED028-B7B7-41F2-A304-3509354B4380

So adding the "Buys" column (or whatever the name gonna be) won't break it on mobile, although we can argue whether that last column should be used by it, or reserved for something else more important in the future (like short or long position?)

stash86:
> Regarding this PR, I think you should only show the buys column if the strategy has position_adjustment_enable set to True.
> PS: I'm planning to add a similar functionality into trade list (not table), that would list out each individual order of the trade.

I already have it coded actually. Code on my fork, on develop branch. Might need some refinement regarding wordings and templates, but it's working

C7BC1C61-CB4A-4FE1-B790-2565FAD26CE4

Gonna send PR of it pieces by pieces

txtr99:
Are you against adding settings for what’s shown to the config.Json file
inside “Telegram”? Similar to the keyboard layout perhaps.

Or to the buy/sell toggles (on/off)

In that case we could enable folks to choose from predefined options they
feel are important to them

We could even allow a strategy level override if they choose to omit this
from the config file, and customize for each strategy.

xmatthias:
> Are you against adding settings for what’s shown to the config.Json file inside “Telegram”? Similar to the keyboard layout perhaps.

i don't think that'll give us what we're looking for. You'll not care about DCA orders unless you're using DCA (so it's a simple check against "has this strategy DCA enabled").
If it's something that will be "questionable" though, i'm not so sure.

We could have different endpoints/commands specifying what you'd like (`/status table`, `/status dca`, ...). That way we'll not prevent you from accessing that information if you've configured it "the other way".

txtr99:
This seems like the right approach, and will be inline with the master
change for the position adjustment functionality where folks that decide
not to use it, will see no change at all in their experience.

Those that do decide to use it, will have the “option”
Of changing their telegram experience via a different command trigger.

stash86:
I have my PR #6259 to do same thing as here, but then only enabled when position adjustment is enabled. I guess most users won't mind additional column if they are using position adjustment.

xmatthias:
I'll close this in favor of #6259, simply because that PR is developed further (it only shows the column header when necessary).

The discussion here is still a good one, and i'd appreciate a PR updating the `/status` command as well.

txtr99:
Excellent thank you!

PR #6252, opened 2022-01-17
Fix eager-loading trade.orders

xataxxx:
## Summary

Eager-loading orders allows access to trade.orders in all callbacks.

## Quick changelog

- Added lazy="joined" to Trade.orders

## What's new?

We can now access previous orders in all custom callbacks like custom_sell, confirm_trade_exit etc.


xmatthias (approved):
LGTM 👍

PR #6254, opened 2022-01-18
Spreadfilter log % fix

SmartManoj:
Both the following are same
```
print(f'{3:.3%}')
print(f'{3 *100 :.3}%')
# 300.000%
```

Here both are mixed.


xmatthias (approved):
LGTM :+1:

PR #6255, opened 2022-01-18
Rework backtesting --no-cahche to --cache=[none, day, week, month].

rokups:
Exploring idea we talked about.


freqtrade/data/btanalysis.py
```diff
@@ -168,6 +170,14 @@ def find_existing_backtest_stats(dirname: Union[Path, str],
             # is safe to assume older files will also not have any metadata.
             break
 
+        if min_backtest_date is not None:
+            backtest_date = datetime.strptime(
+                f'{filename.name}', 'backtest-result-%Y-%m-%d_%H-%M-%S.json')
```
xmatthias: I'm not sure how you imagine this to work (i don't think we can use the filename).

assume the following:
i use `backtesting` yesterday - and continue today with  `backtesting --cache week`.
It'll load yesterday's results - and store them into today's file (so far, so good).

I'm now using `--cache day` - so i'd expect only results that were _backtested_ today to be considered.
The one from yesterday however got "moved over" to today's file - so it'll pass this filter (i can move a backtest result over weeks/months this way).


I think we'll need to extend the metadata file with the "backtest-date" - storing when the original backtest was made (and obviously move this date unchanged to new metadata files).

rokups: Gah well that is a very good point 😬

freqtrade/configuration/configuration.py
```diff
@@ -276,8 +276,8 @@ def _process_optimize_options(self, config: Dict[str, Any]) -> None:
         self._args_to_config(config, argname='backtest_breakdown',
                              logstring='Parameter --breakdown detected ...')
 
-        self._args_to_config(config, argname='no_backtest_cache',
-                             logstring='Parameter --no-cache detected ...')
+        self._args_to_config(config, argname='backtest_cache',
+                             logstring='Parameter --cache detected ...')
```
xmatthias: You might wanna have this output the value (see below how to).

freqtrade/optimize/backtesting.py
```diff
@@ -747,21 +763,17 @@ def start(self) -> None:
...
 
         # Load previous result that will be updated incrementally.
         # This can be circumvented in certain instances in combination with downloading more data
-        if self.timerange.stopts == 0 or datetime.fromtimestamp(
-           self.timerange.stopts, tz=timezone.utc) > datetime.now(tz=timezone.utc):
-            self.config['no_backtest_cache'] = True
-            logger.warning('Backtest result caching disabled due to use of open-ended timerange.')
-
-        if not self.config.get('no_backtest_cache', False):
+        min_backtest_date = self._get_min_cached_backtest_date()
```
xmatthias, line 773: Just to spark your interest :wink: 
i think it might be possible to move this whole logic to before loading data

so if you're re-running (intentionally or not) the same backtest, you might get a result VERY quickly ...

xmatthias (approved):
I think this is the nicer interface than the `--no-cache` - great idea!

LGTM :+1:

PR #6258, opened 2022-01-19
Pass dimensions to generate_estimator

italodamato:
## Summary

The number of optimizable dimensions is needed within `generate_estimator` in order to create isotropic kernels for the GaussianProcessRegressor

## Quick changelog

- passing `dimensions` to `generate_estimator` in `hyperopt.py`
- passing `dimensions` to `generate_estimator` in `hyperopt_auto.py`

## What's new?

Example:
```
def generate_estimator(dimensions):
	from skopt.learning import GaussianProcessRegressor
	from skopt.utils import cook_estimator
	from skopt.learning.gaussian_process.kernels import (RBF, Matern, RationalQuadratic, ExpSineSquared, DotProduct, ConstantKernel, WhiteKernel)
	kernel_bounds = (0.0001, 10000)
	kernel = ConstantKernel(1.0, kernel_bounds) * Matern(
							length_scale=np.ones(len(dimensions)),
                				length_scale_bounds=[kernel_bounds for d in dimensions], nu=2.5)
	kernel += ConstantKernel(1.0, kernel_bounds) * Matern(
                				length_scale=np.ones(len(dimensions)),
               				length_scale_bounds=[kernel_bounds for d in dimensions], nu=1.5)

	return cook_estimator("GP", space=dimensions, kernel=kernel, n_restarts_optimizer=2)
```

I'm not sure if there's a better way to achieve this. 

TODO: Update docs accordingly

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -367,7 +367,7 @@ class Hyperopt:
         }
 
     def get_optimizer(self, dimensions: List[Dimension], cpu_count) -> Optimizer:
-        estimator = self.custom_hyperopt.generate_estimator(dimensions)
+        estimator = self.custom_hyperopt.generate_estimator(dimensions=dimensions)
 
         acq_optimizer = "sampling"
         if isinstance(estimator, str):
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -367,7 +367,7 @@ class Hyperopt:
         }
 
     def get_optimizer(self, dimensions: List[Dimension], cpu_count) -> Optimizer:
-        estimator = self.custom_hyperopt.generate_estimator()
+        estimator = self.custom_hyperopt.generate_estimator(dimensions)
 
         acq_optimizer = "sampling"
         if isinstance(estimator, str):
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,3 @@ target/
 !config_examples/config_ftx.example.json
 !config_examples/config_full.example.json
 !config_examples/config_kraken.example.json
-docker-compose.yml
-.DS_Store
-user_data/data/.gitkeep
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,6 @@ target/
 !config_examples/config_ftx.example.json
 !config_examples/config_full.example.json
 !config_examples/config_kraken.example.json
+docker-compose.yml
+.DS_Store
+user_data/data/.gitkeep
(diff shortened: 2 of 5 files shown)


freqtrade/optimize/hyperopt.py
```diff
@@ -367,7 +367,7 @@ def _get_results_dict(self, backtesting_results, min_date, max_date,
         }
 
     def get_optimizer(self, dimensions: List[Dimension], cpu_count) -> Optimizer:
-        estimator = self.custom_hyperopt.generate_estimator()
+        estimator = self.custom_hyperopt.generate_estimator(dimensions)
```
xmatthias: I'm not sure about implementing it this way - it's a change breaking all strategies that implement this function.

a potentially better way might be to simply assign dimensions to custom_hyperopt. there's little to no difference if it's used as `self.dimensions` or `self.custom_hyperopt.dimensions` - and you could access it this way:

``` python
class MyAwesomeStrategy(IStrategy):
    class HyperOpt:
        def generate_estimator(self)
            ... = len(self.dimensions)
```

italodamato: Yes, I'm sure there's a better way to accomplish this. Would you mind being a bit more specific about your idea? Are you suggesting to initialize `self.dimensions` [here](https://github.com/italodamato/freqtrade-1/blob/a4dbdb549d772b488305cb12727c5a8375b0cb2d/freqtrade/optimize/hyperopt_interface.py#L36)?

xmatthias: no not really
if you look at the `hyperopt.py` file - there's only 4(ish) usages of `self.dimension`.

except for the first (which needs to be moved down a bit within that function), these can be changed to `self.custom_hyperopt.dimension` - so you'll have it automatically available.

`get_optimizer()` also don't "really" needs it as argument (it's available as class-attribute anyway, so that's something that's been historically this way) - and it'll be passed by reference anyway

xmatthias: hmm, the suggested approach above does not work as we never create an instance of the HyperOpt childclass, therefore i guess your approach is correct 👍 

the only thing:
would you mind calling the function with keyword args (basically `(dimensions=dimensions)` - as well as updating the docs for this (basically search for `generate_estimator`).

I'd want to add a `**kwargs` argument to the function definition as well - so we can easily add new parameters without breaking existing strategies should that be necessary in the future.

xmatthias (approved):
Nice improvement :+1: 

ps: if you think i'm wrong on a Review - simply point out that my proposal doesn't work :laughing: it's not guaranteed that i'll always look at it within reasonable time to figure that out myself ...

PR #6259, opened 2022-01-19
add "# Buys" column to status table

stash86:

## Summary

Add "# Buys" column to status table, if the user enable position adjustment. If not, then the "# Buys" column won't be added. And the value there will be total number of buy(s), so even with the first buy, the number there will show 1

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

For bot that have position adjustment enabled, add another column at the end of the status table named '# Buys', to show current total buys of each trades.

Bot without position adjustment
image

Bot with position adjustment (the bot in the screenshot below is using old draft code, so the column name is different)
image


xmatthias (approved):
Nice improvement, thanks a lot :+1:

PR #6260, opened 2022-01-19
Add max_buy_position_adjustment as attribute

stash86:

## Summary

Add max_buy_position_adjustment as attribute so it can be used for telegram messages.

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?


freqtrade/strategy/interface.py
```diff
@@ -108,6 +108,7 @@ class IStrategy(ABC, HyperStrategyMixin):
 
     # Position adjustment is disabled by default
     position_adjustment_enable: bool = False
+    max_buy_position_adjustment: int = 1
```
xmatthias: the use of this (and how it's named) will only depend on the strategy, and how the strategy implements it (the documentation is just a sample, not the default behavior). 
We could have different samples, using different calculation methods (also ignoring this completely).
Users are expected to take the sample - and modify it (using whatever logic they chose).

By adding this as configuration element (suggesting it is applied internally), i think it can be confusing and potentially dangerous - if you assume your setting comes from the configuration, but the strategy you're using is really using different naming (or calculates differently altogether).

stash86: I'll make it to behave similar to max_open_trades. -1 for unlimited adjust buy (which is default), which can either be set to another value on strat file if needed, or let the coder create their own logic.

if the max limit is set to a positive integer, the bot will check internally for number of buys. If it's still below the limit, then adjust_position will be called. If not, adjust_position won't be called at all. So this snippet from the docs won't be needed anymore

```filled_buys = trade.select_filled_orders('buy')
        count_of_buys = len(filled_buys)
        
        # Allow up to 3 additional increasingly larger buys (4 in total)
        # Initial buy is 1x
        # If that falls to -5% profit, we buy 1.25x more, average profit should increase to roughly -2.2%
        # If that falls down to -5% again, we buy 1.5x more
        # If that falls once again down to -5%, we buy 1.75x more
        # Total stake for this trade would be 1 + 1.25 + 1.5 + 1.75 = 5.5x of the initial allowed stake.
        # That is why max_dca_multiplier is 5.5
        # Hope you have a deep wallet!
        if 0 < count_of_buys <= self.max_buy_position_adjustment:```
        
So if a user prefer to not use that attribute, they can leave it on default (which is -1) and create their own check logic.

**Comment by stash86** [comment]
gonna close this now, and come back with new PR after I tested my new idea

**Comment by xmatthias** [comment]
> gonna close this now, and come back with new PR after I tested my new idea

You could also leave it open - i think i understood now what you're trying to do (that part is just not yet visible - hence i didn't understand the intend).
I actually think the idea is great, as we'll save freqtrade from calling the callback 100ds of times for no good reason.

**Comment by stash86** [comment]
Okay, new change is up. Basically just moving the check to the front of the process, so that if the max adjust buy has been reached, it don't need to call all those stake amount checks etc.

And thinking of whether I would want to move this check to the front as well
```
if current_profit > -0.05:
       return None```

xmatthias:
> And thinking of whether I would want to move this check to the front as well

No - this is something that's purely usecase related.
you seem to be using it for DCA (increasing a losing position).

but the feature of buying more can also be used to increase winning positions (that's probably the better approach, at least from a risk perspective).
While we don't have an example for that - there's currently nothing that's preventing this behavior.

stash86:
Yeah, thinking of trying it as well rn.

Anyway, the change worked as expected

freqtrade/rpc/rpc.py
```diff
@@ -136,7 +136,10 @@ def _rpc_show_config(config, botstate: Union[State, str],
             'ask_strategy': config.get('ask_strategy', {}),
             'bid_strategy': config.get('bid_strategy', {}),
             'state': str(botstate),
-            'runmode': config['runmode'].value
+            'runmode': config['runmode'].value,
+            'position_adjustment_enable': config.get('position_adjustment_enable', False),
+            'max_buy_position_adjustment': (config['max_buy_position_adjustment']
```
xmatthias: I do appreciate adding this here - please also add it to `ShowConfig` (in `api_schemas.py`) to complete this.

Otherwise it'll not be available via API.

xmatthias (changes requested):
As documented [here](https://www.freqtrade.io/en/stable/developer/#developer-setup) and [in the contributing guidelines](https://github.com/freqtrade/freqtrade/blob/develop/CONTRIBUTING.md)- please consider running the "common" tools locally.
You'll find that that will be more efficient (you'll not have to wait for CI to fail to make minor adjustments).

that'd be `flake8`, `mypy freqtrade` `isort .` and `pytest` (although tests seem to pass now).

stash86:
> As documented [here](https://www.freqtrade.io/en/stable/developer/#developer-setup) and [in the contributing guidelines](https://github.com/freqtrade/freqtrade/blob/develop/CONTRIBUTING.md)- please consider running the "common" tools locally. You'll find that that will be more efficient (you'll not have to wait for CI to fail to make minor adjustments).
> 
> that'd be `flake8`, `mypy freqtrade` `isort .` and `pytest` (although tests seem to pass now).

Sorry. On it rn

stash86:
these tests are failed, but I don't do any changes that supposedly affecting them. So can I ignore them? The 3 other tests passed

============================================= short test summary info ==============================================
FAILED tests/test_directory_operations.py::test_create_userdata_dir_and_chown - AssertionError: assert 1 == 0
FAILED tests/test_plotting.py::test_generate_candlestick_graph_no_signals_no_trades - ValueError: Invalid propert...
FAILED tests/test_plotting.py::test_generate_candlestick_graph_no_trades - ValueError: Invalid property specified...
FAILED tests/test_plotting.py::test_generate_profit_graph - ValueError: Invalid property specified for object of ...
================ 4 failed, 1749 passed, 1 skipped, 48 deselected, 71 warnings in 225.83s (0:03:45) =================

freqtrade/rpc/rpc.py
```diff
@@ -136,7 +136,11 @@ def _rpc_show_config(config, botstate: Union[State, str],
...
```
xmatthias: unfortunately, the way the defaults are set, this causes a crash in webserver mode (start freqtrade with `freqtrade webserver` with a config that does not contain this key and login to the UI).

Not all freqtrade modes load a strategy right away - so to properly set the default, we'll have to have it as part of the jsonschema (including a default value) in `constatnts.py`.

docs/strategy-callbacks.md
```diff
@@ -611,7 +611,7 @@ class DigDeeperStrategy(IStrategy):
     # ... populate_* methods
```
xmatthias, line 616: we'll want to have some explanation in this section (above this codeblock, obviously) - explaining that the callback won't be called when `max_buy_position_adjustment` is set and reached for a trade.

xmatthias (changes requested):

freqtrade/freqtradebot.py
```diff
@@ -473,6 +473,14 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
         If the strategy triggers the adjustment, a new order gets issued.
         Once that completes, the existing trade is modified to match new data.
         """
+        if self.strategy.max_buy_position_adjustment > -1:
+            filled_buys = trade.select_filled_orders('buy')
+            count_of_buys = len(filled_buys)
+            if count_of_buys > self.strategy.max_buy_position_adjustment:
```
xmatthias: Are you sure this comparison is correct?

assuming we set `max_buy_position_adjustment` to 3 (so we allow 3 buy orders - for example 3x 100$ for a total of 300$ stake)
after the initial trade entry, `count_of_buys` will be 1
after the 2nd order, it'll be 2
after the 3rd order, i'll be 3 (we've now reached our 3x 100$ investment)

This comparison will however only result to true on the 4th order - basically having the investment as 4x 100$.

Now it'll depend if we do count the initial buy or not. 
Personally, i think it feels more intuitive to define the overall entries (including the initial call) ... otherwise i have to calculate my total stake per pair as `(1 + max_buy_position_adjust) * stake` (assuming an even stake, obviously).

stash86: It depends on the name of the setting, in my opinion. max_buy_position_adjustment for me refers to the extra buys that can only happen if we enable position adjustment, hence the naming.

If we want to go for total number of buys, then naming it max_buys_per_trade would be better?

xmatthias: that's correct - i guess both are valid, we just need to be clear in the documentation.
i've also asked for additional opinions in discord ...

xataxxx: This could be written a lot simpler like this, because we have nr_of_successful_buys:
`if trade.nr_of_successful_buys > self.strategy.max_buy_position_adjustment:`

Also, since our trade already contains "nr_of_successful_buys" then it would make sense to name the max variable as `max_buys_per_trade`

xmatthias: that's "danger zone" i think (from a naming perspective)

calling it `max_buys_per_trade` will have a different impact than `max_buy_position_adjustment` ... as i'd expect the first to include the initial order, while the 2nd wouldn't.

For me both work fine - but we'll need to be clear on what we want (and also have that clear in the documentation).

xataxxx: I'd be ok with both names. 
Also wanted to point out that we also have /forcebuy that allows buying additional orders and that would not (and should not) honor this max limit, since it's an end-user initiated Force buy (I always imagine the user doing a Jedi mind-trick pose when using that).

xmatthias: The name is now Fine based on the description provided.

This was more about making sure this is what we want.

docs/strategy-callbacks.md
```diff
@@ -653,8 +655,7 @@ class DigDeeperStrategy(IStrategy):
             return None
 
         filled_buys = trade.select_filled_orders('buy')
-        count_of_buys = len(filled_buys)
-        
+        count_of_buys = trade.nr_of_successful_buys
```
xmatthias, line 661: I found that these 2 lines are still necessary - as they're used as part of the conditions below (even though it's no longer used for the conditional). 
I've therefore re-added them.

freqtrade/strategy/interface.py
```diff
@@ -108,6 +108,7 @@ class IStrategy(ABC, HyperStrategyMixin):
 
     # Position adjustment is disabled by default
     position_adjustment_enable: bool = False
+    max_buy_position_adjustment: int = -1
```
xmatthias: thinking about this last-second (right before merging, basically) - i think we should not name it `max_buy_position_adjustment` - but instead `max_entry_position_adjustment`.

With `feat/short` turning up sometime in the future, using `buy` and `sell` terminology will be dangerous (`buy` no longer automatically means _entry_ - but could also mean _exiting a trade_).


It's naming, but something i **try** to have with everything new we add - but i'm not always thinking about it (especially in more complex PR's, which require a lot of other stuff to think about).

xmatthias (approved):
LGTM :+1: 

you'll notice that i've renamed the setting to `max_entry_position_adjustment` - as this will prepare us for short trading - where we'll no longer assume "buy = entry".

The CI failure is not related to this PR .... however it'll prevent docker-images to be updated for now.

freqtrade/constants.py
```diff
@@ -371,7 +371,9 @@
             'type': 'string',
             'enum': AVAILABLE_DATAHANDLERS,
             'default': 'jsongz'
-        }
+        },
+        'position_adjustment_enable': {'type': 'boolean', 'default': False},
```
xataxxx, line 375: Both of these parameters shouldn't be in the config schema!

If they are in the schema and have defaults then they overwrite strategy, meaning people without `"position_adjustment_enable": true` in their config file will have it turned off regardless if the strategy says it should be True.

xmatthias: already fixed in fb3c67d86b64c6d603de87a8b156e78398a661b7 :)

xmatthias: good spot - fixed in e7409e74c2c4798c78559d47eb69842202fca288

stash86: But it's not behaving that way? I'm using it live that way, and it work just fine (I don't enable the position adjustment on config.json, only on strategy file). The default don't overwrite my strategy

stash86: Ah never mind. Just read perk's reply. It caused problem on backtest.

xmatthias: it'll cause the same problem in live as well iirc - as the setting will come from the configuration already...

PR #6264, opened 2022-01-20
add buy_tag to trade tooltip in plots

italodamato:
## Summary

add buy_tag to trade tooltip in plots, it makes it easier to troubleshoot strategies

## Quick changelog

- added the trade `buy_tag` to the tooltip shown in plots


xmatthias:
I'd recommend running the "common" tools we have locally first [documented here](https://github.com/freqtrade/freqtrade/blob/develop/CONTRIBUTING.md#before-sending-the-pr).

Passing tests/CI is one core requirement ... and it's usually pretty simple (and a lot faster locally, where you don't have to wait for 15 minutes for CI to tell you what's wrong).

xmatthias (commented):
am i correct in assuming that the only change to the json file is the addition of one "buy_tag" field? (it's difficult to compare that file as it's all in one line ...).

italodamato:
> am i correct in assuming that the only change to the json file is the addition of one "buy_tag" field? (it's difficult to compare that file as it's all in one line ...).

Yes just added `buy_tag` on the second trade

xmatthias (approved):
LGTM :+1: 

Nice enhancement, thanks a ton! :100:

PR #6272, opened 2022-01-22
Add more info on Telegram's status message

stash86:

## Summary

Add more info into telegram's status message

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

Add more information into status message, to help users not having to look into frequi and/orr scroll up the telegram chat logs for these info
1. Sell reason
2. Number of buys, if position adjusment enabled
3. Open datetime
4. Close datetime
5. Details of each buy, if the trade have more than 1 buy order

The result would be something like this
```
Trade ID: 50 (since 5 days ago)
Current Pair: QRDO/USDT
Amount: 24.7304 (102.86388509 USDT)
Sell Reason: roi
Number of Buy(s): 4
Open Rate: 4.15941049
Close Rate: 4.17370000
Open Date: 2022-01-16 17:15:24
Close Date: 2022-01-20 14:51:48
Current Rate: 4.17370000
Close Profit: -0.06%
Stoploss: 0.00000000 (-100.00%)
Stoploss distance: -4.17370000 (-100.00%)
  
Buy #1:
Buy Amount: 2.9542 (14.61147320 USDT)
Average Buy Price: 4.946
  
Buy #2: at -9.96% avg profit
(5 days 0 hours and 38 minutes ago)
Buy Amount: 3.281 (14.61127730 USDT)
Average Buy Price: 4.4533 (-9.96% from 1st buy rate)
Order filled at: 2022-01-17 05:12:16
(0d 11h 56m 52s from previous buy)
  
Buy #3: at -8.79% avg profit
(4 days 13 hours and 54 minutes ago)
Buy Amount: 6.1527 (26.30024159 USDT)
Average Buy Price: 4.2745854 (-13.57% from 1st buy rate)
Order filled at: 2022-01-17 15:56:19
(0d 10h 44m 3s from previous buy)
  
Buy #4: at -14.42% avg profit
(2 days 9 hours and 13 minutes ago)
Buy Amount: 12.3425 (47.34089300 USDT)
Average Buy Price: 3.8356 (-22.45% from 1st buy rate)
Order filled at: 2022-01-19 20:37:13
(2d 4h 40m 54s from previous buy)
```


xataxxx:
Great idea! Wanted to point out that maybe it would make more sense to add to_json() method for individual orders in the Order class and then use that inside the trade to_json? This would allow reusing the json structure later in various other RPC endpoints.

freqtrade/rpc/telegram.py
```diff
@@ -369,6 +369,47 @@ def _get_sell_emoji(self, msg):
...
+                lines.append("({})".format(cur_buy_date
+                                           .humanize(granularity=["day", "hour", "minute"])))
+                lines.append("*Buy Amount:* {} ({:.8f} {})"
+                             .format(cur_buy_amount, filled_trades[str(x)]["cost"], base_currency))
+                lines.append("*Average Buy Price:* {} ({:.2%} from 1st buy rate)"
+                             .format(cur_buy_average, price_to_1st_buy))
+                lines.append("*Filled at:* {}".format(filled_trades[str(x)]["order_filled_date"]))
+                lines.append("({}d {}h {}m {}s from previous buy)"
+                             .format(days, hours, minutes, seconds))
+        return lines
```
xmatthias: While we surely don't have 100% line by line coverage - that's a too large chunk to leave completely untested, as we risk having it break in the future with some changes (we'd need to manually test this - which is simply not feasible for changes that don't "seem" to touch this code).

freqtrade/rpc/telegram.py
```diff
@@ -396,17 +437,31 @@ def _status(self, update: Update, context: CallbackContext) -> None:
             messages = []
             for r in results:
                 r['open_date_hum'] = arrow.get(r['open_date']).humanize()
+                r['filled_buys'] = r.get('filled_buys', [])
```
xmatthias: what's the point of this?
we either assign `'filled_buys'` in `_rpc_trade_status` - or we don't (so we can't use it).

it's the method that assigns results is not user-provided - so we can rely on it having "whatever" fields we assign to it there.
I assume it's working around broken tests - but i'd rather fix these tests than have some crutch code in the real code - as it can cover up potential failures / problems.

freqtrade/persistence/models.py
```diff
@@ -345,6 +361,7 @@ def to_json(self) -> Dict[str, Any]:
             'max_rate': self.max_rate,
 
             'open_order_id': self.open_order_id,
+            'filled_buys': buys_json,
```
xmatthias: i'd name this `filled_trade_entries` - as it could in theory also contain sells (in case of shorts) in the future (and having it correct in the first place is one less area we need to touch / break in the future).

That said - i'm currently unsure - is there a benefit / drawback to having all (filled) orders available here directly (which would eventually expose it this way via API), and filter where needed?

for most telegram calls (like /status) - there'd currently be no difference (once you sold, it ain't in `/status` anymore - and having sells in `/trade <id>` might not be a bad thing?).

freqtrade/persistence/models.py
```diff
@@ -165,6 +165,16 @@ def update_from_ccxt_object(self, order):
                 self.order_filled_date = datetime.now(timezone.utc)
         self.order_update_date = datetime.now(timezone.utc)
 
+    def to_json(self) -> Dict[str, Any]:
+        return {
+            'cost': self.cost if self.cost else 0,
+            'amount': self.amount,
+            'price': self.price,
+            'average': round(self.average, 8) if self.average else 0,
+            'order_filled_date': self.order_filled_date.strftime(DATETIME_PRINT_FORMAT)
```
xmatthias, line 180: I'd always use a "_timestamp" attribute here as well (aligned to how we do it for the `trade.to_json()`).

These methods (`to_json`) are intended to be directly used by the API - and parsing a date from a string in clients is quite errorprone (and sometimes expensive).

I'd also add some more fields like `side` (or better `ft_order_side`) and `order_type` as well as `filled` (that's some i know by memory - not necessarily all) - as this will not just be limited to telegram, but will also be used by the API - where it makes sense to show some of these.

freqtrade/persistence/models.py
```diff
@@ -282,6 +292,12 @@ def close_date_utc(self):
         return self.close_date.replace(tzinfo=timezone.utc)
 
     def to_json(self) -> Dict[str, Any]:
+        fill_buy = self.select_filled_orders('buy')
+        buys_json = dict()
+        if len(fill_buy) > 0:
+            for x in range(len(fill_buy)):
+                buys_json[str(x)] = fill_buy[x].to_json()
```
xmatthias: Reminder to self: 
Investigate impact on backtesting filesize - as this will be exported there, too.

freqtrade/rpc/telegram.py
stash86: Okay, gonna remove this and debug the test file instead

freqtrade/persistence/models.py
stash86: Well, actually we might want to have all filled orders (buy and sell) here, since some people still using limit sell, and might want to have details of the sells as well

Just to be sure, we don't have `/trade <id>` currently, right? Or do I miss it?

xmatthias: Apparently not for telegram (there's no reason we could not have that in the future...) - but the api has a `/trade/<id>` endpoint (although for the API, the `/trades` endpoint is the more relevant as you can get more data there).

xataxxx: There is /status 123
Returns:
Trade ID: 123 (since a day ago)
Current Pair: XLM/BUSD
Amount: 9261.0 (2402.8153 BUSD)
Open Rate: 0.25945527
Current Rate: 0.26100000
Current Profit: 0.44%
Stoploss: 0.18758200 (-29.00%)
Stoploss distance: -0.07341800 (-28.13%)

freqtrade/persistence/models.py
xmatthias: Don't think this is relevant for backtesting - backtesting only writes "selected" columns to file.

freqtrade/persistence/models.py
xmatthias: yes, but it won't work for closed trades iirc - so if trade `123` was closed, you'll not see that.

stash86: It's still shown? That's how I get the example on the top.

If you noticed, on the example I posted above, there are sell related info, like `Sell reason`, `Close date`

stash86: /status <trade_id> works for open and closed trades as long as I have been using FT

xmatthias: that's ... not incorrect apparently ...

we maybe should adjust naming a bit - showing "current_rate" in that case is quite confusing - as i'm currently not quite sure if that's the actual close rate, or today's rate (which is irrelevant for a trade closed 2 months ago).

stash86: Yeah, we can hide `current rate` if the trade is closed. Maybe some other info as well can be hidden, like the distance of time (since xxx ago), and the stoploss and stoploss distance?

xmatthias: yeah - but i'd separate that part from this PR ... as it's not really related to this change.

xmatthias (approved):
LGTM :+1: 
Nice enhancement

PR #6275, opened 2022-01-22
Extract timedout from ftbot

xmatthias:
## Summary
Extract timeout handling from freqtradebot class.

Simplifies the tradebot class.

* Adds `current_time` argument for easier usage of this callback.

### DIFF
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -188,7 +188,8 @@ class IStrategy(ABC, HyperStrategyMixin):
         """
         return dataframe
 
-    def check_buy_timeout(self, pair: str, trade: Trade, order: dict, **kwargs) -> bool:
+    def check_buy_timeout(self, pair: str, trade: Trade, order: dict,
+                          current_time: datetime, **kwargs) -> bool:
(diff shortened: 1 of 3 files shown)


freqtrade/strategy/interface.py
```diff
@@ -852,6 +856,29 @@ def min_roi_reached(self, trade: Trade, current_profit: float, current_time: dat
...
+        ordertime = arrow.get(order['datetime']).datetime
+        if timeout is not None:
+            timeout_unit = self.config.get('unfilledtimeout', {}).get('unit', 'minutes')
+            timeout_kwargs = {timeout_unit: -timeout}
+            timeout_threshold = current_time + timedelta(**timeout_kwargs)
+            timedout = (order['status'] == 'open' and order['side'] == side
+                        and ordertime < timeout_threshold)
+            if timedout:
+                return True
+        time_method = 'check_sell_timeout' if order['side'] == 'sell' else 'check_buy_timeout'
```
rokups: could just as well use `self.check_sell_timeout` and `self.check_buy_timeout` instead of strings.

PR #6276, opened 2022-01-22
Add support for shorts in strategy.stoploss_from_absolute()

clover-es:
## Summary

Add support for shorts in strategy.stoploss_from_absolute()

Solves # TODO-lev for "stoploss_from_absolute"

## Quick changelog

Add support for shorts in strategy.stoploss_from_absolute()
Add short trade test cases for relevant test.
Change strategy.stoploss_from_open() short parameter name to follow same naming as in the other helper (I've seen most times `is_short` before `for_short`, so I though it was better to stick to that parameter name. Made a separate commit, in case this change might be rejected.

### DIFF
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -283,11 +283,11 @@ class AwesomeStrategy(IStrategy):
 
         # evaluate highest to lowest, so that highest possible stop is used
         if current_profit > 0.40:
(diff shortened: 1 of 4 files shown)


xmatthias (approved):
I've added some lines to the documentation we previously missed (also in the prior PR) - so it's clear how these methods should be called

LGTM :+1:

PR #6279, opened 2022-01-23
Fix position adjustment not counting older orders because of missing average field value

### LINKED ISSUE #6268: Problems with first order after enabling DCA
## Describe your environment

  * Operating system: ____ Mac
  * Freqtrade Version: ____ docker-730d2e35
  
## Describe the problem:

After setting up a Freqtrade instance in dry mode, with a copy of a live db, and enabling DCA i get:

* Wrong order management
* Wrong average cost calculation
* Wrong profit

### Steps to reproduce:

1. make a copy of a live db, FT stable version, with some open trades
2. setup a dry develop image of FT with the live db copy
3. enable DCA
  
### Observed Results:

  * What happened?
 
1. FT (the strategy) rebuys a coin
2. FT (the strategy) sells the coin (only the amount from last buy) at some price that is not even near the average and declares some incorrect profit (taking in consideration the total amount for the coin)
3. FT closes the whole trade from db (note the amount sold was only the amount from last buy ) 
  

2. FT (the strategy) sells the coin (total amount) at some price higher than the average cost and declares the real profit
3. FT closes the whole trade from db only if it completely sells the coin

  ```


✓ Binance: Bought WRX/USDT (#66)
Buy Tag: 39 
Amount: 15577.9 
Open Rate: 1.53600000
Total: (23951.68188285 USDT)

✓ Binance: Bought WRX/USDT (#66)
Buy Tag: 39 
Amount: 7590.60000000
Open Rate: 0.94800000
Total: (7195.889 USDT)

  ✳ Binance: Sold WRX/USDT (#66)
Profit: 0.01% (profit: 0.77424120 USDT / 0.774 USD)
Buy Tag: 39 
Sell Reason: sell_profit_t_0_3 ( 39 )
Duration: 49 days, 15:20:18 (71480.3 min)
Amount: 7590.60000000
Open Rate: 0.94800000
Close Rate: 0.95000000
  ```

xataxxx:
## Summary

When moving from older live database to newer DCA enabled dry-mode, user encountered a bug where their orders without average price were not counted for after DCA orders kicked in.

Solve the issue: #6268 

## Quick changelog

- Fix by using price if average is not available
- Added unit-test to confirm

## What's new?

Additionally added a check to make sure order is closed.

### DIFF
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -584,10 +584,10 @@ class LocalTrade():
                 continue
 
             tmp_amount = o.amount
-            tmp_price = o.average or o.price or 0.0
+            tmp_price = o.average or o.price
             if o.filled is not None:
                 tmp_amount = o.filled
-            if tmp_amount > 0.0 and tmp_price is not None and o.status == 'closed':
+            if tmp_amount > 0.0 and tmp_price is not None:
                 total_amount += tmp_amount
                 total_stake += tmp_price * tmp_amount
 
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -587,7 +587,7 @@ class LocalTrade():
             tmp_price = o.average or o.price or 0.0
             if o.filled is not None:
                 tmp_amount = o.filled
-            if tmp_amount > 0.0 and tmp_price is not None:
+            if tmp_amount > 0.0 and tmp_price is not None and o.status == 'closed':
                 total_amount += tmp_amount
                 total_stake += tmp_price * tmp_amount
 
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -577,18 +577,19 @@ class LocalTrade():
 
         total_amount = 0.0
         total_stake = 0.0
-        for temp_order in self.orders:
-            if (temp_order.ft_is_open or
-                    (temp_order.ft_order_side != 'buy') or
-                    (temp_order.status not in NON_OPEN_EXCHANGE_STATES)):
+        for o in self.orders:
+            if (o.ft_is_open or
+                    (o.ft_order_side != 'buy') or
+                    (o.status not in NON_OPEN_EXCHANGE_STATES)):
                 continue
 
-            tmp_amount = temp_order.amount
-            if temp_order.filled is not None:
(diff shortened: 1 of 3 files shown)


freqtrade/persistence/models.py
```diff
@@ -577,18 +577,19 @@ def recalc_trade_from_orders(self):
...
 
-            tmp_amount = temp_order.amount
-            if temp_order.filled is not None:
-                tmp_amount = temp_order.filled
-            if tmp_amount > 0.0 and temp_order.average is not None:
+            tmp_amount = o.amount
+            tmp_price = o.average or o.price or 0.0
+            if o.filled is not None:
+                tmp_amount = o.filled
+            if tmp_amount > 0.0 and tmp_price is not None and o.status == 'closed':
```
xmatthias: i don't think you should check for `'closed'` here explicitly. That will exclude partially filled (but cancelled ) orders - and this filter is already in place in L583.

freqtrade/persistence/models.py
```diff
@@ -577,18 +577,19 @@ def recalc_trade_from_orders(self):
...
+                    (o.ft_order_side != 'buy') or
+                    (o.status not in NON_OPEN_EXCHANGE_STATES)):
                 continue
 
-            tmp_amount = temp_order.amount
-            if temp_order.filled is not None:
-                tmp_amount = temp_order.filled
-            if tmp_amount > 0.0 and temp_order.average is not None:
+            tmp_amount = o.amount
+            tmp_price = o.average or o.price or 0.0
```
xmatthias: If you don't fall back to 0.0 - you can check for `None` and ignore that order - which is simpler / makes the intend clearer.
If the order didn't have a price, nor an average, it can't be considered.

by falling back to 0.0, you'd add a stake of 0.0 below - which is pointless (but kindof hides the intend you had to exclude that order).

xataxxx: Agreed. Fix implemented.

freqtrade/persistence/models.py
xataxxx: Yes, this would fail with partial fills. My bad, fixed.

xmatthias (approved):
Great, thanks for the quick fix! :+1:

PR #6282, opened 2022-01-23
Fix missing order time data in backtesting.

xataxxx:
## Summary

Current backtesting creates orders without order_date, order_filled_date, order_update_date values.
This fails when strategy is attempting to use those values to calculate the time-difference between last order and current time.

## Quick changelog

- Added missing time info 

## What's new?

Now backtesting is able to calculate the time-difference between last buy and current tick.


xmatthias (approved):

xmatthias (approved):
LGTM :+1: 

on a side note: you might wanna create feature branches for PR's, not create pull requests from develop directly, otherwise you might encounter problems when pulling upstream (in case a PR cannot be merged soon)

Issue #6009, opened 2021-11-30
`HealthCheck` API endpoint


MNThomson:
## Describe the enhancement

In the docs, I'm not seeing any `/api/v1/healthcheck` endpoint or whatnot (heartbeat specific, not referencing `ping`). Potentially returning the time of last heartbeat and any errors (if present).

This would be highly useful for monitoring and ensuring nothing crazy is going wrong. This would make monitoring from another server much easier and with some simple scripts, could alert the user immediately if either the server or the bot is down.

xmatthias:
It's unclear what you'd expect about "crazy going wrong".
if something is "crazy going wrong" - the bot will stop working, which means ping will stop responding.
obviously it could check for the database connection - but that's about it.

Other errors (like exchange errors, api key errors, whatever else ...) are pretty difficult to reliably detect, as they can come in quite many different forms, and (if known) are also handled appropriately to avoid crashing the bot - so they won't bubble up to an overarching exception handler.

MNThomson:
Fair point. Would be nice for peace of mind to get the last heartbeat time.

eeddaann:
@xmatthias, what do you think about creating Prometheus exporter which will expose vital metrics?
I'm not sure if I should instrument the source code or should I decouple it and develop the exporter as a different project which will query the api.
what do you think?

xmatthias:
I've got no idea what a "prometheus exporter" would imply - which makes it quite difficult to make an informed decision.

PR #6283, opened 2022-01-23
/health api and telegram endpoint

xataxxx:
## Summary

Add ability to see the last process timestamp, meaning when was the last time freqtradebot internal process method finished.

Solve the issue: #6009 

## Quick changelog

- Added RPC endpoint that returns JSON {'last_process':'1970-01-01 00:00:00'}
- Added Telegram /health command

## What's new?

Ability to see last process timestamp.
We return 1970-01-01 00:00:00 by default until the first process cycle begins.


xataxxx:
Oops, wrong code, will try again with proper codebase.

PR #6294, opened 2022-01-24
/health api and telegram commands to return last processing time

### LINKED ISSUE #6009: `HealthCheck` API endpoint

xataxxx:
## Summary

Add ability to see the last process timestamp, meaning when was the last time freqtradebot internal process method finished.

Solve the issue: #6009 

## Quick changelog

- Added API / RPC endpoint that returns JSON {'last_process':'1970-01-01 00:00:00'}
- Added Telegram /health command

## What's new?

Ability to see last process timestamp.
We return 1970-01-01 00:00:00 by default until the first process cycle begins.


freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -382,3 +382,7 @@ class BacktestResponse(BaseModel):
 class SysInfo(BaseModel):
     cpu_pct: List[float]
     ram_pct: float
+
+
+class Health(BaseModel):
+    last_process: datetime
```
xmatthias, line 388: Especially for the API, i think we would want a `last_process_ts` as well (as int - as we do with other endpoints).

otherwise we rely on the client to correctly parse the timestamp - which is usually painful and quite errorprone as i've experienced myself.

freqtrade/freqtradebot.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self._exit_lock = Lock()
         LoggingMixin.__init__(self, logger, timeframe_to_seconds(self.strategy.timeframe))
 
+        self.last_process = datetime.utcfromtimestamp(0.0)
```
xmatthias: using `utcfromtimestamp` is dangerous, as it generates a timezone-native object.

if you run `datetime.utcfromtimestamp(0.0).timestamp()` - then the result is -3600 (my timezone is +1).

freqtrade/rpc/rpc.py
```diff
@@ -1030,3 +1031,11 @@ def _rpc_sysinfo() -> Dict[str, Any]:
             "cpu_pct": psutil.cpu_percent(interval=1, percpu=True),
             "ram_pct": psutil.virtual_memory().percent
         }
+
+    def _health(self) -> Dict[str, str]:
+        last_p = self._freqtrade.last_process
+        return {
+            'last_process': str(last_p),
+            'last_process_loc': last_p.astimezone(tzlocal()).strftime(DATETIME_PRINT_FORMAT),
```
xmatthias, line 1039: I've now made some changes to this ...
especially this part can be argued ... this will move the timestamp to "local time" (in server timezone) ...

xataxxx: Thanks, been buzy with life/work, I can sort this out in the weekend though.

xmatthias: I've pushed "something" already (i'm sometimes faster fixing stuff than commenting "please fix" :laughing: )

 ... this is more a "do we want it this way" question

xmatthias (approved):

xmatthias (approved):
LGTM :+1: 

We'll try this with "current timezone" for telegram - we'll see if this will raise questions ...

PR #6303, opened 2022-01-26
Added note to docker_quickstart.md

Wings22Actual:
## Summary

Added note that 'docker run --rm' can be used instead of 'docker-compose run --rm' for some commands

## Quick changelog

-   line 128-132

### DIFF
--- a/docs/docker_quickstart.md
+++ b/docs/docker_quickstart.md
@@ -125,11 +125,12 @@ All freqtrade arguments will be available by running `docker-compose run --rm fr
 
 !!! Note "`docker-compose run --rm`"
     Including `--rm` will remove the container after completion, and is highly recommended for all modes except trading mode (running with `freqtrade trade` command).
-    
-!!! Note that "`docker-compose run --rm`" will require a compose file to be provided.
-Some freqtrade commands that don't require authentication such as `list-pairs` can be run with "`docker run --rm`" instead.
-For example `docker run --rm freqtradeorg/freqtrade:stable list-pairs --exchange binance --quote BTC --print-json`. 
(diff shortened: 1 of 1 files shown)


docs/docker_quickstart.md
```diff
@@ -125,6 +125,11 @@ All freqtrade arguments will be available by running `docker-compose run --rm fr
 
 !!! Note "`docker-compose run --rm`"
     Including `--rm` will remove the container after completion, and is highly recommended for all modes except trading mode (running with `freqtrade trade` command).
+    
+!!! Note that "`docker-compose run --rm`" will require a compose file to be provided.
+Some freqtrade commands that don't require authentication such as `list-pairs` can be run with "`docker run --rm`" instead.
+For example `docker run --rm freqtradeorg/freqtrade:stable list-pairs --exchange binance --quote BTC --print-json`. 
+This can be useful for fetching exchange information to add to your `config.json` without affecting your running containers.
```
xmatthias:
```suggestion
??? Note "Using docker without docker-compose"
    "`docker-compose run --rm`" will require a compose file to be provided.
    Some freqtrade commands that don't require authentication such as `list-pairs` can be run with "`docker run --rm`" instead.
    For example `docker run --rm freqtradeorg/freqtrade:stable list-pairs --exchange binance --quote BTC --print-json`. 
    This can be useful for fetching exchange information to add to your `config.json` without affecting your running containers.
```

Note boxes require quite specific syntax to work / render correctly ... 
the best in that case is usually to use `mkdocs serve` - and navigate to the localhost version to investigate if the rendering is correct or not.

xmatthias (approved):
I've slightly updated formatting to have linebreaks render properly

PR #6306, opened 2022-01-27
add `/forcelong` and `/forceshort` commands

xmatthias:
## Summary
Add `/forcelong` (same as `/forcebuy`) and `/forceshort` capabilities to telegram

## Quick changelog

- Add `/forceshort`
- Add `/forcelong`
- add Additional parameters to API to support short trades (endpoint is now named `/forceenter`, but `/forcebuy` still works.

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -7,6 +7,7 @@ import json
 import logging
 import re
 from datetime import date, datetime, timedelta
+from functools import partial
 from html import escape
 from itertools import chain
 from math import isnan
@@ -23,6 +24,7 @@ from telegram.utils.helpers import escape_markdown
 from freqtrade.__init__ import __version__
 from freqtrade.constants import DUST_PER_COIN
 from freqtrade.enums import RPCMessageType
+from freqtrade.enums.signaltype import SignalDirection
 from freqtrade.exceptions import OperationalException
 from freqtrade.misc import chunks, plural, round_coin_value
 from freqtrade.persistence import Trade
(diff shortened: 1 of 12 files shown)


freqtrade/rpc/telegram.py
```diff
@@ -113,7 +116,8 @@ def _init_keyboard(self) -> None:
                                  r'/stopbuy$', r'/reload_config$', r'/show_config$',
                                  r'/logs$', r'/whitelist$', r'/blacklist$', r'/bl_delete$',
                                  r'/weekly$', r'/weekly \d+$', r'/monthly$', r'/monthly \d+$',
-                                 r'/forcebuy$', r'/edge$', r'/help$', r'/version$']
+                                 r'/forcebuy$', r'/forcelong$', r'/forceshort$',
+                                 r'/edge$', r'/help$', r'/version$']
```
samgermain, line 120: I kinda feel like there should only be a `/forceenter` and `forceexit` just to clean things up, I don't really see the need to separate orders into long and short

xmatthias: the problem is - when you need to type this in telegram, it becomes very clumsy to type, so UX is bad.

if you use `/forceenter` - you need another argument (usability of additional arguments suck, as they can be wrong and you won't know until you have to retype it).

so it'd become `/forceenter long XRP/USDT 1.5` .... as opposed to `/forcelong XRP/USDT 1.5` - which is shorter to type (this is something to look out for, considering telegram is "mostly" used on phones!).

for the API, i agree, and that's how it's implemented.

`/forceexit` works as "one" command as the direction there is clear from the trade id

PR #6307, opened 2022-01-27
Remove shift in analyzed dataframe columns

### LINKED ISSUE #6234: Lag between trigger and order execution for strategy callbacks
## Describe your environment

  * Operating system: MacOS 12.0.1
  * Python Version: n/a
  * CCXT version: n/a
  * Freqtrade Version: freqtrade docker-6c4b2614
  
## Your question

I've been experimenting with the newly released position adjustment feature (using `adjust_trade_position`) and noticed the entry price (i.e. `current_rate`) is delayed by a couple of bars compared to the actual trigger.

For instance, if the last candle is retrieved as:
```
dataframe, _ = self.dp.get_analyzed_dataframe(trade.pair, self.timeframe)
candle = dataframe.iloc[-1].squeeze()
```

by returning a valid stake size, the actual order will be placed at the open price of the second bar after the trigger. 

<table>
<tr>
	<td>candle #
	<td>open rate
	<td>trigger
	<td>order rate
	<td>triggered by
<tr>
	<td>1
	<td>100
	<td>buy
	<td>
	<td>
<tr>
	<td>2
	<td>110
	<td>
	<td>110
	<td>candle # 1
<tr>
	<td>3
	<td>120
	<td>
	<td>
	<td>
<tr>
	<td>3
	<td>130
	<td>adjust_trade
	<td>
	<td>
<tr>
	<td>4
	<td>140
	<td>
	<td>
	<td>
<tr>
	<td>5
	<td>150
	<td>
	<td>150
	<td>candle # 3
<tr>
</table>

I think the ideal logic would be to execute the order at the open price right after the trigger bar has been spotted (which is the current behaviour for buy signals triggered using the `populate_buy_trend` method).

I can see why this is happening (e.g. the callbacks are called once per bar and the result will be used only in the following bar) and that a quick fix could be to just use the `timeframe-detail` argument. However, I think the best way to handle this would be to use the "previous open_rate" as `current_rate` since that should approximate better the live execution. Another issue is that the timeframe-detail argument is not supported in hyperopt.

xmatthias:
## Summary
Remove shift in buy/sell signals when using `get_analyzed_dataframe` in backtesting

Closes #6234 

## Quick changelog

- Add test
- remove shift

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -275,15 +275,19 @@ class Backtesting:
             # Trim startup period from analyzed dataframe
             df_analyzed = processed[pair] = pair_data = trim_dataframe(
                 df_analyzed, self.timerange, startup_candles=self.required_startup)
+            # Update dataprovider cache
+            self.dataprovider._set_cached_df(pair, self.timeframe, df_analyzed)
+
+            # Create a copy of the dataframe before shifting, that way the buy signal/tag
+            # remains on the correct candle for callbacks.
+            df_analyzed = df_analyzed.copy()
+
             # To avoid using data from future, we use buy/sell signals shifted
             # from the previous candle
             df_analyzed.loc[:, 'buy'] = df_analyzed.loc[:, 'buy'].shift(1)
             df_analyzed.loc[:, 'sell'] = df_analyzed.loc[:, 'sell'].shift(1)
             df_analyzed.loc[:, 'buy_tag'] = df_analyzed.loc[:, 'buy_tag'].shift(1)
             df_analyzed.loc[:, 'exit_tag'] = df_analyzed.loc[:, 'exit_tag'].shift(1)
-
-            # Update dataprovider cache
-            self.dataprovider._set_cached_df(pair, self.timeframe, df_analyzed)
 
             df_analyzed = df_analyzed.drop(df_analyzed.head(1).index)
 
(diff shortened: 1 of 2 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -275,16 +275,20 @@ def _get_ohlcv_as_lists(self, processed: Dict[str, DataFrame]) -> Dict[str, Tupl
             # Trim startup period from analyzed dataframe
             df_analyzed = processed[pair] = pair_data = trim_dataframe(
                 df_analyzed, self.timerange, startup_candles=self.required_startup)
+            # Update dataprovider cache
+            self.dataprovider._set_cached_df(pair, self.timeframe, df_analyzed)
+
+            # Create a copy of the dataframe before shifting, that way the buy signal/tag
+            # remains on the correct candle for callbacks.
+            df_analyzed = df_analyzed.copy()
```
xmatthias, line 283: @rokups you think this (combined with the comment above) is clear enough, or should we use a different variable name to make it clear that we're using a copy from here onwards?

rokups: Looks clear enough to me. Im not sure we will ever be in a situation asking ourselves why our changes do not propagate to dataframe passed to strategies. Because there should be no such changes

xmatthias: no, i'm more worried about removal of this "copy" as it would appear unnecessary (maybe as part of refactorings or whatnot), not so much about "does not propagate".

rokups: Comment states the reason why copy exists. Why would anyone think it is unnecessary when comment says otherwise?

xmatthias: because i don't always read comments :rofl:

PR #6310, opened 2022-01-28
Fix ETH duplicate in CoinGecko

Verbalinsurection:
## Summary

Fix for duplicate entry for "ETH" in CoinGecko

## Quick changelog

- Select 'ethereum' on CoinGecko for stake_currency 'ETH'


xmatthias (approved):
if more "common" currencies encounter this duplicate symbol problem, i think we'll have to do this slightly differently (using a reusable mapping, most likely). but as long as it's "only" ETH - we should be fine.

LGTM :+1:

PR #6312, opened 2022-01-28
Short informative decorator

xmatthias:
## Summary
Fix some todo-Lev - improve informative handling with futures

### DIFF
--- a/freqtrade/strategy/informative_decorator.py
+++ b/freqtrade/strategy/informative_decorator.py
@@ -1,4 +1,5 @@
-from typing import Any, Callable, NamedTuple, Optional, Union
+from dataclasses import dataclass
+from typing import Any, Callable, Optional, Union
 
 from pandas import DataFrame
 
@@ -10,7 +11,8 @@ from freqtrade.strategy.strategy_helper import merge_informative_pair
 PopulateIndicators = Callable[[Any, DataFrame, dict], DataFrame]
 
 
-class InformativeData(NamedTuple):
+@dataclass
+class InformativeData:
     asset: Optional[str]
     timeframe: str
     fmt: Union[str, Callable[[Any], str], None]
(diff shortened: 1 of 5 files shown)


freqtrade/strategy/informative_decorator.py
```diff
@@ -54,12 +56,11 @@ def populate_indicators_1h(self, dataframe: DataFrame, metadata: dict) -> DataFr
     _timeframe = timeframe
     _fmt = fmt
     _ffill = ffill
+    _candle_type = CandleType.from_string(candle_type) if candle_type else None
```
rokups, line 61: be nice if we could also use an enum instead of having to use a string

freqtrade/strategy/interface.py
```diff
@@ -156,6 +157,10 @@ def __init__(self, config: dict) -> None:
                 if timeframe_to_minutes(informative_data.timeframe) < strategy_timeframe_minutes:
                     raise OperationalException('Informative timeframe must be equal or higher than '
                                                'strategy timeframe!')
+                if not informative_data.candle_type:
+                    informative_data = InformativeData(
+                        informative_data.asset, informative_data.timeframe, informative_data.fmt,
+                        informative_data.ffill, config['candle_type_def'])
```
rokups: Looks like named tuple becomes a wrong type. I would turn `InformativeData` to a simple class and modify one field

freqtrade/strategy/informative_decorator.py
xmatthias: you can for sure use an enum (`CandleType(CandleType.FUTURES)` returns CandleType.FUTURES) - but it's user-provided, so we can't rely on it - therefore we assume it's a string and "convert from string".

freqtrade/strategy/interface.py
xmatthias: hah - the fact that we're not python 3.7 only (actually 3.8 only) - we can now use Dataclasses ... this seems to be a good pick for that :+1:

PR #6316, opened 2022-01-29
exchange.get_max_pair_stake_amount

samgermain:
Gets the maximum tradeable stake_amount, and makes sure that the stake_amount doesn't exceed the max amount

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -46,7 +46,7 @@ DEFAULT_DATAFRAME_COLUMNS = ['date', 'open', 'high', 'low', 'close', 'volume']
 # it has wide consequences for stored trades files
 DEFAULT_TRADES_COLUMNS = ['timestamp', 'id', 'type', 'side', 'price', 'amount', 'cost']
 TRADING_MODES = ['spot', 'margin', 'futures']
-MARGIN_MODES = ['cross', 'isolated']
+MARGIN_MODES = ['cross', 'isolated', '']
 
 LAST_BT_RESULT_FN = '.last_result.json'
 FTHYPT_FILEVERSION = 'fthypt_fileversion'
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -691,8 +691,13 @@ class Exchange:
         pair: str,
         price: float,
         stoploss: float
-    ) -> Optional[float]:
-        return self._get_stake_amount_limit(pair, price, stoploss, 'max')
+    ) -> float:
+        max_stake_amount = self._get_stake_amount_limit(pair, price, stoploss, 'max')
+        if max_stake_amount is None:
+            # * Should never be executed
+            raise OperationalException(f'{self.name}.get_max_pair_stake_amount should'
+                                       'never set max_stake_amount to None')
+        return max_stake_amount
 
     def _get_stake_amount_limit(
         self,
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -9,7 +9,7 @@ import logging
 from copy import deepcopy
 from datetime import datetime, timedelta, timezone
 from math import ceil
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Literal, Optional, Tuple, Union
 
 import arrow
 import ccxt
@@ -677,33 +677,59 @@ class Exchange:
         else:
             return 1 / pow(10, precision)
 
-    def get_min_pair_stake_amount(self, pair: str, price: float, stoploss: float,
-                                  leverage: Optional[float] = 1.0) -> Optional[float]:
+    def get_min_pair_stake_amount(
+        self,
+        pair: str,
+        price: float,
+        stoploss: float,
+        leverage: Optional[float] = 1.0
+    ) -> Optional[float]:
+        return self._get_stake_amount_limit(pair, price, stoploss, 'min', leverage)
+
+    def get_max_pair_stake_amount(
+        self,
+        pair: str,
+        price: float,
+        stoploss: float
+    ) -> Optional[float]:
+        return self._get_stake_amount_limit(pair, price, stoploss, 'max')
+
+    def _get_stake_amount_limit(
+        self,
+        pair: str,
+        price: float,
+        stoploss: float,
+        limit: Literal['min', 'max'],
+        leverage: Optional[float] = 1.0
+    ) -> Optional[float]:
+
+        isMin = limit == 'min'
+
         try:
             market = self.markets[pair]
         except KeyError:
             raise ValueError(f"Can't get market information for symbol {pair}")
 
-        min_stake_amounts = []
+        stake_limits = []
         limits = market['limits']
-        if (limits['cost']['min'] is not None):
-            min_stake_amounts.append(
+        if (limits['cost'][limit] is not None):
+            stake_limits.append(
                 self._contracts_to_amount(
                     pair,
-                    limits['cost']['min']
+                    limits['cost'][limit]
                 )
             )
 
-        if (limits['amount']['min'] is not None):
-            min_stake_amounts.append(
+        if (limits['amount'][limit] is not None):
+            stake_limits.append(
                 self._contracts_to_amount(
                     pair,
-                    limits['amount']['min'] * price
+                    limits['amount'][limit] * price
                 )
             )
 
-        if not min_stake_amounts:
-            return None
+        if not stake_limits:
+            return None if isMin else float('inf')
 
         # reserve some percent defined in config (5% default) + stoploss
         amount_reserve_percent = 1.0 + self._config.get('amount_reserve_percent',
@@ -718,9 +744,9 @@ class Exchange:
         # for cost (quote, stake currency), so max() is used here.
         # See also #2575 at github.
         return self._get_stake_amount_considering_leverage(
-            max(min_stake_amounts) * amount_reserve_percent,
+            max(stake_limits) * amount_reserve_percent,
             leverage or 1.0
-        )
+        ) if isMin else min(stake_limits)  # TODO-lev: Account 4 max_reserve_percent in max limits?
 
     def _get_stake_amount_considering_leverage(self, stake_amount: float, leverage: float):
         """
@@ -2118,40 +2144,6 @@ class Exchange:
         else:
             raise OperationalException(
                 "Freqtrade only supports isolated futures for leverage trading")
-
-    def get_max_amount_tradable(self, pair: str, price: float) -> float:
-        '''
-        Gets the maximum amount of a currency that the exchange will let you trade
-        '''
-        try:
-            market = self.markets[pair]
-        except KeyError:
-            raise ValueError(f"Can't get market information for symbol {pair}")
-
-        market = self.markets[pair]
-        limits = market['limits']
-        max_amounts = []
-
-        if (limits['cost']['max'] is not None):
-            max_amounts.append(
-                self._contracts_to_amount(
-                    pair,
-                    limits['cost']['max'] / price
-                )
-            )
-
-        if (limits['amount']['max'] is not None):
-            max_amounts.append(
-                self._contracts_to_amount(
-                    pair,
-                    limits['amount']['max']
-                )
-            )
-
-        if not max_amounts:
-            return float('inf')
-        else:
-            return min(max_amounts)
 
 
 def is_exchange_known_ccxt(exchange_name: str, ccxt_module: CcxtModuleType = None) -> bool:
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -935,6 +935,7 @@ class Exchange:
 
     def create_order(
         self,
+        *,
         pair: str,
         ordertype: str,
         side: str,
@@ -967,7 +968,7 @@ class Exchange:
                 side,
                 amount,
                 rate_for_order,
-                params
+                params,
             )
             self._log_exchange_response('create_order', order)
             order = self._order_contracts_to_amount(order)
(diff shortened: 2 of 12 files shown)


xmatthias:
i guess the correct place to call it is very similar to `get_min_pair_stake_amount()` ... as we'll have to have the final amount capped limited to both sides (upper and lower).

freqtrade/exchange/exchange.py
```diff
@@ -723,9 +723,12 @@ def get_min_pair_stake_amount(self, pair: str, price: float, stoploss: float,
...
         # See also #2575 at github.
-        return self._get_stake_amount_considering_leverage(
-            max(min_stake_amounts) * amount_reserve_percent,
-            leverage or 1.0
+        return min(
+            self._get_stake_amount_considering_leverage(
+                max(min_stake_amounts) * amount_reserve_percent,
+                leverage or 1.0
+            ),
+            min(max_stake_amounts)
```
xmatthias: i don't think that'll work (i assume you misunderstood me).

what we try to do is "you can trade between 15$ and 1_000_000$" (assuming that's the maximum) - so the stake-amount must be between these 2.

What this function is now doing:
`min(15, 1_000_000)` - which makes the whole "max" irrelevant - assuming max is always above min - which i would strongly assume).

We'll still need the separate function (or rename this function and return a tuple) - but it's 2 values we'll need to properly have stake-amount between min and max values.

samgermain:
> i guess the correct place to call it is very similar to `get_min_pair_stake_amount()` ... as we'll have to have the final amount capped limited to both sides (upper and lower).

I don't think that'll work actually, the max_amount is the max after leverage is applied, but I believe the min_stake amount is calculated before leverage is applied

xmatthias:
> I don't think that'll work actually, the max_amount is the max after leverage is applied, but I believe the min_stake amount is calculated before leverage is applied

It's a tradeoff anyway - we need leverage to determine minimum stake (look at the comment before `get_min_pair_stake_amount` in freqtradebot.py), but we need the actual stake to determine leverage.
the same exact thing will apply for max-stake - "maybe" an option is to call it twice - once to get an estimate - once to get the final value (after calling the strategy callbacks) ... 

Point is - we need all of these values - and all of these values can be set by the user ... 
but to set them with an "informed" decision - you need minimum/maximum values to know what you're actually allowed to set.

samgermain:
> > I don't think that'll work actually, the max_amount is the max after leverage is applied, but I believe the min_stake amount is calculated before leverage is applied
> 
> It's a tradeoff anyway - we need leverage to determine minimum stake (look at the comment before `get_min_pair_stake_amount` in freqtradebot.py), but we need the actual stake to determine leverage. the same exact thing will apply for max-stake - "maybe" an option is to call it twice - once to get an estimate - once to get the final value (after calling the strategy callbacks) ...
> 
> Point is - we need all of these values - and all of these values can be set by the user ... but to set them with an "informed" decision - you need minimum/maximum values to know what you're actually allowed to set.

I don't know if it matters but I'm looking at the max amount, not the max stake amount, because the exchange will not let you place an order above the max amount

xmatthias:
> I don't know if it matters but I'm looking at the max amount, not the max stake amount, because the exchange will not let you place an order above the max amount

The currency doesn't really matter (if it's stake or amount) - it's the identical problem at the top and bottom ... you also cannot place below the minimum amount, and you can't place above the maximum amount.

In Reality, it's the aggregation of both limits (cost and amount) at both sides (lower limit, higher limit) - and the one that's applied will differ based on the current price (there's been a few samples where minimum amount was 50+$ - as the currency was raising a few 100% in days prior to the "problem" day) - and exchanges don't adjust these limits daily - usually at max monthly or rather - as necessary - so we must look at both cost and amount maximum (aligned to how we do it for the minimum).

The only difference with the max is that it'll need to be applied after leverage - which is not the case for the min (leverage will not reduce the amount).

samgermain:
@xmatthias I guess the max stake amount would be a stake amount with no leverage applied

samgermain:
@xmatthias I think the newest changes might be all that's needed (besides a few more tests). The only thing that concerns me is there's no check when the the max amount limit is `None` inside the market, but it doesn't look like freqtrade does anything about this case for the min amounts either

samgermain:
@xmatthias Let me know what you think of the PR, I'll add tests if you think it's good

freqtrade/freqtradebot.py
```diff
@@ -543,12 +543,19 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
         min_stake_amount = self.exchange.get_min_pair_stake_amount(trade.pair,
                                                                    current_rate,
                                                                    self.strategy.stoploss)
-        max_stake_amount = self.wallets.get_available_stake_amount()
+        max_stake_amount = self.exchange.get_max_pair_stake_amount(trade.pair,
+                                                                   current_rate,
+                                                                   self.strategy.stoploss)
+        if max_stake_amount is None:
+            # * Should never be executed
+            raise OperationalException(f'max_stake_amount is None for {trade}')
```
xmatthias: Is this even possible? 
i don't think there's any path where a call to `max_pair_stake_amount` can return None - so this is really pointless.
i do this too for debugging, sometimes - but the endresult shouldn't have that (otherwise we'd need the same thing for every function we call - which we clearly don't and shouldn't).

if you think we need this check (I don't think we do) - it should raise a DependencyException. The difference is - OperationalException stops the bot - while Dependency will continue with the next pair.

freqtrade/exchange/exchange.py
```diff
@@ -677,33 +677,59 @@ def price_get_one_pip(self, pair: str, price: float) -> float:
...
+    ) -> Optional[float]:
+        return self._get_stake_amount_limit(pair, price, stoploss, 'max')
+
+    def _get_stake_amount_limit(
+        self,
+        pair: str,
+        price: float,
+        stoploss: float,
+        limit: Literal['min', 'max'],
+        leverage: Optional[float] = 1.0
```
xmatthias, line 707: I would maybe add a `*,` argument here - forcing the use of keyword arguments after that - that'll make sure it's not called wrongly / with mistaking arguments

that's a nitpick though - overall i like this concept :+1:

freqtrade/freqtradebot.py
samgermain: It's not possible, it's because `Exchange._get_stake_amount_limit` can return `None` when `limit='min'`, but I moved the check to the Exchange class inside `get_max_pair_stake_amount`. It's there to prevent programmer and mypy errors, but I can also just set `# ignore` inside `get_max_pair_stake_amount`

freqtrade/exchange/exchange.py
samgermain: I don't really understand this, doesn't a star argument mean you can pass a variable number of arguments in?

xmatthias: there's a difference - `*args, **kwargs` simply collects "all" arguments provided, and provides `args` as list, and `kwargs` as dict (with provided keywordarguments as content).

if you define a function as this:
`def whatever(*, pair, x):` - then all arguments after the * are required as kwargs
so a call `whatever(pair, x)` will be a syntaxerror, while `whatever(pair=pair, x=x)` will  work (it's the only way to have this work, actually).

samgermain:
I wrote a test inside `execute_entry` and it looks like `max_stake_amount` is being ignored

```
        # In case of too high stake amount
    
        order['status'] = 'open'
        order['id'] = '55672'
    
        mocker.patch.multiple(
            get_max_pair_stake_amount=MagicMock(return_value=500),
        )
        freqtrade.exchange.get_max_pair_stake_amount = MagicMock(return_value=500)
    
        assert freqtrade.execute_entry(pair, 2000, is_short=is_short)
        trade = Trade.query.all()[9]
        trade.is_short = is_short
>       assert trade.stake_amount == 500
E       assert 713.0 == 500
E        +  where 713.0 = Trade(id=10, pair=ETH/USDT, amount=71.30000000, is_short=False, leverage=1.0, open_rate=10.00000000, open_since=2022-02-03 07:20:11).stake_amount

tests/test_freqtradebot.py:950: AssertionError
```

freqtrade/constants.py
```diff
@@ -46,7 +46,7 @@
 # it has wide consequences for stored trades files
 DEFAULT_TRADES_COLUMNS = ['timestamp', 'id', 'type', 'side', 'price', 'amount', 'cost']
 TRADING_MODES = ['spot', 'margin', 'futures']
-MARGIN_MODES = ['cross', 'isolated']
+MARGIN_MODES = ['cross', 'isolated', '']
```
samgermain: @xmatthias I think you removed this, it looks like the `test_commands` tests fail without it

xmatthias: what i changed was the variable name to MARGIN_MODES - but i didn't touch the content ... 

image

samgermain: I forgot this change hadn't be committed yet, I changed it on the `todos` branch, sorry

samgermain:
@xmatthias Everything should be good I think, let me know if you want anything else updated

freqtrade/freqtradebot.py
```diff
@@ -821,10 +825,13 @@ def cancel_stoploss_on_exchange(self, trade: Trade) -> Trade:
...
-            self, pair: str, price: Optional[float], stake_amount: float,
-            side: str, trade_side: str,
-            entry_tag: Optional[str],
-            trade: Optional[Trade]) -> Tuple[float, float]:
+        self, pair: str, price: Optional[float], stake_amount: float,
+        side: str, trade_side: str,
+        entry_tag: Optional[str],
+        trade: Optional[Trade]
+    ) -> Tuple[float, float]:
+        # TODO: This method has no tests
```
xmatthias: I'm not sure about this ...
it might not have explicit tests, only testing this behaviour - but the content itself is tested as part of execute_entry

tests/test_freqtradebot.py
```diff
@@ -743,6 +743,7 @@ def test_execute_entry(mocker, default_conf_usdt, fee, limit_order,
         (open_rate - (wallet_balance / position)) / (1 - (mm_ratio + taker_fee_rate))
         (10 - (2 / 1)) / (1 - (0.01 + 0.0006)) = 8.085708510208207
     """
+    # TODO: Parametrize this test
```
xmatthias: This test is parametrized.
parametrizing it further will only obscure what it's actually testing.

if any, it should be split up into multiple tests.

freqtrade/exchange/exchange.py
```diff
@@ -677,33 +677,64 @@ def price_get_one_pip(self, pair: str, price: float) -> float:
...
+        stoploss: float,
+        leverage: Optional[float] = 1.0
+    ) -> Optional[float]:
+        return self._get_stake_amount_limit(pair, price, stoploss, 'min', leverage)
+
+    def get_max_pair_stake_amount(
+        self,
+        pair: str,
+        price: float,
+        stoploss: float
```
xmatthias: I think stoploss should be hard-set to 0 for the max-case (and not provided to `get_max_stake_amount` at all).

the reason to have stoploss part of this method is the following:
if you place a buy order near the very bottom (say - for 11$ with 10$ as minimum limit)  - and then have stoplos 20% - that stop would mean a sell for 9$ - which the exchange will no longer allow as it's below minimium limits

at the top however, this is not relevant (if you bought 1_000_000 - you can still sell after a 20% drop (even though you should've sold earlier, but that's not freqtrade's problem).

The value itself (calculation result of `amount_reserve_percent`) is ignored anyway

samgermain: > I think stoploss should be hard-set to 0 for the max-case (and not provided to get_max_stake_amount at all).

A stoploss of 0 isn't correct for shorts, if it doesn't matter though I can just set it to that

> at the top however, this is not relevant (if you bought 1_000_000 - you can still sell after a 20% drop (even though you should've sold earlier, but that's not freqtrade's problem).

xmatthias: > A stoploss of 0 isn't correct for shorts

It's not about shorts/not shorts - but about stoploss values per se for the upper bound.

Max defines the upper limit (in your example, 5000). It does not matter if your stoploss is 5% or 50% - the upper bound will still be 5000. This is different to the behavior on the lower bound, where we factor in the stoploss to be able to  exit the trade (see example above).

If you carefully look at the implementation for max -  you'll notice that stoploss is used for calculations - but the result of the calculation is ignored in case of the "max" calculation. It's therefore only relevant for "min" - which is correct (but makes the stoploss parameter irrelevant to calls that use "max" as "limit".

samgermain: I know that the `max_reserve_percent` isn't used for the max amount, but I don't know if that's correct behaviour

xmatthias: if any, it'd be ROI that's needed, not stop 
Say we have a coin that trades from 10$ to 5000$ per order ...
we enter at minimum 15$ - and at maximum 5000$ ...
the problem with the 5k is - if it goes in our favor (it's now worth 5500$) - we might not be able to sell it anymore ...

long/short is not really relevant i think - as it's stake-currency we're talking here (mainly)

freqtrade/optimize/backtesting.py
```diff
@@ -414,14 +414,16 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
 
     def _get_adjust_trade_entry_for_candle(self, trade: LocalTrade, row: Tuple
                                            ) -> LocalTrade:
-
+        # TODO: Write tests
```
xmatthias: Can you explain what this refers to?
this method is tested by `test_backtesting_adjust_position.py`. It's not directly called - but that's not really necessary, as the whole backtesting flow is tested.

xmatthias (approved):
:+1:

PR #6330, opened 2022-01-31
Removed unnecessary CCXT market checks

samgermain:
I've been unifying the markets for all exchanges on CCXT. There may be a few niche exchanges that are missing limits/contractSize, but by the end of the week they'll all be unified.


xmatthias (approved):
Great :+1:

PR #6332, opened 2022-01-31
Binance max lev

samgermain:
binance.get_max_leverage correctly calculates the leverage

### DIFF
--- a/freqtrade/exchange/binance.py
+++ b/freqtrade/exchange/binance.py
@@ -155,11 +155,11 @@ class Binance(Exchange):
             except ccxt.BaseError as e:
                 raise OperationalException(e) from e
 
-    def get_max_leverage(self, pair: Optional[str], margin: Optional[float]) -> float:
+    def get_max_leverage(self, pair: str, stake_amount: Optional[float]) -> float:
         """
         Returns the maximum leverage that a pair can be traded at
         :param pair: The base/quote currency pair being traded
-        :margin: The total value of the traders collateral in quote currency
+        :stake_amount: The total value of the traders collateral in quote currency
         """
         if pair not in self._leverage_brackets:
             return 1.0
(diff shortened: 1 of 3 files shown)


freqtrade/exchange/binance.py
```diff
@@ -155,19 +155,23 @@ def fill_leverage_brackets(self):
...
         pair_brackets = self._leverage_brackets[pair]
         max_lev = 1.0
         for [min_amount, margin_req] in pair_brackets:
+            lev = 1/margin_req
+            nominal_value = stake_amount * lev
             if nominal_value >= min_amount:
-                max_lev = 1/margin_req
+                max_lev = lev
+            else:
+                break
```
xmatthias: I'm not sure i fully understand this calculation ... 
basically, if 0.01 is provided as `margin_req` - then it becomes `1 / 0.01 = 100`.
I'll now multiply my 100$ with that 100x leverage to get the nominal value - which we then compare with the leverage brakets (?) 

I'm also just now looking at [this binance help page](https://www.binance.com/en/support/announcement/34801a0c405a4b058f9ae18a1a34cad3) - which seems to complicate matters - as it seems like the comparison should be against both "before" and "after" notional value (after meaning after leverage is applied, i assume).

The API documentation on the other hand makes it pretty much unclear if this is talking about "before" or "after" stake amounts ...

samgermain: I think it was wrong, so I updated it

----------------------------

With the new method

if your stake_amount is 1000 and you have brackets like
```
[
    0, 0.1, # leverage = 1/0.1 = 10, nominal_value = 1000 * 10 = 10 000
    10 000, 0.25, # leverage = 1/0.25 = 4, nominal_value = 1000 * 4 = 4000
    1 000 000, 0.5, # leverage = 1/0.5 = 2, nominal_value = 1000 * 2 = 2000
]
```

So on the first bracket, we check if the nominal value is less than the min of the next bracket, the value is 10 000, and the next brackets min is 10 000, so we move up a bracket, now your nominal value is 4 000, which does not exceed the nexts brackets min, which is 1 000 000, so we return the leverage that we calculated for this bracket

xmatthias (approved):

PR #6333, opened 2022-01-31
New config

samgermain:
Updated `new-config` command to add `trading_mode` and `margin_mode` to a new config.

I thought it was safest to just set them as default to spot


config_examples/config_binance.example.json
```diff
@@ -1,95 +1,92 @@
...
+  "max_open_trades": 3,
+  "stake_currency": "BTC",
+  "stake_amount": 0.05,
+  "tradable_balance_ratio": 0.99,
+  "fiat_display_currency": "USD",
+  "timeframe": "5m",
+  "dry_run": true,
+  "cancel_open_orders_on_exit": false,
+  "trading_mode": "spot",
+  "margin_mode": null,
```
xmatthias: i'd apreciate if we could not change the whole file (maybe disable your autoformatter, or set it to 4 spaces for these files).
That's intentional - as it makes the file easier to read for a non-developer.

Also - i think margin_mode is really called "collateral" at the moment - so we will need to be very careful to have these different things aligned - so either rename it as part of this PR to margin_mode - or create it here "wrongly" - to adjust it later (otherwise we risk having parts of the code use one, and parts use the other key).

freqtrade/templates/base_config.json.j2
```diff
@@ -13,6 +13,8 @@
     "fiat_display_currency": "{{ fiat_display_currency }}",{{ ('\n    "timeframe": "' + timeframe + '",') if timeframe else '' }}
     "dry_run": {{ dry_run | lower }},
     "cancel_open_orders_on_exit": false,
+    "trading_mode": "spot",
+    "margin_mode": null,
```
xmatthias: I think this is pointless at the moment.

My expectation based on the "new config" title was that this would ask the user to select a trading mode.
As that's not the case, best remove these keys from the template for the moment. 

Even if they are in the template file, we cannot rely on the values to be there. 
If we need defaults - we have to set them in the schema in `constants.py`.

config_examples/config_binance.example.json
samgermain: I'm going to rename it after the `isolated-liq` pr get's merged

xmatthias:
let me ask a different question here - what's the point of this PR? (what are you trying to archive?)

having a 1:1 copy of configurations for futures makes little sense.
all documentation has the user use `new-config`  to create settings - and that's the settings-template that's the most maintained.

It makes sense to have the configuration for "full" - as that contains settings usually not set via configuration (so it serves as reference for users).
I'm even inclined to say that all example configs (except for full and one other that's used in CI) are not really necessary (unrelated to shorts/futures) - as they are basically just the output of running `freqtrade new-config` selecting appropriate values - so i'm not quite happy to add even more sample configs unless we have a really good reason to - as each additional file is additional work should "anything" change.

---

What's worse - the samples you provide - especially for futures - are non-working (not just because of the margin_type key).

I think a better approach is to remove all new samples (eventually reconsider existing samples - but I'd leave that for outside of the futures scope/branch), and instead add questions for "market type" to the "new config"  command (obviously depending on the selected exchange - as there's no point in asking "market type" for an exchange that only supports spot markets).

config_examples/config_okex_futures.example.json
```diff
@@ -0,0 +1,93 @@
...
+            "BAT/USDT:USDT",
+            "BCH/USDT:USDT",
+            "EOS/USDT:USDT",
+            "IOTA/USDT:USDT",
+            "LINK/USDT:USDT",
+            "LTC/USDT:USDT",
+            "NEO/USDT:USDT",
+            "XMR/USDT:USDT",
+            "XRP/USDT:USDT",
+            "XTZ/USDT:USDT"
```
samgermain: I thought this might be useful just for the pairlist

xmatthias: when i commented the above, this was not the pairlist in this file (which is why i said - it's wrong).

that said - i don't think there's much benefit to it (not enough to justify more sample files we fail to maintain properly, anyway ...).
you can get the same pairlist with `freqtrade list-pairs --exchange okex --trading-mode futures` (which will actually list ALL pairs - which a sample config will never be able to).

freqtrade/commands/build_config_commands.py
```diff
@@ -103,21 +103,32 @@ def ask_user_config() -> Dict[str, Any]:
             "message": "Please insert your display Currency (for reporting):",
             "default": 'USD',
         },
+        {
+            "type": "confirm",
+            "name": "trading_mode",
+            "message": "Do you want to trade Perpetual Swaps (perpetual futures)?",
+            "default": False,
+            "filter": lambda val: '"futures"' if val else '"spot"',
+        },
```
xmatthias: I would inverse the sequence for this.

asking for the exchange first - and then determine if we support futures/margin with it or not - that way we can avoid asking for futures if it's not necessary.

samgermain:
This test is failing, I'm not really sure what the error message is trying to tell me, or what to do with it

```
INFO     freqtrade.commands.build_config_commands:build_config_commands.py:245 Writing config to `coolconfig.json`.
INFO     freqtrade.commands.build_config_commands:build_config_commands.py:246 Please make sure to check the configuration contents and adjust settings to your needs.
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

mocker = <pytest_mock.plugin.MockerFixture object at 0x129e8e190>
caplog = <_pytest.logging.LogCaptureFixture object at 0x12a99a8e0>, exchange = 'bittrex'

    @pytest.mark.parametrize('exchange', ['bittrex', 'binance', 'kraken', 'ftx'])
    def test_start_new_config(mocker, caplog, exchange):
        wt_mock = mocker.patch.object(Path, "write_text", MagicMock())
        mocker.patch.object(Path, "exists", MagicMock(return_value=True))
        unlink_mock = mocker.patch.object(Path, "unlink", MagicMock())
        mocker.patch('freqtrade.commands.build_config_commands.ask_user_overwrite', return_value=True)
    
        sample_selections = {
            'max_open_trades': 3,
            'stake_currency': 'USDT',
            'stake_amount': 100,
            'fiat_display_currency': 'EUR',
            'timeframe': '15m',
            'dry_run': True,
            'trading_mode': 'spot',
            'exchange_name': exchange,
            'exchange_key': 'sampleKey',
            'exchange_secret': 'Samplesecret',
            'telegram': False,
            'telegram_token': 'asdf1244',
            'telegram_chat_id': '1144444',
            'api_server': False,
            'api_server_listen_addr': '127.0.0.1',
            'api_server_username': 'freqtrader',
            'api_server_password': 'MoneyMachine',
        }
        mocker.patch('freqtrade.commands.build_config_commands.ask_user_config',
                     return_value=sample_selections)
        args = [
            "new-config",
            "--config",
            "coolconfig.json"
        ]
        start_new_config(get_args(args))
    
        assert log_has_re("Writing config to .*", caplog)
        assert wt_mock.call_count == 1
        assert unlink_mock.call_count == 1
>       result = rapidjson.loads(wt_mock.call_args_list[0][0][0],
                                 parse_mode=rapidjson.PM_COMMENTS | rapidjson.PM_TRAILING_COMMAS)
E       rapidjson.JSONDecodeError: Parse error at offset 262: Invalid value.

tests/commands/test_build_config.py:71: JSONDecodeError
```

freqtrade/commands/build_config_commands.py
```diff
@@ -107,19 +107,29 @@ def ask_user_config() -> Dict[str, Any]:
             "type": "select",
             "name": "exchange_name",
             "message": "Select exchange",
-            "choices": [
+            "choices": lambda x: [
+                Separator("------ Spot and Perpetual Swaps/Futures -------"),
                 "binance",
+                "gateio",
+                Separator("----------------- Spot Only -------------------"),
```
xmatthias: I'd either leave sorting of the exchanges as it was - or have them sorted alphabetically.

This way it looks bad - and if we think about it from a user perspective - (s)he'll already have the exchange in mind (basically you use wherever you have an account) - so this information here is pointless.

Having accounts on most exchanges is something very few people do - so this is just another information that'll be more confusing than helpful.
We will have the "supported" exchanges in the docs - but this will mean that for every new exchange, we'll have to change ordering here - which will fail sometime in the future for sure.

freqtrade/commands/build_config_commands.py
```diff
@@ -196,7 +206,11 @@ def ask_user_config() -> Dict[str, Any]:
     if not answers:
         # Interrupted questionary sessions return an empty dict.
         raise OperationalException("User interrupted interactive questions.")
-
+    answers["margin_mode"] = (
+        '\n    "margin_mode": "isolated",'
+        if answers.get("trading_mode") == 'futures'
+        else ''
+    )
```
xmatthias: are we sure we want `margin_mode` to be missing from the configuration file if spot is selected?

I'd still provide it (but empty, maybe?) - which will make it easier on an existing config to switch to futures later on.

samgermain: Sure, I kind of just thought your didn't want it https://github.com/freqtrade/freqtrade/pull/6333#discussion_r795740253

freqtrade/commands/build_config_commands.py
samgermain: People are more likely to choose it if it's at the top, so I sorted them by their level of support, they might also only have an account on coinbase or something and none of the exchanges listed

freqtrade/commands/build_config_commands.py
xmatthias: hah no - that comment was against hard-coding 'spot' and "null" for it without asking the user for "what do you want" - but explicitly setting it to spot - and have the user "manually" change it to futures (they can still do this if they picked wrong - but it shouldn't be the default).

There's nothing wrong with having the key there - but the pr name ("new-config") suggested that we'd be asking users "you wanna trade futures?" - and then configure it properly (which this PR didn't do at the time i wrote this comment).

freqtrade/commands/build_config_commands.py
xmatthias: Please change sorting to alphabetic.
I think freqtrade should be neutral on this.

we list exchanges we support here for "easy selection" - and each exchange listed will get the same level of support (if something is broken on 2 exchanges, we're not going to fix gate.io, but leave kraken broken for weeks assuming the fix is possible). If we need to prioritize, then it's based on "users using that exchange" - which would put gate.io close to last (i only know a handful of people using gate.io - but 100ds using kucoin).

What you refer to is functionality (not support) - but even in that case your ordering is wrong - gate.io does not have stoploss orders, nor market orders - so that'd be close to last when it comes to listing.
The sequence is therefore highly subjective (you think about futures to be "the important" feature - but users may not think that as they care more about other features and are not interested in futures)

samgermain: Gateio has stoploss orders on CCXT, is there something in freqtrade that would need to be programmed?

xmatthias: Stoploss orders are not unified in ccxt AFAIK - so it'll need to be (slighlty) different code for every exchange.

If you can point me to "how" to do stoploss orders for gateio, i'll surely be adding this (to the develop branch, though).

xmatthias (approved):
:+1:

PR #6340, opened 2022-02-01
freqUI: fix can't import backtest with missing datetime data

TheJoeSchr:
## Summary
Fix: `ValueError: NaTType does not support strftime` when importing backtest results and wanting to watch charts.

By making datetime import of backtest results more robust it is first checking if variable is a valid datetime

I got the Error:

```shell
...
  File "/home/joe/projects/freqtrade/.venv/lib/python3.10/site-packages/fastapi/encoders.py", line 115, in jsonable_encoder
    jsonable_encoder(
  File "/home/joe/projects/freqtrade/.venv/lib/python3.10/site-packages/fastapi/encoders.py", line 47, in jsonable_encoder
    return encoder_instance(obj)
  File "/home/joe/projects/freqtrade/freqtrade/rpc/api_server/api_schemas.py", line 358, in <lambda>
    datetime: lambda v: v.strftime(DATETIME_PRINT_FORMAT),
  File "pandas/_libs/tslibs/nattype.pyx", line 75, in pandas._libs.tslibs.nattype._make_error_func.f
ValueError: NaTType does not support strftime
```

and I couldn't see any data graphs:
image

I'm not quite sure if checking for specific types is the right approach. Checking if datetime is None or NaT or null would be maybe better, but checking for NaT seems to need a special import of panda, so we could use `pd.isnull(datetime)`. I wasn't sure if importing `panda` in this file would have been the right approach or if this needs to stay "lightweight".
See also https://stackoverflow.com/questions/49435317/how-to-test-if-a-variable-is-pd-nat


freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -7,6 +7,11 @@
 from freqtrade.enums import OrderTypeValues
 
 
+def dtype_string(dtype):
+    result = str(dtype.dtype) if type(dtype) == "datetime" else type(dtype).__name__
```
xmatthias: i'd not do this as strings ... `type(dtype)` - but with `isinstance` instead.

freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -355,7 +360,9 @@ class PairHistory(BaseModel):
 
     class Config:
         json_encoders = {
-            datetime: lambda v: v.strftime(DATETIME_PRINT_FORMAT),
+            datetime: lambda v: v.strftime(DATETIME_PRINT_FORMAT)
+            if dtype_string(v) in ["timedelta64", "datetime64", "datetime", "Timestamp"]
```
xmatthias: Did you actually look at which field/row is "NaT"?  (maybe you can provide a simple strategy that causes this?)

this is a pretty broad approach - and i'm not sure what you need to do to a dataframe to get a "NaT" entry in your date column (it almost certainly means your strategy is messing with the dataframe - like adding lines - which means the strategy will probably raise errors in live.

Bloodhunter4rc:
this happens also on live/dry for me but only on kucoin (docker-a733a74d), same strategy running on binance usdt + busd + ftx is not affected. someone on discord (https://discord.com/channels/700048804539400213/702584313354518659/938940692808491039) mentioned its related to pandas? 

> Traceback (most recent call last):
  File "/home/ftuser/.local/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 377, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/uvicorn/middleware/message_logger.py", line 82, in __call__
    raise exc from None
  File "/home/ftuser/.local/lib/python3.9/site-packages/uvicorn/middleware/message_logger.py", line 78, in __call__
    await self.app(scope, inner_receive, inner_send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/applications.py", line 212, in __call__
    await super().__call__(scope, receive, send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/middleware/cors.py", line 92, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/middleware/cors.py", line 147, in simple_response
    await self.app(scope, receive, send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/routing.py", line 656, in __call__
    await route.handle(scope, receive, send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/routing.py", line 259, in handle
    await self.app(scope, receive, send)
  File "/home/ftuser/.local/lib/python3.9/site-packages/starlette/routing.py", line 61, in app
    response = await func(request)
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/routing.py", line 234, in app
    response_data = await serialize_response(
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/routing.py", line 138, in serialize_response
    return jsonable_encoder(
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/encoders.py", line 66, in jsonable_encoder
    return jsonable_encoder(
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/encoders.py", line 101, in jsonable_encoder
    encoded_value = jsonable_encoder(
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/encoders.py", line 115, in jsonable_encoder
    jsonable_encoder(
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/encoders.py", line 115, in jsonable_encoder
    jsonable_encoder(
  File "/home/ftuser/.local/lib/python3.9/site-packages/fastapi/encoders.py", line 47, in jsonable_encoder
    return encoder_instance(obj)
  File "/freqtrade/freqtrade/rpc/api_server/api_schemas.py", line 358, in <lambda>
    datetime: lambda v: v.strftime(DATETIME_PRINT_FORMAT),
  File "pandas/_libs/tslibs/nattype.pyx", line 75, in pandas._libs.tslibs.nattype._make_error_func.f
ValueError: NaTType does not support strftime

xmatthias:
@Bloodhunter4rc please provide us with a more complete log (basically - a few lines before the exception and the complete exception stack).

it's possibly related to pandas - but i'm yet to encounter this - so if you know how to reliably (!) reproduce this that'd be great (maybe consider opening it as issue instead ...)

Bloodhunter4rc:
there is just

> 2022-02-05 09:50:44 - freqtrade.worker - INFO - Bot heartbeat. PID=1, version='docker-a733a74d', state='RUNNING'
> 2022-02-05 09:51:51 - freqtrade.worker - INFO - Bot heartbeat. PID=1, version='docker-a733a74d', state='RUNNING'
> 2022-02-05 09:52:20 - uvicorn.error - ERROR - Exception in ASGI application

before, thats the complete exception from the log, if i can narrow it down i will, seems hard though
i can confirm that reverting to pandas-1.3.5 i dont get the exception, and its displayed

TheJoeSchr:
Are you just looking at the charts or do you have a strategy "loaded"?


Does this happen with `freqtrade webserver` or `freqtrade trade --dry-run`?

__

Bloodhunter4rc:
freqtrade trade , on live mode , just happens once i click a pair on the chart section in frequi on every pair

xmatthias:
you can add `        "verbosity": "info",` to the api_server configuration - that should list the endpoint iirc.

if i can get my hand on a (simple) strategy that reliably (!) reproduces this i'm more than happy to have a look

Bloodhunter4rc:
info did not show more output.

i narrowed mine down to informative pairs (if i remove all informative pairs its not happening), you can use this simple example strategy i posted in a other issue it happens on this one also.

Its only happening on kucoin for me, so not sure why , i disabled the proxy there just to make sure its not the proxy but it happens still.


```
import numpy as np  # noqa
import pandas as pd  # noqa
from pandas import DataFrame

from freqtrade.strategy import IStrategy, merge_informative_pair, informative

import talib.abstract as ta

class example(IStrategy):
    INTERFACE_VERSION = 2

    # Minimal ROI designed for the strategy.
    # This attribute will be overridden if the config file contains "minimal_roi".
    minimal_roi = {
        #"4320": 0.01,
        #"300": 0.01,
        #"60": 0.04, # 0.01
        #"30": 0.06, # 0.02
        "0": 0.50 # 0.04
    }

    # Optimal stoploss designed for the strategy.
    # This attribute will be overridden if the config file contains "stoploss".
    stoploss = -0.06 # -0.05

    # Trailing stoploss
    trailing_stop = True
    # trailing_stop_positive = 0.005
    trailing_stop_positive = 0.0005
    trailing_stop_positive_offset = 0.015
    #trailing_stop_positive_offset = 0.015

    # Optimal timeframe for the strategy.
    timeframe = '5m'

    # Run "populate_indicators()" only for new candle.
    process_only_new_candles = False

    # These values can be overridden in the "ask_strategy" section in the config.
    use_sell_signal = False
    sell_profit_only = True
    ignore_roi_if_buy_signal = True

    startup_candle_count: int = 200

    # Optional order type mapping.
    order_types = {
        'buy': 'limit',
        'sell': 'limit',
        'stoploss': 'market',
        'stoploss_on_exchange': False
    }

    order_time_in_force = {
        'buy': 'gtc',
        'sell': 'gtc'
    }

    plot_config = {
        # Main plot indicators (Moving averages, ...)
        'main_plot': {
            'tema': {},
            'sar': {'color': 'white'},
        },
        'subplots': {
            # Subplots - each dict defines one additional plot
            "MACD": {
                'macd': {'color': 'blue'},
                'macdsignal': {'color': 'orange'},
            },
            "RSI": {
                'rsi': {'color': 'red'},
            }
        }
    }

    @informative('1d')
    def populate_indicators_1d(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        dataframe['rsi'] = ta.RSI(dataframe, timeperiod=14)

        return dataframe

        """
        Adds several different TA indicators to the given DataFrame

        Performance Note: For the best performance be frugal on the number of indicators
        you are using. Let uncomment only the indicator you are using in your strategies
        or your hyperopt configuration, otherwise you will waste your memory and CPU usage.
        :param dataframe: Dataframe with data from the exchange
        :param metadata: Additional information, like the currently traded pair
        :return: a Dataframe with all mandatory indicators for the strategies
        """

        # Momentum Indicators
        # ------------------------------------

        return dataframe

        """
        Based on TA indicators, populates the buy signal for the given dataframe
        :param dataframe: DataFrame populated with indicators
        :param metadata: Additional information, like the currently traded pair
        :return: DataFrame with buy column
        """
        dataframe.loc[ # Breakout: TEMA crosses over longer TEMA
            (
                (dataframe['rsi'] > 80) & (dataframe['rsi_1d'] > 80)
            ),
            'buy'] = 1
 
        return dataframe

        """
        Based on TA indicators, populates the sell signal for the given dataframe
        :param dataframe: DataFrame populated with indicators
        :param metadata: Additional information, like the currently traded pair
        :return: DataFrame with buy column
        """
        dataframe.loc[
            (
                (dataframe['rsi'] < 1) & (dataframe['rsi_1d'] < 1)
            ),
            'sell'] = 1
        return dataframe

```

freqtrade/rpc/api_server/api_schemas.py
xmatthias: I think the correct place to fix this would be in `rpc.py` - where we also replace `np.NaN` with None.

That said - it appears like currently it's not possible to convert pd.NaT to "anything" - so i've opened an issue against pandas (https://github.com/pandas-dev/pandas/issues/45836).
I'd want to wait on the outcome of that issue before deciding if/how we implement a fix.

TheJoeSchr: I think that is for sure the sensible approach, but so far it's not even sure *if* they will fix that (glimpsing at your discussions of it there)

And for now this part would still be broken, even though we have a "fix" and another one (fixing it in `rpc.py`) lined up already.

How about we implement want of this band-aid fixes for now, reference in the comments this PR and also write a unit test with the strategy @Bloodhunter4rc provided as a litmus test?

Once pandas fixes it proper, we can just switch over do it then

xmatthias: so far, we've barely given pandas time to actually respond (the issue is barely 2 days old).

if any, i'd use the approach in `rpc.py`. Updating the formatter (as you're proposing here) will only fix one particular issue - but may still fail should a column contain NaN / inf.

I don't think we'll need the strategy for that though - as we can construct a fake dataframe that proves the "failure" without issues.

TheJoeSchr: :+1: 

okay, I will look into the `rpc.py` thing, worst case I learn something.

even better if the strategy is not needed, just wanted to make sure that there is a test in place to verify the proper fix once the band-aided would needed to be ripped of.

what are the properties of the fake dataframe to force the "failure"? just any datetime colume containing NaT/NaN/inf?

xmatthias: a datetime-column with `pd.NAT`. (maybe also with some valid dates).

Also, maybe a numeric column which includes inf/-inf or `pd.np.NaN` ... as i think the same bug is there as well (that's why i say - the current fix doesn't fix all instances of the problem).

patosG:
The bug occurs with the BB_RPL_TSL strategy : https://github.com/jilv220/BB_RPB_TSL
which uses informatives pairs as well
hope it helps

freqtrade/rpc/api_server/api_schemas.py
TheJoeSchr: refactored (byusing `isnull()`) and rebased it

I'm not sure if it's still needed, but I guess it doesn't hurt if json_encoder is a little bit more robust?

TheJoeSchr:
Added a band-aid fix for now and left a comment what to try once the pandas issue is fixed

TheJoeSchr:
> a datetime-column with pd.NAT. (maybe also with some valid dates).
Also, maybe a numeric column which includes inf/-inf or pd.np.NaN ... as i think the same bug is there as well (that's why i say - the current fix doesn't fix all instances of the problem).

we could us the same logic for `inf/-inf` or `NaN` properly, by replacing them with an empty string as well, so it's json compatible. But I'm not sure if this is useful, because so far there don't seem to be any issues with those?

freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -355,7 +356,9 @@ class PairHistory(BaseModel):
 
     class Config:
         json_encoders = {
-            datetime: lambda v: v.strftime(DATETIME_PRINT_FORMAT),
+            datetime: lambda v: v.strftime(DATETIME_PRINT_FORMAT)
+            # needed for aslong NaT doesn't work with strftime
+            if not isnull(v) else "",
```
xmatthias: you should take a pick - EITHER we fix it here (which i think is wrong) - or we fix it in rpc (which is also currently wrong in my opinion).

freqtrade/rpc/rpc.py
```diff
@@ -963,6 +963,24 @@ def _convert_dataframe_to_dict(strategy: str, pair: str, timeframe: str, datafra
...
+                # because if replaced with `None`
+                # it will be casted into NaT again
+                dataframe[date_column] = dataframe[date_column].apply(
+                    lambda x: '' if isnull(x) else x)
+
+            """
+            try this if above pandas Issue#45836 is fixed:
+            https://github.com/pandas-dev/pandas/issues/45836
+            """
+            # dataframe = dataframe.replace({NaT: None})
```
xmatthias: I don't think this is a correct fix (or rather, a complete fix).

If you look at both my issue in pandas, as well as the one that's linked there, you'll notice that the same exact issue is also present for `NAN`.

The proper fix (fixing the issue on both date and numeric columns) is posted in the 2nd issue (https://github.com/pandas-dev/pandas/issues/45601#issuecomment-1028480734) - and would result in the below line to be changed to `<...>.astype(object).replace(...`.

currently, this is only fixing the issue for dates - but if a numeric column contains `inf` - pandas would first replace that wit NaN (This still works) - but would then fail to convert NAN to None (None is not a valid numeric numpy type).
Therefore the correct approach (as pointed out in the linked comment) is to first convert the column to object - and then eventually values to `None` (which is not a valid pandas numeric type).
I don't like it as it's broad, and will change the type of more columns than necessary - but the impact is limited as we're working on a copy of the dataframe anyway.

TheJoeSchr: I agree that this is a partial fix, I mentioned it in a comment further down. But you know the saying, perfect is the enemy of good and I learned a little bit more about ft webserver side.

But I don't mind if you don't want to merge this in yet, I just thought I'd share some temporary solutions if somebody needs it.

Talking about this:
> first convert the column to object - and then eventually values to None

Do you want me to implement this or should we wait for a better fix of pandas team?

xmatthias: My point here is - if we implement a temporary fix - i'd want it to be complete - fixing both issues (even though no-one ran into the 2nd one yet - that's only a matter of time as the pandas-team confirmed that this regression is present.

freqtrade/rpc/api_server/api_schemas.py
TheJoeSchr: Same as the in the comment down below. I don't mind if it's not shared, just thought I would share it. 

Maybe it helps somebody or if this takes longer to fix at pandas side, we can rethink which way to pick.

I just wanted to try my hands at both sides of the issue.

freqtrade/rpc/rpc.py
TheJoeSchr: Agree.

Tell me when or if I should do start working on this

xmatthias: seems like the fix works - but i've simplified it a bit anyway and added a test :+1:

xmatthias (approved):
LGTM :+1: 

as mentioned in the other comment - i've modified the fix a bit.
also, when adding comments - don't use multiline strings - these are evaluated at runtime (python sees and runs them), but instead use real comments (lines prefixed with #).
While the impact is negligible in most cases (it's just memory for one more string) - it can sum up quickly if we ain't careful.

TheJoeSchr:
> LGTM :+1: 
> 
> as mentioned in the other comment - i've modified the fix a bit.
> also, when adding comments - don't use multiline strings - these are evaluated at runtime (python sees and runs them), but instead use real comments (lines prefixed with #).
> While the impact is negligible in most cases (it's just memory for one more string) - it can sum up quickly if we ain't careful.
> 

Oh, I thought we were waiting for pandas, must have misunderstood. Anyways good for people who want to have it working now!

Regarding Multiline comments: gotcha! I just did it so it stands out like a sore thumb, because it should be temporary


Cheers

TheJoeSchr:
The comment on line #973 is now wrong

> Replace NaT with empty string

I think it can be deleted all together, it was just meant to help the reviewer and what it does is clear from the code.

I would still leave the "why" we do this part of the comment though

I think it's easier if you do this instead of me opening another PR?

PR #6341, opened 2022-02-01
Plotting: add alias `--backtest-filename` for `--export-filename`

TheJoeSchr:
## Summary
makes it easier to discover how to use this argument

This is just an initial approach. If it seems to be good, I can also port it over to plot-dataframe etc.

If there is a better way to do this, please let me know...

### DIFF
--- a/freqtrade/commands/cli_options.py
+++ b/freqtrade/commands/cli_options.py
@@ -182,11 +182,17 @@ AVAILABLE_CLI_OPTIONS = {
 
     ),
     "exportfilename": Arg(
-        '--export-filename',
-        help='Save backtest results to the file with this filename. '
-        'Requires `--export` to be set as well. '
-        'Example: `--export-filename=user_data/backtest_results/backtest_today.json`',
-        metavar='PATH',
+        "--export-filename",
+        help="Save backtest results to the file with this filename. "
+        "Requires `--export` to be set as well. "
+        "Example: `--export-filename=user_data/backtest_results/backtest_today.json`",
+        metavar="PATH",
+    ),
+    "backtestfilename": Arg(
+        "--backtest-filename",
+        help="Use backtest results from this filename."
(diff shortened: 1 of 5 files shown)


freqtrade/commands/cli_options.py
```diff
@@ -182,11 +182,17 @@ def __init__(self, *args, **kwargs):
...
-        'Example: `--export-filename=user_data/backtest_results/backtest_today.json`',
-        metavar='PATH',
+        "--export-filename",
+        help="Save backtest results to the file with this filename. "
+        "Requires `--export` to be set as well. "
+        "Example: `--export-filename=user_data/backtest_results/backtest_today.json`",
+        metavar="PATH",
+    ),
+    "backtestfilename": Arg(
+        "--backtest-filename",
```
xmatthias: I'd clearly not use 2 different arguments / configuration keys for the same thing.
That's pretty dangerous, and is bound to fail assuming one comes from the configuration (then you got both variables available - and will wonder why the wrong file is loaded).

you can assign a 2nd argument to the same key eventually, and maybe change the help text to something like `Filename for backtesting results" - so it works for both loading and saving.

xmatthias (changes requested):

freqtrade/commands/cli_options.py
TheJoeSchr: you mean like this? I haven't tested commit 298815d yet, I wanted to see if I understood you right

xmatthias: Yes exactly

TheJoeSchr:
I think this is ready now?

I did this:

- rebased on current `develop`
- fixed flake8 linter issue
- tested `--backtest-filename` w/ `plot-profit`
- tested `--export-filename` w/ `plot-dataframe`
- tested `--backtest-filename` w/ `plot-profit`
- tested `--export-filenames` w/ `plot-dataframe`

xmatthias (approved):
For sure 👍

PR #6343, opened 2022-02-01
added support for waves.exchange

peterwilli:

## Summary

This change adds Waves.exchange support to Freqtrade.

## Quick changelog

- Add support for Waves.exchange by creating a custom class.

## What's new?

Support for Waves.exchange.  Note that in order for this to work, the latest CCXT is needed, i sent a bunch of changes to CCXT to fix support for Waves including the ability for market orders.

The changes are merged, so when Freqtrade bumps it's CCXT version it should be ok:
Merges to CCXT for Waves.exchange support

Tested both testnet/mainet + backtest, dry run and even live trading.

### DIFF
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -34,6 +34,7 @@ BAD_EXCHANGES = {
 MAP_EXCHANGE_CHILDCLASS = {
     'binanceus': 'binance',
     'binanceje': 'binance',
+    'wavesexchange': 'waves',
 }
 
 
--- a/freqtrade/exchange/waves.py
+++ b/freqtrade/exchange/waves.py
@@ -1,0 +1,28 @@
+""" Waves exchange subclass """
+import logging
+from typing import Dict, Optional
+
+from freqtrade.exchange import Exchange
+
+
+logger = logging.getLogger(__name__)
+
+
+class Waves(Exchange):
+    """
+    Waves exchange class. Contains adjustments needed for Freqtrade to work
+    with this exchange.
+
+    Please note that this exchange is not included in the list of exchanges
+    officially supported by the Freqtrade development team. So some features
+    may still not work as expected.
+    """
+
+    _ft_has: Dict = {
+        "ohlcv_candle_limit": 1440,
+    }
+
+    # There seems to be no minumum?
+    def get_min_pair_stake_amount(self, pair: str, price: float,
+                                  stoploss: float) -> Optional[float]:
+        return 0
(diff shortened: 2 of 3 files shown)


freqtrade/exchange/waves.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+
+    _ft_has: Dict = {
+        "ohlcv_candle_limit": 1440,
+    }
+
+    # There seems to be no minumum?
+    def get_min_pair_stake_amount(self, pair: str, price: float,
+                                  stoploss: float) -> Optional[float]:
+        return 0
```
samgermain, line 28: I don't think this method will change any behaviour because nothing already happens when the minimum is None, so you should probably get rid of it, as well as `Optional`

freqtrade/exchange/common.py
```diff
@@ -34,6 +34,7 @@ def _get_logging_mixin():
 MAP_EXCHANGE_CHILDCLASS = {
     'binanceus': 'binance',
     'binanceje': 'binance',
+    'wavesexchange': 'waves',
```
xmatthias, line 37: This mapping is pointless.

it's only there for binance since binance has 3 different ccxt exchange-names which map to the same freqtrade exchange class.
unless waves has 2 different ccxt names (hint: it doesn't) - please use ccxt naming to avoid unnecessary confusion.

peterwilli: Thanks for clarifying

xmatthias: @peterwilli are you planning on finishing this PR (make tests, pass, address review comments)?

Otherwise i'll have to close this, as i do not have an account on waves.exchange - and currently do not plan on testing/adding support for this exchange myself.

peterwilli: I didn't really plan to work more on this so this might get some delays, however the download bug I'll push to CCXT today.

xmatthias: > I didn't really plan to work more on this so this might get some delays

I'm not sure how to interpret that - i'm fine with leaving this open _if you plan on making it mergable_ in the foreseeable future - but there's no point if it's a "dump and run" PR - where the work would now fall onto maintainers to finish the PR - simply because i see no incentive to include an exchange with this little volume.

It's not in a merge-able state at the moment (due to the wrong naming - as well as merge conflicts) and the other comment that was pointed out.

As it then doesn't provide any more benefit other than "knows the precise amount of candles" - i'm not sure if this PR makes much sense for an exchange this small, as it'll still be usable with freqtrade as it's an exchange supported by ccxt, just without knowing the precise number of candles.

peterwilli: Alright well you can close it. I'm sorry for the confusion. It wasn't meant as a dump & run PR, I just lost interest due to some reasons I rather not talk about.

In any case, I'll fix the download part on the CCXT side.

I'll remake a new PR when things change.

xmatthias: That's perfectly fine, life/circumstances change.

PR #6344, opened 2022-02-01
Don't allow data from the future to be stored (waves.exchange)

peterwilli:
## Summary

In this change we cut off all data past the current time when `download-data` is executed.

## Quick changelog

- Filter data past current time

## What's new?

When running `download-data` with Waves.exchange, the exchange APi returns data past the current time at the end of the candles list, having all values set to `null`, causing Freqtrade to interpolate over it and keep the same value over and over.

Example:
```
2022-02-01 16:16:48,748 - freqtrade.strategy.interface - DEBUG - Populating indicators for pair WX/USDN.
                          date      open      high       low     close  volume  SHB          macd
0    2022-01-23 16:20:00+00:00       NaN       NaN       NaN       NaN     0.0  0.0           NaN
1    2022-01-23 16:25:00+00:00       NaN       NaN       NaN       NaN     0.0  0.0           NaN
2    2022-01-23 16:30:00+00:00       NaN       NaN       NaN       NaN     0.0  0.0           NaN
3    2022-01-23 16:35:00+00:00       NaN       NaN       NaN       NaN     0.0  0.0           NaN
4    2022-01-23 16:40:00+00:00       NaN       NaN       NaN       NaN     0.0  0.0           NaN
...                        ...       ...       ...       ...       ...     ...  ...           ...
2665 2022-02-01 22:25:00+00:00  1.203377  1.203377  1.203377  1.203377     0.0  0.0  6.661338e-16
2666 2022-02-01 22:30:00+00:00  1.203377  1.203377  1.203377  1.203377     0.0  0.0  6.661338e-16
2667 2022-02-01 22:35:00+00:00  1.203377  1.203377  1.203377  1.203377     0.0  0.0  6.661338e-16
2668 2022-02-01 22:40:00+00:00  1.203377  1.203377  1.203377  1.203377     0.0  0.0  6.661338e-16
2669 2022-02-01 22:45:00+00:00  1.203377  1.203377  1.203377  1.203377     0.0  0.0  6.661338e-16
```

Note that at the time, it wasn't `2022-02-01 22:45:00` yet. I now fixed this by using Python's native `filter` function.
There's no breaking changes noticed, except if you have your clock set wrong (but if you do that, you can't really access SSL protected sites to begin with, which is basically every exchange, so that would have broken something already).

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1290,6 +1290,10 @@ class Exchange:
                     p, _, new_data = res
                     if p == pair:
                         data.extend(new_data)
+        # Some exchanges show data past the current time and set all values to null.
+        current_time = arrow.utcnow().int_timestamp * 1000
+        data = filter(lambda item: item[0] < current_time, data)
+        
         # Sort data again after extending the result - above calls return in "async order"
         data = sorted(data, key=lambda x: x[0])
         return pair, timeframe, data


freqtrade/exchange/exchange.py
```diff
@@ -1290,6 +1290,10 @@ async def _async_get_historic_ohlcv(self, pair: str, timeframe: str,
                     p, _, new_data = res
                     if p == pair:
                         data.extend(new_data)
+        # Some exchanges show data past the current time and set all values to null.
+        current_time = arrow.utcnow().int_timestamp * 1000
+        data = filter(lambda item: item[0] < current_time, data)
```
xmatthias, line 1295: TBH - i'm not certain this is something that should be fixed at freqtrade level.

Please open this as issue or PR against ccxt. This would be the only exchange this has been observed - so i think it makes more sense to fix this at ccxt - so other projects won't run into the same problem.

I'm happy to merge this only if ccxt explicitly says they're refusing to have this in their code.

peterwilli: I'll do that. I was debating whether or not this is a ccxt issue. I'll try it there today!

xmatthias: that's not always clear - i always try to think "will this be a problem for other projects using ccxt, too" - and if the answer is yes - i try to fix it there.

As you'll notice above - i'm not per se excluding this from being added - just think it'll make more sense in ccxt.

xmatthias: @peterwilli any update on this?

xmatthias:
I'll close this for the moment.
Feel free to resubmit should modification change be rejected by ccxt (please include the link to that issue/PR).

freqtrade/exchange/exchange.py
peterwilli: > @peterwilli any update on this?

Hey xmatthias I'm sorry I got distracted with another crypto coin. I'll resubmit a PR today to CCXT about the downloading part.

peterwilli:
CCXT-equivalent PR is sent to here: https://github.com/ccxt/ccxt/pull/12215

Issue #6224, opened 2022-01-14
Feat/short issues for binance futures


fridary:
* Operating system: Mac os 11.6.1
  * Python Version: 3.8.12
  * CCXT version: 1.67.31
  * Freqtrade Version: 2021.12
  
@samgermain
Well, I have just downloaded `feat/short` branch and wonder how you use that as there are many fixes need to make. On what exchange do you trade?
I will start sequentially after the launch of the fresh project, user_dir and config: `python -m freqtrade trade -c config.json`. Config added:
```
"trading_mode": "futures",
"collateral": "isolated",
```

1) `freqtrade.exceptions.OperationalException: Freqtrade does not support isolated futures on Binance`
I suppose we can uncomment lines in `exchange/binance.py`:
```
_supported_trading_mode_collateral_pairs: List[Tuple[TradingMode, Collateral]] = [
        # TradingMode.SPOT always supported and not required in this list
        # TODO-lev: Uncomment once supported
        # (TradingMode.MARGIN, Collateral.CROSS),
        (TradingMode.FUTURES, Collateral.CROSS),
        (TradingMode.FUTURES, Collateral.ISOLATED)
    ]
```
2) `freqtrade.freqtradebot - WARNING - Unable to create trade for ETH/USDT: Could not set margin mode due to BadRequest. Message: binance marginType must be either isolated or crossed`
In `exchange.py set_margin_mode()` change `self._api.set_margin_mode(pair, collateral.value, params)` to `self._api.set_margin_mode(collateral.value, pair, params)`

3)
```
2022-01-13 17:11:35 DEBUG urllib3.connectionpool._make_request() https://fapi.binance.com:443 "POST /fapi/v1/marginType HTTP/1.1" 400 53
2022-01-13 17:11:35 WARNING freqtrade.exchange.common.wrapper() set_margin_mode() returned exception: "Could not set margin mode due to AuthenticationError. Message: binance {"code":-4046,"msg":"No need to change margin type."}". Giving up.
2022-01-13 17:11:35 WARNING freqtrade.freqtradebot.enter_positions() Unable to create trade for ETH/USDT: Could not set margin mode due to AuthenticationError. Message: binance {"code":-4046,"msg":"No need to change margin type."}
```
in `exchange.py set_margin_mode()` quick fix replace
```
except (ccxt.NetworkError, ccxt.ExchangeError) as e:
            raise TemporaryError(
[...]

PR #6346, opened 2022-02-02
Minor fixes for feat/short

samgermain:
Solves issues 3, 4, 5 and 6 of https://github.com/freqtrade/freqtrade/issues/6224

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1880,7 +1880,7 @@ class Exchange:
             return
 
         try:
-            self._api.set_margin_mode(pair, margin_mode.value, params)
+            self._api.set_margin_mode(margin_mode.value, pair, params)
         except ccxt.DDoSProtection as e:
             raise DDosProtection(e) from e
         except (ccxt.NetworkError, ccxt.ExchangeError) as e:
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1786,7 +1786,7 @@ class Exchange:
 
         try:
             funding_history = self._api.fetch_funding_history(
-                pair=pair,
+                symbol=pair,
                 since=since
             )
             return sum(fee['amount'] for fee in funding_history)
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -877,7 +877,12 @@ class Exchange:
 
     # Order handling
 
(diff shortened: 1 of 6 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -1834,7 +1853,7 @@ def _set_leverage(
             return
 
         try:
-            self._api.set_leverage(symbol=pair, leverage=leverage)
+            self._api.set_leverage(symbol=pair, leverage=round(leverage))
```
xmatthias: are you sure set_leverage must be a full number? (to be precise - are you sure that's the case for all exchanges?).

freqtrade/exchange/exchange.py
```diff
@@ -882,21 +882,38 @@ def _lev_prep(self, pair: str, leverage: float):
...
+        leverage: float,
+        reduceOnly: bool,
+        time_in_force: str = 'gtc',
+    ) -> Dict:
         params = self._params.copy()
         if time_in_force != 'gtc' and ordertype != 'market':
             param = self._ft_has.get('time_in_force_parameter', '')
             params.update({param: time_in_force})
+        if reduceOnly:
+            params.update({'reduceOnly': True})
```
xmatthias, line 902: is "reduceOnly" an official ccxt parameter which is supported for all exchanges, and eventually mapped to whatever naming other exchanges use?

samgermain: Yeah, it's supposed to be anyway, if there's an exchange where that's not the case it should be updated

xmatthias: while not relevant for this - i think this is a point that's somewhat lacking in ccxt - you say it's supposed to be unified - but the documentation does not even contain the word `reduceOnly`.

That makes it difficult to rely on it (i do trust it's true - but it's a pointless question to begin with as it should simply be stated in the docs) ...  it means you've just forced users to open an issue asking this (which seems clear for ccxt team - but obviously not for bystanders / users).

this is not the only case of this mind you ... but something i think should be improved ...

samgermain: @xmatthias So I just checked, and it's not that common across the library yet, a number of exchanges just take the `reduceOnly` parameter (including binance, ftx and okex), but `gateio` is the only one I can see that takes `reduceOnly` and changes it to the exchange specific, `reduce_only` so that would be why it's not documented.

It is the intent though

PR #6349, opened 2022-02-02
Todos

samgermain:


freqtrade/freqtradebot.py
```diff
@@ -1272,7 +1272,8 @@ def handle_cancel_enter(self, trade: Trade, order: Dict, reason: str) -> bool:
             # to the order dict acquired before cancelling.
             # we need to fall back to the values from order if corder does not contain these keys.
             trade.amount = filled_amount
-            # TODO-lev: Check edge cases, we don't want to make leverage > 1.0 if we don't have to
+            # * Check edge cases, we don't want to make leverage > 1.0 if we don't have to
+            # * (for leverage modes which aren't isolated futures)
```
samgermain, line 1276: In isolated futures, the leverage is always the amount of collateral you put in, if you add more collateral, it goes in with the same amount of leverage, so your leverage doesn't change depending on if an order is filled all the way or not

freqtrade/optimize/backtesting.py
```diff
@@ -497,7 +501,7 @@ def _get_sell_trade_entry(self, trade: LocalTrade, sell_row: Tuple) -> Optional[
         sell_candle_time: datetime = sell_row[DATE_IDX].to_pydatetime()
 
         if self.trading_mode == TradingMode.FUTURES:
-            # TODO-lev: Other fees / liquidation price?
+            # TODO-lev: liquidation price?
```
samgermain, line 504: Futures has funding_fees + regular trading fees that you would see in spot, unless there's a liquidation. Are we going to set our liq price to 5% above the actual liq price?

freqtrade/persistence/models.py
```diff
@@ -569,7 +569,7 @@ def update(self, order: Dict) -> None:
                 payment = "BUY" if self.is_short else "SELL"
                 # * On margin shorts, you buy a little bit more than the amount (amount + interest)
                 logger.info(f'{order_type.upper()}_{payment} has been fulfilled for {self}.')
-            # TODO-lev: Double check this
+            # TODO-lev: Is anything else needed here?
```
samgermain: @xmatthias I'm asking you this, do we need to add anything here? I'm not sure why I put it here, but I think I wasn't sure if it was complete

samgermain:
@xmatthias Most of the other TODOs I don't really know what to do about, or don't really understand the part of the code where they are, so do you think you could take a look through the remaining ones and solve which ones are obvious to you?

- And don't worry about the longer ones like "Create a meaningful short strategy"

Thanks

freqtrade/freqtradebot.py
```diff
@@ -1339,13 +1340,13 @@ def _safe_exit_amount(self, pair: str, amount: float) -> float:
...
         """
-        # TODO-lev Maybe update?
         # Update wallets to ensure amounts tied up in a stoploss is now free!
         self.wallets.update()
         trade_base_currency = self.exchange.get_pair_base_currency(pair)
         wallet_amount = self.wallets.get_free(trade_base_currency)
         logger.debug(f"{pair} - Wallet: {wallet_amount} - Trade-amount: {amount}")
-        if wallet_amount >= amount:
+        if wallet_amount >= amount or self.trading_mode is TradingMode.FUTURES:
+            # A safe exit amount isn't needed for futures, you can just exit/close the position
```
xmatthias, line 1350: I think this introduces a bug, while hiding bigger problems.

First of all, the balance (`/balance` via telegram) will be reported wrong (this is failing ONLY because the balance is not correct - otherwise it'd be working as is) - and will be missing the position that is open, most likely understating overall balance (so if you've got 100% of your capital in positions, /balance will report 0 balance).

It'll also make freqtrade fail/crash/error if the position was closed on-exchange by the user (this is a scenario that's quite likely - and we need to account for at least by not failing - not necessarily by closing the position in freqtrade correctly).

freqtrade/optimize/backtesting.py
```diff
@@ -333,8 +333,12 @@ def _get_ohlcv_as_lists(self, processed: Dict[str, DataFrame]) -> Dict[str, Tupl
                     df_analyzed.loc[:, col] = 0 if not tag_col else None
 
             # Update dataprovider cache
-            self.dataprovider._set_cached_df(pair, self.timeframe, df_analyzed, CandleType.SPOT)
-            # TODO-lev: Candle-type should be conditional, either "spot" or futures
+            self.dataprovider._set_cached_df(
+                pair,
+                self.timeframe,
+                df_analyzed,
+                CandleType.FUTURES if self.trading_mode == TradingMode.FUTURES else CandleType.SPOT
```
xmatthias: i'd use the variable from the config instead of this (`candle_type_def`) - which will have the correct candle type without having this logic all over the place.

freqtrade/freqtradebot.py
samgermain: What should change?

xmatthias: in my understanding, you introduced this line to avoid having exiting fail because "balance is too low" (or whatever the error is called).
That might even be fine (i'm not sure it is - but i guess we can leave it as workaround for now).

To me, the proper fix is however to actually have balance show up "somehow" in our wallet (it's a future we own - but we still own it - and currently it'll look like you're broke if you're in a position).

i'm not sure yet how to approach this in detail - but `fetch_balance()` will not return the amount of the positions (at least not for all exchanges).
We might therefore need to have an alternative way to get the proper balance (i'd however do that completely separate, not in this PR).

an example of the problem:
i have a wallet of 1000$
i enter a long position with 2x leverage and go "all in" - so now i have a position of 2000$ in ETH.
if i now run `/balance` - i'll have "0.0" as my balance - since the balance is not available - and positions don't show up. Instead, i should have 2000 ETH (maybe with a hint that it's a 2x position) ... 


Having this would automatically/implicitly fix this problem - and is needed for correct "/balance" output, too.

samgermain: Where's the part that needs to change? We could just call `fetchPositions` and sum (contracts*contractSize*price), we would have to call `fetchTickers` also if we want the real price instead of the markPrice

xmatthias: i don't know (as said above, i'm not sure how to approach this yet).
it'll be either wallet updating - or in the exchange (some addition to the get_balances method) ... which is really just syntactic suggar (where you put it).

i think it'll depend how we want/need it - as i'm not sure we can/should add position amounts directly to "ETH free" - as that'd be technically wrong (we do have an ETH position - but we can't use that ETH for anything else ...).
 
Maybe for this PR - add it as a TODO to the above line (in my understanding that line should be done differently once balance reporting is corrected) - and add it as a todo task in the dashboard.

freqtrade/exchange/binance.py
```diff
@@ -100,6 +100,7 @@ def stoploss(self, pair: str, amount: float, stop_price: float,
                 side=side,
                 amount=amount,
                 price=rate,
+                leverage=leverage,
```
xmatthias: This is wrong - this `create_order` call is the ccxt one - ccxt does not have a "leverage" parameter on `createOrder`.

i'm also not sure we'll really need to set leverage on a stop order - as we're just "closing a position" - so leverage is really already defined.

freqtrade/persistence/models.py
xmatthias: I would think this has to do with the comment 2 lines above here - but i don't think there's much we can do exactly here - other than obviously verifying that we paid back everything we loaned ... but that won't happen in this file for sure.

freqtrade/optimize/backtesting.py
xmatthias: I'm not certain i understand this (and if you mean this only for backtesting, or for live mode too).

i mean  - would we assume that liquidation fees are 5%, and therefore have backtesting "pay that as a fee" - or would we force the stop 5% above liquidation price to ensure we're not running into a real liquidation? (again - only for backtesting, or also for live??)

freqtrade/plot/plotting.py
```diff
@@ -431,44 +460,15 @@ def generate_candlestick_graph(pair: str, data: pd.DataFrame, trades: pd.DataFra
...
-                    size=9,
-                    line=dict(width=1),
-                    color='red',
-                )
-            )
-            fig.add_trace(sells, 1, 1)
-        else:
-            logger.warning("No sell-signals found.")
+    longs = create_scatter(data, 'enter_long', 'green', 'up')
+    exit_longs = create_scatter(data, 'exit_long', 'red', 'down')
```
xmatthias: To be honest, i'd split the changes to the plotting modules out of here to a separate PR (if you're keen on completing them - for me, i could live with not having futures support for plotting for the moment).

these changes suggest it's working - but it really isn't, and produced results are wrong.

freqtrade/freqtradebot.py
richardjozsa: @xmatthias Your idea could work, I went after the endpoint what we need, if we can use `fetchPositions` , that should give back the `'conracts` which is the number of asset what we bought. Proly if we put that number there, that should work. So in your example we will get information about how many ETH we bought for that 2000 USDT. in this way we don't need any problematic calculation. With this we can put it into our balance. BTW I just want to note that here that CCXT documentation is confusing in case of this table, they write "float" in the description, while their examples are decimals?!
image

freqtrade/optimize/backtesting.py
samgermain: > or would we force the stop 5% above liquidation price to ensure we're not running into a real liquidation? (again - only for backtesting, or also for live??)

Yes, for live, dry-run, backtesting, and maybe give the user some control over this parameter?

freqtrade/plot/plotting.py
samgermain: I reverted the changes, the commit is there if you want to use it, I've never even used plotting so I won't try to figure it out this way

sha: 1f74cfe841d1eede5a4670104e572eb91edf14b4
message: "plot.generate_candlestick_graph Added short equivelent, separating plotting scatter creation to a function"

**Reversion**
sha: 99b8a8ca79547ae03d31f83130e911cabb34c71b
message: "Revert "plot.generate_candlestick_graph Added short equivelent, separating plotting scatter creation to a function"

This reverts commit 0abba7f9b7299ba3c45df6a2ba6e35ad6a19c5a0." (I rebased so this doesn't match)

samgermain:
>  or would we force the stop 5% above liquidation price to ensure we're not running into a real liquidation?

Yes, for live, dry-run backtesting

freqtrade/persistence/models.py
samgermain: We only need to do that on margin trading anyway

freqtrade/exchange/binance.py
samgermain: Right, I just got it confused with the freqtrade `createOrder`

freqtrade/optimize/backtesting.py
xmatthias: yeah, something like "liquidation_offset" as parameter might work

xmatthias (approved):
I'll merge this as is - as the other open topics are both bigger changes - where i think it makes sense to do separate PR's.

freqtrade/plot/plotting.py
xmatthias: Thanks - i've extracted the "forward" commit to the branch `short_plot` (otherwise i'm 100% sure we'll lose it).
I only know it ain't working - i didn't look into the details yet as to why (i suspect it's loading the wrong data) - but didn't check further and i'm not sure if that's the only problem

PR #6354, opened 2022-02-04
wrote ccxt_compat.test_get_max_leverage_spot test_get_max_leverage_fu…

samgermain:


samgermain:
CCXT needs to be updated, the failed test is a bug I just fixed today
https://github.com/ccxt/ccxt/pull/11857

xmatthias:
> CCXT needs to be updated, the failed test is a bug I just fixed today [ccxt/ccxt#11857](https://github.com/ccxt/ccxt/pull/11857)

we can always force-bump ccxt (i've done that in this branch a few times already iirc) - i'd also increase the "min ccxt version" in `setup.py` in that case, as we're now incompatible with older ccxt versions (which is a fact i really REALLY don't care about :laughing: )

setup.py
```diff
@@ -42,7 +42,7 @@
     ],
     install_requires=[
         # from requirements.txt
-        'ccxt>=1.66.32',
+        'ccxt>=1.72.29',
```
xmatthias, line 45: Please also bump to this or a latter version in `requirements.txt`, otherwise ccxt will be installed twice - first the one from requirements, and then this one (not just in CI, but also in regular installs).

xmatthias (approved):
:+1:

PR #6358, opened 2022-02-05
Update setup.sh / correct minimum required python version

Bloodhunter4rc:
## Summary

Just came across this when i tried to run it with python3.7


xmatthias (approved):
Great catch! :+1: 

I've actually found a few more places where this was missing in the docs

PR #6373, opened 2022-02-07
Leverage tiers

samgermain:
Uses methods from these PRs

[Okx fetchLeverageTiers](https://github.com/ccxt/ccxt/pull/11938)
[Binance fetchLeverageTiers](https://github.com/ccxt/ccxt/pull/11937)
[fetchLeverageTiers documentation](https://github.com/ccxt/ccxt/pull/11941)

- Redid fill_leverage_brackets
- moved binance.get_maintenance_ratio_and_amt & binance.get_max_leverage to base exchange class
- changed ccxt method of getting the leverage brackets
- rename leverage brackets to leverage_tiers
- Okex can now get the maintenance amount


samgermain:
@xmatthias A few of the tests are failing for `exchange.get_max_leverage` because it looks like `self._api.has['fetchLeverageTiers']` always evaluates to true, but it needs to be false for `gateio`

samgermain:
@xmatthias 

With the Okex implementation of `get_max_pair_stake_amount`, `test_execute_entry` is now failing with `E           TypeError: '<' not supported between instances of 'float' and 'MagicMock'` even with `get_max_pair_stake_amount` in `mocker.patch.multiple`

freqtrade/exchange/exchange.py
```diff
@@ -690,13 +692,14 @@ def get_max_pair_stake_amount(
...
         price: float,
+        leverage: float = 1.0
     ) -> float:
         max_stake_amount = self._get_stake_amount_limit(pair, price, 0.0, 'max')
         if max_stake_amount is None:
             # * Should never be executed
             raise OperationalException(f'{self.name}.get_max_pair_stake_amount should'
                                        'never set max_stake_amount to None')
-        return max_stake_amount
+        return max_stake_amount / leverage
```
samgermain, line 705: Just wanted to make sure you noticed this, it not be needed, or might be used incorrectly if we use the response wrong, but it makes sense

xmatthias:
> @xmatthias
> 
> With the Okex implementation of `get_max_pair_stake_amount`, `test_execute_entry` is now failing with `E TypeError: '<' not supported between instances of 'float' and 'MagicMock'` even with `get_max_pair_stake_amount` in `mocker.patch.multiple`


The thing you need to understand is what you're mocking.
the mock is against `Exchange` - if Okex (or any other exchange) overwrites this, then that'll also overwrite the mock.
You get a MagicMock exception since ccxt as a whole is mocked via MagicMock when using `get_patched_exchange` (basically accepting whatever you call `_api.xxxx()` with and returning a new magicmock for each new call).
This means, the "okex" mock is missing/wrong. A quick fix (the only one i know without looking too deep) is to patch the function explicitly in the okex class (in addition to the generic mock against `Exchange`.

freqtrade/exchange/exchange.py
```diff
@@ -1852,26 +1855,119 @@ def _get_funding_fees_from_exchange(self, pair: str, since: Union[datetime, int]
...
             return 1.0
 
+        if self._api.has['fetchLeverageTiers']:
+
+            # Checks and edge cases
+            if stake_amount is None:
+                raise OperationalException(
+                    'binance.get_max_leverage requires argument stake_amount')
+            if pair not in self._leverage_tiers:
+                tiers = self.get_leverage_tiers_for_pair(pair)
```
xmatthias: If you use any "active" calls here (calls that need the ccxt object) - then we cannot use this method in backtesting/hyperopt.

The same will also become a problem for calculating the liquidation price - but for backtesting - it is a hard requirement to first gather all data from the exchange - and only then start backtesting (it's because of hyperopt, as well as for performance reasons - as we might end up calling the same method 1000nds of times for the same pair).


I'm also not sure why this change is necessary - the prior method seemed to contain what's needed (at least for binance). 
The traffic / memory requirement also seems to heavily increase (we're going from 1kb response to a 16kb response (which we will also have to keep in memory) in case of binance for example).

tests/exchange/test_binance.py
```diff
@@ -162,168 +171,369 @@ def test_stoploss_adjust_binance(mocker, default_conf, sl1, sl2, sl3, side):
...
-                     [200000000.0, 0.25],
-                     [300000000.0, 0.5]],
-        "ZEC/USDT": [[0.0, 0.01],
-                     [5000.0, 0.025],
-                     [25000.0, 0.05],
-                     [100000.0, 0.1],
-                     [250000.0, 0.125],
-                     [1000000.0, 0.5]],
+    api_mock.fetch_leverage_tiers = MagicMock(return_value={
+        'ADA/BUSD': [
```
xmatthias, line 168: Is there a substancial difference between this data and the data in conftest? (like - are there fields that are binance specific and we use)?

if it's only different pairs - i'd keep the data in one place (which is conftest - not individual test files).

freqtrade/exchange/exchange.py
```diff
@@ -2149,6 +2235,66 @@ def dry_run_liquidation_price(
...
                 "Freqtrade only supports isolated futures for leverage trading")
 
+    @retrier
+    def get_leverage_tiers_for_pair(self, pair: str):
+        # When exchanges can load all their leverage tiers at once in the constructor
+        # then this method does nothing, it should only be implemented when the leverage
+        # tiers requires per symbol fetching to avoid excess api calls
+        if (
+            self._api.has['fetchLeverageTiers'] and
+            not self._ft_has['can_fetch_multiple_tiers'] and
```
samgermain: Regarding [this comment](https://github.com/freqtrade/freqtrade/pull/6373#discussion_r801281879), whether we access the api or not will depend on whether we can fetch all the tiers for all markets at once or not, some exchanges(Okex) can only fetch them one market at a time, we also only fetch the tier if we have not fetched it before, which means I'll probably have to add a flag for `false` in the leverage_tiers dictionary if we can't get it. 

Binance can fetch them all at once, so it will just execute the `else` branch in this method

tests/exchange/test_binance.py
samgermain: `conftest` only has the output of this method, here I'm testing if we get the expected output from a given input

tests/exchange/test_okex.py
```diff
@@ -0,0 +1,166 @@
...
+    default_conf,
+    mocker,
+):
+    api_mock = MagicMock()
+    default_conf['trading_mode'] = 'futures'
+    default_conf['margin_mode'] = 'isolated'
+    default_conf['dry_run'] = True
+    api_mock.fetch_leverage_tiers = MagicMock(return_value={
+        'SHIB/USDT:USDT': [
+            {
```
samgermain: Here is another instance where I'm testing expected inputs and outputs, but with the Okex api response (Okex doesn't have the maintenance amount in info)

tests/exchange/test_binance.py
xmatthias: is that input exchange-specific to binance? I'd assume ccxt will return the same(ish) for all exchanges?

tests/exchange/test_okex.py
xmatthias: Do we need "info"? why?

in my experience, "info" is the raw exchange response - which i'd assume ccxt is abstracting away.
or is some field missing in the unification (i was assuming this is a new endpoint you added, so should have everything we need).

freqtrade/exchange/exchange.py
xmatthias: That'd mean we can't support okex futures (nor any other exchange that has this requirement), as we have to disable backtesting for exchanges that use "online" calls.

I'd instead have a "fetch_for_pairs" (or whatever) method we can call with the whitelist of backtesting, which then populates the fields correctly (obviously it'll be slow - but only at the start).

also - is this data for okex changing that often? maybe we can cache it the same way we do the binance data?



the way you're implementing it here is problematic in another way too and needs to be changed.
assuming we try to buy the same pair 10 times - we'd to 10x the same request, which is unnecessary - so once fetched, it should be added to the cached data.

tests/exchange/test_binance.py
samgermain: The data in `info` is exchange specific to binance, and we use the variable `cum` in info

tests/exchange/test_okex.py
samgermain: The binance response contains `cum` in info, which is the maintenance amount, which is needed in the binance liquidation price formula, the way we we're calculating it before gave the wrong numbers, and I thought it was easier just to use the number that the exchange returns. This here is just testing that `maintAmt` gets set to `None` with the Okex response. 

The reason `cum` is only in info is that `binance` is the only exchange that returns it, so it doesn't make sense to put it in the unified response, but it is required for the binance liquidation price formula

freqtrade/exchange/exchange.py
samgermain: > That'd mean we can't support okex futures (nor any other exchange that has this requirement), as we have to disable backtesting for exchanges that use "online" calls.

It's a private endpoint for binance, so we store the data in `binance_leverage_tiers.json`, we could do the same for Okex, 

> I'd instead have a "fetch_for_pairs" (or whatever) method we can call with the whitelist of backtesting, which then populates the fields correctly (obviously it'll be slow - but only at the start).

Yeah I'd like to make the functionality a bit better, so that there's less of a delay when you actually need to make a trade, I'm not sure if that needs to be done in this PR, it's more of an enhanced functionality type thing, because we can just load the tiers the same way we do with binance

> also - is this data for okex changing that often? maybe we can cache it the same way we do the binance data?

It already is, the methods that call this method only do so if the pair isn't already saved in `_leverage_tiers`

xmatthias: all of these "static" data endpoints (`get_min/max_amount/leverage`) are supposed to be used by backtesting as well (even though that may not be implemented yet for all instances).

This means - we must be certain which methods are going to call the exchange and which ones aren't.

The way `get_max_pair_stake_amount()` for okex is implemented - it's potentially (almost certainly on the first call per pair) going to call the exchange - which will simply result in an error when running hyperopt.

The same problem (but worse) happens in `get_max_leverage()` - where it can happen on every exchange.

Fetching things like max/min amounts and leverage must be offline calls using data that's available (ideally from markets - or cached elsewhere), otherwise they cannot be used in backtesting - which means backtesting would not work.

For okex - maybe a "call for required pairs" (like - `load_leveraged_tiers_for_pairs()`) can be added which is called for backtesting so we can fetch the data as required.

i don't really have an answer for the remainder though, as you MUST be careful where you add online calls to not break backtesting support (which you clearly didn't care in this PR so far).

samgermain: The okex leverage tiers are loaded the same way as binance now, it would probably be a better idea to load all the ones you need at the beginning though to get up to date information (maybe give an option for binance too? To load all the leverage tiers via the private endpoint, and default to the json file if the private endpoint won't work?). I don't know if that needs to be done immediately though

----------------------

For reference, I just wrote a script that called `fetchLeverageTiers` for every linear swap symbol, I split the list of symbols in 2 and ran it twice, so we could run it for most pairlists from the beginning

```
import ccxt
import json


okex = ccxt.okex()
okex.loadMarkets()

markets = okex.markets
swaps = []
for symbol, market in markets.items():
    if (market["swap"] and market["linear"]):
        swaps.append(market["symbol"])

swaps = swaps[:len(swaps)//2]
# swaps = swaps[len(swaps)//2:]

tiers = {}
for symbol in swaps:
    res = okex.fetchLeverageTiers(symbol)
    tiers[symbol] = res[symbol]

filename = './okex_leverage_tiers.json'

with open(filename, 'w+') as file:
    file.write(json.dumps(tiers))
```

samgermain:
@xmatthias I think this PR should mostly be acceptable by now, let me know if you want anything different. I'm going to keep working on the compat tests though

freqtrade/exchange/exchange.py
samgermain: > The traffic / memory requirement also seems to heavily increase (we're going from 1kb response to a 16kb response (which we will also have to keep in memory) in case of binance for example).

We don't store the whole response in memory, and the calculation of max leverage was incorrect before, I don't think you can calculate the max leverage from just the maintenance margin ratio, the maintenance amount calculation was wrong too I think.

samgermain: Technically we could get rid of the `max` field, if we saved the last max value to the markets, and then always checked what the min of the next tier was (because it would be equal to max).

We could also save it as an array where we just know what each index is if you think that the memory requirement will be an issue

The old method didn't give enough info though,

freqtrade/exchange/okex.py
```diff
@@ -46,3 +47,37 @@ def _lev_prep(
...
+            symbols = []
+
+            for symbol, market in markets.items():
+                if (market["swap"] and market["linear"]):
+                    symbols.append(market["symbol"])
+
+            tiers = {}
+            for symbol in symbols:
+                res = self._api.fetchLeverageTiers(symbol)
+                tiers[symbol] = res[symbol]
```
samgermain: @xmatthias So we could probably do something like where we load all the leverage tiers for okex at the beginning, is there anything special we need to do to run this at the beginning of backtesting or hyperopt?

I got rid of the get_leverage_tiers_for_pair method

xmatthias: at the moment there isn't - but my idea was to run a method (when futures mode) which passes the pairlist in - so we only fetch the pairs necessary. Obviously that method is "empty" for most exchanges (or disabled?) and only fetches the data for exchanges which need it (i guess based on the _ft_has entry ...)

freqtrade/exchange/okx.py
```diff
@@ -46,3 +47,37 @@ def _lev_prep(
...
+            markets = self.markets
+            symbols = []
+
+            for symbol, market in markets.items():
+                if (market["swap"] and market["linear"]):
+                    symbols.append(market["symbol"])
+
+            tiers = {}
+            for symbol in symbols:
+                res = self._api.fetchLeverageTiers(symbol)
```
xmatthias: This concerns me a bit from a performance perspective.

it's fetching data for all pairs (unfiltered). 
It'll delay startup (that might be acceptable) - but it'll also delay "`reload_markets()`" - where it could lead to losses (or missed profits) (assume this method runs for 2 minutes - that's 2 minutes freqtrade will not do anything else.

I'd rather use some intelligence here ... like
* only for pairs with matching stake-currency (that should bring the list down quite some)

freqtrade/exchange/okx.py
```diff
@@ -46,3 +47,37 @@ def _lev_prep(
...
+        pair_tiers = self._leverage_tiers[pair]
+        return pair_tiers[-1]['max'] / leverage
+
+    def load_leverage_tiers(self) -> Dict[str, List[Dict]]:
+        if self.trading_mode == TradingMode.FUTURES:
+            markets = self.markets
+            symbols = []
+
+            for symbol, market in markets.items():
+                if (market["swap"] and market["linear"]):
```
xmatthias: please use the exchange provided methods for this (in this example - `market_is_future`).
this will keep the logic in one place and will avoid errors in the future.

For example for okex - if we need to check for both swap AND linear - then this should be added to `market_is_future` too.
I've adjusted this now - but an adjustment to `market_is_future` may be necessary (either for okx only, or for all exchanges).

xmatthias: In corellation to this there's also the only failing test (`test_get_markets()`) - which is failing due to the changes in conftest.
"probably" - the fix is in `market_is_future` - but before touching this i'd like your opinion - as you know more about the market-structure in ccxt than i do (basically: "do we need to filter on `type=swap + 'linear' == True` for all exchanges?")

samgermain: > In corellation to this there's also the only failing test (`test_get_markets()`) - which is failing due to the changes in conftest. "probably" - the fix is in `market_is_future` - but before touching this i'd like your opinion - as you know more about the market-structure in ccxt than i do (basically: "do we need to filter on `type=swap + 'linear' == True` for all exchanges?")

Linear is USDT(or BUSD, its margined in whatever the quote currency is) margined, and inverse is coin margined, we haven't written all the functionality for coin margined yet, so for freqtrade in it's current state, it's necessary to filter for `linear` but it's not necessary in general

The `linear` symbols look like `BTC/USDT:USDT`(settled in USDT) or `ETH/BUSD:BUSD`(settled in BUSD) while the `inverse` ones look like `BTC/USD:BTC`(settled in BTC) or `ETH/USD:ETH` (settled in ETH)

freqtrade/exchange/okx.py
samgermain: > * only for pairs with matching stake-currency

`linear` will do that on okex, because they're all USDT margined, it's about 90 some markets on `okex`, what you wrote is probably better though

we should filter just for the pairlist, but I don't know how to access that from here

tests/exchange/test_ccxt_compat.py
```diff
@@ -341,6 +341,7 @@ def test_get_max_leverage_spot(self, exchange):
 
     def test_get_max_leverage_futures(self, exchange_futures):
         futures, futures_name = exchange_futures
```
xmatthias, line 327: I do think that this test is called (and running) for the mentioned exchanges ... ?

tests/exchange/test_okx.py
```diff
@@ -0,0 +1,301 @@
...
+
+    assert exchange.get_max_pair_stake_amount('BNB/BUSD', 1.0) == 30000000
+    assert exchange.get_max_pair_stake_amount('BNB/USDT', 1.0) == 50000000
+    assert exchange.get_max_pair_stake_amount('BTC/USDT', 1.0) == 1000000000
+    assert exchange.get_max_pair_stake_amount('BTC/USDT', 1.0, 10.0) == 100000000
+
+    assert exchange.get_max_pair_stake_amount('TTT/USDT', 1.0) == float('inf')  # Not in tiers
+
+
+# def test_load_leverage_tiers_okx(default_conf, mocker):
```
xmatthias: ?

tests/exchange/test_exchange.py
```diff
@@ -4220,3 +4231,154 @@ def test_get_max_pair_stake_amount(
...
+
+def test_load_leverage_tiers(mocker, default_conf, leverage_tiers):
+    api_mock = MagicMock()
+    api_mock.fetch_leverage_tiers = MagicMock()
+    type(api_mock).has = PropertyMock(return_value={'fetchLeverageTiers': True})
+    default_conf['dry_run'] = False
+    default_conf['trading_mode'] = 'futures'
+    default_conf['margin_mode'] = 'isolated'
+
+    ccxt_exceptionhandlers(
```
xmatthias, line 4313: are you intentionally only testing the exception handlers? (i noticed firstly because `leverage_tiers` is not used)

In theory, we could have the "return" statement missing in the function - and we'd never notice ... 
while now it's a plain "passthrough" of ccxt, nothing guarantees that stays this way ...

tests/exchange/test_exchange.py
```diff
@@ -4220,3 +4231,154 @@ def test_get_max_pair_stake_amount(
...
+@ pytest.mark.parametrize('pair,value,mmr,maintAmt', [
+    ('ADA/BUSD', 500, 0.025, 0.0),
+    ('ADA/BUSD', 20000000, 0.5, 1527500.0),
+    ('ZEC/USDT', 500, 0.01, 0.0),
+    ('ZEC/USDT', 20000000, 0.5, 654500.0),
+])
+def test_get_maintenance_ratio_and_amt(
+    mocker,
+    default_conf,
+    leverage_tiers,
```
xmatthias, line 4412: Is `leveraged_tiers` not needed? if it's not - i'd remove it to keep tests clean

tests/exchange/test_binance.py
```diff
@@ -148,7 +148,15 @@ def test_stoploss_order_dry_run_binance(default_conf, mocker):
     (1501, 1499, 1501, "sell"),
     (1499, 1501, 1499, "buy")
 ])
-def test_stoploss_adjust_binance(mocker, default_conf, sl1, sl2, sl3, side):
+def test_stoploss_adjust_binance(
+    mocker,
+    default_conf,
+    leverage_tiers,
```
xmatthias: if we don't need a fixture - please don't include it.

VScode is showing me unused parameters via color highlight - which makes it quite easy to notice.

freqtrade/exchange/okx.py
xmatthias: we're not looking at "linear" or not to determine futures (nowhere afaik) - but at `ccxt_futures_name` - which is "future" for okex.

I guess we'll have to "enhance" `market_is_future` to include "linear" - but it's unclear to me if we need to do that for okx only, or for all exchanges.

tests/exchange/test_ccxt_compat.py
samgermain: Right, it won't work for gateio yet, and we'll need to update CCXT

freqtrade/exchange/okx.py
xmatthias: For the moment, i'd go with "point out it's going to be slow".

the actual pairlist is not available when the exchange is first loaded (loading volumepairlist requires the exchange).
it currently takes ~45 seconds - which is okish (slow but acceptable).
if we see it's causing problems, we can think of something else...

tests/exchange/test_okx.py
samgermain: I need the mock the first and second response of an api call, and I can't remember how thats done

freqtrade/exchange/okx.py
samgermain: > we're not looking at "linear" or not to determine futures (nowhere afaik) - but at `ccxt_futures_name` - which is "future" for okex.
> 
> I guess we'll have to "enhance" `market_is_future` to include "linear" - but it's unclear to me if we need to do that for okx only, or for all exchanges.

All the markets have to be linear, we haven't written functionality for inverse markets

freqtrade/exchange/okx.py
```diff
@@ -46,3 +47,48 @@ def _lev_prep(
...
+            # Be verbose here, as this delays startup by ~1 minute.
+            logger.info(
+                f"Initializing leverage_tiers for {len(symbols)} markets. "
+                "This will take about a minute.")
+
+            for symbol in sorted(symbols):
+                res = self._api.fetch_leverage_tiers(symbol)
+                tiers[symbol] = []
+                for tier in res[symbol]:
+                    tiers[symbol].append(self.parse_leverage_tier(tier))
```
xmatthias: I think we'll need to be careful with this
this test fails because we call `parse_leverage_tier()` both here and in `fill_leverage_tiers()`
Obviously parsing the same thing twice on the same object will cause it to fail as we're significantly changing the structure as part of this method.

freqtrade/exchange/gateio.py
```diff
@@ -31,7 +31,7 @@ class Gateio(Exchange):
         # TradingMode.SPOT always supported and not required in this list
         # (TradingMode.MARGIN, MarginMode.CROSS),
         # (TradingMode.FUTURES, MarginMode.CROSS),
-        (TradingMode.FUTURES, MarginMode.ISOLATED)
+        # (TradingMode.FUTURES, MarginMode.ISOLATED)
```
xmatthias: Is this intentional? gate.io works better than okx (which is failing completely the moment you try to buy)...

freqtrade/exchange/okx.py
```diff
@@ -19,15 +22,17 @@ class Okx(Exchange):
         "ohlcv_candle_limit": 300,
         "mark_ohlcv_timeframe": "4h",
         "funding_fee_timeframe": "8h",
+        "can_fetch_multiple_tiers": False,
     }
 
     _supported_trading_mode_margin_pairs: List[Tuple[TradingMode, MarginMode]] = [
         # TradingMode.SPOT always supported and not required in this list
         # (TradingMode.MARGIN, MarginMode.CROSS),
```
xmatthias, line 30: I think we should only change this once we know we can place orders on the exchange (missed line, the actual comment should be 1 line below)

okex fails when creating an order (while i've fixed one error for now, it's still not really working)

freqtrade/exchange/gateio.py
samgermain: `gateio.fetchLeverageTiers` isn't implemented yet, but I just figured out how to implement it, and it will probably be merged today or tommorow

freqtrade/exchange/okx.py
samgermain: Whats the error you're receiving?

freqtrade/exchange/gateio.py
xmatthias: i did do some futures trading for gate.io

leverageTiers may be required in the end to get details right - but to test with smallish stake i don't see why we couldn't (unless this PR breaks gate.io again - which i didn't really test).

freqtrade/exchange/okx.py
xmatthias: multiple?
i've fixed one (set_leverage called with the wrong params), ran into the next one ... simply put - it's clear that freqtrade never made a single live futures trade on okex - hence i don't think we're ready to uncomment this line.

While we could use this PR to fix these problems (and i did to some extend) - i don't think we should - as this should really be about "leverage" / "leverage tiers", not about "let's get okx working".
That'd be also an "okish" approach - but it would delay a PR that's quite complete as is now i think.

xmatthias: I've fixed these errors now ... it's additional parameters that are needed for creating orders in OKX (as well as a different parameter set for `set_leverage()`.

freqtrade/exchange/gateio.py
samgermain: > i did do some futures trading for gate.io
> 
> leverageTiers may be required in the end to get details right - but to test with smallish stake i don't see why we couldn't (unless this PR breaks gate.io again - which i didn't really test).

I added the version of `get_max_leverage` and `get_maintenance_ratio_and_amt` that work for gateio as gateio methods, so that they override the version on exchange.py. Unless someone plans to trade over 10,000 at 100x leverage it should be accurate for gateio actually

freqtrade/exchange/exchange.py
```diff
@@ -1891,7 +1891,7 @@ def parse_leverage_tier(self, tier) -> Dict:
         return {
             'min': tier['notionalFloor'],
             'max': tier['notionalCap'],
-            'mmr': tier['maintenanceMarginRatio'],
+            'mmr': tier['maintenanceMarginRate'],
```
xmatthias, line 1894: According to the tests against exchanges - this change seems wrong?

I had also tested this "live" - and it did work for okex - but i don't think it had maintenanceMarginRate ...

samgermain: It's `maintenanceMarginRate` on binance, and it's supposed to be `maintenanceMarginRate`. I posted a [PR](https://github.com/ccxt/ccxt/pull/11985) to CCXT fixing this

This was my mistake, sorry, I guess we should just wait until that PR gets merged

xmatthias: ok then it makes sense
Once it's merged, please bump the ccxt version - if everything goes well, tests should pass then.

xmatthias: @samgermain 
looking at the error in the first CI try (`ccxt.base.errors.RateLimitExceeded: okx {"msg":"Requests too frequent.","code":"50011"}`) - maybe okx rate-limits in ccxt are too low (maybe okx changed them recently or the weight for the endpoint is higher than expected)?

We're intentionally not overwriting rateLimit - as ccxt limits are usually OK.

samgermain: The [api endpoint](https://www.okx.com/docs-v5/en/#rest-api-public-data-get-position-tiers) for `fetchLeverageTiers` is set up to send a request no more than once every 200 ms which is what is specified in Okx's docs



If we have two PRs being tested at once though, then would we be sending requests from the same api on multiple instances of the bot?

xmatthias: for sure - but it all runs on the github actions infrastructure - and i don't think it'll be the same exit IP's (that's surely a possibility - but a pretty low one).

It's however a risk we MUST take if we want to test against live exchange endpoints.
we also try to avoid that by killing prior runs on the same branch - so having 2 run in parallel (which actually reach this stage) requires 2 runs to happen from different branches.

kroitor: @xmatthias if you can paste a shortened snippet to reproduce the rate limiting error in py – we will investigate and fix that in ccxt, thx!

samgermain: > requires 2 runs to happen from different branches.

That's what I was referrin to, we got another CCXT member to test out the rate limiter and he also thinks its from running two PRs at the same time also. 

Is it possible to limit this test so that we don't have to test it on python 8, 9 and 10 on windows, mac and linux? I feel like we would run into less problems if we limited it to one operating system. 

We could even set up a russian roulette situation for it to have a 1/3 chance of running, so then the probability of it not being tested all 9 times would be (2/3)^9 = 2.6%

xmatthias: `--longrun` is only running on 3.9 - and only on linux. Can't limit it more than this, really.

the errors are visible in this [ci run](https://github.com/freqtrade/freqtrade/runs/5226543636?check_suite_focus=true). 
It's running on github action cloud workers - which is a shared IP range in Azure datacenters afaik - so in theory it could've been others running tests against okx as well running into this problem.

as far as i can tell, it was our only CI job running at the time (when the job was running, this was the only CI job which had extensive calls against okx).

I'd say we monitor this going forward in feat/short - it was just a "quick suspicion" - i've also not ran into it locally so far.

PR #6384, opened 2022-02-10
test_ccxt__get_contract_size

samgermain:
It looks like FTX is the only exchange that's failing, and it looks like FTX isn't using the unified symbols for some reason, it's still using the symbol like `BTC-PERP`, which was updated on CCXT a while ago


xmatthias:
your local ccxt version is outdated.
the screenshot contains the version - 1.63.x (we're now at 1.73.x or so).

xmatthias (approved):
LGTM :+1:

PR #6386, opened 2022-02-10
Added liquidation_buffer to freqtradebot

samgermain:
Cuts the liquidation price short by a default of 5% that can be adjusted in the users config, to help prevent liquidation and liquidation fees to the user


freqtrade/freqtradebot.py
```diff
@@ -103,8 +103,8 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self._exit_lock = Lock()
         LoggingMixin.__init__(self, logger, timeframe_to_seconds(self.strategy.timeframe))
 
+        self.liquidation_buffer = float(self.config.get('liquidation_buffer', '0.05'))
```
xmatthias, line 106: can you please also add this to the documentation?

also - if you specify the schema in `constants.py` (as number) - you'll not need to parse it explicitly as float (assuming the fallback is also a float, obviously).

This way (without schema) we risk it failing "mid-buy" - if it can't be parsed to a float.

I also don't see a need to do this "on init" (basically assigning it to the freqtrade object) - but we can do the config lookup while executing the trade.

freqtrade/freqtradebot.py
```diff
@@ -756,6 +756,14 @@ def execute_entry(
         funding_fees = self.exchange.get_funding_fees(
             pair=pair, amount=amount, is_short=is_short, open_date=open_date)
         # This is a new trade
+        if isolated_liq:
+            liquidation_buffer = abs(enter_limit_filled_price -
+                                     isolated_liq) * self.liquidation_buffer
+            isolated_liq = (
+                isolated_liq - liquidation_buffer
+                if is_short else
+                isolated_liq + liquidation_buffer
```
xmatthias: I think the calculation should be separated into a separate function (not in `freqtradebot.py`). 
We'll need it in backtesting as well ... so i'd maybe place it in the exchange class

docs/leverage.md
```diff
@@ -69,6 +69,18 @@ One account is used to share collateral between markets (trading pairs). Margin
...
+*Defaults to `0.05`.*
+
+A ratio specifying how large of a safety net to place between the liquidation price and the stoploss to prevent a position from reaching the liquidation price
+
+Possible values are any floats between 0.0 and 0.99
+
+**ex:** If a trade is entered at a price of 10 coin/USDT, and the liquidation price of this trade is 8 coin/USDT, then with `liquidation_buffer` set to `0.05` the minimum stoploss for this trade would be 8 + ((10 - 8) * 0.05) = 8 + 0.1 = 8.1
+
+!!! Danger "A `liquidation_buffer` of 0.0, or a low `liquidation_buffer` is likely to result in liquidations, and liquidation fees"
+Currently Freqtrade is able to calculate liquidation prices, but does not calculate liquidation fees. Setting your `liquidation_buffer` to 0.0, or using a low `liquidation_buffer` could result in your positions being liquidated. Freqtrade does not track liquidation fees, so liquidations will result in accurate profit/loss results for your bot. If you use a low `liquidation_buffer`, it is recommended to use `stoploss_on_exchange`.
```
xmatthias:
```suggestion
Currently Freqtrade is able to calculate liquidation prices, but does not calculate liquidation fees. Setting your `liquidation_buffer` to 0.0, or using a low `liquidation_buffer` could result in your positions being liquidated. Freqtrade does not track liquidation fees, so liquidations will result in inaccurate profit/loss results for your bot. If you use a low `liquidation_buffer`, it is recommended to use `stoploss_on_exchange` if your exchange supports this.
```

freqtrade/exchange/exchange.py
```diff
@@ -135,13 +135,18 @@ def __init__(self, config: Dict[str, Any], validate: bool = True) -> None:
...
         self.margin_mode: Optional[MarginMode] = (
             MarginMode(config.get('margin_mode'))
             if config.get('margin_mode')
             else None
         )
+        self.liquidation_buffer = config.get('liquidation_buffer', 0.05)
+        if self.liquidation_buffer < 0.0:
+            raise OperationalException('Cannot have a negative liquidation_buffer')
+        if self.liquidation_buffer > 0.99:
+            raise OperationalException('Liquidation_buffer must be below 0.99')
```
xmatthias: This is a setting that's coming from the configuration - so it should be checked right there (adding this to `constants.py` - as mentioned here.

please allow the configuration schema to do the validation (you can set min-max values in the configuration).

xmatthias (changes requested):
Please add the validation to the json schema in `constants.py`.

Also - i would ask you to please read the comments i add to reviews. 
It's honestly very tiring and frustrating to ask for changes - have changes implemented, and then have to ask for the **same exact thing** again.
I'm mentioning this as this is happening quite frequently in your PR's - which is wasting a lot of time (both your's and mine) and could easily be avoided.

freqtrade/freqtradebot.py
samgermain: I feel like I just like it being inside to exchange.init more than doing a config lookup when executing the trade, but I can move it if you want

xmatthias: It's a configuration setting - having it not in the config validation will make it crash with "ugly" errors (float conversion failed for example).

e.g. `"liquidation_buffer": "abc",`
I know this is wrong - but it's user input, so we have to validate it.

xmatthias (approved):
LGTM :+1:

italodamato:
is it possible to add some logging for this parameter as well as `futures_funding_rate`?

xmatthias:
`futures_funding_rate` is a fallback that's only used if the exchange doesn't provide a value for that.

If it's there - it's in your config - so best look there.
I don't see us logging this - it'll only be confusing for no good reason
The same is true for `liquidation_buffer`. It's got a sensible default (5%) that's documented.

The only way to change it is though your config - so if you can read logs - the config is just nearby :)

italodamato:
I just think it'd be useful to provide some form of feedback to make sure the settings are getting applied correctly.

xmatthias:
if you think about it - then that's not really true.

You'd have to apply that to all settings (the ones mentioned here are not special, nor really important) - which are around 150 - not considering freqAI.
if you however get a startup log of 150+ messages - you'll not even bother to look through that and try and see if "my setting applied correctly" - so it becomes pointless noise.

The things that are logged are the ones that can be set in either config or strategy.

PR #6391, opened 2022-02-11
feature: add buy tag to forcebuy

lukasgor:

## Summary

Adding the possibility to provide a buy tag for forcebuy

## Quick changelog

- when calling forcebuy method from API you can include the buy tag as well

## What's new?

See above

### DIFF
--- a/freqtrade/rpc/api_server/api_schemas.py
+++ b/freqtrade/rpc/api_server/api_schemas.py
@@ -280,6 +280,7 @@ class ForceBuyPayload(BaseModel):
     price: Optional[float]
     ordertype: Optional[OrderTypeValues]
     stakeamount: Optional[float]
+    buy_tag: Optional[str]
 
 
 class ForceSellPayload(BaseModel):
--- a/freqtrade/rpc/api_server/api_schemas.py
+++ b/freqtrade/rpc/api_server/api_schemas.py
@@ -280,7 +280,7 @@ class ForceBuyPayload(BaseModel):
     price: Optional[float]
     ordertype: Optional[OrderTypeValues]
     stakeamount: Optional[float]
-    buy_tag: Optional[str]
+    entry_tag: Optional[str]
 
 
 class ForceSellPayload(BaseModel):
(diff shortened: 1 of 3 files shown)


freqtrade/rpc/rpc.py
```diff
@@ -717,7 +717,8 @@ def _exec_forcesell(trade: Trade) -> None:
             return {'result': f'Created sell order for trade {trade_id}.'}
 
     def _rpc_forcebuy(self, pair: str, price: Optional[float], order_type: Optional[str] = None,
-                      stake_amount: Optional[float] = None) -> Optional[Trade]:
+                      stake_amount: Optional[float] = None,
+                      buy_tag: Optional[str] = None) -> Optional[Trade]:
```
xmatthias, line 721: while `buy_tag` is generally a user-controlled field (it comes from the strategy) - i fear this has potential negative side-effects.

In theory, you could insert a malicious string here, which might (!) be shown in a ui component - effectively resulting in a "stored XSS" attack.

That's not possible with other fields i'm aware of - so it's something i'll need to think about a bit more.

freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -280,6 +280,7 @@ class ForceBuyPayload(BaseModel):
     price: Optional[float]
     ordertype: Optional[OrderTypeValues]
     stakeamount: Optional[float]
+    buy_tag: Optional[str]
```
xmatthias: In light of current proceedings in `feat/short` - i'd ask you to name this `entry_tag` instead.

while it'll create a short-term misalignment (the API is sending `entry_tag` - but it end's up in buy_tag, and the client get's "buy_tag" back) - i think it's better to have it right immediately, rather than renaming it a month later.

lukasgor: sure, just updated it.

freqtrade/rpc/rpc.py
lukasgor: good point, I can add some method that will strip out malicious tags from this string maybe? Let me know if you have any other idea

xmatthias: that'll be pretty difficult - there's too much possibility to obfuscate tags or whatnot.
it'll also depend on the usecase - in freqUI, vue will render this as string and you have to jump through hoops to have it render a "real" injected script-tag ...

but freqUI is potentially not the only API consumer - and it'll be difficult to protect against unknown consuming languages.

xmatthias (approved):
Let's keep it as is for now ...

but we may add additional filters should problems arise
LGTM :+1:

PR #6394, opened 2022-02-12
todo removal

samgermain:


freqtrade/edge/edge_positioning.py
```diff
@@ -159,7 +159,7 @@ def calculate(self, pairs: List[str]) -> bool:
         logger.info(f'Measuring data from {min_date.strftime(DATETIME_PRINT_FORMAT)} '
                     f'up to {max_date.strftime(DATETIME_PRINT_FORMAT)} '
                     f'({(max_date - min_date).days} days)..')
-        #  TODO-lev: Should edge support shorts? needs to be investigated further...
+        #  * Should edge support shorts? needs to be investigated further (add enter_short exit_short)
```
samgermain: We won't be implementing edge immediately, so I thought this was fine to remove as a todo

freqtrade/plugins/protections/stoploss_guard.py
```diff
@@ -51,7 +51,6 @@ def _stoploss_guard(self, date_now: datetime, pair: str = None) -> ProtectionRet
         # if pair:
         #     filters.append(Trade.pair == pair)
         # trades = Trade.get_trades(filters).all()
-        # TODO-lev: Liquidation price?
```
samgermain, line 54: I believe the liq will just be recorded as a stoploss, so we shouldn't have to worry about it here

freqtrade/edge/edge_positioning.py
xmatthias: yes - at least as "TODO-lev". I've readded it as regular TODO - otherwise it's a regular comment and won't be found if someone is looking for "tasks".

PR #6397, opened 2022-02-12
edited todo-lev comments

samgermain:


freqtrade/commands/data_commands.py
```diff
@@ -160,7 +160,6 @@ def start_list_data(args: Dict[str, Any]) -> None:
     from freqtrade.data.history.idatahandler import get_datahandler
     dhc = get_datahandler(config['datadir'], config['dataformat_ohlcv'])
 
-    # TODO-lev: trading-mode should be parsed at config level, and available as Enum in the config.
```
samgermain: I don't think we should have a is_short_strategy flag that's at the config, I think that just the presence of enter_short exit_short should be enough, and we can through an exception if it's spot.

It would be a good idea to be able to turn off shorts or longs entirely based on the market though (if the market goes bull, turn off short, if it goes bear turn off long), but that wouldn't be in the config, it would be dynamic within the strategy

freqtrade/enums/candletype.py
```diff
@@ -8,8 +8,10 @@ class CandleType(str, Enum):
     MARK = "mark"
     INDEX = "index"
     PREMIUMINDEX = "premiumIndex"
-    # TODO-lev: not sure this belongs here, as the datatype is really different
+
+    # TODO: Could take up less memory if these weren't a CandleType
     FUNDING_RATE = "funding_rate"
+    BORROW_RATE = "borrow_rate"  # * unimplemented
```
samgermain: If margin ever gets implemented, borrow_rates should be downloaded the same as funding rates, right now CCXT has `fetch_borrow_rate_history` for FTX and Binance (but live interest data still isn't merged, and there's no method to pay interest yet)

freqtrade/freqtradebot.py
```diff
@@ -1049,7 +1049,7 @@ def handle_stoploss_on_exchange(self, trade: Trade) -> bool:
         Check if trade is fulfilled in which case the stoploss
         on exchange should be added immediately if stoploss on exchange
         is enabled.
-        # TODO-lev: liquidation price always on exchange, even without stoploss_on_exchange
+        # TODO: liquidation price always on exchange, even without stoploss_on_exchange
```
samgermain: I have no solution for this, maybe demand that stoploss on exchange be used for futures? Is there any reason to not use it if it's implemented?

freqtrade/freqtradebot.py
```diff
@@ -1736,7 +1736,7 @@ def fee_detection_from_trades(self, trade: Trade, order: Dict, order_amount: flo
                 trade.update_fee(fee_cost, fee_currency, fee_rate, order.get('side', ''))
 
         if not isclose(amount, order_amount, abs_tol=constants.MATH_CLOSE_PREC):
-            # TODO-lev: leverage?
+            # * Leverage could be a cause for this warning, leverage hasn't been thoroughly tested
```
samgermain, line 1739: from discord

sam -  @richardjozsa You've been running freqtrade in futures mode right? Have you been getting any of this warning in your log output "Amount {amount} does not match amount {trade.amount}"
richardjozsa -  of course I'm using every time the same setup from config and strategy side, that worked earlier, when I used my crude solutions,margin_mode: "isolated", trading_mode: "futures.". I haven't got that message, that message dissapared after you repaired it.

freqtrade/optimize/backtesting.py
```diff
@@ -127,8 +127,6 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self.config['startup_candle_count'] = self.required_startup
         self.exchange.validate_required_startup_candles(self.required_startup, self.timeframe)
 
-        # TODO-lev: This should come from the configuration setting or better a
-        # TODO-lev: combination of config/strategy "use_shorts"(?) and "can_short" from the exchange
```
samgermain: https://github.com/freqtrade/freqtrade/pull/6397/files#r805210092

freqtrade/optimize/backtesting.py
```diff
@@ -538,7 +536,7 @@ def _get_sell_trade_entry(self, trade: LocalTrade, sell_row: Tuple) -> Optional[
         sell_candle_time: datetime = sell_row[DATE_IDX].to_pydatetime()
 
         if self.trading_mode == TradingMode.FUTURES:
-            # TODO-lev: liquidation price?
+            # TODO: liquidation price?
```
samgermain: I don't remember why I wrote this here

freqtrade/templates/sample_short_strategy.py
```diff
@@ -15,7 +15,7 @@
 import freqtrade.vendor.qtpylib.indicators as qtpylib
 
 
-# TODO-lev: Create a meaningfull short strategy (not just revresed signs).
+# TODO: Create a meaningfull short strategy (not just revresed signs).
```
samgermain, line 18: Reversed signs is probably good enough for merge, once it's merged a meaningful strategy will show up soon

freqtrade/commands/data_commands.py
xmatthias: >  we can through an exception if it's spot.

Please start thinking about things from a user perspective ("how would that feel as a user").

If you throw an exception once a signal appears - you'd crash the bot potentially days into running the bot - which from a UX perspective - is VERY bad.

if you run spot there's 2 ways - either we ignore `enter_short` completely (no warning, just not done) - or we have a flag similar to `use_custom_stoploss` (for example) - which acts as "optin" for shorts - and blocks execution on spot markets (so you're aware that you're running a short strategy on spot markets).

Throwing an exeption if a signal on the wrong market appears is out of the question.

freqtrade/exchange/exchange.py
```diff
@@ -1528,7 +1528,7 @@ def refresh_latest_ohlcv(self, pair_list: ListPairsWithTimeframes, *,
         :return: Dict of [{(pair, timeframe): Dataframe}]
         """
         logger.debug("Refreshing candle (OHLCV) data for %d pairs", len(pair_list))
-        # TODO-lev: maybe depend this on candle type?
+        # TODO: maybe depend this on candle type?
```
xmatthias: i think this can be removed completely.
this method IS now depending on candle_type ... if it means something else, i don't understand the comment and it should be clarified. If it's not clear today, it'll also not be clear next month when someone looks through regular todo's.

freqtrade/optimize/backtesting.py
xmatthias: If you don't think it's necessary, please remove it.

we should only move TODO-lev to TODO if it's something we can postpone for later. If we don't know why it's there - it's either outdated, or we must analyze/fix it.

freqtrade/commands/data_commands.py
xmatthias: That said - the above TODO was about something else (moving the parsing of `trading_mode` from all over the place to one central location in `configuration.py` ) - which is not done and should therefore be kept there.

freqtrade/freqtradebot.py
xmatthias: I'd actually leave this for the moment (as TODO-lev so we don't forget).

if we assume that we set out "liquidation price offset" to 5% above the real liquidation (so we don't actually encounter a liquidation), then freqtrade will also handle this off exchange - _assuming_ we treat that price as stoploss (or rather - stoploss becomes `max(stoploss, liquidationprice)`).

freqtrade/freqtradebot.py
xmatthias: this is the fallback to fee fetching - and will only apply if  the order did not include fees.
I don't think that's the case on binance (although it may depend if it filled immediately).

So this is already a pretty rare case ...

freqtrade/freqtradebot.py
samgermain: > I'd actually leave this for the moment (as TODO-lev so we don't forget).
> 
> if we assume that we set out "liquidation price offset" to 5% above the real liquidation (so we don't actually encounter a liquidation), then freqtrade will also handle this off exchange - _assuming_ we treat that price as stoploss (or rather - stoploss becomes `max(stoploss, liquidationprice)`).

There's the possibility that the price blows through the 5% buffer though, especially if you're using high leverage

xmatthias: we have the same possibility for stoploss on exchange ... (we define the stop and the limit (by default 1% below) - which stops can blow through for sure.

That's also "not really" handled for stoploss on exchange either - so i guess keeping it as regular TODO will be fine (i'd however ask you to add the "blow through" comment to the file, so we know what it's about when looking at this months from now).

freqtrade/freqtradebot.py
samgermain: Ok, we should probably just leave the comment there, and once this is merged into stable it probably wouldnt be necessary anymore, this is just a heads up in case we arrive here

freqtrade/freqtradebot.py
```diff
@@ -1358,7 +1358,6 @@ def _safe_exit_amount(self, pair: str, amount: float) -> float:
         trade_base_currency = self.exchange.get_pair_base_currency(pair)
         wallet_amount = self.wallets.get_free(trade_base_currency)
         logger.debug(f"{pair} - Wallet: {wallet_amount} - Trade-amount: {amount}")
-        # TODO-lev: Get wallet amount + value of positions
```
samgermain: I actually don't think we need to do anything here, the amount that's in positions shows up in `USED`, same as the amount which is in a limit order

```
DOT: { free: 0, used: 0, total: 0 },
  BTC: { free: 0, used: 0, total: 0 },
  SOL: { free: 0, used: 0, total: 0 },
  BNB: { free: 0, used: 0, total: 0 },
  ETH: { free: 0, used: 0, total: 0 },
  ADA: { free: 0, used: 0, total: 0 },
  USDT: { free: 58.54689751, used: 1082.52790274, total: 1113.9274002 },
  XRP: { free: 0, used: 0, total: 0 },
  BUSD: { free: 0.07715642, used: 0, total: 0.07715642 },
  timestamp: undefined,
  datetime: undefined,
  free: {
    DOT: 0,
    BTC: 0,
    SOL: 0,
    BNB: 0,
    ETH: 0,
    ADA: 0,
    USDT: 138.54689751,
    XRP: 0,
    BUSD: 0.07715642
  },
  used: {
    DOT: 0,
    BTC: 0,
    SOL: 0,
    BNB: 0,
    ETH: 0,
    ADA: 0,
    USDT: 1002.52790274,
    XRP: 0,
    BUSD: 0
  },
  total: {
    DOT: 0,
    BTC: 0,
    SOL: 0,
    BNB: 0,
    ETH: 0,
    ADA: 0,
    USDT: 1113.9274002,
    XRP: 0,
    BUSD: 0.07715642
  }
}
```

xmatthias: that's wrong.

assuming you're in an XRP/USDT position - then the value we're looking for here is the XRP value - which is 0 in your case (there's no other value that would make sense either - so i'm taking XRP as sample).
it may show up as "used USDT" - but that's irrelevant - as we look for "currency" - not "amount.

It's currently not impacting because of the "trading_mode==Futures" below - but unless you can also remove the "or" condition below (which means something in the remaining code changed), this will have to stay as TODO-lev.

freqtrade/enums/candletype.py
xmatthias: Maybe add it as commented value here then.

this enum is also used as "valid informative pair candle types" - so having an unimplemented value here can become problematic if someone tries to use it.

freqtrade/freqtradebot.py
samgermain: I don't really understand what the problem is, it looks like this method checks for if you have enough currency to safely exit, but that's not an issue on futures, you can always close a position, so you don't need a safe exit amount

xmatthias: letme repeat what i said in another PR (again, it's REALLY tiring to explain the same thing to you over and over again!):
your assumption is that nothing else happens (like manual interference) - which we can't fully rely on.

What if the user closed the position manually? or it was liquidated? in both cases, the position would no longer be available, causing an exception (and potential other errors) while closing.

We're CURRENTLY excluding this check for futures - but that's a temporary measure until we use fetch_position to get the correct balance.

xmatthias: The main point here is - without fetching positions - output of `/wallet` is wrong too (we discussed this last time as well).
once we have wallets uptodate - we can also check if we reallly still own the position.

freqtrade/optimize/backtesting.py
xmatthias: I am honestly not convinced about the way for this.

do we really just ignore short files should we encounter them in a spot backtest/live run?
That's a decision in the end ... but i feel like usability / chance for error are pretty high with this approach

freqtrade/freqtradebot.py
samgermain: > it's REALLY tiring to explain the same thing to you over and over again!):

You have to be more polite. It's discouraging to hear that my work is a burden to you and it makes me want to stop contributing. I read all of your comments and think about them. When you didn't understand funding fees, I used every method I had to explain them to you.

> The main point here is - without fetching positions - output of /wallet is wrong too (we discussed this last time as well)

I understood that the wallet output wouldn't reflect what contracts we were holding, but in this context this didn't seem like a problem, because this method is just to determine if we have a safe exit amount

--------

I do everything I can when reading Github comments, but if you want some tips on how to make explanations ultra clear, I recommend using more markdown syntax for things like

- bullet points
- **bolded words**
- sharing images

xmatthias: I do believe that you read what i write - but it's either a understanding (didn't understand the content / intend) or a forgetting (forgot what you read a few minutes/hours/days later).


What i mean is this discussion: https://github.com/freqtrade/freqtrade/pull/6349#discussion_r798268236, which is a replica of the discussion above, just from 2ish weeks ago.
Using more markdown will not help it, as the **content won't change** (the discussion will still be the same). using bold feels like screaming - which i'm trying to avoid in review comments.


And this again _proves my point_ - i feel like i'm _repeating myself_ over the same exact topic quite often recently - often within the same PR. it's not the work that is a burden, but the "need to tell you x times" - which is simply an attention thing (it seems like you're flying over comments, and then forgetting half of it by the time you're done). 
i'm trying to tell you that it's tiring/frustrating for reviewers when they have to repeat themselves again and again (either within a PR, or across multiple PR's ) on the same exact thing.


Once wallet includes positions, i don't see what would be wrong by including it here, too. 
Removing the TODO-lev feels like a "my goal is to get rid of the TODO-lev's, no matter what" - as it's in fact not done, and should stay there as reminder that we wanted to remove this.

I've tried the case where the position on exchange disappeared / was smaller - and received exceptions from the `create_order` method.
Freqtrade didn't crash, but it was also not properly handled, as we do know beforehand that the position is no longer open - at least once wallets output is correct.

PR #6400, opened 2022-02-13
trade-adjustment for short trades

xmatthias:
## Summary
Improve trade-adjustment support to also support different trade directions (-> short).

This is still lacking leverage support (needs to be tested) - and has therefore some new TODO-lev's.

TODO:
* [x]  check that liquidation-price is properly updated (needs update after a buy fills)

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -726,6 +726,7 @@ class FreqtradeBot(LoggingMixin):
             amount = safe_value_fallback(order, 'filled', 'amount')
             enter_limit_filled_price = safe_value_fallback(order, 'average', 'price')
 
+        # TODO: this might be unnecessary, as we're calling it in update_trade_state.
         interest_rate, isolated_liq = self.leverage_prep(
             leverage=leverage,
             pair=pair,
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -510,7 +510,7 @@ class FreqtradeBot(LoggingMixin):
         """
         # TODO-lev: Check what changes are necessary for DCA in relation to shorts.
         if self.strategy.max_entry_position_adjustment > -1:
-            count_of_buys = trade.nr_of_successful_buys
+            count_of_buys = trade.nr_of_successful_entries
             if count_of_buys > self.strategy.max_entry_position_adjustment:
                 logger.debug(f"Max adjustment entries for {trade.pair} has been reached.")
                 return
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -103,7 +103,6 @@ class FreqtradeBot(LoggingMixin):
         self._exit_lock = Lock()
         LoggingMixin.__init__(self, logger, timeframe_to_seconds(self.strategy.timeframe))
 
-        self.liquidation_buffer = float(self.config.get('liquidation_buffer', '0.05'))
         self.trading_mode: TradingMode = self.config.get('trading_mode', TradingMode.SPOT)
         self.margin_mode_type: Optional[MarginMode] = None
         if 'margin_mode' in self.config:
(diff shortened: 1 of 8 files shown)


freqtrade/persistence/models.py
```diff
@@ -861,6 +861,8 @@ def recalc_trade_from_orders(self):
                 total_stake += tmp_price * tmp_amount
 
         if total_amount > 0:
+            # TODO-lev: This should update leverage as well -
+            # as averaged trades might have different leverage
```
samgermain: Would this work?
```
for o in self.orders:
    total_leverage += o.leverage * o.filled/total_amount
```

xmatthias: I guess ... 
the problem is - `Orders.leverage` is not filled and is always 1 - since ccxt does not return leverage as part of order responses so the leverage parsing always falls back to the column default.
to be clear, I don't think that's a ccxt problem (exchange response doesn't contain leverage either) - but a problem of how we assign the column.

i guess we'll have to simply add the appropriate leverage when creating the order ... 
so it would work here and in tests, but not in reality...

samgermain: >  so it would work here and in tests, but not in reality...

Is it a problem if it doesn't work in reality? That sounds like an issue

xmatthias: > Is it a problem if it doesn't work in reality

you bet :laughing: 
it's also mentioned in #6224 - which points out that leverage in the orders table is always 1 (that's how i discovered it first, and can confirm that all my live databases have leverage=1 in the orders table, too).

so far, that's not a problem as it's a unused field - but with the above change, it wouldn't be, so we should have correct content in the field.

xmatthias: @samgermain thinking about this, a much better question arises - is that even possible?
i buy 100$ worth of "COIN" with 5x leverage, and then again 200$ worth of coin at 3x leverage (so my average leverage would be 4.5).

That's however not a valid leverage value - so i THINK what's going to happen is that the call to "3x leverage" will reduce my initial position to be at 3x leverage - so now i have 300$ at 3x leverage.

samgermain: > That's however not a valid leverage value -

It's valid on some exchanges

> That's however not a valid leverage value - so i THINK what's going to happen is that the call to "3x leverage" will reduce my initial position to be at 3x leverage - so now i have 300$ at 3x leverage.

On binance futures, in isolated mode, you can't reduce your leverage, if you increase a position for $100 at 5x leverage to 10x leverage, you now have a position for $50 at 10x leverage

In cross mode, if you decrease a position for $100 at 20x leverage to 10x leverage, you now have a position for $200 at 10x leverage

---------------------

Binance will also only have one leverage value for all trades with that currency, the same is true for gateio, and okex, but on a few exchanges like kucoinfutures, you can set multiple trades on the same market for different leverages and margin modes.

xmatthias: hm, for the moment, i think we'll then simply use the "existing" leverage of a trade in case of an adjustment.
otherwise this will become too fragmented (obviously it can be changed for kucoin eventually - but i don't see kucoin support at the beginning anyway).

freqtrade/freqtradebot.py
```diff
@@ -1596,9 +1598,19 @@ def update_trade_state(self, trade: Trade, order_id: str, action_order: Dict[str
...
                 trade = self.cancel_stoploss_on_exchange(trade)
+                # TODO: Margin will need to use interest_rate as well.
+                _, isolated_liq = self.leverage_prep(
+                    leverage=trade.leverage,
+                    pair=trade.pair,
+                    amount=trade.amount,
+                    open_rate=trade.open_rate,
+                    is_short=trade.is_short
+                )
+                if isolated_liq:
```
xmatthias, line 1613: @samgermain calling leverage_prep here is the right place (it's when the order filled - so we now know the "real" amount).

The question is though - would this be sufficient overall? (so we could remove the call while placing the order? (where the amount may still change, in case of timedout orders) ?

samgermain: > 

I think the point of the method was to avoid flake8 "This method is too complicated" errors. The error might be gone if I comment out the leverage parts though

---

I don't really understand your question, does that answer it?

xmatthias: well - no.
I'm asking you (and myself as well) if we really need to call this method from 2 places in freqtradebot.

I think calling it here (when the order is effectively filled) is sufficient.
It's not relevant while placing the order (not really, anyway) - only while handling the trade afterwards ...

xmatthias: I'll add a todo there to doublecheck - it's not something we need to check/decide now i think.

PR #6416, opened 2022-02-14
Update windows_installation.md

froggleston:
Update links to include just the cpp build tools instead of the 4GB full Visual Studio link.

## Summary

Update docs to include C++ build tools link for Windows


docs/windows_installation.md
```diff
@@ -54,6 +54,8 @@ error: Microsoft Visual C++ 14.0 is required. Get it with "Microsoft Visual C++
 
 Unfortunately, many packages requiring compilation don't provide a pre-built wheel. It is therefore mandatory to have a C/C++ compiler installed and available for your python environment to use.
 
-The easiest way is to download install Microsoft Visual Studio Community [here](https://visualstudio.microsoft.com/downloads/) and make sure to install "Common Tools for Visual C++" to enable building C code on Windows. Unfortunately, this is a heavy download / dependency (~4Gb) so you might want to consider WSL or [docker compose](docker_quickstart.md) first.
+You can download the Visual C++ build tools from [here](https://visualstudio.microsoft.com/visual-cpp-build-tools/). However, the easiest way is to download install Microsoft Visual Studio Community [here](https://visualstudio.microsoft.com/downloads/) and make sure to install "Common Tools for Visual C++" to enable building C code on Windows. Unfortunately, this is a heavy download / dependency (~4Gb) so you might want to consider WSL or [docker compose](docker_quickstart.md) first.
```
xmatthias: i'm kindof fine with this if we're sure this works.

I assume this is taken from the python output when it's failing - which used to link to a either broken link or to tools that didn't include everything necessary (i think it was the latter).

If we are sure it works, i'd actually change the wording to prefer the build-tools over visual studio community.

xmatthias (approved):
I've updated wording a bit, and specified what specifically needs to be installed (the installer has quite some options)...

LGTM :+1:

PR #6417, opened 2022-02-14
Attempt fix for #6261

### LINKED ISSUE #6261: Buy candle ROI trap
(So we do not forget)

I bumped into a backtesting trap related to roi getting filled on buy candle. Consider this situation:
image

Current behavior:
1. Fist candle requests an order placement below candle close using `custom_entry_price()`.
2. Second candle does reach said level and order is considered filled.
3. Second candle fell by more than active ROI, which means there is enough room for price to go up and get filled by ROI.

ROI (on second candle) essentially considers price to move like so: open -> low (order fill) -> high (roi) -> close. This is too optimistic and hardly ever happens.

How it should work:
If buy (second, in the example above) candle hits ROI target:
* Only consider a successful exit for longs on red candle, if entry price >= open.
* Only consider a successful exit for shorts on green candle, if entry price <= open.

This would always assume the pessimistic outcome:
  * Red candles move open->high(only chance for ROI of longs on entry candle, or try stoploss first for shorts)->low->close
  * Green candles move open->low(only chance for ROI of shorts on entry candle, or try stoploss first for longs) ->high->close

xmatthias:
## Summary
Don't accept 0 duration ROI with custom pricing on red candles with price < open.

closes #6261

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -357,6 +357,15 @@ class Backtesting:
                     # use Open rate if open_rate > calculated sell rate
                     return sell_row[OPEN_IDX]
 
+                if (
+                    trade_dur == 0
+                    and trade.open_rate < sell_row[OPEN_IDX]  # trade-open > open_rate
+                    and sell_row[OPEN_IDX] > sell_row[CLOSE_IDX]  # Red candle
+                ):
+                    # ROI on opening candles with custom pricing can only
+                    # trigger if the entry was at Open or lower.
+                    # details: https: // github.com/freqtrade/freqtrade/issues/6261
+                    raise ValueError("Opening candle ROI on red candles.")
                 # Use the maximum between close_rate and low as we
                 # cannot sell outside of a candle.
                 # Applies when a new ROI setting comes in place and the whole candle is above that.
@@ -414,7 +423,10 @@ class Backtesting:
             trade.close_date = sell_candle_time
 
             trade_dur = int((trade.close_date_utc - trade.open_date_utc).total_seconds() // 60)
-            closerate = self._get_close_rate(sell_row, trade, sell, trade_dur)
+            try:
+                closerate = self._get_close_rate(sell_row, trade, sell, trade_dur)
+            except ValueError:
+                return None
             # call the custom exit price,with default value as previous closerate
             current_profit = trade.calc_profit_ratio(closerate)
             order_type = self.strategy.order_types['sell']
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -359,12 +359,15 @@ class Backtesting:
 
                 if (
                     trade_dur == 0
-                    and trade.open_rate < sell_row[OPEN_IDX]  # trade-open > open_rate
+                    # Red candle (for longs), TODO: green candle (for shorts)
                     and sell_row[OPEN_IDX] > sell_row[CLOSE_IDX]  # Red candle
+                    and trade.open_rate < sell_row[OPEN_IDX]  # trade-open below open_rate
+                    and close_rate > sell_row[CLOSE_IDX]
                 ):
                     # ROI on opening candles with custom pricing can only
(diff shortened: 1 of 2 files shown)


rokups:
I think we can rewrite it better. On red candle guaranteed up-move is in both wicks, so we can safely sell if ROI price is not within candle body. It will cover both upper and lower wick.
```py
if (
    trade_dur == 0
    and sell_row[OPEN_IDX] > sell_row[CLOSE_IDX]  # Red candle (for longs), TODO: green candle (for shorts)
    and sell_row[CLOSE_IDX] > close_rate > sell_row[CLOSE_IDX] # ROI inside candle body
):
  raise ...
```

xmatthias:
> I think we can rewrite it better. On red candle guaranteed up-move is in both wicks, so we can safely sell if ROI price is not within candle body. It will cover both upper and lower wick.
> 
> ```python
> if (
>     trade_dur == 0
>     and sell_row[OPEN_IDX] > sell_row[CLOSE_IDX]  # Red candle (for longs), TODO: green candle (for shorts)
>     and sell_row[CLOSE_IDX] > close_rate > sell_row[CLOSE_IDX] # ROI inside candle body
> ):
>   raise ...
> ```

that doesn't really make sense (neither the description, but less so the code ... )

rokups:
Ok maybe code is bit off 😁 But idea is this: consider a long position and a red candle. In a red candle we can only reliably say that price went up in the candle wicks. For all we know price may go in order of OHLC, so we can fill ROI only in open->high range and low->close range. So we need both trade open and close to be inside wicks to allow considering such a roi sell, or reject it otherwise.

xmatthias:
i think having high== close (or low==close) is quite the exception (it's often very slightly different - which won't tell you anything then). 
We can obviously check for this - but i don't think it'll hit very often (and it's also not really the point of #6261) the way i understood it.

My idea was a different (broader?) one:
when red candle, and "custom-price" pushed price < open price - don't allow immediate sell (as we assume high was before low in this case).

If the candle is full (no wicks) - or has small wicks is not really important i'd think.

rokups:
But then there are candles with large wicks. Imagine a huge pinbar with a tiny red body. ROI on such candle is perfectly viable if we enter on open as wick would be quite big.

xmatthias:
i'm not against "price == open" - only against price < open (if the custom price below the open price).
what if it's a long red candle with a tiny wick of 1 tick both at the top and bottom? your logic wouldn't catch it - but my price close to "low" would still have caused ROI.

rokups:
I think i confused you with a bad example. What i really meant is this:
```py
if (
    trade_dur == 0
    and sell_row[OPEN_IDX] > sell_row[CLOSE_IDX]  # Red candle (for longs), TODO: green candle (for shorts)
    and not (
        (close_rate > sell_row[OPEN_IDX] and trade.open_rate >= sell_row[OPEN_IDX])
        or (close_rate <= sell_row[CLOSE_IDX] and trade.open_rate < sell_row[CLOSE_IDX])
    )
):
  raise ...
```
only allow long roi on buy candle if buy and sell happen in the wicks of red candle. Does that make more sense? 😁

xmatthias:
> only allow long roi on buy candle if buy and sell happen in the wicks of red candle. Does that make more sense?

not really? (i don't see a difference to my code other than checking a few impossible conditions).

i mean - if i buy at/above open on a red candle - it's a given that ROI sell must happen above open (otherwise it's not a ROI sell) - so half of the first OR condition is pointless.
basically the same happens with the other part (if close_rate is below  close, open_Rate on a ROI can NEVER have been higher). 
That makes the condition unnecessarily complicated (and negated at that ...).

In my opinion, it's quite simple - we disallow all sells where open_rate is below Open - and close_rate is above close.
 It's quite a simple condition once you remove all the unnecessary stuff :D

rokups:
I guess you finally saw the difference 😉 Your old code did not allow sells on the bottom wick of the red candle and latest commit fixed that in a way more elegant way than i suggested. awesome 😁👍🏻

PR #6423, opened 2022-02-16
Futures wallet amount

### LINKED ISSUE #6224: Feat/short issues for binance futures

samgermain:
Fetches positions and sets the size of the positions, as the wallet amount for that currency for futures

closes #6224

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1158,50 +1158,28 @@ class Exchange:
             balances.pop("total", None)
             balances.pop("used", None)
 
-            if self.trading_mode == TradingMode.FUTURES:
-
-                open_orders_response: List[dict] = self._api.fetch_open_orders()
-                open_orders: dict = {}
-                for order in open_orders_response:
-                    symbol: str = order['symbol']
-                    open_orders[symbol] = order
-
-                positions: List[dict] = self._api.fetch_positions()
-                for position in positions:
-                    symbol = position['symbol']
-                    market: dict = self.markets[symbol]
-                    size: float = self._contracts_to_amount(symbol, position['contracts'])
-                    side: str = position['side']
-                    if size > 0:
-
-                        if symbol in open_orders:
-                            order = open_orders[symbol]
-                            order_amount: float = order['remaining']
-                            order_side: str = order['side']
-                            if order_side == 'buy' or order_side == 'long':
-                                order_amount = 0
-                        else:
-                            order_amount = 0
-
-                        if side == 'long' or side == 'buy':
-                            currency = market['base']
-                            free = size - order_amount
-
-                            balances[currency] = {
-                                'free': free,
-                                'used': order_amount,
-                                'total': size,
-                            }
-                            balances['free'][currency] = free
-                            balances['used'][currency] = order_amount
-                            balances['total'][currency] = size
-
             return balances
         except ccxt.DDoSProtection as e:
             raise DDosProtection(e) from e
         except (ccxt.NetworkError, ccxt.ExchangeError) as e:
             raise TemporaryError(
                 f'Could not get balance due to {e.__class__.__name__}. Message: {e}') from e
+        except ccxt.BaseError as e:
+            raise OperationalException(e) from e
+
+    @retrier
+    def get_positions(self) -> List[Dict]:
+        if self._config['dry_run'] or self.trading_mode != TradingMode.FUTURES:
+            return []
+        try:
+            positions: List[Dict] = self._api.fetch_positions()
+            self._log_exchange_response('fetch_positions', positions)
+            return positions
+        except ccxt.DDoSProtection as e:
+            raise DDosProtection(e) from e
+        except (ccxt.NetworkError, ccxt.ExchangeError) as e:
+            raise TemporaryError(
+                f'Could not get positions due to {e.__class__.__name__}. Message: {e}') from e
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -342,7 +342,7 @@ class Exchange:
 
     def get_pair_base_currency(self, pair: str) -> str:
         """
-        Return a pair's quote currency
+        Return a pair's base currency
         """
         return self.markets.get(pair, {}).get('base', '')
 
@@ -1157,6 +1157,39 @@ class Exchange:
             balances.pop("free", None)
             balances.pop("total", None)
             balances.pop("used", None)
+
+            if self.trading_mode == TradingMode.FUTURES:
+
+                open_orders_response: List[dict] = self._api.fetch_open_orders()
+                open_orders: dict = {}
+                for order in open_orders_response:
+                    symbol: str = order['symbol']
+                    open_orders[symbol] = order
+
+                positions: List[dict] = self._api.fetch_positions()
+                for position in positions:
+                    symbol = position['symbol']
+                    market: dict = self.markets[symbol]
+                    size: float = self._contracts_to_amount(symbol, position['contracts'])
+                    side: str = position['side']
+                    if size > 0:
+
+                        if symbol in open_orders:
+                            order_amount: float = open_orders[symbol]['remaining']
+                        else:
+                            order_amount = 0
+
+                        if side == 'short':
+                            currency: str = market['quote']
+                        else:
+                            currency = market['base']
+
+                            if currency in balances:
+                                balances[currency] = {
+                                    'free': size - order_amount,
+                                    'used': order_amount,
+                                    'total': size,
+                                }
 
             return balances
         except ccxt.DDoSProtection as e:
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1175,19 +1175,26 @@ class Exchange:
                     if size > 0:
 
                         if symbol in open_orders:
-                            order_amount: float = open_orders[symbol]['remaining']
+                            order = open_orders[symbol]
+                            order_amount: float = order['remaining']
+                            order_side: str = order['side']
+                            if order_side == 'buy' or order_side == 'long':
+                                order_amount = 0
                         else:
                             order_amount = 0
 
                         if side == 'long' or side == 'buy':
                             currency = market['base']
+                            free = size - order_amount
 
-                            if currency in balances:
-                                balances[currency] = {
-                                    'free': size - order_amount,
-                                    'used': order_amount,
-                                    'total': size,
-                                }
+                            balances[currency] = {
+                                'free': free,
+                                'used': order_amount,
+                                'total': size,
+                            }
+                            balances['free'][currency] = free
+                            balances['used'][currency] = order_amount
+                            balances['total'][currency] = size
 
             return balances
         except ccxt.DDoSProtection as e:
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1179,9 +1179,7 @@ class Exchange:
                         else:
                             order_amount = 0
 
-                        if side == 'short':
-                            currency: str = market['quote']
-                        else:
+                        if side == 'long' or side == 'buy':
                             currency = market['base']
 
                             if currency in balances:
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1168,7 +1168,7 @@ class Exchange:
             raise OperationalException(e) from e
 
     @retrier
-    def get_positions(self) -> List[Dict]:
+    def fetch_positions(self) -> List[Dict]:
         if self._config['dry_run'] or self.trading_mode != TradingMode.FUTURES:
             return []
         try:
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -1338,7 +1338,6 @@ class FreqtradeBot(LoggingMixin):
         trade_base_currency = self.exchange.get_pair_base_currency(pair)
         wallet_amount = self.wallets.get_free(trade_base_currency)
         logger.debug(f"{pair} - Wallet: {wallet_amount} - Trade-amount: {amount}")
-        # TODO-lev: Get wallet amount + value of positions
         if wallet_amount >= amount or self.trading_mode == TradingMode.FUTURES:
             # A safe exit amount isn't needed for futures, you can just exit/close the position
             return amount
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -1335,10 +1335,13 @@ class FreqtradeBot(LoggingMixin):
         """
         # Update wallets to ensure amounts tied up in a stoploss is now free!
         self.wallets.update()
+        if self.trading_mode == TradingMode.FUTURES:
+            return amount
+
         trade_base_currency = self.exchange.get_pair_base_currency(pair)
         wallet_amount = self.wallets.get_free(trade_base_currency)
         logger.debug(f"{pair} - Wallet: {wallet_amount} - Trade-amount: {amount}")
-        if wallet_amount >= amount or self.trading_mode == TradingMode.FUTURES:
+        if wallet_amount >= amount:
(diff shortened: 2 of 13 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -1153,6 +1153,37 @@ def get_balances(self) -> dict:
...
+                    size: float = self._contracts_to_amount(symbol, position['contracts'])
+                    side: str = position['side']
+                    if size > 0:
+
+                        if symbol in open_orders:
+                            order_amount: float = open_orders[symbol]['remaining']
+                        else:
+                            order_amount = 0
+
+                        if side == 'long' or side == 'buy':
```
samgermain: Is it fine if we don't do anything for shorts? Shorts mean your selling the base, which means you're getting the quote. This will be returned from the balance api endpoint as used quote (USDT), and we don't really want to say that it's free, because it's locked up in the position.

samgermain:
@xmatthias I just need a thumbs up that this will work and I'll start writing tests for it.

freqtrade/exchange/exchange.py
```diff
@@ -1153,6 +1153,44 @@ def get_balances(self) -> dict:
...
+                            free = size - order_amount
+
+                            balances[currency] = {
+                                'free': free,
+                                'used': order_amount,
+                                'total': size,
+                            }
+                            balances['free'][currency] = free
+                            balances['used'][currency] = order_amount
+                            balances['total'][currency] = size
```
xmatthias: One thing is for sure - we'll not need both ways (we're actually dropping the latter way with "pop()" above, so you'll need to be careful as this will most likely crash as `balances['free']` does no longer exist as dict.
I know ccxt returns it this way - but we only look at the "currency" part (in the end it's the same content ...).

---

i'm not sure i'd do it this way though (basically messing with the currency endpoint/response).

AFAIK, we only use `get_wallets()` in one place (`wallets.py`) - so i'd keep the "position" endpoint separate (from an exchange class perspective), and instead move the logic to `wallets.py` (we'll still need the exchange class, but it'll be the usual thin wrapper around the ccxt method).

My initial idea on this was to add a 4th field (on top of free/used/total) to the Balance object - `position` - which can be positive or negative (that way we can correctly capture shorts too), and which contains the position we're currently in.

Now i'm not sure if that "position" part should contain the currency (like `DOT` in `DOT/USDT:USDT`) - or the whole pair (`DOT/USDT:USDT`) - as in theory, we could (in the future) support expiring futures, where we'd have multiple "DOT" pairs with different expiry - and the position is really a `DOT/USDT:USDT` position, not a DOT position (that's slightly different to spot markets - where you could take the DOT you bought and "run" with them to your offline wallet.

Hope this makes somewhat sense ... 
If you want i can have a look at this topic as well - as I've had a pretty clear idea about this (which will fail once i start to implement it, but we'll see ...)

samgermain: > My initial idea on this was to add a 4th field (on top of free/used/total) to the Balance object - position - which can be positive or negative (that way we can correctly capture shorts too), and which contains the position we're currently in.

I don't understand how this would be used, would the part that needs it just check if there's enough in position as well?

>  as in theory, we could (in the future) support expiring futures, where we'd have multiple "DOT" pairs with different expiry

If they have a different expiry they will have a different symbol, the expiry date is included in delivery future symbols **ex:** `BTC/USDT:USDT-220315`

> If you want i can have a look at this topic as well - as I've had a pretty clear idea about this (which will fail once i start to implement it, but we'll see ...)

If you don't feel swamped with too much other stuff, I feel like I'm running out of ways to be an asset

What pieces do you think are left before the branch "works"

xmatthias: > I don't understand how this would be used, would the part that needs it just check if there's enough in position as well?

Wallets are used for 2 things
* check if we have enough
* Report it via RPC.

There is however quite some difference between "is my position there" versus "do i own enough DOT to sell in spot markets" (one being that we need to use the full symbol for positions - as you point out above).

> I feel like I'm running out of ways to be an asset

The best way to test i feel at this point is "real life tests".
This can be as easy as running a bot with a "nothing" strategy (no signal) - but setting leverage to "whatever" ...
and then use forcebuy to enter a trade (maybe with different order types) - look at the values in freqUI (are values as expected, stoploss, profit calculations, ...)
maybe try a "tight" stoploss (especially on exchange) - so it adjusts frequently, ...

I've done this in a very limited fashion for gate.io and okx (but will need to repeat that once the liquidation branch is ready) - but binance has efficiently blocked my futures access, so i won't be able to do any tests there.

xmatthias:
@samgermain is it intentional / expected that `fetch_positions()` returns inconsistent results?

like - on gate.io - i get historical results (e.g. the one below)
```
{
    "info": {
        "value": "0",
        "leverage": "5",
        "mode": "single",
        "realised_point": "0",
        "contract": "ETH_USDT",
        "entry_price": "0",
        "mark_price": "2896.41",
        "history_point": "0",
        "realised_pnl": "0",
        "close_order": None,
        "size": "0",
        "cross_leverage_limit": "0",
        "pending_orders": "0",
        "adl_ranking": "6",
        "maintenance_rate": "0.005",
        "unrealised_pnl": "0",
        "user": "6266643",
        "leverage_max": "100",
        "history_pnl": "0",
        "risk_limit": "500000",
        "margin": "0",
        "last_close_pnl": "0",
        "liq_price": "0"
    },
    "symbol": "ETH/USDT:USDT",
    "timestamp": None,
    "datetime": None,
    "initialMargin": 0.0,
    "initialMarginPercentage": None,
    "maintenanceMargin": 0.0,
    "maintenanceMarginPercentage": 0.005,
    "entryPrice": 0.0,
    "notional": 0.0,
    "leverage": 5.0,
    "unrealizedPnl": 0.0,
    "contracts": 0.0,
    "contractSize": 0.01,
    "marginRatio": None,
    "liquidationPrice": 0.0,
    "markPrice": 2896.41,
    "collateral": 0.0,
    "marginType": "isolated",
    "side": None,
    "percentage": None
}
```
while on okx i get no results (i currently don't have a position open, so i assume i only get open positions, but closed positions are not included ...).

Also - from the above - how can i determine that this was closed? 
only through notional/contracts being 0.0?

samgermain:
> like - on gate.io - i get historical results (e.g. the one below)

That's how it is on Gate.io because that's what the API returns, I'll see if we should update that, but I'm guessing no because we usually don't want to reduce data from the api

> Also - from the above - how can i determine that this was closed? only through notional/contracts being 0.0?

Yeah that's how you can do that

xmatthias:
I'm just asking because for orders, ccxt has different endpoints for open/closed orders, and i would assume exchanges have that for positions as well (i'm pretty sure i can get "closed_positions" from the okx api, as i can see them in their app, too, but have no way through ccxt it seems (except for using private, non-unified endpoints, obviously)).

While for gate.io that would then go to the same Endpoint (maybe with filters?) - it has the potential to reduce the traffic on other exchanges by only asking for "Open positions" - or "closed positions" (whatever is wanted).

samgermain:
> I'm just asking because for orders, ccxt has different endpoints for open/closed orders, and i would assume exchanges have that for positions as well (i'm pretty sure i can get "closed_positions" from the okx api, as i can see them in their app, too, but have no way through ccxt it seems (except for using private, non-unified endpoints, obviously)).
> 
> While for gate.io that would then go to the same Endpoint (maybe with filters?) - it has the potential to reduce the traffic on other exchanges by only asking for "Open positions" - or "closed positions" (whatever is wanted).

If getting the closed positions is a feature you need I can implement it, it's probably just not implemented yet because we have other things that have been a higher priority

samgermain:
@xmatthias It looks like gateio `fetchPositions` will be left as is, if you need to closed positions for the other exchanges let me know

PR #6424, opened 2022-02-16
partial sell

mkavinkumar1:
## Summary

partial sell for dca
we are selling the last rebuy and subtracting the profit from the stake amount of previous rebuy and calculating its new avg open price
Related issue #1519

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -56,7 +56,7 @@ WEBHOOK_FORMAT_OPTIONS = ['form', 'json', 'raw']
 
 ENV_VAR_PREFIX = 'FREQTRADE__'
 
-NON_OPEN_EXCHANGE_STATES = ('cancelled', 'canceled', 'expired')
+NON_OPEN_EXCHANGE_STATES = ('cancelled', 'canceled', 'closed', 'expired')
 
 # Define decimals per coin for outputs
 # Only used for outputs.
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -56,7 +56,7 @@ WEBHOOK_FORMAT_OPTIONS = ['form', 'json', 'raw']
 
 ENV_VAR_PREFIX = 'FREQTRADE__'
 
-NON_OPEN_EXCHANGE_STATES = ('cancelled', 'canceled', 'closed', 'expired')
+NON_OPEN_EXCHANGE_STATES = ('cancelled', 'canceled', 'expired')
 
 # Define decimals per coin for outputs
 # Only used for outputs.
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -478,6 +478,15 @@ class FreqtradeBot(LoggingMixin):
 
         if stake_amount is not None and stake_amount < 0.0:
             # We should decrease our position
+            #TODO : debug 
+            open_sell_order=trade.select_order('sell', True)
+            if open_sell_order: 
+                msg = {
+                'type': RPCMessageType.WARNING,
+                'status':'bug open_order_id is None'
+            }
+                self.rpc.send_msg(msg)
+                return 
             self.execute_trade_exit(trade, current_rate, sell_reason=SellCheckTuple(
                 sell_type=SellType.CUSTOM_SELL), sub_trade_amt=-stake_amount)
 
@@ -1228,7 +1237,7 @@ class FreqtradeBot(LoggingMixin):
         self.strategy.lock_pair(trade.pair, datetime.now(timezone.utc),
                                 reason='Auto lock')
 
-        self._notify_exit(trade, order_type, sub_trade=bool(sub_trade_amt))
+        self._notify_exit(trade, order_type, sub_trade=bool(sub_trade_amt), order=order)
         # In case of market sell orders the order can be closed immediately
         if order.get('status', 'unknown') in ('closed', 'expired'):
             self.update_trade_state(trade, trade.open_order_id, order,
@@ -1245,9 +1254,9 @@ class FreqtradeBot(LoggingMixin):
         # Use cached rates here - it was updated seconds ago.
         current_rate = self.exchange.get_rate(
             trade.pair, refresh=False, side="sell") if not fill else None
-        if order:
-            profit_rate = safe_value_fallback(order, 'average', 'price')
-            amount = safe_value_fallback(order, 'filled', 'amount')
+        if sub_trade:
+            amount = order.get('filled') or order.get('amount') or 0
+            profit_rate = order.get('average') or order.get('price') or 0
             profit = trade.process_sell_sub_trade(order, is_closed=False)
             open_rate = trade.get_open_rate(profit, profit_rate, amount)
             open_cost=open_rate * amount * (1+ trade.fee_open)
@@ -1377,7 +1386,7 @@ class FreqtradeBot(LoggingMixin):
 
         order = self.handle_order_fee(trade, order)
 
-        trade.update(order, sub_trade=sub_trade)
+        trade.update(order)
         trade.recalc_trade_from_orders()
         Trade.commit()
 
@@ -1390,9 +1399,10 @@ class FreqtradeBot(LoggingMixin):
 
         if not trade.is_open:
             self.handle_protections(trade.pair)
+        sub_trade = order.get('filled') != trade.amount
         if order.get('side', None) == 'sell':
             if send_msg and not stoploss_order and not trade.open_order_id:
-                self._notify_exit(trade, '', True, sub_trade=sub_trade)
+                self._notify_exit(trade, '', True, sub_trade=sub_trade, order=order)
         elif send_msg and not trade.open_order_id:
             # Buy fill
             self._notify_enter(trade, order, fill=True, sub_trade=sub_trade)
@@ -1486,6 +1496,13 @@ class FreqtradeBot(LoggingMixin):
 
         if len(trades) == 0:
             logger.info("Applying fee on amount for %s failed: myTrade-Dict empty found", trade)
+            msg = {
+                'type': RPCMessageType.WARNING,
+                'status': "fees bug"
+
+                    
+            }
+            self.rpc.send_msg(msg)
             return order_amount
         fee_currency = None
         amount = 0
@@ -1543,3 +1560,4 @@ class FreqtradeBot(LoggingMixin):
         return max(
             min(valid_custom_price, max_custom_price_allowed),
             min_custom_price_allowed)
+5
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -1271,7 +1271,7 @@ class FreqtradeBot(LoggingMixin):
             'order_type': order_type,
             'amount': amount,
             'open_rate': open_rate,
-            'close_rate': trade.close_rate,
+            'close_rate': profit_rate,
             'current_rate': current_rate,
             'profit_amount': profit,
             'profit_ratio': profit_ratio,
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -701,7 +701,7 @@ class FreqtradeBot(LoggingMixin):
             'stake_amount': trade.stake_amount,
             'stake_currency': self.config['stake_currency'],
             'fiat_currency': self.config.get('fiat_display_currency', None),
-            'amount': order.get('filled') or order.get('amount')
+            'amount': order.get('filled') or order.get('amount'),
             'open_date': trade.open_date or datetime.utcnow(),
             'current_rate': current_rate,
             'sub_trade': sub_trade,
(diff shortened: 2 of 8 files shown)


xmatthias:
You really shouldn't make pull requests from your own develop branch, but instead create a separate branch.

Usually, you'll want to keep develop updated with upstream (this repository) - so you have the possibility to easily compare for changes offline.
I've seen many messed up PR's/branches because of this - especially if it's one we expect to not be "fix typo and merged" - but to be work in progress.

I'm also not sure if this is a good PR for a first-time contributor - it'll not be as easy as you think it is - and will require quite some understanding of freqtrade internals, so you might be better of looking for some of the issues labeled "good first issue" instead.

mkavinkumar1:
[log.txt](https://github.com/freqtrade/freqtrade/files/8094864/log.txt)

mkavinkumar1:
for testing dca order values
[a.py.txt](https://github.com/freqtrade/freqtrade/files/8102241/a.py.txt)

freqtrade/persistence/models.py
```diff
@@ -116,7 +116,7 @@ class Order(_DECL_BASE):
     ft_order_side = Column(String(25), nullable=False)
     ft_pair = Column(String(25), nullable=False)
     ft_is_open = Column(Boolean, nullable=False, default=True, index=True)
-
+    is_recovered = Column(Boolean, nullable=False, default=False)
```
xmatthias: What do you imagine this variable to be used for?

The naming suggests that you're trying to say "this recovered and is positive again" - please keep in mind that this is ONE usecase for this - but the most likely a lot more profitable one is to increase stake in positions which are going for you - where "is_recovered" is simply a wrong name for.

xmatthias:
To be honest, for the moment, this feels like you're poking a problem with a stick hoping that something good falls out of it, but without really understanding the problem space.

I don't see any concept on how you're going to have profit calculations even remotely correct.
This way, it MAY work somehow for your own purposes (if you don't care if calculations/results are correct), but is nothing we can accept.

Please outline a few points either in a comment below, or by updating the initial post, on
* how you're going to keep track of realized profit
* how you're going to handle colliding signals (you got a buy order open and unfilled, and the strategy would like to place a sell order or viceversa)
* how your approach will handle the following trade sequence: "buy 100, buy 100, sell 100, buy 150, sell 100, sell remainder"

mkavinkumar1:
**we are selling last rebuy as whole**

The realized profit is used to average the previous buy and the last buy is marked as processed.
The strategy should only partial sell when a rebuy is there.
So, instead sell 100, sell remainder, we sell 150.

freqtrade/persistence/models.py
mkavinkumar1: what about is_processed?

stash86:
models.py, line 482, won’t latest buy would be orders[-1]? And orders[-2] would give you exception when the bot have only 1 buy.

And you need to give example of how this would be used on real world

mkavinkumar1:
> models.py, line 482, won’t latest buy would be orders[-1]? And orders[-2] would give you exception when the bot have only 1 buy.
orders[-1] - current buy
orders[-2] - last buy(If 1 rebuy then this will avg price of initial buy )

freqtrade/persistence/models.py
xmatthias: It's not about the name alone - i don't understand what you aim to do with this.

xmatthias:
> The strategy should only partial sell when a rebuy is there.

A functionality like this one will not be accepted. 
For freqtrade, we aim to keep our functionalities open for mutliple different usecases. 
This is especially important in the partial buy/sell space - as there is a multitude of ways to utilize this (take partial profit, DCA, buy more when in profit,  ...).

The way you explain it, it is aimed specifically at your strategy (sell the last bought stake).
This is not generally applicable - and therefore not something we will merge to freqtrade.

A PR should for sure enable this behaviour - but _cannot_ be limited to this functionality.

---

If you aim to have this merged to freqtrade, please answer **all** questions asked in https://github.com/freqtrade/freqtrade/pull/6424#issuecomment-1046041615 

Especially the last part - i expect to be a supported way. If you don't aim to support that, you can close this PR (or i will) - as we will not merge a partial solution which may prevent proper implementation.

mkavinkumar1:
> * how you're going to keep track of realized profit

The realized profit is used to average the previous buys and they are marked as realized unless there is some amount left.

> * how you're going to handle colliding signals (you got a buy order open and unfilled, and the strategy would like to place a sell order or vice versa)

Only one signal should be active like rebuys

> * how your approach will handle the following trade sequence: "buy 100, buy 100, sell 100, buy 150, sell 100, sell remainder"
buy1 -> 100, 
buy2 -> 100
sell1 -> 100

calculates profit from buy2 and buy1 stake amount is adjusted and buy1 avg price is calculated and buy2 is marked as realized

buy3 -> 150,
sell2 -> 100
calculates profit from buy3 and buy3 stake amount is adjusted and buy3 avg price is calculated. 
sell3 -> remaining will sell as usual.

freqtrade/persistence/models.py
```diff
@@ -116,6 +116,7 @@ class Order(_DECL_BASE):
     ft_order_side = Column(String(25), nullable=False)
     ft_pair = Column(String(25), nullable=False)
     ft_is_open = Column(Boolean, nullable=False, default=True, index=True)
+    is_realized = Column(Boolean, nullable=True, default=False)
```
xmatthias: I don't think you can use this to track if an order is realized.

this will fail very quickly (assume the following scenario):
* buy 100
* buy 100
* sell 50
* sell 100
* sell 50

after the first sell, none of the buy orders is fully realized - but yet you realized 50$ (let's ignore the profit for now).

Similarily, you could only enter once, and have multiple exits 
* buy 100
* sell 50
* sell 50

As such, the "realized" status of a buy signal is not relevant.

mkavinkumar1: ->buy1 - 100
->buy2 - 100

after sell1 - 50
->buy1 - 100
->buy2 - 50 & open rate is reduced using the profit

after sell2 - 100
->buy1 - 50 & open rate is reduced using the profit
->buy2 - 0 & marked as realized
->sell3 - as usual (whole sell)

mkavinkumar1: > I don't think you can use this to track if an order is realized.

May be we can set this as is_fully_realized

xmatthias: As mentioned in https://github.com/freqtrade/freqtrade/pull/6424#discussion_r810504496 already - i didn't understand what you're trying to track - and now that i understand it - i think its not going to work in slightly more complex samples.

What difference would it make if something is "partially" sold vs. "fully sold"? 

What you really need to track are 2 things: 
* how much is still owned (what's the position size)
* how much profit was realized (absolute and relative (?))

I think both should be tracked on the trade object, not on the Order object.
When recalculating - you simply do "+" on the buys, and "-" on the sells - probably in the sequence of buys - so you can track realized correctly based on the prior average.

even if after 3 buys and sells, the first buy would be fully realied (but the 2nd isn't)  - you'll still need that first buy to properly calculate the average.
Excluding / ignoring an offer is an edge-case (only when you trade in exact lotsizes) - which is unrealistic for 2 reasons
* lot-size is in coin amount - we're buying and selling in quote amounts (stake currency)
* every value is possible as long as it's tradable (and owned) 

Think about the following: 
we invest 100$ in XRP at a rate of 1$
we invest another 100$ in XRP at a rate of 0.9 (we now own ~211 XRP).

If we now sell 100$ in XRP at a rate of 1.1$ - we've sold 90 XRP - so we hold 121 XRP - so while our size (100$) never changed - we traded 3 different lot-sizes.

mkavinkumar1: > even if after 3 buys and sells, the first buy would be fully realied (but the 2nd isn't) - you'll still need that first buy to properly calculate the average.

We are processing the buys in the reverse order.

mkavinkumar1: Buy 1 : 100 coins @ price 1.00 = $100.00
Buy 2 : 111 coins @ price 0.90 = $99.90

-------------------------

If we sell 111 coins @ 1.1 price = $122.10, then profit is 24.67
Buy 1 : 100 coins @ price 0.75 = $75.33
Buy 2 is fully realized.

-------------------------

If we sell 100 coins @ 1.1 price = $110.00, then profit is 22.22
Buy 1 : 100 coins @ price 1.00 = $100.00
Buy 2 : 11 coins @ price -1.12 = $-12.32

-------------------------

If we sell 121 coins @ 1.1 price = $133.10, then profit is 25.67
Buy 1 : 90 coins @ price 0.71 = $64.33
Buy 2 is fully realized.

-------------------------

[a (2).py.txt](https://github.com/freqtrade/freqtrade/files/8115519/a.2.py.txt)

xmatthias: You will still need to track price/profit properly / correctly independent of the "buy was realized" or not - so this is covering an edge-case that's not really relevant - as you need to run the "full" calculation anyway.

Please also add extensive unit-tests covering the different cases discussed here (and more) - always checking if profit aligns, ...).

you can use for example `test_position_adjust` as a template - as that tests this behavior properly.

freqtrade/freqtradebot.py
```diff
@@ -1238,16 +1238,26 @@ def execute_trade_exit(
...
         current_rate = self.exchange.get_rate(
             trade.pair, refresh=False, side="sell") if not fill else None
-        profit_ratio = trade.calc_profit_ratio(profit_rate)
+        if order:
+            profit_rate = safe_value_fallback(order, 'average', 'price')
+            amount = safe_value_fallback(order, 'filled', 'amount')
+            profit = trade.process_sell_sub_trade(order, is_closed=False)
+            open_rate = trade.get_open_rate(profit, profit_rate, amount)
+            open_cost=open_rate * amount * (1+ trade.fee_open)
+            profit_ratio =  ( open_cost + profit)/open_cost - 1
```
xmatthias: Well let me be very clear - NO.

We will not distribute order-parsing throughout the bot - especially not if it's just for sending notifications.

The values you require here will also be needed by other RPC methods (like /profit) - therefore they must be part of the trade object. The pure fact that you're reading stuff from the order here (even require the order dict here) means that you're doing it wrong.

xmatthias (changes requested):

freqtrade/freqtradebot.py
mkavinkumar1: its to track last buy order

mkavinkumar1: The profit is subtracted from the stake_amount. We are not taking profit unless it is a full sell. We are giving the profit to the trade itself.

xmatthias: > We are not taking profit unless it is a full sell.

the moment the first sell order fills, this is "taken profit". if that's for a full buy order, or 10% of the initial buy is not relevant - as such, i don't understand your statement - and think that the "taken profit" needs to be captured properly in the database (and notifications should then use the data from there).

mkavinkumar1: The profit is used to adjust the position of the trade like when rebought

xmatthias: please
a) make sure to merge develop into your branch. It's currently conflicting which automatically implies there must be changes done.
b) please do not use the order dict in models.py (outside of the Order object). Latest changes in develop use the `Order` object - which gives you type safety. Your code would currently crash/cause problems on certain exchanges (you can't rely on average to be filled correctly on all exchanges).

freqtrade/constants.py
```diff
@@ -56,7 +56,7 @@
 
 ENV_VAR_PREFIX = 'FREQTRADE__'
 
-NON_OPEN_EXCHANGE_STATES = ('cancelled', 'canceled', 'closed', 'expired')
+NON_OPEN_EXCHANGE_STATES = ('cancelled', 'canceled', 'expired')
```
xmatthias: This is introducing a bug, please revert.

Please avoid changes to areas of the bot you don't understand (or without analyzing it's full impact).

xmatthias (changes requested):
If you have any plans on getting this merged, please avoid mixing things that don't belong here (like fixing supposed bugs which you only suspect, or other changes that are not directly related to the feature you're trying to introduce).

**Please note:**
Please see the below as a kind (but last) warning:
Your current attitude is very bad and makes collaboration with you VERY difficult.

I have no problem closing this PR simply because of lack of collaboration, independent of which content it'll try to introduce - as the lack of collaboration is wasting time (you'll run into problems and bugs suggestions told you to not do).

While we are happy to get contributions - they need to be done in a productive, collaborative way - which you're currently not showing (you're trying to force your way without even trying to listen to other arguments).

Please change your attitude / behaviour (and revert changes that are irrelevant to this PR) - **or i'll close this PR**.

mkavinkumar1:
The mess is because I am currently running this branch in live mode.
Will clean up ASAP.

xmatthias:
I do hope that you still have that branch locally in a backup branch ... it's gone from your repository at any rate (which caused the PR to close - without possibility to reopen).

otherwise you might try to use `git reflog` to try and get the changes back (i think you deleted / recreated your repository??).
there's also other ways - but i'd rather not go that route if not necessary.

PR #6427, opened 2022-02-17
moved interest to exchange class method

samgermain:
Is this possible? I don't know how to import these methods into models


freqtrade/exchange/binance.py
```diff
@@ -358,3 +360,23 @@ def dry_run_liquidation_price(
         else:
             raise OperationalException(
                 "Freqtrade only supports isolated futures for leverage trading")
+
+    @classmethod
```
xmatthias, line 364: This won't work as class methods i think.

We're using `exchange_resolver` to get the correct class per exchange(name) - so by using a class-method you'll actually not gain anything (you can't use `Exchange.method()` - but must use `exchangeinstance.method()`.
That said - i'm also not certain if this is something we can have at the moment.

Importing anything from the exchange module to persistence will almost certainly lead to a cyclic import (once you've been there, you'll not want to go back there ... )- so the only way to get this to work is by not mixing `models.py` with exchange.

You also can't/shouldn't attach (or pass) the exchange-object to a trade method - which would be an option i guess - but not a very nice one.

The only thing that jumps to mind right now is to maybe pass the function to the models method that needs to calculate interest (basically `trade.whatever(some,param, exchange.interest)`). - which you can then call (as it's a calllable that's passed in).

It's something that might work - but i'm not sure if we SHOULD be doing this considering where we use it (both `calc_profit` and `calc_profit_ratio` would be affected, and both are used throughout both trading, RPC, backtesting and hyperopt ... so it might be problematic to change it everywhere safely (the safely is the problem here).

PR #6431, opened 2022-02-18
set_margin_mode BadRequest error squelch

samgermain:
Solves the issue with setting the margin mode to what it's already set to

This issue is only fixed in JS on CCXT, I don't know when it will be fixed for python, I would like asssign the error to e, and then check what the message of e is and only squelch it if the message is the specific error message that we don't like, but I don't know how to instantiate an instance of Exchange with api keys for testing and then run the method

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1894,6 +1894,8 @@ class Exchange:
         except (ccxt.NetworkError, ccxt.ExchangeError) as e:
             raise TemporaryError(
                 f'Could not set leverage due to {e.__class__.__name__}. Message: {e}') from e
+        except ccxt.BadRequest:
+            return
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 


freqtrade/exchange/exchange.py
```diff
@@ -1894,6 +1894,8 @@ def _set_leverage(
         except (ccxt.NetworkError, ccxt.ExchangeError) as e:
             raise TemporaryError(
                 f'Could not set leverage due to {e.__class__.__name__}. Message: {e}') from e
+        except ccxt.BadRequest:
+            return
```
xmatthias, line 1898: This error is way to broad considering what we intend to do (hide the error for Binane only).

i think there's a multitude of reasons there could be a "BadRequest" - maybe not on binance, but this would apply to all exchanges.
"hiding" these in general can hide problems we'd encounter for other exchanges.

xmatthias (changes requested):
I noticed that you've got a PR working in https://github.com/ccxt/ccxt/pull/11936 ... 
i'd probably wait for the outcome of that before implementing a "kill all" error handler which could have unintended side-effects.

freqtrade/exchange/exchange.py
samgermain: How can I properly test this method with my api keys so that I can see the structure of the error and make it more specific

xmatthias: Run freqtrade with your API keys in debug mode? 
i mean - it's a simple configuration - and then you just input the regular command as "debug command" in vscode ...

PR #6432, opened 2022-02-18
fixed stake amount

mkavinkumar1:


xmatthias (changes requested):
I'd be grateful if you could explain why you think this change is necessary.

it modifies test inputs, after which test output is changed as well.
I'd like to understand why you think this is necessary / what this improves over the previous test (i'm not in favor of merging stuff without understanding the reason behind it).

mkavinkumar1:
because stake amt is lower than amt*open rate

xmatthias (approved):
i'm sure you'll find many more instances of this throughout the tests - as it's not really testing for "are the numbers correct" - but "is the notification shown correctly".

PR #6435, opened 2022-02-19
Dockerfile to use in rootless environment.

mahiso:
## Summary

Dockerfile to use in rootless environment.

## Quick changelog

- Adds a Dockerfile 

## What's new?

A Dockerfile for use in a Docker rootless environment.

https://www.docker.com/blog/experimenting-with-rootless-docker/

All ftuser specific statements have been removed so that the application runs as root in the container, root is mapped to the user of the host.

This has the particular advantage for mounting the user_data directory from the host, so that the files in it are created with the rights of the host user. So it is easier to manage the user_data from the host and e.g. to keep it in Git.

### DIFF
--- a/docker/Dockerfile.rootless
+++ b/docker/Dockerfile.rootless
@@ -1,0 +1,52 @@
+FROM python:3.9.9-slim-bullseye as base
+
+# Setup env
+ENV LANG C.UTF-8
+ENV LC_ALL C.UTF-8
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONFAULTHANDLER 1
+ENV PATH=/root/.local/bin:$PATH
+ENV FT_APP_ENV="docker"
+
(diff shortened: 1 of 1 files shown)


xmatthias:
All security recommendation recommends to not run docker containers as root, as it's possible to escape the docker container and archive root at the host level.

We used to use root containers - but have since migrated away from that.
Adding this file would therefore be a step back.

Mounted `user_data` directories will have the of the initial host user (UID 1000), therefore i don't see any benefit for this.

Please understand that we will not add another dockerfile we need to maintain to support an _experimental feature_ in docker.
Feel free to resubmit should this become the default docker installation path, or at least no longer be an experimental feature - AND have gained some mainstream adoption (so far, i find this VERY niche, similar to using user namespaces - which has been in docker for years, but almost noone seems to use it).

PR #6440, opened 2022-02-20
Add leverage strategy to new-strategy command

clover-es:
## Summary

Add leverage strategy to new-strategy command.

## Quick changelog

- Modify base template adding new shorts signal.

## What's new?

It's important to note that both 'long' and 'short' calls have been added in this example strategy, 'short' signals are commented out.

TODO:
Test passed, but check if something new is needed.

### DIFF
--- a/freqtrade/templates/base_strategy.py.j2
+++ b/freqtrade/templates/base_strategy.py.j2
@@ -129,13 +129,15 @@ class {{ strategy }}(IStrategy):
                 (dataframe['volume'] > 0)  # Make sure Volume is not 0
             ),
             'enter_long'] = 1
-        # Uncomment to use shorts (remember to set margin in trading mode in configuration)
-        '''dataframe.loc[
+        # Uncomment to use shorts (Only used in futures/margin mode. Check the documentation for more info)
+        """
+        dataframe.loc[
             (
                 {{ sell_trend | indent(16) }}
                 (dataframe['volume'] > 0)  # Make sure Volume is not 0
             ),
-            'enter_short'] = 1'''
+            'enter_short'] = 1
+        """
 
         return dataframe
 
@@ -152,12 +154,14 @@ class {{ strategy }}(IStrategy):
                 (dataframe['volume'] > 0)  # Make sure Volume is not 0
             ),
             'exit_long'] = 1
-        # Uncomment to use shorts (remember to set margin in trading mode in configuration)
-        '''dataframe.loc[
+        # Uncomment to use shorts (Only used in futures/margin mode. Check the documentation for more info)
+        """
+        dataframe.loc[
             (
                 {{ buy_trend | indent(16) }}
                 (dataframe['volume'] > 0)  # Make sure Volume is not 0
             ),
-            'exit_short'] = 1'''
+            'exit_short'] = 1
+        """
         return dataframe
     {{ additional_methods | indent(4) }}
--- a/freqtrade/templates/base_strategy.py.j2
+++ b/freqtrade/templates/base_strategy.py.j2
@@ -129,6 +129,13 @@ class {{ strategy }}(IStrategy):
                 (dataframe['volume'] > 0)  # Make sure Volume is not 0
             ),
             'enter_long'] = 1
+        # Uncomment to use shorts (remember to set margin in trading mode in configuration)
+        '''dataframe.loc[
+            (
+                {{ sell_trend | indent(16) }}
+                (dataframe['volume'] > 0)  # Make sure Volume is not 0
+            ),
+            'enter_short'] = 1'''
 
         return dataframe
 
@@ -145,5 +152,12 @@ class {{ strategy }}(IStrategy):
                 (dataframe['volume'] > 0)  # Make sure Volume is not 0
             ),
             'exit_long'] = 1
+        # Uncomment to use shorts (remember to set margin in trading mode in configuration)
+        '''dataframe.loc[
+            (
+                {{ buy_trend | indent(16) }}
+                (dataframe['volume'] > 0)  # Make sure Volume is not 0
+            ),
+            'exit_short'] = 1'''
         return dataframe
     {{ additional_methods | indent(4) }}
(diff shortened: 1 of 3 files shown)


freqtrade/templates/base_strategy.py.j2
```diff
@@ -37,6 +37,9 @@ class {{ strategy }}(IStrategy):
     # Check the documentation or the Sample strategy to get the latest version.
     INTERFACE_VERSION = 2
 
+    trading_mode: "{{ trading_mode }}"
+    margin_mode: "{{ margin_mode }}"
```
xmatthias: these arguments can't be set from the strategy (and i think that makes sense to keep it this way).

You'll have to have the exchange configured correctly (like - amounts in the futures wallet, API keys enabled for futures, ...) - so you have to have the config set correctly.

freqtrade/templates/subtemplates/sell_trend_full.j2
```diff
@@ -1,3 +1,8 @@
...
-(dataframe['tema'] > dataframe['bb_middleband']) &  # Guard: tema above BB middle
-(dataframe['tema'] < dataframe['tema'].shift(1)) &  # Guard: tema is falling
+dataframe.loc[
+    (
+        (qtpylib.crossed_above(dataframe['rsi'], self.sell_rsi.value)) &  # Signal: RSI crosses above sell_rsi
+        (dataframe['tema'] > dataframe['bb_middleband']) &  # Guard: tema above BB middle
+        (dataframe['tema'] < dataframe['tema'].shift(1)) &  # Guard: tema is falling
+        (dataframe['volume'] > 0)  # Make sure Volume is not 0
+    ),
+'sell'] = 1
```
xmatthias: This should use `exit_long` - `sell` and `buy` should be considered deprecated (although we're currently not warning about it, and will probably continue to support it for quite some time).

Same in the other "full/minimal" templates

freqtrade/commands/deploy_commands.py
```diff
@@ -80,8 +84,10 @@ def start_new_strategy(args: Dict[str, Any]) -> None:
         if new_path.exists():
             raise OperationalException(f"`{new_path}` already exists. "
                                        "Please choose another Strategy Name.")
-
-        deploy_new_strategy(args['strategy'], new_path, args['template'])
+        if "leverage" in args:
+            deploy_new_strategy(args['strategy'], new_path, args['template'], args['leverage'])
+        else:
+            deploy_new_strategy(args['strategy'], new_path, args['template'])
```
xmatthias: you should be able to use `args.get('leverage'], False)` to simplify this ...

freqtrade/commands/deploy_commands.py
```diff
@@ -63,6 +65,8 @@ def deploy_new_strategy(strategy_name: str, strategy_path: Path, subtemplate: st
                                                "sell_trend": sell_trend,
                                                "plot_config": plot_config,
                                                "additional_methods": additional_methods,
+                                               "trading_mode": 'futures' if leverage else 'spot',
+                                               "margin_mode": 'isolated' if leverage else ''
```
xmatthias: i'd not take such assumptions ... (also see other comment about *_mode coming from the strategy - so this may not be necessary ... ).

By assuming that "Leverage == futures" - you're making it more difficult to support margin - which could also use leverage (but cannot rely on this automatism anymore).

freqtrade/commands/cli_options.py
```diff
@@ -463,6 +463,13 @@ def __init__(self, *args, **kwargs):
         choices=['full', 'minimal', 'advanced'],
         default='full',
     ),
+    "leverage": Arg(
```
xmatthias: I'm not sure about using an additional argument for this ...

My initial idea here was to use the `--template` flag, and have it included in the advanced template (just as all the other callbacks).
Do you see any problem with that approach? (maybe we can add a `leverage` mode to that if you think this must be separate).

It's just as likely that a user will be using leverage than that they need to write a custom exit (or most other callbacks for that matter) - so for alignment, we would have to have a flag for every callback (which would be quite bad from a user perspective).

An alternative (if we want to make a customized strategy) would be to "ask" the user interactively (similar to `new-config`) - but i'm not sure if the time investment for that is justified - as it's  just as easily get the full template and delete what's not needed.

clover-es: Thanks for your review. Now that I think about it again, maybe this is as simple as, changing signal  `buy` for `enter_long` in normal templates, and create a new template with only `enter_short` signals instead. I was unaware about the deprecation of `buy` in favour of `enter_long`. But again, you could have both `enter_long` and `enter_short` in a strategy, so how do you solve that using only the `--template` argument?

freqtrade/commands/deploy_commands.py
clover-es: Ok, this can be removed now that I understand that normal strategies can also contain "enter_long" signals.

freqtrade/templates/base_strategy.py.j2
clover-es: Yes, should leave this to the config file.

freqtrade/commands/cli_options.py
xmatthias: is it a problem to have both ways? (maybe the "enter_short" commented out with a comment that this will apply for shorts).

It's anyway intended as a template - not necessarily as something you can use directly - so modifications are kindof expected.

freqtrade/templates/base_strategy.py.j2
```diff
@@ -129,6 +129,13 @@ class {{ strategy }}(IStrategy):
                 (dataframe['volume'] > 0)  # Make sure Volume is not 0
             ),
             'enter_long'] = 1
+        # Uncomment to use shorts (remember to set margin in trading mode in configuration)
```
xmatthias: i'd rather point this to the documentation
margin is not supported (futures is) - but it may require additional configuration - not just the flick of a flag - and having that information in the template directly may cause it to be outdated (maybe we discover something we need to update)...

xmatthias (approved):

PR #6456, opened 2022-02-24
Lev tier update

samgermain:
Sorry to update this again, we split `fetchLeverageTiers` into 2 methods since it was still pretty new, one that can fetch multiple markets, and one where a symbol is required (`fetchMarketLeverageTiers`). I thought it was useful to move okx's implementation to the base exchange class so that it could also be used for `kucoinfutures` (and possibly other exchanges if it comes up)

- All exchanges with `fetchLeverageTiers` also have `fetchMarketLeverageTiers`
- `okx` and `kucoinfutures` have `fetchMarketLeverageTiers` but not `fetchLeverageTiers`

https://github.com/ccxt/ccxt/pull/12090


freqtrade/exchange/exchange.py
```diff
@@ -1859,18 +1858,54 @@ def _get_funding_fees_from_exchange(self, pair: str, since: Union[datetime, int]
...
+                except ccxt.DDoSProtection as e:
+                    raise DDosProtection(e) from e
+                except (ccxt.NetworkError, ccxt.ExchangeError) as e:
+                    raise TemporaryError(
+                        f'Could not load leverage tiers due to {e.__class__.__name__}.'
+                        f'Message: {e}'
+                    ) from e
+                except ccxt.BaseError as e:
+                    raise OperationalException(e) from e
+            elif self.exchange_has('fetchMarketLeverageTiers'):
```
xmatthias, line 1907: I have to say - i do agree with flake8 - this method is too complex and should be split into 2 (it's really 2 methods ...).

I'd create 3 methods out fo this - one that selects the proper method and calls one of the 2 underlying methods for each ccxt method.

samgermain:
`gateio.fetchLeverageTiers` was merged into CCXT

https://github.com/ccxt/ccxt/pull/11948

freqtrade/exchange/exchange.py
samgermain: I took a different approach that made more sense to me, but am willing to change it

xmatthias (approved):

xmatthias:
as all failures are random (and hopefully fixed in feat/short) - i'll merge this anyway as the changes themselfs are ok and do NOT raise errors.

PR #6458, opened 2022-02-25
Hide sell_reason if a trade is still open

stash86:

## Summary

Hide sell_reason if a trade is still open

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?
There are cases where sell_reason isn't empty but a trade is still open, for example when a sell limit order timed out.


freqtrade/rpc/telegram.py
```diff
@@ -448,7 +448,8 @@ def _status(self, update: Update, context: CallbackContext) -> None:
                     "*Current Pair:* {pair}",
                     "*Amount:* `{amount} ({stake_amount} {base_currency})`",
                     "*Entry Tag:* `{buy_tag}`" if r['buy_tag'] else "",
-                    "*Exit Reason:* `{sell_reason}`" if r['sell_reason'] else "",
+                    "*Exit Reason:* `{sell_reason}`"
+                    if (r['sell_reason'] and not r['is_open']) else "",
```
xmatthias: I don't think checking for `is_open` is correct (basically, it'll not have the impact you hope it does).

is_open is only set to False in one place - in `trade.close()` - which is only called once the trade is "really really closed" (when the order is filled). 

after that, there'll be no more checking for cancel and whatnot - so the trade is technically closed - and won't be reopened unless you mess with the database manually.

stash86: So the story was
Someone is on gateio, limit sell order. A sell order was fully canceled, but if you check /status, it would show you the sell reason, because apparently canceled sell order don't remove the sell reason from the db.

So my intention is only show the sell reason once a trade is really really closed. So I believe the change proposed is doing that?

xmatthias: It's not showing it while the sell-order is open (so you'll see "we're selling" - but not why).

Now that's obviously a point of discussion - if that's intended it's fine with me - but from the description - it didn't appear to be - and i think it's good to show the sell-reason for trades with open sell orders.

stash86: Fair, I didn't think of that case. So what would be the best behaviour?

If a sell trade still open, show the reason. If a sell trade closed but the trade still not really really closed, hide the sell reason?

xmatthias: I think the only reason sell_reason should be hidden is if it's set - but the sell-order has been canceled.

A potentially better fix for this might be in the cancel-order section (where all other values are reset, also, but sell_reason is apparently missing) - so it's simply no longer set - which will automatically fix this (and have it correct in the database, too).

xmatthias: I've moved the fix to the "order-cancellation" now so when the order is cancelled, sell_reason is reset.
this is better not just for this notification, but also for freqUI - as well as to keep the database clean/correct.

stash86: thanks

xmatthias (approved):

PR #6466, opened 2022-02-26
Leverage stoploss

samgermain:
Stoploss accounts for leverage

----------------------

A long trade that opens 
- at a price of `1.0` pair/USDT
- **with 5.0 leverage**
- a stoploss percent of 0.05

Would have a stoploss value of 0.99 pair/USDT

-------------------

A long trade that opens 
- at a price of `1.0` pair/USDT
- **with 1.0 leverage**
- a stoploss percent of 0.05

Would have a stoploss value of 0.95 pair/USDT

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -1065,7 +1065,11 @@ class FreqtradeBot(LoggingMixin):
 
         # If enter order is fulfilled but there is no stoploss, we add a stoploss on exchange
         if not stoploss_order:
-            stoploss = self.edge.stoploss(pair=trade.pair) if self.edge else self.strategy.stoploss
+            stoploss = (
+                self.edge.stoploss(pair=trade.pair)
+                if self.edge else
+                self.strategy.stoploss / trade.leverage
+            )
             if trade.is_short:
                 stop_price = trade.open_rate * (1 - stoploss)
             else:
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -1128,9 +1128,6 @@ def test_handle_stoploss_on_exchange(mocker, default_conf_usdt, fee, caplog, is_
     assert trade.is_open is False
     caplog.clear()
 
-    # TODO-lev: Test 2 identical orders but one with leverage, 1 without, and test that the
-    # leveraged trade is hit, but the other trade is not
-
     mocker.patch(
         'freqtrade.exchange.Binance.stoploss',
         side_effect=ExchangeError()
--- a/tests/test_freqtradebot.py
+++ b/tests/test_freqtradebot.py
@@ -1128,6 +1128,9 @@ def test_handle_stoploss_on_exchange(mocker, default_conf_usdt, fee, caplog, is_
     assert trade.is_open is False
     caplog.clear()
 
+    # TODO-lev: Test 2 identical orders but one with leverage, 1 without, and test that the
+    # leveraged trade is hit, but the other trade is not
+
     mocker.patch(
         'freqtrade.exchange.Binance.stoploss',
         side_effect=ExchangeError()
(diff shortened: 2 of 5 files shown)


tests/test_freqtradebot.py
```diff
@@ -1128,6 +1128,9 @@ def test_handle_stoploss_on_exchange(mocker, default_conf_usdt, fee, caplog, is_
     assert trade.is_open is False
     caplog.clear()
 
+    # TODO-lev: Test 2 identical orders but one with leverage, 1 without, and test that the
+    # leveraged trade is hit, but the other trade is not
```
samgermain: Why are we expecting the stoploss to be hit in the code a few lines above this?

xmatthias: it's a stoploss on exchange - so the exchange returns "status=closed"  when asking for the order.

Price didn't have to move - and i'm not sure we must adapt this part for leverage either, as the exchange will still raise the stop when necessary (what we should test is that the value we place it is properly calculated)

freqtrade/freqtradebot.py
```diff
@@ -1065,7 +1065,11 @@ def handle_stoploss_on_exchange(self, trade: Trade) -> bool:
 
         # If enter order is fulfilled but there is no stoploss, we add a stoploss on exchange
         if not stoploss_order:
-            stoploss = self.edge.stoploss(pair=trade.pair) if self.edge else self.strategy.stoploss
+            stoploss = (
+                self.edge.stoploss(pair=trade.pair)
+                if self.edge else
+                self.strategy.stoploss / trade.leverage
+            )
```
samgermain, line 1072: Comment for the link

tests/test_freqtradebot.py
samgermain: I was just trying to test [these changes](https://github.com/freqtrade/freqtrade/pull/6466/files#r815318556), but given what you said I don't think that would be tested here

samgermain:
This failing test keeps showing up in random builds across different PRs 

`FAILED tests/rpc/test_rpc_apiserver.py::test_api_trade_single[True] - assert ...`

It didn't happen locally when running `pytest --random-order-bucket=module --random-order-seed=560592` with python3.9

xmatthias:
> This failing test keeps showing up in random builds across different PRs
> 
> `FAILED tests/rpc/test_rpc_apiserver.py::test_api_trade_single[True] - assert ...`

"should" be fixed in feat/short - it's an odd random failure i've got no clue why it happens.

samgermain:
> > This failing test keeps showing up in random builds across different PRs
> > `FAILED tests/rpc/test_rpc_apiserver.py::test_api_trade_single[True] - assert ...`
> 
> "should" be fixed in feat/short - it's an odd random failure i've got no clue why it happens.

K, I rebased, in case you pulled

samgermain:
> > This failing test keeps showing up in random builds across different PRs
> > `FAILED tests/rpc/test_rpc_apiserver.py::test_api_trade_single[True] - assert ...`
> 
> "should" be fixed in feat/short - it's an odd random failure i've got no clue why it happens.

It looks like it's still happening

freqtrade/freqtradebot.py
xmatthias: which link?
ah nevermind

xmatthias:
> It looks like it's still happening

just need to forcepush harder :rofl: 
seems like i got it now though

freqtrade/freqtradebot.py
xmatthias: i actually think i would've missed this ... :+1:

tests/test_freqtradebot.py
xmatthias: there's a difference - the changes are fine and will change how/where the stoploss will be placed.
That'll not change the "stop on exchange" handling though

xmatthias (approved):
Nice one :+1:

PR #6467, opened 2022-02-26
Liquidation price in backtesting

samgermain:
- backtesting._enter_trade gets liquidation_price and adds it to new trades
- backtesting._leverage_prep created

-------------

I think this might be all that's needed (other than tests), but

- For the conditional `if trade is None:` within `backtesting._enter_trade`, what happens when `trade` is not `None` / whats the situation where this happens
    - Is it fine if we only do stuff with the `liquidation_price` when `trade` is `None`?
- For the conditional `if stake_amount and (not min_stake_amount or stake_amount > min_stake_amount):`, what happens when it's false, I noticed we still return the trade, what is this trade, because it looks like we don't have enough stake to create it.


xmatthias:
> * For the conditional `if trade is None:` within `backtesting._enter_trade`, what happens when `trade` is not `None` / whats the situation where this happens
>   
>   * Is it fine if we only do stuff with the `liquidation_price` when `trade` is `None`?
> * For the conditional `if stake_amount and (not min_stake_amount or stake_amount > min_stake_amount):`, what happens when it's false, I noticed we still return the trade, what is this trade, because it looks like we don't have enough stake to create it.

Both your questions relate to modify_position.

`_enter_trade` is called twice potentially - once to create a new trade (trade is None) - and once to increase the position (trade is a Trade object - therefore we need to return that in case of problems).

The latter point is therefore fine as is - but the first i'm not sure - i assume liquidationprice will change if the position size changes.

freqtrade/optimize/backtesting.py
```diff
@@ -130,6 +130,7 @@ def __init__(self, config: Dict[str, Any]) -> None:
         # TODO-lev: This should come from the configuration setting or better a
         # TODO-lev: combination of config/strategy "use_shorts"(?) and "can_short" from the exchange
         self.trading_mode: TradingMode = config.get('trading_mode', TradingMode.SPOT)
+        self.margin_mode: MarginMode = config.get('trading_mode', MarginMode.NONE)
```
xmatthias: wrong key?

samgermain:
> The latter point is therefore fine as is - but the first i'm not sure - i assume liquidationprice will change if the position size changes.

Yeah that's correct, I updated the code

freqtrade/optimize/backtesting.py
```diff
@@ -682,14 +727,23 @@ def _enter_trade(self, pair: str, row: Tuple, direction: str,
...
+                    orders=[],
                 )
 
             trade.adjust_stop_loss(trade.open_rate, self.strategy.stoploss, initial=True)
 
+            if self.trading_mode == TradingMode.FUTURES:
+                if isolated_liq is None:
+                    raise OperationalException(
+                        f'isolated_liq is none for {pair} while trading futures, '
+                        'this should never happen')
```
samgermain: @xmatthias You mentioned once that you only want to use `assert` in tests, but it might be good to use assert instead of all these conditional checks with `OperationException`s. [I asked mypy about it a while ago and that was their advice](https://github.com/python/mypy/issues/10785)

Using assert implies it should never happen

------------

e.g. `assert isolated_liq is not None`

samgermain:
@xmatthias The tests I added are faiiling with

```
        # leverage = 5
        # ep1(trade.open_rate) = 0.001
        # position(trade.amount) = 60000
        # stake_amount = 300 -> wb = 300 / 5 = 60
        # mmr = 0.01
        # cum_b = 0.01
        # side_1: -1 if is_short else 1
        #
        # Binance, Short
        # ((wb + cum_b) - (side_1 * position * ep1)) / ((position * mmr_b) - (side_1 * position))
        # ((60 + 0.01) - ((-1) * 60000 * 0.001)) / ((60000 * 0.01) - ((-1) * 60000)) = 0.00198036303630
        trade = backtesting._enter_trade(pair, row=row, direction='long')
>       assert isclose(trade.isolated_liq, 0.0019803630363036304)
E       TypeError: must be real number, not NoneType

tests/optimize/test_backtesting.py:582: TypeError
```

I think it's because there's not enough enouhg in the wallet to make an order, but I don't know how to set that amount

freqtrade/optimize/backtesting.py
```diff
@@ -591,6 +592,42 @@ def _get_sell_trade_entry(self, trade: LocalTrade, sell_row: Tuple) -> Optional[
         else:
             return self._get_sell_trade_entry_for_candle(trade, sell_row)
 
+    def _leverage_prep(
```
xmatthias: is this method different to the one in freqtradebot, or just a plain copy?
maybe we have a possibility to extract this to a neutral location if it's identical? (it looks identical at first glance ...)

freqtrade/optimize/backtesting.py
xmatthias: Assert statements can be disabled on a interpreter level with global variables. [how to](https://realpython.com/python-assert-statement/#disabling-assertions-in-production-for-performance).

so when you use asserts - you also assume that it doesn't matter if they are disabled.

I'm not sure about doing these checks this way - you could also simplify this with `if not isolated_liq: set_liquidation()`.

we should be certain enough about our implementation of `_leverage_prep` to be sure the above never happens (Futures + no liquidation returned). 
If we are not - then the implementation needs to be improved.

xmatthias:
> I think it's because there's not enough enouhg in the wallet to make an order, but I don't know how to set that amount

Tests are not that different from regular mode / backtesting - so dry_run_wallet should be the way to go.

freqtrade/optimize/backtesting.py
samgermain: It's identical

samgermain: We could separate it if we pass `exchang` `trading_mode` and `margin_mode` as a parameter

samgermain:
> > I think it's because there's not enough enouhg in the wallet to make an order, but I don't know how to set that amount
> 
> Tests are not that different from regular mode / backtesting - so dry_run_wallet should be the way to go.

Sorry, I don't understand what you mean

tests/optimize/test_backtesting.py
```diff
@@ -562,6 +564,57 @@ def test_backtest__enter_trade(default_conf, fee, mocker) -> None:
...
+    mocker.patch('freqtrade.exchange.Exchange.get_fee', fee)
+    mocker.patch("freqtrade.exchange.Exchange.get_min_pair_stake_amount", return_value=0.00001)
+    mocker.patch("freqtrade.exchange.Exchange.get_max_pair_stake_amount", return_value=float('inf'))
+    patch_exchange(mocker)
+    default_conf_usdt['stake_amount'] = 'unlimited'
+    default_conf_usdt['max_open_trades'] = 2
+    default_conf_usdt['trading_mode'] = 'futures'
+    default_conf_usdt['margin_mode'] = 'isolated'
+    default_conf_usdt['stake_currency'] = 'BUSD'
+    default_conf_usdt['exchange']['pair_whitelist'] = ['.*']
```
xmatthias: I'm not sure about this. It seems that despite having 2 `:USDT` pairs in the list - we don't get them selectable for this test.

I suspect it's because of how `is_future()` works for binance (the exchange this is currently testing) - so the only pair "left" is the SOL/BUSD:BUSD one.

I assume the values are changing intentionally - but i've left it failing for now as i'm not sure what your thought behind the comparisons was.

I've had to split the test in 2 - as we need to initialize Backtesting with proper trading_mode (the error you were seeing had nothing to do with wallet sizes, but with backtesting not being in "futures" mode).

freqtrade/exchange/exchange.py
```diff
@@ -2055,6 +2055,42 @@ def _set_leverage(
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 
+    def get_interest_rate(self) -> float:
+        """
+        Calculate interest rate - necessary for Margin trading.
+        """
+        return 0.0
+
+    def get_liquidation_price(
```
xmatthias, line 2065: I moved (and renamed and simplified) this method to be in the exchange class (which i think is where it belongs).
there's little point in having it return a Tuple if the 2nd value has no meaning. 

It complicates usage beyond reason - especially considering that the interest-rate part is not done (and may even require different arguments once we get to that).

freqtrade/exchange/exchange.py
```diff
@@ -2055,6 +2055,42 @@ def _set_leverage(
...
+            self,
+            pair: str,
+            open_rate: float,
+            amount: float,  # quote currency, includes leverage
+            leverage: float,
+            is_short: bool
+    ) -> Optional[float]:
+
+        if self.trading_mode in (TradingMode.SPOT, TradingMode.MARGIN):
```
samgermain: I think the `TradingMode.MARGIN` should be removed. It's not accurate, we just haven't written the code yet, so the OperationalException is the best outcome from this method

freqtrade/exchange/exchange.py
```diff
@@ -2055,6 +2055,42 @@ def _set_leverage(
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 
+    def get_interest_rate(self) -> float:
+        """
+        Calculate interest rate - necessary for Margin trading.
```
samgermain: Interest rates aren't calculated, they're given by the exchange

xmatthias: well then the comment is slightly wrong.
However, my assumption was that we might need to calculate it for dry-run. 
At any rate, we'll HAVE to cache it for backtesting - otherwise we'd go to the exchange for every trade in backtesting - but only get the same rate back every time.

freqtrade/optimize/backtesting.py
xmatthias: we can also just move it to the exchange class directly :laughing:  it got everything we need and more.

freqtrade/exchange/exchange.py
samgermain: > well then the comment is slightly wrong. However, my assumption was that we might need to calculate it for dry-run. At any rate, we'll HAVE to cache it for backtesting - otherwise we'd go to the exchange for every trade in backtesting - but only get the same rate back every time.

I don't believe it would ever be calculated, I think that it's something that's kind of like price, where you either have it or you don't

tests/optimize/test_backtesting.py
samgermain: > I'm not sure about this. It seems that despite having 2 `:USDT` pairs in the list - we don't get them selectable for this test.

What exactly are you not sure about? What are the :USDT pairs in the list?

> I suspect it's because of how is_future() works for binance (the exchange this is currently testing) - so the only pair "left" is the SOL/BUSD:BUSD one.

You mean `market_is_future`? Which part of it? The part that checks if the type is `swap`/`future` or the part that checks if it's `linear`

> I assume the values are changing intentionally - but i've left it failing for now as i'm not sure what your thought behind the comparisons was.

I don't really know what you're talking about, what values and how are they changing?

When I wrote the test for testing backtesting, I was just trying to check that the `isolated_liq` was being set to what was expected, and I didn't really understand everything else that was going on in the test (because it was already there, and I just wanted to add the liquidation_price part)

xmatthias: The tests are currently failing because "isolated_liq" get's assigned a different value than the expected one.
If you did the math manually and it's now the outcome different - there is something wrong (it seems like you did - the formula is listed right above).
The problem is - the numbers used in the sample (like position) are not correct.

If you looked at the result and just copy/pasted it - the numbers simply need updating - which i don't think is the case.

> What are the :USDT pairs in the list?

The pairs would be ADA/USDT:USDT and ETH/USDT:USDT iirc - but the `market_is_future` of the binance class is only allowing SOL/BUSD:BUSD to pass through.
it's not necessarily a problem - and i don't think we can make them pass both the regular and binance version of "market_is_future" (simply because binance is using "future" instead of swap).

in theory, we could set both future and swap = True (which is not even necessarily wrong) - i'm checking now what impact on other tests this would have - but it's not really a problem to test with BUSD either - as long as it's not a problem throughout tests.

xmatthias: Nevermind the pairs - could please double-check the the formula?
i've now updated all values, but the outcome of the formula in the comment is (Totally) different to the real outcome.
maybe the formula is not for binance - or it's wrong (or implemented wrong) - all i know is that the result is ... different.

samgermain: > Nevermind the pairs - could please double-check the the formula? i've now updated all values, but the outcome of the formula in the comment is (Totally) different to the real outcome. maybe the formula is not for binance - or it's wrong (or implemented wrong) - all i know is that the result is ... different.

The amount was just missing a `0` at the end

xmatthias: so what you're saying is - i've been too stupid to copy a value from the debug window into the editor?

fine ... works for me :laughing: 

except that there's still differences in the 5th decimal (you get 0.000807 - while the calculation yields 0.0008176). While the difference is slim - it might point to a problem (it's at least something to be suspicious about).

samgermain: > except that there's still differences in the 5th decimal (you get 0.000807 - while the calculation yields 0.0008176). While the difference is slim - it might point to a problem (it's at least something to be suspicious about).

I didn't notice that, I'm checking it right now, should hopefully have an answer in 20 minutes

xmatthias: no rush - i also only noticed accidentally - doublechecking if i'm totally mad
your results are quite close though - so it's for sure just a minor thing (my results were WAY off)

samgermain: > no rush - i also only noticed accidentally - doublechecking if i'm totally mad your results are quite close though - so it's for sure just a minor thing (my results were WAY off)

The difference is because of the liquidation buffer, I updated the comments

samgermain: I also added the formula for the freqtrade liquidation price to the docs

xmatthias: Great spot! :+1:

xmatthias (commented):
:+1:

xmatthias (approved):
:+1:

PR #6473, opened 2022-02-27
Slightly reduce verbosity when reload_conf is issued

mkavinkumar1:
old state was compared instead of new

### DIFF
--- a/freqtrade/worker.py
+++ b/freqtrade/worker.py
@@ -86,7 +86,9 @@ class Worker:
         # Log state transition
         if state != old_state:
 
-            if State.RELOAD_CONFIG not in (state, old_state):
+            if state != State.RELOAD_CONFIG or (
+              old_state == State.RELOAD_CONFIG and
+              state != State.RUNNING):
                 self.freqtrade.notify_status(f'{state.name.lower()}')
 
             logger.info(
--- a/freqtrade/worker.py
+++ b/freqtrade/worker.py
@@ -86,7 +86,7 @@ class Worker:
         # Log state transition
         if state != old_state:
 
-            if state != State.RELOAD_CONFIG:
+            if State.RELOAD_CONFIG not in (state, old_state):
                 self.freqtrade.notify_status(f'{state.name.lower()}')
 
             logger.info(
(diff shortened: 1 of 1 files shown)


freqtrade/worker.py
```diff
@@ -86,7 +86,7 @@ def _worker(self, old_state: Optional[State]) -> State:
         # Log state transition
         if state != old_state:
 
-            if old_state != State.RELOAD_CONFIG:
+            if state != State.RELOAD_CONFIG:
```
xmatthias: I'm not sure i understand the point of this (what this tries to accomplish).

What this changes is when this is executed (before/after the reload). 
Please explain which messages / in which scenario you think this makes a difference.

mkavinkumar1:
image
The second msg is not necessary

xmatthias (changes requested):
It's incorrect that one of these messages is not needed.

All 3 messages show one particular thing:
1. we received your request to reload the configuration (this is a telegram response - not a state change) - and will not show if you reload via API.
2. we are reloading the configuration (so the bot will now be down for some time)
3. we are now running again.

The mid-state was okish (it changed the wording of the 3rd messages) - but removing this message is not desired, as it hides useful information.

The time between the 2nd and 3rd point can vary greatly depending on the exchange (some exchanges take some time to start up) - so while it might be fast on the configuration you're attempting this - it's not given that that's the case.


image

You're now also hiding the fact of what the new state is. It could be running (default config) - but it could also be stopped (if startup-state is stopped in the configuration).

mkavinkumar1:
> You're now also hiding the fact of what the new state is. It could be running (default config) - but it could also be stopped (if startup-state is stopped in the configuration).

the new state is already ignored.

xmatthias:
> the new state is already ignored.

Which is why i said - the "middle ground" you had was better :)

mkavinkumar1:
> The time between the 2nd and 3rd point can vary greatly depending on the exchange

But the time between 1st and 2nd is instant only. Will the 2nd one fail after 1st?

xmatthias:
> But the time between 1st and 2nd is instant only.

No it's not - it'll depend on the exchange and configuration / setup. 

As mentioned above - the wording can change (which was the first commit) - but the remainder is not desired (so there will be 2 messages - one for "reloading" (start) - and one for "reloaded/running" (end)) - also when reloading via API (which will remove the very first message, which is a response).

The current instance moves further away from that - and is therefore an unwanted change.

PR #6481, opened 2022-02-28
partial sell

mkavinkumar1:
Previous PR #6424 
updating failed unit tests


freqtrade/freqtradebot.py
```diff
@@ -692,23 +701,25 @@ def _notify_enter(self, trade: Trade, order: Dict, order_type: Optional[str] = N
...
+            'exchange': trade.exchange.capitalize(),
             'pair': trade.pair,
             'limit': open_rate,  # Deprecated (?)
             'open_rate': open_rate,
             'order_type': order_type,
             'stake_amount': trade.stake_amount,
             'stake_currency': self.config['stake_currency'],
             'fiat_currency': self.config.get('fiat_display_currency', None),
-            'amount': safe_value_fallback(order, 'filled', 'amount') or trade.amount,
+            'amount': order.get('filled') if fill else order.get('amount'),
```
xmatthias: Why would you change this? 
the prior line was the "safe" version - which made sure a value was sent.

This will fail if filled = None in the order dict - as you'd pass None to the object.

freqtrade/freqtradebot.py
```diff
@@ -1192,7 +1205,7 @@ def execute_trade_exit(
             # Emergency sells (default to market!)
             order_type = self.strategy.order_types.get("emergencysell", "market")
 
-        amount = self._safe_exit_amount(trade.pair, trade.amount)
+        amount = sub_trade_amt or self._safe_exit_amount(trade.pair, trade.amount)
```
xmatthias: you can do this IF you make sure that sub_trade_amt is really lower than trade amount (which i currently don't see anywhere).

you're also mixing stake amount with base amount (the return value of `adjust_trade_position` is in stake/quote amount).

freqtrade/freqtradebot.py
```diff
@@ -1227,24 +1240,41 @@ def execute_trade_exit(
         self.strategy.lock_pair(trade.pair, datetime.now(timezone.utc),
                                 reason='Auto lock')
 
-        self._notify_exit(trade, order_type)
+        self._notify_exit(trade, order_type, sub_trade=bool(sub_trade_amt), order=order)
```
xmatthias: There is no reason to use the order dict here.
please use `order_obj` to get type safety and avoid nasty bugs (ccxt/exchange order return values can be "pretty" empty depending on the exchange and endpoint...).

freqtrade/persistence/models.py
```diff
@@ -636,24 +705,17 @@ def calc_profit_ratio(self, rate: Optional[float] = None,
...
-            # Just in case, still recalc open trade value
-            self.recalc_open_trade_value()
-            return
-
         total_amount = 0.0
         total_stake = 0.0
         for o in self.orders:
             if (o.ft_is_open or
                     (o.ft_order_side != 'buy') or
+                    o.is_fully_realized or
```
xmatthias: I've said it before, and will say it again.
tracking `is_fully_realized` is pointless. 

you need to track 2 things - realized profit (profit you already sold) - and open profit (profit/loss that's still open).
Especially when recalculating - you need to look at both and update both - as both realized, as well as non-realized will have changed.

By ignoring realized orders, you're potentially skewing / falsifying the result of subsequent orders - assuming the exit size was not identical to the entry size.

The method of tracking "is_fully_realized" will ONLY work if you always trade identical sizes (you enter in steps of 100, exit in steps of 100) - the moment you mix and match values (enter 100, 100, 200,  exit 120) , this will result in wrong results / calculations.

As such - this is covering an **edge-case** (without properly checking this is said edge-case).

freqtrade/persistence/models.py
```diff
@@ -119,6 +119,7 @@ class Order(_DECL_BASE):
     ft_order_side: str = Column(String(25), nullable=False)
     ft_pair: str = Column(String(25), nullable=False)
     ft_is_open = Column(Boolean, nullable=False, default=True, index=True)
+    is_fully_realized = Column(Boolean, nullable=True, default=False)
```
xmatthias: This property is pointless and tracks an irrelevant fact, please remove.

You will need additional columns to track realized profit (for ease of access from RPC methods) - but **not** on the Order object.

xmatthias (changes requested):

Mohbad123:
ATM pullout

Mohbad123 (commented):
I need atm pullout

freqtrade/freqtradebot.py
mkavinkumar1: Updated

freqtrade/freqtradebot.py
mkavinkumar1: Updated.

freqtrade/freqtradebot.py
mkavinkumar1: Updated

freqtrade/persistence/models.py
mkavinkumar1: It was adjusted to different lots.

freqtrade/persistence/models.py
mkavinkumar1: The profit is used to adjust open rate like when we increase stake.

freqtrade/freqtradebot.py
```diff
@@ -478,9 +478,22 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
...
+                    'status': 'bug open_order_id is None'
+                }
+                self.rpc.send_msg(msg)
+                return
+            # Starategy should return amount in base currency to avoid dust
+            amount = -stake_amount
+            if amount > trade.amount:
+                logger.info("Amount is higher than available.")
+                return
+            self.execute_trade_exit(trade, current_rate, sell_reason=SellCheckTuple(
```
xmatthias: Strategy will not return this in base currency - but in stake currency.

Doing it the way you propose would mean that to increase values - the same method (`adjust_trade_position`) would have to return:
* positive values **in stake currency** to increase amount
* negative values **in base currency** to reduce positions

That's a terrible, terrible interface, and very prone to user mistakes - and as such, not an interface i'm willing to accept into freqtrade.

it's freqtrade's (and as PR creator, your) responsibility to convert the stake-amount to the corresponding base amount at the current sell rate.

You can avoid dust by returning slightly more than what you own.

Once the "remaining stake" get's to below minimum tradable amount (usually around 10$ - but it can also be 50$ on some exchanges - use `get_min_pair_stake_amount()` to determine this value)) - you'll have to either reject the partial sell (with "remaining would be too small") - or sell "the remaining bit".

mkavinkumar1: Updated

freqtrade/persistence/models.py
mkavinkumar1: Added a unit test for that.

mkavinkumar1: added two more unit tests for profit calculation

Profit calculation checker
[aa.py.txt](https://github.com/freqtrade/freqtrade/files/8179725/aa.py.txt)

PhaniKanagala (approved):

freqtrade/rpc/telegram.py
```diff
@@ -235,17 +235,31 @@ def _format_buy_msg(self, msg: Dict[str, Any]) -> str:
...
+        message += f"*Total:* `({round_coin_value(total, msg['stake_currency'])}"
 
         if msg.get('fiat_currency', None):
-            message += f", {round_coin_value(msg['stake_amount_fiat'], msg['fiat_currency'])}"
+            message += f", {round_coin_value(total_fiat, msg['fiat_currency'])}"
 
         message += ")`"
+        if msg.get('sub_trade'):
+            bal = round_coin_value(msg['stake_amount'], msg['stake_currency'])
+            message += f"\n*Balance:* `({bal}"
```
xmatthias, line 257: I don't think the wording of "Balance" is correct in this context.
What i assume you mean is "Remaining" (i sold 100XXX, and still own 150 XXX)?

especially in the case of a partial sell though - I'd like to see 2 things in the notification

* Realized profit
* unrealized / still open profit

Both of the above would also make sense in the `/status` context / message.

freqtrade/optimize/backtesting.py.dev
```diff
@@ -0,0 +1,991 @@
+# pragma pylint: disable=missing-docstring, W0212, too-many-arguments
```
xmatthias: I assume this file should either be removed or renamed to `backtesting.py` (if it contains the changes needed for backtesting support of this).

it's a good reminder though that backtesting support for partial sells seems to be completely absent at the moment (this file will not be used by freqtrade).

freqtrade/rpc/telegram.py
mkavinkumar1: As we used the profit to average the price of previous buy, realized profit can't be calculated for subsequent sells.

For example,
1   buy 100 @ 11
2   sell 50 @ 8
3   sell 50 @ 16

What is the realized and unrealized profit after order 2?
What is the realized profit after order 3?

The final profit is 94.25

mkavinkumar1: As we used the profit to average the price of previous buy, realized profit can't be calculated for subsequent sells.

For example without fees,
1   buy 2 @ 10
2   buy 2 @ 5
3  sell 2 @ 6
4   sell 2 @ 9

What is the realized and unrealized profit after order 2 and 3?
What is the realized profit after order 4?

The final profit is 0

mkavinkumar1: For example without fees,
1   buy 2 @ 10
2   buy 2 @ 5
3  sell 2 @ 6
4   sell 2 @ 9

What is the realized and unrealized profit after order 2 and 3?
What is the realized profit after order 4?

The final profit is 0

freqtrade/optimize/backtesting.py.dev
mkavinkumar1: For example without fees,
1   buy 2 @ 10
2   buy 2 @ 5
3  sell 2 @ 6
4   sell 2 @ 9

What is the realized and unrealized profit after order 2 and 3?
What is the realized profit after order 4?

The final profit is 0

PhaniKanagala (approved):

freqtrade/rpc/telegram.py
xmatthias: well after order 2 it's easy - realized profit is 0 (you didn't sell anything - so i don't think this should even be shown) and unrealized is whatever the result is (something negative, i suppose).

after 3 - your maximum position was 30$ - you sold half of it for a total of 18$ (your average entry price was 7.5$ - your sell was at 6$/pce - so your profit will be ~-3$ (and whatever this results to in relative terms) <- this is the realized profit
You'll still have some unrealized profit - which will be the remaining position vs. current price.

after 4 - realized profit will be "whatever the outcome" (so 0 in this case) - and there's no unrealized profit (you're no longer in a position).

---

I think you're too focused on this one usecase though (DCA with full lotsizes only) - what for the following scenario (increasing position size because it's going "our way")

1 buy 5 @ 10$
2 buy 5 @ 12$
3 sell 2 @ 15$
4 buy 5 @ 13$
5 sell remaining 13 @17$

After order 3 - you'll have some realized/materialized profit (you've taken profit - so you MUST keep track of how much you earned in this trade already).
You then rebuy on a small pullback - before selling for a substancial gain.

mkavinkumar1: 1 buy 5 @ 28$
2 buy 5 @ 10$
3 sell 2 @ 11$
4 buy 5 @ 13$
5 sell remaining 13 @17$

Here after order 3, I calculated profit from last order price($10) instead of avg price($12). So the profit is $2 instead of -$2 and subtracted it from the stake amount of the previous orders to reduce the avg price. If like order 2 and 3 happens multiples times, the stake amount will be reduced. Also it reduces the chance of hitting stoploss as we used the profit instead of just tracking.

mkavinkumar1: Used quote amount as 1 for the sake of example.

1 buy 1 @ $10
2 buy 1 @ $5
3 sell 1 @ $8



Here, I calculated profit from last order price($5) instead of avg price($7.5). So the profit is $3 instead of $.5 and subtracted it from the stake amount of the previous orders and calculated new avg price.

So the new avg price is $7 instead of $7.5.

If the stoploss sets for -6% after 1 hour, and the price comes to $7 , the trade will be saved by the new avg price.

mkavinkumar1: > your sell was at 6$/pce - so your profit will be ~-3

Here I considered the profit is $2.
In user perspective, we are selling it in profit as we bought at lower price previously.

mkavinkumar1: > I'd like to see 2 things in the notification
> 
> * Realized profit
> * unrealized / still open profit
> 
> Both of the above would also make sense in the `/status` context / message.

So at every partial sell, the profit includes the previous profits. As we subtract the profit from stake amount, the realized profit reduces the unrealized profit.

Same happens if we rebuy at higher prices.

mkavinkumar1: > your sell was at 6$/pce - so your profit will be ~-3

Here I considered the profit is $2.
In user perspective, we are selling it in profit as we bought at lower price previously.

mkavinkumar1: > I'd like to see 2 things in the notification
> 
> * Realized profit
> * unrealized / still open profit
> 
> Both of the above would also make sense in the `/status` context / message.

So at every partial sell, the profit includes the previous profits. As we subtract the profit from stake amount, the realized profit reduces the unrealized profit.

Same happens if we rebuy at higher prices and selling in different lot sizes.


This way it will be easier for strategies to process the trades else the strategy will need to process the realized profits and stoploss and roi won't be reflect realized profits

mkavinkumar1: Maybe we can sort the orders list by prices

freqtrade/optimize/backtesting.py.dev
mkavinkumar1: Updated

freqtrade/rpc/telegram.py
xmatthias: > Used quote amount as 1 for the sake of example.

By simplying stake to 1 - you're ignoring the complexity of different stake sizes. While equal size is a possibility we should be prepared for, please assume this as an exceptional case - not "what we base our calculations on".


> Here, I calculated profit from last order price($5) instead of avg price($7.5). So the profit is $3 instead of $.5 and subtracted it from the stake amount of the previous orders and calculated new avg price.

I don't think this is correct behaviour, and this will fail once you use different stake sizes per order (how would this behave if the "last" order is smaller than the amount sold?).

You should instead use the (weighted) average price of all orders so far - obviously subtracting already sold amounts (so the average open price may change as more amount is sold).

> Maybe we can sort the orders list by prices

Orders should be sorted by date - never by price.


> So at every partial sell, the profit includes the previous profits. As we subtract the profit from stake amount, the realized profit reduces the unrealized profit.

It's a point of discussion if the regular profit column should contain all profit (realized + unrealized) - and one i would leave up to you for now (assuming it's properly outlined in the documentation).

I'd however still want to see "realized" profit as a separate datapoint  - as this corresponds to "locked in" profit - which is an important point (if this trade goes completely against me - how much profit did i already lock in).

mkavinkumar1: > how would this behave if the "last" order is smaller than the amount sold?

It will take the previous order price for the remaining amount.

mkavinkumar1: > so your profit will be ~-3$

Shouldn't it be subtracted from the stake amount?
So that the ROI and stoploss will be adjusted.

freqtrade/persistence/models.py
mkavinkumar1: Done

freqtrade/rpc/telegram.py
mkavinkumar1: Done

freqtrade/persistence/models.py
```diff
@@ -179,6 +180,7 @@ def update_from_ccxt_object(self, order):
         self.amount = order.get('amount', self.amount)
         self.filled = order.get('filled', self.filled)
         self.average = order.get('average', self.average)
+        self.initial_average = order.get('average', self.initial_average)
```
xmatthias, line 183: This is on the order object, which corresponds to one exchange order.

The average of the order will not change - so this seems like a pointless column.

freqtrade/persistence/models.py
```diff
@@ -504,6 +514,64 @@ def update_trade(self, order: Order) -> None:
...
+            if is_closed:
+                b_order.filled -= amount
+                b_order.order_update_date = datetime.now(timezone.utc)
+            sell_amount -= amount
+            profit += self.calc_profit2(avg_rate, sell_rate, amount)
+            realized_profit += self.calc_profit2(buy_rate, sell_rate, amount)
+        if is_closed:
+            b_order2 = orders[idx]
+            amount2 = b_order2.safe_amount_after_fee
+            b_order2.average = (b_order2.average * amount2 - profit / (1 + self.fee_open)) / amount2
```
xmatthias, line 551: the average corresponds to the **order's average** - how it filled on the exchange - and the same is true for the whole order object.
Therefore this is simply wrong - the average fill price of the order will **NEVER** change based on other orders.

This is changing it in the database (not sure if that's intentional ...) - so running this same method twice in a row will get you different results

freqtrade/persistence/models.py
```diff
@@ -504,6 +514,64 @@ def update_trade(self, order: Order) -> None:
...
+            b_order2.order_update_date = datetime.now(timezone.utc)
```
xmatthias: This HAS to correspond to the date when the order was updated with a ccxt order response - not when you messed with it.
This is true for almost all values of the order object (that are assigned as part of the parse_ccxt method).

Anything else will be considered a bug.

tests/test_freqtradebot.py
```diff
@@ -4581,6 +4590,318 @@ def make_sure_its_651(*args, **kwargs):
...
+        ),
+        (
+            ('buy', 100, 3),
+            ('buy', 100, 7),
+            ('sell', 100, 11),
+            ('buy', 150, 15),
+            ('sell', 100, 19),
+            ('sell', 150, 23),
+        )
+    ]
```
xmatthias: I do like this concept - i'd however

* move this to a parameter (`@pytest.mark.parametrize`) so it's separate tests, which simplifies debugging should one case fail
* Add several "expected" values for each order (may also be `None` if it doesn't make sense) to check the value after each order
*  Evaluate and therefore double-check the trade's (amount, profit, realized profit, unrealized profit, ...) after each step (that's the expected values from above).

xmatthias (changes requested):
Your handling / updating of order objects will introduce quite severe bugs (i commented on **some** of the affected lines, but there's others doing this stuff as well).

please do not update **anything** on the order object. It's there to reflect the order-status of the exchange.
for example, updating the average of the order would mean that you now assume that the fill price of the order changed - which is simply not true, and therefore a bug.
It'll also have unintended side-effects you're clearly not aware - like that orders will no longer show up correctly (as they were made) in neither telegram nor freqUI - making it impossible to verify if the calculations done were correct or not, as the underlying data-base did change.

---

Honestly - this feels like a PR with very little concept on how things should work, and as an attempt to just poke at it "until some valid solution falls out".

Some parts of the code show that you didn't properly test stuff.
That's something i really hate in PR's - as it's pushing the testing to the Reviewer / or end user - which in a project this size, will simply not work - and therefore we have to be strict in reviews.
It will also slow down all subsequent reviews - as it's clear that the changes made can't be trusted - which means a way more meticulous - and therefore more timeconsuming review.

---

Let me repeat the question i asked in the previous PR - are you really sure that this is a PR you should be working on based on your experience in programming/development, working with freqtrade and trading?

It's a pretty complex PR (we knew that from the start, and i warned you right away).
The direction seems pretty wrong in many instances - which shows quite clear non-understanding of how freqtrade works.

Let me suggest this again: maybe start with another, smaller PR (or a few) - to get a better understanding of how freqtrade works first.

xmatthias:
maybe consider approaching the calculation topic from the other side (full trade closed).
then you'd want to
* (weighted) average all buys
* (weighted) average all sells
which makes out the "final" profit. 
This is the approach now, too (except there's only one sell) - and i don't expect this "final" logic to change, either.
Now if we move back from this - and have - say - 10% (100X) still in the trade (we sold 90% or 900X) - then i'd assume it's the same approach for the closed profit.

for the "still open" amount - i don't have a very strong opinion yet, really - it can be either "what's left" - or "all assuming the remainder sells at current rate". 
Now thinking about it - i think this last approach is the best - as you'll see "realized profit: 1% (10$) - unrealized  profit 2% (20$)".
so if the trade closes "right now", the final profit is aligned to the last shown unrealized profit of 2%.

mkavinkumar1:
For future readers, the approach considered was

When we partial sell, the profit is calculated from the price of last buys and it is used to adjust the price of the previous buys. The filled amounts are reduced from the last buys. This way ROI and stoploss will acknowledge the profits.

tests/test_freqtradebot.py
mkavinkumar1: Updated

freqtrade/persistence/models.py
mkavinkumar1: Updated

freqtrade/persistence/models.py
mkavinkumar1: It is changed when profit is given to the previous orders.

To calculate realised profit of the trade, a copy of it is taken.

freqtrade/persistence/models.py
mkavinkumar1: It is changed when profit is given to the previous orders.

This statements wil only run once when the sell order is closed.

xmatthias: Let me repeat what i said above (and in the change statement).

The order object is not to be updated - it's there to reflect the state of the order on the exchange.

It does therefore not matter when or how often this code is running - it's modifying fields in the database it should not.

the **order's average price did not change** - and it never will - as it's an Order on the exchange.

This is not something to argue about - it's a simple fact and i will not accept anything that will change this.

xmatthias (changes requested):
If i'm honest - i'm getting really frustrated with this PR (or rather, with the behaviour/attitude you're showing).

I think i've been very clear above about the fact that the **order object is not to be updated**, yet you ignore that and simply push on as if i didn't say anything, reasoning with "it's only called once".

While i'm happy to get contributions - i also expect that review comments are followed - especially on points that completely violate the point of a database object.

---

As you don't seem to understand how contributions to open source works let me explain one thing:

As a maintainer, accepting a Pull request means that i'm accepting responsibility to maintain this code going forward.
This means, if pull requests try to change things we clearly don't want changed (for example changes to the order object's values) - we will request changes.

Not following this will lead to the rejection of the PR - and repeated ignoring of comments and requests for change will lead to the closure of the PR to avoid wasting our time on contributions that would create problems to maintainability in the future.

---

To be clear on further process so it's no surprise:

Should you request another review without having removed all code that modifies the order object "at will" (so it's no longer reflecting the values on the exchange) - i will understand this as "i ignore what reviewers say and how the project works - and try to push my way through" - which will lead to the closure of the PR.
Should you reopen the same PR without the requested changes - it will lead to a ban from the freqtrade organization.

I really hope it does come to neither of the above - but i will not waste any more time reviewing things which clearly introduce problems.

mkavinkumar1:
For a trade with stake amount $100 left after partial sell with profit $5, and the ROI is 10% , then the final profit is $15 greater than expected $10. Do you have any reason why the partial sell profit should not be subtracted from the stake amount?

xmatthias:
> For a trade with stake amount $100 left after partial sell with profit $5, and the ROI is 10% , then the final profit is $15 greater than expected $10. Do you have any reason why the partial sell profit should not be subtracted from the stake amount?

For the remaining stake - you'll "fake" an average entry price that's below the lowest "real" entry.

This is wrong and will cause problems and side-effects you're clearly not aware of - like sudden ROI triggers as the remaining relative profit "spiked" for no good reason (which means 2 sells in very short succession - so why sell only part of the position in the first place?) - and potentially also due to side-effects with trailing stop (although i didn't do the math to verify problems there).
It'll also create a false sense of profit (it might show a remaining profit of 50% - while in reality we only gained 10% on the whole position) - leading to confusion and disappointment.

We've discussed this already, both here and on discord.
To me, the approach you're taking with the calculation is simply wrong - and i've seen no good reason NOT to use the overall average (The only reason you brought forward so far was to show the "profit spike" - which is however missleading - and we will therefore not use).

Please accept my **decision** that we will use the real average of all entries, so the "average entry" will only be modified by additional entries - but not by exits.

mkavinkumar1:
Let's take the BTC example
Buy 10  @  $ 50k
Buy 2 @  $ 30k 
Sell 1  @ $ 33k

If the last 2 orders happens for 10 or more  times, and if we use the average price, then the partial sells will be at huge loss. Shouldn't the loss be taken into account for ROI?

xmatthias:
> Let's take the BTC example
> Buy 10 @ $ 50k
> Buy 2 @ $ 30k
> Sell 1 @ $ 33k


After the 2nd buy - you own 12 BTC - bought at an average rate of 46_667$.

Your sell for 1 BTC will therefore be at a loss from 46_667 to 33_000 - or 13_000$ -  or -29% (`(33_000 * 1) / (46_667 * 1) -1`).
From a calculation perspective, the open average will not change (why would it - it's all one bucket of BTC's bought at an average price of 46_667$. you can't determine if you sold the bitcoin you bought at 30k or 50k - so both decisions would be wrong).

if you sell another 1 BTC for the same price (again 33k$) - this will be again for a loss of -29% - or 13_000$.

If you sold the remainder at 33k$ as well (so you sold all 12 BTC at 33k$), then your overall loss will also be -29% - or `((33_000 * 12) / (46_667 * 12)) - 1` - or ~-29%.

The same behaviour is expected when selling 10x 1 BTC at 33_000$ - which again will result to ~29% loss (both per exit, and for the total trade.

mkavinkumar1:
> you can't determine if you sold the bitcoin you bought at 30k or 50k - so both decisions would be wrong).

We can choose any. If a higher one is choosen, it will increase the price of lower one and vice versa but decrease . Here last order is choosen. We can sort them by price but you didn't like that. Maybe the option can be given to strategy. 



> If the last 2 orders happens for 10 or more times (x)

You missed this.

> If you sold the remainder at 33k$

If you sold at 2% ROI, then it is not, as we didn't acknowledge the losses ($13k * x)

xmatthias:
> You missed this.

No i did not (or didn't understand what you mean). The behaviour will be the same if you sell 10x 1 BTC or 1x 10 BTC. I've updated the post above to reflect that.

let's reverse this - please do the math for your example using the math you'd be using, please including both relative and absolute profit at each step (just as i did above).

mkavinkumar1:
> if you sell another 1 BTC

Not the existing one. We buy again 2 BTC @ $ 30k (2nd order). It will reduce the average price.

And we sold 1 BTC again. Now the loss is slightly lesser due to the new average price but still a loss.

The loss will happen until the average the price cames to $33k. After that it's profit.

> like sudden ROI triggers as the remaining relative profit "spiked" for no good reason (which means 2 sells in very short succession)

If ROI is met, then there is no reason to hold. If it is higher, it will help that. 

> - so why sell only part of the position in the first place?

It's upto the strategy. If ROI not set, then the strategy want to sell some at low profit and remaining at higher profits.

mkavinkumar1:
The concept is like [Revenue sharing](https://www.investopedia.com/ask/answers/010915/how-does-revenue-sharing-work-practice.asp) instead of equal share we are fully sharing.

xmatthias:
I don't see how your link relates to this - it'll still be the account holder who get's 100% of the profits (minus fees).

---

let's reverse this - please do the math for your example using the math you'd be using, please including both relative and absolute profit at each step (just as i did above).

mkavinkumar1:
If the first sell is -10% and the final sell is 2% as per ROI.
then the profit is 2% of final sell amount - 10% of first sell amount / total amount != 2% as expected from ROI. 

Currently reply from phone. Will add the math ASAP.

mkavinkumar1:
> The concept is like [Revenue sharing](https://www.investopedia.com/ask/answers/010915/how-does-revenue-sharing-work-practice.asp) instead of equal share we are fully sharing.

[Partial profit](https://www.investopedia.com/articles/active-trading/022315/when-and-how-take-profits-options.asp#:~:text=It%20allows%20him%20to%20retain%20%24500%20capital%20%28out%20of%20the%20initial%20capital%20of%20%24800%20to%20buy%2010%20contracts%20at%20%2480%29)

The article mentions the point but not used to keep the calculations simple.

mkavinkumar1:
> let's reverse this - please do the math for your example using the math you'd be using, please including both relative and absolute profit at each step (just as i did above).

**output:**
``` avg price after rebuy (1) is 46.67; loss : -13.67, stake:513.33, amt:11
 avg price after rebuy (2) is 44.10; loss : -24.77, stake:529.23, amt:12
 avg price after rebuy (3) is 42.09; loss : -33.86, stake:547.14, amt:13
 avg price after rebuy (4) is 40.48; loss : -41.33, stake:566.67, amt:14
 avg price after rebuy (5) is 39.17; loss : -47.50, stake:587.50, amt:15
 avg price after rebuy (6) is 38.09; loss : -52.59, stake:609.41, amt:16
 avg price after rebuy (7) is 37.19; loss : -56.78, stake:632.22, amt:17
 avg price after rebuy (8) is 36.43; loss : -60.21, stake:655.79, amt:18
 avg price after rebuy (9) is 35.79; loss : -63.00, stake:680.00, amt:19
 avg price after rebuy (10) is 35.24; loss : -65.24, stake:704.76, amt:20
 ```

**Code:**
```price = 50
amt = 10
stake = amt * price
loss=0
for i in range(10):
	amt+=2
	price = (stake + 2*30)/amt

	amt-=1
	loss += 33 - price
	stake = amt * price
	print(f' avg price after rebuy ({i +1}) is {price:.2f}; loss : {loss:.2f}, stake:{stake:.2f}, amt:{amt}')```

xmatthias:
This sample is 
* using different prices and quantities than the one you proposed above
* not doing any sells (amount is only ever raising, never falling)
* not calculating the profit / remaining profit after sells

So it's basically not showing anything from the calculation we'd want to see.
So basically, it's showing the existing logic for increasing size - not the logic for what you are attempting to implement.

xmatthias:
Can you please run the CI tools locally before pushing [as described here](https://github.com/freqtrade/freqtrade/blob/develop/CONTRIBUTING.md#before-sending-the-pr).
That'll avoid unnecessarily running CI 100ds of times with "fix flake" - and will reduce the amount of commits, especially the "fix flake" ones.

mkavinkumar1:
https://pastebin.com/mX0xGyJM

>using different prices and quantities than the one you proposed above

@ line 1 first order; line 7 second order; line 10 3rd order

> not doing any sells (amount is only ever raising, never falling)

*@ line 9 amt is reduced


> * not calculating the profit / remaining profit after sells

*@line 10

xmatthias:
I have been pretty clear what i'd like to see (https://github.com/freqtrade/freqtrade/pull/6481#issuecomment-1064455749, https://github.com/freqtrade/freqtrade/pull/6481#issuecomment-1064786626, https://github.com/freqtrade/freqtrade/pull/6481#issuecomment-1064370749).

Your script does not show any relative profit - and does not show any output between buy and sell - which means it's not "after each step" - but after every 2 steps.

mkavinkumar1:
> Your script does not show any relative profit - and does not show any output between buy and sell - which means it's not "after each step" - but after every 2 steps.

image

xmatthias:
@mkavinkumar1 this latest script/calculation proposal seems to make sense (after some smaller modifications to be more verbose - to provide more clarity).

The entry price will only change after new entries - while the profit calculation seems correct.

Please note that the calculation will not change anything on review comments like https://github.com/freqtrade/freqtrade/pull/6481#discussion_r822294330. Modifying the order object is not something we aim to - as it'll introduce bugs - and will cause things to no longer work correctly.

mkavinkumar1:
Waiting for atleast 10 results in the following poll.

https://discord.com/channels/700048804539400213/702592028403564614/952427747593031690

xmatthias:
The pull you're running is irrelevant, as it's a question provided without any context.

It's results are therefore irrelevant to me - and i will not accept a PR where an exit modifies the entry price.
You've shown nothing that convinced me of downsides to this other than "you think it's wrong".

I will NOT have you force your calculation  - it'll be me who needs to maintain/support this thereafter. 
As maintainer - i need to accept responsibility for every line of code i review and subsequently merge - therefore it's the maintainer's opinion that will be the final decision.

While opinions can change - you did nothing to change mine - as both methods of calculation work out in the end - therefore a decision was taken [6 days ago](https://github.com/freqtrade/freqtrade/pull/6481#issuecomment-1062974671).

---

I've been very clear (here https://github.com/freqtrade/freqtrade/pull/6481#issuecomment-1062974671, here https://github.com/freqtrade/freqtrade/pull/6481#pullrequestreview-903874042 and here https://github.com/freqtrade/freqtrade/pull/6481#issuecomment-1066104475), yet you seem to fail to respect pull request reviews.

Posting a pull about a decision taken shows clear disrespect - and i'll therefore take the consequences i've anounced above.

I'll therefore **close this PR** now as i finally lost my patience with your **uncollaborative behaviour**.

You are free to **resubmit this PR** under the following **conditions**:
* you will change your behaviour, and not ignore PR feedback
* PR feedback from this PR will be implemented as requested
* we will use average pricing as explained in the last few comments

Unless you can proof why the average pricing is wrong/not working - there will be no further discussion about how the price will be calculated.

Failing to comply to the above (i.e. you resubmit the PR without the requested changes) will lead to me **removing your access** to the freqtrade organization.
I hope it doesn't come to this point and would look forward to have you contribute this - but i've wasted enough time on this with someone who is not willing to respect Pull request reviews or decisions.

PR #6487, opened 2022-03-01
setup.sh install gettext for mac

samgermain:
I uninstalled all my brew installations and tried running the setup script, and got that the package `gettext` needs to be installed on mac


xmatthias (approved):
Interesting find :100: 

I'll have to trust you in this as i have no way of verifying a macos installation. :+1:

PR #6489, opened 2022-03-03
gateio market orders futures

samgermain:
Gateio futures allows market orders, just not spot

### DIFF
--- a/freqtrade/exchange/gateio.py
+++ b/freqtrade/exchange/gateio.py
@@ -37,6 +37,7 @@ class Gateio(Exchange):
     def validate_ordertypes(self, order_types: Dict) -> None:
         super().validate_ordertypes(order_types)
 
-        if any(v == 'market' for k, v in order_types.items()):
-            raise OperationalException(
-                f'Exchange {self.name} does not support market orders.')
+        if self.trading_mode != TradingMode.FUTURES:
+            if any(v == 'market' for k, v in order_types.items()):
+                raise OperationalException(
+                    f'Exchange {self.name} does not support market orders.')


xmatthias (approved):
i guess this is a point we'll need to look at separately, anyway (not just for gate.io)

i suspect we'll find several instances of some limitations being different (like spread-filter not working for binance futures).

samgermain:
> i guess this is a point we'll need to look at separately, anyway (not just for gate.io)
> 
> i suspect we'll find several instances of some limitations being different (like spread-filter not working for binance futures).

Binance, Gateio, Kraken, and Kucoin have separate api endpoints for their futures exchanges
Okex and FTX use the same endpoints.

In particular though, gateio decided to make their spot and futures api's basically nothing alike, bybit was also horrible in that regard, kraken wasn't great

xmatthias:
No what i mean is - which features are identical - which aren't.
even if the endpoint is identical - that doesn't mean that the same order-types / time-in-force, ... apply to both.

samgermain:
> No what i mean is - which features are identical - which aren't. even if the endpoint is identical - that doesn't mean that the same order-types / time-in-force, ... apply to both.

It usually does

samgermain:
@xmatthias This error still happens                     `'Exchange Gate.io does not support market orders.')`

xmatthias:
for non-futures markets, that seems to be the correct approach?

It's technically ccxt's fault. the setting for `createMarketOrder` is set to false - even if i start the exchange in futures mode. 

https://github.com/ccxt/ccxt/blob/ac9f9fe5bf6a78dfb37ef354d268d4da20cd8ee2/js/gateio.js#L42

so ccxt is telling freqtrade "i cannot trade market orders on this exchange". 
While we do have an explicit check for gate.io - the main driver is ccxt (ccxt will have to handle our order after all).

I assume this "has" would have to change based on the init type (if we start with default_mode swap - the "has" entries would need to look different for gate.io - which apparently isn't happening).

PR #6492, opened 2022-03-03
Merge develop into feat/short

samgermain:
There was a lot to merge, so I want you(@xmatthias ) to double check before it's merged in.


freqtrade/exchange/binance.py
```diff
@@ -22,6 +19,7 @@ class Binance(Exchange):
 
     _ft_has: Dict = {
         "stoploss_on_exchange": True,
+        "stoploss_order_types": {"limit": "stop_loss_limit"},
```
xmatthias, line 24: This will need separate adjustments for futures (i've been aware of this when i added it to develop).
Either we change this to have the keys as "limit_futures" or similar - or we add an optional `stoploss_order_types_futures` entry which maps futures order-types (in binance, "stop_loss_limit" is wrong for futures but should be "stop") and falls back to the original if the order-types are identical.

freqtrade/exchange/binance.py
```diff
@@ -52,82 +50,6 @@ def stoploss_adjust(self, stop_loss: float, order: Dict, side: str) -> bool:
...
-            # TODO: Name limit_rate in other exchange subclasses
-            rate = stop_price * limit_price_pct
-        else:
-            rate = stop_price * (2 - limit_price_pct)
-
-        ordertype = 'stop' if self.trading_mode == TradingMode.FUTURES else 'stop_loss_limit'
-
-        stop_price = self.price_to_precision(pair, stop_price)
-
-        bad_stop_price = (stop_price <= rate) if side == "sell" else (stop_price >= rate)
```
xmatthias, line 76: we need to bring the adjustments we did to this method to the new stoploss method in the exchange.
It's basically an extended copy of this method.

this line is one example - but it'll affect several lines here

freqtrade/rpc/telegram.py
```diff
@@ -389,46 +389,52 @@ def _get_sell_emoji(self, msg):
...
         """
         Prepare details of trade with entry adjustment enabled
         """
-        lines = []
+        lines: List[str] = []
+        if len(filled_orders) > 0:
+            first_avg = filled_orders[0]["safe_price"]
+
         for x, order in enumerate(filled_orders):
+            if order['ft_order_side'] != 'buy':
```
xmatthias: mmmmh ... will need adjustment - potentially by adding a new argument to either this method - or a new field to the json method writing the order to make it clear if it's an entry or exit order.

freqtrade/rpc/telegram.py
```diff
@@ -459,7 +465,7 @@ def _status(self, update: Update, context: CallbackContext) -> None:
             messages = []
             for r in results:
                 r['open_date_hum'] = arrow.get(r['open_date']).humanize()
-                r['num_entries'] = len(r['filled_entry_orders'])
+                r['num_entries'] = len([o for o in r['orders'] if o['ft_order_side'] == 'buy'])
```
xmatthias: same here - using `'buy'` will be wrong. maybe a "entry=True" to the order-json would be an idea?

xmatthias (commented):
The merge conflicts themselfs were not the problem - but the things that don't show up as conflict (like the modifications to the stoploss method).

While i did fix some of the points (not all!) - it's not done completely i think (but i ran out of time - and wanted to push anyway).

If you have time before me - please double-check the changes to the stoploss method very carefully - especially the ones related to futures.

freqtrade/freqtradebot.py
```diff
@@ -639,7 +639,6 @@ def execute_entry(
                 entry_tag=enter_tag, side=trade_side):
             logger.info(f"User requested abortion of buying {pair}")
             return False
-        amount = self.exchange.amount_to_precision(pair, amount)
```
xmatthias, line 642: We don't need this here as it's called in create_order, too (and there the call is correct, calculating also the contract sizes).

xmatthias: @samgermain would you mind having a look at this again (not this line in particular)

we seem to have only one instance of `amount = self.amount_to_precision(pair, self._amount_to_contracts(pair, amount))` - all others omit the `amount_to_contracts` part.

is that correct (for example for dry-run)? and is it correct to not need this to place stoploss orders?
depending on the outcome, we might be able to roll back 685820cc12368bef41244900cc8e1af94e59bd55

samgermain:
@xmatthias I also noticed that you removed `binance.stoploss`, stoploss isn't unified for `ccxt.binance` yet. I have a [pr for it](https://github.com/ccxt/ccxt/pull/11852/files), but you still need to pass in a type of `stop` or `stop_loss` or `stop_loss_limit` or whatever it is, the type in the response object is also `stop` or `stop_loss` or `stop_loss_limit`

xmatthias:
> @xmatthias I also noticed that you removed `binance.stoploss`, stoploss isn't unified for `ccxt.binance` yet. I have a [pr for it](https://github.com/ccxt/ccxt/pull/11852/files), but you still need to pass in a type of `stop` or `stop_loss` or `stop_loss_limit` or whatever it is, the type in the response object is also `stop` or `stop_loss` or `stop_loss_limit`

not a problem - this is taken care of (or at least i'm aware of it): https://github.com/freqtrade/freqtrade/pull/6492#discussion_r819277920

We can mostly unify stoploss in freqtrade - which is easier than to copy the same function to every method.

xmatthias (commented):
I'm going to merge this as i don't see anything left to do for this merge.

PR #6507, opened 2022-03-07
Update order time in force to use entry/exit wording

xmatthias:
## Summary
Update "time in force" settings to use entry/exit wording

### DIFF
--- a/freqtrade/configuration/config_validation.py
+++ b/freqtrade/configuration/config_validation.py
@@ -6,7 +6,7 @@ from jsonschema import Draft4Validator, validators
 from jsonschema.exceptions import ValidationError, best_match
 
 from freqtrade import constants
-from freqtrade.enums import RunMode
+from freqtrade.enums import RunMode, TradingMode
 from freqtrade.exceptions import OperationalException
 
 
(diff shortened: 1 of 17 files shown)


freqtrade/configuration/config_validation.py
```diff
@@ -207,3 +208,24 @@ def _validate_ask_orderbook(conf: Dict[str, Any]) -> None:
...
+        if conf.get('trading_mode', TradingMode.SPOT) != TradingMode.SPOT:
+            raise OperationalException(
+                "Please migrate your time_in_force settings to use 'entry' and 'exit'.")
+        else:
+            logger.warning(
+                "DEPRECATED: Using 'buy' and 'sell' for time_in_force is deprecated."
+                "Please migrate your time_in_force settings to use 'entry' and 'exit'."
+            )
+            time_in_force['entry'] = time_in_force.pop('buy')
+            time_in_force['exit'] = time_in_force.pop('sell')
```
xmatthias, line 231: @samgermain this segment of code is what i always meant when i said "compatibility" code.

It ensures that users simply doing an update on an existing bot will not encounter a "cannot load" situation - and therefore a smooth upgrade process.

obviously they'll still need to update this over time as we may remove the compatibility layer at some point - but it's not a hard transition that simply breaks a working spot bot - as updates are often done in a "i got 5 minutes" - and then encountering a broken system is a very frustrating experience.
I've been there often enough with other projects - and don't want my project(s) to be like this.

PR #6513, opened 2022-03-09
Gateio stoploss on exchange

samgermain:
## Summary

Enables stoploss on exchange for Gateio. 
Creating a stoploss order on gateio only requires the `stopPrice` parameter to be passed via `params` to the CCXT `createOrder` method, so the common method for creating a stoploss order should work

## Quick changelog

- gateio.fetch_stoploss_order
- gateio.cancel_stoploss_order

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -210,6 +210,9 @@ OKX requires a passphrase for each api key, you will therefore need to add this 
 
 ## Gate.io
 
+!!! Tip "Stoploss on Exchange"
+    Binance supports `stoploss_on_exchange` and uses `stop-loss-limit` orders. It provides great advantages, so we recommend to benefit from it by enabling stoploss on exchange..
+
 Gate.io allows the use of `POINT` to pay for fees. As this is not a tradable currency (no regular market available), automatic fee calculations will fail (and default to a fee of 0).
 The configuration parameter `exchange.unknown_fee_rate` can be used to specify the exchange rate between Point and the stake currency. Obviously, changing the stake-currency will also require changes to this value.
 
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -211,7 +211,7 @@ OKX requires a passphrase for each api key, you will therefore need to add this 
 ## Gate.io
 
 !!! Tip "Stoploss on Exchange"
-    Binance supports `stoploss_on_exchange` and uses `stop-loss-limit` orders. It provides great advantages, so we recommend to benefit from it by enabling stoploss on exchange..
+    Gate.io supports `stoploss_on_exchange` and uses `stop-loss-limit` orders. It provides great advantages, so we recommend to benefit from it by enabling stoploss on exchange..
 
 Gate.io allows the use of `POINT` to pay for fees. As this is not a tradable currency (no regular market available), automatic fee calculations will fail (and default to a fee of 0).
 The configuration parameter `exchange.unknown_fee_rate` can be used to specify the exchange rate between Point and the stake currency. Obviously, changing the stake-currency will also require changes to this value.
--- a/freqtrade/exchange/binance.py
+++ b/freqtrade/exchange/binance.py
@@ -23,6 +23,13 @@ class Binance(Exchange):
         "l2_limit_range": [5, 10, 20, 50, 100, 500, 1000],
     }
 
+    def stoploss_adjust(self, stop_loss: float, order: Dict) -> bool:
+        """
+        Verify stop_loss against stoploss-order value (limit or price)
+        Returns True if adjustment is necessary.
+        """
+        return order['type'] == 'stop_loss_limit' and stop_loss > float(order['info']['stopPrice'])
+
     async def _async_get_historic_ohlcv(self, pair: str, timeframe: str,
                                         since_ms: int, is_new_pair: bool = False,
                                         raise_: bool = False
--- a/freqtrade/exchange/binance.py
+++ b/freqtrade/exchange/binance.py
@@ -23,13 +23,6 @@ class Binance(Exchange):
         "l2_limit_range": [5, 10, 20, 50, 100, 500, 1000],
     }
 
-    def stoploss_adjust(self, stop_loss: float, order: Dict) -> bool:
-        """
-        Verify stop_loss against stoploss-order value (limit or price)
-        Returns True if adjustment is necessary.
-        """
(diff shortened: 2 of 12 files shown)


xmatthias (commented):
you will need to implement `stoploss_adjust` for sure to have gate.io stoploss functional (currently it'd raise an OperationalException - as is also explained in the `stoploss()` docstring

Assuming it'll work with that change - i'd want to add this to the documentation as well - 2 areas that jump to mind are both `stoploss.md` - as well as `exchanges.md` - where we have it documented for the other exchanges.

samgermain:
> you will need to implement `stoploss_adjust` for sure to have gate.io stoploss functional (currently it'd raise an OperationalException - as is also explained in the `stoploss()` docstring
> 
> Assuming it'll work with that change - i'd want to add this to the documentation as well - 2 areas that jump to mind are both `stoploss.md` - as well as `exchanges.md` - where we have it documented for the other exchanges.

I don't really understand what the stoploss_adjust is for, but it doesn't look like checking the type is necessary on binance, so I just moved the binance implementation to the exchange class and removed the type check

docs/exchanges.md
```diff
@@ -210,6 +210,9 @@ OKX requires a passphrase for each api key, you will therefore need to add this
 
 ## Gate.io
 
+!!! Tip "Stoploss on Exchange"
+    Binance supports `stoploss_on_exchange` and uses `stop-loss-limit` orders. It provides great advantages, so we recommend to benefit from it by enabling stoploss on exchange..
```
xmatthias:
```suggestion
    Gate.io supports `stoploss_on_exchange` and uses `stop-loss-limit` orders. It provides great advantages, so we recommend to benefit from it by enabling stoploss on exchange..
```

freqtrade/exchange/exchange.py
```diff
@@ -795,7 +795,7 @@ def stoploss_adjust(self, stop_loss: float, order: Dict) -> bool:
         Verify stop_loss against stoploss-order value (limit or price)
         Returns True if adjustment is necessary.
         """
-        raise OperationalException(f"stoploss is not implemented for {self.name}.")
+        return stop_loss > float(order['stopPrice'])
```
xmatthias: is `stopPrice` now unified across all exchanges in ccxt?

What will ccxt fill here if for some reason, there was no stop-price in the exchange response? will it be 0.0, or None, or missing (if an exchange doesn't have this part implemented yet?)

So far, we kept this method in the specific classes as it wasn't unified, so even if the method itself is mostly identical - we have to explicitly double-check if it's really working for this exchange.

freqtrade/exchange/binance.py
```diff
@@ -23,13 +23,6 @@ class Binance(Exchange):
         "l2_limit_range": [5, 10, 20, 50, 100, 500, 1000],
     }
 
-    def stoploss_adjust(self, stop_loss: float, order: Dict) -> bool:
-        """
-        Verify stop_loss against stoploss-order value (limit or price)
-        Returns True if adjustment is necessary.
-        """
-        return order['type'] == 'stop_loss_limit' and stop_loss > float(order['info']['stopPrice'])
```
xmatthias: The check for order-type was kindof intentional - as i don't really trust the exchange to return the correct order (especially binance in high-volume times ...).

If we think we can combine this to the exchange class, then please also look at  kucoin and huobi, which have very similar (if not identical) implementations.

xmatthias (changes requested):
Please make sure to actually test this (not just via tests, but via real orders!) - to make sure it's actually working.

What i usually test for:
* Placing an order
* cancel an order on exchange (order should be replaced  after a few seconds)
* order cancelling (hope for a raising pair, unfortunately)
* stop order hit (pull stoploss VERY close so it get's hit within reasonable time).

This can be tested with very minimal capital (i use 20$ stake for this - so my loss on this test is usually < 1$.

---

In it's current state, it's crashing the moment it tries to place the stoploss order - which makes it clear that this was never tested.

freqtrade/exchange/gateio.py
```diff
@@ -31,4 +32,18 @@ def validate_ordertypes(self, order_types: Dict) -> None:
...
+            order_id=order_id,
+            pair=pair,
+            params={'stop': True}
+        )
+
+    def cancel_stoploss_order(self, order_id: str, pair: str, params={}) -> Dict:
+        return self.cancel_order(
+            order_id=order_id,
+            pair=pair,
+            params={'stop': True}
```
xmatthias, line 49: This does not work - the implementation in CCXT differs from what you're passing here (i assume it's missaligned in ccxt ...).

https://github.com/ccxt/ccxt/blob/a88cd68ffab7382b5767ff41c57724d3829a4d4e/js/gateio.js#L3115

Compared to fetch_order:

https://github.com/ccxt/ccxt/blob/a88cd68ffab7382b5767ff41c57724d3829a4d4e/js/gateio.js#L2933

So either, we have to use `isStop` here - or ccxt needs to align the cancel method to the fetch-method and use `stop`.

freqtrade/exchange/exchange.py
samgermain: I'll add a `createStopOrder` method to CCXT, so that we can check for it
This is the progress on `stopPrice` https://github.com/ccxt/ccxt/issues/12044

freqtrade/exchange/gateio.py
samgermain: https://github.com/ccxt/ccxt/pull/12329

freqtrade/exchange/exchange.py
samgermain: This is no longer in exchange though, I just created one for gateio

samgermain:
still need to test this out

xmatthias:
> still need to test this out

Unless you've got some more concerns, i think this is good to be merged now considering the ccxt version in develop is already bumped to the "fixing version".
I've just done a nice cycle for this - with stoploss behaving exactly the way i expected it to.

xmatthias (approved):
LGTM :+1: 

(i'll hold of with merging to give you time to answer to the above question).

samgermain:
> > still need to test this out
> 
> Unless you've got some more concerns, i think this is good to be merged now considering the ccxt version in develop is already bumped to the "fixing version". I've just done a nice cycle for this - with stoploss behaving exactly the way i expected it to.

no i'm good

PR #6521, opened 2022-03-11
change develop versioning

TheJoeSchr:
…on number

## Example of what should happen:

Running `pip install -e .`:
```shell
  Running setup.py develop for freqtrade
Successfully installed freqtrade-2022.2.1.dev0
```

## What actually happens now:
```shell
  Running setup.py develop for freqtrade
Successfully installed freqtrade-develop
```

## Summary

I had troubles installing the `develop` branch locally as editable (`pip install -e .`) when using `poetry`. after some research I found out that using `"develop"` as version number is not the canonical way to do it:

from: https://peps.python.org/pep-0440/#developmental-releases
and https://peps.python.org/pep-0440/#implicit-development-release-number

I glanced that there might be 2 steps required to "fix" this:
1. use `dev` instead of `develop`
   > If used as part of a project’s development cycle, these developmental releases are indicated by including a developmental release segment in the version identifier:
`X.Y.devN    # Developmental release`

2. additionally it might be better to append the current version number also in the develop branch. `X.Y.` in aboves quote. I'm thinking if this might bring trouble with git branches. But it seems to me, this should be fixed whenever branches get merged or rebased.

After changing to `-dev` notation schema, when I use `pip install -e .` it automatically adds `0` to the version, as described here:
> Development releases allow omitting the numeral in which case it is implicitly assumed to be 0. The normal form for this is to include the 0 explicitly. This allows versions such as 1.2.dev which is normalized to 1.2.dev0.
from: https://peps.python.org/pep-0440/#implicit-development-release-number

This indicates to me, that is is working now in a canonical way.

### DIFF
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -1,5 +1,5 @@
 """ Freqtrade bot """
-__version__ = 'develop'
+__version__ = '2022.2.1-develop'
 
 if __version__.endswith('develop'):
     try:
@@ -29,5 +29,4 @@ if __version__.endswith('develop'):
             versionfile = Path('./freqtrade_commit')
             if versionfile.is_file():
                 __version__ = f"docker-{versionfile.read_text()[:8]}"
-        except Exception:
-            pass
+        except Exception: pass
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -1,12 +1,11 @@
 """ Freqtrade bot """
 __version__ = 'develop'
 
-if __version__ == 'develop':
-
+if __version__.endswith('develop'):
     try:
         import subprocess
 
-        __version__ = 'develop-' + subprocess.check_output(
+        __version__ = __version__ + '-' + subprocess.check_output(
             ['git', 'log', '--format="%h"', '-n 1'],
             stderr=subprocess.DEVNULL).decode("utf-8").rstrip().strip('"')
 
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -1,7 +1,7 @@
 """ Freqtrade bot """
-__version__ = '2022.2.1-develop'
+__version__ = '2022.2.1-dev'
 
-if __version__.endswith('develop'):
+if __version__.endswith('dev'):
     try:
         import subprocess
 


freqtrade/__init__.py
```diff
@@ -1,12 +1,11 @@
 """ Freqtrade bot """
-__version__ = 'develop'
-
-if __version__ == 'develop':
+__version__ = '2022.2.1-dev'
```
xmatthias, line 2: if you think it's necessary to rename "develop" to "dev" - then maybe.
Our versions are postfixed with the git version (so develop becomes develop-11c76c3c8 - or in the renamed case, `dev-11c76c3c8`).

There's 0 benefit on having a version number there other than merge conflicts and more work (the need to do an unnecessary commit on develop after doing a release).

Poetry is not a standard installation tool in my eyes.
If it has problems with certain formats, best open an issue there.

Installation using pip (`pip install -e .`) works perfectly fine with using both develop - as well as develop-<shahash>.

TheJoeSchr: Hmm, I think I might have miscommunicated.

It's not so much that poetry is not working, it's more that freqtrade is doing it the wrong way (according to python docs linked above). Poetry just has exposed this, any other tool that is working with version numbers might balk at this as well, because it's normalised.

Any ways, I don't really care, I just got into this and fixed it as much as I could while keeping it simple. I understand that you don't want to add extra work for virtual no gain. If it happens to be needed to change, this might be a beginning.

Maybe it would be interesting to look how other version their development versions. Again, it's just the develop branch, so one could also argue it doesn't matter at all. I just think, if it doesn't matter, one could also try to follow standard notation if this is the canonical way to do it and other tools might expect it to respect it.

xmatthias: If it's auto-incrementing (basically the way we had it - but with slightly different formatting) - i'm not per against it in general.
I'm however not a fan of "auto-incrementing" versions via CI (the way ccxt does it) - as this creates 1000nds of ci commits with "version bump" - which can make repositories explode quite quickly.


For the develop version - it has shown beneficial to have the commit in the version.
This allows (without any effort) to immediately identify "is this person running the latest version" or not - and if not - from when the version is - by simply using the commit in the value on either github search or `git log <commit>`.

While i'm aware of the python documentation above - it's mainly relevant if you push develop versions to pypi - which we do not (and probably will not do anytime soon).


---

your proposal of having a version `2022.2.1-dev-24f480b4c` would however also not respect the above links - as the last part (after dev) should be a number (2022.2.1-dev0`) - which will however make it impossible to determine "at a glance" which precise version we're talking about.

TheJoeSchr:
I don't think autoincrement is the way, for reason you stated. If I would
do it, I would increment it manually, whenever you do the stable release or
something like that. It's not as spamming and whenever a merge
conflict occurs, it's a sign that the branch is quite out of date anyways

Re: `freqtrade --version` this still works like you want, it just says now
e.g. 2022.2.1-dev-<commit>

Re: `dev0` like I said in my description (and the documentation), the `0`
after `...dev` gets autoadded (by pip?).

Re:
> it's mainly relevant if you push develop versions to pypi - which we do
not (and probably will not do anytime soon).

That's part true, I think it's also relevant to be standard confirm
whenever you need to know version. Eg. Installation, package management,
etc.


I think the least work if you want to be standard confirm is to just call
the develop version something like `N-dev`. `N` being an arbitrary number,
maybe 0 or 2022




__

xmatthias:
> I think the least work if you want to be standard confirm is to just call the develop version something like `N-dev`. `N` being an arbitrary number, maybe 0 or 2022

I'm not sure i understand the benefits of being standard-confirm _for the develop version_.
it's made for developers / people who aim to be on the cutting edge of the bot.

If you install the develop version using python standard tooling (poetry is not!) - you'll use `pip install -e .`.
This works perfectly fine, and i'm yet to encounter any drawbacks to this.

also using pip for further commands (e.g. `pip show freqtrade`) will show you the version (`Version: develop`).

Now if non-standard tooling / alternative package managers have problems with that this is one point - but i'm not sure it's a point i particularly care about, as pip (the standard python package management) is working fine with it. as it works with pip it's a bug in poetry (or whatever other tool) in my view - if it creates problems there.

if pip were to throw some error or warning about this i'd look at it differently - but as far as i'm aware it's not.

TheJoeSchr:
Yeah, like I said, it needs to be weighted, if this is something even
reasonable for the develop version. If it turns out yes, this PR might be a
first step.

Then it needs to be decided, should the version increment at all, or is
something like a static  `0-dev` instead of `develop` good enough. Next
decision, should it increment. If yes, automatically or manually. One way
to do it would be to increment it by CI once stable gets released


__

PR #6537, opened 2022-03-14
Add BT's leverage and short calculation

adrianceding:

## Summary

BT supports leverage and short

Solve the issue: #___

## Quick changelog

- Add leverage close rate calculate
- Add short close rate calculate

## What's new?

Add leverage and short close rate calculate


freqtrade/optimize/backtesting.py
```diff
@@ -354,38 +354,61 @@ def _get_ohlcv_as_lists(self, processed: Dict[str, DataFrame]) -> Dict[str, Tupl
 
     def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTuple,
                         trade_dur: int) -> float:
+        leverage = trade.leverage or 1.0
+        is_short = trade.is_short or False
+        filled_dur = int((trade.close_date_utc - trade.buy_filled_date_utc).total_seconds() // 60)
```
xmatthias: I'd kindly ask you to not mix these 2 things.

it's one thing to adjust the calculation for leverage.
but this change (filled_dur, the new column, ...) is something entirely different which does not belong into this PR in my opinion - which should aim to only fix the calculation for leverage.

Assuming you're interested in keeping this part as a separate PR - it's also done wrongly - you'll not need a new column on the trade object. The date is set on the order - so you can get it via `trade.orders[0].order_filled_date`.

adrianceding: ok. I'll move out the fill_date stuff later

xmatthias: I'm not saying it won't need fixing - but i'd rather not have it combined here - as it'll make the review more difficult (it'll mix 2 things which both affect calculation).

adrianceding: I understand. Keeping it clean is what every coder likes.Although I'm not very good at this.😄

freqtrade/optimize/backtesting.py
```diff
@@ -358,6 +358,94 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
         Get close rate for backtesting result
         """
         # Special handling if high or low hit STOP_LOSS or ROI
+        is_short = trade.is_short or False
+        if is_short:
+            return self._get_short_close_rate(sell_row, trade, sell, trade_dur)
+        else:
+            return self._get_long_close_rate(sell_row, trade, sell, trade_dur)
+
+    def _get_short_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTuple,
```
xmatthias: I'm not entirely sure i'm happy with the split between short and long close rate.

It may seem to simplify the logic - but in reality, it'll introduce a (quite high) risk of the 2 slowly drifting apart over time (the code is not next to each other - if people attempt to fix one bug, they're less likely to go someplace "completely different" (different method in this case) to try and fix the other problem.


Assuming you've been forced into this for "flake8 complexity" reasons - how about splitting the method by sell_type instead?
That'd keep logic we'd expect to behave identically in the same vicinity - while grouping the code logically.

freqtrade/optimize/backtesting.py
```diff
@@ -358,6 +358,94 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
+                if (
+                    not self.strategy.use_custom_stoploss and self.strategy.trailing_stop
+                    and self.strategy.trailing_only_offset_is_reached
+                    and self.strategy.trailing_stop_positive_offset is not None
+                    and self.strategy.trailing_stop_positive
+                ):
+                    # Worst case: price reaches stop_positive_offset and dives down.
+                    stop_rate = (sell_row[OPEN_IDX] *
+                                 (1 - abs(self.strategy.trailing_stop_positive_offset) +
+                                  abs(self.strategy.trailing_stop_positive)))
```
xmatthias: Something makes me believe that this should also include leverage (and this shows the problem mentioned above already - the same is true for the long variant ... but there's no link to that.

freqtrade/optimize/backtesting.py
```diff
@@ -358,6 +358,94 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
+                else:
+                    # Worst case: price ticks tiny bit above open and dives down.
+                    stop_rate = sell_row[OPEN_IDX] * (1 + abs(trade.stop_loss_pct / leverage))
+                    assert stop_rate > sell_row[HIGH_IDX]
```
xmatthias: This will have to use `LOW_IDX` - otherwise we can ONLY get prices outside of the candle.

freqtrade/optimize/backtesting.py
```diff
@@ -359,77 +359,121 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
+                # If open_rate is < open, only allow sells below the close on red candles.
+                raise ValueError("Opening candle ROI on red candles.")
+
+            # Use the maximum between close_rate and low as we
+            # cannot sell outside of a candle.
+            # Applies when a new ROI setting comes in place and the whole candle is above that.
+            if is_short:
+                return max(min(close_rate, sell_row[HIGH_IDX]), sell_row[LOW_IDX])
+            else:
                 return min(max(close_rate, sell_row[LOW_IDX]), sell_row[HIGH_IDX])
```
xmatthias: i don't think this condition is necessary.
Both are doing "min between close/high" - and "max between close/low" - just in the reverse sequence.

As we always assume that "high > low" (and i think that's a safe assumption) - they are actually behaving identically.

freqtrade/optimize/backtesting.py
```diff
@@ -359,77 +359,121 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
...
-                    # use Open rate if open_rate > calculated sell rate
-                    return sell_row[OPEN_IDX]
+        # Set close_rate to stoploss
+        return trade.stop_loss
+
+    def _get_close_rate_for_roi(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTuple,
+                                trade_dur: int) -> float:
+        is_short = trade.is_short or False
+        leverage = trade.leverage or 1.0
+        side_1 = -1 if is_short else 1
```
xmatthias, line 420: That's a very neat way to do this without duplicating all formulas :+1:

xmatthias (approved):
Thanks a lot! :+1: 

i'll be working on tests for this separately (already started) - so i'll add this to feat/short for now.

Thanks for your involvement! :100:

PR #6540, opened 2022-03-15
correcting docs for pricing of ask strategy

mkavinkumar1:


docs/includes/pricing.md
```diff
@@ -88,9 +88,9 @@ When selling with the orderbook enabled (`ask_strategy.use_order_book=True`), Fr
 
 #### Sell price without Orderbook enabled
 
-When not using orderbook (`ask_strategy.use_order_book=False`), the price at the `ask_strategy.price_side` side (defaults to `"ask"`) from the ticker will be used as the sell price.
```
xmatthias, line 91: Why would we remove the default value here? I think that's quite important information - assuming this property is not explicitly set - as it'll define what the user can expect as default behavior.
I think it's also important to be clear that this section ONLY applies when orderbook is false - which is now also gone (i know it's in the title -it's not very explicit what we expect).

docs/includes/pricing.md
```diff
@@ -88,9 +88,9 @@ When selling with the orderbook enabled (`ask_strategy.use_order_book=True`), Fr
...
+The following section uses `side` as the configured `ask_strategy.price_side`.
 
-When not using orderbook (`ask_strategy.use_order_book=False`), Freqtrade uses the best `side` price from the ticker if it's below the `last` traded price from the ticker. Otherwise (when the `side` price is above the `last` price), it calculates a rate between `side` and `last` price.
+When not using orderbook (`ask_strategy.use_order_book=False`), Freqtrade uses the best `side` price from the ticker if it's above the `last` traded price from the ticker. Otherwise (when the `side` price is below the `last` price), it calculates a rate between `side` and `last` price.
```
xmatthias: Maybe in the "Otherwise" sentence, we can also include a mention of `bid_last_balance`.

while it's explained in detail below - i find this pretty vague reading this sentence now "calculates a rate between" ... which rate? 

maybe something like this?

```suggestion
When not using orderbook (`ask_strategy.use_order_book=False`), Freqtrade uses the best `side` price from the ticker if it's above the `last` traded price from the ticker. Otherwise (when the `side` price is below the `last` price), it calculates a rate between `side` and `last` price based on `ask_strategy.bid_last_balance`.
```

xmatthias (commented):
Would you mind also checking the "buy price without orderbook enabled" part? 

I know this is just a reversed copy of that - so assuming the logic here is wrong, it's most likely also wrong in the buy section.

---

I also better not look at this code section too closely in general - i don't think the way it's made makes sense anymore - as the idea was to have the price between last and bid - or between last and ask - depending on bid/ask configuration (not depending on buying or selling).

so when we use "ask" - we should have price between last and ask - and when we use bid - price should really be between last and bid - irrespective of the side we're trading (which is currently not the case).

To be clear - that's however just ranting at the side for myself - not something I'll want/require from a PR that aims to fix the documentation - so feel free to just ignore this part.

xmatthias (approved):

PR #6542, opened 2022-03-15
[develop] Check version with endswith

TheJoeSchr:
Deleted previous PR by accident, it seems our conversation has been lost. sorry

I suggest for now to only merge a42d9b7f6a92caa49b0d0098aa3fb5fc1452904a (because it only makes it more reliable)and probably 15a4c19228819c5fd8d8fabd14b0e03c1a047053 (because it's a first step toward canonical versioning of develop versions)

### DIFF
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -1,5 +1,5 @@
 """ Freqtrade bot """
-__version__ = 'dev'
+__version__ = 'develop'
 
 if 'dev' in __version__:
     try:
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -1,27 +1,14 @@
 """ Freqtrade bot """
-__version__ = 'develop'
+__version__ = 'dev'
 
-if __version__ == 'develop':
-
+if 'dev' in __version__:
     try:
         import subprocess
 
-        __version__ = 'develop-' + subprocess.check_output(
+        __version__ = __version__ + '-' + subprocess.check_output(
             ['git', 'log', '--format="%h"', '-n 1'],
             stderr=subprocess.DEVNULL).decode("utf-8").rstrip().strip('"')
 
-        # from datetime import datetime
-        # last_release = subprocess.check_output(
-        #     ['git', 'tag']
-        # ).decode('utf-8').split()[-1].split(".")
-        # # Releases are in the format "2020.1" - we increment the latest version for dev.
-        # prefix = f"{last_release[0]}.{int(last_release[1]) + 1}"
-        # dev_version = int(datetime.now().timestamp() // 1000)
-        # __version__ = f"{prefix}.dev{dev_version}"
-
-        #  subprocess.check_output(
-        #     ['git', 'log', '--format="%h"', '-n 1'],
-        #     stderr=subprocess.DEVNULL).decode("utf-8").rstrip().strip('"')
     except Exception:  # pragma: no cover
         # git not available, ignore
         try:


xmatthias:
> Deleted previous PR by accident, it seems our conversation has been lost. sorry

It's not lost - just in the prior PR: #6521

> 
> I suggest for now to only merge [a42d9b7](https://github.com/freqtrade/freqtrade/commit/a42d9b7f6a92caa49b0d0098aa3fb5fc1452904a) (because it only makes it more reliable)and probably [15a4c19](https://github.com/freqtrade/freqtrade/commit/15a4c19228819c5fd8d8fabd14b0e03c1a047053) (because it's a first step toward canonical versioning of develop versions)

Merging just part of a PR is not really possible - it's an "all or nothing" approach. The alternative would be cherry-picking - but i usually don't cherry-pick out of PR's - as thatt would leave the PR in a odd state (hanging, most likely with merge conflicts).

maybe as a intermediate way to support both ways as much as possible without further update:

`if __version__.contains('dev'):`

This should cover both develop - as well as canonical version (which again - i don't see a point in having for the develop version - as we're not aiming to release develop versions to pypi).

TheJoeSchr:
Ah, sorry for double opening then!

So if you don't do cherry-pick, I squashed the commits and added your `contains()` idea, that's probably the smartest thing to do.

If I understood you correctly, you still wanted to keep `__version__ == 'develop'` BUT I still changed it to `dev`, because I thought, whats the point of this commit then. Is there any reason why it can't be `dev`? are there other parts hardcoded to check for `develop`? I didn't find any with a quick ripgrep...


ALSO I remove the dead code, that big commented out section. I saw that you last touched that 2 years ago and I figured if it hasn't reactivated by now, it's good to go.

Tell me if you think otherwise or if I overstepped and I will change it accordingly...

TheJoeSchr:
wait, there isn't a `contains` in python

TheJoeSchr:
fixed

freqtrade/__init__.py
```diff
@@ -1,27 +1,14 @@
 """ Freqtrade bot """
-__version__ = 'develop'
-
-if __version__ == 'develop':
+__version__ = 'dev'
```
xmatthias: in case you're wondering - it's this line which makes me doubt - and have a look at basically everything you didn't change :laughing:

TheJoeSchr: Sorry, I don't understand? What to you doubt? What specific about this line?

Also everything I didn't change is quite a lot 😋

xmatthias: My problem is the change from develop to dev.
I'm not sure if we use that someplace - and it'll change the way uses see the version - which may cause additional questions "i'm on the develop branch - but my version is now dev..." ? 

i'll probably revert that line for now - as it provides no benefit over the current method.

xmatthias (approved):

SmartManoj:
The changes don't change anything.

We should prepend {latest_stable_version}- to develop versions.
So that we quickly identify which month develop version we use instead of looking at that commit hash.
And poetry will install with [this](https://github.com/python-poetry/semver/issues/5#issuecomment-746527505) config

We don't need to change to dev as it will be considered as semver+branch+hash

```
__version__ = '2022.2'
__version__ += '-develop'
```
And we can comment second line in stable versions.

xmatthias:
> The changes don't change anything.

This is correct and intentional.

> So that we quickly identify which month develop version we use instead of looking at that commit hash.

There is no benefit to this. 
To properly know which version someone is using, we must know the precise commit (maybe exactly the problematic line was changed in the meantime).
this is either a release (well known state) - or a commit. `2022.2-develop` would be a range of between 100 and 1000 commits - and is therefore unsuitable for problem solving.

The only potentially acceptable way was described in this comment https://github.com/freqtrade/freqtrade/pull/6521#discussion_r825253019, which would postfix the version with the commit (we will continue to have the commit in the version).

That would however not provide the benefits - as that's still not work with poetry.
As such, changing the version would not provide any benfits, but with the drawback of more unnecessary and pointless work (incrementing the develop-version as well).


---

Freqtrade's main usage is as app, not as library. 
As such - compatibility to non-standard tooling like poetry is **not relevant** to me. 
Installation works fine with pip (which is python's standard tooling) - therefore if it's a problem for poetry, it must be considered a bug in poetry (it does however work fine with the stable version), and my advise then is to stick to standard tooling.

TheJoeSchr:
Glad to see in 777af5517d you changed your mind. What changed?

xmatthias:
pip changed.

i'm fine if this is the new standard - but i'll not change because "random edgecase tool" needs us to do something different :)

TheJoeSchr:
That's why I put the references to official documentation into the original PR #6521 though
> found out that using "develop" as version number is not the canonical way to do it:
> from: https://peps.python.org/pep-0440/#developmental-releases
> and https://peps.python.org/pep-0440/#implicit-development-release-number

not because of edgecase tool, put because it's the right thing to do and will create troubles if not changed

PR #6543, opened 2022-03-16
Add support for storing buy candle indicator rows in backtesting results

froggleston:
## Summary

Backtest results contain trade candle information, but do not allow easy access to the analysed dataframe for indicators and their values. Users would typically have to re-analyse the dataframe via the strategy to calculate these values again after backtesting. This means that helper scripts such as [buy_reasons](https://github.com/froggleston/freqtrade-buyreasons) are very slow.

## What's new?

This PR simply stores the buy signal rows of the backtested analysed dataframe in a new dict so that tools can access this after backtesting is complete, saving time and effort. The added memory and compute footprint is minimal, and this feature is turned off by default. Configs can add a new parameter, `enable_backtest_signal_candle_export` to turn this feature on.


freqtrade/optimize/backtesting.py
```diff
@@ -136,6 +137,10 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self.config['startup_candle_count'] = self.required_startup
         self.exchange.validate_required_startup_candles(self.required_startup, self.timeframe)
 
+        self.enable_backtest_signal_candle_export = False
+        if self.config.get('enable_backtest_signal_candle_export', None) is not None:
+            self.enable_backtest_signal_candle_export = bool(self.config.get('enable_backtest_signal_candle_export'))
```
xmatthias: This can be greatly simplified ... 

```suggestion
        self.enable_backtest_signal_candle_export = self.config.get('enable_backtest_signal_candle_export', False)
```

If you want to ensure the parameter is a bool, please do so using the json validator in `constants.py`.

I would also want to ensure this is disabled in hyperopt - as this WILL have an impact on performance eventually.

xmatthias (changes requested):
One thing that for sure has to change (this causes the request changes" tick) is the fact that this is neither tested nor documented.
The documentation part makes it "dead" code, which only very few people will know about - and the "not tested" part can cause this to break with future changes without anyone noticing.

---

Before moving forward with this however, let's first discover / clarify what you're actually after.
After quickly looking at the linked script, it's unclear to me why you need the whole "Opening candle" in the first place (apart from the fact that the logic fails in case of multi entries).

as far as i could identify _with a quick glance_, you're only looking at `buy_tag` and `sell_reason`.

That information should be in the trade object (also in backtesting). 
It may or may not be stored as part of the backtesting result - but I'd like to understand why you think you need the "opening candle" - or if what you really need is already available - but may just not be stored.

We also have a "sell-reasons" output already - the name suggests that your script provides basically the same but for buy-reasons - which beggs the question why it's not simply added to freqtrade itself to run "inline" with backtesting.

froggleston:
Thanks for the feedback as always!

The information isn't in backtesting AFAIK - you get the buy_tag, but not the indicators that resulted in that tag. buy_reasons helps you tune indicators by showing indicator values at the signal candle, e.g. 

image

Sell reasons are not handled in terms of indicators yet, but something i'm considering so isn't something I'd get into yet.

Disabling this for hyperopting sounds sensible.

xmatthias:
You may want to add such a screenshot to your repository too - to make it clear what the tool actually aims to do 😆

froggleston:
Point taken 😄

froggleston:
Screenshot added to the repo. I'll get on with writing some tests and documentation.

froggleston:
Just working through the updates and requests....

freqtrade/optimize/backtesting.py
froggleston: Addressed in latest commits

froggleston:
@xmatthias Have I approached the testing part correctly? I would appreciate if it could be checked against correctness/standards. Just the documentation to go

froggleston:
Now added documentation. If this is all OK, I can commit the buy_reasons.py script into the main freqtrade/scripts folder, if this makes sense. If not, we can simply link to my github repo. I'd like to see it in the main ft repo though. I think it would be a nice addition.

moinonin:
This would be a great feature. Sometimes you have backtested and backtested and all you want to do is identify those loses and just unbuy the heck out of them :dancers:

froggleston:
> 

You can use my buy_reasons.py script already, but the issue is that it's super slow. All this PR is to make it *much* quicker.

tests/optimize/test_optimize_reports.py
```diff
@@ -201,6 +202,27 @@ def test_store_backtest_stats(testdatadir, mocker):
     assert str(dump_mock.call_args_list[0][0][0]).startswith(str(testdatadir / 'testresult'))
 
 
+def test_store_backtest_candles(testdatadir, mocker):
+
+    dump_mock = mocker.patch('freqtrade.optimize.optimize_reports.file_dump_pickle')
```
xmatthias: this test seems fine at first glance.

i'd however also add a 2nd test which uses `tmpdir` instead of `testdatadir` - and which doesn't mock file-writing.
then you can test "does it really write a file" - and also "can i read the written file" - which is quite important :laughing: 

That's however slightly complicated by the fact that freqtrade doesn't have a method to actually read this file back - so that part can't really sanely be tested for now (see other comment on that).

docs/strategy_analysis_example.md
```diff
@@ -250,3 +250,75 @@ fig.show()
 ```
 
 Feel free to submit an issue or Pull Request enhancing this document if you would like to share ideas on how to best analyze the data.
+
+## Analyse the buy/entry and sell/exit tags
```
xmatthias: This is the only file you shouldn't put this documentation to.

it's auto-generated from the jupyter notebook via [this method](https://www.freqtrade.io/en/latest/developer/#updating-example-notebooks) ... 
so if you think it belongs here - it should also go to the notebook.
That doesn't make too much sense though - as it's only description - not much code to explore interactively.

 i think you might want to create a new page for this in the data-analysis section instead

docs/strategy_analysis_example.md
```diff
@@ -250,3 +250,75 @@ fig.show()
...
+
+It can be helpful to understand how a strategy behaves according to the buy/entry tags used to
+mark up different buy conditions. You might want to see more complex statistics about each buy and
+sell condition above those provided by the default backtesting output. You may also want to
+determine indicator values on the signal candle that resulted in a trade opening.
+
+We first need to enable the exporting of trades from backtesting:
+
+```
```
xmatthias:
```suggestion
``` bash
```

docs/strategy_analysis_example.md
```diff
@@ -250,3 +250,75 @@ fig.show()
...
+Before running your next backtest, make sure you either delete your old backtest results or run
+backtesting with the `--cache none` option to make sure no cached results are used.
+
+If all goes well, you should now see a `backtest-result-{timestamp}_signals.pkl` file in the
+`user_data/backtest_results` folder.
+
+Now run the buy_reasons.py script, supplying a few options:
+
+```
+./scripts/buy_reasons.py -c <config.json> -s <strategy_name> -t <timerange> -g0,1,2,3,4
```
xmatthias: If it's documented this way - then the script should also be included (otherwise we'd need some hint on how to get it there).

I'm however a bit concerned about throwing a new, 500+ line file into the scripts directory.
Scripts are usually not covered by tests - so it can break at any time, especially since it imports internal modules/packages from freqtrade.

Maybe you do have a version in another branch that does utilize the pickle-file already (and is cleaned up from the rest) - so we can actually see "how much" of that script is still left?

---

Instead of a script (we only really have the scripts folder for the rest client - as that intentionally doesn't depend on anything from freqtrade).
i think this potentially should be a subcommand (`freqtrade ...`) - which would make it testable, as well as (if done right) usable from within the jupyter notebook, or maybe even "inline" to backtesting via an additional flag.

I'm however fully aware that converting it to that may obviously be a bit more work than just dumping a script though ...

I would however see this as a secondary step (separate PR) ... to keep this one nice and small.

freqtrade/misc.py
```diff
@@ -86,6 +87,21 @@ def file_dump_json(filename: Path, data: Any, is_zip: bool = False, log: bool =
...
+    Dump object data into a file
+    :param filename: file to create
+    :param data: Object data to save
+    :return:
+    """
+
+    if log:
+        logger.info(f'dumping pickle to "{filename}"')
+    with open(filename, 'wb') as fp:
+        pickle.dump(data, fp)
```
xmatthias: i'm ... quite unhappy with the use of pure pickle. it does not create a transportable dataformat (you can't read the file from other languages - and it'll break should pandas (or anything that's been dumped) be updated in the meantime - potentially silently so you won't notice it loaded wrong content).

If we assume we must use pickle (as format), then i'd prefer to use `joblib.dump`.
we use that in hyperopt ... and it's quite good when dumping pandas/numpy arrays

[some comparison](https://stackoverflow.com/questions/12615525/what-are-the-different-use-cases-of-joblib-versus-pickle)

freqtrade/optimize/backtesting.py
```diff
@@ -1070,6 +1077,26 @@ def backtest_one_strategy(self, strat: IStrategy, data: Dict[str, DataFrame],
         })
         self.all_results[self.strategy.get_strategy_name()] = results
 
+        if self.backtest_signal_candle_export_enable and \
```
xmatthias: This should be moved to a separate function that's only called here.

It's not "really" needed by `backtest_one`, so it should be treated separately i think.
it'll also make the dependencies clearer (what's needed as input, ...) - as these have to be passed in as arguments.

freqtrade/optimize/backtesting.py
```diff
@@ -1070,6 +1077,26 @@ def backtest_one_strategy(self, strat: IStrategy, data: Dict[str, DataFrame],
...
+
+                pairdf = preprocessed_tmp[pair]
+                resdf = results['results']
+                pairresults = resdf.loc[(resdf["pair"] == pair)]
+
+                if pairdf.shape[0] > 0:
+                    for t, v in pairresults.open_date.items():
+                        allinds = pairdf.loc[(pairdf['date'] < v)]
+                        signal_inds = allinds.iloc[[-1]]
+                        signal_candles_only_df = signal_candles_only_df.append(signal_inds)
```
xmatthias: I wonder what the impact of this is ...

how much would this method take for a backtest of - say 1000 trades?
would that also cause the "fragmented dataframe" warning?

tests/optimize/test_optimize_reports.py
froggleston: OK i'll code in an actual write. The buy_reasons script is the reading part, which I can also add (see below)

docs/strategy_analysis_example.md
froggleston: OK, I'll move it to another new page

docs/strategy_analysis_example.md
froggleston: I think i'll do as you suggest - splitting out the core functionality into a formal `freqtrade` command. I've done this for the notebook so it shouldn't be too much of a problem. We don't need the whole separate script.

Inline to backtesting would be fun! I'll leave that for another PR though.

freqtrade/misc.py
froggleston: We don't have to use pickling I guess. It was just the simplest way to get a working proof-of-concept. I'll take a look at joblib to see any differences.

In terms of improvement to the idea, would you suggest simply dumping the dict of dataframes as just that - a json/csv representation?

freqtrade/optimize/backtesting.py
froggleston: ok I'll move it out

freqtrade/optimize/backtesting.py
froggleston: 1000 trades is only 1000 signal candles, so it's not that much of a problem IMO.

docs/strategy_analysis_example.md
xmatthias: you might want to leave both of these for a separate PR - that way we can add this mostly "as is" for now - and have a followup PR that adds a new command ...

freqtrade/misc.py
xmatthias: json/csv might introduce a problem of precision.
if you dump a dataframe to csv without some additional magic, it'll lose precision after 12ish decimals (it'll be cutoff iirc).

while not a problem for most prices - it can be a problem for some indicators - which would then appear to be 0.0, instead of 0.000000000000055 ...

docs/strategy_analysis_example.md
froggleston: OK, sounds good. I'll commit the buy_reasons.py script to the main FT scripts folder? Or simply link to my repo for now?

freqtrade/misc.py
froggleston: Ah OK, makes sense. I'll continue to check out joblib then.

docs/strategy_analysis_example.md
xmatthias: link to your repo for now - that way we can avoid confusion i think

freqtrade/misc.py
froggleston: Suggestion implemented in [b3cb722](https://github.com/freqtrade/freqtrade/pull/6543/commits/b3cb7226467c359b0748cb4866f8f8fcfa649a53)

docs/strategy_analysis_example.md
froggleston: Fixed in [b3cb722](https://github.com/freqtrade/freqtrade/pull/6543/commits/b3cb7226467c359b0748cb4866f8f8fcfa649a53)

docs/strategy_analysis_example.md
froggleston: Moved docs in [b3cb722](https://github.com/freqtrade/freqtrade/pull/6543/commits/b3cb7226467c359b0748cb4866f8f8fcfa649a53)

tests/optimize/test_optimize_reports.py
froggleston: Test implemented in [b3cb722](https://github.com/freqtrade/freqtrade/pull/6543/commits/b3cb7226467c359b0748cb4866f8f8fcfa649a53)

freqtrade/optimize/backtesting.py
froggleston: Implemented in [f92997d](https://github.com/freqtrade/freqtrade/pull/6543/commits/f92997d3789c6eb8b9116e4a55dac2b9b7112aa7)

froggleston:
OK I think I've implemented all the suggestions!

freqtrade/optimize/backtesting.py
froggleston: Oh, or did you mean move it out completely and generate it outside `backtest_one`? I think it's better inside backtest_one as it doesn't mean you're going to forget to generate the signal candles each time you want to process subsequent backtests...

xmatthias: i think you will have to call it from here ... so my intend is what you implemented... stop doubing yourself after implementing correctly ;)

freqtrade/constants.py
```diff
@@ -380,6 +380,7 @@
         },
         'position_adjustment_enable': {'type': 'boolean'},
         'max_entry_position_adjustment': {'type': ['integer', 'number'], 'minimum': -1},
+        'backtest_signal_candle_export_enable': {'type': 'boolean'},
```
xmatthias: while reading through the documentation, i did notice that this has not really good usability.

It's based on the assumption that we set this for every backtest (hence via config) - which is most likely wrong / unnecessary (you'll potentially test 20 things before looking at this in detail).

i'd rather combine this with the export flag  - so we can provide `--export signals` - which will then export both trades and signals ... that would allow this to be provided via cli argument - which should simplify only occasional use.

That'll obviously also require some additional changes ... if you let me know what you think about this i can look at it (i roughly know where this should be applied).

froggleston: Sounds great! I'll take a look and see how that works and implement it

xmatthias: well in the end, you just have to search for where `'export' == trades` is used via config (should only be backtesting, and obviously the constants). You'll then have to "augment" that to check both scenarios - as i'd assume that `--export signals` exports both (exporting only signals without trade does not make sense).

froggleston: Yup - looking through it now

xmatthias (approved):
LGTM :+1: 

Looking forward to the followup-PR which will add this functionality natively ;)

PR #6545, opened 2022-03-16
Partial exit using average price

### LINKED ISSUE #1519: position mgmt (Position stacking / pyramiding)

mkavinkumar1:
Old PR #6481 

buy 10 coins @ price $10
buy 15 coins @ price $5
Now avg price is $7
sell 15 coins @ price $6

If we use avg price for the calculation (current PR), the loss is $15 and avg price is still $7. Then all partial sells would be in loss until the current price > average price.

If we use last order price (old PR), the profit is $15 and average price is $10. And after sharing the profit to previous buys, the new avg price is $8.5

Discord discussions :
https://discord.com/channels/700048804539400213/702592028403564614/950612052798500904


ajhepple:
> And ater sharing the profit to previous buys, the new avg price is $8.5

This makes no sense to me. The average price is an average _purchase_ price and doesn't change after a _sale_.

mkavinkumar1:
> > And ater sharing the profit to previous buys, the new avg price is $8.5
> 
> This makes no sense to me. The average price is an average _purchase_ price and doesn't change after a _sale_.

https://www.investopedia.com/articles/active-trading/022315/when-and-how-take-profits-options.asp#:~:text=It%20allows%20him%20to%20retain%20%24500%20capital%20%28out%20of%20the%20initial%20capital%20of%20%24800%20to%20buy%2010%20contracts%20at%20%2480%29

freqtrade/freqtradebot.py
```diff
@@ -470,17 +472,25 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
         stake_amount = strategy_safe_wrapper(self.strategy.adjust_trade_position,
                                              default_retval=None)(
             trade=trade, current_time=datetime.now(timezone.utc), current_rate=current_rate,
-            current_profit=current_profit, min_stake=min_stake_amount, max_stake=max_stake_amount)
+            current_profit=current_profit, min_stake=min_stake_amount,
+            max_stake=max_stake_amount, current_entry_rate=current_entry_rate,
```
xmatthias: The moment you change a strategy interface part around - this has to be done consistently (which means wherever it appears, including the documentation and templates).

I'm however not entirely certain if this will make sense anyway.
the way "get_rate" is currently implemented will cause it to make two requests to the exchange.
Freqtrade would be doing this in every iteration - so we have to consider the performance impact of this - to get prices which are usually less than 0.1% apart.
Depending on the configuration - you may even get the same rate twice (if both sides are using ASK, for example).

While the "2 calls" problem can probably be worked around by changing how that method works - i'd not want to have that part included in this PR - as it's not really related - and will require an additional configuration check to make sure that both sides are either using orderbook - or both sides are not.

freqtrade/freqtradebot.py
```diff
@@ -1234,24 +1253,39 @@ def execute_trade_exit(
...
         Sends rpc notification when a sell occurred.
         """
-        profit_rate = trade.close_rate if trade.close_rate else trade.close_rate_requested
-        profit_trade = trade.calc_profit(rate=profit_rate)
         # Use cached rates here - it was updated seconds ago.
         current_rate = self.exchange.get_rate(
             trade.pair, refresh=False, side="sell") if not fill else None
-        profit_ratio = trade.calc_profit_ratio(profit_rate)
+        if sub_trade:
+            assert order is not None
```
xmatthias: please don't use assert in regular code (outside of tests)
Would this ever be hit, freqtrade would crash/restart as assertionexceptions are not captured.

even worse - if that was the intend (to have it crash) - it can be disabled at an interpreter level - so assert statements are simply skipped.

freqtrade/freqtradebot.py
```diff
@@ -470,17 +472,25 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
...
         if stake_amount is not None and stake_amount > 0.0:
             # We should increase our position
             self.execute_entry(trade.pair, stake_amount, trade=trade)
 
         if stake_amount is not None and stake_amount < 0.0:
             # We should decrease our position
-            # TODO: Selling part of the trade not implemented yet.
-            logger.error(f"Unable to decrease trade position / sell partially"
-                         f" for pair {trade.pair}, feature not implemented.")
+            amount = -stake_amount / current_exit_rate
```
xmatthias: i think using `abs()` here will make more sense - with the same result.

I had to look twice why we need to use "-" here - making it ABS will force it to be positive - and be clear that that's the actual intend, not to have amount be negative.

freqtrade/freqtradebot.py
```diff
@@ -1373,9 +1409,6 @@ def update_trade_state(self, trade: Trade, order_id: str, action_order: Dict[str
         self.handle_order_fee(trade, order_obj, order)
 
         trade.update_trade(order_obj)
-        # TODO: is the below necessary? it's already done in update_trade for filled buys
-        trade.recalc_trade_from_orders()
-        Trade.commit()
```
xmatthias, line 1658: The comment was about the "recalc" line - not about the commit.
The commit IS needed, and missing it is clearly wrong.

freqtrade/persistence/models.py
```diff
@@ -630,27 +674,23 @@ def calc_profit_ratio(self, rate: Optional[float] = None,
...
+            if o.ft_is_open or not o.filled:
                 continue
 
             tmp_amount = o.safe_amount_after_fee
-            tmp_price = o.average or o.price
-            if o.filled is not None:
-                tmp_amount = o.filled
+            tmp_price = o.safe_price
+            is_sell = o.ft_order_side != 'buy'
+            side = [1, -1][is_sell]
```
xmatthias: please use `1 if is_sell else -1`. 
Creating a list and then using the index of that makes it unnecessarily complex.

(same below)

freqtrade/persistence/models.py
```diff
@@ -679,9 +719,9 @@ def select_order(
         :param is_open: Only search for open orders?
         :return: latest Order object if it exists, else None
         """
-        orders = self.orders
+        orders = [o for o in self.orders if o.filled or o.ft_is_open]
```
xmatthias: I don't think this is correct.

This is `select_order()` - not `select_open_order` or similar - so i expect to get **all** orders for that side (independent if they are open or not).
`is_open` is a parameter that is checked below - so having it here is clearly wrong.

tests/test_freqtradebot.py
```diff
@@ -227,8 +226,7 @@ def test_edge_overrides_stoploss(limit_buy_order_usdt, fee, caplog, mocker,
     freqtrade.strategy.min_roi_reached = MagicMock(return_value=False)
     freqtrade.enter_positions()
     trade = Trade.query.first()
-    oobj = Order.parse_from_ccxt_object(limit_buy_order_usdt, 'ADA/USDT', 'buy')
-    trade.update_trade(oobj)
```
xmatthias: What's the reason to remove this and instead fake open_rate?

tests/rpc/test_rpc_telegram.py
```diff
@@ -421,7 +421,9 @@ def test_daily_handle(default_conf, update, ticker, limit_buy_order, fee,
 
     # Simulate fulfilled LIMIT_BUY order for trade
     oobj = Order.parse_from_ccxt_object(limit_buy_order, limit_buy_order['symbol'], 'buy')
-    trade.update_trade(oobj)
+    trade.open_rate = oobj.safe_price
+    trade.amount = oobj.safe_amount_after_fee
+    trade.recalc_open_trade_value()
```
xmatthias: if this change is necessary, i think it suggest a problem (either with the test - or with the code that's causing it).

the trade will have an open_rate here already (as well as an amount) - so calling `update_trade()` with the buy order will update the values accordingly.
That obviously doesn't just apply here - but also to all other instances where the same change was applied.

freqtrade/freqtradebot.py
```diff
@@ -470,17 +472,25 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
...
-            # TODO: Selling part of the trade not implemented yet.
-            logger.error(f"Unable to decrease trade position / sell partially"
-                         f" for pair {trade.pair}, feature not implemented.")
+            amount = -stake_amount / current_exit_rate
+            if trade.amount - amount < min_stake_amount:
+                logger.info('Remaining amount would be too small')
+                return
+            if amount > trade.amount:
+                logger.info(f"Amount is higher than available. {amount} > {trade.amount}")
+                return
```
xmatthias: I'm not entirely sure if this is good behavior - we could just assume this as "sell remaining" - so if i still own 10 - and the method returns "-11" - we'd simply exit the whole position.

I'd be interested in knowing the reason for deciding onto this behaviour?

freqtrade/freqtradebot.py
```diff
@@ -1263,29 +1297,29 @@ def _notify_exit(self, trade: Trade, order_type: str, fill: bool = False) -> Non
...
             'fiat_currency': self.config.get('fiat_display_currency', None),
+            'sub_trade': sub_trade,
         }
-
-        if 'fiat_display_currency' in self.config:
-            msg.update({
-                'fiat_currency': self.config['fiat_display_currency'],
-            })
+        if sub_trade:
+            msg['cumulative_profit'] = trade.realized_profit
```
xmatthias: is there a downside to always provide this?
if it's a simple scenario (full buy followed by a full sell) - this would simply be 0 - correct?

freqtrade/persistence/models.py
```diff
@@ -127,6 +127,7 @@ class Order(_DECL_BASE):
     side = Column(String(25), nullable=True)
     price = Column(Float, nullable=True)
     average = Column(Float, nullable=True)
+    initial_average = Column(Float, nullable=True)
```
xmatthias: I think this is a leftover from the earlier attempt and can safely be removed (it's not referenced anywhere)

freqtrade/persistence/models.py
```diff
@@ -854,6 +895,7 @@ class Trade(_DECL_BASE, LocalTrade):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        self.realized_profit = 0
```
xmatthias: I don't think this is necessary.

it is defined as 0 for LocalTrade - and the database default will take care of it in the database scenario.
Did you encounter this to be 0? If yes - then we'll need to look at this - as i don't understand what makes this different from the 20 or so other columns we have and which don't need this.

tests/test_persistence.py
```diff
@@ -1638,32 +1642,6 @@ def test_recalc_trade_from_orders_ignores_bad_orders(fee):
     assert trade.open_trade_value == 2 * o1_trade_val
     assert trade.nr_of_successful_buys == 2
 
-    # Just to make sure sell orders are ignored, let's calculate one more time.
-    sell1 = Order(
```
xmatthias: I'd want to have this kept there - as it can test that the entry price will not be modified by exits.

It's possible that some asserts below no longer make sense - but the case itself is valid and should be kept.

On the contrary to removing this - i'd want to have an additional test showing more complex behaviours (like mixing/matching of buys/sells - specifically testing the behavior of the calculation part.
While the test in `test_freqtradebot` is good - it also includes the whole "bot complexity" - whereas here we can focus purely on the calculation.

freqtrade/persistence/models.py
```diff
@@ -504,14 +513,49 @@ def update_trade(self, order: Order) -> None:
...
+            self.stake_amount = self.open_rate * self.amount
+            self.realized_profit += profit
+
+        self.close_profit_abs = profit
+        self.close_profit = sell_stake_amount / (sell_stake_amount - profit) - 1
+        self.recalc_open_trade_value()
+        if is_non_bt:
+            Trade.commit()
+
+    def calc_profit2(self, open_rate: float, close_rate: float,
```
xmatthias: What's the difference between this `calc_profit2()` and the exiting `calc_profit()`?

once you resolve all the sub-calls that are made, they should come to the same formula - or at least to the same result using a different formula?

freqtrade/persistence/models.py
```diff
@@ -504,14 +513,49 @@ def update_trade(self, order: Order) -> None:
...
+                     amount: float) -> float:
+        return float(Decimal(amount) *
+                     (Decimal(1 - self.fee_close) * Decimal(close_rate) -
+                      Decimal(1 + self.fee_open) * Decimal(open_rate)))
+
+    def get_open_rate(self, profit: float, close_rate: float,
```
xmatthias: I'm not sure i understand the aim of this.
the trade open rate should be stored at `open_rate` at all times. 

why would we need a different calculation for just this one message?

freqtrade/persistence/models.py
```diff
@@ -487,12 +490,18 @@ def update_trade(self, order: Order) -> None:
             self.amount = order.safe_amount_after_fee
             if self.is_open:
                 logger.info(f'{order.order_type.upper()}_BUY has been fulfilled for {self}.')
-            self.open_order_id = None
+            # condition to avoid reset value when updating fees
+            if self.open_order_id == order.order_id:
```
xmatthias: Can you maybe explain the condition as to when you'd expect to hit this (and obviously also the sell version of it)?

updating fees happens right after the buy - so it's quite unlikely that an open sell order already exists?

tests/test_freqtradebot.py
```diff
@@ -4632,6 +4641,346 @@ def make_sure_its_651(*args, **kwargs):
...
+            ('sell', 50, 5),
+        ),
+        (
+            # amount, open_rate, stake_amount, cumulative_profit, realized_profit
+            (100.0, 10.0, 1000.0, 0.0, None,),
+            (200.0, 12.5, 2500.0, 0.0, None,),
+            (150.0, 12.5, 1875.0, -28.0625, -28.0625,),
+            (50.0, 12.5, 625.0, 713.8125, 741.875,),
+            (50.0, 12.5, 625.0, 713.8125, 336.625,),
+        )
```
xmatthias: From an test org point of view - i think i would combine these two arrays into one long array - where the first 3 columns are "order" - and the remainder is price.

That will make it a lot easier for the "reader" to immediately identify after which step we expect which result - while also removing the need to `res[idx]` all asserts.

freqtrade/persistence/models.py
```diff
@@ -504,14 +513,49 @@ def update_trade(self, order: Order) -> None:
...
+    def process_sell_sub_trade(self, order: Order, is_closed: bool = True,
+                               is_non_bt: bool = True) -> None:
+        sell_amount = order.safe_amount_after_fee
+        sell_rate = order.safe_price
+        sell_stake_amount = sell_rate * sell_amount * (1 - self.fee_close)
+        if sell_amount == self.amount:
+            if is_closed:
+                self.close(sell_rate)
+                if is_non_bt:
+                    Trade.commit()
```
xmatthias: This seems unnecessarily complex. 
if you remove all commits from this method, and don't return after `process_sell_sub_trade()` in `update_trade()` you'll get the same result - without the need to explicitly differentiate between bt and non-bt - allowing you to remove the `is_non_bt` parameter.

freqtrade/freqtradebot.py
```diff
@@ -459,7 +459,9 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
                 return
         else:
             logger.debug("Max adjustment entries is set to unlimited.")
-        current_rate = self.exchange.get_rate(trade.pair, refresh=True, side="buy")
```
xmatthias: While not directly on this line - you'll need to change how `max_entry_position_adjustment` is handled/interpreted above, otherwise your "partial sell" will not be called once `max_entry_position_adjustment` is reached.

I'd probably update it to be called all the time - ignoring the buy if `max_entry_position_adjustment` is reached. That'll obviously also have to be updated in the docs.

xmatthias (commented):
This seems a lot better now than the previous attempt. :+1: 

there's quite a few doubts i have here and there, but overall it looks quite promising.

obvious things that are still missing is the documentation - but it's safe to leave that to the end - to a point where we're happy with the code.

freqtrade/freqtradebot.py
mkavinkumar1: # for mypy only; order will always be passed during sub trade
https://stackoverflow.com/a/68471435/5033247

tests/test_freqtradebot.py
mkavinkumar1: In update trade, open_rate and amount is set and recalc_trade_from_orders is not used there is when only one buy order. Now the condition is removed. The first order id doesn't match with limit_buy_order_usdt, so the open_rate won't change. Also the order amount was less than the first order amount. So when selling, it considers it as a partial sell. We can change the order id but some trades doesn't have orders.

tests/rpc/test_rpc_telegram.py
mkavinkumar1: https://github.com/freqtrade/freqtrade/pull/6545/files/1e9cbd579f28ce5425ce2f68f0b164c2f81ba153#r830439888

freqtrade/freqtradebot.py
mkavinkumar1: If the strategy thinks it has higher amount, then the strategy is wrong. Freqtrade shouldn't simply ignore that.

freqtrade/freqtradebot.py
mkavinkumar1: we only provide when it is a partial sell.

freqtrade/persistence/models.py
mkavinkumar1: Getting None for this.  Remaining columns doesn't use its initial value. 
When using this solution, error occurred when creating model {something with indexes)
https://stackoverflow.com/a/20648661

freqtrade/persistence/models.py
mkavinkumar1: calc_profit uses open trade value which can't be used for partial sells. calc_profit2 is a generalized method. we can safely remove calc_profit in further updates.

freqtrade/persistence/models.py
mkavinkumar1: if updating fees is failed. Also it is better to be safe than sorry.

freqtrade/freqtradebot.py
mkavinkumar1: when using bid strategy when selling, 70% of trades are timed out. 
current_entry_rate, current_exit_rate can be fetched using kwargs var. As current_rate is kept for now, it won't affect interface. Keeping entry_rate now.

freqtrade/freqtradebot.py
mkavinkumar1: done

freqtrade/freqtradebot.py
mkavinkumar1: reverted

freqtrade/persistence/models.py
mkavinkumar1: u swapped values. updated

freqtrade/persistence/models.py
mkavinkumar1: used for previous PR as filled is reduced. Now the condition is always true when filled is not changed. reverted.

freqtrade/persistence/models.py
mkavinkumar1: yep; removed;

tests/test_persistence.py
mkavinkumar1: removed as it is tested already in previous test. Added again.

freqtrade/persistence/models.py
mkavinkumar1: used for previous PR, removed.

tests/test_freqtradebot.py
mkavinkumar1: Planning to do different modes of Partial sells. So that time, input can be reused. used zip.

freqtrade/persistence/models.py
mkavinkumar1: updated

freqtrade/freqtradebot.py
mkavinkumar1: updated

freqtrade/freqtradebot.py
```diff
@@ -470,17 +464,32 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
...
+                else:
+                    logger.debug("Max adjustment entries is set to unlimited.")
             self.execute_entry(trade.pair, stake_amount, trade=trade)
 
         if stake_amount is not None and stake_amount < 0.0:
             # We should decrease our position
-            # TODO: Selling part of the trade not implemented yet.
-            logger.error(f"Unable to decrease trade position / sell partially"
-                         f" for pair {trade.pair}, feature not implemented.")
+            amount = abs(stake_amount) / current_exit_rate
```
SmartManoj: Strategy should return value as Decimal. Else 
```
a = Decimal(float(Decimal(-13.533) * Decimal(.3134))) / Decimal(.3134)
print(float(a))
>>> -13.533000000000001
```

freqtrade/freqtradebot.py
xmatthias: there's better methods to satisfy mypy (like `if order is not None`).

freqtrade/persistence/models.py
xmatthias: for sure possible - as said - i know it's working - but it's unnecessarily complex

freqtrade/freqtradebot.py
xmatthias: i do understand that from the code - but is there a downside to providing it all the time?
it'd make both this method, as well as methods using this simpler, as they can rely on this being available.

freqtrade/freqtradebot.py
stash86: If the bot has max rebuy of 3, and then it has used it all and partial sell the last order, should the bot be assumed to have 1 open rebuy slot, or it's not allowed to rebuy again?

mkavinkumar1: It is not allowed now. If we want, we can traverse buys in reverse order (a config is need to determine this), and check buy amount with total sell amount and reducing count

freqtrade/freqtradebot.py
mkavinkumar1: > is there a downside to always provide this? 

Nope;  Need to add in all test cases 

> if it's a simple scenario (full buy followed by a full sell) - this would simply be 0 - correct?

Yes

freqtrade/freqtradebot.py
mkavinkumar1: updated

freqtrade/freqtradebot.py
xmatthias: that's not going to happen (it's obviously up to the strategy's disgression), but you can't rely on that - which is why i said - are we sure we'd not want to allow the strategy to use a "bigger" rate to exit with the final value?

freqtrade/freqtradebot.py
xmatthias: i'm not sure if we need an additional config for this considering that this is mostly a "convenience" functionality - but also something that can easily be handled within the strategy.
We have too many config options barely anyone knows how to use already - no point in making every single question configurable i think.

As such, i think it's more a point on "do we want it this way or that".
As long as the behaviour is documented, i don't see this as a breaking point - it's just "it's like this" ...

freqtrade/persistence/models.py
xmatthias: no, it's not reverted but remained identical.

freqtrade/persistence/models.py
xmatthias: why can't open_trade_value be used? 
the only fields used for that (recalculation) are `amount` and `open_rate`.

Both will have to be updated throughout the trade to reflect the current average buy rate - as well as the current amount we still own.

freqtrade/persistence/models.py
xmatthias: Please also remove the unnecessary parameter `is_non_bt` from the function header.

tests/test_freqtradebot.py
xmatthias: Then we need to update the order-id of the mocked trade - and eventually add orders for trades which don't have orders.

This is however clearly wrong - both here and in `test_rpc_telegram`.

freqtrade/persistence/models.py
SmartManoj: if amount is 50 and open_rate is 1, open_trade_value = 50
if we sell 1 coin at close rate 2, calc_profit2 returns 1 (fees not considered) 
calc_profit returns (100 - 50 = 50 ( close_trade_value - open_trade_value)

freqtrade/persistence/models.py
SmartManoj: updated

freqtrade/persistence/models.py
SmartManoj: not sure why. again reverted

tests/test_freqtradebot.py
SmartManoj: updated

tests/rpc/test_rpc_telegram.py
SmartManoj: updated

freqtrade/exchange/exchange.py
```diff
@@ -1174,6 +1177,28 @@ def get_rate(self, pair: str, refresh: bool, side: str) -> float:
...
 
+    def get_rates(self, pair: str, refresh: bool) -> Tuple[float, float]:
+        buy_rate = sell_rate = None
+        if not refresh:
+            buy_rate, sell_rate = self._buy_rate_cache.get(pair), self._sell_rate_cache.get(pair)
+
+        bid_strategy = self._config.get('bid_strategy', {})
+        ask_strategy = self._config.get('ask_strategy', {})
+        order_book = ticker = None
+        if bid_strategy.get('use_order_book', False) and ('use_order_book' in bid_strategy):
```
xmatthias: you can safely omit the 2nd part - it's not doing anything, really.

if any, you'd need to check if it's inside before access (exploiting condition shortcircuiting) - but that's not necessary when using `.get()`.

freqtrade/exchange/exchange.py
```diff
@@ -1174,6 +1177,28 @@ def get_rate(self, pair: str, refresh: bool, side: str) -> float:
 
         return rate
 
+    def get_rates(self, pair: str, refresh: bool) -> Tuple[float, float]:
```
xmatthias: We'll for sure want a test for this, testing the different behaviours we can expect.

While i guess the code will work now - we need to ensure it's not breaking in the future without us noticing.

overall, i like how it's implemented though :+1:

freqtrade/exchange/exchange.py
```diff
@@ -1174,6 +1177,28 @@ def get_rate(self, pair: str, refresh: bool, side: str) -> float:
...
+        buy_rate = sell_rate = None
+        if not refresh:
+            buy_rate, sell_rate = self._buy_rate_cache.get(pair), self._sell_rate_cache.get(pair)
```
xmatthias: I am in general not a particular friend of double-assignment like this, especially if the source is from different variables.

Splitting it into 2 lines will make the code more readable.

freqtrade/freqtradebot.py
```diff
@@ -452,14 +453,8 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
...
-        if self.strategy.max_entry_position_adjustment > -1:
-            count_of_buys = trade.nr_of_successful_buys
-            if count_of_buys > self.strategy.max_entry_position_adjustment:
-                logger.debug(f"Max adjustment entries for {trade.pair} has been reached.")
-                return
-        else:
-            logger.debug("Max adjustment entries is set to unlimited.")
-        current_rate = self.exchange.get_rate(trade.pair, refresh=True, side="buy")
+        current_entry_rate, current_exit_rate = self.exchange.get_rates(trade.pair, True)
+        current_rate = current_entry_rate  # backward compatibilty
```
xmatthias: You can save this assignment, and instead pass it to the callback directly (basically the same code - but in the function call).

SmartManoj: updated

freqtrade/exchange/exchange.py
SmartManoj: updated

freqtrade/exchange/exchange.py
SmartManoj: updated

freqtrade/exchange/exchange.py
SmartManoj: added tests. is there any other behaviors need to test?

freqtrade/freqtradebot.py
SmartManoj: Should we use this function?
```
def div(a, b):
    p = Decimal(str(a)).as_tuple().exponent
    return round(float(Decimal(a) / Decimal(b)), -p)
```

SmartManoj: Should we need this function?
```
def div(a, b):
    p = Decimal(str(b)).as_tuple().exponent
    return round(float(Decimal(a) / Decimal(b)), -p)
 ```
 
 FYI, `price_to_precision` not works for `FEG/USDT` as it has price in 12 decimal points, the function truncates to 10 digit.

File: tests/exchange/test_exchange.py [hunk 25]
```diff
@@ -1996,16 +2002,19 @@ def test_get_sell_rate(default_conf, mocker, caplog, side, bid, ask,
...
-    ('buy', 'ask', None, 4, 4,  0, 4),  # ask not available
-    ('buy', 'ask', None, None, 4,  0, 4),  # ask not available
-    ('buy', 'bid', 6, None, 4,  0, 5),  # bid not available
-    ('buy', 'bid', None, None, 4,  0, 5),  # No rate available
-    ('sell', 'ask', None, 4, 4,  0, 4),  # ask not available
-    ('sell', 'ask', None, None, 4,  0, 4),  # ask not available
-    ('sell', 'bid', 6, None, 4,  0, 5),  # bid not available
-    ('sell', 'bid', None, None, 4,  0, 5),  # bid not available
-])
+get_ticker_rate_error = [
```
xmatthias: You might wanna combine the definitions of these variables (and move them at the top).

we've now got 3 - spread throughout the file - which will make it difficult to find them should we need them (and it's unclear that they belong together).

tests/test_freqtradebot.py
xmatthias: having more tests this way will not improve readability but reduce that - as the action -> corresponding result are on different lines.

While using zip treats how you work with the data - my point here is about the definition per se - which is unclear as is.

freqtrade/exchange/exchange.py
xmatthias: Please don't write "updated" if that's not true - while you changed it for the "none" assignment (The one that was kind of OK as it's just initializing the variable) - it's not changed on this line (where i see it as worse.

tests/exchange/test_exchange.py
```diff
@@ -2081,6 +2091,76 @@ def test_get_sell_rate_exception(default_conf, mocker, caplog):
...
+    api_mock = MagicMock()
+    api_mock.fetch_l2_order_book = order_book_l2
+    api_mock.fetch_ticker = MagicMock(
+                 return_value={'ask': ask, 'last': last, 'bid': bid})
+    exchange = get_patched_exchange(mocker, default_conf, api_mock)
+
+    pair = "ETH/BTC"
+
+    # Test regular mode
+    rate = exchange.get_rate(pair, refresh=True, side="sell")
```
xmatthias: i think your intend here is to call `get_rates()` (the new method) - not `get_rate()`.

happens throughout this test ... 
I'm not sure if it wouldn't make more sense to combine this with the other test though - as get_rates is providing both buy and sell price, it might make more sense to test them together (our expectation is in the end to get both results at the same time).

freqtrade/optimize/backtesting.py
```diff
@@ -383,14 +383,17 @@ def _get_close_rate(self, sell_row: Tuple, trade: LocalTrade, sell: SellCheckTup
 
     def _get_adjust_trade_entry_for_candle(self, trade: LocalTrade, row: Tuple
                                            ) -> LocalTrade:
+        current_entry_rate = current_exit_rate = row[OPEN_IDX]
+        current_rate = current_entry_rate
```
xmatthias: There's little point in this assignment.
Best assign to one variable (probably `current_rate`) - and use that throughout the method.

using 3 names for the same thing will not improve clarity (the only reason to do this) - but actually reduce it - as it's unclear at first glance at the wrapper lines that we're passing in the same value 3 times.

freqtrade/optimize/backtesting.py
```diff
@@ -399,6 +402,16 @@ def _get_adjust_trade_entry_for_candle(self, trade: LocalTrade, row: Tuple
                 self.wallets.update()
                 return pos_trade
 
+        if stake_amount is not None and stake_amount < 0.0:
+            amount = -stake_amount / current_rate
```
xmatthias: Please use `abs()` here as well

freqtrade/optimize/backtesting.py
```diff
@@ -399,6 +402,16 @@ def _get_adjust_trade_entry_for_candle(self, trade: LocalTrade, row: Tuple
...
+            if amount > trade.amount:
+                logger.info(f"Amount is higher than available. {amount} > {trade.amount}")
```
xmatthias: Please abstain from logging in backtesting.

it'll be meaningless log messages (you're testing 2000 trades, and will not know what you tested with), and will slow down hyperopt (output is a deal breaker there, as the terminal is usually the slowest part).

freqtrade/optimize/backtesting.py
```diff
@@ -467,32 +480,38 @@ def _get_sell_trade_entry_for_candle(self, trade: LocalTrade,
...
+            ft_order_side="sell",
+            side="sell",
+            order_type=order_type,
+            status="open",
+            price=closerate,
+            average=closerate,
+            amount=amount or trade.amount,
+            filled=0,
+            remaining=trade.amount,
+            cost=trade.amount * closerate,
```
xmatthias: if you use `amount or trade.amount` above - the same should be used for the calculation of cost.

freqtrade/freqtradebot.py
```diff
@@ -1234,24 +1262,38 @@ def execute_trade_exit(
...
         Sends rpc notification when a sell occurred.
         """
-        profit_rate = trade.close_rate if trade.close_rate else trade.close_rate_requested
-        profit_trade = trade.calc_profit(rate=profit_rate)
         # Use cached rates here - it was updated seconds ago.
         current_rate = self.exchange.get_rate(
             trade.pair, refresh=False, side="sell") if not fill else None
-        profit_ratio = trade.calc_profit_ratio(profit_rate)
+
+        # second condtion is for mypy only; order will always be passed during sub trade
```
xmatthias:
```suggestion
        # second condition is for mypy only; order will always be passed during sub trade
```

freqtrade/freqtradebot.py
```diff
@@ -612,14 +631,17 @@ def execute_entry(self, pair: str, stake_amount: float, price: Optional[float] =
             trade.open_order_id = order_id
 
         trade.orders.append(order_obj)
-        trade.recalc_trade_from_orders()
+        if pos_adjust:
+            trade.recalc_trade_from_orders()
+        else:
+            trade.recalc_open_trade_value()
```
xmatthias: This one leaves me with a questionmark for myself (i think it's wrong / pointless).

i think `trade.recalc_trade_from_orders()` should also work for the initial entry - so why do we need to split this?
More specifically - once the order fills, we call `update_trade()` - which for buy orders - will call  `trade.recalc_trade_from_orders()` anyway.

tests/rpc/test_rpc_telegram.py
```diff
@@ -698,7 +708,9 @@ def test_monthly_handle(default_conf, update, ticker, limit_buy_order, fee,
 
     trades = Trade.query.all()
     for trade in trades:
-        trade.update_trade(oobj)
+        trade.open_rate = oobj.safe_price
+        trade.amount = oobj.safe_amount_after_fee
```
xmatthias: I do fear that this will lead to more problems down the road.

A proper method (using update_trade/update_order) seems to work for other tests?

SmartManoj: updated

freqtrade/freqtradebot.py
SmartManoj: updated

freqtrade/freqtradebot.py
SmartManoj: Yes, it is not needed here. removed

freqtrade/optimize/backtesting.py
SmartManoj: updated

freqtrade/optimize/backtesting.py
SmartManoj: updated

freqtrade/optimize/backtesting.py
SmartManoj: updated

freqtrade/optimize/backtesting.py
SmartManoj: updated

tests/exchange/test_exchange.py
SmartManoj: updated. 
Then we need to make data for the 4 variations. Wrote the test with intend of checking the API call count.

tests/exchange/test_exchange.py
SmartManoj: moved to top. expected value is also there. So can't combine them.

freqtrade/exchange/exchange.py
SmartManoj: not seen that. updated

tests/test_freqtradebot.py
SmartManoj: updated

freqtrade/optimize/backtesting.py
SmartManoj: updated

tests/exchange/test_exchange.py
xmatthias: That was expected :+1:

tests/exchange/test_exchange.py
xmatthias: As said above - i'm not sure ;) It's now testing `get_rates()` - so that's fine i'd think :)

freqtrade/persistence/models.py
```diff
@@ -630,27 +660,24 @@ def calc_profit_ratio(self, rate: Optional[float] = None,
...
-                    (o.status not in NON_OPEN_EXCHANGE_STATES)):
+            if o.ft_is_open or not o.filled:
                 continue
 
             tmp_amount = o.safe_amount_after_fee
-            tmp_price = o.average or o.price
-            if o.filled is not None:
-                tmp_amount = o.filled
+            tmp_price = o.safe_price
+            is_sell = o.ft_order_side != 'buy'
```
xmatthias: Unfortunately this overlapps with the merge of `feat/short` - so there's now quite some merge conflicts - so please merge develop into your branch and fix merge-conflicts as they come up.

The problematic points will however not be the overlaps (that's where you look at) - but areas like this line (exemplary - but can happen to all other lines as well).

After the merge - please go over the "still changed lines" . Areas of focus should go to lines like this (which previously used `'buy'` or `'sell'`) - which should be updated to `trade.entry_side` and `trade.exit_side` respectively).

I don't think this will appear in too many cases (at least from a quick glance) - but it's still something to be on the lookout for to avoid nasty, difficult to track bugs (you'd increase the position instead of decreasing, potentially).

SmartManoj: https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/persistence/models.py

`entry_side` is only used in `Order.to_json()` ; not present in `Trade` object

`entry_side = 'sell' if self.is_short else 'buy'`. Is this right?

xmatthias: the name is `enter_side` ... my bad.

SmartManoj: enter_side , entry_side (used in Order.to_json() only), entry_rate
Which is more accurate and consistence with other vars - enter or entry?

xmatthias: unfortunately, both are used at the moment - and with a merge as big as the feat/short one - it was simply missed.

freqtrade/exchange/gateio.py
```diff
@@ -39,6 +42,11 @@ class Gateio(Exchange):
         (TradingMode.FUTURES, MarginMode.ISOLATED)
     ]
 
+    def _init_ccxt(self, exchange_config: Dict[str, Any], ccxt_module: CcxtModuleType = ccxt,
+                   ccxt_kwargs: Dict = {}) -> ccxt.Exchange:
+        ccxt_kwargs.update({"timeout": 15000})
+        return super()._init_ccxt(exchange_config, ccxt_module, ccxt_kwargs)
+
```
xmatthias: I don't see why this would be relevant / related to this PR, please remove it.

Gateio's api performs quite well and fast in my tests - if you're having timeout problems, then it's most likely a network problem - not something we will enforce at a freqtrade level.
You can configure this in the settings - but clearly not here.

mkavinkumar1: https://github.com/freqtrade/freqtrade/runs/5818814987?check_suite_focus=true

xmatthias: Most likely, this suggests that gate.io blocked the github actions IP range.
Increasing the timeout for all users will not be the solution.

Either way - it's not related to this PR - therefore please revert these changes.

If this is a common problem  (one that can be reproduced independent of CI) - then please file an issue with ccxt.
we'll not mess with timeouts in freqtrade in this regards.

SmartManoj: Happened in GCP too. Will fill an issue at `ccxt`.  Currently have no logs.

xmatthias (changes requested):
Please avoid commiting .orig files.

It seems to be a side-effect of merging (?) - so i'd suggest to use a proper editor (e.g. vscode) which will support you in comparing files cross-branch (e.g. "compare working tree with develop").

SmartManoj:
> Please avoid commiting .orig files.
> 
> It seems to be a side-effect of merging (?) - so i'd suggest to use a proper editor (e.g. vscode) which will support you in comparing files cross-branch (e.g. "compare working tree with develop").

Using Meld. [git mergetool add those.](https://stackoverflow.com/questions/1251681/git-mergetool-generates-unwanted-orig-files). Disabled

.gitignore
```diff
@@ -1,107 +1,110 @@
...
+user_data/notebooks/*
+freqtrade-plot.html
+freqtrade-profit-plot.html
+freqtrade/rpc/api_server/ui/*
+
+# Macos related
+.DS_Store
+
+# git files after merging
+*.orig
```
xmatthias: this is something that's specific to you / your workflow - not to git.
We'll not add custom problems here - you can exclude them in your personal / per device gitignore.

Therefore please revert this change.

Please avoid unnecessary changes (like this one). it's wasting both our time - mine to look at it - your's to revert the changes.

> Using Meld. [git mergetool add those.](https://stackoverflow.com/questions/1251681/git-mergetool-generates-unwanted-orig-files). Disabled

Then use a better tool which doesn't leave artifacts around - or be careful what you stage.

we'll not be adding random gitignore stuff because some tool is missbehaving.

Apart from doing a unwanted change to gitignore - you didn't actually remove the orig files.

.pre-commit-config.yaml
```diff
@@ -11,7 +11,35 @@ repos:
...
+        files: ^scripts/
+        entry: mypy scripts
+        additional_dependencies: 
+        -    types-cachetools==5.0.0
+        -    types-filelock==3.2.5
+        -    types-python-dateutil==2.8.10
+        -    types-requests==2.27.15
+        -    types-tabulate==0.8.6
+        -    types-urllib3==1.26.11
+
```
xmatthias: Pleas don't mix unmerged, rejected PR's.

.gitignore
```diff
@@ -15,6 +15,9 @@ freqtrade/rpc/api_server/ui/*
 # Macos related
 .DS_Store
 
+# git files after merging
+*.orig
+
```
xmatthias: Please fix your tooling to not keep backups.
That's a sign of a miss-configured tooling on your side (we recommend vscode or pycharm - neither of the two will write these files).

As such - there's no reason to exclude this in freqtrade's gitignore file.
Also make sure to remove all `.orig` files.

xmatthias (changes requested):
@SmartManoj If you expect me to look at this PR - please clean it up and do a review of all changes yourself first, and **remove all things that do not belong here** (everything that is not connected to adding this feature does NOT belong into this PR).

I'm pretty sure you'll see an additional thing or tho if you think through all the changes.

Once you're done with the above (and there's no merge conflicts), please click the "re-request a review" button at the top right to request another review to confirm that what was requested has been changed (throughout the PR - this does NOT only apply to lines where i commented).

SmartManoj:
9f3be72 is a temporary commit. Without that, pre-commit not worked. Will remove ASAP.

.gitignore
SmartManoj: https://stackoverflow.com/questions/12366150/how-to-delete-orig-files-after-merge-from-git-repository#comment62315290_12366207

Isn't?

xmatthias: I don't know what tooling you use that it keeps .orig files (i've never seen that, and i've used quite some merge-tools over the years).

that thread seems to suggest using `git config --global mergetool.keepBackup false` (https://stackoverflow.com/a/34149790/1770304) - but i'm not sure if this fixes it as it's dependent on your setup.

Even if we were to change the gitignore (we're not) - this PR is clearly not the right place for such a change.

xmatthias:
> Will remove ASAP.

@mkavinkumar1 @SmartManoj Are you planning to address the requested changes from [here](https://github.com/freqtrade/freqtrade/pull/6545#pullrequestreview-937466723) - which basically means you do a review yourself (look through all changed files) - and the new diff (github PR -> files) should **only contain changes which are actually desired** - all other undesired/unrelated changes are rolled back to their initial state, undesired files are removed etc.

I expect the next push to this branch to have a clean state - which makes it possible to review, and potentially "mergable" assuming no additional things are found.

Otherwise i'll consider this PR as abandoned / assume that you have no intend to actually contribute this - as that would mean you're ignoring requested changes.

SmartManoj:
Yes, after #6779 and resolving merge conflicts. On which order you will review the files?

xmatthias:
> On which order you will review the files?

The order of review will not matter for the PR - everything interlocks / needs to play together - therefore i expect everything to be ready for review.
I think the main parts that are currently open are the unwanted files / unwanted changes. Do a review of all files currently in the PR yourself - and make sure the diff is clean (as requested twice above, now).

---

Simply put - imagine yourself in the Reviewers / maintainers shoes and determine - "what is missing if i were to merge this (and with that, accept responsibility over the code i just merged)."

I think this will show you a great deal of "oh, this change is unnecessary - or "this commented statement was just there for debugging purposes, but isn't relevant anymore" - and will avoid a lot of (really unnecessary) requested changes - which are easily spottable if you look at the things you changed before actually pushing/submitting a PR.
I suggest to do this for every PR - you'll see that your PR's will improve (fewer changes requested) through this mindset.

You can do this quite easily even before doing a PR with vscode's "gitlense" extension, which allows a "compare working tree with ..." command - showing all changes - but also github itself provides a very good view for this.

SmartManoj:
I think there is no need to remove the unrelated changes if it is in upstream as the base is changed when merging upstream?
#6792

xmatthias:
If it is in upstream (part of https://github.com/freqtrade/freqtrade - develop) - then no (so it's aligned to the target branch).

If it is a pending PR - then it is unrelated to this PR until it's merged - and as such - shall not be included here.
Even if you think it's a minor change - there may be changes requested - or the PR might be rejected if we consider the change to be not good. 


A PR should stand "on it's own" - and it's changes will be considered based on the diff as is.
As you'll not know if/when other PR's are merged, including them in another PR is bad practice.

SmartManoj:
Waiting for #6823

SmartManoj:
@anyone Feel free to reuse the code if you feel the PR is moving slowly.

freqtrade/freqtradebot.py
```diff
@@ -511,38 +512,59 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
...
+            self.execute_entry(trade.pair, stake_amount, None and current_entry_rate,
+                               trade=trade, is_short=trade.is_short)
 
         if stake_amount is not None and stake_amount < 0.0:
             # We should decrease our position
-            # TODO: Selling part of the trade not implemented yet.
-            logger.error(f"Unable to decrease trade position / sell partially"
-                         f" for pair {trade.pair}, feature not implemented.")
+            # Strategy should return value as Decimal for accuracy.
+            amount = abs(float(Decimal(stake_amount) / Decimal(current_exit_rate)))
```
xmatthias, line 568: While the calculations themselfs are covered by tests - you're missing tests for the most critical part - which is this whole if block, as well as the corresponding one in backtesting - where the strategy method is interfacing with the (tested) logic.

For backtesting - i'd recommend to simply add tests to `test_backtest_detail.py` which cover this scenario.

freqtrade/exchange/exchange.py
```diff
@@ -1444,8 +1444,9 @@ def fetch_l2_order_book(self, pair: str, limit: int = 100) -> dict:
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 
-    def get_rate(self, pair: str, refresh: bool,
-                 side: EntryExit, is_short: bool) -> float:
+    def get_rate(self, pair: str, refresh: bool,  # noqa: max-complexity: 13
```
xmatthias: Please don't noqa methods. If you need to add complexity, consider refactoring them to reduce complexity you added. 
This doesn't apply only here, but also in other cases where this was added.

freqtrade/persistence/trade_model.py
```diff
@@ -1017,6 +1068,7 @@ class Trade(_DECL_BASE, LocalTrade):
     open_trade_value = Column(Float)
     close_rate: Optional[float] = Column(Float)
     close_rate_requested = Column(Float)
+    realized_profit = Column(Float, default=0.0)
```
xmatthias, line 1090: For added columns - please make sure they're also part of the migration (and that the migration triggers).
otherwise this will: 
* fail on opening the first trade with an existing database
* fail when the next migratio is triggered (this column will not be migrated)

xmatthias (commented):
This PR is now clearly moving in the right direction :+1: - but it's still some ways to go.

I've only looked at a few files now - but what i've seen looks mostly fine.

freqtrade/optimize/backtesting.py
```diff
@@ -280,8 +280,8 @@ def load_bt_data_detail(self) -> None:
 
             if unavailable_pairs:
                 raise OperationalException(
-                        f"Pairs {', '.join(unavailable_pairs)} got no leverage tiers available. "
```
SmartManoj, line 290: @xmatthias Should I revert style changes too?

freqtrade/freqtradebot.py
SmartManoj: Any test cases added for rebuys in `test_backtest_detail.py`?

freqtrade/optimize/backtesting.py
xmatthias: it doesn't belong here - HOWEVER - it's just one line, so it's fine here.

It is however very problematic if a whole file changes (most every line) due to formatting changes - as it makes identifying "real" changes near impossible / very timeconsuming - drawing away the focus from reviewing the changed logic (and how it ties into all the rest of the bot) to "is here a change in logic - or just in formatting".

freqtrade/freqtradebot.py
xmatthias: backtesting rebuy testcase is in `test_backtesting_adjust_position.py`.
i'm however not sure if that'll be easier or more difficult - as this relies on the return value from the TestStrategy - so messing with that MAY change a lot of other tests, too.

SmartManoj: trade details for `StrategyTestV2`?

xmatthias: seems so - but it's better to use StrategyTestV3  - as that's the new strategy format. I'll update the existing test to that as well.

freqtrade/freqtradebot.py
```diff
@@ -512,39 +513,59 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
...
-            current_profit=current_profit, min_stake=min_stake_amount,
-            max_stake=min(max_stake_amount, stake_available))
+            trade=trade, current_time=datetime.now(timezone.utc), current_rate=current_entry_rate,
+            current_profit=current_entry_profit, min_stake=min_entry_stake,
+            max_stake=min(max_entry_stake, stake_available),
+            current_entry_rate=current_entry_rate, current_exit_rate=current_exit_rate,
+            current_entry_profit=current_entry_profit, current_exit_profit=current_exit_profit,
+            min_entry_stake=min_entry_stake, min_exit_stake=min_exit_stake,
+            max_entry_stake=min(max_entry_stake, stake_available),
+            max_exit_stake=min(max_exit_stake, stake_available)
```
xmatthias, line 541: This is a very dangerous change the way it is now.
Not because of the parameters, but because it's done completely inconsistently.
backtesting does not provide all parameters (so strategies relying on that will malfunction)- and none of the new parameters are in the documentation, nor in the function header (`interface.py`)

This method now has way too many parameters. It's not useful this way.

`max_exit_stake` is automatically the current trade amount  (so the parameter is irrelevant).
`min_exit_stake` should also be removed - it's simply `min_stake` - the difference in rate is too minimal to matter, so this only clutters the interface.

i also don't think the `current_*_profit` and `current_*_rate` parameters are necessary.
Depending on the config, their value will be identical - and having this many parameters will only confuse people using this.

freqtrade/optimize/backtesting.py
```diff
@@ -494,16 +494,20 @@ def _get_close_rate_for_roi(self, row: Tuple, trade: LocalTrade, exit: ExitCheck
...
         stake_available = self.wallets.get_available_stake_amount()
         stake_amount = strategy_safe_wrapper(self.strategy.adjust_trade_position,
                                              default_retval=None)(
             trade=trade,  # type: ignore[arg-type]
-            current_time=row[DATE_IDX].to_pydatetime(), current_rate=row[OPEN_IDX],
+            current_time=row[DATE_IDX].to_pydatetime(), current_rate=current_rate,
             current_profit=current_profit, min_stake=min_stake,
-            max_stake=min(max_stake, stake_available))
+            max_stake=min(max_stake, stake_available),
+            current_entry_rate=current_rate, current_exit_rate=current_rate,
```
xmatthias, line 518: Inconsistent parameters across the bot.

freqtrade/persistence/trade_model.py
```diff
@@ -612,6 +631,35 @@ def update_trade(self, order: Order) -> None:
             raise ValueError(f'Unknown order type: {order.order_type}')
         Trade.commit()
 
+    def process_exit_sub_trade(self, order: Order, is_closed: bool = True) -> None:
+        exit_amount = order.safe_amount_after_fee
```
xmatthias: Note to self: This is missing a test.

rokups:
I cant really make sense of `rel_profit` in `test_recalc_trade_from_orders_dca`. For simplicity lets look at no-fees test:
```
{
    # tuple 1 - side, amount, price
    # tuple 2 - amount, open_rate, stake_amount, cumulative_profit, realized_profit, rel_profit
    'orders': [
        (('buy', 100, 10), (100.0, 10.0, 1000.0, 0.0, None, None)),
        
        // ✓ Avg price (100 * 10 + 100 * 15) / (100 + 100) = 12.5
        // ✓ Total stake: 100 * 10 + 100 * 15 = 2500                 
        (('buy', 100, 15), (200.0, 12.5, 2500.0, 0.0, None, None)),
        
        // ✓ Total stake: 2500 - 50 * 12.5 = 1875
        // ✓ PnL: (12 - 12.5) * 50 = -25
        // x Order Profit: ((12 - 12.5) / 12.5) * (50 / 200) = -0.01
        (('sell', 50, 12), (150.0, 12.5, 1875.0, -25.0, -25.0, -0.04)),
        
        // ✓ Total stake: 1875 - 100 * 12.5 = 625
        // ✓ PnL: (20 - 12.5) * 100 = 750
        // ✓ Cumulative profit: -25.0 + 750 = 725
        // x Order Profit: ((20 - 12.5) / 12.5) * (100 / 200) = 0.3
        // x Total Profit: -0.01 + 0.3 = 0.29
        (('sell', 100, 20), (50.0, 12.5, 625.0, 725.0, 750.0, 0.56)),

        // ✓ Total stake: 625 - 50 * 12.5 = 625
        // ✓ PnL: (5 - 12.5) * 50 = -375
        // x Order Profit: ((5 - 12.5) / 12.5) * (50 / 200) = -0.15
        // x Total Profit: -0.01 + 0.3 - 0.15 = 0.14
        (('sell', 50, 5), (50.0, 12.5, 625.0, 725.0, -375.0, 0.56)),
    ],
    // Total spent: 2500.0
    // Total received: 600 + 2000 + 250 = 2850
    // Total profit: 2850 - 2500 = 350
    // Total profit percent: 1 - (2850 / 2500) = 0.14
    'end_profit': 350.0,
    'fee': 0.0,
    'end_profit_ratio': -0.60,  // What is that?
},
```
Bits marked with `x` are my calculations that i think are correct, but they do not match what is in the test. It seems to me `rel_profit` should be % profit relative to _part_ of the position sold, not relative to entire position. Following that logic, in the end we can just add up % PnL of all orders and we arrive at the same total profit as it is calculated from absolute dollar amounts of all buys and sells. In this example trade makes 14% profit and this is what we should get when we add profit % of all sell trades i would say.

xmatthias:
Thanks for having a look at this!
well the current relative profit calculation assumes the following:
`// Order Profit: ((12 - 12.5) / 12.5) = -0.04` (without considering full stake).


> `// x Total Profit: -0.01 + 0.3 - 0.15 = 0.14`

i find this questionable - you're aggregating relative profits - which may be on the same level (say - all 3 were -50 amount).

> `// x Order Profit: ((20 - 12.5) / 12.5) * (100 / 200) = 0.3`

Can you explain the reason for the `(100 / 200)` part? i assume the 200 is the total stake we ever owned.
I think that's problematic - you can't assume a buy buy sell sell sell sequence.

if we look at the 3rd test (i'll make a copy without fees later) - what would be the value to use for the 2nd sell? (after the partial rebuy) - would we need to use 200 (we never owned more than 200)  - or 300 (because all buys combined were 300).

rokups:
For the sake of history: as already mentioned on discord `(100 / 200)` is a fraction of total stake that was sold, so profit is adjusted to be relative to only sold part. My reasoning for that is this: consider you do a partial sell for profit of X% and another partial sell at break-even price so for 0% profit. In that case `(100 / 200)` adjustment makes sense as it gives us real profit of that particular order which happens to be final profit of entire trade.

You are also right that with interleaved buys and sells its not so easy ;) I suppose we can look at "relative profit" from different angles and we may both not be wrong here. You were right we cant just add up profit % when buys and sells are interleaved as well, but maybe we dont have to...

In the end i think we should clearly define what is `rel_profit` and what is `end_profit_ratio`. I can not quite make sense of these numbers at the moment.

My interpretation is this:
- `rel_profit` should be % profit of each individual order. `((order_sell_price - avg_price) / avg_price) * (coin_sell_amount / coin_current_amount)`
- `end_profit_ratio` should be total % profit calculated independently of `rel_profit`. `(total_stake_received / total_stake_spent) - 1`

New no-fees test is very helpful indeed here. If you look carefully, total USD spent is 3250 and total USD received is 6450. Test correctly states 3200 USD profit. However `end_profit_ratio = 1.09`. What is this ratio of? Trade clearly made 98% profit. I just cant connect this ratio to anything.

xmatthias:
`end_profit_ratio` is clear wrong - especially in the first test - where absolute profit is positive, but relative profit is strongly negative.
It's curently calculated as the relative profit of the last order - so for the first testcase - `(5 - 12.5) / 12.5`.
now that's obviously wrong ...

xmatthias (approved):
I think this PR is now as good as it gets - further potential issues will be fixed during the develop period.

As this PR contains too many commits (and too many of them with pointless changes which had to be reverted) - it will be squash-merged, to keep freqtrade history somewhat clean.

LGTM :+1:

PR #6548, opened 2022-03-17
Change how close trade value calculated

stash86:

## Summary

Change how close trade value calculated, to prepare for partial sell

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

Close trade value now gonna be calculated as filled exit order + (unsold coin amount x current rate )


freqtrade/persistence/models.py
```diff
@@ -590,9 +590,19 @@ def calc_close_trade_value(self, rate: Optional[float] = None,
...
 
-        sell_trade = Decimal(self.amount) * Decimal(rate or self.close_rate)  # type: ignore
-        fees = sell_trade * Decimal(fee or self.fee_close)
-        return float(sell_trade - fees)
+        filled_exit_orders = self.select_filled_orders('sell')
+        filled_amount = 0
+        filled_exit_trade = 0
+        for order in filled_exit_orders:
+            filled_amount += order.filled
+            filled_exit_trade += (order.price * order.filled)
```
xmatthias: Please use safe_filled / safe_price ... you can't 100% rely on filled being there.

xmatthias (commented):
To be honest, i'm not sure i understand this change / what you aim to archive with this.

The calculation will "not really" change in most cases - and there is no handling of "what's left", so it seems kinda pointless at this stage. 
The only case where this may change something is when small amounts are left (dust) - but i'm not sure on the impact this will have to this instance - in unfortunate cases, it may drag your profit way down - while in reality, you may have 8$ of dust left to "just" convert to bnb.


As it's touching calculations - i'd also want to have tests which show the expected new behaviour - which i guess will differ in some instances (where clearly no tests exist so far).

stash86:
`remaining_amount = self.amount - filled_amount` to track unsold coin

The new calculation will follow this equation of P/L at any time 
`P/L = sum of filled exit + (unsold coin * current rate) - open_trade_value` which is different from current formula of `P/L = (amount * current rate) - open_trade_value`

xmatthias:
> The new calculation will follow this equation of P/L at any time

that's correct - but it's a state you can't get to at the moment.
once a sell trade is filled - the order is closed.

Your query will not give you partially filled sell orders which are still open - only once they are closed.

partial sell orders cannot be canceled - so the only case where this will actually apply is once the trade is filled - but is lacking the "dust" part which it was forced to leave behind.
It may have an impact if you cancel the partially filled order on exchange - but i'm not sure that'll be handled properly elsewhere.

stash86:
Buy 10 coins @$10
Buy 20 coins @$7
open_trade_value = 240

When current rate is $6, P/L = 0 + (30 x 6) - 240 = -$60
When current rate is $9, sell 15 coins. P/L = (15 x 9) + (15 x 9) - 240 = $30
The price drop to $7, current P/L = (15 x 9) + (15 x 7) - 240 = 0
The price drop again to $5, current P/L = (15 x 9) + (15 x 5) - 240 = -$30
The price rose to $20 and sell the remaining, final P/L = (15 x 9 + 15 x 20) + 0 - 240 = $195

stash86:
Not really looking at partially filled limit order currently, only forced partial sell. Current PR won't affect any partially filled limit sell order

stash86:
And the "dust" part should be covered in the formula, on the `unsold amount x current rate` part.

Ofc when a trade marked as closed, current_rate can be changed to final sell rate if you want to

xmatthias (commented):
I'd want to put this on hold for now - as it seems to be adding partial sells as well in the last commits.

#6545 is further in this regards (it also updates all notifications, ...) - so while that PR is not complete - i don't think it makes sense to do the same work twice - as it would have to lead to the rejection of one of the 2 PR's (and therefore wasted time).

maybe a better alternative is to have a look at the other PR and help identify potential issues - things we haven thought about yet.

PR #6553, opened 2022-03-19
Add proposed leverage for the judgment of min_stake_amount and max_stake_amount

adrianceding:

## Summary

Add proposed leverage for the judgment of min_stake_amount and max_stake_amount

## Quick changelog

- Use leverage in config to assist in judging min_stake_amount and max_stake_amount

## What's new?

When balance is less than min_stake_amount. And xleverage is greater than min_stake_amount later. You can set leverage in the config file to adjust min_stake_amount and max_stake_amount. This is useful in small amount tests

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -600,8 +600,9 @@ class FreqtradeBot(LoggingMixin):
         trade_side = 'short' if is_short else 'long'
         pos_adjust = trade is not None
 
+        proposed_leverage = self.config.get('leverage', 1.0)
         enter_limit_requested, stake_amount = self.get_valid_enter_price_and_stake(
-            pair, price, stake_amount, side, trade_side, enter_tag, trade)
+            pair, price, stake_amount, side, trade_side, enter_tag, trade, proposed_leverage)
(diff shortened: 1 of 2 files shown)


freqtrade/freqtradebot.py
```diff
@@ -600,8 +600,9 @@ def execute_entry(
         trade_side = 'short' if is_short else 'long'
         pos_adjust = trade is not None
 
+        proposed_leverage = self.config.get('leverage', 1.0)
```
xmatthias, line 603: rather than adding an additional config property - i would revert the sequence of calls - so "leverage" is called before stake-amount (that was my idea for this problem)

We can take the "assumed" stake-amount (already available based on default settings) - and use that to calculate min/max leverage.
if we look at the leverage brackets - these are usually multiple 1000$ wide - so it's more than doubtful that your default calculation says "20$" - and your custom_stake_amount returns 20.000$ (which would drop it in different max-leverage buckets).

on binance for example, a sample bucket would be 0-5000, 5000-25000, 25_000- 100_000 - so i think the gap is wide enough to take a simple guess.

PR #6558, opened 2022-03-20
Recursively search subdirectories in config['user_data_dir']/strategies for a strategy

samgermain:
## Summary

Recursively search subdirectories in user_data/strategies for a strategy

## Quick changelog

- `IResolver.build_search_paths` takes a list of extra directories in `extra_dirs` instead of just a single extra directory
- `StrategyResolver._load_strategy` recursively finds the subdirectories in `user_data/strategies` and passes them to `IResolver.build_search_paths` via  `extra_dirs`

## What's new?

Strategy files can be in a sub-directory inside `config['user_data_dir']/strategies`

e.g.

`user_data/strategies/old`
`user_data/strategies/private`
`user_data/strategies/berlinguyinca`
`user_data/strategies/discord`
`user_data/strategies/freqtrade-strategies/user_data/strategies`
`user_data/strategies/lookahead_bias`

--------------

For me personally, this will be really helpful, and keep me from having duplicate strategy files, and I imagine I'm not the only one who might have a problem like this


samgermain:
I didn't write any tests, because I couldn't find tests for `StrategyResolver._load_strategy` or `IResolver.build_search_paths`, but I tried it out and it works

xmatthias:
> For me personally, this will be really helpful, and keep me from having duplicate strategy files, and I imagine I'm not the only one who might have a problem like this

I'm not sure how this would keep you from having duplicate files.
Assuming you have a strategy - `StrategyA` - which is stored in `strategies/StrategyA.py` - then this PR would allow you to have the same strategy in `strategies/something/StrategyA.py` - so the filename is identical.
The file is however still a duplicate.
You should anyway name the strategies differently (unique strategy names) - otherwise you'll have problems with the resolver anyway - as you can't rely on which file is loaded.

---

I'm actually quite worried about the impact of this change. 
Assuming people have a "old" directory - with copies of their strategies - we'd now detect both of these strategies - and would load "one" of them - most likely without the user noticing.

samgermain:
> I'm not sure how this would keep you from having duplicate files.

This is my strategies folder, a few of those folders are github repos, but freqtrade can't find the strategy unless they're one level up, so I need to move the strategies up, but they're still part of the repo, so they need to be in there.

I didn't realize there was an extra_dir argument beforehand. 

I'm trying to add a `-r` option for recursive strategy search, but I'm having trouble figuring out how to access the variable

xmatthias:
you can simply use `--strategy-path` for this - specifying `--strategy-path user_data/strategies/berlinguyinca` (or whatever other dir) to load strategies from that folder.

samgermain:
> you can simply use `--strategy-path` for this - specifying `--strategy-path user_data/strategies/berlinguyinca` (or whatever other dir) to load strategies from that folder.

Yeah but I like the recursive option more, it's a lot less tedious, with the `-r` option, it would only search recursively if you sepecify `-r`

freqtrade/commands/cli_options.py
```diff
@@ -82,6 +82,11 @@ def __init__(self, *args, **kwargs):
         help='Reset sample files to their original state.',
         action='store_true',
     ),
+    "recursive": Arg(
```
xmatthias: you'll have to add this to `configuration.py` "someplace" (just look for a spot near the other strategy options).

options will not automatically transfer to the configuration - it's a one-line addition though.

samgermain:
I'm now getting
```
main.py: error: unrecognized arguments: -r
```
and
```
  File "/path/freqtrade/freqtrade/resolvers/strategy_resolver.py", line 170, in _load_strategy
    if config['recursive_strategy_search']:
KeyError: 'recursive_strategy_search'
```

xmatthias:
You're not assigning it to any command.

you'll need to be careful about "is that key there at all times" too (which is why you get the key error).
There's routes which won't guarantee a full configuration (for example when loading a strategy in a jupyter notebook).

while we can require some keys to be there if they are necessary - this is not one of these.

xmatthias:
You're not assigning it to any command.

you'll need to be careful about "is that key there at all times" too (which is why you get the key error).
There's routes which won't guarantee a full configuration (for example when loading a strategy in a jupyter notebook).

while we can require some keys to be there if they are necessary - this is not one of these.

freqtrade/commands/cli_options.py
```diff
@@ -82,6 +82,11 @@ def __init__(self, *args, **kwargs):
         help='Reset sample files to their original state.',
         action='store_true',
     ),
+    "recursive_strategy_search": Arg(
+        '-r', '--recursive_strategy_search',
```
xmatthias: I'd not give this command a "1 letter alias" 

if you rely on this, you'll usually want to have it in the configuration so it always applies.
For consistency, the command should also be `--recursive-strategy-search` (with dash, not underscore), as all other arguments also use it that way.

Obviously the configuration setting will remain as is.

samgermain:
> You're not assigning it to any command.
> 
> you'll need to be careful about "is that key there at all times" too (which is why you get the key error). There's routes which won't guarantee a full configuration (for example when loading a strategy in a jupyter notebook).
> 
> while we can require some keys to be there if they are necessary - this is not one of these.

I can't figure out where I would edit the config, in Arguments._parse_args I could maybe assign ` parsed_arg.config['recursive_strategy_search']` but that's all I've got

xmatthias:
I think you're looking for the wrong thing.

`--recursive-strategy-search` will not be available for any (sub)command. 
While you can assign it in the configuration directly - you can't run `freqtrade trade --recursive-strategy-search` at this point.

For this - you need to assign it to the appropriate commands in `arguments.py`

freqtrade/commands/arguments.py
```diff
@@ -12,7 +12,7 @@
 
 ARGS_COMMON = ["verbosity", "logfile", "version", "config", "datadir", "user_data_dir"]
 
-ARGS_STRATEGY = ["strategy", "strategy_path"]
+ARGS_STRATEGY = ["strategy", "strategy_path", "recursive_strategy_search"]
```
xmatthias, line 15: This is what you were missing.

in general, i'd however think that this is mostly a "config" setting - which you set in your config and forget about.
you'll usually not want to run `--reqcursive-strategy-search` with every command - assuming that's your standard way to work.

freqtrade/commands/cli_options.py
xmatthias: The argument name has not been changed - so it's now the only argument to use `_` in it's name.

Also, did you actually test this with a command line? 

`freqtrade trade  --recursive_strategy_search -s SampleStrategy` yields `freqtrade trade: error: argument --recursive_strategy_search: expected one argument`.

freqtrade/commands/arguments.py
xmatthias, line 15: I think this should be added to strategy-list as well (below) - which defines this separately (as it can't take a `strategy` argment).

samgermain: What do you mean exactly? It looks like when looping through the strategy_list in `backtesting.__init__` and `backtesting.start`, `load_strategy` and then `_load_strategy` is called, which already accounts for `recursive_strategy_search`

xmatthias: the command where it's missing is `freqtrade list-strategies`. That's a few lines below this one ... just search for `strategy_path` in this file.

xmatthias (approved):
I hope i got them all - but this is what i meant.

`list-strategies` must also support this - otherwise it'll be highly confusing to load

Thanks for the idea and most of the work.

LGTM :+1:

samgermain:
> I hope i got them all - but this is what i meant.
> 
> `list-strategies` must also support this - otherwise it'll be highly confusing to load
> 
> Thanks for the idea and most of the work.
> 
> LGTM 👍

Thanks for finishing this up

PR #6561, opened 2022-03-20
Unified taker and maker's stake amount calculation logic

adrianceding:

## Summary

Unified taker and maker's stake amount calculation logic

Solve the issue: #___

## Quick changelog

- Unified taker and maker's stake amount calculation logic

## What's new?

Unify the calculation logic of the stake amount of taker and maker in the contract

### DIFF
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -847,7 +847,10 @@ class LocalTrade():
 
     def recalc_trade_from_orders(self):
         # We need at least 2 entry orders for averaging amounts and rates.
+        # TODO: this condition could probably be removed
         if len(self.select_filled_orders(self.enter_side)) < 2:
+            self.stake_amount = self.amount * self.open_rate / self.leverage
+
             # Just in case, still recalc open trade value
             self.recalc_open_trade_value()
             return
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -661,18 +661,15 @@ class FreqtradeBot(LoggingMixin):
                                self.exchange.name, order['filled'], order['amount'],
                                order['remaining']
                                )
-                stake_amount = order['cost']
                 amount = safe_value_fallback(order, 'filled', 'amount')
                 enter_limit_filled_price = safe_value_fallback(order, 'average', 'price')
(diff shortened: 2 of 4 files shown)


xmatthias (commented):
I think this will fix only 2 out of 3 cases.

* [x] market orders (`order_status == 'closed'`)
* [x] limit orders that filled immediately (`order_status == 'closed'`)
* [ ] limit orders which don't fill immediately

I think the correct place for this calculation might be `trade.update_trade()` - that would cover all 3 cases with one line of code - as all orders will have to go through there.

Interestingly, this is handled for multi-buys implicitly (look at `recalc_trade_from_orders()`) - so AFTER the 2nd buy stake_amount will be correct again - it's however wrong for cases with only one buy.

adrianceding:
Can we simply remove
`if len(self.select_filled_orders(self.enter_side)) < 2:`
 to fix this

xmatthias:
> Can we simply remove `if len(self.select_filled_orders(self.enter_side)) < 2:` to fix this

in theory - in theory the logic SHOULD also work for one order - but we'll have to think it through ... ;) 
Iirc, this was added to keep "regular" behaviour unchanged - even though it shouldn't change for 1 buy order either.
That said - it's possible that this can break 100ds of tests due to unaligned orders we use to fill trades
 ... so best give it a try?

**edit** the proposed change will break ~42 tests.
I'll have a look on how to approach this best.

freqtrade/persistence/models.py
```diff
@@ -847,7 +847,10 @@ def calc_profit_ratio(self, rate: Optional[float] = None,
 
     def recalc_trade_from_orders(self):
         # We need at least 2 entry orders for averaging amounts and rates.
+        # TODO: this condition could probably be removed
         if len(self.select_filled_orders(self.enter_side)) < 2:
+            self.stake_amount = self.amount * self.open_rate / self.leverage
```
xmatthias, line 852: I've moved this here for the moment.

While the "proper" approach would be to remove this condition and run the same code throughout - it's breaking too many tests at the moment due to "mock missmatch" (we mock a order that's slightly different to what we'd get here).

I have another idea on how to handle this - but i'll not add that to feat/short anymore, therefore we'll take this shortcut for the moment.

PR #6563, opened 2022-03-20
Optimize only new points

italodamato:
## Summary

Enforce points returned from `self.opt.ask` have not been already evaluated

## Quick changelog

- build `asked` iteratively in `hyperopt.py` to ensure the returned points are new

## What's new?

Steps:

    1. Try to get points using `self.opt.ask` first
    2. Discard the points that have already been evaluated
    3. Retry using `self.opt.ask` up to 3 times
    4. If still some points are missing in respect to `n_points`, random sample some points
    5. Repeat until at least `n_points` points in the `asked_non_tried` list
    6. Return a list with length truncated at `n_points`


freqtrade/optimize/hyperopt.py
```diff
@@ -410,6 +410,34 @@ def prepare_hyperopt_data(self) -> None:
...
+            if i < 3:
+                asked = self.opt.ask(n_points=n_points)
+            else:
+                asked = self.opt.space.rvs(n_samples=n_points * 5)
+            asked_non_tried += [x for x in asked
+                                if x not in self.opt.Xi
+                                and x not in asked_non_tried]
+            i += 1
+        if asked_non_tried:
+            return asked_non_tried[:n_points]
```
xmatthias: Wouldn't you have to check the length here, too? otherwise we risk getting fewer points than requested in some edge-case scenarios (where the space is almost exhausted).

with n_points=10 - i could in theory request 100 random points - but never get > n_points - so `asked_non_tried` would still be only 5 or so.

It's an edge-case for sure (probably one you never hit) - but i'm not entirely sure if it's handled correctly downstream, as until now, we assumed that `.ask()` get's me the amount of points i asked for.

freqtrade/optimize/hyperopt.py
```diff
@@ -410,6 +410,34 @@ def prepare_hyperopt_data(self) -> None:
...
+        i = 0
+        asked_non_tried: List[List[Any]] = []
+        while i < 100 and len(asked_non_tried) < n_points:
+            if i < 3:
+                asked = self.opt.ask(n_points=n_points)
+            else:
+                asked = self.opt.space.rvs(n_samples=n_points * 5)
+            asked_non_tried += [x for x in asked
+                                if x not in self.opt.Xi
+                                and x not in asked_non_tried]
```
xmatthias: i think this is potentially problematic - and may throw away too many points - resulting in too many "random picks" - or too many points not tried

---

For simplicity's sake let's assume i have 10 cores (so `n_points=10`).
I ran this optimization for 100 epochs already (so i do have "some" values already tried).

Now if `.opt.ask()` returns 5 point that were already tried - then ask for new points.
we'd however ask again for 10 points - and then (below) remove the remaining points (assuming ask_not_tried has now 15 elements).

As a result - we'd get 10 points - but 5 additional points were retrieved - but never calculated.

I'm not sure how skopt handles this - will it "know" that we never "told" the result to these 5 points - or will it assume they were tried and bad - and therefore never try these again (which would be problematic).

---

on top of that (and that's potentially more worrying) - looking at the code for "Optimizer" - it appears to implement a cache (`self.cache_`) - which would return the same points again for the same inputs, UNLESS `.tell()` is used - which will evict the cache.

italodamato: >As a result - we'd get 10 points - but 5 additional points were retrieved - but never calculated.

I'm not sure what you mean by this. The `ask` method is just a way to get the points that the Optimizer "thinks" will minimize the obj function. But you can also input(ie `tell`) the numbers manually without using the `ask` method.

> it appears to implement a cache (`self.cache_`)

so this means we can call `ask` just once instead of 3 times


In my tests, I'm getting awesome results instead of getting stuck in a local minimum like before, this helps increase the 'exploration' side of things.

freqtrade/optimize/hyperopt.py
italodamato: yes makes sense

freqtrade/optimize/hyperopt.py
xmatthias: > I'm not sure what you mean by this. The `ask` method is just a way to get the points that the Optimizer "thinks" will minimize the obj function. But you can also input(ie `tell`) the numbers manually without using the `ask` method.

That's clear to me ... but i looked "into" the ask function (not looking for this though - but it was a random observation).

https://github.com/scikit-optimize/scikit-optimize/blob/a2369ddbc332d16d8ff173b12404b03fea472492/skopt/optimizer/optimizer.py#L384-L385
https://github.com/scikit-optimize/scikit-optimize/blob/a2369ddbc332d16d8ff173b12404b03fea472492/skopt/optimizer/optimizer.py#L419

As far as i could tell, the cache is only reset when you use `.tell()` in between 2 calls to `.ask()`.

What will have happened in your tests is therefore most likely that "real" random points were selected (the else branch a few lines below this comment-line). That's not per se wrong - but makes calling `.ask()` multiple times obsolete.

italodamato: An alternative is to set `self.opt.cache_ = {}` manually after each `ask`
https://github.com/scikit-optimize/scikit-optimize/blob/a2369ddbc332d16d8ff173b12404b03fea472492/skopt/optimizer/optimizer.py#L525

xmatthias: > An alternative is to set `self.opt.cache_ = {}` manually after each `ask` https://github.com/scikit-optimize/scikit-optimize/blob/a2369ddbc332d16d8ff173b12404b03fea472492/skopt/optimizer/optimizer.py#L525

Yes that's correct - i'm just pointing out flaws in the current logic.- HOW that should be fixed is a different topic.

Potentially, we could assume that the local minima is exhausted - so we'll need fresh random points.
This would mean that with 5 valid points, and 2 invalid points - we would get 2 new random points (simply adding what's missing) - and combine these 2 new random points with the 5 points we got from .ask (now it's an either/or situation).

freqtrade/optimize/hyperopt.py
```diff
@@ -410,6 +410,35 @@ def prepare_hyperopt_data(self) -> None:
...
+        3. Retry using `self.opt.ask` up to 3 times
+        4. If still some points are missing in respect to `n_points`, random sample some points
+        5. Repeat until at least `n_points` points in the `asked_non_tried` list
+        6. Return a list with length truncated at `n_points`
+        '''
+        i = 0
+        asked_non_tried: List[List[Any]] = []
+        while i < 100 and len(asked_non_tried) < n_points:
+            if i < 3:
+                self.opt.cache_ = {}
```
xmatthias, line 435: Did you also observe if this also improved results - or if they got worse again? 

my suspicion here is that it may be beneficial to use completely random points once you hit this ... which is what happened before this line was added - where you described that results were better ...

italodamato: I'm not seeing a big difference, to be honest. My main issue was that, after a certain point, the ask method keeps returning the same point (or few points) over and over again because it doesn't have any interest in exploring other strategies to get to the minimum. I found it to be an issue, especially when using just a few parameters.

The downside of this approach is that right now it's not clear when the optimizer is running random points and when it's running 'optimal' points. It'd be great to store this information and display it in the progress table somehow.

xmatthias: is there a benefit to knowing this? (it's something you can't influence anyway) ... 

i mean - if my results are "best" based on random points or "real optimization" ... do i really care?
in theory the very first point could be the best possible point within this parameter space ...

freqtrade/optimize/hyperopt.py
```diff
@@ -410,6 +410,35 @@ def prepare_hyperopt_data(self) -> None:
...
+        1. Try to get points using `self.opt.ask` first
+        2. Discard the points that have already been evaluated
+        3. Retry using `self.opt.ask` up to 3 times
+        4. If still some points are missing in respect to `n_points`, random sample some points
+        5. Repeat until at least `n_points` points in the `asked_non_tried` list
+        6. Return a list with length truncated at `n_points`
+        '''
+        i = 0
+        asked_non_tried: List[List[Any]] = []
+        while i < 100 and len(asked_non_tried) < n_points:
```
xmatthias: Unfortunately this shows some very bad behavior once a space is "really" exhausted.

I've stress-tested this with 100 epochs and a space of 40 possible points (basically one int space). 
This means starting from epoch 40, it was impossible to get new points.

This led to endless loops of "new random points" (~100 per iteration) - which led to hyperopt taking quite a long time (a considerably longer than on develop) - which is the opposite from what we'd hope.

maybe it's just that 100 cycles/attempts are too much (considering that we'll be getting `3 x "real new points" + 97 * points * 5` - there MUST be some random points available a lot earlier.

freqtrade/optimize/hyperopt.py
italodamato: You might care because it might be a sign to stop earlier in case we just evaluate random points. The assumption is that random points should help the optimizer get interested in new areas of the search space. If the random points are not able to "dissuade" the optimizer from the current minimum, it might be a sign that we found actual absolute minimum.

freqtrade/optimize/hyperopt.py
italodamato: Thanks for checking this. I have reduced the number of loops. Are you able to test it with the current settings?

xmatthias: i'll need to do more tests in different scenarios - but as you'll know yourself - testing things in hyperopt is quite timeconsuming.

the above example (way too small a space of 40 points) shows as follows:

* ~1m for 100 epochs on current develop
* ~5m for 100 epochs on this branch

Now it's obviously a edge-case - one which actually should implement early stopping due to space exhaustion (which is something that scikit-optimize doesn't support afaik).

We might be in a position to do so with this method - but i'm not sure on the logic for that. After how many consecutive _random duplicate_ points would we start to consider a space 100% exhausted?


a more interesting test will be to test it with a bigger space - which i hope to find time for over the weekend.

italodamato: I've created a quick script to check exactly that: 

```python
import random
import pandas as pd
import numpy as np

def avg_cycles(tot_combinations, samples_per_cycle, n_iterations=20):
    stopped_at_i = []
    for _ in range(n_iterations):
        already_seen = []
        i=0
        while len(already_seen)<tot_combinations:
            samples = random.sample(range(tot_combinations), samples_per_cycle)
            already_seen += [s for s in samples if s not in already_seen]
            i+=1
        stopped_at_i.append(i)

    return np.mean(stopped_at_i)

df1 = pd.DataFrame(range(0, 2000, 100), columns=['n'])

for n in [2, 4, 6, 8, 10]:
    df1[f'samples_per_cycle_1/{n}'] = df1.apply(lambda x: avg_cycles(int(x['n']), int(x['n']//n)), axis=1)
    
ax = df1.set_index('n').plot(figsize=(10, 5), title='Average number of cycles to sample all values, depending on sample size as a fraction of tot values')
ax.set_xlabel('Number of total values')
ax.set_ylabel('Average number of cycles before sampling all values')

```
Which produces this chart 
image

I'd say that the number of cycles doesn't depend on the space size as long as we sample the number of points as a percentage of the total space size. Since we can compute the space size, we can just set the random samples to a fraction of the total space.

italodamato:
I'm not sure why it failed

italodamato:
I'm not sure why it failed

freqtrade/optimize/hyperopt.py
```diff
@@ -409,6 +409,52 @@ def prepare_hyperopt_data(self) -> None:
...
+        5. Repeat until at least `n_points` points in the `asked_non_tried` list
+        6. Return a list with length truncated at `n_points`
+        '''
+        def unique_list(a_list):
+            seen = []
+            for x in a_list:
+                key = repr(x)
+                if key not in seen:
+                    seen.append(key)
+            return [eval(x) for x in seen]
```
xmatthias: I think we should try to avoid using `eval` if possible (and i think it's possible above, and don't understand why it's made like this).
it's a quite dangerous method (it executes whatever you pass in).

in my understanding of the code - we show the representation (based on the assumption that it will come out as `"float(5)"` or something in a similar format).

We'll then append it to the seen list - and before returning, eval it (basically executing the above string - so `float(5)` turns into `5.0`).

This is unfortunately bound to fail / raise errors since
* you can't rely on all types to implement `__repr__` as `Type(value)`
* even if they implement it this way - nothing guarantees that this will give you the correct value (maybe the correct initialization would be `Type(value, somethingelse)`

I'm not sure why you convert it to string (and back) - it'll create the same object twice (performance problem) - and potentially result in value changes (if the `__repr__` limits precision shown, for example).

italodamato: Makes sense. I used that because the list is a list of lists that is unhashable if you try to transform it into a set. I've found another way to achieve the same with a simple for loop (just pushed).

italodamato: Makes sense. I used that because the list is a list of lists that is unhashable if you try to transform it into a set. I've found another way to achieve the same with a simple for loop (just pushed).

italodamato:
I don't think the failure is related to the change

xmatthias:
> I don't think the failure is related to the change

definitely not - the failure came from gateio responding with "Internal server errors" ...

xmatthias (approved):
Looks like a good improvement. 
We can look at "early stopping" if the space is exhausted separately - but i think this might enable us to do something in this regard
 
LGTM :+1:

PR #6575, opened 2022-03-21
taker or maker passed correctly for entry and exit orders

samgermain:
Freqtrade statically passes `maker` to get fees for an order, this new method isn't perfect (limit orders can still be taker orders) but its better than what was there before. I couldn't figure out the places to update the fee if it's a stoploss order, all I could see was stoploss on exchange, which should only be for live anyway

I can't get the test to pass because Trade.open_fee returns the MagicMock instead of the actual return value

```
    def _calc_open_trade_value(self) -> float:
        """
        Calculate the open_rate including open_fee.
        :return: Price in of the open trade incl. Fees
        """
        buy_trade = Decimal(self.amount) * Decimal(self.open_rate)
>       fees = buy_trade * Decimal(self.fee_open)
E       TypeError: conversion from MagicMock to Decimal is not supported

freqtrade/persistence/models.py:568: TypeError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/sam/Documents/dev/freqtrade/freqtrade/freqtrade/persistence/models.py(568)_calc_open_trade_value()
-> fees = buy_trade * Decimal(self.fee_open)
(Pdb) self.fee_open
<MagicMock id='4929478464'>
(Pdb) self.fee_open.return_value
0.0025
```

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -570,32 +570,21 @@ class FreqtradeBot(LoggingMixin):
             logger.info(f"Bids to asks delta for {pair} does not satisfy condition.")
             return False
 
-    def get_fee(
+    def is_maker(
         self,
-        pair: str,
         side: str,
-        is_short: bool = False,
     ):
         '''
-        :param pair: Unified CCXT symbol
-        :param side: One of entry, exit or stoploss
-        :param is_short: True for short trades
+        :param side: One of entry, or exit # TODO: stoploss
         '''
         order_type = self.config['order_types'][side]
         if order_type == 'limit':
-            if (
+            return (
                 side == 'entry' and self.config['entry_pricing']['price_side'] == 'same' or
                 side == 'exit' and self.config['exit_pricing']['price_side'] == 'same'
-            ):
-                taker_or_maker = 'maker'
-            else:
-                taker_or_maker = 'taker'
+            )
         else:
-            taker_or_maker = 'taker'
-        return self.exchange.get_fee(
-            symbol=pair,
-            taker_or_maker=taker_or_maker
-        )
+            return True
 
     def execute_entry(
         self,
@@ -696,6 +685,8 @@ class FreqtradeBot(LoggingMixin):
         funding_fees = self.exchange.get_funding_fees(
             pair=pair, amount=amount, is_short=is_short, open_date=open_date)
         if trade is None:
+            open_maker = self.is_maker('entry')
+            close_maker = self.is_maker('exit')
             trade = Trade(
                 pair=pair,
                 base_currency=base_currency,
@@ -704,8 +695,10 @@ class FreqtradeBot(LoggingMixin):
                 amount=amount,
                 is_open=True,
                 amount_requested=amount_requested,
-                fee_open=self.get_fee(pair, 'entry'),
-                fee_close=self.get_fee(pair, 'exit'),
+                fee_open=self.exchange.get_fee(symbol=pair, taker_or_maker=('maker' if open_maker else 'taker')),
+                fee_close=self.exchange.get_fee(symbol=pair, taker_or_maker=('maker' if close_maker else 'taker')),
+                fee_open_is_maker=open_maker,
+                fee_close_is_maker=close_maker,
                 open_rate=enter_limit_filled_price,
                 open_rate_requested=enter_limit_requested,
                 open_date=open_date,
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -583,11 +583,9 @@ class FreqtradeBot(LoggingMixin):
         '''
         order_type = self.config['order_types'][side]
         if order_type == 'limit':
-            enter_side = 'ask' if is_short else 'bid'
-            exit_side = 'bid' if is_short else 'ask'
             if (
-                side == 'entry' and self.config['bid_strategy']['price_side'] == enter_side or
-                side == 'exit' and self.config['ask_strategy']['price_side'] == exit_side
+                side == 'entry' and self.config['entry_pricing']['price_side'] == 'same' or
+                side == 'exit' and self.config['exit_pricing']['price_side'] == 'same'
             ):
                 taker_or_maker = 'maker'
             else:
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -570,6 +570,35 @@ class FreqtradeBot(LoggingMixin):
             logger.info(f"Bids to asks delta for {pair} does not satisfy condition.")
             return False
 
+    def get_fee(
+        self,
+        pair: str,
+        side: str,
+        is_short: bool = False,
+    ):
+        '''
+        :param pair: Unified CCXT symbol
+        :param side: One of entry, exit or stoploss
+        :param is_short: True for short trades
+        '''
+        order_type = self.config['order_types'][side]
+        if order_type == 'limit':
+            enter_side = 'ask' if is_short else 'bid'
+            exit_side = 'bid' if is_short else 'ask'
+            if (
+                side == 'entry' and self.config['bid_strategy']['price_side'] == enter_side or
+                side == 'exit' and self.config['ask_strategy']['price_side'] == exit_side
+            ):
+                taker_or_maker = 'maker'
+            else:
+                taker_or_maker = 'taker'
+        else:
+            taker_or_maker = 'taker'
+        return self.exchange.get_fee(
+            symbol=pair,
+            taker_or_maker=taker_or_maker
+        )
+
     def execute_entry(
         self,
         pair: str,
@@ -664,12 +693,10 @@ class FreqtradeBot(LoggingMixin):
             enter_limit_filled_price = safe_value_fallback(order, 'average', 'price')
 
         # Fee is applied twice because we make a LIMIT_BUY and LIMIT_SELL
-        fee = self.exchange.get_fee(symbol=pair, taker_or_maker='maker')
         base_currency = self.exchange.get_pair_base_currency(pair)
         open_date = datetime.now(timezone.utc)
         funding_fees = self.exchange.get_funding_fees(
             pair=pair, amount=amount, is_short=is_short, open_date=open_date)
-        # This is a new trade
         if trade is None:
             trade = Trade(
                 pair=pair,
@@ -679,8 +706,8 @@ class FreqtradeBot(LoggingMixin):
                 amount=amount,
                 is_open=True,
                 amount_requested=amount_requested,
-                fee_open=fee,
-                fee_close=fee,
+                fee_open=self.get_fee(pair, 'entry'),
+                fee_close=self.get_fee(pair, 'exit'),
                 open_rate=enter_limit_filled_price,
                 open_rate_requested=enter_limit_requested,
                 open_date=open_date,
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -682,9 +682,10 @@ class LocalTrade():
             self.fee_open_cost = fee_cost
             self.fee_open_currency = fee_currency
             if fee_rate is not None:
+                current_rate_ratio = self.fee_close / self.fee_open
                 self.fee_open = fee_rate
-                # Assume close-fee will fall into the same fee category and take an educated guess
-                self.fee_close = fee_rate
+                # The new open/close fees will be as proportional as the old fees
+                self.fee_close = fee_rate * current_rate_ratio
         elif self.exit_side == side and self.fee_close_currency is None:
             self.fee_close_cost = fee_cost
             self.fee_close_currency = fee_currency
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -285,9 +285,11 @@ class LocalTrade():
     fee_open: float = 0.0
     fee_open_cost: Optional[float] = None
     fee_open_currency: str = ''
+    fee_open_is_maker: str = True
     fee_close: float = 0.0
     fee_close_cost: Optional[float] = None
     fee_close_currency: str = ''
+    fee_close_is_maker: str = True
     open_rate: float = 0.0
     open_rate_requested: Optional[float] = None
     # open_trade_value - calculated via _calc_open_trade_value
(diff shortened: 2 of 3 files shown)


freqtrade/freqtradebot.py
```diff
@@ -582,8 +593,6 @@ def execute_entry(self, pair: str, stake_amount: float, price: Optional[float] =
             amount = safe_value_fallback(order, 'filled', 'amount')
             enter_limit_filled_price = safe_value_fallback(order, 'average', 'price')
 
-        # Fee is applied twice because we make a LIMIT_BUY and LIMIT_SELL
-        fee = self.exchange.get_fee(symbol=pair, taker_or_maker='maker')
```
xmatthias, line 667: I think keeping this "inline" will make more sense - as it does not make sense to assign the close_fee a different value - as the value is in ~80% of the cases wrong.

CCXT is returning the exchange default value as far as i know (not the one tied to your particular account including rebates, VIP levels etc.).
Therefore, the moment you use your account "a bit" - your real fees will be different (e.g. Paying in BNB on binance gives you quite some rebates).

As such - there is logic in place to make the close-fee guess to be identical than the entry-fee (we know you paid 0.05% instead of the default 0.1% - so it's pretty safe to assume you'll get the same rebate on closing, too).
So assigning close_fee here will have only temporary impact - with even more questionable benefits (even if we know you're using market orders to exit - the fee will still be different).

samgermain: > as it does not make sense to assign the close_fee a different value

Why? What if it is a different value

> as the value is in ~80% of the cases wrong.

Why?

> Therefore, the moment you use your account "a bit" - your real fees will be different (e.g. Paying in BNB on binance gives you quite some rebates).

How is anything related to that changed? I don't think I changed anything related to custom fee rates in the config. If it's wrong now, it would have been wrong before too, unless something calls `Trade.update_fee` that I'm not seeing

xmatthias: > Why? What if it is a different value

It's overwritten the moment the order fills - so you write it here and it's valid potentially only for seconds.
Also the opening fee will be wrong in the same 80% of the cases if you trade live - simply because you use real money - and exchanges have different fees depending on the usage of the account - which ccxt will not return until you placed an order.

> Why?

I don't think it's wrong now - but it doesn't sense to assign different fees for open and close if different order types are used - if we know for a fact that it'll be overwritten the moment the order fills with the fee the order filled with.

As such - I'd have kept it as "get open fee - assume that same fee for closing" (which would align it to "after the order filled", too). 
It's as good a guess as any - and just as wrong as the new approach.

samgermain: In dry run it's calculating the wrong fees, that's how I noticed it, and that's why I'm changing it, strategies that use market orders don't have accurate dry runs

samgermain: Those are all market orders, so they're going to be maker, and should be 0.0004, if my open orders are limit though with exit orders being market (which is common with a philosophy of being slow to get in but quick getting out) then the fee rate for open orders would probably be 0.0002 but 0.0004 for exit orders, unless you're placing a limit order at a price above the ask price

samgermain: > and exchanges have different fees depending on the usage of the account 

CCXT actually has the fee tiers for a number of exchanges, it doesn't have them for binance spot, but I could fix that, it wouldn't matter unless the user gives their tier anyway because dry run doesn't access private methods

xmatthias: it wasn't clear that you're talking dry-run only with this - what I'm talking about above is live.

the updating of fees should not happen in dry-run - so there we'd be safe (my main thinking was about live though - where it'll be overwritten as fees are updated).

samgermain: > it wasn't clear that you're talking dry-run only with this - what I'm talking about above is live.
> 
> the updating of fees should not happen in dry-run - so there we'd be safe (my main thinking was about live though - where it'll be overwritten as fees are updated).

Does it make sense to you that limit orders would mostly be maker, and market orders would mostly be taker? 

Also, I added the order book side logic to the get_fee method

samgermain: even for live they shouldn't be the same, if ones a taker order and the other is maker they'll be different, we need to fetch both for live

xmatthias: My point is a different one - and obviously you didn't get that yet.

I'm not arguing that the fees are identical - what i'm saying is that whatever we assume a fee to be **before an order fills** is an estimation - which is most likely wrong (wrong in ~80% of the cases - at least on binance).

---

while the fees may be different - they will ALSO be different by user account.
you may not be paying the same fees on binance than i do, because my volume might be above yours, or because I'm holding more BNB - or because i have fewer referals or whatever else counts into this equation (it's different by exchange).

While you're talking about "tiered" fees - that's all nice and well - but we'll still not get the **correct fees** until an order is executed (fees estimated based on the user's account parameters). 
That's not a ccxt limitation, but an exchange API one - as I'm yet to see such an endpoint from the exchange.


---

So in a **mixed** case (limit for entry, market for exit) 
- we will take an assumption at entry (this is what you're messing with now) (say - `0.001 for entry`, and `0.002 for exit`).
- Once the order fills - we'll know the real rate the exchange applied (this may or may not be different to the value we get via `get_fee()` in reality, we did pay `0.0005` (half of the original fee) - and in BNB (this is not relevant, we convert it to a ratio).

As the order filled - we now know what the entry rate was - so `update_fee` will update the close-rate to `0.0005`.
This means, your prior estimation (`0.002` for exit) was pointless  as it is overwritten.

Now we have 2 possibilities at this point - we either assume that the initial assumption was correct (`0.002` for the exit) - so we remove the "set close fee" line.
or we assume that the adapted fee is more appropriate - so we leave it as is, and assume `0.0005` is our new exit fee.

The problem is - both are potentially wrong - as the ordertype is different, we might encounter `0.0015` as exit-fee (maybe rebate is lower for market orders).

This is why i think setting the close-fee differently is pointless (at least when looking at live) - as we simply do not know the real fee until an order is executed (and even then, it's only a guess for the next order, as circumstances could change and therefore different rates may apply for the exit).


> Does it make sense to you that limit orders would mostly be maker, and market orders would mostly be taker?

This will depend on the pricing side - limit orders can also pay taker fees if it becomes a "marketable limit order" - which basically means that a buy price above the spread was selected. Then we define the "worst" possible price the order will fill - but if the current best ask is below the price, then we'll get that price (but pay taker fees).
Market orders are simpler - there you know that you pay the taker fee.

freqtrade/freqtradebot.py
```diff
@@ -474,7 +474,7 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
 
         if stake_amount is not None and stake_amount > 0.0:
             # We should increase our position
-            self.execute_entry(trade.pair, stake_amount, trade=trade)
+            self.(trade.pair, stake_amount, trade=trade)
```
xmatthias: I assume that's not an intended change ... as i don't think that's valid python syntax ...

freqtrade/freqtradebot.py
samgermain: > This is why i think setting the close-fee differently is pointless (at least when looking at live) - as we simply do not know the real fee until an order is executed (and even then, it's only a guess for the next order, as circumstances could change and therefore different rates may apply for the exit).

Well a better solution would be to multiply the open fee by maker/taker if open is taker and close is maker, and taker/maker if open is maker and close is taker

> This will depend on the pricing side - limit orders can also pay taker fees if it becomes a "marketable limit order" - which basically means that a buy price above the spread was selected. Then we define the "worst" possible price the order will fill - but if the current best ask is below the price, then we'll get that price (but pay taker fees).

Yeah I know, but the best we can do in dry_run is assume maker or taker based on the config settings in ["bid_strategy"]["price_side"] and ["ask_strategy"]["price_side"]

My fixes in this PR are for dry run though

samgermain: > As the order filled - we now know what the entry rate was - so update_fee will update the close-rate to 0.0005.
This means, your prior estimation (0.002 for exit) was pointless as it is overwritten.

I'm not sure if I know exactly what code your talking about, but in Trade.update_fee, I made it so that the exit fee gets set to the open fee, but the open and exit fee are proportial as there were before, so setting the close_fee with the `0.0002` value because it helps to determine what the size of the close fee is compared to the open fee

samgermain:
i'm not really sure what to do about stoploss, I was thinking about adding another fee_stoploss to Trade but thought this might be confusing in the database, and then checking what the sell reason is and assigning either close_fee or stoploss_fee

freqtrade/persistence/models.py
```diff
@@ -531,9 +531,10 @@ def update_fee(self, fee_cost: float, fee_currency: Optional[str], fee_rate: Opt
             self.fee_open_cost = fee_cost
             self.fee_open_currency = fee_currency
             if fee_rate is not None:
+                current_rate_ratio = self.fee_close / self.fee_open
```
xmatthias: you cannot safely rely on this method being called only once (There are several scenarios where it'll be called multiple times, and also reach this line) per trade.

If this is called 2 or 3 times - then you'll encounter "disappearing" fees - as the close-fee will each time be updated lower - as the gap widens.

xmatthias (changes requested):
Unfortunately, this approach won't work, and is therefore mostly pointless.

I've had to check this first - but your changes will not work (also/especially not on dry-run) - as you'll run into the same problem than what i described for live above - which is that the fee will be overwritten once the order fills.

You should be able to inject a different logic into `exchange.add_dry_order_fee()` - where you'll also have the proper order-logic available (using `/forcebuy` via API can use different orders!).

In it's current state however, the impact you're generating will last seconds on average (until the order fills) only to be then overwritten by the "default" that comes from `get_fee()`.

The probably simplest mode will be to add the appropriate parameters to the call to get_fee there - as such, i think to archive the goal of this PR - as it'll all be handled by the exchange class.

BKC42042 (approved):

freqtrade/freqtradebot.py
```diff
@@ -679,8 +695,10 @@ def execute_entry(
                 amount=amount,
                 is_open=True,
                 amount_requested=amount_requested,
-                fee_open=fee,
-                fee_close=fee,
+                fee_open=self.exchange.get_fee(symbol=pair, taker_or_maker=('maker' if open_maker else 'taker')),
```
xmatthias, line 698: Please take a look at the exchange's `add_dry_order_fee()` method.

As pointed out before - doing this here is pointless, and the fee  assigned here will only have an effect _until the order fills_ - after which it's overwritten with the value we assign via "dry-run order" handling (see `exchange.add_dry_order_fee()`).


I'm not against this change in general (so we use the proper fee type) - but the fee assigned in this place is wrong.

freqtrade/persistence/models.py
```diff
@@ -1079,9 +1087,11 @@ class Trade(_DECL_BASE, LocalTrade):
     fee_open = Column(Float, nullable=False, default=0.0)
     fee_open_cost = Column(Float, nullable=True)
     fee_open_currency = Column(String(25), nullable=True)
+    fee_open_is_maker = Column(Boolean, default=True)
     fee_close = Column(Float, nullable=False, default=0.0)
     fee_close_cost = Column(Float, nullable=True)
     fee_close_currency = Column(String(25), nullable=True)
+    fee_close_is_maker = Column(Boolean, default=True)
```
xmatthias, line 1094: I'm not sure about the usefulness of these columns.

If you consider realistic order handling (the fee comes from the **filled order** - both in live and dry-run!) - then you'll not be able to reliably fill these columns (so the content may be wrong - and as such, the usage in the calculation may be wrong.

freqtrade/persistence/models.py
```diff
@@ -682,9 +684,15 @@ def update_fee(self, fee_cost: float, fee_currency: Optional[str], fee_rate: Opt
             self.fee_open_cost = fee_cost
             self.fee_open_currency = fee_currency
             if fee_rate is not None:
+                if self.fee_open_is_maker == self.fee_close_is_maker:
+                    current_rate_ratio = 1
+                elif self.fee_open_is_maker and not self.fee_close_is_maker:
+                    current_rate_ratio = 2
+                elif not self.fee_open_is_maker and self.fee_close_is_maker:
+                    current_rate_ratio = 0.5
```
xmatthias, line 692: i don't think you can generalize the ratio in this way across exchanges - which makes this calculation wrong now.

It may be true for the exchange you're using (it probably isn't - if we look at binance at this moment - then fees are identical for maker/taker in the "lowest level" tier).

image

The calculation takes the assumption that taker/maker difference is 50% - but that's only true for VIP9 users on binance (volume over 5 Billion BUSD!)...

xmatthias (changes requested):
Unfortunately, this is still wrong - and the improvements you think apply to dry-run don't really apply in a real dry-run scenario - as the value is overwritten by the filled order fees (where the difference is not applied).

Most of this should therefore be moved to the exchange class - and used by dry-run filling (as well).

I'm also strongly in doubt about the "half maker fee" approach - as it seems misalligned to real market fees (at least looking at binance). 
It may be correct for some exchanges, but not for others - as such it'll decrease accuracy (binance has the biggest FT userbase afaik), instead of increasing it.

freqtrade/persistence/models.py
samgermain: Right, we should pass in the taker and maker ratio instead.

It's true for binance futures, but yeah I was generalizing too much

freqtrade/freqtradebot.py
samgermain: I don't know how to do it correctly, but when I look at my dry run results, they all have the maker fee when using a market order strategy, when they shouldn't

xmatthias: well that'll not change with this PR :laughing: at least not in it's current state.

As said - apply THIS change (or something similar) to the method pointed out above - and it should work.

samgermain:
Reopened on https://github.com/freqtrade/freqtrade/pull/6780

Although that PR doesn't cover everything that this PR does. I think that there would still be a benefit to estimating the close fee as being equal to open fee only if the side of entry and exit orders are the same and otherwise take the ratio of 

- taker/maker if exit=='taker' and entry='maker'
- maker/taker if exit=='maker' and entry='taker'

but the way I've written it in this PR, it looks like it always gets overwritten

xmatthias:
> but the way I've written it in this PR, it looks like it always gets overwritten

yes - that's my point all along 😆 
will look at the other PR shortly - but a first glance looks promising.

samgermain:
> > but the way I've written it in this PR, it looks like it always gets overwritten
> 
> yes - that's my point all along 😆 will look at the other PR shortly - but a first glance looks promising.

If it can be not overwritten though, I think the logic in this PR would be best, except if maker is 0%. If entry is `maker` and exit is `taker` that can mean an entry price of 0 on gateio, and exit price of 0.05%

PR #6576, opened 2022-03-21
funding_fee tests

samgermain:
Funding fee tests for cases where there's 0 funding fees

Right now this PR should fail because of the [reasons mentioned here](https://github.com/freqtrade/freqtrade/commit/ddce28c12db4a6b8ca0b50c564f3578feddcc074#r69159950). I'm not sure why we round that date down


freqtrade/exchange/exchange.py
```diff
@@ -2302,11 +2302,10 @@ def _fetch_and_calculate_funding_fees(
         timeframe = self._ft_has['mark_ohlcv_timeframe']
         timeframe_ff = self._ft_has.get('funding_fee_timeframe',
                                         self._ft_has['mark_ohlcv_timeframe'])
-        open_date = timeframe_to_prev_date(timeframe, open_date)
 
         if not close_date:
             close_date = datetime.now(timezone.utc)
-        open_timestamp = int(open_date.timestamp()) * 1000
+        open_timestamp = int(timeframe_to_prev_date(timeframe, open_date).timestamp()) * 1000
```
xmatthias, line 2308: @samgermain Basically like this.

Obviously this required also some changes to other test-cases - and i removed excessive decimals (which makes the diff in the test-file look worse than it is).

Please have a look at the changes i made - and if you're happy with it / agree with the change, simply merge it.

samgermain: I'm fine with it, but it looks like the new tests I added with 0 funding fees are failing

xmatthias: no - the failing tests are in a different test file (one i didn't run locally, apparently...) 😆

xmatthias: How bout now? :laughing:

Issue #6583, opened 2022-03-23
Could Kucoin 42900 be applied to `retrier`? 


SmartManoj:
## Describe your environment

  * Operating system: ____
  * Python Version: _____ (`python -V`)
  * CCXT version: _____ (`pip freeze | grep ccxt`)
  * Freqtrade Version: e7418cdcdb9cf7916ff0a5464113c20817fe7486 (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
  
## Your question

#5847 is only applied for `retrier_async`.

xmatthias:
Please get used to fill out the issue template. Repeat offenders to this will have their issues closed without further comment.
We're asking some very basic questions - by deleting the issue template, you're explicitly choosing not to answer our questions - which makes me doubt why you think we should chose to answer your questions.

---

we could - but as mentioned in https://github.com/freqtrade/freqtrade/pull/6584#pullrequestreview-919207451 - i don't see why - as this problem is only affecting ohlcv endpoints as far as i know.
Blindly applying it to all endpoints can lead to permanent exchange bans (we're disrespecting their rate-limits), so if any, we'll have to be selective of the endpoints we KNOW it affects (there is none other than ohlcv i'm aware of - so for the moment, i see this as "won't fix".

PR #6584, opened 2022-03-23
kucoin temp fix for retrier

SmartManoj:

## Summary

Solves #6583

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

### DIFF
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -121,19 +121,30 @@ def retrier(_func=None, retries=API_RETRY_COUNT):
         @wraps(f)
         def wrapper(*args, **kwargs):
             count = kwargs.pop('count', retries)
+            kucoin = args[0].name == "Kucoin"  # Check if the exchange is KuCoin.
             try:
                 return f(*args, **kwargs)
             except (TemporaryError, RetryableOrderError) as ex:
                 msg = f'{f.__name__}() returned exception: "{ex}". '
                 if count > 0:
-                    logger.warning(msg + f'Retrying still for {count} times.')
+                    msg += f'Retrying still for {count} times.'
                     count -= 1
                     kwargs.update({'count': count})
                     if isinstance(ex, (DDosProtection, RetryableOrderError)):
-                        # increasing backoff
-                        backoff_delay = calculate_backoff(count + 1, retries)
-                        logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
-                        time.sleep(backoff_delay)
+                        if kucoin and "429000" in str(ex):
+                            # Temporary fix for 429000 error on kucoin
+                            # see https://github.com/freqtrade/freqtrade/issues/5700 for details.
+                            _get_logging_mixin().log_once(
+                                f"Kucoin 429 error, avoid triggering DDosProtection backoff delay. "
+                                f"{count} tries left before giving up", logmethod=logger.warning)
+                            # Reset msg to avoid logging too many times.
+                            msg = ''
+                        else:
+                            backoff_delay = calculate_backoff(count + 1, retries)
+                            logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
+                            time.sleep(backoff_delay)
+                    if msg:
+                        logger.warning(msg)
                     return wrapper(*args, **kwargs)
                 else:
                     logger.warning(msg + 'Giving up.')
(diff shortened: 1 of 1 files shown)


xmatthias (commented):
Neither this nor the linked issue describe where you're actually seeing this (which endpoint).

based on my information, 42900 is ONLY happening on the ohlcv endpoint, and means "you can retry immediately".

If it happens on other endpoints, please provide logs of this.
this is a temporary workaround, specifically made to handle a known and well defined problem (on the ohlcv endpoint).
We've had confirmation from the kucoin support for the ohlcv endpoint specifically. The way it's applied currently (to all endpoints) is only there because we use the `async_retrier` only where the error actually happens.

Applying it blindly to other endpoints can result in a permanent ban by the exchange, so unless you can back up this PR with some logs - i'll consider this as introducing a bug / potentially problematic behaviour.

SmartManoj:
Happens in `fetch_l2_order_book`
Log: https://github.com/mikekonan/exchange-proxy/issues/40#issue-1176625610



> Applying it blindly to other endpoints can result in a permanent ban by the exchange

https://github.com/freqtrade/freqtrade/issues/5700#issuecomment-955632324

Here they told generally and they got reports for ohlcv endpoint only.

freqtrade/exchange/common.py
```diff
@@ -121,19 +121,30 @@ def decorator(f):
...
-                    logger.warning(msg + f'Retrying still for {count} times.')
+                    msg += f'Retrying still for {count} times.'
                     count -= 1
                     kwargs.update({'count': count})
                     if isinstance(ex, (DDosProtection, RetryableOrderError)):
-                        # increasing backoff
-                        backoff_delay = calculate_backoff(count + 1, retries)
-                        logger.info(f"Applying DDosProtection backoff delay: {backoff_delay}")
-                        time.sleep(backoff_delay)
+                        if kucoin and "429000" in str(ex):
```
xmatthias, line 134: You may want to add a test for this (similar to `test__async_kucoin_get_candle_history` - but for a sync endpoint).

While the code does work now - that can't be guaranteed for the future without test - and especially in this area of the bot, it's critical to be able to rely on the code to work correctly.

xmatthias:
I've done some local tests, and the error is not happening (not even when bombarding the exchange with requests) - the error then is a different one - not 429000. 
Now that's clearly not deterministic (maybe they have stronger servers in my region, and worse in other regions, or this week is performing better).

> Log: [mikekonan/exchange-proxy#40 (comment)](https://github.com/mikekonan/exchange-proxy/issues/40#issue-1176625610)

The log clearly shows that it's through the proxy - not for direct calls, as such it could just as well be some feature/bug of the proxy you're using (most likely not respecting the api limits).

For now - i don't see any sign that this is actually happening in a **supported configuration** (proxied setups are not), so don't expect this to be merged unless you're able to provide some logs where this is happening in a supported setup.
There's also been no reports in any issue or on discord about this so far - all reports were for the ohlcv endpoint only.

This could just as well be caused by the proxy not respecting the exchange's rate-limits.
in a proxy setup, calls come from multiple clients. 
If you run multiple bots (clients) on the same server, you may run into exchange rate limit problems.

If you run through the proxy, it's the proxy's responsibility to take care of this. If it's not - then this may lead to this error.

SmartManoj:
https://github.com/mikekonan/exchange-proxy/issues/40#issuecomment-1075120830
The orderbook call is not proxied and only one bot is running.

SmartManoj:
> Log: [mikekonan/exchange-proxy#40 (comment)](https://github.com/mikekonan/exchange-proxy/issues/40#issue-1176625610)
> 
> >

Is the error comes for sync or async version?

freqtrade/exchange/common.py
```diff
@@ -85,7 +85,7 @@ def calculate_backoff(retrycount, max_retries):
 def retrier_async(f):
     async def wrapper(*args, **kwargs):
         count = kwargs.pop('count', API_RETRY_COUNT)
-        kucoin = args[0].name == "Kucoin"  # Check if the exchange is KuCoin.
+        kucoin = args[0].name.lower() == "kucoin"  # Check if the exchange is KuCoin.
```
xmatthias, line 88: Please align it to how it's done in the async retrier now (using proper casing).

Also note that the test is still outstanding as requested below.

xmatthias:
As I've never seen this reported anywhere in plain freqtrade setups (no proxy), this will only clutter the freqtrade codebase with useless, probably unmaintained (and currently untested) code.

As such, this is not an improvement / fix - but a liability for this project.

on top of the above, you don't seem inclined on finishing this PR (feedback ignored for several months - while popping up other PR's so you're clearly still active) - i assume this is no longer relevant and will close this PR.

PR #6585, opened 2022-03-23
Removed old datetime keys and added timestamp

SmartManoj:

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?


tests/conftest.py
```diff
@@ -1140,7 +1139,7 @@ def limit_buy_order_canceled_empty(request):
             'id': '1234512345',
             'clientOrderId': None,
             'timestamp': arrow.utcnow().shift(minutes=-601).int_timestamp,
-            'datetime': arrow.utcnow().shift(minutes=-601).isoformat(),
+            'timestamp': arrow.utcnow().shift(minutes=-601).int_timestamp,
```
xmatthias: This (and a few occurances below as well) are clearly incorrect.

The ccxt response has both `timestamp` as well as `datetime` (formatted in iso format). 
Your change here proposes to have the `timestamp` key twice - which does not make sense.

xmatthias (changes requested):
It makes sense to have timestamp added where it's currently missing (the first few lines).

Removing datetime / replacing it by having timestamp twice is clearly wrong, as both fields are in the ccxt orders response - and simplify reading and debugging the test as a "readable" datetime is available next to the timestamp.

tests/conftest.py
SmartManoj: updated

tests/conftest.py
```diff
@@ -1394,7 +1399,7 @@ def tickers():
         'BLK/BTC': {
             'symbol': 'BLK/BTC',
             'timestamp': 1522014806072,
-            'datetime': '2018-03-25T21:53:26.720Z',
+            'datetime': '2018-03-25T21:53:26.072Z',
```
xmatthias: that's for sure a typo in the small numbers.

bigger problem is that 2 tests are now failing.

SmartManoj: updated

xmatthias (approved):
LGTM :+1:

tests/conftest.py
SmartManoj: > that's for sure a typo in the small numbers.

Isn't 072 is the right value? Previously value was the typo.

xmatthias: yeah you're right - if you instead of "updated" write "but this is the correct value now" i'd have had a closer look (with such "chore" pr's that's sometimes problematic).

fixed in 60f2a12bd9104fb7233fafda6052102265f89013

PR #6588, opened 2022-03-25
Add amount_to_contracts and order_contracts_to_amount to exchange.stoploss

samgermain:
I believe we only need to add these methods when we're actually interacting with the exchange, we convert the amount to the number of contracts, but after making the order, we immediately convert the contracts back to the amount, so the number only needs to change for that brief period that we're actually making the order. It's not needed within `create_dry_run_order` then because we're not actually sending an order to the exchange, there's no need to convert it, and then convert it back immediately, because that step where we send the amount that the exchange reads isn't there

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -402,7 +402,7 @@ class Exchange:
         return trades
 
     def _order_contracts_to_amount(self, order: Dict) -> Dict:
-        if 'symbol' in order:
+        if 'symbol' in order and order['symbol'] is not None:
             contract_size = self._get_contract_size(order['symbol'])
             if contract_size != 1:
                 for prop in ['amount', 'cost', 'filled', 'remaining']:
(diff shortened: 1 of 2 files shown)


xmatthias:
Seems like the gateio response is not as sophisticated as we'd like.

It's missing basically every info - so "converting back" is failing at the moment as symbol is None (the order is however correctly placed).

```
{'id': '15814774', 'clientOrderId': None, 'timestamp': None, 'datetime': None, 'lastTradeTimestamp': None, 'status': None, 'symbol': None, 'type': None, 'timeInForce': None, 'postOnly': False, 'side': None, 'price': None, 'stopPrice': None, 'average': None, 'amount': None, 'cost': None, 'filled': None, 'remaining': None, 'fee': None, 'fees': [], 'trades': [], 'info': {'id': '15814774'}}
```

Don't worry for now though - i'll fix this via "live debugging" ...

samgermain:
> Seems like the gateio response is not as sophisticated as we'd like.
> 
> It's missing basically every info - so "converting back" is failing at the moment as symbol is None (the order is however correctly placed).
> 
> ```
> {'id': '15814774', 'clientOrderId': None, 'timestamp': None, 'datetime': None, 'lastTradeTimestamp': None, 'status': None, 'symbol': None, 'type': None, 'timeInForce': None, 'postOnly': False, 'side': None, 'price': None, 'stopPrice': None, 'average': None, 'amount': None, 'cost': None, 'filled': None, 'remaining': None, 'fee': None, 'fees': [], 'trades': [], 'info': {'id': '15814774'}}
> ```
> 
> Don't worry for now though - i'll fix this via "live debugging" ...

Right, when you create an order on gateio they only return the order id in the response

xmatthias (approved):
Works nicely :+1:

PR #6590, opened 2022-03-26
Fix gateio fee handling for futures

xmatthias:
## Summary
Futures orders do incurr a fee when they are executed.
This is however not part of the api response in some cases (it appears for users without a ref id).

Therefore updating of fees is failing.

## Quick changelog

- Download trading_fees (private endpoint - per user)
- use trading_fees to inject this to the gateio order

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -1300,6 +1300,27 @@ class Exchange:
             raise OperationalException(e) from e
 
     @retrier
+    def fetch_trading_fees(self) -> Dict[str, Any]:
+        """
+        Fetch user account trading fees
+        Can be cached, should not update often.
+        """
+        if (self._config['dry_run'] or self.trading_mode != TradingMode.FUTURES
+                or not self.exchange_has('fetchTradingFees')):
+            return {}
+        try:
+            trading_fees: Dict[str, Any] = self._api.fetch_trading_fees()
+            self._log_exchange_response('fetch_trading_fees', trading_fees)
+            return trading_fees
+        except ccxt.DDoSProtection as e:
+            raise DDosProtection(e) from e
+        except (ccxt.NetworkError, ccxt.ExchangeError) as e:
+            raise TemporaryError(
+                f'Could not fetch trading fees due to {e.__class__.__name__}. Message: {e}') from e
+        except ccxt.BaseError as e:
+            raise OperationalException(e) from e
+
+    @retrier
     def fetch_bids_asks(self, symbols: List[str] = None, cached: bool = False) -> Dict:
         """
         :param cached: Allow cached result
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -93,6 +93,7 @@ class Exchange:
         self._api: ccxt.Exchange = None
         self._api_async: ccxt_async.Exchange = None
         self._markets: Dict = {}
+        self._trading_fees: Dict[str, Any] = {}
         self._leverage_tiers: Dict[str, List[Dict]] = {}
         self.loop = asyncio.new_event_loop()
         asyncio.set_event_loop(self.loop)
@@ -453,7 +454,7 @@ class Exchange:
             self._load_async_markets()
             self._last_markets_refresh = arrow.utcnow().int_timestamp
             if self._ft_has['needs_trading_fees']:
-                self.trading_fees = self.fetch_trading_fees()
+                self._trading_fees = self.fetch_trading_fees()
 
         except ccxt.BaseError:
             logger.exception('Unable to initialize markets.')
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -75,6 +75,7 @@ class Exchange:
         "mark_ohlcv_price": "mark",
         "mark_ohlcv_timeframe": "8h",
         "ccxt_futures_name": "swap",
+        "needs_trading_fees": False,  # use fetch_trading_fees to cache fees
     }
     _ft_has: Dict = {}
     _ft_has_futures: Dict = {}
@@ -451,6 +452,9 @@ class Exchange:
             self._markets = self._api.load_markets()
             self._load_async_markets()
             self._last_markets_refresh = arrow.utcnow().int_timestamp
+            if self._ft_has['needs_trading_fees']:
+                self.trading_fees = self.fetch_trading_fees()
+
         except ccxt.BaseError:
             logger.exception('Unable to initialize markets.')
 
--- a/freqtrade/exchange/gateio.py
+++ b/freqtrade/exchange/gateio.py
@@ -1,6 +1,7 @@
 """ Gate.io exchange subclass """
 import logging
-from typing import Dict, List, Tuple
+from datetime import datetime
+from typing import Dict, List, Optional, Tuple
 
 from freqtrade.enums import MarginMode, TradingMode
 from freqtrade.exceptions import OperationalException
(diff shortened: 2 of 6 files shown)


freqtrade/exchange/exchange.py
```diff
@@ -1299,6 +1304,27 @@ def fetch_positions(self) -> List[Dict]:
         except ccxt.BaseError as e:
             raise OperationalException(e) from e
 
+    @retrier
+    def fetch_trading_fees(self) -> Dict[str, Any]:
+        """
+        Fetch user account trading fees
+        Can be cached, should not update often.
+        """
+        if (self._config['dry_run'] or self.trading_mode != TradingMode.FUTURES
```
samgermain, line 1313: why only futures?

freqtrade/exchange/gateio.py
```diff
@@ -42,6 +46,26 @@ def validate_ordertypes(self, order_types: Dict) -> None:
...
+            # Therefore we patch the response with fees if it's not available.
+            # An alternative also contianing fees would be
+            # privateFuturesGetSettleAccountBook({"settle": "usdt"})
+
+            pair_fees = self._trading_fees.get(pair, {})
+            if pair_fees and pair_fees['taker'] is not None:
+                order['fee'] = {
+                    'currency': self.get_pair_quote_currency(pair),
+                    'cost': abs(order['cost']) * pair_fees['taker'],
+                    'rate': pair_fees['taker'],
```
samgermain: why always taker? I think we can determine if its taker or maker if `create_time == finish_time` in the `info` of the order, if they're equal it's a taker order

samgermain: Taker Order

```
{
  id: '...',
  clientOrderId: 'api',
  timestamp: 1648300969636,
  datetime: '2022-03-26T13:22:49.636Z',
  lastTradeTimestamp: 1648300969636,
  status: 'closed',
  symbol: 'XRP/USDT:USDT',
  type: 'limit',
  timeInForce: 'GTC',
  postOnly: false,
  side: 'buy',
  price: 0.88,
  stopPrice: undefined,
  average: 0.8272,
  amount: 1,
  cost: 8.272,
  filled: 1,
  remaining: 0,
  fee: undefined,
  fees: [],
  trades: [],
  info: {
    id: '...',
    contract: 'XRP_USDT',
    mkfr: '0',
    tkfr: '0.0005',
    tif: 'gtc',
    is_reduce_only: false,
    create_time: '1648300969.636',
    finish_time: '1648300969.636',
    price: '0.88',
    size: '1',
    refr: '0.3',
    left: '0',
    text: 'api',
    fill_price: '0.8272',
    user: '',
    finish_as: 'filled',
    status: 'finished',
    is_liq: false,
    refu: '2436035',
    is_close: false,
    iceberg: '0'
  }
}
```

samgermain: I suppose the taker vs maker won't work here for partially filled orders

samgermain:
I think we should use the `privateFuturesGetSettleAccountBook` endpoint, it returns the exact fees

```
% gateio privateFuturesGetSettleAccountBook '{"settle": "usdt"}' | head -n 15
2022-03-26T13:12:50.051Z
Node.js: v14.17.0
CCXT v1.77.19
gateio.privateFuturesGetSettleAccountBook ({"settle": "usdt"})
2022-03-26T13:12:51.747Z iteration 0 passed in 232 ms

      time |          change |          balance |                  text | type
------------------------------------------------------------------------------
1648300350 |      -0.0056535 |  11.544234439041 | ADA_USDT:141231591425 |  fee
1648299540 |      -0.0041445 |  11.549887939041 | XRP_USDT:141228822644 |  fee
1648284921 |       -0.004151 |  11.554032439041 | XRP_USDT:141186548408 |  fee
1648281600 |       -0.000834 |  11.558183439041 |              XRP_USDT | fund
1648252800 |       -0.000824 |  11.559017439041 |              XRP_USDT | fund
1648224000 |      -0.0008319 |  11.559841439041 |              XRP_USDT | fund
```

I also just realized that `text` is the market id, followed by the order id, so if it's split into multiple orders, you can get all the ones where the number following the colon is the order id

freqtrade/exchange/exchange.py
xmatthias: because fee detection from "trades" works in spot markets.

using `fetch_my_trades()` i can get the appropriate "fee" entry (e.g.:)
That's not working for futures, as there the `fetch_my_trades` endpoint doesn't contain fees.

```
[{'info': {'id': '3011005184',
   'create_time': '1647285410',
   'create_time_ms': '1647285410867.203000',
   'currency_pair': 'SOL_USDT',
   'side': 'sell',
   'role': 'maker',
   'amount': '0.25',
   'price': '79.1649',
   'order_id': '132746780837',
   'fee': '0',
   'fee_currency': 'USDT',
   'point_fee': '0.03958245',
   'gt_fee': '0'},
  'id': '3011005184',
  'timestamp': 1647285410867,
  'datetime': '2022-03-14T19:16:50.867Z',
  'symbol': 'SOL/USDT',
  'order': '132746780837',
  'type': None,
  'side': 'sell',
  'takerOrMaker': 'maker',
  'price': 79.1649,
  'amount': 0.25,
  'cost': 19.791225,
  'fee': {'cost': 0.0, 'currency': 'USDT'},      <<< we will read this
  'fees': [{'currency': 'USDT', 'cost': 0.0}]}]
```

xmatthias:
> I think we should use the `privateFuturesGetSettleAccountBook` endpoint, it returns the exact fees

I've been thinking about this, too (hence the comment about an alternative), but i'm hesitant if that will work properly / stable.
If we can't find the `<contract>:orderid` combination in the settlebook - can we then rely on the fact that it had 0 fees, or did we just miss timing (too early so it's not yet showing up?)


It'll also add an additional call - not that this would be relevant for gateio which has pretty nice ratelimits - but it's a delay still.

I've done the math, comparing the result of the method now with the result of `privateFuturesGetSettleAccountBook` - and it's a 100% match.

freqtrade/exchange/gateio.py
xmatthias: actually, i didn't want to go down that path yet (intentionally).

it would "potentially" be the correct way to do - but i don't think detection from timestamps will work at all times (assuming the exchange may have a slight delay from time to time in high pressure situations).

samgermain:
> > I think we should use the `privateFuturesGetSettleAccountBook` endpoint, it returns the exact fees
> 
> I've been thinking about this, too (hence the comment about an alternative), but i'm hesitant if that will work properly / stable. If we can't find the `<contract>:orderid` combination in the settlebook - can we then rely on the fact that it had 0 fees, or did we just miss timing (too early so it's not yet showing up?)
> 
> It'll also add an additional call - not that this would be relevant for gateio which has pretty nice ratelimits - but it's a delay still.
> 
> I've done the math, comparing the result of the method now with the result of `privateFuturesGetSettleAccountBook` - and it's a 100% match.

It is a match, but we'll have to figure out a way to determine if it's maker or taker, and I don't know if we can do that correctly with orders that don't all fill at once

freqtrade/exchange/exchange.py
samgermain: > because fee detection from "trades" works in spot markets.
> 
> using `fetch_my_trades()` i can get the appropriate "fee" entry (e.g.:) That's not working for futures, as there the `fetch_my_trades` endpoint doesn't contain fees.
> 
> ```
> [{'info': {'id': '3011005184',
>    'create_time': '1647285410',
>    'create_time_ms': '1647285410867.203000',
>    'currency_pair': 'SOL_USDT',
>    'side': 'sell',
>    'role': 'maker',
>    'amount': '0.25',
>    'price': '79.1649',
>    'order_id': '132746780837',
>    'fee': '0',
>    'fee_currency': 'USDT',
>    'point_fee': '0.03958245',
>    'gt_fee': '0'},
>   'id': '3011005184',
>   'timestamp': 1647285410867,
>   'datetime': '2022-03-14T19:16:50.867Z',
>   'symbol': 'SOL/USDT',
>   'order': '132746780837',
>   'type': None,
>   'side': 'sell',
>   'takerOrMaker': 'maker',
>   'price': 79.1649,
>   'amount': 0.25,
>   'cost': 19.791225,
>   'fee': {'cost': 0.0, 'currency': 'USDT'},      <<< we will read this
>   'fees': [{'currency': 'USDT', 'cost': 0.0}]}]
> ```

But this is a general exchange method, not a gateio method, so maybe for other exchanges this isn't correct

freqtrade/exchange/gateio.py
samgermain: > actually, i didn't want to go down that path yet (intentionally).
> 
> it would "potentially" be the correct way to do - but i don't think detection from timestamps will work at all times (assuming the exchange may have a slight delay from time to time in high pressure situations).

It's kind of the best we have though, unless we just want to do limit(with order_side=bid for buys and order_side=ask for sells ) = maker, market = taker

xmatthias:
> It is a match, but we'll have to figure out a way to determine if it's maker or taker, and I don't know if we can do that correctly with orders that don't all fill at once

how does a "plain" maker order show up in the settleaccount endpoint?
based on the fee paid (0) - i'd assume it'll not be there (which is a problem).

assuming that'll not be there - how can we be sure if we paid 0 fee - or if the detection somehow didn't work?

samgermain:
@xmatthias 

So `fetchMyTrades` tells you if it's a taker or maker order, sorry I didn't see that

```
[
  {
    info: {
      size: '26',
      order_id: '...',
      id: '...',
      role: 'taker',
      create_time: '1646863377.3267',
      contract: 'XRP_USDT',
      price: '0.7631'
    },
    id: '...',
    timestamp: 1646863377326,
    datetime: '2022-03-09T22:02:57.326Z',
    symbol: 'XRP/USDT:USDT',
    order: '...',
    side: 'buy',
    takerOrMaker: 'taker',
    price: 0.7631,
    amount: 26,
    cost: 198.406,
    fee: { cost: undefined, currency: 'GT' },
    fees: [ { currency: 'GT', cost: undefined } ]
  },
  ```

freqtrade/exchange/gateio.py
xmatthias: we can get this from my_trades - good spot! :+1:

freqtrade/exchange/exchange.py
xmatthias: well most exchanges provide fee as part of the (filled) order response.

we only ever query `fetch_my_trades` for exchanges that don't (binance, gateio futures) that i know off) .
until now - fees detected from here were working fine - and i'll worry about a wrong fee case once i encounter them as being wrong.

xmatthias:
@samgermain that's a nice spot - what do you think now?

samgermain:
> @samgermain that's a nice spot - what do you think now?

looks good

PR #6603, opened 2022-03-28
Fix using future data to fill when use timeout

adrianceding:

## Summary

Fix using future data to fill when use timeout

Solve the issue: #___

## Quick changelog

- Modify the processing order of timeout

## What's new?

Current order:
1. Process buys.
2. Process entry orders.
3. Create sell orders (if any)
4. Process sell orders.
5. Cancel expired buy/sell orders.

Timeframe: 5m
If i set buy timeout to 10s.

**2** create the buy order.
**5** think there is no timeout

To next loop timeframe
**2** thinks that the order can be filled. But in fact, this should be a cancelled order


xmatthias:
I don't really see how this would fix the problem you're seeing. 

using a timeout < timeframe should be handled separately for sure (this scenario can only be filled in the original candle or must timeout). 
But i don't really see how changing the sequence will archive this. Won't this simply move the problem from buying to selling (or viceversa) ?

adrianceding:
Sorry. Now this should work?

adrianceding:
Yes, I used the dumbest method. Improved

xmatthias:
> Yes, I used the dumbest method. Improved

That's not true - your original method was actually better from both an efficiency standpoint (does not do the same calculation twice) - as well as from a logic perspective.

It was meant / intended as a question - but i think i've now understood the logic/problem you're seeing.
As such - i've rolled the last 2 commits back (and instead added an update to the documentation for this).

adrianceding:
But from a performance point of view. Shouldn't the performance be better by only looping once?
The same functionality can be achieved both ways. But the previous code requires an extra loop

xmatthias:
that depends on the scenario i think - we loop more - but don't do date math more often instead (which is init a timedelta object and add it to a date).

the bigger part is the logic one - doing the "cancel" at the beginning aligns it with the logic in the main loop.
while that's not 100% relevant, as the main trading loop will run every x seconds - while backtesting will only be called once per candle.

PR #6604, opened 2022-03-28
Corrected test_create_order

SmartManoj:

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

### DIFF
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -1098,7 +1098,7 @@ def test_create_order(default_conf, mocker, side, ordertype, rate, marketprice, 
     exchange = get_patched_exchange(mocker, default_conf, api_mock, id=exchange_name)
 
     order = exchange.create_order(
-        pair='ETH/BTC', ordertype=ordertype, side=side, amount=1, rate=200)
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
LGTM 👍

SmartManoj:
Previously, rate was passed for a market order with `createMarketBuyOrderRequiresPrice:None` and it is ignored.
It would be better to rename rate to api_rate or rate_for_order? and revert the changes to check whether it is ignored or not?

xmatthias:
rate is not technically ignored - it's still used to determine the amount, also on market orders.

Testing the "non-use" was for sure never the intend in this test - so i think this change is fine.

SmartManoj:
> rate is not technically ignored - it's still used to determine the amount, also on market orders.

amount is passed to `create_order`

xmatthias:
it still isn't asserted if rate is passed to the api call or not - so that's clearly not the intend of the test.

SmartManoj:
It is asserted.
https://github.com/freqtrade/freqtrade/blob/ddb0254999f920dbcd07abe366185cc4b4b9bd3e/tests/exchange/test_exchange.py#L1110

PR #6608, opened 2022-03-29
Add few lines to docs

stash86:

## Summary

Add more info on shorting setup

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?


docs/includes/pricing.md
```diff
@@ -116,3 +116,23 @@ Assuming both buy and sell are using market orders, a configuration similar to t
 ```
 
 Obviously, if only one side is using limit orders, different pricing combinations can be used.
+
+### Futures market's order pricing
+
+When trading on futures market. orderbook must be used because there is no ticker data for futures.
```
xmatthias: This is not correct in general - but only valid for binance futures (as their futures ticker data does not contain prices).

If any, it should therefore be moved to the `exchanges.md` to the binance section - with a specific "futures" reference / subtitle...

docs/strategy-customization.md
```diff
@@ -235,7 +235,7 @@ def populate_entry_trend(self, dataframe: DataFrame, metadata: dict) -> DataFram
     Short-trades need to be supported by your exchange and market configuration!
     Please make sure to set [`can_short`]() appropriately on your strategy if you intend to short.
 
-    ```python
+```python
```
xmatthias: I know markdown lint tools will flag this - but it's intentionally indented - as that will move it into the (collapsed) "note" box once you render the documentation

docs/configuration.md
```diff
@@ -100,6 +100,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `trailing_only_offset_is_reached` | Only apply trailing stoploss when the offset is reached. [stoploss documentation](stoploss.md). [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*  <br> **Datatype:** Boolean
 | `fee` | Fee used during backtesting / dry-runs. Should normally not be configured, which has freqtrade fall back to the exchange default fee. Set as ratio (e.g. 0.001 = 0.1%). Fee is applied twice for each trade, once when buying, once when selling. <br> **Datatype:** Float (as ratio)
 | `trading_mode` | Specifies if you want to trade regularly, trade with leverage, or trade contracts whose prices are derived from matching cryptocurrency prices. [leverage documentation](leverage.md). <br>*Defaults to `"spot"`.*  <br> **Datatype:** String
+| `can_short` | Specifies if the bot can do futures shorting trade. <br>[Strategy Override](#parameters-in-the-strategy) <br>*Defaults to `false`.*  <br> **Datatype:** Boolean
```
xmatthias: This parameter should be treated similar to `INTERFACE_VERSION`.

neither of the 2 can be set via the configuration (intentionally) - which is why i didn't add it to this part of the docs (as it's not really set via the configuration).

It's also documented in the strategy part, so i'm not sure if it makes sense to have it elsewhere.

docs/strategy-customization.md
```diff
@@ -256,7 +256,26 @@ def populate_entry_trend(self, dataframe: DataFrame, metadata: dict) -> DataFram
...
+            (
+                (qtpylib.crossed_below(dataframe['rsi'], 70)) &  # Signal: RSI crosses below 70
+                (dataframe['tema'] > dataframe['bb_middleband']) &  # Guard
+                (dataframe['tema'] < dataframe['tema'].shift(1)) &  # Guard
+                (dataframe['volume'] > 0)  # Make sure Volume is not 0
+            ),
+            ['enter_short', 'enter_tag']] = (1, 'rsi_cross')
+
+        return dataframe
+```
```
xmatthias: Instead of this - maybe add a small side-sentence above to `This method will also define a new column, "enter_long", ` - which might be something like `... (which is a mandatory column)`.

Alternatively add it to the "enter short trades" notes box.

Otherwise, we have 3 consecutive "note" boxes - which in my experience leads to people not reading it's content (like, not at all).

docs/strategy_migration.md
```diff
@@ -18,6 +18,7 @@ If you intend on using markets other than spot markets, please migrate your stra
   * `sell` -> `exit_long`
   * `buy_tag` -> `enter_tag` (used for both long and short trades)
   * New column `enter_short` and corresponding new column `exit_short`
+  * `enter_long` MUST be set even if the strategy is a shorting strategy.
```
xmatthias: That's not a change, really - "buy" was also a mandatory column, and simply remained exactly that. Now this may change a little later (maybe before we move this to stable) - but honestly, i think a pure short strategy is quite rare - you'll usually want to exploit both sides of the market depending on market conditions - which is why i see this as an edge-case condition.


If you think it should be explicitly stated, then please add it to the extended documentation below - not to the quick summary.

stash86: this has been covered in entry signal rules on strategy-customization, so I removed this

xmatthias (approved):
I've updated it to the new wording so it matches feat/short in it's current state (merged 20 minutes ago - so don't worry) 

Nice one :+1:

PR #6615, opened 2022-03-29
Add selection buttons for trades to forcesell cmd in telegram

cyberjunky:
## Summary

Add the option to select a trade for forcesell via telegram client, similar to forcebuy option.

Solve the issue: #___ (new feature)

## Quick changelog

Date:   Tue Mar 29 19:19:07 2022 +0200

    Add selection buttons for trades to forcesell cmd in telegram

Date:   Tue Mar 29 19:41:49 2022 +0200

    Allow forcesell to be a valid keyboard option

## What's new?

forcesell


cyberjunky:
I will update tests tomorrow.

cyberjunky:
Can you tell me how i run the tests locally?
Are there mode developer notes available somewhere?

xmatthias:
> Can you tell me how i run the tests locally? Are there mode developer notes available somewhere?

I was just going to suggest the same :laughing: 

ofc - [in the documentation](https://www.freqtrade.io/en/latest/developer/) - or in contributing.md in the root directory of the repository.

freqtrade/rpc/telegram.py
```diff
@@ -114,8 +114,9 @@ def _init_keyboard(self) -> None:
                                  r'/stopbuy$', r'/reload_config$', r'/show_config$',
                                  r'/logs$', r'/whitelist$', r'/blacklist$', r'/bl_delete$',
                                  r'/weekly$', r'/weekly \d+$', r'/monthly$', r'/monthly \d+$',
-                                 r'/forcebuy$', r'/forcelong$', r'/forceshort$',
-                                 r'/edge$', r'/health$', r'/help$', r'/version$']
+                                 r'/forcebuy$', r'/forcesell$', r'/edge$', r'/health$', r'/help$',
```
xmatthias: You'll need to be careful here - this removes `/forcelong` (among others) - which is clearly not the intend (i assume it's a side-effect of merging develop).

freqtrade/rpc/telegram.py
```diff
@@ -196,6 +197,8 @@ def _init(self) -> None:
                                  pattern='update_sell_reason_performance'),
             CallbackQueryHandler(self._mix_tag_performance, pattern='update_mix_tag_performance'),
             CallbackQueryHandler(self._count, pattern='update_count'),
+            CallbackQueryHandler(self._forcebuy_inline, pattern=r"\S+\/\S+"),
+            CallbackQueryHandler(self._forcesell_inline, pattern=r"[0-9]+\s\S+\/\S+"),
```
xmatthias: I'll have to ask you to align the naming of the new segments to the new pattern - which will be `forceenter` and `forceexit` instead of *buy and *sell.

Usually, `buy` should map 1:1 to `enter` (same with sell -> exit) - as we used to only do long trades - so an entry was always a buy.

freqtrade/rpc/telegram.py
cyberjunky: Thanks for being so patients, will fix and try to be more careful ;-)

freqtrade/rpc/telegram.py
cyberjunky: No problems, I just used the naming which was already there, but will rename it.

cyberjunky: O I see you renamed a lot, just as I grasped to understand the code. What about the RPC calls?
Is this done on purpose, doesn't look very consistent naming?

``` 
    def _rpc_forceexit(self, trade_id: str, ordertype: Optional[str] = None) -> Dict[str, str]:
..
    def _rpc_force_entry(self, pair: str, price: Optional[float], *,
```

xmatthias: it's more about the name than the "format" itself. It's inconsistent - yes - but it's also not really part of a public interface.
That said, it's also simple to rename due to that :laughing:

tests/rpc/test_rpc_telegram.py
```diff
@@ -1208,15 +1208,6 @@ def test_forcesell_handle_invalid(default_conf, update, mocker) -> None:
     assert msg_mock.call_count == 1
     assert 'not running' in msg_mock.call_args_list[0][0][0]
 
-    # No argument
```
xmatthias, line 1215: I'd  appreciate a quick test on this...

maybe look at `test_forceenter_no_pair` for an example - while it doesn't test the "full flow" - it can still test most parts of it - at least making sure there's no broken code (in the future ...)

xmatthias (approved):
That's a great feature! :+1: 

I've added a few tests, and cleaned the code up a bit (it actually broke `/forcebuy` as it was)

cyberjunky:
No thanks, sorry for the lack of time and internal freqtrade code knowledge to fix/improve it myself.
And I was unlucky with the developer branch changes in between. ;-P

 I hope I have some more time in the future, I certainly have some other ideas.

xmatthias:
> And I was unlucky with the developer branch changes in between. ;-P

Absolutely - and i'm sorry for that (but there's never a good time for a big change - there'll always be some open PR's around).

> I certainly have some other ideas.

looking forward to these - we'll not merge another "big change" anytime soon i think :laughing:

PR #6618, opened 2022-03-30
Add trading mode to list-pairs and list-markets output in docs

samgermain:
Solves this: https://trello.com/c/wXWIwjA2

I sent the output of all these commands to a file on develop and feat/short and ran command-line `diff` to see where the output was different between the two branches, and these were the only places where the docs weren't updated


xmatthias:
> these were the only places where the docs weren't updated

That's ... surprising at best - i was expecting commands to be pretty missaligned
Even better - thanks for taking care of this! :+1:

PR #6624, opened 2022-04-01
docs for shorting

samgermain:
## Summary

Adds docs for whats required to short

## Quick changelog

- docs for can_short strategy setting
- docs for allowed `trading_mode` and `margin_mode` when shorting

### DIFF
--- a/docs/leverage.md
+++ b/docs/leverage.md
@@ -22,6 +22,8 @@ Shorting is not possible when trading with [`trading_mode`](#understand-tradingm
 
 For a strategy to short, the strategy class must set the class variable `can_short = True`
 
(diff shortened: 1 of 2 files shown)


docs/leverage.md
```diff
@@ -13,6 +13,17 @@
     Please only use advanced trading modes when you know how freqtrade (and your strategy) works.
     Also, never risk more than what you can afford to lose.
 
+
+Please read the [strategy migration guide](https://www.freqtrade.io/en/latest/strategy_migration/#strategy-migration-between-v2-and-v3) to migrate your strategy from a freqtrade v2 strategy, to v3 strategy that can short and trade futures
```
xmatthias: we shouldn't use absolute links here. This would link to "latest" - but should that ever change, then the stable docs are suddenly out of date.

Fixed - but putting this here for future reference

xmatthias (approved):

PR #6627, opened 2022-04-01
Added bot_start callback to strategy interface

samgermain:
A bot_start callback for the strategy than is called once when freqtrade starts running. Differs from `__init__` in that `dataprovider` and `wallets` have been assigned and are available within `bot_start` but not within `__init__`

@xmatthias If you don't like this approach, we can also try passing the dataprovider and wallets to the `__init__` method. I just think there should be a way to access this information in a method that is only called once

This callback also has the benefit of not needing to call `super().__init__(config)`, which might be easier for novices

---------------

If you do like this approach, I'll add the documentation and tests for it

### DIFF
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -39,7 +39,10 @@ class AwesomeStrategy(IStrategy):
         Called only once after bot instantiation.
         :param **kwargs: Ensure to keep this here so updates to this won't break your strategy.
         """
-        self.dp['remote_data'] = requests.get('https://some_remote_source.example.com')
+        if self.config['runmode'].value in ('live', 'dry_run'):
+            # Assign this to the class by using self.*
+            # can then be used by populate_* methods
+            self.remote_data = requests.get('https://some_remote_source.example.com')
 
 ```
 ## Bot loop start
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -24,11 +24,11 @@ Currently available callbacks:
 
 ## Bot start
 
-A simple callback which is called once when the bot starts.
+A simple callback which is called once when the strategy is loaded.
 This can be used to perform actions that must only be performed once and runs after dataprovider and wallet are set
 
 ``` python
-import asyncio
+import requests
 
 class AwesomeStrategy(IStrategy):
 
@@ -42,7 +42,7 @@ class AwesomeStrategy(IStrategy):
         if self.config['runmode'].value in ('live', 'dry_run'):
             # Assign this to the class by using self.*
             # can then be used by populate_* methods
-            self.remote_data = requests.get('https://some_remote_source.example.com')
+            self.cust_remote_data = requests.get('https://some_remote_source.example.com')
 
 ```
 ## Bot loop start
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -34,17 +34,12 @@ class AwesomeStrategy(IStrategy):
 
     # ... populate_* methods
 
-    async def some_asynchronous_task(self):
-        self.dp['remote_data'] = requests.get('https://some_remote_source.example.com')
-
     def bot_start(self, **kwargs) -> None:
         """
         Called only once after bot instantiation.
         :param **kwargs: Ensure to keep this here so updates to this won't break your strategy.
         """
-        self.loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(self.loop)
-        self.loop.run_until_complete(self.some_asynchronous_task())
+        self.dp['remote_data'] = requests.get('https://some_remote_source.example.com')
 
 ```
 ## Bot loop start
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -146,11 +146,11 @@ See [Dataframe access](strategy-advanced.md#dataframe-access) for more informati
 
 ## Custom stoploss
 
-Called for open trade every throttling iteration (roughly every 5 seconds) until a trade is closed.
+Called for open trade every iteration (roughly every 5 seconds) until a trade is closed.
 
 The usage of the custom stoploss method must be enabled by setting `use_custom_stoploss=True` on the strategy object.
 
-The stoploss price can only ever move upwards - if the stoploss value returned from `custom_stoploss` would result in a lower stoploss price than was previously set, it will be ignored. The traditional `stoploss` value serves as an absolute lower level and will be instated as the initial stoploss (before this method is called for the first time for a trade).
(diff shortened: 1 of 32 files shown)


freqtrade/strategy/interface.py
```diff
@@ -206,6 +213,13 @@ def bot_loop_start(self, **kwargs) -> None:
         """
         pass
 
+    def on_whitelist_update(self, **kwargs) -> None:
```
xmatthias: while certainly a possibility - i'm not convinced on the usability of this callback.
first of all, it's not giving you any info (like prior and new pairlist).
but more importantly -
 using `bot_loop_start()` the strategy can keep track of this itself pretty easily (that's like 2 lines of code).

Knowing that this is the result of a "i'd need this" - i don't think this will be utilized enough to justify pulling it along for now.

samgermain: What do you think of just the `bot_start` callback?

xmatthias: `bot_start` can/will provide benefits for sure if it's implemented completely.

I understand this as a PoC for now - the final version will obviously have to also call the callback in other scenarios (backtesting, edge, ...).

freqtrade/freqtradebot.py
```diff
@@ -121,6 +121,8 @@ def update():
                     self._schedule.every().day.at(t).do(update)
         self.last_process = datetime(1970, 1, 1, tzinfo=timezone.utc)
 
+        self.strategy.bot_start()
```
samgermain, line 125: I can't think of why this would need to be in edge, but maybe it should be above `# Initializing Edge only if enabled`?

xmatthias: it's not about edge itself (maybe i wasn't 100% clear) - but about the assumption that users can rely on this to be ran when the strategy initializes.

Imagine a usecase where you load your machine learning model in this function (we assume this to be the `__init__`  for users - no? (otherwise it'd be kinda pointless).

if you run edge (edge-cli, to be precise) - then you'd also want to run this - otherwise your populate_* methods would fail - as your model would not be loaded.

we'd probably also want this for plotting ... but it's no problem to omit it in some areas if we have that documented (like we do for bot-loop_start - which also ain't called in the plotting scenario.

tests/test_plotting.py
```diff
@@ -58,6 +58,7 @@ def test_init_plotscript(default_conf, mocker, testdatadir):
     assert "ohlcv" in ret
     assert "TRX/BTC" in ret["ohlcv"]
     assert "ADA/BTC" in ret["ohlcv"]
+    assert default_conf['strategy'].bot_started is True
```
samgermain: I get why this is failing, but I'm not sure how I would test this out for plotting

tests/strategy/test_default_strategy.py
```diff
@@ -32,6 +32,7 @@ def test_strategy_test_v3(result, fee, is_short, side):
     assert type(indicators) is DataFrame
     assert type(strategy.populate_buy_trend(indicators, metadata)) is DataFrame
     assert type(strategy.populate_sell_trend(indicators, metadata)) is DataFrame
+    assert strategy.bot_started is True
```
samgermain: Not sure why this would be failing

tests/optimize/test_edge_cli.py
```diff
@@ -94,6 +94,7 @@ def test_edge_init(mocker, edge_conf) -> None:
     assert edge_cli.config == edge_conf
     assert edge_cli.config['stake_amount'] == 'unlimited'
     assert callable(edge_cli.edge.calculate)
+    assert edge_conf['strategy'].bot_started is True
```
samgermain: I get why this is failing, but I'm not sure how I would test this out for edge

tests/optimize/test_backtesting.py
```diff
@@ -312,6 +312,7 @@ def test_backtesting_init(mocker, default_conf, order_types) -> None:
     get_fee.assert_called()
     assert backtesting.fee == 0.5
     assert not backtesting.strategy.order_types["stoploss_on_exchange"]
+    assert backtesting.strategy.bot_started is True
```
samgermain, line 315: not sure why this would be failing

docs/strategy-callbacks.md
```diff
@@ -21,6 +22,31 @@ Currently available callbacks:
...
+        self.dp['remote_data'] = requests.get('https://some_remote_source.example.com')
+
+    def bot_start(self, **kwargs) -> None:
+        """
+        Called only once after bot instantiation.
+        :param **kwargs: Ensure to keep this here so updates to this won't break your strategy.
+        """
+        self.loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self.loop)
+        self.loop.run_until_complete(self.some_asynchronous_task())
```
xmatthias: This is a non-working example.

Please do NOT use asyncio for examples (especially not if they are faulty / non-working). The last line will block - so the bot will fail to start.

Therefore, please keep the example small and simple (for example do some fake-code to load some machine learning model).

docs/strategy-callbacks.md
```diff
@@ -34,17 +34,12 @@ class AwesomeStrategy(IStrategy):
...
-
     def bot_start(self, **kwargs) -> None:
         """
         Called only once after bot instantiation.
         :param **kwargs: Ensure to keep this here so updates to this won't break your strategy.
         """
-        self.loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(self.loop)
-        self.loop.run_until_complete(self.some_asynchronous_task())
+        self.dp['remote_data'] = requests.get('https://some_remote_source.example.com')
```
xmatthias: please make sure your samples are actually working.
using `dp` here will result in an exception - as dp is a class, and you're using it like a dict.

use `custom_info` as shown in the "storing information" documentation section.

tests/test_plotting.py
xmatthias: i don't think you can reliably test this - not the way plotting is setup (the strategy object is a local object in that context).

You could try to mock or patch the method and see if it's been called (that's what you're really after) - but i'm not sure that'll work with the strategy loader correctly.
it's most likely not worth the effort though - we don't have to have this tested for all cases i think.

tests/optimize/test_backtesting.py
xmatthias: it's not ... ?

tests/optimize/test_edge_cli.py
xmatthias: `assert edge_cli.strategy.bot_started is True` ?

tests/strategy/test_default_strategy.py
xmatthias: you can't test this here. It's initializing the strategy directly, independent of both strategy loader, as well as all other bot modules.
As such, it's expected that this is not called here.

xmatthias (approved):
slightly updated the sample, and moved the call in edge to the right class to avoid having it called twice.

LGTM :+1:

PR #6628, opened 2022-04-01
Fixes CoinGecko typos in fiat_convert

topscoder:
## Summary

Fixes log messages containing wrong naming for CoinGecko.

Solve the issue: #NaN

## Quick changelog

- Ensures consistent naming for CoinGecko in fiat_convert.py
- Ensures consistent naming for CoinGecko in test_fiat_convert.py


xmatthias (approved):
I'd not go as far as calling them typos - i don't think they were determined in HOW their name should be spelled in the past.

That said - thanks a lot for this! :100: 

LGTM :+1:

PR #6630, opened 2022-04-02
checking exchange name with lower

SmartManoj:
## Summary

args[0].name is `KuCoin`  | [ccxt name](https://github.com/ccxt/ccxt/blob/0afa036fbb04b9ea940bb3344d43891bf7b13b92/python/ccxt/kucoin.py#L32
)

config exchange name is changed here
https://github.com/freqtrade/freqtrade/blob/0b40c37cc54dbf97ff8ed1b82330e9d158d91d6b/freqtrade/configuration/configuration.py#L181
Where does `self.args` comes from?

### DIFF
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -102,7 +102,7 @@ def calculate_backoff(retrycount, max_retries):
 def retrier_async(f):
     async def wrapper(*args, **kwargs):
         count = kwargs.pop('count', API_RETRY_COUNT)
-        kucoin = args[0].name == "Kucoin"  # Check if the exchange is KuCoin.
+        kucoin = args[0].name.lower() == "kucoin"  # Check if the exchange is KuCoin.
         try:
             return await f(*args, **kwargs)
         except TemporaryError as ex:
--- a/freqtrade/exchange/common.py
+++ b/freqtrade/exchange/common.py
@@ -102,7 +102,7 @@ def calculate_backoff(retrycount, max_retries):
 def retrier_async(f):
     async def wrapper(*args, **kwargs):
         count = kwargs.pop('count', API_RETRY_COUNT)
-        kucoin = args[0].name.lower() == "kucoin"  # Check if the exchange is KuCoin.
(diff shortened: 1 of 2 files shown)


freqtrade/exchange/common.py
```diff
@@ -102,7 +102,7 @@ def calculate_backoff(retrycount, max_retries):
 def retrier_async(f):
     async def wrapper(*args, **kwargs):
         count = kwargs.pop('count', API_RETRY_COUNT)
-        kucoin = args[0].name == "Kucoin"  # Check if the exchange is KuCoin.
+        kucoin = args[0].name.lower() == "kucoin"  # Check if the exchange is KuCoin.
```
xmatthias: It's interesting that this was not noticed before ... as this effectively disables this fix completely.

considering the frequency this will be called, i think we should prefer to use the proper casing in the string instead.
That'll avoid converting a static (!) string to lowercase whenever we fetch ohlcv data.

xmatthias (commented):
There is no "self.args" here.

It's a decorator - so all arguments will be the arguments passed to the method.
As this decorator is used _exclusively_ on exchange methods, the first argument will always be the exchange instance (self - in the method itself).

This decorator would however fail if we decorate a static method with it - as then the first argument may not have a .name property.

xmatthias (approved):
Still find it interesting that noone noticed...

Good spot anyway :+1:

SmartManoj:
Used lower in case if it is changed in ccxt.


> exchange instance

https://github.com/freqtrade/freqtrade/blob/0b40c37cc54dbf97ff8ed1b82330e9d158d91d6b/freqtrade/exchange/exchange.py#L217

Exchange.name comes from config 

In config, exchange name is changed here
https://github.com/freqtrade/freqtrade/blob/0b40c37cc54dbf97ff8ed1b82330e9d158d91d6b/freqtrade/configuration/configuration.py#L181
I'm asking, here where does `self.args` comes from?

SmartManoj:
Used lower in case if it is changed in `ccxt`

xmatthias:
well its assigned in the `__init__` of this class.
just trace how that's instanciated - but i'm not sure why that's really relevant.

PR #6633, opened 2022-04-03
ROI and exit_tags matched to enter_tags

samgermain:
## Summary

Enables the ability to set roi and sell signals for trades, dependent on what the buy signal for that trade was

## Quick changelog

- interface.minimal_roi can now be indexed by the enter_tag, so the property could look like
- 
```
{
    '0': '0.02',  # default roi, used if there is no enter_tag, or the enter_tag is not a key within interface.minimal_roi
    '30': '0.00',    # default roi
    'bb_lower': {    # if the enter_tag for the trade is `'bb_lower'`
        '0': '0.03',
        '30': '0.00',
    },
    'bb_upper_short': {    # if the enter_tag for the trade is `'bb_upper_short'`
        '0': '0.05',
        '30': '0.00',
    }
}
```

- using `interface.tags`(name can change) you can set all the allowable exit signals for trades that were triggered by specific `enter_signals`

e.g.

```
tags = {
    'bb_lower': ['bb_upper'],
    'bb_upper_short': ['bb_lower', 'low_rsi']
}
```

If the enter_tag is not a key within `tags` then any sell_signal can be used for this trade

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -922,19 +922,18 @@ class FreqtradeBot(LoggingMixin):
             analyzed_df, _ = self.dataprovider.get_analyzed_dataframe(trade.pair,
                                                                       self.strategy.timeframe)
 
-
             (enter, exit_, exit_tag) = self.strategy.get_exit_signal(
                 trade.pair,
                 self.strategy.timeframe,
                 analyzed_df,
-                is_short=trade.is_short,
+                is_short=trade.is_short
             )
 
         logger.debug('checking exit')
         exit_rate = self.exchange.get_rate(
             trade.pair, side='exit', is_short=trade.is_short, refresh=True)
 
-        enter_tag = '' # TODO
+        enter_tag = # TODO
         if self._check_and_execute_exit(
             trade=trade,
             exit_rate=exit_rate,
@@ -1119,7 +1118,6 @@ class FreqtradeBot(LoggingMixin):
             exit_=exit_,
             force_stoploss=self.edge.stoploss(trade.pair) if self.edge else 0,
             enter_tag=enter_tag,
-            exit_tag=exit_tag,
         )
 
         if should_exit.exit_flag:
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -922,18 +922,19 @@ class FreqtradeBot(LoggingMixin):
             analyzed_df, _ = self.dataprovider.get_analyzed_dataframe(trade.pair,
                                                                       self.strategy.timeframe)
 
+
             (enter, exit_, exit_tag) = self.strategy.get_exit_signal(
                 trade.pair,
                 self.strategy.timeframe,
                 analyzed_df,
-                is_short=trade.is_short
+                is_short=trade.is_short,
             )
 
         logger.debug('checking exit')
         exit_rate = self.exchange.get_rate(
             trade.pair, side='exit', is_short=trade.is_short, refresh=True)
 
-        enter_tag = # TODO
+        enter_tag = '' # TODO
         if self._check_and_execute_exit(
             trade=trade,
             exit_rate=exit_rate,
@@ -1118,6 +1119,7 @@ class FreqtradeBot(LoggingMixin):
             exit_=exit_,
             force_stoploss=self.edge.stoploss(trade.pair) if self.edge else 0,
             enter_tag=enter_tag,
+            exit_tag=exit_tag,
         )
 
         if should_exit.exit_flag:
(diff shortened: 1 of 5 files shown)


freqtrade/freqtradebot.py
```diff
@@ -931,17 +931,27 @@ def handle_trade(self, trade: Trade) -> bool:
...
-                is_short=trade.is_short
+                is_short=trade.is_short,
             )
 
         logger.debug('checking exit')
         exit_rate = self.exchange.get_rate(
             trade.pair, side='exit', is_short=trade.is_short, refresh=True)
-        if self._check_and_execute_exit(trade, exit_rate, enter, exit_, exit_tag):
+
+        enter_tag = '' # TODO
```
samgermain: Not sure how this part would be done yet

freqtrade/strategy/interface.py
```diff
@@ -124,6 +124,8 @@ class IStrategy(ABC, HyperStrategyMixin):
     # Definition of plot_config. See plotting documentation for more details.
     plot_config: Dict = {}
 
+    tags: Dict[str, List[str]] = {}
```
xmatthias: i'm not sure i understand the intend  with this

do you assume users list all tags here (as well as in the code) - and freqtrade ignores them if they're missing?

I think that'll be a heavily breaking change the way it's implemented at the moment, breaking basically everyone's strategy assuming they use entry/exit tags.

samgermain: This is how it would work

using interface.tags(name can change) you can set all the allowable exit signals for trades that were triggered by specific enter_signals

e.g.

```
tags = {
    'bb_lower': ['bb_upper'],
    'bb_upper_short': ['bb_lower', 'low_rsi']
}
```

If the enter_tag is not a key within tags then any sell_signal can be used for this trade

so if they put nothing inside `tags` then nothing will be different than now, if they put a enter_tag inside `tags` then the only sell signals that can sell that trade are the ones listed in `tags[name_of_enter_tag]`, trades opened with any other enter_tag can sell by whatever means and aren't affected by `tags`

xmatthias: i fear a bit that this might lead to non-explicable "non-sells" (they are explainable if you look at them in VERY much detail - but non-obvious - which means difficult for users to grasp - which will mean a LOT of questions and a LOT of wasted time with very little benefit).

Consider the following:
Plotting functions (both `plotting.py` and freqUI) will show all sell signals (where sell_signal == 1 in the dataframe).
You can't really change this either - as it will depend what you're looking for if this is desired or not.

If you're now in a trade and it's not sold (but you see a sell signal) - you'll have no idea why - so from a support perspective, this might turn into a nightmare / unsupportable functionality.
You can obviously dig into the strategy, trying to identify this - but without access to the strategy (which often happens in issues) - this is near impossible to figure out.

This can also be done quite easily from within custom-exit callback - where it'll 
* not show misleading signals in the charts
* provide even more fine grained control (like - only do this if profit is > 1%)

So i think an improved example for `custom_exit()` will already satisfy this - in a better way as well (it'll leave control to the user - without resulting in confusing charts).

---

ROI is "simpler" in this regard - as it's not explicitly shown anywhere - but it'll be kindof missaligned then (why can we do this for ROI, but not for all other properties that control bot selling, like stoploss, use_sell_signal, ...).

samgermain: > ROI is "simpler" in this regard - as it's not explicitly shown anywhere - but it'll be kindof missaligned then (why can we do this for ROI, but not for all other properties that control bot selling, like stoploss, use_sell_signal, ...).

I was going to add stoploss, the goal of this PR was to be able to combine multiple strategies into one, so that if you have a strategy thats consistently profitable by small amounts, then you can include that strategy with another well performing strategy, and keep combining strategies that are profitable by a small amount into a bigger strategy thats profitable by a higher amount

xmatthias: well honeslty - i do see enter_tag as somewhat informative only - "some info the user wanted there to identify the signal". 

Internal freqtrade logic should not be based on it in my opinion. 
for now - i'm not 100% convinced on the benefits this will bring (not compared to the complexity it'll add for the user).
This is in part true for ROI (i can see some benefit there) - but definitely for the "tags" idea - which should clearly be done differently - within the user's strategy.

samgermain: > well honeslty - i do see enter_tag as somewhat informative only - "some info the user wanted there to identify the signal".
> 
> Internal freqtrade logic should not be based on it in my opinion. for now - i'm not 100% convinced on the benefits this will bring (not compared to the complexity it'll add for the user).

That was the original intention, but I don't see any reason to not expand the usage of it.

The roi part would be necessary if you wanted to combine high volatility strategies with low volatility strategies, as you would be looking at different kinds of roi

xmatthias: > That was the original intention, but I don't see any reason to not expand the usage of it.

For the moment - i do.
The way it is currently is slightly wrong - it's tied to the trade object - so you'll only have this signal on "initial entry".
In a DCA scenario, you may want to have the tag per order. 
Tying any logic to `enter_tag` (the initial one) now would prevent future enhancements in this area.

ROI per tag can be "faked" pretty easily with custom-sell - so i don't see this as providing enough benefit to block future, more useful enhancements.

PR #6635, opened 2022-04-03
added partial fills for dry run

SmartManoj:
## Summary

If we buy 2 coins at $2.5 and in order book only 1 coin is left for $2.5 and then the order should be partially filled. Previously it would be fully filled.


freqtrade/exchange/exchange.py
```diff
@@ -877,42 +877,67 @@ def get_dry_market_fill_price(self, pair: str, side: str, amount: float, rate: f
 
         return rate
 
-    def _is_dry_limit_order_filled(self, pair: str, side: str, limit: float) -> bool:
+    def _fill_dry_limit_order(self, pair: str, side: str,  # noqa: max-complexity: 13
```
xmatthias: Whenever you need to `noqa` something, then that's a warning sign that it should be done differently - maybe split into multiple functions.

The function itself doesn't look "that" complex though - are you sure this is really needed in it's current state?

freqtrade/exchange/exchange.py
```diff
@@ -877,42 +877,67 @@ def get_dry_market_fill_price(self, pair: str, side: str, amount: float, rate: f
...
-                    return True
-        except IndexError:
-            # Ignore empty orderbooks when filling - can be filled with the next iteration.
-            pass
-        return False
+                obd = ob['bids']
+            try:
+                logger.debug(
+                    f"{pair} checking dry {side}-order: "
+                    f"{limit=}, {given_amount=}, {order_book_top=}")
```
xmatthias, line 896: I'll need to double-check if the minimum python version we're requiring did include this or not (3.8 at this point).

If it includes this - then this is fine. 
I remember that this used to "not work" on older python versions - but i forgot which version this was introduced :(

tests/test_freqtradebot.py
```diff
@@ -3382,7 +3392,8 @@ def test_execute_trade_exit_with_stoploss_on_exchange(
         price_to_precision=lambda s, x, y: y,
         stoploss=stoploss,
         cancel_stoploss_order=cancel_order,
-        _is_dry_limit_order_filled=MagicMock(side_effect=[True, False]),
+        _fill_dry_limit_order=MagicMock(
+            side_effect=SideEffect(lambda *_: _[-2:], lambda *_: (None, 0))),
```
xmatthias, line 3396: can't this also be done by using an array passed to side_effect?

I know we use this in other places - but most likely not with functions.

freqtrade/exchange/exchange.py
SmartManoj: It is a brand new feature in 3.8
https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

tests/test_freqtradebot.py
SmartManoj: [side-effect is overloaded to serve a different purpose for iterables.](https://stackoverflow.com/questions/64992044/pytest-mock-multiple-calls-of-same-method-with-different-side-effect/64992350#:~:text=side-effect%20is%20overloaded%20to%20serve%20a%20different%20purpose%20for%20iterables)

freqtrade/exchange/exchange.py
SmartManoj: It's not need now.

freqtrade/exchange/exchange.py
xmatthias: Great, thanks for digging this out. I've been uncertain if this was added in 3.8 or 3.9.

however nothing in 3.8 is "brand new" anymore :laughing:

freqtrade/exchange/exchange.py
```diff
@@ -877,42 +877,67 @@ def get_dry_market_fill_price(self, pair: str, side: str, amount: float, rate: f
...
+        """
+        Returns average price and filled amount
+        """
         if not self.exchange_has('fetchL2OrderBook'):
-            return True
-        ob = self.fetch_l2_order_book(pair, 1)
-        try:
+            return limit, given_amount
+        for order_book_top in (None, 1000):
+            ob = self.fetch_l2_order_book(pair, order_book_top)
```
xmatthias, line 888: Passing "None" as value here is not supported and leads to an exception - and i'm actually surprised mypy isn't screaming at you - as tye type-hint on `fetch_l2_order_book()` defines it as int - not `Optional[int]`.

I'm anyway not sure why you're using None as first argument (or why you're looping at all)

xmatthias:
> If we buy 2 coins at $2.5 and in order book only 1 coin is left for $2.5 and then the order should be partially filled. Previously it would be fully filled.

Thinking about this a bit more ... are you sure this PR will provide the benefit we hope from this?

This partial fill will mostly happen on low volume pairs (you have to use quite some size to have this happen on BTC).
Assuming the above case - it's correct that the order will be partially filled.
Freqtrade will however retry to fill this order (or the remainder of it) after one iteration (~5s later in default settings).
Assuming the mid to low-volume pair - let's assume that the orderbook didn't move (no volume / price changes).
Then we'd be filling the remaining 1 coin at 2.5$ - as we didn't really consume this coin (dry-run does not have market-impact - which live would - so the best bid/ask would've moved in live).

Even if the gap is wider (we buy 10 coins instead of 1 - with only one available),  it'd just be more iterations it'll take to fill the order.

The big problem here is - we can't really track this reliably, as we won't know if the orderbook replenishes or not - even if we keep track of "we used 1 coin at 2.5$" - we'll not know if the 1 coin we still see in the orderbook the same coin (same seller/buyer) - or if this has been really filled/cancelled, and a new buyer/seller popped up to fill our order.

freqtrade/exchange/exchange.py
xmatthias, line 888: Passing `None` as value here is not supported and leads to an exception - and i'm actually surprised mypy isn't screaming at you - as tye type-hint on `fetch_l2_order_book()` defines it as int - not `Optional[int]`.

I'm anyway not sure why you're using None as first argument (or why you're looping at all). 
If you'd like to fetch one level only, i'd use `1` instead of None to get just one level (i'm not sure why though - as that might lead to 2 requests which are not really necessary).

xmatthias (changes requested):

freqtrade/exchange/exchange.py
SmartManoj: Coz mypy is only checking for freqtrade and scripts folder.

SmartManoj:
> you have to use quite some size to have this happen on BTC)

Order book contains .002 BTC (8$) amount also. If the price difference is high and the market moves opposite, it can happen on any pairs.

> Freqtrade will however retry to fill this order (or the remainder of it) after one iteration (~5s later in default settings).

Once timedout, it will exit with the current filled amount if any else only it will retry.

freqtrade/exchange/exchange.py
xmatthias: mypy is checking all files within the freqtrade directory (including this one).
I'm therefore not sure what you mean with your comment.

It's failing to properly detect this - which is probably a bug in mypy (caused by the decorator it appears - as removing the `@retrier` decorator causes mypy to point out errors).

Even if mypy is failing to detect this (this can also happen in other cases, where the type on the function is missing for example) - that's no excuse for submitting untested PR's.

xmatthias:
> Once timedout, it will exit with the current filled amount if any else only it will retry.

The point is - the effort this code goes through (code-complexity) to simulate a edge-case scenario is most likely too much to be worth including this.

To have the scenario you describe happen - you'll have to
* have the orderbook be lower than your order
* either
  * have the price move away from your order within one cycle
  * have the timeout in this cycle

If the 0.002 BTC remain on the orderbook for 20 seconds - then you'll have 4x 0.02 filled on the same order - as we've seen the same order again and again - and filled on it again and again (which will clearly not happen in real trading).

so this is attempting hard to simulate a behavior that's unlikely to happen (all while still getting it wrong in all but one edge-case - the one described above).

---

you also **clearly didn't test this PR** in action (the code raises exceptions the moment it get's hit the first time).
If you expect me to pursue this further (properly review this) - please show me that you've seen this work in dry run at least once (logs, messages, ...) With reasonable stake size.

Without that (consider it as proof that you did test the PR this time) - i'll not be accepting this PR - and will be closing this PR within  week.

xmatthias:
As mentioned above - i'll not accept PR's where it's clear that the contributor did not test it.

While i also test PR's before merging - i'll not accept blindly contributed code (or point out every line of flaw).
It MUST be a given that the PR is in a working state. We can help with tests, or to clean up things - but when it becomes clear that the function has not been ran once outside of unittests, the PR becomes a liability, and therfore a waste of time.

PR #6636, opened 2022-04-03
added pre-commit

SmartManoj:
## Summary

[pre-commit - A framework for managing and maintaining multi-language pre-commit hooks.](https://pre-commit.com/)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pycqa/flake8
+    rev: '4.0.1'
+    hooks:
+    -   id: flake8
+        args: 
+        -   max-line-length = 100,
+        -   max-complexity = 12
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v0.942'
+    hooks:
+    -   id: mypy
+        args: [--ignore-missing-imports]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,11 +5,30 @@ repos:
     rev: '4.0.1'
     hooks:
     -   id: flake8
-        args: 
-        -   max-line-length = 100,
-        -   max-complexity = 12
+        stages: [push]
+
 -   repo: https://github.com/pre-commit/mirrors-mypy
     rev: 'v0.942'
     hooks:
     -   id: mypy
-        args: [--ignore-missing-imports]
+        stages: [push]
+
+-   repo: https://github.com/pycqa/isort
+    rev: '5.10.1'
+    hooks:
+      - id: isort
+        name: isort (python)
+        stages: [push]
+
+# https://github.com/pre-commit/pre-commit/issues/761#issuecomment-394167542
+-   repo: local
+    hooks:
+    -   id: pytest
+        name: pytest
+        entry: venv/bin/pytest
+        language: script
+        pass_filenames: false
+        # alternatively you could `types: [python]` so it only runs when python files change
+        # though tests might be invalidated if you were to say change a data file
+        always_run: true
+        stages: [push]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,30 +5,17 @@ repos:
     rev: '4.0.1'
     hooks:
     -   id: flake8
-        stages: [push]
+        # stages: [push]
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
     rev: 'v0.942'
     hooks:
     -   id: mypy
-        stages: [push]
+        # stages: [push]
 
 -   repo: https://github.com/pycqa/isort
     rev: '5.10.1'
     hooks:
       - id: isort
         name: isort (python)
-        stages: [push]
-
-# https://github.com/pre-commit/pre-commit/issues/761#issuecomment-394167542
--   repo: local
-    hooks:
-    -   id: pytest
-        name: pytest
-        entry: venv/bin/pytest
-        language: script
-        pass_filenames: false
-        # alternatively you could `types: [python]` so it only runs when python files change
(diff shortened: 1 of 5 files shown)


xmatthias:
While i like the concept of pre-commit hooks - i'm personally not a huge fan of them (on big projects).

it will run on every commit - slowing down the development process if you do frequent commits.
flake8 + mypy take between 1s and 5-6s per run.

unless they do something clever to speed the runs up - i don't think we'll be able to use this - simply for performance reasons.

The "compatibility" is often only relevant once you're ready to push/publish things - not in intermediate commits.


i also have my doubts about the configuration itself. 
* flake8 is configured in `setup.cfg`. If we need to duplicate the configuration - then we will not be using precommit (also because your configuration is slightly different to the one we use in setup.cfg).
* mypy requires the path as argument when running manually - and will refuse to run without it - as such i think the provided configuration is wrong / non-functional.
* isort is completely missing - bug will break ci just as much.

 
Assuming the above are all "fixable" - please "developer setup instructions" to the developer docs (`developer.md`) - as well as `contributing.md`.

SmartManoj:
> flake8 + mypy take between 1s and 5-6s per run.

[ Pre-commit will run all hooks against currently staged files only.](https://pre-commit.com/#:~:text=this%20will%20run%20all%20hooks%20against%20currently%20staged%20files.)



> it will run on every commit - slowing down the development process if you do frequent commits.

Updated for pre-push.



> * flake8 is configured in `setup.cfg`. If we need to duplicate the configuration - then we will not be using precommit 
Didn't get.

> (also because your configuration is slightly different to the one we use in setup.cfg).

[pre-commit won’t ever pass untracked files to flake8 so excluding .git / .tox / etc. is unnecessary.
](https://flake8.pycqa.org/en/latest/user/using-hooks.html#:~:text=pre-commit%20won%E2%80%99t%20ever%20pass%20untracked%20files%20to%20flake8%20so%20excluding%20.git%20%2F%20.tox%20%2F%20etc.%20is%20unnecessary.) So gave separate config. Removed as there is no harm to pass.



> * mypy requires the path as argument when running manually - and will refuse to run without it - as such i think the provided configuration is wrong / non-functional.

It is passed by pre-commit internally. [Config source].(https://github.com/pre-commit/mirrors-mypy#using-mypy-with-pre-commit)

Added `isort`

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+    hooks:
+    -   id: pytest
+        name: pytest
+        entry: venv/bin/pytest
+        language: script
+        pass_filenames: false
+        # alternatively you could `types: [python]` so it only runs when python files change
+        # though tests might be invalidated if you were to say change a data file
+        always_run: true
+        stages: [push]
```
xmatthias, line 34: This one - clearly no.
pytest runs for ~1-2 minutes - so DX for this will be bad.

CONTRIBUTING.md
```diff
@@ -20,6 +20,8 @@ Best start by reading the [documentation](https://www.freqtrade.io/) to get a fe
 
 ## Before sending the PR
 
+Do the following if you disabled pre-commit hook when commiting.
```
xmatthias: while it tells you what to do without pre-commit - it's not documenting _how_ to set it up.

While `setup.sh` is nice - you can't rely on everyone using the `setup.sh` script - especially developers tend to install things slightly differently depending on their local setup.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pycqa/flake8
+    rev: '4.0.1'
+    hooks:
+    -   id: flake8
+        stages: [push]
```
xmatthias: > [Pre-commit will run all hooks against currently staged files only.](https://pre-commit.com/#:~:text=this%20will%20run%20all%20hooks%20against%20currently%20staged%20files.)

If you provide good arguments - then still reconfiguring it without waiting for an answer is ... premature ;) 

I actually liked the fact of running flake8 at every commit more after your argument / link.

The point of having pre-commit is to avoid "fix flake8" (or fix mypy") commits. 
assuming precommit does a decent job running it only on "changed" / staged files performance impact should be minimal.

if we only run it on push - we'll simply end up with that commit at the end - not removed.

.pre-commit-config.yaml
xmatthias: pytest is the one i'm not sure about ... 
pytest runs for ~2 minutes - and you can't really run it for "only changed files" - as it can't know which files affected which tests for certain (i doubt pre-commit tracks coverage by test and only runs what's necessary).

Therefore i think pytest should be either "optin" (so you have to do `git push --extra pytest` (if that's possible) - or not part of pre-commit at all.

If you know you're only fixing a documentation line (think - new contributor) - you'd not want to run 2 minutes of CI (you know you didn't change a code file).

xmatthias:
> It is passed by pre-commit internally. [Config source].(https://github.com/pre-commit/mirrors-mypy#using-mypy-with-pre-commit)

With the limitations described in this link - i'm not sure mypy makes sense.
It'd only evaluate internal mypy - but you can still have mypy wrong for everything imported - which basically removes 70-80% of the benefits it'd aim to bring.

.pre-commit-config.yaml
xmatthias: To summarize - i'd leave `flake8`, `mypy` and `isort` running on commit - and remove pytest completely for the moment.

Pytest is simply too slow to run here (also the configuration below fails on installations that don't use `venv` as directory).

xmatthias (approved):
I've removed the non-working parts from this.

I'd honestly appreciate if you could - before opening new PR's - make sure that the existing PR feedback is addressed.
Otherwise you'll have a ton of PR's open - without actually finishing anything.

it's not helping me at all if i need to finish every PR for you - and that's also something that will not happen for most PR's.

PR #6651, opened 2022-04-05
show reserved balance

SmartManoj:

## Summary

show reserved balance

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -239,7 +239,7 @@ class Telegram(RPCHandler):
             f"{emoji} *{msg['exchange']}:*"
             f" {enter_side['entered'] if is_fill else enter_side['enter']} {msg['pair']}"
             f" (#{msg['trade_id']})\n"
-            )
+        )
         message += f"*Enter Tag:* `{msg['enter_tag']}`\n" if msg.get('enter_tag', None) else ""
         message += f"*Amount:* `{msg['amount']:.8f}`\n"
         if msg.get('leverage') and msg.get('leverage', 1.0) != 1.0:
@@ -418,7 +418,8 @@ class Telegram(RPCHandler):
                 if prev_avg_price:
                     minus_on_entry = (cur_entry_average - prev_avg_price) / prev_avg_price
 
-                dur_entry = cur_entry_datetime - arrow.get(filled_orders[x-1]["order_filled_date"])
+                dur_entry = cur_entry_datetime - \
+                    arrow.get(filled_orders[x - 1]["order_filled_date"])
                 days = dur_entry.days
                 hours, remainder = divmod(dur_entry.seconds, 3600)
                 minutes, seconds = divmod(remainder, 60)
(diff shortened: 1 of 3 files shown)


SmartManoj:
Did I reduced cyclomatic complexity?

freqtrade/rpc/telegram.py
```diff
@@ -418,7 +418,8 @@ def _prepare_entry_details(self, filled_orders: List, base_currency: str, is_ope
                 if prev_avg_price:
                     minus_on_entry = (cur_entry_average - prev_avg_price) / prev_avg_price
 
-                dur_entry = cur_entry_datetime - arrow.get(filled_orders[x-1]["order_filled_date"])
+                dur_entry = cur_entry_datetime - \
+                    arrow.get(filled_orders[x - 1]["order_filled_date"])
```
xmatthias, line 422: please avoid unnecessary linebreaks (especially ones with `\` - which should not be used if possible). 
this line seems to be unchanged once it's reverted to one line.

freqtrade/wallets.py
```diff
@@ -214,6 +214,18 @@ def get_total_stake_amount(self):
...
+        """
+        Return the total currently reserved balance in stake currency.
+        Calculated as
+        (<open_trade stakes> + free amount) - tot_profit - available_capital
+        """
+        available_capital = self._config.get('available_capital') or 0
+        tot_profit = Trade.get_total_closed_profit()
+        open_stakes = Trade.total_open_trades_stakes()
+        available_balance = self.get_free(self._config['stake_currency'])
+        return available_balance + open_stakes - tot_profit - available_capital
```
xmatthias, line 227: This function is only covering half of the options.

There's 2 possibilities to limit the used capital - either via `available_capital`, or via `tradable_balance_ratio`.
You can have a look at the logic in `get_total_stake_amount()` (right above this function).

freqtrade/rpc/telegram.py
```diff
@@ -840,14 +841,20 @@ def _balance(self, update: Update, context: CallbackContext) -> None:
                             f"*{curr['currency']}:*\n"
                             f"\t`Available: {curr['free']:.8f}`\n"
                             f"\t`Balance: {curr['balance']:.8f}`\n"
-                            f"\t`Pending: {curr['used']:.8f}`\n"
-                            f"\t`Est. {curr['stake']}: "
-                            f"{round_coin_value(curr['est_stake'], curr['stake'], False)}`\n")
+                            f"\t`Pending: {curr['used']:.8f}`\n")
+                        if curr['stake'] == curr['stake']:
+                            reserved = self._rpc._freqtrade.wallets.get_reserved_stake_amount()
+                            curr_output += f"\t`Reserved: {reserved}`\n"
```
xmatthias, line 847: I think we'd want to have an update to `telegram.md` for this as well - explaining what "Reserved" actually means.
it's not 100% clear from the call alone (i've had to dig through the code to figure out what you actually mean).

freqtrade/rpc/telegram.py
```diff
@@ -840,14 +841,20 @@ def _balance(self, update: Update, context: CallbackContext) -> None:
                             f"*{curr['currency']}:*\n"
                             f"\t`Available: {curr['free']:.8f}`\n"
                             f"\t`Balance: {curr['balance']:.8f}`\n"
-                            f"\t`Pending: {curr['used']:.8f}`\n"
-                            f"\t`Est. {curr['stake']}: "
-                            f"{round_coin_value(curr['est_stake'], curr['stake'], False)}`\n")
+                            f"\t`Pending: {curr['used']:.8f}`\n")
+                        if curr['stake'] == curr['stake']:
+                            reserved = self._rpc._freqtrade.wallets.get_reserved_stake_amount()
```
xmatthias, line 846: This should probably be moved to `rpc.py` - and added to the stake-currency entry (instead of calling this here).

That way, it'll also be available for the API.

freqtrade/rpc/telegram.py
SmartManoj: It's a [bug in pycodestyle](https://github.com/PyCQA/pycodestyle/issues/248#issuecomment-213658514)

SmartManoj: flake8 ignores E226 due to [this](https://github.com/PyCQA/pycodestyle/issues/248). 

> unchanged
`x-1` to `x - 1`

xmatthias: hm - i noticed that these were ignored - but i assumed that's common for flake8?

xmatthias: Good catch - on a separate note we should try to enable this at some point.

we don't configure `ignore = ` for flake8 - so the default applies - which apparently hides all warnings - so configuring the "seemingly default" throws a TON of warnings ...

SmartManoj: > Good catch

Using `Sublime Text`  with `anaconda` package which uses `autopep8`

> please avoid unnecessary linebreaks (especially ones with `\` - which should not be used if possible).

But `black` uses braces.

xmatthias (changes requested):
Maybe the review wasn't clear - the above comments (done several weeks ago) were requests for changes.

Please work on **finishing existing PR's**, rather than creating new, (pointless) PR's if you aim to help this project.
Pointless Pr's changing random stuff around without _visible benefit_ just waste people's time - and are not helping this project.

xmatthias:
I'll assume that there is no longer interest in completing this PR.
if that's not the case, please address the above concerns and either reopen this, or create a new PR with the changes.

PR #6653, opened 2022-04-05
renamed forceentry forceexit

SmartManoj:

## Summary

renamed forceentry forceexit to force_entry force_exit for readability
renamed emergencyexit to emergency_exit for consistency

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?


SmartManoj:
Weird. one filled was missed

freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -316,7 +316,7 @@ class ForceEnterPayload(BaseModel):
     entry_tag: Optional[str]
 
 
-class ForceExitPayload(BaseModel):
+class Force_exitPayload(BaseModel):
```
xmatthias: We will not name classes like this.

Naming for Classes in python should be in `CamelCase` - which this was, therefore this is simply wrong.

xmatthias (changes requested):
In general, i'd suggest to ask (issue, discord) before doing a bigger change like this if there's no existing issue to avoid wasting time.

I'm not sure i agree with this in general that this IS in fact better naming.

---

For the telegram commands - i will definitely not accept this.
`/force_exit` is a lot worse from a usability perspective (think phone) than `/forceexit` the moment you have to type it - therefore we'll surely not be accepting that part, as this will make usability worse, not better.

---

For the other things - please consider that people are already using this / have migrated to this state.

Even though it's not in a release version yet - i don't really see the benefit of using these as "split" terminology. It'll force people to migrate again - which would mean that we'll also have to have double-migrations in place internally.

from `forcesell` to `forceexit` - which is already available
and then again from `forceexit` to `force_exit` - which is missing here - but would need to be added as well.

---

For **user-facing** things (all configs / strategy / tag/...) - i will only be able to accept this if you can provide a _clear, working migration_ for both the "prior" state, the "mid" state (the one in develop now) - and the "new" state.

We will obviously be able to remove the "mid" state after the release (assuming people in develop will upgrade frequently) - but for an initial step until the release, it's a must.

---

For internal things - which are **not exposed** to the user - it's probably fine for the most part (it actually doesn't matter how they are named and can be changed at will, as long as we change them everywhere).


---

There is also some time constraint on this - so if you intend to pursue this further, i'd appreciate quick implementation of the requested points. 
The longer the "mid" state takes, the more people will be affected - which means it's less likely that we're going to force people through this "double" migration.

freqtrade/rpc/api_server/api_schemas.py
SmartManoj: > `/force_exit` is a lot worse from a usability perspective (think phone) than `/forceexit` the moment you have to type it - therefore we'll surely not be accepting that part, as this will make usability worse, not better.

Added alias `/fe` as time is considered here.

SmartManoj:
> `/force_exit` is a lot worse from a usability perspective (think phone) than `/forceexit` the moment you have to type it - therefore we'll surely not be accepting that part, as this will make usability worse, not better.

Added alias `/fe` as time is considered here.

freqtrade/rpc/api_server/api_schemas.py
SmartManoj: updated

docs/telegram-usage.md
```diff
@@ -171,8 +171,8 @@ official commands. You can ask at any moment for help with `/help`.
 | `/locks` | Show currently locked pairs.
 | `/unlock <pair or lock_id>` | Remove the lock for this pair (or for this lock id).
 | `/profit [<n>]` | Display a summary of your profit/loss from close trades and some stats about your performance, over the last n days (all trades by default)
-| `/forceexit <trade_id>` | Instantly exits the given trade  (Ignoring `minimum_roi`).
-| `/forceexit all` | Instantly exits all open trades (Ignoring `minimum_roi`).
+| `/force_exit <trade_id>` | Instantly exits the given trade  (Ignoring `minimum_roi`).
+| `/force_exit all` | Instantly exits all open trades (Ignoring `minimum_roi`).
```
xmatthias: The alias is fine (should be added here, too) - but i'd still keep it at `/forceexit`, not `/force_exit`.

That aligns it to the majority of other commands (like `/forcelong` and `/forceshort`) - and using underscores on a phone is simply bad UX.

docs/telegram-usage.md
```diff
@@ -285,7 +285,7 @@ Starting capital is either taken from the `available_capital` setting, or calcul
 > **BINANCE:** Long ETH/BTC with limit `0.03400000` (`1.000000 ETH`, `225.290 USD`)
 
 Omitting the pair will open a query asking for the pair to trade (based on the current whitelist).
-Trades crated through `/forceentry` will have the buy-tag of `forceentry`.
+Trades crated through `/force_entry` will have the buy-tag of `force_entry`.
```
xmatthias:
```suggestion
Trades crated through `/forcelong` will have the buy-tag of `force_entry`.
```

freqtrade/rpc/api_server/api_v1.py
```diff
@@ -135,13 +135,13 @@ def show_config(rpc: Optional[RPC] = Depends(get_rpc_optional), config=Depends(g
     return resp
 
 
-# /forcebuy is deprecated with short addition. use ForceEntry instead
-@router.post('/forceenter', response_model=ForceEnterResponse, tags=['trading'])
+# /forcebuy is deprecated with short addition. use Force_entry instead
+@router.post('/force_enter', response_model=ForceEnterResponse, tags=['trading'])
```
xmatthias: for the API - please don't use this naming.
it's not used by people - and for an api, `/forceenter` is clearer.

This obviously applies not just here - but also for the rest-client, as well as `/forceexit`.

docs/telegram-usage.md
xmatthias: Thinking about the alias a bit more - `/fe` could be both `forceentry` or `forceexit` - so it might actually be missleading ...

mkavinkumar1: > Thinking about the alias a bit more - `/fe` could be both `forceentry` or `forceexit` - so it might actually be missleading ...

What about /fx?

SmartManoj: > Thinking about the alias a bit more - `/fe` could be both `forceentry` or `forceexit` - so it might actually be missleading ...

How about /fx?

xmatthias: yeah - that'll work :+1:

freqtrade/rpc/api_server/api_v1.py
SmartManoj: Changed. But https://stackoverflow.com/q/10302179

docs/telegram-usage.md
SmartManoj: > using underscores on a phone is simply bad UX.

What about for `/reload_config`?

Also it's not hard to type. In Gboard long press 'g' will put '_' when long press for symbols enabled.
Or https://www.computerworld.com/article/3048523/gboard-android-typing-shortcuts.html#:~:text=Gboard%20shortcut%20%232,could%20ever%20desire.

(like `/forcelong` and `/forceshort`
It's better to rename them too.
And alias can be added as
/fn force_enter
/fl force_long
/fs force_short

docs/faq.md
```diff
@@ -79,7 +79,7 @@ You can use "current" market data by using the [dataprovider](strategy-customiza
 
 ### Is there a setting to only SELL the coins being held and not perform anymore BUYS?
 
-You can use the `/stopbuy` command in Telegram to prevent future buys, followed by `/forceexit all` (sell all open trades).
+You can use the `/stopbuy` command in Telegram to prevent future buys, followed by `/force_exit all` (sell all open trades).
```
xmatthias:
```suggestion
You can use the `/stopbuy` command in Telegram to prevent future buys, followed by `/forceexit all` (sell all open trades).
```

docs/rest-api.md
```diff
@@ -145,10 +145,10 @@ python3 scripts/rest_client.py --config rest_config.json <command> [optional par
...
 | `delete_lock <lock_id>` | Deletes (disables) the lock by id.
 | `profit` | Display a summary of your profit/loss from close trades and some stats about your performance.
-| `forceexit <trade_id>` | Instantly exits the given trade  (Ignoring `minimum_roi`).
-| `forceexit all` | Instantly exits all open trades (Ignoring `minimum_roi`).
-| `forceenter <pair> [rate]` | Instantly enters the given pair. Rate is optional. (`forcebuy_enable` must be set to True)
-| `forceenter <pair> <side> [rate]` | Instantly longs or shorts the given pair. Rate is optional. (`forcebuy_enable` must be set to True)
+| `force_exit <trade_id>` | Instantly exits the given trade  (Ignoring `minimum_roi`).
+| `force_exit all` | Instantly exits all open trades (Ignoring `minimum_roi`).
+| `force_enter <pair> [rate]` | Instantly enters the given pair. Rate is optional. (`forcebuy_enable` must be set to True)
+| `force_enter <pair> <side> [rate]` | Instantly longs or shorts the given pair. Rate is optional. (`forcebuy_enable` must be set to True)
```
xmatthias: Please revert the changes to rpc method naming consistently, not just in some places.

freqtrade/rpc/telegram.py
```diff
@@ -115,7 +115,7 @@ def _init_keyboard(self) -> None:
                                  r'/stopbuy$', r'/reload_config$', r'/show_config$',
                                  r'/logs$', r'/whitelist$', r'/blacklist$', r'/bl_delete$',
                                  r'/weekly$', r'/weekly \d+$', r'/monthly$', r'/monthly \d+$',
-                                 r'/forcebuy$', r'/forcelong$', r'/forceshort$',
+                                 r'/forcebuy$', r'/forcelong$', r'/forceshort$', r'/forceexit$',
```
xmatthias: This is wrong here at this point in time.

issuing `/forceexit` without any parameter will not do anything other than showing a warning that either "all" or a trade number must be specified.

freqtrade/rpc/telegram.py
```diff
@@ -1359,23 +1359,24 @@ def _help(self, update: Update, context: CallbackContext) -> None:
...
             "------------\n"
             "*/start:* `Starts the trader`\n"
             "*/stop:* Stops the trader\n"
             "*/stopbuy:* `Stops buying, but handles open trades gracefully` \n"
-            "*/forceexit <trade_id>|all:* `Instantly exits the given trade or all trades, "
+            "*/force_exit <trade_id>|all:* `Instantly exits the given trade or all trades, "
             "regardless of profit`\n"
-            f"{forceenter_text if self._config.get('forcebuy_enable', False) else ''}"
+            "*/fe <trade_id>|all:* `Alias to /force_exit`"
+            f"{force_enter_text if self._config.get('forcebuy_enable', False) else ''}"
```
xmatthias: consistency ...

freqtrade/rpc/api_server/api_v1.py
```diff
@@ -135,13 +135,13 @@ def show_config(rpc: Optional[RPC] = Depends(get_rpc_optional), config=Depends(g
     return resp
 
 
-# /forcebuy is deprecated with short addition. use ForceEntry instead
+# /forcebuy is deprecated with short addition. use Force_entry instead
```
xmatthias: while it's only a comment - it should still be correct.

docs/telegram-usage.md
```diff
@@ -285,7 +286,7 @@ Starting capital is either taken from the `available_capital` setting, or calcul
 > **BINANCE:** Long ETH/BTC with limit `0.03400000` (`1.000000 ETH`, `225.290 USD`)
 
 Omitting the pair will open a query asking for the pair to trade (based on the current whitelist).
-Trades crated through `/forceentry` will have the buy-tag of `forceentry`.
+Trades crated through `/forcelong` will have the buy-tag of `force_entry`.
```
xmatthias:
```suggestion
Trades created through `/forcelong` will have the buy-tag of `force_entry`.
```

Not your typo - but still wrong :laughing:

docs/stoploss.md
```diff
@@ -52,17 +52,17 @@ The bot cannot do these every 5 seconds (at each iteration), otherwise it would
 So this parameter will tell the bot how often it should update the stoploss order. The default value is 60 (1 minute).
 This same logic will reapply a stoploss order on the exchange should you cancel it accidentally.
 
-### forceexit
+### force_exit
 
-`forceexit` is an optional value, which defaults to the same value as `exit` and is used when sending a `/forceexit` command from Telegram or from the Rest API.
+`force_exit` is an optional value, which defaults to the same value as `exit` and is used when sending a `/force_exit` command from Telegram or from the Rest API.
```
xmatthias: consistency. For the rpc methods - use the proper notation.

xmatthias (changes requested):
quite honestly - this is not really helping much - but making it more difficult for reviewers, as the requested changes are done inconsistently, wasting quite some time, as the same thing needs to be highlighted multiple times.

Please get used to doing a _review of your own PR_ (this means - go through the diff in github and look for inconsistencies yourself).
Almost all points in the above comments are easily spottable if you'd really look at what you're submitting as PR.

While i did comment on some of the lines where it's wrong - please look through the whole PR to identify other areas where the change was done wrongly (where it's affecting rpc method naming) - and revert that yourself.

SmartManoj:
> as the requested changes are done inconsistently

Sorry, I thought all files mentioned in the requested changes

freqtrade/rpc/telegram.py
```diff
@@ -1359,23 +1359,25 @@ def _help(self, update: Update, context: CallbackContext) -> None:
...
             "_BotControl_\n"
             "------------\n"
             "*/start:* `Starts the trader`\n"
             "*/stop:* Stops the trader\n"
             "*/stopbuy:* `Stops buying, but handles open trades gracefully` \n"
+            # TODO: forceenter forceshort forcelong missing
             "*/forceexit <trade_id>|all:* `Instantly exits the given trade or all trades, "
             "regardless of profit`\n"
-            f"{forceenter_text if self._config.get('forcebuy_enable', False) else ''}"
+            "*/fe <trade_id>|all:* `Alias to /force_exit`"
```
xmatthias:
```suggestion
            "*/fe <trade_id>|all:* `Alias to /forceexit`"
```

docs/configuration.md
```diff
@@ -435,7 +435,7 @@ Configuration:
     If `stoploss_on_exchange` is enabled and the stoploss is cancelled manually on the exchange, then the bot will create a new stoploss order.
 
 !!! Warning "Warning: stoploss_on_exchange failures"
-    If stoploss on exchange creation fails for some reason, then an "emergency exit" is initiated. By default, this will exit the trade using a market order. The order-type for the emergency-exit can be changed by setting the `emergencyexit` value in the `order_types` dictionary - however, this is not advised.
+    If stoploss on exchange creation fails for some reason, then an "emergency exit" is initiated. By default, this will exit t_he trade using a market order. The order-type for the emergency-exit can be changed by setting the `emergency_exit` value in the `order_types` dictionary - however, this is not advised.
```
xmatthias:
```suggestion
    If stoploss on exchange creation fails for some reason, then an "emergency exit" is initiated. By default, this will exit the trade using a market order. The order-type for the emergency-exit can be changed by setting the `emergency_exit` value in the `order_types` dictionary - however, this is not advised.
```

xmatthias (approved):
LGTM :+1: 

Thanks for taking the time to align this.

PR #6658, opened 2022-04-06
renamed enter-side

SmartManoj:

## Summary

renamed enter-side

### DIFF
--- a/freqtrade/persistence/models.py
+++ b/freqtrade/persistence/models.py
@@ -371,7 +371,7 @@ class LocalTrade():
         return self.close_date.replace(tzinfo=timezone.utc)
 
     @property
-    def enter_side(self) -> str:
+    def entry_side(self) -> str:
         if self.is_short:
             return "sell"
         else:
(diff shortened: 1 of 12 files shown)


SmartManoj:
For enter_long and enter_short, long_entry and short_entry would be appropriate. Suggestions?

freqtrade/persistence/models.py
```diff
@@ -371,7 +371,7 @@ def close_date_utc(self):
         return self.close_date.replace(tzinfo=timezone.utc)
 
     @property
-    def enter_side(self) -> str:
+    def entry_side(self) -> str:
```
xmatthias, line 380: Please add a (temporary) compatibility layer here (for strategies using this already with develop).

This means - a property named `enter_side` (which calls `entry_side`). 
It should have the docstring of "DEPRECATED, please use entry_side` - and a TODO with "remove me after 2022.5".

xmatthias (commented):
> For enter_long and enter_short, long_entry and short_entry would be appropriate. Suggestions?

I don't see any benefit from this renamal - therefore please leave that as is.
We've been working on feat/short for almost a year - and this part has been changed roughly half a year ago - enough time for anyone to step in and suggest a naming change.

We'll not change it again at this point.

freqtrade/persistence/models.py
xmatthias: I'm impatient - fixed it myself ... :laughing:

xmatthias (approved):
Nice job :+1: 

PR #6664, opened 2022-04-07
Use 3.10 if found and install 3.10 if python not installed

SmartManoj:
It's better to use 3.10 now.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -25,7 +25,7 @@ function check_installed_python() {
         exit 2
     fi
 
-    for v in 9 10 8
+    for v in 10 9 8
     do
         PYTHON="python3.${v}"
         which $PYTHON
@@ -35,7 +35,7 @@ function check_installed_python() {
             return
         fi
     done
-
+    PYTHON="python3.10"
     echo "No usable python found. Please make sure to have python3.8 or newer installed."
     exit 1
 }


setup.sh
```diff
@@ -25,7 +25,7 @@ function check_installed_python() {
         exit 2
     fi
 
-    for v in 9 10 8
+    for v in 10 9 8
```
xmatthias, line 28: > It's better to use 3.10 now.

Python 3.10 had a memory problem in january - which is why the docker-image has been downgraded to 3.9.
It's also happening on 3.10 (at least on my machine).

There's since been a ton of dependency updates, as well as new 3.10 versions - but i never had the time to verify if the problem is now gone or not.

Therefore, i will only be able to accept this if you are **certain** this problem is gone (and actually did some tests for it) - in which case we could also upgrade the main docker image again.
simply leave the bot running overnight and observe memory usage compared to a 3.9 installation with the same setup - this can be performed either via a 3.10 docker image (we don't build these at the moment) - or via a native installation.

setup.sh
```diff
@@ -35,7 +35,7 @@ function check_installed_python() {
             return
         fi
     done
-
+    PYTHON="python3.10"
```
xmatthias, line 38: Why add this here? it seems pointless - we exit 2 lines after this ...

SmartManoj: If no python versions are installed, it will install python3.8 as it is last processed.

setup.sh
SmartManoj: Seems it may be fixed. https://bugs.python.org/issue46347

setup.sh
xmatthias: Freqtrade will install no python version itself. 2 lines after this, the script will exit if no python version is installed, so this line does nothing.

xmatthias:
> Seems it may be fixed. https://bugs.python.org/issue46347

Thanks - it's now clear that you didn't test it - but are only **suspecting** that it's fixed without actually testing it.

For a project like freqtrade, that's a too risky approach. Should this not be the problem that we were seeing, then we'd return to a memory leak, affecting quite some people.

As such - we will not be accepting this at this point. 
Once it's properly tested, we will surely look at preferring 3.10 over 3.9 - but for now, the circumstances for this are not given yet.

For the future, please make sure to test PR's before posting them.

PR #6668, opened 2022-04-09
Fixed setup.sh

erdieee:
Running ./setup.sh on the commit (ebcb530d4f0303ca0cb164296b92cd8ea269d032) led to the error:
./setup.sh: line 92: syntax error near unexpected token `then'
./setup.sh: line 92: `    if [[ $dev =~ ^[Yy]$ ]] then'

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -89,12 +89,13 @@ function updateenv() {
     fi
     echo "pip install completed"
     echo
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
Great catch! :+1:

PR #6672, opened 2022-04-09
Update .pre-commit-config.yaml

SmartManoj:
## Summary

Previously when there is no change in python files, pre-commit `mypy` runs successfully.
Now tested after files changed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,34 @@ repos:
     rev: 'v0.942'
     hooks:
     -   id: mypy
-        # stages: [push]
+        pass_filenames: false  # suppress the normal filename passing
+        files: ^freqtrade/  # filter the files down to a specific subdirectory
+        # pre-commit only supports running at the root of a repo since that's where
+        # git hooks run.  but it also allows running arbitrary code so you can
+        # step outside of those bounds
+        # note that `bash` will reduce your portability slightly
+        verbose: true
+        entry: mypy freqtrade
+        additional_dependencies: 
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -11,7 +11,34 @@ repos:
...
+        # note that `bash` will reduce your portability slightly
+        verbose: true
+        entry: mypy freqtrade
+        additional_dependencies: 
+        -    types-cachetools==5.0.0
+        -    types-filelock==3.2.5
+        -    types-python-dateutil==2.8.10
+        -    types-requests==2.27.15
+        -    types-tabulate==0.8.6
+        -    types-urllib3==1.26.11
```
xmatthias, line 28: These dependencies will not update automatically via dependabot, which means they'll become outdated within weeks, resulting in wrong results.

Dependency versions shall be read from the requirements file.

If **this** is the requirement to successfully run mypy in pre-commit - then we'll be removing mypy from the pre-commit hook, as this will be a dependency nightmare, without providing visible benefit.

xmatthias (changes requested):

xmatthias:
we will  not pin dependencies in multiple configuration files.
Dependabot will update dependencies in `requirements-*.txt`.
It will however not update dependencies in this file (source: https://github.com/dependabot/dependabot-core/issues/1524).

As pointed out above - the dependencies would become missaligned within weeks - turning this into a dependency nightmare.

Feel free to submit a better approach to this if you think mypy is currently not running - alternatively remove the mypy hook - but this approach we will not take.

PR #6675, opened 2022-04-10
Docs: update plotting doc to show strategy option is mandatory

zolbayars:
## Summary

Providing a strategy name is mandatory to run the plotting, yet the [documentation page](https://www.freqtrade.io/en/stable/plotting/) doesn't say so. 

## Quick changelog

- Add the `--strategy` option to the main example command
- Remove an unnecessary tip

## What's new?

If you run the `plot-dataframe` command without providing a strategy, it fails with `ERROR - No strategy set. Please use --strategy to specify the strategy class to use`. So I think the strategy option should be provided for the main example (and there's no need for that tip about "almost certainly" wanting to specify a strategy) 🙂

### DIFF
--- a/docs/plotting.md
+++ b/docs/plotting.md
@@ -96,7 +96,7 @@ Strategy arguments:
 Example:
 
 ``` bash
-freqtrade plot-dataframe -p BTC/ETH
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
Nice spot :+1:

PR #6689, opened 2022-04-14
Avoid ignoring the case precision amount = 0

kokimame:

## Summary

Avoid skipping rounding with precision amount = 0
Precision amount = 0 seems to be a valid input to decimal_to_precision but 0 is False in boolean and left out at the if.

Solve the issue: #NaN


xmatthias (approved):
That's a great catch - i've improved the test case around this to ensure we'll not be breaking this again in the future

LGTM :+1:

PR #6692, opened 2022-04-16
Feature: Readjust Entry Order

eSeR1805:

## Summary

The goal of this is to allow the user to readjust the limit orders (if not already filled fully/partially) on each subsequent new candle after initial trade trigger.

## Quick changelog

- added new callback to strategy interface
- added new functionality to live/dry bot modes
- added new functionality to backtesting feature
- added documentation entries for new functionality

## What's new?

User can instruct the bot, on each new candle after initial trade trigger, to adjust(cancel/replace) the current limit order so that it reflects the target limit rate. 
For example: expecting a pullback to SMA200(target price) after a breakout of this level(entry trigger). Candles following the initial trigger will naturally have a different value on SMA200. User can decide if they want to pursue their price target higher or lower.

## Current concerns
- no dedicated test(s)
- interaction with position adjust(DCA) in backtest
- naming not quite uniform across the components


freqtrade/freqtradebot.py
```diff
@@ -1140,33 +1141,78 @@ def check_handle_timedout(self) -> None:
...
+        :param trade: Trade object.
+        :return: None
+        """
+        analyzed_df, _ = self.dataprovider.get_analyzed_dataframe(trade.pair,
+                                                                  self.strategy.timeframe)
+        latest_candle_open_date = analyzed_df.iloc[-1]['date'] if len(analyzed_df) > 0 else None
+        latest_candle_close_date = timeframe_to_next_date(self.strategy.timeframe,
+                                                          latest_candle_open_date)
+        # Check if new candle
+        if order_obj and latest_candle_close_date.replace(tzinfo=None) > order_obj.order_date:
```
xmatthias: Please use `order_obj.order_date_utc` for date comparisons. 
That'll avoid having to cast the "other side" throughout the code.

freqtrade/freqtradebot.py
```diff
@@ -1140,33 +1141,78 @@ def check_handle_timedout(self) -> None:
...
+                                                         default_retval=proposed_rate)(
+                pair=trade.pair, current_time=datetime.now(timezone.utc),
+                proposed_rate=proposed_rate, entry_tag=trade.enter_tag,
+                side=trade.entry_side)
+            # check if user has requested entry limit adjustment
+            if proposed_rate != adjusted_entry_price:
+                # cancel existing order
+                self.handle_cancel_enter(trade, order, constants.CANCEL_REASON['REPLACE'],
+                                         allow_full_cancel=False)
+                stake = self.wallets.get_trade_stake_amount(trade.pair, self.edge)
```
xmatthias: i actually don't think this is correct. The code will work - but usage of this is most likely wrong.

It'll work "most of the time" assuming no concurrency and no partially filled orders - but:

* if an additional stake became available - then the "new" entry might have a higher stake than what you had initially (which is potentially unwanted). It'll depend on the configuration - but is tricky to fully test or even think through, as you can have quite some different scenarios (all setting combinations + actual trade scenarios).
* if the order filled partially (say 80/100) - then you're now creating a new order for another 100 units - so your final position is 180 - instead of the 100 you aimed for.

I think the better approach might be to capture the "full" value of the initial order (say you were going for 100 - and filled 80) - then you'd try to place a new order for 20 - so in the end, your "full" stake is filled

freqtrade/strategy/interface.py
```diff
@@ -467,6 +467,28 @@ def adjust_trade_position(self, trade: Trade, current_time: datetime,
         """
         return None
 
+    def readjust_entry_price(self, trade: Trade, pair: str, current_time: datetime,
+                             proposed_rate: float, entry_tag: Optional[str],
+                             side: str, **kwargs) -> float:
```
xmatthias: i somehow don't like the naming - but lack a better name for now.

it somehow looks too much like "read just" ... which won't make any sense.
maybe splitting it (re_adjust_entry_price) makes it better.

That said - i would think that what this method is actually doing is not readjusting the price - but readjusting the order - therefore something like  `adjust_open_order()` might also make sense?

freqtrade/freqtradebot.py
```diff
@@ -1140,33 +1141,78 @@ def check_handle_timedout(self) -> None:
...
+        # Check if new candle
+        if order_obj and latest_candle_close_date.replace(tzinfo=None) > order_obj.order_date:
+            # New candle
+            proposed_rate = self.exchange.get_rate(
+                trade.pair, side='entry', is_short=trade.is_short, refresh=True)
+            adjusted_entry_price = strategy_safe_wrapper(self.strategy.readjust_entry_price,
+                                                         default_retval=proposed_rate)(
+                pair=trade.pair, current_time=datetime.now(timezone.utc),
+                proposed_rate=proposed_rate, entry_tag=trade.enter_tag,
+                side=trade.entry_side)
```
xmatthias: maybe a stupid question - but isn't the most important parameter missing (`order` this would replace)?

Sure, we can get the order through `trade.orders` - but especially if you're in a DCA scenario, you'll need to quite carefully get the "right" order - using `trade.orders[0]` will not do it then...

The doc sample even shows it as well (at least partially) - you're using `trade.open_date_utc` ... but that'll only work for the first attempt - depending on what you're trying to archive - you might replace the order every 5 minutes for 30 minutes - but you'll need to know the offset from the current order - not from the initial order.

docs/strategy-callbacks.md
```diff
@@ -381,14 +382,14 @@ class AwesomeStrategy(IStrategy):
...
+
         return new_exitprice
 
 ```
 
 !!! Warning
-    Modifying entry and exit prices will only work for limit orders. Depending on the price chosen, this can result in a lot of unfilled orders. By default the maximum allowed distance between the current price and the custom price is 2%, this value can be changed in config with the `custom_price_max_distance_ratio` parameter.  
-    **Example**:  
+    Modifying entry and exit prices will only work for limit orders. Depending on the price chosen, this can result in a lot of unfilled orders. By default the maximum allowed distance between the current price and the custom price is 2%, this value can be changed in config with the `custom_price_max_distance_ratio` parameter.
+    **Example**:
```
xmatthias: you'll need to be careful with trailing whitespace in markdown.

At least these 2 are "Meaningful" whitespaces - if there's 2 at the end, it'll be "new paragraph" - so now we're no longer rendering new lines, but all on one line ...

xmatthias (commented):
Just a quick glance for now - overall it looks quite promising :+1: 

all comments are basically suggestions for improvement, pointing out potential problems.

docs/strategy-callbacks.md
eSeR1805: Yep. I'll adjust linter settings and revert these unwanted changes. Thanks!

freqtrade/freqtradebot.py
eSeR1805: Honestly my initial approach was targetting just the first buy of the trade. From this perspective I've named the method to have `entry_price`. It's becoming obvious that this would be useful to be paired with DCA mode as well so I'll have to give that some though.
As for `order` as parameter I'm really not sure if the user would want that object available or not. Again maybe in the context of DCA it makes more sense to have it as well. Since the cancelation and new order are handled internally I feel it doesn't really matter for the implementation of this feature so whatever we choose to share through the strategy interface is fine by me. So I'd say that this should ultimately be your decision and I'm looking forward for you feedback on it.

freqtrade/freqtradebot.py
eSeR1805: I see your point. For this reason I'm considering changing the return of the new strategy callback to have: `float price` + `bool confirmation`. Adding this extra return will remove the need of comparing proposed stake vs new user suggested stake as a means of detecting user intention, thus removing the need of queriying `get_trade_stake_amount`. Added benefit is that it would also allow user to more easily indicate a sort of `timeout` for these readjusted orders.
Looking forward for your feedback on this.

freqtrade/strategy/interface.py
eSeR1805: Initially it was `adjust_entry_price` and I wanted to add emphasis on the `re`occuring nature. `adjust_open_order` is fine for me 😛

freqtrade/freqtradebot.py
eSeR1805: Updated through [17da4ca](https://github.com/freqtrade/freqtrade/pull/6692/commits/17da4ca09939ae56256d7a5d71edf7ff1babbdcd). Thanks!

freqtrade/freqtradebot.py
eSeR1805: Seems I've also confused stake and rate in my previous comment. I agree, replacement order should use the previous attempted(canceled) stake ammout. LE: will be using `order.remaining` as new order stake should there be a partial fill of the previous order.

eSeR1805: Afterthough on `bool confirmation`. Probably not needed, we could use `None` to indicate that we don't want order adjustment or use as pseudo timeout.

freqtrade/strategy/interface.py
eSeR1805: Considering we're taking `entry` out of the name, should this handle both entries and exits?
LE: implementation and usage wise I think it's simpler/better to have them separated: `adjust_entry_price`, `adjust_exit_price`

eSeR1805:
Updated dry/live functionality as per suggestions and personal solution.

Outstanding items:
- ~~issue where trade appears open and order filled after user requested cancellation~~ Fixed through [76558f2](https://github.com/freqtrade/freqtrade/pull/6692/commits/76558f284f15c50264dba74f83bc58ff756d37c8)
- ~~backtesting flow not updated (to consider DCA interaction)~~
- tests

freqtrade/freqtradebot.py
xmatthias: you may be targetting the first order - but there's no limitation in the code - so it'll anyway apply for all orders.

The point of sending the order with it is to have the prior price available (as well as the last adjustment date).
if you assume the same order needs 2-3 adjustments until it fills - then you'll need the time since last order - not since initial order. DCA will also benefit from this - but it's not the main concern.
as such - i'd definitely add the order object as parameter.

eSeR1805: As always, thanks for the feedback!

> you may be targetting the first order - but there's no limitation in the code - so it'll anyway apply for all orders.

Indeed there was no code to apply this limitation (in live/dry modes). Calling it a fortunate mistake. Moving forward I think the desired functionality is to check/replace/cancel all open entry orders (either partially of fully unfilled).

> if you assume the same order needs 2-3 adjustments until it fills - then you'll need the time since last order - not since initial order.

This example was not a usecase I've considered, but I do see how it would add extra flexibility for the end user. Also it would allow end user to check filled amount of the current order and make decisions based on that. Thus,
I've added the Order object to the input param list through: [3166739](https://github.com/freqtrade/freqtrade/pull/6692/commits/3166739ec9a1f211c72f922b8157619d56538148)

freqtrade/strategy/interface.py
```diff
@@ -467,6 +467,32 @@ def adjust_trade_position(self, trade: Trade, current_time: datetime,
...
+
+        :param pair: Pair that's currently analyzed
+        :param trade: Trade object.
+        :param order: Order object
+        :param current_time: datetime object, containing the current datetime
+        :param proposed_rate: Rate, calculated based on pricing settings in exit_pricing.
+        :param entry_tag: Optional entry_tag (buy_tag) if provided with the buy signal.
+        :param side: 'long' or 'short' - indicating the direction of the proposed trade
+        :param **kwargs: Ensure to keep this here so updates to this won't break your strategy.
+        :return float: New entry price value if provided
```
xmatthias, line 497: What would be the return value to "leave the order alone" ? returning proposed_rate (if yes, that should be documented here, in the template, as well as in the doc page ...)

edit: please see other comment first - and then have the result explained here and in the other mentioned places

freqtrade/freqtradebot.py
```diff
@@ -1140,33 +1141,85 @@ def check_handle_timedout(self) -> None:
...
+                trade=trade, order=order_obj, pair=trade.pair,
+                current_time=datetime.now(timezone.utc), proposed_rate=proposed_rate,
+                entry_tag=trade.enter_tag, side=trade.entry_side)
+
+            full_cancel = False
+            cancel_reason = constants.CANCEL_REASON['REPLACE']
+            if not adjusted_entry_price:
+                full_cancel = True
+                cancel_reason = constants.CANCEL_REASON['USER_CANCEL']
+            if proposed_rate != adjusted_entry_price:
```
xmatthias: please rethink the interface a bit (or explain it better to me)

it's currently unclear to me how to get the bot to do the following actions:
1. leave order as is
2. cancel order from this callback (i assume that's None?)
3. set new order at proposed rate
4. set the new order at any random given rate - say 225$

(the above should serve as a thinking exercise - on what states we'd want to get to).

with the current implementation / interpretation of the return values - you can't get to all 4 states.
1 will only work if you return "proposed_rate" (hence this comment is on this line) - but that means you can't use the "new best rate" as entry price (point 3) - but instead must slightly modify it by 0.0000001 so it's no longer the same price.

freqtrade/optimize/backtesting.py
```diff
@@ -850,6 +859,21 @@ def run_protections(self, enable_protections, pair: str, current_time: datetime)
...
+        """
+        for order in [o for o in trade.orders if o.ft_is_open]:
+            if order.side == trade.entry_side and current_time > order.order_date_utc:
+                # cancel existing order
+                del trade.orders[trade.orders.index(order)]
+
+                # place new order
+                self._enter_trade(pair=trade.pair, row=row, trade=trade,
+                                  direction='short' if trade.is_short else 'long',
+                                  readjust_req=True)
```
xmatthias: i don't think this is correct.

you're calling this method when timeout did not happen, so you can't automatically assume to delete the order - what if the callback returned to "leave the same order in place"?.

I'll have to think about the approach for backtesting a bit - but i don't think that calling the callback from within `_enter_trade()` is a good approach here.

a (untested) suggestion):
try to call the callback here - if the response is to replace the order - delete it and call `_enter_trade()` with both the trade object, and the price (instead of the new `readjust_req` parameter).

freqtrade/strategy/interface.py
xmatthias: having separate callbacks is fine (and you don't need to have them both at this point, either - as i think for exits, people will usually prefer to exit quickly).

freqtrade/freqtradebot.py
xmatthias: i'm not sure this is final yet (see other comment) ...

freqtrade/freqtradebot.py
eSeR1805: **On current implementation:**
> 1. leave order as is

 Indeed user would have to return `proposed_rate`

> 2. cancel order from this callback (i assume that's None?)

Yup. User would have to return `None` to just cancel the order (with no replacement)

> 3. set new order at proposed rate

Can't do. As you've noticed it would need a slight modification to `proposed_rate`. Used to think that it's not a "big deal" but if I consider BTC pairs it might be a significant difference.

> 4. set the new order at any random given rate - say 225$

User simply returns the desired rate for the new order. If I'm missing something here regarding fees or such, I'm sorry, I'm probably oblivious to such things at the moment.

**Proposed solutions:**
1. Simple hackish (probably confusing a bit for end user?):
To leave order as is user would need to `return -proposed_stake`
To use the "new best rate" user would need to `return proposed_stake`

2. Add an extra `bool` return to better specify intent:
> 1. leave order as is

`return proposed_stake, False`
> 2. cancel order from this callback (i assume that's None?)

`return proposed_stake, None`
> 3. set new order at proposed rate

`return proposed_stake, True`
> 4. set the new order at any random given rate - say 225$

`return 225, True`

If there's a simpler/better way that I've missed while replying to this please let me know. As always thank you for the feedback and looking forward for your thoughts on my proposals.

freqtrade/strategy/interface.py
eSeR1805: Understood. Will do another round of docs and info updates once we settle for a solution to best support all 4 cases from the other comment.

freqtrade/freqtradebot.py
xmatthias: > User simply returns the desired rate for the new order. If I'm missing something here regarding fees or such, I'm sorry, I'm probably oblivious to such things at the moment.

fees don't matter here - we're talking about the order price - not profits.
my point in this was - if that new desired rate is (accidentally) the new proposed rate, it would also not work.

---

i think both solutions CAN confuse the user, the first due to random negative return values - the 2nd by using tuples (tuples in python are NOT that easy to grasp - and sometimes error-prone if you're not careful).

i think the big problem comes from supporting cancellation out of this callback. 
if you take this out - then `None` leaves the order in place - and every other return value sets a new rate.

what's the main thought behind supporting cancellation from here as well? it'll complicate both freqtrade's code (you now have 2 callbacks which can do the same thing) - and potentially confuse users (which one should i use??).

what benefit would it provide over the `custom_entry_timeout()` callback?

eSeR1805: > what benefit would it provide over the `custom_entry_timeout()` callback?

Nothing appart from slight convenience. I imagine whomever wants to use this with multiple branches and different timeouts for each would have some repeated logic both in `custom_entry_timeout()` and `adjust_entry_price()`. Not that big of a deal probably.

But yeah, I see your point. Better to have redundant code in strat (with low probability of occuring anyway) than have it in freqtrade code.

xmatthias: no one said defining such a thing is easy :laughing: 

the only thing i see as REALLY problematic at this point is to leave the order in place with `proposed_rate`. we can leave the order in place if the order price is returned (maybe this is the point we're looking for?)

Thinking about this a bit more - the current code would anyway replace the order in the wrong instance.
assume an open order at 99$ (it's not filled).
price now moved to 100$ (so the proposed rate is 100$)
if the strategy now return 99$ again - we'd currently cancel and replace the order - which is clearly unnecessary.
if instead, i return 100$ (to move the order higher) - then we'd NOT replace the order - which seems wrong (i'd have to send 100.0001$ to have it replaced).

so maybe the correct approach would be to pass in a "current_order_rate" - which acts as "leave order in place". Then i think we would be able to cover all of the above cases.

eSeR1805: Oh this sounds great!

freqtrade/freqtradebot.py
```diff
@@ -1215,7 +1273,8 @@ def handle_cancel_enter(self, trade: Trade, order: Dict, reason: str) -> bool:
         if isclose(filled_amount, 0.0, abs_tol=constants.MATH_CLOSE_PREC):
             logger.info(f'{side} order fully cancelled. Removing {trade} from database.')
             # if trade is not partially completed and it's the only order, just delete the trade
-            if len(trade.orders) <= 1:
+            open_order_count = len([order for order in trade.orders if order.status == 'open'])
+            if open_order_count <= 1 and allow_full_cancel:
```
xmatthias, line 1280: I'm not sure this is correct in a DCA scenario (an order would be closed but filled, so you can't delete the trade).
You might wanna check this via `trade.select_filled_orders()` (or `trade.nr_of_successful_entries`).

eSeR1805: Updated through https://github.com/freqtrade/freqtrade/pull/6692/commits/f9977c26e7f682346f875f30fd8b11a9f19cc867
Thanks!

freqtrade/optimize/backtesting.py
eSeR1805: Took another swing at backtesting end https://github.com/freqtrade/freqtrade/pull/6692/commits/ad0c5d944034228f8e17ba658b3b3e1d8ff4af72. Not sure if I've considered all cases but it should be better now. I guess DCA is still main concern to get right.

freqtrade/freqtradebot.py
eSeR1805: Addressed this through https://github.com/freqtrade/freqtrade/pull/6692/commits/17650d7e6025385f8d9dbd3dbe44d097b50ce9eb.

freqtrade/strategy/interface.py
eSeR1805: Docs also updated through https://github.com/freqtrade/freqtrade/pull/6692/commits/17650d7e6025385f8d9dbd3dbe44d097b50ce9eb.

freqtrade/freqtradebot.py
xmatthias: i can see the change in the commit - but not in the "uptodate branch" - so i guess there's a problem with the develop merge, which reverted this change (?)

freqtrade/optimize/backtesting.py
```diff
@@ -661,7 +661,8 @@ def get_valid_price_and_stake(
                 proposed_rate=propose_rate, entry_tag=entry_tag,
                 side=direction,
             )  # default value is the open rate
-            # We can't place orders higher than current high (otherwise it'd be a stop limit entry)
+
+            # We can't place orders higher than current high (otherwise it'd be a stop limit buy)
```
xmatthias: i guess that's an unintended change?

freqtrade/optimize/backtesting.py
```diff
@@ -960,13 +998,16 @@ def backtest(self, processed: Dict,
 
                 for t in list(open_trades[pair]):
                     # 1. Cancel expired entry/exit orders.
-                    if self.check_order_cancel(t, current_time):
-                        # Close trade due to entry timeout expiration.
+                    order_cancel = self.check_order_cancel(t, current_time)
+                    # 2. Replace/cancel (user requested) entry orders.
+                    order_replace = self.check_order_replace(t, current_time, row)
+                    if order_cancel or order_replace:
```
xmatthias: Behavior in live (freqtradebot.py) and backtesting (here) will be slightly different.

Live will only call the new callback if the order was NOT cancelled. Here, you always call both callbacks. 

I also think that this condition can lead to odd circumstances 
e.g. check_cancel cancels the order - and order_replace replaces it.
As `order_cancel` will then still be True - the trade will be deleted - so the whole "replace" logic was for nothing.


You can probably get arount this by combining the two into the `if` statement - if the first evaluates to true, the 2nd will not be executed (eager execution)

freqtrade/freqtradebot.py
eSeR1805: Erm... I see it in https://github.com/freqtrade/freqtrade/pull/6692/files#diff-a5ba1096e60674c3beaeff004ded33639f377283eed9fcaaeeb20b518f038473R1199
So in `replace_order()` before the call to `handle_cancel_enter()`. Is it not where you expected this to be?

freqtrade/optimize/backtesting.py
eSeR1805: Yeah. Looks like my messup from when merging.

freqtrade/optimize/backtesting.py
eSeR1805: Yeah. I've picked up on it also but thougth it to be harmless. My thinking was this:
1. if `check_order_cancel()` deletes the order from the trade it will no longer execute the callback from `check_order_replace()` since the loop will not find any open orders
2. if `check_order_cancel()` decides to delete the trade alltogether then it doesn't really matter what `check_order_replace()` does (and it will not do much because of 1.)
3. my main priority here was to have timeout check execute before replacement logic

freqtrade/optimize/backtesting.py
eSeR1805: Addressed through https://github.com/freqtrade/freqtrade/pull/6692/commits/9d205132d0fb54df99a7f2b5d0a59520e07eab1b

freqtrade/freqtradebot.py
xmatthias: actually no, i didn't.
i guess it'll work this way, too.

I was expecting it to be there in place of `open_order_count` (right above here) if i'm honest - as checking for the amount of orders with status=open is pretty pointless, as there can only be one open order.

i'd have moved that condition "here" - and left "allow_full_cancel" as simple boolean (that way it's clear also from within the function if we're replacing the order, or cancelling it for good).

freqtrade/optimize/backtesting.py
xmatthias: ah right - that might work - but i always find such implicit logic (the 2nd won't do anything because there's no order left) dangerous ...

eSeR1805: Addressed this through: https://github.com/freqtrade/freqtrade/pull/6692/commits/4e43194dfeefb135e3a567659dba605bb501b10d.

tests/test_freqtradebot.py
```diff
@@ -2450,15 +2450,117 @@ def test_check_handle_timedout_entry(
...
+    cancel_order_mock = MagicMock(return_value=limit_buy_cancel)
+    patch_exchange(mocker)
+    mocker.patch.multiple(
+        'freqtrade.exchange.Exchange',
+        fetch_ticker=ticker_usdt,
+        fetch_order=MagicMock(return_value=old_order),
+        cancel_order_with_result=cancel_order_mock,
+        get_fee=fee
+    )
+    freqtrade = FreqtradeBot(default_conf_usdt)
```
xmatthias: you might wanna use `get_patched_freqtradebot()` instead - it'll get exchange/rpc mocked and therefore out of the way.

freqtrade/optimize/backtesting.py
```diff
@@ -867,28 +873,76 @@ def run_protections(
...
+            # cancel existing order whenever a new rate is requested (or None)
+            if requested_rate == order.price:
+                # assumption: there can't be multiple open entry orders at any given time
+                return False
+            else:
+                del trade.orders[trade.orders.index(order)]
+
+            # place new order if None was not returned
+            if requested_rate:
+                self._enter_trade(pair=trade.pair, row=row, trade=trade,
```
xmatthias, line 939: I'm not entirely sure this call is correct (not sure it's wrong either - but the outcome is wrong)

We expect to update an existing trade - but the open rate is not updated (assuming it's the first trade).
I suspect the real problem is in `recalc_trade_from_orders()` ... (which we'd expect to update the rate i guess) - but especially in the "1 order" scenario, it's not doing that at the moment (which will probably need to change).

The test i'v added (fails ATM) shows (if you debug it) that the `open_rate` is left at the initial rate - so there is clearly something fishy going on here. 
In reality, this test will not have 10% profit .

eSeR1805: Proposed solution for this added through https://github.com/freqtrade/freqtrade/pull/6692/commits/dbecc097dffdb54b8b78c10ce52c99283c00d4b0

eSeR1805: Personally I'm not particularly happy with needing this:
`latest_order_in_trade.id is not None):`
I'm assuming is irrelevant for normal operation but for testcases it seems that orders being injected, they might not have an `id` populated.

freqtrade/persistence/models.py
```diff
@@ -866,9 +866,17 @@ def calc_profit_ratio(self, rate: Optional[float] = None,
         return float(f"{profit_ratio:.8f}")
 
     def recalc_trade_from_orders(self):
+        filled_orders_count = len(self.select_filled_orders(self.entry_side))
+        latest_order_in_trade = self.select_order(self.entry_side, True)
+        # No fills but newer order
+        if (filled_orders_count == 0 and latest_order_in_trade is not None and
+                latest_order_in_trade.id is not None):
+            # after ensuring there is a populated order
+            if latest_order_in_trade.id > 1:
```
xmatthias: Whatever you do here (or why) - it's wrong.

`.id` is an irrelevant element of the order (it's the sequence of the table) - and it does NOT correspond to the order ID for this trade - but simply the sequence of all orders (so a trade could have order.id 200 - and it's still the only order for this trade).

It'll be 1 for the very first trade (and in tests that only provide one order) - but nothing else.

what you need to check here instead is that `.price` is populated (that can be empty)

eSeR1805: Thanks for the feedback. Updated once more through: https://github.com/freqtrade/freqtrade/pull/6692/commits/ae01afdd0f3bd9f63b3b39d8bb22ae8a3a38830b
Hope I've understood your intent correctly.

freqtrade/optimize/backtesting.py
xmatthias: I fear i did misslead you (or rather - didn't analyze properly before posting the above) - sorry for that.
`open_rate` is properly set in `update_trade()` - which covers the live scenario.

This method is however not called in backtesting - so the setting of open rate should be moved to backtesting - where the (first) order fills.

eSeR1805: Yes somehow I've missed the failing freqtradebot tests, or ommitted to run them. Would have been maybe obvious to me also that it would have negative impact in live/dry side. Reverted model changes through: https://github.com/freqtrade/freqtrade/pull/6692/commits/25c74e26d1a59699e8ae5e54f94d5fe0cb47a3d4
Another swing at this added through: https://github.com/freqtrade/freqtrade/pull/6692/commits/2bed0eab0cc5433800ac8de8b68bb0ad2d7ca1f3

xmatthias:
I did a very quick test today - with a "mean" strategy (place order way off spot - and then adjust up every minute).

This seems to be working nicely for the most part (see orders below).

image

Some things to consider (how this will behave):
* Trade_open_date will still remain at the initial open date (when the first order was placed).
* Open-rate is currently NOT adjusted until the first order fills - leading to the below scenario (that's not necessarily a problem - just something we should be aware is happening). Unrelated to this PR, i think behavior here should change (don't show profit until the first order fills) - but that's clearly out of scope for this PR.

image

eSeR1805:
> Trade_open_date will still remain at the initial open date (when the first order was placed).

Well this kinda makes sense for me but I see why it would be confusing. Thanks for adding mention in the docs for it!
> Open-rate is currently NOT adjusted until the first order fills - leading to the below scenario (that's not necessarily a problem - just something we should be aware is happening). Unrelated to this PR, i think behavior here should change (don't show profit until the first order fills) - but that's clearly out of scope for this PR.

Indeed that's not great and telegram reporting is even more confusing if I recall correctly. Fixing this sounds pretty straightforward so I'll have a look at it and start a new PR once I get a bit of time.

xmatthias (approved):
Great improvement :100: 

LGTM :+1:

PR #6694, opened 2022-04-17
switch notionalFloor -> minNotional and notionalCap -> maxNotional

samgermain:
This will be needed once this PR gets merged in https://github.com/ccxt/ccxt/pull/12873

### DIFF
--- a/freqtrade/exchange/binance_leverage_tiers.json
+++ b/freqtrade/exchange/binance_leverage_tiers.json
@@ -2,90 +2,90 @@
   "RAY/USDT": [
     {
       "tier": 1,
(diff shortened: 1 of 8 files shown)


xmatthias (changes requested):
As this is a breaking change dependant on the ccxt version:
Once the PR is merged, please push another commit which bumps the min-required ccxt version to the version that contains this change (in `setup.py`) - and obviously the requirements.txt version, too.

All the rest is fine (in the end, we'll need to adjust to the way ccxt provides this info).

xmatthias (approved):
LGTM :+1: 

also replaces #6701

PR #6704, opened 2022-04-20
Added renovate bot

SmartManoj:

## Summary

Added renovate bot for auto updating pre-commit.

Reference : https://github.com/dependabot/dependabot-core/issues/2040#issuecomment-954658300

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,49 @@ repos:
     rev: 'v0.942'
     hooks:
     -   id: mypy
-        # stages: [push]
+        pass_filenames: false  # suppress the normal filename passing
+        files: ^freqtrade/  # filter the files down to a specific subdirectory
+        # pre-commit only supports running at the root of a repo since that's where
+        # git hooks run.  but it also allows running arbitrary code so you can
+        # step outside of those bounds
+        # note that `bash` will reduce your portability slightly
+        verbose: true
+        entry: mypy freqtrade
+        additional_dependencies: 
+        # Splitting dashes and items on separate lines is intentional,
+        # enables Renovate's pip_requirements manager to update this file.
+        -
+            types-cachetools==5.0.0
+        -
+            types-filelock==3.2.5
+        -
+            types-python-dateutil==2.8.10
+        -
+            types-requests==2.27.15
+        -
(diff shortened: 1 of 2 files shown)


xmatthias (changes requested):
We are using Dependabot for dependency management - and i don't see why we should be adding another dependency managing service.

mypy will work just fine as is for all internal modules/packages- and will only have problems for external  calls - where a manual run of mypy might be necessary.
It's however not something that justifies adding another dependency management (which may eventually cause other conflicts, missaligned dependencies, ...).

If any - then we can use a [standalone github action](https://github.com/marketplace/actions/pre-commit-autoupdate) for this, which creates PR's [which has been linked in the same thread you link above](https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/).

---

I'm still doubting the usefulness of mypy in the pre-commit context in general.
In my understanding, it'll ONLY run on files that are staged (this makes it fast).

This means, if i change `exchange.py` - it'll run `mypy exchange.py`. 
This implicitly means, it'll not check all the other files which actually rely on the return types of this file (for example `freqtradebot.py`),  therefore rendering it useless.

If on the other hand, we force it to run `mypy freqtrade` (What you're proposing here)- then we'll have the performance penalty of running mypy on the whole project (independent of the type libraries) - which will take longer (11s in a recent test) - which is more than what i deem acceptable - therefore I'll surely not accept any thing that will delay `git commit` by 11 seconds due to some pre-commit hook.

xmatthias:
As the comment you linked to clearly states - it's a creative approach to **trick renovate into doing this** - not a supported way from renovate. 

We'll not use a non-supported approach like this, which could break at any time (most likely without us noticing anytime soon, as the updates would simply no longer be applied).

SmartManoj:
> If any - then we can use a [standalone github action](https://github.com/marketplace/actions/pre-commit-autoupdate) for this, which creates PR's [which has been linked in the same thread you link above](https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/).

https://github.com/dependabot/dependabot-core/issues/2040#issuecomment-954656421

> In my understanding, it'll ONLY run on files that are staged (this makes it fast).

For mypy, the whole project will run.
And the stage param is push, it will run only when pushing.

PR #6707, opened 2022-04-21
Ref: timeseries friendly merge_ordered in merge_informative_pair func…

koradiyakaushal:
time-series friendly pd.merge_ordered instead of pd.merge in merge_informative_pair function to include merge and ffill in same method and 2.5xfaster

https://pandas.pydata.org/docs/user_guide/merging.html#timeseries-friendly-merging


xmatthias (approved):
That's a great find! thanks a lot :100: 

LGTM :+1:

PR #6715, opened 2022-04-24
Add relative drawdown

### LINKED ISSUE #6655: Account Drawdown is not calculated correctly if drawdown is at start of simulation
## Describe your environment

  * Operating system: Windows 10
  * Python Version:  Python 3.8
  * CCXT version: 1.57.38
  * Freqtrade Version: develop-973107c
 
## Describe the problem:

When a strategy start on loss, that initial loss doesn't seem to be counted for the account drawdown calculation
Simulation conditions: 
Initial Balance = 1000 USDT 
Minimal Balance = 496.588 USDT (50% drawdown)
Accused Drawdown = 9%

```
 =============== SUMMARY METRICS ================
| Metric                 | Value               |
|------------------------+---------------------|
| Backtesting from       | 2019-01-01 00:00:00 |
| Backtesting to         | 2021-09-30 23:55:00 |
| Max open trades        | 1                   |
|                        |                     |
| Total/Daily Avg Trades | 617 / 0.62          |
| Starting balance       | 1000 USDT           |
| Final balance          | 12034.484 USDT      |
| Absolute profit        | 11034.484 USDT      |
| Total profit %         | 1103.45%            |
| Trades per day         | 0.62                |
| Avg. daily profit %    | 1.10%               |
| Avg. stake amount      | 1775.784 USDT       |
| Total trade volume     | 1095658.453 USDT    |
|                        |                     |
| Best Pair              | BTC/USDT 288.07%    |
| Worst Pair             | BTC/USDT 288.07%    |
| Best trade             | BTC/USDT 31.93%     |
| Worst trade            | BTC/USDT -13.74%    |
| Best day               | 1354.786 USDT       |
| Worst day              | -583.621 USDT       |
| Days win/draw/lose     | 194 / 574 / 236     |
| Avg. Duration Winners  | 18:34:00            |
| Avg. Duration Loser    | 8:49:00             |
| Rejected Entry signals | 0                   |
| Entry/Exit Timeouts    | 0 / 0               |
|                        |                     |
| Min balance            | 496.588 USDT        |
| Max balance            | 12034.484 USDT      |
| Drawdown (Account)     | 9.13%               |
| Drawdown               | 656.284 USDT        |
| Drawdown high          | 6191.971 USDT       |
| Drawdown low           | 5535.688 USDT       |
| Drawdown Start         | 2021-07-07 11:25:00 |
| Drawdown End           | 2021-07-17 04:10:00 |
| Market change          | 1065.27%            |
================================================
```

Attaching the profit plot for better visualization:

image

### Steps to reproduce:

  1. Have your strategy loose a lot of trades at first
  2. Pickup profit
  3. Observe drawdown at the end of the simulation
  
### Observed Results:

  * What happened? Drawdown accused was 9%, while my account was 50% under at some point
  * What did you expect to happen? My conception for drawdown is to be the loss you would initially have if you enter at the worst possible time. That, of course, could also happen with your initial stake before you start having any profits.

nicolaspapp:
## Summary

Add a relative drawdown metric in backtesting an profit plotting.
Solve the issue: #6655

## Quick changelog

- Modified `_calc_drawdown_series` to add a column for relative drawdown and included starting balance in the calculation when present
- Added strating balance in parameters on `calculate_underwater` and `calculate_max_drawdown` 
- Added `MaxDrawDownRelativeHyperOptLoss` Hyperopt loss function
- Added `Max % of account underwater` into backtest stats 
- Added a new subplot into `add_underwater` but with relative drawdown that will be shown in proft plots

## What's new?

The concept of absolute drawdown added recently is very useful for simulations and optimizations, but when doing long simulations and using dynamic stake, absolute drawdown might masquerade big drawdowns at the start of the simulation.
Showing relative drawdown as another metric,  helps identifying those cases.

Small example:

image

This image shows an account dradown of 9% based on highest absolute dradown, but hides an account drawdown of about 37% earlier in the simulation.


freqtrade/optimize/optimize_reports.py
```diff
@@ -760,11 +764,12 @@ def text_table_add_metrics(strat_results: Dict) -> str:
                                              strat_results['stake_currency'])),
 
             # Compatibility to show old hyperopt results
-            ('Drawdown (Account)', f"{strat_results['max_drawdown_account']:.2%}")
+            ('Max % of account underwater', f"{strat_results['max_relative_drawdown']:.2%}"),
+            ('Absolute Drawdown (Account)', f"{strat_results['max_drawdown_account']:.2%}")
```
xmatthias: If we change backtesting output (not against that) - we should also document it in `backtesting.md` - explaining what the new metrics contain (this is especially important in this case - as we now have X drawdown metrics which all show slightly different things.

docs/hyperopt.md
```diff
@@ -563,7 +565,8 @@ Currently, the following loss functions are builtin:
 * `SharpeHyperOptLossDaily` - optimizes Sharpe Ratio calculated on **daily** trade returns relative to standard deviation.
 * `SortinoHyperOptLoss` - optimizes Sortino Ratio calculated on trade returns relative to **downside** standard deviation.
 * `SortinoHyperOptLossDaily` - optimizes Sortino Ratio calculated on **daily** trade returns relative to **downside** standard deviation.
-* `MaxDrawDownHyperOptLoss` - Optimizes Maximum drawdown.
+* `MaxDrawDownHyperOptLoss` - Optimizes Maximum absolute drawdown.
+* `MaxDrawDownRelativeHyperOptLoss` - Similar as the above, but also optimizes Maximum relative drawdown.
```
xmatthias: I think this may need some more explanation on how this relative drawdown is calculated.

just from the term it's unclear - and could be 2-3 different things

freqtrade/plot/plotting.py
```diff
@@ -158,12 +159,15 @@ def add_profit(fig, row, data: pd.DataFrame, column: str, name: str) -> make_sub
 
 
 def add_max_drawdown(fig, row, trades: pd.DataFrame, df_comb: pd.DataFrame,
-                     timeframe: str) -> make_subplots:
+                     timeframe: str, starting_balance: number) -> make_subplots:
```
xmatthias: I think the proper typehint for this (and the other functions) is float, not number (this will also avoid importing Numpy just for typehints).

tests/data/test_btanalysis.py
```diff
@@ -362,3 +362,38 @@ def test_calculate_max_drawdown2():
     df = DataFrame(zip(values[:5], dates[:5]), columns=['profit', 'open_date'])
     with pytest.raises(ValueError, match='No losing trade, therefore no drawdown.'):
         calculate_max_drawdown(df, date_col='open_date', value_col='profit')
+
+
+@pytest.mark.parametrize('profits,relative,highd,lowd,result,result_rel', [
+    ([0.0, -500.0, 500.0, 10000.0, -1000.0], False, 3, 4, 1000.0, 0.090909),
+    ([0.0, -500.0, 500.0, 10000.0, -1000.0], True, 0, 1, 1000.0, 0.5),
```
xmatthias: I've added a test case for the 2 scenarios we are looking at - with the (very simple) test numbers, and the result looks quite promising.

The thing that surprised me here is that absolute drawdown remains 1000 - even though "relative" is used.
In my understanding, all returned values should align - so if i use `relative` - i get "high/low" from the initial drawdown (the biggest one relative wise, relative to balance) - including dates, and absolute drawdown also from this period.

currently, we get the 500$ drawdown (for all values _except_ the absolute drawdown - which is quite odd - as all other values are from the initial drawdown period.

xmatthias (commented):
I've added 2 commits adding an explicit test-case for this. It will simplify testing this as it detaches it from strategy logic and data and whatnot - and purely focuses on the calculation at hand.

I've not looked at everything in detail yet (especially not the plotting part) - but i think we first need to get the calculation the way we desire it to be.

tests/data/test_btanalysis.py
nicolaspapp: Yes you are right, absolute drawdown result was not matching the relative result when the `relative` flag was being used, I just added a commit to fix that

docs/hyperopt.md
nicolaspapp: Updated.

xmatthias (approved):
Great :+1: 

Thanks a lot for this! :+1:

PR #6726, opened 2022-04-25
Move df append to pd concat

froggleston:
## Summary

Small update to remove deprecation warnings - replace df.append function with pd.concat in backtesting.py


xmatthias (approved):
fun thing: it'll only fix the warning - performance will not be improved.

LGTM :+1:

froggleston:
Yup, this is just a warning removal PR ;)

PR #6728, opened 2022-04-26
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -324,6 +324,8 @@ jobs:
   notify-complete:
     needs: [ build_linux, build_macos, build_windows, docs_check, mypy_version_check ]
     runs-on: ubuntu-20.04
+    permissions:
+      repository-projects: read  
     steps:
 
     - name: Check user permission
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,6 +309,9 @@ jobs:
           webhookUrl: ${{ secrets.DISCORD_WEBHOOK }}
 
   cleanup-prior-runs:
+    permissions:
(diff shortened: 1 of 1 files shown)


xmatthias:
This definitely makes sense - i think we'll also need to add it to the next job in the workflow (although i'm not sure which permission that'll need).

I do have a few questions about this, as i find the documentation quite confusing.

The way it's configured now - would that mean that all other permissions are set to None (not defined = restricted) - or are we just explicitly granting these permissions - while all other permissions would also be there?

naveensrinivasan:
> The way it's configured now - would that mean that all other permissions are set to None (not defined = restricted) - or are we just explicitly granting these permissions - while all other permissions would also be there?

The ones within the Job are restricted to those permissions, and other permissions within those jobs are set to none. 

>If you specify the access for any of these scopes, all of those that are not specified are set to none.

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

HTH

xmatthias (approved):
Great - thanks for your contribution and for the clarification! :+1:

PR #6729, opened 2022-04-26
Add 'exit_tag' parameter to 'custom_exit_price' callback.

rokups:
`custom_entry_price` already had `entry_tag` so i added `exit_tag` to `custom_exit_price`. This allows having different exit prices based on exit condition in `custom_sell`.


xmatthias (approved):
Nice :+1:

PR #6747, opened 2022-04-29
Temp fix till auto updating mypy additional dependencies

SmartManoj:
## Summary

As discussed here, 
https://github.com/freqtrade/freqtrade/pull/6704#issuecomment-1107439849, let the devs use their own pre-commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,6 +98,9 @@ target/
 .pytest_cache/
 .mypy_cache/
 
+# temp
+.pre-commit-config.yaml
+
 #exceptions
 !*.gitkeep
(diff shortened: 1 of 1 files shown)


xmatthias:
We will not .gitignore a file that's commited to the repository.

You should also pay attention if your contributions actually make sense, and if things did change since you last looked at things.
Mypy is now included (with dependencies) in the pre-commit file - including a ci step that'll ensure we're keeping dependencies uptodate.

Posting redundant / pointless PR's (like this one) are not helping this project, but are actually wasting people's time.
Thanks.

PR #6748, opened 2022-04-29
Update PULL_REQUEST_TEMPLATE.md

SmartManoj:
# Summary 
Added instructions as comments


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,17 @@
...
+<!-- Explain in one sentence the goal of this PR -->
 
 Solve the issue: #___
 
 ## Quick changelog
 
-- <change log 1>
-- <change log 1>
+- <!-- <change log 1> -->
+- <!-- <change log 1> -->
```
xmatthias: These 2 should not be comments - they're intended to be replaced.

xmatthias (changes requested):

xmatthias (approved):
Let's give this a try - i doubt it'll change much. 
There'll still be some people who care, and properly specify what they'd like to change
and others, which simply dump the PR without any description.

PR #6771, opened 2022-05-02
Update setup.sh

talentoscope:
Added curl to dependencies for Debian systems

-- Just a small change that corrects the setup.sh file, as curl is required but not installed by the bash setup script by default


xmatthias (approved):
LGTM :+1:

PR #6773, opened 2022-05-03
Freqai: an interface for users to build/train/backtest predictive models

robcaulk:
## Summary

Add an interface for users to easily create and interface their own predictive models with Freqtrade strategies. 

Solve the issue: #___

## Quick changelog

Add a freqai interface with datahandling tools.

## What's new?

- Build data handling tools
- Provide examples
- Furnish data analysis tools
- Add doc


freqtrade/optimize/backtesting.py
```diff
@@ -204,6 +204,12 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
         """
         self.progress.init_step(BacktestState.DATALOAD, 1)
 
+        if self.config['freqaimodel']:
+            self.required_startup += int((self.config['freqai']['train_period']*86400) /
```
xmatthias, line 208: if you assume certain attributes to be in the config, they must be properly specified in constants so schema validation takes place.

otherwise, all options can be optional, and you can't trust anything to be there.
(i'd go the validated route though - otherwise it can be tough)

xmatthias:
Closing this in favor of #6774.

PR #6774, opened 2022-05-03
Freqai: an interface for users to build/train/backtest predictive models

robcaulk:
## Summary

SUPERCEDED by #6832 

Add an interface for users to easily create and interface their own predictive models with Freqtrade strategies

Solve the issue: #___

## Quick changelog

Add a freqai interface with datahandling tools.

## What's new?

-Build data handling tools
-Provide examples
-Furnish data analysis tools
-Add doc

TODO:

- [x] add automatic retraining during live (which may need to use data-download if training period extends beyond exchange candle limits).
- [x] decide on location of freqai parameters and validate them
- [x] finish doc
- [ ] add ability to accept open ended timerange
- [ ] use .get throughout datakitchen.py/freqai_interface.py for any optional config parameters 
- [x] let user run config from anywhere on the system
- [ ] add unit tests


freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,155 @@
+
+import os
```
xmatthias: I would try to not use `os.path` (and consorts) whenever possible, but instead rely on [pathlib.Path](https://docs.python.org/3/library/pathlib.html).
That obviously applies not just to this file but overall - it'll also allow you to pass around paths as Path object instead of strings - which is usually simpler / less errorprone once you're used to it.

We're doing that everywhere else already - and os.path is pretty tricky to get right for both windows and linux/mac.

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(self, config: Dict[str, Any]) -> None:
+
+        self.config = config
+        self.freqai_info = config['freqai']
+        self.data_split_parameters = config['freqai']['data_split_parameters']
+        self.model_training_parameters = config['freqai']['model_training_parameters']
+        self.feature_parameters = config['freqai']['feature_parameters']
+        self.full_path = (str(config['user_data_dir'])+
+                            "/models/"+self.freqai_info['full_timerange']+
+                            '-'+self.freqai_info['identifier'])
```
xmatthias: would fail on windows - please use Pathlib (as pointed out above).

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,155 @@
...
+        :dataframe: Full dataframe coming from strategy - it contains entire
+        backtesting timerange + additional historical data necessary to train 
+        the model.
+        :metadata: pair metadataa coming from strategy. 
+        """
+        self.pair = metadata['pair']
+        self.dh = DataHandler(self.config, dataframe, self.data)
+
+        print('going to train',len(self.dh.training_timeranges),
+            'timeranges:',self.dh.training_timeranges)
```
xmatthias: please avoid using print statements except for "user summary" output (like the backtesting tables).

If it's worthy output (which should be shown to the user) - please use `logger.info()` (or some other appropriate loglevel).
If it's just internal debug output while testing - i'd remove it completely (or comment it at least).

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,155 @@
...
+    def make_labels(self, dataframe: DataFrame) -> DataFrame:
+        """
+        User defines the labels here (target values).
+        :params:
+        :dataframe: the full dataframe for the present training period
+        """
+
+        return dataframe
+
+    def train(self, unfiltered_dataframe: DataFrame, metadata: dict) -> Any:
```
xmatthias, line 153: You should make this an @abstractmethod
that will force the inheriting class to override this - and i don't think calling using Any as model will work down the strategy.

freqtrade/freqai/data_handler.py
```diff
@@ -0,0 +1,439 @@
...
+from sklearn.model_selection import train_test_split
+from sklearn.metrics.pairwise import pairwise_distances
+import datetime
+from typing import Any, Dict, List, Tuple
+import pickle as pk
+from freqtrade.configuration import TimeRange
+
+SECONDS_IN_DAY = 86400
+
+class DataHandler:
```
xmatthias: I would somehow rename this if possible (both the class and the file)
we have an `IDataHandler` (which is the base class for json/hdf5/... hanlders) - and having a (almost identical) name is quite confusing, especially considering that this does A LOT more.

This is more like  Model Preparation or data preparation than just a "datahandler" (it does PCA, and probably much more i didn't look at yet in detail...)

If we fail to come up with a better name, then maybe `ModelDataHandler` would already be better.

xmatthias (commented):
i'd ask you to have flake8 passing (there's currently some 300+ errors) - it'll make looking at it a lot easier  (and will be a necessity anyway).


That will avoid a lot of "can we format this differently please" comments - which are obviously wrong - and will simplify reading through the code - as it's somewhat familiar codestyle.
it's not about CI passing or not at this point - but about general readability of the code - which will surely be increased by that.

freqtrade/freqai/data_handler.py
robcaulk: Thanks for the comments, all great points - I will get to work.

docs/freqai.md
```diff
@@ -0,0 +1,265 @@
...
+## Configuring the bot
+### Example config file
+The user interface is isolated to the typical config file. A typical Freqai
+config setup includes:
+
+```json
+    "freqai": {
+                "timeframes" : ["5m","15m","4h"],
+                "full_timerange" : "20211220-20220220",
+                "train_period" : "month",
```
robcaulk: These to `period` parameters are now indicated as integers based on a unit of days. This needs to be reflected in the doc.

docs/freqai.md
```diff
@@ -0,0 +1,265 @@
...
+                        "BNB/USD"
+                ],
+                "train_params" : {
+                        "period": 24,
+                        "shift": 2,
+                        "drop_features": false,
+                        "DI_threshold": 1,
+                        "weight_factor":  0,
+                },
+                "SPLIT_PARAMS" : {
```
robcaulk: `SPLIT_PARAMS` and `CLASSIFIER_PARAMS` have been renamed, and should be reflected in the doc.

docs/freqai.md
```diff
@@ -0,0 +1,265 @@
...
+
+no. `timeframes` * no. `base_features` * no. `corr_pairlist` * no. `shift`_
+3 * 3 * 3 * 2 = 54._
+
+### Deciding the sliding training window and backtesting duration
+
+`full_timerange` lets the user set the full backtesting range to train and
+backtest through. Meanwhile `train_period` is the sliding training window and
+`backtest_period` is the sliding backtesting window. In the present example,
+the user is asking Freqai to train and backtest the range of `20211220-20220220` (`month`).
```
robcaulk: Period parameters are now integers instead of strings. Needs to be reflected in the doc.

docs/freqai.md
```diff
@@ -0,0 +1,265 @@
...
+using the typical trade command:
+
+```bash
+freqtrade trade --strategy FreqaiExampleStrategy --config config_freqai.example.json --training_timerange '20211220-20220120'
+```
+
+Where the user has now specified exactly which of the models from the sliding window
+that they wish to run live using `--training_timerange` (typically this would be the most
+recent model trained). As of right now, freqai will
+not automatically retain itself, so the user needs to manually retrain and then
```
robcaulk: Typo on "retrain"

docs/freqai.md
```diff
@@ -0,0 +1,265 @@
...
+computational effort and the fit to the training data. If a user has a GPU
+installed in their system, they may benefit from changing `task_type` to `GPU`.
+The `weight_factor` allows the user to weight more recent data more strongly
+than past data via an exponential function:
+
+$$ W_i = \exp(\frac{-i}{\alpha*n}) $$
+
+where $W_i$ is the weight of data point $i$ in a total set of $n$ data points._
+
+`drop_features` tells Freqai to train the model on the user defined features,
```
robcaulk: `drop_features` is not currently in the branch, should be removed from doc or added to branch depending on if the functionality should be retained.

freqtrade/freqai/data_handler.py
robcaulk: How about FreqaiDataHandler?

freqtrade/freqai/freqai_interface.py
robcaulk: I added the `@abstractmethod` but I am not sure what to use in place of `Any` for the return value on `predict()` since we do not know the model class type that the user will want to return. 

Maybe you have a tip for how to handle that?

xmatthias: the return typing is fine i guess (maybe some scikit model supertype could work ...?) - it's about the `return Any` statement itself, really.

If it's an abstract method, you can leave the body empty without problem - so i'd do that.

freqtrade/freqai/data_handler.py
xmatthias: it's a possibility (better than just DataHandler for sure).

i'm not particularly thrilled by the `DataHandler` part in the name, really (but lack a better term for now)

it is handling data for sure - but it's really preparing / handling the dataset for the model ... so maybe something around DataSet could also work (naming is hard - and also not a problem we need to solve now i think).

rokups:
Some feedback:
- "freqai*" -> "ai*" would be cleaner
- config in docs is outdated
- switching stake amount requires chaging "btc_pair", not sure if something could be done about it
- self.dh.load_data(self.dh.model_path) does not need path parameter
- logging should use f-strings for clarity, also i get lots of exceptions about not all arguments being converted when logging
- This happens when trying to train for very first time:
```
! 2022-05-05 15:07:52,158 - freqtrade.resolvers.iresolver - WARNING - Path "/home/rk/src/freqtrade/user_data/freqaimodels" does not exist.
  2022-05-05 15:07:52,167 - freqtrade - ERROR - Impossible to load FreqaiModel 'ExamplePredictionModel'. This class does not exist or contains Python code errors.
  2022-05-05 15:07:52,168 - root - INFO - Took 2.65s.
  2022-05-05 15:07:52,226 - freqtrade.exchange.exchange - INFO - Closing async ccxt session.
! (venv) ~/src/freqtrade % ls /home/rk/src/freqtrade/user_data/freqaimodels
! ExamplePredictionModel.py  __pycache__
```
- Another exception:
```
  File "/home/rk/src/freqtrade/user_data/strategies/FreqaiExampleStrategy.py", line 146, in populate_indicators
    dataframe['target_mean'], dataframe['target_std']) = self.model.bridge.start(dataframe, metadata)
  File "/home/rk/src/freqtrade/freqtrade/freqai/freqai_interface.py", line 100, in start
    self.model = self.dh.load_data()
  File "/home/rk/src/freqtrade/freqtrade/freqai/data_handler.py", line 95, in load_data
    self.data_dictionary["train_features"] = pd.read_pickle(
AttributeError: 'DataHandler' object has no attribute 'data_dictionary'
```

I was unable to do anything from start to the end so far. Stuck with `data_dictionary` exception.

robcaulk:
> Some feedback:
> 
> * "freqai*" -> "ai*" would be cleaner
> * config in docs is outdated
> * switching stake amount requires chaging "btc_pair", not sure if something could be done about it
> * self.dh.load_data(self.dh.model_path) does not need path parameter
> * logging should use f-strings for clarity, also i get lots of exceptions about not all arguments being converted when logging
> * This happens when trying to train for very first time:
> 
> ```
> ! 2022-05-05 15:07:52,158 - freqtrade.resolvers.iresolver - WARNING - Path "/home/rk/src/freqtrade/user_data/freqaimodels" does not exist.
>   2022-05-05 15:07:52,167 - freqtrade - ERROR - Impossible to load FreqaiModel 'ExamplePredictionModel'. This class does not exist or contains Python code errors.
>   2022-05-05 15:07:52,168 - root - INFO - Took 2.65s.
>   2022-05-05 15:07:52,226 - freqtrade.exchange.exchange - INFO - Closing async ccxt session.
> ! (venv) ~/src/freqtrade % ls /home/rk/src/freqtrade/user_data/freqaimodels
> ! ExamplePredictionModel.py  __pycache__
> ```
> 
> * Another exception:
> 
> ```
>   File "/home/rk/src/freqtrade/user_data/strategies/FreqaiExampleStrategy.py", line 146, in populate_indicators
>     dataframe['target_mean'], dataframe['target_std']) = self.model.bridge.start(dataframe, metadata)
>   File "/home/rk/src/freqtrade/freqtrade/freqai/freqai_interface.py", line 100, in start
>     self.model = self.dh.load_data()
>   File "/home/rk/src/freqtrade/freqtrade/freqai/data_handler.py", line 95, in load_data
>     self.data_dictionary["train_features"] = pd.read_pickle(
> AttributeError: 'DataHandler' object has no attribute 'data_dictionary'
> ```
> 
> I was unable to do anything from start to the end so far. Stuck with `data_dictionary` exception.

Hey, ya some bugs popped up during my recent flake8 appeasement. But I am fixing them at the moment - will push working version to you soon :-)

freqtrade/freqai/data_handler.py
```diff
@@ -353,8 +345,8 @@ def principal_component_analysis(self) -> None:
         pca2 = PCA(n_components=n_keep_components)
         self.data["n_kept_components"] = n_keep_components
         pca2 = pca2.fit(self.data_dictionary["train_features"])
-        logger.info("reduced feature dimension by", n_components - n_keep_components)
-        logger.info("explained variance", np.sum(pca2.explained_variance_ratio_))
+        logger.info("reduced feature dimension by %s", n_components - n_keep_components)
+        logger.info("explained variance %f", np.sum(pca2.explained_variance_ratio_))
```
rokups: Bruh use f-strings xD
```cpp
logger.info(f"explained variance {np.sum(pca2.explained_variance_ratio_)}")
```

freqtrade/freqai/data_handler.py
robcaulk: FreqaiDataKitchen ;)?

robcaulk:
> Some feedback:
> 
> * "freqai*" -> "ai*" would be cleaner
> * config in docs is outdated
> * switching stake amount requires chaging "btc_pair", not sure if something could be done about it
> * self.dh.load_data(self.dh.model_path) does not need path parameter
> * logging should use f-strings for clarity, also i get lots of exceptions about not all arguments being converted when logging
> * This happens when trying to train for very first time:
> 
> ```
> ! 2022-05-05 15:07:52,158 - freqtrade.resolvers.iresolver - WARNING - Path "/home/rk/src/freqtrade/user_data/freqaimodels" does not exist.
>   2022-05-05 15:07:52,167 - freqtrade - ERROR - Impossible to load FreqaiModel 'ExamplePredictionModel'. This class does not exist or contains Python code errors.
>   2022-05-05 15:07:52,168 - root - INFO - Took 2.65s.
>   2022-05-05 15:07:52,226 - freqtrade.exchange.exchange - INFO - Closing async ccxt session.
> ! (venv) ~/src/freqtrade % ls /home/rk/src/freqtrade/user_data/freqaimodels
> ! ExamplePredictionModel.py  __pycache__
> ```
> 
> * Another exception:
> 
> ```
>   File "/home/rk/src/freqtrade/user_data/strategies/FreqaiExampleStrategy.py", line 146, in populate_indicators
>     dataframe['target_mean'], dataframe['target_std']) = self.model.bridge.start(dataframe, metadata)
>   File "/home/rk/src/freqtrade/freqtrade/freqai/freqai_interface.py", line 100, in start
>     self.model = self.dh.load_data()
>   File "/home/rk/src/freqtrade/freqtrade/freqai/data_handler.py", line 95, in load_data
>     self.data_dictionary["train_features"] = pd.read_pickle(
> AttributeError: 'DataHandler' object has no attribute 'data_dictionary'
> ```
> 
> I was unable to do anything from start to the end so far. Stuck with `data_dictionary` exception.

I *think* all these feedbacks are resolved in recent commits. Much appreciated your time on testing this. Soon I hope to engage some "beta testers" from #general to help identify common user issues and catch errors like these ones.

freqtrade/freqai/data_handler.py
robcaulk: Havent used this logger before - I honestly just quick searched how it was done elsewhere in the codebase and replicated. As is it seems to work, or maybe I am missing something?

rokups: python is notorious for million ways to format a string. what you found is old code that nobody updated. f-strings are just simpler way to format stuff as you have a single string with variables (or expressions) embedded into it

robcaulk: No doubt about that - anyways using f strings makes it easier (more elegant?) to pass the flake8 line length boss for longer strings.

freqtrade/constants.py
```diff
@@ -99,380 +145,414 @@
...
-                'allowed_risk': {'type': 'number'},
-                'stoploss_range_min': {'type': 'number'},
-                'stoploss_range_max': {'type': 'number'},
-                'stoploss_range_step': {'type': 'number'},
-                'minimum_winrate': {'type': 'number'},
-                'minimum_expectancy': {'type': 'number'},
-                'min_trade_number': {'type': 'number'},
-                'max_trade_duration_minute': {'type': 'integer'},
-                'remove_pumps': {'type': 'boolean'}
+        "freqai": {
```
robcaulk, line 436: @xmatthias can you check this? I added `config['freqai']` to constants in an effort to pass the unittests, but still doesnt seem to pass:

https://github.com/freqtrade/freqtrade/runs/6307348305?check_suite_focus=true#step:8:7483

robcaulk: (also accidentally black formatted the file, but it doesnt appear to cause any bugs here, we can revert it if you prefer)

xmatthias: yeah - for now i'd ask you to revert it. the file should format fine with autopep8 - black changes too much for a PR like this - making Reviews near impossible (almost every line in this file is changed (mostly due to quotes) - which is an irrelevant change for this PR - but it took me near 2 minutes to reach the things you actually changed (the things i SHOULD be looking at).

I don't think that unittests will fail because of this (this does validation - it will NOT fill the schema for you unless explicitly instructed).

freqtrade/optimize/backtesting.py
```diff
@@ -204,12 +204,12 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
         """
         self.progress.init_step(BacktestState.DATALOAD, 1)
 
-        if self.config['freqaimodel']:
-            self.required_startup += int((self.config['freqai']['train_period']*86400) /
-                                    timeframe_to_seconds(self.config['timeframe']))
+        if self.config['freqai']['train_period'] > 0:
```
xmatthias: You should use the following here to allow this to pass (not just unittests) with any config that is not using freqAI (does not have that configured).

```suggestion
        if self.config.get('freqai', {}).get('train_period') > 0:
```

Using this "optional" style will not be required throughout i guess - but for interfacing parts - always assume people have an existing configuration - and that behavior should not break.

freqtrade/constants.py
robcaulk: Reverted, and added changes on top.

freqtrade/optimize/backtesting.py
robcaulk: Thanks, changed it.

freqtrade/freqai/freqai_interface.py
robcaulk: Thanks, fixed.

robcaulk:
~~Hmm, it seems to pass mac and windows but not on linux. Any insight @xmatthias ?~~

It fixed itself...

The error is not very helpful:
https://github.com/freqtrade/freqtrade/runs/6324327210?check_suite_focus=true#step:8:278

xmatthias:
that was for sure CI having troubles. The "current" failure is a fluke - but for now, not worth looking at (90% of CI passes).

freqtrade/templates/FreqaiExampleStrategy.py
```diff
@@ -0,0 +1,194 @@
...
+        return df
+
+    def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
+
+        # the configuration file parameters are stored here
+        self.freqai_info = self.config["freqai"]
+        self.pair = metadata['pair']
+
+        # the model is instantiated here
+        self.model = CustomModel(self.config)
```
xmatthias, line 141: Shouldn't the instantiation happen in something like `bot_start()` - which is only running once per run?

The current location would run this once for every pair in backtesting - and once for every pair per iteration on live ... 

That said - if we assume this as a "builtin" feature of freqtrade - we could take care of this "in the backend" - assuming we detect a freqAI config ...

freqtrade/templates/FreqaiExampleStrategy.py
```diff
@@ -0,0 +1,194 @@
...
+                if pair in whitelist_pairs:
+                    continue  # avoid duplication
+                informative_pairs.append((pair, tf))
+        return informative_pairs
+
+    def populate_any_indicators(self, pair, df, tf, informative=None, coin=""):
+        """
+        Function designed to automatically generate, name and merge features
+        from user indicated timeframes in the configuration file. User can add
+        additional features here, but must follow the naming convention.
```
xmatthias, line 66: which convention? I looked at the docs - but nothing jumped out at me for this ... ?

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,202 @@
...
+        self.feature_parameters = config["freqai"]["feature_parameters"]
+        # self.backtest_timerange = config["timerange"]
+
+        self.time_last_trained = None
+        self.current_time = None
+        self.model = None
+        self.predictions = None
+
+    def start(self, dataframe: DataFrame, metadata: dict, strategy: IStrategy) -> DataFrame:
+        """
```
xmatthias, line 44: Do i understand this right that this will either predict - or train - based on the runmode (for now)?
How would you imagine this to work with automatic retraining?

I think this should be splitup to a method doing "train" and "predict()" (naming might be different ...) - assuming the training in live will also use the sliding window approach (??)

freqtrade/templates/ExamplePredictionModel.py
```diff
@@ -0,0 +1,159 @@
...
+        # optional additional data cleaning
+        if self.feature_parameters["principal_component_analysis"]:
+            pca_components = self.dh.pca.transform(filtered_dataframe)
+            self.dh.data_dictionary["prediction_features"] = pd.DataFrame(
+                data=pca_components,
+                columns=["PC" + str(i) for i in range(0, self.dh.data["n_kept_components"])],
+                index=filtered_dataframe.index,
+            )
+
+        if self.feature_parameters["remove_outliers"]:
```
xmatthias, line 146: This is entering danger zone.

as this is now part of the "user-defined" model - i'm not sure if having it in the config makes sense (it could also be defined in this file).

The problem is - we only have / check X values for correctness (only the ones we know / use in this example).
People might however have 20 different desires for this - and placing all in the config might become problematic.

if we assume this to be "automatic" things - then the default shouldn't have it in the example file - but someplace in the backend.
if we assume the user modifies it (also what it does) - then the config is not the right place. 

It's difficult to control what people put within the `remove_outliers` IF statement (you can't, really)- so you can no longer rely on the fact that it does what the description says - especially if people start sharing different models.

freqtrade/freqai/freqai_interface.py
robcaulk: No, I should correct your understanding:

`start()` will train a new model if it cannot find a pretrained model. **Live** and **backtesting**.
if `start()` does find find a pretrained model, it will simply load that model and inference (use predict()) it. **Live** and **backtesting**

The only difference for **live** is that if it finds an existing model, it checks the amount of time that has elapsed since the end of that training period. If the elapsed time is greater than the config user defined `backtest_period`, then it trains a new model and points any future inferences (predict()) to the new model. 

So your understanding that backtesting = training and live = predicting is incorrect. We train and predict in both backtesting and live. In other words, we need `start()` to detect if we need training/predicting in both backtesting and live. 

Please let me know if it makes sense, and if not, let me know if I can clarify things more.

freqtrade/templates/FreqaiExampleStrategy.py
robcaulk: The issue with putting this in `bot_start()` is that it would carry around too much unnecessary data in ram. Imagine we are  training on 2 years of data for each of 30 pairs, your suggestion would keep all that data duplicated in ram. And what is worse, we may not even need that data again for a while if the user is simply inferencing a model saved to the drive. This class is just to help the strategy locate existing models to load them and make predictions with them, or to know whether or not it should train a new model and save it. As soon as we finish training/making a prediction, we can let python clean this up and move to the next pair because we have saved any relevant model/metadata to disk to be loaded later.

Correct me if my understanding of how the data would be held in ram is incorrect. But if it is true, then we still could move this line to `bot_loop_start()` in an effort to make the strategy "cleaner" for the user.

freqtrade/templates/ExamplePredictionModel.py
robcaulk: Yeah, I agree with what you are saying here. Let me look into transitioning these userconfigurable options over to the hyperparameters type setup so that they can put them in the strategy directly. 

Maybe ALL freqai stuff should go in there...I need to spend time learning how it works and then I will come back to with a proposed solution.

xmatthias: maybe - maybe not - that's ... difficult to determine.

i think you'll want at least some setup parameters in the config - so freqtrade can determine when to retrain.

freqtrade/templates/FreqaiExampleStrategy.py
xmatthias: i think you shouldn't think about these 2 (model loading, data loading) as attached / combined.

If i'm running a model (live) - i'd probably want to keep the model in memory - avoiding the "disk read" for the model.
now training is (obviously) different - you'll only want to load the data "when needed" - but i don't really see why the two can't be (at least partially) decoupled.

My point here is for live mode, mostly - if i assume i run a model on 20 pairs - we'd be loading 20 models for every pair on every iteration (or at least for every candle). 
Assuming that a trained model will not be huge (it could be - but default models won't) - i would assume keeping that in memory would be fine?

freqtrade/freqai/freqai_interface.py
xmatthias: thanks - it "kindof" makes sense now - except that it doesn't - or rather - it raises more questions.

In my understanding, the goal is to move the "live retraining" to a separate thread/process - to avoid impacts on "live trades".

By calling start (and training there) from this part of the strategy - we're effectively retraining "in-line" (or in-process) - slowing "this iteration" down by "training + prediction" (however long this is).

this might not be a problem - but depending on the trade state (close to a stoploss, maybe) - we might be missing stops/exits while the model is being retrained as the bot-loop is blocked - which is clearly not a safe way for this.

Maybe this can set a flag for the retraining (this is where the "non-repeated initialization" comes into play) - which will let freqtrade know that "now we need to retrain everything, please spawn new processes for me".
Obviously there's alternative approaches (like a "diff-thread-callback") which is used for this "out of process" processing (that's not the point of this comment) - but my point is - having this in-line in live (!) is probably not the best idea

robcaulk: >By calling start (and training there) from this part of the strategy - we're effectively retraining "in-line" (or in-process) - slowing "this iteration" down by "training + prediction" (however long this is).
>having this in-line in live (!) is probably not the best idea

You are 100% correct :-) Moving the training to a separate thread is now on the road map - I will do that after I sort out the parameters.

freqtrade/templates/FreqaiExampleStrategy.py
robcaulk: Yeah it is use-by-use I supposed. However, the ultimate need is speed in the end, and if we can reduce the wait time to prediction by keeping pretrained model ready to go - then it would be worth it. 

May need a bit of architecture work - or it might be only a couple lines. I am adding it to the roadmap.

PR #6778, opened 2022-05-03
Add bot_loop_start() call in plotting.py

### LINKED ISSUE #6776: Support `bot_loop_start` when calling `plot-dataframe`
## Describe your environment
(if applicable)

  * Operating system: MacOS Monterey 12.3.1 (21E258)
  * Python Version: 3.8.9 (`python -V`)
  * CCXT version: 1.80.61 (`pip freeze | grep ccxt`)
  * Freqtrade Version: develop-4d5cdf0b8 (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)

TLDR: Suggest making the `bot_loop_start` callback work when using `plot-dataframe' CIL command. 

Motivation: 
- I use `bot_loop_start` to import external informative signals for my strategy
- This works well for live trading and backtesting
- A key requirement of workflow is to export dataframe-plots for inspection of the backtest

Ask:
- Perhaps this is already supported and just my bad. PLMK.
- When running `plot-dataframe` on a strategy that uses `bot_loop_start` an error is returned indicating that none of the external informative signals were defined and hence not available in `populate_indicators`. Note: backtests and trading live work without error for this. 

PLMK if anything else I could provide. And if there is a simple fix - a code pointer and I could look into implementing. 

Many thanks

markdregan:
## Summary
plotting.py was missing a call to strategy.bot_loop_start() resulting in strategies using this callback to not work.

Made changes and confirmed plotting now works for strategies using bot_loop_start() callback.

Solve the issue: #6776

## Quick changelog

- add strategy.bot_loop_start() callback to plotting.py

## What's new?

Strategies that use bot_loop_start() callback can now use the `plot-dataframe` CIL command.

### DIFF
--- a/freqtrade/plot/plotting.py
+++ b/freqtrade/plot/plotting.py
@@ -634,6 +634,7 @@ def load_and_plot_trades(config: Dict[str, Any]):
     exchange = ExchangeResolver.load_exchange(config['exchange']['name'], config)
     IStrategy.dp = DataProvider(config, exchange)
     strategy.bot_start()
+    strategy.bot_loop_start()
     plot_elements = init_plotscript(config, list(exchange.markets), strategy.startup_candle_count)
     timerange = plot_elements['timerange']
     trades = plot_elements['trades']


xmatthias (approved):
Great, thanks a lot for fixing! :+1:

PR #6779, opened 2022-05-03
cleaned up backtesting

SmartManoj:
## Summary

Solves [this bug ](https://github.com/freqtrade/freqtrade/runs/6282019573?check_suite_focus=true).

[Reference](https://stackoverflow.com/questions/41516287/python-del-does-not-work-as-destructor)
[
]()

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -144,6 +144,7 @@ class Backtesting:
         self.init_backtest()
 
     def __del__(self):
+        logger.info('Inside Backtesting.__del__()')
         self.cleanup()
 
     @staticmethod
@@ -151,6 +152,7 @@ class Backtesting:
         LoggingMixin.show_output = True
         PairLocks.use_db = True
         Trade.use_db = True
+        logger.info('Cleaned Backtest')
 
     def init_backtest_detail(self):
         # Load detail timeframe if specified
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -144,7 +144,7 @@ class Backtesting:
         self.init_backtest()
 
     def __del__(self):
-        logger.info('Inside Backtesting.__del__()')
+        logger.debug('Inside Backtesting.__del__()')
         self.cleanup()
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


xmatthias (changes requested):
The least I'd expect from a PR changing more than 1 line is a short description describing the change (or what it's trying to fix if it's not linking to an open issue).

This can include a link to additional information - but ONLY a link to a stack-overflow question (which is more discussion and question than answer) will not suffice to describe the problem you're trying to fix, and if i need to research / guess the problem you're trying to fix before actually reviewing, then this will only lead to missunderstandings - as i might interpret the question slightly different, or read a different answer than you did.

Therefore, please update your first post with a description of the problem you're seeing / trying to solve.

tests/conftest.py
```diff
@@ -33,6 +33,12 @@
 logging.getLogger('').setLevel(logging.INFO)
 
 
+@pytest.fixture(autouse=True)
+def run_around_tests():
+    yield
+    # assert Trade.use_db
```
xmatthias: An empty fixture will not bring any benefit - so it's pointless code that needs executing - resulting in a slightly longer execution time.

this should either reset `use_db` (assuming that's the bug you're trying to solve) to ensure this is never in the wrong state after a test (backtesting is not the only part of the tests that can cause this to be left wrong if one test fails)

or the fixture should be removed.

freqtrade/optimize/backtesting.py
```diff
@@ -144,13 +144,15 @@ def __init__(self, config: Dict[str, Any]) -> None:
...
     def __del__(self):
+        logger.info('Inside Backtesting.__del__()')
         self.cleanup()
 
     @staticmethod
     def cleanup():
         LoggingMixin.show_output = True
         PairLocks.use_db = True
         Trade.use_db = True
+        logger.info('Cleaned Backtest')
```
xmatthias: If you really think this output is needed (I think it's not) - it should be debug level - not info.
otherwise it'll be additional (for the user pointless) output that will be visible after running backesting.
Same applies to the change above.

xmatthias:
If i say "there's no point in a empty fixture" - then commenting that part will still have the pointless code there.

The only relevant line was added directly to 45b328af2, i'll therefore close this as it's no longer needed.

SmartManoj:
> If i say "there's no point in a empty fixture" - then commenting that part will still have the pointless code there.
> 
Just kept for future reference

PR #6780, opened 2022-05-04
Dry run taker or maker fees

samgermain:
Freqtrade statically passes maker to get fees for an order, this new method isn't perfect (limit orders can still be taker orders) but its better than what was there before.


freqtrade/exchange/exchange.py
```diff
@@ -785,6 +785,26 @@ def _get_stake_amount_considering_leverage(self, stake_amount: float, leverage:
 
     # Dry-run methods
 
+    def taker_or_maker(
+        self,
+        order_reason: Literal['entry', 'exit', 'stoploss'],  # TODO: stoploss
+    ):
+        order_type = self._config['order_types'][order_reason]
+        if order_type == 'market' or order_reason == 'stoploss':
```
samgermain: I'm not sure if the stoploss assumption is correct, are freqtrade stoploss limit orders always taker orders, or sometimes maker orders? Theres no config price_side for it so I"m not sure

xmatthias: Depends on the exchange (and the settings)
kraken only has stop market orders - so you know what you are (taker)

binance only has stop limit orders - but "default" settings set the limit price 1% below the stop - so it's reasonable to assume that you're a taker as well. 
users can adjust the limit value - and the market can "gap down" - so you'll really only know if you're taker or maker once the order fills.

other exchanges offer both limit and market options - so you can obviously have both.

---

In reality (independent of stoploss) you'll know this once the order fills (also a limit order can be using taker fees).

Maybe the following assumption provides a better / more realistic result:
If the order fills right when it's created - it's taker fees
If it "stays there" for at least one cycle - it's maker fees.


**Edit** formatting - the last point just jumped at me while looking at this PR

samgermain: > binance only has stop limit orders 

binance futures can do both

samgermain: > Maybe the following assumption provides a better / more realistic result:
> If the order fills right when it's created - it's taker fees
> If it "stays there" for at least one cycle - it's maker fees.

how would I check if a cycle has been passed?

xmatthias: based on the call location (where the "close limit order" is called from).
There will only be one location where the order is created - just mark that with a "first" parameter (or whatever) - and if it fills there, it's taker (in reality independent of the order type), otherwise it's maker fees.

Stoploss orders might be slighty different / more complex ... but i think having this right for regular buy/sell orders will already be a significant improvement.

> binance futures can do both

That's not the point - it's more that there can be all sorts of such things

samgermain: > That's not the point - it's more that there can be all sorts of such things

I was just letting you know

xmatthias (approved):
I've now found the time to implement how i proposed it above.

if an order fills immediately - it's a taker order.
if the order stays on the "virtual orderbook" for at least one cycle - it'll be a maker order - as someone else "matched" our order. This should align us mostly to how real exchanges work.
Differences will obviously depend on the exchanges - as not all exchanges provide different rates (binance does not).

PR #6782, opened 2022-05-04
Revoked Api announce

lulu-angry:
Hello Freqtrade Dev Team,

I think it be interesting to implement a function that lets you know that your API key has been revoked through telegram or other?
About a announce-like or a warning in telegram.


xmatthias:
Creating a PR for this is clearly wrong. 
If any, this should be an issue - but it's also not clear what you mean.

please open an issue, filling out the issue template - and specify a bit more pecisely what it is you mean exactly.

PR #6786, opened 2022-05-06
Profit reporting fix

eSeR1805:
## Summary

Perform profit reporting only when at least one entry order is filled.
This is just to get the ball rolling on this and discuss the approach. 
LE: occurred to me that an alternate solution cointained to `rpc` might be better.

Solves the issue: Not really an issue but `adjust_entry_order` PR can make it more common that profit is reported although order is still awaiting fill. 
## Quick changelog

- `calc_profit` and `calc_profit_ratio` return 0.0 when no filled orders in the trade
- `_rpc_status_table` to allow reporting of 0 fiat profit.

## What's new?

Nothing new. Just adjustment of existing logic.

## TODO
- [ ] resolve test failures (profit missmatch) - likely due to orders being populated as unfilled


xmatthias:
You could probably just do a `git rebase origin/develop` on this branch (assuming origin is the freqtrade/freqtrade remote) - to get the "prior" commits out.

alternatively - you can use `git cherry-pick <lastcommit>` on a new branch to move your changes over to a new branch.

PR #6792, opened 2022-05-08
consistent exchange name

SmartManoj:
## Summary

consistent exchange name on notify methods


freqtrade/freqtradebot.py
```diff
@@ -831,7 +831,7 @@ def _notify_enter(self, trade: Trade, order: Dict, order_type: Optional[str] = N
             'type': msg_type,
             'buy_tag': trade.enter_tag,
             'enter_tag': trade.enter_tag,
-            'exchange': self.exchange.name.capitalize(),
+            'exchange': trade.exchange.capitalize(),
```
xmatthias, line 834: i honestly don't see what benefit this brings. Freqtrade won't support multiple exchanges (and we don't aim to - within one process).

So please describe why you think this is better than using the current approach - otherwise i don't see why this would be better than the prior approach.

Also note:
one sentence (repetition of the title) is not a good PR description.

Especially for a PR that provides no visible benefit, it will be beneficial to see WHY you think this change is necessary. 
I currently don't see this as improvement - it's just an irrelevant change - so i'm not sure why this should be accepted / preferred over the current approach.

xmatthias (changes requested):

mkavinkumar1:
> notify methods

https://github.com/freqtrade/freqtrade/blob/8a6a6ec9111d4a42d9fd1d1f60df0133bbba7747/freqtrade/freqtradebot.py#L834
https://github.com/freqtrade/freqtrade/blob/8a6a6ec9111d4a42d9fd1d1f60df0133bbba7747/freqtrade/freqtradebot.py#L1505

First returns Gate.io
Second returns Gateio

xmatthias (approved):

PR #6793, opened 2022-05-08
logged balance details

SmartManoj:
Logged balance details when balance is lower than stake


xmatthias (changes requested):
I think you've contributed often enough to freqtrade now to know:
please put a brief description of the change in the PR request.

You've also been told about this in https://github.com/freqtrade/freqtrade/pull/6779 - so i expect people to learn / improve. Failure to do so will lead to a ban in the end.

xmatthias (approved):

PR #6797, opened 2022-05-08
then -> than

lukeingalls:
## Summary

Fix grammar in docs

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Great spot, thanks for taking the time to fix this! :100:

PR #6811, opened 2022-05-10
corrected minor "typo" in formatting

DJCrashdummy:
sorry for opening such a trivial merge request, but if i wait for more to sum up, i'll probably never find resp. think about it again.


xmatthias (approved):
Also / especially small PR's are priceless - as it's a formatting issue most people (including myself) would probably ignore.

LGTM 👍

PR #6814, opened 2022-05-11
 minor polish for explanation of `backtesting --breakdown`

DJCrashdummy:
why, of course! **after** merging an absolutely minor correction, i found an other mistake in the docs which should really be corrected.
anyway... i used it, to also expand the explanation a little bit to make things even clearer.

- corrected the command to fit the explanation
- added a little explanation how to read the weekly & monthly breakdown


xmatthias (approved):
Do not worry about creating PR's ;)

PR #6823, opened 2022-05-13
cleaned up backtesting

SmartManoj:
## Summary

Solves the [bug](https://github.com/freqtrade/freqtrade/runs/6425715015?check_suite_focus=true)


SmartManoj:
@xmatthias The future reference I [was talking ](https://github.com/freqtrade/freqtrade/pull/6779#issuecomment-1120363380) about 

This should be in docs too.
And [this](https://stackoverflow.com/a/57379818) too to find erroneous test cases

xmatthias (approved):
Great - thanks for taking the time tracking this down! :+1: 

LGTM

PR #6827, opened 2022-05-14
Fix: Refresh SL on entry order replacement

eSeR1805:
## Summary

Fix backtesting behavior for entry order adjustment feature.

Solve the issue: Stoploss latched based on first order.

## Quick changelog

- Update `_enter_trade` so that `open_rate` is refreshed when a `LocalTrade` object is supplied with 0 filled orders;
- Update `check_order_replace` to refresh stoploss when entry order is replaced;
- Update `adjust_stoploss` to allow refreshing the SL
- Added testcases

## What's new?

Nothing new. Fix for BT side of PR: https://github.com/freqtrade/freqtrade/pull/6692

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -780,8 +780,6 @@ class Backtesting:
                     # interest_rate=interest_rate,
                     orders=[],
                 )
-            elif trade.nr_of_successful_entries == 0:
-                trade.open_rate = propose_rate
 
             trade.adjust_stop_loss(trade.open_rate, self.strategy.stoploss, initial=True)
 
@@ -814,11 +812,11 @@ class Backtesting:
                 remaining=amount,
                 cost=stake_amount + trade.fee_open,
             )
+            trade.orders.append(order)
             if pos_adjust and self._get_order_filled(order.price, row):
                 order.close_bt_order(current_time, trade)
             else:
                 trade.open_order_id = str(self.order_id_counter)
-            trade.orders.append(order)
             trade.recalc_trade_from_orders()
 
         return trade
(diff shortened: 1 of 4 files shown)


freqtrade/optimize/backtesting.py
```diff
@@ -780,6 +780,8 @@ def _enter_trade(self, pair: str, row: Tuple, direction: LongShort,
                     # interest_rate=interest_rate,
                     orders=[],
                 )
+            elif trade.nr_of_successful_entries == 0:
+                trade.open_rate = propose_rate
```
xmatthias: We do assign open_rate in `close_bt_order()` in `trade_model.py` when the order closes.

Thinking about that a bit more - i think that is the right place to adjust the stoploss - we have a condition for "first order filled" there anyway ... so that'll be the more obvious place.

xmatthias (commented):
I actually don't think this fully fixes the bug you're trying to solve (not for dry/live anyway).

~For a limit order that is not immediately filled - this might work as that the next iteration will adjust the stoploss (it'll actually adjust it until the first order fills ...) (except that it doesn't - as the method is never called with "refresh=True" from non-backtesting code).~

~If the order fills after the adjustment (you adjusted beyond the spread) - then you'll never get `adjust_stop_loss()` called with `nr_successful_entries == 0` - and the stoploss will not be adjusted.~

~i think the proper handling for this should be once the order is closed (both for BT and for live).
`recalc_trade_from_orders()` has such code - so maybe calling the stoploss code from there might make more sense?~

Edit:
This isn't a problem in live - Live will only call adjust_stoploss once the first order is filled. I've adjusted a test to actually show this.

xmatthias (approved):
LGTM :+1:

PR #6832, opened 2022-05-15
Freqai: an interface for users to build/train/backtest predictive models and run them live

robcaulk:
## Summary

Add an interface for users to easily create and interface their own predictive models with Freqtrade strategies.

replaces #6774 

## What's new?

-Build data handling tools
-Provide examples
-Furnish data analysis tools
-Enable live auto retraining
-Add doc

## TODO

- [x] add automatic retraining during live (which may need to use data-download if training period extends beyond exchange candle limits).
- [x] decide on location of freqai parameters and validate them. Put them as hyperoptable parameters?
- [x] finish doc
- [x] add ability to accept open-ended timerange
- [x] use .get throughout datakitchen.py/freqai_interface.py for any optional config parameters 
- [x] let user run config from anywhere on the system
- [x] move live training to separate thread
- [x] Set a "DefaultPredictionModel" instead of just providing an "ExamplePredictionModel." 
- [x] Auto collect features based on what user feeds into the strategy (i.e. remove the "base_features" from the config)
- [x] Store pretrained model in memory for live so that the model doesnt need to be loaded before prediction (speed considerations)
- [x] Download all data in the corr_pairlist automatically? (user needs to know that they need all data for corr_pairlist in addition to the whitelist, meanwhile download-data only grabs whitelist pairs)
- [x] Automatically detect when user changes indicators, and then tell FreqAI to retrain if no model matches. Right now it will simply error out if it has a trained model with the config['freqai']['identifier'] and the given --timerange because it will load the old model but feed in new sets of indicators - which will not work. 
- [x] Lock the model during training so that the bot doesnt try to access it at the same time that FreqAI is saving the new one.
- [x] Suppress catboost folder 
- [x] ensure fluidity of reusing models created during backtesting for live (possible right now but needs to be more userfriendly)
- [x] remove assertions
- [x] Add some prioritization of untrained pairs to avoid case where one pair dominates separate thread due to alignment of training times and train period.
- [x] Improve prioritization so that we simply hold a queue of pairs needing training, as one is trained, it goes to end of queue.
- [x] Let user train high frequency subdaily during backtesting (already available in live)
- [x] Add functionality to let users indicate a desired file storage format during dry/live
- [x] Increase performance by only making predictions on most recent candle from strategy provided dataframe. 
- [x] Ensure frequi is displaying true historical predictions instead of simply updating those with newly trained models.  
- [x] Auto purge obsolete models to enable sustained high frequency retraining
- [x] Add expiration check to models so that FreqAI isnt inferencing old models on new data. 
- [x] Make config params more descriptive
- [x] Set up base regression class to avoid unnecessary user code duplication (thanks @xmatthias )
- [x] add unit/regression tests
- [x] Make LightGBM the default model (because it compiles easily on ARM)
- [x] Reduce the forced inheritence (remove @abstractmethod and introduce a secondary generic template to be inherited from instead of inheriting directly from freqai_interface.py) 
- [x] report training progress (like "pair 2/5, training period 5/22")
- [x] Identify "data missing at end" in backtesting (using open end timerange with 1d backtest_period and data stopping 1 day early should not crash).
- [x] Filter all features from the dataframe which is returned to the strategy (to help Frequi in the case of large feature numbers)
- [x] Add a separator in model_filename to avoid possible regex complications with coins ending in integers
- [x] Remind users in DOC that VolumePairlist is not supported (and describe why).
- [x] Describe `target_roi` in doc
- [x] reformat doc to a more typical parameter - description list.  
## Wishlist 

- [x] Add a "follow_mode" functionality so that one "follower" instance of Freqtrade just grabs predictions stored on a centralized server, while another "leader" instance crunches the predictions and saves them. 
- [x] Allow follower to exist on a separate system which is rsynced to the leader
- [x] Restructure the existing background thread data-download. Currently the background thread needs to ping the exchange API at the same time as the foreground botloop, which may introduce ratelimit issues and latency problems. It is preferable for Freqai to simply collect the normal botloop candles and store them to be grabbed for retraining. *Currently under construction in feat/freqai-datarehaul branch
- [x] Return a continuous prediction confidence (perhaps the dissimilarity index directly) and use it to inform the stake size.
- [x] Improve return values from FreqAI to strategy so that we can accommodate any user data they want (right now fixed to prediction values and target stats)
- [x] have a separate thread constantly scanning pairs for retraining - right now it can only train one pair per candle. 
- [x] Add docker accessibility (thanks @xmatthias)
- [x] Add multi-target training
- [x] Add classifier 
- [x] Add live trade database analysis/model feedback  

## To be added in separate future PRs:
- Dynamically set the target forecasting period to best suit the recorded performance of the most recent training model. This may be achieved by minimizing the candle distance between the predictions and the recorded performance.
- Add functionality for adding external API data sources (needs proper cleaning, merging, storing, backtesting, live/dry debugging)
- Move all indicator duplication into FreqAI, away from strategy? Right now users need to keep some permanent loops in their strategy, but it would be preferable if this were hidden from user so they have less code to learn)
- Allow user to feed funding rate into model training/prediction


freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,267 @@
...
+        corr_dataframes, base_dataframes = self.dh.load_pairs_histories(new_trained_timerange,
+                                                                        metadata)
+
+        unfiltered_dataframe = self.dh.use_strategy_to_populate_indicators(strategy,
+                                                                           corr_dataframes,
+                                                                           base_dataframes,
+                                                                           metadata)
+
+        self.model = self.train(unfiltered_dataframe, metadata)
+        self.dh.save_data(self.model)
```
xmatthias: You'll need to mutex a few things here i think (not necessarily on this line - but also on others where appropriate ...).

as this is running in a different thread noow, you'll have to consider that freqtrade will read the model while the new model is being written.

it's pretty easy with `from threading import Lock` - and then wrapping it in a lock instance (look at `freqtradebot.py` for an example).

robcaulk: Thanks, I will put this on on the todo list.

freqtrade/freqai/data_kitchen.py
```diff
@@ -68,14 +69,33 @@ def __init__(self, config: Dict[str, Any], dataframe: DataFrame, live: bool = Fa
                 config["freqai"]["backtest_period"],
             )
 
+    def assert_config(self, config: Dict[str, Any], live: bool) -> None:
+        assert config.get('freqai'), "No Freqai parameters found in config file."
```
xmatthias: Please don't use "assert" in non-test code (basically outside of the `tests/` directory.

It can be disabled via Python flag (`python -O`) - converting all of these asserts into "noop" statements.

robcaulk: Ok thanks. Which method would you prefer that I use instead to verify the config has the necessary information?

robcaulk: I guess I can just test for it and then log an error if it doesnt pass?

xmatthias: I'm not certain you'll need explicit asserts for this (at least not for all options - maybe only for freqai (the parent component) - as you can't require that without breaking all existing configs).

a properly configured json schema (`constants.py` - `'required'` within the `freqai` object) usually fills this need already (check if things are there, correct type, within a certain range, ...) - _unless_ it's some complex dependencies (if A is set, B can only be X or Y, otherwise it can be Z too) ...

freqtrade/freqai/freqai_interface.py
robcaulk: After further thought and reconsideration of the new code structure (I made some significant changes to how the data is handled in recent days) this is not an issue. 

Here is  why: the new model file has a different name and is being saved in a different directory with a different name. FreqAI is controlling where to look for the model file eachtime `populate_indicators()` is executed for a pair. The code is written right now so that if a new thread is spawned to train a model, FreqAI continues on to load the old model and give predictions back to the strategy. That separate thread will finish training sometime in the future and write a new file in a new directory. Next time `populate_indicators()` is called for the same pair, FreqAI directs the newest model file to be loaded.

At the moment, I dont see anywhere where we need to mutex anything, but I will keep it in mind as code develops.

freqtrade/freqai/data_kitchen.py
robcaulk: Moved validation to `config_validation.py`. Should be resolved.

freqtrade/commands/data_commands.py
```diff
@@ -50,7 +50,13 @@ def start_download_data(args: Dict[str, Any]) -> None:
     exchange = ExchangeResolver.load_exchange(config['exchange']['name'], config, validate=False)
     markets = [p for p, m in exchange.markets.items() if market_is_active(m)
                or config.get('include_inactive')]
-    expanded_pairs = expand_pairlist(config['pairs'], markets)
+    if config.get('freqai') is not None:
+        assert config['freqai'].get('corr_pairlist'), "No corr_pairlist found in config."
```
robcaulk: Should be removed now that we validate `corr_pairlist` in `conf_validatation.py`

docs/freqai.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+### Building the feature set
+
+Features are added by the user inside the `populate_any_indicators()` method of the strategy 
+by prepending indicators with `%`:
+
+```python
+    def populate_any_indicators(self, metadata, pair, df, tf, informative=None, coin=""):
+        informative['%-''%-' + coin + "rsi"] = ta.RSI(informative, timeperiod=14)
```
robcaulk: Remove duplicate `%-`

docs/freqai.md
```diff
@@ -0,0 +1,413 @@
...
+    def populate_any_indicators(self, metadata, pair, df, tf, informative=None, coin=""):
+
+
+        # Add generalized indicators here (because in live, it will call only this function to populate 
+        # indicators for retraining). Notice how we ensure not to add them multiple times by associating
+        # these generalized indicators to the basepair/timeframe
+        if pair == metadata['pair'] and tf == self.timeframe:
+            df['%-day_of_week'] = (df["date"].dt.dayofweek + 1) / 7
+            df['%-hour_of_day'] = (df['date'].dt.hour + 1) / 25
+
```
robcaulk, line 247: Missing closing ```

robcaulk:
@xmatthias, flake8 is displeased with the complexity of `start_download_data()` as evidenced here:

https://github.com/freqtrade/freqtrade/runs/6592776080?check_suite_focus=true#step:9:7

This left me no (apparent) choice but to create an additional function in `pairlist_helpers.py` to accommodate the auto downloading of corr_pairlist data from the config. Let me know if this is an issue, and if so, please advise your preferred method.

robcaulk (commented):
@xmatthias I am gonna need you to review this commit - it feels quite hacky to do it this way despite the fact that it works. 

The issue is that freqai is using `refresh_backtest_ohlcv_data()` to download data on the fly in live/dry. This requires the instantiation of an Exchange object, but when we are in 'futures' mode, the Exchange instantiation automatically loads the leverage_tiers, which takes....too long and we clearly have no need for them. 

Please advise if there is a better way to download the data without needing to instantiate an Exchange. 

Another idea may be to have a new RunMode.FREQAILIVE ...

Thanks for your time.

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1002 @@
...
+
+        self.model_filename = "cb_" + coin.lower() + "_" + str(int(trained_timerange.stopts))
+
+        # self.freqai_config['live_trained_timerange'] = str(int(trained_timerange.stopts))
+        # enables persistence, but not fully implemented into save/load data yer
+        # self.data['live_trained_timerange'] = str(int(trained_timerange.stopts))
+
+    def download_new_data_for_retraining(self, timerange: TimeRange, metadata: dict) -> None:
+
+        exchange = ExchangeResolver.load_exchange(self.config['exchange']['name'],
```
xmatthias: I assume this is for the live runs?  (if yes, maybe add it as comment via docstring?)

Why are you creating a new instance here instead of using the one that exists already?
by creating a new instance, you're creating problems with rate-limiting (it's now a new, separate rate-limiter) - and with that potentially problems for both instances - as they might refresh both (assuming your refresh here goes "across" a new candle time - then both would reload data at the same time - very likely locking you out (depending on the exchange, but still a very real possibility).

the below used method (`refresh_backtest_ohlcv_data()`)  was due for removal, and was only "left" for edge. Now i don't mind having this used here - but Edge does not recreate a new exchange class - so the regular rate-limit stuff can still fly onwards.

you may need to rethink the architecture a bit for training (i'd not want to expose the exchange class directly to the interface) - but i don't think that'll be a problem ?

robcaulk: The reason I instantiated a dummy exchange was becaue I didn't find a way to access FreqtradeBot.exchange without passing it to the strategy and then into FreqAI. Is there an easier way to gain access to FreqtradeBot.exchange from inside the strategy? 

P.S. edge is instantiated inside FreqtradeBot, so it is passed the exchange directly. FreqAI is not, it is instantiated inside the strategy.

robcaulk: scratch that, I can use the dataprovider like this:

strategy.dp._exchange to grab it.

robcaulk: After further private discussion, we concluded that the cleanest solution is to restructure the handling of training data in dry/live background threads. This is a relatively intensive rehaul, so it has been added to the wishlist.

However, currently we are not observing any ratelimiting issues associated with pinging the exchange on the background thread (in parallel with the foreground bot loop). So we will keep the existing "hack" until we have time to restructure the data handling.

Resolving this thread accordingly.

requirements-freqai.txt
```diff
@@ -0,0 +1,8 @@
+# Include all requirements to run the bot.
+-r requirements.txt
+
+# Required for freqai
+scikit-learn==1.0.2
+scikit-optimize==0.9.0
+joblib==1.1.0
+catboost==1.0.4
```
xmatthias: It seems like installation of catboost on ARM is not possible as they don't provide ARM wheels via pypi.

are there instructions / workarounds for that available (if yes, they should be in the docs) ...

robcaulk: Noted, I will make sure to make LightGBM the default prediction model, and add some info about this in the docs ;)

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1257 @@
...
+        `expiration_hours` in the configuration file.
+        :params:
+        trained_timestamp: int = The time of training for the most recent model.
+        :returns:
+        bool = If the model is expired or not.
+        """
+        time = datetime.datetime.now(tz=datetime.timezone.utc).timestamp()
+        elapsed_time = (time - trained_timestamp) / 3600  # hours
+        max_time = self.freqai_config.get('expiration_hours', 0)
+        return elapsed_time > max_time
```
xmatthias: I think the default here is wrong.  (or should be checked differently)
if it's not set in the config - it'll default to 0 - expiring the model immediately - so you can only get buys right after the training.

it's also kindof the only parameter that's in hours - all others are in days - so configuration is kindof inconsistent (although i'd also not want to configure this in days ...).

robcaulk: Ya I think the default value is incorrect here. Thanks for noticing that!

Re: hours
This is toward the new more descriptive naming scheme that all config parameters will take. It will include units :-) so it will be train_period_days etc. I will convert all in one fell swoop when I get a chance (as the todo list indicates). 

If you have any recommendations for renaming config parameters, feel free to drop them here. In my opinion most will be getting new more descriptive names.

freqtrade/freqai/prediction_models/LightGBMPredictionModel.py
```diff
@@ -0,0 +1,127 @@
...
+        :params:
+        :data_dictionary: the dictionary constructed by DataHandler to hold
+        all the training and test data/labels.
+        """
+
+        eval_set = (data_dictionary["test_features"], data_dictionary["test_labels"])
+        X = data_dictionary["train_features"]
+        y = data_dictionary["train_labels"]
+
+        model = LGBMRegressor(seed=42, n_estimators=2000, verbosity=1, force_col_wise=True)
```
xmatthias: i think instead of hardcoding some params here which are configurable in catboost (e.g. `n_estimators`) - you might wanna use `**self.model_training_parameters` instead. While the parameters will not align - it'll allow changes iva the config more easily

robcaulk: Yes, absolutely. Thanks for pointing it out - I'll add it to my next commit. 

You are very welcome to make these kinds of changes, btw ;)

freqtrade/templates/FreqaiExampleStrategy.py
```diff
@@ -0,0 +1,342 @@
...
+            if set_generalized_indicators:
+                df["%-day_of_week"] = (df["date"].dt.dayofweek + 1) / 7
+                df["%-hour_of_day"] = (df["date"].dt.hour + 1) / 25
+
+                # user adds targets here by prepending them with &- (see convention below)
+                # If user wishes to use multiple targets, a multioutput prediction model
+                # needs to be used such as templates/CatboostPredictionMultiModel.py
+                df["&-s_close"] = (
+                    df["close"]
+                    .shift(-self.freqai_info["feature_parameters"]["period"])
```
xmatthias: shouldn't `shift()` here be using the "shift" parameter? that would seem to make more sense to me at first glance ... ?

robcaulk: Thanks for asking, this is another example of why the config parameters  need a fresh renaming based on a consistent and explicit naming scheme.  I reckon we've put that name change off long enough  - I'll do that next :-)

The answer to your question is "no.": 

For records-sake: `shift` is a parameter that adds previous candles to as new features to the existing feature set. e.g. shift 2 will take all features created by the user and shift them once and twice and add both of those shifts to the featureset automatically. 

`period`, on the other hand, is the `label_period_candles` (probably the future name of this one) which is the distance into the future  to look for setting the labels (also referred to as targets).

xmatthias: i did understand that - it solely wasn't clear why we'd shift it back by the same amount than "rolling" - it's however clear now after thinking about it a bit more

tests/freqai/conftest.py
```diff
@@ -0,0 +1,68 @@
...
+    return freqaiconf
+
+
+def get_patched_data_kitchen(mocker, freqaiconf):
+    dd = mocker.patch('freqtrade.freqai.data_drawer', MagicMock())
+    dk = FreqaiDataKitchen(freqaiconf, dd)
+    return dk
+
+
+def get_patched_strategy(mocker, freqaiconf):
```
xmatthias: i'd name this a bit more explicit - the name kindof suggests that it'll load a generically patched strategy - maybe `get_patched_freqai_strategy()` ?

tests/freqai/test_freqai.py
```diff
@@ -0,0 +1,95 @@
...
+
+#     base = np.random.normal(20, 2, size=size)
+
+#     date = pd.date_range(start, periods=size, freq=f'{tf_mins}min', tz='UTC')
+#     df = pd.DataFrame({
+#         'date': date,
+#         'open': base,
+#         'high': base + np.random.normal(2, 1, size=size),
+#         'low': base - np.random.normal(2, 1, size=size),
+#         'close': base + np.random.normal(0, 1, size=size),
```
xmatthias: you might want to be careful with the `'close'` part - it must be between high/low - which ain't the case right now (obviously depending on the random state)

tests/freqai/test_freqai.py
```diff
@@ -0,0 +1,95 @@
...
+    with pytest.raises(OperationalException, match=r"backtest_period_days must be positive"):
+        dk.create_fulltimerange("20220101-20220201", -1)
+
+
+def test_split_timerange(mocker, default_conf):
+    freqaiconf = freqai_conf(copy.deepcopy(default_conf))
+    freqaiconf.update({"timerange": "20220101-20220401"})
+    dk = get_patched_data_kitchen(mocker, freqaiconf)
+    tr_list, bt_list = dk.split_timerange("20220101-20220201", 30, 7)
+    assert len(tr_list) == len(bt_list) == 9
```
xmatthias: i think you may want to use separate asserts here.

The moment this test fails for the first time, you'll refactor it to that anyway - as only then pytest will show you which value caused the problem.

robcaulk: True!

tests/freqai/test_freqai.py
robcaulk: Unused code, resolving convo.

freqtrade/commands/cli_options.py
```diff
@@ -647,4 +647,14 @@ def __init__(self, *args, **kwargs):
         nargs='+',
         default=[],
     ),
+    "freqaimodel": Arg(
+        '--freqaimodel',
+        help='Specify a custom freqaimodels.',
+        metavar='NAME',
+    ),
```
xmatthias, line 654: maybe a stupid question - but especially the model is something that i assume is very tightly coupled with the strategy you're actually running. 
would it make sense to have that model be part of the strategy class (from a definition point of view)?

setup.sh
```diff
@@ -77,7 +77,15 @@ function updateenv() {
         fi
     fi
 
-    ${PYTHON} -m pip install --upgrade -r ${REQUIREMENTS} ${REQUIREMENTS_HYPEROPT} ${REQUIREMENTS_PLOT}
+    REQUIREMENTS_FREQAI=""
```
xmatthias, line 80: Note to self:
we'll need to rework this potentially - if "develop" is selected, all other questions are pointless as that installs hyperopt, plotting AND freqAI dependencies.

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,592 @@
...
+        :param:
+        unfiltered_dataframe: Full dataframe for the current backtest period.
+        dk: FreqaiDataKitchen = Data management/analysis tool assoicated to present pair only
+        :return:
+        :predictions: np.array of predictions
+        :do_predict: np.array of 1s and 0s to indicate places where freqai needed to remove
+        data (NaNs) or felt uncertain about data (i.e. SVM and/or DI index)
+        """
+
+    def make_labels(self, dataframe: DataFrame, dk: FreqaiDataKitchen) -> DataFrame:
```
xmatthias: i think this method should either be an abstract method (force implementation by children) - or should return the dataframe - or should be removed.

in it's current state - it's doing neither of the two - so if i'm not implementing this, make_labels will return None (the typehint suggests it's supposed to return the dataframe) - potentially breaking other things.

I suspect this is no longer necessary - as it doesn't seem to be used anywhere?

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,592 @@
...
+        for storing, saving, loading, and analyzing the data.
+        :params:
+        :unfiltered_dataframe: Full dataframe for the current training period
+        :metadata: pair metadata from strategy.
+        :returns:
+        :model: Trained model which can be used to inference (self.predict)
+        """
+
+    @abstractmethod
+    def fit(self) -> Any:
```
xmatthias: While the docstring suggests a parameter - it's not defined. 
All childs implement this method with parameter (at least at first glance)- so i suppose it's simply missing here?

mypy should catch this in theory - but i've seen this detection fail in practice several times already.

freqtrade/commands/cli_options.py
robcaulk: Actually, they are very much decoupled. The user could run the same strategy on CatboostPredictionModel and LightGBMPredictionModel, or even a neuralnetwork. The models are agnostic to the strategy.

There is one strategy consideration that must be known (and it is already commented into the example strategy). If the user decides to create multiple targets, they would need to use the CatboostPredictionMultiModel (or LightGBMPredictionMultiModel) if they want to get multiple targets back. 

So it seems the current structure is best.

freqtrade/resolvers/freqaimodel_resolver.py
```diff
@@ -0,0 +1,57 @@
...
+        if not freqaimodel_name:
+            raise OperationalException(
+                "No freqaimodel set. Please use `--freqaimodel` to "
+                "specify the FreqaiModel class to use.\n"
+            )
+        if freqaimodel_name in disallowed_models:
+            raise OperationalException(
+                f"{freqaimodel_name} is a baseclass and cannot be used directly. User must choose "
+                "an existing child class or inherit from this baseclass.\n"
+            )
```
xmatthias: is this necessary? creating an instance of BaseRegressionModel should fail as it's not implementing all abstract methods anyway (like - missing the .fit() method).
i didn't look at the error that comes out then though - maybe that's too confusing?


If we decide to keep it - i'd reword it to "Please choose ..." instead of "User must choose ..." - so it's better aligned to the remaining wording (like for example 2 lines above).

freqtrade/strategy/interface.py
```diff
@@ -548,6 +548,23 @@ def version(self) -> Optional[str]:
...
         return None
 
+    def populate_any_indicators(self, basepair: str, pair: str, df: DataFrame, tf: str,
+                                informative: DataFrame = None, coin: str = "",
+                                set_generalized_indicators: bool = False) -> DataFrame:
+        """
+        Function designed to automatically generate, name and merge features
+        from user indicated timeframes in the configuration file. User can add
+        additional features here, but must follow the naming convention.
+        Defined in IStrategy because Freqai needs to know it exists.
```
xmatthias: I'd add a comment here that says that this method is ONLY used / called if freqAI is enabled.
That could replace the "defined here because" ... sentence.

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1318 @@
+import copy
+import datetime
+import json
+import logging
+import pickle as pk
```
xmatthias: What's the reason / benefit of using pickle in some cases, and  `joblib.dump`/load  in different cases (within the same file)? 
i'd assume that joblib (and it's implicit cloudpickle library for pickle) is the preferred way - (mainly since we do have that available /installed alread)

freqtrade/freqai/data_drawer.py
```diff
@@ -0,0 +1,360 @@
...
+        Create or dictionary for each follower to maintain unique persistent prediction targets
+        """
+        follower_name = self.config.get("bot_name", "follower1")
+        whitelist_pairs = self.config.get("exchange", {}).get("pair_whitelist")
+
+        exists = (
+            Path(self.full_path / str("follower_dictionary-" + follower_name + ".json"))
+            .resolve()
+            .exists()
+        )
```
xmatthias: I think this can be simplified (a lot, actually).
explicitly using "str" is not necessary (it's already a string - and if you use fstring for the concatenation, you're guaranteed a string output, too).
`.resolve()` should be unnecessary - and `.exists()` should be replaced with `is_dir()` or `.is_file()` (depending if you expect this to be a directory or file).

I'd also extract the "path" generation into a separate function (or generated at class init) - the same segment (`Path(self.full_path / str("follower_dictionary-" + follower_name + ".json")`) is there at least 3 times (3 times visible at current scroll level! - twice in the same function) - extracting that will reduce code duplication - and with that also the potential for problems / typos.

freqtrade/optimize/backtesting.py
```diff
@@ -206,6 +206,11 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
         """
         self.progress.init_step(BacktestState.DATALOAD, 1)
 
+        if self.config.get('freqai') is not None:
```
xmatthias: is freqai assumed to be active/enabled the moment a freqai key is found?
there's no `freqai.enabled` - like we have it with telegram/rpc methods - so it can easily be disabled ? 

Not saying that's necessarily a problem - more a question if you will.

freqtrade/plugins/pairlist/pairlist_helpers.py
```diff
@@ -40,3 +40,15 @@ def expand_pairlist(wildcardpl: List[str], available_pairs: List[str],
...
                 raise ValueError(f"Wildcard error in {pair_wc}, {err}")
     return result
+
+
+def dynamic_expand_pairlist(config: dict, markets: list) -> List[str]:
+    if config.get('freqai', {}):
+        corr_pairlist = config['freqai']['feature_parameters']['include_corr_pairlist']
+        full_pairs = config['pairs'] + [pair for pair in corr_pairlist
+                                        if pair not in config['pairs']]
+        expanded_pairs = expand_pairlist(full_pairs, markets)
```
xmatthias: I think you will want to add `corr_pairlist` pairs **after** the `expand_pairlist()` call.

Assuming i use `include_corr_pairlist` of `.*/USDT` - which means all USDT pairs available on the exchange (now that's excessive as it'll be a few 100 pairs on binance - but other wildcard pairs can cause similar scenarios).

This sounds good in practice - but is a moving target, depending on pair listings / delistings.
In my understanding, this would then break freqai, as the number of features would be changing over time - depending on listings / delistings of the exchange

freqtrade/resolvers/freqaimodel_resolver.py
robcaulk: The default error message isn't descriptive enough for the user to understand what went wrong. This is just an attempt to guide them in the right direction.

Yeah, we should make the message more consistent with others by changing the phrasing to "Please choose."

freqtrade/optimize/backtesting.py
robcaulk: At the moment, yes, it is assumed enabled. For consistency, it makes sense to add an `enabled` parameter similar to the other config methods.

freqtrade/freqai/data_drawer.py
robcaulk: Good points - I will clean this up :-)

freqtrade/optimize/backtesting.py
xmatthias: i'm ... undecided on this.
on one hand, it would be aligned - on the other hand - it's near impossible to simply "disable" freqAI while continuing to use the same strategy.

This reminds me of another thing - "how" does `--strategy-list` work when used to compare a freqAI and non-freqAI strategy?

freqtrade/freqai/data_kitchen.py
robcaulk: pickle is used for dataframes primarily (and the PCA model because it seemed to need it for some reason). Joblib dump/load couldn't handle these objects.

If cloudpickle is already installed with joblib (I did not know about this), then it should be as easy as dropping it in:

`import cloudpickle as pk`

I will test it and make the change if it works.

freqtrade/freqai/freqai_interface.py
robcaulk: I think I deleted this function today or yesterday. It is superceded by users creating labels in their strategy by prepending `&`  to columns in their dataframe.

freqtrade/optimize/backtesting.py
robcaulk: Yes, but - the point is not necessarily to use the same strategy but to use the same config for a FreqAI strat and a non FreqAI strat. 

RE: `--strategy-list`
That is a scary question, mostly because my head isn't into that side of the code, but I suspect there are data loading conflicts if FreqAI is not the first strategy on the list (I've never tested it, for the record). I have half a mind to simply disallow its use if `freqai` is enabled.

freqtrade/freqai/data_kitchen.py
xmatthias: cloudpickle is vendored in joblib - check hyperopt.py on how to import it ... 
i actually think that the `dump` and `load` functions you import are either direct re-exports from cloudpickle - or wrappers for it.

robcaulk: Yup, it all worked. Just needed to change the protocol to `cloudpickle.DEFAULT_PROTOCOL`

docs/freqai.md
```diff
@@ -0,0 +1,687 @@
...
+| `learning_rate` | A common parameter among regressors which sets the boosting learning rate. <br> **Datatype:** float.
+| `n_jobs`, `thread_count`, `task_type` | Different libraries use different parameter names to control the number of threads used for parallel processing or whether or not it is a `task_type` of `gpu` or `cpu`. <br> **Datatype:** float.
+
+### Return values for use in strategy
+Here are the values you can expect to receive inside the dataframe returned by FreqAI:
+
+|  Parameter | Description |
+|------------|-------------|
+| `&-s*` | user defined labels in the user made strategy. Anything prepended with `&` is treated as a training target inside FreqAI. These same dataframe columns names are fed back to the user as the predictions. For example, the user wishes to predict the price change in the next 40 candles (similar to `templates/FreqaiExampleStrategy.py`) by setting `&-s_close`. FreqAI makes the predictions and gives them back to the user under the same key (`&-s_close`) to be used in `populate_entry/exit_trend()`. <br> **Datatype:** depends on the output of the model.
```
xmatthias: is the parameter `&-` - or `&` - or `&-s`? The parameter sample makes me believe that `&-s` is needed, but the description talks about `&` only ... ? 

Same with the next point i guess

docs/freqai.md
```diff
@@ -0,0 +1,687 @@
...
+|  |  **Model training parameters** 
+| `model_training_parameters` | A flexible dictionary that includes all parameters available by the user selected library. For example, if the user uses `LightGBMPredictionModel`, then this dictionary can contain any parameter available by the `LightGBMRegressor` [here](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMRegressor.html). If the user selects a different model, then this dictionary can contain any parameter from that different model.  <br> **Datatype:** dictionary.
+| `n_estimators` | A common parameter among regressors which sets the number of boosted trees to fit <br> **Datatype:** integer.
+| `learning_rate` | A common parameter among regressors which sets the boosting learning rate. <br> **Datatype:** float.
+| `n_jobs`, `thread_count`, `task_type` | Different libraries use different parameter names to control the number of threads used for parallel processing or whether or not it is a `task_type` of `gpu` or `cpu`. <br> **Datatype:** float.
+
+### Return values for use in strategy
+Here are the values you can expect to receive inside the dataframe returned by FreqAI:
+
+|  Parameter | Description |
```
xmatthias: This table seems to be missing the `%-` and `%%-` cases cases completely ... 
i understand it's supposed to define what's returned - but `%` documentation is kinda scarse and hard to find ...

docs/freqai.md
```diff
@@ -0,0 +1,687 @@
...
+freqtrade trade --strategy FreqaiExampleStrategy --config config_freqai.example.json --freqaimodel ExamplePredictionModel
+```
+
+By default, Freqai will not find find any existing models and will start by training a new one 
+given the user configuration settings. Following training, it will use that model to predict for the
+duration of `backtest_period_days`. After a full `backtest_period_days` has elapsed, Freqai will auto retrain 
+a new model, and begin making predictions with the updated model. FreqAI backtesting and live both
+permit the user to use fractional days (i.e. 0.1) in the `backtest_period_days`, which enables more frequent 
+retraining. But the user should be careful that using a fractional `backtest_period_days` with a large
+`--timerange` in backtesting will result in a huge amount of required trainings/models.
```
xmatthias: This paragraph looks/sounds wrong with the new parameter naming (could be me misunderstanding it though) - and i think should be using `live_retrain_hours` - or `train_period_days`

docs/freqai.md
```diff
@@ -0,0 +1,687 @@
...
+    "freqai": {
+        "fit_live_prediction_candles": 300,
+    }
+```
+
+If the user sets this value, FreqAI will initially use the predictions from the training data set
+and then subsequently begin introducing real prediction data as it is generated. FreqAI will save 
+this historical data to be reloaded if the user stops and restarts with the same `identifier`.
+
+<!-- ## Dynamic target expectation
```
xmatthias: ?? is this not yet implemented?

xmatthias (commented):
Just a few "docs" comments ... 
i also did some minor tweaks to wording / formatting in the docs.

docs/freqai.md
robcaulk: Yes, you are right - this needs to be updated now that we have the new parameter `live_retrain_hours`. Nice catch.

docs/freqai.md
robcaulk: It is implemented already in `fit_labels()`. The difference is that `fit_labels()` assumes a gaussian distribution in the labels a priori. I have another function, associated to this doc paragraph, which  basically just fits the labels on a bunch of different distribution types and then picks the best fitted one to provide parameters back for. That function is commented out at the end of `data_kitchen.py`.

That approach is just not quite as robust and requires yet more user understanding of an already difficult concept. Fitting a statistical distribution to the training labels and using that to understand inference prediction rarity is a HUGE leap for almost everyone who has tested FreqAI so far. Asking them to now understand quantiles and understand that the return parameters are no longer `mean` and `variance` and are instead based on an esoteric statistical distribution...it is too much for *this PR*. I will reintroduce the idea in a more concise way in the future - and we will start returning `&-s*_stat1` `&-s*_stat2` etc instead of `&-s_mean/std`. 

Further, this auto-distribution fitting approach(commented at the end of `data_kitchen.py`) relies on this library "Fitter" which is NOT a library we should/can depend on. I am working on writing our own custom auto-fitter - but that needs better testing and performance considerations.

tl;dr fitting labels with gaussian is good and working well for now. In the future, we will have our own auto-distribution fitter - but not before this merges into develop/stable.

docs/freqai.md
robcaulk: `%%` is indeed missing from the documentation. I've gone ahead and renamed the table so that we can include all "expected dataframe keys."

docs/freqai.md
robcaulk: Thanks for the proof reading ;)

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1308 @@
...
+        self.label_list: List = []
+        self.model_filename: str = ""
+        self.live = live
+        self.pair = pair
+        self.svm_model: linear_model.SGDOneClassSVM = None
+        self.keras = self.freqai_config.get("keras", False)
+        self.set_all_pairs()
+        if not self.live:
+            self.full_timerange = self.create_fulltimerange(
+                self.config["timerange"], self.freqai_config.get("train_period_days")
```
xmatthias, line 80: i think we should add some "explicit" failure around backtesting without `--timerange`?
it's failing with keyerror if that's not given - and it's not obvious why it fails to the user...

xmatthias: i think we should add some "explicit" failure around backtesting without `--timerange`?
it's failing with keyerror if that's not given - and it's not obvious why it fails to the user...

freqtrade/freqai/prediction_models/CatboostPredictionModel.py
```diff
@@ -0,0 +1,47 @@
...
+        test_data = Pool(
+            data=data_dictionary["test_features"],
+            label=data_dictionary["test_labels"],
+            weight=data_dictionary["test_weights"],
+        )
+
+        model = CatBoostRegressor(
+            allow_writing_files=False,
+            verbose=100,
+            early_stopping_rounds=400,
```
xmatthias: Shouldn't these 2 parameters be moved to the configuration - so users can define them dynamically?
the Catboost test was failing partially because of this - as it would define `verbose` twice with the default config ...

config_examples/config_freqai_futures.example.json
```diff
@@ -0,0 +1,97 @@
+{
```
xmatthias: I think it would make more sense to provide freqAI configurations as partial samples.

they should only contain the `"freqai"` key - and can then be loaded via `add_config_files` in other parameters.
The freqAI part has no (meaningful) difference between spot and futures pairs.
instead, we should provide configs based on provided models (LightGBM, Catboost) - as they require slightly different `"model_training_parameters"`.

freqtrade/freqai/prediction_models/BaseRegressionModel.py
```diff
@@ -0,0 +1,112 @@
...
+        User uses this function to add any additional return values to the dataframe.
+        e.g.
+        dataframe['volatility'] = dk.volatility_values
+        """
+
+        return dataframe
+
+    def train(
+        self, unfiltered_dataframe: DataFrame, pair: str, dk: FreqaiDataKitchen
+    ) -> Tuple[DataFrame, DataFrame]:
```
xmatthias: The return hint is clearly wrong here (we're not returning a tuple).
That obviously makes the docstring wrong as well ... (also applies to BaseTensor model)

freqtrade/freqai/data_drawer.py
```diff
@@ -0,0 +1,371 @@
...
+            return_null_array: bool = Follower could not find pair metadata
+        """
+        pair_in_dict = self.pair_dict.get(pair)
+        data_path_set = self.pair_dict.get(pair, {}).get("data_path", None)
+        return_null_array = False
+
+        if pair_in_dict:
+            model_filename = self.pair_dict[pair]["model_filename"]
+            trained_timestamp = self.pair_dict[pair]["trained_timestamp"]
+            coin_first = self.pair_dict[pair]["first"]
```
xmatthias: This segment could use some rework.
`pair_in_dict` actually contains the pair dict already (if the pair exists, obviously) - yet we're doing the dict lookup again several times below, which is ... subobtimal from a performance perspective.
i think there's two options - either you use `pair in self.pair_dict` to populate `pair_in_dict` (no need for dict lookup) and use a separte variable to cache the pair-dict - or reuse the variable within the if block.

freqtrade/freqai/data_drawer.py
```diff
@@ -0,0 +1,371 @@
...
+
+        if not data_path_set and self.follow_mode:
+            logger.warning(
+                f"Follower could not find current pair {pair} in "
+                f"pair_dictionary at path {self.full_path}, sending null values "
+                "back to strategy."
+            )
+            return_null_array = True
+
+        return model_filename, trained_timestamp, coin_first, return_null_array
```
xmatthias: `coin_first` (as well as `trained_timestamp` and `model_filename`) can be unbound (it's not set if `pair_in_dict` and `follow_mode` are both false). This can lead to very subtle crashes of "used before initialized" errors.

freqtrade/freqai/data_drawer.py
```diff
@@ -0,0 +1,371 @@
...
+        Attach the return values to the strat dataframe
+        :param dataframe: DataFrame = strategy dataframe
+        :return: DataFrame = strat dataframe with return values attached
+        """
+        df = self.model_return_values[pair]
+        to_keep = [col for col in dataframe.columns if not col.startswith("&")]
+        dataframe = pd.concat([dataframe[to_keep], df], axis=1)
+        return dataframe
+
+    def return_null_values_to_strategy(self, dataframe: DataFrame, dk) -> None:
```
xmatthias: it's quite unfortunate that Datadrawer and datakitchen produce a cyclic import if we try to add typehints here.

Having the typehints here would greatly help mypy to identify potential issues - now they're all implicit "Any" - which is obviously a "can do everything and nothing" type.

freqtrade/freqai/data_drawer.py
xmatthias: coin_first can be unbound (it's not set if `pair_in_dict` and `follow_mode` are both false). This can lead to very subtle crashes of "used before initialized" errors.

freqtrade/freqai/data_drawer.py
```diff
@@ -0,0 +1,589 @@
...
+        """
+        # dynamic df returned to strategy and plotted in frequi
+        mrv_df = self.model_return_values[pair] = pd.DataFrame()
+
+        for label in dk.label_list:
+            mrv_df[label] = pred_df[label]
+            mrv_df[f"{label}_mean"] = dk.data["labels_mean"][label]
+            mrv_df[f"{label}_std"] = dk.data["labels_std"][label]
+
+        if self.freqai_info.get("feature_parameters", {}).get("DI_threshold", 0) > 0:
```
xmatthias: I think you can safely assume `feature_parameters` to be present based on the schema definition - and as such, access it via `freqai_info['feature_parameters']`.
For obvious reasons, that's a faster call than `.get()` ... which is unnecessary here (at last in thee first segment.)

obviously also applies below / elsewhere ...

freqtrade/freqai/data_drawer.py
```diff
@@ -0,0 +1,589 @@
...
+            else:
+                delete_dict[coin]["num_folders"] += 1
+                delete_dict[coin]["timestamps"][int(timestamp)] = dir
+
+        for coin in delete_dict:
+            if delete_dict[coin]["num_folders"] > 2:
+                sorted_dict = collections.OrderedDict(
+                    sorted(delete_dict[coin]["timestamps"].items())
+                )
+                num_delete = len(sorted_dict) - 2
```
xmatthias, line 362: wouldn't a time-based approach be better / more reliable?
Currently i think this works by removing all files (sorted by timestamp) except for the last 2 - but in reality, you'll know old a model is allowed to be.


How will this impact startup? assume i have set my bot to retrain every hour - will i still get chart predictions for 6 hours ago? (that's 6 models ago).
How about after a restart of the bot (for whatever reason)?

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1112 @@
...
+            logger.warning(
+                "SVM outlier removal not currently supported for Keras based models. "
+                "Skipping user requested function."
+            )
+            if predict:
+                self.do_predict = np.ones(len(self.data_dictionary["prediction_features"]))
+            return
+
+        if predict:
+            assert self.svm_model, "No svm model available for outlier removal"
```
xmatthias: Please don't assert in "non-test" code - asserts can be deactivated via a CLI flag of python ... so you can't rely on this to work on all cases.

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,572 @@
...
+        self.follow_mode = self.freqai_info.get("follow_mode", False)
+        self.dd = FreqaiDataDrawer(Path(self.full_path), self.config, self.follow_mode)
+        self.lock = threading.Lock()
+        self.identifier = self.freqai_info.get("identifier", "no_id_provided")
+        self.scanning = False
+        self.keras = self.freqai_info.get("keras", False)
+        if self.keras and self.freqai_info.get("feature_parameters", {}).get("DI_threshold", 0):
+            self.freqai_info["feature_parameters"]["DI_threshold"] = 0
+            logger.warning("DI threshold is not configured for Keras models yet. Deactivating.")
+        self.CONV_WIDTH = self.freqai_info.get("conv_width", 2)
```
xmatthias, line 81: This parameter (as well as keras) is neither documented, nor part of the schema.

The 2nd part is not necessarily a problem (but means you must type-check it before usage) - but yet i'd refer that to the schema.

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,572 @@
...
+    @threaded
+    def start_scanning(self, strategy: IStrategy) -> None:
+        """
+        Function designed to constantly scan pairs for retraining on a separate thread (intracandle)
+        to improve model youth. This function is agnostic to data preparation/collection/storage,
+        it simply trains on what ever data is available in the self.dd.
+        :param strategy: IStrategy = The user defined strategy class
+        """
+        while 1:
+            time.sleep(1)
```
xmatthias, line 145: What's the reason to sleep a whole second here?

freqtrade/freqai/data_kitchen.py
```diff
@@ -85,9 +85,10 @@ def __init__(
                 config["freqai"]["backtest_period_days"],
             )
 
-        db_url = self.config.get('db_url', None)
-        self.database_path = '' if db_url == 'sqlite://' else str(db_url).split('///')[1]
-        self.trade_database_df: DataFrame = pd.DataFrame()
+        if self.live:
+            db_url = self.config.get('db_url', 'sqlite://')
+            self.database_path = '' if db_url == 'sqlite://' else str(db_url).split('///')[1]
```
xmatthias: It's unsafe to assume tripple `/` in the dburl.
That's only valid for sqlite url's.
Using postgres or mysql will get you a keyerror on the `[1]` - as the split won't yield a 2nd element.

freqtrade/optimize/backtesting.py
```diff
@@ -207,6 +207,11 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
...
+            self.required_startup += int(self.config.get('freqai', {}).get('startup_candles', 1000))
```
xmatthias: Wouldn't defaulting to 0 here make more sense?

if it's not set - we should assume that the value is 0 - not 1000 ...

freqtrade/freqai/data_kitchen.py
xmatthias: Apart from that it's also breaking CI - as database_path is no longer assigned - and therefore a "use before definition" error.

freqtrade/freqai/data_kitchen.py
```diff
@@ -87,9 +87,13 @@ def __init__(
...
 
         if self.live:
-            db_url = self.config.get('db_url', 'sqlite://')
-            self.database_path = '' if db_url == 'sqlite://' else str(db_url).split('///')[1]
-            self.trade_database_df: DataFrame = pd.DataFrame()
+            db_url = self.config.get('db_url', 'None')
+            self.database_path = Path(db_url)
+            self.database_name = self.database_path.parts[-1]
+        else:
+            self.database_path = Path('None')
```
xmatthias: a better approach (especially considering the check you need to do) would be to assign  the following  - but before the `if`.
That'd enable mypy to tell you where you need to check if the database_path is not set properly.

```
        self.database_path: Optional[Path] = None
```

freqtrade/freqai/prediction_models/CatboostClassifier.py
```diff
@@ -0,0 +1,44 @@
...
+
+from catboost import CatBoostClassifier, Pool
+
+from freqtrade.freqai.prediction_models.BaseRegressionModel import BaseRegressionModel
+
+
+logger = logging.getLogger(__name__)
+
+
+class CatboostClassifier(BaseRegressionModel):
```
xmatthias: Is there a LightGBM counterpart, too? (just asking - it's not a requirement)

freqtrade/freqai/freqai_interface.py
```diff
@@ -0,0 +1,629 @@
...
+        ):
+            dk.principal_component_analysis()
+
+        if self.freqai_info["feature_parameters"].get("use_SVM_to_remove_outliers", False):
+            dk.use_SVM_to_remove_outliers(predict=False)
+
+        if self.freqai_info["feature_parameters"].get("DI_threshold", 0):
+            dk.data["avg_mean_dist"] = dk.compute_distances()
+
+        if self.freqai_info["feature_parameters"].get("use_DBSCAN_to_remove_outliers", False):
```
xmatthias, line 403: Should this parameter be in the example config, too?

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1215 @@
...
+        ]
+        return dataframe[to_keep]
+
+    def get_current_trade_database(self) -> None:
+
+        if self.database_path is None:
+            logger.warning('No trade database found. Skipping analysis.')
+            return
+
+        data = sqlite3.connect(self.database_name)
```
xmatthias: Freqtrade does support other database engines as well (postgres, mariadb).
using sqlite3 will therefore fail in such instances (this is therefore introducing a bug for such instances - especially as no check for "is this a sqlite url" is in place).
so you either need to account for these instances (careful as psycopg2 and pymysql are not required by default, so imports may fail) - or handle this differently (which is what i'd suggest).

You should be able to use something around `Trade._session.bind` - which you can pass to `pd.from_sql()` - and which will reuse the existing connection.
Obviously it'll not work for backtesting - but that'll not work with a database anyway i'd think.

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1215 @@
...
+    def split_timerange(
+        self, tr: str, train_split: int = 28, bt_split: int = 7
+    ) -> Tuple[list, list]:
+        """
+        Function which takes a single time range (tr) and splits it
+        into sub timeranges to train and backtest on based on user input
+        tr: str, full timerange to train on
+        train_split: the period length for the each training (days). Specified in user
+        configuration file
+        bt_split: the backtesting length (dats). Specified in user configuration file
```
xmatthias: what's `dats` ? is it supposed to be data? (heavy nitpick!)

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1215 @@
...
+    """
+
+    def __init__(
+        self,
+        config: Dict[str, Any],
+        live: bool = False,
+        pair: str = "",
+    ):
+        self.data: Dict[Any, Any] = {}
+        self.data_dictionary: Dict[Any, Any] = {}
```
xmatthias: Is there a way to better type this?
data_dictionary seems to have a limited set of keys ... so i think this could be typed via `TypedDict` [docs](https://docs.python.org/3/library/typing.html#typing.TypedDict).
Tthat'd enable mypy (and also editors) to properly help us detect bugs ... 
I know we're not using this elsewhere - but i think it would help us a lot in this case - where "any/any" is used to type the dictionary...

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1215 @@
...
+        elif max_timeframe_chars[-1] == "m":
+            if len(max_timeframe_chars) == 2:
+                additional_seconds = max_period * 60 * int(max_timeframe_chars[-2])
+            elif len(max_timeframe_chars) == 3:
+                additional_seconds = max_period * 60 * int(float(max_timeframe_chars[0:2]))
+            else:
+                logger.warning(
+                    "FreqAI could not detect max timeframe and therefore may not "
+                    "download the proper amount of data for training"
+                )
```
xmatthias: I'm not sure what this does - but i don't think it's correct.

the code is using the "last" value of "include_timeframes" - which is a list.
so the value could be '1h', '15m', or anything else (it's also not guaranteed that the list is sorted properly and the highest value is "highest"), so i can break the logic by using the following list (in this sequence): 

```
            "include_timeframes": [
                "3m",
                "15m",
                "1h",
                "5m"
            ],
```
(i just added 5m at the end, as that's the one i added last ... )

assuming i understood the intend correctly, i think you should loop this list - and use `timeframe_to_seconds()` to convert it to seconds - picking the max from these - and then do whatever transformation is needed.

freqtrade/freqai/data_drawer.py
```diff
@@ -0,0 +1,642 @@
...
+        """
+        with open(self.historic_predictions_path, "wb") as fp:
+            cloudpickle.dump(self.historic_predictions, fp, protocol=cloudpickle.DEFAULT_PROTOCOL)
+
+    def save_drawer_to_disk(self):
+        """
+        Save data drawer full of all pair model metadata in present model folder.
+        """
+        with open(self.pair_dictionary_path, "w") as fp:
+            json.dump(self.pair_dict, fp, default=self.np_encoder)
```
xmatthias: I assume this is running freqently / has a lot of data passing through?
In which case, i'd suggest using rapidjson.dump for it (see `misc.py`)- it's a LOT faster - and i think can handle numeric data (like numpy) semi-automatically if set properly.

freqtrade/freqai/data_kitchen.py
xmatthias: @robcaulk i don't really see this as resolved if i'm honest.

It's still not supporting other engines - and it's also not reusing available methods (there's a method to load trade data from DB - as well as a (threaded) session that exists for the lifetime of the bot.

Assuming you see this as temporary workaround (a workaround i clearly don't like):
Also error handling around this is clearly not sufficient - if something between `.connect()` and `.close()` fails - you'll keep the session open, potentially causing a lock on the database (sqlite is pretty prone to locking problems ...).

freqtrade/freqai/data_kitchen.py
```diff
@@ -0,0 +1,1205 @@
...
+            return
+
+        data = sqlite3.connect(self.database_name)
+        query = data.execute("SELECT * From trades")
+        cols = [column[0] for column in query.description]
+        df = pd.DataFrame.from_records(data=query.fetchall(), columns=cols)
+        self.trade_database_df = df.dropna(subset='close_date')
+        data.close()
+
+    def np_encoder(self, object):
```
xmatthias: I don't think this method is used (not from datakitchen, anyway).

so it can either be removed - or refactored into a "non-class" method - so it can be used wherever needed.

freqtrade/freqai/freqai_interface.py
```diff
@@ -579,6 +588,28 @@ def fit_live_predictions(self, dk: FreqaiDataKitchen, pair: str) -> None:
 
         return
 
+    def inference_timer(self, do='start'):
```
xmatthias, line 593: I think this would be something that might be moved someplace else in the future.

"slow" strategies can have the same effect, with the same risk.
That said - i'm not sure if we really need this - setting `ignore_buying_expired_candle_after` can already do this automatically - automatically ignoring "late" entries.

xmatthias (approved):
Great job / contribution! :100: 

LGTM :+1: 

I assume this will be a first version, which will still be gradually improved.
Let's hope we don't have to do too many breaking changes from now on - as it'll impact more people.

one wish for the future would be to improve test coverage of the freqAI module - while it has some tests covering certain areas - it's not well covered overall - which could become a problem down the line.

PR #6851, opened 2022-05-16
BT: Reporting canceled trade entries

eSeR1805:
## Summary

Backtest report to include trade entry cancelations count.

Solve the issue: No issue

## Quick changelog

- Track trade entry cancelations
- Report trade entry cancelation
- Update testcases that care about BT report

## What's new?

The goal is to have the backtest report include information on the number of trade entries that were canceled due to user request through `adjust_entry_price`. If for the first order in a Trade user returns `None` in the aforementioned callback then the trade will get canceled completely.

*Note: Since it's a trivial change don't think any extra testcases are needed for this but I'd like your feedback on it.


xmatthias (approved):
Great :100: 

Nice improvement, LGTM :+1:

PR #6854, opened 2022-05-17
BT: Reporting canceled/replaced entry orders

eSeR1805:
## Summary

Further expanding on PR:https://github.com/freqtrade/freqtrade/pull/6851. Report canceled/replaced orders also

Solve the issue: No related issue.

## Quick changelog

Update testcases that care about BT report
- Track entry order cancelations/replacements
- Report entry order cancelation/replacements
- Update reporting section to only show when order adjustment is used
- Update testcases that care about BT report
- Update related documentation

## What's new?

BT report to also show information on entry order cancelations/replacements. BT report to include this section only when `adjust_entry_price` is used (this is determined by seeing at least one order cancelation).
Samples with and without this extra info in the report:
image
image


xmatthias (approved):
LGTM :+1:

PR #6857, opened 2022-05-17
Add support for fudging unavailable funding rates, allowing backtesti…

froggleston:
…ng of timeranges where futures candles are available, but rates are not

## Summary

This is to allow backtesting of longer historical timeranges when using futures data. It is often the case that futures candles and marks are available from exchanges, but the funding rates are not. This breaks backtesting for anything but the most recent months. This PR introduces the `futures_funding_rate` config option that will substitute a user-specified value to allow backtesting to work. Realistically, this should be set to 0.

## Quick changelog

- Change history_utils.py to support the case that if hist.empty isn't true, allow creation of an empty dataframe to hold funding rates
- Change backtesting.py to detect when no rates are available and populate the empty dataframe with the user-specified value
- amend docs

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -230,6 +230,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `dataformat_trades` | Data format to use to store historical trades data. <br> *Defaults to `jsongz`*. <br> **Datatype:** String
 | `position_adjustment_enable` | Enables the strategy to use position adjustments (additional buys or sells). [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*<br> **Datatype:** Boolean
 | `max_entry_position_adjustment` | Maximum additional order(s) for each open trade on top of the first entry Order. Set it to `-1` for unlimited additional orders. [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `-1`.*<br> **Datatype:** Positive Integer or -1
+| `futures_funding_rate` | User-specified funding rate to be used when historical funding rates are not available from the exchange. This does not overwrite real historical rates. It is recommended that this be set to 0 unless you are testing a specific coin and you understand how the funding rate will affect freqtrade's profit calculations. [More information here](configuration.md) <br>*Defaults to None.*<br> **Datatype:** Float
 
 ### Parameters in the strategy
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -230,7 +230,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `dataformat_trades` | Data format to use to store historical trades data. <br> *Defaults to `jsongz`*. <br> **Datatype:** String
 | `position_adjustment_enable` | Enables the strategy to use position adjustments (additional buys or sells). [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*<br> **Datatype:** Boolean
 | `max_entry_position_adjustment` | Maximum additional order(s) for each open trade on top of the first entry Order. Set it to `-1` for unlimited additional orders. [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `-1`.*<br> **Datatype:** Positive Integer or -1
-| `futures_funding_rate` | User-specified funding rate to be used when historical funding rates are not available from the exchange. This does not overwrite real historical rates. It is recommended that this be set to 0 unless you are testing a specific coin and you understand how the funding rate will affect freqtrade's profit calculations. [More information here](configuration.md) <br>*Defaults to None.*<br> **Datatype:** Float
+| `futures_funding_rate` | User-specified funding rate to be used when historical funding rates are not available from the exchange. This does not overwrite real historical rates. It is recommended that this be set to 0 unless you are testing a specific coin and you understand how the funding rate will affect freqtrade's profit calculations. [More information here](leverage.md) <br>*Defaults to None.*<br> **Datatype:** Float
 
 ### Parameters in the strategy
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -230,7 +230,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `dataformat_trades` | Data format to use to store historical trades data. <br> *Defaults to `jsongz`*. <br> **Datatype:** String
 | `position_adjustment_enable` | Enables the strategy to use position adjustments (additional buys or sells). [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*<br> **Datatype:** Boolean
 | `max_entry_position_adjustment` | Maximum additional order(s) for each open trade on top of the first entry Order. Set it to `-1` for unlimited additional orders. [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `-1`.*<br> **Datatype:** Positive Integer or -1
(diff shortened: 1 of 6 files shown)


docs/configuration.md
```diff
@@ -230,6 +230,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
 | `position_adjustment_enable` | Enables the strategy to use position adjustments (additional buys or sells). [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `false`.*<br> **Datatype:** Boolean
 | `max_entry_position_adjustment` | Maximum additional order(s) for each open trade on top of the first entry Order. Set it to `-1` for unlimited additional orders. [More information here](strategy-callbacks.md#adjust-trade-position). <br> [Strategy Override](#parameters-in-the-strategy). <br>*Defaults to `-1`.*<br> **Datatype:** Positive Integer or -1
+| `futures_funding_rate` | User-specified funding rate to be used when historical funding rates are not available from the exchange. This does not overwrite real historical rates. It is recommended that this be set to 0 unless you are testing a specific coin and you understand how the funding rate will affect freqtrade's profit calculations. [More information here](configuration.md) <br>*Defaults to None.*<br> **Datatype:** Float
```
xmatthias: that "more info here" link is wrong if you ask me - it PROBABLY would like to point to the leverage.md page .. ?

freqtrade/optimize/backtesting.py
```diff
@@ -275,8 +275,28 @@ def load_bt_data_detail(self) -> None:
...
+                                 'high': 'high_mark',
+                                 'low': 'low_mark',
+                                 'volume': 'volume_mark'},
+                        inplace=True)
+
+                    self.futures_data[pair] = mark_rates_dict[pair]
+                else:
+                    self.futures_data[pair] = mark_rates_dict[pair].merge(
+                        funding_rates_dict[pair], on='date',
+                        how="inner", suffixes=["_fund", "_mark"])
```
xmatthias: I'm not sure this covers what you're actually after.

Assuming you're using OKX - which only provides funding rates for the last few months:
* this would work if you backtest with `timerange 20220101-20220301` (as the timerange stops before the funding data starts)
* if you backtest `20220101-20220501` - then funding rates dataframe will not be empty - but you'll have the "last few months" (i guess from April forward - so you may have 1000 mark candles, and 300 funding rate candles). that'll empty the mark rates for the period from 20220101-20220401 (it's an inner join) - and be "more wrong" than using the user-defined rate for for the period where "data is not available".

froggleston: FTX right now, but the same principle applies.

- Yes, correct, and any pairs within a longer timerange where empty funding rate df exists.
- Yes, true. I'll try to address this. Any pointers before I start?

docs/configuration.md
froggleston: Ah yes, my mistake

xmatthias: that's why we do reviews ;)

froggleston:
I've implemented an outer join to merge the full df with the user specified funding rate: 

image

froggleston:
Ah i've noticed a problem - the mark and rate are switched. 2 secs.

froggleston:
empty rate df:

image

outer merged df:

image

froggleston:
Not sure why the test is failing....

xmatthias:
> Not sure why the test is failing....

This is a fluky test - it fails randomly and i have no idea why

froggleston:
Yeah, I couldn't see anything obvious

freqtrade/optimize/backtesting.py
```diff
@@ -275,8 +275,34 @@ def load_bt_data_detail(self) -> None:
...
+                                 'volume': 'volume_mark'},
+                        inplace=True)
+
+                    self.futures_data[pair] = mark_rates_dict[pair]
+                else:
+                    if "futures_funding_rate" in self.config:
+                        self.futures_data[pair] = mark_rates_dict[pair].merge(
+                            funding_rates_dict[pair], on='date',
+                            how="outer", suffixes=["_mark", "_fund"]).fillna(
+                                self.config.get('futures_funding_rate'))
```
xmatthias: is it intentional that you fill "everything" missing with `futures_funding_rate`?

assuming you have some mark candles missing (but funding-rate rows are there for whatever reason) - this would fill the mark columns with it as well.

while "funding_rate" is oscillating around 0 (usually) - mark prices are in "candle price" - so for btc, somewhere around 30k at the moment - so this could heavily taint results

freqtrade/optimize/backtesting.py
```diff
@@ -275,8 +275,34 @@ def load_bt_data_detail(self) -> None:
...
-                    mark_rates_dict[pair], on='date', how="inner", suffixes=["_fund", "_mark"])
+
+                if (pair in mark_rates_dict
+                        and len(funding_rates_dict[pair]) == 0
+                        and "futures_funding_rate" in self.config):
+                    mark_rates_dict[pair]["open_fund"] = self.config.get('futures_funding_rate')
+                    mark_rates_dict[pair]["close_fund"] = 0.0
+                    mark_rates_dict[pair]["high_fund"] = 0.0
+                    mark_rates_dict[pair]["low_fund"] = 0.0
+                    mark_rates_dict[pair]["volume_fund"] = 0.0
```
xmatthias: You can safely ignore these columns - only `open_fund` and `open_mark` are actually relevant to this dataframe.

froggleston: OK we can remove those - i thought they might be necessary for something downstream so left them in

freqtrade/optimize/backtesting.py
froggleston: Ah I meant to only fill funding rate. I'll adjust that to the series instead of the dataframe

froggleston: Done

freqtrade/optimize/backtesting.py
froggleston: And done

freqtrade/optimize/backtesting.py
```diff
@@ -275,8 +275,30 @@ def load_bt_data_detail(self) -> None:
...
+                                 'volume': 'volume_mark'},
+                        inplace=True)
+
+                    self.futures_data[pair] = mark_rates_dict[pair]
+                else:
+                    if "futures_funding_rate" in self.config:
+                        self.futures_data[pair] = mark_rates_dict[pair].merge(
+                            funding_rates_dict[pair], on='date',
+                            how="outer", suffixes=["_mark", "_fund"])['open_fund'].fillna(
+                                self.config.get('futures_funding_rate'))
```
xmatthias: This is ... actually wrong.

it's selecting a series - and then assigns the series to `futures_data`.
I'm refactoring this a bit now to move this to the exchange class so all the "combine" logic is in one place - you'll see what i mean. That'll also allow tests on it - which i think is quite important (this is what identified this bug in the first place) :laughing: 

I'd appreciate if you could give a look at the changes i made - maybe have a run with "your" scenario, too - to see if this is what you're after.

xmatthias: This is ... actually wrong.

it's selecting a series - and then assigns the series to `futures_data`.
I'm refactoring this a bit now to move this to the exchange class so all the "combine" logic is in one place - you'll see what i mean. That'll also allow tests on it - which i think is quite important (this is what identified this bug in the first place) :laughing: 

I'd appreciate if you could give a look at the changes i made - maybe have a run with your scenario (where you've encountered this problem), too - to see if this is what you're after (tests should also show that - but can never replace a real test - and maybe my data is slightly different than yours).

freqtrade/optimize/backtesting.py
xmatthias: no - good job ;)

xmatthias: no - good job ;)

freqtrade/optimize/backtesting.py
froggleston: OK awesome - let me have a look later. Thanks!

froggleston: This looks fine to me - i don't see any issues yet my end...

froggleston:
Your changes look good to me - I don't see any issues my end as yet. 👍

xmatthias (approved):
LGTM :+1:

PR #6858, opened 2022-05-17
Removed "is not None"

SmartManoj:
## Summary

[For Performance](https://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none#comment9527936_7816439)

freqtrade\configuration\configuration.py:461 reduced checks

freqtrade\persistence\trade_model.py:fee_updated - reduced code


xmatthias:
checking explicitly for `is not None` is usually done intentionally - and you'll need to be 110% sure that the variable you're checking is not "Falsy" (`0`, `[]`, `''`) to avoid unexpected behaviour. 

As tests agree with me - this is an invalid change we will not be accepting.

SmartManoj:
> that the variable you're checking is not "Falsy"

That's why it was in the draft stage.

PR #6859, opened 2022-05-17
Removed None in dict.get()

SmartManoj:
## Summary
https://stackoverflow.com/a/12631641

Extra Changes: 
freqtrade\freqtradebot.py:70
freqtrade\plugins\pairlistmanager.py:31


xmatthias:
I'm not sure what your aim with this is. 

It's clear that using `.get('', None)` is not _necessary_ - but it's better to be explicit than implicit, and using an expicit `None` makes it clear that the result will be None - and not something else in a type someone might guess.

This is neither providing a performance benefit - nor any other benefit i can see.

Therefore (to not close this issue outright) - please point out why you think this is an improvement.
i'll not waste my time reviewing a PR that's decreasing readability, just to boost someones github stats (i assume that's the reason for doing such pointless PR's).

SmartManoj:
>  it's better to be explicit than implicit

Not in this case, as "It is superfluous in this case." which makes the default value of that argument pointless.

>  someone might guess.

Violates "In the face of ambiguity, refuse the temptation to guess."

> performance benefit 
```
Python 3.11.0a7 (main, Apr  5 2022, 21:27:39) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import dis
>>> a = {}
>>> dis.dis("a.get(2)")
  1           0 LOAD_NAME                0 (a)
              2 LOAD_METHOD              1 (get)
              4 LOAD_CONST               0 (2)
              6 CALL_METHOD              1
              8 RETURN_VALUE
>>> dis.dis("a.get(2, None)")
  1           0 LOAD_NAME                0 (a)
              2 LOAD_METHOD              1 (get)
              4 LOAD_CONST               0 (2)
              6 LOAD_CONST               1 (None)
              8 CALL_METHOD              2
             10 RETURN_VALUE
```
> decreasing readability

Nope. If the second argument is present, time taken for human to process that info and finding that info is not useful is waste.


Also I don't see any example for this Zen. And found [Explicit isn’t better than implicit](https://mina86.com/2021/explicit-isnt-better-than-implicit/)


> just to boost someones github stats

https://github.com/isaacs/github/issues/627
Why should a sane person do that?

froggleston:
Violates "In the face of ambiguity, refuse the temptation to guess."

This is actually contradicting what you are suggesting. By removing the explicit reference to None, you are adding temptation to guess.

The performance "gains" from this PR will be negligible compared to other processes.

SmartManoj:
> you are adding temptation to guess.

There is nothing to guess. It returns `None` if not found. One should understand the schematics of the language before reading that. What is the point of default arguments?

froggleston:
To provide explicit references to what is default, including None.

SmartManoj:
> What is the point of default arguments?

Isn't that there is no need to pass the default value to that argument?

>  The performance "gains" from this PR will be negligible compared to other processes.

Even speeding up a millisecond is important in crypto trading.
Else it would violate the second virtue of the [three virtues of a great programmer](https://thethreevirtues.com).

SmartManoj:
```
def order(price, amount):
	pass
```

`order(price=1, amount=1)` is better than `order(1,1)`. 

I think this is the right example for the Zen "Explicit is better than implicit."

xmatthias (approved):
In some instances, it's irrelevant if we use `, None` explicitly or implicitly.

Especially when directly comparing with None though i prefer to have it explicit, and I've therefore reverted a few instances.
The performance gain is really not worth it - as it's marginal (it's one additional AST call), and you'll not feel it (like - not at all) in any operation (you'd not feel this difference if this were running 10.000 times in a loop).

PR #6874, opened 2022-05-22
Buy reasons

froggleston:
## Summary

Integrate buy_reasons into FT with a CLI command

## Quick changelog

- Amend init, cli options, arguments, and configuration
- Add new entryexitanalysis.py

## What's new?

Following on from the generation of signal candles from FT with `--export=signals`, this PR adds the `freqtrade analysis` cli command, where the command needs a config, a strategy, and various options to tune the output:

`freqtrade analysis -c config.json --strategy SampleStrategy --indicator_list rsi,profit_abs --analysis_groups 0,1,2,3`

NB still needs tests


froggleston:
I also need to fix the first tabular output - it misses out info that I think it something to do with the merging.

freqtrade/commands/analyze_commands.py
```diff
@@ -0,0 +1,60 @@
...
+    config = setup_utils_configuration(args, method)
+
+    no_unlimited_runmodes = {
+        RunMode.BACKTEST: 'backtesting',
+    }
+    if method in no_unlimited_runmodes.keys():
+        from freqtrade.data.btanalysis import get_latest_backtest_filename
+
+        btfile = get_latest_backtest_filename(config['user_data_dir'] / 'backtest_results')
+        signals_file = f"{os.path.basename(os.path.splitext(btfile)[0])}_signals.pkl"
```
xmatthias: Please use `pathlib.Path` instead of `os.path` - it's ... more resiliant for multi-os scenarios (windows in particular)

freqtrade/commands/analyze_commands.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if (not os.path.exists(config['user_data_dir'] / 'backtest_results' / signals_file)):
```
xmatthias:
```suggestion
        if (not (config['user_data_dir'] / 'backtest_results' / signals_file).exists()):
```

froggleston:
Looks fine - just the kucoin test problem again. I'll work on some tests now.

froggleston:
Some random tests failing again. Everything seems to be done though I think.

tests/data/test_entryexitanalysis.py
```diff
@@ -0,0 +1,94 @@
...
+        '--datadir', str(testdatadir),
+        '--analysis_groups', '0',
+        '--strategy',
+        'StrategyTestV3Analysis',
+    ]
+    args = get_args(args)
+    start_analysis_entries_exits(args)
+
+    captured = capsys.readouterr()
+    assert 'enter_tag_long' in captured.out
```
xmatthias: That last CI failure was most likely caused by this new test (even though the failure was in a different place).

Please either use the `backtesting_cleanup` fixture (imported from test_backtesting) - or call `Backtesting.cleanup()` manually after the test to ensure backtesting is cleaned up properly.

froggleston:
Oh Kucoin.

tests/data/test_entryexitanalysis.py
```diff
@@ -0,0 +1,110 @@
...
+    default_conf.update({
+        "use_exit_signal": True,
+        "exit_profit_only": False,
+        "exit_profit_offset": 0.0,
+        "ignore_roi_if_entry_signal": False,
+        'analysis_groups': "0",
+        'enter_reason_list': "all",
+        'exit_reason_list': "all",
+        'indicator_list': "rsi"
+    })
```
xmatthias, line 29: looking at the coverage - we should either have multiple tests (with different analysis_groups) - or provide more groups in this test (or parametrize it to use different combinations - but i suspect then asserting may become difficult - but maybe it's not? ).

freqtrade/commands/arguments.py
```diff
@@ -415,3 +419,9 @@ def _build_subcommands(self) -> None:
                                               parents=[_common_parser])
         webserver_cmd.set_defaults(func=start_webserver)
         self._build_args(optionlist=ARGS_WEBSERVER, parser=webserver_cmd)
+
+        # Add backtesting analysis subcommand
+        analysis_cmd = subparsers.add_parser('analysis', help='Analysis module.',
```
xmatthias: I'm not entirely sure i like the "analysis" name for the command.
it's pretty generic - while it's really only analyzing backtest results - so maybe `backtesting-analysis` would be more appropriate, and less missleading?

freqtrade/commands/cli_options.py
```diff
@@ -614,4 +614,36 @@ def __init__(self, *args, **kwargs):
               "that do not contain any parameters."),
         action="store_true",
     ),
+    "analysis_groups": Arg(
+        "--analysis_groups",
```
xmatthias: Please use dash (`-`) for arguments (so the above would become `--analysis-groups`).
That'll keep commandline-arguments aligned throughout freqtrade - otherwise they're too easy to miss-type (if you need to also remember _ or -.)

freqtrade/data/entryexitanalysis.py
```diff
@@ -0,0 +1,259 @@
...
+                available_inds = []
+                for ind in indicator_list.split(","):
+                    if ind in bigdf:
+                        available_inds.append(ind)
+                ilist = ["pair", "enter_reason", "exit_reason"] + available_inds
+                _print_table(bigdf[ilist], sortcols=['exit_reason'], show_index=False)
+        else:
+            _print_table(bigdf[columns], sortcols=['pair'], show_index=False)
+    else:
+        print("\\_ No trades to show")
```
xmatthias, line 185: I seem to run into this consistently (i did run backtesting with `--export signals`).

the output is not helpful in determining what is wrong (i did use buy and sell reasons).

```
2022-05-26 19:28:59,221 - freqtrade.data.btanalysis - INFO - Loading backtest result from ../xmatt/backtest_results/backtest-result-2022-05-26_19-26-07.json
2022-05-26 19:28:59,228 - freqtrade.data.entryexitanalysis - INFO - Loaded signal candles: ../xmatt/backtest_results/backtest-result-2022-05-26_19-26-07_signals.pkl
2022-05-26 19:28:59,228 - freqtrade.data.entryexitanalysis - INFO - Processing None : 2 pairs
\_ No trades to show

```

**edit** it was the missing `--strategy` argument - see below ... but output is clearly not helpful :laughing:

freqtrade/data/entryexitanalysis.py
```diff
@@ -0,0 +1,259 @@
...
+        logger.info(f"Processing {strategy_name} : {len(pairlist)} pairs")
+
+        for pair in pairlist:
+            if pair in signal_candles[strategy_name]:
+                analysed_trades_dict[strategy_name][pair] = _analyze_candles_and_indicators(
+                                                              pair,
+                                                              trades,
+                                                              signal_candles[strategy_name][pair])
+    except Exception:
+        pass
```
xmatthias: I don't think we should swallow all exceptions blindly.
if we expect certain exceptions (in my case it's a keyerror due to a missing `-s` argument) - then it should also be shown unless we expected that error beforehand.

freqtrade/commands/analyze_commands.py
```diff
@@ -0,0 +1,64 @@
...
+    from freqtrade.data.entryexitanalysis import process_entry_exit_reasons
+
+    # Initialize configuration
+    config = setup_analyze_configuration(args, RunMode.BACKTEST)
+
+    logger.info('Starting freqtrade in analysis mode')
+
+    process_entry_exit_reasons(Path(config['user_data_dir'], 'backtest_results'),
+                               config['exchange']['pair_whitelist'],
+                               config['strategy'],
```
xmatthias: I'm not sure about this - you're probably better of passing the whole config into this method - and then handle it there.

For example (what i ran into):
i did a backtest of one strategy. 
so there is really no need to use `--strategy` on this command again as the backtest-result can determine that only one strategy is available - and use that strategy for the output - simplifying the user process for the user.

freqtrade/data/entryexitanalysis.py
```diff
@@ -0,0 +1,259 @@
...
+def process_entry_exit_reasons(backtest_dir: Path,
+                               pairlist: List[str],
+                               strategy_name: str,
+                               analysis_groups: Optional[str] = "0,1,2",
+                               enter_reason_list: Optional[str] = "all",
+                               exit_reason_list: Optional[str] = "all",
+                               indicator_list: Optional[str] = None):
+
+    try:
+        trades = load_backtest_data(backtest_dir, strategy_name)
```
xmatthias: Consider using load_backtest_stats here (first) - that way you can derive the strategy.

If we assume users MUST insert a strategy for this command explicitly - we should check for that.

freqtrade/data/entryexitanalysis.py
```diff
@@ -0,0 +1,255 @@
...
+                    trades_red = pd.merge(trades_red, trades_inds, on='signal_date', how='outer')
+                except Exception as e:
+                    print(e)
+        return trades_red
+    else:
+        return pd.DataFrame()
+
+
+def _do_group_table_output(bigdf, glist):
+    if "0" in glist:
```
xmatthias: i'm not against this naming (0, 1, 2) in general - but please put a comment here as well for each option - otherwise we have to "jump" to the cli page to actually see what `0` does ...

freqtrade/data/entryexitanalysis.py
```diff
@@ -0,0 +1,255 @@
...
+                       'profit_tot', 'wl_ratio_pct', 'avg_win', 'avg_loss']
+
+        sortcols = ['total_num_buys']
+
+        _print_table(new, sortcols, show_index=True)
+    if "1" in glist:
+        new = bigdf.groupby(['enter_reason']) \
+                   .agg({'profit_abs': ['count', 'sum', 'median', 'mean'],
+                         'profit_ratio': ['sum', 'median', 'mean']}
+                        ).reset_index()
```
xmatthias: At a quick glance, it seems like all but "0" could be refactored / combined as 90% of the code (except for the groupby) is identical.
using the group-by as argument or variable (e.g. `x = ['enter_reason']`) - using that as groupby (`...groupby(x).`) - as well as for the columns `new.columns = x + ['num_buys', ...]`).

The first instance (`0`) seems slightly different - but for the other 4, i think this can reduce the code to almost 1/3rd of the current code.

freqtrade/commands/analyze_commands.py
froggleston: I wasn't aware I could pass the whole config in here. I can get on this.

freqtrade/data/entryexitanalysis.py
froggleston: What's the backtest output? You should only get this if there are indeed no trades to show.

xmatthias: all i did was run `freqtrade analysis` without `--strategy` ...  (see other comments)

the backtest output itself was fine - but this ended up with backtestresult being empty (load error due to None as strategy) ...

freqtrade/data/entryexitanalysis.py
froggleston: Yeah, i've now improved this a lot - thanks for the hints

xmatthias: no push? (intentional maybe?) ... ;) just saying :laughing:

freqtrade/commands/arguments.py
froggleston: Done

freqtrade/commands/cli_options.py
froggleston: Done

freqtrade/data/entryexitanalysis.py
froggleston: Using load_backtest_stats now

freqtrade/data/entryexitanalysis.py
froggleston: Done

freqtrade/commands/analyze_commands.py
froggleston: Given the backtests_stats use now, I decided against this - I prefer being explicit about the config options required for the function?

freqtrade/data/entryexitanalysis.py
froggleston: Done

freqtrade/data/entryexitanalysis.py
froggleston: Done

freqtrade/data/entryexitanalysis.py
froggleston: aaaaand done

tests/data/test_entryexitanalysis.py
froggleston: That might be a bit pointless with the current simplicity of the trades in the backtest. I'd have to add more trades. It's possible it'll just take me a bit longer to do.

froggleston: Done

xmatthias: to be honest - that's also changed with the simplification of the whole method ...

freqtrade/commands/analyze_commands.py
xmatthias: works for me  this was more a suggestion ;)

freqtrade/commands/arguments.py
```diff
@@ -415,3 +419,10 @@ def _build_subcommands(self) -> None:
                                               parents=[_common_parser])
         webserver_cmd.set_defaults(func=start_webserver)
         self._build_args(optionlist=ARGS_WEBSERVER, parser=webserver_cmd)
+
+        # Add backtesting analysis subcommand
+        analysis_cmd = subparsers.add_parser('backtesting-analysis',
+                                             help='Backtest Analysis module.',
+                                             parents=[_common_parser, _strategy_parser])
```
xmatthias: i don't think this command is using any of the strategy methods (`--strategy`, `--strategy-path`, `--recursive-strategy-search`) anymore - so i think we can/should remove _strategy_parser from here, so it's no longer a possible (but then unused) argument.

froggleston: Ah true - I forgot to remove that.

freqtrade/commands/analyze_commands.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    config = setup_utils_configuration(args, method)
+
+    no_unlimited_runmodes = {
+        RunMode.BACKTEST: 'backtesting',
+    }
+    if method in no_unlimited_runmodes.keys():
+        from freqtrade.data.btanalysis import get_latest_backtest_filename
+
+        btfile = Path(get_latest_backtest_filename(config['user_data_dir'] / 'backtest_results'))
```
xmatthias: Looking at this now - i think we may wan't to use `exportfilename` (from the config object) instead, and add `exportfilename` to the arguments (so you get `--backtest-filename <asdf>`).

that'll allow to specify explicit backtesting files (also older ones) - which is currently impossible, and will remove the need to have `backtest_results` distributed all around.

Please be aware that 'backtest_results' can be a directory (default) - or a file (if provided explicitly) - so that'll need separate handling if you do additional checking - BUT `load_backtest_stats()` supports both implicitly.

froggleston: Yeah I used to do this way back in the buy_reasons days. No problem putting it back in. I'll get on it this week.

froggleston: So you want the default use of `backtest_results/[last_filename]` to be removed entirely and only rely on exportfilename? Happy to do this, but just checking.

xmatthias: If you look at `configuration.py` (search for `exportfilename`) - you'll notice that:

* exportfilename is a directory (pointing to `user_data/backtest_results`)
* unless `--backtest-filename <filename>` is provided (which should usually be `user_data/backtest_results/backtest-result-filename.json`)

As such, i think the correct approach is 
* check if its a directory - if yes - use "get_last_filename" to determine the latest filename.
* if it's a file - load that file (that way users can provide another, older backtest file, too).

froggleston: Hmm would there be a problem resolving the strategy name from the backtest_results metadata if people supply their own exportfilename? Or are there backtest metadata/stats files written wherever the exportfilename is?

xmatthias: it's always alongside that.
on very old versions (prior to the meta file) that may not be the case - but i don't think THAT old versions can still be shown correctly anyway

froggleston: ah OK, so:

`freqtrade\commands\analyze_commands.py` `process_entry_exit_reasons(config['exportfilename'], ...`

And that's literally it? get_backtest_stats() already checks whether its a dir/file....

froggleston:
Looks like the tests fail due to the output files being 1 sec apart:

`backtest-result-2022-06-14_16-01-40_signals.pkl`
vs
`backtest-result-2022-06-14_16-01-41_signals.pkl`

xmatthias:
> Looks like the tests fail due to the output files being 1 sec apart:
> 
> `backtest-result-2022-06-14_16-01-40_signals.pkl` vs `backtest-result-2022-06-14_16-01-41_signals.pkl`

That's a bug - but in a different area.
I didn't realize you're regenerating the timestamp for the file - then it could happen to "real" users as well (it's not a test problem) - noone guarantees that the backtest file is not generated at xx:xx:40.999 - and then the clock ticks ahead (obviously) ...

froggleston:
> That's a bug - but in a different area.
I didn't realize you're regenerating the timestamp for the file - then it could happen to "real" users as well (it's not a test problem) - noone guarantees that the backtest file is not generated at xx:xx:40.999 - and then the clock ticks ahead (obviously)

Yeah, I realise this but I thought it would be good to keep the output as close as possible to the existing files generated. Should it be refactored to simply the same date, or closest minute?

xmatthias:
> Yeah, I realise this but I thought it would be good to keep the output as close as possible to the existing files generated. Should it be refactored to simply the same date, or closest minute?

The proper fix is to simply reuse the same time. 3c62df6b86c4749991ea751e0b567ea3df7585ac fixes that (it's not REALLY related to this PR).

freqtrade/commands/cli_options.py
```diff
@@ -614,4 +614,36 @@ def __init__(self, *args, **kwargs):
...
+    "analysis_groups": Arg(
+        "--analysis-groups",
+        help=("grouping output - "
+              "0: simple wins/losses by enter tag, "
+              "1: by enter_tag, "
+              "2: by enter_tag and exit_tag, "
+              "3: by pair and enter_tag, "
+              "4: by pair, enter_ and exit_tag (this can get quite large)"),
+        nargs='?',
+        default="0,1,2",
```
xmatthias: Looking at usage of this (and the other commands) - i think it would be better to use `nargs='+'`.

That'll allow users to specify `--analysis-groups 1 2 3` (no comma) - and you'll be given a list instead of a string (so no need to split the string anymore).
That would align this better to how other parameters (`trade_ids`, `timeframes` (for downloading) are handling multi-input. 

that's obviously not limited to this command ... 

As the number of selections is limited, maybe using `choices=...` will also help to limit user errors.

freqtrade/data/entryexitanalysis.py
```diff
@@ -0,0 +1,229 @@
...
+
+            # 3: profit summaries grouped by pair and enter_tag
+            if g == "3":
+                group_mask = ['pair', 'enter_reason']
+
+            # 4: profit summaries grouped by pair, enter_ and exit_tag (this can get quite large)
+            if g == "4":
+                group_mask = ['pair', 'enter_reason', 'exit_reason']
+
+            new = bigdf.groupby(group_mask).agg(agg_mask).reset_index()
```
xmatthias: You'll want to handle "group_mask is empty" here.
using `choices=[]` in the command specification will probably not suffice, as it is still possible to define this in a configuration file

i (accidentally) used `--analysis-groups 5` as argument, which has this command crashing.

xmatthias (approved):
can't find anything more ... therefore LGTM :+1:

PR #6883, opened 2022-05-23
Change default value of process_only_new_candles to True since False …

robcaulk:
…is an uncommon usecase employed only by expert strategy devs

## Summary

`process_only_new_candles` should be `true` by default since using `false` is an uncommon use case for expert strat devs only.

## Quick changelog

Invert all definitions of `process_only_new_candles` in repo

## Breaking change? Here is why I don't think so:

For the majority of users who do not define `process_only_new_candles,` they will see no apparent change, except that this is going to save computational effort (and help with their rate limit if they are using informatives iinm). For the experts who are ensuring that they are processing tick data, they will see a change if they have not set this value in their script, but presumably if they are managing tick data then they are expert enough to be aware of this kind of change.


eSeR1805:
I think plenty of people use frequi + dry run to iterate/debug strategies. Having `process_only_new_candles` set to `False` as default allows them to see signals from the past as soon as they start the run whereas having it set to `True` would need them to wait for entry/exit triggers to actually happen on new candles. Not a rant, just something to be aware of.

robcaulk:
> I think plenty of people use frequi + dry run to iterate/debug strategy. Having `process_only_new_candles` set to False as default allows them to see signals from the past as soon as they start the run whereas having it set to `True` would need them to wait for entry/exit triggers to actually happen on new candles. Not a rant, just something to be aware of.

I was un-aware of this usecase. In that case, this group of users would need to explicitly define `process_only_new_candles` as `false` if this PR were to be accepted.

xmatthias:
> I think plenty of people use frequi + dry run to iterate/debug strategies. Having `process_only_new_candles` set to `False` as default allows them to see signals from the past as soon as they start the run whereas having it set to `True` would need them to wait for entry/exit triggers to actually happen on new candles. Not a rant, just something to be aware of.

@eSeR1805 I'm not entirely sure about this usecase or the expected problems from this. maybe you can describe the expected impact a bit more?

`process_only_new_candles` has no impact on freqUI as far as I'm aware. 
All candles will be analyzed once (when they are issued / seen by freqtrade) - and only once _in this mode_ - and are then cached (in analyzed form) in memory until the next candle arrives. This is identical to the other mode - but there we update the cache on every iteration. FreqUI is then accessing this cache (independent of this flag).

To me / my knowledge, the only expected impact (which is however a breaking change) is where the user expects to run analysis on every iteration - for example when using the orderbook for entries (which is also outlined above).


> and help with their rate limit if they are using informatives iinm

Informative pairs are downloaded only once per candle (just as regular pairs) - so i don't see any impact here unless the strategy is using orderbook / ticker data.

eSeR1805:
> @eSeR1805 I'm not entirely sure about this usecase or the expected problems from this. maybe you can describe the expected impact a bit more?

@xmatthias Tried a quick 'always buy' experiment and checked frequi. Indeed in both cases buy signals appear on past candles also so I'm thinking either it was a side effect in the past or I've misinterpreted the causality of some behavior. But certainly there were cases were `frequi` would not show signals from past candles. I'll try some other means to reproduce it.

xmatthias (approved):
I've wanted to change this for quite some time as it'll reduce processing need - and people who need this will have to opt in.

It will however not apply to most existing strategies anyway, as this is part of the default template - so unless it's been explicitly removed, i expect this to be in most strategies.

LGTM :+1:

PR #6906, opened 2022-05-30
Params to instance

### LINKED ISSUE #6871: Unique optimization values per pair
Operating system: Debian
Python Version: 3.9.12
CCXT version: 1.83.10
Freqtrade Version: develop-d5486f17d

As every market has unique arithmetic and behavior it's illogical to assume that the same indicator settings will apply perfectly for a group of pairs.

Some crypto markets have a strong correlation, yet the actual performance of the indicators needs to be perfectly fitted (as possible) per market.

Therefore I would like to suggest the following:

1. As unique parameters were introduced with buy/sell hyperopt settings, introducing a per-pair flag (rather in boot or in config) to define rather FT optimizes a per-pair scenario
2. A per-pair scenario should be guided using {self.quote} or similar in strategy for proper identification of each dataframe
3. the saved hyperopt JSON file should include per-pair parameters
4. Backtesting should be transparent 

Having this feature I believe will greatly improve FT performance in live/dry trading

xmatthias:
## Summary

Allow parameters to be initialized per instance.

Closes #6871 

## Quick changelog

- Load parameters after `bot_start()`, and from `self`, not from `cls`

### DIFF
--- a/docs/advanced-hyperopt.md
+++ b/docs/advanced-hyperopt.md
@@ -98,6 +98,23 @@ class MyAwesomeStrategy(IStrategy):
 !!! Note
     All overrides are optional and can be mixed/matched as necessary.
 
+### Dynamic parameters
+
+Parameters can also be defined dynamically, but must be available to the instance once the * [`bot_start()` callback](strategy-callbacks.md#bot-start) has been called.
+
+``` python
+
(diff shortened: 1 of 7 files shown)


rokups (approved):
Nothing stands out to me. If it works it must be OK :+1:

PR #6908, opened 2022-05-31
Persistent storage of user-custom information

eSeR1805:
## Summary

The goal is to allow user to persistently store custom information.
Credits to @nightshift2k for original implementation.

Solve the issue: Not targetting any issue.

## Quick changelog

- Add database model and middleware
- Add access functions for both modes

## What's new?

Idea is to have key-value pairs serialized and persistently stored within the database. Similarly to orders, key-value pairs are associated with trade entries.

## TODO
- [x] Add documentation
- [x] Add dedicated tests


raphant:
Perhaps adding a `default=None` to `get_kval()` would make sense. Using the example from the docs:
```python
if pair == 'BTC/USDT' and entry_tag == 'long_sma200' and side == 'long' and (current_time - timedelta(minutes=10) > trade.open_date_utc and order.filled == 0.0:
    dataframe, _ = self.dp.get_analyzed_dataframe(pair=pair, timeframe=self.timeframe)
    current_candle = dataframe.iloc[-1].squeeze()
    # store information about entry adjustment
    existing_count = trade.get_kval(key='num_entry_adjustments')
    if not existing_count:
        existing_count = 1
    else:
        existing_count += 1
    trade.set_kval(key='num_entry_adjustments', value=existing_count)

    # adjust order price
    return current_candle['sma_200']
```
Would then be:
```python
# store information about entry adjustment
existing_count = trade.get_kval(key='num_entry_adjustments', default=0)
existing_count += 1
trade.set_kval(key='num_entry_adjustments', value=existing_count)
# adjust order price
return current_candle['sma_200']
```

eSeR1805:
@raph92 I see your point and I don't have any strong opion on either of the 2 ways of handling inexistend kval. Guess your suggestion would make end user life slightly easier. Not by much though 😄   

LE: Backtracked to this. `get_kvals()` retrieves `List[KeyValue]` so specifying a default value doesn't work here.

freqtrade/persistence/keyvalue.py
```diff
@@ -0,0 +1,57 @@
...
+class KeyValue(_DECL_BASE):
+    """
+    KeyValue database model
+    Keeps records of metadata as key/value store
+    for trades or global persistant values
+    One to many relationship with Trades:
+      - One trade can have many metadata entries
+      - One metadata entry can only be associated with one Trade
+    """
+    __tablename__ = 'keyvalue'
```
xmatthias: i'd want to name this table differently - maybe `trade_userdata` - or maybe `trade_customdata`?

freqtrade/persistence/trade_model.py
```diff
@@ -915,6 +918,12 @@ def select_filled_or_open_orders(self) -> List['Order']:
                 or (o.ft_is_open is True and o.status is not None)
                 ]
 
+    def set_kval(self, key: str, value: Any) -> None:
```
xmatthias: i understand what you're trying to say ("set key value" - but shorter) - however i think for a user (no prorammer / developer) this will be confusing (what's a kval, exactly?)

maybe using something a bit more custom - like trade_ (not just for this - but also for the classes / table) - which however makes it clearer what this is (storage for user data tied to trades).

freqtrade/rpc/telegram.py
```diff
@@ -1517,6 +1520,47 @@ def _show_config(self, update: Update, context: CallbackContext) -> None:
...
+                        f"*Trade ID:* `{result['ft_trade_id']}`",
+                        f"*Type:* `{result['kv_type']}`",
+                        f"*Value:* `{result['kv_value']}`",
+                        f"*Create Date:* `{result['created_at']}`",
+                        f"*Update Date:* `{result['updated_at']}`"
+                    ]
+                    # Filter empty lines using list-comprehension
+                    messages.append("\n".join([line for line in lines if line]))
+                for msg in messages:
+                    self._send_msg(msg)
```
xmatthias: You will have to be careful with message length.

as you don't know what the content may eventually be - you could end up with 4000+ characters in one line - which is the limit per message iirc.
Now that would not crash the bot - but simply not send the message - but maybe we can handle that correctly/properly?

freqtrade/persistence/trade_model.py
```diff
@@ -1118,6 +1128,10 @@ def delete(self) -> None:
         Trade.query.session.delete(self)
         Trade.commit()
 
+        for kval in self.keyvalues:
+            KeyValue.query.session.delete(kval)
+        KeyValue.query.session.commit()
```
xmatthias: the sequence of this must be reversed - deleting the trade first will fail as long as a keyvalue is still available as there is a referential key - or the "trade" deletion will automatically delete all kvalues through the foreign key with "cascade all" (i'm not sure if that properly works on sqlite ...).


If it's not for you - make sure both tables migrate again (or use a new database) - as then you'll have the foreign key missing. 

Sqlalchemy will not migrate your tables automatically if they already exist - even if they have a different schema.

freqtrade/persistence/models.py
```diff
@@ -57,6 +58,8 @@ def init_db(db_url: str) -> None:
     Trade.query = Trade._session.query_property()
     Order.query = Trade._session.query_property()
     PairLock.query = Trade._session.query_property()
+    KeyValue._session = scoped_session(sessionmaker(bind=engine, autoflush=True))
+    KeyValue.query = KeyValue._session.query_property()
```
xmatthias: i'm ... skeptical about this (this could end up deadlocking the bot) in odd cases

The assumption with a 2nd session is that no KeyValue object will ever be touched by the "main" session - as all access goes through `trade.set_kval()` and `trade.get_kval()`.
The way the models are currently setup, that's an incorrect assumption as sqlalchemy lazy-loads relationships by default.

Put that into your `populate_*` methods (or any other method, really)

```python
# Use any filter you like - but you must use .get_trades for this to get a non-detached trade object
trade = Trade.get_trades([Trade.id == 1])

for kvals in trade.keyvalues:   # This will lazy-load this array via an additional select
    print(kvals)
```

Now i'm fully aware that through this method, you could also all trade data (this is a full, rw trade object) - but if users do this, one must assume it's intentional and if it breaks, it's not freqtrade's fault.

freqtrade/optimize/backtesting.py
```diff
@@ -151,6 +151,7 @@ def cleanup():
         LoggingMixin.show_output = True
         PairLocks.use_db = True
         Trade.use_db = True
+        KeyValues.use_db = True  # ???
```
xmatthias: What's the question here ?

docs/strategy-advanced.md
```diff
@@ -11,7 +11,7 @@ If you're just getting started, please be familiar with the methods described in
 !!! Tip
     You can get a strategy template containing all below methods by running `freqtrade new-strategy --strategy MyAwesomeStrategy --template advanced`
 
-## Storing information
+## Storing information (Non-Persistent)
```
xmatthias: To be honest - if we have persistent storing - i'm not sure if it makes sense to keep this segment in the documentation.

docs/strategy-advanced.md
```diff
@@ -40,6 +40,74 @@ class AwesomeStrategy(IStrategy):
...
+    def bot_loop_start(self, **kwargs) -> None:
+        for trade in Trade.get_open_order_trades():
+            fills = trade.select_filled_orders(trade.entry_side)
+            if trade.pair == 'ETH/USDT':
+                trade_entry_type = trade.get_kvals(key='entry_type').kv_value
+                if trade_entry_type is None:
+                    trade_entry_type = 'breakout' if 'entry_1' in trade.enter_tag else 'dip'
+                elif fills > 1:
+                    trade_entry_type = 'buy_up'
+                trade.set_kval(key='entry_type', value=trade_entry_type)
```
xmatthias: I actually think this is a bad sample.

`bot_loop_start` runs for every loop (~every 5 seconds) - and will update the value accordingly every 5s.
the filter is on open orders - but if users assume this as "good" approach for open trades, too - you can easily cause quite some load on the database, simply by updating it with the "sme" data again and again.

it's also not "really" doing what you think it does - as it'll never be called for trades which fill immediately (market orders, marketable limit orders).

docs/strategy-advanced.md
```diff
@@ -40,6 +40,74 @@ class AwesomeStrategy(IStrategy):
...
+        if pair == 'BTC/USDT' and entry_tag == 'long_sma200' and side == 'long' and (current_time - timedelta(minutes=10) > trade.open_date_utc and order.filled == 0.0:
+            dataframe, _ = self.dp.get_analyzed_dataframe(pair=pair, timeframe=self.timeframe)
+            current_candle = dataframe.iloc[-1].squeeze()
+            # store information about entry adjustment
+            existing_count = trade.get_kvals(key='num_entry_adjustments').kv_value
+            if not existing_count:
+                existing_count = 1
+            else:
+                existing_count += 1
+            trade.set_kval(key='num_entry_adjustments', value=existing_count)
```
xmatthias: You could also just count the orders that were cancelled (`trade.orders` ...) ... but i guess it serves as sample for users still.

docs/strategy-advanced.md
```diff
@@ -40,6 +40,74 @@ class AwesomeStrategy(IStrategy):
...
+            # adjust order price
+            return current_candle['sma_200']
+
+        # default: maintain existing order
+        return current_order_rate
+
+    def custom_exit(self, pair: str, trade: Trade, current_time: datetime, current_rate: float, current_profit: float, **kwargs):
+
+        entry_adjustment_count = trade.get_kvals(key='num_entry_adjustments').kv_value
+        trade_entry_type = trade.get_kvals(key='entry_type').kv_value
```
xmatthias: what if `entry_type` was never set - either because it's an old trade (before this was added) - or because the setting didn't work (see other comment) ... i guess it'd fail with "NoneType does not have kv_value" ?

freqtrade/optimize/backtesting.py
eSeR1805: I don't understand the purpose of setting `use_db` to True in `backtesting:cleanup()`. Just didn't add up for me.

docs/strategy-advanced.md
eSeR1805: Fair point. I can chop that out completely.

docs/strategy-advanced.md
eSeR1805: Yeah I've not trial tested any of the documentation snippets. When I wrote it the goal was just to illustrate some usage models. Definitely will review them and update accordingly once we settle on the core fuctionality and implementation.

freqtrade/persistence/trade_model.py
eSeR1805: Well I'm not sure if the users you've mentioned would actually see/care about this 😀 Also docs mention there is a key involved for storing/retrieving data. But I see your point and I have no problem with finding better naming for this feature or better documenting it.

freqtrade/rpc/telegram.py
eSeR1805: Thanks for the fedback on this. I was not aware of a TG message limit.
I see two options at the moment. Not sure which is the best approach here but I'm leaning towards the 1st:
1. Abort attempt to send message and log a warning (additionally maybe even inform about this occurence through TG message)
2. Truncate `kv_value` representation in the message and inform both through TG message and logger warning.

freqtrade/persistence/trade_model.py
eSeR1805: Fair point. This was not directly tested so not sure how it will behave. Will follow up when I have more info on it.

freqtrade/optimize/backtesting.py
xmatthias: These are set on the class - not on the instance.
If you don't reset it - half of (randomized) tests will fail - as it's set in the wrong state - and "non-backtesting" tests don't use the database.

eSeR1805: Oh I see now. Thanks for clarifying! Will cleanup this

freqtrade/rpc/telegram.py
xmatthias: either way works for me (even splitting in theory - but the result could not be what the user wants depending on the content).
i'm just pointing it out (the message limit is available as a constant in the telegram file).

freqtrade/persistence/trade_model.py
eSeR1805: Addressed through https://github.com/freqtrade/freqtrade/pull/6908/commits/c420304b33b94f67533864345201893ea28f2b9b .

freqtrade/rpc/telegram.py
eSeR1805: Addressed through https://github.com/freqtrade/freqtrade/pull/6908/commits/8494bea64f0b7534553b28c15bf1b5a1a791e383.

freqtrade/persistence/keyvalue.py
eSeR1805: Addressed through https://github.com/freqtrade/freqtrade/pull/6908/commits/9fdb8b07accd04b4f629d075d52d6339e02ebaa2.

freqtrade/persistence/trade_model.py
eSeR1805: Addressed through multiple commits.

freqtrade/optimize/backtesting.py
eSeR1805: Addressed through https://github.com/freqtrade/freqtrade/pull/6908/commits/9fdb8b07accd04b4f629d075d52d6339e02ebaa2.

bfiics (commented):
Love that! Keep up the good work

Axel-CH:
@eSeR1805  Is this still a WIP? This feature will be a huge benefit

xmatthias:
it's got no activity in over a year now ...

i'd consider it stalled at best - to the margin of "should probably be closed and reopened if still interested in contributing . . . "

raphant:
@xmatthias I see you have made commits into feature_keyval_storage. Is it safe to say that you are planning on taking this further?
Are there any blockers that the community can help with?

xmatthias:
@raphant For the most part, i think this mostly PR is complete now (though in hindsight - i did probably change at least every other line, so a fresh start might've been a better approach :laughing: )

More testing would help -  i did test some cases - but others will always have other usage ideas - which might trigger some edge-cases i didn't think to test so far (though in reality, there's only 4 different cases - live, backtest, storing native info (int, str) and random info (dict, dataframes) ... an every combination thereof.

xmatthias (approved):
finally (and after years) 
LGTM :+1:

PR #6914, opened 2022-05-31
Leverage tiers async

xmatthias:
## Summary

Load okx leverage tiers async

@samgermain maybe you could have a look at this (or rather - at the ratelimits for this) - if you run `freqtrade list-pairs --exchange okx --trading-mode futures` with this (requires a minimal exchange defining stake_currency to USDT).

I get something between 5 and 7 DDOS errors during this load - which suggests that the weight for this endpoint is slightly wrong.


samgermain:
@xmatthias Did your rate limit problems get fixed?

PR #6919, opened 2022-06-01
feat: add support for discord notification

ShahAnuj2610:
## Ref
https://discord.com/developers/docs/resources/webhook#execute-webhook

## look and feel
image

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -2,7 +2,6 @@
 Freqtrade is the main module of this bot. It contains the class Freqtrade()
 """
 import copy
-import json
 import logging
 import traceback
 from datetime import datetime, time, timezone
@@ -10,7 +9,6 @@ from math import isclose
 from threading import Lock
 from typing import Any, Dict, List, Optional, Tuple
 
-import requests
 from schedule import Scheduler
 
 from freqtrade import __version__, constants
@@ -35,6 +33,7 @@ from freqtrade.rpc import RPCManager
 from freqtrade.strategy.interface import IStrategy
 from freqtrade.strategy.strategy_wrapper import strategy_safe_wrapper
 from freqtrade.wallets import Wallets
+
 
 logger = logging.getLogger(__name__)
 
@@ -380,9 +379,9 @@ class FreqtradeBot(LoggingMixin):
             except ExchangeError:
                 logger.warning(f"Error updating {order.order_id}.")
 
-    #
-    # BUY / enter positions / open trades logic and methods
-    #
+#
+# BUY / enter positions / open trades logic and methods
+#
 
     def enter_positions(self) -> int:
         """
@@ -490,9 +489,9 @@ class FreqtradeBot(LoggingMixin):
         else:
             return False
 
-    #
-    # BUY / increase positions / DCA logic and methods
-    #
+#
+# BUY / increase positions / DCA logic and methods
+#
     def process_open_trade_positions(self):
         """
         Tries to execute additional buy or sell orders for open trades (positions)
@@ -580,16 +579,16 @@ class FreqtradeBot(LoggingMixin):
             return False
 
     def execute_entry(
-            self,
-            pair: str,
-            stake_amount: float,
-            price: Optional[float] = None,
-            *,
-            is_short: bool = False,
-            ordertype: Optional[str] = None,
-            enter_tag: Optional[str] = None,
-            trade: Optional[Trade] = None,
-            order_adjust: bool = False
+        self,
+        pair: str,
+        stake_amount: float,
+        price: Optional[float] = None,
+        *,
+        is_short: bool = False,
+        ordertype: Optional[str] = None,
+        enter_tag: Optional[str] = None,
+        trade: Optional[Trade] = None,
+        order_adjust: bool = False
     ) -> bool:
         """
         Executes a limit buy for the given pair
@@ -623,9 +622,9 @@ class FreqtradeBot(LoggingMixin):
 
         if not pos_adjust and not strategy_safe_wrapper(
                 self.strategy.confirm_trade_entry, default_retval=True)(
-            pair=pair, order_type=order_type, amount=amount, rate=enter_limit_requested,
-            time_in_force=time_in_force, current_time=datetime.now(timezone.utc),
-            entry_tag=enter_tag, side=trade_side):
+                pair=pair, order_type=order_type, amount=amount, rate=enter_limit_requested,
+                time_in_force=time_in_force, current_time=datetime.now(timezone.utc),
+                entry_tag=enter_tag, side=trade_side):
             logger.info(f"User requested abortion of buying {pair}")
             return False
         order = self.exchange.create_order(
@@ -747,11 +746,11 @@ class FreqtradeBot(LoggingMixin):
         return trade
 
     def get_valid_enter_price_and_stake(
-            self, pair: str, price: Optional[float], stake_amount: float,
-            trade_side: LongShort,
-            entry_tag: Optional[str],
-            trade: Optional[Trade],
-            order_adjust: bool,
+        self, pair: str, price: Optional[float], stake_amount: float,
+        trade_side: LongShort,
+        entry_tag: Optional[str],
+        trade: Optional[Trade],
+        order_adjust: bool,
     ) -> Tuple[float, float, float]:
 
         if price:
@@ -886,9 +885,9 @@ class FreqtradeBot(LoggingMixin):
         # Send the message
         self.rpc.send_msg(msg)
 
-    #
-    # SELL / exit positions / close trades logic and methods
-    #
+#
+# SELL / exit positions / close trades logic and methods
+#
 
     def exit_positions(self, trades: List[Any]) -> int:
         """
@@ -1060,10 +1059,10 @@ class FreqtradeBot(LoggingMixin):
         # Finally we check if stoploss on exchange should be moved up because of trailing.
         # Triggered Orders are now real orders - so don't replace stoploss anymore
         if (
-                trade.is_open and stoploss_order
-                and stoploss_order.get('status_stop') != 'triggered'
-                and (self.config.get('trailing_stop', False)
-                     or self.config.get('use_custom_stoploss', False))
+            trade.is_open and stoploss_order
+            and stoploss_order.get('status_stop') != 'triggered'
+            and (self.config.get('trailing_stop', False)
+                 or self.config.get('use_custom_stoploss', False))
         ):
             # if trailing stoploss is enabled we check if stoploss value has changed
             # in which case we cancel stoploss order and put another one with new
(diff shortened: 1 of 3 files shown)


freqtrade/freqtradebot.py
```diff
@@ -1543,6 +1544,43 @@ def _notify_exit(self, trade: Trade, order_type: str, fill: bool = False) -> Non
...
+                {'name': 'Profit', 'value': profit_trade, 'inline': True},
+                {'name': 'Profitability', 'value': '{:.2f}%'.format(profit_ratio * 100), 'inline': True},
+                {'name': 'Stake currency', 'value': self.config['stake_currency'], 'inline': True},
+                {'name': 'Fiat currency', 'value': self.config.get('fiat_display_currency', None), 'inline': True},
+                {'name': 'Buy Tag', 'value': trade.enter_tag, 'inline': True},
+                {'name': 'Sell Reason', 'value': trade.exit_reason, 'inline': True},
+                {'name': 'Strategy', 'value': trade.strategy, 'inline': True},
+                {'name': 'Timeframe', 'value': trade.timeframe, 'inline': True},
+            ],
+        }]
```
xmatthias: Letme put this very quickly - not this way.

For notifications - we use the `rpc` package (rpc folder).
If you think notifications for discord are necessary - then this should be implemented as an additional module (similar to telegram or webhooks).

This does not mean that all the same functionalities must be supported (it can be partial integration or whatnot) - but it must follow the same schema (you may also implement only the `send_msg()` method).

This also means, that there will be no changes necessary to other parts of the bot. 
At a first glance, all of these fields should be available in the payloat provided to send_msg as well, therefore i don't see any change necessary to this file.

freqtrade/freqtradebot.py
```diff
@@ -1818,3 +1856,22 @@ def get_valid_price(self, custom_price: float, proposed_price: float) -> float:
...
+            }
+
+            headers = {
+                "Content-Type": "application/json"
+            }
+
+            try:
+                requests.post(webhook_url, data=json.dumps(payload), headers=headers)
+            except Exception as e:
+                logger.error(f"Error sending discord message: {e}")
```
xmatthias: Wrong location - please implement this properly as separate RPC module (similar to telegram).

As you seem to be using webhooks for this - i'm not sure if this is necessary at all - or if it can be handled via the (already available) webhooks - which you can fully configure via configuration (no changes necessary).

freqtrade/freqtradebot.py
```diff
@@ -379,9 +380,9 @@ def handle_insufficient_funds(self, trade: Trade):
             except ExchangeError:
                 logger.warning(f"Error updating {order.order_id}.")
 
-#
-# BUY / enter positions / open trades logic and methods
-#
+    #
+    # BUY / enter positions / open trades logic and methods
+    #
```
xmatthias: Please remove all unnecessary formatting changes.

they're not relevant to this PR - and formatting in many parts is done intentionally as is.
This does not affect only this segment - but all segments where only whitespace was changed.

xmatthias (changes requested):
This is clearly the wrong way - we cannot accept it in this form.

If you think this needs to be a separate module (done via code) - then please implement it properly - similar to telegram / webhook integration - so all notification methods are in one place. Otherwise we'll end up with notification code all over the codebase - which is not a maintainable way to implement things.

---

Before investing more time here though, please think about (and explain) if this is really necessary, or if the same can be accomplished by configuring [integrated webhooks](https://www.freqtrade.io/en/stable/webhook-config/) (you're using webhooks, too - so i don't see why they'd be different).

what this would then turn into is a documentation topic - where we could have a predefined configuration segment for "how to configure a nice discord webhook notification".

The builtin Webhooks notification method support different formats (including json) - so unless you have good reasons to do it the "code" way (please explain why) - i see this as a configuration only topic, as the same notification should be possible via native webhooks already.

freqtrade/freqtradebot.py
ShahAnuj2610: Is it possible to pass this fields in webhooks?

ShahAnuj2610: > This is clearly the wrong way - we cannot accept it in this form.
> 
> If you think this needs to be a separate module (done via code) - then please implement it properly - similar to telegram / webhook integration - so all notification methods are in one place. Otherwise we'll end up with notification code all over the codebase - which is not a maintainable way to implement things.
> 
> Before investing more time here though, please think about (and explain) if this is really necessary, or if the same can be accomplished by configuring [integrated webhooks](https://www.freqtrade.io/en/stable/webhook-config/) (you're using webhooks, too - so i don't see why they'd be different).
> 
> what this would then turn into is a documentation topic - where we could have a predefined configuration segment for "how to configure a nice discord webhook notification".
> 
> The builtin Webhooks notification method support different formats (including json) - so unless you have good reasons to do it the "code" way (please explain why) - i see this as a configuration only topic, as the same notification should be possible via native webhooks already.

Correct. If there's a way to send in exact format like this webhooks, then it can be documented.

My idea was to make life easier to not go through that hassle and just provide webhooks_url and enabled property like telegram. If that's the route we want to go, then I'm happy to refactor and make a separate rpc for discord

xmatthias: > Is it possible to pass this fields in webhooks?

I assume this should be possible - the webhook defines the complete content that's going to be sent.

From the docs sample:
```
"webhookstatus": {
            "text": "Status: {status}"
        }
```

If you modify this a bit, it could come out as 
```
"webhookstatus": {
            "title": "Trade #{trade_id}, {profit_amount}",
            "fields": ....
        }
```

Now i'm not entirely sure IF/how it works (i never played with this myself) - so if it's not working (or not the way you expect it to), we can either modify things in the webhook to make it work (if it's simple / nonbreaking) - or explicitly support discord (own file / class, subclassing `RPC`) - similar to the current webhook implementation (but designed for discord).

ShahAnuj2610: I would go with the route of creating separate rpc. For now, will just include these `fill` notification and then we could go on expanding. Wdyt?

xmatthias: Sure, as mentioned, the extend is mostly up to you. 

obviously more is better - but not a requirement for starters.

ShahAnuj2610:
@xmatthias I've refactored into discord rpc. pls check once

ShahAnuj2610:
## profit case
image

froggleston:
This would be improved by letting the user configure the message as per other rpc calls, e.g. use the config to read in which fields should be included, something like: `if field_enums.field in config['discord']['fields']:`

ShahAnuj2610:
> 

I would recommend smth like this. So the config defines both fields and its order
```python
            # get fields order from config and remove other fields from embeds
            fields_order = self.config['discord']['fields_order']
            for embed in embeds:
                embed['fields'] = [
                    embed['fields'][i] for i in fields_order if i < len(embed['fields'])]
```
wdyt?

froggleston:
> 

dict keys aren't guaranteed to remain in the order they are inserted IIRC so you might not end up with the same ordering in `fields_order` each time

ShahAnuj2610:
> > 
> 
> dict keys aren't guaranteed to remain in the order they are inserted IIRC so you might not end up with the same ordering in `fields_order` each time

Embed[fields] is an array

freqtrade/rpc/discord.py
```diff
@@ -0,0 +1,108 @@
...
+            'profit_ratio': profit_ratio,
+            'buy_tag': trade.enter_tag,
+            'enter_tag': trade.enter_tag,
+            'sell_reason': trade.exit_reason,  # Deprecated
+            'exit_reason': trade.exit_reason,
+            'open_date': trade.open_date,
+            'close_date': trade.close_date or datetime.utcnow(),
+            'stake_currency': self.config['stake_currency'],
+            'fiat_currency': self.config.get('fiat_display_currency', None),
+        }
```
xmatthias, line 49: that's a ... dangerous docstring - i'm sure it's correct today - but it's also something that can change easily with new functionality being added - and i'm sure everyone (including myself) will forget to update this. 
It's surely helpful while developing - but not for the "final" thing - where it'll become outdated (and with that misleading) pretty quickly

freqtrade/rpc/discord.py
```diff
@@ -0,0 +1,108 @@
...
+            headers = {
+                'Content-Type': 'application/json',
+            }
+            try:
+                requests.post(
+                    self.config['discord']['webhook_url'],
+                    data=json.dumps(payload),
+                    headers=headers)
+            except Exception as e:
+                self.logger.error(f"Failed to send discord message: {e}")
```
xmatthias, line 108: you may want to slightly refactor this - splitting message generation from sending the message (basically using a similar pattern than the webhook thing).

you'll see that it will make this simpler to test - as you'll just get "embeds" (or payload?)  as parameter - and can assert if it's content is what you expect.

You may even be able to subclass `Webhook` instead (this really only provides a better way to configure the webhooks) - and use it's `_send_msg()` method for the actual network sending.

freqtrade/rpc/discord.py
```diff
@@ -0,0 +1,108 @@
...
+        }
+        """
+        self.logger.info(f"Sending discord message: {msg}")
+
+        # TODO: handle other message types
+        if msg['type'] == RPCMessageType.EXIT_FILL:
+            profit_ratio = msg.get('profit_ratio')
+            open_date = msg.get('open_date').strftime('%Y-%m-%d %H:%M:%S')
+            close_date = msg.get('close_date').strftime(
+                '%Y-%m-%d %H:%M:%S') if msg.get('close_date') else ''
```
xmatthias, line 58: Please use `DATETIME_PRINT_FORMAT` (imported from `freqtrade.constants`). It's the same dateformat - but it's not "everywhere" in the code.

xmatthias (commented):
> dict keys aren't guaranteed to remain in the order they are inserted

within python - that's no longer true, starting with 3.7, python dicts guarantee sort order.
that said - i don't think it's guaranteed while reading from the config - so we can't rely on this.

> I would recommend smth like this. So the config defines both fields and its order

having things configurable is surely an option - but the defaults should be "good for 90% of the users".
I don't see configurability of this as an absolute must though - also because it'll complicate certain things (what if i want the 2nd row to only contain 2 fields instead of 3 to improve readability for myself?)

---

For the PR itself, it's missing the following things ATM (2 of them are documentation)
* sample (disabled) entry to `config_full.example.json` (so people can easily copy/paste configs - which seems to be liked)
* doc page explaining how to do things
* tests (nothing fancy - just something so we can easily see if this is breaking in the future for some reason - basically just feeding in some message - and expecting a "request" to the configured hook url). `test_rpc_webhook.py` does something very similar ... ;)

freqtrade/rpc/discord.py
froggleston: And not all trades have all fields, e.g. leverage. This is why some kind of simple default per CandleType/Trade.is_short, and then user configurability would be good

xmatthias:
As this PR is disallowing maintainer edits, it has been merged  in 9c65fad73fbd8ee77eb1c4616ed6c9008dd95b8f.

Additional changes include full customization, tests and a documentation segment.

PR #6944, opened 2022-06-07
give extra info on rate origin for confirm_trade_*

gaugau3000:
Documentation :
-> Take into consideration the market buy/sell rates use case for the confirm_trade_entry and confirm_trade_exit callback function explanation


xmatthias (changes requested):
the change itself is fine - but must be done throughout the codebase.

you're changing the docstring only in the documentation - this is just a copy of the same docstring in the interface - as well as in the advanced template.

xmatthias (approved):

Issue #6945, opened 2022-06-08
Fatal Exception - No module named 'pymysql' in Docker container


clutch70:
## Describe your environment

  * Operating system: Ubuntu Server 18.04
  * Freqtrade Version: freqtrade 2022.4.2
  

## Describe the problem:

When trying to leverage an SQLAlchemy string in the docker container, an error is thrown reporting that pymysql is missing. My peasantly Googling (https://stackoverflow.com/questions/50333650/install-python-package-in-docker-file) makes me think that the module is missing from the dockerfile. If I'm on to a thing here I'm happy to do up a PR to add to line 40 of the dockerfile.

### Relevant code exceptions or logs

docker-compose service when error is thrown

  ```
ft-400-NFI-static-eth-kucoin-buy:
    image: freqtradeorg/freqtrade:stable
    restart: unless-stopped
    container_name: ft-400-NFI-static-eth-kucoin-buy
    volumes:
      - "./user_data:/freqtrade/user_data"
    command: >
      trade
      --logfile user_data/logs/400-NFI-static-eth-kucoin-buy.log
      --logfile syslog:1.1.1.1:514
      --db-url mysql+pymysql://user:password@localhost:3306/400_NFI_static_eth_kucoin_buy
      --config user_data/configs/400-NFI-static-eth-kucoin-buy-config.json
      --strategy NostalgiaForInfinityX
  ```

Note: Please copy/paste text of the messages, no screenshots of logs please.

  ```
 2022-06-08 02:03:55,200 - freqtrade.resolvers.exchange_resolver - INFO - Using resolved exchange 'Kucoin'...
2022-06-08 02:03:55,218 - freqtrade.commands.trade_commands - ERROR - No module named 'pymysql'
2022-06-08 02:03:55,219 - freqtrade.commands.trade_commands - ERROR - Fatal exception!
Traceback (most recent call last):
  File "/freqtrade/freqtrade/commands/trade_commands.py", line 18, in start_trading
    worker = Worker(args)
  File "/freqtrade/freqtrade/worker.py", line 35, in __init__
    self._init(False)
  File "/freqtrade/freqtrade/worker.py", line 52, in _init
    self.freqtrade = FreqtradeBot(self._config)
  File "/freqtrade/freqtrade/freqtradebot.py", line 69, in __init__
    init_db(self.config.get('db_url', None), clean_open_orders=self.config['dry_run'])
[...]

PR #6947, opened 2022-06-08
#6945 Add pymysql to Dockerfile

clutch70:
## Summary

Adds pymysql to the dockerfile

Solve the issue: #6945 

## Quick changelog

Add && pip install --user --no-cache-dir pymysql on line 40

## What's new?

Support external SQL from Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ COPY --chown=ftuser:ftuser requirements.txt requirements-hyperopt.txt /freqtrade
 USER ftuser
 RUN  pip install --user --no-cache-dir numpy \
   && pip install --user --no-cache-dir -r requirements-hyperopt.txt \
-  && pip install --user --no-cache-dir pymysql
+  && pip install --user --no-cache-dir pymysql \
 
 # Copy dependencies to runtime-image
 FROM base as runtime-image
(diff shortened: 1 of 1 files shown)


xmatthias:
It's actually intentionally not included, keeping the docker images to a minimum size.

As the [SQL documentation](https://www.freqtrade.io/en/stable/sql_cheatsheet/#mariadb-mysql) describes - packages for additional sql databases need to be installed separately.
For docker, this can be easily accomplished by following [the docker segment about custom dependencies](https://www.freqtrade.io/en/stable/docker_quickstart/#additional-dependencies-with-docker-compose).

While i do understand that you'd like to have this builtin, a majority of users are using sqlite, which is working perfectly fine.

Each of these is an optional dependency - one that only edge-cases will require.
but where would this stop? 
we should install mariadb, postgres, maybe mssql and oracle as well, because someone may intend to use it at some point? In theory, freqtrade supports all database systems that sqlalchemy supports - but that won't mean we'll have to install all drivers for all of these systems.

---

While i appreciate the attempt to fix the issue yourself - i don't currently see us cluttering the dockerfile with a dependency (thus increasing the image size) that will be unused by 98% of the users.

clutch70:
Yep makes sense. Thanks for the pointer on using the custom dockerfile!!

PR #7045, opened 2022-07-05
replace the word "abortion" with "denied" in log messages

robcaulk:
## Summary

Replace the word "abortion" with "denied" in log messages (and improve consistency with new entry/exit terminology). 

The word "abortion" is not used properly here, I can't even find a secondary definition for "abortion" that would fit. The principle connotation for abortion is clearly medical/political and thus, improper here. Further, the grammar itself of this message is bizarre. 

Perhaps we could replace "User requested abortion of buying PAIR" with "User aborted buying PAIR." But if we are going to make the change, we might as well increase the accuracy/consistency. 

The log info stems from `confirm_trade_entry/exit()`. If the confirm_trade_entry/exit returns False, then it is "denying" the entry.

So it would make most sense to say the user "denied", the entry/exit.

Either way, the use if `buying` in the message is certainly inaccurate now that we are sticking to entry/exit terminology with the new short/long features.

Let me know your thoughts. 

Solve the issue: #___

## What's new?

```bash
2022-07-05 10:49:28,417 - freqtrade.freqtradebot - INFO - User requested abortion of buying EGLD/USDT
```

Becomes

```bash
2022-07-05 10:49:28,417 - freqtrade.freqtradebot - INFO - User denied entry for EGLD/USDT.
```

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -634,7 +634,7 @@ class FreqtradeBot(LoggingMixin):
                 pair=pair, order_type=order_type, amount=amount, rate=enter_limit_requested,
                 time_in_force=time_in_force, current_time=datetime.now(timezone.utc),
                 entry_tag=enter_tag, side=trade_side):
-            logger.info(f"User rejected entry for {pair}.")
+            logger.info(f"User denied entry for {pair}.")
             return False
         order = self.exchange.create_order(
             pair=pair,
@@ -1465,7 +1465,7 @@ class FreqtradeBot(LoggingMixin):
                 time_in_force=time_in_force, exit_reason=exit_reason,
                 sell_reason=exit_reason,  # sellreason -> compatibility
                 current_time=datetime.now(timezone.utc)):
-            logger.info(f"User rejected exit for {trade.pair}.")
+            logger.info(f"User denied exit for {trade.pair}.")
             return False
 
         try:
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -634,7 +634,7 @@ class FreqtradeBot(LoggingMixin):
                 pair=pair, order_type=order_type, amount=amount, rate=enter_limit_requested,
                 time_in_force=time_in_force, current_time=datetime.now(timezone.utc),
                 entry_tag=enter_tag, side=trade_side):
-            logger.info(f"User requested abortion of buying {pair}")
+            logger.info(f"User rejected entry for {pair}.")
             return False
         order = self.exchange.create_order(
             pair=pair,
@@ -1465,7 +1465,7 @@ class FreqtradeBot(LoggingMixin):
                 time_in_force=time_in_force, exit_reason=exit_reason,
                 sell_reason=exit_reason,  # sellreason -> compatibility
                 current_time=datetime.now(timezone.utc)):
-            logger.info(f"User requested abortion of {trade.pair} exit.")
+            logger.info(f"User rejected exit for {trade.pair}.")
             return False
 
         try:


froggleston:
I would prefer rejected over denied. Even "User rejected confirmation of trade entry" might be even more explicit and the wording ties to the `confirm_` function naming scheme...

robcaulk:
> 

Agreed about being more explicit, but why use "rejected" over "denied?" Denied is a clear antonym to "confirm". Reject is an antonym of "accept." But this function is not accepting entry/exit, it is confirming entry/exit. 

For the explicit message, I like:

`Entry/exit denied by user in confirm_trade_entry/exit for {pair}`

Which I guess was suggested by @jimmynixx

xmatthias:
Honestly - i have no preference either way.

I do clearly appreciate the removal of "abortion" - which is probably some abuse of a non-english native for "abort" (probably it was me :laughing:)

---

What I've been thinking about in the past days though is if verbosity of this should be reduced to 1 message per candle.
currently, if you "abort" entry (say unconditionally - for the duration of the candle) - then you'll be spammed with the same message (For the same pair) ~every 5 seconds - clogging up (and causing log-rotation) quite quickly - especially if you have many signals and use this as a main source to prevent entries.

The downside to this is that you'll no longer see when exactly this was allowed to "pass" (entry was confirmed) - HOWEVER - the next messages would then be about order placement - so i guess that can serve as "confirmation" - too.

Opinions, please ;)

robcaulk:
> Honestly - i have no preference either way.
> 
> I do clearly appreciate the removal of "abortion" - which is probably some abuse of a non-english native for "abort" (probably it was me laughing)
> 
> What I've been thinking about in the past days though is if verbosity of this should be reduced to 1 message per candle. currently, if you "abort" entry (say unconditionally - for the duration of the candle) - then you'll be spammed with the same message (For the same pair) ~every 5 seconds - clogging up (and causing log-rotation) quite quickly - especially if you have many signals and use this as a main source to prevent entries.
> 
> The downside to this is that you'll no longer see when exactly this was allowed to "pass" (entry was confirmed) - HOWEVER - the next messages would then be about order placement - so i guess that can serve as "confirmation" - too.
> 
> Opinions, please ;)

From my experience, more log the better - so I would vote against making this message more opaque (e.g. limiting to once per candle) in favor of reducing verbosity. 

On the other hand, if someone is relying on confirm_trade_* to constantly be avoiding entry/exit on many pairs on all candles - maybe they are the type of person who will want to clean the log. Then maybe it makes sense to add a setting in the stratgey similar to `process_new_candles_only`. Something like `limit_log_info_to_once_per_candle`?

froggleston:
Hmm OK yeah I see your point about `confirm` and `denied` @robcaulk - sounds fair.

If there is an option to verbose-ify the messages for debugging every bot loop, instead of the proposed new default of every candle, at least then people have the ability to get the more granular messages.

xmatthias:
If any, i'd log it on debug level - so if you REALLY wanna see it - you need to use debug logging (which will have other, more verbose side-effects too).
Having such things as "configurable" really provides little benefit - i bet if i'd not have asked but simply changed it noone would've even noticed :laughing: 

even assuming you only have 1 signal active - on a 5m candle, you'd get ~60 (60 * 5m / 5s) messages that all say exactly the same throughout the candle - so i don't really see any benefit here (it's not even about logging more, really).

froggleston:
Sure, debugging level sounds great and 100% agree that the current verbosity is pretty.... intense so a change would be welcome.

xmatthias (approved):
As we did agree on "denied" - let's accept this as is :+1: 

Thanks a lot for the mindful change! :100:

PR #7072, opened 2022-07-11
Improve FreqAI documentation

lolongcovas:
## Summary

Add explaination about the training parameters.


docs/freqai.md
```diff
@@ -408,6 +408,11 @@ It is common to want constant retraining, in whichcase, user should set `live_re
 
 ### Controlling the model learning process
 
+Depending on what AI model to be used, these parameter names could be different. For example, the accepted parameters for the `Catboost`
+models are `data_split_parameters`, `n_estimators` and etc. For the model like SVM regression model, the accepted parameters are different.
```
robcaulk: This is untrue, `data_split_parameters` is not unique to Catboost. `data_split_parameters` are used for all FreqAI models which require a data split. Do you mean `model_training_parameters`? The dict entry for `model_training_parameters` in the user config is common for all models, but the contents of that particular dict entry depends on the type of model being employed (catboost vs lightgbm vs tensorflow).

docs/freqai.md
```diff
@@ -425,6 +430,8 @@ where $W_i$ is the weight of data point $i$ in a total set of $n$ data points._
 Finally, `period` defines the offset used for the `labels`. In the present example,
 the user is asking for `labels` that are 24 candles in the future.
 
+Note: since we work time series data and want to train a AI model to predict the future, the validation/test data should be the "future" by a given training data. Thus, we strongly recommend to disable `shuffle` parameter during the cross-validation steps. For more detailed explaination, visit [here](https://medium.com/@soumyachess1496/cross-validation-in-time-series-566ae4981ce4).
```
robcaulk: Generally this is true, however this depends on the model construction and the autocorrelation in time of the datapoints (which depends on the feature engineering). Since FreqAI is aimed at general use cases and fully flexible feature engineering, the note should avoid deterministic language like "should" and "strongly recommend." Something like:

"Typically in time-series forecasting, the training data is non-shuffled to avoid biasing training with future data. This can be accomplished with `shuffle` in `data_split_parameters` set to False. "

The article on cross-validation is valid, but for this particular point it is irrelevant. If you or someone else introduces cross-validation functionality into FreqAI, then perhaps linking to the article in the doc will be useful. 

Also: generally, Freqtrade documentation aims to avoid the use of the word "we."

robcaulk (changes requested):
Thanks for adding this to the doc. Please see my comments in the review.

docs/freqai.md
lolongcovas: Yes, you are right. I put wrongly the `data_split_parameters` which is quite common for the /val/test dataset splits. I wanted to say the training parameters could be different for one ML approach from another. So, feel free to add the correspond training parameters

docs/freqai.md
lolongcovas: Ok thanks for the recommendation. One thing I noticed is that the `shuffle` parameter only makes sense for the cross validation and achieve a model quite generalized for the future predictions. However, in my trials, the trained model with `shuffle=false` and `test set = 0.33` predict badly the future candles. why? the data used for the training are quite old and I used the most recent data as test set. Once said that, we only care our model predicts correctly the target value for the `label_period_candles` future candles. So yes, we could train the model with all data (`data_split_parameters.test_size = 0`).

One idea, the `.fit()` method could allow the possibility of hyperparameter tuning by predicting correctly the next `label_period_candles` target value and use these hype params to train with the whole data. This is only feasible for high end machines do the job within the `timeframe` (e.g. 5m).

robcaulk: yes we can train on all data without breaking any out for testing, it needs a bit of exploration (I’m running some dry runs on this type of training strategy).

Cross validation is a feature that Freqai needs, it is something I am trying to keep in mind while building the architecture. 

We’ve discussed this idea a bit about having a dynamic `label_period_candles` which is optimized based on recent model prediction performance. Its definitely something I’d like to see in the code at some point since it removes a parameter from the user side.

docs/freqai.md
robcaulk: This PR makes me think it could be useful to split out some sections for various models. “Common catboost regressor considerations” “common lightgbm considerations” “common tensorflow considerations” etc. Where we discuss some freqai related interactions which each library.

docs/freqai.md
lolongcovas: > Typically in time-series forecasting

I think the article is still valid because we allow user to test the model on the test set (`"test_size": 0.25`). In order to avoid the misleading model performance, we should disable the `shuffle`.

robcaulk (commented):
Fix english, generalize the writing, improve clarity.

PR #7085, opened 2022-07-14
freqai: fix issue when bot restarts with same identifier, does not load predi…

wagnercosta:
## Summary

When the bot restarted, the prediction history was not loaded, using the same identifier


robcaulk:
Lgtm!

PR #7105, opened 2022-07-18
Feat/freqai

lolongcovas:
## Summary

Speed up the data weights computation by using numpy vectorization.
add logs for start and end day of training data.


freqtrade/freqai/prediction_models/BaseRegressionModel.py
```diff
@@ -39,7 +39,11 @@ def train(
         :model: Trained model which can be used to inference (self.predict)
         """
 
-        logger.info("--------------------Starting training " f"{pair} --------------------")
+        start_date = unfiltered_dataframe["date"].iloc[0]
+        end_date = unfiltered_dataframe["date"].iloc[-1]
+        logger.info("-------------------- Starting training " f"{pair} --------------------")
+        logger.info("-------------------- Using data "
+                    f"from {start_date} to {end_date}--------------------")
```
robcaulk: I like the idea of giving feedback to the user for the trained date range - but this is the incorrect information. Here you report the dates associated with the unfiltered data. Trained data is on the filtered dataframe. 

Also, I dont think reporting down the second is useful here - it only clutters the log. Trained data is always at the scale of days, so I think days months years is the only important information.

I will make these changes and merge it. In the future, try to keep topics separated in PRs. HEre we have the weights setting combined with logging training dates - so one cannot be pushed until both are ready.

lolongcovas: hei, thanks for the modification and merge. I didnt have time to review it. next time i will create unit separated PRs. +1

PR #7108, opened 2022-07-19
Added description heading to README.md

rzrymiak:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@
 [![Documentation](https://readthedocs.org/projects/freqtrade/badge/)](https://www.freqtrade.io)
 [![Maintainability](https://api.codeclimate.com/v1/badges/5737e6d668200b7518ff/maintainability)](https://codeclimate.com/github/freqtrade/freqtrade/maintainability)
 
+## Description
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,8 @@
 [![Documentation](https://readthedocs.org/projects/freqtrade/badge/)](https://www.freqtrade.io)
 [![Maintainability](https://api.codeclimate.com/v1/badges/5737e6d668200b7518ff/maintainability)](https://codeclimate.com/github/freqtrade/freqtrade/maintainability)
 
+## Description
+
```
xmatthias, line 9: This part is unnecessary, adding an unnecessary header, moving the content lower - which is not desired.

The other section is fine as is - but please revert this part of the change.

xmatthias: This part is unnecessary, adding an unnecessary header, moving the content lower - which is not desired.

The other section is fine as is - but please revert this part of the change.

xmatthias (changes requested):

xmatthias (approved):
LGTM :+1:

PR #7114, opened 2022-07-22
freqAI: Bugfix/avg mean dist

lolongcovas:
## Summary

Solve the issue of average distance computation by ignoring the sample itself distance values


xmatthias (changes requested):
Please use a clean branch for this.
you should be able to get to this with `git rebase upstream/feat/freqai` (assuming your upstream (for this repository) is named upstream). After that, do a `git push -f`.

The result should be that this PR is updated with new history - which will contain only one (the latest) commit.

PR #7174, opened 2022-08-04
Fix typo

stash86:
adjust_trade_position should return stake_amount, not amount

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
nice catch :+1:

PR #7192, opened 2022-08-07
Add script for patching conda libta-lib

AchmadFathoni:
## Summary

Add script for patching  conda libta-lib with modification of `build_helpers/install_ta-lib.sh` to fix this failed test:
```
______________________________________________________________________ test_talib_bollingerbands_near_zero_values ______________________________________________________________________

    def test_talib_bollingerbands_near_zero_values():
        inputs = pd.DataFrame([
            {'close': 0.00000010},
            {'close': 0.00000011},
            {'close': 0.00000012},
            {'close': 0.00000013},
            {'close': 0.00000014}
        ])
        bollinger = ta.BBANDS(inputs, matype=0, timeperiod=2)
>       assert bollinger['upperband'][3] != bollinger['middleband'][3]
E       assert 1.25e-07 != 1.25e-07

tests/test_talib.py:14: AssertionError
=================================================================================== warnings summary ===================================================================================
``` 

## What's new?

- Now user that install freqtrade with conda  can use script to patch libta-lib.
- Updated the installation documentation.


docs/installation.md
```diff
@@ -326,6 +326,13 @@ python3 -m pip install --upgrade pip
 python3 -m pip install -e .
 ```
 
+Patch conda libta-lib
+
+```bash
+cd build_helpers
+bash install_ta-lib_conda.sh
```
xmatthias: I don't really see this as applicable to a conda environment (at least not correctly).

To me, there is no reason to use conda in a linux/macos environment (where python / build dependencies are easy to install) - so the main usecase for conda is windows - where installation is simplified to some degree.
As such, a linux only script provides no benefit.

One benefit would be to have a windows script for this - but you don't really want to go there (build dependencies on windows are a pain), and instead reuse the available windows wheels.

For linux however, installing ta-lib can also be done via the available script - and will work just fine for a conda environment.

AchmadFathoni: OK, should the documentation be updated to tell the conda user how to patch ta-lib? To prevent new user confusion.

AchmadFathoni: btw you can close this PR.

xmatthias (approved):
it's clearly necessary for conda installs, too - but i don't like having scripts duplicated - as they will drift apart over time.

I've updated this now to reuse the existing script - which simplifies usage.

PR #7209, opened 2022-08-09
Sorted whitelist in rpc

ecoppen:
## Summary
Added `sorted()` to line 883 so that when executing `/whitelist` command the pairings will be in alphabetical order making it faster to manually identify whether the pair you are confirming/looking for is present.

## Quick changelog
883: `'whitelist': sorted(self._freqtrade.active_pair_whitelist)`

## For the future
Separate the quote asset that's repeated `n` pairing times, on the assumption that only one quote asset can be used per freqtrade instance to save the redundant data taking up so much space.

### DIFF
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -884,7 +884,7 @@ class RPC:
         """ Returns the currently active whitelist"""
         res = {'method': self._freqtrade.pairlists.name_list,
                'length': len(self._freqtrade.active_pair_whitelist),
-               'whitelist': self._freqtrade.active_pair_whitelist
+               'whitelist': sorted(self._freqtrade.active_pair_whitelist)
                }
         return res
 
--- a/tests/rpc/test_rpc.py
+++ b/tests/rpc/test_rpc.py
@@ -1183,7 +1183,7 @@ def test_rpc_whitelist(mocker, default_conf) -> None:
     ret = rpc._rpc_whitelist()
     assert len(ret['method']) == 1
     assert 'StaticPairList' in ret['method']
-    assert ret['whitelist'] == default_conf['exchange']['pair_whitelist']
+    assert ret['whitelist'] == sorted(default_conf['exchange']['pair_whitelist'])
 
 
 def test_rpc_whitelist_dynamic(mocker, default_conf) -> None:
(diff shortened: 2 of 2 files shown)


stash86:
Personally I prefer my whitelist to be sorted the way it really is, which in my case by volume. I would prefer this added as additional setting in the config, so that people who want to want their whitelist shown alphabetically can just enable it on config

xmatthias (changes requested):
Hey there,

Thanks for your contribution! :100: 

The pairlist ordering is something that's quite important for freqtrade operations.
In case of multiple signals triggering at the same time (quite a frequent occurance - as crypto pairs often show similar moves), pairs early in the pairlist will get a trade, while latter pairs will no longer get a trade slot (`max_open_trades`), and it's signal is therefore ignored.

As such - there is quite some information within the pairlist ordering itself - which this PR would remove from the user.
As such, this is unfortunately not a contribution we're able to accept, as it would remove the user's ability to see the pairlist sorting that's actually used.

If you think this is **really** necessary, adding a `/whitelist sorted` to telegram (should only impact telegram, not the API!) might be a possibility that's non-breaking and not removing information.

I'll keep this open still for a bit to see your thoughts on this.

ecoppen:
If that’s preferred more widely then I can change this around. Do note this is just for rpc purposes and would not affect the whitelist shown in console/logs.

Alternatively It could be `pair (rank)` e.g. `btc/usdt (1)` and eventually `base: usdt` followed by `btc (1), doge(3), eth (2)`, but this would probably be more confusing.

ecoppen:
It’s not _really_ necessary however I understand the issues you have both presented, I’ll close this and head back to the drawing board. Thank you!

xmatthias:
> Do note this is just for rpc purposes and would not affect the whitelist shown in console/logs.

it's clear that this was for RPC usages only - but when i look at the whitelist (using volumepairlist, obviously) - i can see "pair X is at spot 3" - which is often surprising as it's a small pair with a huge spike in volume.

Removing the sorting from rpc would lead to confusion as people do expect the sequence there (and if it's not as desired, they need to change their pairlist setup) ...

ecoppen:
Thanks for the confirmation of this, I've reattempted this with your suggestions. The comment before that you're referring to wasn't directed your way - I haven't mastered replying to a specific comment 😅

PR #7211, opened 2022-08-10
Optional /whitelist args - sorted, nobase

ecoppen:
## Summary
Added two optional arguments to `/whitelist` command in Telegram:

- `sorted` for alphabetical order
- `nobase` to remove the base currency

## Quick changelog
- `/whitelist sorted` for alphabetical whitelist
- `/whitelist nobase` to remove the base currency from each pairing in the whitelist
- `/whitelist sorted nobase` for an alphabetical whitelist without base currencies listed

## Other notes
I gave the testing a go but it might need further adjustment? 😅

## Example 
image

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -1371,7 +1371,7 @@ class Telegram(RPCHandler):
             if context.args:
                 if "sorted" in context.args:
                     whitelist['whitelist'] = sorted(whitelist['whitelist'])
-                if "nobase" in context.args:
+                if "baseonly" in context.args:
                     whitelist['whitelist'] = [pair.split("/")[0] for pair in whitelist['whitelist']]
 
             message = f"Using whitelist `{whitelist['method']}` with {whitelist['length']} pairs\n"
@@ -1493,8 +1493,8 @@ class Telegram(RPCHandler):
             "*/fx <trade_id>|all:* `Alias to /forceexit`\n"
             f"{force_enter_text if self._config.get('force_entry_enable', False) else ''}"
             "*/delete <trade_id>:* `Instantly delete the given trade in the database`\n"
-            "*/whitelist [sorted] [nobase]:* `Show current whitelist. Optionally in "
-            "order and/or without the base currency.`\n"
+            "*/whitelist [sorted] [baseonly]:* `Show current whitelist. Optionally in "
+            "order and/or only displaying the base currency of each pairing.`\n"
             "*/blacklist [pair]:* `Show current blacklist, or adds one or more pairs "
             "to the blacklist.` \n"
             "*/blacklist_delete [pairs]| /bl_delete [pairs]:* "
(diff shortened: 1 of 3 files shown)


tests/rpc/test_rpc_telegram.py
```diff
@@ -1458,6 +1458,23 @@ def test_whitelist_static(default_conf, update, mocker) -> None:
...
+    context = MagicMock()
+    context.args = ['nobase']
+    msg_mock.reset_mock()
+    assert ("Using whitelist `['StaticPairList']` with 4 pairs\n"
+            "`ETH, LTC, XRP, NEO`" in msg_mock.call_args_list[0][0][0])
+
+    context = MagicMock()
+    context.args = ['nobase', 'sorted']
+    msg_mock.reset_mock()
+    assert ("Using whitelist `['StaticPairList']` with 4 pairs\n"
```
xmatthias, line 1479: You'll have to call the whitelist again.
just resetting the mock - and then asserting won't work properly (and is causing the ci failure)

applies to each of the asserts above.

freqtrade/rpc/telegram.py
```diff
@@ -1368,6 +1368,12 @@ def _whitelist(self, update: Update, context: CallbackContext) -> None:
         try:
             whitelist = self._rpc._rpc_whitelist()
 
+            if context.args:
+                if "sorted" in context.args:
+                    whitelist['whitelist'] = sorted(whitelist['whitelist'])
+                if "nobase" in context.args:
+                    whitelist['whitelist'] = [pair.split("/")[0] for pair in whitelist['whitelist']]
```
xmatthias: A currency pair has the following naming: `base/quote[:settle]` (settle only for futures).

Keeping the above in mind, think the selected naming is pretty confusing.

it's called nobase - but in reality, it's removing the quote currency (and potentially the settle currency), leaving the base-currency in place, so either "baseonly" or "noquote" would be more fitting (i'd prefer the positive version though, as the only thing left will be the base).

xmatthias (changes requested):
Looks great and works as expected (even though CI still disagrees, the problem there is really with the test).

Please also add the new comands to the `telegram.md` documentation page - so people can also find out about the new commands.

tests/rpc/test_rpc_telegram.py
ecoppen: Thanks for this guidance, this is the area I'm least confident with so should be resolved now if it's as simple as I understood it to be.

freqtrade/rpc/telegram.py
ecoppen: I can't believe I've mixed these two up, especially after referring to them in the correct way on discord, perhaps I second guessed myself. I have changed to `baseonly` following your recommendation.

ecoppen:
> Please also add the new comands to the `telegram.md` documentation page - so people can also find out about the new commands.

Hopefully this is `telegram_usage.md` otherwise I've updated the wrong one.

ecoppen:
Thanks for fixing the tests!

xmatthias (approved):
Tests are ... often confusing for a lot of people - and it's no problem to ask for help, considering an attempt was made :1st_place_medal: 

LGTM :+1: 

Thanks for your contribution! :100:

PR #7214, opened 2022-08-11
Allowing custom RPC messages to be sent from strategy

emiliobasualdo:
I have **not** included tests yet. I want to know what you guys think.
## Summary
I would like to send custom RPC messages from inside the strategy. 
Basically I would like to receive notification regarding certain conditions that happen along the execution of the strategy.

## Quick changelog

- IStrategy is given the RPCManager.send_msg function
- Custom msgs are allowed in Telegram

### DIFF
--- a/freqtrade/enums/rpcmessagetype.py
+++ b/freqtrade/enums/rpcmessagetype.py
@@ -19,6 +19,8 @@ class RPCMessageType(Enum):
 
     STRATEGY_MSG = 'strategy_msg'
 
+    CUSTOM_MSG = "custom_msg"
+
     def __repr__(self):
         return self.value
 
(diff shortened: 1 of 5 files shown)


xmatthias:
Hi There,

Thanks for your contribution.
This is however already possible in the current develop version (will be included in the next release).

You can find more information how to do this from the strategy in the [latest documentation (deep link)](https://www.freqtrade.io/en/latest/strategy-customization/#send-notification).

As such - there's unfortunately no point in continuing this PR - as it'd add duplicate functionality - done slightly differently - but without the protections in place to avoid spamming users.

emiliobasualdo:
Well, thats great news!
Best regards and thanks for the fast response.

Issue #7138, opened 2022-07-28
Activate bot trading during predefined hours


darioneto:
Not sure if there is any better solution than cron job to start/stop the docker daemon? Can you please advise what is the best approach to keep the bot active only during predefined hours on daily basis?

xmatthias:
I've transferred this to the main repository, as it's really not related to the technical library.

I'm not sure what your aim is with this - but freqtrade is really designed to be continuously running.
Assuming you stop freqtrade every day at a certain time - it may or may not have positions open, which it'll then no longer follow (until the restart). You might therefore miss potential exit opportunities (or stoplosses)- so i really think this approach is wrong.


Most likely, what you want is a date-based guard  - which prevents entering at certain times of the day. This can be easily accomplished by comparing the dataframe date with something else (for example `dataframe['date'].dt.hour < 20` (in UTC) - _untested_)

darioneto:
Thanks for your reply. As long as the stop loss is sent to the exchange I'm not concerned with switching it off for a certain amount of hours. What would be the difference between dataframe validation or calling via API to stop/start?

xmatthias:
As said above - stop stops all bot operations - so the bot evaluation will no longer run.
if you physically stop it (kill the process) - or call `/stop` is equivalent.
the only difference is that you can restart it remotely from `/stop` - which you can't if you kill the process. Then you need "something" to bring the process back up.

This includes not just stoploss - but also take profits - so if your pair had a 50% profit overnight - dropping back to -5% when you switch it on - it's as if that never happened - but you missed 50% profits.

darioneto:
I understand, So in this case the best approach would be to use dataframe. So let's say for example that I need to turn off the bot between 1 am and 2 am UTC every day would this work? 
                 ............
                (dataframe['volume'] > 0) &
                ( dataframe['date'].dt.hour > 2) & (dataframe['date'].dt.hour < 1  )

In the  populate_entry_trend validation formula

xmatthias:
you should change your wording / thinking from "turn it off" to "disable entry signals". That'll make your goal clearer - and will help in formulating your statement.

from afar (without really testing it) - it will work. 
i'd recommend to backtest this (maybe also with a wider range) - so you can see the effects on this (i assume this is based on "big whales trade mostly at this point ..." or "My signal is weak at that time ..."

darioneto:
[...]

PR #7222, opened 2022-08-13
TradindTime protection & Telegram channel notification support

farmage:
## Summary
Sorry, I'm not a GitHub guru, so all changes in one PR. If you know an excellent way to separate it, I'll very much appreciate it if you drop me some links to documentation.

TradingTime
I made it as Protection. I agree, that it's not the best way to disable new trades exactly at "stop time", but it works well enough for me, so it may be useful for someone else.
How to use:
```
    @property
    def protections(self):
        return [
            {
                "method": "TradingTime",
                "lookback_period_candles":1,
                "start_time": "04:00", # UTC
                "end_time": "20:05" # UTC
            },            
    ]
```

The second change is for telegram
I needed notification of finished orders in a public channel, so I added some modifications to the "telegram.py" module
To use it, the user needs to add these parameters to the config:
```
{
    "use_channel_notification": true,
    "channel_notification_chat_id": "@youre_channel_name"
}
```
Filled orders, profit, and daily stats for 21 days will be available to the public.

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (changes requested):
Hi,

First off - thanks for your contribution.

This will have to be 2 PR's (not going to accept it this way, sorry).
although i don't see what benefit the telegram change would bring - so that'll need to have a good reason as to why this would actually be needed.

A guide for the github PR workflow can be found [here](https://guides.github.com/introduction/flow/), which i find a quite nice guide on how to do proper pull requests (the changes you do should be based on the upstream develop.

Assuming you did proper commits (atomic - limited to one change at a time without mixing the 2 features together in single commits) - you should be able to bring your changes to the new branch  (see PR workflow) by using `git cherry-pick <commitid>`.

xmatthias:
Asside from the **how**:
As already explained in https://github.com/freqtrade/freqtrade/issues/7138#issuecomment-1213802362 - 
This protection will not work the way you expect it to - protections are only called once a trade closes - so this will not work for the first trade - and will assume at least one trade per pair per day (which must close within the "disallowed period") to lock trading.

We have a [developer documentation](https://www.freqtrade.io/en/latest/developer/#protections) section about how protections work - which explain when each method is called (only when trades close).

Based on that - it's impossible to do what you're trying to do with protections - but a "daytime based" protection must be implemented as guard within the strategy - which however pretty easy to accomplish with `dataframe['date'].dt.hour`.

I'll therefore close this PR - as this way is not feasible for the goal you're trying to accomplish - and therefore also makes continued work on this pointless.

darioneto:
@xmatthias , the proposed solution is not working in practice for some reason. I have this deployed and strangely enough, the trades are still being sent out at a full hour. 

for example, if you add 
```
...
                (dataframe['date'].dt.hour != 1) &
                (dataframe['date'].dt.hour != 2)
...
```
you will notice the trade is being sent out a few seconds after the full hour mark.

As long as this would cover the theoretical time frame as expected it's a good enough solution, however, it's not working for me. 
not sure if this could be due to the fact this setting is in place
`    process_only_new_candles = False`

darioneto:
actually, small correction, looking at this history this is not working at all, not only at the full hour. Trades are still being executed during the expected suspension window. The time of the container is at UTC as well at the (dataframe['date'].dt.hour != 1) statement

xmatthias:
The most likely reason is the other conditions. 
there's nothing wrong with the statement per se - but if you have a combination of and (&) and or (|) combinations, brackets can become an issue.

darioneto:
There are only "&" simple one-liner statements 

```
                (df_1["x"][0] > 8) &
                (df_2["x"][0] < 1) &
                (df_3["x"][0] < 0.01) &   
                (df_4["x"][0] < 0.01) &  
                (dataframe['date'].dt.hour != 1) &
                (dataframe['volume'] > 0) &
                (dataframe['date'].dt.hour != 2)
```

from that above, I was expecting trades now to be executed between 1:00:00 - 2:59:59 UTC
which is not the case.

xmatthias:
well you're comparing with "!=" - so the expectation should be that no trades are executed within that window - but trades are executed throughout the remainder of the day.

darioneto:
this is precisely what I expect to happen however this is not the case. Trades are still being executed regardless

xmatthias:
The above will not generate signals between 1 and 3 am UTC.
You can also verify this using plotting / freqUI plots.

If you see signals during that period, then you've most likely got duplicate strategies (or you're modifying a wrong strategy file alltogether).


This will work well for entries - for exits, there's other reasons (roi, stoploss, ...) - which will not be impacted by signals directly.
for that, you'll need to use the "confirm_exit" callbacks - and prevent exits there (if that's really what you need).

darioneto:
This is matched based on the Entry Tag which is the only one that responds with these conditions. There are no other tags in the strategy that comes with the trade.
each content of dataframe.loc[] has different conditions with a different tag.

xmatthias:
with the information you provide - it's impossible to help you. 

this is also the wrong place for this. 
if you have problems, please open an issue describing your problem in detail, with a full, minimum reproducible strategy (basically only this signal).

PR #7236, opened 2022-08-16
Fix input shape for LighGBMClassifier

robcaulk:
Ameliorate the following warning:

```
freqtrade/.env/lib/python3.10/site-packages/sklearn/preprocessing/_label.py:98: DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples, ), for example using ravel().
  y = column_or_1d(y, warn=True)
  ```
  
  And add `sample_weight` and `eval_sample_weight` arguments to LightGBMClassifier/Regressor prediction models.


xmatthias (approved):
LGTM :+1:

PR #7244, opened 2022-08-17
extract download-data from freqai to prepare for future async changes

robcaulk:
## Summary

Move automated datadownload out of FreqAI to prepare for future async changes.


freqtrade/freqai/data_kitchen.py
```diff
@@ -1027,3 +1002,78 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
...
+    :param timerange: TimeRange = The full data timerange for populating the indicators
+                                    and training the model.
+    :param dp: DataProvider instance attached to the strategy
+    """
+    all_pairs = copy.deepcopy(
+        config["freqai"]["feature_parameters"].get("include_corr_pairlist", [])
+    )
+    for pair in config.get("exchange", "").get("pair_whitelist"):
+        if pair not in all_pairs:
+            all_pairs.append(pair)
```
xmatthias: i think this can be reduced to a call to `dynamic_expand_pairlist()` - which odes exactly this logic for `download-data`.
The sequence will be (slighlty) different - but it'll be aligned - and also support regex whitelists.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1027,3 +1002,78 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
...
+
+    max_period = config["freqai"]["feature_parameters"].get(
+        "indicator_max_period_candles", 20
+    ) * 2
+    additional_seconds = max_period * max_tf_seconds
+
+    trained_timerange.startts = int(
+        time - config["freqai"].get("train_period_days", 0) * SECONDS_IN_DAY
+    )
+    trained_timerange.stopts = int(time)
```
xmatthias: the result of this seems to be unused?

freqtrade/freqai/data_kitchen.py
```diff
@@ -1027,3 +1002,78 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
...
+    max_tf_seconds = 0
+    for tf in timeframes:
+        secs = timeframe_to_seconds(tf)
+        if secs > max_tf_seconds:
+            max_tf_seconds = secs
+
+    max_period = config["freqai"]["feature_parameters"].get(
+        "indicator_max_period_candles", 20
+    ) * 2
+    additional_seconds = max_period * max_tf_seconds
```
xmatthias: i'm ... not sure if this is correct - and i think it will result in a lot more data being downloaded than what's actually necessary.

let's assume the following include_timeframes:
`['3m',  '1d']` - as well as `indicator_max_period_candles` (how is this different to the existing `startup_candle_count`?) of 100.

The result in `additional_seconds` would be 100 * 86400 (100 days)  - which is then applied as timerange for all timeframes.
this means - we'd be downloading 100 days of 3m data - but in reality, we'd only need 100 * 60 * 3 for the 3m timeframe and 100 * 60 * 60 for the 1h timeframe - which is significantly less than the specified "additional seconds".

i therefore doubt we'll need 100 days of 3m data.

a potential solution would be to call `refresh_backtest_ohlcv_data()` in a loop of timeframes - calculating the "startup_candle_count" per timeframe instead of using the longest for everything.

robcaulk: The calculation here is correct. But you are right we are downloading more data than is necessary for lower tfs.

>a potential solution would be to call refresh_backtest_ohlcv_data() in a loop of timeframes - calculating the "startup_candle_count" per timeframe instead of using the longest for everything.

Yeah, you are right. This would reduce amount of data necessary, I will check into improving it.

>indicator_max_period_candles (how is this different to the existing startup_candle_count?

`indicator_max_candles` is the max period used in the indicator calculation. Typically, this should match the maximum value of `indicator_periods_candles` list. In the case you describe, the user has set `['3m', '1d']` to `include_timeframes` and then let's say they set `[10,20]` to `indicator_periods_candles`. FreqAI loops on all their indicators, for all the `include_timeframes` for all the `indicator_periods_candles`.

So that means, the user asked for indicators to be computed for the `1d` timeframe with timewindows of 10 and 20. 

Now FreqAI needs to ensure that the first candle of training does not have a NaN, and that means ensuring that the there is 20 days available for computing the 1d indicators with a 20 day timeperiod :-). That gets added to the user defined `train_period_days` (because we need non NaN at the beginning of the training period).  

This *needs* to be considered for all timeframes since all timeframes use all timewindows. Yes, this is where the data expansion increases quickly. I don't recommend setting 100 to your `indicator_max_period_candles` unless you actually use 100 somewhere or have it included in the `indicator_periods_candles`. 

They are completely different. From the doc:

```
startup_candles | Number of candles needed for backtesting only to ensure all indicators are non NaNs at the start of the first train period.  Datatype: positive integer.
-- | --
```

```
indicator_max_period_candles | The maximum period used in populate_any_indicators()  for indicator creation. FreqAI uses this information in combination  with the maximum timeframe to calculate how many data points it should  download so that the first data point does not have a NaN  Datatype: positive integer.
-- | --
```

In other words, startup_candles (used in FreqAI) is only used in backtesting, this is where the user needs to manually tell FreqAI how many candles will be needed for the full backtesting period. 

`indicator_max_period_candles` is typically the maximum of `indicator_periods_candles` but could be larger in the case where the user has a single large indicator placed outside the `indicator_periods_candles` loop.

xmatthias: i honestly still don't see the difference between `startup_candles` and `indicator_max_period_candles`. 

Sure, the documentation could be changed to include freqAI for `startup_candles` - but documentation is made to cover "as is" - if usecase needs to be expanded - so be it.

Think about what `startup_candles` really does (letme rephrase it to make it's purpose clear):
"Number of candles needed to ensure that enough candles are available PRIOR to the relevant timeperiod to avoid NaN values at the beginning of the timeperiod."

for live, it's irrelevant as only the very last candle actually matters - freqai / backtesting are different - but right now, in my understanding - both serve the purpose of having no NAN candles in the dataset - hence - what would i set `indicator_max_period_candles` - vs. `startup_candles`? 

i currently can't come up with a setup where one would be 10, butt the other 100 - that'd mean that one of the 2 is misconfigured / not matching the remaining configuration.

now if both are relevant / used in freqAI is a different topic - i'm also not saying "change this now" - but i don't understand where the difference is - other than indicator_max_period_candles only applying to freqAI.

robcaulk: `indicator_max_period_candles` is in units of the highest TF candles which is used for dry/live (perhaps 1d, or 4h, typically always greater than the config['timeframe']) when FreqAI automatically computes the number of "startup_candles" (in units of the config['timeframe']) to ensure no NaNs. As you can see here, it is added to `train_period_days` and everything is converted to the base tf candle.

The user doesnt actually need to know much about any of that - all they need to know is "what is the biggest timeperiod I am asking the strategy to compute my indicators on?"

 `startup_candles` is in the base tf candle, and it includes train_period_days inside of it. 
 
 The user needs to calculate this themselves. The [docs demonstrate](https://www.freqtrade.io/en/latest/freqai/#deciding-the-sliding-training-window-and-backtesting-duration) how the user can calculate `startup_candles`

Is it still not clear?

xmatthias: well to me it's still just another way to specify `startup_candle_count` (sorry, i meant the existing value before - as we now seem to have 3(**!**) different ways to specify the same thing) - which is also in candles (not minutes or days).

The outcome of `indicator_max_period_candles` is "i need `indicator_max_period_candles` more candles prior to the earliest date (train date in this case).
Which is exactly what `startup_candle_count` is also for.

---

Let's tackle this via an example maybe:

assuming i calculate ONLY ta.RSI(dataframe, 14)  in my populate_any_indicators, and i use 3m and 1h in my informative timeframes.

What is the correct value of `indicator_max_period_candles`, `startup_candle_count` and `startup_candles`?

robcaulk: `indicator_max_period_candles` = 14. ie. the maximum period passed to talib in `populate_any_indicators()`. 

`startup_candles` can only be defined if we also know the `train_period_days`. Let's assume `train_period_days` = 10, then:

`startup_candles` = `indicator_max_period_candles` * 60 minutes/candle / 3 minutes/basecandle + `train_period_days` * 1440 minutes/day / 3 minutes/basecandle = 5080

The user doesn't need to deal with this calculation in dry/live, they only need to know "what is the maximum period I am passing to talib in `populate_any_indicators()`. It is computed for them. 

startup_candle_count is whatever the user has in the strategy.

During backtesting only, FT will set startup_candle_count += startup_candles if FreqAI is enabled.

The reason we have config['freqai']['startup_candles'] is because `startup_candle_count` is limited by the max candles allowed by exchange.  


The only change I am willing to accept here would be to add another "startup function" in the `load_freqAI_model()` function where we would compute the startup_candles automatically in the backtesting case. This was not done before because we were trying to avoid touching too much of the FT startup when we initiated development on FreqAI.

xmatthias: logically (if `startup_candle_count` is assigned correctly) - it'd be 14 (since we use 14 startup candles).

> During backtesting only, FT will set startup_candle_count += startup_candles if FreqAI is enabled.

This would then get you 28 startup candles - but i don't really see why you'd need that - as your max. lookback here would be 14 - and after 14 candles, you've got all indicators filled.

properly setting `startup_candle_count` will suffice - which makes `indicator_max_period_candles` redundant.

---

So my initial understanding was correct - that `indicator_max_period_candles` is not needed.
`startup_candles` can (and should!) be calculated internally (it's way to complicated for the user, they'll just set it to 10.000 and hope it'll work - if not, increase some more).
`indicator_max_period_candles` is just a copy of `startup_candle_count` - which apparently i overlooked while reviewing.

It's not a problem for this PR - but something i'd like to see simplified.

> The reason we have config['freqai']['startup_candles'] is because startup_candle_count is limited by the max candles allowed by exchange.

You can always circumvent this if freqAI is enabled. That's not a problem since freqAI is part of freqtrade.
Assuming freqAI as external module is in my opinion wrong, and will lead to a lot of such technical debt / problems.

robcaulk: >logically (if startup_candle_count is assigned correctly) - it'd be 14 (since we use 14 startup candles)

Well no, we need 5080 startup_candles in our hypothetical scenario. What am I missing?

> it's way to complicated for the user, they'll just set it to 10.000 and hope it'll work - if not, increase some more

agreed, that was why I was writing an auto-datadownload for backtesting...but we know that also can introduce tangential problems :-). But yes, computing this automatically in load_freqai_model() is the best way to do it. 

I will add it to this PR. 

Thanks for the thorough review.

xmatthias: > Well no, we need 5080 startup_candles in our hypothetical scenario. What am I missing?

This was about `indicator_max_period_candles` - not about `startup_candles`.

`startup_candles` is (to me) something with a clearly defined formula where all points are known by the bot - which means it can (and should) be calculated internally (i think we both agree on that).

---

`indicator_max_period_candles` is the one that bothers me - in the hypothetical scenario above, it'd be set to 14, too (the maximum "window" used in the indicator) - so after 14 candles, no NAN values are present anymore.
`startup_candle_count` defines the same thing - in the same way (in candles).
the only reason to have `indicator_max_period_candles` is to circumvent limitations imposed by  `startup_candle_count` - but i don't see why these can't be disabled for freqAI bots (that'd be justified - and would remove duplicate configuration of the same value).

robcaulk: >the only reason to have indicator_max_period_candles is to circumvent limitations imposed by startup_candle_count 

It isn't true though, `indicator_max_period_candles` is required for FreqAI to know how many total startup_candles to automatically compute. The user input is the maximum period given to talib + `include_timeframes`. 

Can you answer this: how would we compute startup_candles without knowing the maximum period (`indicator_max_period`) going into talib + the maximum timeframe in `include_timeframes` ?

The reason I ask you to answer this question now is because when I make the automatic startup_candles function, it will *still* require the user to input a `indicator_max_period`. Otherwise there is no way to calculate the `startup_candle`(_count).

xmatthias: `startup_candle_count` needs to contain the maximum period going into talib (or any other indicator).
`indicator_max_period_candles` is the same - but for freqAI.

calculating `startup_candles` is using (based on the formula from the docs):
>  ( 4 hours * 25 max period * 60 minutes/hour + 30 day train_period_days * 1440 minutes per day ) / 5 min (base time frame) = 1488.

moving this into formula land again:
`(max_timeframes * startup_candle_count * 60 + train_period_days + 1440) / timeframe`.

if the 25 (in the sample) comes from `indicator_max_period_candles` - or from `startup_candle_count` does not matter.

freqtrade/freqai/data_kitchen.py
robcaulk: Fixed

freqtrade/freqai/data_kitchen.py
robcaulk: Fixed.

freqtrade/freqai/data_kitchen.py
robcaulk: I attempted to remove `indicator_max_period_candles` and `startup_candles` in favor of keeping the standard strategy set `startup_candle_count`. You should probably review it.

freqtrade/optimize/backtesting.py
```diff
@@ -244,6 +244,21 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
...
+        if not self.config['freqai']['enabled']:
+            return self.required_startup
+        else:
+            if not self.config['startup_candle_count']:
+                raise OperationalException('FreqAI backtesting module requires strategy '
+                                           'set startup_candle_count.')
+            tf_seconds = timeframe_to_seconds(timeframe)
+            train_candles = self.config['freqai']['train_period_days'] * 86400 / tf_seconds
+            # multiplied by safety factor of 2 because FreqAI users
+            # typically do not know the correct window.
```
xmatthias: :rofl: 
while i don't disagree with this assumption per se - i'm not sure we should have a code comment that refers to our users this way.

if we REALLY assume users don't know this - then 2x won't be enough (it'll be left at 30, while using SMA200) .

freqtrade/freqai/data_kitchen.py
```diff
@@ -1027,3 +1002,74 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
         if self.unique_classes:
             for label in self.unique_classes:
                 self.unique_class_list += list(self.unique_classes[label])
+
+# Methods called by interface.py (load_freqai_model())
+
+
+def download_all_data_for_training(timerange: TimeRange,
```
xmatthias: I would move the calculation of the "proper" timerange into this method - and to that as a loop in there.

The initial problem i pointed out in https://github.com/freqtrade/freqtrade/pull/7244#discussion_r948254871 is still there (although the discussion in that review drifted in another direction) - forcing us to download a LOT more than what is actually necessary, since we use the start-date from the highest timeframe - which could be 365 days  (so we download 1y of 1m data) - while we only would need 365 minutes of 1m data.

freqtrade/optimize/backtesting.py
```diff
@@ -244,6 +244,21 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
...
+            total_candles = self.required_startup * 2 + train_candles
```
xmatthias: I do wonder why mypy ain't flagging this - `train_candles` is a float (after the above division) - so the result of this will also be a float.

While it's not a huge problem - i'm also not sure all data-providers will handle numbers with a decimal part gracefully in all instances.
i'd simply conver it explicitly to int in this line - that'll also ensure the output is not 200.0123 candles ...

freqtrade/data/dataprovider.py
```diff
@@ -92,7 +92,7 @@ def historic_ohlcv(
                 'timerange') is None else str(self._config.get('timerange')))
             # Move informative start time respecting startup_candle_count
             timerange.subtract_start(
-                timeframe_to_seconds(str(timeframe)) * self._config.get('startup_candle_count', 0)
+                self.get_required_startup_seconds(str(timeframe))
```
xmatthias: the explicit str is unnecessary here.
we expect "5m" for example - if it ain't a string, it'll fail anyway - the only difference will be if it fails here - or from within timeframe_to_seconds.

docs/freqai.md
```diff
@@ -113,7 +113,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
 | `stratify_training_data` | This value is used to indicate the stratification of the data. e.g. 2 would set every 2nd data point into a separate dataset to be pulled from during training/testing. <br> **Datatype:** positive integer.
-| `indicator_max_period_candles` | The maximum *period* used in `populate_any_indicators()` for indicator creation. FreqAI uses this information in combination with the maximum timeframe to calculate how many data points it should download so that the first data point does not have a NaN <br> **Datatype:** positive integer.
+| `indicator_max_period_candles` | **Deprecated in favor of** strategy set `startup_candle_count`, however, both configuration parameters provide the same functionality; the maximum *period* used in `populate_any_indicators()` for indicator creation (timeframe independent). FreqAI uses this information in combination with the maximum timeframe to calculate how many data points it should download so that the first data point does not have a NaN <br> **Datatype:** positive integer.
```
xmatthias: that's a ... dangerous (and wrong) statement.

`indicator_max_period_candles` is no longer referenced in the code - so it's not deprecated, but removed - and does not / no longer provide the same functionality.

freqtrade/data/dataprovider.py
```diff
@@ -105,6 +105,17 @@ def historic_ohlcv(
...
+    def get_required_startup_seconds(self, timeframe: str) -> int:
+        tf_seconds = timeframe_to_seconds(timeframe)
+        base_seconds = tf_seconds * self._config.get('startup_candle_count', 0)
+        if not self._config['freqai']['enabled']:
+            return base_seconds
+        else:
+            train_seconds = self._config['freqai']['train_period_days'] * 86400
+            # multiplied by safety factor of 2 because FreqAI users
+            # typically do not know the correct window.
+            return base_seconds * 2 + int(train_seconds)
```
xmatthias: i wonder if this could be combined with `get_required_startup()` from backtesting.
They both use the same logic - with the only difference being that this returns seconds, while this returns candles - so maybe moving the function from backtesting here, and calling that could reduce code duplication?

freqtrade/optimize/backtesting.py
```diff
@@ -244,6 +244,21 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
         self.progress.set_new_value(1)
         return data, self.timerange
 
+    def get_required_startup(self, timeframe: str) -> int:
+        if not self.config['freqai']['enabled']:
+            return self.required_startup
+        else:
+            if not self.config['startup_candle_count']:
+                raise OperationalException('FreqAI backtesting module requires strategy '
+                                           'set startup_candle_count.')
```
xmatthias:
```suggestion
                raise OperationalException('FreqAI backtesting module requires the strategy '
                                           'to set startup_candle_count.')
```

docs/freqai.md
```diff
@@ -113,7 +113,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
 | `indicator_periods_candles` | A list of integers used to duplicate all indicators according to a set of periods and add them to the feature set. <br> **Datatype:** list of positive integers.
```
xmatthias: I think we should have a check (or automatism to adjust `startup_candle_count`) for this.

`startup_candle_count` MUST be bigger or equal to `indicator_periods_candles`.
If i look at the sample strategy - changing this value in the (sample) config from 20 to 50 would break backtesting, as `startup_candle_count` would still remain at 20.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1027,3 +1002,74 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
         if self.unique_classes:
             for label in self.unique_classes:
                 self.unique_class_list += list(self.unique_classes[label])
+
+# Methods called by interface.py (load_freqai_model())
```
xmatthias: I think it might make sense to create a new file for these 2 methods.

that'll keep data_kitchen clean - and focused on the class without containing additional non-class methods.
i know we do this in other cases (exchange) too - but it's not a good pattern - and one i'm trying to get rid of...

xmatthias (commented):
There's ... more comments here than i expected...
but there's nothing really serious, most things are suggestions for improvements.

freqtrade/data/dataprovider.py
robcaulk: Hah, I wondered the same thing but I just kept that code unchanged - unsure why it is originally doing that. Any ideas why?

freqtrade/freqai/data_kitchen.py
robcaulk: Leave it to another PR for now, please. :-). Let's sort out this backtesting stuff...it is, not so straight forward already as you can see.

freqtrade/data/dataprovider.py
xmatthias: not really - but code changes over time - so most likely, this was initially a non-str input ...

freqtrade/optimize/backtesting.py
```diff
@@ -244,6 +244,21 @@ def load_bt_data(self) -> Tuple[Dict[str, DataFrame], TimeRange]:
         self.progress.set_new_value(1)
         return data, self.timerange
 
+    def get_required_startup(self, timeframe: str) -> int:
+        if not self.config['freqai']['enabled']:
```
xmatthias: This must be `self.config.get('freqai', {}).get('enabled')`.
otherwise it's crashing with keyerror when you run with a "non-freqAI" enabled configuration.

freqtrade/freqai/data_kitchen.py
robcaulk: Resolving this convo to get it out of the way in the PR conversation.

freqtrade/data/dataprovider.py
xmatthias: leave the `str()` there for now - it's needed for mypy.

freqtrade/optimize/backtesting.py
robcaulk: Ah I hit this during pytesting :-)

freqtrade/data/dataprovider.py
robcaulk: I combined them and removed the redundant code from both `backtesting.py` and `data_kitchen.py` :-)

freqtrade/optimize/backtesting.py
robcaulk: Yeah, better to be consistent and trust the users with their own choices. I removed the comment *and* the factor of safety.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1027,3 +1002,53 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
...
+
+    for tf in config["freqai"]["feature_parameters"].get("include_timeframes"):
+        timerange = TimeRange()
+        timerange.startts = int(time)
+        timerange.stopts = int(time)
+        startup_candles = dp.get_required_startup(str(tf))
+        tf_seconds = timeframe_to_seconds(str(tf))
+        timerange.subtract_start(tf_seconds * startup_candles)
+        new_pairs_days = int((timerange.stopts - timerange.startts) / SECONDS_IN_DAY)
+        # FIXME: now that we are looping on `refresh_backtest_ohlcv_data`, the function
```
robcaulk, line 1041: @xmatthias This is an issue we need to solve if we want to auto-download data heterogeneously.

freqtrade/freqai/data_kitchen.py
robcaulk: I guess I decided to do it in this one. I resolve this conversation in favor of the other one I started regarding the same topic.

docs/freqai.md
robcaulk: Agreed - thanks for pointing that out.

docs/freqai.md
robcaulk: Yup, good idea.

freqtrade/freqai/data_kitchen.py
robcaulk: Done

freqtrade/freqai/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+    for tf in config["freqai"]["feature_parameters"].get("include_timeframes"):
+        timerange = TimeRange()
+        timerange.startts = int(time)
+        timerange.stopts = int(time)
+        startup_candles = dp.get_required_startup(str(tf))
+        tf_seconds = timeframe_to_seconds(str(tf))
+        timerange.subtract_start(tf_seconds * startup_candles)
+        new_pairs_days = int((timerange.stopts - timerange.startts) / 86400)
+        # FIXME: now that we are looping on `refresh_backtest_ohlcv_data`, the function
```
robcaulk: @xmatthias This is an issue we need to solve if we want to auto-download data heterogeneously.

config_examples/config_freqai.example.json
```diff
@@ -9,6 +9,7 @@
     "dry_run": true,
     "timeframe": "3m",
     "dry_run_wallet": 1000,
+    "startup_candle_count": 20,
```
xmatthias: Dangerous - having this in the configuration means whatever you set in the strategy is ignored (config > strategy) ...

freqtrade/freqai/utils.py
```diff
@@ -0,0 +1,138 @@
...
+                                    and training the model.
+    :param dp: DataProvider instance attached to the strategy
+    """
+
+    if dp._exchange is not None:
+        markets = [p for p, m in dp._exchange.markets.items() if market_is_active(m)
+                   or config.get('include_inactive')]
+    else:
+        # This should not occur:
+        raise OperationalException('No exchange object found.')
```
xmatthias: if you reverse the condition - you can raise OperationalError` only once (in the if branch) and omit the else (and the check again further below).

That check will then be valid throughout the method.

freqtrade/freqai/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    time = datetime.now(tz=timezone.utc).timestamp()
+    data_load_timerange = TimeRange()
+
+    timeframes = config["freqai"]["feature_parameters"].get("include_timeframes")
+
+    max_tf_seconds = 0
+    for tf in timeframes:
+        secs = timeframe_to_seconds(tf)
+        if secs > max_tf_seconds:
+            max_tf_seconds = secs
```
xmatthias, line 65: if you wanna be fancy - we could also do this via the following comprehension

```suggestion
    max_tf_seconds = max(timeframe_to_seconds(tf) for tf in timeframes)
```

Not sure if that really improves readability though ...

freqtrade/freqai/utils.py
```diff
@@ -0,0 +1,138 @@
...
+    logger.info(f'FreqAI auto-downloader using {max_period} startup candles.')
+
+    additional_seconds = max_period * max_tf_seconds
+
+    data_load_timerange.startts = int(
+        time
+        - config["freqai"].get("train_period_days", 0) * 86400
+        - additional_seconds
+    )
+    data_load_timerange.stopts = int(time)
```
xmatthias: You can't really (reliably) assign a timeRange object like this.

wherever we use this - we're checking for `timerange.starttype == 'date'` (or at least we should) - and by not assigning that, it might mean that it'll not use the timerange given.

You can instead use `Timerange('date', 'date', startts, stopts)` to create the instance once the values are clear (instead of creating an empty instance and then reassigning attributes).

tests/freqai/test_freqai_backtesting.py
```diff
@@ -48,10 +48,10 @@ def test_freqai_backtest_load_data(freqai_conf, mocker, caplog):
...
-    del freqai_conf['freqai']['startup_candles']
-    backtesting = Backtesting(freqai_conf)
-    with pytest.raises(OperationalException,
-                       match=r'FreqAI backtesting module.*startup_candles in config.'):
-        backtesting.load_bt_data()
+    # del freqai_conf['freqai']['startup_candles']
+    # backtesting = Backtesting(freqai_conf)
+    # with pytest.raises(OperationalException,
+    #                    match=r'FreqAI backtesting module.*startup_candles in config.'):
+    #     backtesting.load_bt_data()
```
xmatthias: i guess this can now be removed?

freqtrade/freqai/utils.py
xmatthias: don't see that as a very big problem though.
there's little preventing us from adding a new param to that method to skip funding rates all together - but then it'll remain the callers responsibility to download them separately.

xmatthias:
Let me know if you think this is good to go - or if you think there's something wrong with the changes i made in the last few commits (i implemented the PR feedback, so we can merge this ASAP, having the most possible time in "develop").

robcaulk:
Thanks for doing that - just ran my "peace of mind" checks. It is as ready as it's ever going to be - let's set it free on develop. I will make announcements in the discord to help people understand that `indicator_max_periods_candles` no longer exists.

xmatthias (approved):
LGTM :+1:

PR #7258, opened 2022-08-19
Add flag to move hyperopt populate_indicators to epoch

### LINKED ISSUE #7080: Add a switch in hyperopt to compute indicators at each epoch instead of at the start
## Describe your environment

  * Operating system: Gentoo Linux 5.10.61
  * Python Version: 3.10.15
  * CCXT version: 1.89.14
  * Freqtrade Version: 2022.6

## Describe the enhancement

I've developed a strategy that makes use of some indicators (expecially MACD, which has 3 parameters: meaning there are 3 nested loops in `populate_indicators`) and informative pairs (with other indicators).

When I run hyperopt, it quickly wipes out my entire RAM + swap (16GB + 20GB) and forces Linux to kill freqtrade's processes. I can't even run it with `-j 1` for more than a bunch of pairs or too large timerange. Anyway, using `-j 1`, of course, slows down the process (balancing out the speedup that comes from precomputing all indicators before starting the first epoch). Moreover, I think that accessing a very large dataframe is probably (not verified - just an intuition) much more slower than accessing a small dataframe (at least for the principle of locality, I think): so creating a large dataframe in `populate_indicators` may slows down the entire hyperopt in complex strategies.

Actually, I've come to a workaround: compute indicators in `populate_entry/exit_trend` instead of in `populate_indicators` (I compute indicators 2 times: one for entry and one for exit, but just for the needed values instead of the entire ranges of values). This is the only way I can run my strategy. Memory consumption is negligible, and execution time seems to be lower (again: not verified numerically, it just seems faster to me - probably due to the reduced access time with a smaller dataframe and in the parallelization of the computation of indicators, since `populate_*_trend` are executed in parallel).

So, I would like to have a switch in the `hyperopt` command to make the call to `populate_indicators` at each epoch instead of computing all indicators at the start. At least, this allows to hyperopt more complex strategies that otherwise would consume too much memory (sometimes even with `-j 1`) or force the user to use only one core slowing down the entire process.

xmatthias:
## Summary

Add flag to move hyperopt populate_indicators to epoch
This can have memory benefits - but will be heavier on the CPU.

closes #7080

## Quick changelog

- allow moving of `populate_indicators()` to the start of each Epoch.
- Automatically disable special hyperopt `.range` behavior - returning only 1 value in this mode.

### DIFF
--- a/docs/hyperopt.md
+++ b/docs/hyperopt.md
@@ -432,10 +432,10 @@ While this strategy is most likely too simple to provide consistent profit, it s
     `range` property may also be used with `DecimalParameter` and `CategoricalParameter`. `RealParameter` does not provide this property due to infinite search space.
 
 ??? Hint "Performance tip"
-    By doing the calculation of all possible indicators in `populate_indicators()`, the calculation of the indicator happens only once for every parameter.  
-    While this may slow down the hyperopt startup speed, the overall performance will increase as the Hyperopt execution itself may pick the same value for multiple epochs (changing other values).
-    As this also has Performance implications, hyperopt provides `--analyze-per-epoch` - which will move the execution of `populate_indicators()` to the epoch process. This will implicitly also change the `.range` functionality to only return the actually used value.
-    You should however try to use space ranges as small as possible. 
+    During normal hyperopting, indicators are calculated once and supplied to each epoch, linearly increasing RAM usage as a factor of increasing cores. As this also has performance implications, hyperopt provides `--analyze-per-epoch` which will move the execution of `populate_indicators()` to the epoch process, calculating a single value per parameter per epoch instead of using the `.range` functionality. In this case, `.range` functionality will only return the actually used value. This will reduce RAM usage, but increase CPU usage. However, your hyperopting run will be less likely to fail due to Out Of Memory (OOM) issues.
+
+    In either case, you should try to use space ranges as small as possible this will improve CPU/RAM usage in both scenarios.
+
 
 ## Optimizing protections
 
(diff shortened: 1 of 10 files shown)


froggleston:
Looks good. I'm not sure ```As this also has Performance implications, hyperopt provides `--analyze-per-epoch` - which will move the execution of `populate_indicators()` to the epoch process``` is as clear as it could be though? What does this actually mean in simple terms? Perhaps provide a quick example or diagram?

xmatthias:
> What does this actually mean in simple terms

well what it says - if you don't remove it from context, but see it as part of the "note" box it's part of ;) 

hyperopt runs `populate_indicators()` once before all epochs - and then reuses this data. 
if you use `.range` - then we'll create all indicator columns and reuse that (e.g. ema5,ema6,ema7, ...).

if the new option is given, it'll run once every epoch - obviously causing a lot more CPU load - while at the same time reducing memory consumption, as only the "necessary" ema is calculated (`.range` returns 1 value)

froggleston:
How about:

```
During normal hyperopting, indicators are calculated once and supplied to each epoch, linearly increasing RAM usage as a factor of increasing cores. As this also has performance implications, hyperopt provides `--analyze-per-epoch` which will move the execution of `populate_indicators()` to the epoch process, calculating a single value per parameter per epoch instead of using the `.range` functionality. In this case, `.range` functionality will only return the actually used value. This will reduce RAM usage, but increase CPU usage. However, your hyperopting run will be less likely to fail due to Out Of Memory (OOM) issues.

In either case, you should try to use space ranges as small as possible as in either situation, this will improve CPU/RAM usage.
```

xmatthias:
@froggleston Thanks a lot - using exactly this wording!

PR #7260, opened 2022-08-19
Example FreqAI hybrid strategy

JohanVlugt:
## Summary

Example strat showing how a typical FT user can augment a normal strategy with a FreqAI indicator (a classifier for up/down prediction).

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/config_examples/config_freqai.example.json
+++ b/config_examples/config_freqai.example.json
@@ -75,7 +75,6 @@
             "weight_factor": 0.9,
             "principal_component_analysis": false,
             "use_SVM_to_remove_outliers": true,
-            "stratify_training_data": 0,
             "indicator_max_period_candles": 20,
             "indicator_periods_candles": [10, 20]
         },
--- a/freqtrade/templates/FreqaiHybridExampleStrategy.py
+++ b/freqtrade/templates/FreqaiHybridExampleStrategy.py
@@ -257,6 +257,3 @@ class FreqaiExampleHybridStrategy(IStrategy):
             'exit_short'] = 1
 
         return df
-
-    def get_ticker_indicator(self):
-        return int(self.config["timeframe"][:-1])
--- a/freqtrade/templates/FreqaiHybridExampleStrategy.py
+++ b/freqtrade/templates/FreqaiHybridExampleStrategy.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 import talib.abstract as ta
 from pandas import DataFrame
+from technical import qtpylib
 
 from freqtrade.strategy import IntParameter, IStrategy, merge_informative_pair
 
@@ -48,7 +49,7 @@ class FreqaiExampleHybridStrategy(IStrategy):
             "indicator_periods_candles": [10, 20]
         },
         "data_split_parameters": {
-            "test_size": 0.33,
+            "test_size": 0,
             "random_state": 1
         },
         "model_training_parameters": {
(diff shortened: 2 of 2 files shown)


robcaulk (approved):

freqtrade/templates/FreqaiHybridExampleStrategy.py
```diff
@@ -0,0 +1,337 @@
...
+                 **kwargs) -> float:
+
+        return 1
+
+    """
+    Supertrend Indicator; adapted for freqtrade, optimized by the math genius.
+    from: Perkmeister#2394
+    """
+
+    def supertrend(self, dataframe: DataFrame, multiplier, period):
```
xmatthias: Sample strategies should not implement indicators via copy/paste. That's bad practice for any strategy - so these strategies clearly should be leading by example in implementing best practices by reusing existing indicator implementations (this also allows for easier fixes should problems/bugs in an implementation be discovered).

supertrend is available in multiple libraries, including technical, pandas-ta (both of which freqtrade depends on) - so our samples should make use of that.

config_examples/config_freqai.example.json
```diff
@@ -75,7 +75,6 @@
             "weight_factor": 0.9,
             "principal_component_analysis": false,
             "use_SVM_to_remove_outliers": true,
-            "stratify_training_data": 0,
```
xmatthias, line 78: is the removal of this intentional?

robcaulk: Yes, setting it to 0 is equivalent to not setting it at all. I like the idea of keeping the example config as simple as possible.

freqtrade/templates/FreqaiHybridExampleStrategy.py
robcaulk: Fair point. I guess the reason this was done is because this supertrend function is far more efficient than the standard libs. 

But an example is an example, we will remove the function and use pandas-ta instead.

xmatthias: if it is (and results are still identical) - it should be contributed to one of the mentioned libraries.

xmatthias: What's worse - this implementation seems incorrect - or at least miss-aligned to how tradingview and pandas_ta calculate supertrend.

results are similar - but not identical (using the same parameters, obviously) - so while it's faster - results may be different to what you'd expect.
now tradingview is not the gold standard - but i do trust the implementation of pandas_ta more than a random implementation we see here - which was (most likely) sourced from an issue comment someplace.

freqtrade/templates/FreqaiHybridExampleStrategy.py
```diff
@@ -0,0 +1,262 @@
...
+                (df['tema'] <= df['bb_middleband']) &
+                (df['tema'] > df['tema'].shift(1)) &  # Guard: tema is raising
+                (df['volume'] > 0)  # Make sure Volume is not 0
+            ),
+            'exit_short'] = 1
+
+        return df
+
+    def get_ticker_indicator(self):
+        return int(self.config["timeframe"][:-1])
```
xmatthias: Can we remove this?
It doesn't seem to be used - and it's faulty code.

it'll return `1` - no matter if the timeframe is 1m, 1h or 1d ... so it'll not work properly once you start changing timeframes around

robcaulk: Done

freqtrade/templates/FreqaiHybridExampleStrategy.py
robcaulk: Thanks for checking on that. Resolved it by removing supertrend all together.

PR #7263, opened 2022-08-20
Okx cache tiers

xmatthias:
## Summary

Okx tier initialization takes too long to effectively backtest.
tiers will not change this often, so we can cache them (for the moment, for at least 1 day).


froggleston:
I use about 130 pairs and haven't found this a problem, but sure, any reduction in my 12 minute backtests would be good! :D

xmatthias:
> I use about 130 pairs and haven't found this a problem, but sure, any reduction in my 12 minute backtests would be good! :D

i had to do 10 backtests in a row today - with a backtest that takes normally 10 seconds ... waiting a minute IS a problem :laughing:

PR #7266, opened 2022-08-20
Extra filename option when loading strategy params

emiliobasualdo:
NO tests were added
What do you guys think?
## Summary
Load strategy params from `strategy-class-name.json` file

## What's new?

This allows us to create small strategy variations in the same strategy file, while loading params from different .json files
ex:
`SampleStrategy.py`
```python
class SampleStrategy1(IStrategy):
    static_param_1 = 0.01
    ...
    ...
  
class SampleStrategy2(SampleStrategy1):
    static_param_1 = 0.02

class SampleStrateg3(SampleStrategy1):
    static_param_1 = 0.03
```
Would allow for this:
```
/strategies
 |
 |- SampleStrategy.py
 |- SampleStrategy1.json
 |- SampleStrategy1.json
 |- SampleStrategy3.json
```
Instead of this:
```
/strategies
 |
 |- SampleStrategy1.py
 |- SampleStrategy2.py
 |- SampleStrategy3.py
 |- SampleStrategy1.json
 |- SampleStrategy1.json
 |- SampleStrategy3.json
```

### DIFF
--- a/freqtrade/strategy/hyper.py
+++ b/freqtrade/strategy/hyper.py
@@ -112,10 +112,13 @@ class HyperStrategyMixin:
         self._ft_load_params(protection_params, 'protection', hyperopt)
 
     def load_params_from_file(self) -> Dict:
-        filename_str = getattr(self, '__file__', '')
-        if not filename_str:
-            return {}
+        filename_str = self.__class__.__name__
         filename = Path(filename_str).with_suffix('.json')
+        if not filename.is_file():
+            filename_str = getattr(self, '__file__', '')
+            if not filename_str:
+                return {}
+            filename = Path(filename_str).with_suffix('.json')
 
         if filename.is_file():
             logger.info(f"Loading parameters from file {filename}")


xmatthias:
I'm not sure what benefit this will give you over multiple strategy files.
Sure you'd have fewer files - but if you're _a bit_ careless about naming your strategies, it'll become heavily confusing as to which file actually belongs to which strategy - as the strategy name and the filename can be something completely different.

Imagine the below setup (not impossible with this change) - and tell me where `SampleStrategy1.json` belongs to ... 

```
/strategies
 |
 |- hello_world.py
 |- SampleStrategy1.json
 |- SampleStrategy1.json
 |- SampleStrategy3.json
 |- SMACRoss.py
```

---

The implementation (aside from tests and docs) is clearly incomplete - as we'd want this to apply globally (also for writing parameter files) - but it's showing what you're after quite clearly.


Your change is however currently taking a few (risky) assumptions:
* If the strategy file is NOT named as the base strategy (`hello_world.py` - containing `SampleStrategy`) - the files don't belong together anymore
* if `SampleStrategy2.py` does exist - but contains `SampleWhateverStrategy` - we'd override this strategy's parameter file (assuming writing works the same way) - breaking a completely unrelated strategy.


I think if we do this, it should be either
* opt in (explicitly define the filename as part of the strategy subclass - default to current behavior?)
* append the strategy-name to the filename (so the filename in your sample would be ` SampleStrategy_SampleStrategy1.json`
* both of the above - allowing a `paramfile` - which is appended (`SampleStrategy_BestParams1.json`)

---

Personally, i think a `paramfile` parameter (which is appended to the strategy filename) might make the most sense (basically combining the 2 options above), as it'd allow further possibilities with this.

This could eventually be exposed via the configuration / cli - so you could have multiple parameter files (For the same strategy) - and run 3 bots in parallel without changing the strategy (same strategy/strategy name) - purely by using different parameter files - and as such, test the results of the best 3 epochs of a hyperopt run.

xmatthias:
Going to close this - as it's seen no progress in several months.

The feedback above is still valid - so i'm more than happy to get this either reopened or re-submitted if we see this progressing further.

PR #7282, opened 2022-08-22
Plug mem leak, add training timer

robcaulk:
## Summary

The `dk.model_filename` string contains the last trained timestamp, which is useful for creating folder/files on disk. However, we also use this for adding entries to our `dd.model_dictionary` which *should* only hold the newest models in RAM (to improve inference performance). However, FreqAI was using a unique key to add each entry - thus resulting in an ever expanding dictionary:

`self.model_dictionary[dk.model_filename] = model`

The solution was quite simple:

`self.model_dictionary[coin] = model`

We use this opportunity to include the training timer, which was constructed to help aid in the investigation of the memleak. This `training_timer()` computes the sum of time spent training one full loop through the pair whitelist. It logs the information for the user to see.

### DIFF
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -421,7 +421,7 @@ class FreqaiDataDrawer:
             )
 
         # if self.live:
-        self.model_dictionary[dk.model_filename] = model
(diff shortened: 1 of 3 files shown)


xmatthias (approved):
Great find! :+1:

PR #7286, opened 2022-08-23
runmode has type `freqtrade.enums.runmode.RunMode` but is used as type `None`

luca-digrazia:
**"filename"**: "freqtrade/configuration/configuration.py"
**"warning_type"**: "Incompatible variable type [9]",
**"warning_message"**: " runmode is declared to have type `freqtrade.enums.runmode.RunMode` but is used as type `None`.",
**"warning_line"**: 31
**"fix"**: Add Optional type

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -28,7 +28,7 @@ class Configuration:
     Reuse this class for the bot, backtesting, hyperopt and every script that required configuration
     """
 
-    def __init__(self, args: Dict[str, Any], runmode: RunMode = None) -> None:
+    def __init__(self, args: Dict[str, Any], runmode: Optional[RunMode] = None) -> None:
         self.args = args
         self.config: Optional[Dict[str, Any]] = None
         self.runmode = runmode


xmatthias:
Which tool is issuing this as warning?

mypy (and pyright for pylance) have been configured to not issue this warning (more details here: https://github.com/python/mypy/issues/12869).
As you're seeing this as warning, i assume you're using a different tool - and which would have to be configured to match the tools we do use.

IF we were to configure this explicitly (no implicit Optional for function arguments) - it would have to be throughout the repository - i do like the implicit method more though - - as it'd be 100ds of `Optional` (just in the exchange class) that would have to be added throughout the code - which makes the function definitions quite long and hard to read.

luca-digrazia:
> Which tool is issuing this as warning?
> 
> mypy (and pyright for pylance) have been configured to not issue this warning (more details here: [python/mypy#12869](https://github.com/python/mypy/issues/12869)). As you're seeing this as warning, i assume you're using a different tool - and which would have to be configured to match the tools we do use.
> 
> IF we were to configure this explicitly (no implicit Optional for function arguments) - it would have to be throughout the repository - i do like the implicit method more though - - as it'd be 100ds of `Optional` (just in the exchange class) that would have to be added throughout the code - which makes the function definitions quite long and hard to read.

Hi, thanks for the message. It is a warning from Pyre. I think in this case `Optional` is enough in the method argument, then you don't have to update the same variable again.

xmatthias (changes requested):
we're not using pyre for freqtrade at the moment - and 
If you run pyre against the whole repository, you'll find 10s (72 to be precise) of the same warnings - so this PR is highly incomplete, fixing one instance of an error that actually appears >50 times.

I'm not sure on such a change - as the benefit it'd provide is basically non-existant (mypy will still detect this correctly) - and on functions with many arguments, this will reduce readability quite some by having the code cluttered by `Optional[]`.
Also, the current approach is in line with current mypy default behavior (the type-checker freqtrade relies on).

---

A better approach will be to align the pyre setup  to add configuration for the "implicit optional" flag in pyre to mirror mypy's default behavior (getting these errors can be accomplished with `mypy freqtrade --no-implicit-optional` (details see in the link above) - assuming such a thing exists for pyre.

luca-digrazia:
ok, got it.

xmatthias:
As mentioned above - as is this is a pointless fix.

If we assume this to be the correct way to type arguments with default None type - this would need to be fixed several more cases, as this same warning will come up several more times.

pyre also seems to flag a lot of other non-issues / false positives - especially around sqlalchemy objects - so it either requires extensive configuration, or is not very mature in collaborating with sqlalchemy.

As we're not using pyre but mypy though - we'll not be investing time into this now.

PR #7296, opened 2022-08-26
Change MinPts calculation in DBSCAN

th0rntwig:
Changed the calculation of minimum points (MinPts) to use for nearest neighbor calculation.

DBSCAN clusters data using two parameters, MinPts and epsilon. MinPts is suggested to be 2x nmb features. But, in cases where the user has constructed their feature set with nmb features < nmb time points or 2x nmb features >= nmb time points, MinPts cannot be calculated this way. 
This PR changes the calculation of MinPts to be 0.25x nmb time points.


xmatthias:
Please familiarize yourself with the [developer setup](https://www.freqtrade.io/en/latest/developer/#developer-setup) part of the documentation. 

it'll allow you to run the tools we use in CI locally, avoiding CI failures and unnecessary delays due to failing CI.

Please understand that we cannot accept PR's with failing CI - and it is the task of the author to ensure CI fails.
in some cases, this may require assistance (which is no problem, just ask for help) - but in 95% of the cases, it's either flake8 or mypy that's failing, which can easily be detected locally.

Thanks

robcaulk (approved):
LGTM

PR #7297, opened 2022-08-26
Conception problem with the total_profit

richardjozsa:
I think here is a conception problem with the total profit, if we `*=` it with the last PnL it means we assume that the last profit effects on this one. In real life it depends on the user eg stake amount can be unlimited.
However the agent will learn better and more if we just add it with `+=`.

### DIFF
--- a/freqtrade/freqai/RL/Base4ActionRLEnv.py
+++ b/freqtrade/freqai/RL/Base4ActionRLEnv.py
@@ -292,7 +292,8 @@ class Base4ActionRLEnv(gym.Env):
             pnl = self.get_unrealized_profit()
 
             if self._position in (Positions.Long, Positions.Short):
-                self._total_profit *= (1 + pnl)
+                #self._total_profit *= (1 + pnl)
+                self._total_profit += pnl
                 self._profits.append((self._current_tick, self._total_profit))
                 self.close_trade_profit.append(pnl)
 
--- a/freqtrade/freqai/RL/Base5ActionRLEnv.py
+++ b/freqtrade/freqai/RL/Base5ActionRLEnv.py
@@ -295,7 +295,8 @@ class Base5ActionRLEnv(gym.Env):
             pnl = self.get_unrealized_profit()
 
             if self._position in (Positions.Long, Positions.Short):
-                self._total_profit *= (1 + pnl)
+                #self._total_profit *= (1 + pnl)
+                self._total_profit += pnl
                 self._profits.append((self._current_tick, self._total_profit))
                 self.close_trade_profit.append(pnl)
 


freqtrade/freqai/RL/Base4ActionRLEnv.py
```diff
@@ -292,7 +292,8 @@ def _update_profit(self, action):
             pnl = self.get_unrealized_profit()
 
             if self._position in (Positions.Long, Positions.Short):
-                self._total_profit *= (1 + pnl)
+                #self._total_profit *= (1 + pnl)
+                self._total_profit += pnl
```
robcaulk, line 296: So I agree with this change, but I think we need to check if the user wants to compound or not based on the `self.config['stake_amount'] `.

robcaulk (changes requested):

freqtrade/freqai/RL/Base4ActionRLEnv.py
xmatthias: `"stake_amount"` - maybe? 

i'm not sure why this is relevant though - that's relevant to open new positions (is my next position equally sized, or slightly bigger or smaller).

Assuming `pnl` is the absolute profit though - i don't see why this is relevant here.
if it's relative profit - it's wrong anyway (in both forms) , as you can't aggregate percentage values with different stakes.

robcaulk: Summing the pnl would assume each trade stake is 1. Which is how the environment is designed. This is why I had it compounding originally - it used the full "trade wallet" (but only for one coin which would be the `stake_amount`). That said...

Looking closer, it is also just plain wrong to `update_profit` this way at *each training step*. This should only be updated when a training trade is closed:

```python 
      if self._position in (Positions.Long, Positions.Short) and action in (Actions.Long_exit.value, Actions.Short_exit.value):
              self._total_profit *= (1 + pnl)
              #self._total_profit *= (1 + pnl)
              self._total_profit += pnl
  ```

richardjozsa: But PnL itself is unrealized profit, so that wouldn't be to right if we just update it when the trade is closed. Maybe we need another function which is update it when the exit signal is formed, but until that because we use unrealized profit here that is right.

robcaulk: Well we can't be summing unrealized profit.

Assume a trade goes perfectly, every tick you sum the unrealized profit at that point during the trade. In the end you add the full profit from the entire trade. It is not correct.

robcaulk: Another bug was located in this function - it only updates profit when the action is a trade entry. Also incorrect.

I have corrected this two bugs, added compounding detection, and refactored the environment inheritance tree in another commit. Please check to ensure you still agree with `_update_profit()`, and make a PR if you feel it is still incorrect. But I am closing this PR since the other commit I made covers this and a lot more.

PR #7304, opened 2022-08-28
Reduce image sizes in freqai doc

th0rntwig:
The file sizes of the images in the FreqAI doc were too large. This PR has reduced image sizes.


robcaulk (approved):
Thanks!

PR #7315, opened 2022-08-29
allow users to properly reverse train-test data ordering

robcaulk:
There are some cases where users may want to train up until the most recent candle, but still avoid over fitting using an evaluation data set. In this case, they can experiment with reversing the train-test data ordering.

New parameter: `reverse_train_test_ordering`

False (typical way):

`shuffle` = False
`train_period_days` = 10
`test_size` = 0.5

`train_features` -> day 1-5 
`test_features` -> day 4-10

where day 10 is the current day. 

True: (unorthodox way enabled with this PR)

`train_features` -> day 4-10 
`test_features` -> day 1-5
 
where day 1 is 10 days ago.


docs/freqai.md
```diff
@@ -114,6 +114,7 @@ Mandatory parameters are marked as **Required**, which means that they are requi
 | `svm_params` | All parameters available in Sklearn's `SGDOneClassSVM()`. See details about some select parameters [here](#removing-outliers-using-a-support-vector-machine-svm). <br> **Datatype:** Dictionary.
 | `use_DBSCAN_to_remove_outliers` | Cluster data using DBSCAN to identify and remove outliers from training and prediction data. See details about how it works [here](#removing-outliers-with-dbscan). <br> **Datatype:** Boolean. 
 | `outlier_protection_percentage` | If more than `outlier_protection_percentage` fraction of points are removed as outliers, FreqAI will log a warning message and ignore outlier detection while keeping the original dataset intact. <br> **Datatype:** float. Default: `30`
+| `reverse_train_test_order` | If true, FreqAI will train on the latest data split and test on historical split of the data. This allows the model to be trained up to the most recent data point, while avoiding overfitting. However, users should be careful to understand unorthodox nature of this parameter before employing it. <br> **Datatype:** bool. Default: False
```
xmatthias, line 117: > However, users should be careful to understand unorthodox nature of this parameter before employing it.

wouldn't you want to explain this somehow / somewhere?
now it's kindof saying "you need to understand what this does - but i'm not gonna tell you in detail" ...

robcaulk: ...I explain precisely what it does here. 

This is not an ML book, it is a generalized toolbox. This is a perfectly sufficient explanation to help them decide if they should google more then experiment with the tools we provide, or back off and let it be.

xmatthias (approved):

PR #7321, opened 2022-08-30
Remove outlier protection from DI 

th0rntwig:
Outlier protection (to keep outlier-identifying algorithms from discarding full datasets) was incorrectly applied to the DI. This bug has now been corrected.


robcaulk (approved):
LGTM

PR #7322, opened 2022-08-30
Add inlier metric

robcaulk:
A custom ML feature for FreqAI users only, the "inlier metric."

The `inlier_metric` is a metric aimed at quantifying how different a prediction data point is from the most recent historic data points. 

User can set `inlier_metric_window` to set the look back window. FreqAI will compute the distance between the present prediction point and each of the previous data points (total of `inlier_metric_window` points). 

This function goes one step further - during training, it computes the `inlier_metric` for all training data points and builds weibull distributions for each each lookback point. The cumulative distribution function for the weibull distribution is used to produce a quantile for each of the data points. The quantiles for each lookback point are averaged to create the `inlier_metric`. 

FreqAI adds this `inlier_metric` score to the training features! In other words, your model is trained to recognize how this temporal inlier metric is related to the user set labels. 

This function does **not** remove outliers from the data set.


docs/freqai.md
```diff
@@ -115,6 +115,8 @@ Mandatory parameters are marked as **Required**, which means that they are requi
...
+| `inlier_metric_window` | If set, FreqAI will add the `inlier_metric` to the training feature set and set the lookback to be the `inlier_metric_window`. Details of how the `inlier_metric` is computed can be found [here](#using-the-inliermetric) <br> **Datatype:** int. Default: 0
+| `noise_standard_deviation` | If > 0, FreqAI adds noise to the training features. FreqAI generates random deviates from a gaussian distribution with a standard deviation of `noise_standard_deviation` and adds them to all data points. Value should be kept relative to the normalized space between -1 and 1). In other words, since data is always normalized between -1 and 1 in FreqAI, the user can expect a `noise_standard_deviation: 0.05` to see 32% of data randomly increased/decreased by more than 2.5%. Good for preventing overfitting. <br> **Datatype:** int. Default: 0
```
xmatthias: i guess it's clear for experts - but how do you get to 32%, starting with a range of -1 to 1 (so a range of 2) - and 5% of deviation. I can get to 2.5% easily (that's just 2 * 0.05) - but the 32% seem randomly picked at first glance (i'm certain they're calculated, but ...)

freqtrade/freqai/data_kitchen.py
```diff
@@ -724,6 +724,113 @@ def rotate_point(origin, point, angle):
...
 
+    def compute_inlier_metric(self, set_='train') -> None:
+        """
+
+        Compute inlier metric from backwards distance distributions.
+        This metric defines how well features from a timepoint fit
+        into previous timepoints.
+        """
+
+        import scipy.stats as ss
```
xmatthias: For (world-) historic reasons (some people are sensitive to this) - i think we should pick a different alias (anything other than `ss` will work fine).

the most obvious one might be `stats` - which would allow you to rewrite the import to `from scipy import stats` (flake will tell you about that, too) :laughing:

freqtrade/freqai/data_kitchen.py
```diff
@@ -724,6 +724,113 @@ def rotate_point(origin, point, angle):
...
+            )
+
+        inlier_metric = pd.DataFrame(
+            data=inliers.sum(axis=1) / no_prev_pts,
+            columns=['inlier_metric'],
+            index=compute_df.index
+        )
+
+        inlier_metric = 2 * (inlier_metric - inlier_metric.min()) / \
+            (inlier_metric.max() - inlier_metric.min()) - 1
```
xmatthias: I'd prefer wrapping this in parentheses instead of `\` linewrapping (you can wrap at any time if the whole statement is in `()`).

I've experienced this as error-prone in some instances - where (no-good) editors may add spaces after this - leading to a very subtle bug you can spend ages debugging - if it happens to result in a valid python line.

freqtrade/freqai/data_kitchen.py
```diff
@@ -843,14 +950,14 @@ def create_fulltimerange(self, backtest_tr: str, backtest_period_days: int) -> s
...
+            #     datetime.now(tz=timezone.utc).timestamp()
             # )
 
         backtest_timerange.startts = (
             backtest_timerange.startts - backtest_period_days * SECONDS_IN_DAY
         )
-        start = datetime.datetime.utcfromtimestamp(backtest_timerange.startts)
-        stop = datetime.datetime.utcfromtimestamp(backtest_timerange.stopts)
+        start = datetime.utcfromtimestamp(backtest_timerange.startts)
+        stop = datetime.utcfromtimestamp(backtest_timerange.stopts)
```
xmatthias: This is ... wrong (sorry for not noticing this before - big reviews ... :unamused: ) - and as such dangerous (even though it'll probably not have a big impact in this location specifically).

using `utcfromtimestamp()` results in a "broken" datetime object - unless your system is in the UTC timezone.
while the time will be in utc - the object will be tz native (in your system timezone).
as a consequence - running `datetime.utcfromtimestamp(backtest_timerange.startts).timestamp()` will return a timestamp shifted by your timezone offset.

The correct way for this is the below:
`datetime.fromtimestamp(timerange.startts, tz=timezone.utc)`

obviously applies to all usages of `utcfromtimestamp()` ...

tests/freqai/conftest.py
```diff
@@ -82,6 +82,38 @@ def get_patched_freqaimodel(mocker, freqaiconf):
...
+    freqai = strategy.freqai
+    freqai.live = True
+    freqai.dk = FreqaiDataKitchen(freqai_conf)
+    freqai.dk.pair = "ADA/BTC"
+    timerange = TimeRange.parse_timerange("20180110-20180130")
+    freqai.dd.load_all_pair_histories(timerange, freqai.dk)
+
+    freqai.dd.pair_dict = MagicMock()
+
+    data_load_timerange = TimeRange.parse_timerange("20180110-20180130")
```
xmatthias: is `timerange` and `data_load_timerange` intentionally identical?

docs/freqai.md
robcaulk: Ya, it is a hard one to describe without a figure. Since we are setting the noise with a gaussian, its shape is controlled by the standard deviation, `noise_standard_deviation` in this case. With gaussian PDFs, 32% of the generated points come from within the 1st standard deviation. User sets the standard deviation with `noise_standard_deviation`. 

I'll add some more...words...haha

freqtrade/freqai/data_kitchen.py
robcaulk: Thanks for noticing, I will fix this and any other occurrences.

freqtrade/freqai/data_kitchen.py
robcaulk: Sure, no problem.

freqtrade/freqai/data_kitchen.py
robcaulk: I agree!

tests/freqai/conftest.py
robcaulk: Ah it makes no difference. I removed the second one.

xmatthias (approved):
LGTM :+1:

PR #7323, opened 2022-08-31
Update and correct descriptions and figure

th0rntwig:
The description of the parameters used for DBSCAN needed updating after the computation update in a previous PR. 
The description of the outlier protection was changed to clarify its effect.
The figure showing the effect of weigh factors during training was updated for cosmetic purposes.


robcaulk:
Looks good, but please roll back the last commit so we can keep the logo full screen width. You will need to do:
```
git reset --hard HEAD~1
git push -f yourrepo HEAD:improv-doc
```

th0rntwig:
Rolled back the commit, the scaled logo should now be gone.

robcaulk (approved):
Thanks!

PR #7324, opened 2022-08-31
Unnecessary lines in Base4, and changes for box space, to fit better for our needs

richardjozsa:


robcaulk:
Thanks for this, in the future, can you start from a clean `feat/freqai-rl-dev` branch? Right now your branch has all sorts of extra commits that have nothing to do with the present PR. it starts to add up and it means I cant make changes to your branch without accepting all your previous incorrect history.

PR #7325, opened 2022-08-31
Improve Freqai backtesting performance

wagnercosta:
## Summary

After training the models, the predictions will now be saved for the subsequent backtesting execution to be faster

## What's new?

Currently, after the model is saved in the first run of the backtesing, the next runs perform the predictions again. This change will save predictions on the first run, making it unnecessary inference on subsequent runs, which will allow backtesting to be faster with saved models and predictions.

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -1,7 +1,6 @@
 import copy
 import datetime
 import logging
-import os
 import shutil
 from pathlib import Path
 from typing import Any, Dict, List, Tuple
@@ -1108,15 +1107,25 @@ class FreqaiDataKitchen:
         :param file_name: h5 file name
         :param root_folder: folder to save h5 file
         """
-        os.makedirs(root_folder, exist_ok=True)
-        append_df.to_hdf(file_name, key='append_df', mode='w')
+        backtesting_root = Path(
+            self.full_path
+            / root_folder
+        )
+        if not backtesting_root.is_dir():
+            backtesting_root.mkdir(parents=True, exist_ok=True)
 
-    def get_backtesting_prediction(self, prediction_file_name: str) -> DataFrame:
+        full_file_path = Path(self.full_path / root_folder / file_name)
+        append_df.to_hdf(full_file_path, key='append_df', mode='w')
+
+    def get_backtesting_prediction(
+        self, root_prediction: str, prediction_file_name: str
+    ) -> DataFrame:
         """
         Retrive from disk the prediction dataframe
         :param prediction_file_name: prediction file full path
         :return:
         :Dataframe: Backtesting prediction from current backtesting period
         """
-        append_df = pd.read_hdf(prediction_file_name)
+        prediction_path = Path(self.full_path / root_prediction / prediction_file_name)
+        append_df = pd.read_hdf(prediction_path)
         return append_df
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -69,6 +69,8 @@ class FreqaiDataKitchen:
         self.label_list: List = []
         self.training_features_list: List = []
         self.model_filename: str = ""
+        self.backtesting_results_path = Path()
+        self.backtesting_prediction_folder: str = "backtesting_predictions"
         self.live = live
         self.pair = pair
 
@@ -808,8 +810,6 @@ class FreqaiDataKitchen:
         else:
             self.full_df = pd.concat([self.full_df, append_df], axis=0)
 
-        return append_df
-
     def fill_predictions(self, dataframe):
         """
         Back fill values to before the backtesting range so that the dataframe matches size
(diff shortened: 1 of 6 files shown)


freqtrade/freqai/data_kitchen.py
```diff
@@ -1089,3 +1098,25 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
...
+    def save_backtesting_prediction(
+        self, file_name: str, root_folder: str, append_df: DataFrame
+    ) -> None:
+
+        """
+        Save prediction dataframe from backtesting to h5 file format
+        :param file_name: h5 file name
+        :param root_folder: folder to save h5 file
+        """
+        os.makedirs(root_folder, exist_ok=True)
```
xmatthias: Please don't use `os.` - but [pathlib](https://docs.python.org/3/library/pathlib.html) for OS operations (Which has a `path.mkdir()` option).

freqtrade/freqai/freqai_interface.py
```diff
@@ -643,6 +665,56 @@ def train_timer(self, do='start'):
...
+        :param dk: FreqaiDataKitchen
+        :trained_timestamp: current backtesting timestamp period
+        :return:
+        :str: prediction file name
+        :str: prediction root path
+        """
+        coin, _ = pair.split("/")
+        prediction_base_filename = f"{coin.lower()}_{trained_timestamp}"
+        root_prediction = f'{dk.full_path}/backtesting_predictions'
+        prediction_file_name = f"{root_prediction}/{prediction_base_filename}_predictions.h5"
```
xmatthias: Please use Pathlib when working with filenames (pathlib's `Path`, to be precise).

This will fail on windows - as `/` is not used as directory separator. pathlib provides options to handle this gracefully - there'll be enough examples of this throughout this repository.

xmatthias (dismissed):

xmatthias:
You should also not be making PR's from your develop branch (for the future though!).
There's some drawbacks (you can't really merge upstream to your branch until this is merged) - and it also has some drawbacks for us in some instances.

freqtrade/freqai/data_kitchen.py
wagnercosta: Thanks for the feedback. Fixed!

freqtrade/freqai/freqai_interface.py
wagnercosta: Thanks for the feedback. Fixed!

freqtrade/freqai/freqai_interface.py
```diff
@@ -637,6 +659,56 @@ def train_timer(self, do='start'):
...
+        returns the path and name of the predictions file
+        :param pair: pair e.g. BTC/USD
+        :param dk: FreqaiDataKitchen
+        :trained_timestamp: current backtesting timestamp period
+        :return:
+        :str: prediction file name
+        :str: prediction root path
+        """
+        coin, _ = pair.split("/")
+        prediction_base_filename = f"{coin.lower()}_{trained_timestamp}"
```
robcaulk: We have `dk.model_filename` which is this same string. Better to use that instead of creating a duplicate.

freqtrade/freqai/freqai_interface.py
```diff
@@ -637,6 +659,56 @@ def train_timer(self, do='start'):
...
+        root_prediction = 'backtesting_predictions'
```
robcaulk: Please assign this as `self.backtesting_results_path` to self in the `__init__()` of `class FreqaiDatakitchen` so that you can refer to it anywhere using `dk.backtesting_results_path`

freqtrade/freqai/freqai_interface.py
```diff
@@ -637,6 +659,56 @@ def train_timer(self, do='start'):
...
+                logger.info("Found backtesting prediction file at %s", prediction_file_name)
+            elif not scanning:
+                logger.info(
+                    "Could not find backtesting prediction file at %s", prediction_file_name
+                )
+            return file_exists
+        else:
+            return False
+
+    def get_backtesting_prediction_file_name(
```
robcaulk: We can remove this function entirely. See the comments below.

freqtrade/freqai/freqai_interface.py
```diff
@@ -637,6 +659,56 @@ def train_timer(self, do='start'):
...
+        :param pair: pair e.g. BTC/USD
+        :param path: path to prediction
+        :return:
+        :boolean: whether the prediction file exists or not.
+        """
+        if not self.live:
+            prediction_file_name, root_prediction = self.get_backtesting_prediction_file_name(
+                pair, dk, trained_timestamp
+            )
+            path_to_predictionfile = Path(dk.full_path / root_prediction / prediction_file_name)
```
robcaulk: e.g. `path_to_predictionfile = Path(dk.full_path / dk.backtesting_results_path / dk.model_filename)`

freqtrade/freqai/data_kitchen.py
```diff
@@ -1060,3 +1068,35 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
         if self.unique_classes:
             for label in self.unique_classes:
                 self.unique_class_list += list(self.unique_classes[label])
+
+    def save_backtesting_prediction(
```
robcaulk, line 1072: Is there a reason you chose to put this outside `save_data()`? It is preferable to do this inside the data_kitchen `save_data()` function, as it is better to avoid adding another function if we don't have to.  

You can use the `if not self.live:` flag inside `save_data()` to ensure this only occurs during backtesting.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1060,3 +1068,35 @@ def get_unique_classes_from_labels(self, dataframe: DataFrame) -> None:
...
+    def get_backtesting_prediction(
+        self, root_prediction: str, prediction_file_name: str
+    ) -> DataFrame:
+        """
+        Retrive from disk the prediction dataframe
+        :param prediction_file_name: prediction file full path
+        :return:
+        :Dataframe: Backtesting prediction from current backtesting period
+        """
+        prediction_path = Path(self.full_path / root_prediction / prediction_file_name)
```
robcaulk: Similar here, we can just use the `dk.backtesting_results_path` and `dk.model_filename` directly instead of passing arguments around.

robcaulk (changes requested):
Thanks a lot for this - I think you have the method down. 

Here I am just avoiding code duplication and reducing points of possible failure.

freqtrade/freqai/data_kitchen.py
wagnercosta: The main reason is because this method is called in freqai_interface after ```pred_df, do_preds = self.predict(dataframe_backtest, dk)``` that only happens some lines below, after the model is saved or reloaded (in case of a pre-existent model). I can change the logic, but I don't know if it's will be better. What do you think?

freqtrade/freqai/freqai_interface.py
wagnercosta: Thanks for the feedback. Fixed!

freqtrade/freqai/freqai_interface.py
wagnercosta: Thanks for the feedback. Fixed!

freqtrade/freqai/freqai_interface.py
wagnercosta: Thanks for the feedback. Fixed!

freqtrade/freqai/freqai_interface.py
wagnercosta: Thanks for the feedback. Fixed!

freqtrade/freqai/data_kitchen.py
wagnercosta: Thanks for the feedback. Fixed!

freqtrade/freqai/data_kitchen.py
wagnercosta: Another idea came to my mind: Do we need to save the model in backtesting? Or could it be something optional defined in the config file? The reason to save the model would be to reuse it, however, we are already saving the predictions. In most cases, I think saving the model will be unnecessary in backtesting. This will improve first run performance too, right? What do you think?

robcaulk: I agree we could add a config param to not save the model if you feel inclined to add it to the PR :)

wagnercosta: Sure. I will add it.

robcaulk (approved):
Looks good to me, nice work - the backtesting community will benefit tremendously from this PR.

PR #7329, opened 2022-09-01
Make the recommendation for Binance/Kucoin blacklisting more accurate.

epigramx:
Now that a recent bug regarding selling `BNB` is fixed, it should be safe to trade it, but with a warning that the user may have to manually maintain extra `BNB`. 
Also the old text implied those features are always enabled so this text makes it clear those fee-related features can be also disabled.
I'm not sure if it's still true that an "eaten by fees" position becomes unsellable but I left that as it is (also I assumed Kucoin is similar).

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -61,8 +61,8 @@ Binance supports [time_in_force](configuration.md#understand-order_time_in_force
 
 ### Binance Blacklist
 
-For Binance, please add `"BNB/<STAKE>"` to your blacklist to avoid issues.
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
> I'm not sure if it's still true that an "eaten by fees" position becomes unsellable but I left that as it is (also I assumed Kucoin is similar).

That's for sure still true - while there is "some" margin for error here - we always have to assume that the BNB trade stays open for days - while you may have 10's or 100ds of trades in parallel.

LGTM :+1:

PR #7331, opened 2022-09-01
Normalise PCA space

th0rntwig:
Normalisation of principal component transformed features.

PCA can be enabled by the user to reduce the feature space by finding the components that describe 99.9% of the variance in the dataset. This transformation is done on normalised [-1,1] data but the transformed data is no longer normalised to [-1,1]. This PR enables normalisation of the transformed data, and ensures that transformed test and prediction data is normalised to the corresponding transformed training data.

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -287,19 +287,26 @@ class FreqaiDataKitchen:
         :returns:
         :data_dictionary: updated dictionary with standardized values.
         """
+
+        df_train_features = data_dictionary["train_features"]
         # standardize the data by training stats
-        train_max = data_dictionary["train_features"].max()
-        train_min = data_dictionary["train_features"].min()
-        data_dictionary["train_features"] = (
-            2 * (data_dictionary["train_features"] - train_min) / (train_max - train_min) - 1
+        train_max = df_train_features.max()
+        train_min = df_train_features.min()
+        df_train_features = (
+            2 * (df_train_features - train_min) / (train_max - train_min) - 1
         )
         data_dictionary["test_features"] = (
             2 * (data_dictionary["test_features"] - train_min) / (train_max - train_min) - 1
         )
 
         for item in train_max.keys():
-            self.data[item + "_max"] = train_max[item]
-            self.data[item + "_min"] = train_min[item]
+            if not [col for col in df_train_features.columns if col.startwith('PC')]:
+                self.data[item + "_max"] = train_max[item]
+                self.data[item + "_min"] = train_min[item]
+            else:
+                # if PCA is enabled and has transformed the training features
+                self.data[item + "_pca_max"] = train_max[item]
+                self.data[item + "_pca_min"] = train_min[item]
 
         for item in data_dictionary["train_labels"].keys():
             if data_dictionary["train_labels"][item].dtype == object:
@@ -320,8 +327,14 @@ class FreqaiDataKitchen:
                     - 1
                 )
 
-            self.data[f"{item}_max"] = train_labels_max  # .to_dict()
-            self.data[f"{item}_min"] = train_labels_min  # .to_dict()
+            if not [col for col in df_train_features.columns if col.startwith('PC')]:
+                self.data[f"{item}_max"] = train_labels_max  # .to_dict()
+                self.data[f"{item}_min"] = train_labels_min  # .to_dict()
+            else:
+                # if PCA is enabled and has transformed the training features
+                self.data[f"{item}_pca_max"] = train_labels_max  # .to_dict()
+                self.data[f"{item}_pca_min"] = train_labels_min  # .to_dict()
+
         return data_dictionary
 
     def normalize_data_from_metadata(self, df: DataFrame) -> DataFrame:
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/data_kitchen.py
```diff
@@ -287,19 +287,26 @@ def normalize_data(self, data_dictionary: Dict) -> Dict[Any, Any]:
...
 
         for item in train_max.keys():
-            self.data[item + "_max"] = train_max[item]
-            self.data[item + "_min"] = train_min[item]
+            if not [col for col in df_train_features.columns if col.startswith('PC')]:
+                self.data[item + "_max"] = train_max[item]
+                self.data[item + "_min"] = train_min[item]
+            else:
+                # if PCA is enabled and has transformed the training features
+                self.data[item + "_pca_max"] = train_max[item]
```
robcaulk: `item` here is going to be unique already. In the case of principal components, `item` will be `PC1` `PC2` etc. Can't we just leave this code without the check then since the keys will be unique already in the coin meta data (`self.data`)?

Or is there another reason why you attach `_pca_max`, and I am missing it?

freqtrade/freqai/data_kitchen.py
```diff
@@ -320,8 +327,14 @@ def normalize_data(self, data_dictionary: Dict) -> Dict[Any, Any]:
                     - 1
                 )
 
-            self.data[f"{item}_max"] = train_labels_max  # .to_dict()
-            self.data[f"{item}_min"] = train_labels_min  # .to_dict()
+            if not [col for col in df_train_features.columns if col.startswith('PC')]:
```
robcaulk: Same as comment above, I dont think we need this check since `item` is unique already wrt principal components or normal features.

freqtrade/freqai/data_kitchen.py
```diff
@@ -331,11 +344,17 @@ def normalize_data_from_metadata(self, df: DataFrame) -> DataFrame:
         :param df: Dataframe to be standardized
         """
 
+        if not [col for col in df.columns if col.startswith('PC')]:
```
robcaulk: If we are lucky, even this function will remain entirely unchanged since `item` is unique. See above comments and please confirm my suspicion.

freqtrade/freqai/data_kitchen.py
```diff
@@ -450,38 +469,42 @@ def principal_component_analysis(self) -> None:
 
         from sklearn.decomposition import PCA  # avoid importing if we dont need it
 
-        n_components = self.data_dictionary["train_features"].shape[1]
-        pca = PCA(n_components=n_components)
+        pca = PCA(0.999)
```
robcaulk, line 468: Nice this is more computationally efficient compared to stable.

freqtrade/freqai/data_kitchen.py
```diff
@@ -450,38 +469,42 @@ def principal_component_analysis(self) -> None:
...
-        n_keep_components = np.argmin(pca.explained_variance_ratio_.cumsum() < 0.999)
-        pca2 = PCA(n_components=n_keep_components)
+        n_keep_components = pca.n_components_
         self.data["n_kept_components"] = n_keep_components
-        pca2 = pca2.fit(self.data_dictionary["train_features"])
+        n_components = self.data_dictionary["train_features"].shape[1]
         logger.info("reduced feature dimension by %s", n_components - n_keep_components)
-        logger.info("explained variance %f", np.sum(pca2.explained_variance_ratio_))
-        train_components = pca2.transform(self.data_dictionary["train_features"])
+        logger.info("explained variance %f", np.sum(pca.explained_variance_ratio_))
```
robcaulk, line 474: I know the old code used these strings, but in general we prefer the use of f-strings. So any chance we touch some old code that doesnt use f-strings, ideally we would convert it. I can do that on this branch. Just a comment for future PRs.

freqtrade/freqai/data_kitchen.py
```diff
@@ -450,38 +469,42 @@ def principal_component_analysis(self) -> None:
...
         if self.freqai_config.get('data_split_parameters', {}).get('test_size', 0.1) != 0:
-            test_components = pca2.transform(self.data_dictionary["test_features"])
+            test_components = pca.transform(self.data_dictionary["test_features"])
             self.data_dictionary["test_features"] = pd.DataFrame(
                 data=test_components,
                 columns=["PC" + str(i) for i in range(0, n_keep_components)],
                 index=self.data_dictionary["test_features"].index,
             )
+            # normalise transformed test feature to transformed training features
+            self.data_dictionary["test_features"] = self.normalize_data_from_metadata(
```
robcaulk, line 499: This is an interesting component of this PR that I hope has a meaningful impact on the principal component feature. Thanks for this.

robcaulk (changes requested):
Generally speaking - this is going to improve principal component analysis feature. 

But I think we get some of this already for "free" without adding all these extra strings to the metadata. See my comment and let me know if I missed something.

freqtrade/freqai/data_kitchen.py
th0rntwig: You're right, the keys are already unique, my mistake!

freqtrade/freqai/data_kitchen.py
th0rntwig: And more readable imo :)

freqtrade/freqai/data_kitchen.py
th0rntwig: Good to know, will do that from now on!

freqtrade/freqai/data_kitchen.py
th0rntwig: I think it will improve the performance of the models when the user enables PCA. Looking forward to seeing if that's correct.

robcaulk (approved):

PR #7337, opened 2022-09-02
Ensure that talib never gets called in parallel

robcaulk:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/templates/FreqaiHybridExampleStrategy.py
+++ b/freqtrade/templates/FreqaiHybridExampleStrategy.py
@@ -174,30 +174,28 @@ class FreqaiExampleHybridStrategy(IStrategy):
     # flake8: noqa: C901
     def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
 
-        # User creates their own custom strat here. Present example is a supertrend
-        # based strategy.
-
         dataframe = self.freqai.start(dataframe, metadata, self)
 
-        # TA indicators to combine with the Freqai targets
-        # RSI
-        dataframe['rsi'] = ta.RSI(dataframe)
+        with self.freqai.analysis_lock:
+            # TA indicators to combine with the Freqai targets
+            # RSI
+            dataframe['rsi'] = ta.RSI(dataframe)
 
-        # Bollinger Bands
-        bollinger = qtpylib.bollinger_bands(qtpylib.typical_price(dataframe), window=20, stds=2)
-        dataframe['bb_lowerband'] = bollinger['lower']
-        dataframe['bb_middleband'] = bollinger['mid']
-        dataframe['bb_upperband'] = bollinger['upper']
-        dataframe["bb_percent"] = (
-            (dataframe["close"] - dataframe["bb_lowerband"]) /
-            (dataframe["bb_upperband"] - dataframe["bb_lowerband"])
-        )
-        dataframe["bb_width"] = (
-            (dataframe["bb_upperband"] - dataframe["bb_lowerband"]) / dataframe["bb_middleband"]
-        )
+            # Bollinger Bands
+            bollinger = qtpylib.bollinger_bands(qtpylib.typical_price(dataframe), window=20, stds=2)
+            dataframe['bb_lowerband'] = bollinger['lower']
+            dataframe['bb_middleband'] = bollinger['mid']
+            dataframe['bb_upperband'] = bollinger['upper']
+            dataframe["bb_percent"] = (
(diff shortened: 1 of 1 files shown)


freqtrade/templates/FreqaiHybridExampleStrategy.py
```diff
@@ -174,30 +174,28 @@ def populate_any_indicators(
...
-            (dataframe["close"] - dataframe["bb_lowerband"]) /
-            (dataframe["bb_upperband"] - dataframe["bb_lowerband"])
-        )
-        dataframe["bb_width"] = (
-            (dataframe["bb_upperband"] - dataframe["bb_lowerband"]) / dataframe["bb_middleband"]
-        )
-
-        # TEMA - Triple Exponential Moving Average
-        dataframe['tema'] = ta.TEMA(dataframe, timeperiod=9)
+        with self.freqai.analysis_lock:
```
xmatthias, line 179: i think this lock should move to the strategy class - so we'll use it for all populate_* calls (independent of freqai).

That'll have it abstracted away (i still don't like users having to take care of locking) - and ta-lib really seems to be dangerous in this regards - which could in theory fall on our feet with other operations, too.

Also (independent of this ...) - we should aim to get a reproducible, minimal sample setup - where we can show this problem to the folks at the [ta-lib python wrapper](https://github.com/mrjbq7/ta-lib/).

there doesn't seem to be an issue for this - and the only mention about threads is in https://github.com/mrjbq7/ta-lib/issues/472#issuecomment-952073436 - which doesn't really go into further details though.

robcaulk: Agreed on all points.

robcaulk: What is the best way to do it? Do we add another decorator to `populate_indicators` in `interface.py`?

robcaulk: I created a barebones issue over at ta-lib - let's see if it gains any traction. If not, I will work on making an MWE for them. 

https://github.com/mrjbq7/ta-lib/issues/540

robcaulk: PR made to demonstrate the issue with an MWE:

https://github.com/mrjbq7/ta-lib/pull/541

xmatthias: i'd ... hold on this for a sec to see how the ta-lib issue progresses. If we see a rather "quick" release - i think we can update the dependency an remove the locks that were required for this.

robcaulk: He released a new version - we just need to test it. But the test I wrote over there is our exact usecase - so there is high probability the new release will work.

xmatthias: we'll not be able to update until updated windows wheels are available (usually here: https://www.lfd.uci.edu/~gohlke/pythonlibs/#ta-lib).

That's usually within 2-3 days ... but the site now says "archived" - and the last update seems to be from June.

so it's likely that we'll not be able to update for now- as the new version can't be reliably installed on windows (ta-lib does not provide wheels themselfs - even though there would be a PR for it available).

robcaulk: In that case, I say we merge this PR and wait. It is  a bandaid but at least we can take the bandaid off (and the others inside FreqAI) in a future PR.

xmatthias: not like this though.
it's a faulty that leaves this up to the user, which will break everyones strategy once we are able to remove this lock (as the lock object will no longer be there).

We'll therefore probably have to wrap `populate_indicators()` (which would move the lock to the strategy interface).

I'm also trying to get ta-lib to build wheels in CI - while there's 2 pr's available there, neither is in an acceptable state (one changes 300 files - the other uses travis, which ta-lib is no longer using).

robcaulk: So you suggest we wrap `populate_indicators()` in a lock? Then all of FreqAI is locked because it is itself, called inside `populate_indicators()`.

I guess this goes back to my original question:

> What is the best way to do it? Do we add another decorator to `populate_indicators` in `interface.py`?

xmatthias:
no longer necessary, since the update of ta-lib in #7406 should fix this as well - without putting freqtrade-internals (like a lock) into the strategy - restricting us from removing this lock in the future.

PR #7340, opened 2022-09-03
Support SIGINT in FreqAI

wizrds:
## Summary

Support shutting down FreqAI and child threads when received SIGINT. This will allow the current training iteration to finish, but will close after without allowing more to start.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -148,6 +148,8 @@ class FreqtradeBot(LoggingMixin):
 
         self.check_for_open_trades()
 
+        self.strategy.ft_bot_cleanup()
+
         self.rpc.cleanup()
         Trade.commit()
         self.exchange.close()
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -169,7 +169,7 @@ class IStrategy(ABC, HyperStrategyMixin):
                         'freqAI is not enabled. '
                         'Please enable it in your config to use this strategy.')
 
-                def cleanup(self, *args, **kwargs):
+                def shutdown(self, *args, **kwargs):
                     pass
 
             self.freqai = DummyClass()  # type: ignore
--- a/freqtrade/strategy/interface.py
+++ b/freqtrade/strategy/interface.py
@@ -168,6 +168,10 @@ class IStrategy(ABC, HyperStrategyMixin):
                     raise OperationalException(
                         'freqAI is not enabled. '
                         'Please enable it in your config to use this strategy.')
+
+                def cleanup(self, *args, **kwargs):
+                    pass
+
             self.freqai = DummyClass()  # type: ignore
 
     def ft_bot_start(self, **kwargs) -> None:
@@ -180,6 +184,12 @@ class IStrategy(ABC, HyperStrategyMixin):
         strategy_safe_wrapper(self.bot_start)()
 
         self.ft_load_hyper_params(self.config.get('runmode') == RunMode.HYPEROPT)
+
+    def ft_bot_cleanup(self) -> None:
+        """
+        Clean up FreqAI and child threads
+        """
+        self.freqai.shutdown()
 
     @abstractmethod
     def populate_indicators(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -76,6 +76,8 @@ class FreqaiDataDrawer:
             self.full_path / f"follower_dictionary-{self.follower_name}.json"
         )
         self.historic_predictions_path = Path(self.full_path / "historic_predictions.pkl")
+        self.historic_predictions_bkp_path = Path(
+            self.full_path / "historic_predictions.backup.pkl")
         self.pair_dictionary_path = Path(self.full_path / "pair_dictionary.json")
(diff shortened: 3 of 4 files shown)


freqtrade/strategy/interface.py
```diff
@@ -168,6 +168,10 @@ def start(self, *args, **kwargs):
                     raise OperationalException(
                         'freqAI is not enabled. '
                         'Please enable it in your config to use this strategy.')
+
+                def cleanup(self, *args, **kwargs):
```
xmatthias: the name of this method should be aligned to the method you call in `ft_bot_cleanup()` - which is shutdown - not `cleanup()`.

freqtrade/freqtradebot.py
```diff
@@ -148,6 +148,8 @@ def cleanup(self) -> None:
 
         self.check_for_open_trades()
 
+        self.strategy.ft_bot_cleanup()
```
xmatthias, line 151: should probably also be called from backtesting ...

xmatthias (commented):
While i understand what you're trying to do - is this really an issue (it's not one i noticed so far)?

the tread is not started as daemon - so once the main thread exits, child threads should clean up, too.
from [the docs](https://docs.python.org/3/library/threading.html#threading.Thread.daemon): 
> the main thread is not a daemon thread and therefore all threads created in the main thread default to daemon = False.
> The entire Python program exits when no alive non-daemon threads are left.

freqtrade/strategy/interface.py
wizrds: Good catch, originally called it cleanup but changed it to shutdown. Forgot about that part.

wizrds:
> While i understand what you're trying to do - is this really an issue (it's not one i noticed so far)?
> 
> the tread is not started as daemon - so once the main thread exits, child threads should clean up, too. from [the docs](https://docs.python.org/3/library/threading.html#threading.Thread.daemon):
> 
> > the main thread is not a daemon thread and therefore all threads created in the main thread default to daemon = False.
> > The entire Python program exits when no alive non-daemon threads are left.

So, yes when CTRL+C was pressed, the child threads would continue as it was an infinite loop. This meant that telegram's `/reload_config` would not work either. As well, @robcaulk had mentioned an elusive bug that coincided with this where people would try and shut down in the middle of pickling the model to disk, which would cause errors. I think key word there is `no alive`, as the `start_scanning` thread would still be alive even when the main thread signaled a close.

xmatthias:
> As well, @robcaulk had mentioned an elusive bug that coincided with this where people would try and shut down in the middle of pickling the model to disk,

nothing will change in this regards though - i can still force-exit the process by pressing ctrl+c multiple times - unless that's explicitly delayed until pickling is finished - i could still corrupt the file.

freqtrade/freqtradebot.py
wizrds: Where would that be called? I can't seem to find where things are cleaned up in the `Backtesting` object.

wizrds:
> > As well, @robcaulk had mentioned an elusive bug that coincided with this where people would try and shut down in the middle of pickling the model to disk,
> 
> nothing will change in this regards though - i can still force-exit the process by pressing ctrl+c multiple times - unless that's explicitly delayed until pickling is finished - i could still corrupt the file.

No it doesn't directly fix it completely, though it should help with some things. At the minimum, at least when SIGINT is sent the first time the user can expect it to shutdown soon, rather than never.

freqtrade/freqtradebot.py
wizrds: Just at the bottom of `Backtesting.start`?

robcaulk: > should probably also be called from backtesting ...

Can you be more specific? FreqAI backtesting is all single threaded. It does not require this kind of shut down sequence.

robcaulk:
> > As well, @robcaulk had mentioned an elusive bug that coincided with this where people would try and shut down in the middle of pickling the model to disk,
> 
> nothing will change in this regards though - i can still force-exit the process by pressing ctrl+c multiple times - unless that's explicitly delayed until pickling is finished - i could still corrupt the file.

It is true, technically. But enabling /reload is important in itself. Which is also a goal here. Also, this is just cleaner and more correct.

With regards to a "full fix" I will eventually add a  backup which will always keep one non-corrupted backup of the `historic_predictions.pkl` available in case of corruption.

freqtrade/freqtradebot.py
xmatthias: well my assumption was that `ft_bot_cleanup()` may be used to clean up other things too over time.

if we discover something else that'll have to be shutdown (with or without freqAI context) - will you remember to check if this is actually called everywhere? I know i probably won't ... 

Leave it for now - @robcaulk  is right, we'll not need this in backtesting for now.

xmatthias:
> But enabling /reload is important in itself. Which is also a goal here. Also, this is just cleaner and more correct.

that's not something that's obvious from the desc - but yeah, that's for sure an important point.

xmatthias (approved):
LGTM :+1: 
(i'll not be merging this for now - as i'd like @robcaulk's approval, too (feel free to merge if you're fine with it).

robcaulk:
Wow this works surprisingly well. I am going to add the `historical_predictions.backup.pkl` then merge it :-)

robcaulk (approved):
LGTM

PR #7343, opened 2022-09-04
Fix issue with Freqai backtesting (different results in same day with different TR)

wagnercosta:
## Summary

Fixed an issue when backtesting with the same models and predictions but with different timeranges the backtesting returns different results for the same day  (breakdown day).


robcaulk (approved):
Nice find!

PR #7362, opened 2022-09-05
improve clarity on data download requirements

robcaulk:
We have had confusion in the main discord regarding how much data should be pre-downloaded for backtesting. This PR aims to help the lost souls who plan to never read it ;)


docs/freqai.md
```diff
@@ -515,10 +515,11 @@ and if a full `live_retrain_hours` has elapsed since the end of the loaded model
 The FreqAI backtesting module can be executed with the following command:
 
 ```bash
-freqtrade backtesting --strategy FreqaiExampleStrategy --config config_examples/config_freqai.example.json --freqaimodel LightGBMRegressor --timerange 20210501-20210701
+freqtrade download-data --config config_examples/config_freqai.example.json --strategy FreqaiExampleStrategy --strategy-path freqtrade/templates -t 3m 15m 1h --timerange 20210315-20210701
```
xmatthias: wouldn't a link (and a comment saying we assume data is available) to the downloading docs suffice?

i'm worried about having this here as people will assume it's always required - which it clearly isn't (you only need this the first time).

apart from that - the command is wrong - download-data ignores the end-date in such a case.

robcaulk: It is a fair point. Link + note will do

xmatthias: they'll not read it anyway :laughing:

xmatthias (approved):
Great :+1:

PR #7363, opened 2022-09-05
Fix hyperopt - freqai

wagnercosta:
## Summary

Fix the hyperopt to work with Freqai

## What's new?

Fix the issue `TypeError: cannot pickle '_thread.lock' object` when hyperopting a freqai strategy


freqtrade/freqai/data_drawer.py
```diff
@@ -155,14 +165,22 @@ def save_historic_predictions_to_disk(self):
...
+                with open(self.pair_dictionary_path, 'w') as fp:
+                    rapidjson.dump(
+                                    self.pair_dict, fp, default=self.np_encoder,
+                                    number_mode=rapidjson.NM_NATIVE)
+        else:
+            # save_lock it's not working with hyperopt
             with open(self.pair_dictionary_path, 'w') as fp:
-                rapidjson.dump(self.pair_dict, fp, default=self.np_encoder,
-                               number_mode=rapidjson.NM_NATIVE)
+                rapidjson.dump(
```
xmatthias: i don't think this will really work / suffice (at least not without introducing additional problems).

this will only work if the freqAI model is guaranteed to not retrain the models - and in the end, not to not call this method.
Hyperopt runs backtesting in several parallel procesess.
This means, you now potentially have X parallel freqAI trainings running (if using `--analyze-per-epoch`, or implementing `freqai.start()` in a  non-default place - see below)
apart from the fact that these will slow each other down as they each scale out to multiple threads/processes themselfs - they'll also write the model to disk once done.
They'll however write the model to disk - using the **same** filename - over and over again.

Ideally - you'll "just" have models that change all the time - but there's nothing preventing the model for ETH/USDT from being written to disk at the same time - causing corruption and/or crashes (probably both).

---

now in theory we can just reject to start hyperopt on a freqAI strategy that also defines `--analyze-per-epoch` - but i think a "2nd layer of defense" should be added - to protect from users implementing freqai not in `populate_indicators()` - but in `populate_entry_trend()` - which is called in parallel processes all the time, independent of `--analyze-per-epoch`.

wagnercosta: After a better analysis, we don't need to call this method in hyperopt (in parallels process). Only in the first run, if the user choose to save models, but this happens before the parallels execution. All things inside `populate_any_indicators` in a Freqai strategy will no be hyperopt. The root problem is that I was using the FreqaiExampleStrategy and we have this call inside the strategy, in `confirm_trade_exit` and `custom_exit` methods:

```
        pair_dict[pair]['extras']["prediction" + entry_tag] = 0
        if not follow_mode:
            self.freqai.dd.save_drawer_to_disk()
        else:
            self.freqai.dd.save_follower_dict_to_disk()
```

@robcaulk , why do we need this calls inside the strategy? This is handled by the backend, right? Maybe something old that we need to clear.

robcaulk: The only goal of this code is to use the trade candle prediction for exit logic. We store it to disk because we want crash resilience. 

This means that it can be removed for backtesting, and the same info can be stored using a dictionary similar to `custom_info` example in the docs: https://www.freqtrade.io/en/stable/strategy-advanced/#storing-information

Or just remove it entirely and let users handle this on their own.

wagnercosta: Thanks Rob. Great. My suggestion for now is to remove this in example strategy and we can create a task to add a new example using `custom_info`. I will continue my tests without this code.

robcaulk: I never liked this code being in the example strategy. We can remove it from the `custom_exit` and the `confirm_trade_exit` . The example strategy should be simple and this just muddies the waters.

wagnercosta: I will remove all code in example strategy that refers to self.freqai.dd

xmatthias: > All things inside populate_any_indicators in a Freqai strategy will no be hyperopt.

that's ... not really true - it's an assumption based on the default/sample strategy - but nothing prevents me (or any other user) from adding another indicator that's based on a hyperoptable parameter to populate_any_indicators, instead of using what's currently defined.

While we do loop on `indicator_periods_candles` (for example) - that's a parameter that could for sure be hyperopted (which would however require retraining the model each time - maybe even requiring a new identifier ...?)
It's not used anywhere internally other than for candle startup calcs - the rest must be done within the strategy - so no-one forces me to use this - instead of a hyperoptable parameter (or set of parameters) to experiment with different lengths.

it's just a sample, really - the  real point is - nothing prevents the usage of parameters in `populate_any_indicators` ...

wagnercosta: I got it! I know that a user can in another strategy try to hyperopt an indicator in `populate_any_indicators`, but I think that we need to inform in docs that this will not work (at this moment). If we hyperopt in `populate_any_indicators`, the process will be extremely slow, because it will require a new training for each combination. I think that the main goal, at this moment, is to allow user hyperopt all other options/spaces except `populate_any_indicators`. This will be very helpful, even if it's not a complete solution yet.

> now in theory we can just reject to start hyperopt on a freqAI strategy that also defines --analyze-per-epoch - 

I agree. This will be necessary.

> to protect from users implementing freqai not in populate_indicators() - but in populate_entry_trend()

What do you mean? Prevent an user to call `self.freqai.start()` in populate_entry_trend()? Couldn't we handle this in the documentation?

xmatthias: we sure can - my point is - we have the samples that call it in one place - but nothing says "we don't support other locations".
Documentation can cover a lot of things - maybe we can add a sample in the freqAI docs about the limits of hyperopt specifically? so it's grouped together ...

robcaulk: > I got it! I know that a user can in another strategy try to hyperopt an indicator in `populate_any_indicators`, but I think that we need to inform in docs that this will not work (at this moment). If we hyperopt in `populate_any_indicators`, the process will be extremely slow, because it will require a new training for each combination. I think that the main goal, at this moment, is to allow user hyperopt all other options/spaces except `populate_any_indicators`. This will be very helpful, even if it's not a complete solution yet.
> 
> > now in theory we can just reject to start hyperopt on a freqAI strategy that also defines --analyze-per-epoch -
> 
> I agree. This will be necessary.
> 
> > to protect from users implementing freqai not in populate_indicators() - but in populate_entry_trend()
> 
> What do you mean? Prevent an user to call `self.freqai.start()` in populate_entry_trend()? Couldn't we handle this in the documentation?

Yes, we cannot allow hyperopt to retrain models. It adds complexity and it is anyways not computationally tractable.

If they want to hyperopt with FreqAI, they are limited to entry/exit criteria. That should simplify this PR and will also reduce the number of questions we can expect in discord about "why the hyperopt is taking so long?"

wagnercosta: > we sure can - my point is - we have the samples that call it in one place - but nothing says "we don't support other locations".

@robcaulk , do you see any situation that will be necessary to call self.freqai outside `populate_indicators()` or `populate_any_indicators()` ?

xmatthias: you won't need to see a situation / come up with a reason to have users try it 😆

robcaulk: No. 

Calling `self.freqai.start()` inside `populate_any_indicators()` would cause an infinite loop. 

We haven't seen users trying to call `self.freqai.start()` anywhere besides the instructed location of `populate_indicators()`. So I don't think this is much of a concern at the moment.

freqtrade/configuration/configuration.py
```diff
@@ -537,3 +538,12 @@ def _resolve_pairs_list(self, config: Dict[str, Any]) -> None:
                 config['pairs'] = load_file(pairs_file)
                 if 'pairs' in config and isinstance(config['pairs'], list):
                     config['pairs'].sort()
+
+    def _check_hyperopt_analyze_per_epoch_freqai(self) -> None:
```
wagnercosta: @xmatthias , I added the check to this file. I don't know if it's the best place.

docs/freqai.md
```diff
@@ -532,6 +534,15 @@ for each pair, for each backtesting window within the expanded `--timerange`.
 
 ---
 
+### Hyperopt
```
wagnercosta, line 539: @robcaulk I appreciate your help to improve the text!

freqtrade/freqai/data_drawer.py
wagnercosta: Well. New version with all considerations addressed!

freqtrade/configuration/configuration.py
xmatthias: I think the right place to check for this is in `config_validation.py` (and checking this against the config, obviously).

Checking arguments won't help much - as the same elements can be set via config - and would currently pass.

wagnercosta: Thanks! Fixed!

freqtrade/freqai/data_drawer.py
wagnercosta: In the end, the change was much simpler than I expected.

freqtrade/configuration/config_validation.py
```diff
@@ -323,6 +324,14 @@ def _validate_pricing_rules(conf: Dict[str, Any]) -> None:
             del conf['ask_strategy']
 
 
+def _validate_freqai_hyperopt(conf: Dict[str, Any]) -> None:
+    freqaimodel = conf.get('freqaimodel')
```
xmatthias: i think you should validate `conf.get('freqai', {}).get('enabled')` here ... checking on freqaimodel is ... odd - to say the least

wagnercosta: You are right! Fixed.

docs/freqai.md
robcaulk: The wording seems good to me!

robcaulk (commented):
Thanks again for this - this is a huge feature for FreqAI :-)

Sorry it is taking me some time to test - I will prioritize testing this tonight. But it seems you two sorted the main issues out already. Thanks @xmatthias for that effort.

freqtrade/templates/FreqaiExampleStrategy.py
```diff
@@ -47,11 +45,6 @@ class FreqaiExampleStrategy(IStrategy):
     startup_candle_count: int = 40
     can_short = False
 
-    linear_roi_offset = DecimalParameter(
```
robcaulk, line 50: I think we need to keep at least one hyperoptable parameter around to show users example usage. I can take care of this :-)

wagnercosta: Just a suggestion, maybe a good example could be hyperopt stddev multiplier. The code below is working for me:

```python
class FreqaiExampleStrategy(IStrategy):
    """
    Example strategy showing how the user connects their own
    IFreqaiModel to the strategy. Namely, the user uses:
    self.freqai.start(dataframe, metadata)

    to make predictions on their data. populate_any_indicators() automatically
    generates the variety of features indicated by the user in the
    canonical freqtrade configuration file under config['freqai'].
    """

    std_dev_multiplier_buy = CategoricalParameter([0.75, 1, 1.25, 1.5, 1.75], default=1.25, space="buy", optimize=True)
    std_dev_multiplier_sell = CategoricalParameter([0.1, 0.25, 0.4], default=0.25, space="sell", optimize=True)
```

```python

        # All indicators must be populated by populate_any_indicators() for live functionality
        # to work correctly.

        # the model will return all labels created by user in `populate_any_indicators`
        # (& appended targets), an indication of whether or not the prediction should be accepted,
        # the target mean/std values for each of the labels created by user in
        # `populate_any_indicators()` for each training period.

        dataframe = self.freqai.start(dataframe, metadata, self)

        for val in self.std_dev_multiplier_buy.range:
            dataframe[f'target_roi_{val}'] = dataframe["&-s_close_mean"] + dataframe["&-s_close_std"] * val
            dataframe[f'sell_roi_{val}'] = dataframe["&-s_close_mean"] - dataframe["&-s_close_std"] * val
            
        return dataframe

    def populate_entry_trend(self, df: DataFrame, metadata: dict) -> DataFrame:

        enter_long_conditions = [df["do_predict"] == 1, df["&-s_close"] > df[f"target_roi_{self.std_dev_multiplier_buy.value}"]]

        if enter_long_conditions:
            df.loc[
                reduce(lambda x, y: x & y, enter_long_conditions), ["enter_long", "enter_tag"]
            ] = (1, "long")

        enter_short_conditions = [df["do_predict"] == 1, df["&-s_close"] < df[f"sell_roi_{self.std_dev_multiplier_buy.value}"]]

        if enter_short_conditions:
            df.loc[
                reduce(lambda x, y: x & y, enter_short_conditions), ["enter_short", "enter_tag"]
            ] = (1, "short")

        return df

    def populate_exit_trend(self, df: DataFrame, metadata: dict) -> DataFrame:
        exit_long_conditions = [
            df["do_predict"] == 1, 
            (df["&-s_close"] < df[f"sell_roi_{self.std_dev_multiplier_buy.value}"] * 
             self.std_dev_multiplier_sell.value)]
        if exit_long_conditions:
            df.loc[reduce(lambda x, y: x & y, exit_long_conditions), "exit_long"] = 1

        exit_short_conditions = [
            df["do_predict"] == 1, 
            (df["&-s_close"] > 
             df[f"target_roi_{self.std_dev_multiplier_buy.value}"] * self.std_dev_multiplier_sell.value)]
        if exit_short_conditions:
            df.loc[reduce(lambda x, y: x & y, exit_short_conditions), "exit_short"] = 1

        return df
```

docs/freqai.md
```diff
@@ -534,6 +536,15 @@ for each pair, for each backtesting window within the expanded `--timerange`.
...
+
+The [Hyperopt](hyperopt.md) module can be executed with some restrictions:
+
+- The `--analyze-per-epoch` hyperopt parameter is not compatible with FreqAI.
+- It's not possible to hyperopt indicators in `populate_any_indicators()` function. This means that the user cannot optimize model parameters using hyperopt. Apart from this exception, it is possible to optimize all other [spaces](hyperopt.md#running-hyperopt-with-smaller-search-space).
+- The [Backtesting](#backtesting) instructions also apply apply to Hyperopt.
```
wagnercosta: @robcaulk , please, could you remove the second apply word at final version? I just realized now

freqtrade/templates/FreqaiExampleStrategy.py
robcaulk: Great idea.

robcaulk (approved):
Tested this on my machine - works quite well. Thanks both of you for the effort on this - I have a feeling many of the FT community will explore this functionality.

docs/freqai.md
```diff
@@ -534,6 +536,31 @@ for each pair, for each backtesting window within the expanded `--timerange`.
 
 ---
 
+### Hyperopt
+
+Users can hyperopt using the same command as typical [hyperopt](hyperopt.md):
+
+```bash
+freqtrade hyperopt --hyperopt-loss SharpeHyperOptLoss --strategy FreqaiExampleStrategy --strategy-path freqtrade/templates --config config_examples/config_freqai.example.json --timerange 20220428-20220507
```
wagnercosta: @robcaulk, the model parameter is missing

PR #7367, opened 2022-09-06
add continual learning to catboost and friends

robcaulk:
Add the ability for Catboost and friends to employ "continual learning." This feature allows users to tell FreqAI to continue training existing models from their most recently trained state. As opposed to the existing paradigm where a new model is trained each retrain.

Users can activate this functionality via:

`"continual_learning": true`

In their config files.


freqtrade/freqai/freqai_interface.py
```diff
@@ -674,7 +675,7 @@ def train(self, unfiltered_dataframe: DataFrame, pair: str, dk: FreqaiDataKitche
         """
 
     @abstractmethod
-    def fit(self, data_dictionary: Dict[str, Any]) -> Any:
+    def fit(self, data_dictionary: Dict[str, Any], dk: FreqaiDataKitchen) -> Any:
```
xmatthias: do we assume we'll have the need to extend this interface further?

if yes - i think we should consider recommending people add either `*args, **kwargs` - or just `**kwargs` to theyr implementation (as we do with strategy methods) - so we can extend the interface at will without breaking people's models over and over again.

Please note: 
the latter approach (`**kwargs` only) means we MUST call all additional arguments as keyword arguments (no longer as positional arguments).

robcaulk: sounds good

freqtrade/freqai/prediction_models/LightGBMRegressorMultiTarget.py
```diff
@@ -31,6 +32,9 @@ def fit(self, data_dictionary: Dict) -> Any:
         eval_set = (data_dictionary["test_features"], data_dictionary["test_labels"])
         sample_weight = data_dictionary["train_weights"]
 
+        if self.continual_learning:
+            logger.warning('Continual learning not supported for MultiTarget models')
```
xmatthias: wouldn't you kindof want to warn about this on startup?
I think with this it's going to be shown once per pair per training ...

robcaulk: I just went ahead and made our own FreqAI regressor to allow us to support continual learning and eval sets on multioutput :-)

freqtrade/freqai/prediction_models/CatboostRegressor.py
```diff
@@ -38,16 +38,13 @@ def fit(self, data_dictionary: Dict) -> Any:
...
         model = CatBoostRegressor(
             allow_writing_files=False,
             **self.model_training_parameters,
         )
 
-        model.fit(X=train_data, eval_set=test_data)
-
-        # some evidence that catboost pools have memory leaks:
-        # https://github.com/catboost/catboost/issues/1835
-        del train_data, test_data
```
xmatthias, line 50: we no longer fear from this memory leak?

freqtrade/freqai/prediction_models/LightGBMRegressorMultiTarget.py
```diff
@@ -28,12 +29,36 @@ def fit(self, data_dictionary: Dict) -> Any:
...
+                eval_sets[i] = (  # type: ignore
+                    data_dictionary["test_features"],
+                    data_dictionary["test_labels"].iloc[:, i]
+                )
+
+        init_model = self.get_init_model(dk.pair)
+        if init_model:
+            init_models = init_model.estimators_
+        else:
+            init_models = [None] * y.shape[1]
```
xmatthias, line 50: None - or NaN ? (just asking ...)

xmatthias (approved):
just 2 questions (which are not THAT important, and can be tackled afterwards) 

LGTM :+1:

freqtrade/freqai/prediction_models/CatboostRegressor.py
robcaulk: That leak was solved with my earlier mem-leak solution :-)

freqtrade/freqai/prediction_models/LightGBMRegressorMultiTarget.py
robcaulk: None !

PR #7370, opened 2022-09-07
Add optional `suffix` parameter to `merge_informative_pair`

wizrds:
## Summary

Add an optional `suffix` parameter to `merge_informative_pair` to specify the informative columns suffix instead of just the timeframe.

### DIFF
--- a/freqtrade/strategy/strategy_helper.py
+++ b/freqtrade/strategy/strategy_helper.py
@@ -31,6 +31,8 @@ def merge_informative_pair(dataframe: pd.DataFrame, informative: pd.DataFrame,
     :param ffill: Forwardfill missing values - optional but usually required
     :param append_timeframe: Rename columns by appending timeframe.
     :param date_column: A custom date column name.
+    :param suffix: A string suffix to add at the end of the informative columns. If specified,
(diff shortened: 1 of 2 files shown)


freqtrade/strategy/strategy_helper.py
```diff
@@ -50,10 +53,17 @@ def merge_informative_pair(dataframe: pd.DataFrame, informative: pd.DataFrame,
 
     # Rename columns to be unique
     date_merge = 'date_merge'
-    if append_timeframe:
+    if append_timeframe and not suffix:
```
xmatthias: we'll clearly need this documented in the docstring - as well as on the doc page (ideally a test), checking that naming comes out as expected.

The code itself looks ok :+1:

xmatthias (changes requested):

freqtrade/strategy/strategy_helper.py
wizrds: Sounds good, can do :+1:

xmatthias (approved):
Great work! 💯 
I just reversed the sequence of the conditions - so we don't have to "and not the other" check on each condition.

LGTM  👍

PR #7374, opened 2022-09-07
Filter constant features

th0rntwig:
If a feature consists of a constant value for all time points normalisation will turn it into nans and the entire dataframe will be dropped. This PR adds a check for constant features and removes them prior to normalisation, logging a warning to the user about which features were removed.


robcaulk:
Thanks for this, I will try to fix the broken test!

th0rntwig:
Thanks!

robcaulk (approved):
Ready to go!

th0rntwig:
Great!

PR #7375, opened 2022-09-07
ensure inlier metric can be combined with other cleaning methods

robcaulk:
Inlier metric can now be stacked on top of SVM and PCA.

### DIFF
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -467,7 +467,7 @@ class IFreqaiModel(ABC):
         if ft_params.get(
             "principal_component_analysis", False
         ):
-            dk.pca_transform(dataframe)
+            dk.pca_transform(self.dk.data_dictionary['prediction_features'])
 
         if ft_params.get("use_SVM_to_remove_outliers", False):
             dk.use_SVM_to_remove_outliers(predict=True)
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -427,6 +427,11 @@ class IFreqaiModel(ABC):
 
         ft_params = self.freqai_info["feature_parameters"]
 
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/freqai_interface.py
```diff
@@ -446,11 +451,6 @@ def data_cleaning_train(self, dk: FreqaiDataKitchen) -> None:
             dk.use_DBSCAN_to_remove_outliers(predict=False, eps=eps)
             self.dd.old_DBSCAN_eps[dk.pair] = dk.data['DBSCAN_eps']
 
-        if ft_params.get('inlier_metric_window', 0):
```
xmatthias, line 449: i'm ... not sure i understand how moving this from "here" to the top of this function will change if you can combine it or not ?

robcaulk: `compute_inlier_metric` adds a feature to the feature set. 

So now we train SVM and the PCA transform on the correct number of features.

freqtrade/freqai/freqai_interface.py
```diff
@@ -467,7 +467,7 @@ def data_cleaning_predict(self, dk: FreqaiDataKitchen, dataframe: DataFrame) ->
         if ft_params.get(
             "principal_component_analysis", False
         ):
-            dk.pca_transform(dataframe)
+            dk.pca_transform(self.dk.data_dictionary['prediction_features'])
```
xmatthias, line 470: Doesn't seem like the dataframe argument is used in this function ...

as it's used in base models - it might be people pass in something and expect that to be used (instead of prediciton_features) (basically a dataframe that's not equal to the one in data_dictionary) - and then get surprising results.
maybe we should remove that argument - if we assume "dataframe magic" to happen in the background... ?

xmatthias (approved):
LGTM :+1:

PR #7376, opened 2022-09-07
Fix freqai backtesting time range issue

wagnercosta:
## Summary

Fix an issue with FreqAI backtesting returning incorrect data frame

## What's new?

Fix an issue that FreqAI is returning the full dataframe (including training) to freqtrade backtesting module and causing incorrect summary stats that uses the total backtesting days.


robcaulk (approved):
LGTM

PR #7391, opened 2022-09-10
Fix FreaqAI backtesting - startup_candle_count bug

wagnercosta:
## Summary

Freqai also needs to send the candles related to startup_candle_count for the FT backtesting module to work properly.


robcaulk (approved):
Voila

PR #7392, opened 2022-09-10
Improve freqai tests by utilizing parametrization

xmatthias:
## Summary
Parametrizing the tests reduces duplication greatly.
all tests (within one test group) were identical except for the model parameter - which does not make sense (and reduced test coverage - as catboost was forgotten ...).


robcaulk (approved):
Thanks a lot for that - agreed this is a much better use of pytest. 

Sorry for adding off-topic commits to this. But while I was reviewing it, I saw some small things that would take no time to add/cleanup.

PR #7394, opened 2022-09-11
add search share button to website

robcaulk:
This adds a "share" symbol in the search bar of the website so users can share exact search queries with one another. 

image


xmatthias (approved):
LGTM :+1:

PR #7395, opened 2022-09-11
Reorganize and improve FreqAI docs

robcaulk:
The FreqAI doc is currently a single page that is too difficult for users to consume/search. The present PR aims to better organize the documentation associated with FreqAI. 

The PR includes improvements to the doc, such as detailed descriptions of the file structure and each of the files generated by FreqAI.


docs/freqai-developers.md
```diff
@@ -0,0 +1,28 @@
...
+
+* CatboostRegressor
+* CatboostRegressorMultiTarget
+* CatboostClassifier
+* LightGBMRegressor
+* LightGBMRegressorMultiTarget
+* LightGBMClassifier
+* XGBoostRegressor
+* XGBoostRegressorMultiTarget
+* XGBoostClassifier
```
xmatthias: Ain't having a list here dangerous - at least to some extend?

assuming we add another model - it's very likely that this will be forgotten.

The list should be there for sure - but directed at users "these are your options".
developer docs should focus on what's important for the developer - this list is not really - at least not in detail.

docs/freqai-configuration.md
```diff
@@ -0,0 +1,283 @@
...
+        "fit_live_prediction_candles": 300,
+    }
+```
+
+If the user sets this value, FreqAI will initially use the predictions from the training data
+and subsequently begin introducing real prediction data as it is generated. FreqAI will save
+this historical data to be reloaded if the user stops and restarts a model with the same `identifier`.
+
+
+## Parameter table
```
xmatthias: I think this document jumps around a bit

we start with configuration - but don't explain options.
then we go to the strategy - but return to the parameters table down here again.

maybe either moving the parameters table up (or the top section down here) - or splitting this document completely into "configuration" and "strategy" might make more sense? (just a thought though)

docs/freqai-configuration.md
```diff
@@ -0,0 +1,283 @@
+# Configuration
+
+## Setting up the configuration file
+
+The user interface is isolated to the typical Freqtrade config file. Although there are plenty of additional parameters that a user can choose from, as highlighted in the [parameter table](#parameter-table), a FreqAI config must at minimum include the following parameters (the example inputs are not requires):
```
xmatthias: > The user interface is isolated to the typical Freqtrade config file

that "isolated" seems missplaced / misworded - I've read the whole paragraph 3 times now - and i still not sure in what way it's isolated. it's a separate section - sure, but ...

(it's also a lie, the "user interface" is not just the config file - but also the strategy - which takes part in freqAI's success/failure just as much).

jackromo888:
``

jackromo888 (approved):
Thanks

docs/freqai-configuration.md
robcaulk: I agree 100%

docs/freqai-developers.md
robcaulk: @th0rntwig fixed this

docs/freqai-configuration.md
robcaulk: We went a different direction . I think users need to be introduced to the config before being hit with a lengthy parameter table. The config is quite readable by design (most parameters have verbs and units). 

Now the param table is its own page - which is easily seen from variohs "heights" in FT doc.

docs/freqai-configuration.md
robcaulk: @th0rntwig fixed this

docs/freqai-data-handling.md
```diff
@@ -0,0 +1,115 @@
+# Data handling
```
xmatthias: I find this "running FreqAI" and "data handling" segments quite confusing / disorganized.

I think the best way would be to move the "filestructure" part to the developer page (regular users will not care about this) - and the rest to running freqAI.

That's based on the assumption of 'i'm new here to play around freqAI - and i'm looking for "how do i run it"' - so "running freqAI" is the place to go. Unfortunately, it only includes half of the information (it's missing model purging, for example)

docs/freqai-data-handling.md
```diff
@@ -0,0 +1,115 @@
...
+
+```json
+    "freqai": {
+        "purge_old_models": true,
+    }
+```
+
+This will automatically purge all models older than the two most recently trained ones to save disk space.
+
+### Returning additional info from training
```
xmatthias: this segment should probably go to "feature engineering".
The way this page starts tells a new user "technical details, not interesting for me" - while this part (and the part above) would be interesting.

docs/freqai-running.md
```diff
@@ -0,0 +1,139 @@
...
+
+!!! Note "Model reuse"
+    Once the training is completed, the user can execute the backtesting again with the same config file and
+    FreqAI will find the trained models and load them instead of spending time training. This is useful
+    if the user wants to tweak (or even hyperopt) buy and sell criteria inside the strategy. If the user
+    *wants* to retrain a new model with the same config file, then they should simply change the `identifier`.
+    This way, the user can return to using any model they wish by simply specifying the `identifier`.
+
+---
```
xmatthias, line 70: Above here should be a good place for the "Downloading data"  part - that'll avoid having the user skip around.

docs/freqai-parameter-table.md
```diff
@@ -0,0 +1,51 @@
...
+| `save_backtest_models` | Save models to disk when running backtesting. Backtesting operates most efficiently by saving the prediction data and reusing them directly for subsequent runs (when the user wishes to tune entry/exit parameters). Saving backtesting models to disk also allows to use the same model files for starting a dry/live instance with the same model `identifier`. <br> **Datatype:** Boolean. <br> Default: `False` (no models are saved).
+| `fit_live_predictions_candles` | Number of historical candles to use for computing target (label) statistics from prediction data, instead of from the training dataset (more information can be found [here](freqai-configuration.md#creating-a-dynamic-target-threshold)). <br> **Datatype:** Positive integer.
+| `follow_mode` | Use a `follower` that will look for models associated with a specific `identifier` and load those for inferencing. A `follower` will **not** train new models. <br> **Datatype:** Boolean. <br> Default: `False`.
+| `continual_learning` | Use the final state of the most recently trained model as starting point for the new model, allowing for incremental learning. <br> **Datatype:** Boolean. <br> Default: `False`.
```
xmatthias: might benefot from a link to continual learning

docs/freqai-feature-engineering.md
```diff
@@ -0,0 +1,165 @@
...
+            "include_shifted_candles": 2,
+            "indicator_periods_candles": [10, 20]
+        },
+        ...
+    }
+```
+
+The `include_timeframes` in the config above are the timeframes (`tf`) of each call to `populate_any_indicators()` in the strategy. In the presented case, the user is asking for the `5m`, `15m`, and `4h` timeframes of the `rsi`, `mfi`, `roc`, and `bb_width` to be included in the feature set.
+
+The user can ask for each of the defined features to be included also for informative pairs using the `include_corr_pairlist`. This means that the feature set will include all the features from `populate_any_indicators` on all the `include_timeframes` for each of the correlated pairs defined in the config (`ETH/USD`, `LINK/USD`, and `BNB/USD` in the presented example).
```
xmatthias: Just a general comment for the future, not necessarily for this PR, but maybe something we can adress going forward:

Throughout the docs, we try to adress the user as the reader (i.e. "you/your"). 

i find the "the user" terminology odd / missplaced in docs - as it assumes that the reader is not necessarily the user. This may sound correct for papers (where you explain to some reader how a user would interact with this) - but is wrong for documentation.

docs/freqai-running.md
```diff
@@ -0,0 +1,139 @@
...
+
+```json
+    "freqai": {
+        "expiration_hours": 0.5,
+    }
+```
+
+In the presented example config, the user will only allow predictions on models that are less than 1/2 hours old.
+
+## Data stratification for training and testing the model
```
xmatthias, line 108: this feels strange here - and i feel it might be better placed in the feature engineering section (which is also necessary for running freqAI, effectively).

docs/freqai-outlier-detection.md
```diff
@@ -0,0 +1,79 @@
+# Outlier detection
```
xmatthias: I'm not sure i understand why this deserves it's own page. 
It'd be better placed under "feature-engineering" if you ask me - right now that feature engineering part is kindof split in 2 - for no good reason, really - as this is also just part of feature engineering.

docs/freqai-data-handling.md
th0rntwig: Agreed. Moving 'file structure' to the Dev page, 'Returning data from training' to 'Feature engineering', and the rest to 'Running FreqAI'.

docs/freqai-data-handling.md
th0rntwig: Removed the 'Data handling' page all together and moved things as per comment above.

docs/freqai-parameter-table.md
th0rntwig: Added.

docs/freqai-running.md
th0rntwig: I disagree. The stratification isn't a feature but a way to affect the training of the model, which is part of how to run FreqAI.

docs/freqai-outlier-detection.md
th0rntwig: I agree. They were on that page before so I'm moving them back.

docs/freqai-feature-engineering.md
th0rntwig: I both agree and disagree with this. Using "you/your" makes the docs more personal which is nice when you read them as a user, but not only users will read the docs and for non-users "you/your" becomes weird. Personally, I prefer to keep it as it is as it sounds more professional and generalized.

xmatthias: It's fine _for now_ - but for the future, i'd appreciate if we can align the freqAI docs to the remaining freqtrade docs, using you/your.
This is a project written by volunteers - for people using it for free. 
Right now, freqAI stands out by using "the user" - so it feels detached from the remaining project, as it's using terminology not matching the remaining project docs.

It intentionally shouldn't sound too professional - and docs that use "the user" always feel like "the write though he is something better, not addressing the reader directly".
You often see this in docs for paid software, too - where it always feels like it's the company talking over your head.

th0rntwig: It's a very fair point and I fully agree that we should not (seemingly) be talking over peoples heads. I have changed "the user/they/their" to "you/your" throughout except for when referring to examples were a "fictive user" is doing things. If I missed anything let me know and I'll fix it in the next commit.

robcaulk: Ok, let's try this method of avoiding you/user as much as possible.

xmatthias (approved):
Great job, thanks for your efforts to both of you! :100: 

LGTM :+1:

PR #7411, opened 2022-09-13
Changing the default OHLCV data from json to hdf5

yu3ufe:
since hdf5 data is faster than json files while processing in backtesting mode, it's better to use as the default data type

### DIFF
--- a/docs/data-download.md
+++ b/docs/data-download.md
@@ -183,15 +183,15 @@ Freqtrade currently supports 3 data-formats for both OHLCV and trades data:
 * `jsongz` (a gzip-zipped version of json files)
 * `hdf5` (a high performance datastore)
 
-By default, OHLCV data is stored as `json` data, while trades data is stored as `jsongz` data.
+By default, OHLCV data and trades data are stored as `hdf5` data.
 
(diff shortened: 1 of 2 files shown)


xmatthias:
While i don't disagree on the fact - this is a breaking change, which would cause a lot of head-aches for people updating without having this configured to json explicitly (freqtrade would fail to find existing json data for no good reason).

this means we'll need a good way of a transition (either automatic conversion, auto-fallback to json or some very clear warning that this might be the problem).
I currently however don't see how this transition would work - and i'm hesitant about auto-conversion - as that'll potentially take quite some time if you have a large dataset (and i don't really want to have "if not found, search for json" code all over the place)

You can also make an argument that jsongz would be smaller from a storage perspective - so that can be more ideal as well if storage size is your restriction. it can also be faster than hdf5 in some instances - but hdf5 beats it if you do selective backtesting (not using the full timerange).

For trades - hdf5 is for sure the wrong format - it's not actively used for anything other than data-downloads, and jsongz should be the default here (if we're looking to change the default).

yu3ufe:
Ok, I'll close this for now until we find a good solution for it

PR #7412, opened 2022-09-14
Fixed a bug that prevents clearing old models

initrv:
## Summary

Fix the issue of purging old models when the "user_data/models/unique-id" directory contains sub-directories with names that do not match a regular expression "sub-train-(\w+)_(\d{10})".
For example, "user_data/models/unique-id/tensorboard".

Solve the issue: #___

## Quick changelog

- Old models are purged correctly if the model dir contains dirs with names that do not match the regex


xmatthias (approved):
Great spot! :100: 

LGTM :+1: 
(ci has a generic error which needs a separate fix).

PR #7421, opened 2022-09-15
Add `spice_rack` to FreqAI

robcaulk:
The goal of the `FreqAI` `spice_rack` is to simplify the user interaction with `FreqAI` to be equivalent to a standard talib indicator. With this in mind, users should treat the `spice_rack` as just another indicator to help their personal strat.

Users of `spice_rack` do not need to worry about setting `FreqAI` parameters, deciding models - they dont need to even concern themselves with feature engineering (no concerns with `populate_any_indicators`). 

**The goal is simplicity.** 

The `spice_rack` is aimed at users who do not wish to deal with setting up `FreqAI` confgs, but instead prefer to interact with `FreqAI` similar to a `talib` indicator. In this case, the user can instead simply add two keys to their config:

```json
    "freqai_spice_rack": true, 
    "freqai_identifier": "spicey-id",
```

Which tells `FreqAI` to set up a pre-set `FreqAI` instance automatically under the hood with preset parameters. Now the user can access a suite of custom `FreqAI` supercharged indicators inside their strategy:

```python
        dataframe['dissimilarity_index'] = self.freqai.spice_rack(
            'DI_values', dataframe, metadata, self)
        dataframe['maxima'] = self.freqai.spice_rack(
            '&s-maxima', dataframe, metadata, self)
        dataframe['minima'] = self.freqai.spice_rack(
            '&s-minima', dataframe, metadata, self)
        self.freqai.close_spice_rack()  # user must close the spicerack
```

Users can then use these column, in concert with all their own additional indicators added to `populate_indicators`, in their entry/exit criteria and strategy callback methods the same way as any typical indicator. For example:

```python
    def populate_entry_trend(self, df: DataFrame, metadata: dict) -> DataFrame:

        df.loc[
            (
                (df['dissimilarity_index'] < 1) &
                (df['minima'] > 0.1)
            ),
            'enter_long'] = 1

        df.loc[
            (
                (df['dissimilarity_index'] < 1) &
                (df['maxima'] > 0.1)
            ),
            'enter_short'] = 1

        return df

    def populate_exit_trend(self, df: DataFrame, metadata: dict) -> DataFrame:

        df.loc[
            (
                (df['dissimilarity_index'] < 1) &
                (df['maxima'] > 0.1) 
            ),

            'exit_long'] = 1

        df.loc[
            (

                (df['dissimilarity_index'] < 1) &
                (df['minima'] > 0.1)
            ),
            'exit_short'] = 1

        return df
```


freqtrade/freqai/utils.py
```diff
@@ -85,6 +93,83 @@ def get_required_data_timerange(
...
+        (s + "_" + tf) for s in ["date", "open", "high", "low", "close", "volume"]
+    ]
+    df = df.drop(columns=skip_columns)
+    if set_generalized_indicators:
+        df["%-day_of_week"] = (df["date"].dt.dayofweek + 1) / 7
+        df["%-hour_of_day"] = (df["date"].dt.hour + 1) / 25
+        df["&s-minima"] = 0
+        df["&s-maxima"] = 0
+        min_peaks = argrelextrema(df["close"].values, np.less, order=80)
+        max_peaks = argrelextrema(df["close"].values, np.greater, order=80)
```
xmatthias, line 165: are you sure this is working here?

iirc (been a while since i looked at this though), this doesn't calculate min/max values in a rolling fashion - but is using the full dataframe at any time - therefore causing an implicit lookahead (and with that - bias in backtesting).

freqtrade/freqai/spice_rack/lightgbm_config.json
```diff
@@ -0,0 +1,36 @@
...
+        "identifier": "spicy-id",
+        "feature_parameters": {
+            "include_timeframes": [
+                "3m",
+                "15m",
+                "1h"
+            ],
+            "include_corr_pairlist": [
+                "BTC/USDT",
+                "ETH/USDT"
```
xmatthias: i'm not sure this will work with a static config (at least not like this).

If i start "spice_rack" (without additional config, obviously) - i'd expect the corr_pairlist to actually work for my scenario.
If i'm now starting this on gateio futures (which doesn't have BTC/USDT - but BTC/USDT:USDT) - it'll fail as the informative pairs are not available.

docs/freqai.md
```diff
@@ -807,3 +807,85 @@ Code review, software architecture brainstorming:
...
+            ),
+            'exit_short'] = 1
+
+        return df
+```
+
+The user does need to ensure their `informative_pairs()` contains the following (users can add their own `informative_pair` needs to the bottom of this template):
+
+```python
+    def informative_pairs(self):
```
xmatthias: i think the docs are missing a crucial piece.

Based on the docs - i can use any strategy (say - the default sample strategy), throw in the 2 config options below - and use the indicators.

That'll however mean that `self.freqai.start()` will never be called (nobody told me to add that to `populate_indicators()`.

as that's "default" code (every freqAI is expected to do the same) - it might be worth moving that to the "backend" - no longer requiring the user to call `.freqai.start()`.

The same could (probably should or must) go for informative-pairs - the moment we ask a user "just copy/paste this segment into your strategy" - it's something that'll become a problem later on.
Imagine we see a bug in this segment or the need to change something for whatever reason (this code is now copy/pasted in everyone's strategy - good luck migrating without a huge mess).

---

I think we should consider moving these 2 things to the backend (as a separate PR - which ideally is merged before this one) - while a breaking freqAI change - it'll be a huge interface simplification.
let me know what you think about that ...

robcaulk: I don't understand, why are you saying that the spice_rack user needs to call `self.freqai.start()`? That's not how spice_rack interface is designed - see how spice_rack calls start in the backend...

Why are you suggesting normal FreqAI users shouldn't have to call freqai.start()? 

Both usecases are well defined and require their respective interactions with spice_rack() or start(). I don't see why we would hide either of these interactions from the user. Please elaborate. 

Re: informative pairs:

While I don't disagree with your point. Well we could do just as you say, and it is easy to add - and I considered it during this PR (after all it's the same code as I'm doing for `populate_any_indicators` 
). However, it means we take full control over `informative_pairs`, and as such, the spice_rack user can't add their own informative needs. Thus, imo it's best to allow them to add what ever they need to `inforative_pairs` since the idea is that spice_rack is just an another indicator.

Maybe an alternative solution is to add a `freqai_informative_pairs()` to `interface.py` and ensure it is always called before the typical `informative_pairs`?

freqtrade/freqai/spice_rack/lightgbm_config.json
robcaulk: Fair point - do you see any way around this without asking the user to define exchange specific config params? Goal here is simplicity, so for me I would keep a set for all supported exchanges and call those accordingly.

freqtrade/freqai/utils.py
robcaulk: It's important to remember that FreqAI targets *need* to look into the future. 

Nonetheless, live/backtesting do have slightly different behaviors (not hugely). For live mode, FreqAI trims the data frame to the training period and feeds it into `populate_any_indicators`

In backtesting,full df gets maxmins from `populate_any`, then the training gets trimmed to the window and we train on those labels. Thus the edge labels of each window are "more accurate". But that edge is no more than 1 label on each side. Typically users are training on thousands of labels per train window so this accounts for less than 1%.

We could try to make backtesting emulate live closer by slicing data frame and then `populate_any` For each window - It may need to be a separate PR though depending on ramifications.

freqtrade/freqai/spice_rack/lightgbm_config.json
xmatthias: >  I would keep a set for all supported exchanges

for binance, that'd then be
BTC/USDT, BTC/BNB, BTC/ETH, BTC/BUSD, BTC/...  (obviously the same combinations with ETH or any other random pair).

You can't realistically maintain this as separate configs (that's 4 configs for binance spot alone - and we currently support 6 spot and 3 futures exchanges ... 

---

you need to do this dynamically i think (which shouldn't be too hard, either - as you should be able to filter`markets` - similar to the filters list-pairs / list-markets do (with both quote and base - obviously using stake-currency).

docs/freqai.md
xmatthias: > However, it means we take full control over informative_pairs, 

that's clearly not the case if done properly (you shouldn't replace the function as you're doing with `populate_any_indicators()` now).
the `@informative` decorator does something similar - so there's one place we can hook this into - and it's well defined

``` pseudo-code
if freqai: 
  # add these pair/timeframe combos unless they're already in the list
```


> Both usecases are well defined and require their respective interactions with spice_rack() or start().

Well i've so far not seen any usecase where `freqai.start()` is not at the top of `populate_indicators()` - so maybe the requirement/usecase for this isn't written anywhere I've looked so far.
based on my information, i therefore don't think that's a requirement (it's dummy code we require people to copy around) - but an interface decision.

robcaulk: Re: decorator : 

Ok I will need to look into it and get back to you with a reply. Sounds like it is a good solution.

Re: location of FreqAI.start()

It can be put anywhere in populate_indicators(), there's nothing forcing it to be set at the top. 

So I'm still confused as to what you are asking for. are you saying we should consider removing these from populate indicators entirely?

freqtrade/freqai/spice_rack/lightgbm_config.json
robcaulk: I'm not sure I agree: why would I need to support *all* stake currencies for a corr_pairlist? All we need is the closest available two pairs to BTC/USDT and ETH/USDT. These are just corr_pairs - they don't need to follow any user stake or anything like that. 

That said, your solution is likely still the right one as we just need to filter the market and find closest string to those two. 

I'll have a look at implementing that.

docs/freqai.md
xmatthias: >  are you saying we should consider removing these from populate indicators entirely?

After a quick review here - my assumption was I'll need `freqai.start()` in indicators. that's not the case - which is fine.
I'm ... putting it up for discussion though. 
For now - it seems to be a "users must copy this to their strategy" solution - which is usually a solution we don't need.

> Ok I will need to look into it and get back to you with a reply. Sounds like it is a good solution.

don't worry - i can implement this (separate PR) as i know roughly where it needs to go

robcaulk: >For now - it seems to be a "users must copy this to their strategy" solution - which is usually a solution we don't need.

Honestly - it is a fair point. We could automate everything from the `populate_indicators` side and the user just magically *gets* the predictions etc inside their dataframe, which they then append their own indicators to. 

My only concern is that it increases opacity which transitively increases the barrier to entry for new users. 

Typical FT user is someone who writes:

```
```

So i like that they can see "ok my dataframe goes into freqai, and comes out with new stuff attached

```
dataframe = self.freqai.start(dataframe)
```

(Where `spice_rack` aims to increase transparency even more)

Now we ask them to essentially "trust" that some stuff is gonna be in their dataframe. I am just unsure about if it will really improve the interface. But I want to hear your thoughts and responses to these points. May be worth asking the community directly - it is their code after all. 

I find the same issue with the `@informative` decorator. It kind "magically" makes stuff happen, where as a long time FT user I liked having control over where and when informatives were accessbile. This was due to three things:

1) harder to relearn a new method which requires reading up on
2) I now have to trust the authors of the decorator that i am getting the dataframe that I wanted.
3) the actual strategy code requires a bit of rethinking to accommodate the @informative decorator.

I therefore, never switched to using the @informative decorator. But, I am also not really using FT strats too intensely these days so my judgement isn't necessarily representative of the majority.

xmatthias: `@informative` is a helper - simplifying the usecase (you no longer need to forget to put the informative pair in 2 locations (where you use it, and in the `informative_pairs()` location).

We didn't remove the old method though - so everyone is free to use whatever they like.
In this case it's however where we ask users to "do the same thing in every strategy" - which provides 0 benefit / 0 flexibility (if you forget this part - it'll fail with probably random errors).

I'll have a go at the indicators stuff - you can then let me know what you think (it shouldn't be a breaking change, either).

> (Where spice_rack aims to increase transparency even more)

for now - that's what i don't really see, but rather the opposite (hiding stuff from the beginner user is not a bad thing though).
For now, it's populating some magic columns for me - but the exact configs are unknown unless i look into the code.

Now that's a problem for another day - but we'll for sure have to have the exact config / setup in the docs - so i could go from a working spice-rack to a regular freqAI strategy once i'm "grown up" - and want to configure my own parameters.

> I now have to trust the authors of the decorator that i am getting the dataframe that I wanted.

this is exactly what spice-rack brings - users have to trust us / freqtrade / freqAI to give them the dataframe they never asked for.
but in the end, it's the same with the whole application - users have to trust the authors / reviewers to not break the application in unexpected ways, change behavior suddenly - or intentionally do malicious stuff. Without the trust - every user must read (and understand) the full sourcecode - while it's a possibility with open source - it's also not the reality of open source.

robcaulk: > Now that's a problem for another >day - but we'll for sure have to have >the exact config / setup in the docs - >so i could go from a working spice->rack to a regular freqAI strategy once >i'm "grown up" - and want to >configure my own parameters.

Agreed 100%

> For now, it's populating some magic > columns for me - but the exact >configs are unknown unless i look >into the code.

> this is exactly what spice-rack >brings - users have to trust us / >freqtrade / freqAI to give them the >dataframe they never asked for.

I'll say this: spice rack lets users ask for an exact indicator and add that to their dataframe as a column. That is pure transparence. Sure it's black box generated data but at least the command lets them decide what they want. It isn't much different with some of those trading view API indicators. the goal is that they look for patterns themselves. As I explicitly point out in the doc - this is meant to be used as a supporting indicator only. 

This is geared toward the users who just want to get a generalized config and continue with their normal TA. Luckily we now have enough experience to set up a generalized config for them. 

I do, however, see spice rack evolving to let the user select from, say, 4 premade configs. I already leave that evolution open here with the config param "freqai_config" which defaults to lightgbm.

robcaulk: > I'll have a go at the indicators stuff - you can then let me know what you think (it shouldn't be a breaking change, either).

Sounds good, I appreciate the help

freqtrade/freqai/spice_rack/lightgbm_config.json
robcaulk: This should be handled with recent commit:

https://github.com/freqtrade/freqtrade/pull/7421/commits/ff300d5c85adcb59d1abfda99bf56e057a8a5950

freqtrade/freqai/utils.py
xmatthias: i'm just ... asking if we're sure it's working here - not saying it's not ;) 

> We could try to make backtesting emulate live closer by slicing data frame and then populate_any For each window - It may need to be a separate PR though depending on ramifications.

I've tried this once (for each candle, actually) - and the performance hit is VERY high - taking you from a 10s backtest to 5-6 minutes easily.

freqtrade/freqai/spice_rack/lightgbm_config.json
```diff
@@ -0,0 +1,37 @@
...
+        "identifier": "spicy-id",
+        "feature_parameters": {
+            "include_timeframes": [
+                "30m",
+                "1h",
+                "4h"
+            ],
+            "include_corr_pairlist": [
+                "BTC/USD",
+                "ETH/USD"
```
xmatthias, line 17: To avoid confusion, you may want to leave this empty / add a comment that this will be replaced.

freqtrade/freqai/freqai_interface.py
```diff
@@ -138,7 +138,7 @@ def start(self, dataframe: DataFrame, metadata: dict, strategy: IStrategy) -> Da
             dk = self.start_backtesting(dataframe, metadata, self.dk)
 
         dataframe = dk.remove_features_from_df(dk.return_dataframe)
-        self.clean_up()
+        # self.clean_up()
```
xmatthias, line 145: ?

freqtrade/freqai/freqai_interface.py
```diff
@@ -685,6 +685,18 @@ def get_init_model(self, pair: str) -> Any:
 
         return init_model
 
+    def spice_rack(self, indicator: str, dataframe: DataFrame,
+                   metadata: dict, strategy: IStrategy) -> NDArray:
+        if not self.spice_rack_open:
+            dataframe = self.start(dataframe, metadata, strategy)
```
xmatthias, line 738: Am i reading this right that this would replace the need to copy `self.freqai.start()`into the strategy's populate_indicators?

docs/freqai.md
```diff
@@ -794,3 +794,68 @@ Code review, software architecture brainstorming:
...
+
+Which tells `FreqAI` to set up a pre-set `FreqAI` instance automatically under the hood with preset parameters. Now the user can access a suite of custom `FreqAI` supercharged indicators inside their strategy:
+
+```python
+        dataframe['dissimilarity_index'] = self.freqai.spice_rack(
+            'DI_values', dataframe, metadata, self)
+        dataframe['maxima'] = self.freqai.spice_rack(
+            '&s-maxima', dataframe, metadata, self)
+        dataframe['minima'] = self.freqai.spice_rack(
+            '&s-minima', dataframe, metadata, self)
```
xmatthias: What are the possible values for the first parameter? (and what do they contain)?

freqtrade/strategy/interface.py
```diff
@@ -145,12 +145,28 @@ def __init__(self, config: dict) -> None:
                 self._ft_informative.append((informative_data, cls_method))
 
     def load_freqAI_model(self) -> None:
-        if self.config.get('freqai', {}).get('enabled', False):
+        spice_rack = self.config.get('freqai_spice_rack', False)
+        if self.config.get('freqai', {}).get('enabled', False) or spice_rack:
+            spice_rack = self.config.get('freqai_spice_rack', False)
```
xmatthias: unnecessary line - 148 already does exactly this ;)

freqtrade/strategy/interface.py
```diff
@@ -145,12 +145,28 @@ def __init__(self, config: dict) -> None:
...
+                self.config = self.setup_freqai_spice_rack(self.config)
             # Import here to avoid importing this if freqAI is disabled
             from freqtrade.freqai.utils import download_all_data_for_training
             from freqtrade.resolvers.freqaimodel_resolver import FreqaiModelResolver
             self.freqai = FreqaiModelResolver.load_freqaimodel(self.config)
-            self.freqai_info = self.config["freqai"]
+            if not self.process_only_new_candles:
+                logger.warning('User set process_only_new_candles to false, '
+                               'FreqAI requires true. Changing to true.')
+                self.process_only_new_candles = True
```
xmatthias, line 161: i understand the need to run population only once per candle - but (especially with spice_rack) - i could see the user desire to combine spice_rack with orderbook data (for example) - which would be disabled by this.

we'll need to look at this again i think (probably separate from this PR) - maybe we can make freqAI independent of this - but have FreqAI only do it's heavy operations once per candle.
(might be as easy as keeping a "last_candle_date" per pair)...

freqtrade/freqai/spice_rack/lightgbm_config.json
```diff
@@ -0,0 +1,37 @@
...
+                20
+            ]
+        },
+        "data_split_parameters": {
+            "test_size": 0,
+            "random_state": 1
+        },
+        "model_training_parameters": {
+            "n_estimators": 800
+        }
```
xmatthias, line 35: looking at this now - i think we should split this parameter per model type - to allow for flexible model switches.

We currently can't run the same strategy/config against catboost - and then switch to lightgbm (or xgboost) without changing this config segment (as the parameters are different).
if we had these named per model (or nested per modelname?) - we could more easily switch around

For example:
``` json
        "model_training_parameters": {
            "LightGBM": {
                "n_estimators": 800
            },
            "Catboost": {"whatever": ...}
        }
```

Will be something for a separate PR though

freqtrade/strategy/interface.py
```diff
@@ -173,6 +190,60 @@ def shutdown(self, *args, **kwargs):
...
             self.freqai = DummyClass()  # type: ignore
 
+    def setup_freqai_spice_rack(self, config: dict) -> Dict[str, Any]:
+        import difflib
+        import json
+        from pathlib import Path
+        auto_config = config.get('freqai_config', 'lightgbm_config.json')
+        with open(Path('freqtrade') / 'freqai' / 'spice_rack'
+                  / auto_config) as json_file:
+            freqai_config = json.load(json_file)
```
xmatthias: This is wrong / will not work depending on the setup. 
it's assuming you run freqtrade from the repository's root directory.

This is probably true for many users - but it's not a must - you can install freqtrade in any environment (also via `pip install freqtrade`) - so the directory "freqtrade" may not be available in the CWD.

You'll have to navigate to that via `Path(__file__).parent` - although i'm not 100% sure that'll work from this location - as you're in the strategy interface - which means the location of "__file__" may be the actual place of the strategy.

I think the best thing would be to move this method to freqai/utils (or a separate file in spice_rack) - and pass in what's needed (most likely just the dataprovider or the exchange object).
That'll also make the "relative path navigation" easier.

freqtrade/freqai/utils.py
```diff
@@ -85,6 +93,83 @@ def get_required_data_timerange(
     return data_load_timerange
 
 
+def auto_populate_any_indicators(
```
xmatthias, line 97: i'm a bit concerned about this - we should call this functions from tests to ensure it's working (and keeps working).
we assume people trust this with just the "flick of a switch" - so we should be sure it'll not break at random ...

freqtrade/freqai/freqai_interface.py
robcaulk: Right. I am hoping the user interface of `spice_rack` is clear in the doc.

freqtrade/freqai/spice_rack/lightgbm_config.json
robcaulk: No, it is used to find closest pairs in the market. It cannot be empty. 

Adding a comment to a json file?

freqtrade/freqai/freqai_interface.py
robcaulk: Coming back to the spice rack for more spices requires that we don't force delete the data kitchen after the first spice is extracted. 

`self.dk/model` gets replaced automatically anyways next time the strategy comes back to access spice rack when `spice_rack_open == False`.

freqtrade/strategy/interface.py
robcaulk: Ya it was duplicated during dev.

freqtrade/strategy/interface.py
robcaulk: Yes, `spice_rack`, and FreqAI in general only work with `process_only_new_candles = True`. Agreed it is beyond this PR.

freqtrade/freqai/spice_rack/lightgbm_config.json
robcaulk: I think you are confused about the target user of the `spice_rack`. `spice_rack` users don't decide the model. The name of the file is `lightgbm_config.json` afterall. They get a pre-made freqai instance under the hood. That is it. 

I don't see why you are trying to give this flexibility - if users want to start setting model params and model types then they can do that by just using normal `FreqAI`. `spice_rack` has one goal **simplicity**. It is for users who don't want to mess with model parameters and configuration files.

robcaulk: And by the way, `n_estimators` *does* work for catboost, xgboost, and lightgbm. This config would technically work for all our current regressors/classifiers.

But that doesnt change the fact that we decide the model type too for the user - so it is a moot point.

freqtrade/strategy/interface.py
robcaulk: Good catch, I will try to work it out.

freqtrade/freqai/utils.py
robcaulk: Yes, this needs a test for sure.

freqtrade/freqai/utils.py
robcaulk: I don't think it would in our case - it all happens inside the `IFreqaiModel` the strategy/backtest sends the full dataframe in and gets the full dataframe out. IT is just  about *when* freqai slices it. I will likely change this in a separate PR because it is objectively better and closer to live.

docs/freqai.md
robcaulk: Yes, the available values need to be listed in the docs with associated descriptions to help users understand it.

freqtrade/strategy/interface.py
robcaulk: This should be resolved in latest commits.

freqtrade/freqai/freqai_interface.py
xmatthias: it's fine if we aim to not clean up - i'm just raising a question if this can become a problem.
Usually when you have a "cleanup" call - it's to clean up (avoid memory leaks) - if that's not the case, no worries

freqtrade/freqai/spice_rack/lightgbm_config.json
xmatthias: it's not 100% relevant for spice_rack (although i think users may want to pick "run on GPU" if one is available) - and maybe not for this particular config combination.

HOWEVER - it is very relevant for regular freqAI :)

freqtrade/freqai/spice_rack/lightgbm_config.json
xmatthias: you can use rapidjson - and add `rapidjson.PM_COMMENTS | rapidjson.PM_TRAILING_COMMAS` as flags (the same config is used to read regular configurations).

apart from the fact that rapidjson is faster to read json files - - it'll also allow comments and trailing commas (a pending pain in json files)

freqtrade/freqai/freqai_interface.py
robcaulk: No this clean_up call is demonstrative only. I'll mark resolved.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1267,3 +1268,11 @@ def check_if_backtest_prediction_exists(
                 f"Could not find backtesting prediction file at {path_to_predictionfile}"
             )
         return file_exists
+
+    def spice_extractor(self, indicator: str, dataframe: DataFrame) -> npt.NDArray:
+        if indicator in dataframe:
```
xmatthias: `... in dataframe.columns:` ?

docs/freqai.md
```diff
@@ -794,3 +794,68 @@ Code review, software architecture brainstorming:
 Beta testing and bug reporting:
 @bloodhunter4rc, Salah Lamkadem @ikonx, @ken11o2, @longyu, @paranoidandy, @smidelis, @smarm,
 Juha Nykänen @suikula, Wagner Costa @wagnercosta
+
+## Using the `spice_rack`
+
+<!-- Dont forget this section during the doc reorg! -->
+
+The `spice_rack` is aimed at users who do not wish to deal with setting up `FreqAI` confgs, but instead prefer to interact with `FreqAI` similar to a `talib` indicator. In this case, the user can instead simply add two keys to their config:
```
xmatthias: (mostly / in part copied from https://github.com/freqtrade/freqtrade/pull/7442#issuecomment-1253213755)

We imagine / advertise spice_rack to be "just another indicator you can use" - so there's nothing really preventing the following (which i think is a valid usecase, btw).

``` python
        dataframe['dissimilarity_index'] = self.freqai.spice_rack('DI_values', dataframe, metadata, self)
        informative_1h['dissimilarity_index'] = self.freqai.spice_rack('DI_values', informative_1h, metadata, self)
```

user would like to have the dissimilarity index - but calculated against the informative 1h pair - which may (or may not) have a different length.

---

we seem to place `spice_rack` as an (intelligent) indicator of sorts - just like ta-lib.
ta-lib can do this ... can spice_rack too (or should it)?

if not - i think we need to change what we're trying to bring here - as it's got quite some limitations ATM (only for the main dataframe, ...) which is not a problem i think.

we just need to be aware that users will find ways to abuse this - (which is a good thing, as this usually results in better code).

freqtrade/freqai/data_kitchen.py
robcaulk: Yes, that would be safer. But as is, it defaults to iterate on the columns iinm.

freqtrade/freqai/spice_rack/lightgbm_config.json
robcaulk: I don't disagree with the idea of adding flexibility in a single config to switch between models (for normal FreqAI) or for adding GPU support for `spice_rack`. But I don't believe either of these are in the scope of the present PR.

As lightgbm is not compiled by default with GPU - and `spice_rack` is designed to run on lightgbm (since it is the most flexible one of the three, running on cpus, arm, and considered to be the "fastest"), I would leave any integration of further models/GPU spice_rack support to future PRs. 

Goal here is to provide a very simplified interface to an additional indicator. We reiterate in the doc that `spice_rack` should *NOT* be used alone. It is but another indicator to support a strat - nothing more - nothing less. Hence highly efficient and low levels of user attention.

robcaulk:
Closing this, as I personally don’t have the interest in continuing further and I haven’t seen any interest from community or other devs.

docs/freqai.md
robcaulk: Sounds like you have some ideas for how you would like this to look. Personally my time ends here on this PR. You are welcome to reopen it if you want it integrated.

PR #7428, opened 2022-09-17
Informative freqai

xmatthias:
## Summary

As discussed in https://github.com/freqtrade/freqtrade/pull/7421#discussion_r973207174 - this will remove the need for freqAI users to have `informative_pairs()` implemented with copy/pasted code.
This is 100% predictable - and we can handle this easily.

That said - users can still implement `informative_pairs()` for other informative pairs.
Leaving a current strategy untouched will also work - as we do a "set/list" at the end, removing potential duplicates.

This brings the advantage that we can now fully test this code - and improve it should this be necessary (as well as the reduced dummy-code need in strategies.

## Quick changelog

* Move informative pairs combination for freqAI strategies to freqtrade code, combining `include_corr_pairlist * include_timeframes`.
* Remove freqai `informative_pairs` samples

### DIFF
--- a/tests/freqai/test_freqai_interface.py
+++ b/tests/freqai/test_freqai_interface.py
@@ -320,7 +320,7 @@ def test_principal_component_analysis(mocker, freqai_conf):
 
 @pytest.mark.parametrize('timeframes,corr_pairs', [
     (['5m'], ['ADA/BTC', 'DASH/BTC']),
-    (['5m'], ['ADA/BTC', 'DASH/BTC']),
+    (['5m'], ['ADA/BTC', 'DASH/BTC', 'ETH/USDT']),
     (['5m', '15m'], ['ADA/BTC', 'DASH/BTC', 'ETH/USDT']),
 ])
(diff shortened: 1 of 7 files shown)


tests/freqai/test_freqai_interface.py
```diff
@@ -315,3 +316,27 @@ def test_principal_component_analysis(mocker, freqai_conf):
     assert Path(freqai.dk.data_path / f"{freqai.dk.model_filename}_pca_object.pkl")
 
     shutil.rmtree(Path(freqai.dk.full_path))
+
+
+@pytest.mark.parametrize('timeframes,corr_pairs', [
+    (['5m'], ['ADA/BTC', 'DASH/BTC']),
+    (['5m'], ['ADA/BTC', 'DASH/BTC']),
```
robcaulk: Is this purposefully duplicated?

robcaulk (approved):
That is sexy.

tests/freqai/test_freqai_interface.py
xmatthias: yeah ... no :D

PR #7431, opened 2022-09-17
Add plot feature importance

initrv:
## Summary

Give the user a function to visually display the importance of features

Solve the issue: #___

## Quick changelog

- Added optional configuration key "plot_feature_importance"

## What's new?
features-plot

### DIFF
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -20,6 +20,7 @@ from freqtrade.exceptions import OperationalException
 from freqtrade.exchange import timeframe_to_seconds
 from freqtrade.freqai.data_drawer import FreqaiDataDrawer
 from freqtrade.freqai.data_kitchen import FreqaiDataKitchen
+from freqtrade.freqai.utils import plot_feature_importance
 from freqtrade.strategy.interface import IStrategy
 
 
@@ -554,6 +555,14 @@ class IFreqaiModel(ABC):
         dk.find_features(unfiltered_dataframe)
 
         model = self.train(unfiltered_dataframe, pair, dk)
+
+        if self.freqai_info["feature_parameters"].get("plot_feature_importance", False):
+            plot_feature_importance(
+                model=model,
+                feature_names=dk.training_features_list,
+                pair=pair,
+                train_dir=dk.data_path
+            )
 
         self.dd.pair_dict[pair]["trained_timestamp"] = new_trained_timerange.stopts
         dk.set_new_model_names(pair, new_trained_timerange)
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -556,20 +556,15 @@ class IFreqaiModel(ABC):
 
         model = self.train(unfiltered_dataframe, pair, dk)
 
-        if self.freqai_info["feature_parameters"].get("plot_feature_importance", False):
-            plot_feature_importance(
-                model=model,
-                feature_names=dk.training_features_list,
-                pair=pair,
-                train_dir=dk.data_path
-            )
-
         self.dd.pair_dict[pair]["trained_timestamp"] = new_trained_timerange.stopts
         dk.set_new_model_names(pair, new_trained_timerange)
         self.dd.pair_dict[pair]["first"] = False
         if self.dd.pair_dict[pair]["priority"] == 1 and self.scanning:
             self.dd.pair_to_end_of_training_queue(pair)
         self.dd.save_data(model, pair, dk)
+
+        if self.freqai_info["feature_parameters"].get("plot_feature_importance", False):
+            plot_feature_importance(model, pair, dk)
 
         if self.freqai_info.get("purge_old_models", False):
             self.dd.purge_old_models()
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -563,7 +563,7 @@ class IFreqaiModel(ABC):
             self.dd.pair_to_end_of_training_queue(pair)
         self.dd.save_data(model, pair, dk)
 
-        if self.freqai_info["feature_parameters"].get("plot_feature_importance", False):
+        if self.freqai_info["feature_parameters"].get("plot_feature_importance", True):
             plot_feature_importance(model, pair, dk)
 
         if self.freqai_info.get("purge_old_models", False):
--- a/freqtrade/freqai/utils.py
+++ b/freqtrade/freqai/utils.py
@@ -191,5 +191,4 @@ def plot_feature_importance(model: Any, pair: str, dk: FreqaiDataKitchen,
         fig = add_feature_trace(fig, fi_df_worst, 2)
         fig.update_layout(title_text=f"Best and worst features by importance {pair}")
 
-        store_plot_file(fig, f"{dk.model_filename}-{label}.html", dk.data_path,
-                        include_plotlyjs="cdn")
+        store_plot_file(fig, f"{dk.model_filename}-{label}.html", dk.data_path)
--- a/freqtrade/freqai/utils.py
+++ b/freqtrade/freqai/utils.py
@@ -1,13 +1,9 @@
 import logging
 from datetime import datetime, timezone
-# for plot_feature_importance
 from pathlib import Path
 
 import numpy as np
 import pandas as pd
-import plotly.graph_objects as go
-import plotly.io as pio
-from plotly.subplots import make_subplots
 
 from freqtrade.configuration import TimeRange
 from freqtrade.data.dataprovider import DataProvider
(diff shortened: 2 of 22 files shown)


initrv:
We can also reduce the .html file size from 4M to ~20kb but make it dependent on external resource using `include_plotlyjs="cdn"` somewhere.

freqtrade/plot/plotting.py
```diff
@@ -614,7 +615,7 @@ def store_plot_file(fig, filename: str, directory: Path, auto_open: bool = False
 
     _filename = directory.joinpath(filename)
     plot(fig, filename=str(_filename),
-         auto_open=auto_open)
+         auto_open=auto_open, include_plotlyjs=include_plotlyjs)
```
xmatthias: we're intentionally importing `plot()` from `plotly.offline` - so we avoid any connection to plotly. 

that got several reasons (mainly around user's privacy) - so i think this setting is counter-productive to the approach taken to this.

robcaulk (approved):
LGTM

freqtrade/freqai/utils.py
```diff
@@ -132,3 +137,58 @@ def get_required_data_timerange(
...
+                    y=fi_df["feature_names"],
+                    orientation='h', showlegend=False
+                ), row=1, col=col
+            )
+        fig = make_subplots(rows=1, cols=2, horizontal_spacing=0.5)
+        fig = add_feature_trace(fig, fi_df_top, 1)
+        fig = add_feature_trace(fig, fi_df_worst, 2)
+        fig.update_layout(title_text=f"Best and worst features by importance {pair}")
+
+        store_plot_file(fig, f"{dk.model_filename}-{label}.html", dk.data_path)
```
xmatthias, line 194: i think this will store the result in the wrong location (depends where we expect it to end up at).

for model storage, we use `dk.data_path / f"{dk.model_filename}_svm_model.joblib"` (For example) - so they end up in a directory like `models/default/sub-train-CHZ_1663224320/...`.

while i can understand that we may not want them in the model dir itself (why not - the plot is JUST valid for this one model) - the current approach will store the model in "cwd" - which is wrong (we only store data in the `user_data_dir`).

The bigger problem is that "cwd" is not accessible on docker setups - so you're now generating plots in the container - which are never exposed (and can't be exposed as that location also contains a ton of other stuff) to the user system.

xmatthias (changes requested):
I don't think this works at the moment.

running the following (without a model) generates an exception - comming out of the new code.
`freqtrade trade --strategy FreqaiExampleStrategy --strategy-path freqtrade/templates --config config_examples/config_freqai.example.json --freqaimodel LightGBMRegressor`

It's basically using the backtesting sample from the docs - but changed to work for the "trade"  command.

the only change i made was to the configuration, changing the n_estimators to 10 so it can be tested quickly.

```
2022-09-18 08:34:08,239 - freqtrade.freqai.data_kitchen - INFO - SVM tossed 53 train points from 4809 total points.
2022-09-18 08:34:08,249 - freqtrade.freqai.data_kitchen - INFO - SVM tossed 25 test points from 2370 total points.
2022-09-18 08:34:08,736 - freqtrade.freqai.base_models.BaseRegressionModel - INFO - Training model on 569 features
2022-09-18 08:34:08,736 - freqtrade.freqai.base_models.BaseRegressionModel - INFO - Training model on 4756 data points

[1]     valid_0's l2: 0.0149567
[2]     valid_0's l2: 0.01391
[3]     valid_0's l2: 0.0130216
[4]     valid_0's l2: 0.0122191
[5]     valid_0's l2: 0.0115642
[6]     valid_0's l2: 0.0108845
[7]     valid_0's l2: 0.0103883
[8]     valid_0's l2: 0.00986342
[9]     valid_0's l2: 0.00938445
[10]    valid_0's l2: 0.00895354
2022-09-18 08:34:09,265 - freqtrade.freqai.base_models.BaseRegressionModel - INFO - --------------------done training 1INCH/USDT--------------------
Exception in thread Thread-1:
  File "/home/xmatt/.pyenv/versions/3.9.7/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/home/xmatt/.pyenv/versions/3.9.7/lib/python3.9/threading.py", line 910, in run
    self._target(*self._args, **self._kwargs)
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/freqai/freqai_interface.py", line 202, in _start_scanning
    self.extract_data_and_train_model(
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/freqai/freqai_interface.py", line 567, in extract_data_and_train_model
    plot_feature_importance(model, pair, dk)
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/freqai/utils.py", line 160, in plot_feature_importance
    models[dk.label_list[0]]
KeyError: '&-s_close'
```

robcaulk:
> I don't think this works at the moment.
> 
> running the following (without a model) generates an exception - comming out of the new code. `freqtrade trade --strategy FreqaiExampleStrategy --strategy-path freqtrade/templates --config config_examples/config_freqai.example.json --freqaimodel LightGBMRegressor`
> 
> It's basically using the backtesting sample from the docs - but changed to work for the "trade" command.
> 
> the only change i made was to the configuration, changing the n_estimators to 10 so it can be tested quickly.
> 
> ```
> 2022-09-18 08:34:08,239 - freqtrade.freqai.data_kitchen - INFO - SVM tossed 53 train points from 4809 total points.
> 2022-09-18 08:34:08,249 - freqtrade.freqai.data_kitchen - INFO - SVM tossed 25 test points from 2370 total points.
> 2022-09-18 08:34:08,736 - freqtrade.freqai.base_models.BaseRegressionModel - INFO - Training model on 569 features
> 2022-09-18 08:34:08,736 - freqtrade.freqai.base_models.BaseRegressionModel - INFO - Training model on 4756 data points
> 
> [1]     valid_0's l2: 0.0149567
> [2]     valid_0's l2: 0.01391
> [3]     valid_0's l2: 0.0130216
> [4]     valid_0's l2: 0.0122191
> [5]     valid_0's l2: 0.0115642
> [6]     valid_0's l2: 0.0108845
> [7]     valid_0's l2: 0.0103883
> [8]     valid_0's l2: 0.00986342
> [9]     valid_0's l2: 0.00938445
> [10]    valid_0's l2: 0.00895354
> 2022-09-18 08:34:09,265 - freqtrade.freqai.base_models.BaseRegressionModel - INFO - --------------------done training 1INCH/USDT--------------------
> Exception in thread Thread-1:
> Traceback (most recent call last):
>   File "/home/xmatt/.pyenv/versions/3.9.7/lib/python3.9/threading.py", line 973, in _bootstrap_inner
>     self.run()
>   File "/home/xmatt/.pyenv/versions/3.9.7/lib/python3.9/threading.py", line 910, in run
>     self._target(*self._args, **self._kwargs)
>   File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/freqai/freqai_interface.py", line 202, in _start_scanning
>     self.extract_data_and_train_model(
>   File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/freqai/freqai_interface.py", line 567, in extract_data_and_train_model
>     plot_feature_importance(model, pair, dk)
>   File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/freqai/utils.py", line 160, in plot_feature_importance
>     models[dk.label_list[0]]
> KeyError: '&-s_close'
> ```

Ya that's just a bug that was caught by CI. Needs to set the dict entry to model :-)

freqtrade/freqai/utils.py
robcaulk: Not really - it should be saving it to dk.data_path, which is with all the model specific files. 

I don't see where or why this is saving to any general directory.

robcaulk: See the function arguments :)
https://github.com/freqtrade/freqtrade/blob/32c3f62934f97520490be1b009a5d0f5dc7dab4f/freqtrade/plot/plotting.py#L604

freqtrade/freqai/utils.py
```diff
@@ -132,3 +137,58 @@ def get_required_data_timerange(
...
```
xmatthias, line 194: we'll probably need some escaping for this. The current filename is `user_data/models/uniqe-id/sub-train-ALGO_1663501785/cb_algo_1663501785-&-s_close.html` - having an `&` in a filename is not really standard, and will cause errors (on linux, you need to quote this filename to work with it - and i don't really want to know how windows will be handling such filenames ...).

robcaulk: I'll remove the `&` from the string. Easy money ;)

freqtrade/freqai/freqai_interface.py
```diff
@@ -562,6 +563,9 @@ def extract_data_and_train_model(
             self.dd.pair_to_end_of_training_queue(pair)
         self.dd.save_data(model, pair, dk)
 
+        if self.freqai_info["feature_parameters"].get("plot_feature_importance", True):
```
xmatthias, line 566: should the default be False?
considering it's also not part of the docs (should probably be fixed ...) - people won't know to look for this - which makes the generated plot missed CPU cycles.

This param should also be added to the json schema (`constants.py`) ... so we know we'll only get a boolean through.

xmatthias (approved):
LGTM now 👍

freqtrade/freqai/freqai_interface.py
robcaulk: What about this point? It made sense to me - just didn’t get around to it yet.

xmatthias: Good catch - overlooked this missing point.
added in ea58c29ded45033eb175d51f7041facb3ec33fa7.

PR #7432, opened 2022-09-17
Normalise distances before Weibull fit

th0rntwig:
## Summary

The pairwise distances should be normalised before fitting a Weibull to their distributions for the fit to be correct.

## Quick changelog

The Inlier metric is calculated based on a Weibull function fitted to the distribution of pairwise distances to previous time points. In the current implementation, the distances are not normalised which can sometimes lead to the Weibull fit being incorrect. This PR adds normalisation to the distances before the fitting.

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -780,6 +780,9 @@ class FreqaiDataKitchen:
         into previous timepoints.
         """
 
+        def normalise(dataframe: DataFrame) -> DataFrame:
+            return (dataframe - dataframe.min()) / (dataframe.max() - dataframe.min())
+
         no_prev_pts = self.freqai_config["feature_parameters"]["inlier_metric_window"]
 
         if set_ == 'train':
@@ -824,6 +827,7 @@ class FreqaiDataKitchen:
         inliers = pd.DataFrame(index=distances.index)
         for key in distances.keys():
             current_distances = distances[key].dropna()
+            current_distances = normalise(current_distances)
             fit_params = stats.weibull_min.fit(current_distances)
             quantiles = stats.weibull_min.cdf(current_distances, *fit_params)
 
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/data_kitchen.py
```diff
@@ -780,6 +780,9 @@ def compute_inlier_metric(self, set_='train') -> None:
         into previous timepoints.
         """
 
+        def normalise(dataframe: DataFrame) -> DataFrame:
+            return (dataframe - dataframe.min()) / (dataframe.max() - dataframe.min())
```
xmatthias: If this function were used in a regular `populate_indicators()` method - it'd be a lookahead bias because it's using the full dataframe ...

Assuming you considered this - please add a docstring here - which should clearly state why/when it's acceptable to use this - as it got a lookahead bias if it's not used with great care.

robcaulk:
I think we need to store the `fit_params` when `set_ == 'train'` and then call those `fit_params` when `set_ == 'prediction'`. This means that we are checking where the prediction feature distances fit into the trained feature distance distributions. 

We can store the fit params in `self.data['fit_params'] = fit_params` in this function when `set_ == 'train'`. Then when just call them when `set_ == 'prediction'` with `fit_params = self.data['fit_params']`. FreqAI is handling everything properly on the back end to ensure that data dictionary is the correct one :-)

freqtrade/freqai/data_kitchen.py
th0rntwig: The inlier metric function runs on the sliced train dataset so there is no issue.

th0rntwig:
> I think we need to store the `fit_params` when `set_ == 'train'` and then call those `fit_params` when `set_ == 'prediction'`. This means that we are checking where the prediction feature distances fit into the trained feature distance distributions.
> 
> We can store the fit params in `self.data['fit_params'] = fit_params` in this function when `set_ == 'train'`. Then when just call them when `set_ == 'prediction'` with `fit_params = self.data['fit_params']`. FreqAI is handling everything properly on the back end to ensure that data dictionary is the correct one :-)

Done!

robcaulk (approved):
LGTM!

PR #7434, opened 2022-09-18
improve train queue system in FreqAI

robcaulk:
The old FreqAI train queue was in-capable of continuing training after reloading from a changed pairlist (while maintaining `identifier` constant). 

This lead to issues where the bot would stop training, and it would simply wait indefinitely without training a pair. 

Current PR switches to a `deque` for the queue and builds queues from existing information to the best of its ability, or if no existing info is available, it builds the queue based on the whitelist.

### DIFF
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -3,6 +3,7 @@ import shutil
 import threading
 import time
 from abc import ABC, abstractmethod
+from collections import deque
 from datetime import datetime, timezone
 from pathlib import Path
 from threading import Lock
@@ -80,6 +81,7 @@ class IFreqaiModel(ABC):
         self.pair_it = 0
         self.pair_it_train = 0
         self.total_pairs = len(self.config.get("exchange", {}).get("pair_whitelist"))
+        self.train_queue = self._set_train_queue()
         self.last_trade_database_summary: DataFrame = {}
         self.current_trade_database_summary: DataFrame = {}
         self.analysis_lock = Lock()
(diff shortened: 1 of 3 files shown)


freqtrade/freqai/freqai_interface.py
```diff
@@ -181,29 +183,36 @@ def _start_scanning(self, strategy: IStrategy) -> None:
...
-                    )
-                    self.train_timer('stop')
-
-            self.dd.save_historic_predictions_to_disk()
+            pair = self.train_queue[0]
+
+            # ensure pair is avaialble in dp
+            if pair not in strategy.dp.current_whitelist():
+                self.train_queue.popleft()
+                logger.warning(f'{pair} not in current whitelist, removing from train queue.')
```
xmatthias, line 191: I guess we should also be able to do this in reverse? (maybe not on every iteration)
so basically - we check for each pair in the current pairlist, if it's in the train-queue - and if it is not - we append it at the end?

Now it's an idea for the future for sure - as that'll also have implications on the data - which might be missing for that new pair.

freqtrade/freqai/freqai_interface.py
```diff
@@ -685,6 +691,30 @@ def get_init_model(self, pair: str) -> Any:
...
+        best_queue = deque()
+
+        pair_dict_sorted = sorted(self.dd.pair_dict.items(),
+                                  key=lambda k: k[1]['trained_timestamp'])
+        for pair in pair_dict_sorted:
+            if pair[0] in current_pairlist:
+                best_queue.appendleft(pair[0])
+        for pair in current_pairlist:
+            if pair not in best_queue:
+                best_queue.appendleft(pair)
```
robcaulk, line 713: This should be `appendright()`

freqtrade/freqai/freqai_interface.py
robcaulk: Sure it just needs to also execute the refresh ohlcv. There is your volume pairlist support if you want it!

xmatthias: absolutely, that's my point 
needs to be looked at separately though.

xmatthias (approved):
LGTM 👍

PR #7441, opened 2022-09-19
Ensure train ordering after restart

robcaulk:
Ensure lowest timestamps get trained first after restart.

The dict is sorted from lowest timestamp to highest timestamp - this means that we want to appendright() so that lower timestamps get trained before higher timestamps.

### DIFF
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -702,8 +702,8 @@ class IFreqaiModel(ABC):
         """
         current_pairlist = self.config.get("exchange", {}).get("pair_whitelist")
         if not self.dd.pair_dict:
-            logger.info('Set fresh train queue from whitelist.')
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/freqai_interface.py
```diff
@@ -703,6 +703,7 @@ def _set_train_queue(self):
         current_pairlist = self.config.get("exchange", {}).get("pair_whitelist")
         if not self.dd.pair_dict:
             logger.info('Set fresh train queue from whitelist.')
+            logger.info(f'Queue: {current_pairlist}')
```
xmatthias: Can we combine this message with the already existing message, please? (same below)

FreqAI already creates too many logmessages - so we should aim to combine them where possible (having 2 messages here will provide 0 benefit).

robcaulk:
Good to go - merge when you are ready.

xmatthias (approved):
LGTM :+1:

PR #7442, opened 2022-09-19
Allow FreqAI to send back individual columns in backtesting (`spice_rack` need)

wagnercosta:
## Summary

Try to let FT trim the backtesting data instead of FreqAI by modifying startup_candle in backtesting module.

FreqAI is designed to replace the entire dataframe in backtesting.

spice_rack simply adds a column to an existing dataframe.

The dataframes are not equal in length because FreqAI is trimming off training data candles at the end of the backtesting calculations so that backtesting statistics would be correct.

Now we need the dataframe in FreqAI to match the length of the dataframe in the backtest, which includes the training data candles. So the trimming needs to happen later, to maintain backtesting statistics accuracy.

Please read the [conversation on discord](https://discord.com/channels/700048804539400213/983285489299779645/1021150889811120158) for background.

Please also review [this PR](https://github.com/freqtrade/freqtrade/pull/7391) to understand why the training data was interfering with backtesting statistics, and needed to be removed.

### DIFF
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -90,8 +90,12 @@ class DataProvider:
         if saved_pair not in self.__cached_pairs_backtesting:
             timerange = TimeRange.parse_timerange(None if self._config.get(
                 'timerange') is None else str(self._config.get('timerange')))
+
+            # In backtesting, the training data was already add
+            add_train_candles = self.runmode in (RunMode.DRY_RUN, RunMode.LIVE)
+            startup_candles = self.get_required_startup(str(timeframe), add_train_candles)
+
             # Move informative start time respecting startup_candle_count
-            startup_candles = self.get_required_startup(str(timeframe))
             tf_seconds = timeframe_to_seconds(str(timeframe))
             timerange.subtract_start(tf_seconds * startup_candles)
             self.__cached_pairs_backtesting[saved_pair] = load_pair_history(
@@ -105,7 +109,7 @@ class DataProvider:
             )
         return self.__cached_pairs_backtesting[saved_pair].copy()
 
-    def get_required_startup(self, timeframe: str) -> int:
+    def get_required_startup(self, timeframe: str, add_train_candles: bool = True) -> int:
         freqai_config = self._config.get('freqai', {})
         if not freqai_config.get('enabled', False):
             return self._config.get('startup_candle_count', 0)
@@ -115,7 +119,9 @@ class DataProvider:
             # make sure the startupcandles is at least the set maximum indicator periods
             self._config['startup_candle_count'] = max(startup_candles, max(indicator_periods))
             tf_seconds = timeframe_to_seconds(timeframe)
-            train_candles = freqai_config['train_period_days'] * 86400 / tf_seconds
+            train_candles = 0
+            if add_train_candles:
+                train_candles = freqai_config['train_period_days'] * 86400 / tf_seconds
             total_candles = int(self._config['startup_candle_count'] + train_candles)
             logger.info(f'Increasing startup_candle_count for freqai to {total_candles}')
             return total_candles
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -143,10 +143,14 @@ class Backtesting:
 
         # Get maximum required startup period
         self.required_startup = max([strat.startup_candle_count for strat in self.strategylist])
-        # Add maximum startup candle count to configuration for informative pairs support
-        self.config['startup_candle_count'] = self.required_startup
         self.exchange.validate_required_startup_candles(self.required_startup, self.timeframe)
 
+        if self.config.get('freqai', {}).get('enabled', False):
+            # For FreqAI, increase the required_startup to includes the training data
+            self.required_startup = self.dataprovider.get_required_startup(self.timeframe)
+
+        # Add maximum startup candle count to configuration for informative pairs support
+        self.config['startup_candle_count'] = self.required_startup
         self.trading_mode: TradingMode = config.get('trading_mode', TradingMode.SPOT)
         # strategies which define "can_short=True" will fail to load in Spot mode.
         self._can_short = self.trading_mode != TradingMode.SPOT
@@ -221,7 +225,7 @@ class Backtesting:
             pairs=self.pairlists.whitelist,
             timeframe=self.timeframe,
             timerange=self.timerange,
-            startup_candles=self.dataprovider.get_required_startup(self.timeframe),
+            startup_candles=self.config['startup_candle_count'],
             fail_without_data=True,
             data_format=self.config.get('dataformat_ohlcv', 'json'),
             candle_type=self.config.get('candle_type_def', CandleType.SPOT)


freqtrade/data/dataprovider.py
```diff
@@ -90,8 +90,12 @@ def historic_ohlcv(
         if saved_pair not in self.__cached_pairs_backtesting:
             timerange = TimeRange.parse_timerange(None if self._config.get(
                 'timerange') is None else str(self._config.get('timerange')))
+
+            # In backtesting, the training data was already add
+            add_train_candles = self.runmode in (RunMode.DRY_RUN, RunMode.LIVE)
```
xmatthias, line 95: that's not a realistic check (it'll always be false).

`historic_ohlcv()` is only called in one place (from `get_pair_dataframe()`), which does not call this method if we're in dry/live mode.

xmatthias (commented):
The first thing i ask my self - is the target branch of add-spice-rack really correct?

if there's a problem in the startup calculation - it should be fixed in develop (so the branch should be based on top of develop - and be pointed at develop) - as i don't see how spice_rack would change behavior in this.

If i'm honest - i will need to play with this - however the backtest that works on the develop branch is failing with an error (that's clearly unrelated to this PR, but is most likely added/caused in spice_rack).

`AttributeError: 'FreqaiExampleStrategy' object has no attribute 'freqai_info'`

That's an error we need to fix in spice_rack (it's a PR that's marked as WIP - so not necessarily expected to work at all times) - but it also clearly makes testing of this pretty difficult - as we first need to move the commit(s) to another, actually working branch.

---

Cherry-picking this commit on top of develop fixes backtesting again - however i'm not entirely happy with the result.

I'm using the sample strategy with sample config (only change is to n_estimators - so training is quick).

```
freqtrade backtesting --strategy FreqaiExampleStrategy --strategy-path freqtrade/templates --config config_examples/config_freqai.example.json --freqaimodel LightGBMRegressor --timerange 20220911-20220912
```

My data starts on august 31st in this case - so if i try this on develop, backtesting this works.

if i run it with this commit on top of develop - i get "no data left after adjusting".
The adjustment candles will be 7240 3m candles - which i don't have (that's ~15 days, i only have 12 days prior to my backtest).

This cannot be a problem - as it's a realistic case, for example when pairs are newly listed and no older data is available.

---

In summary - i'm not entirely sure what this PR is actually aiming to fix - as (at least at a first glance) - behavior in develop seems correct, as the actual start date is at the defined timerange start.

robcaulk:
> The first thing i ask my self - is the target branch of add-spice-rack really correct?
> 
> if there's a problem in the startup calculation - it should be fixed in develop (so the branch should be based on top of develop - and be pointed at develop) - as i don't see how spice_rack would change behavior in this.
> 
> If i'm honest - i will need to play with this - however the backtest that works on the develop branch is failing with an error (that's clearly unrelated to this PR, but is most likely added/caused in spice_rack).
> 
> `AttributeError: 'FreqaiExampleStrategy' object has no attribute 'freqai_info'`
> 
> That's an error we need to fix in spice_rack (it's a PR that's marked as WIP - so not necessarily expected to work at all times) - but it also clearly makes testing of this pretty difficult - as we first need to move the commit(s) to another, actually working branch.
> 
> Cherry-picking this commit on top of develop fixes backtesting again - however i'm not entirely happy with the result.
> 
> I'm using the sample strategy with sample config (only change is to n_estimators - so training is quick).
> 
> ```
> freqtrade backtesting --strategy FreqaiExampleStrategy --strategy-path freqtrade/templates --config config_examples/config_freqai.example.json --freqaimodel LightGBMRegressor --timerange 20220911-20220912
> ```
> 
> My data starts on august 31st in this case - so if i try this on develop, backtesting this works.
> 
> if i run it with this commit on top of develop - i get "no data left after adjusting". The adjustment candles will be 7240 3m candles - which i don't have (that's ~15 days, i only have 12 days prior to my backtest).
> 
> This cannot be a problem - as it's a realistic case, for example when pairs are newly listed and no older data is available.
> 
> In summary - i'm not entirely sure what this PR is actually aiming to fix - as (at least at a first glance) - behavior in develop seems correct, as the actual start date is at the defined timerange start.

It seems you do not have the background on the bug:

https://discord.com/channels/700048804539400213/983285489299779645/1021150889811120158

please review that conversation. 

The reason this is targeted for `spice_rack` branch is because it is unnecessary without `spice_rack`. Otherwise, this was already solved by `FreqaiDataKitchen.remove_training_from_backtesting()`. I.e. there is no reason to push this to develop *before* `spice_rack` goes to develop - it doesn't fix anything in develop.

xmatthias:
Considering that spice-rack is just an automatic/pre-configuration of freqAI for users who want to have this done "automatically" - i don't really see why spice_rack would need a special fix. That suggests a bug in spice-rack - and the way this works currently seems to break / change behavior of non-spice_rack configs in unintended ways.

robcaulk:
> Considering that spice-rack is just an automatic pre-configuration of freqAI - i don't really see why spice_rack would need a special fix. That suggests a bug in spice-rack - and the way this works currently seems to break / change behavior of non-spice_rack configs in unintended ways.

It is pretty simple - FreqAI is designed to replace the entire dataframe in backtesting.

`spice_rack` simply adds a column to an existing dataframe.

The dataframes are not equal in length because FreqAI is trimming off training data candles at the end of the backtesting calculations so that backtesting statistics would be correct.

Now we need the dataframe in FreqAI to match the length of the dataframe in the backtest, which includes the training data candles. So the trimming needs to happen later, to maintain backtesting statistics accuracy. 

Please read the conversation on discord for background. 

Please also review [this PR](https://github.com/freqtrade/freqtrade/pull/7391) to understand why the training data was interfering with backtesting statistics, and needed to be removed.

robcaulk:
> That suggests a bug in spice-rack 

Ya... that is what we are fixing with this PR :-) I am so confused.

> and the way this works currently seems to break / change behavior of non-spice_rack configs in unintended ways.

We agree that PRs don't get merged if they break other stuff - especially the CI... which is currently broken on this PR for good reason. so I am confused as to what the issue is here.

We are using this PR to find the best solution. Please feel free to help us find that solution.

xmatthias:
> We are using this PR to find the best solution. Please feel free to help us find that solution.

It's difficult to even attempt to find a solution if 
* the code to test is non-functional 
* the problem you're trying to fix is completely unclear - and must be pieced together from a discord conversation.

robcaulk:
> > We are using this PR to find the best solution. Please feel free to help us find that solution.
> 
> It's difficult to even attempt to find a solution if
> 
> * the code to test is non-functional
> * the problem you're trying to fix is completely unclear - and must be pieced together from a discord conversation.

I admit the original description/title was lacking information. But it was updated and in my opinion the PR and problem is now well described/titled. 

If you don’t wish to dive in, that is perfectly ok - i just ask you to please allow us solve the problem and get a passing CI before critiquing the very existence of the PR.

xmatthias:
i'm (still) not convinced this properly addresses the problem you see with `spice_rack`.
While it's messing with the backtesting startup candles - it's not directly addressing the actual problem.
what we need is to ensure (at all times) that the returning values match (on both length and dates) the dataframe the strategy uses / passed in.

for backtesting (for example) - it could be the case that users call this from a callback (for whatever reason) - so you're no longer getting the full dataframe, but a dataframe truncated up to this point in time.
While an edge-case (one we can probably document to "Not do" as it has other negative implications, too) - i think it potentially shows a slightly different problem.

we're getting a dataframe in the call to `spice_rack()` - the most sensible approach would probably be to ensure right there that the returning series has the same length - and more importantly, is for the same dates.

---

We imagine / advertise spice_rack to be "just another indicator you can use" - so there's nothing really preventing the following (which i think is a valid usecase, too).

``` python
        dataframe['dissimilarity_index'] = self.freqai.spice_rack(
            'DI_values', dataframe, metadata, self)
        informative_1h['dissimilarity_index'] = self.freqai.spice_rack(
            'DI_values', informative_1h, metadata, self)
```

user would like to have the dissimilarity index - but calculated against the informative 1h pair.

To quote from the spice_rack docs:

> The `spice_rack` is aimed at users who do not wish to deal with setting up `FreqAI` confgs, but instead prefer to interact with `FreqAI` similar to a `talib` indicator.

we seem to place spice_rack as an indicator - just like ta-lib.
ta-lib can do this ... can spice_rack too (or should it)?
if not - i think we need to change what we're trying to bring here - as it's got quite some limitations atm (only for the main dataframe, ...)

---

To be very clear - this is not just about this pr per se - but about the problem it's aiming to fix with `spice_rack` (based on the way i understood it - and what i've seen further after thinking about it).

robcaulk:
Miscommunications left and right. Closing this in favor of a new push on develop, as requested by @xmatthias.

PR #7443, opened 2022-09-19
protect against unforeseen issues in scanning thread

robcaulk:
The training thread is vulnerable to unforeseen and rare issues. Such as 

```
catboost/libs/metrics/metric.cpp:5119: All train targets are equal
```
Here we log it as an error and allow the thread to keep going, avoiding a full breakdown of the training thread just because a single pair had an issue.

### DIFF
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -205,9 +205,13 @@ class IFreqaiModel(ABC):
 
             if retrain:
                 self.train_timer('start')
-                self.extract_data_and_train_model(
-                    new_trained_timerange, pair, strategy, dk, data_load_timerange
-                )
+                try:
+                    self.extract_data_and_train_model(
+                        new_trained_timerange, pair, strategy, dk, data_load_timerange
+                    )
+                except Exception as msg:
+                    logger.warning(f'Training {pair} raised exception {msg}, skipping.')
+
                 self.train_timer('stop')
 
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/freqai_interface.py
```diff
@@ -205,9 +205,13 @@ def _start_scanning(self, strategy: IStrategy) -> None:
...
                 self.train_timer('start')
-                self.extract_data_and_train_model(
-                    new_trained_timerange, pair, strategy, dk, data_load_timerange
-                )
+                try:
+                    self.extract_data_and_train_model(
+                        new_trained_timerange, pair, strategy, dk, data_load_timerange
+                    )
+                except Exception as msg:
+                    logger.warning(f'Training {pair} raised exception {msg}, skipping.')
```
xmatthias, line 213: i think you need to be careful with logging.
this way, you'll only get the exception type (for example ValueError) - but not the actual explanation.

not necessarily a problem - but can hinder actual debugging / error search.

xmatthias (approved):
LGTM :+1:

freqtrade/freqai/freqai_interface.py
robcaulk: This exception is hard to predict - could arise from any library the user brings in. I guess we could try to wrap this similar to the strategy wrapper so we can get slightly more detailed exceptions?

PR #7454, opened 2022-09-22
Ensure the DF has the same size in backtesting FreqAI

wagnercosta:
## Summary

Currently, in develop branch, we are removing training data from DF before sending to backtest module (function `remove_training_from_backtesting`  in `freqai/data_kitchen.py`).  This results in different dataframe sizes after backtesting is completed, which can cause other issues.

This PR is intended to remove this function, and still ensure that backtesting does not give wrong stats by not correctly computing the backtest start date when considering the dataframe's training data.

The new code calculates the amount of candles for the training period and adds these candles to startup_candle_count. This will ensure that the start date shown in the backtesting result is correct.


robcaulk (approved):
This LGTM

xmatthias (approved):
Removed a leftover comment which was no longer relevant

LGTM :+1: Great job! :100:

PR #7455, opened 2022-09-22
Allow use of --strategy-list with freqai, with warning

paranoidandy:
## Summary

Trying to run backtesting with freqai and multiple strategies via `--strategy-list` currently exits with an `OperationalException`. This PR changes this to allow backtesting multiple strategies in one command, but with a user warning that all strategies must share identical `populate_any_indicators`.

## Quick changelog

- Replace raising `OperationalException` with a user warning.
- Rewrite test.freqai.test_freqai_backtesting.test_freqai_backtest_start_backtest_list to assert the warning is produced.

### DIFF
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -430,6 +430,16 @@ class FreqaiDataDrawer:
 
         return
 
+    def load_metadata(self, dk: FreqaiDataKitchen) -> None:
+        """
+        Load only metadata into datakitchen to increase performance during
+        presaved backtesting (prediction file loading).
+        """
+        with open(dk.data_path / f"{dk.model_filename}_metadata.json", "r") as fp:
+            dk.data = json.load(fp)
+            dk.training_features_list = dk.data["training_features_list"]
+            dk.label_list = dk.data["label_list"]
+
     def load_data(self, coin: str, dk: FreqaiDataKitchen) -> Any:
         """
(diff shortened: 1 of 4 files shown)


xmatthias:
@robcaulk wouldn't a side-effect of this be that `freqai.start()` is called twice in backtesting? (once per strategy).

so in unlucky cases, this might mean that the model is trained twice ... no? (this was the main reason to disable this in the first place).

robcaulk:
> @robcaulk wouldn't a side-effect of this be that `freqai.start()` is called twice in backtesting? (once per strategy).
> 
> so in unlucky cases, this might mean that the model is trained twice ... no? (this was the main reason to disable this in the first place).

With the new prediction saving feature, all subsequent calls to `freqai.start()` will just load the predictions from disk. This is why we decided to try to bring this feature back. This is also why it warns that `populate_any_indicators is **REQUIRED** to be the same across all strategies`

Now I am thinking about ways to enforce it beyond just giving the user a warning.

xmatthias (approved):
For me that's fine - obviously having a way to enforce it would be better - but i guess not required right away.

robcaulk (approved):
Good to go

robcaulk:
> For me that's fine - obviously having a way to enforce it would be better - but i guess not required right away.


We now enforce identical `populate_any_indicators` by loading in the metadata from the existing training and comparing feature lists to current dataframe.

PR #7457, opened 2022-09-23
Add elapsed time to Freqai training logs

aemr3:
## Summary

It's difficult to see elapsed time while training. This helps to understand training performance.

## What's new?

Added elapsed time to the training log.

### DIFF
--- a/freqtrade/freqai/base_models/BaseClassifierModel.py
+++ b/freqtrade/freqai/base_models/BaseClassifierModel.py
@@ -1,4 +1,5 @@
 import logging
+from time import time
 from typing import Any, Tuple
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


xmatthias (commented):
Looking at the necessary change for this - would it not make sense to move this logging to the parent class (basically to `freqai_interface.py`) - around the call to `.train()` ... 
that'd make logging align throughout different models - which is quite difficult to actually ensure this way.

aemr3:
> Looking at the necessary change for this - would it not make sense to move this logging to the parent class (basically to `freqai_interface.py`) - around the call to `.train()` ...
> that'd make logging align throughout different models - which is quite difficult to actually ensure this way.

I didn't want to do that, as we already have logging inside these classes. Most people will not use custom made models (including me). So I am not sure really why these logs should go into interface right now. Just wanted to commit a simple change.

xmatthias (approved):
> I didn't want to do that, as we already have logging inside these classes. Most people will not use custom made models (including me). So I am not sure really why these logs should go into interface right now. Just wanted to commit a simple change.

It's random observations we sometimes see during review.
if you don't look at this diff in the context of this change - it's not immediately clear that the same logging happens in every class  - but with this it becomes very apparent.

you'll notice that i didn't "request changes" - with the intend to leave it up to you to decide if you're open to do this (slightly bigger) improvement or not.


LGTM :+1:

PR #7459, opened 2022-09-23
Fix typo in Configuration docs for Producer

wizrds:
## What's new?

There is a typo in the `configuration.md` docs page that still uses the old terminology `External Signals`.


xmatthias (approved):
Nice catch - thanks! :+1:

PR #7460, opened 2022-09-23
Bug fix in External Message Consumer

wizrds:
## Summary

Catch `websockets.exceptions.ConnectionClosedError` which is caused in rare occasions when a Producer is shut down


xmatthias (approved):
LGTM :+1:

PR #7474, opened 2022-09-25
Fix bug in API Server WebSocket

wizrds:
## Summary

This fixes the error when a client would disconnect while the background thread in the API Server would be broadcasting data.


xmatthias (approved):
nice catch, LGTM 👍

PR #7475, opened 2022-09-26
FreqAI - Backtest live/ready models

wagnercosta:
## Summary

Add a new feature for backtest live (ready) models - FreqAI

## What's new?

This feature will allow, through the `--freqai-backtest-live-models` parameter, to run the backtest with ready-made models (which may have been generated in other backtests or in dry/live run). The code will calculate the timerange according to the saved models. Each model will perform the backtesting according to the training end timestamp (start date) until the training end timestamp of the next model (end date) and so on. For the last model, the period of the previous model will be used for the execution.

## TODO
- [X] Code draft version to allow early feedback
- [x] Run additional tests with saved models
- [x] Test pairs with different amount of models in the directory
- [x] Add documentation/code comments (new functions)
- [x] Add documentation
- [x] Add dedicated tests


freqtrade/freqai/data_kitchen.py
```diff
@@ -1095,15 +1119,15 @@ def check_if_new_training_required(
 
         return retrain, trained_timerange, data_load_timerange
 
-    def set_new_model_names(self, pair: str, trained_timerange: TimeRange):
+    def set_new_model_names(self, pair: str, timestamp_id: int):
```
wagnercosta, line 1145: @robcaulk , I changed this param because for training models, we are using the training end timestamp and for the new backtest_live_models features we need to use the start backtest timestamp. Using timestamp_id, we will handle this in a generic way.

freqtrade/freqai/freqai_util.py
```diff
@@ -0,0 +1,115 @@
+"""
```
xmatthias: What's the difference between `freqtrade.freqai.utils` and `freqtrade.freqai.freqai_utils` ? (both files are utils and contain some utilities ...)

freqtrade/commands/cli_options.py
```diff
@@ -668,4 +668,10 @@ def __init__(self, *args, **kwargs):
         help='Specify additional lookup path for freqaimodels.',
         metavar='PATH',
     ),
+    "freqai_backtest_live_models": Arg(
+        '--freqai-backtest-live-models',
+        help='Run backtest with ready models.',
+        action='store_true',
+        default=False,
```
xmatthias: Be careful with setting defaults here.

if you do this means you will not have the possibility to set this parameter via configuration.
A better approach to define defaults is in `constants.py` via validation (if necessary) - or by assuming it's non-mandatory and checking with `config.get(paramname, default)`.

freqtrade/configuration/config_validation.py
```diff
@@ -334,6 +335,26 @@ def _validate_freqai_hyperopt(conf: Dict[str, Any]) -> None:
...
+                '--freqai-backtest-live-models parameter.')
+
+        if freqai_backtest_live_models and not freqai_enabled:
+            raise OperationalException(
+                'Using --freqai-backtest-live-models parameter is only '
+                'supported with a FreqAI strategy.')
+
+        if freqai_enabled and not freqai_backtest_live_models and not timerange:
+            raise OperationalException(
+                'Please pass --timerange if you intend to use FreqAI for backtesting.')
```
xmatthias, line 376: huh now that's some extensive logic as to when we can and cannot do something.

I'm not sure users will understand this...

freqtrade/freqai/freqai_interface.py
```diff
@@ -572,7 +603,7 @@ def extract_data_and_train_model(
         model = self.train(unfiltered_dataframe, pair, dk)
 
         self.dd.pair_dict[pair]["trained_timestamp"] = new_trained_timerange.stopts
-        dk.set_new_model_names(pair, new_trained_timerange)
+        dk.set_new_model_names(pair, int(new_trained_timerange.stopts))
```
xmatthias: There's no need to force-int this ...

freqtrade/freqai/data_kitchen.py
```diff
@@ -450,6 +454,31 @@ def split_timerange(
         # print(tr_training_list, tr_backtesting_list)
         return tr_training_list_timerange, tr_backtesting_list_timerange
 
+    def split_timerange_live_models(
+        self
+    ) -> Tuple[list, list]:
+
+        tr_backtesting_list_timerange = []
+        pair = self.pair.split("/")[0].split(":")[0]
```
xmatthias: i'm not sure what you're trying to do here - but it's for sure wrong.

a pair is either `BTC/USDT` - or `BTC/USDT:USDT`.

the first expression will (in both instances) give you `BTC` - so the 2nd is irrelevant.
It's however also not a pair, but just the base asset ... so i'm not if this is really giving the desired outcome (as well as if it's even necessary).
if the expectation is BTC - then the 2nd split is unnecessary.

The variable name is also missleading (maybe `asset` would be better) - as this is no longer a pair.

freqtrade/freqai/data_kitchen.py
```diff
@@ -450,6 +454,31 @@ def split_timerange(
...
+        if pair not in self.backtest_live_models_data["pairs_end_dates"]:
+            raise OperationalException(
+                f"Model not available for pair {self.pair}. "
+                "Please, try again after removing this pair from the configuration file."
+            )
+        pair_data = self.backtest_live_models_data["pairs_end_dates"][pair]
+        model_end_dates = []
+        backtesting_timerange = self.backtest_live_models_data["backtesting_timerange"]
+        for end_date in pair_data:
+            model_end_dates.append(end_date)
```
xmatthias: List-comprehension would yield better performance ... ;) 

you'll still need the final append - but still ...

freqtrade/freqai/freqai_interface.py
```diff
@@ -280,6 +305,12 @@ def start_backtesting(
                 dk.append_predictions(append_df)
             else:
                 if not self.model_exists(dk):
+                    if dk.backtest_live_models:
+                        raise OperationalException(
+                            "Training models is not allowed "
+                            "in backtest_live_models backtesting "
+                            "mode"
+                        )
```
xmatthias: This will surely fit on fewer lines ...

that said - ain't we a bit deep into the process here to still operationalError? 
wouldn't half of the magic have already happened if we discover that the last pair of the pairlist doesn't have models available?

freqtrade/freqai/freqai_interface.py
```diff
@@ -253,25 +257,46 @@ def start_backtesting(
...
+                tr_backtest_startts_str = datetime.fromtimestamp(
+                                                tr_backtest.startts,
+                                                tz=timezone.utc).strftime(DATETIME_PRINT_FORMAT)
+                tr_backtest_stopts_str = datetime.fromtimestamp(
+                                                tr_backtest.stopts,
+                                                tz=timezone.utc).strftime(DATETIME_PRINT_FORMAT)
+                logger.info(
+                    f"No data found for pair {pair} "
+                    f" from {tr_backtest_startts_str} "
+                    f"to {tr_backtest_stopts_str}. "
```
xmatthias: in general i'd try to use as few lines as possible (which means - avoid unnecessary line wrapping ...)

freqtrade/optimize/backtesting.py
```diff
@@ -25,6 +25,7 @@
 from freqtrade.exceptions import DependencyException, OperationalException
 from freqtrade.exchange import (amount_to_contract_precision, price_to_precision,
                                 timeframe_to_minutes, timeframe_to_seconds)
+from freqtrade.freqai import freqai_util
```
xmatthias: I'm a bit concerned about this import.

the assumption is that freqAI is an optional module (just as hyperopt) - and all it's dependencies are optional as well.
While i think it works now - importing this here creates a direct dependency on the `freqtrade.freqai` package to always resolve / load (even with freqAI dependencies missing).

one small change (which will not be seen in the context of this import), adding an import to `freqai/__init__.py` can break this - which is something that'll not be captured by tests (we require all dependencies for tests).

If we need this import, it should be moved to a "freqAI vettet" area - which we know is only running when freqAI is installed / used.
maybe look at the imports in `interface.py` for ideas (basic idea is to move the import to the function that uses the stuff - which is behind the "paywall" of "freqai.enabled=true").

freqtrade/freqai/freqai_util.py
```diff
@@ -0,0 +1,115 @@
...
+    if (
+        finish_timestamp < int(datetime.now(tz=timezone.utc).timestamp()) and
+        datetime.now(tz=timezone.utc).strftime('%Y%m%d') != stop.strftime('%Y%m%d')
+    ):
+        # add 1 day to string timerange to ensure BT module will load all dataframe data
+        end_date_string_timerange = stop + timedelta(days=1)
+
+    backtesting_string_timerange = (
+        f"{start.strftime('%Y%m%d')}-{end_date_string_timerange.strftime('%Y%m%d')}"
+    )
```
xmatthias: I'm not sure on the usefulness of this.

While it's currently used for backtesting - that part could easily be done by using the "TimeRange" object from below in combination with an else.(config.timerange should not be used elsewhere in backtesting).

It's also assigned to `backtest_live_models_data` - but this part / entry is never used - so i think we can save the 2nd value (as well as the conversion to that).

freqtrade/freqai/freqai_util.py
```diff
@@ -0,0 +1,115 @@
...
+    return backtesting_timerange, backtesting_string_timerange, pairs_end_dates
+
+
+def get_pairs_timestamps_training_from_ready_models(models_path: Path) -> Dict[str, Any]:
+    """
+    Scan the models path and returns all pairs end training dates (timestamp)
+    :param models_path: FreqAI model path
+
+    :returns:
+    :pairs_end_dates: Dict with pair and model end training dates info
```
xmatthias: it's interesting that people seem to think :returns: is the proper syntax (i've noticed this elsewhere, too)- it's actually :return: <value>.
Freqtrade "mostly" follows this (it's not always noticed in reviews).

Also, the actual variable name you're returning is missleading info - that's not visible to the outside / caller, and if that were renamed, it'll become an outdated fragment very soon.

xmatthias (commented):
Just a quick glance over this, really  ... 
assuming it works from a freqAI perspective - this seems nice - although i think it'll kindof collide with the `purge_models` option (?) - which will leave only the last 2 models around. This might need tuning in the future to leave the last X days around ... 

Backtesting currently assumes a min. model-duration of 1d (iirc) - from the description, i read that this aims to follow live as close as possible - so we'd get a model every 1 hour? 
will model expiration (`expiration_hours`) be considered for backtest?

as you point out already, docs are missing (and will be strongly needed - considering all the limitations (looking at `config_validation`) ...

wagnercosta:
> Just a quick glance over this, really ... assuming it works from a freqAI perspective - this seems nice - although i think it'll kindof collide with the `purge_models` option (?) - which will leave only the last 2 models around. This might need tuning in the future to leave the last X days around ...
> 
> Backtesting currently assumes a min. model-duration of 1d (iirc) - from the description, i read that this aims to follow live as close as possible - so we'd get a model every 1 hour? will model expiration (`expiration_hours`) be considered for backtest?
> 
> as you point out already, docs are missing (and will be strongly needed - considering all the limitations (looking at `config_validation`) ...

I think we can handle `purge_old_models` in docs. For example, an user can set to retrain a model each 7 days, set `purge_old_models` to true and still use this feature.

Actually we don't have a logic to verify model expiration, but I think it's not necessary at this moment and also can be documented.

Actually this code doesn't have a logic to ensure at least 1 day of BT and I will add this logic to this PR.

freqtrade/freqai/freqai_util.py
wagnercosta: I created another file to call functions from FreqAI and from BT module and avoid duplicate code. If I use the `freqtrade.freqai.utils` file, we have a circular reference error when calling functions from `freqtrade.freqai.data_kitchen` because it's imported in `freqtrade.freqai.utils` too. Maybe a best solution is move this file to a generic module? What do you suggest? The goal is a file with FreqAI functions that can be called inside FreqAI module and other modules too.

freqtrade/configuration/config_validation.py
wagnercosta: The timerange check was inside `freqtrade.freqai.data_kitchen` file. I moved to this file, to keep all validations together. In summary.

- Timerange is required for a regular FreqAI backtest.
- Backtest live models will calculate timerange, and we don't need this param.
- Backtest live models only works with FreqAI

I agree that we need to document well the rules.

xmatthias:
> Actually this code doesn't have a logic to ensure at least 1 day of BT and I will add this logic to this PR.

don't get me wrong though - i'm just asking questions based on my configuration - which is retraining once per hour (so to align, i'd need a backtest that uses 24 models per day).

> I think we can handle purge_old_models in docs. 

oh for sure - my idea was to maybe in a future improvement to have an additional parameter together with `purge_old_models` which keeps 4 or 6 models at a time (while still cleaning up behind itself to not keep growing indefinitely).

freqtrade/freqai/data_kitchen.py
wagnercosta: You are right. I need only the asset. I will fix the code and var name.

freqtrade/freqai/freqai_interface.py
wagnercosta: You're right. It was just an additional check to guarantee that the model will not be trained, but we would only get to this situation due to a verification failed before. I will remove this check.

robcaulk:
> Just a quick glance over this, really ... assuming it works from a freqAI perspective - this seems nice - although i think it'll kindof collide with the `purge_models` option (?) - which will leave only the last 2 models around. This might need tuning in the future to leave the last X days around ...

Using this requires a conscious effort by the user to collect *many* models during live. Thus, it is not colliding, it just means a user wont find much joy in this feature unless the user knows "hey I wanna fill my drive up so I can later backtest this directly."

Yes we can change `purge_old_models` to be an integer instead of a boolean, with default set to 2. But, I don't see it in the scope of this PR (despite it being a simple change). 

> Backtesting currently assumes a min. model-duration of 1d (iirc) - from the description, i read that this aims to follow live as close as possible - so we'd get a model every 1 hour? will model expiration (`expiration_hours`) be considered for backtest?

I am unsure where you read this (maybe there is an antiquated docstring somewhere that needs updating?). It *was* the case long ago. But I since improved backtesting (a couple months back) so that people can backtest at what ever subdaily frequency they desire.

freqtrade/freqai/freqai_util.py
wagnercosta: I refactored the code, but `config.timerange` is used in freqai. I think that for now, we need this config. Example  below:

https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/freqai/data_kitchen.py#L88

xmatthias: mmh i see - nice catch :+1:

freqtrade/configuration/config_validation.py
xmatthias: i'm not questioning the logic / need (even if it wouldn't have been in data_kitchen - doing sanity checks is still fine.
it's just a lot of "this combination, not that combination" :laughing: and without docs - it takes a moment to grasp :laughing:

docs/freqai-running.md
```diff
@@ -73,12 +73,24 @@ Backtesting mode requires [downloading the necessary data](#downloading-data-to-
 
 To allow for tweaking your strategy (**not** the features!), FreqAI will automatically save the predictions during backtesting so that they can be reused for future backtests and live runs using the same `identifier` model. This provides a performance enhancement geared towards enabling **high-level hyperopting** of entry/exit criteria.
 
-An additional directory called `predictions`, which contains all the predictions stored in `hdf` format, will be created in the `unique-id` folder.
```
wagnercosta, line 76: @robcaulk I think this was incorrect. Or your idea is to change de folder name? Currently, the code creates a folder called `backtesting_predictions`.

robcaulk: You are right, thanks for changing that.

freqtrade/freqai/freqai_util.py
robcaulk: I see the issue. Can we move the data_kitchen specific functions directly into `data_kitchen.py`?

robcaulk (changes requested):
This is looking quite nice - thank you.

I see one big issue - which is actually a current issue with normal backtesting as well. If the user uses `fit_live_predictions()`  (adds `fit_live_predictions_candles` to his/her config) then the backtesting module will not simulate it - and any information that `fit_live_predictions` adds to the return dataframe in live runs, will not be available to the strategy in backtests.

As I say, this actually not a problem with the current PR - it is an existing issue with backtesting. But maybe we can try to wrap it into this PR. 

The solution may be for us to create a fake `self.dd.historical_predictions` dictionary the same way live does , and to feed it into the `fit_live_predictions()` function as we are iterating through our timeranges in the backtesting loop...

I am not 100% how to handle this - maybe we can try to set a meeting to brainstorm best ways to do that. It seems like an important improvement to backtesting + live backtesting functionalities.

freqtrade/freqai/freqai_util.py
wagnercosta: Move `plot_feature_importance` to `data_kitchen.py` right? and remove this import `from freqtrade.freqai.data_kitchen import FreqaiDataKitchen`? I will modify this to keep just one util file.

robcaulk: No, I meant to move anything that any function from your freqai_utils that is called by data_kitchen, should go directly into data kitchen. Then we avoid any circular imports because data kitchen won’t need to import the utils file. Then you can move all your other freqai_utils into the normal utils file. 

Let me know if it presents an issue.

wagnercosta:
> This is looking quite nice - thank you.
> 
> I see one big issue - which is actually a current issue with normal backtesting as well. If the user uses `fit_live_predictions()` (adds `fit_live_predictions_candles` to his/her config) then the backtesting module will not simulate it - and any information that `fit_live_predictions` adds to the return dataframe in live runs, will not be available to the strategy in backtests.
> 
> As I say, this actually not a problem with the current PR - it is an existing issue with backtesting. But maybe we can try to wrap it into this PR.
> 
> The solution may be for us to create a fake `self.dd.historical_predictions` dictionary the same way live does , and to feed it into the `fit_live_predictions()` function as we are iterating through our timeranges in the backtesting loop...
> 
> I am not 100% how to handle this - maybe we can try to set a meeting to brainstorm best ways to do that. It seems like an important improvement to backtesting + live backtesting functionalities.

yes, we can brainstorm about it.
Other option to discuss is to call a new method after this point:

https://github.com/freqtrade/freqtrade/blob/backtest_live_models/freqtrade/freqai/freqai_interface.py#L146

We have the full dataframe that will be returned to BT module. We can recalculate the gaussian distribution according with `fit_live_predictions_candles` param.

robcaulk:
> > This is looking quite nice - thank you.
> > I see one big issue - which is actually a current issue with normal backtesting as well. If the user uses `fit_live_predictions()` (adds `fit_live_predictions_candles` to his/her config) then the backtesting module will not simulate it - and any information that `fit_live_predictions` adds to the return dataframe in live runs, will not be available to the strategy in backtests.
> > As I say, this actually not a problem with the current PR - it is an existing issue with backtesting. But maybe we can try to wrap it into this PR.
> > The solution may be for us to create a fake `self.dd.historical_predictions` dictionary the same way live does , and to feed it into the `fit_live_predictions()` function as we are iterating through our timeranges in the backtesting loop...
> > I am not 100% how to handle this - maybe we can try to set a meeting to brainstorm best ways to do that. It seems like an important improvement to backtesting + live backtesting functionalities.
> 
> yes, we can brainstorm about it. Other option to discuss is to call a new method after this point:
> 
> https://github.com/freqtrade/freqtrade/blob/backtest_live_models/freqtrade/freqai/freqai_interface.py#L146
> 
> We have the full dataframe that will be returned to BT module. We can recalculate the gaussian distribution according with `fit_live_predictions_candles` param.

No doubt, but this means the user must write 2 functions, one for live and one for backtesting. It would be preferable for backtesting to simply reuse `fit_live_predictions()` - this way there would be no chance of a typo or mismatch in the calculations. 

I think this is beyond the scope of the present PR though - it would require machinery to handle `fit_live_predictions()` in backtesting. So for now I would leave it be.

freqtrade/freqai/freqai_util.py
wagnercosta: Done!

freqtrade/optimize/backtesting.py
robcaulk: This is resolved.

robcaulk (approved):
LGTM

freqtrade/freqai/data_kitchen.py
```diff
@@ -1350,6 +1374,109 @@ def check_if_backtest_prediction_is_valid(
...
+            all_models_end_dates.sort(reverse=True)
+            finish_timestamp = all_models_end_dates[0] + \
+                (all_models_end_dates[0] - all_models_end_dates[1])
+
+        all_models_end_dates.append(finish_timestamp)
+        all_models_end_dates.sort()
+        start_date = (datetime(*datetime.fromtimestamp(min(all_models_end_dates),
+                      timezone.utc).timetuple()[:3], tzinfo=timezone.utc))
+        end_date = (datetime(*datetime.fromtimestamp(max(all_models_end_dates),
+                    timezone.utc).timetuple()[:3], tzinfo=timezone.utc))
```
xmatthias, line 1440: am i correct in assuming that this will trim the model start/end date to days?
so if my live model was running from 2022-10-08 23:00:00 - 2022-10-09 01:00:00 (2 hours)- i'd get `2022-10-08 - 2022-10-10` (Now suddenly 2 days) - basically augmenting the actual backtest range quite a bit beyond the actual model availability ...?

is there a particular reason for this? (backtesting can handle intra-day start times through `--timerange <epoch>-<epoch>` - though it's no longer a documented behavior...)

wagnercosta: currently, Freqai needs a string timerange in config (see this code: https://github.com/freqtrade/freqtrade/blob/backtest_live_models/freqtrade/freqai/data_kitchen.py#L1035). The main purpose of this new code is to return this format to keeps the compatibility with existent FreqAI backtest behavior. 

But, this new feature will only make the predictions using the timestamp models periods. Even if the BT dataframe contains all candles, predictions and possibly trades will only happen on candles calculated based on the models timestamp. The other candles outside the period will not have the model predictions.
According with this rule in description of this PR:

> The code will calculate the timerange according to the saved models. Each model will perform the backtesting according to the training end timestamp (start date) until the training end timestamp of the next model (end date) and so on. For the last model, the period of the previous model will be used for the execution.

wagnercosta: @xmatthias Sorry. Forget my previous comment. I'm going to do some tests with the timerange with epochs.

xmatthias: Just asking ;) for now it's good as is :+1: 

let's keep the epoch version for a later enhancement - this PR was "lying around" for long enough.

xmatthias (approved):
LGTM :+1: 

Nice work :100:

PR #7481, opened 2022-09-26
:pencil2: fixed typo from StaticPairlist to StaticPairList

chusri:
## Summary

There is one typo in the error message, when the user copies the suggestion message to use in their configuration, it will throw an error.

## Quick changelog

- update error message StaticPairlist to StaticPairList


robcaulk:
You will also need to go change the tests to match the new string:

https://github.com/freqtrade/freqtrade/actions/runs/3124809239/jobs/5069408830#step:8:444

```python
                       match=r'VolumePairList not allowed for backtesting\..*StaticPairlist.*'):
```

at lines 437 and 469 tests/optimize/test_backtesting.py` fail since they look for the old string.

xmatthias (approved):
Great catch! :100: 

LGTM :+1:

PR #7492, opened 2022-09-28
Shutdown Subproc Env on signal

wizrds:
Support cleaning up and shutting down any sub processes started by stable baselines when a signal is received.


xmatthias (approved):
Looks like a nice and simple solution :+1: 

Don't worry about tests - they seem to be random failures (which i hope to have fixed in develop) - but anyway not related to the changes here.

robcaulk (approved):
Thanks a lot for this

PR #7495, opened 2022-09-28
Set train-test-split parameters shuffle=False as default and remove stratification

th0rntwig:
## Summary

Updating input parameter for the splitting of features into training and testing datasets:
* Ensure shuffle=False for splitting data into train and test.
* Remove stratification option.

## Quick changelog

Splitting input features into train and test with shuffle=True (default for sklearn train_test_split()) removes the chronological ordering of the data. This PR sets shuffle=False as the default input to train_test_split() to ensure that the user is aware of the data being shuffled by having them set shuffle=True in the config.
This PR also removes stratification of the data during train-test splitting as a parameter directly mentioned in the docs, as it is not implemented for shuffle=False, and its use-case is only for classification problems. Using it correctly requires the user to do some reading up on its functionality. The docs already points the interested user to the documentation for the input parameter of train_test_split() with a direct url link.


robcaulk (approved):
LGTM

PR #7500, opened 2022-09-29
Profit Limit Protection

smarmau:
## Summary

Adds a customisable profit limit protection to freqtrade 

## What's new?

Allows you to set a customisable profit limit (ie stop trading for 12 hours once 1% total profit has been reached. 

First PR so all feedback taken onboard.


freqtrade/plugins/protections/profit_limit.py
```diff
@@ -0,0 +1,91 @@
...
+
+        trades = Trade.get_trades_proxy(is_open=False, close_date=look_back_until)
+        # trades = Trade.get_trades(filters).all()
+        if len(trades) < self._trade_limit:
+            # Not enough trades in the relevant period
+            return None
+
+        profit = sum(
+            trade.close_profit for trade in trades if trade.close_profit
+            )
```
xmatthias: You'll need to be careful with this - aggregating percentages like this is problematic and only works in very specific cases (when all stakes are identical).

assume i make the following 3 trades:
* 1% on a 100$ trade (abs 1$)
* 0.5% on a 1000$ trade (abs 50$)
* -2% on a 50$ trade (-1$)

My profit sum (as calculated above) of these 3 trades is -0.5%.
my actual profit is +50$.

the percentages need to be a weighted sum ... We do this for example in the /daily endpoint - so you might get ideas on an approach for this there.

A (simpler) alternative would be to use absolute values. It'll not scale if we assume the account constantly grows - but it would make the calculation / logic a lot simpler.

smarmau: Thanks xmatthias, I hadn't considered that. Code updated to the simpler absolute methodolgy.

freqtrade/plugins/protections/profit_limit.py
```diff
@@ -0,0 +1,86 @@
...
+        look_back_until = date_now - timedelta(minutes=self._lookback_period)
+
+        trades = Trade.get_trades_proxy(is_open=False, close_date=look_back_until)
+
+        if len(trades) < self._trade_limit:
+            # Not enough trades in the relevant period
+            return None
+
+        profit_sum = sum(trade.close_profit_abs for trade in trades if trade.close_profit_abs)
+        stake_sum = sum(trade.stake_amount for trade in trades)
```
xmatthias, line 53: usage of `trade.stake_amount` is (also) problematic as it will update on partial entries / exits.

we could therefore have an extreme case of profit taking where:
* enters 10000$
* exits 99%  of the trade (with a profit of - say 200$  (which is 2% of the 10.000$) - leaving the remaining stake_amount at 100$

the remaining stake (100$) would now be used for the above calculation - so the profit would seem to be 200% - but in reality - it's been 2%.

docs/includes/protections.md
```diff
@@ -139,6 +141,27 @@ def protections(self):
     This Protection applies only at pair-level, and will never lock all pairs globally.
     This Protection does not consider `lookback_period` as it only looks at the latest trade.
 
+#### Profit Limit
+
+`ProfitLimit` uses all trades for a pair within `lookback_period` in minutes (or in candles when using `lookback_period_candles`) to determine the overall profit ratio.
+If that ratio is greater than `profit_limit`, all pairs will be locked for `stop_duration` in minutes (or in candles when using `stop_duration_candles`).
```
xmatthias, line 147: I'm not entirely sure if the current stop-duration concept (the way we use it for other protections) works for this limitation.

A common usecase I've heard is the possibility to "stop trading for today" once a certain profit has been reached.
I'd however not know how to configure this at this point - if i use "stop_duration" 180 (3h - as in the example) - it'd work fine if i reached the profit at 9pm - but it'll fail if i have the required 1% profit already at 3am - as the bot would continue trading after 6am (3 hours after the lock).

xmatthias (commented):
I'm not convinced on the idea of this.
It makes sense to stop trading if you're losing (bad market conditions) - but to stop trading when winning sounds counter-intuitive.

It's not up to me to decide this (I'll not be a user of this protection, ...) - i do however think that the calculation will need some more thought to cover different trade scenarios (see separate comment).
We many also need to expand the "lock  duration" possibility - either limited to this protection - or applicable for all protections (which might be simpler in the long run ...).

gaugau3000:
In fine when if you look at past profit/loss you trade the equity curve. The problem with this that all the trades are independant and have the same probabibility to appends so use past trades is irrevevant (we say trades have no serial correlation). It is like say in roulette this number happens twice i will not put money on it the third time. Below the ressources : 
-> https://www.youtube.com/watch?v=O-kZQUf297U (empiral)
-> https://qoppac.blogspot.com/2015/11/random-data-evaluating-trading-equity.html (study)

Have a nice day.

PR #7508, opened 2022-09-30
Fix feature list match for PCA

aemr3:
## Summary
Currently pca transformation is not working correctly because feature list doesn't match with training features. 
This is an attempt to solve the following error messages we get now if we enable PCA:

```
File "/home/ubuntu/Temp/freqtrade/freqtrade/freqai/freqai_interface.py", line 432, in check_if_feature_list_matches_strategy
    dk.find_features(dataframe)
File "/home/ubuntu/Temp/freqtrade/freqtrade/freqai/data_kitchen.py", line 885, in find_features
    raise OperationalException("Could not find any features!")
freqtrade.exceptions.OperationalException: Could not find any features!
2022-09-30 06:24:17,563 - freqtrade.strategy.interface - WARNING - Unable to analyze candle (OHLCV) data for pair EOS/USDT: Could not find any features!
```

### DIFF
--- a/freqtrade/freqai/base_models/BaseClassifierModel.py
+++ b/freqtrade/freqai/base_models/BaseClassifierModel.py
@@ -92,7 +92,7 @@ class BaseClassifierModel(IFreqaiModel):
         filtered_df = dk.normalize_data_from_metadata(filtered_df)
         dk.data_dictionary["prediction_features"] = filtered_df
 
-        self.data_cleaning_predict(dk, filtered_df)
+        self.data_cleaning_predict(dk)
 
         predictions = self.model.predict(dk.data_dictionary["prediction_features"])
         pred_df = DataFrame(predictions, columns=dk.label_list)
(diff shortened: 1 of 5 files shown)


robcaulk (approved):
I appreciate the effort you spent debugging this, thank you. I knew this one wasnt as straightforward as other PRs due to the nature of how we are checking feature lists for the user and how we do it slightly different depending on live, backtest, train, predict, and backtest-load-from-file...

Needless to say, this one was a bit surgical, and I took the opportunity to fix `inlier_metric` here in addition to the referenced PCA bug.

There is one issue with your approach: with your method the user could try to reuse the wrong PCA model and get "lucky" with the number of new principal components. In this case, FreqAI would silently let the user reuse that old model. 

But fixing it properly exposes an architectural weakness in FreqAI that I would like to ameliorate. Here I have decided that we will simply enforce the raw features engineered off of `populate_any_indicators` output to be identical.  We sacrifice whether or not `inlier_metric` was added so that we can maintain the backtesting performance benefits (where we dont want to recompute `inlier_metric`), but luckily the regressor itself has its own error checking for feature string names/count. 

I would ultimately prefer that we generalize adding additional features in the way that `inlier_metric` does. Right now it is the only function that does this but it would be nice if it were more robust and less spread across functions. But that is for a different PR.

robcaulk (approved):
LGTM

PR #7511, opened 2022-10-02
Fix typos and correct/improve descriptions

th0rntwig:
## Summary

Fixed some random typos and corrected/improved some descriptions for clarity.


docs/freqai-running.md
```diff
@@ -1,18 +1,18 @@
...
+# Running `FreqAI`
 
 There are two ways to train and deploy an adaptive machine learning model - live deployment and historical backtesting. In both cases, `FreqAI` runs/simulates periodic retraining of models as shown in the following figure:
 
 ![freqai-window](assets/freqai_moving-window.jpg)
 
 ## Live deployments
 
-FreqAI can be run dry/live using the following command:
+`FreqAI` can be run dry/live using the following command:
```
xmatthias: i'm not sure about this - FreqAI is not really code - but a name.
so wrapping it in ` seems wrong to me - and rendering will most likely look better without the ticks.

th0rntwig: I was about to change all occurrences to the non `` version because I agree with you, it's not code, but there seemed to be more `` than non `` in the previous version so I went the other way. Imo it renders well in most places, except for in, e.g., the parameter table where it gets mixed in with actual code and it becomes somewhat cluttered. 

@robcaulk  what is your opinion? I believe you were the one starting the ``?

xmatthias: we're also not using `freqtrade` throughout the docs, unless we refer to the command that's executed.
It's not just how it renders - but how people will perceive it.

if we wrap names like this - it'll be indistinguishable from actual code segments / things where we aim to highlight it's a variable or a setting.

xmatthias (changes requested):

docs/freqai-running.md
robcaulk: Yes, I was the one that changed them all to have backticks. Ok, we can change them back.

xmatthias: backticks are fine (and should be used) if we talk about class names - options, settings or commands.

so `freqai` (all lowercase) is perfectly fine and correct (the config segment) - but `freqAI` does not refer to any of the above ;)

xmatthias (approved):
Great 👍 

fine for me ...

PR #7512, opened 2022-10-02
add close price and date to `historic_predictions.pkl`

robcaulk:
Add close price and date at each prediction data point in the persistent `historic_predictions.pkl` to enable performance tracking and deeper post processing options.

### DIFF
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -294,6 +294,12 @@ class FreqaiDataDrawer:
             for return_str in rets:
                 df[return_str].iloc[-1] = rets[return_str]
 
+        # this logic carries users between version without needing to 
+        # change their identifier
+        if 'close_price' not in df.columns:
+            df['close_price'] = 0
+            df['date_pred'] = 0
+
         df['close_price'].iloc[-1] = strat_df['close'].iloc[-1]
         df['date_pred'].iloc[-1] = strat_df['date'].iloc[-1]
 
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -257,7 +257,7 @@ class FreqaiDataDrawer:
 
     def append_model_predictions(self, pair: str, predictions: DataFrame,
                                  do_preds: NDArray[np.int_],
-                                 dk: FreqaiDataKitchen, len_df: int) -> None:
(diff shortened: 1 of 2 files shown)


freqtrade/freqai/data_drawer.py
```diff
@@ -293,6 +294,9 @@ def append_model_predictions(self, pair: str, predictions: DataFrame,
             for return_str in rets:
                 df[return_str].iloc[-1] = rets[return_str]
 
+        df['close_price'].iloc[-1] = strat_df['close'].iloc[-1]
+        df['date_pred'].iloc[-1] = strat_df['date'].iloc[-1]
```
xmatthias, line 304: shouldn't close and date already be part of the dataframe when it comes in? 

will this be a breaking change - requiring identifier resets?

I'm also not sure on the "post processing options".
if i look at the below - then `.df` has a rolling length of the incoming dataframe.
That's for sure correct, but it'll mean that in the end, the data will age out.

robcaulk: Historic predictions dataframe is not the same as the incoming strategy dataframe. Nor is it the same as the returned dataframe . As mentioned in the PR description - historic_predictions.pkl is persistent . You can use code search to see how that is the case and where it is saved and how it is saved.

if users update versions they need to ensure a fresh identifier.

robcaulk: Historic predictions is an indefinite record of the bot’s run - we don’t phase historic predictions out. This is data for post processing.

freqtrade/freqai/data_drawer.py
```diff
@@ -293,6 +294,15 @@ def append_model_predictions(self, pair: str, predictions: DataFrame,
             for return_str in rets:
                 df[return_str].iloc[-1] = rets[return_str]
 
+        # this logic carries users between version without needing to 
+        # change their identifier
+        if 'close_price' not in df.columns:
+            df['close_price'] = 0
+            df['date_pred'] = 0
```
xmatthias: I think this field should have a different type / different default assignment (maybe np.NaN?)
Right now - we'd create an empty column of type int (most likely np.int64).

the actual date column is however of type pd.Timestamp ... so the below 2 lines will cause the type of this field to be object (it's mixed type now) ...

robcaulk: good point

xmatthias (approved):
LGTM :+1:

PR #7289, opened 2022-08-25
Add reinforcement learning module to FreqAI

robcaulk:
## Summary

This PR aims to add a flexible framework for deploying reinforcement learning models with FreqAI. 

## What is RL and why does FreqAI need it?

Reinforcement learning involves two important components, the *agent* and the training *environment*. During agent training, the agent moves through historical data candle by candle, always making 1 of a set of actions: Long entry, long exit, short entry, short exit, neutral). During this training process, the environment tracks the performance of these actions and rewards the agent according to a custom user made `calculate_reward()` (here we offer a default reward for users to build on if they wish). The reward is used to train weights in a neural network. 

A second important component of the proposed RL implementation is the use of *state* information. State information is fed into the network at each step, including current profit, current position, and current trade duration. These are used to train the agent in the training environment, and to reinforce the agent in dry/live. **Freqtrade is a perfect match for this reinforcing mechanism since this information is readily available in live deployements.** 

Reinforcement learning is a natural progression for FreqAI, since it adds a new layer of adaptivity and market reactivity that Classifiers and Regressors cannot match. However, Classifiers and Regressors have strengths that RL does not have such as robust predictions. Improperly trained RL agents may find "cheats" and "tricks" to maximize reward without actually winning any trades. For this reason, RL is more complex and demands a higher level of understanding than typical Classifiers and Regressors.

## How the present PR opens up a wide range of creative possibilities for Freqtrade users

With the current framework, we aim to expose the training environment to the user via the common "prediction model" file (i.e. CatboostClassifier, LightGBMRegressor, etc.). Users inherit our base environment in this file, which allows them to override as much or as little of the environment as they wish. 

We envision the majority of users focusing their effort on creative design of the `calculate_reward()` function, while leaving the rest of the environment untouched. Other users may not touch the environment at all, and they will only play with the configruation settings and the powerful feature engineering that already exists in FreqAI. Meanwhile, we enable advanced users to create their own model classes entirely (we offer an example of how an advanced user can begin doing something like that.)

The framework is built on stable_baselines3 (torch) and openai gym for the base environment class. But generally speaking, the model class is well isolated. Thus, the addition of competing libraries can be easily integrated into the existing framework (albeit with some basic assistance from core-dev). For the environment, it is inheriting from `gym.env` which means that a user would need to write an entirely new environment if they wish to switch to a different library. 

## Who is working on it?

If you are interested in helping to test, develop, and improve this Freqtrade feature, please join our discussions in the discord:

[...]

PR #7538, opened 2022-10-05
improve freqai tests

robcaulk:
Add coverage and slim down the freqai config to reduce test time. 

These are the same tests proposed by #7289, with the goal of isolating them and confirming that the large RL changes do not interfere with existing functionality.


tests/freqai/test_freqai_datakitchen.py
```diff
@@ -71,22 +71,22 @@ def test_use_DBSCAN_to_remove_outliers(mocker, freqai_conf, caplog):
     freqai = make_data_dictionary(mocker, freqai_conf)
     # freqai_conf['freqai']['feature_parameters'].update({"outlier_protection_percentage": 1})
     freqai.dk.use_DBSCAN_to_remove_outliers(predict=False)
-    assert log_has_re(r"DBSCAN found eps of 2\.3\d\.", caplog)
+    assert log_has_re(r"DBSCAN found eps of 1.75", caplog)
```
xmatthias: fails on arm ... 
`INFO     freqtrade.freqai.data_kitchen:data_kitchen.py:717 DBSCAN found eps of 1.74. `

i think this should be kept at 1 decimal (also keep in mind that `"."` in regex matches "everything" - so the above would match `1.75` - as well as `1075` or `1s75`).

robcaulk: Ah missed it...

Should be done now.

xmatthias (approved):
Great! 

LGTM 👍

PR #7544, opened 2022-10-06
Remove constant labels from prediction

### LINKED ISSUE #7542: Feature shape mismatch during prediction
## Describe your environment

  * Operating system: Ubuntu 22.02 via WSL2 on Windows 11
  * Python Version: 3.8
  * CCXT version: Specified in Freqtrade 2022.9.1
  * Freqtrade Version: 2022.9.1
  
## Describe the problem:
When running FreqAi, som of the features seems to get removed by data kitchen because of occuring constant values. This is causing a feature shape missmatch during training. It could be that some features are not constant when predictions start. 
  
### Observed Results:

  * See logs below.
[log.txt](https://github.com/freqtrade/freqtrade/files/9726325/log.txt)

th0rntwig:
## Summary

Match prediction labels to training labels if constant features were removed from training data.
https://github.com/freqtrade/freqtrade/issues/7542#issue-1399878192

## Quick changelog

Constant features bring no information to the ML model and are therefore removed. This PR ensures that constant features are removed from prediction data so that the labels match the training data.

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -460,6 +460,18 @@ class FreqaiDataKitchen:
 
         return df
 
+    def check_pred_labels(self, df_predictions: DataFrame) -> None:
+        """
+        Check that prediction feature labels match training feature labels.
+        :params:
+        :df_predictions: incoming predictions
+        """
+        train_labels = self.data_dictionary["train_features"].columns
+        pred_labels = df_predictions.columns
+        if len(train_labels.difference(pred_labels)) != 0:
+            self.data_dictionary["prediction_features"] = df_predictions[train_labels]
+        return
+
     def principal_component_analysis(self) -> None:
         """
         Performs Principal Component Analysis on the data for dimensionality reduction
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -241,6 +241,7 @@ class FreqaiDataKitchen:
             self.data["filter_drop_index_training"] = drop_index
 
         else:
+            filtered_df = self.check_pred_labels(filtered_df)
             # we are backtesting so we need to preserve row number to send back to strategy,
             # so now we use do_predict to avoid any prediction based on a NaN
             drop_index = pd.isnull(filtered_df).any(axis=1)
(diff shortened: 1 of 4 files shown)


tests/freqai/test_freqai_interface.py
```diff
@@ -312,6 +320,8 @@ def test_follow_mode(mocker, freqai_conf):
     freqai.dd.load_all_pair_histories(timerange, freqai.dk)
 
     df = strategy.dp.get_pair_dataframe('ADA/BTC', '5m')
+    for i in range(5):
```
robcaulk: I don't think it is necessary to add this to all the functions - only `start_backtesting()` and `extract_data_and_train()` should be sufficient.

freqtrade/freqai/data_kitchen.py
```diff
@@ -460,6 +460,18 @@ def slice_dataframe(self, timerange: TimeRange, df: DataFrame) -> DataFrame:
...
+    def check_pred_labels(self, df_predictions: DataFrame) -> None:
+        """
+        Check that prediction feature labels match training feature labels.
+        :params:
+        :df_predictions: incoming predictions
+        """
+        train_labels = self.data_dictionary["train_features"].columns
+        pred_labels = df_predictions.columns
+        if len(train_labels.difference(pred_labels)) != 0:
+            self.data_dictionary["prediction_features"] = df_predictions[train_labels]
```
robcaulk: Can you put a log warning here, so that we as devs can see when this starts cropping up in bug reports (also doesn't hurt for the user to realize something "went wrong")

```
logger.warning(f'Removed {num_diffs} features from prediction features, these were likely considered constant values during most recent training')
```

Then you can assert the log existence in the tests that you edited (e.g. inside `test_start_backtesting()`, note: ensure to add `caplog` to the arguments of `test_start_backtesting(caplog)`) :

```
    assert log_has_re(
        "Removed 5 features from prediction features",
        caplog,
    )
 ```

robcaulk (changes requested):
Thanks a lot for taking the initiative on this bug - I like the fix. A few comments before it is ready to go.

tests/freqai/test_freqai_interface.py
th0rntwig: Kept only the check in `start_backtesting()` as there's no direct way of modifying the dataframe in the other test.

freqtrade/freqai/data_kitchen.py
th0rntwig: Adding the log warning and assert in the test showed that the shape check was placed incorrectly: the shape check was made after normalization which caused key errors. I now moved the check to `filter_features()` in `data_kitchen` which means it gets done before normalization.

Also added an assert log in `test_start_backtesting()` about dropping the constant features from the test data.

robcaulk (approved):
LGTM

PR #7550, opened 2022-10-07
Reset dataframe index after slice

aemr3:
## Summary

Heikinashi candles was throwing a KeyError exception due to the dataframe index. Added reset_index to sliced dataframes in get_base_and_corr_dataframes() for populate_any_indicators().

Thanks @xmatthias @robcaulk


robcaulk (approved):
Thanks for that, LGTM

PR #7554, opened 2022-10-08
Add tensorboard for catboost

initrv:
## Summary

The purpose of this PR is to add the ability to use the tensorboard with catboost models.
https://catboost.ai/en/docs/features/visualization_tensorboard

## Quick changelog

- Added necessary parameters to catboost prediction_models
- Added freqai requirement for tensorboard

## What's new?

The user can explore catboost models using ```tensorboard --logdir /path/to/model```
catboost-tensorboard


freqtrade/freqai/prediction_models/CatboostClassifier.py
```diff
@@ -32,8 +32,9 @@ def fit(self, data_dictionary: Dict, dk: FreqaiDataKitchen, **kwargs) -> Any:
         )
 
         cbr = CatBoostClassifier(
-            allow_writing_files=False,
+            allow_writing_files=True,
```
xmatthias, line 36: @robcaulk didn't we intentionally disable this setting as catboost then started to write random stuff in the current directory?

We can surely enable this if necessary - but we need to make sure it'll only write to the proper directory (the model directory) - not in random directories.

robcaulk: Yes, but since here we set `train_dir`,  it directs the output to that directory (the model directory).

robcaulk (commented):
I added a doc to help users understand usage (and this will replace the reinforcement learning tensorboard doc). 

I also decided that we should isolate the files into their own directory so that it keeps the model folder looking "clean." (catboost dumps quite a few files for this). 

Other models can now place their tensorboard logs into the same directory and tensorboard should have no problem finding them (e.g. RL).

docs/freqai-running.md
```diff
@@ -142,6 +142,20 @@ dataframe['outlier'] = np.where(dataframe['DI_values'] > self.di_max.value/10, 1
...
+Catboost models benefit from tracking training metrics via Tensorboard. You can take advantage of the FreqAI integration to track training and evaluation performance across all coins and across all retrainings. Tensorboard is activated via the following command:
+
+```bash
+cd freqtrade
+tensorboard --logdir user_data/models/unique-id
+```
+
+where `unique-id` is the `identifier` set in the `freqai` configuration file. This command must be run in a separate shell if the user wishes to view the output in their browser at 127.0.0.1:6060 (6060 is the default port used by Tensorboard).
+
+![tensorboard](assets/tensorboard.jpg)
```
xmatthias, line 156: @robcaulk We're missing the asset for this.

robcaulk: Just added it, should be good to go.

robcaulk (approved):
LGTM!!

PR #7555, opened 2022-10-09
Better Message WebSocket performance

wizrds:
## Summary

This PR aims to improve the overall performance of the Message Websocket. First, we move away from a naive broadcasting method to a queue-per-client method which should help latency between messages by not waiting on slow connections to send other messages. Next, the broadcasted dataframes have been changed from the full dataframe to the most recent candle. The initial request still requests the full dataframe to begin with, but then any subsequent dataframe messages are appended to the data already there instead of overwriting it. This should guarantee that dataframe messages (besides the initial messages) to be incredibly small.

### DIFF
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -9,7 +9,7 @@ from collections import deque
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Tuple
 
-from pandas import DataFrame
+from pandas import DataFrame, concat
 
 from freqtrade.configuration import TimeRange
 from freqtrade.constants import Config, ListPairsWithTimeframes, PairWithTimeframe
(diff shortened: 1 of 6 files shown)


xmatthias (changes requested):
I'd appreciate if PR's could be focused on one thing - so this PR will have to be split into at least 2
* change how queue works
* try to change how we transfer dataframes (this will need a LOT more work and is currently unstable - see below).

---

You're thinking about this WAY too simple (focused on the case you tested - with a stable pairlist).
if we think about what producer/consumer mode aims to do - then it's "transfer a dataframe from the producer to the consumer". 

While this does NOT imply that we must transfer whole dataframes  at each point - it does imply that freqtrade (as provider of producer/consumer mode) ensures that the dataframes are identical.

After just flying over the code - i can already tell that this is not the case in too many scenarios, and will therefore render this feature unstable - and as such unusable.

Some (probably not all) scenarios that will fail the consumer with different dataframes:
* changes to the producer's pairlist (the dataframe will be of length1 -  if it doesn't crash outright)
* missing candles (e.g. exchange downtimes, or producer bot internet outages)

---

based on this - i think it'd be best to for now remove the "partial" dataframe code from this PR - so we can look at the queue code in detail. 
you can obviously submit another PR which includes partial transmission - yet i am not convinced it's a good idea, as the "ensure the consumer dataframe is identical/equal" point is probably not something that can be archived easily without a lot of complications.

PR #7557, opened 2022-10-09
Add metric tracker to FreqAI

robcaulk:
An optional metric tracker which collects train timings, inference timings, and cpu load for each asset. The tracker writes results to disk in the model folder, so that the user can post-process results, or so that FreqAI can resume where it left off in case of a crash.


freqtrade/freqai/freqai_interface.py
```diff
@@ -658,7 +660,7 @@ def fit_live_predictions(self, dk: FreqaiDataKitchen, pair: str) -> None:
 
         return
 
-    def inference_timer(self, do='start'):
+    def inference_timer(self, do: str = 'start', pair: str = ''):
```
xmatthias: you might wanna type "do" as `Literal['start', 'stop']` ... that way mypy will ensure nobody calls this with something else.

freqtrade/freqai/data_drawer.py
```diff
@@ -110,7 +139,18 @@ def load_drawer_from_disk(self):
...
+    def load_metric_tracker_from_disk(self):
+        """
+        Tries to load an existing metrics dictionary if the user
+        wants to collect metrics.
+        """
+        if self.freqai_info.get('write_metrics_to_disk', False):
+            exists = self.metric_tracker_path.is_file()
+            if exists:
+                with open(self.metric_tracker_path, "r") as fp:
+                    self.metric_tracker = json.load(fp)
```
xmatthias: If we write with rapidjson - we should also use rapidjson (with the same flags) for loading to avoid problems and data-discrepancies after a few rounds of read/write (multiple restarts, basically).

freqtrade/freqai/data_drawer.py
```diff
@@ -78,25 +79,53 @@ def __init__(self, full_path: Path, config: Config, follow_mode: bool = False):
...
+
+            self.metric_tracker[pair][metric].append(value)
+
+    def collect_metrics(self, time_spent: float, pair: str):
+        """
+        Add metrics to the metric tracker dictionary
+        """
+        load1, load5, load15 = psutil.getloadavg()
+        cpus = psutil.cpu_count()
+        self.update_metric_tracker('train_time', time_spent, pair)
```
xmatthias, line 123: how often do we assume this is called at max? (per second / minute / hour / ...).

we have to treat freqtrade as a long-running process - which could be running for days/weeks/months without a restart.
this (or rather, update_metric_tracker) has every sign of a potential memory leak.

while the information in it seems negligible at first glance - i'm not sure it is if this is called frequently enough. 
we save a 4 floats (8bytes) per call. Assuming this is called once per second, then the object-structure will grow by `2_764_800` bytes (~2 mb) per day/pair - on a datastructure that's kept in memory - and loaded from disk on restart.

freqtrade/freqai/data_drawer.py
```diff
@@ -78,25 +79,53 @@ def __init__(self, full_path: Path, config: Config, follow_mode: bool = False):
...
+        General utility for adding and updating custom metrics. Typically used
+        for adding training performance, train timings, inferenc timings, cpu loads etc.
+        """
+        with self.metric_tracker_lock:
+            if pair not in self.metric_tracker:
+                self.metric_tracker[pair] = {}
+            if metric not in self.metric_tracker[pair]:
+                self.metric_tracker[pair][metric] = []
+
+            self.metric_tracker[pair][metric].append(value)
```
xmatthias: Opposing my other comment on data size - i don't think what you're currently saving makes sense.

it'll end as an array of "train_time, cpu1, cpu5, cpu15" - per pair - but it's lacking any time-reference (you'll have a datapoint, but you'll only know it's the 15th datapoint (measured from when?), not "this happened at 11:45 on the 15th october".

This will make correlating this to anything quite difficult / impossible - which makes me doubt why we'd need to save this info in the first place.

freqtrade/freqai/data_drawer.py
robcaulk: This is called once per train per coin and once per pair list inference (once per base candle). So the amount depends on train frequency and base timeframe.

freqtrade/freqai/data_drawer.py
robcaulk: We don't care about *when* these are called. These are for understanding statistical representations of these metrics.

Inference timings vector wont be the same length as train timings vector.

freqtrade/freqai/data_drawer.py
xmatthias: so assuming we're on the 1m timeframe, we're down to ~46kb/day/pair - which i guess is ok, especially considering this needs to be explicitly activated.

freqtrade/freqai/data_drawer.py
xmatthias: i'm not sure that'll hold true.

assuming train/inference timings change after 100 entries - you'd want to corelate this to some other measure to identify why this happened.

I've seen this often enough at work - time-based data without time reference will end up as useless data.

robcaulk: Well it is easy enough to add the timestamp just in-case, however each list needs its own timestamp - so it doubles the disk space. But as you checked earlier - disk space is not an issue at this write frequency.

freqtrade/freqai/data_drawer.py
robcaulk: Done!

freqtrade/freqai/freqai_interface.py
robcaulk: Done!

xmatthias (approved):
LGTM :+1:

PR #7558, opened 2022-10-09
Refactor broadcasting in Message Websocket

wizrds:
## Summary

This PR aims to increase performance with slow connections in the message websocket.

## Quick changelog

- Update WebSocketChannel with a queue attribute and relay method for sending messages in the queue
- Rewrite ChannelManager broadcast to append messages to the client's queue instead of directly sending
- The builtin websockets package ping interval has been disabled completely, as we take care of pinging ourselves.
- The `wait_timeout` has been updated to 30 seconds instead of 300 to better support reverse proxies closing connections if no data is sent after a certain time.

## What's new?

The WebSocketChannel now has a `relay` method that is started as a task when instantiated, which runs in the background sending messages from the queue for that specific channel. This ensures that whenever a message is broadcasted, there's not an increase in latency for clients that have poor connection or not receiving data.

### DIFF
--- a/freqtrade/rpc/api_server/api_ws.py
+++ b/freqtrade/rpc/api_server/api_ws.py
@@ -1,10 +1,10 @@
 import logging
 from typing import Any, Dict
 
-import websockets
 from fastapi import APIRouter, Depends, WebSocketDisconnect
 from fastapi.websockets import WebSocket, WebSocketState
 from pydantic import ValidationError
+from websockets.exceptions import WebSocketException
 
 from freqtrade.enums import RPCMessageType, RPCRequestType
 from freqtrade.rpc.api_server.api_auth import validate_ws_token
@@ -115,10 +115,7 @@ async def message_endpoint(
                     # Process the request here
                     await _process_consumer_request(request, channel, rpc)
 
-            except (
-                WebSocketDisconnect,
-                websockets.exceptions.WebSocketException
-            ):
+            except (WebSocketDisconnect, WebSocketException):
                 # Handle client disconnects
                 logger.info(f"Consumer disconnected - {channel}")
             except RuntimeError:
@@ -126,7 +123,7 @@ async def message_endpoint(
                 # RuntimeError('Cannot call "send" once a closed message has been sent')
                 pass
             except Exception as e:
-                logger.info(f"Consumer connection failed - {channel}")
+                logger.info(f"Consumer connection failed - {channel}: {e}")
                 logger.debug(e, exc_info=e)
             finally:
                 await channel_manager.on_disconnect(ws)
--- a/freqtrade/rpc/api_server/api_ws.py
+++ b/freqtrade/rpc/api_server/api_ws.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Any, Dict
 
+import websockets
 from fastapi import APIRouter, Depends, WebSocketDisconnect
 from fastapi.websockets import WebSocket, WebSocketState
 from pydantic import ValidationError
@@ -102,7 +103,6 @@ async def message_endpoint(
     """
     try:
         channel = await channel_manager.on_connect(ws)
-
         if await is_websocket_alive(ws):
 
             logger.info(f"Consumer connected - {channel}")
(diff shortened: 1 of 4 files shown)


freqtrade/rpc/api_server/ws/channel.py
```diff
@@ -34,6 +35,8 @@ def __init__(
         self._serializer_cls = serializer_cls
 
         self._subscriptions: List[str] = []
+        self.queue: asyncio.Queue[Dict[str, Any]] = asyncio.Queue(maxsize=32)
```
xmatthias, line 38: this means we have at max. 32 messages per consumer in the queue, right?

freqtrade/rpc/api_server/ws/channel.py
```diff
@@ -95,6 +105,20 @@ def subscribed_to(self, message_type: str) -> bool:
...
+        """
+        Relay messages from the channel's queue and send them out. This is started
+        as a task.
+        """
+        while True:
+            message = await self.queue.get()
+            try:
+                await self._send(message)
+                self.queue.task_done()
+            except RuntimeError:
```
xmatthias, line 124: would it make sense to move the "sleep" here - so we can send unlimited messages from the main queue, and are eventually limiting performance here?

freqtrade/rpc/api_server/api_ws.py
```diff
@@ -115,26 +115,34 @@ async def message_endpoint(
                     # Process the request here
                     await _process_consumer_request(request, channel, rpc)
 
-            except WebSocketDisconnect:
+            except (
+                WebSocketDisconnect,
+                websockets.exceptions.WebSocketException
+            ):
```
xmatthias: If you import the `WebSocketException` directly, it'll make this line format nicer.
we don't use websockets elsewhere in this file - so it would make this part a bit nicer to look at / read

freqtrade/rpc/api_server/api_ws.py
```diff
@@ -115,26 +115,34 @@ async def message_endpoint(
...
-            except Exception as e:
-                logger.info(f"Consumer connection failed - {channel}")
-                logger.exception(e)
+            except RuntimeError:
                 # Handle cases like -
                 # RuntimeError('Cannot call "send" once a closed message has been sent')
+                pass
+            except Exception as e:
+                logger.info(f"Consumer connection failed - {channel}")
+                logger.debug(e, exc_info=e)
```
xmatthias: we're not interested why a connection failed?
i think at least logging the exception type on a visible log-level would make sense... ?

freqtrade/rpc/external_message_consumer.py
```diff
@@ -62,7 +62,7 @@ def __init__(
         self.enabled = self._emc_config.get('enabled', False)
         self.producers: List[Producer] = self._emc_config.get('producers', [])
 
-        self.wait_timeout = self._emc_config.get('wait_timeout', 300)  # in seconds
+        self.wait_timeout = self._emc_config.get('wait_timeout', 30)  # in seconds
```
xmatthias, line 65: > The wait_timeout has been updated to 30 seconds instead of 300 to better support reverse proxies closing connections if no data is sent after a certain time.

I'm not sure about this point. To me, this means that the reverseproxy should be reconfigured.

It'll increase our ping frequency from 1/5m to 2/1m - which is quite a lot more - so i think 30 is too low.

freqtrade/rpc/api_server/ws/channel.py
wizrds: Yes, the internal queue the `websockets` package uses defaults to 32 as well, so this is set to match that.

freqtrade/rpc/api_server/ws/channel.py
wizrds: I agree, I will make this change

freqtrade/rpc/api_server/api_ws.py
wizrds: Oops good catch

xmatthias (approved):
LGTM 👍

PR #7560, opened 2022-10-10
Update freqai-spice-rack.md

smarmau:
Instructs newer users to place the code calling spice_rack in populate_indicators


robcaulk (approved):
Thanks for that

PR #7569, opened 2022-10-10
Add XGBoost random forest predictors to freqai

Silur:
As gradient boosting underperforms on noisy data - which is to be expected on low timeframes, this PR includes random forest predictors for FreqAI that are generally more robust against overfitting when training data is expected to be noisy. Thankfully the XGB API is general enough so it only needed minimal changes based on the GB model.

## Changelog
 - add `freqtrade/freqai/prediction_models/XGBoostRFClassifier.py`
 - add `freqtrade/freqai/prediction_models/XGBoostRFRegressor.py`

### DIFF
--- a/freqtrade/freqai/prediction_models/XGBoostRFClassifier.py
+++ b/freqtrade/freqai/prediction_models/XGBoostRFClassifier.py
@@ -1,0 +1,85 @@
+import logging
+from typing import Any, Dict, Tuple
+
+import numpy as np
+import numpy.typing as npt
+import pandas as pd
+from pandas import DataFrame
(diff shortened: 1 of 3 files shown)


robcaulk:
Interesting, thanks for this. Do you want to give a whack at the tests? (I think it may be as easy as just adding the string names to our existing lists)

https://github.com/freqtrade/freqtrade/blob/develop/tests/freqai/test_freqai_interface.py#L32
https://github.com/freqtrade/freqtrade/blob/develop/tests/freqai/test_freqai_interface.py#L113

robcaulk (approved):
Thanks for this!

PR #7571, opened 2022-10-10
document user_data/freqaimodels

### LINKED ISSUE #7570: Allow freqai to pull prediction_models from user_data
Currently, only classes present in `freqai/prediction_models` are available for backtesting/trading.
Allowing the user to define a custom model to be used with `--freqaimodel` would allow more flexibility.

xmatthias:
## Summary

Improve usability of freqaimodels

Solve the issue: #7570 

## Quick changelog

- auto-create freqaimodel  directory when using `create-userdir`
- document freqaimodel directory as a location we load models from.


robcaulk (approved):
Thanks for that - we had a user confused by this last week.

PR #7572, opened 2022-10-11
Test WebSocket client fix

wizrds:
## Summary

This is a few small fixes to the Testing WebSocket client under scripts

## Quick changelog

- Changed timestamp calculations from millisecond to microsecond to better match the readable delta
- Disabled internal ping interval as we ping explicitly


xmatthias (approved):
LGTM 👍

PR #7582, opened 2022-10-13
Add model info

th0rntwig:
## Summary

Added information about the available ML models.

## Quick changelog

The ML libraries currently available in FreqAI were, in previous versions of the doc, mentioned but not described nor were they links to their corresponding docs. This PR adds a short description of the libs, giving the user some key words that they can use to search for more info, and links to the lib docs. This PR also adds descriptions on how to set the labels for regressor and classification models, to help users not fall into the trap of using the wrong label type.


docs/freqai-configuration.md
```diff
@@ -204,17 +204,44 @@ If this value is set, FreqAI will initially use the predictions from the trainin
 
 ## Using different prediction models
 
-FreqAI has multiple example prediction model libraries that are ready to be used as is via the flag `--freqaimodel`. These libraries include `Catboost`, `LightGBM`, and `XGBoost` regression, classification, and multi-target models, and can be found in `freqai/prediction_models/`. However, it is possible to customize and create your own prediction models using the `IFreqaiModel` class. You are encouraged to inherit `fit()`, `train()`, and `predict()` to let these customize various aspects of the training procedures.
+FreqAI has multiple example prediction model libraries that are ready to be used as is via the flag `--freqaimodel`. These libraries include `CatBoost`, `LightGBM`, and `XGBoost` regression, classification, and multi-target models, and can be found in `freqai/prediction_models/`.
```
xmatthias, line 207: we should be cautious about pointing people to internal code directories (like `freqtrade/freqai/prediction_models`). 
That's not available easily depending on the install method (basically in any method which doesn't do a checkout first, like docker or `pip install freqtrade`).

The most likely result will be that we'll have to list all builtin models explicitly someplace in the docs.

xmatthias (approved):
Great enhancement! :+1: 
Fixed a few spelling mistakes (my spell-check plugin works great! :laughing: ) - and it's good to go :+1:

docs/freqai-configuration.md
th0rntwig: Good point. I do think listing all in the docs is not a bad thing, although it's something that needs to be maintained. But listing and linking to the lib docs gives users more power over how to use the software so it's definitely worth it.

th0rntwig:
> Great enhancement! 👍 Fixed a few spelling mistakes (my spell-check plugin works great! 😆 ) - and it's good to go 👍

Thanks for fixing that! Seems I am in desperate need of a spell-check plugin of my own..

PR #7593, opened 2022-10-15
Fix constant PCA

th0rntwig:
## Summary

Bug fix for KeyError() trying to remove constant features with PCA enabled.

## Quick changelog

Constant features are removed as they do not contribute to the model. When PCA is enabled, the constant features that were removed no longer matches the df keys and throws an error. This PR disables the removal of constant features if PCA is enabled, as PCA will "remove" them by itself (it only keeps components influencing the variance of the data and a constant feature has little influence).

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -206,11 +206,15 @@ class FreqaiDataKitchen:
 
         drop_index = pd.isnull(filtered_df).any(axis=1)  # get the rows that have NaNs,
         drop_index = drop_index.replace(True, 1).replace(False, 0)  # pep8 requirement.
+        ft_params = self.freqai_config["feature_parameters"]
         if (training_filter):
-            const_cols = list((filtered_df.nunique() == 1).loc[lambda x: x].index)
-            if const_cols:
-                filtered_df = filtered_df.filter(filtered_df.columns.difference(const_cols))
-                logger.warning(f"Removed features {const_cols} with constant values.")
+            if not ft_params.get(
+                "principal_component_analysis", False
+            ):
+                const_cols = list((filtered_df.nunique() == 1).loc[lambda x: x].index)
+                if const_cols:
+                    filtered_df = filtered_df.filter(filtered_df.columns.difference(const_cols))
+                    logger.warning(f"Removed features {const_cols} with constant values.")
             # we don't care about total row number (total no. datapoints) in training, we only care
             # about removing any row with NaNs
             # if labels has multiple columns (user wants to train multiple modelEs), we detect here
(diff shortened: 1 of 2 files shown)


freqtrade/freqai/data_kitchen.py
```diff
@@ -206,11 +206,15 @@ def filter_features(
 
         drop_index = pd.isnull(filtered_df).any(axis=1)  # get the rows that have NaNs,
         drop_index = drop_index.replace(True, 1).replace(False, 0)  # pep8 requirement.
+        ft_params = self.freqai_config["feature_parameters"]
         if (training_filter):
-            const_cols = list((filtered_df.nunique() == 1).loc[lambda x: x].index)
-            if const_cols:
-                filtered_df = filtered_df.filter(filtered_df.columns.difference(const_cols))
-                logger.warning(f"Removed features {const_cols} with constant values.")
+            if not ft_params.get(
```
robcaulk: I think we still want to check for constant columns when we filter for training data (even when `principal_component_analysis` is true). This occurs *before* PCA, and we want to feed the non constant cols into PCA. 

Correct me if I am missing something.

robcaulk (changes requested):

freqtrade/freqai/data_kitchen.py
th0rntwig: The idea with this was that since PCA basically looks for variance in the data, a constant feature would be "automatically deselected" and we wouldn't need to find it and remove it before, reducing computational power (very slightly but still). But depending on future additions to the code, making this check agnostic to manipulations of the feature labels would indeed be the better way of doing it. I'll make the required changes!

freqtrade/freqai/data_kitchen.py
```diff
@@ -241,7 +245,8 @@ def filter_features(
             self.data["filter_drop_index_training"] = drop_index
 
         else:
-            filtered_df = self.check_pred_labels(filtered_df)
+            if len(self.constant_features_list):
```
robcaulk: I don't see how this will ever be anything besides an empty list. 

We train in a separate thread to the prediction (inferencing). You are attaching `self.constant_feature_list` to the temporary `dk` object. If you want information to persist from training to inference, you need to use `self.data` because `dk.data` is loaded in to the data kitchen eachtime a temporary data kitchen is created for a new coin.

th0rntwig:
I changed the storage location for the constant column labels to `datakitchen.data()`. This caused the `test_normalize_data` to throw an error as it checked for the length of `datakitchen.data()`. That test now checks for entries `_max` and `_min` (substrings of the entries added in `normalize_data`) in the dict keys instead of its length.

robcaulk (approved):
LGTM

PR #7596, opened 2022-10-16
Fix counting available trade slots in backtesting.

### LINKED ISSUE #7592: Backtest incorrectly processes timed out orders
## Describe your environment

  * Operating system: Windows 10
  * Python Version: 3.10.8 
  * CCXT version: 1.93.98
  * Freqtrade Version: 2022.9.1

## Describe the problem:

I wrote a simple strategy that places limit orders every candle. Price calculations are based on previous candles (such indicators as ATR, Supertrend, nPOC, etc.) So the basic idea is to place a long order below the current price and wait for it to be filled. If it doesn't fill during the current candle then cancel it. On new candle make recalculations and place a new order. Strategy uses 1 pair and opens max 1 trade.
During backtesting I noticed that orders were placed every two candles, so some orders were skipped. This issue doesn't depend on the timeframe.

### Steps to reproduce:

test_strategy.py:
```
import logging
from datetime import datetime
from functools import reduce
from typing import Optional

from pandas import DataFrame

from freqtrade.strategy import (IStrategy)

class TestStrategy(IStrategy):

    startup_candle_count = 10
    stoploss = -0.03
    minimal_roi = {"0": 100}
    can_short: bool = False

        return dataframe

        long_conditions = [dataframe['volume'] > 0]
        if long_conditions:
            dataframe.loc[
                reduce(lambda x, y: x & y, long_conditions),
                'enter_long'] = 1

        return dataframe

    def populate_exit_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        return dataframe

    def custom_entry_price(self, pair: str, current_time: datetime, proposed_rate: float,
                           entry_tag: Optional[str], side: str, **kwargs) -> float:
        message = f'>>> ENTRY >>> time: {current_time}'
        logger.info(message)
        return 1000.0
```

test-config.json:
```
{
  "dry_run": true,
  "initial_state": "running",
  "trading_mode": "futures",
  "margin_mode": "isolated",
  "futures_funding_rate": 0.0,
  "process_only_new_candles": true,
  "max_open_trades": 1,
  "stake_currency": "USDT",
  "stake_amount": "unlimited",
  "tradable_balance_ratio": 0.99,
  "stoploss": -0.03,
  "timeframe": "1m",
  "order_types": {
    "entry": "limit",
    "exit": "limit",
    "stoploss": "market",
    "stoploss_on_exchange": false
  },
  "ignore_buying_expired_candle_after": 10,
  "unfilledtimeout": {
    "entry": 50,
    "exit": 50,
    "unit": "seconds"
  },
  "exchange": {
    "name": "okx",
    "pair_whitelist": [
      "BTC/USDT:USDT"
    ]
  },
  "entry_pricing": {
    "use_order_book": true,
    "order_book_top": 1,
    "price_side": "other",
    "price_last_balance": 0.0,
    "check_depth_of_market": {
      "enabled": false,
      "bids_to_ask_delta": 1
    }
  },
  "exit_pricing": {
    "use_order_book": true,
    "order_book_top": 1,
    "price_side": "other"
  },
  "pairlists": [
    {
      "method": "StaticPairList"
    }
  ]
}
```

backtest command:
`freqtrade backtesting -c ./user_data/test-config.json --strategy TestStrategy -p BTC/USDT:USDT --timeframe 1m --timerange=20221013-`
  
### Observed Results:

  * What happened?
 Strategy uses 1m timeframe. Look at the orders time, they were placed every 2 minutes instead of 1.
```
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:09:00+00:00
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:11:00+00:00
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:13:00+00:00
```

```
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:09:00+00:00
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:10:00+00:00
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:11:00+00:00
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:12:00+00:00
test_strategy - INFO - >>> ENTRY >>> time: 2022-10-13 07:13:00+00:00
```

### Results
In trade mode everything works fine, so I started to debug backtesting and found the reason of such behavior. When the backtest cancels the order, it [only decrement the open_trade_count](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/optimize/backtesting.py#L1126), but later uses [open_trade_count_start to check for available trading slots](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/optimize/backtesting.py#L1138). As a result, it always gets one slot less than actually available (zero in my case). 
I just added `open_trade_count_start -= 1` to the _Close trade_ block (after line 1126) and now it works as expected. All tests passed successfully and I wrote another test for the described case.

Please check everything I wrote in this bug report. And if this is really a bug, then I have a ready-to-commit code with a fix and a test:)

iprogger:
## Summary

When the backtest cancels an order, it should decrement the open_trade_count_start value, which is later used to check for available trading slots.

Solve the issue: #7592

## Quick changelog

- decrement the `open_trade_count_start` when canceling an order
- add a new test strategy with a custom entry price
- add a test for the issue

### DIFF
--- a/freqtrade/optimize/backtesting.py
+++ b/freqtrade/optimize/backtesting.py
@@ -1123,6 +1123,7 @@ class Backtesting:
                     if self.manage_open_orders(t, current_time, row):
                         # Close trade
                         open_trade_count -= 1
+                        open_trade_count_start -= 1
                         open_trades[pair].remove(t)
                         LocalTrade.trades_open.remove(t)
                         self.wallets.update()
--- a/tests/optimize/test_backtesting.py
+++ b/tests/optimize/test_backtesting.py
@@ -799,6 +799,35 @@ def test_backtest_one(default_conf, fee, mocker, testdatadir) -> None:
                 t["close_rate"], 6) < round(ln.iloc[0]["high"], 6))
 
 
+def test_backtest_timedout_entry_orders(default_conf, fee, mocker, testdatadir) -> None:
+    # This strategy intentionally places unfillable orders.
+    default_conf['strategy'] = 'StrategyTestV3CustomEntryPrice'
+    default_conf['startup_candle_count'] = 0
+    # Cancel unfilled order after 4 minutes on 5m timeframe.
+    default_conf["unfilledtimeout"] = {"entry": 4}
+    mocker.patch('freqtrade.exchange.Exchange.get_fee', fee)
+    mocker.patch("freqtrade.exchange.Exchange.get_min_pair_stake_amount", return_value=0.00001)
+    mocker.patch("freqtrade.exchange.Exchange.get_max_pair_stake_amount", return_value=float('inf'))
+    patch_exchange(mocker)
+    backtesting = Backtesting(default_conf)
+    backtesting._set_strategy(backtesting.strategylist[0])
+    # Testing dataframe contains 11 candles. Expecting 10 timed out orders.
+    timerange = TimeRange('date', 'date', 1517227800, 1517231100)
+    data = history.load_data(datadir=testdatadir, timeframe='5m', pairs=['UNITTEST/BTC'],
+                             timerange=timerange)
+    min_date, max_date = get_timerange(data)
+
+    result = backtesting.backtest(
+        processed=deepcopy(data),
+        start_date=min_date,
+        end_date=max_date,
+        max_open_trades=1,
+        position_stacking=False,
+    )
+
+    assert result['timedout_entry_orders'] == 10
+
+
 def test_backtest_1min_timeframe(default_conf, fee, mocker, testdatadir) -> None:
     default_conf['use_exit_signal'] = False
     mocker.patch('freqtrade.exchange.Exchange.get_fee', fee)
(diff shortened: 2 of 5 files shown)


iprogger:
Oops, I ran pytest locally only for test_backtesting.py. Making a fix...

xmatthias (approved):
quite an edgecase - only reproduces (Reliably) if you use one pair / one max_open_trades ... but you did a great job analyzing, and then fixing this :100: 

Nice spot and fix! :+1:

iprogger:
Thank you! Yeah, I'm a simple man, trade only btc😄 But it also reproduces with more pairs and max_open_trades, I just tested it with `pairs=['UNITTEST/BTC', 'ETH/BTC', 'LTC/BTC']` and `max_open_trades=3`.

xmatthias:
yes - but it doesn't if max_open_trades is > number of pairs :)

PR #7607, opened 2022-10-18
Docker compose now in docker

matteoettam09:
## Summary

Docker compose is now part of docker.

```
docker-compose 
```
does not work anymore.

## Quick changelog

just removed all the problematic *-*

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -20,7 +20,7 @@ Please do not use bug reports to request new features.
   * Operating system: ____
   * Python Version: _____ (`python -V`)
   * CCXT version: _____ (`pip freeze | grep ccxt`)
-  * Freqtrade Version: ____ (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
+  * Freqtrade Version: ____ (`freqtrade -V` or `docker compose run --rm freqtrade -V` for Freqtrade running in docker)
   
 Note: All issues other than enhancement requests will be closed without further comment if the above template is deleted or not filled out.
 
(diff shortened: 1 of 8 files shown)


xmatthias:
hi,

have you got a source for this? (which version of docker this was added as builtin command, if it's installed automatically, ...)

As far as i could find, it's a _plugin_ to docker, which is not available by default, but needs to be [installed explicitly](https://docs.docker.com/compose/install/linux/).

docker-compose can still be installed as [standalone version](https://docs.docker.com/compose/install/other/#install-compose-standalone) - or even via pip  - which used to be the case so far, too, and is basically how the installation for docker-compose worked until now.

matteoettam09:
Hello, 
sorry. Indeed, you are right: now docker compose can be installed as a plugin of docker, which makes the command 
```
docker compose
```
work. So it is not installed naturally as part of docker, I was wrong.

And I didn't dig into the fact that you could still install a standalone version associated with the 
```
docker-compose 
```
command.

Sorry for the wrong request, feel free to close it.

xmatthias:
i assume the plugin route is the future of docker compose - but we do have to mention this explicitly (has to have the plugin installed, ...).

I'm not decided on this yet - so i'd like to keep this open for a bit to think on the best road for this case.

rzhb:
It's `Docker Desktop` now.
`Docker Desktop` contains `Docker CLI client` `Docker Compose`. After installed, `docker` `docker compose` `docker-compose` all should work.

I made a pull request to @matteoettam09 's branch.

xmatthias:
> It's `Docker Desktop` now.

That's unfortunately incorrect, and only true for Windows (and potentially macOS).

On linux, installation will differ between distribution and installation method - so stating it's "Docker Desktop" is incorrect (also on windows there's alternative methods to get docker working ...).

rzhb:
I don't think there is any difference. 
This is a "quick start". And the current link actually referred to "Docker Desktop" not "Docker CE".
Docker Desktop works for Linux Windows macOS.
Just check [What's included in Docker Desktop?](https://docs.docker.com/desktop/)

xmatthias (approved):
did some more rewording ...

LGTM :+1:

PR #7612, opened 2022-10-20
avoid redundant indicator population for corr_pair list

robcaulk:
Currently, the FreqAI framework populates the indicators for the `corr_pairlist` redundantly for all pairs for each candle. This PR removes the redundant indicator population and therefore improves time to prediction (reduces over all inference timing). 

Note: this same logic cannot be applied to the population of indicators prior to model training because we cannot make assumptions about train timings and therefore we'd not be certain that the corr_pair dataframe is updated. In contrast, when FreqAI inferences all pairs for each candle - FreqAI knows that the corr_pair dataframe is equally updated to the base asset and can therefore reuse the populated dataframe.


freqtrade/freqai/data_kitchen.py
```diff
@@ -1126,13 +1126,57 @@ def set_all_pairs(self) -> None:
...
+        in a dictionary to be reused and attached to other pairs.
+        :params:
+        :dataframe: fully populated dataframe (current pair + corr_pairs)
+        :return:
+        :corr_dataframes: dictionary of dataframes to be attached to other pairs in same candle.
+        """
+        corr_dataframes: Dict[str, DataFrame] = {}
+        pairs = self.freqai_config["feature_parameters"].get("include_corr_pairlist", [])
+
+        for pair in pairs:
```
th0rntwig, line 1155: I think we can improve performance by not looping through the pairs, both here and in `attach_corr_pair_columns`. 
Here we could simply do:
` corr_pairs = [pair.split('/')[0] for pair in pairs]`
  `corr_dataframes = dataframe.filter(regex='|'.join(corr_pairs))`
And in `attach_corr_pair_columns`:
`dataframe = pd.concat([dataframe, corr_dataframes], axis=1)`
` if len(dataframe.columns.unique()) < len(dataframe.columns):`
            `dataframe = dataframe[[dataframe.columns.unique()]]`

robcaulk: I'm, not sure the improvement in performance is worth the reduced readability here...we are talking about maybe miliseconds gained, but the proposed code is nearly impossible to come back to or for a new person to understand.

robcaulk:
I am confident in this PR, as I (and others) have extensively tested it and benchmarked the performance gains. 

The main concern is that it technically will require users to start from a new identifier, because it reorganizes the order of the corr_pair features in the training dataframe. 

I do not have a solution for this - and I do not believe it should block its acceptance. Coding anything around this would add unnecessary complexity to the code base just to bridge a small gap for current users.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1126,13 +1126,57 @@ def set_all_pairs(self) -> None:
...
+                                 corr_dataframes: Dict[str, DataFrame],
+                                 current_pair: str) -> DataFrame:
+        """
+        Attach the existing corr_pair dataframes to the current pair dataframe before training
+        :params:
+        :dataframe: current pair strategy dataframe, indicators populated already
+        :corr_dataframes: dictionary of saved dataframes from earlier in the same candle
+        :current_pair: current pair to which we will attach corr pair dataframe
+        :return:
+        :dataframe: current pair dataframe of populated indicators, concatenated with corr_pairs
```
xmatthias: Please use the default docstring formatting as explained in [contributing](https://github.com/freqtrade/freqtrade/blob/develop/CONTRIBUTING.md#additional-styles-applied) - and used throughout the codebase.

freqtrade/freqai/freqai_interface.py
```diff
@@ -680,6 +687,7 @@ def inference_timer(self, do: Literal['start', 'stop'] = 'start', pair: str = ''
                                    " avoid blinding open trades and degrading performance.")
                 self.pair_it = 0
                 self.inference_time = 0
+                self.get_corr_dataframes = True
```
xmatthias: I don't think this is a good place to reset this variable.

This is a timing function - having it do anything other than timing (and performance gathering) seems like an unintended side-effect to me - especially since this seems operation-critical.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1126,13 +1126,57 @@ def set_all_pairs(self) -> None:
...
+                    ready for training
+        """
+        pairs = self.freqai_config["feature_parameters"].get("include_corr_pairlist", [])
+
+        for pair in pairs:
+            if current_pair not in pair:
+                dataframe = pd.concat([dataframe, corr_dataframes[pair]], axis=1)
```
xmatthias: Can you explain this logic?
`pairs` is a list of pairs (e.g. `['BTC/USDT', 'XRP/USDT', ...]`)

do we assume `current_pair` to be `BTC/USDT`? (then why not use == for comparison?)

no matter what the intend here is (or the content of current pair) - using `in` for pair comparisons is quite certainly wrong.
Crypto pairs are often named similarily - you could have `BTC/USD` and `BTC/USDT` - the first is "in" the latter - but that's not the comparison you're looking for.

There's also a pair `T/USDT` - so every coin that ends in `T` would have the condition "true" - which would most certainly have unintended side-effects.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1126,13 +1126,57 @@ def set_all_pairs(self) -> None:
...
+        :dataframe: fully populated dataframe (current pair + corr_pairs)
+        :return:
+        :corr_dataframes: dictionary of dataframes to be attached to other pairs in same candle.
+        """
+        corr_dataframes: Dict[str, DataFrame] = {}
+        pairs = self.freqai_config["feature_parameters"].get("include_corr_pairlist", [])
+
+        for pair in pairs:
+            coin = pair.split('/')[0]
+            pair_cols = [col for col in dataframe.columns if coin in col]
```
xmatthias: what'd be the side-effect if this matches a wrong column (which seems very likely ...)?

Based on the sample in the other comment - if my informative coin is `T/USDT` ([available on binance](https://www.binance.com/en/trade/T_USDT?)) - we'd have every column that contains a capital T (could be `BTC` - could be an indicator that i decided to name `-TRIPPLE` - and therefore would match on every column)

freqtrade/freqai/data_kitchen.py
```diff
@@ -1173,10 +1217,13 @@ def use_strategy_to_populate_indicators(
...
-                    if pair in i:
-                        continue  # dont repeat anything from whitelist
+
+        # ensure corr pairs are always last
+        for i in pairs:
+            if pair in i:
+                continue  # dont repeat anything from whitelist
+            for tf in tfs:
+                if pairs and do_corr_pairs:
                     dataframe = strategy.populate_any_indicators(
```
xmatthias: I don't quite like variable naming here ... it's unclear what `i` is (is it a string? a list of strings? ... - someone not knowing freqAI insideout would have to look up the type in the config ... ) - maybe we can enhance the variables with typehints to help this?

xmatthias (commented):
> The main concern is that it technically will require users to start from a new identifier, because it reorganizes the order of the corr_pair features in the training dataframe.
> 
> I do not have a solution for this - and I do not believe it should block its acceptance. Coding anything around this would add unnecessary complexity to the code base just to bridge a small gap for current users.

I've never said this would be a blocking factor - we just need to be aware of this.

there's however some other concerns about how comparisons on pairs are done ... - see other comments

freqtrade/freqai/data_kitchen.py
robcaulk: the `if current_pair not in pair:` is a check to ensure that if a pair exists simultaneously on the `whitelist` and the `corr_pairlist`, we will not add the same indicators to itself. That is all.

For the point about `T/USDT`, it is a fair point. And the only way around this would be to switch FreqAI away from using `coin = pair.split('/')[0]` in `populate_any_indicators` to simply using the full `pair` string. 

But - that will be a massive breaking change, everyone will need to update their `populate_any_indicator` functions, presumably with the following edit:

```python
coin = pair #.split('/')[0]  <-- removing the split call *should* be enough.  
```

Although small - it still would require everyone to do it.

freqtrade/freqai/data_kitchen.py
robcaulk: While `i` is indeed a non-descriptive iterator name. I don't see this point about someone looking up config types. `i` is just an iterator of `pairs`, which I think is obvious to even new python users. 

We can change the iterator name, but how exactly do you want to type hint it?

xmatthias: You can put `pairs: List[str] = self.config...` where pairs is created, so mypy (and hence also developers using a semi-decent IDE) know what type it is.

robcaulk: Sure, type hinting pairs, no problem. I thought you asked "it is unclear what `i` is"

freqtrade/freqai/data_kitchen.py
xmatthias: `T/USDT` is an extreme sample - i've never heared of it before - and probably will never again unless my bot happens to buy it - but it's as good as any to serve as an example.
ETH/USDT and xETH/USDT (Wrapped ETH iirc) is a similar example - and many other derivatives which have (intentionally or not) a similar name).

in hindsight it's a bit unfortunate that we use `f"%-{coin}rsi-period_{t}"` - not `f"%-{coin}-rsi-period_{t}"` - as that would give us a clear indication as to where the coin name is "finished" and where the indicator name starts.


with indicator naming - we do take a LOT of faith in users to keep the same nomenclature (lowercase, same identical syntax, ...) - which is in the samples - but not documented - and also not really a requirement ...

let's take a diff. example of a problematic "in" condition:
we use  `informative[f"{coin}bb_lowerband-period_{t}"] = bollinger["lower"]` ... but maybe the user deletes (and recreates) this and it becomes `informative[f"{coin}bb_LOWERBAND-period_{t}"] = bollinger["lower"]` (uppercased lowerband). (obvoiusly this could happen with any coin/indicator - as there's too many coins to count - with most possible 3/4/5 letter symbols).

Now we'd have a match on `BAND/USDT` (where the coin is `BAND` - which is a valid, listed coin on binance).
The naming convention only specifies the starting of a column (`%- / `%%-` / `&-`) - not the rest of the name.

It's not a problem as long as the "naming" remains what it is - just a column name.
the moment we attempt to put bigger meaning into the name (like figuring out the pair name) - we might encounter problems - where the naming will no longer be reliable.

---

While I'm not sure if it'll fix all problems, maybe using `%-{coin}` (and all other documented counterparts) as search-term for columns could be an option? so the column has to start with one of these (3?) signs to be identified as "belongs to this coin".

It'll still not fix all problems (`T` / `TRX` -> `%-Tmfi`, `%-TRXmfi`) - but will be slightly safer than `T in column_name` ...

freqtrade/freqai/freqai_interface.py
robcaulk: This is a function which tracks how many pairs have passed. It is all handled internally - it’s not a user controlled function. 

For the present PR we need to know how many pairs have passed. 

So if we don’t use the function that already checks how many pairs have passed, then we need to duplicate that function - and all the code contained within it. 

Tldr this is the exactly right location for this flag to get set.

freqtrade/freqai/data_kitchen.py
xmatthias: well that too `i` is kinda a bad name for anything :laughing: 
(it's actually a pair i think - though it could also be "coin" ... )

freqtrade/freqai/data_kitchen.py
robcaulk: Fixed.

freqtrade/freqai/data_kitchen.py
robcaulk: This is now handled inthe following way:

Users are encouraged via the example strat and docs, to use the full pair string when making their features in `populate_any_indicators()`. Our backend looks for the full pair string in one of the following forms:

```python
valid_strs = [f"%-{pair}", f"%{pair}", f"%_{pair}"]
```

If none of these are matched, but the user has pairs in their `include_corr_pairlist`, then FreqAI will log a warning:

```python
                logger.warning("Couldn't cache corr_pair dataframes for improved performance. "
                               "Consider ensuring that the full coin/stake, e.g. XYZ/USD, "
                               "is included in the column names when you are creating features "
                               "in `populate_any_indicators()`.")
```

And then revert to old behavior.

robcaulk: This is now handled inthe following way:

Users are encouraged via the example strat and docs, to use the full pair string when making their features in `populate_any_indicators()`. Our backend looks for the full pair string in one of the following forms:

```python
valid_strs = [f"%-{pair}", f"%{pair}", f"%_{pair}"]
```

If none of these are matched, but the user has pairs in their `include_corr_pairlist`, then FreqAI will log a warning:

```python
                logger.warning("Couldn't cache corr_pair dataframes for improved performance. "
                               "Consider ensuring that the full coin/stake, e.g. XYZ/USD, "
                               "is included in the column names when you are creating features "
                               "in `populate_any_indicators()`.")
```

And then revert to old behavior.

freqtrade/freqai/data_kitchen.py
robcaulk: This is resolved by typing the relevant lists and changing `i` to `corr_pair`.

freqtrade/freqai/data_kitchen.py
robcaulk: See other comment for explanation on the new string checking method.

docs/freqai-feature-engineering.md
```diff
@@ -2,7 +2,10 @@
...
+Low level feature engineering is performed in the user strategy within a function called `populate_any_indicators()`. That function sets the `base features` such as, `RSI`, `MFI`, `EMA`, `SMA`, time of day, volume, etc. The `base features` can be custom indicators or they can be imported from any technical-analysis library that you can find. One important syntax rule is that all `base features` string names are prepended with `%-{pair}`, while labels/targets are prepended with `&`. 
+
+!!! Note
+    Adding the full pair string, e.g. XYZ/USD, in the feature name enables improved performance for dataframe caching on the backend. If users elect *not* to add the full pair string in the feature string, FreqAI will operate in a reduced performance mode.
```
xmatthias:
```suggestion
    Adding the full pair string, e.g. XYZ/USD, in the feature name enables improved performance for dataframe caching on the backend. If users select *not* to add the full pair string in the feature string, FreqAI will operate in a reduced performance mode.
```

freqtrade/freqai/freqai_interface.py
```diff
@@ -363,10 +364,10 @@ def start_live(
...
         self.model = self.dd.load_data(metadata["pair"], dk)
 
-        with self.analysis_lock:
-            dataframe = self.dk.use_strategy_to_populate_indicators(
-                strategy, prediction_dataframe=dataframe, pair=metadata["pair"]
-            )
+        dataframe = dk.use_strategy_to_populate_indicators(
+            strategy, prediction_dataframe=dataframe, pair=metadata["pair"],
+            do_corr_pairs=self.get_corr_dataframes
+        )
```
xmatthias, line 371: we no longer use  `analysis_lock` here ... does that mean we can/could remove that lock completely? (in my understanding, the problem we tried to solve here was fixed in ta-lib)

or is it still necessary in some other place? (it's used once more in `extract_data_and_train_model()`).

freqtrade/templates/FreqaiExampleStrategy.py
```diff
@@ -63,45 +63,43 @@ def populate_any_indicators(
         :param informative: the dataframe associated with the informative pair
         """
 
-        coin = pair.split('/')[0]
```
xmatthias, line 66: We should do the same improvement to FreqaiExampleHybridStrategy (basically align it to the new behavior)

docs/freqai-feature-engineering.md
robcaulk: In fact, elect is the correct word here :-)

freqtrade/freqai/freqai_interface.py
```diff
@@ -738,6 +744,29 @@ def _set_train_queue(self):
...
+        """
+
+        if self.get_corr_dataframes:
+            self.corr_dataframes = dk.extract_corr_pair_columns_from_populated_indicators(dataframe)
+            if not self.corr_dataframes:
+                logger.warning("Couldn't cache corr_pair dataframes for improved performance. "
+                               "Consider ensuring that the full coin/stake, e.g. XYZ/USD, "
+                               "is included in the column names when you are creating features "
+                               "in `populate_any_indicators()`.")
+            self.get_corr_dataframes = not bool(self.corr_dataframes)
```
xmatthias, line 762: I'm not sure i understand the logic used for `get_corr_dataframes` (when is it expected to be True, when false?)

from my understanding (please correct me where/if i've got it wrong):
* it's True at the beginning of each loop (it's set in "inference.stop" - so it'll be empty afterwards)
* it's it's False if we were able to extract columns from a populated DF (this line)
* it's kept at True if the above was successful
* if it's False - we use cached columns (?) -> else branch in this function

in my understanding, this won't be used by the training loop? do we have to somehow ensure that's not regressing in the future somehow - by adding some additional new functionality or whatnot ... ?

docs/freqai-feature-engineering.md
xmatthias: both would be correct - but i find "select" to be simpler language. 
Always think that a non-native english speaker should understand this, too.

robcaulk: Select is not correct, I am sorry. You are welcome to change it to "choose" or "decide."

freqtrade/freqai/freqai_interface.py
robcaulk: Sure, we can add the analysis lock removal to this PR.

freqtrade/templates/FreqaiExampleStrategy.py
robcaulk: Done.

freqtrade/freqai/freqai_interface.py
robcaulk: If `self.corr_dataframes` is empty, then it will keep `self.get_corr_dataframes` as true in `use_strategy_to_populate_indicators` (`do_corr_pairs = True`). This means that it will always calculate the indicators for the corr_pair dfs for each main pair. This is the behavior of current develop. 

If `self.corr_dataframes` is not empty, it means that `extract_corr_pair_columns_from_populated_idicators` successfully found the corr_pairs in the column names. So it sets `self.get_corr_dataframes` to False, because, it just got them and cached them. All subsequent pairs during the present candle will `use_strategy_to_populate_indicators` with `do_corr_pairs = False` (it wont calculate the indicators).

At the end of the candle, `inference_timer` will set `self.get_corr_dataframes` to true and that means that the first pair to get inferenced during the next candle will calculate the full set of indicators, cache them, and then the cycle continues.

None of this has anything to do with training. This is only for inference.

docs/freqai-feature-engineering.md
xmatthias: `elect` is wrong in my opinion.

As you think my proposal is wrong, Pick any wording you think is appropriate and i'll have a look at that (Please keep language simplicity in mind - for non-english native speakers).

this branch now also seems to have merge conflicts ...

xmatthias (changes requested):

docs/freqai-feature-engineering.md
robcaulk: Ok changed to "decide"

xmatthias (approved):
LGTM :+1:  

let's hope it gives us great performance gains! :100:

PR #7615, opened 2022-10-20
Add price jump warning

### LINKED ISSUE #7407: Backtesting with missing candels and 1000% higher price afterwards
======== SUMMARY METRICS ======
| Metric                      | Value                      |
|-----------------+--------------|
| Starting balance          | 1000 BNB           |
| Final balance               | 8859.982 BNB    |
| Total profit %              | 786.00%             |

So much profit.... It's because of missing candles

image

**How can I avoid this problem in backtesting?** 
If I have an open position, the candels are missing, afterwards the price is 1000% higher than before.

xmatthias:
## Summary

Add price jump warning - alerting people about potential unrealistic gains.

Closes #7407

## Quick changelog

- price jump warning

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -102,6 +102,12 @@ If this happens for all pairs in the pairlist, this might indicate a recent exch
 
 Irrespectively of the reason, Freqtrade will fill up these candles with "empty" candles, where open, high, low and close are set to the previous candle close - and volume is empty. In a chart, this will look like a `_` - and is aligned with how exchanges usually represent 0 volume candles.
 
+### I'm getting "Price jump between 2 candles detected"
+
+This message is a warning that the candles had a price jump of > 30%.
+This might be a sign that the pair stopped trading, and some token exchange took place (e.g. COCOS in 2021 - where price jumped from 0.0000154 to 0.01621).
+This message is often accompanied by ["Missing data fillup"](#im-getting-missing-data-fillup-messages-in-the-log) - as trading on such pairs is often stopped for some time.
+
 ### I'm getting "Outdated history for pair xxx" in the log
 
 The bot is trying to tell you that it got an outdated last candle (not the last complete candle).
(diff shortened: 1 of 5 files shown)


robcaulk (commented):
Just curious, but are we adding unnecessary overhead with this? We expect this to be hit 0.01% of the time, or less. People trading established pairs wouldn't need it at all, right?

xmatthias:
> Just curious, but are we adding unnecessary overhead with this? We expect this to be hit 0.01% of the time, or less. People trading established pairs wouldn't need it at all, right?

"established pair" is a broad term ...

also an established pair could have a token swap - think SHIB - it's kinda been around a while now - but it's surely possible that at some point, they decide "we have too many 0 in our number - so let's exchange all tokens at 1000:1 - which would effectively boost price by 1000%.
freqtrade won't have this info - so would assume that that pair just spiked 1000%

it's an edgecase, sure - but one nobody will be checking for while backtesting, while it'll greatly exaggerate wins (though you should notice if 1 pair has a 1000% trade).

I'm not against ignoring this request entirely (closing this and the corresponding issue as won't fix) - as it's an edge-case that WILL show up in backtesting as "very high fluke" the moment you look at any other metric than "total profit"

---

it's also not "really" relevant for live - so maybe moving this to the datahandler (so it only runs on data we load from disk) would make more sense?

robcaulk (approved):
LGTM

popperwin:
Hello!
Thanks for this feature, it can be useful. 
However I was backtesting on ETH/USDT on binance for year 2022, and I had a price jumps warning without any "Missing data fillup".

I looked at the code and it seems that the condition `warn_price`  in the function `_check_empty_df` doesn't apply the timerange selected; so the warning is triggered even if the price gap is not in the timerange of the backtesting.  The warning can be very misleading in that cases.
 
In my case the drop were between  `2017-08-22 03:55:00`  to `2017-08-22 04:00:00`, but my backtest timerange is from `2022-01-01` to `2022-12-31`

xmatthias:
nice spot - will be updated shortly.

PR #7618, opened 2022-10-22
Update function in FreqAI interface to record FreqAI config params

wizrds:
## Summary

~~Supports users with multiple config files being copied to the models directory.~~ 

This now only makes a record of certain config parameters in a `run_params.json` file. The parameters recorded are:
`freqai, timeframe, stake_amount, stake_currency, max_open_trades, pairs`.

## Quick changelog

- ~~Record all configs instead of just the first~~
- ~~Changed from `shutil.copy` to `shutil.copyfile`~~
- Serialize certain config params to a `run_params.json` file

## What's new?

~~Currently, the function `set_full_path` records the configs for reproducibility. However it assumes only one config is used and copies the first one. This PR changes the behavior to include **ALL** configs. Also, when ran in a Docker Swarm setting with the config added to swarm, it is mounted in the container as read only. When copied over with `shutil.copy` it includes permission bits, causing the next run to fail with a permission error. This is fixed by using `shutil.copyfile`.~~

Since we create the file completely ourselves there are no permission issues.

### DIFF
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -531,7 +531,7 @@ class IFreqaiModel(ABC):
         Creates and sets the full path for the identifier
         """
         self.full_path = Path(
-            self.config["user_data_dir"] / "models" / f"{self.freqai_info['identifier']}"
+            self.config["user_data_dir"] / "models" / f"{self.identifier}"
         )
         self.full_path.mkdir(parents=True, exist_ok=True)
 
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -61,15 +61,16 @@ class IFreqaiModel(ABC):
             "data_split_parameters", {})
         self.model_training_parameters: Dict[str, Any] = config.get("freqai", {}).get(
             "model_training_parameters", {})
+        self.identifier: str = self.freqai_info.get("identifier", "no_id_provided")
         self.retrain = False
         self.first = True
         self.set_full_path()
+        self.record_configs()
         self.follow_mode: bool = self.freqai_info.get("follow_mode", False)
         self.save_backtest_models: bool = self.freqai_info.get("save_backtest_models", True)
         if self.save_backtest_models:
             logger.info('Backtesting module configured to save all models.')
         self.dd = FreqaiDataDrawer(Path(self.full_path), self.config, self.follow_mode)
-        self.identifier: str = self.freqai_info.get("identifier", "no_id_provided")
         self.scanning = False
         self.ft_params = self.freqai_info["feature_parameters"]
         self.keras: bool = self.freqai_info.get("keras", False)
@@ -526,14 +527,24 @@ class IFreqaiModel(ABC):
         return file_exists
 
     def set_full_path(self) -> None:
+        """
+        Creates and sets the full path for the identifier
+        """
         self.full_path = Path(
             self.config["user_data_dir"] / "models" / f"{self.freqai_info['identifier']}"
         )
         self.full_path.mkdir(parents=True, exist_ok=True)
-        shutil.copy(
-            self.config["config_files"][0],
-            Path(self.full_path, Path(self.config["config_files"][0]).name),
-        )
+
+    def record_configs(self) -> None:
+        """
+        Records configs in the full path for reproducibility
+        """
+        self.config_record_path = self.full_path / Path("configs")
+        self.config_record_path.mkdir(parents=True, exist_ok=True)
+
+        for config in self.config["config_files"]:
+            dest_config_path = self.config_record_path / Path(config).name
+            shutil.copyfile(config, dest_config_path)
 
     def extract_data_and_train_model(
         self,
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import threading
 import time
@@ -21,7 +20,7 @@ from freqtrade.exceptions import OperationalException
 from freqtrade.exchange import timeframe_to_seconds
 from freqtrade.freqai.data_drawer import FreqaiDataDrawer
 from freqtrade.freqai.data_kitchen import FreqaiDataKitchen
-from freqtrade.freqai.utils import plot_feature_importance
+from freqtrade.freqai.utils import plot_feature_importance, record_params
 from freqtrade.strategy.interface import IStrategy
 
 
@@ -97,7 +96,7 @@ class IFreqaiModel(ABC):
         self._threads: List[threading.Thread] = []
         self._stop_event = threading.Event()
 
-        self.record_params()
+        record_params(config, self.full_path)
 
     def __getstate__(self):
         """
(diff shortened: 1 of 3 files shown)


freqtrade/freqai/freqai_interface.py
```diff
@@ -526,14 +527,24 @@ def model_exists(self, dk: FreqaiDataKitchen) -> bool:
...
+    def record_configs(self) -> None:
+        """
+        Records configs in the full path for reproducibility
+        """
+        self.config_record_path = self.full_path / Path("configs")
+        self.config_record_path.mkdir(parents=True, exist_ok=True)
+
+        for config in self.config["config_files"]:
+            dest_config_path = self.config_record_path / Path(config).name
+            shutil.copyfile(config, dest_config_path)
```
xmatthias: I'm a bit concerned about this (not this PR per se, but the whole functionality of "copy config someplace else").
If we assume a live configuration, we're effectively copy/pasting people's exchange keys to random places (within the user directory).
People may take extra precautions (permissions, ..., for example) for their keys - which we now basically void by copying the config to other places.

@robcaulk can you remind me of the purpose of doing a copy of the config files? (why do we need a copy of the config file in the model directory) ?

tests/freqai/test_freqai_interface.py
```diff
@@ -19,7 +19,7 @@
 
 def is_arm() -> bool:
     machine = platform.machine()
-    return "arm" in machine or "aarch64" in machine
+    return "arm" in machine or "aarch74" in machine
```
xmatthias: well this clearly has to be changed back - the architecture on an ARM is aarch64 - not aarch74.

xmatthias (changes requested):

tests/freqai/test_freqai_interface.py
wizrds: Whoops didn't mean to do that

freqtrade/freqai/freqai_interface.py
robcaulk: You make a good point. 

The reason we copy the config over is because it is a record of the run. This enables improved reproducibility, ease of parameter exploration (easy to see which parameters were associated with a set of results).

we could parse it to save everything except the sensitive key information?

wizrds: Would it make more sense to only save certain information such as the Freqai config, timeframe, etc?

xmatthias: if reproducibiliy is the goal, then i think copying json files is the wrong approach - neglecting most possibilities freqtrade offers to configure things (you're also omitting the strategy - which is also a big part of reproducing (or not) a result.

i can see several ways where the above simply won't work
*  `-c config.json -c secret/config.json -c ../someother/config.json` (both names are identical ...) - this could go on with 20 directories (we would store one config.json - but not all of them)
* `export FREQTRADE__FREQAI__TRAIN_PERIOD_DAYS=20` (while the config defines this as 15) - so your "reproducibility" would show a value of 15 - but the bot runs with 20 (obviously this is possible for every value)
* `--timeframe 1m` (config / strategy defines the timeframe as 5m - but the runtime argument overwrites this
* config not defining timeframe, but the strategy defines the timeframe (that's how i run all my bots) - so you'd not capture the actual timeframe used (this is obviously true for many parameters - but timeframe is the most relevant one i think

the "running" config (basically the `config` object we pass around throughout the bot) will contain the above info - compiled in the right sequence - so that's the only "single" source that actually REALLY defines what's being used by the bot.

Now simply serializing this will be difficult - as it'll contain 
* sensitive information
* non-serializable objects like paths

While we can identify the first pretty easily - i think it's better to pick what we really need (what's actually relevant to reproducibility). That'll avoid potentially leaking new sensitive information (new keys) - and will be simpler than to identify the non-json objects we keep in config (like all the Path objects).

robcaulk: I understand Freqtrade is perhaps too flexible on that front of parameter setting. I’m not aiming to solve that “issue” anytime soon. I think you are a bit gun shy here. You forgot to point out that the entire source code version is also changeable underneath ;). 

The point is to just have a record of parameters stored with the results/models. This includes pairlist, pairlist filter, stake size etc.

Sure copy the strat over too, that’s a good idea. I had it in earlier versions of FreqAI, im happy to bring it back.

The solution here is to simply copy the full config without any secret-containing sub dictionaries. Am I missing something?

xmatthias: > I think you are a bit gun shy here

I'm not sure i understand your point (especially, i don't understand the point of getting personal - after i pointed out that a PR is not fulfilling the goal you're having).

I'm pointing out flaws in the current logic - which is NOT fulfilling your need of having "a record of parameters used" (at least not reliably).
The new logic (copy all additional configs) will make it worse - as you might end up overwriting the most important config in case of naming collisions.

It may work for you personally by copying the configuration - but that's not the case for most users (it's rare to see a configuration containing timeframe - that's usually coming from the strategy).
That's different with the freqAI examples, but they're really the exception (should probably be changed to align behavior here) - the recommendation is to have the timeframe in the strategy - not in the configuration - as most often, the strategy will only work for certain timeframes.

> The solution here is to simply copy the full config without any secret-containing sub dictionaries. Am I missing something?

If that's the solution - then the _current_ behavior is correct - and the new behavior (what this PR is introducing) would revert this by including all files.

robcaulk: It seems the best solution is to simply wait for full initialization of the `IFreqaiModel` object, and then grab the `freqai` relevant parameters from the configuration dictionary:

`freqai, timeframe, stake_amount, max_open_trades, pairs`

Now we no longer need to save any particular config file, and we avoid copying secrets unnecessarily.

wizrds: @xmatthias 
I'm not sure what you mean by the current way is correct, as it just grabs the first config file and copies that. Which could be anything, the full config, your telegram config file, etc.

xmatthias: > I'm not sure what you mean by the current way is correct

That's a direct answer to the sentence quoted - which says "The solution here is to simply copy the full config without any secret-containing sub dictionaries.". 
if this would be the aim (it isn't!) - then current behavior would be correct (it's based on the assumption that secrets are in subdirectories - which is not the case).

> wait for full initialization of the `IFreqaiModel` object

do we need to wait for this, or just until the configuration is "complete"?
basically - is freqAI changing settings in the configuration if it determines some invalid state, or is it just warning / failing on these conditions?

robcaulk: IFreqaiModel has some automatic "reorientation" to help ensure users aren't getting themselves into trouble. So it needs to wait until initialization finishes on the IFreqaiModel object.

robcaulk (approved):
LGTM

freqtrade/freqai/utils.py
```diff
@@ -191,3 +193,28 @@ def add_feature_trace(fig, fi_df, col):
...
+    }
+
+    with open(params_record_path, "w") as handle:
+        rapidjson.dump(run_params, handle, indent=4, default=np_encoder,
+                       number_mode=rapidjson.NM_NATIVE)
+
+
+def np_encoder(self, object):
+    if isinstance(object, np.generic):
+        return object.item()
```
xmatthias: I don't think the config object will contain numpy objects (if it does, that's clearly not expected and i'd be curious which key that was) ... so i'm not sure why we need that (i think using "str" as default for dump should work as well).

freqtrade/freqai/utils.py
```diff
@@ -191,3 +193,28 @@ def add_feature_trace(fig, fi_df, col):
...
+        "pairs": config.get('exchange', {}).get('pair_whitelist')
+    }
+
+    with open(params_record_path, "w") as handle:
+        rapidjson.dump(
+            run_params,
+            handle,
+            indent=4,
+            default=str,
+            number_mode=rapidjson.NM_NATIVE
```
xmatthias: i think you will want to use `rapidjson.NM_NATIVE | rapidjson.NM_NAN` as number_mode.

`max_open_trades` will be `inf` if set to -1 in the configuration... 
while it's a non-common config nowadays - it's still a supported mode (static stake, but increasing number of trades until you run out of money).


Other than this - i think this is good.
we do have a check in place that verifies the file is written/copied, right?

wizrds: Do you mean check if the path exists at the end of the record function?

xmatthias: well basically yes.

from log coverage, i can see the function is called ...  (actually by almost every freqai test - which is expected).

a simple test (simply removing the write from the function) however shows that it's actually irrelevant what this function does - as long as it doesn't error ... 

i think to protect ourselfs from future regressions, it might be good to do a very simple assert in one of the tests if "has this file been written" ...

xmatthias (approved):
LGTM :+1:

PR #7621, opened 2022-10-23
Improved WebSocketChannel API

wizrds:
## Summary

Improved WebSocketChannel API with a more unified approach without needing throttling outside of the mechanism.

## Quick changelog

- Update sending via the WebSocketChannel and ChannelManager to manage throttling

## What's new?

This change allows a more readable and sensible solution to sending messages via a WebSocketChannel. Right now, we must add sleep calls throughout the code to throttle how fast we send messages via a channel. With this PR, the channel has better management of large pairlists by waiting a certain amount of time for the relay function to drain it's outgoing messages, effectively throttling how fast it sends messages in one place.

### DIFF
--- a/freqtrade/rpc/api_server/api_ws.py
+++ b/freqtrade/rpc/api_server/api_ws.py
@@ -75,7 +75,7 @@ async def _process_consumer_request(
         # Format response
         response = WSWhitelistMessage(data=whitelist)
         # Send it back
-        await channel_manager.send_direct(channel, response)
+        await channel_manager.send_direct(channel, response.dict(exclude_none=True))
 
     elif type == RPCRequestType.ANALYZED_DF:
         limit = None
@@ -90,7 +90,7 @@ async def _process_consumer_request(
         # For every dataframe, send as a separate message
         for _, message in analyzed_df.items():
             response = WSAnalyzedDFMessage(data=message)
-            await channel_manager.send_direct(channel, response)
+        await channel_manager.send_direct(channel, response.dict(exclude_none=True))
 
 
 @router.websocket("/message/ws")
--- a/freqtrade/rpc/api_server/api_ws.py
+++ b/freqtrade/rpc/api_server/api_ws.py
@@ -90,7 +90,7 @@ async def _process_consumer_request(
         # For every dataframe, send as a separate message
         for _, message in analyzed_df.items():
             response = WSAnalyzedDFMessage(data=message)
-        await channel_manager.send_direct(channel, response.dict(exclude_none=True))
+            await channel_manager.send_direct(channel, response.dict(exclude_none=True))
 
 
 @router.websocket("/message/ws")
--- a/freqtrade/rpc/api_server/api_ws.py
+++ b/freqtrade/rpc/api_server/api_ws.py
@@ -1,4 +1,3 @@
-import asyncio
 import logging
 from typing import Any, Dict
 
@@ -11,6 +10,7 @@ from freqtrade.enums import RPCMessageType, RPCRequestType
 from freqtrade.rpc.api_server.api_auth import validate_ws_token
 from freqtrade.rpc.api_server.deps import get_channel_manager, get_rpc
 from freqtrade.rpc.api_server.ws import WebSocketChannel
+from freqtrade.rpc.api_server.ws.channel import ChannelManager
 from freqtrade.rpc.api_server.ws_schemas import (WSAnalyzedDFMessage, WSMessageSchema,
                                                  WSRequestSchema, WSWhitelistMessage)
 from freqtrade.rpc.rpc import RPC
@@ -37,7 +37,8 @@ async def is_websocket_alive(ws: WebSocket) -> bool:
 async def _process_consumer_request(
     request: Dict[str, Any],
     channel: WebSocketChannel,
-    rpc: RPC
+    rpc: RPC,
+    channel_manager: ChannelManager
 ):
     """
     Validate and handle a request from a websocket consumer
@@ -72,9 +73,9 @@ async def _process_consumer_request(
         whitelist = rpc._ws_request_whitelist()
 
         # Format response
-        response = WSWhitelistMessage(data=whitelist)
+        response = WSWhitelistMessage(data=whitelist).dict(exclude_none=True)
         # Send it back
-        await channel.send(response.dict(exclude_none=True))
+        await channel_manager.send_direct(channel, response)
 
     elif type == RPCRequestType.ANALYZED_DF:
         limit = None
@@ -88,10 +89,8 @@ async def _process_consumer_request(
 
         # For every dataframe, send as a separate message
         for _, message in analyzed_df.items():
-            response = WSAnalyzedDFMessage(data=message)
-            await channel.send(response.dict(exclude_none=True))
-            # Throttle the messages to 50/s
-            await asyncio.sleep(0.02)
+            response = WSAnalyzedDFMessage(data=message).dict(exclude_none=True)
+            await channel_manager.send_direct(channel, response)
 
 
 @router.websocket("/message/ws")
@@ -116,7 +115,7 @@ async def message_endpoint(
                     request = await channel.recv()
 
                     # Process the request here
-                    await _process_consumer_request(request, channel, rpc)
+                    await _process_consumer_request(request, channel, rpc, channel_manager)
 
             except (WebSocketDisconnect, WebSocketException):
                 # Handle client disconnects
--- a/freqtrade/rpc/api_server/api_ws.py
+++ b/freqtrade/rpc/api_server/api_ws.py
@@ -73,7 +73,7 @@ async def _process_consumer_request(
         whitelist = rpc._ws_request_whitelist()
 
         # Format response
-        response = WSWhitelistMessage(data=whitelist).dict(exclude_none=True)
+        response = WSWhitelistMessage(data=whitelist)
         # Send it back
         await channel_manager.send_direct(channel, response)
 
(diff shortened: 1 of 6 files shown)


freqtrade/rpc/external_message_consumer.py
```diff
@@ -270,6 +270,11 @@ async def _receive_messages(
                     logger.debug(f"Connection to {channel} still alive...")
 
                     continue
+                except (websockets.exceptions.ConnectionClosed):
+                    # Just eat the error and continue reconnecting
+                    logger.warning(f"Disconnection in {channel} - retrying in {self.sleep_time}s")
+                    await asyncio.sleep(self.sleep_time)
```
xmatthias, line 276: are you sure this will actually try to reconnect?
we're not "really" calling `.connect()` again - only `wait_for()` ...

freqtrade/rpc/api_server/ws/proxy.py
```diff
@@ -15,6 +15,10 @@ class WebSocketProxy:
     def __init__(self, websocket: WebSocketType):
         self._websocket: Union[FastAPIWebSocket, WebSocket] = websocket
 
+    @property
+    def raw(self):
```
xmatthias: i find `.raw` quite generic, and it's completely unclear that it's going to return the websocket object - maybe `.raw_websocket` or something might be better...

freqtrade/rpc/api_server/api_ws.py
```diff
@@ -88,10 +89,8 @@ async def _process_consumer_request(
 
         # For every dataframe, send as a separate message
         for _, message in analyzed_df.items():
-            response = WSAnalyzedDFMessage(data=message)
-            await channel.send(response.dict(exclude_none=True))
-            # Throttle the messages to 50/s
-            await asyncio.sleep(0.02)
+            response = WSAnalyzedDFMessage(data=message).dict(exclude_none=True)
+            await channel_manager.send_direct(channel, response)
```
xmatthias: Is there a downside to moving `.dict()` into the `send_direct` method? 

That'd obviously assume we can type-hint `send_direct()` with `WSMessageSchema` - but we shouldn't really be sending different messages anyway other than Ping anyway - so i think all messages SHOULD be typed already (i'm not sure about broadcast ...)

i don't really like the fact that "response" is now no longer typed - as it can introduce bad practices once we add additional message types

freqtrade/rpc/api_server/ws/channel.py
```diff
@@ -170,9 +189,8 @@ async def disconnect_all(self):
         Disconnect all Channels
         """
         with self._lock:
-            for websocket, channel in self.channels.copy().items():
-                if not channel.is_closed():
-                    await channel.close()
+            for websocket in self.channels.copy().keys():
+                await self.on_disconnect(websocket)
 
             self.channels = dict()
```
xmatthias: i think this shouldn't be done (setting channels to an empty dict here).

assuming one of the channels issues an exception on close (maybe "closed already" due to some timing issue) - then we'd "lose" all other channels - and never try to disconnect from them again.

obviously our assumption is that the channels dict is empty after this method - but if it's not for whatever reason (maybe a parallel reconnect) - this could become dangerous.

freqtrade/rpc/api_server/ws/channel.py
```diff
@@ -24,6 +24,8 @@ def __init__(
         self,
         websocket: WebSocketType,
         channel_id: Optional[str] = None,
+        drain_timeout: int = 3,
+        throttle: float = 0.01,
```
xmatthias, line 29: I'm not sure i like these as arguments.
It makes you think "configurability per channel" - but it really isn't.

for a constant, i think setting them directly in __init__ would be better.

freqtrade/rpc/api_server/ws/channel.py
```diff
@@ -160,6 +178,7 @@ async def on_disconnect(self, websocket: WebSocketType):
         with self._lock:
             channel = self.channels.get(websocket)
             if channel:
+                logger.info(f"Disconnecting channel {channel}")
```
xmatthias, line 182: intentional on info? (not a problem necessarily - i guess a disconnect won't happen this often?)

xmatthias (commented):
I definitely like the current approach better.
It'd be great if we could also improve typehints a bit - especially the `ChannelManager` 's last 2 methods seem to take whatever we give them ... but we actually know what the parameters are.

freqtrade/rpc/external_message_consumer.py
wizrds: Yes it works just like the ping error does, by breaking out of that loop and continuing the outer loop on `_create_connection`.

freqtrade/rpc/api_server/ws/proxy.py
wizrds: Understood, I had the same thought originally.

freqtrade/rpc/api_server/ws/channel.py
wizrds: Yes this was intentional

freqtrade/rpc/api_server/api_ws.py
wizrds: I don't necessarily see a downside. I agree, I will make these changes.

freqtrade/rpc/api_server/ws/channel.py
wizrds: This is really no different than the `serializer_cls` parameter as well. I don't know about being configurable right now, but potentially in the future it's a possibility. What do you think?

freqtrade/rpc/api_server/ws/channel.py
```diff
@@ -170,36 +190,29 @@ async def disconnect_all(self):
...
-            message_type = data.get('type')
-            for websocket, channel in self.channels.copy().items():
-                if channel.subscribed_to(message_type):
-                    if not channel.queue.full():
-                        await channel.send(data)
-                    else:
-                        logger.info(f"Channel {channel} is too far behind, disconnecting")
-                        await self.on_disconnect(websocket)
+            for channel in self.channels.copy().values():
+                if channel.subscribed_to(message.type):
```
xmatthias: hm - seems like the type we pass in here ain't of type `WSMessageSchema` - but of type dict.

the call to `.broadcast` isn't typed - so mypy won't help us on that ...

```
2022-10-24 19:54:03,749 - freqtrade.rpc.api_server.webserver - ERROR - Exception happened in background task: 'dict' object has no attribute 'type'
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/rpc/api_server/webserver.py", line 200, in _broadcast_queue_data
    await self._ws_channel_manager.broadcast(message)
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/rpc/api_server/ws/channel.py", line 204, in broadcast
    if channel.subscribed_to(message.type):
AttributeError: 'dict' object has no attribute 'type'

```

wizrds: I have no idea how this slipped past me. I did type `broadcast` so really it should've caught this.

xmatthias: it's the call to broadcast that's not typed (the queue has no type - so it's implicitly returning ANY)

freqtrade/rpc/api_server/webserver.py
```diff
@@ -127,10 +128,10 @@ def shutdown(cls):
         cls._has_rpc = False
         cls._rpc = None
 
-    def send_msg(self, msg: Dict[str, str]) -> None:
+    def send_msg(self, msg: Dict[str, Any]) -> None:
         if self._ws_queue:
             sync_q = self._ws_queue.sync_q
-            sync_q.put(msg)
+            sync_q.put(WSMessageSchema(**msg))
```
xmatthias: we assume this is cheap, right?

I'll run some more tests with this - as i don't trust pydantic to be fast in all instances (especially if the object becomes bigger)... but i assume it'll be ok.

wizrds: I believe it's either this or `parse_obj` and iirc they both are the same except `parse_obj` will ignore unknown fields

xmatthias: yes, the "how" is not the point, if any, it's the "if"...
the alternative i see would be to revert to prior behavior, passing in a dict to the send_direct message (not a WSMessageSchema object) - as i've clearly missed that we're currently not doing this transformation for broadcasts.

wizrds: I'm not sure I follow, this line is the transformation of dict to WSMessageSchema for broadcast

xmatthias: I've decided to revert this to be a dict. It'll avoid type-checking outgoing messages, where we know (or at least assume we know) what the content is. As we expect this to potentially be called 200 (once per pair) times per iteration (worst case) - this can have a noticable impact.

freqtrade/rpc/api_server/api_ws.py
```diff
@@ -90,7 +90,7 @@ async def _process_consumer_request(
         # For every dataframe, send as a separate message
         for _, message in analyzed_df.items():
             response = WSAnalyzedDFMessage(data=message)
-        await channel_manager.send_direct(channel, response.dict(exclude_none=True))
+            await channel_manager.send_direct(channel, response.dict(exclude_none=True))
```
xmatthias, line 93: @wizrds thanks for spotting this! :+1:

xmatthias (approved):
Nice improvement!
LGTM :+1:

PR #7642, opened 2022-10-25
Update freqai-running.md

smarmau:
Updated the follower section of the docs to include parameters to make the config schema validate on start up for follow_mode


robcaulk (approved):
LGTM

PR #7644, opened 2022-10-25
Support for multi target multi-class classification (FreqaiMultiOutputRegressor for Classification)

markdregan:
## Summary

This PR extends the FreqaiMultiOutputRegressor template for multi target multi-class classification.

## Quick changelog

- FreqaiMultiOutputClassifier created based on FreqaiMultiOutputRegressor
- Added `predict` and `predict_proba` to FreqaiMultiOutputClassifier to address squeeze issue
- CatboostClassifierMultiTarget created from CatboostRegressorMultiTarget
- Added MultiTargetClassifierTestStrategy and config_test.json temporarily for testing

## What's new?

Ability to build freqai strategies that demand multi target classifiers (optionally multi-class and multi-target).

Under hood sklearn MultiOutputClassifier is enhanced in same way FreqaiMultiOutputRegressor - allowing user to pass model specific params to the underlying models within MultiOutputClassifier. This enables per model evaluation for Catboost. 

## TODO:
- [ ] Fix parallel execution with `multitarget_parallel_training` = True. Currently not working.
- [x] More testing across num classes and num targets
- [x] Test simple multi class multi target strategy


markdregan:
@robcaulk - I have a draft PR for multi-target / multi-class strat working. However it is not working when `multitarget_parallel_training` is True. I'm getting the error:

```freqtrade.freqai.freqai_interface - WARNING - Training 1INCH/USDT raised exception PicklingError. Message: Could not pickle the task to send it to the workers., skipping.```

I'll look into more later in week. If this twigs anything with you LMK.

**Comment by robcaulk** [comment]
> @robcaulk - I have a draft PR for multi-target / multi-class strat working. However it is not working when `multitarget_parallel_training` is True. I'm getting the error:
> 
> `freqtrade.freqai.freqai_interface - WARNING - Training 1INCH/USDT raised exception PicklingError. Message: Could not pickle the task to send it to the workers., skipping.`
> 
> I'll look into more later in week. If this twigs anything with you LMK.

Very nice. One thing you can change to get more information is to edit the logger line (213 freqai_interface.py) for that in `start_scanning`:

```python
                except Exception as msg:
                    logger.warning(f"Training {pair} raised exception {msg.__class__.__name__}. "
                                   f"Message: {msg}, skipping.")
```

```python
                except Exception as msg:
                    logger.exception(f"Training {pair} raised exception {msg.__class__.__name__}. "
                                   f"Message: {msg}, skipping.")
```


Otherwise, I'll try to test it when I get a chance in the coming days!

user_data/strategies/MultiTargetClassifierTestStrategy.py
```diff
@@ -0,0 +1,244 @@
+import logging
```
xmatthias: we shouldn't be having files in user_data. If a user happens to have a file with the same name, this would cause an update conflict,  many users inexperienced with git will not be able to fix on their own.

The name suggests that it's a test strategy (used for unittests??) - so `tests/strategy/strats/` might be an appropriate location ...

markdregan: I'll remove these. Was adding to PR so testing of a dummy strategy would be easier by the reviewer. But this makes sense.

markdregan:
I'm also experiencing the same pickle error when using `FreqaiMultiOutputRegressor` and `multitarget_parallel_training`. Including the error below from the `FreqaiMultiOutputRegressor` example. LMK if this error flags any root cause with you @robcaulk. 

```
2022-10-27 11:27:21,425 - freqtrade.freqai.freqai_interface - ERROR - Training ALGO/USDT raised exception PicklingError. Message: Could not pickle the task to send it to the workers., skipping.
joblib.externals.loky.process_executor._RemoteTraceback: 
"""
  File "/Users/mregan/Dev/litmus-04/freqtrade/.env/lib/python3.10/site-packages/joblib/externals/loky/backend/queues.py", line 125, in _feed
    obj_ = dumps(obj, reducers=reducers)
  File "/Users/mregan/Dev/litmus-04/freqtrade/.env/lib/python3.10/site-packages/joblib/externals/loky/backend/reduction.py", line 211, in dumps
    dump(obj, buf, reducers=reducers, protocol=protocol)
  File "/Users/mregan/Dev/litmus-04/freqtrade/.env/lib/python3.10/site-packages/joblib/externals/loky/backend/reduction.py", line 204, in dump
    _LokyPickler(file, reducers=reducers, protocol=protocol).dump(obj)
  File "/Users/mregan/Dev/litmus-04/freqtrade/.env/lib/python3.10/site-packages/joblib/externals/cloudpickle/cloudpickle_fast.py", line 632, in dump
    return Pickler.dump(self, obj)
  File "stringsource", line 2, in _catboost._PoolBase.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__
"""

The above exception was the direct cause of the following exception:

  File "/Users/mregan/Dev/litmus-04/freqtrade/freqtrade/freqai/freqai_interface.py", line 209, in _start_scanning
    self.extract_data_and_train_model(
  File "/Users/mregan/Dev/litmus-04/freqtrade/freqtrade/freqai/freqai_interface.py", line 573, in extract_data_and_train_model
    model = self.train(unfiltered_dataframe, pair, dk)
  File "/Users/mregan/Dev/litmus-04/freqtrade/freqtrade/freqai/base_models/BaseRegressionModel.py", line 66, in train
    model = self.fit(data_dictionary, dk)
  File "/Users/mregan/Dev/litmus-04/freqtrade/freqtrade/freqai/prediction_models/CatboostRegressorMultiTarget.py", line 71, in fit
    model.fit(X=X, y=y, sample_weight=sample_weight, fit_params=fit_params)
  File "/Users/mregan/Dev/litmus-04/freqtrade/freqtrade/freqai/base_models/FreqaiMultiOutputRegressor.py", line 52, in fit
    self.estimators_ = Parallel(n_jobs=self.n_jobs)(
  File "/Users/mregan/Dev/litmus-04/freqtrade/.env/lib/python3.10/site-packages/joblib/parallel.py", line 1098, in __call__
    self.retrieve()
  File "/Users/mregan/Dev/litmus-04/freqtrade/.env/lib/python3.10/site-packages/joblib/parallel.py", line 975, in retrieve
    self._output.extend(job.get(timeout=self.timeout))
  File "/Users/mregan/Dev/litmus-04/freqtrade/.env/lib/python3.10/site-packages/joblib/_parallel_backends.py", line 567, in wrap_future_result
    return future.result(timeout=timeout)
  File "/opt/homebrew/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "/opt/homebrew/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
_pickle.PicklingError: Could not pickle the task to send it to the workers.
```

xmatthias:
not immediately - however from experience with pickling problems (hyperopt . . . ) i can tell you that pickling complex objects is no easy task.

You basically cannot have anything with a handle open (files, sockets, ...).
to unpickle, the object also need to be resolvable (for example with `freqtrade.freqai.someclass`).

Joblib has a nice function for this last part (`wrap_non_picklable_objects()`) which we (have to) use for hyperopt ... so you may want to try to wrap the content of your `delayed` function into this (look at hyperopt for a precise example on that).

robcaulk:
@markdregan I don’t quite understand why you are modifying that twice. We already have the parallelized training across models in the `n_jobs` parameter:

https://github.com/freqtrade/freqtrade/blob/54c7122cc3d936d53c15465e8681452dd63afb4a/freqtrade/freqai/base_models/FreqaiMultiOutputRegressor.py#L52

freqtrade/freqai/prediction_models/CatboostClassifierMultiTarget.py
```diff
@@ -0,0 +1,74 @@
...
+        fit_params = []
+        for i in range(len(eval_sets)):
+            fit_params.append({
+                'eval_set': eval_sets[i], 'init_model': init_models[i],
+                'log_cout': sys.stdout, 'log_cerr': sys.stderr,
+            })
+
+        model = FreqaiMultiOutputClassifier(estimator=cbc)
+        thread_training = self.freqai_info.get('multitarget_parallel_training', False)
+        if thread_training:
```
robcaulk, line 70: This is trying to do something that is already handled in the base class.

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
```diff
@@ -0,0 +1,127 @@
...
+        -------
+        y : {array-like, sparse matrix} of shape (n_samples, n_outputs)
+            Multi-output targets predicted across multiple predictors.
+            Note: Separate models are generated for each predictor.
+        """
+        check_is_fitted(self)
+        if not hasattr(self.estimators_[0], "predict"):
+            raise ValueError("The base estimator should implement a predict method")
+
+        y = Parallel(n_jobs=self.n_jobs)(
```
robcaulk: This is new, we don’t parallelize prediction in the current regressor (although I think it’s a great idea to do so), but I’d start by matching current regressor functionality to help you debug this mult threading issue - then you can introduce new functions like this after you are confident they aren’t the root of your multi threading problem.

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if sample_weight is not None and not has_fit_parameter(
+            self.estimator, "sample_weight"
+        ):
+            raise ValueError("Underlying estimator does not support sample weights.")
+
+        if not fit_params:
+            fit_params = [None] * y.shape[1]
+
+        self.estimators_ = Parallel(n_jobs=self.n_jobs)(
```
robcaulk, line 60: I’d let the base class logic handle this for you. 

Another point here is that this doesn’t necessarily increase time to trained models. It’s adding overhead on Al already highly parallelizable algorithm. Many beta testers ran tests using n_jobs or not, they found no legitimate speed-up. It could be due to using small processors (<=24 cores).


I think the only real usecase for n_jobs is with multiple GPUs installed in the system. Then you’d see tremendous benefit. But this is currently a rare configuration in Freqtrade community, so I wouldn’t prioritize this over the principal benefits of the general PR (multi classification).

markdregan: I copied this from `FreqaiMultiOutputRegressor` directly. My understanding was that this was your solution to `MultiOutputRegressor` not allowing fit_params be passed to underlying models. I must be misunderstanding something?

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
```diff
@@ -0,0 +1,127 @@
...
+            classes corresponds to that in the attribute :term:`classes_`.
+
+            .. versionchanged:: 0.19
+                This function now returns a list of arrays where the length of
+                the list is ``n_outputs``, and each array is (``n_samples``,
+                ``n_classes``) for that particular output.
+                """
+        check_is_fitted(self)
+        results = np.hstack([estimator.predict_proba(X) for estimator in self.estimators_])
+        return np.squeeze(results)
```
markdregan, line 86: Question @robcaulk on my approach here. For both `predict` and `predict_proba` I copied the code from `MultiOutputClassifier` and made a tiny modification to add `np.squeeze()` - because the output was of dimension (1, num_samples, num_features). This was causing issues in `freqai_interface` when `predict()` is called. So I decided to instantiate `predict()` and `predict_proba()` in  use `np.squeeze()` in `FreqaiMultiOutputClassifier` to address this issue. Which then returned results in dimensions (num_samples, num_features). 

Just explaining my thinking here. And why this code is duplicated from the underlying base class. Is there a simplier way to achieve this without me copy pasting the code from the underlying class? Perhaps use `super().predict()` and just modify the output of that? (not experienced coder so apologies if missing basics here)

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
markdregan: Sounds good. I can remove. First see my comment above. Main reason I copied this from underlying base class was because I wanted to modify the dimensions/shape of the output from `FreqaiMultiOutputClassifier`. I think there must be a cleaner way for me to do this.

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
robcaulk: Yes, this is the functionality you duplicated in your `CatboostClassifierMultiTarget.py`. Best to just leave the functionality here in the base class, same as the code you copied this from.

freqtrade/freqai/prediction_models/CatboostClassifierMultiTarget.py
markdregan: This was copied from `FreqaiMultiOutputRegressor`. I was assuming I should base `FreqaiMultiOutputClassifier` off this identically with few minor modifications. I'm misunderstanding something here?

robcaulk: Ah Im wrong here, forget my comment. This is correctly matching the other CatboostMultiOutputRegressor.

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
robcaulk: Sorry I was mistaken here. You are correct.

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
markdregan: I've simplified in most recent commit to use super().

markdregan:
Discussed with @robcaulk - the issue with `multitarget_parallel_training` is shared across Regressor and Classifier and will be resolved via a separate PR. So I won't address in this PR.

This PR is working with `multitarget_parallel_training = false`. 

I've minimized duplicate code in `predict` and `predict_proba` using super().

I've added OperationalException if duplicate class labels are detected across targets. Not sure if this is the preferred way to raise exceptions in FT. LMK.

freqtrade/freqai/base_models/FreqaiMultiOutputClassifier.py
robcaulk: This seems good to me.

freqtrade/freqai/prediction_models/CatboostClassifierMultiTarget.py
```diff
@@ -0,0 +1,74 @@
...
+
+from freqtrade.freqai.base_models.BaseClassifierModel import BaseClassifierModel
+from freqtrade.freqai.base_models.FreqaiMultiOutputClassifier import FreqaiMultiOutputClassifier
+from freqtrade.freqai.data_kitchen import FreqaiDataKitchen
+
+
+logger = logging.getLogger(__name__)
+
+
+class CatboostClassifierMultiTarget(BaseClassifierModel):
```
robcaulk, line 16: Is this much different for `LightGBMClassifierMultiTarget` and `XGBoostClassifierMultiTarget`? 

I prepared the tests to be ready to test those as well - assuming they are easy to create.

markdregan: It should be very similar. I'd need to check if same modification of predict and predict_proba is needed.

Thanks for adding tests.

robcaulk: I got it working with lightgbm, xgboost was complaining.

If a user comes to us asking for XGBoost multiclassifier -then I will sort it out. Otherwise - we let it be for now.

robcaulk (approved):
LGTM thanks!

robcaulk:
> Discussed with @robcaulk - the issue with `multitarget_parallel_training` is shared across Regressor and Classifier and will be resolved via a separate PR. So I won't address in this PR.
> 
> This PR is working with `multitarget_parallel_training = false`.
> 
> I've minimized duplicate code in `predict` and `predict_proba` using super().
> 
> I've added OperationalException if duplicate class labels are detected across targets. Not sure if this is the preferred way to raise exceptions in FT. LMK.

Just to conclude - the use of `multitarget_parallel_training` is a bit of a useless feature for 99% of applications. That is why it is not documented or advertised. This is because the decision tree algorithms are already extremely well parallelized - so running multiple trainings sequentially on all the threads is probably going to be identical to running multiple trainings in parallel. Except the multiple trainings in parallel adds additional parallelization overhead.

This changes for people who may have multiple GPUs per system. But that is currrently not a common use-case. If we have users come to us saying they want the `*MultiTarget` to work better with multi-gpu systems - then we can investigate solving the pickling problem bringing this feature back. 

For now, this works quite well and achieves a very wide use-case for alot of current users. So we merge it :-)

PR #7647, opened 2022-10-26
Combine stop logic

xmatthias:
## Summary

Combine the stop-logic of different exchanges, which is actually almost identical.

## Quick changelog

- use common adjust_stop function instead of one per exchange with identical logic.


robcaulk (approved):
LGTM

xmatthias:
won't merge this before the release though ...

wizrds:
LGTM also 😁

PR #7659, opened 2022-10-29
Remove loop of normalization from metadata

aemr3:
## Summary
Normalization from metadata was way slow if the features count was high since it was looping through whole dataframe columns. Moved calculation to pandas to make it faster.

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -354,11 +354,12 @@ class FreqaiDataKitchen:
         :param df: Dataframe to be standardized
         """
 
-        train_max = []
-        train_min = []
-        for item in df.keys():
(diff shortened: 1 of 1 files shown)


robcaulk (approved):
Really great work finding that! 

I made a small tweak and tested it. LGTM!

aemr3:
> Really great work finding that!
> 
> I made a small tweak and tested it. LGTM!

I was just going to change variables like this:

```
train_max = pd.Series(
    data=[self.data[f"{item}_max"] for item in df.keys()],
    index=df.keys()
)
train_min = pd.Series(
    data=[self.data[f"{item}_min"] for item in df.keys()],
    index=df.keys()
)
```

But I guess creating a list with known length is better. Thanks.

PR #7662, opened 2022-10-30
Fix missing f-string from PR #7611

markdregan:
## Summary

I mistakenly removed an f-string before submitting PR #7611

Solve the issue: PR #7611

## Quick changelog

- Added f-string to L963

## What's new?

Ability to use extra_returns in backtest.


robcaulk (approved):
LGTM

PR #7689, opened 2022-11-02
add integrated tests for PCA and DBSCAN

robcaulk:
Here we add PCA and DBSCAN in-situ tests (without adding any additional test timing). This should catch hard to find data transformation issues before they make it to develop.


xmatthias (approved):
LGTM :+1:  (i'll never complain about better / improved tests) :laughing:

PR #7692, opened 2022-11-02
Fix Memory Leak in Websockets

wizrds:
## Summary

This is a hotfix for a memory leak in the WebSockets regarding an error bubbling up to the `broadcast_queue_data` method and stalling the broadcasting thread. This also updates some code blocks regarding channel disconnecting, as well as some more quality of life improvements.

## Quick changelog

- Update Channel Disconnecting
- Call `queue.task_done` in `_broadcast_queue_data`
- On finishing the `_broadcast_queue_data` function, the `self._ws_queue` is set to None to prevent leaking
- Refactor `channel.send` to avoid using `queue.put` resulting in a RuntimeError when a queue is full
- The consumer now displays the latency to the producer from a ping

### DIFF
--- a/freqtrade/rpc/api_server/webserver.py
+++ b/freqtrade/rpc/api_server/webserver.py
@@ -197,6 +197,7 @@ class ApiServer(RPCHandler):
                 # Get data from queue
                 message: WSMessageSchemaType = await async_queue.get()
                 logger.debug(f"Found message of type: {message.get('type')}")
+                async_queue.task_done()
                 # Broadcast it
                 await self._ws_channel_manager.broadcast(message)
         except asyncio.CancelledError:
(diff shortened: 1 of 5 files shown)


freqtrade/rpc/api_server/webserver.py
```diff
@@ -210,6 +211,9 @@ async def _broadcast_queue_data(self):
             # Disconnect channels and stop the loop on cancel
             await self._ws_channel_manager.disconnect_all()
             self._ws_loop.stop()
+            # Avoid adding more items to the queue if they aren't
+            # going to get broadcasted.
+            self._ws_queue = None
```
xmatthias, line 216: This is the killer feature - as it'll mean that even if this loop crashes for some reason - it'll never start to leak due to queue getting full.

xmatthias (approved):
Nice fix :100: 

I've improved wording on the waiting a bit - i feel otherwise it may be dangerous that someone things "well the queue can do that on it's own" in a few months (might even be me) - and revert it to prior behavior.

I've given it a spin and wasn't able to reproduce the issue we've seen yesterday - but it's obviously a timing issue, so not seeing it doesn't guarantee it won't happen (Though logic tells me it's fixed).

LGTM :+1:

Issue #7666, opened 2022-10-30
Back test results inconsistent with FreqAI


samee-sys:
## Describe your environment

  * Operating system: Windows 10
  * Python Version: 3.10.4 
  * Freqtrade Version: 2022.9.1
  

## Describe the problem:

I am using a Freqai strategy. with same configuration and same strategy, I am back-testing two timeperiods. 
1) 2022-09-07 to 2022-09-23 - result is 5.35% profit. Note down the weekly breakdown. Week 11/09: 4.816, 18/09: 15.542
2) 2022-09-07 to 2022-09-29 - result is -2.53%. The inconsistency is in the weekly breakdown. Week 11/09: -6.259, 18/09: 6.99

As you can see only the last week should have different result as config and strategy is same. But all weeks have different results. Refer attached screenshots for weekly breakdown and results.

![image](https://user-images.githubusercontent.com/58465165/198883315-abd9b880-db36-40cc-a494-2887624931b7.png)

![image](https://user-images.githubusercontent.com/58465165/198883364-042d19ef-bddb-47b8-aecf-0f8960ae1883.png)

Request to have a look and resolve this issue. Thank you.
### Steps to reproduce:

  1. Create freqtrade config and strategy with 1 hours timeframe.
  2. Backtest the strategy with 3 weeks time.
  3. Backtest with the same time period + one more week i.e 4 weeks.

wagnercosta:
Hi,

Thanks for reporting this.
I'm working on this issue and I believe I've managed to reproduce the problem. Please, could you inform the contents of the `freqai` section of your config file?

samee-sys:
@wagnercosta thanks for looking into the issue. Below is the content from Freqai section:
```
"freqai": {
        "enabled": true,
        "purge_old_models": true,
        "train_period_days": 90,
        "identifier": "bbtrend45",
        "feature_parameters": {
            "include_timeframes": [                
                "1h"
            ],
            "include_corr_pairlist": [
                "BTC/USDT"
            ],
            "label_period_candles": 2,
            "include_shifted_candles": 3,
            "DI_threshold": 0.9,
            "weight_factor": 0.9,
            "principal_component_analysis": false,
            "use_SVM_to_remove_outliers": true,
            "indicator_periods_candles": [9, 21, 55],
        },
        "data_split_parameters": {
            "test_size": 0,
            "random_state": 1
        },
        "model_training_parameters": {
            "n_estimators": 800
        }
    }
```

[...]

PR #7695, opened 2022-11-03
Fix inconsistent backtesting results - FreqAI

wagnercosta:
## Summary

This PR fix an issue for backtesting with inconsistent values when the user is using the `backtest_period_days` parameter with a value greater than 1  and when performing backtesting with different timeranges.

Solve the issue: #7666

## What's new?

The first backtesting execution save prediction file based on the training timestamp. When the `backtest_period_days` parameter is greater than 1, or when not specifying it in the configuration file (default: 7), for the next execution with a new timerange the saved file is inconsistent with the new timerange period. This PR adds a new check to verify if the number of predictions is the same length of backtesting slice dataframe for the training period.


robcaulk (approved):
LGTM!

PR #7699, opened 2022-11-04
FreqAI: Reorder data cleaning

th0rntwig:
## Summary

Change the order of data manipulation in `data_cleaning_train()` and `data_cleaning_predict()` in `data_kitchen` so that outlier identification and removal is done before calculation of dissimilarity index (`DI`), and `PCA` transformation.

## Quick changelog

The input data is manipulated, according to user input, before being used for training/testing/predicting. This PR changes the order of such data manipulation so that all outlier detection/removal is done prior to calculation of the dissimilarity index (`DI`), and so that `PCA` transformation is done as the last step. This reordering increases consistency as all currently available outlier detection/removal is now done prior to the calculation of the `DI`, and the `PCA` transformation. This reordering also means that outliers no longer have an influence on the `DI` or `PCA`.
This required some refactoring of the code, and changes to some tests, to ensure that un-transformed (no `PCA`) data remains available for outlier detection/removal using `DBSCAN` (if enabled by the user).

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -273,11 +273,15 @@ class FreqaiDataKitchen:
         test_labels: DataFrame,
         train_weights: Any,
         test_weights: Any,
+        train_df_no_transf: DataFrame = DataFrame(),
+        test_df_no_transf: DataFrame = DataFrame()
     ) -> Dict:
 
         self.data_dictionary = {
             "train_features": train_df,
+            "train_features_no_transf": train_df_no_transf,
             "test_features": test_df,
+            "test_features_no_transf": test_df_no_transf,
             "train_labels": train_labels,
             "test_labels": test_labels,
             "train_weights": train_weights,
@@ -289,7 +293,7 @@ class FreqaiDataKitchen:
 
     def normalize_data(self, data_dictionary: Dict) -> Dict[Any, Any]:
         """
-        Normalize all data in the data_dictionary according to the training dataset
+        Normalize all data in the data_dictionary according to the training dataset.
         :param data_dictionary: dictionary containing the cleaned and
                                 split training/test data/labels
         :returns:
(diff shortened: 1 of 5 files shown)


freqtrade/freqai/data_kitchen.py
```diff
@@ -682,7 +693,10 @@ def use_DBSCAN_to_remove_outliers(self, predict: bool, eps=None) -> None:
         if predict:
             if not self.data['DBSCAN_eps']:
                 return
-            train_ft_df = self.data_dictionary['train_features']
+            if self.data_dictionary["train_features_no_transf"].empty:
```
aemr3, line 696: I get KeyError here if I restart after training. Tried with and without PCA.

```  
File "/freqtrade/freqtrade/freqai/data_kitchen.py", line 723, in use_DBSCAN_to_remove_outliers
    if self.data_dictionary["train_features_no_transf"].empty:
KeyError: 'train_features_no_transf'
```

freqtrade/freqai/data_kitchen.py
```diff
@@ -495,6 +499,9 @@ def principal_component_analysis(self) -> None:
         """
 
         from sklearn.decomposition import PCA  # avoid importing if we dont need it
+        self.data_dictionary["train_features_no_transf"] = self.data_dictionary["train_features"]
+        self.data["training_features_list_no_transf"] = \
```
aemr3, line 503: Is this used somewhere? It looks like unnecessary.

freqtrade/freqai/data_kitchen.py
```diff
@@ -918,9 +933,13 @@ def check_if_pred_in_training_spaces(self) -> None:
         and avoid making predictions on any points that are too far away
         from the training data set.
         """
+        if self.data_dictionary["train_features_no_transf"].empty:
```
aemr3, line 936: This also gives KeyError if DBSCAN is not used and after a restart.

robcaulk:
While I agree with the rationale behind this PR - the thing that becomes most clear to me is how fragile this component of FreqAI currently is. 

I think we need to step back and restructure the `FreqaiDataKitchen` so that it starts to behave more like a "Freq"-model. In the sense of, we should be keeping all model specific objects together (e.g. trained model file, PCA transform, SVM model, train_df, metadata). Right now they are scattered, and when we want to make a simple change like this PR, we end up needing to touch a bunch of tentacles throughout DK and DD. In the new "Freq"-model paradigm, serialization and reinitialization from crash becomes *even better* (even tho crash resilience is already one of FreqAI's strong points). Further, DataKitchen needs a cleaning (pun intended), because it currently holds a bunch of backtesting logic that we should separate into a new file. 

So I am going to close this in preference of a restructuring that follows a more modular logic described above. This is certainly on my to-do list, but if you feel like taking a whack at it - I can give you as much support as possible to make it happen.

PR #7701, opened 2022-11-04
add option to force single precision

robcaulk:
Add option for users to force single precision for the full dataframe. The results are:

* decreased disk usage (~70%)
* decreased ram usage (~70%)
* faster training (~20% LightGBM, 0% Catboost, X% XGBoost)

To activate, add:

```
    "freqai": {
       ...
        "convert_df_to_float32": true
      ...
   }
```

to the FreqAI config.


freqtrade/freqai/data_kitchen.py
```diff
@@ -1344,3 +1347,24 @@ def check_if_backtest_prediction_is_valid(
                 f"Could not find backtesting prediction file at {path_to_predictionfile}"
             )
             return False
+
+    def reduce_dataframe_footprint(self, df: DataFrame) -> DataFrame:
```
xmatthias: To be quite frank - i think this method should not be "here" (part of data_kitchen, maybe not even freqAI).

first of all - it's not using any "self" - so it's independent of data_kitchen (more like standalone util) .

but MUCH more important than this - i think every strategy can benefit from this (not just freqAI).

as such - i think we should 
* be moving this to "some" generic place (maybe a new file in `util` ?) 
* move the setting to the root of the configuration (out of the freqAI realm)

Obviously the method will still be called by freqAI for `populate_any_...` dataframes - but i'd also aim to make it part of the regular strategy interface - where we call this right after the regular "populate_*" calls (optional based on the setting, obviously).

we don't have to do that implementation in this PR (even though i think it would be a great improvement) - but i think by placing the method (and corresponding setting) strategically would be nice.

freqtrade/freqai/data_kitchen.py
```diff
@@ -1344,3 +1347,24 @@ def check_if_backtest_prediction_is_valid(
...
+        start_mem = df.memory_usage().sum() / 1024**2
+        print("Memory usage of dataframe is {:.2f} MB".format(start_mem))
+
+        for col in df.columns[1:]:
+            col_type = df[col].dtype
+            if col_type != object:
+                if str(col_type)[:3] == "int":
+                    df[col] = df[col].astype(np.int32)
+                else:
+                    df[col] = df[col].astype(np.float32)
```
xmatthias: i think you'd want an `elif` here with `isinstance(df[col], np.floating)` (i'd also do a comparison against `np.integer` above in a similar fashion).

you can't outright assume that everything that's not an int is a float - it could just as well be a bool (explicitly assigned type) - or a string (`dataframe['dir'] = 'raising'`)

robcaulk: looks resolved.

robcaulk:
G2G in my opinion. Anyone have any reservations on this?

freqtrade/freqai/data_kitchen.py
xmatthias: @robcaulk I don't see why you've resolved this.

I've asked this to be moved from Datakitchen to a place where it's generally accessible, as well as the config to be moved to the "general config" - not freqAI section of the configuration.
Marking things as resolved without comment or action is NOT how PR comments work. 
That's "ignoring a PR comment" - which is not how i'd like to collaborate (neither on this project nor elsewhere).

Now letme ask for this in a clearer way (hoping to make it clear what i would like):

Please move this to a generic place (probably a new file in `utils/`, or in converter (see #7703 for a proper place to place a generic function (one that's not tied to freqAI specifically)).

xmatthias (changes requested):

freqtrade/freqai/data_kitchen.py
robcaulk: Thanks for the feedback. You seemed to indicate that your comment was optional - but it would appear that was miscommunicated. 

I have made the adjustments that you requested.

xmatthias: yeah - the "optional" part was meant for the 3rd part - which is implement it for non-freqAI strategies right away.

freqtrade/constants.py
```diff
@@ -159,6 +159,7 @@
         'ignore_buying_expired_candle_after': {'type': 'number'},
         'trading_mode': {'type': 'string', 'enum': TRADING_MODES},
         'margin_mode': {'type': 'string', 'enum': MARGIN_MODES},
+        'convert_df_to_32bit': {'type': 'number', 'default': False},
```
xmatthias: what do you think about a more ... generic name for the parameter?

it's true that we're converting to 32bit for now - but we could just as well convert to 16bit ints (for example) in the future - so maybe a name like `reduce_df_footprint` might keep more possibilities open without explicitly stating the final size.

robcaulk: Sure go for it!

xmatthias (approved):
LGTM :+1:

PR #7702, opened 2022-11-04
move write_metrics_to_disk to proper place in param table

robcaulk:
`write_metrics_to_disk` was located in the incorrect part of the parameter table. This MR moves it to the correct location.


xmatthias (approved):
:+1:

PR #7707, opened 2022-11-06
backtesting - use full detail timeframe

### LINKED ISSUE #7529: backtesting with timeframe-detail fills orders on "long" timeframe
## Describe your environment

  * Operating system: All
  * Python Version: 3.10
  * CCXT version: 1.93.90
  * Freqtrade Version: 2022.9

## Describe the problem:

Seems there is incorrect handling of the custom_stoploss in backtesting. When timeframe detail used, the stoploss call does not match the expected price changes within the timeframe.
Example:
- timeframe=1h, timeframe-detail=1m
- enter trade and stop loss transactions will happen at a single 1h candle (let's say, candle starts at 0:00)
- enter trade transaction is in the middle of the candle (e.g. 0:45)
- custom_stoploss seems to be activated before actual enter trade, so seems this function does not respect the timeframe-detail parameter (assume stop loss condition was satisfied only between 0:30-0:35)
- observed trade executed at the price from 0:45 and stop loss price taken from 0:30

### Steps to reproduce:

  1. request a transaction that will be executed in the middle of the candle
  2. define a custom-stoploss,
  3. run a backtesting with timeframe-detail
  4. incorrect stop loss transaction may be observed when stop loss conditions satisfied at the candle before the moment when transaction should be executed
  
### Observed Results:

  * What happened?
 Observed the stop loss call that seems to be relevant to the data which is valid before the transaction execution.
 Stop loss should be executed obviously only after transaction execution.

### LINKED ISSUE #7768: Backtest vs dryrun difference
Operating system: Windows Server 2016
Python Version: Python 3.8.3
CCXT version: I did try thi, but could not get a result when entered pip freeze (there was no ccxt on the list)
Freqtrade Version: 2022-08
  
I did run some backtesting and hyperopting on 4h candle strategies. I did use following options for hyperopt and backtesting to make sure I sue 1 minute details on top of the 4h data. 
               "--timeframe-detail", 
               "1m", 

I had the 1m and 4h data on downloaded from Binance. My entry was at 97.9% price of the previous closed 4h candle. I did run the dryrun trading and also backtested the same period afterwards. I face the following issue. The strategy I used purchased BONDUSDT based on the closed candle of 2022-11-12 8pm. It bought on 4.283 at 2022-11-13 00:06. In the dryrun mode it could not sell for more than 4 days. BUT the backtest results said that it could sell at 4.337 rate. I did manually check it and see that the price never went up to 4.337 after 2022-11-13 00:06. How could it that the backtest says that it could sell (we had 1m timedetail used in the backtest).

I did try to find similar case here, but could not. Thank you!

Thank you for your help in advance.
Peter

xmatthias:
## Summary

closes #7529

## Quick changelog

- Use detail timeframe for order fills and exit condition checks
- Entry's order placement will only occur on the main timeframe (filling on the detail)
- Entry and exit fills will occur based on the detail timeframe - assuming custom pricing is used (otherwise all orders fill immediately).

## caveats

* this PR is expected to slow down backtesting with `--timeframe-detail` slightly (depending on the number of open trades) - since it's now "branching" out into the detail timeframe for all candles, not just for candles during open trades.
* There is no significant impact expected on backtests without `--timeframe-detail`.

### DIFF
--- a/docs/backtesting.md
+++ b/docs/backtesting.md
@@ -583,7 +583,8 @@ To utilize this, you can append `--timeframe-detail 5m` to your regular backtest
 freqtrade backtesting --strategy AwesomeStrategy --timeframe 1h --timeframe-detail 5m
 ```
 
-This will load 1h data as well as 5m data for the timeframe. The strategy will be analyzed with the 1h timeframe - and for every "open trade candle" (candles where a trade is open) the 5m data will be used to simulate intra-candle movements.
+This will load 1h data as well as 5m data for the timeframe. The strategy will be analyzed with the 1h timeframe, and Entry orders will only be placed at the main timeframe, however Order fills and exit signals will be evaluated at the 5m candle, simulating intra-candle movements.
+
 All callback functions (`custom_exit()`, `custom_stoploss()`, ... ) will be running for each 5m candle once the trade is opened (so 12 times in the above example of 1h timeframe, and 5m detailed timeframe).
 
 `--timeframe-detail` must be smaller than the original timeframe, otherwise backtesting will fail to start.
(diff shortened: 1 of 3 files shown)


froggleston:
Did a quick test of this yesterday - nothing broke :)

PR #7708, opened 2022-11-06
Improve iteration logic

xmatthias:
## Summary

Slightly improve iteration logic, avoiding iterations that hit the spot between a new candle issuance and when the candle is ready on exchange (between :00 and :01).
Hitting that spot currently means that new candles would only be fetched at :05 (with default settings) - which is unnecessarily late.


robcaulk (approved):
LGTM

PR #7724, opened 2022-11-09
Support git and local changes in dev containers

### LINKED ISSUE #7723: VS Code Dev Container has no access to the local source code
## Describe your environment

  * Operating system: Mac OS Ventura 13.0
  * Python Version: [container environment]
  * CCXT version: [container environment]
  * Freqtrade Version: source code in branch develop

## Describe the problem:

I'm not sure if it is a bug or a feature :) but ... I'm using VS Code for development and I noticed that devcontainer.json file is thre. Which is quite a good idea for a new developer like me to start with the project. So, I opened the dev container in vs code. Then I faced the following problem. 
- The code which I see in the dev container is not in sync with my local code
- I have no access to git, so I can not actually develop in this dev container

### Steps to reproduce:

  1. Reopen the dev container in vs code
  2. Edit any file
  3. Reopen local code
  4. No changes on the local
  
### Observed Results:

  * I can not really use the dev container as it is supposed to be used
  * I expect when I change the code, I can also create a branch, and commit and also switch back to local and see the changes locally (as it is designed with vs code dev containers)

  ```
   no logs
  ```

wassertim:
## Summary

This PR fixes the problem with git and changes sync in vs code dev container feature

Solve the issue: #7723

## Quick changelog

- postCreateCommand is changed by adding symlink to freqtrade .env directory from the workspace/freqtrade (to have the install environment in the source folder)
- workspace folder changed from /freqtrade to /workspaces/freqtrade
- local sources mount to /workspaces/freqtrade

## What's new?

Now you can actually develop in a vs code dev container

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,7 +15,8 @@
     // Uncomment to connect as a non-root user if you've added one. See https://aka.ms/vscode-remote/containers/non-root.
     "remoteUser": "ftuser",
 
-    "postCreateCommand": "pip install --user -e . && freqtrade create-userdir --userdir user_data/",
+    "onCreateCommand": "pip install --user -e .",
+    "postCreateCommand": "freqtrade create-userdir --userdir user_data/",
 
     "workspaceFolder": "/workspaces/freqtrade",
 
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,9 +15,7 @@
     // Uncomment to connect as a non-root user if you've added one. See https://aka.ms/vscode-remote/containers/non-root.
     "remoteUser": "ftuser",
 
-    "postCreateCommand": "ln -s /freqtrade/.env /workspaces/freqtrade/.env && freqtrade create-userdir --userdir user_data/",
-
-    "workspaceFolder": "/workspaces/freqtrade",
+    "postCreateCommand": "ln -s /freqtrade/.env /workspaces/freqtrade/.env && freqtrade create-userdir --userdir user_data/",    
 
     "settings": {
         "terminal.integrated.shell.linux": "/bin/bash",
@@ -35,7 +33,6 @@
         "ms-python.vscode-pylance",
         "davidanson.vscode-markdownlint",
         "ms-azuretools.vscode-docker",
-        "vscode-icons-team.vscode-icons",
-        "github.copilot",
+        "vscode-icons-team.vscode-icons",        
     ],
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -11,12 +11,13 @@
     "mounts": [
         "source=freqtrade-bashhistory,target=/home/ftuser/commandhistory,type=volume"
     ],
+    "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/freqtrade,type=bind,consistency=cached",
     // Uncomment to connect as a non-root user if you've added one. See https://aka.ms/vscode-remote/containers/non-root.
     "remoteUser": "ftuser",
 
-    "postCreateCommand": "freqtrade create-userdir --userdir user_data/",
+    "postCreateCommand": "ln -s /freqtrade/.env /workspaces/freqtrade/.env && freqtrade create-userdir --userdir user_data/",
 
-    "workspaceFolder": "/freqtrade/",
+    "workspaceFolder": "/workspaces/freqtrade",
 
     "settings": {
         "terminal.integrated.shell.linux": "/bin/bash",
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -11,12 +11,13 @@
     "mounts": [
         "source=freqtrade-bashhistory,target=/home/ftuser/commandhistory,type=volume"
     ],
+    "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/freqtrade,type=bind,consistency=cached",
     // Uncomment to connect as a non-root user if you've added one. See https://aka.ms/vscode-remote/containers/non-root.
     "remoteUser": "ftuser",
 
-    "postCreateCommand": "freqtrade create-userdir --userdir user_data/",
+    "postCreateCommand": "ln -s /freqtrade/.env /workspaces/freqtrade/.env && freqtrade create-userdir --userdir user_data/",
```
xmatthias: the `.env` linking seems wrong (maybe you can explain your reasoning on this??).

Why should we use a virutalenv within a container which has freqtrade installed already (That's the whole point of using the devcontainer - to have all dependencies contained within it).

Assuming you're on windows - then the virtual environment would even be broken (python venv vs. linux docker container).

.devcontainer/devcontainer.json
```diff
@@ -11,12 +11,13 @@
     "mounts": [
         "source=freqtrade-bashhistory,target=/home/ftuser/commandhistory,type=volume"
     ],
+    "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/freqtrade,type=bind,consistency=cached",
```
xmatthias, line 14: is this actually necessary? 
I've tried to create a devcontainer config in vscode on an (almost) empty directory - and the only keys required were "name", "image" and "features" (which we apparently don't use - as we use the freqtrade upstream docker image with requirements-dev installed).

.devcontainer/devcontainer.json
```diff
@@ -35,5 +36,6 @@
         "davidanson.vscode-markdownlint",
         "ms-azuretools.vscode-docker",
         "vscode-icons-team.vscode-icons",
+        "github.copilot",
```
xmatthias: while i do like copilot personally - that's clearly too controversial to force onto future developers by default.
everyone can install this on his own within the container - but it should clearly not be part of the default config - at least not at this point.

.devcontainer/devcontainer.json
wassertim: I remember I could not find the workspaces folder without it. I will check again

.devcontainer/devcontainer.json
wassertim: Yes your are right. This one I just forgot to clean up, when I was preparing the PR

.devcontainer/devcontainer.json
wassertim: Sure, I will try. If you mount the source folder and not the folder from the docker image, you lose the installation, which is done during docker build. So in this case the installation is done in /freqtrade dir, which has the .env. But the local sources don't have it and that is correct since you don't want to install the dependencies all over again on your local machine, neither through this setup.

wassertim: So it is kind of a workaround to have the source mounted and the freqtrade command in place

.devcontainer/devcontainer.json
wassertim: It actually does not work without this binding. So it has to stay

.devcontainer/devcontainer.json
wassertim: > Assuming you're on windows - then the virtual environment would even be broken (python venv vs. linux docker container).

It should not be platform specific, because it is a command in a linux container, should work the same on windows and mac and linux

wassertim:
Thanks for such quick review @xmatthias !

.devcontainer/devcontainer.json
xmatthias: the `.env` directory is coming from the host (is mounted) - so this is clearly wrong (you shouldn't activate or use a virtual environment from within the container).

i think IF we're moving the directory from /freqtrade to /workspace/freqtrade - then a simple `pip install --user -e .`on that directory is the proper approach.

.devcontainer/devcontainer.json
xmatthias: i'm not sure about that. The devcontainer is also supposed to work on github codespaces - which it does in it's current form just fine (without any modification).

i still have to try the local version - but last time i tried, it worked just fine.

.devcontainer/devcontainer.json
xmatthias: The correct command to put here (considering the other changes) seems to be `pip install --user -e .`.
That'll make sure the correct freqtrade version is used (not the one in `/freqtrade`).

wassertim: Yes, it seems it is the way to go. This is a very good remark.

xmatthias (approved):
Great find :100: 

i did one minor tweak, which is to split the installation into the "onCreateCommand" hook - which seems to run earlier / less frequently if running on codespaces.

LGTM :+1:

PR #7728, opened 2022-11-10
Simplify timerange handling

xmatthias:
## Summary

This PR greatly simplifies timerange handling by moving common operations into the timerange class itself (where they belong).

This was long overdue - and is now becoming more relevant than ever - since freqAI is using the timerange object in more ways than it was originally intended - which results in (the same) logic spread throughout the codebase.

## Quick changelog

- Move some logics to timerange object.


robcaulk (approved):
This is much more concise. Thanks for that!

PR #7730, opened 2022-11-11
fix loading of metric tracker from disk

robcaulk:
Fixes bug where DataDrawer found the metric tracker on disk, but overwrites it with an empty disk soon after.

### DIFF
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -87,6 +87,7 @@ class FreqaiDataDrawer:
             self.create_follower_dict()
         self.load_drawer_from_disk()
         self.load_historic_predictions_from_disk()
+        self.metric_tracker: Dict[str, Dict[str, Dict[str, list]]] = {}
         self.load_metric_tracker_from_disk()
         self.training_queue: Dict[str, int] = {}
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/data_drawer.py
```diff
@@ -87,6 +87,7 @@ def __init__(self, full_path: Path, config: Config, follow_mode: bool = False):
             self.create_follower_dict()
         self.load_drawer_from_disk()
         self.load_historic_predictions_from_disk()
+        self.metric_tracker: Dict[str, Dict[str, Dict[str, list]]] = {}
```
xmatthias, line 90: Just one minor comment (to be addressed in a future PR).

would we be able to reinitialize (i think it needs to be done twice ...?) a data-drawer in a test (just reinit) after a tracker file has been written without huge performance impact - so we can then assert if `metric_tracker` is not empty (has been loaded without being overwritten)?

whenever initialization logic depends on such intristic sequences i see a high risk of future regression if it's not asserted properly.

No test will fail, and no "regular" behavior will fail either when testing a PR - so even a manual test will not notice this UNLESS you explicitly test for this.

xmatthias (approved):
LGTM :+1:

PR #7735, opened 2022-11-12
Improve python GC behavior

xmatthias:
## Summary
As explained in https://www.youtube.com/watch?v=p4Sn6UcFTOU - this PR will tune the GC collector settings to not run as often.

Impact of this becomes clear when running `freqtrade show-trades` against a huge database - where i've seen 20-30% speedup (without increased memory consumption) without any further action.

Running with GC Debug mode (`gc.set_debug(gc.DEBUG_STATS)`) shows how often GC is actually running for a simple command as the one above.


epigramx:
I think I see roughly a 5 to 6% improvement, on backtesting, but it wasn't a very controlled test.

xmatthias:
> I think I see roughly a 5 to 6% improvement, on backtesting, but it wasn't a very controlled test.

with "improvement" - you mean speed improvement (runs faster)? or increased / decreased memory consumption?

epigramx:
> with "improvement" - you mean speed improvement (runs faster)? or increased / decreased memory consumption?

speed, (with command 'time')

epigramx:
it terms of memory I didn't see much of a difference, so I don't know what the default setting is even doing as a trade-off, probably nothing useful due to high efficiency of the code tested.

PR #7736, opened 2022-11-12
Update README.md

ghost:
Removed the FTX option, it went bankrupt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,6 @@ Please read the [exchange specific notes](docs/exchanges.md) to learn about even
 
 - [X] [Binance](https://www.binance.com/)
 - [X] [Bittrex](https://bittrex.com/)
-- [X] [FTX](https://ftx.com/#a=2258149)
 - [X] [Gate.io](https://www.gate.io/ref/6266643)
 - [X] [Huobi](http://huobi.com/)
 - [X] [Kraken](https://kraken.com/)


README.md
```diff
@@ -28,7 +28,6 @@ Please read the [exchange specific notes](docs/exchanges.md) to learn about even
 
 - [X] [Binance](https://www.binance.com/)
 - [X] [Bittrex](https://bittrex.com/)
-- [X] [FTX](https://ftx.com/#a=2258149)
```
xmatthias, line 31: I'm not per se against removing this (although i'd have waited to see how it'll play out).

At any rate - we'll also want to remove it from index.md - which contains the same list.

ghost: Ah, i forgot the index.md.
I mean, most of the FTX users lost their money and the CEO made FTX look like it was hacked and transferred the users' crypto money to a different address. I don't know how it'll play out but what would change your mind(removing it later) after this point?

xmatthias: so far - the hack is rumor (it's not confirmed what really happened - although i guess there is a certain base assumption ...).

funds are lost (for now) - that's correct (and i'm affected myself, too) - but it's unclear what will happen precisely - that'll have to be seen over the coming days/weeks.

a "proper" removal would remove all code / references to ftx (it's effectively no longer operational at this point - with the API being down, too now).
We'll do that in due time however - once the final fate of FTX is official (and not just some twitter rumor).

xmatthias:
i think you did missunderstand my intend above - this PR is fine if we can also remove it from the index.md. That's not "remove it all" - but "remove publicity" for FTX - which is perfectly fine.

xmatthias (approved):
Will fix index.md in a subsequent commit.

LGTM :+1:

PR #7737, opened 2022-11-12
FreqAI - Backtesting enhancements and bug fix

wagnercosta:
## Summary

This PR brings enhancements to the 'backtest_live_models' feature and fixes a bug in backtesting (FreqAI)

## What's new?

- Fix a bug that was generating 1 candle offset in predictions
- The backtesting prediction file now has the candle date and it's merged by date to avoid other possible bugs
- Apply `fit_live_predictions` to backtesting if it's enabled.
- Update `freqai_backtest_live_models` to use `historic_predictions.pkl`. This will allow user to use the `freqai_backtest_live_model` and keep the `purge_old_models` as true, using less disk space. Other advantage is to reuse all live dataframe data, resulting a faster backtesting execution. 
- Change backtesting prediction files from hdf5 to feather format
- ~~Change backtesting predictions files to 1 file per pair~~: (Keeping it here just for history. Tests with this change did not bring significant performance improvements. In addition, it has the disadvantage that if the file per pair is corrupted, it is necessary to redo all the predictions for that pair, instead of redoing just the one for a specific model).

## TODO

- [X] Code first version to receive early feedback
- [x] Refactor and document new functions
- [x] Add documentation related to new features
- [x] Add dedicated tests


freqtrade/freqai/freqai_interface.py
```diff
@@ -301,14 +314,38 @@ def start_backtesting(
...
+        :param dataframe: DataFrame = strategy passed dataframe
+        :param metadata: Dict = pair metadata
+        :param dk: FreqaiDataKitchen = Data management/analysis tool associated to present pair only
+        :return:
+            FreqaiDataKitchen = Data management/analysis tool associated to present pair only
+        """
+        pair = metadata["pair"]
+        dk.return_dataframe = dataframe
+        self.dk.set_backtesting_live_dataframe_path(pair)
+        saved_dataframe = self.dk.get_backtesting_live_dataframe()
```
robcaulk: Is there a reason you are using `self.dk` instead of `dk` here? 

I know it is confusing that they both exist. It is *usually* safer to always use `dk` instead of `self.dk`. But if you have a particular reason for using `self.dk`, please let me know.

freqtrade/freqai/freqai_interface.py
```diff
@@ -147,12 +152,20 @@ def start(self, dataframe: DataFrame, metadata: dict, strategy: IStrategy) -> Da
...
+                dk = self.start_backtesting_from_live_saved_files(
+                    dataframe, metadata, self.dk)
+                dataframe = dk.return_dataframe
 
-        dataframe = dk.remove_features_from_df(dk.return_dataframe)
         self.clean_up()
         if self.live:
             self.inference_timer('stop', metadata["pair"])
+            if self.save_live_data_backtest:
+                dk.save_backtesting_live_dataframe(dataframe, metadata["pair"])
```
robcaulk: It seems to me you are rebuilding `historical_predicions.pkl`. Is it possible to use `historic_predictions.pkl` directly instead of building/saving more dataframes? Or perhaps `historical_predictions.pkl` is missing a key piece of information? If so, can we simply add that information to `historical_predictions.pkl`? This way we may be able to consolidate machinery. 

Let me know your thoughs.

robcaulk (changes requested):
This is really clean work, thank you for the effort.

freqtrade/freqai/freqai_interface.py
wagnercosta: I think it's possible to use `historical_predicions.pkl`.
The main reasons I decided to save the DF again are:

- I can easily get dry/live start and end date. My code is just recording the current candle and `historical_predictions.pkl` brings all historical candles from the dataframe. We can count candle numbers, but I don't know if that's the best way to handle it.
- We have separate files per pair and I think it's better to deal with small files in dry/live to attach current candles. The backtest loop is per pair as well, which can bring more performance.
- The new code is relatively simple and isolated, which can bring more flexibility in the future, in case we have different requirements for the 2 features.

I wanted to share what I thought, but honestly it's not a final opinion, as I know the advantages of leaving it unified. With these points what do you think? If you prefer, I can adapt the code to use `historical_predictions.pkl`.

robcaulk: I don’t quite understand. Why do you say it “brings all candles?” It’s all sitting in the `self.dd.historic_prédictions` Dictionary, which is in ram. It has the datetime objects, so you can merge or slice or take the last candle, whatever you need. Is there some performance issue with using self.dd.historic_predictions?

Why do you say “it is per pair?” Self.dd.historic_predictions is also a dictionary where each key is “per pair” to access the per pair dataframes. Can you explain why this is not a sufficient structure so as to warrant writing new files per pair?

wagnercosta: Code updated to use `self.dd.historic_prédictions`

freqtrade/freqai/freqai_interface.py
```diff
@@ -672,7 +705,8 @@ def fit_live_predictions(self, dk: FreqaiDataKitchen, pair: str) -> None:
         for label in full_labels:
             if self.dd.historic_predictions[dk.pair][label].dtype == object:
                 continue
-            f = spy.stats.norm.fit(self.dd.historic_predictions[dk.pair][label].tail(num_candles))
+            f = spy.stats.norm.fit(
+                self.dd.historic_predictions[dk.pair][label].fillna(0).tail(num_candles))
```
robcaulk: This is dangerous - but maybe there is a good reason to do it? I cant imagine we will have NaNs very often in the historic_predictions dataframe - so is this only for startup? 

Otherwise, this modifies the statistics.

wagnercosta: Fixed.

robcaulk (approved):
Good to go

PR #7753, opened 2022-11-14
Improve the training by randomly setting the starting point

richardjozsa:
This is an improvement for all kind of RL stuff, the starting point sets in the first third of the data which ensures that the agent does not only optimizes the policy or Q values to the choosen timerange. That means this keeps the users safe throught now allowing to create badly set up models.


robcaulk:
While I don't disagree with the idea, it seems something is failing in the branch which has to do with the addition of the Monitor:

https://github.com/freqtrade/freqtrade/actions/runs/3469246698/jobs/5796007196#step:8:559

I guess you have it working locally - but this branch has a bug somewhere.

PR #7756, opened 2022-11-16
Support SSL in WebSocket connection

wizrds:
## Summary

Added support for a `secure` option in the producer's list to use the `wss` scheme for producers behind TLS termination

## Quick changelog 

- Added `secure` option in EMC
- Updated docs to include option
- Updated `ws_client` script to support

### DIFF
--- a/docs/rest-api.md
+++ b/docs/rest-api.md
@@ -389,11 +389,12 @@ Now anytime those types of RPC messages are sent in the bot, you will receive th
 }
 ```
 
-#### Reverse Proxy and Websockets
+#### Reverse Proxy setup
 
-There are some quirks when using a reverse proxy with the message websocket endpoint. The message websocket endpoint keeps a long-running connection open between the Rest API and the client. It's built on top of HTTP and uses the HTTP Upgrade mechanism to change from HTTP to WebSockets during connection. There are some challenges that a reverse proxy faces when supporting WebSockets, such as WebSockets are a hop-by-hop protocol, so when a proxy intercepts an Upgrade request from the client it needs to send it's own Upgrade request to the server, including appropriate headers. Also, since these connections are long lived, the proxy needs to allow these connections to remain open.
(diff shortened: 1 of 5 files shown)


freqtrade/constants.py
```diff
@@ -512,6 +512,7 @@
                                 'minimum': 0,
                                 'maximum': 65535
                             },
+                            'secure': {'type': 'boolean', 'default': False},
```
xmatthias, line 515: i'm not against this - but if we think about supporting this - we will have to also have some instructions on how to actually configure the server side.

We currently don't even mention a reverse proxy in the docs - but if the client side supports wss (which you can't get on the server without a reverse proxy) - we need to also show people how to configure that.

this can be as simple as an nginx snippet we know is working to serve both freqUi and websockets...

docs/rest-api.md
```diff
@@ -389,6 +389,23 @@ Now anytime those types of RPC messages are sent in the bot, you will receive th
...
 
+#### Reverse Proxy and Websockets
+
+There are some quirks when using a reverse proxy with the message websocket endpoint. The message websocket endpoint keeps a long-running connection open between the Rest API and the client. It's built on top of HTTP and uses the HTTP Upgrade mechanism to change from HTTP to WebSockets during connection. There are some challenges that a reverse proxy faces when supporting WebSockets, such as WebSockets are a hop-by-hop protocol, so when a proxy intercepts an Upgrade request from the client it needs to send it's own Upgrade request to the server, including appropriate headers. Also, since these connections are long lived, the proxy needs to allow these connections to remain open.
+
+When using Nginx, the following configuration is required for WebSockets to work:
+```
+proxy_http_version 1.1;
+proxy_set_header Upgrade $http_upgrade;
+proxy_set_header Connection $connection_upgrade;
```
xmatthias: In a default nginx setup 

```
Unknown "$connection_upgrade" Variable
```

I'll have a look to improve this section.

wizrds: Whoops :eyes: Guess I missed something when copy pasting that. I fixed it.

xmatthias (approved):
LGTM :+1:

PR #7759, opened 2022-11-17
Fix Pandas to_json memory leak

wizrds:
## Summary

This PR implements a small fix for a memory leak that occurs in `pandas.DataFrame.to_json` with datetime objects, as well as another small QOL fix for WebSocketChannel.

## Quick changelog

- Changed using `to_json` to `to_dict` and special datetime handling
- Updated WebSocketChannel to not send if closed and changed closing sequence

Some more information in regards to the leak: [https://github.com/pandas-dev/pandas/issues/24889](https://github.com/pandas-dev/pandas/issues/24889), and [https://github.com/pandas-dev/pandas/issues/40443](https://github.com/pandas-dev/pandas/issues/40443)

### DIFF
--- a/freqtrade/misc.py
+++ b/freqtrade/misc.py
@@ -262,7 +262,10 @@ def dataframe_to_json(dataframe: pandas.DataFrame) -> str:
     :param dataframe: A pandas DataFrame
     :returns: A JSON string of the pandas DataFrame
     """
-    return dataframe.to_json(orient='split')
+    # https://github.com/pandas-dev/pandas/issues/24889
+    # https://github.com/pandas-dev/pandas/issues/40443
+    # We need to convert to a dict to avoid mem leak
+    return dataframe.to_dict(orient='tight')
 
 
 def json_to_dataframe(data: str) -> pandas.DataFrame:
@@ -271,7 +274,7 @@ def json_to_dataframe(data: str) -> pandas.DataFrame:
     :param data: A JSON string
     :returns: A pandas DataFrame from the JSON string
     """
-    dataframe = pandas.read_json(data, orient='split')
+    dataframe = pandas.DataFrame.from_dict(data, orient='tight')
     if 'date' in dataframe.columns:
         dataframe['date'] = pandas.to_datetime(dataframe['date'], unit='ms', utc=True)
 
--- a/freqtrade/misc.py
+++ b/freqtrade/misc.py
@@ -10,7 +10,8 @@ from typing import Any, Dict, Iterator, List, Mapping, Union
 from typing.io import IO
 from urllib.parse import urlparse
 
-import pandas
+import orjson
+import pandas as pd
 import rapidjson
 
 from freqtrade.constants import DECIMAL_PER_COIN_FALLBACK, DECIMALS_PER_COIN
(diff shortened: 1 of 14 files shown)


scripts/ws_client.py
```diff
@@ -101,7 +101,7 @@ def json_deserialize(message):
     :param message: The message to deserialize
     """
     def json_to_dataframe(data: str) -> pandas.DataFrame:
-        dataframe = pandas.read_json(data, orient='split')
+        dataframe = pandas.DataFrame.from_dict(data, orient='tight')
```
xmatthias: it's now no longer leaking, which is great.

What i don't like are 2 things:
* it's a **breaking change** - as the transferred message slightly changed - which means once a producer updates to this - all consumers need to follow suite immediately, or will not be able to receive any more messages. I think we can counter this quite easily in the producer by running something around `str(orjson.dumps(df.to_dict(orient='')))` - which will revert it back to a string.
* message size almost doubles (in my 1000 columns test, it jumps from 13Mb to >19mb)

a quick output of the prior / after 

```
dumps: {"index":[0],"columns":["date","open","high","low","close","volume"],"data":[[1593907200000,20.993428306022466,22.855164004851282,18.345739767921774,22.516458162430492,199.76584662527665]]}
to_json: {"columns":["date","open","high","low","close","volume"],"index":[0],"data":[[1593907200000,20.993428306,22.8551640049,18.3457397679,22.5164581624,199.7658466253]]}
```

The point that changes is the precision in numbers.
I don't think we can (or should) counter this - it's more an awareness thing - not something "really" problematic.
The reason for this is that apparently, `to_json()` is doing some [funky data conversion](https://stackoverflow.com/questions/62927218/pandas-dataframe-with-float64-incorrectly-exported-by-to-json-function) - so the prior version was slightly changing decimals between producer and consumer - while the new mode won't.

freqtrade/misc.py
```diff
@@ -262,7 +262,10 @@ def dataframe_to_json(dataframe: pandas.DataFrame) -> str:
     :param dataframe: A pandas DataFrame
     :returns: A JSON string of the pandas DataFrame
     """
-    return dataframe.to_json(orient='split')
+    # https://github.com/pandas-dev/pandas/issues/24889
+    # https://github.com/pandas-dev/pandas/issues/40443
+    # We need to convert to a dict to avoid mem leak
+    return dataframe.to_dict(orient='tight')
```
xmatthias: i think we could/should use the following here:

``` python
def dataframe_to_json(dataframe: pandas.DataFrame) -> str:
    """
    Serialize a DataFrame for transmission over the wire using JSON
    :param dataframe: A pandas DataFrame
    :returns: A JSON string of the pandas DataFrame
    """
    # https://github.com/pandas-dev/pandas/issues/24889
    # https://github.com/pandas-dev/pandas/issues/40443
    # We need to convert to a dict to avoid mem leak
    def default(z):
        if isinstance(z, Timestamp):
            return z.timestamp() * 1e3
        raise TypeError

    return str(orjson.dumps(dataframe.to_dict(orient='split'), default=default), 'utf-8')
```

this will keep the consumers compatible (no change in the message formatting).

xmatthias (approved):
Thanks a lot for all your effort and help in tracking this down! took a while - and i must say - the result is quite unexpected.

LGTM :+1:

PR #7771, opened 2022-11-21
Refactor WebSocket API for performance

wizrds:
## Summary

This PR revises the message websocket endpoint and WebSocketChannel API's to increase performance by getting rid of queues completely. Before, the WebSocketChannel kept a main queue for communication between the main thread and the API Server thread, and a queue for each connection holding 32 outgoing messages. This caused the bot to use a lot of memory. This also increases code readability and tools to easily extend the server protocol.

## Quick changelog

- Get rid of all queues for the websockets in the API Server
- Update WebSocketChannel API for readability and full asynchronous
- Add async context manager for easily managing channels
- Ensure connections have no impact on each other
- Add methods for easily running concurrent tasks with the channel
- Add dynamic timeout for disconnecting connections that are too slow
- Warn users if performance of the connections is too poor for the amount of pairs/consumers


freqtrade/rpc/api_server/api_ws.py
```diff
@@ -74,65 +77,28 @@ async def _process_consumer_request(
...
-        # Do nothing
-        pass
-    except Exception as e:
-        logger.error(f"Failed to serve - {ws.client}")
-        # Log tracebacks to keep track of what errors are happening
-        logger.exception(e)
-    finally:
-        if channel:
-            await channel_manager.on_disconnect(ws)
+    async with create_channel(websocket) as channel:
```
xmatthias: There's still a PRETTY nasty exception if i connect to a websocket with an invalid token (also on 48242ca02 - which is the last commit in this branch ATM).
it's curious though - as in my understanding - with an invalid token, we shouldn't even get this far into the method (so line 100 shouldn't be reached).

i can however 100% reliably reproduce this by simply using a wrong token on the consumer.

```
2022-11-24 07:09:46,596 - uvicorn.error - INFO - ('127.0.0.1', 47896) - "WebSocket /api/v1/message/ws?token=secret_ws_t0ken.1" 403
2022-11-24 07:09:46,596 - uvicorn.error - INFO - connection failed (403 Forbidden)                    2022-11-24 07:09:46,598 - uvicorn.error - INFO - connection closed
2022-11-24 07:09:46,598 - freqtrade.rpc.api_server.ws.channel - INFO - Disconnected from channel - WebSocketChannel(3ba2829a, ('127.0.0.1', 47896))                                                         2022-11-24 07:09:46,599 - uvicorn.error - ERROR - Exception in ASGI application
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 230, in run_asgi
    result = await self.app(self.scope, self.asgi_receive, self.asgi_send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__                                                               return await self.app(scope, receive, send)                                                         File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/uvicorn/middleware/message_logger.py", line 86, in __call__                                                              raise exc from None
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/uvicorn/middleware/message_logger.py", line 82, in __call__
    await self.app(scope, inner_receive, inner_send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/fastapi/applications.py", line 270, in __call__                                                                          await super().__call__(scope, receive, send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/applications.py", line 124, in __call__                                                                        await self.middleware_stack(scope, receive, send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/middleware/errors.py", line 149, in __call__                                                                   await self.app(scope, receive, send)                                                                File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/middleware/cors.py", line 76, in __call__
    await self.app(scope, receive, send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/routing.py", line 706, in __call__
    await route.handle(scope, receive, send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/routing.py", line 341, in handle
    await self.app(scope, receive, send)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/routing.py", line 82, in app
    await func(session)
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/fastapi/routing.py", line 287, in app
    await dependant.call(**values)
  File "/home/matt/freqtrade/freqtrade/rpc/api_server/api_ws.py", line 100, in message_endpoint
    async with create_channel(websocket) as channel:
  File "/home/matt/.pyenv/versions/3.10.4/lib/python3.10/contextlib.py", line 199, in __aenter__
    return await anext(self.gen)
  File "/home/matt/freqtrade/freqtrade/rpc/api_server/ws/channel.py", line 231, in create_channel
    await channel.accept()
  File "/home/matt/freqtrade/freqtrade/rpc/api_server/ws/channel.py", line 130, in accept
    return await self._websocket.accept()
  File "/home/matt/freqtrade/freqtrade/rpc/api_server/ws/proxy.py", line 73, in accept
    return await self._websocket.accept()
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/websockets.py", line 99, in accept
    await self.send(
  File "/home/matt/.pyenv/versions/3.10.4/envs/trade_3104/lib/python3.10/site-packages/starlette/websockets.py", line 87, in send
    raise RuntimeError('Cannot call "send" once a close message has been sent.')
RuntimeError: Cannot call "send" once a close message has been sent.
2022-11-24 07:09:51,066 - DummyDataStrategyTracker - INFO - RAM: 332.82 MB
```

wizrds: So the denying because of invalid token is a little different than a regular endpoint. The actual endpoint function is still run, but when the invalid token is detected it has simply closed the connection already. That's where the RuntimeError comes in from sending an "accept" message when the connection has already sent a "closed" message. This is fixed in the most recent commit [101dec4](https://github.com/freqtrade/freqtrade/pull/7771/commits/101dec461e40c2b8ed15a7075bb4b7dc9099c7b2).

wizrds: Is there a reason that the log message when a websocket is denied `debug` level? I'd imagine it would be best as `info` as users should know when they're getting hit with incorrect tokens. (This very well could've been my fault, I just don't remember if we decided on it already.)

xmatthias: Shouldn't we then check on `if token:`  and skip the whole body if it's not provided?
there seems to be little point in calling `create_channel` if the socket is closed already

Now how you log a denied is ... questionable.
you already get an implict error from uvicorn - so i think we put it to debug as otherwise, we'd have 3 log-lines for one bad connect (connect, denied, (both from uvicorn) and "our denied").

wizrds: I think just marking the channel closed if unable to connect would be the way to go as it keeps things simple and readable, while accomplishing the same thing.

wizrds: > Now how you log a denied is ... questionable. you already get an implict error from uvicorn - so i think we put it to debug as otherwise, we'd have 3 log-lines for one bad connect (connect, denied, (both from uvicorn) and "our denied").

I think in this case we might as well just remove that call altogether.

freqtrade/misc.py
```diff
@@ -301,3 +303,44 @@ def remove_entry_exit_signals(dataframe: pd.DataFrame):
...
+            for item in iter:
+                # This runs outside the event loop thread, so we
+                # must use thread-safe API to talk to the queue.
+                asyncio.run_coroutine_threadsafe(q.put(item), loop).result()
+        except Exception as e:
+            exception = e
+        finally:
+            asyncio.run_coroutine_threadsafe(q.put(_END), loop).result()
+
+    threading.Thread(target=iter_to_queue).start()
```
xmatthias: That's not a good solution if i'm honest, i'd prefer we can remove this - at least for now.

This has the potential to spawn of 100ds of threads without control (or tracking) - in addition to having to spawn one thread per consumer (request - if i'm reading the code right).

So far, it's unclear to me what happens to this thread if the consumer disconnects while it's still sending these messages.
i assume the thread will stall and remain running (though not doing anything) forever (further messages can't be sent, so the generator is never fully consumed) - but it'll clearly need full testing (also abuse testing).
I suspect can also cause the bot to start >100 threads - simply by sending 100 requests for "analyzed_df" - which (given enough requests) - will cause it to crash.
potentially even by "requesting - but instead of waiting for answer,disconnect ... causing the threads to stay around forever.

---

That PR was 95% done - i'd appreciate if we could not extend the scope of PR's.

This can surely be looked at as a separate PR (where the scope of the PR is and remains this exact task/topic) - but in this PR - please remove.

xmatthias: I've reverted this for now for the above reasons.

xmatthias (approved):
Nice work! :100:  (looking forward to the future enhancements we've been discussing).

LGTM :+1:

Issue #7452, opened 2022-09-22
Script Install Failure :: No module named `wheel`


eclipsed4utoo:
## Describe your environment

  * Operating system: MacOS 12.2.1 (M1 Mac)
  * Python Version: 3.8.9 (`python -V`)
  * CCXT version: 1.93.87 (`pip freeze | grep ccxt`)
  * Freqtrade Version: 2022.8 (`freqtrade -V` or `docker-compose run --rm freqtrade -V` for Freqtrade running in docker)
  
Note: All issues other than enhancement requests will be closed without further comment if the above template is deleted or not filled out.

## Describe the problem:

I'm attempting to install the bot and all of the dependencies. I'm going through the script installation from the documentation by running the `./setup.sh -i` command. After running that command, and after the script checks/installs a bunch of dependencies, it fails out when it gets to the `bdist_wheel`.

### Steps to reproduce:

  1. Clean install
  2. Run `./setup.sh -i`
  
### Observed Results:

  * What happened? Script errors
  * What did you expect to happen? The script to not error.

### Relevant code exceptions or logs

Note: Please copy/paste text of the messages, no screenshots of logs please.

  ```
  Collecting progressbar2==4.0.0
  Using cached progressbar2-4.0.0-py2.py3-none-any.whl (26 kB)
Collecting catboost==1.0.6
  Using cached catboost-1.0.6-cp310-none-macosx_10_6_universal2.whl (22.0 MB)
Collecting lightgbm==3.3.2
  Using cached lightgbm-3.3.2.tar.gz (1.5 MB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/private/var/folders/z_/7fjfs5sd20jg1vjw2nt2y6640000gn/T/pip-install-_dm_ygwh/lightgbm_aa381f0549c04c028c5a5a1bc94b6095/setup.py", line 17, in <module>
          from wheel.bdist_wheel import bdist_wheel
      ModuleNotFoundError: No module named 'wheel'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

[...]

PR #7791, opened 2022-11-23
Enable --use-pep517 flag for freqai

aemr3:
## Summary

Installing freqtrade with freqai on Apple M1 chip requires some manual installation procedures. The problematic package is currently lightgbm. This flag fixes this problem. 

Solve the issue: #7452 

## What's new?

Enabled --use-pep517 flag for freqai installation.

## References
https://github.com/pypa/pip/issues/8559

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -82,7 +82,7 @@ function updateenv() {
     dev=$REPLY
     if [[ $REPLY =~ ^[Yy]$ ]]
     then
-        REQUIREMENTS_FREQAI="-r requirements-freqai.txt"
+        REQUIREMENTS_FREQAI="-r requirements-freqai.txt --use-pep517"
     fi
 
     ${PYTHON} -m pip install --upgrade -r ${REQUIREMENTS} ${REQUIREMENTS_HYPEROPT} ${REQUIREMENTS_PLOT} ${REQUIREMENTS_FREQAI}


xmatthias:
Interesting - how come this is only a problem on M1's? 
are there no wheels available for that platform?

what's lightgbm's opinion on this topic (is there an issue for this)? i assume this could be solved by lightgbm changing  something in their setup procedure?

aemr3:
> Interesting - how come this is only a problem on M1's? are there no wheels available for that platform?
> 
> what's lightgbm's opinion on this topic (is there an issue for this)? i assume this could be solved by lightgbm changing something in their setup procedure?

Yeah, there are no wheels for some packages. Here is the list: pandas-ta, SQLAlchemy, TA-Lib, tables, blosc, py_find_1st, python-rapidjson, sdnotify, lightgbm, tornado, docopt, grpcio

On M1 mac these packages are installed from source with their setup script. But lightgbm cannot complete because of this wheel problem. Other packages don't have this problem.

You're right, if lightgbm fixes it we will not have to enable this flag. They are waiting to build and test arm64 wheels on M1 Macs in a free CI service:) See:
https://github.com/microsoft/LightGBM/issues/5328

xmatthias:
OK Got it

> They are waiting to build and test arm64 wheels on M1 Macs in a free CI service:) See:

Which seems pretty lame - considering it's a project that's backed/initiated by microsoft - the owner of 2 free cloud CI services (github actions / azure CI).
it's not something you can help with ... just a generic comment on the topic.

---

i'll trust this will install fine - as i lack an M1 to actually test this.

aemr3:
> i'll trust this will install fine - as i lack an M1 to actually test this.

I have tested this with clearing pip cache and it did work. Maybe we can get another confirmation. There is a conversation going on discord.

robcaulk (approved):
thanks for testing this

PR #7795, opened 2022-11-25
Add date selection arguments to backtest-analysis printout

froggleston:
## Summary

Users may want to select only a subset of trades to print after analysis with `backtesting-analysis`.

## Quick changelog

- add `date_start` and `date_end` arguments to _print_results in entryexitanalysis.py
- tidy up complex function (flake8)

## What's new?

- two simple arguments to the print_results function to allow subsetting of the main backtesting-analysis dataframe by start and/or end date


freqtrade/data/entryexitanalysis.py
```diff
@@ -152,25 +152,44 @@ def _do_group_table_output(bigdf, glist):
...
+        df = df.loc[(df['exit_reason'].isin(exit_reason_list))]
+
+    return df
+
+
 def _print_results(analysed_trades, stratname, analysis_groups,
                    enter_reason_list, exit_reason_list,
-                   indicator_list, columns=None):
+                   indicator_list, columns=None,
+                   date_start=None, date_end=None):
```
xmatthias: I don't quite see how this will satisfy the first point of "a user wants to select a subset"...

it's a private method which is not exposed to the user (and should not be used directly).
maybe you forgot to commit the parts where the user can provide his selection via `freqtrade backtesting-analysis` ?

froggleston: Yup this was the first step for the notebook update (so i can use the actual FT code instead of my buy_reasons script still). I'll add the user selection in the next PR. I'll be publishing a version of the notebook that others can use in the next few days!

xmatthias: i don't see why we need this _as is_. users can't use it.
it's adding parameters to an internal method (it starts with a `_`) - which are not used.

Please add the user interface part here, too, so we can test and look at it all at once.

also, if you consider this as a public facing method - then the name of the method is clearly wrong.
it's intended use is within this file, not to be imported someplace else (both the naming, and the `_` are clear signs of that).

froggleston:
OK I've implemented the rest of the changes to allow users to specify dates, and added a test. I also refactored the _print_results private function into two public functions. 

I also simplified the `process_entry_exit_reasons(config: Config):` to take a config object instead of the increasingly numerous individual arguments.

Although now you've made me wonder if all these functions should just be public.....

freqtrade/commands/cli_options.py
```diff
@@ -658,6 +658,16 @@ def __init__(self, *args, **kwargs):
...
+    "analysis_date_start": Arg(
+        "--analysis-date-start",
+        help=("Start date to filter trades for analysis (inclusive). "
+              "e.g. '20220101'"),
+    ),
+    "analysis_date_end": Arg(
+        "--analysis-date-end",
+        help=("End date to filter trades for analysis (exclusive). "
+              "e.g. '20220131'"),
+    ),
```
xmatthias: is there any reason we can't stick with what we got (`--timerange`) - but need new arguments that are exclusive to this (but seemingly do the same exact thing, with almost the identical syntax).

froggleston: Erm, good question 😄

froggleston: OK, done. Thanks - not sure why I didn't think of that!

freqtrade/configuration/configuration.py
```diff
@@ -462,6 +462,9 @@ def _process_analyze_options(self, config: Config) -> None:
         self._args_to_config(config, argname='indicator_list',
                              logstring='Analysis indicator list: {}')
 
+        self._args_to_config(config, argname='timerange',
+                             logstring='Filter trades by timerange: {}')
+
```
xmatthias, line 467: The timerange parameter is already present - no need to add it again - it'll cause duplicated log messages...) - as well as having no reason

docs/utils.md
```diff
@@ -744,6 +746,12 @@ optional arguments:
   --indicator-list INDICATOR_LIST [INDICATOR_LIST ...]
                         Comma separated list of indicators to analyse. e.g.
                         'close,rsi,bb_lowerband,profit_abs'
+  --analysis-date-start YYYYMMDD
+                        Start date to filter trades for analysis (inclusive). e.g.
+                        20220101
+  --analysis-date-end YYYYMMDD
+                        End date to filter trades for analysis (exclusive). e.g.
+                        20220131
```
xmatthias: might need a refresh?

froggleston: Ah good spot. Done.

freqtrade/data/entryexitanalysis.py
```diff
@@ -152,37 +153,55 @@ def _do_group_table_output(bigdf, glist):
...
+        res_df = _select_rows_by_tags(res_df, enter_reason_list, exit_reason_list)
 
-        if enter_reason_list and "all" not in enter_reason_list:
-            bigdf = bigdf.loc[(bigdf['enter_reason'].isin(enter_reason_list))]
+    return res_df
 
-        if exit_reason_list and "all" not in exit_reason_list:
-            bigdf = bigdf.loc[(bigdf['exit_reason'].isin(exit_reason_list))]
+
+def print_results(res_df, analysis_groups, indicator_list):
```
xmatthias, line 189: One final note - if you consider this function as public interface - we should consider renaming it to something more telling.
if you import it in a notebook, you'll no longer remember what "print_results" was 3 cells down.

if instead, it's called something like "print_bt_analysis_result" or similar - it'll be much clearer.

I'll leave that however for the future...

xmatthias (approved):
Nice enhancement :100: 

LGTM :+1:

freqtrade/data/entryexitanalysis.py
froggleston: Ah yeah that's a good idea. I'll wrap that into the next update.

PR #7799, opened 2022-11-26
Use default `extra_returns_per_train`

robcaulk:
The initial build of `extra_returns_per_train` were set to 0. However, users may have default values set in their configs - and therefore we ensure those default values are set during the initial build of `dd.historic_predictions`


xmatthias (approved):
LGTM :+1:

PR #7801, opened 2022-11-26
Fix 4ac update_total_profit

initrv:
## Summary
_update_total_profit() must be executed before "self._position = Positions.Neutral" because _update_total_profit() calls get_unrealized_profit(), which returns 0 if position is already neutral and total_profit is not updated


robcaulk (approved):
Good catch!

PR #7807, opened 2022-11-27
[WIP] Add tensorflow support to FreqAI

robcaulk:
This PR aims to provide a standard `tensorflow` example for users to build on. In particular, the proposed example neural network is a convolutional neural network. 

To-do:

- [ ]  documentation
- [ ]  beta testing 
- [ ]  debug multiple targets
- [ ]  add tests

If you want to help beta-test/develop this feature, please feel free to join us in the FreqAI discord https://discord.gg/3SRkynuMkw


robcaulk:
This MR is superceded quite effectively by our new PyTorch module. 

While Tensorflow remains easily added as shown in this MR, I don't think it is a good idea right now to support it in the source code the same way we do with the PyTorch module. There are a variety of reasons for this:

- Tensorflow adds yet *more* dependencies which means more install options, more docker image variations, more test if statement flags,
- Tensorflow dependencies can conflict negatively with torch dependencies. It is typically better to orient dependencies toward one or the other, as opposed to having both installed simultaneously
- The above two issues might be worth the effort in a case where Tensorflow had a strong usecase - but unfortunately it is becoming much less useful for hobby level ML (which is the main user-base of FT + FreqAI). PyTorch is leagues beyond TF ins terms of usability, readability, documentation, debugability, deployability etc etc.

For these reasons, I am closing this MR. If users wish to use TF, it should be easy enough for them to import it into their custom IFreqaiModel, as shown in this MR. We simply do not aim to give TF first class treatment right now. This may change in the future if TF has a large performance gap or provides better functionality for something important in FreqAI.

PR #7808, opened 2022-11-27
Fix custom reward function link

aemr3:
## Summary

Fixed custom reward function links in the RL doc. The links were incorrect.
Fixed some typos.


robcaulk (approved):
Thanks for that!

PR #7809, opened 2022-11-27
Improve the RL learning process

richardjozsa:
Improve the RL learning process by selecting random start point for the agent, it can help to block the agent to only learn on the selected period of time, while improving the quality of the model.


robcaulk (approved):
LGTM thanks!

PR #7810, opened 2022-11-28
Add more calculations for backtest metrics

### LINKED ISSUE #1907: Enhancement - Add Sharpe Ratio for strategies backtesting

stash86:
## Summary

Add more calculations for backtest metrics

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

Add expectancy, sharpe, sortino, and calmar calcs to be used in backtest report table, like this (which will be a separate PR later)
```
================== SUMMARY METRICS ==================
| Metric                      | Value               |
|-----------------------------+---------------------|
| Backtesting from            | 2021-09-01 00:00:00 |
| Backtesting to              | 2022-11-22 00:00:00 |
| Max open trades             | 4                   |
|                             |                     |
| Total/Daily Avg Trades      | 45 / 0.1            |
| Starting balance            | 190 BUSD            |
| Final balance               | 292.386 BUSD        |
| Absolute profit             | 102.386 BUSD        |
| Total profit %              | 53.89%              |
| CAGR %                      | 42.19%              |
| Sortino                     | 5.22                |
| Sharpe                      | 1.88                |
| Calmar                      | 152.39              |
| Profit factor               | 7.99                |
| Expectancy                  | 1.24                |
| Trades per day              | 0.1                 |
| Avg. daily profit %         | 0.12%               |
| Avg. stake amount           | 44.854 BUSD         |
| Total trade volume          | 2018.441 BUSD       |

```


xmatthias (changes requested):
This is great i think.

what we'll definitely want is CI passing (seems to be only flake8, and maybe mypy??), as well as adding the new outputs to the `backtesting.md` documentation page.

In the future, i think we can extract these calculations from the hyperopt-loss functions - but for now i don't think that's something we'll have to do here.

mripani (commented):
This looks good, I'm wondering about the risk free rate crypto approach. Because in theory in the Sharpe Ratio formula the returns are adjusted by the Risk Free rate.

froggleston:
> This looks good, I'm wondering about the risk free rate crypto approach. Because in theory in the Sharpe Ratio formula the returns are adjusted by the Risk Free rate.

IMO this can be solved simply by adding documentation at this stage. Adding in support for adjust rates is a larger piece of work and could likely be done after this PR.

xmatthias:
> which will be a separate PR later

i only realized now that you intend to do the actual using of the methods later.

I don't think that's a good idea - as it'll mean we'll have to accept this "blind".
There's no tests for it - and it's not used anywhere.

Effectively this means that we need to trust our eyes on the review that the code is correct - without the possibility to run the code in a freqtrade context (neither in tests, nor via a command).

Obviously it's possible to run the methods manually, but as the "use from backtesting" is almost certainly a 3-4 line addition (per metric) - i think that should be done as part of this PR.

That way, we can also make sure that calculation "quirks"  (like the risk-free approach) is actually documented.

freqtrade/data/metrics.py
```diff
@@ -190,3 +192,131 @@ def calculate_cagr(days_passed: int, starting_balance: float, final_balance: flo
...
+                      min_date: datetime, max_date: datetime) -> float:
+    """
+    Calculate sortino
+    :param trades: DataFrame containing trades (requires columns profit_ratio)
+    :return: sortino
+    """
+    if (len(trades) == 0) or (min_date is None) or (max_date is None) or (min_date == max_date):
+        return 0
+
+    total_profit = trades["profit_ratio"]
```
xmatthias: This is literally the worst column to use for "any" calculation that includes a ".sum()".

The main problem is that it's completely ignoring position sizing - so a 5% win will always weight in equally - even if one was on a 10$ stake, and the other on a 1000$ stake.

I realize that the hyperopt-loss functions do the same thing - but with dynamic positions - it's simply wrong (also in loss functions).
Obviously this applies not just to this line - but to all functions that use this column.

to the total profit "properly", you can for example use `results['profit_abs'].sum() / start_balance` - which is calculated as `profit_total` in backtesting.

xmatthias: i've done this for calmar now to demonstrate what i mean

stash86: done

stash86:
> > which will be a separate PR later
> 
> i only realized now that you intend to do the actual using of the methods later.
> 
> I don't think that's a good idea - as it'll mean we'll have to accept this "blind". There's no tests for it - and it's not used anywhere.
> 
> Effectively this means that we need to trust our eyes on the review that the code is correct - without the possibility to run the code in a freqtrade context (neither in tests, nor via a command).
> 
> Obviously it's possible to run the methods manually, but as the "use from backtesting" is almost certainly a 3-4 line addition (per metric) - i think that should be done as part of this PR.
> 
> That way, we can also make sure that calculation "quirks" (like the risk-free approach) is actually documented.

done on optimize_report. So now you would see the new metrics like the one on the screenshot

freqtrade/data/metrics.py
```diff
@@ -190,3 +192,121 @@ def calculate_cagr(days_passed: int, starting_balance: float, final_balance: flo
...
+        return 0
+
+    total_profit = trades['profit_abs'] / starting_balance
+    days_period = max(1, (max_date - min_date).days)
+
+    expected_returns_mean = total_profit.sum() / days_period
+
+    trades['downside_returns'] = 0
+    trades.loc[total_profit < 0, 'downside_returns'] = (trades['profit_abs'] / starting_balance)
+    down_stdev = np.std(trades['downside_returns'])
```
xmatthias: looking at this more closely i think this is actually wrong.

the std-dev is calculated against all trades (say 100) - assuming a 50:50 win/loss rate - we'd get 100 rows - with 50x 0 - and the others with "whatever their negative value".
That's making the std-dev appear smaller than it really is.

i've pushed a fix for this ...

xmatthias (approved):
LGTM :+1:

PR #7823, opened 2022-11-28
Set model_type in base RL model

aemr3:
## Summary

This PR aims to fix an error when user tries to use a custom model saved in freqaimodels folder. Currently it is assumed that custom model file name includes "Reinforcement". But if the user chooses to use "MyCoolRLModel", as instructed on docs, model saving and loading will not work.


robcaulk (approved):
That is a much better way to do it...

PR #7828, opened 2022-11-30
bring back market side setting in get_state_info

robcaulk:
Market side was missing from the `get_state_info()` in RL. This MR brings it back.


xmatthias (approved):
LGTM :+1:

PR #7833, opened 2022-11-30
binance restricted locations and server location

gaugau3000:
## What's new?
Explain in the doc that server location in binance restricted country will lead to API error.
Cf : https://github.com/ccxt/ccxt/issues/15872

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -54,6 +54,9 @@ This configuration enables kraken, as well as rate-limiting to avoid bans from t
 
 ## Binance
 
+!!! Warning "Server location and geo-ip restrictions"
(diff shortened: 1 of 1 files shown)


xmatthias:
it's quite telling that CI has exactly this problem on this particular PR.

Github actions CI runs on us Azure servers - what's interesting though is that binance seems to be blocking these requests (only to the futures API) only during the US night, but not all day long.
if it'd be regulations enforced, it would have to be blocked all day long ...

stash86:
it's now blocking spot as well, and for some IPs, the block is permanent. 3 out of 5 of my VPSes located in USA has been permanently blocked. All on spot market

xmatthias (approved):
LGTM :+1:

Issue #6402, opened 2022-02-13
Cannot open config.json with Docker and Jupyter Lab


gabrieleiacono:
## Describe your environment

  * Operating system: Debian 5.15.15-2 (2022-01-30) x86_64
  * Python Version: Python 3.9.9
  * CCXT version: 1.68.20
  * Freqtrade Version: freqtrade 2022.1
  
## Your question
I can't open any config.json using Configuration.from_files() from inside a Jupyter intance that runs inside a container (created as described [here](https://www.freqtrade.io/en/stable/docker_quickstart/#data-analysis-using-docker-compose)) with the user_data folder mounted as a volume. The volume is mounted correctly, but from_files() cannot open them throwing the OperationalException reported below.

Steps reproduce this behaviour:
1. Download the docker-compose file from the repository inside a new folder (let's call it ft_userdata)
2. `docker-compose pull`
3. `docker-compose run --rm freqtrade create-userdir --userdir user_data`
4. `docker-compose run --rm freqtrade new-config --config user_data/config.json`
5. Get [docker-compose-jupyter.yml](https://github.com/freqtrade/freqtrade/blob/develop/docker/docker-compose-jupyter.yml) and [Dockerfile.jupyter](https://github.com/freqtrade/freqtrade/blob/develop/docker/Dockerfile.jupyter) inside ft_userdata (same level of docker-compose.yml)
6. `docker-compose -f docker-compose-jupyter.yml up`
7. Open JupyterLab and create a new Python notebook in the root folder 
8. Create a cell and change the pwd with `%cd user_data`. Make sure that the config.json created during step 4 is present
9. Try to load the config.json by creating another cell with the following content:
```python
from freqtrade.configuration import Configuration

config = Configuration.from_files(["config.json"])
```
 

### Relevant code exceptions or logs

  ```
chown: cannot access '/freqtrade/user_data/user_data': No such file or directory

---------------------------------------------------------------------------
OperationalException                      Traceback (most recent call last)
Input In [2], in <module>
      2 from freqtrade.configuration import Configuration
      4 # Customize these according to your needs.
      5 
      6 # Initialize empty configuration object
      7 # config = Configuration.from_files([])
      8 # Optionally, use existing configuration file
----> 9 config = Configuration.from_files(["config.json"])
     11 # Define some constants
     12 # config["timeframe"] = "5m"
     13 # # Name of the strategy class
     14 # config["strategy"] = "SampleStrategy"
     15 # Location of the data
[...]

PR #7835, opened 2022-12-01
update strategy_analysis_example.ipynb

rzhb:
* Add download-data command
* Change directory
* Fix relative config path #6402
* Use default data-format

### DIFF
--- a/freqtrade/templates/strategy_analysis_example.ipynb
+++ b/freqtrade/templates/strategy_analysis_example.ipynb
@@ -7,7 +7,7 @@
     "# Strategy analysis example\n",
     "\n",
     "Debugging a strategy can be time-consuming. Freqtrade offers helper functions to visualize raw data.\n",
-    "The following assumes you work with SampleStrategy, data for 5m timeframe from Binance and have downloaded them into the data directory in the default location."
+    "The following assumes you work with SampleStrategy, data for 5m timeframe from Binance and have downloaded them into the data directory in the default location, using command like `freqtrade download-data --exchange binance --trading-mod spot --pairs BTC/USDT --days 7 -t 5m`."
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


freqtrade/templates/strategy_analysis_example.ipynb
```diff
@@ -23,22 +23,36 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "import os\n",
     "from pathlib import Path\n",
+    "\n",
+    "# Change current working directory from `somedir/freqtrade/user_data/notebooks` to project root `somedir/freqtrade`, so relative paths remain consistent.\n",
+    "if not Path(\"LICENSE\").is_file():\n",
+    "    os.chdir(\"../../\")\n",
```
xmatthias: I'm not sure why this is necessary - i think it hints at a wrong jupyter startup location rather than something that should be done in the example notebook.

the assumption is that jupyter (or jupyter lab) is started at the repository's root - which contains `user_data/` - in which case the above is wrong (or at least unnecessary).

i don't really see why we should force users to change directory (to a potentially non-existing directory).
if a non-standard startup location is chosen, i'd assume that the user also knows how to change directory.

This code is also failing if it's starting from within user_data - as then it'd go up one level too much. As such - i don't currently see this as something we should be adding to the notebook example.

rzhb: This is already somewhere in the freqtrade document, but not convenient, https://www.freqtrade.io/en/stable/data-analysis/#change-directory-to-root

> the assumption is that jupyter (or jupyter lab) is started at the repository's root

As I can see, Jupyter start at the notebook file location by default, which is `freqtrade/user_data/notebooks`, not the repository' root.

xmatthias (approved):
I've modified this a bit - but LGTM now :+1:

PR #7837, opened 2022-12-01
freqAI backtesting - Perfomance improvement

wagnercosta:
## Summary

This PR brings performance improvements when backtesting freqAI through saved prediction files. The improvement is significant when there is a high number of reference models.

Example: (413 models = Backtesting freqAI loop reduced from 82.59604382514954 seconds to 2.093254804611206 seconds)


robcaulk (approved):
LGTM

PR #7842, opened 2022-12-02
Change to broadcasting single candles in Producer/Consumer

wizrds:
## Summary

This PR re-implements the logic required for changing the Producer to only broadcast the most recent candle when calculated as opposed to the entire Dataframe. This reduces the message sizes significantly and helps increase network performance, all while keeping the same resource usage on the Consumer.

## Quick changelog

- Re-implement EMC using new WebSocketChannel API's
- Add methods to DataProvider for appending candles to Producer data
- Update `analyzed_df` request on Producer side to allow specifying the pair
- Add mechanism for EMC to request data from Producers when needed

### DIFF
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -7,7 +7,7 @@ Common Interface for bot and strategy to access data.
 import logging
 from collections import deque
 from datetime import datetime, timezone
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple
 
 from pandas import DataFrame, concat
 
@@ -165,7 +165,7 @@ class DataProvider:
         timeframe: str,
         candle_type: CandleType,
         producer_name: str = "default"
-    ) -> Union[bool, int]:
+    ) -> Tuple[bool, int]:
         """
         Append a candle to the existing external dataframe
 
@@ -179,22 +179,22 @@ class DataProvider:
         if producer_name not in self.__producer_pairs_df:
             # We don't have data from this producer yet,
             # so we can't append a candle
-            return False
+            return (False, 0)
 
         if pair_key not in self.__producer_pairs_df[producer_name]:
             # We don't have data for this pair_key,
             # so we can't append a candle
-            return False
+            return (False, 0)
 
         # CHECK FOR MISSING CANDLES
-        # return int
+        # return (False, int > 0)
 
         existing_df, _ = self.__producer_pairs_df[producer_name][pair_key]
         appended_df = self._append_candle_to_dataframe(existing_df, dataframe)
 
         # Everything is good, we appended
         self.__producer_pairs_df[producer_name][pair_key] = appended_df, last_analyzed
-        return True
+        return (True, 0)
 
     def _append_candle_to_dataframe(self, existing: DataFrame, new: DataFrame) -> DataFrame:
         """
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -9,7 +9,7 @@ from collections import deque
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Tuple
 
-from pandas import DataFrame, concat
+from pandas import DataFrame, concat, to_timedelta
 
 from freqtrade.configuration import TimeRange
 from freqtrade.constants import Config, ListPairsWithTimeframes, PairWithTimeframe
@@ -176,24 +176,30 @@ class DataProvider:
         """
         pair_key = (pair, timeframe, candle_type)
 
-        if producer_name not in self.__producer_pairs_df:
+        if (producer_name not in self.__producer_pairs_df) \
+           or (pair_key not in self.__producer_pairs_df[producer_name]):
             # We don't have data from this producer yet,
-            # so we can't append a candle
-            return (False, 999)
-
-        if pair_key not in self.__producer_pairs_df[producer_name]:
-            # We don't have data for this pair_key,
-            # so we can't append a candle
-            return (False, 999)
-
-        # CHECK FOR MISSING CANDLES
-        # Calculate difference between last candle in local dataframe
-        # and first candle in incoming dataframe. Take difference and divide
-        # by timeframe to find out how many candles we still need. If 1
-        # then the incoming candle is the right candle. If more than 1,
-        # return (False, missing candles - 1)
+            # sor we don't have data for this pair_key
+            # return False and 1000 for the full df
+            return (False, 1000)
 
         existing_df, _ = self.__producer_pairs_df[producer_name][pair_key]
+
+        # CHECK FOR MISSING CANDLES
+        timeframe_delta = to_timedelta(timeframe)  # Convert the timeframe to a timedelta for pandas
+        local_last = existing_df.iloc[-1]['date']  # We want the last date from our copy of data
+        incoming_first = dataframe.iloc[0]['date']  # We want the first date from the incoming data
+
+        candle_difference = (incoming_first - local_last) / timeframe_delta
+
+        # If the difference divided by the timeframe is 1, then this
+        # is the candle we want and the incoming data isn't missing any.
+        # If the candle_difference is more than 1, that means
+        # we missed some candles between our data and the incoming
+        # so return False and candle_difference.
+        if candle_difference > 1:
+            return (False, candle_difference)
+
         appended_df = self._append_candle_to_dataframe(existing_df, dataframe)
 
         # Everything is good, we appended
@@ -212,7 +218,7 @@ class DataProvider:
             existing = concat([existing, new])
 
         # Only keep the last 1500 candles in memory
-        existing = existing[-1500:] if len(existing) > 1000 else existing
+        existing = existing[-1500:] if len(existing) > 1500 else existing
 
         return existing
 
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -17,6 +17,7 @@ from freqtrade.data.history import load_pair_history
 from freqtrade.enums import CandleType, RPCMessageType, RunMode
 from freqtrade.exceptions import ExchangeError, OperationalException
 from freqtrade.exchange import Exchange, timeframe_to_seconds
+from freqtrade.misc import append_candles_to_dataframe
 from freqtrade.rpc import RPCManager
 from freqtrade.util import PeriodicCache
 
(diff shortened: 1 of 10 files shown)


freqtrade/rpc/external_message_consumer.py
```diff
@@ -336,16 +366,58 @@ def _consume_analyzed_df_message(self, producer_name: str, message: WSMessageSch
...
 
-        # Add the dataframe to the dataprovider
-        self._dp._add_external_df(pair, df,
-                                  last_analyzed=la,
-                                  timeframe=timeframe,
-                                  candle_type=candle_type,
-                                  producer_name=producer_name)
+        logger.debug(f"Received {len(df)} candle(s) for {key}")
+
+        if len(df) >= 999:
```
xmatthias: This is clearly wrong.

a dataframe from OKX (for example) will have at max. 300 rows - as they don't provide more history in one request. 

if we assume we get more than 1 candles in a partial dataframe, then a  safe value for this request would be 100 (obviously we can't request more than 100 partial candles, then).

freqtrade/rpc/rpc.py
```diff
@@ -1058,23 +1058,43 @@ def _rpc_analysed_dataframe(self, pair: str, timeframe: str,
...
                                                pair, timeframe, _data, last_analyzed)
 
-    def __rpc_analysed_dataframe_raw(self, pair: str, timeframe: str,
-                                     limit: Optional[int]) -> Tuple[DataFrame, datetime]:
-        """ Get the dataframe and last analyze from the dataprovider """
+    def __rpc_analysed_dataframe_raw(
+        self,
+        pair: str,
+        timeframe: str,
+        limit: Optional[int] = None
```
xmatthias: Pretty pointless default if all calls use it with the parameter...

freqtrade/rpc/external_message_consumer.py
```diff
@@ -336,16 +366,58 @@ def _consume_analyzed_df_message(self, producer_name: str, message: WSMessageSch
...
+                pair,
+                df,
+                last_analyzed=la,
+                timeframe=timeframe,
+                candle_type=candle_type,
+                producer_name=producer_name
+            )
+
+            if not did_append:
+                logger.debug("Holes in data or no existing df, "
```
xmatthias: I think we should log this in info or warning mode.

i understand you made this for debugging purposes - but it's a pretty important info - as it can hint at "your producer is too slow, and we're gonna make that worse by requesting it all again" ...

freqtrade/data/dataprovider.py
```diff
@@ -150,6 +150,80 @@ def _add_external_df(
...
+        Append a candle to the existing external dataframe
+
+        :param pair: pair to get the data for
+        :param timeframe: Timeframe to get data for
+        :param candle_type: Any of the enum CandleType (must match trading mode!)
+        :returns: False if the candle could not be appended, or the int number of missing candles.
+        """
+        pair_key = (pair, timeframe, candle_type)
+
+        if (producer_name not in self.__producer_pairs_df) \
```
xmatthias: I'd prefer to use parentheses `()` instead of `\` to wrap lines in if statements.
`\` is pretty easy to mess up unintentionally if you switch between environments (windows/linux lineendings ... )

freqtrade/data/dataprovider.py
```diff
@@ -150,6 +150,80 @@ def _add_external_df(
...
+            # or we don't have data for this pair_key
+            # return False and 1000 for the full df
+            return (False, 1000)
+
+        existing_df, _ = self.__producer_pairs_df[producer_name][pair_key]
+
+        # CHECK FOR MISSING CANDLES
+        timeframe_delta = to_timedelta(timeframe)  # Convert the timeframe to a timedelta for pandas
+        local_last = existing_df.iloc[-1]['date']  # We want the last date from our copy of data
+        incoming_first = dataframe.iloc[0]['date']  # We want the first date from the incoming data
```
xmatthias: we assume the dataframe has at least one row ... maybe should be mentioned in the docstring

freqtrade/data/dataprovider.py
```diff
@@ -150,6 +150,80 @@ def _add_external_df(
...
+
+        # We have received this candle before, update our copy
+        # and return True, 0
+        if local_last == incoming_first:
+            existing_df.iloc[-1] = dataframe.iloc[0]
+            existing_data = (existing_df.reset_index(drop=True), _)
```
xmatthias: we actually use `_` here - can we please give it a name? (`_` can easily be overwritten accidentally, as it's common python convention to use this for things we don't use).

freqtrade/data/dataprovider.py
```diff
@@ -150,6 +150,80 @@ def _add_external_df(
...
+
+        :param existing: The full dataframe you want appended to
+        :param new: The new dataframe containing the data you want appended
+        :returns: The dataframe with the new data in it
+        """
+        if existing.iloc[-1]['date'] != new.iloc[-1]['date']:
+            existing = concat([existing, new])
+
+        # Only keep the last 1500 candles in memory
+        existing = existing[-1500:] if len(existing) > 1500 else existing
```
xmatthias: You must reindex after this - otherwise the index will start floating from 1-1501, 2-1502, ... - as it's not reset again.

this has shown to create chaos / crashes in some indicators (haikin-ashi for one) ...

freqtrade/rpc/external_message_consumer.py
```diff
@@ -336,16 +366,58 @@ def _consume_analyzed_df_message(self, producer_name: str, message: WSMessageSch
...
+                             f"requesting {n_missing} candles "
+                             f"for {key} from `{producer_name}`")
+
+                self.send_producer_request(
+                    producer_name,
+                    WSAnalyzedDFRequest(
+                        data={
+                            "limit": n_missing,
```
xmatthias: I'm ... not entirely sure this is a good idea - i'd rather go the "full DF" route - at least for now.

We assume we only hit this if there was some outage - or if the producer was too slow. Based on this - you could get the 1m candle at :59.999 (instead of at :01). by the time we get to request a new DF (or by the time the producer get's to work on this request) - we might be missing one additional candle - so we might end up in a request/re-request loop - always falling 1 candle short - but might never actually get to a "now we have it all".

IF we take this approach - i think you'll want to request 1 overlapping candle - which will then have to be replaced (we don't know if the candle changed mid-candle ... ), which will obviously make the logic in dataprovider a bit more complex (you would now have to replace the last candle - but append subsequent candles).

wizrds: Per our conversation over Discord, we've decided to go the route of updating the `_add_external_candle` method to support overlapping dataframes. Also, always requesting `n_missing + 1`.

freqtrade/data/dataprovider.py
```diff
@@ -150,6 +151,80 @@ def _add_external_df(
...
+            )
+        ]
+        new_candles = dataframe[
+            ~dataframe['date'].isin(
+                existing_df['date']
+            )
+        ]
+
+        if overlapping_candles:
+            existing_df = concat([old_candles, overlapping_candles], axis=0)
```
xmatthias: i think instead of filtering 3 times (and then concatting twice, once here and once in the other method) - it would be much simpler (and safer to some degree) to do something like this:

```
# Drop all duplicate candles
existing_df1 = existing_df[existing_df['date'] < incomming_first]
...
```

If the above modifies the existing dataframe - then there was an overlapping candle, so we technically can never get a candle difference.
if we get a difference - this also automatically implies that the dataframe was not modified - then we obviously return and request more.

After that, a simple concat will suffice (we removed all duplicates first, so we won't create duplicates) (obviously followed by a limit to the 1500 candles).

---

the current method has the downside that, in theory, it could drop canldes from the "middle". 
We assume that ain't going to happen as we control the producer - but we don't really - so with "in" filtering, we could end up with an unsorted dataframe (if the incomming dataframe itself had leaks).
It's also 2 more pandas operations than necessary.

freqtrade/data/dataprovider.py
```diff
@@ -150,6 +151,80 @@ def _add_external_df(
         self.__producer_pairs_df[producer_name][pair_key] = (dataframe, _last_analyzed)
         logger.debug(f"External DataFrame for {pair_key} from {producer_name} added.")
 
+    def _add_external_candle(
+        self,
```
xmatthias: I think that we'll want to have some good, explicit tests for this method.

We knew already that it will contain some slightly more complex logic - and (and that's the main thing) - it's not something where edge-cases are easily testable as it's dependant on upstream / out of process data.

freqtrade/data/dataprovider.py
xmatthias: I've implemented the changes i mentioned above - as the current method was failing with several errors (just checkout 5811363 to see these errors).

based on the tests - i think the current method will work just fine - though i'm not sure if this shouldn't be the only entrypoint for emc.
we do have "is full dataframe" handling here, too - so it's kinda redundant to move this decision to emc.

wizrds: Your changes LGTM, and I appreciate the thorough test you've added.

> based on the tests - i think the current method will work just fine - though i'm not sure if this shouldn't be the only entrypoint for emc. we do have "is full dataframe" handling here, too - so it's kinda redundant to move this decision to emc.

I'm not sure I follow. What do you mean the only entrypoint for emc? Are you saying any `analyzed_df` message should just call the `_add_external_candle` method no matter the size?

xmatthias: it's a thought so far, nothing i investigated deeper.

in the end, we don't need 2 entry-points to add an external dataframe. 
we can take the decision "is this a complete one" - or an "aggregatable one" within this function - which will simplify the logic a lot (right now we do this decision twice, once in EMC to call `add_external_df()` - and once within this method - if the resulting, truncated dataframe was empty.

wizrds: I'm not sure I agree. To me, the logic for deciding what to do with messages should stay in the EMC.

xmatthias: then this whole function is misplaced where it is now.

Part of this decision is already here - but we now take this decision twice, which means it's splitting the logic between this file and emc - where an initial decision is taken

freqtrade/rpc/external_message_consumer.py
```diff
@@ -411,8 +411,7 @@ def _consume_analyzed_df_message(self, producer_name: str, message: WSMessageSch
                 # Set to None for all candles if we missed a full df's worth of candles
                 n_missing = n_missing if n_missing < FULL_DATAFRAME_THRESHOLD else 1500
 
-                logger.warning("Holes in data or no existing df, "
-                               f"requesting {n_missing} candles "
+                logger.warning("Holes in data or no existing df, requesting {n_missing} candles "
```
wizrds: Shouldn't the warning content be an f string?

wizrds: Never mind, I see now that it was fixed

wizrds:
LGTM :+1:

xmatthias (approved):
Great work! 

LGTM :+1:

PR #7843, opened 2022-12-03
freqai RL agent info during training

smarmau:
## Summary

Adds metrics to RL training within freqai to the console during callbacks

Original author of code @stm on discord 

Example:

```-----------------------------------------
| _actions/               |             |
|    Hold                 | 4           |
|    Long_enter           | 10          |
|    Long_exit            | 9           |
|    Neutral              | 2           |
|    Short_enter          | 13          |
|    Short_exit           | 13          |
|    _Invalid             | 71          |
|    _Unknown             | 36          |
|    action               | 1           |
| _reward/                |             |
|    mean_trade_duration  | 2.55        |
|    total_profit         | 0.988       |
|    total_reward         | 427         |
| _state/                 |             |
|    current_profit_pct   | 5.59e-05    |
|    position             | 1           |
|    trade_duration       | 8           |
| rollout/                |             |
|    ep_len_mean          | 1.51e+03    |
|    ep_rew_mean          | 3.83e+03    |
| time/                   |             |
|    fps                  | 61          |
|    iterations           | 13          |
|    time_elapsed         | 27          |
|    total_timesteps      | 1664        |
| train/                  |             |
|    approx_kl            | 0.006526591 |
|    clip_fraction        | 0.0102      |
|    clip_range           | 0.2         |
|    entropy_loss         | -1.54       |
|    explained_variance   | 0.116       |
|    learning_rate        | 0.00025     |
|    loss                 | 94.3        |
|    n_updates            | 120         |
|    policy_gradient_loss | -0.00613    |
|    value_loss           | 208         |
-----------------------------------------

### DIFF
--- a/freqtrade/freqai/RL/BaseReinforcementLearningModel.py
+++ b/freqtrade/freqai/RL/BaseReinforcementLearningModel.py
@@ -13,9 +13,11 @@ import torch as th
 import torch.multiprocessing
 from pandas import DataFrame
 from stable_baselines3.common.callbacks import EvalCallback
+from stable_baselines3.common.callbacks import BaseCallback
 from stable_baselines3.common.monitor import Monitor
 from stable_baselines3.common.utils import set_random_seed
 from stable_baselines3.common.vec_env import SubprocVecEnv
+from stable_baselines3.common.logger import HParam
 
 from freqtrade.exceptions import OperationalException
 from freqtrade.freqai.data_kitchen import FreqaiDataKitchen
@@ -155,6 +157,8 @@ class BaseReinforcementLearningModel(IFreqaiModel):
         self.eval_callback = EvalCallback(self.eval_env, deterministic=True,
                                           render=False, eval_freq=len(train_df),
                                           best_model_save_path=str(dk.data_path))
+        
+        self.tensorboard_callback = TensorboardCallback()
 
     @abstractmethod
     def fit(self, data_dictionary: Dict[str, Any], dk: FreqaiDataKitchen, **kwargs):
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by robcaulk — CHANGES_REQUESTED** [review]
I like the idea, but the implementation has some blocking points:

* the hard coded `hparam_dict` has parameters which are unique to some policies (PPO) and non existent for others (DQN). Since we support the user selecting between all these various policies in the configuration file, we need this to be equally flexible, otherwise it will break for all the policies it was not designed for.

One solution would be to reduce the parameters to only params shared between all policies, then define this barebones dict(`hparam_dict`) at the `ReinforcementLearner` level so users can see how they can add additional pieces of information if they want. 

Another issue is the typing is really difficult to do (rooting from the same issue as described above). We could certainly `type: ignore` all of it, but it becomes hard to debug and carry around in the future. I'll put some effort in to see what the primitives are and see if we can get some standardized typing going in this thing.

File: freqtrade/freqai/RL/BaseReinforcementLearningModel.py [hunk 1]
```diff
@@ -156,6 +157,8 @@ def set_train_and_eval_environments(self, data_dictionary: Dict[str, DataFrame],
                                           render=False, eval_freq=len(train_df),
                                           best_model_save_path=str(dk.data_path))
 
+        self.tensorboard_callback = TensorboardCallback()
```
robcaulk: This can go inside the `BaseReinforcementLearner.__init__()` so that we don't need to replicate it throughout.

freqtrade/freqai/RL/BaseReinforcementLearningModel.py
```diff
@@ -398,3 +401,51 @@ def _init() -> gym.Env:
         return env
     set_random_seed(seed)
     return _init
+
+
+class TensorboardCallback(BaseCallback):
```
robcaulk: Might be best to move this to it's own file - keeping multiple classes in a single file starts to get disorganized quickly.

freqtrade/freqai/prediction_models/ReinforcementLearner.py
```diff
@@ -88,6 +88,33 @@ class MyRLEnv(Base5ActionRLEnv):
         User can override any function in BaseRLEnv and gym.Env. Here the user
         sets a custom reward based on profit and trade duration.
         """
+        def reset(self):
+
+            # Reset custom info
+            self.custom_info = {}
```
robcaulk: I think we can put this in the `BaseEnvironment` so that it doesnt need to be defined for all use-cases.

smarmau:
Thanks Rob, as discussed on discord, I'll leave it with you to make this as generable and maintainable as possible. Appreciate the discussions and help.

freqtrade/freqai/RL/BaseReinforcementLearningModel.py
robcaulk: Looks like the multi_proc variant is our constraint here - we can't store the callback object inside the multiproc env, since there are a bunch of envs duplicated across various processes. Instead we need to keep the callback stored in the model, as you did here.

robcaulk (approved):
LGTM, thanks for the contribution!

PR #7860, opened 2022-12-05
Ensure base tf to be include_timeframes

aemr3:
## Summary

Added functionality to check if base timeframe is included in include_timeframes. If it's not included then we prepend base timeframe into include_timeframes. 

Also made model_training_parameters optional, removed n_estimators from freqai config example to allow easy start with RL.

This is current error when you forget to include base timeframe include_timeframes:

```
2022-12-05 23:24:05,895 - freqtrade.strategy.strategy_wrapper - ERROR - Unexpected error '3m' calling <bound method IStrategy._analyze_ticker_internal of <Example.Example object at 0x13bd506a0>>
  File "/Temp/freqtrade/freqtrade/strategy/strategy_wrapper.py", line 27, in wrapper
    return f(*args, **kwargs)
  File "/Temp/freqtrade/freqtrade/strategy/interface.py", line 748, in _analyze_ticker_internal
    dataframe = self.analyze_ticker(dataframe, metadata)
  File "/Temp/freqtrade/freqtrade/strategy/interface.py", line 726, in analyze_ticker
    dataframe = self.advise_indicators(dataframe, metadata)
  File "/Temp/freqtrade/freqtrade/strategy/interface.py", line 1221, in advise_indicators
    return self.populate_indicators(dataframe, metadata)
  File "/Temp/freqtrade/var/strategies/Example.py", line 169, in populate_indicators
    dataframe = self.freqai.start(dataframe, metadata, self)
  File "/Temp/freqtrade/freqtrade/freqai/freqai_interface.py", line 140, in start
    dk = self.start_live(dataframe, metadata, strategy, self.dk)
  File "/Temp/freqtrade/freqtrade/freqai/freqai_interface.py", line 364, in start_live
    self.dd.update_historic_data(strategy, dk)
  File "/Temp/freqtrade/freqtrade/freqai/data_drawer.py", line 664, in update_historic_data
    self.current_candle = history_data[dk.pair][self.config['timeframe']].iloc[-1]['date']
KeyError: '3m'
```

### DIFF
--- a/freqtrade/freqai/utils.py
+++ b/freqtrade/freqai/utils.py
@@ -233,3 +233,23 @@ def get_timerange_backtest_live_models(config: Config) -> str:
     dd = FreqaiDataDrawer(models_path, config)
     timerange = dd.get_timerange_from_live_historic_predictions()
     return timerange.timerange_str
+
+
+def ensure_base_tf_in_include_timeframes(config: Config) -> Config:
+    """
+    Ensure that the base timeframe is included in the include_timeframes list
+    :param config: Configuration dictionary
+
+    :return config: Configuration dictionary
+    """
+    feature_parameters = config.get('freqai', {}).get('feature_parameters', {})
+    include_timeframes = feature_parameters.get('include_timeframes', [])
+
+    if config['timeframe'] in include_timeframes:
+        return config
+
+    include_timeframes = [config['timeframe']] + include_timeframes
+    config.get('freqai', {}).get('feature_parameters', {}) \
+        .update({**feature_parameters, 'include_timeframes': include_timeframes})
+
+    return config
--- a/freqtrade/freqai/utils.py
+++ b/freqtrade/freqai/utils.py
@@ -233,23 +233,3 @@ def get_timerange_backtest_live_models(config: Config) -> str:
     dd = FreqaiDataDrawer(models_path, config)
     timerange = dd.get_timerange_from_live_historic_predictions()
     return timerange.timerange_str
-
-
-def ensure_base_tf_in_include_timeframes(config: Config) -> Config:
-    """
-    Ensure that the base timeframe is included in the include_timeframes list
-    :param config: Configuration dictionary
-
-    :return config: Configuration dictionary
-    """
-    feature_parameters = config.get('freqai', {}).get('feature_parameters', {})
-    include_timeframes = feature_parameters.get('include_timeframes', [])
-
-    if config['timeframe'] in include_timeframes:
-        return config
-
-    include_timeframes = [config['timeframe']] + include_timeframes
-    config.get('freqai', {}).get('feature_parameters', {}) \
-        .update({**feature_parameters, 'include_timeframes': include_timeframes})
-
-    return config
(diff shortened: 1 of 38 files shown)


freqtrade/freqai/utils.py
```diff
@@ -233,3 +233,23 @@ def get_timerange_backtest_live_models(config: Config) -> str:
...
+
+    :return config: Configuration dictionary
+    """
+    feature_parameters = config.get('freqai', {}).get('feature_parameters', {})
+    include_timeframes = feature_parameters.get('include_timeframes', [])
+
+    if config['timeframe'] in include_timeframes:
+        return config
+
+    include_timeframes = [config['timeframe']] + include_timeframes
```
xmatthias: Why is this a requirement?

i think i should be fine running on 5m data (main tf) - but have `"include_timeframes": ["15m", "30m", "1h"]`.

The include_timeframes can't be smaller than the main timeframe (so changing main to 30m won't work - but this is validated) - but i'm not aware why my above setup wouldn't be valid ?

aemr3: > Why is this a requirement?
> 
> i think i should be fine running on 5m data (main tf) - but have `"include_timeframes": ["15m", "30m", "1h"]`.
> 
> The include_timeframes can't be smaller than the main timeframe (so changing main to 30m won't work - but this is validated) - but i'm not aware why my above setup wouldn't be valid ?

It is not valid, currently including base timeframe in include_parameters is a requirement. But there is no validation. You just get KeyError if you try to run without including base timeframe in your include_timeframes config. 

This function will allow users to forget base timeframe in include_timeframes. 

I could do this without modifying config here. But there were a lot of loops for getting data and other stuff so this was an easy method to do it. 

If this doesn’t look ideal to you we can of course change all loops of include_timeframes one by one in all files.

xmatthias: interesting, seems like i never ran into this so far, but i thought i had tested exactly this.

To be clear - I'm not questioning the method, just the requirement - as it's new to me.

Only thought then -  you might want to do this in `_validate_freqai_include_timeframes` - so all related logic is in one place.

aemr3: You mean updating config in that function or throwing an error if base timeframe not in include_timeframes?

I think since those validation functions doesn’t return anything updating config inside there is not possible without a change?

aemr3: By the way we can still remove this requirement by changing all loops of include_timeframes.

xmatthias: `dict.update()` will mutate the original dictionary. 
Otherwise your method would break tons of stuff, as then we'd have 2 configurations around - one you modified, and one that was there before.

> By the way we can still remove this requirement by changing all loops of include_timeframes.

i'm still not a fan of this ... :laughing:

aemr3: > `dict.update()` will mutate the original dictionary.

Ok, I will move the logic there. Thanks for the info.

robcaulk:
Looks like the test is failing, any ideas why?

aemr3:
> Looks like the test is failing, any ideas why?

Yeah, fixed it. It is working now. My last commit passed somehow on my tests. It was incorrect.

robcaulk (approved):
LGTM

PR #7861, opened 2022-12-05
Add support for collating and analysing rejected signals in backtest

### LINKED ISSUE #7870: Export backtest analysis to csv
## Describe your environment

  * Operating system: Linux mint Una
  * Python Version: 3.10.4
  * CCXT version: 2.0.96
  * Freqtrade Version: 2022.11.dev-2c1330a4e
  
## Your question

How do we export backtesting analysis to csv file with command `freqtrade backtesting-analysis`? I would like to have my exit-reasons and indicators in a spreadsheet.

Thanks in advance

froggleston:
## Summary

Currently, backtesting only outputs the number of rejected signals, and doesn't allow any analysis of which pairs or dates. This PR adds in support to track these rejected signals, and a new backtesting-analysis option, `--rejected` to output a table of rejected signals.

As this could be a large list, there is a new option to backtesting-analysis, `--analysis-to-csv`, which will output CSV files for each table specified in the analysis command. It will by default output these files in the `user_data/backtest_results` folder unless specified with the new `--analysis-csv-path` option.

## Quick changelog

- Add support in `backtesting.py` to store rejected signalsif the `--export signals` option is supplied.
- Add support in `backtesting-analysis` to turn rejected table outputting on and off, and to save tables direct to CSV

## What's new?

The `--export signals` flag will now also produce a dataframe of rejected signals, i.e. the pair, date, and entry_tag. When enabled, this will write a .pkl file containing the rejected signal dataframe, as well as the existing signals.pkl file. This file is read when using the `backtesting-analysis` command with the `--rejected` flag, producing a table like below:

```
Rejected Signals:
+----------+---------------------------+----------------+
| pair     | date                      | enter_tag      |
|----------+---------------------------+----------------|
| SOL/USDT | 2022-11-08 18:00:00+00:00 | buy_1          |
| SOL/USDT | 2022-11-08 18:05:00+00:00 | buy_1          |
| ONE/USDT | 2022-11-08 19:35:00+00:00 | buy_3          |
+----------+---------------------------+----------------+
```
## What's missing?

A specific test for the rejected output. I can't work out how to get a mock set up to represent a rejected signal.


froggleston:
Not sure why `recordfilename = 'export.txt'` is coming through as a str when this worked before... Checking...

Update: typo :kek:

freqtrade/optimize/backtesting.py
```diff
@@ -1073,20 +1088,22 @@ def backtest_loop(
...
+                if trade:
+                    # TODO: hacky workaround to avoid opening > max_open_trades
+                    # This emulates previous behavior - not sure if this is correct
+                    # Prevents entering if the trade-slot was freed in this candle
+                    open_trade_count_start += 1
+                    # logger.debug(f"{pair} - Emulate creation of new trade: {trade}.")
+                    LocalTrade.add_bt_trade(trade)
+                    self.wallets.update()
+            else:
+                self._collate_rejected(pair, row)
```
xmatthias, line 1114: This is not a rejected trade - it's simply a signal that's not acted on (consider it skipped).

The only time we "reject" a trade is via `confirm_trade_entry()`.

docs/advanced-backtesting.md
```diff
@@ -114,3 +115,37 @@ For example, if your backtest timerange was `20220101-20221231` but you only wan
 ```bash
 freqtrade backtesting-analysis -c <config.json> --timerange 20220101-20220201
 ```
+
+### Printing out rejected trades
+
+Use the `--rejected` option to print out rejected trades.
```
xmatthias: You will clearly have to explain what "rejected" trades are.

to me - these are trade that 
* got rejected by the user via confirm_trade_entry
* got rejected by the market - by not filling the order (so the initial entry is canceled).

I think you mean neither of these (at least that's what the code suggests)... so the term rejected is probably wrong.

freqtrade/optimize/backtesting.py
froggleston: ~~But these are called "Rejected Trades" in the backtesting output. Either this should be "Skipped Entry Signals" in the report, or Rejected Trades should truly reflect user-rejected trades with `confirm_trade_entry()`~~

Ah these are now "Rejected Entry Signals" in the report

docs/advanced-backtesting.md
froggleston: ~~See above...~~
OK this is fair. I can adjust this to be `--rejected-signals`

freqtrade/optimize/backtesting.py
xmatthias: rejected signals will work - sure (completely changes the context).

i only read the title at first - and it didn't contain what i expected AT ALL :laughing:

froggleston: Yeah I can see the terminology was incorrect. Amended!

docs/advanced-backtesting.md
froggleston: I'll commit this later.

freqtrade/commands/cli_options.py
```diff
@@ -634,7 +634,7 @@ def __init__(self, *args, **kwargs):
               "3: by pair and enter_tag, "
               "4: by pair, enter_ and exit_tag (this can get quite large)"),
         nargs='+',
-        default=['0', '1', '2'],
+        default=[],
```
xmatthias, line 639: is this intentional? the way it acts now is to echo "nothing" by default ...

freqtrade/data/entryexitanalysis.py
```diff
@@ -186,38 +207,65 @@ def prepare_results(analysed_trades, stratname,
...
-            headers='keys',
-            tablefmt='psql',
-            showindex=show_index
+    if to_csv:
+        if csv_path is not None:
+            safe_name = Path(csv_path,
+                             name.lower().replace(" ", "_").replace(":", ""))
+        else:
+            safe_name = Path("user_data",
+                             "backtest_results",
```
xmatthias: This is wrong - and will ignore settings for `userdir` in both CLI (`--user-dir xxxx`) as well as via config.

freqtrade/data/entryexitanalysis.py
```diff
@@ -43,9 +51,9 @@ def _process_candles_and_indicators(pairlist, strategy_name, trades, signal_cand
         for pair in pairlist:
             if pair in signal_candles[strategy_name]:
                 analysed_trades_dict[strategy_name][pair] = _analyze_candles_and_indicators(
-                                                              pair,
-                                                              trades,
-                                                              signal_candles[strategy_name][pair])
+                                                            pair,
+                                                            trades,
+                                                            signal_candles[strategy_name][pair])
```
xmatthias: i think this should be indented 4 spaces > to the prior line.
Then we don't have to move it by 2 spaces every time a name changes...

tests/data/test_entryexitanalysis.py
```diff
@@ -191,8 +191,18 @@ def test_backtest_analysis_nomock(default_conf, mocker, caplog, testdatadir, tmp
...
+                    ['--analysis-groups', "0", "1", "2",
+                     '--timerange', "20180129-20180130"]
+                    )
     start_analysis_entries_exits(args)
     captured = capsys.readouterr()
     assert 'enter_tag_long_a' in captured.out
     assert 'enter_tag_long_b' not in captured.out
+
+    # test rejected - how to mock this?
+    # args = get_args(base_args + ['--rejected-signals'])
```
xmatthias: backtest results are not mocked for this test (it's simply running backtesting and uses whatever results).

best have a look if the backtest above returns such signals (i don't know - this is a test you added iirc).

if it doesn't - try to either move the timerange around to hit a rejected signal - or try disabling sell signal (for example), or disable ROI.
The way backtest is running above doesn't make that simple though - as you never get access to the strategy object from the test.


I'd however suggest to split the 2 tests apart - so you can keep the existing test unchanged, while still searching for "valid rejections" on a new backtest setup - which may no longer have ROI sells ...

freqtrade/configuration/configuration.py
```diff
@@ -465,6 +465,15 @@ def _process_analyze_options(self, config: Config) -> None:
         self._args_to_config(config, argname='timerange',
                              logstring='Filter trades by timerange: {}')
 
+        self._args_to_config(config, argname='analysis_rejected',
+                             logstring='Analyse rejected signals: {}')
+
+        self._args_to_config(config, argname='analysis_to_csv',
+                             logstring='Store analysis tables to CSV: {}')
+
+        self._args_to_config(config, argname='analysis_csv_path',
```
xmatthias, line 474: I'm not sure i'd make this configurable.
it adds little benefit in my eyes (it still won't allow you to customize the filename) - so backtest_result is as good a name as any.

freqtrade/commands/cli_options.py
```diff
@@ -658,6 +658,21 @@ def __init__(self, *args, **kwargs):
...
+    "analysis_to_csv": Arg(
+        '--analysis-to-csv',
+        help='Save selected analysis tables to individual CSVs',
+        action='store_true',
+    ),
+    "analysis_csv_path": Arg(
+        '--analysis-csv-path',
+        help=("Specify a path to save the analysis CSVs "
+              "if --analysis-to-csv is enabled. Default: user_data/basktesting_results/"),
+    ),
```
xmatthias, line 677: Any reason we can't combine this with `export_csv` (see above in this file)? 
maybe there's good reasons to split it - maybe it's because you didn't notice ... ?

freqtrade/data/entryexitanalysis.py
```diff
@@ -148,11 +157,23 @@ def _do_group_table_output(bigdf, glist):
...
+                _print_table(new, sortcols, name=f"Group {g}:",
+                             to_csv=to_csv, csv_path=csv_path)
             else:
                 logger.warning("Invalid group mask specified.")
 
 
+def _do_rejected_signals_output(rejected_signals_df, to_csv=False, csv_path=None):
+    cols = ['pair', 'date', 'enter_tag']
+    sortcols = ['date', 'pair', 'enter_tag']
+    _print_table(rejected_signals_df[cols],
```
xmatthias, line 176: I find the output quite odd in case there's no rejected signal.
i've had to tripple-check to see why i get no output before realizing that i probably had no rejected signals in my backtest.

maybe adding an output "someplace" which says "no rejected signals" would make this userfriendlier?

freqtrade/data/entryexitanalysis.py
```diff
@@ -186,38 +207,65 @@ def prepare_results(analysed_trades, stratname,
...
-            showindex=show_index
+    if to_csv:
+        if csv_path is not None:
+            safe_name = Path(csv_path,
+                             name.lower().replace(" ", "_").replace(":", ""))
+        else:
+            safe_name = Path("user_data",
+                             "backtest_results",
+                             name.lower().replace(" ", "_").replace(":", ""))
+        data.to_csv(f"{str(safe_name)}.csv")
```
xmatthias: I think you'll want some comment here - making the user identify that "something" happened.

simply running this results in some output not telling me that happened ANYTHING.

freqtrade/data/entryexitanalysis.py
```diff
@@ -247,7 +306,10 @@ def process_entry_exit_reasons(config: Config):
 
                 print_results(res_df,
                               analysis_groups,
-                              indicator_list)
+                              indicator_list,
+                              rejected_signals=rej_df,
+                              to_csv=to_csv,
+                              csv_path=csv_path)
```
xmatthias, line 320: Based on the fallback used for `csv_path` - csv_path can never be `None` - making huge parts of the downstream logic pointless.

freqtrade/data/entryexitanalysis.py
```diff
@@ -186,38 +207,65 @@ def prepare_results(analysed_trades, stratname,
...
-    print(
-        tabulate(
-            data,
-            headers='keys',
-            tablefmt='psql',
-            showindex=show_index
+    if to_csv:
+        if csv_path is not None:
+            safe_name = Path(csv_path,
+                             name.lower().replace(" ", "_").replace(":", ""))
```
xmatthias: the generated names are quite ... nothing-saying - and will overwrite themselfs whenever running it again.
while this CAN be intentional (save disk space) - i don't think it should be (if they overwrite - we don't need a csv output at all).

effectively, the files will be useless the moment they are written - as you'll not remember 2 days from now what the csv file was for (and there's 0 content telling you that it was for strategyX - version Y).

As such - i'd question the csv export per se - do we really think this is useful / necessary?

freqtrade/optimize/optimize_reports.py
```diff
@@ -45,29 +45,41 @@ def store_backtest_stats(
...
 
 
+def store_backtest_signal_candles(
+        recordfilename: Path, candles: Dict[str, Dict], dtappendix: str) -> Path:
+    return _store_backtest_analysis_data(Path(recordfilename), candles, dtappendix, "signals")
+
+
+def store_backtest_rejected_signals(
+        recordfilename: Path, trades: Dict[str, Dict], dtappendix: str) -> Path:
+    return _store_backtest_analysis_data(Path(recordfilename), trades, dtappendix, "rejected")
```
xmatthias: Considering how these 2 functions are used - it might make more sense to have just 1 function - which calls the same underlying function twice with diff. arguments.

That'll keep the part in backtesting file cleaner - as it's only one call - abstracting away the logic as to WHAT you're doing with the files below.
it's not like there's an option to call only one or the other ...

freqtrade/optimize/backtesting.py
```diff
@@ -1282,12 +1300,33 @@ def _generate_trade_signal_candles(self, preprocessed_df, bt_results):
...
-                    signal_candles_only_df = pd.concat([signal_candles_only_df, signal_inds])
+                    signal_candles_only_df = pd.concat([
+                        signal_candles_only_df.infer_objects(),
+                        signal_inds.infer_objects()])
 
                 signal_candles_only[pair] = signal_candles_only_df
 
         self.processed_dfs[self.strategy.get_strategy_name()] = signal_candles_only
 
+    def _generate_rejected_signals(self, preprocessed_df, rejected_dict):
```
xmatthias: I think this function should be moved out of the backesting class.

It can just as well return the object - but would keep this file cleaner (and the function itself could be tested without side-effects).
That obviously also applies to the above "generate_trade_signals_candle" function - so feel free to leave this for a future refactoring.

freqtrade/commands/cli_options.py
froggleston: Yes, this is because even if someone just wanted the rejected signals, they get everything else whether they like it or not.

freqtrade/data/entryexitanalysis.py
froggleston: OK I'll work on that.

freqtrade/configuration/configuration.py
froggleston: OK so you suggest simply using user_data/backtest_result even if export-dir is supplied?

freqtrade/commands/cli_options.py
froggleston: I did notice, but it would require changing the description in the command, which I wasn't sure you'd prefer. I can easily rework this to use export_csv.

freqtrade/data/entryexitanalysis.py
froggleston: Sure, I can add this.

freqtrade/data/entryexitanalysis.py
froggleston: The overwriting was intentional - you can run the analysis multiple times to get the output you want, and in this process could generate tens or hundreds of CSVs.

We've already had one ticket requesting CSV export, and printing 1000+ lines to stdout isn't feasible when you might have thousands of rejected signals, so I think it's a valid feature.

freqtrade/optimize/optimize_reports.py
froggleston: OK, sounds fair.

freqtrade/optimize/backtesting.py
froggleston: You might have to give me an example, as we have the similar function for generating signal candle dataframes. If we remove this rejected functionality, is there a rationale for removing the signal generation? I'm not sure how I'd get access to the backtesting information without having this generation in the backtesting class itself...

freqtrade/commands/cli_options.py
xmatthias: well the example (hyperopt.csv) is dedicated to where it's currently used.
if it's used elsewhere, too - there's no reason to not change this to something more generic.

freqtrade/optimize/backtesting.py
xmatthias: If you look at the function body - then the ONLY reference to `self` is in the very last line (twice).

So instead of that assignment - you do a `return rejected_candles_only` - which makes this a function without side-effects (data is passed in and something else comes out - nothing else is "hurt" in the process).

freqtrade/commands/cli_options.py
froggleston: Would you recommend using the `exportfilename` option for this also, like the `export_csv` above?

xmatthias: well iirc - then `export_csv` is a path relative to CWD - so it would not need 2 options for this (you can just pass `userdata/backtests/whatever.csv`).
Might need checking though - as it's been ages since i touched the csv code

freqtrade/optimize/optimize_reports.py
froggleston: Would you suggest using an enum for the type? e.g. `freqtrade.enums.BacktestingAnalysisOutputType.SIGNAL_CANDLES`

xmatthias: what for? (maybe i'm missing some context ...) 

it's not really optional (at least not right now)
if you call one - you'll also call the other.

My point above was to use one method - and simply call `_store_backtest_analysis_data` twice from within the same method - so i don't really see a usecase for an enum?

froggleston: Ah sorry, I was overthinking it. Understood 👍

freqtrade/configuration/configuration.py
xmatthias: well using export_dir is fine i guess? so the file is where all the other results are.

Currently - we're allowing someone to set "store this in cwd" (`--analysis_csv_path $(pwd)`) - but the filename will be static. 
So i have no way to prevent overwriting, unless i want a separate directory for every file.

if any, then a "csv path+name" argument would make more sense  - allowing me to customize the name (if i want to prevent the overwriting for specific results).

froggleston (commented):
Everything looks good. Tested with fresh clone and setup.sh, rejected output works, as does the CSV exports. I see the CSV export CLI options are still in, separate from the other export CLI option. I guess this would be approached in a subsequent PR. So, LGTM!

xmatthias:
>  I see the CSV export CLI options are still in, separate from the other export CLI option. I guess this would be approached in a subsequent PR

this was actually intentional - as the new options allow to specify a directory - while the existing ones refer to a specific file (which won't work in this case - as it can potentially generate multiple CSV files).

xmatthias (approved):
LGTM :+1:

froggleston:
Thanks for finishing this off - I've been procrastinating on it for too long! Great to see it included.

PR #7862, opened 2022-12-05
New websocket message "new_candle"

xmatthias:
## Summary
Issue a new message whenever a new candle is issued.
this will trigger a minimal notification once a new candle is analyzed - and therefore ready for an UI to be picked up.

Since the UI needs both a different format, as well as the dataframe for just one pair, this is the most efficient way.

PR on freqUI to follow a little later - adding this functionality to the UI as well.

### DIFF
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -101,7 +101,12 @@ class DataProvider:
         """
         return self.__producer_pairs.get(producer_name, []).copy()
 
-    def _emit_df(self, pair_key: PairWithTimeframe, dataframe: DataFrame, new_candle: bool) -> None:
+    def _emit_df(
+        self,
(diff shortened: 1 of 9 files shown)


wizrds (approved):
LGTM. One thing I noticed right away was if/when we decide to make use of this message type in a consumer then we'll need to define it in the `ws_schemas.py` file (which is mainly used to validate incoming messages not outgoing).

xmatthias:
> One thing I noticed right away was if/when we decide to make use of this message type in a consumer then we'll need to define it in the `ws_schemas.py` file (which is mainly used to validate incoming messages not outgoing).

This would be true for every message there is (we can/could also subscribe to `entry_fill` - which is not "really" typed at the moment).
I don't disagree that we might have to type these in the future (also for own own sake) - but then that would apply to all messages we send around - not just this one :laughing:

PR #7866, opened 2022-12-07
Cleanup tensorboard callback

initrv:
## Summary

Callback cleanup and simplification to make it easier to add new metrics

### DIFF
--- a/freqtrade/freqai/RL/Base5ActionRLEnv.py
+++ b/freqtrade/freqai/RL/Base5ActionRLEnv.py
@@ -49,7 +49,7 @@ class Base5ActionRLEnv(BaseEnvironment):
         self._update_unrealized_total_profit()
         step_reward = self.calculate_reward(action)
         self.total_reward += step_reward
-        self.tensorboard_metrics[self.actions._member_names_[action]] += 1
+        self.tensorboard_log(self.actions._member_names_[action])
 
         trade_type = None
         if self.is_tradesignal(action):
--- a/freqtrade/freqai/RL/Base5ActionRLEnv.py
+++ b/freqtrade/freqai/RL/Base5ActionRLEnv.py
@@ -49,6 +49,7 @@ class Base5ActionRLEnv(BaseEnvironment):
         self._update_unrealized_total_profit()
         step_reward = self.calculate_reward(action)
         self.total_reward += step_reward
+        self.tensorboard_metrics[self.actions._member_names_[action]] += 1
 
         trade_type = None
         if self.is_tradesignal(action):
--- a/freqtrade/freqai/prediction_models/ReinforcementLearner.py
+++ b/freqtrade/freqai/prediction_models/ReinforcementLearner.py
@@ -100,6 +100,7 @@ class ReinforcementLearner(BaseReinforcementLearningModel):
             """
             # first, penalize if the action is not valid
             if not self._is_valid(action):
+                self.tensorboard_log("is_valid")
                 return -2
 
             pnl = self.get_unrealized_profit()
--- a/freqtrade/freqai/prediction_models/ReinforcementLearner.py
+++ b/freqtrade/freqai/prediction_models/ReinforcementLearner.py
@@ -108,15 +108,12 @@ class ReinforcementLearner(BaseReinforcementLearningModel):
             # reward agent for entering trades
             if (action == Actions.Long_enter.value
                     and self._position == Positions.Neutral):
-                self.custom_info[Actions.Long_enter.name] += 1
                 return 25
             if (action == Actions.Short_enter.value
                     and self._position == Positions.Neutral):
-                self.custom_info[Actions.Short_enter.name] += 1
                 return 25
             # discourage agent from not entering trades
             if action == Actions.Neutral.value and self._position == Positions.Neutral:
-                self.custom_info[Actions.Neutral.name] += 1
                 return -1
 
             max_trade_duration = self.rl_config.get('max_trade_duration_candles', 300)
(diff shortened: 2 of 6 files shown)


freqtrade/freqai/prediction_models/ReinforcementLearner.py
```diff
@@ -100,7 +100,6 @@ def calculate_reward(self, action: int) -> float:
             """
             # first, penalize if the action is not valid
             if not self._is_valid(action):
-                self.custom_info["Invalid"] += 1
```
robcaulk, line 103: Why are you removing this one?

freqtrade/freqai/prediction_models/ReinforcementLearner.py
```diff
@@ -131,22 +130,21 @@ def calculate_reward(self, action: int) -> float:
...
 
             # close short
             if action == Actions.Short_exit.value and self._position == Positions.Short:
                 if pnl > self.profit_aim * self.rr:
                     factor *= self.rl_config['model_reward_parameters'].get('win_reward_factor', 2)
-                self.custom_info[f"{Actions.Short_exit.name}"] += 1
+                self.custom_info[Actions.Short_exit.name] += 1
                 return float(pnl * factor)
 
-            self.custom_info["Unknown"] += 1
```
robcaulk, line 151: And this one?

freqtrade/freqai/prediction_models/ReinforcementLearner.py
```diff
@@ -109,15 +108,15 @@ def calculate_reward(self, action: int) -> float:
             # reward agent for entering trades
             if (action == Actions.Long_enter.value
                     and self._position == Positions.Neutral):
-                self.custom_info[f"{Actions.Long_enter.name}"] += 1
+                self.custom_info[Actions.Long_enter.name] += 1
```
robcaulk: Should we add a note in the documentation telling users how they can use `custom_info`?

robcaulk (commented):
Hey, looks good but I have a few questions regarding removals.

freqtrade/freqai/prediction_models/ReinforcementLearner.py
aemr3: > Should we add a note in the documentation telling users how they can use `custom_info`?

How about renaming `custom_info` to something like `metrics`?

freqtrade/freqai/prediction_models/ReinforcementLearner.py
initrv: I originally added the "invalid", "hold" and "unknown" values when I tested the maskable ppo. I think logging actions that are not explicitly specified in model actions can be confusing for users and should not be present by default. But I can return them back if you think otherwise.

freqtrade/freqai/prediction_models/ReinforcementLearner.py
initrv: With this PR, the user just needs to define "self.custom_info['string'] = value" anywhere and the metric will appear in the tensorboard under _custom. Incrementable values as action counters must be predefined in reset()

initrv: > > Should we add a note in the documentation telling users how they can use `custom_info`?
> 
> How about renaming `custom_info` to something like `metrics`?

Renaming a dictionary is a good point, because the documentation defines custom_ variables as the recommended naming for user-defined variables, not built-in ones.

robcaulk: Best would be `tensorboard_metrics` or something to indicate that the value will be available in tensorboard. Also, we already have a `metric_tracker` which tracks resource usage in freqai.

freqtrade/freqai/prediction_models/ReinforcementLearner.py
robcaulk: If a user wants to understand how the agent navigated the reward, it makes sense to see how many times it made an invalid choice. Having stats on how many times the agent decided to hold us also important no? I don’t see how this would confuse a user.

 I also don’t quite understand the connection to maskable ppo.

initrv: Please look at the new commit. I propose to remove all metrics from the reward function, and move the new built-in action counters to step(). In my maskable ppo example, I don't want to track the invalid and unknown actions in the tensorboard because they are always zero. But if we define a counter on reset, it will show up even if I remove the metric from the reward function. So I think it's better to display the bare minimum and leave the user the freedom to add whatever they want

freqtrade/freqai/RL/Base5ActionRLEnv.py
```diff
@@ -49,6 +49,7 @@ def step(self, action: int):
         self._update_unrealized_total_profit()
         step_reward = self.calculate_reward(action)
         self.total_reward += step_reward
+        self.tensorboard_metrics[self.actions._member_names_[action]] += 1
```
robcaulk: Why would you prefer to have this here instead of inside the below `is_tradesignal()` logic?

While the action count *before* is_tradesignal is absolutely valuable, I am not sure it is *more* valuable than knowing the true trade signal action counts. Now the user will see many actions that were never acted upon due to the is_tradesignal protection.

robcaulk: Never mind, you can ignore this comment. Maybe we add another entry to the dictionary that is "acted upon" counts or something.

robcaulk (approved):
This is a nice one ;)

Issue #4493, opened 2021-03-06
UserPairlist handler to manage "Manual" pairlists


xmatthias:
## Describe the enhancement

we could have a `UserPairlist` (name to define - this one is probably not good) pairlist, which will either provide a full pairlist (simliar to staticpairlist) - or inject additional pairs to the pairlist.

This `UserPairlist` get's the pairlist from a (new) database table - allowing full control to RPC by adding and removing pairs from this userpairlist at will (it's empty by default, but persisted across restarts by beeing in the database).

as the pairlist is refreshed quite often (once every iteration, unless the pairlist implements some caching) - users could simply add / remove your pair, and it'd magically appear without changing configuration or whatever.

obviously blacklist (and such) will still apply as usual, so you can't whitelist pairs that are in a blacklist - but that shouldn't really be a problem i'd hope.

It can act as both a "generator" (first position of the pairlists) - or as filter (which will add the pairs in it's list).

aiforinvesting:
@xmatthias 
How come this does not work; providing all USDT pairs in the whitelist of config.json. and populating indicator is the default RSI.

    def populate_buy_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        """
        Based on TA indicators, populates the buy signal for the given dataframe
        :param dataframe: DataFrame populated with indicators
        :param metadata: Additional information, like the currently traded pair
        :return: DataFrame with buy column
        """
        tickers = []
        with open('tickers.txt', 'r') as filehandle:
            for line in filehandle:
                # remove linebreak which is the last character of the string
                ticker = line[:-1]
                # add item to the list
                tickers.append(ticker)

        dataframe.loc[
            (
                # Signal: RSI 
                (dataframe['rsi'] > 0)  &
                (metadata['pair'] in tickers)
            ),
            'buy'] = 1

        return dataframe

xmatthias:
I guess this approach could work (assuming you have no bugs) - but it's not what this issue is about, but  about providing a way to manage the pairlist "online" - either via API or via telegram.

Investigating why your code will or will not is however clearly not part of this issue.

epigramx:
The dream would be to have full (and easy) programmatic freedom. But I wonder if a good first version is an `EditPairList` that can be set at the end of the chain and it has only three string settings: append; prepend and remove.
So for example after one uses all the current filters as normal at the end it would have "remove these pairs" or/and "append these pairs" and/or "prepend these pairs".

PR #7867, opened 2022-12-07
Add Remotepairlist

Bloodhunter4rc:
## Summary

Introduce a Pairlist which receives the pairs from a Remote Website / URL as string (python list) or json, this could provide one pairlist for multiple instances of Freqtrade, filter the Pairlist not within the freqtrade instance but Prefilter it. Reduce API Calls.

Additional to that it can load a pairlist from a local json file specified.

## Quick changelog

- Added Remotepairlist

## What's new?

**Pairlist Config-Remote:**
```
    "pairlists": [
        {
            "method": "RemotePairList",
            "number_assets": 20,
            "pairlist_url": "http://pairlist.robot.co.network/test.json",
            "refresh_period": 1800
        }
    ],
```

**Pairlist Config-Local:**
```
    "pairlists": [
        {
            "method": "RemotePairList",
            "number_assets": 20,
            "pairlist_url": "file:///test.json",
            "refresh_period": 1800
        }
    ],
```

**Remote json file:**
`
{
  "pairs": ["BTC/USDT","ETH/USDT"],
  "info": "Pairlist fetched with Remotepairlist",
  "refresh_period": 1800
}
`

**Remote php/txt:**
`
["BTC/USDT","ETH/USDT","BNB/USDT","XRP/USDT","DOGE/USDT","ADA/USDT","MATIC/USDT","DOT/USDT","LTC/USDT","SHIB/USDT","SOL/USDT","TRX/USDT","UNI/USDT","AVAX/USDT","LINK/USDT","ATOM/USDT","ETC/USDT","XMR/USDT","XLM/USDT","BCH/USDT","QNT/USDT","ALGO/USDT","FIL/USDT","NEAR/USDT","APE/USDT","VET/USDT","HBAR/USDT","ICP/USDT","FLOW/USDT","EGLD/USDT","EOS/USDT","LUNC/USDT","AXS/USDT","TWT/USDT","SAND/USDT","XTZ/USDT","AAVE/USDT","THETA/USDT","LDO/USDT","CHZ/USDT","MANA/USDT","APT/USDT","CAKE/USDT","FTM/USDT","ZEC/USDT","XEC/USDT","MKR/USDT","KLAY/USDT","DASH/USDT","NEO/USDT"] # Test
`

`2022-12-07 18:12:22,960 - RemotePairList - INFO - Searching 20 pairs: ['BTC/USDT', 'ETH/USDT', 'XRP/USDT', 'DOGE/USDT', 'ADA/USDT', 'MATIC/USDT', 'DOT/USDT', 'LTC/USDT', 'SHIB/USDT', 'SOL/USDT', 'TRX/USDT', 'UNI/USDT', 'AVAX/USDT', 'LINK/USDT', 'ATOM/USDT', 'ETC/USDT', 'XMR/USDT', 'XLM/USDT', 'BCH/USDT', 'QNT/USDT']
2022-12-07 18:12:22,960 - RemotePairList - INFO - Pairlist fetched with Remotepairlist Fetched in 0.088381 seconds.`

## Todo

- [x] Docs
- [x] Unit Test


xmatthias (commented):
i'm a bit sceptical on what the usecase for this might be.
I'm sure it's useful for you personally somehow - but will it be useful for the overall community?

you'll either need to support many possible formats (you won't know **how** an unknown, remote source will provide their pairlist), and either try to detect that format, or have it configurable (which will also be a maintenance nightmare) 

or it'll need to be a format we dictate (which begs the question where that remote format would come from). 
The format your test config uses doesn't appear to be one from a generic provider, but a format quite tailored to how you'd like to have it - but i doubt any provider will use this format "just because".

If it's coming from an upstream freqtrade setup - then this already exists in the [producer pairlist](https://www.freqtrade.io/en/latest/plugins/#producerpairlist) - which will (on top of the pairlist) also allow you to use the calculated indicators from the upstream instance.

Bloodhunter4rc:
> i'm a bit sceptical on what the usecase for this might be. I'm sure it's useful for you personally somehow - but will it be useful for the overall community?

The idea behind this was different, to use several external APIs from a website hosted by yourself or a third party to allow the user to filter their pairlist. Filtering and sorting by indicators, such as trending pairs or for example pairs with a strong Pearson Correlation, can be done by the user to create their own pairlist. This allows for greater flexibility as the user has complete control over the filtering and sorting criteria. 

It also can be used on exchanges that may not support a Volumepairlist or have issues with frequent API calls.

> you'll either need to support many possible formats (you won't know **how** an unknown, remote source will provide their pairlist), and either try to detect that format, or have it configurable (which will also be a maintenance nightmare)
> 
> or it'll need to be a format we dictate (which begs the question where that remote format would come from). The format your test config uses doesn't appear to be one from a generic provider, but a format quite tailored to how you'd like to have it - but i doubt any provider will use this format "just because".
> 

I agree it is not maintainable to support many formats, Json is widely used and freqtrade users are familiar with it (freqtrade configuration / backtest results) the structure of the proposed json is very simple and could be documented. The "text/html" part could be removed.

> If it's coming from an upstream freqtrade setup - then this already exists in the [producer pairlist](https://www.freqtrade.io/en/latest/plugins/#producerpairlist) - which will (on top of the pairlist) also allow you to use the calculated indicators from the upstream instance.

For me that is a different usecase.

xmatthias:
> The idea behind this was different, to use several external APIs from a website hosted by yourself or a third party to allow the user to filter their pairlist. Filtering and sorting by indicators, such as trending pairs or for example pairs with a strong Pearson Correlation, can be done by the user to create their own pairlist. This allows for greater flexibility as the user has complete control over the filtering and sorting criteria.

Such a service is not available as far as i know.

---

My main point remains - it's an idea - but without upstream - it's a pointless addition.

Unless there is a (popular and trust-worthy) upstream service available (the format would probably have to be aligned to that) - i don't see us accepting this, as it would result in "dead" code - which almost nobody could use (there's no upstream available) - but we'd still carry around until someone finally decides to remove it.


if our goal is to give the user full, interactive control of the pairlist, then this is clearly the wrong approach.
in that case, you should direct your efforts towards implementing #4493 - which would enable  adding/removing pairs interactively via freqUI.

Bloodhunter4rc:
Maybe this is then not useful for the overall community. I created pairlist.robot.co.network for exactly that reason. I thought it would not be too hard to create a JSON file with pairs and remotely fetch it or locally load it, but that is not what most people are going to do. I don't want to push my service so this was kept as generic as possible. It's also not popular and i dont have that aim for it (a new service can never be thats kinda chicken and egg situation). #4493 is not what i was looking for, but it can potentially be more beneficial for the overall community.

xmatthias:
> I created pairlist.robot.co.network for exactly that reason
> I don't want to push my service so this was kept as generic as possible.

i don't really see this as a problem necessarily - as long as we're clear / open about that (the initial post doesn't really mention any of that).

To me - the URL looked like your custom hosted file (not usable for the general public as it's a static file - which it probably is).
I honestly don't look at the root of a provided config to see "what may be findable there".

Assuming your aim is to keep that service going forward (available to users ...) - i'd welcome if you could reopen (and finish) this PR.

---

I'm not sure if we'll really need 2 ways for now - parsing HTML is really NOT a good approach (it's usually something that's quite prone to fail).

Bloodhunter4rc:
> > I created pairlist.robot.co.network for exactly that reason
> > I don't want to push my service so this was kept as generic as possible.
> 
> i don't really see this as a problem necessarily - as long as we're clear / open about that (the initial post doesn't really mention any of that).
> 
> To me - the URL looked like your custom hosted file (not usable for the general public as it's a static file - which it probably is). I honestly don't look at the root of a provided config to see "what may be findable there".
> 
> Assuming your aim is to keep that service going forward (available to users ...) - i'd welcome if you could reopen (and finish) this PR.
> 
> I'm not sure if we'll really need 2 ways for now - parsing HTML is really NOT a good approach (it's usually something that's quite prone to fail).

I am planning to keep the service running (for all users).

I removed the html part, should a example be provided with my page for usage in the docs? Or keep it out of the docs? I did keep it out for now.

It is my first pull request ever so i got some questions regarding the unit test(s):

Is it ok to keep fetch_pairlist inside the PairlistHandler or will this have to be moved somewhere else? Should the test be added to the existing test_pairlist, if so the previous question remains where to put that function to call it for the test? Or should code be duplicated for the test?

xmatthias:
> Is it ok to keep fetch_pairlist inside the PairlistHandler or will this have to be moved somewhere else?

not sure where you'd move it to - it's the core part of this pairlist - as long as we make sure it's properly cached (not called whenever we call the pairlist) it'll be fine.

tests - i think you'll want separate tests - as you'll want to properly mock the network call.

Bloodhunter4rc:
I added a few unit tests, but i had issues mocking the response when using a context manager with python.requests. When logging the response it was always the MagicMock __enter__ of the context manager, which i am not sure how to handle.

Also not sure how to handle the existing tests which fail now due to the new parameter on RemotePairlist:
`FAILED test_pairlist.py::test__whitelist_for_active_markets[whitelist1-is not compatible with exchange-RemotePairList] - freqtrade.exceptions.OperationalException: `pairlist_url` not specified. Please check your configuration for "pairlist.config.pairlist_url"`

xmatthias:
You'll notice that the fix for the tests is quite / rather simple.

They're not supposed to run for this pairlist (Which requires special treatment / mocking) - so we can simply disable them.

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -0,0 +1,166 @@
...
+        :param tickers: Tickers (from exchange.get_tickers). May be cached.
+        :return: new whitelist
+        """
+
+        # Validate whitelist to only have active market pairs
+        pairlist = self._whitelist_for_active_markets(pairlist)
+        pairlist = self.verify_blacklist(pairlist, logger.info)
+        # Limit pairlist to the requested number of pairs
+        pairlist = pairlist[:self._number_pairs]
+        self.log_once(f"Searching {self._number_pairs} pairs: {pairlist}", logger.info)
```
xmatthias: We should remove this from the pairlists. 
It's quite very misleading (it was misleading in VolumePairlist while it was there) - as it's not reflecting the finally used pairlist.

it's issued at the root level now - once ALL filters have ran - reflecting the actual pairlist that will be used.

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -0,0 +1,166 @@
...
+
+            if "application/json" in str(content_type):
+                jsonparse = response.json()
+                pairlist = jsonparse['pairs']
+                info = jsonparse.get('info', '')
+            else:
+                raise OperationalException('RemotePairList is not of type JSON abort ')
+
+            self._refresh_period = jsonparse.get('refresh_period', self._refresh_period)
+            self._pair_cache = TTLCache(maxsize=1, ttl=self._refresh_period)
```
xmatthias: Why are you initializing the cache several times? 

this line does only make sense once - in `__init__()`.

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -0,0 +1,166 @@
...
+                                    timeout=self._read_timeout)
+            content_type = response.headers.get('content-type')
+            time_elapsed = response.elapsed.total_seconds()
+
+            if "application/json" in str(content_type):
+                jsonparse = response.json()
+                pairlist = jsonparse['pairs']
+                info = jsonparse.get('info', '')
+            else:
+                raise OperationalException('RemotePairList is not of type JSON abort ')
```
xmatthias: Is this really intentional?

`OperationalException` causes the bot to stop (and restart in case of docker).

I think getting invalid responses can be a temporary problem - so maybe stopping is fine on the first attempt - but should be handled differently on subsequent requests? (especially considering that we're handling regular request exceptions differently).

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -0,0 +1,166 @@
...
+        :param pairlist: pairlist to filter or sort
+        :param tickers: Tickers (from exchange.get_tickers). May be cached.
+        :return: new whitelist
+        """
+
+        # Validate whitelist to only have active market pairs
+        pairlist = self._whitelist_for_active_markets(pairlist)
+        pairlist = self.verify_blacklist(pairlist, logger.info)
+        # Limit pairlist to the requested number of pairs
+        pairlist = pairlist[:self._number_pairs]
```
xmatthias: This whole function is odd, and i think it's misused / misinterpreted.

while you use it to filter your pairlist (which is fine) - it's actually _intended_ to be called by the pairlist manager in cases when the pairlist is not the first pairlist (it's not the pairlist generator, just a filter).

If that were the case (there is another pairlist before this), then this would simply replicate behavior of an offsetpairlist (more or less) - completely ignoring the actual remote URL (it won't cross-check against that URL, nor will it add missing pairs from the remote source, nor will it even call the URL, since in that case, `gen_pairlist()` is never called, and subsequently also `fetch_pairlist()`).

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -0,0 +1,166 @@
...
+            time_elapsed = response.elapsed.total_seconds()
+
+            if "application/json" in str(content_type):
+                jsonparse = response.json()
+                pairlist = jsonparse['pairs']
+                info = jsonparse.get('info', '')
+            else:
+                raise OperationalException('RemotePairList is not of type JSON abort ')
+
+            self._refresh_period = jsonparse.get('refresh_period', self._refresh_period)
```
xmatthias: i'm not sure i like this - do we really want a remote source (a source we don't control) to update our cache settings?
it can be the maximum frequency allowed - but just because a source allows a refresh every minute doesn't mean that's what i want for my strategy.

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -0,0 +1,166 @@
...
+                    raise ValueError(f"{self._pairlist_url} does not exist.")
+            else:
+                # Fetch Pairlist from Remote URL
+                pairlist, time_elapsed, info = self.fetch_pairlist()
+
+        pairlist = self.filter_pairlist(pairlist, tickers)
+        self._pair_cache['pairlist'] = pairlist.copy()
+
+        if time_elapsed != 0.0:
+            self.log_once(f'{info} Fetched in {time_elapsed} seconds.', logger.info)
```
xmatthias: I'd be very careful with this.
we're efectively directly logging remote content.

the pairlist is different, as that's filtered first (so only actual valid pairs would be logged) - but here we could get anything (also malicious) - thinking back to log4J last year - that's not something i'd want to find bot users in ...

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: The idea was that a remotesource maybe would not like it to have the pairlist refreshed every 1 minute, if the remote pairlist for example only updates every 30 minutes, since that would be a waste of resources and no point, thinking about it the better way would be to allow the minimum refresh period to be set from the remote end and if its higher keep the local setting. + Maybe logging that change or current refresh_period on a update.

Would that be ok? Allow the remote end only to change it on the first fetch, so it will be noticed?

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: It was done to reinitialize the cache with the new refresh_period (ttl) from remote end, however it would be best to init it after the first fetch then, and not every iteration.

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: I agree, stopping on the first attempt would be better, additional requests should not be stopped out.

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: Did not give it much thought as valid json was parsed then allready, but yes it could cause issues.

We could check for a max-length of 256 chars, only alphanumeric and wrap it in a try: except: clause, would that be sufficient or would it have to be left out?

xmatthias: what i mean is - what's the (expected) benefit of logging an additional info?

Limiting to something is fine - but once we need to do full input validation (or do whatever to be sure it's not problematic) i'd question if we "really" need it.

Bloodhunter4rc: it could return additional statistics, or return filtered pairs for example i used it for 

"info":"Fear and Greed:26, BTC:16723.52USD, BTC Dominance:39.14%, BTC price changed by +0.29% from yesterday to today. Updated:13:01UTC."

or return the sorting / filters used:
"info":" # Pairs: 40, Filtered: 14, Sort by marketcap (50) desc + rsi_14 (40) desc

its not per se needed but it could give information about the pairlist.

xmatthias: keep in mind that it's information that's only shown in the logs - so you'll need to actively look for it to actually see this.

freqtrade/plugins/pairlist/RemotePairList.py
xmatthias: i'm not against changing this per se (if the config doesn't specify a refresh_period, for example) - but i think the "final say" should be with the strategy - at least if it's higher there.
it should also be respected on an ongoing basis (not just at startup) 
we assume freqtrade to be a process that's running for days/weeks - so if the minimum changed from 1m to 5m, we should respect this in a timely manner, not the next time the process is restarted.

---

just because an api allows a request every minute doesn't mean i actually want a new pairlist every minute - so i think the "this defines the minimum - but use the config setup if it's higher" will be best.

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -39,12 +39,13 @@ def __init__(self, exchange, pairlistmanager,
                 'for "pairlist.config.pairlist_url"')
 
         self._number_pairs = self._pairlistconfig['number_assets']
-        self._refresh_period = self._pairlistconfig.get('refresh_period', 1800)
+        self._refresh_period: int = self._pairlistconfig.get('refresh_period', 1800)
         self._keep_pairlist_on_failure = self._pairlistconfig.get('keep_pairlist_on_failure', True)
-        self._pair_cache: TTLCache = TTLCache(maxsize=1, ttl=self._refresh_period)
+        self._pair_cache: Optional[TTLCache] = None
```
xmatthias: i think this makes it (way) worse 
the TTLCache is still recreated every now and then - but now we have to also check it for None - which seems to add unnecessary complexity.

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        Short whitelist method description - used for startup-messages
+        """
+        return f"{self.name} - {self._pairlistconfig['number_assets']} pairs from RemotePairlist."
+
+    def process_json(self, jsonparse) -> Tuple[List[str], str]:
+
+        pairlist = jsonparse.get('pairs', [])
+        remote_info = jsonparse.get('info', '')[:256].strip()
+        remote_refresh_period = jsonparse.get('refresh_period', self._refresh_period)
```
xmatthias: you might need to `int()` this here.
it's a remote source - can't rely on refresh_period to actually be an integer (could be quoted - could be a character).
it'll obviously need exception handling - but most likely outside of this method.

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: its no must i will remove it from the code to be safe

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: it will have to be recreated in case the remote refresh_period (minimum) changes as theres no way to change a existing cache ttl? So i removed the None Checks again, and let it recreate it once needed. (At start once if its lower than the fetched refresh_period and on additional changes of that)

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: it is cast to int now with exception handling, which just outputs the type, due to it could just remotely be used for any malicious code if displaying the actual exception values with the error.

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: the refresh_period is now increased to the one sent from the server (minimum) if it changes and logged

xmatthias (approved):
LGTM :+1:

monkeyden:
Not sure which of you this question is for but, how does this feature work with (if at all) whitelist?  I have a local file configured

`"pairlists": [
        {
            "method": "RemotePairList",
            "number_assets": 20,
            "pairlist_url": "file:///user_data/data/alt_futures_bybit.json"
        }
    ],`

The code that has a problem with it seems to be this, from backtsting.py
`if len(self.pairlists.whitelist) == 0:
            raise OperationalException("No pair in whitelist.")`

How to configure whitelist when the list of tradable assets is dynamically driven?

Bloodhunter4rc:
> Not sure which of you this question is for but, how does this feature work with (if at all) whitelist? I have a local file configured
> 
> `"pairlists": [ { "method": "RemotePairList", "number_assets": 20, "pairlist_url": "file:///user_data/data/alt_futures_bybit.json" } ],`
> 
> The code that has a problem with it seems to be this, from backtsting.py `if len(self.pairlists.whitelist) == 0: raise OperationalException("No pair in whitelist.")`
> 
> How to configure whitelist when the list of tradable assets is dynamically driven?

Same like a VolumePairList , RemotePairList does not support Backtesting.

PR #7868, opened 2022-12-07
Fix add state info bt

robcaulk:
`add_state_info` was not properly deactivated during backtesting. This PR checks the dataprovider during initialization of the RL environment to ensure that `add_state_info` will only be active during dry/live runs.

### DIFF
--- a/freqtrade/freqai/RL/BaseEnvironment.py
+++ b/freqtrade/freqai/RL/BaseEnvironment.py
@@ -83,6 +83,7 @@ class BaseEnvironment(gym.Env):
         if dp:
             self.live = dp.runmode in (RunMode.DRY_RUN, RunMode.LIVE)
         if not self.live and self.add_state_info:
+            self.add_state_info = False
             logger.warning("add_state_info is not available in backtesting. Deactivating.")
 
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/RL/BaseEnvironment.py
```diff
@@ -188,7 +195,7 @@ def _get_observation(self):
         """
         features_window = self.signal_features[(
             self._current_tick - self.window_size):self._current_tick]
-        if self.add_state_info:
+        if self.add_state_info and self.live:
```
xmatthias, line 198: isn't this (additional) check redundant, considering the checks in line 85 - which deactivate `add_state_info` ?

xmatthias (approved):

freqtrade/freqai/RL/BaseEnvironment.py
robcaulk: Yes, in the current form it is. But I wrote this for explicit readability to other developers. i.e. "we CANNOT use state info unless we are live." Further, if a user decides to take control of `self.add_state_info` somewhere in their own custom model, then this check will no longer be redundant.

PR #7883, opened 2022-12-11
fix bug in MultiOutput* with conv_width = 1

robcaulk:
MultiOutputClassifier/Regressor was not working with single candle predictions due to the flat shape of 1D np.arrays being incompatible with multi column dataframe creation.

This MR fixes it by reshaping the prediction array if the user is using `conv_width == 1` (the default setting).


xmatthias (dismissed):
LGTM :+1:

xmatthias (approved):

PR #7898, opened 2022-12-14
fix doc minimal_roi

initrv:
## Summary

Extra brackets removed


froggleston (approved):
Looks OK to me. Good spot.

PR #7899, opened 2022-12-14
Ensure data provider is passed to multiproc envs

robcaulk:
This PR fixes a bug associated with `add_state_info` not automatically deactivating in backtesting runs. 

It also fixes a bug where the `self.live` information was not properly passed to the `_multiproc` env. 

Finally, it improves code readability by packing large amounts of env passed args into a dictionary.

### DIFF
--- a/freqtrade/freqai/RL/BaseEnvironment.py
+++ b/freqtrade/freqai/RL/BaseEnvironment.py
@@ -44,8 +44,8 @@ class BaseEnvironment(gym.Env):
 
     def __init__(self, df: DataFrame = DataFrame(), prices: DataFrame = DataFrame(),
                  reward_kwargs: dict = {}, window_size=10, starting_point=True,
-                 id: str = 'baseenv-1', seed: int = 1, config: dict = {},
-                 env_info: dict = {}):
+                 id: str = 'baseenv-1', seed: int = 1, config: dict = {}, live: bool = False,
+                 fee: float = 0.0015):
         """
         Initializes the training/eval environment.
         :param df: dataframe of features
@@ -67,12 +67,12 @@ class BaseEnvironment(gym.Env):
         if self.config.get('fee', None) is not None:
             self.fee = self.config['fee']
         else:
-            self.fee = env_info.get('fee', 0.0015)
+            self.fee = fee
 
         # set here to default 5Ac, but all children envs can override this
         self.actions: Type[Enum] = BaseActions
         self.tensorboard_metrics: dict = {}
-        self.live = env_info.get('live', False)
+        self.live = live
         if not self.live and self.add_state_info:
             self.add_state_info = False
             logger.warning("add_state_info is not available in backtesting. Deactivating.")
--- a/freqtrade/freqai/RL/BaseEnvironment.py
+++ b/freqtrade/freqai/RL/BaseEnvironment.py
@@ -56,7 +56,8 @@ class BaseEnvironment(gym.Env):
         :param id: string id of the environment (used in backend for multiprocessed env)
         :param seed: Sets the seed of the environment higher in the gym.Env object
         :param config: Typical user configuration file
-        :param env_info: Environment info dictionary, used to pass live status, fee, etc.
+        :param live: Whether or not this environment is active in dry/live/backtesting
+        :param fee: The fee to use for environmental interactions.
         """
         self.config = config
         self.rl_config = config['freqai']['rl_config']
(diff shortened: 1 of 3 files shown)


freqtrade/freqai/prediction_models/ReinforcementLearner_multiproc.py
```diff
@@ -37,14 +37,14 @@ def set_train_and_eval_environments(self, data_dictionary: Dict[str, Any],
         env_id = "train_env"
         self.train_env = SubprocVecEnv([make_env(self.MyRLEnv, env_id, i, 1, train_df, prices_train,
                                         self.reward_params, self.CONV_WIDTH, monitor=True,
-                                        config=self.config) for i
+                                        config=self.config, dp=self.data_provider) for i
```
xmatthias: are you sure this actually works? (especially looking at dry/live...).

in my understanding of python - creating a sub-process means that all objects need to be pickled, and will then be unpickled in the other process. Maybe `SubprocVecEnv` (or whatever is responsible for creating the subprocesses) uses some different mechanics under the hood so it works - but standard python / joblib it will not.

this is impossible for objects with open sockets (open network connections) - which dataprovider will have implicitly by having an exchange attached to it.

This is untrue in hyperopt (for that very same reason) - so i

aemr3: It doesn't work yes, I am pushing another commit to fix it.

aemr3: Basically we are now passing an info dict to base environment for additional data, and the backtesting bug was because of the reset_env being called before setting add_state_info to false.

freqtrade/freqai/RL/BaseEnvironment.py
```diff
@@ -59,32 +56,28 @@ def __init__(self, df: DataFrame = DataFrame(), prices: DataFrame = DataFrame(),
...
         :param config: Typical user configuration file
-        :param dp: dataprovider from freqtrade
+        :param env_info: Environment info dictionary, used to pass live status, fee, etc.
         """
         self.config = config
         self.rl_config = config['freqai']['rl_config']
         self.add_state_info = self.rl_config.get('add_state_info', False)
         self.id = id
-        self.seed(seed)
-        self.reset_env(df, prices, window_size, reward_kwargs, starting_point)
```
robcaulk, line 69: good find on that one - thank you

freqtrade/freqai/prediction_models/ReinforcementLearner_multiproc.py
```diff
@@ -34,17 +35,23 @@ def set_train_and_eval_environments(self, data_dictionary: Dict[str, Any],
         train_df = data_dictionary["train_features"]
         test_df = data_dictionary["test_features"]
 
+        env_info = {"live": False}
+        if self.data_provider:
+            env_info["live"] = self.data_provider.runmode in (RunMode.DRY_RUN, RunMode.LIVE)
```
robcaulk: I think we have `self.live` already available here since `ReinforcmentLearner*` is a child of `IFreqaiModel` see [here](https://github.com/freqtrade/freqtrade/blob/fix/multiproc-dp/freqtrade/freqai/freqai_interface.py#L133).

```python
env_info["live"] = self.live
```

robcaulk (commented):
Thanks, just a couple comments. I'll add the change for loading more arguments into the dict, that should help clean up the code a bit here.

freqtrade/freqai/prediction_models/ReinforcementLearner_multiproc.py
aemr3: That's better, thanks.

freqtrade/freqai/RL/BaseEnvironment.py
```diff
@@ -59,32 +56,28 @@ def __init__(self, df: DataFrame = DataFrame(), prices: DataFrame = DataFrame(),
         :param id: string id of the environment (used in backend for multiprocessed env)
         :param seed: Sets the seed of the environment higher in the gym.Env object
         :param config: Typical user configuration file
-        :param dp: dataprovider from freqtrade
+        :param env_info: Environment info dictionary, used to pass live status, fee, etc.
```
aemr3: I think we will need to update function doc as well.

xmatthias (approved):
LGTM :+1:

PR #7906, opened 2022-12-16
fix base4 env done condition

initrv:
## Summary

A small fix to make the episodes termination state identical in the base environments


robcaulk (approved):
LGTM

PR #7908, opened 2022-12-16
Add 3 Action RL Env

aemr3:
## Summary

This PR adds Base3ActionRLEnv to the RL module.  Also can_short param is added to the BaseEnvironment.
This environment can be used with both long only strategies (i.e spot markets) and short enabled strategies.


robcaulk (changes requested):
Thanks for this. It needs two more components:
1) we need to test the can_short = False. I think we can do that by creating a variant of:
`tests/strategy/strats/freqai_rl_test_strat.py`

2) we should add documentation for the 3 environments, and some note to indicate that the 3Ac can also work with long only bots. 

I will try to sort these both out, then I will ask you to review them :-)

Rob

aemr3:
> Thanks for this. It needs two more components:
> 
> 1. we need to test the can_short = False. I think we can do that by creating a variant of:
>    `tests/strategy/strats/freqai_rl_test_strat.py`
> 2. we should add documentation for the 3 environments, and some note to indicate that the 3Ac can also work with long only bots.
> 
> I will try to sort these both out, then I will ask you to review them :-)
> 
> Rob

Thank you for the changes. It looks good to me. Are you going to change the strategy also? Maybe we can also modify reward function to show long only version.

robcaulk (approved):
LGTM

PR #7921, opened 2022-12-19
remove binance ad from website

robcaulk:
Binance may or may not follow the same path as FTX. I see no reason to drive open-source users to any particular exchange like this. 

~~Not to mention this violates RTD.~~


xmatthias (changes requested):
you can't remove the overrides/main.html file without breaking tons of stuff on the documentation (the version selector at the bottom for example).

To remove just the AD (Which i understand is what you aim to do with this) - just remove that section of the document - but not the rest.

> Not to mention this violates RTD.

it doesn't - RTD terms doesn't prevent other ads on the page afaik.

robcaulk:
> you can't remove the overrides/main.html file without breaking tons of stuff on the documentation (the version selector at the bottom for example).
> 
> To remove just the AD (Which i understand is what you aim to do with this) - just remove that section of the document - but not the rest.
> 
> > Not to mention this violates RTD.
> 
> it doesn't - RTD terms doesn't prevent other ads on the page afaik.

Which section of which document? I first tried to only remove:

```html
    <!-- Load binance SDK -->
    <script async defer src="https://public.bnbstatic.com/static/js/broker-sdk/broker-sdk@1.0.0.min.js"></script>

    <script>
      window.onload = function () {
        var sidebar = document.getElementById('widget-wrapper')
        var newDiv = document.createElement("div");
        newDiv.id = "widget";
        try {
          sidebar.prepend(newDiv);

          window.binanceBrokerPortalSdk.initBrokerSDK('#widget', {
            apiHost: 'https://www.binance.com',
            brokerId: 'R4BD3S82',
            slideTime: 4e4,
          });
        } catch(err) {
          console.log(err)
        }
      }
    </script>
```
    
But it left a big ugly empty white box on the website.
    
So please elaborate, in particular, can you please be more explicit about what is "the section" and what is "the document?"

xmatthias:
The "document" will be the one you removed in this PR - but full removal is wrong (as mentioned above).

The section looks "roughly correct" - but is most likely incomplete - as it'll leave the container behind (as you say so yourself).

> But it left a big ugly empty white box on the website.

that's the container. 
i don't know by memory what to look for.

If the expectation from a reviewer is "you must remove line yyy in document xxx" - then that's a sign that the PR is out of your comfort zone (not sure how good your HTML / JS is). 
That's also perfectly fine - not everyone needs to be an expert in anything.

I'm happy to look at it eventually when i find time (it's no priority to me at the moment) - but definitely not based on this PR (at least not in it's current state) - as the only commit is 100% wrong - and will therefore need full reverting - so basing changes on top of this is more work than starting from a clean develop branch.

robcaulk:
> it's no priority to me at the moment

Well, I think that says it all.

xmatthias:
> Well, I think that says it all.

Quite frankly - you're proposing a PR - and upon receiving a review that changes are needed - I'm asked to "what lines do i need to change - what i tried didn't work".
Guess what - i don't know the solution precisely without actually looking at both the source - as well as how it renders after some parts of the change (and then analyze which parts still need to be removed).
This basically results in a "i need to do it" outcome - where i can then either tell you precisely which changes you need to make - or do the change myself (which will be faster).

As stated above - I'm fine with that.

if you feel upset about the fact that something you want to change is not important enough to me to either fix immediately, or spoon-feed you the solution "right now" - resulting in "I'm upset and close this pr"- then apparently it's not something important anyway.

moinonin:
"Binance may or may not follow the same path as FTX. I see no reason to drive open-source users to any particular exchange like this." 

First of all lets address this why the ad was put in the first place before thinking of how to remove. Is it helping in contributing to the main developer? is it bringing some income? I don't think removing it is the most difficult thing. Agreeing to remove or not is rather the difficult one IMHO.

robcaulk:
@moinonin good question.

In fact - we have discussed this at length in DM. The ad supports CCXT, it was put there based on a mutual agreement a couple years ago (I am just relaying my understanding, it could be incorrect). 

Clearly FT has a good reason to help keep CCXT running. But at this point it is, in my opinion, it is inappropriate to direct users to "auto-invest" in binance. So we also discussed the idea of simply changing the ad to a cloud provider solution (maybe vultr or oracle). But it seems that removing it entirely is simpler (and especially now that CCXT is no longer as small of a project).

PR #7922, opened 2022-12-20
Add an option drop last shifted labels while backtesting

aemr3:
## Summary

Added an option to drop last shifted labels while backtesting. This helps to simulate dry runs better.

- [ ] Update docs


aemr3:
Closing this due to a different change that will be done.

PR #7925, opened 2022-12-20
(WIP) - FreqAI Backtesting - fix call to populate_any_indicators

wagnercosta:
## Summary

Currently in the FreqAI backtest, the `populate_any_indicators` function is called only once for the entire dataframe. This PR aims to change the behavior so that the labels are calculated according to the sliced dataframe, simulating more closely the behavior in dry/run live.

- [x] Create draft solution
- [ ] Discuss the proposed solution
- [ ] Test different scenarios
- [ ] Fix tests
- [ ] Documentation

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -1184,6 +1184,7 @@ class FreqaiDataKitchen:
         pair: str = "",
         prediction_dataframe: DataFrame = pd.DataFrame(),
         do_corr_pairs: bool = True,
+        set_only_targets: bool = False
     ) -> DataFrame:
         """
         Use the user defined strategy for populating indicators during retrain
@@ -1222,7 +1223,8 @@ class FreqaiDataKitchen:
                 dataframe.copy(),
                 tf,
                 informative=base_dataframes[tf],
-                set_generalized_indicators=sgi
+                set_generalized_indicators=sgi,
+                set_only_targets=set_only_targets
             )
 
         # ensure corr pairs are always last
@@ -1235,7 +1237,8 @@ class FreqaiDataKitchen:
                         corr_pair,
                         dataframe.copy(),
                         tf,
-                        informative=corr_dataframes[corr_pair][tf]
+                        informative=corr_dataframes[corr_pair][tf],
+                        set_only_targets=set_only_targets
                     )
 
         self.get_unique_classes_from_labels(dataframe)
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -149,12 +149,13 @@ class IFreqaiModel(ABC):
         # the concatenated results for the full backtesting period back to the strategy.
         elif not self.follow_mode:
             self.dk = FreqaiDataKitchen(self.config, self.live, metadata["pair"])
-            dataframe = self.dk.use_strategy_to_populate_indicators(
-                strategy, prediction_dataframe=dataframe, pair=metadata["pair"]
-            )
+
+            # dataframe = self.dk.use_strategy_to_populate_indicators(
+            #     strategy, prediction_dataframe=dataframe, pair=metadata["pair"]
+            # )
             if not self.config.get("freqai_backtest_live_models", False):
                 logger.info(f"Training {len(self.dk.training_timeranges)} timeranges")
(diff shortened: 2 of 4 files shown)


robcaulk (changes requested):
I appreciate the time you spent to pull together a solution to the problem. As you and I discussed in DM, I am going to try to summarize some of our conclusions below.

I think there are two clear points that this PR reveals: 
* backtesting *needs* a fix that allows users to recompute targets for each window (exactly what you have proposed here).
* **BUT** there is no way to implement this fix (in a computationally performant way) without asking the user to make large adjustments to their `populate_any_indicators()`.

So if we agree on these two things, then to me it is an opportunity to improve the user interface.

We may consider something like splitting off the target definition into its own function (then we control how and when it is called on the back-end, thus gaining the performance and backtesting bug fix simultaneously). It also feels more organized, instead of hiding label definition behind a flag (`set_generalized_indicators()`). In fact, I have seen multiple users who believe that they need to change the default argument of `set_generalized_indicators` to `True` in order for that part of the code to get executed. Obviously this indicates that the users are also not familiar with Python basics, but I think it highlights how confusing the current `populate_any_indicators()` is in general. 

While doing this big change, we might as well take advantage of how it will all be broken. I'd love to move the `include_shifted_candles` and the df merging/col dropping into our `freqai/utils.py` so that they are single function calls instead of having the user handle them here (these two functionalities should never by changed by the user, so it makes sense to make the user side more readable and abstracted). If we want to be really slick we can even abstract away the loop on `indicator_periods_candles`. 

Finally I guess since we are making all these structural changes, I'd like to rename the function from `populate_any_indicators()` to something more descriptive (e.g. `feature_engineering()` or `signal_analysis()`.)  

I am more than happy to get my hands dirty in this re-org and do most of the things I have suggested here, but please let me know if you have any suggestions or if you would like to take the first stab at it.

aemr3:
> While doing this big change, we might as well take advantage of how it will all be broken. I'd love to move the `include_shifted_candles` and the df merging/col dropping into our `freqai/utils.py` so that they are single function calls instead of having the user handle them here (these two functionalities should never by changed by the user, so it makes sense to make the user side more readable and abstracted). If we want to be really slick we can even abstract away the loop on `indicator_periods_candles`.
> 
> Finally I guess since we are making all these structural changes, I'd like to rename the function from `populate_any_indicators()` to something more descriptive (e.g. `feature_engineering()` or `signal_analysis()`.)
> 
> I am more than happy to get my hands dirty in this re-org and do most of the things I have suggested here, but please let me know if you have any suggestions or if you would like to take the first stab at it.

I agree, this is a nice opportunity to make those changes. It will significantly improve user experience. Having shifted candles and merge logic inside feature engineering function is difficult to understand for new-comers. Just want to know would it be possible to have a section like set_generalized_indicators so that additional custom feature extraction can be executed only once?
+1 to feature_engineering()

xmatthias:
> Finally I guess since we are making all these structural changes, I'd like to rename the function from `populate_any_indicators()` to something more descriptive (e.g. `feature_engineering()` or `signal_analysis()`.)

Is your intend to have this as a breaking change (users MUST update their strategy if they update freqtrade to a version beyond X) - or will there be a compatibility layer to keep users running their already existing strategies? (this almost certainly has a cost of more complexity in the  calling code).

In general, i think we should try to avoid breaking changes as much as possible.
That always results in a bad user experience (i wanted to do a simple update, now everything is broken and i need to understand whatever changes i need to make ...).

robcaulk:
> > Finally I guess since we are making all these structural changes, I'd like to rename the function from `populate_any_indicators()` to something more descriptive (e.g. `feature_engineering()` or `signal_analysis()`.)
> 
> Is your intend to have this as a breaking change (users MUST update their strategy if they update freqtrade to a version beyond X) - or will there be a compatibility layer to keep users running their already existing strategies? (this almost certainly has a cost of more complexity in the calling code).
> 
> In general, i think we should try to avoid breaking changes as much as possible. That always results in a bad user experience (i wanted to do a simple update, now everything is broken and i need to understand whatever changes i need to make ...).

Ideally a deprecation warning for a month or two followed by full removal. But depends on back end complexity. My current sense is that it won’t be too difficult to carry both methods simultaneously using some pretty basic, but difficult to read, if statements (with the deprecation log warning), that just change the path of the code back end code.

robcaulk:
This PR is superceded by #7950 .

PR #7936, opened 2022-12-23
add rl_config.device param to support use apple chip GPU

threadshare:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/freqai/prediction_models/ReinforcementLearner.py
+++ b/freqtrade/freqai/prediction_models/ReinforcementLearner.py
@@ -58,10 +58,16 @@ class ReinforcementLearner(BaseReinforcementLearningModel):
                              net_arch=self.net_arch)
 
         if dk.pair not in self.dd.model_dictionary or not self.continual_learning:
+            kwargs = self.freqai_info.get('model_training_parameters', {})
+
+            # set device if device is not None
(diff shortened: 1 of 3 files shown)


freqtrade/freqai/prediction_models/ReinforcementLearner.py
```diff
@@ -58,10 +58,16 @@ def fit(self, data_dictionary: Dict[str, Any], dk: FreqaiDataKitchen, **kwargs):
...
+
+            # set device if device is not None
+            if self.device != '':
+                kwargs['device'] = self.device
+
             model = self.MODELCLASS(self.policy_type, self.train_env, policy_kwargs=policy_kwargs,
                                     tensorboard_log=Path(
                                         dk.full_path / "tensorboard" / dk.pair.split('/')[0]),
-                                    **self.freqai_info.get('model_training_parameters', {})
+                                    **kwargs
```
xmatthias, line 70: I actually think that what you try to accomplish can actually be done already.

you can put the following in your config - which will be the Kwargs to the models.
As these vary by model - it's better to keep this generic.

``` json
"model_training_parameters": {
    "device": "GPU"
}
```

PR #7951, opened 2022-12-27
strategy_updater

hippocritical:
## Summary

This change is targeted at people who want to upgrade their strategy files. It uses python => ast to achieve that.
This currently works like list-strategies in that you give it a list of strategies that you want to have it convert.
It makes backups of the converted files (but currently not by timestamp, so be careful of firing it multiple times without backing it up.)

Solve the issue: 
People don't upgrade their strategies to the v3 syntax

## Quick changelog

added strategy_utils command and code 

## What's new?

This will solve the issue that people refuse updating their stategies (hopefully)


freqtrade/strategy/strategy_updater.py
```diff
@@ -0,0 +1,213 @@
+import ast
+import os
+import shutil
+
+
+class strategy_updater:
```
xmatthias: (nitpick): ClassNames "should" be in CamelCase in python

freqtrade/strategy/strategy_updater.py
```diff
@@ -0,0 +1,213 @@
...
+            _rename_dict)
+
+        # return the modified code without executing it
+        return updated_code
+
+    # function that uses the ast module to update the code
+    def modify_ast(node, _name_mapping, _function_mapping, _rename_dict): # noqa
+        # create a visitor that will update the names and functions
+        class NameUpdater(ast.NodeTransformer):
+            def generic_visit(self, node):
```
xmatthias: does this HAVE to be a nested class?

it has 2 downsides i can see so far
* it increases indentation by 8 spaces for no good reason (which automatically wraps more lines than we'd want)
* field names are now confusing 
  (what is `node` in the function below? the `node` from `modify_ast()` is no longer accessible, as it's covered by the `node` definition in this method ...)

freqtrade/strategy/strategy_updater.py
```diff
@@ -0,0 +1,213 @@
...
+                if hasattr(node.slice, "elts"):
+                    for elt in node.slice.elts:
+                        if isinstance(elt, ast.Constant) and \
+                                elt.value in strategy_updater.rename_dict:
+                            elt.value = strategy_updater.rename_dict[elt.value]
+                return node
+
+            def visit_Constant(self, node):
+                # do not update the names in import statements
+                if node.value in \
```
xmatthias: we should try to not wrap lines with `\` (if possible, obviously).
i've seen that break in mixed windows/linux environments - so it suddenly simply no longer works - therefore i became somewhat allergic against this :laughing: 

in if statements, it's better to wrap it all in parentheses `()`  (obviously not just in this line)

freqtrade/strategy/strategy_updater.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def start(self, strategy_obj: dict) -> None:
+        """
+        Run strategy updater
+        It updates a strategy to v3 with the help of the ast-module
+        :return: None
+        """
+
+        self.cwd = os.getcwd()
+        self.strategies_backup_folder = f'{os.getcwd()}user_data/strategies_orig_updater'
```
xmatthias: please use [pathlib](https://docs.python.org/3/library/pathlib.html) for directory operations. You'll find that it has almost all methods available as os.<directoryoperation>.

This line will be wrong on windows (wrong slash) - so it'll either fail to open - or create a file in the cwd named as the string above (not sure - didn't use os.path for 2-3 years on windows).

freqtrade/strategy/strategy_updater.py
```diff
@@ -0,0 +1,213 @@
...
+    rename_dict = {'buy': 'enter_long', 'sell': 'exit_long', 'buy_tag': 'enter_tag'}
+
+    def start(self, strategy_obj: dict) -> None:
+        """
+        Run strategy updater
+        It updates a strategy to v3 with the help of the ast-module
+        :return: None
+        """
+
+        self.cwd = os.getcwd()
```
xmatthias: without looking what's done with this - it's wrong.

`cwd` is technically  irrelevant (at least at this nesting level within freqtrade) - user_data could be `(cwd)/user_data`  (the default) - but it could just as well be `(cwd)/../whatever/...` - given as `freqtrade <subcommand> --user-data ../whatever` ...

xmatthias (commented):
didn't play with it yet, really - just did a quick fly through the code

Assuming it does what you say (which i do trust it does) - looks quite good already!

freqtrade/strategy/strategy_updater.py
hippocritical: those \ ones were primarily just cause either pycharm or github argued about too many characters in a row ...
yeah will be fixing that as much as possible
to me, who cares if a row is 105 characters instead of max 100

freqtrade/strategy/strategy_updater.py
hippocritical: it s just used to get the base folder of freqtrade itself, but yeah - will look into how others did it inside freqtrade

hippocritical:
pushed a new commit with several fixes, tests will come when the base code looks ok.

hippocritical:
Some strategies produce the following error message (as an example for moved modules):
freqtrade.resolvers.iresolver - WARNING - Could not import [FILE] due to 'cannot import name 'CategoricalParameter' from 'freqtrade.strategy.hyper' and thereby cannot be loaded since the starting point is strategy-list and the class name is an empty string for this object. (the import now is freqtrade.strategy)

How to solve this issue best? 
Currently the IResolver._get_valid_object naturally fails and just returns an empty class name and thereby will never be updated.

Solution1:
edit the _get_valid_object function to still get the class name
Solution2:
make an own function with ast that reads the class names and use that
Solution3:
?

xmatthias:
> edit the _get_valid_object function to still get the class name

you can't (not reliably, anyway).
Since the import fails - python won't go any further (it'll never get to the line with the class).
The only way to read the class name would then be to "grep" for the class (`grep -e '^class \S+\:$`) (basically string search)  - but that's kinda flaky / not reliable.

i'm not sure AST can work around a failing import either - in my understanding, it still requires a somewhat valid python file to work from - so i think with a failing import, you'll struggle to actually get it to AST in the first place.

---

for a first iteration, i'd clearly just skip wrong imports. 
it's going to be complex, flaky logic at best...

hippocritical:
yes ast needs "valid code", but in the start I disabled the full checking of the code.
It is that it does not NEED to be fully valid anymore.
Meaning by that : it does not check imports, it does not calculate through logic, etc.
so it could load the file containing "import aModuleThatDoesntExist"

Well ok, as you said this is for the future-me. Will keep the import change in it for the future and set a market that it is not triggering currently. Then going at tests and after that docs.

hippocritical:
added a ton of tests split into digestible chunks

requirements.txt
```diff
@@ -57,3 +57,23 @@ schedule==1.1.0
...
+scipy~=1.9.3
+xgboost~=1.7.2
+catboost~=1.1.1
+lightgbm~=3.3.3
+astor~=0.8.1
+ta~=0.10.2
+finta~=1.3
+tapy~=1.9.1
+matplotlib~=3.6.2
+PyYAML~=6.0
```
xmatthias: this ... doesn't seem to be an intentional change.

It'll duplicate several dependencies we have in other requirements files (for example the plot and freqAI dependencies).

i think the only valid new dependency is `astor`. the others are either in other requirements files  (freqai, pytest, ...) - or freqtrade doesn't depend on them (e.g. finta).

We'll not install all possible ta libraries - nor will we force every dependency on every user (e.g.: torch is >500MB - and only needed if you use freqAI).

all dependencies we depend on will be installed with `pip install -r requirements-dev.txt` - but for example pytest is not necessary for regular users (non developers).

freqtrade/strategy/strategyupdater.py
```diff
@@ -0,0 +1,260 @@
...
+            os.makedirs(strategies_backup_folder)
+
+        # backup original
+        # => currently no date after the filename,
+        # could get overridden pretty fast if this is fired twice!
+        # The folder is always the same and the file name too (currently).
+        shutil.copy(source_file, target_file)
+
+        # update the code
+        new_code = StrategyUpdater.update_code(self, old_code)
```
xmatthias: i'm surprised this does even work.

update_code is not a static method (and we're in an instance method anyway) - so the correct way to call this would be 
`new_code = self.update_code(old_code)`

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -0,0 +1,42 @@
...
+                found = True
+                break
+
+        if not found:
+            print(f"strategy {strategy_obj['name']} could not be loaded or found and is skipped.")
+
+    for filtered_strategy_obj in filtered_strategy_objs:
+        # Initialize backtesting object
+        instance_strategy_updater = StrategyUpdater()
+        StrategyUpdater.start(instance_strategy_updater, config, filtered_strategy_obj)
```
xmatthias: same here, effectively ...
```suggestion
        self.start(config, filtered_strategy_obj)
```

freqtrade/strategy/strategyupdater.py
```diff
@@ -0,0 +1,260 @@
...
+
+        source_file = strategy_obj['location']
+        print(f"started conversion of {source_file}")
+        strategies_backup_folder = Path.joinpath(config['user_data_dir'], "strategies_orig_updater")
+        target_file = Path.joinpath(strategies_backup_folder, strategy_obj['location_rel'])
+
+        # read the file
+        with open(source_file, 'r') as f:
+            old_code = f.read()
+        if not os.path.exists(strategies_backup_folder):
```
xmatthias: are you sure this is the correct test for this?

i think `strategies_backup_folder.is_dir()` would be more fitting - "exists" also yields true if it's a file that exists.

freqtrade/strategy/strategyupdater.py
```diff
@@ -0,0 +1,260 @@
...
+            f.write(new_code)
+        print(f"conversion of file {source_file} successful.")
+
+    # define the function to update the code
+    def update_code(self, code):
+        # parse the code into an AST
+        tree = ast.parse(code)
+
+        # use the AST to update the code
+        updated_code = self.modify_ast(self, tree)
```
xmatthias:
```suggestion
        updated_code = self.modify_ast(tree)
```

xmatthias:
gave this a spin on a pretty simple (but old) strategy (nothing fancy, a few parameters ...) - which resulted in a `KeyError`:

i also noticed that it's taking quite a long time (if you look at the time above, it's almost 1 minute for a 260 line file).


```
2023-01-02 08:46:07,332 - freqtrade.configuration.configuration - INFO - Using strategy list of 1 strategies     
2023-01-02 08:46:07,332 - freqtrade.configuration.config_validation - INFO - Validating configuration ...        
started conversion of /home/xmatt/devel/cryptos/freqtrade/user_data/strategies/TurtleStrategyX3p.py              
                                                                                                                                            
                                                                                                                                            
2023-01-02 08:46:53,507 - freqtrade - ERROR - Fatal exception!                                                                              
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/main.py", line 42, in main                                      
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/commands/strategy_utils_commands.py", line 42, in start_strategy_update
    StrategyUpdater.start(instance_strategy_updater, config, filtered_strategy_obj)                                       
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 72, in start             
    new_code = StrategyUpdater.update_code(self, old_code)                                                                                  
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 84, in update_code     
    updated_code = self.modify_ast(tree)                                                                                                    
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 92, in modify_ast               
    NameUpdater().visit(tree)                                                                                                               
  File "/home/xmatt/.pyenv/versions/3.10.8/lib/python3.10/ast.py", line 412, in visit                                       
    return visitor(node)                                                                                                                    
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 109, in generic_visit
    self.check_fields(field_value)                                                                                                          
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 117, in check_fields
    self.visit(item)                                                                                                                        
  File "/home/xmatt/.pyenv/versions/3.10.8/lib/python3.10/ast.py", line 412, in visit                         
    return visitor(node)                                                                                                                    
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 219, in visit_ClassDef
    return self.generic_visit(node)                                                                                                         
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 109, in generic_visit
    self.check_fields(field_value)                                    
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 117, in check_fields         
    self.visit(item)                                                                                                                        
  File "/home/xmatt/.pyenv/versions/3.10.8/lib/python3.10/ast.py", line 412, in visit      
    return visitor(node)                                              
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 174, in visit_FunctionDef
    self.check_args(node)                                                                                                                     File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 137, in check_args
    self.check_args(node.args)
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/strategy/strategyupdater.py", line 141, in check_args                                     arg.arg = StrategyUpdater.name_mapping[arg.arg]                                                                                         
KeyError: 'self'

```

hippocritical:
yes it takes ages now since we are forced to go through all child nodes recursively again to really get EVERYTHING in there. Going through there curated will not get some things that are nested over and over.
It is still possible to make it go faster, just need to add a check if this is actually an ast.* node and not some string or something like that which makes no sense to check.

Adding those tests might have broken the main code, currently investigating to get both working.

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -39,4 +39,4 @@ def start_strategy_update(args: Dict[str, Any]) -> None:
     for filtered_strategy_obj in filtered_strategy_objs:
         # Initialize backtesting object
         instance_strategy_updater = StrategyUpdater()
-        StrategyUpdater.start(instance_strategy_updater, config, filtered_strategy_obj)
+        self.start(config, filtered_strategy_obj)
```
xmatthias: well no - `instance_strategy_updater.start()` - self refers to ... "nothing" in this case (it's not a class)

xmatthias (commented):
> yes it takes ages now since we are forced to go through all child nodes recursively again to really get EVERYTHING in there.

Doing it recursively is not a problem - but maybe checking if we're still "in this file" - or in some dependencies might make sense?

hippocritical:
have to rewrite the whole shebang, this gets out of hand ... too many conditionals etc. 

Rewriting with ast.walk, this might take some time ... on hold until then.

hippocritical:
found that astor.to_source(tree) makes some code ugly, like this:
    {'params': {'trailing': {'trailing_stop': False,
        'trailing_stop_positive': False, 'trailing_stop_positive_offset': 
        0.0, 'trailing_only_offset_is_reached': False}, 'buy': {
        'atrs_range': 1, 'rsi_buy_hline': 61, 'window_range': 16}, 'sell':
        {}, 'protection': {}, 'roi': {'0': 0.696, '10216': 
        0.40800000000000003, '26070': 0.143, '41881': 0}, 'stoploss': {
        'stoploss': -0.254}}}

There is an option to indent properly again but it errors out currently. Will be fixed at a later point.

hippocritical:
I am confident that this version will be workable.
The conversion time for a simple strategy is 0.0-ish seconds
and for a complicated one 1.x-ish seconds.
Added timers for conversion of a single file to the code base to illustrate that.

hippocritical:
I m still feeling that we should make a version of the IResolver function that does not check the validity of a strategy.
Things that will be depreceated and then made an error (example: ticker_interval) will not be able to be updated.
Might be a thing after the first version is up and running ... I m just not sure about how to do that without screwing the normal procedure and without producing duplicate code.

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -0,0 +1,40 @@
...
+    # Import here to avoid loading backtesting module when it's not used
+    from freqtrade.strategy.strategyupdater import StrategyUpdater
+
+    config = setup_utils_configuration(args, RunMode.UTIL_NO_EXCHANGE)
+
+    strategy_objs = StrategyResolver.search_all_objects(
+        config, enum_failed=False, recursive=config.get('recursive_strategy_search', False))
+
+    filtered_strategy_objs = []
+    for args_strategy in args['strategy_list']:
```
xmatthias: what's the expectation of running `freqtrade strategy-updater` (without arguments)? 

currently it's crashing with `TypeError: 'NoneType' object is not iterable` - in this line.
i'd almost assume that if `--strategy-list` is not given, we'll simply update all strategies?

It's more a question - i'm not sure what your thoughts on this are.

xmatthias (commented):
Speed definitely improved a TON - and the result now seems workable.

At the same time, it's however also removing all comment lines from the file.
I understand they're not part of the code per se - but removing them can have quite some impact.
My test-strategy had some explaining comments, as well as (currently inactive) code - which is now simply gone.

(left is new, right is old)
image


Result formatting seems changed, too. 
That's not per se a problem - but surprising - and honestly, quite ugly (and wrong considering pep8).

image

The line can wrap (it's not 100 chars long though - so why wrap at 80?) - but if it wraps, then the next line needs 4 more spaces - as is, the ) is aligned with the next line, which violates pep8.

dataframe formatting becomes ... bad (there's no other way to put it).

image

(new version again, not in diff view)
image


The new part is ... illegible. without a ton of manual formatting, this is impossible to work on that strategy further. It's surely usable - but the moment you doubt "did i use this condition, or that" - you can't digest it anymore.

That last part however makes the updater pretty much pointless in my view - as it's a lot faster to do the transition manually - as that will not 
* remove potentially important comments
* mess up formatting so the dataframe conditions are no longer leggible 

Now obviously it'll depend on the usecase - if the goal is "i want X strategies transitioned so i can test them" - then this might work (it's not broken code - just very ugly code).

I do see the main usecase however with users who didn't transition to the new format yet (for whatever reason) - and with users that find an interesting (maybe old) strategy and would like to test it. This last point however also contains "potentially add new conditions" - which this format makes impossible without a lot of manual work.

---

Ideally (and this was my expectation) - this would only modify "what's necessary".
so a "diff" between source and destination would be minimal - keeping the code structure (including comments and unnecessary user formattings) intact.

that's not quite what this is doing - and my assumption always was that in order to properly support this, we'd need to use string replace - as all AST based tools would drop comments, as they're non-executable code.

---

One potential "fix" for the formatting is to run a good (!) autoformatter on the code (assuming we can't write the code back with only impacted lines changed?).

I've tried to run black against the result - and it improves formatting quite a lot - but that won't fix the missing comments - which i see as quite crucial to have a good user experience, and to be able to continue to use the (converted) strategy going forward.

hippocritical:
update:
They are currently considering my issue.
They mentioned parso too
https://parso.readthedocs.io/en/latest/index.html
This PR is on hold until they either decline the implementation of comments back via some unparse function or go through with it.
If they deny me that issue being solved I will very likely look into parso, but for the meantime ... waiting game.

hippocritical:
Update: they are working on an update which includes round-trip with comments.
Appears there are many edge cases which have to be worked around, meaning it will take some time.

xmatthias:
is there an issue or PR we can look at to track progress?

hippocritical:
yes there is
https://github.com/t3rn0/ast-comments/issues/6

hippocritical:
update: ast_comments looks nicely now
Ast itself on the other hand does change the logic internally from if elif elif to if else if
This is only happening once and the logic itself still works, just the structure looks funny when you have 50 + tabs now, one for each buy condition of nfi.

This should be an issue with ast itself, dunno why it does that currently - investigating why this happens.

Side node: since there is no mentioning of any empty rows because it is just parsing and unparsing the code itself the empty rows will be gone which the user did input for formatting.
The code itself still has to be "beautified" but that is the last thing that will be done.

Maybe this function should only be used for the ones who are "too lazy" to port something to futures etc?
Stating this since manually updating will keep their layouts and empty rows etc which cannot be preserved.

grafik

xmatthias:
Don't necessarily see this as an issue.

it's intended as a helper - but i don't think you'll be able to make it work in 100% of the cases - or make it never change any other line - at least not with the current approach.
it's a small subset of strategies that make extensive use of if/elif. 
It might be an inconvenience - but also something that's quite quickly fixed, too.

hippocritical:
the aim for this weekend is that I can provide you a workable version that could be tested.
Found 1-2 things that still needs patching, then others could test out the version too.
I wouldn't mind to have it inside a freqtrade branch itself instead of mine, just did it like that since this was the path of least resistance.

hippocritical:
if there are things missing that still don't get replaced properly I d need to get directions which.
if you need me to do something to move this into a freqtrade branch then just say so, first PR so I might have done things wrong.

xmatthias:
It's now clearly working a lot better :100: 

---

one thing it's still doing is messing with comment indentation if the next line is dedented.
If you don't notice this right away - and wanna come back to this later, it can sure add some complexity / head scratching as to why this is now broken.
This might be something ast_comments can look at (?) - as the comment was indented - so i'd expect indentation to remain identical... 

image

It's also moving comments out of statements - which (especially for multi-line dataframe assignments) is not ideal.
I suspect there won't be a real solution to this though - as the comment is not part of the statement.

image

Inline-comments are moved out of the line, and to a separate line.
again, nothing too surprising - and i'm not sure if it's problematic.

image

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -0,0 +1,55 @@
...
+    :return: None
+    """
+
+    config = setup_utils_configuration(args, RunMode.UTIL_NO_EXCHANGE)
+
+    strategy_objs = StrategyResolver.search_all_objects(
+        config, enum_failed=False, recursive=config.get('recursive_strategy_search', False))
+
+    filtered_strategy_objs = []
+    if hasattr(args, "strategy_list"):
```
xmatthias: there's something phishy going on with this

```
$ freqtrade strategy-updater --strategy-list TurtleStrategyX3p
2023-02-21 07:03:02,023 - freqtrade - INFO - freqtrade 2023.2.dev-08ca0f7c0
2023-02-21 07:03:02,023 - freqtrade.loggers - INFO - Verbosity set to 0
2023-02-21 07:03:02,024 - freqtrade.configuration.configuration - INFO - Using user-data directory: /home/xmatt/devel/cryptos/freqtrade/user_data ...
2023-02-21 07:03:02,024 - freqtrade.configuration.configuration - INFO - Using data directory: /home/xmatt/devel/cryptos/freqtrade/user_data/data ...
2023-02-21 07:03:02,024 - freqtrade.configuration.configuration - INFO - Using strategy list of 1 strategies
2023-02-21 07:03:02,024 - freqtrade.configuration.config_validation - INFO - Validating configuration ...
Conversion of sample_strategy.py started.
Conversion of sample_strategy.py took 0.0 seconds.
Conversion of TurtleStrategyX3p.py started.
Conversion of TurtleStrategyX3p.py took 0.0 seconds.
Conversion of StrategyB1.py started.
Conversion of StrategyB1.py took 0.0 seconds.
Conversion of xxyy.py started.
Conversion of xxyy.py took 0.0 seconds.
Conversion of xxyy1.py started.
Conversion of xxyy1.py took 0.0 seconds.
Conversion of StrategyB2.py started.
Conversion of StrategyB2.py took 0.0 seconds.

```

i've clearly passed in `--strategy-list` - but i got all strategies updated.

requirements.txt
```diff
@@ -57,3 +57,5 @@ schedule==1.1.0
 #WS Messages
 websockets==10.4
 janus==1.0.0
+
+astor~=0.8.1
```
xmatthias: `ast_comments` is missing - causing tests to fail.

hippocritical:
since we can't use astor anymore (which is a wrapper of ast, as ast_comments is) and it has the pretty code option it will be ugly at first now.
We would have to do your own beautification for code then.

xmatthias:
in the current version, i didn't find the code to be too ugly if i'm honest - most lines were not touched at all, or very minimally edited.

my comment on the "astor" line was more about "ast_comments is missing" - not "don't use astor". 
If astor is no longer used it for sure should be removed - but that wasn't my intend with the comment posted.

hippocritical:
yes, on it :) just an explanation why things would not be possible with ast_comments but were with astor.

Doesnt hurt the code itself it will still be working.

Currently working on the hasattr(strategy_list) thing and it apparently only returns true if the thing checked is a string. If it is a list is always returns false no matter if it s there or not.
So changed it to "if strategylist in args" and it works now ... will place a commit likely today.

freqtrade/strategy/strategyupdater.py
```diff
@@ -0,0 +1,253 @@
...
+    def visit_FunctionDef(self, node):
+        node.name = self.check_dict(StrategyUpdater.function_mapping, node.name)
+        self.generic_visit(node)
+        return node
+
+    def visit_Attribute(self, node):
+        if (
+                isinstance(node.value, ast_comments.Name)
+                and node.value.id == 'trades'
+                and node.attr == 'nr_of_successful_buys'
```
xmatthias: Am i reading this right that this is explicitly updating `trades.nr_successful_buys` to `trades.nr_successful_entries`?

I think the more common occurance for this will be `trade.nr_successful_buys` (singular - as is the name for the `trade` parameter) - which would currently fail.

freqtrade/strategy/strategyupdater.py
```diff
@@ -0,0 +1,253 @@
...
+                    ):
+                        child.value = ast_comments.parse('3').body[0].value
+        self.generic_visit(node)
+        return node
+
+    def visit_Subscript(self, node):
+        if isinstance(node.slice, ast_comments.Constant):
+            if node.slice.value in StrategyUpdater.rename_dict:
+                # Replace the slice attributes with the values from rename_dict
+                node.slice.value = StrategyUpdater.rename_dict[node.slice.value]
```
xmatthias, line 221: What scenario would this cover?

I've noticed that tests don't hit it - which stuns me, as i think there's tests for (almost) everything now.

xmatthias (approved):

PR #7970, opened 2023-01-03
Enhanced delete_lock() function

HichamBerbache:
added error handling

### DIFF
--- a/scripts/rest_client.py
+++ b/scripts/rest_client.py
@@ -124,7 +124,11 @@ class FtRestClient():
         :param lock_id: ID for the lock to delete
         :return: json object
         """
-        return self._delete(f"locks/{lock_id}")
+        if not self.locks().get(lock_id):
+            logger.warning("Lock id not found!")
+            return None
+        else:
+            return self._delete(f"locks/{lock_id}")
 
     def daily(self, days=None):
         """Return the profits for each day, and amount of trades.


scripts/rest_client.py
```diff
@@ -124,7 +124,11 @@ def delete_lock(self, lock_id):
         :param lock_id: ID for the lock to delete
         :return: json object
         """
-        return self._delete(f"locks/{lock_id}")
+        if not self.locks().get(lock_id):
+            logger.warning("Lock id not found!")
+            return None
+        else:
+            return self._delete(f"locks/{lock_id}")
```
xmatthias, line 131: I'm not entirely sure about this.

i think i'd prefer to have the script simply fail (which it does now - or eventually catch the 404 error that occurs if we attempt to delete a lock that doesn't exist - rather than first querying ALL locks (which could be 1000nds) just to find out if we actually can delete a lock or not.

it's also converting this command to a "2 request" command - which doesn't really make much sense.

xmatthias:
Based on the current output (`{"lock_count": 0, "locks": []}`) - i don't think we'll be changing how the script works.

The script is meant as a quick way to control the bot - doing 1 request per call (at max 2 - should we decide to move away from basic auth).

if you intend to do more complex workflows, the client can be imported from the script, and single methods can then be called as necessary.

PR #7971, opened 2023-01-03
Update FreqaiExampleStrategy.py

paranoidandy:
Change can_short to True to enable shorting

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/templates/FreqaiExampleStrategy.py
+++ b/freqtrade/templates/FreqaiExampleStrategy.py
@@ -39,7 +39,7 @@ class FreqaiExampleStrategy(IStrategy):
     use_exit_signal = True
     # this is the maximum period fed to talib (timeframe independent)
     startup_candle_count: int = 40
-    can_short = False
+    can_short = True
 
     std_dev_multiplier_buy = CategoricalParameter(
         [0.75, 1, 1.25, 1.5, 1.75], default=1.25, space="buy", optimize=True)


freqtrade/templates/FreqaiExampleStrategy.py
```diff
@@ -39,7 +39,7 @@ class FreqaiExampleStrategy(IStrategy):
     use_exit_signal = True
     # this is the maximum period fed to talib (timeframe independent)
     startup_candle_count: int = 40
-    can_short = False
+    can_short = True
```
xmatthias, line 42: I think the idea to have this as "false" was to allow the default sample strategy to also work on spot markets (freqtrade will not allow you to load a short strategy in spot markets).

As such - i'm not entirely sure if we should really be changing this to True - as it'll automatically exclude all spot markets.

(ci failures on macOS can be ignored - that's because the branch (as well as the target branch) is not uptodate with develop, which includes a fix for macOS ci).

robcaulk: I would vote for taking the change and setting it to True. This strategy is not designed for one side of the market, it is designed for both. 

Many people are confused as to why the strategy may have strange performance - this is likely one of (many) causes.

xmatthias: In principle I'm fine with that - just pointing out _why_ we set this to False in the first place :) 
requiring long/short support will however be problematic for quite a lot of people (US, europe) - as bigger exchanges started to lock us out from their futures markets.

it's anyway pointed at a non-develop branch but wagner's  - so will probably experience another review together with that branch.

wagnercosta: I agree with you guys. I vote for setting it to True, but I understand Matthias concerns and I think we can easily add a spot strategy freqAI example (or add a comment inside the current strategy giving some orientation).

xmatthias: works for me :+1:

PR #7972, opened 2023-01-03
Fix Backtesting  Analysis Column Wrong

Undertoned:
## Summary

I find wrong column when i analyze my trade data , so just fix it

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/data/entryexitanalysis.py
+++ b/freqtrade/data/entryexitanalysis.py
@@ -52,7 +52,7 @@ def _process_candles_and_indicators(pairlist, strategy_name, trades, signal_cand
     return analysed_trades_dict
 
 
-def _analyze_candles_and_indicators(pair, trades:pd.DataFrame, signal_candles:pd.DataFrame):
+def _analyze_candles_and_indicators(pair, trades: pd.DataFrame, signal_candles: pd.DataFrame):
     buyf = signal_candles
 
     if len(buyf) > 0:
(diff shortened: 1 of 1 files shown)


froggleston:
I'm not sure what was "wrong" here? I understand adding the function argument typing, but what is the purpose of rearranging the column order?

freqtrade/data/entryexitanalysis.py
```diff
@@ -120,7 +120,7 @@ def _do_group_table_output(bigdf, glist):
 
         else:
             agg_mask = {'profit_abs': ['count', 'sum', 'median', 'mean'],
-                        'profit_ratio': ['sum', 'median', 'mean']}
+                        'profit_ratio': ['median', 'mean', 'sum']}
             agg_cols = ['num_buys', 'profit_abs_sum', 'profit_abs_median',
                         'profit_abs_mean', 'median_profit_pct', 'mean_profit_pct',
                         'total_profit_pct']
```
xmatthias, line 126: @froggleston 
i was doubting the same at first - but:
if you compare the changed fields (`agg_mask`) with `agg_cols` - you'll notice that the sequence is wrong.
(total_profit_pct != profit_ratio.mean) ...

froggleston: Oh great spot @Undertoned ! LGTM.

xmatthias (approved):
great spot! :+1:

PR #7975, opened 2023-01-04
handle data gaps between FreqAI and DP better

robcaulk:
FreqAI was naive in data gap handling by simply skipping data updates if there was no matching date in the dataprovider and the historical FreqAI data. 

This PR ensures that data will always be appended, even if there is no matching date in the historical FreqAI data and the data provider. It logs a warning to the user so that they are aware that a gap in historical data may exist. However, this gap should be isolated to a single incident (if it ever occurs). 

In plain words, if there is no matching date in the historical freqai data and the dataprovider, then it will check to see if this is due to a missing candle or if it is because the bot has missed updates for a full dataprovider length of candles (e.g. 1000 for binance). In the former case, it appends the most recent candle from the dataprovider. In the latter case, it will append the entire dataprovider dataframe.

### DIFF
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -645,12 +645,20 @@ class FreqaiDataDrawer:
                             + 1
                         )
                     except IndexError:
+                        index = -1
+                        if history_data[pair][tf].iloc[-1]['date'] < df_dp['date'].iloc[0]:
+                            index = 0
+                        else:
+                            index = -1
                         logger.warning(
-                            f"Unable to update pair history for {pair}. "
-                            "If this does not resolve itself after 1 additional candle, "
-                            "please report the error to #freqai discord channel"
+                            f"No common dates in historical data and dataprovider for {pair}. "
+                            f"Appending dataprovider to historical data (full? {not bool(index)})"
+                            "but please be aware that there is likely a gap in the historical "
+                            "data.\n"
+                            f"Historical data ends at {history_data[pair][tf].iloc[-1]['date']} "
+                            f"while dataprovider starts at {df_dp['date'].iloc[0]} and"
+                            f"ends at {df_dp['date'].iloc[0]}."
                         )
-                        return
 
                     history_data[pair][tf] = pd.concat(
                         [
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -661,7 +661,7 @@ class FreqaiDataDrawer:
                                 f"ends at {df_dp['date'].iloc[0]}."
                             )
 
-                    hist_df = pd.concat(
+                    history_data[pair][tf] = pd.concat(
                         [
                             hist_df,
                             df_dp.iloc[index:],
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -627,12 +627,12 @@ class FreqaiDataDrawer:
 
             for pair in dk.all_pairs:
                 for tf in feat_params.get("include_timeframes"):
-
+                    hist_df = history_data[pair][tf]
                     # check if newest candle is already appended
                     df_dp = strategy.dp.get_pair_dataframe(pair, tf)
                     if len(df_dp.index) == 0:
                         continue
-                    if str(history_data[pair][tf].iloc[-1]["date"]) == str(
+                    if str(hist_df.iloc[-1]["date"]) == str(
                         df_dp.iloc[-1:]["date"].iloc[-1]
                     ):
                         continue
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/data_drawer.py
```diff
@@ -645,12 +645,19 @@ def update_historic_data(self, strategy: IStrategy, dk: FreqaiDataKitchen) -> No
                             + 1
                         )
                     except IndexError:
+                        if history_data[pair][tf].iloc[-1]['date'] < df_dp['date'].iloc[0]:
+                            index = 0
+                        else:
+                            index = -1
```
xmatthias, line 653: Assuming you have this case (`index = -1`) - it'd mean that the dataprovider data is older than the available data, right? (we got a key error - so it's not identical).

Why are we then still appending the very last candle?

wouldn't that result in the following date series (1h candles ...)

```
2022-01-02 00:00
2022-01-02 01:00
2022-01-02 02:00  <-- up to here, we get the data from "history_data"
2022-01-01 04:00  <-- this is appended from `df_dp` - as index is `-1` - the very last candle will be appended.
```

I think this case needs different handling (continue if index=-1) - to avoid having a timeseries in the wrong sequence.

xmatthias (commented):
> In plain words, if there is no matching date in the historical freqai data and the dataprovider, then it will check to see if this is due to a missing candle or if it is because the bot has missed updates for a full dataprovider length of candles (e.g. 1000 for binance). In the former case, it appends the most recent candle from the dataprovider. In the latter case, it will append the entire dataprovider dataframe.

Wouldn't you want to either exclude that pair - or restart the bot (so the data is refreshed) if you're missing 1000+ candles?
that's definitely a sign of a "sick" bot  ...

freqtrade/freqai/data_drawer.py
```diff
@@ -645,12 +645,19 @@ def update_historic_data(self, strategy: IStrategy, dk: FreqaiDataKitchen) -> No
                             + 1
                         )
                     except IndexError:
+                        if history_data[pair][tf].iloc[-1]['date'] < df_dp['date'].iloc[0]:
```
xmatthias: aside from this PR, i noticed that we use `history_data[pair][tf]` about 5 times in this for-loop alone.

While dict lookups are not "that" expensive - they're also not free - and i think it might improve readability if we can change that to a "single variable", instead of a nested dict lookup.

freqtrade/freqai/data_drawer.py
xmatthias: @robcaulk you missed this? ;)

freqtrade/freqai/data_drawer.py
robcaulk: Good point, changed.

freqtrade/freqai/data_drawer.py
robcaulk: > Assuming you have this case (index = -1) - it'd mean that the dataprovider data is older than the available data, right? (we got a key error - so it's not identical).

No, why would the data provider have older data than the `historic_data`? The historic data is constructed from the data provider each time the data provider comes in. Further, the `historic_data` is constructed newly every time the bot starts. A point can only be added to historic_data if it was done so from the dataprovider. This means that the second case of `index=-1` means that the pair just doesnt have that candle for some reason - (target case being it was offline for say 300 OKX candles) the best way to get back on track is to start adding dataprovider candles. There would be a gap in the data, and we log this quite concisely for the user to see, but there would be no case where the data provider is adding an older candle to the historic_data. We can add a check for it if youwant, but I do not see how it would occur. 

The goal here is resilience and getting the bot back on track in case it was disconnected from the internet for a dataprovider df length period of time. I dont see the issue here. 

> Why are we then still appending the very last candle?

To get the bot back on track.

> wouldn't that result in the following date series (1h candles ...)

I dont see any possibility of this occurring. 

> if you're missing 1000+ candles? that's definitely a sign of a "sick" bot ...

Sure, we log the message clearly for the user to understand. The goal here is resilience and getting the bot back on track in case it was disconnected from the internet for a dataprovider df length period of time.

freqtrade/freqai/data_drawer.py
xmatthias: @robcaulk forgotten to push? or do we aim to change this at a later point?

freqtrade/freqai/data_drawer.py
xmatthias: > (target case being it was offline for say 300 OKX candles)

This would be the first case though (`index=0`). Where historic data ends on 2023-01-23 - and dp data starts on 2023-02-01 (exaggerated example, obviously).

>>if you're missing 1000+ candles? that's definitely a sign of a "sick" bot ...
> Sure, we log the message clearly for the user to understand.

we do however continue operations for that pair - so the assumption is that a user is constantly reading every log message, and will eventually (in a timely manner) handle warnings in the log.
Neihter freqAI nor freqtrade will care if it's trained on data with a gap of 1000 candles ... but will the result be valid?

> > wouldn't that result in the following date series (1h candles ...)
> 
> I dont see any possibility of this occurring.

this WILL happen if we ever end in the "else" case of that condition.
It's not likely to happen (If we assume it can't happen - why is there an else case with different handling??)
Should it happen, then current behavior is wrong - potentially resulting in the above shown series.

i'm worried about the "else" case - which granted, is unlikely to happen - but will mean that historic data goes from 2023-01-01 to 2023-02-01 - but dataprovider data starts before 2023-01-01 - and potentially ends before that, too (we don't check this).
Yet we're using the very last dataprovider candle (without checking if it's actually after the historic data).

If we assume "won't ever happen" - then we should raise an exception which causes the bot to restart (this is undefined / completely wrong data - something must've gone wrong - the only safe approach is a restart - causing a new redownload of data - combined with fresh dataprovider data).

Alternatively, the else branch should void existing historic data - and start from scratch with what dataprovider offered.

> The goal here is resilience and getting the bot back on track in case it was disconnected from the internet for a dataprovider df length period of time.

That's a good thing obviously - but we still also need to protect from unlikely cases.

robcaulk: Ok, I am going with raising an `OperationalException` to force the user to reload a "sick bot." 

We agree that this should happen so infrequently (maybe never) that it shouldnt be an issue.

freqtrade/freqai/data_drawer.py
robcaulk: Not only did I forget to push, I somehow forgot to save it as well. Anyways should be in latest commit.

freqtrade/freqai/data_drawer.py
xmatthias: If you carefully read above - i've raised questions about the `if` side - but my concern was always more about the else block. 

i think the `OperationalException` is now placed in the wrong block (if, not else).
it should be either in both (no condition?) - or only in the else case (depending on what we want to accomplish)

in my understanding: the "if" block means:
* historic data stops before the dataprovider data starts -> data will have gaps

Failing here is fine :+1: 

---

the else block on the other hand:
* historic data stops **after** the dataprovider data ends (otherwise we'd not be in an index error)

this 2nd case is **very** problematic (and is most likely a theoretical case). 
I'm not even sure how we can get to the 2nd case (most likely, something funky with the clock happened) - but i know current behavior "use the last new candle and append that" is clearly wrong.

robcaulk: > the else block on the other hand:

    historic data stops after the dataprovider data ends (otherwise we'd not be in an index error)

Not necessarily, all we are doing is looking for a match of the *last* historical data candle with *any* candle in the dataprovider. If it is missing, it means that the dataprovider is missing that candle which could happen for a variety of reasons. However we have checked that the *last* historical data candle *is* in the same timerange as the dataprovider (because it is after the first dataprovider candle). Since we don't have a match, we simply want to get back on track and take the latest dataprovider candle and append it. 

In other words, the rarity of the "else" event (maybe a downtime from the exchange or something), means it is more important to keep the bot up instead of forcing a reload or something more drastic.

xmatthias: I've been missreading this as "we merge the whole dataframe and take the last date" - so i've been overlooking a `.iloc[-1]` in the try/catch block.

After a short discord discussion we've justified that now :laughing: 

Sorry for the confusion ...

xmatthias (approved):
LGTM :+1:

PR #7976, opened 2023-01-04
Hyperopt Max open trades

### LINKED ISSUE #5963: Add ability to hyperopt max_open_trades
## Describe your environment
N/A

## Describe the enhancement
Currently it seems the max_open_trades can have significant impact on profit. The only way I have been able to find a good number is to run several backtests one at a time and changing the setting and then comparing the results.
I'm wondering about the likelihood of adding this as a new hyperopt space. I realize HO is all strategy specific, and max_open_trades is bot specific, so this might be a lot of work.

adarkforce:
## Summary

This pull request is an attempt to add the ability to hyperopt `max_open_trades` and to set `max_open_trades` from the strategy.

Solve the issue: #5963

## Quick changelog

- `max_open_trades` can be set in the strategy class
- added the `trades` space to the hyperopt to optimize `max_open_trades`
- added an ovverridable `trades_space` function to customize the range of `max_open_trades` to optimize

## What's new?

- The hyperopt class has a new space (`trades_space`) used to evaluate the optimal `max_open_trades` parameter.
- Hyperopting will export save `max_open_trades` along with the other parameters.
- Loading a configuration file without `max_open_trades` will fallback to the value set in the strategy (-1 if not set).
- Setting `max_open_trades` = -1 in the strategy or in the config means that `max_open_trades` is unlimited.
- The config schema for BACKTESTING does not require `max_open_trades` anymore
- `use_max_market_positions` is ignored in hyperopt when optimizing `max_open_trades`
- Unlimited `max_open_trades` is ignored during hyperopting if `stake_amount` is set to unlimited


freqtrade/optimize/backtesting.py
```diff
@@ -920,7 +920,7 @@ def handle_left_open(self, open_trades: Dict[str, List[LocalTrade]],
                 trade.close(exit_row[OPEN_IDX], show_msg=False)
                 LocalTrade.close_bt_trade(trade)
 
-    def trade_slot_available(self, max_open_trades: int, open_trade_count: int) -> bool:
+    def trade_slot_available(self, max_open_trades: int | float, open_trade_count: int) -> bool:
```
xmatthias: assuming this change is done to fix a mypy error ... (and wherever this type changed elsewhere ...)

whatever is triggering the fact that `max_open_trades` can be a float  - it's wrong and needs to be fixed "there".
you can't have 2.5 trades open - it's either 2 - or 3 - therefore must be an integer.

aside from that - `int | float` is only valid in python 3.10 and up - however we also support python 3.8 - so `Union[int, float]` would be correct syntax (but is still wrong - as this value shouldn't be a float).

adarkforce: I wanted to consider the `max_open_trades` type in the strategy as `int` and `float('inf')`, since when the config is parsed the `-1` is converted to `float('inf')`.

But if I understand well i should allow only for integer values for `max_open_trades` in the strategy (so that -1 == infinity), and so loading a strategy with no value for `max_open_trades` in the config and with `max_open_trades = -1` set in the strategy should result in a config with `max_open_trades = float('inf')`, and vice versa when loading strategy with a config that has max_open_trades == float('inf') I should set -1 in the strategy?

xmatthias: the problem with supporting INF is that you can only support that if the stake-amount is fixed (it's not unlimited).

this is checked during regular startup - but (probably) not while hyperopting - so you could go from "100" to "inf" - where 100 is still valid, but inf isn't.
It's not that different from -1 though - as it's the meaning that's invalid.

freqtrade/commands/cli_options.py
```diff
@@ -251,7 +251,8 @@ def __init__(self, *args, **kwargs):
     "spaces": Arg(
         '--spaces',
         help='Specify which parameters to hyperopt. Space-separated list.',
-        choices=['all', 'buy', 'sell', 'roi', 'stoploss', 'trailing', 'protection', 'default'],
+        choices=['all', 'buy', 'sell', 'roi', 'stoploss',
+                 'trailing', 'protection', 'default', 'trades'],
```
xmatthias: i'm torn on naming here (default should go last, but that's a small nitpick).

"trades" as space name seems fine at first glance - but the more i think about it makes me feel it's not the right word.

all spaces are optimizing trades to some degree (when to enter, when to exit, when to not enter (protections) ...) - so it can be misunderstood quite easily (and it had me confused while looking through the rest of the changes, too).

We're optimizing trade count (or trade slots, or max_open_trades) - so i find trades almost too generic. 

(before acting/changing - let me know your thoughts on this!)

freqtrade/optimize/hyperopt.py
```diff
@@ -328,6 +341,9 @@ def generate_optimizer(self, raw_params: List[Any]) -> Dict[str, Any]:
             self.backtesting.strategy.trailing_only_offset_is_reached = \
                 d['trailing_only_offset_is_reached']
 
+        if HyperoptTools.has_space(self.config, 'trades'):
+            self.max_open_trades = params_dict['max_open_trades']
```
xmatthias: are you sure this works throughout the bot in all cases?

the strategy is not reloaded on every hyperopt loop - so the sync between parameter and config will not happen implicitly (and is also not necessary for all other attributes).

`config['max_open_trades']` is used throughout the bot - for example in `wallets.py` to calculate dynamic stakes.
So while this will hyperopt max_open_trades - the stake-amount will remain identical 

example with a 1000 USDT wallet, unlimited stake, and an initial value of 3 (from the config).
with that we get 1000/3 (roughly 333$)$ trades.
If the value is changed to 1 - we'd expect the stake size to be 1000$ (all in on every trade), or a 100$ stake if the trade-amount is bumped to 10.

In my tests, the first stake was always 329.73 - independent of the actually selected trade count - so if i'd start with a 1 max_open_trades value - i'd never get different results - as the stake-amount won't change - implicitly also maxing out all trade slots..

freqtrade/optimize/hyperopt.py
```diff
@@ -118,11 +119,14 @@ def __init__(self, config: Config) -> None:
...
         # Use max_open_trades for hyperopt as well, except --disable-max-market-positions is set
         if self.config.get('use_max_market_positions', True):
-            self.max_open_trades = self.config['max_open_trades']
+            self.max_open_trades = self.config['max_open_trades'] \
+                if self.config['max_open_trades'] != float('inf') else -1
         else:
             logger.debug('Ignoring max_open_trades (--disable-max-market-positions was used) ...')
             self.max_open_trades = 0
 
+        print("Strategy max open trades", self.max_open_trades)
```
xmatthias: don't think we should print here ... seems like a debug remain?

freqtrade/optimize/hyperopt.py
adarkforce: Sorry for replying late. Yes, I've realized that `max_open_trades` is accessed throughout the code only by reading at the config. I need to make sure that for backtesting the `max_open_trades` is only the one passed to the `backtest` function, and there's no path leading to read it from the config.
If it is not the case (as it seems) I will try to see what will be a more elegant solution of the two:
1. Passing `max_open_trades` to all the functions needing it called from the `backtest` function
2. Provide a clone of the config with a different `max_open_trades` for each backtesting epoch.

What do you think about it? 
I will also try to cover those cases with some tests. 
Thanks for the help. :D

freqtrade/optimize/hyperopt.py
adarkforce: Yes, I'm removing it :D

freqtrade/optimize/hyperopt.py
xmatthias: > Sorry for replying late

don't think about it that way - that's ... just wrong :)  2 days is NOT late.
it's not our job - so having other priorities in life is perfectly fine. I can't guarantee a review within a day (or whatever), either.

---

i think option 2 will be preferable, at least short-term, otherwise you end up passing `strategy.max_open_trades` around to tons of places.
You'll struggle with option 1 if you attempt to do this "partially" - as there's a ton of "call from a call" scenarios, so even if a function is not directly used in backtesting - it may still be called implicitly (Like the functions in wallets, for example).

in theory, it SHOULD work without leaking to the other epochs in that batch (without copying the config) as hyperopt pickles/unpickles the objects, so it's technically different object id's in each epoch - though i didn't test if the config might leak or not.
it should be easy to test within the strategy with `print(f'{self.max_open_trades}, {self.config["max_open_trades"]}')` or some variation of this ... with the expectation that they both yield the same result every time.

The problem with copying the config is that then you'll need to re-init all objects with the copied config (for example wallets) - which we'd want to avoid if possible.

freqtrade/commands/cli_options.py
adarkforce: Yeah it felt to disruptive to use a multi-word arg name, since they were all one word lol. 
While `trade_slots` or `trade_count` are good names (and better alternatives to `trades`, since they are more explicative and specific), they may be too distant from the parameter that we are optimizing: `max_open_trades`.
So `max_open_trades` or `open_trades` will be the best options in my opinion.
I would also consider `mot`, which also doesn't violate the one word rule :D

freqtrade/optimize/hyperopt.py
xmatthias: The new changes look to work as expected after a very quick test. :+1: 

i'll try to make some time to test and review properly soon.

freqtrade/optimize/hyperopt_interface.py
```diff
@@ -191,6 +191,16 @@ def trailing_space(self) -> List[Dimension]:
             Categorical([True, False], name='trailing_only_offset_is_reached'),
         ]
 
+    def trades_space(self) -> List[Dimension]:
```
xmatthias: I think at least the function here (it's the function that users can override) should get a better name, like `max_open_trades_space()`.

it'll make it clearer i'd think - and there's no requirement to have the function a "one word" name.

freqtrade/optimize/backtesting.py
```diff
@@ -1226,13 +1225,11 @@ def backtest_one_strategy(self, strat: IStrategy, data: Dict[str, DataFrame],
...
-        if self.config.get('use_max_market_positions', True):
-            # Must come from strategy config, as the strategy may modify this setting.
-            max_open_trades = self.strategy.config['max_open_trades']
-        else:
+        if not self.config.get('use_max_market_positions', True):
             logger.info(
                 'Ignoring max_open_trades (--disable-max-market-positions was used) ...')
-            max_open_trades = 0
+            self.strategy.max_open_trades = -1
+            self.config.update({'max_open_trades': float('inf')})
```
xmatthias: for my understanding mostly ...
why are we assigning different values to strategy and config?

while `strategy.max_open_trades` won't be used for the most part - won't this create a problem should that value be used? (it's clearly unexpected to have the 2 values diverge).

adarkforce: There is a little ambiguity with typing since the possible values for `max_open_trades` in the config are positive integers and float('inf') (which is a float), while the ones in the strategy are integers starting from -1 (so only `int`). 

I am going to check if it is better to have also the `max_open_trades` from the config being only `int`, with -1 meaning infinite, so that the two values will always match. What do you think about it? 

And also, yes, it is true, all of the reading of the `max_open_trades` parameter are done from the config, `max_open_trades` in the strategy is only used to sync with the config when loading the strategy, so I can avoid updating the strategy.max_open_trades while hyperopting / backtesting, but I thought it would have been better to have them synced (even if it is quite hardcoded).

xmatthias: i think setting it in both is better - if it's available, people will use it in the strategy for "whatever" (like `stake_amount / self.max_open_trades` (or whatever).
Results would differ if it's set "wrongly" during hyperopt.

it's very sad that there's no "int or inf" type - using -1 will have you do several `< max_open_trades OR -1` throughout ...
i didn't think about the "inf" case though - might be better to update to float (or a custom type we create with an explanation (it'll still be an alias for float though).

freqtrade/constants.py
```diff
@@ -681,3 +681,4 @@
 BidAsk = Literal['bid', 'ask']
 
 Config = Dict[str, Any]
+IntOrInf = float
```
xmatthias, line 684: Nice workaround :+1:

xmatthias (approved):
Great enhancement! :100: 

Also thanks for taking the time to do the `IntOrInf` workaround / fix!

LGTM :+1:

adarkforce:
> Great enhancement! 💯
> 
> Also thanks for taking the time to do the `IntOrInf` workaround / fix!
> 
> LGTM 👍

Thank YOU @xmatthias, for all the great feedbacks. 
Now the only thing left to do would be to modify the `trades` argument to the `space` option, what are your thoughts on it?

xmatthias:
> Now the only thing left to do would be to modify the trades argument to the space option, what are your thoughts on it? 

not sure i understand what you mean ?

freqtrade/commands/cli_options.py
adarkforce: Sorry @xmatthias , I didn't explain it well in the last message. I was talking about the `trades` space name, regarding this review. I haven't change it yet, so one would call `--space trades` to hyperopt `max_open_trades`.

xmatthias: i'm aware of this (was fully aware of it when we merged) - and lacking any idea on a better term - i don't think we should change it again at this point.

PR #7987, opened 2023-01-07
update calmar, sharpe, and sortino hyperopt losses to use latest formula

stash86:
## Summary

update calmar, sharpe, and sortino hyperopt losses to use latest formula

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/optimize/hyperopt_loss/hyperopt_loss_calmar.py
+++ b/freqtrade/optimize/hyperopt_loss/hyperopt_loss_calmar.py
@@ -11,7 +11,7 @@ from typing import Any, Dict
 from pandas import DataFrame
 
 from freqtrade.constants import Config
-from freqtrade.data.metrics import calculate_max_drawdown
+from freqtrade.data.metrics import calculate_calmar
 from freqtrade.optimize.hyperopt import IHyperOptLoss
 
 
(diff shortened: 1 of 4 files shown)


stash86:
seems like the sortino calculation is the one not passing the test. Once I commented sortinohyperoptloss from the test, the test pass.

The trouble part is the down_stdev part

xmatthias:
> The trouble part is the down_stdev part

That's correct. the sample results have 2 negative trades - both with an absolute loss of -0.2 (relative loss -0.1).
from a calculation perspective, this means losses have a STD of 0.

more than this test troubles me the fact that this is quite easily archivable - if we hit 5x static stoploss - then we have a STD of 0 (all losses are equal in size).

What's the expected sortino ratio for that? we're currently returning 100 - which is clearly wrong - as we can't calculate this.
i feel this makes sortino quite easy to break - as long as downside returns are spread thin - sortino will always be high, independently if these losses are small or large.

xmatthias (approved):
Improving the test case is valid just as well :laughing: 


Great, thanks for this! :+1:

stash86:
Lol yeah, but your point is valid. A high ratio doesn't always means good performance, users should be warned of this "trick"

PR #8004, opened 2023-01-11
Revert DataFrame serialization in Producer mode (>pandas 2.0)

wizrds:
## Summary

This PR reverts the use of `dataframe.to_dict` to `dataframe.to_json` in serialization. This fixes an issue caused by the existence of `@informative`. We originally changed it to bypass a memory leak that was caused by `dataframe.to_json` upstream in the pandas library, however that no longer seems to be an issue. Further testing will be done.


xmatthias:
> This fixes an issue caused by the existence of @informative

I'm not sure what that issue would be.
Behavior of `@informative` can be easily done without that decorator (it's really just a shortcut for "get / merge" an informative dataframe. 
The result is a regular dataframe without bells and whistles. I don't really see what problem that would cause - and if it does - then i'm not sure why to_json would fix that - as that's technically basically doing the same.

Without further information, this looks like a "shot in the dark" fix - without actually understanding the problem (not to say we shouldn't revert this workaround if we are certain pandas fixed their memory-leak - but i'm questioning the reason for doing so).

---

Have you got some references (for example in the pandas changelog) that mention the fix of this mem-leak that i overlooked?
The last few changelogs don't really mention this (this should be outlined as a bugfix).

one of the referenced issues  (in the now removed code) is still open - while the other was closed in 2019 (with people claiming it's still an issue in 2022) - so i don't really see evidence that it's been fixed by anything other than accidentally by some other change (which means it can be reintroduced any time).

wizrds:
> > This fixes an issue caused by the existence of @informative
> 
> I'm not sure what that issue would be. Behavior of `@informative` can be easily done without that decorator (it's really just a shortcut for "get / merge" an informative dataframe. The result is a regular dataframe without bells and whistles. I don't really see what problem that would cause - and if it does - then i'm not sure why to_json would fix that - as that's technically basically doing the same.
> 
> Without further information, this looks like a "shot in the dark" fix - without actually understanding the problem (not to say we shouldn't revert this workaround if we are certain pandas fixed their memory-leak - but i'm questioning the reason for doing so).
> 
> Have you got some references (for example in the pandas changelog) that mention the fix of this mem-leak that i overlooked? The last few changelogs don't really mention this (this should be outlined as a bugfix).
> 
> one of the referenced issues (in the now removed code) is still open - while the other was closed in 2019 (with people claiming it's still an issue in 2022) - so i don't really see evidence that it's been fixed by anything other than accidentally by some other change (which means it can be reintroduced any time).

Unfortunately this is a bit of a shot in the dark type of fix, I'm still unable to figure out why this is happening. I've reviewed the informative code quite a bit and there honestly shouldn't be anything funky changing with the DataFrame to cause the serialization error. Like we have discussed before, my original thought was it had to do with a DataFrame being nested and the use of `dataframe.to_dict` didn't recursively change the DataFrames, though it seems unlikely to be the case as `@informative` doesn't change anything other than merging in columns. I created the PR early on to kind of allow other users to make use of it if they *absolutely* needed informative functions in their producers. 

I was unable to find any references to the memory leak being fixed as well, thus I planned on testing the memory usage over a long period of time. It's a bit harder to notice any leaks since the introduction of the single candle change that happened a while ago, as now we don't get large bursts of used memory like we did before. From the testing I've done so far though it seems like it isn't leaking anymore or much at all if it is. I will be reviewing everything extensively when I can before fully committing to this path.

xmatthias:
> It's a bit harder to notice any leaks since the introduction of the single candle change that happened a while ago, as now we don't get large bursts of used memory like we did before

just as an idea - maybe it can be made to leak by continuously reconnecting (which will force a full transmission again and again?

xmatthias:
from a "informative" perspective - #8008 will fix that.

As said above, that doesn't mean we shouldn't consider removing the fix if the root cause has been fixed  (but that should be the reason, not to fix a not well understood, different problem) - but that'll need extensive testing unfortunately.

xmatthias:
Appears not fixed with current pandas versions (https://github.com/pandas-dev/pandas/pull/50794).

It's however a merged PR - so assuming that plays out - we should be able to revert this fix after the next pandas upgrade.

wizrds:
> Appears not fixed with current pandas versions ([pandas-dev/pandas#50794](https://github.com/pandas-dev/pandas/pull/50794)).
> 
> It's however a merged PR - so assuming that plays out - we should be able to revert this fix after the next pandas upgrade.

Was this released? I can't find it in the change logs for the last 2 releases at all

xmatthias:
This could now be relevant again.

The linked issue / PR is closed, and i don't really see anything in the pandas issue tracker related to this.

xmatthias:
> Was this released? I can't find it in the change logs for the last 2 releases at all

i'd think so? the PR points at "milestone 2.0", and it's not been thrown out.

---

> Fixed a memory leak in [DataFrame.to_json()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html#pandas.DataFrame.to_json) and [Series.to_json()](https://pandas.pydata.org/docs/reference/api/pandas.Series.to_json.html#pandas.Series.to_json) when serializing datetimes and timedeltas ([GH40443](https://github.com/pandas-dev/pandas/issues/40443))

[the pandas 2.0 changelog](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html) does include this (search for 40443)

wizrds:
> > Was this released? I can't find it in the change logs for the last 2 releases at all
> 
> i'd think so? the PR points at "milestone 2.0", and it's not been thrown out.
> 
> > Fixed a memory leak in [DataFrame.to_json()](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html#pandas.DataFrame.to_json) and [Series.to_json()](https://pandas.pydata.org/docs/reference/api/pandas.Series.to_json.html#pandas.Series.to_json) when serializing datetimes and timedeltas ([GH40443](https://github.com/pandas-dev/pandas/issues/40443))
> 
> [the pandas 2.0 changelog](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html) does include this (search for 40443)

Pfff I was looking on bug fixes :roll_eyes:

xmatthias (approved):

PR #8007, opened 2023-01-11
fix "--version": needs to change working directory

TheJoeSchr:
before calling `git`. otherwise it would display git commit id from the
directory where you are calling `freqtrade` from instead of freqtrade's
current commit id

## Summary

    `--version:` needs to change working directory before calling `git`. otherwise it would display git commit id from the
    directory where you are calling `freqtrade` from instead of freqtrade's
    current commit id

## Reproduce
    - into a git repo other then freqtrade: `cd my-strategies`
    - (optional) `pip install -e ../freqtrade`
    - call `freqtrade --version`: note hash after "dev-"
    - check `git log -n 1` => will be above hash
    - double-check: go to your `freqtrade/` and check `git log -n 1` there, will be different

### DIFF
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -3,11 +3,13 @@ __version__ = '2023.1.dev'
 
 if 'dev' in __version__:
     try:
+        import os
         import subprocess
+        freqtrade_basedir = os.path.dirname(os.path.abspath(__file__))
 
         __version__ = __version__ + '-' + subprocess.check_output(
             ['git', 'log', '--format="%h"', '-n 1'],
-            stderr=subprocess.DEVNULL).decode("utf-8").rstrip().strip('"')
+            stderr=subprocess.DEVNULL, cwd=freqtrade_basedir).decode("utf-8").rstrip().strip('"')
 
     except Exception:  # pragma: no cover
         # git not available, ignore
--- a/freqtrade/__init__.py
+++ b/freqtrade/__init__.py
@@ -2,10 +2,10 @@
 __version__ = '2023.1.dev'
 
 if 'dev' in __version__:
+    from pathlib import Path
     try:
-        import os
         import subprocess
-        freqtrade_basedir = os.path.dirname(os.path.abspath(__file__))
+        freqtrade_basedir = Path(__file__).parent
 
         __version__ = __version__ + '-' + subprocess.check_output(
             ['git', 'log', '--format="%h"', '-n 1'],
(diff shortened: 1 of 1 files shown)


freqtrade/__init__.py
```diff
@@ -3,11 +3,13 @@
 
 if 'dev' in __version__:
     try:
+        import os
         import subprocess
+        freqtrade_basedir = os.path.dirname(os.path.abspath(__file__))
```
xmatthias: Please use [Pathlib](https://docs.python.org/3/library/pathlib.html) - it provides the same functionalities - but in a safer method (safe as in - works cross-platform including windows).

I consider the os module as legacy when it comes to path handling - it had too many bugs in the past when it came to windows.

xmatthias (changes requested):

freqtrade/__init__.py
TheJoeSchr: done

TheJoeSchr:
What do you think about adding this as well?
It would display the current branch name of the freqtrade repo as well. This is how I actually caught this bug, because names are easier to compare than hashes for humans

```python

        __version__ = __version__ + '-' + subprocess.check_output(
            ['git', 'symbolic-ref', '--short', 'HEAD'],
            stderr=subprocess.DEVNULL, cwd=freqtrade_basedir).decode("utf-8").rstrip().strip('"')
```

xmatthias:
> What do you think about adding this as well?

nah not really.
first - you'll not know how a branch is going to be spelled - so it'll violate Pep440 (no longer installable).
it's also potentially sensitive information (it's a private development branch - but i require support with something).

the hash is not for everyday users - but for us when helping people with issues. 
knowing the branch won't help (it won't matter if you're on develop, or created a new branch named "myNewCoolFeature" but made no change) - it's either a known commit (we know what's in and can support that) or it isn't (even if the branch is called develop - you can have custom commits on it with unknown content.

xmatthias (approved):
LGTM :+1:

TheJoeSchr:
> > What do you think about adding this as well?
> 
> nah not really. first - you'll not know how a branch is going to be spelled - so it'll violate Pep440 (no longer installable). it's also potentially sensitive information (it's a private development branch - but i require support with something).
> 
> the hash is not for everyday users - but for us when helping people with issues. knowing the branch won't help (it won't matter if you're on develop, or created a new branch named "myNewCoolFeature" but made no change) - it's either a known commit (we know what's in and can support that) or it isn't (even if the branch is called develop - you can have custom commits on it with unknown content.

I concur! :+1:

PR #8008, opened 2023-01-11
Fix websockets for dataframes with NaT entries

xmatthias:
## Summary

Fix serialization of dataframes with NaT in producer/consumer mode.


wizrds (approved):
LGTM :+1:

PR #8031, opened 2023-01-16
Add a new analysis group to output stats grouped by exit_tag

froggleston:
## Summary

In backtesting-analysis there was no option to view only exit tag performance. This PR adds analysis group 5 to add this functionality.

## Quick changelog

- Add analysis_group 5
- Update docs and tests

## What's new?

An additional table in the backtesting-analysis output that provides the exit-tag aggregated data.

image


xmatthias (approved):
LGTM 👍

PR #8033, opened 2023-01-18
update config-freqai-example to match latest binance futures pair syntax

stash86:
## Summary

update config-freqai-example to match latest binance futures pair syntax

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Great spot! :+1:

PR #8080, opened 2023-01-30
Fixes gym issue

leonardocustodio:
If we set setuptools to version 65.5.0 it solves the gym installation issue.
We can remove this when gym is upgraded.
https://github.com/freqtrade/freqtrade/issues/8078

PR #8083, opened 2023-01-31
Gym aid

### LINKED ISSUE #8078: Running setup.py install for gym did not run successfully.
Running setup.py install for gym ... error
  error: subprocess-exited-with-error

  

## environment

  * Operating system: ubuntu-22.04.1-live-server
  * Python Version: python3
  
## Describe the problem:
Hello ! 
I try to setup freqtrade on ProxmoxVM-ubuntu-22.04.1-live-server

get this error :

× Running setup.py install for gym did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      error in gym setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: legacy-install-failure

× Encountered error while trying to install package.
╰─> gym

note: This is an issue with the package mentioned above, not pip.
hint: See above for output from the failure.
Failed installing dependencies

leonardocustodio:
Refer to https://github.com/freqtrade/freqtrade/pull/8080

@xmatthias , do you think this way could be any useful for newcomers?

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -31,7 +31,7 @@ The easiest way to install and run Freqtrade is to clone the bot Github reposito
     The clock on the system running the bot must be accurate, synchronized to a NTP server frequently enough to avoid problems with communication to the exchanges.
 
 !!! Error: Running setup.py install for gym did not run successfully.
-    If you get an error related with gym we suggest you to downgrade it to version 65.5.0 you can do it with the following command:
+    If you get an error related with gym we suggest you to downgrade setuptools it to version 65.5.0 you can do it with the following command:
     ```bash
     pip install setuptools==65.5.0
     ```
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -30,6 +30,12 @@ The easiest way to install and run Freqtrade is to clone the bot Github reposito
 !!! Warning "Up-to-date clock"
     The clock on the system running the bot must be accurate, synchronized to a NTP server frequently enough to avoid problems with communication to the exchanges.
 
+!!! Error: Running setup.py install for gym did not run successfully.
+    If you get an error related with gym we suggest you to downgrade it to version 65.5.0 you can do it with the following command:
+    ```bash
+    pip install setuptools==65.5.0
+    ```
+
 ------
 
 ## Requirements
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -30,7 +30,7 @@ The easiest way to install and run Freqtrade is to clone the bot Github reposito
 !!! Warning "Up-to-date clock"
     The clock on the system running the bot must be accurate, synchronized to a NTP server frequently enough to avoid problems with communication to the exchanges.
 
-!!! Error: Running setup.py install for gym did not run successfully.
+!!! Error "Running setup.py install for gym did not run successfully."
     If you get an error related with gym we suggest you to downgrade setuptools it to version 65.5.0 you can do it with the following command:
     ```bash
     pip install setuptools==65.5.0
--- a/setup.sh
+++ b/setup.sh
@@ -49,7 +49,8 @@ function updateenv() {
     source .env/bin/activate
     SYS_ARCH=$(uname -m)
     echo "pip install in-progress. Please wait..."
-    ${PYTHON} -m pip install --upgrade pip
+    # Setuptools 65.5.0 is the last version that can install gym==0.21.0
+    ${PYTHON} -m pip install --upgrade pip wheel setuptools==65.5.0
     read -p "Do you want to install dependencies for dev [y/N]? "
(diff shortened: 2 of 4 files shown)


leonardocustodio:
Or maybe just make an option in the menu to "fix" that. Because it still needs the fixed version in the requirement (to work in setup.sh). I think the reason for that is that it runs in a fresh venv. Even if you have the correct version before if that is not present, it will not work. So only fixing in there is enough to make it work on setup.sh
But if you want to make the installation system-wide. You need to downgrade, so maybe just make an option for the ones who want to make the system-wide install? Or even another sh file temp_fix_gym.sh ?

setup.sh
```diff
@@ -1,6 +1,30 @@
...
+  stoolversion=${stoolversion##* }
+  stoolversion=$(echo $stoolversion | grep -Eo "^[0-9]+")
+
+  if [ $stoolversion -gt 65 ]; then
+    echo "You are using a new version of setuptools. This is not supported by gym at the moment."
+    read -p "Do you want to downgrade setuptools to a version that is supported by gym? [y/N] "
+
+    if [[ $REPLY =~ ^[Yy]$ ]]
+    then
+      ${PYTHON} -m pip install setuptools==65.5.0
```
xmatthias: If i'm honest, then my approach would've been to force this downgrade if freqai-rl (or dev) is selected (we do something similar with cython for raspberry installations).

It's not really optional - and we know for a fact that installation will fail if this requirement is not matched - so we're basically asking the user "would you like to have the installation fail - or not?" - without saying it out loud 😆 

It's not a global downgrade anyway, and will only impact the current virtual environment - so it's not problematic i'd think.

leonardocustodio: So that's a funny thing. In the fresh, newly created venv the version is actually `59.6.0` it doesn't even pass the check if that check is after `source .env/bin/activate`. Maybe we can only leave it in the freqai requirements for the setup.sh. It will still not work if trying to install system-wide (through `pip install -r ....`). It will only work in setup.sh.

leonardocustodio: Still again even if the version is `59.6.0` in venv if the version is not fixed in the requirement it fails to build gym.

leonardocustodio: Ok, can't get simpler than that, just let people know in the docs haha

docs/installation.md
```diff
@@ -30,6 +30,12 @@ The easiest way to install and run Freqtrade is to clone the bot Github reposito
 !!! Warning "Up-to-date clock"
     The clock on the system running the bot must be accurate, synchronized to a NTP server frequently enough to avoid problems with communication to the exchanges.
 
+!!! Error: Running setup.py install for gym did not run successfully.
```
xmatthias: Hah the docs check got this (not me)
the correct syntax is without a ":" 

```suggestion
!!! Error "Running setup.py install for gym did not run successfully."
```

setup.sh
xmatthias: that obviously works 😆 

i did a few tests (always on clear systems) - and found a working method for the setup.sh script.
combined with the documentation - that will have to suffice.

xmatthias (approved):
LGTM 👍

docs/installation.md
leonardocustodio: Nice catch by the ci

setup.sh
leonardocustodio: Looks great, thank you for getting to the bottom of this :)

mathlabelle:
hi, what is the final solution to get gym to install? i deleted everything and started fresh with a new ubuntu 22.04 install and the script installation method still gives me the same gym issue.  **

for newbies like me.

-reinstall ubuntu 22.04 
-run :
# update repository
sudo apt-get update

# install packages
sudo apt install -y python3-pip python3-venv python3-dev python3-pandas git curl

pip install setuptools==65.5.0 --no-cache-dir 

- carry on with the freqtrade installation guide using the developt branch

xmatthias:
it'll work on develop - but obviously not on the stable / released version - at least not at this moment.

Installing setuptools upfront won't work - as the `setup.sh` script creates a new virtual environment - activates that - and then uses whatever setuptools are in there. 
the updated setup.sh script does account for that and eventually downgrades setuptools.

mathlabelle:
true, my bad i did install the develop since i could  nt get the stable version to work.
BTW in no mean i am complaining , i am really gratefull for the chance to use freqtrade. 
i was simply trying to give a in plain english way to install the new version and have the freqai working for new python learners.
just making sure it didn't get received as a negative comment.
cheers

xmatthias:
No worries - the way i understood it was as question ...

PR #8088, opened 2023-02-01
Update README.md

Ezrahel:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,10 @@ first. If it hasn't been reported, please
 ensure you follow the template guide so that the team can assist you as
 quickly as possible.
 
+For every [issue](https://github.com/freqtrade/freqtrade/issues/new/choose) created, kindly follow up and mark satisfaction or reminder to close issue when equilibrium ground is reached.
+
+--Maintain github's community policy--
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -165,6 +165,10 @@ first. If it hasn't been reported, please
 ensure you follow the template guide so that the team can assist you as
 quickly as possible.
 
+For every [issue](https://github.com/freqtrade/freqtrade/issues/new/choose) created, kindly follow up and mark satisfaction or reminder to close issue when equilibrium ground is reached.
+
+--Maintain github's community policy--
```
xmatthias: reading this ... maybe there should be a link to that community policy if we assume people to follow it? 

searching for `Maintain github's community policy` on google yields some results - neither of which is "clearly" the answer we're looking for (all titles are worded differently ...).

Ezrahel: I found one that communicate the intent of my write up, you can check it out here>> https://docs.github.com/en/site-policy/github-terms/github-community-code-of-conduct

xmatthias: I found that - but if we point to it from the readme, it should be a clickable link 😉

xmatthias (approved):
LGTM 👍

PR #8101, opened 2023-02-02
[kucoin] manage kucoin numeric password passed as environment variabl…

### LINKED ISSUE #8100: Kucoin passord should be a string also if it is passed with an environment variable
## Describe your environment

  * Operating system: docker
  * Python Version: _____ (`python -V`)
  * Freqtrade Version: develop
  
## Describe the problem:

When you use docker version of freqtrade and you want to pass kucoin password with an environment variable, it is converted in number but validator says that it should be a string

### Steps to reproduce:

  1. prepare docker environment
  2. configure to use real kucoin account
  3. pass kucoin password as environment variable
  
### Observed Results:

Validator refuse to accept configuration because kucoin password should be a string and not a numeric value

lucaforni:
…e as a string

## Summary

Solve the issue: #8100 

## Quick changelog

- exchange password value passed as an environment variable is not converted as numeric value

### DIFF
--- a/freqtrade/configuration/environment_vars.py
+++ b/freqtrade/configuration/environment_vars.py
@@ -32,7 +32,7 @@ def flat_vars_to_nested_dict(env_dict: Dict[str, Any], prefix: str) -> Dict[str,
     :param prefix: Prefix to consider (usually FREQTRADE__)
     :return: Nested dict based on available and relevant variables.
     """
-    no_convert = ['CHAT_ID']
+    no_convert = ['CHAT_ID', 'PASSWORD']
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
LGTM 👍 

Small workflow optimization tip for next time (don't take this as critique - just as small time saver):
if it's such a small fix - **and** you submit the fix literally 1 minute after the issue - skip the issue and put the content of it into the PR. 

It might be different if you either don't know the fix / need us to look at it - or if the fix needs some discussion upfront - but here that's clearly not the case.

Anyway - grazie mille 💯

PR #8102, opened 2023-02-02
setup.sh: checks if git directory is dirty before bothering user with…

TheJoeSchr:
… potentially scary question

## Summary
As discussed on discord. Setup should not ask to reset if there are no changes, so unexperienced Users dont get scared off.

Also turns it to "keep changes", so experienced User can just smash yyyyy while waiting for the 10seconds to be over.

win-win?

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -190,16 +190,17 @@ function update() {
     updateenv
 }
 
-# Reset Develop or Stable branch
 function check_git_changes() {
     if [ -z "$(git status --porcelain)" ]; then
         echo "No changes in git directory"
-        return 0
+        return 1
     else
         echo "Changes in git directory"
-        return 1
+        return 0
     fi
 }
+
+# Reset Develop or Stable branch
 function reset() {
     echo_block "Resetting branch and virtual env"
 
--- a/setup.sh
+++ b/setup.sh
@@ -221,8 +221,6 @@ function reset() {
                     echo "- Hard resetting of 'stable' branch."
                     git reset --hard origin/stable
                 fi
-            else
-                echo "Reset ignored because you are not on 'stable' or 'develop'."
             fi
         fi
     else
--- a/setup.sh
+++ b/setup.sh
@@ -191,25 +191,37 @@ function update() {
 }
 
 # Reset Develop or Stable branch
+function check_git_changes() {
+    if [ -z "$(git status --porcelain)" ]; then
+        echo "No changes in git directory"
+        return 0
+    else
+        echo "Changes in git directory"
+        return 1
+    fi
+}
 function reset() {
     echo_block "Resetting branch and virtual env"
 
     if [ "1" == $(git branch -vv |grep -cE "\* develop|\* stable") ]
     then
+        if check_git_changes; then
+            read -p "Keep your local changes? (Otherwise will remove all changes you made!) [Y/n]? "
+            if [[ $REPLY =~ ^[Nn]$ ]]; then
 
-        read -p "Reset git branch? (This will remove all changes you made!) [y/N]? "
-        if [[ $REPLY =~ ^[Yy]$ ]]; then
+                git fetch -a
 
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -191,25 +191,37 @@ function update() {
...
-
-            if [ "1" == $(git branch -vv | grep -c "* develop") ]
-            then
-                echo "- Hard resetting of 'develop' branch."
-                git reset --hard origin/develop
-            elif [ "1" == $(git branch -vv | grep -c "* stable") ]
-            then
-                echo "- Hard resetting of 'stable' branch."
-                git reset --hard origin/stable
+        if check_git_changes; then
```
xmatthias, line 209: unfortunately,  this change ain't working .

this is bash - you can't assume that "1" is truthy, and 0 is falsy.
Running this script shows:

```
----------------------------
Resetting branch and virtual env
----------------------------
Changes in git directory
- Deleting your previous virtual env

```


It clearly detected "changes in git directory" - but it ignored the remaining "if" block - as the check is invalid.

xmatthias (changes requested):
I don't disagree with this - and the script can surely be improved further, too.

setup.sh
TheJoeSchr: damn you chatgpt :D

it should work, should according to SO as well: https://stackoverflow.com/questions/5431909/returning-a-boolean-from-a-bash-function

Not a bash hero, so if you got any ideas, feel free to pass them along

xmatthias: quite honestly, i'd have believed that as well by just looking at the code - but after testing - i saw it ain't working (i'm always skeptical about bash comparisons ...).

TheJoeSchr: `0 is true` and `1 is false` in bash, btw. 

because it's analogue to the exit value of the process. just remembered my unix classes

TheJoeSchr: fixed it now

freqtrade/exchange/exchange.py
```diff
@@ -1833,7 +1833,7 @@ async def _async_get_historic_ohlcv(self, pair: str, timeframe: str,
         )
         input_coroutines = [self._async_get_candle_history(
             pair, timeframe, candle_type, since) for since in
-            range(since_ms, until_ms or (arrow.utcnow().int_timestamp * 1000), one_call)]
+            range(since_ms if since_ms else 0, until_ms if until_ms else (arrow.utcnow().int_timestamp * 1000), one_call)]
```
xmatthias: i'd rather not have that check included, as it's not really necessary.

you should be able to quite easily remove this by running `git rebase -i origin/develop` - and selecting "drop" on the middle commit, followed by a `git push -f`.

setup.sh
xmatthias: Works now :+1:

freqtrade/exchange/exchange.py
TheJoeSchr: yeah no worries, thanks for writing it out, very thoughtful

setup.sh
```diff
@@ -190,26 +190,39 @@ function update() {
...
+                    echo "- Hard resetting of 'develop' branch."
+                    git reset --hard origin/develop
+                elif [ "1" == $(git branch -vv | grep -c "* stable") ]
+                then
+                    echo "- Hard resetting of 'stable' branch."
+                    git reset --hard origin/stable
+                fi
+            else
+                echo "Reset ignored because you are not on 'stable' or 'develop'."
             fi
```
xmatthias: I don't think we'll need this else branch. 
the message shown is missleading / wrong (i selected "Y" - but my branch was stable).

TheJoeSchr: good catch. removed it

xmatthias (approved):
Very nice 💯 
That script is getting more love in the last week alone than it got in the past 3-4 years 🤣 

LGTM 👍

PR #8109, opened 2023-02-04
Pass metadata dictionary to feature_engineering_* and set_freqai_targets()

robcaulk:
Users may have specific needs for fine controlling the feature engineering/targets based on the present pair/timeframe/period of the freqai automated feature engineering algorithm. 

This PR aims to provide that for them, in addition to documentation to help explain how it works. The goal is that it looks and feels similar to the existing freqtrade `metadata` dictionary.

### DIFF
--- a/docs/freqai-feature-engineering.md
+++ b/docs/freqai-feature-engineering.md
@@ -28,6 +28,10 @@ It is advisable to start from the template `feature_engineering_*` functions in 
 
         All features must be prepended with `%` to be recognized by FreqAI internals.
 
+        Access metadata such as the current pair/timeframe/period with:
+
+        `metadata["pair"]` `metadata["tf"]`  `metadata["period"]`
+
         :param df: strategy dataframe which will receive the features
         :param period: period of the indicator - usage example:
         dataframe["%-ema-period"] = ta.EMA(dataframe, timeperiod=period)
@@ -75,6 +79,10 @@ It is advisable to start from the template `feature_engineering_*` functions in 
         Features defined here will *not* be automatically duplicated on user defined
         `indicator_periods_candles`
 
+        Access metadata such as the current pair/timeframe with:
+
+        `metadata["pair"]` `metadata["tf"]`
+
         All features must be prepended with `%` to be recognized by FreqAI internals.
 
         :param df: strategy dataframe which will receive the features
@@ -98,6 +106,10 @@ It is advisable to start from the template `feature_engineering_*` functions in 
         This function is a good place for any feature that should not be auto-expanded upon
         (e.g. day of the week).
 
+        Access metadata such as the current pair with:
+
+        `metadata["pair"]`
+
         All features must be prepended with `%` to be recognized by FreqAI internals.
 
         :param df: strategy dataframe which will receive the features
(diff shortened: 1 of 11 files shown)


freqtrade/templates/FreqaiExampleStrategy.py
```diff
@@ -161,6 +173,10 @@ def set_freqai_targets(self, dataframe, **kwargs):
         Required function to set the targets for the model.
         All targets must be prepended with `&` to be recognized by the FreqAI internals.
 
+        Access metadata such as the current pair with:
```
xmatthias, line 181: I guess the "metadata" kwarg should be in the header definition for the functions (in the sample strategies, as well as documentation).

Right now we say "access via metadata" - but unless you implicitly know that metadata is a 2nd argument - you'll just assume it's "magically" there - which is not the case.

docs/freqai-feature-engineering.md
```diff
@@ -161,6 +177,19 @@ You can ask for each of the defined features to be included also for informative
 In total, the number of features the user of the presented example strat has created is: length of `include_timeframes` * no. features in `feature_engineering_expand_*()` * length of `include_corr_pairlist` * no. `include_shifted_candles` * length of `indicator_periods_candles`
  $= 3 * 3 * 3 * 2 * 2 = 108$.
 
+
+ ### Gain finer control over `feature_engineering_*` functions with `metadata`
+
+ All `feature_engineering_*` and `set_freqai_targets()` functions are passed a `metadata` dictionary which contains information about the `pair`, `tf` (timeframe), and `period` that FreqAI is automating for feature building. As such, a user can use `metadata` inside `feature_engineering_*` functions as criteria for blocking/reserving features for certain timeframes, periods, pairs etc.
+
+ ```py
+def feature_engineering_expand_all(self, dataframe, period, **kwargs):
```
wagnercosta, line 186: I think it's necessary to change the method signature to includes the `metadata`param or change the example to read from `kwargs`, right?

freqtrade/freqai/data_kitchen.py
```diff
@@ -1247,17 +1247,21 @@ def populate_features(self, dataframe: DataFrame, pair: str, strategy: IStrategy
         tfs: List[str] = self.freqai_config["feature_parameters"].get("include_timeframes")
 
         for tf in tfs:
+            metadata = {"pair": pair, "tf": tf}
             informative_df = self.get_pair_data_for_features(
                 pair, tf, strategy, corr_dataframes, base_dataframes, is_corr_pairs)
             informative_copy = informative_df.copy()
 
             for t in self.freqai_config["feature_parameters"]["indicator_periods_candles"]:
+                metadata["period"] = t
```
wagnercosta: The `period` is already sent as a `t` parameter. Do you think it's necessary to add in metadata?

wagnercosta (commented):
This PR will help a lot the users to customize FreqAI strategy features!

freqtrade/freqai/data_kitchen.py
robcaulk: Good point, I removed it to avoid redundancy.

freqtrade/templates/FreqaiExampleStrategy.py
robcaulk: Ok, I have added metadata to the docstrings as well as the example arguments.

wagnercosta (dismissed):
LGTM

xmatthias (approved):
LGTM :+1: 

I've slightly improved the typehints - as it'll allow users to fully benefit the editor's auto complete capability.
it's not 100% related to this PR (it's been missing from other fields, too) - but i've seen it here and it's quite related (same functions being touched ...).

PR #8147, opened 2023-02-10
Add pair to environment for access inside calculate_reward

robcaulk:
We have a user in the discord who wants to start using feature values to help make reward estimates. This PR should enable that for all users.

Now this can be included in the `calculate_reward()`

```python
            # you can use feature values from dataframe
            rsi_now = self.df["%-rsi-period-_shift-1_{self.pair}_{self.config['timeframe']}"].iloc[self._current_tick]

            # reward agent for entering trades
            if (action in (Actions.Long_enter.value, Actions.Short_enter.value)
                    and self._position == Positions.Neutral):
                if rsi_now < 40:
                    factor = 2
                else:
                    factor = 1
                return 25 * factor
```

@xmatthias I am doing a `per-file-ignores` because I disagree that the template `calculate_reward()` function in `BaseReinforcementLearningModel.py` is too complex. Instead, the goal is to show all the various ways a user could play with `calculate_reward()`, which means lots of if then statements (since we need to demonstrate a variety of methods to them). I am sure you will have a different approach that you prefer, go ahead and implement that if so.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,5 +49,3 @@ exclude =
     __pycache__,
     .eggs,
     user_data,
-per-file-ignores =
-    freqtrade/freqai/RL/BaseReinforcementLearningModel.py:C901
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,3 +49,5 @@ exclude =
     __pycache__,
     .eggs,
     user_data,
+per-file-ignores =
+    freqtrade/freqai/RL/BaseReinforcementLearningModel.py:C901
--- a/docs/freqai-reinforcement-learning.md
+++ b/docs/freqai-reinforcement-learning.md
@@ -177,7 +177,7 @@ As you begin to modify the strategy and the prediction model, you will quickly r
                 factor = 100
 
             # you can use feature values from dataframe
-            rsi_now = self.df[f"%-rsi-period-10_shift-1_{self.pair}_"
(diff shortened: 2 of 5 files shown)


setup.cfg
```diff
@@ -49,3 +49,5 @@ exclude =
     __pycache__,
     .eggs,
     user_data,
+per-file-ignores =
+    freqtrade/freqai/RL/BaseReinforcementLearningModel.py:C901
```
xmatthias: You can do excludes inline, in the file itself at the top of the file (or at the head of the function, if you'd prefer to exclude just one function).

doing it here is kinda odd - and will fail "at random" should we ever rename or move the file.

robcaulk: Oh, if we can do it per function then that solves it. I will ask chatGPT to google that for me and then I will implement it.

xmatthias: >  I will ask chatGPT to google that for me and then I will implement it. 

chatGPT can't really google
but if you search through the codebase for C901 you'll see that we have 2 instances of that already

https://github.com/freqtrade/freqtrade/blob/d47d8c135b0f3fcac430e1ad225c55d7465583a2/freqtrade/rpc/api_server/api_backtest.py#L29-L31

robcaulk: >chatGPT can't really google

oh you thought that was serious hah.

xmatthias: well - no 😆

xmatthias (approved):
LGTM :+1:

PR #8148, opened 2023-02-11
Add explicit warning that supported price types gonna differ

stash86:
## Summary

Add explicit warning that supported price types gonna differ on each exchanges

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
:100:

PR #8152, opened 2023-02-11
Typo on freqai docs

Shadyzpop:
## Summary

1. `a the` - there is an extra "a" before `the features`
2. `historic` - it should be "historical" to match the correct adjective form.


xmatthias (approved):
Great eye for detail! :+1:

PR #8171, opened 2023-02-13
Remove deepcopy protection, replace it with a warning message

raphaelstar:
## Summary

The `trade` parameter is copied with `deepcopy` to protect the user from unexpected behavior. This: 

- produces a large drawback in performance, especially on strategies with a lot of DCA orders.
- prevents advanced strategies that require mutating `trade`.

This PR removes the deep copy and replaces it with a warning message that is displayed during `dry_run` and `live` modes. The mechanism to check for mutation of `trade` is done by using the `sqlalchemy` engine by checking if `trade` has been flagged as dirty, indicating a mutation.

## Quick changelog

- remove deepcopy, add a warning message if trade have mutated.

## What's new?

The following benchmark shows a speedup of 150%.
<details>
  <summary>Config:</summary>

```json
{
  "max_open_trades": 1,
  "stake_currency": "USDT",
  "stake_amount": 100,
  "tradable_balance_ratio": 0.99,
  "timeframe": "5m",
  "dry_run": true,
  "dry_run_wallet": 1000000,
  "position_adjustment_enable": true,
  "cancel_open_orders_on_exit": false,
  "trading_mode": "futures",
  "margin_mode": "isolated",
  "unfilledtimeout": {
    "entry": 10,
    "exit": 10,
    "exit_timeout_count": 0,
    "unit": "minutes"
  },
  "entry_pricing": {
    "price_side": "same",
    "use_order_book": true,
    "order_book_top": 1,
    "price_last_balance": 0.0,
    "check_depth_of_market": {
      "enabled": false,
      "bids_to_ask_delta": 1
    }
  },
  "exit_pricing": {
    "price_side": "same",
    "use_order_book": true,
    "order_book_top": 1
  },
  "exchange": {
    "name": "okx",
    "key": "",
    "secret": "",
    "ccxt_config": {},
    "ccxt_async_config": {},
    "pair_whitelist": [
      "LTC/USDT:USDT"
    ],
    "pair_blacklist": [
    ]
  },
  "pairlists": [
    {
      "method": "StaticPairList"
    }
  ],
  "bot_name": "freqtrade",
  "initial_state": "running",
  "force_entry_enable": false,
  "internals": {
    "process_throttle_secs": 5
  }
}

```

</details>

<details>
  <summary>Strategy:</summary>

``` python
from typing import Optional

import numpy as np  # noqa
import pandas as pd  # noqa
from pandas import DataFrame

from freqtrade.strategy import (IStrategy)

class SampleStrategy(IStrategy):
    INTERFACE_VERSION = 3

    can_short: bool = False
    minimal_roi = {
        "600000": -1
    }
    stoploss = -1.0
    # Trailing stoploss
    trailing_stop = False
    timeframe = '5m'
    # Run "populate_indicators()" only for new candle.
    process_only_new_candles = True

    # These values can be overridden in the config.
    use_exit_signal = True
    exit_profit_only = False
    ignore_roi_if_entry_signal = False

    startup_candle_count: int = 30

    # Optional order type mapping.
    order_types = {
        'entry': 'limit',
        'exit': 'limit',
        'stoploss': 'market',
        'stoploss_on_exchange': False
    }

    order_time_in_force = {
        'entry': 'GTC',
        'exit': 'GTC'
    }

    def informative_pairs(self):
        return []

        return dataframe

        dataframe.loc[
            :,
            'enter_long'] = 1

        dataframe.loc[
            :,
            'enter_short'] = 0

        return dataframe

    def populate_exit_trend(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
        dataframe.loc[
            :,

            'exit_long'] = 0

        dataframe.loc[
            :,
            'exit_short'] = 1

        return dataframe

    def adjust_trade_position(self, trade, current_time,
                              current_rate: float, current_profit: float,
                              min_stake: Optional[float], max_stake: float,
                              current_entry_rate: float,
                              current_exit_rate: float,
                              current_entry_profit: float,
                              current_exit_profit: float, **kwargs) -> Optional[
      float]:
      return 100
```
</details>

Run:

`freqtrade backtesting --config user_data/config.json --strategy-path user_data/strategies --strategy SampleStrategy --timeframe 5m --cache none --timerange=20230205-20230210`

Previously took 2 minutes, now takes 45 seconds.

### DIFF
--- a/freqtrade/strategy/strategy_wrapper.py
+++ b/freqtrade/strategy/strategy_wrapper.py
@@ -5,6 +5,7 @@ from typing import Any, Callable, TypeVar, cast
 from freqtrade.exceptions import StrategyError
 from freqtrade.persistence import Trade
 
+
 logger = logging.getLogger(__name__)
 
 
--- a/freqtrade/strategy/strategy_wrapper.py
+++ b/freqtrade/strategy/strategy_wrapper.py
@@ -1,10 +1,9 @@
 import logging
-from copy import deepcopy
 from functools import wraps
 from typing import Any, Callable, TypeVar, cast
 
 from freqtrade.exceptions import StrategyError
-
+from freqtrade.persistence import Trade
 
 logger = logging.getLogger(__name__)
 
@@ -23,7 +22,16 @@ def strategy_safe_wrapper(f: F, message: str = "", default_retval=None, supress_
         try:
             if 'trade' in kwargs:
                 # Protect accidental modifications from within the strategy
-                kwargs['trade'] = deepcopy(kwargs['trade'])
+                trade = kwargs['trade']
+                if isinstance(trade, Trade):
+                    if trade in Trade.query.session.dirty:
+                        Trade.commit()
+                    return_vals = f(*args, **kwargs)
+                    if trade in Trade.query.session.dirty:
+                        logger.warning(f"The `trade` parameter have changed "
+                                       f"in the function `{f.__name__}`, this may "
+                                       f"lead to unexpected behavior.")
+                    return return_vals
             return f(*args, **kwargs)
         except ValueError as error:
             logger.warning(
--- a/tests/strategy/test_interface.py
+++ b/tests/strategy/test_interface.py
@@ -838,10 +838,10 @@ def test_strategy_safe_wrapper(value):
 
 
 @pytest.mark.usefixtures("init_persistence")
-def test_strategy_safe_wrapper_trade_copy(fee):
+def test_strategy_safe_wrapper_trade_mutate_warning(fee, caplog):
     create_mock_trades(fee)
 
-    def working_method(trade):
+    def working_method_mutate(trade):
         assert len(trade.orders) > 0
         assert trade.orders
         trade.orders = []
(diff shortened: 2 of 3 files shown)


freqtrade/strategy/strategy_wrapper.py
```diff
@@ -23,7 +23,16 @@ def wrapper(*args, **kwargs):
         try:
             if 'trade' in kwargs:
                 # Protect accidental modifications from within the strategy
-                kwargs['trade'] = deepcopy(kwargs['trade'])
+                trade = kwargs['trade']
+                if isinstance(trade, Trade):
+                    if trade in Trade.query.session.dirty:
+                        Trade.commit()
```
xmatthias, line 29: This will add random commits at random points in the process - i'm not entirely sure that this won't have unintended side-effects.

xmatthias (changes requested):
For live, this is out of the question. 
Disallowing direct modifications of the trade object is fully intentional (there's been severe problems in the past), and there are no plans on removing this.

in live, the performance gain will also be marginal and therefore neglectible - compared to the risk you run on a modified trade-object - it's not worth it.

---

considering the performance gains you're seeing - i guess we can skip the deepcopy if we're in backtesting/hyperopt - the best way to check for that is probably through `Trade.use_db` - which will be False in Backtesting - (or via `is_instance`, as you're doing currently, but in reverse). 

But for live - this is a change we're not going to make.

xmatthias:
I don't really see why you deleted this - i think with the proposed changes (removing the block only for backtesting) - this could be a nice enhancement.

raphaelstar:
I need trade mutation. Forbidding it forces me to fork freqtrade to fit my needs, I can't contribute and maintain two diverging versions.

xmatthias:
quite honestly - if you **need** trade mutation - then you're doing something wrong.
It will surely have other, unintended side-effects.

You're obviously free to maintain your own fork - when reporting problems / issues in the future, please make sure to test if it's actually a problem with freqtrade or with your fork - this includes not reusing a database that you did trade-mutations on.

PR #8178, opened 2023-02-16
Update binance_leverage_tiers.json

ShahAnuj2610:
Update binance_leverage_tiers.json

### DIFF
--- a/freqtrade/exchange/binance_leverage_tiers.json
+++ b/freqtrade/exchange/binance_leverage_tiers.json
@@ -1,103 +1,5 @@
 {
-  "1000LUNC/BUSD:BUSD": [
-    {
-      "tier": 1.0,
-      "currency": "BUSD",
-      "minNotional": 0.0,
-      "maxNotional": 5000.0,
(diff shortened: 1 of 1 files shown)


xmatthias:
From afar, i'd say this file was generated with either an old version of ccxt, or with a faulty configuration (which we also did have on the documentation ...).
Pairs are downgraded from (for example) `BTC/USDT:USDT` to `BTC/USDT` - which does no longer correspond to the correct naming freqtrade futures pairs - and is as such breaking tons of tests (as well as wouldn't work for freqtrade, either).

I've updated the docs to reflect the correct mode to initialize ccxt to generate this file (which is with "defaultType": "swap").
The best approach will be to do a clean commit on top of develop and resubmit that new commit as new PR. Thanks

Just to double-confirm this - the script to generate this file is in the [docs](https://www.freqtrade.io/en/latest/developer/#update-binance-cached-leverage-tiers).

xmatthias:
For now - don't mind - updated them in 31c7b3e136a53ff402c51ab096cc24b76b9029ee.

However thanks for your effort!

PR #8181, opened 2023-02-16
* fixed loading existing model for backtest

amalysh:
Fixed check for existing RL model.

### DIFF
--- a/freqtrade/freqai/RL/BaseReinforcementLearningModel.py
+++ b/freqtrade/freqai/RL/BaseReinforcementLearningModel.py
@@ -320,6 +320,22 @@ class BaseReinforcementLearningModel(IFreqaiModel):
 
         return prices_train, prices_test
 
+    def model_exists(self, dk: FreqaiDataKitchen) -> bool:
+        """
+        Given a pair and path, check if a model already exists
+        :param pair: pair e.g. BTC/USD
+        :param path: path to model
+        :return:
+        :boolean: whether the model file exists or not.
+        """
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/RL/BaseReinforcementLearningModel.py
```diff
@@ -320,6 +320,22 @@ def build_ohlc_price_dataframes(self, data_dictionary: dict,
...
 
+    def model_exists(self, dk: FreqaiDataKitchen) -> bool:
+        """
+        Given a pair and path, check if a model already exists
+        :param pair: pair e.g. BTC/USD
+        :param path: path to model
+        :return:
+        :boolean: whether the model file exists or not.
+        """
+        path_to_modelfile = Path(dk.data_path / f"{dk.model_filename}_model.zip")
```
xmatthias, line 331: @robcaulk would it make sense to have the extension dynamic / in a variable someplace? (.zip for RL models, .joblib for joblib models, .h5 for keras models (where the same thing would be broken, too)) ? 

That'd avoid overwriting / duplicating a whole function just because 3-5 letters are different ...

robcaulk:
Thanks for finding this issue @amalysh , but I think as @xmatthias points out - it is better to avoid duplicate code. I have tried to help answer this suggestion in #8182 . Please take a look, and possibly test that branch. Thank you.

freqtrade/freqai/RL/BaseReinforcementLearningModel.py
robcaulk: Please see #8182 for a suggested approach that avoids duplicate code.

PR #8184, opened 2023-02-17
Feature market direction

### LINKED ISSUE #8011: New variable in Strategy class that can be changed via telegram when bot is running
Variable name :
market_direction = "long" , "short" , ("even" or "side" - one of these 2, which you think fits better as description when market doesn't have a clear direction) 
Default startup value will be set in the strategy class, with one of the above 3 parameter.

Access and change value (in running instance) via telegram through command:
/marketdir long
/marketdir short
/marketdir side
so methods that will use market_direction or self.market_direction variable will get the updated (via telegram) value.

This will allow to easy change strategy behavior from phone when you don't have access to PC, when market condition are changing.

LangLazy:
## Summary

Adds new user set optional field which represents market direction, as well as a telegram command allowing users to change this field.

Solve the issue: #8011 

## Quick changelog

- Added new command ` \marketdir <param>` to change the marketdir variable

## What's new?
- Added new field `market_direction `to the `IStrategy `class 
- Added a new enum `MarketDirection`
- Added new command ` \marketdir <param>` to change the marketdir variable
- Added a test to check if the functionality works as intended

The goal of this PR is to allow users to have an internal variable they can control to signify what they think the market conditions are. The default value for this if not set by the user is `MarketDirection.NONE` meaning if a user has not chosen to set up this value, it will not hold any meaning or influence anything else.

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -129,7 +129,8 @@ class Telegram(RPCHandler):
             r'/weekly$', r'/weekly \d+$', r'/monthly$', r'/monthly \d+$',
             r'/forcebuy$', r'/forcelong$', r'/forceshort$',
             r'/forcesell$', r'/forceexit$',
-            r'/edge$', r'/health$', r'/help$', r'/version$', r'/marketdir (long|short|even|none)$'
+            r'/edge$', r'/health$', r'/help$', r'/version$', r'/marketdir (long|short|even|none)$',
+            r'/marketdir$'
         ]
         # Create keys for generation
         valid_keys_print = [k.replace('$', '') for k in valid_keys]
@@ -1495,8 +1496,9 @@ class Telegram(RPCHandler):
             "*/count:* `Show number of active trades compared to allowed number of trades`\n"
             "*/edge:* `Shows validated pairs by Edge if it is enabled` \n"
             "*/health* `Show latest process timestamp - defaults to 1970-01-01 00:00:00` \n"
-            "*/marketdir [long | short | even | none]:* `Updates the user managed variable"
-            " that represents the current market direction` \n"
+            "*/marketdir [long | short | even | none]:* `Updates the user managed variable "
+            "that represents the current market direction. If no direction is provided `"
+            "`the currently set market direction will be output.` \n"
 
             "_Statistics_\n"
             "------------\n"
@@ -1691,16 +1693,28 @@ class Telegram(RPCHandler):
         :return: None
         """
         if context.args and len(context.args) == 1:
-            new_market_dir = context.args[0]
-            if new_market_dir == "long":
-                self._rpc._update_market_direction(MarketDirection.LONG)
-            elif new_market_dir == "short":
-                self._rpc._update_market_direction(MarketDirection.SHORT)
-            elif new_market_dir == "even":
-                self._rpc._update_market_direction(MarketDirection.EVEN)
-            elif new_market_dir == "none":
-                self._rpc._update_market_direction(MarketDirection.NONE)
+            new_market_dir_arg = context.args[0]
+            old_market_dir = self._rpc._get_market_direction()
+            new_market_dir = None
+            if new_market_dir_arg == "long":
+                new_market_dir = MarketDirection.LONG
+            elif new_market_dir_arg == "short":
+                new_market_dir = MarketDirection.SHORT
+            elif new_market_dir_arg == "even":
+                new_market_dir = MarketDirection.EVEN
+            elif new_market_dir_arg == "none":
+                new_market_dir = MarketDirection.NONE
+
+            if new_market_dir is not None:
+                self._rpc._update_market_direction(new_market_dir)
+                self._send_msg("Successfully updated market direction"
+                               f" from *{old_market_dir}* to *{new_market_dir}*.")
             else:
-                raise RPCException("Invalid market direction provided")
+                raise RPCException("Invalid market direction provided. \n"
+                                   "Valid market directions: *long, short, even, none*")
+        elif context.args is not None and len(context.args) == 0:
+            old_market_dir = self._rpc._get_market_direction()
+            self._send_msg(f"Currently set market direction: *{old_market_dir}*")
         else:
-            raise RPCException("Invalid usage of command /marketdir.")
+            raise RPCException("Invalid usage of command /marketdir. \n"
+                               "Usage: */marketdir [short |  long | even | none]*")
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -129,7 +129,7 @@ class Telegram(RPCHandler):
             r'/weekly$', r'/weekly \d+$', r'/monthly$', r'/monthly \d+$',
             r'/forcebuy$', r'/forcelong$', r'/forceshort$',
             r'/forcesell$', r'/forceexit$',
-            r'/edge$', r'/health$', r'/help$', r'/version$', r'/marketdir$'
+            r'/edge$', r'/health$', r'/help$', r'/version$', r'/marketdir \d+$'
         ]
         # Create keys for generation
         valid_keys_print = [k.replace('$', '') for k in valid_keys]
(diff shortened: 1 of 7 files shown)


LangLazy:
Let me know if the unit test and the way I implemented the functionality follow the style of the rest of the code. I would love to get any feedback on what I could do better.

freqtrade/rpc/telegram.py
```diff
@@ -129,7 +129,7 @@ def _init_keyboard(self) -> None:
             r'/weekly$', r'/weekly \d+$', r'/monthly$', r'/monthly \d+$',
             r'/forcebuy$', r'/forcelong$', r'/forceshort$',
             r'/forcesell$', r'/forceexit$',
-            r'/edge$', r'/health$', r'/help$', r'/version$'
+            r'/edge$', r'/health$', r'/help$', r'/version$', r'/marketdir$'
```
xmatthias: i think this is pointless.

it'll allow you to create a button `/marketdir` (without anything after it) - but if you call `/marketdir` without argument, it'll do nothing based on the handler, which requires 1 argument.

freqtrade/rpc/telegram.py
```diff
@@ -1677,3 +1678,26 @@ def _send_msg(self, msg: str, parse_mode: str = ParseMode.MARKDOWN,
...
+        """
+        Handler for /marketdir.
+        Updates the bot's market_direction
+        :param bot: telegram bot
+        :param update: message update
+        :return: None
+        """
+        if context.args and len(context.args) == 1:
+            new_market_dir = context.args[0]
+            match new_market_dir:
```
xmatthias: You can't use match on freqtrade for now (probably not for years).

We do support python 3.8-3.10 - match was only added in 3.10, so this will fail on all prior versions (which tests confirm as "invalid syntax" errors ...

freqtrade/rpc/telegram.py
```diff
@@ -1677,3 +1678,26 @@ def _send_msg(self, msg: str, parse_mode: str = ParseMode.MARKDOWN,
...
+                case "long":
+                    self._rpc._freqtrade.strategy.market_direction = MarketDirection.LONG
```
xmatthias: conceptually, i'd probably move the actual assignment to the rpc class (which would be mostly aligned how other methods in this regards are setup).

That'll make this (eventually) available easily for other rpc methods (namely, the API) - should we decide to add support for that there as well.
Not a hard requirement - but adding this since you asked :)

xmatthias (commented):
we'll clearly need some documentation - otherwise it's a pointless feature nobody will know how to use.

freqtrade/rpc/telegram.py
```diff
@@ -129,7 +129,7 @@ def _init_keyboard(self) -> None:
             r'/weekly$', r'/weekly \d+$', r'/monthly$', r'/monthly \d+$',
             r'/forcebuy$', r'/forcelong$', r'/forceshort$',
             r'/forcesell$', r'/forceexit$',
-            r'/edge$', r'/health$', r'/help$', r'/version$'
+            r'/edge$', r'/health$', r'/help$', r'/version$', r'/marketdir \d+$'
```
xmatthias: `\d+` is wrong - as that is regex speak for a number - but we expect `"long", "short", "even", ...`).

Possible solutions will be either generally allowing a string `/marketdir \s+$` - or limiting to allowed strings  `/marketdir (long|short|...)$` (both solutions untested - but most likely somewhat working).

I'd suggest to actually try this (this defines the "valid" entries for [telegram keyboard config](https://www.freqtrade.io/en/stable/telegram-usage/#create-a-custom-keyboard-command-shortcut-buttons)).
You can also use [regex101](https://regex101.com/) to make finding the correct regex easier.

The rest looks fine (from a code perspective) - but is missing docs  ;)

LangLazy: Thanks for the pointers!

I was just a bit confused by what you meant in the last part. Do you suggest implementing the regex you wrote in place of what I have? I didn't quite get what you meant by (not sure what "this" is referencing)

> (this defines the "valid" entries for [telegram keyboard config](https://www.freqtrade.io/en/stable/telegram-usage/#create-a-custom-keyboard-command-shortcut-buttons))

Could you elaborate a bit on what you meant by this?

xmatthias: "this" -> the regex used.

there's a pretty well defined list of acceptable commands - (e.g. `/marketdir long`) - which are currently disallowed (\d+ matches numbers only).
`/marketdir 2` (which the current would allow) is not among the actually working commands ...

xmatthias (changes requested):
From a technical perspective, this works quite well. Now it's the details (mainly around usability).

what i think is still missing:
* documentation on how to actually use this within the strategy (i know it's `self.market_direction` - but how will a user know, purely from looking at the docs, without being aware of this functionality / this PR?)
* command feedback
* an entry in the `/help` section of the telegram commands (this helps users looking at it remote only to know WHY the command didn't work)
 
ref 2:
image

a user get's no feedback that the direction was set (misstyping the command has the same effect) -  neither is the currently set direction visible when not passing in an argument (there could be another option for this - but this was the quickest way i could find)

To better explain what i mean:

image

after the 2 above comments - what is market-dir? and how can i tell (as a user)?

LangLazy:
Thanks for the feedback!

I have since added command feedback and more descriptive error messages. Let me know if these fix the usability issues. 
Moreover, I added the ability for users to do `/marketdir` which would allow them to find out what market direction is currently set.

I had a couple of questions about your above comments.

I had already added the information about the `/marketdir` command to the `/help` command before you made your last comment, was there somewhere else I needed to update for the `/help` command?

Also regarding

> documentation on how to actually use this within the strategy (i know it's self.market_direction - but how will a user know, purely from looking at the docs, without being aware of this functionality / this PR?)

Where in the documentation should I add that this variable is available for users? I tried looking around but the closest I could find was the page on `strategy-customization` and even then I did not really know where would be appropriate for my change to go.

xmatthias:
for now, we'll keep the whole documentation part together in the telegram page.

there's not too much to add other than "use it like this" - and i see this as an edge-case scenario, which very likely not many people will actually use.

LangLazy:
Ah, I see, thanks for pointing that out. I think with that last change everything you brought up for review should be addressed. Let me know if anything needs to be updated before this is merged :)

xmatthias (approved):
LGTM :+1: 

I do see this as beta feature though. If we see this as creating problems (strategies using this will not really produce reproducible results) - or otherwise as problematic, this might be removed in the future.

VietnameseRick:
Hi guys, are there any example on how to implement `self.market_direction` to the strategy ?

xmatthias:
well usage will be as simple as `self.market_direction == MarketDirection.LONG` (for example).

How you're using that within the context of your strategy is up to your imagination.
Personally, i have my doubts on the benefit of this - as this is something that's set externally (at random, basically), which you can't backtest due to that.

VietnameseRick:
> well usage will be as simple as `self.market_direction == MarketDirection.LONG` (for example).
> 
> How you're using that within the context of your strategy is up to your imagination. Personally, i have my doubts on the benefit of this - as this is something that's set externally (at random, basically), which you can't backtest due to that.

I can't share you much details about how I would use it but to be honest, it does help really much, market always change and the bias of trader is what make them out performance compare to bot but `self.market_direction` is what really change the game. Sometime the bias come from the human is better than the indicator

pilchor:
Hello,

Does anyone have a working strategy example that uses `self.market_direction`?

When I try to use it, no trades are opened, even though `self.market_direction` is correctly set to `'even'` or `'long'` (as seen in the logs).

Here’s a snippet of my code:

```python
    logger.info(f"Current market direction: {self.market_direction}")
    
    if self.market_direction in ['long', 'even']:
        dataframe.loc[
            (
                (dataframe['close'] > dataframe['senkanA']) &
                (dataframe['close'] > dataframe['senkanB']) &
                (dataframe['close'] > dataframe['tenkan']) &
                (dataframe['senkanB'] > dataframe['senkanA']) &
                (dataframe['tenkan'] > dataframe['kijun'])
            ),
            'enter_long'] = 1

    return dataframe

xmatthias:
well market-direction is one part - but you'll still have to have the scenario where all your conditions are true.

Assuming your logic triggers - then it'll set enter_long - but only if **all** conditions within that statement match.
if you struggle to get this running - try with a simpler condition which triggers more frequently? (a condition comparing 5 lines can have days / weeks without trade.

Issue #8187, opened 2023-02-19
Can't use journald because systemd python package is outdated


AchmadFathoni:
## Describe your environment

  * Operating system: Arch Linux
  * Python Version:  3.10.0
  * CCXT version: 2.7.12
  * Freqtrade Version: 2023.1

## Describe the problem:
Using `--logifle journald` after installing systemd python package as stated in [manual](https://www.freqtrade.io/en/stable/advanced-setup/#logging-to-journald) will produce python error.

### Steps to reproduce:

  1. Install systemd python package `pip install systemd`
  2. Run the bot `freqtrade trade --logfile journald`
  
### Observed Results:

  * What happened? Bot throw error and stop working. systemd python package version is old and only works with python 3.6. The new package by same author has  renamed as `cysystemd`.
  * What did you expect to happen? Bot will run without error while sending log to journald.

### Relevant code exceptions or logs
```
2023-02-19 22:04:58,298 - freqtrade - INFO - freqtrade 2023.1
2023-02-19 22:04:58,424 - freqtrade.worker - INFO - Starting worker 2023.1
2023-02-19 22:04:58,424 - freqtrade.configuration.load_config - INFO - Using config: config.json ...
2023-02-19 22:04:58,426 - freqtrade.loggers - INFO - Verbosity set to 0
2023-02-19 22:04:58,426 - freqtrade.commands.trade_commands - ERROR - module 'collections' has no attribute 'Mapping'
2023-02-19 22:04:58,426 - freqtrade - ERROR - Fatal exception!
Traceback (most recent call last):
  File "/run/media/toni/Windows/Shared/Source/freqtrade/freqtrade/commands/trade_commands.py", line 18, in start_trading
    worker = Worker(args)
  File "/run/media/toni/Windows/Shared/Source/freqtrade/freqtrade/worker.py", line 37, in __init__
    self._init(False)
  File "/run/media/toni/Windows/Shared/Source/freqtrade/freqtrade/worker.py", line 50, in _init
    self._config = Configuration(self._args, None).get_config()
  File "/run/media/toni/Windows/Shared/Source/freqtrade/freqtrade/configuration/configuration.py", line 42, in get_config
    self.config = self.load_config()
  File "/run/media/toni/Windows/Shared/Source/freqtrade/freqtrade/configuration/configuration.py", line 84, in load_config
    self._process_logging_options(config)
[...]

PR #8198, opened 2023-02-20
Fix outdated systemd related exception text.

AchmadFathoni:
## Summary

Complement e9c64c58 that solves the issue: #8187 

## What's new?

Now the user will prompted by python exception to install `cysystemd` instead `systemd` package.


xmatthias (commented):
hah nice catch! 👍

xmatthias (approved):

PR #8203, opened 2023-02-21
Add buffer_train_data_candles feature

robcaulk:
Buffer the start and end of the timerange. This is used *after* the indicators are populated.
The main example use is when predicting maxima and minima, the argrelextrema function  cannot know the maxima/minima at the edges of the timerange. To improve model accuracy, it is best to compute argrelextrema on the full timerange and then use this function to cut off the edges (buffer) by the kernel.

In another case, if the targets are set to a shifted price movement (the FreqaiExampleStrategy), this buffer is unnecessary because the shifted candles at the end of the timerange will be NaN and FreqAI will automatically cut those off of the training dataset.

Activate with:

```
"buffer_train_data_candles": 100
```

In the `feature_parameters` config. Defaults to 0.


xmatthias (approved):
LGTM :+1:

PR #8095, opened 2023-02-02
remove follow mode in favor of producer consumer

robcaulk:
Producer consumer mode replaces follow mode for all applications. Removing follow mode with this PR to clean the code up and avoid possible confusion by new users.

PR #8210, opened 2023-02-22
Allow user to control number of historical model files

robcaulk:
`purge_old_models` now takes an integer to indicate how many models should be kept. Otherwise old boolean type functionality remains the same in case users do not wish to change their configs.

```
| `purge_old_models` | Number of models to keep on disk (not relevant to backtesting). Default is 2, which means that dry/live runs will keep the latest 2 models on disk. Setting to 0 keeps all models. This parameter also accepts a boolean to maintain backwards compatibility. <br> **Datatype:** Integer. <br> Default: `2`.
```

This PR also cleans up some remnants from #8095.


docs/freqai-parameter-table.md
```diff
@@ -15,7 +15,7 @@ Mandatory parameters are marked as **Required** and have to be set in one of the
 | `identifier` | **Required.** <br> A unique ID for the current model. If models are saved to disk, the `identifier` allows for reloading specific pre-trained models/data. <br> **Datatype:** String.
 | `live_retrain_hours` | Frequency of retraining during dry/live runs. <br> **Datatype:** Float > 0. <br> Default: `0` (models retrain as often as possible).
 | `expiration_hours` | Avoid making predictions if a model is more than `expiration_hours` old. <br> **Datatype:** Positive integer. <br> Default: `0` (models never expire).
-| `purge_old_models` | Delete all unused models during live runs (not relevant to backtesting). If set to false (not default), dry/live runs will accumulate all unused models to disk. If <br> **Datatype:** Boolean. <br> Default: `True`.
+| `purge_old_models` | Number of models to keep on disk (not relevant to backtesting). Default is 2, dry/live runs will keep 2 models on disk. Setting to 0 keeps all models. If <br> **Datatype:** Boolean. <br> Default: `2`.
```
xmatthias:
```suggestion
| `purge_old_models` | Number of models to keep on disk (not relevant to backtesting). Default is 2, dry/live runs will keep 2 models on disk. Setting to 0 keeps all models. If <br> **Datatype:** Integer. <br> Default: `2`.
```

In my understanding, boolean is supported for backward compatibility - but our aim is to have people use ints for clarity ?

robcaulk: Thanks for catching this, I also updated a bit more.

xmatthias (approved):
LGTM :+1:

PR #8212, opened 2023-02-23
setup.sh: actually install all dependencies

TheJoeSchr:
I thought if asked to install all dev dependencies, all should actually be installed. or are all requirements from each `requirements-*.txt` duplicated inside `requirenemts-dev.sh`? I don't assume so?

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -62,6 +62,10 @@ function updateenv() {
     if [[ $REPLY =~ ^[Yy]$ ]]
     then
         REQUIREMENTS=requirements-dev.txt
+        REQUIREMENTS_PLOT="-r requirements-plot.txt"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -62,6 +62,10 @@ function updateenv() {
     if [[ $REPLY =~ ^[Yy]$ ]]
     then
         REQUIREMENTS=requirements-dev.txt
+        REQUIREMENTS_PLOT="-r requirements-plot.txt"
```
xmatthias, line 65: requirements-dev.txt actually includes all other dependencies via `-r requirements-xxx.txt` already - so if you run only `pip install requirements-dev.txt` - you'll get ALL dependencies (from all requirements files) except the doc ones - which are irrelevant.

TheJoeSchr:
Closed then, sorry about that

PR #8216, opened 2023-02-24
Close training and eval environments in ReinforcementLearner_multiproc

### LINKED ISSUE #8213: Memory leak when using ReinforcementLearner_multiproc. Endless Python processes generation.
## Describe your environment

  * Operating system: Linux Mint 21.1
  * Python Version: 3.10.10(`python -V`)
  * CCXT version: 2.8.17 (`pip freeze | grep ccxt`)
  * Freqtrade Version: docker-2023.2.dev-e56bf067 (`freqtrade -V` or `docker compose run --rm freqtrade -V` for Freqtrade running in docker)
  
## Describe the problem:

When using the _**ReinforcementLearner_multiproc.py**_ model, over time, a large number of Python instances are launched, which are closed only if the container is stopped or manually killed.

### Steps to reproduce:

1. Start container using appropriate config and strategy using **_ReinforcementLearner_multiproc_**
2. Open System Monitor.
3. Watch how Python instances are created when training starts, but not deleted when it ends.
  
### Observed Results:

When training starts for a particular ticker, a large number of Python instances are added, as expected.
When the training of the next ticker starts, new Python instances are added additionally, while the old ones are not deleted, although they show 0% CPU load.
In my opinion, Python instances should be deleted after training is complete, before new ones are added.

I was able to fix the issue on my setup by adding the following lines of code to the **_set_train_and_eval_environment_** method of the **_ReinforcementLearner_multiproc_** class, before the functions that create the training environments.

```
if self.train_env :
    self.train_env.close()
if self.eval_env :
    self.eval_env.close()
```

robcaulk:
Addresses #8213, not ready for merge until further testing/verification.  

Tested and looks good to go.

### DIFF
--- a/freqtrade/freqai/prediction_models/ReinforcementLearner_multiproc.py
+++ b/freqtrade/freqai/prediction_models/ReinforcementLearner_multiproc.py
@@ -34,6 +34,11 @@ class ReinforcementLearner_multiproc(ReinforcementLearner):
         train_df = data_dictionary["train_features"]
         test_df = data_dictionary["test_features"]
 
+        if self.train_env:
+            self.train_env.close()
+        if self.eval_env:
+            self.eval_env.close()
+
         env_info = self.pack_env_dict(dk.pair)
 
         env_id = "train_env"


dblyzniuk (approved):
This solved the problem for my system. No negative effects were observed. Tested for 48 hours.

robcaulk:
> This solved the problem for my system. No negative effects were observed. Tested for 48 hours.

Thanks for testing that, I confirm the fix.

xmatthias (approved):
great find! :+1:

PR #8220, opened 2023-02-24
removed redundant dependencies from environment.yml

asuiu:
## Summary
Currently the `environment.yml` contains unversioned dependencies, which is a risk of breaking functionality due to untested versions of dependencies.

Additionally, the `requirements.txt` contains the versioned dependencies, so it's safer to replace all the redundantly specified dependencies in `environments.yml` with `-r requirements.txt` according to DRY principle.

I've also added `libta-lib=0.4.0` from conda-forge channel, which is required by `ta-lib` from `requirements.txt`.

## What's new?

Improved automatic dependency management of the project.


environment.yml
```diff
@@ -3,73 +3,9 @@ channels:
...
-        - pytest-random-order
-        - ccxt
-        - flake8-tidy-imports
-        - -e .
-    # - python-rapidjso
+  - python>=3.8,<=3.10
+  - pip
+  - libta-lib=0.4.0 # this is required for ta-lib from requirements.txt
+  - pip:
+    - -r requirements.txt
```
xmatthias: `requirements.txt` is not the full equivalent to the dependencies listed before.
We'll be missing some dependencies - which will result in a barebones installation (no plotting, no hyperopt...). 
Now i don't mind that necessarily - but it's probably a partially unintended change.

---

Now assuming the above is just a differently named ta-lib - then we'll end up with ta-lib installed twice (with potentially conflicting versions).
`requirements.txt` includes ta-lib - which WILL be installed (pip ignores dependencies already installed by conda). 

With that - the whole conda-file has become pointless.

environment.yml
```diff
@@ -3,73 +3,9 @@ channels:
...
-        # - py_find_1st
-        - tables
-        - pytest-random-order
-        - ccxt
-        - flake8-tidy-imports
-        - -e .
-    # - python-rapidjso
+  - python>=3.8,<=3.10
+  - pip
+  - libta-lib=0.4.0 # this is required for ta-lib from requirements.txt
```
xmatthias: [Conda forge](https://anaconda.org/conda-forge/libta-lib) has basically no information about this package.
If we assume this is an official ta-lib library - then i'd expect a link to it from the ta-lib documentation ... 

I'll not be accepting unknown dependencies (from unknown, potentially malicious sources).

xmatthias (changes requested):
Yes, dependencies here are not pinned, which is not ideal. 
Best complain to dependabot for not being able to update conda environment files :smile: 
I'm more than happy to update the dependabot config to update the dependencies here (which would mean we could have pinned dependencies) - which is (was last time i checked) however impossible.

---

This pr is making the conda environment file pointless in my eyes.
The way you changed it to now will be installing everything from pip.
A better approach would then be to simply remove this file and change the documentation to either not use conda, or create an empty conda environment (without environment file), and then install everything from pip.

The only real reason to use the conda-file is if you're on windows and have trouble installing packages from pypi which don't include wheels (most prominently, ta-lib, but i think there's a few others) - where conda provides binary packages, while pypi does not.

With this change - this only usecase became pointless / irrelevant - so i don't really see what sense the conda file still makes.

environment.yml
asuiu: Yes, make sense. I've removed the `environment.yml` and updated the documentation as advised.

environment.yml
asuiu: 1. I noticed that the `libta-lib` library appears to have come from the same source as the popular `ta-lib` package from conda-forge. It's interesting to note that `libta-lib` has twice the number of downloads as `ta-lib`, and both packages seem to be quite popular.
2. While it's not entirely clear what you mean by "unknown source," it's worth noting that some of the packages used already in `requirements.txt` might be considered as such and possess similar risks, which is very hard to mitigate as manual analysis is quite tedious. A good practice might be to rely on automated security checkers of dependencies such as Safety or Dependabot from Github, which seems to be already used here.
3. It's also worth noting that the ta-lib library, which is already included in requirements.txt, has a link to this conda-forge library.

asuiu:
@xmatthias - thanks for taking time to review!

I've gone ahead and made the changes you suggested, so whenever you have a chance, I'd really appreciate it if you could take another look and let me know what you think. Thanks again!

environment.yml
xmatthias: for every package in `requirements.txt` - there's a corresponding pypi site - which links to the original source.
For all smaller packages (not major ones like pandas/numpy, ...) - we did this analysis before adding them (is it a popular package / still actively maintained).
now these things change and we don't look at this on an ongoing basis - but pypi makes this far easier than conda.


maybe it's "lack of experience with conda" - but i don't find this information for conda packages - all conda-forge packages seem to be "magically there" (without description) - without obvious link to the actual source project. so if i'm the first to publish a package on conda-forge - it's name is mine.

asuiu:
Hi @xmatthias!  
I noticed that you left a comment on the pull request, but I noticed that it hasn't been merged yet. Is there anything else that needs to be addressed before it can be merged? Let me know how I can assist in getting the pull request approved.

xmatthias:
> Hi @xmatthias! I noticed that you left a comment on the pull request, but I noticed that it hasn't been merged yet. Is there anything else that needs to be addressed before it can be merged? Let me know how I can assist in getting the pull request approved.

not really, no, sorry
i've been busy getting some other things fixed (and life, too, occasionally) - i'll review, and probably merge shortly

xmatthias (approved):
Thanks for your time!

LGTM :+1:

PR #8231, opened 2023-02-26
Fix dependencies & tests to run on Windows

asuiu:
## Summary
Fix dependencies and tests to run on Windows.

Currently some of the dependencies are incompatible with Windows, as well as some tests fail to run on Windows and this PR fixes to make all tests pass by ignoring those incompatible, and not installing incompatible packages on Windows.

## Quick changelog

- fix `requirements-freqai-rl.txt` to not install on Win platform incompatible packages;
- fix tests to skip running incompatible tests on Win platform;

## What's new?

No new features/functionalities.

### DIFF
--- a/requirements-freqai-rl.txt
+++ b/requirements-freqai-rl.txt
@@ -3,8 +3,8 @@
 
 # Required for freqai-rl
 torch==1.13.1
-stable-baselines3==1.7.0
+stable-baselines3==1.7.0; sys_platform != "win32"
 sb3-contrib==1.7.0
 # Gym is forced to this version by stable-baselines3.
 setuptools==65.5.1 # Should be removed when gym is fixed.
-gym==0.21
+gym==0.21; sys_platform != "win32"
--- a/tests/freqai/test_freqai_interface.py
+++ b/tests/freqai/test_freqai_interface.py
@@ -1,5 +1,6 @@
 import platform
 import shutil
+import sys
 from pathlib import Path
 from unittest.mock import MagicMock
 
@@ -16,6 +17,9 @@ from freqtrade.plugins.pairlistmanager import PairListManager
 from tests.conftest import create_mock_trades, get_patched_exchange, log_has_re
 from tests.freqai.conftest import get_patched_freqai_strategy, make_rl_config
 
+if sys.platform.startswith("win"):
+    pytest.skip("Skipping these tests as it depends on stable-baselines3==1.7.0 which doesn't support Windows",
+                allow_module_level=True)
 
 def is_arm() -> bool:
     machine = platform.machine()
--- a/tests/rpc/test_rpc_apiserver.py
+++ b/tests/rpc/test_rpc_apiserver.py
@@ -2,6 +2,7 @@
 Unit test file for rpc/api_server.py
 """
 import logging
+import sys
 import time
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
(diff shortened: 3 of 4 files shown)


requirements-freqai-rl.txt
```diff
@@ -3,8 +3,8 @@
 
 # Required for freqai-rl
 torch==1.13.1
-stable-baselines3==1.7.0
+stable-baselines3==1.7.0; sys_platform != "win32"
```
xmatthias, line 6: I disagree with this.

CI clearly shows that both installation, as well as tests are running fine on windows (from python 3.8-3.10).
It's therefore most likely related to your specific setup.

tests/freqai/test_freqai_interface.py
```diff
@@ -16,6 +17,9 @@
 from tests.conftest import create_mock_trades, get_patched_exchange, log_has_re
 from tests.freqai.conftest import get_patched_freqai_strategy, make_rl_config
 
+if sys.platform.startswith("win"):
```
xmatthias, line 20: We will not be skipping the whole freqAI tests on windows.
that's both unnecessary, as well as danerous.

best fix your setup instead.

tests/rpc/test_rpc_apiserver.py
```diff
@@ -1794,6 +1795,7 @@ def test_api_backtest_history(botclient, mocker, testdatadir):
     assert result2['backtest_result']['strategy'][strategy]
 
 
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Throws OSError on Windows")
 def test_health(botclient):
     ftbot, client = botclient
```
xmatthias, line 1800: The actual problem is not windows (look at CI for this).

I've seen this before though - and it's if you're running a test (on windows) on a negative timezone (i suspect in rare edgecases, it can also happen in live mode in these circumstance).

I'm happy to get a PR fixing that particular issue properly - but skipping this test is not the fix.

xmatthias (changes requested):
I disagree with basically every line in this.

it's fixing something that's related to your setup (not to windows specifically) - and skipping a whole segment of tests for no good reason (again - your specific setup).

The health part is very similar - it's environment specific (and pretty sure it's affecting you) - but skipping that test is the wrong fix.

asuiu:
@xmatthias 
Thank you for taking time to review. I apologize for the confusion and for overlooking the fact that there is already a working Windows Continuous Integration pipeline. In this context, as you mentioned, there is no point in skipping those tests since the pipeline works. It looks that is was something wrong with my set-up, software we use is not perfect. With the presence of passing Windows CI pipeline - this PR just doesn't make any sense.

Once again, I apologize for any inconvenience or confusion that my PR may have caused, and I appreciate your willingness to review and explain the issue.

I close this PR.

xmatthias:
As mentioned the "health" test is a different topic - which i struggle to recreate as it requires windows (which i don't use) on a negative timezone offset (basically US) - which i'm not in.

The fix for THAT would be great - it'll be in `rpc.py -> _health()` - where `astimezone()` is creating the rpoblem - and might need slightly different handling for `0` values (i traced it this far last time i looked at this)

asuiu:
> As mentioned the "health" test is a different topic - which i struggle to recreate as it requires windows (which i don't use) on a negative timezone offset (basically US) - which i'm not in.
> 
> The fix for THAT would be great - it'll be in `rpc.py -> _health()` - where `astimezone()` is creating the rpoblem - and might need slightly different handling for `0` values (i traced it this far last time i looked at this)

@xmatthias 
I have created a separate issue for this, for easier tracking: https://github.com/freqtrade/freqtrade/issues/8244

PR #8245, opened 2023-02-27
Fix last_process related bug in RPC.health (BUG-#8231)

### LINKED ISSUE #8244: RPC tests fail on Windows with local TimeZone UTC+2
## Describe your environment

  * Operating system: Windows 11
  * Python Version: Python 3.10.9 (`python -V`)
  * CCXT version: 2.8.17 (`pip freeze | grep ccxt`)
  * Freqtrade Version: freqtrade 2023.3.dev-44b100507 
 

## Describe the problem:

Two tests related to RPC fail on Windows where local timezone is set to UTC +X, where X>=2

### Steps to reproduce:

  1. Run tests on Windows 11 with TimeZone UTC +X, where X>=2
  
### Observed Results:

  * Tests fail

`rpc.test_rpc.test_rpc_health` with next log:
```python
tests\rpc\test_rpc_apiserver.py:1796 (test_health)
botclient = (<freqtrade.freqtradebot.FreqtradeBot object at 0x000001E75CA2F9A0>, <starlette.testclient.TestClient object at 0x000001E75CC9ACB0>)

    def test_health(botclient):
        ftbot, client = botclient
    
>       rc = client_get(client, f"{BASE_URI}/health")

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = tzlocal(), dt = datetime.datetime(1970, 1, 1, 0, 0, tzinfo=tzlocal())

    def _naive_is_dst(self, dt):
        timestamp = _datetime_to_timestamp(dt)
>       return time.localtime(timestamp + time.timezone).tm_isdst
E       OSError: [Errno 22] Invalid argument

C:\Users\ASU\miniconda3\envs\freqtrade\lib\site-packages\dateutil\tz\tz.py:260: OSError
```

and `rpc.test_rpc.test_rpc_health` with log:
```python
tests\rpc\test_rpc.py:1249 (test_rpc_health)
mocker = <pytest_mock.plugin.MockerFixture object at 0x00000240AB91AFE0>
default_conf = {'amend_last_stake_amount': False, 'cancel_open_orders_on_exit': False, 'candle_type_def': <CandleType.SPOT: 'spot'>, 'datadir': WindowsPath('D:/Work/freqtrade/tests/testdata'), ...}

    def test_rpc_health(mocker, default_conf) -> None:
    
        freqtradebot = get_patched_freqtradebot(mocker, default_conf)
        rpc = RPC(freqtradebot)
>       result = rpc._health()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = tzlocal(), dt = datetime.datetime(1970, 1, 1, 0, 0, tzinfo=tzlocal())

    def _naive_is_dst(self, dt):
        timestamp = _datetime_to_timestamp(dt)
>       return time.localtime(timestamp + time.timezone).tm_isdst
E       OSError: [Errno 22] Invalid argument

C:\Users\ASU\miniconda3\envs\freqtrade\lib\site-packages\dateutil\tz\tz.py:260: OSError
```

asuiu:
## Summary

Fixes bug in RPC.health related to the TimeZone that was reproducing only on Windows and on UTC-pozitive timezones.

Solve the issue: #8231 

#### Note: 
The main concern with this fix is the contract of the `/health` endpoint which should be able to handle `null`s in the json response.
As the alternative solution with less-risk impact is to change the default timestamp of `last_process` return to `1970-01-02 00:00:00`, that would fix the problem for all time zones, but that wouldn't be semantically correct since the last_process actually is undefined, so to maintain a logical consistency - using a null/None value would be a better choice.

Please let me know if the clients of the `/health` endpoint aren't able to handle the `null`s, and I'll change back the contract to datetime strings.

## Quick changelog

- `FreqtradeBot.last_process` by default is None now
- `RPC._health` method named changed to `health()` since this is public method, used by external clients.
- `Health` api schema updated to match the semantics of the returned data.
- tests fixed to pass

## What's new?

The `/health` endpoint now will return `null` for last_process timestamps if the bot hasn't been started yet.

### DIFF
--- a/freqtrade/rpc/api_server/api_schemas.py
+++ b/freqtrade/rpc/api_server/api_schemas.py
@@ -456,5 +456,5 @@ class SysInfo(BaseModel):
 
 
 class Health(BaseModel):
-    last_process: datetime
-    last_process_ts: int
+    last_process: Optional[datetime]
+    last_process_ts: Optional[int]
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -127,7 +127,7 @@ class FreqtradeBot(LoggingMixin):
                 for minutes in [0, 15, 30, 45]:
                     t = str(time(time_slot, minutes, 2))
                     self._schedule.every().day.at(t).do(update)
-        self.last_process = datetime(1970, 1, 1, tzinfo=timezone.utc)
+        self.last_process: Optional[datetime] = None
 
         self.strategy.ft_bot_start()
         # Initialize protections AFTER bot start - otherwise parameters are not loaded.
--- a/freqtrade/rpc/api_server/api_v1.py
+++ b/freqtrade/rpc/api_server/api_v1.py
@@ -346,4 +346,4 @@ def sysinfo():
 
 @router.get('/health', response_model=Health, tags=['info'])
 def health(rpc: RPC = Depends(get_rpc)):
-    return rpc._health()
+    return rpc.health()
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -89,7 +89,7 @@ class RPC:
     # Bind _fiat_converter if needed
     _fiat_converter: Optional[CryptoToFiatConverter] = None
 
-    def __init__(self, freqtrade: "FreqtradeBot") -> None:
+    def __init__(self, freqtrade) -> None:
         """
         Initializes all enabled rpc modules
         :param freqtrade: Instance of a freqtrade bot
--- a/freqtrade/rpc/rpc.py
+++ b/freqtrade/rpc/rpc.py
@@ -89,7 +89,7 @@ class RPC:
     # Bind _fiat_converter if needed
     _fiat_converter: Optional[CryptoToFiatConverter] = None
 
-    def __init__(self, freqtrade) -> None:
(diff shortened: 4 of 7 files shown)


freqtrade/rpc/api_server/api_schemas.py
```diff
@@ -456,5 +456,5 @@ class SysInfo(BaseModel):
 
 
 class Health(BaseModel):
-    last_process: datetime
-    last_process_ts: int
+    last_process: Optional[datetime]
+    last_process_ts: Optional[int]
```
xmatthias, line 460: @froggleston  i know you use the /health endpoint in frogtrade9000 ... is this going to cause problems? 
This would happen anywhere between start of the bot and the first iteration .... so only for a quite limited amount of time.

froggleston: No, I only use sysinfo in frogtrade9000 so no problems I can see.

xmatthias (approved):
LGTM :+1: 

thanks for taking the time to fix this!

PR #8272, opened 2023-03-03
Make strategy.bot_loop_start run once per candle in backtest

paranoidandy:
## Summary

This PR changes the backtesting behaviour of bot_loop_start from once per backtest to once per candle

## Quick changelog

- strategy_safe_wrapper(self.strategy.bot_loop_start, supress_error=True)() call moved
- updated docs

## Considerations

Docs may need extra changes to explain the change in behaviour from previously expected?


docs/bot-basics.md
```diff
@@ -57,10 +57,10 @@ This loop will be repeated again and again until the bot is stopped.
 
 * Load historic data for configured pairlist.
 * Calls `bot_start()` once.
-* Calls `bot_loop_start()` once.
 * Calculate indicators (calls `populate_indicators()` once per pair).
 * Calculate entry / exit signals (calls `populate_entry_trend()` and `populate_exit_trend()` once per pair).
 * Loops per candle simulating entry and exit points.
+  * Calls `bot_loop_start()` strategy callback.
   * Check for Order timeouts, either via the `unfilledtimeout` configuration, or via `check_entry_timeout()` / `check_exit_timeout()` strategy callbacks.
```
xmatthias, line 67: I'm ... not sure i like this (don't get me wrong - the doc corresponds to the change) after seeing the logic / sequence of events change (which is why I've put this comment here).

---

i think `bot_loop_start()` needs to be called one additional time (where it was, basically).

Assuming my (current) `bot_loop_start()` implementation was to load some additional data (maybe into a variable) - which i'm then using in my `populate_*()` methods to merge it to each pair (think about the sample we have in the docs - but with a backtesting fallback loading a historic file from disk).

this could be moved to `bot_start()` - but that will be wrong for live (it's now no longer reloading) - so i now need the logic split across 2 different methods to have roughly identical behavior in backtesting and live.

maybe the `bot_loop_start()` call can be kept initially as well - so it'll be called "one additional time".

freqtrade/optimize/backtesting.py
```diff
@@ -1159,6 +1158,7 @@ def backtest(self, processed: Dict,
         while current_time <= end_date:
             open_trade_count_start = LocalTrade.bt_open_open_trade_count
             self.check_abort()
+            strategy_safe_wrapper(self.strategy.bot_loop_start, supress_error=True)()
```
xmatthias: i think if we assume this callback to be called in backtesting over time - it might make sense to add a `current_time` argument to it (it's currently not needed as it's only called once in backtesting).

otherwise we'll know "it's been called 200 times" - but we won't know "it's now 2022-12-15 22:05:00" ...

docs/bot-basics.md
paranoidandy: I might need help with adapting tests as they're quite opaque to me. For instance:

https://github.com/freqtrade/freqtrade/blob/9573974c4798982a87c9e70a7a341ac4235ba956/tests/optimize/test_backtesting.py#L379

Should this be removed or changed to call_count == [numer_of_candles + 1] and if so it's not immediately obvious to me how to get that with the mocker here.

xmatthias: depends - assuming we reinstate the "startup" call before all populate_* calls - then i don't think we'll need changes to tests - so best leave that to the end.

xmatthias (approved):

PR #8273, opened 2023-03-04
Stop from open lev

### LINKED ISSUE #8129: stoploss_from_open has some ambiguous with leverage
## Describe your environment

  * Operating system: ____ Linux
  * Python Version: _____ (`python -V`) 3.10
  * CCXT version: _____ (`pip freeze | grep ccxt`) 1.95
  * Freqtrade Version: ____ (`freqtrade -V` or `docker compose run --rm freqtrade -V` for Freqtrade running in docker)
  
## Describe the problem:

In the doc explanation,  profit and stoploss is calculated with leverage.

And this function `stoploss_from_open` (in freqtrade/strategy/strategy_helper.py)  may wrong when enter short with leverage.  Because it's profit_rate may large than 1 , and  but open_relative_stop still less than 1, and stoploss be always negative, and function return 0.0. But in fact, it has positive profit.

### Steps to reproduce:

  1. _____
  2. _____
  3. _____
  
### Observed Results:

  * What happened?  
       stoploss_from_open may have  ambiguous  when used leverage
       In function dive leverage,  or argument explicitly explanates  without leverage
``` Python
              def stoploss_from_open(
                  open_relative_stop: float,
                  current_profit: float,
                  is_short: bool = False
                  ) -> float:
                  """
                  Given the current profit, and a desired stop loss value relative to the open price,
                  return a stop loss value that is relative to the current price, and which can be
                  returned from `custom_stoploss`.
              
                  The requested stop can be positive for a stop above the open price, or negative for
                  a stop below the open price. The return value is always >= 0.
              
                  Returns 0 if the resulting stop price would be above/below (longs/shorts) the current price
              
                  :param open_relative_stop: Desired stop loss percentage relative to open price
                  :param current_profit: The current profit percentage
                  :param is_short: When true, perform the calculation for short instead of long
                  :return: Stop loss value relative to current price
                  """
              
                  # formula is undefined for current_profit -1 (longs) or 1 (shorts), return maximum value
                  if (current_profit == -1 and not is_short) or (is_short and current_profit == 1):
                      return 1
              
                  if is_short is True:
                      stoploss = -1 + ((1 - open_relative_stop) / (1 - current_profit))
                  else:
                      stoploss = 1 - ((1 + open_relative_stop) / (1 + current_profit))
              
                  # negative stoploss values indicate the requested stop price is higher/lower
                  # (long/short) than the current price
                  return max(stoploss, 0.0)
```

  ```
  // paste your log here
  ```

xmatthias:
## Summary
stoploss_from_open adjusted for leverage

closes #8129

## todo

- [x] ensure calculation is what we expect
- [x] Documentation

### DIFF
--- a/tests/strategy/test_strategy_helpers.py
+++ b/tests/strategy/test_strategy_helpers.py
@@ -177,26 +177,30 @@ def test_stoploss_from_open(side, profitrange):
     ("long", 0.1, 0.2, 1, 0.08333333),
     ("long", 0.1, 0.5, 1, 0.266666666),
     ("long", 0.1, 5, 1, 0.816666666),  # 500% profit, set stoploss to 10% above open price
+    ("long", 0, 5, 10,  3.3333333),  # 500% profit, set stoploss break even
+    ("long", 0.1, 5, 10,  3.26666666),  # 500% profit, set stoploss to 10% above open price
+    ("long", -0.1, 5, 10,  3.3999999),  # 500% profit, set stoploss to 10% belowopen price
 
     ("short", 0, 0.1, 1, 0.1111111),
     ("short", -0.1, 0.1, 1, 0.2222222),
     ("short", 0.1, 0.2, 1, 0.125),
     ("short", 0.1, 1, 1, 1),
+    ("short", -0.01, 5, 10, 10.01999999),  # 500% profit at 10x
 ])
 def test_stoploss_from_open_leverage(side, rel_stop, curr_profit, leverage, expected):
 
-    stoploss = stoploss_from_open(rel_stop, curr_profit, side == 'short')
+    stoploss = stoploss_from_open(rel_stop, curr_profit, side == 'short', leverage)
     assert pytest.approx(stoploss) == expected
     open_rate = 100
     if stoploss != 1:
         if side == 'long':
-            current_rate = open_rate * (1 + curr_profit)
-            stop = current_rate * (1 - stoploss)
-            assert pytest.approx(stop) == open_rate * (1 + rel_stop)
+            current_rate = open_rate * (1 + curr_profit / leverage)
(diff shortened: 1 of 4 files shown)


tests/strategy/test_strategy_helpers.py
```diff
@@ -177,26 +177,30 @@ def test_stoploss_from_open(side, profitrange):
...
     ("long", 0.1, 5, 1, 0.816666666),  # 500% profit, set stoploss to 10% above open price
+    ("long", 0, 5, 10,  3.3333333),  # 500% profit, set stoploss break even
+    ("long", 0.1, 5, 10,  3.26666666),  # 500% profit, set stoploss to 10% above open price
+    ("long", -0.1, 5, 10,  3.3999999),  # 500% profit, set stoploss to 10% belowopen price
 
     ("short", 0, 0.1, 1, 0.1111111),
     ("short", -0.1, 0.1, 1, 0.2222222),
     ("short", 0.1, 0.2, 1, 0.125),
     ("short", 0.1, 1, 1, 1),
+    ("short", -0.01, 5, 10, 10.01999999),  # 500% profit at 10x
```
xmatthias, line 188: assuming the below sample of an open-rate at 100$ - what's the expectation with `stoploss_from_open(-0.01, curr_profit)` at 10x?

The current implementation will yield 100.1 - which results in 1% "own capital at risk" below the open (1% / 10x).
as this is "stoploss_from_open" - this could also be interpreted as "price distance" - but i feel that would be strange behavior, as all other stops are  leverage adjusted.

Undertoned: I see the code annotation for  `open_relative_stop` argument is `Desired stop loss percentage relative to open price`, so I feel the same that  `stops with leverage adjusted`  is more suitable than `price distance`. And It's can also be consistent with other stops.

Issue #7499, opened 2022-09-29
Support take profit order same as "stoploss_on_exchange" on Binance 


liaogx:
## Describe your environment

 * Operating system: MacOS Monterey 12.5.1
 * Python Version: 3.8.5
 * CCXT version: 1.92.20
 * Freqtrade Version: freqtrade 2022.8 stable

## Describe the problem

In some cases, bots may miss take profit target if prices spike or drop in very short period, 1s or 2s, shorter than default `process_throttle_secs: 5`, this problem could be avoided if `take profit` order could be set as `stoploss_on_exchange` order.

## Describe the enhancement

Binance futures support three kinds of TP and SL orders
- trigger price + limit price

- trigger price + market price

- position TP/SL

Support take profit order same as the `stoploss` order on Binance. 😄

xmatthias:
While i'll mark this as enhancement - it'll probably stay in this state for quite a long time (we have enhancement issues that are 4 years and older ...) - as this will require quite some heavy changes to internal workings - and is currently not something we're actively working on.

asuiu:
The tested workaround can be found here: 
https://github.com/freqtrade/freqtrade/issues/7414

xmatthias:
> The tested workaround can be found here: #7414

Your workaround may be tested in a standalone context - but it's not working in a freqtrade context. 
Your comment there has been hidden as it's missleading freqtrade users down a path that will not work.
(the comment can still be shown - but i hope it's a sign to people to read the subsequent comment, explaining that it'll not work).

Axel-CH:
Quick update on the subject, I saw that CCXT now support TP/SL order. you can see that [here](https://docs.ccxt.com/#/README?id=stoploss-and-takeprofit-orders-attached-to-a-position)
@xmatthias Do you think that could be implemented and supported in Freqtrade?
 What are the key steps/things to looks in order to implement that?
I willing to work on it the next week

xmatthias:
@Axel-CH ccxt was never the blocking point for this, and it's not the case now either.

Depending on the exchange, supporting both a TP/SL order will result in either one (binnace OCO) or 2 orderid's (almost all other exchanges) that need to be tracked.

A first step can be the removal of all "open_order_id" occurances by iterating over a (to define) property of "open orders" for that trade object, using the orders we track in the database.
Implicitly, this would enable multiple orders to be open per trade - but i'd keep that limitation - at least for the moment.

[...]

PR #8296, opened 2023-03-09
feat: Added price_rounding modes in config

asuiu:
## Summary
This PR adds new functionality to configure the price rounding mode for new orders.

## Quick changelog

- adds `exchange.price_rounding_mode` config option
 
## What's new?
Currently the orders are sent by applying a ROUND_UP rounding to the price, which on some systems might be an issue due to the ways the CPython represents the `float` numbers, leading to wrong price roundings and unexpected behavior. The safest rounding type for built-in Python float numbers is the ccst.ROUND mode, which currently can't be changed without changing the current functionality of freqtrade.
I have added the rounding mode as an option to config, and it can set one of 3 available ccxt rounding modes.

The rounding mode defaults to ccxt.ROUND_UP, and it can be changed by setting `exchange.price_rounding_mode` config parameter to the one of the 3 supported by `ccxt` rounding modes: `TRUNCATE, ROUND, ROUND_UP`.

The feature is also covered by unit-tests.

### DIFF
--- a/freqtrade/constants.py
+++ b/freqtrade/constants.py
@@ -5,6 +5,8 @@ bot constants
 """
 from typing import Any, Dict, List, Literal, Tuple
 
+from ccxt import ROUND, ROUND_DOWN, ROUND_UP, TRUNCATE
+
 from freqtrade.enums import CandleType, PriceType, RPCMessageType
 
 
@@ -50,6 +52,8 @@ DEFAULT_DATAFRAME_COLUMNS = ['date', 'open', 'high', 'low', 'close', 'volume']
 # it has wide consequences for stored trades files
 DEFAULT_TRADES_COLUMNS = ['timestamp', 'id', 'type', 'side', 'price', 'amount', 'cost']
 TRADING_MODES = ['spot', 'margin', 'futures']
+PRICE_ROUND_MODES = [TRUNCATE, ROUND, ROUND_UP, ROUND_DOWN]
+DEFAULT_PRICE_ROUND_MODE = ROUND_UP
 MARGIN_MODES = ['cross', 'isolated', '']
 
 LAST_BT_RESULT_FN = '.last_result.json'
@@ -476,7 +480,9 @@ CONF_SCHEMA = {
                 'outdated_offset': {'type': 'integer', 'minimum': 1},
                 'markets_refresh_interval': {'type': 'integer'},
                 'ccxt_config': {'type': 'object'},
-                'ccxt_async_config': {'type': 'object'}
+                'ccxt_async_config': {'type': 'object'},
+                'price_rounding_mode': {'type':    'integer', 'enum': PRICE_ROUND_MODES,
+                                        'default': DEFAULT_PRICE_ROUND_MODE}
             },
             'required': ['name']
         },
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -19,9 +19,9 @@ from ccxt import TICK_SIZE
 from dateutil import parser
 from pandas import DataFrame, concat
 
-from freqtrade.constants import (DEFAULT_AMOUNT_RESERVE_PERCENT, NON_OPEN_EXCHANGE_STATES, BidAsk,
-                                 BuySell, Config, EntryExit, ListPairsWithTimeframes, MakerTaker,
-                                 OBLiteral, PairWithTimeframe)
+from freqtrade.constants import (DEFAULT_AMOUNT_RESERVE_PERCENT, DEFAULT_PRICE_ROUND_MODE,
+                                 NON_OPEN_EXCHANGE_STATES, BidAsk, BuySell, Config, EntryExit,
+                                 ListPairsWithTimeframes, MakerTaker, OBLiteral, PairWithTimeframe)
 from freqtrade.data.converter import clean_ohlcv_dataframe, ohlcv_to_dataframe, trades_dict_to_list
 from freqtrade.enums import OPTIMIZE_MODES, CandleType, MarginMode, TradingMode
 from freqtrade.enums.pricetype import PriceType
@@ -143,7 +143,9 @@ class Exchange:
             if config.get('margin_mode')
             else MarginMode.NONE
         )
-        self.liquidation_buffer = config.get('liquidation_buffer', 0.05)
+        self.liquidation_buffer = config.get("liquidation_buffer", 0.05)
+        self._price_rounding_mode = exchange_config.get("price_rounding_mode",
+                                                        DEFAULT_PRICE_ROUND_MODE)
 
         # Deep merge ft_has with default ft_has options
         self._ft_has = deep_merge_dicts(self._ft_has, deepcopy(self._ft_has_default))
@@ -732,10 +734,11 @@ class Exchange:
 
     def price_to_precision(self, pair: str, price: float) -> float:
         """
-        Returns the price rounded up to the precision the Exchange accepts.
-        Rounds up
+        Returns the price rounded to the precision the Exchange accepts.
+        The default price_rounding_mode in conf is ROUND_UP
         """
-        return price_to_precision(price, self.get_precision_price(pair), self.precisionMode)
+        return price_to_precision(price, self.get_precision_price(pair),
+                                  self.precisionMode, self._price_rounding_mode)
 
     def price_get_one_pip(self, pair: str, price: float) -> float:
         """
--- a/freqtrade/exchange/exchange_utils.py
+++ b/freqtrade/exchange/exchange_utils.py
@@ -2,11 +2,12 @@
 Exchange support utils
 """
 from datetime import datetime, timedelta, timezone
-from math import ceil
+from math import ceil, floor
 from typing import Any, Dict, List, Optional, Tuple
 
 import ccxt
-from ccxt import ROUND_DOWN, ROUND_UP, TICK_SIZE, TRUNCATE, decimal_to_precision
+from ccxt import (DECIMAL_PLACES, ROUND, ROUND_DOWN, ROUND_UP, SIGNIFICANT_DIGITS, TICK_SIZE,
+                  TRUNCATE, decimal_to_precision)
 
 from freqtrade.exchange.common import BAD_EXCHANGES, EXCHANGE_HAS_OPTIONAL, EXCHANGE_HAS_REQUIRED
 from freqtrade.util import FtPrecise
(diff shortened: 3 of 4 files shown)


asuiu:
@xmatthias 
as a side note, I'd propose to change the default rounding mode from `ROUND_UP` to `ROUND`. It's pretty dangerous by adding a risk of placing the orders with the wrong prices, and I can't imagine any general valid use-case where it might be useful, although there might be some that I miss. Anyway if there's a good purpose for it - it would be great to add those into the comments of the [`freqtrade.exchange.exchange.Exchange.price_to_precision`](https://github.com/freqtrade/freqtrade/blob/29dfb5c169a062e52c38b6304f8482a02b6ad4cb/freqtrade/exchange/exchange.py#L733) method.

freqtrade/exchange/exchange.py
```diff
@@ -732,10 +734,11 @@ def amount_to_precision(self, pair: str, amount: float) -> float:
...
     def price_to_precision(self, pair: str, price: float) -> float:
         """
-        Returns the price rounded up to the precision the Exchange accepts.
-        Rounds up
+        Returns the price rounded to the precision the Exchange accepts.
+        The default price_rounding_mode in conf is ROUND_UP
         """
-        return price_to_precision(price, self.get_precision_price(pair), self.precisionMode)
+        return price_to_precision(price, self.get_precision_price(pair),
+                                  self.precisionMode, self._price_rounding_mode)
```
xmatthias, line 741: this won't work reliably throughout the bot.

`exchange.price_to_precision()` is only used from within the exchange class (it's a local helper, just as the other methods around here).

neither from trade-model, nor from backtesting we'll call this (or from trade_model, we won't be able to call this, either) - but we call the actual method instead.

As is - it's therefore incomplete - and will only apply to to probably less than half of the actually used locations.

freqtrade/exchange/exchange_utils.py
```diff
@@ -219,35 +220,48 @@ def amount_to_contract_precision(
...
     :return: price rounded up to the precision the Exchange accepts
-
     """
     if price_precision is not None and precisionMode is not None:
-        # price = float(decimal_to_precision(price, rounding_mode=ROUND,
-        #                                    precision=price_precision,
-        #                                    counting_mode=self.precisionMode,
-        #                                    ))
         if precisionMode == TICK_SIZE:
+            if rounding_mode == ROUND:
```
xmatthias, line 244: using Round (especially with tick-size exchanges) is pretty dangerous - as it doesn't really define the direction of the rounding.

freqtrade/constants.py
```diff
@@ -476,7 +480,9 @@
                 'outdated_offset': {'type': 'integer', 'minimum': 1},
                 'markets_refresh_interval': {'type': 'integer'},
                 'ccxt_config': {'type': 'object'},
-                'ccxt_async_config': {'type': 'object'}
+                'ccxt_async_config': {'type': 'object'},
+                'price_rounding_mode': {'type':    'integer', 'enum': PRICE_ROUND_MODES,
```
xmatthias, line 484: i don't think that making this configurable is a good idea.
You'll get into the "how to" discussion very quickly.

maybe i'd like to round down on buys, and round up on sells ... - or "up on entries" "down on exits" (which is different from the prior in case of shorts).

xmatthias (changes requested):
which scenario are you encountering problems due to this?

Assuming a standard configuration (strategy does NOT modify the price) - then the only place where this rounding would actually apply is  stoploss - where rounding up is correct (in case of long - see discussion in #2763) - but for shorts, probably rounding down would be better.
All other prices would come directly to the orderbook - which is already in "exchange precision".

---

as is - it's incompletely applying the configuration - and therefore clearly wrong (but before fixing this - we should determine if configuration is the right approach).

---

Making this configurable is probably a pretty bad idea IMHO.
We use the method "standalone" (not through the exchange class) in several places - some of which don't have config available, neither the exchange.
The only way around this without passing config to places it doesn't belong would be to set / update a magic singleton you then use within the method - but that's not really a good approach.

---

To me, the correct rounding (up / down) will depend on the application (stoploss long/short, entry/exit) - and should NOT be configurable by the user (at least not for stops).
I'm not claiming that it's correct as is (especially when looking at shorts) - but a switch changing the direction application-wide is wrong, too.

freqtrade/exchange/exchange_utils.py
asuiu: Can you please elaborate more on this? Maybe I miss some information on properties of "tick-size" exchanges or how it's used?

freqtrade/constants.py
asuiu: I agree, I also wouldn't add it as a configuration, and would change the default rounding to ROUND, instead of ROUND_UP, to avoid redundant configuration and discussions.

Do you think that it's okay to change the default rounding from ROUND_UP to ROUND?

I think this would simplify this PR a lot; with the exception of the tests that would be broken by this change :)

asuiu:
@xmatthias 

> which scenario are you encountering problems due to this?

I'm implementing the feature to offer possibility to the strategy add TakeProfit and StopLoss protections for open positions, where price and moment of placing of these orders would be triggered and defined by strategy. This functionality seems to be described in this Issue: https://github.com/freqtrade/freqtrade/issues/7499 although author is not clear on how exactly these orders would be defined - by strategy - or by config.

> Assuming a standard configuration (strategy does NOT modify the price) - then the only place where this rounding would actually apply is stoploss - where rounding up is correct (in case of long - see discussion in https://github.com/freqtrade/freqtrade/pull/2763) - but for shorts, probably rounding down would be better.

Yes, as you've mentioned - this is inconsistent behavior for Long and Short positions, so in order to correctly place the SL you have to trick the percentages in the config somehow - I don't know how yet.
So making it consistent for Futures probably would be a better choice - from my perspective; Ultimately, as you mention - the rounding up for SL on Long is correct, which implies that rounding up on SL for Short - is wrong behavior.

> as is - it's incompletely applying the configuration - and therefore clearly wrong 

I'm new to freqtrade and most probably miss lots of things and architecture details, so would be grateful if you could advise a better option/idea.

> (but before fixing this - we should determine if configuration is the right approach).

I didn't find a better option to make it work, and with not lots of code changes across the freqtrade, so would be thankful if you can come with some alternatives.

asuiu:
> To me, the correct rounding (up / down) will depend on the application (stoploss long/short, entry/exit) - and should NOT be configurable by the user (at least not for stops).
I'm not claiming that it's correct as is (especially when looking at shorts) - but a switch changing the direction application-wide is wrong, too.

@xmatthias, I agree with these, and my initial proposal of using the normal ROUND would fix this dilemma.  Are you ok with changing the default rounding mode?

freqtrade/exchange/exchange_utils.py
xmatthias: TICK_SIZE is a rare method - fortunately used by very few exchanges.

You could for example get tick value of 0.5 - which would mean only multiples of 0.5 are valid prices (100, 100.5, 101, ...) - but not 100.1, 100.2, ...). 

As a consequence - using "round" of 100.2 (or 100.4) would ALWAYS round down to 100, while 100.7 (or 100.55) would always round to 101 ... which is however incorrect depending on the usage (which is why I'm in general not a fan of using ROUND).

Depending on the  price of the pair (VERY low priced pairs) - tick-size can also be 1 or 5 - or 2 ... so 2,4,6, ... are valid - but 5,7 is not (obviously decimals would also not be allowed in this mode) - which makes it even less predictable / worse - as the actual difference increases.

xmatthias:
> I agree with these, and my initial proposal of using the normal ROUND would fix this dilemma. Are you ok with changing the default rounding mode?

I don't see / agree with this (at least not entirely). 
Simply using ROUND by itself will make this problem worse.

ROUND will not take the order direction (buy/sell | entry/exit) into account - and is therefore in my opinion in general the wrong behavior  (it's 100% wrong for stop orders).

the default for the function may be changed to ROUND (default applies to regular buy/sell orders) - even if I'm inclined to say that it should still be direction-dependent (as well as intend dependent - but that may go too far).
For a limit buy order, i may want the price to be rounded down (if my intend is for price to retrace to my price).
If my intend is however to "fill immediately" - then "round_up" would be the better choice (i accept an eventually marketable limit order to get filled immediately - at that defined price or better).

The same goes for sell orders - but in reverse. 
Now that's probably going too far - as this would require quite a lot of interpretation into a simple, single price - but is the reason why i think that making this configurable is a bad idea.

---

For everything stop-related, we'll need to explicitly specify ROUND_UP or ROUND_DOWN (depending on the trade direction) to avoid problems with order-placement and increased risk (especially on VERY low priced pairs, 1 tick can be several %).
That's currently not the case - and I'm clearly open for improvements in that regard. That's been forgotten / overlooked with the introduction of shorts - but so far has also not been causing problems (at least not as far as i can see).

For non-stop orders (so regular orders) - i guess `ROUND` could be used - but it's not a simple "change the default" - as that would break the logic for stops.

> as is - it's incompletely applying the configuration - and therefore clearly wrong

simply search for the method name (with a leading space, excluding tests) and you'll see what i mean. 
The "configured setting" would only be used in some instances, but not in others (for example, it'd be ignored for stop prices) - as these calls don't go through the exchange class.

---

regarding #7499, we should discuss this separately. There's a few (bigger) things that need to happen before we can even start to look at that.

freqtrade/exchange/exchange_utils.py
asuiu: Oh, so if I understand right, it's dangerous when the TICK_SIZE is not a power of 10 due to its complexity for an average human brain as its not used to the "non-decimal" rounding, and it would be slightly more complex for a human to compute the normal rounding vs a ROUND_UP/ROUND_DOWN, and this is the source of danger.

I don't agree that a slight increase in complexity would be a valid reason to implement a functionality in a semantically wrong way, as this would lead to even more confusion and would make system even more complex and prone to unexpected behavior.

asuiu:
> regarding https://github.com/freqtrade/freqtrade/issues/7499, we should discuss this separately. There's a few (bigger) things that need to happen before we can even start to look at that.

Can you point more exactly what are the things that need to happen before that?

I've already implemented it on my own fork, so far just for the real-time trading, but I've tested it with Binance, and a freqtrade strategy, and it seems to work fine with Binance and the rest of freqtrade components.

xmatthias:
Binance is a special case with OCO orders. While we can support binance, a implementation here will have to be scalable to other exchanges. most other exchanges implement this in a different way (2 separate, independent orders) - so handling for this will be different. 

basically, the whole concept of "open_order_id" must be removed, and something around "handle all open orders" will need to be used.

It's however not related to this pr at all (so this is clearly the wrong place to discuss this) - and i'll therefore be marking this, as well as the prior post as off-topic.

freqtrade/exchange/exchange_utils.py
xmatthias: it's not the complexity that makes it difficult - but the oddity of using round- which can lead to huge steps (with round, in unexpected directions - as it'll NOT be towards the nearest next point necessarily).
The human brain has little to do with this - as a regular user will not really come to touch this logic.

independent of tick-size - using ROUND is wrong for most cases (see the other comments).

asuiu:
> For everything stop-related, we'll need to explicitly specify ROUND_UP or ROUND_DOWN (depending on the trade direction) to avoid problems with order-placement and increased risk (especially on VERY low priced pairs, 1 tick can be several %).

@xmatthias 
I understand that the above approach specifies ROUND_UP or ROUND_DOWN depending on the trade direction to avoid problems with order placement and increased risk. However, I wanted to bring to your attention that there are profitable strategies that require precise definition of the StopLoss price. Especially for VERY low priced pairs, rounding up long StopLoss orders could be 100% wrong and harmful for those strategies.
 
Luckily, this is just an optional config setting - and since by default the current behavior won't be changed and won't affect currently working strategies, we could extend the freqtrade capabilities to be used by a larger pool of users and companies, and benefit from contributions from them, like is this PR.

So, assuming that the freqtrade's intention is to be used by as many users as possible, in order to speed up this PR's merge, I'd propose to let the users to decide what's the best way to use the protection orders and introduce this feature as optional setting, as it's done in this PR;
Ultimately, it's clear that this setting is required by at least one user/company, so there's a good chance it might be required by others;

Since we didn't find a better idea than to make it an optional config setting, @xmatthias can you please confirm that there are no other blockers for this feature and advise what other changes I'd need to perform in order to align it with the rest of the functionalities.

@xmatthias, thanks again for your time and explanations!

xmatthias:
> Since we didn't find a better idea than to make it an optional config setting, @xmatthias can you please confirm that there are no other blockers for this feature and advise what other changes I'd need to perform in order to align it with the rest of the functionalities.

I disagree and am not sure  where you read this. That's absolutely not what i said above.

Maybe it wasn't fully clear from the above - so letme repeat my intention above and be very clear so we don't move in the wrong direction here:
we will **not** be adding this as optional configuration, as explained in detail in https://github.com/freqtrade/freqtrade/pull/8296#issuecomment-1461759769. 

First of all, because your changes do exactly not do what you actually want (the only place where the rounding is not changed is for stoplosses).
But more importantly, because a simple, plain configuration setting will NOT satisfy the needs we have.

The way i currently see this is that the correct approach is (this is a great PR apart from the configurable things, don't get me wrong):
* Fix stoploss calculations by using `ROUND_DOWN` for short stoplosses (long  stoplosses are correct as is). This will not be optional, i'm happy to get this updated in this PR, or separately (otherwise i can implement this myself at a later point, and would surely reuse many parts of this PR).
In my opinion, no other approach for stoplosses is acceptable, and everything else must be considered a bug.
* either leave regular, non-stop orders as is (always `ROUND_UP` - which is potentially wrong), or change them to `ROUND` (i don't have a strong opinion on this, and i don't think it really matters much). Obviously for this, the default approach can be changed to `ROUND` - but for all stoploss usages - we will supply an explicit value, which is either `ROUND_UP` or `ROUND_DOWN` - depending on the trade direction.

I don't really see a point in having a configuration for this - and IF there really is one (though i don't really see a point in it), then it will definitely **not apply to stoploss**.

---

To explain my reason for the strong opinion on stops:
With a very tight stoploss (which some people seem to find due to too much backtesting), it's already quite possible to get "stoploss would trigger immediately" warnings. By allowing the user to round in the wrong direction (based on your very crude configurations, also without ANY chance for users to actually use it correctly in different scenarios), this problem would become much worse (especially on real shitcoins, with VERY low prices and ticksizes > 1%). 

---

> Luckily, this is just an optional config setting - and since by default the current behavior won't be changed and won't affect currently working strategies, we could extend the freqtrade capabilities to be used by a larger pool of users and companies, and benefit from contributions from them, like is this PR.

I'm more than happy to get contributions and more users - but that should not introduce bugs (or allow the user to configure the bot to have bugs).

---

If you think different rounding is necessary - then the need is wrong.
with `custom_stoploss` (assuming you're using this) - you can very precisely control the stop. with a KNOWN rounding method (up for long stops, down for short stops) - you can very well control this, and get to the precise price you need - if that's REALLY what you need (which i doubt as the stop should control the risk you're willing to take, not the precise price of the stop - but is a different topic entirely).

---

> Ultimately, it's clear that this setting is required by at least one user/company, so there's a good chance it might be required by others;

if this were applied to stops - then it would introduce buggy behavior. While i'm happy to get contributions - for every PR, i'll also have to think _"by accepting this, i accept responsibility to support and maintain this for ALL users going forward"_. 
As such, i will not be accepting potentially problematic behavioral changes unless there's a real, non-neglectible benefit - which honestly, i don't see in this case.

xmatthias:
Closing in favor of #8386 - since this PR does not allow pushes from maintainers (That PR does build on top of this PR directly though, as most parts were great (except for the configuration part)).

It's implementing the changes as commented above - using ROUND for regular orders, and ROUND_UP / ROUND_DOWN (respectively) for stop orders, depending on their direction.

PR #8297, opened 2023-03-09
Feat/add pytorch model support

Yinon-Polak:
## Summary
This update includes changes to the loading/saving methods of DataDrawer and IFreqaiModel,
and the addition of new classes for a PyTorch model trainer and a PyTorch multi-class classifier.

## What's new?
- The PyTorchModelTrainer class has been added to contain the general training loop logic, loading, and saving methods.
- The PyTorchClassifierMultiTarget class has been added, which currently uses a simple MLP PyTorch model serving as an example.


freqtrade/freqai/base_models/PyTorchModelTrainer.py
```diff
@@ -0,0 +1,184 @@
...
+    ) -> Dict[str, DataLoader]:
+        """
+        Converts the input data to PyTorch tensors using a data loader.
+        """
+        data_loader_dictionary = {}
+        for split in ["train", "test"]:
+            labels_shape = data_dictionary[f"{split}_labels"].shape
+            labels_view = labels_shape[0] if labels_shape[1] == 1 else labels_shape
+            dataset = TensorDataset(
+                torch.from_numpy(data_dictionary[f"{split}_features"].values).float(),
```
robcaulk: Since we are starting with a dataframe, it is preferable to use `torch.tensor()` as opposed to `from_numpy()`.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
```diff
@@ -0,0 +1,184 @@
...
+        self.model_meta_data = model_meta_data
+        self.device = device
+        self.max_iters = max_iters
+        self.batch_size = batch_size
+        self.eval_iters = eval_iters
+
+        if init_model:
+            self.load_from_checkpoint(init_model)
+
+    def fit(self, data_dictionary: Dict[str, pd.DataFrame]):
```
robcaulk: It seems to me you are doing the training loss twice unnecessarily. See comments below.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
```diff
@@ -0,0 +1,184 @@
...
+
+    @torch.no_grad()
+    def estimate_loss(
+            self,
+            data_loader_dictionary: Dict[str, DataLoader],
+            data_dictionary: Dict[str, pd.DataFrame]
+    ) -> Dict[str, float]:
+
+        self.model.eval()
+        epochs = self.calc_n_epochs(
```
robcaulk: This should be moved to *before* the training loop - its value should not be changing inside the training loop so it is unnecessary cycles.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
```diff
@@ -0,0 +1,184 @@
...
+            n_obs=len(data_dictionary["test_features"]),
+            batch_size=self.batch_size,
+            n_iters=self.eval_iters
+        )
+        loss_dictionary = {}
+        for split in ["train", "test"]:
```
robcaulk: Unless I am missing something, you appear to compute loss on the entire train dataset and then again, in the subsequent loop (line 76). These values will be the same wont they? So it appears to be wasted cycles. 

Seems to me you only need to do `test` here.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
```diff
@@ -0,0 +1,184 @@
...
+        )
+        for epoch in range(epochs):
+            # evaluation
+            losses = self.estimate_loss(data_loaders_dictionary, data_dictionary)
+            logger.info(
+                f"epoch ({epoch}/{epochs}):"
+                f" train loss {losses['train']:.4f} ; test loss {losses['test']:.4f}"
+            )
+            # training
+            for batch_data in data_loaders_dictionary["train"]:
```
robcaulk: Isn't `data_loaders_dictionary` filled with `dataloader` objects? Does this work to treat `dataloader` objects directly as batch data? 

It would be more classical to enumerate here on the train dataloader (which is precisely how you did it below in `estimate_loss()`).

freqtrade/freqai/prediction_models/PyTorchClassifierMultiTarget.py
```diff
@@ -0,0 +1,162 @@
...
+        int: The maximum number of iterations to run during training.
+        int: The batch size to use during training.
+        float: The learning rate to use during training.
+        int: The number of training iterations between each evaluation.
+        dict: A dictionary mapping class names to their corresponding indices.
+        dict: A dictionary mapping indices to their corresponding class names.
+        """
+
+        super().__init__(**kwargs)
+        model_training_parameters = self.freqai_info["model_training_parameters"]
```
robcaulk: `model_training_parameters` is reserved for external library `kwargs` that pass to the model object initializer directly. Howerver, I guess we can leave this as is since you are creating your own model object here (your `PyTorchMLPModel`). 

One thing I would ask is that you please type these attributes e.g.:

`self.learning_rate: float = model_training_parameters.get("learning_rate", 3e-4)`

freqtrade/freqai/freqai_interface.py
```diff
@@ -563,8 +563,11 @@ def model_exists(self, dk: FreqaiDataKitchen) -> bool:
             file_type = ".joblib"
         elif self.dd.model_type == 'keras':
             file_type = ".h5"
-        elif 'stable_baselines' in self.dd.model_type or 'sb3_contrib' == self.dd.model_type:
+        elif ('stable_baselines' in self.dd.model_type or
+              'sb3_contrib' == self.dd.model_type or
+              'pytorch' == self.dd.model_type):
```
robcaulk:
```suggestion
        elif (self.dd.model_type in ["stable_baselines", "sb3_contrib", "pytorch"]):
```

freqtrade/freqai/data_drawer.py
```diff
@@ -446,7 +446,9 @@ def save_data(self, model: Any, coin: str, dk: FreqaiDataKitchen) -> None:
             dump(model, save_path / f"{dk.model_filename}_model.joblib")
         elif self.model_type == 'keras':
             model.save(save_path / f"{dk.model_filename}_model.h5")
-        elif 'stable_baselines' in self.model_type or 'sb3_contrib' == self.model_type:
+        elif ('stable_baselines' in self.model_type or
+              'sb3_contrib' == self.model_type or
+              'pytorch' == self.model_type):
```
robcaulk:
```suggestion
        elif (self.model_type in ["stable_baselines", "sb3_contrib", "pytorch"]):
```

freqtrade/freqai/prediction_models/PyTorchMLPModel.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class PyTorchMLPModel(nn.Module):
+    def __init__(self, input_dim: int, hidden_dim: int, output_dim: int):
+        super(PyTorchMLPModel, self).__init__()
+        self.input_layer = nn.Linear(input_dim, hidden_dim)
+        self.hidden_layer = nn.Linear(hidden_dim, hidden_dim)
+        self.output_layer = nn.Linear(hidden_dim, output_dim)
+        self.relu = nn.ReLU()
+        self.dropout = nn.Dropout(p=0.2)
```
robcaulk: Why not make `p` another kwarg for people to play with? 

Could go even further and parameterize the number of layers using a loop in `forward`. Just a suggestion, you don't need to do it if you prefer not to.

robcaulk (changes requested):
Thanks for this, it is a great contribution that the community will surely appreciate.

I have mostly minor comments. However, my main concern is surrounding the duplicated train loss calculation. Unless I miss something, this should be corrected. 

Finally - are you comfortable doing the tests? I can help you with those if you need, but I want to give you the opportunity to do them and learn if you wish. Just let me know.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
Yinon-Polak: from_numpy() automatically inherits input array dtype. On the other hand, torch.Tensor is an alias for torch.FloatTensor.

Therefore, if you pass int64 array to torch.Tensor, output tensor is float tensor and they wouldn't share the storage. torch.from_numpy gives you torch.LongTensor as expected.

https://stackoverflow.com/questions/48482787/pytorch-memory-model-torch-from-numpy-vs-torch-tensor

In PyTorchClassifierMultiTarget we convert the label column type from str to int. So i think it's better this way, you can control the data types outside PytorchModelTrainer.

robcaulk: But `from_numpy` requires a numpy array. It is great for numpy arrays because it avoids making a copy. The point about "sharing storage" in the link you sent, only applies if you are starting from a numpy array. 

However, we are not starting with a numpy array, we are starting with a pandas dataframe, so it will convert from dataframe.values into a numpy array, and then from numpy array to torch tensor. 

`torch.tensor` skips the creation of the numpy array. 

But I have not timed it, and I doubt it is incredibly impactful since it is once per train. So do as you wish :-)

Yinon-Polak: I don't think there's a way to convert directly from DataFrame to pytorch tensor.

freqtrade/freqai/prediction_models/PyTorchMLPModel.py
Yinon-Polak: yes, that should be added.
i think something like the baselines3 convention should be used here.

```
freqai: { 
     ...,
     "trainer_kwargs": {
         "max_iters":  100
         "batch_size": 64
         "learning_rate": 3e-4
         "model_kwargs": {
             "n_hidden":  1024,
             "drop_out": 0.2,
        }, 
     },
}
```

ignoring "model_training_parameters" completely.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
Yinon-Polak: inside the training loop the model parameters are being updated every batch iteration.
so, the loss there is'nt representing the loss on the entire data.

i'm updating the eval iteration to go through all the train/test data once.
also, adding max_n_eval_batches for the case that the train data is very large, and the user wants just a sample of few batches to estimate the loss.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
Yinon-Polak: the train and eval are equivalent. 
the only difference is that in the eval i do it on both splits versos the train where i explictly only take train.

freqtrade/freqai/prediction_models/PyTorchClassifierMultiTarget.py
Yinon-Polak: will add types.

I expand on the usage of model_training_parameters in the last comment. we can use something else just for the clarification.

freqtrade/freqai/data_drawer.py
Yinon-Polak: this statements are not equivalent.
'stable_baselines' in self.model_type vs self.model_type in ["stable_baselines", "sb3_contrib", "pytorch"].
pushing this change for "sb3_contrib", "pytorch" only.

freqtrade/freqai/freqai_interface.py
Yinon-Polak: this statements are not equivalent.
'stable_baselines' in self.dd.model_type vs self.dd.model_type in ["stable_baselines", "sb3_contrib", "pytorch"].
pushing this change for "sb3_contrib", "pytorch" only.

Yinon-Polak:
@robcaulk I can write the tests, is there a reference for model testing ?

freqtrade/freqai/prediction_models/PyTorchMLPModel.py
robcaulk: Using a standard convention is good, but why would we ignore `model_training_parameters`? It fulfills the purpose and it is well known in FreqAI software for serving such purpose. Adding a new dictionary that does the same thing as an existing one is not easy to support.

freqtrade/freqai/freqai_interface.py
robcaulk: Please keep the suggested change, but you are correct that we simply need to add a "3" to the end of `stable_baselines3`. Please follow same convention as suggested in the DataDrawer.

freqtrade/freqai/data_drawer.py
robcaulk: Please keep the suggested change, but you are correct that we simply need to add a "3" to the end of stable_baselines3

freqtrade/freqai/prediction_models/PyTorchClassifierMultiTarget.py
robcaulk: As I mention, I think it is best to leave this as is since you are creating your own model object. It works in this case.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
robcaulk: Technically you enumerate on the dataloader object below, and you do not explicitly enumerate on the object here. 

As you mention, you do the same thing in both places, so for consistency and readability it is preferred to use the same syntax in both places to convey this point. Then I would lobby for explicitly enumerating on the dataloader here.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
robcaulk: Ok I see. In that case, what is the use of that train loss in comparison to simply reporting the  sum of train loss encountered during the most recent epoch?

I agree that the test loss important to do in this fashion - I am more concerned about the value of this train loss. Let me know how you would use that value and how it would help you more than just using epoch train loss summation and validation loss reporting.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
robcaulk: There is not. There are only ways to avoid creating numpy arrays, but it is not impactful since it is once per train - either way is fine.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
Yinon-Polak: i see what you mean, i'll add it.

freqtrade/freqai/base_models/PyTorchModelTrainer.py
Yinon-Polak: just for comparing the train to test loss. no other uses.
the difference between what you suggest to what i did will only be relevant in big datasets.
we can use the last epoch.

freqtrade/freqai/prediction_models/PyTorchMLPModel.py
Yinon-Polak: it makes it more clear what are the trainer args and what are the model args.
but i see what you mean, i'll keep `model_training_parameters`.

tests/freqai/test_freqai_interface.py
```diff
@@ -85,6 +86,21 @@ def test_extract_data_and_train_model_Standard(mocker, freqai_conf, model, pca,
     if 'test_3ac' in model or 'test_4ac' in model:
         freqai_conf["freqaimodel_path"] = str(Path(__file__).parents[1] / "freqai" / "test_models")
 
+    if 'PyTorchClassifierMultiTarget' in model:
+        model_save_ext = 'zip'
+        freqai_conf['freqai']['model_training_parameters'].update({
+            "n_hidden": 1024,
```
robcaulk: Looks good, but one point is that we need these tests to be super short (freqai already takes a disproportionate time to test), so we should bring these parameters down to their bare minimum. So maybe 64 hidden_dim, batch_size 4? Iters 5?

tests/freqai/test_freqai_interface.py
```diff
@@ -52,7 +52,8 @@ def can_run_model(model: str) -> None:
     ('ReinforcementLearner_multiproc', False, False, False, True, False, 0),
     ('ReinforcementLearner_test_3ac', False, False, False, False, False, 0),
     ('ReinforcementLearner_test_3ac', False, False, False, True, False, 0),
-    ('ReinforcementLearner_test_4ac', False, False, False, True, False, 0)
+    ('ReinforcementLearner_test_4ac', False, False, False, True, False, 0),
+    ('PyTorchClassifierMultiTarget', False, False, False, True, False, 0)
```
robcaulk: Does the target assignment work the same as the other classifiers we have? If so, you can move this to the function below with the other classifiers where it loads a classifier test strat. 

If it doesnt use the same target assignment method as the other classifiers, is it because of the encoding? If we can't get around the encoding, then we would need its own test strat (you could copy and paste the existing classifier test strat and add the encoding to the target setting.)

freqtrade/freqai/prediction_models/PyTorchMLPModel.py
```diff
@@ -0,0 +1,53 @@
...
+from torch import Tensor
+
+
+logger = logging.getLogger(__name__)
+
+
+class PyTorchMLPModel(nn.Module):
+    def __init__(self, input_dim: int, output_dim: int, **kwargs):
+        super(PyTorchMLPModel, self).__init__()
+        hidden_dim: int = kwargs.get("hidden_dim", 1024)
```
robcaulk: Now the fun part, are you able to help document these parameters for users? If not, I can help you with it. But I want to let you do it if you'd like to practice doing so.

In either case, we should consider where to place the documentation for the PyTorch stuff. Ideally we would have a PyTorch section which shows users how to select different models which parameters are available for each model. Also, the doc should be written in a way that would encompass future buildouts (like adding a PyTorchRegressor).

freqtrade/freqai/prediction_models/PyTorchClassifierMultiTarget.py
```diff
@@ -0,0 +1,164 @@
...
+        :raises ValueError: If self.class_names is not defined in the parent class.
+
+        """
+
+        if not hasattr(self, "class_names"):
+            raise ValueError(
+                "Missing attribute: self.class_names "
+                "set self.freqai.class_names = [\"class a\", \"class b\", \"class c\"] "
+                "inside IStrategy.set_freqai_targets method."
+            )
```
robcaulk: This is a nice error, thanks for adding that. 

I guess this will need to be added to a new classifier test strat (I elaborated on this in the other comment). But it will also need a bit of documentation for users to understand. I can help with documentation if you would like.

robcaulk (commented):
Thanks, getting close. I was wondering if we should move `PyTorchMLPModel` to a new folder `freqai/pytorch/models` and do a string loading for the PyTorchMLPModel using `import_module` like we do already with the stablebaselines3 stuff in RL. 

This would open the possibilities for users to simply define it in their freqai config, and we could (in the future) provide a selection of model types for users to play with.

Additionally, I would like to use the same machinery for adding a second trainer in the future (not necessarily with this PR).

Yinon-Polak:
> Thanks, getting close. I was wondering if we should move `PyTorchMLPModel` to a new folder `freqai/pytorch/models` and do a string loading for the PyTorchMLPModel using `import_module` like we do already with the stablebaselines3 stuff in RL.
> 
> This would open the possibilities for users to simply define it in their freqai config, and we could (in the future) provide a selection of model types for users to play with.
> 
> Additionally, I would like to use the same machinery for adding a second trainer in the future (not necessarily with this PR).

it think that before we move to importing the model by string we should generalize the trainer. 
so we could import it as string as well. 

the main thing i think is still missing is moving all the moving parts downward.
as i see it the order of classes the users should modify is in increasing order (1. most won't change, 4. most will change):
1. BasePytTorchModel.
2. PyTorchClassifierMultiTarget / PyTorchRegressor (WIP).
3. PytorchModelTrainer.
4. PytorchModel.

currently the initialization of the optimizer class & loss class in the PyTorchClassifierMultiTarget / PyTorchRegressor, is suboptimal as people would probably would like to change those very often.

robcaulk:
> > Thanks, getting close. I was wondering if we should move `PyTorchMLPModel` to a new folder `freqai/pytorch/models` and do a string loading for the PyTorchMLPModel using `import_module` like we do already with the stablebaselines3 stuff in RL.
> > This would open the possibilities for users to simply define it in their freqai config, and we could (in the future) provide a selection of model types for users to play with.
> > Additionally, I would like to use the same machinery for adding a second trainer in the future (not necessarily with this PR).
> 
> it think that before we move to importing the model by string we should generalize the trainer. so we could import it as string as well.
> 
> the main thing i think is still missing is moving all the moving parts downward. as i see it the order of classes the users should modify is in increasing order (1. most won't change, 4. most will change):
> 
>     1. BasePytTorchModel.
> 
>     2. PyTorchClassifierMultiTarget / PyTorchRegressor (WIP).
> 
>     3. PytorchModelTrainer.
> 
>     4. PytorchModel.
> 
> 
> currently the initialization of the optimizer class & loss class in the PyTorchClassifierMultiTarget / PyTorchRegressor, is suboptimal as people would probably would like to change those very often.

I agree with your order of "modifiablility." These additional objects, the `PytorchModel` and the `PytorchModelTrainer`, are in many ways analogous to our `MyRLEnv` class in the `ReinforcementLearner`. For example, most people dont need or want to touch the `ReinforcementLearner` (i.e. `PyTorchClassifierMultiTarget` in this PR), but they do want to modify pieces of the Environment such as the `calculate_reward`. The "pieces" in the present PR are the `PytorchModelTrainer` and the `PytorchModel`. 

We handle this in the RL module by adding a subclass to the `ReinforcementLearner` called the `class MyRLEnv` [details here](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/freqai/RL/BaseReinforcementLearningModel.py#L364) which is a class that inherits/controls the entire environment (so `calculate_reward` is in there). 

For the present PR, we could follow a similar pattern, and add a `MyTrainer` subclass and a `MyTorchModel` subclass inside the `PytorchClassifierMultiTarget`. The appeal of this method means that users can simply inherit their custom class as follows:

```python
class MyCoolTorchClassifier(PytorchClassifierMultiTarget):

    class MyTrainer(PytorchAutoRegressiveTrainer):
        """
        Class that modifies our autoregressive trainer template
        """
        def method_override():
        
    class MyTorchModel(PytorchModel | torch.nn.module ):
        """
        Class that defines or tweaks the network architecture
        """
        
```

Now the `MyCoolPytorchClassifier` class knows it has methods for `MyTorchModel` and a `MyTrainer` which it then uses for initializing its model and trainer inside the `BasePytorchModel`. Example usage in RL [here](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/freqai/RL/BaseReinforcementLearningModel.py#L152).

The main benefit of this structure is that the user still defines a single custom  `--freqaimodel MyCoolTorchClassifier` which contains all their customizations and inherits the entire `IFreqaiModel` object. This is analogous to how the `IStrategy` is inherited for users when they make their custom strategies with `--strategy MyCoolStrategy`.

The downside of this structure is that it can be a little confusing for super new people to realize that they would need to define these subclasses inside their custom class. And they need to realize that these subclasses are overriding existing subclasses, so the subclass names `MyTorchModel` and `MyTrainer` cannot change (just like overriding any other method in Python).

Thoughts? @xmatthias would also like to get your opinion on the matter if you have time.

freqtrade/freqai/prediction_models/PyTorchClassifierMultiTarget.py
Yinon-Polak: yeah, definitely needs a separate test.
also, is the proper place for the additional documentation is the in main PyTorchClassifierMultiTarget documentaion:
```python
class PyTorchClassifierMultiTarget(BasePyTorchModel):
    """
    A PyTorch implementation of a multi-target classifier.
    Important: the user must set `class_names` atrribute in the strategy class inside `set_freqai_targets` method.
     we use class_names here to:
      - initialize the model `output_dim` param
      - create a class mapping form str to int, as pytorch dos'nt support str tensors.
    """
```

robcaulk: While the docstrings are important for code readability. Users use our full documentation available online https://www.freqtrade.io/en/stable/freqai/. 

These files are modifiable in `docs/freqai*.md` 

The torch parameters would go into the existing parameter table. But we need to add another section (maybe at the end of Configuration section?) that describes the pieces of the PyTorch module. This would certainly include things like the "need to add the `self.freqai.class_names` in the `set_freqai_targets()` function." Or "How to create a custom `PyTorchModel`".

Yinon-Polak:
I think it's better to decouple the model and trainer from the specific usage (MyCoolTorchClassifier),
so that we'll be able reuse the same trainer/model in other usecases.

i thought of getting rid of the fit method in the PyTorchClassifierMultiTarget / PyTorchRegressor.
making this classes as an predict only class.

then the user inherits is PyTorchClassifierMultiTarget / PyTorchRegressor and implemetns the fit method,
something like this:

```python
class MLPPyTorchClassifierMultiTarget(PyTorchClassifierMultiTarget):

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        model_training_params = self.freqai_info.get("model_training_parameters", {})
        self.trainer_kwargs: Dict[str, any] = model_training_params.get("trainer_kwargs", {})
        self.model_kwargs: Dict[str, any] = model_training_params.get("model_kwargs", {})

    def fit(self, data_dictionary: Dict, dk: FreqaiDataKitchen, **kwargs) -> Any:
        """
        User sets up the training and test data to fit their desired model here
        :param data_dictionary: the dictionary constructed by DataHandler to hold
                                all the training and test data/labels.
        :raises ValueError: If self.class_names is not defined in the parent class.

        """

        class_names = self.get_classes_names()
        self.convert_label_column_to_int(data_dictionary, dk, class_names)
        n_features = data_dictionary["train_features"].shape[-1]
        model = PyTorchMLPModel(
            input_dim=n_features,
            output_dim=len(class_names),
            **self.model_kwargs
        )
        model.to(self.device)
        optimizer = torch.optim.AdamW(model.parameters(), lr=self.learning_rate)
        criterion = torch.nn.CrossEntropyLoss()
        init_model = self.get_init_model(dk.pair)
        trainer = PyTorchModelTrainer(
            model=model,
            optimizer=optimizer,
            criterion=criterion,
            model_meta_data={"class_names": class_names},
            device=self.device,
            init_model=init_model,
            **self.trainer_kwargs,
        )
        trainer.fit(data_dictionary)
        return trainer
```

this address the need to have flexibility with the trainer and model initialization.  
here the only thing that user needs to copy, from one implantation to the next
is (in the classifier one):
```python
    class_names = self.get_classes_names()
    self.convert_label_column_to_int(data_dictionary, dk, class_names)
```
this is important as we can have the `self.calss_name` attribute only when calling the fit method.
in the regressor one i don't think we'll need to copy anything. 

pros:
- decuples trainer/model
- more readable

cons:
- fit and predict separation creates dependency between between the model predictions to the prediction post processing done in predict.

robcaulk:
> I think it's better to decouple the model and trainer from the specific usage (MyCoolTorchClassifier), so that we'll be able reuse the same trainer/model in other usecases.
> 
> i thought of getting rid of the fit method in the PyTorchClassifierMultiTarget / PyTorchRegressor. making this classes as an predict only class.
> 
> then the user inherits is PyTorchClassifierMultiTarget / PyTorchRegressor and implemetns the fit method, something like this:
> 
> ```python
> class MLPPyTorchClassifierMultiTarget(PyTorchClassifierMultiTarget):
> 
>     def __init__(self, **kwargs):
>         super().__init__(**kwargs)
>         model_training_params = self.freqai_info.get("model_training_parameters", {})
>         self.trainer_kwargs: Dict[str, any] = model_training_params.get("trainer_kwargs", {})
>         self.model_kwargs: Dict[str, any] = model_training_params.get("model_kwargs", {})
> 
>     def fit(self, data_dictionary: Dict, dk: FreqaiDataKitchen, **kwargs) -> Any:
>         """
>         User sets up the training and test data to fit their desired model here
>         :param data_dictionary: the dictionary constructed by DataHandler to hold
>                                 all the training and test data/labels.
>         :raises ValueError: If self.class_names is not defined in the parent class.
> 
>         """
> 
>         class_names = self.get_classes_names()
>         self.convert_label_column_to_int(data_dictionary, dk, class_names)
>         n_features = data_dictionary["train_features"].shape[-1]
>         model = PyTorchMLPModel(
>             input_dim=n_features,
>             output_dim=len(class_names),
>             **self.model_kwargs
>         )
>         model.to(self.device)
>         optimizer = torch.optim.AdamW(model.parameters(), lr=self.learning_rate)
>         criterion = torch.nn.CrossEntropyLoss()
>         init_model = self.get_init_model(dk.pair)
>         trainer = PyTorchModelTrainer(
>             model=model,
>             optimizer=optimizer,
>             criterion=criterion,
>             model_meta_data={"class_names": class_names},
>             device=self.device,
>             init_model=init_model,
>             **self.trainer_kwargs,
>         )
>         trainer.fit(data_dictionary)
>         return trainer
> ```
> 
> this address the need to have flexibility with the trainer and model initialization. here the only thing that user needs to copy, from one implantation to the next is (in the classifier one):
> 
> ```python
>     class_names = self.get_classes_names()
>     self.convert_label_column_to_int(data_dictionary, dk, class_names)
> ```
> 
> this is important as we can have the `self.calss_name` attribute only when calling the fit method. in the regressor one i don't think we'll need to copy anything.
> 
> pros:
> 
>     * decuples trainer/model
> 
>     * more readable
>       cons:
> 
>     * fit and predict separation creates dependency between between the model predictions to the prediction post processing done in predict.

I think your proposed structure is superior to my suggestion above :-)

robcaulk:
I am also realizing now that we will need to edit the `./setup.sh` file so that the user message says "Do you also want dependencies for freqai-rl (~700mb additional space required) [y/N]? " to instead say "Do you also want dependencies for freqai-torch blah blah."

Similarly we would be changing the docker images to be `Docker.freqai_torch` instead of `Docker.freqai_rl`

The question is, is that sufficient or do you @xmatthias prefer we split out stablebaselines3 and gym stuff away from the Torch base image? I would say this is OK without splitting them - they are pretty light. Only issue is that they could hold back Torch only builds (especially with Torch 2.0 out now).

xmatthias:
> Thoughts? @xmatthias would also like to get your opinion on the matter if you have time.

I'm glad you have agreed on a better structure for this already. I'm not a huge fan of nested classes - and adding more for different usecases would bloat the interface quite some - and make it rather difficult to transfer knowledge between different model types (not necessarily copy/paste - but knowing how to write a model).


> I am also realizing now that we will need to edit the `./setup.sh` file so that the user message says "Do you also want dependencies for freqai-rl (~700mb additional space required) [y/N]? " to instead say "Do you also want dependencies for freqai-torch blah blah."
> 
> Similarly we would be changing the docker images to be `Docker.freqai_torch` instead of `Docker.freqai_rl`
> 
> The question is, is that sufficient or do you @xmatthias prefer we split out stablebaselines3 and gym stuff away from the Torch base image? I would say this is OK without splitting them - they are pretty light. Only issue is that they could hold back Torch only builds (especially with Torch 2.0 out now).

I'm not sure i understand the reason for this. As far as i can see, no new dependency has been added for this PR (we depend on Torch already).

`freqai_rl` will also need torch - so you can't get rid of this (at least it's imported in the current `BaseReinforcementLearningModel`).
I'd keep to as few questions as possible - maybe rephrase it to ask "Do you also want dependencies for freqai-rl/ freqai-torch (~700mb additional space required) [y/N]? " - but i'm clearly against adding yet another question - which would actually install the SAME dependency.

I also don't really see why we would rename the dockerfile. Different docker-files are to not bloat everything - but as long as freqai_rl will require torch - i don't see us having a "freqai_torch" image (or dockerfile) - at least i don't see a reason for that.


> Only issue is that they could hold back Torch only builds (especially with Torch 2.0 out now).

if torch 2.0 and gym (or sb3) conflict, then we'll not be able to use torch2.0 until the other tools support it.

robcaulk:
> > Thoughts? @xmatthias would also like to get your opinion on the matter if you have time.
> 
> I'm glad you have agreed on a better structure for this already. I'm not a huge fan of nested classes - and adding more for different usecases would bloat the interface quite some - and make it rather difficult to transfer knowledge between different model types (not necessarily copy/paste - but knowing how to write a model).
> 
> > I am also realizing now that we will need to edit the `./setup.sh` file so that the user message says "Do you also want dependencies for freqai-rl (~700mb additional space required) [y/N]? " to instead say "Do you also want dependencies for freqai-torch blah blah."
> > Similarly we would be changing the docker images to be `Docker.freqai_torch` instead of `Docker.freqai_rl`
> > The question is, is that sufficient or do you @xmatthias prefer we split out stablebaselines3 and gym stuff away from the Torch base image? I would say this is OK without splitting them - they are pretty light. Only issue is that they could hold back Torch only builds (especially with Torch 2.0 out now).
> 
> I'm not sure i understand the reason for this. As far as i can see, no new dependency has been added for this PR (we depend on Torch already).
> 
> `freqai_rl` will also need torch - so you can't get rid of this (at least it's imported in the current `BaseReinforcementLearningModel`). I'd keep to as few questions as possible - maybe rephrase it to ask "Do you also want dependencies for freqai-rl/ freqai-torch (~700mb additional space required) [y/N]? " - but i'm clearly against adding yet another question - which would actually install the SAME dependency.
> 
> I also don't really see why we would rename the dockerfile. Different docker-files are to not bloat everything - but as long as freqai_rl will require torch - i don't see us having a "freqai_torch" image (or dockerfile) - at least i don't see a reason for that.
> 
> > Only issue is that they could hold back Torch only builds (especially with Torch 2.0 out now).
> 
> if torch 2.0 and gym (or sb3) conflict, then we'll not be able to use torch2.0 until the other tools support it.

I did not suggest adding new docker files or new questions. On the contrary, I suggested we change the name of them to something more descriptive of what people are getting. 

i.e. it makes sense that if I want RL I get the `Docker.freqai_torch` and I answer "yes" to "Do you want torch?" 

It makes no sense if I dont want RL (I only want Torch) to need to use `Docker.freqai_rl` and answer "yes" to "Do you want RL?"

tests/freqai/test_freqai_interface.py
Yinon-Polak: moved this test to `test_extract_data_and_train_model_Classifiers`

Yinon-Polak: added this line to freqai_test_classifier.py inside set_freqai_targets method, declaring class_names
```python
self.freqai.class_names = ["down", "up"]
```
i use this attribute in the classifier to:
1. create Dict[str, int] maaping, to encode decode the str class names to int, as it is necessary by torch modules.
2. initialize the model with proper number of output classes

freqtrade/freqai/prediction_models/PyTorchClassifierMultiTarget.py
Yinon-Polak: > I guess this will need to be added to a new classifier test strat (I elaborated on this in the other comment). But it will also need a bit of documentation for users to understand. I can help with documentation if you would like.

Added this line to freqai_test_classifier.py inside set_freqai_targets method, declaring class_names
```python
self.freqai.class_names = ["down", "up"]
```
I use this attribute in the classifier to:
1. create Dict[str, int] maaping, to encode decode the str class names to int, as it is necessary by torch modules.
2. initialize the model with proper number of output classes

xmatthias:
> i.e. it makes sense that if I want RL I get the `Docker.freqai_torch` and I answer "yes" to "Do you want torch?"
> 
> It makes no sense if I dont want RL (I only want Torch) to need to use `Docker.freqai_rl` and answer "yes" to "Do you want RL?"

This implies a base understanding of "what you get" already if you ask me.
Using Reinforcement Learning does not automatically imply the usage of Torch - so if i look at it from a new user's perspective - i'll not know the answer if you ask "do you want torch?", and will probably answer with "no" (there's alternatives to torch when using reinforcement Learning) - i want reinforcement learning - not torch.

i'm happy with rephrasing of the question to something that includes both (RL and Torch) - but using torch alone will be wrong.

In the end, it comes down to documentation more than anything - but i don't see the need to modify the dockerfile names at this point. that's just work for no good reason.

Lz1688:
我我就想知道一个问题啦，你们这么多人去点

你们这么多人去欺负我一个妹妹，你们这样子很光荣吗？我就想请问一下嘛，先不管说这些手机呢哈，我们的设备我们是怎么用的啦？但是我们当初是真的有交钱呐，我妹她也是真实的身份呐。而且也是同意的啦，填写也是这样子啦，你们不要跟我说什么有的没的啦今天你们几个公司的人一上来就把他当成是犯人，对吗？

获取Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

tests/freqai/test_freqai_interface.py
```diff
@@ -85,6 +87,22 @@ def test_extract_data_and_train_model_Standard(mocker, freqai_conf, model, pca,
     if 'test_3ac' in model or 'test_4ac' in model:
         freqai_conf["freqaimodel_path"] = str(Path(__file__).parents[1] / "freqai" / "test_models")
 
+    if 'PyTorchMLPRegressor' in model:
+        model_save_ext = 'zip'
+        freqai_conf['freqai']['model_training_parameters'].update({
```
robcaulk: Can we make this a function or fixture so that it is place in one location (probably in the conftest.py). That way we can make modifications to it in one place without needing to ensure the same change is made across all the tests.

Yinon-Polak: did exactly that in previous commit

build_helpers/publish_docker_multi.sh
```diff
@@ -64,6 +65,7 @@ docker build --cache-from freqtrade:${TAG_FREQAI} --build-arg sourceimage=${CACH
 docker tag freqtrade:$TAG_PLOT ${CACHE_IMAGE}:$TAG_PLOT
 docker tag freqtrade:$TAG_FREQAI ${CACHE_IMAGE}:$TAG_FREQAI
 docker tag freqtrade:$TAG_FREQAI_RL ${CACHE_IMAGE}:$TAG_FREQAI_RL
+docker tag freqtrade:$TAG_FREQAI_RL ${CACHE_IMAGE}:$TAG_FREQAI_TORCH
```
robcaulk: @xmatthias please have a look at this to ensure I am complying with your wishes. The goal here is not to push a new image, but to simply add another tag to the RL image.

robcaulk (dismissed):
LGTM, really nice work @Yinon-Polak 

@xmatthias please wait on merge.

build_helpers/publish_docker_multi.sh
xmatthias: well i don't mind this necessarily (though it's pointless) - but it'll have no impact, as it's creating a new tag on the cache image (the one nobody knows about, and more importantly, nobody should be using).

I'll have a look at it later

freqtrade/freqai/freqai_interface.py
```diff
@@ -83,6 +83,7 @@ def __init__(self, config: Config) -> None:
         self.CONV_WIDTH = self.freqai_info.get('conv_width', 1)
         if self.ft_params.get("inlier_metric_window", 0):
             self.CONV_WIDTH = self.ft_params.get("inlier_metric_window", 0) * 2
+        self.class_names: List[str] = []  # used in classification subclasses
```
xmatthias, line 86: This comment suggests that `class_names` is used throughout the classifier models - but i don't think that's true?

is it intentional that we have these only implemented in the new Torch models?

build_helpers/publish_docker_multi.sh
xmatthias: Docker part is fine now - and should create a new tag.

freqtrade/freqai/freqai_interface.py
Yinon-Polak: Yes it is intentional. 
We use for:

1. create Dict[str, int] maaping, to encode decode the str class names to int, as it is necessary by torch modules.
2. initialize the model with proper number of output classes

XGBoost, CatBoost have their own built-in methods for dealing with class targets of type str. This attribute could be used with any model that doesn't have built-in method.

Yinon-Polak: We can change the comment to somthing like:
used in pytorch classification subclasses

On the other end, 
1. i think it's a good practice to declare class names in the strategy.
2. There will be a future PR with tensorboard logger that will be excessible to all models. 
Then it might be more stable/convenient to use `class_names` attribute in XGBoost/CatBoost classifiers logging.

xmatthias: The fact that "it's intentional" is enough for me already :+1: 

I assumed that the other models have this builtin - just making sure ;)

robcaulk (approved):
LGTM

PR #8307, opened 2023-03-11
Improve tensorboard_log

initrv:
## Summary

This PR aims to make it easier to use tensorboard_log() and let you categorize the metrics.
Categories are very handy when you have a lot of metrics.

Additionally, it seems to me that the ability to increase the metric by a certain value is useless.
And in order to log a specific value, we have to specify inc=False every time.

## Quick changelog

- Added "category" parameter, allows you to visually separate metrics in the tensorboard.
- Removed "inc" parameter, increment metric if no value is specified.

### DIFF
--- a/docs/freqai-reinforcement-learning.md
+++ b/docs/freqai-reinforcement-learning.md
@@ -248,13 +248,13 @@ FreqAI also provides a built in episodic summary logger called `self.tensorboard
             """
             def calculate_reward(self, action: int) -> float:
                 if not self._is_valid(action):
-                    self.tensorboard_log("is_valid")
+                    self.tensorboard_log("invalid")
                     return -2
 
 ```
 
 !!! Note
-    The `self.tensorboard_log()` function is designed for tracking incremented objects only i.e. events, actions inside the training environment. If the event of interest is a float, the float can be passed as the second argument e.g. `self.tensorboard_log("float_metric1", 0.23)` would add 0.23 to `float_metric`. In this case you can also disable incrementing using `inc=False` parameter.
+    The `self.tensorboard_log()` function is designed for tracking incremented objects only i.e. events, actions inside the training environment. If the event of interest is a float, the float can be passed as the second argument e.g. `self.tensorboard_log("float_metric1", 0.23)`. In this case the metric values are not incremented.
 
 ### Choosing a base environment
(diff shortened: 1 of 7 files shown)


robcaulk:
Thanks for this, it looks well thought out. But my main concern is that it *looks* like a breaking change. In other words, as soon as we push this to stable, anyone using the logger with `inc=False` will start getting an error. 

Can you confirm my concern, and if so, is there a way we can avoid this?

initrv:
Yes, unfortunately, this is a breaking change for some users. I'm not excited about it, but we can leave the `inc` attribute in place and not use it, for backward compatibility purposes. This solution also makes it difficult to use positional arguments, and in the future would still like to remove the extra argument
```python
def tensorboard_log(self, metric: str, value: Optional[Union[int, float]] = None,
                        inc: Optional[bool] = None, category: str = "custom"):
...
:param inc: (deprecated) sets whether the `value` is incremented or not
...
```
It would be nice to make a deprecation warning, but it would be very spammy and looks redundant for such a small function

robcaulk:
> Yes, unfortunately, this is a breaking change for some users. I'm not excited about it, but we can leave the `inc` attribute in place and not use it, for backward compatibility purposes. This solution also makes it difficult to use positional arguments, and in the future would still like to remove the extra argument
> 
> ```python
> def tensorboard_log(self, metric: str, value: Optional[Union[int, float]] = None,
>                         inc: Optional[bool] = None, category: str = "custom"):
> ...
> :param inc: (deprecated) sets whether the `value` is incremented or not
> ...
> ```
> 
> It would be nice to make a deprecation warning, but it would be very spammy and looks redundant for such a small function

I think of any removal, this is the most ideal type. I don't know a single person who is using it. If they are, they would surely pop up quickly and the solution is fast and easy. 

I would vote for just plain removal of the `inc` kwarg. Thoughts @xmatthias ?

xmatthias:
> I would vote for just plain removal of the `inc` kwarg. Thoughts @xmatthias ?

i'm fine with that - if somebody used it, it might be better to have it fail than to silently ignore the parameter (which is what we're doing at the moment, with the non-used `inc` kwarg).

robcaulk (approved):
LGTM

PR #8327, opened 2023-03-13
Support for python 3.11 via setup.sh script

skinner12:
## Summary

Use freqtrade with latest python 3.11

## What's new?

Support for __python 3.11__

Now it's possible to use python 3.11 when run `./setup.sh`

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -25,7 +25,7 @@ function check_installed_python() {
         exit 2
     fi
 
-    for v in 10 9 8
+    for v in 11 10 9 8
     do
         PYTHON="python3.${v}"
         which $PYTHON


xmatthias:
Thanks for this, and for helping to make this project better :100: 

Unfortunately, we have several dependencies (mainly freqAI related) that are incompatible with 3.11 at this point.
Using 3.11 via the install script would therefore result in a partial (or faulty) installation if a user selects freqAI dependencies, without really providing a reason why.

I think it's better to not detect 3.11 in this case - as the alternative would be head-scratching after a (seemingly successful) installation.

xmatthias (approved):
with the last dependency (catboost) supporting 3.11, we can finally fully support this.

LGTM :+1:

PR #8336, opened 2023-03-16
Added the latest Gymnasium version 0.28(will be released shortly),

richardjozsa:
## Summary

It is a highly experimental setup, but to be prepared when changes will happen, I wanted to create this PR.

So the gymnasium 0.28 will be released shortly, and than SB3 and SB3 contrib will adapt to it to be able to use finally the latest version of gym/gymnasium. The link for the branches are the PR-s what will be released but in a later period of time. I have tested them for days, they are working perfectly.

Solve the issue: #___

## Quick changelog

-Base files adapted to gymnasium
-Experimental requirements added

## What's new?

Faster trainings, a wider feature range for the RL part of the package.


requirements-freqai-rl.txt
```diff
@@ -3,8 +3,10 @@
 
 # Required for freqai-rl
 torch==1.13.1; python_version < '3.11'
-stable-baselines3==1.7.0; python_version < '3.11'
-sb3-contrib==1.7.0; python_version < '3.11'
+#until these branches will be released we can use this
+git+https://github.com/Farama-Foundation/Gymnasium@main
+git+https://github.com/DLR-RM/stable-baselines3@feat/gymnasium-support
+git+https://github.com/Stable-Baselines-Team/stable-baselines3-contrib@feat/gymnasium-support
```
xmatthias: I've marked that PR as "WIP" for now. It'll not prevent people from using it / testing this - but we'll not be adding dependencies which MUST be installed directly from github to actually work.

Once the versions are merged (and released) - please update this PR to the pinned pypi versions that include this support.

richardjozsa: Of course, I thought the same!

xmatthias: This should actually now be available through `sb3_contrib==2.0.0a4` - according to https://github.com/DLR-RM/stable-baselines3/issues/871 - as well as `stable-baselines3==2.0.0a5`.

in my understanding, this would mean we can remove all of the above git installs ... 

(Updating this is now becoming crucial, as the lastest pip version refuses to install gym==0.21 - causing all CI to fail).
I think, we can therefore also depend on the alpha versions if we need (the breaking point so far was the installation from github) - obviously assuming that everything works as we expect.

richardjozsa: I will test thoose  today and let you know the results and modify this PR!

richardjozsa: Updated the questionable lines, tested, working perfectly with the changes of this PR!

freqtrade/freqai/RL/TensorboardCallback.py
```diff
@@ -44,7 +45,10 @@ def _on_training_start(self) -> None:
     def _on_step(self) -> bool:
 
         local_info = self.locals["infos"][0]
-        tensorboard_metrics = self.training_env.get_attr("tensorboard_metrics")[0]
+        if isinstance(self.training_env, SubprocVecEnv):
+            tensorboard_metrics = self.training_env.get_attr("tensorboard_metrics")[0]
+        else:
+            tensorboard_metrics = self.training_env.tensorboard_metrics
```
xmatthias: @robcaulk according to debugging one of the tests, this is of type  `VecMonitor` - not of type BaseEnvironment.
I suspect the correct type for training_env would therefore be `Union[SubprocVecEnv, VecMonitor]` ...

robcaulk: It depends, if they use `_multiproc` then it will be a VecMonitor/SubProcVecEnv. If they are using the regular `ReinforcementLearner` then it will be of type `BaseEnvironment`. Yes, this is the downside of trying to support both methods, fun typing.

robcaulk (approved):
Really nice, thanks for the work on this ;)

PR #8361, opened 2023-03-21
featherdatahandler: implement trades_store/_trades_load

TheJoeSchr:
as discussed on discord. for now it's implemented using pandas functions. I haven't tested it against a big backtesting, but it's better than not existing


TheJoeSchr:
I was thinking, in most usecases we will convert to pandas DataFrame eventually, so we can't overcome this memory load anyways. Maybe it would be better to accept it and rewrite the interface to use DataFrame. That way we don't duplicate the trades as `List` and as `DataFrame`

xmatthias:
> in most usecases we will convert to pandas DataFrame eventually, 

That's true for converting to ohlcv ... but it's clearly not true for downloading trades data (which is the only case where we actually SAVE the dataframe) - where appending to a list has big benefits over appending / concatenating dataframes.

xmatthias:
@TheJoeSchr anything that's preventing this from being "no longer a draft" ? 

i think the PR is actually mostly fine (i've reverted a few non-related formatting changes ..)

TheJoeSchr:
@xmatthias no it's fine from my point of view. I thougt I needed to do a big test and never got the time. But it works and is significantly faster on my usecases than hdf5, so ready to merge

xmatthias (approved):
great - then let's have this ;) 

as you correctly pointed out "somewhere" - it's the same approach we use with hdf5 - which is also converting it to a dataframe before writing / after reading.
As such, it'll just give us the feather performance - without drawbacks on top of hdf5.

LGTM :+1:

PR #8369, opened 2023-03-22
backtest - lookahead_analysis

hippocritical:
this PR is a continuation of https://github.com/freqtrade/freqtrade/pull/8313
pulling the strategy-updater into freqtrade broke backtest_bias_tester which was a branch of a branch => the strategy_updater in my fork so I just moved to dev and kept the code the same for now. Improving it now slowly.

## Summary

This PR aims to solve the issue of "hey is this strategy having any backtest bias?"
Many more strategies than thought beforehand are infested with bias without the programmers even knowing it.
It is not easy to detect look ahead bias when you are starting and even for veteran programmers it sometimes can't be detected easily.

This link shows how many strategies already are infested by backtest look ahead bias and that are even still circulating - many of them are even live without the user knowing it.
http://frequenthippo.ddns.net:3000/d/4CwBZom4z/freqtrade-dryruns_v2?orgId=1&viewPanel=7278

## Quick changelog

- Implement backtest bias tester as new command

## What's new?
The graphic shows how the logic of the backtest bias tester works.
1. Make it have 20 fully opened and closed trades by running an initial backtest
2. check every backtest trade entry signal and exit signal
3. check the last row of a dataframe to see if it cheats by changing the indicators which look into the future
4. report what it found in either dataframe (including indicators that are biased) and entry / exit signals.
5. output result in CSV file (coming in the next PR)

grafik

## Plan for the future (this pr is not done, just a first draft!)

Putting this out there so people can comment on if they don't like something.
Plan:
- make variables adaptive like timerange and max_try_signals
- create a proper list of results at the end of a --strategy-list run
- output to specified .csv-file ('cause I am using that currently myself for the grafana server and so others can have a list of that too for later use or sharing things)
- tests
- docs

Current command which is tested with is as follows:
backtest_lookahead_bias_checker
--config=user_data/configs_backtest/config_baseline.json
--config=user_data/pairlists/binance_spot/USDT/monthly/monthly_90_USDT_0,01_minprice_20220901.json
--strategy-list=A1
--timeframe=5m

ps: currently I dont like the name of the command since it's way too long. If you have any suggestions then please go ahead :)


freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,241 @@
+# pragma pylint: disable=missing-docstring, W0212, line-too-long, C0103, unused-argument
```
xmatthias: well now that's an argument for sure - but isn't the "backtest lookahead bias" better fitted in the "optimize" directory?
also from the command perspective - it's `backtest-` - not `strategy-` ... 

it's just a feeling though - let me know your thoughs as to why you moved it to the strategy dir, maybe?

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,241 @@
...
+
+import pandas
+
+from freqtrade.configuration import TimeRange
+from freqtrade.data.history import get_timerange
+from freqtrade.exchange import timeframe_to_minutes
+from freqtrade.optimize.backtesting import Backtesting
+
+
+class backtest_lookahead_bias_checker:
```
xmatthias: Class namess should be `CamelCase`, not `snake_case` ...

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,241 @@
...
+
+            # register if buy signal is broken
+            if not self.report_signal(
+                    self.entry_varholder.result,
+                    "open_date", self.entry_varholder.compared_dt):
+                self.current_analysis.false_entry_signals += 1
+
+            # register if buy or sell signal is broken
+            if not self.report_signal(self.entry_varholder.result,
+                                      "open_date", self.entry_varholder.compared_dt) \
```
xmatthias: can we ... not use `\` to separate if statements ... wrapping the whole condition in "()" will make it look way better, you'll see (not just here - but also below / elsewhere).

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: I asked some months ago, frog said that he would group both under strategy_utils so I did
If you want to have the bias checker under optimize then sure, you re the boss I am indifferent about it

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: yes, this basically is just a "throw it in there and have a 1st draft"
will change that to fit the coding guidelines you gave in the strategy updater

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: the pylint line will be gone next time.

xmatthias: i'm ... not complaining about pylint lines - we have them at random places - but don't actually use pylint - so it's technically irrelevant to have it or not :laughing: 

---

well i'm ... indecisive - as it'll fit in both areas - but i think things from the strategy package are more likely to be imported (also by mistake, eventually) - so i think keeping that "emptier" makes more sense

freqtrade/commands/arguments.py
```diff
@@ -116,8 +116,14 @@
 
 NO_CONF_ALLOWED = ["create-userdir", "list-exchanges", "new-strategy"]
 
-ARGS_STRATEGY_UTILS = ["strategy_list", "strategy_path", "recursive_strategy_search"]
+ARGS_STRATEGY_UPDATER = ARGS_COMMON_OPTIMIZE + ["strategy_list"]
```
xmatthias: you might want to revert the content of the variable (not necessarily the variable name change, which is good).

args_common_optimize has way too many arguments which are useless for the strategy-updater.

hippocritical:
all changes currently have been locally integrated. will check the functionality today and push

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: just say so, will leave it in there until you decided either way :)

hippocritical: that pylint sneaked in due to pycharm asking for if it should enter those and I clicked to see what happens - then didnt delete it again

hippocritical:
another in-between-state. Will add more functionality now.
Currently when testing with futures-data it fails and don't know why (yet)

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,226 @@
...
+                                  f"{str(self_value)} != {str(other_value)}")
+
+    def prepare_data(self, varHolder, pairs_to_load):
+        prepare_data_config = copy.deepcopy(self.local_config)
+        prepare_data_config['timerange'] = (str(self.dt_to_timestamp(varHolder.from_dt)) + "-" +
+                                            str(self.dt_to_timestamp(varHolder.to_dt)))
+        prepare_data_config['pairs'] = pairs_to_load
+        self.backtesting = Backtesting(prepare_data_config)
+        self.backtesting._set_strategy(self.backtesting.strategylist[0])
+        varHolder.data, varHolder.timerange = self.backtesting.load_bt_data()
```
xmatthias: > another in-between-state. Will add more functionality now.
> Currently when testing with futures-data it fails and don't know why (yet)

from afar, i'd say because you're missing the call to `.load_bt_data_detail()`...

it's a guess - but since you're not using `.start()` - a pretty good one :laughing:

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,226 @@
...
+    class Analysis:
+        def __init__(self):
+            self.total_signals = 0
+            self.false_entry_signals = 0
+            self.false_exit_signals = 0
+            self.false_indicators = []
+            self.has_bias = False
+
+        total_signals: int
+        false_entry_signals: int
```
xmatthias: I really don't like this being a nested class (nor the above)
there's little reason for that - and it makes usage kinda odd (same for the class above, obviously).

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,226 @@
+import copy
+from copy import deepcopy
+from datetime import datetime, timedelta, timezone
+
+import pandas
```
xmatthias: i think it'd be better to import just the dataframe (`from pandas import DataFrame`).

if we "must" import the whole pandas library - then please use the pandas convention of `as pd` ... :)

freqtrade/commands/arguments.py
```diff
@@ -116,8 +116,15 @@
 
 NO_CONF_ALLOWED = ["create-userdir", "list-exchanges", "new-strategy"]
 
-ARGS_STRATEGY_UTILS = ["strategy_list", "strategy_path", "recursive_strategy_search"]
+ARGS_STRATEGY_UPDATER = ["strategy_list"]
```
xmatthias: i do like the variable name better - but `"strategy_path", "recursive_strategy_search"` should be left where they are ... 
using a different strategy path (or recursive search) is a quite common usecase...

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: varHolder and Analysis are now outside

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: done

freqtrade/commands/arguments.py
hippocritical: done

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: done

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: done

TheJoeSchr:
passing in a "unfinished" timerange fails, e.g. `--timerange 20230401-`
```
└> freqtrade backtest-lookahead-bias-checker --strategy-list Volumio --timerange 20230401-
2023-04-13 10:59:30,965 - freqtrade - INFO - freqtrade 2023.4.dev-e5e63d5be
2023-04-13 10:59:30,965 - freqtrade.configuration.load_config - INFO - Using config: user_data/config.json ...
2023-04-13 10:59:30,966 - freqtrade.loggers - INFO - Verbosity set to 0
2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Parameter --timerange detected: 20230401- ...
2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Using user-data directory: /home/joe/projects/ft/freqtrade-hippocritical/user_data ...
2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Using data directory: /home/joe/projects/ft/freqtrade-hippocritical/user_data/data/binance ...
2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Using strategy list of 1 strategies
2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Parameter --cache=day detected ...
2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Filter trades by timerange: 20230401-
2023-04-13 10:59:30,966 - freqtrade.exchange.check_exchange - INFO - Checking exchange...
2023-04-13 10:59:30,973 - freqtrade.exchange.check_exchange - INFO - Exchange "binance" is officially supported by the Freqtrade development team.
2023-04-13 10:59:30,973 - freqtrade.configuration.configuration - INFO - Using pairlist from configuration.
2023-04-13 10:59:30,973 - freqtrade.configuration.config_validation - INFO - Validating configuration ...
2023-04-13 10:59:30,999 - freqtrade.resolvers.iresolver - WARNING - Could not import /home/joe/zk/efforts/ft/volumio-strategy/user_data/strategies/MockVolumio.py due to 'unexpected indent (MockVolumio.py, line 116)'
2023-04-13 10:59:30,999 - freqtrade.resolvers.iresolver - WARNING - Could not import /home/joe/zk/efforts/ft/volumio-strategy/user_data/strategies/Prototype 'temp'.py due to 'name 'IStrategy' is not defined'
Bias test of Volumio.py started.
Parsing of parsed_timerange failed. exiting!
checking look ahead bias via backtests of Volumio.py took 0.0 seconds.
```

hippocritical:
> passing in a "unfinished" timerange fails, e.g. `--timerange 20230401-`
> 
> ```
> └> freqtrade backtest-lookahead-bias-checker --strategy-list Volumio --timerange 20230401-
> 2023-04-13 10:59:30,965 - freqtrade - INFO - freqtrade 2023.4.dev-e5e63d5be
> 2023-04-13 10:59:30,965 - freqtrade.configuration.load_config - INFO - Using config: user_data/config.json ...
> 2023-04-13 10:59:30,966 - freqtrade.loggers - INFO - Verbosity set to 0
> 2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Parameter --timerange detected: 20230401- ...
> 2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Using user-data directory: /home/joe/projects/ft/freqtrade-hippocritical/user_data ...
> 2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Using data directory: /home/joe/projects/ft/freqtrade-hippocritical/user_data/data/binance ...
> 2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Using strategy list of 1 strategies
> 2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Parameter --cache=day detected ...
> 2023-04-13 10:59:30,966 - freqtrade.configuration.configuration - INFO - Filter trades by timerange: 20230401-
> 2023-04-13 10:59:30,966 - freqtrade.exchange.check_exchange - INFO - Checking exchange...
> 2023-04-13 10:59:30,973 - freqtrade.exchange.check_exchange - INFO - Exchange "binance" is officially supported by the Freqtrade development team.
> 2023-04-13 10:59:30,973 - freqtrade.configuration.configuration - INFO - Using pairlist from configuration.
> 2023-04-13 10:59:30,973 - freqtrade.configuration.config_validation - INFO - Validating configuration ...
> 2023-04-13 10:59:30,999 - freqtrade.resolvers.iresolver - WARNING - Could not import /home/joe/zk/efforts/ft/volumio-strategy/user_data/strategies/MockVolumio.py due to 'unexpected indent (MockVolumio.py, line 116)'
> 2023-04-13 10:59:30,999 - freqtrade.resolvers.iresolver - WARNING - Could not import /home/joe/zk/efforts/ft/volumio-strategy/user_data/strategies/Prototype 'temp'.py due to 'name 'IStrategy' is not defined'
> Bias test of Volumio.py started.
> Parsing of parsed_timerange failed. exiting!
> checking look ahead bias via backtests of Volumio.py took 0.0 seconds.
> ```

should be fixed

hippocritical:
things to do:
check freqai strategies for compatibility
add docs
add tests

hippocritical:
@xmatthias 
Is there a solution for an edge case where the bias happens only for one candle and during an exit?
Currently the dataframe has to be elongated by one candle to allow proper exiting which will give the strategy one more candle for bias if it wants to look into the future for exits.

Strategy-bound-Exits are (just so the statement is clear) only allowed at a non-last-candle whereas force-exits happen always at the very last candle.
I want the last candle to be there at the point of the exit and not one longer.

So any exits that are only catching one more candle into the future will not be caught since the dataframe data still is the same in the look ahead scope of the exit itself.

=> would there be a possibility to disable that check and leave an exit just "hanging" for the last candle so the strategy can decide to either sell or not sell at the very last candle for the timerange?
Currently freqtrade force-exits at the last candle which is proper - just not for my case while bias checking.

Can't move it simply to advise_all_indicators since there are no exit_long's (etc) yet.
A bias for one candle on the exit-condition will currently never be caught.

xmatthias:
I'm not entirely sure what you're after - but leaving trades "open" at the end of the backtesting period would not have the effect you hope for.

As the trade is not closed - it'll simply be "skipped" - and act as if it were deleted (though it'll still influence other, parallel trades, as it did exist while it was open).

I'm not sure how you'd imagine this to work exactly (or what you'd like backtesting to do with these "left open" trades -  but a simple "allow keeping trades open" will NOT do what you expect it will.

We also forward-shift backtest signals to align with live, since a signal generated on the candle for 10:00 can only be acted on at 11:00 (but it's still at the 10:00 candle) - therefore we must forward-shift this to avoid constant lookbacks.

That'll mean this "very last candle signal" (on the very last candle) is actually gone from the dataframe.

---

Assuming the above would not be a problem - then 
One (potential) alternative which could improve backtest handling in general - and would probably do what you'd hope could be to close the trade as "exit_signal" on close (using close rate instead of the next open) if it's the very last candle, and that last candle has a (dataframe based) exit signal.

I'd not see that as part of this PR (must be done separately) - and i'm also not sure if it'd really work (or how it would work) due to the shifting of the signals for backtesting purposes.

hippocritical:
> I'm not entirely sure what you're after 

This is only for the case that the bias is only one candle long.
When an entry is set that looks into the future by one candle it would not be found otherwise, since at the moment the backtest bias checker is forced to have at least one candle in addition where the trade can force-close. Entries or exits at will are forbidden in the very last candle in a backtest.

It sounds like a real head scratcher to implement for freqtrade and this would only hit an edge case I suppose so it is not ultra vital - but on long timeframes like 1h this could be a "viable" way to game the backtest I suppose.

Currently working on implementing tests for the backtest bias checker and really unsure about several things but we ll see, this commit is quite a bit more complex than the old one regarding tests qq

After the tests are implemented it should be ready for you to have a closer look at. The code itself except the tests are already done - at least I feel that way with freqai and "classic" strategies working both in futures and spot.

xmatthias:
i'm ... not entirely sure you even need to run a backtest for this if i'm completely honest - at least not for this particular scenario.
The last candle is special in backtesting - as we do need to make sure we handle these scenarios correctly.

can't you just run the strategy analysis with the "closing candle" and "closing candle +1" (more or less) - and determine if the signal disappeared (or rather, reappeared)? or if anything else in the dataframe changed?

this should cover pretty much every exit scenario i think (except for maybe custom_exit - but even then we make sure to not actually provide a forward-looking pair if regular, documented methods are used).

> After the tests are implemented it should be ready for you to have a closer look at. 

you may want to change approach here - should changes be required (i didn't look at it yet - don't take this as spoiler), you'll have double work for the testing - adjusting tests + the functional code.

I think it might make more sense to look at it beforehand - that way tests can be done against the "more or less finalized" version (doesn't prevent necessary changes should tests bring up some problems) - which is usually simpler  / faster than adapting again.

hippocritical:
will then have to check wether I can get the fully fledged signal with the last candle dunno atm.
(including populate_entry_trend, confirm_trade_entry... the whole chain)
some could have bias in confirm_trade_entry too to cheat the system.

This might explain the dilemma a little more.
It picture shows what and how it is done.
One trade is verified with three characteristics (of that one is done double, dataframe checking)
- Dataframe checking for each entry and exit, compare to the original backtest, same timestamp
- entry starting from the very beginning of the timerange until the entry candle +1 (setting the entry-timestamp-candle as last candle does not work since it then would deny the entry)
- exit starting from the very beginning of the timerange until the exit candle +1 (setting the exit-timestamp-candle as last candle does not work since then it would always force-exit not verifying any exits by the strategy itself)

So if there is a way to set the timerange on either the entry or the exit candle directly and still have the entries and exits triggered correctly in the analysis then no problem.
My problem is that the entry+1 and exit+1 is forced and at least for me there is no way around it.

So if the bias is only looking one candle into the future and the bias happens in the populate_entry_trend or populate_exit_trend respectively (not in the dataframe!) then there would be the issue with non-detected bias.
The dataframe itself would not be an issue, I could just set the comparison to the very last timestamp, but would prefer to have both timestamps be the same and even be able to catch the bias inside populate_entry_trend or populate_exit_trend (yes there is confirm_trade_entry and conform_trade_exit too, I just mean that the signal has to register fully).

in essence the end = exit+1candle is the issue there

grafik

hippocritical:
> you may want to change approach here - should changes be required (i didn't look at it yet - don't take this as spoiler), you'll have double work for the testing - adjusting tests + the functional code.

If that s the case then this is ready for review.
The only thing that s left except from tests is the issue of 1-candle-lookahead-bias

Would rather do the tests at the end since those are seeming to be quite a chore and more complicated than the strategy-updater by far.

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -53,3 +57,135 @@ def start_conversion(strategy_obj, config):
...
+            bias_checker_instances.append(
+                initialize_single_lookahead_bias_checker(filtered_strategy_obj, config, args))
+    elif 'strategy' in args and args['strategy'] is not None:
+        for strategy_obj in strategy_objs:
+            if strategy_obj['name'] == args['strategy']:
+                bias_checker_instances.append(
+                    initialize_single_lookahead_bias_checker(strategy_obj, config, args))
+                break
+    else:
+        processed_locations = set()
```
xmatthias: i think not providing `--strategy`, nor `--strategy-list`  should fail (aligned to backtesting)

Running through all strategies without ANY filter should not be the default - and can lead to huge runs (i have probably 100ds of strategies in my list - and running through all of them is NEVER what i intend to do - if any, i've forgotten the argument).

if we MUST have a "run for all strategies unfiltered" logic (i don't think we should) - then i'd prefer an explicit flag instead - like `--all-strategies` or similar. that'll avoid unintended runs against the whole strategy directory.

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -53,3 +57,135 @@ def start_conversion(strategy_obj, config):
...
+
+        for filtered_strategy_obj in filtered_strategy_objs:
+            bias_checker_instances.append(
+                initialize_single_lookahead_bias_checker(filtered_strategy_obj, config, args))
+    elif 'strategy' in args and args['strategy'] is not None:
+        for strategy_obj in strategy_objs:
+            if strategy_obj['name'] == args['strategy']:
+                bias_checker_instances.append(
+                    initialize_single_lookahead_bias_checker(strategy_obj, config, args))
+                break
```
xmatthias: i suspect you can simplify this whole block a ton by simply using something around (in pseudo code):
Right now, it's the same logic, somehow nested rather deeply - causing potential errors in (future) refactorings - should something in the content change.

```
if args['strategy_list']:
    strategy_list = args['strategy_list']
elif args['strategy']:
    strategy_list = [args['strategy_list']]

# followed by the regular logic - which now applies ALWAYS to a list.


```

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+        self.full_varHolder = None
+
+        self.entry_varHolder = None
+        self.exit_varHolder = None
+        self.entry_varHolders = []
+        self.exit_varHolders = []
+        self.backtesting = None
+        self.minimum_trade_amount = None
+        self.targeted_trade_amount = None
+        self.failed_bias_check = True
```
xmatthias: This is somehow an odd pattern
`BacktestLookaheadBiasChecker` is not reused as far as i can tell - so the pattern will always be:

call `__init__()` - setting everything to nothing
call `start()` - change half of the settings to "something", which could already be done here if (at least config) were passed into the constructor.

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+import shutil
+from copy import deepcopy
+from datetime import datetime, timedelta, timezone
+
+from pandas import DataFrame
+
+from freqtrade.configuration import TimeRange
+from freqtrade.data.history import get_timerange
+from freqtrade.exchange import timeframe_to_minutes
+from freqtrade.optimize.backtesting import Backtesting
```
xmatthias: considering this is importing Backtesting - i'd move this to the optimize directory.

we try to not import from .optimize modules throughout the bot - so dependencies for that can remain optional.
while not relevant for backtesting, definitely relevant for hyperopt

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -4,9 +4,13 @@
 from pathlib import Path
 from typing import Any, Dict
 
+import pandas as pd
+from tabulate import tabulate
+
 from freqtrade.configuration import setup_utils_configuration
 from freqtrade.enums import RunMode
 from freqtrade.resolvers import StrategyResolver
+from freqtrade.strategy.backtest_lookahead_bias_checker import BacktestLookaheadBiasChecker
```
xmatthias: From an import logic perspective - please move this import to the function where it's used (look at `optimize_commands.py` for samples of that).

the imports at the top level  run for every function (if i run `freqtrade trade`, this line is called) - but this should not be the case for "optimize" functions (for example hyperopt dependencies are not installed by default) - and can lead to random errors (random as in - we didn't foresee them, but they constantly reproduce) if not 100% of the depencencies are installed.

while not necessarily a problem with this particular function - it can still trigger the optimize module to load - which is unnecessarily heavy if these functions are not even used.

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+        self.strategy_obj = strategy_obj
+
+        # first make a single backtest
+        self.full_varHolder = VarHolder()
+
+        # define datetime in human-readable format
+        parsed_timerange = TimeRange.parse_timerange(config['timerange'])
+
+        if parsed_timerange.startdt is None:
+            self.full_varHolder.from_dt = datetime.utcfromtimestamp(0)
```
xmatthias: please don't use this method, it's simply wrong.
it returns a 0 timestamp in utc - but automatically converts it to local time.
This can be no problem - but can also fail if you're in a negative offset timezone.

instead, `datetime.fromtimestamp(0, tz=timezone.utc)` should be used.

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+        else:
+            self.full_varHolder.from_dt = parsed_timerange.startdt
+
+        if parsed_timerange.stopdt is None:
+            self.full_varHolder.to_dt = datetime.now()
```
xmatthias: If you use `tz=timezone.utc` here, you should have all (hope i got them all) timezones in utc timezone, making some workarounds unnecessary.

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+
+                            if not self.current_analysis.false_indicators.__contains__(col_name[0]):
+                                self.current_analysis.false_indicators.append(col_name[0])
+                                print(f"=> found look ahead bias in indicator {col_name[0]}. " +
+                                      f"{str(self_value)} != {str(other_value)}")
+
+    def prepare_data(self, varHolder, pairs_to_load):
+
+        # purge previous data
+        abs_folder_path = pathlib.Path("user_data/models/uniqe-id").resolve()
```
xmatthias: `uniqe-id` is assumed to be the id used ? 

If we assume that's the "backtest bias reporter" - then we should also explicitly set this.
otherwise it's taking the assumption that `"identifier": "uniqe-id",` is set - which we shouldn't hard-code or assume here.

We should also (probably not just here) assume that `Path('user_data')` is the user-data.
i can set this to `--userdir ../my_user_data` - which should be respected / used.

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+            if len(self.entry_varHolders) >= 10:
+                pass
+            # check if the indicators themselves contain biased data
+            self.analyze_indicators(self.full_varHolder, self.entry_varHolder, result_row['pair'])
+            self.analyze_indicators(self.full_varHolder, self.exit_varHolder, result_row['pair'])
+
+        if (self.current_analysis.false_entry_signals > 0 or
+                self.current_analysis.false_exit_signals > 0 or
+                len(self.current_analysis.false_indicators) > 0):
+            print(" => " + self.local_config['strategy_list'][0] + ": bias detected!")
```
xmatthias: i assume the usage of `print` here is intentional?

in general, i'd use logging instead unless there's good reasons against it.
Assuming this can run for a longer period - it'll allow implicit piping to file (`--log-file`) - while print output will be to stdout - which will be gone if you close the terminal (and didn't redirect output).

At any rate, i'd prefer f-strings instead for string concatenation - they feel more modern (and have some implicit type-conversion that can help, too).

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -53,3 +57,135 @@ def start_conversion(strategy_obj, config):
...
+
+def initialize_single_lookahead_bias_checker(strategy_obj, config, args):
+    print(f"Bias test of {Path(strategy_obj['location']).name} started.")
+    start = time.perf_counter()
+    current_instance = BacktestLookaheadBiasChecker()
+    current_instance.start(config, strategy_obj, args)
+    elapsed = time.perf_counter() - start
+    print(f"checking look ahead bias via backtests of {Path(strategy_obj['location']).name} "
+          f"took {elapsed:.1f} seconds.")
+    return current_instance
```
xmatthias: I think you shouldn't return `current_instance` here.

Assuming you run this for several strategies - it'll keep all Backtest objects (including the data they loaded) in memory (they're assigned to `self.backtest` in `BacktestLookaheadBiasChecker()`).

as far as i can tell, only `.current_analysis` - and the strategy name/identifier itself are going to be used - so that's what should be returned.

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+        if pathlib.Path.exists(abs_folder_path):
+            shutil.rmtree(abs_folder_path)
+
+        prepare_data_config = copy.deepcopy(self.local_config)
+        prepare_data_config['timerange'] = (str(self.dt_to_timestamp(varHolder.from_dt)) + "-" +
+                                            str(self.dt_to_timestamp(varHolder.to_dt)))
+        prepare_data_config['exchange']['pair_whitelist'] = pairs_to_load
+
+        self.backtesting = Backtesting(prepare_data_config)
+        self.backtesting._set_strategy(self.backtesting.strategylist[0])
```
xmatthias: Am i correct in reading the code that there will be one `BacktestLookaheadBiasChecker` per strategy backtested?
if yes - what parameters (in the config / backtesting) are we expecting to change between runs?

I'm asking myself if we can't reuse the backtest object (and the data loaded) - currently, we'll load data from disk several times per strategy - multiplied by strategy.
data loading is often the slowest part in backtesting - so reducing the frequency of doing this can greatly help in speeding things up.

freqtrade/commands/strategy_utils_commands.py
hippocritical: --all-strategies are not strictly necessary
it will fail on no strategies mentioned next time.

xmatthias: It's me telling you that I'm not completely against it if you see the need - but would like to have it explicit.
obviously if there's no need, we won't need it 😆

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -53,3 +57,135 @@ def start_conversion(strategy_obj, config):
...
+                (df['filename'] == row_data['filename']) &
+                (df['strategy'] == row_data['strategy'])
+                ] = pd_series
+        else:
+            # Add new row
+            df = pd.concat([df, pd.DataFrame([row_data], columns=df.columns)])
+
+        return df
+
+    if Path(args['exportfilename']).exists():
```
xmatthias: Fails if `--export-filename` is not provided as argument.

```
2023-04-24 08:16:14,561 - freqtrade - ERROR - Fatal exception!                                                                              Traceback (most recent call last):                                                                                                          
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/main.py", line 42, in main                                                            
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/commands/strategy_utils_commands.py", line 108, in start_backtest_lookahead_bias_check
er                                                                                                                                          
    export_to_csv(args, bias_checker_instances)                                                                                             
  File "/home/xmatt/devel/cryptos/freqtrade/freqtrade/commands/strategy_utils_commands.py", line 158, in export_to_csv                
    if Path(args['exportfilename']).exists():                                                                                               
  File "/home/xmatt/.pyenv/versions/3.10.8/lib/python3.10/pathlib.py", line 960, in __new__              
    self = cls._from_parts(args)                                                                                                            
  File "/home/xmatt/.pyenv/versions/3.10.8/lib/python3.10/pathlib.py", line 594, in _from_parts                                             
    drv, root, parts = self._parse_args(args)                                                                                               
  File "/home/xmatt/.pyenv/versions/3.10.8/lib/python3.10/pathlib.py", line 578, in _parse_args                          
    a = os.fspath(a)                                                                                                                        
TypeError: expected str, bytes or os.PathLike object, not NoneType    
```

The probably correct way is to create a regular config object and use that - instead of using args - which is pretty difficult to handle reliably (and will ignore settings coming from a config).

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+
+        varHolder.indicators = self.backtesting.strategy.advise_all_indicators(varHolder.data)
+        varHolder.result = self.get_result(self.backtesting, varHolder.indicators)
+
+    def start(self, config, strategy_obj: dict, args) -> None:
+
+        # deepcopy so we can change the pairs for the 2ndary runs
+        # and not worry about another strategy to check after.
+        self.local_config = deepcopy(config)
+        self.local_config['strategy_list'] = [strategy_obj['name']]
```
xmatthias: By doing this, you implicitly enforce the user to specify `--timeframe xx` (because of how backtesting interprets this) - but for single strategy backtests, that'd not be necessary - so i suspect simply using `"strategy"` (and assignment not as list) might be a better idea.

freqtrade/commands/strategy_utils_commands.py
```diff
@@ -53,3 +57,135 @@ def start_conversion(strategy_obj, config):
...
+                        'biased_entry_signals': inst.current_analysis.false_entry_signals,
+                        'biased_exit_signals': inst.current_analysis.false_exit_signals,
+                        'biased_indicators': ",".join(inst.current_analysis.false_indicators)}
+        csv_df = add_or_update_row(csv_df, new_row_data)
+
+    print(f"saving {args['exportfilename']}")
+    csv_df.to_csv(args['exportfilename'], index=False)
+
+
+def initialize_single_lookahead_bias_checker(strategy_obj, config, args):
```
xmatthias: You will find that adding type-hints to function headers greatly helps reduce bugs (on all function headers, not just this one).

mypy can help identify potential problems - and code completion will work better, too - as the editor (if setup correctly) will be able to assist you, and suggest the correct variable name

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+
+        self.backtesting = Backtesting(prepare_data_config)
+        self.backtesting._set_strategy(self.backtesting.strategylist[0])
+        varHolder.data, varHolder.timerange = self.backtesting.load_bt_data()
+        self.backtesting.load_bt_data_detail()
+
+        varHolder.indicators = self.backtesting.strategy.advise_all_indicators(varHolder.data)
+        varHolder.result = self.get_result(self.backtesting, varHolder.indicators)
+
+    def start(self, config, strategy_obj: dict, args) -> None:
```
xmatthias: You may want to split this function up a bit (into multiple subfunctions, i'd suggest)

a function that's 100 lines long is ... VERY difficult to digest / read through - making it pretty difficult to determine what is happening at each step.

All variables are also instance variables (`self.xxx`) - so the actual scope of where (for example the loop) is running is not relevant - and can easily be extracted to a separate method (which may call other methods in the process).

freqtrade/strategy/backtest_lookahead_bias_checker.py
```diff
@@ -0,0 +1,252 @@
...
+                break
+
+            # if force-sold, ignore this signal since here it will unconditionally exit.
+            if result_row.close_date == self.dt_to_timestamp(self.full_varHolder.to_dt):
+                continue
+
+            self.current_analysis.total_signals += 1
+
+            self.entry_varHolder = VarHolder()
+            self.exit_varHolder = VarHolder()
```
xmatthias: You may want to have these as local variables instead. That'll make their actual scope clear (they are only valid within the loop).

xmatthias (commented):
> ps: currently I dont like the name of the command since it's way too long. If you have any suggestions then please go ahead :)

maybe try `lookahead-bias-checker` (or `lookahead-analysis`) instead? 
While it's using backtesting under the hood (so having backtestability IS a requirement) - i don't think we will need to explicitly mention this.

Overall, i do like how it works though - there's a few things to iron out (see exception above) - but it's a great functionality!

freqtrade/commands/strategy_utils_commands.py
hippocritical: resolved
--all-strategies would be dreadfully slow and tons of people would be complaining sometimes it taking multiple days depending on the amount I bet

freqtrade/commands/strategy_utils_commands.py
hippocritical: I would still want to keep the pre-filtering of the strategies so if they cannot be properly loaded by freqtrade they won't even reach the backtest bias checker

unified --strategy and --strategy_list into one list and that is gone through as you suggested.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: those variables get put there since pycharm else warns me about it 
"instance attribute defined outside __init__"

can remove it if you like, please advise

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: does this mean that only that file should be moved or the content of strategy_utils_commands should be moved over to somewhere else in addition?
(that file got moved already)

freqtrade/commands/strategy_utils_commands.py
hippocritical: sure, put the import line into the function itself

freqtrade/strategy/backtest_lookahead_bias_checker.py
xmatthias: well i'm not necessarily against these lines - but about the usage pattern this creates (this, in combination with whee they're overwritten).

i think you should pass config into the constructor - and fill these variables properly (sorry if that didn't come across ...).

freqtrade/strategy/backtest_lookahead_bias_checker.py
xmatthias: This file should definitely be moved. 

strategy_utils_commands will have another action taken to mitigate that (see comment there)

freqtrade/commands/strategy_utils_commands.py
hippocritical: I am confused, the arg exportfilename is not introduced by me but is just re-used.
ARGS_BACKTEST = ARGS_COMMON_OPTIMIZE + ["position_stacking", "use_max_market_positions",
                                        "enable_protections", "dry_run_wallet", "timeframe_detail",
                                        "strategy_list", "export", "exportfilename",
                                        "backtest_breakdown", "backtest_cache",
                                        "freqai_backtest_live_models"]

It was already available in other parts, so adding it in the config would make no sense now ?
(will better catch the errors there)

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: moved to utcnow, stating "now" only happens once and here.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: unique-id was a solution after a talk with drafty / your_secret_admirer on the freqai discord channel.
He suggested that solution. Here a link of the conversation:
https://discord.com/channels/700048804539400213/983285489299779645/1097244068561555607

If you have any better solution just say so, purging unique-id feels wrong to me but he said it's ok so I settled with his suggestion.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: switched to f-strings and logging.info
yes, f-strings are much better.
everywhere except from the the tabulate print in the end

freqtrade/commands/strategy_utils_commands.py
hippocritical: this won't work since the data as file-name, class-name etc are read by text_table_bias_checker
if you want .current_analysis only then these variables then the read variables should be moved but I don't see the point since the bulk of the data is current_analysis anyways.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: the only thing that is carried over between backtests and slice-backtests and between strategies is the result file for the outputs.
By design nothing should be allowed to even be moved between anything.
Every backtest or partial backtest should be made on a clean slate.

Until now the optimization of data-loading never was in the scope. Yes the data can be taken from the full timerange that is already loaded and just cut off instead of being loaded each backtest / partial backtest.

Looking into it.

freqtrade/commands/strategy_utils_commands.py
hippocritical: yeah only tested with --exportfilename. Next version checks for the variable and only then engages the function of saving into the csv file

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: switched to --strategy

freqtrade/commands/strategy_utils_commands.py
hippocritical: > > ps: currently I dont like the name of the command since it's way too long. If you have any suggestions then please go ahead :)
> 
> maybe try `lookahead-bias-checker` (or `lookahead-analysis`) instead? While it's using backtesting under the hood (so having backtestability IS a requirement) - i don't think we will need to explicitly mention this.
> 
> Overall, i do like how it works though - there's a few things to iron out (see exception above) - but it's a great functionality!

picked lookahead-analysis

hippocritical: > You will find that adding type-hints to function headers greatly helps reduce bugs (on all function headers, not just this one).
> 
> mypy can help identify potential problems - and code completion will work better, too - as the editor (if setup correctly) will be able to assist you, and suggest the correct variable name

sure, already doing it here and there but found it as "useful but optional"
will make it a requirement now.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: yeah sure

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: done

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: Moved everything to optimize_commands but didn't want to blow up the file with those secondary functions.
Shifted those to a sub-class of the file lookahead_analysis and static methods.

Found this to be a little cleaner than just stuffing everything into optimize commands.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: sure

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: Yes, moved the self.entry_varHolder and exit too only as self-reference of the self.entry_varHolders[idx]
The for loop will care for picking of which varHolders to use

Left         self.current_analysis.total_signals += 1 in place so even if something fails it reports the right lines out without any additional lines (it s just a +=1 so it really doesnt take any significant computational power even though when everything works ok it is some wasted cycles)

freqtrade/strategy/backtest_lookahead_bias_checker.py
xmatthias: my point is - we're taking 2 assumptions here.
* user-data is in `./user_data` (starting from current dir) - which it may or may not be.
* identifier is set to "unique-id"

The first is a (easily solved) problem. 
now for the 2nd, i'm not sure what the intend is (also not from the linked conversation) 

Are you trying to (intentionally) retrain models for each backtest? (i assume so - as the timerange will be different each time, and we'd like to start fresh).
if yes - then using a hardcoded identifier may work - but **only** under the assumption that you assign that same value to the config (which is not set).

therefore your "delete" may delete something non-existant (and more importantly, non-used) - as freqAI will store the models in `user_data/models/someotherIdentifier` - as `someotherIdentifier` is set as `identifier` in the freqAI config.

freqtrade/strategy/backtest_lookahead_bias_checker.py
xmatthias: you're not wrong in that approach given what the command aims to do.
it was just a suggestion / thought i had to speed things up slightly.

i'm fine with both "won't do" - as well as "let's try optimization at a later point in a sep. PR"

freqtrade/commands/strategy_utils_commands.py
xmatthias: it wasn't added by you, no - but if it's not provided, argparse won't guarantee it's there (it has no default).
The better / safer way here is probably to simply use the config object - where this setting IS set correctly.

hippocritical: I caught it properly so if it s not there it will not fire - and not catch on fire and burn down the house

Will look into how to set it into config

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: I've been looking into how to do it, but unsure how to not make it crash completely.

Would prefer to have this fixed in an "optimization PR" after the first one went through.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: > Are you trying to (intentionally) retrain models for each backtest? (i assume so - as the timerange will be different 

each time, and we'd like to start fresh).
yes, throw away anything old

Might be better that we force the user to define some ID they want instead of "unique-id"

freqtrade/commands/strategy_utils_commands.py
hippocritical: I currently check if it is there and only then read it - it does not produce errors that way, if that s ok with you

xmatthias: i think it's better to use "config" instead of args (config is parsed and available anyway).
that way, you'll get "automatic" handling of expected path modifications (like the default for export_filename if no explicit path is given, but a different user_dir is selected).

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: using config.user_data_dir now
using the config.freqai.identifier now

It looks like it defaults to "example" and still doesn't crash if the user doesn't define it.
The program will still try to purge the previous folder if exists.

freqtrade/strategy/backtest_lookahead_bias_checker.py
hippocritical: Tried, failed - would like to postpone this until the first version is done.

freqtrade/commands/strategy_utils_commands.py
hippocritical: sure, looking into it

hippocritical:
maybe it would be better if the dataframe would be deleted right after the conversion happened.
In the end after everything went through only the results matter - not the dataframe itself.

Assuming we have a list of 100s of strategies
(if you can do someone will do it)
then it would slowly blow up the RAM...

xmatthias:
> maybe it would be better if the dataframe would be deleted right after the conversion happened. In the end after everything went through only the results matter - not the dataframe itself.
> 
> Assuming we have a list of 100s of strategies (if you can do someone will do it) then it would slowly blow up the RAM...

for sure - but this again might be a "future optimization" point - it CAN be important - but is not necessarily crucial i'd think

hippocritical:
building tests is a hell hole currently, working on it on and off every now and then but have no motivation to do since I found myself inside dependency traps over and over ...

Could you lend me a hand with that? qq

xmatthias:
>I found myself inside dependency traps over and over 

Can you elaborate?

hippocritical:
I don't get how the heck to do a test with --minimum_trade_amount and --targeted_trade_amount to initialize it correctly
tried several times, all failures.
My issues generally are that I don't get how tests internally work and how to start a lookahead-analysis via tests properly.

Could you create the base line for a test for lookahead-analysis so it just starts working? I could then build on that and test the properties of the lookahead-analysis after that. Somehow it does not want to go into my head how to do the tests there.

hippocritical:
dunno why github thinks those additional files (which were from the main repo) are now edited by myself ... 
those shouldn't even be there to begin with since I just stated in pycharm "accept their version"

docker-compose.yml
```diff
@@ -6,15 +6,6 @@ services:
     # image: freqtradeorg/freqtrade:develop
     # Use plotting image
     # image: freqtradeorg/freqtrade:develop_plot
-    # # Enable GPU Image and GPU Resources (only relevant for freqAI)
```
xmatthias: This seems like a merge gone bad/wrong (accepting the wrong version).

It's reverting a change we did recently - and i assume it's the same case with several other files, too (this wasn't that big of a PR).

hippocritical: reverted to your state, should be fixed now. sorry

hippocritical:
Those tests should have all been done now, if you have any more requirements then please say so.

next in line: docs along with some gotchas and tips/tricks and how it works generally (like the picture in the very start, but updated)

hippocritical:
how to convey that you can use items from backtesting and freqai without showing all the info / args of those again?

Would a link to both pages be enough and just point to the added arguments ?

I would like to point to both backtesting and freqai (since you are allowed to lookahead-analyze both types of strategies) and then explain the quirks of the command itself.
Example: 
- how the lookahead analysis works
- what commands you should use as a minimum
- why --cache=none is enforced 
- why --targeted_trade_amount is enforced 
- why --minimum_trade_amount exists
- ...

hippocritical:
There is an edge case where we have constrained buys - since the 2ndary backtests (verifying each entry/exit) could be differing when only using one pair instead of all - it would be best to enforce the max_open_trades to the same amount as pairs.

This can not easily be conveyed to the user, will see what to work with best ... Else maybe users not aware how it works internally by 100% (meaning almost everyone) might geht false positives in terms of biased strategies.

This is truly an edge case but it could still happen - since with a constrained trade amount and the initial test the buy could be denied and then in a 2nd entry/exit run it could then be triggered.

xmatthias:
I'm not sure i agree on that being an edgecase.

it's a very common case - and i think what we want to test is best conveyed by setting max_open_trades to the pairlist length (obviously also considering dry-run-amount to not run out of money ...).
Doing that is more documentation than anything - but not something i'd see as problematic, really.

hippocritical:
will extend the docs in the next step and from a code base this should(?) be done
After that it would be nice if a native speaker checked spelling and everything else in the added part for the docs.
love the speed improvements btw ^.^

freqtrade/optimize/lookahead_analysis_helpers.py
```diff
@@ -119,25 +119,22 @@ def add_or_update_row(df, row_data):
...
-        current_instance.start()
-        elapsed = time.perf_counter() - start
-        logger.info(f"checking look ahead bias via backtests "
-                    f"of {Path(strategy_obj['location']).name} "
-                    f"took {elapsed:.0f} seconds.")
-        return current_instance
-
-    @staticmethod
-    def start(config: Config):
+    def calculate_config_overrides(config: Config):
```
xmatthias, line 123: You might wanna add an explicit test for this.
It's nice and isolated, which makes this a pretty easy test to write - but we'll want to ensure this remains correct for the future - as it's a critical piece of functionality to have the bias checker work correctly.

hippocritical: true, sure

hippocritical:
the tests that fail are not within the lookahead-analysis now

hippocritical:
@xmatthias dunno what's still missing, my local tests come back successfully but some tests fail on github - but some OS/Python combos go through flawlessly.

xmatthias:
almost certainly, they'll fail if you run them with the same `--random-order` value as the failing test.

I'll have a look sometime later this week (considering it's my log modifications / tests that cause the issue).

xmatthias (approved):
Great enhancement! :100: 

LGTM :+1:

PR #8383, opened 2023-03-25
Convert historic_predictions.pkl to parquet for performance

robcaulk:
## Summary

The goal of this PR is to alleviate saving/loading performance issues for expanding historic_predictions.pkl files on extended FreqAI dry runs. 

As this dataframe is growing indefinitely for the duration of a single `identifier`, it is clear that any active post processing will begin suffering more and more due to pkl loading performance slowdowns. By using Parquet to store this dataframe, we speed up saving time as well, which will reduce computational loads - saving cycles for more important components such as training/inference. 

We already have a large number of users who are running their own dry runs with a valuable `historic_predictions.pkl` growing in size. This PR *also* converts their pkl to a parquet without them having to do any additional work.

### DIFF
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -223,7 +223,8 @@ class FreqaiDataDrawer:
             value.to_parquet(filename)
 
         # create a backup
-        shutil.copytree(self.historic_predictions_folder, self.historic_predictions_bkp_folder)
+        shutil.copytree(self.historic_predictions_folder,
+                        self.historic_predictions_bkp_folder, dirs_exist_ok=True)
 
     def save_metric_tracker_to_disk(self):
         """
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -169,11 +169,7 @@ class FreqaiDataDrawer:
 
         if exists:
             try:
-                for file_path in self.historic_predictions_folder.glob("*.parquet"):
-                    key = file_path.stem
-                    key.replace("_", "/")
-                    self.historic_predictions[key] = pd.read_parquet(file_path)
-
+                self.load_historic_predictions_from_folder()
                 logger.info(
                     f"Found existing historic predictions at {self.full_path}, but beware "
                     "that statistics may be inaccurate if the bot has been offline for "
@@ -182,10 +178,7 @@ class FreqaiDataDrawer:
             except EOFError:
                 logger.warning(
                     'Historical prediction files were corrupted. Trying to load backup files.')
-                for file_path in self.historic_predictions_folder.glob("*.parquet"):
-                    key = file_path.stem
-                    key.replace("_", "/")
-                    self.historic_predictions[key] = pd.read_parquet(file_path)
+                self.load_historic_predictions_from_folder()
                 logger.warning('FreqAI successfully loaded the backup '
                                'historical predictions files.')
 
@@ -204,6 +197,18 @@ class FreqaiDataDrawer:
             )
 
         return exists
+
+    def load_historic_predictions_from_folder(self):
+        """
+        Try to build the historic_predictions dictionary from parquet
+        files in the historic_predictions_folder
+        """
+        for file_path in self.historic_predictions_folder.glob("*.parquet"):
+            key = file_path.stem
+            key.replace("_", "/")
+            self.historic_predictions[key] = pd.read_parquet(file_path)
+
+        return
 
     def save_historic_predictions_to_disk(self):
         """
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -75,8 +75,7 @@ class FreqaiDataDrawer:
         self.full_path = full_path
         self.historic_predictions_path = Path(self.full_path / "historic_predictions.pkl")
         self.historic_predictions_folder = Path(self.full_path / "historic_predictions")
-        self.historic_predictions_bkp_folder = Path(
-            self.full_path / "historic_predictions_backup")
+        self.historic_predictions_bkp_folder = Path(self.full_path / "historic_predictions_backup")
         self.pair_dictionary_path = Path(self.full_path / "pair_dictionary.json")
         self.global_metadata_path = Path(self.full_path / "global_metadata.json")
         self.metric_tracker_path = Path(self.full_path / "metric_tracker.json")
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -74,8 +74,10 @@ class FreqaiDataDrawer:
         self.historic_predictions: Dict[str, DataFrame] = {}
         self.full_path = full_path
         self.historic_predictions_path = Path(self.full_path / "historic_predictions.pkl")
-        self.historic_predictions_bkp_path = Path(
-            self.full_path / "historic_predictions.backup.pkl")
+        self.historic_predictions_path_parquet = Path(
+            self.full_path / "historic_predictions.parquet")
+        self.historic_predictions_bkp_path_parquet = Path(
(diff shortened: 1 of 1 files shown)


freqtrade/freqai/data_drawer.py
```diff
@@ -176,24 +179,34 @@ def load_historic_predictions_from_disk(self):
...
+                self.historic_predictions = pd.read_parquet(
+                    self.historic_predictions_bkp_path_parquet)
                 logger.warning('FreqAI successfully loaded the backup historical predictions file.')
 
+        elif not exists and convert:
+            logger.info("Converting your historic predictions pkl to parquet"
+                        "to improve performance.")
+            with Path.open(self.historic_predictions_path, "rb") as fp:
+                self.historic_predictions = cloudpickle.load(fp)
+            self.historic_predictions.to_parquet(self.historic_predictions_path_parquet)
```
xmatthias: what made you pick parquet over feather? (don't see this as critique / request for change)

based on my [tests](https://www.freqtrade.io/en/latest/data-download/#dataformat-comparison) - feather is both slightly smaller, as well as slightly faster ... (and the interface to it is effectively identical, except for a different function name (read_feather -> read_parquet, ...).

robcaulk: I read that Feather is not good for long-term storage. If you think they have alleviated that issue, then I would be perfectly happy to switch to feather.

freqtrade/freqai/data_drawer.py
```diff
@@ -163,37 +164,67 @@ def load_historic_predictions_from_disk(self):
...
                     "an extended period of time."
                 )
             except EOFError:
                 logger.warning(
-                    'Historical prediction file was corrupted. Trying to load backup file.')
-                with self.historic_predictions_bkp_path.open("rb") as fp:
-                    self.historic_predictions = cloudpickle.load(fp)
-                logger.warning('FreqAI successfully loaded the backup historical predictions file.')
+                    'Historical prediction files were corrupted. Trying to load backup files.')
+                self.load_historic_predictions_from_folder()
```
xmatthias, line 180: Nothing changed between the call in the try block and here - why would we expect loading to work now (it's not changing the path to a "backup dir" or something similar...)


that said - i think it might make sense to move the try/catch block into the loop within `load_historic_predictions_from_folder` instead.
if we assume a corrupted file - we might still be able to load the remaining files?

freqtrade/freqai/data_drawer.py
```diff
@@ -163,37 +164,67 @@ def load_historic_predictions_from_disk(self):
...
+                logger.warning('FreqAI successfully loaded the backup '
+                               'historical predictions files.')
+
+        elif not exists and convert:
+            logger.info("Converting your historic predictions pkl to parquet"
+                        "to improve performance.")
+            with Path.open(self.historic_predictions_path, "rb") as fp:
+                self.historic_predictions = cloudpickle.load(fp)
+            self.save_historic_predictions_to_disk()
```
xmatthias, line 189: would/should we delete the backup file after this step?
it's no longer written to - so once it's saved in the new format, it would stick around "forever" ... ? 
potentially even causing problems if the parquet files were reset - but this file forgotten ?

freqtrade/freqai/data_drawer.py
```diff
@@ -163,37 +164,67 @@ def load_historic_predictions_from_disk(self):
...
     def save_historic_predictions_to_disk(self):
         """
         Save historic predictions pickle to disk
         """
-        with self.historic_predictions_path.open("wb") as fp:
-            cloudpickle.dump(self.historic_predictions, fp, protocol=cloudpickle.DEFAULT_PROTOCOL)
+
+        self.historic_predictions_folder.mkdir(parents=True, exist_ok=True)
+        for key, value in self.historic_predictions.items():
+            key = key.replace("/", "_")
```
xmatthias, line 219: I'm not convinced this will work. 
what about futures pairs (on windows)? 

can't name a file `ETH_USDT:USDT` on windows - as `:` is reserved for dir letters.
Best use `pair_to_filename()` from `misc` instead - which will handle this safely.

freqtrade/freqai/data_drawer.py
```diff
@@ -163,37 +164,67 @@ def load_historic_predictions_from_disk(self):
...
         return exists
 
+    def load_historic_predictions_from_folder(self):
+        """
+        Try to build the historic_predictions dictionary from parquet
+        files in the historic_predictions_folder
+        """
+        for file_path in self.historic_predictions_folder.glob("*.parquet"):
+            key = file_path.stem
+            key.replace("_", "/")
```
xmatthias, line 207: same as below - but in reverse (`idatahandler.rebuild_pair_from_filename()` has logic for this ... but it's not readily usable iirc due to it's location, so would need either copy/paste or a small refactoring to extract it from that class, potentially into misc alongside the pair_to_filename method)

freqtrade/freqai/data_drawer.py
```diff
@@ -74,8 +74,9 @@ def __init__(self, full_path: Path, config: Config):
         self.historic_predictions: Dict[str, DataFrame] = {}
         self.full_path = full_path
         self.historic_predictions_path = Path(self.full_path / "historic_predictions.pkl")
-        self.historic_predictions_bkp_path = Path(
-            self.full_path / "historic_predictions.backup.pkl")
+        self.historic_predictions_folder = Path(self.full_path / "historic_predictions")
+        self.historic_predictions_bkp_folder = Path(
+            self.full_path / "historic_predictions_backup")
```
xmatthias: Nitpick: fits onto one line 

```suggestion
        self.historic_predictions_bkp_folder = Path(self.full_path / "historic_predictions_backup")
```

freqtrade/freqai/data_drawer.py
```diff
@@ -675,7 +706,7 @@ def get_timerange_from_live_historic_predictions(self) -> TimeRange:
         Returns timerange information based on historic predictions file
         :return: timerange calculated from saved live data
         """
-        if not self.historic_predictions_path.is_file():
+        if not self.historic_predictions_folder.exists():
```
xmatthias, line 708: `.exists()` or `.is_dir()`?

freqtrade/freqai/data_drawer.py
```diff
@@ -163,37 +164,67 @@ def load_historic_predictions_from_disk(self):
         Locate and load a previously saved historic predictions.
         :return: bool - whether or not the drawer was located
         """
-        exists = self.historic_predictions_path.is_file()
+        exists = self.historic_predictions_folder.exists()
```
xmatthias, line 166: `.is_dir()` ? exists returns true in both "is a file" or "is a directory" cases

freqtrade/freqai/data_drawer.py
robcaulk: Yeah good point, but I’m scared to do this as it is a very valuable file. Maybe we delete the main pkl but we leave the backup pkl, since the backup pkl won’t interfere with future operations.

freqtrade/freqai/data_drawer.py
robcaulk: This was a copy paste mistake, good catch.

freqtrade/freqai/data_drawer.py
robcaulk: Alright we can change it in this case.

freqtrade/freqai/data_drawer.py
xmatthias: probably yes - the way it is now i think it's quite dangerous that (for whatever reason) a new migration might happen - which would overwrite the new (more precious as containing more data) files ...

robcaulk:
I’m closing this because I’m not seeing the performance gains I was hoping to see in my testing. It therefore feels like an intrusive MR without the expected benefits.

I’ll be happy to reopen this if we want to explore it more in the future.

PR #8388, opened 2023-03-26
Bug fix: FreqAI backtest target setting

robcaulk:
## Bug

This PR fixes a backtesting bug found by @JohanVlugt and @samsam in the FreqAI discord. In brief, setting targets in backtesting does not permit using the `metadata["pair"]` in the column's fstring. 

## Cause 
We remove special chars ( the colon ":") from the dataframe columns before the sub train dataframes are split up and before the final targets are created for each sub train. 

## Fix
Remove the special chars from the sub train dataframes only, and do it *after* the targets are set, but *before* training. (All because LightGBM training doesnt like the special chars).

### DIFF
--- a/tests/freqai/test_freqai_interface.py
+++ b/tests/freqai/test_freqai_interface.py
@@ -432,10 +432,12 @@ def test_plot_feature_importance(mocker, freqai_conf):
     freqai = strategy.freqai
     freqai.live = True
     freqai.dk = FreqaiDataKitchen(freqai_conf)
+    freqai.dk.live = True
     timerange = TimeRange.parse_timerange("20180110-20180130")
     freqai.dd.load_all_pair_histories(timerange, freqai.dk)
 
-    freqai.dd.pair_dict = MagicMock()
+    freqai.dd.pair_dict = {"ADA/BTC": {"model_filename": "fake_name",
+                                       "trained_timestamp": 1, "data_path": "", "extras": {}}}
 
     data_load_timerange = TimeRange.parse_timerange("20180110-20180130")
     new_timerange = TimeRange.parse_timerange("20180120-20180130")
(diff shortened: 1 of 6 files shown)


wagnercosta (dismissed):
LGTM 👍

tests/freqai/test_freqai_interface.py
```diff
@@ -425,6 +432,7 @@ def test_plot_feature_importance(mocker, freqai_conf):
     freqai = strategy.freqai
     freqai.live = True
     freqai.dk = FreqaiDataKitchen(freqai_conf)
+    freqai.dk.live = True
```
xmatthias, line 435: seems like this change (or one in the actual files) causes the windows-test to fail as a mock object is used as path (causing an invalid path).
on linux/macos, that'll be fine as that accepts "almost everything" as valid path - but windows is a bit picky here.

xmatthias: image

clearly a test-issue ... 
the mock is set from load_data in line 510 from datadrawer  - where `dk.model_filename = self.pair_dict[coin]["model_filename"]` ... (and pair_dict is a magicmock, which is assigned right in that test.

Now it's clearly caused by the switch to "live = True" - otherwise this path is not called .... i'm not entirely sure what the actual fix is though, other than something VERY specific for this test (or not touching `pair_dict`) - neither of which i'm sure is a good idea.

xmatthias: Running that test on linux/mac also (somewhat) causes this - though it does not fail.

have a clean checkout ... it'll generate a new directory named `MagicMock` ... ;) 

image

xmatthias (approved):
LGTM :+1:

PR #8423, opened 2023-04-01
make trade_type value more explicit, add profit to trade_history dict

robcaulk:
The goal of this PR is to add useful information to the reinforcement learning environment `self.trade_history` list. This list is refreshed for each episode, so it can be used inside the `calculate_reward()` function to evaluate all the trades that were made during any particular episode. 

We also remove docstrings in favor of people reading code. These kinds of docstrings are outdated quickly (as we saw), and I prefer not to maintain them.


xmatthias (approved):
LGTM :+1:

PR #8426, opened 2023-04-02
Add sb3 learn progress bar

initrv:
## Summary

The purpose of this PR is to provide the user with progress bar functionality for stable-baselines3 and sb3-contrib out of the box
https://stable-baselines3.readthedocs.io/en/master/guide/callbacks.html#progressbarcallback

## Quick changelog

- `progress_bar` param for rl_config
- The `tqdm` and `rich` libraries have been added to the requirements

## What's new?

 `8% ━━━╺━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 7,918/96,736  [ 0:02:15 < 0:19:29 , 76 it/s ]`


requirements-freqai-rl.txt
```diff
@@ -8,3 +8,6 @@ sb3-contrib==1.7.0; python_version < '3.11'
 # Gym is forced to this version by stable-baselines3.
 setuptools==65.5.1 # Should be removed when gym is fixed.
 gym==0.21; python_version < '3.11'
+# Progress bar for stable-baselines3 and sb3-contrib
+tqdm==4.65.0; python_version < '3.11'
+rich==13.3.3; python_version < '3.11'
```
xmatthias: Why both? 

rich does provide a progress bar [on it's own](https://rich.readthedocs.io/en/stable/progress.html) - so this seems pretty much redundant.

If stable baselines requires both, please open an issue / bug report against them (i consider this a bug, adding unnecessary dependencies for no good reason).

Freqtrade also depends on `progressbar2` - and i REALLY don't see freqtrade depending on 3 different progressbar providers for no good reason - so this will need to change (at least based on currently available information).

freqtrade/freqai/prediction_models/ReinforcementLearner.py
```diff
@@ -71,7 +71,8 @@ def fit(self, data_dictionary: Dict[str, Any], dk: FreqaiDataKitchen, **kwargs):
 
         model.learn(
             total_timesteps=int(total_timesteps),
-            callback=[self.eval_callback, self.tensorboard_callback]
+            callback=[self.eval_callback, self.tensorboard_callback],
+            progress_bar=self.rl_config.get('progress_bar', False)
```
xmatthias, line 75: We do use progressbars in freqtrade in other places - and i know that it tends to mess up regular output (and requires quite some messing around to not break regular output).

which output (stdout / stderr, ...) is this outputting?

How will it impact actual logfiles (`--log-file somelog.log`)?

What im mainly looking for here is 
* Is the progressbar included in the logfile?
* Will the log formatting break afterwards or while the progressbar runs (with output between progress bar steps) (for the different providers like console, docker console and file output)

xmatthias (commented):
I don't like us adding 2 (!!) progressbar providers  for one feature.
(we already have one - why can't progressbar2 be used??)
I'm fine with adding ONE of the 2 dependencies (preferably rich) - but definitely not both.

If stable-baselines requires both - then that's a non-thought out feature to me  (or rather, a bug in stable-baselines) - and i don't like adding alpha features (unless stable-baselines can provide some very good explanation for this technical debt).

requirements-freqai-rl.txt
initrv: Unfortunately yes, both libraries are required. it looks like sb3 uses `tqdm`, which uses `rich` to make the progress bar "richer". Here is the Import and ImportError from sb3:

```python
from tqdm.rich import tqdm
...
if tqdm is None:
    raise ImportError(
        "You must install tqdm and rich in order to use the progress bar callback. "
        "It is included if you install stable-baselines with the extra packages: "
        "`pip install stable-baselines3[extra]`"
    )
```
https://tqdm.github.io/docs/rich/
https://github.com/DLR-RM/stable-baselines3/blob/470771b5c2142f0813b6bb1afb9143c960bf09a0/stable_baselines3/common/callbacks.py#L16

xmatthias: I don't really see why it wouldn't be possible to use rich directly (see linked documentation above).

Seems like a bug to me - or rather non-thoughtful usage of dependencies.
please report that to sb3. 
If any, it'll allow us to dump one of the libraries soon, otherwise, i hope to get some good explanation as to why they think it's a good idea to use rich through a different library (which doesn't seem to make sense to me).

freqtrade/freqai/prediction_models/ReinforcementLearner.py
initrv: The progress bar has no impact on the log files using `--log-file`.
All ft messages are displayed correctly, it seems to me.
But I have only checked linux native console output. 

However, when new messages arrive, the progress bar leaves its old copy on the screen.
It's not a big problem for me, but I can understand if this is unacceptable to freqtrade.

See screenshot:
scr

xmatthias: this is unfortunately mostly what i expected, and makes it pretty much useless if you ask me.
There's clearly ways around it - so i think to have this accepted, we'd want this to behave properly (basically similar / the same way as it does with hyperopt).

Leonardbrett:
> I don't like us adding 2 (!!) progressbar providers  for one feature.
> (we already have one - why can't progressbar2 be used??)
> I'm fine with adding ONE of the 2 dependencies (preferably rich) - but definitely not both.
> 
> If stable-baselines requires both - then that's a non-thought out feature to me  (or rather, a bug in stable-baselines) - and i don't like adding alpha features (unless stable-baselines can provide some very good explanation for this technical debt).

This woman offers fast and legit investment. Connect with Charlotte on 🅦🅗🅐🅣🅢🅐🅐🅟+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

Leonardbrett:
My first investment with Charlotte give me profits of over $7,000 and ever since then she has never failed to deliver and I can even say she's the must sincere broker I have ever known'' communicate with her to be a successful investor. On WhatsApp:+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

freqtrade/freqai/prediction_models/ReinforcementLearner.py
xmatthias: inspired by this (rich progressbars just look so much nicer) , i've created #8461 - using rich as progressbar for hyperopt.

Some of the changes that required also impacted logging compatibility - which i suspect MAY benefit this PR as well.

@initrv, Would you be so kind to re-test the progressbar behavior with the updated PR (I've pushed a merge from develop for this) - basically doing the same test than in the screenshot above.
I'd hope that the progressbar now stays "at the bottom" - no longer re-rendering due to log messages (unless you press random keys ... ) - but didn't test that so far.

initrv: That's great, thanks @xmatthias ! I confirm that the progress bar now works much better and stays at the bottom until training finished. As long as you don't resize the terminal window or press keys :)

xmatthias: The resizing is a ... different problem - but honestly, not one i care about :laughing:

Same with pressing keys in a "read only" terminal - shoot yourself in the foot? sure, go on - we won't stop you :laughing:

freqtrade/constants.py
```diff
@@ -600,6 +600,7 @@
                         "policy_type": {"type": "string", "default": "MlpPolicy"},
                         "net_arch": {"type": "array", "default": [128, 128]},
                         "randomize_starting_position": {"type": "boolean", "default": False},
+                        "progress_bar": {"type": "boolean", "default": False},
```
xmatthias: should we default this to true? 
it seems to be a nice feature. if someone is bothered by it, stopping it would still be an option ? 
just a thought ... i'm fine with it either way

xmatthias:
@robcaulk your opinion? 

Technically, i'm fine with this as it is now - though i'm not sure on the default of "progress_bar" - i think having this as true might be beneficial and capture a wider usergroup "automatically" ? (we can always change this later, though).
People not wanting this can still disable it, then (so we don't force it onto users).

xmatthias (dismissed):

freqtrade/constants.py
robcaulk: I would vote for defaulting to true, I will make the change

robcaulk:
> @robcaulk your opinion?
> 
> Technically, i'm fine with this as it is now - though i'm not sure on the default of "progress_bar" - i think having this as true might be beneficial and capture a wider usergroup "automatically" ? (we can always change this later, though). People not wanting this can still disable it, then (so we don't force it onto users).

I like this, thanks for putting that effort in, users will find this helpful for tracking large trainings.

robcaulk (approved):

PR #8427, opened 2023-04-02
Typo fix constants

initrv:
## Summary

Typo fix constants `randomize_startinng_position` -> `randomize_starting_position`


xmatthias (approved):
Great eye for detail! :+1: 

I wonder how one would discover such a typo - as this is not necessarily an area of the code where you'd read through on a daily basis ... :laughing:

PR #8459, opened 2023-04-08
Add initial bot start time to /profit endpoint 

### LINKED ISSUE #2885: Feature/enhancement- adding bot uptime.
I would like to suggest a feature, adding bot uptime to be viewed via telegram alerts  it would reset if the bot is restarted, with the `/stats` option as suggested by @xmatthias .

e.g if my strategy was based on a higher time frame maybe 4h, and the market is currently going against the rules, so it would just not trade for days or maybe a week, in that case i would want to know just how long the bot is up (and if i need to re-think my strategy)
same goes if my strategy has too many entry conditions

xmatthias:
## Summary
If a bot is not trading stright away, it's currently impossible to determine it's actual start time.
we can simply record this date (will default to the first trade open date) - so we have this correct in the future.

closes #2885

## Quick changelog

- Add general purpose, bot-wide, persistent key value store
- add functionality to record the bot start time
- add bot start time to /profit output

### DIFF
--- a/docs/telegram-usage.md
+++ b/docs/telegram-usage.md
@@ -279,6 +279,7 @@ Return a summary of your profit/loss and performance.
 >   ∙ `33.095 EUR`
 >
 > **Total Trade Count:** `138`
+> **Bot started:** `2022-07-11 18:40:44`
 > **First Trade opened:** `3 days ago`
 > **Latest Trade opened:** `2 minutes ago`
 > **Avg. Duration:** `2:33:45`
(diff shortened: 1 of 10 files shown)


Leonardbrett:
My first investment with Charlotte give me profits of over $7,000 and ever since then she has never failed to deliver and I can even say she's the must sincere broker I have ever known'' communicate with her to be a successful investor. On WhatsApp:+𝟭𝟰𝟳𝟱𝟮𝟰𝟮𝟲𝟯𝟯𝟴

PR #8491, opened 2023-04-12
Remotepairlist - fix continous fetching every x bot_loop seconds

Bloodhunter4rc:
## Summary
When an empty pairlist is returned from remote pairlist will fetch the pairlist every bot_loop_seconds due to empty cache.

## What's new?
Store a dummy value in cache, to adhere to the refresh_period

### DIFF
--- a/freqtrade/plugins/pairlist/RemotePairList.py
+++ b/freqtrade/plugins/pairlist/RemotePairList.py
@@ -142,9 +142,9 @@ class RemotePairList(IPairList):
         """
 
         if self._init_done:
-            if self._pair_cache.get('pairlist') != [None]:
-                pairlist = self._pair_cache.get('pairlist')
-            else:
+            pairlist = self._pair_cache.get('pairlist')
+            if pairlist == [None]:
+                # Valid but empty pairlist.
                 return []
         else:
             pairlist = []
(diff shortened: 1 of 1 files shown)


freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -142,7 +142,10 @@ def gen_pairlist(self, tickers: Tickers) -> List[str]:
         """
 
         if self._init_done:
-            pairlist = self._pair_cache.get('pairlist')
+            if self._pair_cache.get('pairlist') == ["Empty"]:
```
xmatthias: Can you maybe explain (in more detail) what you're trying to actually fix?
adding a dummy item seems ... odd (and potentially dangerous/error-prone)

maybe using None instead (type-safety might help in that case) could also work?

Bloodhunter4rc: when a empty pairlist is returned for example: {"pairs":[],"refresh_period":1800} , it will keep fetching the pairlist from the url every bot_loop_seconds (default 5) due to the cache being empty

if it is set to "None" the cache wont reset, i need a cache item for a empty pairlist to still use the caching functionality, i can change it to List[None], that would work, if thats any better? In the End we return straight away if its [None] or ["Empty"] the same happens, just thats then a list with None

xmatthias: i think i got it
I've simplified the code a bit, and it'll also avoid a stupid race-condition if cache expires right between the 2 calls (which i DID ran into before - and therefore know we must avoid :laughing: )

xmatthias (approved):
LGTM :+1:

PR #8494, opened 2023-04-13
Bug fix: ensure data is on same device as model

robcaulk:
@smarmau reported a bug in discord saying that the data was not properly located to the same device as the model. This PR attempts to ameliorate this issue.


xmatthias (approved):
I'll approve this as conceptually, it looks correct.
However, I have no actual possibility to test this due to lack of appropriate GPU.

Please get additional feedback from @smarmau (or someone else who can actively test this) if this is fixes it before merging, thanks.

PR #8499, opened 2023-04-15
Telegram. Fixed the blacklist removal message

bkamuz:
## Summary

There is an issue with an error message when removing a pair from the blacklist that is not on the blacklist. Example below:

```
/bl_delete BTC/*

Error adding BTC/* to blacklist: Pair BTC/* is not in the current blacklist.

Blacklist contains 1 pairs
BTC/USDT

```

## Quick changelog

- `action` argument added to the `send_blacklist_msg()` function

## What's new?

The necessary strings have been added to get the correct answer in the telegram. Now answer looks like:
`Error deleting BTC/* from blacklist: Pair BTC/* is not in the current blacklist.`

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -1413,12 +1413,12 @@ class Telegram(RPCHandler):
         Handler for /blacklist
         Shows the currently active blacklist
         """
-        self.send_blacklist_msg(self._rpc._rpc_blacklist(context.args))
+        self.send_blacklist_msg(self._rpc._rpc_blacklist(context.args), ['adding', 'to'])
 
-    def send_blacklist_msg(self, blacklist: Dict):
+    def send_blacklist_msg(self, blacklist: Dict, action):
         errmsgs = []
         for pair, error in blacklist['errors'].items():
-            errmsgs.append(f"Error adding `{pair}` to blacklist: `{error['error_msg']}`")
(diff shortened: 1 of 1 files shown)


freqtrade/rpc/telegram.py
```diff
@@ -1413,12 +1413,13 @@ def _blacklist(self, update: Update, context: CallbackContext) -> None:
         Handler for /blacklist
         Shows the currently active blacklist
         """
-        self.send_blacklist_msg(self._rpc._rpc_blacklist(context.args))
+        self.send_blacklist_msg(self._rpc._rpc_blacklist(context.args), 1)
 
-    def send_blacklist_msg(self, blacklist: Dict):
+    def send_blacklist_msg(self, blacklist: Dict, des):
```
xmatthias: Please don't use non-typed (and more importantly, non-descriptive) variables.

xmatthias (changes requested):
First off (for next time) - you shouldn't use your develop branch for PR's.
That's bad practice, and can lead to some subtle errors down the road.

Now for this PR - i think it's okish (shouldn't take too long - so we can work around "PR from main branch" issues.

Though please don't use anonymous (variable name not saying what it's supposed to do) and wrongly typed variables. 

Using an integer here can have odd side-effects - what happens in the method if i pass in a "2" ?

xmatthias (approved):
LGTM 👍

Issue #8515, opened 2023-04-17
BadRequest Exception when Loading Leverage Tiers for GLMR/USDT:USDT on OKX Exchange


wassertim:
Describe your environment
Operating system: ubuntu
Python Version: 3.10.10
CCXT version: 3.0.23
Freqtrade Version: freqtrade docker-2023.3.dev-298f5685

Describe the problem:
I'm receiving "Could not load leverage tiers for GLMR/USDT:USDT due to BadRequest" exception when running a strategy with the OKX exchange.

Steps to reproduce:

- Set up a strategy with the OKX exchange, trading mode "futures" and margin mode isolated
- Run the strategy
- Observe the BadRequest exception in logs

Observed Results:

- The bot fails to load leverage tiers for the GLMR/USDT:USDT market on OKX exchange, raising a BadRequest exception.
- I expected the bot to successfully load leverage tiers and continue its operation.

Relevant code exceptions or logs
```
2023-04-17 15:44:36,828 - freqtrade.exchange.exchange - INFO - Initializing leverage_tiers for 127 markets. This will take about a minute.
2023-04-17 15:44:37,285 - freqtrade.exchange.common - WARNING - get_market_leverage_tiers() returned exception: "Could not load leverage tiers for GLMR/USDT:USDT due to BadRequest. Message: okx {"code":"51000","data":[],"msg":"Parameter uly  error"}". Retrying still for 4 times.
2023-04-17 15:44:37,832 - freqtrade.exchange.common - WARNING - get_market_leverage_tiers() returned exception: "Could not load leverage tiers for GLMR/USDT:USDT due to BadRequest. Message: okx {"code":"51000","data":[],"msg":"Parameter uly  error"}". Retrying still for 3 times.
2023-04-17 15:44:38,393 - freqtrade.exchange.common - WARNING - get_market_leverage_tiers() returned exception: "Could not load leverage tiers for GLMR/USDT:USDT due to BadRequest. Message: okx {"code":"51000","data":[],"msg":"Parameter uly  error"}". Retrying still for 2 times.
2023-04-17 15:44:39,030 - freqtrade.exchange.common - WARNING - get_market_leverage_tiers() returned exception: "Could not load leverage tiers for GLMR/USDT:USDT due to BadRequest. Message: okx {"code":"51000","data":[],"msg":"Parameter uly  error"}". Retrying still for 1 times.
[...]

PR #8516, opened 2023-04-17
Implement blocklist for failed leverage #8515

wassertim:
## Summary

Solve the issue: #8515

## Quick changelog

- Add a check for failed leverage symbols in exchange config

## What's new?

In this PR I try to overcome the bug I reported earlier, when running a strategy on OKX futures, which tries to load an incorrect symbol. This fixed the problem for me, but I don't know if it is the best way to do it. If there are any suggestions, I would be happy to apply them in this PR.

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -2374,6 +2374,8 @@ class Exchange:
                 symbols = []
 
                 for symbol, market in markets.items():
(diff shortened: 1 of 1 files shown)


xmatthias:
I appreciate your  involvement in this (and in the project).

This is however not the right approach for this problem. 
It'll not scale (should their API show further differences like this one), and make everyone's bot fail with the next random pair they add in mistake.

I'll have a proper fix for this ready soon.

PR #8521, opened 2023-04-18
docs: use helper function `stoploss_from_absolute` in strategy callba…

TheJoeSchr:
…cks "absolute" example
  ## Summary

Use "official" helper function instead of "own" formula so if something changes example have higher changes to still work by using the maintained method via `stoploss_from_absolute` helper function.

`def stoploss_from_absolute()` also has a `is_short` argument, which I incorporated.

But docs say, it doesn't matter if return value is positive or negative:
>The absolute value of the return value is used (the sign is ignored), so returning 0.05 or -0.05 have the same result, a stoploss 5% below the current price.

and all that `is_short` does is this:
```py
    if is_short:
        stoploss = -stoploss
```
so maybe it's not even needed.

### DIFF
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -352,7 +352,7 @@ class AwesomeStrategy(IStrategy):
 
         # Convert absolute price to percentage relative to current_rate
         if stoploss_price < current_rate:
-            return (stoploss_price / current_rate) - 1
+            return stoploss_from_absolute(stoploss_price, current_rate, is_short=trade.is_short))
 
         # return maximum stoploss value, keeping current stoploss price unchanged
         return 1
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -352,7 +352,7 @@ class AwesomeStrategy(IStrategy):
 
         # Convert absolute price to percentage relative to current_rate
         if stoploss_price < current_rate:
-            return stoploss_from_absolute(stoploss_price, current_rate, is_short=trade.is_short))
+            return stoploss_from_absolute(stoploss_price, current_rate, is_short=trade.is_short)
 
         # return maximum stoploss value, keeping current stoploss price unchanged
         return 1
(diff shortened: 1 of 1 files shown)


TheJoeSchr:
also just realized the example above is using `leverage` with `stoploss_from_open()` call which doesn't exist (anymore). so decided to fix that as well

xmatthias:
> also just realized the example above is using `leverage` with `stoploss_from_open()` call which doesn't exist (anymore). so decided to fix that as well

Not sure where you got that from - but it's wrong.
not providing leverage to this function will cause wrong behavior, resulting in wrongly placed stoplosses.

docs/strategy-callbacks.md
```diff
@@ -318,11 +318,11 @@ class AwesomeStrategy(IStrategy):
 
         # evaluate highest to lowest, so that highest possible stop is used
         if current_profit > 0.40:
-            return stoploss_from_open(0.25, current_profit, is_short=trade.is_short, leverage=trade.leverage)
+            return stoploss_from_open(0.25, current_profit, is_short=trade.is_short)
```
xmatthias: Why would we remove this? it's in the function and people SHOULD use it if they use leverage (and want to avoid problems). It also doesn't hurt non-leveraged trades - as the value will be 1.

So i don't see why we should remove it from the docs -  the "new" docs are now misleading ... 

https://github.com/freqtrade/freqtrade/blob/caf524c68584995e5ef6e4d2ba165ca4279e264f/freqtrade/strategy/strategy_helper.py#L86-L91

xmatthias (changes requested):

docs/strategy-callbacks.md
TheJoeSchr: ah, seems I checked on the wrong branch which wasn't rebased on current dev. sorry

docs/strategy-callbacks.md
```diff
@@ -352,7 +352,7 @@ class AwesomeStrategy(IStrategy):
 
         # Convert absolute price to percentage relative to current_rate
         if stoploss_price < current_rate:
-            return (stoploss_price / current_rate) - 1
+            return stoploss_from_absolute(stoploss_price, current_rate, is_short=trade.is_short))
```
xmatthias:
```suggestion
            return stoploss_from_absolute(stoploss_price, current_rate, is_short=trade.is_short)
```

xmatthias (approved):
:+1:

PR #8533, opened 2023-04-21
Create pylint.yml

MFG1PIKACHU0408:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,23 @@
+name: Pylint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10"]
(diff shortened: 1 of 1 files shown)


xmatthias (changes requested):
We're not using pylint - so why would we need CI for that?

personally, i find it way too aggressive / picky. 
I'm sure it can be fined tuned to work work better - but that's currently not the case for this repository. The combination of ruff/flake8, in combination with mypy is sufficient in my eyes, and i don't see what pylint can bring on top of this.

This can sure be argued for and against, so I'm curious about what your reasoning is to bring pylint to the project.

---

On top of that - we'd want PR's to be directed against the repository's main branch - which is NOT stable.
It should also not be done from the main branch of your repo, but from a feature branch.

xmatthias:
Currently no plans on adding pylint.

PR #8565, opened 2023-04-28
GPU Enable in docker-compose

vinistation:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


robcaulk:
Thanks for this :)

For the docker compose, I think we should make a second file that is just for freqai_rl/torch and has the gpu stuff inside of it. But for that we will also need comment from @xmatthias, he probably has some specific setup he’d prefer with regards to docker compose.

docker-compose.yml
```diff
@@ -26,3 +26,12 @@ services:
       --db-url sqlite:////freqtrade/user_data/tradesv3.sqlite
       --config /freqtrade/user_data/config.json
       --strategy SampleStrategy
+    # Enable GPU Image and GPU Resources
+    #image: freqtradeorg/freqtrade:develop_freqaitorch
```
xmatthias: Might make sense to move to the top (with top being somewhere in the middle, next to image)? below the command it's mostly missleading, and will probably lead to a lot of "oversight" questions, where 2 `image:` lines are still active, as they're not next to each other.

xmatthias (commented):
For future PR's - please create a feature branch in your fork when doing a PR. Doing PR's from the fork's develop branch is not good practice, and has some odd side-effects sometimes.

---

i'm "mostly" fine with both approaches (comment in the current file vs. new file) - but if we opt for a new file, please put it into the `docker/` directory.

The new file might give us some more "freqAI-preconfig" options as well - but as said, fine with both approaches

Aside from the position of this segment - would it make sense to also add some hint about this to the docs so people are aware this actually exists and is possible without reading through the compose file ... ?

robcaulk:
The title of this PR is superceded by #8580 . That means that the denormalization is now properly handled in develop.

However, this GPU integration to the docker file is still open and waiting to be handled. I am changing the title of this PR accordingly.

freqtrade/freqai/base_models/BasePyTorchRegressor.py
```diff
@@ -47,4 +47,5 @@ def predict(
         y = self.model.model(x)
         y = y.cpu()
         pred_df = DataFrame(y.detach().numpy(), columns=[dk.label_list[0]])
+        pred_df = dk.denormalize_labels_from_metadata(pred_df)
```
xmatthias, line 50: > That means that the denormalization is now properly handled in develop.

@robcaulk does this mean this line should be reverted, right?
The merge conflict is exactly on this line, too).

robcaulk: Yeah it’s a just a resolution on the merge conflict. They should look the same so just keep incoming.

robcaulk (approved):
LGTM

PR #8567, opened 2023-04-30
sexy logging in console

LinXueyuanStdio:
## Summary

beatuiful logging style

## Quick changelog

- add ColoerFormatter
- configure console_handler

## What's new?
截屏2023-04-30 12 04 05

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
         additional_dependencies:
           - types-cachetools==5.3.0.5
           - types-filelock==3.2.7
-          - types-requests==2.28.11.17
+          - types-requests==2.29.0.0
           - types-tabulate==0.9.0.2
           - types-python-dateutil==2.8.19.12
           - SQLAlchemy==2.0.12


robcaulk:
Hello,

Maybe you could specify the weak points of the code that this PR attempts to improve upon. It seems you were dissatisfied with error handling, so you added a new log to help. But it’s not clear to me what was insufficient about the “old” logging/trace back tracking. Can you help us to understand your rationale here?

Thanks,

rob

LinXueyuanStdio:
I'm so sorry that the second commit is pushed by mistake....

LinXueyuanStdio:
OK. Finally I revert and squash to one commit.

freqtrade/loggers/__init__.py
```diff
@@ -5,12 +5,16 @@
...
 from freqtrade.loggers.buffering_handler import FTBufferingHandler
-from freqtrade.loggers.std_err_stream_handler import FTStdErrStreamHandler
+from freqtrade.loggers.std_err_stream_handler import FTStdErrStreamHandler, ColorFormatter
 
 
 logger = logging.getLogger(__name__)
 LOGFORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
 
+console_handler = FTStdErrStreamHandler()
+console_handler.setLevel(logging.DEBUG)
```
xmatthias: This is clearly a no-go.

we'll not hard-code logging to debug. Verbosity levels will be comming via `-v` flag provided.

freqtrade/loggers/std_err_stream_handler.py
```diff
@@ -1,7 +1,64 @@
...
+
+    blue = "\x1b[34m"
+    cyan = "\x1b[36;21m"
+    green = "\x1b[32;21m"
+    orange = "\x1b[33;21m"
+    grey = "\x1b[38;21m"
+    yellow = "\x1b[33;21m"
+    red = "\x1b[31;21m"
+    bold_red = "\x1b[31;1m"
+    reset = "\x1b[0m"
```
xmatthias: In my experience, statically defined color-codes are not working reliably. Some consoles have difficulty rendering colored lines, which will mean uses or these consoles have broken output with no way to fix it.

i also don't see a way to actually disable this - which means we force it onto users, who may have color blindness disabilities - or other problems with funky looking colors.

---

how reliable is this cross-console? (mac, linux, windows (powershell and cmd and new terminal), docker containers) ??

Can you please provide screenshots showing the above scenarios are working correctly

xmatthias (changes requested):
I'm not per se against color logging - but i'm not sure if this is the right way.

First off - it doesn't look like in the screenshot, but most texts have an underline now (which is unnecessary, and makes it hard to read, as now almost everything contains an underline).

image

I also don't find this style "beautiful" - but actually rather ugly. 
unless the style becomes WAY more beautiful / less intrusive, please see this as complete rejection of this PR (it MAY have potential - but i'm not sure on the approach taken, either (see other comment).

There's a few general issues with the theme (most likely not complete)

* Log level is unclear (it's somehow coded into the colors - but it's absolutely unclear what level a message is).
* Unnecessary underlines of both date and text makes  no sense and makes it harder to read (as well as having it seem like a link).
* Color selection is odd (at best) - why is the date always orange - which also corresponds to the warning level? is every date then a warning??
* It's showing file name instead of the module (which we show intentionally)

PR #8580, opened 2023-05-01
Add transformer to FreqAI

robcaulk:
The goal of this PR is to provide FreqAI users with the option of running a transformer with positional encoding on their feature sets. 

One key difference between this model and all the other models in FreqAI is that it does *not* use `include_shifted_candles`, instead we pass a 3D tensor to the transformer. This means that users who were using `include_shifted_candles: 5` before, would now start using `conv_width: 5` to achieve the same input parameter space. 

Besides this difference, the `PyTorchTransformer` is also capable of handling multiple targets in a single model. Although we do have this ability already in FreqAI with our `*MultiTarget` variants, those are simply building `n` models under the hood. PyTorchTransformer is training one model on `n` outputs, which may be beneficial for performance, but could also hinder the prediction if one of the easy-to-fit targets if another target is difficult to fit. It is on the user to decide if this is appropriate or not for their custom feature engineering. 

Running the transformer is the same as all other FreqAI models:

```
freqtrade trade --strategy FreqaiExampleStrategy --freqaimodel PyTorchTransformerRegressor --config config.json
```


requirements-freqai-rl.txt
```diff
@@ -2,7 +2,7 @@
 -r requirements-freqai.txt
 
 # Required for freqai-rl
-torch==1.13.1; python_version < '3.11'
+torch==2.0.0; python_version < '3.11'
```
xmatthias: wasn't torch 2.0 supposed to be adding support for 3.11? 

Might be worth a try if it installs and works properly on 3.11 (basically removing the constraint in the requirements file)?

richardjozsa: I use python 3.11 and pytorch 2.0.0, and pandas 2.0.0 in my latest test setups. It indeed works properly, but I have not installed normal Freqai requirements like xgboost and catboost etc, because I'm into RL, but python 3.11 and pytorch 2.0.0 works perfectly. :)

richardjozsa:
This is pretty good!! I have had the same idea to implement transformers and self attention layers, just haven't had time to do it. Thank you for it. :)

freqtrade/freqai/prediction_models/PyTorchTransformerRegressor.py
```diff
@@ -0,0 +1,139 @@
...
+
+        n_features = data_dictionary["train_features"].shape[-1]
+        n_labels = data_dictionary["train_labels"].shape[-1]
+        model = PyTorchTransformerModel(
+            input_dim=n_features,
+            output_dim=n_labels,
+            time_window=self.window_size,
+            **self.model_kwargs
+        )
+        model.to(self.device)
```
xmatthias, line 75: I noticed your post about this in discord ...

i assume it will make sense to add the hint about `torch.compile(model)` to the docs ... 
It's a good hint - and something regular users will not be aware without having it documented someplace.

robcaulk: Good point, I added this to the documentation.

requirements-freqai-rl.txt
robcaulk: I also have tested with 3.11 + pytorch 2.0.0. Did not notice any issues.

I removed the constraint from the requirements.txt.

xmatthias (approved):
LGTM :+1: 

Not sure if you (or the beta-testers) want/need to do further tests) - feel free to merge when you think it's fully ready.

xmatthias (approved):
LGTM :+1: 

Not sure if you (or the beta-testers) want/need to do further tests - feel free to merge when you think it's fully ready.

PR #8587, opened 2023-05-02
Remove dependency workarounds in place for gym

xmatthias:
## Summary

Remove workarounds we had to put in place to support the outdated gym dependency.

with #8336 merged, these should no longer be necessary (ci will tell for certain).


robcaulk (approved):
LGTM

PR #8589, opened 2023-05-02
Add missing dependencies to setup.py

alxtrkhv:
## Summary

Discrepancy between `requirements.txt` and `setup.py` broke pypi package (2023.4 at least), `pip install -e .`, `poetry add`, etc

Solve the issue: None

## Quick changelog

- Added missing dependencies from `requirements*.txt` files to `setup.py`

## What's new?

`pip install -e .` works like a charm, presumably pypi package works as well.

Hi!

I'm quite new to this project and haven't managed to fully wrap my mind around it yet so was uncertain about any constraints required for those packages and took the most straightforward approach just to add missing packages as is. 

`pip install packaging` and `pip install ast-comments` was enough to get a help message from the cli but I'm not sure if that's truly enough.

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,9 +21,11 @@ freqai = [
 
 freqai_rl = [
     'torch',
+    'gymnasium',
     'stable-baselines3',
-    'gym==0.21',
-    'sb3-contrib'
+    'sb3-contrib',
+    'setuptools',
+    'tqdm'
 ]
 
 hdf5 = [
--- a/setup.py
+++ b/setup.py
@@ -32,11 +32,20 @@ hdf5 = [
 develop = [
     'coveralls',
     'mypy',
+    'ruff',
+    'pre-commit',
     'pytest',
     'pytest-asyncio',
     'pytest-cov',
     'pytest-mock',
     'pytest-random-order',
+    'isort',
+    'time-machine',
+    'types-cachetools',
+    'types-filelock',
+    'types-requests',
+    'types-tabulate',
+    'types-python-dateutil'
 ]
 
 jupyter = [
--- a/setup.py
+++ b/setup.py
@@ -91,7 +91,15 @@ setup(
         'aiofiles',
         'schedule',
         'websockets',
-        'janus'
+        'janus',
+        'ast-comments',
+        'aiohttp',
+        'blosc',
+        'cryptography',
+        'httpx',
+        'python-dateutil',
+        'tables'
+
     ],
     extras_require={
         'dev': all_extra,
--- a/setup.py
+++ b/setup.py
@@ -12,9 +12,11 @@ hyperopt = [
 
 freqai = [
     'scikit-learn',
+    'joblib',
     'catboost; platform_machine != "aarch64"',
     'lightgbm',
-    'xgboost'
+    'xgboost',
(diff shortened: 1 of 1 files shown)


xmatthias:
I'm not entirely sure what problem this is fixing (though it's good to have dependencies aligned).

While the dependencies should be aligned (in theory - but including the restrictions) - none of our installation methods will use a pure `setup.py` based installation - but all documented ways will first do a `pip install -r requirements<-*>.txt` run (depending on necessary dependencies).

Installing from pip is not part of the documentation (precisely for this reason) - and running / using freqtrade without matching dependencies (based on the requirements files) is considered an invalid /incorrect installation which didn't follow the documentation.

Dependency pins are important (but can't be auto-updated in a setup.py context) - therefore all documented installation methods install the requirements files explicitly first.

alxtrkhv:
Thank you very much for such a quick response 

I'd like to elaborate using a couple of examples: 

1. First one would be a pypi package that provides incomplete list of dependencies - it's not documented in the docs but it's there on pypi and [it's supported by the team(?)](https://pypi.org/project/freqtrade/) if I'm not mistaken - `pip install something` is one of the most idiomatic things in python

2. There are some tools like [pipx](https://github.com/pypa/pipx) and [poetry](https://github.com/python-poetry/poetry) that simplify python dependency management for different use cases: while `pipx` helps to install cli tools using isolated environments and PATH, `poetry` is used for general python project management - virtual environment chores, dependency resolution, building/packaging, etc - however it is not requirements.txt-friendly at the moment. 

   The best thing about those tools that they require virtually zero effort from developers - just 
a valid setup.py or pyproject.toml, or pypi package - both setup.py and pypi package are already there, just a bit outdated.

   While both those cases are undocumented and thus not explicitly supported by the team, I believe that both the project and its users could benefit from such an improvement.

3. Also there is [an example in the docs](https://www.freqtrade.io/en/latest/installation/#install-python-dependencies) with `pip install -e .` and without `pip install -r requirements*.txt` (could be a typo, I'm not sure)


I understand that there are non-python requirements that still should be documented and cannot be easily packaged even using `pipx` or `poetry`, but I think that such change would give more freedom and flexibility to the users in terms of UX and deployment pipelines.

xmatthias:
3 is definitely a hickup - as that's not intentional, and will (in almost all cases) result in an invalid installation.

i'd like to install from pypi - but due to ta-lib's current install problems (and not providing wheels), that's a long-shot - and simply running `pip install freqtrade` may give you a working freqtrade - but only if some pre-work was done (e.g. the ta-lib C lib was installed beforehand, ...).

it's also the case that sometimes "too uptodate" dependencies can break freqtrade - so unless dependencies are hard-pinned (with ==) - we can't really rely (or guarantee) that the installation works.

---

Now talking about "automated updates" - it seems like i was wrong.
a quick test reveals that dependabot CAN update dependencies in setup.py (https://github.com/xmatthias/test_python_dependabot/pull/1) ...

xmatthias (approved):
I'll merge this for now as it's improving the situation (it's mostly fixing the disconnect) - but it'll not fix the base problem - as some dependencies installed with this will be incompatible as they're not pinned properly.

PR #8619, opened 2023-05-09
Bug fix `live_retrain_hours`

robcaulk:
The introduction of `buffer_train_data_candles` truncates the training data timerange. This `timerange.stopts` is used after training to define the training timestamp, which is compared later against the `live_retrain_hours` to decide if a new model should be trained or not. 

If the user buffered the training data, then this `timerange.stopts` was artificially pushed backwards, causing `live_retrain_hours` to improperly think that a new train should be started, even though a model was recently trained within the `live_retrain_hours` period. 

This PR fixes the problem.

Thanks @smarmau and @davedavesen for reporting and investigating.

### DIFF
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -622,9 +622,9 @@ class IFreqaiModel(ABC):
             strategy, corr_dataframes, base_dataframes, pair
         )
 
-        new_trained_timerange = dk.buffer_timerange(new_trained_timerange)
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
LGTM 👍

robcaulk:
Waiting on @smarmau confirmation before merge on this one.

smarmau:
> Waiting on @smarmau confirmation before merge on this one.

Confirmed fix is good. Thanks @robcaulk and @xmatthias

Issue #8621, opened 2023-05-10
Forceenter input and error message seems not working 


polyvios-bc:
## Describe your environment

  * Operating system:  OSX
  * Python Version: 3.9.6
  * Freqtrade Version:  latest stable
  
## Your question

Trying to do a /forceenter rest api call to a running bot passing just the pair and direction that I want but seems its failing with no obvious reason.

Input payload: 
`{'pair': 'JASMY/USDT', 'side': 'short', 'price': None}
`

Response:
` {'detail': [{'loc': ['body'], 'msg': 'field required', 'type': 'value_error.missing'}]}`

According with the documentation here:
(https://www.freqtrade.io/en/stable/rest-api/#available-endpoints)

forceenter
    Force entering a trade

        :param pair: Pair to buy (ETH/BTC)
        :param side: 'long' or 'short'
        :param price: Optional - price to buy

It should work

xmatthias:
you can omit price (not send it) - but i don't think sending `None` is valid.

While i don't remember how i did it - you might wanna check the implementation in [freqUI](https://github.com/freqtrade/frequi) - where i know this is working fine.

polyvios-bc:
Yeah first i tried and without the price same error unfortunately.

polyvios-bc:
Seems like on freqUI you are using forcebuy -->

```
 async forceentry(payload: ForceEnterPayload) {
        if (payload && payload.pair) {
          try {
            // TODO: Update forcebuy to forceenter ...
            const res = await api.post<
              ForceEnterPayload,
              AxiosResponse<StatusResponse | TradeResponse>
            >('/forcebuy', payload);
            showAlert(`Order for ${payload.pair} created.`, 'success');

            return Promise.resolve(res);
          } catch (error) {
            if (axios.isAxiosError(error)) {
              console.error(error.response);
              showAlert(`Error occured entering: '${error.response?.data?.error}'`, 'danger');
            }
            return Promise.reject(error);
          }
        }
```

polyvios-bc:
Also forgot to mentioned I am just using the rest_client.py from your scripts  https://github.com/freqtrade/freqtrade/blob/develop/scripts/rest_client.py#L272

froggleston:
This sounds like a similar problem to the `limit` argument for pair_candles, where a check needed to be added to ensure the None isn't passed: https://github.com/freqtrade/freqtrade/blob/develop/scripts/rest_client.py#L335

[...]

PR #8622, opened 2023-05-10
Add check for None prices in forceenter REST API script

froggleston:
Add fix for forceenter to avoid passing None prices back to the API

## Summary

To fix the problem raised in https://github.com/freqtrade/freqtrade/issues/8621

## Quick changelog

- Add check to avoid passing `price: None` to API backend

### DIFF
--- a/scripts/rest_client.py
+++ b/scripts/rest_client.py
@@ -281,7 +281,7 @@ class FtRestClient():
                 "side": side,
                 }
         if price:
-            params['price'] = price
+            data['price'] = price
         return self._post("forceenter", data=data)
 
     def forceexit(self, tradeid, ordertype=None, amount=None):
--- a/scripts/rest_client.py
+++ b/scripts/rest_client.py
@@ -279,8 +279,9 @@ class FtRestClient():
         """
         data = {"pair": pair,
                 "side": side,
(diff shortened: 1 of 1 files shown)


xmatthias:
Are we sure that's actually a problem? (i don't mind the fix, don't get me wrong - just trying to understand if it's really necessary)

running `python scripts/rest_client.py --config api.json forceenter XRP/USDT long` seems to work just fine in the current develop.

yes, it's sending ` {'pair': 'BTC/USDT', 'side': 'long', 'price': None}` - but it's accepted by the backend just fine

froggleston:
> Are we sure that's actually a problem? (i don't mind the fix, don't get me wrong - just trying to understand if it's really necessary)
> 
> running `python scripts/rest_client.py --config api.json forceenter XRP/USDT long` seems to work just fine in the current develop.
> 
> yes, it's sending ` {'pair': 'BTC/USDT', 'side': 'long', 'price': None}` - but it's accepted by the backend just fine

Obviously if this is a non-fix, then I'm happy to see this closed unmerged.

xmatthias (approved):
I'm not sure it's a problem, but it's also not hurting to have this.

LGTM :+1:

PR #8623, opened 2023-05-10
Add Tensorboard logger for PyTorch and XGBoost

robcaulk:
Right now FreqAI only supports Tensorboard logging for Reinforcement Learning and Catboost.

Meanwhile, the PyTorch module is simply logging losses to command line. This makes the module nearly useless since users are unable to visualize their performance. This PR adds tensorboard support for PyTorch so that users can now  see clearly how well their models are performing, and how well they are adapting from train to train.

We also take the opportunity to add support for XGBoost users, who have been somewhat blind on their training for a while now. 

One caveat is that users *must* have the Torch FreqAI version installed for TB to properly work. If they do not, they will not gain the ability to use tensorboard, but the design of the PR ideally makes this transparent to the user.


freqtrade/freqai/tensorboard.py
```diff
@@ -0,0 +1,77 @@
...
+import xgboost as xgb
+
+
+logger = logging.getLogger(__name__)
+
+
+class BaseTensorboardLogger:
+    def __init__(self, logdir: str = "tensorboard", id: str = "unique-id"):
+        logger.warning("Tensorboard is not installed, no logs will be written."
+                       "Use ensure torch is installed, or use the torch/RL docker images")
```
xmatthias: Might need some rewording i'd think ;) 

**edit** Same below

tests/freqai/test_freqai_datakitchen.py
```diff
@@ -194,3 +194,4 @@ def test_get_full_model_path(mocker, freqai_conf, model):
 
     model_path = freqai.dk.get_full_models_path(freqai_conf)
     assert model_path.is_dir() is True
+    shutil.rmtree(Path(freqai.dk.full_path))
```
xmatthias: Not sure i understand this?

Assuming this is to delete some data this is creating - then this should be set to tmpdir (use as argument, it's a fixture in pytest) before the test / calls and set the base path to that- so it's only stored in the users temporary  directory (which is cleaned up automatically by pytest).

there's still another test that kinda leaves a `tensorflow` directory at the root (i know you're aware of that) - which suggests it's misconfigured to not use `user_data` (not necessarily the test - but the actual functionality), but current dir instead - which is kinda wrong.

tests/freqai/test_freqai_interface.py
```diff
@@ -552,8 +552,6 @@ def test_get_state_info(mocker, freqai_conf, dp_exists, caplog, tickers):
 
     if is_mac():
         pytest.skip("Reinforcement learning module not available on intel based Mac OS")
-    if is_py11():
```
xmatthias: is the "is_mac" check still valid?  (i'm not sure)

i think with recent updates to both sb3 and torch, this might be working on macos, too? (can for sure be looked at separately, and probably should ...)

freqtrade/freqai/__init__.py
```diff
@@ -0,0 +1,14 @@
+# ensure users can still use a non-torch freqai version
+try:
+    from freqtrade.freqai.tensorboard import TensorBoardCallback, TensorboardLogger
+    TBLogger = TensorboardLogger
+    TBCallback = TensorBoardCallback
+except ModuleNotFoundError:
+    from freqtrade.freqai.tensorboard import BaseTensorBoardCallback, BaseTensorboardLogger
+    TBLogger = BaseTensorboardLogger  # type: ignore
+    TBCallback = BaseTensorBoardCallback  # type: ignore
```
xmatthias: i **think** i understand what you're trying to do here : 
the initial import should fail if torch is not installed - and fall back to the `BaseTensorboardLogger`.

That's however not what's happening - this import will always work (unless xgboost is not installed, in which case both try and except blocks will fail as they both import from the same location).

Instead, you'll hit a `ModuleNotFound` exception once TensorboardLogger is initialized - as the import of torch is only happening once you're creating an instance - but at that point, there's no "graceful fallback"  - so it'll simply fail.


What you'll want to do instead is to split the 2 logics  into 2 files (`Base*` in one file, "actual" in the other) - and move the torch import to the top of the 2nd file (so it's failing on import).

Then this logic will work / be effective, and the fallback to "base" will actually work (the imports will then come from 2 separate files).

---

on top of that, i'd suggest to move this whole logic (callbacks + the content of this file) into it's own package (directory) - maybe called `tensorboard/...` - so it's clear that these 3 files belong together.

It'll obviously change the imports to `from freqtrade.freqai.tensorboard import xxx` ... but i think that's better than having it top-level - where it might cause problems with circular imports (you're importing from the __init__.py on files within that package - which can in some cases lead to problems).

freqtrade/freqai/tensorboard.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def log_scaler(self, tag: str, scalar_value: Any, step: int):
+        return
+
+    def close(self):
+        return
+
+
+class BaseTensorBoardCallback(xgb.callback.TrainingCallback):
```
xmatthias: Might make sense to import `xgb.callback.TrainingCallback` explicitly (from xgboost.callback import TrainingCallback` ? 
we're now importing xgb directly, but only ever use this one subclass (and a child of that subclass) in this file.

tests/freqai/test_freqai_datakitchen.py
robcaulk: Tensorflow? We dont even have tensorflow in develop. I dont know which test you are referring to.

tests/freqai/test_freqai_interface.py
robcaulk: We can test, but for me only way to test it is via CI since I do not have access to a mac.

tests/freqai/test_freqai_datakitchen.py
xmatthias: yeah typo (or brain lapsus) - tensorboard is the directory.

robcaulk: I think I have resolved all lingering files.

tests/freqai/test_freqai_interface.py
xmatthias: mmmh, turns out, we can't CI on macos ... (see segfault)

tests/freqai/test_freqai_interface.py
```diff
@@ -271,6 +280,7 @@ def test_start_backtesting(mocker, freqai_conf, model, num_files, strat, caplog)
     strategy.freqai_info = freqai_conf.get("freqai", {})
     freqai = strategy.freqai
     freqai.live = False
+    freqai.activate_tensorboard = test_tb
```
xmatthias, line 278: could this be set directly in the freqai_config fixture? so it applies to all tests implicitly (also future tests, saving us trouble in the past)? 

or are we not "fully" loading the config here - so this would be skipped? 
(obviously in theory applies to all tests that explicitly set activate_tensorboard)

xmatthias (dismissed):
Seems like the faulty macos test wasn't the one i was trying to exclude, but a different one (i didn't even touch) :laughing: 

LGTM :+1:

tests/freqai/test_freqai_interface.py
robcaulk: Good point, I think we can set it in the fixture.

xmatthias (approved):
LGTM (even more :laughing: ) :+1:

PR #8631, opened 2023-05-12
Clarify expectations about the FreqAI + Freqtrade tool

robcaulk:
We have marked uptick in users approaching our project with high expectations that the test strategies/rewards provided in the source code should be beating the market for them.

This PR aims to clarify the purpose of test/example strategies/rewards, all over the place. After this is pushed, only users who avoid documentation/docstrings all together will be the ones confused about how the tool works. All other users should be aware of the responsibility that they have for finding externally developed strategies or developing their own strategies. 

The PR also tries to add helpful tips/hints to users to help them create better more useful rewards/strategies.


xmatthias (approved):
wait, it's not a "switch on and be rich" approach? :scream: :rofl: 
joking, obviously...

LGTM :+1:

PR #8654, opened 2023-05-17
Integrate ArcticDB as history data storage

ShuxiangCao:
## Summary

Integrated ArcticDB as a history data storage option. Its remarkable capabilities enable effortless preservation and retrieval of historical data for backtesting from S3 compatible block storage services. See [Arcticdb website](https://arcticdb.io/).

## Quick changelog

- Added a data handler for ArcticDB (ohlcv only).

## What's new?

- Allows saving downloaded data to ArcticDB.
- Allows reading ArcticDB data source.


xmatthias:
For the moment, i don't see this as something we're likely to add - though further reasoning as to **why** you think this should be added can obviously change this (it's not a technical point so far, but a "why" point).

---

ArcticDB is a project that's 2 (!) months old (first commit was on march 15th). 
i'm sure it AIMS to be the go-to datastore for data science - but with 2 months, 300 stars and 12 contributors, i doubt this is a very mature project, and it clearly isn't there (also considering 

we use the data storage layer for 1 main purpose - store candle data, and retrieve it for backtesting/hyperopt/freqAI. Fast.
We therefore now recommend feather - which is a great format, and allows for great read times (especially compared to some other formats), while keeping a relatively small disk size.

I don't really see how this can be accomplished by an S3-like storage - which is _by definition_ remote - and will mean that for every backtest, we'll download data from a random s3 bucket.

I'm sure this will _technically_ work (without having looked at the code) - but it'll be 100 times slower than a local filestore (no matter if you're on a fast connection, or on a VPS). 

I'm therefore not sure what benefit you think this will bring to freqtrade, so maybe you can outline the usecase, and why you think this is a must (and some performance comparisons against a locally hosted feather would be nice, too).

ShuxiangCao:
Thank you for your quick response.

- ArcticDB, established in 2018 (See [history here](https://docs.arcticdb.io/history/)), is a robust and proven database solution. It is managed by the Man group and has been effectively incorporated into their time series analysis system. I have strong confidence in its quality.

- I recommend the integration of ArcticDB with freqtrade for use on the AWS service. While running freqtrade locally and accessing S3 for backtesting data could potentially yield slower speeds, integrating with AWS services, such as Sagemaker, allows for the creation and deployment of a highly scalable pipeline.

- Another significant advantage of this integration is its streamlined file system management. The existing solutions includes feather, requires you to either download the data or mount the folder to a existing file system. With arcticdb and S3, you could establish a nearly stateless container (although the user strategy folder would still require mounting), eliminating concerns about supplying the datasource file system. This would allow you to deploy a large number of backtesting jobs to the cluster by merely specifying the parameters through arguments.

I'll quickly fix the import issue and I look forward to hearing your thoughts.

xmatthias:
So far, i don't see a benefit to the majority of the freqtrade users - which will be iterating (and hopefully imporving) on their strategy - not "spin up a ton of backtests".
It can have some edge-case usecases - but for now, I'm not convinced that I'll want to accept maintenance for this - as it'll serve a very niche usecase - and can potentially cause quite some overhead, with questionable benefits (see next paragraph).

S3 storage can quite easily be mounted as drive in an aws environment - which makes it the problem of the user wanting to use such a solution, and won't add unnecessary dependencies and code, and with that maintenance to this project.

---

As requested above, please provide some performance tests for this comparing to "local feather" (feel free to test both locally and on AWS if you think that makes a big difference).

---

**edit:** After looking at the license, I'm even less convinced. Before looking at this further, please provide some (**independent**!) source which confirms that their odd "business source license" is actually compatible with GPL3 - the license we use in this project.

It's not a commonly used license (it's the first time i hear from this license) - and on a quick search, i couldn't find any reference if the "Business Source License" is compatible to ANY open source license or not (it's not itself an open source license - but that's not automatically a blocker).

ShuxiangCao:
Okay, the license seems to be a problem, and I'm not sure I have the expertise to conclusively assess its compatibility.  I'm stepping back for now. But, if we need this feature in the future, I'm ready to help again. Meanwhile, I'll see how else I can help with freqtrade.

xmatthias:
Quite honestly - licenses are a pain.
i'm no lawyer or expert in this, either - which makes me even more cautious.

i only saw it accidentally  - but licenses are ... dangerous for any project :(

PR #8677, opened 2023-05-22
Stable baselines 3 seeding update

richardjozsa:
## Summary

After the gymnasium update of stable baselines 3 packages there were issues with seeding which caused reproducibility issues, they patched and fixed it, but because of it we need to align the reset function.

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
This is fine for me (ignore the failing test - that's not related to this change). :+1: 

@robcaulk - any objections?

robcaulk (approved):

robcaulk:
good for me too

PR #8692, opened 2023-05-26
Outsource data pipeline handling to improve flexibility

robcaulk:
The weakest point of FreqAI is currently the difficulty adding new data pipeline steps. For example, if a user wants to change the normalization methods, they must be very familiar with the DataKitchen. And what is worse, they likely wont be able to make it flexible enough for pushing upstream for the community. This will lead to their fork becoming quickly obsoleted. Not good.

This weakness originates from some early design decisions associated with automating pipeline creation from the config (e.g. `use_SVM_to_remove_outliers`) and some pretty interesting complexities in adaptive modeling in the Freqtrade environment. There were a few major short-comings missing from SKLearn's pipeline back when we created FreqAI:

- propagating point removal across X, y and sample_weights
- dynamic feature removal
- propagating proper feature names through the dynamic removal
- propagating proper feature names after dimension change (e.g. PCA)
 
 Additionally, FreqAI has a variety of methods that are quite useful and are not found in SKLearn's library such as the `DissimilarityIndex`, a self adapting DBSCAN, or even wild experiments like the inlier metric. 
 
Next, Freqtrade requires a way of passing information back to the strategy about trustworthiness of a single prediction. This constraint originates from a need to *always* have a prediction point for all candles (making the dataframe sizes match). This means during inference we actually do not want to *remove* outliers, we only want to signal an outlier. Enter `do_predict`. Another missing item from SKLearn's pipeline. 
 
 Finally, after observing and building around the DataKitchen, it became clear a few months back that we needed more normalization methods and we wanted to introduce more fun transformations from SKLearn et al. So we needed to build a library that would allow for easily inheriting SKLearn methods, while maintaining the modularity of the SKLearn pipeline - but keeping all the previously mentioned constraints of the FT + FreqAI adaptive modeling environment. It turns out that these coherence requirements exist in other adaptive modeling applications as well - so we might as well build a centralized library for this kind of thing.
 
 This PR eliminates most if not all data processing from the FreqAI source code, opting to outsource to a new library called [datasieve](https://github.com/emergentmethods/datasieve). 

Doc excerpt:

Users should see no change from the standard configuration interface. However, users who want to start building their own datapipelines can now do so with `define_data_pipeline()`:

### Customizing the pipeline

Users are encouraged to customize the data pipeline to their needs by building their own data pipeline. This can be done by simply setting `dk.feature_pipeline` to their desired `Pipeline` object inside their `IFreqaiModel` `train()` function, or if they prefer not to touch the `train()` function, they can override `define_data_pipeline`/`define_label_pipeline` functions in their `IFreqaiModel`:

!!! note "More information available"
    FreqAI uses the the [`DataSieve`](https://github.com/emergentmethods/datasieve) pipeline, which follows the SKlearn pipeline API, but adds, among other features, coherence between the X, y, and sample_weight vector point removals, feature removal, feature name following. 

```python
from datasieve.transforms import SKLearnWrapper, DissimilarityIndex
from datasieve.pipeline import Pipeline
from sklearn.preprocessing import QuantileTransformer, StandardScaler
from freqai.base_models import BaseRegressionModel


class MyFreqaiModel(BaseRegressionModel):
    """
    Some cool custom model
    """
    def fit(self, data_dictionary: Dict, dk: FreqaiDataKitchen, **kwargs) -> Any:
        """
        My custom fit function
        """
        model = cool_model.fit()
        return model

    def define_data_pipeline(self) -> Pipeline:
        """
        User defines their custom feature pipeline here (if they wish)
        """
        feature_pipeline = Pipeline([
            ('qt', SKLearnWrapper(QuantileTransformer(output_distribution='normal'))),
            ('di', ds.DissimilarityIndex(di_threshold=1)
        ])

        return feature_pipeline
    
    def define_label_pipeline(self) -> Pipeline:
        """
        User defines their custom label pipeline here (if they wish)
        """
        label_pipeline = Pipeline([
            ('qt', SKLearnWrapper(StandardScaler())),
        ])

        return label_pipeline
```

Here, you are defining the exact pipeline that will be used for your feature set during training and prediction. You can use *most* SKLearn transformation steps by wrapping them in the `SKLearnWrapper` class as shown above. In addition, you can use any of the transformations available in the [`DataSieve` library](https://github.com/emergentmethods/datasieve). 
 
 TODO:
 - [x] Ensure classifier works with pipeline
 - [x] Ensure Reinforcement Learning works with pipeline
 - [x] Fix all tests
 - [x] Document how a user would create their own pipeline in the IFreqaiModel
 - [x] Document how a user would create a custom step in a pipeline
 - [x] Document how a user would add an SKLearn step to the pipeline
 - [x] Add noise transform to Datasieve
 - [x] Add a FreqAI specific step as an example/new feature
 - [x] Ensure thread count is propagated to pipeline
 - [x] Add migration guide
 - [x] Throw warnings that direct to migration guide
 - [x] Allow users to continue using deprecated code with full backwards compatibility (but with noisy warnings)]


freqtrade/freqai/base_models/BaseClassifierModel.py
```diff
@@ -50,21 +50,30 @@ def train(
         logger.info(f"-------------------- Training on data from {start_date} to "
                     f"{end_date} --------------------")
         # split data into train/test data.
-        data_dictionary = dk.make_train_test_datasets(features_filtered, labels_filtered)
+        d = dk.make_train_test_datasets(features_filtered, labels_filtered)
```
xmatthias: I agree the variable can/should be made shorter - but i'd not make it 1 letter.
In my experience, 1 letter variable names can lead to bugs if they're used outside of loops (basically everything except `for i in range(20)`.

Maybe `ds` (for dataset) could be a viable alternative?

(Obviously this ocurrance is a sample, and applies to all other models, too)

freqtrade/freqai/transforms/quantile_transform.py
```diff
@@ -0,0 +1,28 @@
...
+    A subclass of the SKLearn Quantile that ensures fit, transform, fit_transform and
+    inverse_transform all take the full set of params X, y, sample_weight required to
+    benefit from the DataSieve features.
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def fit_transform(self, X, y=None, sample_weight=None, feature_list=None, **kwargs):
+        super().fit(X)
```
xmatthias: assuming this is a class that will be subclassed further - shouldn't this be `self.fit(X)`? (it's not relevant for this implementation, but if this classes "fit" method does something more ... ?

freqtrade/freqai/transforms/quantile_transform.py
```diff
@@ -0,0 +1,28 @@
...
+        X = super().transform(X)
+        return X, y, sample_weight, feature_list
+
+    def fit(self, X, y=None, sample_weight=None, feature_list=None, **kwargs):
+        super().fit(X)
+        return X, y, sample_weight, feature_list
```
xmatthias: i think this class might need some more explanation.

at first glance, it seems like a bug to return 3/4 variables without modification - but i think it's intentional - i just don't understand the reason for that at the moment.

freqtrade/freqai/data_drawer.py
```diff
@@ -513,13 +514,19 @@ def load_data(self, coin: str, dk: FreqaiDataKitchen) -> Any:  # noqa: C901
...
+            dk.feature_pipeline = self.meta_data_dictionary[coin]["feature_pipeline"]
+            dk.label_pipeline = self.meta_data_dictionary[coin]["label_pipeline"]
         else:
             with (dk.data_path / f"{dk.model_filename}_metadata.json").open("r") as fp:
                 dk.data = rapidjson.load(fp, number_mode=rapidjson.NM_NATIVE)
 
             dk.data_dictionary["train_features"] = pd.read_pickle(
                 dk.data_path / f"{dk.model_filename}_trained_df.pkl"
             )
+            with (dk.data_path / f"{dk.model_filename}_pipeline.pkl").open("rb") as fp:
```
xmatthias: These filenames should probably be constants, so miss-types will be avoided.

The problem of not having it as constants is right in this version/commit - where `f"{dk.model_filename}_feature_pipeline.pkl"` is saved, while `f"{dk.model_filename}_pipeline.pkl"` is loaded (which will obviously not exist). (label_pipeline.pkl seems correct).

freqtrade/freqai/base_models/BaseTensorFlowModel.py
```diff
@@ -1,70 +0,0 @@
-import logging
```
xmatthias, line 1: for my own understanding, is this generic cleanup of an unused base-class, or related to the datapipeline introduction / change somehow?

tests/freqai/test_freqai_datakitchen.py
```diff
@@ -9,9 +9,9 @@
 from freqtrade.data.dataprovider import DataProvider
 from freqtrade.exceptions import OperationalException
 from freqtrade.freqai.data_kitchen import FreqaiDataKitchen
-from tests.conftest import get_patched_exchange, log_has_re
-from tests.freqai.conftest import (get_patched_data_kitchen, get_patched_freqai_strategy,
-                                   make_data_dictionary, make_unfiltered_dataframe)
+from tests.conftest import get_patched_exchange  # , log_has_re
+from tests.freqai.conftest import make_unfiltered_dataframe  # make_data_dictionary,
+from tests.freqai.conftest import get_patched_data_kitchen, get_patched_freqai_strategy
```
xmatthias:
```suggestion
from tests.conftest import get_patched_exchange
from tests.freqai.conftest import make_unfiltered_dataframe
from tests.freqai.conftest import get_patched_data_kitchen, get_patched_freqai_strategy
```
?

docs/freqai-feature-engineering.md
```diff
@@ -209,15 +209,67 @@ Another example, where the user wants to use live metrics from the trade databas
...
+```
+
+Here, you are defining the exact pipeline that will be used for your feature set during training and prediction. If you have a custom step that you would like to add to the pipeline, you simply create a class that follows the DataSieve/SKLearn API. That means your step must have a `fit()`, `transform()`, `fit_transform()`, and `inverse_transform()` method. You can see examples of this in the `freqtrade.freqai.transforms` module where we use SKLearn `QuantileNormalization` to create a new step for the pipeline.
+
+As there is the `feature_pipeline`, there also exists a definition for the `label_pipeline` which can be defined the same way as the `feature_pipeline`, by overriding `define_label_pipeline`.
+
+!!! note "Inheritence required"
+    While most SKLearn methods are very easy to override, as shown in freqtrade/freqai/transforms/quantile_transform.py, they still need to include passing X, y, and sample_weights through all `fit()`, `transform()`, `fit_transform()` and `inverse_transform()` functions, even if that means a direct pass through without modifications.
+
+<!-- ## Data dimensionality reduction with Principal Component Analysis
```
xmatthias: What's the assumption with a 20 line comment blob in the docs?

It's either wrong/superceded, so we can remove it, or it can be shown? (or is this temporary??)

freqtrade/freqai/freqai_interface.py
```diff
@@ -503,68 +505,41 @@ def check_if_feature_list_matches_strategy(
...
-            dk.compute_inlier_metric(set_='train')
-            if self.freqai_info["data_split_parameters"]["test_size"] > 0:
-                dk.compute_inlier_metric(set_='test')
-
-        if ft_params.get(
-            "principal_component_analysis", False
-        ):
-            dk.principal_component_analysis()
+        if ft_params.get("principal_component_analysis", False):
+            dk.feature_pipeline.steps += [('pca', ds.DataSievePCA())]
```
xmatthias: (not tied to this line in particular, just noticed it while reading through this part)

From an interface perspective, i think you'd not want to expose `.steps` in a writing manner (instead, make it private).

using a `dk.feature_pipeline.append(('pca', ds.DataSievePCA()))` will almost certainly be better, and will allow for more flexibility in the future.

Aside from that - you're using a tuple - what's the first entry, what the 2nd?  (i assume (name, instance)?)

Is the name assumed to be unique? or can i do the following without negative side-effects? (assumption is, I'm lazy and forgot to update the step name).

``` python
 Pipeline([
            ('something', ds.DataSieveVarianceThreshold(threshold=0)),
            ('something', ds.DataSieveMinMaxScaler(feature_range=(-1, 1)))
            ])
```

How will it behave if the tuple is missing the name, and is just 1 element long? (yes, it's probably violating the interface, but are there checks for that?)
Based on the docs, this is a somewhat user-facing interface - so it'll have to be pretty stable (assumptions on "User will provide a 2 element tuple" are usually voided upon ~enemy~ user contact

(this might be checks you'd want to do in an `.append()` method ...)

freqtrade/freqai/prediction_models/XGBoostRegressor.py
```diff
@@ -8,6 +8,9 @@
 from freqtrade.freqai.tensorboard import TBCallback
 
 
+# from datasieve.pipeline import Pipeline
+# from freqtrade.freqai.transforms import FreqaiQuantileTransformer
+
```
xmatthias: Intentionally commented? still something that's left to do?
leftovers ?

docs/freqai-feature-engineering.md
```diff
@@ -209,15 +209,67 @@ Another example, where the user wants to use live metrics from the trade databas
...
+    def define_data_pipeline(self, dk: FreqaiDataKitchen) -> None:
+        """
+        User defines their custom eature pipeline here (if they wish)
+        """
+        from freqtrade.freqai.transforms import FreqaiQuantileTransformer
+        dk.feature_pipeline = Pipeline([
+            ('qt', FreqaiQuantileTransformer(output_distribution='normal'))
+        ])
+
+        return
```
xmatthias: from an interface perspective, this (and the corresponding label_pipeline method) feels odd, and doesn't really fit into most other interface methods we have.

It's based on "create side-effects" only - while returning nothing, which is often problematic (the side-effects, not so much the "return nothing).

I feel that from a user perspective (and alignment with most other strategy/freqai interface methods), it would be better to return an array of pipeline steps  - which the backend then assembles into an actual Pipeline.

That'll also somewhat abstract the `Pipeline()` object, as the user doesn't need to know about it, really.
What the user cares about is the list of pipeline steps - not "needs to perform A, B, C".

freqtrade/freqai/transforms/quantile_transform.py
robcaulk: Yes it is intentional, it is a stand alone functional library that pipes through these four variables from start to finish. All methods must carry all four variables through no matter what, even if they don't modify them or simply pass them through to the next step. Some steps may modify one of the variables, others may modify all 4, some may do something special with one of them. The only constraint of it all is that we pass all 4 through all the time on all steps.

Unless you know of a better way to handle something like this?

freqtrade/freqai/transforms/quantile_transform.py
robcaulk: It is not the case, so SKLearn's `fit()` is not the same as the `self.fit()`. In this case it is because we want to carry these 3 extra variables through our pipeline. 

If you have a recommended way to change this I am all ears - but this seemed to be the most effective approach to keep SKLearn around and keep a familiar API, while introducing the possibility of making actions on the other variables.

freqtrade/freqai/base_models/BaseClassifierModel.py
robcaulk: Really, I'd prefer to shorten the keys of the `data_dictionary`, but I think it would break too many people's code doing so.

But we can use `ds` instead.

freqtrade/freqai/base_models/BaseTensorFlowModel.py
robcaulk: Yeah, that is right. It is unused, since we decided against introducing TensorFlow in favor of pushing and supporting PyTorch + dependencies.

freqtrade/freqai/base_models/BaseClassifierModel.py
xmatthias: yeah, user facing stuff (like keys) are very hard to change

freqtrade/freqai/transforms/quantile_transform.py
xmatthias: mmh, that's not exactly what i meant, maybe i should've written `self.fit(X, y, sample_weight, feature_list, **kwargs)` so it's clear which method i think should be called. (it's the `super()` part i'm concerned about, not the variables / absence thereof).

in my understanding, `fit_transform()` is shorthand to `.fit() & .transform()` (conceptually - which is somewhat confirmed by the calls to `super().fit()` and `super().transform()` in this method).

As such, i could inherit from this class, implement logic _only_ in `.fit()` (the rest of the methods is already implemented, as this is not an abstract interface).
I'd then expect my `fit_transform()` method to call my newly created logic in `.fit()` (obviously the same applies to the `.transform()` method below.
by calling `super().fit()` here - changes to the implementation of `.fit()` will be ignored when calling `.fit_transfrom()`.

robcaulk: Yeah, you are right. In fact, that is how we already handle the more interesting transforms, like the `SVMOutlierExtractor` (the conversion of our `use_svm_to_remove_outliers()`)  [here](https://github.com/emergentmethods/datasieve/blob/main/datasieve/transforms/svm_outlier_extractor.py#L19).

 This FreqaiQuantileTransform is a very simple example of how to get an SKLearn element into the FreqAI pipeline.  Without a doubt, putting `self.fit(); self.transform()` into `fit_transform()` would be superior for this example (Ill swap it over) as it would allow users to continue inheriting from this example if they wished. Which is why we do exactly that with `SVMOuterlierExtractor`. However, we agree that running this simple code with `self.fit()` or `super.fit()` yields the same results. 

The simplicity of this example doesnt help convey they general idea of carrying `X, y, sample_weights, feature_list` around together. So I would implore you to check out the `SVMOutlierExtractor` linked above - it would help illustrate the ultimate needs/goals.

xmatthias: Yeah, that's what i mean.

obviously for **this** implementation, it's irrelevant, but further inheriting from this (only implementing what's needed) might have this suddenly buggy without good reason.

---

Another interesting point the link to the `svm_outlier_extractor` had me think about is that `.fit()` actually returns a value (i assume X?) - which is however not used.
this will have no impact as long as the original `X, y, ...` are mutated - but if for some reason, a new list is assigned / returned (for whatever reason) instead of mutating the input, then that new list would be ignored, effectively / potentially causing faulty/wrong behavior.
Again, most likely no impact in this particular instance (accidentally), but not really suitable and potentially problematic for a functional interface.

Effectively, this might mean that this line should potentially be something around the below (obviously, not just here, but potentially also in the datasieve repo / implementations)

```
X, y, sample_weight, feature_list = self.fit(X, y, sample_weight, feature_list, **kwargs)
# ...
```

robcaulk: It depends entirely on each step of the pipeline. If the user thinks that their `fit()` should modify `X`, then they would return and pass it along. For SVMOutlierExtractor `fit()` simply fits the model and does not `transform()` anything. This is the typical case in SKLearn as well, `fit()` does not mean `transform()`. If you want to also `transform`,then you call `fit_transform()`. 

But it doesnt stop users from doing anything they want inside `fit()` if they think that `fit()` should be doing more to prepare for the subsequent `fit()` functions in other steps. 

For example, PCA really doesnt work unless it is `fit_transform` or `transform`. This is because the PCA step changes the dimensionality of the data, so subsequent steps *depend* on this transformation to be properly fit. Else they are useless. 

Building pipelines gets tricky when incorporating a wide range of methods and when trying to accommodate a variety of input/output needs.

freqtrade/freqai/transforms/quantile_transform.py
xmatthias: ah wait, this class needs to correspond to the interface of what the datasieve Pipeline class expects, not to the interface of sklearn?

wouldn't you want to subclass from datasieve instead (most likely an Abstract interface defining the methods, and arguments that it expects)?
If that Parent defines proper / decent types, mypy should be able to capture eventual contract violations in that case.
Obviously QuantileTransformer will then have to be created as part of the constructor for this class - but it's not "QuantileTransformer -> FreqaiQuantileTransformer" is not really parent/child if the parameters for the same method changed - but "PipelineType -> FreqaiQuantileTransformer" is.

This would also allow to check for correct implementation (`issubclass(cls, PipelineType)` when the pipeline is created - not once it's running and hits a runtime error.

The way this is defined now only works mostly accidentally i fear (most likely due to missing sklearn types) - and is very prone to break on mypy with ["Return type "str" of "foo" incompatible with return type "int" in supertype "A""](https://mypy.readthedocs.io/en/stable/error_code_list.html?highlight=override#check-validity-of-overrides-override) (or a corresponding error for method arguments).

freqtrade/freqai/freqai_interface.py
robcaulk: Yes, thanks for asking and for the recommendation. I added this check so now datasieve has checks to ensure the user doesnt reuse the same string twice [see here](https://github.com/emergentmethods/datasieve/blob/v0.1.1/datasieve/pipeline.py#L49). 

I also added the `.append()` method and converted code to use that.

freqtrade/freqai/prediction_models/XGBoostRegressor.py
robcaulk: I kept these commented so users could see how they might define their own pipeline in the IFreqaiModel. I am still thinking about better ways to convey this to users.

docs/freqai-feature-engineering.md
robcaulk: I wasnt sure if we would remove it or not - in the end it is removed.

docs/freqai-feature-engineering.md
robcaulk: I agree on one point and disagree on another. 

I agree it feels better to return the pipeline, and that is what I have done. Now any user who is starting from one of our `IFreqaiModel` classes will see

```
        dk.feature_pipeline = self.define_data_pipeline()
        dk.label_pipeline = self.define_label_pipeline()
```

Which they can see is automating the construction of the `Pipeline` object. If they want, it is easy to see that they can remove the `self.define_data_pipeline()` and define it themselves following the `Pipeline` API (which is identical to SKLearn so it is a "universal" pipeline definition approach).

I disagree that they this `define_data_pipeline()` should have yet another function somewhere assembling steps. This function is *already* the automation step. Users should see the definition clearly and be able to define the pipeline themselves.

freqtrade/freqai/freqai_interface.py
```diff
@@ -505,42 +507,47 @@ def check_if_feature_list_matches_strategy(
...
 
-        dk.feature_pipeline.fitparams = dk.feature_pipeline._validate_fitparams(
-            {}, dk.feature_pipeline.steps)
+        sigma = self.freqai_info["feature_parameters"].get('noise_standard_deviation', 0)
+        if sigma:
+            feature_pipeline.append(('noise', ds.Noise(sigma=sigma)))
 
-        # if self.freqai_info["feature_parameters"].get('noise_standard_deviation', 0):
-        #     dk.pipeline.extend(('noise', ds.Noise()))
+        feature_pipeline.fitparams = feature_pipeline._validate_fitparams(
```
xmatthias: is this a mandatory call a user is expected to make? 
if yes, why is the method "private" (_ prefixed methods are considered private by convention).
Also if yes - should this call be abstracted away from the user (by moving it to "after" the calls)?

Also, is there some further docs to datasieve? i've looked at the readme (couldn't find other docs) - which doesn't seem to mention this method at all ?

docs/freqai-feature-engineering.md
xmatthias: > I disagree that they this define_data_pipeline() should have yet another function somewhere assembling steps. This function is already the automation step. Users should see the definition clearly and be able to define the pipeline themselves.

I'm not sure where/how i've suggested this (it's not in the above comment - at least i don't interpret my this way).

i'd think the change is fine - but this last paragraph has me confused :laughing:

docs/freqai-feature-engineering.md
```diff
@@ -209,57 +209,77 @@ Another example, where the user wants to use live metrics from the trade databas
...
+    ])
+```
 
-![inlier-metric](assets/freqai_inlier-metric.jpg)
+But users will find that they can add PCA and other steps just by changing their configuration settings, for example, if you add `"principal_component_analysis": true` to the `feature_parameters` dict in the `freqai` config, then FreqAI will add the PCA step for you resulting in the following pipeline:
+
+```py
+from datasieve.transforms import SKLearnWrapper, DissimilarityIndex, PCA
+from datasieve.pipeline import Pipeline
+dk.feature_pipeline = Pipeline([
```
xmatthias: might need removal of `dk.` with the new structure? (also above)

freqtrade/freqai/prediction_models/XGBoostRegressor.py
xmatthias: Documentation won't be sufficient you think?

freqtrade/freqai/transforms/quantile_transform.py
robcaulk: I think this is resolved with the new datasieve api

docs/freqai-feature-engineering.md
robcaulk: No, if they are defining the pipeline directly, then they will want to set it to `dk.feature_pipeline`. 

If they want to use the `define_data_pipeline()` function then they would follow the instructions below.

freqtrade/freqai/freqai_interface.py
robcaulk: No, this is only if the pipeline is assembled step by step with `.append()` which is not an expected user facing interaction. The default `define_data_pipeline()` is constructing the pipeline for them based on configuration settings, so we need to instantiate it and dynamically define it. This is not how users should be interacting with the pipeline (I dont know why they would want to do it this way instead of just defining their pipeline).

Users would define the full pipeline at once, as we show in the documentation. In which case they wouldn't be interacting with this method.

freqtrade/freqai/prediction_models/XGBoostRegressor.py
robcaulk: I find it is quite rare for users to read the documentation, unfortunately. The more hints we install in various locations the more likely they are to "catch on."

xmatthias: I'm not a fan of commented code throughout just as "samples".

People are not expected to read through the code unless they're developers. If they're developers, i expect they can read documentation. If they don't read documentation, they'll also not find it here.

freqtrade/freqai/freqai_interface.py
xmatthias: I think then this should be done differently here, too.
there's nothing preventing you from creating a list we append the single entries to - and calling `Pipeline(pipelinesteps)` at the end of the function - which will make this call obsolete - and will leave control over necessary validations to datasieve (you can enhance this later on eventually).

It's a private, undocumented method on an external (code is not in this repo) module - so it shouldn't be called directly (otherwise it should be made public).

docs/freqai-feature-engineering.md
xmatthias: why would they (what's the usecase to set this directly)? (why do we need to have 2 interfaces for this?)

I think it's better for both usability(easier to explain to users), as well as maintainability to have 1 interface for this.
i can see no benefit in allowing the user to set this in multiple ways that won't result in head-scratching (why are there 2 ways, which one should i use??) and maintenance problems.

and in which method / part of which code would that code segment go? If we assume both interfaces as valid - then the docs will need some clarification for sure - as based on the code and descriptions here, i have no clue where this should be placed.

robcaulk: Really? I am struggling to find where the misunderstanding is in the current docs:


## Customizing the pipeline

Users are encouraged to customize the data pipeline to their needs by building their own data pipeline. This can be done by simply setting `dk.feature_pipeline` to their desired `Pipeline` object inside their `IFreqaiModel` `train()` function, or if they prefer not to touch the `train()` function, they can override `define_data_pipeline` in their `IFreqaiModel`:

```py
    from datasieve.transforms import SKLearnWrapper, DissimilarityIndex
    from datasieve.pipeline import Pipeline
    from sklearn.preprocessing import QuantileTransformer
    def define_data_pipeline(self) -> Pipeline:
        """
        User defines their custom eature pipeline here (if they wish)
        """
        feature_pipeline = Pipeline([
            ('qt', SKLearnWrapper(QuantileTransformer(output_distribution='normal'))),
            ('di', ds.DissimilarityIndex(di_threshold=1)
        ])

        return feature_pipeline
```

Here, you are defining the exact pipeline that will be used for your feature set during training and prediction. Here you can use *most* SKLearn transformation steps by wrapping them in the `SKLearnWrapper` class.

As there is the `feature_pipeline`, there also exists a definition for the `label_pipeline` which can be defined the same way as the `feature_pipeline`, by overriding `define_label_pipeline`.

robcaulk: As the doc says, if the user doesnt want to touch `train()` then they can simply override `define_data_pipeline()`. 

There are going to be many ways to skin a cat here, the goal is to convey that we have an easy way for them to override the pipeline without touch deep internals: hence `define_data_pipeline`. But in the end its just setting `dk.feature_pipeline` in the `train()` function fro them, so they are welcome to set that `dk.feature_pipeline` however they wish if they are building their own `train()` of the IFreqaiModel. 

`fit()` `train()` `predict()` are user facing functions that we expect power-users to wish to modify. Now we are adding `define_data_pipeline()` to that list for a user who doesnt need/want to redefine `train()` just to get a custom pipeline.

freqtrade/freqai/prediction_models/XGBoostRegressor.py
robcaulk: Ok I have removed it.

docs/freqai-feature-engineering.md
robcaulk: In fact, we have three ways for them to modify the pipeline. With the primary method being simply changing flags inside their configuration file :-)

xmatthias: Yes - and 3 is clearly too much. :) 

the flags are fine (and needed for compatibility) - but i think we should focus on one method, and have only one of these methods defined/documented.

I don't see a valid reason to have both - and it'll only be confusing for people.
having too much choice is not good - and i struggle to see **any** benefit of defining this in `train()`  over using the dedicated method.

That'll be a clear way for the user to approach this.
sure, overriding internals is always an option - but we should have ONE way to do this via config (we do) and ONE way to do this via code. Then it's clear which way we want to guide users.

Compare this with the definition of the entry/exit signals
nothing prevents you from defining everything in `populate_indicators()` - but we provide users with a clear guideline of "this should go here". It's not a technical necessity - but one clear way to approach this in a uniform way.

> Really? I am struggling to find where the misunderstanding is in the current docs:

The part you highlight here - yes.
The other part isn't really clear though - and will only lead to confusion and missunderstandings.

robcaulk: Ok, then I will convert the docs to follow your request. We simply remove the discussion about defining the pipeline with `dk.feature_pipeline=` and assume a user who is interacting with the internals can already see this clearly and understand this. Instead we focus the docs to show that a user would use the `define_data_pipeline` method.

xmatthias: we should work based on the assumption that people do share their code / custom implementations of things we consider as "user facing".

Providing ONE clear way to do things will make it easier for people to understand what others did / tried to do, and improve based on that.

if we assume `train()` as user facing, then a common call to `define_data_pipeline()`  (with assignment to the dk object) is fine.

if that call to `define_data_pipeline()` is implemented by some users - but missing in favor of direct definition in other objects (created by other users, but shared), it'll make it VERY difficult for people to share, as it's 2 conflicting methods to do the same thing (If i define it directly, `define_data_pipeline ()` will not be called, which would be unexpected).

---

As said, i don't see a benefit of the "direct" definition other than "just another way". yes that's just another way, and there's probably 20 more - but guiding users  through one way will make it easier to use, for the user, as well as easier to support from our side (they'll always have used it identically).

freqtrade/freqai/data_drawer.py
robcaulk: Done.

docs/freqai-feature-engineering.md
robcaulk: Ok fixed, now the documentation highlights two methods for building a pipeline. From the config, or going in and customizing it yourself.

freqtrade/freqai/freqai_interface.py
robcaulk: You are right, easier to build the list and then instantiate the object at once at the end. Done.

freqtrade/freqai/freqai_interface.py
```diff
@@ -991,3 +968,26 @@ def predict(
         :do_predict: np.array of 1s and 0s to indicate places where freqai needed to remove
         data (NaNs) or felt uncertain about data (i.e. SVM and/or DI index)
         """
+
+    # deprecated functions
+    def data_cleaning_train(self, dk: FreqaiDataKitchen, pair: str):
+        """
+        throw deprecation warning if this function is called
+        """
+        ft = "https://www.freqtrade.io/en/latest"
```
xmatthias: This is used at least twice - and defined twice (here and in the next function below) - so it should be extracted by at least one level (to only define this once) - maybe even to `constants.py` (though i'll leave that up to you).

I also think it should point to stable.
While it'll become an invalid link until the next release - i think users should for the most part be using "stable" - not "latest" from a docs perspective.
Latest can contain features not yet available with the release version - which in my experience is what most people are using.

freqtrade/freqai/prediction_models/XGBoostRFRegressor.py
```diff
@@ -44,7 +45,10 @@ def fit(self, data_dictionary: Dict, dk: FreqaiDataKitchen, **kwargs) -> Any:
 
         model = XGBRFRegressor(**self.model_training_parameters)
 
+        model.set_params(callbacks=[TBCallback(dk.data_path)], activate=self.activate_tensorboard)
```
xmatthias, line 48: Is this change related to pipelines? (all changes in this file, really)?

if yes, then i think i misunderstood how pipelines work, or rather what they're supposed to do . . .

(don't get me wrong, it's ok to have it though in general i'd have preferred it as separate change - it'd just surprise me if this impacts pipeline behavior).

requirements-freqai.txt
```diff
@@ -10,3 +10,4 @@ catboost==1.2; 'arm' not in platform_machine and (sys_platform != 'darwin' or py
 lightgbm==3.3.5
 xgboost==1.7.5
 tensorboard==2.13.0
+datasieve==0.1.4
```
xmatthias: should be added to setup.py as well.
if we require a specific min-version - best using `>=0.1.4` or similar (it'll usually not be a pinned version ther).
it'll have to be there to facilitate installs from pypi, for example `pip install freqtrade[freqai]`.

While not a documented install method (thanks ta-lib...), it's still one that's used by some people.

freqtrade/freqai/base_models/BasePyTorchClassifier.py
```diff
@@ -149,3 +161,57 @@ def get_class_names(self) -> List[str]:
             )
 
         return self.class_names
+
+    def train(
```
xmatthias, line 165: Is there any particular difference between this and BaseTorchRegressor i'm missing? 

The 2 methods seem to be identical (unless i'm overlooking some detail where a parameter is different) and both inherit from the same parent - so i think they should be / remain in the parent class.

freqtrade/freqai/prediction_models/XGBoostRFRegressor.py
robcaulk: No, just tossed it in there when I noticed it. It is just adding tensorboard to XGBRFRegressor.

freqtrade/freqai/base_models/BasePyTorchClassifier.py
robcaulk: Yes, the regressor also has a pipeline for the labels. Meanwhile the classifier does not.

freqtrade/freqai/RL/BaseReinforcementLearningModel.py
```diff
@@ -107,27 +110,37 @@ def train(
...
-        self.df_raw = copy.deepcopy(data_dictionary["train_features"])
+        self.df_raw = copy.deepcopy(dd["train_features"])
         dk.fit_labels()  # FIXME useless for now, but just satiating append methods
 
         # normalize all data based on train_dataset only
         prices_train, prices_test = self.build_ohlc_price_dataframes(dk.data_dictionary, pair, dk)
 
-        data_dictionary = dk.normalize_data(data_dictionary)
+        dk.feature_pipeline = self.define_data_pipeline(threads=dk.thread_count)
+        dk.label_pipeline = self.define_label_pipeline(threads=dk.thread_count)
```
robcaulk: This should be removed, RL doesnt need a label pipeline.

freqtrade/freqai/freqai_interface.py
robcaulk: Done, moved to `constants.py`

requirements-freqai.txt
robcaulk: Done

freqtrade/freqai/RL/BaseReinforcementLearningModel.py
```diff
@@ -236,18 +248,19 @@ def predict(
...
-        self.data_cleaning_predict(dk)
+        dk.data_dictionary["prediction_features"], outliers, _ = dk.feature_pipeline.transform(
+            dk.data_dictionary["prediction_features"], outlier_check=True)
 
         pred_df = self.rl_model_predict(
             dk.data_dictionary["prediction_features"], dk, self.model)
         pred_df.fillna(0, inplace=True)
 
+        if self.freqai_info.get("DI_threshold", 0) > 0:
+            dk.DI_values = dk.feature_pipeline["di"].di_values
```
xmatthias: Shouldn't this test be `if 'di' in dk.feature_pipeline:`? 

It'll work for regular cases - but assuming a user uses a config with `DI_threshold` set - but implements `define_data_pipeline` - then this can cause an implicit crash, as the configuration does no longer correspond to the generated pipeline.

(obviously not just in this file...)

robcaulk: good catch, fixed.

robcaulk:
This is ready to go on my side.

xmatthias (approved):
LGTM :+1:

PR #8696, opened 2023-05-27
allow no / empty minimal_roi

xmatthias:
## Summary

Allow complete disabling of minimal_roi by providing an empty dictionary (which is also the default now).


froggleston (approved):
Looks good, and will clear up some long term confusion about how to work around it currently.

PR #8760, opened 2023-06-11
Add MaskablePPO support

initrv:
## Summary

The goal of this PR is to add ```sb3_contrib``` ```MaskablePPO``` basic support to FreqAI RL.

Solve the issue: 
```"Environment does not support action masking. Consider using ActionMasker wrapper"```
when using the ```model_type``` ```MaskablePPO```

## Quick changelog

- Action masking support during learning

## What's new?

```MaskablePPO``` works out of the box

### DIFF
--- a/freqtrade/freqai/RL/BaseEnvironment.py
+++ b/freqtrade/freqai/RL/BaseEnvironment.py
@@ -141,6 +141,9 @@ class BaseEnvironment(gym.Env):
         Unique to the environment action count. Must be inherited.
         """
 
+    def action_masks(self) -> list[bool]:
+        return [self._is_valid(action.value) for action in self.actions]
+
     def seed(self, seed: int = 1):
         self.np_random, seed = seeding.np_random(seed)
         return [seed]
--- a/freqtrade/freqai/RL/BaseEnvironment.py
+++ b/freqtrade/freqai/RL/BaseEnvironment.py
@@ -2,7 +2,7 @@ import logging
 import random
 from abc import abstractmethod
 from enum import Enum
-from typing import Optional, Type, Union
+from typing import List, Optional, Type, Union
 
 import gymnasium as gym
 import numpy as np
@@ -141,7 +141,7 @@ class BaseEnvironment(gym.Env):
         Unique to the environment action count. Must be inherited.
         """
 
-    def action_masks(self) -> list[bool]:
+    def action_masks(self) -> List[bool]:
         return [self._is_valid(action.value) for action in self.actions]
 
     def seed(self, seed: int = 1):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by robcaulk — COMMENTED** [review]
Thanks for this, if I understand correctly, `action_masks()` is being called somewhere during training. But dont we need to also add `action_masks` argument to the `predict()` method when we inference into the trained model? 

I am mostly just reviewing [this doc](https://sb3-contrib.readthedocs.io/en/master/modules/ppo_mask.html) to put the pieces together.

**Comment by initrv** [comment]
The model now ignores invalid actions during training, but is still capable of producing them during prediction. We can construct the current predition mask depending on ```market_side``` from ```state_info``` and pass it to ```predict(action_masks=)```.
 However, the use of masks is optional outside of learning. And I haven't managed with a proper implementation that would account for all models and scenarios, so I have submitted the PR without masked predictions for now.

But I have some ideas about it. In my own environments, I modified ```_is_valid``` to call it outside and pass the position regardless of the state of the environment

```python
def is_valid(self, action: int, position: float = None) -> bool:
    if not position:
        position = self._position.value
    ...
```

So we can build a mask during predict() in a similar way as in the environment
```python
def _action_masks_predict(self, market_side: float) -> List[bool]:
    return [self.train_env.is_valid(action.value, market_side) for action in Actions]
```

```python
model.predict(action_masks=self._action_masks_predict(market_side))
```

I don't think that's a good way to do it. We need to take several factors into account such as the type of user environment, whether vectorization is used, whether DCA is needed, etc.

robcaulk (approved):
LGTM

robcaulk:
> The model now ignores invalid actions during training, but is still capable of producing them during prediction. We can construct the current predition mask depending on `market_side` from `state_info` and pass it to `predict(action_masks=)`. However, the use of masks is optional outside of learning. And I haven't managed with a proper implementation that would account for all models and scenarios, so I have submitted the PR without masked predictions for now.
> 
> But I have some ideas about it. In my own environments, I modified `_is_valid` to call it outside and pass the position regardless of the state of the environment
> 
> ```python
> def is_valid(self, action: int, position: float = None) -> bool:
>     if not position:
>         position = self._position.value
>     ...
> ```
> 
> So we can build a mask during predict() in a similar way as in the environment
> 
> ```python
> def _action_masks_predict(self, market_side: float) -> List[bool]:
>     return [self.train_env.is_valid(action.value, market_side) for action in Actions]
> ```
> 
> ```python
> model.predict(action_masks=self._action_masks_predict(market_side))
> ```
> 
> I don't think that's a good way to do it. We need to take several factors into account such as the type of user environment, whether vectorization is used, whether DCA is needed, etc.

Yeah I see your point, in this case we can leave it as is.

PR #8779, opened 2023-06-13
Feature: use order object instead of open_order_id field

Axel-CH:
## Summary

The Goal of this PR is to prepare the possibility to allow multiple orders to be opened for a Trade. As @xmatthias stated [here](https://github.com/freqtrade/freqtrade/issues/7499#issuecomment-1585515491)

Solve the issue: #7499

## Quick changelog
Removed open_order_id property in Trade class
Add open_orders property in Trade class
Update test_update_limit_order

## What's new?
One test can work with the new open order management. To go further we need to validate the changes/direction.

`pytest tests/persistence/test_persistence.py::test_update_limit_order`


Axel-CH:
First Questions:

**1/ Order groups?**
    Do you think that ORM requests complexity could be reduced with a new Order property "group"
        ->Add property to Order "ft_action" or "ft_order_group" ? [entry, exit, stoploss]
        -> The Goal is to answer in a more efficient way to thooses questions
            -> Do I have exit orders open? how many? what is the sum amount of thooses exit order?

**2/ Existing bugfix outdated?**
    In trade_model.py::update_trade function; What is this "condition to avoid reset value when updating fees" a bugfix?
    -> Do you think this bugfix will be relevant after the open_order_id removal?


**3/ What about trade.stoploss_order_id**
    ->trade.stoploss_order_id" seem to fall in the same category of implied limitations of "trade.open_order_id"
    What do you think? (Will not touch that in that PR, just want to have you thought)

freqtrade/persistence/migrations.py
```diff
@@ -294,6 +295,28 @@ def fix_old_dry_orders(engine):
...
 
+        # Update current Order where
+        # -current Order is open
+        # -current Order trade_id not equal to current Trade.id
+        # -current Order not stoploss
+        # -order_id not equal to current Trade.stoploss_order_id
+        # -current Order is dry
+
+        # NEW WIP
+        stmt = update(Order).where(
```
xmatthias: What's the reason for this update? (i understand what it does, but it seems unnecessary unless i miss something?)

dry-run orders can stay open without problems - and will be loaded from the database.

xmatthias:
From a title perspective (not necessarily from the code) i think this goes too far.
i'd want to still keep the limit to "1 open order at a time" for now - even though it'll be enforced by code only, not by "we only have one place to store it".

> **1/ Order groups?** Do you think that ORM requests complexity could be reduced with a new Order property "group" ->Add property to Order "ft_action" or "ft_order_group" ? [entry, exit, stoploss] -> The Goal is to answer in a more efficient way to thooses questions -> Do I have exit orders open? how many? what is the sum amount of thooses exit order?

I'd not worry too much about request complexity at this point - we do start from an open trade when checking orders - which implicitly loads all orders through the link (trade.orders evaluates immediately except in one case).

Your properties are therefore simple list comprehensions on objects in memory (not pushed to the DB) - and should therefore not really impact performance.

Aside from that - i think you can get this with ft_order_side - even though that'll be buy/sell/stop - not entry/exit/stop.



> **2/ Existing bugfix outdated?** In trade_model.py::update_trade function; What is this "condition to avoid reset value when updating fees" a bugfix? -> Do you think this bugfix will be relevant after the open_order_id removal?

I assume you're talking about the "Got different open_order_id" log - which shouldn't really matter at this point.
we'll see if it causes problems while running / testing this - but i'd not worry about this too much rn

> **3/ What about trade.stoploss_order_id** ->trade.stoploss_order_id" seem to fall in the same category of implied limitations of "trade.open_order_id" What do you think? (Will not touch that in that PR, just want to have you thought)

we WILL have to touch / remove that, too.
i'll leave it up to you if we do that now or later - but it's a requirement for OCO order support in my opinion.

freqtrade/persistence/migrations.py
Axel-CH: I was trying to replicate existing code but without "open_order_id" and the new logic. You mean I can leave migration functions unchanged and it will be remove after?

xmatthias: well i'd leave it for now and work on regular functionality (backtesting, trading).

There's WAY more important thing to do than THIS small update (which is per se not REALLY necessary unless you migrate from a VERY old database).

freqtrade/freqtradebot.py
```diff
@@ -1436,8 +1446,8 @@ def cancel_all_open_orders(self) -> None:
         Trade.commit()
 
     def handle_cancel_enter(
```
Axel-CH, line 1456: What do you think about that? as the order_id is not explicitly in the trade object, I need to pass it to the function. I don't find this very elegant.

Axel-CH: @xmatthias

freqtrade/freqtradebot.py
```diff
@@ -1320,34 +1321,41 @@ def manage_open_orders(self) -> None:
...
-            except (ExchangeError):
-                logger.info('Cannot query order for %s due to %s', trade, traceback.format_exc())
-                continue
 
-            fully_cancelled = self.update_trade_state(trade, trade.open_order_id, order)
-            not_closed = order['status'] == 'open' or fully_cancelled
-            order_obj = trade.select_order_by_order_id(trade.open_order_id)
+                fully_cancelled = self.update_trade_state(trade, open_order.order_id, order)
+                not_closed = order['status'] == 'open' or fully_cancelled
+                order_obj = trade.select_order_by_order_id(open_order.order_id)
```
xmatthias: I think you can omit this now, and use `open_order` instead.

`open_order` is already the order-object - so this is really irrelevant.

freqtrade/freqtradebot.py
xmatthias: I'm not entirely sure i see  / understand the problem? 
you'll always need a way to know WHICH order to cancel.

imagine this with true multi-order support (and the call from "replace-order") - you'll need to know if it's entry order A or B that you'd like to cancel.
i guess you could use the order object instead of the order-id - but that won't really make a difference, and will result in `order.order_id` instead of `order_id` from a usage perspective.

Axel-CH: Yes I know that it is required to have the order id. It's just a bit confusing to have to order related parameters.
I it's not a big deal for you I will leave it like that. I have plenty of tests to fix

freqtrade/freqtradebot.py
Axel-CH: Good catch

freqtrade/freqtradebot.py
```diff
@@ -1236,11 +1236,16 @@ def handle_stoploss_on_exchange(self, trade: Trade) -> bool:
...
-            # Trade has an open Buy or Sell order, Stoploss-handling can't happen in this case
-            # as the Amount on the exchange is tied up in another trade.
-            # The trade can be closed already (sell-order fill confirmation came in this iteration)
-            return False
+        print("***************open_orders DEBUG***************")
+        print(f"trade.open_orders: {trade.open_orders}")
+        print(f"trade.open_orders_count: {trade.open_orders_count}")
+        print(f"trade.open_entry_or_exit_orders_count: {trade.open_entry_or_exit_orders_count}")
+
+        # if trade.open_entry_or_exit_orders_count != 0 or not trade.is_open:
```
Axel-CH: The function is not working with that code block, From my understanding this issue of tied capital is only present on spot, can you confirm? Do you have an idea of how we can achieve this? Why not letting the exchange return an error and handling it?

freqtrade/persistence/trade_model.py
```diff
@@ -468,6 +469,23 @@ def safe_quote_currency(self) -> str:
         except IndexError:
             return ''
 
+    @hybrid_property
+    def open_orders_count(self) -> int:
+        return len(self.open_orders)
+
+    @hybrid_property
```
xmatthias: what's the reason to use `@hybrid_property` over `@property`? (i think the latter should be used).

in my understanding, the "hybrid" is needed to define a property that works both at instance and class level - which doesn't seem to make sense here (what would `Trade.open_orders_count` even mean?

freqtrade/freqtradebot.py
xmatthias: the "tied up" issue is really depending on the exchange, and how they actually implement the stop order types.

on binance, it's only spot - on futures, i think it depends if the exchange supports reduce-only orders.
if they don't - then you risk reversing your position accidentally by having both a stop and limit order open (and triggered).

i don't really see why this shouldn't be working here - we're in a trade scoped method and have the trade object available ... 
if `trade.open_entry_or_exit_orders_count` is not returning the right value, then there's a problem with that method i'd assume.

having the exchange fail it intentionally is problematic and not a solution.
it'll never succeed place a stoploss, but _attempt_ to place one at any iteration (until an eventually order tying up the capital is cancelled or closed). That'll lead to TONS of bad log messages - as well as potentially problems with exchange ratelimits.

freqtrade/persistence/trade_model.py
```diff
@@ -680,24 +700,33 @@ def update_trade(self, order: Order) -> None:
...
-            if self.open_order_id == order.order_id:
-                self.open_order_id = None
-            else:
-                logger.warning(
-                    f'Got different open_order_id {self.open_order_id} != {order.order_id}')
+            # if self.open_order_id == order.order_id:
+            #    self.open_order_id = None
+            # else:
+            #    logger.warning(
+            #        f'Got different open_order_id {self.open_order_id} != {order.order_id}')
```
xmatthias: you can't perform these tests (and assignment) anymore - so i don't think they're relevant any longer (both here and above).

freqtrade/persistence/trade_model.py
Axel-CH: You are right will change that

freqtrade/freqtradebot.py
Axel-CH: Ok I will check more deeply but the `trade.open_entry_or_exit_orders_count` seem to work fine, maybe it not adequately used yet.

***************open_orders DEBUG***************
trade.open_orders: [Order(id=1, order_id=mocked_limit_sell_usdt, trade_id=1, side=None, filled=0.0, price=2.2, order_type=None, status=None), Order(id=2, order_id=100, trade_id=1, side=None, filled=0.0, price=2.0, order_type=stop_loss_limit, status=open)]
trade.open_orders_count: 2
trade.open_entry_or_exit_orders_count: 1

xmatthias: did you notice it in tests - or in actually running the bot?
I'm asking because with tests - it could just as well be caused by inadequate test setup.

We've improved this over time - but not all tests require 100% valid trades - so they're not always perfectly aligned.

tests/test_freqtradebot.py
```diff
@@ -6302,7 +6286,7 @@ def test_position_adjust2(mocker, default_conf_usdt, fee) -> None:
 
     trade = Trade.session.scalars(select(Trade)).first()
     assert trade
-    assert trade.open_order_id is None
+    assert trade.open_entry_or_exit_orders_count == 0
```
xmatthias: You might wanna introduce a new property in `has_open_orders`.
this is quite a common check i think - so it'll be handy (even if we might only use it in tests).

freqtrade/freqtradebot.py
Axel-CH: I've got this logs from the tests, I will check the test

tests/test_freqtradebot.py
Axel-CH: Yes but in that case it could be more like `has_entry_or_exit_orders` as some time we don't want to take stop loss in account.

xmatthias: well i'd exclude stop orders anyway.

They need a separate property - as i know no location (at least not by memory) where we check both open_order_id as well as stoploss_order_id in the same place  ;)

freqtrade/freqtradebot.py
Axel-CH: In fact pytest tests/test_freqtradebot.py::test_handle_stoploss_on_exchange is working but it is other test cases using the function that are failing, I think it could be more related to bad mock order initialisation than an issue within the function.

tests/test_freqtradebot.py
Axel-CH, line 1: @xmatthias  I'm having issues fixing the test `test_handle_sle_cancel_cant_recreate` could you take a look at it to check if the test is fine?

freqtrade/persistence/trade_model.py
```diff
@@ -1296,6 +1327,90 @@ class Trade(ModelBase, LocalTrade):
...
+
+    @has_open_orders.expression
+    def has_open_orders(cls) -> int:
+        return (
+            select(func.exists())
+            .where(Order.ft_is_open.is_(True))
+            .where(Order.ft_order_side != "stoploss")
+            .where(Order.ft_trade_id == cls.id)
+            .as_scalar()
+        )
```
xmatthias: I can't think of a case where it's faster to go through the database than to do a in-memory list comprehension to get the same result.

as mentioned before  - `.orders` is a list that's filled in all cases when querying the trade from the database.

That applies to basically all methods below as well that use the same pattern.
to me, all of these functions should be implemented in the `LocalTrade` class only - so the same logic will work for backtesting, as well as live.

tests/test_freqtradebot.py
```diff
@@ -2888,11 +2888,15 @@ def test_adjust_entry_maintain_replace(
     assert freqtrade.strategy.adjust_entry_price.call_count == 1
 
     # Check that order is replaced
+    # TODO replace order at a price that can't be fullfilled
```
Axel-CH: @xmatthias  How can I create an order that is not immediately filled? Probably by setting an entry price far from market price, But I'm not yet familiar with the mocking process of current market price, Any advise or example test? (looking for one at the moment)

freqtrade/persistence/trade_model.py
Axel-CH: Fixing that today, I'm suspecting this causing some of the the 15 failing tests on test_freqtradebot

freqtrade/persistence/trade_model.py
```diff
@@ -1296,6 +1326,40 @@ class Trade(ModelBase, LocalTrade):
...
+
+        return len(open_buy_or_sell_orders)
+
+    @property
+    def open_orders_ids(self) -> list:
+        open_orders_ids_wo_sl = []
+        for oo in self.open_orders:
+            if (oo.ft_order_side not in ['stoploss']):
+                open_orders_ids_wo_sl.append(oo.order_id)
+        return open_orders_ids_wo_sl
```
xmatthias: Considering Trade is a subclass of LocalTrade - i think these are duplicates (quite dangerous!) / not needed? (unless i miss some different logic).

tests/test_freqtradebot.py
xmatthias: The problem here is a different one - one i'm not sure will be straightforward to fix as this PR (as it'll taint the actual changes).

`open_trade`  is incomplete - the INTEND is to have an open exit order - not an open entry order (at least not the way the order is setup - which makes it closed the moment "any" check is applied).

`open_trade_usdt` does this correctly - and will therefore be better suited for this test - but changing the test from one to the other fixture might require other changes, too (as it'll behave differently) - and should be done separately.

freqtrade/rpc/rpc.py
```diff
@@ -763,21 +763,22 @@ def _rpc_reload_trade_from_exchange(self, trade_id: int) -> Dict[str, str]:
...
         fully_canceled = False
-        if trade.open_order_id:
-            order = self._freqtrade.exchange.fetch_order(trade.open_order_id, trade.pair)
+        for oo in trade.open_orders:
+            order = self._freqtrade.exchange.fetch_order(oo.order_id, trade.pair)
 
             if order['side'] == trade.entry_side:
                 fully_canceled = self._freqtrade.handle_cancel_enter(
-                    trade, order, CANCEL_REASON['FORCE_EXIT'])
+                    trade, order, oo.order_id, CANCEL_REASON['FORCE_EXIT'])
```
xmatthias, line 798: you'll need to be careful here
changing if with for will change the context - and `fully_canceled` will have the result of the last order - which may or may not be correct behavior

Axel-CH: True, I changed that

freqtrade/persistence/trade_model.py
```diff
@@ -1418,7 +1450,7 @@ def get_open_order_trades() -> List['Trade']:
         Returns all open trades
         NOTE: Not supported in Backtesting.
         """
-        return cast(List[Trade], Trade.get_trades(Trade.open_order_id.isnot(None)).all())
+        return cast(List[Trade], Trade.get_trades([Trade.open_orders_count != 0]).all())
```
xmatthias: This is wrong, and causes at least one test to fail (namely `test_cancel_all_open_orders`), and can potentially lead to quite some inefficiencies, maybe even bugs or oddities when it comes to orders going "missing" (or orders being processed multiple times).

`Trade.open_orders_count` is a property - so while this looks good at first - the resulting query is the following (notice the empty where clause).


```
'SELECT trades.id, trades.exchange, trades.pair, trades.base_currency, trades.stake_currency, trades.is_open, trades.fee_open, trades.fee_open_cost, trades.fee_open_currency, trades.fee_close, trades.fee_close_cost, trades.fee_close_currency, trades.open_rate, trades.open_rate_requested, trades.open_trade_value, trades.close_rate, trades.close_rate_requested, trades.realized_profit, trades.close_profit, trades.close_profit_abs, trades.stake_amount, trades.max_stake_amount, trades.amount, trades.amount_requested, trades.open_date, trades.close_date, trades.stop_loss, trades.stop_loss_pct, trades.initial_stop_loss, trades.initial_stop_loss_pct, trades.stoploss_order_id, trades.stoploss_last_update, trades.max_rate, trades.min_rate, trades.exit_reason, trades.exit_order_status, trades.strategy, trades.enter_tag, trades.timeframe, trades.trading_mode, trades.amount_precision, trades.price_precision, trades.precision_mode, trades.contract_size, trades.leverage, trades.is_short, trades.liquidation_price, trades.interest_rate, trades.funding_fees \nFROM trades \nWHERE true'
```

Axel-CH: You are right this is not good and also for another reason "Trade.open_orders_count != 0 " is not similar to "Trade.open_order_id.isnot(None)" as It will also return stoploss orders.

Axel-CH: Do you think Trade.is_open.is_(True) filter could be good? or we need to checks withing orders array of Trade

xmatthias: i didn't "really" look at the details of where it's used.

currently (develop) - the desire is to select only trades with an active (non stoploss) order. 
If we change it to "is_open" - then i think the whole method should be removed / replaced with something else - as i think we do have a method to get "open trades".

freqtrade/persistence/trade_model.py
```diff
@@ -589,11 +619,11 @@ def to_json(self, minified: bool = False) -> Dict[str, Any]:
             'is_short': self.is_short,
             'trading_mode': self.trading_mode,
             'funding_fees': self.funding_fees,
-            'open_order_id': self.open_order_id,
             'amount_precision': self.amount_precision,
             'price_precision': self.price_precision,
             'precision_mode': self.precision_mode,
-            'orders': orders,
+            'orders': orders_json,
+            'open_orders': open_orders_json
```
xmatthias: Maybe stupid question - but why exactly do we need `open_orders` ? 

`orders` already includes open orders - so it's just a question of filtering this.
Adding both however means that the same open order object will be serialized twice - for every RPC call - which is every few seconds if freqUI is open ...

Axel-CH: Yes you are right it is duplicated data, open orders could be retrieved by filtering from 'orders'

freqtrade/rpc/telegram.py
```diff
@@ -663,10 +663,10 @@ async def _status_msg(self, update: Update, context: CallbackContext) -> None:
                              ("`({stop_loss_ratio:.2%})`" if r['stop_loss_ratio'] else ""))
                 lines.append("*Stoploss distance:* `{stoploss_current_dist:.8g}` "
                              "`({stoploss_current_dist_ratio:.2%})`")
-                for order in r.get('open_orders', []):
+                if r.get('open_orders'):
                     lines.append(
-                        f"*Open Order:* `{order}`"
-                        + "- `{exit_order_status}`" if r['exit_order_status'] else "")
+                        "*Open Order:* `{open_orders}`"
```
Axel-CH, line 668: We don't need to show details for each order individually?

xmatthias: we do - but "open_orders" does NOT contain what you think it does.
it's created in `rpc.py` - which is already combining all orders into one combined string.

Axel-CH: I see, thank you for the clarification

xmatthias (approved):
LGTM :+1:

PR #8790, opened 2023-06-16
Add links to more FreqAI learning content

robcaulk:
As have written in depth content describing various components of freqai, with pictures, code sample descriptions, strategy files and configs. We would like to make sure FreqAI users can find this content if they are willing to dive deeper into the rabbit hole. Here we add the links.


xmatthias (approved):
LGTM :+1:

PR #8819, opened 2023-06-24
Remotepairlist - add blacklist mode

Bloodhunter4rc:
## Summary

Add blacklist mode to remotepairlist, as requested in discord: (https://discord.com/channels/700048804539400213/824719649417199646/1097212913393074197)

## Quick changelog

- Add Blacklist Mode
- Adhere to Number of Assets

## What's new?

Add Blacklist Mode

## TODO

- [x] Add Blacklist
- [x] Allow Wildcard


freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -250,6 +251,25 @@ def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
...
+                    merged_list.append(pair)
+                else:
+                    filtered.append(pair)
+            if filtered:
+                self.log_once(f"Blacklist - Filtered out pairs: {filtered}", logger.info)
+
+        else:
+            raise OperationalException(
+                '`mode` not configured correctly. Supported Modes: '
+                'are "whitelist","blacklist"')
```
xmatthias: You might wanna move this to __init__.py, so it's failing immediately, not after potentially minutes (depending what comes before RemotePairlist).

xmatthias (commented):
Seems like a great enhancement :+1: 

Can you please have a look at the tests as well? should be easy enough to derive from the existing tests ...

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: moved it to init, and added tests

Bloodhunter4rc:
I will add the expand_pairlist helper to allow wildcards.

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -250,6 +258,20 @@ def filter_pairlist(self, pairlist: List[str], tickers: Dict) -> List[str]:
         :return: new whitelist
         """
         rpl_pairlist = self.gen_pairlist(tickers)
-        merged_list = pairlist + rpl_pairlist
-        merged_list = sorted(set(merged_list), key=merged_list.index)
+        merged_list = []
+        filtered = []
+
+        if self._mode == "whitelist":
+            merged_list = pairlist + rpl_pairlist
```
xmatthias: i know this didn't really change - but is this intentional? 

Assuming i use "volumepairlist 50" - and then "remotePairlist 5 (currently with a local file containing 3 pairs)" - would i then expect 55 pairs, or rather 5 pairs (volumepairlist further filtered onto pairs actually in the remote pairlist)?

All other pairlists would have this act as a filter.
It's not necessarily a problem to have it act this way - but it's not immeditately clear that it'll not act as a filter if it's not in "generator" mode.

xmatthias (changes requested):
I've taken the liberty to add the new parameter to the "available parameters" function - so it can be used through the UI.
Feel free to update the comment if you think it could be different.

image



Unfortunately, "blacklist" is ignored if the pairlist is used as a generator, resulting in the list to be used as whitelist (although explicitly configured as blacklist).

Testing with the following pairlist configuration - i'd expect it to either throw an error at me (a blacklist cannot be a generator, only a filter!) - or to produce no pairs (not necessarily nice behavior, as it's clearly not a desired state).

``` json
[
  {
    "method": "RemotePairList",
    "mode": "blacklist",
    "number_assets": 20,
    "pairlist_url": "file:///mememe.json",
    "refresh_period": 1800,
    "read_timeout": 60
  }
]
```

To me, it's a question of an additional check in `__init__`  - failing this as invalid configuration - but maybe this was done intentionally?

freqtrade/plugins/pairlist/RemotePairList.py
Bloodhunter4rc: It was intentional, actually i thought about adding a option to chose if it is a filter or a addition to the current whitelist.. i added that now, as it makes sense to want both, filter is the default action now  as thats how other pairlists behave.

Edit: nice i like the gui addition, that was an oversight i will add the init check

freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -57,6 +58,23 @@ def __init__(self, exchange, pairlistmanager,
...
+
+        if len(pairlists) == 1 and self._mode == 'blacklist':
+            raise OperationalException(
+                'At `blacklist` mode RemotePairList requires an additional '
+                'Pairlist and cannot be used on its own.')
+
+        if pairlists[0]['method'] == "RemotePairList" and self._mode == 'blacklist':
+            raise OperationalException(
+                'At `blacklist` mode RemotePairList can not be on the first '
+                'position of your pairlist.')
```
xmatthias: You should be able to use `pairlist_pos` to check for this in one go
if the position is 0, it's the first pairlist - and can't be used in blacklist mode.

the "on it's own" check is really irrelevant then if you think about it.

xmatthias (approved):
Great enhancement :100:  

LGTM :+1:

PR #8835, opened 2023-06-26
make sure default PCA behavior reduces parameter space size

robcaulk:
PCA by default does not reduce the parameter space (i.e. number of features), but  users typically want to use it to reduce the feature count in an effort to improve training/inference computational efficiency.

We set the default behavior to keeping 0.999 explained variance. Power users who want more fine tuned control over the pipeline should still simply build their own by inheriting the `define_data_pipeline()` function.


xmatthias (approved):
LGTM :+1:

PR #8863, opened 2023-07-07
Bump datasieve 0.1.7

robcaulk:
Bump to new datasieve version allowing all higher versions of pandas.


xmatthias (approved):
Thanks for the prompt update! :100: 

LGTM :+1:

PR #8899, opened 2023-07-12
Change default dataformat to feather

xmatthias:
## Summary

Change the default dataformat to feather, based on tests, the most efficient dataformat.

This will be a breaking change for users who don't have "dataformat_ohlcv" configured explicitly, but full backwards compatibility can be archived by using the following in the configuration:

``` json
{
    "dataformat_ohlcv": "json",
    "dataformat_trades": "jsongz"
}
```

in case the data is still in json format, we recommend to convert it using the `convert-data` subcommand.

## Quick changelog

- Change default dataformat to feather (for both ohlcv and trades data)


tomjrtsmith:
Hi Mathias, this has me a bit confused. This is a breaking change that
requires that Freqtrade installs all have the following in their config
files to remain running? Is this correct?

{
    "dataformat_ohlcv": "json",
    "dataformat_trades": "jsongz"
}


 If so, where's the best place to put it?

stash86:
https://github.com/freqtrade/freqtrade/blob/develop/config_examples/config_full.example.json

tomjrtsmith:
Thanks Stephan.

xmatthias:
A better approach is to convert the data to feather. 
Json is the default as it's the oldest format we support - but at the same time - it's also the slowest format we support (both reading and writing). We have a [comparison table](https://www.freqtrade.io/en/stable/data-download/#dataformat-comparison) - after seeing the results, you'll understand the reason for that change.

PR #8903, opened 2023-07-13
Freqai pytorch bugfixes

Yinon-Polak:
## Summary

Few bug fixes and refactoring of freqai PyTorch module.

## Quick changelog

- fixed bug where tensors were not being moved to device
- fixed bug load model to current device raised exception
- changes in data convertor to initialize with specific dtype and directly on device 
- adding optional n_epochs hyper param


freqtrade/freqai/torch/PyTorchModelTrainer.py
```diff
@@ -182,8 +181,8 @@ def save(self, path: Path):
             "pytrainer": self
         }, path)
 
-    def load(self, path: Path):
-        checkpoint = torch.load(path)
+    def load(self, path: Path, device: Optional[str] = None):
```
robcaulk: I am not sure I see the use-case for adding this argument. `map_location` is useful for loading a model that was trained on GPU into a computer that does not have a GPU, so that the model can be inferenced. This scenario is not something a FreqAI user would find themselves in. 

We hold all models in RAM so this `load` is only called when a user calls `/reload` on their bot. Then they would load the models back into ram and continue where they left off (presumably on the same computer). Now you could say that a user might want to switch computers, but this introduces many other issues that FreqAI is not designed to support.

Second issue - how it is currently setup, the user can never actually access this argument. `data_drawer.py` the `trainer.load()` is called without any additional arguments right now. 

Let me know what your intention is for this argument, and let me know if I missed a possible use-case for FreqAI users. Otherwise, I would prefer not to introduce this without clear reasoning.

freqtrade/freqai/torch/PyTorchModelTrainer.py
```diff
@@ -40,23 +39,27 @@ def __init__(
...
         :param batch_size: The size of the batches to use during training.
-        :param max_n_eval_batches: The maximum number batches to use for evaluation.
         """
         self.model = model
         self.optimizer = optimizer
         self.criterion = criterion
         self.model_meta_data = model_meta_data
         self.device = device
         self.max_iters: int = kwargs.get("max_iters", 100)
+        self.n_epochs: Optional[int] = kwargs.get("n_epochs", None)
```
robcaulk: I am glad that we are switching to let users define epochs directly from the config. I felt that the original way was a bit hard for users to control/understand.

I would suggest we actually use `n_epochs` by default instead of `max_iters`. So I would swap `kwargs.get("max_iters", None)` and `kwargs.get("n_epochs", 10)`.


 Can we also add this to the parameter list documentation so that users know they can access this?

freqtrade/freqai/torch/PyTorchModelTrainer.py
```diff
@@ -72,55 +75,51 @@ def fit(self, data_dictionary: Dict[str, pd.DataFrame], splits: List[str]):
...
+                xb = xb.to(self.device)
+                yb = yb.to(self.device)
                 yb_pred = self.model(xb)
                 loss = self.criterion(yb_pred, yb)
 
                 self.optimizer.zero_grad(set_to_none=True)
                 loss.backward()
                 self.optimizer.step()
-                self.tb_logger.log_scalar("train_loss", loss.item(), i)
+                self.tb_logger.log_scalar("train_loss", loss.item(), batch_counter)
```
robcaulk, line 95: Thanks for finding this bug!

freqtrade/freqai/torch/PyTorchModelTrainer.py
```diff
@@ -72,55 +75,51 @@ def fit(self, data_dictionary: Dict[str, pd.DataFrame], splits: List[str]):
...
     ) -> None:
         self.model.eval()
-        n_batches = 0
-        for i, batch_data in enumerate(data_loader_dictionary[split]):
-            if max_n_eval_batches and i > max_n_eval_batches:
-                n_batches += 1
-                break
+        for _, batch_data in enumerate(data_loader_dictionary[split]):
             xb, yb = batch_data
             xb.to(self.device)
```
robcaulk: Above you set `xb = xb.to(self.device)` but here I think you forgot to change it. We need to do it here too (thanks for finding that bug too).

robcaulk (commented):
Thanks for these improvements :-) Please let me know if I missed something with respect to how you want to use `map_device` to help FreqAI users.

freqtrade/freqai/torch/PyTorchModelTrainer.py
Yinon-Polak: yes, this is intended for switching  between a "training" machine to "trading" vm.
what's preventing one from running a start on a gpu vm and train a model.
then load this model on a different a different vm (cpu only) where the bot will run ? 

i know that currently there is no explicit calling for it. but the reason that the trainer has two loading function, `load` and `load_from_checkpoint`. is to allow this flexibility. training large model's on cpu vm will create a huge bottleneck.

also, this won't effect the standard use case.

freqtrade/freqai/torch/PyTorchModelTrainer.py
Yinon-Polak: I'll add `n_epochs` to the docs.

I am considering removing the defaults altogether since they were never intended to function as "good" default values. What are your thoughts on this?

freqtrade/freqai/torch/PyTorchModelTrainer.py
robcaulk: If we want to support this use-case, then we need to document how a user would do it, and the arguments need to be used somewhere in the the code. In other words, how would a user train on GPU, then configure FreqAI to deploy on CPU? This should be capable using the FreqAI config.

One issue is that if a user does do this, then it means that they do not want adaptive modeling. They are deploying with a static model. That is not necessarily a problem - and it is currently possible by setting `live_retrain_hours` to a high number. But it is not the common use case at the moment. 

To be clear, I am not against adding this kind of functionality. It is just that with your current modifications it is only half done. We need to add the full functionality for users.

freqtrade/freqai/torch/PyTorchModelTrainer.py
robcaulk: You can remove them if you like. In this case we would want to catch the values if they are undefined and yield a descriptive error.

Yinon-Polak: I've updated the docs.
as for the defaults.., i'll leave for to the next PR

freqtrade/freqai/torch/PyTorchModelTrainer.py
Yinon-Polak: I removed this for now, will try to add this functionality in the future.

robcaulk:
Thanks for this, I am sorry for the delay in the final review. Looks great.

robcaulk (approved):

PR #8912, opened 2023-07-15
Add expectancy to RPC calls and telegram

froggleston:
## Summary

Add expectancy, expectancy rate and winrate to the RPC trade statistics and corresponding telegram command (`/profit`)

## Quick changelog

- Add winrate, expectancy and expectancy_rate calculations into rpc.py `_rpc_trade_statistics`
- Add corresponding values into telegram /profit message
- Add tests


freqtrade/rpc/rpc.py
```diff
@@ -576,6 +596,9 @@ def _rpc_trade_statistics(
             'winning_trades': winning_trades,
             'losing_trades': losing_trades,
             'profit_factor': profit_factor,
+            'winrate': winrate,
+            'expectancy': expectancy,
+            'expectancy_rate': expectancy_rate,
```
xmatthias: if you intend to expose these through the API, too (which we should) - you'll need to add them to api_schemas for the corresponding endpoint.

froggleston: Ah I knew I forgot something!

xmatthias: be sure to double-check api tests - not sure how this endpoint is tested ... but the corresponding one might fail after this.

froggleston: I swear if I get another trailing whitespace precommit I'll explode.

xmatthias: fix your editor? trailing commits can be removed by vscode automatically ... 
just make sure to disable it for *.md - 2 trailing spaces there are a paragraph :laughing:

xmatthias (commented):
Great enhancement :100: 


Unfortunately, it seems to be non-functional at the moment, so it's difficult to test this further.
When trying this via telegram, i get the following error:

```
2023-07-16 16:14:35,624 - freqtrade.rpc.telegram - WARNING - Telegram NetworkError: Can't parse entities: can't find end of the entity starting at byte offset 591! Trying one more time.
2023-07-16 16:14:35,655 - freqtrade.rpc.telegram - WARNING - TelegramError: Can't parse entities: can't find end of the entity starting at byte offset 591! Giving up on that message.
```

froggleston:
Looks like I was missing a single backtick so the newline was throwing off the markdown processor.

freqtrade/rpc/rpc.py
```diff
@@ -576,6 +589,9 @@ def _rpc_trade_statistics(
...
```
xmatthias: image

i've tested this on my test database (dry run test trades - often with durations of only minutes) - most of them are force-enter / force-exits - and sometimes there's big losers :laughing: 

I struggle to find an explanation for the expectancy values provided.
expectancy-ratio should clearly be negative (this strategy is losing BIG) - but (if i understand this correctly) - it's 4.5 - which i don't really understand.

At any rate (not just for me) - we should have the metrics explained in telegram.md in the /profit section ...

froggleston: If you use the absolute loss value in the following line does it looks right?

`((loserate / 100) * abs(mean_losing_profit))`

froggleston: OK I've realised my calculations are the wrong way around (expectancy vs ratio). Give me a couple of minutes.

froggleston: Sorry, was getting on with real work. This should all be correct now.

freqtrade/rpc/telegram.py
```diff
@@ -873,7 +877,9 @@ async def _profit(self, update: Update, context: CallbackContext) -> None:
                 f"*{'First Trade opened' if not timescale else 'Showing Profit since'}:* "
                 f"`{first_trade_date}`\n"
                 f"*Latest Trade opened:* `{latest_trade_date}`\n"
-                f"*Win / Loss:* `{stats['winning_trades']} / {stats['losing_trades']}`"
+                f"*Win / Loss:* `{stats['winning_trades']} / {stats['losing_trades']}`\n"
+                f"*Winrate:* `{winrate:.2f}%`\n"
```
xmatthias: this could be compresse into one line `Win / Loss: 26 / 178 (12.75%)` to reduce overall linecount.

I'm not entirely sure if that makes sense for the /profit output though - which is not overly long anyway.

freqtrade/rpc/rpc.py
```diff
@@ -521,6 +523,17 @@ def _rpc_trade_statistics(
 
         profit_factor = winning_profit / abs(losing_profit) if losing_profit else float('inf')
 
+        mean_winning_profit = (winning_profit / winning_trades) if winning_trades > 0 else 0
+        mean_losing_profit = (abs(losing_profit) / losing_trades) if losing_trades > 0 else 0
+
+        winrate = (winning_trades / closed_trade_count) * 100 if closed_trade_count > 0 else 0
+        loserate = (100 - winrate)
```
xmatthias: These 2 variables should be as ratio (0-1), not as % (0-100)

winrate is multiplied by 100 - and subsequently both values are `/100` seveal times for subsequent calculations.
For an API consumer, it's also simpler to use it if it's a ratio - showing "52%" is really a formatting question, not how the value should bein the API - and in python, can be Accomplished quite easily with `:.2%}` instead of `:.2f}%`.

froggleston: Done!

xmatthias (approved):
Great enhancement :+1:

PR #8934, opened 2023-07-18
Update docker-compose-freqai.yml

robcaulk:
The `--freqai-model` argument is incorrect. It should be `--freqaimodel`.


xmatthias (approved):
Great spot :+1:

PR #8937, opened 2023-07-19
Add built-in trade fields in indicator list docs

froggleston:
## Summary

There are some trade- and candle-related fields that are always available to output on the indicator-list so have updated the docs to include the most commonly used ones.

## Quick changelog

- updated advanced-backtesting.md to include commonly-used fields available for indicator-list, e.g. profit_abs, min_rate, max_rate, OHLCV etc


xmatthias (approved):
LGTM :+1:

PR #8941, opened 2023-07-21
Map BUSD to correct coingecko id

SitoCH:
## Summary

Using BUSD as stake currency generates a WARNING because Coingecko returns 2 entries for the same symbol.

## Quick changelog

- added to `coingecko_mapping` the mapping `'busd': 'binance-usd'`

### DIFF
--- a/freqtrade/rpc/fiat_convert.py
+++ b/freqtrade/rpc/fiat_convert.py
@@ -25,6 +25,7 @@ coingecko_mapping = {
     'bnb': 'binancecoin',
     'sol': 'solana',
     'usdt': 'tether',
+    'busd': 'binance-usd',
 }
 
 


stash86:
This is interesting, because I have been using BUSD and never have such issue. I even just checked my bots, no such issue

SitoCH:
That's weird, I also manually tested Coingecko's API and this is part of the response:

```
{
  "id": "binance-usd",
  "symbol": "busd",
  "name": "Binance USD"
},
{
  "id": "binance-usd-linea",
  "symbol": "busd",
  "name": "Binance USD (Linea)"
},
```

xmatthias (approved):
That coin only listed 3 days ago on coingecko - so any bot started (or restarted) prior to that wouldn't show this problem.

image

Sad we have to do such workarounds - but the PR is good.

LGTM :+1:

PR #8943, opened 2023-07-21
merge to use  expectancy and expectancy ratio from data/metrics

stash86:
## Summary

merge to use  expectancy and expectancy ratio from data/metrics, instead of having two separate functions

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/data/metrics.py
+++ b/freqtrade/data/metrics.py
@@ -202,7 +202,7 @@ def calculate_expectancy(trades: pd.DataFrame) -> Tuple[float, float]:
     """
 
     expectancy = 0
-    expectancy_ratio = float('inf')
+    expectancy_ratio = 100
 
     if len(trades) > 0:
         winning_trades = trades.loc[trades['profit_abs'] > 0]
--- a/freqtrade/data/metrics.py
+++ b/freqtrade/data/metrics.py
@@ -198,7 +198,7 @@ def calculate_expectancy(trades: pd.DataFrame) -> Tuple[float, float]:
     """
     Calculate expectancy
     :param trades: DataFrame containing trades (requires columns close_date and profit_abs)
-    :return: expectancy
+    :return: expectancy, expectancy_ratio
     """
 
     expectancy = 0
--- a/freqtrade/data/metrics.py
+++ b/freqtrade/data/metrics.py
@@ -201,23 +201,28 @@ def calculate_expectancy(trades: pd.DataFrame) -> float:
     :return: expectancy
     """
 
-    if len(trades) == 0:
-        return 0
+    expectancy = 0
+    expectancy_ratio = float('inf')
 
-    winning_trades = trades.loc[trades['profit_abs'] > 0]
-    losing_trades = trades.loc[trades['profit_abs'] < 0]
-    profit_sum = winning_trades['profit_abs'].sum()
-    loss_sum = abs(losing_trades['profit_abs'].sum())
-    nb_win_trades = len(winning_trades)
-    nb_loss_trades = len(losing_trades)
+    if len(trades) > 0:
+        winning_trades = trades.loc[trades['profit_abs'] > 0]
+        losing_trades = trades.loc[trades['profit_abs'] < 0]
+        profit_sum = winning_trades['profit_abs'].sum()
+        loss_sum = abs(losing_trades['profit_abs'].sum())
+        nb_win_trades = len(winning_trades)
+        nb_loss_trades = len(losing_trades)
 
-    average_win = (profit_sum / nb_win_trades) if nb_win_trades > 0 else 0
-    average_loss = (loss_sum / nb_loss_trades) if nb_loss_trades > 0 else 0
-    winrate = (nb_win_trades / len(trades))
-    loserate = (nb_loss_trades / len(trades))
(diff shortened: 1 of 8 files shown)


stash86:
@xmatthias can you help me figuring out why my test failed at that point? I believe 'profit_abs' column exist, since same thing is used for drawdown calculation below it

freqtrade/optimize/optimize_reports/bt_output.py
```diff
@@ -233,8 +233,8 @@ def text_table_add_metrics(strat_results: Dict) -> str:
             ('Calmar', f"{strat_results['calmar']:.2f}" if 'calmar' in strat_results else 'N/A'),
             ('Profit factor', f'{strat_results["profit_factor"]:.2f}' if 'profit_factor'
                               in strat_results else 'N/A'),
-            ('Expectancy', f"{strat_results['expectancy']:.2f}" if 'expectancy'
-                           in strat_results else 'N/A'),
+            ('Expectancy Ratio', f"{strat_results['expectancy_ratio']:.2f}" if 'expectancy_ratio'
```
xmatthias: is the output now "Expectancy Ratio" instead of  Expectancy - or was this labeled wrongly?

At any rate, we should update the corresponding backtesting.md page, too.
Maybe it also makes sense to output both values - so it's aligned to `/profit` ?

freqtrade/data/metrics.py
```diff
@@ -200,28 +200,54 @@ def calculate_expectancy(trades: pd.DataFrame) -> float:
...
+    :return: expectancy ratio
+    """
+
+    expectancy_ratio = float('inf')
+
+    if len(trades) > 0:
+        winning_trades = trades.loc[trades['profit_abs'] > 0]
+        losing_trades = trades.loc[trades['profit_abs'] < 0]
+        profit_sum = winning_trades['profit_abs'].sum()
+        loss_sum = abs(losing_trades['profit_abs'].sum())
```
xmatthias, line 211: quite a lot of code duplication between this and the function above (not just on these 4 lines)

maybe combining the 2 functions into one, similar to how it was in `rpc.py` makes more sense, as then most of the calculation values can be reused?

freqtrade/rpc/rpc.py
```diff
@@ -523,20 +524,20 @@ def _rpc_trade_statistics(
...
-        loserate = (1 - winrate)
 
-        expectancy, expectancy_ratio = self.__calc_expectancy(mean_winning_profit,
-                                                              mean_losing_profit,
-                                                              winrate,
-                                                              loserate)
+        # expectancy, expectancy_ratio = self.__calc_expectancy(mean_winning_profit,
+        #                                                       mean_losing_profit,
+        #                                                       winrate,
+        #                                                       loserate)
```
xmatthias: we'll def. want to remove the function definition of `__calc_expectancy` itself.
It's now unused as far as i can tell.

freqtrade/data/metrics.py
```diff
@@ -200,28 +200,54 @@ def calculate_expectancy(trades: pd.DataFrame) -> float:
...
+    average_win = (profit_sum / nb_win_trades) if nb_win_trades > 0 else 0
+    average_loss = (loss_sum / nb_loss_trades) if nb_loss_trades > 0 else 0
+    winrate = (nb_win_trades / len(trades))
+    loserate = (nb_loss_trades / len(trades))
+    expectancy = (winrate * average_win) - (loserate * average_loss)
 
     return expectancy
 
 
+def calculate_expectancy_ratio(trades: pd.DataFrame) -> float:
```
xmatthias: Can you help me understand this?

the function that used to be `calculate_expectancy` is now `calculate_expectancy_ratio` - and the old one is now returning the actual expectancy (not the ratio), right?

freqtrade/optimize/optimize_reports/bt_output.py
stash86: Yes, I will make it print both as how it's in /profit

freqtrade/data/metrics.py
stash86: sure

freqtrade/data/metrics.py
```diff
@@ -194,32 +194,35 @@ def calculate_cagr(days_passed: int, starting_balance: float, final_balance: flo
...
     :param trades: DataFrame containing trades (requires columns close_date and profit_abs)
-    :return: expectancy
+    :return: expectancy, expectancy_ratio
     """
-    if len(trades) == 0:
-        return 0
 
-    expectancy = 1
+    expectancy = 0
+    expectancy_ratio = float('inf')
```
xmatthias: i'm ... sceptical about using `inf` here

it's prone to cause `ValueError: Out of range float values are not JSON compliant` - and is the expectancy really infinite if we did no trade, or is it rather 0?

stash86: I'm using frog's default value of inf. Personally, I prefer 100.

xmatthias: Then let's go with that.
backtesting will be ... picky otherwise :laughing:

xmatthias (approved):
Great, thanks! :100: 

LGTM :+1:

PR #8968, opened 2023-07-25
Update showcase.md / Change Domain due to recent dns issues

Bloodhunter4rc:
Change of domain due to recent dns issues.

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Now personally, i'd prefer if the domain were available through https (we can talk about this in discord, maybe) - but that's personal preference mostly

LGTM :+1:

PR #8997, opened 2023-08-01
add second alternatives instead of using .range

stash86:
## Summary

add second alternatives instead of using .range in hyperopt

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Great addition - thanks!👍

PR #9014, opened 2023-08-03
bugfixes and false-positives for lookahead-analysis

hippocritical:
## Summary

This PR targets to fix a few things
1.) remove the checking of force-exited trades by jumping over those
2.) removing false-positives in edge cases
3.) adding a proper error message for if people did not enter a timerange

### What was already fixed:
1.) removal of checking of force-exits

### For 2.) I need some help. This is the configuration that produces those false-biases:
https://drive.google.com/drive/folders/1Apd4RMlfI6qx1AIOPUojQKjjaPvmXoZ2?usp=sharing
Configuratoin with bias:
lookahead-analysis
--timerange=20220101-20220701
--timeframe=5m
--config=user_data/configs_backtest/config_baseline.json
--config=user_data/config_test_lookahead_bias.json
--config=user_data/configs_backtest/config_ccxt200.json
--config=user_data/configs_backtest/config_filetype_jsongz.json
--strategy-list=ElliotV8HO
--targeted-trade-amount=100
(the other configs do not alter the result at all, just to show the full command but they are not necessary)

str(current_time)=="2022-01-23 15:10:00+00:00" and pair=="VELO/USDT" (interface.py, row 1247-ish)

the issue with that kucoin false positive is freqtrade itself, trailing_stoploss should hit at
str(current_time)=="2022-01-23 15:10:00+00:00" and pair=="VELO/USDT" (interface, row 1247-ish)
but it doesn't.

When all pairs (82) are gone though, then it shows bias. Meaning that the initial trade at 15:10 ends there with trailing_stoploss, but the partial (starting at the same datetime but ending at 15:15, 1 candle after the exit would happen) ends with a force-exit.
this is weird, especially when only running VELO/USDT then there is no bias, not changing anything at all.

I dug in deeply again but didn't have a breakthrough. Could you please give some assistance?

### 3.) will be done at the side, currently focussing on 2.)


xmatthias:
As a test, i did the following individual backtests:

```
1640995200-1642967100   :23 19:45
1640995200-1642967400   :23 19:50
1640995200-1642967700   :23 19:55
1640995200-1642968000   :23 20:00
```

3 of the 4 backtests show the folllowing combination: (with the end-date of the LOCG trade changing as it's a force exit).

image

The exact numbers on all other trades do match (including profit and close-date).

The very first (ending at 19:45) migh be considered the most interesting one. 
This one is "missing" the 5m VELO trade - but the open date corresponds exactly to the end date - so this is expected.

---

Maybe as a pointer ... based on my debugging, it might have something to do with this (notice the debug outputs).

image

It's checking for 19:50 - with the trade ending at 19:55 - but i can find no (manual) backtest range (see the ones above i tried) that does a VELO exit at 19:55 - so this is something that's only really reproducible within the analysis execution

---

**edit**
i CAN reproduce this by backtesting with `--pairs VELO/USDT` - which causes the trade to look like this:

image

For now, i have no logical explanation why another trade would influence the EXIT date of this trade.

xmatthias:
Turns out, it's fees. If Velo is the first pair in the pairlist, fees are 0.003 - otherwise it's 0.002 (probably due to another pair being "first" in the list.

kucoin seems to be the only exchange where ccxt provides different default-rates per pair, at least according to my tests across most the supported exchanges.

Backtesting takes the fee from the first pair in the pairlist - which changes as backtesting is reinitialized again and again.
I suspect we could simply take the very first fee - and apply that throughout the test (assigning it to `config['fee']` will suffice).

hippocritical:
If there s anything to be changed yet, please say so.
Thanks for looking into it!

freqtrade/optimize/lookahead_analysis_helpers.py
```diff
@@ -137,6 +137,16 @@ def calculate_config_overrides(config: Config):
                         'just to avoid false positives')
             config['dry_run_wallet'] = min_dry_run_wallet
 
+        if 'fee' not in config or config['fee'] != 0.02:
+            logger.info('fee was not set to a fixed value of 0.02. ')
+            config['fee'] = 0.02
```
xmatthias: wow, a fee of 2% ... that's not only excessive - it'll change results on a TON of strategies ... almost all of which will be negative with that kind of fee (you start with ~-2% on every trade).

While we _can_ default to a specific value - we should accept whatever value a user supplies (currently, my explicit config would be thrown away).
If i use `fee=0.001` in the config because i know my fee level on the exchange will be that - then no command should decide "we use something different" - unless it's the live exchange telling you you're wrong :laughing: 



---

If i'm honest, i'd implement this differently. 
Let the "first" backtest run through - and use the fee value from there (`backtesting.fee`) to assign it to the config going forward, so all subsequent, "individual" backtests then use that first value.

Fees can be quite different depending on markets / exchanges - and forcing one specific value across all will accomplish little - and will potentially break tons of strategies that rely on a specific "low" fee.

hippocritical: whoops
well which value of fees is used literally doesn't matter since we don't want to get how good / bad a trade or strategy is but just to get if this has bias when cutting the timerange
fixating to 0 fees would be proper too and we might catch a little more since fees just offset the profit and maybe some buys/sells do some weird stuff with very very tight trailing etc

xmatthias: that's correct - but 2% is ... excessive, maybe try 0.001 (0.1%) ... that shouldn't impact actual results too much.

I'd still keep a "user desire" though - as you say - as long as it's defined, we don't "really" care what it is.

hippocritical: yes 2% is excessive, that was a brain fart

hippocritical: the amount of fees really does not matter in this case
it just matters that the tests are reproducible and that the fees are static

hippocritical:
tested with the issue-config again and no issues anymore. 
=> no false-positive, so your solution works (obviously)
LGTM

xmatthias (approved):
LGTM :+1:

PR #9030, opened 2023-08-08
Increase bybit ohlcv_candle_limit to 1000

jansmets:
Bybit supports fetching up to 1000 candles in a single request.

bybit API docs:
https://bybit-exchange.github.io/docs/v5/market/kline
 <html>
<body>

limit | false | integer | Limit for data size per page. [1, 1000]. Default: 200
-- | -- | -- | --

</body>
</html>

CCXT : 
https://github.com/ccxt/ccxt/blob/a4d71c478ce7905a82f73bf82dff1349d1bec166/ts/src/bybit.ts#L2348

Following test snippet returns 1000 candles:
```
#!/usr/bin/python
import pandas as pd
import ccxt
from datetime import datetime

exchange = ccxt.bybit()

for timeframe in ['1m']:
    df = pd.DataFrame(
        exchange.fetch_ohlcv(
            symbol='BTC/USDT',
            timeframe=timeframe,
            since=exchange.milliseconds() - (1*365*24*3600*1000),
            limit=1000
        ),
        columns = ['Time', 'Open', 'High', 'Low', 'Close', 'Volume']
    )

    df['Time'] = [datetime.fromtimestamp(float(time)/1000) for time in df['Time']]
    df.set_index('Time', inplace=True)
    print(df)
```

### DIFF
--- a/freqtrade/exchange/bybit.py
+++ b/freqtrade/exchange/bybit.py
@@ -27,7 +27,7 @@ class Bybit(Exchange):
     """
 
     _ft_has: Dict = {
-        "ohlcv_candle_limit": 200,
+        "ohlcv_candle_limit": 1000,
         "ohlcv_has_history": True,
     }
     _ft_has_futures: Dict = {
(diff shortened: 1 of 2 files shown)


jansmets:
I'm going to have a look at the UT.

xmatthias (approved):
seems like spot was updated to use the v5 api, too.
when i tested v5, that wasn't the case yet.

LGTM :+1:

xmatthias:
You might wanna "not" delete that PR though ... it isn't merged yet :fearful: 


**edit** sorry not true - don't really understand why github says YOU closed this - as it's been merged (i can see the merge commit in my local history, as well as on the repository).

Issue #8763, opened 2023-06-12
Reinfocment Learning model failed to run trade on train data with continual_learning


CryptoRahino:
## Describe your environment

  * Operating system: ubuntu 22
  * Python Version: 3.10.0 (`python -V`)
  * CCXT version: ccxt==3.0.97 (`pip freeze | grep ccxt`)
  * Freqtrade Version: 2023.5 (`freqtrade -V` or `docker compose run --rm freqtrade -V` for Freqtrade running in docker)
  
Note: All issues other than enhancement requests will be closed without further comment if the above template is deleted or not filled out.

## Describe the problem:

after running RL training on backtest, trade is failing with error `Observation spaces do not match` 
this happens only when using  continual_learning=true.

 

### Steps to reproduce:

  1. configure continual_learning=true
  2. run backtest on RL
  3. run trade on RL
  

```2023-06-12 05:24:56,332 - freqtrade.freqai.freqai_interface - ERROR - Training ADA/USDT:USDT raised exception ValueError. Message: Observation spaces do not match: Box(-1.0, 1.0, (1, 185), float32) != Box(-1.0, 1.0, (1, 188), float32), skipping.
Traceback (most recent call last):
  File "/opt/freqtrade/freqtrade/freqai/freqai_interface.py", line 242, in _start_scanning
    self.extract_data_and_train_model(
  File "/opt/freqtrade/freqtrade/freqai/freqai_interface.py", line 640, in extract_data_and_train_model
    model = self.train(unfiltered_dataframe, pair, dk)
  File "/opt/freqtrade/freqtrade/freqai/RL/BaseReinforcementLearningModel.py", line 129, in train
    model = self.fit(data_dictionary, dk)
  File "/opt/freqtrade/freqtrade/freqai/prediction_models/ReinforcementLearner.py", line 75, in fit
    model.set_env(self.train_env)
  File "/home/rahino/anaconda3/envs/freqtrade_orig1/lib/python3.10/site-packages/stable_baselines3/common/base_class.py", line 500, in set_env
    check_for_correct_spaces(env, self.observation_space, self.action_space)
  File "/home/rahino/anaconda3/envs/freqtrade_orig1/lib/python3.10/site-packages/stable_baselines3/common/utils.py", line 230, in check_for_correct_spaces
    raise ValueError(f"Observation spaces do not match: {observation_space} != {env.observation_space}")
[...]

PR #9105, opened 2023-08-25
Convert add_state_info warning into a OperationalException

robcaulk:
We have quite a few users who are backtesting with RL + adding `add_state_info: true` to their `rl_config`s. This is not permitted in FreqAI, since state info is only available in dry/live deployments.

Despite the fact that we have documented this in the `rl_config` parameter table of the [docs](https://www.freqtrade.io/en/stable/freqai-parameter-table/#reinforcement-learning-parameters), and we also log a warning to the [user console](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/freqai/RL/BaseEnvironment.py#L84), users are still confused. This is evidenced by multiple issues posted lately:

https://github.com/freqtrade/freqtrade/issues/7982
https://github.com/freqtrade/freqtrade/issues/8763

@xmatthias and I discussed multiple solutions to this problem - including storing some global metadata during backtesting and then enforcing it during dry/live. But that just encourages users to continue avoiding docs/logs. I prefer to raise an `OperationalException` telling users that the parameter they are trying to use is not available to them and that they need to deactivate it if they intend to continue.

### DIFF
--- a/freqtrade/freqai/RL/BaseEnvironment.py
+++ b/freqtrade/freqai/RL/BaseEnvironment.py
@@ -11,6 +11,8 @@ from gymnasium import spaces
 from gymnasium.utils import seeding
 from pandas import DataFrame
 
+from freqtrade.exceptions import OperationalException
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
LGTM :+1:

PR #9135, opened 2023-09-02
Fix nulls in indicator list output and add expectancy ratio

froggleston:
This PR addresses two issues in the `backtesting-analysis` command:

a) If two entry signals were detected on the same signal candle, this resulted in one correct entry, and one null row. This didn't affect any results or calculations, but it would be confusing for a user and is resolved.
b) Add expectancy ratio per entry tag in the analysis group 0 table output.


xmatthias (approved):
LGTM :+1:

PR #9137, opened 2023-09-02
Implment weekly/monthly RPC endpoints

froggleston:
This PR adds the `/weekly` and `/monthly` RPC calls to the main API endpoints and `rest_client.py` script. This brings the API endpoints into line with the available telegram calls.


froggleston:
Nice additions @xmatthias - I wouldn't have seen those.

xmatthias (approved):
not true, you would have once the "monthly" endpoitn was correct  - as then it wasn't working as it was no longer providing a date :laughing:

LGTM :+1:

PR #9152, opened 2023-09-04
Add recursive-analysis sub-command

stash86:
## Summary
Add recursive-analysis sub-command to test indicators of potential recursive issue (value changes when startup candle change)

Still to-do:
1. [x] Tests
2. [x] Documentations
3. [x] Maybe combine it to lookahead-analysis

Solve the issue: #___


freqtrade/optimize/recursive_analysis.py
```diff
@@ -0,0 +1,224 @@
...
+            # remove folder and its contents
+            if Path.exists(path_to_current_identifier):
+                shutil.rmtree(path_to_current_identifier)
+
+        prepare_data_config = deepcopy(self.local_config)
+        prepare_data_config['timerange'] = (str(self.dt_to_timestamp(varholder.from_dt)) + "-" +
+                                            str(self.dt_to_timestamp(varholder.to_dt)))
+        prepare_data_config['exchange']['pair_whitelist'] = pairs_to_load
+
+        backtesting = Backtesting(prepare_data_config, self.exchange)
```
xmatthias, line 122: What exactly (from backtesting) do you need? 
My initial assumption was that it's only analyzing indicators (?) which might not even need a backtesting class (?)

xmatthias (commented):
from a high level view, the whole `RecursiveAnalysis` class seems to share a great deal of code with lookahead-bias (not everything obviously - but there's functions that are  90% identical, if not more).


i think some refactoring will greatly reduce complexity / code duplication (which undoubly will cause deviation over time) - and with that increase maintainability greatly.

`RecursiveAnalysisSubFunctions` might be a similar candidate, too - though it's doesn't seem to share that much code - at least it's not immediately obvious

freqtrade/optimize/recursive_analysis.py
stash86: good question. I just took hippo's code and modify them, so not sure if I can do things without calling backtesting. On quick glance, at least I need backtesting to get the data to calculate indicators

stash86:
Options that I have in mind
1. Combine my code to hippo's code, and add flag to switch between lookahead and recursive
2. Create new class as main class that hold those similar functions

xmatthias:
well clearly option 2
option 1 would/will result in too much "if recursive" - they're somewhat identical - but not THAT similar  :laughing:

freqtrade/optimize/recursive_analysis.py
xmatthias: leave this for now - i'll look at it in detail

in theory, you shouldn't - plotting also doesn't need it (and i don't think you need funding rates or detail timeframe - as you don't really run a backtest afterwards)

stash86:
tests and docs are done. I have created BaseAnalysis to put similar functions there (not many).

froggleston (changes requested):
This looks really good!

My only suggestion would be to include an example output table in the docs to help users understand what the different output values mean.

froggleston (dismissed):

freqtrade/templates/sample_strategy.py
```diff
@@ -77,7 +77,7 @@ class SampleStrategy(IStrategy):
     exit_short_rsi = IntParameter(low=1, high=50, default=30, space='buy', optimize=True, load=True)
 
     # Number of candles the strategy requires before producing valid signals
-    startup_candle_count: int = 30
+    startup_candle_count: int = 170
```
xmatthias: I assume this is the outcome of running the analysis against this strategy?

freqtrade/commands/arguments.py
```diff
@@ -467,3 +470,14 @@ def _build_subcommands(self) -> None:
 
         self._build_args(optionlist=ARGS_LOOKAHEAD_ANALYSIS,
                          parser=lookahead_analayis_cmd)
+
+        # Add recursive_analysis subcommand
+        recursive_analayis_cmd = subparsers.add_parser(
+            'recursive-analysis',
+            help="Check for potential look ahead bias.",
```
xmatthias: I think we might need a better help string.
This is identical to lookahead_analysis - but the command does not do exactly that.

docs/recursive-analysis.md
```diff
@@ -0,0 +1,91 @@
+# Recursive analysis
```
xmatthias, line 1: We'll need to decide what to do with this.
The way it's currently configured, it'll NOT show up in the left-side nav - and will therefore only be reachable either through direct links - or through the link in strategy-customization.

i'm ... tempted to combine this with "lookahead analysis" - as the commands are strongly related, and i think we already have a "too long" nav ...

docs/recursive-analysis.md
```diff
@@ -0,0 +1,92 @@
...
+
+What happens if the calculation is using only the latest 500 candles? Then instead of 999, the `steps` value at last closed candle is 499. The difference of the value means your backtest result can differ from your dry/live run result.
+
+The `recursive-analysis` command requires historic data to be available. To learn how to get data for the pairs and exchange you're interested in,
+head over to the [Data Downloading](data-download.md) section of the documentation.
+
+This command is built upon backtesting since it internally chains backtests to prepare different lengths of data and calculates indicators based on the downloaded data.
+This does not run the strategy itself, but rather uses the indicators it contains. After multiple backtests are done to calculate the indicators of different startup candle values (`startup_candle_count`), the values of last rows across all backtests are compared to see how much variance they show compared to the base backtest.
```
xmatthias: i think we need to work on wording here (thoughout this doc).
while it's using the backtest class currently - it's not actually executing a backtest - but runs the populate_* methods (actually only `populate_indicators()`).

a "backtest" would assume a full run (`backtest.start()`) - which ain't what's actually happening ... 
We might also need a warning someplace that indicators calculated in `populate_entry_trend()` are ignored for this analysis.

freqtrade/templates/sample_strategy.py
stash86: Kinda. It's supposed to be 200 actually. Gonna update it

docs/recursive-analysis.md
stash86: Personally, I felt like Backtesting, Hyperopt, Lookahead and Recursive analysis should all be combined into 1 main menu.

xmatthias: you mean something like this

```
Backtesting
    Backtesting
    hyperopt
    lookahead analysis
    recursive analysis
```

not against this per se - though i don't actually like the behavior of expanding boxes, as it'd require 2 clicks to actually go to backtesting (for example).

stash86: another place I have in mind is `Advanced Topics`

xmatthias: but you'd keep it as 2 separate pages? or do something similar than what we do with plugins ... ?

stash86: keep as separate pages.

freqtrade/optimize/recursive_analysis_helpers.py
```diff
@@ -0,0 +1,106 @@
...
+        print(table)
+        return table, headers, data
+
+    @staticmethod
+    def calculate_config_overrides(config: Config):
+        if 'timerange' not in config:
+            # setting a timerange is enforced here
+            raise OperationalException(
+                "Please set a timerange. "
+                "A timerange of 20 candles are enough for recursive analysis."
```
xmatthias: interesting, this seems to contradict to what we're saying in the docs ... i think we need to adjust either this, or the docs (i think the docs are ... more correct in this regard).

> It is recommended to set a long timerange (at least 5000 candles) so that the initial indicators' calculation that is going to be used as a benchmark has very small or no recursive issues itself. For example, for a 5m timeframe, a timerange of 5000 candles would be equal to 18 days.

freqtrade/optimize/recursive_analysis.py
```diff
@@ -0,0 +1,182 @@
...
+        logger.info(f"Check indicators at {check_time}")
+        # logger.info(f"vs {part_timerange} with {part.startup_candle} startup candle")
+
+        compare_df = base_row_check.compare(part_last_row)
+        if compare_df.shape[0] > 0:
+            # print(compare_df)
+            for col_name, values in compare_df.items():
+                # print(col_name)
+                if 'other' == col_name:
+                    continue
```
xmatthias, line 94: what's the reason to exclude `'other'` from this test ? 

`dataframe['other'] = ta.RSI(dataframe, timeperiod=50).shift(-1)` is ok - but `dataframe['rsi'] = ta.RSI(dataframe, timeperiod=50).shift(-1)` is not? :D

freqtrade/optimize/recursive_analysis_helpers.py
stash86: Ah yes. On it rn

freqtrade/optimize/recursive_analysis.py
stash86: we skip 'other' because `.compare` will create 2 columns  'self' and 'others', and we only need to go through it just once

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.compare.html

xmatthias: ah ok perfect - thanks for the explanation

xmatthias (approved):
We might want to do some deduplication of the 2 helper methods - but i'm intentionally leaving this for a future enhancement.
This command is too valueable to not merge now.

LGTM :+1:

froggleston:
LET'S GOOOOOOOOOOOO! Great addition!

PR #9179, opened 2023-09-12
chore: fix bug associated with leaving FreqAI offline for more than 1candle

robcaulk:
The current reload behavior of FreqAI interferes with user's ability to track historic predictions in FreqUI. This is because we are naively taking the tail of the historic_predictions df and using that to populate the initial FreqUI dataframe (the strategy dataframe), after the reload.

This PR fixes this by ensuring we force 0 values for any missed predictions while the user held the bot offline. So now the prediction would like this after reloading (assuming the below user left the bot offline for 6 candles):

image

### DIFF
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -294,9 +294,9 @@ class FreqaiDataDrawer:
         if len(common_dates.index) > 0:
             new_pred = new_pred.iloc[len(common_dates):]
         else:
-            logger.error("No common dates found between new predictions and historic predictions. "
-                         "You likely left your FreqAI instance offline for more than "
-                         f"{len(dataframe.index)} candles.")
+            logger.warning("No common dates found between new predictions and historic "
+                           "predictions. You likely left your FreqAI instance offline "
+                           f"for more than {len(dataframe.index)} candles.")
 
         df_concat = pd.concat([hist_preds, new_pred], ignore_index=True, keys=hist_preds.keys())
 
--- a/freqtrade/freqai/data_drawer.py
+++ b/freqtrade/freqai/data_drawer.py
@@ -263,23 +263,55 @@ class FreqaiDataDrawer:
             self.pair_dict[metadata["pair"]] = self.empty_pair_dict.copy()
             return
 
-    def set_initial_return_values(self, pair: str, pred_df: DataFrame) -> None:
+    def set_initial_return_values(self, pair: str,
+                                  pred_df: DataFrame,
+                                  dataframe: DataFrame
+                                  ) -> None:
         """
         Set the initial return values to the historical predictions dataframe. This avoids needing
         to repredict on historical candles, and also stores historical predictions despite
         retrainings (so stored predictions are true predictions, not just inferencing on trained
-        data)
+        data).
+
+        We also aim to keep the date from historical predictions so that the FreqUI displays
+        zeros during any downtime (between FreqAI reloads).
         """
 
-        hist_df = self.historic_predictions
-        len_diff = len(hist_df[pair].index) - len(pred_df.index)
-        if len_diff < 0:
-            df_concat = pd.concat([pred_df.iloc[:abs(len_diff)], hist_df[pair]],
-                                  ignore_index=True, keys=hist_df[pair].keys())
+        new_pred = pred_df.copy()
+        # set new_pred values to nans (we want to signal to user that there was nothing
+        # historically made during downtime. The newest pred will get appeneded later in
+        # append_model_predictions)
+        new_pred.iloc[:, :] = np.nan
+        new_pred["date"] = dataframe["date"]
+
+        hist_preds = self.historic_predictions[pair].copy()
+        # rename date_pred column to date so that we can merge on date
+        hist_preds = hist_preds.rename(columns={"date_pred": "date"})
+
+        # find the closest common date between new_pred and historic predictions
+        # and cut off the new_pred dataframe at that date
+        common_dates = pd.merge(new_pred, hist_preds, on="date", how="inner")
+        if len(common_dates.index) > 0:
+            new_pred = new_pred.iloc[len(common_dates):]
         else:
-            df_concat = hist_df[pair].tail(len(pred_df.index)).reset_index(drop=True)
+            logger.error("No common dates found between new predictions and historic predictions. "
+                         "You likely left your FreqAI instance offline for more than "
+                         f"{len(dataframe.index)} candles.")
+
+        df_concat = pd.concat([hist_preds, new_pred], ignore_index=True, keys=hist_preds.keys())
+
+        # remove last row because we will append that later in append_model_predictions()
+        df_concat = df_concat.iloc[:-1]
+        # any missing values will get zeroed out so users can see the exact
+        # downtime in FreqUI
         df_concat = df_concat.fillna(0)
-        self.model_return_values[pair] = df_concat
+
+        # rename date column back to date_pred
+        df_concat = df_concat.rename(columns={"date": "date_pred"})
+
+        self.historic_predictions[pair] = df_concat
+
+        self.model_return_values[pair] = df_concat.tail(len(dataframe.index)).reset_index(drop=True)
 
     def append_model_predictions(self, pair: str, predictions: DataFrame,
                                  do_preds: NDArray[np.int_],
--- a/freqtrade/freqai/freqai_interface.py
+++ b/freqtrade/freqai/freqai_interface.py
@@ -645,11 +645,11 @@ class IFreqaiModel(ABC):
         If the user reuses an identifier on a subsequent instance,
         this function will not be called. In that case, "real" predictions
         will be appended to the loaded set of historic predictions.
-        :param df: DataFrame = the dataframe containing the training feature data
-        :param model: Any = A model which was `fit` using a common library such as
-                      catboost or lightgbm
+        :param pred_df: DataFrame = the dataframe containing the predictions coming
+            out of a model
         :param dk: FreqaiDataKitchen = object containing methods for data analysis
         :param pair: str = current pair
+        :param strat_df: DataFrame = dataframe coming from strategy
         """
 
         self.dd.historic_predictions[pair] = pred_df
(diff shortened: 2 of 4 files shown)


freqtrade/freqai/data_drawer.py
```diff
@@ -263,23 +263,55 @@ def set_pair_dict_info(self, metadata: dict) -> None:
...
+        hist_preds = hist_preds.rename(columns={"date_pred": "date"})
+
+        # find the closest common date between new_pred and historic predictions
+        # and cut off the new_pred dataframe at that date
+        common_dates = pd.merge(new_pred, hist_preds, on="date", how="inner")
+        if len(common_dates.index) > 0:
+            new_pred = new_pred.iloc[len(common_dates):]
         else:
-            df_concat = hist_df[pair].tail(len(pred_df.index)).reset_index(drop=True)
+            logger.error("No common dates found between new predictions and historic predictions. "
```
xmatthias: wouldn't `warning` be more appropriate, considering both wording, as well as what it's supposed to signal?

freqtrade/freqai/data_drawer.py
```diff
@@ -263,23 +263,55 @@ def set_pair_dict_info(self, metadata: dict) -> None:
...
+                         "You likely left your FreqAI instance offline for more than "
+                         f"{len(dataframe.index)} candles.")
```
xmatthias:
```suggestion
                         f"{len(dataframe)} candles.")
```
?

freqtrade/freqai/data_drawer.py
```diff
@@ -263,23 +263,55 @@ def set_pair_dict_info(self, metadata: dict) -> None:
...
-            df_concat = pd.concat([pred_df.iloc[:abs(len_diff)], hist_df[pair]],
-                                  ignore_index=True, keys=hist_df[pair].keys())
+        new_pred = pred_df.copy()
+        # set new_pred values to nans (we want to signal to user that there was nothing
+        # historically made during downtime. The newest pred will get appeneded later in
+        # append_model_predictions)
+        new_pred.iloc[:, :] = np.nan
+        new_pred["date"] = dataframe["date"]
+
+        hist_preds = self.historic_predictions[pair].copy()
```
xmatthias, line 287: In/after this line - what's the difference between `hist_preds` and `pred_df`? 

In my current understanding understanding, `set_initial_historic_predictions()` is assigning that same `pred_df` dataframe to exactly this dict, which effectively means at this point we have 3 copies dataframes of length X,
one empty, and 2 with identical (but copied) contents.

---

(leaving the above aside for now) - is it necessary to do a `.copy()` here?
`self.historic_predictions[pair]` is assigned below  - so i assume the prior dataframe is outdated, so mutating that would not matter, anyway (were not in "multi-threading" land here yet as far as i can tell?)

xmatthias (commented):
Interesting that this whole method isn't covered by tests - but i think with the length it now reached, it will definitely deserve a test or two.
That's even more important considering we're fixing a bug - which usually means we'll want to avoid accidentally reintroducing that at a later point.

Aside from that, it'll also make sure that random dependency updates (pandas, looking at you) won't introduce deprecations or behavior changes we're unaware of.

freqtrade/freqai/data_drawer.py
robcaulk: Why? Explicit is better.

freqtrade/freqai/data_drawer.py
robcaulk: All of FreqAI is multithreaded. We don't want simultaneous access to this while it has a changed column name.

robcaulk: `pred_df` is a single vector of predictions coming out of the model for the current inference only.

`historic_predictions` contains all historic predictions + auxillary info that are appended and sent back to strat. See `append_model_predictions()` function just below.

robcaulk:
> Interesting that this whole method isn't covered by tests - but i think with the length it now reached, it will definitely deserve a test or two. That's even more important considering we're fixing a bug - which usually means we'll want to avoid accidentally reintroducing that at a later point.
> 
> Aside from that, it'll also make sure that random dependency updates (pandas, looking at you) won't introduce deprecations or behavior changes we're unaware of.

I'm not sure how one would cover this in a test, it requires running freqai for multiple candles, then stopping for multiple candles, then restarting.

If you would like to see a test for it, please make a suggestion on how you would like to see the test written.

freqtrade/freqai/data_drawer.py
robcaulk: We can switch this to a `warning`

freqtrade/freqai/data_drawer.py
xmatthias: > All of FreqAI is multithreaded. We don't want simultaneous access to this while it has a changed column name.

This part / function is running in the main thread though (called via `start()` -> `start_live()` - which starts the thread, but then continues to this function) - which is why i'm asking.

I think i got the confusion about `pred_df` now - it's the odd code segment in `set_initial_historic_predictions()` (see link below) - which messes with references in an odd way - and should probably be cleaned up to be clearer.

https://github.com/freqtrade/freqtrade/blob/628963c20704ca7d47f4d2ec79cfaaf048fc8c82/freqtrade/freqai/freqai_interface.py#L655-L656

xmatthias:
Running freqAI and stopping and waiting and starting again would be a full integration test - which is not what i asked for. (you can fake the waiting with time_machine - but i don't think that makes sense here, as the actual system time is irrelevant for this test).
A simple unit-test will suffice, which isolates this method from the rest (it has no dependencies other than 2 external dicts, so is an ideal candidate for an isolated unit test).


The main logic here (and the thing that's most prone to breaking) is the merging - which means you can setup a test which
* has a dataframe that ends at X - assign that to `self.historic_predictions[pair]`
* have a new dataframe that ends at X + 5 (or whatever) ... and has overlapping candles (that's important, the way i understand the logic) - but different actual values (you'll want to distinguish which value came from where)
* call this function directly from the test, and afterwards assert that
  * `historic_predictions[pair]` has no holes, and ends at X+5-1, and has the expected columns (no 'date, but a 'date_pred' column, for example).
  * `model_return_values[pair]` has the same dataframe, but truncated to "dataframe" length

Now you can obviously game step 2 and have have no overlapping candles - which should log the "No common dates" logmessage (which you can also assert for). 
This 2nd test can at the same time test if the remaining behavior (result of  `historic_predictions[pair]`) is what we actually expect.

Hope this is the suggestion you've been asking for.

freqtrade/freqai/data_drawer.py
robcaulk: I can see where the confusion would originate.

So we have the following events in FreqAI which are designed to accommodate FreqUI viewing.

First time a model is created and inferenced, we inference on the full set of exchange candles. We save those as "predictions" (but all except the last one are technically fake predictions). We set those as the first `historic_predictions` dataframe for that pair. We never do this fake prediction thing again after the first call. The goal is simply to fill out FreqUI as a starting point for users.

From then on, we only append to historic_predictions and then return the `tail` of that which is equivalent in length to the strategy dataframe (the exchange candles). 

The current PR is about reloading FreqAI, we need to load in those historic predictions, but we need to identify how many candles we were "offline for" so we can show in FreqUI that there were no predictions duringthose offline candles, and so that future candles are placed properly in FreqUI.

robcaulk:
> Running freqAI and stopping and waiting and starting again would be a full integration test - which is not what i asked for. (you can fake the waiting with time_machine - but i don't think that makes sense here, as the actual system time is irrelevant for this test). A simple unit-test will suffice, which isolates this method from the rest (it has no dependencies other than 2 external dicts, so is an ideal candidate for an isolated unit test).
> 
> The main logic here (and the thing that's most prone to breaking) is the merging - which means you can setup a test which
> 
>     * has a dataframe that ends at X - assign that to `self.historic_predictions[pair]`
> 
>     * have a new dataframe that ends at X + 5 (or whatever) ... and has overlapping candles (that's important, the way i understand the logic) - but different actual values (you'll want to distinguish which value came from where)
> 
>     * call this function directly from the test, and afterwards assert that
>       
>       * `historic_predictions[pair]` has no holes, and ends at X+5-1, and has the expected columns (no 'date, but a 'date_pred' column, for example).
>       * `model_return_values[pair]` has the same dataframe, but truncated to "dataframe" length
> 
> 
> Now you can obviously game step 2 and have have no overlapping candles - which should log the "No common dates" logmessage (which you can also assert for). This 2nd test can at the same time test if the remaining behavior (result of `historic_predictions[pair]`) is what we actually expect.
> 
> Hope this is the suggestion you've been asking for.

Thanks for the suggestion - I have implemented it as you described. Let me know if you were thinking of something else.

freqtrade/freqai/data_drawer.py
xmatthias: well i'm not confused, i'm just pointing out that this is running in the main thread, which contradicts your above statement of "all of freqAI is multithreaded" - as this part is not :)


The code i highlighed above is clearly not clean code, making that function difficult to read for no good reason (it creates 3 aliases for the same object), which i'd like you to fix - but i understand you'd like to look at this at a later point / different PR :+1:

robcaulk: We can chang it now, now problem. But the reason for the alias was actually to reduce line lengths and make the function easier to  read, all while manipulating the original df inside `self.historic_predictions[pair]`. I don’t believe we can assign pred_df to self.historic and then continue using pred_df below with the expectation that it’s mutating self.historic. 

Looking forward to your comment.

xmatthias:
```
 self.dd.historic_predictions[pair] = pred_df 
 hist_preds_df = self.dd.historic_predictions[pair] 
```

what the above does is assign `pred_df` to `self.dd.historic_predictions[pair]` - and then `self.dd.historic_predictions[pair]` to `hist_preds_df`, which creates an alias.
It's however not creating a copy or new reference (at no point in these 2 lines) - it's just storing a reference to pred_df in a dictionary (eventually adding a new key to that dictionary) - and then creating a new alias called `hist_preds_df`.

essentially, the following will work, and show what i mean (the comment is there as i don't think tripple-comparisons are allowed in python)

``` python
assert id(pred_df) == id(hist_preds_df) # == self.dd.historic_predictions[pair]
```

i assume it's intentional to FIRST assign it before mutating (won't this be problematic with multi-threaded accesses, where you could read this dataframe at any point within this function?) - but it's still the same object that's being mutated, so using pred_df throughout the function should be fine, too.

robcaulk: Sounds like you have an idea of how you want this function to look - go ahead and make your changes. 

This function is called once in the entire lifetime of a single `identifier`.

xmatthias (approved):
LGTM :+1:

PR #9182, opened 2023-09-13
Change the cache of rangestability to 1 day

stash86:
## Summary

Change the default cache of RangeStability to 1 day, since there is no point on having the cache at 24 minutes. And this make the cache same as VolatilityFilter, which also using 1d candles and cached at 1 day

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Agreed (can't say it differently :laughing: ).

PR #9187, opened 2023-09-15
Add FullTradesPairlist

### LINKED ISSUE #9175: Reduce Pairlist when max_open_trades is reached
## Describe the enhancement

I would like a strategy option like: limit_pairs_with_max_trades
Which reduces the pairlist to only the active open trades, so if your max_open_trades = 5 and all trade slots are filled, the pairlist should be consisting only of those 5 pairs, which will reduce data usage, cpu usage, and increase the calculation speed for those 5 open trades by a bit. In what file should this be handled (reducing the pairlist)?

Or it could be just a pairlist which checks free_open_trades and returns nothing/open traded pairs if its full, which could be in the last position of the pairlist?

stash86:
## Summary

Add FullTradesPairlist to shrink whitelist when trade slots are full.

Still to do

- [x] Tests
- [x] Freqai-related check/warning
- [x] Make sure the explanation in docs is clear

Solve the issue: #9175 

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/config_examples/config_full.example.json
+++ b/config_examples/config_full.example.json
@@ -70,6 +70,7 @@
     },
     "pairlists": [
         {"method": "StaticPairList"},
+        {"method": "FullTradesFilter"},
         {
             "method": "VolumePairList",
             "number_assets": 20,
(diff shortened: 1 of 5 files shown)


stash86:
@robcaulk if you can help me to add the FullTradesFilter to existing freqai config check

xmatthias:
> @robcaulk if you can help me to add the FullTradesFilter to existing freqai config check

I'm not entirely sure that's necessary ...
it's an length "modifying" pairlist - as are several others. 
We mention this in the docs, but (unless i'm missing something) - we're not actively preventing usage, neither of Volumepairlist, nor of any of the other filters.

stash86:
That's good, because that means all tasks done :D

xmatthias (approved):
Great enhancement :+1:

PR #9188, opened 2023-09-16
Add trade param to custom entry price

Axel-CH:
Add trade parameter to custom_entry_price callback

## What's new?
Allow user to use trade object within the custom_entry_price callback, for example to set an entry price related to number of successful exits.
Note: I had some issues running tests, will fix bug if any, tomorrow

### DIFF
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -520,7 +520,7 @@ class AwesomeStrategy(IStrategy):
 
     # ... populate_* methods
 
-    def custom_entry_price(self, pair: str, current_time: datetime, proposed_rate: float,
+    def custom_entry_price(self, pair: str, trade: 'Trade', current_time: datetime, proposed_rate: float,
                            entry_tag: Optional[str], side: str, **kwargs) -> float:
 
         dataframe, last_updated = self.dp.get_analyzed_dataframe(pair=pair,
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -522,7 +522,7 @@ class AwesomeStrategy(IStrategy):
 
     # ... populate_* methods
 
-    def custom_entry_price(self, pair: str, trade: 'Trade', current_time: datetime, proposed_rate: float,
+    def custom_entry_price(self, pair: str, trade: Optional['Trade'], current_time: datetime, proposed_rate: float,
                            entry_tag: Optional[str], side: str, **kwargs) -> float:
 
         dataframe, last_updated = self.dp.get_analyzed_dataframe(pair=pair,
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -511,7 +511,7 @@ Each of these methods are called right before placing an order on the exchange.
     If your custom pricing function return None or an invalid value, price will fall back to `proposed_rate`, which is based on the regular pricing configuration.
 
 !!! Note
-    Within this function you will have access to the Trade object, giving you the flexibility to adjust your current entry or exit price in relation to the current trade status. This object is available as soon as the first entry order associated with the trade is created.
+    Using custom_entry_price, the Trade object will be available as soon as the first entry order associated with the trade is created, for the first entry, `trade` parameter value will be `None`.
 ### Custom order entry and exit price example
 
 ``` python
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -510,6 +510,8 @@ Each of these methods are called right before placing an order on the exchange.
 !!! Note
     If your custom pricing function return None or an invalid value, price will fall back to `proposed_rate`, which is based on the regular pricing configuration.
 
+!!! Note
+    Within this function you will have access to the Trade object, giving you the flexibility to adjust your current entry or exit price in relation to the current trade status. This object is available as soon as the first entry order associated with the trade is created.
 ### Custom order entry and exit price example
 
 ``` python
--- a/docs/strategy-callbacks.md
+++ b/docs/strategy-callbacks.md
@@ -512,6 +512,7 @@ Each of these methods are called right before placing an order on the exchange.
 
 !!! Note
     Using custom_entry_price, the Trade object will be available as soon as the first entry order associated with the trade is created, for the first entry, `trade` parameter value will be `None`.
+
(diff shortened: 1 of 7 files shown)


freqtrade/freqtradebot.py
```diff
@@ -937,7 +937,7 @@ def get_valid_enter_price_and_stake(
             # Don't call custom_entry_price in order-adjust scenario
             custom_entry_price = strategy_safe_wrapper(self.strategy.custom_entry_price,
                                                        default_retval=enter_limit_requested)(
-                pair=pair, current_time=datetime.now(timezone.utc),
+                pair=pair, trade=trade, current_time=datetime.now(timezone.utc),
```
xmatthias: Trade-object is potentially `None`  here (on the initial entry) - so the type on the callback is clearly wrong.

That's something that will ned to be pretty well documented, too  - and is most likely also the reason you're encountering problems with tests.

Axel-CH:
I will fix the flake8, but I did not found a fix for "LocalTrade" / "Trade" type issue for the moment. Will test more options tomorrow

freqtrade/freqtradebot.py
Axel-CH: Yes I supposed, which type should I set? any?

Axel-CH: from what i see I should put it as Optional

freqtrade/optimize/backtesting.py
```diff
@@ -738,7 +738,7 @@ def get_valid_price_and_stake(
         if order_type == 'limit':
             new_rate = strategy_safe_wrapper(self.strategy.custom_entry_price,
                                              default_retval=propose_rate)(
-                pair=pair, current_time=current_time,
+                pair=pair, trade=trade, current_time=current_time,
```
Axel-CH: @xmatthias I don't get why 'LocalTrade' type cause an isse for 'custom_entry_price', and it is working fine for 'custom_exit_price', the only difference is see there is that it is 'Optional'. What do you think?

xmatthias: LocalTrade != Trade ... and LocalTrade is the parent (this has performance reasons and can't really be changed unfortunately).

it's a typing problem, really - but we don't want to use LocalTrade directly in the strategy (also not as types) - as that has other downsides.
wrap line - add type ignore (other backtesting.py callbacks will have the same).

Axel-CH: I will do that now, Thank you

docs/strategy-callbacks.md
```diff
@@ -510,6 +510,8 @@ Each of these methods are called right before placing an order on the exchange.
 !!! Note
     If your custom pricing function return None or an invalid value, price will fall back to `proposed_rate`, which is based on the regular pricing configuration.
 
+!!! Note
+    Within this function you will have access to the Trade object, giving you the flexibility to adjust your current entry or exit price in relation to the current trade status. This object is available as soon as the first entry order associated with the trade is created.
```
xmatthias: I think wording of this note is quite ... non-specific
This documentation segment covers both entry and exit functions - however this limitation only applies to the entry function (which is not immediately clear from this).

i'd also word this differently / invert  how it's worded - with the assumption that 70-80% of the users will only use one entry - where this is irrelevant.

docs/strategy-callbacks.md
```diff
@@ -520,7 +522,7 @@ class AwesomeStrategy(IStrategy):
 
     # ... populate_* methods
 
-    def custom_entry_price(self, pair: str, current_time: datetime, proposed_rate: float,
+    def custom_entry_price(self, pair: str, trade: 'Trade', current_time: datetime, proposed_rate: float,
```
xmatthias:
```suggestion
    def custom_entry_price(self, pair: str, trade: Optional['Trade'], current_time: datetime, proposed_rate: float,
```

This same applies to everywhere this is defined in the docs ...

freqtrade/strategy/interface.py
```diff
@@ -405,6 +406,7 @@ def custom_entry_price(self, pair: str, current_time: datetime, proposed_rate: f
         When not implemented by a strategy, returns None, orderbook is used to set entry price
 
         :param pair: Pair that's currently analyzed
+        :param trade: trade object.
```
xmatthias:
```suggestion
        :param trade: trade object (None for initial entries).
```

freqtrade/strategy/interface.py
```diff
@@ -395,7 +395,8 @@ def custom_stoploss(self, pair: str, trade: Trade, current_time: datetime, curre
         """
         return self.stoploss
 
-    def custom_entry_price(self, pair: str, current_time: datetime, proposed_rate: float,
+    def custom_entry_price(self, pair: str, trade: Trade,
```
xmatthias:
```suggestion
    def custom_entry_price(self, pair: str, trade: Optional[Trade],
```

Not doing this will make people believe it's always given (also the editor will believe this, leading to mistakes in strategy development).

freqtrade/freqtradebot.py
```diff
@@ -937,7 +937,9 @@ def get_valid_enter_price_and_stake(
             # Don't call custom_entry_price in order-adjust scenario
             custom_entry_price = strategy_safe_wrapper(self.strategy.custom_entry_price,
                                                        default_retval=enter_limit_requested)(
-                pair=pair, current_time=datetime.now(timezone.utc),
+                pair=pair,
+                trade=trade,  # type: ignore[arg-type]
```
xmatthias: There's no reason to ignore this here if typehints are correct (using `trade: Trade` is incorrect - as that implies the value cannot be None, which is incorrect).

docs/strategy_migration.md
```diff
@@ -271,7 +271,7 @@ New string argument `side` - which can be either `"long"` or `"short"`.
 
 ``` python hl_lines="3"
-    def custom_entry_price(self, pair: str, current_time: datetime, proposed_rate: float,
+    def custom_entry_price(self, pair: str, trade: 'Trade', current_time: datetime, proposed_rate: float,
```
xmatthias: you don't want to have this in the "before" section i'd think ...

xmatthias (approved):
Great enhancement :100: 

LGTM :+1:

PR #9191, opened 2023-09-17
Bug: FreqAI fit_live_predictions()

robcaulk:
The latest PR #9179 introduced a bug that only crops up when users engage `fit_live_predictions()` to analyze their historic predictions on the fly. The issue was a creation of duplicate `date_pred` columns, which introduced type checking issues for some users (depends on how the user wants to use `self.historic_predictions` inside `fit_live_predictions()`).

This PR fixes the duplicate column problem.

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -244,6 +244,14 @@ class FreqaiDataKitchen:
                 f"{self.pair}: dropped {len(unfiltered_df) - len(filtered_df)} training points"
                 f" due to NaNs in populated dataset {len(unfiltered_df)}."
             )
+            if len(unfiltered_df) == 0 and not self.live:
+                raise OperationalException(
+                    f"{self.pair}: all training data dropped due to NaNs. "
+                    "You likely did not download enough training data prior "
+                    "to your backtest timerange. Hint:\n"
+                    "https://www.freqtrade.io/en/stable/freqai-running/"
+                    "#downloading-data-to-cover-the-full-backtest-period"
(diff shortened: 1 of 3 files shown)


freqtrade/freqai/data_kitchen.py
```diff
@@ -244,6 +244,14 @@ def filter_features(
                 f"{self.pair}: dropped {len(unfiltered_df) - len(filtered_df)} training points"
                 f" due to NaNs in populated dataset {len(unfiltered_df)}."
             )
+            if len(unfiltered_df) == 0 and not self.live:
+                raise OperationalException(
```
xmatthias, line 248: What was current behavior in backtesting? Didn't it simply log the error and move on?

i wonder if this won't this hinder a longer-term backtest (say, all of 2022) with pairs that only listed in April 2022? (that's a valid backtesting scenario - and one that should be supported, as it's quite common for coins to be listed).

robcaulk: No, if there is no data, the backtest fails with a zero division error. Backtests only work if the user has the data. Therefore, you cannot backtest a coin on all of 2022  if you only have data from April. You need to start your backtest in probably May 2022 (depends on how much training data the user asks for).

xmatthias: Ok, i'll open a seperate issue for this (it's not directly related to this PR).

That's a scenario freqtrade supports, so it should be a scenario freqAI supports as well.

xmatthias (approved):
LGTM 👍

PR #9204, opened 2023-09-19
Update exchanges.md

alexpvpmindustry:
## Summary
changed some text in the docs.

## Quick changelog

- Malaysia is a soverign country... parted with Singapore in 1965. [1]
- updated the list of banned countries for Binance. [2]
- also fix some formating

[1] from wiki: https://en.wikipedia.org/wiki/Singapore
image

[2] from https://www.binance.com/en/legal/list-of-prohibited-countries
image

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -55,7 +55,7 @@ This configuration enables kraken, as well as rate-limiting to avoid bans from t
 ## Binance
 
 !!! Warning "Server location and geo-ip restrictions"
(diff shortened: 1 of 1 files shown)


xmatthias (approved):
I'm pretty sure we copied the list of countries from binance (which would suggest they did have the "Malaysia (Singapour)" part there at the time.

That's besides the point though - it changed, so should we.
LGTM  :+1:

alexpvpmindustry:
All good!

PR #9209, opened 2023-09-21
fix broken link

alexpvpmindustry:
its a broken link in the docs


xmatthias (approved):
Great catch :+1: 

Thanks for taking the time to also fix it :100:

PR #9217, opened 2023-09-24
Update data-download.md

froggleston:
This PR fixes the data-download.md docs to reflect the recent change of default data format for OHLCV and trades data to feather.

### DIFF
--- a/docs/data-download.md
+++ b/docs/data-download.md
@@ -154,13 +154,13 @@ freqtrade download-data --exchange binance --pairs ETH/USDT XRP/USDT BTC/USDT --
 
 Freqtrade currently supports the following data-formats:
 
+* `feather` - a dataformat based on Apache Arrow
 * `json` -  plain "text" json files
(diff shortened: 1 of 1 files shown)


docs/data-download.md
```diff
@@ -203,15 +203,15 @@ time freqtrade list-data --show-timerange --data-format-ohlcv <dataformat>
 
 |  Format | Size | timing |
 |------------|-------------|-------------|
+| `feather` | 72Mb | 3.5s |
 | `json` | 149Mb | 25.6s |
 | `jsongz` | 39Mb | 27s |
 | `hdf5` | 145Mb | 3.9s |
-| `feather` | 72Mb | 3.5s |
 | `parquet` | 83Mb | 3.8s |
```
xmatthias, line 210: i'm not against reordering - prior order was by "date support added" ... but now it seems kinda mixed (maybe by timing would make sense? which would move both parquet and hdf5 in front of json ...)

xmatthias (approved):

PR #9241, opened 2023-09-29
chore: protect users against process_only_new_candles=False

robcaulk:
We often see users attempting to use FreqAI with `process_only_new_candles=False`. This is not supported by FreqAI (and will not be without major rehaul of FreqAI and major improvements to customer hardware).

This PR fixes this and prevents users from wasting time with a strategy that accidentally has process_only_new_candles set to false.


xmatthias (approved):
LGTM :+1:

PR #9261, opened 2023-10-02
Fix visual issues and docs

stash86:
## Summary

Fix visual issues and docs

Solve the issue: #___

## Quick changelog

- No table should be printed when there is no variance
- Fix wordings on log messages
- Fix some lines on docs for backtesting (data format) and recursive analysis (add arguments explanations)

### DIFF
--- a/freqtrade/optimize/analysis/recursive_helpers.py
+++ b/freqtrade/optimize/analysis/recursive_helpers.py
@@ -102,6 +102,5 @@ class RecursiveAnalysisSubFunctions:
             RecursiveAnalysisSubFunctions.text_table_recursive_analysis_instances(
                 RecursiveAnalysis_instances)
         else:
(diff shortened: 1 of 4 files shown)


freqtrade/optimize/analysis/recursive_helpers.py
```diff
@@ -101,6 +103,5 @@ def start(config: Config):
                 RecursiveAnalysis_instances)
         else:
             logger.error("There were no strategies specified neither through "
-                         "--strategy nor through "
-                         "--strategy-list "
+                         "--strategy "
```
xmatthias: You might want to update the first line, too, to be singular.

it's now kinda odd, and reads like there's something missing.

xmatthias (approved):
great! 

LGTM :+1:

PR #9267, opened 2023-10-04
Update stoploss order management

Axel-CH:
The Goal of this PR is to prepare the possibility to have more advanced stoploss management for a Trade

Solve the issue: https://github.com/freqtrade/freqtrade/issues/7499

## Steps

- [x] Remove "stoploss_order_id" occurences in code
- [x] Replace by appropriate function / logic
- [x] Update / Fix unit tests

## What's new?
**stoploss_order_id** will be removed from the codebase, replaced by the usage of **orders** in **trade** object.


freqtrade/freqtradebot.py
```diff
@@ -904,18 +904,18 @@ def execute_entry(
...
-                    trade.stoploss_order_id, trade.pair, trade.amount)
-                self.update_trade_state(trade, trade.stoploss_order_id, co, stoploss_order=True)
-
-                # Reset stoploss order id.
-                trade.stoploss_order_id = None
-            except InvalidOrderException:
-                logger.exception(f"Could not cancel stoploss order {trade.stoploss_order_id} "
-                                 f"for pair {trade.pair}")
+        if trade.has_open_sl_orders:
+            for o in trade.orders:
```
xmatthias: you will want to use a helper here i'd think (similar to the `.open_orders` one, but for stoploss orders).
Otherwise you may end up distributing this same logic (loop all, skip most) throughout the bot,

Axel-CH: Yes, I hesitated but I pushed like that to get your opinion 👍🏽

freqtrade/persistence/trade_model.py
```diff
@@ -364,8 +364,6 @@ class LocalTrade:
     # percentage value of the initial stop loss
     initial_stop_loss_pct: Optional[float] = None
     is_stop_loss_trailing: bool = False
-    # stoploss order id which is on exchange
-    stoploss_order_id: Optional[str] = None
     # last update time of the stoploss order on exchange
     stoploss_last_update: Optional[datetime] = None
```
xmatthias: @Axel-CH Now a very good question is - what do we do with `stoploss_last_update`?

in it's current state, it refers to `stoploss_order_id`. 
If we assume we can "X" stop orders, which one would it refer to? the oldest? newest?  don't think that's going to work (but removing this should be simple/easy, as it's really only used in 2 places (at least from an operational standpoint).

can an attribute in the order object replace this (most likely `order.order_date` ).

Axel-CH: I think using the most recent order_date of from all stoploss orders related to that trade is a good option

freqtrade/freqtradebot.py
```diff
@@ -1325,6 +1299,41 @@ def handle_trailing_stoploss_on_exchange(self, trade: Trade, order: Dict) -> Non
                     logger.warning(f"Could not create trailing stoploss order "
                                    f"for pair {trade.pair}.")
 
+    def manage_trade_stoploss_orders(self, trade, stoploss_orders):
```
xmatthias: This will need a typehint - but more importantly - a decision on the type of stoploss_orders.

the method assumes it's a `Order` object ... but the calling method passes in a list of exchange order responses (which don't have a `.status`).

Axel-CH: Yes, I see, what type you think we should manage here? exchange or freqtrade orders?

xmatthias: i'm ... not entirely sure about the whole `manage_trade_stoploss_orders` (as in - why it was extracted, why it needs different logic outside of the original loop (though i can imagine this somewhat).

it's probably going to be the exchange order dict - as the actual order obj is currently not available - and we really don't use it for much here.

freqtrade/freqtradebot.py
```diff
@@ -1219,7 +1219,7 @@ def handle_stoploss_on_exchange(self, trade: Trade) -> bool:
         logger.debug('Handling stoploss on exchange %s ...', trade)
 
         stoploss_orders = []
-        for slo in trade.sl_orders:
```
Axel-CH: Nice catch!

Axel-CH (commented):
Thank you for fixing and updating the tests, great enhancements. If you have any specific tests cases or remark, please let me know, I will work on it tomorrow

tests/test_freqtradebot.py
```diff
@@ -1811,23 +1823,37 @@ def test_handle_stoploss_on_exchange_trailing_error(
...
     mocker.patch(f'{EXMS}.fetch_stoploss_order',
                  return_value=stoploss_order_hanging)
+    time_machine.shift(timedelta(minutes=50))
     freqtrade.handle_trailing_stoploss_on_exchange(trade, stoploss_order_hanging)
     assert log_has_re(r"Could not cancel stoploss order abcd for pair ETH/USDT.*", caplog)
 
     # Still try to create order
     assert stoploss.call_count == 1
+    # TODO: Is this actually correct ? This will create a new order every time,
+    assert len(trade.open_sl_orders) == 2
```
xmatthias, line 1848: @Axel-CH the thing i don't "really" like is this.

Essentially, the cancellation failed - yet we simply create a new order, resulting in 2 open stop orders.
It's not a new behavior either i think - but i'm not entirely sure what the correct approach is (why did cancelling fail? most likely, placing a new order will fail, too - as the amount was sold ...)

another, detailed look at the test (and the corresponding code, obviously) will definitely help.

Axel-CH: I think what is missing is the fact that "cancel_stoploss_on_exchange" was not updated . Imo it should cancel all stoploss on exchange orders. Could be replaced by a "cancel_all_stoploss_on_exchange" that will loop over all existing sl and maybe in the future by "cancel_last_stoploss_on_exchange" that will cancel only the last updated stoploss order in order to keep others sl orders.
Also, "cancel_stoploss_on_exchange" is still using and updating "trade.stoploss_order_id" wich should be replaced.

Axel-CH: If the cancellation failed, it could be because the order do not exist, in that case it is good to create a new one.
It could also be because the order was already filled, in that case we should check if it is a partial fill, and if it is, create a new sl order for remaining quantity.

xmatthias: well that's my point, yes
It's however not changed behavior  so i'm pretty uncertain if it's a good idea to look at this as part of this PR.
The only thing changing now is that we "know" that there's now 2 open SL orders, and won't "forget" the prior order ...

Axel-CH: Yes of course, I agree, as this PR is the first step to better stoploss handling it is not required to manage that edge case here

xmatthias: yeah, so what's left is essentially "more tests" ...

(also, i'll not merge this before the next release, to get more time in testing this on a wider audience)

Axel-CH: Ok, by more tests you mean live or unit tests?

xmatthias: live ... essentially testing the ACTUAL behavior, with the randomness of some calls just failing or whatever ... :laughing: 

Unit tests do work fine - but will only bring us that far :laughing:

Axel-CH: I will do live testing today on bybit

Axel-CH: Tested today stop-loss on exchange (bybit future):
- [x] Initial SL
- [x] Replace with custom stoploss
- [x] Remove on exchange by user / replaced by bot

I was not able to trigger an emergency exit, but i'm getting the same behavior on develop. No issue found on my side so far

Axel-CH: I also tested stop-loss managed in freqtrade and it worked fine, LGTM 👍

xmatthias (approved):
I did a few tests myself with it - and behavior looks identical than it was before.

LGTM :+1:

Issue #9273, opened 2023-10-07
Bybit: Sucessfully cancelled entry order (unfilled timeout) does not remove trade due to incorrect response from exchange


stevanStevic:
## Describe your environment

  * Operating system: Ubuntu 22.04
  * Python Version: 3.10.11
  * CCXT version: 4.0.88
  * Freqtrade Version: freqtrade 2023.10-dev-178c0972f
  
Note: All issues other than enhancement requests will be closed without further comment if the above template is deleted or not filled out.

## Describe the problem:

When I run freqtrade in live mode on Bybit exchange with very small unfilled timeout (e.g. 5 or 10s) and large stake amount, sometimes order gets cancelled due to timeout. When cancelling order [Bybit returns only id](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/exchange/exchange.py#L1346) of the order which is being cancelled. Therefore we [have to fetch the order specifically](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/exchange/exchange.py#L1352) to determine it's latest status. However, Bybit doesn't update status on time, and returns `status == open` even though it is cancelled on exchange. Every next call to fetch that order id returns proper status, which is cancelled. This causes freqtrade not to delete the trade because of [this condition](https://github.com/freqtrade/freqtrade/blob/develop/freqtrade/freqtradebot.py#L1541) and creates stop loss order for position which is not filled.

### Steps to reproduce:

  1. Configure live trading on Bybit with isolated futures with large stake amount on coin with lower volume.
  2. Use unfilled timeout configuration as follows:
```json
     "unfilledtimeout": {
        "entry": 5,
        "exit": 10,
        "exit_timeout_count": 0,
        "unit": "seconds"
    }
```
  4. Check the rest of the configuration in the attached log and configure as such.
  5. Wait for signal and order to be created and cancelled (sometimes it might as well be filled).

  
### Observed Results:

  * What happened?
 
After closing the entry order Bybit returns `'status': 'open'`. 
```
2023-10-07 19:55:11,899 - urllib3.connectionpool - DEBUG - https://api.bybit.com:443 "POST /contract/v3/private/order/cancel HTTP/1.1" 200 141
[...]

PR #9274, opened 2023-10-07
Bybit: Fetch order one more time to allow exchange to set correct value

stevanStevic:
## Summary

Solve the issue: #9273

## Quick changelog

- Re-fetch the cancelled order in order to give enough time to bybit to update order status correctly to canceled.

### DIFF
--- a/freqtrade/exchange/bybit.py
+++ b/freqtrade/exchange/bybit.py
@@ -7,7 +7,8 @@ import ccxt
 
 from freqtrade.constants import BuySell
 from freqtrade.enums import CandleType, MarginMode, PriceType, TradingMode
-from freqtrade.exceptions import DDosProtection, OperationalException, TemporaryError
+from freqtrade.exceptions import (DDosProtection, InvalidOrderException, OperationalException,
+                                  TemporaryError)
 from freqtrade.exchange import Exchange
 from freqtrade.exchange.common import retrier
 from freqtrade.util.datetime_helpers import dt_now, dt_ts
@@ -232,4 +233,26 @@ class Bybit(Exchange):
         ):
             # Canceled orders will have "remaining=0" on bybit.
             order['remaining'] = None
+        return order
+
+    def cancel_order_with_result(self, order_id: str, pair: str, amount: float) -> Dict:
+        order = super().cancel_order_with_result(order_id, pair, amount)
+
+        # Currently bybit does not update status of the cancelled order on time.
+        # Even though endpoint returns proper response and code 200, the status stays open.
+        # This re-fetches the order giving vyvit enough time to update the status.
+        if order.get("status") == "open" and order.get("filled") == 0.0:
+            try:
+                order = self.fetch_order(order_id, pair)
+            except InvalidOrderException:
+                logger.warning(f"Could not fetch cancelled order {order_id}.")
+                order = {
+                    "id": order_id,
+                    "status": "canceled",
+                    "amount": amount,
+                    "filled": 0.0,
+                    "fee": {},
+                    "info": {},
+                }
+
         return order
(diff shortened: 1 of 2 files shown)


freqtrade/exchange/bybit.py
```diff
@@ -233,3 +234,25 @@ def fetch_order(self, order_id: str, pair: str, params: Dict = {}) -> Dict:
             # Canceled orders will have "remaining=0" on bybit.
             order['remaining'] = None
         return order
+
+    def cancel_order_with_result(self, order_id: str, pair: str, amount: float) -> Dict:
+        order = super().cancel_order_with_result(order_id, pair, amount)
+
+        # Currently bybit does not update status of the cancelled order on time.
+        # Even though endpoint returns proper response and code 200, the status stays open.
+        # This re-fetches the order giving vyvit enough time to update the status.
```
stevanStevic:
```suggestion
        # This re-fetches the order giving bybit enough time to update the status.
```

freqtrade/exchange/bybit.py
```diff
@@ -233,3 +234,25 @@ def fetch_order(self, order_id: str, pair: str, params: Dict = {}) -> Dict:
             # Canceled orders will have "remaining=0" on bybit.
             order['remaining'] = None
         return order
+
+    def cancel_order_with_result(self, order_id: str, pair: str, amount: float) -> Dict:
+        order = super().cancel_order_with_result(order_id, pair, amount)
+
+        # Currently bybit does not update status of the cancelled order on time.
+        # Even though endpoint returns proper response and code 200, the status stays open.
+        # This re-fetches the order giving bybit enough time to update the status.
```
xmatthias, line 243: I'm not entirely sure this is true.

i've seen delays in order status updates of up to 8 minutes (very high volume day) on bybit (though on the old API version) - so this will not actually give you any guarantee that the status effectively changes.

freqtrade/exchange/bybit.py
```diff
@@ -233,3 +234,25 @@ def fetch_order(self, order_id: str, pair: str, params: Dict = {}) -> Dict:
...
+        if order.get("status") == "open" and order.get("filled") == 0.0:
+            try:
+                order = self.fetch_order(order_id, pair)
+            except InvalidOrderException:
+                logger.warning(f"Could not fetch cancelled order {order_id}.")
+                order = {
```
xmatthias, line 249: I don't think you'll need this here.

I understand it's a copy/paste from the upstream method - but it really isn't necessary (it's a workaround for exchanges that do a "/delete" on canceled orders, and don't remember them seconds after they're canceled).

if the order disappears (which would mean bybit forgets about the order - which bybit doesn't do, at least for several days) - then you'll get a "canceled" response from the parent method already - never actually hitting this branch.

xmatthias (commented):
Considering that this seems to be a more prevalent problem with order cancels (see linked code below) - maybe we should fix this differently ? 

https://github.com/freqtrade/freqtrade/blob/094984eb2f03610f282698dad62c4c0b0f5b0efc/freqtrade/freqtradebot.py#L1527-L1535


Potentially, the above code could be moved to the exchange class - with a setting to determine if (and how often) we retry for each exchange (this is not necessary for all exchanges)  (with a minimum of 1 retry for every exchange UNLESS it raises the InvalidOrder exception).


**edit** Further looking at the linked issue, this isn't an actual fix. 
It's a workaround that accidentally works at the moment (low volume markets) - but won't reliably fix this - as it's based around "the exchange updates this state within 1 call without delay" - which is not really reliable.

stevanStevic:
Hey @xmatthias 

> It's a workaround that accidentally works at the moment (low volume markets) - but won't reliably fix this - as it's based around "the exchange updates this state within 1 call without delay" - which is not really reliable.

and

> You can't assume that cancels happen immediately - that assumption may hold for your tests, but isn't applicable reliably (not for bybit, and not for other exchanges).
Also because you repeat - but in high-volume times, that repeat will not be sufficient.

With this being said, I agree this wouldn't be a proper way to fix it, and it could happen on other exchanges as well. 

I just tested with your fix https://github.com/freqtrade/freqtrade/commit/897c14e2e5f3725337747a9e093e0c8c2039677d, and this solves the issue, more systematically. Thanks for fixing. 

I will close this PR.

PR #9290, opened 2023-10-09
Fix: Makes sure fetched `trades` exists before using it

TheJoeSchr:
## Summary

Simple fix for when `trades` could not be downloaded by `fetchTrades`, debug should check if `trades` exist before accessing it via array index selector

### DIFF
--- a/freqtrade/exchange/exchange.py
+++ b/freqtrade/exchange/exchange.py
@@ -2223,6 +2223,9 @@ class Exchange:
                 )
                 trades = await self._api_async.fetch_trades(pair, since=since, limit=1000)
             trades = self._trades_contracts_to_amount(trades)
+
+            if len(trades) > 0:
+                logger.debug( "Fetched trades for pair %s, datetime: %s (%d).", pair, trades[0]['datetime'],  trades[0]['timestamp']  )
             return trades_dict_to_list(trades)
         except ccxt.NotSupported as e:
             raise OperationalException(


freqtrade/exchange/exchange.py
```diff
@@ -2223,6 +2223,9 @@ async def _async_fetch_trades(self, pair: str,
                 )
                 trades = await self._api_async.fetch_trades(pair, since=since, limit=1000)
             trades = self._trades_contracts_to_amount(trades)
+
+            if len(trades) > 0:
+                logger.debug( "Fetched trades for pair %s, datetime: %s (%d).", pair, trades[0]['datetime'],  trades[0]['timestamp']  )
```
xmatthias, line 2228: not sure what this is fixing ... ? 
it's just adding a conditional log statement - but won't change the return value or whatever ... ?

TheJoeSchr: It's fixing a crash via exception when `trades` are empty but `logger.debug` access them via `trades[0]`, you see?

My thoughts behind it are that a `debug` log shouldn't crash a program

xmatthias: That's correct - but considering the diff - both the if, and the log message are new (so there's currently no debug output - or at least not in this area of the code).

therefore I'm not sure what this intends to fix, as it's only adding lines, but not replacing anything.

TheJoeSchr: > but considering the diff - both the if, and the log message are new (so there's currently no debug output -

oh, that's what you mean. I'm sorry, I must have written this `logger.debug` statement a few months ago myself. I thought it was original freqtrade code.

PR #9304, opened 2023-10-15
have to use bitwise operator on series comparison, otherwise ambiguous error is thrown

stash86:
## Summary

have to use bitwise operator on series comparison, otherwise ambiguous error is thrown

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Nice catch :+1:

PR #9305, opened 2023-10-15
optimize dtypes for hyperopt and backtesting to decrease memory usage

AliSayyah:
## Summary

optimizes the dtypes of historical data to decrease ram usage.

## Quick changelog

- updated `pandas` to version 2.1.1 to prevent loss of meaningful decimals.
- added a function to downcast dataframes using `pd.to_numeric`.
- added `Bottleneck` to requirements. pandas recommend this library for `Accelerates certain types of nan by using specialized cython routines to achieve large speedup.`

## What's new?
decreases RAM usage especially if lots of indicators are used. pandas defaults to `Float64` but most columns can be downcasted to `Float32`.

## Problems?
two failing tests could be looked into deeper to understand the cause of failure. I could use some help to understand the cause of this behavior.  haven't seen anything else.


freqtrade/data/history/history_utils.py
```diff
@@ -145,12 +171,12 @@ def refresh_data(*, datadir: Path,
...
-    timeframe: str,
-    timerange: Optional[TimeRange],
-    data_handler: IDataHandler,
-    candle_type: CandleType,
-    prepend: bool = False,
+        pair: str,
+        timeframe: str,
+        timerange: Optional[TimeRange],
+        data_handler: IDataHandler,
+        candle_type: CandleType,
```
xmatthias: Before looking at this any further, please remove all non-relevant changes from this PR.

This is not limited to this section / file - please do a review yourself and identify all sections where this comment can apply, reverting these changes accordingly.

requirements.txt
```diff
@@ -1,11 +1,13 @@
 numpy==1.26.0; platform_machine != 'armv7l'
 numpy==1.25.2; platform_machine == 'armv7l'
-pandas==2.0.3
+pandas==2.1.1
```
xmatthias: Pandas was reverted intentionally, and the version 2.1.1 is not one we will be able to use (it contains a bug that causes tons of (false) deprecation warnings.

We'll update pandas when it's ready - but surely not as part of this PR.

requirements.txt
```diff
@@ -1,11 +1,13 @@
...
 pandas-ta==0.3.14b
 
 ccxt==4.1.8
 cryptography==41.0.4
 aiohttp==3.8.6
+# for increasing pandas performance
+Bottleneck==1.3.7
```
xmatthias: I don't see this used.

We'll not force people to  install random, unused dependencies.

**edit** just seen this:
>  pandas recommend this library for 

Please provide a source for this statement.

xmatthias (changes requested):
See other comments. 
will have a proper look once these are adressed.

requirements.txt
AliSayyah: https://pandas.pydata.org/docs/getting_started/install.html#performance-dependencies-recommended

requirements.txt
AliSayyah: yes, of course, this PR won't work as intended before that.

freqtrade/data/history/history_utils.py
AliSayyah: reverted these changes.

requirements.txt
xmatthias: I'm not entirely sure what would've changed in a 2.1.1 that influences this PR? 

Aside from that - Then we'll have to postpone looking at this further, at least until 2.1.2 is released.

The problematic bug:
https://github.com/pandas-dev/pandas/issues/55025
a comment near the end promises this to be included in 2.1.2.

The problem is visible if you run pytest (or a strategy) - causing tons of deprecation warnings due to string assignments - which are however creating the column - causing a wrong warning.

freqtrade/data/history/history_utils.py
```diff
@@ -116,6 +120,28 @@ def load_data(datadir: Path,
     return result
 
 
+def optimize_dtypes(data: DataFrame) -> None:
```
xmatthias: While taking a different approach - i don't really see this differ too much from `reduce_dataframe_footprint` (in `converter.py`)? 

Maybe that function can be improved instead of introducing a new function doing something VERY similar somewhere unfitting (this function shouldn't be in history_utils for sure).

requirements.txt
xmatthias: and that's self-importing, or imported by pandas ... ? 

looking at the bottleneck documentation only shows explicit samples, but doesn't seem to mention that it's auto-used by pandas ?

freqtrade/data/history/history_utils.py
AliSayyah: Haven't seen that function before. yes, the idea is the same. improving that function would work too.

requirements.txt
AliSayyah: prior to `2.1.1`, using `pd.to_numeric` in some scenarios has a chance of losing meaningful decimals. This has been addressed in `2.1.1`.

freqtrade/strategy/interface.py
```diff
@@ -1379,6 +1383,8 @@ def advise_entry(self, dataframe: DataFrame, metadata: dict) -> DataFrame:
         logger.debug(f"Populating enter signals for pair {metadata.get('pair')}.")
 
         df = self.populate_entry_trend(dataframe, metadata)
+        if self.config.get('runmode') not in [RunMode.DRY_RUN, RunMode.LIVE]:
+            optimize_dtypes(dataframe)
```
xmatthias: I think this will have an adverse effect than what you intend.

I can imagine calling this after `populate_indicators()` - but if you call it after "entry_trend" AND `exit_trend` - you'll call this twice for every hyperopt iteration (within each epoch), after an initial call before the hyperopt epochs (that's the one from `populate_indicators` - which i can see making sense).


As each iteration does operations on every column (therefore, for indicator columns, performing it a total of 3 times without actually changing anything) - it'll have an adverse affect, causing a lot more work than what would actually be necessary, thereby negating the slight benefit a different type MAY have.

requirements.txt
xmatthias: unfortunate :laughing: 

all 2.1.x have that string column problem - and each "deprecation" warning they issue causes a slight delay.
Maybe alternative methods (other than `.to_numeric()`) can have a similar effect? for example `.as_type()` ?

freqtrade/data/history/history_utils.py
xmatthias: > Haven't seen that function before.

imagined :) 
that's why we do reviews ;)

requirements.txt
AliSayyah: as per pandas 2.1.1 documentation: 
```
compute.use_bottleneck : bool

Use the bottleneck library to accelerate if it is installed, the default is True

 Valid values: False,True [default: True] [currently: True]
 ```
[reference](https://pandas.pydata.org/docs/reference/api/pandas.set_option.html#pandas.set_option)

File: freqtrade/strategy/interface.py (hunk 5 above)
Comment (AliSayyah) [reply freqtrade/strategy/interface.py]: so to be clear, use are suggesting removing the function call from `advise_entry` and `advise_exit`, right?
and yes, the main gain is from optimizing the indicator columns. I have seen a slight RAM decrease after adding it to exit and entry methods but nothing major.

File: requirements.txt (hunk 2 above)
Comment (AliSayyah) [reply requirements.txt]: well, `as_type()` could be used to do the same thing but downcasting needs a bunch of validation to prevent precision loss. I think relying on `pd.to_numeric()` would be better in this case because it takes some considerations into account.

File: freqtrade/strategy/interface.py (hunk 5 above)
Comment (xmatthias) [reply freqtrade/strategy/interface.py]: yes (unless you have compelling tests to show otherwise).

Doing such optimization is always a tradeoff - you're trading higher RAM for longer CPU time (more operations to reduce ram).

up to some point, the tradeoff may be in favor of "do more CPU" (so do the conversation) - but in this case, i think your overall gain (more speed for hyperopt) - the RAM decrease will probably not be worth it, as you'll trade in "more CPU" once every epoch - which is 100 or 1000'ds of times (depending on -e in hyperopt).

Comment (AliSayyah) [reply freqtrade/strategy/interface.py]: Is there any benchmarking guideline for freqtrade so we can decide on this tradeoff better?
if there isn't, how can I showcase credible results for these gains I'm mentioning? What was your previous approach to this kind of PRs?

**Comment by AliSayyah** [comment]
hyperopt logs for SampleStrategy with one pair. could give a good insight about gains.
```
2023-10-15 16:42:11,744 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 9.63 MB
2023-10-15 16:42:11,750 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 6.42 MB
2023-10-15 16:42:11,750 - freqtrade.optimize.backtesting - INFO - Loading data from 2021-08-31 07:20:00 up to 2023-09-01 00:00:00 (730 days).
2023-10-15 16:42:11,750 - freqtrade.optimize.hyperopt - INFO - Dataload complete. Calculating indicators
2023-10-15 16:42:11,901 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 33.72 MB
2023-10-15 16:42:11,941 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 20.07 MB

```

**Comment by AliSayyah** [comment]
my custom strategy with lots of indicators and 60 pair: (these are for indicator calculation step)
```
2023-10-15 18:12:39,643 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 154.48 MB
2023-10-15 18:12:39,794 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 120.13 MB
2023-10-15 18:12:51,101 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 149.66 MB
2023-10-15 18:12:51,239 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 86.38 MB
2023-10-15 18:13:02,388 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 149.66 MB
2023-10-15 18:13:02,520 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 86.38 MB
2023-10-15 18:13:13,591 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 149.66 MB
2023-10-15 18:13:13,725 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 86.38 MB
2023-10-15 18:13:24,864 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 149.66 MB
2023-10-15 18:13:25,004 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 86.38 MB
2023-10-15 18:13:36,377 - freqtrade.data.history.history_utils - INFO - Memory usage of dataframe is 149.66 MB
2023-10-15 18:13:36,517 - freqtrade.data.history.history_utils - INFO - Memory usage after optimization is: 85.58 MB
...
```
huge memory gains can be seen.

File: freqtrade/data/converter/converter.py [hunk 6]
```diff
@@ -253,27 +253,31 @@ def convert_ohlcv_format(
...
+        if issubclass(df[col].dtypes.type, numbers.Integral):
+            # unsigned integers
+            if df[col].min() >= 0:
+                df[col] = pd.to_numeric(df[col], downcast="unsigned")
+            # signed integers
+            else:
+                df[col] = pd.to_numeric(df[col], downcast="integer")
+        # other real numbers
+        elif issubclass(df[col].dtypes.type, numbers.Real):
+            df[col] = pd.to_numeric(df[col], downcast="float")
```
xmatthias: is the float behavior still the same (float == float32) ? 

now for obvious reasons, we'll need to performance-test (and compat-test) this pretty well - as it's going to have quite some impact on freqAI.

xmatthias:
> my custom strategy with lots of indicators and 60 pair: (these are for indicator calculation step)

on how many candles? unfortunately that's not visible in the logs - so 730 days can be 730 candles, or 1m candles - where it'll be  north of 730_000 candles.
it's also not immediately clear what each log-entry is for.

In reality, we'll want to benchmark 3 things i think to have something comparable
* With this enhancement "as is" - in 3 places in the strategy (eventually also with intermediate results - it's not immediately clear what the logs above show
* without the calls in populate_entry and exit (which is how i'd apply this)

for each, we'd also want the timing (how long did it take to reduce the size once or 3 times).

I'd ignore hyperopt directly - we can interpolate hyperopt from backtesting results - as we know that it'll simply execute the 2nd and 3rd step (`populate_(entry, exit)_trend()`) over and over again.

freqtrade/data/converter/converter.py
```diff
@@ -253,27 +253,31 @@ def convert_ohlcv_format(
...
 
     logger.debug(f"Memory usage after optimization is: "
                  f"{df.memory_usage().sum() / 1024**2:.2f} MB")
-
-    return df
```
xmatthias, line 301: any particular reason to change the interface this way?
i think being explicit (this is being changed and is therefore reassigned) would be better from an interface perspective (even though i'm fully aware that it'll actually mutate the original dataframe).

requirements.txt
xmatthias: hi @AliSayyah 
pandas finally released the 2.1.2 version, which is what we now depend on.

with that, i think we can start looking at this actively again.

AliSayyah: Hi @xmatthias 
Thank you for reminding me of the release. yes, I agree. How should we proceed?

xmatthias: probably
fix conflict
fix ci

then i'll have a spin to see if this works well for all intended usecases, or needs some change / adjustment? 

(this last point will depend on time - i won't have time to do some tests before the weekend ... which essentially (also) defines the sequence) :laughing:

AliSayyah: Hey @xmatthias, got caught up with mid-terms but I've sorted out those conflicts. Can you please give the CI workflow a thumbs up?

xmatthias (commented):
i'm not a huge fan of how this is done (mostly, the change in "history_utils).

Debugging failing test shows the reason:

image

The "first" `df.head()` is 

```
    # Additions at the top of the page / top of the function to fix output
    pd.set_option('display.precision', 15)
    pd.set_option('display.max_columns', 1000)
    pd.set_option('display.expand_frame_repr', False)
```

the open/high/low/close values change.
The reason is probably clear - as it's a rounding issue - but it highlights the reason (and importance) to exclude ohlcv columns.

While this is a small absolute change, it's no longer corresponding to the original exchange candles - without the ability for the user to opt-out of this.

i think we should remove  the call in this location (allow loading of the data "as is").
in all other cases, the function should be called with `skip_original` - to not modify the exchange data.

freqtrade/data/history/history_utils.py
```diff
@@ -103,7 +103,9 @@ def load_data(datadir: Path,
                                  data_handler=data_handler,
                                  candle_type=candle_type,
                                  )
+
         if not hist.empty:
+            hist = reduce_dataframe_footprint(hist)
```
xmatthias, line 118: i'm not a huge fan of putting this here (mostly, because it's non optional, but also because it changes values in a wrong way - see other comment for details).

xmatthias:
After having taken this for another spin - i've decided that we'll not be accepting this.

in my tests, performance of the new approach is worse (both in time and "reduced size" metrics).

in my test, which is essentially a copy of the test-case, but with 600k+ rows:

from `56.31 MB` to `43.51 MB` in ~ 27ms using the current method
from `56.31 MB` to `46.07 MB` in ~49ms using the new method.

Now reducing the original columns (ohlcv) is not an option, and not something we'd consider - so i've intentionally removed that part for the test.

I'll salvage one thing - which is the automatic reduction in size for backtesting - which i'll be adding separately (it's not worth reverting everything else just to accept 3 lines). 
this did happen in 2d194995bc8f7d7a8e15462f26aa3076b84f1b8a.

PR #9339, opened 2023-10-23
fix: bug in startup candle offset

robcaulk:
After we improved the FreqUI viewing for FreqAI restarts #9179 , we introduced a bug with the candle offset in the startup. 

Thanks to @smidelis for finding this bug and doing all the the leg work for resolution.


xmatthias (approved):
Great find :+1:

PR #9364, opened 2023-10-31
Optimized + Updatedd

ghost:
Done some small fixes and optimizations.

### DIFF
--- a/build_helpers/pre_commit_update.py
+++ b/build_helpers/pre_commit_update.py
@@ -29,11 +29,11 @@ mypy_repo = [repo for repo in f['repos'] if repo['repo']
 
 hooks = mypy_repo[0]['hooks'][0]['additional_dependencies']
 
-errors = []
-for hook in hooks:
-    if hook not in type_reqs:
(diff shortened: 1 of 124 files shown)


xmatthias:
I'm sorry, but this will not be accepted - not as is, anyway.

While _some_ individual enhancements may be good (probably not all are, at least in my experience using automated toolings) - getting this as a blob that's changing 100+ files at once is unacceptable, as the burden  / time to review is too high, especialy given that it's not going to provide any visible benefits to the users.

---

I'm not per se against such changes, so feel free to resubmit in smaller chunks (not necessarily individual files, but grouped by code section or directory, with a max of 10-20 individual changes) and we can have a look.

Please also make sure to not change tests and code at the same time / in the same PR. A change like this shouldn't require tests to change and viceversa - otherwise it modified behavior.

Please also don't use your own "develop" branch for pull requests - as that's your default branch, this can have some unintended side-effects, (including, but not limited to auto-closing your PR if you update your fork, ...).

froggleston:
Welcome to the brave new world of pseudo-devs submitting widespread unfocused changes to code by AI.

PR #9377, opened 2023-11-04
Set custom roi within custom_exit method

moinonin:
The PR allows users to set custom roi within the custom_exit method. This means that you can set a much higher roi in the main table e.g.
```
    minimal_roi = {
        "0": 0.32
    }
```
You may then want to use a different roi table as the asset risk progresses. In that case you can then do the following for instance:
        
```
ROI = {
            "0": 0.086,
            "6": 0.064,
            "37": 0.018
        } # within custom_exit method
```
If you are using MACD indicator to exit, you can place the custom roi afterwards like so:
```
if last_candle['macd'] > last_candle['macdsignal']:
    self.minimal_roi = ROI

```

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (changes requested):
I'll have a look once it'll actually show a diff in the PR - as it is now, there's no point in even running CI, as this file will be ignored completely.

I have no clue what the actual change is - but you've clearly commited the file wrongly (`interface.py` is now in the root of the repostory, where it won't be used by the bot). It should instead replace the existing interface.py file - which (if done corectly) - will highlight the _effective changes_ done to this file as part of the "files" tab in this PR.

moinonin:
Hey @xmatthias apologies for that error. The file is now removed from root, and placed in its proper dir freqtrade/strategy/interface.py. I only converted one of the inter-able variables into an integer.

stash86:
You might want to reset your PR. Now there's 599 changed files

xmatthias:
That's the (or one) problem with doing a PR from your own main branch.

you think you "sync" your Fork - and instead, it's messing up your whole PR - potentially auto-closing it.

I'd propose to look at the diff (click on the files tab) of the PR yourself when you think you're done - that tab should only contain the changes you INTENDED to do. now it's 599 files - which is clearly not the intent.

moinonin:
Indeed it was not my intention to modify 599 lines. Actually one insertion in one line. The only reason one cannot invoke roi within custom_exit is because of this line in interface.py:

```
1280        roi_list = [x for x in self.minimal_roi.keys() if x <= trade_dur]
```

Which works when x is converted into integer like so:

```
1280        roi_list = [x for x in self.minimal_roi.keys() if int(x) <= trade_dur]

```
That's all I wanted to change :dango:

xmatthias:
> Indeed it was not my intention to modify 599 lines

it's not 599 lines, but 599 **files** - which makes it way worse (not sure WHAT you did to the repo to get that :laughing: )

---

Assuming that's the change (without looking at it in code, just in the above comment):

i don't see how that would change anything. 
There's validation in place intentionally - and nothing preventing you from assigning the dict with int as keys (which is what the normalization does).
Messing with `minimal_roi` from within the strategy will skip that validation, most likely resulting in malfunctionings as some assumptions are no longer valid.

---

Aside from that - your whole logic of "reassigning roi" will fail the moment you have more than one open trade. `minimal_roi` is a global (class) variable - and is therefore valid for all pairs.

if the strategy  "decides" to change roi to "whatever" for one pair - then this new value will apply to every other pair as well - potentially (most likely) resulting in immediate exits for these other pairs.

As you seem to use custom_exit already - you should instead simply do something like the following:

```
...
        if current_profit > 0.2 and (current_time - trade.open_date_utc).days >= 1:
            return 'roi_2'
        if  if current_profit > 0.1 and (current_time - trade.open_date_utc).hours >= 1:
           return 'roi_1'
      ...
```

obviously with the values as you see fit.

xmatthias:
Based on the above assumed change - and the messed up history - i don't think we'll be able to accept this - at least not with this history (it'll make it seem like every file was changed on this date - which messed up all git history).

I don't think the intend of why you're trying change this will work - it MAY work limited to your isolated case / strategy - but in reality, it won't scale to other configs (it may even have unintended side-effects on your strategy).
We can for sure discuss this further - but for now, i don't see this as useful change - as it will have no effect - at least no intended one (it'll potentially slow backtesting slightly though).

PR #9400, opened 2023-11-07
Update process to analyse and manage open trade in priority

Axel-CH:
## Summary
This feature allow freqtrade, at each process loop to manage in priority open trades when "adjust_trade_position" is enabled. With this feature, the bot will not wait the full whitelist analysis to manage open trades. This feature is usefull when there is a substantial quantity of pairs in whitelist and user need to quickly manage open trades after a candle close.

With this feature I was able to divide the trade management lag by 3.

## Quick changelog

- add new method manage_open_trades to prevent duplicate code in process loop
- manage first open trades when position adjustment is enable

## What's new?
This PR improve position trade management time for open position live

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -225,8 +225,36 @@ class FreqtradeBot(LoggingMixin):
         strategy_safe_wrapper(self.strategy.bot_loop_start, supress_error=True)(
             current_time=datetime.now(timezone.utc))
 
-        self.strategy.analyze(self.active_pair_whitelist)
+        open_trade_pairs_analyzed = []
+        if self.strategy.position_adjustment_enable:
+            open_trade_pairs = [ot.pair for ot in trades]
+            self.strategy.analyze(open_trade_pairs)
 
+            open_trade_pairs_analyzed = open_trade_pairs
+            self.manage_open_trades()
+
+            # Check if we need to adjust our current positions before attempting to buy new trades.
+            with self._exit_lock:
+                self.process_open_trade_positions()
+
+        remaining_pairs_to_analyse = [pair for pair in self.active_pair_whitelist if pair not in open_trade_pairs_analyzed]
+
+        self.strategy.analyze(remaining_pairs_to_analyse)
+        self.manage_open_trades()
+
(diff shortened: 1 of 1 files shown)


freqtrade/freqtradebot.py
```diff
@@ -225,25 +225,23 @@ def process(self) -> None:
...
-        # Check if we need to adjust our current positions before attempting to buy new trades.
-        if self.strategy.position_adjustment_enable:
+            # Check if we need to adjust our current positions before attempting to buy new trades.
             with self._exit_lock:
                 self.process_open_trade_positions()
 
+        remaining_pairs_to_analyse = [pair for pair in self.active_pair_whitelist if pair not in open_trade_pairs_analyzed]
+
+        self.strategy.analyze(remaining_pairs_to_analyse)
+        self.manage_open_trades()
```
xmatthias, line 243: Seems like you're calling `manage_open_trades` twice per iteration - which we clearly wouldn't want, as that will call all callbacks twice per loop - potentially altering behavior in some odd strategies.

xmatthias (commented):
> With this feature I was able to divide the trade management lag by 3.

You should also note that it's at the expense of delaying entry to potential new trades by a much higher factor (which i hope you can provide based on your tests).

freqtrade/freqtradebot.py
Axel-CH: Yes it is the case, do you think we could manage_open_trade, except the ones that was already managed. As the goal of this feature is to manage in priority open trades, it think two call is required. If you think it is not worth it I can scrap this PR and focus on other way to speed up live trades management.

xmatthias: I don't understand the need for this call.

there is no need for this in my eyes, it'll handle the SAME trades again (there can't have been new trades opened between the first and 2nd call (except through force-entry - where it won't matter if they're handled the next iteration).

Axel-CH:
After successful experiments, i'm scrapping this in favor of dataframe multiprocessing custom indicators populate and merging. With the gained speed introduced by parallelization, prioritizing trade entry or management is no longer relevant.

PR #9407, opened 2023-11-11
add entries, exits, and mix_tags API endpoints

stash86:
## Summary

add entries, exits, and mix_tags API endpoints

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Great enhancement! 

I've added some small tests - so we won't accidentally break the output - but - great job - one i had on my "todo" list for a long time.
Thanks! :100:

PR #9408, opened 2023-11-11
Update freqai-feature-engineering.md

konradbeck:
I cannot find any reference to 'include_periods_candles'

## Summary

Solve the issue: feature_engineering_expand_all() mentions 'using indicator_periods_candles', feature_engineering_expand_basic() mentions not expanding on a parameter I am unable to find in the source.

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

Documentation will reference the correct parameters.


xmatthias (approved):
Great spot! :+1: 

I've found another instance of the "wrong" wording, which i've also fixed.

PR #9421, opened 2023-11-13
Update strategy-customization.md

konradbeck:
Shouldn't the timeperiod be 100 when defining a ema100?

## Summary

Solve the issue: correct indicator definition

## Quick changelog

- <change log 1>
- <change log 1>


stash86:
Good catch. I mistakenly changed it when I changed the startup

xmatthias (approved):
LGTM 👍

PR #9423, opened 2023-11-14
Clarify load option in hyperopt docs

froggleston:
## Summary

This PR will update the hyperopt docs to make the `load` (and `optimize`) docs clearer. 

## What's new?

* A more specific subsection in the Parameter Types section regarding what `optimize` and `load` do.
* A tip section that notes `load` also affects parameter values in backtesting

### DIFF
--- a/docs/hyperopt.md
+++ b/docs/hyperopt.md
@@ -14,7 +14,7 @@ To learn how to get data for the pairs and exchange you're interested in, head o
 
 !!! Note
     Since 2021.4 release you no longer have to write a separate hyperopt class, but can configure the parameters directly in the strategy.
-    The legacy method is still supported, but it is no longer the recommended way of setting up hyperopt. 
+    The legacy method is still supported, but it is no longer the recommended way of setting up hyperopt.
     The legacy documentation is available at [Legacy Hyperopt](advanced-hyperopt.md#legacy-hyperopt).
 
 ## Install hyperopt dependencies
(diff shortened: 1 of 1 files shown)


froggleston:
(Sorry for the extra diff cruft due to whitespace removal. Consider the tidying up a bonus! 😄 )

docs/hyperopt.md
```diff
@@ -198,7 +198,7 @@ Hyperopt will then spawn into different processes (number of processors, or `-j
 
 For every new set of parameters, freqtrade will run first `populate_entry_trend()` followed by `populate_exit_trend()`, and then run the regular backtesting process to simulate trades.
 
-After backtesting, the results are passed into the [loss function](#loss-functions), which will evaluate if this result was better or worse than previous results.  
+After backtesting, the results are passed into the [loss function](#loss-functions), which will evaluate if this result was better or worse than previous results.
```
xmatthias, line 201: Unfortunately it's ... not just "tidying" - not in all instances, anyway. 
2 spaces have a special meaning in markdown, and mean "new line without new paragraph" - so this will now change (quite drastically) how it renders.

from 
image

to 
image

(all mashed together) - ignore the highlight, that's me searching for the same section

For most changes - it's fine (like in the code blocks, ...) - though in the (few) cases where the following regex matched - we need to revert this (regex `\s\s$` - basically 2 spaces followed by a newline)

---

i assume you're using something like the following setting - and i'd advise to add the markdown section (the below is vscode config ...).
Obviously this will help with future changes - not for "this time" ...


```
    "files.trimTrailingWhitespace": true,
    "[markdown]": {
        "files.trimTrailingWhitespace": false,
    },
```

froggleston:
Oh, that makes sense, and that snippet is really useful. I'll revert my changes, use your setting, and then likely make a new PR to keep things easier.

xmatthias:
Reverting the few affected lines would've been fine though - but up to you (just make sure to NOT lose the actual enhancement - as that was great! ;) )

PR #9429, opened 2023-11-15
Clarify hyperopt docs for parameter options

froggleston:
## Summary

This PR will update the hyperopt docs to make the load (and optimize) docs clearer.

## What's new?

A more specific subsection in the Parameter Types section regarding what optimize and load do.
A tip section that notes load also affects parameter values in backtesting


xmatthias (approved):
Great! :+1:

PR #9435, opened 2023-11-18
Improve trade close profit assertions

xmatthias:
## Summary

Improve calculation of intermediate, relative "realized" profit.

Changes to existing orders will only apply once the next order is filled (or if `recalculate_from_orders()` is somehow called for other reasons).


stash86 (dismissed):
the rel_profit ratio are all okay

PR #9436, opened 2023-11-19
Remove order list from /status, and add new /order command

stash86:
## Summary

The order list was appended at the end of /status reply with assumption that users will use reasonable number of position adjustment. With current situation where users use unlimited number, the message will become too long. It's better to split them into two separate commands then.

Still to-do

- [x] Fix old test
- [x] Add new test
- [x] Documentation

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


freqtrade/rpc/telegram.py
```diff
@@ -652,9 +694,9 @@ async def _status_msg(self, update: Update, context: CallbackContext) -> None:
                         "*Open Order:* `{open_orders}`"
                         + ("- `{exit_order_status}`" if r['exit_order_status'] else ""))
 
-            lines_detail = self._prepare_order_details(
-                r['orders'], r['quote_currency'], r['is_open'])
-            lines.extend(lines_detail if lines_detail else "")
+            # lines_detail = self._prepare_order_details(
+            #     r['orders'], r['quote_currency'], r['is_open'])
+            # lines.extend(lines_detail if lines_detail else "")
```
xmatthias: is this PR still [wip] ?? looks pretty much complete if i'm honest ...

aside from that, we would probably want to remove these lines (instead of commenting them) - obviously also applies to the test file ...

stash86: Forgot to remove the WIP tag :D. Gonna remove the comments

freqtrade/rpc/telegram.py
```diff
@@ -652,9 +694,6 @@ async def _status_msg(self, update: Update, context: CallbackContext) -> None:
                         "*Open Order:* `{open_orders}`"
                         + ("- `{exit_order_status}`" if r['exit_order_status'] else ""))
 
-            lines_detail = self._prepare_order_details(
-                r['orders'], r['quote_currency'], r['is_open'])
-            lines.extend(lines_detail if lines_detail else "")
```
xmatthias, line 657: looks great! :+1: 

The only question i asked myself while playing with it is - should the regular "status" message include a hint about "you can get order details via `/order <orderid>` - basically "promoting" this command ...
i'm not sure i'd otherwise realize that there's a command with more details ...

xmatthias (approved):
LGTM :+1:

Issue #9432, opened 2023-11-17
freqai - timerange bug in dataprovider.py with long training period


hamzz91:
## Describe your environment

  * Operating system: windows
  * Python Version: 3.11
  * CCXT version: 4.1.8
  * Freqtrade Version: freqtrade 2023.10-dev-cfda8516e
  

## Describe the problem:

I use a freqai strategy with 5m timeframe with "train_period_days"= 180 . If I put "1d" in include_timeframes, the lib fails

### Steps to reproduce:

  1. downlad the data : `freqtrade download-data -c .\config_examples\config_freqai.example_modif.json --timeframes 5m 1d --days 1800`
  2. Use the FreqaiExampleStrategy from the template and the attached config file [config_freqai.example_modif.json](https://github.com/freqtrade/freqtrade/files/13394210/config_freqai.example_modif.json) and any model (catboostRegressor for example)
  3. run the bot (I used the webserver, not sure if it's related)
  
### Observed Results:

  Using the debugger, I could see the "timerange" object of the dataprovider gets corrupted (the properties "startfmt" and "startdt" throw an error). This happens after the line 318 in the dataprovider.py file : `timerange.subtract_start(tf_seconds * startup_candles)`
The "startts" variable becomes negativ.

### Relevant code exceptions or logs

Note: Please copy/paste text of the messages, no screenshots of logs please.

  ```
2023-11-17 16:59:06,997 - freqtrade.rpc.api_server.api_backtest - ERROR - Backtesting caused an error: [Errno 22] Invalid argument
Traceback (most recent call last):
  File "E:\PycharmProjects\cryptobot\freqtrade\rpc\api_server\api_backtest.py", line 84, in __run_backtest_bg
    min_date, max_date = ApiBG.bt['bt'].backtest_one_strategy(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\PycharmProjects\cryptobot\freqtrade\optimize\backtesting.py", line 1291, in backtest_one_strategy
    preprocessed = self.strategy.advise_all_indicators(data)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\PycharmProjects\cryptobot\freqtrade\strategy\interface.py", line 1335, in advise_all_indicators
    return {pair: self.advise_indicators(pair_data.copy(), {'pair': pair}).copy()
[...]

PR #9450, opened 2023-11-20
Bug: freqai backtesting startup_candle_count handling

robcaulk:
Backtesting with FreqAI introduces a bug with large timeframes as indicated by this issue #9432. This bug is not present in dry/live operations.

The function `get_required_startup()` assumes that the `startup_candle_count` is set in the highest timeframe candle size (so in #9432 it assumes that `startup_candle_count` is in days. However, FreqAI (and FT normal) uses `startup_candle_count` inside  `load_bt_data()` to get the full training period loaded in for the user on the **base timeframe**. So currently we are setting this `startup_candle_count` for the base timeframe including the training period as an effort to ensure `load_bt_data()` gets the full training data for the base timeframe. But after setting `startup_candle_count` to this large basetimeframe value, the `get_required_startup()` would use this higher `startup_candle_count` value for higher timeframes in combination with the user set `train_period_days`. Although for the current typical FreqAI usecases, this was not noticeable (usually users stick to 5m, - 4h timeframes together), it could be catastrophic for users who have unexpectedly broad ranges of parameters set. For example, we see thousands of days would become "required" for a config where the user set a base timeframe of `5m` with a high `train_period_days` in combination with a high maximum `include_timeframes`. In #9432, the user set `train_period_days` to 180 and also uses `include_timeframes` of `1d`.

### DIFF
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -332,15 +332,14 @@ class DataProvider:
         if not freqai_config.get('enabled', False):
             return self._config.get('startup_candle_count', 0)
         else:
-            startup_candles = self._config.get('startup_candle_count', 0)
             indicator_periods = freqai_config['feature_parameters']['indicator_periods_candles']
             # make sure the startupcandles is at least the set maximum indicator periods
-            self._config['startup_candle_count'] = max(startup_candles, max(indicator_periods))
+            needed_candles = max(indicator_periods)
             tf_seconds = timeframe_to_seconds(timeframe)
             train_candles = 0
             if add_train_candles:
                 train_candles = freqai_config['train_period_days'] * 86400 / tf_seconds
-            total_candles = int(self._config['startup_candle_count'] + train_candles)
+            total_candles = int(needed_candles + train_candles)
             logger.info(f'Increasing startup_candle_count for freqai to {total_candles}')
             return total_candles
 
--- a/freqtrade/data/dataprovider.py
+++ b/freqtrade/data/dataprovider.py
@@ -332,14 +332,15 @@ class DataProvider:
         if not freqai_config.get('enabled', False):
             return self._config.get('startup_candle_count', 0)
         else:
+            startup_candles = self._config.get('startup_candle_count', 0)
             indicator_periods = freqai_config['feature_parameters']['indicator_periods_candles']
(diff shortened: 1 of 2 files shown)


wagnercosta (approved):
I've conducted some tests, and everything seems good to me

xmatthias (commented):
The point i'm currently concerned is backtesting L1211 - where `_set_dataframe_max_index` is using self.required_startup.

I'm dumping this here without full investigation - but as a reminder to look at potential impacts (this can mess up how strategies can retrieve already analyzed dataframes). 
While this may be an edgecase for freqAI - it's not something that's explicitly prohibited, so should not allow "look in the future" behaviors.

robcaulk:
> The point i'm currently concerned is backtesting L1211 - where `_set_dataframe_max_index` is using self.required_startup.
> 
> I'm dumping this here without full investigation - but as a reminder to look at potential impacts (this can mess up how strategies can retrieve already analyzed dataframes). While this may be an edgecase for freqAI - it's not something that's explicitly prohibited, so should not allow "look in the future" behaviors.

Thanks @xmatthias , I checked into your concern. We do not use `get_analyzed_dataframe()` anywhere in FreqAI, so there will be no issues.

People can still use this in their strategy, and still it will not affect their FreqAI functionality.

xmatthias:
> People can still use this in their strategy, and still it will not affect their FreqAI functionality.

@robcaulk that's the point where i disagree.
imagine the following method in a freqAI strategy:

``` python
    def custom_stake_amount(self, pair: str, current_time: datetime, current_rate: float,
                            proposed_stake: float, min_stake: Optional[float], max_stake: float,
                            leverage: float, entry_tag: Optional[str], side: str,
                            **kwargs) -> float:

        dataframe, _ = self.dp.get_analyzed_dataframe(pair=pair, timeframe=self.timeframe)
        current_candle = dataframe.iloc[-1].squeeze()
        ....
        return proposed_stake

```

(or any other callback that uses an analyzed dataframe, for example to place a stoploss at "atr * 2").
This WILL affect their freqAI backtesting results - as the dataframe you get in this callback is now wrong - leading to lookahead bias (you get the WRONG last candle).

you're right that non-freqAI strategies will not be affected - but it's still introducing a new bug for everyone using callbacks with freqAI.

PR #9485, opened 2023-11-30
exposing bot_name to discord rpc field

stash86:
## Summary

Sometimes users will use one same strategy on multiple bots. Having bot_name available will make it easier to diferentiate which bot send which message

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
Nice :+1:

PR #9493, opened 2023-12-03
chore: fix future warning on pandas

robcaulk:
Latest pandas versions are throwing a future warning:

```
freqtrade/freqai/data_drawer.py:400: FutureWarning:

The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
```

This PR fixes it.


xmatthias (approved):
LGTM :+1:

PR #9494, opened 2023-12-03
Add bot_name to discord messages

froggleston:
## Summary

Discord messages do not contain the bot_name, making it difficult to see which bot is raising those messages. This PR adds this field to send_msg()

## Quick changelog

- add the bot_name config option to the discord rpc, defaulting to 'freqtrade' as per the config docs


xmatthias:
how's this differ from #9485 - other than having added it "the other way" around (essentially before/after other options seems reversed).

froggleston:
Well, it isn't 😄 I didn't see that PR come in. I'll close this one.

PR #9505, opened 2023-12-04
Fix multiple new Pandas 2.1 future warnings in FreqAI startup

robcaulk:
Pandas 2.1 brought with it a slew of future warnings. Two more fixed here including:

```
FutureWarning:

The behavior of pd.concat with len(keys) != len(objs) is deprecated. In a future version this will raise instead of truncating to the smaller of the two sequences
```

And

```
You are trying to merge on datetime64[ns, UTC] and object columns for key 'date_pred'. If you wish to proceed you should use pd.concat
```


smarmau:
Confirmed fixed issues had with existing model and pandas 2.1. 

Thanks Rob.

xmatthias (approved):
LGTM :+1: 

(ignore the random error - that's cause the proxy is acting out ATM).

PR #9515, opened 2023-12-07
In partial exit, do full exit if remaining == 0

stash86:
## Summary

In partial exit, do full exit if remaining == 0. Current behavior reject the exit even if remaining is gonna be zero.

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


perkmeister:
I think it should check for dust or below minimum trading amount remaining and fully exit.

freqtrade/freqtradebot.py
```diff
@@ -685,7 +685,7 @@ def check_and_call_adjust_trade_position(self, trade: Trade):
                 return
 
             remaining = (trade.amount - amount) * current_exit_rate
-            if min_exit_stake and remaining < min_exit_stake:
+            if min_exit_stake and (0 < remaining < min_exit_stake):
```
xmatthias: Looking at this check due to the backtesting failure, i don't think it's correct.
we want to skip if
- remaining is negative (user provided a "too high" stake)
- remaining is between 0 and min_exit_stake.

We want it to pass however if remaining is 0 exactly.

as such - we need to either explicitly check for "remaining = 0" - or exclude 0 from the check somehow.

xmatthias: @stash86 would be great if you could have a look at both the above comment, as well as on the actual change i did.

stash86: sure

docs/strategy-callbacks.md
```diff
@@ -760,21 +760,30 @@ The `position_adjustment_enable` strategy property enables the usage of `adjust_
...
+
 Additional entries are ignored once you have reached the maximum amount of extra entries that you have set on `max_entry_position_adjustment`, but the callback is called anyway looking for partial exits.
 
-Position adjustments will always be applied in the direction of the trade, so a positive value will always increase your position (negative values will decrease your position), no matter if it's a long or short trade.
+### Decrease position
 
-Modifications to leverage are not possible, and the stake-amount returned is assumed to be before applying leverage.
+The strategy is expected to return a negative stake_amount (in stake currency) for a partial exit.
+Returning `-trade.stake_amount` results in a full exit.
+Returning a value more than `-trade.stake_amount` (so remaining stake_amount would become negative) will result in the bot ignoring the signal.
```
stash86: @xmatthias won't this line incorrect since in current code, the bot will reduce the amount to match trade_amount, which means the exit will be successful in the end

xmatthias: partially - the idea was to make this a hard requirement.

i removed that part in the backtesting code - but apparently forgot to remove it in the live section ...

stash86:
anything else needs to be done for this PR?

xmatthias:
> anything else needs to be done for this PR?

I think we're good now ... but the docs had a typo / wrong information until recently :laughing:

xmatthias (approved):
LGTM :+1:

PR #9559, opened 2023-12-16
Improve log message when no good hyperopt result is found 

stevanStevic:
## Summary

This PR improves log message for the user by differentiating between closing the hyperopt too soon and not finding result good enough.

## Quick changelog

- Differentiate log message for closing the hyperopt too soon and not finding result good enough.

## What's new?

When running hyper optimization, if no good result is found `current_best_epoch` variable is not updated. Example of such run can be seen in screenshot below:

image

This leads to print `No epochs evaluated yet, no best result.` which is not true. 

This PR differentiates between closing hyperopt too quick or not having a good result actually, by checking if there are some epochs saved which means they've been evaluated.

### DIFF
--- a/freqtrade/optimize/hyperopt.py
+++ b/freqtrade/optimize/hyperopt.py
@@ -637,6 +637,8 @@ class Hyperopt:
 
             HyperoptTools.show_epoch_details(self.current_best_epoch, self.total_epochs,
                                              self.print_json)
+        elif self.num_epochs_saved > 0:
+            print("No good result found for given optimization function.")
         else:
             # This is printed when Ctrl+C is pressed quickly, before first epochs have
             # a chance to be evaluated.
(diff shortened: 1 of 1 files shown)


xmatthias:
out of interest - what (lossfunction) did you use to get no best epoch?

i've just tried with 3-4 different ones - and they all give the first as "best" - at least with my current test strategy ...

(it's not something that'll really affect this PR - but i'm curious, as i currently struggle to even test the change :laughing: )

stevanStevic:
Well, I too have stumbled upon this by chance, so it's not the usual behavior.

Setup was following:
- I used `OnlyProfitHyperOptLoss` loss function. 
- The strategy was custom PSAR trend based on [TradingView`s SAR](https://www.tradingview.com/support/solutions/43000502597-parabolic-sar-sar/) with 3 parameters.
- I was optimizing those 3 parameters in `buy` space. 
- It was done with ADAUSDT on `1m` timeframe with `"--timerange 20231020-20231120"`
- The output on 10 epochs you can see in PR description.

If this could help you to reproduce it, here it is.

xmatthias (approved):
Yep found it, it happens whenever all "objective" values are `> 100` - which is the start value.

For the update itself - LGTM :+1:

PR #9576, opened 2023-12-18
Make freqai example strat *even* simpler

robcaulk:
The FreqAI example strat includes hyperopt parameters right now, which likely confuses new users. We remove those here.

We also relax buy and sell requirements to make sure it buys and sells a lot, to avoid the inevitable complaints "no buys" in the discords.

Warning of "not for production" remains highlighted in the strat, and highlighted in red in the documentation. Nothing more we can do about that part.


xmatthias (approved):
LGTM :+1:

PR #9577, opened 2023-12-18
Add SKLearnRandomForestClassifier

robcaulk:
We've seen evidence in local testing (going back even multiple years to early FreqAI tests) to suggest that the RandomForestClassifier is quite effective.

There is really no good reason not to make this model easily available to users. This PR adds the functionality.


xmatthias (approved):
Nice :+1:

PR #9613, opened 2023-12-30
Rpc/fix telegram forcebuy

xmatthias:
## Summary

Fix problems with db session handling when shifting to an async excutor.

## Quick changelog

- Fix odd bug which can return cached responses.


Heistergand (approved):
I originally mentioned this error on discord. I tested the scenario in a new freqtrade environment using the fix_telegram_forcebuy branch. It works as expected.

Issue #4917, opened 2021-05-11
Permission issues on docker images


snussik:
While running the container on linux I stated to meet the following issue:
```
PermissionError: [Errno 13] Permission denied: '/freqtrade/user_data/config.json'
```
I think that It may be cause of I have to run docker compose under sudo:
`
sudo docker-compose up -d
`
But I don't think its a reason.

xmatthias:
using sudo with compose is always wrong and points to an installation problem.

The docker-images recently changed to run as user (no longer as root).
As such, they can no longer read files which have wrong permissions (like root).

Fixing this can be as simple as running `docker-compose run --rm freqtrade create-userdir --userdir user_data` again (which will attempt to fix permissions in the user directory).

If this does not fix your issue and you'll require more assistance, then please fill out the issue template (the one you deleted when opening this issue) - otherwise i will not be able to provide any further help as i have to take guesses about your environment.

snussik:
`Fixing this can be as simple as running docker-compose run --rm freqtrade create-userdir --userdir user_data again (which will attempt to fix permissions in the user directory)` this command and `docker-compose build` after helped.

Buy the way, I faced up with this bug too: #4839 and think that they are of the same nature. In my workflow I don't create files on server - just upload them and it (I think) causes permission issues.

alexander171294:
If you are using fedora, or SELinux distributions, and you see permission error running freqtrade create-userdir 

```
2021-11-16 00:35:56,668 - freqtrade.configuration.directory_operations - WARNING - Could not chown user_data
2021-11-16 00:35:56,668 - freqtrade - ERROR - Fatal exception!
Traceback (most recent call last):
  File "/freqtrade/freqtrade/main.py", line 37, in main
    return_code = args['func'](args)
  File "/freqtrade/freqtrade/commands/deploy_commands.py", line 26, in start_create_userdir
    userdir = create_userdata_dir(args["user_data_dir"], create_dir=True)
  File "/freqtrade/freqtrade/configuration/directory_operations.py", line 69, in create_userdata_dir
    subfolder.mkdir(parents=False)
  File "/usr/local/lib/python3.9/pathlib.py", line 1323, in mkdir
    self._accessor.mkdir(self, mode)
PermissionError: [Errno 13] Permission denied: 'user_data/backtest_results'
ERROR: 1
```

Then, you have this problem: https://stackoverflow.com/questions/24288616/permission-denied-on-accessing-host-directory-in-docker

[...]

PR #9645, opened 2024-01-06
Fix: Add create-userdir command to docker-compose for permission handling

simwai:
## Summary

This PR adds a command to the Docker Compose file to create the user directory with the correct permissions, which solves a common permission error.

Solve the issue: #4917

## Quick changelog

- Added `create-userdir --userdir user_data` command to Docker Compose file.

## What's new?

This PR improves the Docker Compose file by adding a command to create the user directory with the correct permissions every time the Docker container starts. This change helps prevent a common permission error that occurs when the Docker container does not have the necessary permissions to access certain files or directories. This change should make the software more robust and user-friendly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,8 +30,11 @@ services:
       - "127.0.0.1:8080:8080"
     # Default command used when running `docker compose up`
     command: >
+      /bin/bash -c "
+      create-userdir --userdir user_data &&
       trade
       --logfile /freqtrade/user_data/logs/freqtrade.log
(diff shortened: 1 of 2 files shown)


stash86:
This is not needed if you don't switch between users. Your issue was that you switch between users.

There is nothing to be fixed on FT side if I'm being honest

simwai:
I didn't switch between users. This prevents a permission denied error for creating files when you run docker-compose with root rights without any manual user switching which helps to avoid some docker pain. It also shouldn't have any downside for people running docker-compose with non-root permissions.

PR #9646, opened 2024-01-06
Fixed docker compose permissions when running it as root user

simwai:
## Summary

This PR adds a command to the Docker Compose file to create the user directory with the correct permissions, which solves a common permission error.

Solve the issue: #4917

## Quick changelog

- Added `create-userdir --userdir user_data` command to Docker Compose file.

## What's new?

This PR improves the Docker Compose file by adding a command to create the user directory with the correct permissions every time the Docker container starts. This change helps prevent a common permission error that occurs when the Docker container does not have the necessary permissions to access certain files or directories. This change should make the software more robust and user-friendly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,8 +30,11 @@ services:
       - "127.0.0.1:8080:8080"
     # Default command used when running `docker compose up`
     command: >
+      /bin/bash -c "
+      create-userdir --userdir user_data &&
       trade
       --logfile /freqtrade/user_data/logs/freqtrade.log
       --db-url sqlite:////freqtrade/user_data/tradesv3.sqlite
       --config /freqtrade/user_data/config.json
       --strategy SampleStrategy
+      "
--- a/docker/docker-compose-freqai.yml
+++ b/docker/docker-compose-freqai.yml
@@ -28,9 +28,12 @@ services:
       - "127.0.0.1:8080:8080"
     # Default command used when running `docker compose up`
     command: >
+      /bin/bash -c "
+      create-userdir --userdir user_data &&
       trade
       --logfile /freqtrade/user_data/logs/freqtrade.log
       --db-url sqlite:////freqtrade/user_data/tradesv3.sqlite
       --config /freqtrade/user_data/config.json
       --freqaimodel XGBoostRegressor
       --strategy FreqaiExampleStrategy
+      "


xmatthias (changes requested):
Permission errors will not be fixed by this command. 
If the directory already exists with wrong permissions, create-userdir will do nothing to fix that.
if the directory didn't exist, then following the docker quickstart guide will suffice.

---

Aside from that - I'm not sure what you're after / what problem you're trying to fix - but one thing is clear - you didn't test your change (it's not working and will break the compose file).


I'd strongly recommend to follow the [docker quick start guide](https://www.freqtrade.io/en/latest/docker_quickstart/#docker-quick-start) - which will teach you how to run create-userdir manually (this should be a manual step, as it's not desired at all times).

I'll be leaving this open for the moment to give you a chance to explain your intend better (especially what particular problem you're trying to fix).

However in it's current state - i do see this as something we're not going to accept (it's not working - which also means you didn't test the change yourself - that should be reason enough).

simwai:
Hey xmatthias, sry for that. I will work out what I did wrong there and why it does not work. You are right that I didn't test it, and that is something I regret. This will never happen anymore.

I didn't follow exactly the quick start guide, which is indeed also not good, but I assumed for a docker-compose -f file.yml up -d I wouldn't need a guide. This lead to wrong permissions on some files/folders which could be prevented. 
I mean why should it be necessary to run the create dictionary command manually before using the docker-compose file.

The intent behind this PR was to avoid a permission denied error when docker-compose is run with root permissions. As you can see in the linked issue #4917 other people had the problem too. I understand that you say running docker-compose with root permissions is a bad thing, but a lot of users like me, don't know how to do it better, and IMO when there is a solution which can help to make it work with root permission or without root permission used it should be used here. I also would include a timezone sync in the next PR I make, because this is also something which makes sense IMO.

PS: I hope you get somehow what I mean.

xmatthias:
Following the documentation is however a requirement.
Not doing so is the actual mistake / problem.

Permission errors (if the directory has different permissions) in docker are being fixed as part of the freqtrade startup process.
Running `create-userdir` as part of every startup can however have very adverse effects - and will actually NOT fix permission problems if the user permissions are wrongly set.

The issue you link to is almost 3 years old - at which point, we didn't have the "Permission fix" yet - so what you're trying to fix has been fixed years ago already.

For the moment / with the information provided, i don't see any future for this PR.

---

As to the timezone sync - we used to have this - however iId advise against that (you can obviously use this on your own - but it's not something we should have by default for all users).

the only "real" effect it'll have is on the log timestamps - however exchanges operate at UTC time, and it's usually simpler / desirable to have logs in UTC timestamp, not in some random, local timestamp, which makes comparing these dates with actual activity quite difficult.

simwai:
Thank you for taking the time to explain your perspective regarding the proposed changes to the docker-compose.yml file and Dockerfile. I appreciate your insights on the potential implications of these modifications.

I understand your concerns about the automatic creation of the user_data directory and the synchronization of the timezone from the Docker container to the host. Your points about the existing solutions for permission errors, the potential adverse effects of running create-userdir at startup, and the benefits of having logs in UTC for consistency and simplicity are well taken.

While my initial intention was to streamline the setup process, I now see how these changes could introduce new complexities and potential issues. I respect your deep understanding of the project and its users’ needs, and I agree that it’s crucial to consider these factors when proposing changes.

Even though the changes I proposed may not be incorporated into the main repository, I’ve learned a lot from this discussion. I’ll continue to use my modified version to suit my specific needs, and I look forward to contributing to freqtrade in other ways in the future.

Thank you again for your time and consideration.

PR #9648, opened 2024-01-07
Pickle Problem

Industrial:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/bin/backtesting
+++ b/bin/backtesting
@@ -2,12 +2,12 @@
 set -e
 echo "> backtesting"
 # freqtrade \
+# --strategy-path /freqtrade/user_data/strategies \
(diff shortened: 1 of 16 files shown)


xmatthias:
To me, this seems like you should instead open an issue considering the only actual information provided is "Pickle Problem" - but the changes itself don't address any pickle-related topics within freqtrade.

If your aim is to submit this as a pull request, then please
* don't use your develop branch as source, but a dedicated branch that's not your main branch
* make sure only files you intended to include are actually included (this PR includes 10 new files we surely won't accept).
* include a description including WHAT you aim archive with the PR.

PR #9659, opened 2024-01-09
Add MarketCapPairList based on CoinGecko

stash86:
## Summary

Add MarketCapPairList based on CoinGecko. For now, it will fetch only list of top 250 marketcap coins.

- [x] Is list of top 250 enough?
- [x] Make sure docs are updated
- [x] Work on test. Might need @xmatthias 's help on how to test this pairlist

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,182 @@
...
+
+            if self._mode == 'top_rank':
+                top_marketcap = marketcap_list[:self._number_assets:]
+
+                for pair in pairlist:
+                    base = pair.split('/')[0]
+                    if base.lower() in top_marketcap:
+                        filtered_pairlist.append(pair)
+                    else:
+                        logger.info(f"Remove {pair} from whitelist because it's not ranked "
```
xmatthias: should use `self.log_once(` ... to avoid logging on every iteration in some odd cases.

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,182 @@
...
+        else:
+            data = self._coingekko.get_coins_markets(vs_currency='usd', order='market_cap_desc',
+                                                     per_page='250', page='1', sparkline='false',
+                                                     locale='en')
+            if data:
+                marketcap_list = [row['symbol'] for row in data]
+                self._marketcap_cache['marketcap'] = marketcap_list
+                can_filter = True
+
+        if can_filter:
```
xmatthias: do we need `can_filter` here ? i think checking for `marketcap_list` again should work, too ?

docs/includes/pairlists.md
```diff
@@ -227,6 +228,29 @@ The optional `bearer_token` will be included in the requests Authorization Heade
 !!! Note
     In case of a server error the last received pairlist will be kept if `keep_pairlist_on_failure` is set to true, when set to false a empty pairlist is returned.
 
+#### MarketCapPairList
+
+`MarketCapPairList` employs sorting/filtering of pairs by their marketcap rank based of CoinGecko. It will only recognize coins up to the coin placed at rank 250. The number of pairs in the resulted pairlist will be slightly different depends on the `mode` defined in the config (available mode are `top_rank` and `total_assets`). The marketcap data from
```
xmatthias: this kinda looks ... incomplete :laughing: 
though i have no idea what you wanted to say in that last sentence ...

docs/includes/pairlists.md
```diff
@@ -227,6 +228,29 @@ The optional `bearer_token` will be included in the requests Authorization Heade
...
+        "refresh_period": 86400
+    }
+]
+```
+
+##### `top_rank` mode
+In this mode, it will return pairlist consist of active and not-blaclisted pairs that are placed at the top `number_assets` rank of the marketcap.
+
+##### `total_assets` mode
+In this mode, it will return pairlist consist of `number_assets` number of active and not-blaclisted pairs sorted by their marketcap rank.
```
xmatthias: blacklisted ? (same above ...)

docs/includes/pairlists.md
stash86: indeed. I might have left it and forgot about it

freqtrade/plugins/pairlist/MarketCapPairList.py
```diff
@@ -0,0 +1,178 @@
...
+                                                     locale='en')
+            if data:
+                marketcap_list = [row['symbol'] for row in data]
+                self._marketcap_cache['marketcap'] = marketcap_list
+
+        if marketcap_list:
+            filtered_pairlist = []
+
+            if self._mode == 'top_rank':
+                top_marketcap = marketcap_list[:self._number_assets:]
```
xmatthias: Are you sure that this is the correct approach (at least to be the default)?

running this with 30 - it's quite surprising that i only get about 20 pairs in the resulting pairlist ... ?

stash86: Well, that's correct, as USDT is ranked 3rd for example. I believe your exchange don't have USDT/USDT pair. On which exchange do you try this? And do you use blacklist?

xmatthias: i understand the reason :laughing: 
but i'm unsure if it's the correct approach (like, shouldn't there be 2 settings?).

assume i'd like to have 40 pairs ... but only from within the top 50 marketcap pairs (so if 12 are not available in my pairlist, my pairlist would only be 38 pairs long, otherwise, i'll always get 40 pairs).

right now, the 2 settings are mutually exclusive (changed via "mode") .. so i have to pick, that i _either_ get to have 40 pairs (total_assets mode) - _or_ pairs from within the top 50 marketcap pairs (top_rank mode).

it's _probably_ possible via some combination of 2 sequential pairlists (didn't test, hence probably) - but with coingecko, that almost certainly isn't good idea, as it'd mean we have 3 coingecko objects around - doing 2x the same request in series (each pairlist is it's own instance) (the 3rd is fiatconvert).

In my initial understanding (before looking at the code in detail) was that `mode` would switch the sorting of the coingecko pairs - essentially one of `market_cap_asc, market_cap_desc, volume_asc, volume_desc, id_asc, id_desc` (taken from the API endpoint docs). if that makes sense is a diff. discussion.

an alternative approach to get the above scenario (and be more flexible overall) could be the following:
* `number_assets` is optional - defining the max length of the pairlist (as it's handled in most other pairlists)
* new setting "marketcap_top" (or whatever better name) defines the marketcap "top" pairs (essentially how many coingecko pairs we'll consider).

that 2nd setting is capped at 250 - and lower values will obviously mean there's fewer values to pick from.

stash86: > assume i'd like to have 40 pairs ... but only from within the top 50 marketcap pairs (so if 12 are not available in my pairlist, my pairlist would only be 38 pairs long, otherwise, i'll always get 40 pairs).

You can do this using stacked filter. First one on `total_assets` mode using 40 `number_assets`. Then second one on `top_rank` mode on 50 `number_assets`. I should add this example in the docs.

> it's probably possible via some combination of 2 sequential pairlists (didn't test, hence probably) - but with coingecko, that almost certainly isn't good idea, as it'd mean we have 3 coingecko objects around - doing 2x the same request in series (each pairlist is it's own instance) (the 3rd is fiatconvert).

Well, the cache is 1d cache, and in theory, assuming the cache expired, only the first pairlist call the API, the second pairlist will use the cache. Don't think it would work differently compared to stacked volumepairlist. Gonna add a log_once when the bot fetch data, so we would know if there is a rare case of double API call

xmatthias: > Well, the cache is 1d cache, and in theory, assuming the cache expired, only the first pairlist call the API, the second pairlist will use the cache

that's incorrect i'm afraid.
The cache is per instance - not a global cache - so if you specify 20 MarketCapPairlists in sequence, then you'll do 20 calls to coingecko (20 identical calls, but still), as each entry into pairlists is a seperate, independent instance.

stash86: ah okay. Let's see if I can change the implementation

xmatthias: i'm ... not sure if turning this into a global (so cache applies for everything) is a good idea.
there's quite a few modes (webserver -> test-pairlist for example) where this can greatly mess up behavior...

xmatthias (approved):
looks pretty great now :100: 

250 assets will def. work for now. if we see tons of requests for more coins, we can pretty easily implement pagination - but i don't feel it's necessary for now.

LGTM :+1:

Immortality-IMT:
Hi, is it possible to get the best of both volume and market cap. Has this been done. For instance xcrypto has volume rank of 7 but a marketcap rank of 17. Giving a combined rank: a weighted average formula could be:
Combined Rank=w1×Market Cap Rank+w2×Volume Rank where w1 and w2​ are the weights assigned to the market cap rank and volume rank, respectively. Ensure that w1+w2=1.

Create the list then loop through the list...
```
$combinedRankList = [];
$weightMarketCap = 0.5; // Weight for market cap rank
$weightVolume = 0.5; //Weight for volume 


 $combinedRank = $weightMarketCap * $crypto['market_cap_rank'] + $weightVolume * $crypto['volume_rank'];
 $crypto['combined_rank'] = $combinedRank;
 $combinedRankList[] = $crypto;  // sort it to the amount of pairs required
```

PR #9660, opened 2024-01-09
Fix failure on using FullTradesFilter on test-pairlist (and maybe other non-trading mode)

stash86:
## Summary

add Try block to catch the failure on using FullTradesFilter on non-dry/live run

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/plugins/pairlist/FullTradesFilter.py
+++ b/freqtrade/plugins/pairlist/FullTradesFilter.py
@@ -48,8 +48,13 @@ class FullTradesFilter(IPairList):
         :return: new allowlist
         """
         # Get the number of open trades and max open trades config
-        num_open = Trade.get_open_trade_count()
-        max_trades = self._config['max_open_trades']
+        try:
+            num_open = Trade.get_open_trade_count()
+            max_trades = self._config['max_open_trades']
+        except AttributeError:
+            # Performancefilter does not work in backtesting.
+            self.log_once("FullTradesFilter is not available in this mode.", logger.warning)
+            return pairlist
 
         if (num_open >= max_trades) and (max_trades > 0):
             return []


freqtrade/plugins/pairlist/FullTradesFilter.py
```diff
@@ -48,8 +48,13 @@ def filter_pairlist(self, pairlist: List[str], tickers: Tickers) -> List[str]:
         :return: new allowlist
         """
         # Get the number of open trades and max open trades config
-        num_open = Trade.get_open_trade_count()
-        max_trades = self._config['max_open_trades']
+        try:
+            num_open = Trade.get_open_trade_count()
```
xmatthias, line 52: What's the particular problem this is trying to solve? (how does it reproduce ...)?

If you look at the `get_open_trade_count()` implementation, then i don't think it will / can fail with that error - as it does have a "backtesting" fallback (which doesn't really work, as in these modes, you'll always get "0" as answer due to the timing / location this is being called ...)

https://github.com/freqtrade/freqtrade/blob/e28b2f44d3ef79c87c0c8f52af94f46355142f1a/freqtrade/persistence/trade_model.py#L1277-L1287

stash86: Got this on `test-pairlist`

```
2024-01-09 05:14:13,502 - freqtrade - ERROR - Fatal exception!
  File "/home/stash/test/freqtrade/freqtrade/main.py", line 42, in main
  File "/home/stash/test/freqtrade/freqtrade/commands/pairlist_commands.py", line 30, in start_test_pairlist
    pairlists.refresh_pairlist()
  File "/home/stash/test/freqtrade/freqtrade/plugins/pairlistmanager.py", line 106, in refresh_pairlist
    pairlist = pairlist_handler.filter_pairlist(pairlist, tickers)
  File "/home/stash/test/freqtrade/freqtrade/plugins/pairlist/FullTradesFilter.py", line 51, in filter_pairlist
    num_open = Trade.get_open_trade_count()
  File "/home/stash/test/freqtrade/freqtrade/persistence/trade_model.py", line 1283, in get_open_trade_count
    return Trade.session.execute(
AttributeError: type object 'Trade' has no attribute 'session'
```

stash86: it's easy to reproduce it, just use this pairlist setting on `test-pairlist` command
```
{
    "pairlists": [
        {
            "method": "VolumePairList",
            "number_assets": 200,
            "sort_key": "quoteVolume",
        },
        {"method": "FullTradesFilter"},
    ],
}
```

xmatthias: You're right - but your fix is only fixing the symptom - not the cause (essentially, it means that other pairlists WILL encounter the same error eventually).

I've created a pr that's actually fixing this for all pairlists in #9665 ... would you mind giving that a look / try? thanks.

stash86:
#9665 works perfectly

PR #9665, opened 2024-01-10
Add No db context to protect certain functions that run in a backtest-like mode

xmatthias:
## Summary

Trade.use_db is not only used by backtesting, but is also assumed for pairlists if the bot is not in live-mode, hence we need a way to set this appropriately.

## Quick changelog

* Extract Trade.use_db setter methods
* Add context manager to set and unset Trade.use_db methods


stash86:
Works perfectly

Issue #5578, opened 2021-09-15
Incremental trade download and candle building


aplainzetakind:
## Describe your environment

  * Operating system, Python, CCTX: Running in docker on linux
  * Freqtrade Version: 2021.8
  
## Your question

I am downloading trades from Kraken. It seems that even if I only fetch a very short time span of trades, all the past trades are loaded into memory in a dataframe, and then new files are produced from there. I have been fetching trades a couple of months at a time starting at the beginning of 2017, and as I approached the present, I started to run out of RAM with 55 million BTC/EUR trades. The process requires roughly 30GB of memory, and because I ran out I had to create a swap file and use that, which of course made everything even slower. Together with producing the candles (I'm creating all up to 1h), it has been running for about 5 hours now. This is quite untenable and I don't see myself updating my historic data any further. My questions are:
- Is this normal or is something wrong on my side?
- If this is indeed normal, is there a design reason for not doing things with streaming file I/O in constant memory? Also, candle conversion seems to be making a pass for each target timeframe; I imagine there'd be room to improve it to produce all target candles in one pass.
- If there are no reasons such approaches are unacceptable, would you be open to patches attempting to improve those points?

xmatthias:
Downloading (and converting) trades data into candles (ohlcv) is not a core feature - but was developed as a neccessity to work around shortcommings in the kraken API (no historic candles).

There's for sure room for improvement, so if you're up to it, feel free to improve this part (it's a time question for me - and as it's only a "kraken-problem" - i see it as not quite high priority.

As you're already pointing out (file streaming, ...) - it's probably not something that's easily / quickly done, and to avoid overlaps / mistakes in the data, there'll be for sure needs to have good tests in place.

That said ... i could also imagine a seperate "convert" command, which splits the conversion logic out (so you don't have to download data "up to now" only because you want additional 1h data (on top of the existing 5m data)).
I think that would also simplify testing (no need to wait for downloading ...) - and should probably be the starting point (obviously NOT with 30GB of data ... you'll become crazy testing with this).

---

The only think i'm not sure is the conversion from trades into different timeframes at the same time.
While it might sound like a good idea, i'm not sure it is:
You'll need to batch the data (say, load weekly/monthly batches) to avoid loading GB's of trades into memory at the same time - and then convert it to candles.

it'll most likely not make sense to persist the candles per batch.
[...]

PR #9690, opened 2024-01-15
Update kraken.py to change pagination type to use timestamp (instead of id)

jasonbhart:
## Summary

Changes Kraken exchange default config to fix system losing track of trade history downloaded prior.

Solve the issue: #5578 

## What's new?

Switches Kraken exchange default config to use 'time' (from 'id') value for "trades_pagination" which fixes oddities surrounding system losing track of what trade history it has downloaded prior when creating historical candles. 

Relevant example CLI command: freqtrade download-data --exchange kraken --dl-trades

The existing config will frequently re-download all trade data for a pair instead of just the missing trade data.  No other code changes.

Reviewing the current Kraken API documentation, it seems the "since timestamp" method is the only method supported anymore, so a "since id" method doesn't seem to be desired behavior.

### DIFF
--- a/freqtrade/exchange/kraken.py
+++ b/freqtrade/exchange/kraken.py
@@ -226,19 +226,3 @@ class Kraken(Exchange):
             fees = sum(df['open_fund'] * df['open_mark'] * amount * time_in_ratio)
 
         return fees if is_short else -fees
-
-    def _trades_contracts_to_amount(self, trades: List) -> List:
-        """
-        Fix "last" id issue for kraken data downloads
-        This whole override can probably be removed once the following
-        issue is closed in ccxt: https://github.com/ccxt/ccxt/issues/15827
-        """
-        super()._trades_contracts_to_amount(trades)
-        if (
-            len(trades) > 0
-            and isinstance(trades[-1].get('info'), list)
-            and len(trades[-1].get('info', [])) > 7
-        ):
-
-            trades[-1]['id'] = trades[-1].get('info', [])[-1]
-        return trades
--- a/freqtrade/exchange/kraken.py
+++ b/freqtrade/exchange/kraken.py
@@ -29,7 +29,7 @@ class Kraken(Exchange):
         "order_time_in_force": ["GTC", "IOC", "PO"],
         "ohlcv_candle_limit": 720,
         "ohlcv_has_history": False,
-        "trades_pagination": "id",
+        "trades_pagination": "time",
         "trades_pagination_arg": "since",
         "mark_ohlcv_timeframe": "4h",
     }
--- a/tests/exchange/test_exchange.py
+++ b/tests/exchange/test_exchange.py
@@ -3489,7 +3489,7 @@ def test_merge_ft_has_dict(default_conf, mocker):
 
     ex = Kraken(default_conf)
     assert ex._ft_has != Exchange._ft_has_default
-    assert ex.get_option('trades_pagination') == 'id'
+    assert ex.get_option('trades_pagination') == 'time'
     assert ex.get_option('trades_pagination_arg') == 'since'
 
     # Binance defines different values


xmatthias:
I'm not sure why you think this would fix #5578 ...
that's completely unrelated (effectively, even unrelated to downloading data itself), but refers to the method of conversion if all trades for a pair (say for 3 years of BTC/EUR) don't fit into memory.

I also don't really have experienced the "downloads all data" with kraken. 
The datadownload seems stable as far as i've tested it - and may feel like it's downloading everything, as even downloading data for one day can take a significant amount of time at kraken's very large ratelimit ...

jasonbhart:
@xmatthias This PR is meant to fix the behavior described in the issue description for #5578 :
"I am downloading trades from Kraken. It seems that even if I only fetch a very short time span of trades, all the past trades are loaded into memory in a dataframe, and then new files are produced from there. " .  I ran into similar, and this change fixed the downloading of all data (instead of incremental) when using trade data to calculate historical candles.  

If you feel it is unrelated, then my mistake, but there is still an issue with the current Kraken config when using 'download-data' with --dl-trades. If you review the Kraken API documentation, youll see timestamp is the method they use with an incremental refresh, and seems to imply other methods are not supported and/or valid (such as using 'id' to denote the starting point of the incremental download).

xmatthias:
From the issue: 

> I am downloading trades from Kraken. It seems that even if I only fetch a very short time span of trades, **all the past trades are loaded into memory in a dataframe**, and then new files are produced from there.

(highlights added for clarity) 

What this refers to is the following process:
* you have 3 years of trades data on your disk, up to the last week, which is missing
* You download the last week's trades data (5 days) (that's the data you're missing)
* download-data will then load all 3 years of trades-data into memory to do the conversation to ohlcv data.

This last step is what the issue talks is about - and is _potentially_ problematic if you're on a lower powered machine - as you can run out of memory (or cause extensive swapping).
This could **in theory** (optionally) be split into chunks, converting the data for example month by month - avoiding a huge memory spike.
how the data is **downloaded**  (what you're trying to change here) through pagination (id or time based) is not related to this issue.

---

That doesn't necessarily mean current behavior is correct ... however in my testing, i can't see anything wrong with it, either.

> there is still an issue with the current Kraken config when using 'download-data' with --dl-trades. If you review the Kraken API documentation, youll see timestamp is the method they use with an incremental refresh, and seems to imply other methods are not supported and/or valid (such as using 'id' to denote the starting point of the incremental download).

I understand your approach to this - but is it an actual problem?
Like - can you show a log where a kraken download had existing data - but deleted and re-downloaded everything? (please make sure that the timerange you're giving to the 2nd download-data call starts AFTER the existing data start date, otherwise it's intended behavior).

I know the documentation talks about time-based pagination - however - the data ALSO contains a "last:" entry - which ccxt is moving to the last array element of the info field, and which we're then using as trade_id (here: https://github.com/freqtrade/freqtrade/blob/cfdf21942d72ae9923c253b18aeac7af18a919a7/freqtrade/exchange/kraken.py#L236-L244).

image
(some logs that show how this works:)
```
2024-01-16 06:44:21,192 - freqtrade.exchange.exchange - DEBUG - Exchange milliseconds: 1705383861192
2024-01-16 06:44:21,192 - freqtrade.exchange.exchange - DEBUG - Fetching trades for pair BTC/USDT, since 1702791861190 (2023-12-17T05:44:21.190000+00:00) ...
2024-01-16 06:44:32,078 - freqtrade.exchange.exchange - DEBUG - Fetching trades for pair BTC/USDT, params: {'since': '1702827632572366277'} 
2024-01-16 06:44:36,985 - freqtrade.exchange.exchange - DEBUG - Fetching trades for pair BTC/USDT, params: {'since': '1702851889795452997'} 
```

Now I'm fully aware that this is also a timestamp - but it's a more precise timestamp - and considering kraken doens't provide an actual trade-id - i'm not entirely sure what we'd gain by switching to the less precise date.

---

If you're convinced that that'll give us benefits (either through simplification, or through enhanced precision), i'm not against it necessarily but it's not a one-line change (at least not the testing / confirmation stuff).

It'd require:
* cleanup of the above mentioned function, as it's no longer necessary and could lead to problems in deduplicating trades
* passing tests
* some Logs which show that the new behavior iterates correctly (i've little doubt about that as the time-based method is proven and tested, but would still like to verify against kraken)
* a comparison of "generated" data with actual data (essentially using the `fetch_ohlcv` method, pick a timeframe, and validate / compare the last 700(ish) candles which kraken provides to ensure we're not suddenly missing trades, which would cause different volume / candles.

jasonbhart:
@xmatthias We good to re-close this without merge?

xmatthias:
yeah, we've got a better fix now :+1:

PR #9703, opened 2024-01-19
feat: allow custom user features to be plotted in backtesting (freqai)

robcaulk:
Currently, FreqAI allows users to monitor custom training features during dry/live in FreqUI by prepending the feature names with `%%`.

However, backtesting does not currently permit custom feature returns. This PR aims to fix that, as well as an improvement to the documentation to make it easier for users to understand how the naming scheme would be managed in their plot.


xmatthias (approved):
:+1:

PR #9704, opened 2024-01-19
fix: logical error in the conditional checks for model classes

ShaneOss:
Fix logical error in the conditional checks for model classes. The `elif` statement that looks for "lightgbm.sklearn" or "xgb" in the model class string is now broken into two separate conditions because the old condition would always evaluate to `True` due to the non-empty string "xgb".

### DIFF
--- a/freqtrade/freqai/utils.py
+++ b/freqtrade/freqai/utils.py
@@ -123,7 +123,7 @@ def plot_feature_importance(model: Any, pair: str, dk: FreqaiDataKitchen,
         elif "xgb" in str(mdl.__class__):
             feature_importance = mdl.feature_importances_
         else:
-            logger.info('Model type does not support for generating feature importances.')
+            logger.info('Model type does not support generating feature importances.')
             return
 
         # Data preparation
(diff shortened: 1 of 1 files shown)


xmatthias (dismissed):
Purely looking at it i know this is the right approach.
I'd want @robcaulk to give his final say though - as i see him as code-owner for this part of the code

LGTM :+1:

freqtrade/freqai/utils.py
```diff
@@ -118,10 +118,12 @@ def plot_feature_importance(model: Any, pair: str, dk: FreqaiDataKitchen,
...
         if "catboost.core" in str(mdl.__class__):
             feature_importance = mdl.get_feature_importance()
-        elif "lightgbm.sklearn" or "xgb" in str(mdl.__class__):
+        elif "lightgbm.sklearn" in str(mdl.__class__):
+            feature_importance = mdl.feature_importances_
+        elif "xgb" in str(mdl.__class__):
             feature_importance = mdl.feature_importances_
         else:
-            logger.info('Model type not support for generating feature importances.')
+            logger.info('Model type does not support for generating feature importances.')
```
robcaulk: If we are going to improve the grammar, let's fix it fully :)

"Model type does not support generating feature importances"

ShaneOss: Thanks Rob. Yeah I noticed that after creating the PR. Cheers!

xmatthias (approved):

PR #9705, opened 2024-01-21
shorten configuration class

PabloRuizCuevas:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

I think making the class shorter makes it more readable, we could also consider making configuration a dataclass instead of using a dictionary.


xmatthias:
please base your change off of the latest develop (that'll potentially include re-submitting this PR - otherwise you'll completely mess up your fork, though i'll include steps for that below, too).


For the "re-submit PR" approach:
you should be able to get there with something around the following (please replace <origin> with the upstream you're using)

```

git checkout -b shorten_config
git rebase develop
# Fix eventual problems - though i don't think there will be any

git push -u <origin> shorten_config
# Re-submit PR from this new branch.
```

---

The alternative (without re-submitting) would be the following:
```
git rebase develop
git push -f
```

However, this will have the drawback of changing your "stable" branch - which you'll then most likely want to delete and recreate (it should work without problems too, though).

---

> we could also consider making configuration a dataclass instead of using a dictionary.

No, dataclass is absolutely the wrong approach for this.
configuration can have arbitrary keys (we won't know what a user submits, could be `stake_amountNOTAKTIVE`) - which will have no effect in the current approach - but would fail when it's a dataclass, as this is not a known key.

While there are potential, other alternatives - i don't think they will provide much benefits over the current approach (at least not considering the time-invest - and that the time can be better spent in a way that actually improves bot behavior / functionality).

PR #9708, opened 2024-01-21
shorten configuration code

PabloRuizCuevas:
## Summary

Solve the issue: #___

## Quick changelog

- <Reduce verbosity>

## What's new?

Remade pr

### DIFF
--- a/freqtrade/configuration/configuration.py
+++ b/freqtrade/configuration/configuration.py
@@ -68,8 +68,10 @@ class Configuration:
         config: Config = load_from_files(self.args.get("config", []))
 
         # Load environment variables
-        env_data = enironment_vars_to_dict()
-        config = deep_merge_dicts(env_data, config)
+        from freqtrade.commands.arguments import NO_CONF_ALLOWED
+        if self.args.get('command') not in NO_CONF_ALLOWED:
+            env_data = enironment_vars_to_dict()
+            config = deep_merge_dicts(env_data, config)
 
         # Normalize config
         if 'internals' not in config:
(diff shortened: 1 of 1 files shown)


freqtrade/configuration/configuration.py
```diff
@@ -68,10 +68,8 @@ def load_config(self) -> Dict[str, Any]:
         config: Config = load_from_files(self.args.get("config", []))
 
         # Load environment variables
-        from freqtrade.commands.arguments import NO_CONF_ALLOWED
-        if self.args.get('command') not in NO_CONF_ALLOWED:
-            env_data = enironment_vars_to_dict()
-            config = deep_merge_dicts(env_data, config)
+        env_data = enironment_vars_to_dict()
+        config = deep_merge_dicts(env_data, config)
```
xmatthias: We'll definitely want to keep this conditional.

it's intentionally there (added recently) - protecting some odd scenarios where environment variables mess with commands that shouldn't load a configuration.

xmatthias (changes requested):
There's one too  many changes (see other comment).

---

For the future, please do Pull requests from a different Branch (not your main branch).

An explanation as to why you should use feature branches can be found [here](https://hynek.me/articles/pull-requests-branch/).

keep this PR as is though (this is for an eventual next time ...) - you can update this PR by simply committing and pushing to this branch again.

freqtrade/configuration/configuration.py
PabloRuizCuevas: Should be solved, thanks for the explanation.

xmatthias (approved):
LGTM 👍

PR #9730, opened 2024-01-25
Improve common mistakes docs

froggleston:
Add more details to the common mistakes section of the strategy customisation docs.


xmatthias (approved):
Great! :100:

PR #9733, opened 2024-01-26
Remotepairlist: fix parsing json exception , add saving to a file of processed pairlist + docs

Bloodhunter4rc:
## Summary

Extend the exception catch to the parsing, so last pairlist is returned.
Add Saving to a file

Solve the issue: https://discord.com/channels/700048804539400213/824719649417199646/1200374477968654446
Parsing Json Exception

## What's new?

Add saving of processed pairlist to a file, to enable other freqtrade instances to read from that file. + preventing unnecessary requests and keeping the pairlist in sync.

### DIFF
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -210,8 +210,7 @@ The `pairlist_url` option specifies the URL of the remote server where the pairl
 
 The `save_to_file` option, when provided with a valid filename, saves the processed pairlist to that file in JSON format. This option is optional, and by default, the pairlist is not saved to a file.
 
-??? Note
-    Example:
+??? Example "Multi bot with shared pairlist example"
     
     `save_to_file` can be used to save the pairlist to a file with Bot1:
 
--- a/docs/includes/pairlists.md
+++ b/docs/includes/pairlists.md
@@ -208,7 +208,42 @@ In "append" mode, the retrieved pairlist is added to the original pairlist. All 
 
 The `pairlist_url` option specifies the URL of the remote server where the pairlist is located, or the path to a local file (if file:/// is prepended). This allows the user to use either a remote server or a local file as the source for the pairlist.
 
-The `save_to_file` option, when provided with a valid filename, saves the processed pairlist to that file in JSON format. This option is optional, and by default, the pairlist is not saved.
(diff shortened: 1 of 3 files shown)


freqtrade/plugins/pairlist/RemotePairList.py
```diff
@@ -236,15 +237,15 @@ def gen_pairlist(self, tickers: Tickers) -> List[str]:
...
-
                         try:
+                            # Load the JSON data into a dictionary
+                            jsonparse = rapidjson.load(json_file, parse_mode=CONFIG_PARSE_MODE)
                             pairlist = self.process_json(jsonparse)
                         except Exception as e:
                             if self._init_done:
                                 pairlist = self.return_last_pairlist()
                                 logger.warning(f'Error while processing JSON data: {type(e)}')
+                                logger.debug(f'Error while processing JSON data: {e}')
```
xmatthias: what's the difference / expectation to log both warning and debug ... ? 
if you're after the reason (like, stacktrace) - you should enable stacktrace instead (or log at exception level, where this is auto-enabled) ... ?

docs/includes/pairlists.md
```diff
@@ -207,6 +208,8 @@ In "append" mode, the retrieved pairlist is added to the original pairlist. All
 
 The `pairlist_url` option specifies the URL of the remote server where the pairlist is located, or the path to a local file (if file:/// is prepended). This allows the user to use either a remote server or a local file as the source for the pairlist.
 
+The `save_to_file` option, when provided with a valid filename, saves the processed pairlist to that file in JSON format. This option is optional, and by default, the pairlist is not saved.
```
xmatthias: i think you might wanna extend / reword this section ...

Starting with this (i red the docs change first) i understood:
"i give 3 bots on the same server the same filename, and they'll identify "do i need to reload this file or can i use it as is" - and then auto-load from here, OR fall back to a network request - refreshing the file.

That's however not really what's happening - the actual assumption is that 
* bot A fetches and saves to file
* bots B and C are configured with this file as "pairlist_url" ... 

Now they both use different syntax (`file:///somefile.json` vs. `./somefile.json`) - which i think will need some more explanation (or maybe a config sample for the 2 bot setups in a `??? Note` block - which defaults to collapsed)  to be actually clear.

xmatthias (approved):
nice

LGTM :+1:

PR #9755, opened 2024-02-01
Update ws_client.py

bvince21:
No changes made. Test for class project

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


stash86:
We would appreciate if you took your class project somewhere else. Thanks

PR #9776, opened 2024-02-07
fix strategy-updater docs

stash86:
## Summary

Fix typo on `strategy-list` explanation

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/docs/utils.md
+++ b/docs/utils.md
@@ -990,11 +990,7 @@ options:
   -h, --help            show this help message and exit
   --strategy-list STRATEGY_LIST [STRATEGY_LIST ...]
                         Provide a space-separated list of strategies to
-                        backtest. Please note that timeframe needs to be set
-                        either in config or via command line. When using this
-                        together with `--export trades`, the strategy-name is
-                        injected into the filename (so `backtest-data.json`
-                        becomes `backtest-data-SampleStrategy.json`
+                        be converted.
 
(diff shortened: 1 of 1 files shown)


docs/utils.md
```diff
@@ -990,11 +990,7 @@ options:
   -h, --help            show this help message and exit
   --strategy-list STRATEGY_LIST [STRATEGY_LIST ...]
                         Provide a space-separated list of strategies to
-                        backtest. Please note that timeframe needs to be set
-                        either in config or via command line. When using this
-                        together with `--export trades`, the strategy-name is
-                        injected into the filename (so `backtest-data.json`
-                        becomes `backtest-data-SampleStrategy.json`
+                        be converted.
```
xmatthias, line 993: well this section is just the output of  `freqtrade strategy-updater --help` ... so by changing this, the documentation is now showing something else in the help output than the actual command.

stash86: I don't think most people will ever use `--help` flag on CLI. Docs is where most people would go to, and get referred to. So I prefer accurate docs over inaccurate docs just to match `--help` output.

Adding a new arg on `cli_options` seems overkill just to have matching output, don't you think so? But if you prefer that route, then I can add a new arg for strategy-updater

xmatthias (approved):

PR #9777, opened 2024-02-07
Kubernetes helm chart tested on minikube for Mac

DerSalvador:
## Kubernetes Helm Chart 

This helm chart is derived from a version running in live mode for 6 months on GKE and modified for using locally on minikube for developers or as an alternative to docker-compose or native scripts.

### DIFF
--- a/docker/helm-chart/strategies/Strategy001.py
+++ b/docker/helm-chart/strategies/Strategy001.py
@@ -1,0 +1,80 @@
+# --- Do not remove these libs ---
+from freqtrade.strategy import IStrategy
+from typing import Dict, List
+from functools import reduce
+from pandas import DataFrame
+# --------------------------------
+import talib.abstract as ta
+import freqtrade.vendor.qtpylib.indicators as qtpylib
+
+class Strategy001(IStrategy):
+    INTERFACE_VERSION = 3
+    '\n    Strategy 001\n    author@: Gerald Lonlas\n    github@: https://github.com/freqtrade/freqtrade-strategies\n\n    How to use it?\n    > python3 ./freqtrade/main.py -s Strategy001\n    '
+    INTERFACE_VERSION: int = 3
+    # Minimal ROI designed for the strategy.
+    # This attribute will be overridden if the config file contains "minimal_roi"
+    minimal_roi = {'60': 0.01, '30': 0.03, '20': 0.04, '0': 0.05}
+    # Optimal stoploss designed for the strategy
+    # This attribute will be overridden if the config file contains "stoploss"
+    stoploss = -0.1
+    # Optimal timeframe for the strategy
+    timeframe = '5m'
+    # trailing stoploss
+    trailing_stop = False
+    trailing_stop_positive = 0.01
+    trailing_stop_positive_offset = 0.02
+    # run "populate_indicators" only for new candle
+    process_only_new_candles = True
+    # Experimental settings (configuration will overide these if set)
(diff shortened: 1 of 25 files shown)


docker/helm-chart/strategies/Strategy001.py
```diff
@@ -0,0 +1,80 @@
+# --- Do not remove these libs ---
```
xmatthias, line 1: We'll clearly not want the strategies in this repo.

The sample-strategies are intentionally separated into a seperate repository - as they should serve as samples (and that should become clear through that).

docker/helm-chart/values-minikube.yaml
```diff
@@ -0,0 +1,226 @@
+---
+config:
```
xmatthias, line 2: Freqtrade uses config files in .json format. 

i don't really see how this fits into the picture. Aside from that, it seems way too opinionated.

docs/kubernetes_quickstart.md
```diff
@@ -0,0 +1,69 @@
...
+This helm chart is like Freqtrade for educational purposes. Any professional use is not recommended. Run Freqtrade in dry-run (see configs).
+Please check the logs after deployment for 
+
+```
+freqtrade.exchange.exchange - INFO - Instance is running with dry_run enabled
+```
+
+## Prerequistes
+
+Look for the appropriate latest version of your preferred OS. Ask ChatGPT/Google to help you.
```
xmatthias, line 12: we will NOT recommend the use of chatGPT in the docs. 

it's answers are flaky at best and strongly depend on the actual question asked.

xmatthias (changes requested):
The strategies are a clear no-go - i'd appreciate if you can remove them in a "commit --amend" - so they're gone from history, too.

Before doing so / investing more time into this - pleaes read onwards though.

---

I'm not per se against having a minimal helm chart / kubernetes deployment approach ... 

however, this is adding 10+ additional files for a pretty much niche case - which by merging, we'd take responsibility to actually support going forward.
The different, yet important things seem spread out across basically all files. I'm sure its' working fine for _your_ usecase - but for someone who didn't write this - it's pretty much impossible to get this up an running quickly - at least while undestanding what they're actually running) - not least because it's taking assumptions which may simply be wrong for another user - like the forced `--freqai-model` parameter.


It's also suggesting that k8s is a good deployment platform for freqtrade - which i personally don't think it is (it's not really sane to run freqtrade on a k8s cluster if you can run it on a simple docker host).
The overhead alone is insane. 

It MAY be fine if you've got a cluster running - but having it part of the freqtrade docs / install instructions suggests that it's a recommended way to run it - which i don't think it should be.

---

For now - i'll leave this as "request changes" due to the strategy topic (i didn't look at the rest in full depth) - but i don't think this is something we're able to accept - not because of the contribution per se - but mainly because i don't see us being able to support this should users have questions about this.


This would make it a "dump and run" merge  - which we'd have to tell users "it's there, hope you can get it working - we'll not be able to help you, good luck" - which pretty much makes it pointless - leading to

* questions we know already we probably can't answer
* frustrated users as they started with an installation method that is seemingly supported, but is not really as they can't really ask questions
* increased maintenance burden - as we need to waste our time explaining users why we have it documented, but can't support it.

Now i'm open for further discussion / other opinions on this - which is why i'm leaving this open for now.

froggleston:
`which is why i'm leaving this open for now` - I guess not @xmatthias :D

PR #9814, opened 2024-02-15
Clarify processing_mode for RemotePairlist docs

froggleston:
No default value is specified in the docs for the `processing_mode`, making it unclear that the default behaviour is `filter` to filter out pairs, rather than append.


xmatthias (approved):
Nice catch :+1:

PR #9818, opened 2024-02-16
Added escaping to enter and exit tags on telegram performance messages.

CaffeinatedTech:
## Summary

Escaping entry and exit tags for telegram messages, as odd characters can break the functionality.

Solve the issue: no issue

## Quick changelog

- add escaping for `enter_tag`, `exit_tag` and `mix_tag` to telegram functions `_enter_tag_performance`, `_exit_reason_performance`, and `_mix_tag_performance`.

## What's new?

When a strategy uses odd characters such as `<, >, &` in the enter or exit tags, then the telegram commands /entries, /exits, and /mix_tags no longer return any result, and a warning is displayed in the log.  I have added escaping using `html.escape` to those tags when the message is being constructed.

```
2024-02-16 10:41:18,142 - freqtrade.rpc.telegram - WARNING - Telegram NetworkError: Can't parse entities: unsupported start tag "" at byte offset 97! Trying one more time.
2024-02-16 10:41:18,406 - freqtrade.rpc.telegram - WARNING - TelegramError: Can't parse entities: unsupported start tag "" at byte offset 97! Giving up on that message.
```

### DIFF
--- a/freqtrade/rpc/telegram.py
+++ b/freqtrade/rpc/telegram.py
@@ -4,7 +4,6 @@
 This module manage Telegram communication
 """
 import asyncio
-import html
 import json
 import logging
 import re
@@ -1376,21 +1375,21 @@ class Telegram(RPCHandler):
             pair = context.args[0]
 
         trades = self._rpc._rpc_enter_tag_performance(pair)
-        output = "<b>Entry Tag Performance:</b>\n"
+        output = "*Entry Tag Performance:*\n"
         for i, trade in enumerate(trades):
             stat_line = (
-                f"{i + 1}.\t <code>{html.escape(trade['enter_tag'])}\t"
+                f"{i + 1}.\t `{trade['enter_tag']}\t"
                 f"{fmt_coin(trade['profit_abs'], self._config['stake_currency'])} "
                 f"({trade['profit_ratio']:.2%}) "
-                f"({trade['count']})</code>\n")
+                f"({trade['count']})`\n")
 
             if len(output + stat_line) >= MAX_MESSAGE_LENGTH:
-                await self._send_msg(output, parse_mode=ParseMode.HTML)
+                await self._send_msg(output, parse_mode=ParseMode.MARKDOWN_V2)
                 output = stat_line
             else:
                 output += stat_line
 
-        await self._send_msg(output, parse_mode=ParseMode.HTML,
+        await self._send_msg(output, parse_mode=ParseMode.MARKDOWN_V2,
                              reload_able=True, callback_path="update_enter_tag_performance",
                              query=update.callback_query)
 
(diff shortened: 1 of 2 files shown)


xmatthias:
I'm ... not entirely convinced that this is the right place to fix this...

i assume this kind of tags may also cause problems in regular buy / sell messages (i never used such odd characters to be honest) - and most likely, will also cause  problems in freqUI (did you test this? can you show screenshots about this not causing issues, maybe - so i don't have to spin up a strategy with odd tags :laughing: )

If the issue is also present for other rendering methods, then the proper fix is most likely to remove non-supported characters from the tags before storing these to the database instead of trying to fix all possible places where this _might_ cause problems (which this pr, is not addressing all i'd suspect problems...).

CaffeinatedTech:
You are correct, there are a couple of other spots in `rpc/telegram.py` that `enter_tag` and `exit_reason` are placed into a message to be sent to telegram.  I can fix those if this is the chosen path.

My concern with altering the user's tags before persisting them is, what else could rely on that exact naming within their strategy?  If we change their tags out from under them, they might get weird errors, or bugs in their strategy.  There is also the matter of already 'dirty' databases, which contain such tags.  That said, the proper way would be to sanitise the `user` input before putting it in the database, who knows what silly things they can do.

It doesn't look like frequi has a problem with these strings, and the console doesn't care.

My situation is that I have a strategy that contains tags with characters that telegram doesn't like.  I got it from a shared resource, so other people definitely have the same strategy.  I hadn't noticed any problem with buy and sell messages, but they may have just failed silently (in the logs I'm not constantly looking at).  Trying to get tag performance stats is where I learned of the issue.

Would you like me to patch these last couple of spots in `rpc/telegram.py` by escaping the string, or should I look into sanitising the values before they hit the DB? - a little point in the right direction would help me with that option.

xmatthias:
i suspect you might've noticed if the entry/exit messages would be missing.

regular messages do to use parse_mode=Markdown - which i don't think has this problem.

That might be an option for these messages, too - as i think the real problem comes from `parse_mode=ParseMode.HTML` 
 (as opposed to `parse_mode=ParseMode.MARKDOWN`) - changing of which will however also require the adaptation of the actual message to markdown format.

we're however not "actually" using any html formatting beyond some `<b>` and `<code>` tags - which should be easily replacable with markdown code (and might even make things more concise / aligned.

>  If we change their tags out from under them, they might get weird errors, or bugs in their strategy.

that's a documentation issue though - we also truncate the string if it's too long - if that's a concern - the docs do have the details. (i do however think that the above approach might be the better one).

CaffeinatedTech:
I have removed the HTML escaping, and updated `_enter_tag_performance`, `_exit_reason_performance`, and `_mix_tag_performance` to use Markdown instead of HTML for the telegram message.

xmatthias (approved):
LGTM :+1:

PR #9862, opened 2024-02-24
[+] Adding requirements and deleting redundant if

Neufal777:
## Summary

Adding pytest requirement

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>

## What's new?

Adding pytest requirement and formatting redundant if statement

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,6 @@
+pytest==6.2.5
+pytest-xdist==2.5.0
+
 numpy==1.26.4
 pandas==2.1.4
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,3 +1,6 @@
+pytest==6.2.5
+pytest-xdist==2.5.0
```
xmatthias, line 2: Why would we force all users to install dev dependencies? (and old ones at that)?

pytest is required to run tests - and is included (and pinned) in requirements-dev.txt.

PR #9066, opened 2023-08-18
[Feature] fetch public trades

TheJoeSchr:
# Orderflow
## Summary

This PR enables to use ccxt fetch_public_trades so freqtrade can use trades data for backtesting and trading
### Quickstart:
- use `--dl-trades` to fetch trades for timerange
- enable using trades in `config.json`
```
"exchange": {
   ...
   "use_public_trades": true,
}
```
- set orderflow processing configuration in `config.json`:
 ```
"orderflow": {
    "scale": 0.5,
    "stacked_imbalance_range": 3,
    "imbalance_volume": 1,
    "imbalance_ratio": 300
  },
```
## TODO
- [x] [docs] write (new?) docs about how to access this new trades data in `populate_indicators` etc
  - [x] `dataframe['trades']`
  - [x] `dataframe['orderflow']`
  - [x] `dataframe['delta']`
- [x] [docs] how to use `--dl-trades` to get trade data
- [x] [docs] which configuration options are available and how to use them
- [x] [tests] unit tests exist, but they are using too big for github test files. so I need to shrink them and rewrite tests to still make sense

from previous feedback by @xmatthias :
- [x] Some diffs are a bit strange at first glance (especially in the exchange class), so expect some questions once i look at it more carefully (might be the webUI doing the diffs oddly if 2 functions are next to each other and are similar enough though) ... 
- [x] and i usually don't like random formatting changes (like - an unnecessary linebreak in a function that's not even touched otherwise)  => fixed with commit 137ee07

- [x] i also assume we'll want to move some things around in the end (converter seems to become quite big now) - but i can eventually help with that.

PR #9878, opened 2024-02-27
chore: extract `timeframe_to_*` functions into `exchange_utils_timeframe.py`

TheJoeSchr:
## Summary

Solves the circular import issue discussed in : #9066 

## What's new?

`exchange_utils_timeframe.py` now contains all `timeframe_to_*` functions which then get re-exported in `exchange/__init__.py`. this extraction helps with importing only the `timeframe_to_*` functions

### DIFF
--- a/freqtrade/exchange/exchange_utils.py
+++ b/freqtrade/exchange/exchange_utils.py
@@ -11,25 +11,12 @@ from ccxt import (DECIMAL_PLACES, ROUND, ROUND_DOWN, ROUND_UP, SIGNIFICANT_DIGIT
 
 from freqtrade.exchange.common import (BAD_EXCHANGES, EXCHANGE_HAS_OPTIONAL, EXCHANGE_HAS_REQUIRED,
                                        SUPPORTED_EXCHANGES)
-from freqtrade.exchange.exchange_utils_timeframe import (timeframe_to_minutes, timeframe_to_msecs,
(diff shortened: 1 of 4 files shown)


freqtrade/exchange/exchange_utils.py
```diff
@@ -11,13 +11,26 @@
 
 from freqtrade.exchange.common import (BAD_EXCHANGES, EXCHANGE_HAS_OPTIONAL, EXCHANGE_HAS_REQUIRED,
                                        SUPPORTED_EXCHANGES)
+from freqtrade.exchange.exchange_utils_timeframe import (timeframe_to_minutes, timeframe_to_msecs,
+                                                         timeframe_to_next_date,
+                                                         timeframe_to_prev_date,
+                                                         timeframe_to_resample_freq,
+                                                         timeframe_to_seconds)
```
xmatthias: Well **if** we remove them from this file - then we shouldn't import them here unless they're actually used (i doubt they are used here ... at least not all of them).

instead - imports should be adjusted accordingly if necessary - which ideally should only be in exchange.py - and eventually in exchange/__init__.py ... otherwise the imports were wrong anyway.

TheJoeSchr: on it

xmatthias (approved):
LGTM :+1:

PR #9910, opened 2024-03-04
update rpc function to include stake currency totals when fiat conver…

syco:
Hi,

simple change to display the status table with stake currency value when the fiat currency converter is not enabled, works by repurposing the fiat currency variables.


xmatthias (changes requested):
The change per se looks OK.

However, we're not accepting PR's directly to the stable branch - but rather to the develop branch (i've changed the base for this now) - which is the base for future releases.

With changed "PR base" (the trager branch)- you can see on the diff (click on the files tab above), we'd now try to merge 24 commits and change the version - which is obviously not what we're looking for (you've made 2 commits, not 24).

I think the best approach will be create a new PR replacing this one - cherry-picking your commits onto the proper branch (you can change the PR base - but not the Source branch in a github PR).

(something around the following)

```
# create a new branch with develop as source
git checkout -b rpc_update develop
# Cherry-pick the commits from your current stable branch
git cherry-pick 78a7d8f5f057986b861f13641b496d00371dfa5d
git cherry-pick ef5188d5c41726abdc1c0855a4e135a80a81f43a
# Push to remote to create a new PR

git push -u <yourupstream> rpc_update

# go to github UI, create new PR
```

alternatively - you can also try to force-push with this history to the existing PR - which should also work - but can be a bit more tricky if you're unfamiliar with git.

syco:
I missed that. Totally understandable you don't merge to stable.
I'll close this one and crrate it again on develop. I think it's the cleaner option. Thanks for taking the time to review it.

PR #9917, opened 2024-03-07
update status table to show total amounts in stake currency

syco:
Hi again,

Same change, but against the develop branch.
It changes the status table rpc to show amounts with stake currency value when the fiat currency converter is not enabled, works by repurposing the fiat currency variables.

Thanks


xmatthias (approved):
LGTM :+1:

PR #9923, opened 2024-03-10
Bump torch

xmatthias:
## Summary

Bump Torch to latest version (apparently dependabot has problems with version-restricted files) , remove constraint on some dependencies

this should enable full 3.12 support


robcaulk (approved):
Thanks a lot, glad 3.12 is easily supported finally.

PR #9941, opened 2024-03-11
Simplify the return value for full exit using adjust trade

stash86:
## Summary

Returning -trade.stake_amount will already trigger the full exit, so not sure why we need such calculation. And the old calc apparently won't result in full exit on backtest, as shown here https://discord.com/channels/700048804539400213/824719649417199646/1216713150452207636

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


docs/strategy-callbacks.md
```diff
@@ -783,7 +783,7 @@ Additional entries are ignored once you have reached the maximum amount of extra
 ### Decrease position
 
 The strategy is expected to return a negative stake_amount (in stake currency) for a partial exit.
-Returning the full owned stake at that point (based on the current price) (`-(trade.amount / trade.leverage) * current_exit_rate`) results in a full exit.
```
xmatthias, line 786: > Returning -trade.stake_amount will already trigger the full exit

Unfortunately, that's incorrect (or rather, is only correct if current_rate == entry-rate).

The check happens on `remaining` - which is calculated as `(trade.amount - amount) * current_exit_rate`.
This won't be `trade.stake_amount` necessarily - and definitely isn't if price moved ever so slightly.

https://github.com/freqtrade/freqtrade/blob/65a09b0d89f5ec313713215fafafbec1c423bf66/freqtrade/freqtradebot.py#L676-L677

I tried, but couldn't find a check which won't have accidental side-effects to support returning stake_amount directly.

We could check for `result == trade.stake_amount` - but there's too much possibility for odd edge-cases - especially once you move into the "higher win" territory - e.g. a position that opened for 100$ - has 150% profit (so it's valued at 250$). Now the user tries to sell 100$ of that as partial sell - so he's returning `-100` (take out the initial stake, let profits run further)  ... the check for "trade.stake_amount == response" would trigger - though the resulting full exit is VERY far away from what the user intended.

You can play with this yourself - simply changing the following test-case to return `-trade.stake_amount` - and you'll see what i mean (the test doesn't use 150% profit - but still shows that -stake_amount != what the docs said before).

https://github.com/freqtrade/freqtrade/blob/65a09b0d89f5ec313713215fafafbec1c423bf66/tests/freqtradebot/test_integration.py#L669-L671

Personally, i think if you need full exits - the simple solution is to go with `custom_exit()`.
However this change is wrong - unless you can also change the linked test above to correspond (and not fail) ... however without side-effects - which is the problem i encountered.

stash86: I thought remaining used to be trade.stake_amount - stake (which is returned from the adjust function?

stash86: I'll try to play with it

xmatthias: it's not - and never really was (at least not as far as i remember this part of the code - though a git blame might proof me wrong - as it's purely based on memory - not on checking😆 )

stash86:
Speaking with perk on discord (start here https://discord.com/channels/700048804539400213/824719649417199646/1216886086547017803) make me realize what I think the partial exit function do is different that what it actually do, and (in my opinion) the docs is also misleading. To summarize my talk with perk, let's use this example.

I bought 100 USDT of coins at a price of 4 USDT each. trade.stake_amount would be 100, trade.amount would be 25. Then the price of the coin will rise to 4.24 USDT (6% profit).

Based of docs
```
if current_profit > 0.05 and trade.nr_of_successful_exits == 0:
        # Take half of the profit at +5%
        return -(trade.stake_amount / 2), 'half_profit_5%'
```

My understanding of how the partial exit should work is (ofc I might be wrong, this is only from memory as well) that let's say current profit is at 6%. The bot will sell -(50/100) * 25, which is 12.5 amount of coins at 4.24 USDT price, which will give me the half profit, because I will get 53 USDT (50 USDT initial stake + 3 USDT profit. let's assume 0 fee for now).

What the bot actually do know is convert the returned stake amount to amount directly. So 50/4.24 = 11.79245283018868. I will get 50 USDT. Not sure where the `half profit` part is. Imo, the docs is misleading.

Based of memory, I believe that the initial idea of the partial exit was the same as my understanding above. And I haven't do the poll, but I would assume most users also think the same, esp since the example in the docs implying that as well.

What do you think?

Update: idea from perk is to add a third values to be returned by the adjust trade. So now it can return stake amount and partial exit tag, we can add is_quote which default to True (or maybe is_amount which default to False), so user can choose which value they actually want to use. So for this current issue, if I want to fully exit, I will write
`return -trade.amount, 'full_exit_from_adjust', False` (assuming the third params is `is_quote`)

xmatthias:
Well if you think the calculation should be changed - happy to review a PR that changes this to how you think it should be.
but i don't think **what** is returned should change (it should be stake amount, with 1 or 2 elements).

---

I'm however clearly against a 3rd parameter.
It'll lead to very confusing behavior - especially since it would be **after** the exit reason.
Think about "you write 5 partial exit logics - and on the last one - you forget the last parameter - now this last exit behaves completely different.

it'll force you to use an exit reason even if you don't need or want one (even if it's an empty string).

---

From a usage perspective - it'd be best to have "amount" for exits - but changing this would mean that entries would also use amount - which is neither intuitive, nor what we need.

while this could be split - it really really shouldn't - as then the sign defines the unit - which is implicit, and anything but intuitive.

stash86:
pushed my proposed change to the partial exit

perkmeister:
I'll add my 2p worth to this. I don't see why this would be confusing to the user, if they're adding the `is_base=True` parameter, which would default to `False` if not added, they should know what they're doing. In my case I'm running a grid bot which has to sell the exact amount of base bought at the grid below, since I can only adjust by quote value it forces me to multiply that by `current_rate` first. I'd much prefer to tell FT how much base to sell (for readabilty if nothing else). I don't understand the argument about forcing exits, that's entirely down to the strategy logic. So this enhancement would give the user the ability to change their trade position either by base amount or quote amount in a completely non-breaking way. Full exits would simply mean returning `-trade.amount` with the `is_base=True` parameter, to me that's much more intuitive than calculating it's quote value using `current_rate` first.

xmatthias:
A 3rd argument will do nothing but confuse users.
I understand it would suit your particular strategy - but i don't see any benefits aside from that (but more complexity from a user perspective, as now the 3rd argument can change the meaning of the first.

you'll end up with code like this:
```
def adjust_trade_position(se.f, ...):
    if <somecondition>: 
        return 100, 'increase'
    if <somecondition>: 
        return -200, 'decrease2x', False
    if <somecondition>: 
        return 200, 'increasex2'
    if <somecondition>: 
        return -100, 'decrease'
    if <somecondition>: 
        return -100, 'decrease',
    if <somecondition - no tag>: 
        return -100
```

It's no longer obvious what happens - one of the lines returns a completely different unit. This being the 3rd argument makes it also difficult to read / understand.

With a glance, you'll think "we increase and decrease in 100$ and 200$ steps"
but in reality, you have one condition in there, that doesn't go -200$ - but instead can mean anything between 20$ (if the price is 0.1) - or 2000$ if price was 10.

as such - i think the approach taken here is the correct one.

tests/freqtradebot/test_freqtradebot.py
```diff
@@ -5460,7 +5460,7 @@ def test_check_and_call_adjust_trade_position(mocker, default_conf_usdt, fee, ca
     assert freqtrade.strategy.adjust_trade_position.call_count == 1
 
     caplog.clear()
-    freqtrade.strategy.adjust_trade_position = MagicMock(return_value=(-10, 'partial_exit_c'))
```
xmatthias, line 5463: I wonder how this worked ... the stake-amount was never -10 something that was owned ...

xmatthias, line 5463: I wonder how this worked ... the stake-amount was never -10 something that was owned ...

perkmeister:
But you're making an assumption people are writing their exit conditions based on dollar amount adjustments. If they knew they were adusting based on base amount they'd have written them differently. For example they may want to sell 20% of their holdings when the profit rises 100% and more if it rises more, and sell everything at 400%, and quote adjusting under other conditions you'd have something like this:
```
def adjust_trade_position(se.f, ...):
    if <profit > 400%>: 
        return -trade.amount, 'sell all base',True
    if <profit i> 200%>: 
        return -trade.amount*0.3, 'decrease base 30%',True
    if <profit > 100%>: 
        return -trade.amount*0.2, 'decrease base 20%',True
    if <somecondition>: 
        return 200, 'increase quote 200'
    if <somecondition>: 
        return -100, 'decrease quote 100'
    if <somecondition>: 
        return -100, 'decrease quote 100'
    if <somecondition - no tag>: #decrease quote 100
        return -100
 ```
Why is this confusing? Your exit conditions match the type of adustment you're doing. It's non-breaking as other strategies already written don't have to change they'll be using quote adjustments which is the default.

stash86:
Take your time to see my proposed change. It would satisfy both usage. You can easily write a simple formula to return equiv stake_amount that corespond to your preferred exit amount

perkmeister:
Looks like you're calculating the amount to sell (in base) without taking into account the current rate (in fact using average price instead). That'll break current strategies which adjust trade position by quote value, and it will actually cost a different amount in quote than you're asking when the trade is placed because that will be done at current price. Unless I've misunderstood?

stash86:
Yes it's gonna be a breaking change, but a good one imo. Make the partial sell simpler and easier to understand. No one really tracking the absolute quote value of profit anyway in their straregy. You are only tracking the profit %.

perkmeister:
Does `current_profit` that's sent to `adust_trade_position()` change with sells? I'm not sure it does, it look to me like it's a partial profit ratio value calculated from `calc_profit_ratio()`.

stash86:
yes it is. current_profit is always relative to current amount only. Any realized profit won't be calculated

perkmeister:
If `current_profit` doesn`t change with sells, how could the example in the docs ever work? That would continually trigger on each bot iteration, it would quickly sell everything wouldn't it?

stash86:
No. We have the count of exit check on the example in the docs

perkmeister:
Doh! Missed that bit. Sorry!

xmatthias:
Personally, i'm fine with the changes made in this PR - as they do reflect the actual _intend_ (how it was designed to work, but apparently didn't).
I did wait for the discussion above to unfold to see different opinions quite intentionally - but i still think the way done here is correct.

I'm however also aware that potentially, this can be a breaking change if people do rely on the precise calculation for this (which i do somehow doubt - as i think this is a glich not many noticed so far since most strategies hunt for small profits - where the difference is pretty small - this only really get's relevant if you get into 20% profitable positions ...).


I don't see an "unit-changing" flag as being useful (as mentioned above already) - as the difference will be `return -trade.amount * 0.3, 'tag', True` - vs. `return -trade.stake_amount * amount, 'tag'` (working from the example above) - where i clearly prefer the clear option - where you can't mess your whole calculation up by a random, 3rd parameter.


---

As for the usefulness of the samples in the docs ... I don't think they're made to be "use as is, provides you profits" (which would assume sensible combinations) - but they should be seen as what they are - samples showing a way to accomplish things - sometimes combining things at random to show different approaches.

xmatthias (approved):
LGTM :+1:

PR #9945, opened 2024-03-12
Nitpicking documentation:

### LINKED ISSUE #9944: Documentation enhancment
## Describe your environment
(if applicable)

  * Operating system: Linux 6.7.9-zen1-1-zen
  * Python Version: 3.11.8
  * CCXT version: 4.2.51
  * Freqtrade Version: 2024.3-dev-4f28e66f9

## Describe the enhancement

This is a nitpick on the documentation. Let us enhance the following points:

- Clarify that the `stoploss_on_exchange` parameter is `order_types.stoploss_on_exchange` in all places.
  While only one `stoploss_on_exchange` exists, it may cause confusion for some people.
- Detail how changing `available_capital`. This is unclear, especially once trades have taken place.
- Clarify what happens BNB runs out with regards to fee reduction on Binance.
- Explain what PrecisionFilter does.

I'll open a small PR to this end.

GeorgiKarapetrov:
## Summary

Solve the issue: #9944

## Quick changelog

- docs/configuration.md
- docs/exchanges.md
- docs/includes/pairlists.md

## What's new?

- Clarify order_types.stoploss_on_exchange
- Detail available_capital ammendments
- Clarify BNB fee reduction when BNB runs out on Binance
- Explain PrecisionFilter

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -68,6 +68,8 @@ Binance supports [time_in_force](configuration.md#understand-order_time_in_force
 For Binance, it is suggested to add `"BNB/<STAKE>"` to your blacklist to avoid issues, unless you are willing to maintain enough extra `BNB` on the account or unless you're willing to disable using `BNB` for fees.
 Binance accounts may use `BNB` for fees, and if a trade happens to be on `BNB`, further trades may consume this position and make the initial BNB trade unsellable as the expected amount is not there anymore.
 
+If not enough `BNB` is availavle to cover transaction fees, then fees will not be covered by `BNB` and no fee reduction will occur. Freqtrade will never buy BNB to cover for fees. BNB needs to be bought and monitored manually to this end.
+
 ### Binance sites
 
 Binance has been split into 2, and users must use the correct ccxt exchange ID for their exchange, otherwise API keys are not recognized.
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -68,7 +68,7 @@ Binance supports [time_in_force](configuration.md#understand-order_time_in_force
 For Binance, it is suggested to add `"BNB/<STAKE>"` to your blacklist to avoid issues, unless you are willing to maintain enough extra `BNB` on the account or unless you're willing to disable using `BNB` for fees.
 Binance accounts may use `BNB` for fees, and if a trade happens to be on `BNB`, further trades may consume this position and make the initial BNB trade unsellable as the expected amount is not there anymore.
 
-If not enough `BNB` is availavle to cover transaction fees, then fees will not be covered by `BNB` and no fee reduction will occur. Freqtrade will never buy BNB to cover for fees. BNB needs to be bought and monitored manually to this end.
+If not enough `BNB` is available to cover transaction fees, then fees will not be covered by `BNB` and no fee reduction will occur. Freqtrade will never buy BNB to cover for fees. BNB needs to be bought and monitored manually to this end.
 
 ### Binance sites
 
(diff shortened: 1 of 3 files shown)


docs/exchanges.md
```diff
@@ -68,6 +68,8 @@ Binance supports [time_in_force](configuration.md#understand-order_time_in_force
 For Binance, it is suggested to add `"BNB/<STAKE>"` to your blacklist to avoid issues, unless you are willing to maintain enough extra `BNB` on the account or unless you're willing to disable using `BNB` for fees.
 Binance accounts may use `BNB` for fees, and if a trade happens to be on `BNB`, further trades may consume this position and make the initial BNB trade unsellable as the expected amount is not there anymore.
 
+If not enough `BNB` is availavle to cover transaction fees, then fees will not be covered by `BNB` and no fee reduction will occur. Freqtrade will never buy BNB to cover for fees. BNB needs to be bought and monitored manually to this end.
```
froggleston: `availavle` incorrectly spelled

docs/includes/pairlists.md
```diff
@@ -369,6 +369,8 @@ As this Filter uses past performance of the bot, it'll have some startup-period
 
 Filters low-value coins which would not allow setting stoplosses.
 
+Namely, pairs are blacklisted if the stop price would be the same as 99% of the stop price due to precision rounding on the exchange.
```
froggleston: Is it "the same as 99%" or "greater than 99%"? I can't remember what it is in the code...

froggleston (commented):
Just a couple of comments, but looks pretty good so far!

docs/includes/pairlists.md
GeorgiKarapetrov: The code is
```python
        # Adjust stop-prices to precision
        sp = self._exchange.price_to_precision(pair, stop_price, rounding_mode=ROUND_UP)

        stop_gap_price = self._exchange.price_to_precision(pair, stop_price * 0.99,
                                                           rounding_mode=ROUND_UP)
        logger.debug(f"{pair} - {sp} : {stop_gap_price}")

        if sp <= stop_gap_price:
```
`sp` cannot be less than `stop_gap_price`, can it? This is why I wrote it this way.

I don't understand this very well so any other clear formulation is welcome.

GeorgiKarapetrov: How about, "Namely, pairs are blacklisted if a gap of one percent or more in the stop price would be caused by precision rounding on the exchange"?

froggleston: Or "Namely, pairs are blacklisted if a variance of one percent or more in the stop price would be caused by precision rounding on the exchange, i.e. `rounded_price <= stop_price * 0.99`" ?

xmatthias: We "might" wanna add a hint on the pointlessness of this in corellation to shorts.

the initial idea was to avoid coins with a value VERY close to their lower trading boundary - not allowing setting of proper stoploss.
this can't happen on shorts ... and on longs in futures - it should in theory trigger a liquidation first ... otherwise how'd dhat coin liquidate ...? 

---

the .99 multiplication is to do "stoploss - 1%" - which is the default for stop limit orders - and gives your actual stop order "some" wiggle room to drop slightly further (stops might not fill at exactly the price they're triggered at).
not sure this is info we need in the docs though ...

GeorgiKarapetrov: I added another small commit with the hint.
I'm happy to drop the whole change to `docs/includes/pairlists.md` if it is not needed in the docs.

---

I guess that the PR will be squashed on merge, right? I don't need to squash these and rebase.

xmatthias (approved):
I'd say - LGTM :+1: 

Thanks for your efforts! :100:

PR #9955, opened 2024-03-15
Feature: trade lifecycle callbacks

Axel-CH:
## Summary
The goal of this PR is to give to the user the possibility to perform actions after trade order filled using new custom callbacks:
**"order_filled()"**

## What's new?
New custom callbacks:
"order_filled()"

Now that users can store persistent custom data, the new callbacks will allow him to set those data at critical trade life cycle and not only within the process loop.

### DIFF
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -478,7 +478,8 @@ class FreqtradeBot(LoggingMixin):
                 trade.close_date = trade.date_last_filled_utc
                 strategy_safe_wrapper(
                     self.strategy.order_filled, default_retval=None)(
-                    pair=trade.pair, trade=trade, current_time=datetime.now(timezone.utc))
+                    pair=trade.pair, trade=trade, order=order_obj,
+                    current_time=datetime.now(timezone.utc))
                 self.order_close_notify(trade, order_obj,
                                         order_obj.ft_order_side == 'stoploss',
                                         send_msg=prev_trade_state != trade.is_open)
@@ -1950,7 +1951,7 @@ class FreqtradeBot(LoggingMixin):
         if order.status in constants.NON_OPEN_EXCHANGE_STATES:
             strategy_safe_wrapper(
                 self.strategy.order_filled, default_retval=None)(
-                pair=trade.pair, trade=trade, current_time=datetime.now(timezone.utc))
+                pair=trade.pair, trade=trade, order=order, current_time=datetime.now(timezone.utc))
             # If a entry order was closed, force update on stoploss on exchange
             if order.ft_order_side == trade.entry_side:
                 trade = self.cancel_stoploss_on_exchange(trade)
--- a/freqtrade/freqtradebot.py
+++ b/freqtrade/freqtradebot.py
@@ -476,10 +476,6 @@ class FreqtradeBot(LoggingMixin):
             if not trade.is_open:
                 # Trade was just closed
                 trade.close_date = trade.date_last_filled_utc
-                strategy_safe_wrapper(
-                    self.strategy.order_filled, default_retval=None)(
-                    pair=trade.pair, trade=trade, order=order_obj,
-                    current_time=datetime.now(timezone.utc))
                 self.order_close_notify(trade, order_obj,
                                         order_obj.ft_order_side == 'stoploss',
                                         send_msg=prev_trade_state != trade.is_open)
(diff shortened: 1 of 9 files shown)


xmatthias (changes requested):
I don't immediately see what benefit these would bring if i'm honest (not the way you implemented them, anyway).

From the naming of the callbacks, i'd expect these to be triggered when the order fills - not when it's placed.


The way this PR is currently: 
For the entry order, it's showing you "i placed an order - which may or may not fill sometime in the future - or may be canceled".
That's irrelevant information, really - just increasing cycle time by pointless function calls if you ask me.
If you're about storing information - you can do the same as part of `confirm_trade_entry()` - which gives you exactly the same things.


For the exit - it's really the same - apart from the fact that the way you currently implement it - you're not guaranteed that the exit call is made for every trade that's exited.
It's not providing **any** benefit the way it is - as it's unreliable information at the wrong moment.

---

Potentially, i can see one callback at the very end of the trade (once the last order closes, closing the trade) - or one that triggers when an order fills (which could include a flag telling the callback that it's closing the trade - but you can probaly get the same info through `trade.is_open`).

with the current implementation however - i don't see benefits and will have to reject it.

Axel-CH:
Hello @xmatthias, thank you for your feedback. About the usage, I will use the first callback to store take profit prices in the entry context. The last one could be used to store custom data when the trade is closed, or send specific trade related notification.

About the timing of those callbacks you are right, There is better places to put them.
About naming / usage what do you think about:
"`trade_position_created`" and "`trade_position_closed`"
"`after_position_created`" and "`after_position_closed`"
"`position_created`" and "`position_closed`"

xmatthias:
Personally - i think there's only one callback in what what you're talking about - which is something around `order-filled`.

A callback on "order placed" doesn't make sense (the order can ba canceled - so no trade ever happened, replaced, or partially filled) - which you'll never know.

Axel-CH:
`order-filled` callback is a great idea, as it will let user decide what to do according to the trade context. Do you think it is better to create a new PR or I can amend this one?

xmatthias:
> `order-filled` callback is a great idea, as it will let user decide what to do according to the trade context. Do you think it is better to create a new PR or I can amend this one?

To be quite frank - i care about the end result - not necessarily about the PR number. :laughing: 

if we think "it's a completely new callback, in a completely different position" - it's most likely better to restart the whole PR (the current commits will place the wrong callbacks in the wrong position) - but i really, really don't care too much 
sorry for beeing not much help in this case

freqtrade/freqtradebot.py
```diff
@@ -1939,6 +1942,10 @@ def update_trade_state(
         trade = self._update_trade_after_fill(trade, order_obj)
         Trade.commit()
 
+        strategy_safe_wrapper(
```
xmatthias: You will probably want to move this to `_update_trade_after_fill` - within the first if.

Axel-CH:
@xmatthias what are you suggesting to test this callback?

xmatthias:
well not sure what you want me to tell you ...

Ensure it's called in the right places. As it's not returning anything - you can't really double-check that part - but you'll still want to ensure it's actually called (the appropriate amount of times) - either by mocking it and using call_count - or by providing a dummy function and setting something on that trade - verifying that's actually set correctly.

Axel-CH:
Testing live, working fine for entry but exits work only in case of emergency exit, will check tomorrow if we need to move the exit case cb.

Axel-CH:
Tested live, and both entry and stoploss 'order_filled' callback are working. Just want to validate that the 'order' parameter provided is corresponding to filled order for normal and stoploss orders.

freqtrade/freqtradebot.py
```diff
@@ -476,6 +476,10 @@ def handle_onexchange_order(self, trade: Trade):
             if not trade.is_open:
                 # Trade was just closed
                 trade.close_date = trade.date_last_filled_utc
+                strategy_safe_wrapper(
+                    self.strategy.order_filled, default_retval=None)(
+                    pair=trade.pair, trade=trade, order=order_obj,
+                    current_time=datetime.now(timezone.utc))
```
xmatthias: I think if you happen to run into this case - then you'll trigger the callback twice / multiple times.

The above "refind" logic will trigger the callback for every (new) order it finds (essentially orders where you manually sold assets on exchange).

If that triggers a trade close - the callback will trigger again - though in that case, there's no "acual" new order fill you can attach - so it'll trigger twice for the very last order.

xmatthias (approved):
LGTM :+1:

PR #9961, opened 2024-03-17
tiny bugfix for lookahead-analysis

hippocritical:
Fixed bugs for lookahead-analysis:
The command fails if trades were not defined inside the config (but in the strategy)
=> change check to .get and adjust logger-comment
The command produces false positives if protections are enabled and they work globally (since the verification checks only work with a single pair)
=> solution: disable all protections (they would just slow down the process and they are not being checked for lookahead anyways)


xmatthias (approved):
LGTM :+1:

PR #9989, opened 2024-03-22
delist schedule

### LINKED ISSUE #9985: ExchangeDelistFilter Feature
## Describe the enhancement

Hey Guys I have written a delist filter for binance 
https://binance-docs.github.io/apidocs/spot/en/#get-symbols-delist-schedule-for-spot-market_data

Not sure if it makes sense or if it actually will help but maybe you think about this and maybe this worthy to take into the mainline
It requires authentication so it might only run in not dry mode.

```python
"""
Exchange Delist filter
"""
import logging
from typing import Any, Dict, List

import pandas as pd

from freqtrade.constants import Config
from freqtrade.exchange.types import Tickers
from freqtrade.persistence import Trade
from freqtrade.plugins.pairlist.IPairList import IPairList, PairlistParameter


class ExchangeDelistFilter(IPairList):

    def __init__(self, exchange, pairlistmanager,
                 config: Config, pairlistconfig: Dict[str, Any],
                 pairlist_pos: int) -> None:

        self._refresh_period = self._pairlistconfig.get('refresh_period', 1800)




    @property
    def needstickers(self) -> bool:
        """
        Boolean property defining if tickers are necessary.
        If no Pairlist requires tickers, an empty List is passed
        as tickers argument to filter_pairlist
        """
        return False

    def short_desc(self) -> str:
        """
        Short allowlist method description - used for startup-messages
        """
        return f"{self.name} - Removing Pairs when delisting."

    @staticmethod
    def description() -> str:
        return "Filter pairs by performance."

    @staticmethod
    def available_parameters() -> Dict[str, PairlistParameter]:
        return {
            **IPairList.refresh_period_parameter(),
        }

    def filter_pairlist(self, pairlist: List[str], tickers: Tickers) -> List[str]:
        """
        Filters and sorts pairlist and returns the allowlist again.
        Called on each bot iteration - please use internal caching if necessary
        :param pairlist: pairlist to filter or sort
        :param tickers: Tickers (from exchange.get_tickers). May be cached.
        :return: new allowlist
        """
        # Get the trading performance for pairs from database


        delistings = self._exchange._api.sapiGetSpotDelistSchedule()
        delisted_pairs = set(sum([d['symbols'] for d in delistings], []))

        for pair in pairlist:
            pair_in_delisting_format = pair.replace('/', "")
            if pair_in_delisting_format in delisted_pairs:
                pairlist.remove(pair)
                self.log_once(f"Removed {pair} from whitelist due to delisting", logger.info)


        return pairlist
````

### PR DESCRIPTION
**Pull request by jakubikan**
## Summary

This will implement exchange function to check whether a pair is delisting 
If it will be delisting it will return the delisting time.

Solve the issue: [9985](https://github.com/freqtrade/freqtrade/issues/9985)

Probably this need some more to have this working in the strategy with the `confirm_trade_entry`
or `custom_trade_exit`

but not sure yet - requesting some review if thats the right way to go 

One thing I'm not sure of yet this will work in the end
Because essentially - what if binance does not activate have the delist schedule api entry with the announcements of the announcements page.
Because when there was the last announcement I looked that the pairs in the Binance app and they have not been marked to be delisted yet. Might als have been a caching issue or something - but not sure 
i guess this can only be really tested with next announcement :smile: 

## Quick changelog

- <change log 1>
- <change log 1>

### DIFF
--- a/freqtrade/exchange/binance.py
+++ b/freqtrade/exchange/binance.py
@@ -217,3 +217,44 @@ class Binance(Exchange):
                     raise OperationalException(e) from e
         else:
             return {}
+
+    def get_spot_delist_schedule(self, refresh: bool) -> list:
+        """
+        Calculates bid/ask target
+        bid rate - between current ask price and last price
+        ask rate - either using ticker bid or first bid based on orderbook
+        or remain static in any other case since it's not updating.
+        :param pair: Pair to get rate for
+        :param refresh: allow cached data
+        :param side: "buy" or "sell"
+        :return: float: Price
+        :raises PricingError if orderbook price could not be determined.
+        """
+        if not refresh:
+            with self._cache_lock:
+                rate = cache_rate.get(pair)
+            # Check if cache has been invalidated
+            if rate:
+                logger.debug(f"Using cached {side} rate for {pair}.")
+                return rate
+
+
+        if conf_strategy.get('use_order_book', False):
+
+            order_book_top = conf_strategy.get('order_book_top', 1)
+            if order_book is None:
+                order_book = self.fetch_l2_order_book(pair, order_book_top)
+            rate = self._get_rate_from_ob(pair, side, order_book, name, price_side,
+                                          order_book_top)
+        else:
+            logger.debug(f"Using Last {price_side.capitalize()} / Last Price")
+            if ticker is None:
+                ticker = self.fetch_ticker(pair)
+            rate = self._get_rate_from_ticker(side, ticker, conf_strategy, price_side)
+
+        if rate is None:
+            raise PricingError(f"{name}-Rate for {pair} was empty.")
+        with self._cache_lock:
+            cache_rate[pair] = rate
+
+        return delistings
--- a/freqtrade/exchange/binance.py
+++ b/freqtrade/exchange/binance.py
@@ -5,7 +5,7 @@ from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
 import ccxt
-from cachetools import TTLCache
+from cachetools import TTLCache, cached
 from threading import Lock
 
 from freqtrade.enums import CandleType, MarginMode, PriceType, TradingMode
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: freqtrade/exchange/exchange.py [hunk 1]
```diff
@@ -187,6 +187,11 @@ def __init__(self, config: Config, *, exchange_config: Optional[ExchangeConfig]
         self._api_async = self._init_ccxt(
             exchange_conf, ccxt_async, ccxt_kwargs=ccxt_async_config)
 
+        if exchange_conf.get('sandboxMode'):
+            self._api.set_sandbox_mode(True)
+            self._api_async.set_sandbox_mode(True)
```
jakubikan: it seems that the delist endpoint is not available on the the testnet 

but not sure if we could keep this since there are quite some cex's out there that allow papertrading

stash86: why don't you use the actual server instead of the testnet? FT has been discouraging the use of testnet because of the limited features in it.

jakubikan: the problem that I'm having is that the delist endpoint is actually an authenticated endpoint 

In dry_mode this will fail since we are not using the authentication there

xmatthias: it's not the first (and won't be the last) feature that ain't fully supported in dry-run mode. 

it's also really a topic of "it's an authenticated endpoint on binance" - if you look at the same / corresponding endpoint on bybit - then it's not authenticated.

freqtrade/exchange/binance.py
```diff
@@ -217,3 +224,58 @@ def load_leverage_tiers(self) -> Dict[str, List[Dict]]:
                     raise OperationalException(e) from e
         else:
             return {}
+
+
+
+    @retrier
+    @cached(cache=TTLCache(maxsize=100, ttl=10), lock=Lock())
```
xmatthias, line 231: you should take your pick, really, when it comes to caching.

actual caching is applied below (in the `get_spot_pair_delist_time()`.
applying caching twice can be dangerous here - as then you risk having the TTL applied twice in unfortunate cases.

That decorator is also something we're not using anywhere else - as it won't allow you to work around the cache and force a refresh - which can be necessary in some instances.

freqtrade/exchange/binance.py
```diff
@@ -217,3 +224,58 @@ def load_leverage_tiers(self) -> Dict[str, List[Dict]]:
...
+    def get_deslist_schedule(self):
+        try:
+
+            delist_schedule = self._api.sapi_get_spot_delist_schedule()
```
xmatthias, line 235: considering that this is a private (direct) endpoint - i think we should have a form to unify it. Usually, that's taken care of by ccxt, but in this case, that unification / abstraction isn't available (yet?).


i know that for example bybit has a similar endpoint (we need to parse the text to make that work) - but i think what the exchange returns should be unified across exchanges (across the exchanges that support it, anyway) - so it's easier to use from a consumer side (not necessarily the strategy - anything outside of the exchange section should be seen as a consumer of this function).

freqtrade/exchange/binance.py
```diff
@@ -217,3 +224,58 @@ def load_leverage_tiers(self) -> Dict[str, List[Dict]]:
...
+        :return: int: delisting time None if not delisting
+        """
+
+        if not pair:
+            return
+
+        cache = self._spot_delist_schedule_cache
+        lock = self._spot_delist_schedule_cache_lock
+
+        schedule_pair = pair.replace('/', '')
```
xmatthias, line 260: ccxt's `.markets` structure should have everything we need to convert symbol to pair (or viceversa).
i also think that the endpoint should be casting to pair already - so we get "DERP/BTC" out of the call - not DERPBTC. simply removing "/" is rather prone to errors in my experience.

it's often best to look at ccxt code on how they do it internally...

jakubikan:
Hey Guys thanks for the input - I will try to continue with it after the east holiday. 
since I have just some limited time online. 
Happy Easter 🐰 😄

freqtrade/exchange/binance.py
stash86: this is what the endpoint give

```
[{
        "delistTime": "1759114800000",
        "symbols": [
            "OMNIBTC",
            "OMNIFDUSD",
            "OMNITRY",
            "OMNIUSDC",
            "OMNIUSDT"
        ]
    }]
```

xmatthias:
Closing this in favor of #12259 - which will provide essentially the same functionality - but is evolved further.

PR #10001, opened 2024-03-25
Update PyTorchModelTrainer.py

robcaulk:
The n_epochs should be defined using the `max` not the `min` function.

## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


xmatthias (approved):
LGTM :+1:

Issue #10009, opened 2024-03-27
Add `max_value` field for VolumePair list


stevanStevic:
## Describe the enhancement

Similar to `min_value` we have for [VolumePair list](https://www.freqtrade.io/en/stable/plugins/#volume-pair-list) I would like to add `max_value` to have upper limit on volume in which bot can trade.

xmatthias:
Sounds like a good first issue ... should be pretty simple to do - and scope is very limited.

ststevic:
> Sounds like a good first issue ... should be pretty simple to do - and scope is very limited.

I'll add it 👍 Wanted to check if okay with you!

xmatthias:
well, no worries, this is what i'd usually consider a simple enhancement :+1:

PR #10011, opened 2024-03-28
fix:Add max_value field for VolumePair list #10009

Dibyanshuaman:
## Summary

Solve the issue: #___

## Quick changelog

- <change log 1>
- <change log 1>


froggleston (changes requested):
This code only adds to the available parameters as a field and doesn't actually implement the feature as far as I can see?

Also, please fill in the PR template with the required information. It's useful to get a summary of the changes in the PR text without looking through code.

xmatthias:
I'll be closing this in favor of #10013 - which provides a complete implementation of this - not just a setting for the webUI that does nothing further. 

anyway thanks for your efforts :+1:

PR #10013, opened 2024-03-28
10009: Add `max_value` to volume pair list

### LINKED ISSUE #10009: Add `max_value` field for VolumePair list

stevanStevic:
## Summary

Solve the issue: #10009 

## Quick changelog

- Add support for `max_value` in VolumePair list which will filter out pairs with quote volume higher than specified.


xmatthias (approved):
Great! LGTM :+1:

PR #10016, opened 2024-03-29
Add support for requests pool size configuration

froggleston:
## Summary

This PR adds support for configuration of the requests session pool size.

## Quick changelog

- add constructor and configuration for pool connection and maxsize properties

## What's new?

When running many bots, it can be easy to saturate the requests session pool. This PR introduces a non-breaking change to allow end users to configure the pool connection and size to avoid requests complaining about socket reuse. If users are running many bots on the same host, increasing maxsize helps with this. Most users would not need to use this feature, but with 10 or 20 bots, you can easily run into the issue of the pool becoming full.

### DIFF
--- a/ft_client/freqtrade_client/ft_rest_client.py
+++ b/ft_client/freqtrade_client/ft_rest_client.py
@@ -21,10 +21,19 @@ logger = logging.getLogger("ft_rest_client")
 
 class FtRestClient:
 
-    def __init__(self, serverurl, username=None, password=None):
+    def __init__(self, serverurl, username=None, password=None,
+                 pool_connections=10, pool_maxsize=10):
 
         self._serverurl = serverurl
         self._session = requests.Session()
+
+        # allow configuration of pool
+        adapter = requests.adapters.HTTPAdapter(
+            pool_connections=pool_connections,
+            pool_maxsize=pool_maxsize
+        )
+        self._session.mount('http://', adapter)
+
         self._session.auth = (username, password)
 
     def _call(self, method, apipath, params: Optional[dict] = None, data=None, files=None):


xmatthias:
Help me understand this ... 
assuming this is for a TUI ... if you have multiple bots - you'll also need multiple FtRestClient classes (assuming different url's / different username / passwords.

Why would you then need to increase the pool size? or would you decrease this - so it's not keeping a too large pool per instance?

froggleston:
As far as I'm aware, requests shares the same pool underneath for the same host, not based on the calling client.

(Yes, you need multiple FtRestClient clients - I run 20+ bots on the same IP so requests manages a single pool for that host)

xmatthias:
> As far as I'm aware, requests shares the same pool underneath for the same host, not based on the calling client.

i think this only applies when you're reusing the same session ... which is not the case (we create a new session each time).
(at least that's how i'm reading the [documentation](https://requests.readthedocs.io/en/latest/user/advanced/#session-objects) - more precisely this section:

> The Session object allows you to persist certain parameters across requests. It also persists cookies across all requests made from the Session instance, and will use urllib3’s [connection pooling](https://urllib3.readthedocs.io/en/latest/reference/urllib3.connectionpool.html). So if you’re making several requests to the same host, the underlying TCP connection will be reused, which can result in a significant performance increase (see [HTTP persistent connection](https://en.wikipedia.org/wiki/HTTP_persistent_connection)).

The code kinda confirms this ... creating a new adapter (which in turn creates a new PoolManager) for every Session.

https://github.com/psf/requests/blob/8dd3b26bf59808de24fd654699f592abf6de581e/src/requests/sessions.py#L448-L449

Now don't get me wrong - that doesn't mean it's a problem ... just trying to understand what we're trying to accomplish - and which direction it should be tuned eventually.

froggleston:
Yes but you only have one Session per host as far as I can see? You get one _reused_ adapter for, say, http://foo and one for http://bar, etc etc? (Or rather multiple sessions, but reusing the same adapter to the pool...)


(Basically, I get pool issues when running many bots on one IP, and if I set maxsize to 100, it goes away) 😄

froggleston:
OK from reading around some more, I guess the issue is that in the FTUI I've got a Session per FtRestClient, but I'm making concurrent requests from multiple threads using that same Session, hence requiring a larger maxsize.

xmatthias:
Well you're assuming default session handling - but `FtRestClient` creates a a custom session - and i can't find any hint (neither in docs nor in code) that this would be the case though - which concerns me a bit - as either means that you're messing with class internals (overriding the session after init) ... 

**edit** Requests (or rather, the session object) is not necessarily thread-safe (at least as far as i'm aware) ... so i suspect you can run into other problems, too.

froggleston:
No, I'm not overriding the Session or messing with anything else. I'm using worker threads in the TUI so am producing a higher concurrency above the default 10 provided by the default Session config. As such, I need to increase the maxsize, and without this PR, I have to add the code I provided in manually. The code I provided is exactly how Session objects are modified to allow different pool sizes etc.

I'm aware that requests isn't thread-safe and other implementations (aiohttp?) might be better, but that would need a whole new FtRestClient implementation 😃

xmatthias (approved):
Let's merge this for now.

Before fully merging to develop, i'll probably convert the parameters to kw only parameters - as that makes it easier to deprecate / remove them in the future - should the implementation details change (like, switching from requests to httpx or something else).

froggleston:
OK, makes sense, thank you!

PR #10041, opened 2024-04-02
docs: created a 404 documentation page

samgermain:
## Summary

Solve the issue: [css doesn't load for 404 pages on the docs](https://github.com/ccxt/ccxt/issues/22032)

## Quick changelog

- adds 404.html
- adds 404.html to static templates of mkdocs.yml

## What's new?

I hope that this issue fixes the confusing 404 pages for the docs, but I don't really know, I couldn't test it, so I'm just guessing that something like this might fix the problem. I created the solution based on [this github post ](https://github.com/squidfunk/mkdocs-material/discussions/5018)

I don't really know how to make changes with mkdocs and don't really know how to test this out, I tried to contribute the best I could though

### DIFF
--- a/docs/overrides/404.html
+++ b/docs/overrides/404.html
@@ -1,0 +1,44 @@
+{% extends "base.html" %}
+
+
+<!-- Navigation -->
+{% block site_nav %}
+
+  <!-- Main navigation -->
+  {% if nav %}
+  {% if page and page.meta and page.meta.hide %}
+  {% set hidden = "hidden" if "navigation" in page.meta.hide %}
+  {% endif %}
+  <div class="md-sidebar md-sidebar--primary" data-md-component="sidebar" data-md-type="navigation" {{ hidden }}>
+    <div class="md-sidebar__scrollwrap">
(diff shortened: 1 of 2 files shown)


xmatthias:
it's a nice find - though i don't think it's the correct fix.
testing this locally renders the page just fine also on develop

image

It's a problem of readthedocs - where it doesn't find any assets - which essentially means the same will happen on any random 404 page, too - unless it brings all assets itself (which i wouldn't do - as then it's sure to break the theme, if not now, then soon in the future).

There's most likely other solutions to this - as i don't find the "current" 404 page that bad - if it renders, anyway.

xmatthias:
Going to close this as it's not the right approach (mkdocs does build a 404.html page out of the box already)  - which can be tested / shown via `mkdocs build && ls -l site/`.

error pages now work since 569a87ca283c4cb26c4826a90d1b3ec7f69c9285 - but it'll probably not properly work for "older" versions ... as these won't be rebuilt and would require an explicit push to each tag.

to test: https://www.freqtrade.io/en/latest/asdfasdf

PR #10050, opened 2024-04-06
FreqAI pandas futures warnings

robcaulk:
This PR aims to prepare for the 3.0 pandas bump.

### DIFF
--- a/freqtrade/freqai/data_kitchen.py
+++ b/freqtrade/freqai/data_kitchen.py
@@ -24,6 +24,8 @@ from freqtrade.strategy import merge_informative_pair
 from freqtrade.strategy.interface import IStrategy
 
 
+pd.set_option('future.no_silent_downcasting', True)
+
 SECONDS_IN_DAY = 86400
 SECONDS_IN_HOUR = 3600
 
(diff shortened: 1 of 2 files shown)


freqtrade/freqai/data_kitchen.py
```diff
@@ -24,6 +24,8 @@
 from freqtrade.strategy.interface import IStrategy
 
 
+pd.set_option('future.no_silent_downcasting', True)
```
xmatthias, line 27: You sure this should've been commited?

that option itself seems deprecated ... (https://github.com/pandas-dev/pandas/issues/57403) ... or at least the functionality behind this, and i don't think we'll use the functionality (it'd be the `downcast` parameter to functions like fillna).

robcaulk: The .replace that we are fixing, uses the silent downcasting. We added the .infer_object to indicate that we want to keep the old behavior for this function moving forward.

So we are happy with behavior and will be happy when 3.0 rolls around. So this setting means we want to operate in that mode now. Which avoids warnings.

xmatthias (approved):
LGTM :+1:

